Issue #1219, opened 2026-07-16
release-please cuts releases for test-only changes (config has no src/main path filter)


JiRaska:
### Affected service / module

`release-please-config.json` (fleet-wide — all 50 registered components)

### Version / commit

main @ 6941e6a8

### What happened?

A **test-only** change cut real releases for four services.

[#1166](https://github.com/JiRaska/open-bank-oss/pull/1166) (`fix(party-service): restore @PactBroker on provider verification`) touched only `src/test/**` files, but in four different modules:

```
openbank-account-service/src/test/kotlin/com/openbank/account/contract/PartyCreatedMessagePactConsumerTest.kt
openbank-kyc-service/src/test/kotlin/com/openbank/kyc/contract/PartyEventMessagePactConsumerTest.kt
openbank-onboarding-service/src/test/kotlin/com/openbank/onboarding/contract/PartyServicePactConsumerTest.kt
openbank-party-service/src/test/kotlin/com/openbank/party/contract/PartyEventPactProviderVerificationTest.kt
```

release-please then opened four release PRs — #1173 (onboarding 0.5.8), #1172 (party 0.9.1), #1170 (kyc 0.10.2), #1171 (account 0.15.2) — each bumping `version.txt` and appending a CHANGELOG entry.

Because each component's changelog quotes the triggering commit verbatim, the entries misattribute the work. `openbank-account-service` 0.15.2 ships this note:

```
### Bug Fixes
* **party-service:** restore @PactBroker on provider verification (unblocks auto-deploy) (#1166)
```

So account-service — a money-path service — publishes a release claiming a fix it never received. No product code changed in any of the four.

### What did you expect?

Per [`CLAUDE.md`](CLAUDE.md):

> A change under `<service>/src/main/**` is released by **release-please** from your Conventional Commit

A change confined to `src/test/**` should cut **no** release. Only party-service was in scope, and arguably nothing should have been released at all.

### Reproduction steps

1. Open a PR changing only a file under `openbank-<svc>/src/test/**` for any registered component.
2. Merge it with a `fix(...)` or `feat(...)` commit.
3. Observe release-please open a release PR for that component.

### Root cause

`release-please-config.json` has no path filter. Every package entry is bare:

```json
"openbank-account-service": { "component": "account-service" },
"openbank-party-service":   { "component": "party-service" },
```

[...]

PR #1195, opened 2026-07-16
feat(vop): add Verification of Payee backend; give control-plane agents episodic memory

JiRaska:
## Summary

Closes the platform audit's most urgent finding: **Verification of Payee had no backend, and the admin UI gated SCT Inst submission on `Math.random()`** — a control that has been legally mandatory since 2025-10-09 and that the platform only appeared to enforce. Adds a real `openbank-vop-service`, gives the six control-plane agents episodic memory on Postgres, and reconciles a handful of documentation/config statements that had drifted away from the code.

## Type of change

- [x] feat (new functionality)
- [x] fix (bug fix)
- [x] docs
- [ ] refactor (no behavior change)
- [ ] perf (performance)
- [ ] chore (build / tooling)
- [ ] security (private until disclosed)
- [ ] breaking change

## Linked issues

Closes #1193 (VoP backend)
Closes #1194 (agent findings → Postgres)
Refs ADR-0171, ADR-0003, ADR-0050, ADR-0118, ADR-0159, ADR-0136

## Changes

**VoP (`openbank-vop-service`, new, money-path)**
- `VopNameMatchPolicy` — pure domain, four EPC outcomes (`match`/`close_match`/`no_match`/`no_data`). Normalises via libs' `MatchKey.normalize` rather than adding a third copy of that normaliser. `CLOSE_MATCH` covers token reordering, initials, legal-form suffixes and a one-character typo.
- Responder side is real: IBAN → account-service (`partyId`) → party-service (`legal_name`/`trading_name`). Requester side is honest rather than faked — no EPC scheme link, so an external IBAN returns `no_data`/`NO_SCHEME_CONNECTIVITY` through a `VopSchemeRoutingPort` seam, mirroring how the rails reach only the clearing simulator.
- Per-requester Valkey rate limit (60/min), OPA bundle + generator, CNPG cluster (`instances: 2`, ADR-0159), Rollout with OPA sidecar and canary, threat model, ADR-0171.
- Admin UI now calls the real backend; the `setTimeout` + `Math.random()` mock is gone.

**Agent findings → Postgres (6 services)**
- `control-liveness-sentinel`, `governance-auditor`, `release-steward`, `docs-truth-agent`, `authz-policy-auditor`, `flaky-test-hunter` converted to the devops-agent pattern: reactive Panache entity + `PostgresFindingRepository`, Flyway V1, CNPG cluster, Testcontainers boot smoke IT. Ports unchanged — adapter swap only.
[...]

PR #1295, opened 2026-07-16
docs: state what actually triggers a release-please release

JiRaska:
## Problem

[`CLAUDE.md`](CLAUDE.md) promises:

> A change under `<service>/src/main/**` is released by **release-please** from your Conventional Commit

That is not what happens, and it cannot be made to happen.

release-please has **no include/allow-list**. The only path mechanism is `exclude-paths`, and per [`commit-exclude.ts`](https://github.com/googleapis/release-please/blob/main/src/util/commit-exclude.ts) it matches a **directory prefix**:

```ts
private isRelevant(file: string, path: string) {
  return path === ROOT_PROJECT_PATH || file.indexOf(`${path}/`) === 0;
}
```

An entry pointing at a file (`openbank-devops-agent/governance.yaml`) is compared as `openbank-devops-agent/governance.yaml/` and never matches — it **fails silently**: config parses, CI stays green, the bump still happens. `normalizePaths()` only trims slashes; it does not help.

So "only `src/main` releases" is inexpressible. Today **318 files across all 50 packages** sit at a package root and every one of them cuts a release:

```
governance.yaml  Dockerfile  build.gradle.kts  detekt-baseline.xml
ktlint-baseline.xml  CLAUDE.md  LICENSE  gradlew  settings.gradle.kts  …
```

Touching a lint baseline ships a version of the service.

## Two live examples

- **account-service 0.15.2** shipped crediting `**party-service:** restore @PactBroker…` — [#1166](https://github.com/JiRaska/open-bank-oss/pull/1166) was test-only across four services and cut four releases ([#1219](https://github.com/JiRaska/open-bank-oss/issues/1219)).
- **devops-agent 0.3.2 → 0.4.0** (a *minor*) for a single `governance.yaml` edit in [#1195](https://github.com/JiRaska/open-bank-oss/pull/1195), captioned *"Features: add Verification of Payee backend; give control-plane agents episodic memory"* — a feature it never received.

## Change

Rewrites the rule to describe the tool we actually have: any file under `<service>/` releases it, except that package's `exclude-paths` (`src/test` once [#1277](https://github.com/JiRaska/open-bank-oss/pull/1277) lands), and notes explicitly that a single file cannot be excluded — so nobody tries and quietly gets nothing.

Docs-only. No behavior change. +8/-3 lines; CLAUDE.md stays lean at 240 lines.

## Why not "fix it properly" instead

The alternative was moving those 318 root files into excludable directories (`governance/governance.yaml` etc.) across 50 packages, and chasing every consumer of those paths — CI scripts, the governance manifest, Docker builds. That is a large, risky change driven by a tool limitation rather than by design. Rejected in favour of honest docs; reopen the question if the release noise gets worse.

## Test plan

- [x] Claim verified against release-please source, not inferred: `isRelevant()` requires a `/` boundary; `normalizePaths()` only trims slashes
- [x] Replayed the matching rule on real data — file-exclude → BUMP (no effect), directory-exclude → SKIP
- [x] 318 root files across 50 packages counted from the working tree
- [x] Markdown only; no config or code touched

Refs #1219, #1277

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -14,9 +14,14 @@ Before a change is "done", run the ship-checklist (`/ship-check` runs the same c
 ADR-0029):
 
 1. **Open a PR.** No direct commits to `main`. Branch `<type>/<scope>-<summary>`; squash-merge via PR.
-2. **Versioning is automatic.** A change under `<service>/src/main/**` is released by **release-please**
-   from your Conventional Commit — so the commit message *is* the changelog. Do **not** hand-edit
-   `version.txt`, `CHANGELOG.md`, or `quarkus.application.version` (it derives from `version.txt`).
+2. **Versioning is automatic.** A change to **any file under `<service>/`** — except the paths listed
+   in that package's `exclude-paths` (currently `src/test`) — is released by **release-please** from
+   your Conventional Commit, so the commit message *is* the changelog. That includes `governance.yaml`,
+   `Dockerfile`, `build.gradle.kts` and the lint baselines: touch one and the service ships a version.
+   release-please offers no include/allow-list (only `exclude-paths`, and only for **directories** —
+   a single file cannot be excluded), so "only `src/main` releases" is not expressible; treat any
+   package-root edit as release-triggering. Do **not** hand-edit `version.txt`, `CHANGELOG.md`, or
+   `quarkus.application.version` (it derives from `version.txt`).
    A module is a released component **iff** it has a `version.txt` (registered in
    [`release-please-config.json`](release-please-config.json) + [`.release-please-manifest.json`](.release-please-manifest.json)).
 3. **API change ⇒ `openapi.yaml` updated + contract test.** Two independent version axes (ADR-0048):


JiRaska:
Closing: superseded by #1320 (`ca36e4cc3`), which rewrote this exact paragraph.

**Your two accurate claims are on `main` — I kept them verbatim in substance:**
- release-please has no include/allow-list, only `exclude-paths`
- exclusion matches **directories** only, so a single file cannot be excluded

I verified both against the v17.x source before reusing them. Nothing of this PR's value is lost.

**What I could not keep** is *"touch one and the service ships a version"*. It asserts the path axis alone triggers a release, and that is false — the commit **type** gates it. A hidden type (`refactor` `docs` `test` `build` `ci` `chore`) renders an empty changelog body, so `changelogEmpty()` returns `undefined` and no release PR is cut, at any path. Verified two ways: by running `release-please@17.10.3`'s `DefaultChangelogNotes` against this repo's real `changelog-sections` (chore-only → SKIP, fix-only → RELEASE), and by a live control — `chore(admin-ui): deploy …` commits have touched `openbank-admin-ui/` since tag `admin-ui-v0.49.0`, release-please has run green since, and no admin-ui release PR exists.

The devops-agent 0.3.2→**0.4.0** example cuts the other way too: a *minor* is only ever produced by a `feat`, so that case shows the type deciding, not the path.

**Why this matters enough to close rather than leave open:** now that #1320 has landed, merging this would *revert* the correction and put the false claim back into the file every contributor and agent reads. An open PR that undoes a merged fix is a trap for whoever drains the queue next — and it nearly caught me an hour ago, when the instruction was "review and merge everything".

`main` now states the rule with both axes:

> release ⟺ (type ∈ {feat, fix, perf, security} **or breaking**) **AND** (a changed file under `<service>/` outside that package's `exclude-paths`)

#1320 also fixed `RELEASE.md:49`, which #1277 had made false an hour earlier ("release-please has no concept of a sub-path exclusion"). That was the docs/config disagreement this PR set out to end — it is ended, just from the other side.

Reopen if you think I've got the type axis wrong; the evidence above is what to attack.

Issue #938, opened 2026-07-12
OPA enforcement bootstrap: anacredit-service, card-issuance-service, interest-service have none


JiRaska:
## What

Split out of #750: while auditing every fleet-wide resourceless M2M `.create` action for a
missing authz rule, three services turned out to have **no OPA enforcement infrastructure at
all**, not just a missing rule:

- `openbank-anacredit-service`
- `openbank-card-issuance-service`
- `openbank-interest-service`

None of them has an `opa-bundle.yaml` ConfigMap, a `gen-*-opa-bundle.sh` generator script, or
any `authz`/`opa.` configuration in `application.yaml`. Compare every other service in
`openbank-infra/gitops/components/*/gen-*-opa-bundle.sh` (accounts, balances, billing, consent,
copilot, customer-edge, fraud-service, fx-service, ledger, lending, notifications, payments/*,
pid, sca) — all 20 have one.

## Why this is a different shape of task than #750

#750's pattern (and #746 before it) was "add one narrow M2M carve-out rule to an *existing*
bundle" — verify the caller, add an `allowed_reasons contains "..." if {...}` block, add
`rest_test.rego` coverage, regenerate one ConfigMap. Bootstrapping a service from zero means:

1. Confirm `@Authorize` annotations are actually present on the service's REST resources (or
   add them if not — check first, don't assume).
2. Author a new `<service>-opa-bundle.yaml` + `gen-<service>-opa-bundle.sh` following the
   established pattern (mounts the shared `rest.rego` base + a service-specific extension).
3. Work out the actual authorization shape per action (which roles, which verified M2M callers
   if any) the same case-by-case way #746/#750 did — do not cargo-cult a blanket grant.
4. Wire `AUTHZ_ENFORCE` into the service's deployment (advisory-mode first, per the fleet's
   established rollout convention — ADR-0034 Phase 5 precedent).
5. Threat-model touch if any of the three is money-path (check `rules.yaml:
   money_path_services`).

## Task

For each of the three services, independently:
- Audit real callers (REST clients / NetworkPolicy ingress allow-list) the way #750's
  settlement.create / lending.create / fx.create analysis did.
- Bootstrap the bundle + generator script.
- Land it in advisory mode first (`AUTHZ_ENFORCE=false`, `advisory: would DENY` logging) before
  flipping to enforce, matching how every other service in this fleet was rolled out.

## Needs ADR

No — extends the existing ADR-0034 OPA rollout to services that were missed, not a new
[...]

Issue #750, opened 2026-07-11
OPA rest.rego: fleet-wide resourceless M2M .create actions have the same authz gap as #746


JiRaska:
## What

Fixing #746 (M2M `sanctions.create` had no `allowed_reasons` rule authorizing it — `operator-on-own-tenant` requires `input.resource`, which a resourceless create action never has) surfaced that this is a **fleet-wide pattern**, not unique to sanctions-service.

A grep for the same shape (`@Authorize(action = "*.create", resource = "")`) across the fleet finds at least:

```
openbank-anacredit-service       anacredit.create
openbank-card-issuance-service   card.create
openbank-fx-service               fx.create
openbank-interest-service         interest.create  (x2)
openbank-ledger-service           ledger.create
openbank-lending-service          lending.create
openbank-notification-service     device.create
openbank-settlement-service       settlement.create
```

`ledger.create` is confirmed live-hitting this exact gap today: ledger-service's boot log shows `advisory: would DENY action=ledger.create resource=null principal=service-account-openbank-services reason=unspecified — proceeding (enforce=false)` on every transaction-service posting — masked only because ledger-service also runs with `AUTHZ_ENFORCE=false` in the real deployment, same as sanctions-service.

## Why #746's fix doesn't cover this

#746's `m2m-sanctions-screening` rule is deliberately scoped to the single `sanctions.create` action, not a generic "any M2M create" grant — because the right authorization shape differs per action:
- `sanctions.create` (run a compliance check) is safe for any M2M service-account caller.
- Some of the `.create` actions above might be more sensitive (e.g. does `card.create` issuing a new card need tighter caller-identity scoping? does `lending.create` disbursing a loan?) and need the same case-by-case analysis #746 did, not a blanket rule.

## Task

For each action in the list above (and any others the grep pattern below turns up — re-run it, this list may be incomplete):
1. Confirm which M2M identity actually calls it (check the calling service's REST client / `@RegisterProvider(OidcClientRequestReactiveFilter::class)` usage) and whether it's currently masked by `AUTHZ_ENFORCE=false` in that service's gitops deployment (check for a live `advisory: would DENY` log line, same as ledger.create above).
2. Decide per-action whether an `m2m-*`-style narrow rule (like #746's) is appropriate, or whether the action needs to stay human-only / needs a different design.
[...]

PR #1187, opened 2026-07-16
ci(opa): discover OPA bundle generators dynamically instead of hard-coding four

JiRaska:
## What

Stage 3 of 3 for #1184. Replaces the hard-coded generator list in `opa-policy.yml` with dynamic
discovery, and corrects the stale `CLAUDE.md` note that said "all five `gen-*-opa-bundle*.sh`".

> [!IMPORTANT]
> **Merge last.** This guard is red until #1185 and #1186 land — that is the guard working. It is
> the first thing that has ever been able to see the existing drift.

## Why

The gate named 4 generators; **25 of the 26** hash `rules.yaml`. The four names were never wrong,
just incomplete — and nothing made the list grow, so every new service was opted out of the gate by
default. That is the defect being fixed, not the specific omissions.

Two hard-coded lists, both replaced:

1. **The `verify` step** regenerated 4 generators and diffed 4 directories.
2. **The `changes` job's path filter** — same 4 component dirs. So a hand-edit to, say,
   `components/ledger/ledger-opa-bundle.yaml` did not even mark the PR policy-relevant, and `verify`
   never ran. Fixing only the step would have left this hole.

Both now derive from `find openbank-infra/gitops/components -name 'gen-*opa-bundle*.sh'`.

## Also: fail loudly on a non-executable generator

A `100644` generator silently no-ops a `./…` loop — indistinguishable from "in sync". This is not
hypothetical: `gen-ledger-opa-bundle.sh` shipped that way and had never once run (fixed in #1185).
The check would be self-defeating without this.

## Implementation notes

- **`while read`, not `mapfile`.** The `Install OPA` step in this same job still branches on
  `darwin`, and macOS ships bash 3.2 where `mapfile` does not exist — it would fail
  `command not found`, leave the array empty, and pass a check that verified nothing. Exactly the
  failure mode this PR exists to remove.
- **Empty discovery is a hard error**, for the same reason.
- The agent-bundle step is left as-is: it runs first and gives a targeted message. The dynamic loop
  also covers `gen-opa-bundle-configmap.sh`; re-running it is idempotent.

## Verification

Ran the guard's exact logic locally against `main`:

- discovers **26** generators (19 unique component dirs; `payments` holds 8);
- exec-check correctly flags `gen-ledger-opa-bundle.sh`;
- after `chmod +x` and regenerating all 26, it reports **36 drifted files** — i.e. it catches
  precisely the drift the old gate could not see;
[...]

PR #1186, opened 2026-07-16
fix(gitops): regenerate stale money-path OPA bundles and auto-patch the card-issuance annotation

JiRaska:
## What

Stage 2 of 3 for #1184. Regenerates the **19 stale money-path** OPA bundles, and fixes
`gen-card-issuance-opa-bundle.sh` so it actually patches its pod-roll annotation.

## Why

`.github/workflows/opa-policy.yml` regenerates and diffs only 4 of the **25** generators that hash
`rules.yaml`. Everything not in that hard-coded list drifts with no CI signal.

## This PR changes no policy logic

Verified mechanically, not by eye — for each of the 18 regenerated bundles:

- the ConfigMap key set is unchanged, and
- every embedded `*.rego` document is byte-identical to its source once comments are stripped.

What moves is the embedded **data** documents the policies read at decision time
(`data.rules.*`, `data.agents.*`, mounted via `subPath`) plus the pod-roll annotation. In other
words: this brings the deployed bundles onto the policy data **the repo already declares on `main`**.

## card-issuance generator defect

`gen-card-issuance-opa-bundle.sh` wrote its ConfigMap but never patched the annotation — it just
printed `NOTE: update the card-issuance-service pod template annotation ... (not auto-patched —
shared multi-service file)` and left it to a human. Its committed annotation is `2def8bbead9b9e1c`
against a real checksum of `838a6bb8c479033f`.

The stated reason (a blind `sed` would hit sibling services' annotations in the shared
`payments-services.yaml`) is real, but the seven sibling payments generators already solve it with an
`awk` substitution scoped to their own resource block. This PR uses that same pattern.

One wrinkle: `card-issuance-service` is a **`Deployment`**, while every payments sibling is a
`Rollout` — so the block matcher accepts both kinds. (Audited the rest: `sepa-instant`, `sepa-payment`
and `swift` key on `^kind: Rollout` and genuinely are Rollouts; the non-payments generators `sed` their
own single-service file, where a blind substitution is safe.)

Without this fix the annotation could never appear in a regeneration diff — so the PR3 guard would
have reported card-issuance "in sync" forever while its pods kept mounting the old bundle.

## Blast radius

Rolls **19 money-path services**. This is the whole reason the work is staged: `anacredit`,
`document-service`, `interest-service` already went in #1185, and the CI guard lands in stage 3 once
the drift is gone.

`components/payments/payments-services.yaml` holds all payments pod templates in one file, so
[...]

PR #1300, opened 2026-07-16
fix(infra): regenerate 21 stale OPA bundles so four-eyes matches rules.yaml

JiRaska:
## What

Regenerates the 21 generated OPA bundles that have no CI coverage, plus their pod-roll checksum annotations. All 21 were stale on `main`.

`opa-policy.yml`'s "build + verify bundle" job regenerates and diffs only **5 of 26** bundles (`agent`, `pid`, `copilot`, `customer-edge`, `notifications`) from a hand-maintained list. The other 21 hash `rules.yaml` and `agents.yaml` as checksum inputs, but nothing ever re-ran them when those inputs changed. Every uncovered bundle had drifted; every covered one was clean.

**CI coverage is deliberately a follow-up PR** so this one stays reviewable as a pure policy correction.

## Drift is data-only — no rego rule changed

Every `.rego` key (`rest.rego`, `agents.rego`, and each per-service `*_rest_ext.rego`) is byte-identical modulo comments across all 21 bundles. No allow/deny rule differs. Two embedded data documents were behind.

**`rules-data.yaml`** — semantic YAML diff of every key rego reads:

| key | state |
|---|---|
| `money_path_services` | unchanged |
| `money_path_action_prefixes` | unchanged |
| `feature_flags` | unchanged |
| `value_bands` | unchanged |
| `four_eyes.verbs` | **missing `convert`, `grant`, `revoke`, `clear`** |

Those four are the actions renamed in the #938 follow-up. `rules.yaml` lists all four as four-eyes-gated, and the matching `@Authorize` actions are live: `fx.convert` ([FxResource.kt:75](openbank-fx-service/src/main/kotlin/com/openbank/fx/infrastructure/rest/FxResource.kt#L75)), `consent.grant` / `consent.revoke` ([ConsentResource.kt:65](openbank-consent-service/src/main/kotlin/com/openbank/consent/infrastructure/rest/ConsentResource.kt#L65), [:136](openbank-consent-service/src/main/kotlin/com/openbank/consent/infrastructure/rest/ConsentResource.kt#L136)), `sanctions.clear` ([SanctionsResource.kt:46](openbank-sanctions-service/src/main/kotlin/com/openbank/sanctions/infrastructure/rest/SanctionsResource.kt#L46)). The deployed bundles did not gate what governance says they gate.

**`agents-data.yaml`** — 6 charters added (`control-liveness-sentinel`, `governance-auditor`, `release-steward`, `docs-truth-agent`, `authz-policy-auditor`, `flaky-test-hunter`). Purely additive: 7 → 13, none removed, none altered.

## Impact: latent, not currently exploitable

Four-eyes enforcement is **off fleet-wide** — `AUTHZ_FOUR_EYES_ENFORCE` is set nowhere in gitops, Terraform, or env, and every service defaults it to `false`. `four_eyes_required` only pauses a call into a `PendingApproval` when `authz.four-eyes.enforce=true`, so the missing verbs gate nothing that would otherwise be gated today.

The exposure is the flip: turning `AUTHZ_FOUR_EYES_ENFORCE=true` on consent, sanctions or fx while trusting `rules.yaml` would silently leave those four actions ungated — a single operator could clear a sanctions hit or grant a TPP consent with no second approver. Landing this first makes that flip safe.

The 6 added charters are inert on the MCP path: they declare `tools.allow` as `{tier, resources}` objects, which `charter_allowed` cannot glob-match against a string tool name, and none of those services call `/tools/call`. Default-deny fails closed either way.

## Verification

- `openbank-infra/opa/build-bundle.sh` passes — `opa check` plus all decision assertions, including the four-eyes ones. 13 charters.
- Regenerated by running each `gen-*-opa-bundle.sh`; no file hand-edited.
- Every changed line attributed to its ConfigMap key to confirm zero non-comment rego changes.

## Review notes

Touches money-path services (ledger, sepa-payment, sepa-instant, domestic-payment, settlement, clearing, swift, transaction, card-issuance, balance, account, lending, fx, consent, sanctions, sca) — 2 approvals per ADR-0030. Threat models already exist for all affected services; no new service is introduced.

The diff is large but mechanical: it is generator output. The reviewable claims are the two tables above.

Refs #750

### DIFF
--- a/openbank-infra/gitops/components/accounts/account-opa-bundle.yaml
+++ b/openbank-infra/gitops/components/accounts/account-opa-bundle.yaml
@@ -9,7 +9,7 @@ metadata:
     app.kubernetes.io/name: account-service
     app.kubernetes.io/part-of: accounts
   annotations:
-    openbank.tech/policy-checksum: "f23d7e86bc3ae564"
+    openbank.tech/policy-checksum: "4a45c174e034e8ae"
 data:
   rest.rego: |
     # SPDX-License-Identifier: Apache-2.0
@@ -202,9 +202,10 @@ data:
     # sanctions.create directly, once AUTHZ_ENFORCE=true.
     #
     # Deliberately action-scoped to sanctions.create ONLY, not a "sanctions." family prefix:
-    # sanctions.review (dismiss/confirm a hit) and the list-refresh actions are real operator
-    # judgment calls, not a service-to-service compliance check, and must stay behind a human
-    # session — widening this to the family would let any M2M caller self-clear a sanctions hit.
+    # sanctions.clear (renamed from sanctions.review, issue #938 follow-up — dismiss/confirm a
+    # hit) and the list-refresh actions are real operator judgment calls, not a service-to-service
+    # compliance check, and must stay behind a human session — widening this to the family would
+    # let any M2M caller self-clear a sanctions hit.
     #
     # Deliberately NOT scoped to one hardcoded client id (unlike edge-service-notification):
     # sanctions.create is "run a compliance check" — not an account-takeover primitive like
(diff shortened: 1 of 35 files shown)


JiRaska:
Closing — this was authored against a stale `origin/main` (base `4cae8e2eb`, ~79 commits behind). The drift it reports was already fixed on main and I did not re-verify against the current tip before opening it.

Already merged:
- #1185 — regenerate stale non-money-path bundles + make `gen-ledger-opa-bundle.sh` executable
- #1186 — regenerate stale money-path bundles
- #1187 — discover generators dynamically instead of hard-coding four
- #1190 — record the OPA bundle sync invariant in `rules.yaml` (`opa_bundle_sync`)

Verified against `b050bd9f3`: regenerating all 26 generators produces **zero diff**. `four_eyes.verbs` already contains `convert`, `grant`, `revoke`, `clear`. No action needed.

Issue #1290, opened 2026-07-16
CNPG instance pods still pull from ghcr.io over NAT — fix via spec.imageName, not the rewrite policy


JiRaska:
## Summary

CNPG instance pods are the single largest group of workloads that still pull images from `ghcr.io`
over the NAT instance: **70 live instance pods**, each with two containers the ECR rewrite never
touches.

```
init      bootstrap-controller   ghcr.io/cloudnative-pg/cloudnative-pg:1.29.1     35.0 MB
container postgres               ghcr.io/cloudnative-pg/postgresql:18.1          246.2 MB
```

They are **deliberately** excluded from `ecr-pull-through-rewrite` — and that exclusion is correct.
From the policy's own comment:

> The CNPG operator reconciles by comparing the running pod's image against the Cluster's
> `spec.imageName`; a mutating webhook that rewrites the image makes the pod permanently "drift" from
> spec, so the operator recreates the instance pod every reconcile (~every 10s) and the cluster never
> reaches Ready.

That was learned the hard way (`pact-broker-db`, `fraud-db` looped when `rewrite-ghcr` landed). **Do
not touch the exclusion.**

## The fix the comment already points at

The same comment says: *"CNPG must own its image ref end-to-end."* It can — via `spec.imageName`.
Setting it to the pull-through ref means **no mutation, no drift, the operator owns the ref**, which is
exactly what CNPG wants:

```yaml
spec:
  imageName: 265175468565.dkr.ecr.eu-north-1.amazonaws.com/ghcr/cloudnative-pg/postgresql:18.1
```

**The pull-through already works for these images** — both repos exist and are populated:

```
ghcr/cloudnative-pg/postgresql       10 tags cached
ghcr/cloudnative-pg/cloudnative-pg    5 tags cached
```

The `bootstrap-controller` initContainer is a second axis: CNPG injects the image the **operator
itself** runs as, so it follows the operator Deployment's image, not `spec.imageName`. Both need doing
to fully close the path.

## Blocker: the docs say this is impossible

`ecr-pull-through-rewrite.yaml` still claims:

> `ghcr.io` — needs GitHub PAT in Secrets Manager; see commented stub in `ecr-pull-through-cache.tf`.
> Left on NAT until creds available.

**Stale.** The `ghcr` pull-through rule has existed since **2026-06-14** and demonstrably works — #1259
pins `configmap-reload` through it and the alloy pods are running on
[...]

PR #1304, opened 2026-07-16
fix(gitops): unbreak pact-broker-db sync, pilot the CNPG ECR ref

JiRaska:
## Linked issues

Refs #1290 — pilot (step 2 of the suggested order).

## 1. This Application has been failing every sync for 16 days

`343aef3f8` declared `storage.size: 1Gi` while the volume was already `5Gi`. CNPG's webhook refuses to
shrink:

```
admission webhook "vcluster.cnpg.io" denied the request:
  spec.storage: Invalid value: "1Gi": can't shrink existing storage from 5Gi to 1Gi
```

Argo has been `sync=Unknown` with a `ComparisonError` since **2026-06-30**, while `health=Healthy` —
because the workload itself is fine. **83 of 84 Applications are Synced; this is the one.** Nothing
surfaced it.

Declared size now tracks the volume. It can grow, never shrink — do not tidy it back down.

## Why 18.1 → 16.14 is not a regression

The `imageName: …postgresql:18.1` declared here **never applied** — no sync ever landed. The live
cluster runs **16.14**.

Unblocking the sync without touching that line would push 18.1 at a live database: a Postgres **major**
version upgrade, 16 → 18, that nobody decided and that has been sitting undelivered for 16 days. That
belongs in its own PR with its own thinking, not as a side effect of fixing a storage field. The tag is
aligned to what actually runs.

## 2. The pilot (#1290)

CNPG instance pods are excluded from `ecr-pull-through-rewrite`, and **correctly so** — the operator
reconciles the running pod's image against `spec.imageName`, so a mutating webhook makes the pod
permanently drift and it gets recreated every ~10s. That is precisely what looped `pact-broker-db` and
`fraud-db` when `rewrite-ghcr` landed. **The exclusion is not being touched.**

Setting the ref in the spec is the mechanism the policy's own comment asks for — *"CNPG must own its
image ref end-to-end"* — with no mutation and nothing to drift from:

```yaml
imageName: 265175468565.dkr.ecr.eu-north-1.amazonaws.com/ghcr/cloudnative-pg/postgresql:16.14
```

Same version, different registry, so this tests **one variable**: does CNPG accept an ECR ref and reach
Ready without the reconcile loop? If yes, the remaining ~70 instance pods can follow.

## Verified before pushing

Server-side dry-run against the **live** webhook — the one that has been rejecting this app:

```
$ kubectl apply --server-side --dry-run=server -f openbank-infra/gitops/components/pact-broker/postgres.yaml
cluster.postgresql.cnpg.io/pact-broker-db serverside-applied (server dry run)
```

The storage rejection is gone and the ECR ref is accepted. The `16.14` tag is cached in the
pull-through (246.2 MB, since 2026-06-15).

## Impact and rollback

One rolling restart of a **single-instance, non-money-path** contract-test database. Rollback is
reverting this commit.

**What to watch after merge** — the failure mode this class produces is loud and specific: if the pod
starts recreating every ~10s and the cluster never reaches Ready, that is the drift loop and the pilot
has failed. Revert immediately; do not try to fix it forward.

### DIFF
--- a/openbank-infra/gitops/components/pact-broker/postgres.yaml
+++ b/openbank-infra/gitops/components/pact-broker/postgres.yaml
@@ -20,10 +20,27 @@ spec:
   monitoring:
     enablePodMonitor: true
   instances: 1
-  imageName: ghcr.io/cloudnative-pg/postgresql:18.1
+  # ECR pull-through instead of ghcr.io — the ADR-0092 pilot for #1290. CNPG instance
+  # pods are (correctly) excluded from the Kyverno ecr-pull-through-rewrite policy: the
+  # operator reconciles the running pod's image against spec.imageName, so a mutating
+  # webhook makes the pod permanently drift and it recreates it every ~10s (that is what
+  # looped pact-broker-db and fraud-db when rewrite-ghcr landed). Setting the ref HERE is
+  # the mechanism the policy's own comment asks for — "CNPG must own its image ref
+  # end-to-end" — with no mutation and nothing to drift from.
+  # Tag stays 16.14 = what this cluster actually runs. The 18.1 that was declared here
+  # never applied (see the storage note below) and adopting it would be a Postgres MAJOR
+  # upgrade — a separate, deliberate decision, not a side effect of changing a registry.
+  imageName: 265175468565.dkr.ecr.eu-north-1.amazonaws.com/ghcr/cloudnative-pg/postgresql:16.14
   storage:
     storageClass: gp3
-    size: 1Gi
+    # 5Gi = the live PVC. It was declared 1Gi in 343aef3f8 while the volume was already
+    # 5Gi, and CNPG's webhook refuses to shrink storage:
+    #   spec.storage: Invalid value: "1Gi": can't shrink existing storage from 5Gi to 1Gi
+    # Every sync of this Application has failed since 2026-06-30 as a result — 16 days,
+    # invisibly, because the workload stayed Healthy and only sync status went Unknown.
+    # That is also why the 18.1 above was never applied. Do not "tidy" this back down:
+    # the declared size must track the volume, which can grow but never shrink.
+    size: 5Gi
   resources:
     requests:
       cpu: 100m


JiRaska:
**Self-review before merge — correcting a claim in the description above.**

The body says *"Same version, different registry, so this tests **one variable**"*. **That is not
accurate**, and the difference is worth stating before this merges.

Digests, checked against ghcr.io and ECR just now:

| | digest |
|---|---|
| ghcr.io `16.14` **today** — index | `sha256:242c0a2e8e87…` |
| ghcr.io `16.14` **today** — linux/arm64 child | `sha256:2acd49778ee0…` |
| ECR pull-through `16.14` — index | `sha256:096c60451f80…` |
| ECR pull-through `16.14` — linux/arm64 child | **`sha256:2acd49778ee0…`** |
| **what `pact-broker-db-1` is running** | **`sha256:f810fe7b1319…`** |

Three conclusions:

1. **The ECR cache is faithful, not stale.** Its arm64 child is byte-identical to what ghcr.io serves
   for `16.14` right now. The index digests differ only because ECR re-serialises the index on
   pull-through — expected, and the content that gets executed is the same.
2. **The running pod is on an older `16.14` build** (`f810fe7b…`, pulled ~5 days ago). The upstream tag
   has moved since. So the restart changes the image **content**, not just the registry — a 16.14
   rebuild, almost certainly base-OS patches, *not* a Postgres version change.
3. **The ECR switch is not what causes that.** Any restart re-resolves the `16.14` tag and would pick
   up the same newer build even if this stayed on ghcr.io. But the test is therefore *not* as isolated
   as the description claims, and reviewers should know that.

The underlying weakness this exposes: **pinning by tag is not deterministic**. A digest pin would be.
That is a fleet-wide convention question (every other CNPG cluster pins by tag too), not something to
change unilaterally in a pilot — noted for #1290.

Nothing here changes the go/no-go: the failure mode to watch for is still the drift loop (pod recreated
every ~10s, cluster never Ready). It just means "image content changed" is an *expected* observation
after merge, not a symptom.

PR #1236, opened 2026-07-16
fix(lending): reverse accrued interest on reschedule, derecognize it on write-off

JiRaska:
Closing audit (`docs/audits/2026-07-16-closing-audit.md`) findings **N-1** and **N-2** — the two accrued-interest defects in lending.

## 1. Reschedule destroyed accrued-but-unpaid installments → income recognized twice

`deleteUnpaid()` dropped every unpaid installment with **no `interestAccrued` filter**. But such a row has already posted an `INTEREST_ACCRUAL` (Dr Interest Receivable / Cr Interest Income). Deleting it means:

- its `INTEREST_SETTLEMENT` can never occur → the receivable is **stranded on the GL forever**, and
- the new schedule re-accrues interest on the same outstanding principal → **the same income recognized twice**.

The trigger is not exotic: forbearance of a delinquent loan is *the* normal reschedule case, and a delinquent loan is exactly the one with accrued unpaid installments.

**Fix:** every accrued unpaid installment posts an `INTEREST_ACCRUAL_REVERSAL` (Dr Interest Income / Cr Interest Receivable) *before* its row is discarded. The installment is cancelled, so the income was never truly earned — both legs back out, and the new schedule re-accrues from scratch.

## 2. Write-off derecognized principal only → orphaned interest receivable

`writeOff` booked `outstandingBalance` (principal). Accrued unpaid interest receivable stayed live on the books forever, on a WRITTEN_OFF loan.

**Fix:** `WRITE_OFF_INTEREST` (Dr Loan Loss Expense / Cr Interest Receivable) alongside the principal `WRITE_OFF`.

Note the deliberate asymmetry with #1: this is **not** an income reversal. That interest *was* legitimately earned when the installment fell due; what failed is collection. So the uncollectible receivable is booked as a credit loss, exactly like the principal exposure. Kept as its own `PostingKind` so an audit trail can split a write-off's loss into principal and interest components.

## Crash safety

Both post to the ledger **before** mutating rows, matching the existing `recordRepayment` / `accrueOne` ordering: a crash between the two replays the same idempotency key on retry and the ledger collapses it to the already-booked journal — exactly-once either way. Reschedule keys carry the persisted generation number, reusing the pattern the forgiveness posting already established, so a retried reschedule cannot double-post.

Deliberately **not** in scope: the service's broader multi-transaction model (audit finding N-3 — reschedule is 3 separate txs). These two flows are made economically correct and retry-safe; the transaction consolidation is a separate change.

## Tests

[...]

Issue #999, opened 2026-07-13
interest-service: withholding tax assembled but never remitted to finanční úřad (ADR-0038 self-acknowledged gap)


JiRaska:
## What

Found via ADR-0160's event-consumer-liveness gate (issue #996 triage): `openbank.interest.accrual.event`, published by `openbank-interest-service`, has zero consumers anywhere in the fleet.

This is not a minor completeness gap. **ADR-0038 explicitly says so itself**: "Actual settlement correctness now depends on a downstream consumer of `interest.withholding.remitted.v1`; until that consumer exists the cash is assembled but not yet remitted" — to the tax authority (finanční úřad), per §38d of the income tax act. The withholding amount is correctly calculated and reserved at accrual time, but nothing ever files the statutory XML or executes the cash leg to the state budget.

## Why it matters

This is a **regulatory compliance gap**, not just a technical one: withheld tax that is never remitted is a real liability sitting unaddressed. It reads exactly like #889 (a fully-built upstream half with no downstream half) but with a compliance dimension #889 didn't have.

## Scope (needs design, not a quick PR)

- A downstream consumer (interest-service itself, or a new/existing service — finrep-service is the closest existing candidate given its FINREP/regulatory-reporting role, worth checking first) needs to:
  1. Consume `openbank.interest.accrual.event` (or a purpose-built `interest.withholding.remitted.v1` per ADR-0038's own naming)
  2. Execute the cash-leg posting to the state budget account
  3. Assemble and file the statutory §38d withholding tax XML on the correct periodic cadence
- This is money-path + regulatory — needs the usual 2-approval + threat-model discipline (`rules.yaml: money_path_services`) if interest-service isn't already in that list (check first), and a dedicated design decision (possibly its own ADR) on which service owns the remittance responsibility.

## Linked issues

Refs #996 (the triage this was found in), Refs ADR-0038

## Notes

Not implementing this in the same PR/session that found it — this needs a real design decision (who owns tax remittance) before code, not a fleet-sweep-style mechanical fix.

JiRaska:
## Recommended design direction (not yet implemented — needs confirmation before starting)

Recommend **against** a new microservice. This fleet already has \`finrep-service\` for regulatory reporting — the correct bounded context, not a new one.

[...]

Issue #747, opened 2026-07-11
PaymentJournalFactory: non-CZK cash-clearing leg posts against the wrong GL account


JiRaska:
## What

Any inbound/outbound payment (not an internal transfer) in a non-CZK currency posts its bank-side cash-clearing journal leg against a hardcoded CZK-only GL account, which fails ledger-service's currency validation.

`openbank-transaction-service/src/main/kotlin/com/openbank/transaction/application/usecase/PaymentJournalFactory.kt`'s `cashClearingLeg()` (around line 114) hardcodes `glAccountId = CUSTOMER_CASH_CLEARING_CZK` (a single CZK-denominated GL account, comment: "1100 Customer Cash Clearing (ASSET, CZK)") regardless of the `ccy` parameter passed in, while the sibling `depositControl()`/`fxPosition()` helpers correctly look up a per-currency account from a map (`DEPOSIT_CONTROL`, `FX_POSITION` — EUR/USD/GBP/CZK all present).

`sameCurrencyLines()` uses `cashClearingLeg` for the "outbound" (source only) and "incoming credit" (target only) cases — i.e. any payment that isn't a same-currency internal transfer between two customer pockets. For those, the `JournalLineRequest` carries the REAL transaction currency (e.g. EUR) but posts against the CZK-only `CUSTOMER_CASH_CLEARING_CZK` account.

## Confirmed live

`POST /api/v1/transactions` with `{"type":"CREDIT","targetAccountId":<EUR account>,"amount":"1000000.00","currencyCode":"EUR",...}` completes the Temporal `PaymentWorkflow` but the `PostJournal` activity gets `422 Unprocessable Entity` from ledger-service (`LedgerRestClient#postJournal`), and after a couple of retries the `LedgerCallGuard` circuit breaker trips, permanently failing the transaction (`status: FAILED`). The byte-for-byte identical request with `currencyCode: CZK` against a CZK account completes normally (`status: COMPLETED`) in ~300ms.

Internal transfers (both source and target set, same currency) don't hit this path — they use `depositLeg` on both sides, no cash-clearing leg at all — so this only affects one-sided inbound/outbound payments in a non-CZK currency.

## Task

[...]

Issue #468, opened 2026-07-07
Pact: contract-drift gate (ADR-0063 Phase 2) + expand coverage to uncovered integration edges


JiRaska:
## Problem

Contract-test coverage is concentrated on the account/balance/ledger/transaction core while most inter-service calls have no contract at all, and committed pacts can silently drift from their consumer tests.

Audit findings (2026-07-07, origin/main):

- 6 committed HTTP pacts in `pacts/`, ~10 consumer services, 7 provider verification tests. That covers roughly a third of the fleet's real integrations.
- Services with REST/event dependencies but **no pact in either direction**: settlement, billing, interest, standing-order, clearing (inbound messages have simulator tests but no consumer contract), kyc→party, dispute→transaction/party, onboarding→party/kyc/sca, sepa-instant→transaction/ledger, sanctions, pid→party.
- #370 fixed the dead `pactbroker.url`-gated provider tests by switching to git-pact (`@PactFolder("../pacts")`), and its "bonus finding" was exactly the remaining hole: **a committed pact was stale** (missing `KYC_STATUS_CHANGED` interaction) and the provider would have verified an incomplete contract. The "regenerate pact on consumer change" convention is a doc comment, not a gate — ADR-0063 Phase 2 (drift check) is still unimplemented.
- `can-i-deploy` (ADR-0092) gates the main-push deploy, which is correct, but it can only gate on pacts that exist.

## Proposed solution

Two workstreams:

1. **Drift gate (ADR-0063 Phase 2)** — a CI check that regenerates consumer pacts from the consumer tests (they already write to `pacts/` via `pact.rootDir`) and fails the PR if the committed JSON differs (same derived-data pattern as the NetworkPolicy and catalog generators: derived artifacts are never hand-edited and never stale).
2. **Coverage expansion sweep** — one PR per uncovered integration edge, priority order:
   - money-path edges first: sepa-instant→transaction/ledger, settlement→ledger, billing→ledger, clearing inbound (pacs.008 as message pact)
   - then platform edges: kyc→party, onboarding→party/kyc/sca, dispute→transaction, interest→balance, standing-order→payment services
   - each edge = consumer test generating a pact in `pacts/` + provider verification via `@PactFolder` (one `@Provider` test per provider — mixed HTTP+message picks the target per interaction in `@BeforeEach`).

## Alternatives considered

[...]

PR #1316, opened 2026-07-16
feat(interest): post the ADR-0033 split so capitalization actually credits the customer

JiRaska:
Implements ADR-0033 §D. Prerequisite for ever enabling the interest crons (#999, `docs/audits/2026-07-16-closing-audit.md` finding I-7).

## Why this had to come first

`capitalize()` recorded the capitalization + withholding rows in interest-service's own tables and posted **nothing** to the ledger — verified: no ledger reference anywhere in `InterestService`, and **nothing consumes `interest.withholding.recorded.v1`**. Meanwhile `WithholdingRemittanceSettlementConsumer` **does** move real money (`transactionClient.initiateTransaction`).

So turning on the accrual/capitalization crons in that state would have made **the bank pay real withholding tax to the finanční úřad on interest it never credited to anyone**, with the GL recording neither the expense nor the customer liability. Dead code at least moves no money. That is why this PR exists and why the crons remain stubs.

## The split

```
DEBIT  4010/4011/4012/4013  Interest Expense <ccy>          gross
CREDIT 2100/2101/2102/2103  Customer Deposit Control <ccy>  net    (subAccountId = accountId)
CREDIT 2200                 Withholding Tax Payable         tax
```

`gross = net + tax` is asserted before the request is built, so the entry balances per currency.

| case | legs |
| --- | --- |
| gross 100.00 CZK, resident individual 15% | Dr `4010` 100.00 · Cr `2100` 85.00 (sub) · Cr `2200` 15 |
| gross 6.99 CZK — `floor(6) × 0.15 = 0`, still `WITHHELD` | **two** legs: Dr `4010` 6.99 · Cr `2100` 6.99 |
| EUR 50.00, `DEFERRED_FX` | Dr `4011` 50.00 · Cr `2101` 50.00 — no 2200 leg |
| zero gross | **no journal**; the capitalization row is still written (matches V6's partial index `WHERE total_accrued <> 0`) |

Zero-amount legs are dropped, never sent. Whole-CZK `DOWN` assessment (daňový řád) makes `taxAmount = 0` the *normal* case below 7.00 CZK — the same reachable state that #1264 had to fix on the remittance side.

## Ledger first, own rows second

`postCreditLeg(cap).flatMap { saveWithOutbox(...) }`, mirroring `LendingService.accrueOne`, and deliberately **outside** `saveWithOutbox`'s `sf.withTransaction` (no network call inside a DB transaction).

A crash between the two leaves the accruals `ACCRUING` and **no** capitalization row, so the retry re-runs the period and replays the **same business-derived idempotency key** — `interest-capitalization-{accountId}-{productId}-{periodTo}`, never `cap.id`, which is a fresh UUID per attempt and would **double-post real money**. The ledger collapses the replay. Exactly-once.

Commit-then-post would strand a credit that exists in interest-service and nowhere in the GL, which no retry can repair because the accruals would already be `CAPITALIZED`.

## V17: two GL findings that contradict the ADR

ADR-0033 §D opens with *"Capitalization currently posts one gross leg"* — **false**; there was no posting to convert.

1. **`2200 Withholding Tax Payable` did not exist at all.**
2. **`4000 Interest Expense` exists but is unusable.** V1 seeds it with `gen_random_uuid()`, so its id differs per deployment and nothing can reference it. This is **issue #468 verbatim** — V15 hit it on `4001 Fee Income` (its own comment: *"every real fee-charge posting 422'd with 'GL account not found'"*) and fixed it by seeding a **new stable-UUID code** rather than editing an applied migration. Same remedy: a new `401x` family.
3. **Interest expense must be per-currency.** `LedgerService` (`:270`) rejects a line whose base currency ≠ its GL account's currency, and `4000` is CZK-only — one shared expense account would 422 the entire EUR entry, killing the customer credit with it. V14 hit exactly this on cash clearing (#747, *"confirmed live: a EUR CREDIT transaction fails there"*).

`2100` (V3, id `…0002`) and `2101`/`2102`/`2103` (V5) already carry stable ids — reused, not re-seeded. V1's `4000` row is left untouched.

## Verified independently, not taken on trust

I re-ran everything myself rather than accept the implementation report:

- **`subAccountId` = the capitalization's `accountId`** — confirmed against four independent precedents (`PaymentJournalFactory.depositLeg`, `BillingJournalFactory`, the ledger's own `JournalEntry.bookedDeltas()` projecting `subAccountId → AccountBookedDelta(accountId, …)`, and `InterestAccrualScenario` in the simulation). Not assumed. This matters because `LedgerService:278` rejects `subAccountId` on non-control legs but **never requires it on control legs** (closing-audit finding L-2) — nothing would have stopped a sub-ledger-blind posting, and the result is a tie-out break at the next 06:00 run.
- **Mutation-tested the two decisive properties**: setting the deposit leg's `subAccountId = null` fails 2 tests; moving the ledger post after `saveWithOutbox` fails 3, including `a ledger failure leaves NO capitalization row`. Green tests are not evidence on their own — that lesson is from #1236 (reverted) and #1246 (self-inflicted, fixed in #1264) earlier today.
- `./gradlew :openbank-interest-service:{test,ktlintCheck,detekt} :openbank-ledger-service:{test,ktlintCheck,detekt}` — **BUILD SUCCESSFUL** (JDK 25), 26 interest tests, 0 failures.
- Diff scope confirmed: only the intended files (an earlier `ktlintFormat` had silently reformatted 12 unrelated baselined files; all reverted).

## Judgement calls worth challenging

1. **The `401x` family instead of `4000`** — forced by #468 + the currency check, but it does leave V1's `4000` as a dead row in the chart. Cleaning that up is a separate concern (nothing references it by UUID; only finrep test fixtures use the code string).
2. **Non-CZK is credited gross** (`DEFERRED_FX`, ADR-0033 §E). The alternative — refuse non-CZK capitalization entirely until the FX withholding design lands — is defensible and would be a smaller surface.
3. **The adapter is bound unconditionally**, not behind a `backend=none` no-op like lending's. A silent no-op default is what produced this very defect. Cost: a deployment without `LEDGER_SERVICE_URL` fails every capitalization — loudly, leaving accruals `ACCRUING` (nothing credited, nothing withheld), which I judge strictly better than silence.
4. **interest-service has no gitops Deployment** (docker-compose only), so there was no manifest to wire. Worth confirming that is expected before this can run in-cluster.
5. `docker-compose.yml` also gained `TRANSACTION_SERVICE_URL`, which was missing — the existing remittance path was pointing at localhost.

## Still out of scope

`accrueAll`/`capitalizeAll` remain stubs; no `@Scheduled` is enabled. This PR makes capitalization *correct*; turning it on additionally needs account enumeration and a balance port — **interest-service has neither**, and neither is designed anywhere. That remains the open product decision behind #999.

### DIFF
--- a/openbank-infra/docker-compose.yml
+++ b/openbank-infra/docker-compose.yml
@@ -1012,6 +1012,10 @@ services:
       QUARKUS_REDIS_HOSTS: redis://:${VALKEY_PASSWORD}@valkey:6379
       KAFKA_BOOTSTRAP_SERVERS: kafka:9092
       MP_MESSAGING_CONNECTOR_SMALLRYE_KAFKA_BOOTSTRAP_SERVERS: kafka:9092
+      # ADR-0033 §D: capitalization posts its credit leg to the ledger. The application.yaml
+      # default is localhost, which resolves to the container itself here.
+      LEDGER_SERVICE_URL: http://ledger-service:8101
+      TRANSACTION_SERVICE_URL: http://transaction-service:8102
     ports:
       - "8125:8125"
     depends_on:
--- a/openbank-interest-service/build.gradle.kts
+++ b/openbank-interest-service/build.gradle.kts
@@ -48,6 +48,12 @@ dependencies {
     testImplementation(libs.testcontainers)
     testImplementation(libs.testcontainers.junit)
     testImplementation(libs.testcontainers.postgresql)
+
+    // Consumer-driven contract for the ledger-service postJournal call the ADR-0033 §D
+    // capitalization credit leg makes (ADR-0063). interest-service is the fifth ledger-posting
+    // consumer; every other one already has a pact with openbank-ledger-service as provider.
+    testImplementation(libs.rest.assured.kotlin)
+    testImplementation(libs.pact.consumer)
 }
 
 kover {
@@ -63,4 +69,23 @@ kover {
             }
         }
     }
+}
+
+// Pact: write the generated consumer contract to pacts/ and forward broker config, matching
+// transaction-service/lending-service/billing-service/settlement-service's tasks.withType<Test>
+// block (ADR-0063 P1/P2). Without pact.rootDir the pact lands in the module's build/pacts and
+// .github/workflows/pact-drift-check.yml never sees it.
+tasks.withType<Test> {
+    systemProperty("pact.rootDir", "${rootProject.projectDir}/pacts")
+    listOf(
+        "pactbroker.url",
+        "pactbroker.auth.username",
+        "pactbroker.auth.password",
+        "pactbroker.enablePending",
+        "pactbroker.providerBranch",
+        "pact.verifier.publishResults",
+        "pact.provider.version",
+        "pact.provider.branch",
+        "pact.provider.tag",
+    ).forEach { key -> System.getProperty(key)?.let { systemProperty(key, it) } }
 }
--- a/openbank-interest-service/ktlint-baseline.xml
+++ b/openbank-interest-service/ktlint-baseline.xml
@@ -143,106 +143,108 @@
         <error line="45" column="29" source="standard:annotation" />
         <error line="45" column="58" source="standard:annotation" />
         <error line="46" column="1" source="standard:spacing-between-declarations-with-annotations" />
-        <error line="46" column="37" source="standard:annotation" />
+        <error line="46" column="40" source="standard:annotation" />
         <error line="47" column="1" source="standard:spacing-between-declarations-with-annotations" />
-        <error line="47" column="33" source="standard:annotation" />
-        <error line="50" column="8" source="standard:annotation" />
-        <error line="52" column="8" source="standard:annotation" />
-        <error line="52" column="43" source="standard:annotation" />
-        <error line="53" column="1" source="standard:spacing-between-declarations-with-annotations" />
-        <error line="53" column="60" source="standard:annotation" />
+        <error line="47" column="37" source="standard:annotation" />
+        <error line="48" column="1" source="standard:spacing-between-declarations-with-annotations" />
+        <error line="48" column="33" source="standard:annotation" />
+        <error line="51" column="8" source="standard:annotation" />
+        <error line="53" column="8" source="standard:annotation" />
+        <error line="53" column="43" source="standard:annotation" />
         <error line="54" column="1" source="standard:spacing-between-declarations-with-annotations" />
-        <error line="54" column="33" source="standard:annotation" />
+        <error line="54" column="60" source="standard:annotation" />
         <error line="55" column="1" source="standard:spacing-between-declarations-with-annotations" />
-        <error line="55" column="34" source="standard:annotation" />
+        <error line="55" column="33" source="standard:annotation" />
         <error line="56" column="1" source="standard:spacing-between-declarations-with-annotations" />
-        <error line="56" column="32" source="standard:annotation" />
+        <error line="56" column="34" source="standard:annotation" />
         <error line="57" column="1" source="standard:spacing-between-declarations-with-annotations" />
-        <error line="57" column="63" source="standard:annotation" />
+        <error line="57" column="32" source="standard:annotation" />
         <error line="58" column="1" source="standard:spacing-between-declarations-with-annotations" />
-        <error line="58" column="68" source="standard:annotation" />
+        <error line="58" column="63" source="standard:annotation" />
         <error line="59" column="1" source="standard:spacing-between-declarations-with-annotations" />
-        <error line="59" column="62" source="standard:annotation" />
+        <error line="59" column="68" source="standard:annotation" />
         <error line="60" column="1" source="standard:spacing-between-declarations-with-annotations" />
-        <error line="60" column="60" source="standard:annotation" />
+        <error line="60" column="62" source="standard:annotation" />
         <error line="61" column="1" source="standard:spacing-between-declarations-with-annotations" />
         <error line="61" column="60" source="standard:annotation" />
         <error line="62" column="1" source="standard:spacing-between-declarations-with-annotations" />
-        <error line="62" column="43" source="standard:annotation" />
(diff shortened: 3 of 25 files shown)


JiRaska:
## Blocked by adversarial review — three CONFIRMED findings, one fatal

Marking draft. **Finding 1 alone means this PR does not work at all**, and I missed it in my own review.

### 1. CRITICAL — every money-bearing capitalization is rejected with HTTP 400

```
InterestService.kt:128   gross = total.setScale(4, HALF_UP)      -> scale ALWAYS 4
LedgerService.kt:95      amount = Money.of(req.amount, req.currencyCode)
Money.kt:17              require(amount.scale() <= currency.defaultFractionDigits)   // == 2
                         require(4 <= 2) -> false -> IllegalArgumentException -> 400
```

Not an edge case — **100% of the path this PR exists to create**, including the worked example in the PR body above (`Dr 4010 100.00`). The PR states as fact, in its body, its commit message and the factory KDoc, that the split posts. It never can.

**Why the tests missed it, and why my mutation-testing didn't cover it:** every test stops one layer short of the boundary. `CapitalizationJournalFactoryTest` ends at the DTO and its `assertBalanced` is a hand-rolled restatement of the ledger's rules that omits `Money`'s scale invariant; every amount assertion uses `isEqualByComparingTo`, which **ignores scale by construction** — the one property that breaks is invisible to the matcher used throughout. `InterestServiceTest` mocks the port. `CapitalizationTransactionIT` injects the repository, not the service.

lending/transaction are immune **by type, not by luck** — they read `.amount.amount` off a `Money`. This PR mirrored lending's port shape and dropped the `Money` that carried the invariant. Fix: carry `Money` in `CapitalizationPosting`, round to currency scale once at the source.

### 2. HIGH — the idempotency key is amount-blind, so "exactly-once on money" fails in the window it was written for

The key pins a *period*, not an economic event, and `findPendingCapitalization` has no lower bound. Crash after the ledger post → backfilled accrual → retry re-queries a **larger** set → same key → ledger replays the old journal → cap row commits at the new gross. interest-service says 120, the GL moved 100 → the remittance pays real tax on interest nobody received. Nothing reconciles interest-service against the GL, so it is silent and permanent.

Putting the amount in the key does **not** fix this — I checked: a changed set mints a new key, the ledger posts a *second* journal, and the orphan stands → double credit. `JournalResponse(id, transactionId, status)` carries no amount, so detecting the mismatch off the response isn't available either.

Fix: **claim the set before posting** (ACCRUING → CAPITALIZING in its own guarded tx → post → write rows + flip CAPITALIZING → CAPITALIZED). The retry then finds the same claimed set, so the amount and the key agree; a backfill lands outside the claim and falls to the next period.

### 3. MEDIUM — negative gross silently skips the GL, and its KDoc's cited authority refutes it

The guard skips `signum() <= 0` and justifies it by pointing at V6's `WHERE total_accrued <> 0`. **V6 excludes zero; the guard excludes zero and everything below.** A negative rate (no CHECK on `annual_rate`, no `@DecimalMin`) commits a cap row saying the customer was charged while the GL records nothing. Zero → no journal stays; negative → refuse loudly.

### Also required

An integration test driving `capitalize` against a **real** `LedgerService`, asserting the posted journal's amounts against the cap row — the test that would have caught #1.

Verified sound and unchanged: the posting direction (`CREDIT 2100` = customer receives, per `PaymentJournalFactory`), `subAccountId` placement, ledger-first ordering, the `401x`/#468 and per-currency/#747 reasoning, and that V17 breaks no chart-assuming query.

Same failure mode as #1236: the tests encode the author's model of the boundary rather than the boundary itself.

JiRaska:
## Round 2 adversarial review — 2 fixed, 1 tracked, 1 noted

Both original CRITICAL/HIGH fixes held up under a second independent pass. Small findings this round:

**Fixed** (pushed): the `LedgerCallGuard` KDoc still described pre-claim-state behavior ("outage leaves accruals ACCRUING" — should be CAPITALIZING, since the claim now commits before the ledger call); and `CapitalizationLedgerBoundaryIT` compared the persisted row to the in-memory value with `isEqualByComparingTo` — the same scale-blind comparator flagged elsewhere in this PR as unable to catch the class of bug round 1 found. `interest_capitalizations.gross_amount/net_amount` are `NUMERIC(20,4)` (pre-existing), so Postgres always round-trips `100.00` back as `100.0000` — numerically sound today (no rounding step in between), but the assertion couldn't have told the difference if there had been one. Now asserts the persisted scale explicitly with a comment naming the gap, instead of silently normalizing it away.

**Tracked, not fixed here** — dormant race: `capitalize`'s retry re-derives tax/net from `taxProfilePort.resolve(accountId)` fresh on every attempt, while `gross` is frozen by the claimed accrual set. If a tax-profile lookup ever becomes account/party-dependent (`TaxPorts.kt` already documents this as a fast-follow) and the profile changes between a crashed attempt and its retry, the ledger's amount-blind idempotency key would replay the first journal while `saveWithOutbox` locally commits the newly-recomputed tax/net — reproducing the row-vs-GL divergence round 1 fixed, through a different vector. Not exploitable today: `DefaultTaxProfileProvider` unconditionally returns the constant `FAIL_SAFE_DEFAULT`, so nothing can actually drift between attempts yet. Filing this so the eventual account→party tax-attribute work doesn't silently reopen it — will open an issue referencing this thread rather than block on a hypothetical.

**Confirmed solid** (tried to break, could not): the backfill race is closed — a backfilled accrual can only ever be `ACCRUING`, never joins an already-claimed `CAPITALIZING` set; concurrent `capitalize()` calls for the same (account,product,periodTo) are safe by row-lock + count-guard, backstopped independently by V6's unique index; a stuck `CAPITALIZING` claim recovers via a plain retry of the same `capitalize` call, no new #1265-shaped wedge; V7's enum/constraint design is sound; the pact is built through the real factory and real tax policy, not a parallel hand-rolled path; nothing outside this diff (FX reval, trial balance, tie-out, FINREP) is touched beyond the V17 account seeding already called out in round 1.

Ready to come out of draft once the tracking issue for the dormant race is filed.

Issue #1284, opened 2026-07-16
Two real faults hidden behind ArgoCD Degraded: document-service has no signing keystore, notifications' KEDA trigger is broken


JiRaska:
**Corrected.** My first version of this issue claimed the apps were healthy and ArgoCD was lying. That was wrong — ArgoCD was right, and my check was broken. Both apps have a genuine failing child.

```
notifications      health=Degraded  since 2026-07-15T17:10:02Z
document-service   health=Degraded  since 2026-07-14T14:08:15Z
```

## 1. document-service has no signing keystore — PAdES signing is silently unavailable

```
ExternalSecret document-service-signing-keystore
  Ready=False  SecretSyncedError: could not get secret data from provider
  remoteRef: key=document-service-signing-seal  property=KEYSTORE_P12_BASE64  (store: vault-kv)
```

The Vault key cannot be read, so the Secret **does not exist at all** (`secrets "document-service-signing-keystore" not found`).

The Deployment consumes it in two places — and both are `optional: true`:

```
env    OPENBANK_SIGNATURE_KEYSTORE_PASSWORD <- document-service-signing-keystore/keystore.password   optional=True
volume signing-keystore                      <- document-service-signing-keystore                    optional=True
```

**That `optional: true` is why nobody noticed.** A non-optional reference fails the pod with `CreateContainerConfigError` (exactly what `vop-oidc` did today, and it got fixed within the hour). An optional one lets Kubernetes drop the mount silently — the pod runs `2/2`, reports `status: UP`, and serves. Whatever PAdES-B signing (ADR-0162 phase 1) does without its keystore, it has been doing it since at least the 14th.

Worth deciding: should a signing keystore ever be `optional`? A document service that cannot sign is arguably a document service that should not start.

## 2. notifications' KEDA autoscaler has been broken for 44 days

```
ScaledObject notification-service   MIN=1 MAX=3   trigger=kafka   READY=False   AGE=44d
  Ready=False   TriggerError: Triggers defined in ScaledObject are not working correctly
  Active=False  ScalerNotActive: Scaling is not performed because triggers are not active
```

`MIN=1` keeps one replica alive, so the service works — it simply can never scale up on Kafka lag. Every ExternalSecret in the namespace is fine; this is the only fault.

## Why this went unseen — and the check that hid it

Both faults are real and ArgoCD reported them correctly. The reason my first pass concluded "everything is healthy" is worth recording, because it is the same trap twice in one investigation:

```
[...]

PR #1320, opened 2026-07-16
docs: correct the release trigger to both axes, record three footguns

JiRaska:
Two documents that disagree with the code, and three footguns that each cost a session today.

Supersedes #1295, which set out to fix the same CLAUDE.md sentence — see the last section.

## 1. RELEASE.md was made false an hour ago

#1277 (`da6a734c3`) gave all 51 packages an `exclude-paths`. `RELEASE.md:49` still says:

> release-please has no concept of a sub-path exclusion

and that a `src/test/**`-only change "**still proposes a release**". Neither holds now. Corrected, plus the mechanics worth knowing: exclusion matches **directory prefixes only** (`file.indexOf(path + '/') === 0`), so a single file cannot be excluded, and a commit touching `src/main` **and** `src/test` still releases — a commit is skipped only if **all** its files match an exclude.

## 2. CLAUDE.md rule 2 was wrong in the other direction

> A change under `<service>/src/main/**` is released by release-please

Understates the path axis (`governance.yaml`, `Dockerfile`, `build.gradle.kts`, the lint baselines all release too) **and omits the type axis entirely**. The honest rule needs both:

> release ⟺ (type ∈ {feat, fix, perf, security} or breaking) **AND** (a changed file under `<service>/` outside that package's `exclude-paths`)

A hidden type (`refactor` `docs` `test` `build` `ci` `chore`) **never** releases, at any path. That matters because it is exactly the lever `rules.yaml: release_scope_mismatch` already advises — "use a non-releasing type when the PR doesn't touch the release-worthy subtree" — and the old text implied no such lever existed.

**Verified against the config, not asserted:**
- 51 packages, each excluding its own `src/test`; admin-ui also excludes `e2e`, and the text now says so (my first draft said just "`src/test`" — the config caught me).
- Live control: `chore(admin-ui): deploy …` commits have touched `openbank-admin-ui/` since tag `admin-ui-v0.49.0`, release-please has run since, and **no admin-ui release PR exists**. Only possible if the type gates the release.

## 3. Three footguns

- **`@ApplicationScoped` is lazy** — an `init {}` warning or go-live `check()` does not run at boot. `PdfBoxPadesSealAdapter`'s "every seal is worthless as evidence" warning had never once appeared in a pod log, and `require-trusted-issuer` — documented as "refuses to **start**" — would have thrown on a *request* instead. Fixed by #1299; the trap generalises. → Kotlin/Quarkus.
- **`optional: true` on a secret ref is a trade, not a default.** Without it: loud `CreateContainerConfigError`, fixed within the hour (vop-oidc, #1232). With it: pod runs `2/2` reporting `UP` while the feature is silently degraded — document-service sealed PDFs with a throwaway cert for two days (#1284). Legitimate when rollout order must never crash-loop the service, but then the fallback needs its own **eager** alarm. → GitOps.
- **An Argo Rollout whose revision 1 never succeeded deadlocks on a dead `stable`.** Neither `retry` nor `promote --full` fixes it — promote skips steps and pauses, not a broken stable. Delete the dead ReplicaSet (verify it owns zero pods first). Cost ~40 minutes on vop-service today. → GitOps.

## On #1295

#1295 aimed at the same CLAUDE.md sentence and got two things genuinely right, which I've kept: there is no include/allow-list, and exclusion is directory-only so a single file can't be excluded.

Where it goes wrong is *"touch one and the service ships a version"* — that asserts path alone triggers a release, which the admin-ui control above disproves. It also cites devops-agent 0.3.2→**0.4.0** as evidence; a *minor* is only ever produced by a `feat`, so that example shows the type deciding, not the path. Merging it would have put two false statements into the file every contributor and agent reads, and contradicted `rules.yaml` — which CLAUDE.md itself declares authoritative. Close it in favour of this, or take this as the rewrite it asked for.

Docs-only; no code or config changes.

Refs #1284, #1299

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -14,9 +14,18 @@ Before a change is "done", run the ship-checklist (`/ship-check` runs the same c
 ADR-0029):
 
 1. **Open a PR.** No direct commits to `main`. Branch `<type>/<scope>-<summary>`; squash-merge via PR.
-2. **Versioning is automatic.** A change under `<service>/src/main/**` is released by **release-please**
-   from your Conventional Commit — so the commit message *is* the changelog. Do **not** hand-edit
-   `version.txt`, `CHANGELOG.md`, or `quarkus.application.version` (it derives from `version.txt`).
+2. **Versioning is automatic, and needs BOTH axes.** release-please cuts a release only when the
+   commit **type** releases (`feat`/`fix`/`perf`/`security`, or breaking) **and** it touches a file
+   under `<service>/` outside that package's `exclude-paths` (today `src/test`, plus `e2e` for
+   admin-ui). The commit message *is* the changelog. Neither axis alone is enough:
+   - A hidden type (`refactor` `docs` `test` `build` `ci` `chore`) **never** releases, at any path
+     — the lever when a PR doesn't change shipped code (`rules.yaml: release_scope_mismatch`).
+   - The path axis is broader than `src/main`: `governance.yaml`, `Dockerfile`, `build.gradle.kts`
+     and the lint baselines all count. There is no include/allow-list — only `exclude-paths`,
+     matching **directories** only, so a single file cannot be excluded.
+
+   Do **not** hand-edit `version.txt`, `CHANGELOG.md`, or `quarkus.application.version` (it derives
+   from `version.txt`).
    A module is a released component **iff** it has a `version.txt` (registered in
    [`release-please-config.json`](release-please-config.json) + [`.release-please-manifest.json`](.release-please-manifest.json)).
 3. **API change ⇒ `openapi.yaml` updated + contract test.** Two independent version axes (ADR-0048):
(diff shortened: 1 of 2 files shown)


JiRaska:
An adversarial review of this PR found three defects **in the PR itself** — fitting, for a change whose thesis is that false docs are the worst outcome. All three verified before fixing; pushed as `525edf469`.

**1. A dangling pointer I created inside my own diff.** `RELEASE.md` kept *"even though CLAUDE.md rule 2 says only `src/main/**` changes should"* — in the very paragraph this PR edits — while the PR rewrites rule 2 to say the opposite. Now cites `rules.yaml: change_requirements.release_scope_mismatch`, which is where that normative claim actually lives and which stays true.

**2. A factual error of exactly the kind this PR exists to remove.** *"A hidden type **never** releases, at any path"* is false: a breaking marker renders a `⚠ BREAKING CHANGES` section, so the body is non-empty and `refactor!` / `chore!` release a **major**. It also contradicted my own line three rows above ("or breaking"). The reviewer verified by running release-please 17.x against this repo's actual `changelog-sections`. Qualified now.

**3. `rules.yaml:245-247` cites "(rule 2)" for a claim this PR moves — deliberately NOT changed.** `rules.yaml` is embedded verbatim by **26** bundle generators, so a one-word comment edit rolls every service's `policy-checksum` and pods. Not worth a fleet roll for a pointer. And the two do not actually conflict: `rules.yaml` states what **should** release (normative, gate-backed); rule 2 states what **does** (mechanical). Worth a follow-up when something else touches `rules.yaml` anyway.

## What the review could not refute

The load-bearing claim — that hidden types don't release — survived execution-level verification. The reviewer ran `release-please@17.10.3`'s `DefaultChangelogNotes` with this repo's real config:

| commits | body lines | result |
|---|---|---|
| chore-only | 1 | **SKIP** |
| test-only | 1 | **SKIP** |
| fix-only | 6 | RELEASE |
| chore+fix | 6 | RELEASE |

`changelogEmpty()` (`base.js:316`) returns `undefined` → no release PR. The 51-package / `src/test` / `e2e` / directory-prefix facts and the `file.indexOf(path + '/') === 0` quote all checked out exactly, and the admin-ui live control held with both alternative explanations ruled out.

## Two known limits, stated rather than papered over

- **The `⟺` has a real exception.** `CommitSplit` runs `includeEmpty: true`, so a commit touching **zero** files applies to every configured package. Live proof: `37d303e` (#480) is an empty commit and released ~22 packages. So the right conjunct isn't strictly necessary. Rare enough that stating it in rule 2 would cost more clarity than it buys — recording it here instead.
- **Footgun (c) is n=1.** The Rollout deadlock mechanism is sound (`retry` only clears an abort; `promote --full` skips steps, not a broken `stable`), and the setup is real in `payments-services.yaml` — but "neither fixes it" and "~90s" generalise from one incident today. The text already says "Seen on vop-service, 2026-07-16"; treat it as a dated observation, not a law.

PR #1358, opened 2026-07-16
feat(governance): add opsmessage.* action namespace + four-eyes gate (ADR-0176 D4/D5)

JiRaska:
## What

First of three PRs implementing **ADR-0176** (operator-initiated customer messaging) end to end. This one adds the OPA infrastructure the other two depend on: a distinct `opsmessage.*` action namespace and a four-eyes gate for it, per D4/D5.

There is currently no operator-facing way to send a customer a message from admin-ui — `notification-service`'s only ingress is Kafka, and only `account-service` publishes today. ADR-0176 designs this deliberately (catalogue templates only, wake-signal-only push, purpose-limited, maker-checker-gated) rather than the naive "textarea wired to an endpoint," for five concrete reasons detailed in the ADR. This PR is the authorization layer that makes the rest safe to build on.

## rest.rego

- **`opsmessage-compose`** — grants `ROLE_OPERATOR`/`ROLE_ADMIN` the `opsmessage.compose` action. Deliberately a new namespace, not `notification.*`: `edge-service-notification` (above) auto-grants every `notification.*`-prefixed action to the customer-edge M2M identity via `startswith`, so naming this `notification.compose` would hand the send capability straight to that M2M identity.
- **`opsmessage-approve`** — grants the same roles `opsmessage.approve`/`opsmessage.reject`. This is the checker side of the maker-checker pair.
- **New `four_eyes_required` disjunct**: `input.action in data.rules.four_eyes.actions`. Only `opsmessage.compose` is in that list — approve/reject are the checker's own decision and are deliberately **not** gated (gating them would need a third approver to approve the approval).

## rules.yaml

New `four_eyes.actions` list, sibling to the existing `four_eyes.verbs`, first entry `opsmessage.compose`. Kept deliberately disjoint per the guardrail comment already on `verbs` (money-path-scope × verb vs. this, an unconditional exact-name match) — `notification-service` joining `money_path_services` instead was considered and rejected in the ADR (drags 2-approval review + threat model + mutation testing + coverage floor + SLO pair onto the whole service to gate one action, and the verb-match would still sweep in M2M callers by verb name alone).

## Correctness check worth calling out

Before adding the new disjunct, I verified empirically (not just by reasoning about Rego semantics) that `input.action in data.rules.four_eyes.actions` on a mock lacking the `actions` key doesn't error — it just leaves that rule body unmatched. All 124 pre-existing tests use a mock without that key; if this had thrown, every one of them would have broken. Ran the full suite before and after: 124/124 → 133/133, zero regressions.

## Tests

9 new `rest_test.rego` cases:
- `test_allow_operator_opsmessage_compose` / `test_allow_admin_opsmessage_compose`
- `test_allow_operator_opsmessage_approve` / `test_allow_operator_opsmessage_reject`
- `test_four_eyes_required_for_opsmessage_compose`
- `test_four_eyes_not_required_for_opsmessage_approve` / `..._reject` (the checker actions are exempt by design)
- `test_deny_edge_service_opsmessage_compose` — confirms the separate namespace actually sidesteps `edge-service-notification`'s `notification.*` match (the whole point of D4)
- `test_deny_customer_opsmessage_compose` — an unprivileged customer session gains nothing here

## Ripple

`find openbank-infra/gitops/components -name 'gen-*opa-bundle*.sh' | sort | xargs -n1 bash` — regenerated all 27 generators. 26 embed `rest.rego`/`rules.yaml` and changed; the agent-service bundle (`gen-opa-bundle-configmap.sh`) only embeds `agents.rego`/`agents.yaml` and correctly did not (confirmed by reading it — matches the documented "25 of 26" shape, now 26 of 27 since the fleet grew a generator). Verified idempotent: a second pass produced zero further changes. 47 files total: 3 source + 26 bundle ConfigMaps + 18 paired Deployment/Rollout checksum-annotation updates (`payments-services.yaml` covers 9 services in one shared file, hence fewer than 26).

Expect this to roll the pods of every service whose checksum moved — that is the point (subPath mounts don't hot-reload).

## Scope

No handler code calls `opsmessage.*` yet — nothing is reachable through this namespace until PR 2 (`notification-service` backend: compose/submit/approve endpoints) lands. This PR only makes it safe to build that on top of.

## Test plan

- [x] `opa test openbank-libs/governance/policies openbank-infra/opa/policies` — 133/133 pass.
- [x] Bundle regen is idempotent (verified via a second pass).
- [x] Confirmed by direct read which generator doesn't embed `rest.rego`/`rules.yaml`, rather than assuming.
- [ ] CI's OPA policy gate (discovers generators dynamically per #1187 — should catch any generator I missed).

Refs `docs/adr/0176-operator-initiated-customer-messaging.md`

### DIFF
--- a/openbank-infra/gitops/components/accounts/account-opa-bundle.yaml
+++ b/openbank-infra/gitops/components/accounts/account-opa-bundle.yaml
@@ -9,7 +9,7 @@ metadata:
     app.kubernetes.io/name: account-service
     app.kubernetes.io/part-of: accounts
   annotations:
-    openbank.tech/policy-checksum: "f87aa3de385e106d"
+    openbank.tech/policy-checksum: "1e6df3341ce2d6a4"
 data:
   rest.rego: |
     # SPDX-License-Identifier: Apache-2.0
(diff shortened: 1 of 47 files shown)


JiRaska:
Pushed a follow-up commit: added `opsmessage.draft` as a separate, ungated action.

Found while wiring the actual endpoints in PR 2 — the maker's draft step (persist content before four-eyes runs) can't reuse `opsmessage.compose` as its action name, since `four_eyes_required` matches purely on `input.action`. Reusing that name would have paused draft creation itself for a second approver, defeating the whole point of splitting a cheap, reversible draft from the gated submit step.

No `rules.yaml` change needed — only a new `rest.rego` allow rule, deliberately absent from `four_eyes.actions`. 135/135 tests passing (was 133/133).

JiRaska:
Superseded by #1368 — its branch contains all 47 files of this PR (the opsmessage.* rego namespace + regenerated OPA bundles) as a strict subset, so merging both would apply the same bundle regeneration twice. Closing in favour of the single PR; no work is lost.

Refs #1368

Issue #461, opened 2026-07-07
Fleet sweep: OSV reports 63 vulnerable dependency versions via gradle/verification-metadata.xml (Scorecard Vulnerabilities 10 → 0)


JiRaska:
## Symptom

Since #333 landed `gradle/verification-metadata.xml`, OSV-Scanner (which Scorecard's Vulnerabilities check runs) reads it as a lockfile and reports **63 known-vulnerable dependency versions** — the check dropped 10 → 0 (scan 2026-07-07, commit 5a09553e).

## Two distinct causes

1. **Stale accumulated entries.** `--write-verification-metadata` only appends; the file still lists e.g. `netty-codec-http` 4.1.100.Final AND 4.1.106.Final AND 4.1.133.Final, `jackson-databind` 2.14.2/2.17.2/2.21.2. Versions no longer resolved by any build stay in the file and OSV flags them forever.
2. **Genuinely outdated resolved versions** (whatever survives a clean regeneration) — mostly Quarkus-managed transitives.

## Per-package breakdown

| Package | CVEs | GHSA ids |
|---|---|---|
| `io.netty:netty-codec-http` | 11 | GHSA-38f8-5428-x5cv, GHSA-57rv-r2g8-2cj3, GHSA-5jpm-x58v-624v, GHSA-84h7-rjj3-6jx4, GHSA-f6hv-jmp6-3vwv, GHSA-fghv-69vj-qj49 … |
| `com.fasterxml.jackson.core:jackson-databind` | 7 | GHSA-5jmj-h7xm-6q6v, GHSA-hgj6-7826-r7m5, GHSA-j3rv-43j4-c7qm, GHSA-rmj7-2vxq-3g9f, GHSA-5hh8-q8hv-fr38, GHSA-9fxm-vc8v-hj55 … |
| `tools.jackson.core:jackson-databind` | 7 | GHSA-5jmj-h7xm-6q6v, GHSA-hgj6-7826-r7m5, GHSA-j3rv-43j4-c7qm, GHSA-rmj7-2vxq-3g9f, GHSA-5hh8-q8hv-fr38, GHSA-9fxm-vc8v-hj55 … |
[...]

Issue #332, opened 2026-07-06
Enable Gradle dependency verification (checksum pinning) — advisory then strict


JiRaska:
## What

The supply chain pins dependency VERSIONS (version catalog, lockfile, Dependabot) but not BYTES. A repository-level tamper or dependency-confusion swap of an already-approved version would resolve silently. This is one of the few OpenSSF Scorecard points the project loses.

## Plan (staged, ADR-0144)

- [ ] Generate + commit `gradle/verification-metadata.xml` (SHA-256 per artifact)
- [ ] Start LENIENT (`org.gradle.dependency.verification=lenient`) — mismatches are warnings while the metadata soaks across all CI lanes + dev platforms; incomplete baseline is safe (lenient never fails)
- [ ] Auto-refresh on Dependabot PRs (`dependabot-verification-metadata.yml`) so the auto-merge flow doesn't break on a bump
- [ ] Flip to `strict` at the `target_enforce_date` once the baseline is complete and clean

## Notes

Gradle dependency verification is a known maintenance cost (every new/updated artifact needs a metadata entry) — the Dependabot refresh workflow covers the common case; manual bumps run `./gradlew --write-verification-metadata sha256 testClasses`.

Part of the SSDLC-hardening audit follow-ups (2026-07-06).

Issue #328, opened 2026-07-06
pacs.008 reader leaks raw SAXParseException/NumberFormatException on malformed inbound XML


JiRaska:
## What

`Pacs008Reader.read()` (openbank-libs-domain) parses XML that arrives over the wire (inbound clearing credit transfers) but does not wrap the DOM parse or the `BigDecimal(amount)` construction. Malformed input escapes as a raw `SAXParseException` or `NumberFormatException` instead of the typed `Pacs008ParseException` the callers catch.

## Impact

At the clearing boundary an untyped throwable becomes an HTTP 500 (and a noisier log/alert) rather than a clean 4xx rejection of a bad message. Not an XXE exposure — the hardening (disallow-doctype-decl etc.) is intact; this is exception-typing/robustness.

## How found

First run of a new Jazzer fuzz target (`fuzz/ossfuzz/Pacs008ReaderFuzzer`, OSS-Fuzz integration): empty / `<` / truncated input crashed on run 1.

## Fix

Wrap parse + amount conversion, rethrow as `Pacs008ParseException` (carrying the cause). Parameterized regression test over malformed inputs.

Part of the SSDLC-hardening audit follow-ups (2026-07-06).

PR #73, opened 2026-07-01
chore(main): release admin-ui 0.39.1

JiRaska:
:robot: I have created a release *beep* *boop*
---

### Bug Fixes

* **admin-ui:** pin postcss to 8.5.16 via npm overrides (CVE-2026-41305) ([#45](https://github.com/JiRaska/open-bank-oss/issues/45)) ([ae96248](https://github.com/JiRaska/open-bank-oss/commit/ae96248f07c087b1e81cb091aeb3129de249dc04))
* **admin-ui:** resolve all 32 open CodeQL findings ([#81](https://github.com/JiRaska/open-bank-oss/issues/81)) ([171790e](https://github.com/JiRaska/open-bank-oss/commit/171790e2ebb8bf7610adda864b410ac6630bf8a4))

---
This PR was generated with [Release Please](https://github.com/googleapis/release-please). See [documentation](https://github.com/googleapis/release-please#release-please).

PR #1367, opened 2026-07-16
fix(deps): close ReDoS CodeQL finding, floor protobuf/otel/handlebars CVEs

JiRaska:
## Summary

Overnight security-dashboard sweep (52 open findings: 3 CodeQL + 49 Dependabot).

**CodeQL — all 3 resolved, 0 remain open:**
- **#331 (js/redos, real fix)** — `service-registry.guard.test.ts`'s gitops metadata scan used `(?:[ \t]+.*\n)*?` per line; on a whitespace-only line `[ \t]+` and `.*` can split the same characters many ways, and CodeQL's own repro string (`\t\t\n` repeated) confirmed exponential backtracking survived the prior `\s`→`[ \t]` fix. Rewritten as a plain line-by-line scan (`nameFromMetadataBlock`) — no regex backtracking possible. `vitest run` on the file: 10/10 pass.
- **#332 (java/constant-comparison, dismissed false positive)** — `VopNameMatchPolicy.matchesWithInitials()`: CodeQL reports `fullTokenAgreements > 0` as always false. The var is mutated inside a lambda passed to `indices.all`, which Kotlin compiles to a boxed `IntRef` — CodeQL's dataflow doesn't track mutation through it. Runtime behavior is correct and already covered by `VopNameMatchPolicyTest`. Dismissed with that justification in the alert.
- **#328 (java/ssrf, dismissed false positive)** — flagged `UpstreamClient.delete()`'s call into `validatedUri()`, which is a documented inline regex host-allowlist barrier built specifically because CodeQL's SSRF barrier recognition doesn't follow a sanitizer across a call-site boundary and re-flags fresh alerts at each new caller (see the function's own kdoc — this exact pattern hit #239-241, #288, #289 before). Dismissed with that justification.

**Dependabot — 2 genuinely fixed, 3 explicitly left open:**

Extends the existing `openbank.dependency-vulnerability-pins` convention plugin (issue #461's pattern: same-line patch/minor floors via `resolutionStrategy.force`, applied fleet-wide):
- `protobuf-java`/`protobuf-javalite`/`protobuf-kotlin`/`protobuf-kotlin-lite` → `4.35.0` (already resolved elsewhere in the graph per `verification-metadata.xml` — same safety bar the file's existing entries use)
- `handlebars` → `4.5.2` in `libs.versions.toml` (its own comment explicitly invites a Dependabot-driven bump)

**~~opentelemetry-api → 1.62.0~~ — attempted, reverted.** First push forced it too; CI caught a real regression (`NoClassDefFoundError: io/opentelemetry/common/impl/ApiUsageLogger`) — 1.62.0 is binary-incompatible with Quarkus 3.37.2's bundled OTel extension, and it broke `test` for effectively every service in the fleet. Reverted in the second commit, verified locally (`document-service`/`vop-service`/`ledger-service` `:test`, real exit code, not tail-masked this time). **Dependabot #42 stays open** — needs a Quarkus platform bump or a 1.6x point release Quarkus 3.37.2 actually ships, not a blind floor to the advisory's version.

**Deliberately not touched — needs a maintainer call, not a 2am force-push:**
- **`io.netty:netty-handler` 4.2.x line (Dependabot #71-73)** — same class of risk the opentelemetry force just proved out: the fleet is floored on 4.1.135.Final elsewhere, and mixing netty major lines (or fighting whatever classpath actually wants 4.2.x) needs tracing before touching.
- **`org.hibernate.reactive:hibernate-reactive-core` (Dependabot #27)** — advisory's patched version is `4.2.1`; fleet is on the `3.x` line under Quarkus 3.37.2. Major-version jump, not a patch — same failure mode as the OTel force, but on the DB layer for every money-path service. Needs a real Quarkus-compatibility check.

## Test plan
- [x] `npx vitest run src/test/service-registry.guard.test.ts` — 10/10 pass
- [x] `./gradlew :openbank-customer-edge:compileKotlin :openbank-document-service:compileKotlin :openbank-vop-service:compileKotlin` — clean, dependency verification enforced
- [x] `./gradlew :openbank-document-service:test :openbank-vop-service:test :openbank-ledger-service:test` — clean after the OTel revert (real exit code 0, not piped through `tail`)
- [x] `./gradlew :openbank-vop-service:ktlintCheck :openbank-customer-edge:ktlintCheck` — clean
- [x] `gradle/verification-metadata.xml` regenerated for `handlebars` via `--write-verification-metadata sha256`
- [x] CI (this PR) — first push was red fleet-wide from the OTel force; second push fixes it, watching for green

Two CodeQL alerts (#332, #328) dismissed via the Security tab with the justification above. Dependabot #71, #72, #73, #27, #42 stay open by design.


JiRaska:
## Follow-up: audited the other 46 Dependabot alerts

Checked every remaining open alert against the *alert-specific* patched-version field (not the advisory-wide list, which can mismatch the actual installed range). Result:

- **41 of them are already fixed** by `resolutionStrategy.force` pins that existed in this repo before tonight (netty 4.1.x line, jackson, grpc-netty-shaded, json-smart, plexus-utils, postgresql) — they're just stale on the dashboard.
- **Root cause of the staleness, found and fixed in the latest commit**: the `Dependency submission` workflow (`.github/workflows/dependency-submission.yml`) has failed on *every* push to `main` since at least 2026-07-14. `org.gradle:github-dependency-graph-gradle-plugin:1.4.1` (pulled by `gradle/actions/setup-gradle`'s dependency-graph extractor) was never in `gradle/verification-metadata.xml`, so dependency verification rejected it before the graph could be generated — GitHub's Gradle dependency graph has been stuck on a stale snapshot, so Dependabot never re-scanned and closed the already-fixed alerts. Fixed by adding its checksums (fetched from `plugins.gradle.org/m2/`, the actual resolution source — Maven Central 404s this coordinate).
- Once this merges and the next `Dependency submission` run succeeds, expect most of those 41 to auto-close within its normal cadence (push-triggered + weekly Monday 04:30 UTC cron) — that's GitHub's own reconciliation, not something this PR does directly.
- Confirmed still genuinely open, unrelated to the graph staleness: **#71-73** (netty-handler 4.2.x), **#27** (hibernate-reactive-core), **#42** (opentelemetry-api) — same three flagged in the PR description as needing a maintainer call.

Issue #1354, opened 2026-07-16
lending-service four-eyes ApprovalStore has no Redis deployed — flipping enforce would break, not fail open


JiRaska:
AuthorizeInterceptor's comment on the ApprovalStore-not-wired branch says the service "proceeds WITHOUT the second-approver gate" — implying a safe fail-open when four-eyes infrastructure is missing. That branch checks `approvalStore.isResolvable` (does a CDI bean exist), not whether the bean actually works.

For `lending-service`, the bean exists — `ApprovalConfig.approvalStore()` is `@Produces`d — but nothing backs it:

- `openbank-lending-service/src/main/resources/application.yaml:64`: `redis.hosts: redis://localhost:6379` (the dev-local default, never overridden)
- `grep -n "REDIS" openbank-infra/gitops/components/lending/lending-service.yaml` → **0 matches**
- `find openbank-infra/gitops/components/lending -iname "*redis*"` → **no file**

Compare `balance-service`, which does this correctly: a real `redis.yaml` Deployment+Service in `openbank-infra/gitops/components/balances/`, and `QUARKUS_REDIS_HOSTS: redis://redis.balances.svc:6379` in its Deployment env.

## Consequence

`isResolvable` returns true for lending (the bean resolves), so `AuthorizeInterceptor` does NOT take the safe "no bean, proceed" logging path. It proceeds to call `ReactiveRedisDataSource` against `localhost:6379` inside the lending pod — which has no Redis. If `AUTHZ_FOUR_EYES_ENFORCE` is ever flipped true for lending, every `lending.disburse` / `lending.collateralRegister` call would hit a Redis connection failure, not the documented "proceeds without the gate, logs an error" fallback.

This is arguably fail-*closed* from a security standpoint (the call breaks rather than skipping the gate) — but it is an undocumented, untested failure mode on a money-path action, discovered by inspection, not by design.

## Fix

Mirror `balances/redis.yaml` + the `QUARKUS_REDIS_HOSTS` env override for lending-service. Given `lending-service` is money-path (`rules.yaml: money_path_services`), this needs a threat-model review per repo convention, not just a gitops PR.

## Related

Found while wiring the same pattern correctly for notification-service (ADR-0176 D5, `opsmessage.compose` four-eyes).

JiRaska:
## Triage: confirmed still live on `origin/main` (282256de8)

Verified every claim in this issue against current code:

[...]

Issue #1325, opened 2026-07-16
Secret-bearing variables bypass TemplateSensitivity via renderTemplate's else branch


JiRaska:
`TemplateSensitivity` (#1179, shipped in #1180) keys secrecy on the **template**. But secrecy is a property of the **variables** — and `renderTemplate` has an `else` branch that dumps every caller-supplied variable into the stored body for any template it does not map.

## The hole

`NotificationConsumer.renderTemplate` maps **6** of the **13** `NotificationTemplate` constants: `ACCOUNT_OPENED`, `TRANSACTION_COMPLETED`, `KYC_APPROVED`, `KYC_REJECTED`, `OTP_CODE`, `WELCOME`.

The other 7 — `ACCOUNT_CLOSED`, `ACCOUNT_FROZEN`, `TRANSACTION_FAILED`, `KYC_DOCUMENT_REQUIRED`, `CONSENT_GRANTED`, `CONSENT_REVOKED`, `PASSWORD_RESET` — hit:

```kotlin
else -> template.name.replace("_", " ").lowercase().replaceFirstChar { it.uppercase() } to
    "<p>${vars.entries.joinToString("<br>") { "${it.key}: ${it.value}" }}</p>"
```

So a caller sending:

```json
{ "template": "ACCOUNT_FROZEN", "variables": { "code": "483920" } }
```

renders `<p>code: 483920</p>`, and `TemplateSensitivity.bodyForStorage(ACCOUNT_FROZEN, body)` returns it unchanged — `ACCOUNT_FROZEN` is not in `SECRET_TEMPLATES`. The secret is stored in cleartext, readable by any `ROLE_OPERATOR`. This is precisely the exposure #1179 set out to close; the allow-list simply cannot express it.

The pinning test in `TemplateSensitivityTest` cannot catch this either, and neither can the review gate the ADR proposes: the leak lives in the branch that has **no per-template code to review**. A reviewer looking at `renderTemplate`'s mapped branches sees nothing wrong.

## Exploitability

**Latent.** The only in-repo publisher is `openbank-account-service`, which sends `TRANSACTION_COMPLETED` with `amount`/`currency`. No caller sends a secret-shaped variable with an unmapped template today. It becomes live the moment any producer does.

[...]

Issue #1182, opened 2026-07-16
sendPush puts rendered body (incl. transaction amount) into the APNs/FCM payload — ADR-0135 §3 violation


JiRaska:
`sendPush` puts the full rendered message body into the APNs/FCM payload:

```kotlin
private fun sendPush(req: NotificationRequest, subject: String, body: String, entity: NotificationEntity): Uni<Void> {
    val pushText = htmlToPlain(body)
    ...
    pushSender.send(PushMessage(platform, token, subject, pushText, mapOf("template" to req.template.name)))
```

ADR-0135 §3 ("Notification payload minimisation") forbids exactly this:

> Push payloads must not contain:
> - Account numbers, IBANs, or balances.
> - Transaction amounts or merchant names.
> - PII beyond a non-identifying reference ID.

## This is live, not theoretical

`openbank-account-service` is the only in-repo publisher on `openbank.notification.requests`, and it publishes `channel: "PUSH"` with `template: "TRANSACTION_COMPLETED"`:

```kotlin
"channel" to "PUSH",
"template" to "TRANSACTION_COMPLETED",
"variables" to mapOf("amount" to amount.toPlainString(), "currency" to currency),
```

which renders to `Transaction of <b>{amount} {currency}</b> completed successfully.` — so the transaction amount reaches the payload. `ACCOUNT_OPENED` would likewise carry `accountNumber`.

Mitigating factor: `FCM_ENABLED` / `APNS_ENABLED` both default to `false`, so a disabled adapter returns *skipped* and nothing egresses in the sandbox today. The defect is in code and in any environment where push is turned on.

## The ADR says this is already done

The ADR-0135 delivery note (updated 2026-06-30) records:

> **Security model** — ✅ Complete: token binding to SCA session, 90-day TTL refresh, explicit logout `DELETE` endpoint (PR #2527), **payload minimisation (title/template only, no PII)**, and per-platform registration limit designed.

It is not complete. The note needs correcting either way, independent of when the code fix lands.

## Fix

ADR-0135 §3 already prescribes the design: the payload carries a generic title plus a non-identifying reference, and the app fetches the detail via an authenticated `GET /api/v1/notifications/{notificationId}` on tap (APNs `content-available: 1` / FCM data message). ADR-0176 D3 adopts this for the new operator-messaging path.

[...]

Issue #1179, opened 2026-07-16
notification-service stores rendered OTP / password-reset bodies readable by any operator


JiRaska:
Two independent problems meet in `notifications.body`.

**1. The body holds the secret.** `OTP_CODE` and `PASSWORD_RESET` render an authentication secret directly into the message body, and `NotificationConsumer` persists it verbatim:

```kotlin
NotificationTemplate.OTP_CODE ->
    "Your OpenBank verification code" to
        "<h2>Verification Code</h2><p>Your code is: <b>${vars["code"] ?: ""}</b>. Valid for 5 minutes.</p>"
```

**2. Any operator can read it.** `GET /api/v1/notifications/{id}` returns `body` and is `@RolesAllowed("ROLE_VIEWER", "ROLE_OPERATOR", "ROLE_ADMIN", "ROLE_SERVICE")`. The shared `rest.rego` `operator-read-any` rule independently grants `.read`/`.list` on **any** resource to every `ROLE_OPERATOR`/`ROLE_ADMIN` principal.

Staff able to read a customer OTP can complete that customer strong authentication — the control ADR-0021 exists to protect. The secret is also spent the moment it is delivered, so keeping it under the service declared retention (`governance.yaml`: 2 years) fails GDPR Art. 5(1)(c) data minimisation.

## Exploitability today

**Latent, not live.** No producer publishes `OTP_CODE` or `PASSWORD_RESET` — the only in-repo publisher is `openbank-account-service`, which sends `TRANSACTION_COMPLETED`. The templates exist and render correctly, so this becomes live the moment OTP delivery is wired through this service. Hence severity:medium rather than high.

Raising it now because two planned changes both make it worse: a per-customer message-history tab in admin-ui would put stored bodies one click from an operator, and operator-composed messages would add free text to the same column.

## Fix

Classify secret-bearing templates in the domain and store a placeholder instead of the rendered secret; delivery keeps the real body. Redact again on read, so two independent controls must both fail. One-off migration clears existing rows.

## Related, deliberately out of scope

[...]

PR #1368, opened 2026-07-16
feat(notification): operator-initiated customer messaging, four-eyes gated

JiRaska:
Refs #1179, #1325, #1354 · **based on #1350** (needs its closed variable schema — this PR's diff is empty until #1350 merges and this retargets to `main`)

Second of three PRs implementing ADR-0176 (operator-initiated customer messaging). `opsmessage.compose`: an operator sends a customer a message from a small, closed, pre-reviewed catalogue, gated by a real maker-checker flow.

## Why not the existing `NotificationTemplate` enum

Found while designing the catalogue, before writing any code: some of its 13 constants (`ACCOUNT_FROZEN`, `KYC_REJECTED`, `TRANSACTION_FAILED`, `CONSENT_REVOKED`) feed the ADR-0059 oversight webhook. An operator manually sending `ACCOUNT_FROZEN` — to test the feature, or by mistake — would raise a **false "account frozen" Slack alert** for an account that isn't frozen. Every other lifecycle template has the same problem in miniature: its existence in the notification history implies a domain event that didn't happen.

`OperatorMessageTemplate` is a new, disjoint, closed enum (`GENERIC_NOTICE`, `SUPPORT_FOLLOWUP` — deliberately minimal; the real catalogue, reusing `document-service`'s `TemplateRepositoryPort` per ADR-0176's corrected Neutral consequence, is B3). Same closed-schema shape #1350 gave `NotificationTemplate`: variables are a constructor argument, `render()` is exhaustive with no `else`, a new constant fails to *compile* until its copy is written.

## Channel: EMAIL only

- **PUSH refused.** `sendPush` already puts rendered content in the APNs/FCM payload — an ADR-0135 §3 violation (#1182, unresolved, blocked on the customer app's fetch-on-tap handling in another repo). A second producer onto that broken path makes it worse, not better.
- **IN_APP refused.** Still a stub that never leaves `PENDING`.
- Real multi-channel delivery is remaining ADR-0176 D2/D3 work.

## Authz: a namespace split that had to survive its own edge case

`opsmessage.compose` / `opsmessage.approval.decide` are their own rego action namespace, not `notification.*` — `edge-service-notification` grants every `notification.*` action to customer-edge's M2M identity via a prefix match.

The namespace split **alone isn't sufficient**, and this is the finding that made D4 more than tidiness: `service-account-openbank-edge` carries `ROLE_OPERATOR` in the realm and is classified `HUMAN` (Keycloak `client_credentials` tokens never produce `principal.type == SERVICE`) — so a rule gated on `HUMAN + ROLE_OPERATOR` alone would silently re-admit that exact M2M identity. Both new allow rules explicitly exclude any `service-account-*` principal id (same idiom `m2m-sanctions-screening` uses for the opposite purpose).

**Regression-tested by removing the exclusion**: exactly the two service-account tests fail, nothing else (135/135 restored after). Not asserted — demonstrated.

## Four-eyes: generalizes an existing precedent, doesn't invent one

`four_eyes.actions` is a new `rules.yaml` list — exact action names, evaluated by `rest.rego` independently of `money_path_scopes`, since notification-service is not and will never be money-path. This is smaller than it sounds: `four_eyes_required` already had a `featureflag.flip` clause matching an exact action name independent of money-path scope, shipped with ADR-0067. This is a third clause in the same shape, not a new mechanism.

**Fleet-wide ripple, expected and required**: `rules.yaml` is embedded in ~25 of 27 OPA bundle generators, so this PR regenerates all 27 — every service's `policy-checksum` annotation moves, rolling those pods on deploy. `opa test`: 135/135 (11 new). `build-bundle.sh` (`opa check --strict` + decision assertions against the real `agents.yaml`): all green.

## Redis: real, unlike the same wiring in lending-service

`ApprovalConfig` / `ApprovalResource` / `ExceptionMappers` mirror `lending-service`'s ADR-0155 wiring closely — with one deliberate deviation: **one decide action with an `approve` boolean**, not ADR-0176 D4's sketched separate `opsmessage.approve`/`opsmessage.reject` — that split was schema-level shorthand in the decision record, not a wire-format commitment, and the proven pattern is one endpoint.

Checking the precedent closely surfaced something outside this PR's scope but too important to sit on: **lending ships that exact wiring with `redis.hosts: localhost:6379` and no Redis deployed anywhere in its gitops.** `ApprovalStore.isResolvable` would be `true` (the bean exists) while every real call fails on connection-refused — not the interceptor's documented "proceeds without the gate, logs an error" fallback, which only fires when the *bean* is absent, not when it's present but unreachable. Filed as #1354, not fixed here (it's lending's money-path threat model, not this PR's call).

notification-service gets it right: a real `redis.yaml` (mirrors `balances/redis.yaml`, the one other service that already does this correctly) + `QUARKUS_REDIS_HOSTS`. `ApprovalStoreWiringIT` proves the whole chain with a real create/find/decide round-trip through Testcontainers Redis — and I verified it's a real guard, not decoration: pointing `quarkus.redis.hosts` at a closed port fails the test.

`AUTHZ_FOUR_EYES_ENFORCE` defaults `false`, matching every other four-eyes-wired service (lending, balance, sepa-payment, sepa-instant, account). The mechanism is wired so it *can* be enforced; flipping it is a deliberate, separate operational decision, not shipped enabled by this PR.

## Verification, split by what actually needs a live OPA sidecar

| Layer | How |
|---|---|
| rego allow/deny matrix + service-account exclusion | `opa test`, 135/135 |
| Closed schema, exhaustive render, persist+deliver | `OperatorMessageServiceIT` — real Postgres + `MockMailbox` |
| Redis/`ApprovalStore` wiring | `ApprovalStoreWiringIT` — real Testcontainers Redis |
| `@Authorize` action string ↔ rego action string | Reflection pin in `NotificationSecurityTest` — the one thing tying Kotlin to rego, and it's untypechecked; a typo compiles fine and fails closed only in a live OPA decision |

**Not covered, and said plainly rather than left implicit**: a live authenticated HTTP round trip through a real OPA sidecar (maker → 202 → checker approves via HTTP → maker retries with `X-Approval-Id` → succeeds). That needs a WireMock OPA stub or a live sidecar in CI — net-new test infrastructure this repo doesn't have yet anywhere, and standing it up for one PR felt like scope creep past "prove B2's own wiring is sound." The *generic* maker/checker mechanism (202/retry/self-approval) is already covered by `AuthorizeInterceptorTest` in `openbank-libs-runtime`; what's genuinely untested here is the full HTTP path specifically for these two new endpoints.

`detekt`, `ktlintCheck`, `test` (92 tests, 0 failures), `quarkusBuild` all green.

## Small, while already touching the file

Fixed a stale comment on `notification-service.yaml`'s Deployment (claimed OIDC is disabled for the pilot; it's been on since customer-edge started proxying authenticated reads) — adjacent to the Redis env var I was already adding there.

### DIFF
--- a/openbank-infra/gitops/components/accounts/account-opa-bundle.yaml
+++ b/openbank-infra/gitops/components/accounts/account-opa-bundle.yaml
@@ -9,7 +9,7 @@ metadata:
     app.kubernetes.io/name: account-service
     app.kubernetes.io/part-of: accounts
   annotations:
-    openbank.tech/policy-checksum: "92074cfb9e487f57"
+    openbank.tech/policy-checksum: "8e3c4c9a57f575bd"
 data:
   rest.rego: |
     # SPDX-License-Identifier: Apache-2.0
@@ -149,6 +149,38 @@ data:
     	role in input.principal.roles
     	some verb in {"list", "read"}
     	endswith(input.action, sprintf(".%v", [verb]))
+    }
+
+    # ADR-0176 D4: operator-initiated customer messaging is its own action namespace
+    # (opsmessage.*), deliberately NOT notification.* — the edge-service-notification rule below
+    # grants every notification.* action to customer-edge's M2M identity via a plain prefix match,
+    # and opsmessage.compose must never be reachable that way. A namespace split alone is not
+    # sufficient, though: service-account-openbank-edge is classified HUMAN (Keycloak
+    # client_credentials tokens never produce principal.type == SERVICE — see
+    # authz_policy.principal_type_service_unreachable in rules.yaml) and carries ROLE_OPERATOR in
+    # the realm (openbank-infra/gitops/components/keycloak/realm-template.json), so a rule gated on
+    # HUMAN + ROLE_OPERATOR alone would silently re-admit that exact M2M identity. The explicit
+    # `service-account-` exclusion below is what actually keeps it out — same idiom as
+    # m2m-sanctions-screening further down, used there for the opposite purpose (identifying M2M,
+    # not excluding it).
+    allowed_reasons contains "operator-compose-message" if {
+    	input.principal.type == "HUMAN"
+    	some role in {"ROLE_OPERATOR", "ROLE_ADMIN"}
+    	role in input.principal.roles
+    	not startswith(input.principal.id, "service-account-")
+    	input.action == "opsmessage.compose"
+    }
+
+    # Checker side of the same four-eyes flow (ADR-0155 mechanism, ADR-0176 D5 trigger). Deciding
+    # an approval is exactly as sensitive as issuing the request it gates, so it gets the identical
+    # HUMAN + operator-role + non-service-account shape, not the broader operator-write-any this
+    # service otherwise has no equivalent of.
+    allowed_reasons contains "operator-decide-message-approval" if {
+    	input.principal.type == "HUMAN"
+    	some role in {"ROLE_OPERATOR", "ROLE_ADMIN"}
+    	role in input.principal.roles
+    	not startswith(input.principal.id, "service-account-")
+    	input.action == "opsmessage.approval.decide"
     }
 
     # Authenticated customers may perform any `customer.*` action (initiate payments, enroll
@@ -291,6 +323,17 @@ data:
     four_eyes_required if {
     	input.action == "featureflag.flip"
     	input.attributes.flag in data.rules.feature_flags.money_path_flags
+    }
+
+    # Exact-action four-eyes, independent of money-path scope (ADR-0176 D5). Generalises the
+    # featureflag.flip clause above into a reusable list (data.rules.four_eyes.actions) rather than
+    # a one-off inline check — for an action whose service will never appear in
+    # money_path_services (opsmessage.compose today), the verb-based clause above can never match,
+    # since it requires deriving a scope FROM money_path_services first. Undefined, not an error,
+    # for any bundle whose rules.yaml predates this key — Rego membership over an undefined
+    # collection simply does not fire.
+    four_eyes_required if {
+    	input.action in data.rules.four_eyes.actions
     }
 
     # ---------------------------------------------------------------------------------------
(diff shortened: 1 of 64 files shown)


JiRaska:
Adversarial code review (10 finder angles + verification + sweep) surfaced 14 findings. Fixed the 2 most severe in a follow-up commit on this branch; the rest are tracked as issues.

## Fixed on this branch

- **#1368 finding 1 (most severe)**: `@Authorize` on `opsmessage.compose` had no `resource` binding — a checkers approval of one message content could silently authorize a retry with completely different content. Fixed with `resource = "#request"`, binding the approval to a content-derived fingerprint (the `data class`'s own `toString()`).
- **#1368 finding 4**: mail-send failure misattribution — a transient DB error after a successful send was silently misreported as a failed send, with the original exception discarded. Split into two independent `onFailure()` handlers so a delivered message is never marked FAILED, and both failure modes are now logged.

Both fixes are in commit `8231e3fc4`, verified by 2 new tests + the full existing suite (99 tests, 0 failures).

## Tracked as issues (not fixed here)

- #1381 — missing required-variable validation
- #1382 — XSS in GENERIC_NOTICE (unescaped operator free text in customer email)
- #1384 — recipient/partyId input-validation gaps
- #1386 — OperatorMessageTemplate has no TemplateSensitivity equivalent (mirrors #1325)
- #1387 — openapi.yaml missing the two new endpoints (mirrors #1303)
- #1388 — stale code comment re: `Ids.newId()`
- #1390 — stale `rules_real` rego test mock
- #1391 — no log signal when four-eyes flagged but not enforced
- #1392 — `NotificationConsumer.sendEmail` has the same mail-misattribution bug this PR fixed in `OperatorMessageService`
- #1393 — 3 DB round-trips where 2 would do
- #1394 — `ExceptionMappers.kt` is the 13th fleet-wide duplicate
- #1395 — `RedisTestResource.kt` missing Docker-availability guard

Also refuted during verification: a suspected `SecurityIdentity`/`SecurityContext` principal-mismatch in the four-eyes self-approval guard — no augmentor in this codebase, both APIs read the same underlying identity, no divergence possible.

PR #1369, opened 2026-07-16
feat(notification-service): add operator-initiated customer messaging (ADR-0176)

JiRaska:
## What

Second of three PRs implementing **ADR-0176** (operator-initiated customer messaging) end to end. Wires the actual compose/submit/approve endpoints in `openbank-notification-service` against the `opsmessage.*` namespace [#1358](https://github.com/JiRaska/open-bank-oss/pull/1358) added.

## Domain

- **`NotificationTemplate.OPERATOR_ACCOUNT_NOTICE`** — the first operator-composable catalogue template, with its own explicit `renderTemplate` case (fixed HTML skeleton, one variable: `referenceId`). Deliberately **not** built on an existing template that falls into the `else` branch: that branch does verbatim, unescaped interpolation of the caller-supplied `variables` map, and building compose on top of it would smuggle back exactly the free-text injection ADR-0176 D2 exists to prevent. (This is issue #1325's territory — a real, separate, pre-existing gap. Not fixed here.)
- **`PushContentPolicy`** (`FULL`/`WAKE_SIGNAL_ONLY`) on `NotificationRequest`, defaulting to `FULL` — every existing call site is unchanged (only `account-service` publishes today, and never sets it). `sendPush` branches on it: `WAKE_SIGNAL_ONLY` carries no body text, only `notificationId` in the data payload, matching ADR-0135 §3 for this new path without touching the pre-existing `TRANSACTION_COMPLETED` violation (retrofitting existing templates is separate, deferred work per the ADR).
- **`OperatorMessage`/`OperatorMessageEntity`** + `V10__create_operator_messages.sql`. `ApprovalStore` only tracks `(action, resourceId, makerId, status)` — no field for the actual message content, and no query to list pending approvals (confirmed: ledger-service's `ApprovalResource` has only `PATCH /{id}`, no list, anywhere in the fleet). This table **is** the persisted draft, and what backs the admin-ui pending-approvals list in PR 3.

## A real design correction along the way

`AuthorizeInterceptor` never runs a four-eyes-gated method's body at all on an un-approved first call — it short-circuits with its own 202 response *before* `ctx.proceed()`. That means there's no hook inside an annotated method to persist content at the exact moment a pending approval gets created; the content has to already exist, keyed by an id the interceptor can carry as `resource`. Two consequences that shaped this PR:

1. **Two REST steps, not one.** `OperatorMessageResource.draft` (own action, `opsmessage.draft`, **not** four-eyes gated — validates and persists) and `.submit` (`opsmessage.compose`, the gated one — the maker's approved retry actually sends). Filed a [follow-up commit on PR 1](https://github.com/JiRaska/open-bank-oss/pull/1358#issuecomment-4996508216) adding `opsmessage.draft` once I hit this — reusing `opsmessage.compose` for both would have paused *drafting itself*.
2. **No separate `DRAFT` row status.** A row is born `PENDING_APPROVAL` and stays that way until it resolves — there's no code path that would ever transition it out of an earlier `DRAFT` state, so modeling one would just be a state nothing exits. `OperatorMessageApprovalResource.reject` explicitly marks the row `REJECTED`, rather than leaving that to the maker's retry: `ApprovalStore` refuses a retry against a non-`APPROVED` decision, so the submit handler's body — the only other place that ever updates this row — never runs for a rejected message.

## REST layer

- **`OperatorMessageResource`**: `POST /api/v1/opsmessages` (draft), `POST /api/v1/opsmessages/{id}/submit` (four-eyes gated — calls `NotificationConsumer.dispatch` directly, now `internal` instead of `private`, reusing render/persist/deliver/oversight unchanged rather than round-tripping through Kafka for a caller already inside this service), `GET /api/v1/opsmessages?status=PENDING_APPROVAL` (reuses the `opsmessage.approve` grant rather than inventing a fourth action name ADR-0176 doesn't name).
- **`OperatorMessageApprovalResource`**: mirrors ledger-service's `ApprovalResource` almost exactly, but split into `.../approve` and `.../reject` (not one `PATCH` + an `approve: Boolean` body) since ADR-0176 D4 names two distinct actions, each with its own rego rule.

## Wiring (mirrors ledger-service's ApprovalStore setup exactly)

`quarkus-redis-client` (first Redis dependency in this service — same as ledger's own first-Redis comment), `ApprovalConfig` producer, `SelfApprovalNotAllowedMapper`/`InvalidApprovalStateMapper` (copied from account-service's exception mappers), `authz.four-eyes.enforce` (defaults `false`; per the ADR, flips `true` only in this service's own sandbox gitops env, gating nothing fleet-wide).

## Test plan

- [x] `:openbank-notification-service:compileTestKotlin` + `ktlintCheck` pass.
- [x] `:openbank-notification-service:quarkusBuild` passes — this is the one that actually validates ArC/CDI wiring (the constructor-injected `ApprovalStore`/`OperatorMessageRepository` in `OperatorMessageApprovalResource`, the new `@Produces` bean); the other two don't catch that class of failure.
- [x] 6 new unit tests on `OperatorMessageResource` (`referenceId` pattern rejection, `MARKETING` rejection, unsupported-template rejection, the happy path, and the submit status guard) — all passing.
- [ ] Full IT-level verification of the actual four-eyes pause/resume (202 → approve → retry → `SENT`) against a live Postgres + Redis is left to CI; not exercised locally in this PR.

## Scope

No admin-ui changes here — PR 3 wires the compose form and pending-approvals panel against these endpoints.

Refs `docs/adr/0176-operator-initiated-customer-messaging.md`

### DIFF
--- a/openbank-notification-service/build.gradle.kts
+++ b/openbank-notification-service/build.gradle.kts
@@ -17,6 +17,11 @@ dependencies {
     implementation(libs.quarkus.flyway)
     implementation(libs.quarkus.jdbc.postgresql)
     implementation(libs.quarkus.smallrye.kafka)
+    // Redis-backed ApprovalStore for the ADR-0155 four-eyes mechanism, extended to
+    // opsmessage.compose by ADR-0176 D5. notification-service has no other Redis surface today
+    // — this is the first Redis dependency in this service (same shape as ledger-service's own
+    // first-Redis-dependency comment on its ApprovalConfig).
+    implementation(libs.quarkus.redis.client)
     implementation(libs.quarkus.smallrye.health)
     implementation(libs.quarkus.scheduler)
     implementation(libs.quarkus.micrometer.registry.prometheus)
--- a/openbank-notification-service/src/main/kotlin/com/openbank/notification/application/NotificationConsumer.kt
+++ b/openbank-notification-service/src/main/kotlin/com/openbank/notification/application/NotificationConsumer.kt
@@ -15,6 +15,7 @@ import com.openbank.notification.domain.model.NotificationChannel
 import com.openbank.notification.domain.model.NotificationRequest
 import com.openbank.notification.domain.model.NotificationStatus
 import com.openbank.notification.domain.model.NotificationTemplate
+import com.openbank.notification.domain.model.PushContentPolicy
 import com.openbank.notification.domain.model.PushPlatform
 import com.openbank.notification.domain.model.TemplateSensitivity
 import com.openbank.notification.infrastructure.persistence.entity.NotificationEntity
@@ -107,7 +108,11 @@ class NotificationConsumer {
             }
     }
 
-    private fun dispatch(req: NotificationRequest): Uni<Void> {
+    // internal, not private: OperatorMessageResource (ADR-0176) calls this directly, in-process,
+    // once four-eyes clears an operator-composed message — reusing render/persist/deliver/
+    // oversight unchanged rather than round-tripping through Kafka for a caller that is already
+    // inside this service.
+    internal fun dispatch(req: NotificationRequest): Uni<Void> {
         val (subject, body) = renderTemplate(req.template, req.variables)
         val entity = NotificationEntity().also {
             it.notificationId = UUID.randomUUID()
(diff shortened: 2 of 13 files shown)


JiRaska:
Superseded by #1368 — both implement the ADR-0176 notification backend, but as competing designs rather than layers: this PR adds its own `operator_messages` table (V10 migration + OperatorMessageEntity/Repository), while #1368 reuses the existing NotificationEntity/NotificationRepository and gates four-eyes through the shared `ApprovalStore` from openbank-libs (covered by its ApprovalStoreWiringIT). Merging both would leave two parallel operator-messaging implementations and a table nothing reads.

Maintainer chose the ApprovalStore design. Closing this one; reopen if the dedicated table turns out to be needed after all.

Refs #1368

PR #1421, opened 2026-07-17
security(ci): submit the Gradle dependency graph on PRs so dependency-review can gate it

JiRaska:
## Summary

A PR introducing a vulnerable Gradle dependency is currently blocked by **no gate** — Gradle is the fleet's primary language and `dependency-review` was blind to it. This wires the integration gradle/actions documents so the gate actually sees Gradle, without introducing lockfiles.

## Type of change

- [x] security (private until disclosed)
- [ ] feat / fix / refactor / perf / docs / chore / breaking change

## Linked issues

Closes #1419

## Changes

- `dependency-submission.yml`: added a `pull_request` trigger with the **same path filter** as the push trigger, so only a PR that can change resolution pays the ~11 min (3 of the last 40 PRs). Skipped on fork PRs via `head.repo.full_name == github.repository`.
- `dependency-review.yml`: `retry-on-snapshot-warnings: true` + `retry-on-snapshot-warnings-timeout: 1200`, and the job's own `timeout-minutes: 10 → 25` so it outlasts the retry window.

## Why this and not `gradle.lockfile`

`dependency-review` only ever diffs **submitted** graphs, and GitHub does not natively parse Gradle — hence `dependency-submission.yml`. It ran only on push to main, so a PR had no head-SHA Gradle snapshot. Nothing else covered the gap:

- **Trivy fs** (security.yml) — Trivy's Java coverage is JAR / pom.xml / `gradle.lockfile` / `*.sbt.lock` ([docs](https://trivy.dev/docs/latest/coverage/language/java/)); this repo has **zero** lockfiles, so Trivy sees no Gradle dependency.
- **CodeQL** — not a dependency scanner, and not a required check.

So detection was post-merge only, via Dependabot alerts on main (**20 open today: 11 high, 9 medium**), remediated by hand-maintained `force()` pins in `openbank.dependency-vulnerability-pins.gradle.kts`.

Lockfiles were the other route (they would let Trivy fs scan Gradle) and were **rejected**: a lockfile per module across 34+ services, regenerated on every dependency change *and* every `force()` edit in the pins plugin, is a permanent merge-conflict tax for a gate `dependency-review` already provides for free once the graph exists.

## Evidence

- Full-fleet resolve **measured at ~670 s**, so the doc's suggested `retry-on-snapshot-warnings-timeout: 600` would have timed out before the snapshot it waits for could exist. Hence 1200.
- Only **3 of the last 40 PRs** touch a dependency manifest → the path filter keeps this off ~93% of PRs.
[...]

Issue #1419, opened 2026-07-17
Gradle dependency graph is never submitted for a PR head, so dependency-review is blind to the fleet's primary language


JiRaska:
Wave 1 of the 2026-07-17 workflow audit flagged this; the mechanism is now confirmed and measured.

## The gap

`actions/dependency-review-action` only ever diffs **submitted** dependency graphs. GitHub natively parses npm/pip/docker/actions manifests, but **not Gradle** — Gradle needs an explicit submission (which is exactly why `dependency-submission.yml` exists).

`dependency-submission.yml` runs only on **push to main** (+ weekly). So on a pull request there is no head-SHA Gradle snapshot, and `dependency-review.yml` — which gates `fail-on-severity: high` — has nothing to say about Kotlin/Java dependencies, the fleet's primary language. A PR introducing a vulnerable Gradle dependency is blocked by **no** PR gate:

- `dependency-review` → no Gradle graph for the head.
- `Trivy fs` (security.yml) → Trivy's Java support covers JAR / pom.xml / `gradle.lockfile` / `*.sbt.lock` only ([docs](https://trivy.dev/docs/latest/coverage/language/java/)), and this repo has **zero** `gradle.lockfile`s.
- CodeQL → not a required check, and not a dependency scanner.

Detection therefore only happens **post-merge**, via Dependabot alerts on main — currently **20 open (11 high, 9 medium)** — with remediation by hand-maintained `force()` pins in `build-logic/src/main/kotlin/openbank.dependency-vulnerability-pins.gradle.kts`.

## Fix

Adopt the integration documented by gradle/actions (`docs/dependency-submission.md`): also run the submission on `pull_request`, and let `dependency-review` retry until the head snapshot lands.

Measured facts that shaped the change:
- The full-fleet resolve takes **~670 s**, so the doc's suggested `retry-on-snapshot-warnings-timeout: 600` would time out before the snapshot exists. Use 1200 (and raise the review job's own `timeout-minutes` above it).
- Only **3 of the last 40 PRs** touch a dependency manifest, so the same path filter as the push trigger keeps the ~11 min cost off ~93% of PRs.

## Deliberately not done

- **`gradle.lockfile`s** (the other route to a PR gate — they would let Trivy fs scan Gradle). Rejected: a lockfile per module across 34+ services, regenerated on every dependency change *and* every `force()` edit in the pins plugin, is a large permanent merge-conflict tax for a gate `dependency-review` already provides for free.
[...]

PR #1422, opened 2026-07-17
[DO NOT MERGE] test: prove the Gradle dependency-review gate fails a vulnerable dep

JiRaska:
## Summary

**Throwaway. Do not merge. Will be closed once the gate's verdict is recorded.**

Stacked on `security/gradle-dependency-review-on-prs` (#1421) and deliberately adds `org.apache.commons:commons-text:1.9` — **Text4Shell**, CVE-2022-42889 / GHSA-599f-7c49-w659, critical — to `openbank-libs-domain`.

## What this proves

#1421 claims a PR that introduces a vulnerable Gradle dependency will now be **blocked**. That claim is worthless unless the gate is seen failing on a real one. So:

- **Expected:** `Dependency submission` runs (this PR touches `build.gradle.kts`, matching the path filter), then `Dependency review` **FAILS** on a high/critical Gradle advisory.
- **Before #1421** the same PR would have passed dependency review untouched — Gradle had no head-SHA graph to diff, so the gate never saw it.

## Linked issues

Refs #1419

### DIFF
--- a/.github/workflows/dependency-review.yml
+++ b/.github/workflows/dependency-review.yml
@@ -14,7 +14,10 @@ permissions:
 jobs:
   dependency-review:
     runs-on: ubuntu-latest  # public repo => hosted runners free+unlimited (FinOps: hosted-first)
-    timeout-minutes: 10
+    # 25 > the retry window below, which must in turn outlast dependency-submission.yml
+    # (measured ~11 min for the full-fleet resolve) or this job would give up before the
+    # graph it is waiting for exists.
+    timeout-minutes: 25
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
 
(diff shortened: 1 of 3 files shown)


JiRaska:
## Verdict: the gate works. Closing.

`dependency-review` **FAILED** this PR on the planted dependency, for the right reason:

```
↪ https://github.com/advisories/GHSA-599f-7c49-w659
##[error]Dependency review detected vulnerable packages.
+ org.apache.commons:commons-text@1.9
```

That is Text4Shell (CVE-2022-42889). Before #1421 this same PR would have passed dependency review untouched — there was no head-SHA Gradle snapshot to diff, so the gate never saw the dependency at all.

### What the test caught that review-by-eyeball would not have

1. **The retry was armed unconditionally**, so `dependency-review` sat **>11 min on a PR touching no manifest** — GitHub answers "No snapshots were found for the head SHA" regardless of whether a submission is coming, so the retry waited out its window for a snapshot that was never going to arrive. That would have taxed ~93% of PRs. Fixed by arming the retry only when the PR touches the same paths `dependency-submission.yml` filters on. Re-measured after the fix: **7 s** on a non-dependency PR.

2. **The window was too tight.** The resolve takes 734 s, but the snapshot only became queryable **~24 min** after the review job started — GitHub indexes the graph *after* the submission API call returns, and that lag is invisible from the submission job's runtime. The run cleared `timeout-minutes: 25` by **48 seconds**. Raised to 1800 s retry / 35 min job.

3. **A near-miss on the proof itself.** This PR's first run also went red — but on a *timeout*, not the vulnerability: force-pushing cancelled the in-flight submission (`cancel-in-progress`), so no snapshot ever landed. A red check is not evidence until you read *why* it is red.

Also confirmed: `Submit fleet dependency graph` ran here (path filter matched `build.gradle.kts`) and was **absent** on #1421 (no manifest touched).

Throwaway branch — closing unmerged, branch deleted.

PR #1427, opened 2026-07-17
ci(governance): submit the Gradle dependency graph on PR heads

### LINKED ISSUE #1419: Gradle dependency graph is never submitted for a PR head, so dependency-review is blind to the fleet's primary language

JiRaska:
## Summary
- `dependency-submission.yml` only ran on `push` to `main` (+ weekly), so `dependency-review.yml` — the PR gate with `fail-on-severity: high` — never had a submitted Gradle graph to diff for a PR head. GitHub parses npm/pip/docker manifests natively; Gradle needs an explicit submission. Result: a vulnerable Kotlin/Java dependency introduced in a PR passed the PR gate silently and only surfaced post-merge via Dependabot.
- Add a `pull_request` trigger to `dependency-submission.yml`, gated on the same manifest path filter as the existing push trigger (measured: only ~3 of the last 40 PRs touch a dependency manifest, so the ~11 min full-fleet resolve stays off ~93% of PRs).
- Wire `retry-on-snapshot-warnings-timeout: 1200` into `dependency-review.yml` so it waits for that concurrently-running submission workflow instead of diffing against a graph missing the new snapshot (measured full-fleet resolve ~670s). Raise the review job's own `timeout-minutes` to 25 so it doesn't self-cancel before the retry window elapses.

This is the mechanical CI-workflow fix suggested in the issue; no permission escalation — `dependency-submission.yml`'s job already runs with `contents: write` (required by the dependency-submission API) on push, this only adds a second trigger for the same job under the same scoped permissions. Fork-PR support is explicitly out of scope (submission cannot succeed with a fork's read-only token; the issue calls this out as a deliberate non-goal for a follow-up).

## Test plan
- [ ] `pull_request` + path-filtered trigger and YAML syntax verified locally (`yaml.safe_load`)
- [ ] On merge, confirm `dependency-submission.yml` fires on a PR that touches `build.gradle.kts`/`libs.versions.toml` and a Gradle snapshot lands in the PR's dependency graph
- [ ] Confirm `dependency-review.yml` picks up that snapshot within the 1200s retry window and diffs it

Closes #1419

### DIFF
--- a/.github/workflows/dependency-review.yml
+++ b/.github/workflows/dependency-review.yml
@@ -14,7 +14,11 @@ permissions:
 jobs:
   dependency-review:
     runs-on: ubuntu-latest  # public repo => hosted runners free+unlimited (FinOps: hosted-first)
-    timeout-minutes: 10
+    # dependency-submission.yml's Gradle graph for this PR head runs concurrently (issue
+    # #1419) and takes ~670s for the full-fleet resolve. retry-on-snapshot-warnings-timeout
+    # below polls for up to 1200s waiting for that snapshot to land — this job's own
+    # timeout must stay comfortably above that or it self-cancels before the retry gives up.
+    timeout-minutes: 25
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
 
@@ -24,6 +28,13 @@ jobs:
           fail-on-severity: high
           comment-summary-in-pr: always
           deny-licenses: GPL-2.0, GPL-3.0, AGPL-3.0
+          # Gradle has no native GitHub dependency-graph parser (unlike npm/pip/docker) —
+          # dependency-submission.yml explicitly submits it for this PR head, but that
+          # submission is a separate, concurrently-running workflow. Retry until the
+          # snapshot lands instead of diffing against a graph that doesn't have it yet
+          # (issue #1419). 1200s: measured full-fleet resolve is ~670s; give headroom.
+          retry-on-snapshot-warnings: true
+          retry-on-snapshot-warnings-timeout: 1200
           # starlette (pulled in transitively via schemathesis==3.39.16's own
           # starlette<1,>=0.13 pin — .github/workflows/requirements/schemathesis.txt)
           # is stuck below the versions that fix these two advisories (patched at
--- a/.github/workflows/dependency-submission.yml
+++ b/.github/workflows/dependency-submission.yml
@@ -27,6 +27,19 @@ on:
       - "openbank-libs/gradle/libs.versions.toml"
       - "settings.gradle.kts"
       - "gradle/wrapper/**"
+  # PR-head submission (issue #1419): dependency-review.yml only ever diffs a SUBMITTED
+  # graph, and GitHub does not natively parse Gradle manifests (unlike npm/pip/docker) —
+  # without this trigger, dependency-review is blind to Kotlin/Java deps on every PR and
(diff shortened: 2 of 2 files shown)


JiRaska:
Superseded by #1421 — same two files, same goal (submit the Gradle graph on PR heads so dependency-review can gate it). Keeping #1421 because it handles three cases this one does not:

- **retry armed conditionally.** This PR sets `retry-on-snapshot-warnings: true` unconditionally; #1421 measured that arming it on a PR touching no manifest makes the job sit >11 min waiting for a snapshot that is never submitted — a tax on the ~93% of PRs the feature has nothing to say about. It gates the retry on a detect-changes step.
- **timeout sized from an end-to-end measurement.** #1421 found the snapshot only becomes queryable ~24 min in (resolve + GitHub's indexing lag), so its 25 min would clear by ~48 s at best — one slow queue from a spurious red. #1421 uses 35.
- **fork PRs.** #1421 skips the submit job on fork PRs, which get a read-only token and would otherwise red every external contribution. This PR has no such guard.

Refs #1421

PR #1434, opened 2026-07-17
fix(notification): use UUIDv7 for consumer notification IDs

### LINKED ISSUE #1388: Stale comment in OperatorMessageService: Ids.newId() does not match NotificationConsumer's rows
Code-review finding on PR #1368.

A comment in `OperatorMessageService.kt` (near the `notificationId` assignment) claims `Ids.newId()` (UUIDv7) "matches NotificationConsumer's existing rows." It does not: `NotificationConsumer.dispatch` (`NotificationConsumer.kt:131`) still mints `notificationId` via bare `UUID.randomUUID()` (UUIDv4), unchanged by PR #1368.

## Cost

A maintainer reading this comment reasonably concludes both write paths into `notifications.notification_id` are ID-format-consistent (e.g. safe to build a UUIDv7-sortable cursor on). Rows from the two paths are actually a mix of UUIDv4 and UUIDv7 — both syntactically valid UUIDs, so nothing catches the false assumption until something depends on sort order or version bits.

## Fix

Either:
- correct the comment to state the two paths currently diverge (and note it as a known inconsistency), or
- migrate `NotificationConsumer.dispatch` to `Ids.newId()` too, per ADR-0106's identifier-intent guard, making the comment true.

Small, low-risk — good first issue.

## Related

PR #1368

pollychen-lab:
## Summary
- switch NotificationConsumer durable notification IDs from bare UUIDv4 to Ids.newId() UUIDv7
- update the OperatorMessageService comment now that both notification write paths use UUIDv7
- assert the persisted consumer notification ID is UUIDv7 in NotificationConsumerIT

Fixes #1388.

## Validation
- JAVA_HOME=/home/polly/Documents/workspace/codex/oss-agent/oss-open-work/jdks/jdk-21.0.11+10 ./gradlew :openbank-notification-service:test --tests com.openbank.notification.integration.NotificationConsumerIT
- JAVA_HOME=/home/polly/Documents/workspace/codex/oss-agent/oss-open-work/jdks/jdk-21.0.11+10 ./gradlew :openbank-notification-service:ktlintCheck
- git diff --check

### DIFF
--- a/openbank-notification-service/src/main/kotlin/com/openbank/notification/application/NotificationConsumer.kt
+++ b/openbank-notification-service/src/main/kotlin/com/openbank/notification/application/NotificationConsumer.kt
@@ -8,6 +8,7 @@ import com.fasterxml.jackson.databind.ObjectMapper
 import com.openbank.libs.audit.AuditEvent
 import com.openbank.libs.audit.AuditEventPublisher
 import com.openbank.libs.audit.AuditResult
+import com.openbank.libs.domain.identifiers.Ids
 import com.openbank.notification.application.port.out.OversightWebhookPublisher
 import com.openbank.notification.application.port.out.PushMessage
 import com.openbank.notification.application.port.out.PushSender
@@ -34,7 +35,6 @@ import org.eclipse.microprofile.reactive.messaging.Incoming
 import org.jboss.logging.Logger
 import java.time.Clock
 import java.time.Instant
-import java.util.UUID
 import java.util.concurrent.Executor
 
 @ApplicationScoped
@@ -131,7 +131,7 @@ class NotificationConsumer {
     private fun dispatch(req: NotificationRequest): Uni<Void> {
         val (subject, body) = renderTemplate(req.template, req.variables)
         val entity = NotificationEntity().also {
-            it.notificationId = UUID.randomUUID()
+            it.notificationId = Ids.newId()
             it.partyId = req.partyId
             it.channel = req.channel.name
             it.template = req.template.name
--- a/openbank-notification-service/src/main/kotlin/com/openbank/notification/application/OperatorMessageService.kt
+++ b/openbank-notification-service/src/main/kotlin/com/openbank/notification/application/OperatorMessageService.kt
@@ -70,12 +70,8 @@ class OperatorMessageService {
 
         val (subject, body) = render(request.template, request.variables)
         // notificationId is the entity's durable, indexed identifier (ADR-0106) — minted via
-        // Ids, a UUIDv7 generator, not a plain random UUID. Diverges from NotificationConsumer's
-        // existing rows (issue #1388): NotificationConsumer.dispatch still mints notificationId
-        // through a bare, unversioned java.util.UUID factory call (version 4), so
-        // `notifications.notification_id` is currently a mix of version-4 and version-7 UUIDs
-        // across the two write paths -- do not assume sort/version-bit consistency across them
-        // without checking NotificationConsumer first.
+        // Ids, a UUIDv7 generator, not a plain random UUID. Matches NotificationConsumer's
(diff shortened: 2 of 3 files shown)


JiRaska:
Thanks for this — the fix itself looks correct and matches the repo convention: `Ids.newId()` (UUIDv7, ADR-0106) is already how `OperatorMessageService.compose()` mints `notificationId`, so this makes `NotificationConsumer.dispatch()` consistent with it, and dropping the now-unused `java.util.UUID` import is correct too (it's not used anywhere else in that file after the change).

The PR currently shows as conflicting with `main` (`mergeStateStatus: DIRTY`). Since this branch is on your fork, I can't push a rebase myself — here's specifically what changed on `main` since you branched, so a rebase should be quick:

- `openbank-notification-service/src/main/kotlin/com/openbank/notification/application/OperatorMessageService.kt`: PR #1430 ("fix(notification): correct stale comment on OperatorMessageService's notificationId") already rewrote the very comment your diff touches. Your branch's base still has the older 3-line version ending "...existing rows, which predate this guard." Current `main` has a longer comment explaining the *existing* v4/v7 split and referencing issue #1388 by number — once you rebase, that whole comment can likely just be deleted/simplified (or trimmed to a short note), since after your change there's no longer a version split between the two write paths for the comment to warn about.
- `NotificationConsumer.kt` and the IT file didn't otherwise change on `main`, so those two files should apply cleanly once the `OperatorMessageService.kt` hunk is resolved.

Could you rebase onto current `main` (`git fetch origin && git rebase origin/main`) and update that comment to reflect that the write paths are now unified? Happy to take another look once it's pushed.

pollychen-lab:
> Thanks for this — the fix itself looks correct and matches the repo convention: `Ids.newId()` (UUIDv7, ADR-0106) is already how `OperatorMessageService.compose()` mints `notificationId`, so this makes `NotificationConsumer.dispatch()` consistent with it, and dropping the now-unused `java.util.UUID` import is correct too (it's not used anywhere else in that file after the change).
> 
> The PR currently shows as conflicting with `main` (`mergeStateStatus: DIRTY`). Since this branch is on your fork, I can't push a rebase myself — here's specifically what changed on `main` since you branched, so a rebase should be quick:
> 
>     * `openbank-notification-service/src/main/kotlin/com/openbank/notification/application/OperatorMessageService.kt`: PR [fix(notification): correct stale comment on OperatorMessageService's notificationId #1430](https://github.com/JiRaska/open-bank-oss/pull/1430) ("fix(notification): correct stale comment on OperatorMessageService's notificationId") already rewrote the very comment your diff touches. Your branch's base still has the older 3-line version ending "...existing rows, which predate this guard." Current `main` has a longer comment explaining the _existing_ v4/v7 split and referencing issue [Stale comment in OperatorMessageService: Ids.newId() does not match NotificationConsumer's rows #1388](https://github.com/JiRaska/open-bank-oss/issues/1388) by number — once you rebase, that whole comment can likely just be deleted/simplified (or trimmed to a short note), since after your change there's no longer a version split between the two write paths for the comment to warn about.
> 
>     * `NotificationConsumer.kt` and the IT file didn't otherwise change on `main`, so those two files should apply cleanly once the `OperatorMessageService.kt` hunk is resolved.
> 
> 
> Could you rebase onto current `main` (`git fetch origin && git rebase origin/main`) and update that comment to reflect that the write paths are now unified? Happy to take another look once it's pushed.

@JiRaska on it, thanks

JiRaska:
Thanks @pollychen-lab — the fix is good and is landing as-is. The only blocker was commit signing: this repo's ruleset requires verified signatures and the commit here is unsigned, so the PR would strand at merge. Re-signed carry with your authorship preserved: see the referencing PR above (Closes #1434 will close this when it merges).

Issue #1444, opened 2026-07-17
15 of 51 CNPG databases have never been backed up — 3 lack a pod-identity association, 12 declare no backup at all


JiRaska:
Found while investigating why `PostgresWALArchiveFailing` had been firing for three days (#1432 follow-on). The three firing alerts turned out to be the visible tip of a much larger gap.

## Summary

**16 of 51 CNPG databases have never been backed up.** No WAL archive, no base backup, no PITR. Two are money-path (`billing-db`, `vop-db`) and two hold audit/compliance data (`authzaudit-db`, `govaudit-db`).

Only **3** of the 16 alert. The other 13 are silently broken.

## Three independent lines of evidence, all agreeing on the same 16

**1. The Terraform allow-list.** `openbank-infra/aws/envs/sandbox-platform/db-backups.tf` declares `aws_eks_pod_identity_association` per `(namespace, service_account)` pair — a hand-maintained map. It lists 36 entries; there are 51 CNPG clusters.

**2. Live pod state.** Credentials arrive via **EKS Pod Identity**, not IRSA and not IMDS (both `EC2NodeClass`es set `httpPutResponseHopLimit: 1`, so pods cannot reach IMDS at all — `inheritFromIAMRole: true` works *only* because Pod Identity injects it):

| pod | in db-backups.tf | live env |
|---|---|---|
| `sanctions-db-2` | declared | **HAS** `AWS_CONTAINER_CREDENTIALS_FULL_URI` + `eks-pod-identity-token` volume |
| `ledger-db-1` | declared | **HAS** |
| `anacredit-db-1` | missing | **NONE** |
| `billing-db-1` | missing | **NONE** |
| `document-service-db-1` | missing | **NONE** |
| `product-catalog-db-1` | missing | **NONE** |

**3. The bucket itself.** `s3://openbank-sandbox-db-backups/` holds 35 prefixes. Exactly these 16 clusters have **zero objects**:

```
anacredit-db          devops-db             goalert-db            releasesteward-db
authzaudit-db         docstruth-db          govaudit-db           sdd-db
billing-db  ***       document-service-db   liveness-db           tpp-registry-db
flakytest-db          glitchtip-pg          product-catalog-db    vop-db  ***
```
`***` = money-path (`rules.yaml: money_path_services`). `authzaudit-db` / `govaudit-db` hold audit evidence.

[...]

Issue #1432, opened 2026-07-17
sanctions-service is DOWN — a PEP_GLOBAL refresh writes ~1.9GB WAL onto a 2Gi volume; CNPG parks the cluster on low disk and nothing alerts


JiRaska:
Found while verifying an unrelated image rollout. **This is a live outage, not a latent risk.**

## Impact

`sanctions-service` has **zero endpoints** — it is down, not degraded. Sanctions screening is an AML control on the money path (ADR-0032, fail-closed).

```
$ kubectl get endpoints -n sanctions
NAME                       ENDPOINTS   AGE
sanctions-service                      42d      <-- empty
sanctions-service-canary               27d      <-- empty
```

Pods sit at `1/2` — the `opa` sidecar is ready, the app container is not; it cannot open a Postgres connection (`org.postgresql.core.PGStream.createSocket`). This is independent of image version: both the old (`sandbox-237c397c`) and the freshly-rolled (`sandbox-ffa5f44c`) pods fail identically.

## Root cause chain

```
sanctions-service           0 endpoints — down
 └ pods 1/2, cannot reach Postgres
    └ sanctions-db-1 = currentPrimary, CrashLoopBackOff, 17 restarts, 3d17h, lastExit=4
       └ "Detected low-disk space condition"
          └ CNPG cluster phase: "Not enough disk space"
```

```
$ kubectl get cluster -n sanctions
NAME           AGE   INSTANCES   READY   STATUS                  PRIMARY
sanctions-db   42d   2           1       Not enough disk space   sanctions-db-1
```

## Two things are wrong, and the second is the durable one

**1. No failover happened.** `currentPrimary=sanctions-db-1` and `targetPrimary=sanctions-db-1` — CNPG is still trying to hold the pod that cannot start as primary. Meanwhile `instancesStatus` reports `healthy: [sanctions-db-2]` and `replicating: [sanctions-db-1]`. A healthy replica has been sitting there while the cluster stayed down.

**2. The volume is too small for its own WAL settings.** This is not "it filled up over time" — the configuration cannot fit:

| | |
|---|---|
| PVC | **2Gi** (gp3) |
| `max_wal_size` | **1GB** — half the volume, before any data |
| database `openbank_sanctions` | 594 MB |
| actual WAL on disk | 47 segments × 16MB = **752 MB** (`pg_wal` 753M > `base` 617M) |
| `wal_keep_size` | 64MB |
[...]

Issue #1184, opened 2026-07-16
OPA bundle sync gate covers 4 of 25 rules.yaml-hashing generators; ~21 services' bundles silently drift


JiRaska:
## Problem

`.github/workflows/opa-policy.yml` ("Verify service OPA bundles + pod-roll annotations are in sync
with the policy source") regenerates and diffs **4** generators:

- `components/pid/gen-pid-opa-bundle.sh`
- `components/copilot/gen-copilot-opa-bundle.sh`
- `components/customer-edge/gen-customer-edge-opa-bundle.sh`
- `components/notifications/gen-notification-opa-bundle.sh`

(plus `components/agent/gen-opa-bundle-configmap.sh` in its own step.)

But **25 of the 26** `gen-*opa-bundle*.sh` generators hash `openbank-libs/governance/rules.yaml` as a
checksum input, and each also embeds `rest.rego`, `agents.rego`, `agents.yaml` and `rules.yaml`
verbatim into its ConfigMap:

```
grep -rl "rules.yaml" $(find . -name "gen-*opa-bundle*.sh")   # -> 25
```

Any edit to a shared policy source therefore leaves the other ~21 services' committed
`*-opa-bundle.yaml` ConfigMaps and their `openbank.tech/policy-checksum` pod-roll annotations stale,
and **CI cannot detect it**. The hard-coded list is the bug: a new service's generator is
opted out of the sync gate by default.

## Current drift

Regenerating every `rules.yaml`-hashing generator against `main` produces a diff in **22 services**
(33 files). The stale content is not limited to the checksum annotation — the embedded
`rules-data.yaml` / `agents-data.yaml` documents that the policies consult at decision time
(`data.rules.*`, `data.agents.*`) are also behind the source. The Rego *code* itself is identical
across all bundles; the drift is in the data documents and comments.

Governance-data specifics and their runtime impact are tracked privately per
`CLAUDE.md` (security issues -> private Security Advisories), not enumerated here.

## Secondary defect

`components/ledger/gen-ledger-opa-bundle.sh` is committed mode `100644` (non-executable) — it fails
`permission denied` when invoked as `./…`, so it has never run in any generator loop. Every other
generator is `100755`.

## Fix (staged by blast radius)

Regenerating a bundle bumps its pod-roll checksum annotation and rolls that service's pods, so this
is split rather than landed as one 22-service roll:

[...]

PR #1452, opened 2026-07-17
fix(infra): grant the 3 missing db-backup pod identities, and gate the list so it cannot rot

JiRaska:
## What was broken

Three CNPG clusters declared `barmanObjectStore` → the managed backup bucket but had **no** `aws_eks_pod_identity_association`, so barman could not authenticate to S3 and every WAL archive failed:

```
Barman cloud WAL archive check exception: Unable to locate credentials   (exit status 4)
```

| cluster | `PostgresWALArchiveFailing` firing since |
|---|---|
| `sdd-db` | 2026-07-14 02:45 |
| `tpp-registry-db` | 2026-07-15 02:32 |
| `vop-db` | 2026-07-16 09:56 |

All three at `severity: critical`, routed to goalert, for up to **three days**. They have no backups and no PITR.

**Root cause:** the hand-maintained `local.db_backup_clusters` map, whose own comment claimed it covered *"all remaining clusters with barmanObjectStore"*. It named 36 entries against 51 live clusters. Nothing linked it to the gitops manifests — a new service copies a sibling's `inheritFromIAMRole: true`, the Cluster reports Ready, the app reports Healthy, and the backup silently never happens. **There is no failure until someone needs a restore.**

## What this does

Adds the three entries, and a CI guard that **asserts** the invariant instead of trusting the comment (`rules.yaml: db_backup_association`, advisory, target 2026-08-15).

## Why a guard and not a discovered `for_each`

That was the original plan. Both in-Terraform routes were tried and are blocked:

- **`data.kubernetes_resources`** returns `objects` as a **dynamic** type → `(known after apply)` → cannot key a `for_each`, which needs plan-time keys. A real `tofu plan` of that design produced:

  ```
  Plan: 0 to add, 0 to change, 35 to destroy.
  ```

  It would have revoked S3 credentials fleet-wide and stopped every backup. A `terraform_data` precondition guard I had written for exactly that scenario fired and stopped the plan — which is the only reason this is a footnote rather than an incident.

- **`fileset` + `yamldecode`** cannot work either: **30 of the 42** files declaring a CNPG Cluster are multi-document YAML, which `yamldecode` does not parse.

A static check also fires at **PR time** rather than at `tofu apply` — earlier than a working `for_each` would have, since nothing applies tofu automatically — and it does not couple the AWS layer to live cluster state.

Same lesson as the OPA generator gate (#1184), where a hard-coded list of 4 silently covered 25 for months.

**Deliberately not diff-scoped** (unlike `db_change` / `api_contract`): the defect is a *missing pair* across two trees, so a PR adding only the gitops half never touches the `.tf` side and a diff-scoped gate would see nothing.

## `goalert-db` is a warning, not an error

It backs up to the legacy, unmanaged `openbank-cnpg-backups-<acct>` bucket and **its backups work** — archiving every 5 minutes, real base backups present, and `PostgresWALArchiveFailing` has never fired for it. That is a governance gap (the bucket's lifecycle, encryption and retention are not set by `db-backups.tf`), not an outage. An earlier draft of this check called it "no backups", which was flatly wrong; the check now reports it accurately and never fails a PR over it.

## Not fixed here

**12 clusters declare no `spec.backup` at all**, so they never even attempt an archive — completely silent, no alert possible. An association does nothing for them; they need a gitops change. Tracked in #1444.

## Verification

| check | result |
|---|---|
| guard against pre-fix `main` | finds **exactly the 3**, matching the 3 live firing alerts |
| guard after the fix | green on the missing-association class |
| `--enforce` with an entry removed | exit **1** |
| `--enforce` restored | exit **0** — the gate bites |
| test suite | **47 green** (9 new) |
| `tofu fmt` / `validate` | clean / Success |
| OPA bundle ripple | `rules.yaml` touched → all 27 generators re-run, output **idempotent** |

New tests cover the two traps that actually bit during development: Strimzi's Kafka CRD also uses `kind: Cluster` (matching on kind alone files Kafka brokers as databases), and multi-doc manifests must parse.

## Linked issues

Refs #1444
Refs #1432

---

## Update: second commit configures the 12-that-had-nothing

The follow-up I flagged (12 clusters with no `spec.backup` at all) is now in this PR — 11 of them; `glitchtip-pg` is deliberately excluded (see below).

Each of the 11 gains, matching the `sanctions-db` reference:
- `backup.barmanObjectStore` → `s3://openbank-sandbox-db-backups/<cluster>`, gzip, 30d retention
- a daily `ScheduledBackup`, staggered across 03:03–03:43
- a matching `aws_eks_pod_identity_association` in `db-backups.tf`

**Plus `max_wal_size: "512MB"` on each.** They had `wal_keep_size: 64MB` and no `max_wal_size` on a 2Gi volume — the exact shape that let a `PEP_GLOBAL` refresh write ~1.9GB of WAL and kill `sanctions-db` (#1432). Adding a base backup without bounding WAL would have handed each of them the same footgun.

Among the 11: `billing-db` (money-path), `authzaudit-db` / `govaudit-db` (audit evidence), `anacredit-db` (regulatory-reporting state, not reconstructible from seed data).

After this the guard is **fully green**: all 49 clusters targeting the managed bucket have an association, `--enforce` passes.

**`glitchtip-pg` excluded on purpose:** its CNPG Cluster is created by the third-party GlitchTip Helm chart, not a plain manifest, so its backup would go through the chart's `valuesObject` — a different change. It holds error-tracking data, not banking data. Follow-up.

### Governance note

`billing-db` is money-path. This PR now touches a money-path database's backup config, so it wants 2 approvals + a threat-model check per ADR-0030. Nothing here changes the money path itself (backup/DR config only, no schema or API surface), but flagging it rather than letting it slip through as infra.


JiRaska:
Checked the FinOps impact of the 11 new backups before touching retention.

Measured against a comparable already-backed-up idle cluster (`interest-db`, 117 MB steady-state across a 27-day retention window ≈ 4.3 MB/day of base backup + WAL combined):

| retention | all 11 DBs, steady-state | monthly cost (S3 Standard) |
|---|---|---|
| 30d (fleet standard — 38 of 39 clusters use it) | ~1.4 GB | **~$0.03/month** |
| 7d (goalert-db precedent) | ~0.3 GB | ~$0.008/month |

The bucket also has a hard 35-day lifecycle expiry regardless of `retentionPolicy`, so 30d already sits under that cap.

Difference between the two is a fraction of a cent per database — below anything a FinOps dashboard would register. Left `retentionPolicy: "30d"` unchanged on all 11 to match the fleet standard rather than trading away restore window (`billing-db` is money-path; `authzaudit-db`/`govaudit-db` hold audit evidence; `anacredit-db` holds regulatory-reporting state not reconstructible from seed data) for a saving too small to matter.

PR #1491, opened 2026-07-17
fix(admin-ui): call the actual merged #1368 opsmessage endpoints, not #1370's original dead ones

JiRaska:
## What

`#1370` (this repo's admin-ui side of ADR-0176) was built against `#1369`'s backend design (draft+submit split, `/api/v1/opsmessages/*`, `referenceId`-only templates). `#1369` was closed as superseded by `#1368`, which shipped a materially different backend design — but `#1370` merged anyway with its original endpoint calls unchanged, so **the compose/approve UI has been live on `main` calling a backend that does not exist.** Confirmed directly: `git show origin/main:openbank-admin-ui/src/lib/api.ts` still had `/api/v1/opsmessages/*` after both PRs landed.

## The real contract, read from the merged backend source (not the PR diff or description)

- **`POST /api/v1/notifications/messages`** (`opsmessage.compose`) — **one** call, not draft+submit. Body: `{partyId, template, recipient, variables}`. The four-eyes retry is the *same request resent* with `X-Approval-Id` — the backend binds `resource` to the request body's own content-derived fingerprint (`OperatorMessageResource`'s own KDoc explains why: there's no pre-existing entity id to bind to for a create-with-rich-body action, so it hashes `request.toString()` instead). There is no draft id in this design — the client must resend the *identical* object, not something merely equivalent.
- **`PATCH /api/v1/notifications/approvals/{id}`** with `{approve: boolean}` — one decide endpoint (mirrors `lending-service`'s own `ApprovalResource`), not the two separate `POST .../approve` + `.../reject` endpoints `#1370` originally called.
- **`recipient`** is a real, operator-supplied, server-validated email address — channel is EMAIL only (PUSH was dropped in `#1368`, blocked on the still-open ADR-0135 §3 payload issue, `#1182`) — not the auto-derived placeholder `#1370` used to satisfy its own now-dead template shape.
- **Two real templates** from `OperatorMessageTemplate.kt`: `GENERIC_NOTICE` (`subject`+`note`), `SUPPORT_FOLLOWUP` (`ticketReference`) — not the single fictional `OPERATOR_ACCOUNT_NOTICE` `#1369` invented.

## A capability gap this surfaced — the "pending approvals" list never had anything to call

Removed the "Messages awaiting approval" list on the Notifications page. It never had a real backend under it in *either* design: `ApprovalStore` itself has no query to list pending approvals (confirmed in both `#1368`'s `ApprovalResource` KDoc and my own earlier research before `#1369` was written) — showing a list page against a capability that doesn't exist would just always render empty or error.

Replaced with a **decide-by-id** form: the maker gives a checker the `approvalId` out of band (chat, verbally — there is genuinely no in-app queue to browse in this design), the checker pastes it in and approves/rejects. This is what the actual backend supports, not a scaled-down version of what `#1370` assumed existed.

## Test plan

- [x] `npm run type-check` — clean.
- [x] `npm run lint` — 0 errors, same 72 pre-existing warnings as the rest of the codebase (no new ones introduced — the rewritten `DecideApproval` component has no `useEffect` at all, since there's no list to load).
- [x] Full `vitest run` — 593/593 passing.
- [x] Cross-checked every request/response shape directly against the merged backend source on `origin/main` (`OperatorMessageResource.kt`, `ApprovalResource.kt`, `OperatorMessageService.kt`, `OperatorMessageTemplate.kt`) — not against `#1368`'s PR diff or its own description, in case of drift between what was proposed and what actually merged.
- [x] Dev server boots clean (Turbopack, `Ready in 353ms`), no console errors beyond the pre-existing dev-mode CSP `eval()` warning, route protection still redirects unauthenticated requests correctly.
- [ ] Full interactive E2E (compose → 202 → a different operator decides by id → retry → sent) needs a live Keycloak + the deployed backend; not achievable in this sandbox — noted rather than silently skipped.

Refs #1368, #1370, `docs/adr/0176-operator-initiated-customer-messaging.md`

### DIFF
--- a/openbank-admin-ui/src/app/notifications/page.tsx
+++ b/openbank-admin-ui/src/app/notifications/page.tsx
@@ -11,7 +11,7 @@ import { useAuth } from '@/lib/auth/useAuth'
 import { hasPermission } from '@/lib/auth/roles'
 import { classifyBffFailure } from '@/lib/services/bff'
 import { DataUnavailable, type UnavailableKind } from '@/components/feedback/DataUnavailable'
-import { opsMessageApi, type OperatorMessage } from '@/lib/api'
+import { opsMessageApi } from '@/lib/api'
 
 const NOTIFICATION_SERVICE = '/api/svc/notification-service'
 
@@ -99,7 +99,7 @@ export default function NotificationsPage() {
         ))}
       </div>
 
-      {canApprove && <PendingOperatorMessages />}
+      {canApprove && <DecideApproval />}
 
       {unavailable && (
         <div className="card" style={{ padding: 0, marginBottom: '16px' }}>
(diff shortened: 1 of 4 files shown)


JiRaska:
Superseded by #1535, which merged the same fix (same core files: api.ts, notifications/page.tsx, parties/[id]/page.tsx, plus a contract test this PR didn't have) while this one was in flight. #1535 is already on main (release admin-ui 0.50.1, #1543). Closing to avoid the now-conflicting duplicate.

PR #1667, opened 2026-07-18
fix(admin-ui): make compliance-controls live-signal join symmetric

JiRaska:
## Summary
- **Superseded in part by #1666** (merged while this PR was open): it already fixed the false-`enforced` data for `encryption-in-transit`/`strong-auth`/`authz-least-privilege` and removed the same-class bad `derivedFrom` on `net-segmentation`, all by removing the `derivedFrom` keys entirely. This PR is now rebased on top of that and reduced to what #1666 didn't touch: the join logic itself.
- `compliance.ts`'s live-signal join was asymmetric — `status: signal ? 'enforced' : c.status` — a `false` signal fell back to the *authored* status instead of demoting it, so it could only ever confirm an authored `enforced`, never disprove one. Harmless today (no remaining control has a `derivedFrom` whose authored status is `enforced`), but the next control that adds one would inherit the same one-way ratchet.
- Fix: make the join symmetric — a `false` signal now demotes an over-claimed `enforced` to `planned`; an already-honest `partial`/`audit`/`planned` status is left untouched.

## Linked issues
Refs #1666 (the data fix this PR builds on). Refs #1673 — filed by me as a `net-segmentation` follow-up before discovering #1666 already fixed it; closing #1673 as a duplicate.

## Test plan
- [x] Standalone simulation of the join logic against: (a) an authored `enforced` + false signal → demotes to `planned`, (b) an honest `partial` + false signal → stays `partial`, (c) `audit` + true signal → promotes to `enforced`, (d) no `derivedFrom` → untouched. All four passed.
- [x] `git diff origin/main -- openbank-libs/governance/compliance-controls.yaml` is empty — no data changes remain in this PR, only the `compliance.ts` join fix.
- [ ] Full `npm run type-check` / dev-server render blocked in this sandbox (no `node_modules` in a fresh worktree, and `.claude/launch.json`'s dev-server env points outside this branch) — reviewer should confirm `/docs/control-tower` still renders correctly.

### DIFF
--- a/openbank-admin-ui/src/lib/governance/compliance.ts
+++ b/openbank-admin-ui/src/lib/governance/compliance.ts
@@ -90,12 +90,16 @@ export async function loadComplianceCatalog(): Promise<ComplianceCatalog | null>
   const authored = parsed.controls ?? []
 
   // Join live status from the derived security posture where declared.
+  // Symmetric: a true signal promotes to `enforced`; a false signal demotes
+  // an authored `enforced` back down (the live signal disproved the claim),
+  // but leaves an already-honest partial/audit/planned status untouched.
   const posture = await loadSecurityPosture()
   const controls: ComplianceControl[] = authored.map(c => {
     if (!c.derivedFrom) return c
     const signal = signalValue(posture, c.derivedFrom)
     if (signal === undefined) return c
-    return { ...c, status: signal ? 'enforced' : c.status, live: true }
+    if (signal) return { ...c, status: 'enforced', live: true }
+    return { ...c, status: c.status === 'enforced' ? 'planned' : c.status, live: true }
   })
 
   // Per-framework coverage roll-up.


JiRaska:
Heads up — this overlaps with #1666, which already merged to `main` and touches the same 3 controls in `compliance-controls.yaml` (`encryption-in-transit`, `strong-auth`, `authz-least-privilege`) plus removed the stale `derivedFrom` on `net-segmentation` too. That's why this branch now shows as conflicting.

Independently arrived at the same root cause (`k8s/base/istio.yaml` unwired, `generate-security-graph.mjs` re-parsing the same dead file) via a parallel session, so the two diverged on approach:

- #1666: fixed only the authored data (`evidence`/`evidenceSource`/`status`), removed the four `derivedFrom: istio.*`/`network.defaultDeny` keys entirely rather than fix the generator — the join logic itself is untouched, so it's now dormant but still asymmetric.
- #1667 (this PR): fixes the same 3 controls' evidence, but *also* makes the `compliance.ts` join symmetric so a false live signal can demote an over-claimed status, not just confirm one — that's a real fix #1666 doesn't have.

Suggest rebasing onto `main` and dropping the now-redundant YAML hunks that duplicate what #1666 already did, keeping the `compliance.ts` symmetric-join fix (and re-adding a `derivedFrom` back onto the relevant controls if that fix depends on one being present to test against — worth checking, since #1666 removed them). Your evidence wording differs slightly in places (e.g. Kafka mTLS via Strimzi vs. ingress-nginx TLS for encryption-in-transit) — worth reconciling which is the more accurate citation before merging.

Not closing this — leaving the reconciliation call to whoever picks it back up.

JiRaska:
Follow-up filed: Refs #1673 (net-segmentation control's live signal reads the wrong file — same root cause, lower severity, intentionally out of scope here).

JiRaska:
Reopening to re-trigger issue-hygiene against the updated PR body.

Issue #1782, opened 2026-07-19
Duplicate party identities have no remediation path


JiRaska:
## Problem

ADR-0072 establishes "1 person = 1 party" and defends it **at creation time**: an RČ blind index (`POST /api/v1/parties/resolve`) and a four-eyes case queue in pid-service whose `LINK_TO_EXISTING` verdict routes an applicant onto an existing party.

Both are pre-creation gates. Neither can act once two parties for the same natural person are already live — and they do become live: onboarding through a channel carrying no RČ, a channel predating the blind index, or an operator creating a party by hand all bypass the dedup key.

## Why it bites

The only writer of `PartyStatus.CLOSED` is GDPR Art. 17 erasure (`DELETE /api/v1/parties/{id}`), which anonymises PII and emits `PARTY_ERASED`. So the only way an operator can retire a duplicate today is to record it as *the data subject exercised their right to be forgotten* — a false compliance statement, destructive of the history a merge is meant to preserve, and acted on by every consumer of that event.

The no-code-change workaround is worse: open matching pockets on the surviving account, move the balance as an ordinary customer payment, close the source account, erase the source party. The ledger then attests that one legal person paid another and was subsequently erased. A duplicate-identity data-quality finding is routine; a money movement that misrepresents what happened is an audit finding.

## Scope

- [x] ADR-0179 — decision record
- [x] `PartyStatus.MERGED` + `merged_into` column + migration
- [x] `POST /api/v1/parties/{id}/merge` with fail-closed preconditions, audited
- [x] `PARTY_MERGED` event, distinct from `PARTY_ERASED`
- [x] Account-ownership guard (fail-closed) so a funded duplicate cannot be retired
- [ ] Ledger `ADJUSTMENT` journal path for the balance sweep, referencing the merge approval (money-path, separate PR)
- [ ] Consumer adoption of `merged_into` per service holding a `partyId` — until then a stale id still resolves to the retired row
- [ ] Four-eyes approval on the merge endpoint itself (currently OPERATOR/ADMIN + audit)
- [ ] Admin-UI action (no party merge or close UI exists today)

## Notes

Not to be confused with `linkKeycloakSub` in customer-edge, whose comment calls itself "the ADR-0072 §5 identity merge" — it only attaches a `KEYCLOAK_ID` external id to an existing party, moves no accounts and retires nothing. That comment is misleading and worth correcting.

JiRaska:
[...]

PR #1783, opened 2026-07-19
feat(party): add duplicate party identity merge (ADR-0179)

JiRaska:
Refs #1782
Refs ADR-0179

## What

Adds an explicit merge operation for duplicate party identities, distinct from GDPR erasure.

## Why

ADR-0072 defends "1 person = 1 party" at *creation* time only — an RČ blind index and a four-eyes case queue that both run before a party exists. Once two parties for one person are live, there is no remediation path.

That gap has a sharp edge. The only writer of `PartyStatus.CLOSED` is Art. 17 erasure, which anonymises PII and emits `PARTY_ERASED`. Retiring a duplicate therefore meant recording a false subject-rights request and destroying the very history the merge exists to preserve. The alternative available without code — move the balance as a customer payment, then erase the source — leaves the ledger attesting that one legal person paid another and was then erased.

## Changes

- `PartyStatus.MERGED` + nullable `merged_into` (V15). Terminal exactly as `CLOSED` is, so a late KYC/AML callback cannot resurrect a retired duplicate. DB enforces `(status = 'MERGED') = (merged_into IS NOT NULL)`; FK is `ON DELETE RESTRICT`.
- `POST /api/v1/parties/{id}/merge` — OPERATOR/ADMIN, authz `party.merge`, audited. Refuses self-merge, an already-merged or erased party on either side, and a merged target (no chains).
- `PARTY_MERGED` event, deliberately not `PARTY_ERASED`. Nothing is anonymised.
- Account guard refusing the merge while the duplicate still owns an open account. **Fail-closed**: an unreachable account-service aborts. Account closure does not verify the balance (ADR-0109 option B), so a wrongly-permitted merge would strand funds on a retired identity with nothing downstream to notice.
- First M2M call from this service, so it brings the fleet-standard `quarkus.oidc-client` block. Token resolves to `ROLE_OPERATOR` — nothing in either Keycloak realm grants `ROLE_SERVICE`.

## Not in this PR

- The balance sweep as a ledger `ADJUSTMENT` journal entry referencing the merge approval (money-path, needs 2 approvals + threat model).
- Consumer adoption of `merged_into`. Until each service holding a `partyId` follows it, a stale id still resolves to the retired row — called out as a negative consequence in the ADR.
- Four-eyes on the merge endpoint itself, and an admin-UI action.

## Verification

- `detekt`, `ktlintCheck`, `test`, `quarkusBuild` all green.
- 9 new unit tests. Guards were mutation-checked: removing the `MERGED` terminal branch and the account guard makes exactly 3 of them fail, then pass again on restore — they are not vacuous.
- API contract 1.6.0 → 1.7.0 (additive). Checked against current `origin/main` and open PRs; no competing claim on the version.

## Reviewer notes

Two unrelated defects were found while researching this and are **not** fixed here:

1. `GdprAggregationAdapter` calls kyc-service and card-issuance-service with **no Authorization header**; both targets are role-protected. The 401 is swallowed as "no data", so Art. 15 subject-access exports silently ship without KYC and card PII.
2. `PartyService` KDoc cites ADR-0073 for the two-key activation gate, but ADR-0073 is about hardware-backed credential storage. Several ADR citations in this service look stale.

### DIFF
--- a/docs/adr/0179-party-identity-merge.md
+++ b/docs/adr/0179-party-identity-merge.md
@@ -1,0 +1,119 @@
+# ADR-0179 — Duplicate party identity merge
+
+Date: 2026-07-19
+Decision-Status: Proposed   <!-- Proposed | Accepted | Superseded by ADR-NNNN | Deprecated | Rejected -->
+Delivery-Status: Planned    <!-- Planned | Partial | Shipped | N/A — decision-only -->
+Author(s): Jiri Raska
+
+## Context
+
+ADR-0072 established "1 person = 1 party" and defends it at *creation* time: a RČ blind index
+(`POST /api/v1/parties/resolve`) plus a four-eyes case queue in pid-service whose verdict
+`LINK_TO_EXISTING` routes an applicant onto the party that already exists.
+
+Both mechanisms are pre-creation gates. Neither can act once two parties for the same natural
+person are already live — and they do become live: a customer onboarding through a channel that
+carries no RČ (or with a differing one), a channel that predates the blind index, or an operator
+creating a party by hand all bypass the dedup key. The platform has no remediation path for that
+state today.
+
+The absence has a sharp edge. The only writer of `PartyStatus.CLOSED` is GDPR Art. 17 erasure
+(`DELETE /api/v1/parties/{id}`), which anonymises PII and emits `PARTY_ERASED`. So the only way
+an operator can currently "retire" a duplicate party is to record it as *the data subject
+exercised their right to be forgotten* — a false compliance statement, destructive of the very
+history the merge is meant to preserve, and acted on by every downstream consumer of that event.
+
+The workaround available without any code change is worse than the gap: open matching currency
+pockets on the surviving account, move the balance across as an ordinary customer payment, close
+the source account, then erase the source party. The ledger then attests that one legal person
+paid another and was subsequently erased. A duplicate-identity data-quality finding is routine;
+a money movement that misrepresents what happened is an audit finding.
+
+## Decision
+
+We will add an explicit, first-class merge operation to party-service, distinct from erasure.
+
+1. **`PartyStatus.MERGED`** joins the status enum, and a nullable `merged_into` (UUID, FK to
+   `parties.id`) column records the surviving party. The pair makes the relation traversable in
+   both directions and distinguishes "retired as a duplicate" from "erased on request".
+   `deriveStatus` treats `MERGED` as terminal exactly as it already treats `CLOSED` — a merged
+   party is never re-opened by a later KYC or AML signal.
+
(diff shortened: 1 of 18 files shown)


JiRaska:
Self-review found two silent defects, both now fixed in d28685dc. Recording them here because both are the same shape — a safety mechanism that fails without saying anything.

**1. The fail-closed guard was fail-open.** `AccountPageBody` named the page fields `items`/`pageInfo` and the account identifier `iban`. account-service actually serializes `CursorPage` as `data`/`pagination`, and the field is `accountNumber`. Jackson does not throw on unmatched names — it leaves the list empty. So `findOpenAccounts` returned no open accounts for every party, and the guard permitted exactly the merge it exists to refuse.

Worth dwelling on: the unit tests all passed, because they mock `PartyAccountGuardPort`. The mock proved the *service* reacts correctly to a non-empty list; nothing proved the *adapter* could ever produce one. `AccountServiceClientContractTest` now builds its payload from the real `CursorPage` type rather than hand-written JSON, so a rename in openbank-libs fails the test instead of quietly disarming the guard. Mutation-checked: renaming `data` back to `items` fails 2 of 3.

**2. A merged party could never be erased.** The V15 CHECK was a biconditional, `(status = 'MERGED') = (merged_into IS NOT NULL)`. GDPR Art. 17 erasure flips `status` to `CLOSED` in place (`PartyRepositoryImpl.anonymize`) without touching `merged_into`, so the UPDATE would have aborted — an Art. 17 request against a retired duplicate would fail outright. Relaxed to the two directions actually wanted: MERGED implies a pointer; a pointer may only appear on MERGED or CLOSED. The pointer is lineage, not personal data, so it survives erasure while the row's PII is anonymized as usual.

**ADR-0179 corrected on a point of substance.** It claimed the balance sweep would be "an ADJUSTMENT journal entry, already four-eyes gated". Neither half was true:

- `PaymentJournalFactory.buildLines` never reads `transaction.type`. DEBIT, TRANSFER and ADJUSTMENT produce byte-identical journals — the type is a label on the transaction row and nothing more, so posting the sweep "as an ADJUSTMENT" buys zero audit distinction from an ordinary customer payment.
- Neither `ledger.create` nor `transaction.create` is in the OPA four-eyes verb list (it covers `post` and `reverse`, which no posting endpoint emits), and `authz.four-eyes.enforce` defaults to `false` in ledger-service anyway.

The distinction the ADR promised has to be built. The mechanical shape is right — the two-deposit-control-leg internal transfer nets the control account to zero and moves only the sub-ledger dimension — but the type carried through to the journal, the reason code, and the approval binding are all missing. The ADR now states this and sketches a dedicated operator-only action rather than reusing `transaction.create`, which is on the M2M payment rails.

Until that lands, the account guard is what prevents a half-finished merge: a duplicate holding an open account cannot be retired at all.

PR #1785, opened 2026-07-19
fix(party): authenticate the GDPR Art. 15 aggregation hops

### LINKED ISSUE #1784: GDPR Art. 15 export silently omits KYC and card PII (unauthenticated aggregation hops)
GdprAggregationAdapter builds raw java.net.http requests with **no Authorization header**:

- GET `{kyc-service}/api/v1/kyc/cases/party/{partyId}` — `KycResource` is `@RolesAllowed(VIEWER, OPERATOR, ADMIN, KYC, KYC_OPENER, KYC_REVIEWER, COMPLIANCE, SERVICE)`
- GET `{card-service}/api/v1/cards/party/{partyId}` — `CardResource` is `@RolesAllowed("ROLE_VIEWER","ROLE_OPERATOR","ROLE_ADMIN")` + `@Authorize(action = "card.list")`

So in any deployed environment both hops return 401. The adapter treats every non-200 as "no data" (null / emptyList + a `log.warnf`), which is indistinguishable from "this party genuinely has no KYC case and no cards". Net effect: **every GDPR Art. 15 subject-access export ships without the KYC and card PII and still reads as successful.**

The class KDoc justified avoiding the MicroProfile REST client "to avoid the ClientHeadersFactory classpath issue (issue #247)". #247 is an unrelated (and merged) admin-ui dossier-status change — that constraint never applied here; auth appears to have been dropped as collateral of the workaround.

### Second, independent blocker

`gen-network-policies.py` derives NetworkPolicy edges from the `http://<svc>.<ns>.svc:<port>` URLs a Deployment declares in env. `party-service.yaml` declares neither downstream URL (they lived only in `application.yaml`), so `party` is absent from both the `kyc-service` and `card-issuance-service` ingress allow-lists. With the VPC CNI agent in standard mode (ADR-0060) the calls are DROPPED — fixing the bearer token alone would not have made them work.

Also latent: the `application.yaml` default pointed at `card-issuance-service.card-issuance:8118`, but card-issuance-service runs in the `payments` namespace.

### Why it was invisible

`GdprAggregationAdapterTest` covered only happy path and generic failure against a loopback stub. It had no assertion that could distinguish 401 from an empty result — the exact failure in play.

### Fix

Follow the fleet M2M pattern (`@RegisterRestClient` + `@RegisterProvider(OidcClientRequestReactiveFilter::class)`), add the `quarkus.oidc-client` block, declare the downstream URLs in gitops, regenerate the network policies, and test 401 separately from absence.

### Compliance question for the PR

Should a 401/403 stay best-effort at all? An auth failure is not absent data; for Art. 15 a silently incomplete export is arguably worse than a loud one.

JiRaska:
Closes #1784

## What was wrong

`GdprAggregationAdapter` built raw `java.net.http` GETs with **no Authorization header** to two role-protected endpoints:

| hop | guard |
|---|---|
| `GET {kyc}/api/v1/kyc/cases/party/{id}` | `@RolesAllowed(VIEWER, OPERATOR, ADMIN, KYC, KYC_OPENER, KYC_REVIEWER, COMPLIANCE, SERVICE)` |
| `GET {card}/api/v1/cards/party/{id}` | `@RolesAllowed("ROLE_VIEWER","ROLE_OPERATOR","ROLE_ADMIN")` + `@Authorize(action = "card.list")` |

Both answered 401 in every deployed environment. The adapter treated any non-200 as "no data" (`null` / `emptyList` + a `log.warnf`), which is indistinguishable from a party genuinely holding no KYC case and no cards. **Every GDPR Art. 15 subject-access export shipped without the KYC and card PII and still read as successful.**

The class KDoc justified the raw client "to avoid the ClientHeadersFactory classpath issue (issue #247)". I checked: #247 is an unrelated, merged admin-ui dossier-status change. That constraint never applied here — auth looks like collateral of a workaround for a problem that was never this one.

## Second, independent blocker (not in the original report)

`gen-network-policies.py` derives NetworkPolicy edges from the `http://<svc>.<ns>.svc:<port>` URLs a Deployment declares in **env**. Both downstream URLs lived only in `application.yaml`, so `party` was absent from the `kyc-service` and `card-issuance-service` ingress allow-lists. With the VPC CNI agent in standard mode (ADR-0060) those calls are DROPPED — **fixing the bearer alone would not have made this work.** Declared both in `party-service.yaml` and regenerated.

Two traps worth recording:
- The generator matches only URLs carrying the literal `.svc` suffix. My first pass wrote `http://kyc-service.kyc:8114`, the regen ran clean and changed nothing — a silent no-op that looks exactly like "already correct". Caught by diffing the regenerated allow-lists rather than trusting the exit code.
- The pre-existing `application.yaml` default named `card-issuance-service.card-issuance:8118`; card-issuance-service actually runs in `payments`.

## Compliance judgement — please weigh in

I did **not** keep 401/403 best-effort. Being *refused* data is not the same as the data being *absent*, and an Art. 15 response that is silently incomplete is indistinguishable from a complete one — the DPO files it and the data subject never learns anything was withheld. So:

- **401/403** → `GdprAggregationAuthException` → **502**, export does not ship.
- **404** (no case / no cards) → degrade, export ships.
- **5xx / unreachable / timeout** → degrade, export ships. A downstream outage must not block a data subject request; the DPO follows up from the log.

If you would rather have a 200 with an explicit `incomplete: true` annotation instead of a 502, say so — it is a one-line change in the mapper, but it puts the burden on every consumer to notice the flag, which is what failed here in the first place.

## API contract

`openapi.yaml` `info.version` 1.6.0 → 1.7.0 (additive `502` response; the `gdpr-export` description was also stale — it said aggregation was "a tracked follow-up" when it had shipped).

> ⚠️ **Version collision:** open PR #1783 also claims `1.7.0` on this same spec. Whichever of us lands second must re-bump to 1.8.0 — a clean text merge here is the trap, since git sees identical lines, not a taken version. #1783 also touches `PartyPorts.kt`, `ExceptionMappers.kt`, `build.gradle.kts` and `application.yaml`, so expect conflicts either way.

## Tests

The old suite drove a loopback `HttpServer` and asserted only "200 → data" and "error → degrade" — it had no assertion that could tell 401 from an empty result, which is precisely why this was invisible. Rewritten against mocked REST clients with 401 and 403 asserted separately from 404, 5xx, and unreachable, on both hops.

## Verification

`detekt`, `ktlintCheck`, `test`, `quarkusBuild` (CDI/ArC wiring — the only step that catches the new `@RestClient` injection) all green locally on JDK 25.

Not verified end-to-end against the live sandbox: that needs the deploy of both this image and the regenerated NetworkPolicies. Worth a real Art. 15 export against a party that *does* hold a KYC case and a card before this is called done — the whole failure class here is "looks fine from every angle except the one nobody checked".

### DIFF
--- a/openbank-infra/gitops/components/kyc/network-policies.yaml
+++ b/openbank-infra/gitops/components/kyc/network-policies.yaml
@@ -41,6 +41,9 @@ spec:
               kubernetes.io/metadata.name: customer-edge
         - namespaceSelector:
             matchLabels:
+              kubernetes.io/metadata.name: party
+        - namespaceSelector:
+            matchLabels:
               kubernetes.io/metadata.name: security-scanner
         - namespaceSelector:
             matchLabels:
--- a/openbank-infra/gitops/components/party/party-service.yaml
+++ b/openbank-infra/gitops/components/party/party-service.yaml
@@ -67,6 +67,15 @@ spec:
                 secretKeyRef:
                   name: party-service-oidc
                   key: OIDC_CLIENT_SECRET
+            # GDPR Art. 15 aggregation hops (ADR-0118 §6). Declared here — not only in
+            # application.yaml — because gen-network-policies.py derives the NetworkPolicy
+            # edges from the env URLs a Deployment declares: without these two entries the
+            # party → kyc / party → card-issuance calls are DROPPED by the VPC CNI agent
+            # even when the M2M bearer is correct.
+            - name: OPENBANK_KYC_SERVICE_URL
+              value: http://kyc-service.kyc.svc:8114
+            - name: OPENBANK_CARD_SERVICE_URL
+              value: http://card-issuance-service.payments.svc:8118
             - name: QUARKUS_OTEL_EXPORTER_OTLP_ENDPOINT
               value: "http://tempo.observability.svc:4317"
             - name: OTEL_TRACES_SAMPLER_ARG
--- a/openbank-infra/gitops/components/payments/network-policies.yaml
+++ b/openbank-infra/gitops/components/payments/network-policies.yaml
@@ -41,6 +41,9 @@ spec:
               kubernetes.io/metadata.name: customer-edge
         - namespaceSelector:
             matchLabels:
+              kubernetes.io/metadata.name: party
+        - namespaceSelector:
+            matchLabels:
               kubernetes.io/metadata.name: security-scanner
         - namespaceSelector:
             matchLabels:
--- a/openbank-party-service/build.gradle.kts
+++ b/openbank-party-service/build.gradle.kts
@@ -21,6 +21,11 @@ dependencies {
     implementation(libs.quarkus.micrometer.registry.prometheus)
     implementation(libs.quarkus.opentelemetry)
     implementation(libs.quarkus.oidc)
+    // The oidc-client filter mints the openbank-services M2M token the GDPR Art. 15 aggregation
+    // hops to kyc-service / card-issuance-service require — both endpoints are @RolesAllowed.
+    implementation(libs.quarkus.rest.client.reactive)
+    implementation(libs.quarkus.rest.client.reactive.jackson)
+    implementation(libs.quarkus.oidc.client.reactive.filter)
     implementation(libs.quarkus.config.yaml)
     implementation(libs.quarkus.smallrye.openapi)
     implementation(libs.quarkus.smallrye.fault.tolerance)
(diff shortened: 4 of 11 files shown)


JiRaska:
Superseded by #1883 — rebased onto current main after this PR's branch (`build.gradle.kts` + `application.yaml`) conflicted with ADR-0179's concurrent M2M client wiring for account-service. Merged both callers into one shared `oidc-client` block. Also fixed two required-check failures that surfaced during rebase: the ADR-0106 identifier-intent guard (GdprAggregationAuthMapper's error id now mints via `Ids.randomId()`) and the ADR-0048 API-contract gate (bumped info.version 1.7.0 -> 1.8.0 for the new 502 response). Now merged as `c0a54fd6e`.

Issue #1797, opened 2026-07-19
Eight live services enforce @Authorize with no OPA PDP — every @Authorize endpoint fails closed


JiRaska:
## Summary

Eight **live** services enforce `@Authorize` while having **no OPA policy decision point** to ask.
`AuthorizeInterceptor` fails closed on an unreachable PDP, so **every `@Authorize`-annotated
endpoint on these services is bricked** — it can never return anything but an error, for any caller.

This violates the invariant `AuthorizeInterceptor` itself documents
(`openbank-libs-runtime/.../authz/AuthorizeInterceptor.kt:55`):

> Advisory must never brick an endpoint when no sidecar is deployed yet.

## Mechanism

Two independent defaults line up badly:

- `AuthorizeInterceptor` declares `@ConfigProperty(name = "authz.enforce", defaultValue = "true")`.
- A service's `application.yaml` either sets `authz.enforce: "${AUTHZ_ENFORCE:true}"` or omits the
  block entirely — both resolve to **enforce**.
- The service's gitops manifest declares **no `opa` sidecar** and never sets `AUTHZ_ENFORCE`.

Result: enforce=true against a PDP at `localhost:8181` that does not exist.

Confirmed live in `openbank-sandbox` (`AUTHZ_ENFORCE` unset in the pod spec, no `opa` container):

| service | namespace | `@Authorize` files | `authz.enforce` default |
|---|---|---|---|
| party-service | party | 10 | `${AUTHZ_ENFORCE:true}` |
| kyc-service | kyc | 9 | `${AUTHZ_ENFORCE:true}` |
| aml-service | aml | 8 | `${AUTHZ_ENFORCE:true}` |
| statement-service | statements | 2 | `${AUTHZ_ENFORCE:true}` |
| tpp-registry-service | tpp-registry | 11 | `${AUTHZ_ENFORCE:true}` |
| standing-order-service | standing-orders | 11 | absent → libs default `true` |
| audit-service | audit | 1 | absent → libs default `true` |
| onboarding-service | onboarding | 6 | absent → libs default `true` |

`standing-order-service` is a distinct sub-case: its gitops manifest **does** declare the sidecar,
but the running pod has no `opa` container — manifest-vs-live drift, not a missing manifest.

Same config shape, currently not deployed (fix before they are): `dispute-service`, `psd2-service`,
`sdd-service`.

[...]

Issue #1322, opened 2026-07-16
OPA service REST extensions have zero test coverage — 22 generators embed rego as bash heredocs


JiRaska:
Found while flipping `AUTHZ_ENFORCE` for the last four services (#266).

## The gap

`opa test` covers `openbank-libs/governance/policies/rest.rego` and `openbank-infra/opa/policies/agents.rego`. It covers **none** of the per-service REST extensions, because those exist only as bash heredocs inside `gen-*-opa-bundle.sh`, never as `.rego` files on disk:

```
22 of 27 generators:  CARD_ISSUANCE_REST_EXT=$(cat << 'REGO' ... REGO)
```

There is nothing for `opa test` to load, so ~22 service-specific policy extensions ship unverified.

## Why it matters — a real bug this hid

`gen-card-issuance-opa-bundle.sh` granted `card.block` to `ROLE_COMPLIANCE` only, while `CardResource.block()` carries `@RolesAllowed("ROLE_OPERATOR", "ROLE_ADMIN", "ROLE_COMPLIANCE")`. `@RolesAllowed` is a **disjunction**; the extension's comment read it as a conjunction. The policy was therefore strictly narrower than the resource it guards, and flipping `AUTHZ_ENFORCE=true` would have 403'd every card block by an operator — the fraud response for a lost/stolen card. Fixed in the PR that closes this issue, but only because the action list was enumerated by hand.

**The evidence-from-traffic path would have missed it.** card-issuance-service is KEDA scale-to-zero with 2 calls in 7 days; shadow-mode denial counts said "low risk, flip it". Only a static enumeration of `@Authorize` actions against the live bundle found it.

## The bug class

Any per-service extension can silently diverge from the resource it guards, in either direction:

- **narrower than `@RolesAllowed`** — the action 403s under enforce (the card.block case; fails closed, breaks a control)
- **wider than `@RolesAllowed`** — the policy grants what the resource would refuse (fails open; Jakarta still catches it, but the policy is then lying about intent)

Neither is caught today by any gate.

## Suggested work

1. **A CI guard** (`rules.yaml: authz_policy`): for each `@Authorize(action = "x")` in a resource, assert the set of principals the rego admits for `x` matches the set `@RolesAllowed` on that method admits. Both sides are declarative, so this is mechanical — and it is the check that would have caught card.block. This is the high-value item.
[...]

PR #1798, opened 2026-07-19
security(party): wire the OPA PDP sidecar so @Authorize stops failing closed

JiRaska:
## What

Gives party-service a working OPA policy decision point, so its `@Authorize` endpoints stop
failing closed.

`openbank-party-service/src/main/resources/application.yaml` sets
`authz.enforce: "${AUTHZ_ENFORCE:true}"`, but `openbank-infra/gitops/components/party/party-service.yaml`
declared no OPA sidecar and never set `AUTHZ_ENFORCE`. `AuthorizeInterceptor` fails closed when the
PDP is unreachable, so every `@Authorize` action on this service returned:

```
HTTP 422 BUSINESS_RULE_VIOLATION
"policy decision point unavailable: OPA call failed: null"
```

`party.update` and `party.consent.update` have been dead for as long as the sidecar has been
missing. It stayed invisible because neither is exercised in sandbox — `party.merge` (ADR-0179,
#1783) was simply the first one anyone called.

Mirrors account-service: OPA `1.17.0` sidecar on `:8181`, bundle mounted from a generated
`party-opa-bundle` ConfigMap, pod-roll checksum annotation kept in sync by the generator.

## Not fixed by disabling enforcement

Setting `AUTHZ_ENFORCE=false` would make the 422 go away, and would be the wrong fix. 22 of the 29
sidecar-equipped services run `AUTHZ_ENFORCE=true` with a working PDP — enforcement is the fleet
norm, and enforcing *without* a PDP is the anomaly. The gap is the missing PDP, so that is what
this wires.

`AUTHZ_ENFORCE=true` is stated explicitly in the manifest. That is **not** a behaviour change — the
pod was already enforcing via the `application.yaml` default. It makes the posture visible next to
the sidecar that now justifies it, the way account-service does.

## Policy

party-service has **four** `@Authorize` actions. Each grant has an in-repo caller; each non-grant
has a stated reason.

| action | granted to | evidence |
|---|---|---|
| `party.merge` | human `ROLE_OPERATOR`/`ROLE_ADMIN` | operator console; no in-repo caller yet |
| `party.update` | human `ROLE_OPERATOR`/`ROLE_ADMIN` | no in-repo caller |
| `party.consent.update` | the above **plus** `service-account-openbank-edge` | `CustomerEdgeResource.kt:1087` |
| `party:resolve` | **nobody** | dead endpoint — see below |

**The `service-account-` exclusion on `operator-party-write` is load-bearing, not boilerplate.**
`AuthorizeInterceptor` never emits `principal.type == "SERVICE"`; a Keycloak client_credentials
token is classified `HUMAN`, and the customer-edge M2M client carries `ROLE_OPERATOR` in the realm.
The obvious one-line rule ("operators may write `party.*`") would therefore hand the **public
customer edge** the ability to merge and retire arbitrary party identities. The exclusion is the
only thing that prevents it, and `test_party_merge_denied_for_edge_m2m` is the assertion that fails
if anyone removes it.

**`party.consent.update` had to be granted to the edge, or this PR would have made things worse.**
`CustomerEdgeResource.updateConsent` (the mobile app's Profile-screen marketing-consent toggle)
calls party-service under a client_credentials token. Wiring the PDP without this grant would have
converted today's 422 into a 403 — the same broken toggle with a new status code. It is scoped to
that one action, not a `party.` family prefix.

**`party:resolve` is deliberately left to default-deny.** Three independent reasons: it has no
in-repo caller (the live ADR-0072 dedup gate is *pid-service's* own `/parties/resolve`,
`CustomerEdgeResource.kt:3100`); its `@RolesAllowed` requires `ROLE_SERVICE`, which no realm client
holds, so JAX-RS rejects M2M callers before the interceptor even runs; and note the **colon** — a
future rule written as `startswith(input.action, "party.")` will silently not match it. Removing the
dead endpoint is an API-surface change (`openapi.yaml` + contract test), so it is not folded into a
security fix.

Four-eyes was considered and rejected for `party.merge`: party-service is not in
`money_path_services`, so the verb-based clause in `rest.rego` cannot fire for it, and the handler
does not consume `AuthzDecision.attributes` — adding it would be decorative. The endpoint already
takes an `approvalReference`.

## Verification

The extension lives on disk rather than in a bash heredoc (the pattern #1322 established, first
adopted by `gen-card-issuance-opa-bundle.sh`) specifically so `opa test` can cover it.

**Unit — `opa test`: 12/12 pass.**

**Mutation-tested, because a passing test proves nothing until it can fail.** Both guard rails were
verified by breaking them:

| mutation | result |
|---|---|
| delete the `not startswith(principal.id, "service-account-")` guard | 3 FAIL — `party_merge_denied_for_edge_m2m`, `party_merge_denied_for_other_m2m`, `party_update_denied_for_edge_m2m` |
| widen the edge grant from `== "party.consent.update"` to `startswith("party.")` | 2 FAIL — `party_merge_denied_for_edge_m2m`, `party_update_denied_for_edge_m2m` |

**End-to-end against the real generated bundle** — the `party-opa-bundle` ConfigMap materialized to
disk exactly as the sidecar mounts it, then queried with `opa eval` on `data.openbank.rest.allow`:

```
ALLOW  party.merge          / HUMAN ROLE_OPERATOR              reason=operator-party-write
ALLOW  party.merge          / HUMAN ROLE_ADMIN                 reason=operator-party-write
ALLOW  party.consent.update / service-account-openbank-edge    reason=service-edge-party-consent-m2m
ALLOW  party.update         / HUMAN ROLE_OPERATOR              reason=operator-party-write
DENY   party.merge          / service-account-openbank-edge     <- takeover guard
DENY   party.update         / service-account-openbank-edge
DENY   party:resolve        / HUMAN ROLE_ADMIN
DENY   party.merge          / HUMAN ROLE_VIEWER
DENY   party.merge          / ANONYMOUS
```

`opa check --strict` on the generated bundle passes; `openbank-infra/opa/build-bundle.sh` still
green.

## Generated artifacts

- All `gen-*opa-bundle*.sh` re-run — **no other service's bundle drifted**, since no shared policy
  source changed.
- `gen-network-policies.py` re-run; diff is exactly two lines on party (opens `8181`, matching
  account-service). Diffed rather than trusting the exit code, per the generator's own caveat.
- Generator committed `100755` — a non-executable generator no-ops the CI loop and looks identical
  to "in sync".

## Post-merge check

```
kubectl get pods -n party -l app.kubernetes.io/name=party-service \
  -o jsonpath='{.items[0].spec.containers[*].name}'   # expect: party-service flagd opa
```
Then `POST /api/v1/parties/{id}/merge` with a `ROLE_OPERATOR` token — expect a business response,
no OPA 422.

## Scope

**This is 1 of 8.** The same gap affects `kyc`, `aml`, `statement`, `tpp-registry`,
`standing-order`, `audit` and `onboarding` — all live, all enforcing with no PDP. Tracked in #1797
rather than fixed here: each needs its own caller audit before its actions can be granted safely,
and `rules.yaml`'s `four_eyes` guardrail explicitly requires per-caller evidence, so this is not a
mechanical sweep. #1797 also records two things this surfaced: no CI guard asserts
"enforces authz ⇒ has a PDP", and the per-service rego suites (this one and card-issuance) do not
actually run in CI.

Refs #1797


JiRaska:
Superseded by #1796 (merged 2026-07-20), which wired the party OPA PDP sidecar with the shared-source generator pattern (zero drift across the other 27 bundles) and advisory-first AUTHZ_ENFORCE=false per the ADR-0034 Phase 5 rollout order. The extra rego test coverage here is worth carrying over — tracked in the enforce-flip follow-up (#1797).

PR #1909, opened 2026-07-23
fix(notification): use UUIDv7 for consumer notification IDs

### LINKED ISSUE #1388: Stale comment in OperatorMessageService: Ids.newId() does not match NotificationConsumer's rows

JiRaska:
Re-signed carry of #1434 by @pollychen-lab (authorship preserved) — the original commit is unsigned and the `main-protection` ruleset requires signed commits, so it would strand with green checks and auto-merge armed.

## Summary (from #1434)
- switch NotificationConsumer durable notification IDs from bare UUIDv4 to `Ids.newId()` UUIDv7 (ADR-0106)
- update the OperatorMessageService comment now that both notification write paths use UUIDv7
- assert the persisted consumer notification ID is UUIDv7 in NotificationConsumerIT

Fixes #1388. Closes #1434.

### DIFF
--- a/openbank-notification-service/src/main/kotlin/com/openbank/notification/application/NotificationConsumer.kt
+++ b/openbank-notification-service/src/main/kotlin/com/openbank/notification/application/NotificationConsumer.kt
@@ -8,6 +8,7 @@ import com.fasterxml.jackson.databind.ObjectMapper
 import com.openbank.libs.audit.AuditEvent
 import com.openbank.libs.audit.AuditEventPublisher
 import com.openbank.libs.audit.AuditResult
+import com.openbank.libs.domain.identifiers.Ids
 import com.openbank.notification.application.port.out.OversightWebhookPublisher
 import com.openbank.notification.application.port.out.PushMessage
 import com.openbank.notification.application.port.out.PushSender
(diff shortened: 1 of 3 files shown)


pollychen-lab:
@JiRaska thanks

Issue #1265, opened 2026-07-16
interest: a mixed-currency accrual set wedges capitalize() forever — the KDoc promises an operator exit that does not exist


JiRaska:
Follow-up to #1246. The refusal it added is the right call; **the KDoc's claim that an operator can resolve it is false as shipped** — there is no such path, and the accruals wedge permanently.

## What #1246 shipped

`InterestService.capitalize` now refuses a pending set spanning several currencies (correctly — summing 100 CZK and 5 EUR into "105" is nonsense, and `WithholdingTaxPolicy` assesses per currency). It documents the exit:

> KDoc: *"an operator must split the product's accruals per currency"*
> error: *"the accruals stay ACCRUING; investigate why one product accrued in several currencies"*

Neither is actionable. There is no operator path.

## The state is reachable (not dead code)

Nothing binds a currency to a product, and nothing validates the pair:

- `InterestResource.accrue` takes `currency` straight from the request body (`AccrualRequest`).
- `InterestRateConfig` (`InterestModels.kt`) has **no currency field** — a product is currency-agnostic by construction.
- `interest_accruals` UNIQUE is `(account_id, accrual_date, product_id)` (`V1__init_interest.sql:33`) — **currency is not in the key**, so two dates under one `(account, product)` may legitimately carry different currencies.
- The column default is `currency CHAR(3) NOT NULL DEFAULT 'EUR'` (`V1:29`) while withholding assesses CZK — so a caller omitting the field silently accrues in EUR.

Repro:
```
POST /api/v1/interest/accrue  {productId:"SAVINGS", currency:"CZK", accrualDate:2026-01-18, ...}
POST /api/v1/interest/accrue  {productId:"SAVINGS", currency:"EUR", accrualDate:2026-01-19, ...}
POST /api/v1/interest/capitalize/{accountId}?productId=SAVINGS   -> 500, forever
```

## There is no way out

| candidate exit | status |
| --- | --- |
| an endpoint that mutates/reverses an accrual | **does not exist** — `InterestResource` has accrue / capitalize / reads / rate-config only |
| `AccrualStatus.REVERSED` | **zero writers** anywhere in `src/main` (verified fleet-wide) |
| `AccrualStatus.SUSPENDED` | **zero writers** anywhere in `src/main` |
| capitalize a subset | `findPendingCapitalization` reads **all** `ACCRUING` rows for the `(account, product)` — the poisoned row is re-read every attempt |

[...]

PR #2043, opened 2026-07-24
fix(interest): freeze tax profile at claim time so a capitalize() retry can't diverge row-vs-GL (#1355)

### LINKED ISSUE #1355: interest: capitalize() retry can silently diverge row-vs-GL if tax-profile resolution ever becomes non-constant
Dormant follow-on to #1316 (interest capitalization credit leg). Not exploitable today — filed so the fix described below doesn't silently reopen a closed money-path defect.

## The gap

`InterestService.capitalize`'s claim-state design freezes `gross` (the claimed accrual set is fixed before the ledger post — a backfilled accrual cannot join it after the fact). It does **not** freeze the withholding tax computation: `taxProfilePort.resolve(accountId)` is called fresh on every attempt inside `capitalizeSet`, including retries.

The ledger's idempotency key is amount-blind (same `interest-capitalization-{accountId}-{productId}-{periodTo}` key regardless of the request body) and a key hit replays the original journal **before any validation, without comparing amounts**. So:

1. Attempt 1: claim set → resolve profile P1 → tax/net computed → post journal J (amounts from P1) → **crash before `saveWithOutbox` commits**.
2. Something changes the account's tax profile (party update, treaty change, exemption granted/revoked) between attempt 1 and the retry.
3. Retry: same claimed set (gross unchanged) → resolve profile **P2** → tax/net recomputed from P2 → same idempotency key → ledger returns J (still P1's amounts, silently) → `saveWithOutbox` commits a capitalization/withholding row using **P2's** tax/net.

Result: the GL reflects P1's split: the interest-service row reflects P2's. Same divergence class as the finding round 1 fixed for the gross-amount case, reopened through the tax-profile axis instead.

## Why it's dormant, not live

`DefaultTaxProfileProvider.resolve()` (`openbank-interest-service/.../infrastructure/tax/DefaultTaxProfileProvider.kt`) unconditionally returns the constant `TaxProfile.FAIL_SAFE_DEFAULT`. Nothing can differ between attempt 1 and the retry today, so the race has no reachable trigger.

`TaxPorts.kt` documents real account→party tax-attribute resolution as a "documented fast-follow." **That fast-follow is what would make this reachable.**

## Fix direction (for whoever picks up the tax-profile fast-follow)

Snapshot the resolved profile (or the computed tax/net) at claim time, alongside the claimed accrual set, so a retry replays the same inputs rather than re-resolving them. The claim-state migration (V7, #1316) is the natural place to extend if a column is needed.

## Ask

When implementing real tax-profile resolution, re-open this and add a test that changes the profile between a crashed attempt and its retry — the equivalent of #1316's "an accrual backfilled after a crashed post cannot make the row and the GL disagree" test, but for the tax axis.

Refs #1316 (round-2 review comment)

JiRaska:
## Why

`capitalize()`'s claim freezes the accrual **set** (gross) before the ledger post, but it did **not**
freeze the withholding computation: `taxProfilePort.resolve(accountId)` was re-called fresh on every
attempt, including retries. The ledger idempotency key is amount-blind, so a retry after a crashed
post replays the **original** journal (P1's amounts) silently, while `capitalizeSet` recomputed
tax/net from a freshly-resolved profile — **P2** if the account's tax attributes changed in between.
The GL then reflects P1's split and the interest-service withholding row P2's — the same row-vs-GL
divergence class #1316 fixed for the gross axis, reopened through the **tax** axis (issue #1355).

Dormant today only because `DefaultTaxProfileProvider.resolve()` returns a constant
(`FAIL_SAFE_DEFAULT`). The account→party tax-attribute resolution fast-follow is what would make it
reachable — this hardens the money path **before** that lands, so the fast-follow can't silently
reopen a closed defect.

## What

Snapshot the resolved tax profile at **claim time** — the durable freeze point, committed before the
ledger post — alongside `claimed_period_to`, and replay it on retry instead of re-resolving:

- **V13**: five nullable snapshot columns on `interest_accruals`
  (`claimed_taxpayer_type`, `claimed_residency`, `claimed_treaty_rate`,
  `claimed_non_cooperating_state`, `claimed_exempt_code`).
- `InterestAccrual` gains `claimedTaxProfile: TaxProfile?`; `claimForCapitalization(ids, periodTo,
  profile)` stamps it; `capitalizeSet` reads it on the retry path via `claimProfile(...)` and only
  calls `TaxProfilePort.resolve` on a fresh set (or as a fallback for a claim already in flight at
  deploy that carries no snapshot — safe while resolution is constant).

Withholding recompute is deterministic given `(gross, profile, toDate)` — all three now frozen — so a
retry's row matches the journal the ledger idempotently replays.

## Tests

- **Unit** (`InterestServiceTest`): a retry with a claimed set carrying a frozen **legal-entity**
  profile commits `NOT_WITHHELD` / zero tax even though `resolve()` would return the 15 %-withheld
  individual default — and `resolve` is **never called** on the retry. Plus: the fresh path freezes
  the resolved profile into the claim.
- **Real-DB IT** (`CapitalizationLedgerBoundaryIT`): a non-default profile (non-resident, treaty rate,
  exempt code) persisted at claim round-trips byte-for-byte back through the V13 columns + mapper +
  claim SQL on `findClaimedForCapitalization`.

## Migration / merge order

- **V13** with a rollback note. ⚠️ **Depends on #1265's V12** (PR #2037) merging first — V13 assumes
  V11→V12→V13 order; landing V13 before V12 would make Flyway reject the later V12 as out-of-order on
  already-migrated DBs. #2037 is already required-green and ahead, so this is the natural order.
- No API/openapi change (internal correctness fix). Threat-model change-log updated (money-path, ADR-0030).

Closes #1355 · Refs #1316


JiRaska:
Ready for money-path review. Checks green (`mergeStateStatus: CLEAN`), no unresolved diffs. Per governance (ADR-0030) this needs **2 approvals + threat model** — `docs/threat-models/openbank-interest-service.md` is present on main. Not self-approving or `--admin`-merging; awaiting a second human reviewer.

Issue #2148, opened 2026-07-25
standing-order daily execution sweep throws HR000068 (reactive Panache off the Vert.x event loop) — no due order ever executes


JiRaska:
## Summary

`StandingOrderExecutionScheduler.sweep()` — the daily `@Scheduled` cron (`0 0 3 * * ?`) that turns due standing orders into payment-execution events — **throws on every run** and enqueues nothing. The standing-order execution pipeline is dead end-to-end, independent of the Kafka-mTLS (#2078) and OPA-sidecar (#2082) gaps just fixed. Confirmed live in sandbox via a full E2E.

## Mechanism

```kotlin
// StandingOrderExecutionScheduler.kt
@Scheduled(cron = "{openbank.scheduler.execution-cron}", ...)
fun sweep(): Unit = runBlocking {          // <-- runs on a Quarkus scheduler executor-thread
    ...
    val count = standingOrderUseCase.executeOrders(today)   // reactive Panache query
}
```

`executeOrders()` → `repo.findDueForExecution(asOf)` is the first Hibernate-Reactive call. `runBlocking` runs the coroutine on the scheduler's `executor-thread`, which carries **no Vert.x context**, so the reactive session blows up:

```
ERROR io.qua.sch.com.run.StatusEmitterInvoker  Error occurred while executing task for trigger
CronTrigger [id=1_...StandingOrderExecutionScheduler#sweep, cron=0 14 23 * * ?, ...]:
java.util.concurrent.CompletionException: java.lang.IllegalStateException:
HR000068: This method should exclusively be invoked from a Vert.x EventLoop thread;
currently running on thread 'executor-thread-1'
```

`findDueForExecution` sits before the per-order try/catch, so the whole sweep aborts — **0 outbox rows**, nothing published, no order ever executes. This is the documented repo footgun ("a reactive Panache repo can't be called from a bare thread — only an HTTP request carries a Vert.x context"), here on the scheduler (non-HTTP) path.

The sibling `StandingOrderOutboxDispatcher.dispatch()` in the SAME service is a **`suspend fun`** (no `runBlocking`) and works — Quarkus dispatches suspend `@Scheduled` methods on a proper context. Only `sweep()` uses `runBlocking`.

## Why it stayed hidden

[...]

Issue #2077, opened 2026-07-24
standing-order cancel/pause returns HTTP 500 (duplicate-key on save)


JiRaska:
## Symptom
`DELETE /customer/v1/standing-orders/{id}` (cancel) and the pause toggle return **HTTP 500**. Pod log (payments ns, `standing-order-service:sandbox-eb0bcf30`):

```
java.sql.SQLException: ERROR: duplicate key value violates unique constraint "standing_orders_pkey" (23505)
```

Confirmed live 2026-07-24 in sandbox while E2E-testing the new ONCE frequency. Pre-existing, unrelated to the ONCE change. Customer app's standing-order cancel (trash icon in `StandingOrderScreen`) and pause toggle are broken — every cancel/pause 500s.

## Root cause
`StandingOrderEntity` has an application-assigned `@Id` (no `@GeneratedValue`). `StandingOrderRepositoryImpl.save()` calls Panache `persist()`, which is **INSERT-only**. Create works; but `pause`/`resume`/`cancel`/`confirmExecution` load an existing row and re-`save()` it, so Hibernate schedules an INSERT for a PK that already exists and flushes `23505`. Same class as the ADR-0126 D3 assigned-`@Id` upsert rule (#1521).

## Fix
Switch `save()` to `merge()` (upsert). Regression IT `StandingOrderSaveUpsertIT` reproduces the exact duplicate-key error before the fix.

Issue #1521, opened 2026-07-17
Fleet sweep: 22 ADRs carry a false Delivery-Status (19 understate shipped work)


JiRaska:
## Summary

An audit of all 169 ADRs' `Delivery-Status` against the actual code on `origin/main` found **22 ADRs whose status is untruthful** — 19 understate what shipped (false-negative, the duplicate-work-inviting kind: #944→#950 was caused by exactly this), 3 overstate. Root cause tracked separately (docs-truth-agent Delivery-Status coverage).

First instance corrected in #1507 / #1509 (ADR-0138). This issue is the fleet tail.

**Method:** grep-based existence check of the artifact each ADR names (service, class, CI gate, migration, manifest). Findings are **hypotheses, not confirmations** — each ADR needs a per-ADR verify before its status is edited. Do **not** batch-edit blind. `ADR-0034` was initially flagged and **cleared on verify** (its "money-path pending" note is correct; a raw `AUTHZ_ENFORCE=true` env count is not live money-path enforcement) — treat every entry with that skepticism.

## Tier 1 — headline `Planned`/`Proposed` but shipped (highest confidence; seen shipping in git log)

- [ ] **ADR-0161** (Planned) — `ObjectStorePort` (libs-domain) + `S3ObjectStore`/`PostgresBlobStore` (libs-runtime) shipped
- [ ] **ADR-0162** (Planned) — full `openbank-document-service` shipped, released 0.8.4
- [ ] **ADR-0169** (Planned) — `CustomerDocumentResource` (D1), SCA `V6__document_signing_binding` (D2), language upsert (D3) shipped
- [ ] **ADR-0176** (Planned) — `OperatorMessageService`/`Resource`, `opsmessage.compose`, admin-ui `MessagesTab` shipped
- [ ] **ADR-0124** (Planned) — `scorecard.yml`, `security.yml` (CodeQL/Trivy), README badge exist; note says "not started"
- [ ] **ADR-0126** (Partial) — D1–D5 all coded; **the ADR body itself marks all five "Shipped"** while the header says Partial (internal contradiction)

## Tier 2 — header `Partial` technically holds, but the delivery NOTE claims a specific phase is pending that actually shipped (verify body wording, not header)

- [ ] **ADR-0019** — 38 services have `docs/` (526 md); note's Phase 4 "pending" is false
[...]

PR #2170, opened 2026-07-25
fix(standing-order): run the daily execution sweep on a Vert.x context (#2148)

JiRaska:
## Bug (money-path, severity:high)
`StandingOrderExecutionScheduler.sweep()` — the daily cron that turns due standing orders into payment-execution events — **threw `HR000068` on every run and enqueued nothing**. The standing-order execution pipeline was dead end-to-end; no due order ever executed (confirmed live in sandbox, #2148).

## Root cause
```kotlin
fun sweep(): Unit = runBlocking { … }   // runs on the bare scheduler executor-thread
```
`runBlocking` runs the coroutine on the scheduler's executor-thread, which carries **no Vert.x context**, so the first Hibernate-Reactive call (`findDueForExecution`) throws `HR000068: This method should exclusively be invoked from a Vert.x EventLoop thread`. It sits before the per-order try/catch, so the whole sweep aborts — 0 outbox rows. The documented repo footgun, here on the non-HTTP scheduler path.

## Fix
`suspend fun sweep()` — Quarkus dispatches a suspend `@Scheduled` method on a Vert.x context, exactly as the sibling `StandingOrderOutboxDispatcher.dispatch()` already does. Removed the unused `runBlocking` import. **3-line behavioural change.**

## Test (the coverage the mocked unit test lacked)
`StandingOrderExecutionSweepIT`: seeds a due ACTIVE order, drives `sweep()` through a **real Vert.x context + real DB**, asserts a `standing_order_outbox` row is enqueued. Ran green vs Testcontainers (`tests=1, failures=0`). The old `StandingOrderServiceTest` mocks the repo, so `executeOrders` never touched a real reactive session and the threading never ran — same masking class as #2077 / #1521. The issue's live sandbox E2E (HR000068 → 0 rows) is the fail-first evidence.

## Local gate
`detekt` + `ktlintCheck` + `compileTestKotlin` green; the IT green vs Testcontainers.

**money-path ⇒ 2 approvals + threat model. Held for review — not auto-merging.**

Refs #2148.

### DIFF
--- a/openbank-standing-order-service/src/main/kotlin/com/openbank/standingorder/infrastructure/scheduler/StandingOrderExecutionScheduler.kt
+++ b/openbank-standing-order-service/src/main/kotlin/com/openbank/standingorder/infrastructure/scheduler/StandingOrderExecutionScheduler.kt
@@ -12,7 +12,6 @@ import io.quarkus.scheduler.Scheduled
 import jakarta.enterprise.context.ApplicationScoped
 import jakarta.enterprise.event.Observes
 import jakarta.inject.Inject
-import kotlinx.coroutines.runBlocking
 import org.eclipse.microprofile.config.inject.ConfigProperty
 import org.jboss.logging.Logger
 import java.time.Clock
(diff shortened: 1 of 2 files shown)


JiRaska:
⚠️ **Duplicate — #2148 was already fixed and merged by PR #2180** (squash `9047a980`), which closed the issue. Two sessions picked up #2148 in parallel.

Flagging rather than closing, since this is another session's branch. Worth diffing before discarding: #2180 chose `suspend fun` over `VertxContextSupport.subscribeAndAwait` deliberately — every reactive `@Scheduled` in the monorepo is `suspend` or returns `Uni`, none uses `VertxContextSupport` in main code, and `subscribeAndAwait` *throws* if ever called from an event-loop thread. It also added `StandingOrderExecutionSweepIT`, which drives the **real cron** via a `@TestProfile` rather than calling `sweep()` directly — a direct call supplies the Vert.x context the scheduler does not, so it would pass against the broken code.

If this branch carries anything #2180 lacks, rebase it onto `main` and keep only that delta. Otherwise it can be closed.

Related follow-ups from #2180: the same defect was live in 4 more schedulers (#2187, fixed in #2190/#2191/#2194, with a CI guard in #2197), and the never-ran historical gap needs an owner decision (#2204).

JiRaska:
Superseded by #2180 (merged 2026-07-25 04:31 UTC), which resolved the same issue (#2148) with an identical production change — `StandingOrderExecutionScheduler.sweep()` converted from `fun … = runBlocking { }` to `suspend fun` so Quarkus dispatches it on a Vert.x context (fixing the HR000068 abort).

This branch is now 186 commits behind `main`; the only delta versus `main` on the scheduler is comment wording, and `main` already carries a more developed `StandingOrderExecutionSweepIT` (via #2180). Nothing unique left to merge, so closing as a duplicate.

Verified: `main` HEAD contains `suspend fun sweep()` + the IT; `git diff origin/main...` on this branch shows only the comment block differs.

PR #2199, opened 2026-07-25
chore(admin-ui): enforce governance.schema.json instead of hand-mirroring it

### LINKED ISSUE #2198: governance.schema.json is advisory — nothing validates governance.yaml against it (ADR-0071)
ADR-0071 says each per-service `governance.yaml` is "validated against `openbank-libs/governance/governance.schema.json`". Nothing does. The schema is referenced only from comments; `openbank-admin-ui/scripts/generate-governance.mjs` re-implements a **subset** of it by hand (REQUIRED list, 3 enums, the schemaName/stateless rule), so the two can and did drift.

## Live drift

Validating all 53 `governance.yaml` files with ajv 8 (2020-12 dialect) against the schema reports 4 failures the CI "Governance manifest gate" does not catch:

- `openbank-clearing-service`
- `openbank-dispute-service`
- `openbank-fx-service`
- `openbank-sanctions-service`

All four: `/lineage must be object`. Each file carries a bare `lineage:` key with no value, which YAML parses to `null`. The generator tolerates it via `decl.lineage ?? undefined`, so the gate stays green and the manifest silently drops the field.

## Fix

Drop the 4 empty keys, and make the schema the single source of truth: compile it with ajv in the generator, derive the required list and enums from it, and surface violations as ordinary `gaps` so the existing gate reports them.

Whole classes the hand-written rules cannot express today: wrong types, unknown top-level keys (`additionalProperties: false`), malformed `lineage` nodes, `minLength` violations.

JiRaska:
Closes #2198

## Problem

ADR-0071 states each per-service `governance.yaml` is "validated against `openbank-libs/governance/governance.schema.json`". Nothing did. The schema was referenced only from comments while `generate-governance.mjs` re-implemented a **subset** of it by hand — a REQUIRED list, three enums, the schemaName/stateless rule — so the two drifted.

**Live drift, on `main` today:** validating all 53 files with ajv 8 (2020-12) reports 4 failures the gate accepts — `clearing`, `dispute`, `fx`, `sanctions` each carry a bare `lineage:` key with no value. YAML parses that to `null`; the schema says `type: object`. `decl.lineage ?? undefined` swallows it, so the gate stays green.

## Change

- Drop the 4 empty `lineage:` keys.
- `generate-governance.mjs` **compiles** the schema with ajv and validates every file against it, and **reads** its REQUIRED list and enums from the schema rather than retyping them. Violations become ordinary `gaps`, so the existing CI gate names them — **no new CI step**.
- ajv `strict: true`, so a future schema keyword the validator cannot handle **throws** rather than being silently skipped. `strictRequired` is the one lever off: it is a heuristic that rejects the schema's `then: { not: { required: ["schemaName"] } }` idiom (schemaName is declared on the parent, not inside the negated subschema).
- The friendly, remedy-carrying messages stay and keep sole ownership of the defects they already name (a `handled` set), so one defect yields exactly one gap.
- **The one remaining hand-mirror is pinned to the schema's verdict.** The generator suppresses ajv's `#/allOf` errors so its own schemaName/stateless message — which spells out the remedy — is what an operator reads. That suppression is the same shape as the defect being fixed, so a second commit compares the two verdicts case-by-case across six documents (stateful, stateless-asserted, neither, both, `stateless: false` with and without a schemaName). Editing the schema's `allOf` branch without following in the generator now fails and names the disagreeing case.
- ajv 8 added as a **devDependency**; lockfile regenerated. The diff is only a hoist — ajv 8 moves to top level, ajv 6 nests under `eslint`/`@eslint/eslintrc`. No other package version moves. Every context that runs the generator (`ci.yml`, `admin-ui-deploy.yml`, `build-push-admin-ui.sh`, the `pretest` hook) does a full `npm ci` including dev deps.

## What the schema now catches that the hand rules never could

Wrong types, unknown top-level keys (`additionalProperties: false`), malformed `lineage` nodes, `minLength` violations.

## Validation against a known-positive

A green run proves nothing on its own, so every new test was confirmed red first.

**1. The schema layer.** With `if (false && !validateSchema(decl))`, four tests go red and the rest stay green:

```
× flags a bare `lineage:` key (null, not an object) — the live drift on 4 services
× flags an unknown top-level key (additionalProperties: false)
× flags a malformed lineage node (relationType missing / out of enum)
× flags a wrongly-typed optional field (evidenceExported must be boolean)
  Tests  4 failed | 10 passed (14)
```

A fifth candidate (empty string vs `minLength: 1`) passed with ajv disabled — it was already covered by the friendly REQUIRED rule, so it proved nothing and was replaced by the `evidenceExported` type case above.

**2. The agreement test.** Disabling the generator's `!stateless && !hasSchemaName` rule turns it red with the case named:

```
AssertionError: neither-stateless-nor-schemaname: generator and governance.schema.json disagree
```

**3. The gate itself**, run with the exact commands from `ci.yml` against a deliberate null `lineage:` appended to a real service file:

```
::error::Governance drift — 1 module(s) ...
  gap: openbank-fx-service: governance.yaml violates governance.schema.json — /lineage must be object
```

## Verification

- `node scripts/generate-governance.mjs` → `53 modules (38 with lineage, 0 gaps)`
- `npm run test` → 39 files, 662 tests passed
- `npm run type-check` → clean
- `npx eslint` on both changed files → clean
- `bash .github/scripts/check-adr-registry.sh` → OK, 187 ADRs
- All of the above re-run after merging current `main` into the branch.

### One flaky test, measured rather than waved off

Two of twelve full-suite runs failed on `docs/adr/[slug]/page.tsx mounts without throwing` (`Test timed out in 5000ms`). Rather than call it unrelated, it was measured: with this PR's 15 added tests **reverted**, the same checkout failed **7 of 12** runs — worse, not better. The flake is pre-existing contention in the shared vitest pool and this PR does not contribute to it. Filed separately with the numbers as #2235.

## Release impact

`chore` — a hidden type, so nothing releases despite touching four `<service>/governance.yaml` paths. Deliberate: this changes a validator and curatorial data, not shipped service behaviour.


JiRaska:
## Review: verified on the merits, held on CI state

Reviewed the 3-dot delta (`git diff origin/main...`) at tip `2356a5b6`, including the
main merge and `ce5b472c`. **No defects found.** Holding only because a required check
has not reported — see the bottom.

### Does it fight #2174? No.
`8633c059` (#2174) is already in this branch's merge base, and the branch:
- leaves `governance.schema.json` byte-identical (no diff),
- touches `ci.yml` **comment-only**, so #2174's rewritten gap reporter (the one that used
  to `TypeError` on `.modules`) is fully intact,
- takes `REQUIRED` from `SCHEMA.required`, which post-#2174 is the 5-element list *without*
  `schemaName` — so the conditional rule is preserved, not re-hardened.

Confirmed behaviourally — stateless services still pass with no `schemaName`, and all three
violation shapes still produce the friendly remedy message rather than ajv's terser wording:

| case | verdict |
|---|---|
| `stateless: true`, no `schemaName` | clean (correct) |
| `stateless: true` + `schemaName` | gap: "a stateless module owns no schema" |
| `stateless: false` | gap: "must be 'true' or omitted" |
| neither stateless nor `schemaName` | gap: "add 'stateless: true' instead…" |

### The 4 pre-existing violations
Fixed by the branch, and the gate is clean over the **whole fleet** on the merged tree:

```
[generate-governance] 53 modules (38 with lineage, 0 gaps)
```

It will **not** go red on merge. Also checked that no `governance.yaml`, the schema, the
generator or the lockfile changed on `main` since this branch's base, so nothing new can
have drifted underneath it.

### Known-positive proof
A gate that has only ever passed is unfalsified, so each path was driven to red and read for
what it *prints*, not just its exit code.

Restoring the four bare `lineage:` keys — the gate fails, exits 1, and **names every module**:

```
::error::Governance drift — 4 module(s) ...
  gap: openbank-clearing-service: ... violates governance.schema.json — /lineage must be object
  gap: openbank-dispute-service:  ... /lineage must be object
  gap: openbank-fx-service:       ... /lineage must be object
  gap: openbank-sanctions-service:... /lineage must be object
```

Each additional violation class, injected one at a time and reverted:

| injected | reported |
|---|---|
| unknown top-level key | `(root) must NOT have additional properties` |
| `evidenceExported: "nope"` | `/evidenceExported must be boolean` |
| `relationType: carrier-pigeon` | `/lineage/upstream/0/relationType must be equal to one of the allowed values` |
| unknown key inside a lineage node | `/lineage/upstream/0 must NOT have additional properties` |
| `dataDomain: banana` | friendly enum message (not the ajv one) — dedup works |

Disabling the ajv layer (`if (false && !validateSchema(decl))`) turns exactly the four new
schema tests red, 10 pass — they genuinely guard the change. And `ce5b472c`'s differential
test is falsifiable too: disabling the hand-mirrored `!stateless && !hasSchemaName` rule
fails with `neither-stateless-nor-schemaname: generator and governance.schema.json disagree`.

> One probe of mine was itself broken and worth flagging: an attempted `relationType` edit
> against `openbank-fx-service` reported "no gap" because that file has no `lineage:` block at
> all, so the edit changed nothing. Re-run against a service that does have one, it flags
> correctly. Same trap this PR is fixing.

### Other checks
`npm ci` clean from the lockfile; the lockfile diff is only the documented hoist (ajv 8 to top
level, ajv 6 under `eslint`/`@eslint/eslintrc`) with no other version moved. `type-check`
clean, `eslint` clean on both touched files, full suite **661 passed / 39 files**. No
`version.txt` or `CHANGELOG.md` touched. `chore` + admin-ui `scripts/` ⇒ nothing releases,
as intended.

> Incidental: `package-lock.json`'s own `version` was stale at `0.55.3` against `package.json`
> `0.56.0` on `main`; regenerating synced it to `0.56.0`. A side effect that corrects existing
> drift, not a hand-edit.

### Why this is held
On tip `2356a5b6` the required check **`all-green` is absent from the rollup entirely**, and
`build (clearing|dispute|fx|sanctions)` plus `CodeQL (java-kotlin)` are **CANCELLED** with
nothing in progress. Arming auto-merge now would leave it parked on a check that will never
report. Re-running CI on this tip should clear it — the content review above is complete and
found nothing blocking.

Issue #2187, opened 2026-07-25
Fleet sweep: runBlocking inside a @Scheduled method — the #2148 HR000068 footgun is live in up to 6 more schedulers, two in ledger-service


JiRaska:
## The defect class

#2148 (PR #2180): `StandingOrderExecutionScheduler.sweep()` was `fun sweep(): Unit = runBlocking { … }`. The Quarkus scheduler invokes a plain (non-`suspend`) `@Scheduled` method on an `executor-thread`, which carries **no Vert.x context**, so the first reactive Panache call throws:

```
HR000068: This method should exclusively be invoked from a Vert.x EventLoop thread;
currently running on thread 'executor-thread-1'
```

The sweep aborted before writing a single row. **No due standing order had ever executed.** It stayed hidden because the unit test mocks the repository, so no real reactive session is ever touched.

## Sweep result

`@Scheduled` + `runBlocking` in the same main-source file, on live `origin/main`:

| service | file | money-path? |
|---|---|---|
| **ledger** | `infrastructure/schedule/TieOutScheduler.kt` | **yes** |
| **ledger** | `infrastructure/schedule/FxRevaluationScheduler.kt` | **yes** |
| **fx** | `infrastructure/schedule/CnbRateIngestionScheduler.kt` | **yes** |
| billing | `infrastructure/scheduler/BillingCycleScheduler.kt` | |
| agent | `application/OversightService.kt` (`scheduledSweep`) | |
| analytics-sink | `infrastructure/reconcile/ReconciliationJob.kt` | |
| standing-order | `infrastructure/scheduler/StandingOrderExecutionScheduler.kt` | fixed in #2180 |

**This list is a grep hypothesis, not a verdict.** Co-occurrence in a file is not the bug: the bug is `runBlocking` wrapping a *reactive* call inside the `@Scheduled` method body. Some of these may wrap non-reactive work, or the `runBlocking` may sit in an unrelated method. Each needs reading before it is called broken.

`TieOutScheduler` deserves looking at first — if it is affected, the daily ledger tie-out has silently never run, which is the same control the balance-reconciliation work (#856) depends on.

## Fix pattern (established by #2180)

[...]

PR #2202, opened 2026-07-25
fix(billing): run the monthly cycle sweep on a Vert.x context (#2187)

JiRaska:
## What

`BillingCycleScheduler.sweep()` was a plain (non-`suspend`) `@Scheduled` method whose body was `runBlocking { runSweep() }` — the same shape as the #2148 standing-order defect. Makes it a `suspend fun` and adds an integration test that drives the **real cron** against a real Postgres.

## Why it is a bug

Quarkus invokes a plain `@Scheduled` method on a bare `executor-thread`, which carries no Vert.x context. The first reactive Panache query on the sweep path is `BillingAssessmentRepositoryImpl.findExisting`, whose first statement is `sf.withSession { … }` on a `Mutiny.SessionFactory` — so the sweep threw:

```
HR000068: This method should exclusively be invoked from a Vert.x EventLoop thread
```

Nothing on that path catches it: `runDiscoveredSweep`'s `try/catch` only wraps the discovery loop, and the CSV-override branch has no catch at all. The monthly cycle would have aborted on its first database touch.

**Blast radius: latent, not live.** The scheduler is disabled by default in every environment (`BILLING_SCHEDULER_ENABLED:false`, and `finops-scaledown/configmap-allowlist.yaml` records it as confirmed disabled), so no fee was ever mis-assessed. This would have fired the day an operator flipped it on — i.e. at go-live.

## Fix

`suspend fun sweep(): Unit = runSweep()`. That is the unanimous fleet convention (#2180, #2190, #2191): every other reactive `@Scheduled` method in the monorepo is `suspend` or returns `Uni`. Quarkus dispatches a suspending `@Scheduled` method on a proper duplicated Vert.x context — visible in the stack: `AbstractCoroutineInvoker` → `VertxDispatcher` → `ContextInternal.dispatch`.

## Test — and proof it can fail

`BillingCycleSweepIT` uses a `@TestProfile` that shrinks the cron to `*/2 * * * * ?` and re-enables `quarkus.scheduler` (`%test` disables it fleet-wide, ADR-0050), then polls a real Postgres (Testcontainers, via the existing `PostgresRedisTestResource`) for the assessment row.

It deliberately does **not** call `sweep()`/`runSweep()` directly: the defect is in *how the framework invokes the method*, so a direct call supplies the very context the scheduler does not, and would pass against the broken code. That is also why the mocked-collaborator `BillingCycleSchedulerTest` never saw it.

Proven by reverting the fix to `runBlocking` and re-running:

```
1 tests 1 fail 0 skip
FAILURE: AssertionFailedError: [a scheduler-dispatched sweep must persist a
billing_cycle_assessment row for billing-sweep-it-account in cycle 2026-07 —
no row means the sweep threw HR000068 off the Vert.x context (#2148 / #2187)]
Expecting value to be true but was false
```

with 123 `HR000068` lines in the captured log (one per 2-second tick). With the fix, the test passes in 0.98 s.

The assertion is scoped to this test's own account id, so nothing another IT leaves in the shared Postgres can affect it. There is no account-service in the profile, so the row is a fail-closed `skipped` assessment (`ACCOUNT_CONTEXT_UNRESOLVED`, same as `BillingCycleServiceIT`) — that a row exists *at all* is the point: it can only have been written from a Vert.x context.

## Money-path

`openbank-billing-service` is on `rules.yaml: money_path_services` → 2 approvals required. Threat model `docs/threat-models/openbank-billing-service.md` already exists and is unaffected: this change moves an existing job onto the correct thread, and adds no endpoint, no authorization decision, no data flow, and no new dependency.

## Scope note

This PR fixes billing only. The other two unresolved rows of #2187's table were traced and **cleared** — `agent-service` has no reactive persistence on its classpath (plain Agroal JDBC by design), and `analytics-sink` reaches only `Dispatchers.IO` + JDK `HttpClient`. Evidence is in the #2187 comment. The CI guard #2187 also asks for is not in this PR: a blanket ban on `runBlocking` in a `@Scheduled` body would fail both of those services, where it is the correct construct, so the rule has to be scoped to modules carrying `quarkus-hibernate-reactive-panache`. Tracked in #2187, which stays open.

## Verification

```
./gradlew :openbank-billing-service:detekt :openbank-billing-service:ktlintCheck \
          :openbank-billing-service:koverVerify :openbank-billing-service:build \
          :openbank-billing-service:quarkusBuild
BUILD SUCCESSFUL
```

No API change (no `openapi.yaml` touch), no DB migration, no event-schema change, no config-key change. `version.txt` untouched — `fix(billing)` under `src/main` releases via release-please.

Refs #2187, #2148.

### DIFF
--- a/openbank-billing-service/src/main/kotlin/com/openbank/billing/infrastructure/scheduler/BillingCycleScheduler.kt
+++ b/openbank-billing-service/src/main/kotlin/com/openbank/billing/infrastructure/scheduler/BillingCycleScheduler.kt
@@ -9,7 +9,6 @@ import com.openbank.billing.application.usecase.BillingCycleService
 import io.quarkus.scheduler.Scheduled
 import jakarta.enterprise.context.ApplicationScoped
 import jakarta.inject.Inject
-import kotlinx.coroutines.runBlocking
 import org.eclipse.microprofile.config.inject.ConfigProperty
 import org.jboss.logging.Logger
 import java.time.Clock
@@ -71,13 +70,27 @@ class BillingCycleScheduler {
 
     private val log = Logger.getLogger(BillingCycleScheduler::class.java)
 
+    // `suspend`, never `runBlocking` (#2148, and its fleet sweep #2187). Quarkus invokes a plain
+    // @Scheduled method on a bare `executor-thread`, which carries no Vert.x context, so
+    // `runBlocking { runSweep() }` ran the first reactive Panache query inside
+    // (`BillingAssessmentRepositoryImpl.findExisting`, via `sf.withSession`) off the event loop and
+    // threw `HR000068: This method should exclusively be invoked from a Vert.x EventLoop thread`.
+    // Nothing here catches that — `runDiscoveredSweep`'s try/catch only wraps the discovery loop,
+    // and the CSV branch has none — so the monthly cycle would have died on its first DB touch the
+    // day an operator flipped `openbank.billing.scheduler.enabled` on. A suspending @Scheduled
+    // method is dispatched by Quarkus on a proper (duplicated) Vert.x context instead.
     @Scheduled(
         cron = "{openbank.billing.scheduler.cron}",
         concurrentExecution = Scheduled.ConcurrentExecution.SKIP,
     )
-    fun sweep(): Unit = runBlocking { runSweep() }
+    suspend fun sweep(): Unit = runSweep()
 
-    /** The cycle-sweep logic, split out from the `@Scheduled` entrypoint so ITs can drive it directly. */
+    /**
+     * The cycle-sweep logic, split out from the `@Scheduled` entrypoint so unit tests can drive it
+     * directly. Note that a direct call cannot see the defect above — it supplies whatever context
+     * the caller is on, which is exactly the context the scheduler does *not* supply; only a
+     * genuinely scheduler-dispatched run proves it (see `BillingCycleSweepIT`).
+     */
     suspend fun runSweep() {
         if (!enabled) {
             log.debug("[billing-cycle-scheduler] Disabled — skipping sweep")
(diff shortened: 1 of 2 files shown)


JiRaska:
Closing as a duplicate — #2194 landed the identical fix (`suspend fun sweep()` + a cron-driving IT) at 8477733 while this branch was building, and the CI guard this PR's description said was still outstanding landed as #2197 at b7450c6. I branched from 398a0433 and did not re-check `origin/main` before opening; that is on me.

Nothing here is worth salvaging on top of #2194 — the fix is character-for-character equivalent, and `BillingCycleSweepVertxContextIT` covers what `BillingCycleSweepIT` did (same `@TestProfile` cron shrink to `*/2 * * * * ?`, same `quarkus.scheduler.enabled` re-enable, same real-Postgres poll). Keeping both would just double a 20-second Testcontainers IT.

Two things from this branch's work that are independently useful, and both are already reflected on `main`, so no follow-up is needed:

- The `agent-service` and `analytics-sink` rows of #2187's table are **false positives**, traced to the classpath rather than judged by grep: agent-service has no Hibernate Reactive at all (plain Agroal JDBC by design), and analytics-sink explicitly drops `hibernate-reactive-panache`/`reactive-pg-client`/`jdbc-postgresql`, reaching only `Dispatchers.IO` + JDK `HttpClient`. In both, `runBlocking` is the *correct* construct and `suspend` would be worse. #2197's `runblocking_allowlist` records exactly this, with the same reasoning.
- I validated #2197's guard against a known-positive rather than trusting its green: reverting billing's `sweep()` to `runBlocking` makes `check-no-runblocking-in-scheduled.py` emit the `::error::` and exit non-zero; on clean `main` it reports `62 @Scheduled method(s) checked across 1604 main-source file(s), 2 allowlisted — clean.` So the gate can actually fail, and its two allowlist entries are both live (a stale entry is itself an error). It is wired into `ci.yml` as an enforced step.

Issue #1919, opened 2026-07-23
Deploy the LLM inference gateway (ADR-0174) — agents call a US provider directly, egress unenforced (ADR-0175)


JiRaska:
ADR-0174 states plainly that the LiteLLM/vLLM gateway topology from ADR-0031 is **"not deployed at all"** — every LLM agent (copilot, agent-service, six ops agents) calls DeepInfra (US) directly, and ADR-0175 names un-enforced LLM prompt egress to a US provider an open data-residency exposure.

Scope:
- Deploy the inference gateway (LiteLLM proxy or equivalent) in-cluster as the single egress point for all LLM traffic
- Route all agents/copilot through it; NetworkPolicy so agent pods cannot reach LLM providers directly
- Per-agent budgets/rate limits + centralized prompt/response audit logging at the gateway (feeds the ADR-0148 audit chain)
- Decide + document the residency posture: EU inference endpoint, or documented derogation with the prompt-content classes that may egress

This is the highest-leverage single deploy in the AI stack: it turns four Partial ADRs (0031, 0148, 0174, 0175) into enforceable reality.

JiRaska:
**State after #2019 + #2024 (2026-07-24):**
- #2018 landed the shared `LlmGatewayPort` seam; #2019 deployed the in-cluster LiteLLM proxy + repointed **2 stub agents** (`control-liveness-sentinel`, `devops-agent`) at `http://litellm.ai-platform.svc:4000/v1` with an egress NetworkPolicy in `ai-platform`.
- **Not yet done (keeps this open):** the real hand-rolled clients (`copilot-service`, `agent-service`) still call DeepInfra directly; fleet-wide routing + the agent-pod egress-deny NetworkPolicy is incomplete.
- **`agents.yaml: model_gateway_as_built` is unchanged (`gateway: none`)** — correctly conservative until the real clients are routed AND verified in the live sandbox (deploy ≠ verified-running). The generated `docs/compliance/eu-ai-act.md` egress-gap section (#2024) therefore still shows the Art.10/GDPR gap, which is accurate.
- **When closing this:** refresh `model_gateway_as_built` (gateway/routing/providers) — that regenerates the AI-Act doc automatically and **restamps ~29 OPA bundles** (agents.yaml is embedded), so land it as a single governance PR and verify the routing against the live sandbox first.

JiRaska:
Three PRs open against this — splitting the routing change from the bundle churn so neither holds the other hostage.

[...]

PR #2209, opened 2026-07-25
fix(agent): route the assistant LLM call through the in-cluster gateway (ADR-0174/0175)

JiRaska:
Closes nothing on its own — this is the **routing half** of #1919. The egress-lock half ships
separately (see *Why the egress lock is not in this PR*).

## What was actually true on `main`

- The gateway **is** deployed and healthy (`litellm-59b9b4c4cf-9j65p`, `1/1 Running`), and
  `copilot-service` **is** live-routed at it (`COPILOT_MODEL_ENDPOINT=http://litellm.ai-platform.svc:4000/v1`
  on the running Deployment), as are six ops agents. The issue comment claiming the real clients
  still call DeepInfra directly is stale.
- `openbank-agent-service` was the genuine remaining bypass: `model-gateway.models[].endpoint` was
  hard-coded to `https://api.groq.com/openai/v1` and the pod held the provider key. Not DeepInfra —
  **Groq** — which is why the "repoint at the existing DeepSeek route" shortcut does not apply.

## Changes

| File | Change |
|---|---|
| `openbank-agent-service/.../application.yaml` | `endpoint` → `${AGENT_MODEL_ENDPOINT:https://api.groq.com/openai/v1}`; `api-key` → `${AGENT_MODEL_API_KEY:${GROQ_API_KEY:}}` |
| `ai-platform/litellm-config.yaml` | new `llama-3.3-70b-versatile` → `groq/llama-3.3-70b-versatile` route |
| `ai-platform/litellm.yaml` + `model-externalsecret.yaml` | `GROQ_API_KEY` projected into the gateway pod from the **existing** `openbank/agent-service` KV path — no new seeding |
| `external-secrets/es-agent-service.yaml` | `AGENT_MODEL_API_KEY` ← `litellm/LITELLM_MASTER_KEY` (same pattern as copilot) |
| `agent/agent-service.yaml` | `AGENT_MODEL_ENDPOINT` + `AGENT_MODEL_API_KEY`; `GROQ_API_KEY` retained for the rollout window only |

The model id is unchanged (`llama-3.3-70b-versatile` is sent verbatim upstream by
`OpenAiCompatibleModelProvider`), so this is an endpoint + key change, never a prompt change.

`gen-network-policies.py` re-run: **no diff** — `platform` is already in litellm's generated ingress
allow-list via copilot.

## Verification

- `ModelGatewayRoutingOverrideTest` (new): asserts `AGENT_MODEL_ENDPOINT` repoints the hosted model
  entry and `AGENT_MODEL_API_KEY` resolves as the backend key. **Proven to fail** against the
  hard-coded endpoint before the change (both tests FAILED, restored, both PASS).
- The nested default `${AGENT_MODEL_API_KEY:${GROQ_API_KEY:}}` was verified against SmallRye Config
  3.17.2 directly (neither set → empty; `GROQ_API_KEY` only → groq value; both → gateway value),
  with the probe validated against a known-positive single-level expression first. The test also
  asserts the resolved value contains no literal `${`, because a non-expanding expression would
  reach the gateway as a bogus key and 401 **silently**.
- `:openbank-agent-service:ktlintCheck detekt test` — BUILD SUCCESSFUL.

## Why the egress lock is not in this PR

`agent-service`'s `policyTypes` is still `Ingress`-only, so nothing prevents the pod reaching a US
provider. That is real and it is #1919's other half — but it **cannot land in the same commit**:

The image the manifest pins today (`sandbox-237c397c`, running now) predates the `AGENT_MODEL_*`
seam. It resolves the backend key from `GROQ_API_KEY` and the endpoint from the baked-in
`https://api.groq.com/openai/v1`. An egress policy that denies internet `:443` would break the
assistant *immediately on merge* and stay broken until release-please cuts the release, the image
builds, and the deploy PR bumps the tag. That is not a guess about the policy — it is what the
running pod does.

So the lock ships as a follow-up whose precondition is checkable: the repointed image is live **and**
`GROQ_API_KEY` has been dropped from `agent-service.yaml` + `es-agent-service.yaml`.

### The outbound-destination list the lock will be built from

Enumerated from `agent-service.yaml` env, `application.yaml`, and the source — each entry is a
source→destination pair, not "the namespace":

| Destination | Port | Where it came from |
|---|---|---|
| `kube-system` CoreDNS | 53/UDP+TCP | every name resolution; proven to work through a `namespaceSelector` rule by the existing `litellm-egress` policy |
| `litellm.ai-platform.svc` | 4000 | **new**, this PR |
| `keycloak.iam.svc` | 8080 | `QUARKUS_OIDC_AUTH_SERVER_URL` — both the resource-server JWKS and the outbound client-credentials token |
| `agent-db-rw.platform.svc` (CNPG) | 5432 | `QUARKUS_DATASOURCE_JDBC_URL`; same namespace, so an egress `podSelector: {}` rule covers it |
| `openbank-cluster-kafka-bootstrap.messaging.svc` + broker pods | 9093 | `KAFKA_BOOTSTRAP_SERVERS` (mTLS listener); the allow must cover the `messaging` namespace, not just the bootstrap Service, because the client reconnects to per-broker advertised addresses |
| `account-service.accounts.svc` / `product-catalog.accounts.svc` | 8100 / 8104 | MCP tool rest-clients |
| `balance-service.balances.svc` | 8103 | MCP tool rest-client |
| `transaction-service` / `clearing-service` / `sepa-instant` (`payments`) | 8102 / 8124 / 8127 | MCP tool rest-clients |
| `ledger-service.ledger.svc` | 8101 | MCP tool rest-client |
| `aml-service.aml.svc` | 8117 | MCP tool rest-client |
| `sanctions-service.sanctions.svc` | 8123 | MCP tool rest-client |
| `fx-service.fx.svc` / `interest-service.interest.svc` / `dispute-service.dispute.svc` | 8119 / 8125 / 8135 | MCP tool rest-clients — registered but the services are not deployed; the edges must still be allowed or onboarding them later fails mysteriously |
| `kube-prometheus-stack-prometheus` / `loki` / `kube-prometheus-stack-alertmanager` / `tempo` (`observability`) | 9090 / 3100 / 9093 / 4317 | observability read tools + OTLP export |
| OPA sidecar | localhost:8181 | same pod — loopback, NetworkPolicy does not apply |

Checked and **not** needed: the Kubernetes API (no kubernetes-client on the classpath), OpenBao (the
SVID CA cert is a static base64 value verified in-process, no runtime call), S3/AWS endpoints.
Three rest-clients (`:8106`, `:8107`, `:8118`) have no manifest env and so resolve to `localhost` —
they are dead, not hidden cross-namespace edges.

The reverse direction (admin-ui → `:8109`, observability → `:8085`, security-scanner → `:8085`,
kubelet probes) is unaffected: NetworkPolicy is connection-tracked, so replies on an established
inbound connection need no egress rule.

Refs #1919

### DIFF
--- a/openbank-agent-service/CLAUDE.md
+++ b/openbank-agent-service/CLAUDE.md
@@ -47,7 +47,10 @@ JAVA_HOME=$(/usr/libexec/java_home -v 25) ./gradlew :openbank-agent-service:test
 
 ## Config gotchas
 
-- `GROQ_API_KEY` must be set for the `llama-3.3-70b-versatile` model; omit for mock-echo.
+- `AGENT_MODEL_API_KEY` (or the legacy `GROQ_API_KEY` fallback) must be set for the
+  `llama-3.3-70b-versatile` model; omit for mock-echo. Deployed, that key is the LiteLLM
+  *virtual* key and `AGENT_MODEL_ENDPOINT` points at the in-cluster gateway — the provider
+  key never lands in this pod (ADR-0174/0175).
 - `AGENT_POLICY_ENFORCEMENT=block` requires a running OPA sidecar — without it the gate
   degrades to advisory and logs a WARN (fail-open by design).
 - `AGENT_OVERSIGHT_ENABLED=true` enables the scheduled compliance-officer sweep (every 30 min).
--- a/openbank-agent-service/src/main/resources/application.yaml
+++ b/openbank-agent-service/src/main/resources/application.yaml
@@ -172,10 +172,18 @@ model-gateway:
       provider: mock
       sensitivity: hosted
     # Groq — free, fast, OpenAI-compatible. `id` is sent verbatim as the upstream model name.
-    # Served by the single `openai-compat` adapter; needs GROQ_API_KEY (see agent.model.openai).
+    # Served by the single `openai-compat` adapter; needs an API key (see agent.model.openai).
+    #
+    # ADR-0174/0175 (#1919): in a deployed environment the endpoint is repointed at the in-cluster
+    # LiteLLM gateway (`http://litellm.ai-platform.svc:4000/v1`) via AGENT_MODEL_ENDPOINT, so the
+    # gateway is the single pod allowed to leave the cluster to a US LLM provider and the provider
+    # key never lands in this pod. The model id is unchanged — litellm-config maps
+    # `llama-3.3-70b-versatile` to `groq/llama-3.3-70b-versatile` — so this is an endpoint + key
+    # change only. The direct-to-Groq URL stays the default for local dev / `./gradlew quarkusDev`,
+    # where there is no gateway to route through.
     - id: llama-3.3-70b-versatile
       provider: openai-compat
-      endpoint: https://api.groq.com/openai/v1
+      endpoint: ${AGENT_MODEL_ENDPOINT:https://api.groq.com/openai/v1}
       sensitivity: hosted
     # The SAME openai-compat adapter serves any OpenAI-shaped backend — just change endpoint+key:
     # - id: qwen2.5                       # self-hosted Ollama for the SELF_HOSTED (sensitive) tier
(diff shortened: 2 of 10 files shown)


JiRaska:
Added a `openbank.tech/litellm-config-revision` annotation to the litellm pod template.

Caught on re-read: LiteLLM parses `--config` **once at startup** and `litellm-config` is mounted as a plain ConfigMap volume, so adding the `llama-3.3-70b-versatile` route would have updated the file and changed nothing — the proxy keeps serving the model list it booted with. agent-service's repointed call would then fail "model not found" while ArgoCD reports Synced and Healthy. Same silent-degradation shape as an `optional: true` secret ref: the failure is invisible from every angle except the one nobody checks.

The annotation makes a config edit roll the pod, mirroring the `openbank.tech/policy-checksum` pattern the OPA bundles already use. It is hand-bumped rather than a content hash because nothing generates this file — the comment says so at the edit site.

Issue #2231, opened 2026-07-25
Every OPA sidecar binds 0.0.0.0:8181 with no auth, and the netpol generator exposes it to every namespace allowed to reach the app port


JiRaska:
## Finding

`gen-network-policies.py` derives ingress ports from a Deployment's containerPorts. Adding an OPA sidecar therefore mechanically adds `8181` to the **same allow-list entry** as the application port — so every namespace permitted to call the service's API is also permitted to call its PDP directly. The sidecar binds `0.0.0.0:8181` and requires no authentication.

This is a **pre-existing fleet pattern across ~20 components on `main`**, not a regression from any single PR. It was noticed while reviewing the #1797 sidecar PRs (#2160, #2164, #2167, #2196) and deliberately not treated as grounds to hold them, since they only follow the established shape.

## Why it matters

The PDP is the component that decides whether a call is allowed. Direct reachability means a caller admitted to the app port can also:
- query the policy surface and enumerate what is and is not granted (an authorization map is useful reconnaissance), and
- depending on what the bundle exposes, evaluate arbitrary input against it.

It does not by itself grant access to data — the interceptor still consults the PDP for the real request — but it removes any confidentiality the policy had, and it widens the blast radius of any future sidecar endpoint that is not read-only.

## Suggested

Fix at the **generator** level so it applies fleet-wide rather than per-service:
- treat sidecar ports distinctly from app ports, and emit `8181` only for the sources that genuinely need it (in practice: nothing outside the pod — the interceptor calls `localhost:8181`), or
- bind the sidecar to loopback if the deployment shape allows it, which removes the question entirely.

Check first whether anything legitimately reaches 8181 cross-pod (a health scrape, the bundle-sync check, security-scanner) before narrowing — this repo has been burned by a NetworkPolicy validated from the wrong vantage point (#1640). Enumerate the real source→destination pairs, both directions.

Validate the generator change against a known-positive: confirm a component that previously emitted 8181 to a wide source set no longer does, and that nothing which needs it lost it.

Refs #1797, PR #2160/#2164/#2167/#2196, #2213.

JiRaska:
Still valid on `origin/main`; open PR **#2232** (`security(infra): stop publishing the loopback-only OPA sidecar port to callers`) targets it at the generator level as this issue asked. Branch `security/netpol-omit-sidecar-only-ports` is live in a worktree, so treat it as in-flight work rather than picking this up independently.

[...]

PR #2232, opened 2026-07-25
security(infra): stop publishing the loopback-only OPA sidecar port to callers

JiRaska:
## What

`openbank-infra/scripts/gen-network-policies.py` derives ingress ports mechanically from **every** container's `containerPorts`. The OPA PDP sidecar (ADR-0034 Phase 5, #1797) declares `containerPort: 8181`, so 8181 landed in the **same** derived ingress rule as the app's HTTP port on **29 components** — publishing the PDP to every namespace allowed to call the app.

For `kyc-service` that was `customer-edge`, `party`, `admin-ui`, `security-scanner`.

## Why it matters

1. **The app reaches the sidecar over loopback.** `OpaSidecarPolicyDecisionPoint.DEFAULT_BASE_URL` is `http://localhost:8181` and most components set no `OPA_URL` at all. Same-pod traffic is not policed by NetworkPolicy, so **8181 needs no cross-namespace ingress whatsoever**.
2. **The sidecar is unauthenticated.** `opa run --server --addr=0.0.0.0:8181 --bundle /bundle` — no `--authentication`, no `--authorization`. From an allowed namespace a pod could:
   - `POST /v1/data/openbank/rest/allow` with arbitrary input — a **policy oracle**;
   - `GET /v1/policies` / `GET /v1/data` — read the whole rego plus the `data.agents` / `data.rules` governance data;
   - `POST /v1/query`, `/v1/compile` — evaluate arbitrary rego, a **CPU-burn DoS** against a money-path pod;
   - `PUT /v1/data/<path>` for any root the mounted bundle's `.manifest` does not own.

   Decision **integrity held**: the bundle roots (`openbank/rest`, `openbank/agents`, `openbank/copilot`, `agents`, `rules`) are write-protected, so `openbank/rest/allow` itself could not be overwritten. The exposure is disclosure, oracle and DoS.

Pre-existing fleet-wide pattern already on `main`, **not** a regression from any single PR.

## Fix — in the generator, never a hand-edit

New `SIDECAR_LOCAL_ONLY_PORT_NAMES = {"opa"}`, applied alongside `INTERNAL_ONLY_PORT_NAMES`: such a port is kept out of `http_ports`, so it gets neither the derived-caller rule nor the admin-ui / ingress-nginx / KEDA rules.

All **38** sidecar port declarations across the fleet are named `opa`, and no Pod/ServiceMonitor scrapes it — so nothing else moves. The regenerated diff is exactly **76 deleted lines (38 × 2)** and nothing else.

## Residual, accepted

The unconditional same-namespace rule (`from: [{podSelector: {}}]`) still reaches 8181 from co-tenant pods. NetworkPolicy cannot express "loopback only", and narrowing that rule to a port list is a separate, larger change. Documented in `openbank-infra/CLAUDE.md`.

Binding OPA to `127.0.0.1` would close it fully but breaks the kubelet `httpGet` readiness/liveness probes (they hit the pod IP, and the distroless OPA image has no shell for an `exec` probe).

## Verification

Ran the generator with the exclusion **disabled** → 29 files regain `port: 8181`; **enabled** → 0. The gate was exercised against a known-positive rather than trusted on a clean pass. Regen is idempotent (second run produces no further drift).

Refs #1797

### DIFF
--- a/openbank-infra/CLAUDE.md
+++ b/openbank-infra/CLAUDE.md
@@ -98,6 +98,21 @@ out of it (they are path-scoped, not less important — several are live-inciden
   workloads), but the reflex stands: verify with
   `kubectl get netpol -n <ns>` / `grep -rl '<workload>-ingress-allow-list' gitops/components/`,
   never with `ls components/<svc>/`.
+- **The generator derives ingress ports from EVERY container's `containerPorts` — a sidecar port
+  that only ever serves loopback must be named into `SIDECAR_LOCAL_ONLY_PORT_NAMES`, or it is
+  published to the app's whole caller set.** The OPA PDP sidecar is the case that forced the rule:
+  the app reaches it at `http://localhost:8181` (`OpaSidecarPolicyDecisionPoint.DEFAULT_BASE_URL`,
+  no `OPA_URL` env in most components), so it needs no cross-namespace ingress at all — yet 8181
+  landed in the SAME derived rule as the app's HTTP port on 29 components, admitting every declared
+  caller namespace (for kyc: customer-edge, party, admin-ui, security-scanner) to an `opa run
+  --server` with no `--authentication` and no `--authorization`. That is a policy ORACLE
+  (`POST /v1/data/openbank/rest/allow` with arbitrary input), full disclosure of the rego plus
+  `data.agents` / `data.rules` (`GET /v1/policies`, `GET /v1/data`), arbitrary rego evaluation
+  (`POST /v1/query`, a CPU-burn DoS on a money-path pod) and writes to any data root the bundle's
+  `.manifest` does not own. Decision *integrity* held — a bundle root is write-protected, so
+  `openbank/rest/allow` itself cannot be overwritten. Fix is in the generator only; never hand-edit
+  a derived `network-policies.yaml`. Residual, accepted: the unconditional same-namespace rule still
+  reaches 8181 from co-tenant pods — NetworkPolicy cannot express "loopback only".
 - **The generator's `URL_RE` requires a literal `.svc` — `http://kyc-service.kyc:8114` does NOT
   match, `http://kyc-service.kyc.svc:8114` does.** Write the short form and the generator exits 0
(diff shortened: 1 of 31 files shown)


JiRaska:
## Independently verified against the live sandbox before merging

This PR narrows ingress on 29 components, so the risk it carries is the #1640 shape: a NetworkPolicy validated from the wrong vantage point, where the in-namespace probe passes and the real caller is blocked. #2231 asks for exactly this check — *"enumerate the real source→destination pairs, both directions"* — so here it is, run against the running cluster rather than inferred from the manifests.

| probe | result |
|---|---|
| Services exposing port 8181 | **0** |
| ServiceMonitors scraping 8181 | **0** |
| Pods with an `OPA*URL` env | **3** — `agent-service`, `ap2-service`, `mcp-service` |
| …of those, pointing anywhere but loopback | **0** (all `http://localhost:8181`) |
| `OpaSidecarPolicyDecisionPoint.DEFAULT_BASE_URL` | `http://localhost:8181` |

So nothing reaches a PDP cross-pod today, and the word doing the work in the PR body — *"no `OPA_URL` env in **most** components"* — holds for all three that have one, not most.

Two things I checked specifically because they would have made this unsafe:

- **A metrics scrape.** If Prometheus scraped 8181 through a Service, narrowing would silently drop the PDP's metrics and nothing would go red — the failure would be a missing series, which is invisible until someone asks a question of it. There is no such Service and no such ServiceMonitor.
- **A cross-pod `OPA_URL` override.** One component pointing at another pod's PDP would be broken by this change and would fail closed on a money-path surface. None does.

The residual the PR already states — the unconditional same-namespace rule still admits co-tenant pods to 8181, because NetworkPolicy cannot express "loopback only" — is accurate and correctly left as accepted rather than papered over.

Arming auto-merge on that basis. Fixing it in the generator rather than in the 29 derived files is the right level, and `SIDECAR_LOCAL_ONLY_PORT_NAMES` makes the next loopback-only sidecar a one-line declaration instead of a repeat of this finding.

Closes #2231.

Issue #2206, opened 2026-07-25
BLOCKER for MCP phase 2: binding real read ports before caller authentication turns get_balance into an unauthenticated read of any account


JiRaska:
## The ordering constraint

The threat model landed in PR #2200 (`docs/threat-models/openbank-mcp-service.md`) establishes that `openbank-mcp-service` today has **no caller authentication at all**:

- `McpEndpoint.resolveContext()` hardcodes `ConsentContext("agent:mcp-anonymous", "none", emptyList())`. It does **not** read the `X-Agent-Id` / `X-Consent-Id` headers **its own KDoc claims it reads.**
- `quarkus.oidc.tenant-enabled: false`, no `@RolesAllowed`, no mTLS.
- The PDP therefore authorizes a **constant**. It is working correctly; it is being asked the wrong question.
- No consent enforcement code exists anywhere in the service. `ReadPorts.kt` delegates the `grantedAccounts` intersection to "the port implementation", and the only implementation is `StubReadPorts`, which enforces nothing.

**This is currently harmless only because the read ports are stubs.** The stub boundary is the load-bearing security control of this service, and no gate protects it.

## The consequence to avoid

Phase 2's stated plan is to swap `StubReadPorts` for real `@RegisterRestClient` adapters. That change needs **no edit to the endpoint and no edit to the policy** — which is exactly what makes it dangerous. The moment it lands, `get_balance(accountId)` becomes an **unauthenticated read of any guessable account id**, and every gate in CI stays green, because nothing in CI knows the stub was the control.

## Therefore

**Caller authentication and consent binding must land BEFORE, or atomically with, the real read ports** — not as a follow-up. Concretely, before any real adapter is bound:

1. Real caller identity (OAuth 2.1 / PSD2 consent binding, per the phase-2 plan) actually read and propagated, not just documented.
2. `resolveContext()` reading the headers its KDoc already claims — or the KDoc corrected, so code and comment stop disagreeing.
3. Consent-scoped account filtering implemented somewhere real, not delegated to an implementation that does not exist yet.
4. Ideally a CI guard that fails if a non-stub `ReadPorts` implementation is wired while `resolveContext` still returns a constant.

## Related policy-input gap

[...]

PR #2316, opened 2026-07-25
feat(mcp): remove the phase-1 placeholder identity and wire real read ports (ADR-0195 step 4)

JiRaska:
## What

The atomic cutover. ADR-0195/#2206 required caller authentication to land **before, or atomically with**, the real read ports — never as a follow-up. Steps 1–3 (#2253/#2262/#2278) prepared everything without changing deployed behavior; this flips it.

## Changes

- **`application.yaml`**: `quarkus.oidc.tenant-enabled=true`. Without this Quarkus never inspects the Authorization header, so `CallerContextResolver` could never see a real `sub`/`consent_id` no matter what a caller sent. No `@Authenticated`/permission policy added — a request with no token still reaches `McpEndpoint` (anonymous `JsonWebToken`); `CallerContextResolver`'s null-check is what denies it, not a framework 401. A present-but-invalid/expired token IS rejected by Quarkus OIDC itself first.
- **`RealAccountReadPort`**: `@Vetoed` removed, `@RestClient` qualifier added to its four downstream-client constructor params (Quarkus's REST client extension produces `@RestClient`-qualified beans, not `@Default` — the earlier `@Vetoed` state never exercised this). Sole `AccountReadPort` bean now.
- **`StubReadPorts.kt` → `StubProposalPort.kt`**: `StubAccountReadPort` deleted. `StubProposalPort` stays (a real `ProposalPort` needs new cross-service API surface — copilot-service's `ActionProposal` domain is internal, not callable — out of scope, flagged in #2262).
- **`McpEndpoint.resolveContext()`**: the `agent:mcp-anonymous` fallback and literal are **gone**. No token ⇒ no `ConsentContext` ⇒ denied ("Authorization unavailable"), same fail-closed path as a malformed token (`CallerIdentitySource.RESOLUTION_FAILED`). `ANONYMOUS_FALLBACK` can no longer fire; kept in the enum as blocker #2206's historical record — the metric that tracked it now reads zero **by construction**.

## Test fallout (all real, not incidental)

- **`McpEndpointTest`**: every protocol/PDP/audit/metrics test needs a validated agent identity now (removing the fallback removes the free ride via an anonymous JWT). Default `endpoint()` helper switched to a token-bearing `TestJsonWebToken`; added dedicated "no token → denied, metered `resolution_failed`" coverage; the old "anonymous_fallback" metering test is now "source=token" (the fallback source is unreachable, not merely untested).
- **`McpAuditEventIT`** (`@QuarkusTest`, real HTTP + real CDI): needs a real-shaped identity too (`@TestSecurity` + `@OidcSecurity`, new `quarkus-test-security-oidc` dep — added to the version catalog). Discovered along the way: `RealAccountReadPort` is now the live `AccountReadPort` bean in this REAL CDI graph, so an "allowed" call tried to reach the real account/balance/transaction/consent-service REST clients, which don't exist in a test JVM — same class of problem `TestPolicyDecisionPoint` already solves for the OPA sidecar. Added `TestAccountReadPort` (`@Alternative @Priority(1)`), same pattern. Added a dedicated "no token" case here too.

## Verification

`detekt` + `ktlintCheck` + `test` (41/41) + `quarkusBuild` all green. Guard #2230 now reports **"placeholder identity removed; real ports permitted"** — the state it exists to gate on.

## Post-merge

ArgoCD rolls the new image + `tenant-enabled=true`. Every current caller has no real agent token yet (no token-minting flow exists), so `tools/call` will consistently return "Authorization unavailable" until one does — the secure default, not a regression (the phase-1 stub's fake-success JSON is retired on purpose).

Refs ADR-0195, #2206, #2230, #2253, #2262, #2278


JiRaska:
Reopening to retrigger CI — push at 5086694a produced zero workflow runs (webhook miss), repo-wide Actions otherwise healthy.

Issue #2322, opened 2026-07-25
tech-debt(copilot): no contract test covers the seven customer-edge calls copilot depends on


JiRaska:
Split out of #2255 (copilot C3). #2255 is closed by documenting + route-set-guarding copilot's **own** served contract; this issue tracks the other half — copilot's **consumed** contract, which nothing verifies today.

## What is unverified

`openbank-copilot-service/src/main/kotlin/com/openbank/copilot/infrastructure/client/CustomerEdgeClient.kt` declares a `@RegisterRestClient(configKey = "customer-edge")` interface with seven routes:

| method | path | client method |
|---|---|---|
| GET | `/customer/v1/accounts` | `listAccounts` |
| GET | `/customer/v1/balances/{accountId}` | `getBalances` |
| GET | `/customer/v1/transactions` | `listTransactions` |
| GET | `/customer/v1/fx/rates` | `getFxRates` |
| GET | `/customer/v1/cards` | `listCards` |
| GET | `/customer/v1/statements/{accountId}` | `listStatements` |
| GET | `/customer/v1/standing-orders` | `listStandingOrders` |

Every copilot read tool (`AccountsTool`, `BalanceTool`, `TransactionTool`, `FxRatesTool`, `CardStatusTool`, `StatementsTool`, `ScheduledPaymentsTool`, `AccountsWithBalancesTool`) is a thin wrapper over one of these, and their unit tests mock the port — so a wrong path, a renamed response field, or a dropped route on the provider side is invisible until runtime, where it degrades every assistant answer that quotes a figure.

This matters more than for an ordinary client: copilot propagates the **customer's** bearer on these calls (on-behalf-of, ADR-0065 / ADR-0089 D5), so the response shapes are the only thing standing between the model and a figure it narrates to the customer. ADR-0089's "the model proposes, the bank disposes" only holds if every figure came from a tool result that really parsed.

## Why a consumer-only pact was deliberately NOT shipped

Per `CLAUDE.md` (Contract tests / Pact): **a consumer pact alone cannot catch a wrong request path** — the Pact mock server answers whatever path the client asks for, so only the provider replay reddens. That is exactly how finrep shipped a call to a ledger path that never existed (#2269). `openbank-customer-edge` cannot replay a pact today:

- no pact dependency and no `@Provider` / `@PactFolder` verification test anywhere in the module;
[...]

Issue #2255, opened 2026-07-25
Prod Readiness: bring every matrix cell to 2 (Verified) — 44 cells, of which 13 are collector probe bugs


JiRaska:
Goal: **every cell of the Prod Readiness matrix at 2 (Verified) or better.**

## Status (2026-07-25, re-verified against `origin/main` @10e6e448 via `openbank-infra/scripts/prod-readiness-collector.py --all --json`)

**All 44 originally-listed below-2 cells are now closed. 0 cells below 2, across all 37 services × 9 dimensions.**

Closed by:
- #2257 — C8 podmonitor namespace gap (anacredit, billing, documents, finrep, sdd, tpp-registry)
- #2266 — collector-probe fixes: C4 stateless-detection, C5 same, C8 substring-match false negative, C1 filename-match
- #2261 — C6+C9 generated runbooks (ap2, billing, document, finrep, mcp, settlement, vop)
- #2291 and follow-on Pact test PRs — C3 contract tests (aml, ap2, card-issuance, copilot, mcp, vop) + finrep openapi.yaml
- individual PRs closing the remaining C1 (agent/copilot port packages, ap2/mcp skeleton growth) and C4 (settlement rollback note) items

## Money-path gate — still correctly NO-GO

`compute_gate` requires **3** on C1/C5/C7 for the 11 money-path services, and 3 comes only from a fresh TTL'd attestation in `openbank-libs/governance/attestations.yaml` (`code_complete`, `restore_drill`, `pentest`). All-2 does not clear this gate. Currently NO-GO: `account balance clearing consent fraud fx ledger lending sca swift transaction`. Those attestations are deliberately out of scope for this issue and must never be written just to make a cell green.

---

<details><summary>Original issue text (baseline @2e5c6d41, superseded above)</summary>

Baseline measured by running `openbank-infra/scripts/prod-readiness-collector.py --all` against a clean worktree of `origin/main` @ 2e5c6d41: **44 cells below 2** across 37 services (11 GO / 26 NO-GO). Per dimension: C1 6 · C3 7 · C4 5 · C5 1 · C6 7 · C8 11 · C9 7.

The 44 are three different problems and must not be treated alike.

### 1. A real production gap — 5 cells (C8)

[...]

Issue #2314, opened 2026-07-25
Nothing compares an openapi.yaml to the code that serves it — 22 of 35 specs declare the wrong port


JiRaska:
## Nothing anywhere compares an openapi.yaml to the code that serves it

ADR-0005 states the fleet is OpenAPI design-first and that "CI lints it with Spectral so docs
cannot drift from runtime". There is no Spectral config in the repo and no workflow references it.
The control named in the ADR does not exist. `.github/scripts/check-api-contract.py` is the only
thing that reads these files, and it checks *version arithmetic between two revisions of the
document* — never the document against the service.

This is the #2280 failure class again: check whether anything reads the artifact at all before
assuming a green covers it.

## Measurement

The cheapest fully mechanical drift signal — the spec's declared dev server port vs
`quarkus.http.port` in the same service's `application.yaml`:

```
match=13  mismatch=22  unparsed=9   (of 44 services with an openapi.yaml)
```

Probe validated in both directions (it returns matches as well as mismatches, so a null result
would have been visible as a bug rather than as evidence of absence).

Mismatching: agent, aml, audit, card-issuance, clearing, consent, dispute, domestic-payment, fx,
interest, kyc, ledger, party, pid, psd2, sca, security-scanner, sepa-instant, sepa-payment,
standing-order, tpp-registry, transaction.

The port is only a proxy for "is this file maintained at all". Where it was checked properly —
aml (#2312) — the schemas were fiction too: invented request fields, invented response fields,
an enum with two values the service rejects.

## Why this is not cosmetic

finrep-service shipped a call to `/api/v1/ledger/trial-balance`, a path that never existed (#2269),
and card-issuance had the same shape (#2283). A spec that nobody checks against the code is the
document those clients are written from.

## Proposal

Every service already depends on `quarkus-smallrye-openapi`.
`quarkus.smallrye-openapi.store-schema-directory` — used nowhere in the repo — writes the generated
document at build time, and CI already builds each changed service. A conformance gate can diff the
committed `openapi.yaml` against the generated one for roughly the cost of a file comparison.

Sequencing note: the gate must land advisory and the 22 specs be corrected first, because a
correction is currently blocked by #2313.

Refs #2312, #2313

JiRaska:
## Prior art landed while this was being written: #2323

`CopilotApiContractTest` (PR #2323, `openbank-copilot-service`) is the first spec-vs-code check in
[...]

Issue #2313, opened 2026-07-25
api-contract gate is unsatisfiable for a spec correction — and forces the false value to be kept


JiRaska:
## The gate cannot be satisfied by a truthful spec correction

Measured on #2312 with the CI-pinned `oasdiff` 1.22.0, running
`.github/scripts/check-api-contract.py --enforce` locally — both ends, not reasoned about:

- `info.version` **1.0.0** -> `::error:: breaking API change requires an info.version MAJOR bump`, exit 1
- `info.version` **2.0.0** -> `::error:: major(info.version)=2 but openbank.api.version=1` and
  `::error:: major(info.version)=2 but the newest served URL major is /api/v1` (ADR-0048 D2/D4), exit 1

There is no third value. The gate assumes the base document described a served contract (ADR-0005
design-first). When the base document is fiction, that assumption is false and the rule becomes
unsatisfiable.

## It does not merely block the fix — it requires the falsehood to be kept

`oasdiff breaking` levels for the correction:

| level | id | n |
|---|---|---|
| 3 (ERR) | `new-required-request-property` | 6 |
| 3 (ERR) | `request-parameter-enum-value-removed` | 2 |
| 3 (ERR) | `new-required-request-parameter` | 1 |
| 3 (ERR) | `request-parameter-default-value-changed` | 1 |
| 2 (WARN) | `response-optional-property-removed` | 9 |
| 2 (WARN) | `request-property-removed` | 6 |

Removing the invented fields is level 2, so it passes. Only *stating the truth* is ERR. And the
obvious dodge — omit rather than misstate — is itself ERR: dropping the wrong `default` is
`request-parameter-default-value-removed`, and dropping the phantom `enum` is
`request-parameter-enum-value-removed` x5. The only green spec is one that repeats
`default: 20` (real: 50) and keeps `CLOSED`/`SAR_FILED` in the `status` enum.

So #2312 ships four constraints as `GATE-CONSTRAINED` `description` prose:

1. `Idempotency-Key` marked `required: false` though `createCase` rejects its absence
2. `customerReference`/`screeningType`/`riskLevel`/`alertCode` absent from `required`
3. `targetStatus`/`decidedBy` absent from `required`
4. `limit` declaring `default: 20`, and the two phantom `status` enum values retained

## What NOT to do

An "this spec was never served" escape hatch, declared in the PR body or a marker file. It is
unverifiable prose — the exact failure class as scoring a contract test by grepping for the word
[...]

PR #2323, opened 2026-07-25
fix(copilot): document the action-confirm endpoint and guard the route set

JiRaska:
Closes the C3 cell for `openbank-copilot-service` (#2255): it had an `openapi.yaml` and no contract test.

## The defect

`ActionConfirmResource` serves `POST /api/v1/copilot/actions/{tokenId}/confirm` — the human-in-the-loop confirmation for a money-path action proposal (ADR-0089 D2), the most consequential call copilot serves — and the published `openapi.yaml` did not mention it at all. Any client generated from that spec is missing the only route that disposes of a proposal, so this is `fix`, not `docs`.

The two paths the spec *did* declare (`/copilot/chat`, `/copilot/chat/stream`) are correct: with `servers: [- url: /api/v1]` they resolve to the served `/api/v1/copilot/chat{,/stream}`. (Worth stating because reading `paths:` without `servers:` makes them look wrong.)

## Spec change

Adds the path with the shapes `ActionConfirmResource` really returns — `200 {status, actionId}`, `401` (authenticated token with no usable subject), `403` (OPA deny **or** token-ownership mismatch, deliberately indistinguishable from not-found in the body), `404` (unknown **or** malformed token id), `422` (expired) — plus the `ActionConfirmation` and `ErrorResponse` schemas. `info.version` 1.0.0 -> 1.1.0.

Classification verified with the gate rather than by arithmetic — `check-api-contract.py --base origin/main --enforce` prints:

```
api-contract gate: openbank-copilot-service: additive change, info.version 1.0.0 -> 1.1.0 — OK (required >= MINOR).
```

`version.txt` / `CHANGELOG.md` untouched — release-please owns the release axis (ADR-0048 D1).

## Test shape, and why

`CopilotApiContractTest` is **static/reflective**: it reads the `@Path` / `@POST` … annotations off the resource classes by reflection and compares that route set with the spec's `paths:` × HTTP-method set, resolved against `servers[0].url`. Chosen over a `@QuarkusTest` + RestAssured shape because the question it answers — *does the served route set equal the declared route set* — needs no HTTP: the runtime shape would have to stub the model gateway for the two chat routes and mint a bearer for the confirm route, buying fidelity this assertion cannot spend. Behaviour is already covered by `ActionConfirmResourceTest` (6 tests) and CDI wiring by `CopilotServiceBootTest`.

The KDoc says plainly what it does not prove: no Quarkus boot, no HTTP, no status codes or bodies, and it cannot catch a route that exists as an annotation but fails at runtime. (An overclaiming KDoc is its own defect — #2283/#2300.)

It is **bidirectional**, which is the point: direction 2 ("declared but not served") is ordinary drift, direction 1 ("**served but not declared**") is the one that would have caught this endpoint — and keeps it caught. The explicit resource-class list is the obvious staleness risk, so a fourth test pins its size to the number of `*Resource.kt` files on disk.

Named `*ContractTest.kt` so the C3 scorer detects the artifact, not the prose (#2291).

## Falsification — all three guards were made to fail, then restored

| perturbation | what the test printed |
|---|---|
| removed `/copilot/chat` from the spec | `[served by a @Path resource but MISSING from openapi.yaml] Expecting empty but was: ["POST /api/v1/copilot/chat"]` |
| added a bogus `/copilot/bogus/never-served` GET to the spec | `[declared in openapi.yaml but NOT served by any @Path resource] Expecting empty but was: ["GET /api/v1/copilot/bogus/never-served"]` |
| added a dummy `DummyProbeResource.kt` | `[a new *Resource.kt must be added to resourceClasses…] actual size is: 3 while expected size is: 2` |

Test count read from the JUnit XML, not the console: `tests="4" skipped="0" failures="0" errors="0"` on the green run, `failures="1"` on each perturbation.

## Out of scope, tracked instead

Copilot also *consumes* seven `/customer/v1/*` routes from `openbank-customer-edge`. `openbank-customer-edge` has no pact dependency, no `@Provider` test and no HTTP boot-test harness, and per `CLAUDE.md` a consumer pact whose provider never replays it cannot catch a wrong path at all — so a consumer-only pact would have been theatre. Raised as **#2322** (`tech-debt`) with the seven routes, the bearer-propagation stake, and what a provider harness in customer-edge needs.

## Gates run locally

- `:openbank-copilot-service:build :openbank-copilot-service:quarkusBuild` — `BUILD SUCCESSFUL`
- `:openbank-copilot-service:koverVerify` — `BUILD SUCCESSFUL`
- `check-api-contract.py --enforce` — exit 0, output above

Refs #2255

### DIFF
--- a/openbank-copilot-service/src/main/resources/openapi.yaml
+++ b/openbank-copilot-service/src/main/resources/openapi.yaml
@@ -7,7 +7,7 @@ openapi: "3.1.0"
 
 info:
   title: OpenBank Copilot Service API
-  version: "1.0.0"
+  version: "1.1.0"
   description: |
     Customer-facing AI assistant (mobile copilot), per ADR-0089.
 
@@ -84,6 +84,63 @@ paths:
         "501":
           description: Capability not yet enabled (feature flag off)
 
+  /copilot/actions/{tokenId}/confirm:
+    post:
+      operationId: confirmAction
+      summary: Confirm a copilot action proposal (human-in-the-loop)
+      description: |
+        Exchanges a one-time proposal token for a confirmed action. This is the
+        human-in-the-loop half of "the model proposes, the bank disposes" (ADR-0089 D2): the
+        assistant never executes a money-path action, it only emits a proposal the customer
+        confirms here from a non-AI-controlled card.
+
+        The token is bound to the customer that owns it — a token confirmed by a different
+        subject is refused, and the response deliberately does not distinguish that case from
+        an unknown token. The confirm is re-checked against the OPA policy gate (ADR-0034,
+        fail-closed: an unreachable PDP denies), and the token is single-use — it is deleted on
+        confirm, on expiry, and on an ownership mismatch is left untouched.
+      security:
+        - bearerAuth: []
+      parameters:
+        - name: tokenId
+          in: path
+          required: true
+          description: The proposal token id handed out with the `ActionProposal` (UUID).
+          schema:
+            type: string
+            format: uuid
+      responses:
+        "200":
+          description: Action confirmed
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ActionConfirmation"
+        "401":
+          description: Authenticated token carries no usable subject
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ErrorResponse"
+        "403":
+          description: OPA policy denied the action, or the token belongs to another customer
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ErrorResponse"
+        "404":
+          description: Token not found (also returned for a malformed token id)
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ErrorResponse"
+        "422":
+          description: Token expired (the proposal must be re-issued by a new turn)
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ErrorResponse"
+
 components:
   securitySchemes:
     bearerAuth:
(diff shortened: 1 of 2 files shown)


JiRaska:
Cross-linking — same root cause, found independently on `openbank-aml-service` under the same
issue (#2255) and merged as #2317 a few minutes before this PR opened.

`CopilotApiContractTest` is the first spec-vs-code check in the repo, and it is worth saying out
loud that **nothing else does this anywhere in the fleet** (#2314). ADR-0005 claims "CI lints it
with Spectral so docs cannot drift from runtime" — there is no Spectral config and no workflow
references it; `check-api-contract.py` only does version arithmetic between two revisions of the
*document*. The cheapest mechanical drift probe across the fleet, spec `servers` port vs
`quarkus.http.port`, reads match=13 / mismatch=22 / unparsed=9.

Two things from the aml side that may save you time:

1. **The api-contract gate is unsatisfiable for a spec correction** (#2313). Measured both ends with
   the CI-pinned oasdiff 1.22.0: at the old version the diff is breaking so a MAJOR bump is
   demanded, and at MAJOR+1 the ADR-0048 D2 invariant rejects it because the served URL major is
   still `/api/v1`. It only bites when you *tighten* something — adding a `required` property or a
   required parameter, removing an enum value, changing a `default` are all ERR. This PR reads as
   additive on a new endpoint, so it passes; the wall is there the moment a correction has to
   narrow an existing declaration. Omitting rather than misstating does not help — dropping a wrong
   `default` and dropping a phantom `enum` are ERR too.

2. **Route-set conformance has a convention trap worth knowing about before this generalises.**
   This service resolves paths against `servers[0].url` = `/api/v1`; `openbank-aml-service` puts the
   full `/api/v1/...` in `paths` and uses `servers` for an absolute dev URL
   (`http://localhost:8117`). A fleet-wide version of this test has to handle both, or it will
   report confident nonsense on half the services.

For the schema half of the problem (invented request/response fields, not just missing routes),
`quarkus-smallrye-openapi` is already a dependency everywhere and
`quarkus.smallrye-openapi.store-schema-directory` is used nowhere — diffing the committed spec
against the generated document would cover routes *and* schemas from the build CI already runs.
Noted as the proposal in #2314; this PR is better prior art than what I wrote there, so I have
referenced it.

Refs #2312, #2313, #2314

PR #2337, opened 2026-07-25
fix(admin-ui): stop displaying the card network name on read-only rows

JiRaska:
No trademark license for VISA/Mastercard yet.

The card detail view (`/cards/[id]`) showed the network in three read-only places: the header subtitle line, a "Síť" row, and the entitlements "Sítě" row.

**The issue-flow network selector in `IssueCardDialog` is deliberately untouched.** There the operator picks a real network to send to card-issuance — it requires one of its enum values (VISA/MASTERCARD/AMEX/UNIONPAY) and the choice has real fee/acceptance consequences. Anonymising that label would block the operator from making a correct decision rather than just hiding a brand name; that's a functional control, not a display of the mark, and out of scope here.

## Verification

```
npm run type-check → clean
npm run lint        → 0 errors (80 pre-existing warnings)
npm test             → 47 files, 757 tests passed
```

Companion: openbank-app PR removing the network wordmark from the customer app's card face.

## Linked issues

N/A — no tracked issue exists for this; it is a direct legal/compliance instruction (no trademark license for VISA/Mastercard yet), acted on immediately.

### DIFF
--- a/openbank-admin-ui/src/app/cards/[id]/page.tsx
+++ b/openbank-admin-ui/src/app/cards/[id]/page.tsx
@@ -191,7 +191,7 @@ export default function CardDetailPage() {
                   <CardStatusChip status={card.status} current />
                 </div>
                 <p style={{ fontSize: '12.5px', color: 'var(--text-secondary)' }}>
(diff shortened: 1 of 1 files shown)


JiRaska:
retrigger

Issue #2312, opened 2026-07-25
openbank-aml-service openapi.yaml describes an API that was never served


JiRaska:
## What

`openbank-aml-service/src/main/resources/openapi.yaml` described an API that was never served.

| Spec declared | Reality |
|---|---|
| `CreateAmlCaseRequest.triggerType`, `.riskScore`, `.notes` | no such identifiers anywhere in the service |
| `AmlCaseResponse.triggerType`, `.riskScore`, `.decision` | same |
| `UpdateDecisionRequest.decision`, `.analystId` | real DTO takes `targetStatus`, `decisionReason?`, `assignedAnalyst?`, `decidedBy` |
| `status` enum includes `CLOSED`, `SAR_FILED` | `AmlCaseStatus` is `OPEN/UNDER_REVIEW/CLEARED/BLOCKED/ESCALATED` |
| `servers: http://localhost:8106` | `application.yaml` binds `quarkus.http.port: 8117` |
| no `partyId` / `screeningType` query params | `AmlCaseResource.listCases` serves both |
| no `Idempotency-Key` header | `createCase` rejects a blank/absent value |
| `PUT /decision` returns an empty `200` | returns an `AmlCaseResponse` |
| `ApiError` is `{code, message}` | shared `com.openbank.libs.api.error.ApiError` is `{traceId, status, code, message, timestamp, details}` |

Ground truth: `AmlCaseDtos.kt`, `AmlCase.kt`, and the provider-verified pact
`pacts/openbank-fx-service-openbank-aml-service.json`.

Found while working #2255, deliberately kept out of #2284 (a `test`-typed PR).

## Also in the same file

`openbank-aml-service/governance.yaml` `lineage.upstream` listed three services that do not touch
this one — `swift-service` (AML appears only in its prose docs), `kyc-service` (publishes
`openbank.kyc.events`) and `sanctions-service` (publishes `openbank.sanctions.screening.event`);
aml consumes neither topic. It omitted the two real API callers, `fx-service` and `agent-service`
(both `@RegisterRestClient(configKey = "aml-service")`), and the one topic it does consume,
`openbank.party.events` from `party-service`.

## Residual

Four true constraints could not be expressed and ship as `GATE-CONSTRAINED` `description` text
instead — see the linked gate issue. This issue is closed by the correction; the residual is not.

PR #2366, opened 2026-07-25
fix(ci): stop the api-contract gate demanding a MAJOR bump it also forbids

### LINKED ISSUE #2313: api-contract gate is unsatisfiable for a spec correction — and forces the false value to be kept

JiRaska:
## Summary

The api-contract gate and the ADR-0048 D2 invariant were **jointly unsatisfiable** for a spec that
never described the running service. Correcting one was red at the old version (oasdiff calls the
diff breaking, a MAJOR is demanded) and red at MAJOR+1 (D2 rejects a major the served URL has not
moved to). Both ends were measured on aml-service with the pinned oasdiff 1.22.0 — there is no
third value.

Worse than blocking the fix, **it required the falsehood**. The only green document aml could ship
kept the phantom `CLOSED`/`SAR_FILED` status values and a `limit` default of 20 against a server
that uses 50, plus four `GATE-CONSTRAINED` concessions.

## Type of change

- [x] fix (bug fix)

## Linked issues

Closes #2313
Refs #2312, #2314, #2360

## Changes

- `.github/scripts/check-api-contract.py` — a fourth class, `correction`: a breaking **document**
  diff in a PR that changes nothing else in the service requires MINOR, not MAJOR. Emits a
  `::notice` naming the count and the first breaking change, so a reclassification is never silent.
- `openbank-infra/scripts/check_api_contract_test.py` — 13 unit tests, picked up by the existing
  `governance-script unit tests` step.
- `openbank-libs/governance/rules.yaml` — the rule and its reasoning.
- 66 gitops files — the usual `rules.yaml` ripple, generated, verified idempotent.

## Reviewer notes — why this is not an escape hatch

I argued against a declared hatch when I raised #2313 and still do: **the discriminator is the diff,
not an assertion.** No marker file, no PR-body token claiming a spec "was never served". ~22 specs
are correction candidates (#2314), so anything a human asserts once becomes the default path
twenty-two times over.

Why it is safe: a spec-only PR leaves the server byte-identical, so no client that works today can
stop working. What "breaks" is a client generated from a document that never described this server
— already broken before the edit. Touch **any** other file in the service and the breaking rule
applies unchanged; only `CHANGELOG.md` and `version.txt` are exempt, because release-please writes
them and neither changes what is served.

It opens no new hole: landing a breaking change as code in one PR and the spec in another already
escapes this gate, which is scoped to spec diffs.

**Proven on the real case, both directions**, rather than in the abstract — the truthful aml spec
(the one I could not ship this morning) restored locally:

| scenario | verdict |
|---|---|
| spec-only | `reclassifying 10 breaking document change(s) … as a CORRECTION`, then `correction change, 1.1.0 -> 1.2.0 — OK`, **exit 0** |
| identical spec + one line appended to `AmlCase.kt` | `breaking API change requires an info.version MAJOR bump`, **exit 1** |

The second is the one that matters. A reclassification that fires too eagerly disables the gate for
genuine breaking changes, so the unit tests are weighted to the negative cases: a Kotlin change
disqualifies, an `application.yaml` change disqualifies, a **test-source** change disqualifies, and
a service-name prefix is not a substring match (`openbank-aml-service-extras/` must not count as
`openbank-aml-service/`). `bump_satisfied("breaking", …)` is asserted to still reject a MINOR.

The probe commits were removed before pushing; `git diff origin/main -- openbank-aml-service` is
empty, so this PR changes only the gate.

## What it still does not check

Whether the corrected document is **true**. #2360 covers the route set; nothing yet covers schemas
(#2314). This change makes a truthful correction *possible* — it does not verify that a given
correction is truthful.

## Follow-up

The six `GATE-CONSTRAINED` concessions in aml (#2317) and copilot (#2353) become removable once this
lands. Separate PR, which will demonstrate the fix in the wild.

## Definition of Done

- [x] Unit tests added — 13, negative-weighted.
- [x] `python3 -m unittest discover -s openbank-infra/scripts -p '*_test.py'` → **98 tests, OK**.
- [x] Full local gate sweep → 34 scripts, 0 non-zero.

## Security checklist

- [x] No new dependency (stdlib only).
- [x] The change makes a merge gate *more* permissive in one narrow, mechanically-determined case;
      the negative tests above are the control on that.

## Compliance impact

- [x] None

## Rollout / Rollback

- Deployment strategy: OPA bundle annotations change, so pods roll — unavoidable for any
  `rules.yaml` edit.
- Rollback plan: revert and re-run the generators.

**Short shelf life** — any competing `rules.yaml`/`.rego` PR regenerates the same bundles and
conflicts this one. Merge with `--auto`; on conflict take `main`'s bundles wholesale and re-run the
generators rather than hand-resolving a checksum.

### DIFF
--- a/.github/scripts/check-api-contract.py
+++ b/.github/scripts/check-api-contract.py
@@ -4,9 +4,20 @@
 For every `openbank-*/src/main/resources/openapi.yaml` changed against the PR
 base, classify the API-contract change from the OpenAPI diff (oasdiff):
 
-    breaking  => info.version MAJOR must move (new URL major, /api/v{N+1})
-    additive  => info.version MINOR (or MAJOR) must move within the same /v{N}
-    editorial => info.version PATCH (or higher) must move
+    breaking   => info.version MAJOR must move (new URL major, /api/v{N+1})
+    correction => a breaking DOCUMENT diff in a PR that changes nothing else in the service:
+                  the served contract is unchanged, so MINOR (see below)
+    additive   => info.version MINOR (or MAJOR) must move within the same /v{N}
+    editorial  => info.version PATCH (or higher) must move
+
+The `correction` class exists because `breaking` and the D2 invariant below are jointly
+unsatisfiable for a spec that never described the running service. Correcting one is red at
+the old version (oasdiff calls it breaking, a MAJOR is demanded) and red at MAJOR+1 (D2 rejects
+it, the served URL major has not moved). Measured both ends on aml-service, #2312/#2313: the
+only green document was one that kept a phantom enum value and a wrong default, so the gate was
+not merely blocking the fix, it was requiring the falsehood. A spec-only PR cannot break a
+working client — the server is byte-identical — so the MAJOR requirement is dropped to MINOR.
+The discriminator is the diff itself, never a declared marker.
 
 and assert the ADR-0048 D2 API invariant on the head revision:
 
@@ -173,17 +184,30 @@ def oasdiff_classify(oasdiff: str, old: Path, new: Path) -> tuple[str, list[str]
     return "none", []
 
 
-REQUIRED_BUMP = {"breaking": "MAJOR", "additive": "MINOR", "editorial": "PATCH"}
+REQUIRED_BUMP = {"breaking": "MAJOR", "correction": "MINOR", "additive": "MINOR", "editorial": "PATCH"}
 
 
 def bump_satisfied(kind: str, old_v: tuple[int, int, int], new_v: tuple[int, int, int]) -> bool:
     if kind == "breaking":
         return new_v[0] > old_v[0]
-    if kind == "additive":
+    if kind in ("additive", "correction"):
         return new_v[0] > old_v[0] or (new_v[0] == old_v[0] and new_v[1] > old_v[1])
     if kind == "editorial":
         return new_v > old_v
     return True
+
+
+# Derived files that cannot change what the service serves, so their presence in a diff does
+# not disqualify a spec correction. release-please writes both.
+BEHAVIOURLESS = {"CHANGELOG.md", "version.txt"}
+
+
+def service_touched_beyond_spec(service: str, spec_rel: str, changed_all: list[str]) -> list[str]:
+    """Files in this service the PR changed other than its openapi.yaml (and derived files)."""
+    return [
+        f for f in changed_all
+        if f.startswith(service + "/") and f != spec_rel and f.rsplit("/", 1)[-1] not in BEHAVIOURLESS
+    ]
 
 
 def main() -> int:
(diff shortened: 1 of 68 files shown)


JiRaska:
Correction to this PR's description: it said the `GATE-CONSTRAINED` concessions were in "aml (#2317) and copilot (#2353)". They were only ever in aml — `git grep -c GATE-CONSTRAINED` over every `openapi.yaml` on main returns exactly one file, six occurrences, all in `openbank-aml-service`. copilot's correction was fully additive (oasdiff ERR 0) and never hit this wall, so it needed no concession.

Field test of the fix is #2373: the same ten ERR-level changes, now reclassified and accepted at 1.1.0 -> 1.2.0.

Issue #2216, opened 2026-07-25
eu-ai-act-registry is not a required check — it went red twice on #2156 and the PR merged with a stale AI Act inventory


JiRaska:
## What happened

PR #2156 added the `ap2-anonymous` charter without regenerating `docs/compliance/eu-ai-act.md`. The `eu-ai-act-registry` workflow **went red on that PR twice**, and the PR **merged anyway** — the workflow is not in `main-protection`'s required status checks.

Result: for the period between that merge and PR #2211, the repo's EU AI Act inventory **omitted an AI system**. The document that exists specifically to enumerate every AI system in the platform was not enumerating one, and the check designed to prevent exactly that had already said so, twice, into a void.

The document itself is now regenerated and correct (verified on `main` after #2211). This issue is about the enforcement gap, not the drift.

## Why this one deserves the required list

Most advisory checks are advisory for good reason. This one is different: with the AI Act deadline of **2026-08-02**, `docs/compliance/eu-ai-act.md` is a regulatory artifact, and it is **generated** — so a red `eu-ai-act-registry` does not mean "someone should look", it means "the committed compliance document does not match reality". There is no judgement call in it.

Currently required: `all-green`, `Validate manifests`, `Gitleaks`, `issue-hygiene`, `OPA policy gate`.

## Suggested

Add `eu-ai-act-registry` to the required status checks, or fold it into `all-green` if that is the established pattern for gates of this kind. Then verify by opening a throwaway PR that deliberately drifts the doc and confirming it cannot merge — a required check nobody has watched fail is the same trap as the gates fixed in #2165 and #2154 this week.

Worth a wider sweep at the same time: which other **generated-artifact** gates are advisory and would let a derived compliance or governance document merge stale? `agent-charter-registry` and `Public-readiness audit` are in the same family.

Refs #1918, PR #2211, PR #2156, #2183.

JiRaska:
Still valid, and there is an open PR that targets exactly this: **PR #2229** (`ci(governance): enforce the EU AI Act + agent-charter drift gates in the required job`) — OPEN as of this pass, not merged. No further scoping needed here; this issue closes when #2229 lands and its failure path has been exercised (a deliberately-drifted doc that cannot merge), per the issue's own acceptance note.

Engineering-ready, blocked only on reviewing/merging #2229.

— automated triage pass

JiRaska:
[...]

Issue #1331, opened 2026-07-16
ADR-0176 force 3 / D6 rest on a false premise; the party-service vs consent-service split-brain needs deciding


JiRaska:
ADR-0176 (#1181) merged with a false premise in force 3 and D6. I wrote it; this is the correction.

**Revised 2026-07-16** after checking every claim against `origin/main` and primary legal sources. The first version of this issue overstated the case in two ways and got one attribution backwards — both corrected below, and called out at the end, because the errors are instructive.

## 1. The premise is false — marketing consent exists and is live ✅ verified

ADR-0176 force 3 states: *"There is no marketing consent to check, only the appearance of one"*, and D6 lists as a precondition *"`marketing_consent` mapped in party-service"*.

There are **two** columns, and the ADR cites the dead one:

| Column | Origin | Status |
|---|---|---|
| `marketing_consent` | `party-service` `V2__compliance_fields.sql:11` | **dead legacy** — `grep` for readers/writers across `party-service/src` returns **0**. The ADR's claim is true *of this column only*. |
| `consent_marketing` | `V12__consent_capture.sql` (#1157) | **live + mapped** — `PartyEntity.kt:90` |
| `consent_marketing_updated_at` | `V13__marketing_consent_updated_at.sql` (#1161) | **live + mapped** — `PartyEntity.kt:96` |

The live one is wired end to end: `PartyEntity` → `PartyRepositoryImpl` (read + write) → `PartyResource` `@PATCH updateConsent` → `UpdateMarketingConsentCommand`, emitting an ADR-0086 audit event; customer-facing revocation via `PATCH /customer/v1/profile/consent` in `openbank-customer-edge`, party-scoped from the JWT.

Timing (both `+02:00`, no timezone trap):

```
dd1d75719 2026-07-16T09:16:15+02:00  feat(party-service,customer-edge): revocable marketing consent (Profile screen) (#1161)
e97e1c7f4 2026-07-16T16:33:34+02:00  docs(adr): add ADR-0176 operator-initiated customer messaging (#1181)
```

The ADR claims that work is *"already in flight"* and that D6 *"does not front-run it"*. It had merged **7 h 17 min earlier**.

[...]

Issue #266, opened 2026-07-05
Roll out OPA enforce mode across the 14 remaining money-path services (ADR-0034 Phase 5)


JiRaska:
## Context

ADR-0034 (unified OPA authorization) is Delivery-Status: Partial. Phase 5 (money-path rollout) has not started (verified while auditing ADR delivery status, 2026-07-05).

## Remaining scope

- [ ] **Phase 5 — money-path enforcement**: 14 money-path services (account, balance, ledger, transaction, sepa-payment, sepa-instant, domestic-payment, clearing-service, swift-service, fx-service, lending-service, sca-service, consent-service, fraud-service, billing-service — see `rules.yaml: money_path_services`) are still running OPA in **advisory** mode, not enforce. Each flip needs 2 approvals + a threat-model update per repo policy. (Note: consent-service specifically is now tracked separately in #263 — ADR-0126 D5.)
- [ ] **D2 — MCP enforcement sidecar**: still advisory; PR #638 open (verify status — may be stale given the repo's public-transition history).
- [ ] **D4 — k8s sidecar deployment**: not yet implemented fleet-wide.

## References
- ADR-0034 (unified OPA authorization)
- Issue #263 (consent-service AUTHZ_ENFORCE flip, ADR-0126 D5)

JiRaska:
## Static coverage analysis — why Phase 5 must not be a blind AUTHZ_ENFORCE sweep

Ran the new `authz-coverage-report.py` (PR #313) against the current fleet. Money-path summary: **22 covered · 25 partial · 9 uncovered** `@Authorize` declarations.

Two structural blockers in `rest.rego` itself (not per-service validation):

1. **No M2M path to money-path endpoints.** A `SERVICE` principal only ever matches `edge-service-notification` (the `notification./device.` families). Every service-to-service call to a money-path write endpoint (e.g. `ledger.create` called by transaction-service) would 403 under enforce. **Proposal:** add a `service-to-service` allow reason whose (caller, callee-action-family) allowlist is derived from the same declared gitops edges `gen-network-policies.py` reads — one source of truth for both the network layer and the authz layer.
2. **Non-read actions with an empty `resource` have no allow path at all** (9 today, e.g. `transaction.reverse`, `ledger.trigger`): `operator-on-own-tenant` requires `input.resource`. Either these endpoints gain a resource binding, or the policy gains an explicit operator rule for non-scoped money-path verbs (four-eyes-gated).

Suggested order:
[...]

PR #2367, opened 2026-07-25
docs(adr): add ADR-0198..0203 for consent, CRM, campaigns and business-plane agents

JiRaska:
## Summary

Six **proposed** ADRs (0198–0203) for the CRM / campaign / agent-collaboration track. Documentation
only — no code, no service, no migration. ADR-0198 records a decision the repo left explicitly open:
issue #1331 closed `COMPLETED` with no comment, while ADR-0176 D6 still reads *"Blocked on the #1331
consent-authority decision"*. The analysis landed; the decision never did.

## Type of change

- [x] docs

## Linked issues

Refs #1331 · Refs #1182 · Refs #2206 · Refs #2216

## Changes

- **ADR-0198 — Marketing consent as a first-class consent-service scope.** Per-channel
  `MARKETING_COMMS_{EMAIL,PUSH,INAPP}` scopes following the `TELEMETRY_RUM` precedent (Art. 7, not
  SCA-gated, 365-day bucket); party-service's `consent_marketing` becomes a projection, the dead V2
  column is dropped in a new migration, and `notification_preferences.marketing_push` is demoted from
  consent to per-channel mute.
- **ADR-0199 — Customer 360 read model in a new crm-service.** Owns no facts; every column derived
  from an existing outbox topic, so ADR-0118 erasure is a projection replay rather than a
  cross-table delete sweep. Closes BIAN rows 11 and 20.
- **ADR-0200 — Campaign journeys as Temporal workflows with consent-gated delivery.** One workflow per
  enrolment; consent re-checked before every send *and* `consent.revoked` delivered as a workflow
  signal, so a journey stops mid-flight instead of at the next batch. Takes up the campaign tool
  ADR-0176 deferred to "its own ADR", keeping its catalogue discipline unchanged.
- **ADR-0201 — Segmentation and next-best-action on the ML decisioning platform.** Second consumer of
  ADR-0139/0140, which is what turns its offline half from a phase-2 intention into a dependency. NBA
  ranks communications and is *barred* from credit outcomes, keeping it outside AI Act Annex III.5(b).
- **ADR-0202 — Agent-to-agent collaboration over proposal events, MCP and Temporal.** An
  `agent.proposal.v1` topic with a closed PII-free envelope, existing MCP read tools, and per-chain
  Temporal orchestration — no A2A framework, because one would become a second control plane beside
  `agents.yaml`.
- **ADR-0203 — Business-plane AI agents.** Six agents, `plane: business`, read-only with a human
  disposition. `fraud-triage-agent` first (its analyst queue already exists); `collections-agent`
  deferred as Annex III.5(b).
- Regenerated the derived `docs/adr/{README.md,DIGEST.md,index.json}`.

## Definition of Done

- [x] Hexagonal layering preserved — no code changed.
- [x] Docs updated (this PR *is* the ADR change).
- [x] `bash docs/adr/gen-index.sh && bash .github/scripts/check-adr-registry.sh` — validator exit 0,
      derived files regenerated and clean after the rebase.
- [x] `.github/scripts/check-license-headers.py` exit 0.
- [x] All internal ADR links resolve (probe validated against a deliberately broken link first, so its
      silence means something).
- [ ] Unit / integration / contract tests — n/a, no code.
- [ ] OpenAPI, Flyway, Avro — n/a, no API, DB or event change in this PR.

## Security checklist

- [x] No new `@Suppress` / `as Any`.
- [ ] Auth / crypto / payment code changed — no code changed. ADR-0202 D3 *documents* the existing
      `principal.id` rule; it does not alter any policy.
- [x] **PII path changed?** No code, but ADR-0199 and ADR-0203 propose new PII processing, and
      ADR-0198 concerns a GDPR Art. 7 consent → suggest `gdpr-review-required`.
- [ ] Cardholder data path changed — no.

## Compliance impact

- [x] GDPR
- [x] DORA
- [ ] PCI DSS · [ ] PSD2 / SCA / RTS · [ ] AML / 5AMLD · [ ] CNB reporting

These are *proposed* decisions, so nothing changes today. What the ADRs record:

**A live gap worth reviewing on its own, independently of whether these ADRs are accepted.**
`NotificationConsumer` gates only the PUSH channel, and only on `marketing_push BOOLEAN NOT NULL
DEFAULT TRUE` — an opt-out, which cannot satisfy Art. 7(1) demonstrability. The EMAIL branch of the
same `when (req.channel)` calls `sendEmail` with no category, preference or consent check at all, so a
`MARKETING`-category email would be delivered with zero consent evaluation anywhere in the path. It is
latent only because the ADR-0176 catalogue has no marketing template — and ADR-0200 is what would add
one. Separately, `admin-ui/src/app/docs/compliance/page.tsx:68` reports marketing consent "ok" citing
the dead V2 column that #1331 verified has zero readers and zero writers.

**DORA:** ADR-0199 and ADR-0200 both reject a commercial vendor (CRM / campaign platform) on the
ADR-0174 exit-strategy and ADR-0175 residency grounds; ADR-0202 rejects an external A2A framework on
the same basis. Those are ICT third-party positions for the register.

**EU AI Act:** ADR-0201's `campaign-next-best-action` and each ADR-0203 agent need an
`ml-systems.yaml` entry, *generated* rather than hand-edited — #2216 recorded that
`eu-ai-act-registry` has gone red on a PR that merged anyway, and an advisory check over a generated
artifact leaves drift mergeable.

## Rollout / Rollback

- Deployment strategy: n/a — documentation only, nothing deployed.
- Rollback plan: revert the commit; the derived index regenerates from the remaining ADRs.
- Data migration reversible: N/A in this PR. ADR-0198 itself carries a rollback note — the projection
  columns are additive and the forwarder is flag-guarded, but the backfill must run *before* the write
  path flips or every opted-in customer silently reverts to denied.

## Reviewer notes

**Read ADR-0198 first.** The other five depend on it, and it is the only one that resolves an
already-open question rather than proposing new capability.

**Blocking dependencies are stated as blocking, not as caveats.** ADR-0200 cannot ship before
ADR-0198, before `IN_APP` becomes real delivery (today `log.infof("IN_APP stub")`, row stays
`PENDING`), and before #1182 for any content-bearing push. ADR-0202 D2 and all of ADR-0203 are blocked
on ADR-0195 / #2206, because mcp-service's `resolveContext()` returns a hardcoded constant — shipping
agent collaboration on top of that would authorize every agent as one identity, which is worse than no
collaboration.

**Two places where the reuse claim is deliberately weaker than it first looks**, because overstating
them would be the more expensive error:
- ADR-0201: only `OnlineFeatureStore` exists, the catalogue is two velocity features, the offline
  half of ADR-0140 is unbuilt (the only offline materialisation in the tree is `FeatureParityIT`), and
  `onnxruntime` is declared in `openbank-fraud-service/build.gradle.kts` alone. "Reuse ONNX serving"
  therefore means *promoting* it into `openbank-libs-runtime`, and that puts a native library into every
  service's dependency graph — a real cost, accepted as the lesser one against duplication.
- ADR-0202: `openbank-libs*` contains no `io.temporal` import; the wiring is duplicated per service as
  `infrastructure/temporal/TemporalClientProducer.kt`. Reuse means copying that pattern.

**ADR numbers 0198–0203 were free at push time** — checked against `origin/main` (only 0197) and every
open PR (0063, 0071, 0196). If a competing ADR PR merges first it takes the number and this branch must
be renumbered; that is the #481-vs-#524 race, and a cleanly-merging identical version line is the trap.

**On `docs` and release-please:** `docs` is a hidden type, so this PR releases nothing, which is
correct — it changes no shipped code.

Three claims in ADR-0198's compliance section (GDPR Art. 7/21, and Act 480/2004 §7's email-only soft
opt-in) are an engineering reading of primary sources carried over from #1331, not legal advice. The
lawful-basis classification should be confirmed with counsel before it is encoded in a ROPA.


JiRaska:
The live gap this PR's compliance section describes is now tracked on its own as #2369, so it can be assessed and fixed independently of whether these ADRs are accepted.

Two things sharpened while writing it up, both verified against `origin/main` @ `0d669339`:

- `NotificationTemplate.category` is an exhaustive `when` over the template enum with **no `MARKETING` branch**, so `NotificationCategory.MARKETING` is currently unreachable. That is what keeps the gap latent — and it means the `MARKETING` branch inside `maybeSendPush` is dead code today, i.e. a control that has never once executed.
- The compiler helps with the wrong half. Adding the first marketing template *is* forced to touch the exhaustive `category` getter, but `when (req.channel)` is already exhaustive over channels, so line 165 silently accepts the new category and calls `sendEmail`. The guard rail points at the mapping and stays silent about the delivery path.

ADR-0198 D4 in this PR proposes gating **before** the channel dispatch rather than per branch, which is what stops the same asymmetry recurring for `IN_APP` and `SMS` once those stubs become real delivery.

JiRaska:
Two more issues split out of this PR's context, both verified against `origin/main` @ `0d669339`:

- **#2371** — GDPR Art. 20 portability is neither implemented nor decided. Art. 15 *is* shipped (`GET /api/v1/parties/{id}/gdpr-export`, aggregating kyc + card-issuance, audited `gdprArticle = "15"`), and ADR-0118 mentions Art. 20 zero times. Relevant to ADR-0199, which cites ADR-0118's erasure model: the same ADR is silent on portability, and a Customer 360 read model is a natural place for that question to resurface.
- **#2372** — `SMS` and `IN_APP` are log-only stubs that return success. This is the blocking dependency ADR-0200 D7 already names. Worth noting for review: `IN_APP` is closer to working than ADR-0176's consequence note implies — rows are persisted before dispatch and the inbox read path does not filter on status, so they already appear in the customer's inbox. What is missing is the terminal status transition and a wake signal, which is a smaller job than "become real delivery" suggests.

Neither changes anything in this PR; ADR-0200 D7 already states both as blocking.

JiRaska:
Ran six independent adversarial review agents against this PR, one per ADR (fresh context each, no knowledge of the authoring session), then re-verified every finding myself against `origin/main` before acting on it. Two findings required rewriting a decision, not just a citation fix:

- **ADR-0199 D4** ("erasure is a replay") — architecturally broken against this fleet's actual Kafka config: every `kafka-topic.yaml` sets `retention.ms: 604800000` (7 days), `cleanup.policy: delete`, not compacted. A full-history replay only works for events younger than a week. `openbank-analytics-sink`, the one existing analog, does NOT rely on replay either (per-row crypto-shredding). Rewrote D4 to the pattern every other ADR-0118 consumer already uses — direct anonymisation on the `PARTY_ERASED` event.
- **ADR-0203** cited a DRAFT, unpublished security advisory by GHSA ID and paraphrased its root cause — verified via the API (`state: draft, published_at: null`) before touching it. This violates this repo's own explicit convention against writing unpatched security specifics to tracked files. Scrubbed to an oblique reference. Also corrected D8: the `resolveContext()` claim was stale — mcp-service now resolves real OAuth tokens (verified against current `main`), the remaining gate is a metric reading zero, not code that no longer exists.

Four smaller corrections: ADR-0198's "checked by the compiler" claim was wrong (no exhaustive match exists; `AISP_SCOPES` is a `setOf` allow-list — D1 now says this must be a reviewed line, not an assumption). ADR-0200 described `IN_APP` as a stub needing a fix; it was fully removed from `NotificationChannel` by #2386 (merged today) — updated to "needs rebuilding," a bigger lift. ADR-0202's #266 citation implied that issue documents a deployed defect it doesn't — retargeted to the guard script's own comment, which distinguishes the two instances correctly. ADR-0201's D5 credit-decisioning boundary gained a concrete mechanism (port-type separation, tracked as its own follow-up) instead of resting on prose alone, and D8's deployed-status guidance was clarified using the registry's own precedent.

Every finding was re-verified independently before I acted on it — Kafka retention configs, the GHSA state via the API, mcp-service's current implementation, `AISP_SCOPES`, issue #266's actual content, and PR #2386's merge — none were taken on the review agents' word alone.

Issue #2372, opened 2026-07-25
notification-service: SMS and IN_APP are log-only stubs that report success — the enum advertises two channels that do not deliver, and no metric would show it


JiRaska:
## Summary

`NotificationChannel` declares four channels — `EMAIL, SMS, PUSH, IN_APP`. Two of them have no
implementation. In `NotificationConsumer`'s dispatch (lines 164–176) they are `log.infof` calls that
return `Uni.voidItem()`:

```kotlin
NotificationChannel.SMS -> {
    log.infof("SMS stub: to=%s template=%s", req.recipient, req.template)
    Uni.createFrom().voidItem()
}
NotificationChannel.PUSH -> maybeSendPush(req, subject, entity)
NotificationChannel.IN_APP -> {
    log.infof("IN_APP stub: to=%s template=%s", req.recipient, req.template)
    Uni.createFrom().voidItem()
}
```

No port, no adapter, no error, no status transition. The type system advertises a delivery capability
the service does not have.

## The two stubs are not equivalent — treat them separately

This matters for what "fix" means, and lumping them together would get one of them wrong.

**`IN_APP` mostly works by accident.** The row is persisted with `status = "PENDING"` *before* dispatch
(line 162), and the inbox read path does **not** filter on status — `NotificationRepository.pageByParty`
/ `findByPartyId` return every row for the party. So an `IN_APP` notification *is* visible in the
customer's inbox via `GET /api/v1/notifications` (proxied through customer-edge). What is actually
missing is (a) the terminal status transition, so the row sits `PENDING` forever, and (b) any wake
signal telling the app something arrived.

**`SMS` is a true silent no-op.** No port, no adapter, and unlike `IN_APP` there is no read surface — an
SMS is delivery-only. A caller gets a persisted row and nothing else, permanently.

## Why it is currently harmless, and why that is the trap

**Nothing requests either channel.** Fleet-wide, the only non-test occurrences of `NotificationChannel.SMS`
and `IN_APP` are the enum declaration, the two stub branches, and `OperatorMessageService`'s comment
explaining why it refuses them. (`git grep` also hits `release-steward`'s `ScanState.IN_APPLICATION` —
substring noise, not a caller.) So this is latent, exactly like the unreachable `MARKETING` category in
#2369.

The trap is that the enum is public and exhaustive. Any service can construct a `NotificationRequest`
with `channel = SMS`, and it **compiles, passes review** — "SMS is a supported channel, it is right there
[...]

PR #2379, opened 2026-07-25
security(sca): remove SMS_OTP — its only implementation logs the OTP instead of delivering it

JiRaska:
## Summary

`SMS_OTP` is removed from `ScaMethod`. Its only implementation, `LoggingNotificationSender.sendSmsOtp`,
has never sent an SMS to anyone — it writes the OTP to the application log and returns. There is no SMS
provider anywhere in the fleet. This closes it by subtraction: a capability that cannot be requested
cannot silently fail to deliver it, and cannot write a live SCA credential to a log sink either.

## Type of change

- [x] security (private until disclosed)
- [x] fix (bug fix)
- [x] breaking change

## Linked issues

None — found during the blast-radius check for #2372 (unrelated `notification-service` channel stub
cleanup). Not filed as a public issue per this repo's routing rule (security holes → private advisory
or, as here, same-PR removal); this PR is the disclosure.

## Changes

- `ScaMethod`: removed `SMS_OTP`. Remaining: `PUSH_NOTIFICATION`, `TOTP`, `BIOMETRIC`.
- `ScaService.initiate` / `verify`: TOTP keeps the OTP generate/store/verify path it already shared
  with SMS_OTP; the SMS-specific dispatch call is gone.
- `NotificationSender` / `NotificationDispatchGuard` / `LoggingNotificationSender`: `sendSmsOtp` removed
  fleet-wide — swept for every caller first, there is exactly one and it is deleted here.
- `openbank-sca-service/src/main/resources/docs/{01-overview,03-api}.{cs,en}.md`: `SMS_OTP` references
  removed from the method enum and the verify-request note.
- Tests: `NotificationDispatchGuardTest`'s SMS case removed; `ScaServiceTest`'s SMS-specific test
  renamed and rewritten for TOTP (same behavior — OTP generate/store, no external dispatch); every
  other `ScaMethod.SMS_OTP` occurrence used only as a generic "some OTP method" fixture switched to
  `ScaMethod.TOTP`.

`ADR-0021` is untouched — a historical decision record referencing SMS_OTP as it existed when written.

## Definition of Done

- [x] Hexagonal layering preserved — no domain framework imports added.
- [x] Unit tests updated — 62 affected tests pass
      (`ScaServiceTest` 50, `NotificationDispatchGuardTest` 1, `ScaChallengeTest` 11).
- [ ] Integration / `@QuarkusTest` classes (`ScaResourceAuthzTest`, `ScaOutboxClaimIT`,
      `ScaPactProviderVerificationTest`) — **could not run in this environment**: Gradle dependency
      verification fails on two Quarkus artifacts' checksums
      (`quarkus-arc-test-supplement-decorator`, `quarkus-resteasy-common-spi`) before any test executes.
      Confirmed pre-existing and unrelated to this change — reproduces identically on plain
      `origin/main` with this branch's changes stashed. Flagging for CI, which may have a working
      dependency-verification cache this sandbox does not.
- [x] `./gradlew :openbank-sca-service:detekt :openbank-sca-service:ktlintCheck` passes.
- [x] No new lint warnings.
- [x] Docs updated (the two `03-api`/`01-overview` doc pairs).
- [ ] OpenAPI spec — n/a; `sca-service`'s `openapi.yaml` does not document `ScaMethod` or
      `preferredMethod` at all (pre-existing gap, not touched here).

## Security checklist

- [x] No secrets, tokens, passwords, or PII added. (The point of this PR is removing a path that wrote
      one to a log.)
- [x] No new `@Suppress` / `as Any`.
- [x] **Auth / crypto / payment code changed** → tagging `security-review-required`. This touches the
      SCA challenge method enum directly.
- [ ] PII path — no.
- [ ] Cardholder data path — no.

## Compliance impact

- [ ] PCI DSS · [ ] DORA · [ ] GDPR · [x] PSD2 / SCA / RTS · [ ] AML / 5AMLD · [ ] CNB reporting

PSD2 RTS on SCA: this removes one of four SCA methods. `PUSH_NOTIFICATION` (default), `TOTP` and
`BIOMETRIC` are unaffected and remain independently spec-compliant knowledge/possession/inherence
factors per method. No customer-facing SCA capability is lost in practice, since `SMS_OTP` could never
complete for a real customer — this removes a method that was already non-functional, it does not
reduce the platform's genuine SCA coverage.

## Rollout / Rollback

- Deployment strategy: rolling, same as any other `sca-service` release.
- Rollback plan: revert the commit. No data migration, no persisted `SMS_OTP` state to reconcile
  (challenges are TTL'd; nothing durable references the removed enum value).
- Data migration reversible: N/A — no schema change.

## Reviewer notes

**This is the disclosure.** No public issue was filed per the repo's routing rule for security holes;
this PR's description is deliberately factual rather than a full writeup, since the vulnerable code is
deleted in the same change. Please confirm no customer-facing surface (mobile app, admin-ui) offers
`SMS_OTP` as a selectable method before merging — I found none (`preferredMethod` is not documented in
`openapi.yaml` at all, and no admin-ui or customer-edge selector references `SMS_OTP`), but that absence
is exactly the kind of claim worth a second pair of eyes on auth-adjacent code.

The dependency-verification failure blocking the `@QuarkusTest` classes is almost certainly a sandbox
artifact (stale local verification metadata cache), not a real supply-chain finding — but I did not
silence or work around it, and it should be confirmed clean in CI before merge rather than assumed.

### DIFF
--- a/openbank-sca-service/src/main/kotlin/com/openbank/sca/application/port/out/ScaPorts.kt
+++ b/openbank-sca-service/src/main/kotlin/com/openbank/sca/application/port/out/ScaPorts.kt
@@ -52,12 +52,10 @@ interface ScaIdempotencyStore {
     suspend fun save(key: String, challengeId: UUID, ttlSeconds: Long)
 }
 
-/** Outbound port for delivering SCA challenges to the user out-of-band (push / SMS). */
+/** Outbound port for delivering SCA challenges to the user out-of-band (push). */
 interface NotificationSender {
 
     suspend fun sendPushNotification(partyId: UUID, challengeId: UUID, message: String)
-
-    suspend fun sendSmsOtp(partyId: UUID, otp: String)
 }
 
 /** Durable store of device credentials enrolled to a party (ADR-0021). */
--- a/openbank-sca-service/src/main/kotlin/com/openbank/sca/application/usecase/NotificationDispatchGuard.kt
+++ b/openbank-sca-service/src/main/kotlin/com/openbank/sca/application/usecase/NotificationDispatchGuard.kt
@@ -20,11 +20,4 @@ class NotificationDispatchGuard(private val notificationSender: NotificationSend
     suspend fun sendPushNotification(partyId: UUID, challengeId: UUID, message: String) {
         notificationSender.sendPushNotification(partyId, challengeId, message)
     }
-
-    @Timeout(2000)
-    @Retry(maxRetries = 2, delay = 200, jitter = 100, retryOn = [Exception::class])
-    @CircuitBreaker(requestVolumeThreshold = 10, failureRatio = 0.5, delay = 5000, successThreshold = 2)
-    suspend fun sendSmsOtp(partyId: UUID, otp: String) {
-        notificationSender.sendSmsOtp(partyId, otp)
-    }
 }
--- a/openbank-sca-service/src/main/kotlin/com/openbank/sca/application/usecase/ScaService.kt
+++ b/openbank-sca-service/src/main/kotlin/com/openbank/sca/application/usecase/ScaService.kt
@@ -159,12 +159,9 @@ class ScaService(
         idempotencyStore.save(idempotencyKey, saved.id, minOf(ttlSeconds, idempotencyTtlSeconds))
 
         when (method) {
-            ScaMethod.SMS_OTP, ScaMethod.TOTP -> {
+            ScaMethod.TOTP -> {
                 val otp = otpGenerator.generate()
                 otpStore.store(saved.id, otp, ttlSeconds)
-                if (method == ScaMethod.SMS_OTP) {
-                    notificationDispatchGuard.sendSmsOtp(command.partyId, otp)
-                }
             }
             ScaMethod.PUSH_NOTIFICATION, ScaMethod.BIOMETRIC -> {
                 notificationDispatchGuard.sendPushNotification(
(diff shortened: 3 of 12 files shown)


JiRaska:
Independent adversarial review (fresh agent, no context from the authoring session) found the fix sound — completeness sweep clean (zero remaining `SMS_OTP`/`sendSmsOtp` outside the immutable ADR-0021), no shared-branch behavior change for TOTP/PUSH_NOTIFICATION/BIOMETRIC, test substitutions are honest rather than a hollow rename.

One real, low-severity gap worth recording rather than silently dropping: `ScaChallengeEntity.method` is `@Enumerated(EnumType.STRING)`. A challenge row persisted with `method='SMS_OTP'` by a pod running the OLD code, read by a pod running THIS code during a rolling deploy, fails `Enum.valueOf` on hydration. `ScaResource` has 13 specific `ExceptionMapper`s but no catch-all, so this surfaces as an unstyled 500 rather than a clean 400/401. The rollback note in this PR's description ("no persisted SMS_OTP state to reconcile — challenges are TTL'd") is not quite accurate: TTL makes the row logically irrelevant, not nonexistent, so a request racing the deploy inside the 300s TTL window can still hit a stale row.

Practical severity is low: `SMS_OTP` never delivered an OTP to a real customer (that is this PR's premise), so no legitimate in-flight challenge exists in that state — only a very recent test/probe request could be affected, for a window measured in seconds around the deploy. Not blocking, but flagging for the record before merge rather than treating the rollback note as fully accurate.

Issue #2392, opened 2026-07-25
A CI gate with no rules.yaml entry is invisible to the ADR-0144 deadline gate — 5 advisory gates ungoverned, 1 dangling citation


JiRaska:
## Problem

`.github/scripts/check-gate-graduation.sh` is the ADR-0144 enforcement mechanism: every `enforced: advisory|planned` rule in `rules.yaml` must carry a `target_enforce_date`, and CI fails once that date passes while the rule still does not block. It works — but it walks **`rules.yaml` and nothing else**. A CI gate that never got a `rules.yaml` entry is structurally invisible to it: no deadline, no graduation pressure, no failure when it stays advisory forever.

That is not hypothetical. Both ADR-0148 gates ran `continue-on-error: true` with no `rules.yaml` entry:

- `check-prompt-registry.py` — went red on `main` from `cffc5ff9` (#2321) and stayed red for a month while every PR passed. The unregistered file was `system.v2.md`, the prompt hardened specifically against injected remediation commands. Fixed and enforced in #2363.
- `check-evals-registry.py` — enforced in #2375.

Neither flip was triggered by any mechanism. Both happened because a person happened to run the script by hand. ADR-0144 exists precisely to make that not be the trigger.

## Scope — this is a class, not two gates

Advisory steps in `ci.yml` whose script name and rule key appear **nowhere** in `rules.yaml`, measured on `bc75095a`:

| gate | rules.yaml entry |
|---|---|
| `check-prompt-registry.py` (now enforced, #2363) | none |
| `check-evals-registry.py` (now enforced, #2375) | none |
| `run-evals.py` — "evals gate replay (ADR-0148, advisory)" | none |
| `check-dotted-mp-messaging-keys.sh` (#686, advisory) | none |
| `check-authz-pdp-parity.py` (#1797, advisory) | **dangling** — see below |

The rest of the advisory steps (`governance-lineage`, `threat-model-diff`, `release-scope-mismatch`, `openapi-route-conformance`, `db-migration`, `db-backup`, `event-schema-compat`) do have entries and are correctly under the deadline gate.

## Separate finding: a dangling citation

`.github/scripts/check-authz-pdp-parity.py:30` says:

> graduates to enforce only once they are cleared — see rules.yaml: authz_pdp_parity.target_enforce_date.

[...]

PR #2393, opened 2026-07-25
docs(governance): fix two gate scripts citing rules.yaml keys that do not exist (#2392)

JiRaska:
## What

Two gate scripts cite `rules.yaml` keys that do not exist. Comment-only change; no gate logic touched.

### 1. `check-authz-pdp-parity.py` — `authz_pdp_parity`

Cited twice: in the module docstring, and as the stated graduation deadline (*"see rules.yaml: authz_pdp_parity.target_enforce_date"*). No such key exists — `grep -n pdp openbank-libs/governance/rules.yaml` returns nothing, and it never has.

The effect is worse than a broken reference. The gate reads as governed by an ADR-0144 deadline, while `check-gate-graduation.sh` — which walks `rules.yaml` and nothing else — cannot see it at all. **A citation to a nonexistent deadline is indistinguishable from having one**, which is exactly how a gate stays advisory forever.

The docstring now says plainly that no rule exists, why that matters, and points at #2392 (the governance half: the whole class of CI gates with no `rules.yaml` entry).

Deliberately **not** fixed by adding the rule instead: that restamps every service's OPA bundle (37 of 38 generators embed `rules.yaml`) and needs a real enforce date, so it belongs in #2392's PR, not a comment fix.

### 2. `check-db-migration.py` — `change_classes.db_change`

Found while checking the first one. The rule exists, but the parent key is `change_requirements`, not `change_classes`. Corrected.

## Measured while here

The parity gate is not merely waiting on a deadline. On main:

```
check-authz-pdp-parity: 37 workload(s) checked; 2 enforce-without-PDP violation(s).
```

`finrep-service` and `onboarding-service` enforce authz by app-default with no `opa` sidecar in their workload. The docstring now names those two blockers instead of the vague "several known violations" it claimed before.

**What that does and does not mean.** My first draft of this PR said every `@Authorize` endpoint on them fails closed. That overstated it, and checking rather than assuming is the whole point: `grep -rn Authorize` over both services returns **nothing** — neither contains a single `@Authorize` annotation today, so nothing is failing closed right now. The violation is a **latent trap** (the first `@Authorize` added to either service bricks that endpoint), not a live outage. Corrected in the commit message and the docstring.

The adjacent question — both services expose real REST resources (`FinrepResource`, `CorepResource`, `OnboardingResource`) with no authorization annotation at all — is a separate finding and needs its own issue, not a comment PR. Note also that #1797, which tracked the enforce-without-PDP class across eight services, is **CLOSED** while these two still trip the gate.

## Verification

- Both scripts parse (`ast.parse`) and run.
- The parity gate reports the identical `37 workload(s) checked; 2 violation(s)` before and after — comment-only, as claimed.
- A scan of every `rules.yaml: <key>` citation across `.github/scripts` + `openbank-infra/scripts` now resolves to a real key. That scan is what surfaced finding #2, and what caught my own first draft: the explanatory text I added initially used the `rules.yaml: <key>` form and so flagged itself. Reworded — a future automated version of this check would otherwise trip on the very comment explaining the problem.

## Linked issues

Refs #2392

### DIFF
--- a/.github/scripts/check-authz-pdp-parity.py
+++ b/.github/scripts/check-authz-pdp-parity.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python3
-"""authz-enforce ⇒ OPA PDP sidecar parity guard (issue #1797, rules.yaml: authz_pdp_parity).
+"""authz-enforce ⇒ OPA PDP sidecar parity guard (issue #1797).
 
 WHY THIS EXISTS: eight live services enforced `@Authorize` (`authz.enforce` defaults true) while
 their gitops manifest declared no `opa` sidecar. AuthorizeInterceptor fails closed when the PDP at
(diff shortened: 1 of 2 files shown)


JiRaska:
The adjacent finding is now filed as #2394 (both services carry zero `@Authorize` annotations, which is *why* the enforce-without-PDP violation is latent rather than live).

Issue #2369, opened 2026-07-25
notification-service: a MARKETING-category EMAIL is delivered with no consent check at all (PUSH is gated only by a default-TRUE flag)


JiRaska:
## Summary

`NotificationConsumer` gates only the **PUSH** channel before delivery, and it gates it on a channel
preference that defaults to on. The **EMAIL** branch of the same dispatch has no category check, no
preference check and no consent check at all. A `MARKETING`-category email would therefore be
delivered with zero consent evaluation anywhere in the path.

This is **latent, not live** — see "Why nothing is broken today" below. It is filed now because the
thing that keeps it latent is about to be removed.

## Evidence

`openbank-notification-service/src/main/kotlin/com/openbank/notification/application/NotificationConsumer.kt`
(line numbers against `origin/main` @ `0d669339`):

```
164                when (req.channel) {
165                    NotificationChannel.EMAIL -> sendEmail(req, subject, body, entity)   # no gate
166                    NotificationChannel.SMS -> {  ...log stub... }
170                    NotificationChannel.PUSH -> maybeSendPush(req, subject, entity)      # gated
171                    NotificationChannel.IN_APP -> { ...log stub... }
```

`maybeSendPush` (line 310) is the only gate in the file, and what it consults is a preference row, not
a consent:

```kotlin
NotificationCategory.MARKETING -> pref?.marketingPush ?: true
```

backed by `V11__notification_preferences.sql`:

```sql
marketing_push BOOLEAN NOT NULL DEFAULT TRUE,
```

whose own comment states the intent plainly — *"A missing row means «all on», so no backfill is
needed"*. That is an opt-out. GDPR Art. 7(1) requires the controller to be able to **demonstrate**
consent, and the absence of a row demonstrates nothing.

Nothing in this service reads the marketing consent that actually exists. That consent is
party-service `consent_marketing` / `consent_marketing_updated_at` (`V12__consent_capture.sql` #1157,
`V13__marketing_consent_updated_at.sql` #1161) — live, mapped on `PartyEntity`, revocable by the
customer through `PATCH /customer/v1/profile/consent`, and audited. It is simply never consulted
before a send.

## Why nothing is broken today, and why that is fragile

`NotificationTemplate.category` (`domain/model/Notification.kt`) is an **exhaustive `when` over the
template enum with no `MARKETING` branch**:

```kotlin
val category: NotificationCategory
    get() = when (this) {
[...]

PR #2419, opened 2026-07-25
docs(adr): add ADR-0205, marketing consent forwarder shape and SCA-exempt activation

JiRaska:
## Summary

Adds ADR-0205: decides three gaps found while trying to build ADR-0198 D3 (party-service marketing
consent forwarder + projection). Each gap was found by reading the actual consent-service code, not
assumed. Documentation only.

## Type of change

- [x] docs

## Changes

- New ADR `docs/adr/0205-marketing-consent-forwarder-and-sca-exempt-activation.md`, `proposed`.
- **D1**: `GDPR_ONLY_SCOPES` allow-list — consents made entirely of GDPR-only scopes activate
  immediately, no SCA challenge. Retroactively fixes `TELEMETRY_RUM`'s identical, previously
  undiscovered gap (zero callers anywhere in the codebase). Mixed requests (GDPR-only + SCA-required
  scope) are rejected outright, not silently downgraded.
- **D2**: one consent aggregate per party covers `MARKETING_COMMS_EMAIL` + `MARKETING_COMMS_INAPP`
  together, matching the single `consent_marketing` boolean. `MARKETING_COMMS_PUSH` stays separate.
- **D3**: a fixed `granteeId` (`"party-service:marketing-comms"`) for this internal grant flow —
  no such convention existed before.
- **D4**: party-service tracks `(partyId, consentId, grantedAt)` locally, populated on
  `ConsentGranted`, consulted on `ConsentRevoked`/`ConsentExpired` — since the revoke event carries no
  `scopes` field, the consumer cannot otherwise know whether a revoked consent was ever a marketing one.
- **D5**: explicitly defers the forwarder implementation itself to its own PR, once D1–D4 exist.

## Definition of Done

- [x] `bash docs/adr/gen-index.sh && bash .github/scripts/check-adr-registry.sh` — exit 0.
- [x] `.github/scripts/check-license-headers.py` — exit 0.
- [x] All internal ADR links resolve.

## Compliance impact

- [x] GDPR — D1 is directly about not over-burdening an Art. 7 consent with an SCA ceremony it
      doesn't need.

## Reviewer notes

Found by trying to implement, not by design review alone: I started building the party-service
projection consumer (the safe-looking, additive-only slice of D3), and it turned out to be
unbuildable correctly without first deciding these three things — most importantly, that
consent-service's `activateConsent` has no SCA-exempt path at all today, which means ADR-0198 D1's
"not SCA-gated" claim doesn't hold against the actual activation mechanism. That's the load-bearing
finding; D2–D4 follow from it once D1 exists.

Next: implement D1 (consent-service) first — it's foundational and fixes the latent `TELEMETRY_RUM`
gap on its own — then D4 (party-service consumer), then the forwarder (D5) as a separate PR.

## Linked issues

Refs #2369


JiRaska:
Independent adversarial review confirmed all three factual claims against `origin/main`, including a corroborating detail I had not found myself: `openbank-infra/gitops/apps/rum-gateway.yaml` explicitly comments that the gateway is "inert until ... the TELEMETRY_RUM consent scope ... are wired" — an independent admission of the zero-callers gap this ADR found. Mixed-scope rejection confirmed to break no current caller.

Two gaps flagged, one already addressed while starting D1 implementation: a `check()` in `Consent.kt`'s companion `init` block now asserts `AISP_SCOPES`/`GDPR_ONLY_SCOPES` disjointness at class-load time rather than relying on code review, per the review's suggestion. The second — a shared `openbank-libs-domain` constant for the `"party-service:marketing-comms"` granteeId string, since both consent-service and party-service depend on that module per ADR-0122 — is worth doing when D4 (the party-service consumer) is built, since that is where both call sites will actually exist; noting it now so it is not forgotten.

PR #2423, opened 2026-07-25
feat(consent): add GDPR_ONLY_SCOPES auto-activation, no SCA challenge (ADR-0205 D1)

JiRaska:
## Summary

ADR-0205 D1: adds `GDPR_ONLY_SCOPES` allow-list to consent-service — a consent request made entirely
of GDPR Art. 7 scopes (`TELEMETRY_RUM`, `MARKETING_COMMS_EMAIL/PUSH/INAPP`) now activates immediately,
no SCA challenge. Fixes `TELEMETRY_RUM`'s identical, previously undiscovered gap as a side effect.

## Type of change

- [x] fix (bug fix — `TELEMETRY_RUM` could never actually activate before this)

## Linked issues

Refs ADR-0205 D1 (`docs/adr/0205-marketing-consent-forwarder-and-sca-exempt-activation.md`, PR #2419)

## Changes

- `Consent.kt`: `GDPR_ONLY_SCOPES` set, mirroring `AISP_SCOPES`. `init { check(...) }` in the
  companion object enforces the two sets stay disjoint at class-load time.
- `ConsentService.createConsent`: if `scopes.all { it in GDPR_ONLY_SCOPES }`, create directly in
  `ACTIVE` status and emit `ConsentGranted` immediately — no `PENDING_SCA`, no `activateConsent` call.
  A request mixing a GDPR-only scope with any other scope throws `ConsentMixedScopeException` rather
  than silently falling back to the SCA-gated path.
- 6 new tests across `ConsentTest`/`ConsentServiceTest`: auto-activate for a multi-scope GDPR-only
  request, auto-activate for a single scope, mixed-scope rejection (and that it never calls `save`),
  an unaffected-SCA-only-request control case, and the disjointness invariant asserted explicitly.
- Fixed one pre-existing test whose mock setup broke under the new branching: `createConsent clamps
  validTo to 365 days for non-AISP scopes` used `TELEMETRY_RUM`, which now routes through the new
  auto-activate path (a different `save()` overload than it mocked) — switched its scope to
  `AGENT_QUERY` to keep testing what it was actually about (the 365-day clamp on the PENDING_SCA path),
  now cleanly separated from the new auto-activate path's own dedicated tests.

## Definition of Done

- [x] Hexagonal layering preserved — domain (`Consent.kt`) and application (`ConsentService.kt`) only.
- [x] Unit tests added/updated — `ConsentServiceTest` 29/29, `ConsentTest` 21/21, all green.
- [x] `./gradlew :openbank-consent-service:detekt :openbank-consent-service:ktlintCheck` passes.
- [x] No new lint warnings.
- [ ] OpenAPI/Flyway — n/a, no API or schema change.

## Compliance impact

- [x] GDPR — Art. 7: an SCA ceremony designed for payment authorization is a disproportionate burden
      on a data-processing opt-in; this is the fix.

## Reviewer notes

**A real regression in a pre-existing test was found and fixed in this PR, worth flagging explicitly
rather than letting it look incidental.** `createConsent clamps validTo to 365 days for non-AISP
scopes` used `TELEMETRY_RUM` as an arbitrary non-AISP example scope. Since `TELEMETRY_RUM` is now in
`GDPR_ONLY_SCOPES`, that test started hitting the new auto-activate branch and failing with a mockk
"no answer found" error (it only stubbed the single-arg `save()` overload; the new branch calls the
two-arg one). Fixed by switching the test's scope to `AGENT_QUERY`, which isolates the concern the
test was originally about. Confirmed this is the true cause, not flakiness or test-order interference,
by bisecting: the failure was 100% reproducible on the full suite and 100% absent running that one
test in isolation, consistent with a real mock-setup mismatch rather than shared state.

This PR is deliberately scoped to consent-service only (D1). Party-service's projection consumer (D4)
and the `UpdateMarketingConsentCommand` forwarder (D5) are separate PRs, per ADR-0205's own D5 note
that the forwarder is out of scope for the ADR itself.

### DIFF
--- a/openbank-consent-service/src/main/kotlin/com/openbank/consent/application/usecase/ConsentService.kt
+++ b/openbank-consent-service/src/main/kotlin/com/openbank/consent/application/usecase/ConsentService.kt
@@ -18,6 +18,7 @@ import com.openbank.consent.domain.event.ConsentGranted
 import com.openbank.consent.domain.event.ConsentRejected
 import com.openbank.consent.domain.event.ConsentRevoked
 import com.openbank.consent.domain.model.Consent
+import com.openbank.consent.domain.model.ConsentScope
 import com.openbank.consent.domain.model.ConsentStatus
 import com.openbank.consent.domain.model.ConsentValidationResult
 import jakarta.enterprise.context.ApplicationScoped
@@ -39,6 +40,11 @@ class ConsentScaChallengeMismatchException(id: UUID, scaSessionId: UUID) :
     RuntimeException("SCA challenge $scaSessionId does not match consent $id")
 class ConsentScaNotCompletedException(id: UUID, scaSessionId: UUID) :
     RuntimeException("SCA challenge $scaSessionId is not completed for consent $id")
+class ConsentMixedScopeException(scopes: Set<ConsentScope>) :
+    RuntimeException(
+        "Cannot mix GDPR-only scopes with SCA-required scopes in one consent request: $scopes. " +
+            "Request the GDPR-only and SCA-required scopes separately (ADR-0205 D1).",
+    )
 
 @ApplicationScoped
 class ConsentService(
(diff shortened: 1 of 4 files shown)


JiRaska:
Independent adversarial review confirmed all six checked claims — mixed/empty-scope handling correct (an empty scope set falls through to `Consent`'s own `require(scopes.isNotEmpty())`, not a new gap), `activateConsent`'s bypass skips only the SCA check (no other invariant was living there), the companion `init{}` disjointness guard fires reliably (Kotlin compiles it into the class's static initializer, forced by the first `AISP_SCOPES`/`GDPR_ONLY_SCOPES` reference `createConsent` already makes), and the "fixed a pre-existing test" claim verified true and non-scope-creeping.

Two follow-ups, both addressed in the latest commit:
- **Coverage gap**: no test proved `activateConsent` still rejects an already-`ACTIVE` GDPR-only consent. Added, plus a `scopes` parameter on the `consent()` test fixture (previously hardcoded to `ACCOUNTS_READ`) to make it possible.
- **Merge-order note**: ADR-0205 (PR #2419) was still open when this PR was opened. Merged #2419 first — decision before implementation, not the other way around. This PR's D1 content now matches an accepted, merged ADR rather than one still in review.

30/30 `ConsentServiceTest`, 21/21 `ConsentTest`, detekt + ktlint clean.

Issue #2425, opened 2026-07-25
Notification pact type-matches channel and template — a wrong channel or template passes verification


JiRaska:
## The weakness

`pacts/openbank-notification-service-openbank-account-service.json` type-matches **every** field of the `TRANSACTION_COMPLETED` notification request:

```json
"$.channel":            { "matchers": [{ "match": "type" }] },
"$.template":           { "matchers": [{ "match": "type" }] },
"$.recipient":          { "matchers": [{ "match": "type" }] },
"$.variables.amount":   { "matchers": [{ "match": "type" }] },
"$.variables.currency": { "matchers": [{ "match": "type" }] }
```

So the contract pins the **shape** and no value. Any string satisfies it.

## Why that matters for these two fields specifically

`channel` decides whether the customer gets a push, an SMS or an email. `template` decides which message they receive. Those are not free-form payload — they are closed vocabularies whose *value* is the whole point of the message.

Measured while adding account-service's `@PactFolder` replay (#2327): changing the producer to emit `"channel": "CARRIER_PIGEON"` leaves the provider verification **green**. Removing the `template` field entirely correctly goes red, so the replay works — it simply is not being asked anything about values.

Contrast with the sibling pact in the same sweep: `openbank-balance-service-openbank-account-service.json` leaves `eventType` with **no** matcher, so it is compared by exact value. That is the right treatment for a discriminator field, and it is already the convention one file over.

## What concretely must happen

In `NotificationRequestMessagePactConsumerTest` (openbank-notification-service), pin the closed-vocabulary fields by value rather than by type:

- `channel` — an exact value, or a regex over the permitted set if the consumer genuinely accepts several
- `template` — an exact value; a notification consumer that would accept any template string is not expressing a contract

Leave `recipient`, `variables.amount` and `variables.currency` as type matchers: those are genuinely free-form per message, and over-specifying them is the coupling Pact exists to avoid.

Then regenerate the pact (`./gradlew :openbank-notification-service:test --tests "*NotificationRequestMessagePactConsumerTest*"`) and commit it in the same PR — `pact-drift-check.yml` enforces that.

## Acceptance criteria

- Emitting a wrong `channel` or a wrong `template` from account-service's `@PactVerifyProvider` producer makes the provider replay **red**. Verify by doing it, not by reading the matchers: this issue exists because a green replay was mistaken for a checked contract.
- The other fields stay type-matched, and the pact still passes unchanged against the real producer.

## Scope note

[...]

PR #2424, opened 2026-07-25
test(balance): replay all three provider pacts before merge, not after

JiRaska:
## Summary

> **Fourth in a stack — merge order #2413 → #2417 → #2422 → this.** Base is `ci/pact-replay-party-service`.

`balance-service` is the provider for **three** committed pacts (six interactions, all HTTP) and its only verification class is `@PactBroker`-sourced and `@EnabledIfSystemProperty(pactbroker.url)`-gated — so it skips on every PR and those contracts were replayed only *after* the merge. Money path: two of the six are transaction-service reserving and releasing payment cover.

## Type of change

- [ ] feat (new functionality)
- [ ] fix (bug fix)
- [ ] refactor (no behavior change)
- [ ] perf (performance)
- [ ] docs
- [x] chore (build / tooling)
- [ ] security (private until disclosed)
- [ ] breaking change

Commit type is `test` — hidden, no release-please bump. No production code changed.

## Linked issues

Refs #2327
Refs #2269
Refs #2338

## Changes

- **New `BalancePactFolderProviderVerificationTest`** — `@PactFolder("../pacts")`, no gating annotation, same `@State` handlers, seeded rows and fixed UUIDs as the broker twin.
- **`KNOWN_UNCOVERED`** loses the three balance pacts: coverage **24 → 27**, backlog **9 → 6**.

## Reviewer notes

**Simpler than the previous two in the stack.** Every interaction is HTTP, so there is no `MessageTestTarget` and no `@PactVerifyProvider` producer — and therefore none of the duplicate-producer ambiguity #2417 and #2422 had to measure.

**Verified from the JUnit XML, not the console** — pact-jvm prints `Not all of the N were verified` on a fully green run:

```
tests=6 failures=0 errors=0
  transaction-service - POST placeHold to reserve funds for a payment
  transaction-service - DELETE releaseHold to release the payment cover
  account-service     - GET single CZK balance for an account
  account-service     - GET all balances for an account
  account-service     - POST initialize opening balance for an account
  mcp-service         - GET the balances the MCP get_balance tool returns verbatim
```

**Falsified rather than assumed:** renaming `BalanceResource`'s `@Path` to `/api/v1/balances-MOVED` turns **all six red** — the #2269 defect class, caught on contracts that would previously have skipped at PR time. Resource restored and the suite re-run green before committing.

[...]

Issue #2327, opened 2026-07-25
16 of 27 pacts have no provider replay before merge — the broker-gated verification skips on every PR


JiRaska:
## The gap

A consumer pact **cannot** catch a wrong request path — the Pact mock server answers whatever path the client asks for, so only the provider replay goes red. That is how finrep-service shipped a call to `/api/v1/ledger/trial-balance`, a ledger route that has never existed (#2269). The rule this repo drew from it is in `CLAUDE.md`: *never add a consumer pact without wiring (or confirming) the provider's `@PactFolder` replay.*

**16 of the 27 committed pacts have no provider replay that runs before merge.** Their only replay is a `@PactBroker` class gated on `@EnabledIfSystemProperty(named = "pactbroker.url", matches = ".+")`, and on a pull request that property is empty:

- `.github/workflows/_service-ci.yml:57` — PR lane runs on `ubuntu-latest` (the Pact Broker has no public ingress, ADR-0056)
- `.github/workflows/_service-ci.yml:72` — `PACT_BROKER_URL` resolves to `''` unless `github.event_name == 'push' && github.ref == 'refs/heads/main'`
- `.github/workflows/_service-ci.yml:439`, `:473` — the empty value is forwarded as `-Dpactbroker.url=""`, so the gate skips the class outright

So for these 16 the contract is replayed only *after* the merge, on the main-push lane. The defect class the gate exists to catch is exactly the one that gets through.

## Which pacts

Providers whose replay is broker-only (each row = the provider that needs the fix, with the pacts it leaves unreplayed pre-merge):

| provider | pacts with no pre-merge replay |
| --- | --- |
| `openbank-transaction-service` | account, domestic-payment, fraud, sepa-instant, sepa-payment (5) |
| `openbank-party-service` | account, kyc, onboarding (3) |
| `openbank-balance-service` | account, transaction (2) |
| `openbank-account-service` | balance, notification (2) |
| `openbank-sca-service` | consent (1) |
| `openbank-fraud-service` | sepa-payment (1) |
| `openbank-fx-service` | transaction (1) |
| `openbank-swift-service` | transaction (1) |

Several are money-path (`transaction`, `balance`, `fraud`, `fx`, `sca`, `swift`).

The other 11 pacts are fine — `openbank-ledger-service` (7), `openbank-kyc-service`, `openbank-aml-service`, `openbank-product-catalog`, `openbank-clearing-simulator` all replay unconditionally from `@PactFolder`.

[...]

Issue #2269, opened 2026-07-25
finrep-service calls a ledger trial-balance path that does not exist — every FINREP/COREP template render fails


JiRaska:
`openbank-finrep-service`'s only outbound client calls a ledger path that does not exist, so **every FINREP and COREP template render fails**. Found while writing the service's missing `openapi.yaml` and its first contract test (#2255, dimension C3 — finrep is the fleet's only C3=0).

## The defect

`openbank-finrep-service/src/main/kotlin/com/openbank/finrep/infrastructure/client/LedgerClient.kt`:

```kotlin
@RegisterRestClient(configKey = "ledger-service")
@Path("/api/v1/ledger")
interface LedgerRestClient {
    @GET
    @Path("/trial-balance")
    fun getTrialBalance(@QueryParam("asOf") asOf: String): Uni<TrialBalanceResponse>
}
```

Effective request path: **`GET /api/v1/ledger/trial-balance`**.

`openbank-ledger-service` exposes no such path. Its `@Path` roots are `/api/v1/control-accounts`, `/api/v1/fx`, `/api/v1/journals`, `/api/v1/journals/approvals`, `/api/v1/ledger/close`, `/api/v1/ledger/fx-revaluation`, and its committed `openapi.yaml` declares exactly two trial-balance paths:

- `/api/v1/journals/trial-balance` — the GL trial balance (`operationId: getTrialBalance`), which is what finrep wants
- `/api/v1/ledger/close/trial-balance` — the fiscal-year-close variant

`openbank-balance-service` calls the correct one, and its committed pact `pacts/openbank-balance-service-openbank-ledger-service.json` pins `/api/v1/journals/trial-balance`.

## Blast radius

`LedgerAdapter.getTrialBalance` is the sole implementation of `LedgerPort`, and both `FinrepResource` (`GET /api/v1/finrep/templates/{templateId}`) and `CorepResource` (`GET /api/v1/corep/templates/{templateId}`) derive every cell from it. F01.01 Balance Sheet, F02.00 P&L and C 01.00 Own Funds therefore all fail against a real ledger. The service is the ADR-0097 regulatory-reporting surface.

## Why nothing caught it

- finrep has **no `openapi.yaml`** and **no contract test** — it is the fleet's only service scoring C3=0. There is no consumer pact pinning the path it calls, so nothing compares the client against the provider's spec.
[...]

PR #2427, opened 2026-07-25
test(account): replay both provider pacts before merge, not after

JiRaska:
## Summary

> **Fifth in a stack — merge order #2413 → #2417 → #2422 → #2424 → this.** Base is `ci/pact-replay-balance-service`.

`account-service` is the provider for **two** committed pacts, both message-only, and its only verification class is `@PactBroker`-sourced and `@EnabledIfSystemProperty(pactbroker.url)`-gated — so it skips on every PR and both contracts were replayed only *after* the merge.

## Type of change

- [ ] fix (bug fix)
- [ ] perf (performance)
- [ ] docs
- [x] chore (build / tooling)
- [ ] breaking change

Commit type is `test` — hidden, no release-please bump. No production code changed.

## Linked issues

Refs #2327
Refs #2269
Refs #2425

## Changes

- **New `AccountPactFolderProviderVerificationTest`** — `@PactFolder("../pacts")`, no gating annotation, same `@State` handlers and `@PactVerifyProvider` producers as the broker twin.
- **`KNOWN_UNCOVERED`** loses both account pacts: coverage **27 → 29**, backlog **6 → 4**.

## Reviewer notes

**Cheapest one in this sweep by a wide margin.** Both pacts are message-only, so a `MessageTestTarget` alone is correct and nothing boots — the whole class runs in **3.2 s**. A plain JVM test, not another `@QuarkusTest` with Testcontainers. The FinOps objection I raised on #2424 does not apply here.

**Verified from the JUnit XML, not the console:**

```
tests=2 failures=0 errors=0  (3.2s)
  notification-service - a TRANSACTION_COMPLETED notification request
  balance-service      - an AccountCreated event
```

**Falsification took three attempts, and the two that did NOT redden are the interesting part** — each looked like a bug and only one was:

| break | result | reading |
| --- | --- | --- |
| `accountType` `CURRENT` → `SAVINGS` | green | **correct.** balance-service's pact carries only `aggregateId`, `currency` and `eventType`; `accountType` is outside the contract entirely |
| `channel` `PUSH` → `CARRIER_PIGEON` | green | **a real weakness.** Every field of the notification pact is a `type` matcher, so it pins shape and no value — including `template` (which message the customer gets) and `channel` (push vs SMS vs email). Filed as #2425; not fixed here, since it is the consumer's contract to tighten |
| remove the `template` field | **red** | as it must be — shape is what this pact pins, and the replay does bite on it |

Worth stating plainly: had I stopped at the first green, I would have reported this replay as falsified when it was not. The contrast is one file over — `openbank-balance-service-openbank-account-service.json` leaves `eventType` with **no** matcher, so it is compared by exact value, which is the right treatment for a discriminator.

**Additive, not a replacement.** Flipping the existing class to `@PactFolder` is exactly what #372 did and #1166 reverted: `can-i-deploy` reads the broker and nothing else, so account-service's consumers were blocked (notification-service #1180 and #1303 both merged clean, built green, then failed the gate). The broker twin is untouched.

## Definition of Done

- [x] Unit tests added/updated. — this PR is the test
- [x] Integration tests added/updated (if service boundary involved). — two provider-side replays
- [x] Contract tests updated (if public API changed). — no pact content changed; they are now replayed pre-merge
- [x] `./gradlew detekt ktlintCheck koverVerify build` passes. — ktlintCheck + compileTestKotlin + the test itself
- [x] No new lint warnings.
- [ ] OpenAPI spec regenerated (if REST API changed). — N/A
- [ ] Flyway migration added + rollback note (if DB changed). — N/A
- [ ] Avro schema versioned backward-compatibly (if event changed). — N/A
- [x] Docs updated (README, ADR if architectural). — KDoc

## Security checklist

- [ ] Auth / crypto / payment code changed? → N/A — test-only
- [ ] PII path changed? → N/A — synthetic pact fixtures
- [ ] Cardholder data path changed? → N/A

## Compliance impact

- [x] None

## Rollout / Rollback

- Deployment strategy: N/A — test-only, no service artifact
- Rollback plan: revert the commit; the broker twin is untouched, so `can-i-deploy` is unaffected either way

### DIFF
--- a/.github/scripts/check-pact-provider-replay.py
+++ b/.github/scripts/check-pact-provider-replay.py
@@ -49,12 +49,10 @@ PACTS = ROOT / "pacts"
 # ~44 generated files and give the PR a short shelf life. The list belongs next to the code that
 # reads it.
 KNOWN_UNCOVERED = {
-    "pacts/openbank-balance-service-openbank-account-service.json",
     "pacts/openbank-consent-service-openbank-sca-service.json",
     # Landed 2026-07-25, after the #2327 audit counted 16 of 27: same three providers
     # (balance/transaction/party), so the same debt, not a new class of it. The estate went
     # 27 -> 33 pacts in a day, which is the argument for a gate rather than another audit.
-    "pacts/openbank-notification-service-openbank-account-service.json",
     "pacts/openbank-sepa-payment-openbank-fraud-service.json",
     "pacts/openbank-transaction-service-openbank-fx-service.json",
     "pacts/openbank-transaction-service-openbank-swift-service.json",
--- a/openbank-account-service/src/test/kotlin/com/openbank/account/contract/AccountPactFolderProviderVerificationTest.kt
+++ b/openbank-account-service/src/test/kotlin/com/openbank/account/contract/AccountPactFolderProviderVerificationTest.kt
@@ -1,0 +1,117 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) OpenBank contributors. Licensed under the Apache License, Version 2.0.
+// See LICENSE in the repository root or https://www.apache.org/licenses/LICENSE-2.0 for details.
+
+package com.openbank.account.contract
+
+import au.com.dius.pact.provider.PactVerifyProvider
+import au.com.dius.pact.provider.junit5.MessageTestTarget
+import au.com.dius.pact.provider.junit5.PactVerificationContext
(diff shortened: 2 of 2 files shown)


JiRaska:
Superseded by https://github.com/JiRaska/open-bank-oss/pull/2446 — rebuilt on current `main`. This one was stacked on a branch that has since been rebased away, so it was CONFLICTING against a base that no longer matches. Identical content, verification and falsification carried over.

Issue #1393, opened 2026-07-17
OperatorMessageService.compose does 3 DB round-trips where 2 would do


JiRaska:
Code-review finding on PR #1368.

`OperatorMessageService.compose()` (`openbank-notification-service/src/main/kotlin/com/openbank/notification/application/OperatorMessageService.kt`) re-queries the full `NotificationEntity` by `notificationId` (SELECT + UPDATE) before flipping status to `SENT` or `FAILED`, instead of a scoped bulk update. Each terminal status transition costs a full-row SELECT (pulling subject/body HTML back out) plus an UPDATE.

`NotificationRepository.markRead`/`markAllRead` (same repository class, `NotificationRepository.kt:51-72`) already establishes the bulk `update("status = ?1 ... where notificationId = ?2", ...)` idiom this code doesn't follow — inconsistent with the file's own established pattern.

## Fix

Replace the find-then-map-then-persist pattern with a scoped bulk `update()` call for both the SENT and FAILED transitions, matching `markRead`'s existing shape. Low risk, no behavior change, cuts one DB round-trip per compose() call.

## Related

PR #1368

PR #2432, opened 2026-07-25
feat(party): project consent-service's marketing consent into consent_marketing (ADR-0205 D4)

JiRaska:
## Summary

ADR-0205 D4: party-service consumes consent-service's outbox (filtered to the fixed
`party-service:marketing-comms` granteeId, ADR-0205 D3) to project marketing-consent grants/
revokes/expiries into `parties.consent_marketing` / `consent_marketing_updated_at`.

## Type of change

## Linked issues

N/A — this PR implements ADR-0205 D4, decided in a merged ADR (`docs/adr/0205-marketing-consent-forwarder-and-sca-exempt-activation.md`), not tracked by a separate backlog issue.

## Changes

- **Migration V16**: `party_marketing_consent(party_id, consent_id, granted_at)` — tracks the
  currently ACTIVE marketing consent per party, one row per party.
- **`MarketingConsentEventConsumer`**: `@Incoming("consent-events-in")` on the whole
  `openbank.consent.events` topic, filters to `granteeId == "party-service:marketing-comms"`
  in-consumer (everything else — every TPP consent — is silently ignored), dispatches
  `ConsentGranted`/`ConsentRevoked`/`ConsentExpired` to the use case. Poison-pill safe.
- **`MarketingConsentProjectionService`**: `applyGranted` always upserts (a fresh grant wins);
  `applyRevokedOrExpired` only clears the projection if the event's `consentId` matches what's
  currently tracked — the mechanism that makes this correct despite `ConsentRevoked` carrying no
  `scopes` field (an out-of-order/late revoke for a superseded consent must not clear a fresh
  re-grant).
- **`PartyMarketingConsentEntity`**: surrogate `PanacheEntity` id + unique `partyId` column,
  matching `PartyEntity`'s own convention — not an application-assigned PK, so it doesn't hit the
  `persist()`-is-INSERT-only pitfall this repo's `CLAUDE.md` documents.
- **`PartyRepository.updateMarketingConsentProjection`**: a scoped `UPDATE ... WHERE partyId = ?`,
  matching `NotificationRepository.markTerminalStatus` (issue #1393) rather than a full
  find-then-mutate-then-save of the `Party` aggregate.
- **Kafka ACL**: `Read + Describe` on `openbank.consent.events` added to party-service's
  `KafkaUser` — required in the real cluster (Strimzi simple authorization denies by default
  per-topic); without it this consumer works in tests and is silently denied in production.

## Definition of Done

- [x] Hexagonal layering preserved — new port/usecase/infrastructure files follow the existing
      party-service structure exactly.
- [x] Unit tests added — 3 use-case tests (grant, matching revoke, **non-matching revoke — the
      load-bearing case**) + 8 consumer tests (dispatch per event type, grantee filtering,
      malformed-JSON and missing-field poison-pill safety). Full `openbank-party-service` suite
      green.
- [x] Integration tests — n/a; consumer/use-case logic is fully covered at the unit level, and
      wiring correctness (Kafka deserialization, migration) is exercised by the fleet's existing
      `@QuarkusTest` dev-services on `./gradlew build`.
- [x] `./gradlew :openbank-party-service:detekt :openbank-party-service:ktlintCheck` passes.
- [x] No new lint warnings — one pre-existing baseline entry duplicated for the new `port.in` file
      (package named `` `in` ``, a Kotlin keyword — the same structural violation already
      grandfathered for `PartyPort.kt`), not a new violation class.
- [x] Flyway migration added (V16) + rollback note in the migration file itself. Additive only
      (new table); rollback is a plain `DROP TABLE`/`DROP SEQUENCE`, safe because this table is a
      projection-tracking aid, not a source of truth — consent-service remains authoritative.

## Compliance impact

- [x] GDPR — this is infrastructure for demonstrable, revocable Art. 7 consent (ADR-0198/0205);
      no new PII is stored beyond a `consentId` and timestamp already implied by the consent's
      existence.

## Rollout / Rollback

- Deployment strategy: rolling. The migration is additive (new table only) and independent of the
  Kafka ACL grant — safe to deploy in either order, but the consumer will silently receive nothing
  until the ACL is applied (GitOps-managed, separate sync wave per `kafka-party-mtls.yaml`'s own
  annotations).
- Rollback plan: revert the commit; `DROP TABLE party_marketing_consent` per the migration's own
  rollback note if needed. No write-path change in this PR (D5 doesn't exist yet), so nothing
  currently produces the events this consumer would process — safe to roll back with zero data
  consequence.
- Data migration reversible: yes, additive-only.

## Reviewer notes

**Deliberately does not touch `UpdateMarketingConsentCommand` or any customer-facing write path.**
That is ADR-0205 D5, its own PR, once this (D4) exists to receive what it will eventually write.
Until D5 lands, this consumer has no producer — it is inert but harmless, exactly the state D1
(#2423) put `TELEMETRY_RUM`/`MARKETING_COMMS_*` scopes in before this PR gave party-service a
reason to care about them.

The `deleteIfMatches` atomic scoped DELETE (not find-then-check-then-delete) is the correctness
mechanism worth double-checking: a concurrent upsert (re-grant) racing this delete cannot lose the
race either way — the delete either matches the row that existed at delete time, or (if a re-grant
already changed `consentId`) matches zero rows and correctly no-ops.

### DIFF
--- a/openbank-infra/gitops/components/party/kafka-party-mtls.yaml
+++ b/openbank-infra/gitops/components/party/kafka-party-mtls.yaml
@@ -5,9 +5,10 @@
 # control; domain-event topics without ACLs stay open under allow.everyone.if.no.acl.found=true.
 #
 # party-service ACL surface:
-#   Write + Describe: openbank.party.events  (own domain event output + outbox relay)
-#   Read  + Describe: openbank.kyc.events    (KYC case-lifecycle gate, ADR-0073)
-#   Read  + Describe: openbank.aml.events    (AML case-lifecycle gate, ADR-0073)
+#   Write + Describe: openbank.party.events    (own domain event output + outbox relay)
+#   Read  + Describe: openbank.kyc.events      (KYC case-lifecycle gate, ADR-0073)
+#   Read  + Describe: openbank.aml.events      (AML case-lifecycle gate, ADR-0073)
+#   Read  + Describe: openbank.consent.events  (marketing-consent projection, ADR-0205 D4)
 #   Read:             group openbank-party-service
 #
 # Cert projection: Strimzi User Operator mints the keystore Secret in `messaging`;
@@ -47,6 +48,12 @@ spec:
       - resource:
           type: topic
           name: openbank.aml.events
+          patternType: literal
+        operations: [Read, Describe]
+        host: "*"
+      - resource:
+          type: topic
+          name: openbank.consent.events
           patternType: literal
         operations: [Read, Describe]
         host: "*"
--- a/openbank-party-service/ktlint-baseline.xml
+++ b/openbank-party-service/ktlint-baseline.xml
@@ -1,5 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <baseline version="1.0">
+    <file name="src/main/kotlin/com/openbank/party/application/port/in/MarketingConsentProjectionUseCase.kt">
+        <error line="5" column="9" source="standard:package-name" />
+    </file>
     <file name="src/main/kotlin/com/openbank/party/application/port/in/PartyPort.kt">
         <error line="5" column="9" source="standard:package-name" />
         <error line="8" column="1" source="standard:no-wildcard-imports" />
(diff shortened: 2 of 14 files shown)


JiRaska:
Independent adversarial review confirmed the load-bearing claim (item 1, the `deleteIfMatches` race protection) holds under direct code inspection — walked the exact out-of-order-revoke scenario against the real `WHERE partyId=? AND consentId=?` SQL and confirmed it correctly no-ops once a re-grant changes the tracked consentId. Entity design, JSON parsing safety, scope (confirmed zero touch on `UpdateMarketingConsentCommand`), and the Kafka ACL necessity (verified against `kafka-consent-mtls.yaml`'s existing Write+Describe grant — the topic already has an ACL, so it's enforced, not open-by-default) all confirmed against `origin/main` directly.

One real finding, fixed in the latest commit: the `occurredAt` fallback to `Instant.now()` was silent — no way to tell a real event timestamp from a fallback apart in `consent_marketing_updated_at`. Now logs a warning naming the payload, with a test confirming the event still applies on this path (it's a real path, not a poison-pill case).

Issue #2414, opened 2026-07-25
mcp-service: propose_payment / ProposalPort has no enforced PROPOSED-only state machine (T-E4)


JiRaska:
## Problem

`propose_payment` -- the money-adjacent tool in `openbank-mcp-service` -- has been a stub since the
service shipped (PR #2104) and is still a stub after the ADR-0195 phase-2 cutover (#2253/#2262/
#2278/#2316), which deliberately left `StubProposalPort` untouched. It returns a literal
`{"phase":"1-stub","status":"PROPOSED"}` note; no maker-checker row is written anywhere.

`agents.yaml`'s `mcp-anonymous` charter declares `requires_human: {every: proposal, sca:
dynamic_linking, scope: consent_granted}`. `agents.rego` says outright: "same as every charter's
`requires_human` block, which no code path reads either." So the propose-only guarantee -- the
claim that this service can never move money -- is currently a property of the stub, not of any
enforced state machine (T-E4 in `docs/threat-models/openbank-mcp-service.md`).

## What's needed

When `ProposalPort` is eventually bound to a real implementation (most likely copilot-service's own
`ActionProposal` domain, per the scope-cut recorded when #2262 investigated this -- copilot's
`ActionProposal` is internal, not a cross-service endpoint today, so this needs its own design, not
a drop-in client):

- A `PROPOSED`-only state machine that this service has structurally no transition out of --
  nothing in `openbank-mcp-service` should be able to reach a disposed/executed state directly.
- SCA dynamic-linking + human disposal enforced at the write boundary (wherever the proposal is
  actually acted on downstream), not merely asserted in `agents.yaml`.
- Re-evaluate `docs/threat-models/openbank-mcp-service.md` §7 (should this service join
  `rules.yaml: money_path_services`?) at that point -- the document's own conclusion names the
  `ProposalPort` binding as the trigger for that decision, explicitly deferred until now.
- Idempotency (T-D2) and server-side argument validation -- amount well-formed non-negative
  decimal, `toIban` a valid IBAN, `currency` ISO 4217 (T-T2) -- neither of which exists today because
  there's no real implementation to validate against yet.

## Why it matters

Nothing in the codebase structurally prevents a future `ProposalPort` implementation from calling
transaction-service directly and moving money on the model's word. The HITL + SCA guarantee this
service's entire risk profile rests on is asserted in YAML, not built into a state machine -- closing
that gap needs to be part of the design from the first line of the real port, not retrofitted after.

[...]

PR #2498, opened 2026-07-26
security(mcp): enforce PROPOSED-only on the call path, not in whichever ProposalPort is bound

JiRaska:
## Refs #2414 — the enforceable half of T-E4. Deliberately partial.

The triage on #2414 recommends keeping this as "a design precondition on whoever binds `ProposalPort`" because the acceptance items only become testable once a real port exists. That is right about the *cross-service* question and I have not touched it. But one item does not need the real port and gets strictly harder to add afterwards: making PROPOSED-only a property of the **call path** instead of a property of the implementation.

### The problem this closes

Right now `StubProposalPort` writes the literal string `PROPOSED` into its own response, and that string is the entire guarantee. So the ADR-0031/ADR-0181 promise — the model proposes, a human plus SCA disposes, money never moves on the model's word — is enforced by one class agreeing to enforce it about itself.

The day a real port is bound (a maker-checker row, or copilot-service's `ActionProposal`) that constant stops being the guarantee, and the failure mode is silent: a proposal that came back `EXECUTED` is serialized to the agent as an ordinary tool success, with nothing anywhere objecting. Adding the gate *now*, while the only binding is a stub, is when it is free.

### What it does

`ProposedOnly.enforce`, called at the `propose_payment` site in `McpToolRegistry`:

- **A whitelist of exactly one value** (`PROPOSED`), not a blacklist of forbidden statuses. A blacklist has to guess the vocabulary a future port will use, and whatever it fails to guess passes silently — which is the same class of defect as the thing being fixed. There is no state past PROPOSED a whitelist-of-one can accidentally admit.
- **Applied to every field named `status` at any depth**, nested objects and array elements included. The likeliest real shape is a wrapper that still says `PROPOSED` around a payment that has already been executed; a root-only check waves exactly that through.
- **A missing status is refused**, never assumed to mean PROPOSED.
- **A non-textual status is refused** rather than skipped, so a numeric or object status cannot slip past unexamined.
- **Fails closed.** A violation throws; `McpEndpoint` maps any throw from a tool to an audited `tool error`. The agent learns nothing about the state it must not reach, and the audit trail records the attempt.

### What it deliberately does NOT do

It does not decide how a real proposal is persisted, which service owns it, or how SCA dynamic linking binds to it. That is the architecture decision in #2414's body and it stays open and owner-gated — hence `Refs`, not `Closes`. The remaining acceptance items (idempotency, argument validation, the `money_path_services` re-evaluation) also still need the real port.

The guarantee this adds is narrower and unconditional: **whatever answers that question cannot hand an agent a disposed proposal without this gate going off first.**

One accepted consequence, called out so it is not a surprise: `propose_payment` echoes the caller's arguments back in its result, so an agent that sends a literal `status: "EXECUTED"` argument gets a `tool error`. Refusing is the conservative direction and the agent is only denying itself.

### Falsification

The registry test was run against the unfixed call site (`ProposedOnly.enforce` removed from `McpToolRegistry`):

```
ProposedOnlyTest > the registry refuses a port that returns a disposed proposal() FAILED
BUILD FAILED
```

The port's `{"status":"EXECUTED"}` was serialized straight back to the caller. Restoring the call makes it green. The direct `ProposedOnly` cases cannot be run against "unfixed code" in the same sense — the class is the fix — so the registry case is the one carrying the falsification, and it is the one that exercises the real path. The opposite direction is asserted too: `the shipped stub port satisfies the invariant`, so the gate cannot make the current binding unusable.

### Gate

`:openbank-mcp-service:detekt ktlintCheck test quarkusBuild` — BUILD SUCCESSFUL.

### DIFF
--- a/openbank-mcp-service/src/main/kotlin/com/openbank/mcp/application/McpToolRegistry.kt
+++ b/openbank-mcp-service/src/main/kotlin/com/openbank/mcp/application/McpToolRegistry.kt
@@ -105,7 +105,9 @@ class McpToolRegistry(
                     arguments.path("limit").asInt(DEFAULT_TX_LIMIT),
                 )
             "list_consents" -> accounts.listConsents(ctx)
-            "propose_payment" -> proposals.proposePayment(ctx, arguments)
+            // The PROPOSED-only invariant is enforced HERE, on the call path, not left to whichever
+            // ProposalPort is bound (T-E4, #2414). See ProposedOnly for why it is a whitelist of one.
+            "propose_payment" -> ProposedOnly.enforce(proposals.proposePayment(ctx, arguments))
             else -> return ToolCallResult(listOf(ToolContent(text = "Unknown tool: $toolName")), isError = true)
         }
         // The charter's `data_scope.pii: masked` (agents.yaml, `mcp-anonymous`) is enforced HERE —
--- a/openbank-mcp-service/src/main/kotlin/com/openbank/mcp/application/ProposedOnly.kt
+++ b/openbank-mcp-service/src/main/kotlin/com/openbank/mcp/application/ProposedOnly.kt
@@ -60,7 +60,7 @@ object ProposedOnly {
     // rather than skipped, so a numeric or object status cannot slip through unexamined).
     private fun collectStatuses(node: JsonNode, into: MutableList<String>): List<String> {
         if (node.isObject) {
-            node.fields().forEach { (name, value) ->
+            node.properties().forEach { (name, value) ->
                 if (name == STATUS_FIELD) {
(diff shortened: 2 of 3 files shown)


JiRaska:
Heads-up on a likely conflict: PR #2481 (`feat(mcp)`, PII masking) also edits `McpToolRegistry.kt`, and both PRs change the same `call()` dispatch — #2481 wraps every tool result in the masker, this one wraps the `propose_payment` branch in `ProposedOnly.enforce`. Textually adjacent, semantically independent: whichever lands second needs the other's wrapper preserved, i.e. the masker applied to the result of the enforced proposal, not instead of it. Flagging it now because the two-concurrent-PRs-on-one-file case passes CI on each side alone.

PR #2556, opened 2026-07-26
feat(admin-ui): add the components/ui primitive layer and one status vocabulary (ADR-0208 D1/D2)

JiRaska:
## Summary
First consolidation tranche under ADR-0208 (#2548).

**Adds `src/components/ui/`, which did not exist at all** — `StatusBadge`, `PageHeader`, `Card`/`StatCard`, `DataTable` — built on the classes already in `globals.css` rather than new CSS, using the `cn()` helper `lib/utils.ts` already exported.

**One status vocabulary (`statusTone`)**, derived from the values that actually appear across `src/app/**/page.tsx` (`ACTIVE` ×29, `UNKNOWN` ×16, `DRAFT` ×11, `CRITICAL` ×10, …) — not an invented taxonomy. Two properties worth reviewing specifically:
- **Fails safe.** An unrecognised status resolves to `neutral`, never `success`. A status nobody taught the helper about must not render as a green tick on a compliance screen.
- **Exact match, no substring.** Substring matching is how `PENDING_FAILOVER` ends up red.

**Migrates `src/app/system/readiness/page.tsx`** as the tranche's proof the primitives fit a real page — **11 raw colour literals → 0**. Its GO/NO-GO gate uses `StatusBadge`, its four summary tiles use `StatCard`, and the 0–3 maturity scale names a `Tone` per level instead of carrying its own hex + `rgba` pair. The scale itself stays local: it is a domain scoring scale, not a lifecycle status, so it does not belong in the shared vocabulary.

## Two deliberate deviations, stated rather than smuggled
1. **Adds 9 classes to `globals.css`** (`.stat-value`/`.stat-label`/`.stat-hint`, six `.tone-text-*`). ADR-0208 says primitives consume the vocabulary rather than grow it — but `.stat-card` only ever styled the *container*, so every page hand-rolled its internals inline, including a colour derived from a literal (`background: ${k.color}18`). The alternative is inline styles inside the shared primitive, which reintroduces the problem one layer down.
2. **The no-raw-hex guard does NOT ship yet.** It would fail on the 45 other pages still carrying literals, and a guard that is red on day one gets disabled rather than obeyed. It ships with the tranche that clears the last of them, exactly as ADR-0208 D2 specifies.

## Test plan
- [x] `npx tsc --noEmit` — exit 0
- [x] `npx eslint src/components/ui src/app/system/readiness/page.tsx src/test/ui-tone.test.ts` — 0 errors (1 pre-existing `set-state-in-effect` warning on an untouched `useEffect`)
- [x] `npx vitest run src/test/ui-tone.test.ts` — 11 passed
- [x] **Verified the tests catch a regression, not just pass**: flipping `statusTone`'s default from `'neutral'` to `'success'` fails 2 of the 11 cases; reverted and confirmed restored.
- [x] Full suite: 771 passed, **5 pre-existing failures** (`finops-taxonomy` ×2, `finops-allocation` ×2, `service-registry.guard` ×1). Confirmed pre-existing by stashing this branch's changes and re-running on clean `origin/main` — the same tests fail there. Not introduced here, not fixed here.

Refs ADR-0208 (merged #2548). N/A — first tranche of a merged ADR; subsequent tranches will each carry their own scope.


JiRaska:
## Visual verification of the review fixes

The three regressions review found were all **invisible to typecheck, lint and the mount-only smoke test** — a wrong CSS class or wrong geometry renders as broken markup that every automated check happily passes. So I verified them by eye rather than declaring them fixed.

Rendered the migrated page to static HTML from a throwaway vitest dump with `globals.css` inlined, served it over HTTP, and measured the DOM:

```
swatchCount: 24                      // 4 services × 5 dims + 4 legend
swatches:      26×26, padding 0px, radius 8px, digits "3","3","2"
smallSwatches: 18×18 ×4,  digits "0","1","2","3"   // all non-zero width
gateBadges:   ["GO","NO-GO","NO-GO","GO"]
```

That is exactly what was broken before: the 18px legend swatch previously had 20px of horizontal padding inside an 18px border-box (zero content width), and the 26px tile rendered as a circle. The screenshot confirms rounded squares with legible digits, the restored `ClipboardCheck` header icon, and StatCard tinting both label and value (GO green, NO-GO red, Money-path amber).

**Two honest limitations of this method**, so nobody over-trusts it later:
1. The dump inlines only `globals.css`, so **Tailwind utilities are absent from it** (`tailwindUtilitiesInDump: false` — `.stat-label` computed `display: block`, not `flex`). Anything depending on `flex`/`items-center`/`gap-*`/`text-right` is NOT covered. I checked those separately against `tailwind.config.ts`: `./src/components/**/*.{js,ts,jsx,tsx,mdx}` is in `content`, so the utilities used by the new primitives do compile.
2. It is a static snapshot — no interaction, no dark mode, no responsive check.

The throwaway dump test was deleted; nothing temporary is committed. Follow-up worth doing before the next tranche touches more pages: this repo's Playwright setup is 1 spec and `continue-on-error: true` (advisory), which is why a CSS regression can reach `main` with everything green.

JiRaska:
## Correction: the "5 pre-existing test failures" claim in this PR was wrong

This PR's test plan said the suite had *"5 pre-existing failures (`finops-taxonomy` ×2, `finops-allocation` ×2, `service-registry.guard` ×1)"* and that I had confirmed them pre-existing by stashing the changes and re-running on clean `main`.

**There are no failures.** Run correctly:

```
$ npm test
Test Files  48 passed (48)
     Tests  781 passed (781)
exit 0
```

Root cause of my error: I invoked `npx vitest run`, which **skips `pretest`** —
`pretest: node scripts/generate-governance.mjs && node scripts/generate-catalog.mjs`.
Those scripts generate `catalog.json` and the governance fixtures the finops and
service-registry tests read, so without them the tests fail with
`ENOENT: ... open 'openbank-admin-ui/catalog.json'`.

Worse than the mistake itself: **my verification shared the mistake.** I "confirmed pre-existing" by stashing this branch's changes and re-running — with the same `npx vitest run`. Of course it failed identically on clean `main`; the probe carried the defect it was supposed to rule out. That is the failure class this repo's own guidance names repeatedly (a check that cannot express the failure reports a plausible wrong answer), and I walked into it after quoting it.

Nothing in the merged diff is affected — `tsc`, `eslint`, the targeted `ui-tone` tests and the visual verification all stand. Only the claim about the suite's baseline was false, and it would have left a future contributor believing admin-ui ships with 5 red tests. Use `npm test`, not `npx vitest run`.

PR #2571, opened 2026-07-26
fix(authz): prohibit the shared M2M identity from role-only writes fleet-wide

JiRaska:
## Summary
Remediates **GHSA-58jq-9hq3-66jr**. `service-account-openbank-services` — the identity nearly every backend service authenticates as — carries `ROLE_OPERATOR` in the realm, and `AuthorizeInterceptor` classifies every `client_credentials` JWT as `HUMAN`. So every `operator-<domain>-write` rule, which checks only `type == "HUMAN"` plus that role, admitted **any** backend service to **any** write in that domain, for any resource. ~19 services.

**Fixed at the `allow` head in the shared `rest.rego`, not per service.** That is the same reasoning the neighbouring `prohibited` block already documents: *"Gating the allow head, not each reason, makes this impossible to bypass by enriching the input."* A per-rule exclusion has to be remembered 19 times and again by whoever writes the 20th, and forgetting it **fails open**.

### What it deliberately does not block
| Kept working | Why |
|---|---|
| **Reads** (`operator-read-any`, `compliance-read-any`) | party-service's GDPR Art. 15 aggregation calls kyc-service and card-issuance-service with this exact identity and depends on them |
| **Writes granted by a reason that NAMES the caller** (`service-consent-m2m-marketing`, `m2m-sanctions-screening`, `service-sca-shared-client-m2m`) | The guard requires that **every** reason admitting the principal be a role-only write reason, so one identity-scoped reason is enough to allow. Naming the caller and enumerating actions is the sanctioned way to grant an M2M write |
| **Human operators/admins** | Keyed on one service-account identity string; no human user can hold it |

Five rego tests, and **verified falsifiable**: deleting the guard flips the two deny tests to FAIL (`PASS: 81/85`). The other three cover the real risk of a fix at the allow head — being too broad. (2 unrelated AI-agent charter tests fail identically on clean `main`: 78/80 baseline vs 83/85 here.)

## A larger finding — reported, not silently shipped as "fixed"
The prohibition matches on the reason **name**, which makes the convention load-bearing. So this adds `check-operator-write-naming.py` — and running it revealed **the convention was never actually followed**. 18 further role-only WRITE rules use other names and therefore **escape this fix**:

`operator-statement-close`, `operator-statement-export`, `operator-statement-close-run-trigger`, `operator-tpp-registry-blacklist`, `supervisor-overdraft-limit`, `dispute-staff-write`, `operator-fx-trigger`, `operator-fx-approval-decide`, `operator-year-close-attest`, `operator-vop-verify`, `operator-pid-resolve`, `operator-party-status`, `operator-anacredit-create`, `operator-standing-order-pause`, `operator-aml-case-update-decision`, `operator-kyc-case-update-check`, `operator-kyc-case-pep-rescreen`, `operator-kyc-case-review-disposition`

**They are baselined, not fixed — and that is deliberate.** The complete fix is to invert the default for that identity (deny everything except an explicit `(identity, action)` allow-list). That list **cannot be guessed**: `AUTHZ_ENFORCE` is `true` on most of those services, so a missing entry is a money-path outage, not a test failure. Each needs its real callers confirmed first. Guessing it here would repeat the SBOM Audit→Enforce incident (a flip on a false precondition that killed 5 workloads for 4 days).

Registered in `rules.yaml: role_only_write_rule_naming` with a `target_enforce_date` and the blocker stated, so it carries graduation pressure rather than sitting advisory forever (the #2392 failure class). `check-advisory-gate-registration.py` passes: *"10 advisory governance gate(s) checked … all registered with a deadline."*

## Test plan
- [x] `opa test rest.rego rest_test.rego` — 83/85; the 2 failures are pre-existing (clean `main`: 78/80, same two)
- [x] **Guard falsifiable, both directions, each verified against a known-positive**: a NEW non-conforming role-only write rule → exit 1; a stale BASELINE entry → exit 1; restored → exit 0
- [x] `check-advisory-gate-registration.py` — passes
- [x] `rules.yaml` parses (`yaml.safe_load`)
- [x] Full `gen-*opa-bundle*.sh` regen included (66 bundle files); **verified idempotent** on a second run, and the guard is present in all 74 generated bundles
- [x] No files changed outside `.github/`, `openbank-libs/governance/` and `openbank-infra/gitops/components/`

Refs GHSA-58jq-9hq3-66jr.

### DIFF
--- a/.github/scripts/check-operator-write-naming.py
+++ b/.github/scripts/check-operator-write-naming.py
@@ -70,29 +70,64 @@ DECLARED_EXCEPTIONS = {
 #
 # Shrinking this set is the work. Adding to it is a regression and the guard says so.
 BASELINE = {
-    "operator-aml-case-update-decision",
-    "dispute-staff-write",
-    "operator-kyc-case-update-check",
-    "operator-kyc-case-pep-rescreen",
-    "operator-kyc-case-review-disposition",
-    "operator-standing-order-pause",
-    "operator-statement-close-run-trigger",
-    "operator-statement-close",
-    "operator-statement-export",
-    "operator-tpp-registry-blacklist",
+    # Verified individually, not assembled by hand: derived by emptying this set and reading what
+    # the check reports. An earlier revision of this list named 18 rules; 10 of those were NOT
+    # reachable by the shared identity at all — they either bar service-accounts outright
+    # (`not startswith(input.principal.id, "service-account-")`, as statement-close/-export do),
+    # gate on a role the shared account does not hold (ROLE_COMPLIANCE / ROLE_CREDIT_RISK /
+    # ROLE_LENDING_OFFICER), or are reads. Overstating a debt list is not a safe error: it hides
+    # the real entries among noise and invites the whole thing being ignored.
     "operator-anacredit-create",
-    "supervisor-overdraft-limit",
-    "operator-fx-trigger",
     "operator-fx-approval-decide",
-    "operator-year-close-attest",
-    "operator-vop-verify",
-    "operator-pid-resolve",
+    "operator-fx-trigger",
     "operator-party-status",
+    "operator-pid-resolve",       # pid.resolve — a lookup; classified here pending confirmation
+    "operator-standing-order-pause",
+    "operator-vop-verify",        # vop.verify — a verification; same caveat
+    "operator-year-close-attest",
 }
 
 REASON_RE = re.compile(r'allowed_reasons\s+contains\s+"([^"]+)"\s+if\s*\{')
-ROLE_GATE_RE = re.compile(r'\brole\s+in\s+input\.principal\.roles\b|"ROLE_(OPERATOR|ADMIN)"')
-IDENTITY_PIN_RE = re.compile(r"input\.principal\.id\s*==")
+# Role gating, in every idiom this fleet actually uses. Review of PR #2571 found the first
+# version missed three: a membership test whose variable is not literally `role`
+# (`some r in {...}; r in input.principal.roles`), a bare role literal with no `some` binding
+# (`"ROLE_PAYMENTS" in input.principal.roles`), and a role set sourced from `data.rules`.
+# A missed idiom here is a role-only write rule that never gets discovered, which is the one
+# failure mode this script exists to prevent.
+# Only the roles the SHARED IDENTITY ACTUALLY HOLDS make a rule reachable by it.
+# openbank-realm.json gives `service-account-openbank-services` exactly
+# `realmRoles: ["ROLE_OPERATOR"]` — nothing else. So a rule gated on ROLE_COMPLIANCE,
+# ROLE_CREDIT_RISK or ROLE_LENDING_OFFICER is unreachable by it no matter how role-only it is,
+# and flagging those was the second over-broad version of this check.
+#
+# Keep this list in step with the realm. If the shared service-account is ever granted another
+# role, add it here — otherwise this guard silently stops covering the rules that role opens.
+SHARED_IDENTITY_ROLES = ("ROLE_OPERATOR",)
+
+ROLE_GATE_RE = re.compile(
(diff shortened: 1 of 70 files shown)


JiRaska:
## Review finding accepted as BLOCKING — the fix would have caused a money-path outage

Independent review blocked the first revision and was correct. Verified myself before changing anything:

**`transaction.create`** — transaction-service's only two reasons are `viewer-transaction-read` and `operator-transaction-write`. It is called with the shared client by **six** services (account-service's welcome-bonus credit, sepa-instant, domestic-payment, swift, interest, sdd), and its own rego states the exposure verbatim: *"RESIDUAL RISK — shared identity, no per-caller narrowing"*. `AUTHZ_ENFORCE: "true"`. My name-matched guard would have returned 403 to all six.

**`settlement.create`** — sole reason `operator-settlement-write`, resourceless, no alternative rule. Same shape, also enforced.

Removing the only path is not a fix for an over-grant on that path.

## What changed
The predicate is no longer a name pattern — it is **data**: `rules.yaml: shared_m2m_write_prohibition.reasons`, opt-in. A reason joins the register only once its service carries a rule that **names the caller and enumerates the actions**, so denying the role-only path removes an over-grant instead of the only path. Deny-where-an-alternative-exists, never deny-and-hope.

- **Registered (6)**: the reasons the review independently confirmed — consent, ledger, balance, fraud, sca, sepa-payment.
- **Deferred (15)**, each with a stated reason. Including four that *do* carry an identity-pinned rule — `fx`, `account`, `clearing`, `domestic-payment` — because **"a rule exists" is not "the rule covers the actions the role-only path was serving"**. My first attempt at this revision listed them on that basis, which would have repeated the same guess one level down; caught it and moved them to deferred. `operator-party-write` is deferred specifically: onboarding-service PUTs `party.kyc-status` via the shared client, latent only because `AUTHZ_ENFORCE` is false there today.

## Other review findings, all addressed
- **Mislabeled test** (`test_deny_other_service_account_write_via_role_only_reason`) — it claimed to prove coverage of a different service-account while reusing the same id, so it demonstrated nothing. Replaced with what it actually tests (role-independence), and the scope is now stated: the guard is keyed on one identity; other service-accounts are deliberately out of scope because their callers have not been enumerated.
- **No end-to-end test** — structurally impossible here: shared `rest.rego` contains zero `operator-*-write` rules (they all live in per-service extensions, outside CI's `opa test` scope). Two new tests cover the properties the opt-in design introduces instead: a role-only write reason *not* in the register still allows (the property that keeps `transaction.create` working), and an absent register key changes no behaviour.
- **Baseline overstated exposure** — `operator-statement-close`/`-export` carry `not startswith(input.principal.id, "service-account-")` and are unreachable by the shared identity; `operator-pid-resolve`/`operator-vop-verify` are lookups. The script's framing is corrected: it is the **discovery** half, and its baseline lists rules that are *invisible as candidates for the register* rather than "reachable today". I have not re-triaged all 18 — that belongs with each service's caller verification.
- **`ROLE_GATE_RE` false negatives** (`r in input.principal.roles` with a non-`role` variable, `"ROLE_PAYMENTS"` alone, `data.rules`-driven role sets) and **`IDENTITY_PIN_RE`** missing the `not startswith(...)` idiom — real gaps, not fixed here. Noted for the follow-up; they widen the discovery set, they do not weaken the prohibition.

## Test plan
- [x] `opa test` — 85/87 (the 2 failures are the pre-existing charter-bridge scope artefact; clean `main` shows the same two)
- [x] `test_allow_role_only_write_reason_not_in_the_register` — **PASS**, which is the direct regression test for the blocking finding
- [x] `check-advisory-gate-registration.py` — exit 0
- [x] `rules.yaml` parses; register = 6 reasons, 15 deferred
- [x] Confirmed `operator-transaction-write` is **absent** from the register in the regenerated transaction bundle
- [x] Full bundle regen, verified idempotent on a second run

PR #2580, opened 2026-07-26
feat(admin-ui): Customer 360 over the analytics silver layer (ADR-0210)

### LINKED ISSUE #2573: ADR-0209 D2/D3: do NOT build crm-service or campaign-service yet — evaluate the analytics-sink read model first
ADR-0209 D2/D3 — **decided NOT to fund yet**, with one thing to evaluate that could make D2 much cheaper. ADR-0209 D6 already states that this ADR sequences work without scheduling it and that three services is a funding decision; this issue records the call.

## Why not yet
- **Nothing in ADR-0199/0200/0201 is built.** `git grep -lIiE 'campaign|\bcrm\b|next-best-action'` over every `openbank-*/src` and `openbank-admin-ui` returns zero files. This is three services' worth of build from a standing start.
- A service here is not just a pod: Deployment + OPA sidecar + a CNPG Postgres + Kafka topics + NetworkPolicies + an OPA bundle + a runbook + backups. ×2 for D2+D3.
- **ADR-0201 D5 is unbuildable as specified regardless.** ADR-0140 phase 1 shipped (`FeatureDefinition`, `OnlineFeatureStore`, `RedisOnlineFeatureStore`) but there is **no offline store, snapshotter or as-of join** anywhere. ADR-0201 D4 names that as its dependency and explicitly forbids the workaround (training on the online store). So the commercially interesting part cannot be reached by funding D2/D3 sooner.

## What to evaluate before committing to a 4th service (the cheap alternative)
ADR-0199 D2 specifies the Customer 360 as "a CQRS projection over the outbox topics that already exist" with its own CNPG Postgres. But **`openbank-analytics-sink` already consumes domain events into ClickHouse** (bronze layer, per its `application.yaml` — `LoggingAnalyticsSink` by default, `ClickHouseAnalyticsSink` when activated), and already implements the ADR-0118 erasure pattern ADR-0199 D4 wants (`ErasureService` doing per-row crypto-shredding, gated by `RetentionPolicies`).

So the question worth answering **before** standing up `crm-service` is: can the Customer 360 read model start as a **query surface over analytics-sink's existing bronze**, rather than a new service with a new database and a new set of consumers? If yes, D2 becomes a read API plus some SQL rather than a service. If no, the reason will be concrete (ADR-0089's rule that customer-facing figures come from the owning service, latency, or the ADR-0199 D3 exclusions) and ADR-0199's shape is vindicated on evidence rather than assumed.

This is a **spike, not a build**: timebox it, and its output is a recommendation plus the reason, appended here.

## What proceeds instead
- #2572 — ADR-0209 D1 (Temporal extraction). Pays back on 14 services regardless of this decision.
- ADR-0209 D5 (campaign-copilot as an MCP tool registration, not a service) — see the sibling issue.

## What must not happen
Starting ADR-0201 D5 (the NBA model). ADR-0209 D4 is explicit and the reason is not preference: without an as-of join the only way to train is on the online store, reintroducing exactly the training/serving skew ADR-0140 exists to prevent — a defect that shows up as a model which validated well and performs worse in production, indistinguishable from ordinary drift.

Refs ADR-0209 D2/D3/D4/D6, ADR-0199, ADR-0200, ADR-0201, ADR-0140.

JiRaska:
## Summary
ADR-0209 D2 asked one question before funding a fourth service: could the Customer 360 read model start as a **query over analytics-sink's existing bronze** instead of ADR-0199's new `crm-service`? Spiked it. The answer is yes, and the evidence is stronger than I expected.

| Verified on `origin/main` | Consequence |
|---|---|
| analytics-sink already consumes `account, transaction, balance, party, kyc, consent, sca, documents, onboarding.funnel, feedback` events | essentially the **exact** list ADR-0199 D2 enumerates — `crm-service` would be a *second* consumer group over the same topics writing a second database |
| `silver_current_state` is a view over `bronze_events` reducing per aggregate via `argMax(…, (aggregate_version, occurred_at, ingested_at))` | **that already IS the CQRS current-state projection** ADR-0199 D2 specifies — deterministic, out-of-order tolerant, already written |
| `silver_as_of` = same reduction with `WHERE occurred_at <= {t}` | point-in-time reconstruction, which ADR-0199 did not offer at all |
| `gitops/components/analytics/` has `clickhouse.yaml`, `clickhouse-init-configmap.yaml`, `clickhouse-auth-externalsecret.yaml` | ClickHouse is **deployed**, not aspirational |
| `ErasureService` / `ErasurePort` / `VaultCryptoErasure` in analytics-sink | ADR-0118 erasure is already implemented, by the same crypto-shredding mechanism ADR-0199 D4 settled on after rejecting replay |

**Both objections ADR-0199 D4 raised against a projection are already answered here.** Retention: bronze is durable ClickHouse, not the 7-day `cleanup.policy: delete` Kafka topics that made a rebuild impossible. Erasure: see above.

So **ADR-0199's decision is right and unchanged** — a derived read model that owns no facts, over the outbox topics. Only *"in a new crm-service"* was wrong, and it was written before anyone checked what analytics-sink already did. D6 supersedes it in mechanism, retains it in principle. ADR-0200's campaign-service, if funded, reads this view; nothing in it depended on the 360 being its own service.

## What this removes from the CRM programme's cost
A service, a CNPG Postgres, a consumer group, and a second ADR-0118 erasure implementation — which was ADR-0209's stated blocker on funding D2.

## The real work, and the thing most likely to go wrong
**D2 — the party key.** `bronze_events` is keyed by `(aggregate_type, aggregate_id)`, so a party event's `aggregate_id` is the partyId but an account or transaction event's is an accountId or transactionId. Assembling one party's view needs account→party (and card→party, case→party) resolution, materialised as a ClickHouse view alongside the existing silver views. It is the only thing this ADR adds to the schema — and its Negative consequence states the risk plainly: **get it wrong and the 360 shows another customer's account, so it needs a test asserting isolation, not just assembly.**

D3 keeps ADR-0199 D3's exclusions verbatim (no authoritative balances, no transaction-level rows, no KYC document content, no risk scores) and adds a staleness field so a consumer sees how old the view is instead of assuming. D4 is operator-facing only — no `customer-edge` route, because that would put a non-authoritative aggregate in front of the customer.

Docs-only. Implementation follows: the account→party view, a read endpoint, and the admin-ui page.

## Test plan
- [x] `bash .github/scripts/check-adr-registry.sh` — content valid
- [x] `bash docs/adr/gen-index.sh` run **last** and committed (it reverts derived files on a failing check); 0210 present in all three derived artefacts (202 ADRs)
- [x] Every substrate claim above verified by reading `origin/main` — the ClickHouse init ConfigMap's view definitions, analytics-sink's `application.yaml` topic list, and the erasure classes — not from any ADR's own front-matter

Closes #2573 — this ADR is the answer to the spike that issue asked for.

Refs ADR-0209 D2, ADR-0199, ADR-0022, ADR-0118, ADR-0089.


JiRaska:
## Implementation added — Customer 360 is now in this PR, not just its ADR

The PR is retitled `feat(admin-ui)` because it now carries admin-ui code. Both release axes matter here (ADR-0029): a `docs` type would not have released admin-ui, so the code would have shipped without a version bump — `release_scope_mismatch`. release-please does the bump post-merge, as it did for #2556.

### It needed no infrastructure at all — which is D1's whole point
| Already existed | Evidence |
|---|---|
| ClickHouse creds in admin-ui | `admin-ui.yaml:121-125` — `CLICKHOUSE_URL` / `_USER` / `_PASSWORD` via the `clickhouse-auth` ExternalSecret |
| The query pattern | `/api/onboarding/funnel-analytics` already talks to the ClickHouse HTTP interface with env creds and always 200s with `available: false` |
| The projection | `silver_current_state` |

So this is **one route + one page** on the ADR-0208 primitives. No crm-service, no CNPG Postgres, no consumer group, no gitops change, no ClickHouse DDL.

### One deviation from the ADR's wording, stated rather than quietly taken
ADR-0210 D2 says the account→party mapping "materialises as a ClickHouse view". It is **SQL in the route** instead, because the ClickHouse schema is applied by an init ConfigMap **on first boot only** — a view added there would not exist on the already-running sandbox cluster, so the page would show nothing. Materialising it is the same follow-up as the deferred gold table.

### Isolation, which is the thing ADR-0210 D2 predicted would go wrong
Transaction events carry **no** `partyId` (only `accountId`), so a party's transactions are reached through the accounts it owns. Widen that and the 360 shows **another customer's account**.

`customer-360.test.ts` therefore asserts on **the SQL the route builds**, not on assembled output — a fixture-based test would pass against a leaking query, because the fixture simply would not contain the other party's rows. **Verified it catches a real leak**: removing the party filter from the `party_accounts` CTE fails it (1 failed / 4 passed); restored clean.

Also asserted: a non-UUID `partyId` is rejected **before any query is issued** (the route builds raw SQL, so the format check is the injection boundary, not a convenience — 4 hostile inputs, 0 queries reached ClickHouse), and the response interface cannot name `balance`/`amount`/`currency`, checked against the source because D3's guarantee is about what the route *cannot* emit.

### Test plan
- [x] `npm test` — **49 files / 791 passed, exit 0**
- [x] `npx tsc --noEmit` — exit 0
- [x] `npx eslint` on the new route, page, test and Sidebar — 0 errors
- [x] Leak probe: widened the CTE → test fails; restored → passes
- [x] **Rendered and looked at it** — dumped to static HTML with `globals.css` inlined and screenshotted: StatCards tinted, `ACTIVE`/`REVOKED` consents distinguished by tone, the `As of` staleness chip present, domain table sorted by event count

### Sandbox
Merging this releases admin-ui and auto-deploy publishes it. Note the known version-lag: release-please's bump is a **separate** commit, so if the deploy picks up the pre-bump SHA it needs re-dispatching with `--ref main`. I will confirm the page is live on the sandbox after merge rather than assuming the deploy landed.

Issue #2330, opened 2026-07-25
Multiple PR-time CI jobs killed by clustered runner-shutdown signal across unrelated services (not the #1218 dind-OOM pattern)


JiRaska:
## Symptom

Between ~12:20 and ~12:26 UTC on 2026-07-25, at least 5 concurrently-open PRs (#2318, #2321, #2229, #2309, #2247, #2316) each had multiple `Services CI` / `Security scan` / `Dependency submission` jobs die mid-build with the identical tail:

```
##[error]The runner has received a shutdown signal. This can happen when the runner service is stopped, or a manually started runner is canceled.
##[error]The operation was canceled.
Cleaning up orphan processes
```

Jobs were killed cleanly mid-task (mid-`compileKotlin`, mid-cache-restore, mid-`test`) across unrelated services and unrelated PRs, clustered within seconds of each other in each run — e.g. #2247's four jobs (fraud-service, fx-service, governance-auditor, interest-service) all died within a 3-second window (12:25:14–12:25:15). One job (#2309's `openbank-customer-edge`) instead saw ~15 Docker Hub registry timeouts pulling a Testcontainers image before failing — a second, possibly related, infra symptom in the same window.

None of the killed jobs showed a compile error, test assertion, or lint violation before the cancellation — every failure was purely infra-side. Confirmed via `gh run rerun` that the underlying commits build cleanly once the runner isn't killed out from under them.

## Why this isn't #1218

#1218 (fixed by #1222) was a `dind` OOM-eviction with request:0 — its signature is *"The self-hosted runner lost communication with the server"* and its kills are staggered by node memory pressure, one pod at a time. This incident's signature is *"received a shutdown signal"* / clean cancellation, and the kills are clustered — multiple unrelated jobs across multiple unrelated PRs dying within a few seconds of each other, in more than one time window today. That points at something acting on the runner pool as a whole (a scale-down, a controller-level pod delete sweep, a deployment rollout of the runner controller) rather than per-pod memory pressure. Worth checking whether #1222's dind request/limit fix regressed, or whether this is a distinct root cause in the ARC runner-pool config / autoscaler.

## Cost

Each hit needed a manual `gh run rerun` (full rerun where jobs showed `cancelled`, since `--failed` is a no-op against `cancelled` conclusions — only `failure` conclusions get retried). Multiplied across ~5 PRs this session alone; likely undercounted since it only surfaces when someone happens to look past "red = my problem" and reads the log tail.

## Ask

[...]

Issue #2177, opened 2026-07-25
Gradle daemon heap is a ratchet with nothing measuring it — the third OOM, and an OOM'd fleet-lint is a vacuous red


JiRaska:
### 1. Nothing measures Gradle heap headroom, so we only learn by going red

`fleet-lint` OOM'd on `main` (run 30142905147, `java.lang.OutOfMemoryError: Java heap space` in Gradle's `MissingTaskDependencyDetector`) with **no code change** — the 11 commits since the last fully-green run touch only gitops/admin-ui/docs/governance/CI. Pure capacity. Fixed by 4g → 6g in PR #2171.

That is now the third instance of the same class:
- `dependency-submission.yml` — `Java heap space`, ratcheted 4g → 6g (red for three days on `main` before anyone noticed).
- `security.yml` — already runs at 6g.
- `fleet-lint` — this one.

Each time the fix is the same and each time we find out by breaking. There is no signal on how close any of these is to its ceiling, and the fleet only grows. `fleet-lint` is the worst case: it is the only workflow that configures **and** executes every module in one invocation, so its peak heap scales with the whole task graph.

**Suggested:** record peak JVM heap per run (Gradle build scan, `-XX:+HeapDumpOnOutOfMemoryError`, or simply logging `Runtime.maxMemory`/GC stats from a build listener) and alert on headroom below a threshold, so the ratchet is anticipated rather than discovered.

### 2. An OOM'd fleet-lint is a vacuous RED, not just a failure

An OOM aborts **part-way through the task graph**. Measured on the same gate:

| run | actionable tasks | reads as |
|---|---|---|
| OOM'd (30142905147) | **455** | "the fleet is broken" |
| green after fix (30143653231) | **920** | fleet actually linted, 58 modules |

So roughly half the fleet was never linted at all, while the check reported a plain red indistinguishable from a real lint violation. On the one gate whose whole purpose is to be un-dismissable, that is the mirror of the vacuous-green problem: the reader cannot tell "your code is bad" from "the runner ran out of memory before reaching your code".

**Suggested:** make the gate distinguish infrastructure failure from lint failure — detect the OOM/abort case and fail with a distinct, explicit message (and ideally a distinct check name or annotation), rather than letting it land in the same red bucket as a real violation.

Found while closing #2107 (PR #2171).

JiRaska:
PR #2479 opened for **problem 2** — deliberately `Refs`, not `Closes`, since problem 1 is untouched.

[...]

Issue #2039, opened 2026-07-24
ARC capacity livelock: maxRunners=6 vs ~30-job full-fleet runs starves the queue for hours


JiRaska:
## Symptom

Full-fleet `services-ci` main-push runs stall for hours: on 2026-07-24 04:44 the oldest queued run was from **20:41** (8h queued). Runs reach ~29/33 jobs done, then **4 jobs (kyc/ledger/lending/libs) sit queued with 0 running**, the run never completes, and later-alphabet jobs (swift, transaction) never get created — so a provider verification that a merged fix needs (e.g. the #1348 swift pact fix) never republishes.

## Root cause — capacity livelock, NOT Karpenter

- `openbank-build` AutoscalingRunnerSet has **`maxRunners: 6`**.
- A full-fleet run (any build-relevant main-push, ADR path-completeness) is **~28–33 jobs**. With 8 such runs queued and only 6 runners, the 6 runners round-robin across dozens of jobs from multiple runs; **no single run ever holds enough runners to finish its matrix**, so runs pile up and the oldest starve.
- Confirmed NOT node/Karpenter: 28 nodes healthy, nodeclaims 6h+, zero pending runner pods, controller + listeners healthy. The `arc-stuck-runner-reaper` CronJob churns but cannot fix a capacity shortfall.

## Immediate relief (done 2026-07-24 ~04:45)

Cancelled 6 stale/superseded queued runs (pre-fix commits, checks moot) to concentrate the 6 runners on the fix-bearing runs. This is a manual band-aid, not a fix.

## Durable options (FinOps tradeoff — owner call)

1. **Raise `maxRunners`** (e.g. 6 → 12–16). Directly ends the livelock. Cost: peak spot/on-demand runner spend during CI bursts roughly doubles+ (currently m7gd/m6gd spot+on-demand mix in eu-north-1a). Only billed while runners are up (scale-to-zero at idle), so the delta is CI-burst-duration × extra-runner-hours, not a standing cost.
2. **Cut full-fleet frequency.** Main-push currently goes full-fleet on ANY build-relevant path for Pact completeness (issue #1208 already flags this saturates the 6-runner pool). A smarter "which providers' pacts could this change affect" selection would shrink most runs from ~30 jobs to a handful. Higher-effort, zero extra runner cost.
3. **Serialize full-fleet runs** (concurrency group so only one runs at a time) — removes the livelock without more runners, at the cost of longer end-to-end latency per run.

[...]

PR #2583, opened 2026-07-26
fix(ci): make the heap sampler's ceiling the real -Xmx, and stop it contradicting the runner-memory warning

JiRaska:
## The finding

#2500 (merged today) wired `gradle-heap-headroom.sh` into the CodeQL `java-kotlin` job — the job #2330 is about — so `-Xmx` would stop being a ratchet discovered by going red. I went to use its output to size a fix for #2330. **The instrument is misreporting, and its advice points the wrong way.**

Verbatim from run [30198139819](https://github.com/JiRaska/open-bank-oss/actions/runs/30198139819), job `89783296232`, two adjacent lines:

```
##[warning] ran with only 499 MiB of available system RAM at its lowest point
          (floor 512 MiB) ... See #2330.
##[warning] peaked at 454% of its configured -Xmx ... Raise -Xmx in this
          workflow's GRADLE_OPTS before it goes red, not after.
```

The runner is out of memory; the next line says give a JVM more of it. Acting on the second makes #2330's `Unable to connect to the child process 'Gradle Worker Daemon 80' ... build machine is extremely loaded` **worse**.

And `454%` is not a real number: a heap cannot exceed its own ceiling — a JVM that did would have thrown `OutOfMemoryError` rather than finishing the build.

## What this changes (instrument only — no build behaviour)

1. **Ceiling = `-XX:MaxHeapSize` from `jcmd VM.flags`**, not `GC.heap_info`'s `reserved`/`capacity`, which is the collector's current region bookkeeping rather than the max heap. That is how a JVM under `-Xmx6g` got a reported ceiling of 1344 MiB.
2. **`used` no longer sums the `Metaspace` / `class space` lines** `GC.heap_info` also prints. Metaspace is not heap and is not bounded by `-Xmx`; counting it put two different quantities in the numerator and denominator.
3. **"Raise `-Xmx`" is gated on the machine having memory to give.** Below `MIN_FREE_MB` the warning instead says reduce the concurrent footprint, and says why.

Plus the number #2330 actually needs and the table could not express: **peak concurrent heap** (max per-sample sum across live JVMs), with an explicit note that the per-JVM rows are a peak-per-pid roll-up over the whole build — the ~40 `GradleWorkerMain` rows are mostly JVMs that lived and died at different times, so reading the table as a snapshot overstates the load while reading one row understates it.

## Falsified, not assumed

| check | result |
|---|---|
| JVM launched `-Xmx512m` | reports **512 MiB**, 88% — sub-100%, where the old path gave 454% in CI |
| real Gradle daemon at `-Xmx3g` | reports **3072 MiB** |
| memory-starved branch | **executed**, printed text read — emits *"Do NOT raise -Xmx here"* |
| exit-code passthrough | `exit 7` → `rc=7`; still never changes a build's verdict |

The starved branch was unreachable on a developer machine (no `/proc/meminfo`), so a `vm_stat` fallback was added to make it exercisable. **A warning path only reachable on a runner is one nobody has ever read — which is precisely how the contradictory advice shipped.**

## Deliberately NOT changed

`maxRunners` (option 1 was already tried and reverted), the CodeQL build scope, `-Xmx6g`, `--max-workers`. The measurement does say the runner is memory-starved — but the instrument producing it was misreporting, so **no footprint change should be justified by this data until the corrected numbers land**. Acting on an unmeasured guess is what #2039 already cost. Analysis posted on #2330.

One further note for #2330: a large share of the `cancelled` CodeQL runs today are this workflow's own `concurrency.cancel-in-progress` superseding runs on rapid-fire main pushes — not livelock. The genuine signal is the 499 MiB floor and the worker-fork timeout.

## Linked issues

Refs #2330, Refs #2039, Refs #2177

### DIFF
--- a/.github/scripts/gradle-heap-headroom.sh
+++ b/.github/scripts/gradle-heap-headroom.sh
@@ -60,36 +60,61 @@ sample_once() {
 
     info="$(jcmd "$pid" GC.heap_info 2>/dev/null)" || continue
 
-    # Sum "used" across the heap regions the collector reports, and take the ceiling from
-    # "reserved"/"capacity" depending on collector. Values arrive as e.g. "used 1234M" or
-    # "used 1234567K".
-    used_kb="$(printf '%s\n' "$info" | grep -oE 'used [0-9]+[KMG]?' \
+    # Heap USED. Only the collector's heap line — `GC.heap_info` also prints "Metaspace
+    # used ..." and "class space used ...", and summing those in inflated every figure by
+    # roughly a gigabyte per JVM. Metaspace is not heap and is not bounded by -Xmx, so
+    # counting it against -Xmx compared two different things (#2330).
+    used_kb="$(printf '%s\n' "$info" \
+      | grep -vE '^[[:space:]]*(Metaspace|class space)' \
+      | grep -oE 'used [0-9]+[KMG]?' \
       | awk '{v=$2; u=substr(v,length(v)); n=v+0;
               if (u=="G") n*=1048576; else if (u=="M") n*=1024;
               s+=n} END {if (NR>0) printf "%d", s}')"
-    max_kb="$(printf '%s\n' "$info" | grep -oE '(reserved|total reserved|capacity) [0-9]+[KMG]?' \
-      | awk '{v=$NF; u=substr(v,length(v)); n=v+0;
-              if (u=="G") n*=1048576; else if (u=="M") n*=1024;
-              if (n>m) m=n} END {if (NR>0) printf "%d", m}')"
+
+    # Heap CEILING — read the effective -Xmx from VM.flags, never from GC.heap_info.
+    #
+    # This is the whole point of the fix. `GC.heap_info`'s "reserved"/"capacity" is the
+    # collector's current region bookkeeping, NOT the max heap, and on a JVM launched with
+    # `-Xmx6g` the old parse reported a 1344 MiB "ceiling" and therefore "454% of its
+    # configured -Xmx" — a figure that is arithmetically impossible for a real heap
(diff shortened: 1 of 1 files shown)


JiRaska:
Two notes after pushing the follow-up commit.

**1. This PR's own CI does not exercise the changed script.** `security.yml`'s `detect-deps` gate only runs CodeQL when a path matching its `CODE_RE` changed, and `.github/scripts/` is not in that pattern — so `CodeQL (java-kotlin, manual)` is **skipped** here and the sampler never runs on this PR. That is correct behaviour (CodeQL analyses Java/Kotlin/TS, not shell) and I am deliberately *not* widening `CODE_RE`, which would add a ~23 min job to every scripts-only PR for no analysis value.

It does mean the corrected numbers land on the next main-push or scheduled run rather than here, so the local falsification is what backs this PR: ground truth against a JVM launched `-Xmx512m` (reports exactly 512 MiB) and a real Gradle daemon at `-Xmx3g` (reports 3072 MiB), plus the memory-starved branch executed and its printed text read. The script is advisory-only and passes the wrapped command's exit code through unchanged, so the blast radius of a mistake here is a wrong table, never a wrong build verdict.

**2. Added a per-pid `-Xmx` cache** (`a97d012`). Reading the ceiling from `jcmd VM.flags` introduced one extra jcmd attach per JVM per sample — roughly 40 JVMs every 5s for 20+ minutes on the CodeQL job. jcmd attaches over a socket and is not free, and adding load to the very job whose memory pressure this script measures would be a self-defeating way to instrument it. `-Xmx` is fixed for a JVM's lifetime, so it is now read once per pid. Re-verified: same 512 MiB against the same ground truth.

PR #2585, opened 2026-07-26
refactor(libs): extract the 14 duplicated TemporalClientProducer copies into openbank-libs-runtime

### LINKED ISSUE #2572: ADR-0209 D1: extract the 14 duplicated TemporalClientProducer copies into openbank-libs-runtime
ADR-0209 D1 — **decided to proceed**. This is the one item in the CRM/campaign sequence whose payback does not depend on CRM ever being funded.

## The facts, measured on `origin/main`
- **14 copies of `infrastructure/temporal/TemporalClientProducer.kt`**, and nothing Temporal-related anywhere under `openbank-libs-domain` or `openbank-libs-runtime`.
- Package-normalised, two randomly-sampled copies differ by **9 lines, all of which are the SPDX header**. The code bodies are identical.
- ADR-0200 D1 already noted the duplication in passing and accepted that campaign-service would copy the pattern "rather than reusing a library that does not exist". It would be the **15th** copy.

## The non-obvious constraint: the licence split
The 14 are **6 Apache-2.0 / 8 AGPL-3.0-only**:

| Apache-2.0 | AGPL-3.0-only |
|---|---|
| transaction-service, statement-service, settlement-service, sepa-payment, fx-service, domestic-payment | release-steward, governance-auditor, flaky-test-hunter, finops-agent, docs-truth-agent, devops-agent, control-liveness-sentinel, authz-policy-auditor |

**The extraction target must be Apache-2.0 `openbank-libs-runtime`.** That works because AGPL modules may consume Apache-2.0 code — and `rules.yaml: agpl_modules.rule` forbids only the reverse ("no Apache-2.0 module may declare project(':openbank-…-agent'…) as a dependency"). Extracting into an AGPL-licensed lib instead would put all 6 Apache services in violation of that rule. The differing headers reflect each *module's* licence (stamped by `scripts/add-license-headers.sh`, path-aware since #2280), not distinct authorship, so no relicensing question arises for identical code.

## Scope
- One `TemporalClientProducer` in `openbank-libs-runtime`, plus its config contract and a worker-registration helper.
- Migrate all 14 call sites; delete the copies.
- Acceptance: every one of the 14 services still builds **and** its workflow tests pass. Per ADR-0209's Negative consequence, this must be **one sweep** — a half-done migration leaves the fleet carrying two Temporal wiring conventions, which is worse than either.

Refs ADR-0209 D1, ADR-0200 D1.

JiRaska:
Closes #2572. Refs ADR-0209 D1, ADR-0200 D1, ADR-0122.

## The issue's premise was wrong — the copies are NOT identical

#2572 states that two randomly-sampled copies "differ by 9 lines, all of which are the SPDX header". Sampling two of fourteen found the one relationship that holds; it does not hold across the set. Hashing all 14 package- and licence-normalised gives **7 distinct texts in 3 behavioural shapes**:

| shape | services | difference |
|---|---|---|
| A (9) | authz-policy-auditor, control-liveness-sentinel, docs-truth-agent, finops-agent, flaky-test-hunter, governance-auditor, release-steward, settlement, transaction | `TemporalConfig`-driven, Micrometer metrics scope attached |
| B (1) | **devops-agent** | `TemporalConfig`-driven, **no metrics scope at all** |
| C (4) | fx, sepa, domestic, statement | `@ConfigProperty`-driven, metrics scope, three different `namespace` defaults |

Plus the 10 `TemporalConfig.kt` copies, each with its own `taskQueue`/`namespace` `@WithDefault`.

Folding B into A silently would have started emitting a new Temporal gRPC client metric series for devops-agent. Per the brief, that divergence is **preserved as config**, not flattened: `openbank.temporal.metrics-enabled` (default `true`), set `false` in devops-agent's `application.yaml` with a comment saying exactly why and how to adopt the fleet norm.

## What changed

- **New**: `openbank-libs-runtime/.../com/openbank/libs/temporal/{TemporalConfig,TemporalClientProducer}.kt`. Framework-touching, so `runtime` not `domain` (ADR-0122). Apache-2.0 is the only valid target — 8 of 14 consumers are AGPL-3.0-only and `rules.yaml: agpl_modules.rule` forbids Apache→AGPL, never the reverse.
- **Deleted**: 14 producer copies + 10 config copies (−880 lines).
- **Migrated**: 24 call sites; no service `build.gradle.kts` needed a change (`:openbank-libs` already re-exports `:openbank-libs-runtime` via `api()`).
- **Config made explicit**: the shared contract cannot carry 14 different `@WithDefault`s, so each service's previous *effective* values are now written into its `application.yaml` (settlement's `server-url`/`namespace`; transaction's whole block).

### Drive-by finding: transaction-service's Temporal yaml has never been read

Its keys sat under `openbank.transaction.orchestration.temporal.*`, but `TemporalConfig` is and always was `@ConfigMapping(prefix = "openbank.temporal")`. Every value there — including the `%test` override — was inert; the service ran on the Kotlin `@WithDefault`s. Moved to the prefix that is actually bound. **Prod is unaffected either way**: the GitOps manifests set `OPENBANK_TEMPORAL_*` env vars, which win over both. Worth noting as its own class of bug — a yaml block that looks authoritative and binds to nothing is invisible to every gate we have.

## Dependency verification

Zero new `gradle/verification-metadata.xml` entries. The first attempt failed on ~27 artifacts: this module imports no Quarkus platform BOM, so a transitive resolve of `temporal-sdk` drags in its *own* grpc 1.54.1 / guava 31.1-jre / protobuf versions, which the fleet pins nowhere. All four new coordinates are therefore declared `isTransitive = false`; each is already pinned (the 14 services resolve them). `@ConfigMapping` needs no declaration — it arrives from the existing `quarkus-arc` compileOnly.

## Verification

- `:<svc>:quarkusBuild` **green for all 14** — CDI wiring surfaces only there, and a shared `@ApplicationScoped` producer consumed across module boundaries is exactly the case that fails there and nowhere else.
- `detekt` + `ktlintFormat` on all 14; `koverVerify`/`detekt`/`ktlintCheck`/`test` on `openbank-libs-runtime`; unit tests green for settlement + transaction (the modules whose tests mock `TemporalConfig`).
- `ktlintFormat` also reformatted 11 *unrelated* fx-service files carrying pre-existing violations — reverted to `origin/main` so the diff stays scoped, and fx re-verified with `quarkusBuild` afterwards.

### New test, falsified

`TemporalClientProducerLazinessTest` asserts that constructing the producer reads **no** config property — which is what precedes dialling `openbank.temporal.server-url` over gRPC, and what every `@QuarkusTest` swapping in an in-process `TestWorkflowEnvironment` depends on.

**Falsified before commit**: replacing `by lazy` with an eager `= run {` turns it red (`failures="1"`, read from the JUnit XML, not the console). It fails on `NoClassDefFoundError: com/google/common/base/MoreObjects` — i.e. construction genuinely tried to build real service stubs and reached for guava, which the deliberately non-transitive test classpath does not carry. Restored to `by lazy`, green again.

## Release

`refactor` is a hidden type, so no release is cut. Deliberate: this is a no-behaviour-change restructure and a 15-module change cannot be expressed in one release scope. Each service picks the shared bean up on its next release.

## Follow-up not done here

A CI guard forbidding a re-introduced per-service `TemporalClientProducer` belongs in `rules.yaml` — but editing `rules.yaml` restamps all ~64 OPA bundles, which would swamp this diff and give it a shelf life of hours. Worth its own small PR.


JiRaska:
## The first cut broke ~40 services. Fixed by splitting the module — pushed.

Worth writing up, because it is a trap with a general shape and the local gate could not have caught it.

**What happened.** Putting `TemporalConfig` + `TemporalClientProducer` in `openbank-libs-runtime` turned 8 `build (openbank-*)` jobs red — `account`, `agent`, `aml`, `analytics-sink`, `ap2`, `anacredit`, `audit` and counting, none of which touch Temporal:

```
Build step ArcProcessor#registerBeans threw an exception:
IllegalArgumentException: Producer method return type not found in index:
io.temporal.client.WorkflowClient
```

**Why.** `openbank-libs-runtime` is Jandex-indexed and *every* service consumes it, so Quarkus Arc resolves the type closure of every producer method in that index — including, for a service with no `temporal-sdk` on its classpath, a return type it cannot find. `compileOnly` keeps the dependency off consumers' **runtime classpaths**; it does not keep the producer out of the **index**. Those are different things and I conflated them.

**The rule**, now recorded in the new module's build file so the next person does not rediscover it: *a Jandex-indexed library every service consumes may only declare producers whose types every service also has.* A producer for an optional dependency belongs in a module only its users depend on.

**The fix.** The code moves verbatim into a new `openbank-libs-temporal` (Apache-2.0, same `agpl_modules.rule` reasoning), and the 14 consumers take an explicit `implementation(project(":openbank-libs-temporal"))`. No `settings.gradle.kts` edit — modules are auto-discovered from any `openbank-*` directory with a `build.gradle.kts`. Per-service Dockerfiles are unaffected: they are only grepped for their `EXPOSE` line, and images bake from Gradle fast-jars.

**Why my local gate missed it, and what I changed about it.** I ran `quarkusBuild` for the 14 services I touched, which is what the CDI-wiring rule says to do — and every one was green, because every one carries `temporal-sdk`. The affected set was *the entire fleet*, and the 14 were precisely the services that could not fail. Editing a shared, indexed library makes every consumer the blast radius, not just the callers you edited. After the split I re-ran `quarkusBuild` on all 14 **and** on all 8 services the first attempt broke, plus `ledger` and `mcp` as controls — 24 green — and `:openbank-audit-service:build`, the local reproducer that showed the exact Arc error, now passes.

Two upsides: the failure was loud, immediate and specific rather than a runtime surprise, and the split is the better structure anyway — `libs-runtime` is "plumbing every service gets", which Temporal is not.

Issue #2598, opened 2026-07-26
analytics-sink: subscribed events land as UNKNOWN/UNKNOWN/unknown — attributable data lost with a green pipeline


JiRaska:
Found while verifying Customer 360 (ADR-0210) against the sandbox rather than against fixtures.

## The defect
Events that analytics-sink **does** subscribe to arrive in `bronze_events` with `aggregate_type = 'UNKNOWN'`, `event_type = 'UNKNOWN'` and `source_service = 'unknown'` — while their payloads are plainly identifiable:

| payload keys (truncated) | what it obviously is | topic in the subscribe list |
|---|---|---|
| `deviceId`, `partyId`, `credentialId: "MFkwEwYHKoZIzj0CAQ…"` | a passkey / device credential | `openbank.sca.events` |
| `documentId`, `templateCode: "RAMCOVA_SMLOUVA_CS"`, `templateVersion`, `sha256` | a document generated | `openbank.documents.document.event` |
| `ceremonyId`, `documentId`, `at` | a signing ceremony step | `openbank.documents.document.event` |

Nothing errors. The rows land, the consumer stays healthy, the pipeline is green — and three columns of attribution are gone. An event that arrives but cannot be attributed is data loss that no alert will ever raise, because from the sink's point of view ingest succeeded.

## Reproduce
```sql
SELECT aggregate_type, event_type, source_service, substring(payload,1,150)
FROM openbank_analytics.bronze_events
WHERE aggregate_type = 'UNKNOWN' OR event_type = 'UNKNOWN';
```
On the sandbox today: 4 rows, all with identifiable payloads.

## Why it matters beyond tidiness
- **Any per-domain query silently under-reports.** Customer 360's domain table groups by `aggregate_type`, so a party's passkey enrolment shows as `unknown` rather than `sca`. The passkey row does carry `partyId`, so it *is* attributed to the right customer — just filed under a domain that does not exist.
- The `silver_current_state` / `silver_as_of` views group by `(aggregate_type, aggregate_id)`, so an UNKNOWN aggregate_type collapses unrelated events from different domains into one bucket.
- It is the same failure shape this repo documents repeatedly: the component cannot express its own failure, so it reports success.

## Second, separate finding in the same data
[...]

PR #2604, opened 2026-07-26
feat(admin-ui): search Customer 360 and Consents by name, not by party UUID

JiRaska:
## What this fixes

Two admin-UI pages exposed their backing store's key as the search box:

- **Customer 360** took a raw `Party UUID`.
- **Consents** defaulted to the grantee lens and offered "By party ID".

Neither is usable without a second tab open on the Parties page. This was reported as *"it does not work"* the first time Customer 360 was used on the sandbox — against a page that was answering correctly. The data model had been allowed to dictate the UX.

## How

`party-service` already owns a trigram name search: `GET /api/v1/parties/search?q=` (ADR-0055, feature flag `party-search` — verified **ENABLED** on the sandbox, `defaultVariant: on`). It is reached through the BFF (ADR-0056) from one shared `PartySearch` component, per ADR-0208: there are two callers, so a copy in each would be two divergent search behaviours over one endpoint.

It resolves a name to a party id, which is what each page queries with. A pasted UUID short-circuits straight to selection, so the operator who already has an id keeps the direct path. **`party-service` remains the only owner of a name index — this adds none.**

Consents now defaults to the **party** lens. The grantee lens stays, because `party-service:marketing-comms` is the one genuinely global view, with a one-click preset for it.

## Second defect, same investigation

A zero-row result rendered as `no_data`, whose copy reads *"the data source is available but does not contain any records yet"*. That was **false** whenever any other party had rows, and is indistinguishable from a broken page.

`available: false` now means one thing: ClickHouse did not answer. A party with no projected events returns `available: true` with empty domains and reads *"this party has no analytics events"*. Same correction for `consent-service` lookups by party and by grantee.

## Verification — every guard run against code it must flag

| Guard | Broken input | Result |
|---|---|---|
| UUID pass-through | disabled the short-circuit | ❌ red |
| empty-result copy | restored the misleading text | ❌ red |
| `available` semantics | pre-fix route | ❌ red (`expected false to be true`) |
| unreachable still unavailable | pre-fix route | ✅ green (ratchet guard) |

All green after the fix. **801/801 tests** (50 files), **0 lint errors** (80 pre-existing warnings, none in changed files), `tsc --noEmit` clean, build compiled. The page is covered by `render-smoke.test.tsx`, which auto-discovers via `import.meta.glob`.

## ADR

ADR-0210 gains two decisions:
- **D8** — the lookup key is not the search surface.
- **D9** — "the source is empty" and "this key has no rows" are different answers and must read differently.

`check-adr-registry.sh`: OK, 202 ADRs.

## Note on sandbox data

The sandbox has only 92 bronze events across 2 parties, and no `ACCOUNT_OPENED` / CONSENT / TRANSACTION rows ever reached bronze — that is [#2598](https://github.com/JiRaska/open-bank-oss/issues/2598), not this PR. After merge the page will render truthfully but sparsely. Search itself is unaffected: it reads `party-service`, which has the full party set.

Refs #2598

### DIFF
--- a/docs/adr/0210-customer-360-as-a-query-over-the-analytics-silver-layer.md
+++ b/docs/adr/0210-customer-360-as-a-query-over-the-analytics-silver-layer.md
@@ -95,6 +95,27 @@ values nor that parity. So ADR-0209 D4 stands unchanged: ADR-0201 D5 does not st
 phase 2 exists. Building segment training on `silver_as_of` would reintroduce precisely the skew
 ADR-0140 was written to prevent, while looking like it had satisfied the prerequisite.
 
+**D8 — The lookup key is not the search surface.** The 360 is keyed by partyId because the silver
+layer is keyed by aggregate id — but that key must never be what the operator types. The first
+implementation exposed a raw `Party UUID` field, which made the page unusable without a second tab
+open on the Parties page, and was reported as "it does not work" the first time it was seen on the
+sandbox. Names are resolved by `party-service`'s existing trigram `/search` (ADR-0055) through the
+BFF, in a shared `PartySearch` component (ADR-0208), which hands back an id; a pasted UUID
+short-circuits straight through, so the direct path survives. `party-service` stays the only owner of
+a name index — this adds none. Generalise: when a page's backing store is keyed by a synthetic id,
+the id is an implementation detail of the query, and letting it become the input is the data model
+dictating the UX.
+
+**D9 — "The source is empty" and "this key has no rows" are different answers and must read
+differently.** Both this route and the Consents page collapsed a zero-row result into the generic
+`no_data` state, whose copy reads *"the data source is available but does not contain any records
+yet"* — false whenever any other party had rows, and indistinguishable from a broken page for the
+operator who hits it. `available: false` now means only that ClickHouse did not answer; a party with
+no projected events returns `available: true` with empty domains and renders as "this party has no
+analytics events". The same correction applies to `consent-service` lookups by party and by grantee.
+Generalise: an empty-state message asserts a fact about the data source, so it is wrong to reuse the
+source-level message for a key-level miss.
+
 ## Alternatives considered
 
 - **Build `crm-service` as ADR-0199 specifies.** Rejected on evidence, not principle: it
--- a/openbank-admin-ui/src/app/api/customer-360/[partyId]/route.ts
+++ b/openbank-admin-ui/src/app/api/customer-360/[partyId]/route.ts
@@ -136,7 +136,12 @@ export async function GET(_req: NextRequest, ctx: { params: Promise<{ partyId: s
 
   try {
     const rows = await chQuery(scopedRowsSql(partyId))
-    if (rows.length === 0) return NextResponse.json(empty(partyId))
+    // A query that succeeded and matched nothing means THIS PARTY has no projected events — it does
+    // NOT mean the data source is empty. Those are different facts and the page renders different
+    // copy for each, so `available` stays true: ClickHouse answered. Returning false here read as
+    // "the source contains no records yet" while the source held events for other parties, which is
+    // the shape an operator cannot tell apart from a broken page.
+    if (rows.length === 0) return NextResponse.json({ ...empty(partyId), available: true })
 
     const byDomain = new Map<string, DomainSummary>()
     const accountIds = new Set<string>()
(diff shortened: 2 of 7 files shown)


JiRaska:
## Self-review found a real defect — and it was the one my own comment claimed to prevent

I reviewed this PR properly rather than trusting the green. A delegated reviewer returned "no issues" after a single tool call; I did not accept that, and the manual pass found this:

**Both pages cleared their results when the query changed. That handles the settled case only.** An already in-flight request settles *afterwards* and puts them straight back.

Measured with a 1.5 s-delayed `consent-service` party lookup interrupted by a lens switch:

```
RACE2_RESULT {"lensNow":"grantee","partyRowsVisible":true,"statCardsVisible":true}
```

The selector read **By grantee** while `MARKETING_COMMS_EMAIL` rows from the **party** query sat on screen — one customer's consents presented as the global marketing view. After the fix:

```
RACE2_RESULT {"lensNow":"grantee","partyRowsVisible":false,"statCardsVisible":false}
```

Every request now carries a generation number and commits only if still current; a lens switch bumps it, which is what actually cancels the flight. The parameter a request was built from (the lens) is read **once at request time**, never from the closure after the `await`.

Applied at all three call sites. Customer 360 was already serialised in practice because its Select buttons disable mid-load — I proved that with a probe rather than assuming it — but that is a UI accident, not a contract, and the failure it would hide is the worst one on that page: another party's data under this party's name.

### While fixing it, I also caught my own failed edit

The first patch of the lens-switch guard **did not apply** (the earlier `lensSelect` extraction had changed the indentation). `tsc` passed, the page compiled, and the guard was simply absent. Only re-running the probe caught it — the same probe printed the same red result twice, which is the only reason I looked instead of declaring it fixed.

## What else this PR now carries

| Item | Why |
|---|---|
| `e2e/party-lookup.spec.ts` (4 specs) | The race is invisible to unit tests. Verified **red** against unguarded code. E2E 9 → 13. |
| `ui-primitives.spec.ts` selects the lens explicitly | It broke here because it relied on the page default this PR deliberately changed. A spec depending on a default asserts two things and fails for the reason that is not its subject. |
| 20px padding on titled cards | `.card` defines **no** padding in `globals.css` (69 files supply it inline); headings rendered flush against the border. Verified by measuring the title's inset inside its card, not by eye. |
| `supersede-deploy-prs.sh` + both deploy workflows | See below. |

## The deploy-PR pile-up, fixed at the cause

`branch-cleanup.yml` already **assumes** this: *"a superseded deploy PR is CLOSED-unmerged (a newer deploy won the race)"*, and it deletes the branch on close. **Nothing ever performed that close.**

Measured today: **7 open admin-ui deploy PRs out of 23 open PRs in the whole repo** — 30% of the review queue, accumulated within three hours, growing by two while I wrote the script. All conflicting, because each rewrites the same manifest line. One of them (#2594) would have **reverted 851 lines of `main`** if merged to clear the backlog. That is what makes the pile dangerous rather than untidy.

The script closes every older open bot deploy PR when a newer one opens — the missing half of an existing design, not a new mechanism. With it exactly one deploy PR is open per prefix, always the newest, always based on current `main`, so the conflict class cannot arise.

Scoped deliberately: bot prefixes only (a hand-cut deploy PR is never touched), same-repo heads only, never the PR just opened, and it **comments before closing** — a silent close is not acceptable even for a bot PR.

It uses `while read`, **not `mapfile`**: `mapfile` is a bash-4 builtin absent from zsh and macOS bash 3.2, where it fails with *command not found* and leaves the array empty — so the script would print "nothing to close" and exit 0. I hit exactly that while testing, and the output looked like a clean result. Exercised via `DRY_RUN` against a known-positive (found 6 of 7) and a known-negative (found 0).

## Gate

801/801 unit · 13/13 E2E · 0 lint errors · `tsc --noEmit` clean · build compiled · `check-adr-registry` OK (202 ADRs). ADR-0210 gains D8, D9, D10.

Both pages were also rendered in a real browser and inspected — that step is what surfaced the card padding, and skipping it is what let three visual regressions through on #2556.

JiRaska:
**Correction: this PR merged commit 1 only. The comment above describes work that is NOT in it.**

Auto-merge was armed here, checks went green on `e2e7a1f0`, auto-merge fired on that SHA, and the second commit (`81eef93f`) was pushed moments later. The branch kept it; this PR merged and closed without it. `grep -c "generation.current"` against `origin/main` returns 0.

The race guard, `e2e/party-lookup.spec.ts`, the card padding and `supersede-deploy-prs.sh` are in #2612. What DID ship here is the name search itself and the `available` semantics fix — those are on `main` and deployed (`sandbox-eab0ff54`).

Leaving the earlier comment in place rather than editing it: it is the evidence for how this reads as complete when it is not.

PR #2629, opened 2026-07-26
fix(analytics-sink): consume Kafka topics that exist

JiRaska:
## The defect

`openbank-analytics-sink` subscribed to `openbank.account.events` and `openbank.transaction.events`. **Neither has ever existed.**

| Subscribed | Reality |
|---|---|
| `openbank.account.events` | ❌ no such topic — real name `openbank.accounts.account.created` |
| `openbank.transaction.events` | ❌ no such topic — real name `openbank.transactions.transaction.initiated` |

Both real topics are declared as `KafkaTopic` CRs, both are already consumed by `audit-service`, and both are produced by `account-service` / `transaction-service`. So audit has been receiving these events all along; analytics never has.

A Kafka consumer subscribed to a nonexistent topic **does not error — it receives nothing.** The sink ran healthy by every available signal: pod ready, consumer group joined, lag zero, metrics flat. Meanwhile **no `ACCOUNT_OPENED` and no `TRANSACTION` row ever reached `bronze_events`**, which silently made ADR-0210 D2's account→party resolution dead code and left Customer 360 structurally unable to show a customer's accounts or transactions.

Same shape as finrep's call to `/api/v1/ledger/trial-balance` — a path that never existed, green tests, broken feature.

This is also the honest correction to something I said earlier while debugging Customer 360: I blamed the missing account rows on SQL casing and fixed that twice. The casing was real but it was never the reason. **The events were not there.**

## Correcting the names makes a latent bug live

Which is why these ship in the same commit:

**1. `inferAggregateType` tested `accountId` first.** A transaction event carries *both* `transactionId` and `accountId`, so every transaction would have landed as `ACCOUNT` the moment the real topic was consumed. Reordered most-specific-first.

**2. `aggregateId` had its own precedence chain**, also preferring `accountId`. That pairs `aggregate_type=TRANSACTION` with an *account* id — and `bronze_events` is keyed `(aggregate_type, aggregate_id)` with `silver_current_state` reducing per that key, so **every transaction on one account would collapse into a single aggregate.** A corrupted read model no single-event test can see. The id is now resolved *from* the resolved type, via a helper kept adjacent so the two cannot drift.

**3. `documentId` / `credentialId` added.** The sandbox held generated-contract, signing-ceremony and passkey-registration events as `UNKNOWN/UNKNOWN/unknown` — identifiable events with three columns of attribution blank.

The inference stays a **heuristic and should not be the mechanism**: the topic name states the domain authoritatively. Reading it needs `@Incoming` to take `Message<String>` with explicit ack/nack, which changes the failure semantics of a working consumer and is not required to fix the misclassification. Deferred deliberately, documented in place — until it lands, any new event shape whose keys are unlisted still becomes `UNKNOWN` silently.

## The gate — making the class impossible

`.github/scripts/check-kafka-topics-exist.py`: every dotted topic name a service references must resolve to a `KafkaTopic` CR in gitops. **45 declared, 35 referenced, 0 unresolved.**

**Enforced from the start, not advisory.** At zero violations there is no judgement left to exercise, so an advisory verdict would only make the next one mergeable. `ALLOWED_ABSENT` is **empty**, and its stale-entry check fails in *both* directions — an exception cannot quietly become permanent.

## Everything verified against inputs it MUST flag

| Check | Input | Result |
|---|---|---|
| consumer tests | pre-fix consumer | **4 of 11 fail** |
| gate | old topic config restored | **exit 1**, naming both topics |
| gate | fixed config | exit 0 |
| `--selftest` | 4 cases, both directions | OK |
| `--selftest` | scan deliberately broken | **fails** |

`--selftest` runs on every CI invocation (the `check-pact-provider-replay.py` convention) so the flagging branch cannot rot once the fleet is clean — a gate that has only ever passed is unfalsified.

`:openbank-analytics-sink:test` **exit 0**, `detekt` + `ktlintCheck` **exit 0** — read from the exit code, not from grepped output, because `./gradlew … | grep` masks it. actionlint finding sets unchanged from `origin/main` (the one `shellcheck` warning is pre-existing, verified by diffing the sets).

## Deliberately not in this PR

The `rules.yaml` registration. Editing that file restamps ~65 OPA bundles, so bundling it would make this churn-heavy and conflict-prone while the substantive fix waits. Nothing requires it: `check-advisory-gate-registration.py` covers *advisory* gates, and this one is enforced.

## What this does not fix

`openbank.consent.events` **does exist** and is subscribed, yet bronze holds zero CONSENT rows. Either no consent event has been produced on the sandbox, or something downstream drops them — not diagnosed here, and it does not block this fix. Tracked under #2598.

Refs #2598

### DIFF
--- a/.github/scripts/check-kafka-topics-exist.py
+++ b/.github/scripts/check-kafka-topics-exist.py
@@ -1,0 +1,184 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: Apache-2.0
+"""Every Kafka topic a service consumes or produces must be a topic that EXISTS.
+
+Why this exists
+---------------
+`openbank-analytics-sink` was subscribed to `openbank.account.events` and
+`openbank.transaction.events`. Neither has ever existed — the real topics are
+`openbank.accounts.account.created` and `openbank.transactions.transaction.initiated`. A Kafka
+consumer subscribed to a nonexistent topic does not error; it simply receives nothing. So the sink
+ran for weeks, healthy by every signal, while no ACCOUNT_OPENED and no TRANSACTION row ever reached
+`bronze_events` — which silently made ADR-0210 D2's account->party resolution dead code, and made
+Customer 360 structurally unable to show a customer's accounts or transactions (#2598).
+
+Nothing was red. Nothing could have been: the consumer's own health, its consumer-group membership,
+its lag and the sink's metrics are all consistent with "this topic has no traffic".
+
+This is the same failure shape as a client pointed at a REST path that does not exist — the one that
+shipped finrep's call to `/api/v1/ledger/trial-balance` (see the contract-test notes in CLAUDE.md).
+There, the fix was to make the provider replay the contract at PR time. Here, the equivalent is to
+compare the name a service asks for against the set of names that are declared to exist.
+
+`openbank-infra/gitops` holds a `KafkaTopic` CR for every topic (42 declared, 42 on the sandbox
+cluster at the time of writing), so the declared set is authoritative and complete. That is what
+makes this checkable statically rather than needing a live broker.
+
+Usage:  check-kafka-topics-exist.py [--enforce]
+Advisory by default (prints ::warning, exits 0) per the repo convention; --enforce fails the build.
+"""
+
+from __future__ import annotations
+
+import argparse
+import pathlib
+import re
+import sys
+
+REPO = pathlib.Path(__file__).resolve().parents[2]
+GITOPS = REPO / "openbank-infra" / "gitops"
+
+# A channel's topic list. Both spellings appear in the fleet:
+#   topics: a,b,c          (multi-topic consumer)
+#   topic: a               (single)
+TOPIC_LINE = re.compile(r"^\s*topics?:\s*(?P<val>[^\s#][^#]*?)\s*(?:#.*)?$")
+
+# Anything templated is not a literal name and cannot be checked here.
+TEMPLATED = re.compile(r"[$&{}*?]|\bSTRING\b")
+
+# Topics whose existence this check cannot assert, with the reason. An entry here is a claim that
+# needs to stay true — not a place to park a failure.
+# EMPTY, and that is the point: with the #2598 topic names corrected, every topic any service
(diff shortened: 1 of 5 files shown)


JiRaska:
## Overlaps #2630 — both are needed, and they collide on one file

Measured, not assumed:

| | #2629 | #2630 |
|---|---|---|
| Topic-name fix (`openbank.account.events` → `openbank.accounts.account.created`) — **the root cause** | yes | no (0 `application.yaml` files touched) |
| `aggregateId` aligned with `aggregateType` — else `TRANSACTION` gets an account id and every transaction on one account collapses into a single aggregate in `silver_current_state` | yes | no |
| `check-kafka-topics-exist.py` CI gate | yes | no |
| Broker-metadata attribution (`Message<String>`) | no | yes |
| Metrics for events that still cannot be attributed | no | yes |

Only overlapping file: `openbank-analytics-sink/.../AnalyticsConsumer.kt`. Both have auto-merge armed, so whichever lands second goes DIRTY.

**#2630 carries the better mechanism.** I deferred exactly that refactor in #2629 and said so in the code: the topic name states the domain authoritatively, and payload-key sniffing means every new event shape silently becomes `UNKNOWN`. #2630 does it properly, so it should win on `AnalyticsConsumer.kt`.

**But #2630 does not fix the root cause.** The sink was subscribed to `openbank.account.events` and `openbank.transaction.events` and **neither topic has ever existed**. No attribution mechanism can classify an event that never arrives — for as long as the sink has run, no `ACCOUNT_OPENED` and no `TRANSACTION` row reached `bronze_events`, which silently made ADR-0210 D2's account→party resolution dead code.

### Suggested order

1. **#2630 first** — it rewrites the consumer's entry point, so it should not have to rebase onto a heuristic it replaces.
2. **#2629 rebased down to its unique content**: the two topic names, the `aggregateId`/type alignment (still needed — that one is about the *id*, not the type inference), and the CI gate. The `inferAggregateType` reorder can go if broker metadata already resolves those cases.

One trap for whoever goes second: **#2630 has a test that reads `application.yaml` and fails when a subscribed topic has no attribution**, and #2629 changes two of those names. In either order the new names need attributions, or that test goes red for the right reason at the wrong time.

PR #2630, opened 2026-07-26
fix(analytics-sink): attribute subscribed events from the broker metadata, and count the ones that still cannot be

JiRaska:
Refs #2598 (asks 1 and 2; ask 3 — the absent ACCOUNT_OPENED / CONSENT / TRANSACTION events — is a separate investigation and the issue stays open for it)

## Root cause

`@Incoming("analytics-events-in") suspend fun consume(payload: String)`.

An outbox-relayed record's **body** is `OutboxEntry.payload` — the bare domain event. The outbox's own addressing never enters that JSON:

| what | where it actually lives |
|---|---|
| event type | `ce-type` Kafka header (`OutboxKafkaHeaders.HEADER_EVENT_TYPE`) |
| aggregate id | the record **key** (`OutboxKafkaHeaders.partitionKey`) |
| producing domain | the **topic** |

A `String` signature discards all three before the mapping runs. So every producer that publishes a bare payload rather than a self-describing envelope lost its attribution, and the classifier's `inferAggregateType` — which only knows `accountId`/`partyId`/`transactionId`/`consentId`/`kycCaseId` — had nothing to work with and returned the `UNKNOWN` sentinel. The missing SCA/DOCUMENT branches are the symptom; the discarded metadata is the cause. Fixing only the branches would have left the same hole open for the next producer.

## Fix, half one — the classifier

* `consume` now takes `Message<String>` and lifts topic / key / `ce-type` into a new `EventAddress`. It acks explicitly in a `finally`: the signature change also switches SmallRye from auto-ack to manual, and a quarantined message would otherwise stall the partition forever.
* `TopicAttribution` **derives** the domain and service from the topic name (`openbank.<domain>.…`, ADR-0003 N3) rather than keeping a list of them. A hand-kept map would silently readmit `UNKNOWN` for the next topic anyone subscribes — the exact failure shape being fixed. Only three genuine exceptions are mapped (`documents`→`DOCUMENT`/`openbank-document-service`, `onboarding`, `feedback`).
* Fallback order is **body-first**, and that asymmetry is the point: the topic is consulted only where the body already yielded `UNKNOWN`, so the change is additive by construction — it can only turn an `UNKNOWN` into a value. Letting the topic win would refile `openbank.balance.events` from `ACCOUNT` to `BALANCE` and split one aggregate across two buckets in the silver views, which group by `(aggregate_type, aggregate_id)`. There is a test pinning exactly that.

The three sandbox shapes from the issue now resolve:

| body | topic | → |
|---|---|---|
| `deviceId`, `credentialId` | `openbank.sca.events` | `SCA` / `sca.device.enrolled` / `openbank-sca-service` |
| `documentId`, `templateCode` | `openbank.documents.document.event` | `DOCUMENT` / `document.generated` / `openbank-document-service` |
| `ceremonyId`, `documentId` | `openbank.documents.document.event` | `DOCUMENT` / `ceremony.step.completed` / `openbank-document-service` |

## Fix, half two — it stops being green about it

This is the half that prevents the next instance. The reason the defect survived is not the missing branch; it is that losing attribution cost **nothing observable** — the row landed, the consumer stayed healthy, and the broken state was indistinguishable from the healthy one, so no alert could ever have fired.

`IngestAttributionMetrics` now:

* increments `openbank_analytics_unattributed_total{field="aggregate_type"|"event_type"|"source_service"}` per unresolved field. The counter is **flat at zero in a healthy pipeline**, so `increase(...) > 0` is a usable alert expression rather than a threshold someone has to guess;
* logs one WARN per affected event naming the topic and the fields, so the offending producer is identifiable from logs without a ClickHouse query.

Deliberately **not** a quarantine. The event is real and its payload intact; diverting it to `dead_letter_events` would turn a partial-attribution problem into actual data loss. Counted-and-kept is the honest outcome.

(The service still has no gitops manifest — never deployed, no `version.txt` — so there is nowhere to land a Prometheus rule yet. The metric is named and shaped so that whoever wires the first manifest has an alert expression ready.)

## The tests fail first, for the stated reason

With only the three resolution chains reverted (types and metrics left in place, so the failure cannot be a compile error):

```
AnalyticsAttributionTest > a passkey enrolment on the sca topic is attributed to SCA, not UNKNOWN() FAILED
AnalyticsAttributionTest > a generated document is attributed to DOCUMENT and the document service() FAILED
AnalyticsAttributionTest > a signing-ceremony step is attributed rather than collapsed into the UNKNOWN bucket() FAILED
AnalyticsAttributionTest > the three sandbox shapes from the issue are all attributable end to end() FAILED
11 tests completed, 4 failed
```

Exactly the four attribution cases go red; the metrics and topic-derivation cases stay green, so they are not merely riding on the same change.

Two anti-vacuity measures in the suite itself:

* the "every subscribed topic resolves" sweep **derives its scope from `application.yaml`**, so adding a topic without an attribution for it fails here — a scope list kept separately from the thing it covers reads as *passing* when it is short, never as *unchecked*;
* a companion case asserts a non-conforming topic returns `null`, so that sweep cannot be green because the function returns something for everything.

## Verification

`./gradlew :openbank-analytics-sink:detekt ktlintCheck test quarkusBuild` — **BUILD SUCCESSFUL**. `quarkusBuild` included because a new `@ApplicationScoped` bean is added; it is service-local (no `openbank-libs` `@Produces`), so the blast radius is this module only.

`RecoveryFlowsTest` updated for the signature change (`Message.of(...)`); its assertions are untouched and it still proves a malformed message is quarantined rather than dropped.

### DIFF
--- a/.github/asvs-l3-baseline.txt
+++ b/.github/asvs-l3-baseline.txt
@@ -185,7 +185,8 @@ V9.1 openbank-infra/gitops/components/security-scanner/security-scanner-service.
 V9.1 openbank-infra/gitops/components/security-scanner/security-scanner-service.yaml:133: plaintext http:// env value http://sepa-instant.payments.svc:8127
 V9.1 openbank-infra/gitops/components/docs-truth-agent/docs-truth-agent.yaml:58: plaintext http:// env value http://litellm.ai-platform.svc:4000
 V9.1 openbank-infra/gitops/components/fraud-service/fraud-service.yaml:114: plaintext http:// env value http://tempo.observability.svc:4317
-V9.1 openbank-infra/gitops/components/notifications/notification-service.yaml:117: plaintext http:// env value http://tempo.observability.svc:4317
+V9.1 openbank-infra/gitops/components/notifications/notification-service.yaml:116: plaintext http:// env value http://consent-service.consent.svc:8106
+V9.1 openbank-infra/gitops/components/notifications/notification-service.yaml:122: plaintext http:// env value http://tempo.observability.svc:4317
 V9.1 openbank-infra/gitops/components/document-service/document-service-service.yaml:87: plaintext http:// env value http://sca-service.sca.svc:8110
 V9.1 openbank-infra/gitops/components/document-service/document-service-service.yaml:93: plaintext http:// env value http://product-catalog.accounts.svc:8104
 V9.1 openbank-infra/gitops/components/document-service/document-service-service.yaml:100: plaintext http:// env value http://party-service.party.svc:8111
--- a/openbank-analytics-sink/src/main/kotlin/com/openbank/analytics/application/AnalyticsConsumer.kt
+++ b/openbank-analytics-sink/src/main/kotlin/com/openbank/analytics/application/AnalyticsConsumer.kt
@@ -10,6 +10,7 @@ import com.openbank.analytics.application.port.out.AnalyticsSink
 import com.openbank.analytics.application.port.out.DeadLetterRecord
 import com.openbank.analytics.application.port.out.DeadLetterSink
 import com.openbank.libs.analytics.AnalyticsEnvelope
+import com.openbank.libs.domain.identifiers.Ids
 import com.openbank.libs.persistence.outbox.OutboxKafkaHeaders
 import io.smallrye.mutiny.Uni
 import io.smallrye.mutiny.coroutines.awaitSuspending
@@ -161,7 +162,8 @@ class AnalyticsConsumer {
         val aggregateType = resolveAggregateType(node, address)
         return AnalyticsEnvelope(
             eventId = node["eventId"]?.asText()?.let { runCatching { UUID.fromString(it) }.getOrNull() }
-                ?: UUID.randomUUID(),
+                    // ADR-0106: a synthesised dedupe key is a durable, indexed identifier -> UUIDv7.
+                ?: Ids.newId(),
             aggregateType = aggregateType,
             // The id MUST agree with aggregateType — bronze_events is keyed (aggregate_type,
             // aggregate_id) and silver_current_state reduces per that key. An independent precedence
@@ -190,6 +192,7 @@ class AnalyticsConsumer {
             payload = PayloadMasker.maskToMap(node["payload"] ?: node),
         )
     }
+
 
     private fun resolveAggregateType(node: JsonNode, address: EventAddress): String = node["aggregateType"]?.asText()
         ?: inferAggregateType(node).takeIf { it != UNKNOWN }
--- a/openbank-analytics-sink/src/main/kotlin/com/openbank/analytics/application/AnalyticsConsumer.kt
+++ b/openbank-analytics-sink/src/main/kotlin/com/openbank/analytics/application/AnalyticsConsumer.kt
@@ -165,17 +165,7 @@ class AnalyticsConsumer {
                     // ADR-0106: a synthesised dedupe key is a durable, indexed identifier -> UUIDv7.
                 ?: Ids.newId(),
             aggregateType = aggregateType,
-            // The id MUST agree with aggregateType — bronze_events is keyed (aggregate_type,
-            // aggregate_id) and silver_current_state reduces per that key. An independent precedence
-            // chain here (it used to prefer accountId over transactionId) would pair
-            // aggregate_type=TRANSACTION with an ACCOUNT id, collapsing every transaction on one
-            // account into a single aggregate in silver — a corrupted read model that no test over
-            // one event can see. Resolve the id FROM the resolved type instead.
-            aggregateId = node["aggregateId"]?.asText()
-                ?: idForType(aggregateType, node)
-                // The outbox partition key IS the aggregate id (OutboxKafkaHeaders.partitionKey).
-                ?: address.key
-                ?: UNKNOWN_SERVICE,
+            aggregateId = resolveAggregateId(node, aggregateType, address),
             aggregateVersion = resolveAggregateVersion(node),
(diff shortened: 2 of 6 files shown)


JiRaska:
## Overlaps #2629 — both are needed, and they collide on one file

Measured, not assumed:

| | #2629 | #2630 |
|---|---|---|
| Topic-name fix (`openbank.account.events` → `openbank.accounts.account.created`) — **the root cause** | yes | no (0 `application.yaml` files touched) |
| `aggregateId` aligned with `aggregateType` — else `TRANSACTION` gets an account id and every transaction on one account collapses into a single aggregate in `silver_current_state` | yes | no |
| `check-kafka-topics-exist.py` CI gate | yes | no |
| Broker-metadata attribution (`Message<String>`) | no | yes |
| Metrics for events that still cannot be attributed | no | yes |

Only overlapping file: `openbank-analytics-sink/.../AnalyticsConsumer.kt`. Both have auto-merge armed, so whichever lands second goes DIRTY.

**#2630 carries the better mechanism.** I deferred exactly that refactor in #2629 and said so in the code: the topic name states the domain authoritatively, and payload-key sniffing means every new event shape silently becomes `UNKNOWN`. #2630 does it properly, so it should win on `AnalyticsConsumer.kt`.

**But #2630 does not fix the root cause.** The sink was subscribed to `openbank.account.events` and `openbank.transaction.events` and **neither topic has ever existed**. No attribution mechanism can classify an event that never arrives — for as long as the sink has run, no `ACCOUNT_OPENED` and no `TRANSACTION` row reached `bronze_events`, which silently made ADR-0210 D2's account→party resolution dead code.

### Suggested order

1. **#2630 first** — it rewrites the consumer's entry point, so it should not have to rebase onto a heuristic it replaces.
2. **#2629 rebased down to its unique content**: the two topic names, the `aggregateId`/type alignment (still needed — that one is about the *id*, not the type inference), and the CI gate. The `inferAggregateType` reorder can go if broker metadata already resolves those cases.

One trap for whoever goes second: **#2630 has a test that reads `application.yaml` and fails when a subscribed topic has no attribution**, and #2629 changes two of those names. In either order the new names need attributions, or that test goes red for the right reason at the wrong time.

JiRaska:
**Update: #2629 merged first** (`488d476e`), so the order ended up the reverse of what I suggested above. This PR now needs a rebase onto `main`. Exactly what changed under you, so the rebase is mechanical:

**1. Two subscribed topic names are different.** `application.yaml` now reads:

```
openbank.accounts.account.created        (was openbank.account.events   — never existed)
openbank.transactions.transaction.initiated  (was openbank.transaction.events — never existed)
```

Your attribution test reads `application.yaml` and fails when a subscribed topic has no attribution, so **both new names need an entry** or it goes red for the right reason at the wrong time. The old names must NOT be re-added: `.github/scripts/check-kafka-topics-exist.py` is now on `main` and enforced, so either name would fail the build with `MISSING openbank.account.events`.

**2. `AnalyticsConsumer.kt` gained two things** — take a view on each rather than resolving the conflict mechanically:

- `inferAggregateType` was reordered most-specific-first, plus `documentId` / `credentialId`. **Your broker-metadata mechanism supersedes this** — if attribution resolves those cases from the topic, drop the heuristic rather than carrying both. Two sources for one answer is how they drift.
- `idForType(type, node)`, resolving `aggregateId` **from** the resolved type. **Keep this regardless.** It is not about type inference: the previous independent precedence chain preferred `accountId`, which pairs `aggregate_type=TRANSACTION` with an account id, and `bronze_events` is keyed `(aggregate_type, aggregate_id)` with `silver_current_state` reducing per that key — so every transaction on one account collapses into a single aggregate. If your mechanism sets the type from the topic, the id still has to follow it.

**3. Six new tests** in `AnalyticsConsumerTest.kt` covering the above. Four of them fail against the pre-#2629 consumer, so they are real constraints, not fixtures — if your mechanism changes the expected `aggregateType` for any of those payload shapes, change the assertion deliberately and say why.

Sorry for the collision — my PR was open before yours and I did not check for parallel work on this file before pushing, which is the actual mistake here.

JiRaska:
## Rebase na origin/main (2026-07-29) — semantic merge decisions

PR byl CONFLICTING, protože main mezitím landnul #2629 (`aa8c20ca0`), který řeší #2598 přes rozšířené body heuristiky, zatímco tato větev řeší #2598 přes broker-metadata attribuci. Obě opravy vznikly paralelně (~6 minut od sebe). Merge zachovává **oba** záměry:

1. **Zachována struktura větve** — `toEnvelope(node, address)` s `EventAddress` (topic/partition-key/ce-type), body-first pořadí fallbacků (deliberate, viz KDoc: topic-first by rebucketoval správně přiřazené eventy).
2. **Aplikován invariant z #2629** — `aggregateId` se řeší Z resolved typu (`idForType`), ne nezávislým precedence řetězcem. Branch původně měla `resolveAggregateId` s accountId-first chain — přesně ten bug, který #2629 opravuje. Partition-key fallback (`address.key`) zůstává jako poslední krok.
3. **Test aktualizován** — `a passkey enrolment...`: #2629 naučil heuristiku `credentialId → PASSKEY`, takže body-first ordering resolve PASSKEY z body (topic se nekonzultuje) a id je `credentialId` (= bucketing, který už žije v bronze). Původní assertion (SCA + partition key) by rebucketovala produkční data — přesně to, co branch KDoc zakazuje.
4. **Druhý commit** (`Ids.newId` místo `UUID.randomUUID`, ADR-0106) aplikován beze změny záměru.

Ověřeno lokálně: `:openbank-analytics-sink:test` — BUILD SUCCESSFUL (včetně E2E metrics testu `the three sandbox shapes from the issue are all attributable end to end`).

Issue #2833, opened 2026-07-31
dependency-review diffs the whole graph as additions: the base SHA almost never has a submitted snapshot


JiRaska:
## Summary

`dependency-review` compares two **submitted** snapshots, looked up by exact SHA. `dependency-submission.yml` is `paths:`-filtered on manifest files for **both** its `push: [main]` and `pull_request` triggers. That filter is correct for the head SHA and wrong for the base SHA, because the two are selected by different rules:

- **head SHA** — a PR only reaches `dependency-review` with a manifest change, so a head snapshot almost always exists.
- **base SHA** — `github.event.pull_request.base.sha` is an arbitrary `main` commit, which has a snapshot only if *that exact commit* also touched a manifest.

Measured on `origin/main`: **1 of the last 50 commits** touches any of `**/build.gradle.kts`, `openbank-libs/gradle/libs.versions.toml`, `settings.gradle.kts`, `gradle/wrapper/**`. So ~98% of possible base SHAs can never have a snapshot.

Result: the action falls back to diffing the **entire head graph as additions** and re-evaluates every pre-existing transitive as if the PR introduced it.

```
The number of snapshots compared for the base SHA (0) and the head SHA (1) do not match.
You may see unexpected additions in the diff.
```

## Evidence

Reproduced on PR #2751 (adds `openbank-campaign-service/build.gradle.kts`), two independent runs:

| run | head SHA | outcome |
|---|---|---|
| [30619408381](https://github.com/JiRaska/open-bank-oss/actions/runs/30619408381) | `33334a2db` | `##[error]Dependency review detected vulnerable packages.` |
| [30622152705](https://github.com/JiRaska/open-bank-oss/actions/runs/30622152705) | `958e8ea6c` | same |

Both logged the `(0) and (1) do not match` line on every retry poll, then `Retry timeout exceeded. Proceeding...` after burning the full `retry-on-snapshot-warnings-timeout: 1800`.

Base SHA is `add0b5e83f20bc7c166df803ca5d55d0f90c6c4c`. It touches no manifest, and `dependency-submission.yml` has **0** runs for it — the retry could never have succeeded, because nothing was ever going to submit that SHA.

[...]

PR #2751, opened 2026-07-30
feat(campaign): campaign-service first slice — deterministic segments, consent-gated Temporal journeys (ADR-0200/0209 D3)

JiRaska:
## Summary

Implements the ADR-0209 D3 first slice of campaign management: a new `openbank-campaign-service` with rule-based versioned segments, per-enrolment Temporal journeys, both ADR-0200 D2 consent mechanisms, D6 suppression, and delivery exclusively through notification-service — plus the first MARKETING catalogue template that makes a marketing send possible at all. No model, no NBA (blocked per ADR-0209 D4 until ADR-0140 phase 2). ADR-0209 D1 (Temporal→libs) was already shipped by #2572 and this service consumes it.

## Type of change

- [x] feat (new functionality)

## Linked issues

Refs #2749 (rollout tracker: gitops component, adverse-state inputs, Studio, ContactPolicyGate)

## Changes

**New service `openbank-campaign-service` (0.1.0, release-please registered):**
- `Campaign` aggregate: `DRAFT → PENDING_APPROVAL → ACTIVE ⇄ PAUSED → CLOSED`; maker/checker as a **domain invariant** (`approver != createdBy`), journeys capped at 5 steps
- `CampaignJourneyWorkflow` per enrolment: workflow id `campaignId+partyId` = idempotency key; steps as activities; `Workflow.sleep` delays with **signal-aware await** (revocation reacts mid-sleep, not at next step)
- **ADR-0200 D2, both mechanisms**: live per-send consent pull (consent-service `party/{id}/grantee/party-service:marketing-comms/active?scope=MARKETING_COMMS_EMAIL`, fail-closed) + `openbank.consent.events` consumer signalling live journeys on `ConsentRevoked`/`ConsentExpired` with a `MARKETING_COMMS_*` scope
- **ADR-0200 D6**: suppression before consent — per-party weekly cap from durable `send_log` + quiet hours (Europe/Prague, midnight-wrapping)
- **ADR-0200 D3**: send = `NotificationRequest` onto `openbank.notification.requests`; no delivery credentials, no recipient PII in the service
- **ADR-0201 D1/D2**: segments as versioned artifacts; closed typed rule DSL rendered to parameterised ClickHouse SQL over the ADR-0210 silver layer — **values never interpolate into query text** (injection test in `SegmentEvaluationTest`); unreachable analytics ⇒ empty cohort
[...]

PR #2848, opened 2026-07-31
fix(ci): diff dependency-review against the merge-base, not pull_request.base.sha

JiRaska:
Second half of the fix for #2833. #2837 (merged) was the first half; on its own it is **necessary but not sufficient**.

## The two conditions

A correct dependency delta needs the base to **carry a submitted snapshot** *and* **be an ancestor of the head**. The original diagnosis in #2833 only recognised the first, and I overstated #2837's completeness on that basis — [corrected here](https://github.com/JiRaska/open-bank-oss/issues/2833#issuecomment-5142890751).

Measured against head `33334a2db` (PR #2751), same head throughout:

| base | snapshot | ancestor | entries |
|---|---|---|---|
| `98fc64c9c` | yes | **yes** | **29** — the true delta |
| `27e83b42c` | yes | no | 1235 — whole graph |
| `add0b5e83` — the real `base.sha` | no | no | 1235 — whole graph |
| `51364286f` — the merge-base | no | yes | 1235 — whole graph |

**Row 2 is the point.** Ancestry alone breaks it, with a perfectly good snapshot present. When the basis is invalid the action silently falls back to reporting the entire head graph as additions, re-flagging every pre-existing transitive as if this PR introduced it — which is how the `netty` and `protobuf` entries in `allow-ghsas` were accreted as permanent fleet-wide suppressions.

`pull_request.base.sha` is the tip of the base branch, so for any PR that is not freshly rebased it is not an ancestor of the head. That is the common case, not an edge case.

## Not the `base-ref` that was already rejected

#2833 tested `base-ref: main` and ruled it out. This is different: `main` resolves to a *tip*, which is not an ancestor of an unrebased head — so it failed for the same reason `base.sha` does. A **merge-base is an ancestor by construction**. The earlier negative result does not transfer.

The two halves compose: #2837 is what makes the merge-base reliably snapshot-bearing from here on.

## Implementation notes

Resolved through the compare API rather than `git merge-base`, because `actions/checkout` is shallow in this job and deepening until the merge-base appears is unbounded. Verified the API agrees with git on a real pair — `.merge_base_commit.sha` for `add0b5e83...33334a2db` returns `51364286f`, byte-identical to `git merge-base`.

`head-ref` is pinned to the value the action already defaults to on `pull_request` — the PR head SHA, which is what the Gradle snapshot is keyed on (the failing run's log named exactly that SHA: `No snapshots were found for the head SHA 33334a2db...`). It makes the pair explicit; it is not a behaviour change.

## Verification

- **Both branches of the new step exercised, not just the happy one.** A real SHA pair resolves and writes the merge-base to `GITHUB_OUTPUT`; an unresolvable base emits the `::warning` and falls back to `base.sha` **without failing the step** (exit 0 confirmed). Failing closed would block every PR on a transient API hiccup; staying silent would reintroduce the #2833 bug one level up — so it warns.
- `actionlint` — clean.
- `yamllint` under the exact config `ci.yml` builds — clean on branch **and** on `origin/main`'s copy. Probe validated: 99 findings under default config, so the clean run is not a silent no-op.
- Parsed the result rather than eyeballing the diff: step order is `checkout → Detect dependency-manifest changes → Resolve the merge-base → Dependency Review`, with `base-ref: ${{ steps.mb.outputs.sha }}`.

## What is NOT yet verified — read before merging

**This PR touches no dependency manifest, so the gate passes trivially on it and proves nothing about the change.** Green here means "not exercised" — exactly the vacuous-green shape #2833 is about.

End-to-end proof needs a manifest-touching PR whose **merge-base is a post-#2837 main commit**. No such PR exists yet, and the two obvious candidates are both ruled out:

- **#2751 is already MERGED** (its branch is deleted), so it can no longer be rebased. An earlier revision of this description named it as the verification candidate; that is no longer possible.
- **#2843** does touch a manifest (`openbank-mcp-service/build.gradle.kts`) and its head has a snapshot, but its merge-base `6a9b9f7c5` landed at 12:14:18Z — **four minutes before** #2837 merged at 12:18:17Z — so it has no snapshot. Measured, both ways: `base.sha...head` = 1235 entries, `merge-base...head` = 1235 entries. The merge-base half cannot help while the snapshot half is missing for that commit.

So the next manifest-touching PR branched from current `main` is the real test. What it must show: a small delta (tens of entries) instead of ~1235. If it still reports the whole graph, this change did not do what it claims.

### Evidence that the precondition now holds

All 6 main commits since #2837 have a `dependency-submission` run, and **none of them touches a manifest** — under the old filter every one would have had zero:

```
2b3cbe66a  completed/success   manifest=0
bebbf84ae  in_progress         manifest=0
57d293801  in_progress         manifest=0
9d84e02b8  in_progress         manifest=0
b325aee34  in_progress         manifest=0
7915cdf73  in_progress         manifest=0
```

Five running concurrently with none cancelling another also validates #2837's per-SHA concurrency change — under the old shared-group rule with `cancel-in-progress`, four of those five would have been killed.

## Still open on #2833

Items 4 and 5 are untouched here: the mismatch still proceeds silently, and the submitted graph still records pre-resolution coordinates that `resolutionStrategy` pins away. Measured on #2751, a correct base cuts findings from ~30 to ~2, **not to 0**.

## Release impact

None. `.github/workflows/` is not under any released package path.

Refs #2833

### DIFF
--- a/.github/workflows/dependency-review.yml
+++ b/.github/workflows/dependency-review.yml
@@ -60,11 +60,64 @@ jobs:
             echo "No dependency manifest touched — no Gradle submission runs for this PR, so nothing to wait for."
           fi
 
+      # Diff against the MERGE-BASE, not `pull_request.base.sha` (issue #2833).
+      #
+      # A correct delta needs the base to satisfy BOTH conditions: carry a submitted
+      # snapshot, and be an ancestor of the head. `base.sha` is the tip of the base
+      # branch, so for any PR that is not freshly rebased it is neither — and when the
+      # basis is invalid the action silently falls back to reporting the ENTIRE head
+      # graph as additions, re-flagging every pre-existing transitive as if this PR
+      # introduced it.
+      #
+      # Measured 2026-07-31 against head 33334a2db (PR #2751), same head throughout:
+      #   base 98fc64c9c  snapshot=yes  ancestor=yes  ->    29 entries (the true delta)
+      #   base 27e83b42c  snapshot=yes  ancestor=NO   ->  1235 entries (whole graph)
+      #   base add0b5e83  snapshot=NO   ancestor=NO   ->  1235 entries  <- the real base.sha
+      # Ancestry alone is enough to break it: row 2 has a perfectly good snapshot.
+      #
+      # #2837 fixed the snapshot half by submitting a graph for every main commit, which
+      # is what makes the merge-base reliably snapshot-bearing from now on. This step is
+      # the other half. NOTE this is NOT the `base-ref: main` that was tried and rejected
+      # in #2833 — `main` resolves to a tip (not an ancestor); a merge-base is an ancestor
+      # by construction.
+      #
+      # Resolved via the compare API rather than `git merge-base`: actions/checkout is
+      # shallow here, and deepening until the merge-base appears is unbounded. Verified
+      # the API agrees with git on a real pair — `.merge_base_commit.sha` for
+      # add0b5e83...33334a2db returns 51364286f, identical to `git merge-base`.
+      - name: Resolve the merge-base to diff against
+        id: mb
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
+          HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+        run: |
+          set -euo pipefail
+          # Fall back to base.sha rather than failing the gate: a resolution hiccup should
+          # degrade to today's (noisier) behaviour, not block every PR. The warning is the
+          # signal that the diff is untrustworthy — silence here would be the #2833 bug
+          # reintroduced one level up.
+          if mb="$(gh api "repos/${GITHUB_REPOSITORY}/compare/${BASE_SHA}...${HEAD_SHA}" \
+                    --jq '.merge_base_commit.sha' 2>/dev/null)" && [ -n "${mb}" ]; then
+            echo "Merge-base of ${BASE_SHA} and ${HEAD_SHA}: ${mb}"
+          else
+            mb="${BASE_SHA}"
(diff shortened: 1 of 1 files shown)


JiRaska:
Closing — **this change is a no-op and rests on a premise I got wrong.** Probe #2862 was built specifically to confirm it and falsified it instead.

## The measurement

#2862 was constructed so that ancestry was the only variable: branched from an older commit so `base.sha` (main's tip) differed from the merge-base, with **both** carrying a snapshot. Predicted 1235. **Got 0**, with no `Retry timeout exceeded`.

Re-running every case with the merge-base computed explicitly:

| base | ancestor of head | merge-base | merge-base has snapshot | entries |
|---|---|---|---|---|
| `add0b5e83` | no | `51364286f` | 0 | 1235 |
| `27e83b42c` | no | `51364286f` | 0 | 1235 |
| `98fc64c9c` | **yes** | `98fc64c9c` | 1 | 29 |
| **`9a725ec72`** | **no** | `d9553237c` | **1** | **0** |
| `d9553237c` | yes | `d9553237c` | 1 | 0 |

Row 4 settles it: a base that is **not** an ancestor still produces a correct delta, because its merge-base has a snapshot.

## What was actually going on

`GET /dependency-graph/compare/{base}...{head}` uses **three-dot** semantics — GitHub resolves the merge-base itself. So the action was always comparing against the merge-base; passing it explicitly as `base-ref` changes nothing.

My earlier "two necessary conditions" analysis was wrong. There is **one** condition: *the merge-base must carry a submitted snapshot*. Every non-ancestor base I had tested happened to have a merge-base without a snapshot, so ancestry and the missing snapshot moved together and I attributed the effect to the wrong one.

## What this means for the real fix

**#2837 was the whole fix, not half of it.** By submitting a graph for every main commit it guarantees the merge-base is snapshot-bearing, which is the only thing that was ever required. It is verified end-to-end twice now — probe #2854 (0 entries, `base.sha == merge-base`) and probe #2862 (0 entries, `base.sha ≠ merge-base`).

Correcting #2833 accordingly. Nothing to merge here.

Refs #2833

PR #2854, opened 2026-07-31
test(ci): end-to-end probe for the dependency-review merge-base fix (#2833)

JiRaska:
**Draft, and intended to be CLOSED rather than merged** unless the comment it adds is judged worth keeping on its own. This exists to falsify #2848, which I have so far only verified in pieces.

## Why a probe was needed

Every earlier verification attempt failed for a different reason:

- **#2751** — already MERGED, branch deleted, cannot be rebased.
- **#2843** — touches a manifest and its head has a snapshot, but its merge-base `6a9b9f7c5` landed at 12:14:18Z, **four minutes before** #2837 merged at 12:18:17Z, so that commit has no snapshot. Measured both ways: `base.sha...head` = 1235, `merge-base...head` = 1235.
- **#2848 itself** — touches no manifest, so the gate passes trivially and proves nothing.

## Why this one is decisive

The change is **comment-only**, so the resolved dependency graph is bit-identical to its parent's. That removes the judgement call that made every other candidate weak:

| behaviour | expected entries |
|---|---|
| correct | **0** — nothing changed |
| broken | **1235** — whole head graph reported as additions |

No middle ground to argue about.

Branch point (the expected merge-base): `73c48273d`, a post-#2837 main commit, so it carries a snapshot.

## What each stage tests

1. **Now, with #2848 unmerged and `base.sha == merge-base`** — this exercises the #2837 half only. Expect 0 entries. If it reports 1235, then #2837 did not deliver what it claims and #2848 rests on a false premise.
2. **After main advances and this branch is pushed again**, `base.sha` moves to the new main tip while the merge-base stays at the branch point — they diverge. That is the state that isolates #2848: without it 1235, with it 0.

Stage 2 is the one that actually tests the merge-base change. Stage 1 is a precondition check.

## Cost

One ~11 min fleet resolution, $0 on hosted runners. Worth stating plainly: the comment this PR adds documents exactly that cost, which is the one genuinely reusable thing here.

Refs #2833

### DIFF
--- a/openbank-simulation/build.gradle.kts
+++ b/openbank-simulation/build.gradle.kts
@@ -16,6 +16,12 @@
 // rung. It exercises the domain + application semantics faithfully (built on the real
 // `openbank-libs` primitives — Money, SagaStateMachine); it cannot catch JVM-threading
 // or OS-level non-determinism (that is the deferred Antithesis option).
+//
+// CI note: editing ANY `build.gradle.kts` — including a comment like this one, which
+// changes no dependency at all — matches the `**/build.gradle.kts` path filter in
+// dependency-submission.yml and dependency-review.yml, so it triggers a full ~11 min
+// fleet dependency resolution. That is deliberate (the filter cannot inspect intent),
+// but it means a comment-only touch here is not free. See issue #2833.
 
 plugins {
     alias(libs.plugins.kotlin.jvm)


JiRaska:
## Stage 1 result: PASS

`dependency-review` on this PR concluded **success** with **0 added packages**.

Side by side with the original broken run on #2751, read from the job logs rather than the check conclusion:

| | probe #2854 | #2751 (broken) |
|---|---|---|
| added packages listed | **0** | **1235** |
| `Retry timeout exceeded` | **0** | 1 |
| `snapshots compared for the base SHA (0)...` | 11 | 40 |
| verdict | did not detect any vulnerable packages | **detected vulnerable packages** |

The comment-only change means the graph is bit-identical to its parent's, so 0 is the only correct answer. It got 0.

### The mismatch warning still appears — and that is fine

Worth writing down, because it reads alarming and I initially misread it as the gate passing without doing a diff.

The `base SHA (0) and the head SHA (1)` line appears 11 times, but it is **transient polling output**, not the final state. Timing:

```
base 73c48273d submission:  12:49:52Z -> 13:00:19Z   (10.5 min)
head 09bf5c240 submission:  12:51:18Z -> 12:57:51Z   ( 6.5 min)
dependency-review:          12:58:04Z -> 13:00:19Z
```

The review started after the head snapshot existed but while the base snapshot was still being produced. `retry-on-snapshot-warnings` polled for ~2 min, the base snapshot landed at 13:00:19Z, and the final comparison used it — hence 0 additions and no `Retry timeout exceeded`.

That is exactly the behaviour #2837 was supposed to buy. Before it, the base snapshot would never appear, the retry would exhaust its full 1800 s window, and the action would proceed on an invalid basis — which is precisely what #2751's log shows (`Retry timeout exceeded. Proceeding...`, then 1235 additions).

**So the distinguishing signal is not the warning, it is whether the retry converges.** A future reader debugging this should compare `Retry timeout exceeded` counts, not warning counts.

### What this does NOT prove

`base.sha == merge-base` on this PR (both `73c48273d`), because the branch was cut from main and no further commit landed before the review ran. So this validates **#2837's half only**.

#2848's merge-base change is still unexercised. To isolate it, `base.sha` must diverge from the merge-base: let main advance, push another commit here so the `synchronize` event carries a newer `base.sha`, and re-read the result. Without #2848 that should regress to 1235; with it, stay at 0.

JiRaska:
Closing — this probe did its job and should not be merged.

## What it established

1. **#2837 works end-to-end in a real CI run.** Comment-only manifest change, so the only correct answer was 0 added packages, and that is what the run produced (vs 1235 on #2751's broken run). No `Retry timeout exceeded`.
2. **The distinguishing signal is retry convergence, not the warning.** The `base SHA (0) and head SHA (1)` line still appears 11× as transient polling output while the base snapshot is being produced (~10 min after its main commit lands). Judging by that warning is misleading — I misread it that way myself. Compare `Retry timeout exceeded` counts instead.
3. **`pull_request.base.sha` is frozen at PR creation.** Pushing an empty commit did NOT move it off `73c48273d` across the `synchronize` event, confirming the behaviour documented for #481 × #524. Consequence: for a freshly-cut branch `base.sha` equals the merge-base permanently, so this probe structurally could not isolate #2848 — which is why #2862 exists, branching from an older commit so the divergence is built in at creation time.

Point 3 is the one worth carrying forward: it means #2848's value scales with how long a PR has been open. On a branch cut minutes ago it changes nothing; on a long-lived branch `base.sha` is arbitrarily far from the merge-base.

## Not merging

The one reusable artefact here is the comment noting that editing any `build.gradle.kts` — even a comment — triggers a ~11 min fleet resolution. That is true and mildly useful, but not worth a commit on its own; if it is wanted it should land in `CLAUDE.md` or alongside the path filter it describes, not buried in one service's build file.

Follow-up measurement continues on #2862. Refs #2833.

Issue #2841, opened 2026-07-31
Spot-kill auto-retry never fires when the reclaim takes only part of a multi-job run (run concludes failure, not cancelled)


JiRaska:
## Summary

`auto-retry-cancelled.yml` (built for #2330) fires only on a **run-level** `cancelled` conclusion:

```yaml
if: >-
  github.event.workflow_run.conclusion == 'cancelled' &&
  github.event.workflow_run.run_attempt == 1
```

A spot reclaim that kills **one job inside a multi-job workflow** does not produce that. The killed job concludes `failure`, and one failed job makes the whole run conclude `failure` — so the retry never matches, and the reclaim still costs a human `gh run rerun`.

Both watched multi-job lanes are exposed: `security.yml` has 6 jobs, `services-ci.yml` fans out per service.

## Measured

**The spot kill that is missed** — `security.yml` run `30627400819`, attempt 1, on PR #2837:

| level | conclusion |
|---|---|
| step `Build (java-kotlin) for CodeQL tracing` | `cancelled` |
| job `CodeQL (java-kotlin, manual)` | **`failure`** |
| run `Security scan` | **`failure`** |

```
##[error]The runner has received a shutdown signal. This can happen when the runner service is
stopped, or a manually started runner is canceled.
##[error]The operation was canceled.
```

The other five jobs (`Trivy`, `CodeQL (javascript-typescript)`, `Detect dependency changes`, two skipped SBOM jobs) succeeded, so the run concluded `failure` and the retry's `if:` never matched. Confirmed no retry run exists for that SHA; it was re-run by hand.

**What the retry actually fired on instead** — the three most recent `cancelled` `security.yml` runs, all on `feat/campaign-gitops`, all attempt 1. Their logs contain `The operation was canceled` and **no shutdown signal**: these are ordinary concurrency supersession from repeated pushes to the same branch, not spot kills.

So in the sampled window the mechanism retried the harmless case (which the workflow's LOOP SAFETY comment explicitly accepts as wasted) and missed the case it was built for.

## Scope of the claim

Being explicit about what is and is not established:

- **Established:** a reclaim taking a *subset* of a multi-job run yields run-level `failure` and is not retried. Directly observed, n=1, with logs.
[...]

PR #2898, opened 2026-07-31
fix(ci): pass -R to gh run rerun — the auto-retry had never once succeeded

JiRaska:
Found by watching for the first live firing of #2892. It fired — and exposed that **the auto-retry from #2330 has never once succeeded.**

## The bug

The job has no `actions/checkout`, so `gh` has no git repository to infer the target repo from:

```
Re-running cancelled run https://github.com/JiRaska/open-bank-oss/actions/runs/30640099943
failed to determine base repo: failed to run git: fatal: not a git repository
##[error]Process completed with exit code 1
```

Every invocation of `gh run rerun` exited 1 at the first line that mattered.

## Not a regression from #2892

Run [30640432089](https://github.com/JiRaska/open-bank-oss/actions/runs/30640432089) at **14:52Z** — five minutes *before* #2892 merged at 14:57Z, on the old workflow — failed on the same line with the same message. The bug is as old as the workflow.

Nothing caught it because **a `workflow_run` job's red is addressed to nobody** — the same blind spot `dependency-submission` hit for three days, which is why that one raises a tracking issue now. This workflow still has no escalation path. Not fixed here; see below.

## It corrects the record in #2841

That issue said the mechanism *"fired on the harmless case and missed the one it was built for"*. The first half was wrong. The three concurrency-superseded `cancelled` runs I cited as *having been retried* were themselves exiting 1. It never fired at all — on either conclusion.

The #2841 fix (retry on a partial spot kill) was still correct and necessary; it just landed on top of a mechanism that could not execute its own final step.

## The fix

`-R "${GITHUB_REPOSITORY}"` on both `gh run rerun` calls. Verified from a non-git directory that this is precisely what resolves it:

```
gh run rerun 999999999                          ->  failed to determine base repo
gh run rerun 999999999 -R JiRaska/open-bank-oss ->  HTTP 404 (reached the API)
```

`gh api` calls in the same step are unaffected — they pass a full `repos/{owner}/{repo}/…` path and need no base-repo resolution.

Deliberately **not** fixed by adding a checkout: cloning the repo to make one API call is pure waste, and the comment in the file now says so, so the next person does not "helpfully" add one.

## Test

All three branches re-run against real SHAs, with a stub that **fails when `-R` is absent** — so the test cannot pass against the unfixed script. The stub was validated first by feeding it a bare rerun and watching it exit 1:

```
validation:        gh run rerun 1              ->  STUB: CHYBI -R (ROZBITE), exit 1
cancelled branch   ->  gh run rerun -R … <id>            ✓
spot-kill branch   ->  gh run rerun -R … <id> --failed   ✓
genuine failure    ->  does not re-run                   ✓
```

## Awkward but worth saying

This is the exact footgun I added to `CLAUDE.md` in #2894 a few minutes earlier:

> **`gh` needs a repo context: outside a checkout it fails with `failed to run git: fatal: not a git repository`** … Pass `-R <owner>/<repo>` explicitly in any script whose working directory is not guaranteed.

I wrote that bullet and then shipped #2892 without applying it, because I restructured an existing call instead of checking whether the original had ever succeeded. "A gate that has only ever passed is unfalsified" has a mirror image: **a step that has only ever gone red is equally unexamined** — I read its red as "did not apply here" rather than "never worked".

## Follow-up not in this PR

`auto-retry-cancelled.yml` still has no escalation path, so the next time it breaks it will again fail silently for as long as nobody looks. `dependency-submission.yml` solved this by raising/refreshing an issue on failure. Worth doing the same here — happy to open it as a separate change.

Refs #2841

### DIFF
--- a/.github/workflows/auto-retry-cancelled.yml
+++ b/.github/workflows/auto-retry-cancelled.yml
@@ -45,6 +45,18 @@
 # dying — indistinguishable from the outside, so we accept one wasted retry on a manually
 # cancelled run over a more clever heuristic that can misfire.
 #
+# `-R` ON EVERY `gh run rerun` IS LOAD-BEARING (issue #2841 follow-up)
+# This job has no `actions/checkout`, so there is no git repository for `gh` to infer the
+# target repo from, and a bare `gh run rerun <id>` dies with
+#   failed to determine base repo: failed to run git: fatal: not a git repository
+# That is not a new hazard: it is how this workflow behaved from the day it was written, so
+# the auto-retry from #2330 had NEVER once succeeded — every invocation exited 1. Nothing
+# caught it because a `workflow_run` job's red is addressed to nobody (the same blind spot
+# `dependency-submission` hit for three days, which is why that one raises an issue now).
+# Verified from a non-git directory: without `-R` the command cannot resolve the repo; with
+# it, the call reaches the API. Do not drop the flag, and do not "fix" this by adding a
+# checkout — cloning the repo to make one API call is pure waste.
+#
 # WATCH LIST
 # Only the lanes that run on the spot pool and hurt when killed mid-flight: the PR services CI
 # and the two slow security lanes. Do NOT add auto-deploy here — it has its own reconcile tick.
@@ -84,7 +96,7 @@ jobs:
             # LOOP SAFETY above describes. `--failed` is a no-op against `cancelled`, so this
             # is always the full re-run.
             echo "Re-running cancelled run ${RUN_URL} (issue #2330)"
-            gh run rerun "${RUN_ID}"
+            gh run rerun -R "${GITHUB_REPOSITORY}" "${RUN_ID}"
             echo "::notice title=spot-kill auto-retry::Re-ran cancelled run ${RUN_URL} (attempt 2 of max 2; a second kill stays for a human)"
             exit 0
           fi
@@ -114,5 +126,5 @@ jobs:
           fi
 
           echo "Re-running ${RUN_URL}: ${killed} job(s) killed mid-step by a runner reclaim (issue #2841)"
-          gh run rerun "${RUN_ID}" --failed
+          gh run rerun -R "${GITHUB_REPOSITORY}" "${RUN_ID}" --failed
           echo "::notice title=spot-kill auto-retry::Re-ran ${killed} spot-killed job(s) in ${RUN_URL} (attempt 2 of max 2; a second kill stays for a human)"
--- a/.github/workflows/auto-retry-cancelled.yml
+++ b/.github/workflows/auto-retry-cancelled.yml
@@ -30,10 +30,24 @@
 #
 # A run-level `failure` is NOT retried on its own — that is the normal outcome of a broken
 # build, and blanket-retrying it would burn runner time and turn a deterministic red into a
-# flaky-looking one. It is retried only when a failed job carries the spot-kill SIGNATURE:
-# at least one `cancelled` step and NO `failed` step. A genuine build failure has a `failure`
-# step followed by `skipped` ones, never a `cancelled` one. Read from the jobs API, so no log
-# download is needed.
+# flaky-looking one. The run is retried when AT LEAST ONE failed job carries the spot-kill
+# SIGNATURE: at least one `cancelled` step and NO `failed` step. A genuine build failure has a
+# `failure` step followed by `skipped` ones, never a `cancelled` one. Read from the jobs API,
+# so no log download is needed.
+#
+# MIXED RUNS ARE THE NORMAL CASE, and the rule above is a per-RUN decision, not a per-JOB
+# filter. `gh run rerun --failed` re-runs every failed job, so a run containing both a
+# spot-killed job and a genuinely broken one re-runs both. That is accepted: the reclaimed job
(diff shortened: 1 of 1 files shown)


JiRaska:
Independently arrived at the same fix from the other direction — a scheduled sweep watching for the first live firing of the #2892 guard. Closing my branch in favour of this one; adding the numbers, since they change how this reads.

**This workflow has never once succeeded.** Not "has a bug" — has never done its job, from its very first run.

```
$ W=repos/JiRaska/open-bank-oss/actions/workflows/auto-retry-cancelled.yml/runs
$ for s in "" "&status=success" "&status=failure" "&status=skipped"; do
    gh api "$W?per_page=1$s" --jq '.total_count'; done
```

| | count |
|---|---|
| total runs | 1069 |
| **success** | **0** |
| failure | 208 |
| skipped (job `if:` false) | 861 |

`208 + 861 = 1069` — every run either never entered the job, or entered it and exited 1. There is no third category. The first run ever recorded, `2026-07-29T21:57:14Z`, already concluded `failure`.

So the 208 are not noise: each one is a run that met the retry criteria and was not retried. #2330's mechanism has been decorative for its entire lifetime, and #2892 extended a code path that had never executed successfully.

Production repro, run [30641343597](https://github.com/JiRaska/open-bank-oss/actions/runs/30641343597) (15:04Z today, `CONCLUSION: cancelled`):

```
Re-running cancelled run https://github.com/JiRaska/open-bank-oss/actions/runs/30640817396 (issue #2330)
failed to determine base repo: failed to run git: fatal: not a git repository (or any of the parent directories): .git
##[error]Process completed with exit code 1.
```

Falsified the fix rather than just confirming it — from a non-git directory, using `gh run view` so nothing is actually re-run:

```
$ cd /tmp/nongit && env -u GH_REPO gh run view 30641343597 --json conclusion
failed to determine base repo: failed to run git: fatal: not a git repository (or any of the parent directories): .git

$ cd /tmp/nongit && GH_REPO=JiRaska/open-bank-oss gh run view 30641343597 --json conclusion
{"conclusion":"failure"}
```

Same resolution path as `gh run rerun`, so `-R` fixes it identically.

One thought, take it or leave it: `GH_REPO: ${{ github.repository }}` in the step's `env` is equivalent and covers any `gh` call added to this step later, whereas `-R` has to be remembered per call site. Not worth a round-trip on this PR — the two call sites here are both covered either way.

The wider lesson is the one this repo already writes down and got bitten by again: a `workflow_run` job's red is addressed to nobody. 208 red runs over two days, on a mechanism whose whole purpose is to act unattended, and the only reason it surfaced is that a scheduled task happened to be reading this workflow's history for an unrelated guard. Worth considering whether this job should raise/refresh an issue on failure the way `dependency-submission` and `fleet-attestation` now do — its output is something others depend on, and its silence is indistinguishable from success.


---

Filed that last paragraph as #2901 so it does not get lost when this merges — escalation only, scoped to `auto-retry-cancelled.yml` (the only `workflow_run` workflow in the fleet, 55 scanned). Explicitly **not** a review comment on this PR: the `-R` fix here is correct and complete on its own, and #2901 does not block it.

JiRaska:
Heads-up, since editing my earlier comment notifies nobody: the escalation suggestion in it is now #2901.

Nothing for you to do here — it is a follow-up, not a review note, and it does not block this PR. The `-R` fix stands on its own.

JiRaska:
## First live firing observed — the discriminator works, and it corrected a claim in this PR

While watching for the first production firing of #2892, it fired. Auto-retry run [30644906717](https://github.com/JiRaska/open-bank-oss/actions/runs/30644906717), reacting to `Services CI` run 30630120640 (main, 63 jobs, concluded `failure`). Real output, read from the log **after the last `##[endgroup]`** — a naive grep matches the echoed script source and reports a hit on every run, which produced two false positives today:

```
Re-running .../30630120640: 1 job(s) killed mid-step by a runner reclaim (issue #2841)
failed to determine base repo: fatal: not a git repository
##[error]Process completed with exit code 1
```

**The discriminator was right.** That run had three failed jobs:

| job | cancelled steps | failed steps | signal |
|---|---|---|---|
| `build (openbank-product-catalog)` | 1 | 0 | `The runner has received a shutdown signal` |
| `build (openbank-mcp-service)` | 0 | 1 | none — genuine failure |
| `all-green` | 0 | 1 | none — downstream |

It found the one reclaimed job and ignored the two real failures. Confirmed against a real mixed run, not a fixture.

**And it confirms this PR's premise.** The guard reached its decision and printed its verdict, then died on the missing `-R`. That was the only thing standing between the decision and an actual re-run — fourth such failure in two hours (14:52, 15:04, 15:45, 15:57), all on the same line.

## Added a commit: the comment overclaimed

The description I wrote in #2892 said the run is retried *"only when a failed job carries the spot-kill signature"*, which reads as though only spot-killed jobs get re-run. They do not — the signature is a per-**run** decision, and `gh run rerun --failed` re-runs every failed job. So this run would have re-run the genuinely broken `mcp-service` too.

That is still the right trade (the reclaimed job needs the retry, the broken one fails again for the same reason, `run_attempt == 1` bounds it), but the wording promised precision the code does not have — and mixed runs are evidently the normal case, not a corner one. The comment now says so, and notes that per-job precision would need `gh run rerun --job <id>`, a larger change deliberately not made here.

Issue #2204, opened 2026-07-25
Data remediation after #2187: ledger tie-out, FX revaluation and CNB rate ingestion have never run — assess and backfill the historical gap


JiRaska:
## Why this exists

#2187 (PRs #2190, #2191, #2194) fixed the `HR000068` scheduler defect. Those PRs make the jobs **run from now on**. They do not address the fact that the jobs **never ran at all**, for as long as each has existed.

Confirmed affected, each with a revert-proof IT showing the job did literally nothing before the fix:

| service | job | what never happened |
|---|---|---|
| ledger | `TieOutScheduler.runTieOut` | the daily ledger tie-out control has never executed |
| ledger | `FxRevaluationScheduler.revalueDaily` | no FX revaluation has ever been posted |
| fx | `CnbRateIngestionScheduler.ingestDailyFixing` | the CNB daily fixing has never been ingested on schedule |
| billing | `BillingCycleScheduler.sweep` | latent only — off by default, no-ops without a batch |

The fx job's own `catch` swallowed the exception into a single ERROR line, so there was no crash and no alert.

## What needs deciding and doing

1. **Establish the window.** For each job, when was it introduced, and has any *other* path produced its output in the meantime? The fx rates may well be arriving via a manual or API route; the tie-out and revaluation may have no substitute at all. Do not assume — check the data.
2. **Assess the financial consequence.** FX revaluation is a posting job: if it has never run, the ledger carries unrevalued FX positions, and the size of that is a number someone has to produce. The tie-out is a *control*, so its absence means the control has never fired, not that data is wrong — but it also means nothing has been verifying what the control exists to verify. That interacts with the ~220k CZK ledger⇄sub-ledger drift already known from the balance-reconciliation work (#856).
3. **Decide on backfill.** Whether to backfill revaluation postings historically, or to revalue as-of-today and document the discontinuity, is a money-path accounting decision — not an engineering one. It needs the owner.
4. **Watch the first live runs.** These crons have never succeeded in production. The first real execution of each should be observed rather than assumed — especially the tie-out, which may report a large first-run discrepancy precisely because it has never run.

## Related structural gap

`quarkus.scheduler.enabled: false` under `%test` in both ledger and billing is *why* this class of defect is structurally invisible in those services: it leaves only direct method calls, which supply the exact Vert.x context the real scheduler does not. The new ITs re-enable it deliberately. Worth checking which other services disable the scheduler in test.

[...]

PR #2917, opened 2026-07-31
fix(fx): point the ČNB fixing ingestion at the URL that actually serves it

JiRaska:
## What

`openbank-fx-service`'s ČNB fixing URL was a **404** and had been for the whole life of the service. The path segment `kurzy-devizoveho-trhu` appears **twice** in ČNB's URL (section, then page); the config carried it once.

Measured from the `fx` namespace on the sandbox:

```
== 404  .../kurzy-devizoveho-trhu/kurzy/denni_kurz.txt              <- configured
== 200  .../kurzy-devizoveho-trhu/kurzy-devizoveho-trhu/denni_kurz.txt   "31.07.2026 #146"
```

`?date=DD.MM.YYYY` works on the correct URL too, so the historical backfill path in `POST /api/v1/fx/cnb/ingest` is mechanically available.

## Why nothing reported it

Every layer was green, and each for a defensible reason:

| layer | what it saw |
|---|---|
| rest-client + `@CircuitBreaker` | a 404 is a valid HTTP response — success |
| `CnbFixingParser` | rejected the 58 KB HTML error page ČNB serves |
| `CnbRateIngestionScheduler` | caught it, logged **one** ERROR line, returned |
| `FxRevaluationService` | no valid `source = CNB` rate, "skipping its revaluation leg", `posted = false` |

The last row is the one that matters: a successful-looking run of a job that revalued nothing. The only evidence anywhere was `fx_rates` not growing, and nothing alerts on a table not growing.

So #2191 made the job *run* (the #2187 `HR000068` fix — the deployed image is `sandbox-398a0433`, which **is** that commit), and it has failed silently every day since. On the sandbox the newest rate of any source is `2026-06-15` (46 days stale) and `source = CNB` has 3 rows, all seeded at the same `created_at` as the synthetic `INTERNAL` seed.

## The gate

A third-party URL in `application.yaml` is unfalsifiable by every test layer this repo has: a unit test stubs the client, an IT serves a local fixture, and a consumer pact answers whatever path it is asked for (#2283). The upstream is a foreign government feed with no pact. Fetching it is the only thing that can be wrong out loud.

`check-external-feeds.py` has two halves in **deliberately different lanes**:

- **DRIFT** — offline, deterministic, **enforced in `Validate manifests`**. Every third-party URL in an `openbank-*/src/main/resources/application.yaml` is either declared in `FEEDS` (probed) or in `NOT_PROBED` (with a reason). The declaration holds **no copy of the URL**: it names the YAML path and the probe reads the value out of the committed file, so the fix and the check cannot drift apart — the vacuity that made `ProductCatalogPactConsumerTest` unable to fail.
- **LIVENESS** — daily, fetches each feed and asserts the payload's **shape**, not its status code. Never blocks a merge; a third party having a bad day must not wedge the repo. A dead feed escalates onto #2204 (refreshing one comment, not stacking one per day).

Exit codes keep `2` (dead feed) and `3` (probe could not run) apart from `0`, because this whole issue is what happens when a broken thing's silence reads as health.

## Falsification

Not trusting a green — each half was run against a case it MUST flag:

- `--self-test`: 10/10, including `cnb_fixing` rejecting the exact ČNB 404 HTML page, an empty 200, a header with no rate lines, and a plausible-but-wrong `YYYY-MM-DD` header.
- drift, undeclared URL: added a bogus `https://…` to the fx YAML → flagged with file:line.
- drift, broken declaration: typo'd `yaml_path` → `no longer resolves … the probe would test nothing`.
- liveness, real internet: `cnb-daily-fixing` **OK**, `cnb-bank-registry` **DEAD (404)**, exit 2.

The first pass at the internal-host rule classified 15 in-cluster `http://svc.ns:port` URLs as third-party feeds; `is_internal` is now narrow and documented, and `https://` is external without exception.

## What this found on the way

**A second dead ČNB URL.** `openbank-customer-edge`'s bank registry `https://apl.cnb.cz/apljerrpp/jerr_banky.xml` is also a 404. `CnbBanksClient` falls back to an embedded `/banks.json`, so it degrades to a **frozen bank list** rather than an outage — which is exactly why nobody noticed. Declared and reported here, **not fixed**: the replacement endpoint needs deciding, not guessing. Filed separately.

## Not done here

- `rules.yaml` is untouched on purpose. Editing it restamps every service's OPA bundle (~44–65 files) and gives this PR a shelf life measured in hours; the rule is enforced in a required check either way. Say the word and it goes in a follow-up.
- The rate **backfill** (`POST /api/v1/fx/cnb/ingest?date=…` per business day) is sequenced *after* this merges and deploys — backfilling against a 404 would do nothing.

Refs #2204, #2187, #2191, #2283.

### DIFF
--- a/.github/scripts/check-external-feeds.py
+++ b/.github/scripts/check-external-feeds.py
@@ -1,0 +1,399 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) OpenBank contributors. Licensed under the Apache License, Version 2.0.
+#
+# External public-feed watch: catch a dead third-party data feed before a money-path job
+# silently stops producing (issue #2204).
+#
+# WHY THIS EXISTS
+#   `openbank-fx-service` ingests the statutory ČNB daily fixing from a public text feed. On
+#   2026-07-31 that feed's configured URL was a 404 — one path segment short — and had been for
+#   the whole life of the service. ČNB serves the 404 as a 58 KB HTML page, `CnbFixingParser`
+#   rejects it, and `CnbRateIngestionScheduler`'s catch swallows the exception into a single
+#   ERROR line. So:
+#
+#     * every layer was green. The scheduler fired (after #2187 fixed HR000068), the rest-client
+#       resolved, the circuit breaker never opened — a 404 is a perfectly good HTTP response.
+#     * the ONLY evidence was an absence: `fx_rates` stopped gaining `source = CNB` rows. Nothing
+#       alerts on a table not growing.
+#     * downstream, `FxRevaluationService` reads `findLatestBySource(... validTo > now)`, gets
+#       `null`, logs "No ČNB rate for EUR/CZK — skipping its revaluation leg", and returns
+#       `posted = false`. A successful-looking run of a job that revalued nothing.
+#
+#   A URL in a config file is not covered by any test in this repo, by construction: a unit test
+#   stubs the client, an IT serves a local fixture, and a consumer pact answers whatever path it
+#   is asked for (the #2283 lesson — only the real provider can falsify a path). The upstream here
+#   is a foreign government feed with no pact and no contract; the only thing that can falsify its
+#   URL is fetching it.
+#
+# WHAT IT CHECKS
+#   1. DRIFT, both directions, offline — this half is PR-BLOCKING.
+#        Every external `http(s)://` URL in a scanned `application.yaml` must be either
+#        (a) declared in FEEDS, so it is probed, or
+#        (b) declared in NOT_PROBED with a reason it cannot be.
+#        A new external dependency added without either is a feed this watch would silently know
+#        nothing about while still passing. This is the repo's derived-scope rule: a gate whose
+#        coverage set is maintained separately from the artifacts it covers reads as *passing*
+#        when the set is short, never as *unchecked*.
+#
+#        FEEDS holds NO copy of the URL. It names the YAML path (`quarkus.rest-client.cnb-feed.url`)
+#        and the probe reads the value out of the committed file. A second copy would move together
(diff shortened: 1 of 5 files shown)


JiRaska:
Follow-up detail found while arming the backfill, worth recording against this PR's own claim.

The PR body says the `@CircuitBreaker` never opened. That is true, and the reason is sharper than "a 404 is a valid response": **at the scheduler's cadence the breaker was structurally unable to open.**

`CnbRateProviderAdapter` declares `requestVolumeThreshold = 4, failureRatio = 0.5`. `CnbRateIngestionScheduler` makes exactly **one** call per day. A rolling window that needs 4 requests before it will evaluate a failure ratio can never be filled by a once-daily caller, so the breaker sat closed through 46 consecutive failures and would have sat closed indefinitely.

Measured accidentally and then deliberately: calling `POST /api/v1/fx/cnb/ingest` by hand several times in a row **did** trip it —

```
org.eclipse.microprofile.faulttolerance.exceptions.CircuitBreakerOpenException:
  com.openbank.fx.infrastructure.client.CnbRateProviderAdapter#fetchWithResilience circuit breaker is open
```

— which is the same adapter, the same upstream and the same 404. The only variable was call volume.

Generalisation worth keeping: **fault-tolerance annotations sized for request traffic are inert on a daily scheduler.** A `requestVolumeThreshold` above 1 on a once-a-day path is not a weaker signal, it is no signal — the resilience posture reads as present in code review and cannot fire in production. Nothing is changed here (opening a breaker would not have alerted anyone either — that is what the feed watch is for), but it is the reason the "resilient adapter" in this file bought nothing against the defect it was facing.

The backfill runner also records this, since a human replaying 36 dates in a loop *will* trip the breaker on the first bad date and should not read that as the feed being down.

Issue #1449, opened 2026-07-17
A failing dependency-graph submission tells nobody — it rotted for 3 days and the new Gradle gate depends on it


JiRaska:
## What happened

From **2026-07-14 to 2026-07-17**, every run of `dependency-submission.yml` died with `Java heap space` in `Resolve fleet dependency graph` (`-Xmx4g` stopped fitting as the fleet grew). The fleet's Gradle dependency graph silently stopped updating for three days. The `-Xmx6g` fix landed on 2026-07-17; the run at 06:43 was the first success since.

## Why nobody noticed

Nothing was wrong with the signal — **every one of those runs went red**. The problem is that a red push-triggered workflow on `main` is addressed to nobody. It is not a required check, it does not block a merge, and no one watches it.

## Why it matters more now

Two things read that graph, and **neither goes red when it is stale**:

- **Dependabot security alerts** stop reflecting reality — no new alerts for the Kotlin/Java fleet, which reads exactly like "no vulnerabilities".
- **The PR-time Gradle dependency gate** (#1419 / #1421) diffs each PR against this graph. A stale base makes the gate **pass without checking anything** — a green check that verified nothing.

## When will it rot again

The heap is a **ratchet with nothing measuring it**: 4g was fine until the fleet outgrew it, and 6g will go the same way. This cannot be predicted from here — so the fix is not a bigger number, it is making the next occurrence arrive as an addressed issue instead of three quiet days.

## Fix

Add a `raise-issue` job to `dependency-submission.yml` that opens (or refreshes) a tracking issue when the submission fails, mirroring `fleet-attestation.yml`'s established escalation pattern — with the heap diagnosis in the body so the next person does not have to re-derive it.

Deliberately **not** gated on `schedule` (unlike fleet-attestation's): the OOM ran red on main **pushes**, and the schedule is only weekly, so a schedule-only alarm would have waited up to 7 days to say anything. `pull_request` **is** excluded — a PR-run failure is already visible on the PR itself.

## Known gaps this does not close

- **Graph staleness with no failure.** If the workflow stops running entirely (e.g. GitHub disables cron on an inactive repo), there is no failure to alarm on. A staleness check would need a threshold above the weekly cadence (>7 days), so detection would be slow; not built.
[...]

PR #2919, opened 2026-07-31
fix(ci): give every checkoutless gh job a repo context — four dead escalation paths

JiRaska:
Follow-up to #2898. That PR fixed one instance; sweeping the rest of `.github/workflows` found **four more, none of which has ever executed**.

## The bug

A job with no `actions/checkout` gives `gh` nothing to infer the target repo from, so every subcommand dies:

```
failed to determine base repo: failed to run git: fatal: not a git repository
```

| workflow | job | calls |
|---|---|---|
| `api-contract-post-merge.yml` | `raise-issue` | `gh issue list` / `create` / `comment` |
| `auto-deploy.yml` | `deploy-signal` | `gh workflow run` |
| `dependabot-auto-merge.yml` | `auto-merge` | `gh pr merge` |
| `dependency-submission.yml` | `raise-issue` | `gh issue list` / `create` / `comment` |

Every one is an escalation or automation path — the code that runs when something has *already* gone wrong. That is why the breakage stayed invisible: these jobs almost never execute, and when they don't execute they don't fail.

## The sharpest one

`dependency-submission.yml`'s `raise-issue` is the **#1449 alarm** — the mechanism this repo points at as the answer to *"a red push-triggered workflow on main is addressed to nobody"*.

It has never fired: no non-PR submission failure in the last 60 runs. So its own breakage was invisible by the same mechanism it exists to defeat. Had the 2026-07-14 three-day `Java heap space` outage happened *after* the alarm was added, it would have failed to raise the issue and the outage would have stayed exactly as quiet.

## Verified, not assumed — in both directions

- `gh issue list`, `gh issue create` and `gh run rerun` were each run from a non-git directory: all fail with the message above; with the repo supplied they reach the API (`HTTP 404` on a bogus id, i.e. resolution worked).
- **The sweep initially flagged five.** `record-deployment-on-merge.yml` already sets `GH_REPO`, so it is fine — and checking for `GH_REPO` as well as `-R` is the only reason it isn't in this diff. Its comment is also where the fix pattern and its rationale come from: *"every run silently records zero services … Pass the repo explicitly so it works API-only."* Someone hit this once and fixed it in one place.
- None of the four has ever run: every occurrence across recent runs is `skipped`, or the job is absent from the run entirely. So there is no counter-evidence that any of them somehow works.

## Choice of fix

Job-level `GH_REPO: ${{ github.repository }}` rather than `-R` on each call — matches the existing precedent in `record-deployment-on-merge.yml` and covers any `gh` call added to these jobs later. #2898's per-call `-R` is equivalent and left alone rather than churned.

## Note on actionlint

`auto-deploy.yml` produces a warning that `openbank-build` is an unknown runner label. **Pre-existing** — 2 occurrences on `main`, 2 on this branch, and no other findings. Not introduced here.

## Verification

- `actionlint` — clean on all four (modulo the pre-existing label warning above)
- `yamllint` under the exact config `ci.yml` builds — clean on branch and on `origin/main`'s copy for all four
- each job re-parsed after editing to confirm `GH_REPO` landed on the job and the step count is unchanged

## Worth doing next, not done here

Nothing stops this recurring. A guard over `.github/workflows` — *a job that invokes `gh` and has no `actions/checkout` must set `GH_REPO` or pass `-R`* — is mechanically checkable and would have caught all five instances at PR time. I'd rather propose it than bundle it: it needs its own falsification pass, including feeding it the `record-deployment-on-merge.yml` shape it must NOT flag.

Refs #2841, #1449

### DIFF
--- a/.github/workflows/api-contract-post-merge.yml
+++ b/.github/workflows/api-contract-post-merge.yml
@@ -105,6 +105,11 @@ jobs:
     permissions:
       contents: read
       issues: write
+    # This job has no actions/checkout, so `gh` has nothing to infer the target repo from
+    # and every subcommand dies with `failed to determine base repo: fatal: not a git
+    # repository`. Set it explicitly (same fix, same reason as record-deployment-on-merge.yml).
+    env:
+      GH_REPO: ${{ github.repository }}
     steps:
       - name: Open or refresh the tracking issue
         env:
--- a/.github/workflows/auto-deploy.yml
+++ b/.github/workflows/auto-deploy.yml
@@ -1282,11 +1282,6 @@ jobs:
     if: always() && needs.changes.outputs.any == 'true' && needs.build-push.result == 'success'
     runs-on: ubuntu-latest
     timeout-minutes: 2
-    # This job has no actions/checkout, so `gh` has nothing to infer the target repo from
-    # and every subcommand dies with `failed to determine base repo: fatal: not a git
-    # repository`. Set it explicitly (same fix, same reason as record-deployment-on-merge.yml).
-    env:
-      GH_REPO: ${{ github.repository }}
     steps:
       - name: Report undelivered deploy
         env:
--- a/.github/workflows/auto-deploy.yml
+++ b/.github/workflows/auto-deploy.yml
@@ -1282,6 +1282,11 @@ jobs:
     if: always() && needs.changes.outputs.any == 'true' && needs.build-push.result == 'success'
     runs-on: ubuntu-latest
     timeout-minutes: 2
+    # This job has no actions/checkout, so `gh` has nothing to infer the target repo from
+    # and every subcommand dies with `failed to determine base repo: fatal: not a git
+    # repository`. Set it explicitly (same fix, same reason as record-deployment-on-merge.yml).
+    env:
+      GH_REPO: ${{ github.repository }}
     steps:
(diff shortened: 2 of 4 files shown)


JiRaska:
## Correction: `auto-deploy.yml` was never broken — reverted in `8faea252a`

The guard I built alongside this PR flagged `auto-deploy.yml:deploy-signal` and I took it at face value. It is a **false positive**. That job has no `gh` invocation at all; the only match is prose inside the notice body it builds:

```
"re-dispatch to recover (\`gh workflow run auto-deploy.yml -f services=<svc>\`), or" \
```

An escaped-backtick markdown code span inside a double-quoted string — unambiguously text, since an escaped backtick in a double-quoted shell string is a literal character, never command substitution.

Setting `GH_REPO` there would have been harmless, which is exactly why it is worth reverting rather than leaving: a workflow carrying an env var *"because a check said so"* is how a guard's false positive becomes permanent, undocumented cargo.

**The remaining three are real and unchanged** — `api-contract-post-merge.yml`, `dependabot-auto-merge.yml`, `dependency-submission.yml`. Verified by running the guard against `origin/main`: it reports exactly those three, and passes on this branch.

The guard now strips escaped-backtick spans and carries this shape as a must-not-flag self-test case. It is in #2922, stacked on this branch — merge this one first.

PR #2763, opened 2026-07-30
feat(mcp): accept staff OBO tokens behind mcp.obo.enabled (ADR-0224 phase 1b)

JiRaska:
## Summary

Second slice of **ADR-0224** — the resource-server half, completing what #2762 (realm config) started:

- `CallerContextResolver` accepts a **HUMAN staff OBO token** behind `mcp.obo.enabled` (default **false**), recognized by four marks, ALL required: `azp` (actor), `sid` (session), `openbank-mcp-service` in `aud`, ≥1 bounded realm role. Missing any → anonymous, fail-closed.
- `McpEndpoint` forwards `principal.type` + `roles` to the shared PDP for **both** `tools/call` and `tools/list` — an OBO call is gated by the same realm-role vocabulary as a REST call (ADR-0034's promise, now wired).
- Honest state: with today's rego, a HUMAN principal is policy-denied for every `query.*` capability (those grants are the ADR-0223 matrix's concern), and consent-scoped tools stay closed to staff tokens (no consent → downstream fail-closed). The plumbing is proven; the capability grants land with 0223/0224-phase-2.
- Audit events carry `azp` as the first act-chain link + `sid` (ADR-0226 D2) — a staff action via MCP is attributable from day one.

**Tests (5 new):** flag-off → anonymous; flag-on → PDP sees `HUMAN` + bounded roles; azp/sid on the audit event; all five missing-mark variants → anonymous even with the flag on.

⚠️ **STACKED PR** — branched off `feat/mcp-audit-channel-population` (#2760, itself stacked on #2755+#2756; ADR branch merged in for the gate). Rebase after parents merge collapses the diff to the single commit. `version.txt` 0.14.0 (per-PR bump; RP reconciles on its release PR).

## Type of change

- [x] feat (flag-gated, inert by default)

## Linked issues

Refs #2750

## Definition of Done

- [x] `:openbank-mcp-service:build` + `koverVerify` green
- [x] Flag defaults to disabled — zero behavior change without `mcp.obo.enabled=true`
- [x] No grant widening: PDP evaluates roles; today's policy denies all HUMAN capability calls
- [x] Commit signed (-s -S), conventional format

Issue #2750, opened 2026-07-30
Channel-agnostic authz + human-operator MCP channel programme (ADR-0223/0224/0225/0226)


JiRaska:
## What needs doing

Tracker for the programme decided in ADR-0223/0224/0225/0226 (admin-UI audit, 2026-07):

1. **ADR-0223** — channel-agnostic authorization hardening: sidecar-only enforcement, retire `operator-read-any` for a role→action matrix as data, graduated four-eyes flip.
2. **ADR-0224** — human-operator MCP channel: RFC 8693 on-behalf-of exchange, bounded role subsets, act chain, action-class propose/dispose.
3. **ADR-0225** — policy-filtered `tools/list` (capability-shaped discovery).
4. **ADR-0226** — cross-channel audit correlation (`channel`, `act_chain`, `session_id`).

Implementation lands as one PR per ADR slice (0225 first — smallest, self-contained).

## Background

Admin-UI audit found: UUID-first search on key backoffice screens, mock data pages presented as real, 16 fragmented ApprovalResources, backend capabilities with no UI (lending/fraud/sdd/statements/billing), a UI/backend role-vocabulary drift (SUPERVISOR/KYC* missing in UI), and `operator-read-any` granting every operator read access to any resource while the UI permission matrix is cosmetic. The ADRs are the security spine of the remediation; UX ADRs (unified search, persona IA, coverage consoles) follow separately.

JiRaska:
## Programme state (2026-07-30, agent session)

### PRs and merge order
1. **#2752** — ADR-0223/0224/0225/0226 docs. CI green. **Merge first** — everything references these ADR numbers.
2. **#2755** — ADR-0225 impl: policy-filtered `tools/list` (capability-shaped discovery, fail-closed, audited). Builds green; ADR branch merged in so the ADR-ref gate passes independently.
3. **#2756** — ADR-0226 D1/D3: audit `channel`/`act_chain`/`session_id` schema + store + Flyway V9 + `GET /api/v1/audit/entries/by-actor/{id}` + OpenAPI 1.4.0 contract. Chain hash untouched (payload carries the fields).
4. **#2760** (draft, stacked on #2755+#2756) — ADR-0226 D2: mcp-service stamps every MCP audit event with channel/actChain/sessionId; RFC 8693 `act` parser (fail-safe).

### Remaining phases — implementation-ready specs

[...]

PR #2955, opened 2026-07-31
fix(mcp): key an agent session by the token's sub so staff OBO resolves

### LINKED ISSUE #2938: mcp-service: staff OBO principal never resolves — session row stores principal.name, resolver compares jwt sub (ADR-0224 D1/D2 inert)
## Summary

The staff OBO MCP channel (ADR-0224) can never authenticate a human. `CallerContextResolver` compares the session row's `subject` against the JWT's `sub` claim, but the session row stores `identity.principal.name`. In production those are two different values — an email and a UUID — so the comparison always fails, the resolver returns `null`, and every staff OBO call silently degrades to **anonymous**.

Verified end-to-end on the sandbox (2026-07-31 18:09Z), against `sandbox-dde5f188` with `MCP_OBO_ENABLED=true`.

## The mismatch

Write side — [`McpSessionResource.kt:70`](https://github.com/JiRaska/open-bank-oss/blob/main/openbank-mcp-service/src/main/kotlin/com/openbank/mcp/infrastructure/rest/McpSessionResource.kt#L70):

```kotlin
it.subject = identity.principal.name        // Quarkus OIDC → preferred_username → "admin@openbank.local"
```

Read side — [`CallerContextResolver.kt:91`](https://github.com/JiRaska/open-bank-oss/blob/main/openbank-mcp-service/src/main/kotlin/com/openbank/mcp/infrastructure/mcp/CallerContextResolver.kt#L91):

```kotlin
val subject = jwt.subject                   // the `sub` claim → "3a046823-5d47-4de1-9f3f-b1b2a953d2cc"
...
if (session.subject != subject) return null // never equal in production
```

## Reproduction (live sandbox, full relay chain replayed by hand)

The chain from [`route.ts`](https://github.com/JiRaska/open-bank-oss/blob/main/openbank-admin-ui/src/app/api/agent/obo-mcp/route.ts) — create session → RFC 8693 exchange → bind jti → call — reproduced with curl. Every leg **succeeded**:

| Leg | Result |
|---|---|
| `POST /api/v1/mcp/sessions` | `201`, row `019fb95d-…`, `subject: admin@openbank.local` |
| token exchange at Keycloak | `200`, `aud=openbank-mcp-service`, `azp=openbank-admin-ui`, `sid` present, TTL 300 s, roles `[ROLE_OPERATOR, ROLE_VIEWER, ROLE_ADMIN]` |
| `PATCH /api/v1/mcp/sessions/{id}/bind` | `200`, `jti` persisted (confirmed in `agent_session`) |
| `POST /mcp` `tools/list` with the OBO token | `{"tools": []}` |

All four resolver marks were present and the session was live and bound — so `findActiveByJti` hit, and the subject comparison is the only remaining gate. The audit line for that exact call proves the degradation:

```
op=mcp.tools.list result=DENIED actor=unknown actorType=AI_AGENT
```

`actorType=AI_AGENT`, not `HUMAN`; `actor=unknown`. An anonymous request in the same second returns the identical empty list — the two are indistinguishable.

## Why every test is green

`McpEndpointTest` sets `"sub" to "jane.operator"` and stubs the session row with `it.subject = "jane.operator"`; `McpSessionLifecycleIT` uses `@TestSecurity(user = "jane.operator")`, which makes `principal.name` **and** `sub` the same string. The two identifiers are only ever equal in the fixture. Nothing in the suite mints a token where `sub` is a UUID and `preferred_username` is an email — i.e. the shape every real Keycloak token has.

This is the class the repo already documents: a test whose fixture collapses two distinct production values into one cannot fail on their divergence.

## Blast radius

- ADR-0224 D1 (HUMAN principal), D2 (session binding, live revocation) and D5 (act chain into the PDP) are all inert in every environment, not just the sandbox.
- ADR-0225's policy-filtered `tools/list` cannot be observed for humans — the filter runs against an anonymous principal, so the empty list looks like correct fail-closed behaviour rather than a broken identity.
- The E2E runbook in #2750 step 2.1 expects "non-empty, policy-filtered tool list (NOT the anonymous empty list)". It gets the anonymous empty list, and the two are only distinguishable in the audit log.
- Live revocation (step 3) cannot be demonstrated: the token was never accepted, so revoking its session changes nothing observable.

## Suggested fix

Store and compare the same identifier on both sides. `sub` is the stable one — `preferred_username` is mutable in Keycloak and is not an identifier the token guarantees. Either:

1. `McpSessionResource` stores `jwt.subject` (add a display field for `principal.name` if the UI needs one), or
2. the resolver compares against `preferred_username` — weaker, since that claim can change under an existing session.

Option 1, plus a regression test whose fixture uses a UUID `sub` and a *different* `preferred_username`. Without that asymmetry in the fixture, the same bug can be reintroduced with the suite green.

Also worth deciding: `AgentSessionResponse.subject` and the `owns()` check in `McpSessionResource` (`session.subject == identity.principal.name`) read the same field, so any change has to move all three together.

JiRaska:
## What

`McpSessionResource` wrote `identity.principal.name` into `agent_session.subject`; `CallerContextResolver` validates an OBO token by comparing that row to `jwt.subject`. In production the two are different values — Quarkus resolves the name claim (`upn`/`preferred_username`, an email in this realm) while `sub` is a UUID — so the comparison never held, the resolver returned `null`, and **every staff OBO call degraded to anonymous**.

Both writers and the ownership guard now read through one `callerSubject()` helper, so they cannot diverge again:

```kotlin
private fun callerSubject(): String =
    (identity.principal as? JsonWebToken)?.subject ?: identity.principal.name
```

`sub` is also the right key on its own merit — `preferred_username` is mutable in Keycloak and would strand a live session on a rename. The `principal.name` fallback covers a non-JWT identity (`@TestSecurity`); production always takes the `sub` branch.

Closes #2938.

## Evidence — reproduced on the sandbox before the fix

Full relay chain from [`route.ts`](https://github.com/JiRaska/open-bank-oss/blob/main/openbank-admin-ui/src/app/api/agent/obo-mcp/route.ts) replayed by hand against `sandbox-dde5f188` with `MCP_OBO_ENABLED=true`. Every leg succeeded and the last one still returned the anonymous answer:

| Leg | Result |
|---|---|
| `POST /api/v1/mcp/sessions` | `201`, row `subject: admin@openbank.local` |
| RFC 8693 exchange | `200`, `aud=openbank-mcp-service`, `azp=openbank-admin-ui`, `sid` ✓, roles ✓, TTL 300 s |
| `PATCH /sessions/{id}/bind` | `200`, `jti` persisted |
| `POST /mcp` `tools/list` | `{"tools": []}` — identical to the anonymous control |

Audit line for that exact call: `op=mcp.tools.list result=DENIED actor=unknown actorType=AI_AGENT`.

Worth noting how this hides: an empty tool list is *also* the correct fail-closed answer for a caller with no grants — which is what #2763 says today's rego gives a HUMAN principal until the ADR-0223 matrix lands. Nothing distinguishes "policy denied everything" from "identity was never accepted" except the audit line.

## Why the suite was green

`McpEndpointTest` used `"jane.operator"` for the `sub` claim **and** for the stubbed row's subject. `McpSessionLifecycleIT` uses `@TestSecurity(user = "jane.operator")`, which makes `principal.name` and `sub` the same string by construction. The two identifiers were only ever equal in the fixtures.

Fixed at the fixture level, because that is where the blindness lives:

- `McpEndpointTest.staffClaims` now carries a UUID `sub` **and** a distinct `preferred_username`.
- `TestJsonWebToken.getName()` resolves `preferred_username` before `sub`, mirroring Quarkus' principal-name resolution. Without that, a fixture cannot express a real token at all.
- Two new unit tests in `McpSessionResourceTest`: a session issued from a JWT identity is keyed by `sub` (and explicitly *not* by the name claim), and a row written in the old shape is **not** recognised as the caller's (403).

`McpSessionLifecycleIT` still exercises the fallback branch — `@TestSecurity` supplies a non-JWT principal — so the regression lives in the unit tests, not the IT.

## Verification — fail-first proven

Run in an isolated `GRADLE_USER_HOME` (see the environment note below for why). Three steps, same command each time: `:openbank-mcp-service:test --tests '*McpSessionResourceTest*' --tests '*McpEndpointTest*'`.

**1 — with the fix: green.**

```
com.openbank.mcp.McpEndpointTest:        tests=35  failures=0  errors=0
com.openbank.mcp.McpSessionResourceTest: tests=6   failures=0  errors=0
BUILD SUCCESSFUL
```

**2 — `callerSubject()` reverted to `identity.principal.name` (pre-fix behaviour), tests untouched: red, and red for the right reason.**

```
com.openbank.mcp.McpSessionResourceTest: tests=6  failures=2

FAIL  a session issued from a JWT identity is keyed by sub, not by the name claim
      expected: "3a046823-5d47-4de1-9f3f-b1b2a953d2cc"
      but was:  "admin@openbank.local"

FAIL  ownership of a sub-keyed session is recognised, and a name-keyed row is not theirs
      expected: 200
      but was:  403
```

That is the production confusion exactly: the row keyed by an email, the resolver reading a UUID.

**3 — fix restored: green again.** `BUILD SUCCESSFUL`, 35 + 6 tests, 0 failures.

One prediction of mine was wrong and is worth recording: I expected `McpEndpointTest` to go red in step 2 as well. It does not, and should not — that test never calls `McpSessionResource`; it stubs the repository with `subject = STAFF_SUB` directly, so it covers the *read* side (the resolver) and is unaffected by reverting the *write* side. Its fixture still had to change: with `"jane.operator"` as both `sub` and the name claim it could not tell the two apart, so it could not guard the read side against the same confusion.

## Environment note (not a change in this PR)

The local runs above needed a `GRADLE_USER_HOME` under `/private/tmp` sharing nothing with `~/.gradle`. Against the shared home the build fails before reaching any test:

```
e: .../openbank-libs-domain/.../OpaSidecarPolicyDecisionPoint.kt:7:30 Unresolved reference 'databind'.
```

Reproduced on a pristine `origin/main` worktree with an unmodified tree, so it is not this branch. Files disappear from `~/.gradle` while the machine is idle: the unpacked Gradle 9.6.1 distribution lost its `lib/` (the wrapper keeps the `.ok` marker, so it considers itself installed and never re-unpacks — `rm -rf ~/.gradle/wrapper/dists/gradle-9.6.1-bin/<hash>` fixes that), and a `jackson-databind-2.22.1.jar` I placed into `modules-2` by hand, hashing to exactly the value `verification-metadata.xml` declares, was **gone again within the hour** while the copy in the isolated home survived. No build deletes a valid artifact.

Also unexplained: `maven-settings-3.9.16.jar` and `smallrye-common-os-2.19.0.jar` fail verification locally with checksums (`441f7230…`, `e04ba519…`) matching neither the committed metadata nor what repo1 and the GCS mirror serve — both return exactly the committed hashes, verified by direct download, with `REPOSILITE_MIRROR_URL` unset. Flagging it for whoever owns this machine; nothing here belongs in the repo.

## Not in scope

No schema change: `agent_session.subject` keeps its type, and existing sandbox rows expire within the 15-minute TTL. `AgentSessionResponse.subject` now surfaces the `sub` UUID rather than an email — no UI reads it today (the relay uses only `id`), but a future operator console wanting a display name should add a separate field rather than change this key back.

### DIFF
--- a/openbank-mcp-service/src/main/kotlin/com/openbank/mcp/infrastructure/rest/McpSessionResource.kt
+++ b/openbank-mcp-service/src/main/kotlin/com/openbank/mcp/infrastructure/rest/McpSessionResource.kt
@@ -22,6 +22,7 @@ import jakarta.ws.rs.Produces
 import jakarta.ws.rs.core.MediaType
 import jakarta.ws.rs.core.Response
 import org.eclipse.microprofile.config.inject.ConfigProperty
+import org.eclipse.microprofile.jwt.JsonWebToken
 import org.eclipse.microprofile.openapi.annotations.Operation
 import org.eclipse.microprofile.openapi.annotations.tags.Tag
 import org.jboss.logging.Logger
@@ -67,7 +68,7 @@ class McpSessionResource(
         val now = Instant.now(clock)
         val session = AgentSessionEntity().also {
             it.id = Ids.newId()
-            it.subject = identity.principal.name
+            it.subject = callerSubject()
             it.roleCeiling = ceiling.joinToString(",", prefix = "[", postfix = "]") { r -> "\"$r\"" }
             it.clientId = request.clientId
             it.purpose = request.purpose
@@ -137,7 +138,23 @@ class McpSessionResource(
     // Broken-access-control guard: the role grant admits every operator, but a session row is
     // its owner's alone — reading or revoking another operator's session needs ROLE_ADMIN.
     private fun owns(session: AgentSessionEntity): Boolean =
-        identity.roles.contains("ROLE_ADMIN") || session.subject == identity.principal.name
+        identity.roles.contains("ROLE_ADMIN") || session.subject == callerSubject()
+
+    /**
+     * The identifier a session row is keyed by: the token's `sub`, never `principal.name`.
+     *
+     * [com.openbank.mcp.infrastructure.mcp.CallerContextResolver] validates an OBO token by
+     * comparing the row's subject to `jwt.subject`, so the row MUST hold the same claim. With
+     * `principal.name` it held Quarkus' name claim (`upn`/`preferred_username` — an email in this
+     * realm) while the resolver read the `sub` UUID: never equal, so every staff OBO call fell
+     * through to anonymous with no error anywhere (#2938). `sub` is also the stable choice —
+     * `preferred_username` is mutable in Keycloak and would strand a live session on rename.
+     *
+     * The `principal.name` fallback covers a non-JWT identity (`@TestSecurity`); production always
+     * takes the `sub` branch. Both writers and [owns] read through here so the two can't diverge.
+     */
+    private fun callerSubject(): String =
+        (identity.principal as? JsonWebToken)?.subject ?: identity.principal.name
 
     private fun forbidden(): Response =
         Response.status(Response.Status.FORBIDDEN).entity(mapOf("error" to "not your session")).build()
--- a/openbank-mcp-service/src/main/kotlin/com/openbank/mcp/infrastructure/rest/McpSessionResource.kt
+++ b/openbank-mcp-service/src/main/kotlin/com/openbank/mcp/infrastructure/rest/McpSessionResource.kt
@@ -153,8 +153,7 @@ class McpSessionResource(
      * The `principal.name` fallback covers a non-JWT identity (`@TestSecurity`); production always
      * takes the `sub` branch. Both writers and [owns] read through here so the two can't diverge.
      */
-    private fun callerSubject(): String =
-        (identity.principal as? JsonWebToken)?.subject ?: identity.principal.name
+    private fun callerSubject(): String = (identity.principal as? JsonWebToken)?.subject ?: identity.principal.name
 
     private fun forbidden(): Response =
         Response.status(Response.Status.FORBIDDEN).entity(mapOf("error" to "not your session")).build()
--- a/openbank-mcp-service/src/test/kotlin/com/openbank/mcp/McpEndpointTest.kt
+++ b/openbank-mcp-service/src/test/kotlin/com/openbank/mcp/McpEndpointTest.kt
@@ -156,7 +156,7 @@ class McpEndpointTest {
         )
         val principal = pdp.queries.single().principal
         assertThat(principal.type).isEqualTo("HUMAN")
-        assertThat(principal.id).isEqualTo("jane.operator")
+        assertThat(principal.id).isEqualTo(STAFF_SUB)
         assertThat(principal.roles).containsExactly("ROLE_OPERATOR")
     }
 
@@ -224,7 +224,12 @@ class McpEndpointTest {
     private val staffSessionId = "3f2a9c41-7b5e-4c8d-9e0f-1a2b3c4d5e6f"
 
     private val staffClaims = mapOf(
-        "sub" to "jane.operator",
+        // A realm token's `sub` is a UUID and its name claim is an email — they are NEVER the same
+        // string. The old fixture used "jane.operator" for both, which is exactly why #2938 (row
+        // keyed by the name claim, token validated against `sub`) passed every test while the
+        // staff channel was dead in every environment.
+        "sub" to STAFF_SUB,
+        "preferred_username" to "jane.operator@openbank.local",
         "azp" to "openbank-admin-ui",
         "jti" to staffSessionId,
         "aud" to "openbank-mcp-service",
@@ -237,7 +242,7 @@ class McpEndpointTest {
     private fun fakeSessionRepo() = object : AgentSessionRepository() {
         override suspend fun findActiveByJti(jti: String, asOf: java.time.Instant) = if (jti == staffSessionId) {
             AgentSessionEntity().also {
-                it.subject = "jane.operator"
+                it.subject = STAFF_SUB
                 it.roleCeiling = "[\"ROLE_OPERATOR\"]"
                 it.clientId = "admin-ui"
                 it.jti = staffSessionId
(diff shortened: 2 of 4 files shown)


JiRaska:
Closing as superseded by #2946 — and, more importantly, **merging this now would reintroduce the
defect both PRs were written to fix.**

Rebased onto `main` first (it was 171 commits behind) so the question could be answered on current
code rather than on a snapshot. The three `McpEndpointTest` failures survived the rebase, so they
are not drift.

## What changed underneath this branch

`#2946` (merged 2026-07-31 19:10) fixed the same UUID-vs-name mismatch **from the reader side**:

```
writer  McpSessionResource        session.subject = identity.principal.name
reader  CallerContextResolver     preferred_username ?: sub   →  compared to session.subject
```

That pair is self-consistent on `main`, and `E2E #2750` pins it.

This branch fixes the mismatch **from the writer side** — `session.subject = jwt.subject` — and does
not touch the reader. Applied on top of #2946 the pair becomes:

```
writer  sub (a UUID)
reader  preferred_username (an email in this realm)
```

which never matches, so every staff OBO call falls through to anonymous with no error anywhere —
precisely the #2938 symptom this PR describes. The tests say so concretely: the PDP is never
queried at all (`pdp.queries.single()` → `NoSuchElementException`).

## The part of this PR that is still right

Its rationale is not wrong, and #2946 does not address it:

> `preferred_username` is mutable in Keycloak and would strand a live session on rename.

On `main` today, renaming a staff user in Keycloak changes `preferred_username`, the reader computes
the new name, the stored row still holds the old one, and the session is silently stranded — the
same anonymous-with-no-error shape, just triggered by an admin action instead of a claim shape.

Filed separately rather than dropped, because it is a real property of the merged design and this
PR is the only place it was written down.

JiRaska:
Superseded by #2946; merging would re-break the writer/reader pair — see the analysis above.

PR #2976, opened 2026-07-31
fix(consent): the expiration sweep runs on a Vert.x context, and the gate now checks the property

### LINKED ISSUE #2913: consent expiration sweep has never run — plain @Scheduled method has no Vert.x context (HR000068)
## Symptom

`ConsentExpirationJob` fails on every run in the sandbox:

```
17:05:00 ERROR [ConsentExpirationJob] consent.expiration.sweep FAILED threshold=2026-07-31T17:05:00Z:
java.lang.IllegalStateException: HR000068: This method should exclusively be invoked from a Vert.x
EventLoop thread; currently running on thread 'executor-thread-1'
```

Cron is `0 5 * * * ?` — hourly at HH:05 — so it fails hourly. Observed at the first firing after a
pod start, which is how I ran into it.

## Cause

`sweepExpiredConsents()` is a plain (non-`suspend`) `@Scheduled` method. Quarkus invokes it on a bare
`executor-thread`, which carries no Vert.x context, and `buildSweepPipeline` immediately calls
reactive Panache (`consentRepo.findExpiredActive`).

This is the documented fleet footgun — the same root cause as the `runBlocking`-in-`@Scheduled`
sweep (#2190/#2191/#2194, #2197), reached without `runBlocking`: the method builds a `Uni` and
`subscribe()`s it directly, so the reactive call still starts on the scheduler's thread. The
established fix is a `suspend fun`.

## Consequence

Expired consents are never transitioned out of ACTIVE. `validTo` passing has no effect on stored
state, so a consent that should have lapsed keeps reading as active to anything that inspects the
row rather than re-deriving expiry. For a consent service that is a compliance-relevant gap, not
just a stuck cron.

Worth checking how far it reaches: whether `GET .../active` re-derives expiry (in which case the
live gate is intact and only the projection is stale) or trusts the stored status. I have not
verified that, and it decides the severity.

## Why nobody noticed

The failure is an `ERROR` log inside the subscriber's failure callback. Nothing goes red: the pod
stays ready, the scheduler keeps firing, and the metric for "sweep ran and expired N" is only emitted
on success (and only when `count > 0`). A job that has never once succeeded looks identical to a job
with nothing to do.

That also means the existing guard (`check-no-runblocking-in-scheduled.py`) does not catch this
shape — there is no `runBlocking` here. A guard keyed on "reactive Panache reachable from a
non-suspend `@Scheduled` method" would; a guard keyed on the `runBlocking` spelling will not.

## Fix direction

Make `sweepExpiredConsents` a `suspend fun` and await the pipeline, per the fleet convention.
`StandingOrderExecutionSweepIT` / `LedgerSchedulerVertxContextIT` are the pattern for testing it —
driving the real cron from a `@TestProfile` that re-enables the scheduler, since a test calling the
method directly supplies the very Vert.x context the scheduler does not, and passes against broken
code.

Refs #2749

JiRaska:
Closes #2913

## The defect

`ConsentExpirationJob.sweepExpiredConsents()` was a plain (non-`suspend`) `@Scheduled` method that built a `Uni` and `subscribe()`d it. Quarkus invokes such a method on a bare `executor-thread` carrying no Vert.x context, so the first reactive Panache call threw `HR000068` — hourly, for the life of the service. **The sweep had never once succeeded**: expired consents stayed `ACTIVE` and no `ConsentExpired` event ever reached a consumer (GDPR Art. 17 / PSD2 RTS Art. 10).

## Why the existing gate could not see it

`check-no-runblocking-in-scheduled.py` was keyed on the `runBlocking` **spelling** from #2148/#2187. This defect has no `runBlocking` anywhere. `subscribe()` reads as "hand this off, it runs elsewhere" and does not — the subscription starts on the *caller's* thread, which is the contextless executor-thread.

So the gate now checks the **property**: a `@Scheduled` method that starts reactive work must be `suspend`, or declared to return `Uni<…>` (the other shape Quarkus dispatches on a proper duplicated context — four services already use it, and they are correctly not flagged).

## What the property check found

Two more jobs in the identical shape, both also never having run:

| job | consequence |
|---|---|
| `DeviceTokenSweepJob` (ADR-0135) | no stale push token was ever retired |
| `NotificationOutboxDeadLetterJanitorJob` (ADR-0050 N5) | no DEAD outbox row was ever purged |

Both call `Panache.withTransaction` behind the port, so both threw `HR000068` on every firing. Both fixed the same way.

## Evidence

- **Fail-first, against the real cron.** `ConsentExpirationSweepVertxContextIT` drives the actual scheduler via a `@TestProfile` with a 2-second cron. Against the old shape it fails with the exact error from the issue: `HR000068: This method should exclusively be invoked from a Vert.x EventLoop thread; currently running on thread 'executor-thread-1'` (`failures="1"`). With the fix, green. A test that calls the method *directly* cannot see this — it supplies the very context the scheduler does not, which is why the five existing `ConsentExpirationJobTest` cases were green against broken code the whole time.
- **The gate was validated against a known positive, in both directions.** Reverting `DeviceTokenSweepJob` alone makes it print an `::error` naming that method; with all three fixed it reports `62 @Scheduled method(s) checked … clean.` A stale `nonsuspend_reactive_allowlist` entry is itself a violation, so an exception cannot quietly outlive its reason.
- `ktlintCheck`, `detekt`, `test`, `koverVerify`, `quarkusBuild` green for both services (run sequentially — two Quarkus test suites in parallel collide on port 8081).

## Notes for the reviewer

- The sweep's cron became configurable (`openbank.consent.expiration-cron`, default unchanged at `0 5 * * * ?`) purely so the IT can shrink it, plus `ConcurrentExecution.SKIP` to match the fleet.
- `rules.yaml` gained `no_reactive_from_nonsuspend_scheduled_body` + the new allowlist, so **all 39 OPA bundle generators restamped** — regenerated as the last step before pushing, per the ADR-0029 note. That is the bulk of the diff (~67 of 74 files) and none of it is hand-edited.
- The three `@Suppress("TooGenericExceptionCaught")` are the fleet idiom for a cron that must not die on one bad tick; each carries its reason inline.


JiRaska:
## Independent check: are these three the whole set?

Ran the widened gate from this branch against **unmodified `main`** — the gate script and
`rules.yaml` copied onto a clean `origin/main` worktree, so the only thing under test is the
detection:

```
::error ConsentExpirationJob.kt:45              sweepExpiredConsents
::error DeviceTokenSweepJob.kt:39               sweepStaleTokens
::error NotificationOutboxDeadLetterJanitorJob.kt:42   purgeDeadLetters
check-no-runblocking-in-scheduled: 63 @Scheduled method(s) checked across 1716 main-source
file(s) — VIOLATIONS above.
```

Exactly the three this PR fixes. Good, but that only proves the gate finds what the gate looks
for, so I went after the residual: **the check reads reactive tokens in the method body, and a
scheduler that merely calls an injected port has none.**

Enumerated every non-`suspend`, non-`Uni` `@Scheduled` method on `main` — 12 — and checked the 9
the gate does not flag, one at a time, rather than assuming the gate's silence:

| method | why it is genuinely fine |
|---|---|
| `statement-service.refresh` | returns `Uni<Void>` — the gate's other accepted shape |
| `agent-service.scheduledSweep` | `runBlocking`, deliberately allowlisted (no reactive persistence on its classpath) |
| `agent-service.refresh` | reads the proposals store, which is blocking Agroal JDBC |
| `analytics-sink.scheduled` | allowlisted — drops the reactive stack entirely |
| `audit-service.register` | registers Micrometer gauges, no persistence |
| `libs-runtime.registerWorkflowLiveness` | gauge registration only |
| `pid-service.refresh` | verifies a signed trusted list into an in-memory trust store |
| `ledger-service.onStart` | `@Observes StartupEvent`, not a scheduler |
| `security-scanner.scheduledScan` | the one that needed real checking — see below |

`security-scanner` **does** carry `hibernate-reactive-panache` and a Panache outbox repository, so
"no reactive stack" is not an argument there. But the scheduled path does not reach it:
`SecurityScannerService` is constructed with a `Clock` and nothing else, and `scanAll` is
`java.net.http` probes writing to in-memory maps. The outbox is touched only by
`SecurityOutboxDispatcher` / `SecurityOutboxBacklogGauge`, whose port is `suspend` throughout.

**So the set is complete** — and the one case where the gate's heuristic could have hidden a fourth
instance was resolved by reading the call graph, not by trusting the green.

Worth recording for whoever extends this later: the check's blind spot is real, just currently
unoccupied. A future `@Scheduled fun sweep() { service.doIt() }` whose `doIt` is reactive would
pass. Closing that needs call-graph analysis rather than a token scan, which is a much bigger
change than this PR should carry.

Issue #2872, opened 2026-07-31
No gate proves a service component can boot — five first-second failures shipped to main unnoticed


JiRaska:
## What happened

Rolling campaign-service into the sandbox (#2749) hit **five** independent defects. Every one of
them kills the pod, the sidecar, or the pod's readiness within the first seconds of the process, and
every one was on `main`, merged, with all required checks green.

| # | PR | Symptom | Mechanically detectable from the repo alone? |
|---|----|---------|-----------|
| 1 | #2851 | `secrets "campaign-service" is forbidden` — service missing from the ESO cert-reader `resourceNames` allow-list | **Yes** — the component declares a `KafkaUser` and an ExternalSecret reading it; the Role's allow-list is a separate hand-kept list |
| 2 | #2858 | `Failed to load config value ... analytics.clickhouse-user` — credentials never wired | Partly — a non-Optional `@ConfigProperty` with `defaultValue = ""` and no env supplying it |
| 3 | #2865 | `InactiveBeanException: ReactiveRedisDataSource` — no `quarkus.redis.hosts`, no Redis workload | **Yes** — the service produces `RedisApprovalStore`; every other such service declares the config and ships a Redis |
| 4 | #2865 | `/bundle/rules-data.yaml: merge error` — OPA ran in directory mode with flat mounts | **Yes** — the bundle files are bit-identical to every healthy service's; only the args and mount paths differ from the fleet shape |
| 5 | #2870 | probes on 8085 refused — no `quarkus.management` block | **Yes** — the Deployment probes and the PodMonitor name a port the service config never opens |

Four of the five are pure repo-internal contradictions: two artifacts in this tree disagree, and
nothing compares them.

## Why the existing gates could not catch it

They each check one artifact in isolation. `Validate manifests` lints YAML and runs the governance
scripts; the service build compiles and runs tests; the OPA gate checks bundle drift; the netpol
generator derives edges. Nothing reads a Deployment and the service config *together* and asks
whether the described process could start. Unit tests can't: `@ApplicationScoped` is lazy, the
management interface is disabled under `%test`, and no test constructs the beans whose config is
missing.

So the failure mode is the familiar one — the controls were green about something they never
examined.

## Proposed checks

Cheap and mechanical, in rough order of value per line of script:

1. **Probe/scrape port has a listener.** For each Deployment, collect the ports named by
   `livenessProbe`/`readinessProbe` and by any PodMonitor selecting it; assert the service's
[...]

PR #3009, opened 2026-08-01
fix(ci): gate that the Deployment and the service config name the same ports (#2872)

JiRaska:
> **Corrected after live verification.** The first version of this PR claimed a defect in
> `openbank-customer-edge` and gated on `quarkus.management.enabled`. **Both were wrong.** See
> "What I got wrong" below. The PR is now the gate only; customer-edge is untouched.

## Summary

Adds the gate for the #2872 class — **the Deployment and the service config must name the same
ports**. Rolling campaign-service into the sandbox hit five defects that each kill the pod, the
sidecar or its readiness in the first seconds, all on `main` with every required check green. Four
were pure repo-internal contradictions: two artifacts in this tree disagree and nothing compares them.

This closes the pair behind defect #5.

## Type of change

- [x] fix (bug fix) — CI tooling only, no service code

## Linked issues

Refs #2872

## What I got wrong, and how it was caught

I claimed `openbank-customer-edge` was broken on `main`: it declares `quarkus.management.port: 8085`
with no `enabled: true`, its Deployment probes `/q/health/{live,ready}` on that port, so — I
reasoned — nothing could be listening. I had ruled out `envFrom`, container env, Dockerfile `ENV`,
`application.properties` and a libs-supplied default, and all 40+ other services set `enabled: true`.
The inference looked safe.

It was wrong. Checked against the live pod:

| Evidence | Result |
|---|---|
| Pod status | `2/2 Running`, **0 restarts**, up 9h |
| Deployed image | `sandbox-20e6d235` — built from a commit whose `application.yaml` has **no** `enabled` key |
| `curl 127.0.0.1:8085/q/health/ready` inside the container | **`status: UP`** |

Quarkus 3.38 brings the management interface up from the port declaration alone. There was no defect.

**Two things followed, both fixed here:**

1. **The gate was wrong, not just the prose.** Requiring `enabled: true` would flag any service
   relying on the same implicit behaviour — false positives on correct services, which is how a gate
   stops being read. It now compares *declared* ports and does not re-derive Quarkus's activation
   rules; a gate that models the framework badly is a second, worse copy of it.
2. **The customer-edge change is dropped.** It was not broken, so `fix(customer-edge)` would have cut
   a release for a non-fix and left a false diagnosis in that service's changelog.

The general lesson, which is the reason the original PR asked for this check: exhaustively ruling out
the ways a setting *could* arrive is not the same as observing the system. The one source I could not
enumerate from the repo was the framework's own default behaviour.

## What the gate does

For every Deployment/Rollout under `gitops/components`, each container mapping to an `openbank-*`
module must have every probed and PodMonitor-scraped port **declared** by its own `application.yaml`
(`quarkus.http.port` or `quarkus.management.port`). Scope derives from the committed manifests and
the tree, so a new service is covered the day its component lands.

**No existing gate could catch this class.** `Validate manifests` lints YAML in isolation; the
service build compiles and runs tests; unit tests are structurally blind, because the management
interface is off under `%test` and `@ApplicationScoped` is lazy.

## Re-validated in three directions after the correction

| Case | Expected | Result |
|---|---|---|
| campaign-service with its management block removed (the real defect #5) | red | ✅ red |
| a port-only declaration — customer-edge's actual shape | clean | ✅ clean |
| the real tree, 53 service containers | clean | ✅ clean |

## This is the cheaper half of the trade, and #2872 says so

Measured against that rollout: four static checks like this would have caught **~3 of 17** defects; a
boot smoke test — image + ephemeral Postgres/Redis + the Deployment's env, assert readiness — would
have caught **~8**, including the missing-config and inactive-bean classes this cannot see. I built
the static check because it is verifiable without Docker on a host whose Gradle cache is currently
corrupting artifacts. **The boot test remains the better answer for the rest.**

## Definition of Done

- [x] Gate added, scope derived, validated in both directions — after the correction.
- [x] No service code touched; no generated artifact hand-edited.
- [ ] Unit tests / OpenAPI / Flyway / contract tests — not applicable.
- [ ] Local `./gradlew` — not run (host cache corruption); irrelevant here, this PR adds no Kotlin.

## Security checklist

- [x] No secrets, no new dependency, no `@Suppress`.

## Compliance impact

- [ ] None — CI tooling.

### DIFF
--- a/.github/scripts/check-probe-port-has-listener.py
+++ b/.github/scripts/check-probe-port-has-listener.py
@@ -29,8 +29,10 @@ are not built here):
   * every port named by a `livenessProbe` / `readinessProbe` / `startupProbe`, and
   * every port a `PodMonitor` selecting that workload scrapes
 
-must be opened by the service's own `application.yaml` — `quarkus.http.port`, or
-`quarkus.management.port` when `quarkus.management.enabled` is true.
+must be DECLARED by the service's own `application.yaml` — `quarkus.http.port` or
+`quarkus.management.port`. The check is deliberately about whether the two artifacts name the same
+ports, not about whether Quarkus would bind them: re-deriving the framework's activation rules here
+would just be a second, less accurate copy of them.
 
 Scope is DERIVED: the containers come from the committed manifests and the modules from the tree, so
 a new service is covered the day its component lands. There is no list to keep in step.
@@ -86,16 +88,21 @@ def service_ports(module: str) -> tuple[set[int], list[str]]:
     if isinstance(http_port, int):
         ports.add(http_port)
 
-    management = quarkus.get("management") or {}
-    # Quarkus only binds the management interface when it is explicitly enabled. A `port` with no
-    # `enabled: true` opens nothing — which looks exactly like a correct config at a glance, and is
-    # the shape that made defect #5 invisible in review.
-    if management.get("enabled") is True:
-        mgmt_port = management.get("port")
-        if isinstance(mgmt_port, int):
-            ports.add(mgmt_port)
-    elif management:
-        notes.append(f"{module}: quarkus.management exists but `enabled` is not true — it binds nothing")
+    # Any DECLARED management port counts, with or without an explicit `enabled: true`.
+    #
+    # An earlier version of this gate required `enabled: true` and flagged openbank-customer-edge,
+    # which declares only a port. That was wrong, and it was disproved against the live pod: the
+    # deployed image is built from a commit with no `enabled` key, the pod is 2/2 with zero
+    # restarts, and `/q/health/ready` on 8085 answers UP. Quarkus 3.38 brings the management
+    # interface up without it. Modelling Quarkus's activation rules here would make this gate a
+    # second, worse copy of them — and a gate that is wrong about the framework produces false
+    # positives on correct services, which is how a gate gets ignored.
+    #
+    # Comparing DECLARED ports still catches the defect this exists for: campaign-service had no
+    # `quarkus.management` block at all, so the probed 8085 matched nothing.
+    mgmt_port = (quarkus.get("management") or {}).get("port")
+    if isinstance(mgmt_port, int):
+        ports.add(mgmt_port)
 
     return ports, notes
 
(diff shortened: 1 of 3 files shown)


JiRaska:
## This duplicates #2984, which merged about an hour ago

All five required checks here are green, so nothing in CI will stop this — but `main` already carries a guard doing the same job:

```
main:       .github/scripts/check-probe-port-listener.py       (#2984, merged 04:15Z)
this PR:    .github/scripts/check-probe-port-has-listener.py
```

Same inputs, same comparison, same defect. Both:

- iterate every Deployment/Rollout under `openbank-infra/gitops/`
- for each `openbank-*` container, take the ports named by its probes **and** by any `PodMonitor` selecting that workload
- require each to be declared by the service's own `application.yaml` (`quarkus.http.port` / `quarkus.management.port`)
- cite the same defect: port 8085 probed and scraped while the service config had no `quarkus.management` block, so nothing listened there

Both reference #2872; #2984 also names #2870.

Two parallel agents appear to have solved the same issue independently.

## Why it is worth not just merging anyway

Merging would put **two guards doing the same work** into `Validate manifests`, so both run on every PR in the repo. Beyond the wasted time, it is two places to maintain and two wordings of one rule that will drift — and when they eventually disagree, the disagreement itself becomes the thing someone has to debug.

## What I would check before deciding

I have not compared them line by line, only their stated scope and inputs. Worth a look before closing either:

- **#2984's version is longer** (237 vs 205 lines) and carries a `DELIBERATE LIMITS` section excluding sidecars (`opa`, `flagd`) and non-`openbank-*` workloads, which have no service config to compare against. If this PR's version lacks that, it may flag them.
- Conversely, if this version covers a case #2984's misses, the right move is to fold that case into the merged one rather than keep both.

Either way the reconciliation lands here, since #2984 is already on `main`.

*Found while draining the PR queue — not merging this one for that reason. Everything else about it is green.*

JiRaska:
Closing as superseded — the same gate landed on `main` from a parallel effort:
`.github/scripts/check-probe-port-listener.py`, registered in the new gate manifest as
`probe-scrape-port-has-a-listener` (gitops group, enforced).

It is the same check for the same defect: the Deployment's probes and the PodMonitor name a
management port that the service's `application.yaml` never opens (#2872, #2870). Both
implementations were written from the same campaign-service rollout.

Theirs is the better one to keep — it registers a `--selftest` invocation alongside `--enforce`, so
the gate is falsified on every run rather than only ever having passed. Mine had no equivalent.

The one thing worth carrying over is a finding, not code: the first run of my version flagged
`openbank-customer-edge`, and **that finding was wrong**. Verified against the live pod — 2/2 ready,
0 restarts, `/q/health/ready` answering on 8085 — because Quarkus 3.38 binds the management
interface from the port declaration alone, with no `quarkus.management` block required. Any
implementation of this check that treats a missing `quarkus.management` block as proof of a missing
listener will report that same false positive. Worth confirming against the version on `main`
before trusting a red from it.

JiRaska:
Superseded by check-probe-port-listener.py on main — see above.

Issue #2365, opened 2026-07-25
Money-path services remain NO-GO pending evidence-backed critical attestations


JiRaska:
The #2255 sweep took the Prod Readiness matrix from **44 cells below Verified / GO 11-of-37** to **0 cells below Verified / GO 26-of-37**. Every derived signal the collector can read from the repo is now at 2 (Verified) or better.

**The eleven money-path services are still NO-GO, and no amount of code will change that.** `compute_gate` demands **3** on the three critical dimensions for a money-path service, and 3 comes only from a fresh TTL'd attestation:

| cell | attestation key | what it asserts |
|---|---|---|
| **C1** Kód | `code_complete` | the service is feature-complete for production, not a partial slice |
| **C5** Zálohy | `restore_drill` | a restore from backup was actually performed and verified |
| **C7** Security | `pentest` | an external/automated penetration test was run |

`openbank-libs/governance/attestations.yaml` is currently **empty on purpose**, and its closing comment is the right posture:

> `# --- zatím prázdné: žádná M-dimenze není dosvědčena (a to je pravda) ---`

Nothing is attested, and that is true. This issue is the backlog for making it untrue *honestly* — 3 attestations × 11 currently-gated services = **33**, each requiring a real event first.

## Scope depends on #2364

That number is sized against the 11 services the matrix gates today. #2364 shows `rules.yaml` declares **20** money-path services, of which the collector gates 11, scores 6 with the lenient gate, and omits 3 entirely. **Land #2364 first** or this backlog is scoped against a population that is too small — the true figure is 20 × 3 = 60.

## The uncomfortable part: nothing verifies an attestation is real

`attestations.yaml` is an honour-system file. The only protection is TTL decay — after `ttl_days` the collector stops counting the bonus and the dimension falls back to its derived score. There is **no CI guard** that an entry corresponds to a real event. (`check-fleet-attestations.sh` sounds related but is not: it verifies cosign-signed CycloneDX SBOM attestations on container images, ADR-0030 D4 — a different mechanism entirely.)

So the file's value rests entirely on nobody writing an entry for an event that did not happen. Given the sweep's recurring theme — a comment scoring as a scrape, a filename as an architecture, the word "contract" as a contract test — this is the one remaining place in the matrix where a sentence *is* the evidence. Two things follow:

[...]

PR #852, opened 2026-07-11
chore(main): release admin-ui 0.44.0

JiRaska:
:robot: I have created a release *beep* *boop*
---

### Features

* **gitops:** ADR-0030 D5 phase 1 — money-path GitOps-sync drift scanner ([#890](https://github.com/JiRaska/open-bank-oss/issues/890)) ([2e3aa4e](https://github.com/JiRaska/open-bank-oss/commit/2e3aa4e6a479cbc3e02d4b9f2a52cdc507272bc8)), closes [#265](https://github.com/JiRaska/open-bank-oss/issues/265)

### Bug Fixes

* **admin-ui:** day-end flags a never-run / stale EoD tie-out instead of "no data" ([#864](https://github.com/JiRaska/open-bank-oss/issues/864)) ([8f1a756](https://github.com/JiRaska/open-bank-oss/commit/8f1a7562adb0ae71c6e80275c7220eaf67211a69)), closes [#855](https://github.com/JiRaska/open-bank-oss/issues/855)
* **admin-ui:** remove unused Link import in iaops page ([#870](https://github.com/JiRaska/open-bank-oss/issues/870)) ([2a44a96](https://github.com/JiRaska/open-bank-oss/commit/2a44a96ef7830c51854ef5550ced1448fac9eb48)), closes [#865](https://github.com/JiRaska/open-bank-oss/issues/865)
* **admin-ui:** show fx-service scale-to-zero state on the FX page ([#851](https://github.com/JiRaska/open-bank-oss/issues/851)) ([d428007](https://github.com/JiRaska/open-bank-oss/commit/d428007896a88b6e0c9365da8caf5028748d51bc)), closes [#759](https://github.com/JiRaska/open-bank-oss/issues/759)

---
This PR was generated with [Release Please](https://github.com/googleapis/release-please). See [documentation](https://github.com/googleapis/release-please#release-please).

PR #3024, opened 2026-08-01
fix(ci): repair the authenticated fuzz harness and derive its money-path scope (#2365)

JiRaska:
## Summary

The authenticated fuzz lane **had never worked** — 0 successful runs, ever. I dispatched it for the
first time, it failed, and the cause turned out to be the realm file's **name**.

`git mv openbank-fuzz-realm.json openbank-realm.json` is the whole fix. Plus the money-path scope
this PR was originally asked for.

## Type of change

- [x] fix (bug fix) — CI only, no service code

## Linked issues

Refs #2365, Refs #852, Refs #266

## Root cause

Keycloak's directory import derives the **realm name from the filename**, `<realm>-realm.json`. The
file was `openbank-fuzz-realm.json`, so Keycloak expected a realm called `openbank-fuzz` while the
JSON declares `openbank`. It imported `openbank` — which is why the log said so, misleadingly — then
could not bind a session to the realm it had been told to expect:

```
Realm 'openbank' imported
ERROR: Failed to start server in (development) mode
ERROR: Session not bound to a realm
```

**The file's contents were always correct and are unchanged here.** The commit is a pure `git mv`.

## I got this wrong first, and the mistake is the more useful part

I hypothesised the service-account user was missing the `email`/`firstName`/`lastName` that KC 26's
user profile requires, built variants under `/tmp`, saw them start, and committed that as the fix.
It failed on CI identically.

The variants worked because I had written them to `/tmp/kc-*/realm.json` — **I renamed the file
without noticing**, since the name looked like an irrelevant detail of scaffolding a test. The one
variable I simplified away was the only one that mattered.

Isolated afterwards, four containers:

| Content | Filename | Result |
|---|---|---|
| variant C | `realm.json` | starts, token, roles |
| variant C — *same bytes* | `openbank-fuzz-realm.json` | ❌ `Session not bound to a realm` |
| **ORIGINAL, untouched** | `openbank-realm.json` | ✅ starts, token, `ROLE_OPERATOR` + `ROLE_COMPLIANCE` |
| profile-field version | `openbank-realm.json` | ✅ same |

Row 3 settles it: the original contents work as soon as the filename matches. The profile fields were
never needed, and the change adding them has been dropped.

A `# WHY THE REALM FILE IS NAMED …` comment is now in the workflow header so nobody renames it back
to something more descriptive in good faith.

## Every check is "does the JWT carry the roles"

Not "did the container start". Dropping the `users` block *also* starts and mints a token, but
`realm_access.roles` comes back as `[default-roles-openbank, offline_access, uma_authorization]` — no
`ROLE_OPERATOR`. The fuzz would `403` on every authenticated endpoint and report a green "no 5xx"
having exercised nothing. That variant is the trap, and only decoding the JWT distinguishes it.

## Scope

Single-service input → matrix derived from `rules.yaml: money_path_services` and each service's own
config. Eligibility is a **property**, not a maintained list.

**16 in, 4 excluded out loud:**

| Excluded | Why |
|---|---|
| `sdd`, `sanctions`, `vop` | `AUTHZ_ENFORCE` defaults **true** — need the OPA sidecar + `rest.rego` (#266) |
| `settlement` | no `jdbc:postgresql` URL — the harness cannot provision its DB |

`fail-fast: false`, `max-parallel: 6`, per-service artifacts, so one service cannot mask another.

## Verification

| Check | Result |
|---|---|
| Failure reproduced locally via the **exact CI invocation** | ✅ |
| Root cause isolated by varying filename vs. content | ✅ (table above) |
| Fix: start → `client_credentials` → JWT roles, locally | ✅ |
| Derive step, **extracted from the YAML** and run | ✅ 16 in, 4 excluded with reasons |
| `actionlint` finding sets vs `main` | ✅ identical |
| **CI: step 8 "Start Keycloak"** on [30686687793](https://github.com/JiRaska/open-bank-oss/actions/runs/30686687793) | ✅ **success — the fix is proven on the runner** |
| CI: remaining steps (service boot, fuzz) | still running at the time of writing |

The step this PR exists to fix is green on CI. The later steps are the lane doing its actual job for
the first time; I will post their outcome below. **A red at the fuzz step would be a finding against
the service, not a defect in this PR** — and it belongs in its own issue, not here.

## Deliberately not promoted to a schedule

The workflow header makes a green manual run the precondition. That now holds for the harness, not
yet for all 16 services. A weekly cron over sixteen never-booted services would manufacture exactly
the "unvetted weekly red" that comment warns against.

## Definition of Done

- [x] Root cause isolated by experiment, not inferred — and the first, wrong diagnosis withdrawn.
- [x] The plausible-but-wrong fix tested and rejected with evidence.
- [x] Scope derived; exclusions printed, never silent.
- [x] Harness fix confirmed on CI (step 8).
- [ ] Unit tests / OpenAPI / Flyway — not applicable; CI only.

## Security checklist

- [x] No real secrets — Keycloak admin and the client secret are job-local throwaways, unchanged.
- [x] No new dependency; schemathesis stays hash-locked.
- [x] `permissions: contents: read` unchanged.

## Compliance impact

- [ ] None directly — supplies evidence toward the `pentest` attestation keeping money-path services
      NO-GO (#2365).


JiRaska:
## ⚠️ Verification run FAILED — this PR is not ready

I said I would report this rather than lean on the local reproduction. Doing that.

**Run [30686184260](https://github.com/JiRaska/open-bank-oss/actions/runs/30686184260)**, dispatched
from this branch at `b6faa5333`:

| Job | Result |
|---|---|
| `derive fuzz scope` | ✅ success — 16 services, 4 excluded with reasons |
| `schemathesis auth (openbank-consent-service)` | ❌ **failure at step 8, "Start Keycloak"** |

Same error as before the fix:

```
Realm 'openbank' imported
ERROR: Failed to start server in (development) mode
ERROR: Session not bound to a realm
```

**The fix was in the run.** Confirmed the realm JSON at `b6faa5333` carries `email`,
`firstName` and `lastName` on the service-account user, so this is not a stale-checkout artefact.

### So the diagnosis is incomplete, not the fix incomplete

Locally, against `quay.io/keycloak/keycloak:26.0`, the same content **starts**, mints a
`client_credentials` token, and the JWT carries `ROLE_OPERATOR` + `ROLE_COMPLIANCE`. On the runner it
does not. That means the missing user-profile fields were **at most part** of the cause, and my
local test differed from CI in some way I have not yet identified — different mount shape (a
directory containing the real filename vs. a lone `realm.json`), different host architecture, or
something else.

Two possibilities, and I am not going to guess between them in the PR body:

1. The profile fields are necessary but not sufficient, and something else in the realm JSON also
   trips KC 26 on that platform.
2. My local reproduction was not equivalent to the CI invocation, which would make the "reproduced
   byte-identical" claim in the description **too strong** — the failure reproduced, but the *fix*
   was validated against a setup that is not the one that matters.

Either way the honest state is: **the harness is still broken and this PR does not fix it yet.**

### What stands regardless

- The lane genuinely had **0 runs, ever**, and its first two runs both failed at the same step. That
  finding is unaffected.
- Variant A (drop the `users` block) is still the trap: it starts and mints a token whose
  `realm_access.roles` lacks `ROLE_OPERATOR`, so the fuzz would 403 everywhere and report green.
  Whatever the eventual fix is, it must be checked by decoding the JWT, not by "Keycloak came up".
- `derive fuzz scope` works: 16 money-path services in, 4 excluded out loud (sdd/sanctions/vop need
  the OPA sidecar per #266; settlement has no JDBC URL).

### Next

Re-run the committed file locally through the *exact* CI invocation — same directory mount, same
filename, same flags — rather than the simplified variant I used. If it passes there too, the
difference is the runner platform and the diagnosis moves to the container logs with `--verbose`.

Marking this draft until a green run exists.

JiRaska:
## Verification run: harness **fixed**, and it immediately found a real bug

Run [30686687793](https://github.com/JiRaska/open-bank-oss/actions/runs/30686687793) from this branch:

| Step | Result |
|---|---|
| `derive fuzz scope` | ✅ 16 services in, 4 excluded with reasons |
| 8. **Start Keycloak (throwaway realm)** | ✅ **success** — the rename fixes it, confirmed on the runner |
| 9. Fuzz authenticated surface | ❌ failure — **a finding, not a harness defect** (below) |
| 10. Upload fuzz reports | ✅ |

The run is red, and that is the lane working. Step 8 is the one this PR exists to fix and it is
green; step 9 got as far as booting the service, obtaining a bearer token, and generating **468 test
cases** — none of which had ever happened before, because Keycloak had never started.

## What it found

**`PATCH /api/v1/consents/approvals/{id}` returns 500 on a `null` JSON body.** Filed as #3029.

```
[500] {"traceId":"8701067a-…","status":500,"code":"INTERNAL_ERROR", …}

curl -X PATCH -H 'Authorization: …' -H 'Content-Type: application/json' \
  -d null http://localhost:8106/api/v1/consents/approvals/1ffa0cda-…
```

That is the four-eyes approval endpoint on a money-path service, and the property being violated —
*no input, however malformed, produces a 5xx* — is precisely the `pentest` dimension keeping
money-path services NO-GO in #2365. The very first working run of this lane produced exactly the kind
of evidence it was built to produce.

## Two harness observations from the same run, also in #3029

1. `GET /api/v1/consents/grantee/{granteeId}` tripped schemathesis's `filter_too_much` health check —
   its parameter constraints are tight enough that the operation is effectively untested.
2. **5 operations repeatedly returned 404**, so the fuzz never reached their core logic. The lane
   boots against an empty database, so the authenticated surface is currently being tested mostly at
   its 404 boundary. Seeding a fixture would deepen this considerably.

Neither is a defect in this PR; both are follow-ups now that the lane runs at all.

## Where that leaves this PR

The harness fix is confirmed on CI, which was the open question. The red belongs to a service bug
that has its own issue, and to the depth limitations above — not to this change.

Ready for review.

JiRaska:
## Not merging: the harness still fails to boot 15 of its 16 targets

The five required contexts are green, so nothing blocks this — but every `schemathesis auth` check except one is failing on `93ab2745a`:

```
success   openbank-clearing-service
failure   account, balance, billing, consent, domestic-payment, fraud, fx,
          interest, ledger, lending, sca, sepa-instant, sepa-payment,
          swift, transaction                                   (15)
```

The failing step is `Fuzz authenticated surface` in all of them, and the failure is **before any fuzzing happens** — the service never comes up:

```
##[error][openbank-ledger-service] failed to boot
  at io.quarkus.vertx.http.runtime.VertxHttpRecorder$WebDeploymentVerticle.setupTcpHttpServer
  at io.vertx.core.http.HttpServer.listen
FAILURE: Execution failed for task ':openbank-ledger-service:quarkusDev'
```

Spot-checked `openbank-fx-service` and `openbank-sca-service` — same step, same `failed to boot`. So this is the harness, not fuzz findings.

`openbank-clearing-service` passing is the interesting part: whatever the cause, it is not uniform across all targets, so it is probably not "the harness is entirely broken" but something these 15 share and clearing-service does not.

## What I could NOT establish

**Why the bind fails.** The exception message is interleaved with the parallel `docker pull` output and I did not isolate it. Could be an occupied port, a missing database, or config — I am not asserting which. Ledger was trying port 8101.

Worth pulling the `fuzz-reports/<svc>-boot.log` artifact rather than reading the job log; the script tails only 60 lines of it on failure.

## Why this is worth blocking on

The PR's stated purpose is to *repair* the authenticated fuzz harness. The harness not booting its targets is the thing being repaired, so a green required set says nothing about whether the change works. These checks are non-required, so it would merge silently.

This is #2365 territory — money-path services staying NO-GO — and every one of the 15 is a money-path service.

*Found while draining the PR queue. Everything else on this PR is green, and the head has not moved since I looked (`93ab2745a`).*

JiRaska:
## Update: #3079 landed the same fix, and a re-run here cannot pick it up

Following on from [my earlier comment](https://github.com/JiRaska/open-bank-oss/pull/3024#issuecomment-5150196428) — still 15 of 16 `schemathesis auth` checks failing on `93ab2745a`.

I was going to re-run CI to see whether #3079 (`make the fuzz harness able to boot reactive-only services`, merged 13:19) had fixed it. **It cannot**, and the reason matters:

```
merge-base of this branch:  aec0ec68b   07:52
#3079 merged as:            450453bc4   13:19
```

A re-run replays the *same* SHA — the merge ref computed when the event fired — not a fresh merge with current `main`. So the fleet build would repeat the identical failure, about an hour for no new information. Not worth spending.

## The bigger issue: this and #3079 changed the same file

Both modify `.github/workflows/api-fuzz-authenticated.yml`. Against current `main` this branch differs by **101 lines**, and the two versions have diverged rather than converged:

```
                      main (post-#3079)    this branch
realm mentions              8                  12
bearer mentions             4                   4
```

That is the third instance today of two parallel efforts landing on the same artifact (#3009 vs #2984, #3103 vs #3079, now this).

## Suggested route

**Merge `origin/main` into this branch**, rather than re-run or rebase:

- it brings in #3079's harness fix, which is what the re-run was meant to test
- it creates a fresh event, so CI runs against the current state
- the conflict in `api-fuzz-authenticated.yml` gets resolved once, by whoever knows which version is intended

Rebase would be the worse option here for the same reason it was on #3079: replaying commits against a base that already carries their squashed equivalent turns a content question into a series of conflicts.

I have not touched the branch — the 101-line divergence is an authoring decision, not a mechanical one, and there is a live worktree on it.

*Head unchanged at `93ab2745a` since I looked; `mergeStateStatus` is `CLEAN`, so nothing here is blocked except the failing non-required checks.*

JiRaska:
## Correcting myself: the merge is clean, so I did it

In [my previous comment](https://github.com/JiRaska/open-bank-oss/pull/3024#issuecomment-5150249847) I said resolving `api-fuzz-authenticated.yml` against `main` would need a conflict resolution — "an authoring decision, not a mechanical one" — and left the branch alone. I tested that before repeating it, and it is wrong: `git merge origin/main` auto-merges the file with **no conflict**.

I also said there was a live worktree on this branch. There is not any more, so nothing is being stomped.

```
before   93ab2745a   merge-base aec0ec68b (07:52, predates #3079)
after    e42f4897e   main merged in, signed
```

## Why merge rather than leave it

#3079 (`make the fuzz harness able to boot reactive-only services`) merged at 11:19 and is exactly the fix for the 15 failing `schemathesis auth` targets reported here. A re-run could never pick it up — it replays the same merge ref — so the branch had to move for the question to be answerable at all.

## What this does and does not prove

The merge is **textually** clean. That is not the same as semantically correct: two parallel efforts edited this workflow independently, and git merges text, not meaning. The workflow still parses, has its 2 jobs, and its longest `run:` block is 3641 chars (the 19000 ceiling from #3135 is not in play). Against `main` the file now differs by `+73/-11`, down from 101 divergent lines.

Whether the harness actually boots its targets is what the pipeline now answers. If the 15 come back green, #3079 was the fix and this is ready; if they do not, the failure is something #3079 did not cover and the fresh logs will say which.

I have not merged the PR — only moved the branch so CI can run against current `main`.

*Overnight queue drain.*

JiRaska:
## Correction: these checks are not PR CI, so pushing did not re-run them

I said the pipeline would answer whether #3079 fixed the 15 boot failures. It will not, and it never would have — `api-fuzz-authenticated.yml` has exactly one trigger:

```yaml
on:
  workflow_dispatch:
```

No `pull_request`. So the 15 red `schemathesis auth` checks on `93ab2745a` came from someone dispatching it by hand, not from this PR's CI. After the merge commit the new head `e42f4897e` has 10 workflow runs and 33 check-runs, and **zero** of them are fuzz:

```
schemathesis / fuzz check-runs on e42f4897e:  0
```

That also means the red was never merge-blocking here — but equally, no push, rebase or re-run can ever turn it green. Someone has to dispatch it.

## Dispatched

I ran it against the new head, scoped and cheap:

```
run 30719467756   ref fix/api-fuzz-authenticated-realm   sha e42f4897e
services      openbank-ledger-service openbank-fx-service openbank-clearing-service
max_examples  10
```

Ledger and fx were both in the failing 15; clearing was the single target that passed. That gives a control pair — if the harness is fixed, all three should now boot; if clearing passes and the other two still fail, the cause is something those two share and #3079 did not address.

## Worth considering separately

A fuzz harness that only runs on manual dispatch produces exactly this situation: a PR whose stated purpose is repairing that harness cannot demonstrate the repair, and its red checks are stale artefacts of whenever someone last ran it. Whether that should stay manual is a real question — but it is a separate one from this PR.

*Overnight queue drain.*

JiRaska:
## The dispatch answered it: #3079 fixed the boot, and what is left are real findings

I dispatched the harness against the merged head. The result reframes this PR — **the services boot now**, and the remaining red is the fuzzer working:

```
                        cases    result
clearing-service          291    291 passed
ledger-service            679    2 unique failures   "Server error: 2"
fx-service                747    1 unique failure    "Server error: 1"
```

Compare with what I reported earlier on `93ab2745a`: *"the failure is before any fuzzing happens — the service never comes up"*. That is no longer the case. Boot works; these are 500s found by fuzzing endpoints that came up fine.

So my framing in the two comments above was wrong from the moment #3079 landed. The red looks identical from the check list either way — a `schemathesis auth` failure — which is exactly why I had to read the artifact rather than the job log.

## The three findings

```
ledger  POST /api/v1/journals               {"idempotencyKey": "", …}   -> 500
ledger  POST /api/v1/ledger/fx-revaluation  ?date=   (empty)            -> 500
fx      POST /api/v1/fx/convert             generated payload           -> 500
```

The ledger `?date=` one has the shape #3104 catalogues (a query parameter present but empty, reaching the handler and throwing). The other two look like body-field validation rather than a missing body, so #3032/#3050/#3044 would not cover them — but I have not confirmed that against those PRs' scopes, so treat it as a lead rather than a classification.

Full repro curls are in the run's artifacts (`api-fuzz-auth-reports-*`), which carry the `-boot.log`, the fuzz log and the JUnit XML per service.

## For whoever picks this up

The run was scoped and cheap on purpose: three services, `max_examples=10`. Ledger and fx were both among the failing 15; clearing was the one that passed. That control pair is what makes the conclusion safe — if this were still a harness problem, clearing would have broken too.

A full-fleet dispatch would likely surface more of these, since 10 examples per endpoint is a thin sweep.

```
run 30719467756   sha e42f4897e
```

*Overnight queue drain.*

PR #3028, opened 2026-08-01
fix(ci): add 38 verification-metadata entries missing on a cold cache

### LINKED ISSUE #3017: verification-metadata.xml misses .pom/.module entries that only a cold-cache resolve needs
## Summary

`gradle/verification-metadata.xml` is missing `.pom` and `.module` metadata entries that only a
**cold-cache** resolve needs. A warm cache never re-verifies them, so the gap is invisible to
every normal build and to CI as it runs today.

**Severity correction.** An earlier revision said a cold-cache build "fails". Today it does
not: `gradle.properties` sets `org.gradle.dependency.verification=lenient`, so a missing entry
is a WARNING and the build goes green. The failure arrives when verification graduates to
strict per ADR-0144. Verified both ways under `-Dorg.gradle.dependency.verification=strict`:
with the entries present `BUILD SUCCESSFUL`, with one removed `BUILD FAILED` on exactly that
artifact.

That makes this lower-urgency but not lower-value: `lenient` is precisely what keeps the gap
invisible until the flag flips, at which point every cold-cache build breaks at once.

Found while trying to build a service image from a fresh Docker layer (#3015, and see #3016 for
why that path exists at all).

## Evidence

First failure, on `:openbank-campaign-service:quarkusGenerateAppModel`:

```
Dependency verification failed for configuration ':openbank-campaign-service:quarkusProdRuntimeClasspathConfigurationPlatform'
  One artifact failed verification: hibernate-platform-7.4.5.Final.pom (org.hibernate.orm:hibernate-platform:7.4.5.Final)
```

This is a genuine absence, not cache corruption. The component block exists but carries only the
`.module` artifact, while the neighbouring `7.4.3.Final` block has both:

```xml
<component group="org.hibernate.orm" name="hibernate-platform" version="7.4.5.Final">
   <artifact name="hibernate-platform-7.4.5.Final.module">   <!-- present -->
   <!-- .pom entry absent -->
```

Adding that one hash — sha256 `81bc3e82…`, validated by cross-checking the downloaded pom against
Maven Central's published `.sha1` (`33ce6db4…`, matched) — moved the build forward and
immediately exposed six more, in the very next module (`:openbank-libs-domain:compileKotlin`):

```
jackson-annotations-2.18.2.module
jackson-base-2.18.2.pom
jackson-bom-2.18.2.pom
jackson-module-kotlin-2.18.2.module
kotlin-reflect-1.8.10.pom
kotlin-reflect-2.1.21.pom
```

Those seven were the visible tip. A cold-cache regeneration of this one service's graph
(empty `GRADLE_USER_HOME`, `-Xmx6g`, `BUILD SUCCESSFUL in 11m53s`) puts the real figure at
**38 missing entries across 31 components**:

| artifact type | missing |
|---|---|
| `.pom` | 19 |
| `.jar` | 11 |
| `.module` | 8 |

**Removed by regeneration: 0** — the additions-only constraint holds at this scope, which matters
given that an earlier fleet-wide attempt live-deleted 1295 valid poms.

An earlier revision of this issue claimed the gap was confined to `.pom`/`.module` metadata
artifacts and that jars were safe because they sit on the path every warm build takes. **That was
wrong** — it generalised from the seven artifacts in the first two failures, and 11 of the 38 are
jars. Do not use that model to scope the fix.

**38 is a lower bound.** It covers `:openbank-campaign-service:quarkusBuild` alone; the fleet
figure is unmeasured. What it does establish is the shape of the work: tens per service rather
than hundreds, and no existing entry needs touching.

## Why this matters beyond #3016

The Dockerfile build is only how it was found. The same failure hits any cold `GRADLE_USER_HOME`:

- a brand-new self-hosted runner, or an ARC pod whose cache volume was recreated
- CodeQL's tracing build, which resolves into its own cache
- a fresh contributor clone

(All of these warn today and fail once strict lands — not the other way round.)

`CLAUDE.md` already records this failure class ("Fresh-cache builds fail dependency-verification
on artifacts nobody has hashed yet"), but as a per-incident workaround. This issue is to close the
gap systematically instead of one artifact at a time.

## Constraints on the fix

The house norm is **additions-only**. A blanket `--write-verification-metadata sha256` rewrites the
whole file: reorder noise, unrelated components, and — measured previously — deletion of legitimate
entries (one earlier attempt live-deleted 1295 valid poms). So:

- regenerate into a **throwaway copy**, diff against the committed file, and apply only the
  **added** components to the original
- `origin="Maven Central"` for hand-added entries (the existing convention: 69 entries already use
  it, vs 5046 `origin="Generated by Gradle"`)
- validate any hand-computed sha256 against Maven Central's published `.sha1` before committing it,
  so a corrupted download cannot be laundered into the gate file

## Acceptance criteria

- [ ] a cold-cache `GRADLE_USER_HOME` build of at least one service completes without a
      dependency-verification failure
- [ ] no existing entry removed or reordered — diff is additions-only
- [ ] a CI job that periodically resolves on a cold cache, so the gap cannot silently reopen
      (without it this regresses the moment a dependency is bumped)

JiRaska:
## What

Adds **38 missing entries** to `gradle/verification-metadata.xml` — 19 `.pom`, 11 `.jar`,
8 `.module`, across 31 components. Nothing is removed or reordered.

## Why

The file is maintained by warm-cache builds. Anything a warm cache does not re-resolve was never
hashed, so a build with an empty `GRADLE_USER_HOME` fails:

```
Dependency verification failed for configuration ':openbank-campaign-service:quarkusProdRuntimeClasspathConfigurationPlatform'
  One artifact failed verification: hibernate-platform-7.4.5.Final.pom (org.hibernate.orm:hibernate-platform:7.4.5.Final)
```

This is a genuine absence, not cache corruption: the `7.4.5.Final` component carried only its
`.module` entry, while the neighbouring `7.4.3.Final` had both `.module` and `.pom`.

Who hits it: a fresh self-hosted runner, an ARC pod whose cache volume was recreated, CodeQL's
tracing build, a new contributor clone.

## How the number was obtained

Measured, not guessed. A cold regeneration of `:openbank-campaign-service:quarkusBuild` against an
empty `GRADLE_USER_HOME` (`BUILD SUCCESSFUL in 11m53s`) produced exactly these 38 additions and
**zero** removals.

Two earlier attempts were discarded, and both are worth knowing about:

- one died on `Java heap space` after 20m35s yet **exited 0**, because the shell chain ended in a
  `cp`. The regenerated file was byte-identical to its input, which reads exactly like "no gap
  found" rather than "the build died"
- one finished in 2m58s and looked clean, but had inherited the 2.3 GB cache the failed run left
  behind. Being warm, it never went back for `hibernate-platform-7.4.5.Final.pom` — the very
  artifact that opened the issue

Both were caught by checking that a **known-missing artifact actually got written**, not by the
exit code and not by the diff. Anyone automating this needs that assertion.

## Trust in the hashes

Every one of the 38 sha256 values was downloaded from Maven Central and cross-checked against the
published `.sha1` before being written here. All 38 matched, and each also matched the value Gradle
generated independently. A corrupted download therefore cannot be laundered into the gate file —
which matters, since this file *is* a security control and "add the hash the build showed you" is
the exact failure mode it exists to prevent.

They carry `origin="Maven Central"`, following the existing convention for hand-added entries
(69 already use it, vs 5046 `origin="Generated by Gradle"`).

## Additions-only

158 lines added, **0 removed**, XML well-formed, no duplicate artifacts, and the resulting artifact
set is exactly the previous set plus those 38.

A full `--write-verification-metadata` rewrite was deliberately **not** committed: it reindents the
entire file (~18.5k lines churned), burying the change in reorder noise, and that approach has
previously deleted legitimate entries. The additions were extracted from the regenerated file and
applied to the committed one.

## Scope

This covers one service's dependency graph. The fleet-wide gap is larger and stays open — this does
not close #3017.

Also unresolved there: `--write-verification-metadata` OOMs at the default heap and needs
`-Dorg.gradle.jvmargs=-Xmx6g`, so "just regenerate it" is not yet an executable instruction.

## Linked issues

Refs #3017


JiRaska:
## Verified by build, with a falsifiable test

Empty `GRADLE_USER_HOME`, `-Dorg.gradle.dependency.verification=strict`:

| run | result |
|---|---|
| this branch | `BUILD SUCCESSFUL in 13m 19s`, `quarkus-app/quarkus-run.jar` produced |
| same, with **one** of the 38 entries removed | `BUILD FAILED in 10m 38s` |

The failure was precisely the removed artifact and nothing else:

```
Dependency verification failed for configuration ':openbank-campaign-service:quarkusProdRuntimeClasspathConfigurationPlatform'
  One artifact failed verification: hibernate-platform-7.4.5.Final.pom (org.hibernate.orm:hibernate-platform:7.4.5.Final) from repository maven
```

The negative run is the part that makes the positive one mean anything. It establishes three things
at once: the `-D` flag really does override `lenient` from `gradle.properties`, the test is capable
of failing, and the 38 entries are both necessary and sufficient for this service under strict.

**Why the negative control was needed.** The first verification run of this branch was green and
proved nothing: `gradle.properties` sets `org.gradle.dependency.verification=lenient`, so
verification failures are warnings, not errors. That build would have passed without this PR
applied at all. A green run under `lenient` carries no information about this change.

The metadata file was restored from a `cp` backup afterwards; `git diff` is clean, so the branch
under review is unchanged by the experiment.

## Still a lower bound

This is `:openbank-campaign-service:quarkusBuild`. Other services will need their own additions —
see #3017.

PR #3043, opened 2026-08-01
feat(observability): make fleet LLM spend and reliability observable in Prometheus

### LINKED ISSUE #3042: AI/LLM spend and reliability are unobservable, and the one alert that claimed otherwise cannot fire
### Summary

Nothing anywhere reports how many tokens this bank's AI features consume, how often an LLM call fails, or whether a given agent's model is switched on at all. The one alert that claimed to could never fire.

### Findings

**1. `FinOpsAgentTokenBudgetExceeded` is inert.** It is written against `langfuse_agent_tokens_total`, `langfuse_model_cost_per_token` and `langfuse_agent_daily_budget_usd`. Langfuse appears in **no manifest anywhere in gitops** and has never been deployed, so none of those three series has ever existed. The alert's own description says "will be absent until the bridge is deployed" — which is the trap: it reads as *pending* rather than as *inert*, and the ADR-0112 D6 detector looks covered on anything that counts rules.

**2. The LiteLLM gateway is not scraped, and cannot usefully be.** Its Prometheus exporter is a LiteLLM **Enterprise** feature (docs.litellm.ai/docs/proxy/prometheus, "Enterprise LiteLLM Only"). So the natural choke point cannot report token spend for us at all.

**3. There is no AI dashboard.** No Grafana board covers LLM cost, throughput or reliability.

**4. `openbank.agent.tokens_total` is an OpenTelemetry span attribute, not a metric.** It lands in Tempo, so it is unavailable to a Prometheus alert or a budget rule. This is easy to misread as existing coverage.

**5. Six of the eight control-plane agents make no LLM call at all.** A survey of every `/chat/completions` reference found eleven sites; six are stub `LlmDiagnosisAdapter`s (finops-agent, governance-auditor, docs-truth-agent, flaky-test-hunter, authz-policy-auditor, release-steward) still carrying the ADR-0112 P4 TODO. Three code paths make real calls.

### Fix

Count tokens client-side on the three real paths via a shared pure-domain port implemented by the existing `DomainMetrics` bean, emitting `openbank_llm_requests_total`, `openbank_llm_tokens_total` and `openbank_llm_call_duration_seconds`. Put price per token in a Prometheus recording-rule price book so a rate-card change is a gitops edit, not a redeploy of every caller. Replace the dead alert with four that depend on nothing undeployed, and add the missing dashboard.

### Follow-ups not in scope here

- Per-agent budgets need LiteLLM virtual keys backed by a Postgres the gateway does not have. The fleet-wide USD/day ceiling is what is enforceable today.
- DeepInfra publishes no DeepSeek-V3.2 price in LiteLLM's map; the nearest sibling (V3.1) is used and labelled `price_source="nearest-sibling"`.
- The six stub adapters remain stubs.

JiRaska:
## Summary

Fleet LLM spend, throughput and reliability are now real Prometheus series, alerts and a dashboard. Previously nothing reported them, and the one alert that claimed to (`FinOpsAgentTokenBudgetExceeded`) was written against `langfuse_*` series from a Langfuse that has never been deployed — so it could not fire, and said so in its own description, which read as *pending* rather than as *inert*.

## Type of change

## Linked issues

Closes #3042
Refs ADR-0112, ADR-0174

## Why client-side and not at the gateway

LiteLLM is the natural choke point and does ship a Prometheus exporter — but it is an **Enterprise** feature (docs.litellm.ai/docs/proxy/prometheus, "Enterprise LiteLLM Only"). Checked against upstream docs before choosing an approach rather than assumed. So tokens are counted in the callers.

## What is actually instrumented

ADR-0174 describes one shared LLM seam, but only two services use it. A survey of every `/chat/completions` reference found eleven sites, of which **six are stubs that make no HTTP call at all** (`LlmDiagnosisAdapter` in finops-agent, governance-auditor, docs-truth-agent, flaky-test-hunter, authz-policy-auditor, release-steward — all still on the ADR-0112 P4 TODO). Three code paths make real calls, and all three are instrumented:

- `openbank-libs-runtime` `OpenAiCompatibleLlmGatewayClient` (devops-agent, control-liveness-sentinel)
- `openbank-agent-service` `OpenAiCompatibleModelProvider`
- `openbank-copilot-service` `OpenAiCompatibleModelProvider`

## Changes

- **New pure-domain port** `LlmCallMetricsPort` (openbank-libs-domain) with the closed outcome vocabulary `success | http_error | exception | not_configured`, implemented by the **existing** `DomainMetrics` CDI bean — not a new producer, which would drag Micrometer into every consumer's Arc type closure.
- The shared client takes it as a defaulted constructor arg (`LlmCallMetricsPort.NONE`) so it stays plain-constructible and an un-wired caller is silent rather than broken. The two CDI providers use field injection, because detekt's `LongParameterList` fires *at* `constructorThreshold`, not above it.
- **`usage` is now parsed.** The client discarded the response's `usage` block entirely (`@JsonIgnoreProperties` swallowed it) — which is why token consumption existed only on the provider's invoice. Nullable, so a provider that omits it cannot turn a good completion into a null.
- **`not_configured` is recorded, not silent.** An agent whose key was never seeded returns `null` forever and logs one warning at startup; nothing else anywhere says so.
- **Price book as recording rules**, not a constant in Kotlin: a rate-card change is then a gitops edit that reloads in place. `groq/llama-3.3-70b-versatile` is priced from BerriAI/litellm's own `model_prices_and_context_window.json` ($0.59/M in, $0.79/M out) — fetched, not recalled. DeepInfra publishes no DeepSeek-V3.2 entry there, so its nearest published sibling (V3.1) is used and every rule, alert and panel carries `price_source="nearest-sibling"` — the estimate is labelled in the data, not in a comment.
- **`openbank:llm_tokens_unpriced_24h`** is the check on the check: a model added to the gateway and forgotten in the price book makes every spend figure quietly low, and a plain sum cannot notice.
- Four alerts replace the dead one: `AiFleetDailySpendHigh`, `AiSpendUnpriced`, `AiCallErrorRateHigh` (excludes `not_configured`, a different failure with a different fix) and `AiCallsNotConfigured`. None depends on anything undeployed.
- `dashboard-openbank-ai.yaml`, 12 panels. Its header panel states in the board itself that cost is derived rather than billed, and that the six stub agents appear nowhere — so absence on this board is not evidence an agent is healthy.

## Series added

```
openbank_llm_requests_total{model,outcome}
openbank_llm_tokens_total{model,kind}            # prompt/completion split — priced differently
openbank_llm_call_duration_seconds{model,outcome}
```

Tags are the closed sets `model` and `outcome`/`kind` only — no prompt, user id, or status code.

## Falsification

- **`promtool check rules` initially FAILED.** `openbank:llm_tokens_unpriced_24h` was written `unless on (model) group_left ()`, and Prometheus rejects a grouping modifier on a set operator. Without running it this file would have been a rule group Prometheus silently refuses to load. Fixed; 11 rules SUCCESS.
- The validator was then fed a deliberately broken expression and reported it, so that SUCCESS is a measurement rather than a hope.
- All 13 dashboard PromQL targets extracted and run through promtool: SUCCESS.
- `./gradlew compileKotlin` on all five touched modules: **BUILD SUCCESSFUL**. (The shared `~/.gradle` on this machine has a corrupt instrumented-jar cache that fails `:build-logic:compilePluginsBlocks` on an *unmodified* tree too — verified against a clean worktree — so this ran against an isolated `GRADLE_USER_HOME`.)
- yamllint with `ci.yml`'s own inline config on all three manifests: clean.

## Definition of Done

- [x] Hexagonal layering preserved — the new port is pure domain, the Micrometer implementation stays in `openbank-libs-runtime`.
- [ ] Unit tests added/updated — none in this PR; the change is instrumentation on paths whose existing tests cover behaviour, and the metric contract is asserted by promtool over the rules and dashboard that consume it. Happy to add `DomainMetrics.recordCall` assertions if reviewers prefer.
- [ ] Integration tests — n/a.
- [ ] Contract tests — n/a, no public API change.
- [x] Compiles clean on every touched module.
- [x] No new lint warnings.
- [ ] OpenAPI / Flyway / Avro — n/a.
- [x] Docs updated — the dashboard documents its own limits in-board; the rules files carry the reasoning.

## Security checklist

- [x] No secrets, tokens, passwords, or PII in code, config, tests, logs. Prompts and completions are never recorded — only counts, and the API key handling is untouched.
- [x] No auth / crypto / payment code changed.
- [x] No PII path changed. Metric tags are a closed low-cardinality set by construction.
- [x] No cardholder data path changed.

## Compliance impact

- [x] DORA — ICT cost and third-party (LLM provider) monitoring.
- [x] EU AI Act adjacent — per-model call volume and failure rate become evidence rather than assertion.

## Deployment note

`rules.yaml` and the `.rego` policies are untouched, so no OPA bundle restamp. The manifests are synced by the `observability-components` Application (`directory.recurse: true`), so no kustomization registration is needed.

### DIFF
--- a/openbank-agent-service/src/main/kotlin/com/openbank/agent/infrastructure/model/OpenAiCompatibleModelProvider.kt
+++ b/openbank-agent-service/src/main/kotlin/com/openbank/agent/infrastructure/model/OpenAiCompatibleModelProvider.kt
@@ -17,6 +17,8 @@ import com.openbank.agent.domain.model.ModelResponse
 import com.openbank.agent.domain.model.ModelUsage
 import com.openbank.agent.domain.model.StopReason
 import com.openbank.agent.domain.model.ToolInvocation
+import com.openbank.libs.llm.LlmCallMetricsPort
+import com.openbank.libs.observability.DomainMetrics
 import jakarta.enterprise.context.ApplicationScoped
 import jakarta.inject.Inject
 import org.eclipse.microprofile.config.ConfigProvider
@@ -47,6 +49,12 @@ class OpenAiCompatibleModelProvider : ModelProvider {
     @Inject
     lateinit var objectMapper: ObjectMapper
 
+    // Field injection, not a constructor arg: detekt's LongParameterList fires AT
+    // constructorThreshold, and this class is instantiated by Arc, so there is no
+    // constructor to widen anyway. Same shape as LoanStageEventConsumer / VopRateLimitFilter.
+    @Inject
+    lateinit var metrics: DomainMetrics
+
     // Resolved lazily (NOT @ConfigProperty-injected): an un-seeded/empty key would otherwise fail
     // config load at boot (SmallRye SRCFG00040 on an empty String binding) and CrashLoop the pod —
     // and break any @QuarkusTest / fresh env where GROQ_API_KEY isn't set. Empty here just degrades
@@ -66,6 +74,9 @@ class OpenAiCompatibleModelProvider : ModelProvider {
     override suspend fun complete(model: ModelDescriptor, request: ModelRequest): ModelResponse {
         val base = (model.endpoint ?: error("model '${model.id}' has no endpoint (base URL) configured"))
             .trimEnd('/')
+        if (apiKey.isBlank()) {
+            metrics.recordCall(model.id, LlmCallMetricsPort.OUTCOME_NOT_CONFIGURED, 0, 0, 0)
+        }
         require(apiKey.isNotBlank()) {
             "agent.model.openai.api-key is empty — set the backend API key (e.g. GROQ_API_KEY) to use model '${model.id}'"
         }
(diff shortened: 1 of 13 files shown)


JiRaska:
**#3065 should merge after this one.**

It adds the preventive half of the same control (LiteLLM virtual keys + per-model ceilings backed by a Postgres) and its comments reference `AiFleetDailySpendHigh` from this PR. No file overlap between them, so this is a readability ordering, not a conflict one.

Issue #2161, opened 2026-07-25
agent-review Claude fallback is dead: OAuth token rejected at the gateway (owner-only credential)


JiRaska:
## Summary

Split out of #2088 (item 3). #2088's 413 defect is fixed by excluding generated artifacts from the reviewer payload and budgeting it to GitHub Models' 8000-token cap. This issue tracks the **other half**: the Claude fallback in `.github/workflows/agent-review.yml` is itself dead, and it needs a credential only the repo owner can mint — so it cannot be fixed by an agent.

## Symptom

When the GitHub Models step fails, `continue-on-error: true` routes to the `Claude review (fallback)` step (`anthropics/claude-code-action`). On PR #2082 it returned, ~2s after `Claude Code initialized`:

```
"type": "result", "subtype": "success", "is_error": true,
"duration_ms": 2115, "num_turns": 1, "total_cost_usd": 0
```

`is_error: true` at `num_turns: 1` with `total_cost_usd: 0` = the **first** model call was rejected at the gateway, before the agent ever ran `gh pr diff`. So this is not a diff-size problem — it is an auth/quota/model-availability rejection of `CLAUDE_CODE_OAUTH_TOKEN`. The secret is present in the run. Reproduced on 3 reruns over ~8h, so not a blip.

## Why it stayed invisible

The fallback only runs when the primary fails, and the primary succeeded on every normal-sized diff — so the standby path had effectively never been exercised. Same "silent standby that is actually dead" shape as the unreachable `merge_queue` trigger (#1465/#1504) and the red-addressed-to-nobody `dependency-submission` run.

With #2088's fix landed the primary now succeeds on exactly the PRs that used to 413, which makes the fallback *less* likely to be exercised — i.e. this stays hidden unless it is fixed deliberately.

## Owner-only action

1. Re-mint the token locally: `claude setup-token` (requires the owner's Claude Pro/Max session — no agent and no CI job can produce it), then update the repo secret `CLAUDE_CODE_OAUTH_TOKEN`.
2. Confirm the refreshed token actually works by forcing the fallback once (e.g. temporarily point the GitHub Models step at a bogus model on a throwaway PR, or dispatch the workflow with the primary disabled) — a green run that never entered the fallback proves nothing.

## Follow-up regardless of the credential

[...]

PR #3044, opened 2026-08-01
feat(libs): fleet-wide guards turning a missing or null JSON body into 400, not 500 (#3038)

### LINKED ISSUE #3038: Money-path: a null or absent JSON body returns 500 on 28 endpoints across 8 services
A missing or `null` JSON request body returns **500 Internal Server Error** instead of a 4xx across
**28 endpoints on 8 money-path services**. One defect class, one fix pattern.

Found by the first full run of `api-fuzz-authenticated.yml` across the money-path matrix
([30687065400](https://github.com/JiRaska/open-bank-oss/actions/runs/30687065400)), on the branch of
#3024 which repairs that harness. Until that run, **this lane had never started**, so no endpoint
here had ever been sent a malformed authenticated request.

## What the failing requests were

| Shape | Count |
|---|---:|
| `-d null` (JSON `null` body) | 16 |
| no body at all | 4 |

Every failure is the same class. Sample:

```
curl -X PATCH -H 'Authorization: …' -H 'Content-Type: application/json' \
  -d null http://localhost:8119/api/v1/fx/approvals/9019b4c2-…

curl -X POST  -H 'Authorization: …' -H 'Content-Type: application/json' \
  http://localhost:8119/api/v1/fx/convert
```

## The findings

| Service | Cases | 500s | Endpoints |
|---|---:|---:|---|
| `openbank-interest-service` | 812 | **8** | `DELETE /api/v1/interest/rates/{id}`<br>`GET /api/v1/interest/accounts/{accountId}/effective-rate`<br>`GET /api/v1/interest/accruals/{accountId}`<br>`POST /api/v1/interest/accrue`<br>`POST /api/v1/interest/accrue/all`<br>`POST /api/v1/interest/capitalize/{accountId}`<br>`POST /api/v1/interest/rates`<br>`POST /api/v1/interest/withholding/remittances` |
| `openbank-balance-service` | 722 | **6** | `PATCH /api/v1/balances/approvals/{id}` ¹<br>`POST /api/v1/balances/reconciliation`<br>`POST /api/v1/balances/{accountId}/credit`<br>`POST /api/v1/balances/{accountId}/debit`<br>`POST /api/v1/balances/{accountId}/holds`<br>`POST /api/v1/balances/{accountId}/initialize` |
| `openbank-swift-service` | 1057 | **4** | `PATCH /api/v1/swift/approvals/{id}` ¹<br>`POST /api/v1/swift`<br>`POST /api/v1/swift/{id}/ack`<br>`POST /api/v1/swift/{id}/reject` |
| `openbank-fx-service` | 745 | **3** | `PATCH /api/v1/fx/approvals/{id}` ¹<br>`POST /api/v1/fx/cnb/ingest`<br>`POST /api/v1/fx/convert` |
| `openbank-sepa-instant` | 615 | **3** | `PATCH /api/v1/sepa-instant/approvals/{id}` ¹<br>`POST /api/v1/sepa-instant`<br>`POST /api/v1/sepa-instant/{paymentId}/recall` |
| `openbank-billing-service` | 192 | **2** | `PATCH /api/v1/fees/approvals/{id}` ¹<br>`POST /api/v1/fees/post` |
| `openbank-fraud-service` | 113 | **1** | `POST /api/v1/fraud/score` |
| `openbank-consent-service` | 821 | **1** | `PATCH /api/v1/consents/approvals/{id}` ¹ |

**Total: 28 unique 500s across 8 money-path services.**

¹ already fixed by #3032 (all sixteen `ApprovalResource.decide` copies). **The rest are not covered
by anything** — they are ordinary POST/PATCH/DELETE handlers with the same non-nullable request
parameter.

## Root cause and fix

The parameter is declared non-nullable in Kotlin (`request: XRequest`). Kotlin null-safety is a
**compile-time** property; JAX-RS/Jackson hands a `null` straight through at runtime, the first field
access throws NPE, and the generic mapper turns it into a 500.

The fix is the pattern this repo already uses in `ConsentResource`, `BillingResource` and elsewhere:

```kotlin
suspend fun handler(…, request: XRequest?): Response {
    requireNotNull(request) { "request body is required" }
```

`libs-runtime`'s `CommonExceptionMappers` maps `IllegalArgumentException` → 400. See #3032 for the
worked example across sixteen files.

## Why this matters beyond the status code

The property schemathesis asserts — *no input, however malformed, produces a 5xx* — is the
public-surface hardening a pentest probes first (ADR-0080), and it is the `pentest` dimension keeping
money-path services NO-GO in #2365. Several of these are money-movement endpoints:
`POST /api/v1/balances/{accountId}/credit`, `/debit`, `POST /api/v1/swift`,
`POST /api/v1/fx/convert`, `POST /api/v1/interest/accrue/all`.

A 500 here is not itself an exploit, but it marks an unhandled path in code that moves money, and it
is exactly the signal an adversarial tester follows first.

## Scope note

Only **8 of 16** services in the matrix got this far — the other 7 failed to boot in the harness and
are tracked separately. `openbank-clearing-service` fuzzed clean over 621 cases, so a green result is
achievable and this is not a harness artefact.

Refs #3024, #3032, #2365, #852

JiRaska:
> **Read the "Critical assessment" section before approving.** I do not think this should merge
> first, or alone — see "Suggested order". Opening it because the work is done and reviewable, not
> because I think it is ready to land.

## Summary

Fleet-wide guards that turn a **missing or `null` JSON request body into 400 instead of 500**, by
reading the handler's own Kotlin nullability rather than applying a blanket rule.

## Type of change

- [x] feat (new functionality) — shared runtime behaviour

## Linked issues

Refs #3038, #3032, #2365

## The defect

The handler declares a non-nullable Kotlin body parameter (`request: FooRequest`). Kotlin
null-safety is a **compile-time** property; JAX-RS/Jackson hands a `null` straight through at
runtime, the first field access throws NPE, and the generic mapper renders 500.

The first full run of the authenticated fuzz lane found **28 of these across 8 money-path services**
(#3038); twenty were exactly `-d null` or a body-less POST, on endpoints including
`POST /api/v1/balances/{accountId}/credit`, `POST /api/v1/swift`, `POST /api/v1/fx/convert`.

## The design decision that makes this safe

**A blanket "null body → 400" rule would be wrong, and provably so.** `BillingResource.reverse`
declares `request: ReverseFeeRequest?` and *deliberately* handles null, returning its own 400 naming
both missing fields. A global rule would pre-empt that — and would be deciding a per-handler question
globally.

So the guards read **Kotlin's own nullability**: the compiler emits
`@org.jetbrains.annotations.NotNull` on a non-nullable JVM parameter. A body is required **iff the
handler said it was** — the same source of truth the bug comes from.

## Two classes, because the shapes are knowable at different moments

| | Where | Why there |
|---|---|---|
| **Absent body** | `ContainerRequestFilter` | Decidable from headers (`hasEntity()`) — cheap, non-blocking. The reader is never invoked, so an interceptor could not see it at all. |
| **Literal `null`** | `ReaderInterceptor` | Only knowable after deserialization. Detecting it in a filter means reading and re-buffering the entity stream — a **blocking read on the I/O thread** on RESTEasy Reactive. |

## The mistake this almost shipped

A `suspend fun` carries a `kotlin.coroutines.Continuation` at the JVM level, **unannotated**. The
first draft's entity detector treated it as the request body — and **the fleet has 346 suspend
handlers**, so every body-less POST would have started answering 400.

Caught by enumerating the fleet, not by re-reading the code. The exclusion is explicit and is the
reason the test file exists.

The detector is conservative everywhere else too: anything it cannot positively identify as an
entity parameter degrades to today's behaviour. A false negative leaves a 500 that already exists;
a false positive rejects valid money-path traffic, which is strictly worse.

## Critical assessment — the case against merging this as-is

**Maximum blast radius.** Jandex auto-registers these into *every* service; they run on *every*
POST/PUT/PATCH in the fleet. I have not executed them once — the build host's shared Gradle cache is
corrupting artifacts, so there is no local run behind this.

**The `@NotNull` assumption is unverified in bytecode.** I reasoned it from how Kotlin emits JVM
annotations; I could not decompile a real handler to confirm. If Kotlin omits it for some shape, the
guard silently does nothing. That is a safe failure — but it means this may simply be inert.

**`hasEntity()` on a chunked request with no `Content-Length`** may not answer what I expect.

**It is a net, not a cure.** The handlers still declare non-nullable parameters. This stops the 500
reaching a client and covers endpoints nobody has fuzzed yet, but the per-handler fix remains the
actual correction. **It does not close #3038.**

## Suggested order

1. **#3032 first** — the sixteen approval handlers. Local, evidenced, small.
2. **Then the remaining per-handler fixes** for #3038's other endpoints.
3. **These guards last**, as a net — at the point where they should catch *nothing*, because every
   known case is already fixed. That is the only way to learn whether they produce false positives:
   turning them on must change nothing.

Before this reaches `main` it also wants one authenticated-fuzz run with the guards enabled. The
pass condition is specific: the 28 known findings come back as **400 instead of 500**, and **no new
rejection appears anywhere**. I do not have that measurement.

## Definition of Done

- [x] Design driven by two counterexamples found in the fleet, not by the happy path.
- [x] Unit tests covering the dangerous negatives (suspend-without-body, nullable parameter,
      annotation-only handlers) — the false-positive cases matter more than the positive one.
- [ ] **Never executed.** No local run, no CI run with the guards active on real traffic.
- [ ] Does not close #3038 — deliberately.

## Security checklist

- [x] No secrets, no new dependency.
- [x] Changes request handling on every money-path service → `security-review-required`.
- [x] Fails open by design: unrecognised shapes behave exactly as today.

## Compliance impact

- [x] DORA — money-path endpoints answering 500 to trivially malformed input.
- [ ] PCI DSS / GDPR / PSD2 / AML / CNB — not applicable.

### DIFF
--- a/openbank-libs-runtime/src/main/kotlin/com/openbank/libs/api/error/RequiredBodyGuards.kt
+++ b/openbank-libs-runtime/src/main/kotlin/com/openbank/libs/api/error/RequiredBodyGuards.kt
@@ -1,0 +1,136 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) OpenBank contributors. Licensed under the Apache License, Version 2.0.
+// See LICENSE in the repository root or https://www.apache.org/licenses/LICENSE-2.0 for details.
+
+package com.openbank.libs.api.error
+
+import jakarta.ws.rs.BadRequestException
+import jakarta.ws.rs.container.ContainerRequestContext
+import jakarta.ws.rs.container.ContainerRequestFilter
+import jakarta.ws.rs.container.ResourceInfo
+import jakarta.ws.rs.core.Context
+import jakarta.ws.rs.ext.Provider
+import jakarta.ws.rs.ext.ReaderInterceptor
+import jakarta.ws.rs.ext.ReaderInterceptorContext
+import java.lang.reflect.Method
+import java.lang.reflect.Parameter
+
+/**
+ * Turns a **missing or `null` JSON request body into 400 instead of 500**, fleet-wide.
+ *
+ * ## The defect
+ *
+ * A resource method declares a non-nullable Kotlin body parameter (`request: FooRequest`). Kotlin
+ * null-safety is a *compile-time* property — JAX-RS/Jackson hands a `null` straight through at
+ * runtime, the first field access throws NPE, and the generic mapper renders it as 500.
+ *
+ * The first full run of the authenticated fuzz lane found **28 of these across 8 money-path
+ * services** (#3038) — `POST /api/v1/balances/{accountId}/credit`, `POST /api/v1/swift`,
+ * `POST /api/v1/fx/convert` among them. Twenty were exactly `-d null` or a body-less POST.
+ *
+ * ## The decision that makes this safe: read Kotlin's own nullability
+ *
+ * A blanket "null body → 400" rule would be wrong, and provably so. `BillingResource.reverse`
+ * declares `request: ReverseFeeRequest?` and *deliberately* handles null, returning its own 400
+ * naming both missing fields. A global rule would pre-empt that with a generic message, and more
+ * importantly would be deciding a per-handler question globally.
+ *
+ * So these guards decide per parameter, from the same source of truth the bug comes from: the
+ * Kotlin compiler emits `@org.jetbrains.annotations.NotNull` on a non-nullable JVM parameter and
+ * `@Nullable` on a nullable one. A body is required **iff the handler said it was**.
+ *
+ * ## Why two classes
+ *
+ * The two shapes are knowable at different moments, and merging them would be a correctness bug:
+ *
+ * - **Absent body** is decidable from headers ([ContainerRequestContext.hasEntity]) — cheap and
+ *   non-blocking, so it belongs in a request filter. The reader is never invoked in this case, so
+ *   an interceptor could not see it at all.
+ * - **A literal `null` body** is only knowable after deserialization. Detecting it in a request
+ *   filter would mean reading and re-buffering the entity stream, which on RESTEasy **Reactive**
+ *   is a blocking read on the I/O thread. A [ReaderInterceptor] runs where blocking is already
+ *   permitted and sees the decoded value directly — `proceed()` returning `null` IS the defect.
+ *
+ * ## This is a net, not a cure
+ *
+ * The handlers still declare non-nullable parameters. This stops the 500 reaching a client, and
+ * covers endpoints nobody has fuzzed yet, but the per-handler fix (nullable parameter +
+ * `requireNotNull`, as #3032 did for the sixteen approval resources) remains the actual
+ * correction. Defence in depth, not a reason to close #3038.
+ */
+internal object RequiredBody {
+
+    val BODY_METHODS = setOf("POST", "PUT", "PATCH")
+
+    private const val NOT_NULL = "org.jetbrains.annotations.NotNull"
+
+    /**
+     * The entity parameter of [method], or null when it declares none.
+     *
+     * Three exclusions, each load-bearing:
+     * - **JAX-RS-annotated** parameters (`@PathParam`, `@QueryParam`, `@Context`, `@BeanParam`, …)
+     *   are never the entity.
+     * - **`kotlin.coroutines.Continuation`** — every `suspend fun` carries one at the JVM level, and
+     *   it is unannotated. Without this exclusion the fleet's 346 suspend handlers would each look
+     *   like they expect a body, and every body-less POST would start answering 400. This is the
+     *   single most dangerous mistake available here.
+     * - **`jakarta.ws.rs.core.*`** container types (`UriInfo`, `HttpHeaders`, …).
+     */
+    fun entityParameter(method: Method): Parameter? = method.parameters.firstOrNull { p ->
+        val jaxRsAnnotated = p.annotations.any {
+            it.annotationClass.qualifiedName?.startsWith("jakarta.ws.rs.") == true
+        }
+        !jaxRsAnnotated &&
+            p.type.name != "kotlin.coroutines.Continuation" &&
+            !p.type.name.startsWith("jakarta.ws.rs.core.")
+    }
+
+    /** True only when the handler declared the body **non-nullable** — i.e. null is a defect here. */
+    fun isRequired(annotations: Array<out Annotation>): Boolean =
+        annotations.any { it.annotationClass.qualifiedName == NOT_NULL }
+}
+
+/**
+ * Rejects a POST/PUT/PATCH carrying **no entity at all** when the matched handler declares a
+ * non-nullable body parameter.
+ *
+ * Post-matching by design — it needs [ResourceInfo] to know what the handler expects, so it must
+ * not be `@PreMatching`. Reads no stream: [ContainerRequestContext.hasEntity] answers from headers,
+ * which is what keeps it safe on the reactive stack.
+ */
+@Provider
+class AbsentBodyRequestFilter : ContainerRequestFilter {
+
+    @Context
+    lateinit var resourceInfo: ResourceInfo
+
+    override fun filter(requestContext: ContainerRequestContext) {
+        if (requestContext.method !in RequiredBody.BODY_METHODS) return
+        if (requestContext.hasEntity()) return
+
+        val method = runCatching { resourceInfo.resourceMethod }.getOrNull() ?: return
+        val entity = RequiredBody.entityParameter(method) ?: return
+        if (!RequiredBody.isRequired(entity.annotations)) return
+
+        throw BadRequestException("request body is required")
+    }
+}
+
+/**
+ * Rejects a body that deserialises to `null` — the JSON literal `null`, which Jackson decodes
+ * happily and hands to a parameter Kotlin believes cannot hold it.
+ *
+ * A nullable parameter passes through untouched, so handlers that deliberately accept an absent
+ * body keep their own behaviour and their own error messages.
+ */
+@Provider
+class NullBodyReaderInterceptor : ReaderInterceptor {
+
+    override fun aroundReadFrom(context: ReaderInterceptorContext): Any? {
+        val value = context.proceed()
+        if (value == null && RequiredBody.isRequired(context.annotations)) {
+            throw BadRequestException("request body is required")
+        }
+        return value
+    }
+}
--- a/openbank-libs-runtime/src/main/kotlin/com/openbank/libs/api/error/RequiredBodyGuards.kt
+++ b/openbank-libs-runtime/src/main/kotlin/com/openbank/libs/api/error/RequiredBodyGuards.kt
@@ -72,7 +72,6 @@ internal object RequiredBody {
 
     private const val CONTINUATION = "kotlin.coroutines.Continuation"
 
-
     /**
      * The entity parameter of [method], or null when it declares none.
      *
--- a/openbank-libs-runtime/src/main/kotlin/com/openbank/libs/api/error/RequiredBodyGuards.kt
+++ b/openbank-libs-runtime/src/main/kotlin/com/openbank/libs/api/error/RequiredBodyGuards.kt
@@ -14,6 +14,8 @@ import jakarta.ws.rs.ext.ReaderInterceptor
 import jakarta.ws.rs.ext.ReaderInterceptorContext
 import java.lang.reflect.Method
 import java.lang.reflect.Parameter
+import kotlin.reflect.KParameter
+import kotlin.reflect.jvm.kotlinFunction
 
 /**
  * Turns a **missing or `null` JSON request body into 400 instead of 500**, fleet-wide.
@@ -36,8 +38,14 @@ import java.lang.reflect.Parameter
  * importantly would be deciding a per-handler question globally.
  *
  * So these guards decide per parameter, from the same source of truth the bug comes from: the
- * Kotlin compiler emits `@org.jetbrains.annotations.NotNull` on a non-nullable JVM parameter and
- * `@Nullable` on a nullable one. A body is required **iff the handler said it was**.
+ * handler's own Kotlin nullability. A body is required **iff the handler said it was**.
+ *
+ * Read via `kotlinFunction`, NOT via `@org.jetbrains.annotations.NotNull`. The compiler does emit
+ * that annotation, but it is declared `@Retention(RetentionPolicy.CLASS)` — it exists in the class
+ * file and is invisible to `Parameter.getAnnotations()` at runtime, so an annotation-based check
+ * can never return true. The first version of this guard was written that way and its own unit
+ * test caught it. `kotlinFunction` reads the `@Metadata` annotation, which is RUNTIME-retained;
+ * `AuthorizeInterceptor` and `FeatureFlagInterceptor` already depend on it fleet-wide.
  *
  * ## Why two classes
  *
@@ -62,7 +70,8 @@ internal object RequiredBody {
 
     val BODY_METHODS = setOf("POST", "PUT", "PATCH")
 
-    private const val NOT_NULL = "org.jetbrains.annotations.NotNull"
+    private const val CONTINUATION = "kotlin.coroutines.Continuation"
+
 
     /**
      * The entity parameter of [method], or null when it declares none.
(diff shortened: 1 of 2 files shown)


JiRaska:
## The `@NotNull` assumption is falsified — the guards are inert as written

The "Critical assessment" lists this as an open risk:

> **The `@NotNull` assumption is unverified in bytecode.** I reasoned it from how Kotlin emits JVM annotations; I could not decompile a real handler to confirm. If Kotlin omits it for some shape, the guard silently does nothing.

I decompiled it. The answer is worse than "some shape" and better than "unknown": Kotlin emits the annotation for **every** shape, into a bytecode attribute that **core reflection never exposes**. `RequiredBody.isRequired()` cannot return `true` for any handler.

### 1. The real handler does carry `@NotNull` — in the Invisible attribute

`javap -v` on the compiled `ApprovalResource.decide` (balance-service, from a local build tree):

```
public java.lang.Object decide(java.lang.String, DecideApprovalRequest, kotlin.coroutines.Continuation<...>);
    RuntimeVisibleParameterAnnotations:      <- jakarta.ws.rs.PathParam
    RuntimeInvisibleParameterAnnotations:
      parameter 0: org.jetbrains.annotations.NotNull
      parameter 1: org.jetbrains.annotations.NotNull
      parameter 2: org.jetbrains.annotations.NotNull
```

The JAX-RS annotations land in **Visible**; Kotlin's `@NotNull` lands in **Invisible**. That split is the whole story.

### 2. Because the annotation itself is `CLASS`-retained

`javap -v` on `NotNull.class` from `annotations-26.1.0.jar`:

```
java.lang.annotation.Retention(
  value=Ljava/lang/annotation/RetentionPolicy;.CLASS
)
```

### 3. And a `CLASS`-retained parameter annotation is invisible to `Parameter.getAnnotations()`

Measured rather than argued — two annotations of my own, plain `javac`, one of each retention on the same method:

```
param 0 declared=@ClassRet (CLASS)     reflection sees: []
param 1 declared=@RuntimeRet (RUNTIME) reflection sees: [RuntimeRet]
```

```
RuntimeVisibleParameterAnnotations:   RuntimeRet
RuntimeInvisibleParameterAnnotations: ClassRet
```

So in

```kotlin
fun isRequired(annotations: Array<out Annotation>): Boolean =
    annotations.any { it.annotationClass.qualifiedName == NOT_NULL }
```

the array never contains it. Both call sites go through reflective annotations — the filter's `entity.annotations` (a `java.lang.reflect.Parameter`) and the interceptor's `context.annotations` (the resource method's parameter annotations, obtained the same way) — so both return early, always. The net catches nothing.

### 4. The PR's own test already says this

This is not a new finding so much as a confirmation of what CI is reporting. `build (openbank-libs-runtime)` is red on:

```
RequiredBodyTest > Nullability > a non-nullable body parameter is required() FAILED
    org.opentest4j.AssertionFailedError at RequiredBodyTest.kt:95
132 tests completed, 1 failed
```

Line 95 is `assertThat(RequiredBody.isRequired(p.annotations)).isTrue()`, asserted against a real Kotlin-compiled handler through real reflection. **That test is the right test** — it states the load-bearing assumption as an assertion instead of trusting it, and it is doing exactly its job by failing. Worth saying plainly, because the tempting next move is to relax it.

### What this does and does not change

It does not touch the **design**, which I think is right: deriving "is a body required" from the handler's own nullability rather than a blanket rule, and the `Continuation` exclusion (346 suspend handlers) are both correct calls. What is wrong is only the *source* the design reads that nullability from.

Two ways out, both real work:

- **Jandex.** Quarkus already builds the index at augmentation, and it does read `RuntimeInvisibleParameterAnnotations` — `MethodInfo.parameters()` carries them. The lookup moves to build time, which also removes per-request reflection.
- **`kotlin.reflect`.** `KFunction.parameters[i].type.isMarkedNullable` reads Kotlin metadata rather than JVM annotations, so it sees the truth. But it is per-request reflection on every POST/PUT/PATCH in the fleet and `kotlin-reflect` is a heavier dependency — this one needs a measurement before it is a proposal.

Tuning the current approach will not work: reading runtime annotations is a dead end by construction, not an under-tuned heuristic.

### One stale blocker to strike from the Definition of Done

> I have not executed them once — the build host's shared Gradle cache is corrupting artifacts, so there is no local run behind this.

That obstacle is gone. It was never corruption: five `openbank-app` self-hosted jobs ran `gradle/actions/setup-gradle` with no `GRADLE_USER_HOME`, so its default `cache-cleanup: on-success` pruned the developer's `~/.gradle` while local builds were reading it. Fixed in JiRaska/openbank-app#322 and #309, with `check-gradle-user-home-isolation.py` guarding both repos (#3004 here). Local builds on this machine work again, so "never executed" is now a choice rather than a constraint.

Same log incidentally shows this branch also hit `Dependency verification failed for configuration ':openbank-libs-domain:compileClasspath' — 4 artifacts` in the warm-up step, which is the cold-cache additions-only case, unrelated to the above.

### Suggested order — unchanged, and now better supported

#3032 and #3050 are the real corrections, both `CLEAN`, and neither depends on this. Landing them first is right independently of what happens here; this PR should stay a draft until the nullability source is replaced and it can be shown to catch something.

JiRaska:
## Jandex works — prototyped it, and the runtime half is the real decision

Following up on the inertness finding above. I ran the Jandex path against real compiled classes rather than reasoning about it, and it does what the design needs. But "rewrite it on Jandex" splits into two questions with different answers, and the second one is where the work is.

### 1. Does Jandex see what reflection cannot? Yes

Same `ApprovalResource.decide` whose `Parameter.getAnnotations()` returns `[]`:

```
method ApprovalResource.decide
  param 0 type=java.lang.String                        annotations=[@PathParam("id"), @NotNull]
  param 1 type=DecideApprovalRequest                   annotations=[@NotNull]
  param 2 type=Continuation<? super Response>          annotations=[@NotNull]
```

`MethodParameterInfo.annotations()` reads `RuntimeInvisibleParameterAnnotations`. The nullability source the design wants is available — just not through core reflection.

### 2. Can `libs-runtime` reach an index of the *service's* classes at runtime? Not today

`alias(libs.plugins.jandex)` is applied in `openbank-libs-domain` and `openbank-libs-runtime` only — **no service applies it**, so `META-INF/jandex.idx` exists for libs classes while the handlers live in the services. Quarkus does index everything at augmentation, but `CombinedIndexBuildItem` is reachable only from an extension's deployment module, and `libs-runtime` is a plain library.

So there are three shapes, and they are not equally priced:

| | what it takes | runtime cost |
|---|---|---|
| **A** — index the services too | jandex plugin in ~35 build files; read every `META-INF/jandex.idx` off the classpath at startup into a `CompositeIndex` | index parse at startup, index held in memory |
| **B** — build-time set | a Gradle task computes the required-body signatures and writes a small resource; runtime reads a `Set<String>` | ≈ none, no Jandex at runtime |
| **C** — Quarkus extension | new deployment module, build step + recorder | none, but the largest change and it changes what `libs-runtime` *is* |

### Prototype of B, run against real classes

~60 lines using Jandex over compiled class directories. Across the nine services that happened to be built on this machine — **1081 classes, 150 JAX-RS handlers, 40 required-body, 1 deliberately nullable, 109 with no entity parameter.**

What it gets right, checked in both directions rather than only the happy one:

| case | result |
|---|---|
| `BillingResource#reverse` — the `ReverseFeeRequest?` counterexample this PR is designed around | **OPTIONAL**, not required ✅ |
| `Continuation` on a `suspend fun` | not mistaken for the body ✅ |
| handlers fixed by #3032 / #3050, scanned at their pre-fix commit | all present in the required set ✅ |

That last row is worth dwelling on: the set the scanner produces at a pre-fix commit **is** the list of bugs those two PRs went on to fix. It reproduces a known answer.

### The prototype found a false positive the reflection version structurally cannot

```
REQUIRED body (2):
  com.openbank.billing.infrastructure.client.LedgerRestClient#postJournal   <- outbound REST client
  com.openbank.billing.infrastructure.rest.ApprovalResource#decide
```

An outbound MicroProfile REST *client* carries the same JAX-RS annotations as an inbound resource. A client is an `interface`; a resource is a concrete class — `if (Modifier.isInterface(ci.flags())) continue` drops it, and billing's handler count goes 13 → 8, required 2 → 1.

The interesting part is not the bug, it is *when* it is visible. The current reflection implementation has the same blind spot and would never show it, because `isRequired()` never returns true. The day someone fixes the nullability source without also fixing this, the guard starts rejecting valid outbound calls. Variant B surfaces it at build time, because the set is a readable artifact you can diff — which is also the answer to "how do we learn whether it produces false positives" from the Suggested order section.

### What I did not do

- **The runtime half of B is unbuilt.** The Gradle task that writes the resource, and the guard that reads a `Set<String>` instead of annotations — that is the remaining work. The scanner is the cheap part.
- **No `UNDECIDED` category.** A parameter carrying neither `@NotNull` nor `@Nullable` (a Java handler, or Kotlin with the annotations suppressed) currently falls into the "no entity" bucket. Production code wants that as an explicit third state that degrades to today's behaviour, matching this PR's fail-open posture.
- Nothing is executed against real traffic. That still needs the authenticated-fuzz run described in this PR's Definition of Done, with the pass condition unchanged: the 28 known findings return 400, and nothing new is rejected.

### One correction to my earlier comment

I wrote that `kotlin.reflect` would mean "per-request reflection on every POST/PUT/PATCH". That is too pessimistic — the answer is constant per `Method`, so a `ConcurrentHashMap<Method, Boolean>` makes it one reflection per handler for the life of the process. The real objection is different: `kotlin-reflect` is declared only in `openbank-libs-domain` and `openbank-libs-testing` today, so `libs-runtime` would drag it onto the runtime classpath of every service (~3 MB) — a worse trade than B, but not for the reason I first gave.

JiRaska:
## Ready — and the mechanism changed since the draft was opened

The guard originally keyed on `@org.jetbrains.annotations.NotNull`. The Kotlin compiler does emit
that on a non-nullable JVM parameter, but it is declared `@Retention(RetentionPolicy.CLASS)`, so it
is **absent from `Parameter.getAnnotations()` at runtime**. `isRequired` could never return true and
both guards were structurally inert — they would have shipped, run on every POST/PUT/PATCH in the
fleet, and never rejected anything.

Verified against the annotation jar rather than from memory:

```
javap -v org/jetbrains/annotations/NotNull.class   →   Retention: CLASS
```

Its own unit test is what caught it. The test was right; the design was wrong.

Now read through `method.kotlinFunction` — the `@Metadata` annotation, which IS RUNTIME-retained.
Not a new dependency: `AuthorizeInterceptor` and `FeatureFlagInterceptor` in this same module
already resolve handlers that way fleet-wide, so the mechanism is one the platform already runs in
production.

Two details that would otherwise be silent bugs:

- `KFunction.parameters` excludes the `Continuation` every `suspend fun` carries at the JVM level
  and includes the receivers, so JVM and Kotlin positions do not line up. Both sides are filtered to
  value parameters before pairing, and a suspend handler has its own test.
- `NullBodyReaderInterceptor` was reading `context.annotations`, which has the identical retention
  problem. It now resolves the matched method via `ResourceInfo`, like the filter does.

The original defect is kept as a regression test — it asserts the runtime parameter carries no
nullability annotation — so nobody reinstates the annotation-based check.

## CI

**85 checks pass, 0 fail.** That breadth is the point: this is a `openbank-libs-runtime` change, so
every service builds against it, and a filter that mis-identified a body parameter would break
handlers fleet-wide rather than in one module.

## Scope note

The per-handler fixes (#3032, #3050, #3057) are merged and remain the actual correction — this is
defence in depth for endpoints nobody has fuzzed yet, not a reason to close #3038.

JiRaska:
## Merging — the fleet-wide blast radius checks out empirically

Two `@Provider` classes in `openbank-libs-runtime` auto-register in **every** service on the classpath, so a `ContainerRequestFilter` here runs on every POST/PUT/PATCH in the bank. That is worth measuring rather than reading, so I did.

### Does the heuristic misfire anywhere in the fleet?

`RequiredBody.entityParameter` takes the *first* parameter that is not JAX-RS-annotated, not a `Continuation`, and not `jakarta.ws.rs.core.*`. If a handler's first such parameter were an injected type (`SecurityIdentity`, `RoutingContext`, `JsonWebToken`, …) rather than a DTO, the guard would treat a legitimately body-less POST as a missing required body and start returning 400.

Scanned every `@POST`/`@PUT`/`@PATCH` handler in `*/src/main/**`:

```
349  handlers total
 95  no unannotated parameter at all      -> guard silent by construction
  0  first unannotated parameter is an injected type   <- would misfire
```

Zero. The fleet consistently injects `SecurityIdentity` as a field (`@Inject lateinit var`), which is also what `CLAUDE.md` recommends for dodging detekt's `LongParameterList` — so the pattern that would break this does not exist here.

### Coverage

`feat(libs)` did rebuild the fleet on this PR — **60 service builds, all green**, not just the two changed files. I checked rather than assumed; I had a note claiming a libs change rebuilds nothing, and it does not apply here.

### Scope of the behaviour change

The guard fires only when all four hold: the method is POST/PUT/PATCH, the request carries no entity, an entity parameter exists, and its Kotlin type is **non-nullable**. A handler declaring `body: Foo?` is untouched. So the change converts 500 into 400 for genuinely required bodies and is a no-op everywhere else.

### Non-required red

`agent-review` is failing — that is #2161 (413 on generated bundles), not this PR. All five required contexts are green.

*Reviewed while draining the queue overnight. Additive only: +314/-0 across 2 files, no deletions.*

JiRaska:
## Closing the gap my own review left — the guard is live at runtime, verified

I merged this after checking it could not *misfire*. I did not check that it *fires*, and those are different claims. Given [the inertness finding above](https://github.com/JiRaska/open-bank-oss/pull/3044#issuecomment-5150042183) — `Parameter.getAnnotations()` never seeing `@NotNull` — that was the more important half, and I skipped it. Filling it in now.

### The merged version does not depend on the annotation

It reads Kotlin metadata instead, and says so in place:

```kotlin
val kFunction = runCatching { method.kotlinFunction }.getOrNull() ?: return false
...
return !kParam.type.isMarkedNullable
```

`kotlinFunction` reads `@Metadata`, which is `RUNTIME`-retained — so the mechanism that made the earlier version inert does not apply here.

### But that swaps one silent failure for another

`kotlinFunction` needs **kotlin-reflect on the runtime classpath**, and `openbank-libs-runtime/build.gradle.kts` does not declare it. The convention plugin applies `org.jetbrains.kotlin.jvm`, which brings `kotlin-stdlib`, not `kotlin-reflect`. Where it is absent, `runCatching { … }.getOrNull() ?: return false` swallows the failure and `isRequired` returns `false` for every handler — inert again, with nothing logged.

The unit tests cannot answer this: they prove `isRequired` works on the **test** classpath, which is not the one the filter runs on.

### Measured on the real runtime classpath

```
:openbank-fx-service:dependencies --configuration runtimeClasspath
    kotlin-reflect  4 occurrences   org.jetbrains.kotlin:kotlin-reflect:2.4.0

:openbank-ledger-service:dependencies --configuration runtimeClasspath
    kotlin-reflect  4 occurrences   org.jetbrains.kotlin:kotlin-reflect:2.4.0
```

Both resolve it transitively at 2.4.0. Control on the same output — `kotlin-stdlib` 19 and 18 occurrences — so the greps are reading real dependency trees, not empty files. (My first attempt returned 0 for *both*, which is what a broken probe looks like: `timeout` does not exist on macOS, exit 127.)

So the guard is live in these services. Two samples, not 34 — and it is transitive rather than declared, which means a dependency change elsewhere could remove it silently.

### Worth a follow-up

Declaring `kotlin-reflect` explicitly in `openbank-libs-runtime` would make this a property rather than a coincidence. It costs nothing (already resolved fleet-wide) and removes a silent-inertness mode that no test layer here can see.

Also: #3104's table still lists this PR as "draft — inert as written". That was true of the version it described, and is no longer true of what merged — worth correcting there so nobody plans around a gap that is closed.

*Follow-up to an overnight merge; the original review is [above](https://github.com/JiRaska/open-bank-oss/pull/3044#issuecomment-5150335120).*

PR #3064, opened 2026-08-01
fix(billing): validate fee query params instead of letting the schema reject them

JiRaska:
Fixes the last undiagnosed 500 from the first authenticated fuzz run (#3038):
`POST /api/v1/fees/post?...&currency=ISO-2022-CN`.

## What was wrong

`validateParams` checked only `isNullOrBlank`. Everything else about these three parameters
was enforced by the database and nowhere else:

```
V1__init_billing.sql:36    currency  CHAR(3) NOT NULL
V1__init_billing.sql:13/14 cycle_id / account_id  VARCHAR(64) NOT NULL
```

An 11-character currency is non-blank, so it passed, reached
`persistWithPostingIntent`, and Postgres rejected the INSERT. Nothing maps that fault, so
it came out as a 500 — a server error for what is squarely a client error.

## The part worth reading

The fault handling *below* the validation looks like it should have absorbed this, and it is
the reason it did not. `accounts.resolve` is `runCatching{}.getOrNull()`, so an unresolvable
account produces a `skipped = true` assessment — which **still carries the caller's raw
currency** — and `assessAndPost` persists skipped assessments by design. The graceful
degradation branch is therefore exactly the branch that reaches the `CHAR(3)` column.

And `BillingCycleService.runCycle` wraps each account in `runCatching`, so the scheduled
sweep already absorbed this: one bad currency logged and the batch continued. Only the
single-account REST entry point propagated. That is why it never showed up in operation —
the path that runs daily is the protected one, and the exposed path is the one nobody calls
by hand.

## The fix

Currency is parsed through `CurrencyCode` (openbank-libs-domain), the fleet's existing
validating value type — `require(code.length == 3)` plus an ISO 4217 lookup. Billing already
depends on it (`BillingJournalFactory`) but did not use it here, so the invariant was
declared in two places (the value type and the DDL) and enforced in neither of the ones a
request passes through. Ids get an explicit 64-character bound against the same DDL.

`CurrencyCode.of` also upper-cases. That is deliberate and is a behaviour change: `czk` and
`CZK` would otherwise be two distinct values in both the column and the
`(cycleId, accountId, currency)` idempotency key, so the same fee could be assessed twice
under two spellings of one currency.

## Tests

`BillingResourceValidationTest` — 8 cases. Each asserts the **pair**: 400 *and* the use case
was never invoked. A 400 alone would still pass if validation ran after the write. Includes
the exact fuzz input, a 3-letter non-ISO code, both over-long ids, an id of exactly 64
(off-by-one guard), lower-case normalisation, the pre-existing blank check, and the
`/fees/assess` twin that shares the code path.

## Spec

The 400 description said "Missing required query parameter", no longer the whole truth.
`info.version` 1.3.0 -> 1.3.1.

The limits are documented in **prose, not as `pattern`/`maxLength`**. oasdiff classifies a
newly narrowed request parameter as breaking, and ADR-0048 ties the spec's major to the
URL's `/api/v{N}` — so expressing them mechanically would demand `/api/v2` for a change that
does not deserve one. Flagging that explicitly because it is a real gap: the machine-checkable
form of this constraint is unavailable at this layer, and the deterministic guard is the unit
test above, not the spec.

## Verification status

Not compiled locally — the shared Gradle cache on this host is still corrupt
(`NoClassDefFoundError: org/jetbrains/kotlin/compilerRunner/CompilerEnvironment` on
`:build-logic:compilePluginsBlocks`, offline or not). CI is the verifier, as with the
preceding PRs in this series. YAML was validated with a parser; line length and brace
balance checked by hand.

Refs #3038

### DIFF
--- a/docs/threat-models/openbank-billing-service.md
+++ b/docs/threat-models/openbank-billing-service.md
@@ -123,6 +123,17 @@ Trust boundaries: every inbound/outbound hop is service↔service over mTLS with
   assessed fee with that idempotencyKey", or 409 "fee exists but was never POSTED — nothing to
   reverse") rather than fabricating a compensating journal against nothing, or against a
   waived/still-pending/failed fee that never moved money in the first place.
+- **Unvalidated input reaching persistence as a 500, and one currency stored under two
+  spellings** → the fee endpoints validated only blankness, so the column definitions in
+  `V1__init_billing.sql` (`currency CHAR(3)`, `cycle_id`/`account_id VARCHAR(64)`) were the only
+  thing asserting the shape of caller input; Postgres' rejection is unmapped, so a client error
+  surfaced as a 500 (found by the authenticated fuzz run, #3038). Two consequences beyond the
+  status code: the error class leaked that the failure came from the database, and a currency
+  differing only in case would have been a *distinct* value in both the column and the
+  `(cycleId, accountId, currency)` idempotency key — a second assessment of the same fee under a
+  second spelling, which the double-charge control above does not catch because the keys differ.
+  Both endpoints now parse currency through `CurrencyCode` (validating + case-normalising) and
+  bound the ids at the column width before anything is persisted.
 
 ## 5. Residual risks / assumptions
 
@@ -176,3 +187,7 @@ Trust boundaries: every inbound/outbound hop is service↔service over mTLS with
   No new trust boundary (same billing→account-service OIDC M2M client as the existing account
   read). Double-gated opt-in (`discovery-enabled` on top of `enabled`, both default off); CSV
   override wins; fail-closed page reads. Rewrote the §5 "no discovery port" residual accordingly.
+- 2026-08-01 — input validation on `POST /api/v1/fees/assess` and `/fees/post` (#3038). No trust
+  boundary change: same callers, same authn/authz, same downstream calls. The change is that
+  caller input is now validated *before* persistence rather than by the schema, closing a 500 and
+  the case-variant idempotency-key gap described in §4.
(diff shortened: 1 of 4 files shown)


JiRaska:
## The test suite was falsified, not just run green

The PR body claims `BillingResourceValidationTest` is the deterministic guard for this fix, so
that claim was checked rather than assumed. Commits `56c95ff73` (probe) and `af6a11122` (revert)
are that check and can be read as a pair.

Why it was needed: this job's **"Upload test reports" step is `if: failure()`**, so a passing run
publishes no JUnit XML, and Gradle prints no test names on success. A green `:test` task therefore
cannot distinguish *"eight cases passed"* from *"eight cases were never collected"* — and Kotlin
JUnit5 silently ignores a test whose inferred return type is not `Unit`, a documented failure mode
in this repo. Every case here is written `fun …(): Unit = runBlocking {`, which is a defence
against that, not a measurement of it.

The probe was an always-false `require` at the top of `validateParams`, chosen so that **all**
eight cases break — a mutation that only breaks the six 400-asserting cases would have left the
two 200-asserting ones unproven. CI then named them:

```
BillingResourceValidationTest > the fuzz input that produced a 500 is now a 400, ... FAILED
BillingResourceValidationTest > a three-letter code that is not ISO 4217 is rejected too() FAILED
BillingResourceValidationTest > an over-long cycleId is rejected before it reaches VARCHAR(64)() FAILED
BillingResourceValidationTest > an over-long accountId is rejected before it reaches VARCHAR(64)() FAILED
BillingResourceValidationTest > an id of exactly the column width is accepted() FAILED
BillingResourceValidationTest > a lower-case currency is normalised before it reaches the use case() FAILED
BillingResourceValidationTest > a blank parameter is still a 400() FAILED
BillingResourceValidationTest > the assess dry-run rejects the same input() FAILED
```

Eight of eight, on eight distinct line numbers. The revert is a `git revert`, and
`git diff 277043fe1 af6a11122` is empty — the tree is byte-identical to the last green commit, so
nothing of the probe survives.

Reviewable leftover: the `if: failure()` upload means this question will be just as unanswerable
for the next test added to this service. Not changing it here — it is a fleet-wide workflow
decision, not a billing one — but flagging it as the reason this took a CI round-trip.

PR #3065, opened 2026-08-01
feat(ai-platform): give the LLM gateway a budget it can actually enforce

JiRaska:
## Summary

`AiFleetDailySpendHigh` (#3043) is a **detective** control — it fires after the money is gone. This is the **preventive** one: the gateway refuses the call.

## Type of change

## Linked issues

Refs #3042, ADR-0112, ADR-0119, ADR-0174, ADR-0175

## Three layers, each biting before the one outside it

| layer | ceiling | answers |
|---|---|---|
| per-agent virtual key | 3–8 USD/day | *who* spent it, and stop just them |
| per-model (in git) | 5–20 USD/day | declarative, reviewable |
| fleet alert | 25 USD/day | last line, not first |

## Why Postgres is a hard prerequisite, not a nice-to-have

Virtual keys are the only mechanism that bounds spend per **caller** rather than per model, and upstream is explicit that they require a `DATABASE_URL` — keys are created through `/key/generate`, never declared in `config.yaml`. Without one the gateway has exactly **one** credential shared by every agent, so:

- "which agent burned the budget" is unanswerable,
- "stop *that* agent" revokes the whole fleet.

A database also makes the counters durable. LiteLLM without one keeps spend in memory, so every pod restart silently resets the budget window to zero — the same *"gauge resets on pod restart"* shape this repo has already been bitten by, and the reason a config-only `max_budget` was rejected as the whole answer.

## Changes

- **CNPG `litellm-db`** in `ai-platform`, following devops-agent's single-owner pattern verbatim: 1 instance, gp3, ECR pull-through `imageName` (so the Kyverno rewrite webhook does not fight the operator's own reconcile loop), bounded WAL, PodMonitor on.
- **`DATABASE_URL` deliberately NOT `optional`**, unlike every other secret ref on this Deployment. A missing provider key degrades to a deterministic fallback, which is safe; a missing `DATABASE_URL` would leave the proxy serving traffic and enforcing **nothing** — a budget that silently does not apply is worse than one never claimed, because the dashboard and the ADR both say the ceiling exists. Reads CNPG's ready-made `uri` key, so there is no URL assembled from parts to drift from the credential.
- **Config revision bumped 2 → 3.** LiteLLM parses `--config` once at startup; without the bump the ConfigMap changes and the running proxy keeps its old model list.
- **Per-model budgets in `litellm-config.yaml`** because they *can* be declarative.
- **`seed-litellm-virtual-keys.sh`** for the per-agent keys, which cannot be. It mirrors the existing `seed-litellm-gateway.sh` rather than becoming a Job: a Job would need the proxy admin credential **and** OpenBao write access in one pod — a strictly larger blast radius than an operator running it once per environment. Idempotent (an existing alias is left alone unless `--rotate`), never prints a key value, and checks `/key/list` first so it cannot report success against a gateway with no database.

## No S3 backup, stated rather than omitted

The manifest records the trade: this holds spend counters and key metadata, not regulated records. Losing it restarts budget windows and forces a re-seed — an operational annoyance, not a retention breach. That calculus changes the day spend history is used as *evidence* rather than as a guardrail, and the comment says so.

## Verification — against the real thing, not the docs

- Booted `litellm v1.74.3-stable` (the deployed tag) with **this ConfigMap's** `config.yaml` against a real Postgres: Prisma migrated the schema, `/health/liveliness` alive in 14s.
- `POST /key/generate` returned a working per-agent key with `max_budget=3.0`, `budget_duration=1d`, scoped to one model — so virtual keys function under this exact config, not just in principle.
- `GET /model/info` confirms **both** per-model ceilings loaded from the ConfigMap (`20.0/1d`, `5.0/1d`). A budget that failed to parse would otherwise be silently absent, which is the failure mode that matters here.
- `shellcheck -S warning` on the new script: clean.
- yamllint with `ci.yml`'s own inline config: clean. Duplicate-key guard: 54 files, clean.
- `gen-network-policies.py` re-run: no diff. CNPG pods are deliberately left unselected by the generator (its own header records why), same as every other single-owner DB in the fleet.

## ⚠️ Honest limitation

**Until each agent is repointed from the shared master key to its own virtual key, the per-agent budgets exist but bind nothing.** The per-model ceilings and the durable counters are live immediately; the per-agent layer needs a credential written per agent, which is an operator step. The script's closing note says this too, so it cannot be mistaken for done.

## Definition of Done

- [x] Hexagonal layering — n/a, infra only.
- [ ] Unit / integration / contract tests — n/a; the executable check is the live LiteLLM + Postgres run above.
- [ ] `./gradlew ...` — n/a, no Kotlin change.
- [x] No new lint warnings (yamllint + shellcheck).
- [ ] OpenAPI / Flyway / Avro — n/a. The schema is Prisma's, migrated by LiteLLM itself at startup.
- [x] Docs updated — the reasoning lives in the manifests and the script header.

## Security checklist

- [x] No secrets, tokens, passwords, or PII in code, config, tests, logs. The script never prints a key value and redacts `.key` from an error response; `DATABASE_URL` is a secret ref.
- [x] New component: a CNPG Postgres, same operator and image path already used by ~10 single-owner clusters in this repo.
- [ ] **Auth-adjacent — reviewer attention.** This introduces a credential-issuing surface (`/key/generate` behind the master key) and a database holding key material. `LITELLM_SALT_KEY` encrypts that material at rest. Worth a second pair of eyes on the master-key handling in the script.
- [x] No PII path changed — the database holds spend counters and key metadata, never prompts.
- [x] No cardholder data path changed.

## Compliance impact

- [x] DORA — ICT third-party cost control moves from detective to preventive, with per-caller attribution.
- [x] EU AI Act adjacent — per-agent spend attribution is the record-keeping half of "which system did what".

## Deployment note

`rules.yaml` and the `.rego` policies are untouched, so no OPA bundle restamp. The CNPG cluster provisions before the Deployment can start (`DATABASE_URL` is a required secret ref, so the pod stays Pending until `litellm-db-app` exists — deliberate). LiteLLM migrates its own schema on first boot.

### DIFF
--- a/openbank-infra/gitops/components/ai-platform/litellm-config.yaml
+++ b/openbank-infra/gitops/components/ai-platform/litellm-config.yaml
@@ -15,6 +15,13 @@ data:
         litellm_params:
           model: deepinfra/deepseek-ai/DeepSeek-V3.2
           api_key: os.environ/DEEPINFRA_API_KEY
+          # Per-MODEL daily ceiling. Declarative and in git, unlike the per-key budgets, which
+          # exist only in the database because upstream creates keys through /key/generate.
+          # 20 USD/day sits just under the 25 USD/day fleet figure AiFleetDailySpendHigh alerts on,
+          # so the preventive control bites BEFORE the detective one pages — an alert that fires
+          # only after the money is gone is a receipt, not a guardrail.
+          max_budget: 20
+          budget_duration: 1d
       # agent-service's admin-UI assistant model (#1919). It sends `model-gateway.models[].id`
       # verbatim, so the model_name MUST stay `llama-3.3-70b-versatile` — that is what repointing
       # agent-service at this gateway costs: an endpoint + key change, nothing else. The Groq key
@@ -24,7 +31,16 @@ data:
         litellm_params:
           model: groq/llama-3.3-70b-versatile
           api_key: os.environ/GROQ_API_KEY
+          # Lower ceiling: this route serves only agent-service's admin-UI assistant, a single
+          # interactive surface, where the ops models serve every scheduled agent. A shared ceiling
+          # would let a runaway ops loop starve the interactive one, or vice versa.
+          max_budget: 5
+          budget_duration: 1d
     general_settings:
+      # Virtual keys and every budget below persist here. Injected from the CNPG litellm-db-app
+      # secret (components/ai-platform/postgres.yaml) — see the DATABASE_URL note in litellm.yaml
+      # for why it is deliberately NOT optional.
+      database_url: os.environ/DATABASE_URL
       # Callers authenticate to LiteLLM with this virtual key (projected from OpenBao), NOT with the
       # upstream provider key — so the DeepInfra key lives only here, never in an agent pod (ADR-0175).
       master_key: os.environ/LITELLM_MASTER_KEY
--- a/openbank-infra/gitops/components/ai-platform/litellm.yaml
+++ b/openbank-infra/gitops/components/ai-platform/litellm.yaml
@@ -26,7 +26,7 @@ spec:
         # would leave agent-service's repointed call failing "model not found" with a green,
         # in-sync ArgoCD — the same silent-degradation shape as an `optional: true` secret ref.
         # BUMP THIS whenever litellm-config.yaml changes; that is what rolls the pod.
-        openbank.tech/litellm-config-revision: "2"
+        openbank.tech/litellm-config-revision: "3"
     spec:
       securityContext:
         runAsNonRoot: true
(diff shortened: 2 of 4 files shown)


JiRaska:
**Merge after #3043.**

Not a code dependency — the seven open observability PRs touch **zero overlapping files**, so there are no textual conflicts in any order. This one is a *semantic* reference: the manifests and the seeding script here describe their budget ceilings as sitting under `AiFleetDailySpendHigh`, which #3043 introduces. Merged first, those comments point at an alert that does not exist yet.

Everything else in this batch is order-independent.

PR #3072, opened 2026-08-01
feat(admin-ui): serve Grafana behind an identity-aware edge gate, not a port-forward

### LINKED ISSUE #3071: Internal tool UIs are port-forward-only: serve them behind an identity-aware edge gate
## Problem

Grafana, GlitchTip's web UI, ArgoCD and the Pact Broker have no Ingress by decision (ADR-0056), so reaching any of them means `kubectl port-forward`. That is genuinely safe — an attacker cannot reach a pre-auth surface that is not routed — and it is also why observability is used less than it should be: an operator already signed in to the console cannot follow an alert to the dashboard that explains it.

The obvious fixes are all worse than the problem:

- Publishing `grafana.open-bank.tech` puts Grafana's login page and its whole `/api/` surface on the internet. Grafana is the richest post-auth pivot in the cluster — it holds datasource credentials for Prometheus, Loki and Tempo.
- Trusting a proxy-injected identity header (`auth.proxy` / `X-WEBAUTH-USER`) makes any pod that can reach the Grafana Service a Grafana admin, with NetworkPolicy as the only control.
- A dedicated identity-aware proxy (oauth2-proxy, Pomerium) is a new public component with its own pre-auth surface and a second definition of "operator" that will drift from the admin-UI's.

## Wanted

An internal tool UI reachable from the console by a signed-in operator, with no pre-auth surface published and no header-trust anywhere.

## Constraint that decides the URL shape

The admin-UI session cookie (`__Secure-authjs.session-token`) is `SameSite=Lax` and **host-only**. A gate on any other hostname would never see it and would deny every request; widening the cookie to `.open-bank.tech` would hand the operator session to every subdomain, including the public GlitchTip ingest host and Keycloak. So the tools have to live under `admin.open-bank.tech` itself — the sub-path is forced, not cosmetic.

## Scope

- [x] Grafana at `admin.open-bank.tech/tools/grafana` behind an nginx `auth_request` gate against the admin-UI session, with Grafana's own Keycloak OIDC kept as a second independent layer
- [x] NetworkPolicy closing the ungated `.svc` path to the Grafana pod
- [x] Kyverno policy rejecting an Ingress in `observability` that is neither gated nor declares why it must stay open to machine callers
- [ ] **Phase 2 — GlitchTip UI.** Not done, and not a small follow-up. `BASE_PATH` exists, but Django `CSRF_TRUSTED_ORIGINS` must accept `admin.open-bank.tech` while `GLITCHTIP_DOMAIN` stays `glitchtip.open-bank.tech` (the domain the displayed DSN is built from — changing it points every device's SDK at the gated host), and its OIDC provider is configured through the Django admin at `/admin/socialaccount/socialapp/`, i.e. runtime database state that GitOps does not declare. Shipping that untested trades a working ingest path for a dashboard link.
- [ ] Phase 3 — ArgoCD, Pact Broker, and anything else currently port-forward-only

## Note on machine callers

Split routing is mandatory where a tool also takes machine traffic. A GlitchTip/Sentry SDK authenticates by DSN and `sentry-cli` by org token; neither can follow a 302 to the console login, so gating a whole host that carries them breaks ingest silently (the SDK swallows the redirect). Those Ingresses stay ungated and now carry an `openbank.io/ungated-machine-callers` annotation stating why.

JiRaska:
Closes #3071

## What

Grafana moves from `kubectl port-forward` to `https://admin.open-bank.tech/tools/grafana`, behind an nginx `auth_request` gate that validates the admin-UI session before nginx proxies anything. **Grafana still has no Ingress of its own** — an unauthenticated request is rejected at the edge and never reaches the workload, so this adds a *routed* path, not an *unauthenticated* one. ADR-0056 holds.

Decision recorded as **ADR-0234** (`accepted` / `partial` — see the GlitchTip section below for why `partial`).

## The four parts

**1. Edge gate** — `openbank-infra/gitops/components/admin-ui/tools-gate.yaml`. A dedicated Ingress carries `/tools/*` plus `auth-url` + `auth-signin`, backed by an `ExternalName` Service into `observability`.

It has to be a *separate* Ingress object: ingress-nginx applies an object's annotations to every location that object declares, so putting `auth-url` on the existing `admin-ui` Ingress would gate `/auth/login` against itself.

**2. Gate endpoint** — `openbank-admin-ui/src/app/api/gate/route.ts`. Returns `204` / `401` / `403`, no body, proxies nothing. nginx maps anything else — including a `302` — to a `500`, which is why this is the one route excluded from the middleware matcher: the middleware answers an unauthenticated request with a redirect to the login page, so the gate would fail closed on exactly the case it exists to reject cleanly. The route repeats the same session + role check itself, and a thrown session decode returns `401`, never an allow.

**3. The tool keeps its own SSO** — Grafana's `auth.generic_oauth` against the `openbank` realm stays on; `root_url` moves to the sub-path and `serve_from_sub_path` is enabled. `auth.proxy` is deliberately **not** enabled and the gate forwards no identity headers: a header-trust setup would make any pod able to reach the Grafana Service a Grafana admin, collapsing two independent layers into one and making a network control load-bearing for authentication. Both layers use the same realm, so the second one costs a silent redirect, not a login.

**4. Behind the edge** — a NetworkPolicy restricts the Grafana pod's ingress to ingress-nginx, `admin-ui` and its own namespace; a Kyverno `ClusterPolicy` rejects an Ingress in `observability` that neither carries `auth-url` nor declares why it must stay ungated. Both Ingresses that exist there today carry the declaration as of this PR, so the policy is `Enforce` from the start rather than an Audit rule that would let the first regression through while reading green.

## Why a sub-path and not `grafana.open-bank.tech`

Not aesthetics. The session cookie is host-only, so a gate on another hostname would see no session and deny everything. Widening it to `.open-bank.tech` would leak the operator session to every subdomain including the public ones — a strictly worse trade than the problem being solved.

## What is NOT in this PR

**GlitchTip's UI is not moved.** It stays port-forward-only, unchanged. `BASE_PATH` exists, but two things are unresolved and neither can be settled from the repo: Django `CSRF_TRUSTED_ORIGINS` must accept `admin.open-bank.tech` while `GLITCHTIP_DOMAIN` stays `glitchtip.open-bank.tech` (the domain the displayed DSN is built from — changing it points every device's SDK at the gated host), and its OIDC provider is Django-admin runtime state, not something GitOps declares. Shipping that untested would trade a working ingest path for a dashboard link. Tracked as phase 2 in #3071.

GlitchTip's ingest Ingress and the RUM gateway are touched only to add the `openbank.io/ungated-machine-callers` declaration — no routing change. An SDK authenticates by DSN and cannot follow a login redirect, so gating a host that carries machine traffic breaks ingest silently.

## Trade-offs accepted

- **The admin-UI becomes an availability dependency of Grafana.** `auth_request` fails closed, so a down gate means unreachable dashboards. Correct direction, and the reason `grafana-local.sh` stays as break-glass.
- **`grafana-local.sh` can no longer complete an OIDC login.** Grafana has exactly one `root_url` and it now carries the public sub-path. The script is rewritten to say so; the port-forward still works for a local Grafana admin login.
- **`/api/gate` is the first route excluded from the middleware matcher** since ADR-0080 P0 set it to "everything except Auth.js and static assets".

## Verification

`npm test` 905/905 · `tsc --noEmit` · eslint · `next build` (the `/api/gate` route is emitted) · yamllint with the CI config · shellcheck · `check-adr-registry.sh` · `check-gitops-secret-refs.py` · `gen-network-policies.py` drift gate — all green after the rebase onto current `main`.

The new wiring assertions were run against inputs they must flag, and **two of them passed against broken input on the first attempt**:

- the middleware-exclusion check matched its own explanatory comment three lines above the matcher, so a whole-file grep stayed green after the exclusion itself was deleted. It now strips comments and reads `config.matcher`.
- `[a-z0-9-]+` on a typo'd `?tool=grafanaX` captured the `grafana` prefix and reported agreement with an Ingress pointing at a tool the gate does not know. The class is now case-insensitive.

Both fixed, all four wiring assertions then went red against the broken input, and green after restore.

## Needs live verification post-deploy

- `auth_request` + `ExternalName` upstream behaviour on ingress-nginx 4.15.1
- the Kyverno policy against a real Kyverno — `admissionWebhooks` is disabled in the sandbox, so an invalid Ingress is silently dropped rather than rejected
- Grafana rendering under the sub-path (assets, live panels)
- **the Keycloak realm import runs on cold start only**, so `openbank-grafana`'s new redirect URI will not appear in the sandbox realm without a manual add or a re-import


JiRaska:
## Update: the Kyverno policy now ships as `Audit`, not `Enforce` (`ad2fa5183`)

Verified against the sandbox rather than assumed. Enforce would have broken two healthy workloads on merge: the policy auto-syncs from `components/kyverno`, but the two machine-caller declarations live in `gitops/apps/{glitchtip,rum-gateway}.yaml`, which nothing syncs — and both of those Applications are `prune: true, selfHeal: true`, so ArgoCD keeps re-applying their rendered Ingress. Enforce would reject ArgoCD's own re-apply of two Ingresses that are correct in git.

```
NAME          HOST                       DECLARED   AUTHURL
glitchtip     glitchtip.open-bank.tech   <none>     <none>
rum-gateway   rum.open-bank.tech         <none>     <none>
```

ADR-0030's SBOM policy went Audit→Enforce on this exact shape and killed five workloads for four days (#1197). The flip is tracked in #3106 with its precondition as a single command, and with a decoy Ingress to exercise the failure path afterwards.

Also verified against the live API server while there: all four new objects pass `kubectl apply --dry-run=server`; the Grafana pod really carries `app.kubernetes.io/name: grafana` (so the NetworkPolicy selects a pod, not nothing); `kube-prometheus-stack-grafana` is ClusterIP :80, matching the ExternalName target and port; and both `ingress-nginx` and `admin-ui` namespaces carry `kubernetes.io/metadata.name`, so the `namespaceSelector`s match.

## Post-merge runbook

Merging this PR alone does **not** make the gate reachable — three of the pieces do not auto-sync. In order:

**1. The admin-UI image.** `/api/gate` is a route in the Next.js build, so it only exists once a new image ships. `admin-ui-deploy.yml` fires on merge, builds, pushes to ECR and **opens a second GitOps PR** bumping the image tag. That PR has to be merged too, or ArgoCD keeps running the old image and every `/tools/*` request is a 500 (nginx's mapping of the middleware's 302).

**2. Auto-syncs, no action needed.** `components/admin-ui` (the tools-gate Ingress + Service), `components/kyverno` (the policy), `components/observability` (the NetworkPolicy), `components/keycloak` (the realm ConfigMap) — all four ArgoCD Applications are `prune: true, selfHeal: true`.

**3. Grafana's `root_url` — manual.** The change is in the `kube-prometheus-stack` **Application object itself** (a Helm-repo source with an inline `valuesObject`), and nothing syncs `gitops/apps/`:

```
kubectl apply -f openbank-infra/gitops/apps/kube-prometheus-stack.yaml
```

Until this runs, Grafana still answers on `root_url=http://localhost:3000` and will redirect `/tools/grafana` away. Same command shape for `glitchtip.yaml` and `rum-gateway.yaml` (the #3106 precondition).

**4. The Keycloak redirect URI — manual, and the one most likely to be missed.** `realm-template.json` is only read on a **cold start** import, so syncing the ConfigMap does not update a running realm. Without this, the Grafana OIDC round-trip fails with `Invalid parameter: redirect_uri` — which looks like a gate bug and is not one. Add `https://admin.open-bank.tech/tools/grafana/login/generic_oauth` to the `openbank-grafana` client's redirect URIs (and `https://admin.open-bank.tech` to its web origins), via the admin console or `kcadm`.

**5. Then test**, signed in to the console:

- `https://admin.open-bank.tech/tools/grafana` → dashboards, no second login prompt (silent OIDC redirect through the existing Keycloak session)
- the same URL in a private window → the console login page, **never** a Grafana login page. That is the whole point: the gate must reject before nginx proxies.
- `curl -so /dev/null -w '%{http_code}' https://admin.open-bank.tech/tools/grafana/` with no cookie → expect a redirect to the console login, not `200`
- a signed-in user without `system:view` (neither `ROLE_ADMIN` nor `ROLE_OPERATOR`) → `403` from the gate, no Grafana content
- GlitchTip ingest still works — a crash from the app still lands. It is on a different host and deliberately ungated, but it is the thing that breaks silently if anything about the Ingress annotations went wrong.

Issue #3076, opened 2026-08-01
SDLC throughput: merge rate halved while CI got faster — the required gate job grows linearly and 40% of PRs are bots


JiRaska:
## What

Merge throughput dropped ~48% week over week while CI itself got no slower. This tracks the four structural causes and the work to remove them.

## Measurements (2026-08-01, `gh` API over merged PRs and workflow runs)

**Throughput and latency**

| week | merged PRs | bot share | human PR cycle median | p90 |
|---|---|---|---|---|
| 2026-W29 | 627 | 36% | 17.1 min | 398 min |
| 2026-W30 | 689 | 39% | 6.4 min | 286 min |
| 2026-W31 (6 days) | 309 | 41% | 24.0 min | 437 min |

~98 merges/day to ~51/day.

**It is not the build.** `services-ci.yml` median duration went 4.4 min (07-12) to 0.8 min (07-31); queue wait median 0.00 min on every sampled day. Nothing is waiting for a runner.

**What did grow, linearly:** `ci.yml`'s `validate` job — the `Validate manifests` required check.

| | `validate` steps | `.github/scripts` files | `ci.yml` median |
|---|---|---|---|
| 4 weeks ago | 35 | 24 | 0.7 min |
| today | 97 (79 in one job) | 80 | 2.4 min |

79 steps, one runner, one checkout, strictly sequential, on a required check that **every** PR pays — including the ~40% that are bot PRs. Every gate added taxes every PR in the repo, permanently, and nothing bounds that.

**Where the PR slots go** (733 merged in the last 7 days):

| kind | n | median | total open-PR time |
|---|---|---|---|
| release-please bot | 194 | 9.3 min | 675 h |
| feat | 94 | 48.0 min | 386 h |
| fix | 101 | 17.7 min | 204 h |
| auto-deploy bot | 92 | 2.3 min | 17 h |

**Run volume:** 7 298 workflow runs on 07-31. In a 1 000-run sample, 27% concluded `skipped`; `Auto-retry spot-killed CI runs` alone was 22% of all runs.

**The tail is human.** p90 is ~7 h and the ten slowest non-bot PRs ran 20–109 h. None of them was waiting on CI.

## Work

- [ ] **1. Shard the gates.** Split `validate` into a matrix so wall time is `max(shard)` not `sum(gates)`. `Validate manifests` must keep that exact job name — it is a required context in the `main-protection` ruleset — so the shards aggregate into it rather than replacing it.
[...]

Issue #2064, opened 2026-07-24
analytics namespace ships with no NetworkPolicy — and the ADR-0081 drift gate misses it (untracked-file blind spot)


JiRaska:
## Problem (two parts)

**1. The `analytics` namespace has no committed NetworkPolicy.** #1974 (business-funnel analytics — ClickHouse + analytics-sink + Kafka mtls) added `openbank-infra/gitops/components/analytics/` but no `network-policies.yaml`. Running `gen-network-policies.py` produces two policies the tree is missing:
- `analytics-sink-ingress-allow-list` (admits admin-ui :8134, observability :8086, security-scanner :8086)
- `clickhouse-ingress-allow-list`

So analytics currently has the default (open) ingress posture, unlike every other component, while its declared GitOps edges say otherwise. ADR-0081: allow-lists are the enforced baseline; a namespace with real cross-ns callers and no policy is a gap.

**2. The drift gate has an untracked-file blind spot (the reason #1 slipped through).** `.github/workflows/ci.yml` (the gen-network-policies drift gate) runs:
```
python3 openbank-infra/scripts/gen-network-policies.py
git diff --exit-code openbank-infra/gitops/components/
```
`git diff --exit-code` only inspects **tracked** files. When the generator emits a policy for a **brand-new** component that has never had one committed, the output is an **untracked** file — invisible to `git diff` — so the gate stays green while the new namespace is unprotected. Only a *modification* to an already-tracked policy trips it (which is why e.g. the copilot repoint #2062, touching the tracked ai-platform policy, was caught but this was not).

## Fix

- Commit the generated `analytics/network-policies.yaml` (run `python3 openbank-infra/scripts/gen-network-policies.py` and add the new file).
- Close the gate blind spot: after regenerating, fail on untracked output too — e.g. `git add -A --intent-to-add openbank-infra/gitops/components/ && git diff --exit-code openbank-infra/gitops/components/`, or an explicit `git status --porcelain` check for untracked `network-policies.yaml` under `components/`.

## Evidence
- `git show origin/main:openbank-infra/gitops/components/analytics/network-policies.yaml` → absent.
- Generator emits it (verified in a clean worktree from origin/main).
- Introduced by #1974 (commit 7a14a799).

Refs ADR-0081, #854.

PR #3077, opened 2026-08-01
ci(governance): declare the gates in a manifest and shard them, so the gate count stops costing wall time

JiRaska:
Refs #3076 (items 1 and 2).

## Why

`Validate manifests` is a required check on every PR, and it was 79 sequential steps in one job. That count is not stable — it grew 35 → 97 in four weeks, taking the job's median from 0.7 to 2.4 min, linearly. Roughly 40% of merged PRs are release-please / auto-deploy bots, so every gate added was charging every PR in the repo, forever, with nothing bounding it.

The second cost was quieter: a gate that IS a workflow step cannot be enumerated. Not by a human, not by someone wanting to run one locally, and not by `check-advisory-gate-registration.py` — which had to regex workflow YAML and guess "is this advisory?" from a step name, and once flagged itself for having the word in its own step name (#2450).

## What changed

**`.github/gates/gates.yaml`** — 67 gates declared, each with `id` / `group` / `mode` / `when` / `needs_base` / `selftest` / `run`.

**`.github/scripts/run-gates.py`** — runs them concurrently within a shard, with a `--list` / `--group` / `--only` / `--self-test` CLI.

**`ci.yml`** — 1 464 → 547 lines. The `validate` job becomes an 8-way matrix (`lint gitops supplychain registry kotlin api data security`) plus a thin aggregator job that keeps the exact name **`Validate manifests`**, because that string is a required context in the `main-protection` ruleset. Aggregating rather than renaming means this ships without a GitHub-side ruleset edit — the same shape as `services-ci.yml`'s `all-green`, for the same reason.

**`check-advisory-gate-registration.py`** — now reads the manifest as well as workflow steps. This was not optional: the 79 steps it used to scan moved out of `ci.yml` in this PR, so without it the gate would have quietly found nothing and gone green. It reads `mode:` directly instead of inferring from a step name. `--expect-manifest` (wired in CI) fails if the manifest contributes zero advisory gates, so a future move cannot empty it silently either.

## Measured — on CI, not locally

Baseline: the old serial `Validate manifests` job over its last 25 successful PR runs on `main` — **median 301 s**, p90 306 s, max 308 s.

This PR, run 30692362431:

| job | wall |
|---|---|
| gates (registry) | 58 s |
| gates (gitops) | 39 s |
| gates (lint) | 26 s |
| gates (data) | 22 s |
| gates (supplychain) | 20 s |
| gates (kotlin) | 18 s |
| gates (security) | 16 s |
| gates (api) | 10 s |
| **Validate manifests** (aggregator) | **3 s** |

**301 s → ~61 s critical path, 4.9x.** The shards total 209 s of wall between them, so the parallelism is doing the work rather than something having quietly stopped running — the shard totals still exceed the old job's *reported* wall because the old one paid one checkout for all 79 steps while eight shards each pay their own.

The floor is now `registry` at 58 s, which is one gate: ADR registry integrity. #3109 takes that gate's dominant check from O(references × ADRs) to O(n), which should make `gitops` the new floor.

## Fidelity

The manifest was extracted **mechanically** from the workflow, not retyped — a transcription is a different program. Verified by set-comparing every `run:` body between `origin/main`'s `ci.yml` and the manifest: 79 original bodies, all accounted for, with exactly two deliberate differences —

- the hash-pinned PyYAML install moved out of the `gen-network-policies` gate (where it sat only because that was the first step needing it) into the shard, since `run-gates.py` itself imports yaml and there is no longer a "first step";
- `check-advisory-gate-registration.py` gained `--expect-manifest`.

Rebased onto #3074 mid-flight; its Loki `--since` scoping was carried into the manifest and re-verified against `origin/main` byte for byte.

## What running them concurrently broke, and how

Three real defects, each found by running the thing rather than reasoning about it:

**Self-test polarity.** The first runner treated a self-test as a known-positive that must FAIL, and duly reported all 11 existing self-tests as "unfalsified" on a repo where all 11 were fine. They are the checkers' own `--self-test` harnesses: they build the fixture, run the gate, assert it went red, and exit **0** when the gate is falsifiable. Both shapes now exist and `selftest_expect:` declares which — there is no signal in the command text, and guessing wrong is silent in the safe-looking direction.

**Shared git index.** `gen-network-policies` and `service-runbook-drift` both `git add --intent-to-add` so a generated-but-never-committed file shows as drift instead of staying invisible as untracked (#2064). Sequentially that was fine; concurrently they race for `.git/index.lock`. Each gate now gets a private `GIT_INDEX_FILE` copy, which also stops any gate leaking staged state into another's `git diff` — coupling the old job had and nobody had reason to notice.

**`${{ }}` in a `run:` body.** One gate carried a workflow expression into the manifest, where nothing expands it, so bash died with `bad substitution`. Fixed via `$BASE_REF` from the shard, and `run-gates.py` now refuses to load a manifest containing one. On its first run that check flagged the comment that *explains* why the expression was removed — same shape as #2450 — so comment lines are stripped first, and the self-test asserts both directions.

## Falsification

Everything here was run against an input it must flag:

- `run-gates.py --self-test` drives a synthetic manifest covering pass, enforced-fail, advisory-fail, both self-test polarities in both verdicts, `pull_request`-only skip, timeout, empty manifest, `needs_base` mismatch in both directions, and the `${{ }}` guard on a command (must flag) and in a comment (must not). It asserts `report()`'s exit code too, so a runner that classified correctly and returned 0 anyway would not pass. It runs in CI as its own gate (`gate-runner-self-test`).
- Verified the self-test itself can fail: removed the comment-stripping and it correctly reported `load() rejected a ${{ }} mentioned only in a comment`.
- `gen-network-policies` under the private index: red on an untracked file under `components/`, green when removed, and the real index left untouched.
- `--expect-manifest`: red with the manifest absent, red with the manifest present but all-enforced, green on the real one.
- `actionlint` on the new `ci.yml` — validated against a deliberately broken copy first (`property "nope" is not defined`), so its silence means something.

## Not in this PR

No `rules.yaml` change: editing it restamps ~65 OPA bundles, which would bury this diff and give the PR a shelf life measured in hours. Items 3–5 of #3076 follow separately.

## Conflicts to expect

Four open PRs add a step to `validate` and will conflict with this by construction: #3019, #3009, #3008, #2980. Migration is mechanical — the step's `run:` body becomes a manifest entry's `run:`, plus `id` / `group` / `mode`, and `selftest` if it has a self-test step. Happy to do the rebase for whichever lands second; the deletion is the whole conflict, so there is nothing to reconcile semantically.

### DIFF
--- a/.github/gates/gates.yaml
+++ b/.github/gates/gates.yaml
@@ -1,0 +1,1227 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) OpenBank contributors. Licensed under the Apache License, Version 2.0.
+#
+# THE GATE MANIFEST — every static governance gate CI runs, declared in one place.
+#
+# WHY THIS FILE EXISTS
+#   These gates used to be 79 sequential steps in ci.yml's `validate` job. That shape had
+#   two costs, both of which compounded with every gate added:
+#
+#     1. SERIAL. One runner, one checkout, 79 steps end to end. Between 2026-07-04 and
+#        2026-08-01 the job went 35 -> 97 steps and its median wall time 0.7 -> 2.4 min,
+#        growing linearly with the gate count. `Validate manifests` is a REQUIRED check on
+#        every PR — including the ~40% that are release-please / auto-deploy bot PRs — so
+#        each new gate taxed every PR in the repo, permanently. Sharding turns that into
+#        max(shard) instead of sum(all).
+#
+#     2. UNADDRESSABLE. A gate WAS a workflow step, so nothing could enumerate the set:
+#        not a human, not a developer wanting to run one locally, and not
+#        check-advisory-gate-registration.py — which had to regex workflow YAML and infer
+#        "is this a gate?" from a step name (and, per its own header, flagged itself once
+#        for having the word `advisory` in a step name, issue #2450).
+#
+#   Declaring them here makes the set enumerable, runnable outside CI, and parallelisable:
+#
+#       python3 .github/scripts/run-gates.py --list
+#       python3 .github/scripts/run-gates.py --group kotlin
+#       python3 .github/scripts/run-gates.py --only domain-purity-gate-hexagonal-invariant
+#       python3 .github/scripts/run-gates.py --selftest-only        # falsify every gate
+#
+# THE CONTRACT — read this before adding an entry
+#   id          stable slug, unique across the file. Used by --only and in the run log.
+#   name        display name, carried verbatim from the step it replaces.
+#   group       which matrix shard runs it. Shards are WALL-TIME buckets, not a taxonomy —
+#               rebalance them freely when one gets slow. A gate and its self-test always
+#               run together, so that pairing is not a constraint on grouping.
+#   mode        `enforced` (a failure fails CI) or `advisory` (a failure is a ::warning and
+#               the run continues). An `advisory` entry MUST be registered in rules.yaml
+#               with a `target_enforce_date`: check-advisory-gate-registration.py reads
+#               THIS FILE, so an advisory gate cannot hide here either (issue #2392).
+#               Note that most gates here are ALSO advisory internally — they print
+#               ::warning and exit 0 unless passed --enforce. `mode:` is about what the
+#               RUNNER does with a non-zero exit; it does not override the script's own
+#               flag, and the two must agree. See rules.yaml: gate_modes.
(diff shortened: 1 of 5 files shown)


JiRaska:
`CodeQL (java-kotlin, manual)` is red on this PR and it is not caused by it — the two-phase build's empty-trace fallback cannot re-compile, diagnosed in #3078. It reproduces on any PR that triggers CodeQL via `.github/workflows/` without changing compilable source. CodeQL is not a required context, so this is not blocking.

Issue #3039, opened 2026-08-01
api-fuzz-authenticated: 7 of 16 money-path services fail to boot in the harness


JiRaska:
Seven of the sixteen money-path services in the authenticated fuzz matrix **never reached the fuzz
step** — they failed to boot in the harness. That is a harness gap, not a service defect, and it is
what currently caps this lane's coverage at 9 of 16.

From run [30687065400](https://github.com/JiRaska/open-bank-oss/actions/runs/30687065400), the first
full-matrix run (branch of #3024).

## Which, and what the log actually says

| Service | Cause as logged |
|---|---|
| `openbank-sepa-payment` | not determined from the log tail |
| `openbank-account-service` | Port already bound: 8080 |
| `openbank-sca-service` | not determined from the log tail |
| `openbank-ledger-service` | not determined from the log tail |
| `openbank-lending-service` | not determined from the log tail |
| `openbank-transaction-service` | not determined from the log tail |
| `openbank-domestic-payment` | not determined from the log tail |

**7 of 16** services never reached the fuzz step.

**"Not determined" means exactly that.** The step tails only the last 60 lines of the boot log, and
for six of the seven the stack trace's root cause is above that window. I am not extrapolating
`account-service`'s cause to the others — one measured instance is not a pattern, and guessing here
would send the fix in the wrong direction.

## The one cause that is established

`openbank-account-service`: **`QuarkusBindException: Port already bound: 8080`**.

The harness runs Keycloak on host port **8080** (`docker run -p 8080:8080`), which is also Quarkus's
default HTTP port. `account-service` declares `quarkus.http.port: 8100` and does not override it in
`%dev`, so something in dev mode still reaches for 8080 — worth understanding rather than assuming.

Regardless of that detail, **moving Keycloak off 8080 removes the whole collision surface** and costs
one line plus the three URLs that reference it:

```yaml
docker run -d --name fuzz-keycloak -p 8543:8080 …
… http://localhost:8543/realms/openbank/…
```

That is worth doing whether or not it explains the other six.

## Suggested next step

1. Raise the boot-log tail from 60 lines, or upload the full `${svc}-boot.log` as an artifact
   unconditionally — the artifact upload already exists but only the tail is echoed. Without that,
   this issue cannot be closed for six of the seven services.
2. Move Keycloak to 8543.
3. Re-run and re-triage.

## Why this matters

`openbank-clearing-service` fuzzed clean over 621 cases and eight other services produced real
[...]

PR #3079, opened 2026-08-01
fix(ci): make the fuzz harness able to boot reactive-only services, and report why when it cannot

JiRaska:
Triages all seven boot failures in #3039 and fixes the four that are harness gaps.

**No re-run was needed.** The full boot logs were already uploaded as
`api-fuzz-auth-reports-<svc>` artifacts from run 30687065400 — only the *echoed* tail was too
short to see the cause. Each log is ~30k lines of dev-mode DEBUG; the root cause sits thousands
of lines above a 60-line window.

## The seven, with causes

| Service | Real cause | Class |
|---|---|---|
| ledger, account, sca, transaction | `Unable to find datasource '<default>' … Bean is not active [AgroalDataSource]` | harness gap — fixed here |
| sepa-payment | `Schema validation: missing table [object_store_blobs]` | service defect |
| lending | `wrong column type … [currency] in table [collateral]` | service defect |
| domestic-payment | `wrong column type … [reject_detail] in table [domestic_payments]` | service defect |

## Correction to the issue

#3039 records `Port already bound: 8080` as account-service's one *established* cause. It is in
**all seven** logs, and in every one of them it is a **consequence**: the frame above it is
`VertxHttpRecorder.startServerAfterFailedStart`. Quarkus raises a fallback error-page server on
its default port *after* a failed start, and Keycloak holds 8080. So the one cause the issue was
confident about is the one artifact that says nothing about why any service failed.

Moving Keycloak to 8543 removes it from every log, which is worth more than the port itself.

## Why the four could not boot

They declare only a reactive datasource URL, so the default Agroal (JDBC) datasource is never
configured — while `flyway.migrate-at-start: true` needs one. Production is fine because the
gitops env supplies `QUARKUS_DATASOURCE_JDBC_URL`; the harness never did. One flag fixes it.

Three of the four additionally carry a `quarkus.flyway.datasource:` block with `db-kind`,
`username` and `jdbc.url` nested under it. That is not a datasource *definition* path, and the
boot error proves Quarkus ignored it — it says the default datasource is absent while that block
sits right there in the file. **These four services cannot boot from their committed
configuration alone**; that is a service-level defect worth its own issue and is exactly the class
#2872 exists for. Not fixed here — this PR keeps to the harness.

## Verification of the scope change

The DB-name regex now accepts any `postgresql://` URL. Checked against all 54 service configs
rather than assumed: 0 services lose derivation, 1 gains it (`settlement`), and exactly 1 value
changes — ledger, `openbank_ledger_it` -> `openbank_ledger`. The old pattern's first match in
ledger's file is an **integration-test** profile, so the unauthenticated lane had been
provisioning ledger's IT database. That is a fix, but it is a behaviour change and belongs in the
open.

I also stated mid-investigation that the unauthenticated lane was silently skipping those four
services. That was wrong — it derived a DB name for them from an IT-profile `jdbc:` URL, so they
were in scope and failed at boot. Only `settlement` was actually skipped.

## Not fixed here

The three schema-validation failures are real entity/DDL drift. `object_store_blobs` is an
`@Entity` in `openbank-libs-runtime` (`PostgresBlobStore`) and only `openbank-document-service`
has a migration creating it — sepa-payment reaches it through the ordinary
`implementation(project(":openbank-libs-runtime"))` that 47 modules share. Whether the other 46
are equally exposed is **not** answerable from these logs: validation aborts at the first
mismatch, so lending and domestic-payment stopped before they would have reached it. Settling
that needs a per-service boot, which is #2872.

Refs #3039, #3024, #2872

### DIFF
--- a/.github/workflows/api-fuzz-authenticated.yml
+++ b/.github/workflows/api-fuzz-authenticated.yml
@@ -83,7 +83,7 @@ jobs:
       - name: Start Keycloak (throwaway realm)
         run: |
           set -euo pipefail
-          docker run -d --name fuzz-keycloak -p 8080:8080 \
+          docker run -d --name fuzz-keycloak -p 8543:8080 \
             -e KC_BOOTSTRAP_ADMIN_USERNAME=admin \
             -e KC_BOOTSTRAP_ADMIN_PASSWORD=admin \
             -v "$PWD/.github/workflows/keycloak:/opt/keycloak/data/import:ro" \
@@ -93,7 +93,7 @@ jobs:
           KC_UP=0
           for i in $(seq 1 60); do
             code="$(curl -s -o /dev/null -w '%{http_code}' \
-              http://localhost:8080/realms/openbank/.well-known/openid-configuration || true)"
+              http://localhost:8543/realms/openbank/.well-known/openid-configuration || true)"
             if [ "$code" = "200" ]; then KC_UP=1; break; fi
             sleep 3
           done
@@ -130,8 +130,9 @@ jobs:
           echo "==> [${svc}] starting quarkusDev with OIDC on"
           ./gradlew ":${svc}:quarkusDev" \
             -Dquarkus.console.basic=true --console=plain --quiet \
+            -Dquarkus.datasource.jdbc.url="jdbc:postgresql://localhost:5432/${DB}" \
             -Dquarkus.oidc.enabled=true \
-            -Dquarkus.oidc.auth-server-url=http://localhost:8080/realms/openbank \
+            -Dquarkus.oidc.auth-server-url=http://localhost:8543/realms/openbank \
             -Dquarkus.oidc.client-id=openbank-services \
             -Dquarkus.oidc.credentials.secret="${OIDC_CLIENT_SECRET}" \
             > "fuzz-reports/${svc}-boot.log" 2>&1 &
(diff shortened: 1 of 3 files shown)


JiRaska:
## Correcting myself: this is mergeable as-is, no rebase needed

I had this down as blocked on a conflict. It is not — `mergeStateStatus` is `CLEAN` and all five required contexts are green.

The conflict I hit was an artefact of **my** attempt to rebase it, not a property of the PR.

## Why the diff looks like it re-lands #3103

The file list shows `security.yml` at `76+/15-`, which is exactly #3103's change — already squash-merged to `main`. That reads like a duplicate, but it is not:

```
security.yml @ merge-base 04dfae61c : a1cffdb1fc6f
security.yml @ origin/main          : 3617cadc824c
security.yml @ this branch          : 3617cadc824c   <- identical to main
```

GitHub computes the PR file list against the **merge-base**, and the merge-base predates #3103's squash. So the branch's `security.yml` work shows as a diff even though its *content* already matches `main` byte for byte. Merging it is a no-op for that file; the real delta is the two fuzz workflows.

## Why rebasing would have been the painful route

`git rebase origin/main` replays each commit, and this branch carries six:

```
bf6eee840  Merge the CodeQL base-ref fix
56e940dbe  fix(ci): resolve the CodeQL base ref properly
ecf650533  Merge fix/codeql-tracing-rerun-tasks into this branch
27c2aef2a  perf(ci): scope CodeQL to the languages a PR actually changed
439e1e43b  fix(ci): make the CodeQL cold re-run actually recompile   <- squashed onto main as #3103
b26d48d54  fix(ci): make the fuzz harness able to boot reactive-only services
```

Replaying `439e1e43b` against a `main` that already has its squashed equivalent conflicts, and resolving that means deciding which version of `security.yml` is authoritative — an authoring decision, not a mechanical one. Merging sidesteps it entirely, which is the stacked-PR guidance in `CLAUDE.md`.

## Context

This is the fix for what I reported on #3024 — the authenticated fuzz harness failing to boot 15 of its 16 targets. Good to see the cause found.

*Reviewed while draining the PR queue; head unchanged at `bf6eee840` since I looked.*

PR #3103, opened 2026-08-01
fix(ci): make the CodeQL cold re-run actually recompile, and fail loudly if it does not

JiRaska:
`CodeQL (java-kotlin, manual)` fails on any PR that touches no Java/Kotlin — it went red on
[#3079](https://github.com/JiRaska/open-bank-oss/pull/3079), which changes two workflow YAMLs and
nothing else.

```
CodeQL could not process any code written in Java/Kotlin.
```

## Two commits, two different problems

**1. Scope CodeQL to the languages the PR changed** (the cost fix, and the one that makes the red
go away).

The `run_codeql` flag was all-or-nothing and listed `.github/workflows/` as a code path. So a
workflow-only PR ran **both** legs: no new coverage, and a guaranteed red on the java-kotlin one,
because there is nothing for the tracer to see. Detection is now per language and the job consumes
a derived matrix — `[]` skips the job outright. Workflows are in neither pattern.

Cheaper than the alternative (forcing a cold recompile so the tracer has something to look at):
this saves the entire job on doc/CI-only PRs, which are frequent here, instead of adding a full
fleet compile to them.

**2. Fix the cold re-run so it actually recompiles** (a latent bug, still reachable).

The step already detects "everything came from cache" and re-runs. **The detection is correct** —
I checked before assuming otherwise: of 123 `compileKotlin` lines in run 30692843630, zero were
bare (executed); all were `FROM-CACHE`/`UP-TO-DATE`/`NO-SOURCE`. The `472 actionable tasks: 294
executed` line is misleading — those are jars and resources, not compilation.

The **remedy** is what fails: `--no-build-cache` bypasses the build *cache* but not Gradle's
UP-TO-DATE checks, and after the first run every compile output is already in `build/`. That
fallback produced `60 UP-TO-DATE, 60 FROM-CACHE, 0 executed` and CodeQL failed anyway.
`--rerun-tasks` is what forces re-execution, and the step now asserts the re-run compiled
something instead of exiting 0 regardless.

Commit 1 makes commit 2 unreachable for workflow-only PRs, but not for the case it was written
for — a code PR whose compiles all restore from cache, e.g. on a re-run.

## Safety of skipping

Not a required status check: `main-protection` requires `all-green`, `Validate manifests`,
`Gitleaks`, `issue-hygiene`, `OPA policy gate`. Verified via the rulesets API, not assumed. No job
`needs:` the codeql job either, so a skip cannot strand a dependent. Weekly and main-push scans are
unchanged.

## Verification

GitHub's default shell is `bash -e {0}`, so `grep -q … && VAR=true` **terminates the step** when it
does not match — the common case here, and it would have left `codeql_matrix` unset while the job's
`if:` still evaluated to true. Every branch is a full `if/fi` for that reason.

The two emitted shell functions were extracted **from the workflow file** (not retyped) and run
against nine file-list fixtures:

```
  workflow YAML only  -> []                                      OK
  docs only           -> []                                      OK
  kotlin only         -> [java-kotlin]                            OK
  gradle only         -> [java-kotlin]                            OK
  typescript only     -> [javascript-typescript]                  OK
  package-lock only   -> [javascript-typescript]                  OK
  both                -> [java-kotlin, javascript-typescript]     OK
  Dockerfile          -> [java-kotlin]                            OK
  THIS PR (yaml only) -> []                                       OK
```

Known-negatives included on purpose — a classifier that only ever returns "run everything" would
pass a positives-only suite. The emitted JSON was also parsed to confirm it is valid input for
`fromJSON`.

Consequence for this PR itself: it changes only `security.yml`, so its own CodeQL job should now
**skip** rather than run and fail. That is the visible proof the change works.

Refs #3079

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -37,6 +37,7 @@ jobs:
     outputs:
       run_sbom: ${{ steps.f.outputs.run_sbom }}
       run_codeql: ${{ steps.f.outputs.run_codeql }}
+      codeql_matrix: ${{ steps.f.outputs.codeql_matrix }}
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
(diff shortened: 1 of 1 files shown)


JiRaska:
## Third commit: the scoping was silently not engaging

Worth reading, because the first version of this PR went **green for the wrong reason**.

After the per-language change, `CodeQL (java-kotlin, manual)` still ran and *passed* on
[#3079](https://github.com/JiRaska/open-bank-oss/pull/3079) — a workflow-only PR that should have
skipped it. Same code produced `codeql matrix: []` on this PR and both languages on that one.

The classifier was not at fault: replaying it locally against that branch's real diff returns `[]`.
The base ref was fetched with `--depth=1`, and a **shallow ref carries no ancestry**, so
`git merge-base` returns nothing, `BASE` is empty, and the step falls through to "run everything".
Whether that happens depends only on how far the branch has drifted from `main` — which is exactly
why it appeared to work on one PR and not the other.

The more useful half is that the fallback was **silent**. "I ran everything because I could not
compute a diff" is indistinguishable from "scoping decided to run everything", so the detection
could stop working entirely and no run would say so. That applies to the pre-existing `run_codeql`
flag too — it had the same quiet `else`.

Fixed by fetching the base ref with an explicit refspec and no `--depth=1`, and by emitting a
`::warning` when `BASE` cannot be resolved.

## Verified at the source, not inferred from "skipping"

Both PRs, current heads, reading only the step output **after the last `##[endgroup]`**:

```
#3079  codeql matrix: []  | ::warning::merge-base occurrences: 0
#3103  codeql matrix: []  | ::warning::merge-base occurrences: 0
```

That last column matters. The naive grep over the whole job log reports **1** on both, because a
job log contains the step's own `run:` script — the `echo "::warning::merge-base …"` line is in the
listing whether or not it executed. Confirmed by counting the same pattern in the pre-`endgroup`
region: exactly 1, i.e. the listing. So the zero above is a measurement, not an absence of
evidence.

Net effect on #3079: 27 check contexts down to 25, the two per-language legs gone rather than run.

PR #3058, opened 2026-08-01
feat(account): delegation-grant enforcement projection (ADR-0232 D3)

JiRaska:
## Summary

First consumer of `openbank.delegation.events` (issue #2990, stacked on #2971): account-service maintains a local `account_delegation_projection` and the authorization guard becomes **owner OR legacy AccountAuthorization OR an ACTIVE in-window delegation grant** — enforcement fully local, never a synchronous call to delegation-service (guarded by a tripwire test).

**⚠️ Stacked PR: base is `feat/delegation-service` (#2971). Retarget to `main` once #2971 merges.**

## Type of change

- [x] feat (new functionality)
- [ ] fix (bug fix)
- [ ] refactor (no behavior change)
- [ ] perf (performance)
- [ ] docs
- [ ] chore (build / tooling)
- [ ] security (private until disclosed)
- [ ] breaking change

## Linked issues

Refs #2990 — delivers AC1–AC7, AC11, AC13, AC14 (+ AC15 account-service half). Card-issuance, savings goals, propose-only (AC8) and onBehalfOf audit threading (AC9/AC10/AC12) are follow-up slices tracked in the issue.

## Changes

- **Projection**: `account_delegation_projection` (+ caps table, Flyway **V16**), `DelegationEventConsumer` on `delegation-events-in` (DLQ + bounded retry, poison-pill ack — PartyEventConsumer pattern), idempotent upsert keyed on grant id; only `ACCOUNT`-scoped events, OFFERED/DECLINED never enforceable.
- **Guard**: `AuthorizationService.isAuthorized` gains the third disjunct via the projection (owner and legacy paths untouched, delegation only ADDS access); new `isAuthorizedForAmount` enforces per-transaction ceilings + currency match for PAYMENT_ONLY. Capabilities stored as strings — unknown future capabilities can't poison the consumer.
- **rules.yaml**: `openbank.delegation.events` allowlist entry removed (AC13) + OPA bundles regenerated.
- **Threat model**: delegated-actor section (AC15). **version.txt 0.16.0 → 0.17.0** (feat).

## Definition of Done

- [x] Hexagonal layering preserved (domain has zero framework imports).
- [x] Unit tests added/updated.
- [x] Integration tests added/updated (if service boundary involved).
- [ ] Contract tests updated (if public API changed).
- [x] `./gradlew detekt ktlintCheck koverVerify build` passes.
- [x] No new lint warnings.
- [ ] OpenAPI spec regenerated (if REST API changed).
[...]

Issue #2990, opened 2026-08-01
Product-service enforcement projections for delegation grants (ADR-0232 D3)


JiRaska:
## Problem

ADR-0232 (#2960) + delegation-service slice 1 (#2971) ship the grant system of record and the `openbank.delegation.events` stream, but **nothing enforces the grants yet**: a delegate with `ACCOUNT_READ_BALANCES` on someone else's account still gets 403 from account-service, because product services only check ownership. Until this lands, delegated access is write-only.

## Proposed solution

Per ADR-0232 D3 — decentralized, event-fed, never synchronous on the money path:

1. **account-service**: consume `openbank.delegation.events` into a local `delegation_projection` table (grantee, resourceType, resourceId, capabilities, limits, validTo, status). Extend `AuthorizationService.isAuthorized` (and the balance/transaction read guards) to `owner OR active-grant(caller, resource, capability, amount)` — the shape already exists (`isAuthorized` today is exactly `account.partyId == partyId || active role match`); this swaps the source, not the check.
2. **card-issuance-service**: same projection + guard for `CARD_VIEW` / `CARD_MANAGE_LIMITS`.
3. **savings goals** (account-service, ADR-0153): `SAVINGS_DEPOSIT` / `SAVINGS_WITHDRAW` / `SAVINGS_PROPOSE_WITHDRAW` handling; propose-only variants land in the owner's approval inbox (ADR-0227).
4. Delegated money-path actions thread `onBehalfOf: grantorPartyId` into the audit chain (ADR-0086/0133) — the customer-edge audit publisher already carries `partyId` (actor) + flat `details`; add `details.onBehalfOf` + `details.delegationId`.
5. Cumulative daily/monthly spend counters live in the enforcing service (the grant carries ceilings, the resource owner counts spend).
6. **Remove** the `openbank.delegation.events` entry from `rules.yaml: event_consumer_liveness.allowlist` in the first consumer PR (the allowlist has an explicit re-triage condition).

### Projection (account-service, card-issuance)

- **AC1** Consumer applies `DelegationActivated` into the projection idempotently (same event redelivered → same row, no duplicate); ordering per aggregate is guaranteed by the outbox partition key (= aggregateId). `DelegationOffered`/`Declined` never create an enforceable row.
[...]

PR #2971, opened 2026-07-31
feat(delegation): delegation-service — customer-to-party access grants (ADR-0232)

JiRaska:
## Summary

First slice of ADR-0232 (#2960): new **openbank-delegation-service** owning the `DelegationGrant` aggregate — customer-to-party, SCA-bound, granular access grants over products (accounts, savings goals, cards) and single objects (payment, statement, document). Full lifecycle OFFERED → ACTIVE → SUSPENDED/REVOKED/EXPIRED/RENOUNCED/DECLINED with a transactional outbox feeding `openbank.delegation.events` for the product-service enforcement projections (D3).

## Type of change

- [x] feat (new functionality)
- [ ] fix (bug fix)
- [ ] refactor (no behavior change)
- [ ] perf (performance)
- [ ] docs
- [ ] chore (build / tooling)
- [ ] security (private until disclosed)
- [ ] breaking change

## Linked issues

Refs #2960 — implements ADR-0232 (delegated access). No tracking issue yet; follow-up slices (product-service projections, customer-edge UX, external disclosure links, AccountAuthorization migration) get their own issues.

## Changes

- **Domain** (`DelegationGrant`): capability matrix per resource type, approvalPolicy (SOLO/ANY_ONE/ALL/N_OF_M), per-tx/daily/monthly ceilings, Exposure (redaction/maxViews/watermark/download) restricted to object-level grants, execution capabilities structurally forbidden on objects, lifecycle transitions with closed-by/reason audit.
- **Application**: offer (grantor SCA `DELEGATION_GRANT` + fail-closed pid-service eligibility: ACTIVE party, KYC FULL for execution / BASIC for read-only per D5), accept (grantee SCA `DELEGATION_ACCEPT`), decline, renounce, revoke (unilateral, D4), suspend/reinstate, list by grantor/grantee, `/check` returning a decision — never the grant.
- **Infrastructure**: Postgres + Flyway V1 (`delegation_grants`, `delegation_capabilities`, `delegation_redaction_rules`, `delegation_outbox`), atomic `FOR UPDATE SKIP LOCKED` outbox claim (#1201 pattern), scheduled dispatcher + hourly expiry sweep, Kafka publisher, SCA + pid resilient REST clients, Redis idempotency on POST, OPA sidecar PDP, dedicated exception mappers.
[...]

PR #3105, opened 2026-08-01
feat(card-issuance): delegation-grant enforcement projection for cards (ADR-0232 D3)

JiRaska:
## Summary

Card-side twin of #3058 (issue #2990, stacked on #2971): card-issuance consumes `openbank.delegation.events` (CARD-scoped) into `card_delegation_projection`, and a new `CardDelegationGuard` answers **holder OR an ACTIVE in-window delegation grant** for VIEW / MANAGE_LIMITS — exposed as `GET /api/v1/cards/{id}/delegation/check` for the customer-edge slice. Enforcement fully local; a tripwire test bans any synchronous delegation-service client.

**⚠️ Stacked PR: base is `feat/delegation-service` (#2971). Retarget to `main` once #2971 merges. Independent of #3058 (account slice) — either can merge first.**

## Type of change

- [ ] fix (bug fix)
- [ ] perf (performance)
- [ ] docs
- [ ] chore (build / tooling)
- [ ] breaking change

## Linked issues

Refs #2990 — card-issuance slice (projection + guard + check endpoint + tests). Remaining after this: savings goals, propose-only (AC8), onBehalfOf audit threading (AC9/AC10), Pact message contracts (AC12).

## Changes

- **Projection**: `card_delegation_projection` (+ caps, Flyway **V7**), `CardDelegationEventConsumer` on `delegation-events-in` (DLQ + bounded retry, poison-pill ack), idempotent upsert per grant id, OFFERED/DECLINED never enforceable, non-CARD events ignored.
- **Guard**: `CardDelegationGuard` — separate bean so the delegation blast radius stays one class wide; PAN/secure-details deliberately outside delegation scope (no capability maps to them).
- **API**: `GET /{id}/delegation/check?partyId&intent=VIEW|MANAGE_LIMITS` reusing the existing `card.read` OPA action (no rego change, mirrors consent-service's `/active` pattern).
- **Config**: `delegation-events-in` channel; group.id `cardissuance-delegation` via msg-override ConfigMap (dotted-key trap convention); KafkaUser ACLs (topic Read, group Read, DLQ Write) in `kafka-scheme-accepted-acl.yaml`.
- **Fleet-alignment fix**: `opa`/`authz` block added to application.yaml — the service had none, so the interceptor's enforce=true default would 503 every @Authorize endpoint when the sidecar is unreachable (this PR's REST IT is the first to exercise one).
- **version.txt 0.14.0 → 0.15.0** (feat). Threat model: delegation section.

## Definition of Done

- [x] No new lint warnings.

Local evidence: `detekt` ✓ `ktlintCheck` ✓ `test` ✓ (**CardDelegationProjectionIT**: real boot, seeded card via Panache, in-memory channel with `runOnVertxContext(true)` — activation → delegate VIEW=true, revocation → false; consumer 6 tests, guard 7 tests, dependency tripwire) `koverVerify` ✓ `quarkusBuild` ✓. V7 rollback: drop the two new tables. OpenAPI yaml: no dedicated spec file in this service (spec is generated); endpoint follows existing annotations.

## Security checklist

      (license, CVE, source).

**Money-path service: 2 approvals required** (card-issuance is in `rules.yaml: money_path_services`; threat model updated). PCI: secure-details/PAN explicitly out of delegation scope — noted in the threat model.

## Compliance impact

- [x] PCI DSS
- [ ] DORA
- [x] GDPR
- [ ] PSD2 / SCA / RTS
- [ ] AML / 5AMLD
- [ ] CNB reporting
- [ ] None

PCI: the change touches card metadata access control — PAN/CVV remain unreachable via delegation by construction (no capability maps to secure-details). GDPR: delegate access is purpose-bound by the grant; revocation propagates in seconds.

## Rollout / Rollback

- Deployment strategy: rolling, after #2971's delegation-service is deployed (topic + ACLs included here; ConfigMap + KafkaUser are sync-wave ordered already).
- Rollback plan: revert the squash; drop `card_delegation_projection*` (additive, no data migration).
- Data migration reversible: yes (drop new tables)

N/A

## Reviewer notes

- The check endpoint is the customer-edge integration point: customer-edge asks here until it holds its own projection — same transitional shape as account's `/check`.
- The `opa`/`authz` application.yaml addition is a latent-fleet-default fix, not a behavior change in deployment (gitops sets AUTHZ_ENFORCE explicitly); it only aligns local/test defaults with the rest of the fleet.
- `CardDelegationGuard` is deliberately not wired into `CardService` mutations in this slice — limit-change endpoints stay operator-only until the customer-edge slice lands the caller-party context.


JiRaska:
## The card projection is the one sibling that still trusts the grant row

Reviewed this against #3143 and #3164, which fix the same defect class elsewhere in the ADR-0232
program. The card side did not get the same treatment.

`CardDelegationGuard.isAuthorized` is holder OR **any** active matching grant:

```kotlin
val card = cardRepository.findById(cardId) ?: return false
if (card.partyId == partyId) return true
return projectionRepository.findActiveByCardAndParty(cardId, partyId)
    .any { it.isActiveOn(now) && it.satisfies(intent) }
```

`CardDelegationProjectionEntity` has `card_id`, `grantee_party_id`, capabilities and a window —
**no grantor column at all**, and `CardDelegationEventConsumer` never reads `grantorPartyId` off the
event. So the guard cannot check who issued the grant even if it wanted to: an ACTIVE row for
`(cardId, partyId)` is authority in itself.

That is defect #2 of the P0 chain on #3164, verbatim. Compare #3143, which added exactly this
check to the account and savings paths and says why in the code:

```kotlin
.filter { it.issuedBy(account.partyId) && it.isActiveOn(now) && it.satisfies(role) }
// a grant only counts when the party who ISSUED it is that owner. Without this the
// disjunct made a grant row authority in itself
```

`issuedBy` is called in three places there. Here there is nothing equivalent to call.

## Is it exploitable?

Not once #3164 lands: it makes `delegation-service` verify the grantor owns the resource, so an
invalid grant row can never be minted, and trusting the row becomes sound. Validate-at-the-writer
is a legitimate design.

But it is the weaker of the two positions, and right now it rests entirely on a fix that is **not
merged and not absorbed into the base** — `git merge-base --is-ancestor fix/delegation-p0-authz
feat/delegation-service` → no. Two of the three consumers chose defence in depth anyway; this one
would be the single place where a future regression in the writer reaches money (a card
capability) with nothing local to stop it.

## Suggestion

Carry `grantorPartyId` on the event into the projection and add the `issuedBy(card.partyId)`
conjunct, mirroring #3143. It is a column, a consumer field and one `&&` — and it makes the
three projections consistent, which matters more than the individual check: a reviewer comparing
them should not have to work out why one of them is different.

Nothing else in this PR is blocking; CI is green.

JiRaska:
## The `gates (kotlin)` red is not this PR's — and it is the merge-order hazard becoming mechanical

#2976 merged at 04:24 today, which put the **widened** `@Scheduled` guard on `main`: it now checks
the property (a scheduled method that starts reactive work must be `suspend` or `Uni`-returning)
rather than the `runBlocking` spelling. Reproduced by merging this PR's base into a local worktree,
which is what CI actually builds:

```
::error openbank-delegation-service/.../DelegationExpirationJob.kt:37
  the @Scheduled method `sweepExpiredGrants` is neither `suspend` nor `Uni`-returning,
  yet starts reactive work … the job silently does nothing
```

That method is in **#2971**, this PR's base — a fourth instance of the HR000068 class after the
three #2976 fixed. #3164 already fixes it (`suspend fun`, with the reasoning in its KDoc); the base
does not have that fix.

So the ordering risk I flagged on #2971 is now enforced by a gate rather than by a reviewer
noticing: `feat/delegation-service` cannot go green until #3164 lands in it. That is a better place
for it to be.

Nothing to do in this branch for that red.

## What was mine

`ktlintTestSourceSetCheck` — the `grant()` fixture's parameter list fits on one line, so the
multi-line form is a `function-signature` violation. Fixed.

`Customer-app dossier` is the fleet-wide one from openbank-app 0.12.0 (#2101), also not this PR's.

JiRaska:
## The `event-contract coverage ratchet` red is correct — and this is the first case it was built for

`gates (data)` fails with:

```
openbank-delegation-service produces to 'openbank.delegation.events' with no event contract.
1 new undocumented producer topic(s). ADR-0006 is a ratchet, not a wish: what already exists is
grandfathered in .github/event-contract-baseline.txt, and nothing new joins it. Adding your topic
to the baseline is not the fix — the baseline only shrinks.
```

That gate is #3019, merged to `main` after this branch was written, and `openbank.delegation.events`
is the **first new producer topic since**. So the red is the ratchet doing its job on its first real
case, not a collision to be worked around.

**I have not baselined it.** Silently adding the entry is precisely what the gate exists to prevent,
and doing it during a conflict cleanup would be the worst possible way to spend the repo's first
exception.

## What the fix needs

`openbank-contracts/openbank-delegation-service/asyncapi.yaml`, declaring the channel, message
schema, producer, consumers, retention and partitioning key. The material is already in the tree —
`DelegationEvents.kt` gives the shape, and every event carries `aggregateType`, `eventType` and an
explicit `version = 1L`:

```kotlin
data class DelegationOffered(
    override val aggregateId: UUID,
    val grantorPartyId: UUID,
    val granteePartyId: UUID,
    val resourceType: DelegationResourceType,
    val resourceId: UUID,
    val capabilities: Set<DelegationCapability>,
    override val occurredAt: Instant,
) { override val eventType = "DelegationOffered"; override val version = 1L }
```

## Why I stopped here rather than writing it

This would be the **first** contract in the repo — the ratchet reports zero today — so it sets the
template for the whole ADR-0006 migration, and #1916 owns that order ("money-path topics first,
BACKWARD compatibility at the registry, producers before consumers"). Delegation is money-path, so
it is a reasonable first, but choosing the shape of the first contract is a design decision, not
conflict cleanup, and #3019 said explicitly that closing this debt is a multi-PR programme it does
not attempt.

Happy to write it as its own PR against #1916 if that is the call. Flagging rather than deciding.

The same red will appear on #3105 and #3143 — all three now carry delegation-service since they
were retargeted to `main`.

Issue #1916, opened 2026-07-23
Complete ADR-0006: schema registry + Avro — Kafka events are raw unversioned JSON, zero .avsc fleet-wide


JiRaska:
ADR-0006 (Apicurio + Avro) is `partial`: **zero `.avsc` files exist fleet-wide**. Kafka events are raw unversioned JSON; compatibility is a build-time Python diff of Kotlin data-class constructors (`check-event-schema-compat.py` says so itself), and nothing validates payloads at runtime. For a bank this is a replay/interop hazard (audit §2.1 ranks it the #1 architecture gap after DR).

Scope (suggested order):
1. Deploy the schema registry (Apicurio) in-cluster
2. Start with the money-path topics (payments, ledger, transaction) — schema-first, BACKWARD compatibility enforced at registry level
3. Producer-side serialization + registry check in CI, consumer rollout after
4. Retire the constructor-diff script for migrated topics once registry compat replaces it

Two independent axes to preserve: event schema version vs release version (same discipline as ADR-0048 for REST).

JiRaska:
First triage pass — **still valid, but scope item 1 is already done** and the issue text is stale on it.

- **Apicurio IS deployed.** `openbank-infra/gitops/apps/apicurio.yaml` + `components/apicurio/{namespace,registry,postgres,network-policies}.yaml`. There is even a consumer: `openbank-analytics-sink/.../schema/ApicurioSchemaCatalogSource.kt` (with an IT). So "deploy the schema registry in-cluster" is not remaining work.
- **The `.avsc` claim holds.** `git ls-tree -r origin/main | grep -c '\.avsc$'` → **0**. Zero Avro schemas fleet-wide; producers still emit hand-built JSON, and `check-event-schema-compat.py` (Kotlin constructor diff) is still the only compatibility signal.

Revised remaining scope: items 2-4 only — schema-first `.avsc` for the money-path topics, registry-enforced BACKWARD compat, producer serialization + CI check, then retire the constructor diff per migrated topic.

Engineering-ready but large and fleet-wide. Recommend a narrow first slice: **one** money-path topic end-to-end (producer + consumer + registry compat + CI) to prove the pattern before any sweep — and note that migrating a topic's wire format is a rolling-deploy compatibility exercise on live money-path events, so it wants an ADR-0006 delivery note update rather than an ad-hoc PR.

_Automated triage pass, verified against `origin/main` 2026-07-26._

JiRaska:
[...]

PR #3143, opened 2026-08-01
feat(account): savings-goal delegation + propose-only withdrawal flow (ADR-0232 D3/D8)

JiRaska:
## Summary

Savings slices of #2990 (stacked on #3058), in two layers:

1. **SAVINGS_GOAL projection + guard (D3)**: SAVINGS_GOAL-scoped grants join the account-service delegation projection with a `resource_type` discriminator (Flyway V17), and `SavingsGoalDelegationGuard` answers **owner OR an ACTIVE in-window grant** for DEPOSIT / WITHDRAW / PROPOSE_WITHDRAW.
2. **Propose-only withdrawal flow (D8 / AC8)**: a delegate holding only `SAVINGS_PROPOSE_WITHDRAW` creates a `WithdrawalProposal` (V18) paired with an ADR-0155 `PendingApproval`; the **owner's SCA-bound decision is the only path to APPROVED**, which emits `SavingsWithdrawalApproved` — the executable instruction for the payments path. The delegate can never decide (owner check + store-enforced segregation of duties) and never executes (no delegate execution endpoint exists).

**⚠️ Stacked PR chain: #2971 → #3058 → this PR. Retarget as the chain merges.**

Design note: savings goals are **account metadata** (columns on `accounts`, ADR-0153), so SAVINGS_GOAL grants key on the owning account id **by convention** and share `account_delegation_projection` with a type column keeping the guards apart.

## Type of change

- [ ] fix (bug fix)
- [ ] perf (performance)
- [ ] docs
- [ ] chore (build / tooling)
- [ ] breaking change

## Linked issues

Refs #2990 — savings-goals slice + AC8. Remaining: onBehalfOf audit threading on the payment execution path (AC9/AC10), Pact message contracts (AC12).

## Changes

- **Projection**: V17 `resource_type` (rollback = drop column); consumer projects ACCOUNT and SAVINGS_GOAL events; cross-type isolation proven in IT (a savings grant never satisfies account READ_ONLY on the same account).
- **Propose-only**: `WithdrawalProposal` aggregate + V18; `SavingsProposalService` (propose/decide/cancel/list); owner-only decide with party- and purpose-bound SCA (`SAVINGS_WITHDRAW_APPROVAL`); `ApprovalStore` (ADR-0155 Redis store) pairing; `SavingsWithdrawalApproved` outbox event on approval.
- **API**: `.../savings-goal/delegation/check` + `.../delegation/proposals` (+ `/decide`, DELETE) — reusing `account.read`/`account.authorize` OPA actions (no rego change); published in openapi **1.7.0**.
- **Infra**: sca-service REST client + config; `ApprovalConfig` (ApprovalStore producer).
- **version.txt 0.17.0 → 0.19.0** (two feat slices). Threat model: cross-type EoP + maker-checker-bypass sections.

## Definition of Done

- [x] No new lint warnings.

Local evidence: `detekt` ✓ `ktlintCheck` ✓ `test` ✓ (**SavingsGoalDelegationIT**, **SavingsProposalIT**: propose → owner-SCA approve → APPROVED E2E; negative: propose without grant → 403; unit: guard matrix, owner-check, store propagation, SCA purpose binding) `koverVerify` ✓ `quarkusBuild` ✓. Contract tests: no consumer of the new endpoints yet (customer-edge slice will add consumer pacts).

## Security checklist

      (license, CVE, source).

**Money-path service: 2 approvals required** (account-service; threat model updated).

## Compliance impact

- [ ] PCI DSS
- [ ] DORA
- [x] GDPR
- [x] PSD2 / SCA / RTS
- [ ] AML / 5AMLD
- [ ] CNB reporting
- [ ] None

Propose-only never moves money without the owner's own SCA (no new exemption surface); the decide ceremony is purpose- and party-bound; approval and executable instruction share one outbox transaction (no approved-but-uninstructed state).

## Rollout / Rollback

- Deployment strategy: rolling with #3058 (V17/V18 apply on top of its V16).
- Rollback plan: revert; V17 droppable column, V18 droppable table (both noted in migration headers).
- Data migration reversible: yes

N/A

## Reviewer notes

- The approval IS the execution trigger: there is deliberately NO delegate-facing execute endpoint; the money leg is the `SavingsWithdrawalApproved` event consumed by the payments path in the AC9/AC10 slice.
- `StubScaChallengeClient` lives in test sources only (sca-service is not part of the IT stack); its SCA answers mirror the production contract.
- The owner check runs before the store's self-approval guard — the store stays defense-in-depth; both paths are unit-tested.


JiRaska:
Brought this branch onto the grantor-ownership fix and closed the same hole in the savings guard.

**What landed here** (fast-forward, no force-push — the branch tip moved `477998c8f` → `5254944e1`, so a local checkout only needs a `pull`):

1. Merge of `fix/account-delegation-grantor-check` ([#3175](https://github.com/JiRaska/open-bank-oss/pull/3175)), which adds `grantor_party_id` to the projection and makes `AuthorizationService` require the grant's issuer to own the account. Clean merge, no conflicts.
2. `SavingsGoalDelegationGuard` gets the same check — and it matters more here. A `SAVINGS_GOAL` grant is keyed on the **owning account's id** (a savings goal is account metadata, ADR-0153), so a grant naming a stranger's account reached this guard exactly as it reached the account guard, and `SAVINGS_WITHDRAW` moves money.

**Why merge and not rebase:** a rebase means force-pushing a branch that has a live worktree on this machine, and `--force-with-lease` does not protect against that. The merge is non-destructive and is the repo's documented pattern for a child PR that needs an unmerged parent's files. Those files drop out of this PR's diff once #3175 lands on `feat/account-delegation-projection`.

**Migrations are unaffected:** #3175 edits `V16` in place (never applied anywhere, and a new `V18` would have made Flyway refuse this PR's `V17` as out of order). `V17` and `V18` here are untouched.

**Verified locally:** `:openbank-account-service:detekt` plus the delegation and savings suites — 37 tests, 0 failures, including `SavingsGoalDelegationIT`, `DelegationProjectionIT` and the two new savings cases (a non-owner's grant is denied for DEPOSIT and WITHDRAW; the identical grant from the owner still passes, which is what shows the check discriminates rather than just denying).

Threat model updated: the savings entry noted that `SAVINGS_GOAL` grants key on the owning account id "by convention" without drawing the consequence.

JiRaska:
Merged the updated base in (only conflict was a hand-bumped `version.txt`, resolved the same way as
#3058 — release-please owns it and this branch has four `feat` commits).

## The propose-only flow is genuinely enforced — except for who the caller is

The maker/checker claim holds where it matters:

```kotlin
if (account.partyId != decidedByPartyId) throw ProposalForbiddenException(...)   // owner only
verifyOwnerSca(decidedByPartyId, scaSessionId)                                    // SCA bound
approvalStore.decide(approvalId, decidedByPartyId.toString(), approve)           // ADR-0155
```

and `verifyOwnerSca` is a real barrier, not a formality — it requires a COMPLETED challenge whose
`partyId` is the owner and whose `purpose` is `SAVINGS_WITHDRAW_APPROVAL`. A delegate cannot decide
their own proposal.

**But the identity being checked comes from the request:**

```kotlin
suspend fun decide(
    @PathParam("accountId") accountId: UUID,
    @PathParam("proposalId") proposalId: UUID,
    @QueryParam("decidedByPartyId") decidedByPartyId: UUID,   // ← caller-supplied
```

`propose` likewise takes `request.delegatePartyId` from the body, and the resource reads no caller
identity at all. So `account.partyId != decidedByPartyId` compares DB state against a value the
caller chose. That is defect C3 of the #3164 P0 chain, in a second service.

**Not remotely exploitable as it stands** — the SCA check is what stops it, since an attacker would
need a completed challenge of the owner's for this exact purpose. The problem is that SCA is then
doing double duty as authentication, and it is the only thing left if that check is ever relaxed,
mocked in a test profile, or reordered.

## The fix is two files away, in this same service

`AccountResource` already does it, and documents the rule:

```kotlin
const val CUSTOMER_PARTY_HEADER = "X-Customer-Party-Id"
// When that header is present the read MUST belong to that party.
…
@HeaderParam(CUSTOMER_PARTY_HEADER) customerPartyId: UUID?,
```

#3164 resolves the caller the same way in delegation-service. So this is not "adopt a new pattern"
— it is a divergence from the pattern this service uses in a sibling file. Taking `decidedByPartyId`
and `delegatePartyId` from `CUSTOMER_PARTY_HEADER` instead of the query/body makes the three
delegation surfaces consistent and removes the caller's ability to name themselves.

I have not made that change: it alters the endpoint contract (and `openapi.yaml`), which belongs to
this PR's author rather than to a conflict merge.

Everything else here is solid — and `issuedBy(account.partyId)` in the guard, wired into three
authorization paths, is the check #3105 was missing entirely.

JiRaska:
## The `event-contract coverage ratchet` red is correct — and this is the first case it was built for

`gates (data)` fails with:

```
openbank-delegation-service produces to 'openbank.delegation.events' with no event contract.
1 new undocumented producer topic(s). ADR-0006 is a ratchet, not a wish: what already exists is
grandfathered in .github/event-contract-baseline.txt, and nothing new joins it. Adding your topic
to the baseline is not the fix — the baseline only shrinks.
```

That gate is #3019, merged to `main` after this branch was written, and `openbank.delegation.events`
is the **first new producer topic since**. So the red is the ratchet doing its job on its first real
case, not a collision to be worked around.

**I have not baselined it.** Silently adding the entry is precisely what the gate exists to prevent,
and doing it during a conflict cleanup would be the worst possible way to spend the repo's first
exception.

## What the fix needs

`openbank-contracts/openbank-delegation-service/asyncapi.yaml`, declaring the channel, message
schema, producer, consumers, retention and partitioning key. The material is already in the tree —
`DelegationEvents.kt` gives the shape, and every event carries `aggregateType`, `eventType` and an
explicit `version = 1L`:

```kotlin
data class DelegationOffered(
    override val aggregateId: UUID,
    val grantorPartyId: UUID,
    val granteePartyId: UUID,
    val resourceType: DelegationResourceType,
    val resourceId: UUID,
    val capabilities: Set<DelegationCapability>,
    override val occurredAt: Instant,
) { override val eventType = "DelegationOffered"; override val version = 1L }
```

## Why I stopped here rather than writing it

This would be the **first** contract in the repo — the ratchet reports zero today — so it sets the
template for the whole ADR-0006 migration, and #1916 owns that order ("money-path topics first,
BACKWARD compatibility at the registry, producers before consumers"). Delegation is money-path, so
it is a reasonable first, but choosing the shape of the first contract is a design decision, not
conflict cleanup, and #3019 said explicitly that closing this debt is a multi-PR programme it does
not attempt.

Happy to write it as its own PR against #1916 if that is the call. Flagging rather than deciding.

The same red will appear on #3105 and #3143 — all three now carry delegation-service since they
were retargeted to `main`.

PR #3156, opened 2026-08-01
security(admin-ui): hold the public demo account to least privilege in each tool, and deny it Alertmanager

JiRaska:
Refs #3071

**Design changed after review feedback: the demo account is no longer denied outright.** It stays admitted to Grafana and Pyrra and is held to the least privilege each tool can express; only Alertmanager is denied, because it has no privilege level to hold it to.

## The exposure

`demo@openbank.local` is a **public** account — its credentials are handed out — and in the sandbox realm it carries:

```
ROLE_OPERATOR ROLE_AUDITOR ROLE_COMPLIANCE ROLE_ADMIN ROLE_DEMO ROLE_PAYMENTS ROLE_VIEWER
```

Those grants exist so the demo can show every console page, and they were harmless while the tools had no route. **ADR-0234 gave them one.** The gate's permission check alone (`system:view` = ADMIN or OPERATOR) then admits the public account to all three — and Grafana maps `ROLE_ADMIN` → Admin, which grants Explore against Prometheus, **Loki** and Tempo, i.e. the raw log stream.

This is a regression I introduced by routing the tools: the roles were already there, the reachability is new.

## What this does instead

| tool | demo | why |
|---|---|---|
| Grafana | **admitted as Viewer** | Viewer sees dashboards but not Explore, so no raw Loki/Tempo |
| Pyrra | **admitted** | read-only by construction — nothing to restrict |
| Alertmanager | **denied** | no role model at all |

Alertmanager is the exception because its UI *is* its API: anything that can load the page can silence or expire an alert platform-wide. There is no read-only Alertmanager to offer, so the choice is "can mute production alerting" or "not admitted" — and for a public account that is not close.

A greyed-out console is a worse outcome than a read-only one — the point of the demo account is that a visitor sees a working platform — so the Sidebar renders a denied tool as a **disabled entry with the demo-account tooltip**, not a live link that 403s. A visibly failing link reads as a broken product.

## Mechanism

**The deny-list is keyed by tool, not global.** Adding a tool then forces the question "what is the least privilege *this one* can express?" instead of silently inheriting an answer. It runs *before* the permission check, because the denied role travels alongside the roles that would pass.

**Grafana's `role_attribute_path` tests `ROLE_DEMO` first.** jmespath `||` short-circuits, so the ORDER is the entire mechanism — placed after the ADMIN test it would never fire. This is the second, independent layer: it holds even if the gate list is later widened.

## Verification

Falsified in both directions, not just run:

- adding `grafana` to the deny-list reddens the *"still admits the demo account"* assertion — so the allow is asserted, not assumed;
- removing the `ROLE_DEMO` branch from `role_attribute_path` reddens the ordering assertion.

24 tests pass; eslint clean. `tsc --noEmit` is extremely slow in this worktree and had not finished at push time — the changes are a `Record` type, one optional field and a boolean expression, and the suite compiles the route and Sidebar through the vitest transform, so the risk is low; I am watching CI's type-check rather than claiming a local pass.

## A correction that belongs here

Applying the Grafana half as an interim mitigation kept silently reverting, and the cause invalidates something I have repeated:

```
$ kubectl -n argocd get app root
path=openbank-infra/gitops/apps  auto={"prune":true,"selfHeal":true}
```

**There IS an app-of-apps syncing `gitops/apps/`.** The comments in `apps/glitchtip.yaml` and `apps/loki.yaml` saying no app-of-apps exists yet are stale, and I repeated the claim in ADR-0234, in the `tool-ingress-gate-policy.yaml` header, in #3072's post-merge runbook and in several PR bodies. The manual applies I ran earlier only appeared to work because those changes were already merged — `root` would have synced them anyway.

Consequences: the "run `kubectl apply -f gitops/apps/…` after merge" steps are unnecessary; an unmerged change applied by hand is reverted within seconds; and therefore **no interim mitigation outside git is available here — merging is the mitigation**. Follow-up sweep to come, kept out of this PR to keep it focused.

### DIFF
--- a/openbank-admin-ui/src/app/api/gate/route.ts
+++ b/openbank-admin-ui/src/app/api/gate/route.ts
@@ -57,6 +57,43 @@ const TOOL_PERMISSIONS: Record<string, Permission> = {
   pyrra: "system:view",
 }
 
+/**
+ * PER-TOOL deny-list: roles refused for a specific tool, whatever else they carry.
+ *
+ * `demo@openbank.local` is a PUBLIC account — its credentials are handed out —
+ * and in the sandbox realm it holds ROLE_ADMIN, ROLE_OPERATOR, ROLE_COMPLIANCE,
+ * ROLE_AUDITOR, ROLE_PAYMENTS and ROLE_VIEWER alongside ROLE_DEMO, so that it can
+ * show every console page. Those roles were harmless while the tools had no route;
+ * ADR-0234 gave them one, and the permission check alone then admits the public
+ * account everywhere.
+ *
+ * The demo account is DELIBERATELY still admitted to Grafana and Pyrra, because a
+ * greyed-out console is a worse outcome than a read-only one — the point of the
+ * account is that a visitor sees a working platform. It is held to the least
+ * privilege each tool can express instead:
+ *
+ *   - Grafana: `role_attribute_path` (kube-prometheus-stack) tests ROLE_DEMO FIRST
+ *     and maps it to Viewer. Viewer sees dashboards; it does not get Explore, which
+ *     is what would otherwise expose the raw Prometheus/Loki/Tempo streams.
+ *   - Pyrra: read-only by construction — there is nothing to restrict.
+ *   - Alertmanager: DENIED, and it is the exception because it has no role model at
+ *     all. Its UI is its API: anything that can load the page can silence or expire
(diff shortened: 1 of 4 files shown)


JiRaska:
Follow-up on the verification note: `tsc --noEmit` has since finished locally with no output, i.e. no type errors. The hedge in the PR body ("had not finished at push time") is now stale — the local type-check passes.

PR #3190, opened 2026-08-01
feat(admin-ui): a four-eyes console for credit compliance pack activation

JiRaska:
## What

A four-eyes activation console for the ADR-0212 jurisdictional credit compliance packs, at
`/lending/compliance-packs` (sidebar: Compliance → *Credit Compliance Packs*, `compliance:view`).

It shows the **active packs**, the **proposals awaiting a checker** (approve / reject with a
reason), and a **maker form** that takes pack JSON.

## Why

ADR-0212 D4 made pack activation a maker-checker control, but the only way to exercise it was
hand-driving `POST /api/v1/lending/compliance-packs/proposals` and `.../decide` with two
separately-minted operator tokens. A control that needs a shell is a control nobody runs.

That matters right now: `openbank.lending.compliance.enforce-pack` cannot be turned on until a pack
is active, because with enforcement on and no active pack **every origination is refused**. The gate
and the only route through it should not live in different places.

## Three things it deliberately does not do

- **It does not re-implement `maker != checker`.** lending-service raises `MakerCheckerViolation`
  (422) and the page renders that message verbatim. A client-side copy of the rule would put a
  weaker check in front of the real one.
- **It does not render a refused read as an empty queue.** A 403 on the pending list is the ordinary
  outcome for a non-compliance operator, and "nothing pending" is the most dangerous thing an
  approvals screen can say wrongly — same reasoning as ADR-0227 D2's `/approvals`.
- **It does not treat an empty active list as unremarkable.** That is precisely the state in which
  `LENDING_ENFORCE_PACK` must not be flipped, so the screen says so in place.

Each of the three is a test.

## Falsification

The suite was run against a deliberately broken page before it was trusted:

- flattening the 422 to a generic `Decision failed` → `renders the maker-checker refusal verbatim` red
- forcing the pending source to `'ok'` regardless of status → `a refused pending read never renders as "nothing pending"` red

`2 failed | 4 passed`, then green again on restore.

## Verification

- `npm test` — 62 files, **942 tests passed** (includes the 6 new ones)
- `npx tsc --noEmit` — clean
- `npm run lint` — 0 errors (the one warning on the new file is the fleet-wide
  `set-state-in-effect` pattern every console page already uses, incl. `/lending`)

No API, DB, event or config change; admin-ui only.

Refs #3168

### DIFF
--- a/openbank-admin-ui/src/app/lending/compliance-packs/page.tsx
+++ b/openbank-admin-ui/src/app/lending/compliance-packs/page.tsx
@@ -1,0 +1,339 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) OpenBank contributors. Licensed under the Apache License, Version 2.0.
+// See LICENSE in the repository root or https://www.apache.org/licenses/LICENSE-2.0 for details.
+
+// ADR-0212 D4 — the four-eyes activation console for jurisdictional credit compliance packs.
+//
+// Why this screen exists: activating a pack was reachable only by hand-driving
+// POST /compliance-packs/proposals and .../decide with two separately-minted operator tokens.
+// A control that needs a shell is a control nobody exercises — and
+// `openbank.lending.compliance.enforce-pack` must stay false until a pack is active, because with
+// enforcement on and no active pack every origination is REFUSED. The gate and the only way
+// through it belong in the same place.
+//
+// Maker != checker is NOT enforced here. The service raises MakerCheckerViolation (422) when one
+// principal tries both halves, and this page renders that refusal verbatim. A client-side copy of
+// the rule would only hide the real control behind a weaker one.
+
+'use client'
+
+import { useCallback, useEffect, useMemo, useState } from 'react'
+import { useSession } from 'next-auth/react'
+import { CheckCircle2, XCircle, Clock, RefreshCw, ShieldCheck, ScrollText, AlertTriangle } from 'lucide-react'
+import { useLanguage } from '@/lib/i18n/LanguageContext'
+import { svcUrl } from '@/lib/services/bff'
+
+/** Mirrors lending-service `PackActivationView`. `listActive()` synthesises id = all-zero UUID for
+ *  every row (it projects the in-memory registry, not a workflow row) — never key a list on it. */
+type PackActivationView = {
+  id: string
+  jurisdiction: string
+  productType: string
+  packVersion: number
+  effectiveFrom: string
+  contentHash: string
+  state: string
+  proposedBy: string
+  decidedBy: string | null
+  decidedAt: string | null
+}
+
+/** A read that was REFUSED must never render as "nothing pending" — same reasoning as /approvals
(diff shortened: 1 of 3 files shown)


JiRaska:
## Merging — the four-eyes enforcement is already on `main`, this is the view over it

A console that POSTs to `/api/v1/lending/compliance-packs`, where an activated pack is used by the lending origination guard immediately, is a control surface over a money-path service. So I checked what this PR actually grants before merging it.

### It adds no backend

```
lending files in this PR: 0
```

`CompliancePackResource` already exists on `main`, and it is where the four-eyes rule lives — `propose` and `decide` are separate endpoints, each with its own OPA action:

```
@RolesAllowed("ROLE_COMPLIANCE", "ROLE_ADMIN")          <- class level
  POST /proposals              @Authorize lending.compliance.propose
  POST /proposals/{id}/decide  @Authorize lending.compliance.decide
  GET  /proposals/pending      @Authorize lending.compliance.read
  GET  /active                 ROLE_COMPLIANCE|ROLE_CREDIT_RISK|ROLE_LENDING_OFFICER|ROLE_ADMIN
```

So the separation of duties is enforced server-side and is not something this PR can weaken. The console surfaces an existing, role-gated, OPA-checked capability.

### Scope

3 files, `+494/-0`, no deletions: one page, a sidebar entry, and a test. `Admin UI` and `Admin UI build` both green, all five required contexts green.

### Non-required red

`agent-review` is failing — that is #2161 (413 on generated bundles), unrelated to this PR.

*Reviewed while draining the queue overnight.*

JiRaska:
## What is actually blocking this — one unused import

Auto-merge is armed and all five required contexts are green, but `mergeStateStatus` is `BLOCKED`. The cause is not a check:

```
required_review_thread_resolution : on   (main-protection ruleset)
unresolved review threads         : 1
```

The single thread is CodeQL on `openbank-admin-ui/src/app/lending/compliance-packs/page.tsx:22`:

> Unused import `XCircle`.

Confirmed — `XCircle` appears exactly once in the whole diff, on the import line itself:

```
import { CheckCircle2, XCircle, Clock, RefreshCw, ShieldCheck, ScrollText, AlertTriangle } from 'lucide-react'
```

Drop `XCircle` from that import and the thread resolves; auto-merge is already armed, so it will land on its own.

## Why I did not just push the fix

There is a live worktree on `feat/admin-ui-compliance-packs` — last commit 23:07, ~69k files touched in the past hour (a build running). Pushing would stomp work in progress, so this is yours.

## The review itself is done

I checked the substance before arming auto-merge: this adds **zero** backend files. `CompliancePackResource` is already on `main` and is where the four-eyes rule lives — `propose` and `decide` are separate endpoints, class-level `@RolesAllowed("ROLE_COMPLIANCE","ROLE_ADMIN")`, each with its own `@Authorize` OPA action. The console surfaces an existing, gated capability rather than granting one.

`agent-review` is red for #2161 (413 on generated bundles), unrelated.

*Overnight queue drain.*

Issue #2895, opened 2026-07-31
Campaign read-only console in admin-ui — business has no way to see what campaigns are doing


JiRaska:
## Problem

There is no campaign UI at all. `openbank-admin-ui` contains zero campaign pages, so the only way to
create, activate or inspect a campaign is the REST API with an M2M token — a developer tool, not
something marketing can use.

Campaign Studio (ADR-0221) is the planned answer, but it is `proposed/planned` and it is a whole
authoring wizard: versioned segment composition, template catalogue, live consent funnel, four-eyes
submission. That is a large piece of work, and it is also **blocked** — segments cannot be evaluated
at all today (#2891), so an authoring UI would compose artifacts that produce empty cohorts.

## Proposal: ship the read-only half first

A console that only *shows* what campaigns are doing. No authoring, no templates, no approval UX.

- Campaign list: name, goal, state (DRAFT/PENDING/ACTIVE/PAUSED/CLOSED), segment ref, created-by,
  approved-by.
- Campaign detail: the steps, and the enrolment roll-up per state.
- Send log per campaign: outcome per party per step — crucially **including the suppression reason**,
  so `SUPPRESSED_CONSENT` and the ADR-0200 D6 frequency/quiet-hours suppressions are visible rather
  than inferred from a zero.

## Why this slice

- It is the part business actually asks for first — "what is running, who got what, and why did
  someone not get it".
- It does not depend on #2891. Read views work the moment any campaign exists, and they are how you
  would *observe* the segment fix landing.
- It sidesteps the four-eyes problem. `submit` → `activate`-by-a-different-approver is designed for
  two people at a screen; exposing it through curl loses the point of ADR-0200 while adding no
  safety. Read-only has no such tension.
- It is the smallest thing that makes the consent gate legible. Today a suppressed send is invisible
  outside the service's own database.

## What it needs from the backend

Most of it already exists: `GET /api/v1/campaigns`, `GET /api/v1/campaigns/{id}`, and
`GET /api/v1/campaigns/{id}/enrolments`.

The gap is the send log — the `send_log` table (campaign_id, party_id, step_order, outcome,
occurred_at) has **no REST endpoint**. That is the one backend addition this needs, and it is the
most valuable column of the lot because `outcome` is where consent suppression shows up.

PII note worth settling in the design, not after: the send log is keyed by `party_id`, so any view
of it is a view of who was marketed to. It should show party identifiers rather than contact details,
[...]

PR #3191, opened 2026-08-01
feat(campaign): show the journey as a funnel, not a table of enums

JiRaska:
Refs #2895. `openapi.yaml` 1.5.0 → 1.6.0.

Direct feedback from the person this console is for: it is too technical to be usable for marketing.
That is fair, and the Steps section was the clearest example — three columns of order / template id /
delay-in-minutes, showing the campaign's **definition**. To learn whether the campaign was *working*
you had to correlate it against the send log by eye, translating `SUPPRESSED_QUIET_HOURS` as you
went.

A marketer's question is never about one row. It is "how many, and why not more" — which is a shape,
not a table: flow and proportion.

## What it looks like now

Each step is a card carrying the step's own funnel: how many it reached, how many were delivered, and
the rest segmented by reason — with the wait between steps drawn as part of the journey rather than
listed in a column.

For a step that reached 1 000 and delivered 600, the bar reads at a glance as 60 % green, 30 % blue
(no consent), 9 % grey (quiet hours), 1 % red (failed), with the same breakdown as labelled chips
underneath.

## Three rules it follows

Each is something the send-log table breaks on purpose, and each is pinned by a test.

- **No enum reaches the screen.** `SUPPRESSED_QUIET_HOURS` is a value the API returns; what a
  marketer needs is "it was night". The raw value stays in `title`, so the screen and the API cannot
  drift apart without someone noticing — relabelling is a translation, not a replacement.
- **Every number is a SQL aggregate**, grouped by `(step, outcome)` in a new
  `GET /campaigns/{id}/journey`. A funnel is read as the whole picture by definition, so one folded
  from a loaded page would understate every campaign larger than that page while looking exactly as
  authoritative. This is the same mistake #3111 fixed for the suppressed headline.
- **A drop-off is not a failure.** Consent withheld, quiet hours and the frequency cap are the
  platform doing its job. Colouring them red teaches people to ignore red — and the one that *is* a
  failure then reads like the others.

## A design-system addition, not a local colour

The bar does **not** reuse `.badge-*`. Those are pale backgrounds tuned so dark text sits on them
legibly, which makes them the wrong fill for a 14px bar carrying no text: at that size the success
and info pastels are barely distinguishable, and each segment's 1px border reads as noise. I verified
this by rendering the component and looking at it — the first version was genuinely hard to read.

`globals.css` gains a `.funnel-bar` / `.funnel-seg-*` vocabulary built from the **same** semantic
variables at full saturation. Meaning stays in one place, and colour still never leaves the
stylesheet (ADR-0208 D2).

## Smaller thing the tests caught

A step that has reached nobody now says so instead of printing "0 delivered of 0". On a journey whose
later steps are still waiting out their delay that is most of the screen, and three zeroes look like
a result rather than the absence of one.

## Verification

Rendered the component to static HTML with the **compiled** stylesheet and inspected it in a browser
— the first attempt, using only `globals.css`, was misleading because the layout depends on utility
classes generated at build time. Worth stating: a screenshot of a partial stylesheet is not a
screenshot of the page.

Suites green locally: `ktlintCheck detekt test` on campaign-service, 940/940 on admin-ui (7 of them
new, covering the three rules above plus the empty-step and delay-formatting cases).

## Not in this PR

The audience number is the enrolment count, labelled as pre-consent and pre-suppression so it cannot
be read as reach. The full ADR-0221 D1 consent funnel (segment → after consent → after caps → after
suppression) needs counts the engine does not expose yet.

### DIFF
--- a/openbank-admin-ui/src/app/api/campaigns/[id]/journey/route.ts
+++ b/openbank-admin-ui/src/app/api/campaigns/[id]/journey/route.ts
@@ -1,0 +1,47 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) OpenBank contributors. Licensed under the Apache License, Version 2.0.
+// See LICENSE in the repository root or https://www.apache.org/licenses/LICENSE-2.0 for details.
+
+// Per-step funnel for the journey view. Every number is a SQL aggregate on the service side; this
+// route adds nothing and computes nothing, so the picture cannot disagree with the send log.
+
+import { NextResponse } from 'next/server'
+import { auth } from '@/auth'
+import { serverSvcUrl } from '@/lib/services/bff'
+
+export const dynamic = 'force-dynamic'
+
+export async function GET(_req: Request, ctx: { params: Promise<{ id: string }> }) {
+  const session = await auth()
+  if (!session?.user?.accessToken) {
+    return NextResponse.json({ error: 'unauthenticated' }, { status: 401 })
+  }
+  const { id } = await ctx.params
+  const path = `/api/v1/campaigns/${encodeURIComponent(id)}/journey`
+  try {
+    const res = await fetch(serverSvcUrl('campaign-service', 'campaign', 8128, path), {
+      headers: { authorization: `Bearer ${session.user.accessToken}` },
+      signal: AbortSignal.timeout(6000),
+      cache: 'no-store',
+    })
+    if (!res.ok) {
+      // An unreadable funnel must not arrive as an empty one: "nobody was reached" is a business
+      // conclusion a marketer would act on, and a 403 or a timeout is not that conclusion.
+      return NextResponse.json(
+        {
+          steps: [],
+          state:
+            res.status === 401 || res.status === 403
+              ? 'unauthorized'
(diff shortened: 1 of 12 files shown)


JiRaska:
## Merging — read-only, and the spec version is clean against live `main`

The campaign area has a known open gap (marketing email has no consent gate), so I checked what this actually adds. It is read-only: a single `@GET` send-log query behind `CampaignSendLogResource`, plus the funnel view. No send path, no consent decision, no migration.

### API contract (ADR-0048)

`openapi.yaml` changes by `+49/-1`, so the version axis matters. Read off **live `main`**, not this branch's base:

```
openbank-campaign-service  info.version   main: 1.5.0   branch: 1.6.0
open PRs touching this spec: this one only
gates (api): success
```

Minor bump for an additive endpoint, and nothing else is racing for the number — so the #3055 failure mode (a version consumed by an already-merged PR) does not apply here.

### Non-required red

`agent-review` is failing for #2161 (413 on generated bundles), unrelated. All five required contexts are green, so this goes in via `--squash --auto`.

*Overnight queue drain.*

PR #3192, opened 2026-08-01
fix(infra): give the five money-path services the JDBC datasource Flyway migrates

### LINKED ISSUE #3080: Four money-path services cannot boot from their committed configuration (no default JDBC datasource + Flyway migrate-at-start)
Four money-path services cannot start from the configuration committed to this repo. They boot in
production only because the gitops env supplies `QUARKUS_DATASOURCE_JDBC_URL`.

| Service | `quarkus.datasource.jdbc.url` | misplaced `quarkus.flyway.datasource:` block |
|---|---|---|
| `openbank-ledger-service` | absent | no |
| `openbank-account-service` | absent | yes |
| `openbank-sca-service` | absent | yes |
| `openbank-transaction-service` | absent | yes |

Each declares only a reactive datasource URL, so the default Agroal (JDBC) datasource is never
configured — while `quarkus.flyway.migrate-at-start: true` requires one. The boot dies with:

```
io.quarkus.runtime.configuration.ConfigurationException: Unable to find datasource '<default>'
  for persistence unit '<default>': Bean is not active: SYNTHETIC bean
  [class=io.agroal.api.AgroalDataSource]
```

Three of the four look like they intended to solve exactly this: they carry a
`quarkus.flyway.datasource:` block with `db-kind`, `username`, `password` and `jdbc.url` nested
underneath. Quarkus does not read that path as a datasource *definition* — `quarkus.flyway.datasource`
names which datasource Flyway should use. The boot error is the proof that it is ignored: it
reports the default datasource missing while that block sits in the same file. So the config reads
as deliberate and does nothing, which is worse than an omission — it answers the question for the
next reader, incorrectly.

**How this surfaced:** all four failed to boot in the authenticated fuzz harness (#3039). The
harness now passes the JDBC URL explicitly ([#3079](https://github.com/JiRaska/open-bank-oss/pull/3079)),
which unblocks the fuzz lane but leaves the services themselves unbootable outside a cluster —
including for a developer running `quarkusDev`.

**Why nothing caught it:** no gate boots a service (#2872). Unit tests mock the repository layer,
and the integration tests use their own `%test` profile with a real `jdbc:` URL — so the profile
that production-shaped config actually uses is the one nothing exercises.

Fix is per-service: give the default datasource a JDBC URL in `application.yaml` (matching the
reactive one), and delete the `quarkus.flyway.datasource:` block or correct it to the datasource
*name* it was meant to be.

Refs #3039, #2872, #3079

JiRaska:
Closes #3080 · Refs #3039, #2872, #3079

## The defect, confirmed on current `main`

Each of these declares only a **reactive** datasource URL, so the default Agroal (JDBC) datasource that `quarkus.flyway.migrate-at-start: true` requires is never configured:

| service | `quarkus.datasource.jdbc.url` | misplaced `quarkus.flyway.datasource:` mapping |
|---|---|---|
| ledger | absent | no |
| account | absent | **yes** |
| sca | absent | **yes** |
| transaction | absent | **yes** |
| **settlement** | absent | no |

**settlement-service is a fifth the issue does not name.** The sweep behind this change enumerated every `migrate-at-start` service in the fleet: 41 have a default JDBC URL, 5 do not. It is money-path too.

## Why the `quarkus.flyway.datasource:` blocks are deleted, not corrected

That key names **which existing datasource** Flyway should use; it does not define one, so the nested `db-kind`/`username`/`jdbc.url` are inert. The block reads as a solved problem and is not — which is worse than the plain omission, because it answers the question for the next reader incorrectly while the boot error names the missing default datasource in the same file.

Same class of thing in ledger: its comment said no such block was needed because *"its JDBC URL comes from the default datasource"* — true in the cluster, false in this file, which had no JDBC URL at all.

## Evidence — booted, not just parsed

`openbank-sca-service` built as a fast-jar and run with `-Dquarkus.profile=prod` and **no database**:

| build | observed |
|---|---|
| **with this fix** | boot proceeds to Flyway, which attempts a JDBC connection: `Connection refused` ×10 (its `connect-retries`), then fails for want of a database |
| **pre-fix** (rebuilt from the reverted config) | **no connection attempt at all** — zero `Connection refused` — straight to `Failed to start application` |

The discriminator is the presence of the JDBC connect attempt: it can only happen if the datasource now exists.

**What I could not do, stated plainly:** quote the `Unable to find datasource '<default>'` line from the pre-fix run. This service's console encoder emits JSON and drops the exception body, and `-Dquarkus.log.console.json=false` did not override the baked config. So the pre-fix half rests on the absence of the connect attempt plus the boot failure, not on the exception text.

## The gate

`check-flyway-default-datasource.py`, registered as `flyway-default-datasource` in `.github/gates/gates.yaml` (`group: data`, `mode: enforced`). Two rules:

1. `migrate-at-start` ⇒ `quarkus.datasource.jdbc.url` present (a `${ENV:default}` expression counts — the literal default is what makes it bootable);
2. `quarkus.flyway.datasource` must be a string, never a mapping.

**Falsified against the real shape:** re-breaking sca-service to the exact pre-fix config makes it print both findings and exit 1; restored, `46 migrate-at-start service(s) — clean`. The `--selftest` covers six cases both ways, including the two near-misses (an env-expression URL, and a legitimate datasource *name*).

It does not replace a boot gate (#2872). It closes the one failure mode that recurred five times and is decidable from a single committed file.

## Note for the reviewer

`%test` profiles override the datasource with their own `jdbc:` URL, which is exactly why no test could see this — so the test suites are unaffected by the change, and CI's per-service builds are the confirmation.

### DIFF
--- a/.github/gates/gates.yaml
+++ b/.github/gates/gates.yaml
@@ -963,6 +963,22 @@ gates:
     run: |
       python3 .github/scripts/check-kafka-topics-exist.py --enforce
 
+  # Five money-path services could not start from their committed configuration: each declared
+  # only a REACTIVE datasource URL, so the default Agroal (JDBC) datasource Flyway migrates was
+  # never configured and the boot died with "Unable to find datasource '<default>'". They ran in
+  # the cluster only because the gitops env supplied QUARKUS_DATASOURCE_JDBC_URL; `quarkusDev`
+  # and the fuzz harness (#3039) got the failure. Three carried a `quarkus.flyway.datasource:`
+  # MAPPING that reads as the fix and is not — that key names which datasource to use, it does
+  # not define one. Decidable from one committed file, so no boot and no cluster.
+  - id: flyway-default-datasource
+    name: "Flyway default datasource (issue #3080, enforced)"
+    group: data
+    mode: enforced
+    selftest: |
+      python3 .github/scripts/check-flyway-default-datasource.py --selftest
+    run: |
+      python3 .github/scripts/check-flyway-default-datasource.py --enforce
+
   - id: kafka-acl-coverage
     name: "Kafka ACL coverage (issue #2598, enforced)"
     group: data
--- a/.github/scripts/check-flyway-default-datasource.py
+++ b/.github/scripts/check-flyway-default-datasource.py
@@ -1,0 +1,146 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: Apache-2.0
+"""A service with `quarkus.flyway.migrate-at-start` must configure the datasource Flyway migrates.
+
+Why this exists
+---------------
+Five services — ledger, account, sca, transaction and settlement, every one of them money-path —
+could not start from the configuration committed to this repo (#3080). Each declared only a
+REACTIVE datasource URL, so the default Agroal (JDBC) datasource was never configured, while
+`quarkus.flyway.migrate-at-start: true` requires one:
+
+    io.quarkus.runtime.configuration.ConfigurationException: Unable to find datasource '<default>'
+      for persistence unit '<default>': Bean is not active: SYNTHETIC bean
+      [class=io.agroal.api.AgroalDataSource]
+
+They booted in the cluster only because the gitops env supplied `QUARKUS_DATASOURCE_JDBC_URL`. A
+developer running `quarkusDev`, and the authenticated fuzz harness (#3039), got the failure.
+
+Three of them looked like they had solved it: they carried a `quarkus.flyway.datasource:` block
+with `db-kind`, `username`, `password` and a nested `jdbc.url`. Quarkus does not read that path as
+a datasource *definition* — `quarkus.flyway.datasource` names WHICH existing datasource Flyway
+should use. So the config answered the question for the next reader, incorrectly, and the boot
+error names the missing default datasource while that block sits in the same file. That shape is
+worse than the plain omission, so it is flagged separately.
+
+WHAT IT CHECKS, per `openbank-*/src/main/resources/application.yaml`:
+
+  1. `quarkus.flyway.migrate-at-start` truthy ⇒ `quarkus.datasource.jdbc.url` must be present
+     (a `${ENV:default}` expression counts — the literal default is what makes it bootable).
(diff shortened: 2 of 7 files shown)


JiRaska:
## Merging on an explicit four-eyes waiver from the repo owner

Five money-path services' `application.yaml` plus a new gate, so I read the change rather than the checks.

### The config is coherent on all five

Parsed each resulting file with a duplicate-key-detecting loader, because SnakeYAML silently keeps only the last of a repeated mapping key:

```
                 dup keys   datasource keys                              flyway.datasource
account              0      db-kind, jdbc, password, reactive, username   absent
ledger               0      db-kind, jdbc, password, reactive, username   absent
sca                  0      db-kind, jdbc, password, reactive, username   absent
settlement           0      db-kind, jdbc, password, reactive, username   absent
transaction          0      db-kind, jdbc, password, reactive, username   absent
```

Nothing was lost — sca's removed `datasource:` block was consolidated, not dropped; `username`, `password` and `db-kind` all survive. Every `jdbc.url` is `${QUARKUS_DATASOURCE_JDBC_URL:…}`, so the cluster env still wins and the literal is only the local-dev default.

### The gate is falsifiable — checked in both directions

Ran the branch's script against `origin/main`, which is the state it exists to catch:

```
origin/main   8 findings, exactly the 5 services   (missing jdbc.url ×5, ineffective flyway.datasource mapping ×3)
branch        46 migrate-at-start services — clean
--selftest    OK: 6 cases, both directions, 54 configs parsed
```

Declaration is consistent too: `mode: enforced` and `run:` actually passes `--enforce`, so it is not enforced-on-paper-only. Without the flag the script emits `::warning` and exits 0.

Two of my own probes were broken before this worked, both reading as "the gate is broken": running the script from a path other than `.github/scripts/` makes it resolve the repo root two levels too high and report `0 application.yaml parsed`, and my first known-positive ran against a stale local checkout rather than `origin/main`.

### State at merge

All five required contexts green, nothing running, no non-required failures, `mergeStateStatus` CLEAN, head unchanged at `fb0b6fec1` from the start of this review.

*Merged at the owner's explicit instruction; money-path PRs are otherwise left for a human during unattended drains.*

Issue #3194, opened 2026-08-01
Deploy PRs born conflicted never get a merge ref, so no CI ever runs — 4 today, 3 closed undeployed


JiRaska:
## Symptom

Four admin-ui deploy PRs today had **zero workflow runs** and were unmergeable from the moment they opened. Three were closed without ever deploying:

```
#3138  0 runs  CLOSED
#3154  0 runs  CLOSED
#3162  0 runs  CLOSED
#3183  0 runs  (recovered by hand, see below)
```

For contrast, the ones that merged the same day had 18 runs each: #3181, #3166, #3163, #3161.

A deploy PR with no runs can never satisfy the five required contexts, so it sits until a later deploy supersedes it and someone closes it. The image it was carrying is simply never deployed.

## Mechanism

A PR that is **already conflicted when it is created** never gets `refs/pull/<n>/merge`. GitHub cannot compute the merge commit, and `pull_request` workflows run against that ref — so no workflow is ever created. Not skipped: absent. Nothing reports, and re-running is not possible because there is no run to re-run.

Measured across all 18 open PRs at the time:

```
#3183   DIRTY     merge ref 0    runs  0     <- born conflicted
#3058   DIRTY     merge ref 1    runs 13     <- became conflicted later
everything else   merge ref 1    runs >0
```

`DIRTY` alone is not the predictor. #3058 is equally conflicted and has a full check set, because its merge ref was created while the PR was still clean and the conflict arrived afterwards when `main` moved. **The distinguishing property is being conflicted at creation time.**

Confirmed causally on #3183 — merging `main` into the branch and pushing gave it a new head:

```
before   refs/pull/3183/merge = 0    workflow runs =  0    DIRTY
after    refs/pull/3183/merge = 1    workflow runs = 12    BLOCKED, then merged
```

## Root cause

`.github/workflows/admin-ui-deploy.yml`

The checkout has no `ref:`, so on a `push` to main it checks out `github.sha` — the commit that triggered the build, not the current tip of `main`:

```yaml
- uses: actions/checkout@... # v7.0.1
  with:
    fetch-depth: 0
```

`peter-evans/create-pull-request` then branches from that HEAD and targets `main`:

```yaml
branch: chore/admin-ui-deploy-${{ github.sha }}
```

[...]

PR #3211, opened 2026-08-02
fix(lending,domestic): align the entities with the DDL their own migrations create

### LINKED ISSUE #3081: Entity/DDL drift in three money-path services — Hibernate schema validation fails against their own migrations
Three money-path services fail Hibernate's schema validation against a database built by their own
Flyway migrations — i.e. the JPA model and the migrations disagree.

| Service | Finding |
|---|---|
| `openbank-sepa-payment` | `Schema validation: missing table [object_store_blobs]` |
| `openbank-lending-service` | `wrong column type encountered in column [currency] in table [collateral]` — migration has `CHAR(3)` |
| `openbank-domestic-payment` | `wrong column type encountered in column [reject_detail] in table [domestic_payments]` — migration has `TEXT` |

Measured in the authenticated fuzz harness (#3039), against a **fresh** Postgres with only that
service's own migrations applied — so this is not environment drift.

**Production does not see it** (`quarkus.hibernate-orm.database.generation: none`, so nothing
validates) and dev mode does — which is the wrong way round: the mismatch is real either way, and
the environment that hides it is the one that matters.

`sepa-payment`'s is more than cosmetic. `object_store_blobs` is an `@Entity` in
`openbank-libs-runtime` (`PostgresBlobStore`), and **only `openbank-document-service` has a
migration creating it**. sepa-payment reaches that entity through the ordinary
`implementation(project(":openbank-libs-runtime"))` that 47 modules share — no direct code
reference, no `index-dependency` entry. An entity whose table does not exist does not fail at
boot under `generation: none`; it fails at first use.

**Open question, deliberately not answered:** whether the other 46 libs-runtime dependants share
the `object_store_blobs` gap. Only sepa-payment reported it, but lending and domestic-payment
aborted validation on an earlier mismatch and would never have reached it — their silence is not
evidence. Settling it needs a per-service boot (#2872).

Refs #3039, #2872

JiRaska:
Closes #3081 · Refs #3039, #2872

## Two real drifts, both wider than reported

The migrations are applied, so their checksums are frozen — the entities are what move.

**lending** — the issue names `collateral`. Validation aborts on the first mismatch, so it could only ever name one. **Five** tables declare `currency CHAR(3)` (`loan_application`, `loan`, `installment`, `collateral`, `loan_provisioning`) while every entity said `length = 3`, which maps to `varchar(3)`. All five pinned with `columnDefinition = "char(3)"`. `settlement_quote` (V10) is genuinely `VARCHAR(3)` and is deliberately left alone.

**domestic-payment** — `reject_detail` is `TEXT` in V1; without `columnDefinition` a String maps to `varchar(255)`.

## The third finding is not a defect — and this answers the issue's open question

The issue asks, deliberately unanswered, whether the other 46 libs-runtime dependants share sepa-payment's `missing table [object_store_blobs]`. Measured statically, no boot needed:

- 43 modules depend on `openbank-libs-runtime`; **37** of those use Hibernate;
- **exactly 1** (`document-service`) has a migration creating `object_store_blobs`.

So all 36 others share it — **and that is by design**, which is why adding 36 migrations would be the wrong fix. `openbank-libs-runtime` carries the Jandex plugin, so the entity is visible to every dependant's scan, and `PostgresBlobStore`'s own KDoc says exactly this: the entity is visible regardless and *"every query against it simply fails until the service's own migration has created the table"*. Under `quarkus.hibernate-orm.database.generation: none` — what production runs — an unused entity is inert. The finding appeared only because the fuzz harness turns validation on.

## What is worth gating

The pairing that genuinely breaks: a service that **injects** the port and resolves to the **Postgres** backend, with no table. That fails at first use — a 500 on whichever request first touches a blob, with nothing before it to say so. `check-object-store-blobs-migration.py`, registered as `object-store-blobs-migration` (`group: data`, `mode: enforced`).

**Its first run flagged a false positive, and that shaped the rule.** `openbank-customer-edge` injects `ObjectStorePort` for ADR-0192 feedback screenshots and sets `objectstore.backend: s3` — it never reads the table. `PostgresBlobStore` carries `@IfBuildProperty(..., "postgres", enableIfMissing = true)`, so absence of the key selects Postgres; that asymmetry is now in the rule, and both verdicts are pinned in the self-test against the real files.

**Its first self-test also passed while the check was vacuous** — worth stating, because it is the failure mode this repo cares about. `creates_table` matched `"create table"` and the table name anywhere in a file, and document-service's V1 *mentions* the table in a comment while creating others. So V1 read as the creator, and deleting the real V4 still reported clean. Fixed by matching the `CREATE TABLE <table>` **statement**, and the self-test now asserts both that document-service is detected as a user and that V1 is not mistaken for the creator. With that, removing V4 flags document-service and restoring it goes clean.

## Verification

`ktlintCheck`, `detekt`, `test` green for both edited services. The gate runs through `run-gates.py --only object-store-blobs-migration`.


JiRaska:
## Merging — and the one red check is a different repo moving

Merged at the owner's explicit instruction. Notes for the record.

### It was born conflicted, which is why it had no CI at all

```
before   refs/pull/3211/merge = 0    workflow runs = 0
after    refs/pull/3211/merge = 1    workflow runs = 12
```

Opened at 04:16 already conflicting with `main`, so GitHub never created the merge ref and no `pull_request` workflow existed to run or re-run (#3194). The zero-check state reads as "waiting", not "broken". Worth noting this is the first instance of that failure in a **human-authored code** PR — the four before it were all deploy bots, which is how the issue was framed.

The only conflict was `.github/gates/gates.yaml`, and it was self-inflicted: #3192 added `flyway-default-datasource` to `main` an hour earlier while this branch was adding `object-store-blobs-migration`. Both additive, both kept — 73 gates, no duplicate ids.

### The new gate is falsifiable

Checked in both directions rather than trusting the green:

```
--selftest   exit 0   60 modules; detects document-service's migration and its postgres default,
                      exempts customer-edge's s3 backend, and does not mistake
                      ledger-service's absent migration for one
--enforce    exit 0   1 object-store user — clean
mode: enforced   run: … --enforce      <- declaration matches behaviour
```

### `Customer-app dossier` is unrelated

Non-required, and not caused by anything here — this PR touches 0 admin-ui files:

```
version: committed="0.11.0" → regenerated="0.12.0"
```

`openbank-app` released 0.12.0 and the committed `app-status.json` still says 0.11.0. It was green on #3192 and #3044 because those ran before that release. It will now be red on **every** PR until the dossier is regenerated — per ADR-0074 the publish PR normally comes from openbank-app's own CI, so this is a heads-up rather than something to fix here.

### On the change itself

Aligning the entities to the DDL — rather than migrating the DDL — is the only available direction, since the migrations are applied and their checksums frozen. Leaving `settlement_quote` alone is right: it is genuinely `VARCHAR(3)`.

Scope note: this fixes 6 columns. A scan of `origin/main` puts the same drift at **123 columns across 40 services**, 17 of them money-path — detail on #3081.

*Overnight queue drain.*

PR #3236, opened 2026-08-02
feat(admin-ui): show a loan application's real path through the ADR-0211 lifecycle

JiRaska:
## What

`/lending/applications/{id}` — one loan application drawn as the ADR-0211 lifecycle it is actually
in: a vertical rail with the **real path** on it (done steps ticked with time and actor, current step
ringed, the rest pending, early exits separated), and the ADR-0214 evidence trail underneath. Reached
from a **Progress ›** link on every row of the origination queue.

## Why

The console lists applications as rows with a status string. That answers *which applications exist*,
not *where is this one, how did it get there, and what is it waiting for* — which is what the credit
desk asks. Nothing in the UI answered it. The only lifecycle picture anywhere is the static
`/docs/bpmn` diagram: generic, not per-application, and stale — it draws four steps against the
sixteen the machine really has (`DRAFT … REFLECTION_PERIOD … EXPIRED`).

## The state list is generated, not typed

`scripts/generate-origination-graph.mjs` derives the states, the terminal set and the edges from
`OriginationState.kt` + `OriginationTransitionPolicy.kt`, wired into `pretest`/`prebuild` — the same
bake pattern as `catalog.json` / `governance.json`, and gitignored like them.

A diagram carrying its own copy of the state list would drift from the machine **while still
rendering perfectly convincingly**. Guards, all falsified against a deliberately broken source:

| broken input | result |
|---|---|
| `allowedTransitions` renamed (regex matches nothing) | `could not locate standard()'s allowedTransitions map`, exit 1 |
| a non-terminal state left with no outgoing edge | `SIGNED is non-terminal but has no outgoing transition`, exit 1 |
| a state label removed | suite red — `every state has a business label` |
| forbidden evidence read treated as empty | suite red — `a restricted evidence read is stated…` |

The generator refuses to emit `{states: [], edges: []}`; the suite refuses a happy path that stops
covering every non-terminal state.

## Vertical because thirteen steps do not fit sideways

The first cut was horizontal. Screenshotted, it needed ~2000px — the operator would scroll sideways
to see "the whole flow", the one thing they asked to see at once. Wrapping was worse: the arrow after
the current step pointed off a row end into nothing and the sequence restarted on the far left, so
reading order stopped matching the process. Three versions were rendered and looked at, not reasoned
about.

## What it deliberately does not do

Decisions and disbursement are **not** buttons here — ADR-0227 D4 puts money-path disposal in the
approval inbox behind SCA. They render **visible and disabled** with the reason and a link, because an
operator who cannot find a control assumes the platform is broken, while a disabled one teaches them
where the control lives. `advance` is exposed: it only walks the graph the machine already validates
and moves no money.

A 403 on the evidence read is ordinary — it is compliance/credit-risk only, while the queue is also
open to operators — and is stated, never rendered as "this application has no history".

The evidence table is not a duplicate of the rail: the rail keys facts by state, so a revisit
(`ASSESSMENT → DOCS_REQUIRED → ASSESSMENT`) collapses to one dot; the table is every transition in
order, which is what an auditor asks for.

## Verification

- `npm test` — 66 files, **973 tests passed** (9 new)
- `npx tsc --noEmit` — clean
- `npm run lint` — 0 errors
- Rendered and screenshotted at each iteration; the generator run against the live tree yields
  **16 states (4 terminal), 41 edges**, matching the Kotlin exactly

admin-ui only — no API, DB, event or config change.

Refs #3168


JiRaska:
## Both admin-ui checks fail on one missing pre-hook

All five required contexts are green, but `Admin UI build` and `Admin UI` are red, and it is a single cause:

```
##[error] src/components/lending/OriginationFlow.tsx(18,19): error TS2307:
          Cannot find module '../../../origination-graph.json'
```

The generator is wired into two of the three entry points:

```
prebuild  … && node scripts/generate-origination-graph.mjs     ✓
pretest   … && node scripts/generate-origination-graph.mjs     ✓
type-check  tsc --noEmit                                       ← no pre-hook
```

npm only runs `pre<name>` for the matching script name, and there is no `pretype-check`. CI calls `npm run type-check` directly (`ci.yml:91`), so `tsc` runs against a tree where `origination-graph.json` has never been generated — and it is gitignored, so it is never there from a checkout either. `Admin UI`'s "Check build result" is downstream of the same failure.

Adding `"pretype-check": "node scripts/generate-origination-graph.mjs"` fixes both.

## Why the existing derived artifacts do not hit this

Worth knowing before reaching for a different fix. `governance.json`, `catalog.json` and `prod-readiness.json` are gitignored exactly the same way, and they never break `tsc` because nothing **imports** them — they are read at runtime:

```ts
// src/app/api/catalog/governance/route.ts
return process.env.OPENBANK_GOVERNANCE ?? path.resolve(process.cwd(), 'governance.json')
```

That pattern is not available here: `OriginationFlow.tsx` is `'use client'`, so it cannot read from the filesystem, and a static import is the right call for a browser component. So this is a genuinely new shape in this codebase — first statically-imported generated artifact — rather than a deviation from the house style.

Which is also why the pre-hook is worth adding rather than just making CI green: **any** type-check outside npm's hook sees the same error — an editor's language server, a bare `tsc`, a fresh clone before the first build. A contributor opening the repo will see a red squiggle on line 18 with no obvious cause.

## The design itself is the right way round

Deriving states and edges from `OriginationState.kt` + `OriginationTransitionPolicy.kt` rather than hand-drawing them, and failing `originationFlow.test.tsx` when a generated state has no human label, means adding a state in Kotlin surfaces as a red test instead of a node quietly labelled with a raw enum name. That is the drift this repo keeps paying for elsewhere — a second copy of a list that renders perfectly convincingly while being wrong.

`agent-review` is red for #2161, unrelated.

*Reviewed from the PR queue; branch untouched.*

Issue #1915, opened 2026-07-23
Graduate supply-chain verification: Kyverno SBOM/signature Audit->Enforce, dependency verification lenient->strict


JiRaska:
Per the 2026-07-16 platform audit (§4.4), the supply chain is ~SLSA Build L3 on the *generation* side, but verification is not a blocking admission gate:

- Kyverno SBOM policy is **Audit**, not Enforce; an image-signature Enforce was reverted in #770
- Gradle dependency verification runs **lenient/warn-only** (deadline 2026-09-30 already set)

Scope:
- Re-attempt Kyverno signature+SBOM Enforce with the lessons from the #1197 `.att` manifest rewrite (cosign attest is additive + any-match while Kyverno is ALL-match — the false-precondition trap from the SBOM-enforce incident)
- Flip dependency verification lenient -> strict per the existing target date
- Add an admission-time rollback runbook so an Enforce regression cannot repeat the 4-day money-path outage

A green `verify-attestation` can be about an earlier build's envelope — use `openbank-infra/scripts/lib/cosign-attest.sh`, never hand-rolled trivy+cosign.

JiRaska:
**Progress (not closing):** the Kyverno headline is DONE on main — both `verify-images-policy.yaml` and `verify-sbom-attestation-policy.yaml` are `validationFailureAction: Enforce` (SBOM graduated 2026-07-12, ADR-0144). **Remaining, owner-gated:** dependency verification is still lenient (`gradle/verification-metadata.xml: verify-signatures=false`, owner deadline 2026-09-30) + the admission-rollback runbook. Suggest narrowing scope to 'dependency-verification strict + rollback runbook'.

JiRaska:
**Status (verified against live gitops, 2026-07-25):** the Kyverno half of this issue is **done**.
- `verify-images-policy.yaml` (image signatures) → **Enforce**. The 2026-06-11 first attempt reverted per #770 (cosign v3 OCI-1.1 referrer sigs kyverno 3.2.6 couldn't discover) was re-fixed by pinning cosign v2 (tag-based .sig) and re-signing the fleet; it's back to Enforce.
- `verify-sbom-attestation-policy.yaml` (SBOM attestation) → **Enforce**, graduated Audit→Enforce 2026-07-12, kept as a *separate* ClusterPolicy so its staging isn't coupled to the signature rule (the #770 lesson).
[...]

PR #3238, opened 2026-08-02
DO NOT MERGE: prove the verification-metadata gate detects an unpinned dependency

### LINKED ISSUE #3218: verification-metadata is never refreshed on human PRs — only Dependabot's
## Summary

`gradle/verification-metadata.xml` is only ever refreshed on **Dependabot** PRs. A human PR that
adds or bumps a dependency never gets its checksums written, so the artifact ships unpinned.

Today that is invisible (`org.gradle.dependency.verification=lenient` → warning, not failure). It
stops being invisible at the ADR-0144 graduation on **2026-09-30** (#1915), when the same PR starts
failing every cold-cache build.

## Evidence

`.github/workflows/dependabot-verification-metadata.yml` is the **only** thing in the repo that
regenerates the file — verified by grepping every workflow on `main` for
`write-verification-metadata`; one hit.

It is gated twice:

```yaml
on:
  pull_request:
    paths: ["**/build.gradle.kts", "**/libs.versions.toml", "gradle/wrapper/**",
            "gradle.properties", "settings.gradle.kts"]
...
if: github.actor == 'dependabot[bot]' && github.event.pull_request.head.repo.full_name == github.repository
```

The path filter is right. The **actor** gate is the hole: a human editing `build.gradle.kts`
triggers the workflow and the job then skips itself.

The actor gate is not a mistake — it exists because the job pushes a commit back to the PR branch,
and doing that for arbitrary contributors is a real trust-boundary decision (see the security note
at the top of that file). So this needs a different mechanism, not a relaxed `if:`.

## Why now

#3199 just fixed the *other* half of this problem: the producer resolved only the `testClasses`
graph, so prod-runtime-classpath artifacts were never hashed at all (root cause of #3017; 38 entries
were missing for one service, #3028). With that closed, the actor gate is the largest remaining way
for an unhashed artifact to reach `main`.

Measured while investigating: a cold `testClasses` regeneration across all 60 modules adds **zero**
entries, so the test graph is complete and Dependabot's coverage of it is working. The gap is
entirely (a) prod-side, now handled by #3199 for Dependabot PRs, and (b) human PRs, which is this
issue.

## Options

1. **Verify instead of write.** On any PR touching Gradle files, run the regeneration into a
   throwaway copy and fail if the committed file would change. No write-back, so no trust boundary
   — the author regenerates locally and commits it themselves. Downside: a manual step, and
   `--write-verification-metadata` needs `-Xmx6g`+ and takes ~20 min cold fleet-wide (16m45s for
   `testClasses` alone), so it must run warm-cached and ideally scoped to changed modules.
2. **Scheduled cold-cache job on `main`.** Catches everything regardless of author, but only
   *after* merge. Already proposed in #3131 as the regression guard; this issue is the argument for
   why it is not optional.
3. **Write-back for human PRs behind the same-repo check only** (drop the actor condition, keep
   `head.repo.full_name == github.repository`). Cheapest, but needs a deliberate decision that
   pushing to a contributor's branch is acceptable — and it still would not cover fork PRs.

Recommendation: 1 as the merge-blocking signal, 2 as the backstop. 3 only if the trust question is
settled explicitly.

## Acceptance criteria

- [ ] a human PR that adds a production-only dependency without updating
      `verification-metadata.xml` is caught **before** merge
- [ ] the check is exercised against a case it must flag — add a dependency, omit the metadata,
      confirm red — not just observed passing
- [ ] whatever runs is sized: measured cold, `testClasses` is 16m45s and
      `testClasses quarkusDependenciesBuild` is 20m10s over 60 modules; CI runs warm
- [ ] resolved before the `lenient` → `strict` flip (#1915, 2026-09-30), since after it this
      failure mode becomes a broken build rather than a silent warning

Refs #1915, #3131, #3199

JiRaska:
## ⚠️ DO NOT MERGE — this branch exists to make a check go red

Adds `org.jsoup:jsoup:1.18.3` to `openbank-campaign-service` and **deliberately does not**
regenerate `gradle/verification-metadata.xml`. jsoup appears nowhere in that file today.

## Why

#3233 added a gate that fails a PR whose dependency changes are not pinned. It merged with only its
**no-op path** exercised in CI: that PR changed nothing under `openbank-*/build.gradle.kts`, so the
gate correctly skipped, and green meant "did not block" — not "can detect".

Detection was verified only on my machine. The part it hinges on, `--refresh-dependencies`, is
precisely the piece whose behaviour differs between a local cache and CI's, and without it the check
reports "no gaps" even with an entry demonstrably deleted. So the one thing that makes the gate real
is the one thing CI has never exercised.

## Expected result

`verification-metadata` **fails**, naming the missing jsoup artifacts.

| outcome | meaning |
|---|---|
| ❌ red, names jsoup | gate works — close #3218, then close this PR unmerged |
| ✅ green | gate is a no-op in CI; #3218 is **not** resolved and #3233 needs rework |

A green result here is the valuable one to discover now rather than on 2026-09-30, when
`org.gradle.dependency.verification` graduates to `strict` (#1915) and this failure mode turns from
a silent warning into every cold-cache build breaking at once.

## Disposal

Close unmerged once the check reports. Nothing here is intended for `main`.

Refs #3218, #3233

### DIFF
--- a/openbank-campaign-service/build.gradle.kts
+++ b/openbank-campaign-service/build.gradle.kts
@@ -8,6 +8,10 @@ plugins {
 }
 
 dependencies {
+    // DELIBERATELY UNPINNED — this dependency is absent from
+    // gradle/verification-metadata.xml on purpose. Do not "fix" it: the point of
+    // this branch is to prove the #3233 gate goes red. Never merge.
+    implementation("org.jsoup:jsoup:1.18.3")
     implementation(enforcedPlatform(libs.quarkus.bom))
     implementation(libs.quarkus.kotlin)
     implementation(libs.quarkus.resteasy.reactive)


JiRaska:
## Purpose served — closing unmerged, as intended

The gate went red and named exactly the unpinned dependency:

```
  missing: jsoup-1.18.3.jar  (org.jsoup:jsoup:1.18.3)
  missing: jsoup-1.18.3.pom  (org.jsoup:jsoup:1.18.3)
```

Gradle completed rather than crashing, so this is a real detection and not the script's
"could not run" path. `--refresh-dependencies` therefore works on the runner's warm cache — the one
thing #3233 could not demonstrate on its own.

#3218 closed on the strength of this run. Nothing here belongs on `main`.

Issue #3178, opened 2026-08-01
Sandbox: 3 money-path rollouts aborted for 6-9 days, auto-deploy failing 83%, Loki out of disk and alloy OOMKilling


JiRaska:
Sandbox health check, 2026-08-01 ~20:00 UTC. **Serving fine — 435 of 468 pods Running, all 55 CNPG clusters healthy, 122/123 Deployments at desired, 90/90 ArgoCD apps Synced.** These are the parts that are not, none of them new today.

## 1. Three money-path Rollouts have been `RolloutAborted` for 6–9 days

| rollout | ready | image commit | age |
|---|---|---|---|
| `payments/sepa-payment` | 1/1 | `sandbox-90d7d3ad` | 9 days |
| `payments/settlement-service` | 1/1 | `sandbox-8bb90924` | 6 days |
| `payments/transaction-service` | 1/1 | `sandbox-a5e5d32a` | 9 days |

`Progressing=False reason=RolloutAborted`. The `stable` ReplicaSet is serving 1/1, so there is **no outage** — which is exactly why this has gone unnoticed: three money-path services have been pinned to week-old images while every subsequent merge appeared to succeed. `currentStepIndex=0` on all three.

## 2. `auto-deploy` is failing 83% of runs, and has been for at least a day

**33 failures / 7 successes** over the last 40 runs. Failures at 00:59, 04:26, 04:32, 07:15, 07:57, 09:09, 09:17, 10:06, 10:50, 10:57 and onward — continuous, not a burst.

Always the same job (`can-i-deploy`) and the same verdict:

```
can-i-deploy: openbank-lending-service NOT deployable [UNVERIFIED]
[openbank-lending-service] MONEY-PATH service blocked by can-i-deploy
[openbank-lending-service] is blocked on openbank-ledger-service, which is NOT part of
this run — a co-deploy cannot help
```

`UNVERIFIED` is classified as self-clearing ("the counterpart verifies minutes later"), but it has not cleared in a day, so the classification is wrong for this case — the 3-hourly reconcile is retrying something no tick can fix. That is the failure shape #2549 warns about for `REGRESSION`, arriving via a class that is not treated as one.

This is the plausible cause of finding 1: if the deploy gate has been blocking, the rollouts never got a newer image to move to.

## 3. `loki-0` — disk full, ~144 restarts

```
error running loki ... write /var/loki/tsdb-shipper-active/scratch/... no space left on device
```

[...]

Issue #2338, opened 2026-07-25
Derive provider-replay coverage from pacts/ — nothing enforces that a committed pact is actually replayed pre-merge


JiRaska:
## Source ADR

ADR-0063 (consumer-driven contract testing, git-pact)

## Kind of follow-up

Policy / gate — encode a rule that is currently only prose in `CLAUDE.md`.

## What concretely must happen

`CLAUDE.md` states the rule: *never add a consumer pact without wiring (or confirming) the provider's `@PactFolder` replay* — because a consumer pact cannot catch a wrong request path, only the provider replay can (#2269). Nothing enforces it. That is how 16 of 27 pacts ended up with a replay that skips on every PR (#2327), and nobody noticed: the checks were green.

Build a check that DERIVES the provider-replay coverage set from the artifacts, and fails when a committed pact has no replay that actually runs pre-merge.

Shape, mirroring `.github/scripts/derive-pact-drift-scope.sh` (#2318):

1. Enumerate `pacts/*.json`. Each filename is `<consumer>-<provider>.json`.
2. Enumerate provider verification classes: `@Provider("<name>")` under `**/src/test/kotlin/**`.
3. For each pact, require at least one `@Provider("<its provider>")` class that runs **unconditionally on a PR**. A class does not count if any of these apply:
   - it is `@PactBroker`-sourced (`_service-ci.yml:72` blanks `PACT_BROKER_URL` off main-push, so `@EnabledIfSystemProperty(named = "pactbroker.url", matches = ".+")` skips it);
   - it carries `@Disabled`, `@DisabledIfEnvironmentVariable`, or any other gating annotation;
   - its module's `build.gradle.kts` excludes it under `if (System.getenv("CI") == "true")` — the `openbank-swift-service` shape.
4. Fail on an orphan class too: a `@Provider("x")` for which no `pacts/*-x.json` exists is either a typo in the provider name or a deleted pact, and both read as coverage.
5. Exceptions live in ONE explicit allow-list that names the pacts each exception strands, checked in both directions so a stale entry is itself a failure — the `UNGATED_PACTS` pattern.

## Why derived rather than a checklist item

This is the same failure mode #2318 fixed one layer down. A gate whose coverage set is maintained separately from the artifacts it covers reads as *passing* when the set is short, never as *unchecked*. A PR-template checkbox or a `rules.yaml` list of "providers that must have a folder replay" would rot exactly the way `pact-drift-check.yml`'s hand-written module list did.

[...]

PR #3241, opened 2026-08-02
feat(ci): give the eight @PactFolder-only providers the @PactBroker half so their verifications reach the broker

### LINKED ISSUE #3232: 8 of 17 pact providers have only a @PactFolder replay, so they never publish a verification result — their consumers can never pass can-i-deploy
`@PactFolder`-only providers never publish a verification result to the Pact Broker, so their consumers stay `UNVERIFIED` forever and `can-i-deploy` blocks them indefinitely. **8 of 17 providers are in that state.**

## Why this is invisible

`check-pact-provider-replay.py` (#2338) requires every committed pact to have a **`@PactFolder`** replay that runs on a PR — and it is satisfied. But a `@PactFolder` test reads the pact from disk and never contacts the broker: it proves the provider still honours the contract, and publishes nothing. The sanctioned pair documented in `CLAUDE.md` is `@PactFolder` **plus** a `@PactBroker` + `@EnabledIfSystemProperty(pactbroker.url)` class, as `openbank-ledger-service` carries. Nothing enforces the second half, so it quietly never got added for these.

The replay gate is green, the pacts are committed, the provider tests pass — and the broker never learns any of it.

## The 8

| provider | consumers left unverifiable |
|---|---|
| `openbank-aml-service` | `openbank-fx-service` |
| `openbank-clearing-simulator` | `openbank-swift-service` |
| `openbank-consent-service` | `openbank-mcp-service` |
| `openbank-customer-edge` | `openbank-copilot-service` |
| `openbank-kyc-service` | `openbank-onboarding-service` |
| `openbank-product-catalog` | `openbank-card-issuance-service` |
| `openbank-sanctions-service` | `openbank-fx-service` |
| `openbank-tpp-registry-service` | `openbank-psd2-service` |

For contrast, these 9 carry the `@PactBroker` half and do publish: `openbank-account-service`, `openbank-balance-service`, `openbank-fraud-service`, `openbank-fx-service`, `openbank-ledger-service`, `openbank-party-service`, `openbank-sca-service`, `openbank-swift-service`, `openbank-transaction-service`.

## Two distinguishable sub-cases

Six of the 8 have **no `main`-branch version in the broker at all** (`/pacticipants/<p>/branches/main/latest-version` → 404): `openbank-aml-service`, `openbank-clearing-simulator`, `openbank-customer-edge`, `openbank-product-catalog`, `openbank-sanctions-service`, `openbank-tpp-registry-service`.

Two — `openbank-kyc-service` and `openbank-consent-service` — **do** have a main version, because they publish one as a *consumer*. That makes them look further along than they are: having a version and being able to verify are different things, and only the second makes them deployable counterparts.

## How it surfaced

The verification reconciler (#3221) reports it every run. `openbank-kyc-service` was its first live dispatch — and the debt could never have cleared, because kyc has no broker-publishing test. #3229 fixed the reconciler to skip providers that cannot publish and name them instead, so the report is now this inventory:

```
CANNOT PUBLISH, not dispatching: openbank-kyc-service ... It needs the @PactBroker half.
NO main VERSION, not dispatching: openbank-aml-service ...
... (6 total)
```

## What to do

Add the `@PactBroker` + `@EnabledIfSystemProperty(named = "pactbroker.url", matches = ".+")` verification class alongside the existing `@PactFolder` one, per the pair `openbank-ledger-service` already carries. The folder test keeps running on every PR with no infra; the broker one runs on main-push and publishes.

Worth considering afterwards: extend `check-pact-provider-replay.py` so a provider that is named in a committed pact must carry **both** halves. The current gate proves the contract is replayed; nothing proves the result ever reaches the broker, which is the half `can-i-deploy` actually reads.

JiRaska:
A `@PactFolder` test replays the COMMITTED pact from disk: it proves the provider
still honours the contract, on every PR, with no infrastructure. It never contacts
the broker, so it publishes nothing — and `can-i-deploy` reads published
verification results, not green test runs.

Eight of seventeen providers had only that half, so the broker never learned they
verify anything and their consumers stayed permanently UNVERIFIED, unable to
deploy. Nothing caught it: check-pact-provider-replay.py (#2338) enforces the
FOLDER half and is satisfied; nothing enforced the broker half, so it quietly
never got added.

  openbank-aml-service            openbank-kyc-service
  openbank-clearing-simulator     openbank-product-catalog
  openbank-consent-service        openbank-sanctions-service
  openbank-customer-edge          openbank-tpp-registry-service

Each gets a sibling class derived MECHANICALLY from its own @PactFolder class —
same Quarkus resources, same @TestSecurity, same target, same @State methods —
with @PactFolder -> @PactBroker and an @EnabledIfSystemProperty(pactbroker.url)
gate, matching the pair openbank-ledger-service already carries. The folder class
keeps running unconditionally, so PR-time coverage is unchanged; the broker class
runs on the main push, where _service-ci.yml sets PUBLISH_RESULTS=true.

Two things the compiler found that reading would not have:

  * customer-edge declares its fixture ids as FILE-LEVEL `internal const val`s.
    Copying the file redeclared every one in the same package — an
    overload-resolution ambiguity on every use. The sibling now references the
    originals and says why, so the next person does not "helpfully" re-add them.
  * All eight compile (`compileTestKotlin`, exit 0, zero errors), ktlint and
    detekt clean. Worth stating that the exit code was read directly and not
    through a pipe — `./gradlew | tail` reports tail's status, which is how a
    failing build reads as green.

Also updates the reconciler's self-test. It asserted on the FLEET's population —
that some provider could publish and some could not — and this change made the
second half false by fixing it. That assertion did its job and then became a
maintenance trap, so it now tests can_publish_verification against synthetic
fixtures (@PactBroker, @PactFolder-only, no test, missing directory) plus a light
"something in this repo can publish" sanity check. It tests the function, not the
estate.

Verified: the reconciler now sees 17/17 providers as able to publish, so
openbank-kyc-service — the one genuine strand it found and could not act on —
becomes dispatchable.

Closes #3232
Refs #3178

### DIFF
--- a/.github/scripts/pact-reconcile-verifications.py
+++ b/.github/scripts/pact-reconcile-verifications.py
@@ -343,22 +343,44 @@ def self_test() -> int:
     if not edges:
         bad.append("edge list empty")
 
-    # can_publish_verification against the REAL repo, in both directions. A version that
-    # answered True for everything would have dispatched the folder-only providers
-    # forever; one that answered False for everything would dispatch nobody and look
-    # exactly like a healthy fleet.
+    # can_publish_verification in BOTH directions, against SYNTHETIC fixtures.
+    #
+    # This used to assert on the fleet's own population — that some provider could publish
+    # and some could not. That caught what it was built to catch and then went red for the
+    # right reason: #3232 added the @PactBroker half to all eight folder-only providers, so
+    # the "some cannot" half became false by being FIXED. An assertion whose truth depends
+    # on the fleet still having the defect is an assertion that must be edited every time
+    # someone improves things, and editing it under time pressure is how it ends up deleted.
+    # Fixtures do not have that problem: they test the function, not the estate.
+    import tempfile
+    with tempfile.TemporaryDirectory() as tmp:
+        tmpp = pathlib.Path(tmp)
+        for name, ann in (("prov-broker", "@PactBroker"), ("prov-folder", '@PactFolder("../pacts")')):
+            d = tmpp / name / "src" / "test" / "kotlin"
+            d.mkdir(parents=True)
+            (d / "T.kt").write_text(f'@Provider("x")\n{ann}\nclass T\n')
+        (tmpp / "prov-none" / "src" / "test").mkdir(parents=True)
+        checks = [
+            ("prov-broker", True, "@PactBroker provider is publishable"),
+            ("prov-folder", False, "@PactFolder-only provider is NOT publishable"),
+            ("prov-none", False, "provider with no test at all is NOT publishable"),
+            ("prov-missing", False, "provider directory that does not exist"),
+        ]
+        for name, want, why in checks:
+            got = can_publish_verification(tmpp, name)
+            mark = "ok " if got == want else "BAD"
+            print(f"  {mark} {why:48s} -> {got}")
+            if got != want:
+                bad.append(why)
+
+    # And a light fleet sanity check: if NOTHING in the real repo can publish, the function
+    # is answering False for everything and no verification would ever be dispatched — which
+    # looks exactly like a healthy fleet.
     providers = sorted({p for _, p in edges})
     pub = [p for p in providers if can_publish_verification(root, p)]
-    folder = [p for p in providers if not can_publish_verification(root, p)]
-    print(f"  {'ok ' if pub else 'BAD'} providers that CAN publish: {len(pub)}")
-    print(f"  {'ok ' if folder else 'BAD'} providers that CANNOT (folder-only): {len(folder)}")
+    print(f"  {'ok ' if pub else 'BAD'} providers in this repo that can publish: {len(pub)}/{len(providers)}")
     if not pub:
-        bad.append("no provider can publish — the check is answering False for everything")
-    if not folder:
-        bad.append(
-            "every provider can publish — either the fleet changed, or the check is "
-            "answering True for everything; verify before removing this assertion"
-        )
+        bad.append("no provider in the repo can publish — the check answers False for everything")
 
     if bad:
         print("\n::error::self-test FAILED: " + "; ".join(bad))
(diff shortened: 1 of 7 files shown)


JiRaska:
## Two of the eight landed 20 minutes ago — and the conflict is prose only

`mergeStateStatus` is `DIRTY`. The cause is #3239 (`feat(aml,sanctions): publish provider verifications`), merged 05:45:55, which added the same two files this PR adds:

```
openbank-aml-service/…/AmlPactBrokerProviderVerificationTest.kt
openbank-sanctions-service/…/SanctionsPactBrokerProviderVerificationTest.kt
```

I compared content rather than reading the diff, because a file list computed against the merge-base cannot tell a duplicate from a divergence:

```
                                    raw diff   code lines (comments stripped)
AmlPactBrokerProviderVerificationTest        77 lines    main 40  ·  this PR 40
SanctionsPactBrokerProviderVerificationTest  81 lines    main 44  ·  this PR 44
```

Every differing line is KDoc. **The code is identical** — same imports, same annotations, same target, same state handlers. So the conflict is real to git and empty in substance: taking `main`'s version for those two files loses nothing.

## What is left is six, and six is exactly right

Scanning `origin/main` for `@Provider` classes with comments stripped — necessary, because these files discuss `@PactBroker` in their own KDoc and a naive grep counts the prose:

```
providers                    17
  both halves                11
  @PactFolder only            6
  two broker classes (the CLAUDE.md collision)   0
```

The six with no broker half are precisely the ones this PR still contributes:

```
openbank-clearing-simulator   openbank-kyc-service
openbank-consent-service      openbank-product-catalog
openbank-customer-edge        openbank-tpp-registry-service
```

So the premise holds, the count is just 20 minutes stale. Resolve by taking `main` for the aml and sanctions files and keeping the other six.

## The gap this closes is the one the folder-half gate cannot see

Worth stating plainly because it is subtle: `check-pact-provider-replay.py` (#2338) enforces that every committed pact has a `@PactFolder` replay running on a PR, and it is satisfied for all 17. But a folder replay reads from disk and publishes nothing, and `can-i-deploy` reads *published verification results*, not green test runs. So a provider could be fully verified on every PR and still leave its consumers permanently `UNVERIFIED` at the broker — green everywhere, undeployable. That is the shape behind the credit-platform deploy sitting at `can-i-deploy` on a ledger pact verification.

## One thing that will not be verified by merging this

The eight new classes carry `@EnabledIfSystemProperty(pactbroker.url)`, and `_service-ci.yml` blanks `PACT_BROKER_URL` outside main-push — so **none of them runs on this PR**. Whether each class's `@State` handlers actually satisfy the broker's pacts is answered only after merge, on the main-push lane.

That is inherent to the broker half rather than a flaw here, but it means the usual reassurance does not apply: a green PR says the six new files compile, not that they verify. If one has a state handler mismatch it surfaces as a red main-push job. Deriving them mechanically from the existing `@PactFolder` siblings is the right mitigation — it is worth confirming that each really is a mechanical transform of its sibling, since that assumption is doing all the load-bearing work.

*Reviewed from the PR queue; branch untouched.*

PR #3249, opened 2026-08-02
chore(infra): make the per-service Dockerfiles honest declarations, not dead recipes

### LINKED ISSUE #3016: 51 service Dockerfiles are structurally unbuildable and are never built
## Summary

Every one of the 51 `openbank-*/Dockerfile` files is structurally unbuildable, and nothing in
CI builds any of them. They read as the build recipe for their service. They are not, and cannot
be - the build stage would fail at Gradle configuration time on every single one.

Found while fixing an uber-jar `COPY` path in three of them (#3015).

## Evidence

`openbank-ledger-service/Dockerfile` is the file usually cited as the correct shape. A real
`docker build` fails in 77 s:

```
Included build '/build/build-logic' does not exist.
```

Measured across the fleet:

| gap | affected |
|---|---|
| no `COPY build-logic` — root `settings.gradle.kts` `includeBuild`s it for `openbank.quarkus-service` | **51 of 51** |
| no `COPY openbank-libs-{domain,runtime,temporal,testing}` — ADR-0122 split them into siblings, so `project(":openbank-libs-domain")` is unresolvable | **51 of 51** |
| uber-jar `-runner.jar` COPY path + hardcoded `-0.1.0-SNAPSHOT-` | 3 of 51 (fixed in #3015) |

Fixing the first two on `openbank-campaign-service` and rebuilding gets past configuration and
then hits the cold-cache dependency-verification gap tracked separately.

## Why nothing noticed

No CI path builds a per-service Dockerfile:

- `auto-deploy.yml` and `ghcr-publish.yml` both `cp .github/workflows/Dockerfile.deploy` into a
  generated context
- `openbank-infra/scripts/build-push-service.sh:8` documents the choice outright —
  *"WHY NOT `docker build -f openbank-<svc>/Dockerfile`"* — because in-image Gradle hits
  download timeouts

All three build host-side and read exactly one thing from the per-service file: the `EXPOSE`
line, used to set the container port (falling back to 8080 when absent). That grep is the file's
only live contract. Everything else is text that has never executed.

This is the "unfalsified gate" shape from `CLAUDE.md`, inverted: not a check that has only ever
passed, but a recipe that has never run and therefore cannot be wrong out loud.

## Proposed fix

Strip the build stage from all 51. Keep `EXPOSE` and the runtime shape, and state in a header
comment that `.github/workflows/Dockerfile.deploy` is the real recipe and that the file is
consumed only for its `EXPOSE` line.

Rationale for not doing the opposite (making all 51 genuinely buildable): it would resurrect a
path the repo deliberately abandoned for documented reasons, and it needs the cold-cache
verification-metadata work as a prerequisite - a large change to a fleet-wide gate file to
support a build nobody runs.

## Acceptance criteria

- [ ] no `openbank-*/Dockerfile` contains a Gradle build stage that cannot run
- [ ] `EXPOSE` preserved in every file, and the port each one reports is unchanged
- [ ] a guard that fails if a per-service Dockerfile regains a build stage, or if one loses its
      `EXPOSE` line (the latter silently degrades every consumer to port 8080)
- [ ] `rules.yaml` records that `Dockerfile.deploy` is the single image recipe

JiRaska:
Closes #3016 · Refs #3015, ADR-0122

## 48, not 51 — and that difference matters

The issue counts all 51 `openbank-*/Dockerfile` as dead. Two are not:

| file | what it actually is |
|---|---|
| `openbank-admin-ui/Dockerfile` | **multi-stage Next.js, built by `openbank-infra/scripts/build-push-admin-ui.sh`** — bakes BUILD_VERSION/BUILD_GIT_SHA, an SBOM-collector stage and test-results.json |
| `openbank-developer-portal/Dockerfile` | static nginx image |
| `openbank-document-renderer/Dockerfile` | WeasyPrint sidecar (ADR-0162 D3) |

**The first pass of this PR stripped `openbank-admin-ui/Dockerfile`, which is a working build.** It was caught only because the branch later conflicted with `main` on that exact file — not by any check I had written.

The reason is worth stating plainly: I wrote the exemption list from *reading the Dockerfiles* rather than from *what builds them*, which is the hand-kept-set failure this repo keeps paying for. The self-test now derives "is really built" from `openbank-infra/scripts/*.sh` and `.github/workflows/*.yml` and fails if any file they name by path is missing from the exemption set.

That derivation immediately over-matched `openbank-product-catalog`: only `Dockerfile.native` is built, and `Dockerfile\b` matches before the dot. Tightened to `Dockerfile(?![.\w])`; comment-only mentions (anacredit's, about the EXPOSE grep) are skipped.

## What the other 49 were

A build stage that could not run: no `COPY build-logic` (the root `settings.gradle.kts` `includeBuild`s it for `openbank.quarkus-service`), no copy of the ADR-0122 `openbank-libs-*` siblings. Nothing builds them — `auto-deploy.yml`, `ghcr-publish.yml` and `build-push-service.sh` all copy `.github/workflows/Dockerfile.deploy` into a generated context and build host-side, reading exactly one thing from the per-service file: its `EXPOSE` line.

Each is now a header explaining what the file is and where the real recipe lives, plus that `EXPOSE`.

## Acceptance criteria

- [x] no `openbank-*/Dockerfile` contains a Gradle build stage that cannot run (48 stripped, 3 genuinely-built exempt)
- [x] `EXPOSE` preserved, and **every port unchanged** — verified per service against `HEAD`: 51 of 51 identical, none missing
- [x] a guard that fails on a regained build stage or a lost `EXPOSE`
- [x] `rules.yaml` records `Dockerfile.deploy` as the single image recipe

## One deliberate deviation

The issue proposes keeping "the runtime shape". I did not. `Dockerfile.deploy` already specifies the base image, user and entrypoint — reproducing them in 49 files would be a second copy of the one thing that *is* built, free to drift, and drift in a file nothing builds cannot be caught. That is the same second-copy failure the issue itself is about. The header points at `Dockerfile.deploy` instead.

## The guard, falsified both ways

`check-service-dockerfiles.py`, gate `service-dockerfiles` (`group: gitops`, `mode: enforced`):

1. no Gradle build stage (`gradlew`, `FROM gradle:`, `--from=build`);
2. exactly one `EXPOSE`.

Rule 2 is the sharper one. The consumers' grep **falls back to 8080** when `EXPOSE` is absent, so losing that line does not fail — it publishes the wrong port silently.

Verified against known positives: restoring a build stage to `ledger-service` prints the finding and exits 1; removing `sca-service`'s `EXPOSE` prints one too; both restored, clean.

The self-test also covers the code-about-code collision this repo keeps hitting: the new header **names** `gradlew` and `build-logic` while explaining their absence, and a naive matcher would flag all 49 files it just created. The check ignores comment lines and the self-test pins that case explicitly.

## Why this PR had no CI for a while

It was `CONFLICTING`, and GitHub builds `pull_request` runs from `refs/pull/N/merge` — a ref that does not exist while a PR conflicts, so **no run is created at all**, for any workflow. `gh pr checks` reports that as `no checks reported`, which reads like a queue delay. Resolving the conflict produced runs within 45 s. Same cause as #3211.

## Note for review

48 Dockerfiles, the guard, `gates.yaml`, `rules.yaml`, and the OPA bundle restamp that a `rules.yaml` edit forces (regenerated as the last step before pushing, per ADR-0029). None of the bundle churn is hand-edited.


JiRaska:
## The "48, not 51" split holds for one of the three, not three

Checked each claimed-live Dockerfile against what actually builds it, rather than against what
references it.

**`openbank-admin-ui` — confirmed live.** `openbank-infra/scripts/build-push-admin-ui.sh:330` is a
real `docker buildx build`, and the script does bake `BUILD_VERSION`/`BUILD_GIT_SHA` and regenerate
`test-results.json` as the PR describes. Catching that on the first pass was the right call.

**`openbank-developer-portal` and `openbank-document-renderer` — not established.** Nothing in
`.github/workflows` or `openbank-infra/scripts` builds either image:

```
grep -lE 'docker.*build|buildx' .github/workflows openbank-infra/scripts   →  no hit for either
```

Their only references are the gitops manifests, and those name an **image**, not a Dockerfile:

```
developer-portal.yaml:25          …/openbank-developer-portal:sandbox-sec1
document-renderer-service.yaml:46 …/openbank-document-renderer:sandbox-ed34a077
```

A running image proves someone built it once; it does not connect that image to the Dockerfile in
this tree. Those two are in exactly the state the PR sets out to end — a recipe nothing runs — just
with a deployed artifact making them look alive. If they are built out-of-band (by hand, or from
another repo), that is worth saying in the file itself, because "there is an image in ECR" is the
evidence a future reader will find and mis-read the same way.

Two ways to settle it, either fine:
- point a workflow at them, which makes the claim true; or
- treat them like the other 48 and record in the header that the image is built out-of-band, naming
  where.

## Correction to my own note

I had recorded "0 of 51 Dockerfiles buildable" from an earlier sweep. `admin-ui` disproves it — the
sweep evidently looked only at the service Dockerfiles. Updated on my side; flagging it because
that number may have informed #3016's framing too.

Unrelated: this PR is currently `DIRTY` against main.

JiRaska:
## Overlaps #3392 — 31 files, different content, opposite conclusion

#3392 (`fix/dockerfiles-runtime-only`, opened 11 hours after this one) rewrites the same per-service `Dockerfile`s against the same #3016 finding. 31 files in common, and the content differs on every one I sampled — so one of these should be closed rather than both merged.

The disagreement is real and worth settling on its merits: this PR leaves a comment plus `EXPOSE` on the grounds that a reproduced runtime stage is *"a copy free to drift invisibly"*; #3392 reproduces it so the file describes the image that is actually built, and adds a gate.

**Your objection survives #3392 as written** — its gate enforces only "no build stage" and "EXPOSE present", so nothing compares the runtime stage it introduces against `Dockerfile.deploy`. Detail and a measurement in [my comment there](https://github.com/JiRaska/open-bank-oss/pull/3392).

*Flagged from the PR queue; branch untouched.*

JiRaska:
## Closing — #3392 is a strict superset on the Dockerfiles

Both PRs rewrite the per-service `Dockerfile`s against the same #3016 finding. Measured rather than assumed, because the file counts point the wrong way (100 vs 54):

```
Dockerfiles touched by #3249   31
Dockerfiles touched by #3392   52
Dockerfiles only in #3249       0
```

Every file this PR fixes, #3392 fixes too, plus 21 more — `admin-ui`, `mcp`, `notification`, `onboarding`, `party`, `pid`, `product-catalog`, `psd2`, `release-steward`, `sanctions` and others.

The 69 files this PR has and #3392 does not are almost entirely **regenerated OPA bundles and service manifests** — the restamp that follows from touching `.github/gates/gates.yaml`, not additional subject matter.

## The one thing not carried over

`check-service-dockerfiles.py`. #3392 ships its own gate, `check-dockerfile-no-build-stage.py`, enforcing two rules (no build stage, `EXPOSE` mandatory). If this PR's checker asserts anything that one does not, that assertion is worth porting — it is the only artifact here without an equivalent there.

## Where the two actually disagreed

Worth recording, since it was a real design question rather than a collision to resolve mechanically. This PR reduced each file to a comment plus `EXPOSE`, arguing a reproduced runtime stage is *"a copy free to drift invisibly"*. #3392 reproduces it so the file describes the image that is actually built.

**Your objection survives the PR that won**: #3392's gate checks only for a build stage and for `EXPOSE`, so nothing compares its runtime stage against `Dockerfile.deploy`. They agree today (same base digest, same `COPY` set, same entrypoint) and nothing keeps them agreeing. I raised that on #3392 — extending the gate to enforce the mirror answers this PR's point rather than overruling it.

Closing as superseded, not as wrong.

PR #3259, opened 2026-08-02
probe: does this content produce CI runs (temporary, will close)

JiRaska:
Temporary diagnostic for the two PRs that received zero workflow runs (#3211, #3249). Same tree as #3249. Will be closed immediately.


JiRaska:
Diagnostic complete — this reproduced the zero-runs symptom of #3211/#3249. Root cause: the PR is CONFLICTING, so GitHub cannot create refs/pull/N/merge and no pull_request workflow ever runs. Closing.

PR #3279, opened 2026-08-02
feat(admin-ui): make the lending console a credit desk, not a table dump

JiRaska:
## What

The lending console was two flat tables of raw enum strings (`PENDING_REVIEW`, `FOUR_EYES`). That
answers *which applications exist* — a question a database answers. It does not answer what an
underwriter, a credit-risk analyst or a lending officer opens a console to find out:

> what is waiting on a decision, how long has it waited, where is the pipeline jammed, and what is
> going wrong in the book

So the primary object is now the **stage**, not the row.

- The ADR-0211 lifecycle drawn as a pipeline: a bubble per stage, **sized** by how much sits in it,
  **tinted** by how long the *oldest* item there has waited (under 24h / 24–72h / over 72h).
- Same animated engine as the service map — `FlowParticle`, `NodeShadow`, `useFlowAnimation` — so
  `prefers-reduced-motion` is honoured and the motion has an explicit pause control.
- Clicking a stage filters the rows beneath it.
- Four headline tiles: book size, applications in flight, anything aging past 72h, loans in trouble.

Stage names come from the generated origination graph, so the picture cannot drift from the machine.
Rows show the human label with the raw state kept as the `title` — the screen and the state machine
can never end up describing different things.

## Two things the first render got wrong

Found by **looking at it**, not by reasoning about it:

| defect | why it mattered |
|---|---|
| terminal stages tinted by age | nine successfully disbursed loans rendered **red**. A loan disbursed four days ago is finished, not four days late — that is a desk being told it has problems it does not have |
| labels truncated, then colliding | `Připraveno k čerpání` → `Připraveno k čer…` is not a name anybody can act on; the two-line fix then overlapped the age line until the age was placed beneath the label block |

## Honesty about the numbers

`/applications/recent` and `/loans/active` are **capped** lists — the server clamps `limit` to
1..100. No figure here may read as a book total: *"12 waiting" when 300 wait is a staffing decision
made on a wrong number.* The cap travels with the count, and a full page says so outright rather
than in a footnote.

Mutations stay absent — decisions, disbursements and write-offs remain in the approval inbox
(ADR-0227 D4).

## Falsification

Each of these turns the suite red (3 of 9):

- re-tinting terminal stages by age
- hiding the cap notice
- counting finished applications as in-flight

## A note on suite cost

Four separate renders of the console pushed **unrelated** files (`card-issue-dialog`,
`service-map-topology`) past their timeouts under parallel load. I checked rather than assumed: they
pass alone on this branch and pass on a clean `origin/main` worktree, so the failures were mine.
Consolidated to two page renders — the suite's cost is shared, so a heavy test is not only its own
problem.

## Verification

From a clean tree (`origination-graph.json`, `governance.json`, `catalog.json` all deleted first):

- `npm test` — **67 files, 983 tests passed**
- `npm run type-check` — clean
- `npm run lint` — 0 errors
- rendered and screenshotted at each iteration

admin-ui only — no API, DB, event or config change.

Refs #3168

### DIFF
--- a/openbank-admin-ui/src/app/lending/page.tsx
+++ b/openbank-admin-ui/src/app/lending/page.tsx
@@ -2,27 +2,39 @@
 // Copyright (c) OpenBank contributors. Licensed under the Apache License, Version 2.0.
 // See LICENSE in the repository root or https://www.apache.org/licenses/LICENSE-2.0 for details.
 
-// ADR-0230 D1: the lending console, read views first — the origination queue (recent
-// applications fleet-wide) and the active-loan portfolio. Mutations deliberately absent:
-// credit decisions, disbursements and write-offs live in the approval inbox (ADR-0227),
-// never as direct writes from this page.
+// The credit desk's working surface (ADR-0230 D1).
+//
+// WHAT WAS WRONG WITH THE OLD SCREEN
+// Two flat tables of raw enum strings. That answers "which applications exist" — a question a
+// database answers. It does not answer the questions an underwriter, a credit-risk analyst or a
+// lending officer actually open a console with:
+//   - what is waiting on a decision, and how long has it waited
+//   - where is the pipeline JAMMED
+//   - how big is the book and what is going wrong in it
+// So the primary object is now the STAGE, not the row, and every number is either actionable or
+// labelled as not-a-total. Rows stay, one level down, filtered by clicking a stage.
+//
+// HONESTY ABOUT THE NUMBERS
+// `/applications/recent` and `/loans/active` are capped lists (the server clamps limit to 1..100).
+// Nothing here may present a capped count as a book total: "12 waiting" when 300 wait is a staffing
+// decision made on a wrong number. The cap travels with the figure, and a full page is called out.
+//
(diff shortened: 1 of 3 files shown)


JiRaska:
Correction to the "note on suite cost" in this PR's description.

I wrote that the `card-issue-dialog` / `service-map-topology` failures "were mine, not
pre-existing". That attribution was not established, and the comparison behind it was uneven: I ran
**two files** on a clean `origin/main` worktree against the **full suite** on this branch. The full
suite is what loads the machine, so the two runs were not measuring the same thing.

Re-checked properly while working on the campaigns console (#3287): a **full** suite run on a clean
`origin/main` worktree fails too — 5 failed / 967 passed, all timeouts, all in unrelated files —
with the host at load average ~80 because a self-hosted Actions runner and Java builds share it.

So the honest position is: those failures are most likely environmental, and this branch's
contribution to them is unproven. Consolidating the four console renders into two was still worth
doing on its own merits (a heavy test spends a shared budget), but it should not be read as a fix
for a regression I demonstrated — I did not demonstrate one.

CI is the arbiter, not the local run.

PR #3284, opened 2026-08-02
feat(infra): give every agent its own budgeted LiteLLM key instead of the shared master key

JiRaska:
## What

Every agent now authenticates to the LiteLLM gateway with **its own budgeted virtual key** instead of the shared master key.

Builds directly on #3276, which restored the database that makes virtual keys possible at all.

## The problem this closes

All four agents used the same credential — and not just any credential. `LITELLM_MASTER_KEY` is the proxy's **admin** key: it mints and revokes keys, it is exempt from every budget, and it was being projected verbatim into four namespaces by ESO.

| question | before | after |
|---|---|---|
| which agent burned the budget? | unanswerable — one key, one spend row | `/spend/logs` splits by `key_alias` |
| stop *that* agent | only lever revokes the whole fleet | revoke one key |
| what enforces a per-caller ceiling? | nothing — only per-**model** ceilings, which cannot tell a runaway scheduler from ordinary interactive use of the same model | LiteLLM refuses the call once the key's daily budget is spent |

None of this produced a failing request, a red dashboard, or a log line. The platform claimed a control it did not have.

## Budgets

Seeded from `openbank-infra/scripts/seed-litellm-virtual-keys.sh`, deliberately layered:

```
devops-agent 3  +  control-liveness-sentinel 3  +  copilot-service 8  +  agent-service 5
  <  per-model ceilings (20 ops / 5 interactive)  <  AiFleetDailySpendHigh (25 USD/day)
```

Each layer bites before the one outside it, so the alert is the last line rather than the first. Budget enforcement is **preventive**; `AiFleetDailySpendHigh` was only ever detective.

## Verified against the running gateway, not asserted

```
devops-agent                 http=200 reply=ok    budget=3.0/1d
control-liveness-sentinel    http=200 reply=ok    budget=3.0/1d
copilot-service              http=200 reply=ok    budget=8.0/1d
agent-service                http=200 reply=ok    budget=5.0/1d
bogus-key                    http=401                          <- negative control
```

The negative control matters: without it a `200` proves only that *something* authenticated, not that the key did. Keys are in OpenBao under `openbank/litellm`; the local capture file was overwritten and removed.

## Two bugs in the seeding script, both found by running it a *second* time

- `/key/list` returns an array of key **strings**, so `.keys[].key_alias` aborted with `Cannot index string with string ("key_alias")`. This could not happen on the run that *created* the keys — an empty `.keys[]` yields nothing and exits 0. The idempotency branch had only ever executed in the one state where it had nothing to do.
- `--rotate` could never work: `/key/generate` refuses a duplicate alias outright (`Unique key aliases across all keys are required`), so rotation must delete first. Same shape — unreachable until a key existed.

Both now exercised against the real gateway in the state they exist for.

## New enforced gate: `agent-virtual-keys`

A master key and a virtual key are the same shape and both authenticate. An agent wired back to the master key therefore *works perfectly* while silently losing its budget, its attribution and its independent revocation — there is nothing for a reviewer to see. That is a gate's job, not a comment's.

Falsified in both directions before wiring in:

- **4** findings against the pre-fix tree, **0** after
- 7 self-test cases: the `ai-platform` exemption (where the master key legitimately lives), a different KV path with the same property name, a non-ExternalSecret document, an empty `data`, a null document
- the runner's **real exit code checked without a pipe** — `1` on a violating tree, `0` on a clean one — so it blocks rather than merely printing `::error::`

## Also

Corrects prose in five manifests that still named `LITELLM_MASTER_KEY`. Stale prose naming a dead identifier is invisible to a text guard forever, so it gets swept with the vocabulary rather than after it.

Refs #3276

### DIFF
--- a/.github/gates/gates.yaml
+++ b/.github/gates/gates.yaml
@@ -159,6 +159,17 @@ gates:
   # PyYAML is not stdlib: try multiple install paths for portability across
   # macOS (system python3 without pip, Homebrew pip3) and Linux ARC runners
   # (pip/pip3 not in PATH). Fall back to checking if pyyaml is already present.
+  - id: agent-virtual-keys
+    name: every agent authenticates with its OWN LiteLLM virtual key, not the shared master key
+    group: gitops
+    mode: enforced
+    run: |
+      # Falsified in both directions before it was wired in: 4 findings against the
+      # pre-fix tree, 0 after, plus 7 self-test cases covering the exemption and the
+      # near-misses. A gate that has only ever passed is unfalsified.
+      python3 .github/scripts/check-agent-virtual-keys.py --self-test
+      python3 .github/scripts/check-agent-virtual-keys.py --enforce
+
   - id: gen-network-policies-drift-gate
     name: "gen-network-policies drift gate (ADR-0081, issue #854)"
     group: gitops
--- a/.github/scripts/check-agent-virtual-keys.py
+++ b/.github/scripts/check-agent-virtual-keys.py
@@ -1,0 +1,133 @@
+#!/usr/bin/env python3
+"""Every agent must authenticate to the LiteLLM gateway with its OWN virtual key.
+
+WHY THIS IS A GATE AND NOT A COMMENT. The master key and a virtual key are the same shape
+(`sk-...`) and both authenticate successfully, so an agent wired to the master key works
+perfectly -- it just has no budget, no attribution, and cannot be revoked without revoking the
+whole fleet. There is no failing request, no red dashboard, and no log line to notice: the only
+symptom is that a control the platform claims to enforce quietly enforces nothing. That is
+precisely the class of regression a text guard can catch and a human review cannot.
+
+THE RULE. In `openbank-infra/gitops/components/`, an ExternalSecret entry whose `remoteRef.key`
+is `litellm` may reference `property: LITELLM_MASTER_KEY` ONLY inside `components/ai-platform/`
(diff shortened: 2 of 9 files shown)


JiRaska:
**Correction to this PR's claim.**

I wrote here that the per-agent budgets are "a preventive control, where `AiFleetDailySpendHigh` was only ever detective". The keys, the budgets and the attribution are all real and verified — but the budgets **could not have fired**, and I did not check that before claiming it.

LiteLLM costs a request from a built-in model-cost table inside the pinned dependency. That table has no entry for `deepseek-ai/DeepSeek-V3.2`, the ops model behind devops-agent, the liveness sentinel and copilot, so every call to it was costed at **$0**:

```
/model/info:  deepseek-ai/DeepSeek-V3.2  in=0  out=0
```

A real 200 response of 13 tokens produced a `/spend/logs` row correctly attributed to `key_alias=devops-agent` with `spend=0.0`. So attribution — the part this PR is actually about — works. Enforcement did not, and neither did the `max_budget: 20` already in git, nor the 25 USD/day fleet alert. All three measure the same meter, and it never moved.

Found by checking whether spend had started to accumulate after this merged, which is the only reason it surfaced: charging zero is not an error condition, so nothing was red anywhere.

Fixed in #3298, which declares both models' prices in git from the providers' own published numbers and adds an enforced gate (`litellm-model-costs`) that treats an explicit zero as a violation rather than a price.

Issue #1759, opened 2026-07-19
WAL archiving broken on ~15 CNPG DBs — #1452/#1444 pod-identity associations declared but never `tofu apply`d


JiRaska:
## Symptom
`PostgresWALArchiveFailing` firing (critical) on multiple CNPG clusters; `ContinuousArchivingFailing` since ~2026-07-17. Barman fails **every** WAL archive:

```
barman-cloud-wal-archive: ERROR: Barman cloud WAL archiver exception: Unable to locate credentials
```

Confirmed affected: `anacredit-db`, `govaudit-db` (governance-auditor), `liveness-db` (control-liveness-sentinel), `product-catalog-db` (accounts). By inspection of `db-backups.tf` the same gap covers the rest of the #1452 + #1444-second-wave list (billing, document-service, authzaudit, devops, docstruth, flakytest, releasesteward, sdd, tpp-registry, vop, …).

**Not affected:** `accounts-db` (original fleet block) archives fine — same IAM role, same bucket. So the role/trust/bucket are healthy; only the newer **Pod Identity associations** are missing in AWS.

## Root cause
`barman-cloud` gets S3 creds via **EKS Pod Identity**, granted per `(namespace, serviceAccount)` by `aws_eks_pod_identity_association` in `openbank-infra/aws/envs/sandbox-platform/db-backups.tf`. #1452/#1444 **added the tf entries** and the matching gitops `barmanObjectStore` + `ScheduledBackup`. The gitops half synced (ArgoCD), so the pods started *attempting* archival — but the terraform half was **never `tofu apply`d**, so the associations don't exist in AWS and the pods have no creds.

Smoking gun: `product-catalog-db-1` has **5d uptime, no restart**, yet began failing exactly at 2026-07-17 15:26 — i.e. when its gitops `barmanObjectStore` synced, not on any pod (re)start. Rules out admission-timing; it is unapplied infra drift.

## Impact
No recovery point for ~15 DBs (incl. regulatory: anacredit / AnaCredit, govaudit). **Not** an imminent outage — WAL disk is low (6–8 segments, <10% on 2Gi PVCs; low-write sandbox DBs) — but every restore-drill would fail and WAL will accrue slowly.

## Remediation
[...]

PR #3290, opened 2026-08-02
fix(infra): campaign-db had no pod-identity association — its backups never worked

JiRaska:
Refs #3071

`PostgresWALArchiveFailing` had been firing ~48h on `campaign/campaign-db`:

```
ContinuousArchiving=False
unexpected failure invoking barman-cloud-wal-archive: exit status 4
archiving write-ahead log file "000000010000000000000001" failed too many times
```

The gitops manifest declares a `barmanObjectStore` (`s3://openbank-sandbox-db-backups/campaign-db`) **and** a `ScheduledBackup`, so Postgres kept trying to archive and kept failing. **Backups were switched on; the permission to perform them never was.**

## Not the failure mode it looked like

There is a prior incident with this exact alert (#1759, 2026-07-19): the association existed and EKS Pod Identity simply missed injecting credentials at pod admission during a node roll — the fix there was a pod restart, and it needed two attempts for three of four clusters.

Checked that first rather than assuming:

- `eks-pod-identity-agent` is **31/31 ready** (in July it was mid-churn)
- restarted `campaign-db-1` anyway — the new pod **still** had no `AWS_CONTAINER_CREDENTIALS_FULL_URI`

Because there was no association to inject. Different cause, different fix.

## Scope — measured, not inferred

```
CNPG clusters with barmanObjectStore : 52
  ContinuousArchiving OK             : 51
  NOT archiving                      : 1   → campaign/campaign-db
```

I first read the Terraform map as holding 13 entries and briefly suspected ~22 databases were unprotected. **That was wrong** — I had `sed`'d a window of the file rather than the whole map, which holds 50. The live measurement is what caught it, before it reached you as a claim.

## The guard for this already existed, and already fired

`check-db-backup-associations` was built by #1760 for precisely this class. Its nightly schedule has been red since the day the gap appeared:

```
failure  2026-08-02 03:29  schedule
failure  2026-08-01 03:29  schedule
success  2026-07-31 05:46  schedule
```

Run against this branch it passes (`✓ every cluster targeting the managed bucket has an association`). Run with the new line removed, it names campaign and prints **the exact entry added here** as the remediation.

So detection was never missing. Two independent controls — the guard and `PostgresWALArchiveFailing` — were both correct and both unread for two days. That is the point-2 thesis in miniature: the problem is not that faults go undetected, it is that detections go unread.

This is also the **fourth** instance of the same gap. `db-backups.tf` already carries a comment about three clusters added by #1444 "never added here, so every WAL archive failed".

## Deploy — merging does not fix it

Terraform only. Two manual steps after merge:

```
tofu apply    # creates the association
kubectl -n campaign delete pod campaign-db-1
```

The restart is required because Pod Identity injects credentials **at admission** and will not retrofit a running pod — the lesson from #1759. Verify afterwards that the new pod actually has `AWS_CONTAINER_CREDENTIALS_FULL_URI` before believing it worked; in July three of four clusters missed injection on the first restart.

### DIFF
--- a/openbank-infra/aws/envs/sandbox-platform/db-backups.tf
+++ b/openbank-infra/aws/envs/sandbox-platform/db-backups.tf
@@ -184,9 +184,25 @@ locals {
     # attempted an archive, so nothing alerted, and they would have had no recovery point the
     # first time anyone needed one. The matching barmanObjectStore + ScheduledBackup + a bounded
     # max_wal_size were added to their gitops manifests in the same change.
-    anacredit        = { namespace = "anacredit", sa = "anacredit-db" }
-    authzaudit       = { namespace = "authz-policy-auditor", sa = "authzaudit-db" }
-    billing          = { namespace = "billing", sa = "billing-db" }
+    anacredit  = { namespace = "anacredit", sa = "anacredit-db" }
+    authzaudit = { namespace = "authz-policy-auditor", sa = "authzaudit-db" }
+    billing    = { namespace = "billing", sa = "billing-db" }
+    # campaign was the ONE cluster in the fleet whose backups had never worked. Its
+    # gitops manifest declares a barmanObjectStore (s3://…/campaign-db) and a
+    # ScheduledBackup, so Postgres kept trying to archive and kept failing —
+    # `barman-cloud-wal-archive: exit status 4`, ContinuousArchiving=False, ~48h of
+    # PostgresWALArchiveFailing. Backups were switched on and the permission to
+    # perform them never was.
+    #
+    # This is NOT the 2026-07-19 failure mode (#1759), where the association existed
+    # and EKS Pod Identity simply missed injecting credentials at admission during a
+    # node roll; there the fix was a pod restart. Checked here first: the agent is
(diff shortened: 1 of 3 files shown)


JiRaska:
Correction to the deploy note above: I wrote "`tofu apply` must run", which reads as a laptop apply. That is wrong for this repo — `platform-tofu.yml` states it explicitly ("is no longer left for a laptop `tofu apply`"). The real path is:

- push/PR → read-only `tofu plan` to the job summary;
- **apply is a manual `workflow_dispatch`** on `main`.

A local attempt confirmed it from the other side: `tofu plan` fails on the required `governance_gh_pat` variable, which is a CI secret I should not be supplying by hand.

Done for this change:
- plan on the merge commit `d65c37b7`: `# aws_eks_pod_identity_association.db_backups_fleet["campaign"] will be created` / `Plan: 1 to add, 0 to change, 0 to destroy.`
- apply dispatched: https://github.com/JiRaska/open-bank-oss/actions/runs/30751506417

The pod restart afterwards is still required and still needs verifying — Pod Identity injects at admission.

Issue #3299, opened 2026-08-02
Actions cache is over its 10 GB ceiling — eviction is the steady state


JiRaska:
## Summary

The repo's Actions cache is at **10.11 GB against a 10 GB ceiling**, across
**75 entries**. It is over the limit, so GitHub is already evicting least-recently-used
entries — continuously, and silently.

Nothing is broken, which is why nobody has noticed: eviction shows up as builds that
re-download dependencies rather than as a red check. The cost is spread across every PR.

## Where the space goes

| key prefix | entries | total |
|---|---|---|
| `gradle-dependencies-v1…` | 6 | 4523 MB |
| `gradle-build-cache-v1…` | 25 | 3584 MB |
| `setup-java-Linux-x64-gradle…` | 1 | 1181 MB |
| `gradle-transforms-v1…` | 4 | 295 MB |
| `node-cache-Linux-x64-npm…` | 1 | 204 MB |
| `node-cache-Linux-arm64-npm…` | 1 | 201 MB |

Top 6 groups account for 96% of the total.

## Why it is worth a look

Every evicted entry is a job that re-resolves its dependencies from scratch. That is slower PR
feedback for everyone, and it makes cache hit rate unpredictable — a build is fast or slow depending
on what else ran that day, which is a bad property for a signal people wait on.

It also quietly weakens any reasoning that assumes a warm cache. That assumption already caused a
real defect in this repo: a regeneration-based check reported "no gaps" on a warm cache while an
entry was demonstrably missing (#3218 / #3233), and the fix was to stop relying on cache state at
all (`--refresh-dependencies`).

## Not a billing problem

To be explicit, since this touches FinOps: the repo is public, so `ubuntu-latest` minutes are free
and unmetered, and GitHub-hosted runners sit outside our VPC, so none of this shows up as NAT
egress. The cost is wall-clock and predictability, not dollars. Worth sizing before spending effort.

## Possible directions

- **Consolidate the `gradle-dependencies-v1` keys.** Several near-identical entries of ~700–800 MB
  suggest one cache per job variant where one shared entry might do.
- **Trim what is cached.** A ~1.2 GB `setup-java` entry is large for a JDK plus wrapper.
- **Shorten retention** for branch-scoped caches; GitHub keeps them 7 days by default and
  short-lived branches keep entries alive well past their usefulness.
- **Measure hit rate first.** None of the above is worth doing blind — the actionable first step is
  finding out how often jobs currently miss.

[...]

Issue #3131, opened 2026-08-01
verification-metadata: measure and close the cold-cache gap for the remaining services


JiRaska:
## Summary

Follow-up to #3017, which is closed as resolved for `openbank-campaign-service` only. Two things
remain: the rest of the fleet is unmeasured, and nothing stops the gap reopening.

## Why today's CI cannot tell you the answer

Runners have warm Gradle caches, and `gradle.properties` sets
`org.gradle.dependency.verification=lenient` — so a missing `verification-metadata.xml` entry is a
**warning**, not a failure. A fully green fleet build is therefore compatible with every service
being broken on a cold cache. #3028 went green on 60 of 60 build checks and that fact carries no
information about cold-cache readiness anywhere.

The gap only becomes visible under two conditions at once: an empty `GRADLE_USER_HOME` **and**
strict verification. Neither holds in CI today.

This is the "green about nothing" shape: `lenient` is precisely what keeps the gap invisible until
the ADR-0144 graduation flips it to strict, at which point every cold-cache build breaks at once —
across the whole fleet, on the same day.

## Scope of the unknown

Measured for one service: **38** missing entries across 31 components. That is the only datapoint.
`openbank-campaign-service` is not obviously representative — services differ in dependency graph
size, and Temporal/ClickHouse/gRPC-heavy modules pull in more.

## Work

- [ ] measure the per-service gap for the remaining services (cold `GRADLE_USER_HOME`,
      `--write-verification-metadata sha256`, diff, take additions only)
- [ ] apply the additions, **additions-only** — a full rewrite reindents the whole file (~18.5k
      lines of churn) and has previously deleted legitimate entries
- [ ] cross-check each new sha256 against Maven Central's published `.sha1` before committing. This
      file is a security control; "add the hash the build showed you" is the failure mode it exists
      to prevent
- [ ] add a scheduled CI job that resolves on a genuinely cold cache under
      `-Dorg.gradle.dependency.verification=strict`, so the gap cannot silently reopen on the next
      dependency bump
- [ ] decide whether that job blocks or merely reports before the ADR-0144 strict graduation

## Practical notes

- `--write-verification-metadata` **OOMs at the default heap** — it died at `Java heap space` after
  20m35s and needs `-Dorg.gradle.jvmargs=-Xmx6g`. "Just regenerate it" is not executable as written.
- **A warm cache cannot be used to scope this.** A warm rerun reported 36 instead of 38 and silently
[...]

PR #3302, opened 2026-08-02
ci(build): nightly verification-metadata sweep + fix the gate on non-Quarkus modules

JiRaska:
### 1. Bug fix for #3233, which is already on `main`

`quarkusDependenciesBuild` only exists on modules applying the `openbank.quarkus-service`
convention plugin. Three modules do not — `openbank-libs-{domain,runtime,temporal}` — and naming
the task for them fails the **entire** Gradle invocation:

```
Cannot locate tasks that match ':openbank-libs-domain:quarkusDependenciesBuild'
as task 'quarkusDependenciesBuild' not found in project
```

So today, any PR touching one of those three build files takes the gate down with it, surfacing as
`verification-metadata check could not run`. Tasks are now selected per module: 57 Quarkus services
get both tasks, the 3 library modules get `testClasses` only — the sole graph they have.

**This was my bug, introduced in #3233.** It survived because #3233 was only ever exercised against
`openbank-campaign-service`, a Quarkus service. The sharded sweep below was the first run that put a
library module into the same invocation, which is how it surfaced.

### 2. The periodic guard (#3131)

The PR gate sees only modules whose `build.gradle.kts` changed; the producer (#3199) runs only on
Dependabot PRs. Neither notices a gap arriving another way — a transitive dependency moving under an
unchanged version range, a platform BOM bump pulling a new artifact, a merge that raced both. This
sweeps the fleet nightly, so such a gap has a ceiling on how long it can sit undetected.

It **files an issue** rather than just going red: an unread red schedule is addressed to nobody,
the same reason `fleet-attestation.yml` escalates.

## Sizing and cost

**8 shards of ~7 modules.** One fleet-wide regeneration needs ~12g of heap (measured: 6g died with
`Java heap space` after 3h40m on the `quarkusBuild` variant), more than a 16 GB runner can safely
hand a single JVM.

Eight rather than four is **free**: the repo is public, so `ubuntu-latest` minutes are unmetered,
and GitHub-hosted runners sit outside our VPC, so no NAT egress is billed.

**No `cache: gradle`, deliberately.** The Actions cache is at **10.36 GB against a 10 GB ceiling
across 67 entries** (#3299) — already evicting. Eight nightly entries would push out caches that PR
builds depend on: slowing everyone else to speed up a job with no deadline. The sweep passes
`--refresh-dependencies` anyway, so a warm cache buys it nothing.

## What is verified

- `actionlint` clean on the new workflow
- 8 shards cover all 60 modules, no overlap, no gaps
- per-module task selection returns `(testClasses, quarkusDependenciesBuild)` for
  `openbank-campaign-service` and `(testClasses,)` for `openbank-libs-{domain,runtime}`
- the underlying script's detection was proven in CI on #3238: it went red naming the unpinned
  `jsoup-1.18.3.jar` / `.pom`

## What is NOT yet verified — stated plainly

The end-to-end run of the **fixed** script across a module set containing a library module has not
completed. Two attempts were killed by the OOM killer on my machine (`exit 143`), which is under
load from parallel agents — that says nothing about a clean runner, but it means I do not have the
local proof in hand yet.

I will post the result as a comment either way. **If it does not detect the deleted entry, this PR
should not merge as-is** and I will say so.

Also unproven: whether a 7-module shard fits a 16 GB runner. `timeout-minutes: 45` and
`fail-fast: false` are set; the first scheduled run is what will tell us. If shards OOM, raising the
shard count further is the cheap lever — it costs nothing on free runners.

## Linked issues

Refs #3131, #3233, #3299

### DIFF
--- a/.github/scripts/check-verification-metadata-complete.py
+++ b/.github/scripts/check-verification-metadata-complete.py
@@ -51,6 +51,22 @@ ARTIFACT_RE = re.compile(r'<artifact name="([^"]*)"')
 # without the cost of a full quarkusBuild — see #3199.
 TASKS = ("testClasses", "quarkusDependenciesBuild")
 
+# `quarkusDependenciesBuild` only exists on modules applying the `openbank.quarkus-service`
+# convention plugin. Three do not — openbank-libs-{domain,runtime,temporal} — and naming
+# the task for them fails the WHOLE invocation with "Cannot locate tasks that match", so a
+# PR touching one of those build files would take the check down with it. They still get
+# `testClasses`, which is the only graph they have.
+QUARKUS_MARKER = "openbank.quarkus-service"
+
+
+def tasks_for(module: str) -> tuple[str, ...]:
+    build_file = pathlib.Path(module) / "build.gradle.kts"
+    try:
+        is_quarkus = QUARKUS_MARKER in build_file.read_text(encoding="utf-8")
+    except OSError:
+        is_quarkus = False
+    return TASKS if is_quarkus else ("testClasses",)
+
 
 def artifact_set(path: pathlib.Path) -> set[str]:
     """Every (component, artifact) pair in the file, as flat 'g:n:v|artifact' keys."""
@@ -79,7 +95,7 @@ def regenerate(modules: list[str]) -> None:
     same outcome as "a gap was found": a crashed build proves nothing either way,
     and reporting it as a gap would send the author chasing entries that are fine.
     """
-    targets = [f":{module}:{task}" for module in modules for task in TASKS]
+    targets = [f":{module}:{task}" for module in modules for task in tasks_for(module)]
     result = subprocess.run(
         ["./gradlew", "--write-verification-metadata", "sha256",
          # Without this the check is vacuous on CI. A warm Gradle cache does not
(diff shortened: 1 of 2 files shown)


JiRaska:
## Temporary commit added to establish the missing proof — in CI, not locally

The last commit is a **comment-only** change to `openbank-libs-domain/build.gradle.kts`. It exists
solely so the `verification-metadata` job picks up a **non-Quarkus** module, which is the exact case
that used to fail the whole Gradle invocation:

```
Cannot locate tasks that match ':openbank-libs-domain:quarkusDependenciesBuild'
```

Why CI rather than my machine: the local run of this proof was killed three times by the OOM killer
(`exit 143`, and once with 14 MB of free memory), on a host loaded by parallel agents. That says
nothing about the fix — but it means I could not obtain the evidence locally. CI is also the more
faithful environment here, since it runs a warm cache with `--refresh-dependencies`, the combination
this check actually depends on.

| result | meaning |
|---|---|
| job passes | per-module task selection works — remove this commit, PR ready for review |
| `Cannot locate tasks…` | the fix does not work; **do not merge** |

**This commit must be removed before merge.** It is marked TEMPORARY in the file, in the commit
message, and here.

JiRaska:
## Proof obtained — temporary commit removed

The `verification-metadata` job ran **with a non-Quarkus module in its set** and passed:

```
changed modules: openbank-libs-domain
check-verification-metadata: regenerating for 1 module(s): openbank-libs-domain
check-verification-metadata: OK — 5168 artifacts, no gaps for the changed modules
```

Both halves matter. The module really was selected — a green job that had silently skipped it would
have looked identical and proved nothing — and there is no `Cannot locate tasks that match`, which
is what this same input produced before the per-module task selection in this PR.

So the bug that is currently on `main` (a PR touching `openbank-libs-{domain,runtime,temporal}`
takes the gate down) is fixed, and demonstrated on the runner rather than argued from the code.

The temporary commit is gone; the PR is back to two files, both under `.github/`.

## Still unproven

Whether a 7-module shard fits a 16 GB runner. This job selected a single module, so it exercised
task selection, not shard sizing. The first scheduled run will answer that; if shards OOM, raising
the shard count is the cheap lever — free on public-repo runners.

PR #3308, opened 2026-08-02
fix(observability): SbomDriftScannerFailing could never clear — it fired on Job history, not the latest run

JiRaska:
Refs #3071

## The alert could never clear

```
expr: kube_job_failed{namespace="admin-ui", job_name=~"sbom-drift-scanner-.*"} > 0
```

A Job object survives in `failedJobsHistoryLimit` long after its pods are garbage-collected, so the series persists and the alert fires until someone deletes the object by hand.

Measured today: firing **critical for 27.8h** off `sbom-drift-scanner-29757790` (failed 2026-07-31 03:10) — while the two runs since then both completed and the scanner reported:

```json
{"ok": true, "driftedServices": [], "noPodServices": []}
```

Its pods were already gone, so the alert pointed at an incident **nobody could investigate any more**, and no operator action could clear it.

That is the worst shape an alert can have: permanently lit, not actionable, and `severity: critical` — so it feeds the "critical is the resting state" problem that makes the alerts which *do* matter get ignored. This one was 3 of the 18 firing criticals.

## The fix

```
last_schedule_time > last_successful_time
```

The most recent scheduled run has not succeeded — and it goes quiet the moment one does. `for: 15m` comfortably exceeds the observed run time (10–20s), so a run merely in flight does not trip it.

**Validated against live metrics before writing it**, not after:

```
sched: 2026-08-02T03:10  (11.6h ago)
succ : 2026-08-02T03:10  (11.6h ago)
proposed expr (sched > succ) = False  → quiet
```

Correct, because the scanner *did* succeed — while the old expression is firing on that same data.

## Note

The sibling `SbomDriftScannerStale` in the same file already used this metric family for exactly this reason ("a scanner that silently stops running is indistinguishable from a clean fleet"). Only `SbomDriftScannerFailing` was left on the sticky one — the right pattern was already in the file, one alert below.


JiRaska:
Closing in favour of a clean branch — this one was cut from my previous feature branch rather than from `origin/main`, so it carried three already-merged commits and, against current main, would have reverted 44 files of other people's work. Replaced by the single-commit version; the change itself is identical.

JiRaska:
Correction to my closing comment: I wrote that this "would have reverted 44 files of other people's work". **That was an overstatement I had not verified.**

Checked afterwards: the branch's copies of e.g. `openbank-infra/gitops/components/ai-platform/litellm.yaml` and `.github/gates/gates.yaml` are byte-identical to `origin/main`. The 44-file list is computed against the merge-base and overstates — a known trap. Real reversion would only occur where content actually diverged, which I did not demonstrate.

What is established, and is reason enough to have closed this:

- the branch carried three commits that were already merged (it was cut from my previous feature branch instead of `origin/main`);
- GitHub refused the merge on genuine conflicts.

The replacement (#3316) is one commit, one file, cut from current `origin/main`.

Worth noting the failure was progressive and silent: #3278 correctly had 2 files, #3290 had 3, #3305 had 4. Those merged without complaint because the carried content matched what was already on main — no-ops. It only surfaced on the fourth branch, once the merge-base had drifted far enough to conflict.

PR #3350, opened 2026-08-02
fix(security-scanner): run Flyway at start — the schema was never created

JiRaska:
## What

`security-scanner` never ran Flyway. Its deployed database has **zero tables**.

## Why

```
$ psql openbank_security -c '\dt'
Did not find any tables.
$ psql openbank_security -c 'select * from flyway_schema_history'
ERROR:  relation "flyway_schema_history" does not exist
```

The service commits `db/migration/V2__create_security_outbox.sql` and `V3`, and has `quarkus-flyway` as a build dependency — but `quarkus.flyway.migrate-at-start` defaults to **false**, it is absent from `application.yaml`, and no deployment env sets it (`env | grep -i flyway` in the pod: nothing). So every write answers `ERROR: relation "security_outbox" does not exist (42P01)` and `/api/v1/security/report` 500s:

```
$ wget --header="Authorization: Bearer <operator>" http://security-scanner-service.security-scanner.svc:8120/api/v1/security/report
HTTP/1.1 500 Internal Server Error
```

Three things kept it invisible: the pod is **Ready** (health probes never touch the schema), the migrations are present and correct in the repo, and the admin-ui console blamed the caller with "scanner HTTP 401" — a *separate* BFF auth defect ([#3336](https://github.com/JiRaska/open-bank-oss/pull/3336)) that masked this one until an authenticated read got far enough to 500.

## How

- `quarkus.flyway.migrate-at-start: true` + `connect-retries`, mirroring sanctions/fraud.
- The datasource is declared in the file with literal local defaults. Flyway migrates the **default** (Agroal/JDBC) datasource, and `check-flyway-default-datasource.py` exists to stop a service that only boots because the environment happened to supply `QUARKUS_DATASOURCE_JDBC_URL` (#3080) — adding `migrate-at-start` without it trips that gate, correctly. (It did, while I was writing this.)
- **The gate gains the inverse rule**, which is the thing that would have caught this: committed migrations must have *something* that runs them — the config key, or `QUARKUS_FLYWAY_MIGRATE_AT_START` in the service's own GitOps manifest, which is `psd2-service`'s deliberate and documented shape and must **not** be flagged. Rule 1 asks "you migrate, but against what?"; rule 3 asks "you wrote migrations, but does anything run them?".

## Verification

- Gate falsified in both directions. Against the pre-fix tree it exits **1** naming `openbank-security-scanner`; the fleet is **clean** after. `run-gates.py --only flyway-default-datasource` → PASS.
- The new rule's selftest drives the **real** rule function rather than re-deriving it, and separately exercises the GitOps lookup against `psd2-service` (must resolve true) and a nonexistent service (must resolve false).
- The rule found `psd2-service` on its first run; investigated and confirmed a **false positive** for "broken" — it injects the env var via GitOps and migrates fine in production — which is why the rule accounts for that path rather than just the config key.

## What happens to the outbox after this

`SecurityOutboxDispatcher` is an **unconditional** `@Scheduled` job — it does not read `openbank.outbox.dispatch-enabled` at all, which is why `gates.yaml` records this service as out of scope of that guard rather than as an exemption. It already runs every 5s today and errors on the missing table:

```
ERROR io.qua.sch.com.run.StatusEmitterInvoker Error occurred while executing task for trigger
  IntervalTrigger [id=1_...SecurityOutboxDispatcher#dispatchScheduledBatch, interval=5000]
Caused by: java.sql.SQLException: ERROR: relation "security_outbox" does not exist (42P01)
```

After this change it will find the table and attempt to publish — and that will fail differently. The service points at Kafka **plain 9092** with no `KafkaUser` and no keystore, so it authenticates as `ANONYMOUS` against a cluster with `authorization: simple`. It is the **only** workload in the fleet on 9092; the other 36 have a `KafkaUser` and use mTLS 9093.

So this is not a regression — one 5s error loop replaces another, and the 500 on `/report` is fixed either way — but the Kafka wiring is a real gap that needs its own change, tracked separately.

## Linked issues

N/A — defect found by walking the deployed console; no issue filed.

### DIFF
--- a/.github/gates/gates.yaml
+++ b/.github/gates/gates.yaml
@@ -1057,7 +1057,7 @@ gates:
   # MAPPING that reads as the fix and is not — that key names which datasource to use, it does
   # not define one. Decidable from one committed file, so no boot and no cluster.
   - id: flyway-default-datasource
-    name: "Flyway default datasource (issue #3080, enforced)"
+    name: "Flyway default datasource, and migrations something actually runs (issue #3080, enforced)"
     group: data
     mode: enforced
     selftest: |
--- a/.github/scripts/check-flyway-default-datasource.py
+++ b/.github/scripts/check-flyway-default-datasource.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 # SPDX-License-Identifier: Apache-2.0
-"""A service with `quarkus.flyway.migrate-at-start` must configure the datasource Flyway migrates.
+"""Flyway must have a datasource to migrate — and committed migrations must have something running them.
 
 Why this exists
 ---------------
@@ -29,6 +29,14 @@ WHAT IT CHECKS, per `openbank-*/src/main/resources/application.yaml`:
      (a `${ENV:default}` expression counts — the literal default is what makes it bootable).
   2. `quarkus.flyway.datasource` must be a NAME (a string), never a mapping. A mapping there is
      the misplaced-definition shape above.
+  3. A service that commits `src/main/resources/db/migration/*.sql` must have SOMETHING that runs
+     them — `quarkus.flyway.migrate-at-start` here, or `QUARKUS_FLYWAY_MIGRATE_AT_START` in its own
+     GitOps manifest (psd2-service's deliberate, documented shape). security-scanner had neither,
+     so its deployed database held zero tables and no flyway_schema_history, and every write
+     answered `relation "security_outbox" does not exist (42P01)` as a 500 — while the pod stayed
+     Ready, because health probes never touch the schema. The inverse of rule 1: that one asks
+     "you migrate, but against what?", this one asks "you wrote migrations, but does anything run
+     them?".
 
 Both are properties of one committed file, so this needs no cluster, no boot and no network.
 It does not replace a real boot test (#2872) — it closes the one failure mode that recurred five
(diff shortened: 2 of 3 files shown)


JiRaska:
Correction, and force-pushed the commit message so the wrong version does not reach `main`.

I originally wrote that `openbank.outbox.dispatch-enabled` is unset here, so "the outbox will never dispatch even once the table exists", and that enabling it would turn a silent no-op into a failing publish loop. **That is wrong on both halves.** `SecurityOutboxDispatcher` is an unconditional `@Scheduled` job that never reads the flag — `gates.yaml` says so explicitly, listing this service as *out of scope* of the dispatch-enabled guard rather than an exemption — and it is already running every 5s in the sandbox, erroring on the missing table.

What is true, and is now in the PR body: after this fix it will find the table and try to publish as `ANONYMOUS` over plain 9092 against a cluster with `authorization: simple`, with no `KafkaUser` and no keystore — the only workload in the fleet not on mTLS 9093. One error loop replaces another; the `/report` 500 is fixed regardless. Filed separately.

I found this only because a later change made me read `gates.yaml` around the dispatch-enabled guard. The original claim came from pattern-matching the fleet convention instead of reading this service's dispatcher — the exact "verify the mechanism you depend on" mistake.

Issue #3354, opened 2026-08-02
fraud-service ONNX has never loaded: glibc-linked native library in a musl (alpine) runtime image


JiRaska:
## Summary

`fraud-service` ships ONNX Runtime (`com.microsoft.onnxruntime:onnxruntime:1.22.0`, ADR-0139 phase-1b) into a **musl** runtime image. The native library is built against glibc, so it has never loaded in any deployed environment — shadow scoring has produced exactly zero scores since the adapter shipped.

Deployed log line:

```
java.lang.ExceptionInInitializerError: Exception java.lang.UnsatisfiedLinkError:
  /tmp/onnxruntime-java.../libonnxruntime.so: Error loading shared library libstdc++.so.6:
  No such file or directory (needed by /tmp/onnxruntime-java.../libonnxruntime.so)
```

The *blast radius* of that failure is fixed separately (the `Error` escaped a field initializer, so the whole bean failed to construct and took `GET /api/v1/fraud/review-queue` down with it). This issue is the remaining half: making the model actually load.

## What was measured

`OrtEnvironment.getEnvironment()` on `linux/arm64`, onnxruntime 1.22.0, three candidate bases:

| Base | Result |
|---|---|
| `eclipse-temurin:25-jre-alpine` (today's `.github/workflows/Dockerfile.deploy`) | `UnsatisfiedLinkError: libstdc++.so.6` |
| `eclipse-temurin:25-jre-alpine` + `apk add libstdc++` | `UnsatisfiedLinkError: ld-linux-aarch64.so.1` |
| `eclipse-temurin:25-jre` (glibc) | **loads** — `OrtEnvironment(name=ort-java, version=1.22.0)` |

The middle row is the important one: **the cheap fix does not work.** Adding `libstdc++` just moves the failure to the glibc dynamic loader. A glibc base is required, not optional.

## Cost

Base image `224MB` (alpine) -> `356MB` (temurin glibc), i.e. **+132MB per service image**.

- ECR storage: each service has its own repository, so the base layer is stored once per repo. ~51 repos x 132MB = **~6.7GB**, at $0.10/GB-month = **~$0.67/month**. Negligible.
- The real cost is not storage but node pull time on cold starts, and a second libc's CVE surface if the fleet ends up split across two bases.

## Options

[...]

PR #3376, opened 2026-08-02
fix(fraud): an ONNX native-load Error 500'd the whole resource, not just scoring

JiRaska:
## What

An `Error` from ONNX Runtime's native loader escaped a **field initializer** in `OnnxFraudModel`, so the bean failed to construct — and took every endpoint sharing its CDI graph down with it, including `GET /api/v1/fraud/review-queue`, a read-only analyst query that never touches a model.

## Why

The class already documents the behaviour it was supposed to have:

> *Session load happens once at construction; a load or verification failure leaves `session` null, so every `scoreShadow` call returns null immediately.*

It could not do that. `loadSession` catches `Exception`, but the failure is `ExceptionInInitializerError` wrapping `UnsatisfiedLinkError` — both `Error`s — and it is thrown one line earlier, by `OrtEnvironment.getEnvironment()` in a field initializer outside any `try`.

From the deployed pod:

```
Caused by: java.lang.ExceptionInInitializerError: Exception java.lang.UnsatisfiedLinkError:
        at com.openbank.fraud.infrastructure.rest.FraudResource.reviewQueue(FraudResource.kt:57)
```

`reviewQueue` is in that stack because the bean is in its graph, not because it uses the model. In the console this rendered as **"fraud-service is unavailable"** on the fraud review queue, against a healthy, running service.

## How

- `loadEnvironment()` catches `Throwable` — the whole point is surviving an `Error` — and returns null. `environment` becomes nullable, `session` follows, and `scoreShadow` already returns null when either is missing.
- `loadSession` widens to `Throwable` for the same reason: `createSession` also crosses into native code.

Construction can no longer throw. Shadow scoring degrades to rules-only, which is what the KDoc always claimed and what ADR-0139 phase-1b's fail-closed floor intends.

**No fraud verdict changes.** `scoreShadow` is shadow-only, and the bundled model encodes the same deterministic logistic the previous pure-Kotlin `BaselineFraudModel` computed. What is lost is only the shadow signal — which was already absent, since the library never loaded.

## Verification

Three new cases in `OnnxFraudModelTest`, each failing against the pre-fix code (the existing "load failure degrades" case only covered a missing *resource*, i.e. an `Exception`):

- `OrtEnvironment.getEnvironment()` mocked to throw `ExceptionInInitializerError(UnsatisfiedLinkError)` → `loadEnvironment()` returns null rather than propagating.
- Same mock → **constructing** `OnnxFraudModel` does not throw, and `scoreShadow` returns null.
- An `OrtEnvironment` whose `createSession` throws `UnsatisfiedLinkError` → `loadSession` returns null.

## Why the runtime image is not changed here

Measured on `linux/arm64`, onnxruntime 1.22.0:

| Base | Result |
|---|---|
| `eclipse-temurin:25-jre-alpine` (today's `Dockerfile.deploy`) | `UnsatisfiedLinkError: libstdc++.so.6` |
| `eclipse-temurin:25-jre` (glibc) | **loads** |

The obvious cheap fix does not work — adding `libstdc++` just moves the failure to the glibc loader — so making ONNX actually load needs a glibc base, which is a fleet-wide image decision with its own risk and its own verification. That is [#3354](https://github.com/JiRaska/open-bank-oss/issues/3354), with the measurements and a cost breakdown.

This PR is worth having independently of that one: a corrupt library or an OOM at load time must never be able to take a read endpoint down with it.

## Local gate

`detekt`, `ktlintCheck`, `test`, `quarkusBuild`. `OnnxFraudModelTest` is **10/10**. One unrelated flake appeared in the full run — `FraudPactFolderProviderVerificationTest` failed to boot with `QuarkusBindException` (port already bound, from another Gradle build running concurrently on this machine); re-run in isolation it passes, `tests=1 failures=0`.

## Linked issues

Refs #3354 — the runtime-image half of the same defect.

### DIFF
--- a/openbank-fraud-service/src/main/kotlin/com/openbank/fraud/infrastructure/ml/OnnxFraudModel.kt
+++ b/openbank-fraud-service/src/main/kotlin/com/openbank/fraud/infrastructure/ml/OnnxFraudModel.kt
@@ -36,6 +36,11 @@ import java.security.MessageDigest
  * Session load happens once at construction; a load or verification failure leaves [session] `null`,
  * so every [scoreShadow] call returns `null` immediately — the ADR-0139 fail-closed floor applies to
  * model *loading* (and now verification), not just per-call inference failures.
+ *
+ * "Failure" includes the native library not loading at all, which is an `Error` and not an
+ * `Exception` — see [loadEnvironment]. Construction of this bean must never throw: it is injected
+ * into the same graph as read-only endpoints that do not use a model, and a bean that fails to
+ * construct takes all of them down with it.
  */
 @ApplicationScoped
 class OnnxFraudModel(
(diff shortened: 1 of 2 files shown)


JiRaska:
Force-pushed one amend: `ktlintMainSourceSetCheck` was red on the first push. Renaming the local to `activeEnvironment` had pushed the `createTensor` call to 123 chars (limit 120), so ktlint split it into ~10 findings. Shortened to `env`.

Worth recording how it got past me: the local gate stops at the first failing task, and `test` ran before `ktlintCheck` — the `test` failure was itself unrelated (a `QuarkusBindException` from a second Gradle build competing for the port on the same machine), so ktlint never executed locally at all. A green-looking task list is not the same as a task list that ran.

Re-verified after the amend: `ktlintCheck` + `detekt` clean, `test` 130/130.

Issue #3266, opened 2026-08-02
domestic-payment: the screening-unavailable path bypasses applySddPolicy, so an outage holds an INTERNAL_CLIENT payment that a real POTENTIAL_HIT would release


JiRaska:
### Affected service / module

openbank-domestic-payment

### Version / commit

main @ 63b6cd8dd (observed live in sandbox, 2026-08-02)

### What happened?

For `transferScope = INTERNAL_CLIENT`, `DomesticPaymentActivitiesImpl.applySddPolicy` deliberately
downgrades a screening `REVIEW` to `CLEAR` — the bank holds full KYC on both parties, so a
`POTENTIAL_HIT` is not actionable (AMLD4 Art. 15–17, simplified due diligence).

But the `ScreeningUnavailableException` handler returns `REVIEW` **directly from the `catch`
block**, before `applySddPolicy` is ever reached:

```kotlin
} catch (ex: ScreeningUnavailableException) {
    log.warnf(ex, "Sanctions screening unavailable for payment %s; returning REVIEW", paymentId)
    openCaseQuietly(payment, AmlCaseRiskLevel.MEDIUM, ALERT_SCREENING_UNAVAILABLE, ex.message, null)
    return@vtx ScreeningDecision.REVIEW          // <- bypasses applySddPolicy
}

val decision = applySddPolicy(payment, ScreeningPolicy.decide(results), paymentId)
```

The policy is therefore inverted in practice for internal client-to-client transfers:

| situation | outcome |
|---|---|
| a real `POTENTIAL_HIT` against a sanctions list | downgraded to `CLEAR` — payment proceeds |
| sanctions-service unreachable, nothing known | held in `REVIEW` — payment stranded |

An infrastructure outage is treated as *more* suspicious than an actual name match. This is what
stranded payment `c05853f8-4259-4c97-b648-9c8a84082789` on 2026-08-02 and left AML case
`856e4839-c3ce-4e3b-a268-5fbfe1842975` (`SCREENING_UNAVAILABLE`, `OPEN`) behind it.

### What did you expect?

The SDD determination is a property of the payment's `transferScope`, so it should apply to every
path that produces a hold — including the unavailability path — rather than only to the path where
a screening result exists.

**Do not simply route the `catch` through `applySddPolicy`.** That would silently release internal
payments un-screened whenever sanctions-service is down, which is the failure mode the fail-closed
[...]

Issue #3264, opened 2026-08-02
sanctions-service: idempotency_key is a UNIQUE constraint, not idempotency — a replay returns 500 and strands the payment


JiRaska:
### Affected service / module

openbank-sanctions-service

### Version / commit

main @ 63b6cd8dd (observed live in sandbox, 2026-08-02)

### What happened?

`sanctions_checks.idempotency_key` carries a UNIQUE constraint but there is **no idempotent-replay
path**. A caller that resends a previously-seen key does not get the stored check back — the insert
hits the constraint, `GenericExceptionMapper` turns it into an unhandled 500, and the caller sees a
transport failure for a check that in fact already succeeded.

This is not theoretical. It stranded a real sandbox payment on 2026-08-02:

1. `openbank-domestic-payment` screened the debtor with key `<paymentId>:debtor`.
2. sanctions-service completed the check and **persisted `status = CLEAR`** at 10:58:10Z.
3. The caller's `@Timeout(5_000)` had already fired, so its `@Retry` replayed the same key.
4. sanctions-service answered 500 (duplicate key) instead of replaying the stored `CLEAR`.
5. The repeated failures opened the caller's circuit breaker; the **creditor was never screened at
   all** (no row in `sanctions_checks`), and the payment was held fail-closed in `RECEIVED`.

So an *idempotency* mechanism turned a successful, retryable check into a permanent hold. The
payment is one of 7 stuck in `RECEIVED`, the oldest since 2026-06-23.

Caller-side defects are filed separately — this issue is the server-side contract.

### What did you expect?

A repeated `idempotency_key` returns **200 with the stored `SanctionsCheck`** (the standard
idempotent-replay semantics this codebase uses elsewhere, e.g. the shared idempotency store in
`openbank-libs-runtime`). Failing that, at minimum a typed 409 the client can distinguish from a
transport fault — never an unhandled 500.

Worth deciding explicitly as part of the fix: whether a replay must also match the request payload
(same name / entityType) and answer 422 on a key collision with different content, rather than
silently returning a check for a different subject.

### Reproduction steps

1. `POST /api/v1/sanctions/checks` with `idempotencyKey = K`, any name. Observe 2xx and a row in
   `sanctions_checks`.
2. `POST` the identical body again with the same `idempotencyKey = K`.
3. Observe 500 instead of the stored result.

Live evidence in sandbox: `sanctions_checks` holds exactly one row for key
[...]

Issue #3208, opened 2026-08-02
auto-retry re-queues a deliberate cancel, so a backlog drain silently needs two passes


JiRaska:
## Summary

`auto-retry-cancelled.yml` re-runs a **deliberately cancelled** run, so draining a queue backlog by
hand needs **two passes** — and nothing tells the operator the first pass was undone.

This is not a claim that the heuristic is wrong. The workflow header already states the trade-off
explicitly:

> Deliberately NOT retried: `action_required` … and runs cancelled by anything other than the runner
> dying — indistinguishable from the outside, so we accept one wasted retry on a manually cancelled
> run over a more clever heuristic that can misfire.

I checked whether a cheap discriminator exists and **it does not**. My first hypothesis — "a run
cancelled while queued never had a runner, so it cannot have been reclaimed" — is false: the runs I
cancelled had dozens of *completed* jobs and only the tail of the service matrix still queued, so
the run reads `queued` at run level while most jobs already ran on real runners. Distinguishing a
human cancel from a reclaim would need the job logs (`The runner has received a shutdown signal`),
which this workflow avoids on purpose.

So the design is sound. What this issue is about is that **the accepted cost was priced for an idle
pool**, and the case where someone cancels is precisely the case where the pool is not idle.

## What happened (2026-08-02, ~04:00 UTC)

The ARC pool had a 19-hour `Services CI` backlog — 23 queued runs, oldest created `09:17Z`, against
~6 runner slots (the capacity problem is #2039). This blocks `can-i-deploy` for
`openbank-lending-service`, because the consumer pact for a commit is published by that service's own
main-push build, which was sitting in that queue.

Cancelled the 19 stale runs (every one for a commit already on `main`). `gh run cancel` returned 0
for all 19. Within seconds **all 19 were back in the queue**:

```
$ gh api repos/JiRaska/open-bank-oss/actions/runs/30693441813 --jq '.status, .run_attempt'
queued
2
```

Six `Auto-retry spot-killed CI runs` runs fired between `04:03:16Z` and `04:03:32Z`, one per
cancelled run, logging:

```
Re-running cancelled run https://github.com/JiRaska/open-bank-oss/actions/runs/30697897176 (issue #2330)
Re-ran cancelled run ... (attempt 2 of max 2; a second kill stays for a human)
```

A second cancel pass stuck, because `run_attempt == 1` is the gate and attempt 2 is the ceiling.
Queue went 23 → 7.

[...]

PR #3377, opened 2026-08-02
fix(libs,domestic): surface the transport fault a resilient call masks

### LINKED ISSUE #3267: Resilient adapters: @Retry over @CircuitBreaker discards the original transport error — only CircuitBreakerOpenException is ever logged
### Affected service / module

openbank-domestic-payment (pattern is fleet-wide — see Notes)

### Version / commit

main @ 63b6cd8dd (observed live in sandbox, 2026-08-02)

### What happened?

When a resilient inter-service adapter degrades, the **original transport error is never logged**.
Only `CircuitBreakerOpenException` reaches the log, and it names nothing about what actually failed.

MicroProfile Fault Tolerance applies the annotations in a fixed order —
`Fallback > Retry > CircuitBreaker > Timeout > Bulkhead` — so `@Retry` sits *outside*
`@CircuitBreaker`. Once the breaker opens mid-retry, the remaining attempts fail with
`CircuitBreakerOpenException`, and that is the exception `@Retry` ultimately propagates. The real
cause (a 500, a timeout, a connection refusal) is discarded with the earlier attempts.

`SanctionsScreeningAdapter.screen` then wraps whatever surfaced in `ScreeningUnavailableException`,
and the caller logs that. The result, for a payment stranded on 2026-08-02, was a log that says the
breaker is open and nothing else — **six hours of domestic-payment logs contained exactly one line
about sanctions**, and it did not name the fault.

Diagnosing it required reading a different service's log to find the actual error
(`duplicate key value violates unique constraint "sanctions_checks_idempotency_key_key"`, filed
separately). Nothing in the owning service pointed there. On a rail where a screening failure holds
customer money, that is the wrong place for the evidence to live.

### What did you expect?

The first failing attempt is logged at WARN with its own exception, before any retry or breaker
state change masks it. At minimum:

- log inside the adapter on each failed attempt (or via a `@Fallback` / `FaultToleranceMetrics`
  handler that sees the underlying throwable), not only at the outer catch;
- keep the original cause attached when wrapping into `ScreeningUnavailableException`, so the chain
  printed at the call site ends in the transport fault rather than in `CircuitBreakerOpenException`;
- emit a metric distinguishing *breaker open* from *call failed*, so the two are separable in
  Grafana without log spelunking.

### Reproduction steps

1. Make sanctions-service return 500 for `POST /api/v1/sanctions/checks`.
2. Submit domestic payments until `@CircuitBreaker(requestVolumeThreshold = 4, failureRatio = 0.5)`
   opens (one payment is enough: 2 names x up to 3 attempts).
3. Read domestic-payment's log. Observe that the 500 appears nowhere — only
   `CircuitBreakerOpenException`.

### Logs / stack traces

```shell
# the ONLY sanctions-related line in 6h of domestic-payment logs:
2026-08-02 10:58:18.041 WARN com.ope.dom.app.wor.DomesticPaymentActivitiesImpl
  Sanctions screening unavailable for payment c05853f8-4259-4c97-b648-9c8a84082789; returning REVIEW
  com.openbank.domestic.application.port.out.ScreeningUnavailableException: Sanctions screening is unavailable; payment held
  Caused by: org.eclipse.microprofile.faulttolerance.exceptions.CircuitBreakerOpenException:
    com.openbank.domestic.infrastructure.client.SanctionsScreeningAdapter#screenWithResilience circuit breaker is open

# what actually happened, only visible in the sanctions-service log:
2026-08-02 10:58:15.463 WARN org.hib.orm.jdb.error
  ERROR: duplicate key value violates unique constraint "sanctions_checks_idempotency_key_key" (23505)
```

### Severity

S3 — degraded service with workaround

### Notes

- Source: `openbank-domestic-payment/src/main/kotlin/com/openbank/domestic/infrastructure/client/SanctionsScreeningAdapter.kt`.
- Not specific to sanctions or to domestic-payment: any adapter stacking `@Retry` over
  `@CircuitBreaker` has this blind spot. Sweep the sibling adapters (`AmlCaseAdapter`,
  `FraudScoringAdapter`, `SettlementAdapter`, `SchemeGatewayAdapter`) and the equivalents in the
  other payment rails; if the pattern is uniform, the fix belongs in a shared helper in
  `openbank-libs-runtime` rather than per-adapter.
- Falsifiability: a test asserting "the adapter throws `ScreeningUnavailableException`" passes
  today and would still pass after the fix. The assertion has to be on **what was logged** — that
  the underlying cause appears — or the change is unverified.

JiRaska:
Closes #3267 · Refs #3264, #3266

## The masking

`@Retry` sits **outside** `@CircuitBreaker` (fixed MicroProfile order: `Fallback > Retry > CircuitBreaker > Timeout > Bulkhead`). Once the breaker opens mid-retry, the remaining attempts fail with `CircuitBreakerOpenException` — and that is what `@Retry` finally propagates. The 500, the timeout, the connection refusal is discarded with the earlier attempts.

The cost, from the issue: a domestic payment held on 2026-08-02, **six hours of the owning service's logs containing exactly one line about sanctions**, naming nothing. The real cause (`duplicate key value violates unique constraint "sanctions_checks_idempotency_key_key"`) was only findable in a *different* service's log. On a rail where a screening failure holds customer money, that is the wrong place for the evidence to live.

## The fix, and where it goes

`reportingFirstFailure` wraps the call body **inside** the annotated method, so it runs per attempt and inside the breaker. Every attempt reports its own fault as it happens, and the throwable is rethrown unchanged so fault tolerance still counts it — swallowing here would turn a held payment into a released one.

The issue asks whether this belongs in a shared helper. Measured, and it does:

| | count |
|---|---|
| files stacking `@Retry` over `@CircuitBreaker` | **57** |
| of those, mentioning a warn/error log at all | **7** |

So it lives in `openbank-libs-runtime`. `SanctionsScreeningAdapter` — the adapter the issue names — is the first consumer and the proof. The other 56 are a follow-up sweep, deliberately **not** a blind 57-file refactor landed inside a money-path PR.

## Why the handler is a parameter, not a logger

This is the part the issue is sharpest about:

> Falsifiability: a test asserting "the adapter throws `ScreeningUnavailableException`" passes today and would still pass after the fix. The assertion has to be on **what was logged** — that the underlying cause appears — or the change is unverified.

Passing the handler in makes that contract assertable directly, with no log capture and no fault-tolerance runtime. `FirstFailureVisibilityTest` pins:

- the **original** throwable reaches the handler (asserted on the `sanctions_checks_idempotency_key_key` message, i.e. the exact fault that was invisible);
- it is rethrown **unchanged** (`isSameAs`) — otherwise `@Retry`/`@CircuitBreaker` would stop counting the failure;
- a success reports nothing;
- an `Error` is reported too.

## `Throwable`, not `Exception`

Deliberate, and the reason is in the code: a `@Timeout` surfaces as an `InterruptedException` on some paths and a breaker rejection is an unchecked `RuntimeException`. Narrowing the catch would silently re-create the blind spot for exactly the faults this exists to reveal. That is also why detekt's `TooGenericExceptionCaught` is suppressed here with a written reason rather than worked around.

## The metric — all three asks are now covered

`ResilientCallMetrics` emits `openbank.resilient.call.failures{adapter,outcome}`, counted **per attempt** from inside the `@CircuitBreaker`-annotated method, so the numbers line up with what fault tolerance actually saw rather than with the single exception that escaped.

Why the two outcomes must be separable, beyond "the issue asked": they mean opposite things. `call_failed` is the dependency misbehaving and is what you page on; `breaker_open` is this service protecting itself, and is a *consequence* of an earlier burst of `call_failed`. A dashboard summing them shows a flat plateau of "sanctions failing" while the dependency may already have recovered — the breaker is simply still open.

Shape follows `LlmCallMetrics`/`DomainMetrics`: `Instance<MeterRegistry>` rather than direct injection, so the bean loads harmlessly in a service with no `quarkus-micrometer` on its classpath and every method is a silent no-op — pinned by a test. A plain bean, **not** a `@Produces`, which would drag Micrometer into the Arc type closure of every consumer of `openbank-libs-runtime` including those that make no resilient call.

Both tags are closed low-cardinality sets: `outcome` is exactly `breaker_open` or `call_failed`, and `adapter` is a compile-time constant per call site — never a URL or an id, which on the money path would be a cardinality incident.

Injected as a **field**, not a constructor parameter: detekt's `LongParameterList` fires AT the threshold, and `@Inject lateinit` is the fleet convention for adding a metrics port to an existing adapter (`LoanStageEventConsumer`, `VopRateLimitFilter`, `McpEndpoint`).

## What this does not do

- **No change to the sibling adapters.** Named above with the measured count so the follow-up is scoped rather than guessed.

## Note for review

This touches `openbank-domestic-payment`, so it is money-path and wants the two approvals per `rules.yaml` / ADR-0030. The behavioural surface is one added WARN per failed attempt; the screening semantics, the fail-closed mapping to `ScreeningUnavailableException`, and the annotation stack are untouched.

`ktlintCheck`, `detekt`, `test` green for `openbank-libs-runtime` and `openbank-domestic-payment`, plus `quarkusBuild` on domestic-payment — CDI wiring is not validated by ktlint or unit tests, and a new `@Inject` field is exactly what that build catches.

### DIFF
--- a/.github/workflows/auto-retry-cancelled.yml
+++ b/.github/workflows/auto-retry-cancelled.yml
@@ -55,9 +55,19 @@
 # (run_attempt 2) never re-triggers, and a run cancelled for a real reason (superseded by a
 # newer push to the same PR — the concurrency group's normal operation) is only retried once,
 # where it fails fast if the branch has moved. Deliberately NOT retried: `action_required`
-# (approval gates are a human's job) and runs cancelled by anything other than the runner
-# dying — indistinguishable from the outside, so we accept one wasted retry on a manually
-# cancelled run over a more clever heuristic that can misfire.
+# (approval gates are a human's job).
+#
+# This paragraph used to end "and runs cancelled by anything other than the runner dying —
+# indistinguishable from the outside, so we accept one wasted retry on a manually cancelled run".
+# HALF of that is now false and the correction cost one API call: a reclaim kills a RUNNING job,
+# so a run whose cancelled jobs never started a step cannot have been reclaimed. That case is now
+# skipped (see the `cancelled` branch below). The other half stands — a human who cancels WHILE
+# jobs are running is still indistinguishable, and is still retried once.
+#
+# The accepted waste was priced for an idle pool, and the case where someone cancels is precisely
+# the case where the pool is NOT idle: draining a 19-hour, 23-run backlog put all 19 runs straight
+# back in the queue, so the drain silently needed two passes and the re-added load landed on
+# everyone else's PRs (#3208, capacity itself is #2039).
 #
 # `-R` ON EVERY `gh run rerun` IS LOAD-BEARING (issue #2841 follow-up)
 # This job has no `actions/checkout`, so there is no git repository for `gh` to infer the
@@ -105,11 +115,32 @@ jobs:
           set -euo pipefail
 
           if [ "${CONCLUSION}" = "cancelled" ]; then
-            # Whole run cancelled. Indistinguishable from the outside between a reclaim and a
-            # human/concurrency cancel, so retry once either way — the accepted waste that
-            # LOOP SAFETY above describes. `--failed` is a no-op against `cancelled`, so this
-            # is always the full re-run.
-            echo "Re-running cancelled run ${RUN_URL} (issue #2330)"
+            # A reclaim kills a RUNNING job. So if not one cancelled job ever started a step,
+            # nothing was running and a reclaim is impossible — this was a queue cancel, and
+            # re-running it puts back exactly the load whoever cancelled was removing (#3208).
+            #
+            # `steps` is empty for a job that was still queued when the run was cancelled;
+            # a job interrupted mid-flight carries its steps with their conclusions. Verified on
+            # both sides against real runs: 30693441813 (operator drain) has 5 cancelled jobs and
+            # 0 with steps, while contemporaneous cancels that hit running jobs report 1-2 with
+            # steps. Attempt-scoped for the same reason the failure branch below is.
+            #
+            # ONE-DIRECTIONAL on purpose. This only rules a reclaim OUT; it never claims one
+            # happened. A human who cancels while jobs are running still looks exactly like a
+            # reclaim and is still retried once — unchanged, and still the accepted waste the
+            # LOOP SAFETY note describes. What it removes is the case the header never priced:
+            # draining a deep queue, where every cancelled job is one that never started, and the
+            # retry silently undid the drain (#3208, #2039).
+            RUNNING="$(gh api "repos/${GITHUB_REPOSITORY}/actions/runs/${RUN_ID}/attempts/1/jobs?per_page=100" \
+                        --jq '[.jobs[] | select(.conclusion == "cancelled") | select((.steps | length) > 0)] | length')"
+            if [ "${RUNNING:-0}" -eq 0 ]; then
+              echo "::notice title=spot-kill auto-retry::NOT re-running ${RUN_URL} — no cancelled job had started a step, so no runner was reclaimed. Treating it as a deliberate cancel (#3208)."
+              exit 0
+            fi
+
+            # At least one job was interrupted mid-flight: a reclaim is possible, so retry once.
+            # `--failed` is a no-op against `cancelled`, so this is always the full re-run.
+            echo "Re-running cancelled run ${RUN_URL} (${RUNNING} job(s) interrupted mid-flight; issue #2330)"
             gh run rerun -R "${GITHUB_REPOSITORY}" "${RUN_ID}"
             echo "::notice title=spot-kill auto-retry::Re-ran cancelled run ${RUN_URL} (attempt 2 of max 2; a second kill stays for a human)"
             exit 0
--- a/openbank-domestic-payment/src/main/kotlin/com/openbank/domestic/infrastructure/client/SanctionsScreeningAdapter.kt
+++ b/openbank-domestic-payment/src/main/kotlin/com/openbank/domestic/infrastructure/client/SanctionsScreeningAdapter.kt
@@ -9,6 +9,7 @@ import com.openbank.domestic.application.port.out.ScreeningUnavailableException
 import com.openbank.domestic.domain.screening.ScreeningMatchStatus
 import com.openbank.domestic.domain.screening.ScreeningResult
 import com.openbank.domestic.domain.screening.ScreeningRole
+import com.openbank.libs.observability.reportingFirstFailure
 import io.smallrye.mutiny.coroutines.awaitSuspending
 import jakarta.enterprise.context.ApplicationScoped
 import jakarta.inject.Inject
@@ -16,6 +17,7 @@ import org.eclipse.microprofile.faulttolerance.CircuitBreaker
 import org.eclipse.microprofile.faulttolerance.Retry
 import org.eclipse.microprofile.faulttolerance.Timeout
 import org.eclipse.microprofile.rest.client.inject.RestClient
+import org.jboss.logging.Logger
 
 /**
  * Resilient, fail-closed adapter over [SanctionsServiceClient] (ADR-0032 §C/§D). Transport faults are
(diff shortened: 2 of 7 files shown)


JiRaska:
I filed #3267 and hit this defect twice in one incident, so I came to fix it and found this PR
instead. It covers what the issue asked for, and the two load-bearing counts hold up — I derived
them independently: **57** files under `openbank-*/src/main/kotlin` stack `@Retry` over
`@CircuitBreaker`, **7** of those mention a `log.warn`/`log.error` at all. Handler-as-parameter is
also the right call for the falsifiability point; asserting on captured log output would have been
a weaker test.

One thing to split out, and one optional follow-up.

## `.github/workflows/auto-retry-cancelled.yml` does not belong in this PR

Confirmed with a 3-dot diff (`origin/main...origin/fix/resilient-adapter-first-failure`), so this is
a real change on the branch and not a stale-base artifact. It rewrites when a cancelled run is
re-run — skipping runs whose cancelled jobs never started a step — and cites #3208 and #2039.

That is a good change. It is also unrelated to fault-tolerance logging, and bundling it here costs
something specific: a reviewer drawn in by "Closes #3267" is reviewing a fault-tolerance helper, and
the CI-retry logic rides along on that attention. It changes behaviour for *everyone's* PRs, not
just this rail, and workflow edits in this repo have their own failure mode — an oversized `run:`
step makes the whole file unparseable with no error anywhere (#3135/#3139). It deserves its own PR
and its own reviewer, not a paragraph inside a money-path-adjacent one.

Worth noting the irony while it is fresh: this PR currently shows `issue-hygiene` and
`Detect changed services` as **CANCELLED** — plausibly the exact case the workflow change is about.

## Optional: a latched breaker is still only visible when something calls

Evidence from the incident, already on #3267: the breaker stayed open **2h15m** despite
`delay = 10_000`, and only a pod restart cleared it. `openbank.resilient.call.failures{outcome=
"breaker_open"}` does surface that — it increments on every attempt — so this is not a gap in what
the issue asked for, and I would not block on it.

The residual case is a latched breaker on a *quiet* path: no calls, no counter movement, and the
breaker is indistinguishable from healthy until the next request finds it open. A state gauge would
answer "is this breaker open right now" without having to make a call to find out. Follow-up at
most, and only if the sweep over the other 56 call sites happens anyway.

JiRaska:
## This PR's own red run is the case its new skip declines to retry

The `Services CI` run that left this PR at `all-green [failure]` was **cancelled**, not failed, and I measured it before re-running:

```
run 30758989721, attempt 1, head 9ad239698 (unchanged since)
  Detect changed services      cancelled   steps = 0
  nightly fleet-health alert   cancelled   steps = 0
  verification-metadata        success     steps = 8
  all-green                    failure     "Verify no service failed"

cancelled jobs with steps started : 0
```

The new guard in this PR is exactly:

```bash
RUNNING="$(… select(.conclusion == "cancelled") | select((.steps | length) > 0) …)"
if [ "${RUNNING:-0}" -eq 0 ]; then
  echo "::notice … NOT re-running … Treating it as a deliberate cancel (#3208)."
```

`.steps | length` is 0 for both cancelled jobs here, so once this merges, a run in the state this PR was in gets skipped, logged as a deliberate cancel, and left red. I re-ran it by hand; nothing would have.

## Why I am reporting it rather than objecting to it

The inference is correct — a reclaim kills a *running* job, so 0 steps genuinely rules out a spot-kill, and not wasting a retry on a human's deliberate cancel is the right call. The gap is that "not a spot-kill" and "deliberate" are not the same set, and this run is in the difference: the head SHA never moved, nobody mentioned cancelling it, and it died 17 seconds in with `verification-metadata` already green.

I cannot establish what cancelled it, so I am not claiming a defect in the heuristic — only that a third case exists between the two it distinguishes, and this PR is a live specimen of it.

## Interaction with #3395

#3395 keys push-to-main concurrency by SHA so main runs stop cancelling each other, and **deliberately keeps the superseding behaviour on PR branches** ("a newer push SHOULD cancel the old run"). So it does not cover this: a PR-branch run cancelled in its concurrency lane still lands here, and after this PR merges nothing retries it.

Worth deciding whether the notice should be an escalation rather than an `exit 0` — a cancelled run that leaves a required context red is not self-healing, and the only signal today is a check nobody is watching.

## On the change itself

No objection to the substance. `FirstFailureVisibility` and `ResilientCallMetrics` land in `libs-runtime` as a plain `@ApplicationScoped` bean and explicitly **not** a `@Produces` — with the reason in place ("a producer would drag Micrometer into the Arc type closure"), which is the fleet-wide footgun `CLAUDE.md` documents. The new dependency is `testImplementation` only, so nothing changes on any service's runtime classpath.

*Reviewed while draining the queue; I re-ran the cancelled run but did not touch the branch.*

Issue #3240, opened 2026-08-02
Three check-* scripts outside gates.yaml: one never runs at all, one is a ci_producer CI never invokes


JiRaska:
## Summary

`CLAUDE.md` says gates are **declared** in `.github/gates/gates.yaml`, not written as workflow steps, and that the manifest is what makes `mode:` (advisory vs enforced) and `selftest_expect:` (which exit code proves falsifiability) explicit rather than inferred. Nine `check-*` scripts sit outside it, and three of them are worth acting on for different reasons.

```
check-*  scripts in .github/scripts   64
  declared in gates.yaml              55
  outside it                           9
```

Of the nine, four are fine — `check-claude-fallback-result.sh`, `check-fleet-attestations.sh`, `check-manifest-types-only.sh`, `check-upstream-unblock.py` are workflow helpers rather than repo gates, and `check-realm-role-parity.py` runs in-cluster from `keycloak-realm-drift.yaml`, which the manifest could not host anyway. `check-openapi-request-schema-conformance.py` runs from `_service-ci.yml`.

## 1. `check-compliance-matrix.py` has never run — it is referenced nowhere

141 lines, written for #2370. Its own docstring states the stakes:

> *the page was wrong in BOTH directions: GDPR rows marked `ok` cited database columns that exist only in a Flyway migration … Hand-corrected rows rot the same way the original ones did — this gate is what stops the next silent drift.*

A repo-wide `git grep` on `origin/main` finds **zero** references outside the file itself. No workflow, no manifest entry, no glob (nothing in this repo iterates `.github/scripts/check-*`; `run-gates.py` reads the manifest, not the directory). So the drift it exists to stop is unstopped, and the compliance page's 46 regulatory-conformance claims are again only as good as the last hand-correction.

This is the repo's own lesson from #2280 — *the gate that never existed beats the unfalsified one* — with the twist that here the gate was actually written.

## 2. `rules.yaml` names a `ci_producer` that CI never invokes

```yaml
ci_producer: ".github/scripts/check-operator-write-naming.py"
```

The surrounding rule is not a small one — it concerns operator-write naming across services where `AUTHZ_ENFORCE` is true, and says in place that *"registering one without that is a money-path outage, not a test failure."*

[...]

Issue #3235, opened 2026-08-02
GlitchTip captures admin-ui errors with no source maps, so every event is unattributable


JiRaska:
## What

GlitchTip captures admin-ui browser errors, but cannot say where any of them happened. Source maps
are not uploaded, so every event stores a minified title and an **empty culprit**.

Measured against the live GlitchTip database:

```
last_seen     | count | title                              | culprit
08-02 04:22   |   3   | TypeError: s.map is not a function | (empty)
08-02 04:07   |   1   | TypeError: s.map is not a function | (empty)
07-11 13:03   |  27   | TypeError: fetch failed            | (empty)
06-26 08:49   |  16   | Error: Minified React error #31…   | (empty)
```

`openbank-admin-ui/next.config.mjs` exports `nextConfig` directly — there is no `withSentryConfig`
wrapper, so `@sentry/nextjs` never uploads the maps it generates.

## Why it is worth fixing

A user reported a blank error screen on the campaign detail page. GlitchTip had the error — three
occurrences, correctly captured and not filtered — and told us `s.map is not a function` with no
file, no line, and no function beyond the minified `s`.

That is the same information the browser console gave, so the tool added nothing over asking the
user for a screenshot. The root cause (a positional mismatch in a `Promise.all` destructuring, #3213)
was found by reasoning from the message, not from the report.

The value of a crash reporter is attribution. Without maps it records that something broke, which is
the half you already know the moment someone complains.

## Scope

- Wrap the config in `withSentryConfig` with source-map upload enabled, and confirm maps are
  **not** served publicly (upload then delete, which is the default `sourcemaps.deleteSourcemapsAfterUpload`).
- Uploading needs an auth token for the GlitchTip project — a credential, so it belongs in the
  existing ExternalSecret flow and CI secret store, not in the repo.
- `NEXT_PUBLIC_GLITCHTIP_RELEASE` already carries a version (`openbank-admin-ui@0.29.0`); the upload
  must use the same release string or the maps will not match the events.

## Verification that would prove it

Not "the build succeeded". Trigger a deliberate client-side throw on a deployed build and confirm the
GlitchTip event names the source file and line. A green upload step proves an upload happened, not
that the symbolication works — the two have been different things before.

## Related

- #3213 — the crash this was discovered through.
- #3214 — closed as incorrect; I had claimed no DSN was configured. It is, and events do arrive.

JiRaska:
[...]

PR #3392, opened 2026-08-02
fix(infra): make the per-service Dockerfiles honest, and keep them that way

### LINKED ISSUE #3016: 51 service Dockerfiles are structurally unbuildable and are never built

JiRaska:
## What

All 52 `openbank-*/Dockerfile` become what they always claimed to be: runtime recipes. Each now
mirrors `.github/workflows/Dockerfile.deploy` exactly, keeping its own `EXPOSE` and its SBOM lines
where it had them. Plus a guard so it stays that way.

**Closes #3016.**

## Why — these files were fiction

Every one carried a Gradle build stage that **could not run**:

| gap | affected |
|---|---|
| no `COPY build-logic` (root `settings.gradle.kts` `includeBuild`s it) | 52 of 52 |
| no `COPY openbank-libs-{domain,runtime,temporal}` (ADR-0122 siblings) | 49 of 52 |
| no `COPY gradle.properties` | 49 of 52 |

Not a theory — `openbank-ledger-service/Dockerfile`, the file usually cited as the correct shape,
fails a real `docker build` in 77 s:

```
Included build '/build/build-logic' does not exist.
```

And nothing builds them anyway. `auto-deploy.yml`, `ghcr-publish.yml` and
`openbank-infra/scripts/build-push-service.sh` each generate their own context from
`Dockerfile.deploy` and read **one** thing from these files: the `EXPOSE` line.
`build-push-service.sh:8` states the reasoning outright — *"WHY NOT `docker build -f
openbank-<svc>/Dockerfile`"* — because in-image Gradle cannot work here.

A file that reads as the build recipe, is treated as authoritative, and cannot run is worse than no
file. This makes them true instead of deleting them, because the `EXPOSE` line is load-bearing.

## EXPOSE preserved byte-for-byte

Every port extracted from `origin/main` and diffed against the result: **52 of 52 unchanged.**

This mattered more than it looks. Losing an `EXPOSE` fails nothing loudly — the deploy pipeline
falls back to port **8080**, and the service comes up unreachable on a port nobody declared.

## Guard, exercised against cases it must flag

`check-dockerfile-no-build-stage.py` runs in the unconditional `Validate manifests` job (not a
path-filtered workflow, so it is genuinely required).

| scenario | result |
|---|---|
| after this change | ✅ `OK — 52 Dockerfiles, runtime-only, all with EXPOSE` |
| re-add `FROM … AS build` | ❌ error, names the file |
| remove an `EXPOSE` | ❌ error, explains the 8080 fallback |
| header prose *about* a build stage | ✅ not flagged — comments are stripped first |

That last row is deliberate. The new header explains the build stage that used to be here, and a
naive scan would flag every file for the sentence describing why it must not exist.

## Pre-existing finding, reported not failed

The guard warns on duplicate `EXPOSE` ports. Two pairs exist on `main` today and **predate this
change**:

- `8128` — `campaign-service`, `customer-edge`
- `8143` — `document-service`, `governance-auditor`

Legal in Kubernetes (separate pods), but more often a copy-paste slip than a decision. Warned rather
than failed, since I cannot tell which from here.

## Risk

Low by construction: nothing in CI or the deploy path builds these files, so the change cannot alter
any shipped image. The one live contract — `EXPOSE` — is verified unchanged and now guarded.


JiRaska:
## This and #3249 are the same job with opposite conclusions

Both rewrite the per-service `Dockerfile`s against the #3016 finding, 31 files in common, 11 hours apart:

```
#3249  chore/strip-dead-dockerfile-build-stages   05:55   100 files  +2123/-1599
#3392  fix/dockerfiles-runtime-only               17:39    54 files  +1329/-1639
```

The content is different on every file I sampled (account, ledger, fx, aml — four for four), so this is genuine divergence, not a duplicate. The disagreement is substantive and worth deciding rather than merging past:

- **#3249** leaves a comment and `EXPOSE`, arguing that reproducing the runtime stage creates *"a second copy nothing builds is a copy free to drift invisibly"*.
- **This PR** mirrors the runtime stage so *"what it says is what actually happens"*, and adds `check-dockerfile-no-build-stage.py`.

## A finding I nearly filed against this, and withdrew

I was going to report that the mirror is already false: these files carry two lines `Dockerfile.deploy` does not have.

```
COPY sbom/bom.json /app/sbom/bom.json
ENV OPENBANK_SBOM_PATH=/app/sbom/bom.json
```

That is wrong, and checking the builders is what showed it — all three append exactly those lines to the generated context at build time:

```
build-push-service.sh:79-80   SBOM_COPY / SBOM_ENV
auto-deploy.yml:489           echo "COPY sbom/bom.json /app/sbom/bom.json" >> "${ctx}/Dockerfile"
ghcr-publish.yml:141          same
```

So these files mirror the **effective** image, not merely `Dockerfile.deploy`'s static text — which is more honest than what I expected to find, not less.

## What does survive

**Nothing enforces the mirror.** The new gate checks two things — no build stage, and `EXPOSE` present. Neither compares the runtime stage to `Dockerfile.deploy`. Today they agree (same base digest `28db6fdf…`, same `WORKDIR`, user, `COPY` set, `ENTRYPOINT`), and nothing keeps them agreeing: bump the base image, change the entrypoint flags or the user in `Dockerfile.deploy`, and 49+ per-service files silently describe an image that is not built.

That is precisely #3249's objection, and it is the one thing that does not resolve itself by picking this PR. It is also cheap to close — the gate already parses these files; comparing the runtime stage against `Dockerfile.deploy` (modulo the appended SBOM lines and `EXPOSE`) turns "honest today" into "honest by construction", which is the difference the two PRs are actually arguing about.

## Suggestion

Pick one and close the other explicitly. If it is this one, extending the gate to enforce the mirror answers #3249 rather than overruling it. Whichever lands, the 31 shared files mean the loser needs closing, not rebasing.

*Reviewed while checking the PR queue; neither branch touched.*

JiRaska:
Blocking finding, with the evidence — `openbank-admin-ui/Dockerfile` is **really built**, and this PR replaces it with a stub.

`openbank-infra/scripts/build-push-admin-ui.sh:330`:

```bash
docker buildx build \
  --platform "${PLATFORM}" \
  --file openbank-admin-ui/Dockerfile \
  --build-arg "BUILD_VERSION=${BUILD_VERSION}" \
  --build-arg "BUILD_GIT_SHA=${GIT_SHA}" \
  --build-arg "BUILD_DATE=${BUILD_DATE}" \
  --tag "${IMAGE}" --push .
```

It is a 372-line multi-stage Next.js build with an `sbom-collector` stage, a baked `test-results.json`, and a BuildKit secret for the GlitchTip source-map upload (#3235). This PR reduces it to 28 lines whose header says the image "is built by `.github/workflows/Dockerfile.deploy` from a `quarkus-app/` produced host-side" — which is a Quarkus fast-jar recipe and cannot build a Next.js app. After this merges, `build-push-admin-ui.sh` builds a container with no application in it.

I know because **I made the identical mistake** in #3249 and only caught it when the branch later conflicted with `main` on that exact file. My exemption list was written from reading the Dockerfiles rather than from what builds them, which is the hand-kept-set failure this repo keeps paying for.

The derivation that fixes it, and the two traps in it:

1. Scan `openbank-infra/scripts/*.sh` and `.github/workflows/*.yml` for `openbank-<svc>/Dockerfile` and exempt anything they name. Anything a build script names by path IS built.
2. Match `Dockerfile(?![.\w])`, not `Dockerfile\b` — `\b` matches before the dot, so `openbank-product-catalog/Dockerfile.native` (built by `product-catalog-native-build.yml`) makes the plain, dead `Dockerfile` read as built.
3. Skip comment-only mentions — `openbank-anacredit-service/Dockerfile` appears only in an `auto-deploy.yml` comment about the EXPOSE grep, and counting it would exempt a genuinely dead file.

On that basis the exempt set is **three**: `openbank-admin-ui` (built by `build-push-admin-ui.sh`), `openbank-developer-portal` (static nginx) and `openbank-document-renderer` (WeasyPrint, ADR-0162 D3). #3016 counts all three among its "51 dead" — it is 48.

One other thing worth checking here: this PR wires the guard into `.github/workflows/ci.yml`, but the gate steps moved into `.github/gates/gates.yaml` (the `gates` job; `validate` is now just an aggregator). A `run:` step added to `validate` will not be picked up the way the other gates are — worth registering it in the manifest instead, with a `selftest:` key so the falsifiability declaration is not lost.

Happy for this PR to be the one that lands — it is a superset on the Dockerfiles and closing #3249 was the right call. This is only about not shipping the admin-ui strip with it.

JiRaska:
## Heads-up: #3414 adds a 53rd Dockerfile your gate rejects

#3414 (`feat(delegation): gitops component + ALL_SERVICES entry`) creates `openbank-delegation-service/Dockerfile` — delegation-service landed in #2971 and had none. It carries a Gradle build stage, so your gate flags it:

```
exit 1
::error:: openbank-delegation-service/Dockerfile: has a `FROM ... AS <stage>` build stage
1 problem(s) across 53 Dockerfiles
```

Independent branches, so neither CI run sees the other. If this merges first, #3414 goes red on the new gate; if #3414 merges first, this PR needs delegation's file too. I have flagged it there as the cheaper place to fix.

Also worth noting the gate would have caught it either way — which is the argument for this PR over #3249 (now closed): a rule that keeps holding as services are added, rather than a one-time sweep.

*Flagged from the PR queue; branch untouched.*

JiRaska:
## Two bookkeeping notes, neither blocking

**The gate is wired in `ci.yml:107`, not declared in `.github/gates/gates.yaml`.** `CLAUDE.md` says gates are declared in the manifest so that `mode:` (advisory vs enforced) and `selftest_expect:` are stated outright rather than inferred from reading a step. Wired this way it also does not appear in `run-gates.py --list` and cannot be run with `--only`.

This is a fresh instance of #3240, which I filed earlier today after finding three `check-*` scripts outside the manifest — one of which (`check-compliance-matrix.py`) has never run at all because nothing references it. This one does run and is enforced, so it is bookkeeping rather than a hole.

**No `--selftest`** — `--selftest` exits 2, unrecognized. So nothing in the repo establishes the gate can fail.

It demonstrably can, though: I ran it against #3414's new `openbank-delegation-service/Dockerfile` and it flagged exactly the right thing.

```
exit 1
::error:: openbank-delegation-service/Dockerfile: has a `FROM ... AS <stage>` build stage
1 problem(s) across 53 Dockerfiles
```

So the failure path is proven, just not repeatably. A `--selftest` with a must-flag and a must-not-flag case would make that a property rather than something I happened to check.

## On merging this over #3249

Closing #3249 was the right call on the measurement — this PR touches 52 Dockerfiles to its 31, and every file it fixed is fixed here too (0 only in #3249). Its 69 extra files were the OPA-bundle restamp from touching `gates.yaml`, not subject matter.

The one thing #3249 raised that this PR does not answer still stands: nothing compares the runtime stage here against `Dockerfile.deploy`. They agree today; the gate checks only for a build stage and for `EXPOSE`. Extending it to enforce the mirror would close that, and is the natural home for it now that this is the surviving approach.

*Reviewed while draining the queue.*

JiRaska:
## This overwrites three Dockerfiles that are not Quarkus services — one of them is built for real

The sweep is `openbank-*/Dockerfile` with no exemption
(`check-dockerfile-no-build-stage.py:61` globs all of them), so all 52 get the same
`Dockerfile.deploy` mirror. Three of them are not Quarkus services, and after this PR all three
claim to be:

```
openbank-admin-ui/Dockerfile          FROM eclipse-temurin:25-jre-alpine
openbank-developer-portal/Dockerfile  FROM eclipse-temurin:25-jre-alpine
openbank-document-renderer/Dockerfile FROM eclipse-temurin:25-jre-alpine
  COPY quarkus-app/lib/ /app/lib/   …
```

### `openbank-admin-ui` — this one breaks a live build

It is a Next.js app and its Dockerfile **is** built, by name, on every admin-ui release:

```bash
# openbank-infra/scripts/build-push-admin-ui.sh
docker buildx build \
  --file openbank-admin-ui/Dockerfile \
  --build-arg "BUILD_VERSION=${BUILD_VERSION}" …
```

On `main` that file is a real multi-stage build (`docs-collector`, `sbom-collector`, the Next.js
stage). Replacing it with `COPY quarkus-app/…` means the next admin-ui build copies a directory
admin-ui never produces. The premise "nothing builds them anyway" is true for the 49 service
Dockerfiles and false for this one.

Worth noting the closed #3249 hit this exact rock and caught itself — its description records that
its first pass stripped `openbank-admin-ui/Dockerfile` and it was put back. That correction did not
survive into this PR.

### `developer-portal` and `document-renderer` — wrong runtime, quieter failure

- `developer-portal` is a static nginx site (`nginxinc/nginx-unprivileged`, `COPY nginx.conf`,
  `COPY site/`).
- `document-renderer` is the WeasyPrint sidecar (ADR-0162 D3) — a two-stage `python:3.14-slim`
  build with `requirements.txt` and `app.py`.

Neither is built in-repo, so nothing goes red — but each file now describes an image that could
never run the process its Deployment expects. That is the same "honest declaration" problem this PR
exists to fix, inverted.

## Suggested shape

Exempt the three in the guard and leave their real recipes alone. The exemption is the honest
artifact: it says "these are not deploy-built Quarkus services", which is exactly true, and it keeps
the guard meaningful for the other 49. If you would rather they carry a header instead, the closed
#3249 has wording for `developer-portal`/`document-renderer` (built out of band, no in-repo
builder) that could be lifted.

The rest of the PR I have no objection to — the measurement of what was broken (52/52 missing
`COPY build-logic`, the 77 s `docker build` failure on ledger) is exactly the right kind of
evidence.

JiRaska:
Following up on my review above — this had already merged when I posted it, so the three flattened Dockerfiles are on `main`. Fix in #3439: restores them from `2a6537e3d`, adds a two-directional `SELF_BUILT` exemption, and adds the missing `re.M` (without it `COPY --from=` and `RUN gradlew` could never match, and the build-stage rule only matched a first-line violation).

PR #3404, opened 2026-08-02
fix(domestic): resolve the debtor's party for an AML case instead of sending the account id

### LINKED ISSUE #3274: aml: every AML case carries an ACCOUNT id in party_id — the ADR-0032 account→party resolution follow-up never landed
### Affected service / module

openbank-domestic-payment (`infrastructure/client/AmlCaseAdapter.kt`) + openbank-aml-service

### Version / commit

main @ 63b6cd8dd (observed live in sandbox, 2026-08-02)

### What happened?

Every AML case opened from the payment path carries an **account id in `party_id`**. The adapter
fills both fields from the same value:

```kotlin
partyId   = command.debtorAccountId,
accountId = command.debtorAccountId,
```

Live row from case `856e4839-c3ce-4e3b-a268-5fbfe1842975`:

```
party_id   | bc9bc850-d3ae-4b45-abc2-de8a3cb8201c
account_id | bc9bc850-d3ae-4b45-abc2-de8a3cb8201c
```

`bc9bc850-…` is an **account** (`accounts.id`, account_number `CZ41…`). That account's real
`party_id` is `07d13ecc-443b-456e-a504-dd8999078a48`. So the case points at a party that does not
exist, and no join from `aml_cases.party_id` to the party register can resolve.

`party_id` is `not null` in the schema, so the column looks populated and healthy. Nothing detects
that the UUID is from the wrong table.

### This is a known, documented shortcut whose follow-up never landed

ADR-0032, `docs/adr/0032-synchronous-sanctions-aml-screening-gate-in-payment-execution.md`:

> The **AML case `partyId`** is a required UUID, but a payment only carries `debtorAccountId` (the
> party is not resolved at the payment boundary). The adapter maps `debtorAccountId` into both
> `partyId` and … Proper account→party resolution (via `account-service`) is a documented
> [follow-up]

and in its consequences:

> `partyId` on the AML case is account-derived until party resolution lands (documented).

Party resolution has since landed elsewhere — ADR-0179 (party identity merge) is Done — but this
adapter was never revisited. The decision record is honest; the follow-up is simply outstanding, and
there is no issue tracking it, which is why it has stayed outstanding.

### Why it matters beyond tidiness

AML case work is **party-scoped**, not account-scoped. With an account id in `party_id`:

- cases cannot be aggregated per customer — a client with three accounts looks like three unrelated
  subjects, which is exactly the pattern structuring detection is supposed to catch;
- SAR filing and periodic review key off the party;
- a customer's case history cannot be assembled, so "has this subject been alerted before" is
  unanswerable from the case store.

The adapter also has `AccountLookupPort` available (`infrastructure/client/AccountServiceClient.kt`),
so the resolution is reachable — but see below before wiring it naively.

### What did you expect?

The AML case carries the debtor's real `partyId`, resolved from `debtorAccountId` via
account-service, with `accountId` kept as the separate field it already is.

Two things the fix has to settle, because the naive version makes things worse:

1. **A lookup on this path must not be able to hold the payment.** `openCaseQuietly` is deliberately
   best-effort — it swallows failures so opening a case can never change a screening verdict. Adding
   a synchronous account-service call inside it reintroduces exactly the availability coupling that
   stranded a payment via sanctions-service (#3264–#3267). Resolve it asynchronously, or let
   aml-service resolve it on ingest, rather than putting another network call on the payment path.
2. **Existing rows.** Every case written to date has an account id in `party_id`. A back-fill is
   possible (`accounts.id → accounts.party_id`) but it rewrites compliance records, so it needs an
   explicit decision and an audit note rather than a quiet `UPDATE`.

### Reproduction steps

1. Trigger any AML case from a domestic payment (a `BLOCK`/`REVIEW` screening verdict, or an
   unavailable sanctions-service).
2. `select party_id, account_id from aml_cases where transaction_id = '<paymentId>'` — the two
   values are identical.
3. `select party_id from accounts where id = '<that value>'` — returns a *different* UUID, the real
   party.

### Logs / stack traces

```shell
# aml_cases row (PII fields omitted)
case_id        | 856e4839-c3ce-4e3b-a268-5fbfe1842975
party_id       | bc9bc850-d3ae-4b45-abc2-de8a3cb8201c   <- an accounts.id
account_id     | bc9bc850-d3ae-4b45-abc2-de8a3cb8201c
screening_type | TRANSACTION_MONITORING
alert_code     | SCREENING_UNAVAILABLE
status         | OPEN
```

### Severity

S3 — degraded service with workaround

### Notes

- Check the other rails before closing: any service constructing `OpenAmlCaseCommand` from an
  account id has the same shape. This was found in domestic-payment only because that is the rail
  that was being debugged.
- Nothing here is caught by a test: `party_id` is a valid non-null UUID, so schema validation,
  serialization and every assertion on "a case was opened" all pass. The assertion that fails is a
  join against the party register.

JiRaska:
Closes #3274 · Refs ADR-0032

## The defect

A payment carries only `debtorAccountId`, and the adapter filled the AML case's required `partyId` with it:

```kotlin
partyId   = command.debtorAccountId,
accountId = command.debtorAccountId,
```

So every case opened from the payment path pointed at a party that does not exist. `party_id` is `not null`, so the column read as populated and healthy, and nothing detected that the UUID came from the wrong table. ADR-0032 documented the account→party resolution as a follow-up; it never landed.

## The fix

`AccountLookupPort.findPartyByAccountId` over the account-service client this service **already has** (`GET /api/v1/accounts/{accountId}`) — no new dependency, config, or network edge.

On a failed lookup the account id is still sent. Losing an AML case over an account-service outage would be worse than an imprecise one — but it is now logged at WARN with the payment and account ids, so those rows are **identifiable** rather than indistinguishable from correctly-resolved ones.

## The test pinned the defect

This is why the issue's "nothing here is caught by a test" was true — it was worse than absence:

```kotlin
assertThat(body.partyId).isEqualTo(cmd.debtorAccountId)   // asserted the wrong behaviour
```

Now it asserts the case carries the **resolved** party *and* that it differs from the account id, so the old shape cannot come back silently. A second case pins the unresolved fallback.

## Two other rails have the identical shape — deliberately not in this PR

The issue says to check before closing. Measured:

| rail | same two lines? | has an account-service client? |
|---|---|---|
| `openbank-domestic-payment` | yes | **yes** — fixed here |
| `openbank-sepa-payment` | yes | no |
| `openbank-sepa-instant` | yes | no |

Neither of the other two can reach account-service at all today, so each needs a new rest-client, its config and the network edge that goes with it. That is a separate change per rail — not three money-path services in one PR. Worth deciding whether to fix them or to have aml-service reject a `partyId` that is not in the party register, which would close the class rather than the instances.

## Verification

`ktlintCheck`, `detekt`, `test`, `quarkusBuild` green for `openbank-domestic-payment`. `quarkusBuild` matters here specifically — the adapter gained an `@Inject` field, and CDI wiring is not validated by ktlint or unit tests.

Money-path service, so two approvals per `rules.yaml` / ADR-0030.

### DIFF
--- a/openbank-domestic-payment/src/main/kotlin/com/openbank/domestic/application/port/out/AccountLookupPort.kt
+++ b/openbank-domestic-payment/src/main/kotlin/com/openbank/domestic/application/port/out/AccountLookupPort.kt
@@ -10,6 +10,17 @@ interface AccountLookupPort {
     /** Returns the partyId that owns [iban], or null if not an internal account. */
     suspend fun findPartyByIban(iban: String): UUID?
 
+    /**
+     * Returns the partyId that OWNS [accountId], or null when it cannot be resolved.
+     *
+     * ADR-0032's documented follow-up. A payment carries only `debtorAccountId`, and the AML
+     * adapter filled the case's required `partyId` with it — so every case opened from the payment
+     * path pointed at a party that does not exist, and no join from `aml_cases.party_id` to the
+     * party register resolved. `party_id` is `not null`, so the column looked populated and
+     * healthy the whole time (#3274).
+     */
+    suspend fun findPartyByAccountId(accountId: UUID): UUID?
+
     /** Returns the accountId for [iban] (the internal creditor account to credit), or null if it
      *  is not an internal account / cannot be resolved. */
     suspend fun findAccountIdByIban(iban: String): UUID?
--- a/openbank-domestic-payment/src/main/kotlin/com/openbank/domestic/infrastructure/client/AccountServiceClient.kt
+++ b/openbank-domestic-payment/src/main/kotlin/com/openbank/domestic/infrastructure/client/AccountServiceClient.kt
@@ -28,6 +28,13 @@ interface AccountServiceRestClient {
         @HeaderParam("Authorization") authorization: String,
         @PathParam("iban") iban: String,
     ): io.smallrye.mutiny.Uni<AccountDto>
+
+    @GET
+    @Path("/{accountId}")
+    fun getById(
+        @HeaderParam("Authorization") authorization: String,
+        @PathParam("accountId") accountId: String,
+    ): io.smallrye.mutiny.Uni<AccountDto>
 }
 
 data class AccountDto(val id: String, val partyId: String)
@@ -71,6 +78,21 @@ class AccountServiceClient(
         }
     } catch (ex: Exception) {
         log.warnf(ex, "Account lookup for IBAN %s failed — treating as external", iban)
+        null
+    }
+
+    @Suppress("TooGenericExceptionCaught")
+    override suspend fun findPartyByAccountId(accountId: UUID): UUID? = try {
+        val token = oidcClient.get().tokens.awaitSuspending().accessToken
+        val dto = httpClient.getById("Bearer $token", accountId.toString()).awaitSuspending()
+        UUID.fromString(dto.partyId)
+    } catch (ex: jakarta.ws.rs.WebApplicationException) {
+        if (ex.response.status != HTTP_NOT_FOUND) {
+            log.warnf(ex, "Party lookup for account %s failed with HTTP %d", accountId, ex.response.status)
+        }
+        null
+    } catch (ex: Exception) {
+        log.warnf(ex, "Party lookup for account %s failed", accountId)
         null
     }
 
--- a/openbank-domestic-payment/src/main/kotlin/com/openbank/domestic/infrastructure/client/AmlCaseAdapter.kt
+++ b/openbank-domestic-payment/src/main/kotlin/com/openbank/domestic/infrastructure/client/AmlCaseAdapter.kt
@@ -4,6 +4,7 @@
 
 package com.openbank.domestic.infrastructure.client
 
+import com.openbank.domestic.application.port.out.AccountLookupPort
 import com.openbank.domestic.application.port.out.AmlCasePort
 import com.openbank.domestic.application.port.out.OpenAmlCaseCommand
 import io.smallrye.mutiny.coroutines.awaitSuspending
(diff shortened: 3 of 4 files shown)


JiRaska:
## Merging — the client call is real end to end, checked rather than assumed

A new cross-service call is where #2269 lives (finrep calling a ledger path that had never existed, green at every layer because a Pact mock answers whatever it is asked). So I traced this one to the provider rather than reading the adapter:

```
path        /api/v1/accounts/{accountId}   AccountResource.kt:70 + :157   openapi.yaml:166
handler     getAccount → account.toResponse()
field       AccountResponse.partyId        openapi schema ✓   AccountDtos.kt:34 val partyId: UUID ✓
```

Both halves hold — the route exists and the response carries the field the adapter reads. Had `partyId` been absent from `AccountResponse`, the parse would have yielded null, the fallback would have fired on every payment, and the only trace would be the WARN this PR adds.

## On the defect itself

`partyId = command.debtorAccountId` put an account UUID in a party column. The part worth keeping in mind: `party_id` is `not null`, so the column read as populated and healthy. There is no null to find, no constraint to violate, and nothing downstream that can tell one UUID from another — the only way to see it is to know which table the value came from.

## On the fallback

Still sending the account id when the lookup fails is the right call, and the reasoning in the PR body is sound — losing an AML case to an account-service outage is worse than an imprecise one. What makes it acceptable rather than a smaller version of the same bug is the WARN carrying the payment and account ids: those rows become **identifiable**, where before they were indistinguishable from correctly-resolved ones.

Worth noting the fix is forward-only — rows already written with an account id in `party_id` stay wrong, and nothing in this PR finds them. Whether that needs a backfill is a separate question from whether this merges.

All five required contexts green, nothing else red.

*Merged at the owner's explicit instruction; money-path PRs are otherwise left for a human.*

Issue #2983, opened 2026-07-31
A shared-library change rebuilds nothing — consumers keep running the old code until something else touches them


JiRaska:
## What happens

`Auto deploy` derives its build set from changed **service** directories. A change under
`openbank-libs-*` matches none of them, so the run ends with:

```
Detect changed services: success
Build + push: skipped
```

No consumer is rebuilt. The library change sits on `main`, and every service that depends on it keeps
running the previous code until some unrelated commit happens to touch that service.

## How it showed up

#2949 fixed the Temporal payload converter in `openbank-libs-temporal`. Merging it rebuilt nothing;
campaign-service only picked it up because I re-dispatched `auto-deploy` for that service by hand.

The other Temporal consumers still have not: `fx-service` is on `sandbox-99114189` and
`statement-service` on `sandbox-e5edf0ce`, both built before the fix. Their pods are green, which is
worth being precise about — **their health says nothing about the change**, because they are not
running it.

## Why it is worth fixing rather than remembering

Two failure modes, and the second is the dangerous one:

1. A fix does not reach its consumers. Visible eventually, annoying, recoverable.
2. A *regression* in a shared library reaches consumers one at a time, weeks apart, each carried in
   by an unrelated PR. The blast radius is the same either way, but the correlation between "the
   library changed" and "this service broke" is gone by the time it surfaces — and the PR that
   triggers the rebuild gets the blame.

The current arrangement also makes any claim of the form "the fleet is running library version X"
untrue by default.

## Options

1. **Map libs → dependents and rebuild them.** The dependency is already declared in each
   `build.gradle.kts`; `./gradlew :svc:dependencies` or a settings-level scan can derive the set.
   Most faithful, and the derived-set approach the pact drift gate already uses.
2. **Rebuild every service on a libs change.** Simple and always correct, but expensive on a fleet
   this size, and the FinOps cost lands on every libs commit.
3. **Leave it, and require a follow-up dispatch in the PR checklist.** Cheapest, and it is what
   happens today by accident rather than by rule — but a checklist step is exactly the thing that
   gets skipped at 01:00.

I lean towards 1, deriving the dependent set rather than maintaining a list, since a hand-kept
mapping would rot the same way the ESO allow-list did (#2851).

Refs #2749, #2949

PR #3407, opened 2026-08-02
fix(gitops): "no libs dependents" is a normal answer, not an exit 1 that kills every deploy

JiRaska:
## What

**Every push-triggered `Auto deploy` run has been dying since 14:56 today**, in `Detect changed services`, with **no output at all**. Nothing has reached the cluster since — including the five PRs merged this evening.

```
$ gh run list --workflow=auto-deploy.yml   # last 12
push:failure  failed_jobs=[Detect changed services]     x9
workflow_dispatch:failure  failed_jobs=[can-i-deploy]   <- unrelated
workflow_dispatch:success
```

`workflow_dispatch` and `schedule` survive only because both branches `exit 0` before they reach the code below.

## Why

`auto-deploy.yml` consumes this script as `LIBS_DEPENDENTS="$( … | bash libs-change-dependents.sh … )"` under `set -euo pipefail`. The script had **two** silent routes to exit 1, and either one aborts the whole step with no diagnostic:

1. **No libs module changed** → `$DEPENDENTS` is empty → `grep -v '^$'` matches nothing and exits 1 → `pipefail` makes that the pipeline's status. This is the commonest push in the repo.
2. `for known …; do [ "$svc" = "$known" ] && echo … && break; done` ends on a **failed** test whenever the last service read is not in `ALL_SERVICES` — so the `for`, the `while` body, the loop and the pipeline all report 1.

Route 2 is also why a trailing `exit 0` does not help: `set -e` aborts **at** the failing pipeline and never reaches it. (I tried exactly that first, and it still exited 1.)

Reproduced against `origin/main`, and after:

```
printf 'openbank-fraud-service/src/main/kotlin/X.kt\n' \
  | bash .github/scripts/libs-change-dependents.sh openbank-fraud-service
before: EXIT=1  (no output)
after:  EXIT=0  (no output)
```

## How

The grep's no-match is guarded, and the comparison becomes an `if` — with no `else` it returns 0 when false, so the loop ends cleanly whether or not the last service matched. The script's contract is "print the dependents, possibly none"; only a real error may be non-zero.

## Why the gate said OK

`--selftest` exercised the matchers **in-process** and never ran the script as a subprocess — so the one thing the caller actually consumes, the **exit status**, was untested. It printed `selftest OK` while the script exited 1 on the commonest input in production.

Three end-to-end cases now cover it (no libs change; dependents outside `ALL_SERVICES`; the positive path). **All three fail against the pre-fix code** — verified by reverting only the two production lines and re-running the new selftest:

```
selftest FAIL: a push touching no libs module exited 1 — 'no dependents' is a normal answer
selftest FAIL: dependents outside ALL_SERVICES made the script exit non-zero
selftest FAIL: a real libs change exited non-zero
```

`run-gates.py --only libs-change-dependents` → PASS. Validated under **bash 3.2** (macOS `/bin/bash`) as well as bash 5. The two `SC2086` findings on the changed line are deliberate word-splitting of a list and predate this change.

## Note on the earlier failures

Auto-deploy was also red this morning (~04:00–05:00) for a **different** reason: those runs have **zero jobs** and `name` equal to the file path rather than `Auto deploy` — the unparseable-workflow signature from #3135. That era is over; today's runs parse. Two eras, two causes; this PR fixes the current one.

## Linked issues

Refs #2983 — the fan-out this script implements.

N/A otherwise — the outage was found while verifying that #3336/#3350/#3371/#3376 had deployed. They had not.

### DIFF
--- a/.github/scripts/libs-change-dependents.sh
+++ b/.github/scripts/libs-change-dependents.sh
@@ -91,7 +91,34 @@ selftest() {
     echo "selftest FAIL: a governance catalog matched the global paths — that is data, not a compile input" >&2
     fail=1
   fi
-  [ "$fail" -eq 0 ] && echo "selftest OK: declaration matcher and both path matchers verified in both directions."
+  # END-TO-END exit status. Everything above tests the matchers in-process, which is why this
+  # script could exit 1 on the commonest input in production while the selftest stayed green: the
+  # caller consumes the script as a command substitution under `set -e`, so its EXIT STATUS is part
+  # of the contract and has to be exercised by actually running it.
+  local out rc
+  # (a) No libs module changed — the overwhelmingly common push. Must be exit 0 and no output.
+  out="$(printf 'openbank-fraud-service/src/main/kotlin/X.kt\n' \
+    | bash "$0" "openbank-fraud-service" 2>/dev/null)" && rc=0 || rc=$?
+  if [ "$rc" -ne 0 ]; then
+    echo "selftest FAIL: a push touching no libs module exited $rc — 'no dependents' is a normal answer" >&2
+    fail=1
+  elif [ -n "$out" ]; then
+    echo "selftest FAIL: a push touching no libs module printed '$out'" >&2
+    fail=1
+  fi
+  # (b) A libs module changed, but NO consumer of it is in ALL_SERVICES — so every comparison in
+  #     the intersection loop fails, including the last one. (An empty argument cannot be used
+  #     here: `${1:?}` rejects it as a usage error, which is a legitimate non-zero.)
+  printf 'openbank-libs-temporal/src/main/kotlin/X.kt\n' | bash "$0" "openbank-not-a-real-service" >/dev/null 2>&1 \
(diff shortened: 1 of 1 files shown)


JiRaska:
Closing as a duplicate of #3406, which landed at 20:11 while this was in review and fixes the same defect.

I compared content rather than merge state, and #3406 is at least as complete:
- it removes the `while` loop entirely (`grep -Fxf <(…) || true`), so **both** silent exit-1 routes disappear — the empty-`$DEPENDENTS` grep and the loop ending on a failed `[ … ]`;
- it already added the end-to-end exit-status self-test cases, which is the part I thought was missing when I re-checked (my first grep looked for my own idiom, `bash "$0"`, and missed theirs, `bash "$me"` — a probe that answers about the wrong thing).

The only thing here that #3406 does not have is a third self-test case (dependents present but none in `ALL_SERVICES`) and an assertion that the no-libs-change path prints *nothing*. Against `grep -Fxf` the first is close to vacuous — the intersection is the implementation — so it is not worth a second PR churning a file two people edited within the hour. Happy to add it if you disagree.

Verified `origin/main` behaves correctly on all three shapes: exit 0 for no libs change, for a real libs change, and for dependents outside `ALL_SERVICES`.

Issue #3382, opened 2026-08-02
auto-deploy: Detect changed services exits 1 with no output, skipping every downstream job on main


JiRaska:
## What

`auto-deploy.yml` / **Detect changed services** step 3 (`Run set -euo pipefail`) exits 1 with
**zero stdout**, which skips every downstream job. Confirmed on two `main` pushes today:

| commit | detect | downstream |
|---|---|---|
| `5ce707b1` (merge of #2971) | **failure** | Build + push, can-i-deploy, Gitops PR, Deploy signal — all `skipped` |
| `0dad2515` | **failure** | same |
| `122c5e4f` | success | ran |

Step conclusions read from `actions/runs/<id>/jobs`, not from the log — the step's `run:` script
is echoed into the log, so grepping it matches strings that never executed.

Effect: a merge to `main` can build and deploy **nothing**, and the only signal is a red run on
`main` that is addressed to nobody. #2971 merged at 17:11 and produced no image.

## Root cause — NOT established

The step ends with no output at all, so it dies early, before its first `echo`. The first
substantial statement is

```bash
LIBS_DEPENDENTS="$(printf '%s\n' "$CODE_FILES" \
  | bash .github/scripts/libs-change-dependents.sh "$ALL_SERVICES")"
```

and under the step's `set -euo pipefail` a non-zero exit from that script kills the step
silently. The script's last pipeline is

```bash
printf '%s\n' $DEPENDENTS | command grep -v '^$' | sort -u | while read -r svc; do
```

where an empty `DEPENDENTS` makes `grep -v` exit 1. That is a *plausible* mechanism, not a
confirmed one: **I could not reproduce it locally** — the script exits 1 on this macOS host for
both the no-libs-change and the with-libs-change input, including inputs CI accepts, so the
local harness cannot discriminate (bash 3.2 here vs bash 5 on the runner). Do not treat the
mechanism above as diagnosed; reproduce it on a Linux runner first.

Whoever picks this up: the falsifying test is an input that MUST produce output
(a real `openbank-libs-runtime/src/main/**` change) — if that also exits 1 on the runner, the
empty-`DEPENDENTS` theory is wrong.

## Separately: openbank-delegation-service can never deploy

`git show origin/main:.github/workflows/auto-deploy.yml | grep -c openbank-delegation-service`
returns **0**. `ALL_SERVICES` is a static list, so the service is not a deploy candidate at all —
the same gap the comment block in that very step documents for billing-service (#597/#601),
[...]

Issue #2992, opened 2026-08-01
GitOps deployment component for openbank-delegation-service


JiRaska:
## Problem

delegation-service (#2971) builds, tests and has governance registration, but **has no deployment component** — it cannot run anywhere outside CI. The money-path SLOs (Pyrra) and rules.yaml classification are already in place, so the service is one gitops component away from a sandbox deploy.

## Proposed solution

Create `openbank-infra/gitops/components/delegation/` mirroring the consent component:

1. `namespace.yaml`, `postgres.yaml` (CNPG, `openbank_delegations` DB), `redis.yaml` (idempotency store).
2. `delegation-service.yaml` — Deployment (fast-jar image via `build-push-service.sh`, `replicas: 1` steady-state; outbox claim is canary-safe via FOR UPDATE SKIP LOCKED), Service, ports 8126/8090.
3. `gen-delegation-opa-bundle.sh` + `delegation-opa-bundle.yaml` + `delegation_rest_ext.rego` (+ test rego) — actions: `delegation.offer|read|list|accept|decline|renounce|revoke|suspend|reinstate|check`. Generator committed mode `100755` and embedding rules.yaml (the #1184 lessons).
4. `network-policies.yaml` (regenerate via `gen-network-policies.py`), `oidc-externalsecret.yaml`, `kafka-delegation-mtls.yaml` (topic `openbank.delegation.events` exists since #2971).
5. ArgoCD application registration + the `Validate manifests` derived artefacts (runbook regeneration, catalog).
6. Pre-first-deploy gates per ADR-0030: threat model review (`docs/threat-models/openbank-delegation-service.md`), `AUTHZ_ENFORCE=false` advisory window with clean `openbank.authz.decisions{outcome="deny",enforced="false"}` report before any enforcement flip.

## Alternatives considered

Deploy without the OPA bundle (authz unconfigured = fail-closed 503 on every @Authorize call in enforce mode) — rejected; the bundle is part of the component, not a follow-up.

## Does this require an ADR?

No — fits existing patterns (ADR-0010, ADR-0034 D5 rollout phases).

## Compliance touchpoints

- DORA — first deploy of a money-path-adjacent service; enters BCP/ICT register operationally
- PSD2 / SCA — production SCA ceremony verification against real sca-service

Money-path: resulting PR needs 2 approvals + threat model reference.

Issue #2991, opened 2026-08-01
Pact contract tests for delegation-service (sca-service, pid-service, provider)


JiRaska:
## Problem

delegation-service (#2971) ships with two outbound REST dependencies — sca-service `GET /api/v1/sca/challenges/{id}` (grant/accept ceremony verification) and pid-service `GET /api/v1/parties/{id}` (D5 eligibility gate) — and eleven provider endpoints, **none of them covered by consumer-driven contract tests** (ADR-0063/0092). A breaking change in sca-service's challenge response or pid-service's party response would only surface at runtime, on the endpoint that mints payment rights.

## Proposed solution

1. **Consumer pact**: delegation-service → sca-service `getChallenge` (mirror `ConsentScaChallengePactConsumerTest`).
2. **Consumer pact**: delegation-service → pid-service `getParty` — pin the minimal shape the eligibility gate reads (`id`, `status`, `kycAttributes.kycLevel`).
3. **Provider verification**: `DelegationPactProviderVerificationTest` replaying consumer pacts that name openbank-delegation-service as provider (the `/api/v1/delegations` surface, once a consumer exists — customer-edge slice).
4. Single `@Provider` test class per provider with per-interaction target selection (the fleet's Pact pitfall: separate HTTP/message classes collide).

## Alternatives considered

WireMock-only integration tests — rejected: they pin the consumer's *assumption* of the provider, not the provider's actual contract; Pact is the repo standard (ADR-0063).

## Does this require an ADR?

No — fits existing patterns (ADR-0063, ADR-0092).

## Compliance touchpoints

None beyond the existing PSD2 / SCA posture (contract tests change no runtime behavior).

PR #3412, opened 2026-08-02
feat(customer-edge): customer routes for delegated access (ADR-0232 D6)

JiRaska:
## Why

`openbank-delegation-service` is on `main` (#2971 + the enforcement projections #3058/#3105/#3143)
and the app still cannot use any of it. The app authenticates against the `openbank-customers`
Keycloak realm, which upstream services do not accept — everything reaches them through
customer-edge — and `git grep -l delegation openbank-customer-edge/src/main` returned **nothing**.

This is one of three independent blockers between "merged" and "usable". The other two are
#2992 (no gitops workload manifest) and #3382 (not in `auto-deploy.yml`'s `ALL_SERVICES`); both
are separate PRs. **This PR alone does not make the feature work** — it makes the edge able to
proxy a service that does not yet run anywhere.

## What

Eight routes under `/customer/v1/delegations` in a new `CustomerDelegationResource`, kept out of
`CustomerEdgeResource` (already `@Suppress("LargeClass")`), matching `CustomerDocumentResource`.

| | |
|---|---|
| `POST /delegations` | offer a grant (SCA-bound) |
| `GET /delegations/shared-by-me` | grants the caller issued |
| `GET /delegations/shared-with-me` | grants received, incl. pending offers |
| `GET /delegations/{id}` | detail |
| `POST /delegations/{id}/accept` | accept, grantee's own SCA |
| `POST /delegations/{id}/decline` | decline |
| `POST /delegations/{id}/renounce` | give an active grant back |
| `DELETE /delegations/{id}` | revoke |

## The shape of the routes is the security property

They deliberately do **not** mirror upstream. delegation-service exposes `/grantor/{partyId}` and
takes `granteePartyId` as a query parameter — both things a customer client must not choose. Here
the caller says *which side of the relationship* they are asking about and the edge supplies *who
they are*, from the JWT, on every route.

Two decisions worth arguing with if you disagree:

- **`offer()` rejects rather than rewrites.** A body naming a different `grantorPartyId` gets 403;
  it is not silently replaced with the caller's. A client sending the wrong id has a bug, and
  quietly issuing a grant the user never asked for is the worse of the two failures. An absent
  field is filled in, since the app has no reason to send it at all.
- **`revoke()` never sends `revokedBy`.** Upstream derives the actor from `X-Customer-Party-Id` for
  customer-scoped calls; the query parameter it still accepts is the bank-initiated, role-gated
  path. Passing it from the edge would re-open exactly what #3164 closed, where the actor was
  caller-supplied and anyone could revoke anyone's grant signed as someone else.

delegation-service re-checks the caller itself (`requireCallerIs`). This is the outer half of that
pair, not a replacement for it — but it is the half the app actually goes through.

## Verification

- 12 unit tests, all naming a party the caller is **not** and proving it never reaches upstream.
  A test asserting only "the URL contains a UUID" would pass against the bug worth catching.
- `detekt` + `ktlintCheck` + `test` green, exit code read directly from Gradle.
- `openapi.yaml` 1.25.0 → 1.26.0 (new endpoints, backward compatible; ADR-0048 classifies from the
  `oasdiff`, this is the expected minor).
- `application.yaml` gains `delegation-service-url`, checked with a duplicate-key-rejecting YAML
  loader — SmallRye silently keeps the last of a repeated mapping key.

**The route-conformance gate was verified falsifiable rather than trusted.** It reports 77 findings
on this branch and 77 on `main`, which on its own proves nothing — the checker might simply not see
the new resource. Removing the delegation paths from `openapi.yaml` takes it to **85**, exactly the
8 route+verb pairs added, and restoring them returns it to 77.

## Not covered here

No Pact contract for the edge→delegation pair (#2991 — delegation-service has zero pacts today, so
there is nothing to replay against yet). No notification on lifecycle transitions, no `onBehalfOf`
audit threading, no object-level disclosure — all unbuilt upstream, and the UI must not promise
them. Cumulative daily/monthly ceilings are carried by the API but enforced nowhere, so a
"5 000 Kč/day" chip in the app would be a claim the platform cannot keep.

Refs #2992, #2991

### DIFF
--- a/openbank-customer-edge/src/main/kotlin/com/openbank/customeredge/infrastructure/rest/CustomerDelegationResource.kt
+++ b/openbank-customer-edge/src/main/kotlin/com/openbank/customeredge/infrastructure/rest/CustomerDelegationResource.kt
@@ -1,0 +1,190 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) OpenBank contributors. Licensed under the Apache License, Version 2.0.
+// See LICENSE in the repository root or https://www.apache.org/licenses/LICENSE-2.0 for details.
+
+package com.openbank.customeredge.infrastructure.rest
+
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.fasterxml.jackson.databind.node.ObjectNode
+import io.smallrye.common.annotation.Blocking
+import jakarta.annotation.security.RolesAllowed
+import jakarta.inject.Inject
+import jakarta.ws.rs.Consumes
+import jakarta.ws.rs.DELETE
+import jakarta.ws.rs.ForbiddenException
+import jakarta.ws.rs.GET
+import jakarta.ws.rs.POST
+import jakarta.ws.rs.Path
+import jakarta.ws.rs.PathParam
+import jakarta.ws.rs.Produces
+import jakarta.ws.rs.QueryParam
+import jakarta.ws.rs.core.MediaType
+import jakarta.ws.rs.core.Response
+import org.eclipse.microprofile.config.inject.ConfigProperty
+import org.eclipse.microprofile.jwt.JsonWebToken
+import java.util.UUID
+
+/**
+ * Customer-facing delegated-access routes (ADR-0232 D6), proxying `openbank-delegation-service`
+ * with the edge M2M token. Without this class the service is merged but unreachable: the app
+ * authenticates against the `openbank-customers` realm, which upstream services do not accept.
+ *
+ * **The party identity is taken from the token on EVERY route and never from the client.** That is
+ * the whole security property of this file, and it is why the routes do not simply mirror the
+ * upstream paths: upstream exposes `/grantor/{partyId}` and takes `granteePartyId` as a query
+ * parameter, both of which a customer client must not be able to choose. Here the caller says
+ * *which side of the relationship they are asking about* ("shared by me" / "shared with me") and
+ * the edge supplies *who they are*.
+ *
+ * delegation-service independently re-checks the same thing via `X-Customer-Party-Id`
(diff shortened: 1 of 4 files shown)


JiRaska:
## Merging — checked the two things a customer-facing delegation surface can get wrong

**Authorization is class-level, not per-route.** `@RolesAllowed("ROLE_CUSTOMER")` sits on line 51, above `class` on 52, so all six routes are covered — worth confirming rather than counting annotations, since one method-level annotation and six routes looks identical in a diff.

**Caller identity comes from the token, not the request.** Every upstream call passes `partyId()` derived from the injected `JsonWebToken`, which becomes `X-Customer-Party-Id` — the header delegation-service's C2 fix (#3164) scopes every customer-facing handler on. So the edge and the service agree on who is asking, and neither trusts the body for it.

Two choices in here are the right ones and are argued in place rather than assumed:

- `getById` adds no check of its own and leans on delegation-service answering **404, not 403**, for a grant the caller is not party to — so a guessed id is not an existence oracle. That matches what C2 implemented upstream.
- `grantorPartyId` is forced to the token's party, and a body naming a *different* grantor is **rejected with 403 rather than silently rewritten**. Quietly issuing a grant the user did not ask for is the worse failure, and the KDoc says so.

## API contract (ADR-0048)

Read off live `main`, not the branch's base:

```
openbank-customer-edge  info.version   main: 1.25.0   branch: 1.26.0
open PRs touching this spec: this one only
```

Minor bump for additive routes, nothing racing for the number — so the #3055 failure mode (a version consumed by an already-merged PR) does not apply.

`agent-review` is red for #2161, unrelated. All five required contexts green.

*Reviewed while draining the queue.*

PR #3414, opened 2026-08-02
feat(delegation): gitops component + ALL_SERVICES entry so delegation-service can deploy

### LINKED ISSUE #2992: GitOps deployment component for openbank-delegation-service

JiRaska:
## Why

`openbank-delegation-service` merged to `main` (#2971, with the enforcement projections
#3058/#3105/#3143) and **could never run**. Two independent blockers, either sufficient alone:

1. `openbank-infra/gitops/components/delegation/` held only the OPA bundle, the rego and a Kafka
   topic — no namespace, no database, no workload, no Service, and no ArgoCD Application pointing
   at the directory. Namespace `delegation` did not exist in the cluster.
2. `auto-deploy.yml` derives its build set from a **static** `ALL_SERVICES` list that did not
   contain the service, so it was never a deploy candidate regardless of what changed — the same
   gap that hit billing (#597/#601), anacredit and six control agents, all documented in that
   workflow's own comment block.

## What

| file | |
|---|---|
| `namespace.yaml` | PSS restricted, app plane |
| `postgres.yaml` | single-owner CNPG `openbank_delegations` + WAL archiving + daily backup |
| `redis.yaml` | idempotency store |
| `external-secret-oidc.yaml` | OIDC client secret from Vault |
| `kafka-delegation-mtls.yaml` | KafkaUser, **Write-only** on `openbank.delegation.events` |
| `kafka-mtls-externalsecrets.yaml` | keystore/truststore projection |
| `delegation-service.yaml` | Deployment + Service + OPA sidecar |
| `apps/delegation.yaml` | ArgoCD Application |
| `openbank-delegation-service/Dockerfile` | `EXPOSE 8126` (read for the container port) |
| `auto-deploy.yml` | `ALL_SERVICES` entry |

The KafkaUser grants **no consumer-group ACL**: this service reads no topic, so granting one would
widen the principal for no caller. Its consumers (account-service, card-issuance) each hold their
own user with their own Read ACL.

## Things that would have failed silently

- **Probes target management port 8090, not 8085.** This service sets `quarkus.management.port`
  itself; the fleet-common 8085 would have produced probes against nothing. The
  `probe-port-has-a-listener` gate confirms it.
- **`OPENBANK_OUTBOX_DISPATCH_ENABLED=true`.** The dispatcher defaults to `false` fleet-wide.
  Without it the grant lifecycle never leaves the outbox table — no error, `attempt_count` stays 0
  — every enforcement projection stays permanently empty, and delegated access simply never works
  while everything reports healthy.
- **`QUARKUS_REDIS_HOSTS` is a hard boot dependency**, not an optimisation: the idempotency store
  makes `ReactiveRedisDataSource` active, and its absence is an `InactiveBeanException` at boot.
- **`AUTHZ_ENFORCE` stays `false`** for the first rollout per the `rules.yaml` guardrail. Flipping
  it before the sidecar is confirmed serving fails closed on every call.

## NetworkPolicies are generated, not written

`openbank-infra/scripts/gen-network-policies.py` derived the delegation policies and the matching
ingress rules in `accounts`, `payments`, `pid`, `sca`, `kafka` and `keycloak` from the svc URLs
declared as env in the Deployment. Every one of those six diffs is delegation being added as an
allowed source — checked individually, no unrelated drift swept in.

## The image pin is a real sha for an image that does not exist yet

Stated in the manifest itself, not only here. delegation-service has never been built, so ECR holds
no tag for it. Argo will report `ImagePullBackOff` until the first build lands, and **that build
must be dispatched explicitly**:

```bash
gh workflow run auto-deploy.yml -f services=openbank-delegation-service
```

The push path only matches `<svc>/src/main` or `build.gradle.kts`, and this PR touches neither, so
merging it does not trigger a build. The auto-deploy gitops PR then rewrites the pin to the built
sha. The `deploy-drift-declaration` gate requires `sandbox-<hex>` and has no escape hatch for
"not yet built", which is why a placeholder was not an option.

## Gates

The whole `gitops` group passes (11/12). `loki-rule-load-test` reports `UNFALSIFIED` — its own
self-test verdict is wrong — and it does so identically against a clean `origin/main` worktree, so
it is pre-existing and unrelated. Filed separately rather than folded in here.

## Still not enough on its own

Merging this plus #3412 (customer-edge routes) plus the dispatched build is what makes delegated
access reachable from the app. Also still open: no Pact contracts (#2991), no `onBehalfOf` audit
threading in the money leg, no notification on lifecycle transitions.

Closes #2992
Refs #3382, #3412


JiRaska:
## The new Dockerfile collides with #3392, and neither PR's CI can see it

This PR adds `openbank-delegation-service/Dockerfile` with a Gradle build stage. #3392 (open, `fix/dockerfiles-runtime-only`) rewrites the other 52 to runtime-only and adds a gate that rejects exactly that shape. I ran its gate against this file:

```
exit 1
::error:: openbank-delegation-service/Dockerfile: has a `FROM ... AS <stage>` build stage.
          Images are built from .github/workflows/Dockerfile.deploy against a host-side quarkus-app/
1 problem(s) across 53 Dockerfiles
```

The two branches are independent, so neither run knows about the other. Whichever merges second goes red:

- **#3392 first** → this PR fails the new gate and needs the Dockerfile rewritten.
- **This PR first** → #3392's own gate finds 1 problem and it has to fix delegation's file too.

Cheapest fix is here and now: make this file match what #3392 lands, before either merges.

## The build stage is dead text regardless

Worth knowing independently of who merges first. Nothing builds these files — `auto-deploy.yml`, `ghcr-publish.yml` and `build-push-service.sh` each generate their own context from `.github/workflows/Dockerfile.deploy` and read exactly one thing from the per-service file: the `EXPOSE` line. So of the 28 lines added here, `EXPOSE 8126` is the only one that reaches anything.

This copy is at least *more* honest than the ones #3016 measured — it does `COPY build-logic`, `gradle.properties` and the ADR-0122 libs siblings, which is precisely what made 0 of 51 of the old ones unbuildable. But "buildable and never built" is still a second definition of how the image is made, free to drift from the one that is real.

## The rest of the PR

No objection — the gitops component, the `ALL_SERVICES` entry, the Kafka mTLS ExternalSecrets and the OIDC secret are what a new service needs to deploy, and delegation-service landed on `main` in #2971 without any of it. This is the follow-through, not new surface.

*Flagged while checking the PR queue; neither branch touched.*

Issue #3349, opened 2026-08-02
libs: the four-eyes self-approval guard is unfalsified — the interceptor test re-implements it in its own fake, and no test ever feeds it a maker approving themselves


JiRaska:
## Summary

The maker-checker self-approval guard — the mechanism that makes four-eyes actually four-eyed —
has **no test anywhere in the repo that feeds it the case it exists to reject**. The production
guard in `RedisApprovalStore.decide` is correct on inspection, but nothing would go red if it were
deleted. It is unfalsified in the sense the root `CLAUDE.md` warns about: every artefact describing
it looks covered.

This is the shared `openbank-libs` four-eyes path, so it is fleet-wide: 18 services reference
`ApprovalStore` / `four_eyes_required` / `X-Approval-Id`, including account, ledger, transaction,
balance, the four payment services, lending, sanctions and clearing.

## The guard

`openbank-libs-runtime/src/main/kotlin/com/openbank/libs/approval/impl/RedisApprovalStore.kt:79`

```kotlin
if (decidedBy == approval.makerId) throw SelfApprovalNotAllowedException(approval.makerId)
```

`openbank-libs-domain/.../approval/ApprovalStore.kt` documents it as deliberately living here
rather than in each REST layer: *"enforced here, not just by the caller's REST layer, so
segregation of duties holds even if a service's endpoint forgets to re-check it."* That design is
right — every service's `ApprovalResource` correctly delegates without re-checking, and so depends
entirely on this one line.

## Why the existing tests do not cover it

**1. `AuthorizeInterceptorTest` re-implements the guard in its test double.**
`openbank-libs-runtime/src/test/kotlin/com/openbank/libs/authz/AuthorizeInterceptorTest.kt:493`
declares `FakeApprovalStore`, and at line 524 that fake carries a verbatim copy of the production
line:

```kotlin
if (decidedBy == approval.makerId) throw SelfApprovalNotAllowedException(approval.makerId)
```

So the interceptor tests exercise a copy, not the code that decides in production. Deleting line 79
of `RedisApprovalStore` leaves this entire suite green.

**2. That copy is never even invoked.** All ~20 test methods in the file were read. They cover:
202 + pending-approval creation, consuming a valid approval id, an approval id belonging to a
DIFFERENT maker, a still-PENDING id, re-deciding an EXECUTED approval, and `markExecuted` outside
APPROVED. None calls `decide()` with the maker as `decidedBy`. Line 524 is dead code inside the
[...]

Issue #2183, opened 2026-07-25
The money-path 2-approval rule is documented, not enforced: main-protection has required_approving_review_count 0


JiRaska:
## Finding

`CLAUDE.md` non-negotiable #7 and `openbank-libs/governance/rules.yaml` state that a change to a **money-path service** requires **2 approvals** plus a threat model. The `main-protection` ruleset does not implement the approval half:

```
rule: pull_request
  required_approving_review_count: 0
  require_code_owner_review: false
  require_last_push_approval: false
```

So a money-path PR merges with **zero reviews** the moment `all-green` passes.

## How it surfaced

PR #2175 (`openbank-balance-service` — listed in `rules.yaml: money_path_services`) was written expecting to be unmergeable without review; the PR body said so explicitly. It merged with 0 reviews via plain `gh pr merge --squash --auto` — no `--admin`, no override flag, nothing bypassed. The rule simply is not there to bypass.

That change was read-only reporting, so the blast radius was nil. The point is that nothing distinguished it from one that moves money.

## Two adjacent observations from the same ruleset read

1. **`strict_required_status_checks_policy: false`** — "require branches up to date" is off. Per the notes in `CLAUDE.md`, this is the *only* remaining lever against the frozen-`base.sha` race (#481 × #524) that works on a personally-owned repo, since merge queue is unavailable (422 `Invalid rule 'merge_queue'`). It is currently disabled.
2. **`Admin UI build` is neither a required check nor inside `all-green`** — it was red on `main` throughout, and `all-green` passed anyway (see #2165). A PR touching admin-ui can merge with its UI never having been built in CI, which is what happened to #2175's UI half.

## This is a decision, not a defect to auto-fix

Tightening branch protection on a single-maintainer repo has a real cost: with one maintainer, `required_approving_review_count: 2` makes money-path changes unmergeable without inviting reviewers, and there is no admin-free path around it. That may well be why it is set to 0.

The options, roughly:
- **Accept and document.** Amend `rules.yaml` and `CLAUDE.md` so the 2-approval rule reads as an aspiration gated on a second maintainer, not as an active control. Cheapest, and stops the docs overstating the controls — which matters for the OpenSSF/audit claims (cf. #1924, whose Gold blockers are the same bus-factor root).
- **Enforce it** and accept that money-path work needs a second human. Requires solving the maintainer-count problem first.
[...]

PR #3420, opened 2026-08-02
docs(governance): say that money-path two-approval is aspirational, not enforced

JiRaska:
## Summary

`rules.yaml` claimed the money-path two-approval rule was enforced by branch protection. It is not,
and never has been. Rather than turn on a requirement this repository cannot satisfy, this states
the enforcement status, the blocker, the compensating controls, and the condition under which it
gets turned on.

## Type of change

- [ ] fix (bug fix)
- [ ] perf (performance)
- [x] docs
- [ ] chore (build / tooling)
- [ ] breaking change

## Linked issues

Refs #2183, #3349

## The measurement

`main-protection` (ruleset `18325357`) today:

```
required_approving_review_count: 0
require_code_owner_review:        false
dismiss_stale_reviews_on_push:    true
```

Every money-path PR merged since #2183 was filed on 2026-07-25 has **zero** approvals — #3378,
#3197, #3057, #3050, #3044, #3032, #3006, #2962, #2374. Nine of nine. A 0% compliance rate over
nine days is not drift; the control does not exist.

`dismiss_stale_reviews_on_push: true` is the part that misleads: the ruleset is configured to
*dismiss* stale approvals while requiring none, which reads as protection to anyone scanning it.

## Why not simply require 2

The repository has exactly one account with write access, and GitHub does not permit approving your
own pull request. Setting the count to 2 — or even to 1 — makes `main` unmergeable except through
`gh pr merge --admin`, and roughly half of all merges are opened by `openbank-gitops-bot` (deploys,
release-please), which cannot obtain an approval at all.

The outcome would not be a stricter control. It would be a permanently unsatisfiable gate that
forces the admin bypass this repo's own rules forbid, and in doing so removes the signal that
anything was overridden. **An admitted gap is a better audit position than an unsatisfiable
requirement routinely bypassed.**

## What this PR does instead

- Deletes the false `# CONTRIBUTING.md + branch protection` comment.
- Records, in place: enforcement status, why it is not simply enabled, the compensating controls
  that ARE independently verifiable (the five required checks, ADR-0030 threat models, the
  money-path-scoped gates in `gates.yaml`, and the runtime four-eyes gate — with a pointer to
  #3349, where that gate's self-approval guard turns out to have no test that ever exercises it).
- Adds machine-readable keys so a future gate can assert this instead of parsing prose:
  `enforcement_status: aspirational_not_enforced`, `enforcement_blocker:
  single_write_access_account`, `enforcement_tracking_issue: 2183`.
- **Revisit trigger keyed to a fact, not a date:** when a second account holds write access, set
  `required_approving_review_count: 1` with `require_code_owner_review: true`, confirm a bot PR can
  still merge, and only then raise money-path to 2.

## Regeneration

All 40 `gen-*opa-bundle*.sh` re-run **after** the final `rules.yaml` edit (#2457's lesson — running
them before a later edit embeds superseded text and reddens the OPA gate on every bundle), then:

- re-run a second time: byte-identical, so the generators are idempotent here;
- 39 bundles confirmed to contain the new `aspirational_not_enforced` text, so the regeneration
  demonstrably propagated rather than producing a vacuous no-diff;
- service manifests differ **only** in the `openbank.tech/policy-checksum` pod-roll annotation
  (`8f455914ec6024b2` -> `6ef160b0682ff4d9`);
- `rules.yaml` parsed with a duplicate-key-rejecting loader — CI never lints this file, and
  SnakeYAML silently keeps the last of a repeated key.

## Reviewer notes

- This PR has a short shelf life: any competing `rules.yaml` or `.rego` change conflicts all 68
  generated files. Merge with `--auto`, not `--admin`.
- The change is prose plus three scalar keys — no policy logic moves. The bundle diff is the same
  text embedded 39 times.
- **Correction to an earlier version of this description**, which suggested opening a gate that
  asserts "a money-path PR carries a threat model for each touched service". That gate already
  exists and is `mode: enforced`: `threat-model-coverage`
  (`openbank-infra/scripts/check-threat-models.py`) derives its scope from
  `rules.yaml: money_path_services` and requires `docs/threat-models/<service>.md` to exist and be
  non-trivial — at least 500 bytes and mentioning STRIDE or one of its categories, so a stub cannot
  satisfy it. Building a second one would have duplicated a live rule, the #3009 failure mode.
  The claim was made without checking; the compensating controls listed in `rules.yaml` are
  therefore *more* enforced than that bullet implied, not less.

- The gap that does remain is narrower, and this PR does not close it: `threat-model-coverage`
  asserts a threat model **exists** for a money-path service, not that it is **current** with
  respect to the change in front of it. The gate nearest to that,
  `threat-model-updated-on-trust-boundary-change`, is `mode: advisory`. Measured today on
  sanctions-service: the threat model existed, satisfied the enforced gate throughout, and covered
  only the import side — it said nothing about `sanctions.clear`, the action `rules.yaml` calls the
  highest-risk in that service, for nine months (fixed in #3378). Flipping that second gate to
  enforced is plausible but should be measured first: an advisory gate switched on blind can block
  the whole fleet.


JiRaska:
## `gates (gitops)` — one line in `gitops/apps/delegation.yaml` fixes it

```
delegation-opa-bundle.yaml: 261935 bytes, 1839 over what client-side apply can carry, but
Application 'delegation' does not set ServerSideApply=true.
```

The bundle embeds `rules.yaml`, so the prose this PR adds lands in all 65 of them — and delegation's
is the one already closest to the ceiling. Fix, matching `accounts.yaml:50`:

```yaml
    syncOptions:
      - CreateNamespace=false
      - ServerSideApply=true      # ← add
```

## Worth knowing why this is not a pre-existing break

I first read it as "main is already broken" and was wrong, so recording the measurement rather than
the conclusion. The gate does **not** compare the YAML file size: it computes `applied_size`, the
compact JSON kubectl writes into `last-applied-configuration`.

```
delegation bundle on main:  file 267846 bytes  →  applied_size 259444
effective ceiling:          262144 − 2048 reserve = 260096
```

So `main` sits ~650 bytes under, and **this PR is what crosses it**. That also means a standalone
"add ServerSideApply" PR would be green against main with or without the line — unfalsifiable — so
the fix belongs here, where the gate can actually prove it.

## On the PR itself

No objection to the substance. `required_approving_review_count: 0` matches what I read from the
ruleset API today, and stating an unenforced control as unenforced is the right direction —
especially given #3164 and #3411 both leaned on "money-path needs two approvals" this week while
nothing was requiring them.

PR #3448, opened 2026-08-02
fix(lending): compliance-pack activation never worked — three defects behind one 403

JiRaska:
The ADR-0212 D4 workflow shipped complete — endpoints, `@RolesAllowed`, the maker-checker invariant, persistence, the console — and had **never once been executed**. OPA denied `lending.compliance.propose` outright (#3402), so the first request ever to reach the service body did so on 2026-08-02, against the sandbox. It found three defects stacked behind that denial.

## 1. No session

`JpaCompliancePackActivationRepository` carried neither `@WithTransaction` nor `@WithSession`:

```
HTTP 422  No current Mutiny.Session found
```

The endpoint maps business-rule failures to 422, so a missing annotation reads as a *domain rejection* rather than a wiring fault.

## 2. Persist, not merge

`save()` called `persistAndFlush` on an entity whose `@Id` is assigned by the application (`Ids.newId()`), not `@GeneratedValue`. A non-null id cannot tell Hibernate transient from detached, so it schedules an INSERT for both — and the **checker** leg, which loads a PROPOSED row, flips its state and saves, dies at flush with `duplicate key value violates ... _pkey`. Same defect as consent-service (#1521) and standing-order (#2079).

## 3. Activate-before-persist

`decide()` called `registry.activate(decided)` **inside the map, before `activations.save()`**. When the save failed, the caller got a 500 and reasonably concluded that nothing had happened — while that pod was already enforcing the pack, with no row to show for it. A restart loses it (boot rehydrates from the table) and sibling replicas never had it, so **the same loan application is judged against different rules depending on which pod answers**.

This one was not looked for. It fell out of a **contradiction between two tests** in a run meant to prove something else: with defect 2 still in place, step 3 failed with 500 while step 4 still found the pack active. A green run of the fixed code would have left it in.

## Coverage

`CompliancePackActivationIT` drives maker and checker through the real REST endpoints against a Testcontainers Postgres and asserts over plain JDBC.

It has to be an integration test. A mocked repository cannot see any of this, and a real-DB test that only *proposes* cannot see defect 2 either — the INSERT is correct on the maker leg; it takes a **second write to the same row**. Maker and checker are therefore separate ordered tests rather than one happy-path helper.

It also asserts what must **not** happen: the maker cannot decide (422, and the row must not move), and the decision must UPDATE rather than insert a second row — which only `count(*)` can distinguish, never a read by id.

## Falsified three ways, not assumed

| run | result |
|---|---|
| fix applied | **4/4 pass** (+ 7/7 existing unit tests still green) |
| `origin/main`, unfixed | **4/4 fail** |
| annotations fixed, `persistAndFlush` kept | **step 3 alone fails, HTTP 500** |

The middle run alone would have been misread: steps 2 and 3 failed only by cascading on an uninitialised `proposalId`, so defect 2 was **hidden behind defect 1, not demonstrated**. The third run is what measures it.

## What this says about the 403

The OPA denial did not act as a gate, it acted as a **cover**. It kept three defects out of sight in code that every test layer called finished, and the feature read as shipped for its whole life. Worth remembering the next time an authorization fix "just" unblocks something.

Refs #3402, ADR-0212 D4

### DIFF
--- a/openbank-lending-service/src/main/kotlin/com/openbank/lending/application/port/out/CompliancePackActivationRepository.kt
+++ b/openbank-lending-service/src/main/kotlin/com/openbank/lending/application/port/out/CompliancePackActivationRepository.kt
@@ -6,7 +6,10 @@ package com.openbank.lending.application.port.out
 
 import com.openbank.lending.infrastructure.persistence.entity.CompliancePackActivationEntity
 import com.openbank.libs.governance.ProposalState
+import io.quarkus.hibernate.reactive.panache.Panache
 import io.quarkus.hibernate.reactive.panache.PanacheRepositoryBase
+import io.quarkus.hibernate.reactive.panache.common.WithSession
+import io.quarkus.hibernate.reactive.panache.common.WithTransaction
 import io.smallrye.mutiny.Uni
 import jakarta.enterprise.context.ApplicationScoped
 import java.util.UUID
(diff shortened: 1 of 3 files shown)


JiRaska:
Read the diff against the claims. The three fixes are right, and the merge one matches the shape this repo has now paid for three times (`Panache.getSession().flatMap { it.merge(entity) }`, as consent #1521 and standing-order #2079). The falsification matrix is the part that makes this reviewable — particularly the third row, since the middle run alone would have credited defect 2 to a cascade.

I checked the one thing a description cannot establish about itself: **whether the new IT actually runs in CI.** It does.

- `openbank-lending-service/build.gradle.kts` has no test filtering — the `excludes` block there is kover coverage filtering, not `Test.filter`.
- The file sits in the default `src/test` source set, so `:test` picks up `*IT.kt` with no extra wiring.
- No `@Disabled`, `@EnabledIf`, or `Assumptions` anywhere in it — nothing that could make it skip itself silently.
- `:openbank-lending-service:test` executed (not `UP-TO-DATE`, not `FROM-CACHE`) for 79 s in this PR's build.

Worth saying how that was established, because my first attempt was worthless: I grepped the job log for `CompliancePackActivationIT`, got 0, and nearly read that as "it did not run". The log names **no** test class at all — grepping for an existing one (`CompliancePackActivationServiceTest`) also returns 0. Silence there is the probe failing, not the test missing.

## One residual, not a blocker

Defect 3's stated harm survives the fix, in the half the reordering cannot reach.

`CompliancePackRegistry` holds an `AtomicReference<Map<PackKey, List<CompiledCompliancePack>>>` — per pod, in memory, rehydrated at boot by `CompliancePackConfig`. After this change the approving pod activates once the row is committed, which closes the phantom case. But a sibling replica still learns about the pack only at **its** next boot, so:

> the same application would be judged against different rules depending on which pod answered

is still reachable — now by replica lag rather than by a failed save.

Not reachable *today*: `openbank-infra/gitops/components/lending/lending-service.yaml` sets `replicas: 1`. Which is exactly why it deserves an issue rather than a comment — the day someone scales lending for load, a four-eyes compliance control starts diverging per pod, and nothing in CI or at runtime would say so. Nothing in the tracker covers it that I could find.

## Pre-existing, flagging while we are here

`decide()` reads, checks `require(entity.state == PROPOSED)`, then writes in a separate reactive context. Two checkers approving concurrently can both pass that guard and both merge. Not introduced here — the read-modify-write shape predates this PR, which only moved `registry.activate` — but it is the four-eyes invariant on a money-path service, so it is worth an explicit decision rather than inheritance.

## Merging

`openbank-lending-service` is in `rules.yaml: money_path_services`, so this needs 2 approvals and I am not one of them. Leaving it for a human. All five required checks are green and it is `MERGEABLE`.

JiRaska:
Filed the residual as #3467 — the per-pod registry, and why `replicas: 1` is the only thing holding it. Also folded in the `decide()` read-then-write note there rather than leaving it in a review comment, since both sit on the same four-eyes path and would likely be picked up together.

Nothing for this PR to do about either; it is a strict improvement on both counts.

PR #3461, opened 2026-08-02
feat(keycloak): add a second ROLE_COMPLIANCE principal so four-eyes is completable

JiRaska:
## The realm had exactly one ROLE_COMPLIANCE account

The OPA rule that permits pack activation (`compliance-lending-desk`) requires `ROLE_COMPLIANCE` for **both** legs, and `CompliancePackActivationService` raises `MakerCheckerViolation` when the maker and checker JWT subjects match. With a single such account the two conditions cannot both be satisfied: **ADR-0212 D4 activation is uncompletable by anyone, in any environment.**

That is the third independent blocker on the same control, after the OPA denial (#3402) and the persistence defects (#3448). Each one looked like the whole problem while it was merely the one in front.

## Naming carries no ordering

Either account may propose and either may decide. The service requires only that the two subjects **differ**, which is the property four-eyes actually needs. Binding "maker" and "checker" to fixed identities would make the control *weaker*, not stronger — it would tell an attacker which single account to take over.

## Credentials

Follows the existing pattern: a `__COMPLIANCE2_USER_PASSWORD__` placeholder here, the value only ever in the secret store. Nothing about it is in this repository.

## Deploying this

**Realm import runs on cold start only**, so merging this does not create the account in a running Keycloak. The live sandbox account was created through the admin API separately; this file is what makes it survive a rebuild.

Verified live, including the negative:

```
compliance@openbank.local     enabled=True  roles=[ROLE_COMPLIANCE, ROLE_VIEWER]
compliance2@openbank.local    enabled=True  roles=[ROLE_COMPLIANCE, ROLE_VIEWER]
distinct subjects: True          <- what MakerCheckerViolation compares
login with the stored password:  OK
login with a wrong password:     400
```

Refs ADR-0212 D4, #3402, #3448

### DIFF
--- a/openbank-infra/gitops/components/keycloak/realm-template.json
+++ b/openbank-infra/gitops/components/keycloak/realm-template.json
@@ -413,6 +413,25 @@
       ]
     },
     {
+      "username": "compliance2@openbank.local",
+      "email": "compliance2@openbank.local",
+      "firstName": "Compliance Two",
+      "lastName": "OpenBank",
+      "enabled": true,
+      "emailVerified": true,
+      "credentials": [
+        {
+          "type": "password",
+          "value": "__COMPLIANCE2_USER_PASSWORD__",
+          "temporary": false
+        }
+      ],
+      "realmRoles": [
+        "ROLE_COMPLIANCE",
+        "ROLE_VIEWER"
+      ]
+    },
+    {
       "username": "service-account-openbank-edge",
       "enabled": true,
       "serviceAccountClientId": "openbank-edge",


JiRaska:
## Imported against a real Keycloak, at the version the cluster runs

`CLAUDE.md` asks for this before a realm-template PR, because import happens on cold start only — a broken template ships silently. Ran it against **26.6.3**, which is what `platform` is actually running (`openbank-keycloak:26.6.3-optimized`), not the `24.0.5` that a stale reference in the tree points at.

```
docker run -v realm.json:/opt/keycloak/data/import/realm.json quay.io/keycloak/keycloak:26.6.3 start-dev --import-realm

INFO [ImportUtils] Realm 'openbank' imported
INFO [io.quarkus]  Keycloak 26.6.3 started in 18.422s
```

All 13 `__PLACEHOLDER__`s substitute to valid JSON, and the realm parses.

## The new principal is correctly formed

```
username        compliance2@openbank.local
enabled         true      emailVerified   true
firstName       "Compliance Two"          lastName  "OpenBank"
requiredActions []
realm roles     ROLE_COMPLIANCE, ROLE_VIEWER
credentials     1 × password
```

`requiredActions: []` is the one worth calling out — the KC 26 trap is a user missing `firstName`/`lastName`/`email`, which imports fine and then answers *"Account is not fully set up"* at login with no hint why. This user has all three, so it does not land there.

The role mapping was read back from the running server rather than from the file:

```
GET users/<id>/role-mappings/realm  →  ['ROLE_VIEWER', 'ROLE_COMPLIANCE']
```

## What I could NOT verify, and why it is not this PR

The end-to-end mint. Every direct-grant attempt against my container returned `invalid_grant / Invalid user credentials` — **including** the pre-existing `compliance@openbank.local`, a user I created fresh via `kcadm` with a password I set myself, and the bootstrap `master/admin`.

Three controls failing identically is the signature of a broken harness, not a broken template: if #3461 were at fault, only the new user would fail. Brute-force detection recorded `numFailures: 0`, so the requests were not even reaching credential validation. I did not chase it further — it is my local setup, and the question it would answer is not about this change.

So: **import and realm state verified, login path unverified.** The failure mode `CLAUDE.md` warns about — a template that imports silently broken — is the one that is covered.

*Verified while working the PR queue; branch untouched.*

PR #3466, opened 2026-08-02
fix(gitops): register the openbank-lending Temporal namespace

### LINKED ISSUE #3475: lending-service polls an unregistered Temporal namespace — no origination workflow can start
Measured in the sandbox on 2026-08-02 while sweeping Loki, Grafana and HolmesGPT for false positives and error-log volume.

`lending-service` polls the Temporal namespace `openbank-lending`. `temporal-namespace-registration.yaml` has never created it — the registered set is `openbank-default openbank-payments openbank-fx openbank-statements openbank-settlement openbank-finops openbank-devops openbank-liveness openbank-governance openbank-release openbank-docstruth openbank-authzaudit openbank-flakytest openbank-campaign`.

**Impact**

- No loan origination workflow can start — the ADR-0126 origination path is dead.
- Both the workflow and the activity poller retry forever, producing **~2,000 ERROR lines/hour** into Loki (third-largest error producer in the estate).

**Why nothing went red**

The worker connects to the frontend fine; only the *poll* is refused. The pod stays `2/2` because readiness does not touch Temporal, and ArgoCD stays `Synced/Healthy`. Log volume is the only observable.

**Third occurrence of the same class**

| when | service | how the list drifted |
|---|---|---|
| go-live | settlement | list claimed `openbank-default`; the worker polls `openbank-settlement` (runbook 0006) |
| first rollout | campaign | Deployment set no `OPENBANK_TEMPORAL_NAMESPACE`, so it fell back to the bare `openbank` |
| now | lending | never added |

Each time the hand-written list was wrong, never the service — so the fix adds a gate that derives the required set from the gitops env override (which wins) falling back to the `application.yaml` default, rather than trusting the list.

JiRaska:
## Linked issues

Closes #3475

## What

`lending-service` polls the Temporal namespace `openbank-lending`. The registration Job has never created it.

## Why it matters

Both the workflow poller and the activity poller retried forever against a namespace Temporal does not hold:

- **no loan origination workflow could start** — the whole ADR-0126 origination path was dead;
- **~2,000 ERROR lines/hour** into Loki, making lending the third-largest error-log producer in the sandbox.

Nothing went red. The worker connects to the frontend fine and only the *poll* is refused, so the pod stayed `2/2` (readiness does not touch Temporal) and ArgoCD stayed `Synced/Healthy`. Log volume was the only observable.

Confirmed live before the fix — the registered set was `openbank-finops`, `openbank-flakytest`, `temporal-system` and friends, with no `openbank-lending`:

```
temporal operator namespace list | grep openbank-lending   # no match
```

## The list is the thing that keeps drifting

Third occurrence:

| when | service | how it drifted |
|---|---|---|
| go-live | settlement | list claimed `openbank-default`; the worker polls `openbank-settlement` (runbook 0006) |
| first rollout | campaign | Deployment set no `OPENBANK_TEMPORAL_NAMESPACE`, so it fell back to the bare `openbank` |
| now | lending | never added |

Each time the LIST was wrong, never the service. So this adds `check-temporal-namespace-registration.py`, which **derives** the required set rather than trusting the list:

1. the gitops `OPENBANK_TEMPORAL_NAMESPACE` env — which **wins**, because it overrides the config default at runtime;
2. otherwise the `${OPENBANK_TEMPORAL_NAMESPACE:<default>}` in the service's `application.yaml`.

Precedence is the campaign lesson in miniature: reading only (2) reports a namespace the service does not use; reading only (1) misses every service that never sets the env.

An **extra** registered namespace is deliberately not an error — an idle namespace costs nothing and retention reaps it, and failing on extras would make the gate a maintenance tax with no defect behind it.

## Falsifiability

`--self-test` covers the shape that has bitten each time, the override precedence, the harmless-extra case, and a namespace named only in a comment (a whole-file grep passes there; this must not).

Verified against the **real** defect, not only the fixture:

```
python3 .github/scripts/check-temporal-namespace-registration.py   # unfixed manifest -> exit 1
python3 .github/scripts/check-temporal-namespace-registration.py   # with the fix     -> exit 0
```

Refs #2021

### DIFF
--- a/.github/gates/gates.yaml
+++ b/.github/gates/gates.yaml
@@ -226,6 +226,23 @@ gates:
     run: |
       python3 .github/scripts/check-podmonitor-namespace-coverage.py
 
+  # The Temporal registration Job's NAMESPACES list is hand-written, and it has now been the
+  # thing that drifted THREE times (settlement, campaign, lending). A missing namespace is
+  # invisible to every signal that is normally checked: the worker connects fine and only the
+  # POLL is refused, so the pod stays Ready, ArgoCD stays Synced/Healthy, and the service simply
+  # never runs a workflow — lending burned ~2,000 ERROR lines/hour into Loki for the whole time,
+  # with no origination possible. The checker derives the required set from the gitops
+  # OPENBANK_TEMPORAL_NAMESPACE env (which wins) falling back to the application.yaml default,
+  # so the list can no longer be shorter than reality without going red. ENFORCED.
+  - id: temporal-namespace-registration
+    name: "Temporal namespaces a worker polls are registered (enforced)"
+    group: gitops
+    mode: enforced
+    selftest: |
+      python3 .github/scripts/check-temporal-namespace-registration.py --self-test
+    run: |
+      python3 .github/scripts/check-temporal-namespace-registration.py
+
   # --since scopes the EXPENSIVE half (pull ~100 MB + boot Loki, ~90 s) to PRs that can
   # actually invalidate it. This job is unconditional and required, so an unscoped run charged
   # every PR in the fleet 90 s for a check about files it did not touch. Direct cost is $0
--- a/.github/scripts/check-temporal-namespace-registration.py
+++ b/.github/scripts/check-temporal-namespace-registration.py
@@ -1,0 +1,218 @@
+#!/usr/bin/env python3
+"""Guard: every Temporal namespace a service actually polls must be in the registration Job.
+
+WHY THIS EXISTS: `temporal-namespace-registration.yaml` carries a hand-written, space-separated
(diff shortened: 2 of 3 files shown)


JiRaska:
Ran the checker's `--self-test` from this branch rather than trusting the green: it prints the MISSING case for a namespace it must flag and a clean case for one it must not, in both directions. That is the falsifiability this repo asks for, and it is why the "third time this list drifted" framing lands — the gate replaces the list-shaped hole rather than patching one instance of it.

The `2,000 ERROR lines/hour with nothing red` detail is the part worth keeping in the file: readiness does not touch Temporal, so the pod stays 2/2 and ArgoCD stays Healthy while no workflow can start. Log volume as the only symptom is precisely the class nothing alerts on.

Merging — all five required checks green, `clean`, no failures.

PR #3468, opened 2026-08-02
fix(gitops): serve /api/v1/balances from the balances namespace

### LINKED ISSUE #3476: /api/v1/balances has returned 503 for ~60 days — Ingress backend names a Service in the wrong namespace

`https://api.open-bank.tech/api/v1/balances` returns **503**, and has for roughly 60 days:

```
/api/v1/balances  -> 503
/api/v1/accounts  -> 401   # same host, auth-gated, i.e. working
```

**Root cause**

An Ingress backend resolves only within the Ingress's own namespace. `accounts/accounts-api` routed `/api/v1/balances` to a Service literally named `balance-service` in namespace `accounts` — and balance-service has always lived in `balances`. nginx logged the consequence ~1,700 times an hour:

```
Error obtaining Endpoints for Service "accounts/balance-service":
  no object matching key "accounts/balance-service" in local store
```

That made `ingress-nginx` the fourth-largest error-log producer in the estate.

**Why nothing caught it**

- balance-service itself is healthy — the fault is in the route, not the service;
- in-cluster callers use `http://balance-service.balances.svc:8103` and were unaffected, so every integration path stayed green;
- no probe covers the public path;
- the Ingress object was `Synced/Healthy` throughout, because **a backend that does not resolve is not a manifest error**.

**Note for whoever picks this up**

The NetworkPolicy half is load-bearing and easy to miss: `balances` has no `ingress-nginx` rule, because `gen-network-policies.py` derives one only where an Ingress backend declares it and the declaration was in the wrong namespace. Without regenerating, moving the route just swaps the 503 for a hang.

JiRaska:
## Linked issues

Closes #3476

## What

`https://api.open-bank.tech/api/v1/balances` has answered **503 for ~60 days**. Measured today:

```
/api/v1/balances  -> 503
/api/v1/accounts  -> 401   # same host, auth-gated, i.e. working
```

## Root cause

An Ingress backend resolves only within the Ingress's **own namespace**. `accounts/accounts-api` routed `/api/v1/balances` to a Service literally named `balance-service` in namespace `accounts` — and balance-service has always lived in `balances`. The rule named a Service that has never existed:

```
Error obtaining Endpoints for Service "accounts/balance-service":
  no object matching key "accounts/balance-service" in local store
```

nginx logged that **~1,700 times an hour**, making `ingress-nginx` the fourth-largest error-log producer in the sandbox.

## Why nothing caught it

- balance-service itself was healthy — the fault is in the route, not the service;
- its in-cluster callers use `http://balance-service.balances.svc:8103` and were unaffected, so every integration path stayed green;
- no probe covers the public path;
- the Ingress object was `Synced/Healthy` throughout, because **a backend that does not resolve is not a manifest error**.

## The fix

One Ingress per owning namespace, sharing the `api.open-bank.tech` host — the shape `payments-api` and `sanctions-api` already use. nginx merges them into a single server block; each declares its own TLS secret, as the two siblings do.

Deliberately **not** an ExternalName Service in `accounts`: `tools-gate.test.ts` asserts the ADR-0234 `grafana-tools` Service stays the only `type: ExternalName` in gitops, because it carries the repo-wide `AVD-KSV-0108` trivy baseline — a second one would silently inherit that exemption.

## The NetworkPolicy half — easy to miss, and it decides whether this works

`balances` had **no** `ingress-nginx` rule. `gen-network-policies.py` derives one only where an Ingress backend declares it, and the declaration was in the wrong namespace, so it was never generated. Without it this change would have swapped the 503 for a hang.

Regenerated with the generator rather than hand-edited (never edit a derived file). The only derived change is exactly the expected one:

```diff
+    - from:
+        - namespaceSelector:
+            matchLabels:
+              kubernetes.io/metadata.name: ingress-nginx
+      ports:
+        - protocol: TCP
+          port: 8103
```

No other component's policies moved, which also confirms the broken backend never produced a rule on the `accounts` side.

### DIFF
--- a/docs/threat-models/openbank-balance-service.md
+++ b/docs/threat-models/openbank-balance-service.md
@@ -173,6 +173,32 @@ also be deleted (nothing else in balance-service depends on it).
 
 ## 7. Change log
 
+- **2026-08-02** — **balance-service is now genuinely reachable from the public edge**, and the
+  honest framing is that the *intent* did not change while the *reality* did. `accounts/accounts-api`
+  had declared `api.open-bank.tech/api/v1/balances` since the Ingress was written, so the exposure was
+  always designed and threat-modelled as public — but an Ingress backend resolves only inside its own
+  namespace, and it named `balance-service` in `accounts`, where no such Service exists. The path
+  returned **503 for ~60 days** and nginx logged `no object matching key "accounts/balance-service"`
+  ~1,700×/hour. The route now lives in `components/balances/ingress.yaml` beside the Service it
+  targets (the shape `payments-api`/`sanctions-api` already use on this host).
+  **Trust-boundary delta:** `balance-service-ingress-allow-list` gains one derived rule,
+  `TCP:8103 FROM ingress-nginx` — the first non-`openbank-*` namespace admitted to the service port.
+  Without it the fix would have swapped the 503 for a hang, so it is load-bearing, not incidental;
+  it was produced by `gen-network-policies.py` (which derives ingress-nginx edges from Ingress
+  backends) rather than hand-written, and no other component's policies moved.
+  **Why the residual risk does not move:** every §4 mitigation on S1/I1/E1 is authn/authz at the
+  application layer — OIDC bearer JWT, role-gated mutations, OPA enforced on every endpoint, and the
+  A1 `X-Customer-Party-Id` ownership scoping — none of which depended on the caller being in-cluster.
+  This is the same posture `account-service` already carries on the same host and the same Ingress
+  class. D1 (DoS) is the one row that changes in practice rather than in principle: the new Ingress
+  carries the same `limit-rps: 20` / `limit-burst-multiplier: 3` / `limit-connections: 10`
+  annotations as its three siblings, deliberately identical so one host does not present four
+  different limits depending on which path a client happens to hit. The pre-existing
+  "Gateway rate-limit — infra scope" residual is now actually exercised instead of theoretical.
+  **What this does NOT change:** no application code, no schema, no role, no OPA policy, no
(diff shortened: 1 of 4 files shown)


JiRaska:
Verified the three load-bearing claims against `main` rather than reading them:

- **The backend pointed at nothing.** The only `name: balance-service` anywhere under `components/accounts/` is the Ingress backend reference itself — there is no Service by that name in the namespace. `balance-service` and `balance-service-canary` are both in `balances`.
- **The sibling shape is real.** `payments` and `sanctions-service` each carry an ingress file inside their own component; `balances` carried none. Moving the route there makes it the third instance of a pattern, not a new one.
- **`openbank-balance-service` is in `rules.yaml: money_path_services`.**

The threat-model delta is the part I would have asked for and it is already there, including the honest framing that the *intent* never changed while the *reality* did — the path was designed as public from the day the Ingress was written and simply never worked. Calling out `TCP:8103 FROM ingress-nginx` as the first non-`openbank-*` namespace admitted to the port, and that it came out of `gen-network-policies.py` rather than a hand edit, is the right level of detail: without it the fix trades a 503 for a hang.

**Not merging this one.** It is money-path and it makes a money-path service publicly reachable in practice for the first time — 2 approvals by `rules.yaml`, and this is a human's call regardless of the count. Two checks are still pending; `mergeable_state` reads `blocked`.

One question rather than an objection, for whoever approves: `limit-rps: 20` is inherited from the siblings deliberately, and the PR says the pre-existing "Gateway rate-limit — infra scope" residual moves from theoretical to exercised. Is 20 rps the right number for a balances endpoint specifically, or is it the right number for *consistency across the host*? The PR argues the second explicitly, which is a defensible answer — worth it being a decision rather than an inheritance.

Issue #2370, opened 2026-07-25
admin-ui compliance page: 46 hand-maintained conformance claims with no gate; six GDPR rows cite columns no code reads


JiRaska:
## Summary

`openbank-admin-ui/src/app/docs/compliance/page.tsx` renders a 7-area, 46-row regulatory compliance
matrix — PSD2, AML, KYC, GDPR, EBA ICT, CNB, FATCA/CRS — from a hardcoded `COMPLIANCE_AREAS: ComplianceArea[]`
literal. 43 of the 46 rows are marked `ok`. **Nothing verifies any of them.** No test, no CI gate, no
generator; the only files referencing this route are navigation (`Sidebar.tsx`, `docs/page.tsx`,
`dashboard/page.tsx`).

The result is not merely optimistic — the page is wrong in **both** directions, which is what shows its
status values are uncorrelated with the code rather than just stale.

## Evidence — six cited columns have no application reader

Each GDPR row cites a database column as its evidence. Six of those columns exist **only** in a Flyway
migration and a generated ER diagram. No Kotlin, TypeScript or SQL outside the migration itself reads or
writes any of them (checked against `origin/main` @ `0d669339`):

| Row (status) | Cited evidence | Where it actually exists |
|---|---|---|
| Marketing consent (`ok`) | `marketing_consent boolean V2` | `party-service V2__compliance_fields.sql` only |
| Explicit consent timestamp (`ok`) | `gdpr_consent_at + gdpr_consent_version V2` | `party-service V2` + ER diagram only |
| Right to erasure (soft delete) (`ok`) | `deleted_at + deletion_reason V2` | `party-service V2` only |
| Data retention policy (`ok`) | `data_retention_until in accounts + parties V2` | migrations + ER diagrams only |
| Data sensitivity classification (`ok`) | `data_sensitivity in audit_entries V2` | `audit-service V2` + ER diagram only |

That is **5 of the 7 GDPR rows**, all green, all evidenced by DDL. A column in a migration is not a
control; it is a place a control could store something. `marketing_consent` specifically was verified in
#1331 to have zero readers and zero writers — the ADR-0176 correction note calls it "dead legacy" — and
the live marketing consent is a *different* pair of columns (`consent_marketing`,
`consent_marketing_updated_at`, #1157/#1161) that this page does not mention.

## The page is also wrong in the opposite direction

**"Right to erasure (soft delete)"** understates a control that genuinely shipped, while citing dead
[...]

PR #3486, opened 2026-08-02
feat(ci): a check-* script that nothing runs is now a build failure

JiRaska:
Closes the class behind #3240, which #3434 closed two instances of.

## What was wrong

`check-compliance-matrix.py` was written for #2370 — with a docstring naming the exact regulatory drift it prevents — and referenced from nowhere. `check-operator-write-naming.py` is named by `rules.yaml` as its own `ci_producer` while nothing invoked it, so reading `rules.yaml` gave every impression the rule was enforced. Neither had ever executed.

A check that has never run is worse than an unfalsified one: it cannot go red, never appears in a run list, and its presence in the tree reads as coverage to anyone grepping for it.

## The rule

Every `.github/scripts/check-*` must be one of:
1. in a `gates.yaml` entry's `run:`/`selftest:`,
2. invoked from a tracked file (workflow step, gitops CronJob, another script),
3. in `HELPERS` with a reason — and a `HELPERS` entry wrong in **either** direction is itself a finding, so the exception list cannot quietly become permanent.

Clean at registration: **73 scripts, 64 in gates.yaml, 9 invoked elsewhere, 0 never run.**

## A mention is not an invocation

This is the whole difficulty, and this repo is bitten by it in both directions. Measured while writing this:

| script | appears in | what that is |
|---|---|---|
| `check-dockerfile-no-build-stage.py` | 40-odd service `Dockerfile`s | comments |
| `check-openapi-request-schema-conformance.py` | 38 OPA bundles | embedded `rules.yaml` text |

Both are the string propagating, not the check executing. A substring rule reports them as covered on that evidence — the code-about-code collision running in its **silent** direction, where a false negative reads exactly like a pass. So a reference counts only in a **command position** on a non-comment line.

**Line-start is deliberately not a command position.** It would match any prose line beginning with the script name, and that error marks an orphan as covered. Being too strict instead produces a loud false orphan costing one `HELPERS` entry. When the failure directions are not symmetric, take the loud one.

## Two defects found by the gate's own known-positive tests

Not by review — by feeding it inputs it must and must not flag. Both are pinned as cases now:

1. The first regex accepted a bare path under `.github/scripts`, so it matched the `ci_producer:` **value** in `rules.yaml`. That version would have been green about precisely the #3240 defect.
2. The first self-test asserted only that *something* resolved as covered — and stayed **green** when I broke the manifest lookup to cover everything unconditionally. It now runs the real resolution against a name referenced nowhere *and* against a registered one, and fails on either.

Verified after fixing, by breaking `coverage()` again:

```
selftest FAIL: a script referenced nowhere was NOT reported as an orphan —
the gate would be green about the defect it exists to find.
```

## Self-assessment — what this does NOT catch

Honestly, so nobody reads more into the green than is there:

- **Reachability.** A script invoked from a job with an always-false condition, or excluded on the PR lane, passes. That is the `@PactFolder`-vs-`@PactBroker` problem (#2327), which needed its own gate; it is not solved here.
- **Correctness or falsifiability of the script it registers.** This answers "is anything wired to it", not "does it work" or "can it go red".
- **Scripts not named `check-*`.** The naming convention is the scope. A gate named otherwise is invisible to this.
- **Comment stripping is line-based** and only handles `#` outside quotes. A YAML block scalar containing a `#` inside a multi-line string could in principle truncate a line; the failure direction there is a false orphan, which is loud.

## Verification

```
run-gates.py --only gate-script-registration   PASS  (self-test + gate, 0 orphans)
run-gates.py --group lint                      PASS=4
```

### DIFF
--- a/.github/gates/gates.yaml
+++ b/.github/gates/gates.yaml
@@ -776,6 +776,27 @@ gates:
   # evidence citation is mechanical, and advisory over a published claim means "the page
   # says something the code does not support" is mergeable. Green on the current tree, and
   # `--self-test` (added with this registration) proves the red is reachable.
+  # The meta-gate: every check-* script must be run by SOMETHING. #3240 found two that were not —
+  # check-compliance-matrix.py, written for #2370 with a docstring naming the exact regulatory
+  # drift it prevents and referenced from nowhere, and check-operator-write-naming.py, which
+  # rules.yaml names as its own ci_producer while nothing invoked it. Both had never executed.
+  #
+  # That is worse than an unfalsified gate: a check that has never run cannot go red, never
+  # appears in a run list, and its presence in the tree reads as coverage to anyone grepping for
+  # it. Registering those two was the fix; this is what stops the third.
+  #
+  # ENFORCED and clean at registration. It flags ITSELF until registered — the same
+  # self-reference check-advisory-gate-registration.py hit in #2450 — which is the cheapest
+  # possible demonstration that its red is reachable.
+  - id: gate-script-registration
+    name: "every check-* script is actually run (#3240, enforced)"
+    group: registry
+    mode: enforced
+    selftest: |
+      python3 .github/scripts/check-gate-script-registration.py --self-test
+    run: |
+      python3 .github/scripts/check-gate-script-registration.py --enforce
+
   - id: compliance-matrix
     name: "compliance-page evidence citations (#2370, enforced)"
     group: registry
--- a/.github/scripts/check-gate-script-registration.py
+++ b/.github/scripts/check-gate-script-registration.py
@@ -1,0 +1,286 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: Apache-2.0
+"""Every `check-*` script must actually be RUN by something. Nothing may be written and forgotten.
+
+WHY THIS EXISTS
+---------------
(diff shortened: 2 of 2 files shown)


JiRaska:
Ran it against the tree with `main` merged in, since I registered a new gate (`gitops-duplicate-resources`, #3474) an hour ago and wanted to confirm the two compose rather than collide:

```
76 check-* script(s) — 67 in gates.yaml, 9 invoked elsewhere or declared, 0 never run.
--self-test: 6 invocation shapes flagged, 4 bare mentions not flagged,
             3 comment cases both ways, orphan reported while a registered script stays covered.
```

The bare-mention cases are the ones that make this worth having. A registration check is exactly the kind that matches the prose *about* a script rather than an invocation of it — this repo has been bitten in both directions (`check-roles-allowed-realm.py` flagging a KDoc quote, `check-advisory-gate-registration.py` flagging its own step name), and a false pass there reads as coverage. Testing both directions of the comment handling is the difference between a gate and a gate-shaped file.

`HELPERS` being wrong in either direction is the right call too — same reason `KNOWN_UNCOVERED` in the pact replay checker is. An exception list that only fails when it is too short becomes permanent.

Merging: 5/5 required green, `clean`, no failures.

PR #3490, opened 2026-08-02
fix(infra): give delegation-db its backup credentials, and enforce the gate that said so

### LINKED ISSUE #3489: delegation-db has no backups — Pod Identity association missing, and the gate that flagged it was advisory

`delegation-db` shipped today with a `barmanObjectStore`, a `ScheduledBackup`, and **no EKS Pod Identity association**. Every WAL archive has failed from the first minute:

```
Barman cloud WAL archive check exception: Unable to locate credentials
```

The database has **no recovery point at all**.

**Diagnosis (not assumed — the same way campaign was confirmed)**

```
accounts-db-1   env | grep AWS_   -> AWS_CONTAINER_CREDENTIALS_FULL_URI, AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE
delegation-db-1 env | grep AWS_   -> (nothing)
```

So this is a missing association, **not** the #1759 admission-injection race that a pod restart fixes. IMDS is unreachable from pods (verified on both), so there is no fallback credential path — the archive could only ever fail.

**The part that matters more than the missing line**

`check-db-backup-associations.py` exists precisely to prevent this. It **flagged delegation-db correctly** and printed the exact one-line fix. The change merged anyway, because the gate is advisory.

This is the **fourth** occurrence of the defect that gate was built to end:

| when | clusters | outcome |
|---|---|---|
| #1444 | `sdd-db`, `tpp-registry-db`, `vop-db` | every WAL archive failed; sdd-db for three days |
| after | `campaign-db` | ~48h of `PostgresWALArchiveFailing` |
| today | `delegation-db` | no backups, and **the gate had already said so** |

An advisory verdict is right where a human still has judgement to exercise. Here there is none left — *"this database declares a backup and cannot possibly authenticate to perform it"* is not a style opinion, it is a database with no recovery point, and there is no failure until someone needs a restore. Advisory only made it mergeable.

**Follow-up not covered by the fix**

`goalert-db` backs up to `s3://openbank-cnpg-backups-265175468565/goalert-db`, outside the Terraform-managed bucket, so its lifecycle, encryption and retention are not set by `db-backups.tf`. That stays a warning deliberately — it is a governance question with a real choice in it — but it is unresolved.

JiRaska:
## Linked issues

Closes #3489

## What

`delegation-db` shipped today with a `barmanObjectStore`, a `ScheduledBackup`, and **no EKS Pod Identity association**. Every WAL archive has failed since the first minute:

```
Barman cloud WAL archive check exception: Unable to locate credentials
```

The database has **no recovery point at all**.

## Diagnosed, not assumed

Confirmed the same way `campaign` was:

```
accounts-db-1    env | grep AWS_  -> AWS_CONTAINER_CREDENTIALS_FULL_URI, AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE
delegation-db-1  env | grep AWS_  -> (nothing)
```

A missing association — **not** the #1759 admission-injection race that a pod restart fixes. IMDS is unreachable from pods (verified on both), so there is no fallback credential path; the archive could only ever fail.

## The half that matters more

`check-db-backup-associations.py` exists precisely to prevent this. It **flagged `delegation-db` correctly** and printed the exact one-line fix. The change merged anyway — the gate was advisory.

Fourth occurrence of the defect that gate was built to end:

| when | clusters | outcome |
|---|---|---|
| #1444 | `sdd-db`, `tpp-registry-db`, `vop-db` | every WAL archive failed; sdd-db for three days |
| after | `campaign-db` | ~48h of `PostgresWALArchiveFailing` |
| today | `delegation-db` | no backups, **and the gate had already said so** |

An advisory verdict is right where a human still has judgement to exercise. Here there is none left: *"this database declares a backup and cannot possibly authenticate to perform it"* is not a style opinion — it is a database with no recovery point, and there is no failure until someone needs a restore. Advisory only made it mergeable.

## Enforcement details

`mode: enforced` **alone would have been vacuous** for this checker — it prints `::warning` and exits 0 unless passed `--enforce`. The repo has paid for that distinction once already, so both halves are set.

Verified falsifiable against the **real** defect, not a fixture:

```
--enforce, unfixed tree      -> exit 1
--enforce, delegation added  -> exit 0
```

`goalert-db` stays a `::warning` under `--enforce` by design: a cluster backing up to a bucket outside Terraform is a governance question with a real choice in it, which is exactly where advisory belongs.

## Apply note

`sandbox-platform` has no CI apply pipeline. The association is applied out-of-band; merging this PR is the mandatory bookkeeping that keeps `main` aware of infra that is already live.

### DIFF
--- a/.github/gates/gates.yaml
+++ b/.github/gates/gates.yaml
@@ -341,13 +341,30 @@ gates:
   # in exactly that state (issue #1444); sdd-db had been failing every WAL archive for three
   # days. NOT diff-scoped: it runs against whole repo state, because the gap is a MISSING
   # pair, and a PR that adds only the gitops half would not touch the .tf side at all.
-  # ADVISORY for now — add --enforce to flip.
+  #
+  # ENFORCED since 2026-08-02, and the reason is that advisory demonstrably did not work: on
+  # that day delegation-db shipped in exactly this state, the gate printed the warning naming
+  # the cluster and the one-line fix, and the change merged anyway. That is the fourth
+  # occurrence of a defect this gate was built to end.
+  #
+  # An advisory verdict is right where a human still has judgement to exercise. Here there is
+  # none left: "this database declares a backup and cannot possibly authenticate to perform
+  # it" is not a style opinion, it is a database with no recovery point, and the failure is
+  # invisible until someone needs a restore. Advisory only made it mergeable.
+  #
+  # Scope note — `mode: enforced` ALONE is vacuous for this checker (the repo has paid for
+  # that once already): it prints ::warning and exits 0 unless passed --enforce, so both
+  # halves are needed and both are set below. Verified falsifiable against the real defect,
+  # not a fixture: --enforce exits 1 on the unfixed tree and 0 with delegation added. The
+  # goalert-db finding stays a ::warning under --enforce by design — a cluster backing up to
+  # a bucket outside Terraform is a governance question with a real choice in it, which is
+  # exactly where advisory belongs.
   - id: db-backup-association-gate
-    name: "db-backup association gate — every backed-up CNPG cluster has pod identity (advisory)"
+    name: "db-backup association gate — every backed-up CNPG cluster has pod identity (enforced)"
     group: gitops
-    mode: advisory
+    mode: enforced
     run: |
-      python3 openbank-infra/scripts/check-db-backup-associations.py
+      python3 openbank-infra/scripts/check-db-backup-associations.py --enforce
 
   # A probed or scraped port must be a port the service opens (issue #2872, item 1).
   #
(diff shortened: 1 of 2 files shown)


JiRaska:
Falsified the enforced gate against the real defect rather than a fixture, since flipping advisory → enforced is only worth anything if the enforced path actually fails:

```
association present   -> exit 0, "✓ every cluster targeting the managed bucket has an association"
association removed   -> exit 1, and it NAMES the cluster plus the exact line to add:
    ::error delegation-db (namespace delegation) declares barmanObjectStore ->
    s3://openbank-sandbox-db-backups/ but has no aws_eks_pod_identity_association...
    Add `delegation = { namespace = "delegation", sa = "delegation-db" }` to local.db_backup_clusters.
restored              -> exit 0
```

The `goalert-db` finding does stay a `::warning` under `--enforce`, as the comment claims — so the flip does not drag an unrelated governance question into the blocking set.

The scope note is the part I would have checked for and it is already written down: `mode: enforced` in `gates.yaml` is vacuous on its own for this checker, because it prints `::warning` and exits 0 unless passed `--enforce`. Both halves are set here. That distinction has cost this repo before — a sweep for `continue-on-error: true` finds one form and silently reports the other eleven as enforced.

And the framing is right: this is the fourth occurrence, and the first where the gate had already said so and the change merged anyway. An advisory verdict is for where a human still has judgement to exercise; "declares a backup and cannot authenticate to perform it" has none left in it.

Merging: 5/5 required green, nothing pending, nothing failing.

PR #3499, opened 2026-08-02
ci(gates): elect one Gradle cache writer and gate the writer budget

JiRaska:
## Summary

The Actions cache pool has been over its 10 GB ceiling continuously, so GitHub evicts LRU
without anything going red. This measures the hit rate first (as #3299 asks), then elects a
single Gradle cache writer per OS+architecture and adds a gate so the writer set cannot grow
back silently.

## Type of change

- [x] chore (build / tooling)

## Linked issues

Refs #3299

## The measurement (acceptance criterion 1)

Not assumed — parsed `setup-gradle`'s own restore/save lines out of the **raw logs of 40
recent runs** (`/actions/runs/<id>/logs`), across Fleet lint, Dependency submission, Security
scan, Services CI and CI.

| entry | hit | miss | hit % | saved |
|---|---|---|---|---|
| Gradle User Home | 16 | 0 | 100% | 16 |
| `gradle-dependencies-v1` | 15 | 1 | **93.8%** | 1 |
| `gradle-build-cache-v1` | 16 | 0 | 100% | **16** |
| `gradle-transforms-v1` | 16 | 0 | 100% | 2 |
| wrapper-zips / kotlin-dsl / groovy-dsl / instrumented-jars / generated-gradle-jars | 16 each | 0 | 100% | 0 |

**127 of 128 extracted-entry restore attempts hit — 99.2%.** The single miss was a ~850 MB
`dependencies` entry.

Two notes on how this number was reached, because both nearly produced a wrong answer:

- **My first probe could not express a miss.** It grepped for `not found` / `empty home`
  wording that `setup-gradle` never prints, so it reported a clean 100%. The real vocabulary
  is `Cache hit for: <key>` and `Did not restore <name> with key ...`. The 1 miss only
  appeared after re-running with the action's actual strings.
- **The raw counts double-count by exactly 2×.** A run's log zip contains both a job-level and
  a step-level copy of every line, differing only in the sub-millisecond timestamp. The table
  above is deduped; the ratios were unaffected.

## What this means

The cache is *working* — this is not a "cache is broken" fix. The cost is what #3299 says it
is: eviction as the steady state makes wall-clock depend on what else ran that day.

**A janitor would not have helped.** Every entry in the pool had been accessed within the
preceding ~30 minutes, i.e. the pool already holds nothing but live entries — GitHub's own LRU
has already deleted everything deletable. The filler is the **write rate**:

- each writing job-run stores a fresh ~200 MB `gradle-build-cache-v1` entry, **every run**
  (16/16 in the sample);
- plus an entire new ~700–850 MB `gradle-dependencies-v1` entry whenever the resolved set
  shifts by even one artifact — the entry is content-addressed over the whole `modules-2`
  tree, so two jobs resolving almost-the-same set still store it twice.

That is how **9 near-identical dependency entries came to hold 6.1 GB — 60% of the pool**.

## Changes

- **`dependency-submission.yml`: `cache-read-only: true`** (was `${{ ref != main }}`, i.e. a
  writer on main). Exactly two jobs write on main — `fleet-lint` and this one, 8 sampled runs
  each. This is `setup-gradle`'s own prescription for this shape ("Select which jobs should
  write to the cache"): elect one writer per OS+arch, make the rest consumers.
  `gradle-home-cache-strict-match` defaults to false, so it still **restores** fleet-lint's
  home across the job boundary.
- **New `.github/scripts/check-gradle-cache-writers.py`** + registration in
  `.github/gates/gates.yaml` (`lint` group, `enforced`, with a self-test).
- No `rules.yaml` / `.rego` edit, so **no OPA bundle restamp** — this PR is 3 files and will
  not conflict with a governance PR.

### Why `dependency-submission` is the right job to demote

It is the only fleet-wide Gradle job **nobody waits on**: not a required check,
push/schedule-triggered, and measured at **4.8–8.5 min against a 30 min timeout**. The cost of
a colder cache is re-downloading the runtime-only artifacts `assemble` pulls beyond
fleet-lint's `testClasses` set — minutes, inside that headroom. `generate-and-submit` resolves
the fleet regardless of cache state, so the #1449 stale-graph alarm is unaffected.

Deliberately **not** touched:

- **CodeQL** — its java-kotlin leg documents a 25–30 min cold penalty and is the most
  preemption-exposed job in the fleet.
- **`auto-deploy` / `ghcr-publish`** — `ubuntu-24.04-arm`, a **Linux-ARM64 cache scope with no
  other writer**, so demoting them means permanently cold, not shared. Their entries are also
  the smallest in the pool (~7 MB across 3).
- **`_service-ci.yml`** — untouched on purpose; its `cache-disabled` on ARC is the ~$200/mo NAT
  egress fix and must not be re-armed.

### The gate, and why it isn't just a comment

`cache-read-only` **defaults to `${{ github.ref_name != default_branch }}`** — so a workflow
that adds `setup-gradle` with *no* cache inputs silently becomes a writer on main. Nothing
counted them. The gate declares all 13 usages with a mode and a required justification, and
fails on drift in **either** direction: undeclared, changed, or stale. A `WRITER_BUDGET`
ratchet set to today's count (5) means growing the writer set is a deliberate act. This is
acceptance criterion 3.

It classifies honestly rather than conveniently — `writes-on-main` (silence, the default),
`writes-always` (a literal `cache-read-only: false`, which *overrides* that default and writes
on feature branches too — `auto-deploy` is the one instance), `read-only`, `disabled`,
`conditional` (an expression decides; `_service-ci` never writes only because two
complementary expressions cover each other, which no per-step rule can see).

## Exercising the failure path

Per this repo's hardest-won CI rule, the gate was run against cases it **must** flag, not just
until it went green:

| case | verdict |
|---|---|
| **revert this PR's own fix** in the real tree | `DRIFT` + `BUDGET`, exit 1 |
| add a new workflow with bare `setup-gradle` | `UNDECLARED` + `BUDGET`, exit 1 |
| declared `read-only` that silently became a writer | flagged |
| an expression mistaken for `read-only` | flagged |
| a declaration with an empty reason | flagged |
| a stale declaration (usage removed) | flagged |
| `cache-read-only: false` vs silence (the subtle distinction) | both flagged when mis-declared |
| a correctly declared consumer / `setup-java` without `cache: gradle` | **not** flagged |
| advisory mode (no `--enforce`) on a failing tree | exit **0**, as the repo's advisory convention requires |

The 10-case `--self-test` runs immediately before the gate via the runner, so the gate is
skipped entirely if its own red becomes unreachable.

## Definition of Done

- [x] `python3 .github/scripts/run-gates.py --group lint` — 5/5 PASS
- [x] `license-header-consistency` and `advisory-gate-registration` gates PASS
- [x] No Gradle build needed or run (CI-config-only change)
- [ ] Unit tests — N/A (the gate's `--self-test` is the test)

## Security checklist

- [x] No new `@Suppress`.
- [x] No new third-party dependency (stdlib + PyYAML, already used by sibling gates).

## Compliance impact

- [x] None

## Rollout / Rollback

- Deployment strategy: N/A (CI configuration).
- Rollback plan: revert the commit. The gate then flags its own revert as `DRIFT`, so the
  revert must also drop the declaration — deliberate, and the point of the two-way check.

## Reviewer notes

**The projected saving is a projection, and I want to say so plainly.** Removing
`dependency-submission`'s writes should remove roughly half the `build-cache` churn (~2 GB) and
some of the 9 dependency entries (~2–3 GB), landing the pool around 7 GB. I could not verify
that from a PR branch — the effect is only observable after several days of main-branch runs,
because branch caches live in a separate scope. **Re-check with
`gh api /repos/JiRaska/open-bank-oss/actions/cache/usage` in ~3 days**; if it has not dropped
below ~9 GB, the next lever is `gradle-home-cache-excludes: caches/build-cache-1` on
`fleet-lint`, which is a straight ~2 GB for a real recompile cost.

Two other honest limits:

- The 16-job sample is a **few hours** of a very busy repo, all on `main`. It does not cover
  the arm64 lanes or a quiet period, and a 1-in-16 miss has a wide confidence interval.
- I did **not** delete any existing cache entries. The pool will drain on its own via the 7-day
  retention; a one-off purge would have been undone by the next hour of CI and would have
  obscured whether the structural change worked.

### DIFF
--- a/.github/gates/gates.yaml
+++ b/.github/gates/gates.yaml
@@ -536,6 +536,23 @@ gates:
     run: |
       python3 .github/scripts/check-workflow-run-step-size.py
 
+  # Gradle Actions-cache writer budget (issue #3299). The pool has been over its 10 GB
+  # ceiling continuously (11.7 GB / 92 entries on 2026-08-02), so GitHub evicts LRU and the
+  # only symptom is somebody's job re-resolving dependencies — never a red check. What
+  # fills it is the number of jobs WRITING: each writing run stores a ~200 MB build-cache
+  # entry and, on any shift in the resolved set, a whole new ~700-850 MB dependencies entry.
+  # Nothing counted the writers, so a workflow adding `setup-gradle` with no cache inputs
+  # silently became one (the default writes on main). This declares every usage's mode and
+  # fails on drift in EITHER direction — new, changed, or removed.
+  - id: gradle-cache-writer-budget
+    name: "Gradle Actions-cache writer budget (issue #3299, enforced)"
+    group: lint
+    mode: enforced
+    selftest: |
+      python3 .github/scripts/check-gradle-cache-writers.py --self-test
+    run: |
+      python3 .github/scripts/check-gradle-cache-writers.py --enforce
+
   # co-deploy set derivation unit test (issue #1985). The classifier above says WHAT KIND
   # of block a service has; this one says whether the block can be resolved by deploying
   # that service at all. Two services that name each other never converge one at a time,
--- a/.github/scripts/check-gradle-cache-writers.py
+++ b/.github/scripts/check-gradle-cache-writers.py
@@ -1,0 +1,416 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) OpenBank contributors. Licensed under the Apache License, Version 2.0.
+# See LICENSE in the repository root or https://www.apache.org/licenses/LICENSE-2.0 for details.
+"""Every Gradle-cache consumer in CI declares, here, whether it may WRITE the cache.
+
+Background (#3299). The repo's GitHub Actions cache sits permanently over its 10 GB
+ceiling — measured 2026-08-02 at 11.7 GB across 92 entries — so GitHub evicts
+least-recently-used entries continuously. Nothing goes red: eviction shows up as a job
+re-downloading dependencies, never as a failed check.
+
+What actually fills it is the WRITE rate, not stale junk. Every entry in the pool had
+been accessed within the preceding ~30 minutes, which means the pool is already nothing
+but live entries and a janitor would have nothing to delete. Each cache-writing job-run
+stores a fresh `gradle-build-cache-v1` entry (~200 MB) and, whenever the fleet's resolved
+dependency set shifts by even one artifact, an entire new `gradle-dependencies-v1` entry
+of ~700-850 MB — the entry is content-addressed over the whole `modules-2` tree, so two
+jobs resolving almost-the-same set still store it twice. That is how 9 near-identical
+dependency entries came to hold 6.1 GB, 60% of the pool.
+
+So the lever is the number of jobs writing, and `setup-gradle`'s own documentation
+prescribes the shape ("Select which jobs should write to the cache"): elect ONE writer
+per OS+architecture and make every other job a consumer. `gradle-home-cache-strict-match`
+defaults to false, so consumers still restore across job boundaries.
+
+WHY A GATE AND NOT A COMMENT
+    The budget is only stable while something notices a new entrant. A workflow that adds
+    `setup-gradle` with no cache inputs writes by default, silently, and the only symptom
+    is somebody else's job getting slower days later. #3131's nightly sweep was written
+    without `cache: gradle` deliberately, and nothing in the tree recorded that intent.
+
+    So this is a DRIFT check in both directions, not an allowlist. Every usage is declared
+    below with the mode its YAML must produce. A new usage fails as undeclared; a declared
+    usage whose YAML changes mode fails as drift; a declared usage that disappears fails as
+    stale. None of the three can be resolved by editing this file alone without saying what
+    changed.
+
+    It does NOT check `runs-on`. Whether the runner is self-hosted is a property of the
+    expression, not of the step, and the ARC lane's NAT-egress reason for disabling the
+    cache entirely is documented at its own call site.
+
+Usage:
+    check-gradle-cache-writers.py             # warn
+    check-gradle-cache-writers.py --enforce   # fail
+    check-gradle-cache-writers.py --self-test # prove it can fail
+    check-gradle-cache-writers.py --list      # print the derived modes (to update DECLARED)
+"""
+
+from __future__ import annotations
+
+import argparse
+import pathlib
+import sys
+import tempfile
+
+import yaml
+
+WORKFLOW_DIR = pathlib.Path(".github/workflows")
+
+SETUP_GRADLE = "gradle/actions/setup-gradle"
+SETUP_JAVA = "actions/setup-java"
+
+# ---------------------------------------------------------------------------
+# THE DECLARED BUDGET
+#
+# key   "<workflow file>::<job id>"
+# mode  what the step's cache inputs must resolve to. One of:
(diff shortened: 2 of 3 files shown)


JiRaska:
## Independent verification — and why I am NOT merging this tonight

I commissioned this work and then checked it rather than taking it on trust. What I confirmed
myself:

- 26/26 checks pass, 8 skipped, commit `verified=true`, `mergeStateStatus: CLEAN`
- the self-reported bug (success message printing a stale writer count) **is** fixed —
  `write_capable(found)` is called dynamically at both call sites, no cached tuple
- `--self-test` exists and covers four cases: new writer, drift, stale entry, missing reason
- the only behavioural change is `dependency-submission.yml` going unconditionally
  `cache-read-only`; the rest is a new gate plus declarations

## Why it waits for a human

**Its headline benefit is a projection, not an observation.** ~7 GB saved is argued, not measured —
branch caches live in a separate scope, so no PR run can demonstrate it. It needs a re-check of
`gh api /repos/JiRaska/open-bank-oss/actions/cache/usage` in ~3 days; if it has not dropped below
~9 GB, the fix was insufficient.

**One load-bearing assumption is unverified, and the author says so:** that fleet-lint's Gradle home
is a good enough superset of what `dependency-submission`'s `assemble` needs. It is argued from task
graphs, not measured. If wrong, that job re-downloads more each run. Sized against a 30-min timeout
at 4.8–8.5 min today, so there is headroom — but "should not break" is doing real work in that
sentence, and this job feeds the PR-time CVE gate (`block_on_cve_severity`), which is a security
control.

**And today already showed what green checks do not cover.** My own #3392 passed 83 checks and broke
the live admin-ui deploy, because no PR check builds that image. This PR's effect likewise only
materialises on `main`, after merge, where no check watches it.

None of that makes the change wrong — I think it is right, and the gate is genuinely falsifiable,
which is more than my Dockerfile guard managed. It makes it a change that deserves a human looking
at it rather than an agent merging its own commission at night.

## Suggested follow-up regardless of this PR

The real defect was that **nobody was watching the number**. A scheduled probe recording
`cache/usage` over time would have caught the overflow before it became the steady state, and this
PR does not add one.

JiRaska:
Falsified it against a real edit rather than the fixture, by flipping one declared consumer to a writer:

```
--enforce, one consumer flipped to cache-read-only: false
  ::error DRIFT   dependency-submission.yml::submit is declared `read-only`
                  but its YAML now resolves to `writes-always`
  ::error BUDGET  6 jobs can write the Gradle cache ... over the budget of 5
  exit 1
--enforce, restored          exit 0
--self-test                  exit 0
```

Both halves are set — `mode: enforced` **and** `--enforce` in `run:`. Worth calling out because `mode: enforced` alone is vacuous for a checker of this shape (it prints `::warning` and exits 0 without the flag), and this repo has shipped that mistake before. My own first probe here ran the script without `--enforce`, read exit 0, and briefly looked like the gate did not fire — it was the probe, not the gate.

The design I would have argued for is what is already here: the check catches the *drift* separately from the *budget*. A budget alone would let a job silently change its write behaviour while the count stayed at 5, which is exactly how a hand-kept list stops meaning anything.

Measuring the hit rate out of 40 runs' raw logs before electing a writer is the right order, and it is the part that makes the budget number defensible rather than arbitrary.

One thing I did not verify: whether 5 is the right ceiling as opposed to merely the current count. The PR frames raising it as a deliberate act that must be justified, which is the correct treatment either way, so this is a note rather than an objection.

Merging: 5/5 required green, `clean`, nothing pending or failing.

Issue #2549, opened 2026-07-26
auto-deploy reconcile can ask the broker about a sha the push lane never published (PENDING_BUILD churn)


JiRaska:
## Problem

A fleet-wide change leaves the sandbox behind indefinitely, and nothing retries. Measured today on `9f138c133` (#2475, the `ROLE_SERVICE` → `ROLE_API` sweep across 29 modules):

| signal | value |
|---|---|
| Services CI main-push run for that sha | **still `queued` after 3 hours** — 4 of 29 service builds done |
| pact versions published for that sha | **none** (`GET /pacticipants/openbank-account-service/versions/9f138c133` → 404) |
| Auto deploy for that sha | `failure` — `can-i-deploy` blocked every money-path service |
| workloads running code that contains the sweep | **4 of 55** |

Nothing here is broken in the "someone shipped a bug" sense. Every piece behaves as designed. The composition is the defect.

## The three parts

**1. The main-push lane serialises on one ARC runner.** `openbank-build` had exactly one runner pod alive (`openbank-build-kw79z-runner-57tbv`) while 29 service builds queued behind it, at roughly 45 min each. A fleet-wide change therefore needs ~a day to publish its pacts — during which every one of those services reports "no pacts for this version".

**2. `can-i-deploy` reads that absence as "not deployable".** Correct by its own contract (ADR-0092): a version with no verification results is not safe to deploy. But the reason is *"the build has not finished yet"*, not *"the contract regressed"*, and the verdict is identical either way.

**3. Auto deploy triggers on `push` only.** It ran at 07:43, found no pacts, failed — and there is no `workflow_run` trigger, no schedule, and no reconcile loop, so **it never retries once the builds finish**. The services stay on the old image until some future commit happens to touch them again. `record-deployment-on-merge.yml` records deployments; nothing re-attempts a blocked one.

So a fleet sweep is guaranteed to leave the fleet behind, and the only thing that clears it is a human noticing and re-dispatching per service.

## What it cost this time — nothing, and that is the point

The sweep is backward-compatible (`ROLE_OPERATOR` stayed in every annotation), so 51 workloads running the old code behave identically. The gap is invisible *this* time. The next fleet-wide change that is not backward-compatible finds the same trap with real consequences.

## Proposed

Roughly in order of cost:

[...]

Issue #1985, opened 2026-07-23
auto-deploy can-i-deploy is per-service — a tangled/patchwork deploy graph needs deploy-together orchestration (#1009 item 3)


JiRaska:
## Problem

`auto-deploy.yml`'s contract gate runs `can-i-deploy --to-environment sandbox` **per service**, checking each candidate against the versions **currently deployed**. This cannot unwind a *tangled* deploy state: when the deployed graph is a patchwork of different `main` shas (as happens after any deploy-freeze), a consumer's pact is verified against its providers at one sha while a **different, older** provider sha is actually deployed — so `can-i-deploy` finds no verified path and blocks, for **every** service in the interdependent cycle, indefinitely. The connected component can only advance if it deploys **together** at a consistent, mutually-verified version, which the per-service gate structurally cannot express.

This is the design gap flagged as **item 3 of #1009** ("a broken pact chain anywhere currently freezes deploys everywhere in the affected provider's fanout").

## Evidence (2026-07-23)

The deploy-freeze from the flaky pact-CLI download (fixed in #1947) left the payments graph as a patchwork: `sepa-payment`/`transaction`@`e3a4c19d`, `fraud`@`31e6692a`, `balance`@`6376c3fc`, `account`@`2799ad80`, `ledger`@`768a6f73` — all different shas. A full-fleet `auto-deploy` dispatch deployed 11 leaf services but left `transaction`, `sepa-payment`, `sepa-instant` blocked, and re-dispatching does **not** converge (each service's latest pact is verified against provider shas that are never the deployed ones). Confirmed `can-i-deploy` for the `{sepa, transaction, fraud}` set at their *latest* (mixed) versions → **blocked**.

`sepa-payment` (missing #844 for ~12 days) was only deployable by a **manual co-deploy**: find a single `main`-push sha where the whole connected component published pacts AND is mutually verified (`6376c3fc`: `can-i-deploy {sepa,transaction,fraud}@6376c3fc --to sandbox` → *"All required verification results are published and successful"*), pin all three gitops manifests to that sha at once (#1979), bypassing the incremental gate. That worked, but it is a hand-driven workaround that requires broker archaeology and gate-bypass each time.

## Ask

Give `auto-deploy` a **deploy-together** mode for the interdependent set, so a frozen/patchwork graph can be caught up without manual co-deploy. Candidate approaches (design, not prescriptive):

[...]

PR #3501, opened 2026-08-02
fix(gitops): a co-deploy set must not be derived from builds that merely haven't finished

JiRaska:
## What

`derive-codeploy-set.py` names the services that must deploy **together**, and the run then prints the command to co-deploy them. It never saw the block **class**, so services that were only *waiting to be built* looked exactly like services that mutually deadlock.

## The observation

Run [30761740836](https://github.com/JiRaska/open-bank-oss/actions/runs/30761740836) (2026-08-02 18:42Z, the scheduled reconcile):

```
[openbank-account-service]       MONEY-PATH blocked by can-i-deploy [PENDING_BUILD]
[openbank-domestic-payment]      MONEY-PATH blocked by can-i-deploy [PENDING_BUILD]
[openbank-fraud-service]         MONEY-PATH blocked by can-i-deploy [PENDING_BUILD]
[openbank-sepa-instant]          MONEY-PATH blocked by can-i-deploy [PENDING_BUILD]
[openbank-sepa-payment]          MONEY-PATH blocked by can-i-deploy [PENDING_BUILD]
[openbank-swift-service]         MONEY-PATH blocked by can-i-deploy [PENDING_BUILD]
[openbank-transaction-service]   MONEY-PATH blocked by can-i-deploy [PENDING_BUILD]
[openbank-card-issuance-service]            blocked by can-i-deploy [PENDING_BUILD]

::error::CO-DEPLOY SET — [account domestic-payment fraud sepa-instant sepa-payment swift
transaction] block each other; no per-service deploy order converges.
Re-run: gh workflow run auto-deploy.yml -f services='…' -f codeploy=true
```

**All eight were `PENDING_BUILD`** — not one `UNVERIFIED`, not one `REGRESSION`. That class means *"no pact version published for this commit yet, the main-push build has not finished"*, and `probe-pact-version.sh`'s own header calls it self-clearing, absorbed by the 3-hourly reconcile.

So the pipeline instructed an operator, with full authority, to run a **weaker check over seven money-path services** because CI was behind.

That inverts the script's stated purpose. Its header says the recurring hand-driven co-deploy bypass **is the risk it exists to replace** — manufacturing the recommendation from transient state is the same bypass with the pipeline's name on it.

## The fix

- `auto-deploy.yml` records the class it *already computed*: `===SERVICE <name>\t<class>`.
- `derive-codeploy-set.py` admits a service to a component only when its block is **durable** (`UNVERIFIED`, `REGRESSION`). A record with no class keeps today's behaviour, so older captures still parse. `UNKNOWN` is excluded from the other direction: an unclassified block is not positive evidence, and a weaker gate needs positive evidence.
- A transient block is **reported, never silently dropped** — a new `PENDING` finding printed as a notice. "No co-deploy set" must not be ambiguous between *nothing is deadlocked* and *the answer is not knowable yet*.

## Verification, and what it does not prove

Three unit cases added to the existing harness (11 total, all pass):

| case | expectation |
|---|---|
| `PENDING_BUILD` pair | **not** a co-deploy set → `PENDING` |
| `UNVERIFIED` pair | still a co-deploy set |
| mixed durable + transient | names no set — half the evidence is "CI is behind" |

All three fail against `origin/main`. **Being precise about why:** the old script cannot parse the class-bearing header at all, so it produces empty output for all three — that is format incompatibility, not by itself proof of the behavioural change. The claim these cases actually pin is the new one: *with the class present*, a transient pair is not named while a durable pair still is. The old behaviour is preserved for class-less input, which case 7 (unchanged) still covers.

`run-gates.py --only co-deploy-set-derivation-unit-test` → PASS.

## A guard caught me mid-change

Adding this pushed the `can-i-deploy` step's `run:` script to **19006** chars and `check-workflow-run-step-size.py` failed the build at its 19000 limit — the #3135 shape, where an oversized script makes the *whole* workflow unparseable with no error. The rationale moved into the script header where it costs nothing; the step is back to **18580**.

## Why this matters tonight

`fraud-service` (#3376) is undeployed and was named in that co-deploy set. Its block is `PENDING_BUILD` — it needs its build to finish, **not** a seven-service money-path co-deploy. This change stops the pipeline recommending the latter.

## Linked issues

Refs #1985, Refs #2549

### DIFF
--- a/.github/scripts/derive-codeploy-set.py
+++ b/.github/scripts/derive-codeploy-set.py
@@ -49,14 +49,19 @@ Usage:
   derive-codeploy-set.py <changed-services-space-separated> < blocks
 
   stdin is a stream of records, one per blocked service:
-      ===SERVICE <name>
+      ===SERVICE <name>[\t<block-class>]
       <that service's verbatim can-i-deploy CLI output>
+
+  The block class is what classify-can-i-deploy-block.sh decided. It is OPTIONAL: without
+  it every blocked service is eligible for a set (the original behaviour). With it, only a
+  DURABLE block counts — see CODEPLOY_CLASSES.
 
   <changed-services> is every service in THIS run (deployable or not), so a counterpart
   can be told apart from one outside the run.
 
 Prints, one finding per line, TAB-separated. Two record types:
     CODEPLOY\t<svc> <svc> ...        a mutually-blocking set that must deploy together
+    PENDING\t<svc> <svc> ...         blocked only because their builds have not finished
     EXTERNAL\t<svc>\t<counterpart>   blocked on a service this run cannot move
 
 Always exits 0 — this is a reporter, not a gate.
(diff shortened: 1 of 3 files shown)


JiRaska:
Ran `test-derive-codeploy-set.sh` rather than trusting the green:

```
ok   no blocked services => no output
ok   PENDING_BUILD pair is not a co-deploy set
ok   UNVERIFIED pair is still a co-deploy set
ok   mixed durable+transient pair names no set
```

The mixed case is the one that matters most and it is covered — a set that is partly durable and partly waiting on CI is the shape most likely to slip through a rule written as "any block counts".

The defect is worth restating because of what it was recommending: on run 30761740836 the tool printed, with full authority, a co-deploy command over seven money-path services, on evidence that meant only "the main-push build has not finished". A co-deploy is a *weaker* check, and this is the tool that exists to replace the hand-driven bypass — so it was inviting exactly the action it was built to make unnecessary. Requiring positive evidence of a durable block, rather than merely the absence of success, is the right inversion.

Excluding `UNKNOWN` from the other direction is the part I would have argued for if it were missing. An unclassified block is not evidence of anything, and the safe default over money-path services is to say so rather than name a set.

Worth noting my own probes here were the weaker instrument: hand-typed records did not reproduce the UNVERIFIED case because the counterpart detection reads the verbatim CLI output, so my approximation quietly produced no set and would have read as agreement. The harness in the PR is what actually measures it.

Merging: 5/5 required green, `clean`, nothing pending or failing.

PR #3508, opened 2026-08-02
fix(governance): embed only the OPA-read subset of rules.yaml in the policy bundles

### LINKED ISSUE #3357: A rules.yaml edit restamps ~65 gitops files, so governance PRs re-conflict faster than they can be resolved
## The problem

Any PR that edits `openbank-libs/governance/rules.yaml` is effectively unmergeable during working hours.

`rules.yaml` is embedded verbatim into every service's OPA bundle ConfigMap and hashed into its pod-roll checksum annotation, so a one-line governance edit restamps ~65 files under `openbank-infra/gitops/components/`. `main` receives `chore(gitops): auto-deploy …` and `chore(admin-ui): deploy …` commits every few minutes, and those touch the **same** component files. The PR therefore re-conflicts continuously.

**Measured on #2971 (delegation-service, 2026-08-02):** resolved three times in one session — take main's side for the generated files, re-run all 39 generators, verify idempotent, push. Each resolution held for **minutes**. Between the second and third, `main` advanced 4 commits during the resolve itself; auto-merge was armed and never got a non-conflicting moment to fire. Every conflict was in generated gitops files — not once in code or in `rules.yaml`.

## Why the obvious workaround does not work

"Leave the regenerated bundles out of the PR and regenerate after merge" cannot pass CI. `OPA policy gate` (a **required** check) runs every generator and then:

```
git diff --exit-code -- openbank-infra/gitops/components
```

So a PR that changes `rules.yaml` without the restamped bundles is stale by construction and goes red. Dropping the `rules.yaml` change instead would drop the point of the PR. There is no way to author such a PR that does not touch the contended files.

## Suggested direction

Move the regeneration to where the contention is not: **regenerate on `main` after merge** rather than requiring it inside the PR.

Roughly: the gate on a PR verifies that the bundles are consistent with the policy source *as of the PR's base*, and a post-merge job on `main` re-runs the generators and commits the restamp (the same shape as the existing `chore(gitops)` bot commits, which already write to `main` unattended).

**The cost has to be stated plainly:** between a `rules.yaml` merge and the post-merge regeneration, `main` carries bundles that do not match their source. That window is the price, and it needs a deliberate answer — at minimum the post-merge job must be non-skippable and must escalate if it fails, or the guarantee the current gate provides is simply lost rather than moved. A red push-triggered workflow on `main` is addressed to nobody (this repo has learned that one before), so it would need to raise or refresh an issue like `fleet-attestation.yml` does.

Alternatives worth weighing first, since both are cheaper:

- **A quiet-window convention** — resolve and land governance PRs when `main` is idle. Works today, needs no change, but it is a human protocol, not a property of the system, and it is exactly the kind of thing that fails on the day it matters.
- **Make the bot's deploy commits not touch bundle files.** The pod-roll checksum annotation lives in each `*-service.yaml` alongside the image pin, which is why an image bump and a policy restamp collide. Separating them (annotation in its own manifest) would remove the overlap without weakening any gate. This is probably the smallest real fix.

I would look at the third option before the first.

## Impact

Not specific to #2971 — it applies to every future `rules.yaml` edit, which is the intended path for governance changes (`rules.yaml` is authoritative; CI enforces it). Merge queue would also solve it, but is unavailable: this repo is owned by a personal account and a `merge_queue` ruleset returns `422 Invalid rule 'merge_queue'` (#1465).

Refs #2971

JiRaska:
## Summary

Every service OPA bundle embedded `openbank-libs/governance/rules.yaml` **verbatim** — 168 KB, 2358 lines — and hashed it into that service's pod-roll checksum annotation. It now embeds `openbank-libs/governance/rules-opa-data.yaml` (21 KB): the top-level keys of `rules.yaml` that some `.rego` actually reads as `data.rules.<key>`, extracted as raw text. An edit to any of the ~30 keys OPA never reads now changes **nothing** under `openbank-infra/gitops/`.

## Type of change

- [x] fix (bug fix)

## Linked issues

Closes #3357

## Changes

- **New** `.github/scripts/gen-rules-opa-data.py` — derives the subset. The key set is read out of the `.rego` sources (`data.rules.<key>`, comments stripped), never hand-listed. Has `--check` and `--self-test`.
- **New** `openbank-libs/governance/rules-opa-data.yaml` — the generated artifact. Never hand-edit.
- 39 of the 40 `gen-*opa-bundle*.sh` repointed at it (one line each); all 39 bundles + pod-roll annotations regenerated.
- `opa-policy.yml` `verify` regenerates the subset **before** the bundle generators and fails on drift; `rules-opa-data.yaml` and the generator added to the path-relevance pattern.
- `rules.yaml: opa_bundle_sync` records the indirection; root + `openbank-infra` `CLAUDE.md` and `check-opa-bundle-apply-size.py`'s docstring corrected (they all asserted the verbatim embed).

## Why this and not the alternatives

Three options were weighed; the full write-up is on #3357.

- **Post-merge regeneration on `main`** (the issue's own suggestion) buys a window where `main`'s bundles do not match their source, on the artifact that decides authorization. It needs a new non-skippable job with its own escalation path, and it moves the guarantee rather than removing the need for it.
- **Splitting the checksum annotation out of `*-service.yaml`** removes the overlap with the auto-deploy bot's image bumps, which is the mechanical half of the conflict. It leaves 39 restamped ConfigMaps that still say nothing, and leaves review dishonest. Worth doing as a follow-up; it is not the root cause.
- **This one** removes the restamps that were never about policy at all. It is also the only one that touches the 262144-byte annotation ceiling.

## Evidence

**Frequency.** Over the last 80 revisions of `rules.yaml`, **16** changed something a policy reads. The other 64 restamped ~78 files for data OPA never consults.

**Semantic equivalence — measured across all 39 bundles, not argued.** For each bundle, `rules-data.yaml` was parsed on `origin/main` and on this branch:

- the five embedded keys (`authz`, `feature_flags`, `four_eyes`, `money_path_action_prefixes`, `money_path_services`) parse to values **deep-equal** to the ones on `main`;
- the 34 removed keys are referenced by **no** `.rego` in the tree;
- no other ConfigMap member changed;
- `opa test` over both policy trees: **168/168**.

The extraction is byte-preserving rather than a YAML re-dump precisely so that claim can be made, and the generator re-asserts per-key deep equality before it writes. The prover was validated against a known-positive: tampering with one bundle's `four_eyes` block makes it report `value of 'four_eyes' changed`.

**Bundle size.** Largest bundle 278762 → 123676 bytes. Bundles over 200 KB: **39 → 0**. `check-opa-bundle-apply-size.py` still passes; its pairing rule is deliberately unchanged — this buys headroom, not an exemption.

**Falsifiability.** `--self-test` passes; `--check` was proven to go red by appending a key to the committed file and restoring from a `cp` backup, and by adding `data.rules.slo` to a throwaway `.rego` (the subset went stale, i.e. the key set really is derived).

**Workflow validated against GitHub, not a parser.** Oracle checked both ways on throwaway branches (since deleted): the edited `opa-policy.yml` produced **0** runs (accepted — its `push` filter is `main`), a deliberately-malformed copy produced **1** run titled `.github/workflows/opa-policy.yml`. `actionlint` finding set is identical to `origin/main`'s (one pre-existing SC2001). yamllint finding *set* on `rules.yaml` is identical to `origin/main`'s.

## Not repaired here, on purpose

`rest.rego` reads `data.rules.shared_m2m_write_prohibition`, but that key is nested under `change_requirements:` in `rules.yaml`, so it is **undefined in the cluster today** and the `prohibited` rule it guards (GHSA-58jq-9hq3-66jr's mitigation) has never fired. It stays undefined; the generator warns about it on every run. Promoting it would be a semantic policy change smuggled inside a mechanical one, and it needs its own PR and its own review. Filed separately.

This also does not make a *policy-relevant* governance edit conflict-free — an edit to one of the five read keys still restamps the fleet, and correctly so.

## Definition of Done

- [x] Derived data is never hand-edited (the new artifact is generated; the gate enforces it).
- [x] `opa test` 168/168; `check-opa-bundle-apply-size.py`, `check-license-headers.py`, `check-gate-script-registration.py`, `check-workflow-run-step-size.py` pass.
- [x] Docs updated (both `CLAUDE.md`s, `rules.yaml`).
- [ ] N/A: no Kotlin, no REST API, no DB, no Avro schema.

## Security checklist

- [x] No new `@Suppress`.
- [x] Authorization data path changed **in transport only** — the values OPA evaluates are proven byte-equal. Please read the equivalence evidence above rather than the 78-file diff.

## Compliance impact

- [x] None — no policy decision changes.

## Rollout / Rollback

- Deployment strategy: rolling. Every service's `policy-checksum` moves once, so the fleet rolls its pods onto an identically-behaving bundle. `subPath` mounts do not hot-reload, which is what the annotation is for.
- Rollback plan: revert the commit and re-run the generators — the bundles are fully derived, so the revert is mechanical.

## Reviewer notes

The diff is 115 files and only 7 are hand-written: the new generator, the new artifact, `opa-policy.yml`, `rules.yaml`, two `CLAUDE.md`s and one docstring. The other 108 are generated — 39 generators changed by one path each, 39 bundles, 39 annotations. Read the equivalence evidence, not the bundles.

Short shelf life by construction. Merging with `--auto`.


JiRaska:
## The one non-mechanical file in this diff: `docs/runbooks/svc-mcp.md`

The `service runbook drift` gate went red on the first run, and it was right — but not because of anything the runbook generator does wrong *now*. Removing the verbatim `rules.yaml` embed **exposed a false positive that had been there the whole time**.

`generate-service-runbooks.py`:

```python
for f in comp.rglob("*.yaml"):
    t = read(f)
    if "barmanObjectStore" in t and (short in t or f.parent.name.startswith(short)):
        return True
```

That is a whole-file substring test. `rules.yaml` line 253 contains the string `barmanObjectStore` (in the prose of the `db-backup association gate` rule), and `mcp` appears in it too — so `components/mcp/mcp-opa-bundle.yaml`, which embedded all 168 KB of it, satisfied **both** halves. `backup_configured("mcp")` therefore returned `True` for a database that has **no `barmanObjectStore` anywhere in its component directory**, and the committed runbook promised on-call:

> **RPO target:** ≤ 5 min (continuous archiving). **RTO target:** ≤ 30 min

for a cluster with no backup at all. The regenerated file now says the truth: RPO/RTO undefined, DR not achievable, enabling the CNPG backup is the blocking prerequisite.

Two notes for the reviewer:

- **This is the repo's own documented failure mode, running in its silent direction** — a check greps a file for a string and matches the *prose about* the thing rather than the thing, and a false negative here reads exactly like a pass. It is the same class as the `middleware.ts` / `?tool=grafanaX` pair in `CLAUDE.md`.
- **The exposure is incidental, and the weakness is not fully fixed.** With `rules.yaml` no longer propagating into 39 ConfigMaps the specific collision is gone, but `short in t` over a whole file is still the wrong test — any future file in a component directory that happens to contain both strings will do it again. Fixing it means parsing for `kind: Cluster` and reading `spec.backup.barmanObjectStore` structurally, the way `prod-readiness-collector.py` already scopes its C5 detection to `Cluster` documents. That is a behaviour change to a different generator and belongs in its own PR, not smuggled into a mechanical one — filed separately.

`mcp-service`'s missing backup is a real gap, independent of this PR and unchanged by it. This PR only stops the runbook lying about it.

PR #3519, opened 2026-08-02
fix(ci): restore three gates a stale branch silently deleted, and stop it recurring

JiRaska:
## An incident I caused, restored — plus the guard that stops it recurring

PR #3492 squash-merged tonight and silently removed **three enforced gates** from `.github/gates/gates.yaml`:

| gate | added by |
|---|---|
| `scheduled-trigger-emitted` | #3484, hours earlier |
| `gitops-duplicate-resources` | #3474 |
| `gate-script-registration` | #3486 |

**Nobody removed them.** #3492 was branched from an older `main`, and every new gate is appended near the same anchor in this one file, so its copy of `gates.yaml` replayed over the others'. GitHub reported the PR `MERGEABLE` / `CLEAN`, all five required checks green. Three enforced gates stopped running and nothing said anything.

That PR was mine, so this is a self-report, not a review finding.

`gates.yaml` is now the highest-collision file in the tree **precisely because the manifest design worked**: adding a gate is one entry in one file, so every parallel agent edits the same region. The success created the hazard.

All three are restored verbatim from the commits that added them, and verified to run and pass.

## 1. `gates-not-deregistered` — the structural fix

Compares the manifest with its own **merge-base**, not the base tip. That distinction matters: comparing against the tip would demand entries a branch cut earlier never had a chance to see, i.e. a false positive on every stale branch. The merge-base asks the only honest question — *did this branch drop something it started with?*

Flags two things:
- a gate id present at the merge-base and gone here;
- a gate downgraded from `enforced` to `advisory` — the quieter half of the same failure, and one no other check can see.

Removal stays possible. It just has to be **said out loud**:

```yaml
# GATE-REMOVED: <id> — <reason>
# GATE-DOWNGRADED: <id> — <reason>
```

so the intent sits in the diff a human reads. A declaration that is wrong in the *other* direction (declared removed, still present) is also a finding — otherwise the markers rot into the permanent exception list this repo keeps paying for.

**Verified by replaying the actual incident**, not by fixtures alone:

```
REPLAY of the real incident (#3492 over #3474/#3484/#3486): 3 finding(s)
  - gate 'gate-script-registration' (enforced) is in the manifest at the merge-base and gone here…
  - gate 'gitops-duplicate-resources' (enforced) …
  - gate 'scheduled-trigger-emitted' (enforced) …
```

### Why `gate-script-registration` was not enough

It catches the *symptom* — a `check-*` script nothing runs — and it did catch two of the three, which is how I found this at all. It cannot see:
- a lost entry whose script is still invoked from somewhere else,
- a gate whose `run:` is an inline command with no script,
- a silent `enforced` → `advisory` downgrade.

The two are complementary, and neither subsumes the other.

## 2. `opa-sidecar-bundle-shape` — #2872 defect 4

campaign-service died on `/bundle/rules-data.yaml: merge error`. The sidecar runs `opa run --bundle /bundle` — **directory mode**, where OPA derives a data document's path from its location in the tree. The bundle files were bit-identical to every healthy service's; only the **mounts** differed. Mounted flat instead of at `/bundle/<root>/data.yaml`, two documents claim the same root, the bundle fails to load, the container dies at boot.

Nothing could see it: the ConfigMap is right, the args are right, the bundle drift gate compares **content**, and the service build knows nothing about mounts. The contradiction lives strictly between a Deployment's `volumeMounts` and the bundle's own `manifest.json`.

Four rules, of which the last is the one worth the effort:

1. `--bundle` is passed at all (without it OPA serves an empty policy and every decision falls through, silently);
2. `<dir>/.manifest` is mounted;
3. no data file mounted **flat** — the exact #2865 shape, which crashes loudly;
4. every data file sits under a root declared in that bundle's `manifest.json` — **OPA ignores data outside a declared root rather than rejecting it**, so the policy evaluates against absent data and no log explains the decision.

Roots come from each bundle's own `manifest.json`, so there is no per-service list and no exception file. **40 sidecars, 0 findings, 0 exceptions.** Proven by flattening one real mount in `account-service.yaml` and watching it reproduce the campaign message, then restoring from a `cp` backup.

## Self-assessment

- **This PR is itself the hazard it describes.** It edits `gates.yaml` from a branch, so if another gate lands before it merges, the same replay could drop it. The new gate protects the *next* one, not this one — merge promptly, and if it conflicts, resolve by taking `origin/main`'s manifest and re-adding these entries rather than replaying this copy.
- **`gates-not-deregistered` cannot see a gate that was never merged** — a PR whose gate is dropped during conflict resolution before it ever reached `main` leaves no trace at any merge-base.
- **It does not check that a gate still does anything.** A gate whose `run:` was gutted to `true` keeps its id and its mode and passes this check. That is the reachability question `gate-script-registration` also declines to answer (#2327's shape).
- **`opa-sidecar-bundle-shape` reads only `openbank-infra/gitops/components`.** A sidecar defined elsewhere, or an image whose name does not contain `opa`, is invisible.
- **Rule 4 goes quiet when no bundle ConfigMap is found** rather than guessing at roots — deliberate, but it means a renamed ConfigMap downgrades rule 4 to nothing while rules 1-3 still run. Pinned as a self-test case so the restraint is intentional and visible.

## Verification

```
check-gates-not-deregistered.py --self-test    OK (9 transitions, both directions, stale declarations)
check-gates-not-deregistered.py --enforce      clean (86 at merge-base, 90 here)
check-opa-sidecar-bundle-shape.py --self-test  OK (6 fixtures incl. the campaign shape + empty tree)
check-opa-sidecar-bundle-shape.py --enforce    clean (40 sidecars)
run-gates.py --only gitops-duplicate-resources PASS   (restored)
run-gates.py --only scheduled-trigger-emitted  PASS   (restored)
run-gates.py --only gate-script-registration   PASS   (restored, 0 orphans)
run-gates.py --group lint                      PASS=5
```

### DIFF
--- a/.github/gates/gates.yaml
+++ b/.github/gates/gates.yaml
@@ -861,6 +861,96 @@ gates:
     run: |
       bash .github/scripts/ensure-ecr-repository.sh --self-test
 
+  # campaign-service's rollout died on `/bundle/rules-data.yaml: merge error` (#2872 defect 4).
+  # The sidecar runs `opa run --bundle /bundle` — DIRECTORY mode, where OPA derives a data
+  # document's path from its location in the tree. The bundle files were bit-identical to every
+  # healthy service's; only the MOUNTS differed. Mounted flat instead of at /bundle/<root>/data.yaml,
+  # two documents claim the same root, the bundle fails to load, and the container dies at boot.
+  #
+  # Nothing could see it: the ConfigMap is right, the args are right, the bundle drift gate compares
+  # CONTENT, and the service build knows nothing about mounts. The contradiction lives strictly
+  # between a Deployment's volumeMounts and the bundle's own manifest.json.
+  #
+  # The rule worth the effort is the quiet one: a data file under an UNDECLARED root is silently
+  # IGNORED by OPA rather than rejected, so the policy evaluates against absent data and no log
+  # explains the decision. Roots come from each bundle's own manifest.json — no per-service list.
+  #
+  # ENFORCED: 40 sidecars, 0 findings, 0 exceptions. Proven by flattening one real mount and
+  # watching it reproduce the campaign message.
+  # RESTORED. This entry was silently lost: #3486 added it and #3492 (branched from an older main,
+  # inserting at the SAME anchor in this file) squash-merged afterwards and took the file with its
+  # own version. GitHub reported both PRs MERGEABLE/CLEAN — a textbook semantic merge conflict, and
+  # the exact hazard this repo documents for parallel agents. The script survived on main
+  # registered nowhere, i.e. the meta-gate became the orphan it exists to detect. It is registered
+  # here again, and from now on its own rule makes that loss a build failure rather than silence.
+  #
+  # Every check-* script must be in this manifest, invoked from a tracked file, or a declared
+  # HELPER. #3240 found two that had never executed — one of them written to stop a regulatory
+  # drift, referenced from nowhere for its whole life.
+  - id: gitops-duplicate-resources
+    name: "No duplicate resource or env-list entry in a gitops manifest (issues #3450/#3460, enforced)"
+    group: gitops
+    mode: enforced
+    selftest: |
+      python3 .github/scripts/check-gitops-duplicate-resources.py --self-test
+    run: |
+      python3 .github/scripts/check-gitops-duplicate-resources.py --enforce
+
+  - id: scheduled-trigger-emitted
+    name: a declared SCHEDULED trigger must actually be emitted (an unreachable schedule fails silently)
+    group: kotlin
+    mode: enforced
+    run: |
+      # Falsified in three directions before wiring in: a new unemitted service fails, a
+      # baselined service that becomes covered ALSO fails (so the list cannot rot), and the
+      # tree as committed passes. 9 self-test cases, including comment-only mentions and
+      # Kotlin's NESTED block comments.
+      python3 .github/scripts/check-scheduled-trigger-emitted.py --self-test
+      python3 .github/scripts/check-scheduled-trigger-emitted.py --enforce
+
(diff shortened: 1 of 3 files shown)


JiRaska:
**Correction to this PR's root-cause analysis.** #3516 (opened independently, now closed as superseded) got it right where I got it wrong, and the difference changes the lesson.

I wrote that #3492 was "branched from an older `main`". It was not. Verified:

```
$ gh pr view 3492 --json baseRefOid --jq .baseRefOid
8712a3f586240f3e094bb88e840c6aaf04b5f62f          # this IS #3486's merge commit
$ git show 8712a3f58:.github/gates/gates.yaml | grep -c 'id: gate-script-registration'
1                                                  # the base DID contain it
```

The base was correct. What went wrong is that I validated the workflow change on a throwaway probe branch (cut from an earlier `main`, following the validate-against-GitHub discipline) and then restaged the files onto the real branch with:

```
git checkout tmp/ecr-probe-52568 -- .github/gates/gates.yaml
```

That copies a **whole file** from a stale branch onto a fresh base. This repo's guardrails already name that trap — *never re-create a PR by copying whole files from a stale checkout; the copy silently reverts everything `main` gained since that branch diverged* — and I walked into it while doing something that felt like extra rigour.

So the transferable rule is not "watch your base". The base was fine, and **that is precisely what made it invisible**: every check you would normally run to catch a stale branch says you are up to date. The rule is: **never move a whole file between branches — re-apply the edit**, or take the base's copy first and re-add your change on top. If you must copy, `git diff origin/main -- <file>` afterwards and read what *else* changed.

Nothing in the shipped code changes — the restorations and both guards are unaffected, and `gates-not-deregistered` catches this regardless of which mechanism produced it (it compares the manifest against its own merge-base, not against a branch's age). Only the "why" in the commit message and PR body above is wrong, and this comment supersedes it.

PR #3520, opened 2026-08-02
test(delegation): pact contract coverage — 6 pacts, 12 replayed interactions, both directions

### LINKED ISSUE #2991: Pact contract tests for delegation-service (sca-service, pid-service, provider)

JiRaska:
## What

`openbank-delegation-service` had **zero pacts**. Every deploy passed the contract gate by having
nothing to check — and it is the service that mints payment rights across party boundaries
(ADR-0232). This adds contract coverage in both directions and wires a provider replay for every
pact it introduces.

Closes #2991.

### Consumer side — the calls delegation-service makes before it will mint a grant

| Provider | Interaction | Pinned |
|---|---|---|
| `openbank-sca-service` | `GET /api/v1/sca/challenges/{id}` | `purpose="DELEGATION_GRANT"`, `status="COMPLETED"` as **values** |
| `openbank-sca-service` | `POST /api/v1/sca/challenges/{id}/consume` | body is `{partyId}` only — the unlinked-consume shape |
| `openbank-pid-service` | `GET /api/v1/parties/{id}` | `status="ACTIVE"` + the **nested** `kycAttributes.kycLevel="FULL"` |
| `openbank-account-service` | `GET /api/v1/accounts/{id}` | `partyId` — the whole ownership verdict |
| `openbank-card-issuance-service` | `GET /api/v1/cards/{id}` | `partyId` |

Values, not `stringType`, wherever the consumer branches on the exact string (#2425).
`DelegationService` gates on `purpose == "DELEGATION_GRANT"` precisely so a payment challenge — or
the grantee's `DELEGATION_ACCEPT` half — can never be spent to mint a grant; a type matcher would
have asked the replay nothing about that. Same reasoning for pid's `kycLevel`: `KYC_RANK`'s keys are
exactly pid's `KycLevel` names, and the client reads that nested field with an elvis to `"NONE"`, so
a moved field is not a crash — it is every offer silently refused.

### Message side — the `openbank.delegation.events` payload

`account-service` and `card-issuance` each pact the two events they project (`DelegationActivated`
creates an enforceable row, `DelegationRevoked` closes one), and `delegation-service` replays them
against the **real** `DelegationEvents` data classes.

The replay is the load-bearing half here. Both consumers parse with `readTree` + `node.path(...)`,
so every field is optional at the type level: rename `capabilities` and the projection records a
grant authorising nothing; rename `resourceType` and the filter drops the event and the projection
goes permanently empty. None of that is visible from the consumer side — the pact mock hands the
consumer whatever the pact says.

### Provider replays — all `@PactFolder`, all running on the PR

- **new** `PidPactFolderProviderVerificationTest` — pid-service had no pact tests at all
- **new** `CardIssuancePactFolderProviderVerificationTest` — nor did card-issuance; its
  `build.gradle.kts` said outright that nothing pacts against it, which ADR-0232 made untrue
- **new** `DelegationEventPactFolderProviderVerificationTest` — message target, plain JVM
- `ScaPactFolderProviderVerificationTest` + broker twin gain a `COMPLETED`/`DELEGATION_GRANT` state
- `AccountPactFolderProviderVerificationTest` + broker twin gain **per-interaction HTTP/message
  dispatch** and now boot Quarkus. The old KDoc predicted exactly this; splitting into a second
  `@PactFolder` class was not an option, since both would load every pact naming the provider.

`check-pact-provider-replay.py`: **40 committed pacts, 40 replayed on a PR, `KNOWN_UNCOVERED` still
empty.** `derive-pact-drift-scope.sh` picks up `:openbank-delegation-service` on its own — no orphan
pact, no ungated pact, no hand-edited scope.

## Falsification — these tests have been shown to fail

A test that has only ever passed is unfalsified. Two mutations, both run locally, verdicts read from
the JUnit XML (pact-jvm's console "Not all of the N were verified" is noise).

**1. Consumer — a mispointed client, the #2269 defect class.** Changed ONE line of *production*
code, `ScaChallengeClient.kt:33`, `@Path("/api/v1/sca/challenges")` → `@Path("/api/v1/sca/challenge")`.
The pact expectations were not touched. Both interactions went red:

```
DelegationScaChallengePactConsumerTest > the challenge lookup returns the party, purpose and status the ceremony gates on FAILED
DelegationScaChallengePactConsumerTest > spending the challenge is a POST to consume that states only the party FAILED
  java.lang.AssertionError: 1 expectation failed. Expected status code <200> but was <500>.
```

This is the asymmetry working: the outgoing request is reflected off the client's `@Path` via
`ClientRoute`, the expectation is a literal. Derive both and the test cannot fail — measured on
#2290, and `finrep-service`'s call to the never-existent `/api/v1/ledger/trial-balance` is what it
costs.

**2. Provider replay — a drifted event field.** Changed `produceAccountActivated`'s
`resourceType` from `ACCOUNT` to `CARD`:

```
TEST-...DelegationEventPactFolderProviderVerificationTest.xml tests 4 failures 1 errors 0
  openbank-account-service - a DelegationActivated event for an account:
    1.1) body: $.resourceType Expected 'CARD' (String) to be equal to 'ACCOUNT' (String)
```

Both mutations reverted; the tree is clean and the suite is back to `TOTAL 9 failures 0`
(delegation-service `*contract*`), with every provider replay green:

| Replay | interactions | result |
|---|---|---|
| `ScaPactFolderProviderVerificationTest` | 3 (2 new + the existing consent one) | 0 failures |
| `PidPactFolderProviderVerificationTest` | 1 | 0 failures |
| `AccountPactFolderProviderVerificationTest` | 3 (1 new HTTP + 2 existing messages) | 0 failures |
| `CardIssuancePactFolderProviderVerificationTest` | 1 | 0 failures |
| `DelegationEventPactFolderProviderVerificationTest` | 4 | 0 failures |

`detekt` + `ktlintCheck` green on all five touched modules.

## What this does NOT cover

Read this before concluding that delegation-service is contract-tested.

1. **Only the GRANT half of the SCA ceremony.** `DELEGATION_ACCEPT` — the grantee's challenge — has
   no pact. Purpose equality exists precisely so a grantor's challenge cannot be spent as the
   grantee's acceptance, and that property is asserted in one direction only.
2. **Every interaction is a 200. There are no negative paths.** A 404 from account-service or
   card-issuance is a *definitive* NOT_OWNED verdict, and sca's 409 is the replay protection the
   consume exists for. Neither is pacted. The branch that distinguishes "someone is trying
   something" from "the provider is down" is exactly the one with no contract.
3. **`PAYMENT` / `STATEMENT` / `DOCUMENT` resource types** return UNVERIFIABLE with no lookup — there
   is nothing to pact, and equally nothing that would notice if one of them quietly gained a wrong route.
4. **Nothing here is about authorization.** Every provider replay runs under `@TestSecurity` with the
   roles handed to it. A pact proves the route and the shape; it says nothing about whether
   delegation-service's real M2M token gets through (ADR-0034). If the OPA rule denies, all five
   calls 403 and every pact in this PR stays green.
5. **The message contract does not prove delivery.** It pins the JSON `DelegationEvents` serializes
   and the two consumers parse. It does not prove the outbox dispatches, the topic names line up,
   the consumer is wired to `delegation-events-in`, or that the Kafka `group.id` reaches the
   connector at all (the dotted-key trap). Green pact + broken channel config is precisely the shape
   of a silently-empty projection.
6. **`validTo` is not asserted.** The example carries `validTo = null`, so the grant-expiry field the
   projections read has no format pinned — only `validFrom` does.
7. **`capabilities` is a `stringType` array.** The replay checks a non-empty array of strings; it does
   NOT check the vocabulary against `DelegationCapability`. A renamed capability would be projected
   as an unknown string the guard never matches — fail-closed, silent, uncaught.
8. **delegation-service has no `@PactBroker` twin** (it is a message provider only, and is not in
   the broker's consumer graph yet), so `can-i-deploy` will see no published verification result
   *for delegation-service as a provider*. If account-service or card-issuance ever publish those
   message pacts to the broker, that is the #3239 shape and a twin is needed then. Documented in the
   class KDoc rather than pre-emptively added. **Note this is now the only remaining half of that
   gap** — see the correction below.
9. **The clearest weakness in this PR: four broker-sourced classes cannot run on the PR lane.**
   `AccountEventPactProviderVerificationTest` is now a `@QuarkusTest` with HTTP dispatch,
   `ScaPactProviderVerificationTest` gained the new state, and
   `PidPactBrokerProviderVerificationTest` / `CardIssuancePactBrokerProviderVerificationTest` are
   new — all four are `@EnabledIfSystemProperty(pactbroker.url)`-gated, so none has ever executed
   against a broker. What is verified is below; first main-push is where they are actually tried.
10. **Nobody pacts delegation-service's own REST API.** The customer edge calls it and is not a pact
    consumer of it, so `openbank-delegation-service/openapi.yaml` remains backed by no contract.
11. **SAVINGS_GOAL is covered by prose, not by an interaction.** It resolves through the same account
    lookup by convention (ADR-0153); nothing tests that the convention still holds.

---

## Correction after review — broker twins for pid-service and card-issuance (`b75e04ae5`)

The review caught something my own gap list understated. I flagged the missing broker twin on the
**delegation** side; the costly half is on the **provider** side, and it would have landed on
delegation-service's deploys, not on anything abstract.

`pid-service` and `card-issuance-service` each had exactly one `@Provider` class, both
`@PactFolder`-sourced. A `@PactFolder` class replays from disk and never contacts the broker, and
`can-i-deploy` reads the broker and nothing else. Merging the consumer pacts as they stood would
have meant delegation-service publishes pacts naming two providers that never publish a
verification — so `can-i-deploy` answers *"no verified pact between openbank-delegation-service and
the version of openbank-pid-service currently in sandbox"* and blocks every deploy, labelled
`PENDING_BUILD`: a name that reads transient for a state that is permanent. That is the #3239 shape
that took fx-service out across four attempts on 2026-08-02, and delegation-service had only just
reached sandbox.

Added the sanctioned pair for both — `@PactBroker` + `@EnabledIfSystemProperty(pactbroker.url)`,
matching aml/sanctions (#3239) and openbank-ledger-service. `sca-service` and `account-service`
already had broker-sourced classes and were left alone.

**States mirrored** (every one, per the `MissingStateChangeMethod` caution — a state the broker
replay cannot satisfy publishes a FAILURE, which is worse than publishing nothing):

| Provider | State mirrored into the twin |
|---|---|
| `openbank-pid-service` | `an ACTIVE party with FULL KYC exists` — full party seed + the `findById` guard (`persist`, not `merge`, on an application-assigned id) |
| `openbank-card-issuance-service` | `a card held by a known party exists` — full `CardEntity` seed + the same guard |

Parity is asserted mechanically, **against the annotation on the class declaration, not a
whole-file grep** — the review's own first probe reported "broker-sourced: yes" for both services
because `git grep '@PactBroker'` matched the KDoc paragraph explaining why the class was *not*
broker-sourced. The checker strips comments, reads the annotation block on the `class` line, and
asserts set equality of `@State` values in both directions. Validated against a known-positive:
renaming one state in the pid twin makes it print `FAIL STATE PARITY` and `FAIL no extra orphan
state`; restored, it exits 0.

**Both `build.gradle.kts` gained the `pactbroker.*` forwarding** to the forked test JVM —
`pid-service` had no `tasks.withType<Test>` block at all. Without it `pactbroker.url` never reaches
the test JVM, the twin stays `@EnabledIfSystemProperty`-skipped on *every* lane including main-push,
and a twin that cannot see the broker is indistinguishable from not having one.

**KDoc on both `@PactFolder` classes updated.** They previously argued why *they* are not
broker-sourced; unamended, that text reads as an argument against the class just added. Both now
say they are one half of a deliberate pair, and that any new `@State` must be added to the twin in
the same commit.

### What is verified about the new classes, and what is not

Verified:
- both modules compile (`compileTestKotlin` green);
- both classes are **discovered** by JUnit and skipped by the gate — `tests=1 skipped=1 failures=0`
  in the emitted JUnit XML. The `--tests` filter also reports `No tests found`, because it selected
  only a fully-skipped container; the **pre-existing, merged `AmlPactBrokerProviderVerificationTest`
  behaves identically under the same probe**, so that is established fleet behaviour, not something
  introduced here. The probe was falsified first: a nonsense filter fails the build;
- the `@PactFolder` replays still pass with a second `@Provider` class present in the module —
  pid 1/1 and card-issuance 1/1, no collision;
- `detekt` + `ktlintCheck` green on both modules;
- `check-pact-provider-replay.py`: **40 pacts, 40 replayed, `KNOWN_UNCOVERED` empty**, with both
  modules now showing the sanctioned `SKIP` (broker) + `RUNS` (folder) pair;
- no pact file changed, so the drift gate is unaffected.

Not verified, stated plainly: **neither twin has ever executed against a broker.** They cannot on
the PR lane — `_service-ci.yml` blanks `PACT_BROKER_URL` (ADR-0056). Their `@State` bodies are
copies of code that *is* exercised by the `@PactFolder` twins, which is the strongest evidence
available pre-merge, but the broker fetch, the state dispatch across multiple consumers' pacts, and
the result publish are all first exercised on main-push. That run is the one to watch.

One incidental finding worth recording: running two Quarkus provider replays in a single Gradle
invocation fails with `Port already bound: 8081` — both bind the default `quarkus.http.test-port`.
That is a probe artifact, not a defect; each passes when run alone, and CI runs services in separate
jobs. I nearly mis-filed it as a regression from this change.

---

## Correction 2 — `validFrom` pact drift (`a481dfa86`)

`pact-drift-check` was red on every run: the committed message pacts carried
`"2000-01-31T14:00:00+01:00"`, CI regenerated `"2000-01-31T13:00:00Z"`. Same instant, different
string, because the artifact encoded the generating machine's timezone.

The cause is narrower than the rendering, and worth stating precisely because the obvious fix is the
wrong one. **The timestamp matcher was already present and already correct** — `$.validFrom` has
always been `{"match": "timestamp", "format": ...}`, so the contract has never asserted the literal
instant. What wrote the machine-dependent value was the *generator* that `datetime(name, format)`
registers when given no example:

```json
"generators": { "body": { "$.validFrom": { "type": "DateTime", "format": "yyyy-MM-dd'T'HH:mm:ssXXX" } } }
```

Fixed with `datetime(name, format, Date, TimeZone)` — the only overload that pins the rendering; the
`Instant` overload still formats with `TimeZone.getDefault()`. Matcher retained, generator gone,
example now the instant the provider replay actually emits (`2026-01-01T00:00:00Z`) instead of a
pact-jvm default from the year 2000. This is not a return to a pinned value: the matcher still
governs verification.

The KDoc now makes the distinction explicit, so the next reader does not "simplify" it in either
direction: `validFrom` is a matcher **because** no consumer branches on the instant;
`eventType`/`resourceType` are `stringValue` **because** `dispatch` compares them to exact literals.

**Verified:** all six pacts regenerated under `TZ=UTC`, `Europe/Prague`, `Asia/Kolkata` (+05:30) and
`Pacific/Chatham` (+12:45) — byte-identical by sha256 every time. The sweep was falsified first:
restoring the 2-arg call makes it flag the card pact and reproduces the exact CI value. CI's own
assertion — regenerate under UTC, `git diff --exit-code -- pacts/` — exits 0.

**No other environment-dependent values.** Checked statically (every string in all six pacts scanned
for TZ offsets, clock-derived dates, epochs, paths, hostnames — clean) and empirically via the
four-timezone sweep, which covers whatever a static scan would have missed. The only other
generators are `Uuid`, whose examples are pact-jvm constants.

Generalisable: **a matcher does not make a pact environment-independent — a generator can still
write a machine-dependent value next to it.** The matcher is what verification reads; the generator
is what the committed artifact records.

### DIFF
--- a/openbank-account-service/src/test/kotlin/com/openbank/account/contract/AccountEventPactProviderVerificationTest.kt
+++ b/openbank-account-service/src/test/kotlin/com/openbank/account/contract/AccountEventPactProviderVerificationTest.kt
@@ -5,6 +5,7 @@
 package com.openbank.account.contract
 
 import au.com.dius.pact.provider.PactVerifyProvider
+import au.com.dius.pact.provider.junit5.HttpTestTarget
 import au.com.dius.pact.provider.junit5.MessageTestTarget
 import au.com.dius.pact.provider.junit5.PactVerificationContext
 import au.com.dius.pact.provider.junit5.PactVerificationInvocationContextProvider
@@ -14,13 +15,34 @@ import au.com.dius.pact.provider.junitsupport.State
 import au.com.dius.pact.provider.junitsupport.loader.PactBroker
 import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule
 import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
+import com.openbank.account.application.port.out.AccountRepository
 import com.openbank.account.domain.event.AccountCreatedEvent
+import com.openbank.account.domain.model.Account
+import com.openbank.account.domain.model.AccountStatus
 import com.openbank.account.domain.model.AccountType
+import com.openbank.account.it.PostgresRedpandaRedisTestResource
+import com.openbank.libs.domain.account.Iban
+import com.openbank.libs.domain.money.CurrencyCode
+import io.quarkus.test.common.QuarkusTestResource
+import io.quarkus.test.junit.QuarkusTest
+import io.quarkus.test.security.TestSecurity
+import io.quarkus.vertx.core.runtime.context.VertxContextSafetyToggle
+import io.vertx.core.Vertx
+import io.vertx.core.impl.ContextInternal
+import jakarta.inject.Inject
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.asCoroutineDispatcher
+import kotlinx.coroutines.launch
+import org.eclipse.microprofile.config.inject.ConfigProperty
 import org.junit.jupiter.api.BeforeEach
 import org.junit.jupiter.api.TestTemplate
 import org.junit.jupiter.api.condition.EnabledIfSystemProperty
 import org.junit.jupiter.api.extension.ExtendWith
+import java.time.Instant
 import java.util.UUID
+import java.util.concurrent.CompletableFuture
+import java.util.concurrent.Executor
+import java.util.concurrent.TimeUnit
 
 /**
  * Provider-side verification for async message contracts published by account-service
@@ -45,11 +67,13 @@ import java.util.UUID
  * `@EnabledIfSystemProperty` keeps it a no-op locally and on the PR lane, where no broker is
  * configured — matching every other broker-based provider test in the fleet.
  *
- * A [MessageTestTarget] alone is correct here: both pacts naming account-service as provider are
- * message-only (balance-service's AccountCreated, notification-service's TRANSACTION_COMPLETED),
- * so unlike `PartyEventPactProviderVerificationTest` there is no HTTP interaction to dispatch to
- * and no Quarkus instance to boot. If an HTTP consumer contract against account-service is ever
- * added, this needs party-service's per-interaction target dispatch.
+ * Per-interaction target dispatch, as `PartyEventPactProviderVerificationTest` does: two of the
+ * three pacts naming account-service as provider are message-only (balance-service's
+ * AccountCreated, notification-service's TRANSACTION_COMPLETED), but delegation-service's
+ * ownership gate (issue #2991, ADR-0232 D7) reads `GET /api/v1/accounts/{id}`, which needs a
+ * running endpoint — hence `@QuarkusTest` + Testcontainers. This mirrors the git-pact twin
+ * exactly; the two must stay identical or the same contract passes from git and fails from the
+ * broker.
  *
  * IMPORTANT: if `AccountCreatedMessagePactConsumerTest` (openbank-balance-service) changes the
  * contract, regenerate the pact JSON (`./gradlew :openbank-balance-service:test --tests
@@ -59,19 +83,60 @@ import java.util.UUID
  *
  * `@IgnoreNoPactsToVerify(ignoreIoErrors)` makes a missing/unreadable pact a skip, not a failure.
  */
+@QuarkusTest
+@QuarkusTestResource(PostgresRedpandaRedisTestResource::class)
+@TestSecurity(user = "pact-verifier", roles = ["ROLE_API", "ROLE_VIEWER", "ROLE_OPERATOR"])
 @Provider("openbank-account-service")
 @PactBroker(enablePendingPacts = "true")
 @IgnoreNoPactsToVerify(ignoreIoErrors = "true")
 @EnabledIfSystemProperty(named = "pactbroker.url", matches = ".+")
 class AccountEventPactProviderVerificationTest {
 
+    companion object {
+        // Must match DelegationAccountOwnershipPactConsumerTest's ACCOUNT_ID / OWNER_PARTY_ID.
+        private val ACCOUNT_ID = UUID.fromString("11111111-2222-4333-8444-555555555555")
+        private val OWNER_PARTY_ID = UUID.fromString("66666666-7777-4888-8999-aaaaaaaaaaaa")
+    }
+
     private val objectMapper = jacksonObjectMapper().registerModule(JavaTimeModule())
+
+    @ConfigProperty(name = "quarkus.http.test-port", defaultValue = "8081")
+    lateinit var testPort: String
+
+    @Inject
+    lateinit var accountRepository: AccountRepository
+
+    @Inject
+    lateinit var vertx: Vertx
+
+    /** See the git-pact twin: Pact-JVM calls `@State` on a thread with no Vert.x context. */
+    private fun runOnVertxContext(block: suspend () -> Unit) {
+        val future = CompletableFuture<Unit>()
+        val duplicated = (vertx.orCreateContext as ContextInternal).duplicate()
+        VertxContextSafetyToggle.setContextSafe(duplicated, true)
+        val dispatcher = Executor { command -> duplicated.runOnContext { command.run() } }.asCoroutineDispatcher()
+        CoroutineScope(dispatcher).launch {
+            try {
+                block()
+                future.complete(Unit)
+            } catch (t: Throwable) {
+                future.completeExceptionally(t)
+            }
+        }
+        future.get(10, TimeUnit.SECONDS)
+    }
 
     @BeforeEach
     fun setTarget(context: PactVerificationContext?) {
+        if (context == null) return
         // Limit the @PactVerifyProvider scan to this package — the default classpath-wide scan
-        // (ClassGraph) throws on the JDK 25 toolchain.
-        context?.target = MessageTestTarget(listOf("com.openbank.account.contract"))
+        // (ClassGraph) throws on the JDK 25+ toolchain.
+        context.target = if (context.interaction.isAsynchronousMessage()) {
+            MessageTestTarget(listOf("com.openbank.account.contract"))
+        } else {
+            HttpTestTarget("localhost", testPort.toInt())
+        }
+        context.addStateChangeHandlers(this)
     }
 
     @TestTemplate
@@ -116,5 +181,26 @@ class AccountEventPactProviderVerificationTest {
             "variables" to mapOf("amount" to "50.00", "currency" to "CZK"),
         )
         return objectMapper.writeValueAsString(request)
+    }
+
+    /** Mirror of the git-pact twin's handler — see there for why the findById guard is needed. */
+    @State("an account owned by a known party exists")
+    fun accountOwnedByKnownParty() = runOnVertxContext {
+        if (accountRepository.findById(ACCOUNT_ID) != null) return@runOnVertxContext
+        accountRepository.save(
+            Account(
+                id = ACCOUNT_ID,
+                accountNumber = Iban("CZ6508000000192000145399"),
+                accountType = AccountType.CURRENT,
+                partyId = OWNER_PARTY_ID,
+                productId = UUID.fromString("99999999-8888-4777-8666-555555555555"),
+                currency = CurrencyCode("CZK"),
+                status = AccountStatus.ACTIVE,
+                openedAt = Instant.parse("2026-01-01T00:00:00Z"),
+                closedAt = null,
+                version = 0,
+            ),
+        )
+        Unit
     }
 }
--- a/openbank-account-service/src/test/kotlin/com/openbank/account/contract/AccountPactFolderProviderVerificationTest.kt
+++ b/openbank-account-service/src/test/kotlin/com/openbank/account/contract/AccountPactFolderProviderVerificationTest.kt
@@ -5,6 +5,7 @@
 package com.openbank.account.contract
 
 import au.com.dius.pact.provider.PactVerifyProvider
+import au.com.dius.pact.provider.junit5.HttpTestTarget
 import au.com.dius.pact.provider.junit5.MessageTestTarget
 import au.com.dius.pact.provider.junit5.PactVerificationContext
 import au.com.dius.pact.provider.junit5.PactVerificationInvocationContextProvider
@@ -14,12 +15,33 @@ import au.com.dius.pact.provider.junitsupport.State
 import au.com.dius.pact.provider.junitsupport.loader.PactFolder
 import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule
 import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
+import com.openbank.account.application.port.out.AccountRepository
 import com.openbank.account.domain.event.AccountCreatedEvent
+import com.openbank.account.domain.model.Account
+import com.openbank.account.domain.model.AccountStatus
 import com.openbank.account.domain.model.AccountType
+import com.openbank.account.it.PostgresRedpandaRedisTestResource
+import com.openbank.libs.domain.account.Iban
+import com.openbank.libs.domain.money.CurrencyCode
+import io.quarkus.test.common.QuarkusTestResource
+import io.quarkus.test.junit.QuarkusTest
+import io.quarkus.test.security.TestSecurity
+import io.quarkus.vertx.core.runtime.context.VertxContextSafetyToggle
+import io.vertx.core.Vertx
+import io.vertx.core.impl.ContextInternal
+import jakarta.inject.Inject
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.asCoroutineDispatcher
+import kotlinx.coroutines.launch
+import org.eclipse.microprofile.config.inject.ConfigProperty
 import org.junit.jupiter.api.BeforeEach
 import org.junit.jupiter.api.TestTemplate
 import org.junit.jupiter.api.extension.ExtendWith
+import java.time.Instant
 import java.util.UUID
+import java.util.concurrent.CompletableFuture
+import java.util.concurrent.Executor
+import java.util.concurrent.TimeUnit
 
 /**
  * Git-pact provider verification for account-service — the half that actually runs before a merge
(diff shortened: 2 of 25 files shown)


JiRaska:
## Blocking finding before merge: this makes delegation-service undeployable

Verified on `test/delegation-pact-coverage`: `pid-service` and `card-issuance-service` each have
exactly one `@Provider` class and both are `@PactFolder`-sourced. Neither can publish a
verification result to the broker.

After this merges, delegation-service publishes consumer pacts naming both as providers. Those
providers verify on the PR lane (good — that is what the `@PactFolder` classes are for) but never
publish, so `can-i-deploy` for **openbank-delegation-service** will report "no verified pact
between openbank-delegation-service and the version of openbank-pid-service currently in sandbox"
and block every deploy of it.

That is the #3239 failure shape exactly. It made `openbank-fx-service` undeployable across four
attempts earlier today, diagnosed only after `aml`/`sanctions` were found to be `@PactFolder`-only.
delegation-service was deployed to sandbox for the first time tonight; merging this as-is takes
that away, and the symptom will read as `PENDING_BUILD`, which is a misleading label for a
permanent state.

**Not an argument against the PR** — the coverage is the right work and the falsification evidence
is exactly what this repo asks for. It needs the second half of the sanctioned pair before it is
safe to merge.

## What is needed

Add the `@PactBroker` + `@EnabledIfSystemProperty(pactbroker.url)` twin for `pid-service` and
`card-issuance-service`, as #3239 did for `aml` and `sanctions`, and as
`openbank-ledger-service` already carries. Two cautions from that PR:

- **Mirror EVERY `@State`.** A broker replay missing one fails with `MissingStateChangeMethod` and
  publishes a FAILURE, which blocks an otherwise healthy pair — worse than publishing nothing.
- The KDoc on the existing `@PactFolder` classes explains why *they* are not broker-sourced. Update
  it so the next reader sees the pair is deliberate, rather than reading the existing text as an
  argument against the class you just added.

`sca-service` and `account-service` already have broker-sourced classes, so they are fine.

## One process note

My first probe for this reported "broker-sourced: yes" for both services — because
`git grep '@PactBroker'` matched the KDoc paragraph explaining why the class is *not*
broker-sourced. Whole-file greps cannot tell the thing from the prose about the thing; the
annotation on the class declaration is the construct to assert.

JiRaska:
Fixed in `b75e04ae5`. You were right, and my own gap list understated it: I flagged the missing twin on the *delegation* side, which is the cheap half. The provider side is the one that costs the deploy.

Confirmed the finding independently before acting, against the annotation on the class declaration rather than a whole-file grep — both `pid-service` and `card-issuance-service` had exactly one `@Provider` class and both were `@PactFolder`-sourced.

**States mirrored — every one:**

| Provider | State |
|---|---|
| `openbank-pid-service` | `an ACTIVE party with FULL KYC exists` — full `Party` seed, `runOnVertxContext` bridge, and the `findById` guard (`PartyRepositoryImpl.save` is `persist`, not `merge`, on an application-assigned id) |
| `openbank-card-issuance-service` | `a card held by a known party exists` — full `CardEntity` seed via `VertxContextSupport`, same guard |

That guard matters more on the broker side than on the folder side, and I noted why in the KDoc: the broker can hand one class several consumers' interactions naming the same state, so re-entry is expected rather than hypothetical.

Parity is asserted mechanically now, in both directions (no missing state, no orphan state), by stripping comments and reading the annotation block on the `class` line. I validated that checker against a known-positive first — renaming one state in the pid twin makes it print `FAIL STATE PARITY` / `FAIL no extra orphan state`; restored, it exits 0.

**One thing you did not ask for but the fix needed:** both `build.gradle.kts` were missing the `pactbroker.*` forwarding to the forked test JVM — `pid-service` had no `tasks.withType<Test>` block at all. Without it `pactbroker.url` never reaches the test JVM, the twin stays `@EnabledIfSystemProperty`-skipped on *every* lane including main-push, and a twin that cannot see the broker is indistinguishable from not having one. Added to both.

KDoc on both `@PactFolder` classes updated as you asked, so the pair reads as deliberate.

**What I can and cannot claim about the new classes.** Verified: both modules compile; both classes are *discovered* and skipped by the gate (`tests=1 skipped=1 failures=0` in the JUnit XML); the `@PactFolder` replays still pass with a second `@Provider` class in the module (pid 1/1, card 1/1 — no collision); detekt + ktlint green; `check-pact-provider-replay.py` reports **40 pacts, 40 replayed, `KNOWN_UNCOVERED` empty**, with both modules now showing the `SKIP` (broker) + `RUNS` (folder) pair; no pact file changed.

Not verified, plainly: **neither twin has ever run against a broker**, and cannot on the PR lane. Their `@State` bodies are copies of code the `@PactFolder` twins *do* exercise, which is the best evidence available pre-merge, but the broker fetch, the state dispatch across multiple pacts, and the result publish are all first tried on main-push. Worth watching that run.

Two process notes back:

- Your grep caution applies to the discovery check too. `--tests '*PactBrokerProviderVerificationTest*'` reports `No tests found` **and still emits JUnit XML** — the filter selected a fully-skipped container. I nearly read that as "not discovered". The pre-existing `AmlPactBrokerProviderVerificationTest` from #3239 behaves identically under the same probe, which is what settled it; I falsified the probe first (a nonsense filter fails the build).
- Running two Quarkus provider replays in one Gradle invocation fails with `Port already bound: 8081` — both bind the default `quarkus.http.test-port`. Probe artifact, not a defect: each is green alone, and CI runs services in separate jobs. I almost filed it as a regression from this change.

JiRaska:
Fixed in `a481dfa86`. Your diagnosis of the *cause* is exactly right — CET-generated artifact, UTC CI — but the choice you offered rests on a premise that turned out to be false, so let me correct it before saying which option I took.

**The datetime matcher was already there.** `$.validFrom` has carried `{"match": "timestamp", "format": "yyyy-MM-dd'T'HH:mm:ssXXX"}` since the first commit. The contract has always asserted shape, never the literal instant. So option 2 as written — "use a Pact datetime matcher so the contract asserts the SHAPE" — would have changed nothing about the drift, and the field is already the way you and I both want it.

**The actual culprit was a generator, not the matcher and not the rendering.** `datetime(name, format)` with no example registers a *generator* alongside the matcher:

```json
"generators": { "body": { "$.validFrom": { "type": "DateTime", "format": "yyyy-MM-dd'T'HH:mm:ssXXX" } } }
```

A generator means the written value is *produced at generation time*, from pact-jvm's default base date rendered in `TimeZone.getDefault()`. That is why the committed string encoded the machine's timezone. Removing the generator is the fix; the matcher had to stay.

**So: the mechanic of option 1, with the semantics of option 2 already in place.** I used `datetime(name, format, Date, TimeZone)` with an explicit UTC instant — the only overload that pins the rendering, since the `Instant` overload still formats with `TimeZone.getDefault()`. Result: matcher retained, generator gone, example now the instant the provider replay actually emits (`2026-01-01T00:00:00Z`) rather than a library default from the year 2000. This is *not* a return to an exact-value assertion — the matcher still governs verification; the literal is only an example.

Your caution about over-matching is recorded where it will be read: the KDoc now states outright that `validFrom` is a matcher *because* no consumer branches on the instant, contrasts it with `eventType`/`resourceType` which are `stringValue` *because* `dispatch` compares them to exact literals, and says in as many words not to simplify those into matchers or widen this one into a pinned value.

**Verification — the sweep, falsified first.** Regenerated all six pacts under `TZ=UTC`, `Europe/Prague` (the original culprit), `Asia/Kolkata` (+05:30) and `Pacific/Chatham` (+12:45, the nastiest offset available): **byte-identical every time**, by sha256. Then I checked the sweep can actually fail — restoring the 2-arg call makes it report the card pact as differing and reproduces the exact CI value `2000-01-31T14:00:00+01:00` with the `DateTime` generator back in place.

CI's own assertion, run the way CI runs it: regenerate under `TZ=UTC`, then `git diff --exit-code -- pacts/` → **exit 0**.

**Other environment-dependent values: none.** I checked two ways rather than eyeballing. Statically, a scan of every string in all six pacts for TZ offsets, clock-derived dates, epochs, absolute paths and hostnames returns clean. Empirically — the stronger evidence, since it catches what I didn't think to look for — the four-timezone sweep above covers all six files, so any locale- or clock-derived value anywhere in them would have shown up. The only other generators are the `Uuid` ones, whose examples are pact-jvm constants (`e2490de5-…`), stable across every regeneration including CI's.

Also re-ran after the change: the delegation message replay is still 4/4 green against the new example, `check-pact-provider-replay.py` still 40/40 with `KNOWN_UNCOVERED` empty, `derive-pact-drift-scope.sh` clean, detekt + ktlint green on both modules.

One thing worth generalising beyond this PR: **a matcher does not make a pact environment-independent — a generator can still write a machine-dependent value next to it.** The matcher is what verification reads; the generator is what the committed artifact records. This one was invisible to every local run precisely because it only differs across timezones, and it would have been red for every future contributor to these modules, not just for me.

Issue #2101, opened 2026-07-24
app-status dossier auto-publish is a silent no-op: PLATFORM_RW_TOKEN not configured in openbank-app


JiRaska:
ADR-0074 invariant 3 transports the customer-app dossier by having **openbank-app**'s `app-status` workflow regenerate `openbank-admin-ui/app-status.json` and open a publish PR against this repo. That auto-publish is currently a **silent no-op**.

## Symptom
On every push to `JiRaska/openbank-app@main`, the `app-status (dossier)` workflow's publish job runs only its **"Skip notice (no platform write token)"** step and exits `success` — no PR is opened. The job is green, so nothing signals the failure.

Observed on run [30118523556](https://github.com/JiRaska/openbank-app/actions/runs/30118523556) (after openbank-app#276): job "Publish to platform repo" = success, but the only step executed was the skip notice.

## Root cause
`.github/workflows/app-status.yml` gates the checkout + regenerate + `create-pull-request` steps on `env.PLATFORM_RW_TOKEN != \'`. The secret `PLATFORM_RW_TOKEN` (needs `contents:write` + `pull-requests:write` on this repo) is **not configured** in openbank-app, so all real work is skipped.

This mirrors the platform-side `OPENBANK_APP_RO_TOKEN` skip (which IS configured) — but here the missing token means dossier changes in openbank-app never propagate on their own.

## Impact
Any curatorial/derived change in openbank-app (`AppConfig.kt`, `version.txt`, `app-status.yaml`) silently fails to reach the committed `app-status.json`. It surfaces only later as a red **Customer-app dossier** check on an unrelated platform PR (exactly how this was found — #2021), then has to be published manually (#2047, #2097).

## Fix
Add a `PLATFORM_RW_TOKEN` secret to `JiRaska/openbank-app` (fine-grained PAT or App installation token scoped `contents:write` + `pull-requests:write` on `JiRaska/open-bank-oss`). Then a manual `workflow_dispatch` / next main push opens the publish PR automatically.

## Hardening (optional follow-up)
A green job that did nothing is the vacuous-green failure class this codebase works to avoid. Consider making the missing-token path emit a `::warning` (or fail if the repo is expected to always publish) rather than a silent `::notice`, so a broken transport is visible.

Refs ADR-0074 (dossier invariant 3). Surfaced via #2021; manual republishes #2047, #2097.

JiRaska:
[...]

PR #3548, opened 2026-08-03
feat(lending): turn on the fail-closed origination guard now the CZ pack is activated

JiRaska:
## Precondition finally met

ADR-0212 D2's fail-closed origination guard has been off since it shipped, because its precondition could not be satisfied by anyone:

| blocker | PR |
|---|---|
| OPA never permitted pack activation | #3402 |
| persistence failed on both legs | #3448 |
| a single `ROLE_COMPLIANCE` account, so four-eyes could not complete | #3461 |

All three are fixed, and the CZ consumer-credit pack v1 is now **active in sandbox**:

```
EXECUTED | CZ | v1 | proposed by compliance@openbank.local
                   | decided  by compliance2@openbank.local
```

`GET /api/v1/lending/compliance-packs/active` lists it; the row is confirmed by a direct read of `compliance_pack_activation`, not only by the response that wrote it.

## What this refuses from now on

Any origination that does not name a `(jurisdiction, productType)` pair with an active pack.

**Customer intake is unaffected.** It supplies `CZ`/`CONSUMER_CREDIT` from configuration rather than from the request — deliberately, so an applicant cannot choose which compliance pack judges them — and that is the pair which is active.

**Backoffice callers on `POST /applications` do pass their own pair**, and anything other than `CZ`/`CONSUMER_CREDIT` will now be refused. That is the guard working, not a regression. Worth knowing before someone reads the first 400 as an outage.

## Why this is a separate PR from #3534

#3534 establishes that customer intake was never actually running — `@ConfigProperty` was ignored, so `LENDING_INTAKE_ENABLED=true` did nothing. Enforcing a compliance gate above a door nobody had opened would prove nothing and would tangle two independent failures in one deploy. This one lands after it.

Refs ADR-0212 D2/D4, #3402, #3448, #3461

### DIFF
--- a/openbank-infra/gitops/components/lending/lending-service.yaml
+++ b/openbank-infra/gitops/components/lending/lending-service.yaml
@@ -125,12 +125,21 @@ spec:
               value: "true"
             - name: OPENBANK_TEMPORAL_SERVER_URL
               value: temporal-frontend.temporal.svc.cluster.local:7233
-            # ADR-0212 D2/D4: fail-closed origination guard. STAYS false until the CZ
-            # reference pack is four-eyes-activated in this environment (compliance-
-            # packs/README.md runbook: seed -> activate -> verify /active -> flip).
-            # Flipping early refuses every origination with no pack present.
+            # ADR-0212 D2/D4: fail-closed origination guard. The runbook's precondition is
+            # met — the CZ consumer-credit pack v1 was four-eyes-activated in sandbox on
+            # 2026-08-02 (proposed by compliance@openbank.local, decided by
+            # compliance2@openbank.local, state EXECUTED) and /compliance-packs/active
+            # lists it.
+            #
+            # What this now refuses: any origination that does not name a (jurisdiction,
+            # productType) pair with an active pack. Customer intake is unaffected — it
+            # supplies CZ/CONSUMER_CREDIT from configuration rather than from the request,
+            # precisely so an applicant cannot pick which pack judges them, and CZ/
+            # CONSUMER_CREDIT is the pair that is active. Backoffice callers on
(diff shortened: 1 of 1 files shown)


JiRaska:
## Precondition verified in the live database, not from the PR body

This flips `LENDING_ENFORCE_PACK` to `true`, so origination starts refusing anything without an active pack. The PR states the runbook precondition is met; that is a claim about live state, which no commit can establish. Queried the sandbox directly:

```sql
select state, jurisdiction, product_type, pack_version, proposed_by, decided_by
from compliance_pack_activation;

EXECUTED | CZ | CONSUMER_CREDIT | 1 | compliance@openbank.local | compliance2@openbank.local
```

All four parts hold: the activation is `EXECUTED`, it is the `CZ`/`CONSUMER_CREDIT` pair, and `proposed_by` and `decided_by` are **different principals** — so the four-eyes ceremony genuinely completed rather than being self-approved.

Worth noting the chain that made that possible: `compliance2@openbank.local` is the second `ROLE_COMPLIANCE` principal added in #3461, which I verified imports cleanly into Keycloak 26.6.3 earlier tonight (`requiredActions: []`, both realm roles present). Before it existed the ceremony had no second human and could not complete — which is why this flag stayed `false`.

## What the flip actually refuses

Customer intake is unaffected: it supplies `CZ`/`CONSUMER_CREDIT` from configuration rather than from the request, and that is the active pair. Backoffice callers on `POST /applications` pass their own `(jurisdiction, productType)` and will be refused for anything else — which is the guard doing its job.

`Customer-app dossier` is red, as it is on every PR right now: `openbank-app` released again and the committed `app-status.json` trails it (#3552, root cause #2101). Non-required and unrelated.

*Verified while draining the queue overnight.*

PR #3566, opened 2026-08-03
ci(services): name a Docker registry pull failure instead of blaming the tests

### LINKED ISSUE #3562: Testcontainers registry-pull failures on the hosted PR lane read as test failures, not infra failures
## What happens

A Testcontainers image pull that fails against Docker Hub is reported to the reader as **named test
failures**, not as an infrastructure failure. On PR #3515, `build (openbank-ledger-service)` showed:

```
LedgerPactProviderVerificationTest > ... FAILED
LedgerSchedulerVertxContextIT > the scheduled FX revaluation reaches its use case() FAILED
```

Those two tests are fine. The cause is ~900 lines further down the same log:

```
com.github.dockerjava.api.exception.InternalServerErrorException: Status 500:
  {"message":"Get \"https://registry-1.docker.io/v2/\": net/http: request canceled ...
Caused by: org.testcontainers.containers.ContainerFetchException
Caused by: org.testcontainers.shaded.org.awaitility.core.ConditionTimeoutException
Retrying pull for image: postgres:16.3-alpine (104s remaining)
```

The registry error appears **32 times** in that log. A re-run passed, confirming it was transient.

## Why it is structural, not incidental

`_service-ci.yml` sets:

```yaml
runs-on: ${{ (github.event_name == 'push' || github.event_name == 'workflow_dispatch')
             && github.ref == 'refs/heads/main' && 'openbank-build' || 'ubuntu-latest' }}
```

so **every PR build runs on `ubuntu-latest`**, while the "Pre-warm Testcontainers image cache via
ECR" step is gated on `runner.environment == 'self-hosted'`. On the PR lane it is skipped entirely:
no ECR pull-through cache, no in-cluster registry-cache, and Testcontainers pulls
postgres/valkey/redpanda/apicurio straight from Docker Hub over a shared hosted-runner egress IP.

Verified: `gh api repos/JiRaska/open-bank-oss/actions/jobs/91553216566 --jq .labels` -> `ubuntu-latest`.

## Why it costs more than the wall clock

The failure is **misdiagnosed by construction**. The signal is not missing — a more legible and
completely wrong signal is printed above it. A reader who trusts the summary goes looking for a bug
in a test that does not have one, and the correct action (re-run) is the one they will not take.

Note also that a re-run **erases the evidence**: a run's top-level `conclusion` reports the latest
attempt only, so the run that produced the log above now reads `success`. Any frequency measurement
of this class has to enumerate `/actions/runs/<id>/attempts/<n>/jobs` per attempt, not trust the run
conclusion.

## Proposed fix

Emit a `::error` naming Docker Hub when a build fails with a registry-fetch signature, so the
diagnosis is at the top of the log instead of buried. Reducing the exposure itself (extending the
ECR pre-warm to hosted runners, scoping the pre-warm to the images a module needs, or a bounded
retry) should be decided on measured frequency and data-transfer cost, not assumed.

JiRaska:
## Summary

A Testcontainers image pull that fails against Docker Hub is reported to the reader as **named test
failures**, with the real cause ~900 lines below. This adds a diagnostic step that names Docker Hub
in a `::error` when a build fails with a registry-fetch signature.

It **deliberately does not** extend the ECR pre-warm to hosted runners. The measurement below is why.

## The defect

On PR #3515, `build (openbank-ledger-service)` (job `91553216566`) showed:

```
LedgerPactProviderVerificationTest > ... FAILED
LedgerSchedulerVertxContextIT > the scheduled FX revaluation reaches its use case() FAILED
```

Neither test is broken. 900 lines down:

```
com.github.dockerjava.api.exception.InternalServerErrorException: Status 500:
  {"message":"Get \"https://registry-1.docker.io/v2/\": net/http: request canceled ...
Caused by: org.testcontainers.containers.ContainerFetchException
```

The registry error appears **32 times** in that log. The failure is misdiagnosed **by construction**:
the signal is not missing, a more legible and completely wrong one is printed above it.

Structural cause — `runs-on` puts every PR build on `ubuntu-latest`, while the "Pre-warm
Testcontainers image cache via ECR" step is gated on `runner.environment == 'self-hosted'`. On the PR
lane it is skipped entirely, so Testcontainers pulls straight from Docker Hub.

## Measurement

Sample: the last **1000 `services-ci.yml` runs**, 2026-07-31T23:40Z -> 2026-08-03T03:48Z (2.17 days).

A run's top-level `conclusion` reports the **latest attempt only**, so a re-run erases this failure
class entirely — the run behind the log above now reads `success`. The sample therefore enumerates
`/actions/runs/<id>/attempts/<n>/jobs` for every attempt, not the run conclusion.

| | count |
|---|---|
| `build (...)` job attempts | **6323** |
| succeeded | 5685 |
| **failed** | **125** |
| cancelled | 289 |
| still running at sample time | 224 |

Runner split — this is the exposed population:

| runner | build-job attempts | share |
|---|---|---|
| `openbank-build` (self-hosted, main lane; ECR pre-warm **runs**) | 4567 | 72% |
| `ubuntu-latest` (hosted PR lane; ECR pre-warm **skipped**) | **1756** | **28%** |

Grep-safety: none of the signatures used to classify (`ContainerFetchException`,
`registry-1.docker.io`, `toomanyrequests`, `Retrying pull for image`) appears anywhere in
`_service-ci.yml`, verified against `origin/main`. So a match in a job log cannot be the echoed
`run:` script listing — the documented trap for this repo.

## FinOps — why option (b) is rejected

Over-the-wire (compressed, linux/amd64) size of the four pre-warmed images, read from the registry
manifests rather than estimated:

| image | compressed |
|---|---|
| `postgres:16.3-alpine` | 92.4 MB |
| `valkey/valkey:7.2-alpine` | 15.8 MB |
| `redpandadata/redpanda:v24.1.2` | 162.3 MB |
| `apicurio/apicurio-registry-mem:2.6.2.Final` | 393.6 MB |
| **total** | **664.1 MB = 0.649 GB** |

**Denominator.** 1756 hosted build jobs / 2.17 days = **809/day = ~24 600/month**.

**Option (b1) — extend the ECR pre-warm to hosted runners via OIDC.** ECR data-transfer-out to the
internet (eu-north-1, first 10 TB) is $0.09/GB. Every hosted build would pull all four images:

> 24 600 builds/mo x 0.649 GB x $0.09/GB = **~$1 437/month**

against a saving measured in **$0**, because this repo is **public** and GitHub-hosted Actions
minutes are free. The re-run costs latency and attention, not money. Rejected.

**Option (b2) — pre-warm only the images the module needs.** Best case (a postgres-only module) is
92.4 MB: 24 600 x 0.0902 GB x $0.09 = **~$200/month**, still against $0 saved, and it needs a
per-module image manifest that nothing derives today — a new hand-maintained list of the very thing
it covers, which is the gate-scope failure mode this repo has already been bitten by. Rejected.

**Option (b3) — a bounded retry around the pull.** Already implemented, by Testcontainers itself: the
reference log shows `Retrying pull for image: postgres:16.3-alpine (104s remaining)` counting down a
120 s budget that then expired. A second retry layer on top of a retry that already failed buys
nothing. Rejected on evidence.

Worth noting the observed error is `Client.Timeout exceeded while awaiting headers` /
`context deadline exceeded`, **not** `toomanyrequests`. So a Docker Hub login — the usual $0 fix,
raising the anonymous per-IP pull limit — would not have helped this failure either.

**Recommendation: option (a) only, $0/month.** The argument does not depend on the exact
registry-failure count: even at the **upper bound** where every one of the 125 failures were registry
failures (2.0% of all build attempts), $1 437/month to avoid a **free** re-run is not a trade worth
making. The real cost here is a reader sent to debug a test that is not broken, and that is fixed by
printing the right sentence. (The exact classification of those 125 is still running and will be
added as a comment; it can only move the number down, i.e. strengthen this conclusion.)

## What this adds

A post-step under `if: failure()` that runs
`.github/scripts/check-testcontainers-registry-failure.py --service <svc>`, which reads the JUnit XML
Gradle already wrote and emits a `::error` naming Docker Hub when a failure carries a registry
signature.

Two deliberate design choices, both about failure modes this repo has paid for:

* **It reads JUnit XML, not a teed console log.** Teeing would put a pipe in the repo's most
  load-bearing command, where a missing `set -o pipefail` silently swallows the build's exit code. A
  step that only reads artifacts already on disk always exits 0 and cannot change the build verdict.
  Bounded trade-off: if the pull failure aborts before any XML is written it prints nothing and
  behaves exactly as today. It never makes the diagnosis worse.

* **It parses the XML and inspects `<failure>`/`<error>` element text, never greps the file.** A
  whole-file grep cannot distinguish the signature from a test name or a comment mentioning it. In
  this direction the collision is a **false positive telling a reader to re-run a genuinely broken
  build**, so the protection is structural rather than comment-stripping.

## Falsification

`--self-test` asserts both directions, and both were broken in place and observed red before being
restored from a `cp` backup:

| break | result |
|---|---|
| primary signature disabled | `SELF-TEST FAIL: registry failure not detected` (exit 1) |
| parsed lookup replaced by whole-file grep | `SELF-TEST FAIL: false positive on prose-about-the-code` (exit 1) |
| restored | `SELF-TEST PASS` (exit 0) |

Also verified locally: `check-workflow-run-step-size` passes (largest run script 18872 / limit 19000;
the new step is small), `check-gate-script-registration` counts the new script as invoked (87
scripts, 0 never run), and `actionlint` finding sets are identical between `origin/main` and this
branch — with the oracle validated against a known-bad file first, so its silence means something.

This PR is itself the GitHub-side validation of the workflow edit: `services-ci.yml` builds a canary
(`openbank-libs`) on a recipe-only change to `_service-ci.yml`, so the modified reusable workflow is
parsed and executed by this PR's own run.

## Linked issues

Closes #3562
Refs #3515

### DIFF
--- a/.github/scripts/check-testcontainers-registry-failure.py
+++ b/.github/scripts/check-testcontainers-registry-failure.py
@@ -1,0 +1,190 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) OpenBank contributors. Licensed under the Apache License, Version 2.0.
+"""
+Name a Docker-registry fetch failure as such, instead of leaving it disguised as a test bug.
+
+THE PROBLEM THIS SOLVES
+-----------------------
+`_service-ci.yml` runs every PR build on `ubuntu-latest` (only the main-push lane gets the
+self-hosted pool). The "Pre-warm Testcontainers image cache via ECR" step is gated on
+`runner.environment == 'self-hosted'`, so on the PR lane it is skipped entirely: there is no
+ECR pull-through cache and no in-cluster registry-cache, and Testcontainers pulls
+postgres/valkey/redpanda/apicurio straight from Docker Hub. When Docker Hub is slow or
+rate-limits the shared hosted-runner egress IP, the pull times out.
+
+What the reader then sees at the top of the log is:
+
+    LedgerPactProviderVerificationTest > ... FAILED
+    LedgerSchedulerVertxContextIT > the scheduled FX revaluation reaches its use case() FAILED
+
+Two named test failures. The actual cause sits ~900 lines further down, as a
+`ContainerFetchException` wrapped in four layers of Quarkus test-resource plumbing. Measured on
+run 30769096077 (job 91553216566, PR #3515): the registry error appears 32 times in that log and
+the reader has no reason to scroll to it, because the summary already offered a plausible and
+completely wrong explanation. The failure is MISDIAGNOSED BY CONSTRUCTION — it is not that the
+signal is missing, it is that a more legible and incorrect signal is printed above it.
+
+This script reads the JUnit XML that Gradle writes for the failing tests and, if the stack traces
+carry a registry-fetch signature, prints a `::error` naming Docker Hub and recommending a re-run.
+
+WHY THE JUnit XML AND NOT THE CONSOLE LOG
+-----------------------------------------
+Reading the XML needs no change to the "Build + test" step. The alternative — teeing the Gradle
+output to a file — puts a pipe in the repo's single most load-bearing command, where a missing
+`set -o pipefail` silently swallows the build's exit code (this repo has been bitten by
+`gradlew | tail` doing exactly that). A post-step that only reads artifacts already on disk
+cannot change the build verdict at all, which is the correct blast radius for a diagnostic.
+
+The trade-off is deliberate and bounded: if the pull failure aborts the task before any JUnit XML
+is written, this prints nothing and the job behaves exactly as it does today. It never makes the
+diagnosis worse, and there is no case where it turns a green build red.
+
+WHY THE SIGNATURES ARE WHAT THEY ARE
+------------------------------------
+`ContainerFetchException` is the discriminating one: Testcontainers raises it only when it cannot
+obtain an image, so it cannot be confused with a container that started and then misbehaved. The
+registry hostnames and `toomanyrequests` are corroborating. Deliberately NOT matched:
+`ConditionTimeoutException` alone (awaitility is used all over the test suite for reasons that
+have nothing to do with Docker) and `InternalServerErrorException` alone (a 500 from any
+docker-java call, including ones a real bug could provoke).
+
+Note the code-about-code hazard the repo documents: a checker that greps a text file matches
+prose ABOUT the string as readily as the string itself, and in this direction — a false POSITIVE
+that tells a reader to re-run a genuinely broken build — that is the expensive mistake. It is
+avoided structurally here rather than by comment-stripping: the scan reads the `<failure>` and
+`<error>` element TEXT out of parsed XML, never the raw file, so a test NAME or a source comment
+mentioning `ContainerFetchException` cannot trigger it. `--self-test` asserts that specific
+non-match.
+
+Exit code is ALWAYS 0. This is a diagnostic, not a gate: it runs under `if: failure()` when the
+job is already failing, and a non-zero exit here would only add a second red step saying nothing
+new.
+
+Usage:  check-testcontainers-registry-failure.py --service <name> [--self-test]
+"""
+
+from __future__ import annotations
+
+import argparse
+import pathlib
+import sys
+import tempfile
+import xml.etree.ElementTree as ET
+
+# The discriminating signature. Testcontainers raises this ONLY on a failure to obtain an image.
+PRIMARY = "ContainerFetchException"
+
+# Corroborating signatures — any one of these alongside PRIMARY sharpens the message, and each is
+# also sufficient on its own because none of them can be produced by anything but a registry call.
+CORROBORATING = (
+    "registry-1.docker.io",
+    "toomanyrequests",
+    "You have reached your pull rate limit",
+    "Retrying pull for image",
+)
+
+
+def failure_texts(xml_path: pathlib.Path) -> list[str]:
+    """Return the text of every <failure>/<error> element, or [] if the file is unparseable.
+
+    Parsing rather than grepping is what keeps a test name or a source comment mentioning
+    `ContainerFetchException` from triggering a false positive.
+    """
+    try:
+        root = ET.parse(xml_path).getroot()
+    except (ET.ParseError, OSError):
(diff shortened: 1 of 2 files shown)


JiRaska:
## Exact classification (the measurement the body left pending)

All 125 failed `build (...)` job attempts in the sample had their logs fetched and scanned for the
registry signatures.

| | count |
|---|---|
| failed build-job attempts | 125 |
| **of which registry-fetch failures** | **5** |
| share of failures | **4.0%** |
| share of all 6323 build-job attempts | **0.079%** |
| wall-clock minutes burned | **69 min** over 2.17 days (~32 min/day) |

The recommendation in the body holds, and by a wider margin than the upper bound it argued from:
the true rate is 0.079%, not the 2.0% worst case.

### The runner split is the interesting part

| runner | registry failures | build-job attempts |
|---|---|---|
| `ubuntu-latest` (pre-warm **skipped**) | **5** | 1756 |
| `openbank-build` (pre-warm **runs**) | **0** | 4567 |

**5 of 5 on the hosted lane, 0 of 4567 on the self-hosted lane.** That is close to a natural
experiment, and it confirms the causal story rather than merely being consistent with it: the ECR
pre-warm is what prevents this, and its `runner.environment == 'self-hosted'` gate is exactly where
the exposure begins. Rate on the exposed population: 5/1756 = **0.28% of hosted builds**.

### What that does to the FinOps case

Extending the pre-warm to hosted runners would cost ~$1 437/month to eliminate ~69 min of re-run
latency per 2.17 days (~950 min/month) — about **$1.51 per minute of CI latency avoided**, against
**$0** of runner spend saved, because hosted Actions minutes are free on a public repo.

So the merged change is the right and only economical one: at 0.079% the problem is not the
frequency, it is that when it does happen the reader is handed a confidently wrong explanation.

### Method note

A run's top-level `conclusion` reports the latest attempt only, so this failure class is erased by a
re-run — the run behind the reference log now reads `success`. Counting it required enumerating
`/actions/runs/<id>/attempts/<n>/jobs` for all 1000 runs. A conventional "list failed runs" query
would have reported far fewer, and would have looked like a complete answer.

Issue #3496, opened 2026-08-02
Cross-AZ data transfer stepped 11x on 2026-07-27 and is now ~$809/mo — the largest EC2 cost in the account


JiRaska:
## What

`EUN1-DataTransfer-Regional-Bytes` (cross-AZ data transfer) stepped from ~$3/day to ~$33/day on **2026-07-27** and has held there for six days. At that run rate it is **~$809/mo**, which makes it:

- the **single largest EC2 line item** in the account, by an order of magnitude over compute;
- **~95%** of the combined `Amazon EC2 - Compute` + `EC2 - Other` spend;
- **~11x** what `openbank-infra/CLAUDE.md` currently records for it ("the only material transfer line is `EUN1-DataTransfer-Regional-Bytes` at $32.64 (~$70/mo)", measured 2026-07-16).

For scale: total AWS spend over the last 30 days is **$853.45**, and *all* EC2 instance usage across every NodePool is ~$116/mo.

## Evidence

`aws ce get-cost-and-usage --granularity DAILY --filter '{"Dimensions":{"Key":"USAGE_TYPE","Values":["EUN1-DataTransfer-Regional-Bytes"]}}'`, 2026-07-12 → 2026-08-01, `AWS_PROFILE=openbank`:

```
2026-07-12   2.53      2026-07-23   2.90
2026-07-13   2.70      2026-07-24   3.27
2026-07-14   3.54      2026-07-25   3.97
2026-07-15   2.64      2026-07-26   5.02
2026-07-16   5.53      2026-07-27  15.88   <-- step begins
2026-07-17   3.69      2026-07-28  32.42
2026-07-18   3.15      2026-07-29  31.77
2026-07-19   6.97      2026-07-30  36.15
2026-07-20   3.20      2026-07-31  36.61
2026-07-21   2.66      2026-08-01  30.86
2026-07-22   2.62
```

A clean ~11x step over two days, then a flat plateau. That is a change in the cluster, not organic growth.

## Why it matters beyond the money

Nothing alerts on this. It was found only while pricing an unrelated Karpenter node-shape change (#3495) and would otherwise have run indefinitely — the same failure class as the ČNB feed and the audit-chain table: a number that stops (or starts) moving, with no gate watching it.

## What to investigate

[...]

Issue #3488, opened 2026-08-02
agent-review fails ~40% of runs with num_turns=1, cost=$0 — the action never runs, and it blocks --auto merges


JiRaska:
## What

`agent-review` fails roughly 40% of the time with the Claude action never running at all, and the failure is indistinguishable from a real review verdict at a glance.

Signature, identical every time:

```
##[error]The transcript reports `is_error: true` (subtype=success, num_turns=1, cost=$0)
##[error]The run used num_turns=1. A real review reads the diff and submits a verdict,
         so it cannot complete in one turn — this is the #2161 signature.
##[error]Process completed with exit code 1
```

`cost=$0` and `num_turns=1` mean no turn was ever billed — the run died before reading anything, so this is not a review that disagreed, it is a review that did not happen.

## It is not transient, and it is not an outage

I called it transient earlier, on the evidence that 9 of 11 PRs passed at the same time. That was wrong: re-running two of the failures produced the identical failure on attempt 2, and successes and failures interleave minute by minute.

20 most recent runs:

```
21:01 success   20:56 failure   20:50 failure(attempt2)   20:42 success
21:00 success   20:55 success   20:49 success             20:41 failure
20:58 success   20:55 failure   20:47 failure(attempt2)   20:35 success
20:57 success   20:54 success   20:44 failure             20:30 failure
```

Two successes and two failures inside the same minute rules out an expired token or exhausted quota. The clustering tracks how many runs are in flight, which points at a concurrency or rate limit on whatever credential the workflow uses — but that is a hypothesis, and the run output is hidden (`show_full_output: false`, correctly, since the transcript would carry the diff).

## Why it costs something

`agent-review` is not a required check, so it does not block the ruleset. It **does** block `gh pr merge --squash --auto`, which refuses on `mergeStateStatus=UNSTABLE` regardless of which check is red. Today that turned several one-line gitops fixes into a manual merge each — and the documented escape from UNSTABLE is `--admin`, which is exactly the override nobody should be reaching for. (Plain `--squash` without `--auto` does work, which is worth knowing but is not obvious.)

The second cost is the one that compounds: a check that is red ~40% of the time for reasons unrelated to the diff is a check people stop reading. That is the alert-fatigue shape, on the PR page.

## What would settle it

[...]

PR #3572, opened 2026-08-03
docs(infra): correct why the default NodePool wedged — the repo shape is not the live one

JiRaska:
## What

Comment-only correction to the note #3570 added. No behaviour change.

## Why

#3570 raised the `default` NodePool cap correctly, but explained the deadlock wrongly. It blamed the xlarge/2xlarge floor set on 2026-08-02 and computed `48 + 4 > 48`.

**That change has never been applied.** Live, 2026-08-03:

```
kubectl get nodepool default
  instance-category  In [c, m, r]
  instance-size      In [large, xlarge, 2xlarge, 4xlarge]
  limits             {cpu: 48, memory: 128Gi}
```

The repo says `m`/`r`, `xlarge`–`2xlarge`, `192Gi`. `platform-tofu.yml` applies only on manual `workflow_dispatch`, so this file and the cluster drift silently whenever nobody dispatches it.

So the smallest node the live pool may provision is still a 2 vCPU `large`, and the true statement is **`48 + anything > 48`** — the pool is simply full (22 nodes, exactly 48/48 vCPU, 124.4Gi against 128Gi). Reverting the floor would not have unwedged anything, which is precisely the wrong conclusion the old text invited.

## Two things the corrected note now carries

- The 22 small nodes are the **steady state**, not a half-finished roll.
- The next apply lands the raised cap **and** the 2026-08-02 shape change together, so it will drift and roll all 22 nodes at once. The surge headroom from #3570 is what makes that roll possible at all.

## Verification

`tofu fmt -check` clean. Diff is one file, comment lines only.

## Linked issues

Refs #3496

### DIFF
--- a/openbank-infra/aws/envs/sandbox-platform/main.tf
+++ b/openbank-infra/aws/envs/sandbox-platform/main.tf
@@ -379,14 +379,27 @@ resource "kubectl_manifest" "nodepool_default" {
       # Deployment read 1/1 Available; the Pending pods were blocked scale-ups,
       # so HA was degraded, not lost).
       #
-      # The deadlock is self-inflicted by the xlarge/2xlarge floor set on
-      # 2026-08-02: the 22 EXISTING nodes are the old `large` shape and already
-      # fill the cap, while the smallest node Karpenter is now allowed to
-      # provision is 4 vCPU. 48 + 4 > 48, so it can neither schedule a new pod
-      # nor surge a replacement — which also means it cannot roll the drift
-      # created by that change, or by the single-AZ pin above. A zone
-      # requirement drifts every node in the pool; without surge headroom that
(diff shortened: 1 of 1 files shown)


JiRaska:
The correction is the valuable half: #3570 raised the cap correctly and explained it with a repo state that is not the live one.

That distinction is worth keeping visible. A comment citing a constraint the cluster does not have sends the next reader to the wrong lever — and unlike a wrong code path, nothing will ever contradict it. The repo cannot tell you that `instance-size In [large, xlarge, 2xlarge, 4xlarge]` is what is actually applied; only `kubectl` can, which is precisely why the wrong explanation survived review.

Merging: 5/5 required green; the failing `agent-review` is the ~40% flake (#3488), unrelated.

PR #3576, opened 2026-08-03
fix(admin-ui): revert the jsdom 30 bump — it disabled the whole test suite

JiRaska:
The admin-ui vitest suite has not run on `main` since `ca2d18a87`. Not failing tests — **no tests at all**:

```
Error: [vitest-pool]: Failed to start forks worker for test files .../src/test/i18n.test.ts
Caused by: TypeError: webidl.util.markAsUncloneable is not a function
  ❯ new CacheStorage  node_modules/undici/lib/web/cache/cachestorage.js:20
  ❯                   node_modules/jsdom/lib/api.js:12

Test Files  no tests
```

## Bisected

```
94a2b71cf   success   chore(admin-ui): deploy sandbox-653d1bed
ca2d18a87   failure   chore(deps-ui)(deps-dev): Bump jsdom      ← first red
eec92b9b0   failure
2866b3bd5   failure
5510c0f83   failure   … and every main commit since
```

The bump is a **major**, and it moves undici transitively:

```
before   jsdom 29.1.1   undici 7.28.0
after    jsdom 30.0.1   undici 8.9.0
```

undici 8 calls `webidl.util.markAsUncloneable`, which the Node 20 that CI runs (`node-version: "20"`) does not provide. jsdom 29 stays on undici 7 and does not.

This PR sets `jsdom` back to `^29.1.1`; `npm install --package-lock-only` resolves undici back to 7.29.0.

## Why it got in

`Admin UI build` is not a required context, so its red did not block the merge. And the symptom is quiet: `Test Files no tests` is one line that reads like a pass unless you are looking for a count. Roughly 1,026 tests stopped running and nothing said so.

## What I could not verify locally

This machine runs Node v26.5.0, which **has** `markAsUncloneable` — so jsdom 30 works here and the failure does not reproduce. CI's Node 20 is the only environment that exercises it, which makes CI the oracle for this change rather than a local run. That is stated rather than papered over: the causal chain above is bisect plus dependency resolution, and the proof is the check going green on this PR.

## The alternative

Moving CI to a Node that satisfies undici 8 would let jsdom 30 stay. That is a larger change with its own blast radius and belongs in its own PR — this one restores the suite tonight.

Refs #3568 (a later ui-minor bump, red for this same reason and not its own fault)

### DIFF
--- a/openbank-admin-ui/package-lock.json
+++ b/openbank-admin-ui/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "openbank-admin-ui",
-  "version": "0.90.2",
+  "version": "0.91.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "openbank-admin-ui",
-      "version": "0.90.2",
+      "version": "0.91.0",
       "dependencies": {
         "@hookform/resolvers": "5.5.7",
         "@radix-ui/react-dialog": "1.1.23",
@@ -58,7 +58,7 @@
         "autoprefixer": "10.5.4",
         "eslint": "9.39.5",
         "eslint-config-next": "16.2.12",
-        "jsdom": "^30.0.1",
+        "jsdom": "^29.1.1",
         "postcss": "8.5.23",
         "tailwindcss": "4.3.3",
         "typescript": "6.0.3",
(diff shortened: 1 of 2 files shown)


JiRaska:
Verified the premise independently against current `main` rather than reading the bisect — pulled the `Admin UI build` job from the latest CI run on `main` (run 30791658939, job 91616321744):

```
Error: [vitest-pool]: Failed to start forks worker for test files .../openbank-admin-ui/...
Caused by: TypeError: webidl.util.markAsUncloneable is not a function
```

Repeated per test file, no worker ever starts. So the suite is not merely red, it produces no result at all — every assertion in it has been unenforced since `ca2d18a87`.

Also checked whether the open Dependabot PR would put it back: #3568 (`ui-minor` group) touches neither `jsdom` nor `undici`. Its `Admin UI` / `Admin UI build` failures are inherited from `main`, so this revert unblocks that PR rather than racing it.

The distinction the PR draws is the one that matters — a major bump that breaks the *runner* does not look like a broken test, it looks like a suite with nothing in it, and "no tests" is the one result a green-or-red check cannot express.

Issue #3352, opened 2026-08-02
default nodepool hits its limit on reservations nobody uses: CPU requested 11x actual, and 36 VPA recommendations sit unread


JiRaska:
`KarpenterNodePoolNearLimit` on the `default` pool has been firing ~19h, and during today's canary wave it stopped being cosmetic: 8 `FailedScheduling` events for money-path pods (`ledger` ×1, `payments` ×7) with

```
Failed to schedule pod, all available instance types exceed limits for nodepool
```

It cleared on its own as old replicas terminated (0 pending pods now), so this is a squeeze during deploys rather than a hard block. But the obvious remedy — raise the pool limit — would be **treating the symptom, and paying for it**.

## The pool is full of reservations nobody draws on

Requests vs actual usage, scoped per nodepool (containers where both were observable):

```
default        CPU req= 28.5  use= 2.65  (  9%)    MEM req= 65.5Gi  use= 37.8Gi  (58%)
runners        CPU req=  3.5  use= 0.21  (  6%)    MEM req=  7.3Gi  use=  6.0Gi  (82%)
runners-warm   CPU req=  0.4  use= 0.03  (  7%)    MEM req=  0.2Gi  use=  0.1Gi  (29%)
observability  CPU req=  1.7  use= 0.53  ( 31%)    MEM req=  7.1Gi  use=  6.0Gi  (83%)
```

On the constrained pool, **CPU is requested ~11× what is used**. Karpenter bin-packs by *requests*, not usage, so those reservations are what forced 22 nodes and then hit the limit.

Pool state for reference:

```
default        44 / 48 CPU (92%)    118 / 128 GiB (92%)    22 nodes
runners        20 / 64 CPU (31%)
observability   4 / 16 CPU (25%)
runners-warm    8 /  8 CPU (100%)   ← at limit by design; see the note at the end
```

## VPA already computed the answer and nobody reads it

There are **36 VPA objects, all 36 with recommendations, and all 36 in `updateMode: Off`**:

```
target cpu distribution: {50m: 19, 25m: 4, 11m: 6, 100m: 5, 323m: 1, 1: 1}
```

Nineteen services are recommended **50m**. The data to fix this has been sitting in the cluster, computed continuously, acted on never — the same shape as the other findings today, where the detection existed and the reading did not.

## Caveats — I am not proposing a fleet-wide cut

Three reasons not to just apply VPA targets across 36 services:

[...]

PR #3583, opened 2026-08-03
fix(observability): stop KarpenterNodePoolNearLimit firing 100% of the time on a fixed-size pool

JiRaska:
## What

Adds `nodepool!="runners-warm"` to `KarpenterNodePoolNearLimit`, and corrects the advertised
threshold on `PodPendingUnschedulable`. One file, no new alerts — this **subtracts** a permanent
false positive.

Refs #3352.

## Why

`#3386` established that `karpenter_nodepools_usage` measures **provisioned capacity, not demand**,
so for a fixed-size pool the usage/limit ratio is a constant rather than a utilisation signal. It
applied the `runners-warm` exclusion to `KarpenterNodePoolAtLimit`. `NearLimit` sits directly above
it in the same file and kept firing on the same pool for the same wrong reason.

`runners-warm` is 2 on-demand nodes with `expireAfter: Never` (`arc-runners.tf`) against a limit of
8 CPU, so it sits at 8/8 = 1.0 forever. `1.0 > 0.85`, so the rule **could never not fire on it**.

Measured 2026-08-03 over the full retained window:

```
KarpenterNodePoolNearLimit, firing:
  pool=runners-warm  res=cpu     103/103 = 100% of retained window
  pool=default       res=cpu      99/103 =  96%
  pool=default       res=memory  100/103 =  97%
```

and in Alertmanager at the same moment it was the **only** `detector=capacity` alert active.

> Note on that measurement: Prometheus retention here is **~8.6h**, not the 7d a range query will
> happily accept. My first pass reported "9%" because it divided by 2016 requested samples instead
> of the 103 that exist. Check the span before trusting a percentage from this cluster.

### What it cost

This is not hygiene. The same morning, during the 04:00Z cap stall (`default` pinned at 48/48, 22
pods Pending), the real alerts fired **correctly and for an hour** — `KarpenterNodePoolAtLimit`
`firing` on `default` from 03:35Z to 04:35Z — and nobody acted on them. The stall was found by a
person looking at pods, which is the exact discovery path ADR-0173 D1 exists to replace.

A capacity warning that is always on trains the reader to skip capacity warnings. Removing the
permanent firer is what makes the real one legible.

## Verification

Both directions, against live Prometheus, at the same instant:

| check | expectation | result |
|---|---|---|
| new expression | `runners-warm` absent | 0 series |
| **old** expression, same instant | `runners-warm` present | **exactly 1 series: `runners-warm`** |
| new expression at threshold `0.5` | `default` still matched | 2 series, both `default` |

The middle row is the one that matters: it shows the change removes precisely the false positive and
nothing else. The third shows the rule is **narrowed, not emptied** — a rule that matches nothing is
the failure mode this repo keeps finding, so it is asserted rather than assumed.

`default` stays covered and demonstrably still works: after #3570 raised the cap it sits at
56/72 = 78%, below this threshold, and the alert resolved on its own.

## Second change: PodPendingUnschedulable says 15m, delivers ~30m

Its two 15-minute clauses **compose rather than overlap** — the expression only becomes true once
the pod is older than 900s, and `for: 15m` then requires it to stay true for another 15m.

Measured: `audit-service-5c5dbb8679-lcp69` was created `03:44:58Z` and the alert first went `firing`
at `04:16Z` — **31 minutes**, against a summary reading "Pending for >15m".

**Behaviour is deliberately unchanged.** Dropping the staleness clause would make it fire twice as
early and roughly twice as often, and that noise cost has not been measured — ordinary rolling
updates park pods in Pending briefly, so this rule sits one threshold away from alerting on every
deploy. This PR fixes the text and records what to measure first.

## Risk

Low, and asymmetric in the safe direction. The only behavioural change is that one permanently-firing
warning on a pool that is idle by construction stops firing. `AtLimit` (critical) on that pool was
already excluded by #3386, so no coverage is lost that #3386 did not already decide to drop. If
`runners-warm` ever becomes elastic, both exclusions need revisiting together — they now state the
same rationale in the same file.

### DIFF
--- a/openbank-infra/gitops/components/observability/prometheus-rules-capacity.yaml
+++ b/openbank-infra/gitops/components/observability/prometheus-rules-capacity.yaml
@@ -37,15 +37,41 @@ spec:
         # prune, or raise the cap deliberately (an ADR-0173 decision), not in a
         # hurry during an incident.
         - alert: KarpenterNodePoolNearLimit
+          # EXCLUDES runners-warm, for exactly the reason KarpenterNodePoolAtLimit below
+          # already does (#3386) — that fix was applied to the AtLimit rule only, and this
+          # rule kept firing on the same pool for the same wrong reason.
+          #
+          # `karpenter_nodepools_usage` measures PROVISIONED capacity, not demand. For a
+          # fixed-size pool the ratio usage/limit is therefore a constant, not a
+          # utilisation signal: runners-warm is 2 on-demand nodes with `expireAfter: Never`
+          # (arc-runners.tf) against a limit of 8 CPU, so it sits at 8/8 = 1.0 forever.
+          # 1.0 > 0.85, so this rule can never NOT fire on it.
+          #
+          # Measured 2026-08-03 over the full retained window (Prometheus retention here is
+          # ~8.6h, not the 7d a naive range query assumes — check the span before trusting
+          # a percentage): runners-warm was firing 103/103 samples = 100%. It was also the
+          # ONLY detector=capacity alert active in Alertmanager at the time.
+          #
+          # That is the cost of leaving it: during the 2026-08-03 04:00Z cap stall the
+          # `default` pool alerts fired correctly and for an hour — AtLimit 03:35Z-04:35Z,
+          # NearLimit 96-97% of the window — and nobody acted on them. A capacity warning
+          # that is always on trains the reader to skip capacity warnings, so the one time
+          # it carried information it read as more of the same. Removing the permanent
+          # firer is what makes the real one legible; this is a SUBTRACT, not a new alert.
+          #
+          # `default` stays covered, and demonstrably still works: after #3570 raised the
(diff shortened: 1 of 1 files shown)


JiRaska:
Confirmed the asymmetry against `origin/main` rather than the description — `prometheus-rules-capacity.yaml` holds exactly two Karpenter alerts, and only one carries the exclusion:

```
KarpenterNodePoolAtLimit    nodepool!="runners-warm"   present
KarpenterNodePoolNearLimit                             absent
```

So #3386's reasoning was applied to one of the pair and the sibling directly above it kept firing on the same pool for the same wrong reason. That is the shape worth naming: a fix reasoned about a *metric* and applied to a *rule*, and the rule next to it inherited neither.

Subtracting a permanent false positive is the right instinct here. An alert that fires 100% of the time on a fixed-size pool is not a weak signal, it is a rule that trains people to skip the board — and it does that to every alert on it, not just this one.

Merging: 5/5 required green; the failing `agent-review` is the ~40% flake (#3488), unrelated to this diff.

PR #3589, opened 2026-08-03
fix(ci): run the admin-ui jobs on the Node version the app ships on

JiRaska:
## main is broken: every admin-ui test job collects zero tests

Since #3569 (`chore(deps-ui)(deps-dev): Bump jsdom`) merged, the `Admin UI` and `Admin UI build`
jobs fail on every PR with 79 identical errors and **no tests run at all**:

```
Error: [vitest-pool]: Failed to start forks worker for test files …/src/test/<any>.test.ts
Caused by: TypeError: webidl.util.markAsUncloneable is not a function
    ❯ new CacheStorage node_modules/undici/lib/web/cache/cachestorage.js:20:17

 Test Files  no tests
      Tests  no tests
     Errors  79 errors
```

## Cause

The declared engines and the CI runtime disagree:

| package | engines |
|---|---|
| `jsdom@30.0.1` | `^22.22.2 \|\| ^24.15.0 \|\| >=26.0.0` |
| `undici@8.9.0` (pulled by jsdom) | `>=22.19.0` |
| CI `setup-node` | **20** |

`markAsUncloneable` does not exist on Node 20, so undici's `CacheStorage` throws while the worker is
starting — before a single test file is imported. `engine-strict` is off, so `npm ci` installed it
without complaint and the mismatch surfaced only as a red job with an empty summary.

## Fix

Pin the admin-ui jobs to **26** — the version the app actually ships on
(`openbank-admin-ui/Dockerfile: FROM node:26-alpine`). CI was testing three majors behind the
runtime, which is the condition that let this land: a dependency legal for production was illegal for
CI, and nothing compared the two.

Three pins move (`ci.yml` admin-ui tests, `ci.yml` app-status which runs `npm ci` against the same
lockfile, `admin-ui-deploy.yml`). `security.yml`'s CodeQL pin stays at 20 — it does not install
admin-ui dependencies, so it is out of scope here rather than deliberately left behind.

## Verification

The failure is reproducible rather than intermittent — identical on two consecutive runs of #3584,
naming a different test file each time only because worker startup order varies. Locally these tests
pass on Node 26 (`node --version` → v26.5.0), which is the same runtime this pins CI to.

What this does **not** prove: that every admin-ui suite is clean on 26 in CI. That is what this PR's
own run establishes, and it is the thing to read before merging — a green here means the workers
started AND the suites passed; the current state on main cannot distinguish those two.

### DIFF
--- a/.github/workflows/admin-ui-deploy.yml
+++ b/.github/workflows/admin-ui-deploy.yml
@@ -118,7 +118,13 @@ jobs:
 
       - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
         with:
-          node-version: '20'
+          # Matches the runtime image (openbank-admin-ui/Dockerfile: node:26-alpine). Pinned at 20
+          # it was three majors behind what ships, and #3569's jsdom bump made that fatal:
+          # jsdom 30 declares engines ^22.22.2 || ^24.15.0 || >=26 and pulls undici 8, whose
+          # cachestorage calls webidl.util.markAsUncloneable. On 20 every vitest worker died
+          # with "Failed to start forks worker", the run collected ZERO tests, and the job
+          # still reported a plain red — engine-strict is off, so npm ci installed it happily.
+          node-version: "26"
           cache: 'npm'
           cache-dependency-path: openbank-admin-ui/package-lock.json
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -85,7 +85,13 @@ jobs:
 
       - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
         with:
-          node-version: "20"
+          # Matches the runtime image (openbank-admin-ui/Dockerfile: node:26-alpine). Pinned at 20
+          # it was three majors behind what ships, and #3569's jsdom bump made that fatal:
+          # jsdom 30 declares engines ^22.22.2 || ^24.15.0 || >=26 and pulls undici 8, whose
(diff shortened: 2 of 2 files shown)


JiRaska:
Complementary to #3576, not competing — worth stating since both landed on the same defect within the hour and a reader could reasonably close one.

#3576 (merged) reverts jsdom 30 → 29, which fixes the breakage now. This fixes why it was reachable: CI runs the admin-ui jobs on Node 20 while the app ships on 26.

Verified both halves against `origin/main` rather than the description:

```
openbank-admin-ui/Dockerfile:185,210,244   FROM node:26-alpine   (deps / build / runtime)
.github/workflows/ci.yml:88,476            node-version: "20"

package-lock.json after #3576:
  jsdom   29.1.1   engines: ^20.19.0 || ^22.13.0 || >=24.0.0    ← 20 is in range
  undici   7.29.0  engines: >=20.18.1
```

So `main` is green again on jsdom 29, and this PR is not needed to unbreak it — it is needed because the tests are exercised on a runtime three majors from the one that serves users, and nothing says so. The next transitive bump that assumes a modern Node reproduces the same failure, and the failure mode is the bad one: `engine-strict` is off, so `npm ci` installs happily and the incompatibility only appears as workers that will not start.

Also confirmed `openbank-admin-ui/package.json` declares no `engines` field at all, so there is nothing in the repo today that would have caught the mismatch either.

One thing I did not verify: whether anything else in `ci.yml` or `admin-ui-deploy.yml` depends on Node 20 specifically. The diff only changes the `setup-node` inputs, and both workflows' admin-ui jobs are the only consumers I could see, but a full green on this PR is the real evidence there.

PR #3594, opened 2026-08-03
fix(fx): give a derived inverse rate its own identity

### LINKED ISSUE #3374: fx: derived inverse rate reuses the source rate's id, so two directions share one identifier
## What

`GET /api/v1/fx/rates/{base}/{quote}` returns the **source rate's `id`** on a derived
inverse pair, so two opposite directions answer under one identifier.

Measured against `sandbox-6138c3a7` in sandbox:

```
GET /api/v1/fx/rates/EUR/CZK
{"id":"dd01a476-b970-4f17-bd49-850862a6d370","baseCurrency":"EUR","quoteCurrency":"CZK",
 "bidRate":24.80000000,"askRate":25.30000000, ...}

GET /api/v1/fx/rates/CZK/EUR
{"id":"dd01a476-b970-4f17-bd49-850862a6d370","baseCurrency":"CZK","quoteCurrency":"EUR",
 "bidRate":0.03952569,"askRate":0.04032258, ...}
```

Same `id`, different `baseCurrency`/`quoteCurrency`/rates. The numbers are correct —
`1/25.30 = 0.03952569`, `1/24.80 = 0.04032258`, bid/ask swapped so the spread still runs
against the client. Only the identity is wrong.

## Why it matters

`FxRate.inverted()` (added in #3189) uses `copy(...)`, which carries `id` over. That was the
right call for shipping the fix — the inverse is not a stored row and has no id of its own —
but it leaves the id no longer identifying what it names:

- A client caching quotes by `id` stores one direction and serves it for the other. The app
  currently keys off the pair, so this is latent, not live.
- A conversion audit record that references the quote `id` cannot be replayed to a direction:
  the same id reads as EUR/CZK or CZK/EUR depending on which way the caller asked. That is the
  half that matters for money-path evidence.

## Options

1. **Derive a deterministic id for the inverse** — e.g. UUIDv5 over `(sourceId, "inverse")`.
   Stable across calls, distinct per direction, no schema change.
2. **Null the `id` on a derived rate** and add `derivedFrom: <sourceId>` to the response.
   Most honest about what the object is; an API-contract change (`openapi.yaml` + version bump).
3. **Persist inverse rows** at ingestion. Removes the special case at read time but doubles the
   table and puts the rounding in the DB, where a later precision change cannot restate it.

Preference is (2) — it makes "this is a derived quote" visible to the caller, which (1) hides.
It costs an `info.version` minor and a contract test.

## Not a regression

Before #3189 `CZK/EUR` was a permanent HTTP 502 — no id at all. This is a defect in new
behavior that is otherwise working, not a break of something that used to be right.

Refs #3189

JiRaska:
Closes #3374.

## The defect

`GET /api/v1/fx/rates/EUR/CZK` and `.../CZK/EUR` returned **the same `id`** with different pairs and inverted numbers. The rates were right (#3189 fixed a permanent 502 there); only the identity was wrong — and an audit record citing that id cannot be replayed to a direction, which is the half that matters for money-path evidence.

## Why not the option the issue preferred

#3374 proposed nulling the `id` and adding `derivedFrom`. That would have failed hard:

```
fx_conversions.rate_id  NOT NULL REFERENCES fx_rates(id)
```

A derived quote has no row, so writing either a synthetic id or `null` breaks **every CZK→foreign conversion** on a foreign-key violation. `FxRate.inverted()`'s KDoc was protecting exactly this — `FxConversion.rateId` must keep pointing at the row the price came from.

So this combines the issue's options (1) and (2):

- the derived quote gets its **own deterministic id**, from `(sourceId, "inverse")` — stable across calls, so still usable as a cache key or audit reference, and distinct per direction;
- **`derivedFrom`** names the stored row explicitly, making "this is a derived quote" readable instead of inferred;
- `FxService` persists `rate.derivedFrom ?: rate.id`, so the price stays traceable to the row. A unit test pins that property directly (`conv.rateId == stored.id` **and** `!= stored.inverted().id`), on both the derived and the direct path.

## Spec

`1.5.0 → 1.6.0`. `ExchangeRateResponse` declared **neither** `id` nor `derivedFrom` before — the endpoint has always served an undocumented `id` — so documenting both is additive. Version re-read off live `origin/main` immediately before committing (still `1.5.0`, no race).

## The test finding, which is the interesting part

The new contract test uses a **synthetic currency pair**, not EUR/CZK, and the reason generalises.

`V1__create_fx.sql:39` seeds a stored **`CZK/EUR`** row. `getRate` tries the direct direction first, so in a freshly migrated database `GET /rates/CZK/EUR` answers from that row and **never derives anything** — the path under test does not run, and the assertion sees a legitimate null `derivedFrom`. Written against the real pair, the test failed for precisely that reason:

```
expected: "73932f33-8a7e-4dbd-ba2f-aa539dcb7f8a"
 but was: "null"
```

That also explains why **#3374 was observable in sandbox and not locally**: the seeded row's `validTo` is `NOW() + INTERVAL '1 day'`, fixed at migration time, and `findLatest` filters `validTo > now`. So the seed ages out after a day — a long-lived environment derives CZK/EUR, a fresh one serves the stored row. Nobody would have reproduced this on a clean database.

Synthetic ISO 4217 codes in the user-assigned `XA`-`XZ` range, so no migration seed, no other test and no real fixing can supply the reverse direction and quietly make the test vacuous.

## Contract-test shape

Not a pact, deliberately. The one committed pact for this provider covers the **stored** direction and pins four fields, none of them `id`; the defect lives in the **derived** direction, which no in-repo consumer calls. A pact for it would be a contract against a fictional consumer. The counterparty for this property is the committed `openapi.yaml`, so that document is what the assertions read — **parsed by JSON-Pointer, never grepped**, so it cannot pass by matching prose (#2291's failure mode). Delete `id` or `derivedFrom` from the schema and the test fails before it issues a request.

The existing pact is untouched and still replayed on every PR by `FxPactFolderProviderVerificationTest`.

## Verification

```
:openbank-fx-service:build          BUILD SUCCESSFUL
:openbank-fx-service:koverVerify    BUILD SUCCESSFUL
:openbank-fx-service:ktlintCheck    BUILD SUCCESSFUL
:openbank-fx-service:detekt         BUILD SUCCESSFUL
```

## Self-assessment

- **`money_path_services` includes `openbank-fx-service`**, so this wants 2 approvals by rule. Expect the merge to be blocked; that is the correct state.
- **Known residual spec drift, deliberately untouched:** `ExchangeRateResponse` also declares `rate` and `validAt`, which the endpoint has never served, and omits nine properties it does serve. Exact set equality is the right end state, but reaching it means *removing* two declared properties, which `oasdiff` classifies as breaking and would force `/api/v2` (ADR-0048 ties the spec major to the URL). That is a separate decision from fixing an id.
- **Not measured:** whether any external consumer already stores the old shared id. In-repo there is none, and the id was undocumented, but I cannot see outside the repo.
- The original authoring stalled mid-gate; I finished it, and the contract-test failure above was found by running the gate rather than by review.

### DIFF
--- a/openbank-fx-service/build.gradle.kts
+++ b/openbank-fx-service/build.gradle.kts
@@ -60,6 +60,10 @@ dependencies {
     // Consumer-driven contract for the aml-service case store (issue #2255, C3): fx is the consumer
     // of POST /api/v1/aml/cases, so this module also generates a pact.
     testImplementation(libs.pact.consumer)
+    // FxRateApiContractTest asserts the running endpoint against the committed openapi.yaml, which
+    // it PARSES rather than greps (#3374). No version: the Quarkus BOM manages it, same as
+    // openbank-ap2-service and openbank-pid-service, which carry the identical test shape.
+    testImplementation("com.fasterxml.jackson.dataformat:jackson-dataformat-yaml")
 }
 
 kover {
--- a/openbank-fx-service/src/main/kotlin/com/openbank/fx/application/usecase/FxService.kt
+++ b/openbank-fx-service/src/main/kotlin/com/openbank/fx/application/usecase/FxService.kt
@@ -291,7 +291,10 @@ class FxService(
         toAmountMinorUnits = toAmountMinorUnits,
         appliedRate = rate.askRate,
         feeMinorUnits = feeMinorUnits,
-        rateId = rate.id,
+        // The STORED row, never the derived quote's own id (#3374). `fx_conversions.rate_id` is
+        // `NOT NULL REFERENCES fx_rates(id)`, so a derived id — which has no row — would fail the
+        // insert on a foreign-key violation for every CZK→foreign conversion.
+        rateId = rate.derivedFrom ?: rate.id,
         status = status,
         createdAt = createdAt,
         settledAt = settledAt,
--- a/openbank-fx-service/src/main/kotlin/com/openbank/fx/domain/model/FxRate.kt
+++ b/openbank-fx-service/src/main/kotlin/com/openbank/fx/domain/model/FxRate.kt
@@ -6,6 +6,8 @@ package com.openbank.fx.domain.model
 
 import java.math.BigDecimal
 import java.math.RoundingMode
+import java.nio.charset.StandardCharsets
+import java.security.MessageDigest
 import java.time.Instant
 import java.util.UUID
 
@@ -23,6 +25,12 @@ data class FxRate(
     val validFrom: Instant,
     val validTo: Instant,
     val createdAt: Instant,
+    /**
+     * The id of the stored quote this one was derived from, or `null` when this IS the stored
+     * quote. Set only by [inverted]; nothing that reads a row out of `fx_rates` ever sets it,
+     * which is what makes "is this a derived quote?" answerable rather than inferred.
+     */
+    val derivedFrom: UUID? = null,
 ) {
     val pair: String get() = "$baseCurrency/$quoteCurrency"
     val midRate: BigDecimal get() = (bidRate + askRate).divide(BigDecimal.TWO)
@@ -44,10 +52,27 @@ data class FxRate(
      * CZK→foreign exchange — a systematic loss, in the customer's favour on one leg and the
      * bank's on the other, which is exactly the kind of error that does not announce itself.
      *
-     * [id] is carried over unchanged: this is a view of the SAME quote, not a new one, and
-     * FxConversion.rateId must keep pointing at the row the price came from.
+     * ## Identity (#3374)
+     *
+     * The first version of this method carried [id] over unchanged, so `EUR/CZK` and `CZK/EUR`
+     * answered under ONE identifier with different pairs and inverted numbers — an id that no
+     * longer identified what it named. A conversion audit record citing that id cannot be replayed
+     * to a direction.
+     *
+     * So the derived quote gets its own [id], [derivedId] over `(sourceId, "inverse")`: stable
+     * across calls (a client may cache by id) and distinct per direction. The link back to the row
+     * is not lost, it is made EXPLICIT in [derivedFrom] — which is the half the caller could not
+     * see before.
+     *
+     * **`FxConversion.rateId` must still reference the STORED row**, and not merely as bookkeeping:
+     * `fx_conversions.rate_id` is `NOT NULL REFERENCES fx_rates(id)`, and a derived id has no row,
+     * so writing it would fail every CZK→foreign conversion on a foreign-key violation. `FxService`
+     * therefore persists `rate.derivedFrom ?: rate.id`. Nulling the id instead — the shape #3374
+     * first proposed — would have hit the same constraint from the other side.
      */
     fun inverted(): FxRate = copy(
+        id = derivedId(id),
+        derivedFrom = id,
         baseCurrency = quoteCurrency,
         quoteCurrency = baseCurrency,
         bidRate = BigDecimal.ONE.divide(askRate, INVERSE_SCALE, RoundingMode.HALF_UP),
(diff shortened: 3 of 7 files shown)


JiRaska:
The `gates (api)` failure here is not your diff — it is the gate matching your KDoc.

```
::error V8.1 openbank-fx-service:.../domain/model/FxRate.kt:103: banned crypto primitive
::error V8.1 openbank-fx-service:.../domain/model/FxRate.kt:104: banned crypto primitive
```

Lines 103 and 104 are the two comment lines explaining why you picked RFC 9562 v8 *instead of* v5:

```
103   * Version 8 rather than the version 5 #3374 suggested because v5 is *defined* as SHA-1, and
104   * a SHA-1 call in a money-path service is a finding every scanner in this repo will raise —
```

Both begin with `*`. The code calls `MessageDigest.getInstance("SHA-256")`. `check-asvs-l3.py`'s `find_v8` scans raw lines with no comment handling, so the sentence predicting the scanner is what the scanner matched — and your prediction was right in a way you did not intend.

Its docstring already claims comments are excluded ("MD5-for-checksums-in-comments are not matches by pattern"), so the property was asserted and never implemented.

Fix in #3608: strip comments before matching, depth-tracking because Kotlin block comments nest, string literals deliberately left in (`getInstance("SHA-1")` is the call the rule exists to catch). Measured fleet-wide as suppressing zero previously-found findings, and its `--self-test` keeps three must-still-flag cases so the fix cannot silently disable V8.1.

Once that lands, rebase and `gates (api)` should go green without you touching the KDoc. Please do not reword it as a workaround — the comment is correct and useful, and rewriting it would leave the gate broken for whoever hits it next.

Your other three failures (`Validate manifests`, `verification-metadata`, `all-green`) I have not diagnosed yet.

JiRaska:
Heads-up — I opened #3741 implementing the same issue (#3374) with the issue's preferred option (2) (`id: null` + `derivedFrom` on derived quotes), before noticing this PR.

Two findings from comparing the approaches:

1. **The ASVS L3 failure here is structural, not fixable by tuning.** `V8.1 no banned crypto primitives` flags `FxRate.kt:103-104` — the `MessageDigest` call that any name-based UUID (v5 or v8) needs. Option (1) cannot pass that gate in a money-path service; option (2) needs no hashing at all.
2. **The FK objection to option (2) doesn't hold.** `fx_conversions.rate_id = rate.id` is taken from the *domain* `FxRate`, and `inverted()` keeps the source row's id internally — so conversions keep referencing a real `fx_rates` row even when the REST DTO nulls the response `id`. The null never reaches persistence.

#3741's use-case layer returns `ResolvedRate(rate, derivedFrom)`, so the `derivedFrom`-naming idea from this PR survives there too — just without a synthetic id. Happy to close mine if you'd rather land this one, but the gate failure above looks like a dead end.

JiRaska:
## Two open PRs fix #3374, with different contracts

Flagging this because both are money-path, both are open, and merging either one leaves the other conflicting on `FxService.kt` and `openapi.yaml`. This is a decision, not a duplicate to close.

| | #3594 (08-03, +461/−9) | #3741 (08-05, +155/−26) |
|---|---|---|
| derived quote's `id` | its own, **computed** | **`null`** |
| how | `derivedId(sourceId)` — RFC 9562 v8 over SHA-256 of `"<ns>:<sourceId>"`, deterministic | not synthesised; `FxRateResponse.of` sets `id = null` when `derivedFrom != null` |
| `derivedFrom` | yes | yes |
| where the change lands | domain (`FxRate`) | REST DTO; domain keeps the source id |
| API contract | `id` stays required | **`id` becomes nullable** |

Both fix the actual defect — two directions answering under one identifier, so a client caching by `id` serves one direction for the other.

## What I would want decided, in order

**1. The foreign key.** #3594's description states that writing either a synthetic id or `null` breaks every CZK→foreign conversion on `FULL REFERENCES fx_rates(id)`. #3741 avoids that by keeping the source id in the domain and nulling only in the response — `FxConversion.rateId` still points at a real row. If that reading is right, #3741 sidesteps a constraint #3594 has to satisfy, and #3594 should say how it does.

**2. Nullable `id` is an API-contract change.** oasdiff has been unkind to nullable-shape changes in this repo before. `openbank-fx-service`'s `info.version` needs to reflect whatever oasdiff classifies this as, read off **live `main`** rather than either branch's base — two racing spec PRs can both claim the same version and both pass the gate, because the gate classifies against creation-time base.

**3. Is a derived quote addressable at all?** This is the real product question under the mechanism. #3594 says yes and pays a hash construction for it; #3741 says no and makes clients carry `derivedFrom` instead. An audit record that must be replayable to a direction wants the first; a client that only ever reads wants neither.

## Unrelated: #3594's `gates (api)` failure is fixed

It was ASVS V8.1 matching #3594's own KDoc — the two comment lines explaining why RFC 9562 v8 was chosen *over* v5-because-v5-is-SHA-1. `check-asvs-l3.py` scanned raw lines with no comment handling. Fixed in #3608 (merged), so a rebase should clear it. Not a reason to prefer #3741.

JiRaska:
Closing as superseded by #3741, which fixes the same defect and also `Closes #3374`.

Not a quality judgement on the diff — this one is arguably the more thorough of the two. The deciding factors:

- **#3741 is CLEAN; this is BLOCKED** on four required checks (`Validate manifests`, `all-green`, `gates (api)`, `verification-metadata`).
- **#3741 carries the threat model.** fx-service is in `rules.yaml: money_path_services`, so ADR-0030 requires `docs/threat-models/openbank-fx-service.md` to be updated; #3741 does, this PR does not, so it would need that added before it could merge anyway.
- Both bump the spec `1.5.0 -> 1.6.0`, so they cannot both land regardless — whichever went second would have to re-bump.

**The designs are contradictory, not duplicates**, and that is worth recording for anyone reading #3374 later. This PR gives a derived quote its own deterministic id (`derivedId(sourceId, "inverse")`, stable across calls) and keeps `id` non-null. #3741 nulls the response `id` for a derived quote and makes the field nullable. I checked whether the nullable form breaks a consumer: it does not — customer-edge's `fxBidRate` reads bid/ask only, and every admin-ui `.id` reference is scheduling code, so nothing today reads the rate id.

**What is being carried over.** The conversion tests here pin a money-path invariant that #3741 gets right by construction but never asserts: `fx_conversions.rate_id` is `NOT NULL REFERENCES fx_rates(id)`, so if a derived id ever reaches that column, every CZK→foreign conversion fails on a foreign-key violation. That belongs on `main` independent of which identity design wins, and I am porting it in a separate PR.

The 267-line `FxRateApiContractTest` is not being ported as-is: it asserts this PR's spec shape (`id` non-null + `derivedFrom`), which is exactly what #3741 changes. If the identity design is revisited, this branch is the reference.

Branch `fix/fx-inverse-rate-identity` is not being deleted — its worktree still exists locally (last commit 2026-08-03, clean).

PR #3600, opened 2026-08-03
fix(admin-ui): stop single-use refresh tokens logging active operators out

### LINKED ISSUE #3619: admin-ui: active operators logged out every ~5 minutes — single-use refresh tokens are spent by concurrent auth() callers
## Symptom

An operator working continuously in the admin portal is logged out after a few minutes and bounced to `/auth/login?error=SessionExpired`. Activity in the tab does not prevent it. The interval matches the realm's `accessTokenLifespan` (300s).

## Cause

Access-token refresh is implemented (`refreshAccessToken` in `openbank-admin-ui/src/lib/auth/authOptions.ts`), but it cannot survive the realm's rotation policy. `openbank-infra/gitops/components/keycloak/realm-template.json` sets `revokeRefreshToken: true` and `refreshTokenMaxReuse: 0`, so each refresh token is strictly single-use.

`auth()` is called independently by `middleware.ts`, ~30 route handlers, and `/api/gate`. Each decodes the same session cookie and, before the fix, ran its own refresh grant. Two consequences:

1. **Stampede.** Parallel server-side `auth()` calls past expiry all present the same refresh token. One grant wins; the rest get `invalid_grant`, set `error: "RefreshAccessTokenError"`, and the middleware redirects to the login page.
2. **Rotation loss.** Only a response that reaches the browser can persist the rotated token. `/api/gate` is an nginx `auth_request` subrequest — nginx discards its `Set-Cookie` — and `auth()` inside a Server Component cannot set cookies at all. Such a call spends the refresh token and discards the replacement, leaving the browser holding a revoked one, so the next refresh is guaranteed to fail.

A secondary contributor: refresh fired strictly after expiry with no skew, so a token still valid when the JWT callback ran could be expired by the time an upstream service validated it.

## Why no test caught it

`src/test/authoptions-refresh.test.ts` exercised a single refresh in isolation against a mock that would answer any number of grants. Neither the single-use property of the refresh token nor concurrent `auth()` callers were modelled, so the suite was green about a flow that could not work against the real realm.

## Fix

PR #3600 — cache the rotation keyed by the refresh token it spent, so concurrent callers share one grant and a caller still carrying the pre-rotation cookie receives the already-rotated token. Failed grants are not cached. Refresh moves 60s ahead of expiry.

JiRaska:
## Problem

Operators are logged out of the admin portal a few minutes into an **active** session. Five minutes is exactly `accessTokenLifespan` in the realm.

Refresh *is* implemented (`refreshAccessToken` in `src/lib/auth/authOptions.ts`), but the realm sets `revokeRefreshToken: true` and `refreshTokenMaxReuse: 0` (`openbank-infra/gitops/components/keycloak/realm-template.json`), making every refresh token strictly single-use. `auth()` is called independently by `middleware.ts`, ~30 route handlers, and `/api/gate` — each decodes the same cookie and each ran its own grant.

- **Stampede.** A dashboard page fires several server-side `auth()` calls at once. Past expiry they all present the same refresh token. One wins; the rest get `invalid_grant`, set `error: "RefreshAccessTokenError"`, and the middleware redirects to `/auth/login?error=SessionExpired`.
- **Rotation loss.** Only a response reaching the browser can persist the rotated token. `/api/gate` is an nginx `auth_request` subrequest — nginx discards its `Set-Cookie`; `auth()` inside a Server Component cannot set cookies at all. Those calls spend the refresh token and discard the replacement, so the browser is left holding a revoked one and the *next* refresh is guaranteed to fail.

## Fix

Cache the rotation keyed by the refresh token it **spent**:

- concurrent callers share one in-flight grant (stampede);
- a caller still carrying the pre-rotation cookie gets the already-rotated token instead of presenting a revoked one (rotation loss);
- a **failed** grant is not cached, so a transient Keycloak outage stays retryable;
- refresh fires 60s **ahead** of expiry — a token valid when the JWT callback runs can still expire before the upstream service validates it;
- a successful refresh clears a stale `RefreshAccessTokenError`.

In-process map; admin-ui runs `replicas: 1`. Multi-replica would need a shared store — stated in the code.

## Verification

Five cases added to `src/test/authoptions-refresh.test.ts`, each modelling a single-use refresh token (second presentation returns `invalid_grant`). Reverting `authOptions.ts` to `main` with the new tests in place turns **four** of them red:

```
× refreshes ahead of expiry, not after it (skew)
× collapses concurrent refreshes of the same token into ONE grant
× answers a caller still holding the SPENT token with the rotated one
× clears a previous RefreshAccessTokenError once a refresh succeeds
```

With the fix: `npm test` → 79 files, 1089 tests passed. `tsc --noEmit` and `eslint` clean.

## Not changed

The realm's rotation settings stay as they are — this is an app-side bug, and a realm-template edit would not reach the running Keycloak anyway (import runs on cold start only).

## Linked issues

Closes #3619

### DIFF
--- a/openbank-admin-ui/src/lib/auth/authOptions.ts
+++ b/openbank-admin-ui/src/lib/auth/authOptions.ts
@@ -92,30 +92,95 @@ function extractRoles(jwt: string): string[] {
   return realmAccess?.roles?.filter(r => r.startsWith("ROLE_")) ?? []
 }
 
-async function refreshAccessToken(token: JWT): Promise<JWT> {
-  try {
-    const url = `${KEYCLOAK_INTERNAL_URL}/realms/${KEYCLOAK_REALM}/protocol/openid-connect/token`
-    const res = await fetch(url, {
-      method: "POST",
-      headers: { "Content-Type": "application/x-www-form-urlencoded" },
-      body: new URLSearchParams({
-        client_id: CLIENT_ID,
-        client_secret: CLIENT_SECRET,
-        grant_type: "refresh_token",
-        refresh_token: token.refreshToken as string,
-      }),
-    })
-    const refreshed = await res.json()
-    if (!res.ok) throw refreshed
+/**
+ * Refresh a minute BEFORE the access token actually expires. A token that is valid at the
+ * moment the JWT callback runs can still be expired by the time the upstream service
+ * validates it, and the realm's clock is not ours.
+ */
+const REFRESH_SKEW_MS = 60_000
+
+/**
+ * How long a completed rotation stays answerable by the refresh token it consumed.
+ * One access-token lifespan (300s) is the window in which stale copies of the cookie can
+ * still arrive; 120s covers the realistic tail without holding grants longer than needed.
+ */
+const ROTATION_CACHE_TTL_MS = 120_000
+
+type Rotation = Pick<JWT, "accessToken" | "accessTokenExpires" | "refreshToken" | "roles">
+
+/**
+ * Keyed by the refresh token that was SPENT, not by user.
+ *
+ * The realm sets `revokeRefreshToken: true` + `refreshTokenMaxReuse: 0`, so every refresh
+ * token is strictly single-use. Two things then break a session that is actively in use:
+ *
+ *  - **Stampede.** `auth()` is called independently by the middleware, ~30 route handlers and
+ *    `/api/gate` (nginx `auth_request`). Each decodes the SAME cookie, so once the access
+ *    token expires they all present the same refresh token at once. One wins; every other
+ *    gets `invalid_grant` and marks the session `RefreshAccessTokenError`, which the
+ *    middleware turns into a redirect to `/auth/login?error=SessionExpired`.
+ *  - **Rotation loss.** Only a response that reaches the browser can persist the rotated
+ *    token. `/api/gate` is an nginx auth subrequest (its `Set-Cookie` is discarded) and
+ *    `auth()` inside a Server Component cannot set cookies at all. Such a call spends the
+ *    refresh token and throws the replacement away, leaving the browser holding a revoked
+ *    one — the next refresh is then guaranteed to fail.
+ *
+ * Both are answered by remembering the result against the spent token: concurrent callers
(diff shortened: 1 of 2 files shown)


JiRaska:
## Verified end-to-end against the live sandbox Keycloak

**1. The premise, read off the running realm** (not the template — realm import is cold-start only, so the two can disagree). `GET /admin/realms/openbank`:

```
accessTokenLifespan   = 300
ssoSessionIdleTimeout = 1800
revokeRefreshToken    = True
refreshTokenMaxReuse  = 0
```

**2. The failure, reproduced.** Presenting one refresh token twice — exactly what parallel `auth()` callers do:

```
attempt 1 http=200 ok
attempt 2 http=400 invalid_grant  Maximum allowed refresh token reuse exceeded
```

**3. The fix, driven against that same Keycloak.** A throwaway harness (not committed — it needs a live realm) calls the real `jwt` callback with a real password-grant token marked expired:

- *three concurrent callers* — all three end with `error: undefined` and the **same** rotated refresh token, and that token is then accepted by Keycloak (`200`);
- *late caller still holding the spent cookie* — the nginx-subrequest case: first call spends the token and its `Set-Cookie` is discarded, second call arrives with the original token and still comes back healthy.

Reverting `authOptions.ts` to `main` with that same harness turns both red against the live realm:

```
× three concurrent auth() callers all end up with a working token
× a LATE caller still holding the spent cookie is not logged out
AssertionError: expected 'RefreshAccessTokenError' to be undefined
```

So the bug and the fix are both demonstrated against the real Keycloak, not only against a mock.

**4. Review.** Adversarial pass over the diff — cache growth and eviction, prune-vs-in-flight race, cross-user key collision, `error: undefined` serialization, unhandled rejection on the shared promise, skew vs `maxAge`/`updateAge`, and the cost of holding a spent token as a key for 120s — found no defects.

`npm test`: 79 files / 1089 tests pass. `tsc --noEmit` and `eslint` clean.

The failing `agent-review` check is the #2161 gateway signature (`is_error: true`, `num_turns=1`, cost $0), not a finding about this PR; that job is non-blocking.

Issue #3485, opened 2026-08-02
OIDC client secret is stored under two conventions (17 shared / 11 per-service) with nothing saying which is right


JiRaska:
## What

One secret, two storage conventions, nothing recording which is intended. Every service authenticates as the **same** realm client, so the KV key name carries no meaning — but getting it wrong is silent and takes the service down.

Measured across every gitops file referencing `OIDC_CLIENT_SECRET` on `main`:

```
17  remoteRef.key: account-service          shared entry
11  remoteRef.key: <their-own>-service      statement, campaign, sdd, sanctions, anacredit,
                                            balance, audit, mcp, tpp-registry, litellm, glitchtip
```

And the thing that makes the split arbitrary rather than meaningful — I checked the per-service ones:

```
statement  -> client-id: openbank-services
campaign   -> client-id: openbank-services
sdd        -> client-id: openbank-services
sanctions  -> client-id: openbank-services
anacredit  -> client-id: ${QUARKUS_OIDC_CLIENT_ID:openbank-services}
balance    -> client-id: openbank-services
audit      -> client-id: openbank-services
```

Same as the 17. There is no dedicated realm client behind any per-service key; all 28 entries hold, or are supposed to hold, the same `openbank-services` secret. The key name is pure storage convention.

## Why it is worth an issue

The failure is silent and expensive, and it just happened.

`openbank-delegation-service` shipped with `remoteRef.key: delegation-service`, an entry nobody had written. ESO answered `Secret does not exist`, the Secret was never created, and the pod sat in `CreateContainerConfigError` for its entire life. Roughly a dozen alerts fired, none of them naming the cause: `SLOMetricAbsent` ×3, `TargetDown`, `DeploymentNoAvailableReplicas`, `KubePodNotReady`, `KubeContainerWaiting`, `KubeDeploymentReplicasMismatch`, `RolloutStuck`, `ArgoCDAppDegraded`, `PostgresWALArchiveFailing`, and two Kyverno criticals. Fixed in #3471 by pointing at the shared entry.

A service added tomorrow gets a coin flip. Choosing the per-service form is not visibly wrong — 11 services do it — and the cost only appears at deploy, as an alert storm pointing anywhere but here.

This also lands squarely in the alert-fatigue problem: one missing KV key produced ~12 alerts, which is the ratio that trains people to ignore the board.

## What to decide

Not proposing an answer — the point is that it should be written down once and enforced, rather than inferred from whichever neighbour gets copied:

[...]

PR #3610, opened 2026-08-03
fix(infra): read the shared OIDC client secret from one KV entry, and gate it

JiRaska:
## What

One credential, one KV entry — declared in `rules.yaml`, enforced by the gate that already owns this area, and the estate converged onto it.

## The measurement

Re-derived from the **deployed** wiring on `main`, not from grep: workload `OIDC_CLIENT_SECRET` env ref → the Secret it names → the ExternalSecret that creates it → `remoteRef.key`.

```
29  remoteRef.key: account-service      the shared entry
10  remoteRef.key: <their-own>-service  anacredit, audit, balance, campaign, agent,
                                        mcp, sanctions, sdd, statement, tpp-registry
```

(39 rather than the issue's 28: counting ExternalSecret *entries* also catches pid-service, whose `secretKey` is `QUARKUS_OIDC_CREDENTIALS_SECRET` while the remote property is `OIDC_CLIENT_SECRET`. delegation-service is on the shared side since #3471. litellm/glitchtip do not read this property at all — glitchtip authenticates as its own realm client.)

## Why shared, and it is not a majority argument

The `openbank` realm defines exactly **one** confidential M2M client, `openbank-services` (`components/keycloak/realm-template.json`). There is no per-service realm client anywhere, and every one of the 10 per-service services carries `client-id: openbank-services` in its `application.yaml`. So a per-service KV entry never held a per-service credential — it held a hand-seeded **copy** of the one credential. `seed-vault-gaps.sh` made that literal: it populated `openbank/audit-service` by copying the value straight out of `account-service`.

That makes the copies cost without isolation:

- **Blast radius is a property of the credential, not the path.** Rotating `openbank-services` in Keycloak invalidates all N copies at once. Splitting the storage buys no isolation it does not already lack.
- **Rotation gets worse with copies.** One Keycloak rotation becomes N KV writes, with nothing enumerating N. The ExternalSecrets README documented rotation as a single `vault kv put openbank/<entry>` — true only under the shared convention.
- **Each copy needs a manual seed step, and that is the step that goes missing.** delegation-service shipped pointing at an entry nobody had written and sat in `CreateContainerConfigError` for its whole life behind ~12 alerts naming everything except the cause (#3471). `es-mcp-service.yaml` still carried a `NOT YET SEEDED` note — same defect, in its silent form, because `optional: true` kept the pod green while its M2M calls would have gone out unauthenticated.

**What it costs:** the ability to rotate one service's credential independently. Nothing has that today and nothing can until per-service realm clients exist. If they ever do, this rule is wrong and must key off the client-id rather than the path — `rules.yaml` says so in place.

**What it does not fix:** the shared entry's path is literally `account-service`, which reads as service-specific. It is the shared entry (the seeding runbook has said so since the Vault cutover), but the name still misleads. Renaming it to something honest needs an operator `vault kv put` plus a same-PR flip of every `remoteRef.key`; after that write it is a one-line change in `rules.yaml`. **Owner's call — not attempted here.**

## Transition risk

The repoint is a gitops-only change: no Vault write, no Keycloak action, nothing owner-gated.

- Both sides hold the same client's secret, so the delivered bytes are unchanged — except where a copy had drifted (already a stale credential) or was never seeded (mcp-service), where this is a fix.
- Env-mounted values are captured at container start, so nothing changes on a running pod until its next roll. There is no lockout window.
- The one way this could break is `openbank/account-service` not holding `OIDC_CLIENT_SECRET` — contradicted by the 29 workloads already reading it, several money-path. One command settles it if you want it settled: `vault kv get -field=OIDC_CLIENT_SECRET openbank/account-service` (do not echo the value).

Residual, stated plainly: after this, all 39 reads depend on one KV entry rather than 29. That is not a new single point of failure — it is the same one credential — but a bad write to that entry now reaches everything at once.

## The gate

Extends the existing `check-oidc-client-secret-wiring.py` with a fourth check. **No new gate id and no `gates.yaml` edit**: `oidc-client-secret-wiring` already owns this area, and a second copy of one rule is its own defect (#3009).

Scope is by property name — a workload authenticating as a *different* realm client (admin-ui, customer-edge, the WebAuthn client) stores its secret under `client-secret`/`kc-client-secret` and is correctly out of scope.

Falsified in both directions **against the real tree**, not only the synthetic self-test:

| probe | result |
|---|---|
| one manifest reverted to a per-service key, in place | exactly 1 `::error`, naming that file; exit 1 |
| restored from a `cp` backup | clean; exit 0 |
| `rules.yaml` exemption added for that entry | suppressed, prints a `::notice`; exit 0 |
| exemption left behind once the entry conforms again | reported as stale; exit 1 (shrink-only) |
| shared key fed to the checker synthetically | not flagged — the check cannot fail conforming manifests |

`--selftest` also refuses to pass on an empty scan, so a broken probe cannot report the estate clean.

`rules-opa-data.yaml` is unchanged — no `.rego` reads the new key, so no OPA bundle restamps.

## Linked issues

Refs #3485, #3471

The issue stays **open**: the KV-path rename is an owner action, and the acceptance criterion "migrated or explicitly exempted" is met for all 10 while the naming follow-up is not.

## Test plan

- `python3 .github/scripts/check-oidc-client-secret-wiring.py --selftest` → OK
- `python3 .github/scripts/check-oidc-client-secret-wiring.py --enforce` → clean, 39 reads against `account-service`
- `python3 .github/scripts/run-gates.py --group gitops` → the netpol drift gate passes on a committed tree; `loki-rule-load-test` fails locally on `BASE_REF: unbound variable` (CI-supplied), unrelated
- `bash -n openbank-infra/scripts/seed-vault-gaps.sh`

### DIFF
--- a/.github/scripts/check-oidc-client-secret-wiring.py
+++ b/.github/scripts/check-oidc-client-secret-wiring.py
@@ -19,16 +19,42 @@ could not authenticate. Nothing had logged it, because the path had never execut
 `fx_conversions` is empty, so it would have failed on the FIRST conversion anyone performed. Same
 shape as the ČNB feed 404 in #2204 — correct-looking until exercised.
 
-WHAT IT CHECKS — all three from committed artifacts, so it needs no cluster access:
+WHAT IT CHECKS — all four from committed artifacts, so it needs no cluster access:
 
   1. A service whose `src/main` contains `OidcClientRequestReactiveFilter` (or the blocking
      variant) MUST have an `OIDC_CLIENT_SECRET` env ref in its Deployment.
   2. That ref must be `optional: false`. An authorization-bearing credential that is allowed to be
      missing does not degrade gracefully; it sends unauthenticated requests.
   3. The Secret it names must be created by an `ExternalSecret` in gitops.
+  4. That ExternalSecret must read the ONE agreed KV entry (`rules.yaml:
+     oidc_client_secret_storage.shared_kv_entry`), not a per-service one. See below.
 
 And the converse, which is what makes it a shape rather than a checklist: a service with NO
 OIDC-filtered client should not carry the env ref at all.
+
+Check 4 — one credential, one KV entry (#3485)
+---------------------------------------------
+Checks 1-3 answer "is this secret delivered?" and cannot answer "delivered from WHERE?". The
+estate had split into two storage conventions for one credential — measured on main 2026-08-02,
+29 ExternalSecret entries read `account-service` and 10 read `<their-own>-service` — with nothing
+recording which was intended. The split is arbitrary rather than meaningful because the `openbank`
+realm defines exactly one confidential M2M client, `openbank-services`: a per-service KV entry
+holds a COPY of the one credential, seeded by copying the shared entry
+(`openbank-infra/scripts/seed-vault-gaps.sh` does that verbatim for audit-service).
+
+That makes the copies pure cost. Rotation is per-CLIENT, so one Keycloak rotation must fan out to
+every copy, with nothing enumerating them; and each copy needs a seed step that only a human
+performs. That step is the one that gets skipped — `openbank-delegation-service` shipped pointing
+at an entry nobody had written, ESO answered `Secret does not exist`, and the pod sat in
+`CreateContainerConfigError` for its entire life behind ~12 alerts that named everything except
+the cause (#3471).
+
+Scope: an entry is IN scope when its `secretKey` or its `remoteRef.property` is
+`OIDC_CLIENT_SECRET` — the name the shared client's secret is stored under everywhere. Workloads
+authenticating as a DIFFERENT realm client (admin-ui, customer-edge, the WebAuthn client) store
+theirs under `client-secret`/`kc-client-secret` and are correctly per-entry; they are out of
+scope, and that is the distinction the property name carries. If per-service realm clients are
+ever introduced this rule becomes wrong and must key off the client-id instead of the path.
 
 Usage:  check-oidc-client-secret-wiring.py [--enforce] [--selftest]
 Advisory by default (prints ::warning, exits 0) per the repo convention; --enforce fails the build.
@@ -44,6 +70,7 @@ import yaml
 
 REPO = pathlib.Path(__file__).resolve().parents[2]
 GITOPS = REPO / "openbank-infra" / "gitops"
+RULES = REPO / "openbank-libs" / "governance" / "rules.yaml"
 FILTER_MARKERS = ("OidcClientRequestReactiveFilter", "OidcClientRequestFilter")
 ENV_NAME = "OIDC_CLIENT_SECRET"
 
@@ -103,6 +130,47 @@ def provisioned_secrets() -> set[tuple[str, str]]:
     return out
 
 
+def storage_policy() -> tuple[str, dict[str, str]]:
+    """(shared KV entry, {externalsecret-name: reason}) from rules.yaml — never a second copy."""
+    block = (yaml.safe_load(RULES.read_text(encoding="utf-8")) or {}).get(
+        "oidc_client_secret_storage",
+    ) or {}
+    shared = block.get("shared_kv_entry")
+    if not shared:
+        raise SystemExit(
+            "rules.yaml: oidc_client_secret_storage.shared_kv_entry is missing — the checker has "
+            "no convention to enforce and would silently pass everything.",
+        )
+    return shared, dict(block.get("exemptions") or {})
+
+
+def shared_client_secret_entries() -> list[tuple[str, str, str, str, pathlib.Path]]:
+    """[(namespace, externalsecret, secretKey, remoteRef.key, manifest)] delivering the shared
+    openbank-services client secret. In scope iff the secretKey or the remote property is
+    OIDC_CLIENT_SECRET — a different realm client stores its secret under a different name."""
+    out: list[tuple[str, str, str, str, pathlib.Path]] = []
+    for path in GITOPS.rglob("*.yaml"):
+        try:
+            docs = list(yaml.safe_load_all(path.read_text(encoding="utf-8")))
+        except (yaml.YAMLError, UnicodeDecodeError):
+            continue
+        for doc in docs:
+            if not isinstance(doc, dict) or doc.get("kind") != "ExternalSecret":
+                continue
+            meta = doc.get("metadata") or {}
+            for entry in (doc.get("spec") or {}).get("data") or []:
+                if not isinstance(entry, dict):
+                    continue
+                remote = entry.get("remoteRef") or {}
+                if ENV_NAME not in (entry.get("secretKey"), remote.get("property")):
+                    continue
+                out.append((
+                    meta.get("namespace", ""), meta.get("name", "?"),
+                    str(entry.get("secretKey")), str(remote.get("key")), path,
+                ))
+    return out
+
+
 def oidc_env_refs() -> dict[str, tuple[str, str, bool, pathlib.Path]]:
     """{workload: (namespace, secret-name, optional, manifest)} for every OIDC_CLIENT_SECRET ref."""
     refs: dict[str, tuple[str, str, bool, pathlib.Path]] = {}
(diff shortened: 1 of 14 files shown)


JiRaska:
Verifying the one caveat this PR flags as unverified, because it is the load-bearing one: *"assumed, not verified: that `openbank/account-service` actually holds `OIDC_CLIENT_SECRET` — if that inference is wrong, this PR breaks 10 services rather than 0."*

It can be settled from the repo, without cluster or Vault access.

**All 29 existing readers read it non-optionally.** Counted across `openbank-infra/gitops/`:

```
ExternalSecrets reading account-service/OIDC_CLIENT_SECRET:
  non-optional = 29
  optional     = 0
```

Sample (`components/aml/oidc-externalsecret.yaml`):

```yaml
secretStoreRef: vault-kv
refreshInterval: 1h
data:
  - secretKey: OIDC_CLIENT_SECRET
    remoteRef:
      key: account-service
      property: OIDC_CLIENT_SECRET     # no `optional:` — required
```

External Secrets Operator fails the sync when a **required** `remoteRef.property` is absent: the target Secret is not created or refreshed, and the workload cannot mount it. So if `openbank/account-service` did not hold `OIDC_CLIENT_SECRET`, 29 ExternalSecrets across aml, billing, consent, delegation, document-service, finrep, fx, interest, kyc, ledger, lending and the rest would sit in a sync error and those services would not be serving.

That upgrades the claim from *"inferred from 29 readers"* to **structurally proven by 29 required reads on a 1-hour refresh in a running fleet**. The 10 services this PR repoints join a path that 29 others already depend on continuously — the change moves them onto the better-evidenced side, not a speculative one.

Worth noting the direction: `optional: true` on a secret ref is the pattern `openbank-infra/CLAUDE.md` warns about, because it hides a missing secret behind a healthy-looking pod — and this PR found exactly that shape in `es-mcp-service.yaml`, whose `NOT YET SEEDED` header was masked by `optional: true`. Here the absence of `optional` is what makes the inference sound.

**What is still not proven**, and I am not claiming otherwise:

- Whether any per-service copy had **drifted** from the shared value. A drifted copy is a stale credential either way, so convergence corrects it, but nobody can show none was deliberately different without reading both.
- That the 29 are currently healthy. I am reasoning from "they would be loudly broken otherwise", which is strong but is not a cluster observation.

The rotation window this PR describes honestly — env-mounted secrets do not hot-reload, so between the KV write and the last pod roll every un-rolled pod presents the old secret — is unchanged by any of the above, and is the thing to sequence money-path-last.

JiRaska:
**Review: 2 findings, none blocking.** Automated agent review — not an approval. PR is `DIRTY`; reviewed as-is, no rebase attempted.

What it does: consolidates ~39 `ExternalSecret` entries that deliver the `openbank-services` client secret onto a single KV entry (`account-service`), records the convention in `rules.yaml: oidc_client_secret_storage`, and extends the enforced `oidc-client-secret-wiring` gate with a fourth check that the KV coordinate matches.

The security argument holds and is worth stating plainly for the reviewer: **this does not widen blast radius.** There is exactly one confidential M2M realm client, so the per-service KV entries were copies of one credential, not per-service credentials. Rotation is per-client, so N copies made rotation an N-way fan-out with nothing enumerating N, and each copy needed a manual seed step — the step that was skipped for delegation-service (#3471) and is still skipped for mcp-service. Consolidating removes cost without removing isolation that never existed.

Findings:

1. **`es-mcp-service.yaml` is repointed to the shared entry, but `openbank-infra/gitops/components/mcp/mcp-service.yaml:109` keeps `optional: true` — and the PR's own text removes the justification for it.** The manifest comment says optional "so the timing gap before that key is seeded doesn't CrashLoop the pod"; after this PR there is no gap, because the shared entry is the one every other service already reads. mcp-service does carry the OIDC filter (`ConsentValidateClient.kt`, `DownstreamClients.kt`), so this is the gate's own check-2 condition: a credential allowed to be missing means the pod stays green while its downstream calls to account/balance/transaction/consent-service go out unauthenticated. The same argument applies to all eight entries in `OPTIONAL_TRUE_PENDING_LIVE_CHECK` (`agent-`, `anacredit-`, `billing-`, `document-`, `interest-`, `mcp-`, `sdd-`, `tpp-registry-service`) — every one is justified as "secret is declared; flip after confirming it synced in `<ns>`", and after this PR they all read one entry whose sync state is a single fact. Emptying that list is a one-PR follow-up and arguably belongs here.

2. **`shared_kv_entry: account-service` bakes an actively misleading name into an enforced rule.** `rules.yaml` is candid that the path is historical and not account-service's own, but the gate now *requires* every ExternalSecret to name it, so the misnomer becomes the enforced convention rather than an artefact. The comment already sketches the fix (operator `vault kv put openbank/openbank-services-oidc`, then a same-PR flip of every `remoteRef.key` and this one line). Worth an issue so it does not calcify — a name that has to be explained in prose is a name the next reader will get wrong.

Verified clean: gate is `mode: enforced` **and** invoked with `--enforce` (the load-bearing half); `selftest:` is a separate field rather than chained into `run:`, with the reasoning documented; the self-test falsifies in both directions on synthetic input (a per-service key flags, the shared key does not, a stale exemption flags) and refuses to pass when any input scan comes back empty — the "broken probe reports clean" case. Scope is keyed on `secretKey`/`remoteRef.property == OIDC_CLIENT_SECRET`, which correctly excludes workloads authenticating as a *different* realm client (admin-ui, customer-edge, WebAuthn) that store theirs under `client-secret`/`kc-client-secret`; the docstring states outright that the rule becomes wrong if per-service realm clients are ever introduced, which is the right caveat in the right place. `exemptions` is empty and shrink-only with a stale-entry ratchet, so it cannot quietly become permanent. `storage_policy()` raises rather than passing vacuously when `shared_kv_entry` is absent.

One cross-PR observation for the reviewer: read this alongside #3812, which measures that a caller holding the shared service-account token reaches 68 declared write actions including `balance.debit`, `ledger.create` and `sepaPayment.create`. This PR establishes there is one KV entry behind that identity. Neither PR is wrong; together they name the single-credential-fleet-wide-money-path-write exposure precisely, and that is the thing to decide on.

Needs human review regardless.

JiRaska:
Left this branch **untouched** while refreshing the rest of the queue: its conflict is a collision of decisions, not of text, and resolving it is not mine to do.

**What collided.** `main` gained #3485 while this branch sat open. Both write the same rule, independently:

| | this branch | `main` today |
|---|---|---|
| `rules.yaml` key | `oidc_client_secret_storage` | `oidc_secret_convention` |
| script | `check-oidc-client-secret-wiring.py` | `check-oidc-secret-convention.py` |
| gate | — | `oidc-secret-convention`, **enforced** |
| conclusion | one shared KV entry, `account-service` | identical |

Same decision, same rationale, even the same prior art (delegation-service, #3471). Merging as-is would put two rules and two scripts checking one property into `Validate manifests` — the #3009 shape, where CI cannot see the duplication because both are green.

**The substantive difference is what each does about the 10 offenders.** `main`'s gate deliberately FREEZES them and says why:

> Changing a live ExternalSecret's remoteRef re-projects the Secret and rolls the pod. Nothing in this repo can see what those 10 KV entries hold, so the switch is safe only if `account-service` is current — likely (28 live consumers) but not verified from here, and being wrong breaks authentication on 10 services at once. […] Migration is tracked on #3485.

This branch performs exactly that migration. Measured, not inferred — the sets are identical:

```
main baselined : 10
#3610 migrates : 11   (the 10, plus external-secrets/README.md)
overlap        : 10
baselined but NOT migrated: none
```

**So there is also a mechanical consequence.** `main`'s script fails on a stale exemption ("a migration done later must delete its baseline entry or the gate fails"). Merging this branch without emptying that `BASELINE` dict turns the enforced `oidc-secret-convention` gate red on `main` — all ten entries at once.

**What it needs, in order:**

1. Someone who can read the KV to confirm `openbank/account-service` currently holds the live `openbank-services` secret. That is the one fact this repo cannot establish, and the whole risk argument rests on it.
2. A decision on which rule survives — I would keep `main`'s (it is merged, gated and enforced) and drop this branch's duplicate key + script.
3. If the migration proceeds, empty `BASELINE` in the same PR.

Happy to reshape it that way on a word — the rebase itself is trivial, the single conflict is this one `rules.yaml` hunk. What I will not do is pick (1) for you.

JiRaska:
Superseded by #4296, and the merge conflict is what surfaced why.

While this branch sat, `main` gained `oidc_secret_convention` in `rules.yaml` plus
`check-oidc-secret-convention.py` — an independent implementation of the same rule, landed with the
ten manifests **baselined rather than migrated**. So this branch's own `oidc_client_secret_storage`
block and `check-oidc-client-secret-wiring.py` are now a second copy of a rule that already exists,
which is exactly the conflict `git` reported. Its `rules.yaml` block also declares
`wired_into: .github/gates/gates.yaml`, and this branch never touched `gates.yaml`.

The manifest half was still the missing piece, and it is preserved verbatim in #4296 — the ten
ExternalSecrets, `seed-vault-gaps.sh`, and the comment corrections (retargeted at the rule key that
actually landed).

#4296 also settles the question this branch could not: #3485 froze the baseline because the repo
cannot see what the KV entries hold. Measured read-only against the live store, all eleven entries
hold a byte-identical value, so the migration cannot change any projected Secret — and the baseline
is emptied rather than frozen.

Closing rather than rebasing: resolving the conflict would mean deleting most of this branch's own
diff, and the remainder is already on a clean base.

PR #3603, opened 2026-08-03
chore(agent): drop the GROQ_API_KEY rollout remnant, record full egress enforcement

JiRaska:
## Summary

Completes the LiteLLM gateway egress enforcement (#1919): the repointed agent-service image is live and verified against the running Deployment (`AGENT_MODEL_ENDPOINT=http://litellm.ai-platform.svc:4000/v1`, image `sandbox-b52f989b`), and the #2210 egress NetworkPolicy is in place — so the direct-to-provider `GROQ_API_KEY` path is dead config, removed here together with the `model_gateway_as_built` truth update (`egress_enforced: partial → full`).

## Type of change

- [x] chore (build / tooling — gitops + governance registry)

## Linked issues

Closes #1919
Refs #3599 (ADR-0175 D3 — sensitive-data routing / EU inference tier, the remaining open residency work)

## Changes

- `components/agent/agent-service.yaml` — `GROQ_API_KEY` env removed (marked "ROLLOUT WINDOW ONLY"; the window is over — verified live, see below).
- `components/external-secrets/es-agent-service.yaml` — matching `GROQ_API_KEY` secretKey entry removed; provider key now exists only in the ai-platform gateway pod.
- `openbank-libs/governance/agents.yaml` — `model_gateway_as_built`: agent-service promoted `not_routed → routed`, `egress_enforced: partial → full`, verification date → 2026-08-02. **Live evidence:** `kubectl get deploy agent-service -n platform` shows `AGENT_MODEL_ENDPOINT=http://litellm.ai-platform.svc:4000/v1` on image `sandbox-b52f989b` and `networkpolicy/agent-service-egress-allow-list` present in the platform namespace.
- `docs/compliance/eu-ai-act.md` — regenerated via `gen-eu-ai-act.py` (drift-gated in the required `Validate manifests` job); the "Partially closed gap" warning is gone, egress now reads `full`.
- 39 OPA bundle ConfigMaps + pod-roll checksum annotations restamped (`agents.yaml` is embedded policy data — the standard fleet restamp, same as #2212).
- `apps/agent.yaml`, `networkpolicy-agent-egress.yaml` — stale comments updated to the verified state.

## Definition of Done

- [x] Hexagonal layering preserved — no application code changed.
- [x] Unit tests — n/a (gitops + governance data).
- [x] `./gradlew detekt ktlintCheck koverVerify build` — n/a (no Kotlin change).
- [x] OpenAPI / Flyway / Avro — n/a.
[...]

Issue #3599, opened 2026-08-03
ADR-0175 D3: sensitive-data routing at the LiteLLM choke point (EU inference tier)


JiRaska:
Follow-up from #1919 (closed once the egress enforcement is recorded as `full`).

**State:** the LiteLLM gateway is deployed, every LLM caller is routed through it, and egress is network-enforced (`model_gateway_as_built.egress_enforced: full`). But `routing: none` — the gateway serves two hosted US models (DeepInfra, Groq) and there is no self-hosted/EU tier, so prompt content of every class leaves the EU under the documented derogation (ADR-0175 §4/§5: synthetic data only).

**ADR-0175 D3 asks for:** a routing decision at the choke point — residency class 1/2 prompts (anything beyond synthetic) must NOT reach a hosted US provider. Concretely:

1. Deploy (or contract) an EU-resident inference option: self-hosted vLLM in-cluster (ADR-0031 D6 target) or an EU AI provider with DPA + regional pinning.
2. LiteLLM routing policy by residency class: sensitive prompts → EU tier only; synthetic/non-sensitive → hosted allowed.
3. Record the routing in `agents.yaml: model_gateway_as_built.routing` (today `none`) so the EU AI Act inventory (gen-eu-ai-act.py) reflects the real control.
4. Copilot (customer-facing, ADR-0089) is the first caller whose prompt class must move off hosted US once real customer data flows — today sandbox is synthetic-only.

Refs ADR-0175 (D3), ADR-0174, ADR-0031 D6, #1919.

JiRaska:
Verified the in-repo state of this while working the governance-review program (2026-08-03):

**Already implemented and falsified (no work needed):**
- The fail-closed routing hook exists in BOTH gateways: `ModelGateway.resolve()` errors on `sensitive=true` when no `SELF_HOSTED` model is registered ("sensitive request but no self-hosted model registered").
- Both copies carry a falsifying test: agent-service `ModelGatewayTest` "sensitive context with no self-hosted model fails closed"; copilot-service ditto ("…instead of using a hosted model").
- Egress is network-enforced at the LiteLLM choke point (`networkpolicy-litellm-egress.yaml`), virtual-key indirection live (ADR-0174).
- `agents.yaml: model_gateway_as_built.routing: none` already records the honest state.

[...]

PR #3300, opened 2026-08-02
fix(agent-service): route model calls only through LiteLLM gateway

JiRaska:
## Summary

Close the ADR-0031 D6/D7 GitOps gaps for platform namespace agent workloads by removing the last direct-provider fallback from `agent-service` manifests and adding an explicit egress allow-list for `copilot-service` so platform LLM callers can reach only the in-cluster LiteLLM gateway and required in-cluster dependencies.

## Type of change

- [x] fix (bug fix)
- [ ] feat (new functionality)
- [ ] refactor (no behavior change)
- [ ] perf (performance)
- [ ] docs
- [ ] chore (build / tooling)
- [ ] security (private until disclosed)
- [ ] breaking change

## Linked issues

N/A — ADR-0031 D6/D7 gap closure (gateway routing + egress lock)

## Changes

- Remove the `GROQ_API_KEY` rollout fallback from `agent-service` GitOps so the workload authenticates to LiteLLM only with its per-agent virtual key.
- Drop the now-stale precondition commentary from the existing `agent-service` egress allow-list to match the routed deployment state.
- Add a new `copilot-service` egress NetworkPolicy that allows only DNS, LiteLLM, Keycloak, customer-edge, Tempo, and Redis.

## Definition of Done

- [x] Hexagonal layering preserved (domain has zero framework imports).
- [ ] Unit tests added/updated.
- [ ] Integration tests added/updated (if service boundary involved).
- [ ] Contract tests updated (if public API changed).
- [ ] `./gradlew detekt ktlintCheck koverVerify build` passes.
- [x] No new lint warnings.
- [ ] OpenAPI spec regenerated (if REST API changed).
- [ ] Flyway migration added + rollback note (if DB changed).
- [ ] Avro schema versioned backward-compatibly (if event changed).
- [ ] Docs updated (README, ADR if architectural).

## Security checklist

- [x] No secrets, tokens, passwords, or PII in code, config, tests, logs.
- [x] No new `@SuppressWarnings`, `as Any`, `@Suppress("...")` without
      justification (state it in this PR).
- [x] No new third-party dependency, OR: dependency review attached
      (license, CVE, source).
- [x] Auth / crypto / payment code changed? → tag `security-review-required`.
- [ ] PII path changed? → tag `gdpr-review-required`.
- [ ] Cardholder data path changed? → tag `pci-review-required`.

## Compliance impact

- [x] DORA
- [ ] PCI DSS
- [ ] GDPR
- [ ] PSD2 / SCA / RTS
[...]

PR #3612, opened 2026-08-03
chore(governance): record agent-service as routed and egress as fully enforced

### LINKED ISSUE #1919: Deploy the LLM inference gateway (ADR-0174) — agents call a US provider directly, egress unenforced (ADR-0175)

JiRaska:
## Summary

Closes the last open item of #1919: `model_gateway_as_built` still claimed agent-service was `not_routed` and `egress_enforced: partial`. Verified against the **live cluster 2026-08-03**: the running Deployment (image `sandbox-b52f989b`) resolves `AGENT_MODEL_ENDPOINT=http://litellm.ai-platform.svc:4000/v1` (direct GROQ path removed by #3300) and `networkpolicy/agent-service-egress-allow-list` (#2210) is present in the platform namespace — the LiteLLM choke point is complete, so the registry now says so (`routed`, `egress_enforced: full`) and the EU AI Act inventory drops its "partially closed gap" warning.

Supersedes #3603 (closed; the GROQ removal it carried landed separately in #3300).

## Type of change

- [x] chore (governance registry + generated artifacts)

## Linked issues

Closes #1919
Refs #3599 (ADR-0175 D3 — sensitive-data routing / EU inference tier, the remaining residency work)

## Changes

- `agents.yaml` — `model_gateway_as_built`: agent-service `not_routed → routed`, `egress_enforced: partial → full`, verification date → 2026-08-03 (evidence in comments).
- `docs/compliance/eu-ai-act.md` — regenerated via `gen-eu-ai-act.py` (drift-gated in `Validate manifests`).
- OPA bundle ConfigMaps + pod-roll checksum annotations restamped (`agents.yaml` is embedded policy data; standard fleet restamp).
- `apps/agent.yaml` — stale `GROQ_API_KEY` comment corrected (that secret entry no longer exists).

## Definition of Done

- [x] No application code changed (registry + generated artifacts only).
- [x] `opa test` + `build-bundle.sh` pass locally (15 charters, all drift assertions green).
- [x] Docs updated — eu-ai-act.md regenerated from the registry of record.

## Security checklist

- [x] No secrets, tokens, passwords, or PII — none added; the change records removal of a key path.
- [ ] Auth / crypto / payment code — not changed.

## Compliance impact

- [x] GDPR / EU AI Act — positive: the AI Act inventory (Art. 10 posture) now records the LLM egress choke point as fully enforced instead of partially; hosted-US residency derogation itself is unchanged and tracked as #3599.

### DIFF
--- a/docs/compliance/eu-ai-act.md
+++ b/docs/compliance/eu-ai-act.md
@@ -89,25 +89,17 @@ not deployed.
 | Property | As-built value | AI Act / GDPR bearing |
 |---|---|---|
 | Gateway / egress choke point | `litellm` | A single in-cluster choke point exists — the place where residency, redaction and an egress NetworkPolicy CAN be enforced. Whether every caller is actually forced through it is the `Egress enforced` row, not this one. |
-| Egress enforced | `partial` | The gateway is enforced for its own egress, but at least one caller's pod can still reach a provider directly — the choke point is bypassable. |
-| Callers routed through the gateway | `copilot-service`, `control-liveness-sentinel`, `devops-agent`, `docs-truth-agent`, `finops-agent`, `flaky-test-hunter`, `governance-auditor`, `release-steward`, `authz-policy-auditor` | Prompt content from these systems has a single auditable egress path. |
-| Callers NOT routed | `agent-service` | Prompt content from these systems reaches a provider directly. |
+| Egress enforced | `full` | Every LLM caller is network-forced through the gateway; the provider is unreachable directly. |
+| Callers routed through the gateway | `copilot-service`, `agent-service`, `control-liveness-sentinel`, `devops-agent`, `docs-truth-agent`, `finops-agent`, `flaky-test-hunter`, `governance-auditor`, `release-steward`, `authz-policy-auditor` | Prompt content from these systems has a single auditable egress path. |
+| Callers NOT routed | none | Prompt content from these systems reaches a provider directly. |
 | Hosted (external) provider(s) | `deepinfra`, `groq` | Prompt content leaves the platform trust boundary — Art. 10 data governance applies. |
 | Sensitive-data routing | `none` | No routing separates sensitive from non-sensitive prompt data. |
 | Budgets enforced at | `charter` | Cost control only — not a data-governance control. |
 | Fallback | `none` | Single provider; on failure the agent degrades to a deterministic path. |
 
-> ⚠ **Partially closed gap (Art. 10 / GDPR).** A hosted external provider is in use
-> behind an in-cluster gateway, so the provider credential and the egress path are
-> centralised and auditable — but the gateway is not yet the *only* way out. Until
-> `egress_enforced` reads `full`, a caller pod can still reach the provider directly
-> and the choke point is a convention, not a control. Sensitive-data routing remains
-> absent either way. Residency, DPA and the synthetic-data licence position are
-> recorded in **ADR-0175**.
-
 ## Provenance
 
-- Source: `openbank-libs/governance/agents.yaml` (sha256 `7feff1674a83d5ea…`, 15 charters)
+- Source: `openbank-libs/governance/agents.yaml` (sha256 `cc7af1dd6eed558d…`, 15 charters)
 - Source: `openbank-libs/governance/ml-systems.yaml` (sha256 `9cd5cb5b20918520…`, 4 non-agent systems)
 - Related: ADR-0031 (agent governance), ADR-0084 (fraud scoring plane), ADR-0139/0140
   (ML decisioning platform), ADR-0141 (model registry), ADR-0142 (credit decisioning),
--- a/openbank-admin-ui/ai-governance-snapshot.json
+++ b/openbank-admin-ui/ai-governance-snapshot.json
@@ -142,7 +142,7 @@
   "facts": {
     "agentCharters": {
       "source": "openbank-libs/governance/agents.yaml",
-      "sha256": "7feff1674a83d5ea",
+      "sha256": "cc7af1dd6eed558d",
       "count": 15,
       "ids": [
         "compliance-officer",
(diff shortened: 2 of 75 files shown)


JiRaska:
**Review: 1 finding, none blocking.** Automated agent review — not an approval.

What it does: flips `model_gateway_as_built` in `agents.yaml` — `agent-service` moves `not_routed` -> `routed`, and `egress_enforced` moves `partial` -> `full` — and restamps every service's OPA bundle plus the derived `eu-ai-act.md` and `ai-governance-snapshot.json`.

I verified both claims against the tree rather than taking the prose:

- **`routed`**: `openbank-agent-service` resolves `AGENT_MODEL_ENDPOINT`, and `agent-service.yaml` points it at `http://litellm.ai-platform.svc:4000/v1` with the comment naming the litellm-config model mapping. Consistent.
- **`egress_enforced: full`**: `openbank-infra/gitops/components/agent/networkpolicy-agent-egress.yaml` exists and declares `policyTypes: [Egress]`, which is the half that was missing when the value was `partial` (the component's `network-policies.yaml` is still `policyTypes: [Ingress]` only, and the separate file is deliberately named so it is not overwritten). The claim is supported.

Finding:

1. **`egress_enforced: full` is a compliance claim resting on one NetworkPolicy file that no gate asserts the existence of, while the code-level default still points at the provider directly.** `openbank-agent-service/src/main/resources/application.yaml:194` reads `${AGENT_MODEL_ENDPOINT:https://api.groq.com/openai/v1}` and `:217` reads `${AGENT_MODEL_API_KEY:${GROQ_API_KEY:}}` — so "the gateway cannot be bypassed" is a property of `networkpolicy-agent-egress.yaml`, not of the workload. Grepping `.github/` for `networkpolicy-agent-egress` or an agent egress allow-list returns nothing: delete that file and CI stays green while `agents.yaml` keeps asserting `full`, `docs/compliance/eu-ai-act.md` keeps publishing it, and the admin-UI snapshot keeps rendering it. That is the #2280 shape — a published claim about a distributed artifact with prose instead of a gate. The narrow fix is a check that every namespace hosting a charter-registered agent has an `Egress` NetworkPolicy selecting it, keyed off `model_gateway_as_built.routed` so its coverage set is derived rather than hand-kept. (Flipping the application.yaml default to the gateway URL, or to empty, would also be honest — the `GROQ_API_KEY` fallback is documented as a local-dev convenience, and local dev is exactly where a default belongs least when the production posture is "no agent holds a provider key".)

Also note the value is dated "verified live 2026-08-03" against a specific image (`sandbox-b52f989b`). That is the right way to write an as-built claim, and it is also a claim with a shelf life — it is a statement about a Deployment, and a Deployment moves.

Verified clean: I regenerated `rules-opa-data.yaml` via `gen-rules-opa-data.py` and re-ran the `gen-*opa-bundle*.sh` scripts for ledger, party, accounts, agent and mcp against this branch — **zero diff**, so the bundles are not stale relative to the governance data they embed, and the pod-roll checksum annotations move with them. That is the failure this PR could most easily have had (edit the data, ship the old policy) and it does not. `agent.yaml`'s ArgoCD comment was updated from `GROQ_API_KEY` to the LiteLLM virtual key, so the manifest prose no longer describes a retired path. The two derived artifacts (`eu-ai-act.md`, `ai-governance-snapshot.json`) are both in the diff rather than left to drift. `mergeStateStatus: CLEAN`.

Needs human review regardless — it changes a published EU AI Act conformance claim.

JiRaska:
Closing in favour of #4078, which carries the same three-line content change redone against current
main.

Not abandoned — the reason is cost, not merit. This PR is 73 conflicts and 314 commits behind,
because 72 of its 75 files are regenerated OPA bundles. Since #3357 the bundles embed
`rules-opa-data.yaml` instead of hashing the governance sources, so a governance-only edit no longer
restamps them: redone today the same change is **three files and no conflicts**. Merging this one
would have been hours of resolving generated output that the current generators produce for free.

The substance was re-verified against the live cluster today rather than inherited from this PR's
2026-08-03 reading, because the registry's own note required exactly that ("Promote it to `routed`
only after re-reading the live Deployment — merged is not deployed"):

```
AGENT_MODEL_ENDPOINT = http://litellm.ai-platform.svc:4000/v1
networkpolicy/agent-service-egress-allow-list   present in ns platform, age 9d
```

Both still hold, so the promotion this PR proposed is correct and #4078 makes it.

One thing this PR would have shipped silently and #4078 does not: `gen-eu-ai-act.py` had no warning
branch for `gateway + egress == full`, so flipping the flag **removed** the hosted-provider warning
from the published EU AI Act inventory altogether — including "prompt content leaves the EU" and the
ADR-0175 residency/DPA pointer, all of which remain true. #4078 adds the missing branch and
falsifies all three against their inputs.

JiRaska:
**Correction to my closing comment above — one of my two reasons was factually wrong.**

I wrote that since #3357 the bundles embed `rules-opa-data.yaml` instead of hashing the governance
sources, so a governance-only edit no longer restamps them, and that this PR's 72 regenerated bundle
files were therefore avoidable churn. That is not true. **42 of the `gen-*opa-bundle*.sh` generators
read `openbank-libs/governance/agents.yaml` directly**, so editing that file restamps essentially the
whole fleet. This PR's 75-file diff was the correct consequence of its own change.

I found out the way it should be found out: #4078 went red on
`build + verify bundle, check ConfigMap sync` with `agent-opa-bundle.yaml is stale`, and it has now
had to carry the same ~74-file restamp this PR carried.

**The close still stands**, on the reason that survives: 314 commits behind with 73 conflicts, and
those conflicts are in generated output that the generators reproduce for free. Re-doing the edit on
current main and regenerating is cheaper and safer than resolving them by hand — but it is not
"three files", and I should not have said so.

Nothing to do here; recorded so the next person does not inherit a wrong mental model of what
touches the bundles.

PR #3613, opened 2026-08-03
fix(delegation): refuse dailyLimit/monthlyLimit — ADR-0232's cumulative ceilings are enforced nowhere

### LINKED ISSUE #3800: Delegation grants accept dailyLimit/monthlyLimit that nothing enforces
## The defect

`POST /api/v1/delegations` accepts `dailyLimit` and `monthlyLimit`. **Nothing enforces them.**

The grant stores them and `DelegationOffered` publishes them, so every layer looks like it
participates — but no projection counts spend against a grant, so a grantor who sets
"max 5 000 Kč/den" has capped nothing. An API that accepts a limit it does not apply is worse than
one that rejects it: the customer believes they constrained their delegate.

## Why enforcement was not the fix

A cumulative counter needs a point where spend is **observed**, and there is none:

- `AuthorizationService.isAuthorizedForAmount` — the only amount-aware account guard — had zero
  production callers when this was found.
- No money-moving service (domestic-payment, sepa-*, swift, standing-order, settlement, clearing,
  transaction, ledger) reads a delegation grant.
- `DelegationOffered` does not carry the cumulative figures to any projection that could count
  against them.

A counter shipped into that would sit at zero and `withinDailyLimit()` would return `true` forever
— the same false promise with a table behind it, and harder to notice because it would look
implemented. Worse, a correct counter must increment in the transaction that debits the account,
and that transaction does not exist yet, so choosing its home now guarantees moving it later.

## What was done instead

Refuse the fields at both delegation-service and customer-edge with
`400 CUMULATIVE_LIMIT_UNSUPPORTED`, ahead of the SCA gate so a doomed request does not spend the
customer's single-use challenge. `perTransactionLimit` is unaffected and is the only limit that
binds. The app never rendered a cumulative-ceiling control, so no client regresses.

## Still open after that PR

- **Existing rows keep their unenforced ceilings.** Each is a grant a customer accepted under a
  false belief. Nobody has identified those grantors or told them.
- ADR-0232 D1/D6 still describes cumulative ceilings as a capability. Until they are built, the
  ADR promises something the platform does not do — the threat model now says so, the ADR does not.
- The same "field exists at every layer but the enforcing one" pattern is unaudited for D4, D5, D7
  and D8.

JiRaska:
## Option (b): reject them at the edge. **ADR-0232 promises a capability this platform does not have.**

`OfferDelegationRequest` accepted `perTransactionLimit`, `dailyLimit` and `monthlyLimit`. Only the
first is enforced. The other two travelled the whole system — request DTO → `DelegationGrant` →
`DelegationGrantEntity` → `DelegationResponse` — and were applied to nothing. A grantor could cap a
delegate at "5 000 Kč/den", get a `201` with the field echoed back, and have every payment checked
against `perTransactionLimit` alone.

This PR refuses both fields on offer (`400 CUMULATIVE_LIMIT_UNSUPPORTED`), at customer-edge and at
delegation-service, and marks them unsupported in both `openapi.yaml`s. **It does not deliver
ADR-0232 D1/D6 — it stops the platform claiming them.** The ADR gains a measured delivery-status
table saying so.

### Why not (a)

Not effort. A cumulative ceiling is only a control at the point where accumulated spend is
**observed**, and this repo has no such point. Measured against `origin/main`:

| Claim | Evidence |
| --- | --- |
| `DelegationOffered` does not carry the two fields | `DelegationService.offer` emits `perTransactionLimit` only — no projection can even learn a daily cap |
| account-service's projection has no equivalent | `DelegatedAccessGrant` has `perTransactionLimitAmount/Currency` and nothing else |
| The only amount-aware account guard has **zero** production callers | `AuthorizationService.isAuthorizedForAmount:75` — every call site is a unit test; the HTTP `/check` endpoint calls the amount-free `isAuthorized` |
| No money-moving service reads a delegation grant | zero grant/projection reads across domestic-payment, sepa-payment, sepa-instant, swift, standing-order, settlement, clearing, transaction, ledger — every `authoriz` hit is the bearer header or an `@Authorize` OPA annotation |
| customer-edge payment initiation is owner-only | `CustomerEdgeResource` `/domestic-payments` and `/sepa-payments` require debtor-owner == JWT party and pass no delegated context; a delegate gets 403 |
| The one delegate flow carrying an amount cannot execute | savings withdrawal *proposal* — `decide` requires the owner + owner SCA, and `SavingsWithdrawalApproved` has no consumer anywhere |

So a spend counter written today would have **no writer**. It would sit at zero forever while
`withinDailyLimit()` returned `true` on every call — the same lie, with a table behind it and the
appearance of a control. That is precisely the "approximate ceiling" the brief rules out.

The second reason is where the counter has to live. To be correct under concurrency it must be
incremented in the **same transaction that debits the account** — anywhere else and you get a
two-phase gap (check passes, counter increments, payment fails → overcount; payment commits, counter
write fails → undercount) that no amount of row locking in account-service's projection DB closes.
That transaction does not exist yet. Building the ceilings means wiring a delegated money path
first, then counting on it, in that order.

### What changed

**delegation-service** (`info.version` 1.1.0 → **1.2.0**)
- `DelegationService.rejectUnenforcedCeilings` refuses either field, placed **ahead of the SCA gate**
  for the same reason ownership and eligibility already are: `consumeChallenge` *spends* the
  challenge, and a request that was always going to be refused must not cost the customer their
  ceremony.
- `DelegationUnsupportedConstraintException` + mapper → 400 with a machine-readable `code`.
  400 not 422: no value of the field would be accepted, so there is nothing to retry with.
- `DelegationGrant.withinLimits` KDoc corrected. It said cumulative tracking was "the enforcing
  product service's job … the grant carries the ceilings, the resource owner counts the spend" —
  a division of labour written as though it were a fact. Nobody counts.
- Reads are deliberately unchanged: a row stored before this still rehydrates and still serializes.
  Hiding data already stored replaces one wrong belief with another; the response schema instead says
  plainly that the number is not a cap.

**customer-edge** (`info.version` 1.26.0 → **1.27.0**)
- The one exception to `offer`'s pass-through. Upstream is the binding gate; this copy exists so the
  customer channel fails on its own terms and the refusal appears in the contract the app reads.

**ADR-0232** — a delivery-status table scoped to D1's numeric constraints.

### Test evidence — falsified, not just green

Every new assertion was run against the broken code it exists to reject:

- guard removed from `DelegationService.offer` → **3 failed / 30**: `offer refuses a dailyLimit…`,
  `…a monthlyLimit and names both fields…`, `…before spending the SCA challenge`. The control
  `offer still accepts a perTransactionLimit` stayed **green** — the asymmetry is the point; a
  service that had simply stopped accepting all limits would pass the first three.
- edge check neutered to `emptyList()` → **1 failed / 14**: `offer rejects a body carrying
  dailyLimit or monthlyLimit`. Its control (`offer passes through a null ceiling and a
  perTransactionLimit`) covers the two ways to get this wrong: `node.has(field)` would reject an
  explicit JSON `null`, and a blanket limit-strip would drop `perTransactionLimit`.

Local gate green on both modules: `detekt ktlintCheck test`.

### Critical assessment of this PR

**Unverified.**
- No integration test drives the refusal through a real HTTP request against a real DB. Both
  refusals are pure input validation on the request object with no persistence involved, so a
  `@QuarkusTest` would exercise Jackson and the mapper and nothing else — but that means the
  **wiring** of `DelegationUnsupportedConstraintExceptionMapper` (that JAX-RS actually selects it,
  and that the body really carries `code`) is asserted nowhere. If the mapper is not picked up, the
  refusal still happens but surfaces as a 500. That is the weakest link here.
- `openbank-app` is a separate repo. Nothing in this PR proves the app does not send these fields
  today; if it does, this turns a silent 201 into a visible 400 for those users. That is the correct
  direction, but it is a behaviour change to a live client I could not test.
- oasdiff was not run locally (no binary). I classified both bumps as **minor** by hand: no schema
  removal, no new required field, additive `deprecated` + `description` + an error `code`. Rejecting
  previously-accepted input is behaviourally breaking even though the spec diff is not, and I chose
  not to force a major, because a major here means `/api/v2` — disproportionate to removing a field
  that never did anything.

**Where the ceiling can still be exceeded.** Everywhere — that is not a residual risk, it is the
state this PR documents. Nothing is now enforced that was not enforced before. A grant stored before
this merge still carries its `dailyLimit`, still returns it over the API, and it still caps nothing.
I did not migrate or null those rows: destroying data to make an inconsistency invisible is the
wrong lever, and the row is evidence of a grant the customer agreed to under a false belief. **Those
rows should be found and their grantors told** — follow-up work this PR does not do.

**What a reader would wrongly assume.**
1. *That per-transaction limits are enforced on real payments.* They are not, in any reachable path.
   `isAuthorizedForAmount` — the only amount-aware account guard — has zero production callers, and
   the only in-repo caller of delegation-service's amount-aware `POST /check` is the **read-only
   admin-ui console probe**. `perTransactionLimit` is enforced in code that nothing invokes. This PR
   does not fix that, and it is arguably the larger defect: it is the field I just told the API to
   keep accepting.
2. *That "delegation is enforced".* The three `GET …/check` guards (account, savings-goal, card) have
   no in-repo caller either. The delegation projection is real and correct, and is currently
   consulted only by the savings-withdrawal proposal path — which terminates in a database row,
   because `SavingsWithdrawalApproved` has no consumer.
3. *That the ADR's delivery-status table covers ADR-0232.* It covers **D1's numeric constraints
   only**. I did not audit D4, D5, D7 or D8, and the same "field exists at every layer but the
   enforcing one" pattern is exactly what I would look for there next.

Refs ADR-0232

Closes #3800


JiRaska:
## Gap 5 closed with evidence — this does not break the app

You flagged that you could not check whether `openbank-app` sends the fields this PR now rejects.
Checked against `JiRaska/openbank-app@origin/main`:

`DelegationApi`'s offer DTO carries `granteePartyId`, `resourceType`, `resourceId`, `capabilities`,
`perTransactionLimit` and `grantScaSessionId`. **No `dailyLimit`, no `monthlyLimit`.** The
`dailyLimit` hits in that repo are card spending limits (`CardSheets.kt`,
`dailyLimitMinorUnits`) — a different feature.

So the refusal cannot turn a working `201` into a `400` for any client that exists today. It only
closes the door before someone walks through it.

## The choice was right for a reason worth restating

The decisive fact is not that the counter is hard — it is that **it would have no writer**.
`isAuthorizedForAmount` has zero production callers, no money-moving service reads a delegation
grant, and `DelegationOffered` does not even carry the daily figure, so no projection could learn
it. A counter shipped today would sit at zero and `withinDailyLimit()` would return `true` forever:
the same false promise with a table behind it, and harder to notice because it would look
implemented.

## Still open, and it is your own gap 1

Nothing proves JAX-RS actually selects the new mapper. If it is not wired the refusal still
happens, but as a `500` — which reads as a broken service rather than a rejected request, and would
be found by a customer rather than by CI. That is worth an IT that drives the endpoint and asserts
`400` plus the `CUMULATIVE_LIMIT_UNSUPPORTED` code, not a unit test on the mapper class.

Issue #3432, opened 2026-08-02
auto-deploy: a workflow_dispatch reconcile can never deploy — every service is blocked by can-i-deploy for a sha the push lane never published


JiRaska:
## What happens

A `workflow_dispatch` reconcile builds every requested service, then deploys **none of them**. Not a
subset — the whole set, every time.

Measured on run `30761923908` (2026-08-02 19:03Z), dispatched with the 12 services
`auto-deploy-reconcile-lag.sh` itself reported as lagging:

```
Detect changed services  success
Build + push             success
can-i-deploy             failure
Gitops PR                success   ← green, and updated nothing
Deploy signal            success
```

`Gitops PR` logged the same line 13 times:

```
[openbank-account-service] blocked by can-i-deploy this run — skipping manifest update
[openbank-billing-service] blocked by can-i-deploy this run — skipping manifest update
… 11 more
```

The `can-i-deploy` error names the cause:

```
can-i-deploy: workflow_dispatch of 9ef6c0ab9… which has no published pact version for
openbank-billing-service — the gate cannot ask about the sha being deployed, and asking about a
different commit is not a gate (#3318)
```

## Why this is not #3318 or #2549

**#3318 is correct and should stay.** Refusing is better than silently asking the broker about a
different commit. **#2549** fixed the reconcile's own logic (#2021, #2590) and was closed this
morning; the lag detection does work — it named 12 stranded services and it was right.

The gap is the interaction. Pacts are published by each service's own **push** lane. A dispatch
builds images tagged with the current `main` sha, for which no service CI ran, so no pact version
exists for that sha — and #3318 then refuses, correctly, for every service in the set. So the
reconcile path cannot succeed by construction: the more services are stranded, the more certain it
is that none of them will be re-driven.

## Why it matters

Reconcile is the only mechanism for re-driving services that missed their push deploy. It was needed
today precisely because #3256 broke push detection for five hours (fixed in #3406) and left 12
services behind, several money-path: `billing`, `sepa-instant`, `psd2`, `transaction`,
`domestic-payment`, `sepa-payment`, `card-issuance`, `swift`, `account`, `fraud`,
`clearing-simulator`, `devops-agent`.

The failure is also quiet in the direction that matters: `Gitops PR` reports **success** while
updating nothing, so the run reads as "reconcile ran" rather than "reconcile deployed nothing".

## Directions (not a decision)

1. **Publish a pact version for the dispatched sha** before the gate runs — the gate then has
[...]

PR #3618, opened 2026-08-03
fix(fraud): move the runtime image to glibc so the ONNX model can actually load

JiRaska:
## What

`openbank-fraud-service` bundles `com.microsoft.onnxruntime` (ADR-0139 phase-1b). Its
`libonnxruntime.so` is linked against **glibc + libstdc++**, and every service image is built from
`.github/workflows/Dockerfile.deploy` on `eclipse-temurin:25-jre-alpine` — **musl**. So
`OrtEnvironment.getEnvironment()` has never once succeeded in a deployed environment, and the model
has scored exactly nothing since the adapter shipped.

This moves the fleet runtime base to glibc, and adds a check that runs the real dynamic loader over
the real `.so` inside the real base image before any image is pushed.

## Root cause, as measured (not inferred)

Live sandbox pod `fraud-service-…` at 07:35 today, `fraud-service` container:

```
java.lang.UnsatisfiedLinkError: /tmp/onnxruntime-java.../libonnxruntime.so:
  Error loading shared library libstdc++.so.6: No such file or directory
  at ai.onnxruntime.OnnxRuntime.load(OnnxRuntime.java:387)
  at ai.onnxruntime.OrtEnvironment.<clinit>(OrtEnvironment.java:34)
```

`ldd` on the shipped `.so`, on `linux/arm64`, onnxruntime 1.22.0:

| base | verdict |
|---|---|
| `eclipse-temurin:25-jre-alpine` (today's) | `libstdc++.so.6`, `libgcc_s.so.1`, `ld-linux-aarch64.so.1` all unresolved, 3099 unresolvable relocations |
| `eclipse-temurin:25-jre` (glibc) | every NEEDED entry resolves |

The cheap fix does not work: adding `libstdc++` to the musl base only moves the failure to the glibc
dynamic loader. A glibc base is required, not preferred.

## What changed

* **`.github/workflows/Dockerfile.deploy`** → `eclipse-temurin:25-jre`, digest-pinned to the
  multi-arch index. Same vendor, same JRE, only the libc changes. `uid 100 / gid 101` are reproduced
  exactly so nothing that assumed the busybox `adduser -S` numbers moves. Cost: +132MB per base
  layer (224MB → 356MB), ~$0.67/month of ECR storage across ~51 repositories.
* **`.github/scripts/verify-image-native-libs.py`** (new) — extracts every linux `.so` the image
  would ship and runs `ldd` on each one **inside the base image it reads out of Dockerfile.deploy**.
  Wired into `auto-deploy.yml`, `ghcr-publish.yml` and `build-push-service.sh`, i.e. all three
  producers. Its scope is DERIVED from the jars a service actually ships, so there is no list of
  "services with native code" to keep in sync.
* **`openbank-infra/scripts/build-push-service.sh`** — this kept a **third** copy of the runtime
  recipe: unpinned (no digest) and still musl. A laptop deploy of fraud-service would have re-shipped
  the exact defect CI had just fixed. It now copies the same file the CI producers copy.
* **`OnnxFraudModelTest`** — the existing assertion was "score is in [0,1]", which a constant would
  satisfy. It now pins the documented logistic exactly, so only a real inference passes.

## Verification

Not "the build went green" — that is precisely what happened for the whole life of this defect.

1. An image built from the new recipe, run as the `openbank` user, loads the runtime and scores the
   documented input: `ENV OK: OrtEnvironment(name=ort-java,version=1.22.0)` / `SCORE: 0.13010845`,
   matching the value the new unit assertion pins.
2. The new check was falsified in both directions against the **real** `openbank-fraud-service`
   fast-jar: red on the old alpine digest, green on the new base, on both `amd64` and `arm64`.
3. Its `--self-test` was falsified too: pointed at a musl `Dockerfile.deploy` it fails; on this
   branch it passes.
4. The new test assertion was falsified by perturbing the intercept — it goes red.

Incidental finding: `liblz4-java.so` (Kafka's LZ4 codec) is also glibc-linked and could not have
loaded on the musl base either. It resolves cleanly on the new one.

## Merging this does NOT deploy it — read before closing #3354

`auto-deploy.yml`'s `Detect changed services` only selects a service when the push touched
`<svc>/src/main`, `<svc>/build.gradle.kts`, or a shared lib that service consumes. A change to
`.github/workflows/Dockerfile.deploy` matches none of those, so this merge rebuilds **zero**
services and every pod keeps running its musl image. The base change reaches a service only on its
next rebuild for some unrelated reason, or on an explicit dispatch:

```
gh workflow run auto-deploy.yml -R JiRaska/open-bank-oss -f services=openbank-fraud-service
```

(`-f services=` is space-separated if more are wanted.) Verify afterwards that the pod's log no
longer carries `UnsatisfiedLinkError`, and only then treat #3354 as closed. Related: the sandbox
pod is currently on fraud-service **0.10.0**, so it does not yet even carry the #3376 blast-radius
fix that shipped in 0.10.1 — `GET /api/v1/fraud/review-queue` was still answering 500 at 07:35
today.

## Not in this PR

49 per-service `openbank-*/Dockerfile` files and 8 `05-operations.*.md` docs still name the alpine
base. Those files are declaration-only (#3016) — nothing builds them — but they are now stale.
Sweeping them needs a deliberate commit-type call, because `<svc>/Dockerfile` and
`src/main/resources/**` are both release-triggering paths and a releasing type there would cut ~50
patch releases for a change that touches no service code. Tracked separately.

## Linked issues

Refs #3354

## Note for the reviewer

This touches `.github/workflows/`, so the merge hook requires a human. That is expected — leaving
the PR open for review rather than working around it.

### DIFF
--- a/.github/scripts/verify-image-native-libs.py
+++ b/.github/scripts/verify-image-native-libs.py
@@ -1,0 +1,337 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) OpenBank contributors. Licensed under the Apache License, Version 2.0.
+# See LICENSE in the repository root or https://www.apache.org/licenses/LICENSE-2.0 for details.
+"""Prove every native .so a service image ships can actually link in the runtime base (#3354).
+
+WHAT WENT WRONG WITHOUT THIS
+    openbank-fraud-service bundles com.microsoft.onnxruntime (ADR-0139 phase-1b). Its
+    libonnxruntime.so is linked against glibc + libstdc++; the runtime base was
+    eclipse-temurin:25-jre-alpine (musl). OrtEnvironment.getEnvironment() therefore threw
+    UnsatisfiedLinkError on the first request in EVERY deployed environment, from the day the
+    adapter shipped, and the model scored exactly nothing.
+
+    Nothing in this repo could see it, and that is the interesting part:
+
+      * the unit tests construct a real OrtEnvironment and assert a real score — but they run on
+        a glibc CI runner, so they are green against a musl image by construction;
+      * the Gradle build, detekt, ktlint and the image build itself never execute a line of
+        native code;
+      * the pod is Ready — the health probes never touch the model;
+      * and the adapter degrades to `null` on purpose, so the only trace is one ERROR line
+        (and, before #3376, a 500 on an unrelated read endpoint).
+
+    A green build proves nothing here. The only thing that can be wrong out loud is running the
+    real dynamic loader, from the real base image, over the real .so the image ships. That is
+    what this does.
+
+HOW
+    1. Read the runtime base from .github/workflows/Dockerfile.deploy — the single source every
+       image producer copies. NEVER give this script its own copy of the base: a second copy
+       moves with the first and keeps passing about an image nobody deploys.
+    2. Scan the service's fast-jar lib/ tree for jars carrying linux natives for the target
+       architecture, and extract those .so files.
+    3. `ldd` each one INSIDE that base image, on that platform.
+    4. Fail on any unresolved dependency.
+
+    Step 2 is what keeps this honest as the fleet changes: the scope is DERIVED from the jars a
+    service actually ships, not from a hand-kept list of "services with native code". A service
+    that gains a native dependency is covered the day it gains it, and one that drops it stops
+    paying for the check with nothing to update.
+
+    Intra-bundle sonames are not findings. onnxruntime ships libonnxruntime4j_jni.so, whose
+    NEEDED entry `libonnxruntime.so.1` is resolved at runtime because the JVM System.load()s its
+    sibling from the same extracted directory first. A missing dependency whose name matches an
+    extracted sibling is therefore ignored — matched against the extracted file set, which is
+    derived, not against a literal.
+
+USAGE
+    verify-image-native-libs.py <lib-dir> [--arch arm64|amd64] [--base IMAGE]
+    verify-image-native-libs.py --self-test
+
+    <lib-dir> is the fast-jar lib/ directory staged into the build context, e.g.
+    openbank-fraud-service/build/quarkus-app/lib. Exits 0 when there is nothing to check.
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import pathlib
+import re
+import shutil
+import subprocess
+import sys
+import tempfile
+import zipfile
+
+REPO_ROOT = pathlib.Path(__file__).resolve().parents[2]
+DOCKERFILE_DEPLOY = REPO_ROOT / ".github" / "workflows" / "Dockerfile.deploy"
+
+# Jar-internal directories holding Linux natives, per Docker platform arch. onnxruntime uses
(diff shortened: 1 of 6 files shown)


JiRaska:
Verified the two operational claims in this PR, and traced one step further — because the obvious worry is that the fix lands and cannot be shipped.

**"Merging deploys nothing" — confirmed.** `auto-deploy.yml`'s `paths:` are `openbank-*/src/main/**` plus the libs build files; `.github/workflows/Dockerfile.deploy` matches none of them, so the merge triggers no deploy. The dispatch in this PR's instructions is required, not optional.

**The dispatch will succeed, and that was not obvious.** Three things interact here and only the third makes it work:

1. `pact-version-tree-equivalent.sh` (from #3590) puts `Dockerfile.deploy` **in scope** (`GITHUB_IN_SCOPE`). So the moment this merges, every service's tree hash changes and **no service is equivalent** to its last pact-published commit. The equivalence path cannot rescue anything here.
2. A `workflow_dispatch` with no published pact version for the dispatched sha hits #3318's `REFUSE` — which is exactly the fleet-wide block described in #3432 (54 of 54 services refused).
3. **But `services-ci.yml` full-fleets on this change.** Its per-module attribution has a catch-all — `*) SCOPE_OK=0 ;;` — and a `.github/workflows/…` path matches none of the explicit cases, so `SCOPE_OK=0` and the decision is *"a changed path is not cleanly attributable -> full-fleet (safety)"*. Every module rebuilds on the merge commit, so pacts **are** published for that sha, `can-i-deploy` can ask about it exactly, and the dispatch is answerable.

I had this wrong on the first pass and it is worth recording why: I read the `RELEVANT` filter, saw `.github/` was not excluded as inert, then assumed the per-module mapping would attribute it to no module and produce an empty matrix. The catch-all is what prevents that, and it is 20 lines below where I stopped reading. The comment above it promises exactly this behaviour — *"the moment a changed path is NOT cleanly attributable to a module or a pact pair, fall back to full-fleet. Never under-build the Pact graph."* — and the code honours it.

So the sequence is: merge → full-fleet build publishes pacts at the merge sha → `gh workflow run auto-deploy.yml -f services=openbank-fraud-service` → answerable. Dispatch soon after the build completes rather than hours later: #3590 measured equivalence windows of 84-192 min, and once `main` moves on, this sha stops being the one `--latest main` resolves to.

**Two things worth a reviewer's attention, neither a defect in this PR:**

- **Blast radius.** `Dockerfile.deploy` is the single recipe all three producers copy, so musl→glibc changes DNS resolution and allocator behaviour for ~51 services. This PR verified *linkage*, not runtime behaviour under load. That is the right scope for the change, but it is the reason it is a human-review PR and not a rubber stamp.
- **`liblz4-java.so` was the second landmine.** Kafka's LZ4 codec is glibc-linked too, and zstd/snappy happen to ship musl builds. The fleet was one `compression.type=lz4` away from a second failure of this class, with no relationship to fraud-service. That is the argument for the new `verify-image-native-libs.py` being worth more than the base-image change itself.

Also confirmed the shared-repo `git stash` use flagged in the report did not strand anything: `git stash list` still shows the owner's `stash@{0}` and the three older feature-branch entries, unchanged.

Issue #3624, opened 2026-08-03
39 non-money-path handlers still answer 500 for a missing required query/header parameter


JiRaska:
Follow-up to #3104, which fixed the money-path half. **39 inbound sites remain**, across 9 non-money-path services. They are baselined in `.github/scripts/check-nonnull-jaxrs-params.py` (gate `nonnull-jaxrs-param-ratchet`, enforced), so no NEW occurrence can merge — this issue is the tail, not a hole.

## The defect

JAX-RS injects `null` for an absent `@QueryParam` / `@HeaderParam`. Kotlin's null-safety is compile-time only, so a non-nullable declared type buys nothing at runtime; it only decides where the failure lands, and every landing is a 500. Full mechanism, including why a `requireNotNull` in the body of a non-suspend handler is dead code, is in the header of that script and in #3104.

The fix shape is fixed and mechanical — declare the parameter nullable, then guard it — but the **decision** is per-handler and is not:

- **truly required** → nullable + `requireNotNull(x) { "query parameter 'x' is required" }`; libs-runtime maps that to 400
- **has a sensible default** → `@DefaultValue`
- **actually optional** → nullable type, handle the absent case

A blanket rule would be wrong for the same reason it was wrong for bodies (#3044).

## What remains

| service | resource | parameters |
|---|---|---|
| openbank-psd2-service | `AisResource`, `PisResource` | `Consent-ID` (×7), `Idempotency-Key` (×4) |
| openbank-card-issuance-service | `CardResource`, `CardDelegationResource` | `X-Operator-Id` (×7), `Idempotency-Key`, `partyId`, `intent` |
| openbank-customer-edge | `CustomerEdgeResource` | `accountId` (×4), `currency` |
| openbank-statement-service | `StatementResource` | `from` (×2), `to` (×2) |
| openbank-pid-service | `PartyResource` | `index`, `type`, `value` |
| openbank-dispute-service | `DisputeResource` | `actor` (×2) |
| openbank-tpp-registry-service | `TppRegistryResource` | `tppId`, `role` |
| openbank-aml-service | `AmlCaseResource` | `Idempotency-Key` |
| openbank-party-service | `PartyResource` | `Idempotency-Key` |

Current list at any time:

```
python3 .github/scripts/check-nonnull-jaxrs-params.py --root . --list
```

## Suggested order

[...]

Issue #3104, opened 2026-08-01
Money-path: a missing non-nullable @QueryParam/@HeaderParam returns 500 — 57 candidate sites in 22 services


JiRaska:
## What is left after #3050 closed #3038

#3038 catalogued 28 money-path 500s from the first authenticated-fuzz run. Three PRs split it:

| | scope | state |
|---|---|---|
| #3032 | 16 `ApprovalResource` copies, null body | MERGED `36ff2ac` |
| #3050 | 12 handlers, null body | MERGED `7af4d19` |
| #3057 | `DateTimeException` unmapped + 10 catch-all `recoverWithItem` overrides | **open** |
| #3044 | fleet-wide guards | draft — inert as written, see its comments |

`Closes #3038` on #3050 closed the issue, but #3050's own body says it fixes 12 of the 28 and
explicitly hands six endpoints back as "a different defect class needing its own diagnosis". Most of
those six are in fact covered by #3057. **This issue is the part neither PR covers.**

## The uncovered class: a non-nullable reference-typed `@QueryParam`

```kotlin
// openbank-interest-service InterestResource.capitalize
fun capitalize(
    @PathParam("accountId") accountId: String,
    @QueryParam("productId") productId: String,      // <- non-nullable, no @DefaultValue
    @QueryParam("toDate") toDate: String?,
)
```

Omit `?productId=` and JAX-RS passes `null`. Kotlin's null-safety is compile-time only, so the
intrinsic null check throws NPE at the method boundary, `GenericExceptionMapper` renders **500**, and
the caller is told the server broke when the truth is a missing query parameter.

This is the same root shape as #3032/#3050 — a non-nullable Kotlin parameter fed null at runtime —
one argument position over. It is **not** fixed by them (they addressed entity bodies) and **not**
fixed by #3057 (a missing parameter produces NPE, not `DateTimeException`, and removing a catch-all
does not add a mapper for it).

Note the primitive case behaves differently and is already handled elsewhere: `year: Int` /
`month: Int` compile to JVM `int`, JAX-RS supplies `0` for a missing value, and `?year=0` then throws
`DateTimeException` — which is exactly what #3057 maps to 400.

## Scope — 57 candidate sites in 22 services, needing triage not a blanket fix

Measured on `origin/main`, counting `@QueryParam` / `@HeaderParam` / `@MatrixParam` declarations
whose Kotlin type is a non-nullable reference type with no `@DefaultValue`:

```
[...]

Issue #526, opened 2026-07-07
Non-deterministic error statuses: service-local ExceptionMappers collide with libs-runtime mappers on the same JDK exception types


JiRaska:
## Problem

`openbank-libs-runtime` auto-registers (via Jandex) `ExceptionMapper<IllegalStateException>` (→ **422** `BUSINESS_RULE_VIOLATION`) and `ExceptionMapper<IllegalArgumentException>` (→ **400** `VALIDATION_ERROR`) in `com/openbank/libs/api/error/CommonExceptionMappers.kt`. At least one service — openbank-ledger-service (`infrastructure/rest/ExceptionMappers.kt`) — registers its **own** mappers for the *same exception types* with *different* status codes (IllegalState → 409, IllegalArgument → 422).

JAX-RS mapper selection between two providers registered for the **same** exception type is non-deterministic — exactly the collision ADR-0049 D4 warns about for `ExceptionMapper<Exception>`, but on the narrower types. Observed live while building the #465 ledger concurrency IT: racing reversals produced **a 409 from one request and a 422 from another for the identical `IllegalStateException`**, flip-flopping between runs. The ledger file's own comment ("These override libs' defaults intentionally") is wrong — there is no defined override order; it's a per-request lottery.

Consequences: unstable API contract (clients can't rely on documented status codes), untestable error paths (assertions on status codes are flaky), and OpenAPI specs that document whichever mapper happened to win during manual testing.

## Proposed solution

1. **Rule**: a service must NOT register an `ExceptionMapper` for a JDK exception type that libs-runtime already maps (`IllegalStateException`, `IllegalArgumentException`, `NoSuchElementException`, `Exception`, `WebApplicationException`). Services that need a different status code introduce a **dedicated domain exception + mapper** (ledger now does this for reversal conflicts: `JournalReversalConflictException` → 409, from the #465 work).
2. **Sweep**: audit the fleet for service-local mappers on JDK types (`grep -rn "ExceptionMapper<Illegal\|ExceptionMapper<NoSuchElement"` over `openbank-*/src/main`), replace each with a dedicated exception type or delete it if libs' default status is acceptable. Ledger's `IllegalStateExceptionMapper`/`IllegalArgumentExceptionMapper` are the known instances.
3. **Guard**: a CI check (or an ArchUnit/Konsist-style test in libs) asserting no service registers a mapper for the libs-owned types — encode in `rules.yaml` per the knowledge-capture spec.

## Alternatives considered

[...]

PR #3625, opened 2026-08-03
fix(money-path): a missing required query/header parameter answered 500, not 400

JiRaska:
## What

A required `@QueryParam` / `@HeaderParam` declared with a non-nullable Kotlin reference type answers **500** when the caller omits it, never 400. A client cannot tell "I sent a bad request" from "the server is broken", and on a money path that difference decides whether they retry. Same root shape as the null-body 500s of #3032 / #3050, one argument position over.

This fixes the **24 money-path sites**, adds an **enforced ratchet gate**, and baselines the 39 remaining with #3624 tracking them.

## The measurement: 57 candidates → 63 sites → 24 fixed here

#3104 counted **57** candidate declarations. Rescanned with a comment-stripping parser the raw count is **115** — and the discriminator that neither number applies is **inbound vs outbound**:

| | count | verdict |
|---|---:|---|
| outbound MicroProfile REST **client interfaces** | 52 | **not a defect** |
| inbound JAX-RS **resource classes** | 63 | the real set |
| ├ money-path | 24 | fixed here |
| └ elsewhere | 39 | baselined, tracked in #3624 |

The 52 are not a defect for two independent reasons: the annotation there describes the request this service **sends**, so the argument comes from our own Kotlin call site and is compile-time checked; and an abstract interface method has no body to hold an intrinsic. A naive grep over-counts by roughly 2x because of them.

## Why the obvious fix is dead code — measured, not assumed

Both shapes compiled and read back with `javap` (Kotlin 2.3.21):

```
fun f(p: String)
  0: aload_1
  1: ldc           "p"
  3: invokestatic  kotlin/jvm/internal/Intrinsics.checkNotNullParameter   <-- offset 0
  6: goto 9                                                               <-- requireNotNull(p)
  9: aload_1                                                                  compiled to NOTHING

suspend fun f(p: String)
  0: aload_1
  1: areturn                        <-- no intrinsic emitted AT ALL
```

Two consequences, and the second is the one worth carrying forward:

1. Writing `requireNotNull(p)` in the body of a **non-suspend** handler is dead code — the intrinsic has already thrown. **The parameter must be declared nullable for any guard to be reachable.** That is what this PR does.
2. A **suspend** handler gets no intrinsic, so `null` flows into the body. Usually it still 500s at the first dereference — but where nothing dereferences it, the request proceeds with a null the signature promised could not exist. That outcome is worse than the 500 and invisible from every angle.

No new exception mapper anywhere: libs-runtime's `CommonExceptionMappers` already maps `IllegalArgumentException` to 400 with the `ApiError` envelope and a traceId. A service-local mapper for a JDK type races the libs one per request (#526).

## Two guards that could never run

sepa-payment, domestic-payment and fx-service already wrote:

```kotlin
require(idempotencyKey.isNotBlank()) { "Idempotency-Key header is required" }
```

Those handlers are `suspend`, so an **absent** header injected null and `null.isNotBlank()` threw NPE — the guard answered 500 in exactly the case it was written for, while a **blank** header correctly gave 400. Now `require(!idempotencyKey.isNullOrBlank())`.

## One site that was already correct

consent's `hasActiveConsent` already answered 400 for an absent `scope`, because `runCatching { ConsentScope.valueOf(scope) }` swallows the NPE and the `requireNotNull` under it fires. Changed anyway so the message names the missing parameter instead of reporting `unknown scope: null`, and so the declared type stops promising non-null — but it is **not** counted as a fixed 500.

## What changed

24 sites across account (6), consent (4), delegation (4), interest (2), lending (2), sdd (2), domestic-payment, fx, sepa-payment, transaction (1 each). Every one classified as **truly required** — none wanted `@DefaultValue`, none turned out to be genuinely optional.

No `openapi.yaml` changes: the specs are static and already declare these parameters. Only the status of an invalid request moves, 500 → 400.

## The gate

`.github/scripts/check-nonnull-jaxrs-params.py`, registered as `nonnull-jaxrs-param-ratchet` (group `kotlin`, **enforced**). `gates.yaml` id count 102 → 103, verified against live `origin/main` immediately before committing.

A **ratchet** rather than advisory: remediation is genuinely per-handler (required / `@DefaultValue` / actually optional) and only the handler knows which, so a flat check cannot be clean, and this repo's own lesson is that an advisory red gets merged past. The 39 remaining sites are baselined **with their service**, so a new occurrence fails **and** a baseline entry that becomes fixed is reported — the list cannot rot in either direction.

### Falsified in both directions before being trusted

Restoring from a `cp` backup, never `reset --hard` (there were 12 files of uncommitted work in the tree):

```
control (fixed tree)                              exit 0
revert one productId to non-nullable              exit 1, ::error naming file+line
restore                                           exit 0, byte-identical to backup

make a baselined site nullable                    exit 1, ::error STALE baseline entry
restore                                           exit 0, byte-identical to backup
```

The self-test is 14 cases and the **must-NOT-flag half is most of them**: the outbound interface, a nullable type, `@DefaultValue`, a Kotlin default, a primitive, a commented-out declaration, a string literal, and prose inside a **nested** block comment naming the annotation. Kotlin block comments nest, so a KDoc containing `/*` does not end at the first `*/`; the scan mirrors that and the fixture proves it. It asserts against the **construct** — the annotation plus the declared type — never a whole-file grep, so it cannot match the prose about the thing.

A 14th case was added after a latent **false green** was found in the gate itself: the inbound/outbound test excluded everything that was not a `class`, so a *named* nested `object` declared above a handler would have silently skipped every handler below it. 0 such sites exist today — the point is that adding one must not disarm the gate. It now excludes only `interface`.

Worth recording how that was caught, because the first attempt was vacuous: the fixture originally used a `companion object`, and reverting the exclusion to the broken form returned **14 passed, 0 failed** — green against the code it was written to reject, because `DECL_RE` needs a name after the keyword and an unnamed companion never becomes the enclosing declaration. The fixture was asserting a hazard that does not exist while the real one went uncovered. With `object Headers` the same injection reports `FAIL must flag flag_after_object`, exit 2. Reading the fixture would not have found this; only running the injection did.

## Threat models (ADR-0030 D2)

The `threat-model updated on trust-boundary change` gate flagged all ten services, correctly — it cannot know from a diff that a change to a money-path inbound REST surface is a status correction rather than a new surface, and that judgement is exactly what the entry records. Ten entries added, written per service because the security reading differs with the parameter:

- **idempotency key** — a *replay* control, and in three services it was partially inoperative (see above).
- **`partyId` / `role` / `intent` / `granteePartyId`** — *inputs to an access decision*. A null reaching the use case is a decision taken on an absent subject rather than a rejected request.
- **`scaSessionId`** — the *evidence* that SCA completed (consent activate, delegation accept).
- **`debitDate`** — the input to sdd's refund-window arithmetic; must be rejected, never defaulted.

Each states the same three things a reviewer needs — no new caller, no new trust boundary, and the guards run **after** the unchanged `@RolesAllowed` / `@Authorize` gates — plus its rollback. delegation-service had no change-log section; one was added rather than bolting the entry onto an unrelated heading.

## Verification

- `ktlintCheck` + `detekt` green locally on all 10 touched services.
- `:openbank-interest-service:build` green locally. A first run failed on `QuarkusBindException: Port already bound: 8081` — a concurrent build on this machine, the documented environmental failure, not this change; re-running it alone was green.
- The remaining 9 service builds are covered by the CI per-service matrix, which is the same gate.
- `run-gates.py --group security` green locally after the threat-model commit (it refuses to run without `PR_DIFF_BASE` rather than passing vacuously — worth knowing before trusting a local green on that one).

## Linked issues

Refs #3104 — the money-path half. Not closed: 39 non-money-path sites remain, tracked in #3624.
Refs #3038, #3050, #3032, #526

### DIFF
--- a/.github/gates/gates.yaml
+++ b/.github/gates/gates.yaml
@@ -1214,6 +1214,35 @@ gates:
     run: |
       bash .github/scripts/check-exception-mapper-collision.sh .
 
+  # A required @QueryParam/@HeaderParam/@MatrixParam declared with a NON-NULLABLE
+  # Kotlin reference type answers 500 when the caller omits it, never 400 — the
+  # caller cannot tell "I sent a bad request" from "the server is broken", and on
+  # a money path that decides whether they retry (issue #3104). Same root shape as
+  # the null-body 500s of #3032/#3050, one argument position over.
+  #
+  # The trap this gate exists to stop being re-learned: writing requireNotNull in
+  # the BODY of a non-suspend handler is dead code. Kotlin emits
+  # Intrinsics.checkNotNullParameter at bytecode offset 0, so the NPE has already
+  # thrown (verified with javap; a suspend function gets no intrinsic at all, and
+  # the null then flows into the body instead). The parameter must be declared
+  # NULLABLE for any guard to be reachable; libs-runtime maps the resulting
+  # IllegalArgumentException to 400 — never add a service-local mapper (#526).
+  #
+  # Ratchet, not a flat fail: remediation is per-handler (required / @DefaultValue
+  # / genuinely optional) and only the handler knows which. The 24 money-path
+  # sites are fixed; the 39 remaining are baselined in the script with their
+  # service, so a NEW occurrence fails and a baseline entry that becomes fixed is
+  # reported too. Self-test first — the must-NOT-flag half is what matters here,
+  # since an outbound REST-client interface carries the identical annotation and
+  # is not a defect (52 of the 115 raw matches on main). ENFORCED.
+  - id: nonnull-jaxrs-param-ratchet
+    name: "non-nullable required JAX-RS param ratchet (issue #3104, enforced)"
+    group: kotlin
+    mode: enforced
+    run: |
+      python3 .github/scripts/check-nonnull-jaxrs-params.py --self-test
+      python3 .github/scripts/check-nonnull-jaxrs-params.py --root .
+
   # openbank.outbox.dispatch-enabled guard (the CLAUDE.md "outbox footgun":
   # a service whose OutboxDispatcher actually gates on this flag silently
   # never dispatches if it's left at its false default — no error,
--- a/.github/scripts/check-nonnull-jaxrs-params.py
+++ b/.github/scripts/check-nonnull-jaxrs-params.py
@@ -171,6 +171,12 @@ def enclosing_type(src: str, offset: int):
     `kind` is the discriminator that matters: an `interface` is an outbound MicroProfile REST
     client, a `class` is an inbound resource. Files that hold both (a resource plus its client)
     resolve correctly because the LAST declaration before the offset wins.
+
+    Callers must exclude on `== "interface"`, NOT on `!= "class"`. A `companion object` declared
+    above a handler inside a resource class resolves to `object`, and excluding everything that is
+    not a `class` would then skip that handler silently — a false GREEN, which is the direction
+    that never announces itself. There are 0 such sites today; the point is that adding one must
+    not disarm the gate.
     """
     last = None
     for m in DECL_RE.finditer(src, 0, offset):
@@ -207,7 +213,7 @@ def scan_source(src: str, service: str, path: str):
             continue
 
         encl_kind, encl_name = enclosing_type(src, m.start())
-        if encl_kind != "class":
+        if encl_kind == "interface":
             continue  # outbound REST client — the caller supplies the argument
 
         yield {
@@ -265,6 +271,19 @@ class DemoResource {
 
     fun allowedPrimitive(@QueryParam("ok_primitive") g: Int): Response = TODO()
 
+    // A NAMED nested object above a handler must not disarm the check for handlers below it. With
+    // the exclusion written as `!= "class"` this site resolves to `object` and is silently skipped
+    // — a false GREEN, the direction that never announces itself. Excluding only `interface` keeps
+    // it flagged. (An unnamed `companion object` is NOT the hazard: DECL_RE requires a name after
+    // the keyword, so it never becomes the enclosing declaration. Measured — the first version of
+    // this fixture used `companion object` and passed against the broken exclusion too, i.e. it
+    // proved nothing.)
+    object Headers {
+        const val OPERATOR = "X-Operator-Id"
+    }
+
+    fun flaggedAfterNestedObject(@QueryParam("flag_after_object") i: String): Response = TODO()
+
     // A commented-out declaration must not count:
     // fun commentedOut(@QueryParam("comment_only") z: String): Response = TODO()
     fun stringLiteralIsNotADecl(): String = "@QueryParam(\\"literal_only\\") y: String"
@@ -277,7 +296,7 @@ interface DemoClient {
 }
 '''
 
-SELF_TEST_EXPECTED_FLAGGED = {"flag_plain", "flag_header", "flag_enum"}
+SELF_TEST_EXPECTED_FLAGGED = {"flag_plain", "flag_header", "flag_enum", "flag_after_object"}
 SELF_TEST_EXPECTED_ALLOWED = {
     "prose_only", "nested_prose", "ok_nullable", "ok_default_ann", "ok_kotlin_default",
     "ok_primitive", "comment_only", "literal_only", "ok_outbound",
--- a/.github/scripts/check-nonnull-jaxrs-params.py
+++ b/.github/scripts/check-nonnull-jaxrs-params.py
@@ -1,0 +1,366 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) OpenBank contributors. Licensed under the Apache License, Version 2.0.
+#
+# A required @QueryParam / @HeaderParam / @MatrixParam declared with a NON-NULLABLE Kotlin
+# reference type answers 500 when the caller omits it, never 400 (issue #3104).
+#
+# THE MECHANISM, AND WHY THE OBVIOUS FIX IS DEAD CODE
+#   JAX-RS injects `null` for an absent parameter. Kotlin's null-safety is compile-time only, so the
+#   declared non-nullable type buys nothing at runtime — it only decides WHERE the failure lands:
+#
+#     fun f(@QueryParam("p") p: String)          -> Intrinsics.checkNotNullParameter at bytecode
+#                                                   offset 0. NPE before the first statement of the
+#                                                   body, every time. GenericExceptionMapper -> 500.
+#
+#     suspend fun f(@QueryParam("p") p: String)  -> NO intrinsic is emitted at all (verified against
+#                                                   the 2.3.21 compiler with javap, for suspending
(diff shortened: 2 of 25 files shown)


JiRaska:
## Status: awaiting human review — auto-merge deliberately not used

`gh pr merge --squash --auto` was refused by the money-path guard:

> BLOCKED: PR #3625 touches money-path/high-risk paths regardless of its title scope. Matched: `AccountResource.kt`, `AuthorizationResource.kt`, `SavingsGoalDelegationResource.kt`. Leave it for human review.

That is the correct outcome and it was not worked around — no `--admin`, no override flag. Ten money-path services, two of the parameters being inputs to an access decision, is exactly the shape that rule exists for. Leaving it for a human.

## What a reviewer should check, in order of what could actually be wrong

1. **The nullable-parameter conversions are behaviour-preserving for a well-formed request.** Every guard runs *after* the unchanged `@RolesAllowed` / `@Authorize` gates — `AuthorizeInterceptor` is `@AroundInvoke`, so authorization is evaluated before the method body. Verified that no `@Authorize(resource = "#x")` expression now points at a parameter this PR made nullable: **0 new**, against 3 pre-existing on `origin/main` (`#request` ×2 from #3050, `#granteeId`, all already nullable there).
2. **The three `require(!x.isNullOrBlank())` rewrites.** These rely on Kotlin's contract smart-cast to keep the later non-null uses compiling. They do — all nine services compile — but it is the only non-mechanical edit in the diff.
3. **`hasActiveConsent` is the one changed handler that was not broken.** It already answered 400 by accident. If you would rather not touch a working handler, that one is separable.

## Verification performed

- `ktlintCheck` + `detekt` — green locally on all 10 touched services.
- `compileKotlin` — green locally on all 9 Kotlin services in the diff.
- Full `:build` (incl. tests) — green locally on interest, lending, sdd, transaction.
- CI per-service build matrix — green for consent, delegation, domestic-payment (and the rest on the current head at time of writing).
- `gates (kotlin)` — green, which is the new ratchet running for real.
- `gates (security)` — green after the threat-model commit. It went red first, correctly, and that is worth noting: the ADR-0030 D2 gate cannot know from a diff that a money-path REST-surface change is a status correction rather than a new surface. Locally it refuses to run at all without `PR_DIFF_BASE` rather than passing vacuously.
- `gates.yaml` id count re-checked against live `origin/main` after a competing gate (`route-exports`) landed: `git merge-tree` yields **104**, so both entries survive the merge and neither is silently dropped.

## Known limitation of the gate, stated rather than discovered later

The baseline key is `service|Class|param`, with no count. A *new* occurrence of an *already-baselined* parameter in the same class — an eighth `X-Operator-Id` on `CardResource` — would not be flagged. That is deliberate: pinning counts would fail the gate on an unrelated refactor that merely moves a handler. The trade is a narrow false-green against frequent false-reds, and the whole of `CardResource` is already in #3624 as work to do.

JiRaska:
### The `agent-review` red here is infrastructural, not a verdict

Read from the job log rather than the console: the `Classify change scope` step died on a git fetch, before it evaluated anything.

```
error: RPC failed; HTTP 503 curl 22 The requested URL returned error: 503
##[error]Process completed with exit code 128.
```

That is a transient GitHub-side 503 during the branch fetch, so the step never reached the money-path / governance classification it exists to perform — the red says nothing about this PR's content. A re-run should clear it. I have deliberately not re-run it myself: this is your branch and your worktree is live, and the PR still needs its two money-path approvals regardless, so a rerun would not unblock the merge on its own.

Flagging it because this PR is now the bottom of a stack. #3658 (psd2 + card-issuance, 21 sites) and #3660 (the other seven services, 18 sites) close out the #3624 tail on top of it, and they carry the `BASELINE` edits for `check-nonnull-jaxrs-params.py`, so they cannot reach `main` before this does. Merge order is #3625 → #3658 → #3660.

JiRaska:
### ⚠️ Before merging this PR: retarget #3658 to `main` first, or it and #3660 are closed irreversibly

`delete_branch_on_merge` is **`true`** on this repo (verified via the API just now), so merging this PR deletes `fix/nonnull-queryparam-400` **regardless of whether `--delete-branch` is passed** — omitting the flag is not protection here. GitHub then closes every PR targeting the deleted branch, and a closed PR whose base is gone can be neither reopened nor retargeted:

- `reopenPullRequest` → `Could not open the pull request`
- `updatePullRequest` → `Cannot change the base branch of a closed pull request`

That would take out **#3658** (base `fix/nonnull-queryparam-400`) and then **#3660** (base `fix/nonnull-params-psd2-cards`) behind it — losing both numbers and their review threads. Same failure as #3055 closing #3063. It is silent at merge time: `gh pr merge` prints nothing about the children and the merge itself succeeds.

**Order that avoids it:**

1. `gh pr edit 3658 --base main`
2. merge this PR
3. `gh pr edit 3660 --base main` after #3658 lands (or leave it stacked on #3658 and retarget then)

I deliberately did **not** pre-retarget #3658 to `main` myself. Its branch contains this PR's money-path commits, so with `base: main` a well-meaning merge of #3658 *first* would put this PR's money-path changes on `main` under my commit message, without the two money-path approvals they require. Keeping the base pointed here makes that structurally impossible. The branch-deletion hazard is recoverable (new PRs off the same branches); an unreviewed money-path merge is not.

JiRaska:
**Merge-order warning — this PR is the parent of a stack, and merging it first will irreversibly close a child.**

```
#3660  fix/nonnull-params-remaining-seven  ->  fix/nonnull-params-psd2-cards
#3658  fix/nonnull-params-psd2-cards       ->  fix/nonnull-queryparam-400   <-- this PR's branch
#3625  fix/nonnull-queryparam-400          ->  main
```

The repo sets `delete_branch_on_merge: true` — verified just now — so merging this PR deletes `fix/nonnull-queryparam-400` regardless of any `--delete-branch` flag. GitHub then **closes every PR targeting that branch**, i.e. #3658.

A closed PR whose base branch no longer exists **cannot be reopened and cannot be retargeted**: `reopenPullRequest` answers *"Could not open the pull request"* and `updatePullRequest` answers *"Cannot change the base branch of a closed pull request"*. The only way back is a rebase onto the new `main` and a **new** PR — losing the number, the review history and the comment thread. The code survives; the review record does not.

The failure is silent at merge time: `gh pr merge` prints nothing about the child and the merge itself succeeds, so it reads as clean from every angle you would normally check. (Same shape as #3055 closing #3063, reopened as #3111.)

**Two safe orders, either is fine:**

1. **Children first** — merge #3660, then #3658, then this one. Each child's base still exists when it merges.
2. **Retarget first** — point #3658 at `main` *before* merging this, then merge in any order.

The agent that opened #3658/#3660 deliberately did **not** pre-retarget them, and the reasoning is worth keeping: retargeting a child to `main` while this PR is unmerged means a wrong-order merge would land this PR's money-path commits on `main` under the child's authorship. Given the choice, it left the recoverable failure (a closed PR) rather than the unrecoverable one (money-path commits merged unreviewed under the wrong PR).

`gh pr list --base fix/nonnull-queryparam-400` before merging will show what is still pointing here.

PR #3629, opened 2026-08-03
build(fleet): retire the alpine base from the 53 files that only declare it

JiRaska:
## Summary

#3618 moves the fleet runtime base from `eclipse-temurin:25-jre-alpine` (musl) to
`eclipse-temurin:25-jre` (glibc). Nothing else builds images, so 53 Dockerfiles and 12 service
docs went on declaring the musl base and nothing broke — which is the problem. This retires
those declarations and adds the gate that stops the second copy drifting again.

**Stacked on #3618** — base branch is `fix/fraud-onnx-glibc-runtime-base`, not `main`. The new
rule compares each `FROM` against `Dockerfile.deploy`, which still says alpine on `main`, so this
cannot go green until the parent lands. Please do **not** merge #3618 with `--delete-branch`: that
closes this PR irreversibly (a closed PR whose base is gone can be neither reopened nor
retargeted). Either retarget this to `main` first, or merge #3618 without the flag.

## Type of change

- [x] chore (build / tooling) — commit type `build`, see *Why `build` and not `fix`* below

## Linked issues

Refs #3354
Refs #3016

## Changes

- **50 `openbank-*/Dockerfile`**, plus `Dockerfile.scanner`,
  `openbank-security-scanner/Dockerfile.prebuilt` and
  `openbank-notification-service/Dockerfile.runtime` — now carry `Dockerfile.deploy`'s `FROM`
  verbatim, and its `groupadd`/`useradd` form. busybox `addgroup`/`adduser` do not exist on the
  glibc base, so copying only the `FROM` would have left a recipe that cannot run.
  `Dockerfile.runtime` keeps its own uid/gid 1001.
- **12 `src/main/resources/docs/05-operations.{cs,en}.md`** — aml, customer-edge, kyc, onboarding,
  pid, product-catalog, sca. Every one also claimed a multi-stage in-Docker Gradle build that
  #3016 removed; pid's and onboarding's even gave a `docker build -f <svc>/Dockerfile .` command
  that cannot work, because the context has no `quarkus-app/`. Corrected together, not just the
  image name.
- **`openbank-fraud-service/src/main/resources/application.yaml` and `OnnxFraudModel.kt`** — both
  said in the present tense that the deploy image *is* alpine. #3618 makes that false the moment
  it lands. The application.yaml block now also states what #3618's `verify-image-native-libs.py`
  does and does not prove: the library can load; nobody has yet observed a deployed pod scoring.
- **`interest-service-http-scaledobject.yaml`** — one stale `jre-alpine` claim in a comment.
- **`check-dockerfile-no-build-stage.py` gains rule 3** + a `--self-test`, and the gate moves from
  an inline `ci.yml` step into `.github/gates/gates.yaml` — see below.

## The gate could not fire on this PR

Worth reading even if you skim the rest. The #3016 gate was an inline step in `ci.yml`'s
`ui-build` job, and that job is conditional:

```yaml
  ui-build:
    needs: changes-ui
    if: needs.changes-ui.outputs.changed == 'true'
```

`changes-ui` diffs `openbank-admin-ui/`, `*/governance.yaml` and the governance schema. **So the
gate that owns Dockerfile shape never ran on a PR that changes only Dockerfiles** — the exact
change it exists to catch. Not skipped-and-reported: the job is absent, and the aggregate check is
green about work it never did.

Measured on this branch's first commit: `Validate manifests` SUCCESS with 3 steps (it is an
aggregator over the `gates (...)` shards), `Admin UI build` SKIPPED, and the 53-Dockerfile diff
therefore checked by nothing. Both the pre-existing #3016 rules and the new rule 3 were vacuous
here.

Fixed in the second commit: declared in `.github/gates/gates.yaml`, group `registry`, `mode:
enforced`, `selftest_expect: pass`. Shards run unconditionally. `gate-script-registration` now
reports 80 in gates.yaml / 0 never run, and `gates-not-deregistered` 102 -> 103.

## Why keep the `FROM` at all, rather than delete it

The obvious long-term fix — stop these files carrying a `FROM`, since only `EXPOSE` is live — is
wrong here, and not obviously so. **They are not read only by the deploy pipeline.**
`openbank-admin-ui/scripts/generate-cluster-topology.mjs` parses
`openbank-ledger-service/Dockerfile` in `imageFacts()` and renders the base image into the
`/docs/cluster` dossier (ADR-0081). With no `FROM` to parse, `img.ok` is false and it falls back
to a **hardcoded `eclipse-temurin:25-jre-alpine` literal** — deleting the second copy would
resurrect the fiction in the UI instead of retiring it.

So: keep the declaration, and make it impossible for it to disagree.

## Falsifiability

Rule 3 reads **comment-stripped** text on purpose. `Dockerfile.deploy`'s header quotes the retired
image three times to explain why it is retired, and every per-service header now names
`Dockerfile.deploy`; a raw-text scan would read that prose as a declaration — the #2450 collision
class. `--self-test` asserts four things against a synthetic tree: the rule fires on a drifted
`FROM`, stays quiet when every `FROM` agrees (prose and all), covers the three files outside the
`openbank-*/Dockerfile` glob, and exempts `SELF_BUILT`.

Also falsified against the **real** tree, not only the synthetic one: reverting
`openbank-vop-service/Dockerfile` to the alpine digest gives `--enforce` exit 1 with the vop file
named; restoring it gives exit 0 (file restored and shasum-verified).

## Deliberately unchanged

`openbank-infra/docker/services/{standalone,root}-service.Dockerfile` still declare an alpine JRE.
Those are **live** — `openbank-infra/docker-compose.yml` builds 20 services from them — and are
self-contained multi-stage uber-jar recipes for local dev, not copies of the fleet base.
`openbank-fraud-service` is not among the compose services, so there is no ONNX exposure there.
Changing them would alter real local builds for no stated reason.

One follow-up filed rather than folded in, since it is a separate defect in a different tree —
**#3630**: the `generate-cluster-topology.mjs` fallback literals still describe a multi-stage
alpine build that has not existed since #3016, and the committed `cluster-topology.json` is stale
in the same way (`multiStage: true`, `buildBase: …25-jdk-alpine`). Rule 3 makes that fallback
unreachable in practice without making it correct.

## Why `build` and not `fix`

`<svc>/Dockerfile` and `<svc>/src/main/resources/**` are both release-triggering paths —
`exclude-paths` in `release-please-config.json` is `src/test` only — so a releasing type
(`feat`/`fix`/`perf`/`security`) on a PR touching ~50 service paths would cut ~50 patch releases
for a change that alters no shipped code.

Verified rather than assumed, both axes:

1. `release-please-config.json`'s `changelog-sections` marks `refactor` `docs` `test` `build` `ci`
   `chore` as `hidden: true`; only `feat` `fix` `perf` `security` are visible. (`rules.yaml`
   `commits` lists `refactor` as `semver_bump: patch`, which is why `build` was chosen over
   `refactor` — no ambiguity under either document.)
2. Empirically: `385d34653` `chore(fraud): …` (#3503) touched `openbank-fraud-service` outside
   `src/test`, and `openbank-fraud-service/CHANGELOG.md` has had no commit since. Same for
   `eef5ac945` `build: fix uber-jar COPY path in three service Dockerfiles` (#3015) — the closest
   precedent to this PR, same type, no release for any of the three services.

## Definition of Done

- [x] Hexagonal layering preserved (no production code changed; two comment corrections).
- [x] Unit tests added/updated — `--self-test` in `check-dockerfile-no-build-stage.py`.
- [ ] Integration tests — N/A, no service boundary involved.
- [ ] Contract tests — N/A, no public API changed.
- [x] No new lint warnings — `actionlint` clean on `ci.yml`; yamllint finding sets identical to
      `origin/main` under the CI config (probe validated against a known-positive duplicate key).
- [ ] OpenAPI / Flyway / Avro — N/A.
- [x] Docs updated.

## Security checklist

- [x] No new `@Suppress`.
- [ ] PII path — unchanged.
- [ ] Cardholder data path — unchanged.

## Compliance impact

- [x] None

## Rollout / Rollback

- Deployment strategy: N/A — no image is built from any file in this diff. The one behavioural
  change is a CI gate.

## Reviewer notes

The gate is the part worth reviewing. The 53 file edits are mechanical and machine-applied; the
judgement calls are (a) keeping the `FROM` rather than deleting it, for the admin-ui reason above,
and (b) the scope of `EXTRA_FLEET_BASE` — three files nothing references today, which drift exactly
like the per-service ones and are now held to the same rule.

### DIFF
--- a/.github/gates/gates.yaml
+++ b/.github/gates/gates.yaml
@@ -1024,6 +1024,21 @@ gates:
     run: |
       python3 .github/scripts/check-gates-not-deregistered.py --enforce
 
+  - id: dockerfile-runtime-only
+    name: "Dockerfiles are runtime-only, declare EXPOSE, share the deploy base (#3016/#3354, enforced)"
+    group: registry
+    mode: enforced
+    selftest: |
+      python3 .github/scripts/check-dockerfile-no-build-stage.py --self-test
+    selftest_expect: pass
+    run: |
+      # This lived as an inline step in ci.yml's `ui-build` job, which runs only when
+      # openbank-admin-ui/, */governance.yaml or the governance schema changed. So the gate
+      # that owns Dockerfile shape could not fire on a Dockerfile-only PR — the exact change
+      # it exists to catch — and reported nothing rather than reporting a skip. Found on
+      # #3629, whose 53 Dockerfile edits skipped it entirely. A shard runs unconditionally.
+      python3 .github/scripts/check-dockerfile-no-build-stage.py --enforce
+
   - id: gate-script-registration
     name: "every check-* script is actually run (#3240, enforced)"
     group: registry
--- a/.github/scripts/check-dockerfile-no-build-stage.py
+++ b/.github/scripts/check-dockerfile-no-build-stage.py
@@ -15,14 +15,30 @@ Nothing builds them. `auto-deploy.yml`, `ghcr-publish.yml` and
 `.github/workflows/Dockerfile.deploy`, and read exactly one thing from these files:
 the EXPOSE line.
 
-So two rules:
+So three rules:
 
   1. No build stage. A `FROM ... AS build` here is either dead text or a second,
      divergent definition of how images are made. Both are worse than nothing.
   2. EXPOSE is mandatory. It is the only live contract — losing it does not fail
      anything loudly, it silently sets the container port to 8080.
+  3. The FROM must be byte-identical to the one in `Dockerfile.deploy` (#3354).
 
-BOTH rules assume the file describes a Quarkus service deployed by that pipeline.
+Rule 3 exists because #3618 moved the fleet runtime base from
+`eclipse-temurin:25-jre-alpine` (musl) to `eclipse-temurin:25-jre` (glibc) — the
+ONNX Runtime `.so` openbank-fraud-service bundles is glibc-linked and had never
+loaded — and 53 files went on declaring the musl base. Nothing broke, which is the
+problem: a wrong FROM here is exactly the "reads as authoritative while being
+fiction" failure #3016 exists to prevent.
+
+Deleting the FROM instead would be worse, and that is not obvious. These files are
+NOT read only by the deploy pipeline: `openbank-admin-ui/scripts/generate-cluster-topology.mjs`
+parses `openbank-ledger-service/Dockerfile` for `imageFacts()` and renders the base
+image into the /docs/cluster dossier (ADR-0081). With no FROM to parse it falls back
+to a HARDCODED `eclipse-temurin:25-jre-alpine` literal — i.e. removing the second
+copy would resurrect the fiction in the UI rather than retire it. So: keep the FROM,
+and make it impossible for it to disagree.
+
+Rules 1 and 2 assume the file describes a Quarkus service deployed by that pipeline.
 Three `openbank-*/` directories are not that, and for them a build stage is the
 whole point (SELF_BUILT below). The first sweep flattened them into the generic
 Quarkus recipe, which broke the admin-ui image build for real: `admin-ui-deploy.yml`
@@ -38,6 +54,7 @@ prevent.
 Usage:
     check-dockerfile-no-build-stage.py            # warn
     check-dockerfile-no-build-stage.py --enforce  # fail
+    check-dockerfile-no-build-stage.py --self-test  # falsify rule 3 both ways
 """
 
 from __future__ import annotations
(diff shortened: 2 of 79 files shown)


JiRaska:
**Verdict: excellent gate craftsmanship, but this PR is one of TWO competing, independent fixes for the same problem — they will collide.** 1 major finding, otherwise clean.

- **This PR is stacked on #3618** (`fix(fraud): move the runtime image to glibc so the ONNX model can actually load`, still OPEN, branch `fix/fraud-onnx-glibc-runtime-base`), which retargets `Dockerfile.deploy` and the fleet to `eclipse-temurin:25-jre@sha256:681c543...` (full Debian-based glibc image). **This same review batch also contains #3700 + #3709**, an independent chain that retargets the identical set of files — `Dockerfile.deploy`, `Dockerfile.scanner`, all 53 per-service Dockerfiles — to `eclipse-temurin:25-jre-ubi10-minimal@sha256:6031e548...` (RHEL-based minimal glibc image), for the *same* stated reason (#3354, ONNX runtime needs glibc). Two independently-authored solutions to the same defect, touching the same lines in the same 53+ files, with different final base images and different user-creation shell syntax (`groupadd --system --gid 101 ... --shell /usr/sbin/nologin` here vs `groupadd -r ... useradd -r -g ...` in #3709). If both merge — in either order — whichever lands second either conflicts outright or silently reverts the first's base-image choice with a clean merge that prints nothing (this repo's own documented failure mode: a merge git calls clean can still discard content). One of these two chains needs to be closed or explicitly rebased onto the other before either merges; right now nothing surfaces that they're racing, since they're on unrelated branch names and neither PR references the other.
- Everything else in this PR is solid and independently verifiable regardless of which base wins: the new `dockerfile-runtime-only` gate (rule 3, "FROM must match `Dockerfile.deploy` byte-for-byte") is genuinely well falsified — `self_test()` checks agreement (no false positive), drift detection (fires and localizes correctly, doesn't over-flag the agreeing file), the non-glob `EXTRA_FLEET_BASE` files are in scope, and `SELF_BUILT` directories are correctly exempt. `strip_comments()` before the `FROM_RE` match is deliberate and explained (the file's own header prose names the retired base three times, which a raw-text scan would misread as a declaration — the exact code-about-code collision class #2450 already burned this repo on).
- The gate's relocation out of `ci.yml`'s inline step into `.github/gates/gates.yaml` under `group: registry` is exactly the fix this repo's own lesson prescribes, and the PR states plainly why: *"Found on #3629, whose 53 Dockerfile edits skipped it entirely"* — i.e. the inline step's conditional host job (`ui-build`, gated on admin-ui/governance paths) never ran on this very PR's own Dockerfile-only diff, which is a real demonstrated instance of the "gate reachable only by the job it sits in" trap, caught in the act.

Verified clean: `check-dockerfile-no-build-stage.py`'s rule 1/2 (no build stage, mandatory EXPOSE) are unaffected; `docs/05-operations.*.md` doc updates in this diff appear to be incidental regenerated artifacts, not hand-edited fiction.

PR #3632, opened 2026-08-03
fix(account): the propose-only savings withdrawal approval could never succeed

JiRaska:
## What this is

Refs #2990 (AC8), ADR-0232 D8.

I was asked to build the propose-only savings flow end to end. **The premise was wrong, and that
is the most useful thing in this PR.** #3143 — which the task described as shipping "the vocabulary
and the guard, not the flow" — actually shipped the whole flow: the `WithdrawalProposal` aggregate,
`SavingsProposalService` (propose / decide / cancel / list), `V18`, the `SavingsWithdrawalApproved`
event, the REST resource, the OpenAPI paths and an E2E IT. The claim that `SavingsWithdrawalApproved`
"returns zero hits fleet-wide" is false; it is in `AccountEvents.kt` on `main`.

So I checked what the shipped flow actually *does*. **The owner-approval path could never succeed
in production**, and separately, the second factor was not single-use. This PR fixes both, adds the
expiry lifecycle that was missing, and does not rebuild what already works.

## The defect: owner approval was structurally impossible

`SavingsProposalService.verifyOwnerSca` pre-checked `challenge.status != "COMPLETED"`.

This is **the same defect as #3537**, in a second service, shipped one day later. A decoupled
challenge (`PUSH_NOTIFICATION` / `BIOMETRIC`) — what an owner approving on their phone actually
produces — sits at `PENDING` while already holding a signature-verified device decision.
`sca-service.verify()` promotes it, and nothing a customer can reach calls `verify`: customer-edge
exposes create / read / decision only, and `decision` records the signed decision *without*
promoting the challenge. The pre-check therefore rejected exactly the challenges the flow depends
on. Every owner approval failed with "is not completed".

**Why no test caught it:** `StubScaChallengeClient` answered `"COMPLETED"`, and every unit fixture
did the same. A fixture reporting a state the real caller cannot be in does not test the flow, it
hides it. Both are now `PENDING`.

### Second, independent defect: the challenge was never spent

`verifyOwnerSca` only ever *read* the challenge — there was no `consume`, and account-service's SCA
client had no `consume` method at all. So one approved challenge authorised an unbounded number of
decisions: approve a 100 CZK proposal, then reuse the same `scaSessionId` to approve every other
PENDING proposal on the account. Single-use is the entire point of a second factor. This would have
survived the first fix on its own, which is why it is called out separately.

The fix is `consume`, matching delegation-service: it resolves a pending decoupled challenge itself,
refuses one never approved or already spent (409), checks the party it is *told* to expect, and
enforces dynamic linking. Approval is still enforced — by the component that owns it. Purpose is
still checked caller-side, because consume is not told the purpose.

Also added `@JsonIgnoreProperties(ignoreUnknown = true)` to the client DTO; delegation-service's
copy documents that sca-service's response carries `method`/`expiresAt`/`consumedAt`/attempt
counters, and account-service's four-field copy would have broken the day that DTO grew.

### Third defect: the challenge could not be created at all

`ScaPurpose` in sca-service is a **closed enum**, and `SAVINGS_WITHDRAW_APPROVAL` was not in it.
`CreateScaRequest.purpose` is typed as that enum, so a customer asking for a savings-withdrawal
challenge got a deserialization failure — the ceremony could not even begin. account-service was
comparing a response string against a purpose no challenge could ever carry.

I found this only because I went looking to *falsify* my own claim that the fix was sufficient. It
is the clearest evidence that the first two fixes alone would still have shipped a dead feature.

This PR therefore touches sca-service to add the enum value and its push-notification copy (the
`when` is exhaustive, so the compiler demanded the branch). Purpose is separate from
`DELEGATION_GRANT`/`DELEGATION_ACCEPT` for the same reason those are separate from each other: a
challenge raised for one act must never be spendable as another, and purpose equality enforces it.

**Pre-existing drift I deliberately did NOT fix:** sca-service's `openapi.yaml` lists
`[PAYMENT_AUTHORIZATION, LOGIN, CONSENT, ACCOUNT_CHANGE]` — three of those four have never been
valid `ScaPurpose` values. I added the new value and left the phantoms in place with a comment,
because deleting enum values from a request schema is a breaking spec change and deserves its own
PR rather than riding along in this one.

## Expiry (was entirely absent)

Proposals were immortal — a delegate's request from six months ago was still approvable. Adds
`EXPIRED`, an `expiresAt` window (7 days), `V19`, and a `@Scheduled` sweep.

Two deliberate choices:

- **The decision path reads the window, not the stored status** (`isExpiredAt`). If approval had
  keyed off `status == EXPIRED`, the gap between the window closing and the sweep next running
  would be an open door. The sweep is bookkeeping for the owner's inbox; it is not the boundary.
- **Expiry is checked before the SCA leg**, so a doomed decision does not burn the owner's one-shot
  factor. Returns 409, not 400 — the request is well-formed and the caller entitled.

The sweep is a `suspend fun`. A plain `@Scheduled` method carries no Vert.x context, so `runBlocking`
around reactive Panache throws `HR000068` and the tick silently does nothing (#2148/#2187).

## Both fixes are falsified, not just green

I reverted each fix and confirmed the new tests go red:

| Reverted to | Result |
|---|---|
| `status == "COMPLETED"` pre-check, no consume | **5 tests FAILED**, incl. both headline tests |
| `fun sweep() = runBlocking { … }` | `HR000068` thrown; `a scheduler-dispatched sweep expires a closed-window proposal` **FAILED** |

`SavingsProposalExpirySchedulerVertxContextIT` drives the **real cron** (shrunk to 2s via a
`@TestProfile`) rather than calling `sweep()`, because a direct call supplies the very context the
scheduler lacks and passes against broken code.

## Service boundary

**The proposal lives in account-service; the ceremony stays in sca-service; money movement stays out
of this PR entirely.** A savings goal is account metadata (ADR-0153) and the `SAVINGS_GOAL`
delegation resource id is the owning account's id by convention, so the proposal is account-service's
data and its guard already lives there. Approval is a *reference* to an sca-service challenge that
account-service consumes but does not own.

---

# Critical assessment of my own design

Reviews that only list strengths are failed reviews. Here is what is weak.

### 1. The execution leg does not exist, and this PR does not add it

**This is the biggest gap and I chose not to close it.** Nothing anywhere consumes
`SavingsWithdrawalApproved` — zero hits fleet-wide, verified. So the end state of this flow is an
event on Kafka and a row reading `APPROVED`. **No money moves.** An owner who approves a withdrawal
gets a UI that says approved and a balance that never changes.

I did not build it because account-service holds no money: a savings "withdrawal" is an ordinary
payment out of the account, and executing it here — via the `WelcomeBonusPort` transaction client
that happens to be injectable — would put debit initiation in a service that is not on the money
path, bypassing screening, limits and the payment path's own idempotency. That consumer belongs in
the payment path and needs its own threat model and 2 approvals (money-path rule).

That is a defensible boundary and still leaves the feature **not end to end**. A reader seeing
"propose-only savings flow" merged would reasonably assume money moves. It does not. This needs a
follow-up issue and I would not describe AC8 as done.

### 2. No customer can reach this flow — customer-edge has no route for it

`grep` for `savings-goal/delegation/proposals` across `openbank-customer-edge/src` and
`openbank-admin-ui/src` returns **nothing**. The endpoints exist only on account-service, reachable
by an operator or M2M caller directly.

This makes the `X-Customer-Party-Id` contract aspirational: `SavingsProposalResource` documents that
header as "stamped by customer-edge", and customer-edge does not proxy these paths, so today nothing
stamps it. The header is still the right design — it is the #3164 C3 fix — but it currently
describes an integration that does not exist.

So the honest status after this PR: the flow is **correct where it exists and reachable in
principle**, not **deliverable to a customer**. Three separate pieces are missing — edge routes, a
UI, and the execution leg — and none of them are in this PR.

### 3. Unverified: the real sca-service `consume` contract

Every test here uses a stub or a mock. I mirrored delegation-service's client shape and its
`ConsumeScaRequest`, but I **never called a running sca-service**, and there is no consumer pact for
this pair. Per this repo's own hard-won rule, a consumer pact alone cannot catch a wrong path —
only a provider replay can, and there is neither. If `/api/v1/sca/challenges/{id}/consume` has moved
or its request body differs, everything here still passes and the flow is dead again in exactly the
way it was dead before. **This is the single most likely place this PR is still wrong**, and it is
uncomfortably close to the class of defect it fixes.

I checked the purpose question rather than leaving it as a suspicion, and it was real — see the
third defect above. That one is now fixed, but note what it implies: my test suite was fully green
*before* I found it, and would have stayed green if I had not gone looking. The stub cannot tell me
the enum is missing. The same blind spot applies to the request/response shape of `consume`, which
I have **not** verified against a running service.

### 4. Failure modes I did not cover

- **`consume` succeeds, then the DB write fails.** The challenge is spent, the proposal stays
  PENDING, and the owner cannot retry — the second factor is gone. There is no compensation. This
  is a real dead-end I left in place; delegation-service has the same shape, which is a reason to
  think it is conventional here, not a reason to think it is correct.
- **Two concurrent `decide` calls on one proposal.** No optimistic locking / version column.
  `approve()` guards on `status == PENDING` in memory, but two requests can both read PENDING. The
  `ApprovalStore` and `consume`'s single-use probably serialise this in practice — *probably* is
  doing real work in that sentence and I did not test it.
- **No amount validation against the account balance or the savings goal.** A delegate can propose
  a withdrawal of any positive amount. Nothing checks funds until the (nonexistent) execution leg.
- **The `PROPOSAL_TTL` of 7 days is invented.** No ADR or product input; it is a hardcoded constant,
  not configurable per-tenant.
- **No event on expiry or rejection.** Only approval emits. A UI cannot react to a proposal expiring
  without polling.
- **`V19`'s backfill sets `NOW() + 7 days` for existing rows** rather than `created_at + 7 days`, to
  avoid retroactively killing live rows. That is a judgement call that gives old proposals a fresh
  window — arguably wrong, and there is realistically no production data yet, so it is untested
  either way.

### 5. What a reader would wrongly assume works

- That approving a proposal **moves money**. It does not — see (1).
- That a **customer can use any of this**. They cannot — see (2).
- That the flow is **verified against sca-service**. It is verified against a stub I wrote, whose
  fidelity is my assumption — see (3).
- That `the sweep leaves an open-window proposal alone` is a real detector. It is a **negative
  control**: it passed under the broken scheduler too, because a sweep that never runs also never
  touches it. Only the other test discriminates.
- That the delegate cannot self-approve because of the owner check. The owner check is one layer;
  `ApprovalStore`'s segregation of duties is the other, and I did not re-verify the store's
  maker/checker enforcement — I inherited that from #3143 and trusted it.

### 6. My new IT adds a Quarkus boot, and the suite is close to its heap ceiling

On a heavily loaded machine (four parallel agent builds) the combined run failed with
`OutOfMemoryError: Java heap space` during Quarkus deployment — the second test in my new IT
reported "Failed to start quarkus". It passes standalone, and the failure is at CDI container boot,
not in any assertion, so the proximate cause is contention rather than this code.

But it is not purely someone else's problem: `@TestProfile` forces a full Quarkus restart, and this
PR takes account-service from two such classes to three. Nothing in the repo sets `maxHeapSize` for
test JVMs, so the suite runs on the default and I have moved it closer to the edge. I have not
raised the heap here — that is a fleet-wide build change and does not belong in a behaviour PR —
but a reviewer should know this suite may now flake under load, and that the honest fix is a heap
setting rather than deleting the test.

### 7. Scope note

`SavingsProposalIT` (the E2E from #3143) now genuinely proves the fix, since the stub it drives
reports `PENDING`. That was free and is the strongest evidence here — but it is one happy path.

---


JiRaska:
Review: 6 findings, 1 blocking

The two defects this PR names are real and the fixes are correct — accepting `PENDING` does **not** weaken the second factor. I traced `consume` on live `origin/main`: `ScaService.consume` refuses a PENDING challenge unless it is decoupled (PUSH/BIOMETRIC), and promotes it only via `verifyDecoupled`, which returns the challenge untouched unless `decisionStore` holds a signature-verified `APPROVED` device decision — so a challenge with **no** decision at all falls through to `status != COMPLETED` → `ScaChallengeNotApprovedException`. Party, `consumedAt`, and dynamic linking are all checked before the spend, and the spend itself is `update "consumedAt = ?1 where id = ?2 and consumedAt is null"` (`ScaChallengeRepositoryImpl.kt:49`) — a real atomic CAS, so two concurrent approvals on one challenge cannot both win. No TOCTOU there.

### Blocking

1. **`openbank-account-service/.../infrastructure/client/ScaChallengeClient.kt:286` (`@RegisterRestClient(configKey = "sca-service")`) registers no OIDC filter, so the new `consume` call — and the existing `getChallenge` — cannot authenticate.** sca-service gates both routes: `ScaResource.kt:253` `@RolesAllowed("ROLE_API","ROLE_OPERATOR","ROLE_ADMIN")` on `GET /challenges/{id}`, and `ScaResource.kt:362` the same on `POST /challenges/{id}/consume`. Every sibling client in the same package registers the filter — `SanctionsServiceClient.kt:16`, and likewise `ProductCatalogClient.kt` / `TransactionServiceClient.kt` — `@RegisterProvider(OidcClientRequestReactiveFilter::class)`, backed by the `quarkus.oidc-client` M2M block already configured in `application.yaml:65-73` ("required by transaction-service"). `ScaServiceRestClient` has neither that annotation nor any header param.
   Concrete: owner POSTs the decide with a genuinely device-approved challenge → account-service calls `POST /api/v1/sca/challenges/{id}/consume` with no `Authorization` header → **401** → caught by the `catch (e: Exception)` in `verifyOwnerSca` → `ProposalScaException` → **400 "SCA challenge … could not be consumed"**. The approval still cannot succeed in any deployed environment. No test can see it: `StubScaChallengeClient` replaces the adapter and never crosses the wire, so the whole HTTP/auth leg is unexercised — the same shape as the `COMPLETED` fixture this PR is fixing. Please either add the provider or state why the token is expected to arrive some other way.

### Non-blocking

2. **`SavingsProposalService.kt:180-184` catches `Exception` around consume and `ProposalExceptionMappers.kt:41` maps every one to 400.** A `TimeoutException` or `CircuitBreakerOpenException` (both configured on `consumeChallenge`) is an infrastructure failure, not a bad request — and a timeout that fires *after* sca-service committed `consumedAt` burns the owner's one-shot factor while returning 400. Correct to not `@Retry`; consider a distinct exception → 503 for the infra cases so the client can tell "your factor was rejected" from "sca-service is unreachable".
3. **Consume-then-transition ordering is fail-closed but has no compensation or record.** `verifyOwnerSca` spends the challenge before `approvalStore.decide` and `proposalRepository.save`. If either then fails, the challenge is gone and the proposal stays PENDING — the owner must raise a fresh challenge. That is the right direction, but the 409 note added to `openapi.yaml` documents only the expiry case; the "spent but not applied" outcome is undocumented and unaudited.
4. **Nothing serialises two approvals of the same *proposal* with two *different* valid challenges.** `decide` is read-modify-write on the proposal row with no version/conditional update, so both callers see PENDING, both consume their own (distinct, valid) challenge, both `approve()`, both merge, and **two** `SavingsWithdrawalApproved` events land in the outbox. Harmless today only because nothing consumes that event — but the spec calls it "the executable instruction for the payments path", so this needs an `@Version` or a status-conditional update before that consumer exists.
5. **`StubScaChallengeClient.kt:59-63`: `consumeCount`, `consumed` and `reset()` are added, the KDoc says "so an IT can prove single-use", and no IT does.** `reset()` has zero call sites and nothing asserts `consumeCount`. Either wire the single-use IT or drop the scaffolding — the PR's own standard is that unexercised test apparatus reads as coverage it isn't.
6. **`SavingsProposalExpiryScheduler.kt:468`: `ConcurrentExecution.SKIP` is JVM-local**, so with >1 replica two pods sweep the same batch (benign — both write EXPIRED), and the `runCatching` in `sweep()` wraps the whole batch, so one failing row abandons the remainder of that tick. Neither corrupts anything; noting for the record.

### Verified clean

`WithdrawalProposalRepositoryImpl` — the id **is** application-assigned (`Ids.newId()`), and both `save` overloads go through `Panache.getSession().flatMap { session.merge(...) }`, never `persist`, so decide/cancel/expire are UPDATEs (the consent-service duplicate-key trap is avoided). `SavingsProposalExpiryScheduler.sweep()` is a `suspend fun`, and `SavingsProposalExpirySchedulerVertxContextIT` drives the **real** cron via a `@TestProfile` (`quarkus.scheduler.enabled=true`, `*/2 * * * * ?`) with literal-only config overrides — not a direct call — plus a negative control; that is the #2148/#2187 shape done right. Expiry is checked before the SCA leg, and `isExpiredAt` reads the window rather than the swept status, so the sweep-lag gap is closed. V19 carries a rollback note, backfills `NOW() + 7 days` (not `created_at + 7d`, which would retro-kill live rows), drops the default afterwards, and its partial index matches the sweep's predicate exactly; `status` is `@Enumerated(STRING)` so `WHERE status = 'PENDING'` is right. `info.version` bumps re-read off live `origin/main` this turn: account 1.7.0 → 1.8.0, sca 1.4.0 → 1.5.0 — both correct and unclaimed. `ProposalResponse` has no `required` block, so adding `expiresAt` and widening the status enum is non-breaking. Threat model updated, and honest about the residual (nothing consumes `SavingsWithdrawalApproved`, no pact on the new edge).

**No already-written data is wrong.** V19 only adds a column with a forward-looking backfill; no existing proposal row is made incorrect by this change, and nothing needs a data-fix migration.

---
Automated agent review — not an approval. account-service is a money-path service (`rules.yaml: money_path_services`), so this needs 2 human approvals and a threat-model review regardless of what this comment says.

JiRaska:
**Review: 3 findings, 0 blocking**

The three claimed defects check out against the code, and the fixes match the claims: the
`status != "COMPLETED"` pre-check is gone, `consumeChallenge` is a real state change on
sca-service, and `SAVINGS_WITHDRAW_APPROVAL` is now in the closed `ScaPurpose` enum so the
challenge can be created at all.

**1. The second factor is single-use but not bound to the proposal (dynamic linking absent).**
`SavingsProposalService.verifyOwnerSca` + `ConsumeScaChallengeRequest(partyId)` — the consume
states only the party, and `ScaPurpose.SAVINGS_WITHDRAW_APPROVAL` carries no `DynamicLinkingData`.
Scenario: the owner has proposals X (100 CZK) and Y (50 000 CZK) both PENDING, opens the app,
approves what they believe is X on their phone, and the client sends that `scaSessionId` with
`proposalId = Y`. It succeeds — party matches, purpose matches, challenge unspent. The amount the
owner saw on the device and the amount that moves are unrelated. Single-use caps the damage at one
misapplied approval; it does not make the approval about the right proposal. The KDoc says this is
tracked separately, which is the right call for this PR, but under PSD2 dynamic linking it is the
substantive remaining gap in the flow and should be named in the follow-up rather than left as a
comment.

**2. Every consume failure collapses to one `ProposalScaException`.**
`verifyOwnerSca` wraps `catch (e: Exception)` around the consume, so sca-service's 409
(already-spent), a wrong-party 4xx, a 2 s timeout and an open circuit breaker are indistinguishable
to the owner and to the operator reading logs. Sharper: with `@Retry` deliberately (and correctly)
omitted, a consume that *succeeded* server-side but exceeded `@Timeout(2000)` burns the challenge
while the proposal stays PENDING — the owner retries, gets 409-as-generic-error, and the only way
forward is a fresh challenge with nothing telling them so. Mapping "already consumed" and
"upstream unavailable" to different responses costs little and is the difference between a
self-service retry and a support call.

**3. The expiry sweep has no cross-replica coordination.**
`SavingsProposalExpiryScheduler` — `ConcurrentExecution.SKIP` is per-instance, so with N replicas
all N run the same `findExpirable(now, 200)` query on the same cron tick and write the same rows.
Harmless as written (`expireStale` uses the single-arg `save`, no outbox row, and `expire()` is a
`check`-guarded PENDING transition, so the loser's write is a no-op or throws inside the same
batch). It stops being harmless the day expiry gains a domain event: that becomes N duplicate
`SavingsWithdrawalExpired` messages with nothing deduplicating them. Worth a comment in the
scheduler at minimum.

Verified clean: `WithdrawalProposalRepositoryImpl.save` goes through
`Panache.getSession().flatMap { session.merge(...) }`, not `persist` — the app-assigned-`@Id`
INSERT-only trap does not apply to the new `EXPIRED` transition. The scheduler is a `suspend fun`
(no `runBlocking`/HR000068), and `SavingsProposalExpirySchedulerVertxContextIT` drives the real
cron from a test profile rather than calling `sweep()` directly, which is the only shape that can
detect that class of bug. Expiry is checked in `decide` *before* the challenge is consumed, so a
doomed decision does not burn the owner's one-shot factor, and `isExpiredAt` reads the window
rather than the swept status, so the sweep is not the boundary. V19 is additive with a sane
backfill (`DEFAULT NOW() + 7 days`, then `DROP DEFAULT`) and a rollback note, and the partial index
matches the sweep's predicate. The `StubScaChallengeClient` change from `COMPLETED` to `PENDING` is
the real fix here — the fixture was the thing hiding the defect. Leaving the three phantom
`ScaPurpose` values in sca-service's spec with a NOTE rather than deleting them is the right call
(deleting request-schema enum values is breaking); openapi bump 1.4.0 -> 1.5.0 is the correct
classification for an added value.

---
Automated agent review — **not an approval**. This touches `openbank-account-service` and
`openbank-sca-service` on an authorization/second-factor path; it needs human review regardless.

PR #3633, opened 2026-08-03
feat: let a delegate pay from a shared account, and audit it as delegated

JiRaska:
## What this does

A delegate can now **pay** from an account shared with them, and every delegated action is
recorded — and queryable — **as delegated**.

Before this PR the delegation stack was complete except at the point where it means anything:
`openbank-delegation-service` was deployed, the grant lifecycle published to
`openbank.delegation.events`, `account-service` built the enforcement projection from it, and
`AuthorizationService.isAuthorizedForAmount` already ANDed the per-transaction ceiling with
`issuedBy(owner)` / `isActiveOn(now)` / `satisfies(role)`.

It had **zero production callers** (re-derived without a pathspec: the only hits on `origin/main`
are the port declaration, the implementation, and one unit test). `onBehalfOf` existed in **no**
`src/main` file fleet-wide — 3 hits, all in `docs/`. So a delegate could not pay, and the
projection was dead code from the money path's point of view.

| | |
|---|---|
| account-service | `authorizeDelegatedPayment` + `GET /api/v1/accounts/{id}/delegation/payment-authorization`, openapi `1.7.0 -> 1.8.0` |
| audit-service | `onBehalfOf`/`delegationId` in the chain + `GET /api/v1/audit/on-behalf-of/{grantorPartyId}`, `V10`, openapi `1.5.0 -> 1.6.0` |
| customer-edge | domestic-payment consults the grant; `GET /customer/v1/delegations/activity`, openapi `1.26.0 -> 1.27.0` |

`Refs #2990` (AC9, AC10), ADR-0232 D3/D5.

---

## 1. Where the authorization check lives, and why

**The decision is in `account-service`. The identity is established in `customer-edge`. Neither
tier does the other's job.**

`account-service` is the only service that holds *both* the delegation enforcement projection
(fed by `DelegationEventConsumer`) *and* the account's true owner. Any other placement needs a
second copy of the projection kept in step with this one — the failure mode this repo keeps
paying for, where two copies of a rule drift and the money-path copy is the stale one. It also
re-evaluates on **every request** rather than trusting a verdict reached once at offer time, which
is what makes a revoked grant stop working immediately.

`customer-edge` contributes the one thing only it can: **who is asking**. The party comes from the
validated `party_id` JWT claim, never from the request body — the fleet's IDOR convention. The
edge then asks, with the amount, and treats anything that is not an explicit `authorized: true`
with a grantor as a refusal.

**Why not `openbank-domestic-payment`?** It runs on an M2M token and does not know the initiating
customer at all — `actorId` there is nullable and frequently null. It has no way to ask the
question. Putting the check there would have meant inventing a way to carry a customer identity
into a service that deliberately does not have one.

**Why a new endpoint rather than widening `/authorizations/check`?** That returns
`{"authorized": bool}` and other callers parse it. This one must return the **grant id and the
grantor**, because a delegated debit is only auditable *as delegated* if the record names the grant
that permitted it — and a grant is revocable, its projection row closed on revoke, so nothing after
the fact can reconstruct which one was live when the money moved. It reuses the `account.read`
Authorize action and the edge-proxy role set already used by `SavingsGoalDelegationResource`, so
**no `rest.rego` change and no policy-bundle restamp** rides along with a money-path change.

**Why not reuse `/entries/by-actor/{actorId}` for the transparency query?** It looks like the same
question and it would have 403'd in production with every test here green. That route requires the
`audit.read` action, which the OPA bundle **deliberately withholds** from the edge principal —
granting it would put regulated evidence behind a service account. The new `/on-behalf-of/` route
carries `audit.customerRead`, whose `edge-service-audit-customer` rule matches on principal id +
action with **no path condition**, so it covers the new path as-is. That reasoning is what makes
the new endpoint correct rather than redundant, and the margin was narrow.

### The flow

```
app  --POST /customer/v1/domestic-payments (JWT: party=delegate)-->  customer-edge
                                                                      |
  GET /accounts/{id}  (X-Customer-Party-Id: delegate)     ------->  account-service -- 404 (ownership guard)
  GET /accounts/{id}/delegation/payment-authorization
      ?partyId=<delegate>&amount=1500.00&currency=CZK     ------->  account-service
                                                          <-------  {authorized, outcome:DELEGATED,
                                                                     delegationId, grantorPartyId}
  GET /accounts/{id}  (X-Customer-Party-Id: GRANTOR)      ------->  account-service -- 200 (IBAN, owner)
  SCA consume  (delegate's own device, unchanged)         ------->  sca-service
  POST /api/v1/domestic-payments (debtorName = GRANTOR)   ------->  domestic-payment
  audit.emit(partyId=delegate, onBehalfOf=grantor, delegationId=..) -->  openbank.customer.audit
```

The re-fetch **as the grantor** deserves calling out. account-service's `X-Customer-Party-Id` guard
is an *ownership* guard — it 404s a delegate by design — so reading the account the delegate was
just authoritatively told they may debit requires asking as its owner. That is not the edge
self-authorizing: the grantor's identity came from the authoritative decision one call earlier and
is never client-supplied. But it does mean a bug in that decision widens into an account read, so
the edge **re-checks that the fetched account's owner is the named grantor** before proceeding.

Two behaviours worth knowing about:

- **`debtorName` is now the account holder's**, not the initiator's. Sending the delegate's name
  would misattribute the transfer on the counterparty's statement and in every downstream AML party
  resolution.
- **SCA is unchanged and belongs to the initiator.** A delegate authenticates as themselves; the
  grant is what makes it their debit to make, not a substitute for a device-signed,
  amount-and-payee-bound challenge. I added no new SCA path, so the #3537
  `status != "COMPLETED"` pre-check trap is not reintroduced — the existing `scaGate` posts
  straight to `/consume` and reads the response family.

### One deliberate narrowing

`isAuthorizedForAmount` **ignores** the legacy `account_authorizations.transaction_limit` — it
returns true on a role match and only ever compares an amount to a *delegation* grant's ceiling.
Harmless while nothing called it; wiring that behaviour to a live debit route would have shipped a
money path on which an operator-set per-transaction ceiling is decoration (a `PAYMENT_ONLY` row
capped at 1 000 CZK would authorise 1 000 000). `authorizeDelegatedPayment` enforces it. Nothing
regresses, because the old method has no callers. A test pins **both** behaviours side by side so
the divergence is a decision on the record, not an accident.

---

## 2. Audit chain evidence

The requirement was to prove the CHAIN entry verifies, not that a field is present. Asserting
`onBehalfOf == grantor` proves nothing about a tamper-evident log — the column could be one a DBA
can `UPDATE` and that assertion would still pass.

`DelegatedActionAuditChainIT` (real Postgres via Testcontainers, event driven through the real
`AuditConsumer`) does three things.

**(a) The chain verifies over the delegated entry.** `verifyChain` recomputes every SHA-256 link
from the row *as the database returns it*:

```
verification.intact    = true
verification.checked   >= 1       // the delegated row was recomputed, not skipped
verification.unchained = 0        // not a pre-V5 row counted-but-unverified
entry.actorId          = <delegate>   // who acted does not change because they were permitted to
entry.onBehalfOf       = <grantor>
entry.delegationId     = <grant>
```

**(b) The falsification — and it found a second control I had not accounted for.** My first attempt
to tamper with the row reported `0 rows updated` and I assumed a broken probe. It was not: `V2`
carries `CREATE RULE no_update_audit AS ON UPDATE TO audit_entries DO INSTEAD NOTHING`, so an
in-place edit is *silently discarded* — no error, zero rows, log unchanged. The test now asserts
that outright (a rule that had been dropped would look identical from the application, and this is
the only assertion anywhere that would notice), then **drops the rule** and performs the edit a
falsified account-sharing story would need:

```
updatePayload(entry, tampered)      -> 0 rows      // control 1: append-only RULE
verifyChain(from = entry)           -> INTACT      // nothing changed

withoutImmutabilityRule { updatePayload(entry, tampered) } -> 1 row
verifyChain(from = entry).intact             = false
verifyChain(from = entry).firstBrokenEntryId = <entry.id>   // names the edited row

restore original bytes -> verifyChain(from = entry).intact = true   // the BREAK was the edit
```

That second half is `V5`'s stated claim — "the chain makes any mutation that bypasses [the rules]
detectable" — which nothing in the repo had ever exercised.

**(c) The transparency query is scoped, not merely non-empty.** Two delegates on one grantor, one
delegate on a *different* grantor, and one direct payment: the grantor's view returns exactly the
first two, narrows correctly by delegate and by grant, and the direct payment never appears.

The edge-side tests were falsified against broken code before being trusted — with
`onBehalfOf` forced to null, 1 test fails; with the authorization fetch stubbed to `null`, 4 fail
(including "a delegate can pay"). Test counts read from the JUnit XML, not the console.

| suite | tests |
|---|---|
| `AuthorizeDelegatedPaymentTest` | 19 |
| `DelegatedActionAuditChainIT` | 4 |
| `DelegatedDomesticPaymentTest` | 13 |
| `CustomerDelegationResourceTest` | 17 (5 new) |
| `CustomerEdgeDelegatedPaymentPactConsumerTest` | 1 (contract) |
| `AccountPactFolderProviderVerificationTest` | 4 (provider replay) |

`detekt ktlintCheck test` green on all three modules (account-service 38 classes, audit-service 13,
customer-edge 27, zero failures).

---

## 2b. Consumer pact — and proof it can fail

The edge reaches account-service through `UpstreamClient`: a plain `java.net.http` client, no
generated stub, no `@Path`, URL assembled by hand. Every other edge test mocks that client, so it
can only ever prove the edge agrees with itself — #2269's defect precisely.

The failure mode here is worse than a broken screen. `fetchDelegatedPaymentDecision` returns `null`
on any non-200, and a null decision is a **refusal**. A wrong path or a misspelled parameter does
not throw: it makes every delegated payment 403 **forever, fail-closed and silent**. Nothing alerts
on a feature that is merely always refused.

`pacts/openbank-customer-edge-openbank-account-service.json` — expected path and query are
**literals**; the outgoing request is produced by calling the real production method against the
Pact mock server, with the real `UpstreamClient` and its real headers (a loopback stub stands in
for the token endpoint, which otherwise swallows into a 502 and the interaction is never
exercised). Deriving both sides from production would be vacuous — they would move together and
the test could never fail (#2290). `fetchDelegatedPaymentDecision` is `internal` for exactly this.

`amount`/`currency` are in the contract deliberately: the per-transaction ceiling is evaluated
**from** the amount, so a pact that omitted it would pass just as happily against a provider that
ignores limits.

Provider replay is `AccountPactFolderProviderVerificationTest` (`@PactFolder`, runs on every PR —
the `@PactBroker` twin does not count, its gate skips off main). The new `@State` seeds **both**
rows the endpoint answers from: the account, which establishes the owner, and the projection grant,
which `issuedBy(owner)` is checked against — seeding one alone silently yields `NO_GRANT`. Ceiling
5000.00 against the contract's 1500.00, so "compared" is distinguishable from "not compared".
Mirrored verbatim into the broker twin per #3520.

Gates: `check-pact-provider-replay.py` reports **41/41** replayed with `KNOWN_UNCOVERED` still
empty; `derive-pact-drift-scope.sh` now derives `:openbank-customer-edge:test` into the
regeneration scope on its own, and regeneration is byte-stable.

### Falsification

Verdicts read from the **JUnit XML** — pact-jvm prints "Not all of the N were verified" on a fully
green run, so the console cannot be trusted here.

| # | break | result |
|---|---|---|
| baseline | — | provider `tests="4" failures="0"` |
| A | contracted path -> `payment-authorisation` | provider **RED**, `failures="1"`: `expected status of 200 but...` (404) |
| B | `partyId` renamed to `partyIdentifier` in the pact | provider **RED**: *`Actual map is missing the following keys: delegationId, grantorPartyId`* |
| C | `partyId` typo'd in **production** code | consumer **RED**: `the edge's request must match the contracted path and query — Expecting actual not to be null` |

**B is the one worth reading.** The provider did not error — it answered a **different decision**.
With the parameter name wrong, `@QueryParam("partyId")` binds null, the endpoint computes
`NO_GRANT`, returns a perfectly valid 200, and the body simply lacks `delegationId`/
`grantorPartyId`. That is the silent-refusal failure mode this contract exists to catch, caught.

All three restored; both modules re-verified green afterwards (account-service 38 classes,
customer-edge 28, zero failures).

---

## 3. Self-review — what I would not accept from someone else

**This is one rail, not "the payment path".** Only `POST /customer/v1/domestic-payments` consults
the grant. **SEPA, SCT Inst and SWIFT are untouched and remain owner-only**, as are standing
orders, SDD mandate creation and pocket transfers. A reader skimming the title will assume
delegated payments work; they work for domestic CZK and nothing else. That is a scope decision I
made, not an oversight — but the right follow-up is to extract `resolveDebitAuthority` into a
shared helper and apply it to every rail, because leaving four near-identical debit guards in one
file is how they drift.

**Nothing here has run against a real deployment.** No end-to-end proof exists that a real
delegate with a real grant can complete a real payment in sandbox. The pact in section 2b closes
the do-the-two-artefacts-agree question, not the does-it-work-in-production one.

~~**No consumer pact between customer-edge and account-service.**~~ **CLOSED — section 2b.** This
was the gap I said should block the PR. It is now a committed contract with a `@PactFolder`
provider replay that runs on every PR, falsified three ways.

**The audit link is a convention, not a contract.** `customer-edge` writes `onBehalfOf` into a
flattened detail map; `audit-service` reads a top-level field of that name. Two string literals in
two services with nothing asserting they match. The IT builds the event JSON by hand *deliberately*
— a shared fixture would let both halves drift together — but "this test goes red if the edge
renames the field" is only true because I wrote both halves today. A schema, or a pact, would make
it true tomorrow.

**The audit trail is best-effort by design, and now carries the only record of a delegation.**
`EdgeAuditPublisher` catches everything and degrades a Kafka outage to an ERROR log — correct for
audit-must-not-break-the-payment, but it means a broker outage produces a **successful delegated
payment with no record that it was delegated**. The payment is durable; the on-behalf-of fact is
not. Nothing alerts on that gap. For a transparency feature a regulator would read as an
accountability control, "best effort" is the wrong durability class, and I did not change it.

**Two round trips added to the refused path, one to the delegated path.** The owner path is
unchanged (one call, as before). A non-owner costs an extra authorization call, and a delegate an
extra account fetch. Not measured under load; there is no circuit breaker and no cache on the
authorization call, so account-service becoming slow now makes delegated payments slow. Fail-closed
means it also makes them *fail*, which is right, but the blast radius grew.

**The projection is eventually consistent and I did not narrow that window.** A revoked grant stops
authorising only once `DelegationEventConsumer` has processed `DelegationRevoked`. A grantor who
revokes in the app and watches a payment succeed a second later is seeing correct-by-design
behaviour that will read to them as a broken revoke. There is no metric on projection lag.

**`V10`'s columns are outside the chain hash.** Consistent with `V9` and documented in the migration
itself, and the underlying facts *are* chained inside `payload` — but it does mean an `UPDATE` of
`on_behalf_of` alone silently changes what the grantor's transparency view shows while
`verifyChain` still reports INTACT. The append-only RULE is the control there, and I now know from
(b) above that it works. A reader who assumes "it's in the audit chain, therefore every column is
tamper-evident" would be wrong, which is why the migration says so in place.

**The refusal classification is one careless line from becoming an oracle.** `outcome` distinguishes
NO_GRANT from LIMIT_EXCEEDED from ACCOUNT_NOT_FOUND, and the *only* thing stopping that from
enumerating other people's accounts and sharing arrangements is the edge collapsing all of them to
one 403. A test pins it today. A future caller that surfaces `outcome` verbatim reintroduces the
hole with no gate to stop it — this deserves a check, and does not have one.

~~**Unverified: the `audit.customerRead` rego grant.**~~ **CLOSED by review** — the bundle's
`edge-service-audit-customer` rule is action-scoped with no path condition, so the new route is
covered. See section 1 for why reusing the existing `/entries/by-actor/` route would instead have
403'd in production while every test here stayed green.

Also unverified: whether the app or admin-ui actually consume the response shapes I chose, and
whether `LEGACY_AUTHORIZATION` rows carrying a `transaction_limit` exist in any real database — if
they do, the narrowing above is a **behaviour change for someone**, not merely a hole not created.

---

Threat models updated for both affected services
(`docs/threat-models/openbank-account-service.md`, `openbank-customer-edge.md`).
`account-service` is a `money_path_services` entry, so this needs 2 approvals.

The pact gap that originally blocked this is closed (section 2b). What remains open and worth a
reviewer's judgement: no sandbox end-to-end run, the other payment rails still owner-only, and the
best-effort durability of the on-behalf-of record.


JiRaska:
**Review: 3 findings, 0 blocking on the authorization model; 1 functional always-refuse bug.**

The server-side authorization story holds up. `AuthorizationService.authorizeDelegatedPayment`
resolves the owner from `accountRepository.findById(accountId).partyId` (never from the request),
and every grant must pass `issuedBy(account.partyId)` + `isActiveOn(now)` + `satisfies(PAYMENT_ONLY)`
+ per-transaction ceiling, on top of a DB-level `active = true and resourceType = 'ACCOUNT'` filter
in `DelegationProjectionRepositoryImpl:44-51`. The grantee identity comes from the edge's validated
`party_id` claim; the edge treats any non-`authorized`/missing-grantor answer as a refusal and
re-verifies `extractOwnerPartyId(ownerJson) == grantor` before debiting. Revoked and expired grants
stop authorizing (`active=false` / `validTo`), subject to the projection lag the description names.

**1. `LEGACY_AUTHORIZATION` short-circuits the delegation branch, so some delegates can never pay
(functional, fail-closed).**
`openbank-account-service/.../usecase/AuthorizationService.kt` — the legacy
`account_authorizations` branch `return`s before the delegation projection is consulted, and that
outcome carries no `grantorPartyId`. In `CustomerEdgeResource.resolveDebitAuthority` the guard is
`if (decision?.authorized != true || grantor == null) -> refuse`. So: party P holds a legacy
`PAYMENT_ONLY` row on account A (no limit) **and** an active `ACCOUNT_INITIATE_PAYMENT` delegation
grant → account-service answers `authorized: true, outcome: LEGACY_AUTHORIZATION` with no grantor →
the edge 403s every payment, permanently, with `delegationOutcome=LEGACY_AUTHORIZATION` in the audit
and nothing anywhere reading as an error. Safe direction, but it is a capability that silently never
works. Evaluating the delegation branch first (or falling through when legacy yields no grantor)
fixes it.

**2. The grant model has only a per-transaction ceiling — no cumulative or velocity cap.**
`DelegatedAccessGrant.withinPerTransactionLimit` is the only amount control, and
`authorizeDelegatedPayment` is stateless per call. A delegate holding a 1 000 CZK ceiling can
initiate an unbounded number of 1 000 CZK domestic payments and drain the account; each one is
individually authorized, individually SCA'd, and individually correct. That may be the intended
scope of ADR-0232, but it is the property a grantor will assume they set a limit against, so it
wants an explicit human decision rather than inheriting the projection's shape.

**3. Revocation is only as fast as `DelegationEventConsumer`, and there is no lag signal.**
Named in the description, repeating it because it is the money-path half: between `DelegationRevoked`
being published and the projection row closing, a payment still authorizes, and nothing measures that
window. Related, minor: because the check re-runs on every attempt, a client retry with the same
`Idempotency-Key` after a revoke is refused at the edge before it can replay the original
domestic-payment result.

Verified clean: grantee identity is never taken from the request body; `outcome` never reaches the
customer wire (single opaque 403 for NO_GRANT / LIMIT_EXCEEDED / ACCOUNT_NOT_FOUND / not-yours);
`debtorName` correctly switches to the account holder; SCA stays bound to the initiator; the
consumer pact's expected path/query are literals while the request is reflected off the real
production method (`fetchDelegatedPaymentDecision`), which is the asymmetry that makes it able to
fail; `Idempotency-Key` forwarding on the domestic rail is unchanged.

---
Automated agent review — **not an approval**. `openbank-account-service` is a money-path service;
this needs its two human approvals regardless of what this comment says.

JiRaska:
Review: 3 findings, 0 blocking

Traced the authorization path end to end. The claim matches the code: the decision is made in account-service on every request, the acting party comes from the JWT, and the edge fails closed.

**1. No cumulative / velocity ceiling — only per-transaction. `openbank-account-service/.../AuthorizationService.kt` (`authorizeDelegatedPayment`, the `withinPerTransactionLimit` disjunct).** A grant capped at 5 000 CZK authorises an unbounded number of 4 999 CZK debits: 20 payments -> 99 980 CZK moved, every one a correct `DELEGATED` decision, and the grantor's only signal is the transparency view after the fact. Sharpening it: `AccountAuthorization.dailyLimit` already exists, is persisted (`AuthorizationService.kt:42`, `AccountAuthorizationEntity.kt:34/37/101`) and is returned by the authorizations API (`AuthorizationDtos.kt:16/34`) — and **nothing in any `src/main` reads it**, before or after this PR. So an operator who sets a daily limit today gets a stored number with no enforcement, and this PR is what turns the surrounding row into a live debit authority. Not blocking (the delegation grant model has no cumulative field at all, so this is a delegation-service change, not a one-line fix here) but it belongs in the threat model as a stated residual, not omitted.

**2. `LEGACY_AUTHORIZATION` is authorised by the provider and refused by its only consumer.** `authorizeDelegatedPayment` returns `authorized = true` with `grantorPartyId = null` for a legacy `account_authorizations` row (`AccountPorts.kt`, `DelegatedPaymentDecision.authorized`), while `CustomerEdgeResource.resolveDebitAuthority` refuses on `grantor == null`. Concretely: an operator grants party B a PAYMENT_ONLY row on A's account, B pays -> account-service says yes, the edge answers 403 "Debtor account does not belong to caller" and audits `delegationOutcome=LEGACY_AUTHORIZATION / DENIED`. Safe direction, and I'd keep it — but the endpoint's own OpenAPI describes `authorized` as "may debit", so the next consumer that honours it inherits a debit path on which `dailyLimit` (finding 1) is unenforced. Worth either dropping LEGACY out of `authorized` or saying in the spec that a null `grantorPartyId` is not a debit permission.

**3. Best-effort audit now carries the only record that a payment was delegated.** `EdgeAuditPublisher.emit` catches everything and degrades a Kafka outage to an ERROR log, so a broker outage yields a durable delegated payment with no `onBehalfOf` anywhere and nothing alerting on the gap. The PR states this; flagging it because "delegated" is the fact a regulator would ask for, and best-effort is the wrong durability class for it. Follow-up issue, not a merge blocker.

Verified clean: the acting party is server-side only — `customer()` resolves from the validated `party_id` claim and `partyId` never comes from the body/header (`createDomesticPayment` -> `resolveDebitAuthority(customer, ...)`), so no IDOR. `DelegatedPaymentAuthorizationResource` is gated `@RolesAllowed(ROLE_API, ROLE_VIEWER, ROLE_OPERATOR, ROLE_ADMIN)` + `@Authorize(action="account.read")`; `Roles.kt` contains no customer role, so a `ROLE_CUSTOMER` token cannot reach it to self-authorize, and `rest.rego`'s `operator-read-any` (verb `read`) covers the edge principal. Ownership + grant are re-evaluated on every call — `issuedBy(account.partyId) && isActiveOn(now) && satisfies(PAYMENT_ONLY)`, against `findActiveByAccountAndParty`, which also filters `active = true` and `resourceType = ACCOUNT`; a revoked grant's row is closed by `closeById` (`update ... active = false`) and the projection repo uses `session.merge`, not `persist`, so the app-assigned-id trap does not apply. The revocation window is the consumer lag only, and is documented. Currency is checked on both ceilings — `DelegatedAccessGrant.withinPerTransactionLimit` refuses on `perTransactionLimitCurrency != currency`, and the new `withinLegacyTransactionLimit` mirrors it. Fail-open: `fetchDelegatedPaymentDecision` returns null on any non-200 or unparseable body and the caller treats null as a refusal — covered by `an unavailable authorization service refuses the payment` and `an unparseable authorization answer refuses the payment`, plus `authorized without a grantor is treated as a refusal`. Attribution is non-repudiable in both directions: `actorId` stays the delegate, `onBehalfOf` is the grantor, `EdgeAuditPublisher` flattens details top-level and drops nulls so `AuditConsumer`'s `node["onBehalfOf"]` reads the real value and a direct payment emits neither key (asserted both ways in `DelegatedDomesticPaymentTest`); the V10 columns sit outside the chain hash but the facts are inside the hashed `payload`, which the migration states in place. `audit.customerRead` is action-scoped with no path condition in `rest.rego:316`, so `/on-behalf-of/` is covered without a bundle restamp. Pact: expected path and query are **literals** in `CustomerEdgeDelegatedPaymentPactConsumerTest` while the request is reflected off production `fetchDelegatedPaymentDecision`, and `AccountPactFolderProviderVerificationTest` (`@PactFolder`, ungated, runs on a PR) seeds both the account and the grant with a ceiling deliberately above the contracted amount. `openbank.edge.audit-service-url` exists in customer-edge `application.yaml:109`. openapi bumps read off live `origin/main`: account 1.7.0 -> 1.8.0, audit 1.5.0 -> 1.6.0, edge 1.26.0 -> 1.27.0 — all three numbers still free. No new outbox entity, no new `@Scheduled`, no edited applied migration. No already-written data is wrong: V10 adds nullable columns and a partial index only, so nothing needs a backfill or correction.

Automated agent review, not an approval. account-service is a `money_path_services` entry — this needs its two human approvals regardless of anything above.

Issue #3345, opened 2026-08-02
Liveness, not readiness: last-successful-run metric + staleness alert for every scheduler and external feed


JiRaska:
Part of #3343 (operational maturity tracker, level 1 → 2).

## What

For every `@Scheduled` job and every external feed: a "last successful run" metric plus an alert on its staleness. Extend `external-feed-watch.yml` into a general scheduler heartbeat.

## Why

- 5 `@Scheduled` methods never ran (HR000068), 3 of them money-path — and no alert noticed.
- The ČNB feed was 404 for the entire life of fx-service while revaluation reported `posted = false` — a job that "ran successfully" while doing nothing.

A green readiness probe says the process is up; it says nothing about whether the scheduled work inside it ever executes. This alert pair would have caught both on the day of deployment.

## Acceptance criteria

- Every `@Scheduled` method emits a last-success timestamp metric (shared library support preferred over per-service hand-rolling).
- Every external feed fetch records last-success; a 404/parse failure marks the run failed.
- Alerts fire on staleness beyond the job's own period (with a grace multiplier), not on process health.
- `external-feed-watch.yml` extended (or superseded) so scheduler heartbeat and feed watch are one mechanism.

JiRaska:
Status update: non-money-path slice is now in PR #3735 (aml, analytics-sink, authz-policy-auditor, docs-truth-agent, finops-agent) with per-service liveness tests and baseline shrink. Local gate passed: detekt + ktlintCheck + koverVerify + build for all five services on JDK 25. Money-path slices remain in the previously opened PRs (#3647, #3703-#3707) and still need owner review; leaving this issue open until those and the remaining baseline entries land.

JiRaska:
Second non-money-path batch is open as PR #3739 (devops-agent, flaky-test-hunter, governance-auditor, release-steward, security-scanner) with the same liveness test pattern and local gate green. PR #3735 remains for human review because the high-risk path guard matched openbank-aml-service; auto-merge was attempted and refused by the guard.

JiRaska:
Coverage inventory measured against `origin/main` today (`python3 .github/scripts/check-scheduler-liveness.py --root .`), so the remaining work is a list rather than a re-survey.

**Gate output on main:** `domain @Scheduled files: 42 (10 with liveness, 64 exempt outbox/allowlist)`, 32 baselined as adoption-pending.

**Of those 32, 16 are already in an open PR:**

| PR | Baselined files it clears |
|---|---|
[...]

Issue #2239, opened 2026-07-25
The #2187 schedulers still have no runtime liveness signal — 4 of 5 emit no gauge, and mechanism 3's paging rule does not exist


JiRaska:
## Why this exists

#2187 fixed five `@Scheduled` jobs that had silently never run, and #2197 added a CI guard so the *code shape* cannot come back. Neither answers the question the incident actually raised: **what would have told us at runtime?** Nothing did, for months, and after both PRs the answer is still very nearly nothing.

PR #2238 fixes one link in that chain (the sentinel queried a gauge name nobody emits). The rest of the chain is still broken, and the two remaining pieces need decisions rather than a typo fix.

## Gap 1 — four of the five repaired jobs emit no liveness gauge at all

`DomainMetrics.registerWorkflowLiveness` (ADR-0160 mechanism 3) exists precisely for this and has **one** adopter fleet-wide:

| job | service | money-path | liveness gauge |
|---|---|---|---|
| `StandingOrderExecutionScheduler.sweep` | standing-order | | ✅ `standing-order-execution` |
| `TieOutScheduler.runTieOut` | ledger | yes | ❌ none |
| `FxRevaluationScheduler.revalueDaily` | ledger | yes | ❌ none |
| `CnbRateIngestionScheduler.ingestDailyFixing` | fx | yes | ❌ none |
| `BillingCycleScheduler.sweep` | billing | yes | ❌ none |

What exists instead, and why it is not equivalent:

- `TieOutFreshnessWatchdog` measures **data** freshness (`ledger_tieout_runs.runAt`, stale after 25h) and would in fact have caught a dead `runTieOut` — but it is **log-only**: it registers no meter, and its KDoc's "pages via Alloy → Loki" is aspirational, since there is no Loki alerting rule anywhere in the repo. It is also itself a `@Scheduled` method in the same JVM, so a scheduler-subsystem failure takes the watchdog and the job together.
- `SubledgerTieOutBreak` (`prometheus-rules-tieout.yaml`) is `increase(openbank_subledger_tieout_break_total[25h]) > 0` — an **increase** rule. A job that never runs produces no increase, so it stays silent. It alerts on the control finding a break, never on the control being absent.
- FX revaluation, ČNB ingestion and the billing cycle have no metric, no watchdog and no rule of any kind. Success and failure both end in a log line.
[...]

PR #3634, opened 2026-08-03
fix(libs-runtime): seed the workflow-liveness age gauge at registration, not EPOCH

JiRaska:
## Summary

Seeds the workflow-liveness age gauge at registration time instead of `Instant.EPOCH`. This is the decision ADR-0237 made for #2239 Gap 2 — a prerequisite for the staleness PrometheusRule: with the EPOCH seed, every fresh pod reads as ~56 years stale until the job's first success, so the rule would fire for up to 24h after every deploy of a daily job.

## Type of change

- [ ] feat
- [x] fix (bug fix)
- [ ] refactor / [ ] perf / [ ] docs / [ ] chore / [ ] security / [ ] breaking change

## Linked issues

Refs #3345, Refs #2239 (Gap 2), ADR-0237

## Changes

- `DomainMetrics.registerWorkflowLiveness` — seed `lastSuccessEpochMillis` from `Instant.now()`; KDoc rewritten to record the ADR-0237 decision and the residual restart-hides-staleness edge (covered by existing availability alerts).
- `DomainMetricsTest` — the `never-succeeded reads as maximally stale` test replaced by `never-succeeded reads only as stale as the pod is old`; "never ran" detection moves to the alert layer (`absent()` + staleness after one grace period).

## Definition of Done

- [x] Unit tests updated; `:openbank-libs-runtime:test --tests DomainMetricsTest` green (JDK 25)
- [x] `ktlintCheck` + `detekt` green on the module
- [ ] Integration/contract tests — N/A (shared metric primitive, no API)
- [x] No new lint warnings

## Security checklist

- [x] No secrets/PII; no new dependency; no auth/crypto/payment code.

## Compliance impact

- [x] None

## Rollout / Rollback

- Deployment: rolls out with each service's next build; semantics change is gauge-only, no persistence.
- Rollback: revert; the only consumer-side effect is the gauge reading high again until first success (the state the staleness rule will tolerate via `for:` once it lands).
- Data migration: N/A

## Reviewer notes

Behavior change is deliberate and narrow: a never-succeeded workflow now reads age≈pod-uptime instead of age≈56y. Consumers: control-liveness-sentinel (ADR-0163 — its findings get the same boot-window noise removal) and the upcoming `openbank.workflow.liveness` PrometheusRule group. The restart-hides-staleness edge (pod restarted more often than 2× interval masks a never-running job) is documented in the KDoc and covered by crashloop/tier1 availability alerts.


JiRaska:
Merged `main` in — this was **143 commits behind**, and its only real red was a stale baseline rather
than anything about this change.

`Trivy` was failing on `aws/envs/sandbox-platform/ecr-service-repositories.tf` —
`AWS-0031 (HIGH): Repository tags are mutable`. That finding belongs to `main`, not to this branch:
`main` baselines `AVD-AWS-0031` in `.trivyignore` with a documented, load-bearing reason (auto-deploy
tags `sandbox-<sha>` and the reconcile path legitimately re-pushes a sha it has already built, which
IMMUTABLE would break). This branch predated the entry.

```
AVD-AWS-0031 in .trivyignore   before: 0   after: 2
```

**41 open PRs are in the same state**, so this is not specific to this one.

Merge verified for silent deletions, since a clean merge can drop a neighbouring list entry with
nothing printed:

```
conflicts                        0
scope vs origin/main             3 files — DomainMetrics.kt, DomainMetricsTest.kt,
                                 LedgerWorkflowLivenessTest.kt (exactly this fix, nothing else)
.release-please-manifest.json    56 entries (unchanged)
.github/gates/gates.yaml         110 ids (unchanged)
the fix itself                   still present in DomainMetrics.kt
```

Gate re-run with `--rerun-tasks` so nothing could report `UP-TO-DATE` without executing —
`:openbank-libs-runtime:ktlintCheck`, `:detekt` and `:test` all visibly ran, BUILD SUCCESSFUL in 34s.

Auto-merge deliberately **not** armed: this changes shared `openbank-libs-runtime` behaviour that
every service's liveness gauge depends on, and it is the fix for the four-state gauge problem
(`Instant.EPOCH` seeding made a fresh pod read as decades stale). That is worth a human look rather
than a merge on green.

Context for a reviewer, from an inventory taken today: the gate reports **42 domain `@Scheduled`
files, 10 with liveness, 64 exempt, 32 adoption-pending** — 16 of the pending already have open PRs.
Until this merges, "deployed but never run" and "ran and is now stale" are not distinguishable, which
is the state the whole ADR-0237 heartbeat exists to remove.

JiRaska:

The seed-at-registration change is falsified in both directions, which is the thing to check for a metrics/gate PR like this:
- `LedgerWorkflowLivenessTest`'s never-run case now asserts age `< TOLERANCE_SECONDS` (was previously asserting the old EPOCH behaviour — the test genuinely reddens against pre-change code).
- The failed-run case adds `Thread.sleep(1100)` before the failing run and asserts age `>= 1.0`, i.e. it proves a swallowed failure does NOT reset the clock to ~0 — the actual failure mode the KDoc says this metric exists to catch. That's a real fed-input, not just an exit-code check.

Verified the referenced infrastructure actually exists on `origin/main` rather than being aspirational: `docs/adr/0237-scheduler-feed-liveness-heartbeat.md` is present, and `openbank-infra/gitops/components/observability/prometheus-rules-workflow-liveness.yaml` exists and references `openbank.workflow.liveness` — so the KDoc's claim that a PrometheusRule now consumes this gauge (superseding the old "no PrometheusRule and no page" caveat) checks out, not a dangling reference.

Verified: the KDoc's documented consumers (ADR-0163 sentinel + ADR-0237 PrometheusRule) exist on main; test assertions were compared against the diff to confirm they'd fail pre-change. Did not independently re-derive the PrometheusRule's alert expression.

JiRaska:
## Review: 1 blocking finding — the red is real and it is this PR's to fix

I swept the queue for PRs whose red is a stale artifact (a gate that no longer fails on `main`, on a branch far behind). **This one is not that.** The failure reproduces from the PR's own change:

```
DevOpsWorkflowLivenessTest > failed dispatch leaves liveness old()                       FAILED
DevOpsWorkflowLivenessTest > scheduler registers gauge at startup and records success…   FAILED
21 tests completed, 2 failed
Execution failed for task ':openbank-devops-agent:test'
```

and the same shape across **five** services: devops-agent, flaky-test-hunter, governance-auditor, release-steward, security-scanner.

That is the expected consequence of the change, not a coincidence. This PR alters `openbank-libs-runtime`'s `DomainMetrics` so the workflow-liveness age gauge is **seeded at registration** instead of starting at `Instant.EPOCH`. Every service's `*WorkflowLivenessTest` asserts the old semantics — a freshly-registered gauge reading as ancient — so they fail by construction once the seeding lands. The tests are correct about the behaviour they were written against; they now describe a state the code no longer produces.

So the fix is in this PR's scope: update the five services' liveness tests alongside the semantic change. Splitting them is not an option — the gauge's meaning and the assertions about it have to move together, or `main` goes red between the two merges.

## Why the change itself is right

Worth saying, because the red makes it look like the PR is the problem. `Instant.EPOCH` as a "nothing has happened yet" sentinel is the exact defect this repo has been unpicking all week — the same shape as `AuditEvent.timestamp`, `FlagExposure.timestamp` and `ApiError.timestamp` (#3882, #3880). A liveness gauge that starts at 1970 reads as *56 years stale* on a service that has merely not run yet, which is indistinguishable from a genuinely dead scheduler — and that is precisely the signal the gauge exists to carry. Seeding at registration makes "registered but never ran" a distinct, readable state.

The tests failing are the cost of fixing it, and they are the right tests to change.

## One thing to check while updating them

Whatever the new assertion becomes, it should still be able to fail on a *dead* scheduler. Asserting the gauge is merely "not EPOCH" after the change would pass for a registered-and-never-run job as readily as for a healthy one — trading a false-stale for a false-fresh. The useful assertion is recency bounded on both sides (registered ⇒ young, ran ⇒ younger), which is what the repo's other timestamp fixes settled on.

Automated agent review, not an approval. Touches `openbank-libs-runtime` and a ledger test, so it needs human review regardless.

JiRaska:
Refreshed against `main` (was ~189 behind, conflict-free). The failing run was from 2026-08-06, so a current verdict was worth having before diagnosing.

**But I do not expect the refresh alone to clear it, and here is the diagnosis in case it does not.** The failures are this PR's **own new tests**, in four modules:

```
DevOpsWorkflowLivenessTest > scheduler registers gauge at startup and records success after workflow dispatch  FAILED
DevOpsWorkflowLivenessTest > failed dispatch leaves liveness old                                              FAILED
  java.lang.AssertionError at DevOpsWorkflowLivenessTest.kt:41
  java.lang.AssertionError at DevOpsWorkflowLivenessTest.kt:85
```

plus the same shape in `openbank-flaky-test-hunter`, `openbank-governance-auditor` and `openbank-release-steward` — i.e. every consumer of the seeded gauge, not one flaky module.

What line 85 asserts is the interesting part:

```kotlin
assertThat(ageOf(registry)).isGreaterThan(FIFTY_YEARS_SECONDS)
```

after a dispatch that threw. So the test encodes the intended contract: **a workflow that has never succeeded must read as ~56 years stale, not as zero**. That is the right contract — it is the "gauge boot-zero is a fourth state" problem, where an unseeded gauge reads `0` and a staleness alert cannot tell "just succeeded" from "never ran". Getting it wrong in the safe-looking direction is exactly how a scheduler that never runs stays invisible.

So the test is asserting the thing the PR exists to deliver, and it is red. Either the seeding does not reach the gauge in this wiring (`registerLiveness()` is called explicitly in the test, so the path is exercised), or the sentinel it seeds is not the `Instant.EPOCH`-derived one the assertion expects. Worth checking whether `DomainMetrics.registryInstance` being injected *after* construction leaves the gauge registered against a different registry than `ageOf()` queries.

Not something I should guess at further from outside — it is your change and the failure is specific. Flagging it because a red on your own new test is the most informative kind: it means the test can express the failure, which is more than most of the queue can say.

JiRaska:
Heads-up: **#4208 appears to supersede this PR**, and it is in better shape.

Same title, same fix ("seed the workflow-liveness age gauge at registration"), opened five days later. Its file set is a strict superset — all **3** of this PR's files are in it, plus 25 more, covering the consumers rather than just `libs-runtime`:

```
shared files: 3
  openbank-libs-runtime/.../observability/DomainMetrics.kt
  openbank-libs-runtime/.../observability/DomainMetricsTest.kt
  openbank-ledger-service/.../schedule/LedgerWorkflowLivenessTest.kt
```

And it covers the four agent modules whose own new tests were failing here — `devops-agent`, `docs-truth-agent`, `flaky-test-hunter`, `governance-auditor` — which is what I reported on this PR earlier: the failures were this PR's *own* tests, asserting the right contract (a workflow that has never succeeded must read ~56 years stale, not zero) against an implementation that did not yet deliver it.

Current state, for comparison:

| | required checks |
|---|---|
| #4208 | ok=4 bad=**0** |
| this PR | ok=4 bad=**1** |

Not closing it — that is yours to decide, and you may be carrying something #4208 does not. But if the seeding work is all of it, this looks like the one to close, and the contract your tests encode is worth checking survives into #4208: *"a scheduler that has never run must not read as freshly successful"* is exactly the distinction that makes a staleness alert mean anything, and it is easy to lose in a wider refactor.

JiRaska:
Closing as superseded by #4208 — same root fix (`DomainMetrics.registerWorkflowLiveness` seeding from `Instant.EPOCH`, ADR-0237 / #2239 Gap 2), not abandoned.

**Evidence, checked before closing:**

- #3634 touches only `openbank-libs-runtime/DomainMetrics.kt` + 2 tests. Its build fails on **18 unrelated services** (security-scanner, sdd-service, release-steward, lending, ledger, interest, governance-auditor, flaky-test-hunter, finops-agent, engagement, domestic-payment, docs-truth-agent, devops-agent, copilot, balance, authz-policy-auditor, analytics-sink, aml) — every one of them a downstream consumer of `libs-runtime` that fails to compile against this PR's narrow change, because #3634 never added the corresponding `WorkflowLivenessMetrics` support / per-service liveness test each of those services needs.
- #4208 does the full job: adds `WorkflowLivenessMetrics.kt` to `libs-domain`, a `*WorkflowLivenessTest.kt` for every one of those 18+ services, the `control-liveness-sentinel` activity wiring, the Prometheus rule, and the ADR doc. Checked fresh just now: **0 failing, 0 pending** — fully green.

`#4208` is `OPEN`, verified immediately before this close.

Issue #3393, opened 2026-08-02
security-scanner is the only workload on Kafka plain 9092: no KafkaUser, ANONYMOUS against authorization: simple


JiRaska:
## Summary

`security-scanner` is the only workload in the fleet talking to Kafka over the **plain 9092** listener, with no `KafkaUser` and no keystore. The other 36 services have a `KafkaUser` and use mTLS on 9093.

```
$ kubectl get deploy -A -o json | jq '... select KAFKA_BOOTSTRAP_SERVERS matches 9092'
security-scanner/security-scanner-service        <- the only one

$ kubectl get kafkauser -n messaging | wc -l
36        <- and none of them is security-scanner
```

The cluster runs `authorization: simple` with no `superUsers`, and the `plain` 9092 listener has **no authentication** — so the scanner's producer authenticates as `ANONYMOUS`, which no ACL grants. Its target topic `openbank.security.scan.event` exists and is consumed by `audit-service`.

## Why nobody noticed

`SecurityOutboxDispatcher` runs every 5s and has never got as far as producing: Flyway was never enabled, so the table it reads does not exist and every tick dies before the emitter is touched (#3350):

```
Caused by: java.sql.SQLException: ERROR: relation "security_outbox" does not exist (42P01)
```

Once #3350 lands, the query succeeds, the emitter is reached for the first time, and the authorization gap becomes the new failure. That PR is still the right change — the `/report` 500 is fixed either way, and one 5s error loop simply replaces another — but this is the next thing to fall over, so it should not be discovered by surprise.

## What needs doing

1. A `KafkaUser` for `security-scanner` with produce ACLs on `openbank.security.scan.event` and `openbank.security.ict.incident` (the two topics its `application.yaml` declares).
2. Mount the resulting keystore/truststore and point `KAFKA_BOOTSTRAP_SERVERS` at `openbank-cluster-kafka-bootstrap.messaging.svc:9093` with `KAFKA_SECURITY_PROTOCOL=SSL`, matching the pattern every other service already uses.
3. Verify by asserting a row actually leaves `security_outbox` (`status` moves off pending, `sent_at` set) — not by the absence of an error, since the dispatcher swallows failures into `markFailedUni` and `attempt_count`.

## Worth checking while in there

Whether the plain 9092 listener should exist at all. With `authorization: simple` and no authentication it grants nothing, so it is not a hole today — but it is an unauthenticated entry point whose only consumer was a service that never successfully used it.

[...]

Issue #3271, opened 2026-08-02
payments: domestic-payment has no Kafka ACL for its own event topic — 13 outbox rows DEAD, and the manifest comment asserts the opposite of the live cluster


JiRaska:
### Affected service / module

openbank-infra (gitops/components/payments) + openbank-domestic-payment

### Version / commit

main @ 63b6cd8dd (observed live in sandbox, 2026-08-02)

### What happened?

domestic-payment cannot publish to its own domain-event topic. Every attempt is refused:

```
TopicAuthorizationException: Not authorized to access topics: [openbank.domestic.payment.events]
```

The `domestic-payment` KafkaUser carries exactly one ACL — `Write, Describe` on
`payment.scheme-accepted`. Nothing grants it its own event topic. The live cluster runs
`allow.everyone.if.no.acl.found: "false"` (`kafka.openbank-cluster.spec.kafka.config`), so an
mTLS principal with *any* ACL list is denied every resource not on it.

**The manifest asserts the opposite.** The header of
`openbank-infra/gitops/components/payments/kafka-scheme-accepted-acl.yaml` says:

> resources without ACLs stay open under allow.everyone.if.no.acl.found=true (see kafka.yaml), so
> each service only needs ACLs for the locked-down scheme-accepted topics — its own domain-event
> topics need none.

That is false against the deployed cluster, and it is the sentence that will send the next person
down the wrong path. `payments/` is also the only namespace with no `kafka-<svc>-mtls.yaml` — every
other component directory (`accounts/`, `aml/`, `consent/`, `campaign/`, …) has one granting the
service its own topics.

This is the same class as #2078, where standing-order's consumer could not join its group for
exactly this reason. That fix added one KafkaUser; it did not correct the comment or sweep the
namespace.

### Consequence, measured

`domestic_payment_outbox` holds **13 rows in `DEAD`** — dispatch retries exhausted, events
permanently unpublished. The row for payment `c05853f8-4259-4c97-b648-9c8a84082789`:

```
event_type    | domestic.payment.created
status        | DEAD
attempt_count | 10
last_error    | DomesticPaymentOutboxDispatcher#publishWithResilience circuit breaker is open
```

Note the current log is **quiet** — 0 denials in the last 15 minutes — precisely because every
affected row has already gone `DEAD` and nothing retries. A quiet log here means the events are
lost, not that the problem resolved.

Downstream of a missing `domestic.payment.*` event: notification-service never tells the customer,
[...]

Issue #2749, opened 2026-07-30
Campaign service rollout: remaining slices after first-slice code PR


JiRaska:
## Source ADR

ADR-0200 (campaign journeys), ADR-0209 D3 (first slice: deterministic segments, EMAIL-only), ADR-0219/0220/0221 (contact-policy gate, engagement, Studio)

## Kind of follow-up

Pending rollout / implementation of an accepted decision

## What concretely must happen

Follow-ups after the first-slice code PR (service + consent-gated journeys + deterministic segments + four-eyes activate):

1. **GitOps component** for openbank-campaign-service: namespace, CNPG postgres, NetworkPolicy, ConfigMap/ExternalSecret, Kafka ACLs (openbank.notification.requests produce, openbank.consent.events consume), KEDA ScaledObject; add to auto-deploy ALL_SERVICES.
2. **Adverse-state suppression inputs** (ADR-0200 D6): fraud-hold/arrears/dispute signals are not in the analytics-sink topic set today — decide the source (silver expansion vs party-service) and wire the D6 exclusions beyond frequency-cap/quiet-hours/party-status.
3. **Admin-ui Campaign Studio** per ADR-0221 (wizard, consent funnel, four-eyes queue integration).
4. **ContactPolicyGate in libs-runtime** per ADR-0219 (promote campaign-scoped suppression to the platform primitive; impression classes come with ADR-0220 surfaces).
5. **ONNX/NBA stays blocked** per ADR-0209 D4 until ADR-0140 phase 2 (offline store) exists — do not start.

## Acceptance criteria

- openbank-campaign-service deploys to sandbox via ArgoCD and passes the boot smoke-test pattern.
- A marketing send path bypassing the consent/contact gate fails CI (contract test, ADR-0219 D4).
- Studio PR merges with the ADR-0221 non-goals intact (no free text, no SQL, no CSV).

## Compliance touchpoints

- [x] GDPR (Art. 7/21 — consent-gated delivery, suppression)
- [x] CNB (Act 480/2004 — commercial communications)

JiRaska:
[...]

PR #3642, opened 2026-08-03
fix(infra): give security-scanner a Kafka identity and move it onto mTLS :9093

JiRaska:
## What this is

`security-scanner` gets a Kafka identity (`KafkaUser` + ACLs + ESO projection) and moves off the anonymous plaintext listener onto mTLS `:9093`, the shape every other client already uses.

Refs #3393.

## Real state, as MEASURED on the live sandbox (2026-08-03)

The issue's premise held on the important point and was wrong on two others. All of this was measured, not read off a manifest:

**The denial is real and total.** The broker logs, once per 30s:

```
Principal = User:ANONYMOUS is Denied operation = DESCRIBE from host = 10.80.86.183
  on resource = Cluster:LITERAL:kafka-cluster for request = DescribeLogDirs
  with resourceRefCount = 1 based on rule DefaultDeny
```

An ANONYMOUS metadata request on `:9092` returns **zero** topic-partitions — not "the two scanner topics are denied", *everything* is. `allow.everyone.if.no.acl.found: "false"` means a resource with no ACL is denied, not open.

**`security-scanner` has never published anything.** Ground truth, not inference — the on-disk log segments:

```
openbank.security.scan.event-0/00000000000000000000.log      0 bytes, Jun 4
openbank.security.ict.incident-0/00000000000000000000.log    0 bytes, Jun 4
```

**But not for the reason the issue gives.** #3350 has landed and is deployed (image `sandbox-9584ae90` is *after* `0a64027a1`); Flyway ran, `security_outbox` exists. The table is nonetheless **empty and always has been** — `SecurityOutboxRepositoryImpl.persistInTransaction` has **no caller anywhere in `src/main`**. So the dispatcher drains an empty table forever and never reaches the emitter. The one live produce path is `IctIncidentService` on `POST /api/v1/ict-incidents`, which is human-triggered. The scanner has not been silently denied — it has not tried.

**It is not the only workload on `:9092`.** Three are, and they differ:

| workload | on 9092 | actually sends? |
|---|---|---|
| `security-scanner-service` | yes | no — never has (this PR) |
| `kafka-ui` | yes | **yes, and every request is refused** — it is the `10.80.86.183` in the log above, and renders an empty cluster rather than an error |
| `payments/settlement-service` | yes | no — it has no `mp.messaging` config at all; the env var is dead |

**The issue's suggested verification does not work.** It proposes asserting a row leaves `security_outbox` with `sent_at` set. `SecurityOutboxDispatcher` wraps `emitter.send(...)` in `Uni.createFrom().item { … }`, which discards the returned `CompletionStage`, then chains `markSentUni` unconditionally. A denied publish would mark the row **SENT**. Verify on the broker side (topic offset / segment size), never on the outbox status.

## What changed

Two commits, deliberately separable — see the ordering section.

**Identity** (`b71dca0`) — changes no client behaviour:
- new `components/security-scanner/kafka-security-scanner-mtls.yaml`: `KafkaUser security-scanner` with `Write, Describe` on `openbank.security.scan.event` and `openbank.security.ict.incident` (the two topics `application.yaml` declares — two `outgoing` channels, no `incoming`, so no consumer-group ACL), plus the keystore + cluster-CA-truststore ExternalSecrets.
- `eso-kafka-cert-reader` Role `resourceNames` += `security-scanner`. This is the entry whose absence produced the #2749/#2872 failure twice: the ExternalSecret sits at `SecretSyncedError`, ArgoCD never advances past sync-wave -1, no pod is created.
- The user is named `security-scanner` (not `security-scanner-service`) because `check-kafka-acl-coverage.py` derives its match from the module name. That makes the ACLs *checkable*, which they were not: the service used to print as an unverifiable `::notice::`. The gate goes 88 → 90 pairs across 36 → 37 users, and the cert-reader gate 38 → 39, both clean, so **no baseline entry is needed in either**.

**Client** (`99b86d9`) — `KAFKA_BOOTSTRAP_SERVERS` → `:9093`, `KAFKA_SECURITY_PROTOCOL=SSL`, keystore/truststore env + mounts.

Neither secret ref is `optional: true`. Deliberate, per `openbank-infra/CLAUDE.md`: a missing projection must pin the pod at `ContainerCreating` (loud) rather than let it boot and silently publish nothing. This service is the argument for loud — it produced nothing for two months and no signal existed anywhere.

**Also corrected: the `allow.everyone.if.no.acl.found=true` claim, everywhere it still stood.** It has been false since #2794, and it is precisely what made a set of missing producer grants read as deliberate until 13 outbox rows went dead (#3271). Three copies survived that cleanup:
- `components/kafka/kafka.yaml` — the *authoritative* file, whose own `authorization:` block said "allow-everyone only for resources with NO ACL (because allow.everyone.if.no.acl.found=true below)" while the line 20 below it says `"false"`.
- `components/payments/kafka-scheme-accepted-acl.yaml` — the #3271 correction landed *underneath* the false sentence without deleting it, so the header stated the claim and then said "this comment used to say the opposite", quoting itself.
- `components/audit/kafka-audit-mtls.yaml` — twice, including a note asserting three topics "stay open under allow-everyone; no ACL change needed", which #3271 had already disproved by granting all three.

The `kafka.yaml` listener comment also claimed "All 33 deployed Kafka clients migrated to mTLS:9093 (ADR-0137 fleet sweep complete)". It now names the three that had not.

## Order an operator must apply this in

ArgoCD's sync-waves encode it (`KafkaUser` -2, `ExternalSecret`s -1, Deployment 0) — but the Role grant lives in a **different Application** (`payments`), so cross-app ordering is not guaranteed. Verify in order; do not skip to 4:

1. `kubectl -n messaging get role eso-kafka-cert-reader -o jsonpath='{.rules[0].resourceNames}'` contains `security-scanner`
2. `kubectl -n messaging get kafkauser security-scanner` is Ready **and** `kubectl -n messaging get secret security-scanner` exists — Strimzi's User Operator mints it on its own reconcile interval, not instantly
3. both ExternalSecrets in `ns/security-scanner` report `SecretSynced`
4. only then does the Deployment roll. **A pod at `ContainerCreating` means step 3 has not finished, not that this change is wrong** — wait, do not roll back

Verify the outcome on the **broker**, not the pod or the outbox: `POST /api/v1/ict-incidents`, then confirm `openbank.security.ict.incident-0/00000000000000000000.log` is no longer 0 bytes, and that no new `User:ANONYMOUS ... Denied` line names the scanner's pod IP.

Rollback = revert the second commit (`:9092`, no SSL). That restores the pre-change behaviour exactly, which is "connects and is denied everything".

## The plain listener stays — and here is what blocks removing it

Closing `:9092` **to security-scanner** is what this PR does. Removing the listener outright is not safe yet: `kafka-ui` still uses it, and it is the only thing actually sending there. It needs its own `KafkaUser` first (its consumer/describe surface is much wider than a producer's, so that is its own change with its own review). `settlement-service`'s env var is dead and can be deleted in passing by whoever does that. Tracked as the #3393 follow-up.

## Verification done here

- `check-kafka-acl-coverage.py --enforce` — clean, 90 pairs / 37 users, and **falsified in place**: renaming the ICT grant to a near-miss made it print the exact error naming `openbank.security.ict.incident` and exit 1; restored from a `cp` backup, back to exit 0. Its own `--selftest` passes.
- `check-kafka-cert-reader-grant.py --enforce` — clean, 39/39, both directions.
- `check-kafka-topics-exist.py` — 0 unresolved.
- `gen-network-policies.py` re-run: **no diff**. The derived broker allow-list keys on namespace, not port (it hardcodes 9092 for every client, including the 35 already on 9093 — a latent defect, not one this PR introduces or can fix).
- `run-gates.py --group gitops` — 19 PASS. The two non-passes are pre-existing and unrelated: `incluster-hostname-resolution` (advisory; 6 findings, none in any file here) and `loki-rule-load-test` (`BASE_REF: unbound variable`, a local-run artifact).

## What could still go wrong

- Strimzi could be slow to mint the secret and the pod sits at `ContainerCreating` for a few minutes. That is the designed failure and step 4 above says so, but an operator who reads it as breakage and reverts gets a service that is once again silently denied.
- The grant is correct for the topics `application.yaml` names **today**. A future producer that derives its topic in code is outside what `check-kafka-acl-coverage.py` can see, by that gate's own stated scope.
- This PR does not make the scanner publish. `security_outbox` still has no writer, so the scan-event topic stays empty until that is wired — the identity is now in place for when it is.

### DIFF
--- a/docs/runbooks/0008-kafka-mtls-scheme-accepted-cutover.md
+++ b/docs/runbooks/0008-kafka-mtls-scheme-accepted-cutover.md
@@ -142,3 +142,72 @@ next image rebuild — currently blocked by the Pact `can-i-deploy` gate, **#266
 Do **not** try to force it via `MP_MESSAGING_INCOMING_*` env vars: SmallRye cannot
 take hyphenated mp.messaging channel config from env and the pod fails to start
 (tried and reverted, #2649 → #2659).
+
+## Stage 6 — remove the `plain:9092` listener (issue #3393)
+
+The last stage of this runbook, and the one that makes the rest of it enforceable
+rather than merely configured. **Everything above this heading was written while
+`allow.everyone.if.no.acl.found` was still `true` and the plaintext listener was
+in place; read it as history, not as instructions.** Specifically:
+
+- "Untouched: every other Kafka client (they keep using `plain:9092`)" and the
+  "Out of scope — cluster-wide gate flip" section were overtaken by **#2794**,
+  which flipped the gate to `false` for the whole cluster.
+- The verification commands that exec into the broker with
+  `--bootstrap-server localhost:9092` no longer have a listener to reach, and
+  the Stage-4 negative test ("an anonymous client on `plain:9092` is DENIED")
+  cannot be run at all any more — its outcome is now structural. A broker-local
+  check needs `:9093` plus a client `--command-config` carrying a KafkaUser
+  keystore.
+
+### What had to be true first
+
+The listener was kept because removing it was assumed to break clients. It did
+not grant anything — a client on it authenticates as `User:ANONYMOUS`, which
+holds no ACL and is denied on every resource once the gate is `false` — so
+"still on 9092" and "cannot use Kafka" were the same statement. Three workloads
+still named it on 2026-08-03, each resolved before the removal:
+
+| workload | state before | resolution |
+| --- | --- | --- |
+| `security-scanner-service` | no `KafkaUser` at all; both topic log segments 0 bytes since creation — it had never published | `KafkaUser security-scanner` + ACLs, moved to `:9093` |
+| `kafka-ui` | the only workload actually speaking on `:9092`; every request denied (`Principal = User:ANONYMOUS is Denied operation = DESCRIBE … based on rule DefaultDeny`, once per 30s poll), rendered as an empty cluster | `KafkaUser kafka-ui` + browse-scoped ACLs, moved to `:9093` |
+| `settlement-service` | `KAFKA_BOOTSTRAP_SERVERS` env with no `mp.messaging` config and no Kafka client in its source — dead value | env deleted |
+
+### Apply
+
+```
+argocd app sync kafka          # Strimzi rolls the broker to drop the listener
+kubectl -n messaging get kafka openbank-cluster -o yaml | yq '.status.listeners[].name'   # expect: tls  (only)
+```
+
+### Verify
+
+1. **Every client is still connected**, i.e. the roll changed nothing for them:
+   ```
+   kubectl get pods -A -l app.kubernetes.io/part-of=openbank --field-selector=status.phase=Running
+   kubectl logs -n <ns> deploy/<svc> --since=10m | grep -iE "SSL handshake|disconnected|TimeoutException"
+   ```
+2. **kafka-ui shows a populated cluster** — topic list, per-topic partitions and
+   consumer groups with lag. That is the observable this whole issue turned on:
+   before, the same screen showed an empty cluster and no error.
+3. **The denial log is gone.** The broker authorizer should no longer print
+   `User:ANONYMOUS is Denied` at all — previously once per 30s, forever:
+   ```
+   kubectl -n messaging logs openbank-cluster-dual-0 --since=10m | grep ANONYMOUS   # expect: nothing
+   ```
+4. **The derived NetworkPolicy names 9093, not 9092.** `gen-network-policies.py`
+   used to hardcode the port; it now derives it from the clients' bootstrap URLs,
+   so a stale port here is a real (if latent) hole rather than a cosmetic one:
+   ```
+   kubectl -n messaging get netpol kafka-broker-ingress-allow-list -o jsonpath='{.spec.ingress[*].ports[*].port}{"\n"}'
+   ```
+
+### Rollback
+
+Re-add the four listener lines to `components/kafka/kafka.yaml` and sync. Note
+what that does and does not buy: it restores *connectivity* on 9092, never
+*authorization* — an ANONYMOUS client is still denied on every resource by the
+gate. So it is only worth doing to quiet a connection-error loop while a real
+client-side problem is fixed, and it is not a path back to a working plaintext
+client. There is no such path; that is the point of the stage.
--- a/openbank-infra/gitops/components/audit/kafka-audit-mtls.yaml
+++ b/openbank-infra/gitops/components/audit/kafka-audit-mtls.yaml
@@ -3,8 +3,10 @@
 # Identity model (ADR-0137): mTLS gives each service exactly one client cert =
 # one KafkaUser = one principal for ALL of its Kafka traffic. The audit-service
 # is a pure consumer of domain events across 10 topics; it only needs Read+Describe
-# ACLs on those topics and Read on its consumer group. Topics not listed here
-# stay open under allow.everyone.if.no.acl.found=true (see kafka.yaml).
+# ACLs on those topics and Read on its consumer group. NOTE: "topics not listed here
+# stay open under allow.everyone.if.no.acl.found=true" stood here and has been false
+# since #2794 — a topic with no ACL is denied, not open, which is exactly how the six
+# subscriptions at the bottom of this list came to be receiving nothing (#3271).
 #
 # clearing.batch.event / security.ict.incident / security.scan.event added (ADR-0160
 # #996 triage): each was published with zero consumers since introduction — audit-service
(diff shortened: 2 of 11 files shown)


JiRaska:
## Independent re-verification (2026-08-06) — no changes made, no new PR

I was asked to work #3393 and found this PR already carries the fix. Rather than open a
duplicate, here is an independent check of (a) the premise still holding on the live cluster
and (b) this branch still being safe to merge 90 commits later.

### Premise still true on the live sandbox

| claim | result |
|---|---|
| `kafkauser/security-scanner` in `messaging` | **NotFound** — 38 KafkaUsers exist, none is it |
| deploy `security-scanner/security-scanner-service` Kafka env | exactly one var: `KAFKA_BOOTSTRAP_SERVERS=…:9092`. No `KAFKA_SECURITY_PROTOCOL`, no keystore |
| `security-scanner` in `role/eso-kafka-cert-reader` resourceNames | **absent** (38 entries) |
| cluster authorization posture | `type: simple`, `allow.everyone.if.no.acl.found=false` |
| workloads still naming `:9092` | 2 — `messaging/kafka-ui`, `security-scanner/security-scanner-service` |

So the deny-by-default reading is right: an ANONYMOUS principal on `:9092` matches no ACL and
is denied. Note the earlier comment listed **three** `:9092` workloads; `payments/settlement-service`
is no longer one of them, so `kafka-ui` (#3652) is the only remaining blocker for retiring the listener.

### Branch is still safe to merge

- Merge-base is 90 commits back, but the ruleset's `strict_required_status_checks_policy` is
  **false**, so the lag does not gate the merge.
- Test-merged `origin/main` into the branch in a throwaway worktree: **clean, no conflict**, and
  the post-merge diff vs `main` is exactly the 5 intended files — nothing silently dropped.
- The one file both sides touched is `security-scanner-service.yaml`; `main` changed only the
  image tag (#3746, `sandbox-9584ae90` → `sandbox-735e8bdc`) and the merged result **keeps
  main's newer tag**. This branch does not roll back the deployed image.
- Both enforced gates re-run on the *merged* tree: `kafka-acl-coverage` clean (90 pairs / 37
  users; security-scanner is now genuinely checked, not an unverifiable `::notice::`) and
  `kafka-cert-reader-grant` clean (39 read / 39 granted).
- Falsified the grant gate rather than trusting its green: deleting the one added
  `security-scanner` line makes it report `38 granted … 1 finding`. It detects the thing it claims to.
- Env-var shape compared line-for-line against `delegation-service` (a working mTLS client):
  identical var names, `/mnt/kafka-{keystore,truststore}/{user,ca}.p12`, `PKCS12`, and the same
  `user.password` / `ca.password` secret keys.
- The two ACL'd topics match `openbank-security-scanner/src/main/resources/application.yaml`
  exactly (`openbank.security.scan.event`, `openbank.security.ict.incident`); two `outgoing`
  channels and no `incoming`, so no consumer-group ACL is correct.

### Does this roll the pods?

**Yes** — the change edits the Deployment's pod template (env + volumeMounts + volumes), so it is
a genuine rollout, not a ConfigMap edit that leaves pods stale. That is why the operator sequence
in the PR body matters: if the ESO projections have not landed, the new pod stays at
`ContainerCreating` (neither secret ref is `optional: true`, deliberately).

### What I did NOT verify

- I did not create the KafkaUser or apply anything — no cluster mutation, no `tofu apply`. So I
  have **not** observed the Strimzi User Operator actually minting `secret/security-scanner`, the
  ExternalSecrets reaching `SecretSynced`, or a message landing on either topic. The operator
  sequence in the PR body is still unwalked.
- I did not re-read the broker's authorizer log myself; the ANONYMOUS `DefaultDeny` line and the
  0-byte segment measurements are from the earlier comment. The config posture that *implies* the
  denial (`simple` + `allowEveryone=false`) I did confirm directly.
- I did not verify the `dataFrom.extract.key: security-scanner` resolves in the
  `kafka-messaging-certs` ClusterSecretStore — that key only exists once the KafkaUser is created.
- ArgoCD sync-wave ordering is declared, not tested; the Role grant lives in the **payments**
  Application, so cross-app ordering is not guaranteed and step 1 of the operator sequence is a
  real manual check, not a formality.

### Blast radius if this is wrong

Confined to `security-scanner`. Worst case the pod cannot mount its keystore and stays at
`ContainerCreating` — it loses its HTTP endpoints (`/report`, `POST /api/v1/ict-incidents`) until
rolled back. It cannot lose Kafka *throughput*, because it has none: nothing has ever published on
either topic and nothing writes `security_outbox`. Rollback is the second commit alone. Nothing
else in the fleet reads these two topics except `audit-service`, which is a consumer of a topic
that has never had a message.

One thing worth flagging separately: while this is unmerged, a denied publish would still be
recorded as **SENT** — `SecurityOutboxDispatcher` discards the `CompletionStage` from
`emitter.send(...)` and chains `markSentUni` unconditionally. #3646 is the fix for that and is
what makes "verify by watching the outbox" meaningful at all; until it lands, verification has to
be read off the broker (segment size / offset), never off outbox status.

### State

All five required contexts are green; `mergeStateStatus` is `UNSTABLE` only because
`agent-review` is red, which is the retired-GitHub-Models gateway failure and is not a required
check. The merge remains **blocked pending human review** because the diff touches a money-path
path (`components/payments/kafka-scheme-accepted-acl.yaml`, where the `eso-kafka-cert-reader`
Role lives). That block is correct and I did not override it.

Issue #3545, opened 2026-08-03
108 single-replica Deployments cannot roll on a full cluster — the old pod holds the resources the new one needs


JiRaska:
## What

**108 of 110 single-replica Deployments in this cluster cannot complete a rollout when the cluster
is full.** Only 2 are configured to terminate the old pod first.

A one-replica Deployment with the default `RollingUpdate` resolves to `maxSurge: 1` /
`maxUnavailable: 0`, so Kubernetes must schedule the new pod *before* stopping the old one. The
rollout needs room for two pods — and the old pod holds exactly the resources the new one needs.
With no spare room anywhere, that is a deadlock the rollout never escapes.

Measured 2026-08-03 on `delegation-service`:

```
FailedScheduling  0/32 nodes are available: 1 Too many pods, 20 Insufficient memory,
                  4 Insufficient cpu, 8 node(s) had untolerated taint(s)
nodepool default  cpu 48/48  nodes 23      <- at the Karpenter limit, cannot add a node
```

It held a security-relevant fix (#3537 — without it no customer can complete a delegation
ceremony) out of production for hours, silently: the Deployment reports healthy because the OLD
pod is Ready, and only the pin-vs-running-image comparison shows anything is wrong.

**It does not drain.** The overnight CI queue emptied and the pod stayed `Pending`.

## Why this is not "just add capacity"

Capacity is the trigger, not the defect. The same cluster will be full again — that is what a
Karpenter limit is for. What makes it a fleet risk is that the failure is:

- **silent**: `kubectl get deploy` shows `1/1 READY` throughout;
- **worst-timed**: it bites exactly when the cluster is under pressure, which is when you are most
  likely to be shipping a fix;
- **unbounded**: `progressDeadlineSeconds` marks the rollout failed but nothing acts on it.

## What the fix is NOT

Applying `maxSurge: 0 / maxUnavailable: 1` everywhere would be wrong. The right split:

- **Single-replica, no-HA services** (the large majority of the 108): terminate-first costs nothing
  they had — they are already unavailable for the moment the pod is replaced. Done for
  delegation-service in #3544.
- **Multi-replica money-path services**: must keep surge. There, surge IS the zero-downtime
  property, and the answer is having the headroom to surge — not trading the property away.

So the work is a classification, not a sweep. A reasonable shape:

1. Derive the list of single-replica Deployments from the gitops tree (not by hand — a hand-kept
   list of the thing you are checking reads as complete because it is long).
[...]

Issue #3467, opened 2026-08-02
compliance-pack activation is per-pod in-memory: scaling lending past 1 replica diverges the four-eyes control


JiRaska:
## What

`CompliancePackRegistry` is per-pod in-memory state with no cross-replica propagation. Scaling `openbank-lending-service` past one replica makes a four-eyes compliance control diverge per pod, silently.

```kotlin
// openbank-libs-domain/.../compliance/CompliancePackRegistry.kt
class CompliancePackRegistry {
    private val active: AtomicReference<Map<PackKey, List<CompiledCompliancePack>>> = ...
    fun activate(proposal: Proposal<CompiledCompliancePack>): CompiledCompliancePack
}
```

Two writers, no reader between them:

- `CompliancePackActivationService.decide()` calls `registry.activate(decided)` on the pod that served the approval.
- `CompliancePackConfig` rehydrates the registry from persisted activations **at boot**.

So a pack approved on pod A is enforced by pod A immediately and by pod B only after pod B restarts. There is no bound on that window and nothing reports it — no metric, no readiness signal, no log on the pods that are behind.

## Why it is not visible today

`openbank-infra/gitops/components/lending/lending-service.yaml` sets `replicas: 1`, so the divergence is unreachable in the current topology. That is the whole risk: the constraint lives in a gitops value, not in the code that depends on it, and nothing connects the two. A routine "lending is slow, scale it to 3" is all it takes, and the symptom — the same loan application judged against different compliance rules depending on which pod answered — is one nobody would attribute to a replica count.

## Where this came from

#3448 fixed the ordering bug that produced the same divergence for a different reason: `registry.activate` ran *before* the row was persisted, so a failed save left a pod enforcing a pack with no row to show for it. That PR closes the phantom-activation half. The replica half is untouched, and #3448's own description states the harm in terms that remain true:

> sibling replicas never had it, so the same application would be judged against different rules depending on which pod answered

## Options

Not proposing one — this needs a decision, and the cheapest correct answer may be the first:

1. **Declare the constraint and enforce it.** If lending is intended to be single-replica, say so where it can be checked — a guard asserting `replicas: 1` for any service holding authoritative in-memory state, so scaling it fails review rather than production.
2. **Propagate on write.** Publish an activation event and have each pod apply it (the outbox already exists in this service).
3. **Read-through.** Drop the in-memory registry and read active packs per decision, with a short TTL cache.

[...]

PR #3644, opened 2026-08-03
fix(lending): converge the compliance-pack registry across replicas

JiRaska:
## What

`CompliancePackRegistry` is per-pod in-memory state with two writers and nothing between them:
`CompliancePackActivationService.decide()` on the pod that served the approval, and
`CompliancePackBootLoader` at boot. A pack approved on pod A was enforced by pod A immediately and by
pod B only after pod B next restarted — an unbounded window in which the same loan application is
judged against different compliance rules depending on which pod answered. On a money-path four-eyes
control that is a divergence, not a stale cache.

This adds `CompliancePackRefresher`: each pod re-reads the activation table on
`lending.compliance.refresh-interval` (default `30s`, `LENDING_COMPLIANCE_REFRESH_INTERVAL`) and
converges onto it.

## Measured blast radius

- **What is per-pod:** one field — `AtomicReference<Map<PackKey, List<CompiledCompliancePack>>>` in
  `CompliancePackRegistry`. Written by `decide()` (approve leg) and the boot loader; read by
  `CompliancePackGuard.checkOriginationAllowed` / `resolveOriginationPack` on every origination and
  by `GET /api/v1/lending/compliance-packs/active`.
- **On a roll the state is NOT lost** — contrary to the issue text. Every activation is a committed
  `compliance_pack_activation` row and boot rehydrates from it. The defect is strictly the gap
  between an approval on one pod and a restart of the others.
- **Is lending pinned to 1 replica because of this?** Not provably. `lending-service.yaml` is a
  `kind: Rollout` with `replicas: 1` and no `ScaledObject`/`HPA` anywhere for lending, and so is
  essentially the whole estate: of 98 workloads under `openbank-infra/gitops` (76 `Deployment`,
  21 `Rollout`, 1 `StatefulSet`), **97 are `replicas: 1`** — the only exception is
  `developer-portal` at 2. So single-replica is the house default, not a lending-specific mitigation,
  which means nothing anywhere records that scaling lending was unsafe.
- **Nothing in CI would have caught it.** The only replica-aware gate is
  `single-replica-rollout-strategy`, which asserts a deployment-mechanics property (an explicit
  `strategy` block, #3545) and has no notion of per-pod state.
- **Other services carry the same shape** (surveyed, not fixed here — see "Not in this PR").

## How

- `CompliancePackRegistry.syncFrom(proposals)` — converges onto the durable activation set and
  returns how many packs this call added, so a replica that *was* behind says so once.
  **Additive, not a wholesale replace**: a pack is never un-activated (ADR-0212 D3), so on every real
  state the two agree; they differ only on the race where a snapshot read a moment before a
  concurrent `activate` would drop the pack from the pod that just approved it, refusing origination
  there until the next refresh. Each row is re-validated as an approved four-eyes decision, so a row
  cannot become enforceable by arriving through the convergence door instead of the approval one, and
  validation happens before anything is applied so a bad row cannot half-converge a pod.
- **Why a periodic read and not an outbox event:** the activation row is already the durable,
  auditable record, written under the same transaction as the four-eyes state change. Propagating
  over the outbox would add a second copy of that fact and a second way to miss it (an unread topic,
  a consumer group that never joined). Polling the record of truth cannot drift from it, and a failed
  refresh simply leaves the pod where it was — the next tick catches up.
- `refresh()` is a **`suspend fun`**: a plain `@Scheduled` method carries no Vert.x context, so a
  reactive Panache call inside `runBlocking` throws `HR000068` and the job silently does nothing
  (#2148/#2187).
- One unreadable row is logged and skipped rather than stopping the others. The boot loader still
  fails loudly on the same input — at boot, refusing to start is the safe answer; here the pod is
  already serving and one bad row must not become a fleet-wide origination outage.
- The row→proposal mapping is now shared by the boot loader and the refresher, so the two cannot
  drift about what a persisted activation means.
- No DB change (the table and its rows already exist) and no API change, so no Flyway migration and
  no `openapi.yaml` bump.

## Proven red, then green

`CompliancePackReplicaConvergenceIT` was written and run **before** the fix existed. The running
application is pod B: it boots against an empty table and approves nothing. Pod A's approval arrives
the only way a sibling pod can ever see it — as a committed row, inserted over plain JDBC — and the
assertion reads what pod B *serves*, through the endpoint the console and the guard use. It is
written entirely against a table and an endpoint that already exist, so it measures the absence of
any convergence mechanism rather than the absence of a new API.

| | result |
|---|---|
| unpatched `main` + test only | **FAILED** — timeout after 30 s, active list still `[]` |
| with `CompliancePackRefresher` | **PASSED**, and the log carries `compliance packs converged from the activation table: 1 pack(s) newly enforceable` |

The row is inserted rather than approved over REST deliberately: approving through pod B's own
endpoint activates the pack as a side effect of the request, which is the path that already worked
and the one that cannot diverge.

Also added: four `CompliancePackRegistryTest` cases for `syncFrom` — converges, is idempotent across
repeated refreshes (a replace-and-fail design would log an error on every tick after the first),
never drops a locally-activated pack, and refuses a self-approved row without partially applying the
batch. Existing `CompliancePackActivationIT` (4 tests) still green.

## Not in this PR — stated so it is not mistaken for done

1. **This does not by itself make lending safe to scale.** It removes this divergence and bounds
   convergence at the refresh interval; it is not a claim that nothing else in the service is
   replica-hostile.
2. **The concurrent-`decide()` race the issue names is untouched.** Two checkers can both pass
   `require(entity.state == PROPOSED)` and both merge. Same four-eyes invariant, different mechanism
   (it needs a conditional update or a row lock), so it belongs in its own change.
3. **The registry remains a per-pod cache with bounded staleness**, not shared state. A read-through
   per decision would need `CompliancePackGuard` to become reactive across every origination call
   site — a much larger change for a control whose inputs are effective-dated and change rarely.
4. **The same pattern exists elsewhere and is worse where it is not persisted.** Surveyed while
   establishing blast radius, filed separately rather than fixed here: security-scanner's DORA ICT
   incident register and scan results, analytics-sink's maker-checker proposal store, backfill report
   and WORM anchor chain head, customer-edge's nearby-payment session store, and agent-service's
   single-use PoP nonce set. Unlike the compliance pack, none of those has a row to rehydrate from.

Refs #3467, #3448, ADR-0212 D4


JiRaska:
Review: 2 findings, 0 blocking

1. `openbank-lending-service/.../compliance/CompliancePackRefresher.kt:74-84` — the per-row `runCatching` only wraps `row.toProposal()` (parse/compile). `registry.syncFrom` then validates the whole batch up-front (`proposals.forEach(::requireApprovedFourEyes)`) and throws `PackActivationException` outside that guard. Scenario: one `compliance_pack_activation` row in state APPROVED/EXECUTED with `decided_by IS NULL` or `decided_by = proposed_by` (a hand-seeded or backfilled row — the API path cannot produce one) → `refresh()` throws on **every** tick, `added` is never returned, and **no** pack converges on any replica, permanently. That is the exact opposite of the intent stated three lines above ("one unparseable row must not stop the others from converging"), and the only symptom is a scheduler error line. Non-blocking because reachability is low (`decide()` enforces maker≠checker, and `CompliancePackBootLoader` would already have failed the pod loudly on the same row), but wrapping the `syncFrom` call in the same `runCatching`, or validating per row before batching, would make the code match the comment.

2. `CompliancePackRefresher.kt:88-96` — the KDoc says the mapping is now shared "so the two cannot disagree about what a row means". A third copy still exists: `CompliancePackActivationService.toProposal()` (`application/usecase/CompliancePackActivationService.kt:~127`). It maps a live entity with its real state/id and is not interchangeable with the new one, so this is not a defect — only a claim slightly stronger than the code. Worth a word in the KDoc so the next reader does not delete one of them.

Verified clean: `refresh()` IS a `suspend fun @Scheduled` (lens 3 — no `runBlocking`/HR000068), and `quarkus.scheduler.enabled: true` in this service's `application.yaml` so the scheduler genuinely runs under test. `CompliancePackReplicaConvergenceIT` drives the REAL cron through a `@TestProfile` shrinking `lending.compliance.refresh-interval` to `1s` with literal constants (no companion-object classloader desync), inserts the sibling row over plain JDBC, asserts only what the REST endpoint serves, and asserts the negative precondition first — it does not call the scheduled method directly. `refresh-interval` is present in `application.yaml` with an env override and is consumed via the `@Scheduled(every = "{...}")` expression, so there is no `@ConfigProperty` optional-field (`SRCFG00040`) exposure. Thread safety: `syncFrom` is a pure union inside `getAndUpdate` that re-derives `added` on each CAS retry, and readers go through `active.get()` on an immutable map, so an origination request mid-swap sees the old or the new map, never a partial one. Additive-only is sound here: there is no un-activate transition (`decide()` only moves PROPOSED→EXECUTED/REJECTED), `findActivated()` selects only APPROVED/EXECUTED so a rejected row never converges, and withdrawal is expressed as `effectiveTo`/a superseding version and evaluated at READ time by `isEffectiveOn` — so a withdrawn pack does stop being enforced despite the map never shrinking. The `openbank-libs-domain` change is purely additive (new `syncFrom` + private helper, no existing signature touched), so the fleet rebuild it triggers is mechanical. No API surface change, so no `openapi.yaml` bump is owed; `version.txt` correctly untouched. No data is written or migrated by this PR, so no already-written rows are wrong. The residual ~30s divergence window is bounded and documented — a design point, not a defect.

JiRaska:
**Review: 2 findings, 0 blocking**

The mechanism does what the description says: each pod polls `findActivated()` (which filters
`state in (APPROVED, EXECUTED)` in SQL) and unions the result into the per-pod registry, so the
approval-to-sibling window becomes bounded by `lending.compliance.refresh-interval` instead of
"until the pod restarts".

**1. The four-eyes re-validation on the convergence path cannot reject anything the mapping
produces.**
`CompliancePackRefresher.kt` — `CompliancePackActivationEntity.toProposal()` hardcodes
`state = ProposalState.EXECUTED` and substitutes `decidedBy = decidedBy ?: "boot:unknown-checker"`.
`CompliancePackRegistry.syncFrom` then calls `requireApprovedFourEyes`, which checks exactly those
two fields. The state check is satisfied by construction; the `decidedBy != null` check is
satisfied by the fallback literal; and `decidedBy != proposedBy` can only fail when the row already
carries a real, self-approving checker. So the KDoc's "a row cannot become enforceable by arriving
through the convergence door instead of the approval one" does not hold for the case that matters:
an `compliance_pack_activation` row that is `APPROVED` with `decided_by IS NULL` — a hand-applied
data fix, a partial write, a future code path that sets state before the checker — converges into
**every** pod as an enforceable pack, and the audit trail records the checker as the literal string
`boot:unknown-checker`. The real gate here is the `state in (APPROVED, EXECUTED)` SQL predicate, not
the four-eyes assertion; the honest fix is to skip (and log) a row whose `decidedBy` is null rather
than manufacture one. (The boot loader shared this fallback before, so the behaviour is not new —
the *claim* about it is.)

**2. Convergence is one-directional, and this PR makes that structural rather than incidental.**
`syncFrom` is a union by `(key, version)` and never removes. That is correct under ADR-0212 D3's
never-un-activate invariant and correctly avoids the drop-a-concurrent-write race, but the net
effect is that a pod which is *ahead* — or holds a pack whose row was later corrected, deleted, or
rejected — never converges down, and now never can through this path. That is the open D3 concern
(no de-activation) inheriting a second enforcement point: previously a restart re-derived the
registry purely from the table, so a removed row did stop being enforced on the next roll; with a
union-only refresher, a pod that picked a pack up keeps it for the life of the process regardless of
what the table says afterwards. Worth stating in the KDoc as the limit of "converges", because
"converge" reads as bidirectional and this is a monotone catch-up.

Verified clean: `refresh()` is a `suspend fun` and the reactive repository call is awaited with
`awaitSuspending()` — no `runBlocking`, so no HR000068; and
`CompliancePackReplicaConvergenceIT` drives the **real** scheduler via a `QuarkusTestProfile` that
shrinks `refresh-interval` to `1s` and uses literals in `getConfigOverrides()` (the two-classloader
footgun), rather than calling `refresh()` directly — a direct call could not have detected the
missing-context class of bug. `ConcurrentExecution.SKIP` plus `delayed = "10s"` keeps the boot
loader and the first refresh from racing. Re-applying the same rows is a genuine no-op, so a
repeated refresh is idempotent, and the `added` counter is reset inside the `getAndUpdate` lambda,
so a CAS retry cannot double-count. Choosing a poll over an outbox event is well argued and avoids
the `openbank.outbox.dispatch-enabled` class of silent no-op. No schema change, so no migration and
no spec bump is correct.

---
Automated agent review — **not an approval**. This changes what compliance rules a money-path
origination is judged against; it needs human review regardless.

JiRaska:
Refreshed against `main` (was 51 behind, conflict-free) — the red looked misattributed and the base was stale.

The failure was:

```
FraudPactFolderProviderVerificationTest > verifyPacts > openbank-sepa-payment - POST fraud score for SEPA payment  FAILED
> Task :openbank-fraud-service:test FAILED
```

This PR touches **nothing under fraud** — `git diff --name-only origin/main...HEAD | grep -i fraud` is empty; it changes lending and `openbank-libs-domain`'s `CompliancePackRegistry`. A lending change cannot alter fraud's provider contract.

And it is not `main` either: `main`'s Services CI is green (run `31208125403`, plus the four before it). So the red was a stale run against a 51-commit-old base, not a defect in this PR and not fleet breakage.

(The `model card verification FAILED (card not found: ml/no-such.card.json)` lines in the same log are a negative-path test's expected output, not failures — worth saying, because they read alarmingly next to a real one.)

Re-running against current `main` now.

JiRaska:
`all-green` FAILURE — status, not a blocker to ignore. Both failing service builds investigated; one was infra noise, the other is not.

## `build (openbank-fraud-service)` — was infra flake, now green

First attempt failed with `UnsatisfiedLinkError: libstdc++.so.6: No such file or directory` (ONNX native lib), unrelated to this PR's diff. Re-ran, passed clean on the same head. No action needed.

## `build (openbank-lending-service)` — real, reproducible `OutOfMemoryError`, not flake

Failed identically across **three separate attempts**, on **two different heads**:

| attempt | head | thread | outcome |
|---|---|---|---|
| 1 | `47c5e2f8` | `Test worker` | `OutOfMemoryError: Java heap space` |
| 2 (rerun) | `47c5e2f8` | `Test worker`, `BatchSpanProcessor_WorkerThread-1`, `vertx-blocked-thread-checker` | same |
| 3 | `07fcdca2` (current head) | `executor-thread-0 (scheduler)`, `vertx-blocked-thread-checker` | same |

Different thread each time, same failure, so it isn't one flaky test — it's the whole `:openbank-lending-service:test` task exceeding the runner's heap, consistently enough to survive both a rebase and a rerun. A re-run will not fix this.

I did not investigate further — didn't check whether this reproduces on `main` without this PR's diff (i.e. whether it's a pre-existing capacity issue this PR merely exposes by triggering the fleet rebuild) or whether the PR's own changes to `CompliancePackRegistry`/`CompliancePackRefresher` grew the test suite's memory footprint. That's the next step before deciding whether the fix belongs in this PR or is a separate runner/Gradle-heap issue.

Logs: [attempt 1](https://github.com/JiRaska/open-bank-oss/actions/runs/31190886280/job/92927768731), [attempt 3 on current head](https://github.com/JiRaska/open-bank-oss/actions/runs/31208676875/job/92983262393).

JiRaska:
**Merged `origin/main` in (`b01817fc7`, signed). This branch was 97 commits behind; it is now current, zero conflicts.**

Scope after the merge is unchanged — 6 files, exactly what this PR owns — and its own content survived: `syncFrom` is present in `CompliancePackRegistry.kt`, `CompliancePackRefresher.kt` is there. Checked by counting rather than eyeballing the diff, since a clean merge in this repo has silently dropped a neighbouring entry before.

**One local test failed and I do not claim to know whether it is real:**

```
CompliancePackReplicaConvergenceIT > a pack approved on a sibling replica becomes active here without a restart()
java.net.SocketTimeoutException: Read timed out
```

35 of 36 passed. That signature is a container/HTTP timeout rather than an assertion failure, and this machine was running several heavy Gradle builds concurrently — so it looks environmental. But "looks environmental" is not evidence, and the honest read is that CI on the fresh base is the arbiter. Flagging it rather than reporting green.

---

**A finding for this PR's body that is not currently stated anywhere, and it strengthens the case.** The issue and this PR both describe the divergence as an inconsistency. It is that in one direction and a **fail-open** in another:

- **First activation → fails CLOSED.** A stale pod's registry returns null, `checkOriginationAllowed` throws `no active compliance pack … origination refused`. Pod A allows, pod B refuses the same request. Confusing, safe.
- **Version upgrade → fails OPEN, silently.** `activePack` returns `maxByOrNull { it.pack.version }` among effective packs. If v2 tightens the rules, the stale pod still holds v1 — **non-null**, so `checkOriginationAllowed` passes on both pods — and `mergePackChecks` merges *v1's* `mandatoryEligibilityRules` into the policy bundle. The loan is decisioned against superseded, weaker compliance rules with **no error, no log, no refusal**.

So the accurate severity is not a four-eyes bypass — maker/checker separation is enforced against the persisted row and is untouched by pod-local state — but a **compliance-rule-version bypass on the money path**, latent behind `replicas: 1` with `LENDING_ENFORCE_PACK=true` live. The additive `syncFrom` in this PR closes that direction too.

**Also worth correcting an assumption in the issue thread:** Redis is *already* wired in lending (`libs.quarkus.redis.client`, `redis.hosts` in `application.yaml`, `RedisApprovalStore` per ADR-0155, and the threat model already names it), so a Redis-backed option would not have been a new dependency. This PR's choice of DB polling still holds on its own merits — the activation row is written in the same transaction as the four-eyes state change, so polling it cannot drift from it, whereas a Redis copy is a second copy with a second way to be missed.

Still needs its 2 human approvals — lending is money-path.

JiRaska:

```
```

JiRaska:
The red on this PR is **one gate**, and the cause is in the diff rather than anywhere upstream.

`gates (gitops)` reports `27 gates PASS=26 FAIL=1` — `scheduler-liveness-adoption (ADR-0237)`,
*NEW domain @Scheduled job without registerWorkflowLiveness*. `Validate manifests` is only the
aggregator over that shard, so it is the same single failure counted twice. Re-ran it against this
branch's current head `5414c336d`, with the CI environment supplied
(`BASE_REF=main PR_DIFF_BASE=$(git merge-base HEAD origin/main)` — without those, diff-scoped gates
manufacture a failure that reads like a real defect):

```
1 gates  FAIL=1
::error title=scheduler-liveness-adoption::gate failed: scheduler liveness adoption (ADR-0237)
```

The gate's own error line does not name the file; the attribution is from the diff —
`CompliancePackRefresher` is the one new `@Scheduled` and it registers no heartbeat.

Worth saying that the gate is *right* here rather than merely bureaucratic, and for this PR's own
reason. `refresh()` logs only when `added > 0`, and the KDoc says so explicitly: *"Silence here is
the healthy state."* So a refresher that stops running entirely produces exactly the same log output
as one that is working — which is the situation this PR exists to end for the registry itself.

The shape the fleet uses (`FraudHoldService` is the closest sibling):

```kotlin
@Inject
lateinit var domainMetrics: DomainMetrics

private var liveness: WorkflowLivenessRecorder? = null

// StartupEvent, not @PostConstruct: @ApplicationScoped is LAZY, so a @PostConstruct would first
// run when the cron first fires and the gauge would be ABSENT until then — and absent is a
// different signal from stale.
fun registerLiveness(@Observes @Suppress("UNUSED_PARAMETER") event: StartupEvent) {
    liveness = domainMetrics.registerWorkflowLiveness(WORKFLOW_NAME, EXPECTED_INTERVAL)
}
```

with `recordSuccess()` as the last statement of `refresh()` — on the completed path only, never in a
`catch`, or the heartbeat asserts the very thing it exists to disprove. `EXPECTED_INTERVAL` should
match the `@Scheduled` default (`LENDING_COMPLIANCE_REFRESH_INTERVAL:30s`); the staleness rule
already applies a 2x multiplier, so this is the schedule and not a tighter SLA.

I did not push this to the branch. `d88656bbb` in the `.claude/worktrees/agent-ad94eb6a36cd4f736`
checkout is **not an ancestor** of the pushed head `5414c336d`, so that working copy either holds
unpushed commits or the branch was rewritten — either way a push from outside could destroy work
that exists in exactly one place.

JiRaska:
`build (openbank-lending-service)` has failed 4 times in a row with the identical signature: `CompliancePackReplicaConvergenceIT > a pack approved on a sibling replica becomes active here without a restart()` throwing `java.net.SocketTimeoutException` at the same line (`CompliancePackReplicaConvergenceIT.kt:128`, the REST GET call itself), preceded each time by `OutOfMemoryError: Java heap space` and repeated `Connection refused: localhost/127.0.0.1:4317` (OTLP exporter).

This is the shared-runner OOM class documented in the repo's own CLAUDE.md ("Self-hosted runners share the machine with a human"), not a logic defect:
- The failure is a raw socket timeout on the HTTP call itself (server unresponsive), not an assertion mismatch on the response body — if the convergence logic were wrong, the 30s poll loop would time out normally and the `assertThat(body).contains(...)` would fail with a clear diff, not throw `SocketTimeoutException`.
- Ran the same test in isolation, locally, fresh (not cached — `TEST-...xml` timestamp confirms a real re-execution): **PASSED, 9.4s, `failures="0" errors="0"`.**
- The refresh scheduling math has ample margin: `delayed = "10s"` + the test profile's `1s` refresh interval leaves ~19s of slack inside the test's own 30s convergence budget.

Triggered a 5th rerun. If it keeps failing the same way, it's a runner-capacity question, not something to chase in the PR's own code further.

PR #3653, opened 2026-08-03
fix(ci): agent-review must not report a review that never happened

JiRaska:
## What

`agent-review` currently reviews **nothing**, and roughly a quarter of its runs report that as a red attributed to #2161 while the rest report it as a green.

The trigger is not #2161 and it is not intermittent: **GitHub Models was fully retired on 2026-07-30**. Every `actions/ai-inference` call now answers

```
410 GitHub Models is temporarily unavailable as part of a scheduled retirement brownout.
```

The primary reviewer is permanently dead. #2161 (the dead OAuth fallback) only decides what happens *next* — and the answer is either a red nobody can act on, or a green.

## Measurement

120 `Agent review` runs, 2026-08-02T15:52Z .. 2026-08-03T11:18Z, classified from **step conclusions** (`gh api .../actions/runs/<id>/jobs --jq '.steps[]'`), not log text:

| outcome | runs |
| --- | ---: |
| deferred-to-human (sensitive scope — no model ever called) | 75 |
| job skipped by the workflow `if:` | 15 |
| primary failed → dead Claude fallback → job **RED** | 27 |
| primary failed → fallback **skipped** → job **GREEN**, nothing reviewed | 2 |
| checkout/git transport 503 | 1 |
| **a review actually submitted, by any model** | **0** |

All 12 primary failures inspected in detail carried the 410, unanimously. So the "~40% failure" in #3488 is **100% of runs that attempted a review**; the successes are runs that never called a model. That also explains the minute-by-minute interleaving that looked like a concurrency limit — it is deferred-vs-attempted, not load.

One measurement note worth recording: the ai-inference step's own conclusion reads `success` even when it failed, because `continue-on-error: true` rewrites the reported conclusion. The discriminator is the downstream gated step — `Submit review (GitHub Models)` skipping means `steps.ghmodels.outcome == 'failure'`.

## Why a new step and not a tighter guard

The two GREEN rows. `check-claude-fallback-result.sh` is gated on the Claude step having **run** (`steps.claude.outcome != 'skipped'`). On a Dependabot or fork PR the OAuth secret is empty in the job context, the fallback is skipped, the guard is skipped with it, and the job finishes green having reviewed nothing. On the PR page that is indistinguishable from a review that happened — and it is the *default* reading the moment the fallback stops running for any reason.

## What this changes

- Adds `.github/scripts/check-agent-review-happened.sh`, a terminal accountability guard driven off step outcomes. It passes only when a review demonstrably happened, or was deliberately not attempted (sensitive scope deferred to a human, which already posts a comment).
- It **names which of the four causes** it was — retired primary, dead OAuth token, the action refusing a bot actor, or a skipped fallback — instead of letting all four read as "the known #2161 issue".
- On failure it **posts the verdict as a PR comment** (best-effort; the token is read-only on Dependabot and fork PRs). A job conclusion alone is a weak carrier and is the only place this state appears today.
- The #2161 transcript check keeps its own step and its own red, unchanged; it gains only an `id` so it can report its verdict to the terminal guard via `steps.fallback_guard.outcome`.
- Workflow header now records the retirement, so the next reader does not re-diagnose it.

Two gates in this repo shaped that design, both found by running them rather than by predicting them:

1. The first push had the terminal guard *calling* `check-claude-fallback-result.sh`. `check-gate-script-registration.py` (#3240, enforced) went red: it deliberately does not count a caller that lives in `.github/scripts`, so the #2161 guard would have read as an orphan nothing runs. Fixed by making both steps workflow-invoked, one passing its outcome to the other.
2. The second push gave that step `continue-on-error: true`, so a single step would own the verdict. `check-advisory-gate-registration.py` (#2392, enforced) went red, correctly: `continue-on-error` turns an enforced guard into an advisory gate, and an advisory gate with no `target_enforce_date` in `rules.yaml` is one that stays advisory forever. Reverted — the step keeps its own red, and the terminal guard is what turns that red into a named cause and a PR comment.

The job stays advisory and is not a required check, so a red still means "no review happened", never "this PR is rejected".

## What this deliberately does NOT do

Restoring an actual reviewer means pointing the primary at a live inference endpoint. That is a credential decision for the owner and is out of scope here. No secret is created, read or printed by this change.

## Falsification

`--self-test` runs 8 cases — 3 that MUST pass and 5 that MUST fail, each taken from a state observed in the sample above, including the exact green-with-no-review case that motivated the change. Verified green under macOS bash 3.2 and bash 5.2. `shellcheck` and `actionlint` clean; `run-gates.py --group registry` is PASS=14 WARN=1 FAIL=0 locally (the warning is a pre-existing advisory about `rest.rego`, untouched here). Largest `run:` step in the workflow is 7587 chars, well under the 19000 limit (#3135); the change shortens the `run:` block rather than growing it, with the reasoning in the script header.

Validated against **GitHub**, not only a YAML parser: the first push produced a run titled `Agent review` (so the file parsed and `name:` was read, not a run titled after the path), and the new step executed for real with substituted values — `step outcomes: scope_skip=true primary=skipped submit=skipped claude=skipped has_claude=true` — passing on the deferred path, which is the correct verdict for a `.github/*` PR.

## Risk

Low. The only behavioural change is that a run which reviewed nothing is now red and says so, where two of the sampled runs were previously green. The workflow is advisory and not a required status check.

## Linked issues

Refs #3488
Refs #2161

### DIFF
--- a/.github/scripts/check-agent-review-happened.sh
+++ b/.github/scripts/check-agent-review-happened.sh
@@ -1,0 +1,232 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) OpenBank contributors. Licensed under the Apache License, Version 2.0.
+#
+# Terminal accountability guard for `agent-review`: a run that reviewed nothing must not
+# be readable as a review (issue #3488; the fallback half is #2161).
+#
+# WHY THIS EXISTS — measured, not assumed
+#   120 `Agent review` runs sampled over 2026-08-02T15:52 .. 2026-08-03T11:18 (classified
+#   from step conclusions, not from log text) contained ZERO model reviews:
+#
+#     75  deferred-to-human (sensitive scope — correct, no model was ever called)
+#     15  job skipped by the workflow `if:`
+#     27  primary reviewer failed -> dead Claude fallback -> job RED
+#      2  primary reviewer failed -> fallback SKIPPED  -> job GREEN, nothing reviewed
+#      1  checkout/git transport 503
+#      0  a review actually submitted by any model
+#
+#   The trigger is not intermittent and it is not a rate limit. Every single call to the
+#   primary reviewer returned:
+#
+#     410 GitHub Models is temporarily unavailable as part of a scheduled retirement brownout.
+#
+#   GitHub Models was fully retired on 2026-07-30 (GitHub Changelog, "GitHub Models is now
+#   retired"). The primary reviewer is therefore permanently dead, not flaky. The apparent
+#   ~40% pass rate in #3488 was runs that never called a model at all: the successes are
+#   deferred-to-human and skipped runs, and 100% of runs that attempted a review failed.
+#   Successes and failures interleaving minute by minute is that mix, not concurrency.
+#
+#   Two of the sampled runs are the reason this file exists rather than a tweak to
+#   check-claude-fallback-result.sh. That guard is gated on the Claude step having RUN
+#   (`steps.claude.outcome != 'skipped'`), so when the fallback is skipped — a Dependabot or
+#   fork PR, where `secrets.CLAUDE_CODE_OAUTH_TOKEN` is empty in the job context — the guard
+#   is skipped too and the job finishes GREEN having reviewed nothing. On the PR page a green
+#   `agent-review` reads exactly like a review that happened. That is the default reading once
+#   the fallback stops running for any reason, so it is the state worth closing.
+#
+# WHAT IT CHECKS
+#   Drives off STEP OUTCOMES, never log text (a job log embeds each step's own `run:` script,
+#   so grepping it matches strings that never executed). Passes only when a review demonstrably
+#   happened or was deliberately not attempted:
+#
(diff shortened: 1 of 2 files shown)


JiRaska:
All five required contexts are green on `b633abc` (`all-green`, `Validate manifests`, `Gitleaks`, `issue-hygiene`, `OPA policy gate`). Single signed commit, `verification.verified = true` on the branch.

Not merged, deliberately: this touches `.github/workflows`, which requires human review regardless of the PR title. `gh pr merge --squash --auto` refuses, and that is the correct final state — no override attempted.

Live validation on this PR's own runs, since `.github/*` classifies as sensitive: `Agent review` [run 30810272213](https://github.com/JiRaska/open-bank-oss/actions/runs/30810272213) shows both guard steps present and `Verify a review actually happened` executing and passing on the deferred path, printing substituted values (`step outcomes: scope_skip=true primary=skipped submit=skipped claude=skipped has_claude=true`). The run is titled `Agent review`, not the file path, so GitHub parsed the modified workflow.

The no-review failure path cannot be exercised on this PR — a sensitive-scope PR never calls a model — which is why it is covered by the 5 must-fail `--self-test` cases instead.

PR #3658, opened 2026-08-03
fix(psd2): answer 400, not 500, for a missing required query/header parameter

JiRaska:
## Summary

21 of the 39 non-money-path handlers in #3624 that answer 5xx (or worse, silently proceed) when a required query/header parameter is absent — the two services the issue names as priority. Each parameter is declared nullable and guarded, so the caller gets a 400 that names what is missing.

**Stacked.** Base is `fix/nonnull-queryparam-400` (#3625), which owns the gate script and the money-path half. This PR cannot and must not reach `main` ahead of it. Retarget to `main` once #3625 lands.

## Type of change

- [x] fix (bug fix)

## Linked issues

Refs #3624, Refs #3625, Refs #3104

## Changes

- **openbank-psd2-service (11 sites).** `Consent-ID` is mandated on every AIS/PIS call by the Berlin Group NextGenPSD2 spec; a TPP that omitted it was told the ASPSP was broken. Guards throw `Psd2RequestFormatException`, **not** `require()`, so the 400 carries this service's Berlin Group `tppMessages` envelope rather than libs-runtime's generic `ApiError` — the shape `ExceptionMappers.kt` already documents (#526), and the same form `ObConsentResource` uses for `X-Request-ID`.
- `PisResource`'s four initiation handlers funnel through one private helper, so both headers are validated once there. That helper already carried `if (idempotencyKey.isBlank()) throw ...` — **a guard that answered 500 in exactly the case it was written for**, because these handlers are `suspend`, no `Intrinsics.checkNotNullParameter` is emitted, and `null.isBlank()` threw NPE. Only a *blank* key ever reached the intended 400.
- **openbank-card-issuance-service (10 sites).** `X-Operator-Id` is the audit attribution on every card lifecycle write (activate/block/cancel/suspend/resume/limits/controls). Deliberately **not** given the `?: "unknown"` fallback `secureDetails` uses — that is a read; these seven change a card's state and must not be attributed to a placeholder operator. `issueCard`'s `Idempotency-Key` had the same never-reachable blank-guard. `CardDelegationResource`'s `partyId`/`intent` name what is being asked, so neither has a defensible default.
- Baseline entries for both services removed from `check-nonnull-jaxrs-params.py` in the same change. The gate errors on a stale entry, which is what keeps code and list in step.

### Classification

All 21 are **truly required**. Defaultable: 0. Genuinely optional: 0.

**Two** of the guards already written here could never run for the absent case, covering **five** of the 21 sites: psd2's lives in the private helper all four initiation handlers funnel through (4 sites), and card-issuance's is in `issueCard` (1 site).

### Testing, and what was actually measured

`Psd2MissingHeaderStatusIT` drives the AIS accounts read over **real HTTP** — the only layer at which this defect exists. A unit test calling the handler supplies the argument JAX-RS does not, so it passes against the broken signature.

Falsified by reverting the fix and re-running. Worth recording: the measured pre-fix status there was **401, not the 500 the issue predicts**. The handler is `suspend`, the null flowed into the body, and the TPP-certificate check short-circuited first. With a valid certificate nothing short-circuits and the null is carried into `GetAccountsQuery.consentId` — the silent-null outcome the gate's own header calls worse than a 500. The 401 was a lie in its own right: it blamed the caller's certificate for a missing header.

The second test is a control that passes both before and after, so the first test's redness is attributable to the guard alone. A `@TestProfile` turns `authz.enforce` off: without it there is no OPA sidecar under test, the PDP fails closed, and both assertions saw **503** — the IT would have been green about nothing.

## Definition of Done

- [x] Contract tests updated (if public API changed). — no pact change needed; both services' provider replays pass unchanged.
- [x] `./gradlew detekt ktlintCheck koverVerify build` passes. — per module, run sequentially (nine parallel Quarkus test JVMs collide on ports and produce `QuarkusBindException`).
- [x] No new lint warnings.
- [ ] OpenAPI spec regenerated (if REST API changed). — no route, parameter set or schema changed; only the status code for an already-invalid request.
- [x] Flyway migration added + rollback note (if DB changed). — n/a
- [x] Avro schema versioned backward-compatibly (if event changed). — n/a
- [x] Docs updated (README, ADR if architectural). — the mechanism is already documented in root `CLAUDE.md` and the gate script header.

## Security checklist

- [x] Auth / crypto / payment code changed? — no. The change is strictly the status code for a request that was already invalid; no authorization decision moves.
- [x] PII path changed? — no.
- [x] Cardholder data path changed? — no. `secureDetails` is untouched.

## Compliance impact

- [x] PSD2 / SCA / RTS — a Berlin Group error response shape is now returned where a 5xx/401 was. Strictly closer to the spec: a missing mandated header is a client error.

### DIFF
--- a/.github/scripts/check-nonnull-jaxrs-params.py
+++ b/.github/scripts/check-nonnull-jaxrs-params.py
@@ -75,18 +75,14 @@ PARAM_ANNOTATIONS = ("QueryParam", "HeaderParam", "MatrixParam")
 # JVM primitives: JAX-RS supplies the zero value, never null, so no NPE is possible here.
 KOTLIN_PRIMITIVES = {"Int", "Long", "Double", "Float", "Boolean", "Short", "Byte", "Char"}
 
-# The tail left after the money-path fix in #3104. Each entry is "<service>|<Class>|<param>".
+# The tail left after the money-path fix in #3625, being worked through in #3624. psd2 and
+# card-issuance are done. Each entry is "<service>|<Class>|<param>".
 # The count per key is not recorded on purpose: several of these repeat the identical parameter
 # across sibling handlers (card-issuance's X-Operator-Id sevenfold), and pinning the count would
 # make an unrelated refactor fail the gate for no defect.
 BASELINE = {
     # openbank-aml-service — POST /api/v1/aml/cases
     "openbank-aml-service|AmlCaseResource|Idempotency-Key",
-    # openbank-card-issuance-service — operator attribution on every card lifecycle operation
-    "openbank-card-issuance-service|CardResource|Idempotency-Key",
-    "openbank-card-issuance-service|CardResource|X-Operator-Id",
-    "openbank-card-issuance-service|CardDelegationResource|partyId",
-    "openbank-card-issuance-service|CardDelegationResource|intent",
     # openbank-customer-edge — the mobile/web BFF
     "openbank-customer-edge|CustomerEdgeResource|currency",
     "openbank-customer-edge|CustomerEdgeResource|accountId",
@@ -98,10 +94,6 @@ BASELINE = {
     "openbank-pid-service|PartyResource|index",
     "openbank-pid-service|PartyResource|type",
     "openbank-pid-service|PartyResource|value",
-    # openbank-psd2-service — Berlin Group headers on AIS/PIS
-    "openbank-psd2-service|AisResource|Consent-ID",
-    "openbank-psd2-service|PisResource|Consent-ID",
-    "openbank-psd2-service|PisResource|Idempotency-Key",
     # openbank-statement-service
     "openbank-statement-service|StatementResource|from",
     "openbank-statement-service|StatementResource|to",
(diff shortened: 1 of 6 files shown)


JiRaska:
### Reviewer note: guard ordering differs between the two psd2 resources, deliberately left alone

`AisResource` runs the `Consent-ID` guard **before** resolving `tppId`; `PisResource` runs it **after**, because that is the order its existing helper already had and I did not want to move an authentication-ish check as a side effect of a status-code fix.

On the deployed path this is not observable: `EidasMtlsFilter` gates every `open-banking/` route at `@Priority(AUTHENTICATION)` and aborts with 401 `CERTIFICATE_MISSING` before either handler is entered, so an uncertified caller never reaches either guard. The in-handler `?: return missingTpp()` is a fallback for the ungated/sandbox case.

I probed the difference under `@QuarkusTest` and got two results I could not reconcile (`X-TPP-ID` alone → 400, `X-TPP-ID` + `Consent-ID` → 401, from requests the filter cannot distinguish). Rather than change ordering on the strength of a probe I do not understand, I left both resources as they are and am flagging it. If you want them aligned, `PisResource`'s order is the one to copy — validation after authentication leaks less — but it would need the IT reworked, since the 400 is then only reachable with an established `tppId`.

Nothing here is a regression: every case that previously answered 500, or a misleading 401, now answers 400 or better.

PR #3660, opened 2026-08-03
fix(customer-edge): answer 400, not 500, for a missing required query/header parameter

JiRaska:
## Summary

The remaining 18 of the 39 sites in #3624, across seven services — and the last of them. `BASELINE` in `check-nonnull-jaxrs-params.py` is now **empty**, so the gate becomes an outright "no occurrences" ratchet rather than a tail-tracking one.

**Stacked.** Base is `fix/nonnull-params-psd2-cards` (#3658), which is itself based on `fix/nonnull-queryparam-400` (#3625). Merge order is #3625 → #3658 → this. Retarget as each lands.

## Type of change

- [x] fix (bug fix)

## Linked issues

Refs #3624, Refs #3625, Refs #3104

## Changes

Same mechanism throughout: JAX-RS injects `null` for an absent parameter, Kotlin's null-safety is compile-time only, and the parameter must be declared **nullable** before any guard in the body can run. A plain `fun` emits `Intrinsics.checkNotNullParameter` at bytecode offset 0 (guaranteed 500, and a guard written in the body compiles to nothing); a `suspend fun` emits no intrinsic, so the null flows in and fails at the first dereference — or silently does not.

**The error shape chosen per service is the one that service already uses for its 400s.** Introducing a second envelope into a resource that answers every other 400 with an explicit `Response` would be the #526 shape-lottery all over again.

- **customer-edge (5)** — `accountId` on four reads/writes plus `currency` on pocket resolve. Answered with this resource's own `badRequest()`, exactly as `listSddMandates` already does for an absent `accountId`; the app parses that shape. All are `@Blocking` plain `fun`s, i.e. the guaranteed-500 half.
- **statement (4)** — `from`/`to` on close and export. A period-close assigns legal sequences and is idempotent on `(account, currency, period)`, so a guessed range would write the **wrong** statement rather than fail; no default is defensible. `format` needed nothing — it already carries `@DefaultValue`.
- **pid (3)** — `resolveByIndex` already returned the right `INVALID_PARAMETER` 400 for a *blank* index and could never return it for an *absent* one. Widened to `isNullOrBlank()`, which keeps the existing response shape for both cases rather than swapping in a new one. `getByExternalId`'s `(type, value)` pair **is** the lookup key.
- **dispute (2)** — `actor` on withdraw/escalate is the attribution recorded against the state change; a default would write a placeholder into the dispute's history.
- **tpp-registry (2)** — `tppId`/`role` name what is being checked; a guessed role answers a question the caller did not ask.
- **aml (1), party (1)** — `Idempotency-Key`. aml's `require(idempotencyKey.isNotBlank())` is another guard that answered 500 in exactly the case it was written for. party's key is what makes a repeated create safe — synthesising one would make every retry a **new party**.

### Classification

All 18 **truly required**. Defaultable: 0. Genuinely optional: 0. Two of the 18 already had a guard that could never run for the absent case.

### Testing

`StatementMissingParamStatusIT` covers the **non-suspend** half over real HTTP — the only layer at which the defect exists. Falsified: with the fix reverted it reports **500**, and its control test (both dates supplied) passes either way, so the redness is attributable to the guard alone. A `@TestProfile` turns `authz.enforce` off; without it the PDP fails closed and every `@Authorize` endpoint answers 503 before the handler is entered.

This required `rest-assured-kotlin` and `quarkus-test-security` on statement-service's **test** classpath — the one dependency change in this PR, and the reason the fleet dependency-resolution job runs.

## Definition of Done

- [x] Contract tests updated (if public API changed). — no pact change needed; tpp-registry's and the others' provider replays pass unchanged.
- [x] `./gradlew detekt ktlintCheck koverVerify build` passes. — per module, run sequentially (parallel Quarkus test JVMs collide on ports and produce `QuarkusBindException`).
- [x] No new lint warnings.
- [ ] OpenAPI spec regenerated (if REST API changed). — no route, parameter set or schema changed; only the status code for an already-invalid request.
- [x] Flyway migration added + rollback note (if DB changed). — n/a
- [x] Avro schema versioned backward-compatibly (if event changed). — n/a
- [x] Docs updated (README, ADR if architectural). — gate script header updated to state the baseline is now empty and why the ratchet machinery stays.

## Security checklist

- [ ] No new third-party dependency, OR: dependency review attached. — two **test-only** additions, both already in the version catalog and already used elsewhere in the fleet (`io.rest-assured:kotlin-extensions`, `io.quarkus:quarkus-test-security`). Nothing reaches a runtime classpath.
- [x] Auth / crypto / payment code changed? — no. customer-edge's IDOR ownership checks are untouched: the guard runs strictly *before* them and every `ownsAccount` call still receives the same value it did.
- [x] PII path changed? — no.
- [x] Cardholder data path changed? — no.

## Compliance impact

- [x] None — no regulatory surface changes; an already-invalid request now gets an accurate status code.

### DIFF
--- a/.github/scripts/check-nonnull-jaxrs-params.py
+++ b/.github/scripts/check-nonnull-jaxrs-params.py
@@ -56,7 +56,10 @@
 #   (@DefaultValue), or actually optional (nullable) — and only the handler knows which. The
 #   money-path set is fixed; the remainder is BASELINED with its issue so a NEW occurrence fails
 #   while the known tail stays visible. A baseline entry that no longer occurs is reported too, so
-#   the list cannot rot in either direction.
+#   the list cannot rot in either direction. As of #3624 the BASELINE is EMPTY — both halves are
+#   remediated — so today this is an outright "no occurrences" gate. Keep the ratchet machinery
+#   anyway: it is what lets the next genuinely-optional-but-unfixed site be recorded rather than
+#   silently suppressed.
 #
 # EXIT CODES
 #   0  no new occurrences, no stale baseline entries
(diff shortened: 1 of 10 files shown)


JiRaska:
## Stack order note (2026-08-06)

This PR is stacked and cannot be merged in isolation — its base branch is `fix/nonnull-params-psd2-cards` (#3658), which itself bases on `fix/nonnull-queryparam-400` (#3625). Both parents are still `OPEN`, unmerged.

Required merge order: **#3625 → #3658 → this (#3660)**. Merging this PR before its parents would pull in their unreviewed commits along with it. Also note all three PRs in the chain touch money-path-adjacent services (aml/dispute/party/customer-edge) and are gated the same way individually — same review requirement applies to each link.

Retarget each PR onto `main` as its parent lands, per the PR body's own note.

PR #3677, opened 2026-08-03
refactor(libs): move framework-bound classes from libs-domain to libs-runtime

### LINKED ISSUE #3670: libs-domain leaks framework imports: @Authorize/@FeatureFlag are InterceptorBinding, OutboxDispatch imports JBoss Logger — ADR-0122
ADR-0122 promises openbank-libs-domain has zero framework imports. Three violations:
1. libs-domain/.../authz/Authorize.kt — jakarta.interceptor.InterceptorBinding (CDI) declared in domain.
2. libs-domain/.../flags/FeatureFlag.kt — same.
3. libs-domain/.../persistence/outbox/OutboxDispatch.kt — org.jboss.logging.Logger.

Also suspicious: OpaSidecarPolicyDecisionPoint sits in libs-domain while doing HTTP+Jackson (infra adapter).

Scope: move the annotations to openbank-libs-runtime keeping package names (consumer imports unchanged), remove the Logger, assess moving OpaSidecarPolicyDecisionPoint, add a CI/ArchUnit guard so libs-domain stays framework-free.

JiRaska:
## Summary

ADR-0122 promises `openbank-libs-domain` has zero framework imports, but the `@Authorize`/`@FeatureFlag` CDI interceptor bindings, a JBoss logging facade in `OutboxDispatch`, and the HTTP adapters `OpaSidecarPolicyDecisionPoint` / `FlagdProvider` all lived in the domain module. They move to `openbank-libs-runtime` **under the same package names**, so no consumer import changes; a new enforced gate keeps the domain module framework-free from now on.

## Type of change

- [ ] fix (bug fix)
- [x] refactor (no behavior change)
- [ ] perf (performance)
- [ ] docs
- [ ] chore (build / tooling)
- [ ] breaking change

## Linked issues

Closes #3670

## Changes

- `authz`: `Authorize`, `OpaSidecarPolicyDecisionPoint` (+`PolicyDecisionException`) → libs-runtime
- `flags`: `FeatureFlag` (+`FeatureDisabledException`), `FlagdProvider`, `LoggingExposurePublisher` (split out of `FlagExposure.kt`; data class + `ExposurePublisher` port stay in domain) → libs-runtime
- `security`: `StaticServiceTokenProvider` (split out of `ServiceTokenProvider.kt`; interface stays in domain) → libs-runtime
- `persistence.outbox`: `OutboxDispatch` → libs-runtime (AGENTS.md already assigned outbox dispatchers to runtime)
- Tests move with their subjects; `FlagExposureTest` split (data-class tests stay, logging-publisher test moves)
- `libs-domain/build.gradle.kts`: drops `compileOnly` jakarta + jboss-logging — the module no longer needs them
- `libs-runtime/build.gradle.kts`: adds `compileOnly jakarta.interceptor-api` (matches the Quarkus BOM line)
- Detekt baseline entries move with their files (domain 122 → 116, runtime 24 → 30)
- One smart-cast reworked to `?.let` in `OpaSidecarPolicyDecisionPoint` (`AuthzQuery` is now cross-module)
- New enforced gate `libs-domain-framework-free` (`.github/scripts/check-libs-domain-purity.sh` + `gates.yaml`) — greps domain sources for framework imports and fully-qualified jakarta annotations; falsification-tested locally (catches both axes, ignores kdoc prose)
- Two kdoc links retargeted to plain text (link target left the module)

## Definition of Done

- [x] Hexagonal layering preserved (domain has zero framework imports) — now *verified*, this PR is the fix.
- [x] Unit tests added/updated (moved with subjects; no behavior change).
- [ ] Integration tests added/updated (if service boundary involved) — N/A, no boundary change.
- [ ] Contract tests updated (if public API changed) — N/A.
- [x] `./gradlew detekt ktlintCheck koverVerify build` passes — `:openbank-libs-domain:build :openbank-libs-runtime:build :openbank-libs-testing:build` + ktlintCheck green locally; consumer compile sanity (`:openbank-ledger-service`, `:openbank-sepa-payment`, `:openbank-agent-service` `compileKotlin`) green.
- [x] No new lint warnings.
- [ ] OpenAPI spec regenerated (if REST API changed) — N/A.
- [ ] Flyway migration added + rollback note (if DB changed) — N/A.
- [ ] Avro schema versioned backward-compatibly (if event changed) — N/A.
- [ ] Docs updated (README, ADR if architectural) — ADR-0122 already describes this end-state; no doc change needed.

## Security checklist

- [x] No new third-party dependency (one compileOnly artifact added to runtime that the platform BOM already ships).
- [ ] Auth / crypto / payment code changed? → tag `security-review-required` — code moved verbatim, no logic change.
- [ ] PII path changed? → tag `gdpr-review-required` — N/A.
- [ ] Cardholder data path changed? → tag `pci-review-required` — N/A.

## Compliance impact

- [ ] PCI DSS
- [ ] DORA
- [ ] GDPR
- [ ] PSD2 / SCA / RTS
- [ ] AML / 5AMLD
- [ ] CNB reporting
- [x] None


JiRaska:
Heads-up — this one can drop a line silently, which is why I am flagging it rather than leaving it to the rebase.

This PR deletes `LoggingExposurePublisher` from `openbank-libs-domain` and moves it to `openbank-libs-runtime`. #3882 (merged) edited that exact block, adding the exposure's own `timestamp` to the log line:

```kotlin
"flag exposure exposureId=%s at=%s flag=%s variant=%s key=%s reason=%s traceId=%s",
exposure.exposureId,
exposure.timestamp,
```

Because your side deletes the whole block, git will merge that **cleanly** and take the deletion — no conflict, no output, exit 0 — and the `at=` field is simply not there in the moved copy. Please re-add it in the libs-runtime version.

Rationale: the log line's own time is when the fallback publisher ran, not when the subject was exposed to the variant — and `FlagExposure.timestamp` previously defaulted to `Instant.EPOCH`, so nothing recorded the real time at all.

Nothing goes red if it is lost: no test asserts the publisher's output, and no consumer parses the log line.

Note #3808 also rewrites this same statement (to `System.Logger`), so the two PRs collide with each other here independently of #3882.

The `Instant.now()` default on the `FlagExposure` data class is in a different hunk and does not conflict.

JiRaska:
**Review: 3 findings, 1 blocking (procedural — this PR and #3808 do the same move).** Automated agent review, not an approval. PR is `DIRTY`; reviewed as-is, no rebase attempted.

The catastrophic case for a PR that moves `@Authorize` and the OPA PDP between modules is that the `@InterceptorBinding` silently stops resolving and every `@Authorize` becomes a no-op with nothing failing. **I checked and this move is structurally sound**: the FQCN stays `com.openbank.libs.authz.Authorize`, `AuthorizeInterceptor` already lives in `openbank-libs-runtime`, both modules apply the jandex plugin, the `libs-domain` copy is deleted rather than duplicated, and all 44 services carrying `@Authorize` in `src/main` depend on `:openbank-libs-runtime`. `OpaSidecarPolicyDecisionPoint`/`AllowAllPolicyDecisionPoint` are plain classes wired per-service by `@Produces`, not CDI beans, so moving the PDP cannot silently swap the production decision point for the allow-all stub — a mis-move would be a compile error, which is the loud direction.

Findings:

1. **BLOCKING (procedural): #3808 in the open queue performs the same move and registers a competing gate for the same invariant.** #3808 moves `Authorize.kt` and `FeatureFlag.kt` to `libs-runtime` identically, cites the same issue (#3670), and adds gate `domain-purity-gate` (`check-domain-purity.py`) covering `openbank-libs-domain` module-wide; this PR adds gate `libs-domain-framework-free` (`check-libs-domain-purity.sh`) covering the same module. Landing both puts two implementations of one rule in `Validate manifests` on every PR, to drift apart later. Worse, the file overlap means a conflict resolution can reinstate the `libs-domain` copy of `Authorize.kt` next to the `libs-runtime` one — two classes with the same FQCN on one classpath is exactly the state where the interceptor binding can stop matching *silently*. Decide which PR owns the move before merging either; if this one wins, #3808's Python gate should be ported over, and vice versa. (`gh pr view` computes the file list against the merge-base, so read the content, not the diff, when comparing.)

2. **`check-libs-domain-purity.sh` has no `selftest:` in its `gates.yaml` entry, and its fully-qualified axis only covers `jakarta.`** The import axis matches `jakarta|io.quarkus|org.jboss|io.smallrye|org.eclipse.microprofile|io.micrometer|software.amazon`; the FQ-reference axis is `@[[:alnum:]:]*jakarta\.` only. So a fully-qualified `@io.quarkus.runtime.annotations.RegisterForReflection` on a `libs-domain` type — a violation that already exists elsewhere in this tree, in `openbank-product-catalog` — passes. That is half of the exact blind spot the guard exists to close, and with no self-test the gate is unfalsified in this repo's sense: it has only ever passed. #3808's Python version covers both axes for all prefixes and ships `--self-test`.

3. **The FQ axis strips only kdoc continuation lines (`^[0-9]+:[[:space:]]*\*`), not `//` comments.** A line comment mentioning `@jakarta.interceptor.InterceptorBinding` — e.g. prose explaining *why* the annotation was moved out — is reported as a violation. Code-about-code precedence needs deciding before the first run, not after; this repo has been bitten by it three times in one session (#2450).

Verified clean: the gate's scope is derived by `find` over the module rather than a hand-kept file list, and it exits 1 (rather than passing vacuously) if the module directory is absent — both correct. Detekt baselines were moved alongside the classes. `PolicyDecisionPoint`/`AuthzQuery` correctly stay in `libs-domain`, so the dependency runs runtime -> domain, the allowed direction. Neither this PR nor #3808 adds an integration test proving `@Authorize` still fires after the move; a `@QuarkusTest` driving real HTTP at a decorated endpoint with a denying PDP and asserting 403 is the only layer that can tell a registered interceptor from an unregistered one.

Needs human review regardless.

JiRaska:
Refreshed against `main` (was 290 behind). Two conflicts, and the second one would have silently reverted a merged fix — recording it because the shape is easy to miss.

**`FlagExposure.kt`.** This PR moves `LoggingExposurePublisher` out of `libs-domain` into `libs-runtime` (a JBoss `Logger` in the domain half is exactly what the new gate exists to stop). Meanwhile `main` **edited** the class where it still lived, adding the exposure timestamp to the log line:

```diff
-"flag exposure exposureId=%s flag=%s variant=%s key=%s reason=%s traceId=%s",
+"flag exposure exposureId=%s at=%s flag=%s variant=%s key=%s reason=%s traceId=%s",
+exposure.timestamp,
```

So the branch's copy in `libs-runtime` is the **pre-fix** body. Taking this side wholesale — the obvious resolution for a move, and what git leaves you inclined to do — deletes `main`'s change with no conflict to warn you, since the two live in different files by then. Resolved by taking the move and porting `main`'s newer body into the moved copy; the moved class now `diff`s clean against `main`'s version.

That fix is not cosmetic next door: the field it logs defaulted to `Instant.EPOCH`, and the log line is the only place it surfaces.

**`.github/gates/gates.yaml`.** Both sides append a gate entry at the same point. Union, then counted rather than eyeballed, because a registry losing an entry here silently deregisters a gate:

```
main=118  branch=106  merged=119  union=119
lost from main: none      lost from branch: none      duplicate ids: none
```

**Verified after resolution:** `check-libs-domain-purity.sh` exits 0 — *102 domain source file(s) checked, all framework-free*; the class is gone from `FlagExposure.kt` (`grep -c` = 0) and nothing still imports it from the domain package.

Not verified from here: that `libs-runtime` compiles. This is a package-preserving move, so consumers should not see it, but the fleet build is the arbiter and it is running now.

JiRaska:
Closing as superseded by #3808, which fixes the same defect (#3670) and is the further-along of the
two.

This is a duplicate, not a difference of approach — compared by content, not by title. Both move
`Authorize.kt` and `FeatureFlag.kt` from `openbank-libs-domain` to `openbank-libs-runtime`, both
touch `FlagExposure.kt`, `ServiceTokenProvider.kt` and `OutboxDispatch.kt`, both edit
`.github/gates/gates.yaml`, and — the part that matters — **both add their own domain-purity
checker**: `check-libs-domain-purity.sh` here, `check-domain-purity.py`/`.sh` there. Merging both
would put two copies of one rule in `Validate manifests`, on every PR, free to drift apart. CI
cannot see that; only comparing the two file lists can.

#3808 survives because it is newer, registers the rule in `rules.yaml`, and carries a
`domain-purity-baseline.txt` — the ratchet this PR does not have. This branch is also 289 commits
behind with conflicts, so keeping it would cost a large rebase for the version that loses anyway.

**There is a third effort on #3670 and it should not be lost with this one.** The local branch
`work/3670-finish` (no PR) carries a commit worth grafting onto #3808:

    fix(libs): report and floor the domain-purity scan size so a collapsed scope cannot read green

That is a guard against the failure mode this repo keeps hitting — a gate that passes because its
scope silently became empty. Neither #3677 nor #3808 has it. Recommend folding it into #3808 before
#3670 is closed; leaving it on an unpublished branch is how it disappears.

Issue #3670 stays open and is the thing to track.

Issue #3672, opened 2026-08-03
Consolidate duplicated application.yaml infrastructure blocks (datasource/oidc/kafka/outbox/log/otel) into a shared base


JiRaska:
38 services copy-paste ~200 lines of identical infra config; only the DB name and a few values differ. Every infra change is a 38-file sweep and drift is guaranteed.

Scope: shared base config (classpath or gitops-layer) + per-service overrides only; CI check for duplicate YAML keys stays; verify SmallRye profile/import semantics on one pilot service before sweeping.

JiRaska:
Re-measured against live `origin/main` (2026-08-06) — **still open, and the debt has grown**.

| measure | issue text (2026-08-03) | measured on `origin/main` today |
|---|---|---|
| services with `src/main/resources/application.yaml` | 38 | **56** |
| total lines across those files | ~7600 implied | **9504** |
| declare `quarkus:` | — | 56/56 |
| declare an `oidc:` block | — | 56/56 |
| declare a `datasource:` block | — | 50/56 |
| mention `log:` | — | 51/56 |
| mention `otel` | — | 45/56 |
| mention `kafka` | — | 40/56 |
| mention `outbox` | — | 32/56 |

No shared-base mechanism has landed: no `application.yaml` on main references a
`smallrye.config.source`/config-import seam, so every one of the 56 is still a full standalone copy.
An infra change is now a **56-file** sweep, not 38.

Method: `git ls-tree -r origin/main` for the path glob, `git show origin/main:<f>` per file (the local
checkout is not authoritative). Probe validated against known positives.

No action taken — this is a re-measurement only.

JiRaska:
## Still true, but no longer the drift risk it was filed as — closing

Re-measured on `origin/main` @ `1a64af98`:

- The duplication is real: `openbank-account-service/src/main/resources/application.yaml` is still 264 lines and every service carries its own datasource/oidc/kafka/outbox/log/otel block.
- **The shared-base mechanism this issue asks for already exists and works**: `openbank-libs-runtime/src/main/resources/application.yaml` (31 lines) is a classpath base every service inherits today — security headers, the management port, and a `"%test"` oidc override. So the SmallRye semantics question ("verify on one pilot service before sweeping") is already answered in the affirmative; anyone doing the sweep just moves more keys into that file.
[...]

PR #3686, opened 2026-08-03
refactor(finrep): pilot a shared base application.yaml from openbank-libs-runtime

JiRaska:
## Summary

Pilot for consolidating duplicated `application.yaml` infrastructure blocks: a base `application.yaml` ships as a resource inside `openbank-libs-runtime` (every service already depends on it), Quarkus merges it at a lower ordinal than the service's own file, and the pilot service (finrep) slims to its service-specific keys. **Verified empirically** — the merge semantics are asserted by a booted `@QuarkusTest`, not assumed from docs.

## Type of change

- [ ] fix (bug fix)
- [x] refactor (no behavior change)
- [ ] perf (performance)
- [ ] docs
- [ ] chore (build / tooling)
- [ ] breaking change

## Linked issues

Refs #3672

## Changes

- `openbank-libs-runtime/src/main/resources/application.yaml` (new): the two blocks verified byte-identical across the sampled fleet — the OWASP security-header set and the management interface (`enabled`, port 8085, `/q` root) — plus the `%test` oidc disable services already repeat individually.
- `openbank-finrep-service/application.yaml`: those blocks removed (they now come from the base); everything service-specific stays.
- `FinrepBootSmokeTest`: new test `dependency application yaml contributes defaults and service yaml still wins conflicts` — asserts a base-supplied value (Strict-Transport-Security), a service `%test` override beating the base (X-Frame-Options SAMEORIGIN over DENY), base management defaults, and an untouched service-local key. 3/3 green.

## Decision note (from the pilot)

- **In the base:** security headers, management interface, `%test` oidc-off. Only what was identical everywhere.
- **Stays per-service:** datasource (per-service DB), oidc client config (varies), kafka channels/topics, otel endpoints, outbox settings (`dispatch-enabled` stays per-service on purpose — flipping the default would change semantics for services without an outbox), ports (must remain visible to gitops-driven NetworkPolicy generation).
- **Fleet sweep plan:** after this lands, adopt per service in small batches (one PR per service, `refactor(<scope>)`), each removing only blocks identical to the base.
- **Risks called out:** (1) services that never declared the security-header/management blocks will now *gain* them — a deliberate baseline upgrade, but it is a behavior change for those services; (2) the base `%test` oidc-off could interact with a service whose tests assume oidc-on without declaring it — CI's full-fleet build (libs change → fleet rebuild) is the safety net; (3) any future edit to the base ripples the whole fleet, so it must stay minimal and universal.

## Definition of Done

- [x] Unit tests added/updated (boot smoke test proves the merge semantics).
- [ ] Integration tests added/updated — N/A (no service boundary change).
- [ ] Contract tests updated — N/A.
- [x] `./gradlew :openbank-libs-runtime:build :openbank-finrep-service:build` green (includes detekt/ktlint/kover).
- [x] No new lint warnings.
- [ ] OpenAPI spec regenerated — N/A.
- [ ] Flyway migration added — N/A.
- [ ] Avro schema versioned — N/A.
- [x] Docs updated — decision note is this PR body; the sweep guide follows with the first batch PR.

## Security checklist

- [x] No new suppressions.
- [ ] Auth / crypto / payment code changed? — No; `%test`-scoped oidc disable is centralized, production auth untouched.
- [ ] PII path changed? — N/A.
- [ ] Cardholder data path changed? — N/A.

## Compliance impact

- [ ] PCI DSS
- [ ] DORA
- [ ] GDPR
- [ ] PSD2 / SCA / RTS
- [ ] AML / 5AMLD
- [ ] CNB reporting
- [x] None — security *headers* baseline strengthens, no compliance surface change.

### DIFF
--- a/openbank-finrep-service/src/main/resources/application.yaml
+++ b/openbank-finrep-service/src/main/resources/application.yaml
@@ -8,21 +8,6 @@ quarkus:
       origins: "http://localhost:3000,http://openbank-admin-ui:3000"
       methods: GET,POST,PUT,DELETE,OPTIONS
       headers: Content-Type,Authorization
-    header:
-      X-Content-Type-Options:
-        value: nosniff
-      X-Frame-Options:
-        value: DENY
-      X-XSS-Protection:
-        value: "1; mode=block"
-      Referrer-Policy:
-        value: strict-origin-when-cross-origin
-      Permissions-Policy:
-        value: "geolocation=(), microphone=(), camera=()"
-      Content-Security-Policy:
-        value: "default-src 'self'"
-      Strict-Transport-Security:
-        value: "max-age=31536000; includeSubDomains"
 
   oidc:
     auth-server-url: ${QUARKUS_OIDC_AUTH_SERVER_URL:http://localhost:8080/realms/openbank}
@@ -43,11 +28,6 @@ quarkus:
   swagger-ui:
     always-include: true
     path: /api/docs
-  management:
-    enabled: true
-    port: 8085
-    host: 0.0.0.0
-    root-path: /q
 
 openbank:
   service:
@@ -63,8 +43,11 @@ openbank:
 
 "%test":
   quarkus:
-    oidc:
-      enabled: false
+    http:
+      test-port: 0
+      header:
+        X-Frame-Options:
+          value: SAMEORIGIN
     management:
       enabled: false
     devservices:
--- a/openbank-finrep-service/src/test/kotlin/com/openbank/finrep/FinrepBootSmokeTest.kt
+++ b/openbank-finrep-service/src/test/kotlin/com/openbank/finrep/FinrepBootSmokeTest.kt
@@ -9,6 +9,7 @@ import io.restassured.module.kotlin.extensions.Given
 import io.restassured.module.kotlin.extensions.Then
 import io.restassured.module.kotlin.extensions.When
 import org.assertj.core.api.Assertions.assertThat
+import org.eclipse.microprofile.config.ConfigProvider
 import org.junit.jupiter.api.Test
 
 /**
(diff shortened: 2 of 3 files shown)


JiRaska:
**Verdict: real, currently-observed regression — the PR's own stated risk #2 has materialized in its own CI run.** 1 blocking finding.

- The PR body's own "Risks called out" section names this exact failure mode: *"the base `%test` oidc-off could interact with a service whose tests assume oidc-on without declaring it — CI's full-fleet build (libs change → fleet rebuild) is the safety net."* The safety net caught it: on this PR's own CI run (`30847077380`), **`build (openbank-campaign-service)` and `build (openbank-mcp-service)` both FAIL** with the identical signature —
  `jakarta.enterprise.inject.UnsatisfiedResolutionException: Unsatisfied dependency for type org.eclipse.microprofile.jwt.JsonWebToken and qualifiers [@Default]`
  in campaign-service's `CampaignRestContractIT` and two suppressed instances in mcp-service's test build. This is the shape of `%quarkus.oidc.enabled=false` (now contributed by `openbank-libs-runtime/src/main/resources/application.yaml`'s new `"%test": quarkus.oidc.enabled: false` block) disabling OIDC for `%test` profile builds in services that never declared that override themselves and whose tests inject `JsonWebToken` expecting OIDC/JWT to be active. Since every service depends on `openbank-libs-runtime`, this block is now fleet-wide, not finrep-scoped, contradicting the PR's framing as a finrep-only pilot with no other-service blast radius. `all-green` and `Validate manifests` are red on this PR because of it — this isn't advisory, it's a real break blocking merge as things stand.
- The fix is narrower than reverting the whole pilot: either drop the `%test` oidc-disable block from the shared base (keep it only in finrep's own `application.yaml` until each service opts in deliberately, per the PR's own stated adoption plan of "one PR per service"), or make campaign-service/mcp-service declare their own `%test` oidc-enabled override before this merges.

Verified clean (on the parts not already broken): the merge-ordinal claim itself is demonstrated rather than assumed — `FinrepBootSmokeTest`'s new `@QuarkusTest` boots real config and asserts a base-supplied value, a service-`%test`-override beating the base, and an untouched service-local key, which is the right way to prove SmallRye config-source precedence rather than trusting docs; the finrep-service build itself is green.

Issue #3679, opened 2026-08-03
interest-service and delegation-service default AUTHZ_ENFORCE=false in application.yaml (deployed value is true — only the manifest enforces it)


JiRaska:
Found while reconciling ADR-0034 with gitops (PR #3678): three services in `rules.yaml: money_path_services` still run OPA authorization in advisory mode:

- **openbank-sdd-service** — posts real debits via TransactionServiceClient for every authorised direct-debit collection (#1478)
- **openbank-interest-service** — posts real GL journals on capitalize()/write-off (#1478)
- **openbank-delegation-service** — four-eyes assessment pending (#938 follow-up)

ADR-0034's 2026-07-17 note claimed all 16 money-path services enforce=true; that was inaccurate and is corrected in PR #3678.

Scope: validate each service's OPA bundle coverage observed in advisory mode (unexpected DENYs would have logged), then flip `AUTHZ_ENFORCE=true` per service, one PR each. Money-path rules apply: 2 approvals + threat-model consideration where the flip changes authorization behavior on a live endpoint.

JiRaska:
## Per-service precondition state (measured 2026-08-06)

Method, so the claims below are falsifiable: pods read live with `kubectl` (read-only); policy decisions evaluated with `opa eval` against the **live** `*-opa-bundle` ConfigMaps pulled out of the cluster, not against the repo copies — with a must-DENY control (`zzz.frobnicate`) and a must-ALLOW control in every run, both of which behaved correctly.

| | sdd-service | interest-service | delegation-service |
|---|---|---|---|
| in `money_path_services` | yes | yes | yes |
| `AUTHZ_ENFORCE` | `false` | `false` | `false` |
| `opa` sidecar declared | **no → yes in #3807** | yes | yes |
| `opa` sidecar live + ready | **no** | yes (`true,true`) | yes (`true,true`) |
| bundle ConfigMap | **none → added in #3807** | `interest-opa-bundle` | `delegation-opa-bundle` |
| service `_rest_ext.rego` | **none → added in #3807** | inline in the generator | `delegation_rest_ext.rego` |
| actions with no allow rule | **5 of 7** → 0 after #3807 | 0 | 0 |
| empty `resource` a real defect? | no | no (see below) | no |
| unreachable `SERVICE`-typed rule | none | none | none |
| **flip safe today?** | **no** | **not yet** | **not yet** |

### sdd-service — the hard blocker, and the one thing this session could actually fix

[...]

Issue #3422, opened 2026-08-02
Converge the four money-path OPA bundles that drifted while apply was blocked (ledger, psd2, balances, consent)


JiRaska:
Four money-path services are serving OPA policy that does not match `main`, and #3415 prevents this recurring but does not converge what has already drifted.

## Measured 2026-08-02 (sandbox)

| app | live checksum | git checksum |
|---|---|---|
| ledger | `ea6e7952f166e3bc` | `8f455914ec6024b2` |
| psd2 | `30a4784ea24dabe8` | `bdfcb407bfb50106` |
| balances | *(absent)* | *(absent)* |
| consent | `3267ec2b5ff04bea` | `87c6711fc8d91ecf` |

Cause and prevention are in #3415: their bundles exceeded the 262144-byte `metadata.annotations` ceiling, so client-side apply was rejected with `Too long` and Argo retried forever while the cluster kept serving the previous policy.

## Why this needs a separate, deliberate step

`ServerSideApply=true` is now set on these Applications, but a resource already applied client-side **conflicts on its first server-side apply**:

```
error: Apply failed with 2 conflicts: conflicts with "argocd-controller" using v1:
- .data.<svc>_rest_ext.rego
- .metadata.annotations.openbank.tech/policy-checksum
```

Clearing that needs one `kubectl apply --server-side --force-conflicts` per bundle. Mechanically trivial — it was done for lending in #3402 and verified — but for these four it **converges live authorization onto policy that nobody has re-reviewed since it was blocked**. Nobody knows how long each has been frozen or what changed in git meanwhile, so the diff has to be read before it is applied, not after.

## What to do, per service

1. `git show origin/main:openbank-infra/gitops/components/<svc>/<svc>-opa-bundle.yaml` vs the live ConfigMap — read the rego diff and confirm every rule in it was intended.
2. Establish when the freeze started (first commit whose bundle crossed the ceiling) and what authorization decisions were made under the stale policy in between.
3. `kubectl apply -f <bundle> -n <ns> --server-side --force-conflicts --field-manager=argocd-controller`.
4. Restart the pods: the OPA sidecar loads its policy at startup and the pod-template checksum annotation will already match, so **the roll does not happen by itself** — this is what made lending look fixed when it was not.
5. Verify against the running sidecar, not the ConfigMap: query `/v1/data/openbank/rest/allow` on the pod, with a known-negative as well as the rule you expect to pass.

## Also worth answering

[...]

Issue #889, opened 2026-07-12
standing-order-service: no consumer for standing-order.due.v1 — standing orders never actually execute


JiRaska:
## Problem

`standing-order-service`'s daily execution flow never actually triggers a payment. It looks
complete end-to-end (CRUD, scheduling, event publish, callback endpoints all exist and have unit
test coverage), but nothing downstream is wired to the event it emits.

Found while investigating the "standing-order -> payment services" edge for the issue #468 pact
coverage sweep (that edge turned out to be a phantom — no real integration exists to
contract-test — see the discussion on #468).

Concretely, today:

1. `StandingOrderExecutionScheduler` (cron, daily 03:00 UTC) finds due standing orders and calls
   `StandingOrderService.executeOrders()`.
2. `executeOrders()` writes a transactional-outbox row, published via
   `KafkaStandingOrderOutboxEventPublisher` (`@Channel("standing-order-events-out")`) to topic
   `openbank.standing-orders.order.event`, event type `standing-order.due.v1` (or
   `standing-order.failed.v1` on failure). Payload carries `orderId`, `paymentType`,
   `debitAccountId`, `creditorIban`, `creditorName`, `creditorBic`, `amountMinorUnits`,
   `currency`, `remittanceInfo`, `idempotencyKey`, `executionDate`.
3. **Nothing consumes this topic.** Grepped `@Incoming` across `openbank-domestic-payment`,
   `openbank-sepa-payment`, `openbank-sepa-instant`, and `openbank-transaction-service` — zero
   matches for this event type or topic.
4. `StandingOrderResource` also exposes `PATCH /{id}/record-execution` and
   `PATCH /{id}/record-failure` — callback endpoints clearly designed for a payment rail to report
   back the outcome. Grepped every payment/transaction service for any `@RegisterRestClient`
   targeting standing-order-service, or any reference to `record-execution`/`record-failure` —
   zero matches anywhere in the repo.

`StandingOrderExecutionScheduler`'s own doc comment claims: "Downstream payment rails
(sepa-payment, domestic-payment) consume standing-order.due.v1 events" — this is **not true** in
the current codebase. The `governance.yaml` lineage manifests on both sides list a
`transaction-service` `relationType: api / description: creates` block, but it's the same
boilerplate pasted verbatim into domestic-payment/sepa-payment/sepa-instant's manifests (who do
have real REST clients to transaction-service) — standing-order-service's copy isn't backed by
any actual client code.

[...]

PR #3688, opened 2026-08-03
fix(governance): classify standing-order and psd2 as money-path services

### LINKED ISSUE #2188: openbank-standing-order-service initiates real SEPA transfers but is absent from rules.yaml money_path_services
## Finding

`openbank-standing-order-service` is **not** listed in `openbank-libs/governance/rules.yaml: money_path_services` and has no `docs/threat-models/` entry — yet it initiates real SEPA transfers via `StandingOrderDueConsumer` (#889).

Consequence: the 2-approval + threat-model requirement did not bind PR #2180, even though the issue it fixed (#2148) was labelled `money-path` and the defect was that no due standing order ever executed. The labels and the enforced list disagree.

## Why it is not a one-line fix

Adding a service to `money_path_services` pulls in:
- a threat model at `docs/threat-models/openbank-standing-order-service.md` (ADR-0030), and
- SLO objects — `check-slo-registry.sh` is a **required** check, so the PR that adds the service goes red until those exist.

So it needs its own PR with all three landing together, not a list edit.

## Worth checking at the same time

Whether any *other* service that moves money is missing from the list. The same reasoning that missed standing-order — it publishes an event rather than calling a payment API directly — would miss its siblings.

Note that the approval half of the rule is currently not enforced at all (#2183), so adding the service to the list buys the threat model and SLO requirements but not the two approvals.

Refs #2148, PR #2180, #2183.

JiRaska:
## What

Adds `openbank-standing-order-service` and `openbank-psd2-service` to `rules.yaml: money_path_services`.

**standing-order** (the issue's finding): `StandingOrderDueConsumer` POSTs a real SCT to sepa-payment `/api/v1/sepa-payments` for every due order (#889) — it initiates money movement like sdd, not adjacent to it, but was absent from the list, so the threat-model / SLO / 2-approval requirements did not bind it (they did not bind PR #2180).

**psd2** (the issue's "worth checking at the same time" sibling sweep): `PaymentInitiationService` (PIS) calls `transactionClient.initiatePayment` for SCT / SCT_INST / domestic / SIPO — the same initiates-real-payments class as sdd. The OPA-bundle converge work (#3422) already treats psd2 as money-path; this makes the enforced list agree. The rest of the sweep (aml, card-issuance, dispute, mcp, ap2, tax-reporting, campaign, customer-edge, analytics-sink, kyc) found no other service that initiates fund movement — mcp/customer-edge only proxy authenticated callers.

## What lands together (a list edit alone would go red)

- `rules.yaml` — both entries with four-eyes assessments + the `AUTHZ_ENFORCE=false` (advisory) caveat. **Both join the #3679 cohort** — this change buys the threat model, the SLOs and the (currently unenforced, #2183) 2-approval rule, not runtime authz enforcement.
- `docs/threat-models/openbank-standing-order-service.md` — new lightweight STRIDE model (ADR-0030 D2). psd2 already ships one.
- `pyrra-slo-money-path.yaml` — governed availability + latency SLO pair per service (`check-slo-registry.py` is required).
- `rules-opa-data.yaml` regenerated + all 36 OPA bundles restamped — `money_path_services` is policy-read data (`data.rules.*`), so every service's bundle + policy-checksum annotation moves. Pods will roll on sync; that is the mechanism working as designed.

## Verified locally

- `check-slo-registry.py` — OK, 23 services / 46 objects
- `check-threat-models.py` — OK, all 23
- `check-podmonitor-namespace-coverage.py` — OK (payments + psd2 namespaces scraped)
- `check-accounting-clock.py --enforce` — OK for both new entries
- `check-finops-tiers.py` — no findings
- yamllint (CI config) — clean on the pyrra manifest and the whole components tree
- bundle regeneration idempotent (second run: no diff); ledger bundle embeds both new entries

The `public-readiness-audit.sh` gitleaks failure is a pre-existing full-history finding, identical without this change.

Closes #2188
Refs #2183, #3679, #3422

### DIFF
--- a/docs/threat-models/openbank-standing-order-service.md
+++ b/docs/threat-models/openbank-standing-order-service.md
@@ -1,0 +1,83 @@
+<!--
+SPDX-License-Identifier: Apache-2.0
+Copyright (c) OpenBank contributors. Licensed under the Apache License, Version 2.0.
+-->
+# Threat model — openbank-standing-order-service
+
+- **Date:** 2026-08-03
+- **Status:** Lightweight STRIDE (ADR-0030 D2). **Money-path-adjacent** (recurring payment instruction, execution initiation) — listed in `rules.yaml: money_path_services` (issue #2188).
+- **Purpose:** Standing orders — recurring payment instruction lifecycle (create/amend/pause/resume/cancel) and the daily due-date sweep that initiates the actual transfers (ADR-0114).
+
+## 1. Scope & purpose
+
+The standing-order service stores a customer's recurring payment instruction and, once a day,
+sweeps for due orders (`StandingOrderExecutionScheduler`), writes a transactional-outbox row per
+due order, and dispatches `standing-order.due.v1` to Kafka. `StandingOrderDueConsumer` (#889)
+picks each due event up and, for a `SEPA_CREDIT` order, initiates the real credit transfer via
+`openbank-sepa-payment`'s `POST /api/v1/sepa-payments` — a gated money-path service that owns the
+irreversible debit (sanctions/AML gate, Temporal workflow, ledger booking). `DOMESTIC`/`INTERNAL`
+rails are not wired yet and are recorded as failures, never silently dropped.
+
+The service is **money-path-adjacent**: it authorises and schedules the instruction and triggers
+the transfer, but the fund movement itself lives downstream — the same class as
+openbank-sdd-service.
+
+## 2. Data flow (DFD)
+
+```
+[Customer Edge / Admin-UI] --HTTPS--> [openbank-standing-order-service]
+                                           |
+                     daily sweep (Quarkus scheduler)
+                                           |---> [outbox -> Kafka: standing-order.due.v1 / standing-order.failed.v1]
+                                           |
+                     SEPA execution   -----|---> [openbank-sepa-payment] (SCT, money-path)
+                     outcome           --- |---> [confirmExecution / recordFailure -> order state + outbox]
+```
+
+- **External entities:** customer-edge (authenticated customer), admin-UI (ROLE_OPERATOR),
+  sepa-payment (M2M callee).
+- **Trust boundaries:** edge → service (OIDC + OPA sidecar, currently AUTHZ_ENFORCE=false
+  advisory); service → Kafka (mTLS via Strimzi); service → sepa-payment (OIDC client credentials,
+  `openbank-services` client, ROLE_OPERATOR accepted by sepa-payment's createPayment).
+- **Assets:** standing-order instructions (debtor/creditor IBAN, BIC, amount, currency, schedule,
+  remittance info), execution history, idempotency keys.
+
+## 3. Authn/Authz
+
+- All REST endpoints: `@RolesAllowed` (ROLE_CUSTOMER for own orders, ROLE_OPERATOR/ROLE_ADMIN for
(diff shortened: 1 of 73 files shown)


JiRaska:
Heads-up on an overlap: #3795 (issue #3333) edits every object in `pyrra-slo-money-path.yaml`, adding `spec.alerting.absent: false` — this deletes the `SLOMetricAbsent` rule family, which is critical-severity and fires on idle services rather than on faults. `check-slo-registry.py` in that PR now **requires** the flag on every SLO object.

So whichever of these two lands second needs one edit: the four new `standing-order` / `psd2` objects here must carry

```yaml
  alerting:
    absent: false
```

or the `slo-registry-consistency` gate goes red with the reason spelled out. Flagging it now because a text merge of the two will be clean and still wrong — the new objects would simply arrive without the block.

JiRaska:
**Review: no blocking findings, 1 note.** Automated agent review — not an approval.

What it does: adds `openbank-standing-order-service` and `openbank-psd2-service` to `money_path_services`, restamping every service's OPA bundle (the key is read as `data.rules.money_path_services`) and adding a threat model for standing-order.

Verified — the two things this class of PR gets wrong:

- **Bundle regeneration is correct and complete.** I re-ran `.github/scripts/gen-rules-opa-data.py` and then the `gen-*opa-bundle*.sh` scripts for ledger, party and accounts against this branch: **zero diff**. So the derived `rules-opa-data.yaml` was regenerated after the final `rules.yaml` edit (not before it, the #2457 trap), and the ~35 bundle ConfigMaps plus their `policy-checksum` pod-roll annotations agree with the data they embed. A PR in this shape that edits governance data and ships stale bundles deploys the *old* policy; this one does not.
- **Rule 7 (money-path services need a threat model, ADR-0030) is satisfied for both.** `docs/threat-models/openbank-standing-order-service.md` is added here; `docs/threat-models/openbank-psd2-service.md` already exists on `main`. Easy to miss that the PR only adds one file for two services.
- **The classification is defensible on the evidence cited.** `StandingOrderDueConsumer` POSTs a real SCT to sepa-payment for every due order, and psd2's `PaymentInitiationService` calls `transactionClient.initiatePayment` — both initiate money movement rather than sitting adjacent to it, which is the same test `openbank-sdd-service` passes. The four-eyes assessments are written out rather than assumed, and both conclude "no operator force-execute endpoint exists to maker-checker gate; add a verb here if one is ever added" — that is the right shape, since a `four_eyes.verbs` entry naming nothing is worse than an explicit absence.

Note (not blocking):

1. **`rules.yaml`'s standing-order comment records "Runs `AUTHZ_ENFORCE=false` (advisory) — joins the #3679 cohort."** Classifying a service money-path raises the review bar on changes to it but does not turn its policy enforcement on; a money-path service whose PDP decisions are advisory is a gap the classification now makes more visible rather than closes. Worth confirming #3679 tracks standing-order and psd2 explicitly, so the classification does not read as more assurance than it delivers.

Minor cosmetic: the `openbank-psd2-service` comment in `rules-opa-data.yaml` terminates mid-sentence at "PaymentInitiationService (PIS) calls" where `rules.yaml` continues. It is a comment in a generated file and the generator round-trips clean, so it changes nothing — flagging only so nobody "fixes" it by hand and breaks the derivation.

`mergeStateStatus: UNSTABLE`. Adding a service to `money_path_services` is a rule-7 change requiring 2 approvals; this needs human review regardless.

Issue #3656, opened 2026-08-03
Implement ADR-0219 ContactPolicyGate — slice 1: gate in libs-runtime + campaign-service adoption


JiRaska:
ADR-0219 (contact classes, durable counters, suppression) is the subordination point for the whole engagement/gamification branch (ADR-0220 D5) and the callable form of ADR-0203 D5's "gated by ADR-0200's suppression rules" citation. Today those rules exist only inline in campaign-service.

**Slice 1 (this issue):**
- `ContactPolicyGate` in `openbank-libs-runtime`: D1's three contact classes (OUTBOUND_SEND / PROMOTIONAL_IMPRESSION / SERVICE_EXEMPT), rolling windows (never calendar), deterministic arbitration order, D3's suppression model + evaluation ordering (suppression → caps/quiet → consent), D5's fail-closed semantics for classes 1–2 — all as domain-tested library code.
- campaign-service adopts the gate at the ADR-0200 D2 per-step call site, replacing its inline cap/quiet-hours/consent block (D4's first mandatory call site).

**Explicitly NOT in slice 1** (tracked as follow-ups, no placeholders shipped):
- D2's Valkey counters + rebuild-on-flush — campaign's send cap already counts from the durable Postgres send log, so the cache-flush burst defect class never applied there; the shared hot-path counter store + rebuild machinery is its own slice.
- D3's suppression STORE (platform list + admin APIs, consent-service-adjacent) — slice 1 ships the model + port; campaign wires the truthfully-empty list until the store exists.
- D4's CI wiring assertion (detekt/AST rule) — needs the call-site inventory from at least two adopters.
- notification-service consent-gate migration to the gate (ADR-0198 D4 choke point).

Refs ADR-0219 (D1–D5), ADR-0220 D5, ADR-0203 D5, #3585.

JiRaska:
Closing — slice 1 has landed, both halves. Verified in substance rather than by file existence, since
"the class exists" and "the class does what the slice specified" are different answers.

**Half 1 — `ContactPolicyGate` in `openbank-libs-runtime`:**

```
openbank-libs-runtime/src/main/kotlin/com/openbank/libs/contact/
  ContactPolicyGate.kt   ContactPolicy.kt   ContactArbitration.kt
```

Checked against each requirement this issue named:

- **D1's three contact classes** — `OUTBOUND_SEND`, `PROMOTIONAL_IMPRESSION`, `SERVICE_EXEMPT`
[...]

PR #3689, opened 2026-08-03
feat(consent): ADR-0219 D3 suppression store — the granular do-not-contact

JiRaska:
## Summary

Implements **ADR-0219 D3 (slice 2 of #3656)**: the suppression list the contact-policy gate evaluates *before* consent now has a real home — consent-service, as the consent-adjacent state it is. A suppression is a stop that is **not** a consent revocation, so a customer can say "not about loans" without switching everything off (and without being re-asked later, because the consent survives). Independent of #3665 (slice 1) — different service, no shared files; the gate's `SuppressionPort` wiring to this API is the small follow-up once #3665 lands.

## Type of change

## Linked issues

Refs #3656 (slice 2 of the ADR-0219 program)

## Changes

- **Domain:** `Suppression` aggregate — ALL/SCOPE/TOPIC validated by construction (an ALL entry with a value would read as scoped while covering everything), ADR-0219 reason codes (customer opt-out, complaint, RM-managed, legal hold, deceased), one-way `revoke`.
- **Persistence:** **V6 migration** `suppressions` (check constraint enforces the value shape; partial index on the gate's read shape — active-by-party; rollback: `DROP TABLE suppressions;`).
- **Events:** `SuppressionCreated`/`SuppressionRevoked` via the transactional outbox in the SAME commit as the state change — the gate's near-real-time invalidation signal (ADR-0219's event-speed rule) can never be lost between commit and send.
- **REST:** `POST /api/v1/suppressions`, `GET /api/v1/suppressions/party/{partyId}` (the gate's read), `POST /api/v1/suppressions/{id}/revoke`; caller identity via SecurityIdentity (consent-service convention); **openapi 1.4.0 → 1.5.0** (additive MINOR, major invariant intact).
- **OPA:** `suppression.manage` is HUMAN-operator-only (shared M2M client excluded — a role-only check would grant every backend service the write); `suppression.read` serves operators AND the gate's M2M callers; 4 new rego tests; consent OPA bundle restamped.
- **Tests:** `SuppressionTest` pins the ALL/SCOPE/TOPIC shape, `covers()` semantics (incl. "a null topic is never topic-suppressed") and one-way revoke.
- **version 0.16.6 → 0.17.0.**

## Definition of Done

- [x] Hexagonal layering preserved (domain has zero framework imports; ports in application, adapters in infrastructure).
- [x] Unit tests added (domain); Pact provider verification passes.
- [x] `./gradlew :openbank-consent-service:ktlintCheck :detekt :test :koverVerify` — BUILD SUCCESSFUL.
- [x] OpenAPI updated (1.5.0). Flyway migration + rollback note. Avro — n/a.
- [x] rego: `opa test` for the consent ext suite — all new rules covered.

## Security checklist

- [x] No secrets/PII beyond identifiers (partyId) — suppression carries no content.
- [x] New actions gated by OPA with the shared-M2M exclusion pattern (mirrors `operator-consent-write`'s hard-won fix).
- [ ] Auth/crypto/payment — no. Cardholder data — no.

## Compliance impact

- [x] GDPR — Art. 21(2): the granular direct-marketing objection now has a store; entries are the minimum state required (Art. 5(1)(c)); erasure hooks inherit consent-service's PARTY_ERASED handling.
- [x] CNB — Act 480/2004 §7 honoured through the gate's class taxonomy (this is the stop-list half).


JiRaska:
Review: 4 findings, 2 blocking

**Blocking**

1. `openbank-consent-service/version.txt:1` — hand-bumped `0.16.6` → `0.17.0`. release-please owns this file (CLAUDE.md non-negotiable #2, ADR-0029): every prior change to it on `main` is a `chore(main): release` commit, none is a manual edit. Scenario: this merges as a `feat(consent)`, release-please independently proposes `0.17.0` from the same commit, and its release PR now either conflicts on the file or is computed against a version that already appears on `main` without a tag — the #3721/#3753 failure, where a number is skipped or a released version has no tag. The version-sync gate checks `version.txt` agreement, not "is this number still free", so it will not catch it. Revert to `0.16.6`; the `openapi.yaml` bump is the one that is correctly hand-owned and should stay.

2. No integration test exists for the new store — `SuppressionTest` (pure domain, no Quarkus) is the only test file. `SuppressionRepositoryImpl.save()` is the exact shape that has already broken this service: an application-assigned `@Id` (`Ids.newId()`, no `@GeneratedValue`) plus a state transition (`revoke`) routed back through `save()`. The code is **correct** — `Panache.getSession().flatMap { session -> session.merge(...) }`, not `persist` — but nothing proves it, and a repository-mocking test structurally cannot: consent-service shipped this precise defect in #1521 and every revoke 500'd. The same gap covers the outbox: `save()` claims "aggregate + outbox row in ONE transaction" and no test asserts the outbox row exists after a real request. Concrete failure this leaves undetected: revert `merge` to `persist` in a later refactor and `POST /api/v1/suppressions/{id}/revoke` fails at flush with `duplicate key value violates suppressions_pkey` — which finding 3 below then reports to the operator as **409 Conflict**, i.e. indistinguishable from "already revoked", so the contact stop can never be lifted and no 5xx alert fires. Add a REST-driven real-DB IT modelled on the sibling `ConsentRevocationOutboxIT`: create → revoke over RestAssured, assert the row state and both `SuppressionCreated`/`SuppressionRevoked` outbox rows with a plain JDBC read.

**Non-blocking**

3. `.../infrastructure/rest/SuppressionResource.kt:63-72` and `:80-89` — both writes wrap the whole use case in `runCatching` and map every throwable to a client error (`create` → 400 with the raw `it.message`, `revoke` → 409 for anything that is not `SuppressionNotFoundException`). So a DB outage, a check-constraint violation, or a failed outbox write is reported as caller error, never as 5xx, and is invisible to error-rate alerting; the raw exception message (potentially Postgres detail) is echoed to the client. Catch `IllegalArgumentException`/`IllegalStateException`/`SuppressionNotFoundException` specifically and let everything else become a 500.

4. `V6__suppressions.sql:23` — the partial index `(party_id) WHERE revoked_at IS NULL` is non-unique, and nothing dedupes on `(party_id, scope, value)`. Repeated `POST /api/v1/suppressions` with the same body creates N active duplicates. `covers()` is fail-closed so evaluation stays correct, but the operator un-suppress flow is not: revoking one entry leaves the others active, the UI shows the stop as lifted, and contact silently stays blocked. Either add a partial unique index or make "duplicates are intentional" explicit and have the revoke path handle the set.

Verified clean: `openbank.outbox.dispatch-enabled: true` is present in consent-service's `application.yaml`, so the new events actually dispatch. The V6 check constraint (`ALL ⇒ value IS NULL`, `SCOPE|TOPIC ⇒ value IS NOT NULL`) matches the `Suppression` `init` invariant exactly, and the partial index matches the gate's real query (`partyId = ?1 and revokedAt is null`) — index and read shape agree. V6 is a brand-new migration with a rollback note and edits nothing already applied; entity property names map onto the migration's snake_case columns under Quarkus' default naming strategy. The rego contains **no** `input.principal.type == "SERVICE"` rule — M2M is correctly modelled as `HUMAN` + `service-account-openbank-services`, and the write rules explicitly exclude that id, so nothing here becomes unreachable dead code when `AUTHZ_ENFORCE` flips. The OPA bundle and the pod-roll `openbank.tech/policy-checksum` annotation are restamped consistently in both `consent-opa-bundle.yaml` and `consent-service.yaml` (`a6813f6436c88169` → `10f6be30a4b6f835`), and the rego is service-local so no other service's bundle is implicated. `openapi.yaml` `info.version` 1.4.0 → 1.5.0 is a correct additive-minor bump and 1.5.0 is still free on live `origin/main` (which reads 1.4.0). Events carry a real clock-derived `occurredAt`, not `Instant.EPOCH`. `@Path("/api/v1/suppressions")` binds directly to the class — no top-level function stealing it. The threat model is updated, as a money-path service requires. Nothing already written is wrong: this is a new table with no backfill and no existing rows touched, so both blocking findings are fixable in code alone.

JiRaska:
**Review: 4 findings, 1 blocking**

**BLOCKING — `SuppressionResource` has no `@RolesAllowed`, only `@Authorize`.**
`openbank-consent-service/.../infrastructure/rest/SuppressionResource.kt` — the class carries
`@Path` + `@ApplicationScoped` and each method an `@Authorize(action = "suppression.manage" |
"suppression.read")`, with no coarse JAX-RS role gate anywhere. Its sibling in the same package,
`ConsentResource.kt:43-50`, carries `@RolesAllowed("ROLE_API", "ROLE_OPERATOR", "ROLE_ADMIN")` at
class level with a comment saying in as many words that `@Authorize` alone is not the gate and that
every method there was previously missing the pairing, so an anonymous caller reached the
interceptor at all. This PR reintroduces exactly that shape one file over.
Why it bites: `application.yaml:164` is `authz.enforce: ${AUTHZ_ENFORCE:false}`. The sandbox
manifest sets it `true`, so OPA does deny there — but the *default is off*, which means in local,
in `%test`, and in any environment that does not set the variable, `POST /api/v1/suppressions` and
`POST /api/v1/suppressions/{id}/revoke` have no authorization gate of any kind. Inputs → wrong
outcome: any caller the HTTP layer admits posts `{partyId: <someone else>, scope: ALL, reason:
DECEASED}` and that party is silently suppressed from all contact, with `createdBy` recorded as
`"unknown"`. Add the class-level `@RolesAllowed` and this drops to nothing.

**2. `runCatching { ... }.getOrElse { 400 }` on `create` turns every failure into a client error
and echoes the internal message.**
Same file — the catch is unbounded, so a DB outage, a check-constraint violation, an outbox write
failure and a genuinely malformed request all return `400 {"error": "<exception message>"}`. An
operator gets "your input was wrong" for an incident that was not theirs, nothing 5xx-shaped ever
reaches the error-rate SLO, and the Postgres/Hibernate message text is returned to the caller.
`revoke` has the same shape with everything-that-is-not-`SuppressionNotFoundException` collapsed to
409. Catch the domain exceptions (`IllegalArgumentException` / the validation type) and let the rest
propagate to the existing `ExceptionMappers`.

**3. Nothing exercises the repository or the REST path — the outbox-atomicity and revoke claims are
unproven.**
The only new test is `SuppressionTest`, which is pure domain. `SuppressionRepositoryImpl` is written
correctly — `Panache.getSession().flatMap { session.merge(...) }`, not `persist`, so the
application-assigned-`@Id` INSERT-only trap that made *this service's* consent revoke/reject/activate
500 on every call does not apply here — but that is a property this PR proves by reading, not by
testing. A mocked-repository test cannot see it, and a bare `@QuarkusTest` thread cannot call the
reactive repo at all (`No current Vertx context found`). The claim that the `SuppressionCreated`
outbox row commits in the same transaction as the aggregate is likewise only assertable by driving
`POST /api/v1/suppressions` over real HTTP and reading both rows back over plain JDBC — the shape
`ConsentRevocationOutboxIT` already uses in this service. One IT of that shape would close both.

**4. No duplicate suppression handling.**
`SuppressionService.create` has no uniqueness check and V6 has no unique constraint, so posting the
same `(partyId, scope, value)` twice yields two active rows. An operator who then revokes "the"
suppression in admin-ui sees a success and the party stays suppressed by the twin. Fail-safe
direction, but it will read as a broken revoke, which is the same complaint class this service has
already been through.

Verified clean: `merge`, not `persist` (finding 3). `openbank.outbox.dispatch-enabled: true` is
already set in this service's `application.yaml:147`, so the new events will actually dispatch. The
rego rules gate on `input.principal.type == "HUMAN"` with the shared-M2M exclusion — no
structurally-dead `"SERVICE"` principal type — and the write action is human-only while the read
admits the gate's M2M caller. V6 is a new table with a check constraint matching the domain's
ALL/SCOPE/TOPIC invariant, a partial index matching the gate's read shape, and a rollback note.
openapi 1.4.0 -> 1.5.0 is the right classification for additive paths.

---
Automated agent review — **not an approval**. Please confirm the `@RolesAllowed` point against the
`Authorize.kt` contract before merging; it is the one thing here I would not wave through.

Issue #3521, opened 2026-08-02
sanctions.clear four-eyes approvals are unscoped: an approval for one check clears any other check by the same maker


JiRaska:
## Summary

`SanctionsResource.review` is gated `@Authorize(action = "sanctions.clear", resource = "")`. The
empty resource makes a granted four-eyes approval a **bearer token for any review by that maker**,
not authority over the one decision a checker looked at.

## The mechanism

`AuthorizeInterceptor.satisfies` matches a supplied approval on the triple (action, resourceId,
maker):

```kotlin
private fun PendingApproval?.satisfies(action: String, resourceId: String?, maker: String): Boolean =
    this != null && status == ApprovalStatus.APPROVED &&
        this.action == action && this.resourceId == resourceId && makerId == maker
```

With `resource = ""` the middle comparison is `null == null` for every call. So:

1. Maker asks to clear check **A**; the call is paused at 202 with an approval id.
2. A checker approves it — believing they authorised clearing check A.
3. The maker retries with that `X-Approval-Id` against check **B**. It satisfies, and B is cleared.

Nothing clever is required: the same header, a different `checkId` in the body. `markExecuted`
makes an approval one-time, so this is one extra clearance per approval, not unlimited — but the
approval a checker granted is not the approval that gets used.

`rules.yaml` calls this the highest-risk action in the service ("a wrongly-cleared true positive is
a real sanctions violation") and it is the reason sanctions is on the money path.

## Why now

This was recorded as a deferred residual (C2 in the service's threat model) with the reasoning
"deferred because no UI drives this flow yet". #3465 gave the flow a UI and #3506 gives the checker
a queue, so the premise of the deferral no longer holds.

## Fix

`resource = "#cmd.checkId"` (ADR-0206 dotted-path extraction; `cmd` is the `ReviewCommand`
parameter). No policy change is needed — verified against `rest.rego` before proposing this:
`matrix-allows`, the rule that grants `sanctions.clear`, keys on principal type, roles and action
only, and no deny rule reads `input.resource`.

Two things the fix depends on, both of which need their own coverage:

- **The extraction fails CLOSED.** An unknown field name yields no resource and silently restores
  the unscoped behaviour, so a test must assert the named field actually resolves — not merely that
  the annotation string changed.
- **The resource half of `satisfies` was itself untested.** `AuthorizeInterceptorTest` covered a
  mismatched *maker* but never a mismatched *resource*, i.e. the exact comparison this fix starts
[...]

PR #3695, opened 2026-08-03
feat(interest): enforce OPA authorization (AUTHZ_ENFORCE=true)

JiRaska:
## Summary

Flips `AUTHZ_ENFORCE` to `true` for openbank-interest-service (issue #3679) — OPA authorization moves from advisory (log-only) to enforced. Money-path service: threat model exists (docs/threat-models/openbank-interest-service.md), 2 approvals required.

## Type of change

- [x] feat (behavior change: authorization now enforced)
- [ ] breaking change

## Linked issues

Refs #3679

## Pre-flip coverage verification (per the issue's phased plan)

- All five @Authorize actions (`interest.list/read/create/delete/trigger`) are granted to ROLE_OPERATOR by the `rules.yaml` role_action_matrix.
- M2M reads pass via `operator-read-any` (shared `openbank-services` client carries ROLE_OPERATOR); **no M2M writer exists** — no service calls interest writes, so the shared-m2m write prohibition bites nobody.
- Declared ingress callers (network-policies): admin-ui (human operator token), customer-edge, agent-service (read-only tool), security-scanner.
- Note for reviewers: customer-edge's service account carries ROLE_OPERATOR, so the matrix technically reaches interest writes from the edge — same effective posture as today's @RolesAllowed (no regression); tightening that is a separate rego decision, not this PR. 
- `./gradlew :openbank-interest-service:test` green locally.
- gen-network-policies drift gate verified unaffected (generator output identical; the local gate failure was the uncommitted-change artifact).

- [ ] Confirm the advisory-window decision log showed no unexpected would-DENYs (live check — not verifiable from the repo).
- [ ] Threat model: docs/threat-models/openbank-interest-service.md

## Security checklist

- [x] No secrets / PII.
- [x] Auth code unchanged — this flips enforcement of the existing policy, it does not change policy logic.
- [x] Auth change → tag `security-review-required`.

## Compliance impact

- [ ] PCI DSS
- [x] DORA — ICT access control now enforced, not advisory
- [ ] GDPR
- [ ] PSD2 / SCA / RTS
- [ ] AML / 5AMLD
- [ ] CNB reporting


JiRaska:
### Independent coverage check — supports this PR, with one caveat about what it buys

Working #3679 from the other end, I measured the policy this flip would activate. Method: `opa eval data.openbank.rest.allow` against the **live** `interest-opa-bundle` ConfigMap pulled from the cluster (not the repo copy), across all seven `interest.*` actions × four principal shapes, with a must-DENY control (`zzz.frobnicate`) and a must-ALLOW control in the same run — both behaved.

Also confirmed live: `interest-service-5986ff5845-ng8dx` runs `interest-service,opa` with readiness `true,true`.

**Every action resolves ALLOW for every principal shape I could enumerate.** Nothing legitimate is newly denied by this flip, so on coverage grounds it is safe.

Two things a reviewer should have in view:

1. **The eight `resource = ""` sites are not the #3521 defect.** `Authorize.kt` documents the empty string as the non-scoped form, and these are genuinely collection-level or body-driven endpoints (`listAllAccruals`, `accrueAll`, `createRateConfig`, the withholding-remittance trio). The #3521 shape was a *scoped* check written as unscoped; that is not what these are. I checked before assuming.

2. **What enforcement does not buy here.** `operator-interest-write` has no `service-account-*` exclusion, and `role_action_matrix.ROLE_OPERATOR` separately grants `interest.create/.trigger/.delete`, so `service-account-openbank-services` — the shared backend client — reaches every interest write either way. `operator-interest-write` sits on the `deferred` list in `shared_m2m_write_prohibition` ("no identity-scoped rule; callers unverified"), and `matrix-allows` structurally cannot join that register at all (#3765). So this flip hardens against an unauthenticated or wrong-role caller, not against the shared client. Worth stating in the PR body so the control is not read as stronger than it is.

**What I could not measure**: the caller inventory is derived from reading this repo, so a caller that exists only in the cluster is invisible to it. Advisory mode has been logging "would DENY" this whole time — reading that log for a representative window is the evidence that closes it. I did not read it.

The generator's own note that `accrueAll`'s "external scheduler does not exist yet" still holds as far as I can see.

Full per-service table on #3679.

JiRaska:
## Automated agent review — PDP is real and coverage holds; 2 findings, 1 of them a factual error in the justification

**Verdict: lower risk than it looks, but the comment this PR writes into the manifest is wrong in a way that matters for the next reader.**

### What I verified against `origin/main`

`interest-service.yaml` already ships the `opa` sidecar (`--addr=0.0.0.0:8181`, `/health?bundles=true` readiness, `interest-opa-bundle` ConfigMap, pod-roll checksum `aef81e896164e05d`), and `OpaSidecarPolicyDecisionPoint` defaults to `http://localhost:8181`. So this is not a fail-closed-on-every-endpoint flip.

`interest.*` actions in use across `InterestResource` + the second resource: `interest.list`, `interest.read`, `interest.create`, `interest.trigger`, `interest.delete` — five, as the comment says. Against `rules-opa-data.yaml` on `main`, `authz.role_action_matrix.ROLE_OPERATOR.grant` contains `interest.create`, `interest.delete`, `interest.trigger`, `interest.list`, `interest.read`; `ROLE_ADMIN` inherits `ROLE_OPERATOR`; `ROLE_COMPLIANCE` has `interest.read`. Plus base `operator-read-any` for the read/list verbs. **All five are covered.** There is no `interest-service` ext rego, and none is needed.

### 1. "no M2M writer exists" is the load-bearing claim and it is not verified — but the fallout is the opposite of an outage

`customer-edge` is a live caller of interest-service (`CustomerEdgeResource.kt:311, :337` — `GET /api/v1/interest/accruals/{accountId}/summary` and `GET /api/v1/interest/accounts/{accountId}/effective-rate`). Both are reads, both map to `interest.read`, and `service-account-openbank-edge` carries `ROLE_OPERATOR` and is classified `HUMAN`, so `operator-read-any` and `matrix-allows` both grant them. **No 403 there.**

The thing to be aware of instead: `matrix-allows` in `rest.rego` has **no `service-account-` exclusion**, so `interest.create`, `interest.delete` and `interest.trigger` are reachable by *any* backend holding an `openbank-services` token, not only by staff. That is pre-existing on `main` and this PR does not widen it — but flipping `AUTHZ_ENFORCE=true` is the moment that grant stops being theoretical and becomes the service's actual access-control posture. `interest.trigger` runs an accrual pass; `interest.delete` removes a rate config. If the intent is staff-only, the fix is the `not startswith(input.principal.id, "service-account-")` idiom in an `interest_rest_ext.rego`, as `delegation_rest_ext.rego` and `consent_rest_ext.rego` do — not something to discover after the flip.

### 2. The comment credits `operator-read-any` with something the matrix already does

New comment: *"reads additionally reach the shared M2M client via operator-read-any"*. True, but incomplete and slightly misleading — `matrix-allows` grants the same reads to the same principal, so removing `operator-read-any` (which ADR-0223 D2(b) is explicitly triaging for retirement) would not change this service's behaviour. Stating the matrix as the primary path and `operator-read-any` as the redundant one would survive that retirement; as written, this comment becomes wrong the day D2(b) lands.

### 3. No advisory-log evidence cited

Same note as #3696: the guardrail asks for an empty would-DENY population in the live decision log before the flip, and the justification here is a static read of the matrix. Static coverage cannot see a caller nobody wrote down. If a caller was missed it answers `403`, and no probe catches it — `startupProbe`/`readinessProbe` are `tcpSocket`, `livenessProbe` is `/q/health/live` on the management port; none traverse `@Authorize`.

### Verified clean
- Sidecar present on `main`, loopback-addressed, app's default PDP URL matches — no 503-on-every-endpoint risk.
- All five `@Authorize` actions have a matching grant; I found no endpoint that starts denying.
- No `SERVICE`-principal-type rule anywhere in interest's policy path.
- One-file diff, no `.rego` and no `rules.yaml` edit — no OPA bundle restamp owed, no bundle drift.

---
*Automated agent review — this is **not** an approval and no approval is implied. Authorization changes need human review regardless of what any automated check says.*

JiRaska:
Reviewed for coverage. **Safe to flip — but the reason has changed since the earlier comments here, and the tables in them no longer describe `main`.**

**Measure against `main`'s bundle, not this branch's.** This branch is 306 commits behind, and its `rules-opa-data.yaml` predates #3951, which promoted `shared_m2m_write_prohibition` to a top-level key so OPA actually reads it. Verified: `grep -c '^shared_m2m_write_prohibition'` is **0** on this branch and **1** on `main`, and `main`'s committed `interest-opa-bundle.yaml` carries it. #3698 (the interest write prohibition) also merged today. Any evaluation done against this branch's embedded bundle is answering about a policy the fleet no longer runs.

Re-run against **`main`'s** committed bundle, realistic input shape (`resource: {type, id}` — `OpaSidecarPolicyDecisionPoint.toInput()` emits nothing else, so a probe that supplies `resource.attributes.tenant` reaches `operator-on-own-tenant`, a rule that is dead in production):

| principal (`ROLE_OPERATOR`) | `interest.create` | `interest.trigger` | `interest.read` | `zzz.frobnicate` |
|---|---|---|---|---|
| a human operator | ALLOW | ALLOW | ALLOW | DENY |
| `service-account-openbank-edge` | **DENY** | **DENY** | ALLOW | DENY |
| `service-account-openbank-services` | **DENY** | **DENY** | ALLOW | DENY |

Earlier comments state these writes are ALLOW for both service accounts via `matrix-allows`. That was true when they were written; it is not true now. The `frobnicate` column is the control — without it the DENYs prove nothing.

**Does the DENY break a real caller?** Checked rather than assumed: no in-repo client calls an interest **write** endpoint (the `*Client.kt` hits under interest-service are its own outbound calls to account/ledger/transaction). agent-service's MCP tools are `interest_accrual_summary`, `interest_get_accruals`, `interest_list_accruals` — all reads, and reads stay ALLOW for the shared M2M identity. The five `@Authorize` actions are `create/delete/list/read/trigger`; the two writes that just closed have no M2M caller in this repo.

So the flip is safe on coverage grounds. What it enforces is now *stricter* than when this PR was opened, and deliberately so.

**PDP parity is clean** — the workload carries the `opa` sidecar on `0.0.0.0:8181`, matching the interceptor's loopback default, so there is no 503-on-every-endpoint risk. Confirmed with `check-authz-pdp-parity.py`.

**Residual risk, unchanged and worth stating plainly:** all of this evidence is static. A caller that exists only in the cluster is invisible to it, and no probe would catch a 403 — startup/readiness are `tcpSocket` and liveness is `/q/health/live` on the management port, none of which traverses `@Authorize`. The advisory-window would-DENY counter (`openbank.authz.decisions{outcome=deny,enforced=false}`) is the one piece of evidence that would close that gap, and no comment here cites it.

**Correction to a claim on this PR:** "there is no `interest-service` ext rego" is wrong — `interest_rest_ext.rego` is in the bundle and defines `operator-interest-write`. It happens to be redundant with `matrix-allows`, which is why that comment's conclusion still stood.

Also checked: no `input.principal.type == "SERVICE"` rule anywhere in this policy path — correctly identity-scoped by `principal.id`, since that type can never be emitted.

JiRaska:
Refreshed against `main` (was 306 behind). The `Trivy` red that blocked `--auto` was **stale**, not a finding against this PR.

Its run is from 2026-08-03 and reported one HIGH: `brace-expansion` `CVE-2026-69152`, installed `5.0.8`, fixed in `1.1.18 / 2.1.4 / 3.0.6 / 5.0.9`. `main`'s `openbank-admin-ui/package-lock.json` now resolves `1.1.18` and `5.0.9` — both fixed. A one-line `AUTHZ_ENFORCE` change to a workload YAML could not have introduced an npm CVE in the first place; the branch was simply carrying a four-day-old dependency tree.

Merge was conflict-free, and I regenerated the OPA bundle afterwards anyway — a clean merge of a generated artifact is not the same as a current one (that exact trap cost #3807 a 246-commit-stale bundle earlier today). Here it changed nothing: the bundle and the workload's `policy-checksum` were already in sync at `a670f22621bb6a00`. The PR's own diff is unchanged — one file, `AUTHZ_ENFORCE: "false" -> "true"` plus its comment.

One note on that comment, which is now slightly behind its own subject: it says the five actions "are granted to ROLE_OPERATOR by the rules.yaml role_action_matrix … and no M2M writer exists". Both halves still hold, but since #3698 merged today the two write actions are additionally **vetoed** for service-account principals at the allow head, so `interest.create` / `interest.trigger` now resolve DENY for `service-account-openbank-edge` and `service-account-openbank-services` rather than relying on there being no caller. Strictly stronger than what the comment describes — worth a line if you touch the file again, not worth a push on its own.

The other red, `agent-review`, is the known-dead GitHub Models integration; it is not a required check.

PR #3696, opened 2026-08-03
feat(delegation): enforce OPA authorization (AUTHZ_ENFORCE=true)

JiRaska:
## Summary

Flips `AUTHZ_ENFORCE` to `true` for openbank-delegation-service (issue #3679) — OPA authorization moves from advisory (log-only) to enforced. Money-path service: threat model exists (docs/threat-models/openbank-delegation-service.md), 2 approvals required.

## Type of change

- [x] feat (behavior change: authorization now enforced)
- [ ] breaking change

## Linked issues

Refs #3679

## Pre-flip coverage verification (per the issue's phased plan)

 - All ten @Authorize actions are granted by `delegation_rest_ext.rego` with caller-narrowed rules: operator writes exclude `service-account-*`, the edge principal is scoped to the customer action set only, `delegation.check` is the sole M2M-reachable action.
- Sole declared ingress caller: admin-ui relaying a human operator's bearer token (network-policies).
- `./gradlew :openbank-delegation-service:test` green locally.
- gen-network-policies drift gate verified unaffected (generator output identical; the local gate failure was the uncommitted-change artifact).

- [ ] Confirm the advisory-window decision log showed no unexpected would-DENYs (live check — not verifiable from the repo).
- [ ] Threat model: docs/threat-models/openbank-delegation-service.md

## Security checklist

- [x] No secrets / PII.
- [x] Auth code unchanged — this flips enforcement of the existing policy, it does not change policy logic.
- [x] Auth change → tag `security-review-required`.

## Compliance impact

- [ ] PCI DSS
- [x] DORA — ICT access control now enforced, not advisory
- [ ] GDPR
- [ ] PSD2 / SCA / RTS
- [ ] AML / 5AMLD
- [ ] CNB reporting

### DIFF
--- a/openbank-infra/gitops/components/delegation/delegation-service.yaml
+++ b/openbank-infra/gitops/components/delegation/delegation-service.yaml
@@ -147,14 +147,13 @@ spec:
             # app dies with InactiveBeanException.
             - name: QUARKUS_REDIS_HOSTS
               value: redis://redis.delegation.svc:6379
-            # Kept ADVISORY (false) for the first rollout: the interceptor evaluates and
-            # records every decision but does NOT block while false, so the "would DENY"
-            # population can be confirmed empty in the live decision log before the
-            # enforce flip (a separate, deliberate follow-up per the rules.yaml guardrail).
-            # Flipping this before the sidecar is confirmed serving means failing closed on
-            # every call.
+            # OPA sidecar enforced (issue #3679). Advisory through the first rollout; coverage
+            # verified before the flip: delegation_rest_ext.rego grants all ten actions with the
+            # caller-narrowed rules (operator writes exclude service accounts, the edge principal
+            # is scoped to the customer action set, delegation.check is the only M2M action), and
+            # the sole declared caller is the admin-ui relaying a human operator's token.
             - name: AUTHZ_ENFORCE
-              value: "false"
+              value: "true"
           resources:
             requests:
               cpu: 100m


JiRaska:
### Independent coverage check — supports this PR, one gate left

Working #3679 from the other end, I measured the policy this flip would activate. Method: `opa eval data.openbank.rest.allow` against the **live** `delegation-opa-bundle` ConfigMap pulled from the cluster (not the repo copy), across all ten `delegation.*` actions × four principal shapes, with a must-DENY control (`zzz.frobnicate`) and a must-ALLOW control in the same run — both behaved, so the result is not vacuous.

Also confirmed live: `delegation-service-848b86b9cb-jb8mb` runs `delegation-service,opa` with readiness `true,true`, so the fail-closed-with-no-PDP failure mode does not apply here.

Every action resolves for its intended principal. The DENYs are exactly the ones `delegation_rest_ext.rego` documents as deliberate:

- `service-account-openbank-services` (shared backend): DENY on offer/accept/decline/renounce/revoke/suspend/reinstate; ALLOW on `delegation.check` only, plus read/list via `operator-read-any`
- `service-account-openbank-edge`: ALLOW on the seven customer actions, DENY on `suspend`/`reinstate`
- real operator/admin: ALLOW on everything

Worth noting for the reviewer: `role_action_matrix` contains **no** `delegation.*` entry, so unlike most services `delegation_rest_ext.rego` is fully load-bearing here and `matrix-allows` cannot leak past its exclusions. That makes this the one of the three services in #3679 where enforcement genuinely changes the security posture rather than restating the role gate — and it is unaffected by #3765.

**The one thing I could not measure**: my caller inventory is derived from reading this repo, so a caller that exists only in the cluster — a script, a hand-minted token, a service outside this repo — is invisible to it. Advisory mode has been logging "would DENY" decisions this whole time; reading that log for a representative window is the evidence that closes it, and it is cheap relative to a money-path 403. I did not read it.

Full per-service table on #3679.

JiRaska:
## Automated agent review — coverage checks out; 2 things to confirm before the flip

**Verdict: no structural defect found.** Unlike the sdd enforce PR (#3697), this one flips a service that already has a working PDP, and the policy genuinely covers all ten actions. The residual risk is operational, not a coverage hole.

### What I verified against `origin/main`

`delegation-service.yaml` already ships the `opa` sidecar (`openpolicyagent/opa:1.17.0`, `--addr=0.0.0.0:8181`, `/health?bundles=true` readiness, six subPath mounts, `delegation-opa-bundle` ConfigMap, pod-roll checksum `3b4f321fc40e98b3`). `OpaSidecarPolicyDecisionPoint` defaults to `http://localhost:8181`, which matches — so this is not a fail-closed-on-every-endpoint flip.

`DelegationResource` has ten `@Authorize` actions: `offer, read, list, accept, decline, renounce, revoke, suspend, reinstate, check`. Against `delegation_rest_ext.rego` on `main`:

- `operator-delegation-write` — `HUMAN` + `ROLE_OPERATOR|ROLE_ADMIN` + `not startswith(principal.id, "service-account-")` + `startswith(action, "delegation.")` → covers all ten for real staff sessions. The service-account exclusion is present and correct (this is exactly what #3697 is missing for sdd).
- `edge-service-delegation` — `service-account-openbank-edge`, enumerated: `offer, read, list, accept, decline, renounce, revoke` (7 of 10). `suspend`/`reinstate` withheld from the edge, which is right — those are back-office/fraud-signal actions.
- `service-delegation-check` — any `service-account-*` for `delegation.check` only. That is the M2M enforcement lookup and it is correctly identity-shaped rather than gated on the unreachable `SERVICE` principal type.

No rule anywhere gates on `input.principal.type == "SERVICE"`. Coverage is complete; I found no action that starts answering 403.

### 1. The PR body's caller claim is narrower than reality — worth correcting, not blocking

The new comment says *"the sole declared caller is the admin-ui relaying a human operator's token."* That is not what the policy assumes: `edge-service-delegation` exists precisely because `service-account-openbank-edge` is a second caller, and `service-delegation-check` admits a third class (any backend doing an enforcement lookup). All three are covered, so the flip is still safe — but if the reviewer takes the one-caller sentence at face value they will not check the other two, and that sentence is the artifact that outlives this PR.

### 2. No advisory-log evidence is cited

`rules.yaml`' guardrail (and the comment this diff replaces) asks for the live decision log to show an empty would-DENY population before the flip. The new comment asserts coverage *by reading the policy*, which is what I did too — and static coverage cannot see a caller nobody wrote down. If the advisory log has been checked, quoting the would-DENY count in the PR body converts this from an argument into evidence; if it has not, that is the one remaining step.

Failure mode if a fourth caller exists: it starts getting `403` on the affected `delegation.*` action. Probes will not catch it — `startupProbe`/`readinessProbe` are `tcpSocket` and `livenessProbe` hits `/q/health/live` on the management port, none of which traverse `@Authorize`. `delegation.check` is the one to watch: it is the enforcement lookup other services make, so a denial there degrades *their* authorization decisions, not just delegation's own API.

### Verified clean
- PDP exists, is served over loopback, and the app's default PDP URL matches the sidecar's listen address — no 503-on-every-endpoint risk.
- All ten `@Authorize` actions have at least one matching allow rule.
- No `SERVICE`-principal-type rule; M2M callers identified by `principal.id`.
- Write rules are not gated on `HUMAN` + `ROLE_OPERATOR` alone — the `service-account-` exclusion is in place, so the shared `openbank-services` client does not inherit delegation writes.
- No `.rego` or `rules.yaml` edit, so no OPA bundle restamp is owed. One-file diff, no bundle drift.

---
*Automated agent review — this is **not** an approval and no approval is implied. Authorization changes need human review regardless of what any automated check says.*

JiRaska:
Reviewed for coverage. **No action becomes unreachable for a caller that legitimately uses it today — safe to flip.**

**Measured against `main`'s committed bundle, not this branch's.** The branch is 306 commits behind and its `rules-opa-data.yaml` predates #3951 (which promoted `shared_m2m_write_prohibition` to a top-level key so OPA reads it at all): `grep -c` gives **0** here and **1** on `main`. Evaluating the embedded copy would answer about a policy the fleet no longer runs. Input shape is the real one — `resource: {type, id}` — since `OpaSidecarPolicyDecisionPoint.toInput()` emits nothing more, and a probe that invents `resource.attributes.tenant` reaches `operator-on-own-tenant`, an action-agnostic rule that is dead in production and will happily allow a nonsense action.

| principal (`ROLE_OPERATOR`) | `delegation.offer` | `delegation.suspend` | `delegation.read` | `delegation.check` | `zzz.frobnicate` |
|---|---|---|---|---|---|
| a human operator | ALLOW | ALLOW | ALLOW | ALLOW | DENY |
| `service-account-openbank-edge` | ALLOW | **DENY** | ALLOW | ALLOW | DENY |
| `service-account-openbank-services` | **DENY** | **DENY** | ALLOW | ALLOW | DENY |

The `frobnicate` row-end is the must-DENY control; without it the DENYs would prove nothing.

**Caller check.** The only in-repo caller is `CustomerDelegationResource` in customer-edge, and `UpstreamClient` defaults `openbank.upstream.client-id = openbank-edge` with no override in `customer-edge.yaml`, so it arrives as `service-account-openbank-edge`. Its routes — shared-by-me, shared-with-me, getById, offer, accept, decline, renounce, revoke — map onto exactly the actions that stay ALLOW. `suspend`/`reinstate` are admin-ui-only and correctly closed to both service identities. No backend service calls a delegation write.

`delegation.check` staying ALLOW for both service accounts matters more than it looks: other services' authorization decisions depend on it, so it is the one to watch if anything does go wrong after the flip.

**PDP parity clean** — the workload carries the `opa` sidecar on `0.0.0.0:8181`, matching the interceptor's loopback default; `check-authz-pdp-parity.py` reports 0 enforce-without-PDP violations. No 503-on-every-endpoint risk.

Worth noting the shape that makes this one solid: `delegation_rest_ext.rego` carries its own `not startswith(principal.id, "service-account-")` exclusion, which I confirmed fires. It does not depend on the shared prohibition register.

**Residual risk, same as its sibling #3695 and stated plainly:** this evidence is static. A caller that exists only in the cluster is invisible to it, and no probe catches a 403 — startup/readiness are `tcpSocket`, liveness is `/q/health/live` on the management port, neither traverses `@Authorize`. The advisory-window would-DENY counter is the evidence that would close it, and no comment here cites it.

No `input.principal.type == "SERVICE"` rule in this path — correctly identity-scoped by `principal.id`.

Issue #1478, opened 2026-07-17
money_path_services registry omits sdd-service, aml-service, interest-service despite real money movement


JiRaska:
## Summary

\`openbank-libs/governance/rules.yaml: money_path_services\` (the single source of truth for
the 2-approval + threat-model requirement, ADR-0030) does not list \`openbank-sdd-service\`,
\`openbank-aml-service\`, or \`openbank-interest-service\` — found while reviewing the #1201
outbox atomic-claim PRs (#1460 sdd, #1468 interest, #1472 aml), none of which required a
second approval or a threat model under the current registry, unlike their siblings
\`openbank-billing-service\`/\`openbank-lending-service\`/\`openbank-account-service\`/
\`openbank-sepa-payment\` (#1462/#1469/#1470/#1471), which are all correctly listed and all
already have a \`docs/threat-models/\` entry.

## Evidence these move real money, not just adjacent to it

- **sdd-service** (#1000, closed): \`SddCollectionDebitConsumer\` posts a real debit via
  \`TransactionServiceClient\` for every authorised direct-debit collection (PR #1027,
  merged). This is not a "gates money movement" service like \`sca\`/\`consent\`/\`vop\`
  (which the registry's comments explicitly reason about) — it directly moves funds.
- **interest-service** (#999, still open): withholding tax is assembled and booked but never
  remitted; the service also runs \`capitalize()\` accrual/write-off postings
  (see #1265, #1245/#1253) that hit the general ledger directly.
- **aml-service**: currently mostly design/consumer work (#1035 still open, no
  \`sanctions.screening.event\` consumer exists yet) — weaker case today than the other two,
  but a sanctions-screening decision (clear/hold) is exactly the kind of maker-checker action
  \`openbank-sanctions-service\` is already listed for. Worth the same registry review once
  #1035 ships a real decision endpoint.

No \`docs/threat-models/\` entry exists for any of the three today either.

## Ask

Review whether \`openbank-sdd-service\` and \`openbank-interest-service\` (at minimum) should be
added to \`money_path_services\`, each with the reasoning comment style already used for every
other entry (see \`openbank-vop-service\`'s entry for the "gates vs moves money" precedent), and
whether a threat model is needed before the next PR touching either service's write path
merges. \`aml-service\` can wait for #1035 to land a real decision endpoint, but is flagged here
so it isn't forgotten.

[...]

Issue #2442, opened 2026-07-25
163 @RolesAllowed sites name a role no realm issues (ROLE_SERVICE ×152 in 28 modules, 16 money-path) — the M2M caller is silently denied


JiRaska:
## Problem

`check-roles-allowed-realm.py` (enforced since #2418) blocks the unreachable case — an `@RolesAllowed` where *every* role is absent from the realm. It only **warns** on the partial case: a dead role name sitting in a list that also names a live one. On `origin/main` that warning reads:

```
::warning title=RolesAllowed realm parity::163 @RolesAllowed site(s) name a role no realm issues alongside a live one — the caller it was reserved for is silently denied (ROLE_SERVICE×152, ROLE_CREDIT_RISK×9, ROLE_LENDING_OFFICER×6).
```

Humans still get in through the live role, so nothing looks broken. The caller the dead name was reserved for — in almost every case the M2M service-to-service path — is denied, silently, and has been since the annotation was written.

## Inventory (measured on `64a1f9be4`)

**`ROLE_SERVICE` — 152 sites, 28 modules.** No realm declares it: not the staff realm, not the customers realm.

| module | sites | | module | sites |
|---|---|---|---|---|
| document-service | 14 | | party-service | 5 |
| dispute-service | 13 | | tpp-registry-service | 5 |
| account-service | 10 | | sdd-service | 4 |
| notification-service | 10 | | statement-service | 4 |
| ledger-service | 9 | | transaction-service | 4 |
| pid-service | 9 | | clearing-simulator | 3 |
| standing-order-service | 9 | | kyc-service | 3 |
| balance-service | 8 | | aml-service | 2 |
| interest-service | 8 | | sepa-instant | 2 |
| sca-service | 8 | | sepa-payment | 2 |
| clearing-service | 7 | | anacredit, consent, domestic-payment, fraud, settlement, **libs-runtime** | 1 each |
| sanctions-service | 7 | | | |

**16 of the 28 are money-path services** (`rules.yaml: money_path_services`): account, balance, clearing, consent, domestic-payment, fraud, interest, ledger, lending, sanctions, sca, sdd, sepa-instant, sepa-payment, settlement, transaction.

[...]

PR #3698, opened 2026-08-04
fix(interest): prohibit service-account principals from interest writes

JiRaska:
## Summary

Follow-up to PR #3695 (interest `AUTHZ_ENFORCE=true`) and issue #3679. The flip surfaced a trust-boundary widening: `operator-interest-write` was role-only, and the `rules.yaml` `role_action_matrix` grants `interest.create/trigger/delete` to `ROLE_OPERATOR`. Two realm service accounts carry that role — `service-account-openbank-edge` (customer-reachable) and `service-account-openbank-services` (shared backend client) — so both could reach these writes via `matrix-allows`. Interest is money-path (posts real GL journals, #1478) and the writes are bank-side actions invoked by admin-ui operators only. This PR fails closed for every service account.

## Type of change

- [ ] feat
- [ ] refactor
- [ ] perf
- [ ] docs
- [ ] chore
- [ ] security (private until disclosed) — flip is already public via #3695; disclosing the exact narrowing shape is safe
- [ ] breaking change

## Linked issues

Refs #3679, #3695

## Changes

**Interest ext — the tightening (two layers, both needed):**
- `operator-interest-write` gains `not startswith(input.principal.id, "service-account-")` (delegation idiom).
- New `prohibited if { startswith(input.principal.id, "service-account-"); input.action in {create,trigger,delete} }` — vetoes the writes at the allow head, so no reason present or future can admit a backend client. Reads are untouched (the edge legitimately serves customer interest views).

**Shape:**
- The ext moves from a generator heredoc to a **standalone** `interest_rest_ext.rego` (delegation pattern) so `opa test` can load it.
- New `interest_rest_ext_test.rego` — 9/9 green, falsified locally (stripping either layer turns 5 tests red).
- Generator regenerated the interest bundle + policy-checksum annotation (interest only — no fleet ripple).

**New CI step (opa-policy.yml):**
- Runs every `*_rest_ext_test.rego` trio (rest.rego + its ext + the test) — same-package cross-contamination requires one invocation per trio. These previously ran only by hand. **First sweep found a stale test:** aml's `test_service_m2m_excluded_from_human_rules` asserted `count(allowed_reasons) == 0` for a `ROLE_COMPLIANCE` M2M read, but base `compliance-read-any` deliberately grants M2M reads ("how real M2M callers fetch cross-service data" — rest.rego). The test only ever passed under its header's obsolete base-less invocation. Fixed to assert the real invariant (exclusion from aml-specific reasons); aml policy itself is unchanged.

**Threat model updated:** docs/threat-models/openbank-interest-service.md carries the trust-boundary-change entry — required by the `threat-model-updated-on-trust-boundary-change` gate.

## Verification evidence

- `opa test` per trio: interest 9/9, aml 10/10 (fixed), delegation/consent/dispute/kyc/psd2/statement/tpp-registry/card-issuance/standing-order/audit — 11 suites, all green.
- Falsification: stripping the ext's prohibition and the service-account exclusion turns 5 interest tests red (as they must).
- Gates locally: gitops group PASS=20 (known local-env artifacts only), security group PASS=11 including `threat-model-updated-on-trust-boundary-change` with `PR_DIFF_BASE=origin/main`.

- [ ] Threat model change is present (section 6, 2026-08-03 entry).
- [ ] 2 approvals.
- [ ] Confirm no legitimate caller is broken — the fleet audit in the ext file says no M2M writer exists; if a new one appears before this lands, register it with an identity-scoped allow reason first.

## Security checklist

- [x] No secrets / PII.
- [x] Auth change → `security-review-required`.
- [x] No policy logic removed; two overlapping fail-closed layers added on top of existing enforcement.

## Compliance impact

- [x] DORA — ICT access control narrowed on a money-path service.


JiRaska:
**Review: hole closed — no blocking findings. This is the PR the other nine should be measured against, and I would land it first.**

**The #3764 hole cannot occur here, because the veto is keyed on the identity *prefix*, not one client id:**

```
prohibited if {
	startswith(input.principal.id, "service-account-")
	input.action in {"interest.create", "interest.trigger", "interest.delete"}
}
```

That closes both `service-account-openbank-edge` and `service-account-openbank-services` — and every future service account — on exactly the three actions `rules-opa-data.yaml:278-280` grants to ROLE_OPERATOR. Traced: for the shared client on `interest.create`, `operator-interest-write` is excluded, `matrix-allows` fires, and this clause vetoes at the `allow` head (`rest.rego:50`). `rest.allow` is false. The siblings **#3740, #3744, #3755, #3758, #3760, #3761** and **#3764** all pin their veto to `service-account-openbank-edge` alone and leave the shared client admitted through `matrix-allows`; this PR is the shape that does not.

**The test is also the strong form.** `test_shared_m2m_may_not_write_despite_matrix_grant` asserts `rest.allow == false` for `service-account-openbank-services` on `interest.create`. That is precisely the assertion whose *absence* lets the hole survive in the siblings, where the equivalent test asserts only that a *reason* is absent. Together with `test_prohibition_fires_for_any_service_account` (which proves the veto, not merely a missing reason) and the two `test_edge_may_read`/`_list` cases (which prove reads survive and that the base+matrix wiring is live), the suite is falsifiable in both directions.

**And it closes the gap that makes every other PR in this family's evidence unverifiable at merge time.** The new `Run per-service rest_ext unit tests` step in `.github/workflows/opa-policy.yml` is, as far as I can tell, the first thing in CI that ever executes a `*_rest_ext_test.rego`: `openbank-infra/opa/build-bundle.sh:37` runs `opa test "$POLICIES_DIR" "$LIBS_POLICIES_DIR"` and nothing walks `openbank-infra/gitops/components/**`, and no `gates.yaml` entry does either. Three details of the step are right for reasons this repo has been burnt by before:
- the file set is **discovered**, not hand-listed (the `pact-drift-check.yml` failure mode);
- **zero discoveries fails closed** with an explicit error, so a wrong `find` pattern cannot render as a pass;
- each ext runs as its own trio, so same-package `allowed_reasons` from different services cannot cross-contaminate.

One consequence worth stating out loud for the merge order: **once this lands, the other nine suites become CI-enforced — and they will all be green with their holes present**, because the weak reason-absent assertion passes either way. A green `opa-policy` check on those PRs after this merges is therefore not evidence the shared-client path is closed. That is an argument for landing this one first *and* fixing the assertion form in the siblings, not for landing it later.

**Verified clean:** `operator-interest-write` gains the `not startswith(...)` exclusion as well as the veto, so both paths are closed rather than one; the `rules_mock` includes the full `data.rules` shape the base policy reads (`money_path_services`, `four_eyes`, `feature_flags`, `shared_m2m_write_prohibition`) rather than only the matrix, which is the most honest mock in the batch; `interest.read`/`interest.list` remain open to the edge and are pinned; the stale `aml_rest_ext_test.rego` header fix rides along correctly; checksum `99e2f4cfcc50c175` consistent between `interest-opa-bundle.yaml` and the pod-roll annotation in `interest-service.yaml`; no `principal.type == "SERVICE"` rule.

---

PR #3692, opened 2026-08-03
ci(gates): make the Admin UI golden-path E2E merge-blocking via Validate manifests

JiRaska:
## Summary

Makes the Admin UI golden-path Playwright suite **merge-blocking** — without any ruleset edit. The suite has been enforced inside `ui-build` since #653 (hosted `ubuntu-latest` only, stable), but the `Admin UI` aggregator was never in a required context, so a red E2E did not block merges. `Validate manifests` (already a required context) now also depends on the `ui` aggregator — the established in-repo pattern for binding a job into the required set.

## Type of change

- [ ] feat (new functionality)
- [ ] fix (bug fix)
- [ ] refactor (no behavior change)
- [ ] perf (performance)
- [ ] docs
- [x] ci
- [ ] security (private until disclosed)
- [ ] breaking change

## Linked issues

Closes #3675 (E2E half; the pitest-matrix half is PR #3687)

## Changes

- `ci.yml` `validate` job: `needs: [gates]` → `needs: [gates, ui]` + a fail-closed step on `needs.ui.result`. The `ui` aggregator already encodes the skip/cancel semantics (#2185): PRs that touch no admin-ui file pass as correctly-skipped; any real build/E2E failure fails the required check.
- `openbank-admin-ui/CLAUDE.md`: the stale "E2E is advisory (continue-on-error)" bullet corrected — the `continue-on-error` only ever covered the system-deps install step; the suite itself is enforced and now merge-bound.

## Why this shape

- The required context name (`Validate manifests`) is unchanged, so no GitHub-side ruleset edit is needed — a ruleset change cannot ship in the same PR as the thing it protects; this one does.
- Path-scoping is preserved: E2E becomes a gate exactly for admin-ui-touching PRs, not a tax on every PR.
- Fails closed on `cancelled`/`skipped`/anything-not-success, mirroring the existing gate-shard step.

## Definition of Done

- [x] Hexagonal layering preserved — CI only.
- [ ] Tests — the PR's own run is the falsification: this PR touches no admin-ui path, so `ui` must pass as skipped and `Validate manifests` must stay green (verifiable in this PR's checks).
- [x] `./gradlew` gates — N/A; `ci.yml` parses (yaml.safe_load).
- [x] No new lint warnings.
- [x] Docs updated (admin-ui CLAUDE.md).

## Security checklist

- [x] No secrets / PII.
- [x] No new suppressions / dependencies.
- [ ] Auth / crypto / payment code — N/A.

## Compliance impact

[...]

PR #3700, opened 2026-08-04
feat(fraud-service): add in-image ONNX serving smoke gate (#3354)

### LINKED ISSUE #3354: fraud-service ONNX has never loaded: glibc-linked native library in a musl (alpine) runtime image

JiRaska:
In-image smoke gate for the fraud-service ONNX serving path (#3354).

This PR is deliberately step 1 of 2. It adds the gate that proves the defect before we swap the whole fleet runtime base.

What changed
- New `OnnxServingSmokeKt` main in `openbank-fraud-service`: loads OrtEnvironment, verifies the bundled ADR-0141 model card, and runs one real inference round-trip with no container, no DB and no config.
- `OnnxFraudModel.MODEL_RESOURCE_PATH` / `MODEL_CARD_PATH` widened to `internal` so the smoke reuses the same load path as the bean (no reimplementation).
- `.github/workflows/Dockerfile.deploy` now accepts `BASE_IMAGE` and auto-detects whether to use `addgroup/adduser` (alpine) or `groupadd/useradd` (ubi-minimal). Default stays the current alpine image, so deployed behaviour is unchanged today.
- New `.github/workflows/onnx-serving-smoke.yml` builds the fraud image from the same `Dockerfile.deploy` twice — once with the historical alpine base (negative control, must fail) and once with `eclipse-temurin:25-jre-ubi10-minimal` (positive control, must pass).

Verification
- Local: `detekt`, `ktlintCheck`, `compileKotlin` for `openbank-fraud-service` pass.
- Local Docker: alpine image → `SMOKE FAIL: OrtEnvironment did not load`; ubi10-minimal image → `SMOKE OK ...`.

Linked issues
Closes #3354 (PR1; fleet-wide base swap tracked for follow-up PR2)

### DIFF
--- a/.github/scripts/check-gradle-cache-writers.py
+++ b/.github/scripts/check-gradle-cache-writers.py
@@ -130,6 +130,12 @@ DECLARED: dict[str, tuple[str, str]] = {
         "disabled",
         "Boot probe; builds one service and needs no cross-run Gradle state.",
     ),
+    "onnx-serving-smoke.yml::smoke": (
+        "read-only",
+        "Consumer; restores fleet-lint's home. Builds one service's quarkusBuild to smoke "
+        "the ONNX serving path inside the shipped image (#3354), on a schedule rather than "
+        "per-push, so it neither needs nor should store a per-run entry.",
+    ),
     "pitest.yml::pitest": (
         "read-only",
         "Demoted from setup-java. Consumer; restores fleet-lint's home. The setup-java "
--- a/.github/workflows/onnx-serving-smoke.yml
+++ b/.github/workflows/onnx-serving-smoke.yml
@@ -6,10 +6,22 @@
 # SAME Dockerfile.deploy the deploy pipeline uses and runs a tiny no-container main inside it
 # that forces the real load path.
 #
-# Two legs run on every PR:
-#   - alpine (the historical base) MUST fail — it is the negative control that proves the gate
-#     can go red, not only green.
-#   - eclipse-temurin:25-jre-ubi10-minimal (glibc) MUST pass — it is the candidate runtime for PR2.
+# ONE leg runs: the image the fleet actually ships, built from Dockerfile.deploy with no
+# override. The base is pinned there (#3618/#3629 settled it on eclipse-temurin:25-jre) and
+# `verify-image-native-libs.py` refuses a second copy of it, so this workflow must not carry one.
+#
+# WHERE THE NEGATIVE CONTROL WENT. An earlier revision built a second leg on the old alpine base
+# and required it to FAIL, proving the gate could go red. That leg is gone, and deliberately:
+#   - it needed `ARG BASE_IMAGE` in Dockerfile.deploy, which is exactly the second copy of the
+#     base that verify-image-native-libs.py forbids — and with the ARG removed, `--build-arg`
+#     is accepted with a warning and silently builds the SAME base, so the leg would have
+#     "proved" a failure while testing the passing image. A negative control that cannot fail is
+#     worse than none.
+#   - the regression it guarded (going back to musl) is now caught on every push by
+#     verify-image-native-libs.py, which runs `ldd` over the bundled .so INSIDE the pinned base.
+# So this gate answers the narrower question that script cannot: does the serving path actually
+# WORK in the shipped image — native library, model, and ADR-0141 card, through the same
+# companion functions the bean uses.
 #
 # The smoke is intentionally not a @QuarkusTest: it needs no DB, no Kafka, no config; it only
 # needs the image filesystem and the bundled model/card.
@@ -31,8 +43,9 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  # Candidate glibc runtime for the fleet-wide base swap (PR2 of #3354).
-  GLIBC_BASE: eclipse-temurin:25-jre-ubi10-minimal@sha256:6031e5480f5d5818252a348980a192af0eab09e3380a55f75927e019f4d8136c
+  # No base pinned here on purpose: Dockerfile.deploy is the single source, and a second copy
+  # is what verify-image-native-libs.py exists to prevent. (This used to pin ubi10-minimal —
+  # the candidate that was NOT adopted, so the gate would have tested an image nobody ships.)
   SERVICE: openbank-fraud-service
   SMOKE_CLASS: com.openbank.fraud.infrastructure.ml.OnnxServingSmokeKt
 
@@ -70,58 +83,24 @@ jobs:
             -Dquarkus.package.jar.type=fast-jar \
             --console=plain
 
-      # Pre-pull both bases so build errors are clearly attributable to the image, not to a
-      # registry timeout, and so the negative-control leg starts from a known state.
-      - name: Pull base images
+      # Pre-pull the base Dockerfile.deploy pins, so a build error is attributable to the image
+      # rather than to a registry timeout. Read from the file — never hardcoded here.
+      - name: Pull the pinned runtime base
         run: |
-          docker pull eclipse-temurin:25-jre-alpine@sha256:28db6fdf60e38945e43d840c0333aeaec66c15943070104f7586fd3c9d1665b0
-          docker pull ${{ env.GLIBC_BASE }}
+          set -euo pipefail
+          base="$(grep -m1 '^FROM ' .github/workflows/Dockerfile.deploy | awk '{print $2}')"
+          echo "runtime base (from Dockerfile.deploy): $base"
+          docker pull "$base"
 
-      # Negative control: the current alpine base is EXPECTED to fail because onnxruntime's
-      # libonnxruntime.so is built against glibc. A "pass" here means the gate is broken.
-      - name: "Smoke: alpine base (must FAIL)"
-        id: alpine
-        continue-on-error: true
+      # The shipped image, built exactly as the deploy pipeline builds it — no --build-arg,
+      # no override. If this fails, the ONNX serving path does not work in what we deploy.
+      - name: "Smoke: the shipped image must serve"
         run: |
           set -euo pipefail
           docker build \
             -f .github/workflows/Dockerfile.deploy \
-            -t fraud-smoke-alpine \
+            -t fraud-smoke \
             "${{ env.SERVICE }}/build"
-          docker run --rm --entrypoint="" fraud-smoke-alpine \
+          docker run --rm --entrypoint="" fraud-smoke \
             java -cp "/app/lib/boot/*:/app/lib/main/*:/app/app/*" \
               ${{ env.SMOKE_CLASS }}
-
-      # Positive control: the candidate glibc base should load the native library, verify the
-      # model card and return a sane shadow score.
-      - name: "Smoke: glibc base (must PASS)"
-        id: glibc
-        run: |
-          set -euo pipefail
-          docker build \
-            --build-arg BASE_IMAGE=${{ env.GLIBC_BASE }} \
-            -f .github/workflows/Dockerfile.deploy \
-            -t fraud-smoke-glibc \
-            "${{ env.SERVICE }}/build"
-          docker run --rm --entrypoint="" fraud-smoke-glibc \
-            java -cp "/app/lib/boot/*:/app/lib/main/*:/app/app/*" \
-              ${{ env.SMOKE_CLASS }}
-
-      # Report the verdict explicitly. The step names above make the raw logs readable; this
-      # step turns the alpine failure into an expected pass and a glibc failure into a real fail.
-      - name: Verify red/green outcome
-        run: |
-          set -euo pipefail
-          alpine_ok="${{ steps.alpine.outcome }}"
-          glibc_ok="${{ steps.glibc.outcome }}"
-          echo "alpine outcome: $alpine_ok (expected: failure)"
-          echo "glibc outcome:  $glibc_ok (expected: success)"
-          if [ "$alpine_ok" != "failure" ]; then
-            echo "::error::negative-control alpine leg did NOT fail — the gate cannot go red"
-            exit 1
-          fi
-          if [ "$glibc_ok" != "success" ]; then
-            echo "::error::positive-control glibc leg did NOT pass — the base swap is not safe"
-            exit 1
-          fi
-          echo "SMOKE GATE OK: alpine fails as expected, glibc passes"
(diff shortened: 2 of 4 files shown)


JiRaska:
**Verdict: excellent gate design (built-in negative control), but two reachability/CI issues need fixing before merge.** 2 findings.

- **This PR's own CI is currently red because of its own new workflow**: `gates (lint)` fails with `UNDECLARED onnx-serving-smoke.yml::smoke uses the Gradle cache (mode read-only) but is not in DECLARED. Add it with a reason — a new writer costs every other job cache headroom (#3299).` (run `30883020563`, job `gates (lint)`). `.github/workflows/onnx-serving-smoke.yml`'s `smoke` job calls `gradle/actions/setup-gradle` with `cache-read-only: true` but was never added to the fleet's Gradle Actions-cache writer-budget declaration. This isn't staleness — it's this PR's own new job tripping an existing enforced gate — and it currently fails `Validate manifests` too, since that job aggregates `gates`.
- **`ONNX serving smoke` is not a required status check** (confirmed against the live `main-protection` ruleset: the five required contexts are `all-green`, `Validate manifests`, `Gitleaks`, `issue-hygiene`, `OPA policy gate` — this new workflow isn't among them, and being `paths:`-filtered it structurally can't become one without the same binding-copy-in-`Validate-manifests` treatment this repo has applied elsewhere, e.g. #3692 in this same batch). The PR is explicit that this is "step 1 of 2" whose job here is to *prove* the defect before #3709's fleet-wide base swap — a reasonable framing for a diagnostic PR — but once #3709 merges and the fleet default flips to glibc, nothing stops a future regression back to an alpine `Dockerfile.deploy` default from merging cleanly, since this smoke workflow can only ever be advisory as currently wired. Worth a follow-up issue (or noting explicitly here) that the binding half is still owed.

Genuinely strong part of the design, worth calling out: the smoke workflow builds in its own falsification rather than assuming it — the alpine leg is a **negative control that must fail** (`continue-on-error: true`, then a `Verify red/green outcome` step that itself fails if alpine's outcome isn't `failure` or glibc's isn't `success`). That's a rare instance in this batch of a gate that can prove it can go red, on day one, without waiting for an incident to exercise it.

Verified clean: `OnnxFraudModel.MODEL_RESOURCE_PATH`/`MODEL_CARD_PATH` are widened to `internal` (not `public`) so the smoke reuses the real load path rather than reimplementing it — avoids the class of drift where a test fixture diverges from the production code path.

JiRaska:
## Rebase attempted and deliberately stopped — this conflict needs an authoring decision, not a resolution

I rebased this onto the post-#3618 `main`. The ADR-registry conflict (`DIGEST.md`, `README.md`, `index.json`) resolved mechanically — took main's copies and re-ran `docs/adr/gen-index.sh`, 239 ADRs indexed. The second conflict is not mechanical, so I aborted rather than pick a side. The branch is untouched.

### The conflict

`.github/workflows/Dockerfile.deploy`. Both PRs changed how this file names its base image, in incompatible ways.

**`main` now (from #3618):**

```dockerfile
FROM eclipse-temurin:25-jre@sha256:681c543d…
```

hardcoded, with a comment that states outright:

> `.github/scripts/verify-image-native-libs.py` runs `ldd` on every bundled linux `.so` INSIDE this exact base before any image is pushed, and fails the push. **It reads the base image from THIS file — never give it a second copy.**

**This PR:**

```dockerfile
ARG BASE_IMAGE=eclipse-temurin:25-jre-alpine@sha256:28db6fdf…
FROM ${BASE_IMAGE}
```

plus portable user creation that works on both alpine and RHEL bases.

### Why I did not just take main's side

The `ARG` is **load-bearing for this PR's whole point**. `onnx-serving-smoke.yml:102` passes `--build-arg BASE_IMAGE=${{ env.GLIBC_BASE }}`, so the gate builds the image twice — once on the default and once on glibc — and thereby proves the failure *and* the fix. That two-direction falsification is the reason this gate is worth having; a gate that only ever builds the base that works cannot show the failure it exists to catch.

Delete the `ARG` and the gate loses a leg. Keep the `ARG` and the Dockerfile carries a second base-image declaration, which is precisely what #3618's comment warns `verify-image-native-libs.py` against — and this PR's default is still the **alpine** digest, now wrong.

So there are three honest ways forward and they are not equivalent:

1. **Keep `ARG`, change its default to the glibc digest**, and confirm `verify-image-native-libs.py` resolves `ARG`-substituted `FROM` lines rather than reading the literal. Smallest change; needs that verifier check to actually hold.
2. **Keep main's hardcoded `FROM`**, and have the smoke gate build its alpine leg from a small dedicated Dockerfile instead of parameterising the production one. Keeps the "one authoritative base" property #3618 wanted, at the cost of a second file that must not drift.
3. **Keep main's `FROM` and drop the alpine leg**, accepting that the gate only proves the glibc base loads. Cheapest, and the weakest — it stops being a falsification and becomes a smoke test.

I would take (1) if the verifier resolves ARGs and (2) if it does not, but that is an authoring call about what this gate is for, so I am leaving it.

### Context that may matter

**#3709 is now closed** — the fleet base decision went to `eclipse-temurin:25-jre` (#3618, merged 09:42; #3629 retires alpine from the 53 declaring files behind it). So the "must accept either base" requirement this PR was written under no longer holds: there is one base now, and the only reason to parameterise is to let *this gate* test the other one.

Also worth knowing while you are in here: the scanner (`Dockerfile.scanner`, `openbank-security-scanner/Dockerfile.prebuilt`), `notification-service/Dockerfile.runtime` and the local templates under `openbank-infra/docker/services/` stay on alpine after #3618/#3629 — #3709 covered them and it is closed. Nothing there bundles a native library today, so nothing is broken, but the next one that does fails exactly the way fraud-service did.

JiRaska:
## Rebase attempted again and stopped again — but the decision is now much clearer

Asked to rebase this. The ADR-registry conflict resolved mechanically (`docs/adr/gen-index.sh`, 239 ADRs). The `.github/workflows/Dockerfile.deploy` conflict is the same one as this morning, and I stopped rather than pick — but the facts around it have changed enough to point at an answer.

### This PR is NOT superseded by #3629

Worth stating first, because it looks that way. `verify-image-native-libs.py` (merged with #3629) and this PR's `OnnxServingSmoke.kt` are complementary:

- **the verifier** runs `ldd` on every bundled `.so` inside the real base — it proves the libraries **link**.
- **this smoke main** runs *inside the built image* and walks the real serving path: `OrtEnvironment` initialises, the model and its ADR-0141 card load **through the same companion functions the bean uses**.

A successful `ldd` does not prove `OrtEnvironment.getEnvironment()` succeeds, nor that the model file parses. The deeper check is worth having.

### What changed about the conflict

This morning the base-image question was open, so `ARG BASE_IMAGE` was a reasonable way to keep the Dockerfile neutral. It is now settled: **#3618 and #3629 merged**, `main` pins

```dockerfile
FROM eclipse-temurin:25-jre@sha256:681c543d…
```

and `verify-image-native-libs.py`'s own header states the constraint outright:

> Read the runtime base from `.github/workflows/Dockerfile.deploy` — the single source every image producer copies. **NEVER give this script its own copy of the base**: a second copy …

`ARG BASE_IMAGE=<alpine digest>` **is** that second copy, and its default is now the wrong image. So resolving in this PR's favour would contradict a merged decision and defeat the verifier that decision shipped.

### Why I still did not just take `main`'s side

Because it is not free, and the cost is the interesting kind. `onnx-serving-smoke.yml:102` passes `--build-arg BASE_IMAGE=${{ env.GLIBC_BASE }}` to build the image **twice** — once on the default (alpine) to prove the failure, once on glibc to prove the fix. Drop the `ARG` and Docker accepts the `--build-arg` for an undeclared argument with only a warning, so **both legs would silently build the same glibc base** and the "proves it fails on alpine" leg would pass while testing nothing. That is precisely the vacuous-green failure this PR was written to prevent, reintroduced by its own rebase.

So the resolution is a design decision, not a merge decision:

1. **Drop `ARG`, and drop the alpine leg deliberately**, documenting that the base is now pinned and `verify-image-native-libs.py` is what guards against regressing to musl. Smallest, and honest — the smoke main then proves "the serving path works on the base we ship", which is the claim that still matters.
2. **Drop `ARG`, keep the alpine leg** by building it from a small dedicated Dockerfile the workflow owns, so the production file keeps its single hardcoded base. Preserves both directions; costs a second file that must not drift.

I would take (1): the alpine leg's job — catching a musl regression — now belongs to the verifier, which does it on every push rather than in one workflow.

Branch left untouched at `f0c922a55`. Happy to implement either once you pick.

JiRaska:
Rebased onto current `main` and took option **(1)** from the analysis above — the alpine negative control is gone, and the workflow now smokes exactly one image: the one the fleet ships.

**Why the choice was forced rather than preferred.** The rebase brought in #3629's `verify-image-native-libs.py`, whose header says outright that it reads the base from `Dockerfile.deploy` and that the script must **never** be given its own copy of it. This PR's `ARG BASE_IMAGE` *was* a second copy. Removing the `ARG` while keeping the alpine leg would have been the worst of the three options available: Docker accepts `--build-arg` for an undeclared `ARG` with nothing but a warning, so both legs would have built the **same** glibc base, and the leg whose entire job is to assert "alpine must fail" would have gone green while testing the image that works. A negative control that cannot fail is worse than none, because it reports coverage it does not have.

Also dropped the `GLIBC_BASE` env, which is a separate latent bug this PR would have shipped: it pinned `ubi10-minimal`, the candidate **not** adopted — #3709 was closed in favour of `eclipse-temurin:25-jre`. The gate would have been proving things about an image nobody deploys. The base is now read out of `Dockerfile.deploy` at run time:

```
base="$(grep -m1 '^FROM ' .github/workflows/Dockerfile.deploy | awk '{print $2}')"
```

which resolves to `eclipse-temurin:25-jre@sha256:681c543d…` today and follows the pin automatically.

**Coverage after the change** — the two checks are complementary, not overlapping:

| claim | who proves it | when |
|---|---|---|
| every bundled `.so` **links** in the pinned base | `verify-image-native-libs.py` (`ldd` inside the base) | every push |
| the serving path **works** — `OrtEnvironment` initialises, model + ADR-0141 card load via the same companion functions the bean uses | this workflow | this workflow's schedule |

`ldd` succeeding says nothing about `OrtEnvironment` initialising or the model parsing, so nothing that the alpine leg uniquely caught has been lost — "don't regress to musl" simply moved to the check that runs more often.

`Dockerfile.deploy` is now byte-identical to `main`; the portable `if/else` this PR carried was only ever scaffolding for a base that had not been picked. Diff is one file.

Merge is yours — `.github/workflows` is gated. Refs #3354, #3618, #3629, #3709.

JiRaska:
Read the verification claim as a claim, not a given — the interesting part is that it is a real repro, not a hypothetical.

```
alpine image        -> SMOKE FAIL: OrtEnvironment did not load
ubi10-minimal image  -> SMOKE OK
```

That is the shape this repo's CLAUDE.md already documents as a real incident class (`OnnxFraudModel` field-initializer `Error`, not `Exception`, from a native library — every endpoint sharing that bean's injection graph goes 500, including read-only ones that never touch the model). Building the same `Dockerfile.deploy` twice with a different `BASE_IMAGE` and asserting the smoke result differs is the right falsification: it proves the defect is the base image, not something else that happened to change alongside it.

Widening `MODEL_RESOURCE_PATH`/`MODEL_CARD_PATH` to `internal` so the smoke reuses the bean's own load path (rather than a parallel copy of the loading logic) is worth calling out — a smoke test with its own re-implementation of "how to load the model" can pass while the real bean's path is still broken.

Correctly scoped as step 1 of 2 — proving the defect before the fleet-wide base swap, rather than bundling both.

Not merging — workflow + Dockerfile.deploy change, human review by rule.

PR #3647, opened 2026-08-03
feat(lending): register workflow liveness on the accrual and provisioning schedulers (ADR-0237)

JiRaska:
## Summary

Register ADR-0160 mechanism 3 workflow liveness on the two money-path schedulers in `openbank-lending-service`. Before this change, a scheduler that silently stopped running left no runtime signal — the liveness gap identified in ADR-0237 and tracked in #3345.

## Type of change

- [x] feat (new functionality)

## Linked issues

Refs #3345

## Changes

- `InterestAccrualScheduler`: inject `DomainMetrics`, register `lending-interest-accrual` liveness (expected interval 24 h) at `StartupEvent`, call `recordSuccess()` in the `.invoke {}` success path; zero-work runs record success.
- `ProvisioningCycleScheduler`: inject `DomainMetrics`, register `lending-provisioning-cycle` liveness (expected interval 720 h) at `StartupEvent`, call `recordSuccess()` in the `.invoke {}` success path.
- Remove both files from `BASELINE` in `.github/scripts/check-scheduler-liveness.py` — the advisory gate (#3636) fails if healed entries remain listed.
- Extend `InterestAccrualSchedulerTest` and `ProvisioningCycleSchedulerTest` with the new `domainMetrics` constructor param (relaxed mock; existing test behaviour unchanged).
- Add `LendingWorkflowLivenessTest`: 5 unit tests asserting gauge registration at startup, expected-interval gauge, age collapse on success, zero-work success, and age staying stale on a failed run.

## Definition of Done

- [x] Hexagonal layering preserved (domain has zero framework imports).
- [x] Unit tests added/updated.
- [x] Integration tests added/updated (if service boundary involved).
- [x] Contract tests updated (if public API changed).
- [x] `./gradlew detekt ktlintCheck koverVerify build` passes.
- [x] No new lint warnings.
- [x] OpenAPI spec regenerated (if REST API changed).
- [x] Flyway migration added + rollback note (if DB changed).
- [x] Avro schema versioned backward-compatibly (if event changed).
- [x] Docs updated (README, ADR if architectural).

## Security checklist

- [x] No secrets, tokens, passwords, or PII in code, config, tests, logs.
- [x] No new `@SuppressWarnings`, `as Any`, `@Suppress("...")` without justification.
- [x] No new third-party dependency.
- [x] Auth / crypto / payment code changed? No.

## Compliance impact

- [x] None

## Rollout / Rollback

- Deployment strategy: rolling
[...]

PR #3703, opened 2026-08-04
feat(balance): register workflow liveness on the reconciliation schedulers (ADR-0237)

JiRaska:
Refs #3345, ADR-0237


JiRaska:
**Verdict: clean, mergeable individually, and correctly composes with #3647 on the shared baseline.** 0 blocking findings.

- Confirmed against the live API: `mergeable: MERGEABLE` (`BLOCKED` only on required-check/review status, not conflicts). Reproduced merging this branch's `.github/scripts/check-scheduler-liveness.py` change together with #3647's onto a fresh `origin/main` worktree — clean auto-merge, no conflict markers, and the resulting `BASELINE` correctly drops both this PR's two entries (`BalanceReconciliationScheduler.kt`, `ReconciliationFreshnessWatchdog.kt`) and #3647's two (lending) while leaving interest's entries (owned by #3705, not in this merge) untouched. So #3703 and #3647 are mutually consistent — no drop, no double-removal.
- `ReconciliationFreshnessWatchdog.checkFreshness()` was already a `suspend fun` before this PR (unchanged) and `BalanceReconciliationScheduler`'s tick stays Mutiny-reactive — neither trips the `@Scheduled` + `runBlocking`/no-Vert.x-context trap this repo has been bitten by repeatedly.
- New `@QuarkusTest`-adjacent liveness tests use `runBlocking` correctly typed as `(): Unit = runBlocking { ... }`, avoiding the silent-JUnit5-drop shape from an inferred non-`Unit` return.

Verified clean: `onStart(@Observes StartupEvent)` is the registration hook for both schedulers (not a lazy `init {}`), consistent with #3705/#3647's pattern in this same rollout.

PR #3705, opened 2026-08-04
feat(interest): register workflow liveness on accrual and capitalization schedulers (ADR-0237)

JiRaska:
Refs #3345, ADR-0237


JiRaska:
**Verdict: correct instrumentation, but this PR is currently stale against `main` — needs a rebase.** 1 blocking finding.

- Confirmed against the live API: `mergeStateStatus: DIRTY`, `mergeable: CONFLICTING`. Traced why: `.github/scripts/check-scheduler-liveness.py`'s `BASELINE` list on this branch predates two entries that already exist on current `main` (`openbank-domestic-payment/...ScreeningRedriveScheduler.kt`, `openbank-ledger-service/...JournalPartitionMaintainer.kt` / `TieOutFreshnessWatchdog.kt`) — this branch was cut before whatever PR added those. Reproduced the merge locally (`git merge origin/feat/interest-workflow-liveness` onto `origin/main` after `origin/feat/balance-workflow-liveness`) and it produces a real, loud conflict in `check-scheduler-liveness.py` rather than a silent drop — so nobody would merge it blindly, but it does need a rebase before it's mergeable at all, independent of #3703/#3647.
- Change itself is sound: both scheduler methods stay Mutiny-reactive (`.onItem().invoke { ... liveness?.recordSuccess() }`), not `runBlocking`, so the documented `@Scheduled` + `runBlocking` HR000068 trap doesn't apply here — this is purely additive instrumentation on the existing reactive pipeline, not a control-flow rewrite.

Verified clean: `onStart(@Observes StartupEvent)` registers the liveness gauge before the first scheduled tick, matching the `@ApplicationScoped`-is-lazy lesson (an `init {}` block would never fire; `@Observes StartupEvent` is the correct hook, not a constructor-time guard).

Issue #3630, opened 2026-08-03
admin-ui cluster dossier: the image-facts fallback (and the committed cluster-topology.json) still describe a multi-stage alpine build removed in #3016


JiRaska:
## What

`openbank-admin-ui/scripts/generate-cluster-topology.mjs` describes the service container image
for the `/docs/cluster` dossier (ADR-0081). Two things in it are stale, in the same way and for
the same reason — both describe the multi-stage in-Docker Gradle build that #3016 removed.

**1. The fallback literals.** `imageFacts()` parses a representative per-service Dockerfile
(`openbank-ledger-service/Dockerfile`, then account, then party). When it cannot parse one, the
output falls back to hardcoded values:

```js
multiStage: img.ok ? img.multiStage : true,
buildBase:  img.ok ? img.buildBase  : 'eclipse-temurin:25-jdk-alpine',
runtimeBase: img.ok ? img.runtimeBase : 'eclipse-temurin:25-jre-alpine',
```

There has been no build stage since #3016 and no alpine base since #3618, so every one of those
three values is now false. A fallback that invents a plausible answer is worse than one that says
it does not know — it renders as fact in the dossier.

**2. The committed `cluster-topology.json` is stale in exactly that shape.** It currently holds
`multiStage: true` and `buildBase: eclipse-temurin:25-jdk-alpine@sha256:5ecfde8e…`, which the
generator has not produced since #3016. It is regenerated on every admin-ui build and baked into
the image, and #2621 deliberately stopped deploy PRs from committing it (the `generatedAt` line
made every concurrent deploy PR conflict), so the committed copy is only the build-context
fallback — but it is what a reader of the repo sees.

## Why it is not covered by #3629

#3629 adds rule 3 to `check-dockerfile-no-build-stage.py`: every declared runtime base must equal
`.github/workflows/Dockerfile.deploy`'s. That keeps the *parsed* path honest, and it is why the
`FROM` was kept in those files rather than deleted — with nothing to parse, `img.ok` goes false
and the fallback above is what renders. So the gate makes the fallback unreachable in practice
without making it correct.

## Suggested fix

Make the fallback state ignorance instead of a guess — e.g. `runtimeBase: null` with the dossier
row rendering as `unknown` / `partial` — so a parse failure reads as a parse failure. Whatever
shape it takes, the invariant worth keeping is the one #3629 argues for: the runtime base is
stated in exactly one place (`Dockerfile.deploy`), and every other mention derives from it.

[...]

PR #3709, opened 2026-08-04
build(infra): move fleet runtime base from alpine/musl to ubi10-minimal glibc

### LINKED ISSUE #3354: fraud-service ONNX has never loaded: glibc-linked native library in a musl (alpine) runtime image

JiRaska:
Closes #3354

Second step of the ONNX alpine/musl fix (see #3700 for the first step, which added the in-image smoke gate and proved the failure on alpine).

This PR moves the whole fleet runtime from the musl-based `eclipse-temurin:25-jre-alpine` image to the glibc-based `eclipse-temurin:25-jre-ubi10-minimal` image:

- every per-service `Dockerfile` (53 files)
- `Dockerfile.scanner` and `openbank-security-scanner/Dockerfile.prebuilt`
- `openbank-notification-service/Dockerfile.runtime`
- local multi-stage templates in `openbank-infra/docker/services/` (`standalone-service.Dockerfile`, deprecated `root-service.Dockerfile`)
- `.github/workflows/Dockerfile.deploy` default base image
- user creation switched from alpine `addgroup/adduser` to RHEL `groupadd/useradd`
- `.github/workflows/Dockerfile.deploy` keeps conditional user creation so it still boots on either base during any transition window
- `openbank-admin-ui/scripts/generate-cluster-topology.mjs` fallbacks and `openbank-admin-ui/cluster-topology.json` regenerated to match the new base
- root `CLAUDE.md` note updated so it no longer points at the old alpine literal

`check-dockerfile-no-build-stage.py` passes locally (53 runtime-only Dockerfiles, all with EXPOSE). A manual `docker run` of the new ubi10-minimal base confirms `groupadd`/`useradd` are present and work.

### DIFF
--- a/.github/workflows/Dockerfile.deploy
+++ b/.github/workflows/Dockerfile.deploy
@@ -1,6 +1,17 @@
-FROM eclipse-temurin:25-jre-alpine@sha256:28db6fdf60e38945e43d840c0333aeaec66c15943070104f7586fd3c9d1665b0
+ARG BASE_IMAGE=eclipse-temurin:25-jre-ubi10-minimal@sha256:6031e5480f5d5818252a348980a192af0eab09e3380a55f75927e019f4d8136c
+FROM ${BASE_IMAGE}
 WORKDIR /app
-RUN addgroup -S openbank && adduser -S openbank -G openbank
+
+# The deploy pipeline historically used an alpine base (musl). Issue #3354 moves the
+# fleet to a glibc runtime, so this Dockerfile must accept either base without a hard
+# rewrite. Alpine ships addgroup/adduser; RHEL-based bases (ubi-minimal) ship
+# groupadd/useradd via shadow-utils. Pick whichever is present and always drop to a
+# non-root user — the smoke gate (onnx-serving-smoke.yml) exercises the same path.
+RUN if command -v addgroup >/dev/null 2>&1 && command -v adduser >/dev/null 2>&1; then \
+        addgroup -S openbank && adduser -S openbank -G openbank; \
+    else \
+        groupadd -r openbank && useradd -r -g openbank openbank; \
+    fi
 USER openbank
 COPY quarkus-app/lib/ /app/lib/
 COPY quarkus-app/*.jar /app/
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -569,8 +569,8 @@ fire from *outside* it, so they stay here:
   `openbank-admin-ui/scripts/generate-cluster-topology.mjs` parses
   `openbank-ledger-service/Dockerfile` in `imageFacts()` and renders the base image into the
   /docs/cluster dossier (ADR-0081) — and when it cannot parse one it falls back to a HARDCODED
-  `eclipse-temurin:25-jre-alpine` literal, so removing the declaration would have resurrected the
-  fiction in the UI instead of retiring it (#3354, #3630). A second reader that is a *generator
+  `eclipse-temurin:25-jre-ubi10-minimal` literal, so removing the declaration would have resurrected
+  the fiction in the UI instead of retiring it (#3354, #3630). A second reader that is a *generator
   in another tree* is invisible from the file, from the deploy pipeline, and from the gate that
   owns the file's shape.
 - **An oversized `run:` script makes the WHOLE workflow unparseable — and GitHub says nothing.**
--- a/Dockerfile.scanner
+++ b/Dockerfile.scanner
@@ -1,6 +1,6 @@
-FROM eclipse-temurin:25-jre-alpine@sha256:28db6fdf60e38945e43d840c0333aeaec66c15943070104f7586fd3c9d1665b0
+FROM eclipse-temurin:25-jre-ubi10-minimal@sha256:6031e5480f5d5818252a348980a192af0eab09e3380a55f75927e019f4d8136c
 WORKDIR /app
-RUN addgroup -S openbank && adduser -S openbank -G openbank
+RUN groupadd -r openbank && useradd -r -g openbank openbank
 USER openbank
 COPY openbank-security-scanner/build/openbank-security-scanner-0.1.0-SNAPSHOT-runner.jar /app/quarkus-run.jar
 EXPOSE 8120
--- a/openbank-account-service/Dockerfile
+++ b/openbank-account-service/Dockerfile
@@ -15,9 +15,9 @@
 # Adding a build stage back here is enforced against by
 # .github/scripts/check-dockerfile-no-build-stage.py.
 
-FROM eclipse-temurin:25-jre-alpine@sha256:28db6fdf60e38945e43d840c0333aeaec66c15943070104f7586fd3c9d1665b0
+FROM eclipse-temurin:25-jre-ubi10-minimal@sha256:6031e5480f5d5818252a348980a192af0eab09e3380a55f75927e019f4d8136c
 WORKDIR /app
-RUN addgroup -S openbank && adduser -S openbank -G openbank
+RUN groupadd -r openbank && useradd -r -g openbank openbank
 USER openbank
 COPY quarkus-app/lib/ /app/lib/
 COPY quarkus-app/*.jar /app/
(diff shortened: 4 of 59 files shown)


JiRaska:
**Verdict: sound, well-motivated (fixes a real ONNX/musl incompatibility per #3700), one doc-accuracy nit.** 0 blocking findings.

- The user-creation `RUN` step correctly branches per distro family (`Dockerfile.deploy` detects `addgroup`/`adduser` vs falls back to `groupadd`/`useradd`; every per-service Dockerfile with a literal `FROM` is switched directly to the RHEL-style `groupadd -r openbank && useradd -r -g openbank openbank`) — consistent across all 53 files, spot-checked several.
- `openbank-admin-ui/cluster-topology.json` was correctly regenerated (new `generatedAt`, `buildBase`/`runtimeBase` and both `detail` strings updated to the ubi10-minimal digest) — this is exactly the reader called out by this repo's own lesson about `generate-cluster-topology.mjs`'s `imageFacts()` parsing a representative Dockerfile, and it was accounted for rather than left to resurrect a stale fiction.
- **Doc-accuracy nit**: the PR body says `generate-cluster-topology.mjs fallbacks ... updated to match the new base`, but the diff never touches that file. Checked the current script on `main`: `imageFacts()` no longer has a hardcoded `eclipse-temurin:...` literal fallback at all — it returns `{ ok: false }` and downstream renders `"nezjištěno / neparsováno"` when it can't parse, so the hardcoded-literal defect this repo's `CLAUDE.md` bullet describes was already fixed (presumably by #3630) before this PR. This PR's edit to that `CLAUDE.md` bullet just swaps the cited literal string (`alpine` → `ubi10-minimal`) rather than updating the description to reflect that the hardcoded fallback no longer exists — so the bullet keeps describing a defect class that's no longer live in code. Not a functional risk (the JSON was correctly regenerated regardless), just worth a follow-up doc correction so the bullet doesn't mislead the next reader into thinking there's still a hardcoded fallback to worry about.
- No unmeasured size/CVE claims in the PR body to flag — it correctly limits itself to the musl/glibc functional justification (ONNX) and a manual `docker run` confirmation that `groupadd`/`useradd` work on the new base, without asserting an image-size or CVE-posture improvement it hasn't measured.

Verified clean: `check-dockerfile-no-build-stage.py`'s constraint (runtime-only Dockerfiles, `EXPOSE` present) is unaffected — no build stage was added back in any of the 53 files; the digest pin (`@sha256:...`) is consistent across every `FROM` change, so this isn't a floating-tag regression.

JiRaska:
**Follow-up finding, same review pass**: this PR's base-image choice (`eclipse-temurin:25-jre-ubi10-minimal`) races an independent, currently-open competing fix for the identical problem: **#3618** (`fix(fraud): move the runtime image to glibc so the ONNX model can actually load`) retargets the same `Dockerfile.deploy` + all 53 per-service Dockerfiles to a *different* glibc base, `eclipse-temurin:25-jre` (full Debian-based, not ubi10-minimal), and **#3629** in this same batch is stacked on top of #3618 with a gate (`dockerfile-runtime-only`, rule 3) that enforces every Dockerfile's `FROM` matches `Dockerfile.deploy` byte-for-byte — which would then enforce the *other* base image fleet-wide. Both chains solve #3354; only one base image should win. Left as-is, whichever of {#3709, #3629} merges second either hard-conflicts or (worse, per this repo's own documented "clean merge that silently drops content" failure mode) quietly overwrites the other's choice with no diff to notice. Worth resolving which base is authoritative (ubi10-minimal vs plain jre) and closing/rebasing the other chain before either merges.

JiRaska:
Closing: the fleet base decision went the other way, and it has landed.

`#3618` merged at 09:42 and `main` now carries

```
FROM eclipse-temurin:25-jre@sha256:681c543d6f36c50f45e9b5226930a46203dcfa351d3670e9d0bdf0dabae53539
```

with `#3629` retiring alpine from the 53 declaring files behind it. Both destinations were glibc, so this PR's diagnosis was right and would equally have fixed the ONNX load — `libonnxruntime.so` needs glibc + libstdc++ and the fleet was on musl. The choice between `25-jre` and `25-jre-ubi10-minimal` was a preference, not a correctness question, and it is now settled on `main`.

Closing explicitly rather than leaving this open, because a stale base-image PR is a live hazard here: rebasing it later would rewrite `FROM` across the fleet to a *different* valid image, and nothing in CI would object — both bases build, both run, both are glibc. Merge order would silently decide the fleet's runtime.

**Two things this PR covered that the winning pair does not**, worth carrying forward rather than losing with it:

- `Dockerfile.scanner`, `openbank-security-scanner/Dockerfile.prebuilt`, `openbank-notification-service/Dockerfile.runtime`
- the local multi-stage templates under `openbank-infra/docker/services/` (`standalone-service.Dockerfile`, the deprecated `root-service.Dockerfile`)

Those build paths stay on alpine after #3618/#3629. Nothing in them bundles a native library today, so nothing is broken — but the next one that does will fail exactly the way fraud-service did, and the failure mode is a bean whose construction throws an `Error`, which surfaces as unrelated endpoints 500-ing (#3376). Worth a follow-up issue rather than a rediscovery.

Also relevant: **#3700** (the in-image ONNX smoke gate) remains the thing that makes any of this falsifiable — it runs the real dynamic loader over the real `.so` inside the real base image before a push. It is being rebased onto the new base now.

PR #3711, opened 2026-08-04
feat(copilot): ADR-0238 T1 — durable Postgres conversation history

### LINKED ISSUE #3710: copilot T1 conversation history: durable Postgres store (ADR-0238 D1)
ADR-0238 T1: add a PostgresConversationStore that replaces Redis for conversation history — durable across sessions, resumable across devices, 90-day rolling TTL, erased at PARTY_ERASED.

**Scope:**
- New Postgres DB for copilot-service (CNPG, follow ADR-0009)
- Flyway V1: conversation_history table
- PostgresConversationStore implementing the existing ConversationStore port
- copilot.conversation-store config to select (memory/redis/postgres)
- %test profile uses memory store (no DB in unit tests)
- PostgresConversationStoreIT with Testcontainers
- gitops: postgres.yaml + ESO + updated Deployment env
- Threat model update (new PII store)
- version 0.8.1 -> 0.9.0

Refs ADR-0238 T1.

JiRaska:
## Summary

ADR-0238 T1 (#3710): adds a **durable Postgres conversation store** that replaces the ephemeral Redis session for conversation history. History now survives pod restarts, is resumable across devices, and carries a 90-day rolling TTL. Wire control: `copilot.conversation-store=postgres` (default; `redis` and `memory` kept for rollback / dev).

## Type of change

- [x] feat (new capability: durable conversation history)

## Linked issues

Closes #3710

## Changes

- **Flyway V1** — `conversation_history` table (`customer_id`, `conversation_id`, `messages_json`, `expires_at`); `UNIQUE (customer_id, conversation_id)` enforces isolation; expiry + lookup indexes.
- **`PostgresConversationStore`** — `load()` with `WHERE expires_at > now()` + `append()` upsert; uses the `vtx` Vert.x bridge (same pattern as `CampaignJourneyActivitiesImpl`) since the ConversationStore interface is synchronous and must bridge to Hibernate Reactive.
- **Wire guard change** — `InMemoryConversationStore` and `RedisConversationStore` now use `@IfBuildProperty(name="copilot.conversation-store", ...)` instead of `copilot.token-store`, so conversation-store selection is independent of token-store selection.
- **`%test` profile** — DevServices auto-starts a Postgres container; `PostgresConversationStoreIT` uses its own Testcontainers container for full isolation (4 tests: empty load, append + read, trim to MAX_MESSAGES, NO_MEMORY_ID never persisted).
- **gitops** — `copilot/postgres.yaml` (CNPG cluster) + `copilot-service.yaml` Deployment env (`REACTIVE_URL`, `JDBC_URL`, `POSTGRES_USER/PASSWORD` from CNPG-generated `copilot-db-app` secret).
- **Threat model** — 2026-08-04 change-log entry (new PII store, STRIDE analysis, open items: retry queue + `PARTY_ERASED` consumer + Art. 17 erasure).
- **version 0.8.1 → 0.9.0** (feat minor).

## Definition of Done

- [x] Hexagonal layering preserved (ConversationStore port unchanged; new adapter in infrastructure).
- [x] 4 new IT tests (Testcontainers Postgres) + all 116 existing tests pass.
- [x] `./gradlew :openbank-copilot-service:ktlintCheck :detekt :test :koverVerify` — BUILD SUCCESSFUL.
- [x] Flyway V1 + rollback note. Threat model updated. No openapi change.

## Security checklist

- [x] Conversation messages are personal data — isolation by `(customerId, conversationId)` UNIQUE key; guessed ids can never cross customer boundary.
- [x] No secrets in git — DB credentials via CNPG-generated secret only.
- [x] `security-review-required` — new PII store (thread model updated, §6 added).

## Compliance impact

- [x] GDPR — new personal data store; purpose-limited to conversation UX; erasure hook (`PARTY_ERASED` consumer) and retry queue are open items (noted in threat model, #3710 for follow-up).
- [x] ADR-0175 residency: CNPG cluster stays in the same region as all other datastores.


JiRaska:
## Gate fix + trap audit (#3710)

Merged current `main` in and pushed `dc6dfff3a`. Of the five reds, **only one was real.**

### The reds

| Check | Verdict |
|---|---|
| `gates (gitops)` / `Validate manifests` | **REAL.** `db-backup association gate` — fixed below. (`Validate manifests` is an aggregator over the gate shards, so it was the same root cause, not a second one.) |
| `Trivy` | Stale. Attempt 1 failed at 08:07; the **re-run at 08:08 succeeded**. Findings were `brace-expansion` (npm transitive) and `AWS-0031` (ECR mutable tags) — this PR touches no npm manifest and no Terraform module. |
| `Customer-app dossier` | Pre-existing fleet drift, **already fixed on `main`** by #3810 at 06:18 today. Cleared by the merge. |
| `dependency-review` | Stale. It read the head before the dependency submission landed; the submission for that sha exists and succeeded on `run_attempt 2`. Issue #2833 is closed. |

### What was actually broken

`copilot-db` declared a `barmanObjectStore` into the Terraform-managed backup bucket but had **no `aws_eks_pod_identity_association`**. Every WAL archive would have failed with `Unable to locate credentials` and the cluster would have had no recovery point — the #1444 failure mode, caught this time before the cluster exists rather than days after. `postgres.yaml` even pointed at `db-backups.tf` in a comment; the entry was never added.

Falsified rather than assumed: the gate FAILs with my line removed and PASSes with it (55 → 56 associations). Full `gitops` shard: `PASS=22 WARN=1`, no FAIL.

> Note for whoever maintains the gate: its suggested-fix string emits the **namespace** as the map key (`platform = {...}`), not the service name. Every existing entry is keyed by service, and `platform` already hosts `agent` and `mcp`, so following the suggestion literally would collide on the second one.

### Audit — traps this change was exposed to

Three were already handled correctly by the author:

- **Panache `persist()` on an application-assigned `@Id`** — handled. `append` uses `Panache.getSession().flatMap { s -> s.merge(upsert) }`. The `subsequent appends accumulate` IT does 12 appends, so it genuinely exercises the update path and would go red against `persist()`.
- **`quarkus.flyway.migrate-at-start`** — set `true`, with a literal `quarkus.datasource.jdbc.url` in the same file. Both halves of the gate satisfied.
- **`VertxContextSupport.subscribeAndAwait`** — normally a footgun (it throws on an event loop), but safe here: `chat` is `runBlocking` on a worker thread and `chatStream` carries `.runSubscriptionOn(Infrastructure.getDefaultWorkerPool())`. **This is load-bearing and undocumented at the call site** — if either endpoint ever becomes `@NonBlocking`, the store breaks at runtime with nothing catching it.

Flyway numbering: `V1` is free — the migration directory does not exist on `main`, and #3711 is the only open PR adding one. Re-checked against live `main` immediately before pushing.

### Retention — a real gap, reported not silently fixed

Conversation history is customer chat transcripts, i.e. a PII store. The issue asks for a 90-day rolling TTL and `PARTY_ERASED` erasure. As it stands:

- **Retention is read-side only.** `load` filters `expires_at > now()`, so an expired conversation stops being *readable* — but nothing DELETEs the row. The message text stays on disk, and in every base backup, indefinitely.
- **There is no sweep.** No `@Scheduled`, no Temporal workflow, nothing. The migration comment claimed one existed ("the cleanup rides the existing Temporal scheduler pattern or a lightweight Quarkus `@Scheduled` job") — that was aspirational, and I have rewritten it to say plainly what is and is not implemented.
- **There is no `PARTY_ERASED` consumer** anywhere in the service.

I did **not** add these — that is new scope and needs a decision, not a drive-by. But the 90-day TTL should not be read as an erasure guarantee until the sweep lands, and I would not call ADR-0238 D1 done without it.

Also corrected: `postgres.yaml` carried campaign-service copy-paste describing a `send_log` and a "marketing audit record" for a database that holds chat transcripts.

### What I did NOT verify

- **I never ran the module build or the IT.** No `ktlintCheck`/`detekt`/`test`. My own diff is `.tf`/`.sql`/`.yaml` only and cannot affect Kotlin compilation, but that means **I have not independently confirmed `PostgresConversationStoreIT` passes** — I only read it. It needs Docker/Testcontainers.
- I did not confirm the IT actually runs in CI. There is no include/exclude filter in `build.gradle.kts`, so Gradle's default detection should pick up an `*IT` class — but I found no Services CI job for copilot at this head to prove it executed. If it does not run, the `merge`-vs-`persist` coverage above is theoretical.
- I did not test the `%test` profile claim from the issue ("uses memory store, no DB in unit tests"). `copilot.conversation-store` stays `postgres` under `%test` and the profile enables devservices, so unit tests appear to need a database. That may be deliberate; I did not change it.
- The `@IfBuildProperty` guards look mutually exclusive (`memory`/`redis` exact-match, `postgres` with `enableIfMissing = true`), but I did not build the service to confirm CDI resolves exactly one `ConversationStore`.
- The Terraform change is unapplied and unplanned — no `tofu plan`, no cluster access. Its correctness is asserted by the gate only.
- `loki-rule-load-test` reports UNFALSIFIED locally (0.0s, missing binary). Unrelated to this PR, and it passes in CI.
- The cross-PR sampling behind the "pre-existing" verdicts is thin for `dependency-review` specifically: no other sampled PR touches a dependency manifest, so their green is vacuous. Re-running that check is the real test.

JiRaska:
Independently checked the retention concern before endorsing the decision not to arm auto-merge, and
it holds — with a sharper framing than "no erasure consumer".

**Seven services already consume `PARTY_ERASED`:**

```
openbank-account-service   openbank-aml-service        openbank-card-issuance-service
openbank-kyc-service       openbank-notification-service  openbank-onboarding-service
openbank-party-service
```

19 Kotlin files in total. So this is not a control the repo has documented and never built — it is
built, and **this new store would be the outlier that does not participate in it**. That is a
materially different (and easier to act on) statement than "nothing implements erasure", which is
what a narrower grep first suggested to me.

The PR is honest about it in the migration itself, which is the right place for it:

```sql
-- `expires_at > now()`, so an expired conversation stops being readable — but nothing DELETES
-- ... Do not read the 90-day TTL as an erasure [guarantee]
```

And the port makes it structural rather than an oversight — `ConversationStore` declares only:

```kotlin
fun load(customerId: String, conversationId: String): List<ChatMessage>
fun append(customerId: String, conversationId: String, newTurns: List<ChatMessage>)
```

There is no delete operation to call, so honouring an erasure would need a signature change, not just
a subscriber. Worth naming because it decides the shape of the follow-up.

Two things this changes about the merge decision, both for the owner rather than for CI:

1. **Rows become unreadable at 90 days but are never removed.** A `TIMESTAMPTZ` filter is a read-side
   control; the data stays. For chat transcripts — free-text customer input, which is exactly where
   unpredictable personal data lands — the difference between "not served" and "not held" is the
   whole question.
2. **Erasure is already an implemented fleet convention**, so adding a store outside it widens an
   existing control's blind spot rather than deferring an unbuilt one.

Agreeing with the author: this is a retention decision, not something to merge on green checks. The
checks being green is now true and beside the point.

PR #3718, opened 2026-08-04
ci(auto-deploy): trigger deploy on openbank-*/version.txt changes

JiRaska:
ci(auto-deploy): trigger deploy on openbank-*/version.txt changes

Release-please bumps a service's `version.txt`, but a release-only commit does not match the `openbank-*/src/main/**` path filter, so the new version had no image built and immediately drifted from what is deployed.

This adds `openbank-*/version.txt` to the auto-deploy push trigger paths so a version bump is treated as a release marker and drives a deploy just like a source change.

This prevents future occurrences of the drift pattern seen in:
- #3597 (openbank-onboarding-service)
- #3363, #3364, #3365, #3366, #3367 (the broader deploy-drift series)

Existing drift on those services is not retroactively deployed by this PR; a one-time manual `workflow_dispatch` of `auto-deploy.yml` for the affected services can close the current gap after this merges.

Closes #3597
Refs #3363 #3364 #3365 #3366 #3367

PR #3726, opened 2026-08-04
chore(ci): surface fallback review error detail (#3488)

### LINKED ISSUE #3488: agent-review fails ~40% of runs with num_turns=1, cost=$0 — the action never runs, and it blocks --auto merges

JiRaska:
Fixes #3488.

The fallback reviewer guard in `.github/scripts/check-claude-fallback-result.sh` already catches the #2161 "died on first model call" shape, but it only prints `is_error`, `num_turns` and `cost`. When the action transcript also carries an `error`/`message` field (the gateway/OAuth rejection symptom from #3488), that detail is dropped, so the next occurrence still has to be debugged by enabling full output on a public repo.

Changes:
- Extract `.error // .message // .output` from the terminal `result` object and include it (truncated to 500 chars, newlines collapsed to spaces) in the GitHub `::error` annotation.
- Add a self-test case for an error-message variant and bump the self-test count.

The script's own `--self-test` passes.

### DIFF
--- a/.github/scripts/check-claude-fallback-result.sh
+++ b/.github/scripts/check-claude-fallback-result.sh
@@ -69,16 +69,24 @@ evaluate() {
     return 1
   fi
 
-  local is_error num_turns cost subtype
+  local is_error num_turns cost subtype error_detail
   is_error="$(printf '%s' "$result" | jq -r '.is_error // false')"
   num_turns="$(printf '%s' "$result" | jq -r '.num_turns // 0')"
   cost="$(printf '%s' "$result" | jq -r '.total_cost_usd // 0')"
   subtype="$(printf '%s' "$result" | jq -r '.subtype // "unknown"')"
+  error_detail="$(printf '%s' "$result" | jq -r '(.error // .message // .output // empty) | tostring | gsub("\\n"; " ") | .[0:500]')"
 
   echo "Claude fallback transcript: is_error=$is_error num_turns=$num_turns total_cost_usd=$cost subtype=$subtype"
 
   if [ "$is_error" = "true" ]; then
-    echo "::error title=Claude fallback failed::The transcript reports \`is_error: true\` (subtype=$subtype, num_turns=$num_turns, cost=\$$cost). The action exits 0 in this case, so without this check the step would read as a SUCCESSFUL review of a PR nobody reviewed. See #2161."
+    # #3488: the OAuth token was rejected at the gateway, but the only symptom was the
+    # num_turns/cost summary. Surface whatever error detail the transcript carries so the
+    # next failure can be diagnosed without enabling full output on a public repo.
+    if [ -n "$error_detail" ]; then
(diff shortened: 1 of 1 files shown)


JiRaska:

```

```

Issue #3734, opened 2026-08-05
Edge/shared service-account principals reach operator writes on 10 enforce=true services — fleet-wide instance of the interest #3698 class


JiRaska:
## What

`service-account-openbank-edge` (customer-reachable via the edge namespace) and `service-account-openbank-services` (shared backend client) both carry **ROLE_OPERATOR** in the realm and are classified `HUMAN` by `AuthorizeInterceptor`. Every `operator-<domain>-write` ext rule that is role-only therefore admits both to money-path writes once `AUTHZ_ENFORCE=true`.

This is the same class fixed for interest in PR #3698 (follow-up to #3695 / this issue's sibling #3679): interest's ext now excludes `service-account-*` from `operator-interest-write` AND carries a `prohibited` clause vetoing the write actions at the allow head. The fleet audit below shows interest was not unique — it was just the first one measured.

The `shared_m2m_write_prohibition` set (`rules.yaml`, GHSA-58jq-9hq3-66jr) only denies the **shared** client (`service-account-openbank-services`), and only for the six graduated reasons. It does nothing for the **edge** client anywhere.

## Audit (main, 2026-08-05)

`AUTHZ_ENFORCE=true` services with OPERATOR-matrix write grants whose write rules carry **no** `service-account-*` exclusion:

| Service | Write rule (shape) | Shared client | Edge client |
|---|---|---|---|
| account | `operator-account-write` — `startswith(input.action, "account.")`, role-only | ADMITTED | ADMITTED |
| balance | `operator-balance-write` — role-only action set | denied (graduated: `service-balance-m2m`) | **ADMITTED** |
| billing | `operator-billing-write` — role-only | ADMITTED (deferred: callers unverified) | ADMITTED |
| fraud | `operator-fraud-write` — `startswith(..., "fraud.")`, role-only | denied (graduated: `service-fraud-scoring`) | **ADMITTED** |
| fx | `operator-fx-write` / `operator-fx-trigger` — role-only | ADMITTED | ADMITTED |
| ledger | `operator-ledger-write` — role-only action set | denied (graduated: `service-ledger-post/-reverse`) | **ADMITTED** |
| lending | `operator-lending-write` — `startswith(..., "lending.")`, role-only | ADMITTED | ADMITTED |
| sca | `operator-sca-write` — `scaChallenge.`/`device.` families, role-only | denied (graduated) | **ADMITTED** |
[...]

PR #3735, opened 2026-08-05
feat(scheduler): register workflow liveness for five non-money-path jobs (ADR-0237)

JiRaska:
## Summary

Registers the ADR-0237 workflow-liveness heartbeat for five non-money-path scheduled jobs and shrinks the adoption baseline accordingly. This makes a silent non-running scheduler visible to the control-liveness sentinel instead of being discovered archaeologically.

## Type of change

- [x] feat (new functionality)
- [ ] fix (bug fix)
- [ ] refactor (no behavior change)
- [ ] perf (performance)
- [ ] docs
- [ ] chore (build / tooling)
- [ ] security (private until disclosed)
- [ ] breaking change

## Linked issues

Refs #3345
Refs #3343

## Changes

- `openbank-aml-service`: register `aml-party-resolution` liveness in `PartyResolutionScheduler` and record success after each completed sweep.
- `openbank-analytics-sink`: register `analytics-reconciliation` liveness in `ReconciliationJob` and record success after a reconciliation result is stored.
- `openbank-authz-policy-auditor`: register `authz-policy-check` liveness in `AuthzPolicyCheckScheduler` after a scheduled Temporal workflow is dispatched.
- `openbank-docs-truth-agent`: register `docs-truth-check` liveness in `DocsTruthCheckScheduler` after a scheduled Temporal workflow is dispatched.
- `openbank-finops-agent`: register `finops-analysis` liveness in `FinOpsAnalysisScheduler` after a scheduled Temporal workflow is dispatched.
- Add per-service `*WorkflowLivenessTest` coverage proving the gauge starts stale, collapses after a successful run, and stays stale on failure.
- Bump `version.txt` for the released services touched here (`aml`, `authz-policy-auditor`, `docs-truth-agent`, `finops-agent`). `analytics-sink` has no `version.txt` and remains outside release-please.
- Remove the five adopted files from `.github/scripts/check-scheduler-liveness.py` baseline.

## Definition of Done

- [x] Hexagonal layering preserved (domain has zero framework imports).
- [x] Unit tests added/updated.
- [ ] Integration tests added/updated (if service boundary involved).
- [ ] Contract tests updated (if public API changed).
- [x] `./gradlew detekt ktlintCheck koverVerify build` passes.
- [x] No new lint warnings.
- [ ] OpenAPI spec regenerated (if REST API changed).
- [ ] Flyway migration added + rollback note (if DB changed).
- [ ] Avro schema versioned backward-compatibly (if event changed).
[...]

PR #3740, opened 2026-08-05
fix(balance): prohibit customer-edge principal from balance writes

JiRaska:
## Summary

First PR from the #3734 fleet audit (graduated services batch). `operator-balance-write` was role-only, and `rules.yaml`'s `role_action_matrix` grants the balance writes to `ROLE_OPERATOR`. The customer-facing edge M2M identity (`service-account-openbank-edge`, HUMAN-classified, ROLE_OPERATOR) was therefore admitted — via base `matrix-allows` — to **the fleet's money-moving primitives** (`balance.credit`/`balance.debit`) even though no edge write path exists. Balance is reachable from the public edge since 2026-08-02 (its own threat-model entry), which makes this the highest-stakes row in the #3734 matrix.

## Type of change

## Linked issues

Refs #3734, #3698 (the interest pattern this follows)

## Changes

**Caller audit (the evidence this is safe):** customer-edge only ever GETs `/api/v1/balances/{accountId}` (three call sites in `CustomerEdgeResource.kt:264,1194,1222`); no `balanceServiceUrl` write call exists anywhere in the edge. No other M2M caller writes via the edge identity.

**Tightening (two layers, both needed):**
- `operator-balance-write` and `supervisor-overdraft-limit` now exclude every `service-account-*` principal (delegation/interest idiom).
- New edge-scoped `prohibited` clause vetoes all seven write actions (`hold`/`holdRelease`/`credit`/`debit`/`initialize`/`reconciliation.run`/`overdraftLimit`) at the allow head — this is the layer that beats the matrix grant, which the rule exclusion alone does not.
- **Deliberately edge-scoped, not interest's all-service-accounts shape:** the shared client IS a legitimate writer here and keeps its identity-scoped `service-balance-m2m` rule (verified by test).

**Shape:** heredoc → standalone `balance_rest_ext.rego` (interest/delegation pattern) so `opa test` can load it. New `balance_rest_ext_test.rego` (12 tests). Generator regenerated bundle + checksum annotation (balance only, no fleet ripple). Threat model updated (ADR-0030 D2).

## Verification evidence

- `opa test` trio: **12/12 PASS**.
- **Falsification:** stripping the exclusion + prohibition turns **7/12 tests red** (each edge-write denial, the prohibition probe, and the role-only-reason exclusion probe).
- Gates locally: security group 11/11 (incl. `threat-model-updated-on-trust-boundary-change`), gitops group green apart from known local artifacts (`loki-rule-load-test` UNFALSIFIED, `incluster-hostname-resolution` advisory WARN — both reproduce on unmodified main).
- Bundle regen run twice, idempotent; checksum `364f3cdd0cf87840`.

- [ ] Threat model change present (§7 change log, 2026-08-05 entry).
- [ ] 2 approvals.
- [ ] Confirm the caller audit: no legitimate edge write path is broken (none exists per the audit; shared-client M2M writes untouched and tested).

## Security checklist

- [x] No secrets / PII.
- [x] Auth change → `security-review-required`.
- [x] No policy logic removed; two overlapping fail-closed layers added. Shared-client write path preserved and pinned by test.

## Compliance impact

- [x] DORA — ICT access control narrowed on the fleet's central money primitive.


JiRaska:

```

```

JiRaska:
**Review: partial — the money-moving primitives are correctly handled, but `balance.reconciliation.run` remains reachable by the shared client via `matrix-allows`.**

**1. `openbank-infra/gitops/components/balances/balance_rest_ext.rego` — the veto is pinned to `service-account-openbank-edge`.**

`rules-opa-data.yaml:252-257` grants ROLE_OPERATOR `balance.credit`, `debit`, `hold`, `holdRelease`, `initialize`, `reconciliation.run`. For `principal.id = service-account-openbank-services`:
- `credit` / `debit` / `hold` / `holdRelease` / `initialize` — admitted by `service-balance-m2m`. **Intended**, correct, and the edge-only veto is the right instrument to avoid 403'ing it.
- `balance.reconciliation.run` — the **only** matrix-granted balance write *not* in `service-balance-m2m`. `operator-balance-write` is now correctly excluded ✅, but `matrix-allows` (`rest.rego:151-155`) fires, the new `prohibited` requires the edge id and does not, and the base `prohibited` (`rest.rego:485-491`) cannot (fact A below; `operator-balance-write` is registered in `shared_m2m_write_prohibition.reasons` and it still would not help — `matrix-allows` is not a listed reason, so the `every` clause is false). Net: **allowed.**

So any backend service on the shared client can kick off an end-of-day reconciliation run. Lower stakes than credit/debit, but it is a scheduled control-run with real reporting consequences, it is not in the identity-scoped grant, and it is the one action here whose exposure the PR does not account for. `balance.overdraftLimit` is correctly closed — it is matrix-granted only to ROLE_SUPERVISOR, which no service account holds, and `supervisor-overdraft-limit` now excludes service accounts.

**2. `balance_rest_ext_test.rego` — `test_shared_m2m_no_longer_admitted_via_role_only_reason` is the #3764 pattern:** reason-absent, on `balance.credit` — an action the shared client must keep, so it can never be the strong form and proves nothing about the residue. `balance.reconciliation.run` has no shared-principal test in either form; `not rest.allow with input as {"principal": services_m2m, "action": "balance.reconciliation.run"}` is red today. Note the in-file comment already anticipates this shape ("the shared client can ...") — the missing piece is the assertion, not the awareness.

Fix: add `balance.reconciliation.run` to a `prohibited` clause keyed on `startswith(input.principal.id, "service-account-")` (no legitimate M2M caller is claimed for it), keeping the existing edge-scoped clause for the rest.

**Verified clean:** all five edge denials use the strong `rest.allow == false` form; `test_prohibition_fires_for_edge_on_write` proves the veto rather than merely a missing reason; `service-balance-m2m` preserved and pinned; the exclusion is applied to `operator-balance-write` **and** `supervisor-overdraft-limit` — the second one is easy to miss and was not missed; `viewer-balance-read` deliberately left without the exclusion (reads, consistent with the base guard's stated read carve-out); `test_edge_may_read` also serves as the base+data wiring proof for `matrix-allows`, which is a nice touch; no `principal.type == "SERVICE"` rule.

---

JiRaska:

```
```

PR #3741, opened 2026-08-05
fix(fx): a derived inverse rate no longer reuses the source rate's id

### LINKED ISSUE #3374: fx: derived inverse rate reuses the source rate's id, so two directions share one identifier

JiRaska:
Closes #3374.

## The defect

`GET /api/v1/fx/rates/EUR/CZK` and `/CZK/EUR` answered with **the same `id`** — `FxRate.inverted()` carries the source row's id over, so both directions shared one identifier. The numbers were right (#3189); only the identity was wrong. A client caching quotes by `id` would serve one direction for the other, and a conversion audit record referencing the quote id cannot be replayed to a direction — the half that matters for money-path evidence.

## The fix — the issue's preferred option (2)

A derived quote answers with **`id: null`** and **`derivedFrom: <source row id>`**:

- `FxUseCase.getRate` now returns `ResolvedRate(rate, derivedFrom)`. The use case is the only place that knows whether the fallback inversion ran — the response pair matches the request either way, so the resource cannot infer it from the pair.
- `FxResource` maps to a new `FxRateResponse` DTO: stored quote → own `id`, `derivedFrom: null`; derived quote → `id: null`, `derivedFrom: <source id>`.
- **The domain is untouched.** `FxConversion.rateId = rate.id` still names a real `fx_rates` row (`fx_conversions.rate_id NOT NULL REFERENCES fx_rates`), because `inverted()` keeps the source id internally. The null lives only in the REST DTO, so CZK→foreign conversions are unaffected.

## Relationship to #3594 (same issue, option 1+2)

#3594 implements a deterministic derived id and is **blocked on the ASVS L3 gate**: any name-based UUID needs `MessageDigest`, and the banned-crypto-primitive check rejects the call site (`FxRate.kt:103-104`, SHA-256 included). Option (2) needs no hashing. #3594's FK objection to option (2) conflates the response DTO with the domain object — `rateId` is taken from the domain `FxRate`, which still carries the source row's id on an inversion. This PR also leaves #3594 free to land the `derivedFrom`-persistence idea later on top, if wanted.

## Verification

- `:openbank-fx-service:test` — full suite green (34 tests incl. the new derived-quote cases in `FxResourceTest` + `FxServiceTest`).
- `ktlintCheck`, `detekt`, `koverVerify`, `quarkusBuild` — green.
- `check-api-contract.py --base origin/main --enforce` — **additive change, `info.version` 1.5.0 → 1.6.0 OK**.
- `run-gates.py --group api` — ASVS L3 (banned crypto primitives) ok; route conformance + pact replay coverage pass.
- Existing consumer pact (`openbank-transaction-service → openbank-fx-service`, `GET /rates/EUR/CZK`) pins only `baseCurrency/quoteCurrency/bidRate/askRate` — unaffected; the stored direction keeps its shape, `derivedFrom` is additive.

### DIFF
--- a/docs/threat-models/openbank-fx-service.md
+++ b/docs/threat-models/openbank-fx-service.md
@@ -78,3 +78,12 @@ directly determines monetary outcomes — a manipulated rate is a financial-loss
     unaffected (keyed on `(source, pair, validFrom)`). Risk: a slightly stale CNB mid rate used in
     `spreadPct` display for at most 3 days. Rate table integrity and conversion accuracy are not affected
     (INTERNAL bid/ask are the settlement rates; CNB mid is display-only). Residual risk = **negligible**.
+- **2026-08-05** — Derived inverse-quote identity (#3374). `GET /api/v1/fx/rates/{base}/{quote}`
+  answered a pair derived by inverting the stored direction with the **source row's `id`**, so both
+  directions shared one identifier — a quote id could not be replayed to a direction (the §4
+  repudiation concern, and §5's "pin the rate id" assumption). The endpoint now answers `id: null` +
+  `derivedFrom: <source row id>` on a derived quote; a stored quote keeps its own id and no
+  `derivedFrom`. No new flow/surface/boundary — same endpoint, same data, honest identity.
+  `FxConversion.rateId` still references the real `fx_rates` row on both paths (FK unchanged), so
+  dispute defense via §5 is preserved. Risk class = **repudiation/auditability (reduced)**.
+  API contract: additive, `info.version` 1.5.0 → 1.6.0.
--- a/openbank-fx-service/src/main/kotlin/com/openbank/fx/application/port/in/FxPorts.kt
+++ b/openbank-fx-service/src/main/kotlin/com/openbank/fx/application/port/in/FxPorts.kt
@@ -4,13 +4,23 @@
 
 package com.openbank.fx.application.port.`in`
 
-import com.openbank.fx.domain.model.*
-import java.math.BigDecimal; import java.time.Instant; import java.util.UUID
+import com.openbank.fx.domain.model.FxConversion
+import com.openbank.fx.domain.model.FxRate
+import com.openbank.fx.domain.model.RateSource
+import com.openbank.fx.domain.model.RateType
+import java.time.Instant
+import java.util.UUID
 
 data class GetRateQuery(val baseCurrency: String, val quoteCurrency: String, val rateType: RateType = RateType.SPOT)
-data class ConvertCommand(val idempotencyKey: String, val partyId: UUID, val accountId: UUID?,
-    val fromCurrency: String, val toCurrency: String, val fromAmountMinorUnits: Long,
-    val partyName: String)
+data class ConvertCommand(
+    val idempotencyKey: String,
+    val partyId: UUID,
+    val accountId: UUID?,
+    val fromCurrency: String,
+    val toCurrency: String,
+    val fromAmountMinorUnits: Long,
+    val partyName: String,
+)
 
 data class GetRateHistoryQuery(
     val baseCurrency: String,
@@ -22,8 +32,16 @@ data class GetRateHistoryQuery(
     val offset: Int = 0,
 )
 
+/**
+ * The outcome of resolving a pair: the [rate] to quote, and — when the pair had to be answered by
+ * inverting the stored direction — [derivedFrom], the id of the stored row the quote was derived
+ * from. The REST layer nulls the response `id` on a derived quote (#3374); the domain [rate] keeps
+ * the source row's id either way, so `FxConversion.rateId` always names a real `fx_rates` row.
+ */
+data class ResolvedRate(val rate: FxRate, val derivedFrom: UUID?)
+
 interface FxUseCase {
-    suspend fun getRate(query: GetRateQuery): FxRate?
+    suspend fun getRate(query: GetRateQuery): ResolvedRate?
     suspend fun getAllRates(): List<FxRate>
     suspend fun getRateHistory(query: GetRateHistoryQuery): List<FxRate>
     suspend fun convert(cmd: ConvertCommand): FxConversion
--- a/openbank-fx-service/src/main/kotlin/com/openbank/fx/application/usecase/FxService.kt
+++ b/openbank-fx-service/src/main/kotlin/com/openbank/fx/application/usecase/FxService.kt
@@ -9,6 +9,7 @@ import com.openbank.fx.application.port.`in`.ConvertCommand
 import com.openbank.fx.application.port.`in`.FxUseCase
 import com.openbank.fx.application.port.`in`.GetRateHistoryQuery
 import com.openbank.fx.application.port.`in`.GetRateQuery
+import com.openbank.fx.application.port.`in`.ResolvedRate
 import com.openbank.fx.application.port.out.AmlCasePort
 import com.openbank.fx.application.port.out.AmlCaseRiskLevel
 import com.openbank.fx.application.port.out.FraudScoreCommand
@@ -79,11 +80,14 @@ class FxService(
      * customer edge turns the resulting null into `fx_rate_unavailable` + HTTP 502, so the app
      * reports a gateway error for what is really a missing derivation.
      *
-     * Inversion swaps bid and ask — see [FxRate.inverted].
+     * Inversion swaps bid and ask — see [FxRate.inverted]. A derived quote is reported with
+     * [ResolvedRate.derivedFrom] naming the stored row, so the REST layer can null the response
+     * `id` (#3374); the domain rate keeps the source id either way, so `FxConversion.rateId`
+     * always references a real `fx_rates` row.
      */
-    private suspend fun resolveRate(base: String, quote: String, type: RateType): FxRate? =
-        rateRepo.findLatest(base, quote, type)
-            ?: rateRepo.findLatest(quote, base, type)?.inverted()
+    private suspend fun resolveRate(base: String, quote: String, type: RateType): ResolvedRate? =
+        rateRepo.findLatest(base, quote, type)?.let { ResolvedRate(it, derivedFrom = null) }
+            ?: rateRepo.findLatest(quote, base, type)?.inverted()?.let { ResolvedRate(it, derivedFrom = it.id) }
 
     override suspend fun getAllRates() = rateRepo.findAll()
 
@@ -101,7 +105,7 @@ class FxService(
         convRepo.findByIdempotencyKey(cmd.idempotencyKey)?.let { return it }
         // Same resolution as the read path: a conversion must not be refused for a pair the bank
         // can price perfectly well from the other side.
-        val rate = resolveRate(cmd.fromCurrency, cmd.toCurrency, RateType.SPOT)
+        val rate = resolveRate(cmd.fromCurrency, cmd.toCurrency, RateType.SPOT)?.rate
             ?: error("No FX rate available for ${cmd.fromCurrency}/${cmd.toCurrency}")
         require(rate.isValid(Instant.now(clock))) { "FX rate expired for ${cmd.fromCurrency}/${cmd.toCurrency}" }
 
--- a/openbank-fx-service/src/main/kotlin/com/openbank/fx/infrastructure/rest/FxResource.kt
+++ b/openbank-fx-service/src/main/kotlin/com/openbank/fx/infrastructure/rest/FxResource.kt
@@ -9,7 +9,9 @@ import com.openbank.fx.application.port.`in`.ConvertCommand
 import com.openbank.fx.application.port.`in`.FxUseCase
 import com.openbank.fx.application.port.`in`.GetRateHistoryQuery
 import com.openbank.fx.application.port.`in`.GetRateQuery
+import com.openbank.fx.domain.model.FxRate
 import com.openbank.fx.domain.model.RateSource
+import com.openbank.fx.domain.model.RateType
 import com.openbank.libs.authz.Authorize
 import jakarta.annotation.security.RolesAllowed
 import jakarta.ws.rs.Consumes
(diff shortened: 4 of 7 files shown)


JiRaska:
## Two open PRs fix #3374, with different contracts

Flagging this because both are money-path, both are open, and merging either one leaves the other conflicting on `FxService.kt` and `openapi.yaml`. This is a decision, not a duplicate to close.

| | #3594 (08-03, +461/−9) | #3741 (08-05, +155/−26) |
|---|---|---|
| derived quote's `id` | its own, **computed** | **`null`** |
| how | `derivedId(sourceId)` — RFC 9562 v8 over SHA-256 of `"<ns>:<sourceId>"`, deterministic | not synthesised; `FxRateResponse.of` sets `id = null` when `derivedFrom != null` |
| `derivedFrom` | yes | yes |
| where the change lands | domain (`FxRate`) | REST DTO; domain keeps the source id |
| API contract | `id` stays required | **`id` becomes nullable** |

Both fix the actual defect — two directions answering under one identifier, so a client caching by `id` serves one direction for the other.

## What I would want decided, in order

**1. The foreign key.** #3594's description states that writing either a synthetic id or `null` breaks every CZK→foreign conversion on `FULL REFERENCES fx_rates(id)`. #3741 avoids that by keeping the source id in the domain and nulling only in the response — `FxConversion.rateId` still points at a real row. If that reading is right, #3741 sidesteps a constraint #3594 has to satisfy, and #3594 should say how it does.

**2. Nullable `id` is an API-contract change.** oasdiff has been unkind to nullable-shape changes in this repo before. `openbank-fx-service`'s `info.version` needs to reflect whatever oasdiff classifies this as, read off **live `main`** rather than either branch's base — two racing spec PRs can both claim the same version and both pass the gate, because the gate classifies against creation-time base.

**3. Is a derived quote addressable at all?** This is the real product question under the mechanism. #3594 says yes and pays a hash construction for it; #3741 says no and makes clients carry `derivedFrom` instead. An audit record that must be replayable to a direction wants the first; a client that only ever reads wants neither.

## Unrelated: #3594's `gates (api)` failure is fixed

It was ASVS V8.1 matching #3594's own KDoc — the two comment lines explaining why RFC 9562 v8 was chosen *over* v5-because-v5-is-SHA-1. `check-asvs-l3.py` scanned raw lines with no comment handling. Fixed in #3608 (merged), so a rebase should clear it. Not a reason to prefer #3741.

JiRaska:
Review: 2 findings, 0 blocking.

**1. Non-blocking — `derivedFrom` is read off the very field the PR declares wrong, one line after inverting.**
`FxService.kt` (`resolveRate`): `rateRepo.findLatest(quote, base, type)?.inverted()?.let { ResolvedRate(it, derivedFrom = it.id) }` — `it` is the **inverted** rate, so `it.id` is the source id only because `FxRate.inverted()` deliberately carries it over. The PR's whole premise is that a derived quote must not answer under the source row's identity, and the provenance link is then taken from exactly that carried-over id. It is correct today and `FxServiceTest` does assert `resolved.derivedFrom == stored.id`, so it is not silent — but it reads backwards. `rateRepo.findLatest(quote, base, type)?.let { ResolvedRate(it.inverted(), derivedFrom = it.id) }` says the same thing and takes the id from the row it actually came from.

**2. Non-blocking — the response shape for one rate is now a hand-mirrored copy, and the sibling endpoints still serialise the domain object.**
`FxRateResponse` (`FxResource.kt`) restates all 13 serialised members of `FxRate` including the computed `pair`/`midRate`/`spread`. `GET /rates` (`getAllRates`) and the history endpoint still return the domain `FxRate` directly. So a field added to `FxRate` later appears on those two endpoints and silently does **not** appear on `GET /rates/{base}/{quote}`, and nothing fails. Worth either a comment on `FxRate` pointing at the DTO, or building the DTO from the domain object by delegation.

**Already-written data: not ambiguous; no migration or backfill needed.** `fx_conversions.rate_id` did point at the stored `X/CZK` row for a CZK→X conversion priced off the inverse — but the direction is fully recoverable from the same row, which stores `from_currency`/`to_currency` and the numeric `applied_rate` alongside it. So no historic conversion is unattributable to a direction, and this PR correctly changes nothing in persistence: `FxConversion.rateId = rate.id` still names a real `fx_rates` row on both paths and the `NOT NULL REFERENCES` FK is untouched. The ambiguity that did exist was in the **response** identity only, which is what this fixes. (The PR body's "a conversion audit record referencing the quote id cannot be replayed to a direction" overstates it slightly for `fx_conversions`, for the reason above — the claim holds for a client caching REST quotes by `id`, which is the real defect.)

**Verified clean:** inversion direction and precision — `FxRate.inverted()` is `bidRate = 1 / askRate`, `askRate = 1 / bidRate` (the sides swap, so the customer is not quoted the wrong side of the spread), at scale 8 `HALF_UP` matching the stored `numeric(18,8)`; unchanged by this PR, and `FxResourceTest`'s new derived-quote case asserts both sides against values computed from the *source* rate rather than reflected off the result. `id: null` breaks no consumer I could find: `openbank-ledger-service`'s `FxServiceClient.FxRateResponse` is an explicit nullable subset that does not model `id`; `openbank-customer-edge` proxies the upstream JSON verbatim (`CustomerEdgeResource.kt:1213`, `:2629`) rather than binding a DTO; admin-ui reads the pair/rates. The committed pact `pacts/openbank-transaction-service-openbank-fx-service.json` pins one interaction, `GET /api/v1/fx/rates/EUR/CZK`, on `baseCurrency`/`quoteCurrency`/`bidRate`/`askRate` only — the **stored** direction, whose `id` stays populated — so the provider replay is unaffected in both shape and value. `openapi.yaml` `info.version` 1.5.0 → 1.6.0 is correct against **live** `origin/main`, which still reads 1.5.0, and the change is additive (`id` was previously undocumented, `derivedFrom` is new, both `nullable: true`). The threat-model entry is present and dated, as a money-path change requires.

This is an automated agent review, not an approval. A money-path change needs human review regardless.

JiRaska:
**Review: 4 findings, 1 blocking.**

Claim vs. code: the read endpoint now tells the truth about identity, and the inversion direction is right. What the change does not do is carry that honesty into the record that a dispute is actually replayed from.

1. **BLOCKING — `convert()` discards `derivedFrom`, so a conversion priced off an inverted rate stores a `rateId` naming the *opposite* direction with nothing recording the inversion.** `FxService.convert` — `val rate = resolveRate(cmd.fromCurrency, cmd.toCurrency, RateType.SPOT)?.rate`; the resolution result is unwrapped and the `derivedFrom` half thrown away. Scenario: customer converts CZK→EUR; only EUR/CZK is stored; the applied ask is `1/24.90 = 0.04016064`; `FxConversion.rateId` is persisted as the **EUR/CZK** row, whose stored bid/ask are 24.90/25.10. A §5 dispute replay that fetches that row and compares its rates against the applied rate finds no match, and nothing in `fx_conversion` says an inversion occurred — the replayer has to infer it from the currency pair. The threat-model entry added in this PR asserts "dispute defense via §5 is preserved" on the grounds that `rateId` still references a real `fx_rates` row; referencing a real row is not the same as identifying the price that was applied, and this is exactly the repudiation concern the PR opens by citing. The read-path fix makes the gap sharper, not smaller: `GET /rates` now refuses to pretend, while the audit record still does. A `rate_derived_from_id` (or a boolean `rate_inverted`) on `fx_conversion` is what closes it.

2. **`derivedFrom` is read off the inverted object, not the source row.** `FxService.resolveRate` — `rateRepo.findLatest(quote, base, type)?.inverted()?.let { ResolvedRate(it, derivedFrom = it.id) }`. `it` is the **inverted** rate; this is correct today only because `FxRate.inverted()` happens to copy the source `id` through — which is the precise behaviour this PR exists to stop relying on. The obvious follow-up (making `inverted()` mint its own id, or null it) silently turns `derivedFrom` into a fabricated identifier and the fix becomes the bug, with `FxServiceTest`'s `assertThat(rate.id).isEqualTo(stored.id)` going red in a way that reads as "the test needs updating". Bind the source first: `findLatest(quote, base, type)?.let { src -> ResolvedRate(src.inverted(), derivedFrom = src.id) }`.

3. **`info.version` 1.5.0 → 1.6.0 is classified additive because the spec never documented `id`, not because the change is additive.** The endpoint previously serialised the `FxRate` domain object directly, so `id` was always present and always a UUID on the wire; `ExchangeRateResponse` in `openapi.yaml` simply never listed it. This PR *adds* `id` to the schema and simultaneously makes it nullable, so `oasdiff` sees two new optional properties and reports minor — while a real client that keys a cache or an audit reference on `id` now receives `null` on every derived pair. The bump is defensible, but the release note should say plainly that `id` can now be absent, because the gate cannot say it.

4. **No test drives the new response shape over real HTTP.** `FxResourceTest` calls `resource.getRate(...)` directly and asserts on the `FxRateResponse` object. The change is a **serialisation** change — a nullable `UUID?` rendered by Jackson — and whether `id: null` is emitted, omitted, or emitted as `"null"` is decided by the serialiser config, not by the data class. That is what a consumer keys on. One `@QuarkusTest` + RestAssured case asserting the JSON body of a derived pair settles it.

Verified clean: the inversion direction is correct and asserted numerically — derived `bid = 1 / source ask`, `ask = 1 / source bid`, so the bank never quotes itself the better side by accident. The direct-hit path is still preferred over the inverse and is asserted with `coVerify(exactly = 0)` on the reverse lookup, so a stored pair can never be silently answered by a derivation. The ČNB branch is correctly marked non-derivable and passes `derivedFrom = null`. I checked the one committed pact for this provider (`pacts/openbank-transaction-service-openbank-fx-service.json`): it pins only `baseCurrency`/`quoteCurrency`/`bidRate`/`askRate` on the **stored** EUR/CZK direction, so the provider replay is unaffected by the new nullable `id`. The wildcard import in `FxPorts.kt` was expanded rather than hand-wrapped, and the threat model was updated in the same PR rather than deferred.

**Already-written data: ambiguous rather than wrong, and finding 1 is why.** Every `fx_conversion` row written for a pair only stored in the opposite direction already carries a `rateId` pointing at that opposite-direction row with no inversion marker — a real row, correct FK, but not sufficient on its own to reconstruct the applied price. Nothing is corrupt and no backfill is possible without first adding somewhere to put the flag, so this is a schema-plus-code follow-up rather than a migration this PR is missing. No rows need repairing to merge this.

Automated agent review, not an approval. Money-path: two human approvals and a threat-model read required regardless — and finding 1 is a question for whoever owns the §4/§5 dispute-replay procedure, since the threat-model entry in this diff makes a claim that the code does not yet support.

JiRaska:
Heads-up, no action needed on this PR.

I closed #3594 as superseded by this one — same defect, both `Closes #3374`, and only one can land since you both bump the spec `1.5.0 -> 1.6.0`. This one is CLEAN and carries the ADR-0030 threat-model update that #3594 lacked; fx-service is money-path, so that would have blocked it regardless.

Worth knowing the two designs are contradictory rather than duplicate, in case #3374 gets revisited: #3594 gave a derived quote its own deterministic id and kept `id` non-null; this PR nulls it and makes the field nullable. I checked whether nullable `id` breaks a consumer — it does not. customer-edge's `fxBidRate` reads bid/ask only, and every admin-ui `.id` reference is scheduling code, so nothing reads the rate id today.

One thing from #3594 worth keeping, which I have put in #3960 against `main` rather than pushing to your branch: `fx_conversions.rate_id` is `NOT NULL REFERENCES fx_rates(id)`, so a derived id reaching that column fails every CZK→foreign conversion on a foreign-key violation. Your `ResolvedRate` split gets this right — the domain rate keeps the source id — but nothing asserts it. #3960 pins the property itself rather than either derivation scheme, so it holds under your design and would have held under #3594's, and it is falsified against an injected regression rather than merely green.

No conflict expected: #3960 touches only `FxServiceTest.kt`, adding two tests below `convert prices a reverse-quoted pair instead of refusing it`. If it lands after this PR and your `getRate` signature change touches the same file, the merge is mechanical.

PR #3742, opened 2026-08-05
fix(fraud): prohibit customer-edge principal from fraud.score

JiRaska:
## Summary

Second PR from the #3734 graduated-services batch. `operator-fraud-write` was role-only over the whole `fraud.*` namespace, and the matrix grants `fraud.score` to `ROLE_OPERATOR` — so the customer-facing edge identity was admitted to the **real-time fraud scoring gate** via `matrix-allows`. Fleet caller audit: **no fraud caller exists in customer-edge at all** (no `fraudServiceUrl` anywhere); the only M2M scorer is the shared backend client (fx-service shadow scoring), already graduated via `service-fraud-scoring`.

## Type of change

## Linked issues

Refs #3734, #3698, #3740 (same pattern)

## Changes

- `operator-fraud-write` now excludes every `service-account-*` principal — because the rule is `startswith(input.action, "fraud.")`, this also closes any FUTURE fraud.* write shape to M2M.
- New edge-scoped `prohibited` clause vetoes `fraud.score` at the allow head — beats the matrix grant, which the rule exclusion alone does not.
- Edge-scoped, not all-service-accounts: the shared client IS the legitimate scorer (`service-fraud-scoring` preserved + pinned by test).
- Out of scope, documented in the ext + threat model: `fraud.review.read` rides base `operator-read-any` (the pre-existing fleet-wide M2M read over-grant, tracked in #3734).
- heredoc → standalone `fraud_rest_ext.rego`; new `fraud_rest_ext_test.rego` (6 tests); bundle + checksum regenerated (fraud only); threat model updated (ADR-0030 D2).

## Verification evidence

- `opa test` trio: **6/6 PASS**. Falsification: stripping both layers → **3/6 red**.
- Bundle regen idempotent; checksum `4dea6d9057bf5462`.

- [ ] Confirm caller audit: no edge fraud path exists; shared-client scoring untouched and tested.

## Security checklist

- [x] No secrets / PII. [x] Auth change → `security-review-required`. [x] No policy logic removed; fail-closed layers added, legit M2M path pinned by test.

## Compliance impact

- [x] DORA — ICT access control narrowed on the payment scoring gate.


JiRaska:

```

```

JiRaska:
**Review: hole closed for the shared client — no blocking findings.**

**The #3764 hole does not reproduce here, for a reason specific to this service.** `fraud.score` is the only `fraud.*` action in the `role_action_matrix` (`rules-opa-data.yaml:266`), and the shared client is the **legitimate** scorer via `service-fraud-scoring` — so the reach `matrix-allows` leaves it is reach it is supposed to have. Everything else in the `fraud.*` namespace is closed to every service account by the widened `not startswith(input.principal.id, "service-account-")` on `operator-fraud-write`, and because that rule is a prefix match on `fraud.`, future write shapes are closed by construction. The edge-pinned `prohibited` clause is the right instrument for the one action where an identity-scoped grant must survive: a blanket `service-account-` veto here would have 403'd fx-service's shadow scoring. **This PR's asymmetric shape is correct** — unlike its siblings, where the same asymmetry leaves a live gap.

**Test strength — no finding, but note why.** `test_shared_m2m_no_longer_admitted_via_role_only_reason` is the weak reason-absent form (`not "operator-fraud-write" in rest.allowed_reasons`) that failed to expose the hole in #3764. Here it happens to be the *right* assertion, because the shared client must still be allowed on `fraud.score` — `not rest.allow` would be false by design, and `test_shared_m2m_may_score_via_identity_rule` covers the positive with `decision.allow == true` plus the reason. So the form is defensible in this file; I flag it only because the same form is load-bearing-and-wrong in #3744, #3755, #3758, #3760 and #3761, and a reader sweeping the family for the pattern should not "fix" this one.

**Two things worth recording rather than acting on:**
1. `operator-fraud-write` is registered in `shared_m2m_write_prohibition.reasons`, which reads as a second layer. It is not one — see fact A below; that guard has never fired in a deployed bundle, and `matrix-allows` would defeat its `every` clause even if it had.
2. The out-of-scope note about `fraud.review.read` riding base `operator-read-any` is accurate and correctly deferred to #3734. Worth pairing with the observation that `OnnxFraudModel`-adjacent read endpoints are the ones an analyst actually uses, so widening the read side later needs its own caller audit.

**Verified clean:** `service-fraud-scoring` preserved and pinned; `test_edge_may_not_score_despite_matrix_grant` uses the strong `rest.allow == false` form and `test_prohibition_fires_for_edge_on_score` proves the veto (not merely the absence of a reason) is what denies; the caller audit ("no `fraudServiceUrl` anywhere in customer-edge") is consistent with what I see in the edge tree; checksum `4dea6d9057bf5462` consistent between bundle and pod-roll annotation; no `principal.type == "SERVICE"` rule — the four matches in the diff are prose explaining unreachability, which is the correct treatment.

---

JiRaska:

```
```

PR #3744, opened 2026-08-05
fix(ledger): prohibit customer-edge principal from ledger writes incl. year-close attest

JiRaska:
## Summary

Third PR from the #3734 graduated-services batch — and the most sensitive one: ledger is the double-entry **book of record**. `operator-ledger-write` and `operator-year-close-attest` were role-only, and the matrix grants `ledger.create/reverse/trigger/replay` to `ROLE_OPERATOR`. The customer-facing edge identity was therefore admitted to journal post/reverse and FX revaluation via `matrix-allows` — and to `ledger.approve` via the attest rule, whose own comment has always said **no SERVICE principal must ever reach it** (human sign-off by design, ADR-0078 D5).

## Type of change

## Linked issues

Refs #3734, #3698, #3740, #3742 (same pattern)

## Changes

**Caller audit:** no `ledgerServiceUrl` exists anywhere in customer-edge — **no edge ledger caller at all**. The legitimate M2M writers (transaction/lending/settlement via the shared client) keep their identity-scoped `service-ledger-post` / `service-ledger-reverse` rules, both pinned by tests.

**Tightening (two layers):**
- `operator-ledger-write` and `operator-year-close-attest` now exclude every `service-account-*` principal. Side effect, aligned with the ext's own documented intent: `ledger.trigger`/`ledger.replay` are now closed to the shared client at the rule layer too ("an un-mapped SERVICE call to either 403s, by design" — the matrix grant covering them is a separate pre-existing over-grant, out of scope).
- New edge-scoped `prohibited` clause vetoes all five write actions at the allow head — **`ledger.approve` included despite no matrix grant**, so no future matrix edit can hand the edge a statutory close event.
- Edge-scoped, not all-service-accounts: the shared client IS a legitimate writer here.

**Shape:** heredoc → standalone `ledger_rest_ext.rego` (the original "book of record" framing comment preserved); new `ledger_rest_ext_test.rego` (12 tests); bundle + checksum regenerated (ledger only); threat model updated (ADR-0030 D2).

## Verification evidence

- `opa test` trio: **12/12 PASS**. Falsification: stripping both layers → **7/12 red**.
- Bundle regen idempotent; checksum `3b5abe84bb0f28f7`.

- [ ] Threat model change present (§8, 2026-08-05 entry). [ ] 2 approvals.
- [ ] Confirm caller audit: no edge ledger path exists; shared-client post/reverse untouched and tested.

## Security checklist

- [x] No secrets / PII. [x] Auth change → `security-review-required`. [x] No policy logic removed; fail-closed layers added, legit M2M paths pinned by tests.

## Compliance impact

- [x] DORA — ICT access control narrowed on the book of record; year-close attestation (statutory sign-off) now structurally M2M-proof.


JiRaska:

```

```

JiRaska:
**Review: partial — `ledger.approve` is genuinely closed, but `ledger.trigger` and `ledger.replay` remain reachable by the shared client via `matrix-allows`.**

**1. `openbank-infra/gitops/components/ledger/ledger_rest_ext.rego` — the veto is pinned to `service-account-openbank-edge`.**

`rules-opa-data.yaml:281-284` grants ROLE_OPERATOR `ledger.create`, `ledger.replay`, `ledger.reverse`, `ledger.trigger`. For `principal.id = service-account-openbank-services`:
- `ledger.create` / `ledger.reverse` — admitted by `service-ledger-post` / `service-ledger-reverse`. **Intended**, verified callers, correct.
- `ledger.trigger` (FX revaluation) and `ledger.replay` — `operator-ledger-write` is now correctly excluded ✅, no identity-scoped rule covers them, but `matrix-allows` (`rest.rego:151-155`) fires, the new `prohibited` does not (wrong id), and the base `prohibited` (`rest.rego:485-491`) cannot (fact A below; also `operator-ledger-write` is registered in `shared_m2m_write_prohibition.reasons`, and that still would not help — `matrix-allows` is not a listed reason, so the `every` clause is false). Net: **allowed.**

So the shared backend client can still trigger an FX revaluation posting run and replay the journal stream against the book of record. Smaller than the sibling PRs' residue, but this is the double-entry book of record and `ledger.replay` is not a benign verb.

`ledger.approve` (year-close attest) **is** correctly closed for both service accounts: it is not in the matrix, so the `not startswith(input.principal.id, "service-account-")` added to `operator-year-close-attest` is sufficient. That was the sharpest item in the PR description and it holds — the rule comment's long-standing "no SERVICE principal must ever reach it" is finally true.

**2. `ledger_rest_ext_test.rego` — `test_shared_m2m_no_longer_admitted_via_role_only_reason` is the #3764 pattern:** it asserts the reason `operator-ledger-write` is absent, on `ledger.create` — an action the shared client is *supposed* to reach, so the strong form would be wrong there and the weak form proves nothing about the residue. The two actions that actually matter, `ledger.trigger` and `ledger.replay`, have no shared-principal test in either form. `not rest.allow with input as {"principal": services_m2m, "action": "ledger.replay"}` is red today.

Fix: add `ledger.trigger`/`ledger.replay` to a second `prohibited` clause keyed on `startswith(input.principal.id, "service-account-")`, or widen the existing veto and carve out `create`/`reverse` for the shared client. #3698 is the family's correct template.

**Verified clean:** all five edge denials use the strong `rest.allow == false` form; `service-ledger-post`/`service-ledger-reverse` preserved and pinned by positive tests; `service-ledger-read` untouched; `test_edge_may_not_attest_year_close` asserts both the denial and that the veto is what produced it — the best-shaped test in the batch; checksum consistent between the bundle and the pod-roll annotation; the removed heredoc's `principal.type == "SERVICE"` prose is carried into the file verbatim and remains prose, not a rule.

---

JiRaska:

```
```

JiRaska:
Refreshed against `main` (was 244 behind) and, while resolving, measured what this PR actually buys. Posting it because the answer is larger than the PR description claims.

**The generator conflict was the interesting one, and it was not a formatting clash.** This branch moves the ledger extension out of a heredoc inside `gen-ledger-opa-bundle.sh` into a standalone `ledger_rest_ext.rego`; meanwhile #3812 landed on `main` and edited that same heredoc, adding `not startswith(input.principal.id, "service-account-")` to `operator-year-close-attest`. Taking this branch's side wholesale is exactly the shape that silently drops a merged security fix.

It does not here — checked rather than assumed. The standalone file already carries that exclusion in **two** rules (`operator-ledger-write` as well as `operator-year-close-attest`), so it is a superset of #3812, not a rollback. #3812's own note explains why it declined the first one: an exclusion on `operator-ledger-write` alone changes nothing, because `matrix-allows` is a base-layer reason no per-service extension can veto. This branch closes that second path with a `prohibited` veto, which is what makes the exclusion load-bearing.

**What that means on today's `main`.** Decisions evaluated with `opa eval` against the two bundles as generated — `main`'s committed one and this branch's after the merge — for `service-account-openbank-edge` with `ROLE_OPERATOR` (the realm role the edge service account actually carries):

| action | `main` today | this branch |
|---|---|---|
| `ledger.create` | **ALLOW** (`matrix-allows`) | DENY |
| `ledger.reverse` | **ALLOW** (`matrix-allows`) | DENY |
| `ledger.trigger` | **ALLOW** (`matrix-allows`) | DENY |
| `ledger.replay` | **ALLOW** (`matrix-allows`) | DENY |
| `ledger.approve` | DENY | DENY (unchanged — #3812 closed it) |
| `ledger.read` | ALLOW | ALLOW (unchanged) |

So on `main` right now the customer-edge principal can post and reverse journal entries on the book of record, trigger FX revaluation and replay historical events. Reads keep working after the change, which is the control that makes the first four meaningful — a bundle that simply denied everything would show the same four DENYs.

I did **not** re-verify the `AUTHZ_ENFORCE` posture of ledger-service, so whether that path is reachable in the deployed pod today is a separate question from whether the policy permits it.

Resolution detail: the two derived files (`ledger-opa-bundle.yaml`, and the pod-roll checksum line in `ledger-service.yaml`) were resolved by taking `main`'s copy and re-running the generator, never by hand-splicing. Bundle now embeds `ledger_rest_ext.rego`; `check-opa-bundle-parses.py` reports 41 bundles, no unloadable policy.

Still wants a human — money-path authz, 2 approvals by rule.

PR #3748, opened 2026-08-05
fix(sca): close role-only M2M path on SCA writes; grant shared client scaChallenge.consume

JiRaska:
## Summary

Fourth PR from the #3734 graduated-services batch — and the one that proves why the caller audit comes first. `operator-sca-write` was role-only over the whole `scaChallenge.*`/`device.*` families, admitting both M2M clients to every SCA write. **But SCA is the row where M2M writers legitimately exist:** delegation-service (grant-accept ceremony) and document-service (DOCUMENT_SIGNING, ADR-0169 D2) both POST `/api/v1/sca/challenges/{id}/consume` via the shared client — and rode the role-only hole. A naive prohibition would have 403'd both ceremonies.

## Type of change

## Linked issues

Refs #3734, #3698, #3740, #3742, #3744

## Changes (ordering is the point — identity-scoped rule FIRST)

1. `service-sca-shared-client-m2m` gains `scaChallenge.consume` — verified callers: `delegation-service/.../ScaChallengeClient.kt` (`consumeChallenge`) and `document-service/.../ScaChallengeClient.kt` (`consume`, DOCUMENT_SIGNING). account-service and consent-service are GET-only (already covered).
2. `operator-sca-write` now excludes every `service-account-*` principal — closing every present and future `scaChallenge.*`/`device.*` write to M2M (incl. `scaChallenge.verify`, documented human-channel-only).
3. **No prohibition clause, unlike the sibling PRs** — the role_action_matrix grants no `scaChallenge.*`/`device.*` write to ROLE_OPERATOR, so `matrix-allows` admits nothing the exclusion doesn't close (verified; documented in the ext).
4. Edge's ceremony actions stay on `service-sca-edge-m2m` (initiate/read/decide/consume); its `device.*` access rides **base** `edge-service-notification` — pinned by test, not duplicated.
- heredoc → standalone `sca_rest_ext.rego`; new `sca_rest_ext_test.rego` (11 tests); bundle + checksum regenerated (sca only); threat model updated (ADR-0030 D2).

## Verification evidence

- `opa test` trio: **11/11 PASS**. Falsification: stripping the exclusion → **4/11 red**; removing the consume widening → the delegation/document ceremony regression test goes **red** (the test that would have caught the outage).
- Bundle regen idempotent; checksum `11b4bfa36a37649b`.

- [ ] Confirm the two verified consume callers (file paths above) — they are why this is not an interest-style prohibition.

## Security checklist

- [x] No secrets / PII. [x] Auth change → `security-review-required`. [x] No policy logic removed; legit M2M ceremony paths widened FIRST, then the role-only hole closed.

## Compliance impact

- [x] DORA + PSD2/SCA — access control on the SCA ceremony narrowed without breaking customer signing flows.

### DIFF
--- a/docs/threat-models/openbank-sca-service.md
+++ b/docs/threat-models/openbank-sca-service.md
@@ -58,6 +58,24 @@ is the **authentication assurance gate** for payments and consent — defeating 
 
 ## 6. Change log
 
+- **2026-08-05** — Close the role-only M2M path on the SCA ceremony; widen the shared-client
+  identity rule to `scaChallenge.consume` FIRST (#3734). `operator-sca-write` was role-only over
+  the whole `scaChallenge.*`/`device.*` families, so both M2M clients (HUMAN-classified,
+  ROLE_OPERATOR) were admitted to every SCA write — including `scaChallenge.verify` (the OTP
+  fallback, documented human-channel-only) and any future action in those families. SCA differs
+  from the other #3734 rows: the edge IS a legitimate ceremony caller (initiate/read/decide/
+  consume via `service-sca-edge-m2m`; `device.*` via base `edge-service-notification`) and the
+  shared client legitimately consumes challenges — delegation-service's grant-accept ceremony
+  and document-service's DOCUMENT_SIGNING ceremony (ADR-0169 D2) both POST
+  `/api/v1/sca/challenges/{id}/consume` and rode the role-only hole until now. So the ordering
+  is the #3734 "identity-scoped rule FIRST" pattern: `service-sca-shared-client-m2m` gains
+  `scaChallenge.consume`, THEN `operator-sca-write` excludes every `service-account-*`
+  principal. No prohibition clause: `rules.yaml`'s matrix grants no `scaChallenge.*`/`device.*`
+  write to ROLE_OPERATOR, so `matrix-allows` admits nothing the exclusion doesn't close (unlike
+  balance/ledger/fraud). Falsified by `sca_rest_ext_test.rego` — stripping the exclusion turns
+  4 of 11 red; removing the consume widening turns the delegation/document regression test red.
+  The ext moved from a generator heredoc to a standalone `sca_rest_ext.rego` so `opa test` can
+  load it. Rollback: revert the ext — the ceremonies keep working via the widened identity rule.
 - **2026-06-11** — Domain metrics + outbox-backlog gauge (ADR-0077 / ADR-0079). New `DomainMetrics`
   call sites: `scaChallengeIssued(method)` after a challenge is persisted in `initiate`, and
   `scaChallengeResolved(method, outcome)` once a challenge reaches a terminal `ScaStatus`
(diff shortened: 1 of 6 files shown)


JiRaska:

```

```

JiRaska:
**Review: hole closed — 1 finding on the grant, none blocking.**

**The #3764 hole does not exist here, and the PR's stated reason is verifiable.** The `role_action_matrix` grants ROLE_OPERATOR exactly one `device.*`/`scaChallenge.*` action — `device.list` (`rules-opa-data.yaml:213`), a read. No SCA *write* is matrix-granted, so `matrix-allows` cannot re-admit anything the widened exclusion on `operator-sca-write` closes, and omitting a `prohibited` clause is correct rather than an oversight. I checked this against the full matrix rather than taking the ext's word for it.

**The tests are strong.** `test_shared_m2m_may_not_decide` and `test_edge_may_not_verify_via_role_only_path` both assert `rest.allow == false`, not reason-absence — this file does not have the #3764 weakness, and it is one of only two in the batch that doesn't. `test_edge_may_enroll_device_via_base_rule` correctly pins the base `edge-service-notification` grant rather than duplicating it.

**Finding — the `scaChallenge.consume` grant is justified but its blast radius is larger than the justification.** `service-sca-shared-client-m2m` is keyed on `input.principal.id == "service-account-openbank-services"`, and that identity is shared by ~23 backend services. The audit that motivates the grant names two callers (delegation-service's grant-accept ceremony, document-service's DOCUMENT_SIGNING per ADR-0169 D2) — but the grant admits *every* service on that client to consume *any* challenge id. So the narrowing the PR describes ("verified callers") is a property of today's call graph, not of the policy. Two things worth stating in the ext or the threat model rather than leaving implicit:
1. What stops a consume from a service that has no business completing an SCA ceremony? If the answer is challenge-to-purpose binding inside sca-service (the challenge carries its own type/subject and the consumer must match), say so and name the class — that is the real control, and rego is only the outer gate. If there is no such binding, this grant is an SCA-bypass primitive for any compromised backend and deserves its own follow-up.
2. This is a *widening* shipped inside a tightening PR. It is correctly sequenced (identity rule first, then the exclusion — the ordering note in the PR body is right, and stripping it would have 403'd both ceremonies), but it is the one change in this batch that adds reach, so it is where a human reviewer's attention should go rather than the exclusion.

The `scaChallenge.read` half of the same rule is unobjectionable.

**Verified clean:** the exclusion is applied to `operator-sca-write` over both the `scaChallenge.` and `device.` families, so future write shapes in either are closed by construction; `service-sca-edge-m2m` keeps the edge's four ceremony actions and is pinned; `scaChallenge.verify` is correctly denied to both M2M identities and allowed to a human operator; the `rules_mock` explicitly sets `shared_m2m_write_prohibition.reasons` to `[]`, which is honest about not relying on the base guard (see fact A — it is inert in the cluster anyway); no `principal.type == "SERVICE"` rule.

---

PR #3750, opened 2026-08-05
fix(consent): widen M2M exclusion on operator-consent-write to all service accounts

JiRaska:
## Summary

Fifth PR from the #3734 graduated-services batch — the smallest correct diff of them all. ADR-0206 D5 already closed `operator-consent-write` to the shared backend client, but the exclusion named `service-account-openbank-services` literally — leaving `service-account-openbank-edge` (customer-facing, HUMAN-classified, ROLE_OPERATOR) admitted to **every** `consent.*` write (`grant`, `activate`, `validate`, …) via the role-only rule.

The fix is a one-line widening to the `service-account-` prefix. Nothing else moves:

- The edge's **legitimate** consent access is exactly `{consent.list, consent.revoke}` — the customer's PSD2 consent screen — already granted by base `edge-service-consent` in the shared `rest.rego` (now pinned by tests here).
- **No prohibition clause needed** (unlike #3740/#3742/#3744): `rules.yaml`'s role_action_matrix grants no `consent.*` write to ROLE_OPERATOR, so `matrix-allows` admits nothing the exclusion doesn't close.
- The shared client keeps its scoped grants (`service-consent-m2m`, `service-consent-m2m-marketing`) untouched.

## Type of change

## Linked issues

Refs #3734, #3698, #3740, #3742, #3744, #3748

## Changes

1. `operator-consent-write`: `not input.principal.id == "service-account-openbank-services"` → `not startswith(input.principal.id, "service-account-")`, with the rule comment updated to record why.
2. `consent_rest_ext_test.rego` +6 tests: edge denied on `consent.grant`/`consent.activate`/`consent.validate` via operator rule and bundle-wide; edge keeps `consent.revoke`/`consent.list` via base `edge-service-consent`.
3. Bundle + checksum regenerated (consent only, `f81864f16f798b9f`, idempotent); threat model updated (ADR-0030 D2).

## Verification evidence

- `opa test` trio: **19/19 PASS** (12 pre-existing + 6 new + 1 naming). Falsification: reverting the exclusion to the shared-only form → **4/19 red** (both edge via-operator-rule tests + both count==0 denial tests).

- [ ] Confirm no caller regression: edge's consent calls (`CustomerEdgeResource.kt` DELETE `/consents/{id}`) ride base `edge-service-consent`, not this ext.

## Security checklist

- [x] No secrets / PII. [x] Auth change → `security-review-required`. [x] No policy logic removed; exclusion only widened.

## Compliance impact

- [x] DORA + PSD2 — access control on the consent lifecycle narrowed to named identities; customer consent screen path preserved.


JiRaska:
**Review: hole closed — no blocking findings. This is one of the two correct PRs in the batch.**

**The #3764 hole does not exist here, and for a checkable reason.** `rules-opa-data.yaml`'s `role_action_matrix` grants no `consent.*` action to any role — I enumerated ROLE_OPERATOR (lines 204-311), ROLE_ADMIN (inherits), ROLE_COMPLIANCE, ROLE_PAYMENTS and the rest; `consent` appears nowhere. So `matrix-allows` cannot fire for a consent write, and the rule-level exclusion is genuinely the whole control. The PR body says exactly this and it checks out. No `prohibited` clause is needed, and adding one would have been noise.

**On the "widen to all service accounts" question — the wildcard is correct in both directions.**
- *Does it catch identities it shouldn't?* Three service accounts exist in the realm (`realm-template.json`): `openbank-edge`, `openbank-services`, `openbank-mcp-service`. `openbank-services` keeps its full scoped surface via `service-consent-m2m` (`consent.read/validate/activate/reject`) and `service-consent-m2m-marketing` (`consent.grant/revoke`) — untouched by this diff. `openbank-mcp-service` carries no realm roles at all, so `operator-consent-write` never fired for it. `openbank-edge`'s legitimate surface is `{consent.list, consent.revoke}` via base `edge-service-consent` — I verified that against the code, not just the PR body: `CustomerEdgeResource.kt:668-698` has exactly one GET `/consents` (list by party) and one DELETE `/consents/{id}` (revoke), and no grant/activate/validate call path exists in customer-edge. Nothing legitimate is caught.
- *Does it miss identities it should catch?* No — the prefix is what Keycloak deterministically produces for `preferred_username` on any client_credentials token, so a future fourth service account is closed by default. That is the right default direction for this rule.

**Tests are strong enough.** `test_edge_denied_entirely_on_grant` and `test_edge_denied_entirely_on_consent_validate` assert `count(allowed_reasons) == 0`, which — with no `prohibited` clause in play — is strictly stronger than `not rest.allow` and does not have the #3764 weakness. `test_edge_cannot_grant_via_operator_rule` / `_activate_` are the weak reason-absent form, but they are stated as complements to the count assertions rather than standing in for them, and the count tests cover the same actions. Both base-rule preservation tests (`edge-service-consent` on revoke/list) are the right regression guard for the one thing this widening could have broken.

**Verified clean:** the diff is exactly the one-line predicate change plus tests, bundle and threat model — no scope creep; checksum `f81864f16f798b9f` consistent between `consent-opa-bundle.yaml` and the pod-roll annotation in `consent-service.yaml`; no `principal.type == "SERVICE"` rule.

One note that does not change the verdict: fact A below means the *base* `rest.rego` guard listing `operator-consent-write` in `shared_m2m_write_prohibition.reasons` — the ADR-0206 D5 control this PR builds on — has never actually fired in a deployed bundle. This PR's exclusion is doing all the work, which is a point in its favour, but the threat model's account of the layering is optimistic.

---

PR #3755, opened 2026-08-05
fix(sepa-payment): close edge M2M path on SEPA writes beyond create/read

JiRaska:
## Summary

Sixth and final PR of the #3734 graduated-services batch — the row with the widest blast radius. `operator-sepa-payment-write` was role-only over the whole `sepaPayment.*` family, **and** `rules.yaml`'s role_action_matrix grants `sepaPayment.{approval.decide, create, handleReturn, transitionStatus}` to ROLE_OPERATOR — which the customer-facing edge service-account carries. So `service-account-openbank-edge` could execute a clearing return (`pacs.004`), force an ops status transition, or **decide a four-eyes approval** on the SEPA rail.

The edge's *legitimate* access is already identity-scoped in this ext: `service-sepa-payment-edge-m2m` grants exactly `{sepaPayment.create, sepaPayment.read}` — verified caller `customer-edge UpstreamClient` POST `/sepa-payments` (after the edge's own ownership guard + SCA gate, ADR-0021). Clearing-simulator keeps `handleReturn` via the shared-client identity rule.

## Type of change

## Linked issues

Refs #3734, #3698, #3740, #3742, #3744, #3748, #3750

## Changes

1. `operator-sepa-payment-write` excludes `service-account-*` principals.
2. New `prohibited` veto: edge denied on `{sepaPayment.transitionStatus, sepaPayment.handleReturn, sepaPayment.approval.decide}` — needed **in addition to** the exclusion because `matrix-allows` in base rest.rego grants those actions to ROLE_OPERATOR without consulting this rule (base gates its allow head on `not prohibited`, so the veto always wins). `sepaPayment.create` deliberately absent — that is the edge's identity-scoped grant.
3. Heredoc → standalone `sepa_payment_rest_ext.rego` + new `sepa_payment_rest_ext_test.rego` (11 tests; the matrix mock mirrors the real `data.rules.authz.role_action_matrix[role].grant` shape, so the matrix interaction is genuinely exercised — a wrong-shaped mock would have made the prohibition unfalsified).
4. Generator patched to cat the standalone file (checksum + body); bundle + checksum regenerated (`f6c4b36dbe0657c8`, idempotent); threat model updated (ADR-0030 D2).

## Verification evidence

- `opa test` trio: **11/11 PASS**. Falsification, both layers: stripping the exclusion → **2/11 red** (via-operator-rule tests); stripping the prohibition → **4/11 red** (the three edge denial tests go green-to-red via the matrix grant + the prohibition-fires test).
- Gates locally: security 11/11 (incl. threat-model gate with `PR_DIFF_BASE`), drift gate PASS, bundle regen idempotent.

- [ ] Confirm caller audit: customer-edge calls only `sepaPayment.create` (+ read); clearing-simulator calls `handleReturn` via the shared client (`CustomerEdgeResource.kt`, `SepaPaymentClient`).

## Security checklist

- [x] No secrets / PII. [x] Auth change → `security-review-required`. [x] No policy logic removed; identity-scoped grants preserved, role-only + matrix paths closed.

## Compliance impact

- [x] DORA + PSD2 — access control on the SEPA rail narrowed to named identities; four-eyes approval decisions no longer reachable by a customer-facing M2M proxy.


JiRaska:
**Review: same hole as #3764 — `sepaPayment.create`, `transitionStatus` and `approval.decide` stay reachable by the shared client.**

**1. `openbank-infra/gitops/components/payments/sepa_payment_rest_ext.rego` — veto pinned to the edge identity.**

`rules-opa-data.yaml:303-306` grants ROLE_OPERATOR all four SEPA writes. For `principal.id = service-account-openbank-services`:
- `operator-sepa-payment-write` — correctly excluded now. ✅
- `service-sepa-payment-shared-client-m2m` — grants `sepaPayment.handleReturn` (intended, verified caller). ✅
- `matrix-allows` — fires on `create`, `transitionStatus`, `approval.decide` (and `handleReturn`). **Not closed by anything.**
- The new `prohibited` requires the edge id → does not fire.
- Base `prohibited` (`rest.rego:485-491`): note this is the one service in the batch whose reason **is** registered (`shared_m2m_write_prohibition.reasons` lists `operator-sepa-payment-write`). It still does not help — twice over. First, the register is undefined in every deployed bundle (fact A). Second, even if it were populated, `allowed_reasons` here is `{"matrix-allows"}` or `{"matrix-allows", "service-sepa-payment-shared-client-m2m"}`, and `matrix-allows` is not a listed reason, so the `every` clause is false either way. **Being in the register is not protection** — worth knowing before the next service is graduated into it on the assumption that it is.

Net: the shared client can initiate SEPA credit transfers (`sepaPayment.create`) and decide four-eyes payment approvals. On a money-path rail, that is the finding.

**2. `sepa_payment_rest_ext_test.rego` — `test_shared_no_operator_rule_on_transition` is the #3764 pattern:** reason-absent rather than `not rest.allow`, on `sepaPayment.transitionStatus`, which is matrix-granted. The strong form is red today. The edge tests all use `not rest.allow` correctly — the asymmetry is only on the shared principal.

Fix: extend the veto to `startswith(input.principal.id, "service-account-")` and carve `sepaPayment.handleReturn` back in for the shared client, or list the three residual actions explicitly. #3698 is the template.

**Verified clean:** the edge's `create`/`read` flow is preserved by `service-sepa-payment-edge-m2m` and pinned; `handleReturn` correctly stays open to the shared client and closed to the edge — the one place in this batch where the two M2M identities are deliberately treated differently, and it is right; `ROLE_PAYMENTS` matches the matrix; no `principal.type == "SERVICE"` rule.

---

PR #3757, opened 2026-08-05
test(ci): give check-verification-metadata-complete.py a --selftest

JiRaska:
Refs #3240 (part 3 — the bookkeeping tail; the gate itself stays in its `services-ci.yml` job, where the Java 25 toolchain and Gradle cache already are — see the issue thread for why moving it into `Validate manifests` would tax every PR with a JDK install + Gradle resolution it does not need).

## What this adds

`check-verification-metadata-complete.py --selftest` — a Gradle-free falsification harness. The gate is genuinely enforced today but had no self-test, so nothing established it can fail; that is the one real gap left from #3240.

The harness exercises exactly what the verdict rests on:

- **`artifact_set`** — pairs only: an artifact before any component and a component with no artifacts drop out.
- **The verdict algebra** (`missing = after - before`) — an added pair is named exactly; identical sets report no gap; **removals are NOT reported** (a narrower task list would "remove" half the file).
- **`tasks_for`** — both plugin-application shapes (`id("openbank.quarkus-service")` AND `alias(libs.plugins.quarkus)`) get both graphs; the comment-only mention — the prose that once classified a non-Quarkus module AS Quarkus — gets `testClasses` only; a missing build file does not crash.

## Falsification (both directions)

- `python3 .github/scripts/check-verification-metadata-complete.py --selftest` → 9 cases, exit 0.
- Mutating the alias regex in a copy → the self-test fails with 1 case, exit 1. The harness is falsifiable, not ceremonial.

`--modules` stays required for real runs (argparse error without it unless `--selftest`); every existing invocation (`services-ci.yml`, the nightly shard guard) is unchanged.

### DIFF
--- a/.github/scripts/check-verification-metadata-complete.py
+++ b/.github/scripts/check-verification-metadata-complete.py
@@ -29,11 +29,13 @@ Two things this deliberately does NOT do:
 Usage:
     check-verification-metadata-complete.py --modules openbank-a,openbank-b
     check-verification-metadata-complete.py --modules "" --enforce   # no-op, exits 0
+    check-verification-metadata-complete.py --selftest               # prove it can fail
 """
 
 from __future__ import annotations
 
 import argparse
+import os
 import pathlib
 import re
 import shutil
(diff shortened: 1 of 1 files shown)


JiRaska:
Ran the harness rather than reading the description:

```
ok: identical sets report no gap
ok: removals are not reported as gaps
ok: convention plugin -> both graphs
ok: alias(libs.plugins.quarkus) -> both graphs
ok: comment-only mention -> testClasses only
ok: no Quarkus plugin -> testClasses only
ok: missing build file -> testClasses only
selftest OK
```

Two of those are the ones that make it worth having. **"removals are not reported as gaps"** is the asymmetry the verdict rests on (`missing = after - before`) and the direction a naive set-diff gets wrong. **"comment-only mention -> testClasses only"** is the comment-vs-code distinction this repo keeps paying for — a plugin named in a comment must not change the task graph, and a harness that only ever saw real plugin declarations would never find out.

Covering both plugin-application shapes matters for the same reason: `id("openbank.quarkus-service")` and `alias(libs.plugins.quarkus)` are the same fact written two ways, and a parser that handles one silently under-scopes the other.

Merging: 5/5 required green, nothing pending, nothing failing.

PR #3758, opened 2026-08-05
fix(account): close edge M2M path on account lifecycle writes

JiRaska:
## Summary

Second batch, first PR (#3734). Account is the widest row of the audit: the role_action_matrix grants **all ten** `account.*` actions to ROLE_OPERATOR, and `operator-account-write` was role-only — so `service-account-openbank-edge` (customer-facing proxy, HUMAN-classified, ROLE_OPERATOR) could close, freeze, unfreeze, delegate-authorize, and **decide four-eyes approvals** on any account.

The edge's verified legitimate access is exactly `account.{create, update}` — onboarding open-account, pocket add/close, savings-goal set/clear (ADR-0104/ADR-0153) — already identity-scoped in `service-edge-account-m2m`. The shared client keeps `account.read` (billing discovery sweep, ADR-0143).

## Type of change

- [x] fix (security tightening on a money-path service)

## Linked issues

Refs #3734, #3698, #3740, #3742, #3744, #3748, #3750, #3755

## Changes

1. `operator-account-write` excludes `service-account-*` principals.
2. New `prohibited` veto: edge denied on `{account.close, account.freeze, account.unfreeze, account.authorize, account.approval.decide}` — needed in addition to the exclusion because `matrix-allows` grants all ten actions to ROLE_OPERATOR without consulting this rule.
3. Heredoc → standalone `account_rest_ext.rego` + new `account_rest_ext_test.rego` (13 tests; matrix mock mirrors the real `data.rules.authz.role_action_matrix[role].grant` shape).
4. Generator patched (checksum + body cat the file); bundle + checksum regenerated (`d3b2a9eb2e3e2060`, idempotent); threat model updated (ADR-0030 D2).

## Verification evidence

- `opa test` trio: **13/13 PASS**. Falsification: stripping the exclusion → **2/13 red**; stripping the veto → **6/13 red** (5 edge denial tests + veto-fires go green-to-red via the matrix grant).
- Gates locally: security 11/11 (incl. threat-model gate with `PR_DIFF_BASE`), bundle regen idempotent.

## Reviewer checklist (money-path)

- [ ] Threat model change present (§6, 2026-08-05 entry). [ ] 2 approvals.
- [ ] Confirm caller audit: customer-edge calls only `account.create` + `account.update` (+ reads); no in-repo M2M caller for close/freeze/unfreeze/authorize (ext's own audit comment).

## Security checklist

[...]

PR #3760, opened 2026-08-05
fix(fx): close edge M2M path on fx.convert; exclude M2M from fx operator rules

JiRaska:
## Summary

Second batch (#3734). FX's three operator rules were role-only, so both M2M service accounts rode them — and the role_action_matrix grants `fx.convert` (the money-moving conversion) to ROLE_OPERATOR, which `service-account-openbank-edge` carries. A customer-facing proxy could execute conversions, trigger CNB ingests, and decide four-eyes approvals on the FX rail.

Both M2M clients are **verified read-only** — edge proxies the published rate sheet; the shared client serves ledger's FX revaluation and agent-service's MCP read tools — so no identity-scoped write grant exists to preserve; their reads stay via `service-fx-edge-m2m` / `service-fx-shared-client-m2m`.

## Type of change

## Linked issues

Refs #3734, #3698, #3740, #3742, #3744, #3748, #3750, #3755, #3758

## Changes

1. `operator-fx-write`, `operator-fx-trigger`, `operator-fx-approval-decide` exclude `service-account-*` principals.
2. New `prohibited` veto on `fx.convert` for the edge — the **only** fx write in the matrix grant, so the only one `matrix-allows` re-admits past the exclusion. `fx.trigger` / `fx.approval.decide` are absent from the matrix grant; the exclusion closes them outright (no veto needed, verified by test).
3. Heredoc → standalone `fx_rest_ext.rego` + new `fx_rest_ext_test.rego` (11 tests; matrix mock mirrors the real `data.rules.authz.role_action_matrix[role].grant` shape).
4. Generator patched; bundle + checksum regenerated (`254b8a7310997ee4`, idempotent); threat model updated (ADR-0030 D2).

## Verification evidence

- `opa test` trio: **11/11 PASS**. Falsification: stripping the exclusions → **4/11 red**; stripping the veto → **2/11 red** (edge convert denial + veto-fires go green-to-red via the matrix grant).

- [ ] Confirm caller audit: edge calls only `fx.{read, list}` (`CustomerEdgeResource` rate-sheet proxy); ledger reads via `FxServiceClient`; agent-service MCP tools are read-only.

## Security checklist

- [x] No secrets / PII. [x] Auth change → `security-review-required`. [x] No policy logic removed; read-only M2M identity grants preserved, write paths closed.

## Compliance impact

- [x] DORA — access control on the FX rail narrowed to named identities; money-moving conversions no longer reachable by M2M proxies.


JiRaska:
**Review: same hole as #3764 — `service-account-openbank-services` can still call `fx.convert`.**

**1. `openbank-infra/gitops/components/fx-service/fx_rest_ext.rego` — the veto covers only the edge identity.**

Trace for `principal.id = service-account-openbank-services`, `action = fx.convert`:
- `operator-fx-write` — correctly excluded by the new `not startswith(input.principal.id, "service-account-")`. ✅
- `service-fx-shared-client-m2m` — scoped to `{fx.read, fx.list}`, does not fire. ✅
- `matrix-allows` (`rest.rego:151-155`) — `rules-opa-data.yaml:267` grants `fx.convert` to ROLE_OPERATOR. **Fires.**
- The new `prohibited` clause requires `input.principal.id == "service-account-openbank-edge"`. **Does not fire.**
- Base `prohibited` (`rest.rego:485-491`) is inert (fact A below); `operator-fx-write` is in `deferred:` anyway, and `matrix-allows` is never a listed reason.

Net: `allowed_reasons = {"matrix-allows"}` → `rest.allow` true. FX conversion — a real money-path action with a live GL effect — stays reachable by the shared backend client.

`fx.trigger` and `fx.approval.decide` **are** genuinely closed for both service accounts (neither is in the matrix, so the rule exclusion alone suffices). The gap is exactly the one action the matrix grants, which is also the one the PR title names.

**2. The shared-client test is doubly weak — wrong form AND wrong action.** `test_shared_no_operator_rule_on_trigger` asserts a *reason* is absent (not `not rest.allow`), and it does so on `fx.trigger` — the action that is **not** matrix-granted, i.e. the one already closed. The single action where the residue lives, `fx.convert`, has no shared-principal test at all. Add `test_shared_denied_convert { not rest.allow with input as {"principal": shared, "action": "fx.convert"} with data.rules as rules_mock }`; it is red today.

Fix, if the caller audit supports it (this file's own comments say the shared client's legitimate fx surface is read/list): widen the veto predicate to `startswith(input.principal.id, "service-account-")`, the shape **#3698** uses.

**Verified clean:** `service-fx-shared-client-m2m` and `service-fx-edge-m2m` are correctly read-scoped and preserved; `ROLE_PAYMENTS` on `operator-fx-write` matches the matrix's `ROLE_PAYMENTS` grants; the `not startswith(...)` exclusion is applied consistently to all three operator rules (a partial application would have been the easy miss here); heredoc→file extraction faithful; no `principal.type == "SERVICE"` rule.

---

PR #3761, opened 2026-08-05
fix(lending): close edge M2M path on the eight matrix-granted lending writes

JiRaska:
## Summary

Second batch (#3734). `operator-lending-write` was role-only over the whole `lending.*` family (20+ actions), and the role_action_matrix grants **eight** lending writes to ROLE_OPERATOR — approve, collateralDecide, collateralRegister, create, disburse, repay, reschedule, writeoff — all reachable by `service-account-openbank-edge`, a customer-facing proxy. A compromised edge token could approve and disburse loans.

The edge's verified legitimate access is exactly `lending.intake` (customer loan application, `CustomerEdgeResource.applyForLoan`) — already identity-scoped in `edge-customer-intake`. No backend M2M caller exists (anacredit consumes Kafka events; ledger posting is outbound from lending).

## Type of change

## Linked issues

Refs #3734, #3698, #3740, #3742, #3744, #3748, #3750, #3755, #3758, #3760

## Changes

1. `operator-lending-write` excludes `service-account-*` principals.
2. New `prohibited` veto: edge denied on the eight matrix-granted writes — needed in addition to the exclusion because `matrix-allows` grants them to ROLE_OPERATOR without consulting this rule. The ~12 non-matrix writes (`acceleration.execute`, `advance`, `default.mark`, `approval.decide`, …) are closed by the exclusion alone (verified by test).
3. Heredoc → standalone `lending_rest_ext.rego` + new `lending_rest_ext_test.rego` (16 tests; matrix mock mirrors the real `data.rules.authz.role_action_matrix[role].grant` shape).
4. Generator patched; bundle + checksum regenerated (`4aab4e2b7d893d33`, idempotent); threat model updated (ADR-0030 D2).

## Verification evidence

- `opa test` trio: **16/16 PASS**. Falsification: stripping the exclusion → **4/16 red**; stripping the veto → **9/16 red** (8 edge write denials + veto-fires go green-to-red via the matrix grant).

- [ ] Confirm caller audit: customer-edge calls only `lending.intake` (+ reads); no in-repo backend REST caller of lending.

## Security checklist

- [x] No secrets / PII. [x] Auth change → `security-review-required`. [x] No policy logic removed; identity-scoped intake grant preserved, role-only + matrix paths closed.

## Compliance impact

- [x] DORA — access control on the lending rail narrowed to named identities; loan approval/disbursement no longer reachable by a customer-facing M2M proxy.


JiRaska:
**Review: same hole as #3764 — the shared client is still admitted to all eight lending writes.**

**1. `openbank-infra/gitops/components/lending/lending_rest_ext.rego:150-166` — the veto is pinned to one identity; `service-account-openbank-services` keeps every write this PR claims to close.**

Trace for `principal.id = service-account-openbank-services`, `action = lending.disburse`:
- `operator-lending-write` — correctly excluded now by the new `not startswith(input.principal.id, "service-account-")`. ✅
- `matrix-allows` (`rest.rego:151-155`) — `rules-opa-data.yaml:221-292` grants ROLE_OPERATOR `lending.approve`, `collateralDecide`, `collateralRegister`, `create`, `disburse`, `repay`, `reschedule`, `writeoff`. The rule needs only `type == "HUMAN"` + a granted role. **Fires.**
- `prohibited` (this PR, ext line ~160) — requires `input.principal.id == "service-account-openbank-edge"`. **Does not fire.**
- Base `prohibited` (`rest.rego:485-491`) — cannot fire (see fact A below); and even with the register populated, `operator-lending-write` is in `deferred:`, and `matrix-allows` is not a listed reason, so the `every` clause would be false regardless.

Net: `allowed_reasons = {"matrix-allows"}`, `not prohibited` holds, `rest.allow` is true. The shared backend client — which nearly every service authenticates as — can approve applications, disburse, decide collateral, repay, reschedule and write off loans, exactly the set the PR title says is closed.

Your own caller audit is what makes the fix safe and one line: the ext states the only in-repo M2M caller is the edge on `lending.intake`. So the veto predicate can be widened to `startswith(input.principal.id, "service-account-")` — which is precisely what the sibling **#3698** does for interest, and is the shape I would take as the family's template.

**2. `lending_rest_ext_test.rego:118-124` is one assertion short of exposing it.** `test_shared_no_operator_rule_on_create` asserts the *reason* `operator-lending-write` is absent, not `not rest.allow`. It passes today with the hole present. Rewrite as `not rest.allow with input as {"principal": shared, "action": "lending.disburse"}` and it goes red immediately. There is currently no test in the file that evaluates `rest.allow` for the shared principal at all — the 11 `test_edge_denied_*` cases all use the strong form, so the asymmetry looks accidental rather than intended.

**Verified clean:** the exclusion added to `operator-lending-write` is correct and complete for the ~12 non-matrix lending writes; `lending.intake` is correctly left out of the veto and `edge-customer-intake` still fires for it (`test_edge_may_intake_via_identity_rule`); the heredoc→file extraction is faithful; checksum `4aab4e2b7d893d33` matches between `lending-opa-bundle.yaml:12` and the pod-roll annotation in `lending-service.yaml:37`; `rules_mock` mirrors the real `data.rules.authz.role_action_matrix[role].grant` shape; no `principal.type == "SERVICE"` rule anywhere in the diff (the only occurrences are prose correctly explaining why it can never fire).

---

PR #3764, opened 2026-08-05
fix(billing): close edge M2M path on all three billing writes

JiRaska:
## Summary

Second batch, final row (#3734). `operator-billing-write` was role-only over the whole `billing.*` family, and the role_action_matrix grants **all three** billing writes — `post`, `reverse`, `approval.decide` — to ROLE_OPERATOR, which `service-account-openbank-edge` carries. A customer-facing proxy could post fees, reverse them, and decide four-eyes approvals.

Caller audit (2026-08-05): **billing has no in-repo M2M caller at all** — no edge URL, no backend REST client (account-service's billing-discovery read is INBOUND from billing; product-catalog fees are read via catalog). So no identity-scoped grant needed preserving; the tightening is pure subtraction.

## Type of change

## Linked issues

Refs #3734, #3698, #3740, #3742, #3744, #3748, #3750, #3755, #3758, #3760, #3761

## Changes

1. `operator-billing-write` excludes `service-account-*` principals.
2. New `prohibited` veto: edge denied on `{billing.post, billing.reverse, billing.approval.decide}` — needed in addition to the exclusion because `matrix-allows` grants all three to ROLE_OPERATOR without consulting this rule.
3. Heredoc → standalone `billing_rest_ext.rego` + new `billing_rest_ext_test.rego` (8 tests; matrix mock mirrors the real `data.rules.authz.role_action_matrix[role].grant` shape).
4. Generator patched; bundle + checksum regenerated (`d5fb31769906ca66`, idempotent); threat model updated (ADR-0030 D2).

## Verification evidence

- `opa test` trio: **8/8 PASS**. Falsification: stripping the exclusion → **2/8 red**; stripping the veto → **4/8 red** (3 edge write denials + veto-fires go green-to-red via the matrix grant).

- [ ] Confirm caller audit: no edge or backend REST caller of billing-service.

## Security checklist

- [x] No secrets / PII. [x] Auth change → `security-review-required`. [x] No policy logic removed; no M2M grant exists to preserve, write paths closed outright.

## Compliance impact

- [x] DORA — access control on fee posting/reversal narrowed to named human operators; four-eyes approval decisions no longer reachable by a customer-facing M2M proxy.


JiRaska:
**Review: 3 findings, 1 worth acting on before merge.**

1. **`openbank-infra/gitops/components/billing/billing_rest_ext.rego:50` — the veto closes the edge identity only; `service-account-openbank-services` still reaches all three billing writes.** Traced end to end on `origin/main`: `rules-opa-data.yaml:235-237` grants `billing.approval.decide`/`billing.post`/`billing.reverse` to ROLE_OPERATOR, which the shared client's service account carries. The new exclusion does strip `operator-billing-write` from it — but the base shared-identity guard (`rest.rego:485`) only fires when **every** admitting reason is in `shared_m2m_write_prohibition.reasons`, and (a) `operator-billing-write` is explicitly in that key's `deferred:` map, not the register, and (b) `matrix-allows` is never a listed role-only reason, so the `every` clause is false and `prohibited` does not fire. Net: shared client posts/reverses fees and decides four-eyes approvals. Not a regression this PR introduces — but the PR's own caller audit ("billing has no in-repo M2M caller at all") is exactly the evidence that makes closing it safe in the same subtraction. Broadening the veto predicate to `startswith(input.principal.id, "service-account-")` would do it.
   `billing_rest_ext_test.rego:70` is one assertion short of exposing this: `test_shared_no_operator_rule_on_reverse` asserts the *reason* is absent, not `not rest.allow`. Change it to the `rest.allow` form and it goes red today.

2. **The 8-test suite does not run in CI.** `openbank-infra/opa/build-bundle.sh:37` runs `opa test` over `openbank-infra/opa/policies` and `openbank-libs/governance/policies` only — nothing walks `gitops/components/**`, and no `gates.yaml` entry does either. The falsification evidence (2/8 and 4/8 red) therefore lives only in the PR body, and the file is free to drift from the bundle it describes. Pre-existing for the other 10 `*_rest_ext_test.rego` files too, so this is a follow-up issue rather than a blocker on this PR.

3. **Asymmetry worth a comment in the file:** the exclusion is broad (`not startswith(id, "service-account-")`, all service accounts) while the veto is narrow (one hardcoded id). A reader will assume both cover the same set. Finding 1 is why they don't.

Verified without finding a problem: the checksum `d5fb31769906ca66` is consistent across `billing-opa-bundle.yaml` and the pod-roll annotation in `billing-service.yaml`; the heredoc→file extraction is byte-identical to the old inline text apart from the two intended additions; `prohibited` is genuinely load-bearing (base `rest.rego:50` gates the `allow` head on `not prohibited`); `input.action in {...}` is valid under the imported `rego.v1`; the threat-model change-log entry is present as ADR-0030 D2 requires.

*Automated agent review — not an approval. `openbank-billing-service` is money-path: the 2-human-approval rule still applies.*

Issue #3737, opened 2026-08-05
Agent swarm coordination (ADR-0244 follow-ups)


JiRaska:
## Summary

ADR-0244 (**merged** 2026-08-05) defines real-time agent swarm coordination: several chartered
agents working the same case concurrently — joining a running investigation via
OPA-gated Temporal signals, commenting and correcting mid-flight — coordinated by a
dedicated `case-coordinator` agent, with every case ending in exactly one HITL
proposal.

## Scope (per ADR-0244 follow-ups)

- [x] `agents.yaml` schema extension: case budgets + swarm-participation capability, `case-open` capability, per-case-class ceilings, contested-rate thresholds (ADR-0244 D6/D9) — **done** (#3771, gate `agent-case-schema` #4009)
- [x] `case-coordinator` chartered agent (AGPL agent-plane module, `rules.yaml: agpl_modules`) — **charter + parity doc done** (#3771); runtime module follows Phase 1
- [x] Temporal case workflow pattern (signals, deterministic pre-emption via `superseded-by-evidence`)
- [ ] Kill-switch → Temporal case-cancellation wiring (ADR-0244 D7)
- [ ] Eval scenarios for convergence quality under the ADR-0148 evals gate
- [x] Admin-ui swarm thread view ADR → **ADR-0246**, PR #3983 (**merged 2026-08-07**; implementation slice follows acceptance)

## Context

Builds on ADR-0202 (agent-to-agent pipeline over proposal events, MCP and Temporal),
which covers hand-offs but not concurrent same-problem collaboration. Governance
invariants from ADR-0031 (agents propose, governance disposes) are preserved.

## Delivery plan (agreed 2026-08-07)

Thinnest end-to-end slice first (one case class, one signal path, read-only UI), then widen. Each phase is independently mergeable/revertable PR(s).

- [x] **Phase 0 — charter + schema (no runtime code)** — **done 2026-08-08** (#3771 schema+charter, #4009 validation gate, #4006 CLAUDE.md knowledge capture; sub-issue #3998 closed): `agents.yaml` case-coordination schema extension (per-case-class budgets, `case-open` / swarm-participation capabilities, concurrent ceilings, contested-rate circuit-breaker threshold); `case-coordinator` charter (ADR-0244 D3) + `docs/agents/case-coordinator.md` (ADR-0156 parity); derived artifacts regenerated (eu-ai-act, OPA bundles); schema-validation gate
[...]

PR #3771, opened 2026-08-05
feat(governance): agent swarm case-coordinator charter and case-budget schema (ADR-0244)

JiRaska:
Adds the governance scaffold for ADR-0244 agent swarm coordination — the first conceptual slice before building the `case-coordinator` module.

## Changes

- `openbank-libs/governance/agents.yaml`
  - New `case_classes` schema with default and per-class budgets (tokens, contributions, wall-clock), concurrency ceilings, and contested-rate thresholds.
  - New `case-coordinator` charter with `case.open`, `case.coordinate`, `case.synthesize`, `case.preempt` capabilities; HITL proposal-only output; explicit audit capture set.
- `openbank-libs/governance/prompts/registry.yaml`
  - Registers `case-coordinator` as `pending` (ADR-0148 backlog) until the module and LLM wiring exist.
- `docs/agents/case-coordinator.md`
  - Narrative charter: mission, why it exists, human oversight, and known gaps.

## Not in this PR

- The actual `openbank-case-coordinator-agent` module (AGPL registration in `rules.yaml` deferred until the module exists, to keep the license-boundary check self-consistent).
- Temporal workflow, signal handlers, OPA capability check, eval scenarios — all follow-up PRs.

## Type of change

- [x] feat

## Linked issues

Refs #3737

### DIFF
--- a/docs/agents/case-coordinator.md
+++ b/docs/agents/case-coordinator.md
@@ -1,0 +1,49 @@
+---
+id: case-coordinator
+plane: control
+adr: ADR-0244
+---
+
+# case-coordinator
+
+## Mission
+
+Owns a single running *case* — a long-running Temporal workflow opened for one disposition target
+(an alert, incident, or proposal lineage). The coordinator invites other chartered agents to
+contribute mid-flight via OPA-gated Temporal signals, enforces the per-case-class budget and
+deadline, judges when the swarm has converged, and emits **exactly one** human-in-the-loop proposal
+per case. It never writes to a business record directly; the only output is a proposal into the
+existing ADR-0031 approval queue.
+
+## Why this agent exists
+
+ADR-0202 covers agent-to-agent *hand-offs*: one agent finishes, another takes over. That pattern
+does not cover several agents working the *same* problem concurrently, correcting each other
+mid-flight, and converging on a better answer faster. The case-coordinator is the boundary that
+keeps that concurrency bounded and safe: one durable workflow owns the case, the budget is declared
+up front, pre-emption is deterministic, and the swarm cannot silently stall or explode in cost.
+Without a dedicated coordinator, the agent that detected a problem would also judge when it is
+solved — a conflict of interest that ADR-0244 D3 explicitly rejects.
+
+## Human oversight
+
+- `every: proposal` — every synthesized case outcome is a proposal into the HITL queue; a human
+disposes before any state changes.
+- `approver_must_differ_from: author` — segregation of duties is preserved even inside the swarm.
+- Case budgets, deadlines, and contested-rate thresholds are declared in `agents.yaml` under
+`case_classes`, so they are reviewable and versioned with the same governance rigor as tool tiers.
+
+## Known gaps (these are real, not aspirational)
+
+- **There is no `openbank-case-coordinator-agent` module yet.** This charter is the governance
+scaffold; the Quarkus service, Temporal workflow, signal handler, and OPA-gated capability check
+are follow-up work tracked by ADR-0244 delivery notes.
+- **No other charter currently grants `case.join` or `case.contribute`.** The coordinator is the only
+swarm participant defined today; other agents will gain swarm capabilities only after the signal
+contract and policy gate are implemented.
+- **The kill-switch-to-Temporal-cancellation wiring is not built.** The global and per-agent kill
+switches can halt agent runtimes today, but canceling an in-flight case workflow requires a new
+hook that does not yet exist.
+- **The LLM synthesis step is undefined.** Whether the coordinator uses an LLM for convergence
+judgement or a deterministic rule set is a downstream decision; this charter reserves the
+`case.synthesize` capability and leaves the implementation open.
--- a/openbank-libs/governance/agents.yaml
+++ b/openbank-libs/governance/agents.yaml
@@ -174,10 +174,58 @@ tool_tiers:
     - secrets.read.raw
 
 # ---------------------------------------------------------------------------------------
+# Case classes for agent swarm coordination (ADR-0244). A case is one long-running
+# Temporal workflow opened for a single disposition target. Default budgets and
+# convergence thresholds live here; per-class overrides are merged over default.
+# Every case names exactly one class. The case-coordinator charter owns convergence
+# judgement and budget enforcement; other charters may be granted case.join (and later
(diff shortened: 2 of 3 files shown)


JiRaska:
**Review: 4 findings, 1 blocking.** Automated agent review — not an approval.

Findings:

1. **BLOCKING — `openbank-libs/governance/agents.yaml`, `case-coordinator.tools`: this is the only charter in the fleet that puts `tier: execute` in an `allow` block, and its execute *deny* is an enumerated blocklist where every peer uses deny-all.** Measured across all charters in the file: `tool_tiers` defines exactly three tiers — `read`, `write_proposal`, `deny`. There is no `execute` tier. Every other agent denies `- tier: execute / resources: ["*"]`. This charter instead writes:
   ```
   allow:  - tier: execute   resources: [case-workflow-signal]
   deny:   - tier: execute   resources: [money.*, gh.pr.*, secrets.*]
   ```
   Anything under `execute` that is not one of those three prefixes is neither denied nor enumerated — allow-by-omission, in the one charter that introduces a tier the tier registry does not define. Reviewed as written rather than as intended, this is the widest tool grant in the file. Either add `execute` to `tool_tiers` with an explicit member list (so `case-workflow-signal` is a *named* tool, not a free-text resource), or keep `deny: - tier: execute resources: ["*"]` as the peers do and rely on the allow entry for the one exception.

2. **The separation-of-duties invariant the charter states in prose is contradicted by its own capability grant.** The description says "The coordinator never coordinates a case it opened; who detected is never who converged" — but `case_capabilities` grants `case.open`, `case.coordinate`, `case.synthesize` and `case.preempt` to this single charter, and the comment adds "The coordinator is the only charter that may open, coordinate, and synthesize a case." No other charter carries any `case.*` capability, so there is no second party for the claimed split. The invariant, if it is real, lives in per-*instance* logic that the schema cannot express and nothing in this PR enforces.

3. **`case_classes` and `case_capabilities` have zero consumers anywhere in the repo.** Grepped `*.py`, `*.sh`, `*.rego`, `*.kt`, `*.ts/tsx`, `*.mjs` outside `node_modules`: nothing reads either key. That is fine for a charter-first change, but it means the budgets below are declarations, not limits — worth saying so in the block comment so a later reader does not treat `tokens_per_case: 500000` as enforced. It also matters for finding 1: an undefined tier is inert today and becomes load-bearing the moment something reads it.

4. **The two budget systems have no stated precedence.** `case_classes.default.budget.tokens_per_case: 500000` with `max_open_cases: 10` sits next to `case-coordinator.limits.tokens_per_run: 50000 / runs_per_day: 48`. Which one binds, and whether a case's tokens are charged against the coordinator's daily cap, is undefined. `fraud-investigation` at 1M tokens × 5 open cases is a 5M-token ceiling with nothing reconciling it against 2.4M/day at the charter layer.

Also worth checking before merge (registry drift, not a security finding): `docs/compliance/eu-ai-act.md` and `openbank-admin-ui/ai-governance-snapshot.json` are both DERIVED from `agents.yaml`, and neither mentions `case-coordinator`. This is the #2216 shape — a new AI system omitted from the EU AI Act inventory while the check that would say so is advisory over a generated artifact. #3612 in this batch updates both; this PR updates neither.

Verified clean: the charter is proposal-only in the way that matters (`write` denied on `"*"`, `requires_human: every: proposal`, `approver_must_differ_from: author`, `record: reason`); `kill_switch: true` is set; `data_scope.pii: masked` and the read scope is metadata-only (`agent-proposals-readonly` explicitly excludes raw payload); `plane: control` and `enforcement: block` match the fleet convention for oversight agents; the charter doc `docs/agents/case-coordinator.md` is present, which the charter-parity gate requires.

Needs human security review regardless — this adds a new autonomous agent charter with an execute-tier grant.

JiRaska:
## Withdrawing the blocking finding — it was wrong

An earlier automated review on this PR claimed case-coordinator is *"the only charter in the fleet granting `tier: execute` in `allow`, for a tier `tool_tiers` does not define, with an enumerated execute deny-list where every peer uses `["*"]`"*. I checked it before acting on it. **The "only charter" part is false**, and it was the part that made it blocking.

Measured against this branch's `agents.yaml` (16 agents, 9 of which declare tiers):

```
defined tool_tiers : deny, read, write_proposal

charters using UNDEFINED tiers: 9 of 9
   finops-agent                   ['execute', 'write']
   devops-agent                   ['execute', 'write']
   control-liveness-sentinel      ['execute', 'write']
   governance-auditor             ['execute', 'write']
   release-steward                ['execute', 'write']
   docs-truth-agent               ['execute', 'write']
   authz-policy-auditor           ['execute', 'write']
   flaky-test-hunter              ['execute', 'write']
   case-coordinator               ['execute', 'write']
```

Every charter that declares tiers uses `execute` and `write`, and none of them is in `tool_tiers`. case-coordinator follows the fleet convention exactly — it is not an outlier, and there is nothing here for this PR to fix that its eight peers have not already been doing.

**The real observation, which is pre-existing and not this PR's to carry:** `tool_tiers` lists three tiers while the charters use five. Either the list is stale, or `execute`/`write` are meant to be rejected and nothing rejects them. Worth its own issue, because the `deny` entries are the ones that read as protection —

```yaml
deny:
  - {tier: write,   resources: ['*']}
  - {tier: execute, resources: ['money.*', 'gh.pr.*', 'secrets.*']}
```

— and a deny keyed on a tier the schema does not know is worth confirming actually denies. I could not establish a runtime consumer that would break: `McpToolRegistry` enforces through charter *capabilities* with deny-by-default, and `tool_tiers` appears in `agentCharters.ts` (admin-ui display) and in ADR prose. So my reading is that this is a schema/doc inconsistency rather than a live hole — but I am stating that as a reading, not a verified fact, because I did not trace every consumer.

**Also worth recording: my first measurement of this was itself wrong**, and for a reason worth knowing in this repo. I read the branch through `FETCH_HEAD`, which is not stable here — concurrent agents and the scheduled drain run `git fetch` continuously, so `FETCH_HEAD` was repointed between two of my commands and one run reported case-coordinator as absent from `agents.yaml` entirely. Pin an explicit ref (`git update-ref refs/tmp/<name> <sha>` from `git ls-remote`) before reading a branch in this repo; `FETCH_HEAD` is a race.

No blocking finding stands against this PR from my side. It remains `governance(...)` scope and needs human review on its merits.

PR #3772, opened 2026-08-05
feat(case-coordinator-agent): create module and register in rules.yaml (ADR-0244)

JiRaska:
Creates the `openbank-case-coordinator-agent` module and wires its ADR-0148 prompt registry entry (ADR-0244 follow-up).

What changed:
- New Quarkus service module `openbank-case-coordinator-agent` with:
  - `LICENSE` (AGPL-3.0-only)
  - `build.gradle.kts` using the fleet `openbank.quarkus-service` convention; packages the registered prompt at build time
  - `governance.yaml` manifest
  - `version.txt` at `0.1.0`
  - `application.yaml` with reactive Postgres + Flyway + Temporal placeholders + model endpoint config
  - `V1__init_casecoordinator.sql` migration for case workflow metadata
  - `CaseCoordinatorResource` with a `/api/v1/case-coordinator/status` health endpoint
  - `CaseCoordinatorConfig` and `LlmGatewayProducer` (ADR-0174 seam)
  - `CaseCoordinatorLlmPort` + `CaseCoordinatorLlmAdapter` that loads `system.v1.md` from the prompt registry
  - `CaseCoordinatorBootSmokeIT` + `PostgresTestResource` boot smoke test
  - `CaseCoordinatorLlmAdapterTest` asserting the registered prompt is sent byte-for-byte and the gateway degrades to null
- Registered the module in `release-please-config.json` and `.release-please-manifest.json`
- Registered the module as AGPL-3.0-only in `openbank-libs/governance/rules.yaml`:
  - `agpl_modules`
  - `license_denylist_exceptions` paths
  - dependency boundary rule string
- Added `openbank-libs/governance/prompts/case-coordinator/system.v1.md`
- Updated `openbank-libs/governance/prompts/registry.yaml` to mark `case-coordinator` as registered with `system.v1`

Local verification:
- `./gradlew :openbank-case-coordinator-agent:build` passes (detekt, ktlint, tests, kover)
- `python3 .github/scripts/check-license-headers.py` passes
- `python3 .github/scripts/check-prompt-registry.py` passes (case-coordinator now registered; 6 remaining pending charters)
- `bash .github/scripts/check-agent-charter-registry.sh` passes
- `bash docs/adr/gen-index.sh && bash .github/scripts/check-adr-registry.sh` passes

Refs #3737
Depends on #3771 (agents.yaml case-coordinator charter and case-budget schema)


JiRaska:
## Automated agent review — release registration is correct; 2 findings, neither blocking

**Verdict: cleanest new-service scaffold in this batch.** Release wiring and the AGPL declaration are both done properly, which is where new-module PRs usually fail here.

### Verified clean — release registration is in lockstep (all three axes)
- `openbank-case-coordinator-agent/version.txt` → `0.1.0`
- `.release-please-manifest.json` → `"openbank-case-coordinator-agent": "0.1.0"`
- `release-please-config.json` → `{"component": "case-coordinator-agent", "exclude-paths": ["openbank-case-coordinator-agent/src/test"]}`

All three present and agreeing, so the `release-registration consistency` gate should pass and the first release will not break. Note for whoever merges: if `main` moves under this branch, re-count the manifest entries after any merge — a git-clean merge silently drops one of two neighbouring additions to that map, which is how the release manifest went 56→55 on 2026-08-02.

### Verified clean — multi-license declaration is complete
The module carries `SPDX-License-Identifier: AGPL-3.0-only` and the PR updates **all three** places `rules.yaml` tracks it: `dependencies.*.paths`, `agpl_modules` (with the `# ADR-0244` marker), and the `rule:` string's project list. That last one is the part usually missed — `check-license-headers.py` enforces `agpl_modules`, but a stale `rule:` string drifts silently. Good.

### 1. No authz posture beyond `@RolesAllowed` — fine today, worth stating explicitly

`CaseCoordinatorResource` exposes exactly one endpoint:

```
@GET @Path("/status")
@RolesAllowed("ROLE_ADMIN", "ROLE_VIEWER")
fun status(): Status
```

No `@Authorize`, no OPA sidecar, no `*_rest_ext.rego`, no gitops component directory. For a status-only slice that is proportionate and I would not ask for a sidecar now. But `governance.yaml` declares upstream lineage on `temporal` and `agent-service` ("policy decision point for tool-call authorization"), i.e. this module is designed to make tool calls — the moment the first real endpoint lands it needs `@Authorize` actions *and* a matching rego rule in the same PR, not after. An `@Authorize` action with no policy rule denies once enforcement is on; an endpoint with neither is unprotected. Worth a line in the PR body or a follow-up issue so slice 2 does not have to rediscover it.

### 2. No `openapi.yaml`, no gitops deployment, no `Dockerfile`

The module ships no `src/main/resources/openapi.yaml`. Per ADR-0048 the API-contract version is an independent axis from `version.txt`; a released component exposing a REST path with no committed spec means the contract gate has nothing to diff and the first API change classifies against nothing. Also no `openbank-infra/gitops/components/case-coordinator/` — so this is repo-only for now and nothing deploys, which is consistent with "create module" but means the boot path is unexercised outside `CaseCoordinatorBootSmokeIT`. Flagging as scope, not as a defect: if slice 1 is deliberately non-deploying, say so in the PR body.

### Also
- `LlmGatewayProducer` + `CaseCoordinatorLlmAdapter`: check the adapter does not construct anything native or fail-prone in a field initializer on an `@ApplicationScoped` bean — a `Throwable` from a field initializer fails bean construction and then fails *every* injection point of that bean, including endpoints that never use it (the `OnnxFraudModel` / `#3376` shape, where a read-only queue endpoint 500'd on every call). A nullable field assigned in a `try` that catches `Throwable` is the house pattern.
- Branch is `DIRTY`. Rebase before merge, and re-verify the manifest entry count afterwards per the note above.

---
*Automated agent review — this is **not** an approval and no approval is implied.*

PR #3797, opened 2026-08-06
feat(delegation): snapshot counterparty display names onto a grant

### LINKED ISSUE #3604: Delegation counterparty is an unidentifiable UUID on the accept screen — the edge has no permitted party lookup
## What the customer sees

In "Sdíleno se mnou" the counterparty is a **truncated UUID**. Someone is asked to accept access to
their money from `33333333`. That is the one screen where the identity is the whole decision.

The app carries a display name only for a grant **it created itself**, smuggled through the
free-text `note` field — which does nothing for the grantee, who is the person that actually needs
it (openbank-app#360 documents this).

## Why it is not a one-liner

`DelegationResponse` returns `grantorPartyId` / `granteePartyId` as bare UUIDs, and
**customer-edge has no permitted way to turn a party id into a name.** The party OPA bundle allows
`service-account-openbank-edge` exactly two actions:

```rego
input.action in {
    "party.consent.update",
    "party:resolve",
}
```

`party:resolve` is **not** an id lookup — it is the ADR-0072 blind-index dedup gate, resolving by
Czech RČ. There is no `party.read` for the edge, and that bundle's own comment says a caller goes
in "only with matching evidence (a real client method + call site), never speculatively".

So the choice is a design decision, not an omission.

## Options

1. **Snapshot the display name on the grant, at offer time.** delegation-service already calls
   pid-service for eligibility, so it can store `grantorName` / `granteeName` on the aggregate.
   No runtime lookup, no new rego, and the name shown is the one that was true when the grant was
   made — which is arguably what an authorisation record should say, since a later rename must not
   silently rewrite who agreed to what. Cost: a stale name after a legitimate rename, and PII at
   rest in one more service (it is already PII-adjacent — it holds who may touch whose money).

2. **Resolve at the edge, scoped to grants the caller is a party to.** Needs `party.read` for
   `service-account-openbank-edge`, which grants the edge read on ANY party — much wider than the
   need. It is not an enumeration oracle *as used* (the caller already holds the relationship), but
   the grant is what gets reviewed later, not the usage.

3. **A narrow purpose-built endpoint** — `GET /api/v1/parties/{id}/display-name` with its own
   action, allowed to the edge and returning only a display name. More moving parts; the least
   over-granted.

Preference is (1): it needs no new authority at all, and a name captured at the moment of consent
is more correct for this record than a live one. (3) is the right answer if the name must track
renames.

## Not just cosmetic

Asking someone to approve access to an account from a party they cannot identify is the failure the
plain-language summary requirement (ADR-0150) exists to prevent. Until this is fixed, the accept
screen shows a truthful sentence about an unidentifiable counterparty.

Refs openbank-app#360, ADR-0232 D6, ADR-0150

JiRaska:
Closes #3604

## What the customer saw

Someone asked to accept a delegation — to hand another party authority over their money — was shown a truncated UUID as the counterparty. That is the one screen where the identity *is* the decision, so the consent could not be informed. Product defect and compliance defect at once (ADR-0150).

## Which layer was actually missing

Not the data, and not a modelling gap at the source. pid-service already returns `coreAttributes.givenName` / `familyName` on **the exact call delegation-service already makes at offer time** for the ADR-0232 D5 eligibility gate. Three things were missing downstream of it:

| Layer | State before |
|---|---|
| `PidPartyResponse` (delegation's pid client) | Did not declare the fields; `@JsonIgnoreProperties` dropped them at the parser |
| `DelegationGrant` / `delegation_grants` | Nowhere to put a name |
| `DelegationResponse` | Returned bare UUIDs |

So this is option **(1)** from the issue — snapshot the name on the grant at offer time — and it needs no new call and **no new authority anywhere**.

## Why snapshot rather than resolve

- **customer-edge gains nothing.** Its party OPA grants stay `party.consent.update` + `party:resolve` (the ADR-0072 blind-index dedup gate, which is not an id lookup). Option (2) would have given the edge `party.read` on *any* party — scoped in usage, unscoped in the grant, and the grant is what gets reviewed. Option (3) would have added a new lookup surface for a value that does not need to be live.
- **It is the more correct record.** An authorisation record should say who agreed to what *at the moment they agreed*. A later rename must not silently rewrite that.

## Disclosure — what the accepting party can now see

The names ride the existing `DelegationResponse`, whose read paths are already party-scoped (threat-model T8): a grant the caller is not party to is **404**, a list for another party is **403**. So they reach only the two parties to the grant and role-gated bank staff. Both parties are, by construction, already identified to each other by the grant itself, so the name discloses nothing the relationship does not.

Threat model gains **T13**, with the residuals stated rather than implied away:
- bank staff (`ROLE_OPERATOR`/`ROLE_ADMIN`) already read any grant and now see the names too — same scope, wider PII surface;
- a snapshot goes stale on a legitimate rename. Deliberate, but it means the label is **not an identity assertion**.

Only `givenName`/`familyName` are materialised; the rest of `coreAttributes` (birthdate, birth number, documents, nationality) is still dropped at the parser.

## Nullable everywhere, and never blank

No backfill: existing rows stay `NULL`. Manufacturing a name for a grant consented to without one would put a value into an authorisation record that was never part of it. Every consumer falls back to the party id — and specifically never to an empty string, because a blank chip on a consent screen reads as a name that *failed to load* rather than one never captured. That distinction is asserted in three places (`ResilientPartyEligibilityClientTest`, `DelegationServiceTest`, `counterpartyLabel`).

## Versioning

- **API contract:** `openapi.yaml` `info.version` **1.1.0 → 1.2.0**. Two optional response properties added; nothing existing changed or removed — a backward-compatible addition, hence minor. Read off live `origin/main` immediately before setting it.
- **Release version:** left to release-please (`feat` + a path outside `exclude-paths`).
- **DB:** `V3__delegation_counterparty_names.sql`, two nullable `varchar(200)` columns, rollback note in the file.

## Contract test

The pid consumer pact now pins `coreAttributes.givenName` / `familyName` as `stringType` — the path and the type are the contract, since unlike `status` / `kycLevel` these are not a closed vocabulary the consumer ranks. Values match pid's existing `@State` seed (synthetic).

The replay that actually backstops it is pid-service's `PidPactFolderProviderVerificationTest` — `@PactFolder`-sourced, ungated, and it already satisfies the new expectation from its existing `@State` seed (synthetic "Pact Verifier"), so the provider needs no behaviour change.

**Correction — a claim in the first version of this description was wrong.** It said `services-ci` builds both sides of a changed pact file, so the replay would run on this PR. That mapping exists **only in the push lane**. The PR lane fans out purely on `^<module>/`, so a diff confined to `pacts/` matches no module and the provider is *not* pulled in. Measured on this PR's own first run: `Detect changed services` decided *"per-module fan-out"* and the job matrix was `build (openbank-delegation-service)` alone — pid-service absent. Left unaddressed, the contract would have been replayed only *after* the merge, which is exactly the #2327 shape.

Addressed here by giving `PidPactFolderProviderVerificationTest` the documentation this change genuinely owes it — its `@State` seed's `givenName`/`familyName` are no longer incidental fixture noise but the values the pact matches — which also puts `openbank-pid-service` in the PR's build set, so the replay runs *before* the merge. Verify by reading the job list, not by trusting this paragraph.

That the PR lane cannot see the pact graph the push lane deliberately does is a real CI gap, wider than this PR (its push-lane counterpart states outright: *never under-build the Pact graph*). Not fixed here — editing shared change-detection logic inside a money-path PR is the wrong place for it, and it deserves its own issue.

## Deliberately out of scope

- `DelegationOffered` does **not** carry the names — an event-contract change the projections have no use for.
- **customer-edge is untouched.** It proxies the delegation JSON verbatim and its spec declares no response schema for these paths, so the field flows through with no edge change and no edge version bump.

## Self-assessment

**Verified:** the one local test failure (`DelegationExpirationSweepIT`) was environmental, not this change — a clean re-run of the module is green (70/70), and the failing run logged `Thread blocked for 27714 ms` under a parallel fleet build. Also verified: pid returns the fields on that route (`PartyResponse.coreAttributes`); the provider `@State` seeds them; the read paths are party-scoped (read the guard in `DelegationService.getDelegation` / `listByGrantor` / `listByGrantee`); the edge's party OPA action set; that a changed pact pulls the provider into the build.

**Assumed / not proven here:** nothing exercises the full offer→accept path against a real pid-service in this diff — the client is mocked in unit tests and the pact mock answers what it is told. The provider replay is what closes that, which is why it is the load-bearing half.

### DIFF
--- a/docs/threat-models/openbank-delegation-service.md
+++ b/docs/threat-models/openbank-delegation-service.md
@@ -11,7 +11,11 @@ is now stated in the row rather than implied away.
 ## Assets
 
 - `DelegationGrant` records — who may do what on whose resources (confidential; contract
-  evidence under the ADR-0118 retention schedule).
+  evidence under the ADR-0118 retention schedule). Since issue #3604 a grant also carries the two
+  counterparty **display names**, snapshotted at offer time from the eligibility lookup — so this
+  service now holds personal names at rest, not only party ids. Nothing else about the party is
+  stored: the pid client materialises `givenName`/`familyName` and drops the rest of
+  `coreAttributes` (birthdate, birth number, documents, nationality) at the parser.
 - Enforcement integrity of the whole platform: every product service's delegation
   projection trusts this service's event stream.
 - SCA ceremony integrity (grant + acceptance).
@@ -39,6 +43,7 @@ is now stated in the row rather than implied away.
 | T10 | Object-level grant used as execution channel | Aggregate invariant: EXECUTION capabilities rejected on object resource types; object grants are read-only disclosure. |
 | T11 | Unauthorized revoke / forged revoker | `revokedBy` is derived: a customer revoke acts as the party the edge authenticated and must be the grantor; the query parameter survives only on the role-gated bank path. Previously it was a required, trusted query parameter that both authorised the act and wrote the audit field, so any authenticated caller could revoke any grant and sign it as somebody else. `suspend`/`reinstate` are additionally narrowed to ROLE_OPERATOR/ROLE_ADMIN and reject a customer-scoped call. |
 | T12 | A grant outliving its `validTo` in the product projections | Hourly expiry sweep flips ACTIVE → EXPIRED and enqueues `DelegationExpired`, so projections close their rows. The sweep is a `suspend fun` (rules.yaml: scheduled_methods): as a plain `@Scheduled` method it ran with no Vert.x context, so every reactive Panache call threw HR000068 into a single ERROR line and the sweep had **never expired a grant**. A grant past `validTo` still reads as inactive to anything that computes it, so the exposure was the projections that need the event. |
+| T13 | Counterparty display name used as a party-name oracle, or leaked to a non-party (issue #3604) | The name is **snapshotted onto the grant at offer time**, never resolved at read time, so no lookup surface is created and customer-edge gains **no new authority** — its party OPA grants stay `party.consent.update` + `party:resolve` (the ADR-0072 blind-index dedup gate, not an id lookup). It rides the existing `DelegationResponse`, which T8 already scopes: a customer sees only grants they are grantor or grantee of, a list for another party is 403, and a grant the caller is not party to is 404. Both parties are by construction already identified to each other by the grant itself, so the name discloses nothing the relationship does not. The two alternatives were rejected for exactly this row: granting the edge `party.read` would let it read ANY party (the usage is scoped, the *grant* is what gets reviewed), and a dedicated display-name endpoint would create a new lookup surface for a value that does not need to be live. Residual: role-gated bank staff (`ROLE_OPERATOR`/`ROLE_ADMIN`) can read grants for any party and now see the names too — unchanged in scope from the ids they already saw, but a wider PII surface; and a snapshot goes stale on a legitimate rename, which is deliberate (an authorisation record states what was true at consent) but means the label is not an identity assertion. |
 
 ## Out of scope (tracked as follow-ups)
 
--- a/openbank-admin-ui/src/app/delegations/[id]/page.tsx
+++ b/openbank-admin-ui/src/app/delegations/[id]/page.tsx
@@ -25,6 +25,7 @@ import { EntityChip } from '@/components/entities/EntityChip'
 import {
   DelegationStatusBadge,
   capabilityLabels,
+  counterpartyLabel,
   formatCeiling,
   type Grant,
 } from '@/components/delegations/GrantView'
@@ -86,10 +87,10 @@ export default function DelegationDetailPage() {
               <dd><DelegationStatusBadge status={grant.status} /></dd>
 
               <dt style={{ color: 'var(--text-tertiary)' }}>{t('Udělil', 'Grantor')}</dt>
-              <dd><EntityChip type="party" id={grant.grantorPartyId} /></dd>
+              <dd><EntityChip type="party" id={grant.grantorPartyId} label={counterpartyLabel(grant.grantorName)} /></dd>
 
               <dt style={{ color: 'var(--text-tertiary)' }}>{t('Obdržel', 'Grantee')}</dt>
-              <dd><EntityChip type="party" id={grant.granteePartyId} /></dd>
+              <dd><EntityChip type="party" id={grant.granteePartyId} label={counterpartyLabel(grant.granteeName)} /></dd>
 
               <dt style={{ color: 'var(--text-tertiary)' }}>{t('Typ zdroje', 'Resource type')}</dt>
               <dd>{grant.resourceType}</dd>
--- a/openbank-admin-ui/src/app/delegations/page.tsx
+++ b/openbank-admin-ui/src/app/delegations/page.tsx
@@ -28,6 +28,7 @@ import { EntityChip } from '@/components/entities/EntityChip'
 import {
   DelegationStatusBadge,
   capabilityLabels,
+  counterpartyLabel,
   formatCeiling,
   type Grant,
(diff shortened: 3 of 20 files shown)


JiRaska:
Verification note for whoever reviews this — delegation is money-path, so it needs two approvals, and the one claim worth checking independently is the visibility one.

**The spec says the names are "only ever returned to a party of the grant (the read paths are party-scoped) or to role-gated bank staff".** That holds structurally:

```
@RolesAllowed("ROLE_API", "ROLE_OPERATOR", "ROLE_ADMIN")
@Authorize(action = "delegation.read", resource = "#id")        getById
@Authorize(action = "delegation.list", resource = "#partyId")   listByGrantor
@Authorize(action = "delegation.list", resource = "#partyId")   listByGrantee
```

Every read binds `resource` to the **specific** grant or party, and there is an explicit `isBankOperator()` for the staff path. Worth saying out loud because the opposite shape is a defect this repo has already had: `@Authorize(..., resource = "")` in sanctions meant a bearer token satisfied any check (#3521). This is not that.

The residual dependency is the usual one and is not a fault of this PR: the binding is only as strong as the rego rule behind `delegation.read` / `delegation.list`, and `AUTHZ_ENFORCE` has to be true for it to bind at all. Both are pre-existing conditions on every endpoint in the service.

**Two design choices here are better than the obvious alternative and worth not "simplifying" in review:**

1. **The name is snapshotted at offer time, not looked up live.** An authorisation record must not be silently rewritten by a later rename — the screen must show what was true when the grant was made. A live lookup would also put a second service in the request path of a consent screen.
2. **Null is explicit, with a documented fallback to the party id.** The spec says clients must not render a blank, because on a consent screen a blank reads as a name that failed to load rather than as an unknown counterparty.

Also note this carries a Flyway migration (`V3__delegation_counterparty_names.sql`) and a threat-model update, so it is not a display-only change.

Authoring context: the agent that wrote this was terminated by an API error before it could report, after the service builds and coverage had passed. The work is complete as far as CI can tell — all five required contexts are green — but nobody wrote a final self-assessment, so this comment is a substitute for it rather than a review sign-off.

JiRaska:
## Retroactive review (post-merge)

This PR merged without human review due to an incident (`gh pr merge --auto` merging immediately because `required_approving_review_count: 0` — see #3817). Doing the review now, after the fact, rather than reverting.

**Content assessment — sound.**
- DB change is additive/nullable (`V3__delegation_counterparty_names.sql`, two nullable varchar(200) columns), explicit no-backfill rationale, rollback note present. Verified this migration has **not** yet reached any deployed environment (no gitops auto-deploy PR touching delegation-service since the merge) — so there was no live-DB exposure window either way.
- Disclosure analysis (T13) is specific: names ride existing party-scoped read paths (404/403 already enforced), no new authority granted, correctly identifies the residual (staff already had grant-read; snapshot can go stale on rename, explicitly labeled as "not an identity assertion").
- Scope discipline: correctly rejected the broader "give customer-edge party.read" alternative in favor of the narrower snapshot-at-offer-time approach — smaller PII/authority surface.
- API versioning correct (1.1.0 → 1.2.0, backward-compatible addition, minor).
- Contract test gap (PR-lane not pulling in pid-service for the pact replay) was caught and fixed by giving the existing `@PactFolder` provider test real documentation tying it to the new fields, rather than being left for after merge.

**Verdict:** approving retroactively. No follow-up correction needed. #3839 (the prepared revert) is being closed as a result — see that PR for the closing note.

**Process note:** this does not excuse the merge happening without review — it was still a bypass of the intended money-path gate, mitigated only because the content happened to be good. See #3817 for the process fix (drain tooling now refuses to call `gh pr merge` on money-path scope, verified working in the subsequent drain passes).

Issue #2945, opened 2026-07-31
Six services rely on a name coincidence for their Kafka group.id — the dotted-key bug from #686 is a no-op there only by accident


JiRaska:
## Nothing is currently broken. This is about why, and how thin the margin is.

#686 fixed the SmallRye dotted-key bug: a `group.id:` written as a YAML leaf key registers only as the quoted property `"group.id"`, which `KafkaConnectorIncomingConfiguration`'s plain lookup never finds, so `group.id` silently falls back to `quarkus.application.name`.

Seven services on `main` **still carry those dotted keys**. I checked each against the fallback value rather than assuming, and the result splits cleanly:

| service | `group.id` in YAML | `quarkus.application.name` | effect |
|---|---|---|---|
| account | `openbank-account-service` | `openbank-account-service` | none — **identical to the fallback** |
| aml | `openbank-aml-service` | `openbank-aml-service` | none — identical |
| balance | `openbank-balance-service` (×2 channels) | `openbank-balance-service` | none — identical |
| document | `openbank-document-service` | `openbank-document-service` | none — identical |
| party | `openbank-party-service` (×3 channels) | `openbank-party-service` | none — identical |
| statement | `openbank-statement-service` | `openbank-statement-service` | none — identical |
| transaction | `transaction-scheme-accepted-cg` | `openbank-transaction-service` | **differs** — but covered by `transaction-service-msg-override` ConfigMap (`config_ordinal=500`), and the YAML says so in a comment |

So six services are correct **by coincidence**: the value the broken key was supposed to set happens to equal the value the fallback produces. The seventh is correct **by design**, via the ConfigMap override pattern #686 established, and documents it in place.

## Why that coincidence is worth a guard

The six have no override ConfigMap, so nothing but the name collision is holding them up. Each of these turns them into silent consumer-group changes, with no error at any layer:

- renaming a service (`quarkus.application.name` moves, the YAML `group.id` does not)
- giving a channel a purpose-specific group id, exactly as transaction-service did — the moment the intended value stops matching the app name, the dotted key stops being a no-op
- adding a second incoming channel that needs its own group

A silent consumer-group change is not a crash. It is a consumer joining a *different* group, re-reading from that group's offset — which on a money-path topic means reprocessing or skipping, discovered later and from the data.

## Suggested guard

Cheap and fully derivable from committed artifacts, no cluster access:

[...]

Issue #686, opened 2026-07-09
fix(kafka): silent group.id/auto.offset.reset YAML dotted-key bug across services


JiRaska:
## Sweep theme
Quarkus's YAML config source (`quarkus-config-yaml` / SmallRye `YamlConfigSource#flattenYaml`) unconditionally quotes any YAML leaf map key containing a literal dot when registering it as a MicroProfile Config property. A YAML key written as `group.id: foo` under `mp.messaging.incoming.<channel>:` registers ONLY as `mp.messaging.incoming.<channel>."group.id"` (literal quotes in the property name) — never as the plain `mp.messaging.incoming.<channel>.group.id` that `KafkaConnectorIncomingConfiguration.getGroupId()` (and the `auto.offset.reset` / `dead-letter-queue.topic` equivalents) actually read. This happens regardless of whether the YAML key itself is quoted, and was confirmed empirically (a throwaway diagnostic test dumping `ConfigProvider.getConfig().propertyNames`) while root-causing the openbank-fraud-service boot-crash.

`group.id` then silently falls back to Quarkus's default (`quarkus.application.name`). If that fallback isn't the identity granted Read/Describe in the service's Strimzi `KafkaUser` ACL, every consumer poll gets a silent `org.apache.kafka.common.errors.GroupAuthorizationException` — fail-silent (no error, no log) until it lands inside Quarkus's synchronous startup window and becomes a fatal crash-loop. `auto.offset.reset` has a quieter but more insidious failure mode: it silently falls back to Kafka's client default (`latest`) instead of the intended `earliest`, so a consumer silently skips backlogged messages on first boot / after any group reset — no crash, no error, just missing data.

## Driving ADR
N/A — this is an operational bug fix (config-defect class), not a policy rollout. No ADR governs Kafka `mp.messaging` YAML shape.

## Audit findings (2026-07-09, 12 services)

**BROKEN — real GroupAuthorizationException risk (group.id fallback ≠ ACL grant):**
- `openbank-audit-service` — `audit-events-in`: intended `audit-service`, fallback `openbank-audit-service`, ACL grants `audit-service` only
- `openbank-notification-service` — `notification-events-in` (intended `notification-service`) + `party-events-in` (intended `notification-service-party`), fallback `openbank-notification-service`, ACL grants only the two specific names
- `openbank-kyc-service` — `party-events-in`: intended `kyc-service-party`, fallback `openbank-kyc-service`, ACL grants `kyc-service-party` only
[...]

PR #3798, opened 2026-08-06
fix(gitops): make the six coincidental Kafka group.id values actually resolve

JiRaska:
## What this is

#2945 asked why six services' Kafka `group.id` values are correct **by coincidence** rather than
by design, and how thin that margin is. This makes them correct by design, for the `group.id` half
only, and leaves the `auto.offset.reset` half to the owner because changing it is a data event, not
a config edit.

## Measured, per channel

Effective values measured against the live sandbox cluster (kafka-exporter
`kafka_consumergroup_current_offset` / `kafka_topic_partition_*_offset`), not inferred. `group.id`
in every case equals `quarkus.application.name`, so the effective group today is the fallback and
happens to match the file.

| service | channel | topic | group (declared == effective) | committed offsets | money-path |
|---|---|---|---|---|---|
| account | party-events-in | openbank.party.events | openbank-account-service | yes (169/169) | yes |
| account | delegation-events-in | openbank.delegation.events | openbank-account-service | **none** | yes |
| aml | party-events-in | openbank.party.events | openbank-aml-service | yes (169/169) | no |
| balance | ledger-events-in | openbank.ledger.journal.posted | openbank-balance-service | yes (571/571) | yes |
| balance | balance-init-in | openbank.accounts.account.created | openbank-balance-service | yes (129/129) | yes |
| document | account-created-in | openbank.accounts.account.created | openbank-document-service | yes (129/129) | no |
| party | kyc-events-in | openbank.kyc.events | openbank-party-service | yes (115/115) | no |
| party | aml-events-in | openbank.aml.events | openbank-party-service | yes (109/109) | no |
| party | consent-events-in | openbank.consent.events | openbank-party-service | yes (10/10) | no |
| statement | account-events-in | openbank.accounts.account.created | openbank-statement-service | yes (129/129) | no |

Each service's KafkaUser ACL grants `Read` on exactly that literal group name, which is the second
independent confirmation that the fallback value is the one actually in use.

## What this changes

A `<svc>-msg-override.yaml` ConfigMap per service (`override.properties`, `config_ordinal=500`,
mounted via `QUARKUS_CONFIG_LOCATIONS`) setting `group.id` for all ten channels — the pattern
`transaction-service` and `kyc-service` already carry. **Every effective value is unchanged**: each
override sets the group the consumer is already committed on. The deploy restarts the six pods and
moves no offset.

The dotted YAML keys are deliberately **not** deleted. Local dev and tests read them fine; only the
deployed path breaks, so deleting them would make local behaviour match neither state.

`check-kafka-dotted-keys.py` drops the "a `group.id` equal to `quarkus.application.name` is
acceptable" exemption in the same change, so the coincidence cannot come back — an override
ConfigMap is now the only shape that passes. Falsified both ways: the new checker reports **10
findings against the pre-change tree and 0 against this one**. Its self-test was also rewritten to
drive the real `scan()` over a synthetic tree; the previous version re-implemented the classifier
inline and so could not exercise `load_overrides()` at all — the one rule that now decides every
`group.id` verdict.

## What this deliberately does NOT change: `auto.offset.reset`

All ten channels declare `auto.offset.reset: earliest` and it is subject to the identical bug, so
the effective value is the connector default (`latest`). Making it effective is an **operational**
decision and is left baselined in the checker.

The precise risk, which is narrower than "forcing earliest replays the topic": `auto.offset.reset`
only applies when a group has **no valid committed offset**. Nine of the ten channels have committed
offsets today, so flipping the setting would be inert *at the moment of deploy*. The exposure is
later — broker `offsets.retention` is the 7d default (no override in the Kafka CR), all seven topics
carry `retention.ms: 604800000`, and these services are KEDA-scaled, so a group idle past the
retention window loses its offsets and would then re-read the whole retained log instead of jumping
to the tail. Worst cases: `openbank.ledger.journal.posted` (411→571 retained, money-path balance
projection) and `openbank.party.events` (157→169, four consumers).

**One channel is genuinely safe and is worth calling out separately:** account's
`delegation-events-in` has **no committed offsets at all** and `openbank.delegation.events` is
**empty** (oldest == latest == 0). Setting `earliest` there is a no-op today by measurement rather
than by luck. It is still not in this PR: it would be a change with no benefit until the topic has
traffic, and it belongs in the same owner decision as the other nine.

## Linked issues

Refs #2945
Refs #686

## Deployment note

The six pods restart. account, balance and party are Argo Rollouts with `replicas: 1` and a canary
strategy, so each progresses through its `setWeight`/`pause` steps (account and balance additionally
run the `openbank-money-path-canary` analysis). Nothing here changes an offset or a group, so a
canary abort is a plain rollback with no data consequence.

## Checks run locally

- `check-kafka-dotted-keys.py` — self-test 6/6, clean against this tree, 10 findings against `main`
- `check-msg-channel-image-parity.py` — 25 channel overrides across 17 services, every one servable
  by the image actually deployed (this is what proves the ten new channel names are real)
- `check-gitops-secret-refs.py` — 522 declared / 385 references, every referenced ConfigMap declared
- `check-single-replica-rollout-strategy.py` — 0 new vs baseline
- `yamllint` over `openbank-infra .github` — clean

No Gradle module is touched, so no service build or `version.txt` is in scope.

### DIFF
--- a/.github/scripts/check-kafka-dotted-keys.py
+++ b/.github/scripts/check-kafka-dotted-keys.py
@@ -17,11 +17,16 @@
 #   `*-msg-override.yaml` ConfigMap carrying `override.properties` with `config_ordinal=500`.
 #
 # WHAT THIS CHECKS, AND WHY THE TWO KEYS DIFFER
-#   `group.id`         — the fallback is `quarkus.application.name`. Six services get away with the
-#                        broken key purely because the value they wanted happens to equal their
-#                        application name. That is correct by coincidence: it breaks the day a
-#                        service is renamed, or a channel wants its own group (exactly what
-#                        transaction-service did). Accepted, but only when the values match.
+#   `group.id`         — the fallback is `quarkus.application.name`. Six services USED TO get away
+#                        with the broken key purely because the value they wanted happened to equal
+#                        their application name — correct by coincidence, not by design: it breaks
+#                        the day a service is renamed, or a channel wants its own group (exactly
+#                        what transaction-service did). All six now carry a `*-msg-override`
+#                        ConfigMap setting the same value (#2945), so an override is the ONLY thing
+#                        that makes a dotted `group.id` acceptable and the coincidence exemption is
+#                        gone. Removing it is what stops the next service re-introducing the
+#                        latent version of this bug: matching the app name would otherwise pass CI
+#                        while resolving through a fallback nobody chose.
 #   `auto.offset.reset`— there is no such coincidence available. The fallback is the connector's own
 #                        default, so a channel declaring a NON-default value and having no override
 #                        is running on the default while its config file says otherwise. This is the
@@ -37,8 +42,9 @@
 #
 # EXIT CODES
 #   0  no new occurrences, no stale baseline entries
-#   1  a new dotted key that neither matches its fallback nor has an override — or a baseline entry
-#      that is now covered and should be removed
+#   1  a dotted `group.id` with no override ConfigMap, a dotted `auto.offset.reset` asking for a
+#      non-default value with no override — or a baseline entry that is now covered and should be
+#      removed
 #   2  the check could not run (PyYAML missing, tree not found). Never conflated with 0.
 #
 # Run:  python3 .github/scripts/check-kafka-dotted-keys.py [--root .] [--self-test]
@@ -151,9 +157,10 @@ def scan(root):
                 if (service, channel, key) in overrides:
                     continue
 
-                # group.id survives when the fallback (quarkus.application.name) is the same string.
-                if key == "group.id" and name is not None and value == name:
-                    continue
+                # NOTE: `group.id == quarkus.application.name` used to be accepted here. It is not
+                # any more (#2945) — see the header. A value that merely happens to equal the
+                # fallback still does not RESOLVE, so accepting it means the config file and the
+                # running consumer agree by accident, which is not a property a gate can keep.
                 # auto.offset.reset survives only when it asks for the connector default anyway.
                 if key == "auto.offset.reset" and value == CONNECTOR_DEFAULT_OFFSET_RESET:
                     continue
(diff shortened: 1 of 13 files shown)


JiRaska:
**Review: 1 finding, 1 blocking**

- **BLOCKING** — `openbank-infra/gitops/components/document-service/document-service-service.yaml` (volumes section, ~line 358-366): the new `msg-override` volume entry was inserted between `signing-keystore`'s `secret.optional: true` and its `items:` list, so `items:` is now nested *under* the new entry instead of under `signing-keystore`:
  ```yaml
  - name: signing-keystore
    secret:
      secretName: document-service-signing-keystore
      optional: true
  - name: msg-override
    configMap:
      name: document-service-msg-override
      items:
        - key: keystore.p12
          path: keystore.p12
  ```
  This is syntactically valid YAML but semantically broken: the `msg-override` ConfigMap (`document-service-msg-override.yaml`) only has a `override.properties` data key, not `keystore.p12` — so the volume mount will fail at pod creation (`configmap … references key … which does not exist`), and `signing-keystore`'s `items` projection (the one thing that made that secret mount usable) is silently dropped. This will CrashLoop/fail to schedule document-service on deploy. Every other service touched by this PR (account, aml, balance, party, statement) appended its `msg-override` volume entry cleanly at the end of the `volumes:` list — only document-service has this insertion-order defect.

**On the `auto.offset.reset` replay question:** No, this PR does NOT flip the effective `auto.offset.reset` for any existing consumer group. Every one of the six new `*-msg-override.yaml` ConfigMaps sets **only** `group.id` (which resolves to the same value the coincidental fallback already produced — verified: account→`openbank-account-service`, aml→`openbank-aml-service`, balance→`openbank-balance-service`, party→`openbank-party-service`, statement→`openbank-statement-service`, all matching `quarkus.application.name`). Each file explicitly documents that `auto.offset.reset: earliest` is *deliberately not set* here, citing the exact mass-replay risk this task flagged, and defers it to a separate owner decision (#2945). The `check-kafka-dotted-keys.py` gate change also correctly keeps the `auto.offset.reset` exemption for the connector-default case and does not weaken it. **No replay risk from this PR as written.**

Verified: the gate's self-test now drives the real `scan()`/`load_overrides()` path against a synthetic tree rather than a retyped copy of the classifier (good — addresses the "self-test can't model what you didn't imagine" failure mode); confirmed ADR-0237 exists and the group.id values in all five other new override ConfigMaps match their intended service's ACL-granted group. Did not verify the ArgoCD sync-wave ordering for `QUARKUS_CONFIG_LOCATIONS`/ConfigMap availability at pod start (no sync-wave annotations visible in this diff, consistent with the other files in the batch).

JiRaska:
Reviewed. **No blocking findings** — and the thing I went looking for is the thing this PR deliberately does not do.

The dangerous half of the dotted-key defect is `auto.offset.reset`. Six services declare `auto.offset.reset: earliest` under a dotted YAML key that never reaches the connector, so the effective value is the connector default; *making it effective* forces `earliest` onto consumer groups currently running as default, which is a mass replay on money-path channels. This PR sets **only `group.id`** and says so in each ConfigMap ("Making it effective is an operational decision, not a config tidy-up"). Correct call.

The second risk is subtler and is where a well-meant version of this change does damage: if an explicit `group.id` differs from the fallback it replaces, the consumer joins a **new** group with no committed offsets, and with the connector's default `latest` it silently skips the backlog rather than replaying it. Checked all six against `quarkus.application.name` on `main`:

| component | `quarkus.application.name` | override sets |
|---|---|---|
| accounts | `openbank-account-service` | `openbank-account-service` |
| aml | `openbank-aml-service` | `openbank-aml-service` |
| balances | `openbank-balance-service` | `openbank-balance-service` |
| document-service | `openbank-document-service` | `openbank-document-service` |
| party | `openbank-party-service` | `openbank-party-service` |
| statements | `openbank-statement-service` | `openbank-statement-service` |

Identical in every case, so the change is behaviourally a no-op today: same group, same offsets, nothing to replay or skip. It converts a coincidence into a guarantee, which is exactly the stated goal — the coincidence breaks the day someone renames a service or gives a channel its own group.

Wiring verified per service rather than assumed, since a mount without the env var (or a ConfigMap name that does not match) is inert and fails silently:

```
account-service.yaml           env=1  mount+volume=2  account-service-msg-override
aml-service.yaml               env=1  mount+volume=2  aml-service-msg-override
balance-service.yaml           env=1  mount+volume=2  balance-service-msg-override
document-service-service.yaml  env=1  mount+volume=2  document-service-msg-override
party-service.yaml             env=1  mount+volume=2  party-service-msg-override
statement-service.yaml         env=1  mount+volume=2  statement-service-msg-override
```

Each ConfigMap's `metadata.name` matches the reference in its workload, and each `metadata.namespace` matches its workload's namespace (`documents` for document-service, which is the one that could plausibly have been wrong).

Also right: **no `application.yaml` key is deleted** (0 service `application.yaml` files touched). Local dev and tests read the dotted key fine — only the deployed path is broken — so removing it would trade one silent failure for another.

The gate change is a ratchet in the right direction: `check-kafka-dotted-keys.py` drops the "group.id equals the application name is harmless" exemption, so the six can no longer regress and a new occurrence fails outright.

One follow-up worth keeping visible rather than folding in here: the six `auto.offset.reset: earliest` declarations are still inert, and the decision to make them effective needs someone who can reason about replaying those topics.

PR #3808, opened 2026-08-06
fix(libs): move CDI interceptor bindings out of libs-domain and close the gate hole that hid them

JiRaska:
## What

`@Authorize` and `@FeatureFlag` — the fleet's most-used annotations, on ~130 files across 40 services — were declared in `openbank-libs-domain` as `jakarta.interceptor.InterceptorBinding`. CDI is framework. ADR-0002 says the domain layer has zero framework imports; ADR-0122 split domain from runtime precisely so that could hold. The **enforced** `domain-purity-gate` was green about it every single run.

This PR closes the gate hole first, then fixes what the corrected gate finds.

## Why the gate missed it — measured, not guessed

Two independent blind spots, and this landed in the intersection of both.

**1. Scope was a package-path segment, not a module.** `find ... -path "*/domain/*"` needs a literal `/domain/` in the path. `openbank-libs-domain`'s own sources live under `com/openbank/libs/authz/`, `.../flags/`, `.../persistence/outbox/` — the module directory is `openbank-libs-domain`, so the character before `domain` is `-`, not `/`. **80 of the module's 107 main sources (75%) were never opened.**

**2. Detection was `^import <prefix>`, and a fully-qualified reference has no import line.** `@jakarta.interceptor.InterceptorBinding` written inline is invisible to an import grep, as is `@get:jakarta.enterprise.util.Nonbinding`. Both annotations used exactly that form. This evasion is *not* libs-domain-specific — it works just as well in every per-service `domain/` package the gate already scanned.

### Both halves falsified separately, against the real tree

Not against a fixture, and with the **old bash gate as a control** in each case:

| known-positive | new gate | old gate (control) |
|---|---|---|
| clean tree | `279 scanned, 19 found, 0 new` → exit 0 | `11 found, 11 new` |
| **scope half**: plain `import io.quarkus.runtime.Startup` in `libs/flags/FlagExposure.kt` (a libs-domain path with no `/domain/` segment) | **1 new** → exit 1 | `11 found, 11 new` — *unchanged, blind* |
| **detection half**: inline `@jakarta.interceptor.InterceptorBinding` in `openbank-billing-service/.../domain/BillingCycle.kt` (a dir the old gate **did** scan) | **1 new** → exit 1 | `11 found, 11 new` — *unchanged, blind* |

The old gate's counts are identical in all three rows: it cannot express either defect. The detection row deliberately uses a directory the old scope already covered, so the two halves are demonstrated independently rather than as one combined change.

## The gate fix

Rewritten as `.github/scripts/check-domain-purity.py`.

- **Scope**: every per-service `*/domain/*` package (unchanged) **plus the whole `openbank-libs-domain` module**, which *is* the domain side of ADR-0122 regardless of how an individual package is spelled. `openbank-libs-runtime` stays module-exempt.
- **Detection**: both the `import` form and any fully-qualified reference, matched on **comment- and string-stripped** source so prose about a framework type is not a violation. Kotlin block comments nest, so the stripper counts depth — a KDoc containing `/*` would otherwise close early and leak the rest of the file back in.
- **`javax.` narrowed** from the blanket prefix to the legacy Jakarta-EE subpackages. The blanket form would have manufactured 16 violations for `javax.crypto` and `javax.xml`, which are JDK, not framework — baselining those would have made the ratchet a lie.
- **`--self-test`, which the bash version never had.** 20 cases, `selftest:` declared in `gates.yaml` (`selftest_expect` defaults to `pass`). Must-flag includes the two exact #3670 shapes; must-not-flag includes the code-about-code cases this repo has been bitten by twice (a line comment naming the framework, a nested KDoc, a string literal) and both package-boundary directions.

### The scan now reports how many files it opened, and fails if that collapses

Widening the scope fixes today's blind spot but leaves the failure *mode* intact: with only `0 new` on stdout, a future scope bug is indistinguishable from a clean tree. That is the whole history of this gate — a green that meant "it never looked".

Output is now `scanned 279 domain-layer file(s); 19 violation(s) found, 0 new, ...`, and the run fails below `MIN_FILES_SCANNED`.

Both numbers are measured, and the measuring mattered twice:

- A first draft guessed the tree at ~470 files and set the floor at 300. The healthy tree scans **279** — the floor rejected a good tree on its first run.
- Corrected to 200, which felt like generous headroom. But reverting `in_scope()` to the old path-only rule takes the count **279 → 201** — so 200 would have sat *just under the exact collapse it exists to catch*. The floor is **250**, chosen against that measurement.

Falsified with the baseline file emptied, proving the floor fires **on its own** rather than riding on the stale-baseline ratchet that also happens to detect this case. Its honest scope is documented at the constant: it is a coarse backstop for a wholesale scope collapse, it cannot see a bug that drops a handful of files, and it is not the primary detector even here — the stale-baseline ratchet is, because it names the file.

## What moved

18 violations -> 8 in libs-domain.

- **`@Authorize` / `@FeatureFlag` -> `openbank-libs-runtime`.** Package names unchanged (`com.openbank.libs.authz`, `com.openbank.libs.flags`), so **not one consumer import moved**. All 40 consumer modules were checked to already depend on `openbank-libs-runtime` — none needed a build-file edit. The ports they name (`PolicyDecisionPoint`, `AuthzQuery`) stay in libs-domain, which is the direction the dependency is allowed to run. **No `@Produces` was added** — the documented libs-runtime failure mode that drags a type into the Arc closure of every consumer, including ones that never use it, is not in play here.
- **`org.jboss.logging.Logger` -> the JDK's `System.Logger`** in `OutboxDispatch`, `FlagExposure` and `StaticServiceTokenProvider`. Same category, same destination: under Quarkus the JDK logger bridges into the JBoss LogManager via JUL. Two of the three were inline fully-qualified references, so nothing had ever seen them — they were not in the issue either.
- **libs-domain's build file now declares no framework dependency at all**, not even `compileOnly`.

## What is baselined, and why that is not smuggling

8 Jackson references: `ApiError`, `CursorPage`, `FlagdProvider`, `OpaSidecarPolicyDecisionPoint`, `CompliancePackParser`. These are **not new code** — they were in the tree the whole time, in the 75% of the module the gate never opened. Widening a gate's scope always uncovers pre-existing violations, and grandfathering them is what makes the widening shippable at all.

Each needs a real refactor rather than a move — a serialization boundary for the two wire DTOs, and moving two HTTP+Jackson infrastructure adapters to libs-runtime (`OpaSidecarPolicyDecisionPoint` is called out in the issue for exactly this). The baseline file states the reason per entry and the ratchet reports a stale entry in either direction, so this cannot quietly become permanent. Tracked on #3670, which stays open.

## Rebased onto current main (was 89 behind)

`origin/main` moved 89 commits while this sat. Merged in, with the documented "a clean merge can silently delete a list entry" check run on every structured file:

- `gates.yaml`: **111 gate ids on main, 111 in the result**, set-diff empty — main's 6 new gates all survived.
- `rules.yaml`: parsed with a duplicate-key-detecting loader — **no duplicate keys** (this file is never linted by CI, and SnakeYAML silently keeps the last of a repeated key).
- One real conflict, in `FlagExposure.kt`: #3882 landed `at=${exposure.timestamp}` in the same log statement this PR was rewriting to `System.Logger`. Resolved to keep **both** — the new field is present in the merged line, not dropped by taking "our" side.

## Verification

Every build below was run with `--rerun-tasks` and the task list read, so no claim rests on an `UP-TO-DATE` task.

- `:openbank-libs-domain:build` + `:openbank-libs-runtime:build` — **47 tasks executed, 0 up-to-date**; includes `AuthorizeInterceptorTest` / `FeatureFlagInterceptorTest` and both `koverVerify` bounds.
- `:openbank-ledger-service:build` + `quarkusBuild` (**money-path**, carries the `AuthzProducer`) — 46 executed.
- `:openbank-party-service:build` + `quarkusBuild` (the tree's **only `@FeatureFlag` consumer**) — 44 executed.
- `:openbank-sepa-payment:quarkusBuild` (**second money-path**) + `detekt` on both libs — 31 executed.
- `quarkusBuild` is the load-bearing one: CDI wiring is not validated by ktlint or unit tests, so an interceptor-binding move is only believable after a real Quarkus augmentation.
- `--self-test`: 20/20.
- `rules.yaml`'s `architecture` key is not in the OPA read set: `gen-rules-opa-data.py` produces no diff and no bundle is restamped.

### The one red check on the previous head was a flake, and that was checked rather than assumed

`build (openbank-pid-service)` failed on `PidOutboxClaimIT > a stale DISPATCHING row is reclaimed instead of stranded forever`. That test calls `repository.claimProcessable(...)` directly and never goes through `OutboxDispatch` — the only file this PR touches in that area, and only its logger line. Run locally on the merged branch, the JUnit XML reports `tests=2 failures=0 errors=0` for that class, including that exact case. It is a timing-sensitive concurrent-claim DB test.

## What this change does NOT cover — stated plainly

- **It is a source-text gate.** A framework type reached through a Kotlin `typealias`, or through a transitive `api()` dependency that puts framework types on the domain's compile classpath with no source-level mention, is invisible to it. Closing that needs bytecode or dependency-graph analysis, which this is not. No coverage is claimed there.
- **32 of the 40 consumer modules were not built locally.** CI builds them. The move is import-neutral by construction (identical package, dependency already present, verified per module), but "verified by construction" is weaker than "compiled", and the distinction is real.
- **The `System.Logger` bridging claim is reasoned, not observed.** Under Quarkus the JDK logger reaches the JBoss LogManager via JUL, so category and format should be unchanged — but no deployed pod's log output was inspected to confirm the rendered line is byte-identical. These are three log statements, one of them a dev-only warning.
- **The `MIN_FILES_SCANNED` floor is coarse**, as documented above: it catches a wholesale scope collapse, not a partial one.
- **The 8 baselined Jackson violations are still violations.** They are grandfathered, not fixed, and #3670 stays open for them.

## Linked issues

Refs #3670


JiRaska:
Heads-up on a small collision, not a blocker.

#3882 (merged) added the exposure's own `timestamp` to the `LoggingExposurePublisher` log line, in the same statement this PR rewrites from `org.jboss.logging.Logger.infof` to `System.Logger`:

```kotlin
"flag exposure exposureId=%s at=%s flag=%s variant=%s key=%s reason=%s traceId=%s",
exposure.exposureId,
exposure.timestamp,
```

Rationale: the log line's own time is when the fallback publisher ran, not when the subject was exposed to the variant — and `FlagExposure.timestamp` previously defaulted to `Instant.EPOCH`, so nothing anywhere recorded the real time.

When you rebase, please carry `at=${exposure.timestamp}` into the interpolated form. Nothing will go red if it is dropped: the field is not asserted by any test on the publisher, and the log line is not parsed by anything.

The `Instant.now()` default on the data class itself is in a different hunk and does not conflict with this PR.

JiRaska:
**Review: 2 findings, none blocking.** Automated agent review — not an approval.

The catastrophic case for this PR is that moving `@Authorize` out of `openbank-libs-domain` silently stops the `@InterceptorBinding` resolving, turning every `@Authorize` annotation on 44 services into a no-op with nothing failing. **I checked that specifically and it does not happen here**, for four independent reasons:

- The FQCN is unchanged (`com.openbank.libs.authz.Authorize` in both places), so no consumer import moves and the binding the interceptor declares is the same type the endpoints carry.
- `AuthorizeInterceptor` already lives in `openbank-libs-runtime` on `main`; the annotation joins it rather than separating from it.
- Both modules apply `alias(libs.plugins.jandex)`, so ArC still gets an index containing the `@InterceptorBinding` declaration.
- The old `openbank-libs-domain/.../authz/Authorize.kt` is deleted, not left behind — no duplicate FQCN on the classpath.
- Every one of the 44 services with `@Authorize` in `src/main` declares `project(":openbank-libs-runtime")`. The only module using `com.openbank.libs.authz` without that dependency is `openbank-authz-policy-auditor`, and its single reference is a path string to `AuthorizeInterceptor.kt` — which this PR does not move.

Findings:

1. **Nothing in the PR *proves* the interceptor still fires.** The reasoning above is mine, from the tree; the diff adds no test. A unit test calling a resource class directly cannot distinguish a registered interceptor from an unregistered one (same class of blindness as `McpEndpointIdentityTest` vs the `@Path` binding bug, #3371). A single `@QuarkusTest` in any consumer service driving a real HTTP request at an `@Authorize`-decorated endpoint with a PDP stubbed to deny, asserting 403, would make this move falsifiable forever and costs one file. The `AllowAllPolicyDecisionPoint` KDoc even says an accidentally-shipped instance is grep-able by `reason = "test-stub"` — assert on that reason and the test also proves the PDP is wired.

2. **`.github/scripts/domain-purity-baseline.txt` leaves `OpaSidecarPolicyDecisionPoint` and `FlagdProvider` in `openbank-libs-domain`** — an HTTP+Jackson adapter and a flag provider — grandfathered rather than moved. That is a defensible scoping call and the file says so, but it means the production PDP still sits on the domain side of the ADR-0122 split. #3677 in this batch moves exactly these two. See the overlap note below.

**Overlap warning (not a finding against this PR, but read it before merging either).** #3677 performs the same `Authorize.kt` + `FeatureFlag.kt` move, adds a *different* gate (`check-libs-domain-purity.sh`, gate id `libs-domain-framework-free`) for the same invariant, and additionally moves `OpaSidecarPolicyDecisionPoint`/`FlagdProvider`. Merging both without re-reading the result risks two gates for one rule, and — the case that actually matters — a conflict resolution that reinstates the `libs-domain` copy of `Authorize.kt` alongside the `libs-runtime` one. Two classes with that FQCN on one classpath is precisely the state in which the interceptor binding can stop matching *silently*. This PR's gate is the stronger of the two (Python, self-tested, whole-module scope, matches fully-qualified references as well as imports); #3677's is bash, `jakarta.`-only on the FQ axis, and has no `selftest:`.

Verified clean: gate registered in `.github/gates/gates.yaml` with `mode: enforced` and a `selftest:`; scope is derived (whole module + per-service `*/domain/*`), not a hand-kept list; detection strips comments and matches both `import` lines and fully-qualified references, closing the two blind spots that let `jakarta.interceptor.InterceptorBinding` sit in the domain module while every run was green; the baseline is documented, shrink-only, and explicitly says "never add entries"; `rules.yaml: architecture.ci_producer` was updated to the new script path.

This needs human review regardless — it touches the annotation that gates 44 services' endpoints.

JiRaska:
**The 7 failing service builds on this PR are not this PR.** I checked four of them; each log contains
only

```
##[error]Failed to resolve action download info. Error: Service Unavailable
```

three times, with no compiler output at all — a GitHub action-resolution outage on 2026-08-06 ~15:48Z.
No later run in the repo shows it, so it is over. `gh run rerun --failed` is the whole remediation.

Flagging it because the failure set (`openbank-libs-domain` plus 6 consumers) reads exactly like "the
`Authorize`/`FeatureFlag` move broke compilation for everyone that imports them", which is the most
plausible way this PR *could* fail. It did not fail that way, and a check count alone cannot tell the
two apart.

Context on scope: #3677 implements the same issue (#3670) and also creates
`openbank-libs-runtime/.../authz/Authorize.kt` and `.../flags/FeatureFlag.kt`, while adding a *second*
gate for the same invariant. Comparison and a suggested split are on #3670. Nothing to change here —
this PR's gate rewrite is the half I would keep.

Issue #3765, opened 2026-08-05
Shared M2M client reaches matrix-granted writes fleet-wide via matrix-allows (post-#3734 residual)


JiRaska:
## Problem

`matrix-allows` in the shared `rest.rego` grants every action in `rules.yaml`'s `role_action_matrix` ROLE_OPERATOR grant list to **any** HUMAN principal carrying ROLE_OPERATOR — including `service-account-openbank-services`, the shared backend M2M client (ledger, billing, account, agent, document, delegation, … all authenticate with one Keycloak client whose service-account carries ROLE_OPERATOR).

The #3734 batch closed this path for the **edge** client per-service (exclusion + `prohibited` veto). But the **shared** client's matrix path remains open fleet-wide:

- The base `shared_m2m_write_prohibition` register (`rules.yaml`) vets the shared client only when its **only** allow reasons are the listed role-only `operator-*-write` reasons. `matrix-allows` is not (and structurally cannot be) in that register — it is a base-layer reason, and adding it would veto every legitimate matrix-granted shared call.
- Rule-level exclusions (`not startswith(input.principal.id, "service-account-")`) do not bind `matrix-allows` — it is a separate rule in base `rest.rego` that never consults per-service ext exclusions.

Concrete consequence after the #3734 batch merges: on ledger, the shared client keeps its legitimate `ledger.post`/`ledger.reverse` identity grants — but `matrix-allows` still admits it to `ledger.yearClose.approve` (and every other OPERATOR-granted write on services without an ext-level shared veto). Same shape on account (7 writes), fx (`fx.convert`), lending (8 writes), billing (3 writes), sepa-payment (`transitionStatus`, `approval.decide`).

## Why this is a fleet decision, not a per-service PR

1. The shared client is **backend-only** (no customer-facing surface), so the blast radius differs from the edge — but it is one credential held by ~20 services; lateral movement after any single service compromise reaches every matrix-granted write fleet-wide.
2. Fix options need a decision:
   - **(a) Per-service ext `prohibited` vetoes for the shared client** — same pattern as the edge batch, ~15 more PRs, each needing caller evidence (some services DO have legitimate shared-client writes via identity rules — those actions stay).
[...]

PR #3697, opened 2026-08-03
feat(sdd): add OPA sidecar, matrix grants, and AUTHZ_ENFORCE=true

JiRaska:
## Summary

Brings openbank-sdd-service to enforced OPA authorization (issue #3679). Unlike interest (#3695) and delegation (#3696) — which had sidecars and grants already — sdd had **no OPA sidecar, no policy bundle, and no write grants** in the role-action matrix, and its gitops comment still claimed "non-money-path advisory" even though rules.yaml has listed sdd as money-path since #1478. This PR is the full fan-out: sidecar + bundle generator + bundle + matrix grants + fleet bundle regeneration + the flip.

## Type of change

- [x] feat (behavior change: authorization now enforced; new sidecar)
- [ ] breaking change

## Linked issues

Refs #3679

## Changes

- `components/sdd/sdd-service.yaml`: OPA sidecar container (docker.io/openpolicyagent/opa:1.17.0, explicit registry prefix), opa-bundle volume + mounts, `policy-checksum` annotation, `AUTHZ_ENFORCE=true`, stale "non-money-path advisory" comment rewritten.
- `components/sdd/gen-sdd-opa-bundle.sh` (new, **mode 100755** — a non-executable generator no-ops the CI discovery loop) + generated `sdd-opa-bundle.yaml`.
- `rules.yaml`: `sdd.approve/authorise/create/delete/update` added to `role_action_matrix.ROLE_OPERATOR.grant` (`sdd.list/read` already present; ADMIN inherits OPERATOR).
- `rules-opa-data.yaml` regenerated FIRST (the matrix is OPA-read data), then all 42 bundle/annotation pairs restamped — the expected, CI-enforced ripple; generators re-run twice, idempotent.

## Pre-flip coverage verification

- All seven @Authorize actions (`SddResource` + `SddMandateQueueResource`) are matrix-granted to ROLE_OPERATOR; no action is left to deny-by-default.
- Callers (network-policies: customer-edge + admin-ui on 8129): the edge's service account carries ROLE_OPERATOR, so matrix grants reach it — matching today's `@RolesAllowed(OPERATOR/ADMIN/PAYMENTS/API)` posture, no regression.
- ROLE_PAYMENTS deliberately NOT granted: SDD mandate management is desk/operator scope, not a payment-rail action (the ROLE_PAYMENTS matrix covers SCT/SEPA/clearing only).
[...]

PR #3812, opened 2026-08-06
security(governance): gate matrix write grants reachable by an M2M service-account

JiRaska:
## What

Makes a write action joining `rules.yaml: authz.role_action_matrix` under a role an M2M
service-account holds a **build failure unless explicitly declared** — and closes the one write
this measurement found that could be closed with no caller risk.

Refs #3765, #3734. **Not** closing either: the reachable set is not empty, and the remaining
entries need the caller inventory those issues call for.

## The mechanism, restated from the measurement

`rest.rego`'s `matrix-allows` grants every matrix entry to any HUMAN carrying that role. M2M
callers authenticate with a Keycloak client_credentials JWT, which `AuthorizeInterceptor`
classifies as HUMAN, and the service-accounts hold `ROLE_OPERATOR` in at least one realm. So a
line in the ROLE_OPERATOR grant list is a grant to a machine — and nothing in the diff says so.

**No policy layer can catch it.** `shared_m2m_write_prohibition` is keyed by reason NAME, and
listing `matrix-allows` there would veto every legitimate matrix-granted call; a per-service
`*_rest_ext.rego` cannot veto a rule that lives in base `rest.rego`. That is why this is a
build gate and not a rego change.

## How it was measured

`opa eval` against the **live bundle ConfigMaps**, materialised into the sidecar's own directory
layout (`rules-data.yaml` → `rules/data.yaml`, `agents-data.yaml` → `agents/data.yaml`,
`manifest.json` → `.manifest`) and queried at `data.openbank.rest.allow` — reading the rego
cannot tell you which of several reasons actually fires. Every run carries a **must-DENY**
(`zzzbogus.write` for the shared account) and a **must-ALLOW** (`account.read` for a staff
operator); a bundle failing either is reported, never scored. 40 bundles; the `agent` bundle
carries no `rest.rego` and is reported as out of scope rather than as a pass.

Three personas, because the realms disagree about who holds what (below):

| principal | reachable WRITE actions on `AUTHZ_ENFORCE=true` services |
|---|---|
| `service-account-openbank-services` + `ROLE_OPERATOR` | **128** |
| `service-account-openbank-services` + `ROLE_API` only | **51** |
| `service-account-openbank-edge` + `ROLE_OPERATOR` | **133** |

`matrix-allows` is a contributing reason on roughly 40 of the 128. Full per-service, per-action
tables with the reason set for each are posted on #3765 and #3734.

## Two findings that contradict what the repo asserts

**1. The shared-M2M write prohibition has never fired, anywhere.**
`data.rules.shared_m2m_write_prohibition` is not emitted into any bundle — the key is nested
under `change_requirements:` in `rules.yaml`, so `gen-rules-opa-data.py` cannot select it, and
its own header says so. Independently confirmed here by `opa eval`: the six "graduated" reasons
in #3734's table (`operator-balance-write`, `-fraud-`, `-ledger-`, `-sca-`, `-sepa-payment-`,
`-consent-`) all still resolve `allow=true` for the shared identity. The audit's
"denied (graduated)" column does not describe the deployed policy. **Not fixed here** —
promoting the key is a semantic policy change that would deny live money-path callers, and it
belongs with the caller inventory, not smuggled into a gate PR.

**2. `ledger.approve` was reachable, against its own documentation.**
The ext rego says year-close attestation "must NEVER be reachable by any SERVICE principal" and
that an un-mapped call "403s, by design". Measured: `allow=true, reason="operator-year-close-attest"`
for both service-accounts. The rule was role-only; the absence of a `service-*` grant was never
sufficient, because deny-by-default only holds for an action no rule reaches.

## What this PR changes

- **`operator-year-close-attest` now excludes `service-account-*`.** Excluded rather than
  identity-pinned: there is no legitimate M2M attestor to name — every `Ledger*Client` in the
  fleet was read (lending, balance, agent, interest, transaction, finrep, billing, settlement)
  and none exposes the attest / accounting-day / closed-period routes. Falsified both ways on
  the regenerated bundle:

  | principal | `ledger.approve` before | after |
  |---|---|---|
  | `alice.operator` (ROLE_OPERATOR) | allow | **allow** |
  | `bob.admin` (ROLE_ADMIN) | allow | **allow** |
  | `service-account-openbank-services` | allow | **deny** |
  | `service-account-openbank-edge` | allow | **deny** |

  A full before/after sweep across all 40 bundles × 3 personas shows exactly one action changing
  state and nothing newly allowed.

- **New gate `matrix-write-grants` (`check-matrix-write-grants.py`), ENFORCED.** A write action
  granted to a role any M2M service-account holds must appear in
  `rules.yaml: shared_m2m_matrix_write_grants.declared`. The register was seeded with the
  measured 68 grants in this same change, so the tree is clean and the gate blocks only what is
  **new** — which is why it is enforced rather than advisory (unlike its sibling
  `operator-write-naming`, which has a live finding and therefore cannot be).
  - Roles are **derived** from every realm JSON in the tree, not listed, so a realm change
    re-scopes the gate instead of silently un-covering it. Inheritance mirrors `matrix_grants`'
    single `inherits` hop.
  - Ratchets both ways: a declared action no longer granted must be deleted.
  - `--self-test` covers 6 cases including the ones easy to get wrong: a read grant must NOT be
    flagged, a write hiding in an inherited parent list MUST be, a role no service-account holds
    is out of scope, and that same role comes INTO scope the moment a realm grants it.
  - **Falsified against the real diff of PR #3697** (five sdd grants, one of them
    `sdd.authorise` — the call that books a debtor debit): 5 `::error::` lines, rc=1. Injected
    into `rules.yaml` in place and restored from a `cp` backup.

- **`operator-year-close-attest` deleted from `check-operator-write-naming`'s BASELINE** — its
  own stale-entry ratchet surfaced this once the rule stopped being role-only. That script is
  otherwise unchanged and still reports its single pre-existing finding (`operator-mcp-session`).

- Threat model E2 corrected (the "NEVER reachable" claim and its Low rating were both wrong
  before this change), plus the fleet-wide lesson in root `CLAUDE.md`.

## Deliberately NOT done

**`matrix-allows` is not gated on `not startswith(input.principal.id, "service-account-")`.**
The measurement is the argument against doing it here: on every `AUTHZ_ENFORCE=true` service,
each matrix-granted write is ALSO granted by that service's own role-only ext rule, so closing
either path alone closes nothing — and `transaction.create`, `settlement.create`,
`sepaPayment.create` and `sctInstPayment.create` have live callers with no identity-scoped rule
to fall back on. That is #3765's (a)-vs-(b) decision and it needs the caller inventory first.

## Blast radius

Regenerating `gen-rules-opa-data.py` and **every** bundle generator (as the last step before
push) produces zero diff outside `components/ledger/` — the new register is a top-level key no
`.rego` reads, so it does not restamp the fleet. `gen-rules-opa-data.py --check` is green.

Deploy effect: the ledger bundle checksum changes, so ledger pods roll and `ledger.approve`
starts denying both service-accounts. No in-repo caller invokes it.

## Review notes

Money-path authorization: **2 approvals by rule**. Auto-merge is deliberately NOT armed.

`python3 .github/scripts/run-gates.py --all` locally: 98 PASS. The failures are environmental
(gates needing `PR_DIFF_BASE`, `sudo`, or a cluster) and reproduce on `origin/main`;
`threat-model-updated-on-trust-boundary-change` passes with `--base origin/main`.


JiRaska:
**Review: 3 findings, none blocking.** Automated agent review — not an approval.

Checked against this repo's actual authz failure mode (`input.principal.type == "SERVICE"` can never fire; M2M callers are classified `HUMAN` and must be identified by `input.principal.id`). **This PR gets that exactly right.** The ledger ext rule adds `not startswith(input.principal.id, "service-account-")` rather than a `principal.type` test, and the gate's docstring states the reason correctly.

Findings:

1. **`check-matrix-write-grants.py:realm_files/service_account_roles` — the role discovery is blind to a service account that has no `users[]` entry, and partial blindness reads as a pass.** Discovery walks `users[]` for `username` startswith `service-account-`. Measured across the four realm JSONs in the tree: `openbank-infra/gitops/components/keycloak/customers-realm-template.json` declares `serviceAccountsEnabled: true` on `customer-edge-admin` and `openbank-edge-webauthn` and has **zero** `users[]` service-account entries; `realm-template.json` enables `openbank-mcp-service` the same way with no `service-account-openbank-mcp-service` user. So three M2M identities are outside the gate's scope entirely. The `if not roles_by_account` guard only catches the total-zero case — it cannot see partial coverage. Concrete failure: grant `ROLE_OPERATOR` to `openbank-mcp-service`'s service account via a scope mapping or the admin console and every one of the 68 declared writes becomes reachable by it with the gate still green. Suggest also reading `clients[].serviceAccountsEnabled` + client/scope role mappings, and failing loudly when a client has service accounts enabled but no discoverable role source.

2. **`effective_grants()` resolves one level of `inherits`, faithfully mirroring `rest.rego` — which means the *policy* has the same limit.** No 2-level chain exists in `authz.role_action_matrix` today (`ROLE_ADMIN -> ROLE_OPERATOR` is the only edge), so this is latent, not live. Worth a comment in `rules.yaml` next to `inherits` so the next person adding a chain knows both the rego and this gate stop at depth 1.

3. **The register documents a live exposure it does not close, and that exposure is larger than the ledger fix.** `shared_m2m_matrix_write_grants.declared` names 68 write actions — including `balance.debit`, `balance.credit`, `ledger.create`, `ledger.replay`, `transaction.create`, `sepaPayment.create`, `lending.disburse`, `party.merge` — reachable by a caller holding one shared service-account token. In the *deployed* realm template that is `service-account-openbank-edge` (`ROLE_OPERATOR`). Read alongside #3610 in this batch, which establishes that there is exactly **one** confidential M2M realm client and that its secret is a single KV entry: compromise of that one entry is fleet-wide money-path write. Declaring it is the right first move; the register should carry a pointer to the issue that decides (a) vs (b), because as it stands a reader could mistake "declared" for "assessed and accepted".

Verified clean: the gate is `mode: enforced` **and** passes `--enforce` (the load-bearing half — `mode:` alone is vacuous, cf. #2392); `selftest`/`selftest_expect: pass` present with six real falsification cases including the read/write split, inheritance, and the both-directions stale ratchet; the role set is derived from realm files rather than hand-listed; the `is_write` heuristic errs toward "write" (safe direction); the ledger OPA bundle and the `policy-checksum` pod-roll annotation were regenerated consistently with the `.rego` edit (`6af7…` -> `de5f…`), and no other service's bundle changed, which is correct since only a per-service ext rule moved; `operator-year-close-attest` was correctly removed from `check-operator-write-naming.py`'s baseline in the same PR, and `ledger.approve` is correctly absent from the matrix register so the ext exclusion is not defeated by `matrix-allows`.

This needs human security review regardless — it changes a money-path service's policy and seeds a 68-entry security register.

PR #3693, opened 2026-08-03
feat(agent): capture the charter-declared model id on every MCP tool-call audit event

JiRaska:
## Summary

Closes the ADR-0031 D5 evidence-chain gap: every `agent.mcp.tool_call` audit event now carries the acting agent's **charter-declared model id**. The gate runs before any LLM call, so the model cannot come from the request — it comes from the charter (agents-as-code, D1), threaded charter → `AgentIdentity` → `AgentPolicyGate.audit()` payload.

## Type of change

- [x] feat (new functionality)
- [ ] fix (bug fix)
- [ ] refactor (no behavior change)
- [ ] perf (performance)
- [ ] docs
- [ ] chore (build / tooling)
- [ ] security (private until disclosed)
- [ ] breaking change

## Linked issues

Closes #3667
Refs #3669 (ADR-0031 D5 capture matrix documents this design)

## Changes

- `agents.yaml`: every charter declares `model:` (the LiteLLM model id the gateway serves — llama-3.3-70b-versatile for control/dev agents, deepseek-ai/DeepSeek-V3.2 for the rest; the two anonymous placeholder identities honestly declare `unknown`).
- `CharterRegistry`: `modelId(agentId)` lookup over the config mapping, `UNKNOWN_MODEL` fallback; `CharterEntry.model()` with `@WithDefault("unknown")` (backward compatible).
- `McpEndpoint.agentIdentity()`: populates `AgentIdentity.modelId` from the registry.
- `AgentPolicyGate.audit()`: payload gains `model_id`.
- `agent-service/application.yaml`: `model:` mirrored into `agent.charter.charters` (same dual structure allowed-capabilities already uses).
- Tests: new `CharterRegistryModelTest` (known agent / unregistered agent), `AgentPolicyGateTest` + `McpEndpointIdentityTest` assertions for the threading.
- 65 regenerated OPA bundle/annotation files — the expected CI-enforced ripple of editing agents.yaml (generators re-run twice, idempotent).

## What this deliberately does NOT do

- No `prompt_hash`/`tool_calls` on the tool_call event — architecturally N/A at gate time; they live on `agent.model.complete` / `agent.run`, already populated (per-operation capture matrix documented in ADR-0031, PR #3678).

## Definition of Done

- [x] Hexagonal layering preserved.
- [x] Unit tests added/updated — `:openbank-agent-service:build` green (incl. detekt/ktlint/kover).
- [ ] Integration tests — MCP path covered by updated identity/gate tests.
[...]

PR #3787, opened 2026-08-05
feat(docs-truth-agent): wire prompt registry and LLM gateway

JiRaska:
feat(docs-truth-agent): wire prompt registry and LLM gateway

Migrates `openbank-docs-truth-agent` onto the ADR-0148 prompt registry and the ADR-0174 shared `LlmGatewayPort`, following the same pattern used by the other control-plane agents.

What changed
- Added `openbank-libs/governance/prompts/docs-truth-agent/system.v1.md` — the registered system prompt for ADR-status-vs-code drift triage.
- Added `LlmGatewayProducer` and `modelId()` plumbing in `DocsTruthAgentConfig` so the adapter calls the shared LLM gateway instead of holding its own HTTP client.
- Rewrote `LlmDiagnosisAdapter` to load the prompt verbatim from the classpath and delegate to `LlmGatewayPort`; keeps deterministic degradation when the gateway is unavailable.
- Added `LlmDiagnosisAdapterTest` asserting the runtime system prompt equals the registry file byte-for-byte (so the `prompt_hash` resolves) and that a null gateway degrades to the placeholder.
- Packaged the prompt via `build.gradle.kts` `processResources`.
- Marked `docs-truth-agent` as `registered` in `prompts/registry.yaml`.
- Regenerated `openbank-admin-ui/ai-governance-snapshot.json`.

Out of scope
- Mechanical `proposeFixDiff` implementation is left for a follow-up; this PR only lands the LLM seam and registry entry.

Verification
- `./gradlew :openbank-docs-truth-agent:build` passes.
- `python3 .github/scripts/check-prompt-registry.py` passes.
- `python3 .github/scripts/gen-ai-governance-snapshot.py --check` passes.

Closes #3786
Refs #1918

PR #3813, opened 2026-08-06
feat(agent): populate model_id on every AI-attributed audit event

### LINKED ISSUE #3667: agent-service: AI-attributed audit is not consistently populated (model_id, model_version, prompt_hash, policy_decision) — ADR-0031 D5
ADR-0031 D5 and agents.yaml `defaults.audit.capture` declare model_id, model_version, prompt_hash, tool_calls, policy_decision on every AI-attributed AuditEvent. The envelope exists but these fields are not consistently populated on the MCP tools/call path (AgentPolicyGate). This breaks the EU AI Act Art. 12 / DORA Art. 17 evidence chain.

Scope:
1. Populate model_id/model_version/prompt_hash/policy_decision wherever the MCP path can legitimately know them, or explicitly model their absence.
2. Tests asserting the fields on every emitted agent AuditEvent.
3. If a field is architecturally unknowable on the MCP path (model_id lives at the LLM gateway, not the policy gate), say so in ADR-0031 instead of silently under-delivering.

JiRaska:
## Summary

ADR-0031 D5 / `agents.yaml: defaults.audit.capture` declare that every AI-attributed
`AuditEvent` names the acting model. It did not. This populates `model_id` on **every**
`actorType = "AI_AGENT"` emitter in agent-service, models absence explicitly rather than
omitting the key, and adds a structural test so a new AI-attributed audit site cannot ship
unattributed.

Supersedes #3693 (idle 3 days, 60 commits behind `main`, red on four gates because the
`agents.yaml`-derived artifacts were never regenerated). Its design — the model id comes
from the **charter**, because the gate runs before any LLM call — is sound and is kept;
this branch re-applies it on current `main`, regenerates the derived artifacts, and extends
it to the four emitters #3693 did not cover.

## The measurement

Per emitter, before this PR — the three questions the issue asks (nothing sets it / the
caller does not know it / the envelope has no field) resolve differently per path:

| Emitter | Operation | `model_id` before | Why |
|---|---|---|---|
| `AgentRunAuditor` | `agent.run` | present | already resolved from the gateway |
| `ModelGateway` | `agent.model.complete` | present (+`model_version`) | the response carries it |
| `AgentPolicyGate` | `agent.mcp.tool_call` | **absent** | nothing set it — the gate runs *before* the LLM call, so it cannot ask the gateway |
| `McpToolRegistry` | `agent.mcp.tool_exec` | **absent** | nothing set it; only `actorId` was in scope |
| `McpToolRegistry` | `featureflag.flip` | **absent** | same |
| `PromptInjectionGuard` | `agent.guardrail.injection` | **absent** | the field existed on `AgentIdentity`'s successor but nothing populated it |
| `ProposalService` | `agent.proposal.created` | **conditionally absent** | `modelId?.let { put(...) }`, and the MCP draft path passed `modelId = null` |

The envelope was never the blocker: `AuditEvent.payload` is a free-form map, so no
`openbank-libs` change was needed. Five of seven paths were "nothing sets it"; one
(`ProposalService`) was "set, but silently dropped when null".

## What changed

- **`agents.yaml`**: every charter declares `model:` — the LiteLLM model id the gateway
  serves. The two anonymous placeholder identities (`mcp-anonymous`, `ap2-anonymous`)
  honestly declare `unknown` rather than inventing one.
- **`CharterRegistry.modelId(agentId)`** + `CharterEntry.model()` (`@WithDefault("unknown")`,
  backward compatible), mirrored into `agent.charter.charters` in `application.yaml`.
- **MCP path**: `McpEndpoint` threads the charter model onto `AgentIdentity`; the policy
  gate records it. `McpToolRegistry` derives it from the charter for the execution outcome,
  the flag flip, and both drafted proposals.
- **Chat/oversight path**: `AgentChatService.run()` resolves the acting model **once** up
  front and carries it on the identity, so the per-step events name the same model as the
  run-level event. Here the value is the *actual* model, not the declaration — the
  identity constants (`ASSISTANT_IDENTITY`, `OVERSIGHT_IDENTITY`) carry none, so without
  this every per-step event would say `unknown` while the run event named a real model.
- **`ProposalService`**: `model_id` is now unconditional, defaulting to `"unknown"`.

### Absence is modelled, not omitted

An unregistered actor records `model_id = "unknown"`. A missing key and an unknown model
are different facts and an auditor cannot tell them apart; the key is always present.

### The charter is a declaration, not an observation

On the MCP tool-call path the recorded model is what the charter says the agent runs, not
what the gateway served — the gate fires before any completion, so nothing better is
available at that point. The observed value is on `agent.model.complete` / `agent.run`,
which share the run's `traceId`. This is stated in the code, not implied.

## Red-then-green

Reverting only the *population* (keeping every signature, so the tests still compile) and
re-running the suite:

```
RED: audit payload includes model_id from identity (ADR-0031 D5)   expected "llama-3.3-70b-versatile" but was null
RED: audit payload carries unknown model_id when identity has no charter model   expected "unknown" but was null
RED: charter model id is threaded into the identity passed to the policy gate   expected "llama-3.3-70b-versatile" but was "unknown"
RED: tool execution outcome carries the acting model id   expected "llama-3.3-70b-versatile" but was null
RED: an unregistered actor is attributed as explicitly unknown, never absent   {"outcome"="SUCCESS","tool"="get_account"} to contain key "model_id"
RED: a guardrail detection carries the acting model id   expected "llama-3.3-70b-versatile" but was null
RED: a drafted proposal records the acting model id rather than null   expected "llama-3.3-70b-versatile" but was null
RED: every AI_AGENT audit event in this service populates model_id   offenders: AgentPolicyGate.kt, PromptInjectionGuard.kt, McpToolRegistry.kt x2
```

8 red before, 148 tests / 0 failures after.

### The structural test, and its own falsification

`AgentAuditAttributionTest` scans this service's sources and fails on any `AuditEvent`
construction carrying `actorType = "AI_AGENT"` whose payload has no `model_id`. That is
what makes "on every path" checkable rather than asserted.

It strips comments first. This was falsified, not assumed: in the reverted run
`PromptInjectionGuard.kt` **is** reported as an offender even though the file still
contains a comment naming `model_id` — the prose about the thing cannot satisfy the check
for the thing.

Stated limitation: the structural check sees the *literal*, so a conditional
`modelId?.let { put("model_id", it) }` would satisfy it while emitting nothing at runtime.
That is exactly the `ProposalService` shape this PR removes, and it is the behavioural
test (`a drafted proposal records the acting model id rather than null`) that catches it.
The two halves are complementary on purpose.

## Hash chain: not affected, no re-baseline

`AuditRepository.chainHash`'s canonical list is
`prevHash | id | eventType | aggregateType | aggregateId | actorId | actorType |
sha256(payload) | sourceService | correlationId | occurredAt | recordedAt`.

The payload contributes **only via its own SHA-256**, so adding a payload key changes the
hash of new rows computed from their own content and leaves every existing row's
recomputation identical. The canonical field list does not move, and this is **not** a
re-baseline decision (contrast #3586, which changed how timestamps are canonicalised and
did strand 1066 rows). Nothing here is backfilled — `audit_entries` is append-only at the
DB and updates are silently discarded.

## No widening of what is recorded

Only the model identifier is added. No prompt, no completion, no tool arguments. Flagged
for the owner, not changed here: `agent.guardrail.injection` already records a `sample`
field — an 80-char fragment of the offending user input or tool result — which predates
this PR and is the one place agent-service puts model *input* into the audit trail.

## Regenerated artifacts (73 files)

Declaring `model:` in `agents.yaml` restamps everything derived from it: the EU AI Act
inventory provenance hash, the admin-ui AI governance snapshot, and every service's OPA
bundle + pod-roll annotation. Generators were re-run **after** the final `agents.yaml`
edit and are idempotent (second run, no further diff). These are exactly the four gates
#3693 was red on.

## Type of change

## Linked issues

Closes #3667
Supersedes #3693

## Definition of Done

- [x] Hexagonal layering preserved (domain unchanged apart from a data-class field).
- [x] Unit tests added; red demonstrated before green.
- [x] `:openbank-agent-service:build detekt ktlintCheck koverVerify` green locally.
- [x] `run-gates.py --group registry`: 14 pass, 1 pre-existing advisory warn
      (`operator-write-naming`, about `rest.rego`, untouched by this branch).
- [ ] Flyway / OpenAPI / Avro — N/A, no schema or API change.
- [ ] Contract tests — N/A, the audit payload is additive JSON.

## Security checklist

- [x] No secrets or PII added; model ids are non-sensitive config.
- [x] No new suppressions, no new dependency.
- [x] No prompt or completion content added to the audit trail.


JiRaska:
Note on the one red check: **`Customer-app dossier`** is failing on `version: committed="0.14.0" → regenerated="0.15.0"`.

That is not caused by this branch and is not one of the five required contexts. The dossier block is derived from the separate `JiRaska/openbank-app` repo, which released 0.15.0; this PR touches no dossier file (`git diff origin/main...HEAD -- "*dossier*"` is empty). The comparison that looks reassuring is stale rather than contradictory — #3787 shows this check green, but its run is from 2026-08-05T15:20Z, before the app release, while this one ran at 2026-08-06T05:40Z. Any PR whose dossier job runs from now on will be red until the derived block is regenerated.

Flagging it here rather than fixing it inline: regenerating a block sourced from another repo is a separate change with a separate reviewer.

JiRaska:
Ready for human review — **not** auto-merged.

Arming `--squash --auto` was refused by the money-path guard: the PR touches `openbank-infra/gitops/components/{accounts,aml,...}` OPA bundles, which are money-path paths regardless of the `feat(agent)` scope in the title. That is the correct final state and I have not overridden it.

For the reviewer, so the 88-file diff reads quickly:

- **15 files are the change**: 8 Kotlin sources, 5 tests, `application.yaml`, `agents.yaml`.
- **73 files are the generated ripple** of adding `model:` to `agents.yaml` — every OPA bundle plus its pod-roll annotation, the EU AI Act provenance hash, and the admin-ui AI governance snapshot. All produced by the committed generators, re-run after the final `agents.yaml` edit and idempotent on a second run. The money-path bundles are in that set; none was hand-edited.
- **No `.rego` was changed** (`git diff origin/main...HEAD -- "*.rego"` is empty) — the bundles move only because they embed `agents.yaml` and hash it. Policy decisions are unchanged.

JiRaska:
## Split feasibility analysis (2026-08-06)

Investigated whether this PR's 81 files decompose into a mergeable-now subset + a gated remainder. **Verdict: no.** The 72 `openbank-infra/gitops/components/**` OPA-bundle/service.yaml files, `openbank-admin-ui/ai-governance-snapshot.json`, and `docs/compliance/eu-ai-act.md` are all confirmed **mechanical** — running the documented regen sequence (`gen-rules-opa-data.py` then each `gen-*opa-bundle*.sh`, plus `gen-ai-governance-snapshot.py`/`gen-eu-ai-act.py`) against this PR's own source state reproduces them byte-for-byte (`git status --porcelain` empty after regen). No hand-tampering found.

But those 72 files are *derived from* `openbank-libs/governance/agents.yaml`'s edit — and that edit (adding a `model:` field per agent charter, e.g. `model: llama-3.3-70b-versatile`) is itself the real, review-worthy content: it's embedded into every fleet OPA bundle, which is why the restamp is fleet-wide. It can't be carved out as "safe" — it's exactly what the gate exists to catch.

The genuine source-of-change is 11 files:
- `openbank-libs/governance/agents.yaml` — the `model:` charter field
- `openbank-agent-service/.../application/AgentPolicyGate.kt` — `audit()` now writes `"model_id" to modelId` into every `agent.mcp.tool_call` audit event (spot-checked directly against the PR diff, confirmed)
- `CharterRegistry.kt`, `McpEndpoint.kt`, `AgentChatService.kt`, `McpToolRegistry.kt`, `ProposalService.kt`, `AgentPolicy.kt`, `application.yaml`
- 5 test files exercising the above

This is squarely inside the audit/policy-decision path the money-path/authz gate protects — not separable from the reviewable content by construction.

**Practical time-saver for whoever reviews this:** you only need to actually read those ~11 files above (particularly `agents.yaml`'s `model:` additions and `AgentPolicyGate.kt`'s `audit()` change). The other 72 files can be confirmed in one command rather than read line-by-line:
```
git checkout <this-pr-branch>
python3 .github/scripts/gen-ai-governance-snapshot.py   # if present under a similar path
python3 .github/scripts/gen-eu-ai-act.py                 # if present under a similar path
git status --porcelain   # expect: empty
```
Empty output confirms all 72 derived files are exactly what the source changes produce, no manual edits smuggled in.

No merge or split action taken — this PR stays as-is, still correctly gated for your review.

JiRaska:

JiRaska:
Reviewed. Well-tested and honestly documented change. Two things to say, and the second is the substantive one.

### 1. Nothing reads the field today — worth reframing the value claim

`grep -E ": *AuditEventPublisher *\{"` across `main` returns exactly **one** production implementation: `LoggingAuditEventPublisher` in libs-runtime. Its `publish()` logs a fixed format string — `eventId, timestamp, actorId, actorType, operation, resourceType, resourceId, result, traceId` — and never touches `event.payload`. So every `model_id` this PR adds goes into a map the sole publisher drops.

The hash-chain argument in the body is about audit-service's persisted rows, reached over Kafka `audit-events-out` — a path nothing in this fleet publishes to today. The `model_id` that *is* genuinely read is a different thing: the `agent_proposals.model_id` **column**, which the `ProposalService` / `McpToolRegistry` half of this PR does improve.

None of that makes the change wrong — it makes it a latent-trap fix rather than an evidence-chain fix, and framing it that way is the difference between "we were losing attribution" and "we would have been, the day a durable publisher lands". The repo has been burned before by treating a field nothing consumes as corruption.

### 2. The model is declared in two places, and the one the runtime reads has 2 of the 15

`CharterRegistry` is `@ConfigMapping(prefix = "agent.charter")` — it reads `application.yaml`, not `agents.yaml`. Counted on this branch:

```
charters gaining `model:` in agents.yaml   15
agent-id entries in application.yaml        2   (ui-assistant, compliance-officer)
occurrences of `mcp-anonymous`              0
```

So `charters.modelId(actorId)` returns `UNKNOWN_MODEL` for the other 13 — including **`mcp-anonymous`**, the fixed principal the whole MCP tool-call path runs as. The `model:` values added to `agents.yaml` are consumed only by generators (EU AI Act inventory, the OPA bundle hash); nothing under `openbank-agent-service/src/main` reads that file.

That is a second hand-maintained copy of the same list with nothing checking agreement — the repo's own rule about never letting a document keep its own copy of a list that lives elsewhere. `check-agent-charter-registry.sh` does not cover it: it scopes ids in `agents.yaml` against `docs/agents/<id>.md` filenames and says outright it does not check content parity.

Either a parity gate on `model:`, or an explicit note that the 13 are intentionally `unknown` for now. As it stands the PR reads as "every AI-attributed event carries a model id" while the MCP path records `UNKNOWN_MODEL`.

### Checked and clean

- **Tests assert real values**, not presence: `isEqualTo("llama-3.3-70b-versatile")` and `isEqualTo(CharterRegistry.UNKNOWN_MODEL)`. The one `containsKey("model_id")` sits *alongside* an equality check, drawing the absent-vs-unknown distinction deliberately. No `isNotNull()` anywhere. The structural test's own stated limitation is correctly identified and backstopped behaviourally.
- **The declared-not-observed drift is disclosed** in `CharterRegistry.kt:44` and `AgentPolicy.kt:12`. No complaint — just note that `application.yaml`, not `agents.yaml`, is the copy that decides.

One adjacent thing, out of scope but on the same line: `AuditEvent.timestamp` still defaults to `Instant.EPOCH`, and none of the seven emitters this PR touches passes one. So an event this PR now attributes to a model is stamped 1970-01-01. Not caused here, but the PR argues absence-vs-unknown is an evidence gap while the same envelope's "when" is a placeholder.

Branch is `CONFLICTING` with `main`; I reviewed the diff and did not resolve it.

JiRaska:
Pushed (\`0b765d379\`). Two things: the branch was badly conflicted with \`main\`, and the model-id parity gap from review is now gated rather than just noted.

**The merge conflict was 73 files, all derived.** OPA bundles, pod-roll policy-checksum annotations, \`docs/compliance/eu-ai-act.md\`, \`openbank-admin-ui/ai-governance-snapshot.json\` — none carried a hand edit; every one is a generator's output disagreeing with a newer generator's output. Verified rather than assumed for the two that also carry real config (\`customer-edge.yaml\`, \`payments-services.yaml\`): diffed each against its own merge-base and confirmed the only lines that differ are \`policy-checksum\` values. Resolved to \`main\`'s side and regenerated as the **last** step before committing — the order this repo has been burned by getting backwards before.

**The model-id gap is now \`check-agent-model-parity.py\`, not a comment.** \`CharterRegistry\` reads \`application.yaml\`, not \`agents.yaml\`; measured on this branch, 13 charters declare a real model there, \`application.yaml\` covered 2.

I checked why that is not live today rather than assuming it is, and the answer changes the fix. \`mcp-anonymous\` and \`ap2-anonymous\` — the two ADR-0181 phase-1 fixed principals every caller runs as — declare \`model: unknown\` in \`agents.yaml\` **itself**, so the two sides already agree for the one caller that exists today. The other 11 ids are separate microservices (\`openbank-authz-policy-auditor\`, \`openbank-docs-truth-agent\`, \`openbank-finops-agent\`, …) or chat-plane agents this service does not yet route traffic for — none calls \`CharterRegistry.modelId\` at all. The gap is prospective: ADR-0126 phase 2 (per-agent OAuth replacing the fixed principal) is what turns it live, the day a real caller authenticates as one of them.

So the gate is a two-way ratchet in the \`check-outbox-has-writer.py\` shape: a **real drift** between the two files for a covered id is always a NEW finding, never baselined. An id with no \`application.yaml\` coverage is baseline debt with a dated reason. A baseline entry that gains coverage, or whose \`agents.yaml\` value changed, or that no longer exists, is flagged — the list can't rot in either direction.

**Falsified, not just self-tested green.** Self-test 6/6 (matching coverage, baselined gap, a real drift, newly-covered-but-still-baselined, unbaselined gap, stale baseline entry — each case independently). Then on the real tree: edited only \`application.yaml\`'s \`ui-assistant\` model to something wrong, confirmed the gate named it exactly, reverted, confirmed clean.

\`test\`/\`ktlintCheck\`/\`detekt\` all ran on \`openbank-agent-service\`: 148 tests, 0 skipped, 0 failures. Registered in \`gates.yaml\` (group \`registry\`, enforced) and confirmed by \`check-gate-script-registration.py\`.

The \`AuditEvent.timestamp\` defaulting to \`Instant.EPOCH\` observation from the earlier review is real and adjacent, but out of scope here — it is a shared \`libs-domain\` default, not something this PR's diff touches, and fixing it would be its own change.

Issue #3818, opened 2026-08-06
Customer-app dossier check red repo-wide: committed derived version 0.14.0 vs app 0.15.0


JiRaska:
`Customer-app dossier` (ADR-0074) is red on every new PR run, repo-wide and time-based rather than PR-specific:

```
derived-block drift:
  version: committed="0.14.0" → regenerated="0.15.0"
```

`JiRaska/openbank-app` released `0.15.0`, so the committed derived block in `openbank-admin-ui/app-status.json` no longer matches the app source. The check is enforced (`--enforce`), so it blocks every PR until the derived block is regenerated.

Evidence it is not PR-specific:
- PR #3813 touches no dossier file and is red (job ran 2026-08-06T05:40Z).
- PR #3787 shows the same check green from a job that ran 2026-08-05T15:20Z — before the app release.
- Failing job: https://github.com/JiRaska/open-bank-oss/actions/runs/31074794116/job/92530347409

Fix: regenerate with `openbank-admin-ui/scripts/generate-app-status.mjs --app-repo <openbank-app>` (derived data, never hand-edited).

Follow-up worth considering separately: the normal path is openbank-app CI publishing the regenerated artefact back to this repo. That publish did not happen for `0.15.0`, which is why the drift reached every open PR instead of landing as its own PR.

JiRaska:
Duplicate — #3811 (open since 05:17Z, `Refs #2101`) already carries the identical fix: the regenerated `app-status.json` blob hash matches byte for byte (`db5532240`). The systemic half — openbank-app CI never publishing the refresh back, so drift reaches every open PR — is being addressed by #3821 (refresh by pull instead of waiting for a push).

Closing as duplicate; my PR #3820 was closed for the same reason.

Issue #3552, opened 2026-08-03
Customer-app dossier red on every PR again: app-status.json pinned at 0.12.0, openbank-app shipped 0.13.0 (4th recurrence — fix #2101)


JiRaska:
`Customer-app dossier` is red again on every PR. Same single field as #3212, one version later:

```
the committed app-status.json no longer matches openbank-app source (1 field(s)). Regenerate & publish.
  derived-block drift:
    version: committed="0.12.0" → regenerated="0.13.0"
```

Observed on #3546 (a comment-only YAML change that cannot touch the dossier), job [91588538761](https://github.com/JiRaska/open-bank-oss/actions/runs/30782053040/job/91588538761), failing step `Regenerate & diff derived block`. `openbank-admin-ui/app-status.json` on `main` still has `derived.version: "0.12.0"` and `asOf: 2026-07-29`; `openbank-app` has shipped 0.13.0.

## This is recurrence #4, not a new fault

| # | Drift | Resolution |
|---|-------|-----------|
| #2047 | — | manual republish |
| #2097 | — | manual republish |
| #3212 | 0.11.0 → 0.12.0 | manual republish, closed 2026-08-02 |
| this | 0.12.0 → 0.13.0 | — |

#3212 was closed **yesterday** and the drift returned within a day. The interval is just the customer app's release cadence, so hand-publishing buys exactly one release of quiet each time.

## Root cause is already filed and still open: #2101

`openbank-app`'s `app-status` workflow gates its checkout/regenerate/`create-pull-request` steps on `env.PLATFORM_RW_TOKEN != ''`. That secret is not configured, so the publish job runs only its "Skip notice (no platform write token)" step and exits `success` — green, no PR, nothing signalled (ADR-0074 invariant 3). Every dossier change in `openbank-app` therefore fails to propagate, and surfaces here later as a red check on an unrelated platform PR.

**This issue should be closed by fixing #2101, not by a fifth manual republish.** Adding `PLATFORM_RW_TOKEN` to `JiRaska/openbank-app` (fine-grained PAT or App installation token, `contents:write` + `pull-requests:write` on `JiRaska/open-bank-oss`) makes the publish PR open on its own, and this stops recurring.

Per rule 6 and the file's own `_comment`, `app-status.json` is DERIVED — hand-editing `derived.version` here makes the gate green while the two repos still disagree about who owns the value, which is how derived data starts drifting for real (#3212 made the same call).

## Why it keeps costing something

[...]

Issue #3212, opened 2026-08-02
Customer-app dossier red on every PR: app-status.json pinned at 0.11.0, openbank-app shipped 0.12.0


JiRaska:
`Customer-app dossier` is failing on **every** PR since ~#3203 — #3207, #3209, #3210 all red on it, #3203 and earlier green.

Cause is one field, and the gate is doing exactly its job:

```
the committed app-status.json no longer matches openbank-app source (1 field(s)). Regenerate & publish.
  derived-block drift:
    version: committed="0.11.0" → regenerated="0.12.0"
```

`openbank-app` released 0.12.0; `openbank-admin-ui/app-status.json` in this repo still says 0.11.0.

Deliberately not hand-fixing it: `app-status.json` is DERIVED (rule 6, and its own `_comment` says so), and per the `app-status` job's header the publish PR is opened by **openbank-app's** CI. Editing the field here would make the gate green while the two repos still disagree about who owns the value — which is how derived data starts drifting for real.

It is not a required check, so nothing is blocked. But it is red on every PR, which is the state in which people stop reading a check.

JiRaska:
Verified against `origin/main` — the drift this reports is already gone:

```
committed openbank-admin-ui/app-status.json  derived.version = 0.12.0
openbank-app                                 version.txt     = 0.12.0
```

I hit the same red on #3211 a few hours ago and regenerated it locally; by the time I went to open a PR, `main` already carried the regenerated file (a parallel agent got there first) and my copy was byte-identical, so there was nothing to push.

Closing as resolved — but the mechanism that produced it is not fixed, and that is **#2101** (`app-status dossier auto-publish is a silent no-op: PLATFORM_RW_TOKEN not configured in openbank-app`). Until that token exists, every openbank-app release re-creates this exact drift, the `Customer-app dossier` check reds on whoever happens to open the next PR, and it reads like their fault. #2101 needs a repo secret only the owner can set.

JiRaska:
Resolved. `openbank-admin-ui/app-status.json` now reads `0.12.0` and `Customer-app dossier` is green on the last five merged PRs (#3499, #3501, #3502, #3503, #3504).

Closing on the outcome rather than on a fix of my own: the correct owner of that value is openbank-app's publish PR, which is what landed it, and hand-editing the derived file here would have made the gate green while the two repos still disagreed about who owns it.

PR #3820, opened 2026-08-06
chore(admin-ui): regenerate customer-app dossier for app 0.15.0

JiRaska:
## What

Regenerates the committed customer-app dossier (`openbank-admin-ui/app-status.json`) so its `derived` block matches the current `JiRaska/openbank-app` release.

```
-    "version": "0.14.0",
+    "version": "0.15.0",
```

Single field; everything else in the derived block already matched.

## Why

`JiRaska/openbank-app` released `0.15.0`, and the ADR-0074 dossier content check runs with `--enforce`, so the `Customer-app dossier` job is red on **every** open PR regardless of what it touches:

```
derived-block drift:
  version: committed="0.14.0" → regenerated="0.15.0"
```

Failing job: https://github.com/JiRaska/open-bank-oss/actions/runs/31074794116/job/92530347409

It is time-based, not PR-specific: #3813 touches no dossier file and is red (job 2026-08-06T05:40Z), while #3787 shows the same check green from a job that ran 2026-08-05T15:20Z — before the app release.

## How

Derived data is never hand-edited (ADR-0029 #7), so this uses the same generator the "Regenerate & diff derived block" CI step invokes, against a clone of the app source:

```
node openbank-admin-ui/scripts/generate-app-status.mjs --app-repo <openbank-app>
```

The private app repo needs read access — CI gets it from `OPENBANK_APP_RO_TOKEN` (and prints the "Skip notice" step when that secret is absent). Locally the `gh` credential for the repo owner is enough to clone it.

## Verification

Enforce check reproduced red before, green after, on the same command CI runs:

```
node scripts/generate-app-status.mjs --app-repo <app> --against app-status.json --enforce --check
```

- before: `::error title=app-status derived drift::` … exit 1
- after: `✓ derived block matches the committed app-status.json`, exit 0

## Linked issues

Refs #3818

## Note

The normal path for this artefact is openbank-app CI regenerating it and opening the publish PR back into this repo. That did not happen for `0.15.0`, which is why the drift reached every open PR instead of landing as its own PR — captured in #3818 as a follow-up.

### DIFF
--- a/openbank-admin-ui/app-status.json
+++ b/openbank-admin-ui/app-status.json
@@ -12,7 +12,7 @@
   "asOf": "2026-07-29",
   "appRepo": "JiRaska/openbank-app",
   "derived": {
-    "version": "0.14.0",
+    "version": "0.15.0",
     "useFakeData": false,
     "edgeBaseUrl": "https://customer.open-bank.tech",
     "keycloakIssuer": "https://kc.open-bank.tech/realms/openbank-customers",


JiRaska:
Duplicate of #3811 — same artefact, byte-identical.

```
git diff --name-only origin/main...origin/<branch>
  #3811  openbank-admin-ui/app-status.json
  #3820  openbank-admin-ui/app-status.json

shasum of the regenerated file on both branches:  e6760acb6daa
```

Both are healthy (merge ref present, 21 checks each), so this is not a rescue — it is two agents
regenerating the same derived artefact 32 minutes apart. #3811 is the earlier one, so closing this
in its favour. Nothing is lost: merging #3811 produces exactly this file.

Worth noting for whoever merges: this is republish **#5** of the same single field (#2047, #2097,
#3212, #3552, and now these). The recurrence itself is addressed by #3821, which moves the refresh
to a pull from this repo — it would have opened this PR by itself, and would not have opened it
twice.

Issue #1269, opened 2026-07-16
Auto-deploy GitOps PRs strand unsigned: sign-commits: true cannot sign with a user PAT, required_signatures blocks the merge


JiRaska:
## Symptom

Auto-deploy GitOps PRs sit open indefinitely with **all required checks green and auto-merge enabled**. Nothing merges, so ArgoCD never sees the new image tag and the fleet silently stops updating. Today seven were stranded at once (#1200, #1237, #1238, #1251, #1252, #1258, #1267), the oldest for ~3.5 hours.

The PRs are blocked by the `main-protection` ruleset's `required_signatures` rule, because the bot-created commit is unsigned:

```
verified=false  reason=unsigned
```

## Impact

- **Deploys stop without any red signal.** Green checks + auto-merge ON reads as "healthy"; the only symptom is an open PR nobody is looking at. Today `openbank-customer-edge` (#1254, consent documents for onboarding) sat undeployed until the PR was re-signed by hand.
- Every deploy since 2026-07-11 has needed a **manual human GPG re-sign** to land. That is silent, recurring toil, and it puts a human in the loop of an automated deploy path.
- It self-perpetuates: merging one deploy PR triggers the next auto-deploy run, which opens another unsigned PR.

## Evidence

`sign-commits: true` has been set on both `auto-deploy.yml` and `admin-ui-deploy.yml` since #806 (2026-07-11), whose title was literally *"sign auto-deploy GitOps commits so required_signatures stops stranding them"*. **It has never worked.**

Sampling every deploy PR since #806 and checking the *branch* commit (not the squash commit on `main` — GitHub signs that one itself, so it always reads `verified=true` and proves nothing):

| committer on branch commit | signed? | what it actually is |
|---|---|---|
| `JiRaska <jiri@iraska.cz>` | never | the bot's API-created commit |
| `jiri.raska <jiri@iraska.cz>` | always | a human's local `git commit --amend -S` |

The two are distinguishable by the committer *name*: `JiRaska` is the GitHub account login (API path), `jiri.raska` is the local git config `user.name`. There is no intermittency — the bot's commit is unsigned 100% of the time, and every "signed" deploy PR is one a human rescued.

From the `Open gitops PR` step log of run `29499271154`, the action does take the API path, and reports the failure itself without failing the step:

```
Creating tree objects for local commit 46f7d4b2ad5969591474179c6a3945c6798934bc
[...]

PR #3821, opened 2026-08-06
ci(admin-ui): refresh the customer-app dossier by pull instead of waiting for a push

JiRaska:
## The recurrence, not the instance

`openbank-admin-ui/app-status.json` is DERIVED from `JiRaska/openbank-app` source (ADR-0074). The
same single field has now drifted five times:

| # | Drift | Resolution |
|---|---|---|
| #2047 | — | manual republish |
| #2097 | — | manual republish |
| #3212 | 0.11.0 → 0.12.0 | manual republish |
| #3552 | 0.12.0 → 0.13.0 | manual republish |
| #3811 | 0.14.0 → 0.15.0 | manual republish, open now |

The interval is the customer app's release cadence, so each manual republish buys exactly one
release of quiet.

## Why it recurs

The transport is a **push**: openbank-app's own `app-status` workflow regenerates and opens a
publish PR here. Its checkout/regenerate/`create-pull-request` steps are gated on
`env.PLATFORM_RW_TOKEN != ''`. That secret is not configured, so the job runs only its skip notice
and exits **success** — green, no PR, nothing signalled (#2101, open; ADR-0074 invariant 3).

So the drift is discovered days later as a red `Customer-app dossier` check on someone else's PR.

## The change

Reverse the direction. This repo already holds **both** credentials the refresh needs:

- `OPENBANK_APP_RO_TOKEN` — reads the app source; `ci.yml`'s dossier check already uses it, which
  is how that check can regenerate and detect the drift in the first place;
- the GITOPS App (`GITOPS_APP_ID` / `GITOPS_APP_PRIVATE_KEY`) — a **signed** self-PR, the same rung
  `auto-deploy.yml` and `admin-ui-deploy.yml` established (#1269/#1276).

So the refresh depends on no secret in the other repo. Daily schedule + `workflow_dispatch`; when
the regenerated artefact matches, it opens nothing.

`sign-commits: true` because `main-protection` enforces `required_signatures`, and the App token
because a PR opened with `GITHUB_TOKEN` does not trigger the required checks (anti-recursion).
`add-paths` is scoped to the one artefact — `npm ci` and the generator both write into the tree.

## The one thing it must not do

Reproduce the silent no-op one repo over. A missing credential **fails**, naming the secret:

```
::error title=app-status::cannot refresh the dossier — missing secret(s): OPENBANK_APP_RO_TOKEN
```

## Scope

- `ci.yml`'s dossier check is **unchanged** — it still fails a PR whose committed dossier disagrees
  with the app source. This removes the reason it keeps being true, it does not weaken the check.
- **#2101 stays worth fixing** and is not closed by this. It is no longer on the critical path for
  this drift.
- #3811 (the current manual republish) is unaffected and should still merge; this workflow would
  have opened that PR by itself.

## Validation

- `actionlint` clean — probe validated against a known-bad file first (it reports
  `unexpected key "run" for step to execute action`), so its silence here is evidence.
- `check-workflow-run-step-size` clean; largest `run:` in this file is 470 chars.
- Pushed to a branch and counted runs: zero, i.e. GitHub parsed it. The workflow declares no push
  trigger, so an accepted file produces no run and a rejected one still would.

Refs #3552, #2101, #3212, ADR-0074


JiRaska:
Supporting evidence for this PR — with one correction that makes the case stronger, not weaker.

The body says the openbank-app publish job *"runs only its skip notice and exits success — green, no PR, nothing signalled"*. That describes an earlier shape of the workflow. Today it is worse in three independent ways, each fatal on its own, and they mask each other in this order:

**1. The push trigger never fires.** The publish job is `on: push` filtered to `version.txt` / `AppConfig.kt` / `app-status.yaml`. `version.txt` is written exclusively by release-please, which runs with `secrets.GITHUB_TOKEN` — and GitHub does not trigger workflows from a push made with `GITHUB_TOKEN` (anti-recursion). Measured, not inferred: the `0.15.0` release commit (author `github-actions[bot]`) has **zero** workflow runs on its SHA — not a skipped one, none at all. The last `push`-event run of that workflow is from 2026-07-31, across five releases (0.11.0 → 0.15.0) that each rewrote `version.txt`.

**2. It targets an archived repository.** Both checkout steps use `repository: JiRaska/open-bank`, which is `archived: true` (last push 2026-07-29). The sibling workflows in the same directory — `edge-contract-sync.yml`, `edge-contract-gate.yml` — already point at `open-bank-oss`; the dossier transport was left behind in that migration. Even with a token and a live trigger, it would open a PR against a read-only repo.

**3. The token is absent**, as the body states — and the job is no longer a silent no-op about it: it now `exit 1`s deliberately, and the last three `push` runs are `failure`, not `success`.

So the drift is not "published quietly to nowhere" — the transport does not execute at all.

**A fourth would surface the moment 1–3 were fixed:** `sign-commits: true` on `peter-evans/create-pull-request` signs only when given a GitHub **App** token — GitHub signs the GraphQL/Contents API path, never a user PAT. With `main-protection` enforcing `required_signatures`, a PAT-driven publish PR would land unsigned and strand with green checks (the #1276 shape).

This is the argument for the pull direction rather than against it: reversing it clears all four at once and depends on no secret in the other repo. Worth doing after this merges — delete `app-status.yml`'s publish job in openbank-app rather than leave a workflow that cannot run, aimed at a repo that no longer accepts writes.

JiRaska:
**Recurrence #6 landed while this PR sat waiting — about an hour after #5 was fixed.**

```
committed="0.15.0" → regenerated="0.16.0"
```

#3811 merged `0.15.0` roughly an hour ago; `openbank-app` has since shipped `0.16.0`, and
`Customer-app dossier` is red again on every open PR I sampled — including this one.

**This corrects the sizing in the PR body.** I wrote that each manual republish "buys exactly one
release of quiet", assuming the customer app's cadence was measured in weeks. Measured over one
session it is **hours**: 0.13.0 → 0.14.0 → 0.15.0 → 0.16.0. At that rate the manual loop barely
outlives the PR that performs it, and the check is red across the whole queue more often than it is
green — which is the state in which people stop reading a check.

Worth separating from the other red on these PRs, since they look identical from the check list: the
41 PRs with a red `Trivy` are pure branch staleness and clear with `git merge origin/main`. This one
does **not** — #3838's branch carries `0.15.0` byte-identical to `main`, and merging changes nothing.
The check is behaving correctly; the artefact is genuinely out of date with its source.

Nothing in this PR changes as a result. Recording it because it turns the argument from "this saves
a recurring chore" into "without this, the check is usually red", and because a reviewer reading the
body's cadence estimate would be reading an understatement I wrote.

PR #3823, opened 2026-08-06
ci: stop three consumer jobs writing 1.1 GB Gradle-home caches into an over-ceiling pool

JiRaska:
## Measured first, as the issue asked

Re-measured against the live API, not the numbers in #3299:

```
active caches   47 entries
total           11 360 214 746 B = 10.58 GB     (ceiling 10 GB)
```

Worse than at filing (10.11 GB / 75 entries) — fewer entries, each larger. Eviction is the steady
state, and it is invisible: it surfaces as a slow build, never as a red check.

Where it goes:

| key prefix | entries | MB |
|---|---:|---:|
| `gradle-dependencies-v1` | 6 | 4017 |
| `setup-java-Linux-x64-gradle` | 3 | **3288** |
| `gradle-sbom` | 1 | 718 |
| `node-cache-Linux-x64-npm` | 3 | 599 |
| `gradle-build-cache-v1` | 12 | 462 |

All three `setup-java-Linux-x64-gradle` entries sit on **PR refs**, ~1.1 GB each.

## Why those three are the churn engine

They come from `actions/setup-java`'s own `cache: gradle`, which **saves** a whole Gradle user home
keyed on a hash of `**/*.gradle*`.

For `services-ci.yml`'s `verification-metadata` job that key **can never hit**. The job runs only
when a module `build.gradle.kts` changed:

```
mods="$(git diff --name-only "$base"...HEAD | grep -E '^openbank-[^/]+/build\.gradle\.kts$' ...)"
```

so the job's trigger *is* a change to the key's input. It wrote a fresh ~1.1 GB entry per PR and
read back only a restore-key partial.

And a warm Gradle home is the thing that check is designed not to lean on — it passes
`--refresh-dependencies`, with its own source stating why:

> delete a known entry, run without `--refresh-dependencies`, and it passes.

`pitest` and `pact-drift-check` are consumers too. None of the three has any reason to be one of the
lanes that *maintains* the cache — `_service-ci.yml`'s self-hosted lane and `auto-deploy.yml`
already do that.

## The change

All three move to `gradle/actions/setup-gradle` (same pinned SHA the repo already uses) with
`cache-read-only: true` — the pattern `api-fuzz.yml` and `api-fuzz-authenticated.yml` established.
They read the shared entry and write nothing.

Expected: **10.58 GB → ~7.3 GB**, i.e. under the ceiling with headroom, so the remaining entries
stop losing a permanent eviction race.

`dependabot-verification-metadata.yml` is deliberately **left alone** — it is the writer, it is
dependabot-gated, and a warm shared entry is what makes the read-only consumers hit.

## Not claimed

- **This does not measure hit rate**, which is the issue's first acceptance criterion. It removes
  three writers whose exact-key hit is impossible or pointless; it does not tell you how often the
  remaining reads hit. That is still open on #3299.
- **The saving is projected, not observed.** 3.29 GB is what those three entries hold right now;
  the pool will not shrink until they age out or are evicted. Re-measure after this merges.
- **There is a real cost**: `--refresh-dependencies` re-checks metadata but still reuses
  checksum-matching artifacts, so a read-only restore is not free-for-free. A cold read-only hit on
  the shared entry should be at least as good as today's restore-key partial, but I have not timed
  it.
- Unrelated and untouched: `services-ci.yml` uses floating `actions/setup-java@v5` /
  `actions/checkout@v4` tags where the repo pins SHAs elsewhere. Out of scope here.

## Validation

- All three parse; `actionlint` finding sets diffed against `origin/main` and are **identical**
  (2 pre-existing shellcheck findings in `services-ci.yml`, both far from this change) — the sets
  are diffed rather than the output post-filtered, so a new finding could not hide.
- `check-workflow-run-step-size` clean.
- All three jobs are literal `ubuntu-latest`, so `check-gradle-user-home-isolation` does not apply.

Refs #3299


JiRaska:
**Update — this PR now also corrects the writer registry, and the correction is the interesting
part.**

`check-gradle-cache-writers.py` (the #3299 budget gate) landed on `main` while this branch was open.
It declares all three of the jobs touched here as `setup-java`, justified as:

> `actions/setup-java` keys its Gradle cache on the build files, so it re-uses one entry across runs
> rather than storing a new one per run; it is not part of the per-run churn this budget limits.

For `services-ci.yml::verification-metadata` that is **false**, and the live pool says so. The job
runs only when a module `build.gradle.kts` changed — i.e. exactly when the key's own input changed —
so an exact-key hit is impossible by construction. Measured 2026-08-06:

```
setup-java-Linux-x64-gradle   1181 MB   refs/pull/3810/merge
setup-java-Linux-x64-gradle   1176 MB   refs/pull/3808/merge
setup-java-Linux-x64-gradle    931 MB   refs/pull/3797/merge
                             = 3.29 GB  (29% of the 10 GB ceiling)
```

Three entries, all on **PR refs** — not one shared entry re-used across runs. And that check passes
`--refresh-dependencies` deliberately, with its own source saying *"delete a known entry, run
without `--refresh-dependencies`, and it passes"* — so a warm Gradle home is precisely what it is
designed not to lean on.

pitest is a weekly scheduled sweep and the keying argument holds better there; pact-drift-check runs
on PRs and is exposed to the same churn. All three are pure consumers either way, so `read-only` is
correct for all three, and each carries its own reason rather than a shared one.

**Budget unchanged: 5/5 write-capable.** This moves entries between two non-write-capable modes.

```
Gradle cache budget OK — 14 declared usages, 5/5 write-capable:
  _service-ci.yml::build, auto-deploy.yml::build-push, fleet-lint.yml::fleet-lint,
  ghcr-publish.yml::publish, security.yml::codeql
```

The gate's own `--self-test` still passes, and `main` was merged in first — scope verified as exactly
the three workflow files plus this script, nothing else.

This PR still requires human review because it touches `.github/workflows/`.

JiRaska:

Checked the specific claim underpinning this PR: that demoting these three jobs to `cache-read-only: true` doesn't strand them without any writer to read from. Confirmed on `origin/main`:
- `_service-ci.yml` (the self-hosted lane cited in this PR's comments as a writer) actually sets `cache-read-only: ${{ runner.environment != 'self-hosted' }}` — i.e. the self-hosted lane writes NO cache at all (`cache-disabled` there), and even the GitHub-hosted PR lane in that file is already read-only. So `_service-ci.yml` is not itself a writer either.
- `auto-deploy.yml` is confirmed as the actual writer: `cache-read-only: false` with the comment "deploy jobs update the cache too" — so there is a real writing lane for the three demoted jobs to read from, just not the one this PR's own comments attribute it to (they credit "`_service-ci.yml`'s self-hosted lane, auto-deploy.yml" together, but only the latter actually writes).

This doesn't change the correctness of the fix — demoting three consumer-only jobs from writing ~1.1 GB each into an over-ceiling pool is still the right move regardless of which lane keeps the shared entry warm — but the in-repo comment in `pact-drift-check.yml`/`pitest.yml`/`services-ci.yml` slightly overstates who the writer is. Worth a one-line correction so a future reader doesn't go looking for a write in `_service-ci.yml` and not find one.

The `verification-metadata` job's rationale is the strongest of the three and independently verifiable from the diff alone: it only runs when a `build.gradle.kts` changed, which is exactly the input the old cache key hashed, so an exact-key hit was structurally impossible — that job was pure waste, not just suboptimal.

Verified: `_service-ci.yml` and `auto-deploy.yml` on `origin/main` for who actually writes vs. reads the shared Gradle-home cache pool, confirming the mechanism this PR depends on exists, and that the fix targets the real writer even though the writer is misattributed in one comment. Did not verify the claimed 3.29 GB / 29%-of-ceiling measurement against a live cache API listing.

PR #3826, opened 2026-08-06
fix(delegation): refuse non-SOLO approvalPolicy — ADR-0232 D8 co-signing is enforced nowhere

### LINKED ISSUE #3800: Delegation grants accept dailyLimit/monthlyLimit that nothing enforces

JiRaska:
## Stacked on #3613 — read that one first

Base is `fix/delegation-reject-unenforced-cumulative-ceilings`, not `main`. #3613 refuses
`dailyLimit`/`monthlyLimit`; this PR refuses the **third** unenforced constraint on the same
aggregate, found by doing the audit #3800 asks for in its last bullet. It reuses #3613's
`DelegationUnsupportedConstraintException` machinery rather than inventing a parallel rule, which
is why it is stacked instead of independent. **Retarget to `main` if #3613 lands first.**

## What I did NOT do

I did not re-implement #3800's headline fix. #3613 already implements it, is green, and carries
`Closes #3800`; a second copy of one rule is the drift failure this repo has been bitten by. I
verified #3613's premise independently instead (below) and then worked the residual nobody had
touched.

## The defect

`approvalPolicy` (`SOLO | ANY_ONE | ALL | N_OF_M`) and `requiredApprovals` are accepted on
`POST /api/v1/delegations`, validated for self-consistency, persisted, echoed back, and rendered
in admin-ui. **Nothing reads either to make a decision.**

Every non-test, non-build match in the repo, so the claim is checkable:

| Layer | Site |
| --- | --- |
| REST DTO (accept) | `DelegationDtos.kt:46-47`, `DelegationResource.kt:107-108` |
| Command | `DelegationUseCases.kt:37-38` |
| Aggregate (field) | `DelegationGrant.kt:85-86` |
| Aggregate (self-consistency only) | `DelegationGrant.kt:117-118` — `N_OF_M` demands `requiredApprovals >= 2` |
| Entity / DDL (persist) | `DelegationGrantEntity.kt:58,61,136-137,179-180`, `V1__init_delegation.sql:10-11,34` |
| DTO / spec (echo) | `DelegationDtos.kt:76-77,99-100`, `openapi.yaml` |
| admin-ui (display) | `delegations/[id]/page.tsx:101`, `GrantView.tsx:24-25` |

The only other hits in the repo are `openbank-governance-auditor`'s `requiredApprovals`, an
unrelated local variable about PR review counts.

The read side is empty in the two decisive places:

1. `DelegationGrant.covers` (`:129-133`) — the answer behind `POST /check` — consults capability
   and `perTransactionLimit` only.
2. `DelegationOffered` does not carry the policy, and account-service's projection
   (`DelegationProjectionEntity.kt`) has no column for it — I listed its columns; they stop at
   `per_tx_limit_amount`/`per_tx_limit_currency`. So no downstream service *could* enforce it.

**The consequence is on a money path.** `SavingsProposalService.decide`
(`openbank-account-service`, `:80-121`) is the ADR-0232 D8 maker-checker and the one flow that
could honour a co-signing rule. It requires `account.partyId == decidedByPartyId`, calls
`approvalStore.decide` once, and emits `SavingsWithdrawalApproved` — a **single** owner decision
releases the withdrawal whatever the grant said. D8's promise that both parents must approve a
shared-goal withdrawal is a number nobody counts, and the DB even constrains it to be >= 2, which
makes it look enforced.

## Why refuse rather than enforce

Same argument #3613 makes, and it holds harder here: the policy is not replicated to any
projection, so the enforcement point cannot even see it. Building a counter first would mean
choosing its home before the delegated money path exists, and it would read as implemented while
returning the same false assurance. Delivering D8 means replicating the policy onto the
projection and counting decisions where the money moves — in that order.

`SOLO` stays accepted: it is the default and promises no second approver, so it is the one value
that is honest today.

## Red first

Tests written and run **before** the production change:

```
DelegationServiceTest > offer refuses every multi-party approvalPolicy, not just N_OF_M() FAILED
    java.lang.AssertionError at DelegationServiceTest.kt:257
DelegationServiceTest > offer refuses an N_OF_M approvalPolicy because no service counts approvals() FAILED
    java.lang.AssertionError at DelegationServiceTest.kt:238
DelegationServiceTest > offer refuses an unenforced approvalPolicy before spending the SCA challenge() FAILED
    java.lang.AssertionError at DelegationServiceTest.kt:275

34 tests completed, 3 failed
```

After the change: 34 pass. The asymmetry is the point — the control
`offer still accepts the default SOLO approvalPolicy` is **green in both runs**, so the tests mean
"refuses multi-party policies", not "stopped accepting `approvalPolicy`".

## Spec

`info.version` 1.2.0 -> 1.3.0. Read off **live `main`** (1.1.0) as well as the parent branch
(1.2.0) immediately before writing this; 1.3.0 is the next free number in the stack. Needs
re-reading if this sits.

The enum values stay listed. Narrowing it to `[SOLO]` is what oasdiff reads as a breaking
restriction — the trap #3613's last commit was fixing — so the refusal is stated in `description`
and `x-unsupported-values` instead.

## Local gate

`:openbank-delegation-service:detekt :ktlintCheck :test` — BUILD SUCCESSFUL, and I checked all
three tasks actually appear in the output rather than trusting the exit code. `:quarkusBuild` also
green (no new bean, but cheap insurance).

## What I did NOT verify — please weigh these

- **No test proves the 400 is what a client actually receives.** These are use-case tests; they
  assert the exception, not the HTTP status. If
  `DelegationUnsupportedConstraintExceptionMapper` were not selected by JAX-RS this would surface
  as a 500. It is #3613's mapper, already carrying `CUMULATIVE_LIMIT_UNSUPPORTED`, so I have added
  no new wiring risk — but neither PR has an IT proving the mapper is registered. This is #3613's
  own self-declared weakest link and I have not closed it.
- **Existing rows are untouched.** Grants already stored with `ANY_ONE`/`ALL`/`N_OF_M` keep
  rehydrating and keep being echoed. Nobody has counted them or identified those grantors. Same
  residual #3800 records for the ceilings.
- **I did not update ADR-0232 D8.** #3613 adds a measured delivery-status section for D1 only;
  D8 still describes co-signing as a capability. Editing the same region would have conflicted
  with the parent. Follow-up, not done here.
- **I did not touch the other audit findings** (below). They are reported, not fixed.
- I did not run the full fleet build, the OPA/governance gates, or any contract replay locally —
  relying on CI for those.

## The rest of the audit #3800 asked for

D4/D5/D7/D8 swept for the same "present at every layer except the enforcing one" shape. Treat the
non-`approvalPolicy` rows as **findings needing their own issues**, not as claims this PR acts on:

| Element | Verdict |
| --- | --- |
| `approvalPolicy` / `requiredApprovals` (D8) | FALSE-PROMISE — **fixed here** |
| `exposure`: `redactionRules`, `maxViews`, `watermark`, `allowDownload` (D7) | FALSE-PROMISE — stored in `V1__init_delegation.sql`, echoed; no view counter, no redaction, no download gate anywhere in the repo |
| `OBJECT_READ` capability (D7) | FALSE-PROMISE — grantable; neither document-service nor statement-service references delegation or consumes the event stream |
| `ACCOUNT_PROPOSE_PAYMENT` capability (D8) | FALSE-PROMISE — grantable; no proposal endpoint, and account-service's own vocabulary omits it. Contrast `SAVINGS_PROPOSE_WITHDRAW`, which is genuinely wired |
| `DELEGATION_MANAGE` capability (D5) | FALSE-PROMISE — grantable; the D5 "statutory representative on a LEGAL_ENTITY grantor" rule has no code, and the eligibility DTO carries no `partyType`, so it cannot be expressed |
| Four-eyes on high-value payment grants (D4) | FALSE-PROMISE flag / ABSENT threshold — `authz.four-eyes.enforce` exists but only interpolates a startup log line; `delegation.offer` is in no `four_eyes.verbs`/`actions` list, and the service wires no `ApprovalStore` |
| Notification to both parties on every transition (D4) | ABSENT — notification-service does not consume `openbank.delegation.events` |
| Sanctions/PEP screening at grant time and on grantee status change (D5) | ABSENT — and structurally impossible today: the service has no `@Incoming` consumer at all, so a grantee later sanctioned keeps every ACTIVE grant until a human suspends |
| Dual consent, revocation, outbox events, KYC eligibility gate (D4/D5) | IMPLEMENTED |
| `SAVINGS_PROPOSE_WITHDRAW` maker-checker (D8) | IMPLEMENTED — except the N_OF_M half this PR refuses |

The ABSENT rows are honest gaps, not this defect class — nothing pretends they work. The
FALSE-PROMISE rows are the ones that mislead a customer.

Refs #3800

### DIFF
--- a/openbank-delegation-service/src/main/kotlin/com/openbank/delegation/application/usecase/DelegationService.kt
+++ b/openbank-delegation-service/src/main/kotlin/com/openbank/delegation/application/usecase/DelegationService.kt
@@ -28,6 +28,7 @@ import com.openbank.delegation.domain.event.DelegationRenounced
 import com.openbank.delegation.domain.event.DelegationRevoked
 import com.openbank.delegation.domain.event.DelegationSuspended
 import com.openbank.delegation.domain.event.EventMoney
+import com.openbank.delegation.domain.model.ApprovalPolicy
 import com.openbank.delegation.domain.model.DelegationCheckResult
 import com.openbank.delegation.domain.model.DelegationGrant
 import jakarta.enterprise.context.ApplicationScoped
@@ -66,6 +67,7 @@ class DelegationResourceOwnershipException(message: String) : RuntimeException(m
 class DelegationUnsupportedConstraintException(val code: String, message: String) : RuntimeException(message) {
     companion object {
         const val CODE_CUMULATIVE_LIMIT_UNSUPPORTED = "CUMULATIVE_LIMIT_UNSUPPORTED"
+        const val CODE_APPROVAL_POLICY_UNSUPPORTED = "APPROVAL_POLICY_UNSUPPORTED"
     }
 }
 
@@ -100,6 +102,7 @@ class DelegationService(
         val now = OffsetDateTime.now(clock)
         requireCallerIs(command.callerPartyId, command.grantorPartyId)
         rejectUnenforcedCeilings(command)
+        rejectUnenforcedApprovalPolicy(command)
         verifyResourceOwnership(command)
         verifyEligibility(command)
         // SCA last of the three gates: it SPENDS the challenge, so a request that was going to be
@@ -369,6 +372,35 @@ class DelegationService(
                 message = "${unenforced.joinToString(" and ")} cannot be accepted: this platform enforces only " +
                     "perTransactionLimit. No service counts cumulative spend against a grant, so a ceiling set " +
                     "here would never be applied to any payment. Omit the field (ADR-0232 D1/D6).",
+            )
+        }
+    }
+
+    /**
+     * ADR-0232 D8's co-signing promise, refused for the same reason as the cumulative ceilings:
+     * nothing counts approvals against a grant.
+     *
+     * `approvalPolicy` is accepted, checked for self-consistency (N_OF_M demands
+     * `requiredApprovals >= 2`), persisted, echoed and rendered — and never read by a decision.
+     * `DelegationGrant.covers` consults capability and `perTransactionLimit` only;
+     * `DelegationOffered` does not carry the policy; account-service's delegation projection has
+     * no column for it. So the one flow that could honour it — `SavingsProposalService.decide`,
+     * the D8 maker-checker — releases the withdrawal on a SINGLE owner decision no matter what
+     * the grantor chose. "Oba rodiče musí schválit výběr" is a number nobody counts.
+     *
+     * SOLO stays accepted: it is the default and it promises no second approver, so it is the one
+     * value that is honest today. Delivering the rest means replicating the policy onto the
+     * projection and counting decisions where the money moves — in that order, or the counter is
+     * a second unread field.
+     */
+    private fun rejectUnenforcedApprovalPolicy(command: OfferDelegationCommand) {
+        if (command.approvalPolicy != ApprovalPolicy.SOLO) {
+            throw DelegationUnsupportedConstraintException(
+                code = DelegationUnsupportedConstraintException.CODE_APPROVAL_POLICY_UNSUPPORTED,
+                message = "approvalPolicy ${command.approvalPolicy} cannot be accepted: no service counts " +
+                    "approvals against a grant, so a co-signing requirement set here would never be applied — " +
+                    "a single owner decision still releases the money. Only SOLO is enforced today. " +
+                    "Omit the field (ADR-0232 D8).",
             )
         }
     }
--- a/openbank-delegation-service/src/main/resources/openapi.yaml
+++ b/openbank-delegation-service/src/main/resources/openapi.yaml
@@ -5,7 +5,7 @@ info:
     Customer-to-party delegated access (ADR-0232): granular, SCA-bound grants over
     products (accounts, savings goals, cards) and single objects (payment, statement,
     document). Lifecycle: OFFERED → ACTIVE → SUSPENDED/REVOKED/EXPIRED/RENOUNCED/DECLINED.
-  version: 1.2.0
+  version: 1.3.0
 servers:
   - url: http://localhost:8126
 tags:
@@ -272,8 +272,21 @@ components:
             enum: [ACCOUNT_READ_BALANCES, ACCOUNT_READ_TRANSACTIONS, ACCOUNT_INITIATE_PAYMENT,
                    ACCOUNT_PROPOSE_PAYMENT, CARD_VIEW, CARD_MANAGE_LIMITS, SAVINGS_DEPOSIT,
                    SAVINGS_WITHDRAW, SAVINGS_PROPOSE_WITHDRAW, OBJECT_READ, DELEGATION_MANAGE]
-        approvalPolicy: { type: string, enum: [SOLO, ANY_ONE, ALL, N_OF_M] }
-        requiredApprovals: { type: integer }
+        approvalPolicy:
+          type: string
+          enum: [SOLO, ANY_ONE, ALL, N_OF_M]
+          default: SOLO
+          description: >-
+            Only SOLO is enforced. Any other value is rejected with 400 APPROVAL_POLICY_UNSUPPORTED:
+            no service counts approvals against a grant, so a co-signing requirement set here would
+            be stored and echoed while a single owner decision still releases the money. The enum
+            values remain listed because ADR-0232 D8 describes them as intended; they are not built.
+          x-unsupported-values: [ANY_ONE, ALL, N_OF_M]
+        requiredApprovals:
+          type: integer
+          deprecated: true
+          description: >-
+            Meaningful only with approvalPolicy=N_OF_M, which is rejected. Nothing counts approvals.
         perTransactionLimit:
           $ref: '#/components/schemas/MoneyDto'
           description: >-
--- a/openbank-delegation-service/src/test/kotlin/com/openbank/delegation/application/usecase/DelegationServiceTest.kt
+++ b/openbank-delegation-service/src/test/kotlin/com/openbank/delegation/application/usecase/DelegationServiceTest.kt
@@ -16,6 +16,7 @@ import com.openbank.delegation.application.port.out.ScaChallengeClient
 import com.openbank.delegation.application.port.out.ScaChallengeSnapshot
 import com.openbank.delegation.domain.event.DelegationOffered
 import com.openbank.delegation.domain.event.EventMoney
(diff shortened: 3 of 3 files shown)


JiRaska:
Independently re-derived both halves of this before endorsing it, since a delegated finding is a
hypothesis until checked.

**No reader.** Every occurrence of `requiredApprovals` / `approvalPolicy` under `src/main` is
plumbing — command DTO, domain model, JPA entity, and the two mapper directions:

```
DelegationUseCases.kt:38        val requiredApprovals: Int? = null
DelegationService.kt:103        requiredApprovals = command.requiredApprovals
DelegationGrant.kt:86           val requiredApprovals: Int? = null
DelegationGrant.kt:117          require(approvalPolicy != N_OF_M || (requiredApprovals != null && requiredApprovals >= 2))
DelegationGrantEntity.kt:61,137,180
```

The only `require` is **self-consistency**, not enforcement — and it is precisely what makes the
field read as enforced to anyone auditing the model.

**The consequence is on the money path.** `SavingsProposalService.decide`
(`openbank-account-service`) gates on ownership alone:

```kotlin
if (account.partyId != decidedByPartyId) {
    throw ProposalForbiddenException("only the account owner can decide a withdrawal proposal ...")
}
```

No approval count, no consultation of the grant. So a grantor who sets `N_OF_M` with
`requiredApprovals = 3` gets a withdrawal released on **one** decision, and the API accepted the
value, validated it, stored it and echoed it back.

Refusing rather than fake-enforcing is the right call, and it is the same remedy #3613 applies to
the amount limits — consistent, not two different answers to one class of defect.

Flagging for whoever reviews: this is stacked on the unmerged #3613. Retarget it to `main` **before**
that parent merges, or a `--delete-branch` on the parent closes this PR irreversibly (#3055/#3063).

JiRaska:
**Review: no blocking findings.** The central factual claim holds — I re-derived it independently rather than reading the table.

`git grep approvalPolicy` over `openbank-delegation-service/src/main` on `origin/main` returns exactly the sites listed, and the decisive point checks out: `DelegationService.kt:107` is the **only** write path that sets `approvalPolicy` (there is no amend/renew endpoint that could bypass the new refusal), while `DelegationGrant.kt:126` is the only rule that reads it — and it is a self-consistency check (`N_OF_M ⇒ requiredApprovals >= 2`), not an enforcement. So "accepted, persisted, echoed, never consulted by a decision" is accurate, and refusing at the single write path is complete rather than partial.

Notes, none of them blockers:

1. **`openapi.yaml` — this is a behavioural breaking change shipped under a MINOR bump.** A client that sent `approvalPolicy: ALL` yesterday gets 400 today. `oasdiff` will not see it because the enum is deliberately left intact and the refusal is expressed in `description` + the non-standard `x-unsupported-values` (nothing reads that key). That is the right call to avoid the enum-narrowing trap, but it means the contract gate's classification is not evidence about compatibility here — worth being explicit that the minor is a judgement, not a measurement.
2. **`info.version: 1.3.0` needs re-reading immediately before merge.** Stacked on #3613, whose base is at 1.2.0 while live `main` is 1.1.0. If #3613 lands after this, or anything else touches the delegation spec meanwhile, 1.3.0 is either a skipped number or a taken one. This repo has been bitten by a version consumed by an **already-merged** PR, which scanning open PRs does not catch — read it off `origin/main` at merge time.
3. **`openapi.yaml:325`** (the response schema's `approvalPolicy: { type: string }`) keeps the old bare form; only the request schema gained the description. Minor doc asymmetry — a client reading the response schema sees no hint that only SOLO is real.
4. Existing stored non-SOLO grants keep rehydrating and being echoed, and nobody has counted them. Acknowledged in the body; it is the residual that matters most to a customer who was told two approvals were required.
5. The residual you flag yourself is the right one: no test proves the 400 rather than a 500 — `DelegationUnsupportedConstraintExceptionMapper` registration is unproven by either this PR or #3613. One `@QuarkusTest` hitting `POST /api/v1/delegations` with `approvalPolicy: ALL` and asserting the status code would close it for both.

Verified without finding a problem: the refusal is placed **before** `verifyResourceOwnership`/`verifyEligibility`/the SCA consume, so the challenge is not spent on a request that was going to be refused — and there is a dedicated test asserting exactly that; `ApprovalPolicy.SOLO` is the default on the command, the DTO and the aggregate, so clients omitting the field are unaffected; the `offer still accepts the default SOLO approvalPolicy` control genuinely distinguishes "refuses multi-party" from "stopped accepting the field", which is the assertion that makes the other three mean something.

*Automated agent review — not an approval. Stacked on #3613; retarget to `main` if that lands first. Money-path 2-human-approval rule still applies.*

PR #3837, opened 2026-08-06
fix(lending): make the compliance-pack four-eyes decision atomic

JiRaska:
## Scope — this is the *other half* of #3467

#3467 names two defects. The replica-divergence one is already addressed by the open **#3644**
(`CompliancePackRefresher`), which I did not duplicate. This PR takes the second, which #3467 and
#3644 both explicitly split out:

> `decide()` reads the row, checks `require(entity.state == PROPOSED)`, then writes in a separate
> reactive context. Two checkers approving concurrently can both pass that guard and both merge.

The two changes do not overlap: #3644 touches `CompliancePackConfig`, `CompliancePackRefresher`,
`CompliancePackRegistry` and `application.yaml`; this touches
`CompliancePackActivationRepository` and `CompliancePackActivationService`. Verified against
`origin/main` at `4ade3a08d`, no shared file.

## The defect

`CompliancePackActivationService.decide()` reads the proposal through `findById` (`@WithSession`),
decides against that snapshot **in memory**, and writes through `save` (`@WithTransaction`). Two
transactions, no lock, no version column, no conditional predicate. Two decisions arriving together
both observe `PROPOSED`, both pass `require(entity.state == PROPOSED)`, and both write.

The harm is not only that the audit trail names one checker while two acted. `decide()` calls
`registry.activate(decided)` on the approve leg **whether or not that leg's write survived**, so a
`REJECTED` row can coexist with a pod that is enforcing the pack the rejection refused — and
`CompliancePackRegistry` has no de-activation path (ADR-0212 D3), so the pod enforces the rejected
rule set until it restarts. On a money-path segregation-of-duties control that is not a stale cache.

## Proven RED first — and the red output

`CompliancePackConcurrentDecideIT` was written and run **before** any production change, against
`origin/main`. It races a real approve against a real reject on the same proposal over real HTTP,
12 rounds with a fresh proposal each round, synchronised on a `CyclicBarrier`.

```
java.lang.AssertionError: [a second decision was accepted on an already-decided proposal]
Expecting empty but was: ["v900 accepted=[200, 200]",
    "v901 accepted=[200, 200]",
    "v902 accepted=[200, 200]",
    "v903 accepted=[200, 200]",
    "v905 accepted=[200, 200]",
    "v906 accepted=[200, 200]",
    "v907 accepted=[200, 200]",
    "v909 accepted=[200, 200]",
    "v910 accepted=[200, 200]",
    "v911 accepted=[200, 200]"]
```

**10 of 12 rounds accepted both decisions.** Rounds 904 and 908 happened to serialise — that is the
test being honest rather than the defect being intermittent.

The assertions were then merged into one list so a run reports every violation class it saw rather
than only the first; the red output above is from the earlier two-assertion form, which is why it
shows only the double-apply class.

## The fix

`compareAndSetDecision` re-tests `PROPOSED` **inside** the UPDATE:

```
update CompliancePackActivationEntity
   set state = :state, decidedBy = :decidedBy, decidedAt = :decidedAt,
       decisionReason = :reason, updatedAt = :updatedAt
 where id = :id and state = :from
```

The database evaluates the predicate and the assignment in one statement, so exactly one of any
number of concurrent deciders claims the row. It returns the row count; `decide()` treats `0` as a
refusal — the same `400` the caller would have received a second later — and only the claimant
reaches `registry.activate`.

Named parameters rather than Panache's positional `update(String, vararg Any)` because
`decisionReason` is nullable and the vararg cannot carry a null.

**Why not a `@Version` column:** it would work, but it needs a Flyway migration on a live table, and
optimistic locking would surface as an `OptimisticLockException` that has to be mapped back to a
sensible status. The predicate is already exactly the invariant (`state == PROPOSED`), costs no
schema change, and the refusal path is the one that already exists.

**Why not a row lock:** `SELECT ... FOR UPDATE` would hold a transaction open across the pack
compile in `toProposal()`, which is the expensive part of this call.

## Verification

| | result |
|---|---|
| `CompliancePackConcurrentDecideIT` on unpatched `main` | **FAILED** — 10/12 rounds double-applied (output above) |
| same test, with the fix | **PASSED**, three independent runs |
| `CompliancePackActivationIT` (existing, 4 tests) | PASSED |
| `CompliancePackActivationServiceTest` (existing 7 + 1 new) | PASSED |
| `:openbank-lending-service:test` (full suite) | PASSED |
| `ktlintCheck`, `detekt` | PASSED |
| `:openbank-lending-service:quarkusBuild` | PASSED (CDI wiring — unit tests do not check it) |
| `threat-model-coverage` gate | PASSED |

The new unit test — `losing the decision race refuses AND leaves the registry untouched` — pins the
part the IT can only observe indirectly: the service decides in memory *before* it writes, so when
the write is refused it is already holding a fully-approved proposal, and it must not activate it.

The in-memory fake repository had to change to model this honestly. It aliased the caller's entity,
so `rows[id].state` was whatever the caller last mutated — a fake testing that field would report a
transition as committed before it was written, which is the exact confusion the conditional update
exists to remove. It now tracks committed state separately.

## No schema, no API change

- **No Flyway migration.** The predicate uses the existing `state` column, so the `migrate-at-start`
  trap does not apply here.
- **No `openapi.yaml` change.** `400` is already the declared response for
  `POST /compliance-packs/proposals/{id}/decide`; the concurrent-decision refusal reuses it via the
  existing `IllegalArgumentException` → 400 mapping in `CompliancePackResource`.
- `version.txt` untouched — release-please owns it. This is a `fix` touching `src/main`, so it
  releases.

## What I did NOT verify, and what could still be wrong

Stated plainly rather than buried.

1. **The IT is a timing test and cannot promise to go red.** It does not widen the window
   artificially — no sleep, no test hook in production code — so a round in which the two requests
   serialise passes vacuously. It cannot produce a false *red* (an accepted second decision on a
   PROPOSED row is a defect however the threads landed), but on a slower or busier runner the
   observed rate could differ from the 10/12 measured here, and in principle a run could be green
   against broken code. **This is the honest answer to "could your test pass against the OLD code":
   yes, with low probability per round, ~1/6 empirically per round and so very unlikely across 12.**
   I chose not to inflate the round count further because each round is a full propose+race+two
   queries and the suite already costs ~2 min.
2. **I did not test more than two concurrent deciders.** The conditional UPDATE is correct for N by
   construction, but only N=2 is measured.
3. **I did not verify behaviour under `READ COMMITTED` vs other isolation levels explicitly.** The
   test ran against the Testcontainers Postgres default (`READ COMMITTED`), which is also the
   deployed default; a single-statement conditional UPDATE is atomic there, but I did not exercise
   any other level.
4. **I did not run this against a real multi-replica deployment.** Both racers hit one JVM. That is
   the right scope — the defect is a database-level lost update, not a JVM one, and the fix is in
   the statement — but it is not a live multi-pod proof.
5. **`registry.activate` is still not transactional with the row.** If the process dies between the
   committed UPDATE and the in-memory activate, this pod is behind until convergence or restart.
   That is #3644's territory, not this PR's, and I have not closed it.
6. **I did not survey other four-eyes decide paths in this service for the same shape.**
   `LendingService.decide()` (collateral) and the origination `lending.approve` path look
   structurally similar and may carry the identical read-check-write defect. I did not measure them,
   so I am not claiming either way — worth its own issue rather than a claim here.
7. **I did not re-verify #3644 is still current** beyond confirming it is open, mergeable, and
   touches no file this PR touches.

## Merge expectation

`openbank-lending-service` is money-path: 2 human approvals by rule. This PR is **not expected to
merge** on my action, and that is the correct outcome.

Refs #3467, #3644, ADR-0212 D4

### DIFF
--- a/docs/threat-models/openbank-lending-service.md
+++ b/docs/threat-models/openbank-lending-service.md
@@ -245,10 +245,16 @@ two authenticated principals to alter.
 | Repudiation — "who activated this rule set?" | Durable Postgres record per activation (`compliance_pack_activation`: maker, checker, reason, timestamps) + canonical SHA-256 `content_hash` pinned into the audit evidence (ADR-0214); unlike the Redis-backed ADR-0155 layer this trail is permanent |
 | DoS — boot bricked by a corrupt activation row | Deliberate fail-loud: boot refuses to start over a corrupt activation rather than originate unprotected; operator remediation is to fix the row (the in-memory registry cannot silently run with a partial rule set) |
 | Elevation of privilege — maker self-approves | Server-side identity from the JWT subject (never request body); segregation enforced twice: `Proposal.approve` and registry re-assertion |
+| Tampering — two decisions arriving together are BOTH applied, and a rejected pack is enforced anyway (#3467) | The decision is a conditional UPDATE (`... where id = :id and state = PROPOSED`), so exactly one of any number of concurrent deciders claims the row and the rest are refused. Previously `decide()` tested the state against a snapshot read in a separate transaction and then wrote unconditionally, so two decisions both passed; worse, the approve leg called `registry.activate` whether or not its write survived, leaving a REJECTED row beside a pod enforcing the pack that rejection refused — and `CompliancePackRegistry` has no de-activation path (ADR-0212 D3), so that pod enforced it until restart. `CompliancePackConcurrentDecideIT` measures the race over real HTTP; on the previous code it observed both decisions accepted in 10 of 12 rounds |
 
 **Residual risk:** pack enforcement ships behind the bootstrap flag (default `false`) until the
 CZ reference pack is seeded and activated — the guard cannot protect origination while off.
 Tracked as the named bootstrap follow-up (ADR-0212 D4).
+
+**Residual risk (concurrency):** the conditional UPDATE makes the *decision* atomic; it does not
+make the in-memory registry shared. A pack approved on one replica still reaches its siblings only
+by the convergence path, so the at-most-once guarantee here is about the durable decision, not about
+every pod agreeing at the same instant (#3467, and see the replica-convergence work tracked there).
 
 ## 9. Customer self-service origination intake (ADR-0211) — STRIDE supplement
 
--- a/openbank-lending-service/src/main/kotlin/com/openbank/lending/application/port/out/CompliancePackActivationRepository.kt
+++ b/openbank-lending-service/src/main/kotlin/com/openbank/lending/application/port/out/CompliancePackActivationRepository.kt
@@ -20,6 +20,15 @@ interface CompliancePackActivationRepository {
     fun findById(id: UUID): Uni<CompliancePackActivationEntity?>
     fun findByState(state: ProposalState): Uni<List<CompliancePackActivationEntity>>
     fun findActivated(): Uni<List<CompliancePackActivationEntity>>
+
+    /**
+     * Apply a four-eyes decision to [entity] **only if** the stored row is still
+     * [ProposalState.PROPOSED], as a single statement. Returns the number of rows claimed: `1` when
+     * this caller won the decision, `0` when someone else already decided it.
+     *
+     * The caller must treat `0` as a refusal and must not perform any side effect of the decision.
+     */
+    fun compareAndSetDecision(entity: CompliancePackActivationEntity): Uni<Int>
 }
 
 @ApplicationScoped
@@ -42,6 +51,37 @@ class JpaCompliancePackActivationRepository :
     override fun save(entity: CompliancePackActivationEntity): Uni<CompliancePackActivationEntity> =
         Panache.getSession().flatMap { it.merge(entity) }
 
+    /**
+     * The decision leg does NOT go through [save].
+     *
+     * `save` is a blind upsert: it writes whatever the caller loaded, whenever the caller gets round
+     * to it. The checker leg reads the row in one transaction (`findById`, `@WithSession`), decides
+     * against that snapshot in memory, and writes in another — so two decisions arriving together
+     * both observe `PROPOSED`, both pass the service's `require(state == PROPOSED)` guard, and both
+     * write. A lost update on a segregation-of-duties control, and worse than a lost update: the
+     * approve leg activates the pack in the in-memory registry whether or not its write survived, so
+     * a REJECTED row can coexist with a pod enforcing the pack that rejection refused.
+     *
+     * Making the state test part of the UPDATE closes the window without a lock, a version column or
+     * a schema change — the database evaluates `state = PROPOSED` and the assignment in one
+     * statement, so exactly one of any number of concurrent deciders can claim the row and the rest
+     * get `0` back. `CompliancePackConcurrentDecideIT` measures it; on the previous code it observed
+     * both decisions accepted in 10 of 12 rounds.
+     */
+    @WithTransaction
+    override fun compareAndSetDecision(entity: CompliancePackActivationEntity): Uni<Int> =
+        Panache.getSession().flatMap { session ->
+            session.createMutationQuery(DECIDE_HQL)
+                .setParameter("state", entity.state)
+                .setParameter("decidedBy", entity.decidedBy)
+                .setParameter("decidedAt", entity.decidedAt)
+                .setParameter("reason", entity.decisionReason)
+                .setParameter("updatedAt", entity.updatedAt)
+                .setParameter("id", entity.id)
+                .setParameter("from", ProposalState.PROPOSED)
+                .executeUpdate()
+        }
+
     @WithSession
     override fun findById(id: UUID): Uni<CompliancePackActivationEntity?> = find("id", id).firstResult()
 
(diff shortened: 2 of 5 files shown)


JiRaska:
Followed up the residual this PR's author flagged and did not measure:

> I did not survey the other four-eyes paths in this service (`LendingService.decide()` for
> collateral, the origination `lending.approve` path). They look structurally identical and may
> carry the same defect — I measured neither, so I claim nothing.

Measured now: **they do**, and the service has no optimistic locking at all. `LendingService.advance`
reads via `findById`, decides in memory on `existing.status`, and writes via `update` — which carries
its own `@WithTransaction`, so read and write are separate transactions — and the write blind-sets
the status with no predicate. `@Version` appears nowhere in `openbank-lending-service` (probe
validated against account-service, which does use it).

Filed as **#3850** rather than widening this PR, since #3837 is already reviewable and scoped, and
the origination path deserves its own reproduction before a fix. #3837's approach —
re-test the expected state inside the UPDATE, row count `0` means refusal — transfers directly.

JiRaska:
**Review: no blocking findings.** The mechanism does what the description claims.

Verified the claim against the code, not the prose:
- `JpaCompliancePackActivationRepository.compareAndSetDecision` is a single HQL mutation with `where id = :id and state = :from` — the predicate and the assignment are one statement, so it is a genuine compare-and-set, not a read-then-write. Named parameters are indeed required (`decisionReason` is nullable).
- `CompliancePackActivationService.decide` treats `claimed != 1` as a refusal **before** `registry.activate`, which is the half that mattered: the loser can no longer enforce a pack whose decision never committed. The new unit test pins exactly that, and the fake was reworked to track committed state separately rather than aliasing the caller's entity — that change is necessary, not cosmetic; without it the fake could not refuse the second decision.
- `listActive()` reads `registry.allActive(...)`, i.e. the in-memory registry — so the IT's second assertion class ("row is REJECTED yet the pack is enforced") is genuinely observing the registry, not re-reading the same table it just asserted on. That was the thing most likely to have made the money-path assertion vacuous, and it isn't.

Three non-blocking notes:

1. `CompliancePackActivationService.kt:106` — a concurrency refusal is an `IllegalArgumentException` and so surfaces as **400**, indistinguishable from "your pack JSON is malformed". A caller cannot tell "retry is pointless, someone decided it" from "your request was wrong". 409 would be the honest status; it costs an `openapi.yaml` bump, which is presumably why you didn't. Worth a conscious decision rather than a side effect of reusing the existing mapping.
2. `CompliancePackConcurrentDecideIT.propose()` writes `content_hash = "race-probe-$version"`, a non-canonical hash. Harmless today because `toProposal()` recompiles from `payload` and never checks it — but ADR-0214 pins the content hash into the audit evidence, so if the decide path ever verifies it this IT fails for a reason that has nothing to do with the race.
3. The IT activates 12 pack versions (900-911) for `CZ/CONSUMER_CREDIT` into a process-wide `CompliancePackRegistry` shared with every other `@QuarkusTest` in the same JVM. `activePack()` returns the max version, so after this class runs the "active CZ consumer-credit pack" is v911. Benign right now — the payloads differ only in the version integer, and `CompliancePackActivationIT` uses 999 — but it is ordering-coupled by luck rather than by design.

Verified without finding a problem: no Flyway migration is needed (existing `state` column); the bulk update sets `updatedAt` explicitly rather than relying on a lifecycle callback; the entity is loaded in a different session so the bulk write bypassing the persistence context is correct here; no file overlap with #3644 (confirmed against its current file list); `save`/`merge` is untouched on the propose path, so the persist-vs-merge trap is not reintroduced.

*Automated agent review — not an approval. `openbank-lending-service` is money-path: the 2-human-approval rule still applies.*

JiRaska:
Second review pass. **No blocking findings.**

**No atomicity claim to fail here** — unlike its sibling #3876, this path writes no outbox. The only side effect is `registry.activate(decided)`, and the diff correctly moves it *behind* a successful `claimed == 1` (`CompliancePackActivationService.kt:108-113`). Cross-replica divergence stays open and is correctly deferred to #3644.

**persist-vs-merge**: clean. `save` is already `Panache.getSession().flatMap { it.merge(entity) }` and untouched; the new `compareAndSetDecision` is a bulk HQL mutation. Neither path re-opens the trap.

**Four-eyes holds — but through a path worth knowing about.** `decide` itself has no `checker != proposedBy` comparison; separation is enforced one layer down in `Proposal.approve`/`reject` (`openbank-libs-domain/.../governance/MakerChecker.kt:62,71`), which throw `MakerCheckerViolation`, mapped to 422 in `CompliancePackResource.kt:60`. So the same principal cannot be maker and checker, and `id = :id and state = :from` binds the approval to the row. I checked this specifically because the service KDoc promises "a DIFFERENT principal" while the method body shows no such test — it is delegated, not missing, and that is worth a pointer in the KDoc so the next reader does not conclude the guard is absent.

One residual, pre-existing: unlike `propose`, `decide` never requires `checker` to be non-blank, so a blank principal would pass the `checker == proposedBy` test and be recorded as `decidedBy = ""`. Gated by `@RolesAllowed`, so low reachability.

**Test layer is right**: `CompliancePackConcurrentDecideIT` races real HTTP against a real DB, and the new unit test covers only what the IT cannot observe (registry untouched on refusal). The fake-repository rework to track committed state separately was necessary, not cosmetic.

PR #3838, opened 2026-08-06
feat(campaign,notification): tell a delivered message from an accepted handoff (ADR-0239)

### LINKED ISSUE #3663: campaign send log cannot distinguish an accepted handoff from a delivered message
Split out of #3581, which fixed the two shipping defects (a party UUID used as the SMTP envelope,
and a refused handoff leaving no send-log row) but deliberately did not answer this.

## The gap

`SendOutcome.SENT` in campaign-service's send log means "the request was accepted onto
`openbank.notification.requests`". It does not mean the customer received anything, and the two can
differ for reasons that are entirely routine on the notification side:

- the ADR-0198 D4 marketing consent gate suppresses the send (`status = SUPPRESSED`)
- the party has no resolvable e-mail address (`status = FAILED`, the fail-closed path added in #3662)
- the mailer refuses or the message bounces (`status = FAILED`)

In every one of those the campaign console still shows SENT, and the notification row that
disagrees is never read back. Two sides, permanently out of step, and only the quiet one is right.
That matters more here than on a transactional rail: the suppression/consent trail is what the
compliance claims rest on, and a marketing funnel is read as a measurement.

## What a fix needs to decide

1. **Correlation.** The notification request payload carries `partyId`, `channel`, `template`,
   `recipient`, `variables` — nothing that identifies the campaign, the step, or the send-log row.
   Either the payload gains a correlation id, or the outcome event carries enough for the campaign
   to find its own row. This is the actual design question.
2. **The event.** A delivery-outcome event from notification-service (SENT / SUPPRESSED / FAILED
   plus a reason), versioned backward-compatibly, that anything — not only campaign — can consume.
3. **The vocabulary.** Whether campaign's log keeps SENT for the handoff and gains a separate
   confirmed state, or renames the handoff state. This ripples into the admin-ui funnel
   (`OUTCOMES` in `openbank-admin-ui/src/app/campaigns/[id]/page.tsx`) and the campaign
   `openapi.yaml`, so it is a contract change, not a rename.
4. **Idempotency.** Delivery is at-least-once and a redelivery persists a fresh notification row,
   so the campaign side must tolerate more than one outcome per attempt.

Worth an ADR rather than a PR — item 1 is a shared-contract decision that outlives this one
consumer.

JiRaska:
Closes #3663. Implements ADR-0239 end to end — the three slices the issue thread named (correlation id, the outcome event, the campaign-side consumer + column) in one vertical change, because none of them is observable without the other two.

## The defect

`SendOutcome.SENT` records that notification-service accepted the request onto `openbank.notification.requests`. It has never meant delivery, and the two diverge routinely: the ADR-0198 D4 consent gate suppresses, the party has no resolvable address (#3662), the mailer refuses. The campaign console showed `SENT` in every one of those cases and the notification row that disagreed was never read back.

**Premise verified against live `origin/main`, not assumed.** Before this branch, `NotificationRequest` carried nothing identifying the campaign; `notification_outbox` had **no writer anywhere in `src/main`** and the `notification-events-out` channel had **no topic binding**, so notification-service produced nothing at all. `SendRecord` had no delivery field. All three still true at `1ff472626`.

## Red first

With the outbox write removed from `NotificationConsumer.markStatus`, the new `NotificationConsumerIT` case fails:

```
NotificationConsumerIT > a completed send emits a correlated delivery-outcome event (ADR-0239 D2)() FAILED
    java.lang.AssertionError at NotificationConsumerIT.kt:312
    no delivery-outcome event was emitted for a completed send (issue #3663)
```

Restored, the same suite is green.

## What changed

**D1 — correlation.** `NotificationRequest.correlationId` (nullable, additive). campaign-service sets it to the send-log row id, which now has to be minted *before* the handoff rather than by `record` after it. Persisted on the notification row (`V12`) as its own column rather than inside the unmapped `metadata` JSONB — a deliberate, documented deviation from the ADR's wording, taken because `metadata` is not mapped on the entity at all and a JSONB key cannot be indexed as cheaply as this lookup wants.

**D2 — the event.** `openbank.notification.outcomes.v1`, written to `notification_outbox` in the *same transaction* as the status change it reports (ADR-0003). This is the first writer that outbox has ever had. Emitted for every terminal transition, correlated or not — it is a shared contract. Its own `NotificationOutcome` enum rather than a reuse of `NotificationStatus`, which carries `PENDING` (not an outcome) and lacks `SUPPRESSED` (a string the consumer has always written to the column).

**D3 — vocabulary.** `outcome` is untouched: it is the frequency cap's input (ADR-0219 D2) and must not change meaning because a mailer bounced. The row gains `deliveryStatus`/`deliveryReason`/`deliveryUpdatedAt` (`V3`), campaign `openapi.yaml` 1.7.0 → 1.8.0, and the console grows a **Delivery** column beside **Handoff**. The stale `SendRecord.outcome` enum (missing `SUPPRESSED_LIST` since ADR-0219) is corrected in the same bump.

**D4 — idempotency.** `DeliveryTransition.next` is a pure function with its own test table. The topic is partitioned by `notificationId`, **not** correlation id, so two outcomes for one send have no guaranteed order — first-terminal-wins does not depend on order; last-write-wins would have. `BOUNCED` after `CONFIRMED` is the one refinement.

Infra: new `KafkaTopic`, a `Write` grant for notification-service (its first), a `Read` grant plus a **separate consumer group** for campaign-service — two channels sharing one group makes two disjoint subscriptions one member set, so a rebalance on either topic churns the other.

## Verification run

- `:openbank-campaign-service:` `detekt` + `ktlintCheck` + `test` — all three appear in the task list, all green.
- `:openbank-notification-service:` `detekt` + `ktlintCheck` + `test` (incl. the Testcontainers `NotificationConsumerIT`) — green. `quarkusBuild` on both, so the new CDI injection point is validated.
- `run-gates.py --group` over `api data kotlin registry lint gitops` — clean. `check-event-contract-coverage.py`: the new producer topic is covered by the new AsyncAPI file, not baselined. `check-kafka-dotted-keys.py`: no new occurrence (the new channels use nested keys).
- admin-ui: `npm test` 81 files / 1100 tests green, `npm run lint` 0 errors.

## What I did NOT verify — read this before merging

- **Nothing ran against a real broker or a real cluster.** The campaign-side consumer has never consumed a Kafka record: its parsing and routing are unit-tested against a fake repository, and `DeliveryTransition` is tested as a pure function, but the `@Incoming` wiring, the new consumer group, the ACLs and the topic exist only on paper here. The end-to-end claim "a suppression turns the console row from PENDING to SUPPRESSED" is **untested** — the strongest thing proven is that each half does its job in isolation.
- **`applyDeliveryOutcome` was never executed against Postgres.** It is reactive Panache read-decide-write in one transaction; per this repo's own war stories that is exactly the shape a mocked repository cannot vouch for. campaign-service has no Testcontainers IT harness comparable to notification's, and building one was out of scope for this PR. If the `find`/mutate path is wrong, every unit test still passes.
- **The V3 and V12 migrations have not been applied to any database**, only compiled against. `migrate-at-start` behaviour for these two services was not re-verified.
- **`BOUNCED` has no producer.** Nothing ingests bounces; it is declared so consumers must already tolerate it. Its presence in the contract is not evidence bounce handling exists, and the AsyncAPI says so in place.
- **`deliveryStatus` will read `PENDING` for every existing row and for every gate-denied send, permanently** — correctly (nothing was handed off), but it means the column is not a health metric and an operator could misread a wall of PENDING as an outage. ADR-0239's own "Negative" section asks for a staleness signal for genuinely stuck sends; that is **not** built here.
- **The request payload is a hand-built map on one side and a data class on the other.** Nothing in the build compares them. `correlationId` was added to both by hand; a future field can drift silently.
- The admin-ui column was verified by tests and lint, **not** by looking at the rendered page.
- `npx tsc --noEmit` reports one pre-existing error about a CI-generated `origination-graph.json`, unrelated to this change.


JiRaska:
**Picked this up after the authoring session stopped mid-fixup and finished it.** Two collisions
with `main` had appeared while the branch was open, and both are silent to git — a clean merge
cannot raise either.

**1. Flyway version collision.** `V3__campaign_stop_condition.sql` landed on `main`; this branch
carried its own `V3__send_log_delivery_status.sql`. Two different V3s is not a merge conflict —
different filenames, no overlapping lines — it is a boot failure. Renumbered to **V4**. Safe
precisely because it has never been applied anywhere: it exists only on this unmerged branch, so no
checksum or schema history depends on it.

Checked the notification-service side separately: `V12__notification_correlation_id.sql` **is** the
next free number there (`V10`/`V11` are already on `main`), so no renumbering was needed. Worth
saying because the obvious probe reads that directory wrong — `V12__` sorts lexically *before*
`V6__`, so a `ls | tail` hides it and makes V12 look like a gap.

**2. The spec version was consumed by an already-merged PR.** The branch claimed `1.8.0`; `main`
holds it now. Moved to **1.9.0**, read off live `main` rather than this branch's base — the
api-contract gate classifies against the creation-time base, so a competing bump that merged first
is invisible from here. The two prose lines naming `1.8.0` as "added in" moved with it: a version
string in a description is a claim to a client about which deployment carries the field, so leaving
them would have documented the wrong release.

**Verification — tasks confirmed to have actually executed**, since Gradle stops at the first
failure and an `UP-TO-DATE` task is one that did not run (my first attempt reported
`BUILD SUCCESSFUL in 4s` having run no tests at all):

```
:openbank-campaign-service:ktlintCheck  :openbank-campaign-service:detekt          (--rerun-tasks)
:openbank-notification-service:ktlintCheck  :openbank-notification-service:detekt
:openbank-campaign-service:test  :openbank-notification-service:test               (--rerun-tasks, 1m09s)
BUILD SUCCESSFUL
```

Post-merge scope checked for silent deletions, since a clean merge can drop a neighbouring list
entry with nothing printed: `.release-please-manifest.json` 56 = 56 entries, `gates.yaml`
108 = 108 ids, diff scope 27 files as intended. All four commits signed.

**Not verified, stated rather than implied:**

- No live broker — the consumer is exercised by unit tests and an IT, not against deployed Kafka.
  The topic and mTLS manifests are declarations here.
- No multi-service end-to-end run; the two halves are tested independently.
- I re-ran both suites but did **not** re-revert the implementation to re-observe the red, so the
  original red-then-green evidence is inherited rather than re-derived by me.
- This touches `openbank-infra/gitops` (new topic + mTLS), so merging changes cluster state.

JiRaska:
## Merge conflicts with `main` resolved (39f5731e0)

`main` moved under this branch (`d5fdaf3f` added a `PUSH` channel value and a `DRY_RUN` send outcome). Four files conflicted. Every one is a **union**, not a pick.

| File | main's intent | this branch's intent | resolution |
|---|---|---|---|
| `domain/model/Enrolment.kt` | multi-line `SendOutcome` with `DRY_RUN` + its rationale KDoc | `DeliveryStatus` enum, `DeliveryTransition` object, delivery fields on `SendRecord` | main's enum kept verbatim; branch's `DeliveryStatus`/`DeliveryTransition` kept beside it. `outcome` (handoff) and `deliveryStatus` (confirmed fate) remain two facts, per ADR-0239 D3 |
| `port/out/Ports.kt` | `channel: Channel` parameter on `requestSend` | required `correlationId: UUID` | signature carries **both** |
| `workflow/CampaignJourneyActivitiesImpl.kt` | `if (dryRun) record(DRY_RUN) else { handoff }` | send-log id minted *before* the handoff so it can go on the wire as the correlation id | id minted before the dry-run branch; `DRY_RUN`, `SENT` and `FAILED` all record on that one id. A dry run hands nothing off, so its `deliveryStatus` stays `PENDING` — correct, no message ever left |
| `resources/openapi.yaml` | `enum: [..., DRY_RUN, ...]`, `channel: enum [EMAIL, PUSH]` | `SUPPRESSED_LIST` (present in `SendOutcome` since ADR-0219, missing from the spec) + its comment | `SENT, DRY_RUN, SUPPRESSED_CAP, SUPPRESSED_QUIET_HOURS, SUPPRESSED_CONSENT, SUPPRESSED_LIST, FAILED`; `[EMAIL, PUSH]` preserved; comment kept |

### The silent drop the merge *did* make

`CampaignJourneyActivitiesImplTest.kt` merged **clean** — no conflict, nothing printed — and was wrong: its mockk stubs carried five `any()` against a now six-parameter `requestSend`. The Kotlin compiler caught it. Fixed in the same commit (plus a ktlint re-wrap of the resulting long line).

### Verification actually run

```
GRADLE_USER_HOME=~/.gradle-local ./gradlew --rerun-tasks \
  :openbank-campaign-service:{ktlintCheck,detekt,test} :openbank-notification-service:test
BUILD SUCCESSFUL in 1m 12s
```
Tasks confirmed present in the output (not `UP-TO-DATE`): `ktlintMainSourceSetCheck`, `ktlintTestSourceSetCheck`, `detekt`, both `:test`. From the JUnit XML: campaign **84 tests**, notification **127 tests**, **0 failures / 0 errors**.

Post-merge assertions:
- `grep -c 'enum: \[EMAIL, PUSH\]'` = **1**; `- DRY_RUN` = 1; `- SUPPRESSED_LIST` = 1
- `.release-please-manifest.json` = **56** entries (unchanged); `.github/gates/gates.yaml` = **110** `- id:` lines (branch had 108, main 110 — the merge *gained* main's two, dropped none)
- `git diff --name-only origin/main` = 27 files, all campaign / notification / their contracts, gitops kafka topic+mTLS, and the admin-ui campaign page. Nothing unexpected.
- `openapi.yaml` `info.version` re-read off **live** `main` immediately before pushing: main `1.8.0`, this branch `1.9.0`. Still the next free version.
- All 5 commits report `verification.verified = true` via the API.

### What I did NOT verify — stated plainly

- **No integration test, no running service.** Both `:test` suites are unit-level. The `V4__send_log_delivery_status.sql` migration was not applied to a real database, and the `NotificationOutcomeConsumer` was not driven by a real broker — the class of defect this repo repeatedly finds only with a real-DB IT is untested here.
- **The `DRY_RUN` + correlation-id interaction is reasoned, not exercised.** No test asserts that a `DRY_RUN` row gets a correlation id it will never receive an outcome for. It is harmless by construction (nothing is published, so nothing can echo the id back), but "harmless by construction" is an argument, not a measurement.
- **`:quarkusBuild` was not run**, so CDI/ArC wiring of the merged constructor (`dryRun` + the outcome consumer) is unproven locally; CI is the first thing to check it.
- **admin-ui was not built or tested.** `campaigns/[id]/page.tsx` auto-merged and I only read the diff; I did not run `npm test` there. If the funnel renders `DRY_RUN` or `SUPPRESSED_LIST` from a hardcoded label map, that map may be short.
- **The contract/pact and OpenAPI gates were not run locally** — the `1.8.0 -> 1.9.0` classification is asserted from reading live `main`, not from an `oasdiff` run.
- **`asyncapi.yaml` and the gitops Kafka manifests** in the diff auto-merged; I inspected the file list, not every hunk.
- Structured-file counts were checked for the two files with known baselines. Other lists the merge touched (`docs/adr/index.json`, `prompts/registry.yaml`, `.trivyignore`) were **not** entry-counted.

### Could a resolution still have dropped intent?

The honest residual risk is `CampaignJourneyActivitiesImpl.kt`, the only conflict where I wrote new control flow rather than concatenating two blocks. If `main` intended a dry run to consume no send-log id at all, my version differs — it records a real row with a real id. I judged that better (the console needs the row), but it is a decision, not a merge. The other three are mechanically checkable and were checked.

Issue #3596, opened 2026-08-03
can-i-deploy --latest main resolves to unbuilt bookkeeping versions, so an untouched service can never converge


JiRaska:
## Summary

`can-i-deploy --latest main` resolves, for a service nobody is touching, to a **version at a commit that never built that service**. Such a version has no pacts and no verifications, so the gate reports `???` and the classifier labels it `PENDING_BUILD` — permanently. The service can never converge, and re-verifying does not help because the pointer moves on the next merge.

## Evidence

`openbank-fraud-service` has been undeployed all day (#3376 merged 2026-08-02 ~17:54Z). Two consecutive reconcile ticks asked about two *different* versions, and neither commit touches the service:

| tick | version the gate asked about | what that commit actually is | fraud files in it |
|---|---|---|---|
| 02:29Z (run 30775760617) | `f000a22` | `fix(ci): shorten the deploy-drift label description…` (#3361) | 0 |
| 06:20Z (run 30784844936) | `0a64027` | `fix(security-scanner): run Flyway at start…` (#3350) | 0 |

Both rows read:

```
openbank-fraud-service | 0a64027... | openbank-transaction-service | ??? | ??? |
##[error]can-i-deploy: openbank-fraud-service NOT deployable [PENDING_BUILD]
```

Between those two ticks I dispatched `verify-provider.yml` for **both** providers of fraud's pairs — `openbank-fraud-service` (provider to sepa-payment) and `openbank-transaction-service` (provider to fraud). Both runs succeeded. The next tick still said `???`, because by then `--latest main` had moved from `f000a22` to `0a64027`.

**That is the trap: verifying "the latest" is chasing a pointer that every subsequent main merge moves.**

## Why it is not just latency

`PENDING_BUILD`'s own message says the reconcile "re-drives it automatically" and `probe-pact-version.sh`'s header calls the class self-clearing. For a service that is being *touched*, that is true — its build publishes a version at the same sha auto-deploy is deploying. For a service that is **not** being touched, it is false: every main merge mints a newer unbuilt version, so the block regenerates faster than any verification can clear it.

This is the mechanism behind the #1985 co-deploy set never converging. Those services are not deadlocked on each other's contracts (all eight were classified `PENDING_BUILD`, see #3501) — they are all stuck behind this same moving pointer.

## What I verified, and what I did not

[...]

Issue #3306, opened 2026-08-02
can-i-deploy compares against the DEPLOYED provider version, but verifications publish against current main — consumers block permanently


JiRaska:
## Summary

`can-i-deploy` compares a consumer against the provider version **deployed in the environment**, but provider verifications are published against **whatever main SHA the verify build ran at**. On a repo where `main` advances every few minutes, those two SHAs never coincide — so a consumer whose provider needs re-verification is blocked permanently, not transiently.

Measured on `openbank-fx-service`: **five** auto-deploy attempts over ~10 hours, every one blocked at `can-i-deploy`, with the fix sitting merged on `main` and its image built and pushed to ECR the whole time.

## Evidence

The gate's own output, unchanged across all five runs:

```
==> can-i-deploy openbank-fx-service (--latest main) → environment sandbox
    no pact version for this commit — falling back to latest/main (ADR-0092)

CONSUMER            | C.VERSION  | PROVIDER                   | P.VERSION | SUCCESS?
openbank-fx-service | 8d321a8... | openbank-aml-service       | ???       | ???
openbank-fx-service | 8d321a8... | openbank-sanctions-service | ???       | ???

There is no verified pact between the latest version of openbank-fx-service with tag main
(8d321a80...) and the version of openbank-aml-service currently in sandbox
```

Two details worth pinning down:

1. **The consumer version never moves.** `8d321a8` is identical in all five runs. It is not the consumer's own change — it is `fix(admin-ui): stop tagging every GlitchTip event with a release nobody ships (#3243)`, a **later, unrelated** commit. The FX fix (#3189, `53b769f`) is an ancestor of it, so the published consumer pact does contain the change. The consumer side was never the problem.

2. **Re-verifying does not converge.** After #3239 gave aml/sanctions broker-sourced verification, both published main versions and `pact-verification-reconcile.yml` stopped listing them. `verify-provider.yml` was then dispatched for both and both succeeded. The next fx deploy still failed with the identical message.

That is the loop:

- `verify-provider` publishes a verification against provider version = **current main SHA at build time**
- `record-deployment` records provider version = **the SHA that was deployed**
- `can-i-deploy` asks about the **deployed** version
- `main` moves between the two, so the deployed version is never one that has a verification

## Why it is not a lag

[...]

Issue #3223, opened 2026-08-02
record-deployment synthesises a contentless broker version, making can-i-deploy permanently unanswerable for its counterparts


JiRaska:
## Summary

`record-deployment-on-merge.yml` marks a service as deployed at a sha for which the broker holds
**no pacts at all** — synthesising the version with an idempotent `PUT` when it 404s. Every consumer
that then asks "is my pact verified against the version of X currently in sandbox" gets `???`
**forever**, because there is nothing at that version to verify and no verification run will ever
target it.

This is the mechanism behind the fleet-wide `can-i-deploy` failure (#3178 measured auto-deploy
failing 83%). `openbank-lending-service` has been undeployable all night and runs 0.11.5 while `main`
is 0.20.2.

Distinct from #1348 (phantom records — services recorded as deployed that never deployed) and #2549
(reconcile asking about an unpublished sha). Here the deployment is **real**; the recorded version is
real-but-**contentless**, which makes the gate permanently unanswerable rather than merely stale.

## Evidence

Read-only broker query (`kubectl port-forward -n pact-broker svc/pact-broker 9292:9292`):

```
GET /pacticipants/openbank-ledger-service/versions/a499cad1594c214f770e2a23a8ccef780d14d86b
version: a499cad1594c214f770e2a23a8ccef780d14d86b
tags: []
pb:pact-versions: []
```

Zero pacts. Not even tagged `main`. And that is the version the gate resolves as "ledger currently in
sandbox" — the pod really does run `openbank-ledger-service:sandbox-a499cad1`.

What is `a499cad1`?

```
a499cad15 chore(admin-ui): deploy sandbox-6bb6bdfc (#3181)
 openbank-infra/gitops/components/admin-ui/admin-ui.yaml | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)
```

A one-line admin-ui YAML change. It touches ledger not at all.

Sampling the deployed version of four services against the broker:

| service | deployed sha | pacts at that version |
|---|---|---|
| ledger-service | `a499cad1` | **0 — contentless** |
| aml-service | `95b652dd` | **0 — contentless** |
| sanctions-service | `95b652dd` | **0 — contentless** |
| fx-service | `99114189` | 2 |

3 of 4. And the failures match exactly — `openbank-fx-service` is blocked by:

```
There is no verified pact between the latest version of openbank-fx-service with tag main
[...]

PR #3859, opened 2026-08-06
fix(ci): derive can-i-deploy's self-clearing promise from the broker, not the verdict

JiRaska:
## What this fixes

`can-i-deploy` block classes end in a **promise**: PENDING_BUILD's *"the 3-hourly reconcile
re-drives it automatically"* and UNVERIFIED's *"normally clears within one reconcile tick"*.
Both were constants attached to a verdict. `classify-can-i-deploy-block.sh` makes no network
call by design, so it could not know whether waiting helps — and for a large share of today's
blocks it does not, structurally.

This PR does **not** change the identifier scheme (that is the owner's call, see below). It
makes the durable/transient distinction come from **evidence**.

## Proven: which workflow writes the contentless versions

`record-deployment-on-merge.yml`'s idempotent `PUT` — the leading hypothesis in #3223, which
was explicitly not proven. It is now, by correlating a specific broker version with the run
that created it, to the second:

| observation | value |
|---|---|
| broker version `openbank-transaction-service` @ `a5e5d32a…` | `createdAt 2026-07-24T08:05:19Z`, `tags: []`, `pb:pact-versions: []` |
| its `deployed-versions` record (still `currentlyDeployed: true`) | `createdAt 2026-07-24T08:05:21Z` |
| run `30077670109`, job `record-deployment`, branch `chore/gitops-codeploy-swift-transaction-1348` | `08:05:18.9  version not in broker yet — creating it via REST (idempotent PUT)` <br> `08:05:19.5  PUT version → HTTP 201` <br> `08:05:21.4  Recorded deployment of openbank-transaction-service version a5e5d32a… to sandbox` |

`HTTP 201` is the decisive part: the version was **created** by that PUT, not found. No other
workflow on `main` calls `record-deployment` or PUTs a pacticipant version.

Two corrections to how this has been described:

- The recorded sha is the **image-tag sha**, not the gitops merge commit. `a5e5d32a` reached
  transaction-service because a reconcile tick rebuilt it at `main`'s tip and #2057 pinned
  `openbank-transaction-service:sandbox-a5e5d32a`. So the sha names a commit that touched the
  service zero times, exactly as #3223 says, but via the reconcile rebuild rather than via the
  branch name.
- On the runs measured tonight the verdict is **PENDING_BUILD**, not UNVERIFIED. On run
  `31080511057` all six blocked money-path services had `PACT_VERSION_PRESENT=no`, so every one
  reported *"its main-push build has not finished … the reconcile re-drives it automatically"*.
  The false promise is therefore not confined to the UNVERIFIED branch, and a fix scoped to that
  branch would have missed every service in that run.

## Measured, read-only, against the live broker

All **seven** distinct counterpart versions named by run `31080511057`'s verdicts carry zero pacts:

```
contentless  openbank-account-service@f409d797,openbank-balance-service@e88849d6,
             openbank-fraud-service@c1f2e5a5,openbank-fx-service@6138c3a7,
             openbank-pid-service@fe48f5b6,openbank-swift-service@fe48f5b6,
             openbank-transaction-service@a5e5d32a
```

That is the whole blocked set, all six of them money-path, every one promised a reconcile that
cannot deliver.

## The change

- **New `.github/scripts/probe-blocking-counterparts.sh`.** Parses the `(counterpart, version)`
  pairs the verdict itself names and asks the broker one question each: does that version have
  any pacts? Reports `contentless | has-pacts | none | unknown`. Every failure degrades to
  `unknown`, never to `has-pacts` — degrading to `has-pacts` would restore the false
  reassurance on no evidence.
- **`classify-can-i-deploy-block.sh`** gains `UNVERIFIABLE`: durable, names the offending
  versions, states that waiting cannot help. It is tested **before** the `PACT_VERSION_PRESENT`
  split (so it beats PENDING_BUILD too) and **after** REGRESSION (an observed verification
  failure still outranks it). When the probe returns `unknown`, the old message is kept but
  marked unproven rather than repeated as fact. The classifier stays a pure function — the
  network call lives in the probe, where a stubbed self-test can drive every branch.
- **`derive-codeploy-set.py`**: `UNVERIFIABLE` is co-deploy-eligible. It is the one class backed
  by a measurement that the block is durable, and a co-deploy is precisely its remedy. Before
  this class these blocks arrived as UNVERIFIED (already eligible) or PENDING_BUILD (excluded
  and mislabelled), so this keeps the honest half of today's behaviour rather than widening it.
- Both unit tests extended; the probe's self-test registered as an **enforced** gate.

**No deploy behaviour changes.** A blocked service stays blocked under every class; only the
label and the message move.

## Validation

- `probe-blocking-counterparts.sh --self-test` — 10 cases, including both directions of
  contentless-dominates and every probe failure mode. Falsified by construction: each case was
  run against the stub before the aggregation was written the way it is.
- Both directions checked against the **live broker** (read-only, port-forward): the contentless
  version answers `contentless`, and `openbank-transaction-service`'s current `main` version
  answers `has-pacts`. A probe that has only seen one answer proves nothing.
- The shell change was exercised as the **bytes extracted from `auto-deploy.yml`** (parsed out
  of the YAML with PyYAML, not retyped): real verdict + live broker → `UNVERIFIABLE`; probe
  script removed → the step survives under `-e` and falls back to today's label with the
  "unproven" qualifier.
- **Workflow validated against GitHub**, oracle validated first: a deliberately unparseable
  `auto-deploy.yml` pushed to a throwaway branch produced **1 run** (named after the file path,
  `failure`); this PR's version at the same head produced **0 runs**. Throwaway branch deleted.
- `run-gates.py --group supplychain`: 19/20 pass; the one failure is `release-scope-mismatch`,
  which refuses to run locally without `PR_DIFF_BASE`.
- `actionlint` finding set on `auto-deploy.yml` is **identical** to `origin/main`'s;
  `shellcheck` clean on the new script; largest `run:` step 18955 chars (limit 19000).
- After merging `main`: `.github/gates/gates.yaml` went 109 → 110 ids, diff is exactly the one
  added entry — checked because a clean merge can silently drop a neighbouring list entry.

## What I did NOT do, and what could still be wrong

- **I did not change what a pacticipant version is.** The owner's own comment on #3223 argues
  for content-addressed versions and states it is an ADR-level call, not unilateral. I did not
  write that ADR either: the proposal already exists in the owner's words on #3223, and a
  second copy authored by me would be me choosing its shape. This PR is complementary — after
  a content-id cutover the probe still answers correctly, and until then it stops the estate
  lying about it.
- **This does not unblock a single deploy.** Zero services deploy that did not before. It
  converts a false "wait" into an accurate "waiting cannot help, here is what has to move".
  If that reads as insufficient, that judgement is correct — the unblocking fix is the
  identifier decision.
- **Parsing risk.** The probe keys on the CLI's `"and the version of <svc> currently in <env>
  (<40-hex>)"` phrasing. A pact-CLI upgrade that rewords it silently yields `none` → today's
  behaviour, not a wrong answer, but also no fix. I did not pin or test against another CLI
  version.
- **`has-pacts` is not proof that waiting helps.** It only rules out this cause. #3306's other
  half — verifications publishing against a sha that is never the deployed one — still blocks
  permanently and will still be labelled UNVERIFIED/PENDING_BUILD here. I did not attempt it.
- **Cost.** Up to three broker calls per counterpart per blocked service, inside the existing
  `can-i-deploy` job. Small, but not measured under a fleet-sized block.
- **Not observed in CI.** The broker has no public ingress and `PACT_BROKER_URL` is blank off
  main-push, so the real probe path cannot run on this PR. Everything above is a self-test, a
  local run against the live broker, or an extraction of the real step — the first production
  evidence will be the first `main` push after merge.
- **The `unknown` fallback is silent about which it was.** A missing script and an unreachable
  broker both land on `unknown`; only the log line distinguishes them.

Refs #3223, #3306, #3596, #3432, #2549, #3454

### DIFF
--- a/.github/gates/gates.yaml
+++ b/.github/gates/gates.yaml
@@ -687,6 +687,23 @@ gates:
     run: |
       bash .github/scripts/probe-pact-version.sh --self-test
 
+  # The counterpart-verifiability probe (#3223). The classifier above says WHY a block
+  # happened; two of its classes end in a promise ("clears within one reconcile tick",
+  # "the reconcile re-drives it automatically") that was a hardcoded constant, not an
+  # observation — and on run 31080511057 all six blocked money-path services carried it
+  # while every counterpart version named was unverifiable by construction. This probe is
+  # what turns that promise into evidence, so its failure directions are what must be
+  # exercised: a contentless counterpart dominating a healthy one, and every probe failure
+  # degrading to `unknown` rather than to the reassuring answer.
+  - id: blocking-counterpart-probe-unit-test
+    name: "can-i-deploy counterpart verifiability probe (issue #3223, enforced)"
+    group: supplychain
+    mode: enforced
+    selftest: |
+      bash .github/scripts/probe-blocking-counterparts.sh --self-test
+    run: |
+      bash .github/scripts/probe-blocking-counterparts.sh --self-test
+
   # Workflow `run:` step size (#3082). GitHub rejects an oversized run script by making the
   # WHOLE workflow unparseable — zero jobs, no error message, every contributor's runs of that
   # file dead until someone reverts. It happened to auto-deploy.yml (#3135 -> #3139) and NOTHING
--- a/.github/scripts/classify-can-i-deploy-block.sh
+++ b/.github/scripts/classify-can-i-deploy-block.sh
@@ -44,6 +44,14 @@
 #     resolve-can-i-deploy-selector.sh. Only (no + workflow_dispatch) is treated specially;
 #     see the NOT_ASKED note below.
 #
+#   COUNTERPART_STATE / COUNTERPART_DETAIL — the caller's probe of the counterpart versions
+#     this verdict names, from probe-blocking-counterparts.sh (#3223):
+#       contentless → at least one of them carries NO pacts, so nothing can ever verify it
+#       has-pacts   → all of them have pacts; a missing verification really is the gap
+#       none        → the verdict names no counterpart version
+#       unknown     → the probe failed; neither durability nor transience is established
+#     Unset behaves as `none`, so a caller that does not probe keeps today's labelling.
+#
 #   PACT_VERSION_PRESENT — the caller's probe of
 #     GET <broker>/pacticipants/<svc>/versions/<sha>:
 #       no      → 404, this commit's build has published nothing yet
@@ -56,6 +64,8 @@
 #   UNVERIFIED    — a version exists but no verification result yet. Usually the
 #                   counterpart provider lagging by minutes; treated as transient, but
 #                   named separately so it is not mistaken for a passed verification.
+#   UNVERIFIABLE  — the block names a counterpart version that carries NO pacts, so no
+#                   verification can ever target it. Durable; waiting cannot help (#3223).
 #   NOT_ASKED     — the gate was never asked (#3454). resolve-can-i-deploy-selector.sh
 #                   REFUSEd to pick a selector, so no can-i-deploy verdict exists at all.
 #                   See the NOT_ASKED note below for why this is not PENDING_BUILD.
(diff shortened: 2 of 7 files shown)


JiRaska:
---

**Update (force-push to a single commit).** The first push failed the required `Gitleaks` check on
rule `curl-auth-user` — it matches the *shape* `-u "$A:$B"` and cannot tell a variable reference
from a literal, so a broker call with no secret in it fails the gate. The pair is now assembled
into one variable first. A fixup commit does not clear it (gitleaks scans the push range), so the
branch was squashed to one signed commit. Local `gitleaks detect` on the branch now reports only
the pre-existing `generic-api-key` in `openbank-libs/governance/agents.yaml`, which this PR does
not touch.

JiRaska:

`probe-blocking-counterparts.sh`'s self-test stubs `curl` rather than the script's own aggregation logic, so it genuinely exercises the parsing/aggregation code path (the thing an earlier repo lesson calls out — a retyped-copy self-test can't see the real logic). Coverage includes the cases that matter for correctness, not just "does it run": `contentless` dominates a mixed contentless+healthy verdict and a mixed contentless+unknown verdict (one unverifiable counterpart is enough — correctly not diluted by a healthy sibling); a short/non-40-hex sha is correctly not read as a version to probe; every probe failure mode (unreachable, non-JSON body) degrades to `unknown`, never to the reassuring `has-pacts`.

`classify-can-i-deploy-block.sh`'s new UNVERIFIABLE branch is tested for priority ordering against the two things that matter: it beats both self-clearing classes (PENDING_BUILD, UNVERIFIED) when the probe found a contentless counterpart, but a REGRESSION (an actual observed verification failure) still outranks it — `test-classify-can-i-deploy-block.sh` asserts this explicitly, correctly refusing to let a bookkeeping explanation mask a real broken contract. The "unproven" qualifier is also tested: when `COUNTERPART_STATE=unknown`, the self-clearing promise text is kept but marked unproven rather than either silently repeated (false confidence) or dropped (losing information).

`derive-codeploy-set.py` adding `UNVERIFIABLE` to `CODEPLOY_CLASSES` is justified with a concrete "before" state (these blocks previously arrived mislabelled as PENDING_BUILD, which was excluded) and tested (`test-derive-codeploy-set.sh` case 11b) to confirm the allow-list didn't lose eligibility when the label changed.

The `auto-deploy.yml` wiring correctly treats the probe as non-fatal — `read -r cp_st cp_dt < <(...) || { cp_st=unknown; cp_dt=""; }` degrades to `unknown` rather than aborting the step if the process substitution somehow doesn't produce full output, consistent with the probe's own documented "always exits 0" contract.

Verified: priority ordering (REGRESSION > UNVERIFIABLE > self-clearing classes) is tested; probe failure modes all degrade to `unknown`, never to a false-positive `has-pacts`; the new gate entry in `gates.yaml` is registered with both `selftest` and `run` pointing at `--self-test`, correctly declared (not an inline conditional-job step, which is the class of gate-registration bug this repo has hit before). Did not verify against a live pact broker that the `_links["pb:pact-versions"]` JQ path matches the broker's actual current API shape.

Issue #3348, opened 2026-08-02
k6 perf baseline as a CI gate that goes red on p95 regression


JiRaska:
Part of #3343 (operational maturity tracker, level 1 → 2).

## What

Make the k6 performance baseline a CI gate: `perf-baseline.yml` must go red on a p95 regression against the recorded baseline.

## Why

The perf reports directory has one file, dated 2026-07-10 — a one-off report, not a trend. A baseline that cannot fail the build is documentation, not a gate; regressions ship silently.

## Acceptance criteria

- `perf-baseline.yml` compares the run's p95 (and agreed secondary metrics) against a committed baseline and fails the workflow beyond the threshold.
- The baseline is versioned in the repo and updated only by an explicit, reviewed PR.
- k6 runs on a defined cadence (per relevant merge to money-path services, or nightly full-fleet) so a trend exists.
- A regression links the offending change range in the failure output.

JiRaska:
Status update: ADR-0243 + product-catalog pilot are in PR #3716. The branch now includes the workflow, k6 smoke script, rules.yaml advisory registration, actionlint/shellcheck cleanup, and local YAML validation. Auto-merge is blocked by the repo guard because the PR touches .github/workflows; it needs human review/merge. After merge, the next step is one manual perf-gate run and then expanding beyond product-catalog.

JiRaska:
2026-08-27 evidence update:

- Latest scheduled money-path baseline run 32823683692 completed green, but its Run k6 baseline step was skipped. It measured no ledger or transaction handler.
- Repository Variables do not contain PERF_LEDGER_URL or PERF_TXN_URL; repository Secrets do not contain PERF_READ_TOKEN.
- PR #7434 supplies the missing workflow/code contract: token only from a GitHub secret, bearer propagation to k6, and explicit not-run evidence when target or credential is absent. It does not claim that the external target/identity is configured.

The issue remains open until an isolated least-privilege target exists, authenticated baselines produce trend evidence, and the stated reviewed regression policy is implemented. This is a real operational gap, not a green workflow.

JiRaska:
## #7434 merged — and the first scheduled run after it still measured nothing

Follow-up to the 2026-08-27 evidence update, which cited PR #7434 as supplying the missing workflow/code contract. It merged that same day (2026-08-27T19:18:22Z). The next scheduled money-path baseline was 2026-09-01, and reading its **step** conclusions rather than the job conclusion:

```
[...]

PR #3871, opened 2026-08-06
ci(perf-gate): escalate a k6 threshold breach to an issue instead of a silent red run

### LINKED ISSUE #3768: perf-gate: threshold breach creates no escalation issue (ADR-0243 follow-up)
## What

ADR-0243 (decision 2) and the `perf-gate.yml` header both state that a threshold
breach during the advisory phase **creates a GitHub issue rather than blocking a
merge**. The merged workflow has no such step: on a k6 threshold breach the job
simply goes red.

A weekly scheduled workflow whose red run is addressed to nobody is a failure mode
this repo has already measured: `dependency-submission.yml` died of
`Java heap space` for three days with every run red while its consumers quietly
read stale data, because nothing escalated. The documented rule is that anything
whose output others depend on needs an escalation path that raises or refreshes an
issue (`fleet-attestation.yml` and `dependency-submission.yml` now do this).

## Acceptance criteria

- A `perf-gate.yml` run failing on the weekly schedule opens a GitHub issue (or
  comments on / reopens an existing open one — no duplicates), naming the service,
  the breached threshold and linking the run.
- A subsequent green run does not auto-close without a note; the issue is the
  triage record.
- Implementation follows the existing raise/refresh pattern used by
  `fleet-attestation.yml` (see its issue-upsert step), not a hand-rolled new one.

## Context

- ADR-0243 (`docs/adr/0243-k6-performance-gate-as-ci-advisory-check.md`), decision 2
- PR #3716 (pilot implementation), issue #3348 (parent tracker), #3343 (operational maturity)
- `openbank-product-catalog` is the only v1 scope; the escalation step must be in
  place before money-path services join the matrix (they raise the stakes of a
  silently red run).

Refs #3348

JiRaska:
## What

ADR-0243 decision 2 and `perf-gate.yml`'s own header both promise that an advisory-phase
k6 threshold breach **creates a GitHub issue rather than blocking a merge**. The merged
workflow had no such step: a breach only turned the weekly scheduled run red.

A red scheduled workflow on `main` is addressed to nobody. This repo has already paid for
that once — `dependency-submission.yml` was red for three days with `Java heap space`
while its consumers quietly read a stale graph (#1449).

Adds a `raise-issue` job in the **same raise-or-refresh shape** `fleet-attestation.yml`
and `dependency-submission.yml` already carry, rather than inventing a second mechanism
that would drift from the first.

- Fires on `failure() && github.event_name == 'schedule'` only — a `workflow_dispatch`
  failure is already in front of whoever dispatched it, and a repo issue for it is noise.
- Recovers the run's own `perf-reports-<service>` artifacts via `gh run download`
  (no new third-party action to pin — the repo uses `download-artifact` nowhere).
- Renders a body naming **each service, the specific thresholds it crossed, and the run
  URL**; comments on an existing open issue instead of filing a duplicate.
- **Never auto-closes.** A later green run adds nothing and does not close it — the issue
  is the triage record, per the acceptance criteria.
- Also written to `$GITHUB_STEP_SUMMARY`, so the run itself carries the diagnosis.

Body rendering lives in `.github/scripts/perf-gate-issue-body.py` so the reasoning sits in
a script header, not a `run:` block — prose there is charged against the 19000-char ceiling
that silently unparses a whole workflow. The step script is 664 chars.

## How the escalation was proven to fire

An escalation path that has never fired is unfalsified, so each layer was driven, and read
for what it **prints**:

1. **k6's threshold encoding was measured, not assumed.** `metrics.<m>.thresholds` maps an
   expression to a bool where **`true` = crossed**. Proven with one real k6 run carrying a
   passing and a failing threshold on the *same* metric, so the mapping cannot be an artifact
   of which way the run went:
   `http_req_duration {"p(95)<0.0001": true, "p(95)<60000": false}` against console
   `✗ 'p(95)<0.0001'` / `✓ 'p(95)<60000'`. A second control (all requests refused, so the
   metric had no samples) reported `false` for an impossibly tight bound — consistent.
2. **The renderer was fed that real summary JSON** (not a fixture) and named exactly the one
   crossed threshold, excluding the two that passed.
3. **The step script was EXTRACTED from the YAML** (parsed out of `jobs.raise-issue.steps`,
   not retyped) and run with a `gh` stub — itself validated against a known-positive first so
   it could not be a silent passthrough. Both branches were exercised:
   - no existing issue → `gh issue create --title ... --body-file ... --label tech-debt --label severity:medium`
   - existing issue → `gh issue comment 4242 --body-file ...` — **no duplicate**.
4. **The self-test** (`--self-test`, 16 assertions) covers the cases the renderer must
   distinguish, including the ones that must NOT read as a regression: no summary at all
   ("nothing was measured"), k6 green but job red elsewhere, corrupt JSON, and an
   unrecognised verdict encoding (which reports `unparseable`, never `no-breach`).
5. **The workflow was validated against GitHub, not just a YAML parser.** The oracle was
   calibrated on a known-bad file first — a deliberately broken workflow pushed to a throwaway
   branch produced a run titled after the file *path* with zero jobs. The real change then
   produced **zero runs** on its head SHA (valid file, no matching trigger). Throwaway branch
   deleted. `actionlint` clean (oracle also validated against a known-bad).

## What I did NOT verify — please weigh this

- **The job has never run on real GitHub infrastructure.** It cannot be: it is gated on
  `schedule`, and forcing it would have filed a real issue in a public repo. So
  `gh run download` on a run whose matrix partially failed, the `actions: read` permission
  being sufficient, and `gh issue list --search` actually matching this title are all
  **unproven in situ**. The first genuine weekly failure is the first real execution.
- **The `gh issue list --search "<title> in:title"` idiom is inherited** from
  `fleet-attestation.yml`/`dependency-submission.yml` verbatim. I did not independently
  confirm GitHub's search indexes a new issue fast enough that two failures in quick
  succession cannot both create one. Inherited risk, not introduced — but not eliminated.
- **k6 version skew.** The polarity was measured on k6 **v2.1.0** (what I had locally);
  `perf-gate.yml` pins **v0.54.0**. The legacy `--summary-export` shape is stable across that
  range, but I could not run 0.54.0. Mitigation: the parser accepts both the bare-bool and a
  `{"ok": bool}` encoding and reports `unparseable` rather than guessing — so a shape surprise
  degrades to "read the log yourself", not to a false green or a false breach.
- **`perf-gate.yml` itself has never had a successful real run** as far as I can see, so the
  artifact layout (`perf-reports-<service>/<service>-summary.json`) is read off the upload step
  in the workflow, not off an observed artifact. If the upload flattens differently, the body
  degrades to "no per-service reports were recovered" — safe, but less useful.
- No gate added, so nothing prevents this escalation from being removed later.
- `gates.yaml` deliberately untouched.

## Merge state

A PR touching `.github/workflows/` requires human review here. That is the correct final
state — no auto-merge was attempted and no override flag used.

Closes #3768
Refs #3348

### DIFF
--- a/.github/scripts/perf-gate-issue-body.py
+++ b/.github/scripts/perf-gate-issue-body.py
@@ -1,0 +1,336 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: Apache-2.0
+"""Render the escalation issue body for a failed `perf-gate.yml` run (ADR-0243 D2).
+
+WHY THIS EXISTS
+---------------
+ADR-0243 decision 2 and `perf-gate.yml`'s own header both say a threshold breach
+during the advisory phase "creates issues rather than blocking merges". The merged
+workflow had no such step: a breach just turned the weekly scheduled run red, and a
+red push/schedule-triggered workflow on `main` is addressed to nobody. This repo has
+already paid for that failure mode once — `dependency-submission.yml` died of
+`Java heap space` for three days, red every run, while its consumers quietly read a
+stale graph (issue #1449). The established remedy is a raise-or-refresh issue job,
+as `fleet-attestation.yml` and `dependency-submission.yml` both carry; this script is
+only the body renderer for perf-gate's copy of that same shape.
+
+WHY A SCRIPT AND NOT AN INLINE `run:` BLOCK
+-------------------------------------------
+Two reasons, both load-bearing in this repo:
+  * Prose in a `run:` block is charged against a hard ceiling. An oversized `run:`
+    step makes the WHOLE workflow file unparseable, with no error from GitHub — every
+    push then yields a run with zero jobs, killing the workflow for every contributor
+    (measured boundary: 20054 chars accepted, 20654 rejected; `gates.yaml` enforces
+    19000 via `check-workflow-run-step-size.py`). Reasoning belongs in a script header.
+  * A `run:` block can only ever be syntax-checked by its own CI. This file has a
+    `--self-test` so the failure path is exercised before it is trusted.
+
+READING k6's `--summary-export` JSON
+------------------------------------
+`metrics.<name>.thresholds` maps a threshold expression to a BOOLEAN, and the
+polarity is the one that reads backwards: **true means the threshold was CROSSED
+(breached, `✗` in k6's own console output); false means it passed (`✓`)**. That is
+not inferred from documentation — it was measured against a real k6 run in which a
+single metric carried one passing and one failing threshold simultaneously:
+
+    http_req_duration {"p(95)<0.0001": true, "p(95)<60000": false}
+    ...console:  ✗ 'p(95)<0.0001' p(95)=1.84ms
+                 ✓ 'p(95)<60000'  p(95)=1.84ms
+
+Both polarities in one observation, so the mapping cannot be a coincidence of which
+way that particular run went. A second control (all requests refused, so
+`http_req_duration` had no samples) showed the same metric reporting `false` for an
+impossibly tight bound — consistent, since a metric with no samples crosses nothing.
+
+The measurement was taken on k6 v2.1.0; `perf-gate.yml` pins v0.54.0. The legacy
+summary-export shape is stable across that range, but rather than bet on it, the
+parser accepts BOTH encodings it could plausibly meet — a bare bool, and a
+`{"ok": bool}` object (where `ok` is the inverse: ok=false means breached) — and
+degrades to "could not classify" rather than guessing. A misread here would report a
+green run as a breach, or worse, a breach as green, and the issue body is the only
+artifact a human reads.
+
+WHAT "FAILED" CAN MEAN
+----------------------
+A red perf-gate job is not necessarily a threshold breach. The k6 step also fails when
+postgres never becomes ready, when the service fails to boot in dev mode, or when the
+k6 binary install fails — in which case there is no summary JSON at all and the run
+measured NOTHING. Reporting that as a performance regression would send the reader
+hunting for a slowdown that never happened, so those cases are classified separately
+and named as such. A service whose summary exists with zero breached thresholds is
+reported too: it means k6 passed and the job died elsewhere.
+
+NON-GOALS
+---------
+This never closes an issue. Acceptance criterion for #3768 is explicit that a later
+green run must not auto-close the record without a note — the issue IS the triage
+trail, and a bot that silently closes it deletes the only evidence anyone was ever
+told. It also never opens a second issue: the title is deliberately constant and
+service-independent, because a per-service title would file a fresh issue each time
+the matrix grows, and an escalation that opens an issue every run is worse than
+silence.
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import pathlib
+import sys
+
+TITLE = "Performance gate: the weekly k6 run is failing (ADR-0243 advisory phase)"
+
+
+def _breached(value: object) -> bool | None:
+    """True = breached, False = passed, None = shape not understood.
+
+    See the module header: a bare bool is k6's summary-export encoding and its
+    polarity is "true == crossed". The dict form is accepted defensively and its
+    polarity is inverted (`ok: false` means the threshold failed).
+    """
+    if isinstance(value, bool):
+        return value
+    if isinstance(value, dict) and isinstance(value.get("ok"), bool):
+        return not value["ok"]
+    return None
+
+
+def classify(report_dir: pathlib.Path, service: str) -> dict:
+    """Return this service's outcome from whatever files its artifact contains."""
+    summary = report_dir / f"{service}-summary.json"
+    boot_log = report_dir / f"{service}-boot.log"
+
+    if not summary.is_file():
+        reason = (
+            "k6 never produced a summary — the run did not reach the load test "
+            "(postgres, service boot, or the k6 install failed). **Nothing was measured.**"
+        )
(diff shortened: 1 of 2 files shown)


JiRaska:

Checked for the failure-path-never-run class of defect this repo has been bitten by repeatedly. `perf-gate-issue-body.py --self-test` exercises real fixtures rather than a happy-path-only case: a mixed-polarity threshold breach (verified against a real k6 run's console output, per the module docstring — good, that's an actual falsification against observed k6 behaviour rather than an assumption about the JSON schema), a not-measured case (missing summary, boot log present), a k6-green-but-job-red case, an unrecognised threshold-verdict encoding, and corrupt JSON — and asserts each is classified into the *correct distinct bucket*, not just "handled without crashing." The one polarity bug this kind of code is prone to (k6's `thresholds` map reads `true` = breached, which is the "reads backwards" direction) is explicitly called out and tested both ways.

Correctly reasons about why a script (not an inline `run:` block) is used, citing the 19000-char workflow-parseability ceiling this batch's own task list flagged as a trap.

Minor, non-blocking observations:
- `gh issue list --search "$TITLE in:title"` — the title contains `(` and `:` characters; GitHub's search endpoint generally tokenizes on these and should still match, but if it ever silently fails to find the existing issue the fallback is a *second* open issue with the same title rather than a hard failure, which would be quietly wrong rather than loud. Not fatal since the else-branch is `gh issue create`, just no negative-search test exists for this specific query.
- `raise-issue` runs on `needs: perf` with `if: failure() && github.event_name == 'schedule'` — correctly scoped to schedule only (workflow_dispatch failures are excluded, matching the stated rationale that a manual run's failure is already visible to the dispatcher).

Verified: self-test fixtures are genuine synthetic k6 artifacts, not retyped assertions; the "never auto-closes" claim is enforced by the workflow only ever calling `gh issue comment` on an existing match or `gh issue create`, with no close path anywhere in the diff. Did not verify GitHub's issue-search tokenization behaviour against the literal title string in a live repo.

JiRaska:
Confirmed the shape claim rather than the description: `raise-issue`/`raise-or-refresh` exists in both `fleet-attestation.yml` and `dependency-submission.yml` on `main`, and this PR's `raise-issue` job follows the same search-then-comment-or-create pattern (`gh issue list --search "$TITLE in:title"`, comment on an existing open issue, create otherwise). Not a new mechanism.

The framing is right: a red scheduled workflow on `main` is addressed to nobody, and this repo has already measured that cost once (`dependency-submission.yml` red for three days on `Java heap space` while consumers read a stale graph, #1449). A weekly perf gate with no escalation is the same failure mode with a longer fuse.

Not merging — workflow change, human review by rule regardless of the diff.

JiRaska:
Reviewed. **No blocking findings** — this is the raise-or-refresh shape the repo already uses, with a genuinely falsified parser. One gap worth closing.

**Checked against the traps this repo has been bitten by, and it avoids all of them:**

- The `if:` is reachable: the workflow declares `on: schedule: cron "0 6 * * 3"`, so `failure() && github.event_name == 'schedule'` is not structurally dead.
- **Raise-or-refresh, not raise-per-run**: constant title, `gh issue list --search "$TITLE in:title"` → comment on a hit, create otherwise. I checked the empty-result branch specifically — `gh`'s `--jq` prints *empty* on no match, not the literal `null` a bare `jq` would, so there is no `gh issue comment null` path.
- **The `workflow_run` attempts trap does not apply**: `raise-issue` is a `needs:` sibling in the same workflow, not a reactor. It never calls `/actions/runs/*/jobs` at all — it recovers evidence via `gh run download "$GITHUB_RUN_ID"`.
- **No log grepping**, so the echoed-`run:`-script trap is out of scope: classification reads structured JSON from `<service>-summary.json`.
- `--body-file` throughout, so the backtick/`()` shell-substitution trap is avoided.
- **The verdict polarity was measured, not assumed** — and that is the one thing that would have silently inverted the whole feature. The docstring records a real run with a passing *and* a failing threshold on the same metric, so it cannot be an artifact of run direction, and an unknown encoding reports `unparseable` rather than `no-breach`. `--self-test`: 16/16, exit 0, including "a passing threshold on a breached metric is NOT reported".

**The gap: a failure *before* the matrix never escalates.** Verified by parsing the workflow —

```
prepare      needs=None   if=None
perf         needs=prepare
raise-issue  needs=perf   if=failure() && github.event_name == 'schedule'
```

If `prepare` fails, or emits an empty matrix (each service can hit one of the four `::warning::perf-gate: … skipped` paths), `perf` is skipped, so `raise-issue` is skipped too — `failure()` is false for skipped upstreams — and the run reads green having measured nothing. The renderer already has a "No per-service reports were recovered" branch for exactly this, and it is unreachable in the empty-matrix case. That is the same "a red addressed to nobody" problem this PR exists to fix, one job earlier. Cheap fix:

```yaml
needs: [prepare, perf]
if: ${{ !cancelled() && github.event_name == 'schedule' && (failure() || needs.prepare.outputs.matrix == '[]') }}
```

**Pre-existing, not yours, but worth knowing:** the thresholds themselves are inherited rather than measured — `product-catalog-smoke.js` uses `p(95)<2000` and its own comment says it mirrors the public synthetic test. A real regression can sit well under 2 s. Untouched by this PR, but the escalation makes a loose bound more consequential, since it now files an issue only for breaches that bound can see.

Also worth noting one negative I turned up while checking the title-search concern raised in-thread: verbatim searches matched real titles containing parentheses and colons, but a title containing `—` and `§` returned nothing. This PR's title has neither — it is a constraint on ever editing it, not a defect.

PR #3876, opened 2026-08-06
fix(lending): make origination state transitions atomic

### LINKED ISSUE #3850: lending origination: advance() decides on a stale read and writes without a predicate — no optimistic locking anywhere in the service
## Summary

`openbank-lending-service` has **no optimistic locking anywhere**, and its origination state machine
decides on a value read in one transaction and writes it back in another. Two concurrent callers can
therefore both advance the same loan application.

This is the same defect class as the compliance-pack `decide()` race in #3467 / #3837, which the
work there explicitly left out of scope:

> I did not survey the other four-eyes paths in this service (`LendingService.decide()` for
> collateral, the origination `lending.approve` path). They look structurally identical and may
> carry the same defect — I measured neither, so I claim nothing.

Measured; they do.

## Evidence

`LendingService.advance` reads, decides in memory on `existing.status`, and writes:

```kotlin
override fun advance(id: LoanApplicationId, actor: String): Uni<LoanApplication> =
    applications.findById(id).flatMap { existing ->
        ...
        val next = OriginationAdvance.nextState(existing.status, mandatoryStepsFor(existing))
        ...
        applications.update(existing.copy(status = result.newState))
```

`findById` and `update` are **separate transactions** — `update` carries its own `@WithTransaction`
— so the decision is taken on data that is already stale by the time it is written. The write has no
predicate on the expected current status:

```kotlin
@WithTransaction override fun update(application: LoanApplication): Uni<LoanApplication> =
    sf.withTransaction { s ->
        s.find(LoanApplicationEntity::class.java, application.id.value).flatMap { e ->
            e!!.status = application.status      // blind overwrite
            ...
```

And there is no optimistic-lock column to catch it:

```
@Version in openbank-lending-service :  none
@Version elsewhere in the fleet      :  openbank-account-service (AccountEntities, AccountRepositoryImpl,
                                        AccountReconciliationRepository)
```

The probe was validated against that known positive first — account-service proves the search works,
so the empty lending result is an absence rather than a broken grep.

Four call sites share the shape: `LendingService.kt:269`, `:305`, `:335`, `:346`.

## Why it matters

The origination flow is the money path's approval gate. A lost update here does not merely
mis-attribute a decision — two concurrent `advance` calls both read the same state, both compute
`nextState` from it, and both write, so a loan application can pass through a state without the
transition that state exists to require. `decidedBy` / `decisionReason` / `decidedAt` are overwritten
by whichever write lands last, so the audit trail records one actor for a decision two actors made.

## Suggested fix

The same remedy #3837 applies to `CompliancePackActivationService`: re-test the expected state inside
the UPDATE and treat a row count of `0` as a refusal, reusing the existing 400 path. A `@Version`
column would also work and matches account-service, but it is a schema change and a wider blast
radius.

## Not verified

- **No concurrency test was run against these paths.** The claim is derived from the transaction
  boundaries and the absence of a lock, not from an observed interleaving. #3837's own IT needed
  12 rounds to see the race ~10 times, so a reproduction here is worth building before the fix
  rather than after.
- I did not check whether the REST layer serialises `advance` per application by some other means
  (a distributed lock, an idempotency key, a single-writer queue). If one exists this is theoretical.
- `TerminationService` and `OriginationDecisionService` were not read; they appear in the same
  four-eyes grep and may share the shape.

Refs #3467, #3837

JiRaska:
## Scope

Issue #3850 says `LendingService.advance` decides on a stale read and writes without a predicate,
and that `openbank-lending-service` has no optimistic locking anywhere. I re-verified every claim
rather than inheriting it, and built the reproduction the issue explicitly did not have.

This is the same defect class as **#3467 / #3837** (`CompliancePackActivationService.decide()`), and
it uses the same remedy. **No file overlap with #3837** — that PR touches
`CompliancePackActivationRepository/Service`, their two tests and
`docs/threat-models/openbank-lending-service.md`; this one touches `LendingService`, the loan-
application repository port and impl, their two unit tests and one new IT. Verified against #3837's
current file list. The threat model is deliberately untouched: both threat-model gates
(`threat-model-coverage`, `threat-model-updated-on-trust-boundary-change --enforce`) pass on this
branch without it, and editing it would have collided with #3837.

## What I verified from the issue, and what was wrong

| claim | verdict |
|---|---|
| `findById` is `@WithSession`, `update` carries its own `@WithTransaction` | **confirmed** |
| `update` blind-sets status via `s.find(...)` + field assignment | **confirmed** |
| `@Version` appears nowhere in `openbank-lending-service` | **confirmed** — `grep -c` is 0; the probe finds it in account-service, balance-service and document-service, so the empty result is an absence, not a broken grep |
| the four call sites are `LendingService.kt:269, :305, :335, :346` | **confirmed**, exactly — `advance`, `expireIfInState`, `advanceIfInState`, `decide` |
| the REST layer might serialise `advance` some other way | **checked, it does not** — no idempotency key, no distributed lock, no queue; `advanceApplication` is a bare `POST` straight into the use case |
| `TerminationService` may share the shape | **confirmed** — `withLoan` is `loans.findById` then `loans.update`, and `LoanRepositoryImpl.update` is the same blind field assignment. **Not fixed here**, see below |
| `OriginationDecisionService` may share the shape | **WRONG, and worth saying so.** It is a pure evaluator: it holds no repository, never reads and never writes. It cannot carry this defect |

Two things the issue did not mention that I found while measuring:

- **`disburse` is the worst instance of the family and is now fixed too.** `bookLoan` saves the loan
  and its schedule, blind-writes `DISBURSED`, and only then posts to the ledger — so two concurrent
  disbursements of one `READY_TO_DISBURSE` application both posted cash. The claim now runs before
  the posting, so the loser pays nothing. It does **not** make disbursement atomic: the loan row is
  still written before the claim, so a refused racer leaves an unreferenced loan behind. That is
  pre-existing (today *both* racers book one) and needs its own change.
- **`update` never persisted the decision-engine fields at all.** It writes only status, `decidedBy`,
  `decisionReason`, `decidedAt`, so `decisionOutcome`, price band, reason codes and the input hash
  computed on the ASSESSMENT leg were dropped, and the re-read blanked them out of the response.
  Also pre-existing, also not fixed here — but it is why `claimTransition` returns the claimed
  application from memory rather than re-reading. The visible consequence is that the ASSESSMENT
  advance response now carries the values the engine computed instead of nulls.

## Proven RED first — verbatim

`OriginationConcurrentAdvanceIT` was written and run against unpatched `origin/main` (`617011b4b`),
**and re-run in its final form against unpatched `main` after the assertions were revised**, so the
red below is the red of the test that ships. Two real `POST /applications/{id}/advance` calls race
over real HTTP on a `CyclicBarrier`, 12 rounds, a fresh application each round, no artificial window
widening.

```
java.lang.AssertionError: [origination steps applied more than once by concurrent advance calls]
Expecting empty but was: ["round 0: 2 records of SUBMITTED -> KYC_PENDING (statuses=[200, 200])",
    "round 0: transitions claimed more than once: {SUBMITTED->KYC_PENDING=2}",
    "round 1: 2 records of SUBMITTED -> KYC_PENDING (statuses=[200, 200])",
    "round 1: transitions claimed more than once: {SUBMITTED->KYC_PENDING=2}",
    ...
    "round 11: 2 records of SUBMITTED -> KYC_PENDING (statuses=[200, 200])",
    "round 11: transitions claimed more than once: {SUBMITTED->KYC_PENDING=2}"]
```

**12 of 12 rounds recorded the same origination step twice.** Every round, both racers.

### The assertion I removed, and why that matters

The first version of the test asserted "exactly one request returns 200". Against the **fixed** code
it went red once in 12 rounds — and inspecting it showed the round was correct: the second racer's
read landed *after* the first commit, so it read `KYC_PENDING` and took the next edge. Two 200s, two
different steps, no lost update. That assertion was a false red waiting to fire, so it is gone. What
the test asserts instead is the invariant that actually separates the defect from honest
interleaving: **no single transition is recorded twice**, read out of the transactional outbox that
`transitionEvidence` writes. On broken code that fires 12/12; on correct code it cannot fire for a
timing reason.

## The fix

`compareAndSetStatus` re-tests the expected state inside the UPDATE:

```
update LoanApplicationEntity
   set status = :to, decidedBy = :decidedBy, decisionReason = :reason, decidedAt = :decidedAt
 where id = :id and status = :from
```

The database evaluates the predicate and the assignment in one statement, so exactly one of any
number of concurrent callers claims the row. `claimTransition` in `LendingService` treats `0` as a
refusal, throwing the same `IllegalStateException` the caller would have got a moment later — which
the resource already maps: **422 on advance** (already declared in `openapi.yaml`), **409 on decide**
and on the other paths. Nothing downstream of a refusal runs: no evidence event, no workflow signal,
no ledger posting. Named parameters because all three decision fields are nullable and a positional
vararg cannot carry a null.

**Why not `@Version`:** it works and matches account-service, but it is a schema change on a live
table shared by many read paths, and optimistic locking surfaces as an `OptimisticLockException`
that then has to be mapped back to a sensible status. The predicate *is* the invariant, costs no
migration, and reuses the refusal path that already exists. Same call #3837 made.

**Why not `SELECT ... FOR UPDATE`:** it would hold a transaction open across the decision engine on
the ASSESSMENT leg, which is the expensive part of that call.

## No schema, no API, no version change

- **No Flyway migration** — the predicate uses the existing `status` column, so the `migrate-at-start`
  trap does not apply.
- **No `openapi.yaml` change** — 422 is already the declared refusal for advance, 409 for the rest.
  So no `info.version` bump and no race with a competing spec PR.
- `version.txt` untouched; release-please owns it. `fix` + `src/main` releases.

## Verification

| | result |
|---|---|
| `OriginationConcurrentAdvanceIT` on unpatched `main` | **FAILED** — 12/12 rounds double-recorded (above) |
| same test, with the fix | **PASSED**, three independent runs |
| `:openbank-lending-service:test` (211 tests) | PASSED |
| `ktlintCheck`, `detekt` | PASSED |
| `koverVerify` | PASSED |
| `:openbank-lending-service:quarkusBuild` | PASSED (CDI wiring — unit tests do not check it) |
| `check-threat-models.py`, `check-threat-model-diff.py --enforce` | PASSED |

Nine existing unit tests moved from stubbing `applications.update` to stubbing the claim. One new
unit test — `losing the advance race refuses and emits no transition evidence` — pins the half the IT
can only observe indirectly: the service computes the transition *before* it writes, so when the
UPDATE claims no row it is already holding a fully-advanced application and must not emit for a step
it did not take.

## What I did NOT verify, and what could still be wrong

1. **The IT is a timing test. It cannot produce a false RED — one origination step recorded twice is
   a defect however the threads landed — but it CAN produce a false GREEN.** It widens no window, so
   a round in which the two requests serialise passes vacuously. The measured rate against broken
   code was 12/12 per round on this machine; on a slower or busier runner it could be lower, and in
   principle a whole run could be green against broken code. That is the honest answer to "could
   your test pass against the old code": yes, with low probability, and the 12 rounds are what makes
   it unlikely rather than impossible.
2. **Only N=2 concurrent callers are measured.** The conditional UPDATE is correct for any N by
   construction; nothing here proves it for N>2.
3. **Only the `SUBMITTED -> KYC_PENDING` edge is raced.** `decide`, `expireIfInState` and `disburse`
   go through the identical `claimTransition` and are covered by unit tests, but no IT races them.
   `disburse` in particular deserves its own concurrency IT and does not have one.
4. **`TerminationService` is measured and NOT fixed.** `withLoan` + `loans.update` on `LoanEntity`
   carries the same read-check-write shape, including `proposeTermination`/`decideTermination`, which
   is another four-eyes control. Fixing it means a second conditional-update method on
   `LoanRepository` and its own reproduction; folding it into a money-path PR that already needs two
   approvals would have made this unreviewable. It needs its own issue and its own red test — I am
   claiming the shape, not a measured race.
5. **The disburse orphan-loan window is still open** (see Scope). The fix strictly reduces harm there
   — one ledger posting instead of two — but does not close it.
6. **Isolation level is the Testcontainers/deployed Postgres default (`READ COMMITTED`) only.** A
   single-statement conditional UPDATE is atomic there; no other level was exercised.
7. **Single JVM, not a multi-pod deployment.** That is the right scope — the defect is a database
   lost update and the fix is in the statement — but it is not a live multi-replica proof.
8. **The decision-engine persistence gap (see Scope) is named, not fixed**, and the response-shape
   change that falls out of `claimTransition` returning the in-memory application is a real, if
   small, behaviour change worth a reviewer's eye.

## Merge expectation

`openbank-lending-service` is money-path: 2 human approvals plus a threat model by rule. **This PR is
not expected to merge on my action, and that is the correct outcome.**

Closes #3850
Refs #3467, #3837

### DIFF
--- a/openbank-lending-service/src/main/kotlin/com/openbank/lending/application/port/out/LendingRepositories.kt
+++ b/openbank-lending-service/src/main/kotlin/com/openbank/lending/application/port/out/LendingRepositories.kt
@@ -14,7 +14,9 @@ import com.openbank.lending.domain.model.LoanStateSummary
 import com.openbank.libs.domain.identifiers.CollateralId
 import com.openbank.libs.domain.identifiers.LoanApplicationId
 import com.openbank.libs.domain.identifiers.LoanId
+import com.openbank.libs.lending.origination.OriginationState
 import io.smallrye.mutiny.Uni
+import java.time.OffsetDateTime
 import java.util.UUID
 
 interface LoanApplicationRepository {
@@ -31,7 +33,30 @@ interface LoanApplicationRepository {
      * pages to add up rows would be the same wrong answer, slower.
      */
     fun summariseByState(): Uni<List<ApplicationStateSummary>>
+
+    /**
+     * Blind write of the decision fields. Correct only where the caller is not deciding anything
+     * from the value it is overwriting — [compareAndSetStatus] is what an origination transition
+     * must use.
+     */
     fun update(application: LoanApplication): Uni<LoanApplication>
+
+    /**
+     * Move the application from [from] to [to] **only if** the stored row is still in [from], as a
+     * single statement. Returns the rows claimed: `1` when this caller won the transition, `0` when
+     * someone else already moved the row on.
+     *
+     * The caller must treat `0` as a refusal and must not perform any side effect of the transition
+     * — no evidence event, no workflow signal (issue #3850).
+     */
+    fun compareAndSetStatus(
+        id: LoanApplicationId,
+        from: OriginationState,
+        to: OriginationState,
+        decidedBy: String?,
+        decisionReason: String?,
+        decidedAt: OffsetDateTime?,
+    ): Uni<Int>
 }
 
 interface LoanRepository {
--- a/openbank-lending-service/src/main/kotlin/com/openbank/lending/application/usecase/LendingService.kt
+++ b/openbank-lending-service/src/main/kotlin/com/openbank/lending/application/usecase/LendingService.kt
@@ -132,7 +132,7 @@ class LendingService(
             is OriginationTransitionResult.Rejected ->
                 Uni.createFrom().failure(IllegalStateException(result.reason))
             is OriginationTransitionResult.Applied ->
-                applications.update(outcome.recorded.copy(status = result.newState))
+                claimTransition(existing.status, outcome.recorded.copy(status = result.newState))
                     .signalWorkflow()
                     .call { _ -> events.emit(outcome.evidence) }
         }
@@ -146,6 +146,44 @@ class LendingService(
     private fun reflectionDaysFor(application: LoanApplication): Int? =
         complianceGuard.resolveOriginationPack(application.jurisdiction, application.productType)
             ?.pack?.reflectionPeriodDays
+
+    /**
+     * Claim an origination transition, or refuse.
+     *
+     * Every caller here has decided [updated] from a snapshot read in an earlier transaction, so it
+     * must not write blindly: the row may have moved on since. The conditional UPDATE re-tests the
+     * state the decision was taken against, and `0` rows claimed means someone else got there first
+     * — the same `IllegalStateException` the caller would have received a moment later, which the
+     * resource already maps (422 on advance, 409 on decide). Nothing downstream of a refusal runs:
+     * no evidence event, no workflow signal (issue #3850).
+     *
+     * On success the claimed application is returned from memory. `update` returned the re-read
+     * entity, which differs in one respect worth naming: `update` writes only status and the three
+     * decision fields, so the ASSESSMENT leg's engine outputs (`decisionOutcome`, price band, reason
+     * codes, input hash) were never persisted and the re-read blanked them out of the response. That
+     * persistence gap is pre-existing and is NOT fixed here — it needs its own change and its own
+     * test. What changes is only that the response now carries the values the engine computed
+     * instead of nulls.
+     */
+    private fun claimTransition(from: OriginationState, updated: LoanApplication): Uni<LoanApplication> =
+        applications.compareAndSetStatus(
+            updated.id,
+            from,
+            updated.status,
+            updated.decidedBy,
+            updated.decisionReason,
+            updated.decidedAt,
+        ).flatMap { claimed ->
+            if (claimed == 0) {
+                Uni.createFrom().failure(
+                    IllegalStateException(
+                        "Concurrent modification: application ${updated.id} is no longer in $from",
+                    ),
+                )
+            } else {
+                Uni.createFrom().item(updated)
+            }
+        }
 
     private fun Uni<LoanApplication>.signalWorkflow(): Uni<LoanApplication> =
         call { app -> workflowPort.stateEntered(app.id, app.status, reflectionDaysFor(app)) }
@@ -287,7 +325,7 @@ class LendingService(
                             is OriginationTransitionResult.Rejected ->
                                 Uni.createFrom().failure(IllegalStateException(result.reason))
                             is OriginationTransitionResult.Applied ->
-                                applications.update(existing.copy(status = result.newState))
+                                claimTransition(existing.status, existing.copy(status = result.newState))
                                     .signalWorkflow()
                                     .emitEvidence(
                                         existing.status.name,
@@ -319,7 +357,7 @@ class LendingService(
                     is OriginationTransitionResult.Rejected ->
                         Uni.createFrom().failure(IllegalStateException(result.reason))
                     is OriginationTransitionResult.Applied ->
-                        applications.update(existing.copy(status = result.newState))
+                        claimTransition(existing.status, existing.copy(status = result.newState))
                             .signalWorkflow()
                             .emitEvidence(
                                 existing.status.name,
@@ -364,7 +402,8 @@ class LendingService(
                     when (val result = machine.apply(transition(existing, existing.status, target, decidedBy))) {
                         is OriginationTransitionResult.Rejected ->
                             Uni.createFrom().failure(IllegalStateException(result.reason))
-                        is OriginationTransitionResult.Applied -> applications.update(
+                        is OriginationTransitionResult.Applied -> claimTransition(
+                            existing.status,
                             existing.copy(
                                 status = result.newState,
                                 decidedBy = decidedBy,
@@ -468,7 +507,14 @@ class LendingService(
                     is OriginationTransitionResult.Rejected ->
                         Uni.createFrom().failure(IllegalStateException(st.reason))
                     is OriginationTransitionResult.Applied ->
-                        applications.update(application.copy(status = st.newState))
+                        // Claimed, not blind-written: without the predicate two concurrent
+                        // disbursements of one READY_TO_DISBURSE application both post cash to the
+                        // ledger. The claim runs before the posting, so the loser pays nothing.
+                        // It does NOT make disbursement atomic — the loan row and its schedule are
+                        // written before the claim, so a refused racer still leaves an unreferenced
+                        // loan behind. That is pre-existing (today BOTH racers book one) and needs
+                        // its own change; see the pull request for #3850.
+                        claimTransition(application.status, application.copy(status = st.newState))
                             .call { savedApp ->
                                 events.emit(
                                     transitionEvidence(
--- a/openbank-lending-service/src/main/kotlin/com/openbank/lending/infrastructure/persistence/repository/LendingRepositories.kt
+++ b/openbank-lending-service/src/main/kotlin/com/openbank/lending/infrastructure/persistence/repository/LendingRepositories.kt
@@ -110,6 +110,58 @@ class LoanApplicationRepositoryImpl @Inject constructor(
             s.persist(e).map { mapper.toDomain(e) }
         }
     }
+
+    /**
+     * The origination transitions do NOT go through [update].
+     *
+     * [update] is a blind write: it stores whatever the caller loaded, whenever the caller gets
+     * round to it. Every origination transition reads the row in one transaction ([findById],
+     * `@WithSession`), computes the next state from that snapshot in memory, and writes in another
+     * — so two `advance` calls arriving together both observe the same status, both compute the
+     * same forward edge, both pass the state machine, and both write. A lost update on the money
+     * path's approval gate: the status column ends up holding the right value, and the transition
+     * has been claimed twice, with two evidence records and two workflow signals for one step
+     * (issue #3850). On `decide()` the same window lets two checkers each cast what each believes
+     * is the deciding four-eyes vote, and the later write owns `decidedBy`/`decisionReason`.
+     *
+     * Making the expected state part of the UPDATE closes the window with no lock, no `@Version`
+     * column and no schema change: the database evaluates `status = :from` and the assignment in
+     * one statement, so exactly one of any number of concurrent callers claims the row and the rest
+     * get `0`. `OriginationConcurrentAdvanceIT` measures it; on the previous code it observed both
+     * advances accepted in 12 of 12 rounds.
+     *
+     * Same remedy, same reasoning as `JpaCompliancePackActivationRepository.compareAndSetDecision`.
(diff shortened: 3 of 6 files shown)


JiRaska:
**Review: 3 findings, none blocking, one behaviour change I'd want stated in the changelog and not only in the PR body.**

The mechanism is sound: `compareAndSetStatus` is one HQL statement with `where id = :id and status = :from`, `claimTransition` refuses on `0` before `signalWorkflow()` and before `events.emit(...)`, and all five former `applications.update(...)` call sites are converted. The IT's choice of invariant — "no single transition recorded twice", read out of the outbox — is the right one, and removing the "exactly one 200" assertion was correct for the reason stated (two 200s taking two different edges is honest serialisation, not a lost update).

1. **`LendingService.kt:150-185` — the ASSESSMENT advance response now returns engine outputs that are not in the database.** `claimTransition` returns the in-memory application, and `compareAndSetStatus` writes only `status`/`decidedBy`/`decisionReason`/`decidedAt` — so `decisionOutcome`, price band, reason codes and input hash come back in the POST response and are `null` on the very next GET. Before this PR the re-read returned nulls in both places: wrong, but *consistently* wrong. For a credit decision, an API that reports an outcome it did not store is arguably a worse failure than one that reports nothing, and admin-ui / any client caching the advance response now disagrees with the record. The PR body names the persistence gap honestly, but describes the response change as "small"; I'd call it the reviewable part of this diff. Either persist those columns in the same conditional UPDATE, or keep the re-read.

2. **`OriginationConcurrentAdvanceIT.submitApplication()` picks the application by `ORDER BY created_at DESC LIMIT 1` for a fixed `proposed_by`.** Rounds are sequential so it is usually right, but if two rounds land in the same `created_at` tick the ordering is unspecified and round N races round N-1's row. It fails loudly (`assertThat(statusOf(id)).isEqualTo(SUBMITTED)` catches it), so this is flakiness, not a false green — reading the id from the `POST /applications` response or a `RETURNING`/`Location` would remove it.

3. **`LoanApplicationRepository.update` now has zero production callers** — it survives as a port method with a docstring warning. Reasonable as a transition, but the next transition added to this service will find `update` sitting there looking like the normal way to write. Consider removing it (or narrowing it to whatever non-transition write still needs it) so the defect cannot be reintroduced by autocomplete.

Verified without finding a problem: `sf.withTransaction` + `@WithTransaction` matches the class's existing convention (`update`, `summariseByState`); nullable named parameters are exercised for real by the advance path in the IT, so the "Hibernate cannot infer a null parameter type" risk is measured, not assumed; the disburse leg claims **before** `ledger.post`, so the loser genuinely pays nothing (the residual orphan-loan row is correctly named as pre-existing and out of scope); the nine converted unit tests all moved from stubbing `update` to stubbing the claim and still assert the same outcomes; `TerminationService` really does carry the same shape and really is untouched, matching the body; no file overlap with #3837.

One thing unchanged and worth knowing: the claim and the outbox emit are still separate transactions, so a crash between them loses the evidence for a transition that did commit. Same as before this PR — noting it so it is not mistaken for something the conditional UPDATE fixed.

*Automated agent review — not an approval. `openbank-lending-service` is money-path: the 2-human-approval rule still applies.*

JiRaska:
Second review pass. No blocking defect — the CAS does what it claims and is tested at the right layer. **One thing I would change before merge**, in the file this PR is titled after.

**The KDoc asserts an atomicity the code does not provide.** `LendingService.kt:202`:

> "…lands in the transactional outbox **in the same commit as the state change**"

Two transaction boundaries, verified on the branch rather than inferred:

- `compareAndSetStatus` carries `@WithTransaction` (`LendingRepositories.kt:135`)
- `emit` → `outbox.persistInTransaction(...)` (`JpaLoanEventEmitter.kt:49`), whose own KDoc says the boundary lives on `LendingOutboxRepositoryImpl`'s `@WithTransaction` — a *second* transaction, reached via `.call { }` after the claim has already committed

Failure scenario: CAS commits `SUBMITTED → KYC_PENDING`, the pod dies before `emit`. The caller sees a 500, the row has already moved, and a retried `advance` re-reads the new status and takes the *next* edge — so the first transition is permanently absent from `credit.application.transition` evidence, with nothing detecting the hole.

This behaviour is **pre-existing and unchanged by the diff**; the finding is that the PR is titled "atomic" and the neighbouring comment says "same commit", which is what the next reader will take as settled. Either correct the wording or fold the outbox write into the CAS transaction — but the two should not disagree.

**Checked and clean:**

- *persist-vs-merge*: `LoanApplicationRepositoryImpl.update` does `find` then `persist`, but the entity is *managed* in that session, so persist is a no-op and the dirty check emits the UPDATE. The documented trap needs a detached entity; this is not it. `save` is genuinely INSERT-only, correct for `apply`.
- *Four-eyes*: `decide` rejects `decidedBy == existing.proposedBy` with both identities from the JWT subject, and the CAS predicate `id = :id and status = :from` binds the approval to the row — no cross-resource replay. The CAS additionally closes the "two checkers each cast the deciding vote" window, which is the point.
- *Test layer*: `OriginationConcurrentAdvanceIT` races two real `POST /advance` calls and reads the assertion out of `lending_outbox` over plain JDBC — the `LendingOutboxWriteIT` pattern. It also asserts `racedStep == 0` as a violation, so a "fix" that dropped the evidence entirely goes red. That is what makes the invariant non-vacuous.

Minor, pre-existing, outside the diff: `transitionEvidence` hand-builds JSON and sanitises only `"` in `reason`. A client-supplied `decision.reason` containing a backslash or newline writes malformed JSON into the outbox, and the IT's regex payload parser would not notice.

JiRaska:
**Scope note: `Closes #3850` will auto-close an issue this PR only partly fixes.** Suggest changing it to `Refs #3850`, or opening the follow-up before merge.

This PR covers the origination `advance()` path. Re-derived on `origin/main` just now, the same read-decide-write shape is untouched in `TerminationService`, across **seven** transitions:

```
:113 settle          :196 markDelinquent   :278 accelerate
:152 withdraw        :203 markDefaulted    :216/:227/:254 forbearance
```

all going through the same `withLoan { … } / loans.update(…)` helper — read in one transaction, decide, write in another, with no predicate on the write. `settle`, `markDefaulted` and `accelerate` are the ones that end a loan and move money, so the failure mode this PR demonstrates on `SUBMITTED→KYC_PENDING` applies there with a larger consequence.

Also worth recording, since it makes the convention question explicit rather than implicit:

| service | files with `@Version` |
|---|---|
| openbank-lending-service | **0** |
| openbank-account-service | 3 |
| openbank-balance-service | 1 |
| openbank-document-service | 5 |

So three services already carry the JPA optimistic-locking convention and lending carries none. This PR's predicated-update (`where id = :id and status = :from`) is a deliberate second mechanism, and that is defensible — it needs no migration and refuses through the existing 422/409 paths — but it should be a stated choice rather than a silent divergence, because the next person fixing `TerminationService` will have two precedents to pick from.

The probe behind that table was validated against the three services that DO use `@Version` before being trusted on the one that reports zero — a count of zero from an unvalidated grep is indistinguishable from a broken grep.

Nothing here blocks this PR. It stays unmergeable by an agent regardless: lending-service is money-path, `reviewDecision` is empty, and it needs two human approvals.

PR #3880, opened 2026-08-06
fix(libs): stamp ApiError.timestamp at construction instead of serving 1970

### LINKED ISSUE #3874: every ApiError in the fleet serves timestamp 1970-01-01T00:00:00Z — Instant.EPOCH default, 0 of 13 call sites set it
## Summary

Every error response the fleet serves carries `"timestamp":"1970-01-01T00:00:00Z"`. Found in the
first error body read while re-measuring #3039:

```json
{"traceId":"78166ae6-d1dc-4a35-a681-5caa08a364a1","status":500,"code":"INTERNAL_ERROR",
 "message":"An unexpected error occurred. Please contact support with traceId=...",
 "timestamp":"1970-01-01T00:00:00Z"}
```

## Cause

```kotlin
// openbank-libs-domain/src/main/kotlin/com/openbank/libs/api/error/ApiError.kt:16
val timestamp: Instant = Instant.EPOCH,
```

A default of `Instant.EPOCH` rather than a required argument, so any construction that omits it
serialises the epoch. Measured across the fleet:

```
files constructing ApiError()            : 13
sites passing `timestamp = ` explicitly  :  0
```

(The probe was validated against a known positive — `timestamp =` does appear in three other Kotlin
main sources — so the zero is a real absence, not a broken search.)

Every visible call site passes `traceId` / `status` / `code` / `message` and stops there:

```
account-service/…/ExceptionMappers.kt:23,42
aml-service/…/ExceptionMappers.kt:19,26
card-issuance-service/…/CardExceptionMappers.kt:28
consent-service/…/ExceptionMappers.kt:16
…
```

## Why it matters

The field is in the published `openapi.yaml` contract of every service, so an external integrator or
the mobile app receives a timestamp that is always the same wrong value. For a banking platform the
practical cost is on the support path this envelope is explicitly designed for: the message says
*"contact support with traceId=…"*, and the one field that would let support correlate that trace to
a moment in time is fixed at 1970. It also silently defeats any client-side or log-side ordering that
trusts it.

It is not a crash, which is exactly why it has survived — nothing fails, and the value is
syntactically valid.

## Suggested fix

Make it non-defaulted (`val timestamp: Instant`) so the compiler finds all 13 sites, or default it to
`Instant.now()`. The former is preferable: a default that is silently wrong is what produced this, and
a second default would only move the failure. `Instant.now()` in a data-class default is also
evaluated per construction, which is right here, but it makes "the caller forgot" indistinguishable
from "the caller meant now".

Note `TIMESTAMPTZ` truncates to microseconds while `java.time.Instant` carries nanoseconds — only
relevant if anything persists this value, which today nothing appears to.

## Not verified

- Whether any consumer actually reads the field. Its uselessness is argued from the contract and the
  support flow, not measured from a client.
- Whether any service overrides it at a layer I did not read (a response filter, an interceptor).
- Whether `openapi.yaml` marks the field required in every service's spec — a change here may or may
  not be an API-contract change under ADR-0048.

Refs #3039

JiRaska:
Closes #3874.

## What was wrong

Every error response the fleet served carried `"timestamp":"1970-01-01T00:00:00Z"`.
`ApiError.timestamp` defaulted to `Instant.EPOCH`, and **none** of the construction sites
passed it. Nothing ever failed on it — the value is syntactically valid and serialises
cleanly — which is why it survived. The cost lands on the support path the envelope itself
points at (*"contact support with traceId=…"*): the one field that would place a trace in
time was a constant.

## Verifying the issue's measurements

Re-measured against live `origin/main` (`d02b5d63`). Two corrections, both minor:

| Claim in #3874 | Measured | Verdict |
|---|---|---|
| `ApiError.timestamp = Instant.EPOCH` at `ApiError.kt:16` | exact | ✅ |
| 0 call sites pass `timestamp =` | 0 | ✅ |
| "files constructing `ApiError()`: 13" | **12** call-site files (the 13th match is `ApiError.kt`, the declaration itself). **28** construction sites across those 12 files. | ⚠️ off by one file; the substance is unchanged |
| "The field is in the published `openapi.yaml` contract of every service" | **2 of 27** specs that reference `ApiError` declare `timestamp` (aml, finrep). The other 25 declare only `{code, message}`-ish envelopes. | ❌ **wrong** |
| "nothing appears to persist this value" | confirmed — it only ever reaches `Response.entity(...)` | ✅ |

The spec correction cuts both ways: the contract exposure is narrower than stated, but 25
specs *under-declare* an envelope they do serve. That is a separate pre-existing gap, not
touched here.

**No API-contract change under ADR-0048.** The declared shape (`type: string`,
`format: date-time`) is unchanged; only the value changes. No `openapi.yaml` is modified,
so there is no `info.version` bump to take.

## The design decision, and why

`timestamp` is now **required**, not re-defaulted to `Instant.now()`. Three reasons:

1. **A default that is silently wrong is what produced this.** `Instant.now()` would make
   every existing call site correct without anyone looking at it, and leave "the caller
   meant now" indistinguishable from "the caller forgot" — the same failure mode, one
   value to the right. The next field added to this envelope faces the identical trap.
2. **The compiler is the enumerator.** Removing the default is what *found* all 28 sites.
   A default would have shipped a green build and told me nothing about whether any given
   site was even reachable.
3. **`libs-domain` is framework-free by design (ADR-0122).** `Instant.now()` in a
   data-class default is a hidden, unmockable clock read, evaluated per construction — it
   would make two `ApiError`s with identical content unequal, and this repo's tests compare
   these objects.

**What stops the next silently-wrong value?** Not the type system — someone can still
write `timestamp = Instant.EPOCH`. What changes is that doing so is now a visible,
deliberate act in a diff rather than an omission nobody can see. Backing that up, the two
tests added here assert **recency**, so a constant of any kind fails them.

The cost is honest: 28 call sites across 12 services, and the churn is the bulk of this diff.

## Proven with RED-first tests

Both assert **recency**, not non-nullness — an `isNotNull` assertion passes against
`Instant.EPOCH`, which is precisely how this survived.

**1. `CommonExceptionMappersTest` — all nine shared mappers.** RED against unmodified code:

```
Expecting all elements of:
  [ApiError(traceId=f9a3f48b-…, status=400, code=VALIDATION_ERROR, message=x, timestamp=1970-01-01T00:00:00Z, details=null),
   ApiError(traceId=e19f4ec7-…, status=422, code=BUSINESS_RULE_VIOLATION, …, timestamp=1970-01-01T00:00:00Z, …),
   … all 9 …]
```

**2. `TemplateWireFormatTest` — the SERVED body over real HTTP** (`@QuarkusTest` +
RestAssured, in finrep, one of the two services whose spec declares the field). This is the
half that proves the wire format, not just the mapper's return value. RED against
unmodified code:

```
Expecting actual:
  1970-01-01T00:00:00Z
to be between:
  [2026-08-06T09:45:47.664057Z, 2026-08-06T09:45:49.676582Z]
```

## Blast radius — please read this part

This changes a shared `openbank-libs-domain` type, so it touches **every service's error
path**. 12 services, **5 of them money-path** (account, consent, domestic-payment, sca,
sepa-payment) — so this needs 2 approvals per `rules.yaml`.

What limits the risk: the change is type-level and compiler-enforced, `./gradlew
compileKotlin` across the **whole fleet** passes (that is the proof no call site was
missed), and no behaviour other than the value of this one field changes.

## ADR-0106 side-effect, handled

Reformatting the one-liner call sites moved 12 existing `UUID.randomUUID()` trace-id
expressions onto newly-added lines, which trips `check-new-random-uuid.sh` (diff-scoped).
I ran that guard locally, saw it fail with all 12, and migrated those sites to
`Ids.randomId()` — the ADR-0106 "as-touched" path (#1699). Guard now passes. Sites in
`PartyResource.kt` whose lines were *not* re-added are deliberately left alone to keep the
diff minimal.

## Also found (not fixed here)

`openbank-libs-domain/.../audit/AuditEvent.kt:43` has the **identical** defect:
`val timestamp: Instant = Instant.EPOCH`, ~25 construction sites in `src/main`, ~2 passing
it, and `AuditEventPublisher.kt` contains no occurrence of the string `timestamp`, so
nothing stamps it centrally either. If that holds under closer inspection it is worse than
this bug, because it is the evidentiary record. `FlagExposure.kt:31` has the same shape.
Deliberately out of scope — this PR is already fleet-wide and money-path.

## Self-evaluation — what I did NOT verify

- **Could the tests pass against the old code?** No. Both were run against unmodified
  production code first and both failed with the epoch value shown above.
- **I did not observe a running pod.** The served-body proof is a `@QuarkusTest`, not a
  request against a deployed environment.
- **Only finrep's served body is asserted end-to-end.** The other 11 services are covered
  by the mapper-level test plus the type system, not by their own HTTP-level assertion.
- **Whole-fleet `compileTestKotlin` did not complete** — it OOM'd the Gradle daemon
  locally (environmental; no `e:` errors were emitted). Whole-fleet `compileKotlin`
  (main sources) *did* pass. Test sources are low-risk here because no test in the repo
  constructs `ApiError` — verified by grep — but CI is the real check on that.
- **Three test failures during the local gate were "Failed to start quarkus"**, the known
  port-contention flake from parallel Gradle builds; all three passed when re-run serially.
  I did not run them against a pristine `origin/main` worktree to double-confirm, because
  the failure mode was a boot error rather than an assertion.
- **The `Ids.randomId()` migration changes which UUID version those trace ids use** (v4
  either way here, so no behavioural change intended) — but I did not test trace-id
  generation itself; no test covers it.
- **Clock skew / test flakiness:** the recency windows are bounded by real `Instant.now()`
  reads (±1s in the HTTP test). A sufficiently slow CI runner could in principle exceed the
  window, though the assertion brackets only the request itself.


JiRaska:
Picked this up after the authoring session stopped mid-fixup and finished it.

**Where it had stopped:** `threat-model-updated-on-trust-boundary-change` had fired — correctly.
Making `ApiError.timestamp` a required argument changes the inbound REST surface of every service
that serves the envelope, so five money-path threat models needed the assessment recorded. The
session had written those five entries and stalled before committing them. Committed now.

**The assessment, since it is the substantive part and not boilerplate:** risk class is information
disclosure and the increase is deliberate and bounded. An error response now reveals the server's
wall-clock second to anyone who can provoke one, including an unauthenticated caller on an endpoint
answering 401/403 through this envelope. Acceptable, because the HTTP `Date` header on the same
response already carries second-resolution UTC — nothing is disclosed that the caller could not
already read — and it is what makes the envelope's own instruction (*"contact support with
traceId=…"*) actionable. Explicitly **not a timing oracle**: the stamp is taken when the error object
is built, not measured against request start, so no per-request duration is exposed.

**Design choice confirmed as the issue recommended** — `val timestamp: Instant`, non-defaulted, so
the compiler finds every call site and an epoch timestamp can now only appear because someone wrote
it. A default of `Instant.now()` would have made "the caller forgot" indistinguishable from "the
caller meant now", which is the shape that produced the original defect.

**Verification:**

```
threat-model-coverage                            PASS
threat-model-updated-on-trust-boundary-change    PASS
:openbank-libs-domain:test  :openbank-libs-runtime:test
:openbank-libs-runtime:ktlintCheck  :openbank-libs-runtime:detekt     (--rerun-tasks, all visibly ran)
BUILD SUCCESSFUL in 35s
```

Delta measured with a 3-dot diff against `origin/main` — **15 files, zero `openapi.yaml`**. The
response shape is unchanged (`string`/`date-time`), so this is not an API-contract bump under
ADR-0048. Both commits signed.

**Not verified:** I did not re-derive the original red test — the recency assertion and its failure
against the old code are the authoring session's evidence, not mine. I also did not run the per-service
suites for the eleven services whose `ExceptionMappers` changed, only the shared libs; the change is
mechanical (adding one argument at each site) and the compiler is the check that matters, but that is
an argument rather than a measurement.

JiRaska:
Merged `main` in (11 commits behind) and re-ran the gate. Clean merge, 0 conflicts, and both fixes
survive:

```
openbank-libs-domain/.../api/error/ApiError.kt:32   val timestamp: Instant          (this PR)
openbank-libs-domain/.../audit/AuditEvent.kt:49     val timestamp: Instant = Instant.now()   (#3882)
```

`--rerun-tasks` over `:openbank-libs-domain:test`, `:openbank-libs-runtime:test`, `ktlintCheck`,
`detekt` — all four visibly executed, BUILD SUCCESSFUL.

**A question for the reviewer that the merge exposed, and I would rather raise it than quietly leave
it.** #3882 fixed the same class of defect in the same file tree and chose the *opposite* remedy:

| type | remedy | consequence |
|---|---|---|
| `AuditEvent` (#3882) | `= Instant.now()` default | a caller that forgets gets "now", which is usually right and is indistinguishable from a caller that meant "now" |
| `ApiError` (this PR) | no default at all | the compiler names every call site; an epoch value can only appear because someone wrote it |

I still prefer the non-defaulted form here, for the reason this issue documents: a silently-wrong
default is exactly what produced `1970-01-01T00:00:00Z` across 13 call sites and survived for the
life of the type. But two shapes for one problem in one library is a legitimate thing to want
consistent, and which way to converge is a judgement call rather than a fact. Worth one look before
merge; happy to switch this to a `now()` default if the preference is uniformity.

**Note on why the earlier `all-green` was red:** it was not a test failure. Four service builds
(`aml`, `agent`, `anacredit`, `account`) reported **`cancelled`**, and `all-green` renders that as
"One or more service builds failed". Same pattern hit #3887 earlier today. The merge above produces a
fresh run, which is the cleaner fix than re-running the stale one.

JiRaska:
## Split feasibility analysis (2026-08-06)

Investigated whether this PR could split into a mergeable libs-only subset + gated per-service remainder. **Verdict: no — stronger than a review-scope issue, this is compiler-enforced.**

`ApiError.timestamp` loses its default entirely:
```kotlin
-    val timestamp: Instant = Instant.EPOCH,
+    val timestamp: Instant,
```
That's a required-constructor-argument change. Every direct `ApiError(...)` call site fleet-wide fails to compile the moment it lands — confirmed via `git grep -l "ApiError(" origin/main` across all `openbank-*/src/main`: exactly 11 non-libs files, and this PR patches all 11, no orphans. All 11 construct `ApiError` **directly** (not via the shared `CommonExceptionMappers` — that only covers the generic IllegalArgument/IllegalState/NoSuchElement/WebApplicationException catch-alls; each service's domain-specific exceptions like `AccountNotFoundException`/`ConsentNotFoundException` build their own `ApiError` inline). Spot-checked directly against the diff: `openbank-account-service/.../ExceptionMappers.kt` adds `timestamp = Instant.now()` at each of its two construction sites — exactly this pattern.

So merging just `openbank-libs-domain/ApiError.kt` + `openbank-libs-runtime/CommonExceptionMappers.kt` would immediately break the build of all 11 services, 9 of them money-path (account/aml/card-issuance/consent/domestic-payment/kyc/party/pid/sca/sepa-payment). There's no deferring the per-service half — it's not a design choice, it's what makes the fleet compile.

**Also worth flagging, orthogonal to the split question:** the diff updates threat-model docs for only 5 of the 10 code-touched services (account, consent, domestic-payment, sca, sepa-payment) — aml, card-issuance, kyc, party, pid have no corresponding threat-model entry here.

**Current state independent of the gate:** `mergeStateStatus: BLOCKED`, CI has 5 `FAILURE`s (kyc-service, party-service, sca-service, security-scanner, statement-service) — all failing at the "Set up job" step before any Gradle task runs, including on two services (security-scanner, statement-service) that have no `ApiError(` call site at all. Reads like a runner/infra flake affecting the whole run rather than a defect in this PR's code, but it means the PR isn't currently mergeable as a whole regardless of the gate question.

No merge or split action taken — stays as-is, still correctly gated.

JiRaska:
Second review pass, focused on blast radius — this is `openbank-libs`, so it reaches every service. **No contract break, and the test shape is right.** One low-severity finding.

**Wire shape is unchanged.** `timestamp` was already declared and already serialized (as `1970-01-01T00:00:00Z`); only the value moves. Of the 27 specs, aml and finrep are the only two declaring it, both `type: string, format: date-time` and **not** required — so nanosecond precision stays spec-conformant. Correctly no `openapi.yaml` and no ADR-0048 bump.

**Nothing pins the old value.** No file under `pacts/` references it, so zero Pact exposure. `openbank-admin-ui/src/types/index.ts:58` declares `ApiError { traceId, status, code, message }` with no `timestamp`, and TS structural typing ignores the extra field. Nothing in Kotlin deserialises `ApiError` — no consumer DTO, no `readValue<ApiError>` — so making the argument required cannot break an inbound path.

**Call-site completeness, verified independently of the body:** `git grep -l 'ApiError(' origin/main -- '*/src/main/*'` returns **13**, and the PR patches all 13. **0** test sources construct it, so `compileTestKotlin` risk is genuinely nil. I also compared the file set at the merge-base against current `origin/main` — identical, so despite being ~175 commits behind, no new call site has appeared that this branch would fail to compile against.

**The `Instant.EPOCH` trap is avoided, and this is the part worth calling out.** Both new tests assert **recency**, not non-nullity:

- `CommonExceptionMappersTest` brackets nine mapper invocations between two real `Instant.now()` reads and asserts `isBetween(before, after)` — all nine shared mappers, not a sample.
- `TemplateWireFormatTest` (finrep) parses the **served HTTP body** and does the same with ±1s.

That second one is what makes the suite non-vacuous: it proves the wire format, not the mapper's return value. The one `isNotNull()` present is a `describedAs` guard *preceding* the recency check, not a substitute for it. Worth stating explicitly because `isNotNull()` passes against 1970 — the exact way this class of bug survived review before.

**Kotlin semantics are right**: `Instant.now()` is written at each call site inside a function body, so it evaluates per invocation. Removing the default rather than defaulting to `now()` is the correct call for a framework-free `libs-domain` (ADR-0122) — a `now()` default is an unmockable clock read that also breaks data-class equality in tests. No `ApiError.copy(` anywhere in the fleet, so the copy-doesn't-re-evaluate-defaults trap is moot.

**The one finding**, low severity, comment-only: `openbank-party-service/.../rest/ExceptionMappers.kt:132-133`. A blanket `UUID.randomUUID()` → `Ids.randomId()` substitution rewrote the comment that explains the exception, inverting it —

> `// The pre-existing mappers above still mint via bare Ids.randomId(); left untouched…`

`Ids.randomId()` *is* the ADR-0106 path, not "bare", and the mappers above were migrated in this same diff, so nothing is left untouched. The ADR-0106 guard strips comments by design, so nothing will ever flag this. One line to delete.

JiRaska:
Unblocked: merged `origin/main` in (`875812701`, signed). The two conflicts were both **append-only threat-model log entries**, not competing edits:

- `openbank-account-service.md` — this branch's 2026-08-06 `ApiError.timestamp` entry against main's 2026-08-03 delegation-projection entry
- `openbank-consent-service.md` — this branch's 2026-08-06 entry against main's 2026-08-07 engagement-service caller entry

Resolved by keeping **both**, in chronological order at the tail. "Ours vs theirs" are not alternatives here — each side is a distinct dated record someone wrote, and picking one would have deleted a threat-model entry with nothing to notice. The conflict was the only thing that surfaced them at all.

Verified by counting entries rather than eyeballing the diff, since this repo has had a clean merge silently drop a neighbouring list entry:

| file | branch | main | merged |
|---|---|---|---|
| account-service | 17 | 20 | **21** |
| consent-service | 4 | 7 | **8** |

Each is main's count plus this branch's one new entry, and both of main's contested entries are present by name. Zero Kotlin conflicts — the 15 code files auto-merged untouched.

Nothing was rebased or force-pushed: `origin/main` was merged **in**, so the owning worktree at `/private/tmp/ob-3874-3506` is simply behind and loses nothing. It was checked before touching this branch — last commit 2026-08-06T16:05, clean tree, nothing written in 30 minutes, and its local head matched both the remote head and the PR head exactly.

For the record on #3874's substance, independently confirmed while working this: the field **is** read — `ApiError` is serialised to the client on every error, so the 1970 value reached every API consumer of every service. That makes it corruption rather than a latent trap, which is what distinguishes it from the other `Instant.EPOCH` in libs-domain. And only 2 of 27 `openapi.yaml` specs declare `timestamp` at all, none marking it `required`, which is both why no ADR-0048 contract bump is needed and why it survived this long.

JiRaska:
**Two more sites, same class — and that is worth naming rather than just fixing.**

`ExceptionMappers.kt:64` (`PaymentNotCompletedMapper`) and `:79` (`DocumentTemplateUnavailableMapper`) landed on `main` after this branch was cut and did not pass `timestamp`. Fixed, matching the file's existing shape.

This is now the **third** batch (account/party/etc. originally, then domestic-payment, now sepa-payment). It is structural, not carelessness: removing the default makes the compiler enumerate every construction site, so **every merge to `main` that adds an `ApiError(` breaks this PR**. The window in which it compiles is bounded by the next unrelated merge that touches an exception mapper.

That is also exactly the argument for the change. Each of these six sites would otherwise have compiled, passed review, and served `1970-01-01T00:00:00Z` to every caller who provoked that status — silently, because `isNotNull()` passes against 1970 and no test asserts recency. The PR is not creating work; it is surfacing work that was already being created invisibly.

**Practical consequence: this should merge soon or it will need a fourth pass.** If it cannot merge promptly, the cheaper shape is to keep the default and add a gate that fails on a construction site omitting `timestamp` — a ratchet instead of a compile-time enumeration — because a ratchet does not go stale when `main` moves.

JiRaska:
`all-green` failing across three re-runs, three different heads, three unrelated services — worth explaining rather than re-running again.

| attempt | head | failed service | cause |
|---|---|---|---|
| 1 | `54573e2` | agent-service, anacredit-service, aml-service, account-service | `cancelled` — concurrency-group cancel |
| 2 | `61506126` | sepa-payment, simulation | not inspected — head moved before I could |
| 3 | `c515b8a` | notification-service | `MarketingGateIT` fails on the just-landed `ContactPolicyGate` wiring (#4224) — unrelated to this PR's `ApiError.timestamp` change |

None of the three failures touch this PR's scope (`openbank-libs-domain` `ApiError`, plus the 12 consuming services' exception mappers). This is a `libs-domain` change, so every push triggers a full fleet rebuild, and the branch is moving faster than that rebuild completes — each attempt catches a different, unrelated service mid-flux on `main`.

Re-running again is unlikely to help while the branch keeps moving. Worth rebasing onto a settled `main` once nothing else is landing in `openbank-notification-service`, rather than chasing a fresh flake each time.

Issue #3881, opened 2026-08-06
copilot erasure consumer cannot receive PARTY_ERASED: no KafkaUser, no Kafka config — and the partyId/sub key may not match once wired


JiRaska:
## Summary

#3875 merged the GDPR erasure path for copilot conversation history — `PartyErasureConsumer`,
`deleteForCustomer`/`deleteConversation`/`deleteExpired`, and a daily retention sweep. The code is
there and tested. **In the deployed environment it cannot receive a single event**, so the control
exists and is not in force.

## Measured, read-only

```
KafkaUsers in the cluster                    : 38
  named copilot                              :  0
copilot-service (platform ns, running 52d)
  env vars containing KAFKA                  : NONE
  kafka/cert volume mounts                   : NONE
gitops components/copilot/                   : no KafkaUser, no msg-override ConfigMap,
                                               no Kafka manifest of any kind
```

For comparison, services that do consume have one: `messaging/account-service` and
`messaging/notification-service` are both `tls` / `simple` / Ready.

The cluster runs `authorization: simple` with `allow.everyone.if.no.acl.found=false`, so a workload
with no KafkaUser and no certificate has no path to the broker at all. The consumer will simply never
be invoked. Nothing errors — which is the point.

## The second half: the key may not match even once wired

The event carries `partyId`; copilot keys history on the OIDC `sub`. The customers realm defines a
separate `party_id` claim backed by a user attribute, and customer-edge falls back to `sub` only when
that attribute is unset. Where the two differ, the consumer deletes **zero rows and logs success**.

That is the more dangerous half, because after wiring it will look like it is working: a subscribed
consumer, a handled event, a clean log, and nothing deleted. The erasure would be provably wrong only
by querying the table.

This was disclosed in #3875's PR body and in `openbank-copilot-service/CLAUDE.md`. Filing it so it
does not live only there — the same reason #3870 existed at all.

## What needs doing

1. A KafkaUser + cert-reading ExternalSecret + Kafka config for copilot, matching the shape
   notification-service uses. Note `check-kafka-cert-reader-grant.py` derives its required set from
   the ExternalSecrets that read `kafka-messaging-certs`, so both sides must move together.
2. Resolve `partyId` → the key history is actually stored under, or store `partyId` alongside `sub`
   on the write path. Until then the consumer's success is not evidence of erasure.
3. A test that asserts rows are **gone by direct query** after the real event — reading through the
[...]

Issue #3870, opened 2026-08-06
copilot conversation history is a durable personal-data store with no deletion path — no PARTY_ERASED consumer, no expiry sweep


JiRaska:
## Summary

#3711 merged, so `openbank-copilot-service` now holds **customer chat transcripts in Postgres with no
deletion path**, and #3710 — the issue that would have carried the follow-up — is closed. The gap is
recorded in exactly one place: a comment inside the migration.

```sql
-- openbank-copilot-service/src/main/resources/db/migration/V1__conversation_history.sql:10
-- no @Scheduled sweep and no PARTY_ERASED consumer in this service yet; both are required by ...
```

That is the "a fix recorded only in a PR description dies with that PR" shape, one level down. Nothing
outside that SQL comment now says this is outstanding.

## What is actually true on `main`

**The port has no delete operation**, so honouring an erasure needs a signature change, not just a
subscriber:

```kotlin
fun load(customerId: String, conversationId: String): List<ChatMessage>
fun append(customerId: String, conversationId: String, newTurns: List<ChatMessage>)
```

**`expires_at` is a read-side filter, not deletion.** Rows stop being *served* after the 90-day
rolling TTL and stay in the table indefinitely. For free-text chat transcripts — which is precisely
where unpredictable personal data lands — the difference between "not served" and "not held" is the
whole question.

**This service is the outlier, not the pioneer.** `PARTY_ERASED` is already consumed by seven
services across 19 Kotlin files:

```
account-service   aml-service        card-issuance-service   kyc-service
notification-service   onboarding-service   party-service
```

So this is not a control the repo has documented and never built. It is built, and a new durable
personal-data store was added outside it — which widens an existing control's blind spot rather than
deferring an unbuilt one. (Worth stating explicitly because a narrower grep first suggested to me
that *nothing* implemented erasure, which would have been the opposite conclusion.)

## What needs doing

1. A delete operation on `ConversationStore` — by customer, and by conversation.
2. A `PARTY_ERASED` consumer in copilot-service that calls it, matching the shape the other seven use.
3. A `@Scheduled` sweep that actually removes rows past `expires_at`, so the 90-day TTL is a retention
   guarantee rather than a read filter.

Trap for (3), since it has bitten five schedulers in this repo already: a plain (non-`suspend`)
`@Scheduled` method carries **no Vert.x context**, so `runBlocking { … }` around a reactive Panache
[...]

PR #3885, opened 2026-08-06
fix(copilot): give the PARTY_ERASED consumer a Kafka identity, a reachable broker and a reachable database

JiRaska:
## Summary

Wires copilot-service onto Kafka so the `PARTY_ERASED` consumer merged in #3875 can actually be
invoked — and fixes two further faults found while verifying it, either of which alone would have
kept the erasure path dead after the Kafka half landed.

**The plumbing is now real. The erasure is still not correct**, and the second half of #3881 is why.
That is measured below, not assumed.

Closes nothing on its own. Refs #3881, #3875, #3870, #3710.

## What was actually broken — three independent faults, none of which errors

I verified #3881's premise rather than inheriting it. It was right, and incomplete.

**1. No Kafka identity (as filed).** No KafkaUser, no certificate, no `KAFKA_*` env, no Kafka
manifest of any kind. The broker runs `authorization: simple` with
`allow.everyone.if.no.acl.found=false`, so a principal-less workload has no path to it and the
consumer is never invoked.

**2. No route to the broker even with an identity.** `networkpolicy-copilot-egress.yaml` is Egress
default-deny and had no rule for `messaging:9093`. A KafkaUser and a mounted keystore are equally
inert against a broker the pod cannot reach. (Broker *ingress* is fine — Strimzi's own
`openbank-cluster-network-policy-kafka` admits :9093 from anywhere; the generated
`kafka-broker-ingress-allow-list` only covers :9092, which is a separate matter and untouched here.)

**3. No route to its own database — so there was nothing to delete from.** The same allow-list has
no rule for `copilot-db:5432`, *and* the datasource URLs name `openbank_copilot` while
`postgres.yaml` creates `openbank_copilots`. `migrate-at-start: true` means Flyway fails the boot,
so **every pod carrying the #3710 conversation store has CrashLooped since the day it was added**;
the sandbox has been serving from a pre-#3710 ReplicaSet the whole time. Observed live: the newer
ReplicaSet's pod at 1/2 Ready with 20 restarts, logging
`Connection error: The connection attempt failed. (Caused by Connect timed out)` on a loop.

An erasure consumer cannot delete rows from a database its own pod has never opened.

## What this PR changes

- `kafka-copilot-mtls.yaml` — KafkaUser `copilot-service` (Read/Describe on `openbank.party.events`,
  Read on group `copilot-service-party`; consumer-only, no Write grant) + the keystore ExternalSecret.
  The cluster-CA truststore ExternalSecret is **deliberately not repeated**: `platform` already has
  one, created by agent-service, whose header says a second client here should share it. Two
  ExternalSecrets with the same `target.name` in one namespace fight over the same Secret.
- `kafka-scheme-accepted-acl.yaml` — `copilot-service` added to `eso-kafka-cert-reader`'s
  `resourceNames`, **in the same commit** as the ExternalSecret that requires it.
  `check-kafka-cert-reader-grant.py` derives its required set from those ExternalSecrets and fails
  both ways, so neither half can land alone. Gate goes 38/38 → 39/39 clean.
- `copilot-service-msg-override.yaml` — `group.id` / `auto.offset.reset` as an external properties
  source with `config_ordinal=500`. A YAML leaf key containing a dot never reaches the connector
  (#686), and env vars are not the workaround either: `party-events-in` is hyphenated, which trips
  the open upstream `SRMSG00071` bug. Same mechanism notification-service and transaction-service use.
- `copilot-service.yaml` — `KAFKA_*` env, keystore/truststore/msg-override mounts,
  `QUARKUS_CONFIG_LOCATIONS`, and the `openbank_copilots` database-name correction.
- `networkpolicy-copilot-egress.yaml` — egress to `messaging:9093` and to `copilot-db:5432`
  (selected by the CNPG-owned `cnpg.io/cluster` label, so the rule stays one dependency wide rather
  than opening the namespace as agent-service's `podSelector: {}` does).

`auto.offset.reset=earliest` — **safe here, and stated deliberately.** The hazard is forcing
`earliest` onto a group *already running* on the default, which rewinds it. `copilot-service-party`
is a brand-new group with no committed offsets, so there is no position to rewind: it starts at the
oldest retained record. `openbank.party.events` has `retention.ms=604800000`, so the bound is seven
days, and picking those up is the point — they are erasures copilot could not act on.

**No `.rego` and no `rules.yaml` key were touched, so no OPA bundle or pod-roll annotation is
restamped.** `gen-network-policies.py` was re-run and produced no diff. Enforced `gitops` gate group:
22 PASS / 2 pre-existing advisory warnings / 1 pre-existing unfalsified self-test, all unrelated to
these files.

## The partyId/sub question — MEASURED, and the answer is worse than filed

#3881 asked for a measurement of the deployed realm before any fix, and warned that a guess here is
worse than an obvious gap. I measured it read-only against the live sandbox, via the Keycloak and
party-service databases. No credential was handled in plaintext and nothing was mutated. Aggregates
only below — the underlying identifiers are personal data and stay out of this repo.

In realm `openbank-customers`, **35 users**:

| | count |
|---|---|
| `party_id` attribute set, and it equals the Keycloak user id (`sub`) | **0** |
| `party_id` attribute set, and it differs from `sub` | **9** |
| no `party_id` attribute at all | **26** |
| …of those 26, whose `sub` is a real party id in party-service | **0** |

So for **0 of 35 users** does the identity copilot stores (`CopilotChatResource.customerSubject()`
returns the JWT `sub`, confirmed) equal the identity the event carries. The ADR-0069 invariant
`sub == partyId` holds for **nobody** in the deployed realm. `deleteForCustomer(partyId)` will match
zero rows for every user, and log a clean success while doing it.

The join was validated against a known-positive before I believed the zero: injecting one real party
id into the same query returns exactly 1 match, so this is a true zero and not a broken cast.

Root cause is visible in the code and consistent with the data: `WebAuthnKeycloakClient.ensureUser`
creates the Keycloak user **without** setting `id`, so Keycloak generates a random UUID and
`party_id` is stored as a separate attribute. Only `seed-test-customer.sh` sets `id = partyId`, and
nothing in the live realm was created that way. `WebAuthnResource` already documents the divergence
in a comment.

**I did not fix this, on purpose.** The candidate fix — key conversation history on the `party_id`
claim with `sub` as fallback, the same resolution customer-edge already uses — changes which
conversation a customer resumes and orphans every existing row, and it is a decision about identity
in a money-path-adjacent service. That belongs in its own PR with a human on it, not smuggled into a
plumbing change. #3881 stays open for it. The measurement is posted there.

One thing I measured and cannot explain, reported as an observation and not a diagnosis: none of the
9 `party_id` attribute values corresponds to a row in party-service's `parties` table either (69
rows, spanning the whole period). So keying on `party_id` alone may not close the gap on its own.

## What still has to happen after merge

Merging this does not put erasure in force. In order:

1. ArgoCD syncs `copilot` and `payments`. The KafkaUser and ExternalSecrets are at sync-waves -2/-1,
   so ArgoCD blocks on them reaching `SecretSynced` before the Deployment rolls — if the
   `eso-kafka-cert-reader` grant were missing this would strand at wave -1 exactly as #2749 did.
2. The Strimzi User Operator mints the `copilot-service` Secret; ESO projects it into `platform`.
3. The Deployment rolls. **This is the first time a pod carrying #3710/#3875 will boot successfully**
   — the netpol and database-name fixes are what make that possible, so expect the conversation
   schema to be created by Flyway on that boot, not before.
4. Only then does the consumer join `copilot-service-party` and start receiving events.

No Vault value is required. Verify with: the KafkaUser Ready, the ExternalSecret `SecretSynced`, the
pod 2/2 with no Flyway retry loop, and the consumer group present on the broker with a committed
offset — the last of these is the only one that proves it is consuming.

## Critical self-evaluation — what I did NOT verify

- **I did not observe a single event flow end to end.** Nothing here is proven against a real
  `PARTY_ERASED`; this is a manifest change verified by gates, code reading and live read-only
  inspection of the cluster's current state. The first real proof is step 4 above.
- **No test asserts rows are gone by direct query.** #3881's item 3 is not addressed. `ConversationErasureIT`
  exists but I did not extend it, and reading through the store cannot distinguish deleted from
  filtered by `expires_at`. Given the key mismatch, such a test written today against the real key
  would fail — which is arguably the right time to write it, but it is not in this PR.
- **I did not verify the `user.password` / `ca.password` Secret keys exist under those exact names**
  for this KafkaUser; I copied the shape from notification-service and agent-service, which work. If
  Strimzi's key naming differed the pod would fail at boot loudly, not silently.
- **I did not check whether any `conversation_history` rows exist** — I could not, because the
  database has never been reachable. The table may not exist yet.
- **The 26 attribute-less users**: I proved their `sub` is not a party id. I did not establish what
  their `sub` *is*, or whether they are reachable customers at all.
- **The `openbank_copilot` → `openbank_copilots` rename assumes the CNPG bootstrap name is the
  intended one.** It is the only database on the cluster, so the alternative — renaming the database
  — would be the larger and more surprising change, but I did not confirm the original intent.

**What breaks if the key mismatch is real: it is real, and what breaks is the erasure itself.** After
this PR the consumer subscribes, receives the event, runs the delete, deletes nothing, and logs
`erased 0 copilot conversation(s)` at INFO. That is the exact failure mode #3881 called the more
dangerous half — a control that looks live in every dashboard and log while conversation transcripts
of an erased party stay on disk and in every base backup. **Do not read a green deploy of this PR as
GDPR Art. 17 coverage for copilot.** It is the prerequisite, not the control.

### DIFF
--- a/openbank-infra/gitops/components/copilot/copilot-service-msg-override.yaml
+++ b/openbank-infra/gitops/components/copilot/copilot-service-msg-override.yaml
@@ -1,0 +1,50 @@
+# Dotted-key YAML config bug workaround for copilot-service (issue #686, issue #3881).
+#
+# Quarkus's YAML config source unconditionally quotes any `mp.messaging.incoming.<channel>`
+# YAML leaf key containing a literal dot when registering it as a MicroProfile Config
+# property. A YAML key written as `group.id: foo` under `mp.messaging.incoming.<channel>:`
+# registers ONLY as `mp.messaging.incoming.<channel>."group.id"` — literal quotes in the
+# property name — never as the plain `...group.id` that
+# `KafkaConnectorIncomingConfiguration.getGroupId()` (and `auto.offset.reset`) actually read.
+# Nothing errors; the connector silently uses its own default.
+#
+# copilot's `application.yaml` therefore deliberately declares NEITHER key — it carries a
+# comment saying they belong here. That is not a stylistic preference. Were `group.id` left to
+# its default it would fall back to `quarkus.application.name` = `openbank-copilot-service`,
+# which the KafkaUser ACL (kafka-copilot-mtls.yaml) does NOT grant Read on, and every poll
+# would take a silent GroupAuthorizationException. The group below and the ACL there are one
+# fact written twice; they must agree.
+#
+# Why a properties FILE and not env vars: `party-events-in` is a HYPHENATED channel name, and
+# quarkusio/quarkus#30106 + smallrye/smallrye-reactive-messaging#2028 document a live,
+# unresolved upstream bug where an env-var override of a single leaf property on a hyphenated
+# channel crashes Quarkus at startup (`SRMSG00071`). notification-service and
+# transaction-service both use this properties-file mechanism for the same reason.
+#
+# auto.offset.reset=earliest — SAFE HERE, and deliberate. The warning in the root CLAUDE.md is
+# about forcing `earliest` onto a consumer group *already running* on the default, which
+# re-reads the topic from the beginning and mass-replays. `copilot-service-party` is a BRAND
+# NEW group with no committed offsets, so `earliest` versus `latest` is the difference between
+# starting at the topic's oldest retained record and starting at its head — there is no
+# existing position to rewind. Starting at the oldest record is what we want: `openbank.party.
+# events` has `retention.ms=604800000` (7 days), so a fresh consumer picks up any PARTY_ERASED
+# emitted in the last week that copilot could not act on because it had no Kafka identity at
+# all. Those erasures are real and were missed; replaying them deletes conversation rows that
+# GDPR Art. 17 says should already be gone. The bound is the retention window, not all history.
+#
+# config_ordinal=500 forces this source to outrank the baked application.yaml (~254).
+# Loaded via QUARKUS_CONFIG_LOCATIONS on the Deployment (copilot-service.yaml).
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: copilot-service-msg-override
+  namespace: platform
+  labels:
+    app.kubernetes.io/name: copilot-service
+    app.kubernetes.io/part-of: platform
+data:
+  override.properties: |
+    config_ordinal=500
+    mp.messaging.incoming.party-events-in.client.id=copilot-service-party-${quarkus.uuid}
+    mp.messaging.incoming.party-events-in.group.id=copilot-service-party
+    mp.messaging.incoming.party-events-in.auto.offset.reset=earliest
--- a/openbank-infra/gitops/components/copilot/copilot-service.yaml
+++ b/openbank-infra/gitops/components/copilot/copilot-service.yaml
@@ -1,5 +1,5 @@
-# copilot-service: the customer-facing AI assistant backend (ADR-0089). Stateless — no DB,
-# no Kafka. The governed reasoning loop runs server-side behind a provider-agnostic model
+# copilot-service: the customer-facing AI assistant backend (ADR-0089). The governed
+# reasoning loop runs server-side behind a provider-agnostic model
 # gateway. READ tools call the customer edge AS the customer (propagated bearer → ownership
 # enforced at the edge, ADR-0065). The HTTP @Authorize gate queries the OPA sidecar
 # (data.openbank.rest.allow) and the copilot tool-dispatch gate queries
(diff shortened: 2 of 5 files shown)


JiRaska:

Checked specifically for scope creep in the Kafka/NetworkPolicy grants (task item 4):
- `kafka-copilot-mtls.yaml`'s KafkaUser ACL grants only `Read, Describe` on `openbank.party.events` plus `Read` on the consumer group `copilot-service-party` — no Write, matching the stated "consumer-only, produces nothing" claim.
- `networkpolicy-copilot-egress.yaml`'s new egress rules are narrowly scoped: the CNPG rule selects `cnpg.io/cluster: copilot-db` (not `podSelector: {}`, which the file's own comment calls out as the alternative it deliberately avoided), and the Kafka rule is namespace-scoped to `messaging` on 9093 only.
- `kafka-scheme-accepted-acl.yaml` change is a single one-line addition of `copilot-service` to an existing hand-kept `resourceNames` list for ESO's `eso-kafka-cert-reader` Role — consistent with the pattern already used by the other services in that list, and the PR reuses (not duplicates) the existing `kafka-cluster-ca-truststore` Secret in `platform`.

On `auto.offset.reset: earliest` for `copilot-service-party`: this is a brand-new consumer group with no prior committed offsets (first time copilot has had any Kafka identity), so the root `CLAUDE.md` mass-replay warning (which is about forcing `earliest` onto a group *already running* on the connector default) does not apply here — the PR's own comment states this distinction correctly.

Also note: this PR fixes the `openbank_copilot` → `openbank_copilots` DB name typo from #3710, which independently explains why the newer ReplicaSet was CrashLooping — worth confirming the Postgres database is actually named `openbank_copilots` (plural) in `postgres.yaml` before merging, since a batch of two DB-adjacent fixes landing together is exactly the kind of thing worth double-checking against the live manifest.

Verified: ACL/NetworkPolicy scope is narrow and consumer-only; the earlier stated replay-risk pattern does not apply to a new group; ESO Role grant pairing (`check-kafka-cert-reader-grant.py`) is satisfied by the matching diff to `kafka-scheme-accepted-acl.yaml`. Did not independently confirm `postgres.yaml`'s actual database name against a live cluster.

PR #3887, opened 2026-08-06
fix(copilot): governance manifest still declared no database after #3711 added one

JiRaska:
## main is red, and it has been since #3711 merged

`Governance manifest gate (ADR-0071, enforced)` fails on `main`:

```
[generate-governance] 56 modules (41 with lineage, 0 unverified databaseName, 2 gaps)
[generate-governance] gaps:
  - openbank-copilot-service: declares ownsNoDatabase: true but owns 1 Flyway migration(s)
    under src/main/resources/db/migration — drop 'ownsNoDatabase' and declare databaseName
  - openbank-copilot-service: declares primaryDatastore='Redis' but owns 1 Postgres Flyway
    migration(s) — the primary store is PostgreSQL
##[error]Governance drift — 2 module(s) missing or with an invalid governance.yaml
```

**The consequence is bigger than a red check.** The step lives in `Admin UI build` and runs *before*
the `Test` step, so when it fails **vitest never executes on `main`**. Measured across the 40 most
recent `ci.yml` runs: 13 success, 19 skipped, **8 failure — all 8 at this same step**. The admin-ui
suite has not been exercised on those commits at all, which is a silent loss of coverage rather than
a visible one.

## The gate is right on both counts

`openbank-copilot-service/src/main/resources/db/migration/V1__conversation_history.sql` exists, and
`application.yaml` configures a `postgresql` datasource with `migrate-at-start: true`. The manifest is
curatorial; the gate derives the contradiction from the tree. #3711 added the store and left the
manifest describing the service as it was before.

## On the plural

`openbank_copilots` is deliberate — it is what `postgres.yaml`'s CNPG `initdb.database` creates. The
datasource URLs have said the **singular** `openbank_copilot` since #3710 added them, which is a
separate and live fault: every pod carrying the conversation store has been CrashLooping in Flyway,
and the sandbox is still served by a pre-#3710 ReplicaSet. That half is fixed by **#3885**. Recording
the canonical name in the manifest is what stops the two drifting apart again without anything
noticing — which is the whole point of ADR-0071's derived check.

## Scope

One file, one hunk. `openbank-admin-ui/governance.json` is gitignored and CI-generated
(`.gitignore:229`), so no derived artefact is committed here.

## Not verified

- I did not run the generator locally — it needs an installed `node_modules` this worktree does not
  have. The change is validated against the gate's own two messages, which name the required shape
  exactly (`drop 'ownsNoDatabase' and declare databaseName`; `the primary store is PostgreSQL`).
- I did not audit the other 55 manifests for the same class of drift. The gate reports 2 gaps and
  both are this service, so there is no second instance today — but that is the gate's word, not an
  independent sweep.
- This does not make copilot's database reachable. It records what the service owns; #3885 is what
  fixes the connection.

Refs #3711, #3885, ADR-0071

### DIFF
--- a/docs/runbooks/svc-copilot.md
+++ b/docs/runbooks/svc-copilot.md
@@ -6,7 +6,7 @@ exercised DR drill, tracked as TTL'd attestations, never faked here. -->
 # Runbook — openbank-copilot-service
 
 > Operational runbook for the `copilot` service. Data domain **platform**,
-> classification **confidential**, datastore **Redis**.
+> classification **confidential**, datastore **PostgreSQL**.
 
 ## Service identity
 
@@ -15,7 +15,7 @@ exercised DR drill, tracked as TTL'd attestations, never faked here. -->
 | Service | `openbank-copilot-service` |
 | HTTP port | `8131` |
 | Data domain | platform |
-| Datastore | Redis (database `—`) |
+| Datastore | PostgreSQL (database `openbank_copilots`) |
 | Classification | confidential |
 | Retention | 1 year |
 | Lineage role | internal |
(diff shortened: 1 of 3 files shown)


JiRaska:
**Second commit — the gate found the next layer itself, which is the gate working.**

After the manifest fix the count went 2 gaps → 1, and the remaining one named the exact residue:

```
openbank-copilot-service: declares databaseName='openbank_copilots'
  but its datasource URL points at 'openbank_copilot'
```

Both dev defaults in `application.yaml` carried the singular from the day #3710 added them. Fixed.

**Scope, so the two PRs are not confused with each other:**

| | `application.yaml` dev defaults | gitops `REACTIVE_URL` / `JDBC_URL` |
|---|---|---|
| this PR | ✅ | — |
| #3885 | — | ✅ |

They are disjoint — #3885 does not touch this file (verified with a 3-dot diff), so there is no
conflict and no duplicated work. **Both are needed.** This one makes `main` green and a local run
correct; #3885 is what makes the deployed pod able to connect. Until #3885 lands, every pod carrying
the conversation store keeps CrashLooping in Flyway with `SocketTimeoutException`, and the sandbox
continues to be served by a pre-#3710 ReplicaSet.

**Worth stating because it changes how the first commit reads:** I opened this PR believing the
manifest was the whole fix. It was not — it was the first of three places the singular/plural
disagreement lived (manifest, dev defaults, gitops env). The governance gate is what walked me
through them one at a time, and that is exactly the behaviour ADR-0071's derived check exists for: a
curatorial manifest that cannot silently disagree with the tree.

`openbank-admin-ui/governance.json` remains gitignored and CI-generated; nothing derived is committed.

PR #3888, opened 2026-08-06
docs(infra): measure NetworkPolicy coverage per pod, and name what a default-deny would sever

JiRaska:
## What this is

Stage 2 of the #2691 staged rollout. **No default-deny is enabled anywhere, no policy is added, nothing is applied.** Stage 1 (#3799) shipped the repo-side enumeration and the flow-log switch; its last exit criterion — "start with a namespace whose workloads are already fully policy-selected" — needs a number that cannot be derived from the repo, and this supplies it.

## Why the repo cannot answer it

Coverage is a property of **pods**, not namespaces. CNPG Postgres instances are created by the CloudNativePG operator: no Deployment for `gen-network-policies.py` to derive from, no component directory, no presence in gitops at all. "This namespace has NetworkPolicies" is the reassuring flattening — and it is wrong in the direction that gets a platform severed.

`openbank-infra/scripts/netpol-coverage.py` is read-only (`kubectl get pods/networkpolicies -o json`), costs nothing, and mutates nothing. Its unit tests assert the **uncovered** side of every case, because the selector bug that matters is the one that over-matches and reports a default-deny as safe.

## Measured, 2026-08-06, 400 running pods

| | Ingress | Egress |
|---|---|---|
| pods selected by at least one policy | **109** | **5** |
| pods selected by none | **291** | **395** |

The "24 namespaces without a policy" figure understates this by an order of magnitude, in the reassuring direction.

## What a default-deny would break today

**In 40 namespaces the only uncovered workload is `postgresql`.** One finding, two faces:

- **Today** every CNPG instance in the fleet accepts a connection from any pod in the cluster on 5432. The allow-lists protect the services and leave the data reachable directly. This is the largest measured exposure in the issue, and it needs a policy that does not exist — not a rollout.
- **Under a namespace default-deny** those same pods are the ones newly cut off, so each of those 40 namespaces would sever its service from its own database. The service does not error usefully: it times out in Flyway at boot and CrashLoops, and nothing in the stack trace says NetworkPolicy.

And it is not cleanly reversible by the usual reflex. `temporal-network-policies.yaml` already records the mechanism from the one namespace that has a default-deny: the CNPG operator polls each instance manager on :8000, default-deny drops the poll, the cluster never reports Healthy, and **the ArgoCD sync for that namespace hangs** on "waiting for healthy state" — so the commit that would relax the policy cannot apply itself. Any namespace default-deny needs two carve-outs first (intra-ns 5432; `cnpg-system` → 8000 + 9187).

**Egress is not a variant of the same change.** An egress policy denies same-namespace traffic too, so it needs the workload's whole dependency graph. Four workloads have one today and one is broken by it right now: `copilot-service-egress-allow-list` lists DNS, LiteLLM, Redis, IAM, customer-edge and OTLP — and no 5432. The pod CrashLoops in Flyway with `SocketTimeoutException: Connect timed out` against a database in its own namespace. One missing rule, one dead service, no diagnostic naming the cause (fix in flight: #3885). Beyond that, **87 pods across 50 namespaces** need the EKS Pod Identity credential endpoint, which no `namespaceSelector` can express — it takes an `ipBlock`, and without it every AWS call fails at credential fetch, looking like an IAM problem for as long as you let it.

## Staged proposal

1. **Not default-deny first — the CNPG allow-list first.** Generalise `temporal-cnpg-operator` into a derived policy for every CNPG cluster. It closes the 5432 exposure on its own merits and is the precondition for anything else.
2. **First namespace: `finrep`.** One pod, no CNPG instance, not money-path, callers already derived, and a failed regulatory report is noticed and re-run rather than lost. Then `analytics`, `developer-portal`. `admin-ui` only later (it is the window you would diagnose a mistake through). **`customer-edge` last of the seven** despite qualifying on the number — it is the public edge and the caller in two of the three `DROPPED` edges stage 1 found; a namespace with known-missing declarations is the wrong place to start.
3. Everything else waits on step 1 plus a week of harvested flow logs.

Verification recipe between namespaces is in the runbook: pods Ready with no added restarts, **ArgoCD `Synced/Healthy`** (the CNPG deadlock probe, not a formality), unbroken Prometheus scrape, the golden path from a real caller rather than a port-forward (a port-forward enters via the API server and proves nothing about pod-to-pod policy), and `netpol-coverage.py` still reporting UNCOV=0. One full scheduling day between namespaces.

## How complete this is — and what breaks if I am wrong

Not very, and the gaps are the dangerous kind. Stated here rather than in a footnote:

- **It is a snapshot of running pods.** A CronJob, a monthly scheduler (interest capitalization, year-close, FINREP/COREP render), a release-time Job, any break-glass path taken by hand — all invisible, and every one is a flow a default-deny severs with no warning until the day it matters.
- **It answers "is this pod selected by a policy", never "does that policy contain every flow this pod needs".** A covered pod can still be missing a rule; `copilot-service` is exactly that.
- KEDA scale-to-zero makes "no pod running" and "no such workload" identical here.
- It reads the cluster, so it says nothing about a namespace that exists in gitops and has not synced.

If I am wrong about a namespace, the failure mode is connection timeouts with no attributable error, across everything in it at once — which is why this PR proposes an ordering and a carve-out rather than a policy set. The honest position on the issue is unchanged: **the allow-lists are not complete enough to enable default-deny fleet-wide.** The seven zero-uncovered namespaces are the only defensible starting points, and even they should wait for the CNPG work.

## Verification done

- `netpol-coverage.py` cross-validated against an independently written ad-hoc probe: both report 109/400. Unit tests: 7 pass, including the CNPG shape and namespace isolation.
- Stage-1 numbers re-measured live: 110 policies, 74 namespaces, 0 default-deny — unchanged.
- `check-network-policy-code-edges.py` re-run: same 4 edges, no drift.
- `check-stale-comment-references.py`: OK.
- No cluster mutation of any kind; no `kubectl apply`, no `tofu apply`.

Refs #2691

### DIFF
--- a/docs/runbooks/0010-networkpolicy-default-deny-measurement.md
+++ b/docs/runbooks/0010-networkpolicy-default-deny-measurement.md
@@ -132,3 +132,167 @@ property of the source tree at all. That set has to come from the flow logs.
 * and only then, one namespace at a time — starting with a namespace whose
   workloads are already fully policy-selected, where default-deny is provably a
   no-op for existing pods.
+
+## The measurement that decides the order: coverage is PER POD
+
+The last exit criterion above needs a number the repo cannot produce. Which pods
+are already policy-selected is a property of the cluster, not of gitops: CNPG
+Postgres instances are created by the CloudNativePG operator, so they carry no
+Deployment for `gen-network-policies.py` to derive from and appear in no
+component directory at all.
+
+`openbank-infra/scripts/netpol-coverage.py` reports it. It is read-only
+(`kubectl get pods/networkpolicies -o json`), costs nothing, and mutates nothing:
+
+```
+python3 openbank-infra/scripts/netpol-coverage.py
+python3 openbank-infra/scripts/netpol-coverage.py --direction Egress
+```
+
+Measured 2026-08-06 against the sandbox, 400 running pods:
+
+| | Ingress | Egress |
+|---|---|---|
+| pods selected by at least one policy | **109** | **5** |
+| pods selected by none | **291** | **395** |
+
+The 24-namespaces-without-a-policy figure understates this by an order of
+magnitude, and understates it in the reassuring direction. 27% of running pods
+are policy-selected for ingress; 1% for egress.
+
+### The 40-namespace blocker: every database is the uncovered workload
+
+In **40 namespaces the ONLY uncovered workload is `postgresql`**. That is one
+finding with two faces, and both matter:
+
+* **Today, without any default-deny**, every CNPG instance in the fleet accepts
+  a connection from any pod in the cluster on 5432. The allow-lists protect the
+  services and leave the data they guard reachable directly. This is the largest
+  measured exposure in the issue and it does not need a rollout to fix — it needs
+  a policy that does not exist.
+* **Under a namespace default-deny** those same pods are the ones newly cut off,
+  so applying default-deny to any of those 40 namespaces severs each service from
+  its own database. The service does not error usefully: it times out in Flyway at
+  boot and CrashLoops, and nothing in the stack trace says NetworkPolicy.
+
+And the failure is worse than an outage, because it is not cleanly reversible by
+the usual reflex. `components/temporal/temporal-network-policies.yaml` already
+records the mechanism, discovered the hard way in the one namespace that has a
+default-deny today: the CloudNativePG operator polls each instance manager on
+:8000, a default-deny drops that poll, the cluster never reports Healthy, and the
+**ArgoCD sync for that namespace hangs** on "waiting for healthy state" — so the
+follow-up commit that would relax the policy cannot apply itself.
+
+Any namespace default-deny therefore needs two carve-outs before it, per
+namespace, not one:
+
+* intra-namespace 5432 for the service → its own instances, and
+* `cnpg-system` → the CNPG pods on 8000 (instance-manager status) and 9187
+  (instance metrics),
+
+which is exactly the shape of `temporal-cnpg-operator`. Generalising that from a
+hand-written one-off into a derived policy for every CNPG cluster is the concrete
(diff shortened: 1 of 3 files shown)


JiRaska:

Checked the specific concern for a coverage measurement: whether it counts policies that merely *exist* versus policies that actually *select pods*. It does the latter — `netpol-coverage.py`'s `selects()` implements the real podSelector grammar (matchLabels AND'd, plus In/NotIn/Exists/DoesNotExist matchExpressions) and `coverage()` only marks a pod covered if some policy's selector actually matches its labels, not merely if a policy exists in the namespace. `netpol_coverage_test.py`'s `test_cnpg_pod_is_uncovered_when_policies_select_only_the_service` is exactly the case this task asked to check for — a namespace full of allow-lists that select the app pod but not its co-tenant CNPG Postgres instance — and it's asserted correctly (Postgres reported uncovered).

The test suite's own framing is worth noting positively: its docstring states the failure mode it guards against is one-directional (a selector bug that over-matches and falsely reports a pod as covered/safe), and every test case asserts both the covered AND uncovered sides rather than only the reassuring one.

The runbook addition is unusually self-critical for this kind of doc — it explicitly lists what the measurement still can't see (CronJobs/scheduled Jobs, KEDA scale-to-zero pods, unsynced namespaces, "covered but missing a rule" cases like the copilot/#3885 example it cites), and concludes the fleet-wide answer is still "no" rather than using the new tooling to declare victory.

Verified: `selects()`/`coverage()` logic against the test suite: empty selector = default-deny shape (selects everything, tested); direction (Ingress/Egress) honoured; no cross-namespace leakage; completed pods excluded. Did not run the script against a live cluster to confirm the reported sandbox numbers (109/291 ingress, 5/395 egress) — those are presented as a point-in-time measurement in the runbook, not something this diff can verify statically.

Issue #1302, opened 2026-07-16
Closing audit P1: no owner of the business date, no period lock below fiscal year, structural false drift


JiRaska:
Tracking issue for the P1 tail of `docs/audits/2026-07-16-closing-audit.md`. **P0 is complete** (see bottom); these are the systemic items P0 deliberately did not touch. Each has file:line detail in the audit report — this issue exists so the work is visible and ordered, not to restate it.

## Why these are grouped

They are not seven independent bugs. Five of them are the same root cause wearing different clothes: **nothing owns the accounting day.** Fix the calendar and the lock, and items 3–5 shrink from "design work" to "wire it up".

---

### 1. No owner of the business date — three clock regimes coexist *(root cause; do first)*

| regime | where |
| --- | --- |
| `Europe/Prague`, hardcoded | ledger core (`LedgerService`, `YearCloseService`), TieOut/FxReval/CnbRate schedulers |
| `UTC` | **every** CDI `ClockProducer` (fx, balance, lending, statement, libs default), plus hardcoded `Clock.systemUTC()` in `BalanceService`, `InterestService`, `WithholdingRemittanceService` |
| **JVM default** | `statement-service` `CloseOrchestrator` (raw `Instant::now` / `LocalDate::now`, bypassing its own injected Clock) + its close cron with **no `timeZone`** |

ledger-service alone runs two regimes: its CDI `ClockProducer` says UTC while `LedgerService`/`YearCloseService` build `Clock.system(BANK_TIME)` in their constructors. Correctness today rests on cron times sitting far from midnight. Moving `openbank.reconciliation.cron` past midnight Prague would silently reconcile the wrong day.

**Wanted:** one authority for "current accounting day" + day state (`OPEN → CUTOFF → TIED_OUT → LOCKED`), consumed by every close scheduler. Kill the three-regime split; give the statement cron an explicit `timeZone`; make `CloseOrchestrator` use its injected Clock.

### 2. No period lock below fiscal year

`LedgerService.postJournal` only guards `requireOpenPeriod(entryDate.year)` (ATTESTED years). `entryDate` is caller-supplied and freely backdatable into days already tied out, reconciled and statement-closed. Reversals inherit the original `entryDate`, so reversing in July rewrites March. **ADR-0078's header claims "End-of-Day, End-of-Month and End-of-Year closes are live (period lock…)" — no day or month lock exists in code.**

[...]

PR #3889, opened 2026-08-06
fix(sanctions): make the scheduled-refresh de-duplication zone-consistent, and retire two accounting-clock exemptions

### LINKED ISSUE #2963: Accounting-clock gate: three money-path sites derive an accounting date from a locally-chosen zone (ADR-0207 D1)
`check-accounting-clock.py` (added in #2962, ADR-0207 D1) bans binding a clock to a locally-chosen
time zone in the domain/application layer of a money-path service. The accounting date is a domain
value with its own calendar and cutoff, not a projection of wall-clock time; deriving it per caller
is what let `openbank-ledger-service` run a UTC `ClockProducer` bean and a
`Clock.system(Europe/Prague)` at the same time, disagreeing about the date for two hours a day, half
the year, with nothing able to detect it.

The gate ships **enforced**, with these three pre-existing sites in its `ALLOWLIST`. They are listed
there rather than silently excluded so each one has to be justified, and the script **fails on a
stale entry** — if a site is fixed and the entry is left behind, CI goes red. The debt cannot
quietly become permanent.

### The three

1. **`openbank-transaction-service` — `domain/settlement/SettlementDateResolver.kt:33`**
   `ZoneId.of(...)` for settlement/booking-date rolling. This one is entangled with issue #1302
   item 3 (reconciliation false drift): the resolver books post-16:00 and weekend payments with a
   next-business-day `bookingDate` while the balance projection applies deltas immediately, ignoring
   `entryDate`. Changing the zone handling in isolation risks moving the drift rather than fixing
   it, so it should be done together with that item, not before.

2. **`openbank-fx-service` — `application/usecase/CnbRateIngestionService.kt:41`**
   `ZoneId.of(...)` for the ČNB fixing day. Genuinely a *market* calendar question (which fixing
   day a rate belongs to), so the fix is to state whether it wants the accounting day or the ČNB
   publication day — they are not always the same, and the current code does not say which it means.

3. **`openbank-sanctions-service` — `application/usecase/SanctionsListService.kt:158`**
   `ZoneId.systemDefault()`. The worst of the three in principle — JVM-default zone means the answer
   depends on the host — but the least consequential in practice, since it is list-refresh
   bookkeeping rather than an accounting date. Should just take the injected clock.

### Done when

Each site either uses `AccountingClock` (or states in a comment why a market/publication calendar is
correct there and is genuinely not an accounting date), and its `ALLOWLIST` entry is removed. The
gate's stale-entry check enforces that the two move together.

Refs #2962, #1302

JiRaska:
Closes #2963. Refs #2962, #1302.

The `check-accounting-clock.py` gate (ADR-0207 D1) was green *because* three money-path sites were
allowlisted — a control that exists and is not in force. This decides each one on its merits rather
than sweeping them, and moves each exemption together with its verdict.

## Per site

| Site | Zone | Verdict |
|---|---|---|
| `openbank-sanctions-service` `SanctionsListService` | `now.zone` (whatever clock is injected; UTC in prod) | **Defect fixed**, entry removed |
| `openbank-transaction-service` `SettlementDateResolver` | `AccountingClock.BANK_ZONE` (Europe/Prague) | **Ownership fixed**, entry removed |
| `openbank-fx-service` `CnbRateIngestionService` | `Europe/Prague`, unchanged | **Justified exception**, entry KEPT with a decided reason |

### sanctions — a real defect, not a style question

`isDueForScheduledRefresh` compares `now` (from the injected clock, UTC in production) with
`lastUpdatedAt` read in `ZoneId.systemDefault()`. The comparison is a same-minute equality on
`(year, dayOfYear, hour, minute)` — meaningful only if both sides are in one zone. Off UTC it was
wrong in **both** directions: a just-completed run reads as not-yet-run (the list re-imports on the
next tick inside the due minute), and an unrelated run offset by exactly the zone offset reads as
already-run (a due refresh is skipped). Taking the zone from `now` makes both sides one zone by
construction. No zone was chosen.

### transaction — same value, one owner

`BANK_ZONE` was a second copy of `Europe/Prague`. The booking date *is* an accounting date, so it
must be the day the ledger and the year-close also think it is; a locally restated zone is exactly
what let two components in `openbank-ledger-service` disagree for two hours a day. The substitution
is value-identical, so **no booking or value date moves**. The 16:00 cut-off is deliberately not
`AccountingClock.cutoff`: that is the accounting-day rollover, this is a submission deadline.

The issue flagged this as entangled with #1302 item 3 (reconciliation false drift). It still is —
and nothing here touches it, precisely because the change moves no dates. #1302 item 3 stays open.

### fx — the question the issue asked, answered

The issue asked whether the ČNB fixing day means the accounting day or the publication day. It is
the **publication** day: a fixing is declared by the ČNB for a named ČNB business day and published
at 14:30 Prague, so the boundaries that bound `validFrom`/`validTo` are the publisher's, not the
bank's books. It equals `AccountingClock.BANK_ZONE` today only because the ČNB is a Prague
institution and this entity keeps its books in Prague; binding it to `AccountingClock` would encode
a coincidence as a dependency and make a future divergence (a non-Czech entity consuming the ČNB
fixing) silent. The allowlist entry is kept and its reason rewritten from an open question to a
decision, with the same reasoning stated in the file.

## What I proved

- **The gate is falsifiable in both directions, by reintroducing each pattern and reading the
  error** — not by trusting the green. `ZoneId.systemDefault()` back into `SanctionsListService`:
  `::error::...:165: ZoneId.systemDefault()`, exit 1. `ZoneId.of("Europe/Prague")` back into
  `SettlementDateResolver`: `::error::...:44: ZoneId.of(...)`, exit 1. Restored: exit 0.
- **Both new sanctions tests fail against the pre-change code.** They force the JVM default zone to
  `Europe/Prague` and restore it in a `finally` — the defect is invisible on a UTC host, which is
  why nothing caught it in a UTC container. Reverting the one-line fix: both FAILED.
- **A vacuous assertion was caught and fixed mid-review.** The first version of the "skips a list
  already refreshed this minute" test passed against the broken code, because an unstubbed
  `repo.findByListType` made mockk throw inside `refresh()`, `scheduledRefresh()` swallowed it, and
  `importList` was never reached — the assertion held for the wrong reason. Stubbing the whole
  refresh path so a wrongly-due list would *succeed* makes it red against the old code.
- **The settlement zone test uses a literal instant**, not one derived from `BANK_ZONE`. Deriving
  both sides moves input and expectation together and asserts nothing. `2026-06-03T22:30:00Z` is
  already 2026-06-04 in Prague; with the cut-off widened to end-of-day (so the zone is the only
  variable) the two zones book different days, asserted explicitly as unequal. Flipping `BANK_ZONE`
  to UTC: both new tests FAILED.
- `--rerun-tasks :<svc>:ktlintCheck :<svc>:detekt :<svc>:test` green for all three services, with
  `ktlintCheck`, `detekt` and `test` each appearing as executed (not `UP-TO-DATE`) in the output.

## What I did NOT verify — read this before approving

- **The transaction change is untestable by construction.** `ZoneId.of("Europe/Prague")` and
  `AccountingClock.BANK_ZONE` are the same value, so no test can distinguish the old code from the
  new by its output, and none here claims to. The added equality assertion catches a *future*
  divergence, not this change. If you want evidence this PR is safe for transaction-service, the
  evidence is that it is a no-op — read the diff, not the tests.
- **The fx zone choice is a decision, not a measurement.** I did not confirm against ČNB
  documentation that the fixing's business-day boundaries are Prague-local, nor consult anyone on
  whether this entity would ever consume the ČNB fixing from a non-Czech book. If the intended
  reading is "the accounting day, and ČNB happens to publish in Prague", then the right change is
  the opposite of the one here and the entry should go. I have stated my reading in the file so it
  can be disagreed with.
- **The fx path remains untested end-to-end**, and this PR does not change that. `#2963`'s ČNB site
  had a dead-URL incident where the feed 404'd for the service's whole life with every layer green;
  nothing here probes the feed, and I deliberately added no fx test rather than add one that would
  be equally vacuous.
- **No integration test.** The sanctions fix is covered by unit tests against a mocked repository.
  The scheduled path itself (`@Scheduled` + Vert.x context) is not exercised here.
- **`ZoneId.systemDefault()` elsewhere.** Scope is the three allowlisted money-path sites. The gate
  covers `domain/`+`application/` of money-path services only; infrastructure and non-money-path
  services are out of scope by design and I did not sweep them.
- **No threat-model update.** I judged none needed — the change removes a host-dependent value and
  moves no money — but that is my judgement and a reviewer may disagree.
- **Blast radius of a re-release.** This touches `src/main` in sanctions and transaction, so
  release-please will cut releases for both. The transaction one carries no behaviour change.

## Merge expectation

Money-path: two human approvals required. This PR is **not** expected to auto-merge, and that is the
correct final state. No override flag was or will be used.

### DIFF
--- a/.github/scripts/check-accounting-clock.py
+++ b/.github/scripts/check-accounting-clock.py
@@ -82,21 +82,14 @@ BANNED = [
 # exemption and its fix move together and this debt cannot quietly become permanent. Tracked in
 # issue #2963; see it for why each is not a one-line change.
 ALLOWLIST: dict[str, str] = {
-    "openbank-transaction-service/src/main/kotlin/com/openbank/transaction/domain/settlement/"
-    "SettlementDateResolver.kt": (
-        "#2963 — settlement/booking-date rolling, entangled with #1302 item 3 (reconciliation false "
-        "drift). Changing the zone handling alone risks moving the drift rather than fixing it."
-    ),
     "openbank-fx-service/src/main/kotlin/com/openbank/fx/application/usecase/"
     "CnbRateIngestionService.kt": (
-        "#2963 — ČNB fixing day may legitimately be a publication calendar rather than the "
-        "accounting day. The code does not currently say which it means; that has to be decided "
-        "before it can be mechanically converted."
-    ),
-    "openbank-sanctions-service/src/main/kotlin/com/openbank/sanctions/application/usecase/"
-    "SanctionsListService.kt": (
-        "#2963 — ZoneId.systemDefault() in list-refresh bookkeeping, not an accounting date. Worst "
-        "in principle (host-dependent), least consequential in practice; wants the injected clock."
+        "#2963 — DECIDED (not deferred): this is the ČNB PUBLICATION calendar, not the accounting "
+        "day. A fixing is declared by the ČNB for a named business day and published at 14:30 "
+        "Prague; the day boundaries that bound validFrom/validTo are the publisher's, not the "
+        "bank's books. Same value as AccountingClock.BANK_ZONE today only because the ČNB is a "
+        "Prague institution and this entity keeps its books in Prague — binding it to the "
+        "accounting zone would encode a coincidence as a dependency. The file states this in place."
     ),
 }
 
--- a/openbank-fx-service/src/main/kotlin/com/openbank/fx/application/usecase/CnbRateIngestionService.kt
+++ b/openbank-fx-service/src/main/kotlin/com/openbank/fx/application/usecase/CnbRateIngestionService.kt
@@ -38,6 +38,25 @@ class CnbRateIngestionService(
     private val clock: Clock,
 ) : CnbRateIngestionUseCase {
 
+    /**
+     * The ČNB **publication** calendar, deliberately NOT the accounting day
+     * ([com.openbank.libs.domain.calendar.AccountingClock.BANK_ZONE]) — issue #2963 asked which of
+     * the two this means, and the answer is the publication one.
+     *
+     * The ČNB declares a fixing for a named business day and publishes it at 14:30 Prague time;
+     * that day's boundaries are properties of the *publisher*, so `validFrom`/`validTo` here mark
+     * the window a published fixing is the current one, not a window in the bank's books. The
+     * distinction is invisible today because the ČNB is a Prague institution and this bank keeps
+     * its books in Prague, so the two constants hold the same value — but they are answers to
+     * different questions and would have to move independently if either ever changed (a
+     * non-Czech entity consuming the ČNB fixing; a ČNB calendar change). Binding this to
+     * `AccountingClock` would make the coincidence look like a dependency.
+     *
+     * This is why the `check-accounting-clock.py` ALLOWLIST entry for this file is KEPT rather
+     * than removed: the gate's rule ("the accounting zone is declared once") is correct and this
+     * value is not the accounting zone. The entry's reason now records a decision instead of an
+     * open question.
+     */
     private val zone: ZoneId = ZoneId.of("Europe/Prague")
 
     private val enabled: Set<String>
--- a/openbank-sanctions-service/src/main/kotlin/com/openbank/sanctions/application/usecase/SanctionsListService.kt
+++ b/openbank-sanctions-service/src/main/kotlin/com/openbank/sanctions/application/usecase/SanctionsListService.kt
@@ -14,7 +14,6 @@ import jakarta.enterprise.context.ApplicationScoped
 import jakarta.inject.Inject
 import jakarta.ws.rs.NotFoundException
 import java.time.Clock
-import java.time.ZoneId
 import java.time.ZonedDateTime
 import java.util.UUID
 
@@ -144,7 +143,16 @@ class SanctionsListService(
         if (now.hour != cronHour || now.minute != cronMinute) return false
 
         val lastRun = lastUpdatedAt ?: return true
-        val lastRunAt = ZonedDateTime.ofInstant(lastRun, ZONE_ID)
+        // Read the last run in the SAME zone as [now] — which is the injected clock's zone, UTC in
+        // production. This comparison is a same-minute equality on (year, dayOfYear, hour, minute):
+        // it is only meaningful if both sides are in one zone, and it used to read `now` from the
+        // clock and `lastRunAt` from ZoneId.systemDefault(). Those agree only when the JVM default
+        // happens to be UTC, so on any other host the de-duplication compared two different
+        // clocks and was wrong in BOTH directions: a just-completed run read as not-yet-run (the
+        // list re-imports on the next tick inside the due minute), and an unrelated run offset by
+        // exactly the zone offset read as already-run (the due refresh is skipped). Not a zone
+        // choice: taking the zone from `now` makes both sides one zone by construction.
+        val lastRunAt = ZonedDateTime.ofInstant(lastRun, now.zone)
         return !(
             lastRunAt.year == now.year &&
                 lastRunAt.dayOfYear == now.dayOfYear &&
@@ -155,6 +163,5 @@ class SanctionsListService(
 
     companion object {
         private val ALLOWED_DAYS = setOf("MON", "TUE", "WED", "THU", "FRI", "SAT", "SUN")
-        private val ZONE_ID: ZoneId = ZoneId.systemDefault()
     }
 }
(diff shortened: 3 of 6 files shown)


JiRaska:
**Review: 2 findings, none blocking. The fix is right; the evidence for it is weaker than the body claims.**

1. **The two new sanctions tests cannot go red on a UTC CI runner, and the falsification claim ("reverting the one-line fix: both FAILED") is host-dependent.** The pre-change code read the zone from `private val ZONE_ID: ZoneId = ZoneId.systemDefault()` in a **companion object** — a static initializer that runs once, at class load, i.e. during the first test in `SanctionsListServiceTest` that touches the service. `withDefaultZone("Europe/Prague")` sets `TimeZone.setDefault` inside the test body, which cannot retroactively change an already-captured `ZoneId`. On your Mac (`Europe/Prague` host default) `ZONE_ID` was Prague regardless, so both tests were genuinely red — but on the UTC CI runner `ZONE_ID` would have been UTC, identical to `now.zone`, and both tests pass against the *unfixed* code. So they are a regression guard only where the developer's host is off UTC; on CI the thing that actually prevents a re-introduction is `check-accounting-clock.py`, which you did falsify properly in both directions. Worth stating in the PR body so nobody later reads these two tests as the guard.

2. **Production blast radius is likely zero today, which the body's "wrong in BOTH directions" framing overstates.** The injected clock is `Clock.systemUTC()` (`SanctionsListService.kt:34` and `openbank-sanctions-service/src/main/kotlin/com/openbank/sanctions/infrastructure/ClockProducer.kt:16`), and the containers are UTC, so `now.zone` and `ZoneId.systemDefault()` agreed in the deployed path. The change is still correct and worth making — it removes a host dependency from a money-path de-duplication, and the defect is real the moment anything runs off UTC or the clock producer changes — but "a due sanctions refresh is skipped" has not been happening in production. Framing it as latent rather than live changes how a reviewer weighs the risk of the release this cuts.

Verified without finding a problem: `now.zone` is genuinely zone-consistent by construction, not a zone *choice* — correct call; `scheduledRefresh` is already a `suspend fun`, so the `HR000068` scheduler trap does not apply and the direct-call tests are not hiding it; `TimeZone.setDefault` is restored in a `finally`, so no later test in the JVM is retuned; the "stub the whole refresh path so a wrongly-due list would succeed" correction is real — without it the mockk failure inside `refresh()` is swallowed by `scheduledRefresh()`'s catch and the assertion holds for the wrong reason, exactly as described; the transaction change is value-identical (`AccountingClock.BANK_ZONE` is `Europe/Prague`) so no booking or value date moves, and the added equality assertion is honestly labelled as a future-divergence guard rather than a proof of this change; the fx allowlist entry is kept with the reasoning restated in the file rather than only in the gate, so the two cannot drift apart.

The fx reading (publication calendar, not accounting day) is a judgement I can't check from the repo — it is the one place a human with domain knowledge is genuinely required, and the PR is right to flag it.

*Automated agent review — not an approval. `openbank-sanctions-service` and `openbank-transaction-service` are money-path: the 2-human-approval rule still applies.*

JiRaska:
Second review pass. **No blocking finding** — the zone fix is right and the scheduler traps do not apply. One thing neither the body nor the existing review covers.

Cleared first: `scheduledRefresh` is a `suspend fun` under `@Scheduled(… concurrentExecution = SKIP)` with no `runBlocking` in the production path, so the HR000068 trap is out of scope. Both new tests are block-bodied, so JUnit5 will actually run them — no `fun x() = runBlocking {}` shape. The direct-call-only test style is an accepted gap here: the fix is pure zone arithmetic, and a direct call cannot hide anything this diff touches.

**The finding: "de-duplication" is a per-JVM guarantee, and there is a window where the fleet has two JVMs.**

The dedup key is `SanctionsList.lastUpdatedAt` compared to `now` on `(year, dayOfYear, hour, minute)` — a **non-atomic read-check-write**: `listSanctionsLists()` → `isDueForScheduledRefresh` → `refresh()` → `repo.markUpdated(...)`. No DB unique constraint, no advisory lock, no `ClusterLock`. (The `SELECT … FOR UPDATE SKIP LOCKED` in this service is in the *outbox* repository — a different path.) `ConcurrentExecution.SKIP` serialises only within one JVM.

`replicas: 1` makes that fine at steady state. But the workload is `kind: Rollout` with a **canary** strategy whose steps are `setWeight: 10` → `pause: 5m` → `setWeight: 30` → `pause: 5m`, so canary and stable coexist for at least ten minutes on every deploy, both running the 60s scheduler. A deploy window overlapping a list's due minute has both pods read `lastUpdatedAt` as stale and both call `importList` for the same list.

**Impact is low, and I checked rather than assumed:** `SanctionsImportService.importList` is upsert-based throughout (`entryRepo.upsertAll(...)` on every batch path), so a duplicate concurrent refresh converges rather than corrupting. So this is not a reason to hold the PR — but the word "de-duplication" in the title reads as a stronger guarantee than the mechanism provides, and on a compliance service that distinction is worth writing down somewhere before someone relies on it.

I confirm the existing review's point 1 as the more consequential one: the removed `private val ZONE_ID = ZoneId.systemDefault()` lived in a `companion object`, so `TimeZone.setDefault` inside a test body cannot retroactively change it — which is exactly why the old tests were green on a UTC runner and told you nothing.

Issue #3344, opened 2026-08-02
Deployed == main: daily drift gate between gitops image tags and version.txt


JiRaska:
Part of #3343 (operational maturity tracker, level 1 → 2).

## What

A daily gate that compares the image tag in `gitops/apps/*.yaml` with `version.txt` on `main` per service, and opens an issue when drift exceeds N days.

## Why

Sandbox ran lending 0.11.5 while main was at 0.20.2 — what runs is not what is tested. Drift today is discovered by accident, not by a check.

## Acceptance criteria

- A scheduled workflow (or extend an existing fleet-health check) diffs deployed image tags against `main`'s `version.txt` for every service.
- Drift > N days opens/updates a `fleet-health`-style issue (dedupe — one open issue per service).
- The gate's own failure (cannot read tags) is itself visible, not silent.

PR #3891, opened 2026-08-06
fix(deploy-drift): reopen recently hand-closed drift issues instead of stacking new ones

### LINKED ISSUE #3597: deploy-drift: openbank-onboarding-service runs 0.5.8, main is 0.5.9
## Deployed != main: `openbank-onboarding-service`

| | |
|---|---|
| deployed | `0.5.8` (tag `sandbox-31e6692a`, 25d old) |
| main | `0.5.9` |
| status | drift |
| manifest | `openbank-infra/gitops/components/onboarding/onboarding-service.yaml` |

deployed 0.5.8 (25d old, 31e6692a5ed2) vs main 0.5.9 — drift beyond 7d

What runs is not what is tested (issue #3344). Deploy the current build and
this watch closes the issue itself. Closing it by hand does not deploy
anything — the watch reopens it while the drift persists. To accept the
drift knowingly, say so in a comment; a silent drift is the only
unacceptable state.

_Automated: `Deployed == main watch` run 31107298613, threshold 7d. Refreshed in place while the drift persists._

JiRaska:
## Why

On 2026-08-06 a human closed #3597 (onboarding-service) at 05:42 UTC while the drift was still active. The 06:30 UTC schedule run (`31077523776`) found no **open** issue for the marker and stacked a brand-new #3833 for the same ongoing drift — one drift, two issues, split history.

A hand-closed `deploy-drift` issue is not a deployed build. Closing the issue tracks a human gesture, not a cluster fact.

## What changes

- The escalate step now pairs drifted services against open issues **and issues closed within the last 30 days**, matched by the `<!-- deploy-drift:<service> -->` marker (first updated-desc page of closed issues, which covers the whole window).
- A recently closed match is **reopened in place with an explanatory comment** instead of creating a duplicate. An open match is still preferred over a closed one; a closed issue older than the window stays closed and a fresh issue is created.
- Issue body copy updated: it no longer suggests that closing the issue by hand `accepts the drift knowingly` — that contract now lives in comments, since the watch reopens hand-closed issues.
- Auto-close lane (close open issues whose service caught up) is unchanged.

## Verification

- YAML parses; the github-script block passes a JS syntax check.
- The PR lane of this workflow (self-test + declaration gate) runs on this PR because the workflow file itself is in its `paths:` filter.
- After merge, the reopen path can be exercised end-to-end: hand-close the current onboarding drift issue and `workflow_dispatch` the watch — it should reopen that same issue with the `Drift persists — reopening` comment rather than create a new one. (Natural exercise would also occur on the next hand-close.)

## Notes

- Current state was reconciled manually to match this intent: #3597 reopened with a comment, #3833 closed as superseded.
- Commit created via the GitHub API, so it is not GPG-signed with a local key.

Refs #3344


JiRaska:
**Verdict: fixes a real, observed incident correctly; the reopen path itself is unexercised pre-merge.** 0 blocking findings, 2 notes.

- Root cause and fix line up: #3597 was hand-closed at 05:42 UTC while onboarding-service still drifted, and the 06:30 run found no *open* issue for the marker and stacked #3833 — one drift, two issues. The fix (index recently-closed issues by the same `<!-- deploy-drift:<service> -->` marker, reopen-in-place within a 30-day window instead of creating a new one) matches that failure mode directly.
- **Unfalsified until merge** (the PR's own "Verification" section says as much): there's no dry-run or fixture exercising the reopen branch — only a JS syntax check and the unrelated PR-lane gate. The stated plan is to hand-close the current onboarding drift issue and `workflow_dispatch` after merge to prove it reopens rather than duplicates. Worth actually doing that before calling this closed, given the gate lens this repo applies elsewhere (a check that has only ever passed is unfalsified) — an inline `github-script` block has no local test harness, so a real dry run is the only falsification available.
- Minor: `closedRecent` fetches only the first `per_page: 100` page sorted by `updated: desc`, justified as "anything older falls outside the reopen window anyway" — but the sort key is `updated_at`, not `closed_at`, and the filter is on `closed_at`. A closed issue that's within the 30-day reopen window by `closed_at` but has no recent unrelated comment bumping `updated_at` could in principle be pushed past page 1 by other closed-and-recently-commented issues under the same label. Low risk at current issue volume for this single label, but the comment's justification doesn't quite hold if that volume grows — worth either sorting by a field closer to what's filtered, or noting the assumption (page-1 coverage requires the label's closed-issue count to stay under ~100 with recent churn).

Verified clean: `openByMarker` is still preferred over `closedByMarker` when both exist (an open issue for the service wins), and the auto-close lane for services that caught up is untouched — this PR only adds a matching lane, it doesn't touch close semantics.

PR #3702, opened 2026-08-04
feat(engagement-service): scaffold the ADR-0220 D3 gamification engine (slice 1)

JiRaska:
## Summary

Scaffolds **`openbank-engagement-service`**, the ADR-0220 D3 gamification engine (refs #3701). ADR-0220 D5 gates the whole engagement/gamification branch on ADR-0219's contact-policy gate (slices 1–2 in #3665/#3689). The service is reviewable and testable now — the 5 safety invariants, the domain model and the event backbone are in place — while the gate wiring waits for #3665 to merge.

## Type of change

- [x] feat (new service)

## Linked issues

Refs #3701

## Changes

**Domain (ADR-0220 D3 safety invariants — domain rules, tested):**

1. **No credit reward**: `EarnSource` enumerates the only permitted sources. Every value in the enum is non-credit-bearing; adding a new source is a visible, reviewed PR.
2. **Opt-in**: `enrolled = false` by default; leaving is one action and **never forfeits** earned points or badges.
3. **No fake urgency**: `Challenge.expiresAt` is nullable and non-null only for genuinely expiring items — invariant 3 is code, not policy.
4. **Yearly cap**: `award` enforces `earnedThisYear + points ≤ yearlyRewardCap`, hard-fail in the domain.
5. **Vulnerable exclusion**: `adverseState = true` blocks awards; the flag is set event-driven from fraud-hold / arrears / open-dispute events.

**Service files:** `EngagementProfile`, `EarnSource` catalogue, `BadgeType`, `Challenge`; `XpEarned`/`BadgeUnlocked`/`StreakUpdated` outbox events; `EngagementService` (award/streak/badge/opt-in/out); REST hub at `/api/v1/engagement`; transactional outbox (same pattern as consent-service); `EngagementBootSmokeIT` (Testcontainers Postgres, CDI wiring, Flyway V1+V2); `EngagementSafetyInvariantTest` + `EngagementServiceTest` (13 tests total); Flyway V1 (profiles) + V2 (outbox); `openapi.yaml`; `governance.yaml`; `version.txt 0.1.0`; release-please registration.

**ContactPolicyGate:** honest-zero empty-suppression-port wiring until #3665 merges — same pattern as campaign-service before #3689, documented in the producer.

## Definition of Done

- [x] Hexagonal layering (domain zero framework imports).
[...]

Issue #3585, opened 2026-08-03
campaign journeys have no branch or stop conditions (ADR-0200 D1), and in-app surfaces (ADR-0220) are a document with no code


JiRaska:
## The gap between the plan and the code

ADR-0200 **D1** defines a campaign as *"a definition (steps, delays, **branch conditions, stop
conditions**)"*. The domain implements steps and delays. Neither condition type exists —
`CampaignStep` is `(order, template, channel, variables, delaySeconds)` and `Campaign` has no
condition field at all.

ADR-0220 (in-app engagement surfaces: `home_banner`, gamification, pre-approved offers) exists as a
decision document and nothing else. `git grep home_banner` matches only the ADR. There is no
surface service, no slot registry, no impression/click/conversion feedback.

So the estate today can send a fixed sequence on two channels. Every workflow tool this is measured
against can also branch, stop early, and render in-app.

## What is genuinely buildable now, and what is not

The mechanism is cheap: `CampaignJourneyWorkflow` already carries a signal
(`consentRevoked()`, ADR-0200 D2) and each step is an activity, so evaluating a predicate before a
step is a small change. **The predicates are the problem**, and they split sharply:

**Backed by data campaign-service already holds** — buildable without inventing anything:
- stop after N sends in this campaign for this party (the send log has it);
- stop after repeated suppression (`SUPPRESSED_CONSENT` / `SUPPRESSED_CAP` are recorded per step);
- branch on the enrolment's own state (which step it reached, what the last outcome was).

**Not backed by anything** — would have to be fabricated:
- **goal reached / converted.** There is no conversion event anywhere: no `conversion`, `goalReached`
  or `converted` producer in campaign-service or crm-service. A campaign's `goal` is free text on the
  definition, not something the platform can observe.
- **clicked / opened.** No impression or click feedback exists. That is ADR-0220 D2's loop, which is
  not built.

Adding a `goalReached()` signal that nothing ever emits would give the authoring screen a condition
that can never fire — the same class of defect as the `SERVICE` principal type in `rest.rego` and the
IN_APP stub channel, both of which read as capability and delivered nothing. ADR-0220 D5 names the
rule for exactly this case: *"an inauthentic placeholder is worse than a missing feature."*

## Proposed sequence

1. **Stop conditions on observable state** (this repo, no new dependency). Definition gains an
   optional stop condition; the workflow evaluates it before each step and exits with a distinct
[...]

PR #3895, opened 2026-08-06
feat(campaign): branch a journey on what the platform can actually observe (ADR-0200 D1)

JiRaska:
Closes nothing on its own — **Refs #3585 (slice 2)**.

## What the issue asked for, and what is actually left of it

#3585 names two things. This PR does **one half of the first one** and deliberately
leaves the rest.

| Issue item | State |
|---|---|
| 1. Stop conditions on observable state | **already on main** — #3635 shipped `StopCondition` + `STOPPED_MAX_SENDS` |
| "stop after repeated suppression" | **already true and not buildable as written** — the workflow terminates on the *first* suppression (`TERMINATED_SUPPRESSED`), so there is no "repeated" to count |
| 2. A conversion signal, sourced honestly | **untouched** — the issue is right that it is an ADR, not a patch |
| 3. Branch conditions | **this PR**, restricted to what is observable today |
| 4. In-app surfaces (ADR-0220) | **untouched** — see below |

## Branch conditions

A step gains an optional `condition`, evaluated after its delay and immediately before
the send. If it does not hold the step is **skipped and the journey continues** — that
is what makes this a branch and not a second flavour of stop condition.

```
IF_PREVIOUS_CONFIRMED       run only if the earlier send was reported delivered
IF_PREVIOUS_NOT_CONFIRMED   the reminder case
```

Two values, both named after the single fact the platform can observe about an earlier
step: its ADR-0239 `DeliveryStatus`. There is **no** `IF_PREVIOUS_OPENED` and no
`IF_GOAL_REACHED` — the issue establishes that no impression, click or conversion signal
exists anywhere in the estate, and a condition nothing can ever make true is the
inauthentic placeholder ADR-0220 D5 refuses (same defect class as the unreachable
`SERVICE` principal type in `rest.rego`).

`PENDING` counts as *not confirmed*, and so does "no predecessor send at all". That is a
real property of an at-least-once feedback channel: a reminder gated on
`IF_PREVIOUS_NOT_CONFIRMED` will also fire for a send whose outcome simply has not landed
yet. Documented on the enum and in the spec rather than hidden — the mitigation is a step
delay long enough for an outcome to arrive.

A skipped step records a `SKIPPED_CONDITION` send-log row, so the console's funnel shows
"this branch was not taken" instead of a hole. It is not a suppression and does not
consume the frequency cap (which counts `SENT`).

## Replay safety — argued, and then tested

The new `previousDeliveryStatus` activity is reachable **only** when `step.condition` is
non-null. A step serialized before the field existed — in a campaign row and in the
Temporal history of a running journey — deserializes with `condition = null`, so an
in-flight execution replays its recorded `loadDefinition` result, takes the identical
path, and cannot emit a command its history does not contain. No `Workflow.getVersion`
gate is therefore needed, and no new workflow type.

That argument rests on one premise, so it is a test rather than a claim:
`JourneyDefinitionLegacyShapeTest` decodes a pre-change payload **through the same
converter the production client is built with** (`TemporalClientProducer.kotlinAwareData
Converter`) and asserts `condition` is null. A stock `DefaultDataConverter` is not
evidence here — it cannot construct a Kotlin data class at all (#2749), which this PR
rediscovered the hard way.

## What is proved, and how

`CampaignJourneyWorkflowTest` drives the **real workflow** in Temporal's test
environment, not the predicates in isolation:

- a party already at the cap gets `deliverStep` **zero** times and ends `STOPPED_MAX_SENDS`;
- the cap stops a 3-step journey between step 0 and step 1;
- a reminder is skipped when the earlier step was `CONFIRMED`, **and step 2 still runs and
  the journey completes** — a branch, not a stop;
- the same definition with `PENDING` delivers that reminder (the opposite branch off the
  same code path — without it the test above would pass against a workflow that skips
  unconditionally);
- `IF_PREVIOUS_CONFIRMED` does not hold for a first step;
- an unconditional step never calls the new activity (the replay property as an assertion).

**Falsified, not just green:** deleting the branch block from the workflow turns exactly
the two decisive branch tests red (`... is skipped when the earlier step was confirmed`,
`IF_PREVIOUS_CONFIRMED does not hold for a first step`). Run locally before the commit.

Local gate, all tasks observed to execute (not `UP-TO-DATE`):
`--rerun-tasks :openbank-campaign-service:ktlintCheck :detekt :test` → BUILD SUCCESSFUL,
`CampaignJourneyWorkflowTest` 7/7, `JourneyDefinitionLegacyShapeTest` 2/2.

## Versions / migrations

- `openapi.yaml` **1.9.0 → 1.10.0**, additive MINOR (ADR-0048). Read off live `origin/main`
  immediately before pushing, not off the branch base.
- **No Flyway migration.** Steps live in the `steps_json` text column (V2), so a new field
  on `CampaignStep` needs no DDL. `V4` remains the latest.
- No event-contract change.
- `version.txt` untouched — release-please owns it.

## What I did NOT do, and what could still be wrong

- **ADR-0220 in-app surfaces are untouched.** They are a whole surface area (slot registry,
  server-decided content, the impression/click/conversion loop), not a slice of this one,
  and #3702 is already open scaffolding the D3 gamification engine. Shipping a banner here
  would be exactly the "demo with fabricated pre-approvals" ADR-0220 D5 forbids.
- **No conversion signal.** Deliberate — it is the issue's own item 2 and needs an ADR
  deciding what "converted" means for this bank.
- **No admin-ui work.** The authoring wizard cannot yet set a condition, and the funnel will
  render `SKIPPED_CONDITION` as the raw enum (both label maps fall back to the raw value, so
  nothing breaks — it is just untranslated). A follow-up.
- **`latestDeliveryStatusBeforeStep` is not covered by a real-database test.** The query
  compiles and the ordering is argued, but only a Postgres IT would prove the HQL and the
  ordering behave as intended; the workflow tests mock the activity. This is the weakest
  point in the PR.
- **Not exercised against a real Temporal server or a live in-flight journey.** The replay
  argument is reasoned + unit-pinned at the serialization boundary, not demonstrated by
  replaying a production history.
- Noticed in passing, **not fixed**: `CampaignResource.create` still hardcodes
  `Channel.EMAIL` for every step, so the `PUSH` channel added by #3584 is unreachable from
  the REST create path. Out of scope here.


JiRaska:
⚠️ **This branch's commit is UNSIGNED and the `main-protection` ruleset requires signatures — it cannot merge as-is.**

The GPG key needs a passphrase that this session has no way to supply (gpg-agent had nothing
cached, and a passphrase is not something an agent should be handling). The fix is one command
from a terminal with the key available:

```
git -C <worktree> commit --amend --no-edit -S
git push --force-with-lease origin HEAD:refs/heads/feat/campaign-journey-branch-stop
```

Auto-merge deliberately NOT armed. Nothing else about the PR is affected.

JiRaska:
Merged `main` into this branch to clear the conflict with #3897, which landed the PUSH channel on the create endpoint an hour ago (`ee9708f2c`). I did this from a separate worktree with a plain non-force push, so nothing on your branch was rewritten — `git pull` in `/private/tmp/ob-3585-23722` will fast-forward you onto it.

**Two files conflicted, both trivially:**

- `CampaignResource.kt` — `StepRequest` auto-merged with both new fields (`channel` from #3897, `condition` from here). Only the call site conflicted, resolved to carry both: `CampaignStep(it.order, it.template, it.channel, it.variables, it.delaySeconds, it.condition)`.
- `openapi.yaml` — `info.version`.

**The version needed a re-bump, and that is the part worth reading.** This branch claimed `1.10.0`. #3897 landed first as `1.11.0`, so `1.10.0` is no longer a free number and `1.11.0` is taken — I set **`1.12.0`**. The api-contract gate classifies against your PR's base, so a number consumed by a PR that merged while yours was open is invisible to it; the conflict resolution is where it has to be caught (ADR-0048, the #481 × #524 race). Also corrected a comment I wrote on the `channel` field: it said "selectable since 1.10.0", now `1.11.0`, which is the version it actually shipped in.

**Verified after the merge** (the "clean merge that silently drops a list entry" class): `.release-please-manifest.json` is 56 entries on both sides, both test suites survived — `98 tests, 0 failures` for the service, and `CampaignRestContractIT` is 15/15 with the three PUSH tests and your journey tests all present. `detekt ktlintCheck koverVerify build` green, including `quarkusBuild`.

Note `main` moved twice more while I was building (#3898, #3899 — the latter bumped `version.txt` to `0.14.0`). This branch does not touch `version.txt`, so that resolves itself on merge; no action needed.

PR #3916, opened 2026-08-06
fix(balance): stop a not-yet-effective credit being spendable before its value date

### LINKED ISSUE #1745: balance: value-date-aware effective-booked balance + daily value-date roll (ADR-0178 Phase 2)
### Context

ADR-0178 Phase 2. balance-service is a projection of the ledger (ADR-0039). The projection
(`LedgerProjectionService.apply`) books each `AccountBookedChangedEvent` delta into `bookedAmount`
**on event receipt, ignoring `entryDate`**. `entryDate` is persisted to `ledger_projection_event`
(dedup + audit) but never defers the booking.

Consequence: a future-value-dated credit (welcome bonus, scheduled interest, forward-value transfer)
is counted in `bookedAmount` — and therefore potentially in `available` — from the moment the journal
is posted, days before the ledger recognises it (the ledger trial balance is value-dated,
`entry_date <= asOf`). A customer could see / spend a not-yet-effective credit, and the customer
balance disagrees with the GL for the whole pre-value-date window.

Phase 1 (#1747) fixed the **reconciliation** to compare on the value-date basis, so this no longer
shows as false drift — but the **read model itself** is still receipt-dated.

### Scope

- Introduce an **effective-booked** figure: booked amount counting only `entry_date <= today`.
- Daily **value-date roll**: a scheduled job promotes matured future-value-dated entries into
  effective-booked at their value date (idempotent; drives a balance event on promotion).
- Derive the customer's spendable `available` from effective-booked, not the raw running total.
- Decide the product semantics explicitly (is a future-value credit visible-but-unspendable, or
  hidden until value date?) and document in ADR-0178.

### Constraints

Money-path (openbank-balance-service): 2 approvals + threat-model refresh; delivered in reversible
phases behind a flag. PSD2 available-funds semantics change — review with the payments owner.

### Acceptance

- `bookedAmount` / `available` presented as-of today equal the ledger deposit-control as-of today by
  construction (not merely reconciled after the fact).
- A future-value-dated credit does not enter `available` before its value date.
- Reconciliation stays green across the pre-value-date window.

Refs ADR-0178, ADR-0039.

JiRaska:
Closes #1745. ADR-0178 Phase 2.

## What is wrong today for a customer

**A payment received after 16:00 Prague, or at a weekend, is spendable days before the general ledger recognises it.**

The chain, all of it on `origin/main` today:

1. `SettlementDateResolver.resolve` rolls `bookingDate` to the **next business day** whenever the request arrives at/after the 16:00 `Europe/Prague` cut-off, or lands on a non-business day. So a future booking date is the *normal* shape of an after-hours payment, not an edge case.
2. `PaymentActivitiesImpl.buildJournalRequest` passes that `bookingDate` to the ledger verbatim as `entryDate`.
3. For an incoming payment `PaymentJournalFactory.sameCurrencyLines` emits `CREDIT deposit-control(sub=target)`, and the ledger's control balance is value-dated — `PanacheJournalRepository` filters `je.entry_date <= :asOf`. So the GL does **not** hold that money until the booking date.
4. `openbank.balance.projection.enabled` is `true` (ADR-0039 Phase D-2 cutover), so `LedgerProjectionService.apply` books the delta on **event receipt**, and `Balance.applyBookedDelta` moves `bookedAmount` **and** `availableAmount` together.
5. `BalanceService.getBalance` with no `asOf` returned that raw row, and `placeHold` guarded on `availableAmount`.

So the customer sees the money and can spend it. The issue frames this as a missing feature ("introduce an effective-booked figure"); it is a **live incorrectness**, and the available/booked distinction was not stale — it was absent on the live read path.

The odd part: the service already contained the correct answer. `sumBookedDeltaAfter` is keyed on `entryDate`, and `GET /{accountId}/{currency}?asOf=<today>` already returned the value-date-correct figure. Phase 1 (#1747) built the value-date basis for **reconciliation** only. The service served the wrong one of its two answers by default.

## What shipped

- `Balance.notYetEffectiveCredit` — Σ of **strictly positive** projected deltas with `entry_date > today` — plus `effectiveBooked()` / `effectiveAvailable()`. **`withReservation` guards on the effective figure**; that is the spend prevention, not the display change.
- `BalanceService` hydrates the tail on every live read and on `placeHold`, using `AccountingClock` (ADR-0207) rather than a locally-chosen zone, so balance and ledger agree on the accounting day.
- `ValueDateRollScheduler` — daily, `suspend fun`, under `ClusterLock`.
- `V12` adds two **partial** indexes (`delta > 0`) on `ledger_projection_event`, which had none on `entry_date`. Rollback note included; index-only, reversible under load.
- `openapi.yaml` `info.version` 1.7.0 → 1.8.0 (additive).

### Two design decisions worth reviewing explicitly

**Credits only, not the net tail.** A pure value-date restatement subtracts the net future tail — which for a future-dated *debit* means **adding money back**, handing the customer funds already committed to an outbound payment. That is a risk decision, not bookkeeping. The query filters `delta > 0` and `Balance`'s `init` rejects a negative tail so it cannot be reintroduced by accident.

**Derived, never materialized.** There is deliberately no `effective_booked` column. The figure is recomputed per read, so it becomes correct on its own the moment the accounting day passes the value date, with no backfill for credits already sitting in the pipeline. The roll therefore only *announces* maturity: **a missed run delays a notification and can never leave a balance wrong.** A materialized column promoted by the job would invert that failure mode — and this is exactly the class of job that silently never runs (#2148/#2187).

## What I proved, and the RED output

`ValueDateRollIT` — real Postgres + Redpanda, seeded through the real projection path, and it **drives the real cron** (`*/2 * * * * ?` via a `QuarkusTestProfile` that re-enables the scheduler, since `%test.quarkus.scheduler.enabled` is `false` service-wide). Literal config values only — the profile loads in a different classloader.

I falsified it by restoring the pre-change semantics exactly (`withValueDateBasis` returning the balance unchanged, i.e. the tail always ZERO) and re-running:

```
ValueDateRollIT > the cover decision refuses to reserve a not-yet-effective credit, and allows the effective part() FAILED
    java.lang.AssertionError:
    Expecting code to raise a throwable.

ValueDateRollIT > a credit booked to a future date is excluded from the spendable balance, and included on its value date() FAILED
    org.opentest4j.AssertionFailedError:
    expected: 2000.00
     but was: 0

3 tests completed, 2 failed
```

"Expecting code to raise a throwable" **is the defect**: on an account holding 3 000.00 effective + 2 000.00 booked three days forward, `placeHold(3000.01)` *succeeded*. Against the fix it fails as `InsufficientFundsException: Insufficient funds: available=3000.0000 (of which 2000.0000 not yet effective), overdraftLimit=0.0000, requested=3000.01`.

The value-date boundary is asserted at `today` exactly, with no clock manipulation: one account with a credit dated `today+3` (excluded), one with the same credit dated `today` (included). All money arithmetic is exact `BigDecimal` and the assertions are on resulting balance **values** (3000.00 / 5000.00 / 2000.00), not on something having run.

`BalanceValueDateTest` covers the domain arithmetic including the overdraft interaction and the negative-tail rejection.

Local gate green: `test` (148) + `ktlintCheck` + `detekt` + `quarkusBuild`; gate groups `kotlin` (16/16), `data`, `api` — including `accounting-clock-gate`, `clock-injection-gate`, `no-runblocking-in-a-scheduled-body`, `db-migration-gate`, `schema-compat-gate`, `openapi-route-conformance`.

## Critical self-evaluation — what I did NOT verify or do

- **Could a test pass against the old code? One of the three, yes.** `the real scheduler announces a credit that matured today` stayed green under the reverted semantics — the scheduler is new, so nothing existed to break. It proves the cron dispatches and reaches a reactive path (the `HR000068` class), which is what it is for; it is *not* evidence for the value-date rule. The two tests that carry that claim are the ones shown failing above.
- **I did not run this against a real environment.** Everything above about the live customer impact is read off `origin/main` source, not observed on a running cluster. I did not query a deployed database for how many accounts currently hold a forward-dated credit, so I cannot say how many customers are affected right now — only that the code path is routine and the config gate is on.
- **`applyDebit` is unchanged** and still guards on the raw `bookedAmount`. The `POST /{accountId}/debit` endpoint therefore does not consult the value-date basis. I judged the hold to be *the* cover decision and did not want to widen a money-path change; if that endpoint is reachable as a spend path, it is a follow-up.
- **`BalanceResponse` in `openapi.yaml` was already fiction** and I did not fix it: it documents `availableBalance`/`currentBalance`/`reservedBalance`, while the resource serialises the `Balance` domain object (`availableAmount`, `bookedAmount`, …). I added `notYetEffectiveCredit` under its **real** serialised name, so the schema is now a mix. Correcting the rest is a separate, contract-affecting change.
- **The daily roll's announcement is not idempotent across days** beyond the `ClusterLock` tick guard — two runs on the same accounting day would publish twice. `BALANCE_UPDATED` with `amount = 0` is informational, so I accepted that rather than adding a promotion-ledger table (which would reintroduce the materialized state this design avoids).
- **`getBalance(asOf = …)` is untouched.** It already rewinds the whole tail; hydrating on top would subtract the credit twice. Reconciliation is likewise untouched — Phase 1 already excludes this tail on both sides, so drift behaviour is unchanged.
- **No threat-model update.** `docs/threat-models/openbank-balance-service.md` exists and this changes a cover decision; I did not judge it a new threat surface (it removes spendability, adds no input), but a money-path reviewer should confirm rather than take my word.
- **Two `detekt-baseline.xml` entries added**, matching the sibling scheduler's existing entry, rather than restructuring `BalanceRepositoryImpl` (12 functions vs a threshold of 11).
- **I did not decide the product question, on purpose.** Whether `bookedAmount` should keep *displaying* the not-yet-effective credit or hide it is left open; ADR-0178 asks for that decision and it needs the payments owner. Both candidates agree on the unspendable half, which is what this ships — so the change is safe under either outcome, but it does not close that part of the issue.

## Merge expectation

Money-path: 2 human approvals + threat-model review. **This should not merge on CI alone, and I have not attempted to bypass that.** The PSD2 available-funds semantics change wants the payments owner specifically.


JiRaska:
Picked this up after the authoring session stalled before it could report. The work itself was
complete and signed; I ran the gate rather than assume it.

```
--rerun-tasks  :test  :ktlintCheck  :detekt     BUILD SUCCESSFUL
```

all tasks observed executing (no `UP-TO-DATE`), branch pushed unchanged — no rebase, no force-push,
no content of mine added.

**What I did NOT verify:** the authoring session's own reasoning and its red-first evidence are its
claims, not mine — I checked that the tree builds and the tests pass, not that the design decision is
right. That is what review is for.

One thing worth knowing before reading this PR's checks: 55 of 71 open PRs are 76+ commits behind
`main`, and on a stale branch **both a green and a red are unreliable** — three reds today were pure
base staleness. Check `git rev-list --count origin/<branch>..origin/main` before diagnosing anything
here.

JiRaska:
**Review: 3 findings, 1 blocking.**

Claim vs. code: the claim is "stop a not-yet-effective credit being spendable". The hold path delivers that. The other spend path does not.

1. **BLOCKING — `debit()` still spends the not-yet-effective credit.** `BalanceService.debit` delegates to `movementPort.applyDebit(...)`, which runs its own funds guard in SQL against the raw `available_amount` column; nothing on that path calls `withValueDateBasis`, and `Balance.effectiveAvailable()` is never consulted. Scenario: account holds 3 000.00 effective + 2 000.00 credited Friday 18:00 with `entry_date = Monday`. `POST /{accountId}/holds` for 4 000.00 is now correctly rejected; `POST /{accountId}/debit` for 4 000.00 still succeeds and the money leaves. The PR body discloses this as a follow-up — but `debit` is a live money-out endpoint on the same resource, so as merged the invariant in the title holds for one of two doors. Either close it here or the issue cannot be called done.

2. **Event `availableAmount` now carries two different meanings.** `ValueDateRollScheduler.announceMaturities` publishes `availableAmount = effective.effectiveAvailable()` / `bookedAmount = effective.effectiveBooked()`, while `placeHold`, `releaseHold`, `credit` and `debit` in `BalanceService` all publish the raw `updated.availableAmount` on the same `BalanceEvent` fields. A consumer reconciling the stream sees the figure jump down at 00:05 and back up on the next unrelated movement — with no field distinguishing which basis it got. Either publish the effective figures everywhere or add the tail as its own field on the event.

3. **`withReservation`'s message relabels `available`.** `Balance.kt` — `"Insufficient funds: available=$spendable (of which $notYetEffectiveCredit not yet effective)"`. `available` in that string is now the effective figure, not `availableAmount`. If this message reaches an API error body (it is wrapped into `InsufficientFundsException` on the debit path), a client parsing or logging it silently changes meaning. Name it `spendable=`.

Verified clean: value-date boundary is `entry_date > :asOf` in `sumNotYetEffectiveCredit` and `= :date` in `findCreditsMaturingOn`, so `today` itself is included in spendable exactly once — no off-by-one and no double-count at the boundary. Accounting day comes from injected `AccountingClock` (bank zone), not a local `ZoneId`, so balance and ledger agree on the cut. Credits-only (`delta > 0`) filter plus the `init` require means a future-dated debit can never add money back. `getBalance(asOf = ...)` is correctly left un-hydrated, so the tail is not subtracted twice. The scheduler is a `suspend fun` under `ClusterLock` with a literal cron — no `runBlocking`/HR000068, no dual-pod double publish. V12 is index-only, partial on an IMMUTABLE predicate, `IF NOT EXISTS`, with a real rollback note. `ValueDateRollIT` drives the real cron over a real DB and was shown failing against reverted semantics, which is the falsification this class of change needs.

**No stored data is wrong.** This is a read-and-decision-time defect: `ledger_projection_event` already holds the correct dated rows and `balances` holds correct receipt-dated running totals. No backfill or data migration is needed — but funds already released early are already gone, so a query for accounts currently holding a forward-dated credit is worth running before/at deploy.

Automated agent review, not an approval. Money-path: this needs its two human approvals and a threat-model read regardless of what CI says.

JiRaska:
Review: 4 findings, 1 blocking.

**1. BLOCKING — the guard is not on every spend path: `POST /{accountId}/debit` still covers from the raw booked figure, and it has a live money-path caller.**
`BalanceResource.kt:135-139` → `BalanceService.debit` (`BalanceService.kt:190-198`) → `BalanceMovementPortImpl.applyDebit` → `Balance.applyDebit` (`Balance.kt:113-116`), which still requires `bookedAmount - amount >= overdraftFloor()`. `withValueDateBasis` is never applied on that path, so the tail is ZERO there by construction.

The PR body raises this and defers it as "*if* that endpoint is reachable as a spend path". It is, and it is not a hold-then-debit path:
- `openbank-settlement-service/.../infrastructure/client/BalanceRestClient.kt:27` declares `@Path("/{accountId}/debit")`;
- `BalanceDebitAdapter.kt:25-42` calls it with `referenceId = "settlement-debit-$settlementId"`;
- `SettlementWorkflowImpl.kt:44` invokes `activities.debitPayer(settlementId)` as the **first** activity — there is no preceding `placeHold`, so no cover decision ever consults `effectiveAvailable()`.

Concrete scenario, the same one the PR's own IT sets up: account holds 3 000.00 effective plus a 2 000.00 credit whose `entry_date` is `today+3` (the routine after-16:00/weekend shape). `placeHold(3000.01)` is now correctly refused. `POST /{accountId}/debit` with 5 000.00 — or a settlement workflow for 5 000.00 — still succeeds, `bookedAmount` goes to 0, and the customer has spent money the general ledger will not recognise for three days. The defect the PR title claims to close survives on the path that moves money directly rather than reserving it.

Since `applyDebit` runs inside `BalanceMovementPortImpl.apply`'s single transaction, the fix is a `sumNotYetEffectiveCredit` read inside that transaction (the V12 index already covers the predicate) feeding a `notYetEffectiveCredit` onto the domain object before `mutate`. If the deliberate decision is that a posted GL movement must never be refused by its projection (the `applyBookedDelta` argument in `Balance.kt:131-135`), then that argument needs to be made explicitly for `applyDebit` too and settlement's use of it audited — but as it stands the two spend paths disagree with no stated reason.

**2. Non-blocking — `BALANCE_UPDATED` now carries two different bases depending on who emitted it.**
`ValueDateRollScheduler.kt` publishes `bookedAmount = effective.effectiveBooked()` / `availableAmount = effective.effectiveAvailable()`, while `BalanceService.credit`/`debit`/`placeHold`/`releaseHold` publish the raw `updated.bookedAmount`/`availableAmount`. A downstream projection consuming the stream sees the effective figure at 00:05, then the receipt-dated figure on the next movement that day — an apparent unexplained jump on an account with a forward-dated credit. Either both should be effective, or the event should carry `notYetEffectiveCredit` as its own field and let consumers subtract.

**3. Non-blocking — the REST responses of `credit`/`debit`/`releaseHold` serialise `notYetEffectiveCredit: 0` even when the tail is non-zero.**
Only `getBalance`, `getBalances` and `placeHold` call `withValueDateBasis`. So `POST /{accountId}/credit` answers `{availableAmount: 5000, notYetEffectiveCredit: 0}` for the very account where `GET /{accountId}/CZK` answers `{availableAmount: 5000, notYetEffectiveCredit: 2000}`. A client that computes spendable funds from whichever response it happens to hold gets a different answer per endpoint, and the openapi text ("Spendable funds are `availableAmount - notYetEffectiveCredit`") is then wrong for three of them.

**4. Non-blocking — two independent construction sites for the accounting clock.**
`ClockProducer.kt` produces an `AccountingClock` CDI bean (which `ValueDateRollScheduler` injects), while `BalanceService.kt:28` builds its own via the `= AccountingClock.bank(clock)` default parameter and never consults the bean. They agree today because both are `bank(...)`, but the type exists precisely so `cutoff` can be configured — and the day it is, the read/cover path and the roll would answer different accounting days, which is the ADR-0207 defect reproduced inside one service. Injecting the produced bean costs nothing.

**Already-written data: not wrong; no backfill or data migration needed.** The figure is derived per read from `ledger_projection_event`, which already persists `entry_date` for every projected delta (V6), so every historic forward-dated credit is correctly excluded from `effectiveAvailable()` the first time the new code reads it. `bookedAmount`/`availableAmount` rows stay receipt-dated by design and are not restated. The one residual is behavioural, not data: holds **already granted** against a not-yet-effective credit before this deploys are not revisited — worth a one-off query of open holds against accounts with a `delta > 0, entry_date > today` row before/at cutover.

**Verified clean:** value-date arithmetic and zone — `AccountingClock.bank` is `Europe/Prague` with a `DAY_START` (midnight) cutoff, so `today()` is exactly the calendar date in the bank zone and compares like-for-like against ledger-produced `entry_date`; the `< cutoff` branch is unreachable at the default, so there is no DST or two-hour window hole. `sumNotYetEffectiveCredit` (`BalanceRepositoryImpl.kt`) filters `e.delta > 0 and e.entryDate > :asOf`, so negative/reversal deltas are never netted in — a future-dated debit stays fully deducted, and `Balance`'s `init` rejects a negative tail so the unsafe direction cannot be reintroduced by a caller; `BalanceValueDateTest` asserts both, including the overdraft interaction, on exact `BigDecimal` values. `placeHold`/`withReservation` correctly guard on `effectiveAvailable()`, and `getBalance(asOf=…)` is correctly left alone (it rewinds the whole tail, so hydrating on top would double-subtract). `ValueDateRollScheduler.rollDaily` is a `suspend fun` under `ClusterLock` with `ConcurrentExecution.SKIP` — no `runBlocking`, so no HR000068 — and `ValueDateRollIT` drives the **real** cron rather than calling it: `FastRollProfile` re-enables `quarkus.scheduler.enabled`, shrinks the cron to `*/2 * * * * ?`, uses literal config values only (classloader trap), and runs against real Postgres + Redpanda seeded through the real projection path. V12's two partial indexes match the two query predicates exactly — `(account_id, currency, entry_date) WHERE delta > 0` for the per-balance tail (equality columns leading, range last) and `(entry_date) WHERE delta > 0` for the roll — both `IF NOT EXISTS`, index-only, with a stated and genuinely reversible rollback note. `openapi.yaml` `info.version` 1.7.0 → 1.8.0 is correct against **live** `origin/main`, which still reads 1.7.0, and the added property is additive + `readOnly`.

This is an automated agent review, not an approval. A money-path change needs human review regardless — here specifically the payments owner on the `applyDebit` question in finding 1.

JiRaska:
Second review pass. The `debit`-path bypass already reported here reproduces — not repeating it. Three things not yet raised, one substantive.

**1. The fix does not reach any out-of-process consumer.** `effectiveAvailable()` has exactly one production caller: `Balance.withReservation`. Verified by grepping the branch — every other hit is a KDoc reference, plus `ValueDateRollScheduler.kt:131`. `BalanceResource` serialises the domain object, so `GET` still answers `availableAmount: 5000` and merely gains `notYetEffectiveCredit: 2000` as a field nobody reads. Concretely, `openbank-account-service/.../client/BalanceServiceClient.kt:99` maps `available = availableAmount` and drops the new field entirely, so account-service and everything behind it keep reporting the pre-fix spendable figure.

The invariant in the title therefore holds only for callers who go through balance-service's own `placeHold`. Either the payload exposes the effective figure, or the consumers need a coordinated change — shipping the field alone leaves the defect live downstream while looking fixed.

**2. A reversed future-dated credit can wedge an account.** `sumNotYetEffectiveCredit` filters `e.delta > 0` (`BalanceRepositoryImpl.kt:79-95`) while `bookedAmount` nets everything. Take a +2000.00 credit with `entry_date = today+3`, then a −2000.00 reversal carrying the same `entry_date` (`applyBookedDelta` takes an arbitrary signed delta, so that is the natural shape): `bookedAmount = availableAmount = 0`, tail = 2000.00, `effectiveAvailable() = −2000.00`, and with a zero overdraft limit **every hold is refused until `today+3`**. Direction is safe — money withheld, never leaked — so not blocking. But the credits-only rationale in `Balance.kt` argues against netting an *independent* future debit; it does not cover netting a reversal of the very credit being excluded. Worth either a same-`transaction_id` netting rule or an explicit "accepted, reversals are rare".

**3. Test scope.** `ValueDateRollIT` is the right shape — real Postgres, real cron, exact `BigDecimal`, red-first evidence in the body. But it drives `BalanceService` via `VertxContextSupport.subscribeAndAwait`, not the REST endpoint, so nothing covers the resource wiring or the serialised shape of the new field — which is exactly where finding 1 lives. `getBalances` hydration (`BalanceService.kt:86`) has no test. Minor: `FastRollProfile` enables the scheduler service-wide and neutralises only `openbank.reconciliation.cron`, leaving `ReconciliationFreshnessWatchdog` armed on its default cron.

**Re-derived and clean** (checked rather than taken from the body): no double-count at the boundary (`entry_date > :asOf` vs `= :date`, so `today` counts once); no hold leak (`withReservation` guards on the effective figure then decrements the raw one, and `availableAmount ≥ spendable` keeps it above `overdraftFloor()`); `getBalance(asOf=…)` correctly left un-hydrated; `rollDaily` is a `suspend fun` under `ConcurrentExecution.SKIP` + `ClusterLock`, so no `HR000068`, and the IT drives the real cron rather than calling the method — the only way to prove that. V12 is index-only, partial on an IMMUTABLE predicate, `IF NOT EXISTS`, with a reversible rollback note; the non-`CONCURRENTLY` lock on a hot projection table is disclosed in the file.

Nit: `ValueDateRollScheduler.kt`'s KDoc references `ValueDateRollSchedulerIT`; the file is `ValueDateRollIT.kt`.

JiRaska:
Pushed the coordinated fix (`d86bcb0df`). The invariant now reaches account-service, not just balance-service's own `placeHold`.

**Balance gains `effectiveAvailableAmount`, a computed property rather than a second function** — `Balance.effectiveAvailable()` is a Kotlin function, and `BalanceResource` serialises the domain object directly, so Jackson never saw it. A `val`'s getter carries the name Jackson needs; a function does not.

First attempt annotated it with `@get:JsonProperty` and the **enforced domain-purity gate rejected the import** — `Balance` is the domain layer and ADR-0002 keeps it framework-free. Caught before it shipped rather than after. No annotation needed: a plain computed property serialises fine.

**`BalanceDto` gains the field, nullable with a null default.** During a rollout this service can be talking to a provider that predates it, and defaulting to zero would report every account as unspendable — wrong in a *new* direction rather than the old one. `toBalanceView()` prefers the new field and falls back to `availableAmount` otherwise. Moved it out of `BalanceServiceClient` into a top-level `internal` function specifically so the mapping is testable without instantiating a REST client — the first version was a private class member and the test could not see it.

`openapi.yaml` documents the field and *why* it exists, version bumped 1.7.0 → 1.8.0 (additive minor, correct for a new optional response field).

**Falsified**, and the test is written to catch the specific hole this PR's neighbouring review comment found: it asserts the mapping, not presence. A `containsKey("effectiveAvailableAmount")` shaped test would have passed against the pre-fix client the moment balance-service started sending the field — same trap as the original defect, just moved one layer. Reverting the fallback to `available = availableAmount` fails exactly the one test asserting that mapping; the other two (booked stays put, an older provider without the field falls back correctly) are correctly unaffected — they assert properties this change doesn't touch.

```
openbank-account-service: 236 tests, 1 skipped (broker-gated pact), 0 failures
openbank-balance-service: 153 tests, 1 skipped (same), 0 failures
```

`test`, `ktlintCheck`, `detekt` all ran on both services — checked the task list, not just the exit code.

The two remaining findings from the second review pass are unaddressed and are yours to weigh: the reversed-future-credit wedge (direction is safe — money withheld, never leaked — so not blocking) and the `ValueDateRollIT` coverage gap around the REST/serialisation layer, which is exactly where this defect lived. Worth closing given what just happened here.

JiRaska:
Pushed (\`a75eef2cd\`) — the \`--admin\` suggestion was another false alarm. Both enforced gates that went red were correct, not flaky.

**\`scheduler-liveness-adoption\`.** \`ValueDateRollScheduler\` is a NEW \`@Scheduled\` job in this PR, and it had no ADR-0237 heartbeat. That gate graduated to enforced (#4180) after this branch was cut, so nothing on the branch could have known — same shape as the fraud-hold and savings-proposal cases from earlier today. The tick catches its own exception, correctly (a failure here costs a notification, not a balance), which means a permanently broken roll was otherwise indistinguishable from a healthy quiet one — \"0 matured\" is the normal case most days. Fixed the same way: \`recordSuccess()\` only on the completed path, registration from \`StartupEvent\`, field-injected \`DomainMetrics\`.

**\`threat-model-updated-on-trust-boundary-change\`.** \`BalanceServiceClient.kt\` now reads and forwards \`effectiveAvailableAmount\` — a real trust-boundary hunk in account-service, and the gate is right to ask about it even though the risk class here is none-new (an existing inbound field from balance-service, which is inside this service's own trust boundary already, not a new edge). Recorded in \`docs/threat-models/openbank-account-service.md\`.

One thing worth knowing if you run this gate locally: it diffs against a **commit**, not the working tree. I wrote the doc update, ran the gate, and it still failed — because the file was uncommitted. Costs nothing once you know it's there.

Verified on the merged, pushed result: \`check-scheduler-liveness.py\` clean (no \`::error\`, only the 16 pre-existing baselined warnings), \`check-threat-model-diff.py --enforce\` passes. \`test\`/\`ktlintCheck\`/\`detekt\` ran green on both services. One unrelated failure along the way — \`AccountOutboxClaimIT\`, \"Failed to start quarkus\", no reference anywhere to anything this PR touches — did not reproduce in isolation, matching this repo's documented flake where a second Gradle build on the same machine takes the port.

PR #3920, opened 2026-08-06
feat(statement): build the SUPERSEDED write path for closed-period restatement

JiRaska:
## What this is

The statement `SUPERSEDED` write path — issue #1302 item 5. Item 5 was the last unbuilt half of that item's first sentence; the byte-identity half of item 5 (`render()` re-queries live data) is **not** addressed here and is called out below.

## What happens today when a closed period must be restated

**Nothing happens — there is no path at all.** This was the question that decided the shape of the work, so it is worth being exact about the three independent layers that each said no:

1. **No entry point.** `git grep 'PeriodCloseStatus.SUPERSEDED'` returns 0 hits fleet-wide (validated against the known positive `PeriodCloseStatus.CLOSED`, which returns hits in `StatementService`, the repository, the simulation scenario and a test). The enum value, `supersedesSequence`, the ER diagram, the lifecycle diagram, both language docs, the published `openapi.yaml` and `customer-edge`'s spec all describe the mechanism. Nothing ever wrote it.
2. **The close path swallows the correction.** `StatementService.closePocket` finds the existing record and returns it through the idempotency branch — so re-running the close after the underlying data is corrected answers `200` with the **stale figures**, and looks like a successful run.
3. **The schema forbade it.** `ux_statement_period_window` (V1) is UNIQUE on `(account_id, pocket_currency, period_from, period_to)`, so a superseding close for the same window is a duplicate-key violation. ADR-0035 §D was not implementable without a migration.

So the fix is to build the path, not to correct one that writes the wrong status. The only remediation available before this PR was hand-written SQL against a legal record.

## What was built

- **`V6__statement_period_restatement.sql`** — replaces the strict window index with a partial unique index on `status <> 'SUPERSEDED'`: at most one *standing* close per window. Superseded rows are retained and stay renderable by their own legal sequence for the full retention. Rollback note in the file, including why the rollback is only possible once superseded rows are removed.
- **`StatementRestatementService`** (new use case, split out of `StatementService` — which was already at detekt's `TooManyFunctions` ceiling). It re-reads booked entries + balance-service's closing balance, reconciles **fail-closed on exactly the same terms as a first close**, and writes the superseding close plus flips the prior record **in one transaction**.
- **`POST /api/v1/statements/{accountId}/{currency}/restate`** — `ROLE_OPERATOR`/`ROLE_ADMIN` only (deliberately not `ROLE_API`: a restatement is an operator action, not something an automated caller should reach). `openapi.yaml` `info.version` 1.0.0 → **1.1.0** (additive), re-read off live `main` immediately before pushing.
- **`account.statement.period.restated.v1`** — a new event type, not a widened `period.closed`. A consumer of `period.closed` keeps its contract; anything that must react to a correction opts in.

Three behaviours that are decisions, not accidents:

| case | behaviour |
| --- | --- |
| recomputed figures identical | returns the standing record, **burns no legal sequence** |
| restatement cannot itself reconcile | `409`, standing close left **untouched** |
| window was never closed | `404` — a restatement never mints a first close |

**Query filters that are part of the fix, not incidental:** `findByPeriod` and `priorClosing` now exclude `SUPERSEDED`. Both rows share `periodTo`, so an unfiltered "latest prior close" picks between them non-deterministically — opening the next period on a balance the bank has already retracted, propagating the error forward forever. There is a dedicated IT for exactly this.

Money arithmetic is `BigDecimal` end to end. The no-op check compares with `compareTo`, never `equals`, so `1075.00` and `1075.0000` are one amount.

## Traps this had to clear

- **`persist()` on an application-assigned `@Id` is INSERT-only.** The status flip loads the row **managed** via `s.find(...)` and mutates it, so Hibernate issues an UPDATE on flush. Routing it through `persist()` would schedule an INSERT and fail with the same duplicate-key error that 500'd every consent-service lifecycle transition. Commented in place.
- **The flip happens *before* the replacement is inserted** — both rows share the window and the partial index admits only one non-superseded row.
- **`audit_entries` is append-only at the DB** (a rule that discards UPDATEs silently). This path writes **no** audit rows: statement-service has no audit writer, and the record of the correction is the retained superseded row plus the outbox event. Nothing here relies on updating an existing audit row, which is the case that would have been silently discarded.
- **A unit test that calls a resource class cannot tell a served route from an unserved one** (the `McpEndpoint` 404 defect). One IT asks the **running app** what it registered via `/q/openapi` — which Quarkus generates from the scanned JAX-RS annotations, not from the committed `openapi.yaml` (that file is outside `META-INF` and is never merged). Includes a known-positive assertion on an endpoint that has always been served.
- Migration number chosen by listing the directory sorted numerically and re-checked against live `main` immediately before pushing (V1–V5 present; V6 free).
- No `ZoneId` / date-zone work was introduced, so the `AccountingClock` gate is not engaged.

## RED first — the actual output

The IT was run against the **pre-V6 schema** (V6 moved aside, everything else in place). All three DB tests failed with the exact expected cause:

```
org.hibernate.exception.ConstraintViolationException: error executing SQL statement
  [ERROR: duplicate key value violates unique constraint "ux_statement_period_window" (23505)]
  [insert into statement_period (...) values ...]
```

That is the known-positive for the migration half. A second, independent falsification for the assertion half: mutating `supersedesSequence = standing.legalSequenceNumber` to `null` in the write path turns the unit test red (`Expecting actual not to be null`) while the other 9 stay green.

Green after: **67 tests, 0 failures**, plus `detekt`, `ktlintCheck`, `koverVerify` and `quarkusBuild` (CDI wiring — the new bean resolves).

Restatement figures are asserted, not just the status: `1000.00 + 100.00 − 25.00 = 1075.00` over 2 entries replacing a stale `1100.00` over 1, with sequence 3 → 4 and `supersedesSequence = 3`. The DB assertions read the rows back with **plain JDBC**, not through the repository or mapper, so a query-filter or mapper mistake cannot hide the persisted truth from the assertion.

## Self-evaluation — what I did NOT verify, and what this does not do

- **Could the tests pass against the old code?** The unit tests could not even compile against `main` (the use case does not exist), so their redness is not evidence of anything; the load-bearing falsification is the IT against the pre-V6 schema, which is real and quoted above. The `/q/openapi` routing test is the one I am least able to falsify cheaply — I confirmed it *passes* with the annotation present and that the path is genuinely absent from the served document's path list before, but I did not run a build with the annotation deliberately detached.
- **Periods already closed are untouched.** V6 only changes indexes; no row is rewritten and no `status` value changes at deploy time. Every existing record stays `CLOSED` and remains the standing close for its window. The new endpoint is opt-in per period. Nothing back-fills or re-examines history — a period that was closed on wrong data stays wrong until an operator restates it, and **nothing in this PR detects that condition**. Deciding *when* a restatement is owed is not built here.
- **No end-to-end HTTP test of the restate endpoint with auth.** The route's registration is proved; its authz behaviour is covered only by the existing reflection-based `StatementSecurityTest` (no endpoint is `@PermitAll`) and not by a live `@TestSecurity` call. The `@Authorize(action = "statement.restate")` action name is **new and has no rego rule**; with `AUTHZ_ENFORCE` off today that is inert, and it will need a policy entry before enforcement is flipped on. I did not add one.
- **The `render()` byte-identity half of item 5 is untouched.** `render()` still re-queries live booked entries at request time, so a re-render of a closed period is still not guaranteed byte-identical — a late-completed entry inside a closed window can still silently change a legal document. Restatement does not fix that; if anything it makes the gap sharper, since there is now a sanctioned way to change the *record* and still no way to freeze the *entries*. Worth its own issue.
- **No concurrency test** on two restatements racing the same window. The partial unique index should serialise them into one winner and one constraint violation, but I did not prove it.
- **Not deployed and not exercised against a real cluster.** No migration has been applied anywhere.
- `openbank-statement-service` is **not** in `rules.yaml: money_path_services` and has no `docs/threat-models/` entry, so the 2-approval + threat-model rule does not mechanically apply. Given this changes how a legal accounting record can be rewritten, it should get human review regardless. **This PR is not expected to self-merge, and that is the correct outcome.**

## Item 4 (FX staleness) is deliberately NOT in this PR

Measured, not assumed — see the separate issue. It is a different service pair (ledger + fx), a different failure mode, and the fix is not additive. Headline measurements:

- `FxRevaluationService`'s port is `suspend fun cnbRate(base: String): BigDecimal?` — **no date, no timestamp**. `FxRateResponse` in `FxServiceClient` deserialises only `baseCurrency/quoteCurrency/bidRate/askRate`, so fx-service's `validFrom`/`validTo` are on the wire and **discarded by ledger before any staleness check could be made**.
- The only bound is server-side and date-blind: `CNB_VALIDITY_DAYS = 3`. Within that window a Friday fixing is used silently for a Monday revaluation. Beyond it the rate is `null` and the leg is skipped — the `posted = false` shape from the ČNB-404 history.
- Idempotency key is `fx-reval-{date}`: date only, no rate identity. A fixing that arrives late for an already-revalued day **can never be incorporated** — re-running returns the original journal entry and reports `posted = true`, a success that changed nothing.
- Backfill always uses the **latest** rate, not the target date's, because `cnbRate(base)` ignores `command.date`.
- **No FX-rate age/freshness metric or alert exists anywhere.** The nearest rule, `FxRevaluationSkippedAllLegs`, matches a log line emitted by **ledger** with a `{namespace="fx"}` selector while the ledger workload deploys to namespace `ledger` — so on the committed evidence it cannot match. I verified that selector and the ledger namespace by hand; I did **not** verify the live Loki label set, so treat it as a strong signal, not proof.
- 2 tests touch `FxRevaluationService`; both stub the rate as a bare value or `null`. **Staleness is untestable at that seam** — which is exactly why adding a test there without first widening the port would be the vacuous kind.

Refs #1302

### DIFF
--- a/openbank-statement-service/src/main/kotlin/com/openbank/statement/application/port/in/StatementUseCases.kt
+++ b/openbank-statement-service/src/main/kotlin/com/openbank/statement/application/port/in/StatementUseCases.kt
@@ -24,6 +24,23 @@ interface ClosePocketUseCase {
     fun closePocketMonth(accountId: UUID, currency: String, from: LocalDate, to: LocalDate): Uni<StatementPeriod>
 }
 
+/**
+ * Restates an already-closed period after a correction to the underlying booked data (ADR-0035 §D,
+ * issue #1302 item 5).
+ *
+ * A close is **immutable**: this never edits the existing record. It re-reads the booked entries and
+ * balance-service's closing balance, reconciles fail-closed exactly as a first close does, and — if
+ * the figures actually changed — writes a **new** period close with the next legal sequence and
+ * `supersedesSequence` pointing at the record it replaces, flipping that record to
+ * [com.openbank.statement.domain.model.PeriodCloseStatus.SUPERSEDED] in the same transaction.
+ *
+ * Idempotent in the sense that matters for a legal sequence: if the recomputed figures are identical
+ * to the standing close, **no** sequence is burnt and the existing record is returned unchanged.
+ */
+interface RestatePeriodUseCase {
+    fun restatePocketPeriod(accountId: UUID, currency: String, from: LocalDate, to: LocalDate): Uni<StatementPeriod>
+}
+
 /** Renders an already-closed period on demand — nothing is stored (ADR-0035 §F.2). */
 interface RenderStatementUseCase {
     fun render(
--- a/openbank-statement-service/src/main/kotlin/com/openbank/statement/application/port/out/StatementPorts.kt
+++ b/openbank-statement-service/src/main/kotlin/com/openbank/statement/application/port/out/StatementPorts.kt
@@ -49,8 +49,29 @@ interface StatementPeriodRepository {
      */
     fun saveWithOutbox(period: StatementPeriod, event: OutboxMessage): Uni<StatementPeriod>
 
+    /**
+     * The **standing** (non-SUPERSEDED) close for a window, or null. Superseded records stay in the
+     * table for retention and remain reachable by [findBySequence]; they must never be returned here,
+     * or a restated period would look closed-with-the-old-figures to the idempotency check and to
+     * [priorClosing].
+     */
     fun findByPeriod(accountId: UUID, currency: String, from: LocalDate, to: LocalDate): Uni<StatementPeriod?>
+
+    /** Finds by legal sequence **including** superseded records — an issued statement page must stay
+     *  reproducible for its full retention (ADR-0035 §D). */
     fun findBySequence(accountId: UUID, currency: String, legalSequence: Long): Uni<StatementPeriod?>
+
+    /**
+     * Atomically supersede [supersededId] and persist [replacement] plus its [event] in ONE
+     * transaction (ADR-0035 §D). Either the whole restatement lands or none of it does — a
+     * SUPERSEDED record with no replacement would leave the pocket with no standing close, and a
+     * replacement without the status flip would violate the one-active-close-per-window index.
+     */
+    fun supersedeAndReplace(
+        supersededId: UUID,
+        replacement: StatementPeriod,
+        event: OutboxMessage,
+    ): Uni<StatementPeriod>
     fun priorClosing(accountId: UUID, currency: String, before: LocalDate): Uni<BigDecimal?>
     fun listForAccount(accountId: UUID): Uni<List<StatementPeriod>>
 
--- a/openbank-statement-service/src/main/kotlin/com/openbank/statement/application/usecase/StatementRestatementService.kt
+++ b/openbank-statement-service/src/main/kotlin/com/openbank/statement/application/usecase/StatementRestatementService.kt
@@ -1,0 +1,202 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) OpenBank contributors. Licensed under the Apache License, Version 2.0.
+// See LICENSE in the repository root or https://www.apache.org/licenses/LICENSE-2.0 for details.
+package com.openbank.statement.application.usecase
+
+import com.openbank.statement.application.port.`in`.RestatePeriodUseCase
+import com.openbank.statement.application.port.out.AccountInfoPort
+import com.openbank.statement.application.port.out.BalancePort
+import com.openbank.statement.application.port.out.BookedEntryPort
+import com.openbank.statement.application.port.out.PocketAccountInfo
+import com.openbank.statement.application.port.out.StatementOutboxMessage
+import com.openbank.statement.application.port.out.StatementPeriodRepository
+import com.openbank.statement.domain.model.CreditDebit
+import com.openbank.statement.domain.model.PeriodCloseStatus
+import com.openbank.statement.domain.model.StatementEntry
+import com.openbank.statement.domain.model.StatementPeriod
+import com.openbank.statement.domain.reconcile.ReconciliationPolicy
+import io.smallrye.mutiny.Uni
+import jakarta.enterprise.context.ApplicationScoped
+import java.math.BigDecimal
+import java.time.Instant
+import java.time.LocalDate
+import java.util.UUID
+
+/**
+ * Sum of credits minus debits over [entries] — the booked net movement (ADR-0035 §E). Shared by the
+ * close and restatement paths so a correction can never compute movement differently from the close
+ * it replaces. BigDecimal throughout: statement figures are exact, never floating point.
+ */
+internal fun netMovementOf(entries: List<StatementEntry>): BigDecimal = entries.fold(BigDecimal.ZERO) { acc, e ->
+    when (e.creditDebit) {
+        CreditDebit.CRDT -> acc.add(e.amount)
+        CreditDebit.DBIT -> acc.subtract(e.amount)
+    }
+}
+
+/**
+ * Restatement of an already-closed statement period (ADR-0035 §D, issue #1302 item 5).
+ *
+ * Split out of [StatementService] rather than folded into it: the two paths share only their
+ * reconciliation rule, and a correction is the one operation in this service that writes two rows.
+ *
+ * **What existed before this class: nothing.** `PeriodCloseStatus.SUPERSEDED` and
+ * `supersedesSequence` were present in the model, the schema, the renderers and the published spec
+ * and had **zero write sites fleet-wide** — so a period closed on wrong data was immutable, and
+ * re-running the close returned the stale record unchanged through the idempotency branch. The only
+ * remediation was hand-written SQL against a legal record.
+ */
+@ApplicationScoped
+class StatementRestatementService(
+    private val accountInfo: AccountInfoPort,
+    private val bookedEntries: BookedEntryPort,
+    private val balance: BalancePort,
+    private val periods: StatementPeriodRepository,
+) : RestatePeriodUseCase {
+
+    /** Clock seam, as in [StatementService]: `closedAt` is stamped once and stored, so renders of the
+     *  superseding page stay deterministic. Overridable in tests; CDI uses the default. */
+    internal var clock: () -> Instant = Instant::now
+
+    /** Opening balance = the prior close's closing, else balance-service's balance the day before.
+     *  `priorClosing` excludes SUPERSEDED rows, so a restated earlier period cannot re-open a later
+     *  one on a figure the bank has already retracted. */
+    private fun openingBalance(accountId: UUID, currency: String, from: LocalDate): Uni<BigDecimal> =
+        periods.priorClosing(accountId, currency, from).flatMap { prior ->
+            if (prior != null) {
+                Uni.createFrom().item(prior)
+            } else {
+                balance.closingBalance(accountId, currency, from.minusDays(1)).map { it.amount }
+            }
+        }
+
+    override fun restatePocketPeriod(
+        accountId: UUID,
+        currency: String,
+        from: LocalDate,
+        to: LocalDate,
+    ): Uni<StatementPeriod> = accountInfo.pocketAccount(accountId).flatMap { account ->
+        periods.findByPeriod(accountId, currency, from, to).flatMap { standing ->
+            if (standing == null) {
+                Uni.createFrom().failure(NoClosedPeriodToRestateException(accountId, currency, from, to))
+            } else {
+                recomputeAndSupersede(account, currency, from, to, standing)
+            }
+        }
+    }
+
+    private fun recomputeAndSupersede(
+        account: PocketAccountInfo,
+        currency: String,
+        from: LocalDate,
+        to: LocalDate,
+        standing: StatementPeriod,
+    ): Uni<StatementPeriod> = bookedEntries.bookedEntries(account.accountId, currency, from, to).flatMap { entries ->
+        openingBalance(account.accountId, currency, from).flatMap { opening ->
+            balance.closingBalance(account.accountId, currency, to).flatMap { reported ->
+                // Fail closed on exactly the same terms as a first close (ADR-0035 §E): a
+                // restatement that cannot itself reconcile must not replace a record that does.
+                when (val r = ReconciliationPolicy.reconcile(opening, netMovementOf(entries), reported.amount)) {
+                    is ReconciliationPolicy.Result.Mismatch ->
+                        Uni.createFrom().failure(
+                            ReconciliationException(account.accountId, currency, from, to, r),
+                        )
+                    is ReconciliationPolicy.Result.Reconciled ->
+                        if (unchanged(standing, opening, r.closingBalance, entries.size)) {
+                            // Nothing to correct — do not burn a legal sequence on a no-op.
+                            Uni.createFrom().item(standing)
+                        } else {
+                            supersede(account, currency, from, to, standing, entries.size, opening, r.closingBalance)
+                        }
+                }
+            }
+        }
+    }
+
+    /** BigDecimal identity by VALUE (`compareTo`), never `equals` — 1075.00 and 1075.0000 are the
+     *  same amount and only `compareTo` says so. */
+    private fun unchanged(
+        standing: StatementPeriod,
+        opening: BigDecimal,
+        closing: BigDecimal,
+        entryCount: Int,
+    ): Boolean = standing.openingBalance.compareTo(opening) == 0 &&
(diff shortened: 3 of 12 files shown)


JiRaska:
Picked this up after the authoring session stalled before it could report. The work itself was
complete and signed; I ran the gate rather than assume it.

```
--rerun-tasks  :test  :ktlintCheck  :detekt     BUILD SUCCESSFUL
```

all tasks observed executing (no `UP-TO-DATE`), branch pushed unchanged — no rebase, no force-push,
no content of mine added.

**What I did NOT verify:** the authoring session's own reasoning and its red-first evidence are its
claims, not mine — I checked that the tree builds and the tests pass, not that the design decision is
right. That is what review is for.

One thing worth knowing before reading this PR's checks: 55 of 71 open PRs are 76+ commits behind
`main`, and on a stale branch **both a green and a red are unreliable** — three reds today were pure
base staleness. Check `git rev-list --count origin/<branch>..origin/main` before diagnosing anything
here.

JiRaska:
**Review: 4 findings, 1 blocking.**

Claim vs. code: the write path does what the description says — the original row is preserved, the replacement carries the next sequence with `supersedesSequence`, and both land in one transaction. The gap is what happens to the periods *after* the one being restated.

1. **BLOCKING — restating a closed period silently breaks the chain of every period already closed after it.** `StatementRestatementService.restatePocketPeriod` recomputes and supersedes exactly one window and stops. Scenario: January closes at 1 100.00 (seq 1). February is then closed — its `openingBalance` is read from `priorClosing` and stored as 1 100.00 (seq 2), reconciled against balance-service, all valid. A correction now restates January to 1 075.00 (seq 3, supersedes 1). February's *stored* record still says `openingBalance = 1100.00` while `priorClosing` for February now answers 1 075.00: the issued February page no longer chains to the standing January page, by exactly the 25.00 that was corrected. Nothing in the diff detects, blocks, or cascades this, and the fail-closed reconciliation cannot catch it — it only ever validates the window being restated. `StatementRestatementIT.the next period opens on the corrected closing balance` looks like it covers this, but it asserts `priorClosing` with **no February row present**, i.e. the one case where the problem cannot occur. Either refuse to restate a window that has a later standing close (a 409 the operator can act on), or cascade. Given these are legally sequenced statement pages under 10y retention, silently leaving an inconsistent chain is the worst of the three options.

2. **`POST .../restate` with a missing `from` or `to` is a 500.** `StatementResource.restate` is a plain (non-`suspend`) `fun` taking `from: String` / `to: String` as non-nullable `@QueryParam`s, so Kotlin emits `checkNotNullParameter` at bytecode offset 0. JAX-RS does not enforce `required: true` on a query parameter — it passes null — so `POST /api/v1/statements/{id}/CZK/restate` with no query string NPEs before the method body and answers 500. `required: true` in the new `openapi.yaml` block makes the fuzz lane generate exactly that call. Declare them `String?` and `require(...)`, or take `LocalDate?` and reject explicitly. (An unparseable-but-present date is fine — `DateTimeParseException` is mapped to 400 by libs.)

3. **No test drives the new route over real HTTP.** `StatementRestatementIT` calls `StatementPeriodRepositoryImpl` directly and separately asserts the path string appears in `/q/openapi`; `StatementServiceTest` calls `StatementRestatementService` against a mocked repository. So the composition actually exercised in production — REST → `RestatePeriodUseCase` → `supersedeAndReplace` — is never run end to end, and finding 2 above is precisely what that gap hides. The `/q/openapi` check is a good instinct and proves registration, but a registered route is not a working one; one RestAssured `POST` with `@TestSecurity(roles=["ROLE_OPERATOR"])` closes both.

4. **Concurrent restatements answer 500, not 409.** `nextLegalSequence` is `MAX+1` under no lock and the endpoint has no idempotency key. Two operators (or one impatient retry against a slow first call) both read seq 3, both mint seq 4, and the second flush violates `ux_statement_period_window_active` — a `duplicate key` surfacing as an unmapped 500. The *sequential* retry case is genuinely idempotent (`unchanged` returns the standing record without burning a sequence), which is the important half; this is the narrower race. Map the constraint violation to 409.

Verified clean: the superseded record is never mutated beyond its `status` — figures, id, `closedAt` and legal sequence are untouched, confirmed by JDBC read-back rather than through the repository, so a mapper or query-filter mistake cannot hide it. `supersedeAndReplace` loads the prior row **managed** via `s.find(...)` and mutates it, rather than round-tripping an application-assigned id through `persist()` — the consent-service duplicate-key defect is correctly avoided, and the IT would catch it if it regressed. The status flip is flushed *before* the replacement is inserted, which is required by the partial unique index. Replacement row + outbox message commit in one `withTransaction`, so no lost-event window. `findByPeriod` and `priorClosing` both gained `status <> SUPERSEDED` while `findBySequence` deliberately did not, so an issued page stays reproducible — all three asserted. `unchanged` compares `BigDecimal` by `compareTo`, so 1075.00 vs 1075.0000 does not burn a sequence. `nextLegalSequence` is `MAX` over all rows including superseded ones, so a sequence number is never reused. A new event type rather than a widened `period.closed`. V6 narrows the unique index to the invariant that actually holds, with a rollback note that correctly states the precondition. Reconciliation fails closed and leaves the standing record untouched, asserted. `openapi.yaml` 1.0.0 → 1.1.0, additive, with both docs translations updated.

**No stored data is wrong today** — `SUPERSEDED` had zero write sites before this, so nothing has ever been restated and no row is corrupt; no migration beyond V6 (which is index-only) is needed. But note finding 1 makes this a forward-looking data-integrity question: the *first* production restatement of a period that has successors is what creates the inconsistent chain, so it should be answered before the endpoint is exposed rather than after.

Automated agent review, not an approval. Money-path: human review required regardless, and finding 1 in particular is a decision for whoever owns the ADR-0035 legal-sequence semantics, not for CI.

JiRaska:
Review: 5 findings, 2 blocking

The core mechanism is right and, unusually, is proven rather than asserted: `supersedeAndReplace` really is one transaction, the prior row really is UPDATEd rather than re-INSERTed, and `StatementRestatementIT` reads the rows back over plain JDBC instead of trusting the mapper. Both blocking findings are about what happens to the record *around* the restatement, not the write itself.

### Blocking

1. **A superseded page does not stay reproducible — `StatementService.render` re-reads booked entries LIVE.** `render(accountId, currency, legalSequence, format)` resolves the stored `StatementPeriod` by sequence (correct — `findBySequence` has no status filter), then calls `bookedEntries.bookedEntries(accountId, currency, period.periodFrom, period.periodTo)` and renders the header from the **stored** figures over the **current** entries.
   Concrete: January closes at sequence 1 with `closingBalance 1100.00, entryCount 1`. The missing 25.00 fee is booked. Restate → sequence 2 says `1075.00 / 2 entries`, sequence 1 flips to SUPERSEDED. Now `GET` the statement at legal sequence 1: the camt.053/MT940 header says closing 1100.00 over one entry, while the entry list rendered underneath is the corrected two entries summing to 1075.00. An internally inconsistent legal page, handed to a customer or an auditor.
   That directly contradicts what this PR promises in three places: `V6__statement_period_restatement.sql` ("remain renderable by legal sequence, so an already-issued statement can still be reproduced **byte-for-byte**"), and both `03-api.en.md` / `03-api.cs.md` ("stays renderable by its own legal sequence for its full retention period"). Before this PR nothing could make booked entries diverge from a stored close, so the claim was vacuously true; restatement is precisely the event that breaks it. Either snapshot the entries (or the rendered artifact) on the superseded record, or retract the reproducibility claim from the migration and both docs pages.

2. **Restating a period leaves every later close's `openingBalance` stale, with nothing to detect it.** `openingBalance` is resolved from `priorClosing` **only at close time** and then stored; there is no cascade and no re-reconciliation of an already-closed later period.
   Concrete: January closes 1100.00 (wrong). February closes with `openingBalance 1100.00` and its own reconciled closing. Restate January → 1075.00. February's stored record still opens on 1075 + 25 = a figure the bank has just retracted, so the chain has a 25.00 discontinuity that no query ever revisits — and `POST /{accountId}/{currency}/restate` accepts any past window, so this is reachable on the first real use of the feature, not an edge case. The `openingBalance` KDoc in `StatementRestatementService` claims `priorClosing` excluding SUPERSEDED means "a restated earlier period cannot re-open a later one on a figure the bank has already retracted" — that holds for a *future* close, not for one already written, which is the case that matters. Either refuse to restate a window that has a later standing close (cheapest, and matches the fail-closed posture everywhere else here), or cascade, or document the operator obligation to restate the tail explicitly.

### Non-blocking

3. **Two concurrent restates surface as a 500, not a 409.** `nextLegalSequence` runs in its own `@WithSession` before the write transaction, so both callers read standing seq 3 and both compute 4; the second `supersedeAndReplace` violates `ux_statement_period_legal_seq` and the raw `ConstraintViolationException` escapes the two `onFailure` recoveries in `StatementResource.restate`. The *outcome* is correct (fails closed, no double supersession), only the status code is wrong.
4. **One operator can retract an issued legal statement.** `@RolesAllowed("ROLE_OPERATOR","ROLE_ADMIN")` + `@Authorize(action = "statement.restate")` — statement-service is not in `money_path_services`, so the four-eyes verb clause can never derive a `statement` scope and nothing gates this. `rules.yaml: four_eyes.actions` already carries the exact precedent: `party.merge`, added because it is "destructive OF IDENTITY and irreversible in practice … it carried only @RolesAllowed(ROLE_OPERATOR, ROLE_ADMIN) + an audit event, which records who did it but stops nobody". Retracting a sequenced statement page is the same shape. Suggest adding `statement.restate` there — after running that block's own guardrail (confirm no M2M caller, which looks true here since the docs deliberately exclude `ROLE_API`).
5. **No audit event for the retraction itself.** `account.statement.period.restated.v1` is an outbox/business event carrying the figures; it does not record *who* authorised the restatement. For an operator action that voids a legal record, the actor belongs in an audit envelope.

### Verified clean

**V6 does exactly what it claims.** `ux_statement_period_window` in `V1__init_statement.sql:27` is a plain `CREATE UNIQUE INDEX`, not a table constraint, so `DROP INDEX IF EXISTS` succeeds rather than erroring on a dependent constraint; `status VARCHAR(16) NOT NULL DEFAULT 'CLOSED'` means the partial predicate `status <> 'SUPERSEDED'` is never NULL, so no row silently escapes the index. It therefore permits {SUPERSEDED, CLOSED} for one window and still forbids two standing closes. Rollback note present and honest about needing the SUPERSEDED rows deleted first.

**The idempotency lookup is not broken by it.** `findByPeriod` gained `AND status <> :sup` in the same commit (and keeps `setMaxResults(1)`), so it cannot match two rows; `priorClosing` gained the same filter; `StatementRestatementIT` asserts both — the standing close resolves to sequence 2 / 1075.00, and February opens on 1075.00. `latestClosedPeriodTo` already filtered on `status = CLOSED`.

**One transaction, and provably so.** `supersedeAndReplace` loads the prior row MANAGED via `s.find`, mutates `status`, flushes, then persists the replacement and the outbox row — all inside a single `sf.withTransaction`. `StatementPeriodEntity.id` is application-assigned, and the code deliberately does *not* round-trip it through `persist()`, so the consent-service INSERT-only duplicate-key trap is avoided. The IT reads both rows back with `DriverManager`/plain SQL rather than through the repository, which is the only thing that could show this — a mocked repo cannot, and the KDoc correctly names removing V6 as its known-positive.

**Chain semantics.** `supersedesSequence = standing.legalSequenceNumber`, `nextLegalSequence` is `MAX(legalSequenceNumber)+1` over **all** rows including superseded ones, so sequences stay monotonic and unique; re-superseding works correctly, since after a restatement the standing record *is* the replacement and a second call supersedes that one, chaining 1←2←3. `unchanged()` compares BigDecimal by `compareTo`, not `equals`, so 1075.00 vs 1075.0000 burns no sequence. Fail-closed reconciliation before any write, and a new event type rather than a silently-widened `period.closed`. `netMovementOf` is a pure extraction.

`info.version` re-read off live `origin/main` this turn: 1.0.0 → 1.1.0, correct and unclaimed. Both docs pages and the openapi entry agree with the code on roles, 404/409 semantics, and the no-op-burns-no-sequence behaviour. The `/q/openapi` route test is a good catch for the annotation-binding class of defect and carries its own known-positive.

**No already-written data is wrong.** V6 replaces an index and rewrites no row; every existing close stays valid. The exposure in finding 2 is about rows a *future* restatement would leave stale, not about anything in the table today.

---
Automated agent review — not an approval. This writes and retracts sequenced legal statement records under CNB retention, so it warrants human review regardless of what this comment says.

PR #3923, opened 2026-08-06
fix(fraud): a wrongly-typed currency answered 500, not 400

JiRaska:
Refs #3039.

## What this fixes

`POST /api/v1/fraud/score` answered **500** for a wrongly-typed `currency`.

Nothing validated `currency`, and Jackson **coerces** a JSON scalar of the wrong type into a `String` rather than rejecting it: `{"currency": false}` deserialises to the five-character `"false"`. That reached `fraud_scores.currency`, a `varchar(3)`; Postgres raised `22001 value too long`, Hibernate surfaced it as a `DataException`, and with no mapper for that type `GenericExceptionMapper` answered 500.

**RED first.** `FraudScoreMalformedBodyIT` sends the fuzzer's exact reproducer body. Before the guard it failed with:

```
org.hibernate.exception.DataException: error executing SQL statement
  [ERROR: value too long for type character varying(3) (22001)]
  [insert into fraud_scores (...) values (...)]
FraudScoreMalformedBodyIT > a wrongly-typed currency field is a client error... FAILED
```

After the guard the whole `:openbank-fraud-service:test` suite is `BUILD SUCCESSFUL`, including a **control** test asserting a well-formed body still scores 200 (so the guard is narrow, and the tests cannot pass against the old code — the RED output above is the evidence).

The guard is `require` -> `IllegalArgumentException` -> 400 via `openbank-libs-runtime`'s `CommonExceptionMappers`. Deliberately **not** a service-local `ExceptionMapper`: libs owns the mappers for JDK types and two mappers for one type are picked non-deterministically per request. No `ApiError` is constructed anywhere in this diff, so the #3880 `timestamp`-required change does not overlap.

**API contract.** `openapi.yaml` gains a documented `400` (additive) and `info.version` goes 1.1.0 -> 1.2.0, read off **live main** immediately before push. The request schema is **not** narrowed with a `pattern` — that is a breaking oasdiff change needing a major bump and a `/api/v2` URL. Which side was wrong: the **validation** side. The spec was silent, not incorrect.

## Re-measurement of run 30804842325 — the issue title is wrong, but so was the correction

Measured via the attempt-scoped endpoint (`/attempts/1/jobs`). **18 jobs, 12 failure / 6 success.** Every one of the 12 fails at the same step, `Fuzz authenticated surface`.

Note on method: `gh api .../jobs/<id>/logs` returned an **empty body for all 12** jobs. I only noticed via `wc -l`; the classification below comes from the run-level log archive (`/attempts/1/logs`), and every `failed to boot` line quoted was filtered through `grep -v '\${'` to exclude the echoed `run:` script.

**5 of 12 still fail to boot** — so the premise correction ("none fails to boot") does not hold beyond the four sampled. These are **harness gaps, not service defects**: the fuzz job provides neither a Temporal server nor, for settlement, a Postgres.

| Service | Class | Evidence |
|---|---|---|
| openbank-lending-service | **boot** | `UnknownHostException: temporal-frontend.temporal.svc.cluster.local` |
| openbank-domestic-payment | **boot** | same Temporal cluster DNS name |
| openbank-transaction-service | **boot** | `Connection refused: localhost:7233` (Temporal) |
| openbank-sepa-payment | **boot** | `Connection refused: localhost:7233` (Temporal) |
| openbank-settlement-service | **boot** | Flyway: `Connection to localhost:5432 refused` |
| openbank-delegation-service | 500 x4 | missing non-null `granteePartyId`/`scaSessionId` — **already fixed by open #3625** |
| openbank-interest-service | 500 x2 | `capitalize` missing non-null `productId` — **#3625 class**; plus empty `productId` on `effective-rate` |
| openbank-fx-service | 500 x1 | `/fx/convert` — fx is in **#3625**'s file list |
| **openbank-fraud-service** | 500 x1 | **this PR** |
| openbank-ledger-service | 500 x2 | `POST /journals` with `lines: [null, ...]`; `fx-revaluation?date=` — **not** the date parse (that path already guards `isNotBlank()`) |
| openbank-swift-service | 500 x1 | `POST /api/v1/swift` with non-ASCII `beneficiaryAccount` — **uncovered, no PR** |
| openbank-balance-service | 500 x3 | `reconciliation?asOf=`, `/credit` with `currency: ""`, `/initialize` with `currency: false` |

**Schema-validation mismatch is not a failure mode.** 9 jobs report it, including `openbank-account-service` which **passed** with 7 such operations. Ledger's 3 are advisory. Only `Server error` and boot failure make a job red.

## A hypothesis I formed and then falsified

I suspected the `DateTimeExceptionMapper` (landed for #3038, and already present at the run's head SHA) did not fire for `suspend` resource methods — balance's `run()` is `suspend` and still 500'd on `?asOf=`, while interest's `Uni`-returning date paths did not.

**Wrong.** A throwaway IT showed `?asOf=not-a-date` -> **400** on the `suspend` method: the mapper works. `?asOf=` (empty) -> 500 for a different reason — the empty string does not reach `LocalDate.parse`, so reconcile proceeds and the downstream ledger call fails. That test is not in this PR because it measured the environment, not a defect.

## Self-evaluation — what I did NOT verify

- **I re-measured all 12** for boot-vs-500, and that classification is solid (quoted, substitution-checked log lines).
- **Only fraud-service was verified by running code.** The other eleven rows are read off the run log plus source inspection. In particular I did **not** empirically confirm that fx, interest and delegation's findings are fully covered by #3625 — I confirmed delegation's by reading #3625's diff (it changes exactly those three params to nullable + `requireNotNull`), and inferred fx/interest from those services being in its file list. Treat those two as *likely* covered, not proven.
- **Balance's 3 and ledger's 2 are unattributed.** Balance's `?asOf=` 500 in my local IT was caused by an unavailable ledger, which may or may not be what the CI harness hit. I did not reproduce ledger's or swift's.
- **The run is 3 days old and 160 commits behind current main.** Some findings may already be gone; I only checked that for the `DateTimeExceptionMapper` (present at run time) and for commits touching the seven services since (one, unrelated).
- I did not run the full repo gate, only `:openbank-fraud-service:` `test`, `ktlintCheck`, `detekt` (all green). No `quarkusBuild`.
- I did not update `docs/threat-models/openbank-fraud-service.md`; this change removes an availability defect and adds no new trust boundary, but the threat-model-diff gate may disagree.

## Merge expectation

fraud-service is money-path: **2 human approvals required, and this PR will not merge without them. That is the correct final state** — no `--admin`.


JiRaska:
**Review: 3 findings, none blocking.**

Claim vs. code: the diff does fix the reported reproducer, and by the right mechanism. The findings below are the same defect class one field over.

1. **`rail` is guarded for blankness but not for length, and `rail` is `varchar(40)`.** `FraudResource.kt` — `require(rail.isNotBlank())`. `V1__create_fraud_scores.sql:9` declares `rail VARCHAR(40) NOT NULL`. Scenario: `{"amount":"1250.00","currency":"CZK","rail":"<41 chars>"}` → passes the guard → `22001 value too long for type character varying(40)` → `DataException` → **500**, exactly the failure this PR exists to remove. The KDoc states the principle correctly ("the bound is the *storage* contract, so it cannot silently drift back") and then applies it to only one of the two string columns it touches. `require(rail.length in 1..40)`.

2. **`amount` is unvalidated against `NUMERIC(20,4)`.** Same file, same function. `{"amount":"1e20","currency":"CZK","rail":"SEPA"}` deserialises fine into `BigDecimal`, then `22003 numeric field overflow` → 500. Also unbounded scale: `"1.000000001"` is 9 decimal places into a `,4` column (Postgres rounds rather than erroring, so that half is cosmetic, but the precision overflow is not). schemathesis generates large numerics routinely, so the fuzz lane is likely to re-open on this one.

3. **Empty body is untested and is the case where a `suspend fun` bites.** `score` is `suspend`, so Kotlin emits no `checkNotNullParameter` for `req`; a `POST` with `Content-Type: application/json` and a zero-length body relies entirely on RESTEasy Reactive/Jackson rejecting it before dispatch. If it does not, `req.validate()` NPEs → 500 — pre-existing (`req.toDomain()` had the same exposure), not introduced here, but `FraudScoreMalformedBodyIT` has four cases and none is the empty body. One more `body("")` case settles it either way for the cost of six lines.

Verified clean: the guard is `require` → `IllegalArgumentException` → `openbank-libs-runtime`'s `IllegalArgumentExceptionMapper` → 400 with the standard `ApiError` envelope (`traceId`/`status`/`code=VALIDATION_ERROR`/`message`) — no second envelope shape introduced into this resource, and no service-local mapper competing with the libs one. Tests drive **real HTTP** (`@QuarkusTest` + RestAssured + `@TestSecurity` against a real Postgres/Redis resource), including a control asserting a well-formed body still returns 200, so the guard is shown to be narrow rather than merely present. Both the type-coercion case (`"currency": false` → `"false"`) and the plain over-long case (`"CZKK"`) are covered, so a fix that only special-cased the boolean would go red. The new top-level `validate()`/`ISO_4217` declarations sit between two plain `data class`es with no annotation above them, so there is no repeat of the `@Path`-binds-to-the-next-declaration trap. `openapi.yaml` `info.version` 1.1.0 → 1.2.0 with a `400` response added — additive, correct minor.

One note rather than a finding: describing the ISO-4217 constraint as prose instead of `pattern:` to avoid a breaking oasdiff classification is a defensible call, but it does mean the spec no longer machine-describes the enforced request contract, and the gate cannot see that the accepted input set genuinely narrowed (lowercase `czk` was accepted before this and is a 400 after). Low risk while the endpoint is inert per its own KDoc; worth stating in the release note.

**No stored data is wrong.** The over-long value never committed — Postgres rejected the INSERT — so `fraud_scores` holds no corrupted rows and no migration is needed. Pure future-behaviour fix.

Automated agent review, not an approval. Money-path: human review required regardless.

JiRaska:
Review: 4 findings, 1 blocking

1. **BLOCKING** — `FraudResource.kt:92` (`validate()`): `rail` is guarded for blankness only, but `fraud_scores.rail` is `VARCHAR(40)` (`V1__create_fraud_scores.sql`). `{"amount":"1250.00","currency":"CZK","rail":"AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"}` (41 chars) is non-blank, passes the guard, reaches the same INSERT, and Postgres raises the same `22001 value too long` -> `DataException` -> `GenericExceptionMapper` -> **500**. That is byte-for-byte the defect this PR exists to close, one field over, on the same endpoint — and the PR's own second test (`an over-long currency…`) was added precisely because a fix that only handles the type-coercion case is incomplete. A `require(rail.length <= 40)` (or an enum/allow-list, since the rail vocabulary is closed: SEPA, SEPA_INSTANT, DOMESTIC, FX) plus a fourth test closes it.
2. `FraudResource.kt:63` (`amount: BigDecimal`) -> `amount NUMERIC(20,4)`: `{"amount":1e21,"currency":"CZK","rail":"SEPA"}` is a valid JSON number, deserialises fine, and fails at flush with `22003 numeric field overflow` — again 500. Same class as (1); a `require(amount.precision() - amount.scale() <= 16 && amount > ZERO)` would bound it. Not marked blocking only because the fuzz reproducer did not reach it; it is reachable by any client.
3. Non-blocking, adjacent: `rail` is also a Prometheus label (`FraudMetricsAdapter.kt:43`, `.tag("rail", rail)`). With only a non-blank check, an authenticated caller sending random rails mints unbounded label cardinality on `openbank_fraud_scores_total`. A closed allow-list fixes (1) and this together.
4. Non-blocking, behaviour change to state explicitly: `^[A-Z]{3}$` now rejects a lowercase `"czk"`, which previously stored fine in `varchar(3)` and returned 200. No live regression — ADR-0084 Phase 1 is inert and no payment surface calls `/fraud/score` yet — but it is a real narrowing, and worth a line in the PR body rather than being discovered by the first caller.

Verified clean: the guard is the FIRST statement of `score()`, before `scoreFraud.score(...)` and therefore before any persistence, so no partial row is written on rejection; `IllegalArgumentException` -> 400 is real for this service — `openbank-libs-runtime`'s `IllegalArgumentExceptionMapper` returns 400 with the `ApiError` envelope, fraud-service registers no `ExceptionMapper` of its own at all (no #526-class non-deterministic collision), and the IT asserts the status over real HTTP rather than trusting the mapper; the openapi change is additive (a new `400` response + prose, no request-side `pattern`, so oasdiff stays non-breaking/minor) and `info.version` on **live `origin/main`** is still `1.1.0`, so `1.2.0` is free and no other open PR touches the fraud spec (checked #3742/#3700/#3618); no already-written data is wrong — the failing INSERT rolled back, so no malformed `fraud_scores` row exists and no migration is needed (rows with a lowercase currency, if any, remain valid under the column).

Automated agent review, not an approval. fraud-service is money-path — this needs human review and 2 approvals regardless.

JiRaska:
Verified against the diff rather than the description. `req.validate()` uses `require` (→ `IllegalArgumentException` → 400 via `openbank-libs-runtime`'s `CommonExceptionMappers`) rather than a service-local `ExceptionMapper` — the right call per this repo's own rule that a second mapper for one exception type resolves non-deterministically per request.

The ISO 4217 bound (`^[A-Z]{3}$`) matches the actual storage contract (`varchar(3)`), so this closes the failure mode rather than just widening it — a looser regex would still let through a value the column rejects.

Not merging — money-path service, human review by rule.

Issue #3914, opened 2026-08-06
Seven domain-event producers send no event time — audit rows record ingest time as business time


JiRaska:
## Summary

#3907 (closing #3883) taught `AuditConsumer` to record **whether** `occurred_at` is the producer's
event time (`EVENT`) or the consumer's ingest time (`INGEST`), and added the counter
`openbank.audit.event.time.missing{source_service}`. It deliberately did **not** fix the producers.

Seven of the 21 topics `audit-service` subscribes to still send no usable event time, so every audit
row for those topics records **ingest time as business time** — the GDPR Art. 30 "when" dimension and
DORA Art. 17 evidence. Under consumer lag or a replay that value is arbitrarily wrong.

## Re-measured producer list

Method — the path, not the field name (the mistake that cost #3883 a wrong conclusion): for each
topic, locate the actual outbox-row writer, confirm the publisher emits `entry.payload` verbatim
(`emitter.sendMessage(Message.of(entry.payload))`), confirm the topic is in `audit-events-in`'s
`topics:` list, and confirm `openbank.outbox.dispatch-enabled: true`. All seven pass all four checks —
these payloads really do reach `AuditConsumer`.

| service | writer | payloads affected | today |
|---|---|---|---|
| clearing-service | `ClearingEventPublisherImpl` | 2 (`batch.settled`, `item.cleared`) | no event time |
| dispute-service | `DisputeService`, `ComplaintService` | 3 builders (`dispute.resolved`, `dispute.remediation_requested`, shared complaint payload) | no event time |
| statement-service | `StatementService`, `CloseOrchestrator` | 2 (`period.closed.v1`, `period.close_failed.v1`) | no event time |
| sanctions-service | `SanctionsRepositoryImpl.saveWithEvent` / `updateWithEvent` | 2 (payload is the serialised `SanctionsCheck`) | no event time |
| lending-service | `LendingService` | 6 of 10 (`loan.disbursed`, `loan.interest_accrued`, `loan.written_off`, `loan.rescheduled`, `loan.stage_changed`, `loan.provisioned`) | no event time |
| sepa-payment | `SepaPaymentActivitiesImpl` (Temporal path) | 5 hand-built payloads | no event time |
| document-service | `DocumentRenderService`, `SignatureCeremonyService` | 2 | names it `at` |

### How this differs from the survey in #3907's PR body

- **lending: "6 of 8" — the 6 is right, the denominator is not.** There are **10** outbox payload
[...]

Issue #3883, opened 2026-08-06
audit-service: AuditConsumer reads occurredAt, the canonical envelope names it timestamp — event time is dropped at ingest


JiRaska:
## Summary

The canonical audit envelope names its time field `timestamp`; `openbank-audit-service`'s `AuditConsumer` reads `occurredAt`. Nothing bridges the two, so the envelope's own notion of *when the operation happened* is discarded at ingest and replaced with ingest time.

Found while fixing #3882 (`AuditEvent.timestamp` defaulted to `Instant.EPOCH`). Not fixed there because the choice of fix is a contract decision, not a defaulting bug.

## Evidence

`openbank-libs-domain/src/main/kotlin/com/openbank/libs/audit/AuditEvent.kt` — the field the KDoc maps to the GDPR Art. 30 "when" dimension:

```kotlin
val timestamp: Instant = Instant.now(),
```

`openbank-audit-service/src/main/kotlin/com/openbank/audit/application/AuditConsumer.kt` — what actually lands in the append-only store:

```kotlin
occurredAt = node["occurredAt"]?.asText()?.let { Instant.parse(it) } ?: Instant.now(clock),
recordedAt = Instant.now(clock),
```

`AuditEvent` has no `occurredAt`, so for any producer serializing the canonical envelope the lookup misses and `occurredAt` collapses to `recordedAt`. The consumer's other field lookups already carry per-producer fallback chains (`requestedBy` / `actorId` / `initiatedByPartyId`, `eventType` / `type`); the time field has none.

## Why it hasn't bitten yet

It is currently unreachable, and that is the only reason. The single `AuditEventPublisher` implementation in the tree is `LoggingAuditEventPublisher` — it logs, it does not serialize to Kafka. So no producer sends a canonical `AuditEvent` to `audit-events-in` today. The KDoc on `AuditEventPublisher` explicitly invites a contributor to add an `@Alternative @Priority` Kafka implementation, and that PR is the one that silently ships the gap.

The existing producers that *do* reach the consumer (`EdgeAuditPublisher` and the various domain event topics) emit their own flat shapes, which is why the consumer's field names look the way they do.

## The failure mode, once reachable

[...]

PR #3926, opened 2026-08-06
fix: send occurredAt on the four non-money-path domain-event producers

JiRaska:
Closes part of #3914. Refs #3883, #3907.

## What

#3907 taught `AuditConsumer` to record **whether** `occurred_at` is the producer's event time
(`EVENT`) or the consumer's ingest time (`INGEST`). It did not fix the producers. This fixes four
of the seven — the four that are **not** money-path, so this PR can actually land.

| service | payloads | occurredAt now is |
|---|---|---|
| dispute | `dispute.resolved`, `dispute.remediation_requested`, shared complaint payload | `dispute.resolvedAt` (both, shared deliberately) / `complaint.updatedAt` |
| statement | `period.closed.v1`, `period.close_failed.v1` | `StatementPeriod.closedAt` / the same clock read as `failedAt` |
| sanctions | screening event, review event | `checkedAt` / `reviewedAt ?: checkedAt` |
| document | `DocumentGenerated`, `SignatureCeremonyCompleted` | renamed from `at` |

**Not in this PR, tracked on #3914:** clearing-service (2 payloads), sepa-payment's Temporal path
(5 payloads), lending-service (6 of 10 payloads). All three are in `rules.yaml:
money_path_services` — 2 human approvals plus a threat model each — so bundling them here would
have stalled the whole change.

## Which instant, and why per payload

A blanket `Instant.now()` at serialisation would have satisfied the metric and been worse than the
`INGEST` it replaced: a wrong-but-present timestamp is indistinguishable from a real one. Each
payload's choice is argued in place at the call site. Two are worth flagging here:

- **dispute's two resolve events share one instant on purpose.** They are emitted in one
  transaction for one indivisible state change; two fresh clock reads would put two "when"s on it.
- **sanctions could not put `occurredAt` on the aggregate.** The payload *is* the serialised
  `SanctionsCheck`, and the screening and review events legitimately mean different moments, so the
  field is added to the serialised node instead. Same reason it is written as an explicit ISO-8601
  string: that service's mapper uses `findAndRegisterModules()` and leaves
  `WRITE_DATES_AS_TIMESTAMPS` on, so `checkedAt` itself serialises as an epoch **number**, which
  `Instant.parse` cannot read.

## RED first

Every assertion was run against the unfixed code and is quoted below verbatim. The shared rule
lives once, in `openbank-libs-testing`'s new `AuditEventTime` — not restated per service, because a
second hand-kept copy of a cross-module agreement moves with the first and keeps passing.

```
DocumentRenderServiceTest > the DocumentGenerated payload carries the generation instant ...
  AssertionFailedError: [audit occurred_at_source for payload {"documentId":"019fd787-...",
  "templateCode":"LOAN_AGREEMENT","sha256":"6682d70d...","at":1768472130.000000000}]
  expected: EVENT but was: INGEST

SignatureCeremonyServiceTest > the SignatureCeremonyCompleted payload carries the completion instant ...
  expected: EVENT but was: INGEST

DisputeServiceTest > both resolve outbox payloads carry the resolution instant ...
  AssertionFailedError: [... {"eventType":"dispute.resolved","reference":"DSP-3001",
  "status":"RESOLVED_CUSTOMER","resolvedAt":"2025-01-15T10:00Z"}]
  expected: EVENT but was: INGEST

ComplaintServiceTest > the complaint payload carries the transition instant ...
  expected: EVENT but was: INGEST

StatementServiceTest > the period-closed payload carries the close instant ...
  expected: EVENT but was: INGEST

CloseOrchestratorTest > the close-failed payload carries the failure instant ...
  AssertionFailedError: [... "reason":"UNKNOWN","detail":"connection refused",
  "failedAt":"2026-06-20T10:00:00Z"}]
  expected: EVENT but was: INGEST

SanctionsReviewUpdateIT > both the screening and the review payloads carry their own instant ...
  AssertionFailedError: [... "checkedLists":["OFAC_SDN"],"reviewedBy":null,"checkedAt":1785668290.0000...]
  expected: EVENT but was: INGEST
```

Green after: 304 tests across the four services, 0 failures. Two `Failed to start quarkus` errors
in a combined run (`DisputeOutboxClaimIT`, `SanctionsPactProviderVerificationTest`) are the known
concurrent-Quarkus-JVM contention on this machine — both pass in isolation, and neither touches
this change. `detekt` + `ktlintCheck` green on all five modules.

## Two things the red run found that I had not predicted

1. **The document-service rename alone would not have worked.** Both test classes build their own
   `ObjectMapper` with a bare `JavaTimeModule`, which keeps `WRITE_DATES_AS_TIMESTAMPS` on, so the
   instant serialised as `1768472130.000000000` — unreadable to `Instant.parse` under either name.
   The mappers now mirror the CDI-injected one, as the fleet convention already requires. Had the
   test not existed, the rename would have shipped looking correct.
2. **`CloseOrchestratorTest`'s stub was hiding the path.** `runs.recordFailure` was stubbed with a
   relaxed mock whose `Uni` never emits, so the `flatMap` that emits the event was never reached —
   a green-looking test asserting nothing. It now echoes a real `CloseFailure` back.

## What I re-derived vs took on trust

**Re-derived myself, per service:** the producer (the actual outbox-row writer, not the port or the
publisher shape), the topic, that the publisher emits `entry.payload` verbatim, that the topic is in
`audit-events-in`'s `topics:` list, and that `openbank.outbox.dispatch-enabled` is `true`. That is
the path #3883 got wrong by comparing two field names across two files without establishing that
either reaches the other.

**Corrections to the survey in #3907's body,** all from re-measuring: lending is 6 of **10** payload
sites (not 8) — the 6 is right, the denominator was not; dispute is **3** payload builders, not 1;
sepa-payment is **5** hand-built payloads in `SepaPaymentActivitiesImpl`, while `SepaPaymentService`
itself was always fine.

**Took on trust:** that the other 14 topics are fine. I spot-checked each producer for an
`occurredAt` write and found one in every case, but I did not trace all 14 paths end to end the way
I traced these seven.

## What this does not prove

The service tests prove the **producer** half: the payload really written carries an event time the
consumer accepts, as the exact instant intended. They do not run Kafka or the consumer. The sanctions
case is the strongest — it reads the payload back out of a real `sanctions_outbox` row through a
real Postgres — but even there the transport is established by inspection, not executed. The
end-to-end success measure stays what #3914 states: these services' rows moving from
`occurred_at_source = 'INGEST'` to `'EVENT'`, and their series in
`openbank.audit.event.time.missing{source_service}` going to zero.

## Backward compatibility

Additive everywhere except document-service, where `at` -> `occurredAt` is a rename. Nothing in the
fleet deserialises those two events off the wire (checked), so no reader breaks; `analytics-sink`,
the second consumer of that topic, starts reading a field it always looked for and never found.

`docs/asyncapi/openbank-events.yaml` needed no change and says something useful: its `EventEnvelope`
has listed `occurredAt` as **required** all along, and dispute's and sanctions' payload schemas both
`allOf` it. The spec was already right; the producers were not. (Statement and document have no
AsyncAPI channel at all. Separately noted on #3914 and not fixed here: the doc gives dispute's
address as `openbank.disputes.dispute.event`, while producer and consumer both use
`openbank.dispute.events`.)

No API change, no DB change, no version.txt edit — release-please classifies from the `fix(...)`
commits, one per service.


JiRaska:
Closes the producer half of #3883 / **#3914**. #3907 made the *absence* of an event time visible
(`audit_entries.occurred_at_source` = `EVENT`|`INGEST`); this makes four producers stop being `INGEST`.

## What changed

| service | instant chosen | why |
|---|---|---|
| dispute | the business instant on every outbox event | — |
| statement | the business instant on both outbox events | — |
| sanctions | `checkedAt` on save, `reviewedAt ?: checkedAt` on update | the instant is per **event**, not per aggregate: the update path *is* the manual disposition, so the review instant is the business event. The fallback is load-bearing — `reviewedAt` is nullable, and a null there would silently cost the audit trail the event time of what `rules.yaml` calls this service's highest-risk action |
| document | renamed the events' `at` → `occurredAt` | per #3907's contract decision: `occurredAt` is the **single** accepted key, because a second accepted spelling is a second silent path — which is how this survived |

Plus `openbank-libs-testing`'s `AuditEventTime`: one shared copy of the rule instead of four
near-identical assertions.

## Success measure

Not "the field appears in the payload" — **the rows move from `INGEST` to `EVENT`**. That is what
#3907's column exists to show, and it is the only check that cannot be satisfied by a
well-formed-but-wrong payload.

## Verification, and a correction I owe this PR

The authoring session stalled immediately before pushing, so I ran the gate myself and **nearly drew
the wrong conclusion twice.**

First run, four suites concurrently:

```
DisputeOutboxClaimIT > two concurrent claims never return the same row()   FAILED
SanctionsPactProviderVerificationTest > ... POST a full-list screen ...     FAILED
```

- **The dispute failure is pre-existing** — the same test fails on clean `origin/main`. Not this
  change.
- **The sanctions "failure" was not one.** The console line is pact-jvm's documented artifact; the
  JUnit XML for that run reads `tests=2 failures=0 errors=0`. Re-run alone: `BUILD SUCCESSFUL`,
  **89 tests, 0 failures**. I had already concluded "the change broke a contract" before reading the
  XML — the repo's own rule (read the verdict from the XML, never the console) is what reversed it.

The original combined failure was environmental: four Quarkus suites in parallel while other work ran
on the same machine.

Clean re-runs, `--rerun-tasks`, tasks observed executing:

```
:openbank-sanctions-service:test   89 tests, 0 failures   BUILD SUCCESSFUL
:openbank-statement-service:test
:openbank-document-service:test    BUILD SUCCESSFUL
:openbank-libs-testing:test
```

## Not verified

- **The three remaining producers are untouched** — lending (6 of 8 payloads), clearing, and
  sepa-payment's Temporal path. Deliberate: each needs a per-payload decision about which instant is
  the business event, and a wrong-but-present timestamp is worse than an honest `INGEST`. They stay
  on #3914.
- **No `EVENT` row observed** — the success measure above is the design, not a measurement; nothing
  was deployed and no database was read.
- `docs/asyncapi/openbank-events.yaml` was flagged as unchecked in #3907 and I did not check it here
  either.
- The five commits are the authoring session's; I verified their content, signatures (`%G?` = `G`)
  and the build, but did not re-derive its per-service survey.

Refs #3914, #3883, #3907

JiRaska:
Review: 2 findings, 0 blocking

1. Non-blocking, but it is the #3882 shape — three of the six new assertions compute `expected` from the object under test, so they prove *agreement*, not that the instant is real. `ComplaintServiceTest.kt:111` (`filed.updatedAt.toInstant()`), `StatementServiceTest.kt:816` (`closed.closedAt`) and `SanctionsReviewUpdateIT.kt:622-623` (`stored.checkedAt` / `reviewed.reviewedAt`) all read the value back off the aggregate the producer serialised. Concrete failure they cannot see: if `Complaint.updatedAt` ever picked up a default (the entity already defaults to `OffsetDateTime.MIN`, `ComplaintEntity.kt:86`) or a transition stopped setting it from the clock, both sides of the assertion move together and the test stays green while every complaint audit row claims 1970/-999999999. `CloseOrchestratorTest.kt:768` and both document tests do it right — a literal (`Instant.parse("2026-06-20T10:00:00Z")`, `FIXED_NOW`). Pinning the other three to the same fixed-clock literal costs nothing and makes them falsifiable.
2. Non-blocking, worth one line in the PR body: `SanctionsRepositoryImpl.kt:122-124` writes `occurredAt` as an explicit ISO-8601 **string** into a tree produced by `jacksonObjectMapper().findAndRegisterModules()` — which leaves `WRITE_DATES_AS_TIMESTAMPS` **on**, so `checkedAt`/`reviewedAt` in the very same payload stay epoch **numbers**. Correct for `AuditConsumer` (that is exactly why the explicit `put` is right), but the payload now carries two encodings for the same type, and any future consumer reading `checkedAt` gets a number. Pre-existing, not introduced here.

Verified clean, item by item:
- **Format match.** `AuditConsumer.eventTime` (`AuditConsumer.kt:108`) is `node["occurredAt"]?.asText()` + `Instant.parse` — `occurredAt` and only `occurredAt`. `Instant.toString()` (sanctions), `OffsetDateTime.toInstant().toString()` (dispute complaint) and the string-interpolated `Instant`s (statement) all render ISO-8601 instants `Instant.parse` accepts, so these payloads flip INGEST -> EVENT as claimed. The two document-service tests' `.disable(WRITE_DATES_AS_TIMESTAMPS)` is not cosmetic — without it the hand-built mapper emits an epoch **number** and the assertion is red even with the field correctly named; the comment in place says so.
- **No 1970 path.** `StatementPeriod.closedAt: Instant` and `SanctionsCheck.checkedAt: Instant` are non-nullable with no default; `SanctionsCheck.reviewedAt: Instant?` is nullable and the `?: checkedAt` fallback is genuinely reachable and correct; `Complaint.updatedAt: OffsetDateTime` is non-nullable and set from the injected clock in every `ComplaintService` transition (lines 71/82/92/104); `Dispute.resolvedAt` is nullable with an `Instant.now(clock)` fallback. No source field can ship an EPOCH under the current code.
- **The `at` -> `occurredAt` rename is safe, but only by luck of who reads it.** Grepped both consumers of `openbank.documents.document.event` (audit-service, analytics-sink): neither reads `at` — audit reads `occurredAt` only, analytics-sink uses the same `occurredAt ?: now` fallback. So removing `at` strands nothing. Note that nothing would have told you: `check-event-schema-compat.py` is scoped to classes declaring `: DomainEvent(`, and `DocumentEvents.kt` are plain data classes, so a genuinely breaking removal here is invisible to the gate. Renaming the two unwired events (`DocumentTemplatePublished`, `DocumentSigned`) in the same file is the right call for exactly the reason the comment gives.
- **Build changes are minimal:** one `testImplementation(project(":openbank-libs-testing"))` per consuming service (4), plus `api(libs.jackson.datatype.jsr310)` in `openbank-libs-testing` because the helper parses JSON and that module has no Quarkus BOM. Nothing lands on a production classpath.
- **`AuditEventTime` is used consistently** — all six new assertions go through `assertRecordedAsEventTime`, none re-implements the rule locally, and the helper asserts both halves (source == EVENT *and* the value), which is what makes finding (1) a sharpening rather than a hole.

Already-written data: **yes, wrong, and this PR does not and cannot fix it.** Every audit row already ingested from `openbank.dispute.events`, `openbank.statement.event`, `openbank.sanctions.screening.event` and `openbank.documents.document.event` carries the consumer's ingest clock as `occurred_at` with `occurred_at_source = INGEST`. `audit_entries` is append-only at the DB (updates are silently discarded), so there is no backfill — the historical rows stay INGEST-flagged, which is at least honest. Worth stating in the PR body so nobody later reads a pre-fix row as a producer timestamp.

Automated agent review, not an approval. The four producers here are non-money-path as the title says, but sanctions screening is the GDPR Art. 30 / DORA Art. 17 evidence path — human review regardless.

JiRaska:
**Review: 3 findings, 0 blocking**

Checked the direction question first, because it is the one that decides whether this PR does
anything: `AuditConsumer.eventTime` on `main` reads `node["occurredAt"]` and only that, parses it as
an `Instant`, and falls back to `Instant.now(clock)` with `OccurredAtSource.INGEST`. So the producer
side is the correct side to fix, and the field being populated is genuinely consumed — by
audit-service and by `openbank-analytics-sink`, which reads the same spelling. Per-payload instant
choices are right: `period.closedAt` (an `Instant`), `dispute.resolvedAt?.toInstant()`,
`reviewedAt ?: checkedAt`, and one clock read shared by `failedAt`/`occurredAt`. No `Instant.EPOCH`
defaults, and the assertions compare to a specific expected instant rather than `isNotNull()`.

**1. `AuditEventTime` is a second copy of the consumer's rule, which is the thing its own KDoc warns
about.**
`openbank-libs-testing/.../audit/AuditEventTime.kt` — it re-implements
`mapper.readTree(payload)["occurredAt"]` + `Instant.parse` in a module `openbank-audit-service` does
not use. "There is exactly one copy of the rule here" is true across *producers* and not true across
the agreement: if `AuditConsumer.eventTime` ever accepts a second spelling, tightens the parse, or
changes the fallback, all seven producer tests stay green against a consumer that no longer behaves
that way. Cheap close: have audit-service's own consumer test assert through the same helper, so the
two artefacts cannot move apart silently. Worth doing before the money-path half of #3914 lands on
top of it.

**2. The document-service change is a payload rename, not an additive field, and the PR presents it
as the latter.**
`DocumentEvents.kt` — `at` -> `occurredAt` on all four events (two of them live on
`openbank.documents.document.event`). Contrast sanctions, where the same goal was met additively and
the KDoc says why. I checked the consumers: audit-service and analytics-sink both read `occurredAt`
and neither reads `at`, and `.github/event-contract-baseline.txt` tracks the topic, not its fields,
so nothing in-repo breaks and no gate will notice. But any consumer outside this repo — or a
silver-layer projection keyed on `at`, per ADR-0210 — starts reading null with no error anywhere.
That deserves a line in the commit body/changelog saying the payload field was renamed, since
`fix:` on this scope reads as compatible.

**3. Interpolated payloads make `occurredAt`'s parseability a property of a type nobody is pinning.**
`StatementService.kt` / `CloseOrchestrator.kt` / `DisputeService.kt` build the JSON as strings, so
`"occurredAt":"${period.closedAt}"` is correct precisely because `closedAt` is an `Instant`
(`StatementPeriod.kt:29`) and `Instant.toString()` is ISO-8601 instant format. The day that field
becomes an `OffsetDateTime` — which is what dispute's `.toInstant()` call exists to work around —
the interpolation still compiles, still emits a plausible `2026-08-07T10:00+01:00`, and
`Instant.parse` rejects it: the row silently reverts to `INGEST`, the exact defect this PR fixes,
with no compile error and no failing assertion outside the fixed-clock unit tests. A
`DateTimeFormatter.ISO_INSTANT.format(...)` (or an explicit `.toInstant()` everywhere, as dispute
does) removes the dependency on a type that is not part of the contract.

Verified clean: producer and consumer actually meet (the field name matches on both sides and
audit-service subscribes to these topics via `audit-events-in`); sanctions writes
`occurredAt.toString()` explicitly as a string rather than letting a `WRITE_DATES_AS_TIMESTAMPS`
mapper emit an epoch number the consumer cannot parse — that reasoning is correct and is documented
in place; the dispute pair deliberately shares one instant for one transaction; the RED-first
evidence quotes real assertion failures with the unfixed payload inline, so the tests are falsified
rather than assumed. Scoping the money-path services out to #3914 is the right call.

---
Automated agent review — not an approval.

JiRaska:
Second review pass. **No new blocking finding** — the two existing reviews cover the ground, and I confirm their key points rather than restating them.

Confirmed: `AuditConsumer.eventTime` reads `node["occurredAt"]` and only that, so this closes the producer half of the #3883 envelope defect; the field has **two** real readers (audit-service and `openbank-analytics-sink`), so it is not write-only. The document-service `at` → `occurredAt` **rename** is the one genuinely breaking edge, and the observation that `check-event-schema-compat.py` cannot see it is correct and worth keeping visible: the gate only inspects classes declaring `: DomainEvent(`, and `DocumentEvents.kt` are plain data classes, so a field removal there is invisible to it. In-repo consumers read neither field, so the risk is out-of-repo consumers only.

Also confirmed, and this is the part that matters most given what this PR is fixing: several assertions compute `expected` from the object under test (`ComplaintServiceTest.kt:111`, `StatementServiceTest.kt:816`, `SanctionsReviewUpdateIT.kt:622-623`), so they assert *agreement*, not recency — the #3882 shape. `ComplaintEntity.kt:86` defaulting to `OffsetDateTime.MIN` is the live path such a test would not notice. No `isNotNull()` anywhere, and `CloseOrchestratorTest.kt:768` plus the document tests use fixed-clock literals, which is the right model.

One latent defect worth keeping from the third review because it is the kind that reappears: `StatementService.kt` and `CloseOrchestrator.kt` build the payload by string interpolation, so `"occurredAt":"${period.closedAt}"` is parseable *only because* `closedAt` happens to be an `Instant`. If that field ever becomes an `OffsetDateTime` it still compiles, emits `…+01:00`, `Instant.parse` rejects it, and the row silently reverts to `INGEST` — reintroducing precisely the defect this PR fixes, with no failing test outside the fixed-clock units. `DateTimeFormatter.ISO_INSTANT.format(...)` closes it for one line.

JiRaska:
**Unblocked: merged `origin/main` in (`fa8e8542f`, signed). It had gone CONFLICTING — a "green and ready" reading taken minutes earlier was already stale.**

One conflict, in `StatementService.kt`, and it is the interesting kind rather than a text clash: `main` renamed `netMovement` to `netMovementOf`, so a plain "keep ours" would have left an **orphaned private function** behind — both call sites in the merged file already say `netMovementOf`, and the old body would have sat there as dead code for `detekt` to reject.

Resolved by dropping the renamed-away function and keeping the `occurredAt` KDoc that sat next to it, which documents `periodClosedEvent` and is this PR's actual content:

> `occurredAt` is `StatementPeriod.closedAt` — the period-close instant this event announces — and NOT a clock read taken while serialising (#3914).

Verified after resolving, not assumed:

| check | result |
|---|---|
| conflict markers remaining | 0 |
| `netMovement` references | only `netMovementOf`, at both call sites |
| the `occurredAt` KDoc | present |
| `:openbank-statement-service:test` | **76 tests, 0 failures, 0 errors, 0 skipped** — read from the 18 JUnit XML files, not the console |
| `ktlintCheck` | in the executed task list |
| `detekt` | re-run with `--rerun-tasks`, BUILD SUCCESSFUL — it did not appear in the first combined run's task list, so a cached green could not stand in for an execution |

Scope after the merge is 21 files against `main`, matching what the PR had before — checked because a merge git calls clean can still drop a neighbouring entry with nothing printed.

Nothing was rebased or force-pushed: `main` was merged **in**, so the worktree at `/private/tmp/ob-3883b-39569` is simply behind and loses nothing.

JiRaska:
**Unblocked (`4074bce73`, signed). Two conflicts in `DisputeService.kt`, and the second one is the dangerous kind.**

**1. Adjacent additions.** `main` added a whole `openedOutboxMessage` function with its own KDoc; this branch added the KDoc for the function *following* the hunk. Kept both.

**2. Two different fields added to the same JSON payload.** `main` added `partyId`, this branch added `occurredAt`, to `dispute.resolved`:

```kotlin
"""{"eventType":"dispute.resolved","disputeId":"${dispute.id}",""" +
    """"reference":"${dispute.reference}","partyId":"${dispute.partyId}",""" +
    """"outcome":"${dispute.remediationOutcome}",""" +
    """"status":"${dispute.status}","resolvedAt":"${dispute.resolvedAt}",""" +
    """"occurredAt":"${dispute.resolvedAt?.toInstant() ?: Instant.now(clock)}"}""",
```

Taking either side would have compiled, passed review, and **silently dropped the other's entire feature** — either the ADR-0220 arrears consumer loses the party it needs, or the audit trail goes back to recording ingest time as business time. Both are in.

**My first resolution broke the build, and that is worth recording.** The shared `/**` sat *above* the conflict marker, so it was consumed as context by main's side and this branch's KDoc came out headless:

```
e: DisputeService.kt:270:6 Syntax error: Expecting member declaration.
```

That failed loudly at compile time. The version of this that does not — a payload silently missing a field — is exactly what case 2 above would have been.

Verified after resolving rather than assumed: `:openbank-dispute-service:test` → **52 tests, 0 failures, 0 errors**, read from the JUnit XML.

Issue #2364, opened 2026-07-25
The prod-readiness collector hardcodes 14 money-path services against rules.yaml's 20 — six are gated leniently and three have no row at all


JiRaska:
`openbank-infra/scripts/prod-readiness-collector.py` decides which services get the stricter money-path gate from a **hardcoded set**, not from `rules.yaml`. Its own comment admits it:

```python
# Money-path services (mirror of rules.yaml: money_path_services). Kept short
# here; the real collector would parse rules.yaml. These get a stricter gate.
MONEY_PATH = {
    "ledger", "transaction", "account", "balance", "sepa-payment",
    "sepa-instant", "domestic-payment", "clearing", "swift", "fx",
    "lending", "sca", "consent", "fraud",
}
```

That copy has drifted, and it drifts in the direction that makes the matrix look better than it is.

## Measured against `origin/main` @ 0d669339

`rules.yaml: money_path_services` — the authoritative list — declares **20** services. The collector's copy names **14**. The matrix actually applies the money-path gate to **11**.

**Six declared money-path services are scored with the LENIENT gate** (all ≥2, no cell required at 3):

`sdd` · `interest` · `billing` · `settlement` · `sanctions` · `vop`

All six currently read **GO**. Under their correct gate they would each need **3** on C1/C5/C7 and would read NO-GO, exactly like the other eleven. So six of the matrix's GO verdicts are a weaker claim than the column implies — and `rules.yaml` annotates several of them as unambiguously money-path (`sdd` "posts a real debit", `interest` "capitalize()/write-off post real GL journals", `settlement` "debit → credit → ledger booking saga", `billing` "fee assess/post/reverse shipped").

**Three declared money-path modules have no row in the matrix at all:**

`openbank-sepa-payment` · `openbank-sepa-instant` · `openbank-domestic-payment`

They are named in the collector's own `MONEY_PATH` set, yet `all_services()` enumerates `REPO.glob("openbank-*-service")` — and these three are not `-service` suffixed. All three carry a `version.txt`, a `governance.yaml`, an `openapi.yaml` and a threat model, so they are fully-fledged released components; they are simply invisible to the scorer. Every headline the matrix produces ("N cells below Verified", "GO x/37") silently excludes the three modules that move SEPA and domestic payments.

## Why this is the same failure class the sweep kept finding

[...]

PR #3929, opened 2026-08-06
feat(governance): guard prod-readiness attestation format and freshness (#2365)

JiRaska:
## What

A format + freshness gate over `openbank-libs/governance/attestations.yaml` — the only
hand-entered input to the production-readiness matrix, and the one file nothing read.

`readiness-attestation-format`, declared in `.github/gates/gates.yaml` (`group: registry`,
`mode: enforced`), backed by `.github/scripts/check-readiness-attestations.py` with a
`--self-test`. Not an inline workflow step: a gate written as one inherits its host job's
`if:` and silently narrows its own scope.

## The defect it found on the way in

The collector looks attestations up by the **short** service name (`consent`, from
`openbank-consent-service`). The committed file said `consent-service:`, which resolves to
`openbank-consent-service-service` and matches nothing.

Measured — renaming that one key:

| | before | after |
|---|---|---|
| consent C7 (Security) | 2 | **3** |
| ledger C5 / C7 | 3 / 3 | 3 / 3 |

A wrong key is invisible from both ends: the file reads as attested, the matrix reads as
underscored, and neither one points at the other. So the honest count is **3 attestations
written, 2 of which ever scored anything**.

**No attestation was added.** Both live `pentest` entries were verified against the runs
they cite before the key was corrected — the ZAP baseline run of 2026-07-26 has one job
named `ZAP baseline / ledger-service`, and the API fuzz run of 2026-07-28 booted
`openbank-consent-service` on port 8106 and fuzzed its OpenAPI. The consent entry's
evidence was always real; only its key was wrong.

## What the gate asserts

Structural properties only — it never reasons about what a word means, and it mirrors the
collector's own line grammar, because the question is not "is this valid YAML" but "does
the consumer see this entry".

- every entry carries `date`, `ttl_days`, `by`, `ref`
- `date` is a real calendar date, and not in the future
- `ttl_days` is a positive integer (absent, the collector silently assumes 365)
- the service key resolves to a directory the collector actually scores
- the attestation key is one of the eight `attestFresh` call sites consult
- `ref` resolves: a URL, a `#issue`, an existing repo path, or `runbook-NNNN`
- every `(service, key)` pair a YAML reader sees is one the collector's grammar actually
  ingests — a block-style entry is valid YAML, reads exactly like an attestation, and is
  skipped by both the collector and a naive mirror of it. Verified against the real tree: a
  block-style `fx.pentest` leaves the count at 3 and now raises an error saying so, instead
  of vanishing.

It cannot prove the attested event happened. Nothing in a repo can. It makes a fabrication
require a deliberate lie — a real date, a resolvable reference, a service that exists —
rather than a careless line.

## FAIL vs WARN — the decision

**Authoring defects FAIL.** A missing field, a bad or future date, a dead service key, an
unknown attestation key, an unresolvable `ref`: each is introduced by the PR that writes it
and fixed by that same PR.

**Decay only WARNS**, and that is deliberate rather than timid. The collector's TTL is
already the real enforcement — after expiry it stops counting the bonus and the dimension
falls back to its derived score, so a lapsed attestation *cannot buy a green cell*. What it
can still do is mislead a human reading the file, and a warning with a countdown addresses
exactly that. Failing the whole fleet's CI the morning a pentest TTL rolls over would
punish every unrelated PR for an operational event none of them can fix — the alert-fatigue
trade this repo has already paid for once.

**Expired for more than 30 days IS an error** (`--stale-fail-days`). At that point it is
not a lapse, it is stale bookkeeping asserting a control that ended a month ago, and the
fix is one line: delete the entry, or re-attest a real event.

## Falsified in both directions, against the real tree

Self-test: **22 cases, all passing** — 18 must-flag (16 error, 2 warning) and 4
must-not-flag, the latter including a commented-out example that must never be read as an
attestation, and an empty file that must report a count of zero. A missing file is itself a
must-flag case.

Injected into the real `attestations.yaml`, then restored:

```
readiness-attestations: checked 6 attestation(s) ...        <- count moved 3 -> 6
::error:: `fx.pentest` is missing ref (every attestation needs date, ttl_days, by, ref)
::error:: service `sepa-payment` does not resolve to openbank-sepa-payment-service ...
::error:: `lending.pentest` ref 'we-ran-one' resolves to nothing -- cite a URL, a #issue, ...
exit=1
```

Decay, both sides of the threshold:

```
::warning:: `fx.dr_drill` EXPIRED 3 day(s) ago (2026-07-20 + 14d) ...      exit=0
::error::   `fx.dr_drill` EXPIRED 184 day(s) ago (2026-01-20 + 14d) ...    exit=1
```

And it does not flag the valid tree: `exit=0`, two freshness warnings, three attestations
counted. **The count prints on every run** — a green reading "0 attestations" would be a
gate that never opened the file, which is the failure mode this repo keeps hitting.

## Also in here

**The collector's TTL arithmetic let a TTL outlive itself.** `attestFresh` approximated a
year as 365 days and a month as 30, so ledger's 21-day pentest, dated 2026-07-26, still
counted on 2026-08-17. Replaced with exact dates. Measured across the decay window — no
score changes today; the two decay dates become 2026-08-17 (ledger) and 2026-08-19
(consent), i.e. the last valid days are 2026-08-16 and 2026-08-18 as the TTLs say.

`rules.yaml: readiness_attestations` records the rule and names the script as its
`ci_producer`. New top-level key, outside the OPA read set — `gen-rules-opa-data.py --check`
reports in sync, so no bundle restamp.

## What this does NOT do

- It does not raise 3/60. Missing evidence stays missing.
- It cannot verify a `ref` describes the event it claims. It checks the anchor resolves.
- Coverage is bounded by what the collector scores: `openbank-sepa-payment`,
  `openbank-sepa-instant` and `openbank-domestic-payment` have no `-service` suffix, so
  `allServices()` scores no row for them and no attestation for them can ever apply. The
  gate says so explicitly rather than passing silently, but it does not fix it (#2364).
- The denominator has moved: `rules.yaml: money_path_services` now lists **21** services,
  not the 20 #2365 was sized against, and the collector still gates 11. So the headline is
  **2 effective of 63**, not 3 of 60.

## Verification

- `run-gates.py --all` on this branch: **112 gates, PASS=103**, versus a pristine
  `origin/main` worktree at **111 gates, PASS=102**. Same 5 FAIL / 1 UNFALSIFIED / 3 WARN
  on both — all PR-context or local-environment gates (`api-contract`, `db-migration`,
  `schema-compat`, `threat-model`, `release-scope`, `loki-rule-load-test`) that need a base
  ref or docker. The delta is exactly this gate.
- `gates.yaml` re-read against live `origin/main` immediately before push: 111 there, 112
  here, and no id present on main is missing from this branch.
- `npm test` in `openbank-admin-ui`: 82 files, **1107 tests passed**.
- `rules.yaml` re-parsed with a duplicate-key-rejecting loader: clean, 44 top-level keys.

Refs #2365


JiRaska:
**Verdict: solid, well falsified.** 0 blocking findings, 1 minor note.

- `.github/scripts/check-readiness-attestations.py` — the self-test has 20 cases, each round-tripped through a real temp sandbox: it not only checks known authoring defects (missing field, bad date, future date, dead service key, unknown attestation key, unresolvable ref) but includes two cross-check cases that specifically probe the gate's own blind spots — a commented-out example that must read as clean, and a block-style YAML entry (valid YAML, but a shape the collector's line grammar never ingests) that must be flagged even though the primary line-grammar parser alone would miss it. That second case is exactly right: `parse()` deliberately mirrors `loadAttestations()`'s line grammar rather than using a general YAML reader, and `yaml_pairs()` exists solely to catch what that mirror can't. Good defense against "the gate agrees with itself and nothing else."
- `KNOWN_KEYS` (8 keys: `code_complete`, `coverage_floor`, `contract_verified`, `restore_drill`, `dr_drill`, `pentest`, `slo_defined`, `oncall`) matches the actual `attestFresh(short, '<key>')` call sites in `collect-prod-readiness.mjs` on `main` exactly — verified by grep, not just asserted in the PR description.
- Real side-fix bundled in: `attestFresh()`'s date math changes from a 365/30-day approximation to exact `Date.parse` arithmetic — the approximation let a 21-day TTL (ledger's pentest, dated 2026-07-26) still count as fresh on 2026-08-17, i.e. the TTL mechanism could outlive itself. Worth calling out explicitly in the PR description as a distinct fix, since it's easy to read past as a refactor inside a diff that's mostly about the new gate.
- Minor: `KNOWN_KEYS` is a second hand-kept copy of the same 8 keys `attestFresh` call sites define — this repo has a standing lesson about not letting a gate's coverage set drift from what it covers (`derive-pact-drift-scope.sh` is the citation for "derive, don't hand-list"). Here it's low risk since a new key added to the collector without updating `KNOWN_KEYS` fails safe (the gate would flag it as "unknown attestation key" until updated, not silently pass), but a comment noting that asymmetry (grep-based CI check, or just a pointer) would make the coupling explicit.

Verified clean: gate is declared in `.github/gates/gates.yaml` under `group: registry`, `mode: enforced`, reachable unconditionally (not an inline step in a path-filtered job); the FAIL-vs-WARN split for decay is deliberate and matches this repo's own alert-fatigue lesson (an expired-but-recent attestation warns, not fails, since the collector's TTL already self-corrects the score).

JiRaska:

JiRaska:
Reviewed. Strong gate — genuinely enforced, 22/22 self-test, real scope handling. **One small hole, reproduced.**

**Clean first, verified:** `run-gates.py --list` shows `registry  readiness-attestation-format  enforced  always  selftest`, declared in `gates.yaml` rather than as an inline step. `selftest_expect: pass` is correct. No `--enforce` flag and no `::warning`-and-exit-0 path — it `sys.exit(1)`s on findings, so `mode: enforced` means what it says. `KNOWN_KEYS` (8) matches the `attestFresh(short, …)` call sites in `collect-prod-readiness.mjs` exactly. The service-key rule is real and the `sepa-payment` / `sepa-instant` / `domestic-payment` no-suffix gap is *reported* rather than passing silently, which is the right handling of the usual scope hole.

**The hole: the YAML cross-check is one-directional.** `yaml_pairs()` exists to catch mirror-vs-YAML divergence, but only `seen - ingested` is checked; the reverse, `ingested - seen`, never is. Injecting a second `ledger:` heading into the real file:

```
readiness-attestations: checked 4 attestation(s) ...
readiness-attestations: format OK; 2 freshness warning(s)
rc=0
```

A YAML reader sees `ledger keys: ['dr_drill']` — `restore_drill` and `pentest` are gone, because SnakeYAML/PyYAML keep the last of a duplicated mapping key. And the collector loses them the same way: `loadAttestations()` does `data[curSvc] = {}` on every `indent === 0 && endsWith(':')` line, so it *resets* the service dict.

So the gate's headline defence — the printed denominator — is itself wrong in this case: it says 4 while both a YAML reader and the collector see 1 for ledger. That is the repo's own duplicate-key/last-wins lesson landing in the one file whose entire purpose is that the consumer must see what a reader sees. One extra loop reporting `ingested - seen` closes it, plus a self-test case for a duplicate service key.

**One asymmetry worth a comment, not blocking:** drift between `KNOWN_KEYS` and the collector fails safe in one direction and open in the other — a key *added* to the collector is flagged as unknown, while a key *removed* from it leaves an attestation for something nothing reads, passing silently.

(A second injection — a duplicate `pentest:` within one service — was caught, though incidentally, via a bad `ref` rather than as a duplicate.)

Issue #3410, opened 2026-08-02
Delegation events omit the validity window and ceilings that both projections enforce


JiRaska:
The delegation events do not carry the grant's validity window or its ceilings, but both
enforcement projections read them — so every projection invents the window it is enforcing.

Measured on current `main`:

```
validFrom / validTo / perTransactionLimit in the event classes … 0 occurrences
read by account-service DelegationEventConsumer …………………………………… 10
read by card-issuance CardDelegationEventConsumer ………………………………  6
DelegationRepositoryImpl:48  payload = objectMapper.writeValueAsString(event)
```

The aggregate has all three (`DelegationGrant.kt`: `validFrom`, `validTo`, `perTransactionLimit`),
and `DelegationService.offer` constructs `DelegationOffered(...)` without them. The outbox payload
is the serialized **event**, so they never reach the wire, and `node.path("validFrom")` on the
consumer side resolves to nothing.

## What each consumer does with the absence

- **`validFrom` → `?: OffsetDateTime.now()`**. A grant dated to start next month becomes
  enforceable the moment its `DelegationActivated` lands. **No backstop.**
- **`validTo` → `null`**, which every guard reads as "no end". A time-limited grant never expires
  locally. Backstopped only by delegation-service's expiry sweep emitting `DelegationExpired`, so
  the window is honoured *late* rather than never — and only while that sweep runs. (It is a
  `suspend fun` since #3164; before that it never ran at all, which is what this would have been
  hiding behind.)
- **`perTransactionLimit` → absent entirely.** No projection can apply a per-transaction ceiling,
  so a capability granted "up to 5 000 CZK" is enforced as unbounded wherever the projection is
  the authority.

## Why it is worth its own issue

None of it is visible from either side alone: the producer looks correct (it emits its own event
faithfully), and the consumers look correct (they read fields with sensible fallbacks). Only the
pair is wrong, which is exactly the class ADR-0006 contracts exist to make visible — and
`openbank-contracts/openbank-delegation-service/asyncapi.yaml`, the repo's first contract, does not
currently mention any of the three (0 occurrences), so it documents the intended shape rather than
the wire.

## Suggested fix

Add `validFrom`, `validTo` and `perTransactionLimit` to the activation-shaped events
(`DelegationOffered`, `DelegationActivated`, `DelegationReinstated`). Additive and BACKWARD-
compatible, so it needs no consumer change to land — the consumers already read them. Then record
[...]

PR #3930, opened 2026-08-06
ci(governance): assert an event contract agrees with the code it describes

JiRaska:
Picked up after the authoring session stalled mid-work (uncommitted tree, no PR). I verified the gate
myself before committing rather than shipping unfinished work on trust.

## What this adds

The existing ADR-0006 ratchet asserts a contract **file** exists for a producer. It cannot see
whether that file still describes the code — the failure this repo keeps finding one layer down: the
artefact is present and the claim is unchecked.

`check-event-contract-code-agreement.py` compares each `openbank-contracts/*/asyncapi.yaml` against
its producing source — channels, message names, payload properties.

## Verification I ran

```
--self-test        self-test: all 6 case(s) detected.       rc=0
against the tree   OK: 2 event contract(s) agree with their
                   producing code (channels, message names
                   and payload properties).                 rc=0
```

The six cases include the **#3410 shape** — a class property present in code and absent from the
contract, i.e. the contract silently understating what it emits.

**It prints the denominator.** `2 event contract(s)` rather than a bare "OK", so a green cannot mean
"it opened none" — the exact way `domain-purity-gate` was green while never looking at 80 of 107
files, and the reason a floor was nearly set *below* the regression it existed to catch earlier
today.

## A stale measurement corrected

The ratchet's own comment claimed *zero* `asyncapi.yaml` and no contracts directory. Two contracts
(delegation, notification) have since been written, so the comment was describing a tree that no
longer exists. Re-measured: **38 producer pairs across 32 services / 36 distinct topics, 2 contracted
and 36 grandfathered.**

What has **not** changed is the part that matters, and the corrected comment says so plainly: the
`.avsc` count is still **ZERO** and no topic has a registered schema. The Apicurio registry ADR-0006
calls for is deployed and running — **the runtime half exists and the contract half does not**, and
every Kafka event on this platform is still raw unversioned JSON.

## Scope — deliberately not the whole issue

#1916 asks for a schema registry and Avro across the estate. This is the measurement plus one
enforceable layer, not that programme. The remaining decisions are ADR-level (Avro vs JSON Schema,
compatibility mode, who owns the registry lifecycle) and are not mine to pick.

## Not verified

- I did not re-derive the authoring session's producer survey; I ran its gate and read its output.
- The gate is **source-text** comparison. A payload assembled dynamically, or a property reached via
  a `typealias`, is invisible to it — the same limit `domain-purity-gate` carries and states.
- 36 topics are grandfathered. This gate says nothing about them; it only holds the two contracts
  that exist to their code.
- No registry interaction: nothing here talks to Apicurio or checks a registered schema.

Refs #1916, ADR-0006


JiRaska:
**Verdict: solid — this is one of the few gates in the batch with real falsification evidence.** 0 blocking findings, 1 minor note.

- `.github/scripts/check-event-contract-code-agreement.py` — the self-test (`--self-test`, wired as `selftest:`/`selftest_expect: pass` in `.github/gates/gates.yaml`) doesn't just assert "the checker ran"; it copies the two real contracted services (delegation, notification) into an isolated temp tree, confirms the UNMUTATED fixture is clean first (catches fixture staleness rather than trusting it), then mutates six distinct ways — a channel address the service doesn't produce, a produced topic with no channel, a message with no producing class, an event type with no message, a documented property absent from the class, and a constructor property the contract omits (explicitly the #3410 defect shape) — and asserts each mutation is caught with the right error substring. That's a real failure-path demonstration, not just an exit-code check.
- `strip_kotlin_comments()` tracks block-comment nesting depth explicitly rather than matching the first `*/` — the exact code-about-code trap this repo has been bitten by twice (`check-roles-allowed-realm.py`, the `matcher` grep on #3072). Good.
- Minor: the checker restricts event-class discovery to `src/main/kotlin` (excludes `src/test`), which avoids the class of false-positive where a test fixture reusing `override val eventType = "X"` gets counted as a producer — worth a one-line comment in the script confirming that's deliberate rather than incidental, since a future edit widening the glob would silently reintroduce it.

Verified clean: gate is declared in `.github/gates/gates.yaml` under `group: data`, `mode: enforced` (not an inline conditional-job step), so it's reachable regardless of which paths changed; `selftest_expect: pass` is set correctly for a checker whose own `--self-test` is the falsifiability harness, not the known-positive itself.

JiRaska:

JiRaska:
Reviewed. Registration and enforcement are exemplary; **one falsifiable hole the self-test does not model, reproduced.**

**Clean first, verified rather than read:** `run-gates.py --list` shows `data  event-contract-code-agreement  enforced  always  selftest` — declared in `gates.yaml` with `id`/`group`/`mode`, not an inline step in a conditional job. `selftest_expect: pass` is right, since the command is the checker's own harness (6/6). Rule 3 is satisfied *by absence*: there is no `--enforce` flag and no `::warning`-and-exit-0 path — the script `sys.exit(1)`s on findings, so `mode: enforced` is not vacuous. Comment stripping tracks block-comment nesting depth, which matters because Kotlin's nest. Scope is derived from `openbank-contracts/*/asyncapi.yaml` with no hand-list, and the 36 grandfathered topics reading as unchecked is stated plainly in the docstring *and* in gates.yaml.

**The hole.** `parse_event_classes` scopes a class body as `src[ctor_end : next_top_level_data_class.start()]`. An event-type literal declared in a **plain `class`** is therefore swallowed into the *previous* `data class`'s span, and check C then compares the contract payload against the wrong class's constructor properties.

Reproduced with a fixture: a DTO `PageRequest(page, size)` above a real producer `class AccountClosed { companion object { const val EVENT_TYPE = "foo.account.closed" } }`, with the contract documenting `page`/`size` — fields the real producer never sends:

```
OK: 1 event contract(s) agree with their producing code (...)
rc=0
```

That is exactly the defect the script's own error string exists to name ("a consumer written against this document reads a field the producer never sends"), and the count-printing defence does not save it — it prints `1 contract(s)`, confidently.

**The shape is live in the tree today.** Two producers use it:

```
openbank-customer-edge/.../onboarding/OnboardingFunnelPublisher.kt:93  const val EVENT_TYPE = "onboarding.funnel.step"
openbank-customer-edge/.../feedback/FeedbackPublisher.kt:141           const val EVENT_TYPE = "feedback.submitted"
```

customer-edge has no contract yet, so nothing is wrong *today* — but spreading contracts is the point of the issue this gate serves, and the day one lands on a service with that shape the gate goes quietly wrong instead of red. Data-class→data-class attribution is correct; I tested that variant and it reports all 8 expected disagreements. The bug is narrow: bound the body at the class's own closing brace rather than at the next `data class`.

Worth adding the "literal on a non-`data class`" case to the self-test at the same time — this is the repo's standing lesson that a self-test cannot model what its author did not imagine, and it is a cheap case to add now that it has a name.

JiRaska:
Pushed (`c82e124e5`). The reported hole is fixed, two more surfaced while fixing it, and **three findings remain that I deliberately did not fix** — they need your decision, not a patch.

## Fixed

**The class body was bounded at the next `data class`, not at its own brace.** So an event-type literal declared by a following non-data class was attributed to the preceding data class, and the contract got compared against the wrong class's properties — reported as agreement rather than a finding. The self-test gains that case and it is falsifying: against the old bounding it is the one case of seven that fails.

**`${CASE_OUTBOX_TOPIC:proposal-events}` is now resolved.** SmallRye expands env-var defaults at boot, so the committed default *is* the address published to. The raw string produced three findings against case-coordinator-agent that were one non-defect — the same topic reported as undocumented *and* unproduced simultaneously. `${VAR}` with no default is left alone: nothing committed to compare.

**A malformed contract is a finding, not a crash.** The `ScannerError` escaped and killed the run, so nothing was reported about any other contract either.

**And that crash was real.** `openbank-contracts/openbank-fraud-service/asyncapi.yaml` has an unquoted backtick-leading `description:` and **is not parseable YAML on `main` today**. Nothing could see it: `check-event-contract-coverage.py` counts contract *files* without parsing their bodies, so it reports "5 with an event contract" including one it cannot read. Quoted here.

Net on the real tree: **5 findings → 3**, self-test 7/7.

## Not fixed, and why

The three survivors are one shape:

```
message `CaseProposal`     declared in the contract, no class declares that event type
message `EngagementEvent`  same
message `FraudHoldChanged` same
```

None is a stale contract. All three services use the **second producer idiom** — the outbox row is built inline, `OutboxMessage(eventType = "fraud.hold_changed", payload = """{...}""")`, with no data class carrying the type. The gate indexes only data classes, so it calls a correct contract "emitted by nothing".

**I tried teaching it that idiom and it made things worse — 3 findings became 5.** The reason is the interesting part: the contract's message *name* is `FraudHoldChanged` while the wire value is `fraud.hold_changed`. The gate assumes name == literal, which holds for notification-service (`EVENT_TYPE = "NotificationOutcome"`, message `NotificationOutcome`) and does not hold for these three. So indexing the literal adds a second finding instead of clearing the first.

That is a **naming convention decision** — should an asyncapi message name equal the wire event type, or is the mapping deliberate? — and it is yours, not something to settle inside a gate fix. Three options, none of which I would pick for you:

1. Adopt one convention and make the contracts or the constants match.
2. Teach the gate the inline idiom **and** an explicit name↔type mapping in the contract (asyncapi has `messageId`/`x-` extensions for exactly this).
3. Baseline the three with reasons, so the gate ratchets from here.

Worth deciding before merge either way: the gate is `enforced`, so as it stands it reddens `main` on the three.

JiRaska:
The three findings this gate reports on its own PR look like a **conflation in the comparison**, not
drift in the contracts. Worth resolving before baselining anything, because the fix is different in
each direction.

The gate asks whether some class declares the AsyncAPI **message name** as its event type:

```
message `FraudHoldChanged` is declared in the contract but no class in
openbank-fraud-service/src/main declares that event type
(`override val eventType = "FraudHoldChanged"` or `const val EVENT_TYPE = "FraudHoldChanged"`)
```

But an AsyncAPI message name is a document identifier, and the event type is a **wire value**. In
fraud-service they are deliberately different, and the contract itself says so:

- `FraudHoldService.kt:167` — `eventType = "fraud.hold_changed"`
- `openbank-contracts/openbank-fraud-service/asyncapi.yaml:91` — *"Literal `fraud.hold_changed` —
  set by `FraudHoldService.emitHoldChanged`"*

So the producer and the contract already agree, in the field that carries the value; the message is
merely named in PascalCase, as AsyncAPI messages conventionally are.

engagement-service is a third shape again — `EngagementEventRepositoryImpl.kt:53` builds the type at
runtime as `"EngagementEvent.${event.type.name}"`, so no literal exists anywhere to match, and a
string search cannot see it. case-coordinator-agent has no `eventType = "…"` literal at all.

Three services, three encodings. That is the same class as the `check-*.py|sh` blind spot found on
#4083 today: a detector whose subject set is a naming convention reports PASS about everything the
convention does not cover, and FAIL about everything that encodes the same fact differently.

Two suggestions, both cheap:
1. Compare against the value the contract states (the `const`/description literal, or a dedicated
   `x-event-type`), not against the message name — then fraud-service passes for the right reason
   rather than by exemption.
2. Feed the gate a producer that genuinely renamed its wire value and confirm it goes RED. Right
   now every one of its three findings is arguably a false positive, so its detection half is
   unfalsified even though its self-test has 7 green cases — those exercise the parser, not this
   comparison.

I have not touched the branch: `/private/tmp/ob-1916-31153` is a live worktree.

PR #3932, opened 2026-08-06
feat(copilot): register workflow liveness on conversation retention scheduler (ADR-0237)

JiRaska:
## What

Registers the ADR-0237 liveness heartbeat on `ConversationRetentionScheduler` (copilot's daily
conversation-retention sweep), plus a test that covers the success, failure and disabled paths.

## Why this was missed

The scheduler landed in #3875 (`227fd95a3`) with no `DomainMetrics.registerWorkflowLiveness`.
ADR-0237 requires the heartbeat in the same PR that adds a domain `@Scheduled` job, and
`check-scheduler-liveness.py` did report it as a **NEW** violation (`::error`, not a baselined
warning) — but `.github/gates/gates.yaml` declares the gate `mode: advisory`, so `run-gates.py`
downgraded it to a WARN and the PR merged. The gate saw it; the mode swallowed it.

## Why it matters for this scheduler in particular

The sweep is exactly the shape the heartbeat exists for:

- it catches its own exceptions on purpose (`@Suppress("TooGenericExceptionCaught")`) so one bad
  tick cannot kill the schedule — so nothing throws;
- `deleted == 0` is silent, and is also what a healthy quiet day looks like;
- the only externally visible artifact of a working run is a log line that is not emitted when
  nothing expired.

A permanently broken sweep and a healthy one are therefore indistinguishable from outside the pod —
the same class of failure as the five schedulers in #2148/#2187 that had never run. Since the rows
it deletes are free-text chat under a GDPR retention promise (#3870, ADR-0118 §5 / ADR-0238), a
silently dead sweep means personal data staying on disk and in every base backup while the service
reports healthy.

## Implementation notes

- `recordSuccess()` is called **only** after `deleteExpired` returns — never from the `catch`, never
  on the `enabled == false` short-circuit. A heartbeat on the failure path would assert precisely the
  thing it exists to disprove.
- Registration observes `StartupEvent` (the `MandateExpiryScheduler` shape) rather than
  `@PostConstruct` (the `FinOpsAnalysisScheduler` shape). `@ApplicationScoped` is lazy, so
  `@PostConstruct` here would first run when the cron first fires — up to a day after boot for a
  daily job — leaving the gauge **absent** for that window, and absent is a different signal from
  stale.
- Workflow name `copilot-conversation-retention`, expected interval 1 day (matches the `0 30 3 * * ?`
  default cron).

## Verification

- `python3 .github/scripts/check-scheduler-liveness.py --root .` — no longer reports the file;
  compliant count 20 -> 21, baseline unchanged at 22 (copilot was never baselined, it was a new
  violation).
- `:openbank-copilot-service:build :openbank-copilot-service:quarkusBuild` green — `quarkusBuild`
  specifically, since constructor-injecting `DomainMetrics` and adding a `StartupEvent` observer are
  CDI wiring changes that unit tests cannot falsify.
- 128 tests, 0 failures, 0 errors (read from the JUnit XML, not the console).
- `ktlintCheck` + `detekt` green on their own after the test run.
- **The new test was falsified against the defect it exists to catch**: moving `recordSuccess()`
  ahead of the `try` block reddens `a swallowed sweep failure leaves the heartbeat stale`
  (`tests="3" failures="1"`), then restored to green.

## Not in scope

The gate's `mode: advisory` in `.github/gates/gates.yaml` is left as-is — 22 baselined schedulers
are still pending adoption (#3345), so flipping it to enforced now would fail every PR. That flip
belongs with the last adoption PR.

Refs #3345


JiRaska:
Review: 1 finding, 0 blocking

1. `openbank-copilot-service/.../retention/ConversationRetentionScheduler.kt:71` — `private var liveness: WorkflowLivenessRecorder? = null` is written by the thread that fires `StartupEvent` and read by the scheduler's worker thread, with no `@Volatile`. In practice Quarkus starts the scheduler after startup observers complete, so there is a happens-before edge and the worst case is bounded anyway (`liveness?.recordSuccess()` is null-safe, so a stale read costs one missed heartbeat, not an NPE). `@Volatile` is a one-word fix and removes the argument entirely — worth taking, given the gauge is the only externally visible difference between a broken and a healthy sweep.

Verified clean: `sweepExpiredConversations()` IS a `suspend fun` and `ConversationStore.deleteExpired` is `suspend` too, so the heartbeat is not asserting the liveness of a job that would HR000068 — this is not the plain-`fun` + `runBlocking` shape from #2148/#2187. `recordSuccess()` sits inside the `try`, after the awaited delete returns, and is genuinely unreachable from the `catch` and from the `if (!enabled) return` short-circuit; both negative paths are asserted, and the failure-path test is the load-bearing one. Registration hangs off `@Observes StartupEvent`, and a CDI observer method forces instantiation of the lazy `@ApplicationScoped` bean at boot — so the lens-9 laziness trap is handled, and the KDoc's reasoning about `@PostConstruct` is correct. The gauge cannot be confused with a live one: `registerWorkflowLiveness` seeds `lastSuccessEpochMillis` at `Instant.EPOCH`, so a registered-but-never-succeeded workflow reports an age of ~1.8e9 seconds, not zero — no boot-zero fourth state — and the test asserts `> 50 years` at registration and `< 5s` after a success rather than an `isNotNull()` that would pass against 1970. Metric names and the workflow tag come from the shared `WorkflowLivenessMetrics` constants on both the producer and the assertion side, so the emitter/consumer name drift that made the sentinel query an empty vector cannot recur here. Constructor is 4 params — well under detekt's `constructorThreshold: 9`. No data is written by this change (observability only), so nothing already written is wrong; no API surface, no migration, no `version.txt` edit.

JiRaska:
**Review: 2 findings, 0 blocking**

Claim matches code. `registerWorkflowLiveness` is bound to `StartupEvent` (not `@PostConstruct`, so
the gauge exists from boot rather than from the first cron fire — correct reasoning for a lazy
`@ApplicationScoped` bean with a daily cron), `recordSuccess()` sits inside the `try` immediately
after `deleteExpired` returns and nowhere else, and the three tests cover success / swallowed
failure / disabled with the failure case asserting the heartbeat stays stale rather than merely
existing.

**1. A deliberately disabled sweep is indistinguishable from a dead one, forever.**
`ConversationRetentionScheduler.kt` — with `copilot.retention.conversation.enabled=false` the gauge
is still registered at startup and never recorded, so `workflow_last_success_age_seconds` climbs
without bound and whatever staleness rule consumes it fires permanently. The third test pins this
as intended behaviour. For a GDPR retention sweep that is arguably the right choice — switching it
off *should* be loud — but as written it produces an alert that can never be resolved by anything
except re-enabling, which is how this estate accumulated resting-critical alerts before. Either
skip registration when `enabled == false` (absent, which the KDoc already argues is a distinct
signal) or say in the KDoc that a disabled sweep alerts by design so the next person to see it does
not silence the rule.

**2. Nothing proves the heartbeat moves in a real pod.**
Both liveness tests call `sweepExpiredConversations()` directly. That is fine for the
`recordSuccess`-placement question they are asking, but it means the whole chain still rests on the
cron actually firing — and a direct call supplies the very Vert.x context the scheduler does not,
which is the documented reason five schedulers in this fleet had never run. `sweepExpiredConversations`
is already a `suspend fun` and this PR does not touch the scheduling, so there is no defect here;
the point is scope: after this PR a permanently-not-firing schedule still looks exactly like a
disabled one (finding 1), and the only artifact that would distinguish them is a `@TestProfile`
driving the real cron with a shrunken expression, as `StandingOrderExecutionSweepIT` /
`LedgerSchedulerVertxContextIT` do. Worth a follow-up rather than this PR.

Note on the root cause, not a finding: the PR body correctly identifies that
`check-scheduler-liveness.py` *did* flag #3875 as a new violation and `gates.yaml`'s
`mode: advisory` downgraded it to a warning. Fixing this one instance leaves that mechanism intact,
so the next scheduler lands the same way. An issue to flip the gate to enforced once the 22-entry
baseline is drained would be the durable half of this change.

Verified clean: `recordSuccess()` is unreachable from the `catch` and from the disabled
short-circuit; `WORKFLOW_NAME` / `EXPECTED_INTERVAL` are `companion object` constants (no
`MagicNumber` exposure) and the 1-day interval matches the `0 30 3 * * ?` default cron; the
constructor is 4 parameters, well under detekt's `constructorThreshold: 9`; the test's workflow-name
constant is a literal rather than a shared reference, and the classloader/companion-object hazard
does not apply since there is no `QuarkusTestProfile` here; the file keeps its AGPL-3.0-only header
matching the rest of copilot-service. `quarkusBuild` being run explicitly is the right check for a
change that adds a constructor injection and a CDI observer.

---
Automated agent review — not an approval.

PR #3951, opened 2026-08-07
security(governance): promote shared_m2m_write_prohibition to a top-level key

JiRaska:
## What this changes

`rest.rego:482` reads `data.rules.shared_m2m_write_prohibition.reasons`. The key sat under `change_requirements:`, and `gen-rules-opa-data.py` selects only **top-level** keys of `rules.yaml` into `rules-opa-data.yaml` — the derived file OPA actually mounts. So the register was undefined in-cluster: `role_only_write_reason` never true, `prohibited`'s `every` clause always false, and the rule that mitigates **GHSA-58jq-9hq3-66jr** had never denied anything.

Measured on `origin/main` before the change:

```
shared_m2m_write_prohibition top-level in rules.yaml : False
                    under change_requirements:       : True
occurrences in derived rules-opa-data.yaml           : 0     <- what OPA reads
referenced by rest.rego                              : line 482
```

and after:

```
present in rules-opa-data.yaml : True
reasons emitted                : operator-consent-write, operator-ledger-write,
                                 operator-balance-write, operator-fraud-write,
                                 operator-sca-write, operator-sepa-payment-write
```

The block moves **verbatim** — `reasons` (6) and `deferred` (15) both preserved, `adr: ADR-0034` kept with it. Verified structurally, not by eye: parsing both revisions and comparing shows the only difference is the key's location, with `rest of file identical: True`.

`UNRESOLVABLE_DECLARED` in `gen-rules-opa-data.py` is emptied in the same commit — the generator refuses to pass otherwise (`::error:: … is declared unresolvable but IS now selectable — the debt is paid`), which is exactly the both-directions check that stopped this becoming permanent. The registry itself stays, so a *new* referenced-but-unselectable key still fails.

## Blast radius — measured, and narrower than the declaration assumed

The retired declaration warned that "128 currently-reachable writes depend on the very reasons the register names, so promotion denies live money-path callers". That is the right instinct but it over-states what this change can do, because `prohibited` requires **every** admitting reason to be in the register — and `matrix-allows` is not, and cannot be.

So the guard can only bite where `matrix-allows` does **not** co-admit. Checking each active reason's domain against `authz.role_action_matrix.ROLE_OPERATOR.grant`:

| reason | matrix grants non-read actions? | effect of this change |
|---|---|---|
| operator-consent-write | none | **guard now fires** |
| operator-sca-write | none | **guard now fires** |
| operator-sepa-payment-write | none | **guard now fires** |
| operator-ledger-write | `ledger.create/replay/reverse/trigger` | inert — `matrix-allows` co-admits |
| operator-balance-write | `balance.credit/debit/hold/holdRelease/initialize/reconciliation.run` | inert — same |
| operator-fraud-write | `fraud.score` | inert — same |

**Three of six become live; three remain inert.** For the three that fire, the register's own entry criterion is that the service already carries a rule naming the caller and enumerating its actions, so denying the role-only path removes an over-grant rather than the only path — `service-consent-m2m`, `service-sca-edge-m2m` / `service-sca-shared-client-m2m`, and `service-sepa-payment-shared-client-m2m` respectively. That is the claim to check hardest before merging; it is the whole safety argument.

## Why the green test suite is not evidence

`opa test` reports **168/168 PASS** — before and after, and it would pass either way. `rest_test.rego` supplies the register it tests against:

```rego
with data.rules.shared_m2m_write_prohibition.reasons as ["operator-ledger-write"]
```

so it is green about a rule given data production has never had. The generator's own header says this. The evidence that this change works is the derived artifact, not the test run.

## Regeneration

`gen-rules-opa-data.py` re-run, then every `gen-*opa-bundle*.sh`, as the last step before pushing. 72 files: the two governance artifacts, the generator, and 69 bundle/annotation files across the fleet. Verified **idempotent** (second run produces no further diff).

## Not merged deliberately

This is governance + authz on money paths. It is also a **policy decision**, as the retired declaration said — it starts denying callers on three domains. Leaving it for human review rather than arming auto-merge.

Refs #3765, Refs #3734. Related: #3812 argues `shared_m2m_write_prohibition` cannot address the `matrix-allows` layer — that is correct and this change does not claim otherwise; the two are complementary, and the table above marks exactly where this one stops.

### DIFF
--- a/.github/scripts/gen-rules-opa-data.py
+++ b/.github/scripts/gen-rules-opa-data.py
@@ -206,14 +206,12 @@ def build() -> tuple[str, list[str]]:
 #
 # So: today's one occurrence is declared with its reason and its issue, and anything NEW fails. A
 # declaration that becomes resolvable is reported too, so this cannot quietly become permanent.
-UNRESOLVABLE_DECLARED: dict[str, str] = {
-    "shared_m2m_write_prohibition": (
-        "nested under `change_requirements:` in rules.yaml, so it cannot be selected as a top-level "
-        "key. Promoting it is a POLICY DECISION, not an edit: 128 currently-reachable writes depend "
-        "on the very reasons the register names, so promotion denies live money-path callers. "
-        "Tracked in #3765/#3734 — remove this entry in the same change that promotes the key."
-    ),
-}
+# Emptied 2026-08-07: `shared_m2m_write_prohibition` was promoted to a top-level key in the same
+# change, so it is now selected into rules-opa-data.yaml and the declaration above it is history
+# rather than debt. The registry stays in place — a NEW referenced-but-unselectable key still
+# fails, and a declaration that becomes resolvable is still reported, which is what forced this
+# entry to be removed here rather than left to rot.
+UNRESOLVABLE_DECLARED: dict[str, str] = {}
 
 
 def report_missing(missing: list[str]) -> int:
--- a/openbank-infra/gitops/components/accounts/account-opa-bundle.yaml
+++ b/openbank-infra/gitops/components/accounts/account-opa-bundle.yaml
@@ -9,7 +9,7 @@ metadata:
     app.kubernetes.io/name: account-service
     app.kubernetes.io/part-of: accounts
   annotations:
-    openbank.tech/policy-checksum: "bb5f1cb65aa5c2c3"
+    openbank.tech/policy-checksum: "6f67742d3aba3c74"
 data:
   rest.rego: |
     # SPDX-License-Identifier: Apache-2.0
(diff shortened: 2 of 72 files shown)


JiRaska:
## Correction to the blast-radius table in the PR description — do not act on the original

The table I published is **wrong**, and wrong in the direction that matters. My matrix check filtered on the prefixes `sca.` and `sepa.`, but the action names are `scaChallenge.*`, `device.*` and `sepaPayment.*`, so both domains were mis-scored. Re-measured against `authz.role_action_matrix.ROLE_OPERATOR.grant` on `origin/main`:

```
consent.*                    -> NONE
scaChallenge.* / device.*    -> device.list  (only)
sepaPayment.*                -> sepaPayment.approval.decide, .create,
                                .handleReturn, .transitionStatus
```

Corrected effect of promoting the key:

| reason | matrix co-admits? | corrected effect |
|---|---|---|
| operator-consent-write | no | **guard fires** |
| operator-sca-write | only `device.list` | **guard fires** for every `scaChallenge.*` write and `device.enroll` |
| operator-sepa-payment-write | yes — the four main actions | **inert** (I previously said it fires) |
| operator-ledger-write | yes | inert |
| operator-balance-write | yes | inert |
| operator-fraud-write | yes | inert |

So sepa-payment is safer than I said, and **sca is the exposure**.

## Why sca is the problem

`sca-service` declares seven actions:

```
device.enroll, device.list, scaChallenge.consume, scaChallenge.decide,
scaChallenge.initiate, scaChallenge.read, scaChallenge.verify
```

The identity-scoped rule for the shared client covers exactly one of them:

```rego
allowed_reasons contains "service-sca-shared-client-m2m" if {
	input.principal.id == "service-account-openbank-services"
	input.action == "scaChallenge.read"
}
```

`device.list` is additionally covered by `matrix-allows`. Everything else — `scaChallenge.consume`, `.verify`, `.initiate`, `.decide`, `device.enroll` — has **no alternative admitting reason** for `service-account-openbank-services`. After promotion the only reason admitting those is `operator-sca-write`, which is in the register, so `every` holds and `prohibited` fires: **403**.

Six services declare a REST client against sca: account, consent, customer-edge, delegation, document, security-scanner. `sca-service.yaml` in gitops runs `AUTHZ_ENFORCE=true`, as do consent and payments — so there is no enforcement-is-off cushion. Any of those callers invoking an sca challenge action via the shared client breaks **on merge**, and SCA gates payments.

A caller sweep mapping each client to its action and principal is running; I will post the result. Until then this PR should be treated as **not ready to merge** — the register's safety claim ("each entry was verified to have a rule that names the caller and enumerates its actions") does not hold for `operator-sca-write` against the shared client, whatever it may hold for the edge principal.

Note the interaction with **#3748**, which grants the shared client `scaChallenge.consume`: if a live caller uses that action, #3748 is a prerequisite for this PR, not an independent change.

I am leaving the original table in the description rather than silently rewriting it, so the record shows what was published and what corrected it.

JiRaska:
## Caller verification done — one real break found and removed from this PR

The register's safety claim was checked against actual callers rather than taken at face value, and it does not hold for sca. Fixed in `65d9874`.

### The break

`sca-service` exposes `device.enroll`, `device.list`, `scaChallenge.consume`, `.decide`, `.initiate`, `.read`, `.verify`. For the shared client, `service-sca-shared-client-m2m` admits **`scaChallenge.read` only**, and `matrix-allows` covers only `device.list`. Everything else has no alternative admitting reason.

**document-service calls the uncovered one.** `openbank-document-service/.../client/ScaChallengeClient.kt`:

```kotlin
@RegisterRestClient(configKey = "sca-service")
@RegisterProvider(OidcClientRequestReactiveFilter::class)   // attaches the M2M token
…
@POST
@Path("/api/v1/sca/challenges/{id}/consume")
```

with `quarkus.oidc-client.client-id: openbank-services` — so it presents `service-account-openbank-services` — and `components/sca/sca-service.yaml` runs `AUTHZ_ENFORCE=true`. Listing `operator-sca-write` would have made every document-signing consume a 403 on merge. The client's own KDoc even names the gate: *"a real, live endpoint — `ScaResource.consume` — gated `@Authorize(action = "scaChallenge.consume")`"*.

`operator-sca-write` is therefore moved to `deferred` with that evidence. It can move back once **#3748** grants the shared client `scaChallenge.consume` — the two PRs have a hard ordering dependency, and it runs #3748 → #3951, not the other way round.

### The other two, resolved

**consent — genuinely verified, stays promoted.** Callers traced: party-service's marketing forwarder (`consent.grant` / `consent.revoke`, covered by `service-consent-m2m-marketing`, and it passes the exact `granteeId` the rule's `resource` binding expects), psd2 and mcp (`consent.validate`), campaign and notification (`consent.validate` — note that endpoint declares `validate`, not a read), customer-edge (edge principal, different rule). No caller of `consent.activate`/`consent.reject` in-repo, and both are covered anyway.

**sepa-payment — inert, but for a reason the register did not state**, so it is now annotated rather than silently relied upon. `service-sepa-payment-shared-client-m2m` covers `sepaPayment.handleReturn` alone; what actually keeps `standing-order`'s `sepaPayment.create` working is that `ROLE_OPERATOR` also grants `create` / `transitionStatus` / `approval.decide`, so `matrix-allows` co-admits and this reason never becomes the only one. That is a shadow-phase artifact scheduled for retirement under ADR-0223 D2(b) — the day those grants go, this entry starts denying a live money path. Recorded in place, because a safety property that holds by accident and is documented as holding by design is exactly how the next one of these gets missed.

### Net effect of the PR now

`prohibited` becomes live for **consent only**. ledger, balance, fraud and sepa-payment are all co-admitted by `matrix-allows` and stay inert. That is a smaller change than the title suggests — but it is the first time the guard is capable of firing at all, and the register is now honest about which entries are load-bearing and which are coasting.

`opa test` still reports 168/168, which as noted earlier proves nothing here: the tests inject their own register. The evidence is the emitted `rules-opa-data.yaml`, which now carries five reasons and not `operator-sca-write`.

JiRaska:
## Rebased onto the fixed `main`; `operator-sca-write` restored

Sequence completed: **#3748** (grants the shared client `scaChallenge.consume`) → **#3971** (repairs the bundle #3748 shipped unparseable) → this PR.

`operator-sca-write` is back in `reasons`. The reason it was deferred is gone: `service-sca-shared-client-m2m` now admits the shared client for `scaChallenge.consume`, so document-service's call keeps an identity-scoped admitting reason and this entry no longer 403s the document-signing gate.

The deferral commit is kept in history rather than squashed away — the record of *why* it was unsafe and what made it safe is worth more than a tidy diff.

## Verified functionally, against the bundle this branch generates

Not by reading the rego — by evaluating the real ConfigMap contents for the shared client (`service-account-openbank-services`, `ROLE_OPERATOR`):

```
scaChallenge.consume   allow=true    reason=service-sca-shared-client-m2m
scaChallenge.read      allow=true    reason=operator-read-any
device.list            allow=true    reason=matrix-allows
scaChallenge.decide    allow=false   allowed_reasons=[]
```

`decide` is denied by *absence* — no rule admits it at all — which is unchanged, fail-closed behaviour, not an effect of the register. `decide`/`initiate`/`verify` and `device.enroll` remain uncovered for the shared client; no in-repo caller invokes them on that principal.

## Where the guard now actually bites

With all six reasons emitted, `prohibited` can fire only where no other reason co-admits:

| reason | co-admitted by | live? |
|---|---|---|
| operator-consent-write | nothing | **yes** |
| operator-sca-write | `service-sca-shared-client-m2m` (read, consume), `matrix-allows` (device.list) | narrowed to the uncovered actions |
| operator-sepa-payment-write | `matrix-allows` | no |
| operator-ledger-write | `matrix-allows` | no |
| operator-balance-write | `matrix-allows` | no |
| operator-fraud-write | `matrix-allows` | no |

The `operator-sepa-payment-write` annotation added earlier still stands and still matters: it is inert only because `ROLE_OPERATOR` grants `sepaPayment.create`/`.transitionStatus`/`.approval.decide`, and standing-order POSTs `sepaPayment.create` through the shared client. When ADR-0223 D2(b) retires those matrix grants, that entry starts denying a live money path.

## Housekeeping

All three commits signed and `verified: true`. Regeneration is idempotent (second run produces no diff). `gen-rules-opa-data.py`'s `UNRESOLVABLE_DECLARED` registry stays in place but empty, so a new referenced-but-unselectable key still fails the gate.

Still deliberately not merged: governance + authz on money paths, and it is a policy decision.

PR #3956, opened 2026-08-07
fix(settlement): dial balance/ledger/party at their real in-cluster Services, enforce the gate

### LINKED ISSUE #3931: 8 gates are advisory: two find nothing and can graduate free, and scheduler-liveness-adoption already let a violation through
## Summary

Of 111 declared gates, **8 are `mode: advisory`** — they detect and block nothing. Two of them find
nothing at all on a full-tree scan today and could graduate to `enforced` for free; the rest are
advisory for reasons worth stating rather than leaving implicit.

Measured on `origin/main`:

| gate | scope | result today |
|---|---|---|
| `dotted-mp-messaging-key-guard` | whole tree | **PASS** |
| `governance-lineage-vs-code-audit` | whole tree | **PASS** |
| `scheduler-liveness-adoption` | whole tree | WARN — live findings |
| `operator-write-naming` | whole tree | WARN — live findings |
| `incluster-hostname-resolution` | whole tree | WARN — live findings |
| `release-scope-mismatch-gate` | `needs_base: required`, `when: pull_request` | WARN (diff-scoped) |
| `db-migration-gate` | `needs_base: required`, `when: pull_request` | PASS — **vacuous here** |
| `schema-compat-gate` | `needs_base: required`, `when: pull_request` | PASS — **vacuous here** |

**The last two PASSes prove nothing** and I am flagging them rather than counting them: they are
diff-scoped, and I ran them with no meaningful diff. A diff gate with an empty diff always passes.
Only the two whole-tree PASSes are real.

## The two that can graduate for free

`dotted-mp-messaging-key-guard` and `governance-lineage-vs-code-audit` scan the whole tree and find
nothing. Flipping them to `enforced` costs nothing today and stops the next occurrence at PR time
instead of in a later audit.

## The three whole-tree WARNs are the interesting ones

Each has live findings, so graduating them would go red immediately — which is exactly why each needs
a decision rather than a flip:

- **`scheduler-liveness-adoption`** — 32 adoption-pending schedulers. **This one already cost
  something today:** #3875 merged a new `ConversationRetentionScheduler` with no heartbeat; the
  ratchet's `::error` branch fired and `mode: advisory` turned it into exit 0, so nothing stopped it.
  A ratchet whose forward half does not block is a ratchet in name only — the whole point is that
  *new* violations fail while the baseline is grandfathered.
- **`operator-write-naming`** — deliberately advisory pending #3765/#3734, and correctly so: enforcing
  today would fail on the live finding rather than on a regression.
- **`incluster-hostname-resolution`** — findings not triaged here.

## Suggested

1. Graduate the two clean whole-tree gates.
2. For `scheduler-liveness-adoption`, split the mode from the baseline: grandfather the 32, enforce on
   anything new. That is what a ratchet is, and it is the difference between #3875 being caught and
   being merged.
3. Leave `operator-write-naming` advisory with its existing reason, and record it — the reason is
   sound but currently lives only in a comment.

## Not verified

- I did not triage `incluster-hostname-resolution`'s findings, so I cannot say whether they are real
  or baseline noise.
- The three diff-scoped gates were not exercised against a real diff, so this says nothing about
  whether they work — only that today's run is uninformative about them.
- Whether an "advisory" gate is advisory via `mode:` or via an `--enforce` flag inside its own script
  is a separate axis I did not audit; 11 of 12 gates in an earlier sweep were advisory *inside* the
  script, so a `mode:` census may undercount.

Refs #3875, #3765, #3734, ADR-0237

JiRaska:
Closes #3931

## What this is

The `incluster-hostname-resolution` gate (added advisory in #1916's follow-up) reported 6 findings. Its triage note in both `gates.yaml` and `rules.yaml` graded all six as dead defaults "on services whose pods override them by env".

That was true of **three**. The other three were live.

## Verification (2026-08-07, sandbox cluster, `origin/main` @ `98a385f68`)

`kubectl get svc -A` — no Service is named `openbank-balance-service`, `openbank-ledger-service`, `openbank-party-service`, `openbank-sanctions-service` or `openbank-product-catalog` in any namespace. The real ones:

| claimed default | real Service | real port |
|---|---|---|
| `openbank-balance-service:8080` | `balance-service` (ns `balances`) | 8103 |
| `openbank-ledger-service:8080` | `ledger-service` (ns `ledger`) | 8101 |
| `openbank-party-service:8090` | `party-service` (ns `party`) | 8111 |
| `openbank-sanctions-service:8123` | `sanctions-service` (ns `sanctions`) | 8123 |
| `openbank-product-catalog:8080` | `product-catalog` (ns `accounts`) | 8104 |

Then, per service, whether the workload overrides the default:

| finding | override present? | verdict |
|---|---|---|
| account-service -> sanctions | `SANCTIONS_SERVICE_URL` on the Deployment | dead default |
| account-service -> product-catalog | `PRODUCT_CATALOG_SERVICE_URL` (+ `_HOST_OVERRIDE`) | dead default |
| card-issuance -> product-catalog | `PRODUCT_CATALOG_SERVICE_URL` (+ `_HOST_OVERRIDE`) | dead default |
| **settlement -> balance** | **none** | **live** |
| **settlement -> ledger** | **none** | **live** |
| **onboarding -> party** | **none** | **live** |

`kubectl -n payments get pod -l app.kubernetes.io/name=settlement-service` carries only `QUARKUS_DATASOURCE_*`, `QUARKUS_OIDC_AUTH_SERVER_URL`, `OPENBANK_TEMPORAL_SERVER_URL`, `SETTLEMENT_GL_*` — no `*_SERVICE_URL`. `kubectl -n onboarding get pod -l app.kubernetes.io/name=onboarding-service` carries no `*_SERVICE_URL` at all.

The clients on the live side are live code: `BalanceDebitAdapter`, `BalanceCreditAdapter`, `LedgerBookAdapter` (all `@RestClient`), and `AbandonedRegistrationCleaner`'s `PartyServiceClient`. So **every settlement debit/credit and every settlement GL posting, plus onboarding's abandoned-registration party lookup, dialled a hostname that resolves nowhere.**

The `LEDGER_SERVICE_URL` / `BALANCE_SERVICE_URL` pair that does exist in `payments-services.yaml` belongs to **transaction-service**, not settlement — that is the shape that makes this easy to misread from a grep.

## The fix

Follows the fleet convention rather than inventing one — a localhost dev default in `application.yaml`, the in-cluster URL from the workload env in gitops. transaction, billing, statement, lending, fx, kyc, sepa-payment, document-service all already do this; the four services here were the outliers.

- **settlement** (`fix`): `BALANCE_SERVICE_URL` / `LEDGER_SERVICE_URL` added to the Rollout, values copied from transaction-service; defaults to `localhost:8103` / `localhost:8101`.
- **onboarding** (`fix`): `PARTY_SERVICE_URL` added to the Deployment; default to `localhost:8111`.
- **account** and **card-issuance** (`refactor`): defaults corrected to `localhost:8123` / `localhost:8104`. No deployed behaviour changes, so deliberately not a releasing type — these only fix local dev and clear the gate.

NetworkPolicies are derived from those env URLs, so `gen-network-policies.py` was re-run: party's ingress allow-list gains `onboarding` (one 3-line hunk). `payments` was already permitted to `balances` and `ledger`, so nothing there changed.

Note the ordering property: the env override lands with the ArgoCD sync, so **the live fix does not wait for a settlement/onboarding image rebuild** — the `application.yaml` half only matters for local dev and for the day someone drops the env.

## Gate graduation

`mode: advisory` -> `enforced` (ahead of its ADR-0144 target of 2026-10-31), with both copies of the triage note corrected to what was measured. `rules.yaml` keeps the lesson rather than deleting it with the findings:

> an advisory gate's "these are all benign" triage is itself an unverified claim, and advisory mode is exactly what removes the pressure to check it.

Grading a finding by the **shape** of the config line instead of by the deployed manifest gets the benign cases right and the live ones wrong, silently. Same class as "a gate that has only ever passed is unfalsified" one level up: here the gate fired correctly and the *triage* was the unfalsified artifact.

**Falsified against a real counter-example, not only the fixture harness**: re-introducing `openbank-ledger-service:8080` into settlement's `application.yaml` on this branch takes the gate from exit 0 to exit 1 with exactly that one finding, and reverting returns it to 0. `--self-test` is 19/19.

`rules-opa-data.yaml` is unchanged — no `.rego` reads `data.rules.incluster_hostnames` — so no OPA bundle restamps.

## Money-path

settlement and account are `rules.yaml: money_path_services`, so **2 approvals** are required. The enforced `threat-model-updated-on-trust-boundary-change` gate fired on both; both threat models are updated in this PR and are explicit that **no trust boundary moved** — same edges, same services, same NetworkPolicies, so no STRIDE row changes. The settlement finding is availability, not confidentiality or integrity: the requests never left the pod.

## ASVS baseline

`V9.1` counts a new plaintext `http://` inter-service env value as new debt, and onboarding -> party genuinely is one (a dead edge becoming a live one). Declared in `.github/asvs-l3-baseline.txt` with the reason; deleting it is the mesh-mTLS work (security baseline Layer 1), not a per-service change. The settlement additions needed no entry — V9.1's identity is `(file, value)` with the line fragment normalised away, and `payments-services.yaml` already carries both URLs.

## Local verification

`run-gates.py --all`: 112 gates, **109 PASS**, 2 WARN, 1 FAIL.

- Both WARNs are advisory and pre-existing on `main` (`release-scope-mismatch` needs `PR_TITLE`; `operator-write naming` flags `rest.rego`'s `operator-mcp-session`, untouched here).
- The FAIL is `api-contract-gate`, which cannot run locally (`sudo: a password is required` installing `oasdiff`). No `openapi.yaml` is in this diff, so there is nothing for it to classify.
- `loki-rule-load-test` needed `BASE_REF` and passes once it is set.

### DIFF
--- a/.github/asvs-l3-baseline.txt
+++ b/.github/asvs-l3-baseline.txt
@@ -205,6 +205,12 @@ V9.1 openbank-infra/gitops/components/billing/billing-service.yaml:70: plaintext
 V9.1 openbank-infra/gitops/components/billing/billing-service.yaml:78: plaintext http:// env value http://ledger-service.ledger.svc:8101
 V9.1 openbank-infra/gitops/components/billing/billing-service.yaml:97: plaintext http:// env value http://tempo.observability.svc:4317
 V9.1 openbank-infra/gitops/components/onboarding/onboarding-service.yaml:134: plaintext http:// env value http://tempo.observability.svc:4317
+# onboarding -> party-service (issue #3931). Genuinely a NEW plaintext edge and declared as
+# such: the rest-client existed but its URL default named `openbank-party-service`, a hostname
+# no Service has, so the call never left the pod. Making it resolve turns a dead edge into a
+# live one, on the same plaintext in-cluster transport as the other 200+ lines here — deleting
+# this line is the mesh-mTLS work (baseline Layer 1), not a per-service change.
+V9.1 openbank-infra/gitops/components/onboarding/onboarding-service.yaml:138: plaintext http:// env value http://party-service.party.svc:8111
 V9.1 openbank-infra/gitops/components/analytics/analytics-sink.yaml:79: plaintext http:// env value http://clickhouse.analytics.svc:8123
 # analytics-sink -> Apicurio schema registry (issue #1916). Same declared debt as every other
 # in-cluster edge here: the Service exposes exactly one port, named `http`, on 8080
--- a/.github/gates/gates.yaml
+++ b/.github/gates/gates.yaml
@@ -431,14 +431,20 @@ gates:
   # "schema-registry:8081" for the life of the service; the Service is apicurio-registry:8080 in
   # `messaging` (issue #1916). Needs no network: the resolvable set is fully declared in gitops.
   #
-  # ADVISORY (rules.yaml: incluster_hostnames, target_enforce_date 2026-10-31) — 6 pre-existing
-  # findings, all bare `openbank-<svc>` rest-client defaults on services whose pods override them
-  # by env. Four of the owners are money-path, so that cleanup takes the two-approval path rather
-  # than riding along with the guard that found it.
+  # ENFORCED since #3931, which drained the 6 pre-existing findings. The advisory note that stood
+  # here claimed all six were dead defaults "on services whose pods override them by env" — that
+  # was true of only THREE (account-service ×2, card-issuance ×1). The other three were live:
+  # settlement-service's Rollout and onboarding-service's Deployment carried no *_SERVICE_URL at
+  # all, so `openbank-balance-service:8080` / `openbank-ledger-service:8080` /
+  # `openbank-party-service:8090` were the EFFECTIVE values in the cluster — names that resolve
+  # nowhere, on wrong ports (8103 / 8101 / 8111). The triage note was written from the shape of
+  # the findings rather than from the deployed manifests, and being advisory is what let it sit.
+  # Generalize: an advisory gate's "these are all benign" note is itself an unverified claim, and
+  # advisory mode removes the pressure that would have checked it.
   - id: incluster-hostname-resolution
-    name: "in-cluster hostnames resolve to a declared Service (issue #1916, advisory)"
+    name: "in-cluster hostnames resolve to a declared Service (issue #1916, enforced)"
     group: gitops
-    mode: advisory
+    mode: enforced
     selftest: |
       python3 .github/scripts/check-incluster-hostnames.py --self-test
     selftest_expect: pass
--- a/docs/threat-models/openbank-account-service.md
+++ b/docs/threat-models/openbank-account-service.md
@@ -86,6 +86,21 @@ not change any existing request's outcome until explicitly flipped.
 
 ## 6. Change log
 
+- **2026-08-07** — No trust boundary moved: the `sanctions-service` and `product-catalog`
+  rest-client **defaults** in `application.yaml` were changed from `http://openbank-sanctions-service:8123`
+  and `http://openbank-product-catalog:8080` to `http://localhost:8123` / `http://localhost:8104`
+  (issue #3931). Neither `openbank-` name is a Service in any namespace, and the product-catalog
+  port was wrong as well (the Service is `product-catalog:8104` in `accounts`). **These defaults
+  were dead in every deployed environment** — the account-service Deployment sets
+  `SANCTIONS_SERVICE_URL` and `PRODUCT_CATALOG_SERVICE_URL` (the latter at the KEDA HTTP
+  interceptor, with `PRODUCT_CATALOG_API_HOST_OVERRIDE`) — so no deployed request path changes,
+  and both edges keep their existing postures (sanctions fails **closed**, product-catalog fails
+  **open**, §6 2026-07-09 and 2026-06-06 entries). The change matters for local dev, where the
+  screening gate previously failed closed against a name that could never resolve, and it clears
+  the last findings blocking `incluster-hostname-resolution` from `mode: advisory` to `enforced`.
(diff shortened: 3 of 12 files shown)


JiRaska:
Closing as superseded by #3974 — not abandoned. Every commit here is already in that branch.

**Verified rather than assumed**, since "superseded" is exactly the claim that is cheap to get wrong:

```
git merge-base --is-ancestor origin/fix/incluster-hostname-resolution-3931 \
                            origin/feat/incluster-hostnames-gitops-env-3966   # -> YES
```

and the two load-bearing changes are intact on the survivor:

- `.github/gates/gates.yaml` — `incluster-hostname-resolution` is `mode: enforced`
- `openbank-libs/governance/rules.yaml` — `enforced: enforced`, `enforced_since: "2026-08-07"`

Five files read as differing between the two branches (`asvs-l3-baseline.txt`, `gates.yaml`, `openbank-account-service.md`, `party/network-policies.yaml`, `payments-services.yaml`). That is #3974 carrying **more** — its own changes plus what arrived when `main` was merged in (#3987's ADR-registry gate, #3104's change-log entry) — not anything from here being dropped. The ancestor check is what settles it; a file-by-file shasum comparison alone would have looked ambiguous.

## Why this one went DIRTY

#3968 squash-merged. #3974 carried #3968's commits, so the same content arrived on `main` under a different SHA — which is what makes a stacked child conflict, and #3956 touches the same `payments-services.yaml` and `party/network-policies.yaml`. Resolving it here would be duplicated work on content that lands via #3974 regardless.

## What still lands

Everything: the settlement and onboarding in-cluster hostname fixes, the account-service and card-issuance dead-default cleanups, the gate graduation to `enforced`, the corrected triage note in both `gates.yaml` and `rules.yaml`, and the two money-path threat-model updates. #3974 carries `Closes #3931`.

The branch is left in place rather than deleted, so this PR's diff stays readable.

PR #3968, opened 2026-08-07
fix(vop): correct both payee-name hop URLs — ports were transposed, namespace never existed

### LINKED ISSUE #3966: vop-service: both ADR-0171 payee-name hops point at non-existent host/port (namespace `parties`, transposed ports)
## Summary

Both hops of vop-service's ADR-0171 payee-name lookup point at hostnames/ports that cannot serve them. Verified on the live sandbox pod on 2026-08-07, not only in the manifest.

`openbank-infra/gitops/components/payments/payments-services.yaml:2980-2983` (the `vop-service` Rollout):

| var | declared | correct | wrong how |
|---|---|---|---|
| `ACCOUNT_SERVICE_URL` | `http://account-service.accounts.svc:8101` | `…accounts.svc:8100` | port — 8101 is ledger-service's |
| `PARTY_SERVICE_URL` | `http://party-service.parties.svc:8100` | `http://party-service.party.svc:8111` | namespace AND port |

The two ports look transposed, and the namespace pluralised.

## Evidence

```
$ kubectl get ns parties
Error from server (NotFound): namespaces "parties" not found

$ kubectl -n payments get pod -l app.kubernetes.io/name=vop-service \
    -o jsonpath='{range .items[0].spec.containers[0].env[*]}{.name}={.value}{"\n"}{end}' | grep -iE 'party|account'
ACCOUNT_SERVICE_URL=http://account-service.accounts.svc:8101
PARTY_SERVICE_URL=http://party-service.parties.svc:8100

$ kubectl -n accounts get svc account-service -o jsonpath='{.spec.ports[*].port}'   # 8100
$ kubectl -n party    get svc party-service   -o jsonpath='{.spec.ports[*].port}'   # 8111
$ kubectl -n ledger   get svc ledger-service  -o jsonpath='{.spec.ports[*].port}'   # 8101
```

The comment directly above the declaration states the intent, which is what makes the values readable as wrong rather than deliberate:

> ADR-0171 §4: the payee name is two hops away — account-service resolves IBAN→partyId, party-service resolves partyId→legal/trading name.

Neither hop can complete. vop-service is money-path (`rules.yaml:742`, IPR Art. 5c), and a VoP that cannot resolve a name is the case where "could not verify" and "verified" must not be conflated — see the existing concern that `.isValid` over-claims.

## Why nothing caught it

`incluster-hostname-resolution` (enforced as of #3956, issue #3931) reads `openbank-*/src/main/resources/application.yaml` **only**, so a hostname living in a gitops workload env is out of its scope entirely. That is the same class the gate exists for, one layer over — and it is the layer the fleet-standard fix moves URLs *into*. The other layers are blind by construction: a unit test stubs the client, an IT serves localhost, a consumer pact answers whatever path it is asked for.

## Suggested fix

1. Correct both env values in `payments-services.yaml`, then re-run `openbank-infra/scripts/gen-network-policies.py` — the allow-lists are derived from these URLs, so party's and accounts' ingress edges for `payments` are currently derived from the wrong ones.
2. Confirm end-to-end from a pod in `payments` rather than by reading the YAML back — an allow-list read as a flat namespace set loses the port dimension.
3. Worth considering as the durable fix: extend `check-incluster-hostnames.py` to the gitops env layer, where it already parses those manifests to build the *declared Service* set. It has both halves of the comparison in hand today and only checks one direction. That would close the scope gap rather than this one instance.

Found while writing up the #3931 triage lesson; not folded into #3956 or #3962, which are a different service and a docs change.

JiRaska:
Closes #3966

## The defect

ADR-0171 §4 resolves a payee name in two hops — account-service maps IBAN→partyId, party-service maps partyId→legal/trading name. **Neither hop could complete, in any environment.**

The two ports were transposed in `application.yaml` from the start, and the gitops env inherited the same swap while additionally pluralising the namespace:

| | declared | correct | wrong how |
|---|---|---|---|
| `ACCOUNT_SERVICE_URL` | `account-service.accounts.svc:8101` | `…accounts.svc:8100` | 8101 is ledger-service's port |
| `PARTY_SERVICE_URL` | `party-service.parties.svc:8100` | `party-service.party.svc:8111` | 8100 is account-service's port; namespace `parties` has never existed |
| `application.yaml` account default | `localhost:8101` | `localhost:8100` | same transposition |
| `application.yaml` party default | `localhost:8100` | `localhost:8111` | same transposition |

Because both layers carry the same swap, **there was no fallback that was right** — the env override and the default were wrong in the same direction.

Verified against the live sandbox pod, not just the manifest:

```
$ kubectl get ns parties
Error from server (NotFound): namespaces "parties" not found
$ kubectl -n payments get pod -l app.kubernetes.io/name=vop-service -o jsonpath='...'
ACCOUNT_SERVICE_URL=http://account-service.accounts.svc:8101
PARTY_SERVICE_URL=http://party-service.parties.svc:8100
$ # account-service=8100  party-service=8111  ledger-service=8101
```

## Why it matters more than a broken connection

The threat model's §3 *Denial of service* says VoP **fails open**: a lookup failure yields `no_data` + a warning, never a hold. So this is a fail-open control that was **failing open 100% of the time** — every `/verify` answered `no_data` and every payment proceeded with a warning, for the life of the service.

And it was unobservable by construction. §3 *Info disclosure* makes an unknown IBAN a **200 + `no_data`** on purpose, so the status code cannot be used to enumerate which IBANs are ours. That same collapse makes a totally broken lookup indistinguishable from a genuine miss. The thing that would have caught it is an alert on the `no_data` **rate** — which is the threat model's own §4 item 2, still open.

## Corroboration that it never worked

`gen-network-policies.py` derives ingress allow-lists from exactly these env URLs, so regenerating produces the tell: **party's allow-list gains `payments` for the first time**. The edge did not exist, so even a corrected hostname would have been dropped by the CNI until this regen. That is independent evidence — it comes from a generator, not from reading the values back.

## Why no gate saw it

- unit tests stub both clients; the IT serves localhost; a consumer pact answers whatever path it is asked
- `incluster-hostname-resolution` (enforced in #3956) reads `application.yaml` **only** and exempts `localhost` — so it can see neither the gitops env nor a wrong port on a right host

That scope gap is the third bullet added in #3967, with this as its worked example.

## ⚠️ Merge-order note — please read if merging alongside #3956

This PR overlaps #3956 on three files. Two are harmless (different regions of `payments-services.yaml` and `.github/asvs-l3-baseline.txt`), but **`openbank-infra/gitops/components/party/network-policies.yaml` is the dangerous shape**: both PRs insert a neighbouring entry into the *same* ingress list (#3956 adds `onboarding`, this adds `payments`), which is exactly the case git merges "cleanly" while keeping only one — no conflict, nothing printed, exit 0.

It is a **derived** file, so do not hand-merge. After whichever of the two lands second:

```bash
git fetch origin main && git rebase origin/main
python3 openbank-infra/scripts/gen-network-policies.py
git diff --stat
```

Then confirm both survived — this must print 2:

```bash
git show origin/main:openbank-infra/gitops/components/party/network-policies.yaml | grep -cE 'metadata.name: (onboarding|payments)$'
```

## Money-path

vop-service is `rules.yaml: money_path_services` (IPR Art. 5c), so **2 approvals**. The enforced `threat-model-updated-on-trust-boundary-change` gate fired and is satisfied: added as open item 7, framed around the fail-open consequence rather than the connection error.

## Local verification

`run-gates.py --all`: 112 gates, **108 PASS**, 3 WARN, 1 FAIL — none attributable to this change.

- `api-contract-gate` (the FAIL) cannot run locally (`sudo: a password is required` installing `oasdiff`); no `openapi.yaml` is in this diff.
- `incluster-hostname-resolution` warns with its 6 pre-existing findings because this branch is based on `main`, which does not yet have #3956's fix.
- `release-scope-mismatch` and `operator-write-naming` are advisory and pre-existing on `main`.

ASVS V9.1 ends at `0 NEW, 0 paid-off`: two baseline lines retired (the transposed pair) and one added for `party-service.party.svc:8111`. The account hop needs no line — V9.1's identity is `(file, value)` and `account-service.accounts.svc:8100` is already baselined for this file.


JiRaska:
## Independently verified — every claim in the description holds

Re-derived from `origin/main` rather than taken on trust, because "never worked in any environment" is a strong claim and this is the payee-verification path.

**Ports, read from each service's own `application.yaml`:**

```
account-service  8100
ledger-service   8101
party-service    8111
```

**VoP's config on `origin/main`:**

```
64:      url: ${ACCOUNT_SERVICE_URL:http://localhost:8101}
67:      url: ${PARTY_SERVICE_URL:http://localhost:8100}
```

So the account hop points at **ledger-service's** port and the party hop points at **account-service's** port. Transposed exactly as described.

**Namespace:** `namespace: parties` appears nowhere in the tree; the party component declares `namespace: party`, and the gitops directory is `components/party/`. So the gitops URL named a namespace that has never existed.

Both hops therefore fail — one by hitting the wrong service, one by resolving a nonexistent namespace. The two-hop IBAN → partyId → name resolution of ADR-0171 §4 could not complete anywhere.

Worth stating plainly for whoever reviews: this repo's own notes already record that VoP "confirmed a WRONG payee" and that `.isValid` was not telling the truth. This is a mechanism for that, and it is the kind of defect no unit test can see — a stubbed client answers on whatever URL it is handed, and a wrong port is only wrong against a real cluster.

This is an automated agent verification of the diagnosis, not an approval; vop is money-path and still needs human review.

PR #3974, opened 2026-08-07
feat(ci): check gitops workload env hostnames, and fix the four it finds

JiRaska:
Refs #3931, #3966

Extends `incluster-hostname-resolution` to the gitops workload env layer — the scope gap documented in #3967 — and fixes the four live defects it finds.

**Stacked on #3956 and #3968**, both merged into this branch so it is green independently. If they land first this diff shrinks to the gate plus the four fixes; if this lands first they become no-ops. Either order converges.

## Why the gate needed this

It read `openbank-*/src/main/resources/application.yaml` only. That is the layer the fleet-standard remedy **moves a host out of**: the house pattern for a bad in-cluster URL is a localhost dev default plus the real URL in the workload env. So fixing a flagged host handed the checked claim to an unchecked file. The header called it "a real hole"; it was larger than described.

## Why it is not just a second file loop

The corroboration tier — a host is believed if some gitops workload env dials it — is **derived from those same env values**. Reusing it here would be circular: every env host would corroborate itself and nothing could ever be flagged. Corroboration is evidence only when the claim and the corroboration come from independently-authored places.

So the env layer resolves against declared Services plus an explicit `HELM_PROVIDED` set — 10 Services created by Helm charts that gitops references as ArgoCD Applications (kube-prometheus-stack, loki, tempo, pyroscope, temporal, openbao, holmesgpt), which no `kind: Service` in this repo names. Each was verified with `kubectl -n <ns> get svc <name>` and its ports recorded in the comment so a reader can re-run the check rather than trust the line.

That list is **checked both ways** — an entry no workload dials any more fails the gate — so it cannot rot into the permanent unchecked scope the header rightly warns about.

## Also fixed: an explicit `.svc` in an undeclared namespace was skipped

The header claimed form (a) was "unambiguous, always checked". The code disagreed: `if ns not in namespaces: continue` treated it as external. But `.svc` is a cluster suffix by definition, so an unknown namespace means it cannot resolve — which is exactly the namespace-typo case, and how three of the four defects hid. Now flagged in **both** layers.

The short `<name>.<ns>` form keeps the old skip: there an unknown second label really is ambiguous with a real domain (`api.github.com` is not `api` in namespace `github`). That asymmetry is why `normalise` now reports whether the suffix was explicit.

## The four defects

All verified against the live sandbox cluster on 2026-08-07.

| who | declared | correct | wrong how |
|---|---|---|---|
| psd2, security-scanner | `tpp-registry-service.tpp.svc:8108` | `…tpp-registry.svc:8108` | namespace `tpp` never existed |
| openbao secret-rotator | `keycloak.keycloak.svc:8080` | `keycloak.iam.svc:8080` | no namespace `keycloak` |
| standing-order, clearing-simulator | `sepa-payment-service.payments.svc:8108` | `sepa-payment.payments.svc:8115` | wrong Service **and** wrong port (8108 is tpp-registry's) |
| vop-service | `party-service.parties.svc:8100` | — | fixed in #3968 |

**Corroborating evidence they never worked:** NetworkPolicies are derived from these env URLs, and regenerating gives tpp-registry's ingress allow-list `psd2` and `security-scanner` for the first time. The edges did not exist, so a corrected hostname alone would still have been dropped by the CNI. That comes from a generator, not from reading the values back.

### One comment corrected rather than deleted

psd2's manifest said the TPP registry lookup was "Not yet deployed → calls 503 until it lands". That was false — `tpp-registry-service` serves 8108 in `tpp-registry`. The note is rewritten in place because **a comment that explains away a symptom is worse than none**: it retires the question. The failing lookup looked accounted-for for as long as anyone read the file.

Small second instance of the same theme in the ASVS baseline: `keycloak.keycloak.svc:8080` carried a V9.1 debt entry only because it did not match the exemption for the in-cluster Keycloak leg — that exemption names the *real* host. Correcting the hostname makes the finding disappear, so the baseline line is deleted, not rewritten. A wrong hostname had been carrying its own ASVS debt for as long as it was wrong.

## Falsification

Not only the fixture harness — a **real counter-example in the tree**: putting `tpp-registry-service.tpp.svc` back into psd2's manifest takes the gate from exit 0 to exit 1 naming that exact env var, and reverting returns it to 0.

`--self-test` is **31/31**, up from 19. The added cases cover the undeclared-namespace rule, the short-form asymmetry that must stay *quiet*, the env layer, `HELM_PROVIDED`, stale detection, and the circularity assertion — a corroborated-only host stays clean in the application.yaml layer and **is** flagged in the env layer. That last pair is the one that would fail if someone ever "simplified" the two layers to share a believed-set.

## Coverage

Before: 56 application.yaml files, 82 hostnames.
After: 56 application.yaml files (77 hostnames) **+ 477 gitops workload env/arg values (468 hostnames)**, against 291 declared Services (+131 corroborated, application.yaml layer only; +10 Helm-provided, env layer) across 59 namespaces.

## Local verification

`run-gates.py --all`: 112 gates, **109 PASS**, 2 WARN, 1 FAIL.

- `api-contract-gate` (the FAIL) cannot run locally (`sudo: a password is required` installing `oasdiff`); no `openapi.yaml` is in this diff.
- `release-scope-mismatch` and `operator-write-naming` are advisory and pre-existing on `main`.
- `threat-model-updated-on-trust-boundary-change` passes — the money-path threat models carried in from #3956/#3968 satisfy it.

The party NetworkPolicy conflict I warned about in #3968 **did** fire on merge here rather than silently dropping a line. Resolved by regeneration, not by hand, and verified: `grep -cE 'metadata.name: (onboarding|payments)$'` prints 2.


JiRaska:
## Review: no blocking findings — falsified against the real historical defect

I did not test this against a synthetic sample. I reintroduced **the exact defect #3968 fixed yesterday** into this branch — VoP's party hop, `party-service.parties.svc:8100`, where the namespace `parties` has never existed — and ran the gate:

```
::warning file=…/payments/payments-services.yaml::in-cluster host 'party-service.parties.svc'
at `vop-service.vop-service.env.PARTY_SERVICE_URL` does not resolve: namespace 'parties' is not
declared anywhere in gitops … Nothing in this repo's test layers can catch this — a unit test
stubs the client and an IT serves localhost — so fix the hostname, or declare the Service.

1 unresolvable in-cluster hostname(s) (0 in application.yaml, 1 in a gitops workload env).
```

Exit 1 under `--enforce` (measured directly, not through a pipe). On the unmodified branch it passes: 56 application.yaml files + 477 gitops env values, 291 declared Services, clean.

So this gate would have caught, at PR time, a defect that shipped and stayed live long enough to be documented as "VoP confirmed a WRONG payee". That is the strongest thing that can be said for a gate here, and the error message itself explains why no test layer could have.

## The circularity argument is the good part

> The corroboration tier — a host is believed if some gitops workload env dials it — is derived from those same env values. Reusing it here would be circular: every env host would corroborate itself and nothing could ever be flagged.

That is exactly right and easy to get wrong. Extending a check to a second file layer while carrying the same corroboration set over would have produced a gate that runs, reports a scope, and cannot fail — the failure mode this repo keeps rediscovering. The output line makes the asymmetry visible too (`+131 corroborated, application.yaml layer only; +10 Helm-provided, env layer`), so a reader can see which tier applied where rather than having to trust it.

## One non-blocking note: annotation severity does not match the outcome

Under `--enforce` the script exits 1 but emits only `::warning::` annotations — zero `::error::` lines. The job fails correctly, but in the Actions UI the finding renders as a warning, which is the one place a reader looks to tell "this blocked the merge" from "this is advisory noise". Every other enforced gate I have looked at in this repo emits `::error::` when it is going to fail.

One-line fix: pick the prefix from the enforce flag, the way `check-agent-case-schema.py` does:

```python
print(f"{'::error::' if enforce else '::warning::'}{finding}")
```

Not blocking — the exit code is what gates the merge, and that is right.

Automated agent review, not an approval. Touches `.github/gates/`, `.github/scripts/` and money-path gitops, so it needs human review regardless.

JiRaska:
Ran the self-test and both trees, not just read the description.

```
this branch --self-test    all cases pass, including the two that generalise #3981's lesson:
                            "corroborated-only host is clean in application.yaml layer" /
                            "the SAME host IS flagged in the env layer (no self-corroboration)"
this branch, full scan     OK — 56 application.yaml + 477 gitops env values, 0 findings

main's checker, full scan  OK — but it never scans the gitops env layer at all, so "OK" here
                            is not evidence of correctness, it's evidence the old scope never
                            looked. Matches the PR's own claim about #3967's gap.
```

The self-test cases matter more than usual here because of exactly the failure this repo hit with the sibling gate (#3981): a naive "believe any host ≥2 workloads dial" rule would have been wrong in both directions on the real tree. This gate instead keeps the two layers from corroborating each other (a host clean only via the application.yaml layer must still be flagged when the SAME host shows up in the env layer), which is the property that makes "OK" here actually mean something.

Not merging — money-path/high-risk paths (account, card-issuance, party network-policies), human review by rule.

JiRaska:
Reviewed. **No blocking findings — the gate is falsifiable, correctly registered, and the load-bearing design decision is right.**

**Falsified, not assumed.** On a worktree of this branch merged with current `main`:

- `--self-test` exits 0, including the pair that matters — *a corroborated-only host is clean in the application.yaml layer* / *the SAME host IS flagged in the env layer (no self-corroboration)*.
- Clean tree with `--enforce`: exit 0, *59 application.yaml files (77 hostnames) + 482 gitops workload env/arg values (473 hostnames), against 295 declared Services (+132 corroborated, application.yaml layer only; +10 Helm-provided) across 60 namespaces*.
- **Known-positive:** editing `psd2-service.yaml:108` from `tpp-registry-service.tpp-registry.svc` to `.tpp.svc` produced exit 1 with `::error … in-cluster host 'tpp-registry-service.tpp.svc' at psd2-service.psd2-service.env.TPP_REGISTRY_SERVICE_URL does not resolve: namespace 'tpp' is not declared anywhere in gitops`. Restoring the file returned exit 0.

That circularity argument — env values cannot corroborate themselves — is the design decision the whole gate rests on, and it is correct.

**Registration and enforcement both check out**, in both places they have to: `gates.yaml:444-452` declares `id: incluster-hostname-resolution`, `group: gitops`, `mode: enforced`, with `selftest` + `selftest_expect: pass`; `run-gates.py --list` confirms it; and the `run:` passes `--enforce`, without which the script warns and exits 0. Not an inline step in a conditional job.

**Note for the thread:** the earlier review comment asking for `::error` rather than `::warning` under `--enforce` is **stale** — the current head already emits `::error`, as the known-positive above shows.

**Two blind spots, measured rather than guessed.** Neither is a defect in the PR; the second is worth a line in the script header.

1. **A wrong port is invisible.** Changing only the port (`…tpp-registry.svc:8108` → `:9999`) left the gate at exit 0 / *every in-cluster hostname resolves*. Worth naming because one of the four defects this PR fixes was a wrong port (`sepa-payment-service.payments.svc:8108` → `sepa-payment.payments.svc:8115`) — the gate catches the hostname half of that defect and would not catch a recurrence of the port half. Nothing compares the recorded ports to `Service.spec.ports`.

2. **`valueFrom` leaves the checked set silently.** The harvest reads `containers[].env[].value` only when it is a string, plus `args[]`. Converting psd2's literal env to a `configMapKeyRef` kept the gate at exit 0, and the only trace was the hostname count dropping 473 → 472. So the fleet-standard remedy for a flagged host — move it into a ConfigMap — removes it from coverage with no complaint. That is the same shape as the scope gap this gate exists to close, one layer out. `envFrom` and concatenated hostnames (`$(SVC_NAME).$(NS).svc`) are unread for the same reason.

JiRaska:
**Unblocked from `CONFLICTING` (`109bfbe3d`, signed). One conflict, and it was append-only prose.**

`docs/threat-models/openbank-account-service.md` — `main`'s 2026-08-05 `operator-account-write` entry against this branch's 2026-08-07 hostname entry. Kept **both**, chronologically. "Ours vs theirs" are not alternatives for a dated log; picking one deletes a threat-model record with nothing to notice.

Verified by counting entries rather than reading the diff, because a clean merge in this repo has silently dropped a neighbouring list entry before:

| | branch | main | merged |
|---|---|---|---|
| account-service entries | 18 | 20 | **21** |
| `.release-please-manifest.json` | — | 58 | **58** |
| `gates.yaml` ids | — | — | 119, no duplicates |

`incluster-hostname-resolution` survives the merge as `mode: enforced`, and the gate PASSes on the merged tree through the real runner. The NetworkPolicy regeneration you flagged as the expected conflict point is **idempotent** against the merged result — `gen-network-policies.py` re-run produced no change beyond what the merge already carried, so nothing was hand-resolved in a generated file.

Nothing rebased or force-pushed. The worktree at `.claude/worktrees/heuristic-grothendieck-5a3ccf` had been dormant since 2026-08-07T09:41 with a clean tree and its head matching the remote exactly.

---

**Independently corroborated, and it found one more than you did.** Working #3978 from the issue side, I built the same fix from scratch before finding this PR, and arrived at **byte-identical** changes to settlement, onboarding and security-scanner `application.yaml`, the same psd2 `tpp` → `tpp-registry` namespace fix, and the same regenerated NetworkPolicies. Two independent derivations converging is decent evidence both are right.

Where this PR is strictly better than what I built:

- It checks **all** gitops env/arg values, not only those substituted into an `application.yaml` `${VAR:default}`. Mine left a hole — `security-scanner-service.yaml`'s env still held `tpp-registry-service.tpp.svc:8108` and my gate went green, because that value backs no `${VAR}`.
- It finds two live defects mine could not: the openbao secret-rotator's `keycloak.keycloak.svc`, and standing-order/clearing-simulator's `sepa-payment-service.payments.svc:8108` (wrong Service *and* wrong port).
- Its per-layer believed-set plus a `kubectl`-verified `HELM_PROVIDED` list, checked in both directions, is the shape root `CLAUDE.md` prescribes. Mine used per-module self-exclusion, which is weaker.

**A bug in the pre-existing gate that both derivations hit independently:** `normalise()` returned `(name, ns)` and `scan` did `if ns not in namespaces: continue`, so an explicit `.svc` host in a namespace that *does not exist* was skipped **as external** — exactly the `parties`/`tpp` typo shape this gate exists to catch. My first real-tree injection of `party-service.parties.svc` was not flagged, and my own self-test had masked it because I had put `parties` in the fixture namespace set. Caught only by the real-tree counter-example, not by the self-test.

**Corroboration that psd2's TPP hop never worked at all:** NetworkPolicies are derived from those env URLs, so regenerating adds `psd2` to tpp-registry's ingress allow-list **for the first time**. The edge did not exist — a corrected hostname alone would still have been dropped by the CNI.

**On #3978's stale numbers**, re-derived: **5 findings / 4 services**, not the issue's 6 / 3, and stale in both directions. Dropped: vop (fixed by #3968), account ×2 and card-issuance — their gitops overrides are correct, so the old gate was counting dead defaults as defects. Added: psd2, which the issue never names.

This still needs its **2 human approvals** — it touches settlement-service, which is money-path. I have not reviewed, approved or merged it.

PR #3980, opened 2026-08-07
ci(agent-review): defer to human review when GitHub Models is retired

JiRaska:
## Summary

GitHub Models was retired on 2026-07-30 and now returns HTTP 410 on every inference call. The Claude fallback is also unavailable until its OAuth token is re-minted (#2161). This change detects the retired primary reviewer, skips the broken fallback, posts an explicit deferral comment on the PR, and exits green so `agent-review` stops failing on every routine PR.

## Type of change

- [ ] fix (bug fix)
- [ ] perf (performance)
- [ ] docs
- [x] chore (build / tooling)
- [ ] breaking change

## Linked issues

Refs #3935, #3488, #2161

## Changes

- `.github/workflows/agent-review.yml`:
  - Add a step that detects when the GitHub Models primary reviewer failed and marks it as retired.
  - Skip the known-dead Claude fallback when the primary is retired.
  - Post a PR comment explaining the deferral and exit green.
  - Skip the terminal "did a review happen?" guard when the deferral path was taken.

## Definition of Done

- [ ] Hexagonal layering preserved (domain has zero framework imports).
- [ ] Unit tests added/updated.
- [ ] No new lint warnings.

_N/A — CI workflow change only; no application code, tests, OpenAPI, DB, or events touched._

## Security checklist

- [ ] No new `@SuppressWarnings`, `as Any`, `@Suppress("...")` without justification (state it in this PR).
- [ ] No new third-party dependency, OR: dependency review attached (license, CVE, source).

## Compliance impact

- [ ] PCI DSS
- [ ] DORA
- [ ] GDPR
- [ ] PSD2 / SCA / RTS
- [ ] AML / 5AMLD
- [ ] CNB reporting
- [x] None

## Rollout / Rollback

- Deployment strategy: N/A — workflow change takes effect on the next PR event.
- Rollback plan: revert the commit / close the PR; the previous workflow behaviour was a red `agent-review` check on every PR.

### DIFF
--- a/.github/workflows/agent-review.yml
+++ b/.github/workflows/agent-review.yml
@@ -353,17 +353,28 @@ jobs:
       - name: GitHub Models review
         id: ghmodels
         if: steps.scope.outputs.skip == 'false'
-        continue-on-error: true   # any failure → Claude fallback
+        continue-on-error: true   # any failure → Claude fallback, unless retired
         uses: actions/ai-inference@a7805884c80886efc241e94a5351df715968a0ad # v2.1.1
         with:
           model: openai/gpt-4o
           prompt-file: /tmp/prompt.txt
           max-tokens: 400
 
+      # GitHub Models was retired on 2026-07-30 and now answers every inference
+      # call with HTTP 410 (#3488). That failure is permanent, not flaky, so
+      # skip the known-dead Claude fallback (#2161) and defer to a human instead.
+      - name: Check for retired primary reviewer
+        id: ghmodels_retired
+        if: steps.scope.outputs.skip == 'false' && steps.ghmodels.outcome == 'failure'
+        run: |
+          set -euo pipefail
+          echo "retired=true" >> "$GITHUB_OUTPUT"
+          echo "::notice title=agent-review::GitHub Models is retired (HTTP 410); deferring to human review."
+
       # --- Submit the GitHub Models verdict ----------------------------------
       - name: Submit review (GitHub Models)
         id: submit
-        if: steps.scope.outputs.skip == 'false' && steps.ghmodels.outcome == 'success'
+        if: steps.scope.outputs.skip == 'false' && steps.ghmodels.outcome == 'success' && steps.ghmodels_retired.outputs.retired != 'true'
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           PR: ${{ github.event.pull_request.number }}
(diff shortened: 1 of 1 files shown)


JiRaska:
## Review: 1 blocking finding

First, to clear a wrong guess of my own: this is **not** a duplicate of the `Defer to human review` step already on `main`. That one fires on `scope.outputs.skip == 'true'` — the sensitive-scope path — and is unrelated. The two are complementary.

### Blocking: `retired=true` is asserted from *any* failure, not from a 410

```yaml
- name: Check for retired primary reviewer
  id: ghmodels_retired
  if: steps.scope.outputs.skip == 'false' && steps.ghmodels.outcome == 'failure'
  run: |
    echo "retired=true" >> "$GITHUB_OUTPUT"
```

The condition is `outcome == 'failure'`, full stop. Nothing inspects the HTTP status. The comment above it states *"That failure is permanent, not flaky"* — but the code does not establish that; it assumes it. A rate limit, a transient 5xx, a network blip, a malformed prompt, or an action-version problem all set `retired=true` identically.

Three things then happen on that unverified premise:

1. **The Claude fallback is skipped** (`steps.ghmodels_retired.outputs.retired != 'true'`) — including in the future when its OAuth token *is* re-minted and it would have worked. The PR's own rationale for skipping it is that it is currently dead; the gate it adds outlives that reason.
2. **The accountability guard is skipped.** This is the serious one:

   ```yaml
   - name: Verify a review actually happened
     if: always() && steps.scope.outcome == 'success' && steps.ghmodels_retired.outputs.retired != 'true'
   ```

   That guard is #3488's whole point — its own comment in this file says *"A run that reviewed nothing must not be readable as a review."* After this change, any transient failure of the primary reviewer disables it and the job goes green. The guard against vacuous green becomes bypassable by exactly the class of event it was written to survive.
3. **A comment is posted asserting GitHub Models was retired**, which may simply be untrue on that run.

### Two ways to fix, both cleaner than branching on `outcome`

- **Verify the 410.** Capture the actual status from the inference step and set `retired=true` only for HTTP 410 (or a documented, matched error string). Anything else keeps the existing fallback + guard path intact. That preserves the accountability property and makes the notice truthful.
- **Or accept the retirement as permanent and delete the path.** If GitHub Models is gone for good, the honest shape is to remove the `GitHub Models review` step and its submit step outright, leave the Claude fallback (or nothing) as the reviewer, and keep `Verify a review actually happened` unconditional. That removes a branch instead of adding one, and it cannot silently reactivate.

The first is smaller; the second matches what the description says is actually true.

### Not blocking

`agent-review` is **not** a required check on this repo — the required contexts are `all-green`, `Validate manifests`, `Gitleaks`, `issue-hygiene`, `OPA policy gate`. So its red never blocks a merge; it makes `gh pr merge` report `UNSTABLE` and adds noise. Worth knowing when weighing the urgency: this is a noise fix, not an unblocking fix, which argues for taking the cleaner of the two options rather than the faster one.

Automated agent review, not an approval; this touches `.github/workflows` and needs human review regardless.

Issue #3318, opened 2026-08-02
can-i-deploy: refuse a workflow_dispatch deploy of a sha with no pact version instead of silently asking about a different commit


JiRaska:
## The problem

`resolve-can-i-deploy-selector.sh` (#3082) asks the precise question — `--version <sha>` — whenever the broker has a pact version for the commit being deployed. When it does not, it falls back to `--latest main`.

That fallback is correct for a **push**: path-scoped CI skips most services on any given commit, so most of the fleet legitimately has no version for that sha, and `--version` would error.

It is **not** correct for a `workflow_dispatch`. A manual `auto-deploy` builds `TAG="sandbox-${GITHUB_SHA::8}"` from the current `main` tip — a commit that `services-ci` never built for that service, so a pact version for it cannot exist. The gate then answers about a *different* commit than the one being pinned into gitops.

The script's own header already states both outcomes, and names the quiet one as the dangerous one:

> * the older version was NOT verified → red, labelled PENDING_BUILD. Noisy, honest, self-clearing.
> * the older version WAS verified → **GREEN, and auto-deploy ships an image built from `$GITHUB_SHA` whose contracts were never verified at all.** Nothing anywhere says so. A contract gate that answers about the wrong version is not a gate.

## Measured

`openbank-fx-service`, 2026-08-02: five manual `auto-deploy` dispatches, all five blocked with an identical verdict about consumer version `8d321a8` — a commit unrelated to the change being deployed. That is the loud half. The silent half has the same cause and ships the image.

Full context, including why re-dispatching cannot converge: #3306.

## Proposed change

On the `no` branch, when `GITHUB_EVENT_NAME == workflow_dispatch`: **fail with an explicit message** instead of falling back to `--latest main`.

Something like: *"the sha being deployed (`<sha>`) has no published pact version for `<svc>`, so the gate cannot ask about it. A manual deploy of an unbuilt sha would be gated on a different commit. Run `services-ci` for this sha, or dispatch a deploy of a sha that has one."*

Push behaviour is unchanged — the fallback stays exactly as it is today.

## Why this one first

It is the smallest change that removes the dangerous-green half entirely, and it converts a silent wrong answer into a loud one. It does make manual deploys of an unbuilt sha fail — deliberately: today they either block confusingly or ship unverified, and neither is better than a clear refusal.

[...]

Issue #2020, opened 2026-07-24
Auto-deploy: a service stranded by a transient can-i-deploy block is never re-driven


JiRaska:
## Symptom

#1990 (feat(notifications): per-party push preferences, commit 6de0d5e9) changed both `openbank-customer-edge` and `openbank-notification-service`. The auto-deploy run built + pushed **both** images, but only `openbank-customer-edge` reached gitops — notification-service kept running the pre-#1990 image (V11 `notification_preferences` absent, `/api/v1/preferences` 404). Fixed by hand in PR #2017.

## Root cause — NOT changed-service detection

The `changes` job emitted `["openbank-notification-service","openbank-customer-edge"]` correctly. The block was the **can-i-deploy contract gate**:

```
There is no verified pact between the latest version of openbank-notification-service with tag
main (c49ceac...) and the version of openbank-account-service currently in sandbox (2799ad80...)
##[error]can-i-deploy: openbank-notification-service NOT deployable
```

notification-service has a real async message pact consuming an event from account-service. #1990's build republished notification's main pact at 20:00; can-i-deploy ran at 20:34 against the account-service **then in sandbox** (`2799ad80`), which had not yet verified that pact. account-service's provider verification landed later. Querying the broker now: `c49ceac ↔ 5bbebf2e verified=true, deployable=true` — a **timing race**, not a contract break.

The per-push run is **one-shot**: blocked → stranded until a human hand-bumps gitops. This is the same class as #1420/#1009 and is a **standing backlog** — ~23 deployable services are currently behind main for the same reason.

## Fix

Scheduled reconcile in `auto-deploy.yml` that re-offers every service whose gitops image tag is behind main to the same build → can-i-deploy → gitops-pr path, so a strand self-heals once the gate clears. The gate is **not** weakened; a still-blocked service stays blocked (and keeps its #1420 escalation).

PR #3992, opened 2026-08-07
ci(auto-deploy): guard manual dispatch and reconcile

JiRaska:
## Summary

Adds a `reconcile` workflow_dispatch input and a guard that rejects manual dispatches of services that did not change in the current HEAD. This prevents the wasted NOT_ASKED / missing-pact failures seen in runs 31149468822 and 31150052893, where services were dispatched from unrelated commits.

## Type of change

- [ ] fix (bug fix)
- [ ] perf (performance)
- [ ] docs
- [x] chore (build / tooling)
- [ ] breaking change

## Linked issues

Refs #3318, #3454, #2020

## Changes

- Added `reconcile` boolean input to `workflow_dispatch` for the `auto-deploy.yml` workflow.
- `reconcile=true` runs `.github/scripts/auto-deploy-reconcile-lag.sh` immediately, re-driving stranded services to their correct per-service commits instead of forcing them onto the current HEAD.
- `workflow_dispatch` with explicit `services` now rejects any service that did not change in HEAD and is not a dependent of a changed shared library.
- Empty `services` is no longer allowed unless `reconcile=true`.
- `can-i-deploy` co-deploy branch is skipped when `reconcile=true`, because the selected services may span multiple SHAs.
- Moved the `CHANGED_FILES` / `LIBS_DEPENDENTS` / `DIRECT_SERVICES` computation earlier so both the push and manual-dispatch paths share it.

## Definition of Done

- [ ] Unit tests added/updated.
- [x] No new lint warnings.

## Security checklist

## Compliance impact

- [ ] PCI DSS
- [ ] DORA
- [ ] GDPR
- [ ] PSD2 / SCA / RTS
- [ ] AML / 5AMLD
- [ ] CNB reporting
- [x] None

## Rollout / Rollback

- Deployment strategy: applies to the next CI run; workflow file change is effective immediately after merge.
- Rollback plan: revert this PR and, if needed, dispatch the affected services from the correct commit.

## Reviewer notes

- `actionlint` reports the same pre-existing custom runner label warning for `openbank-build` that exists on `main`; no new lint warnings were introduced by this change.
- This changes the behavior of manual dispatches: operators can no longer list arbitrary services at an unrelated commit. They should use `reconcile=true` to recover stranded services, or dispatch from the commit that actually changed the service.


JiRaska:
Hi @JiRaska, this change adds a `reconcile` input and a guard for manual `workflow_dispatch` in `auto-deploy.yml`.

The guard rejects dispatches that list services not changed in the current HEAD (and not dependents of a changed shared library). This prevents the wasted runs we saw in 31149468822 and 31150052893, where services were dispatched from an admin-ui / docs commit, leaving the broker with no pact version for that SHA and producing `NOT_ASKED` / missing-pact can-i-deploy verdicts.

If ops need to recover stranded services at a non-service commit, `reconcile=true` runs the scheduled reconcile script on demand, which re-drives each service to its correct per-service commit instead of forcing everything onto HEAD.

Refs: #3318, #3454, #2020.

Please review when you have a moment.

JiRaska:
## Review: no blocking findings — the cited failures reproduce exactly as described

Both runs named in the description are real and say what the PR says they say:

```
run 31149468822  failure  event=workflow_dispatch  head=78988c633
run 31150052893  failure  event=workflow_dispatch  head=403def459
```

and the cause, from 31149468822's own log:

```
[NOT_ASKED] workflow_dispatch of 78988c633… which has no published pact version for
openbank-pid-service — the gate cannot ask about the sha being deployed, and asking about a
different commit is not a gate (#3318)

This is NOT self-clearing: services-ci is path-scoped, so a sha that did not touch
openbank-pid-service never gains a version, and re-running the same dispatch cannot change that.
```

That last sentence is the whole justification for the guard. The failure mode is not a flake to retry — a manual dispatch of a service the current HEAD did not touch can *never* clear, because the pact version it needs will never exist for that sha. Refusing at dispatch time is the only thing that helps; the alternative is a run that builds, pushes and then throws the images away.

## The design holds up on the point I checked hardest

A guard that only rejected would be worse than the disease: legitimately redeploying an unchanged service is a real need (a stranded gitops tag, exactly issue #2020). This PR keeps that path open rather than closing it —

```
if [ "${{ github.event.inputs.reconcile }}" = "true" ]; then
    … auto-deploy-reconcile-lag.sh …
```

`reconcile=true` re-drives each stranded service to *its own* correct commit instead of building everything at HEAD, and the refusal fires only on the blank-services-without-reconcile combination. So the escape hatch is the semantically right one, not a bypass flag.

`libs-change-dependents.sh`, which the changed-service computation calls, is present on `main`.

## One thing worth a second look before merging

`CHANGED_FILES` comes from `git diff --name-only HEAD~1 HEAD`. On a squash-merged `main` that is the right window, but it is one commit wide — a dispatch aimed at a service changed two commits back is refused even though it genuinely changed recently. That may well be intended (the pact version is per build-sha, so only HEAD's build is answerable), and if so it is worth a sentence in the error message, because the operator's next thought will be "but I *did* change it". Not blocking; the refusal is still the safe direction.

Automated agent review, not an approval. This touches `.github/workflows`, so it needs human review regardless — merging it changes what executes on the self-hosted runners.

JiRaska:
The red `Validate manifests` here is one gate, and the finding is bigger than this PR.

```
::error file=.github/workflows/auto-deploy.yml::job 'can-i-deploy', step 'can-i-deploy
(per changed service)' has a 19149-character `run:` script (limit 19000)
```

Measured on both sides, parsing the YAML rather than counting the diff:

| | that step's `run:` |
|---|---|
| `origin/main` | **18955** chars |
| this branch | 19149 chars |

So `main` sits **45 characters** under the ceiling. This PR adds 194 and crosses it, but any PR that adds more than 45 characters to that step does the same — including a one-line comment. The gate is doing its job; the step is the problem.

Why that matters more than a normal lint: the ceiling is not the gate's invention. Above roughly 20054 characters GitHub stops parsing the **whole workflow file** — no jobs, no error message, `name:` never read, and it renders as an ordinary red run while every contributor's deploys are dead until someone reverts it (that is what #3135/#3139 cost). The 19000 limit is the guard band, and `main` has 0.2% of it left on the fleet's deploy path.

Two ways forward, and the choice is yours:

1. **Shave ~200 characters out of what this PR adds.** Unblocks this PR today. Leaves the next one to hit the same wall, with less margin.
2. **Extract the step into `.github/scripts/can-i-deploy.sh` and call it.** What the gate's own message prescribes ("Move the logic into `.github/scripts/` and call it; prose belongs in the script's header"), and it retires the cliff for everyone rather than moving it. Bigger diff, and it touches the deploy path, so it wants care and a human.

I have not pushed anything to this branch. It merges cleanly with `main` (0 conflicts), but refreshing it would not change the verdict — the failure is real on the current head, not stale. Say the word and I will do (2) as its own PR so this one can rebase onto a step that has room again.

PR #4003, opened 2026-08-07
chore(governance): regenerate the EU AI Act inventory and AI-governance snapshot

JiRaska:
## main is red

#3771 added the ADR-0244 `case-coordinator` charter to `agents.yaml` without regenerating the two
artifacts derived from it. Both enforced drift gates now fail **on main itself**:

```
gate failed: EU AI Act inventory drift (ADR-0148 rule #7, enforced)
gate failed: AI governance snapshot drift (ADR-0029 D3 / ADR-0031, enforced)
```

`Validate manifests` is a required status check and aggregates those shards, so it fails with them.
Confirmed red on both `3208109c` (#3771's own push) and `d16dc102` (the release commit after it).

Every PR that merges main in inherits this. The three open PRs that are still green are green only
because they predate #3771 — the first rebase turns them red.

## What is here

Regenerated by `gen-eu-ai-act.py` and `gen-ai-governance-snapshot.py`, not hand-edited (rule 6).
The whole diff is the one new agent:

- inventory: one row for `case-coordinator`
- snapshot: the two content hashes, control agents `10 -> 11`, total `15 -> 16`, pending-eval `1 -> 2`

## Not included, on purpose

`check-operator-write-naming` also prints an error on main, for `rest.rego`'s `operator-mcp-session`
(GHSA-58jq-9hq3-66jr). That gate is `mode: advisory`, so it is **not** why main is red — I initially
read its `::error::` output as a blocking failure and that was wrong; the gate manifest is what
settles it, not the log line. The fix is a security decision between renaming the rule, pinning
`input.principal.id`, or a declared exception with a reason, and belongs in its own PR.

Refs #3771

### DIFF
--- a/docs/compliance/eu-ai-act.md
+++ b/docs/compliance/eu-ai-act.md
@@ -41,6 +41,7 @@ it can move past *Proposed*.
 | `docs-truth-agent` | control | Limited / minimal risk | — | produces proposals only; a human dispositions every effect (not autonomous decision-making on a natural person). |
 | `authz-policy-auditor` | control | Limited / minimal risk | — | produces proposals only; a human dispositions every effect (not autonomous decision-making on a natural person). |
 | `flaky-test-hunter` | development | Limited / minimal risk | — | produces proposals only; a human dispositions every effect (not autonomous decision-making on a natural person). |
+| `case-coordinator` | control | Limited / minimal risk | — | produces proposals only; a human dispositions every effect (not autonomous decision-making on a natural person). |
 
 ### Non-agent ML and statistical systems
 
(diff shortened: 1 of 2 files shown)


JiRaska:
Overlap note: this PR fixes `eu-ai-act-inventory-drift` + `ai-governance-snapshot-drift` but not the third red gate from #3771 — the `OPA policy gate` / "verify bundle, check ConfigMap sync" failure needs `openbank-infra/gitops/components/agent/agent-opa-bundle.yaml` + `agent-service.yaml` regenerated too (that is what blocks auto-deploy gitops PR #4000). #4004 carries all four files (shared two are byte-identical generator output) and has auto-merge armed — if it lands first, this one rebases down to nothing.

PR #4002, opened 2026-08-07
fix(governance): regenerate OPA bundles and EU AI Act inventory after agents.yaml change (#3771)

JiRaska:
## Summary

#3771 (ADR-0244 case-coordinator charter + case-budget schema) changed `openbank-libs/governance/agents.yaml` without regenerating its derived artifacts. `agents.yaml` is embedded verbatim into 40 OPA bundle ConfigMaps (`rules.yaml: opa_bundle_sync`) and `docs/compliance/eu-ai-act.md` is generated from it (ADR-0148 rule #7) — both enforced gates (`OPA policy gate`, `eu-ai-act-inventory-drift` in `Validate manifests`) fail against current `main`. This PR regenerates everything from the committed source; no hand edits.

## Type of change

- [x] fix

## Linked issues

Refs #3998 (Phase 0 of #3737 — the charter/schema half was delivered by #3771; this completes its derived-artifact tail) · Refs #3771

## Changes

- `docs/compliance/eu-ai-act.md` — regenerated via `python3 .github/scripts/gen-eu-ai-act.py` (16 charters incl. case-coordinator)
- 40 `*-opa-bundle.yaml` + 31 workload `policy-checksum` annotations — regenerated via every `gen-*opa-bundle*.sh` under `openbank-infra/gitops/components`
- `openbank-admin-ui/ai-governance-snapshot.json` — regenerated via `python3 .github/scripts/gen-ai-governance-snapshot.py` (third derived artifact #3771 left stale: prompts/registry.yaml changed with the case-coordinator prompt entry)

## Verification

- `check-eu-ai-act.sh` → OK, in sync
- `check-agent-charter-registry.sh` → OK, 16 agents full parity
- `check-opa-bundle-parses.py` → OK, 40 bundles
- `check-opa-sidecar-bundle-shape.py` → clean
- `check-opa-bundle-apply-size.py` → OK
- Scope: `git diff origin/main...HEAD --name-only` = only the three derived-artifact classes above
- `gen-ai-governance-snapshot.py --check` → in sync

Note for reviewers: `check-eu-ai-act.sh` restores the derived doc to HEAD on a failing run (line 25), the same trap class as `check-adr-registry.sh` (CLAUDE.md `### ADR registry`) — regenerate, commit, THEN check. A CLAUDE.md bullet covering this sibling script follows separately.

## Definition of Done

- [x] Derived data regenerated from source, not hand-edited
- [x] No secrets, tokens, passwords, or PII

Issue #3998, opened 2026-08-07
Phase 0: swarm case-coordination schema + case-coordinator charter (ADR-0244)


JiRaska:
## Phase 0 — swarm case coordination: charter + schema (no runtime code)

First slice of the swarm delivery plan in #3737, enacting the ADR-0244 follow-ups: *"`agents.yaml` extension (case budgets, swarm-participation and `case-open` capabilities, per-case-class ceilings and contested-rate thresholds); `case-coordinator` charter."*

## Scope

1. **`openbank-libs/governance/agents.yaml` schema extension**
   - New `case_coordination` block: per-case-class defaults (token budget, wall-clock deadline, fleet-wide concurrent-open ceiling, contested-rate circuit-breaker threshold) — ADR-0244 D6/D9
   - Pilot case class `incident-triage` only (non-money-path; agreed defaults: 10 concurrent, 1 open/hour/agent, 24 h TTL)
   - New capabilities: `case-open` (D9), swarm participation (`case-contribute`), coordination (`case-coordinate`, `case-invite`, `case-synthesize`)
   - `case-open` granted to **no** existing agent in this phase (runtime wiring is Phase 1); the schema + coordinator charter exercise the new fields
2. **`case-coordinator` charter** (ADR-0244 D3/D8)
   - New `agents.yaml` entry: control plane, coordination capabilities only, own limits + kill-switch scope; the agent that opens a case never coordinates it
   - `docs/agents/case-coordinator.md` narrative (ADR-0156 parity gate; frontmatter id/plane/adr)
3. **Derived artifacts regenerated**
   - `docs/compliance/eu-ai-act.md` (`gen-eu-ai-act.py` — derived from agents.yaml)
   - All OPA bundle ConfigMaps (agents.yaml is embedded verbatim, `rules.yaml: opa_bundle_sync`)
4. **Schema-validation gate** — new `check-agent-case-schema.py` in `gates.yaml` (enforced, with `--self-test`): case-class invariants (budgets positive, contested threshold in (0,1), `case-open` only on agents whose charter names a case class, coordinator never holds `case-open`)

## Explicitly out of scope

- Any runtime/Kotlin code, Temporal workflow, REST endpoint (Phase 1)
- Granting `case-open` to detector agents (lands with Phase 1 runtime wiring)
- admin-ui thread view (Phase 3)

## Acceptance

- `check-agent-charter-registry.sh` OK (parity incl. new narrative doc)
[...]

PR #4006, opened 2026-08-07
docs(governance): generalize the derived-file restore trap and enumerate agents.yaml derived tails

JiRaska:
## Summary

Follow-up promised in #4002: generalizes the ADR-registry restore-trap bullet to all derived-file checks (`check-eu-ai-act.sh` restores its doc to HEAD on failure too — it reverted my first regen before staging), and adds the missing enumeration of `agents.yaml`'s THREE derived tails (eu-ai-act.md, 40 OPA bundles + checksum annotations, ai-governance-snapshot.json) so the next agents.yaml PR regenerates everything in one go instead of red-gating main the way #3771 did.

## Type of change

- [x] docs

## Linked issues

Refs #3998 · Refs #3737 (Phase 0 knowledge capture)

## Changes

- `CLAUDE.md` `### ADR registry` — two additions: sibling-script restore trap with the universal ordering rule (regenerate → stage+commit → check), and the three-derived-tails enumeration pointing at #4002 as the template.

## Definition of Done

- [x] Docs only; no secrets, no PII.


JiRaska:
## Review: no findings — both claims verified, and I hit this trap today

**The restore trap is real.** `check-eu-ai-act.sh:25`:

```bash
git -C "$ROOT" --no-pager diff -- "$DOC" >&2 || true
git -C "$ROOT" checkout -- "$DOC" 2>/dev/null || true
exit 1
```

So running the check after regenerating but before staging **discards the regeneration** and reports the file as stale. The natural workflow — regenerate, verify with the check, then commit — destroys the work and tells you nothing was done. The ordering rule this PR documents (regenerate → stage+commit → check) is not a style preference; it is the only order that survives.

**The enumeration is right.** `agents.yaml` is read by **42** generators: 40 bundle generators under `gitops/components/*/gen-*`, plus `gen-eu-ai-act.py` and `gen-ai-governance-snapshot.py`. That matches the three tails this PR names (the doc, the 40 bundles + their pod-roll checksum annotations, and the snapshot).

**Why this is worth having in `CLAUDE.md` rather than an issue:** I walked into exactly this today on #3772. `agents.yaml` gained the case-coordinator charter, the bundle-sync gate failed naming `agent-opa-bundle.yaml`, so I regenerated *that one* and pushed — and the next run failed again, this time on the service bundles, because `agents.yaml` is a checksum input to all 26 per-service generators, not just the agent one. The first error message pointed at one bundle and read as "one bundle is stale", when the true statement was "every bundle is stale". A reader who knows there are three tails does the whole set the first time; a reader who does not follows the error messages one at a time.

The same shape produced #4002, which regenerated a subset against an already-moving base and had to be closed.

Docs-only, `CLAUDE.md` only. Automated agent review, not an approval — the file is governance-adjacent so the merge guard will ask for a human regardless.

PR #4010, opened 2026-08-07
feat(transaction): resolve card merchants to a name and a shop location (D5)

JiRaska:
Server half of D5 from the customer-app contract draft (openbank-app#396). A `merchant_catalog` table maps a normalised acquirer descriptor to the merchant's public trading name and shop coordinates; `GET /api/v1/transactions` resolves it once per page.

Honesty rules, enforced not just documented:
- Exact match on the normalised key — no fuzzy matching, so `ALZA VETERINARY` can never collect Alza.cz's coordinates. No match = no `merchant` object = client renders what it renders today.
- `NON_NULL`, so a response with nothing to add is byte-identical to before. Serialising `"merchant": null` broke the sepa-payment Pact verification — that check caught a real wire change.
- `description` passed through untouched; disputes and SPAYD use the raw acquirer text.
- Geo null for card-not-present merchants; CHECK constraint keeps lat/lon both-or-neither.
- Catalogue holds public business data (where a shop is), never cardholder location.

11 normalisation tests. Service tests, detekt and ktlint green locally.


JiRaska:
## Review: no blocking findings — the load-bearing claims check out

Verified the three that carry real consequences rather than reading the description:

**Exact match, no fuzzy.** `MerchantCatalogRepository:34` — `find("descriptorKey in ?1", keys)`, an `IN` over normalised keys. No `LIKE`, no similarity. So `ALZA VETERINARY` cannot collect Alza.cz's coordinates, which is the failure this rule exists to prevent — and it is enforced by the query shape, not by a convention someone can later relax.

**One query per page, not per row.** `TransactionResource:80` — `merchantCatalog.findByDescriptors(page.data.map { it.description })` returns a `Map`, and `page.toResponse(merchants)` consumes it. A statement page of 50 rows is one lookup, not 50. Worth noting because the natural implementation of this feature is the N+1 one.

**The privacy boundary is structural, not documentary.** `V16` holds `descriptor_key`, `clean_name`, `logo_url`, `category`, `lat`/`lon`, `city`, `country` — the *shop's* location keyed by acquirer descriptor. There is no cardholder column, and no join that could introduce one: the key is the merchant descriptor, so the table cannot express "where this customer was". The `merchant_catalog_geo_complete` CHECK (lat and lon both-or-neither) plus the range CHECKs mean a half-populated coordinate cannot exist either.

That last distinction is the one I would have looked hardest at, and it holds: this is "where the shop is", not "where the cardholder was". Those are one join apart in a design that does it carelessly, and this one does not offer the join.

## The Pact detail is worth keeping in the description

> `NON_NULL`, so a response with nothing to add is byte-identical to before. Serialising `"merchant": null` broke the sepa-payment Pact verification — that check caught a real wire change.

That is the interesting sentence in this PR. A nullable field serialised as `null` *is* a wire change even though every consumer would ignore it, and the provider replay caught it. This repo has spent a lot of effort establishing that a committed pact nobody replays is worthless (#2327, #2338); here the replay earned its keep on a change nobody would have flagged by eye.

## Not verified

The catalogue's *contents*. The schema cannot hold cardholder data, but nothing here checks that whoever populates `merchant_catalog` puts genuinely public business data in it — that is an operational property, not a schema one. Worth a line in the runbook about where the data comes from and who reviews it, if that does not exist yet.

Automated agent review, not an approval. transaction-service is money-path, so it needs human review regardless.

PR #4020, opened 2026-08-07
feat(card-issuance): make the card controls real, and add category rules (D3)

JiRaska:
D3 (per-category limits + gambling block) from openbank-app#396, plus the thing that had to exist first.

**Found while implementing:** the channel controls shipped since V5 — contactless / online / ATM / abroad — are stored, returned by the API, and **read by nothing**. There is no card authorization path in the repo at all. A customer switching off "payments abroad" gets a 200 and no protection.

This adds the decision point: `CardAuthorizationPolicy` (pure, 15 unit tests, every branch reachable without an acquirer) behind `POST /cards/{id}/authorizations`, plus `GET|PUT /cards/{id}/category-limits` and a served `GET /cards/category-taxonomy` so no client hardcodes MCC sets.

Deliberate calls, argue with any of them: OTHER is limitable but never blockable; CHIP_AND_PIN has no toggle (lock-out); missing acquirer country = domestic; unknown card declines rather than 404s; `spendTracking: false` because per-category spend is not tracked yet and a zero must not render as a measured ring.

Spend totals are caller-supplied (no authorization history here) — that boundary is in the threat model §4a, not implicit. Service tests, detekt, ktlint and route-conformance green locally.

### DIFF
--- a/docs/threat-models/openbank-card-issuance-service.md
+++ b/docs/threat-models/openbank-card-issuance-service.md
@@ -59,6 +59,37 @@ All mutations require `Idempotency-Key` or `X-Operator-Id` header; resource-leve
 | **E**oP | Viewer escalates to issue or block | Distinct roles enforced at JAX-RS layer (`@RolesAllowed`) + OPA; deny-by-default |
 | **T**ampering | Race between `suspend` and `block` on same card | Domain `require` guards throw `IllegalArgumentException` on illegal state; a concurrent block on a SUSPENDED card is valid by design; concurrent suspend+block both succeed on ACTIVE → last writer wins idempotently (both move toward frozen state) — acceptable; optimistic locking would make this exact (see §5) |
 
+## 4a. Card authorization decision point (D3) — STRIDE supplement
+
+Three new surfaces: `GET /api/v1/cards/category-taxonomy`, `GET|PUT /api/v1/cards/{id}/category-limits`,
+and `POST /api/v1/cards/{id}/authorizations` — the decision that answers whether one card
+authorisation is approved.
+
+**The pre-existing defect this closes.** The channel controls (`contactlessEnabled`,
+`onlineEnabled`, `atmEnabled`, `abroadEnabled`) have been stored on `cards` and returned by the API
+since V5, and **no code anywhere read them to decide anything**. A customer who switched off
+"payments abroad" changed a boolean, got a 200, and their card kept working abroad. That is worse
+than not offering the control: a security control that reports success without acting invites the
+customer to stop taking other precautions. Everything below exists to make those toggles real.
+
+| STRIDE | Threat | Mitigation |
+|---|---|---|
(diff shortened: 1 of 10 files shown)


JiRaska:
## The premise is verified — the channel controls really do enforce nothing

I checked the claim that made this PR necessary, because it is the kind of statement that is easy to assert and easy to be wrong about. It holds, and it is stronger than "no authorization path exists":

```
conditional uses of contactlessEnabled / onlineEnabled / atmEnabled / abroadEnabled
in any openbank-*/src/main/**/*.kt on origin/main:   ZERO
```

The fields appear in 10–15 files each, which is what makes this easy to miss — but every occurrence is plumbing: `CardPorts`, `CardEntity`, `CardMapper`, `CardEvents`, and in `CardService.kt` they are positional arguments passed through (lines 120, 128). Nothing branches on them. The only file in card-issuance matching `authoriz` is `CardDelegationResource.kt`, which is delegation, not card authorization.

So a customer who switches off "payments abroad" gets a `200`, sees the toggle stay off, and has exactly the protection they had before: none. That is worse than not shipping the toggle, because the UI makes a promise the platform does not keep.

## This is the third instance of the same shape today

Worth naming, because it is not three coincidences:

1. **`dailyLimit` / `monthlyLimit` on delegation** — stored, echoed in `DelegationResponse`, enforced nowhere. #3613 landed today and took the honest option: refuse the fields rather than keep claiming them.
2. **`AccountAuthorization.dailyLimit`** — stored and API-exposed, read by nothing in `src/main` (surfaced while reviewing #3633).
3. **Card channel controls** — this PR.

The common mechanism is that a field can travel request DTO → domain → entity → response and be *fully plumbed* without any code ever comparing it. Every layer looks correct in isolation; the absence is only visible if you ask "does anything branch on this?" — which no test asks, because a test that sets the flag and reads it back passes.

**This PR takes the right option for its case**: it adds the decision point (`CardAuthorizationPolicy`, pure, behind `POST /cards/{id}/authorizations`) rather than removing the controls. Given card controls are a headline customer feature, that is the better half of the choice #3613 faced — but it is the same choice, and both are honest. What would not have been honest is leaving them as-is.

## On the deliberate calls listed in the description

`spendTracking: false` is the one I would keep as written. Rendering a zero as a measured ring when per-category spend is not tracked would manufacture exactly the false assurance this PR exists to remove — the same error one layer up. Stating it in the response rather than defaulting silently is the right shape.

`OTHER` limitable but never blockable, and `CHIP_AND_PIN` having no toggle, both read as lock-out protection rather than arbitrary policy; worth keeping the reasoning in the threat model where it is, since a future reader will otherwise "fix" them.

## What I did not verify

That the caller-supplied spend totals boundary is safe in practice. The description is explicit that no authorization history exists here and totals come from the caller, and the threat model carries it — but "the caller supplies the number the limit is checked against" is a trust boundary worth a human's eyes, not mine. That is the one thing I would read closely before merging.

Automated agent review, not an approval. card-issuance is money-path (it holds the PAN vault key), so this needs human review regardless.

PR #4030, opened 2026-08-07
fix(finrep): open the management port the probes have always asked for

JiRaska:
finrep-service's pod can never become ready: the gitops manifest probes and scrapes `port: management` (8085), and `application.yaml` never enabled Quarkus's management interface, so kubelet got connection refused.

The probe **path** was fine — without a management port Quarkus keeps `/q/health` on the main HTTP port, so the endpoint exists, just not on 8085. That is why this looks like a healthy service that never passes readiness.

Fixed on the service side, not the manifest side, because the fleet convention (ledger, transaction, …) is `management.enabled: true, port: 8085, root-path: /q` — the manifest was written against it and is right. Retargeting the manifest to 8140 would have made finrep the odd one out and broken the scrape convention to work around four missing lines.

`check-probe-port-listener`: 1 finding → clean across all 113 probed/scraped ports. `%test` already disables management, so tests are unaffected (green locally).

Found while working on #4010, whose gitops gate this was blocking.

### DIFF
--- a/openbank-finrep-service/src/main/resources/application.yaml
+++ b/openbank-finrep-service/src/main/resources/application.yaml
@@ -23,6 +23,20 @@ quarkus:
 
   log:
     level: INFO
+
+  # The management interface the Kubernetes probes and the Prometheus scrape target.
+  #
+  # This block was missing, and its absence is why the finrep pod could never become ready: the
+  # gitops manifest probes and scrapes port 8085 (`port: management`), Quarkus never opened it, and
+  # kubelet got connection refused. Without a management port Quarkus keeps /q/health on the main
+  # HTTP port, so nothing was wrong with the probe PATH — only with the port it was asked for.
+  # 8085 matches every other service in the fleet (ledger, transaction, …), so the manifest was
+  # right and the service configuration was the half that never landed.
+  management:
+    enabled: true
+    port: 8085
+    host: 0.0.0.0
+    root-path: /q
   smallrye-openapi:
     path: /q/openapi
   swagger-ui:


JiRaska:
## This contradicts #4017, which merged at 09:03 — and the evidence mostly favours #4017

These two PRs make directly opposed factual claims about the same service, so I checked both against `main` and the cluster rather than picking one.

**#4017** (merged) said port 8085 *is* open because #3686 moved the management block into `openbank-libs-runtime`'s shared `application.yaml`, and taught `check-probe-port-listener.py` to read that base. **This PR** says management "was never enabled" and the pod "can never become ready", and re-declares the block in finrep.

### What I verified

- **#3686 is merged** (08:34) and it **removed** `management: {enabled, port: 8085, host, root-path}` from `openbank-finrep-service/src/main/resources/application.yaml`. So finrep genuinely has no management block of its own today — that part of this PR's diagnosis is right.
- **The shared base does declare it**: `openbank-libs-runtime/src/main/resources/application.yaml` carries `management: port: 8085, root-path: /q` on `main`, and finrep depends on `:openbank-libs-runtime`.
- **The inheritance is tested, and specifically for management** — `FinrepBootSmokeTest:48-52` asserts `quarkus.management.port == 8085`, `root-path`, and `enabled` resolve. (I initially thought the test only asserted an HTTP header; that was my misreading — it asserts the header *and* the management triple.)
- **The running pod is `1/1 Running`** and has been for **14 hours**, so it predates #3686 entirely. It is serving the old image with the old, explicit config — which means it is evidence about neither PR.

So "the pod can never become ready" is not true of the pod that exists, and the mechanism this PR says is missing is present and covered by a test.

### The gap that is real, and why it is worth saying out loud

`FinrepBootSmokeTest` proves the **config values resolve**. It does not prove the management interface is **listening on 8085**: the readiness assertion at line 25 goes through RestAssured against the `@QuarkusTest` HTTP port, not 8085. Config resolution and a bound socket are different claims, and `quarkus.management.enabled` has build-time semantics — the value resolving inside a test JVM is strong evidence that augmentation sees it, not proof that the packaged fast-jar opens the port.

Nobody has deployed finrep on the post-#3686 config yet. Until that happens, "8085 is open" rests on a config assertion, and this PR's counter-claim rests on reasoning from the removed block. Neither side has the observation.

### Suggestion

Rather than merging this and re-declaring the block — which would restore the duplication #3686 deliberately removed, and hand the next service adopting the base the same false red on an enforced gate — the cheaper and more decisive move is to **make the smoke test assert the socket**: hit `http://localhost:8085/q/health/ready` directly instead of only reading config values. That converts the open question into a test, and it would fail loudly today if the interface really is not opening.

If that assertion goes red, this PR is right and should merge as-is. If it stays green, the block should stay out of finrep and this PR closed.

I would not merge it on the current evidence, and I would not close it either — it is a reasonable challenge to an assumption that has never been observed end to end. It just needs one test to settle it.

Automated agent review, not an approval.

PR #4039, opened 2026-08-07
feat(card-issuance): model the single-use card lifecycle (D1 server preparation)

JiRaska:
Server half of D1 from openbank-app#396. Adds terminal status `CONSUMED`, `CardClosedReason`, `expiresAt`, the `consume()`/`expireUnused()` transitions, and migration V9.

`CONSUMED` is deliberately not CANCELLED: "cancelled" currently covers a customer closing a card, a lost card, and a disposable card doing exactly what it promised. One of those is a security signal — "your disposable card was just used" is this product's fraud alarm and must be distinguishable from routine closure.

An unused card that times out becomes EXPIRED / `VALIDITY_EXPIRED`, never CONSUMED — nothing was spent, and the difference is what the customer is told.

`CONSUMED` joins `TERMINAL_STATUSES`, which excludes a card from PAN-vault backfill. `CardPanVaultBackfillTest` pins that set and failed here; updated deliberately with the reasoning, not relaxed.

**Explicitly not included:** the authorize-once guarantee itself, which lives at the card processor. Until that is configured a SINGLE_USE card is a virtual card with a validity window — the status machine is ready, the guarantee is not. Noted in the threat model §4b so nobody ships "authorises once" copy on the strength of this.

8 lifecycle tests; service tests, detekt and ktlint green locally.

### DIFF
--- a/docs/threat-models/openbank-card-issuance-service.md
+++ b/docs/threat-models/openbank-card-issuance-service.md
@@ -90,6 +90,31 @@ reads only this service's own tables.
 which is the defect, so a rollback should be paired with disabling the customer-facing toggles
 rather than leaving them visibly ineffective.
 
+## 4b. Single-use card lifecycle (D1) — STRIDE supplement
+
+Adds the terminal status `CONSUMED`, a `closedReason`, and an `expiresAt` validity window for
+SINGLE_USE cards. No new endpoint: the card list and detail responses carry the new fields.
+
+**Why a distinct status rather than reusing CANCELLED.** Until now "cancelled" covered a customer
+closing a card, a card reported lost, and a disposable card doing exactly what it promised. Those
+are three different things a customer is owed three different sentences about — and one of them is
+a security signal. "Your disposable card was just used" is the fraud alarm for this product; it
+cannot be distinguished from routine closure if the status is the same.
+
+| STRIDE | Threat | Mitigation |
+|---|---|---|
+| **E**oP | A consumed card is re-provisioned with a live PAN and spends again | `CONSUMED` is in `TERMINAL_STATUSES`, which the PAN-vault backfill excludes. A pinned test asserts the exact contents of that set, so growing it silently is not possible — that assertion caught this change and had to be updated deliberately |
+| **T**ampering | A card is marked CONSUMED without having been used, hiding a decline or a failure | `consume()` requires `cardType == SINGLE_USE` **and** `status == ACTIVE`, and is driven by the processor's lifecycle event rather than by a customer or operator request. Any other type or status throws |
+| **R**epudiation | The customer cannot tell why their card stopped working | `closedReason` distinguishes `SINGLE_USE_CONSUMED` from `VALIDITY_EXPIRED`, `LOST_OR_STOLEN` and `CUSTOMER_CANCEL`; an unused card that times out becomes EXPIRED, never CONSUMED, because nothing was spent |
+| **I**nfo disclosure | Revealing the PAN of a card that already spent | Unchanged and still correct: `CONSUMED` is terminal, so the existing "card not live" guard on secure details refuses it |
+| **D**oS | A disposable card issued and forgotten stays a live PAN indefinitely | `expiresAt` bounds the window even if the card is never presented; a partial index supports finding those cards without weighing on the hot path |
+
+**Not in this change, and load-bearing:** the authorize-once guarantee itself lives at the card
+processor. This models the outcome so the customer can see it; it does not enforce single use. Until
+the processor is configured, a SINGLE_USE card is a virtual card with a validity window — the status
+machine is ready, the guarantee is not. Anything that presents it to customers as "authorises once"
+before the processor side lands would be claiming a control the bank does not yet have.
+
 ## 5. Residual risks / assumptions
 
 - **No optimistic locking today.** `Card` lacks a `version` column; two concurrent lifecycle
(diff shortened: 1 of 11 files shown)


JiRaska:
Read against the repo's own `Instant.EPOCH` lesson before approving — worth stating why it does not apply here rather than leaving it silent.

```kotlin
fun consume(now: Instant = Instant.EPOCH) = also { ... }.copy(updatedAt = now, ...)
fun expireUnused(now: Instant = Instant.EPOCH) = also { ... }.copy(updatedAt = now, ...)
```

`Instant.EPOCH` as a default has bitten this repo before — `AuditEvent.timestamp` and `FlagExposure.timestamp` both defaulted to it, 23 of 25 call sites took the default, and no test caught it because the assertions checked non-nullity rather than recency (#3882).

**This one is not that, and I checked why:** neither `consume()` nor `expireUnused()` has a caller anywhere in this diff. No use case, resource, or consumer invokes them — the PR is explicit that this is "D1 server preparation," domain-only, with the processor-driven trigger not yet wired. Every call site in the diff is a test, and all nine pass `now` explicitly rather than relying on the default.

So the default is inert today, but it is a footgun for whoever adds the real caller in D2: `card.consume()` with no argument compiles, returns a card CONSUMED at 1970-01-01, and nothing here would catch it — there is no recency assertion pattern in this PR to inherit, because there is no production caller yet to assert against. Worth requiring `now: Instant` (no default) before D2 lands, or leaving a comment on the function stating that the default exists only for test convenience and must never reach a real call site.

Everything else here checks out against the diff, not just the description:
- `TERMINAL_STATUSES` is pinned by `CardPanVaultBackfillTest`, and the PR states that test failed and was updated deliberately — the change-detector working as intended, not relaxed.
- `consume()`'s two `require`s (`SINGLE_USE` + `ACTIVE`) are each covered by a dedicated test iterating every other enum value, not just one negative case.
- The STRIDE supplement's claim that the authorize-once guarantee is NOT in this change is honest about scope — nothing in the diff claims otherwise, and the rollback note (a CONSUMED card must be remapped to CANCELLED before an older build can read it) is the right kind of detail to write down before it is needed.

Not merging — `openbank-card-issuance-service` is not in `rules.yaml: money_path_services`, but it issues live PANs and the diff includes a migration and a threat-model STRIDE addition, which is squarely a human's call regardless of the automated gates.

JiRaska:
Reviewed. **One blocking defect**, otherwise sound.

### The update path silently drops both new columns

`CardRepositoryImpl.applyFrom(card)` is the field-by-field copier used on the **existing-row branch** of `save()`. Verified by enumerating what it actually copies on this branch:

```
status maskedPan cardholderName embossedName expiryDate dailyLimitMinorUnits
monthlyLimitMinorUnits contactlessEnabled onlineEnabled atmEnabled abroadEnabled
currency deliveryAddress activatedAt blockedAt blockedReason updatedAt
```

18 fields, and **neither `expiresAt` nor `closedReason`**. Both are new in this PR (39 added lines reference them; neither exists on `main`), and `toEntity()` — the INSERT branch — was updated. So the two persistence paths now disagree.

Failure scenario once D2 wires the processor event: `card.consume(now)` → `save()` → the row is written with `status='CONSUMED'` and **`closed_reason` still NULL**. The justification for the whole change — CONSUMED must be distinguishable from CANCELLED so the fraud alarm reads correctly — rests on a column that never persists on any transition. Same for setting `expiresAt` on an already-issued card, which leaves the new partial index `idx_cards_expires_at` indexing nothing.

Nothing can see it today: `SingleUseLifecycleTest` is pure-domain and `CardPanVaultBackfillTest` mocks the repository. Two lines fixes the copier; the durable fix is a real-DB IT that round-trips a consumed card, the `ConsentRevocationOutboxIT` pattern. That is also where Codecov's 25% on `CardMapper.kt` is pointing.

### Non-blocking, but name it before D2 lands

The terminal-state guard is **read-then-write, not compare-and-set**. `consume()` guards `status == ACTIVE` and `expireUnused()` guards `status in LIVE_STATUSES`, so CONSUMED is genuinely terminal in-model — but `save()` does `find(id)` then mutates the managed entity, with no `WHERE status='ACTIVE'` predicate and no `@Version`. Two concurrent processor events can both read ACTIVE and both pass `require()`. Today that costs only a duplicated outbox row, since the status ends CONSUMED either way — so **not** a live double-spend. It is the gap D2 inherits, and the PR is honest that the authorize-once guarantee lives at the processor.

### Checked and clean

- **Reactive Panache `persist()` trap**: avoided, and documented in place — `persist` only on a genuine insert, mutate-managed-entity on update.
- **Migration**: `V9` is free on `main` (V1–V8 present), purely additive (`ADD COLUMN IF NOT EXISTS` ×2 + partial index), rollback note present and correct about the CONSUMED→CANCELLED remap.
- **Domain state machine**: 8 tests, negatives iterating every other enum value. Good coverage of the layer it covers.

Minor: `CardMapper` uses `runCatching { CardClosedReason.valueOf(r) }.getOrNull()`, so an unrecognised DB value degrades silently to "no reason" rather than failing loudly. Defensible for forward-compatibility, worth a comment saying so deliberately.

JiRaska:
Pushed the fix for the blocking finding (`af9b761ae`).

**The two lines**, plus a test so it cannot recur on the next column. `applyFrom` now writes `expiresAt` and `closedReason`, and moved next to `toEntity` in `CardMapper.kt` so a plain unit test can call it without a Vert.x context.

Naming those two fields in a test would only close this instance, so `CardMapperUpdateParityTest` compares the two persistence paths over **every** mutable property by reflection. A column added to one path and not the other fails whether or not anyone remembers the test exists. `notUpdatable` is the deliberate half — `id`, `idempotencyKey`, `partyId`, `accountId`, `productCode`, `cardType`, `network`, `createdAt` and the two vault columns — so adding a column forces an explicit decision about which side of the line it falls on rather than defaulting to silence.

**Falsified both ways, and the first attempt was not good enough.** Removing the two lines failed naming only `closedReason`:

```
applyFrom() does not write [closedReason], but toEntity() does.
```

`expiresAt` matched by coincidence, because my fixture used the same value on both rows — the exact case the fixture's own comment claimed to rule out. Caught only by reading *which* fields the failure named instead of accepting the non-zero exit. With the rows differing on every updatable column:

```
applyFrom() does not write [closedReason, expiresAt], but toEntity() does.
```

Against the fixed code: 1 test, 0 failures, 0 skipped (read from the JUnit XML, not the console). `:ktlintCheck` and `:detekt` green on the module.

**Left open deliberately** — the second finding stands and is not something I should decide unilaterally: the terminal-state guard is read-then-write, not compare-and-set. `save()` does `find(id)` then mutates the managed entity, with no `WHERE status='ACTIVE'` predicate and no `@Version`, so two concurrent processor events can both read ACTIVE and both pass `require()`. Today that costs a duplicated outbox row rather than a double-spend, since the status ends CONSUMED either way — but it is the gap D2 inherits, and the PR is honest that the authorize-once guarantee lives at the processor. Worth deciding before D2 rather than after.

Issue #3053, opened 2026-08-01
app-status.json drift is advisory over a generated artifact, so the console ships a stale customer-app dossier (0.4.0 vs 0.9.7)


JiRaska:
## What

The `app-status` check reports drift on every PR and does not block, so the drift is mergeable:

```
app-status check (advisory→ENFORCE) against app-status.json
  derived: version=0.9.7 useFakeData=false edge=https://customer.open-bank.tech
  capabilities: 18 (live=13, partial=4, planned=1)
::error title=app-status derived drift::the committed app-status.json no longer matches openbank-app source (1 field(s)). Regenerate & publish.
  derived-block drift:
    version: committed="0.4.0" → regenerated="0.9.7"
```

Observed on #3037, which touches nothing related to the customer app.

## Why this shape is wrong, not just noisy

`openbank-admin-ui/app-status.json` is a **generated** artifact — `generate-app-status.mjs` derives
it from `openbank-app` source. On a hand-written artifact an advisory red means "someone should
look". On a generated one it means "the committed document does not match reality", and there is no
judgement left to exercise. Advisory only makes the mismatch mergeable.

This is the same failure the repo already recorded for `eu-ai-act-registry`: it went red twice on
#2156, the PR merged anyway, and the EU AI Act inventory omitted an AI system until someone
regenerated it (#2216).

The concrete cost here: the admin console's customer-app dossier claims version `0.4.0` while the
app is on `0.9.7` — five minor versions of capability drift in a page whose whole purpose is to state
what the customer app can currently do (ADR-0074).

## Two separable problems

1. **The committed artifact is stale.** Regenerate and publish:
   `node openbank-admin-ui/scripts/generate-app-status.mjs --app-repo ../openbank-app`
2. **The gate cannot stop it recurring.** Deciding between: enforce it (and accept that a publish PR
   from `openbank-app` CI must land before an unrelated admin-ui PR can merge), or accept the drift
   explicitly and stop printing `::error` on every unrelated PR. The current middle state trains
   everyone to ignore a red that is sometimes real — the alert-fatigue pattern already identified as
   a root cause in this repo.

Note the enforcement direction interacts with a known constraint: a `paths:`-filtered workflow can
never be a required status check, so if this is to block, the binding copy of the check has to live
in the unconditional `Validate manifests` job in `ci.yml`, the way `adr-registry` does.

Found while merging #3037.

PR #4053, opened 2026-08-07
ci(app-status): scope the dossier check to drift a PR could have caused

### LINKED ISSUE #3988: Customer-app dossier reddens every open PR on each openbank-app release — 5 issues closed, cause never addressed
## What

`Customer-app dossier` goes red on **every open PR** each time `openbank-app` ships, and stays red until a human regenerates `app-status.json` or the nightly refresh ticks. It has recurred at least five times in six days. Each recurrence gets its own issue, gets fixed by regenerating, and gets closed — so the class reads as solved five times over while nothing about it changed.

```
#3212  CLOSED  Customer-app dossier red on every PR: app-status.json pinned at 0.…
#3552  CLOSED  Customer-app dossier red on every PR again: app-status.json pinned…
#3818  CLOSED  Customer-app dossier check red repo-wide: committed derived versio…
#2101  CLOSED  app-status dossier auto-publish is a silent no-op: PLATFORM_RW_TOKEN…
#3053  CLOSED  app-status.json drift is advisory over a generated artifact…
```

Regenerations on `main`, all of them the same edit:

```
08-01 22:15   08-03 06:09   08-05 11:44   08-06 11:09   08-06 16:48   08-07 06:48
```

## The mechanism, measured

The refresh automation exists and works — `app-status-refresh.yml` runs on `cron: "37 4 * * *"` and its last six runs all succeeded. The problem is not that it is broken. It is that **its cadence and `openbank-app`'s release cadence differ by an order of magnitude**:

```
openbank-app releases      0.16.1  08-06 14:42
                           0.17.0  08-06 22:22
                           0.18.0  08-07 03:01     <- three in ~12 h
open-bank-oss refresh      once daily, 04:37 UTC
```

So after every release there is a window — up to 24 h — in which every PR in this repo is red on a check about a **file none of them touched**, in a **repo none of them changed**.

Today that is 10 of the 55 open PRs I could enumerate.

## Why it matters beyond the noise

`Customer-app dossier` is not a required context, so it does not block the ruleset. It **does** make `mergeStateStatus` UNSTABLE, which makes auto-merge refuse — and the documented way out of UNSTABLE is the admin override, which nobody should be reaching for. So a cadence mismatch in another repository quietly pushes contributors toward a bypass.

And it is the alert-fatigue shape on the PR page: a check that is red for reasons unrelated to the diff, repeatedly, is a check people learn to skip — including on the occasions it is right.

## What to decide

Not proposing one; they trade off differently:

1. **Push, don't poll.** `openbank-app` fires a `repository_dispatch` on release and the refresh runs immediately. Closest to correct, and blocked by the same missing token as #2101 — worth confirming whether a dispatch-scoped token is a smaller ask than `PLATFORM_RW_TOKEN`.
2. **Raise the cadence.** Hourly instead of daily shrinks the window ~24x. Cheap, no new secret, does not eliminate the class.
3. **Scope the check to PRs that could have caused the drift.** This is the one I would argue for on principle: the drift is *never* attributable to the PR being tested, so failing that PR attributes it wrongly. Let a scheduled job own freshness and fail loudly on its own schedule (raising or refreshing an issue, as `fleet-attestation.yml` does), and let the PR check assert only what the PR can affect.

Option 3 does not weaken anything #3053 established. That issue's point was that advisory-over-generated makes drift *mergeable*; making the check fail on a schedule rather than on unrelated PRs keeps it enforcing while pointing the failure at someone who can act on it.

## Acceptance

- A release in `openbank-app` does not turn unrelated PRs in this repo red.
- Drift is still detected and still fails something — with the failure addressed to a party who can fix it, not to whoever happened to open a PR next.

JiRaska:
## Linked issues

Closes #3988

## What is wrong

`Customer-app dossier` regenerates `app-status.json` from `openbank-app`'s source and fails on drift — on **every** pull request.

That drift is caused by `openbank-app` shipping. It is never caused by a PR in this repository. So the check reports a real defect to whoever happened to open a PR next.

Measured 2026-08-07:

```
openbank-app releases      0.16.1  08-06 14:42
                           0.17.0  08-06 22:22
                           0.18.0  08-07 03:01      three in ~12 h
open-bank-oss refresh      cron "37 4 * * *"        once daily

result                     10 of 55 open PRs red on a file none of them touched
```

The class has recurred five times in six days. Each recurrence got its own issue, was fixed by regenerating, and was closed — so it reads as solved five times over:

```
#3212  #3552  #3818  #2101  #3053     all CLOSED
```

Regenerations on `main`, all the same edit:

```
08-01 22:15   08-03 06:09   08-05 11:44   08-06 11:09   08-06 16:48   08-07 06:48
```

## Why this loses no coverage

Freshness is already owned, and already loud. `app-status-refresh.yml` regenerates daily and **opens its own PR** on drift — #3747, #3867, #3917 and #3952 are those PRs. The automation is not broken; its cadence simply differs from `openbank-app`'s by an order of magnitude.

What that scheduled job cannot see is a PR that hand-edits `app-status.json` or changes `generate-app-status.mjs`. That is precisely what stays checked here.

So the split is: drift the PR could have authored → checked at PR time; drift the PR cannot have authored → checked on a schedule, by a job that fails on its own behalf and raises a PR someone can merge.

This does not weaken what #3053 established. That issue's point was that advisory-over-generated makes drift *mergeable*. Nothing here becomes advisory — the comparison still exits 1 on drift, `--enforce` and the `set -o pipefail` are untouched.

## The job stays unconditional

Deliberately. A gate behind a conditional job is **absent**, not skipped: the aggregate check goes green and nothing says a gate was not consulted (#3629, where `check-dockerfile-no-build-stage.py` could not run on a Dockerfile-only PR for exactly this reason).

Only the comparison steps are gated. The `Customer-app dossier` context always reports.

## Verification

Four cases, run in a throwaway worktree against the real tree:

| PR touches | expected `owns` | got |
|---|---|---|
| an unrelated `.rego` (the ten red today) | false | false |
| `openbank-admin-ui/app-status.json` | true | true |
| `openbank-admin-ui/scripts/generate-app-status.mjs` | true | true |
| neither | false | false |

The two `true` cases are the ones that make this worth having — a scoping change that only satisfied the "don't fire" half would have turned the check off rather than aimed it.

Also checked:

```
actionlint      branch 0 findings, main 0 findings — identical sets
YAML            parses; app-status job has 9 steps, 5 scope-guarded
run-step size   largest unchanged at 18955 chars
```

## Two things noticed in passing, not changed here

**`BASE_REF` goes through `env:`** rather than being interpolated into the script. The neighbouring `changes-ui` job still interpolates `github.base_ref` directly. That is not exploitable for a base ref, so it is left alone — but new code should not copy the pattern.

**`auto-deploy.yml`'s `can-i-deploy` step is 18955 chars against a 19000 limit.** That is 45 characters of headroom on a gate whose whole purpose is to stop an oversized `run:` making the workflow unparseable — and the real GitHub ceiling is ~20054, at which point the workflow silently stops existing for everyone. One more comment paragraph in that step is enough. Worth a separate look; this PR does not touch it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -450,12 +450,51 @@ jobs:
       APP_RO_TOKEN: ${{ secrets.OPENBANK_APP_RO_TOKEN }}
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          fetch-depth: 0
+
+      # Does THIS PR touch the dossier or its generator?
+      #
+      # Drift between the committed app-status.json and openbank-app's source is caused by
+      # openbank-app SHIPPING, never by a PR here. Comparing on every PR therefore reports a
+      # real defect to whoever happened to open one next: measured 2026-08-07, openbank-app
+      # released three times in twelve hours (0.16.1, 0.17.0, 0.18.0) against a daily refresh
+      # tick, and 10 of 55 open PRs were red on a file none of them touched. It has recurred
+      # five times in six days (#3212, #3552, #3818, #2101, #3053 — all closed by regenerating).
+      #
+      # Freshness is already owned, and already loud: app-status-refresh.yml regenerates daily
+      # and opens its own PR (#3747, #3867, #3917, #3952 are those PRs). What that job CANNOT
+      # see is a PR that hand-edits app-status.json or changes the generator — which is exactly
+      # what stays checked below. So this narrows the scope to the only drift a PR can author,
+      # and loses no coverage (#3988).
+      #
+      # The JOB stays unconditional on purpose. A gate behind a conditional job is absent
+      # rather than skipped, the aggregate check goes green, and nothing says a gate was not
+      # consulted (#3629). Only the comparison is scoped; the context always reports.
+      - name: Could this PR have caused drift
+        id: scope
+        shell: bash
+        env:
+          BASE_REF: ${{ github.base_ref }}
+        run: |
+          set -euo pipefail
+          git fetch origin "${BASE_REF}:refs/remotes/origin/${BASE_REF}" --no-tags
+          N=$(git diff --name-only "origin/${BASE_REF}...HEAD" -- \
+                openbank-admin-ui/app-status.json \
+                openbank-admin-ui/scripts/generate-app-status.mjs | wc -l | tr -d ' ')
+          echo "owns=$( [ "$N" -gt 0 ] && echo true || echo false )" >> "$GITHUB_OUTPUT"
+          echo "files this PR changes in the dossier's own scope: $N"
+
+      - name: Not this PR's drift
+        if: steps.scope.outputs.owns == 'false'
+        run: |
+          echo "::notice title=app-status::This PR touches neither app-status.json nor its generator, so it cannot have caused dossier drift. Freshness against openbank-app is owned by app-status-refresh.yml, which opens its own regeneration PR (#3988)."
 
       # Read-only checkout of the customer-app source so the generator can derive
       # facts (AppConfig.kt + version.txt + app-status.yaml). openbank-app pins no
       # ref → its default branch, the published state the committed copy should match.
       - name: Checkout customer-app source
-        if: env.APP_RO_TOKEN != ''
+        if: env.APP_RO_TOKEN != '' && steps.scope.outputs.owns == 'true'
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           repository: JiRaska/openbank-app
@@ -468,7 +507,7 @@ jobs:
         run: echo "::notice title=app-status::OPENBANK_APP_RO_TOKEN not set — skipping customer-app dossier content check (ADR-0074)."
 
       - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
-        if: env.APP_RO_TOKEN != ''
+        if: env.APP_RO_TOKEN != '' && steps.scope.outputs.owns == 'true'
         with:
           # Matches the runtime image (openbank-admin-ui/Dockerfile: node:26-alpine). Pinned at 20
           # it was three majors behind what ships, and #3569's jsdom bump made that fatal:
@@ -481,7 +520,7 @@ jobs:
           cache-dependency-path: openbank-admin-ui/package-lock.json
 
       - name: Install (admin-ui, for the yaml parser)
-        if: env.APP_RO_TOKEN != ''
+        if: env.APP_RO_TOKEN != '' && steps.scope.outputs.owns == 'true'
         working-directory: openbank-admin-ui
         run: npm ci --no-audit --no-fund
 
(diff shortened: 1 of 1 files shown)


JiRaska:
Reviewed. **Clean — approve.** Narrowing a check's scope is the risky direction, so I ran it rather than reading it.

Extracted the `scope` step's `run:` from the branch's `ci.yml` with PyYAML (not retyped — a transcription is a different program) and ran it in a throwaway worktree against four real HEADs with `BASE_REF=main`:

| HEAD | printed | `owns` |
|---|---|---|
| `origin/main` unchanged | `…in the dossier's own scope: 0` | `false` |
| +1 line in `app-status.json` | `…: 1` | `true` |
| +1 line in `generate-app-status.mjs` | `…: 1` | `true` |
| +1 line in `README.md` | `…: 0` | `false` |

Both positive cases fire, so this aims the check rather than switching it off.

**What it stops catching, enumerated rather than assumed.** `generate-app-status.mjs` reads only `APP_REPO` files (`AppConfig.kt`, `version.txt`, `app-status.yaml`) plus the committed `app-status.json` via `--against`; `ADMIN_UI` / `PLATFORM_REPO` only derive those paths. So the two-file scope list is exactly complete for PR-authored drift. The one residual input outside it is the `yaml` npm package and the node version — a bump there could in principle move the output, and would be caught by the next daily refresh rather than at PR time. Not worth blocking.

**Structurally right, and this is the part that could easily have been wrong:** the job keeps `if: github.event_name == 'pull_request'` at *job* level, so `github.base_ref` is never empty under `set -euo pipefail`, and only the five comparison steps are gated. The `Customer-app dossier` context therefore always reports — avoiding the absent-gate class where a skipped job reads as a pass.

**One follow-up worth its own issue, not a blocker:** after this lands, `app-status-refresh.yml` is the sole owner of freshness, and its red is addressed to nobody — it has no issue raise-or-refresh step, unlike `fleet-attestation.yml` and `dependency-submission.yml`, which the repo added escalation to for exactly this reason. Previously the noisy PR-time red *was* the escalation path, however wrong the addressee. The vacuous-green direction is covered (it fails loudly on missing secrets, and a stale token reddens the checkout), so this is about who hears it, not whether it fires.

Minor: `fetch-depth: 0` is genuinely required — a shallow base fetch has no merge-base — but it is a full clone of a ~283 MB `.git` on every PR this job runs on.

JiRaska:
One item in this PR's body is now out of date, and it is worth correcting because it reads as an open hazard:

> `auto-deploy.yml`'s `can-i-deploy` step is 18955 chars against the 19000 gate limit … 45 characters of headroom

That is fixed in **#4120** (OPEN, `MERGEABLE`, `CLEAN`), which extracts the step into `.github/scripts/can-i-deploy-gate.sh`. The step drops from **18955 to 245 characters**; the largest remaining `run:` in the file is 10359. Fidelity was proven rather than asserted — 305 lines against the inlined 305, and diffing the extracted body against `origin/main`'s step with only the three `${{ }}` substitutions applied yields exactly one hunk, an obsolete comment that told the reader to keep itself short.

Reached this while working #3988 from the issue side; no duplicate PR was opened, and this PR's own approach is sound.

**On what remains of #3988** — this PR delivers item 1 of the three its triage specified, and the other two get *more* important once it lands, not less:

- **Item 2** — `app-status-refresh.yml` opens a regeneration PR on drift but escalates nothing. With the PR-time check correctly scoped away, that scheduled job becomes the **only** thing that notices staleness. A red scheduled workflow is addressed to nobody; this repo has that written down after `dependency-submission` sat red for three days while everything downstream quietly read a stale graph.
- **Item 3** — the cron is still `37 4 * * *` on `main`, i.e. daily, and its in-file comment justifies that with "a release happens every few weeks". The measured cadence is three releases in ~12 hours.

Neither blocks this PR. Flagged on #3988 as the live remainder so the issue is not closed on item 1 alone — the failure mode this issue exists to record is precisely a symptom being closed five times without the cause moving.

PR #4067, opened 2026-08-07
fix(gitops): repin engagement-service to an image that exists and is attested

JiRaska:
## What

One line in `openbank-infra/gitops/components/engagement/engagement-service.yaml`:

```diff
- image: …/openbank-engagement-service:sandbox-7a66c233d
+ image: …/openbank-engagement-service:sandbox-fac75d19
```

## Why

#4048 pinned `sandbox-7a66c233d`. That is a real commit, but no build ever produced an image from it:

```
$ aws ecr describe-images --repository-name openbank-engagement-service \
    --image-ids imageTag=sandbox-7a66c233d
ImageNotFoundException
```

The tag that **does** exist is `sandbox-fac75d19` — the merge commit of #4048 itself — and it is
complete. I checked the signature and attestation against that tag's **digest**, rather than taking
their presence in the repository listing as proof they belonged to it:

```
sandbox-fac75d19
  digest  sha256:4e015b458a22c1dfa0767c8392300d8306a82c3d9bad6125a76c8d4f22b719bf
  .sig    pushed 2026-08-07T16:42:33
  .att    pushed 2026-08-07T16:43:15
```

So this is a wrong sha in a pin, not a missing pipeline — the correct image was built, signed and
attested three minutes after the merge, and has been sitting unused.

## What it unblocks

Since the bad pin landed at 16:39, `Verify fleet attestations` has failed on **every** PR:

```
FLEET ATTESTATION GATE: FAIL — 1 of 61 declared image(s) not deployable.
No image-provenance policy may graduate Audit -> Enforce while this gate fails
```

The gate had been green on `main` for five consecutive daily runs before that (most recently
2026-08-07T03:38Z), so this is a clean regression with a single cause. It is not a required check,
which is why it read as background red on unrelated PRs — I found it on #4060, a one-line CNPG volume
change that touches no image at all.

## Verification after merge

The gate itself is the check — it enumerates all 61 declared images and must report no failures. Worth
confirming separately that the workload now starts: `kubectl get pod -A | grep engagement` currently
returns nothing.

Closes #4065

PR #4068, opened 2026-08-07
fix(engagement): wire the OPA PDP sidecar — main is red and the whole queue is blocked (#1797)

### LINKED ISSUE #4066: main is red: engagement-service enforces authz with no PDP sidecar, blocking every open PR (#1797)
## Summary

`main` is currently **red** on the enforced gate `authz-enforce ⇒ PDP sidecar parity (#1797)`, and because `gates (security)` aggregates into `Validate manifests` — a required check — **every open PR inherits the failure**. The queue cannot merge until this is fixed.

```
authz-pdp-parity: engagement-service enforces authz (app-default) but its workload declares
no `opa` PDP sidecar — every @Authorize endpoint fails closed (HTTP 503/422, issue #1797).
check-authz-pdp-parity: 40 workload(s) checked; 4 skipped; 1 enforce-without-PDP violation(s).
```

Observed on `main` at `6b3594b2b` (run 31191032017) and reproduced on an unrelated PR branch (#3700, run 31191453469) — the failure is identical and independent of the PR, which is how it was found.

## Cause

`6b3594b2b` — #4054, `security(engagement): add edge-service-engagement authz policy` — added `@Authorize` to `SurfaceResource.kt` (2 sites) and restamped every *existing* service's OPA bundle. `engagement` has no bundle of its own to restamp, and none was added:

```
openbank-infra/gitops/components/engagement/
  engagement-service.yaml      <- containers: [engagement-service]  (no `opa`)
  external-secret-oidc.yaml  namespace.yaml  network-policies.yaml  postgres.yaml
```

Compare `components/party/`, which carries `gen-party-opa-bundle.sh` + `party-opa-bundle.yaml` alongside its workload. `AUTHZ_ENFORCE` is set nowhere in the engagement workload, so the app default applies — hence the gate's `(app-default)`.

## Runtime impact, separate from the CI block

This is not only a gate problem. The workload is deployed (`openbank-engagement-service:sandbox-7a66c233d`), so both `@Authorize` endpoints on `SurfaceResource` are **already failing closed in the sandbox** — the PDP they call does not exist in the pod. The gate is reporting a live outage, not a hypothetical one.

## Fix

Wire the sidecar, mirroring `components/party/`:

1. `gen-engagement-opa-bundle.sh` + the generated `engagement-opa-bundle.yaml`
2. `opa` container + bundle volume in `engagement-service.yaml`

Setting `AUTHZ_ENFORCE=false` is the gate's other sanctioned exit, but it trades a closed door for an open one on a service whose PR two commits earlier exists specifically to authorize it — so it is the wrong direction here unless the surfaces are meant to be public.

## Why it was not caught pre-merge

Worth a follow-up on its own: the gate is enforced and did run on #4054's PR, so either it passed there (the `@Authorize` and the workload met only on merge) or it was red and merged anyway. Whichever it is, it is the interesting half.

JiRaska:
## Why this is urgent

`main` is red on the **enforced** gate `authz-enforce ⇒ PDP sidecar parity` (#1797). That gate lives in `gates (security)`, which aggregates into `Validate manifests` — a **required** check. So every open PR inherits the failure and the queue cannot merge until this lands. Reproduced on an unrelated branch (#3700) with an identical message, which is how it surfaced.

Closes #4066.

## What happened

#4054 (`security(engagement): add edge-service-engagement authz policy`, now main's tip) added `@Authorize` to `SurfaceResource` and restamped every **existing** service's OPA bundle. `engagement` had no bundle to restamp and none was added, leaving it the fleet's only enforce-without-PDP workload.

## The comment that made it look safe

`engagement-service.yaml`'s header listed the sidecar as deliberately deferred:

> the platform's `default allow := false` denies both actions fail-closed the moment enforcement IS wired in — this just means OPA-level policy isn't consulted yet, not that it's open.

Not open is right. Deferred is not. Measured:

- `AuthorizeInterceptor` declares `@ConfigProperty(name = "authz.enforce", defaultValue = "true")`
- `openbank-engagement-service/src/main/resources/application.yaml` overrides it nowhere — hence the gate's `(app-default)`
- `AuthorizeInterceptorTest` names the outcome: **`enforce=true, no PDP bean throws 503`**

So both `@Authorize` endpoints answer 503 instead of serving. The comment is corrected in place rather than deleted — the reasoning is the part worth not repeating.

## What this does

Mirrors `components/notifications` (the no-`rest_ext` shape: engagement's rule is in the **shared** `rest.rego`, so no per-service extension is needed):

| file | |
|---|---|
| `gen-engagement-opa-bundle.sh` | new; uses `awk 1` per source rather than a bare `sed`, since a source with no trailing newline welds the next `data:` key onto its last line and silently swallows it (#3748, fixed #3971) |
| `engagement-service-opa-bundle.yaml` | new, generated, checksum `207d5deced0ca13e` |
| `engagement-service.yaml` | `opa` container, bundle/tmp volumes, pod-roll checksum annotation, corrected header |

`AUTHZ_ENFORCE=false` is the gate's other sanctioned exit and is the wrong one here — the commit two before this exists specifically to authorize these surfaces.

## Verification

Each against the artifact, not the source:

```
check-authz-pdp-parity.py   40 workload(s), 0 violation(s)   (was 1)   exit 0
check-opa-bundle-parses.py  41 bundle(s), no unloadable policy (was 40) exit 0
```

And the decision itself, via `opa eval` over the policy **extracted from the generated ConfigMap** — the bytes the sidecar mounts, not the source tree — with principal `service-account-openbank-edge`:

```
engagement.surface.read        -> {"allow":true, "reason":"edge-service-engagement"}
engagement.surface.recordEvent -> {"allow":true, "reason":"edge-service-engagement"}
engagement.surface.delete      -> false
```

The third line is the control. The grant is scoped to the two actions ADR-0220 D1 names rather than to an `engagement.` family, so a bundle that allowed everything would fail this check — without it the first two lines prove only that something answered.

## Review notes

- authz + gitops change, so it wants a human by rule. Flagging that the *urgency* (a blocked queue) is exactly the pressure under which that rule matters most.
- ArgoCD applies `components/engagement/` recursively — there is no file list to register the new ConfigMap in, same as every other bundle.
- The sidecar's resources match the fleet (`25m/32Mi` request, `100m/64Mi` limit).
- The image is still the bootstrap placeholder `sandbox-7a66c233d`, which Kyverno rejects until auto-deploy pushes a real tag. That is #4054's pre-existing state and untouched here; it does mean the 503s above are the *admitted-pod* behaviour rather than something currently observable in the sandbox.

Refs #1797, #4054, ADR-0034 D5, ADR-0220 D1.


JiRaska:
**Your parity fix is correct — I verified it independently. The PR is blocked by one trailing blank
line in the generated bundle, and the generator will put it back.**

## The fix works

Ran the gate against your branch:

```
check-authz-pdp-parity: 40 workload(s) checked; 4 skipped; 0 enforce-without-PDP violation(s).
```

So the sidecar wiring clears `#1797` and this PR is the one that unblocks the queue.

## What is actually red

`gates (lint)` → `yamllint`, and it is **one error in one of this PR's own new files**:

```
FAIL [lint] yamllint
  openbank-infra/gitops/components/engagement/engagement-service-opa-bundle.yaml
  ##[error] 2035:1 [empty-lines] too many blank lines (1 > 0)
```

Everything else in that gate's output is 28 pre-existing `comments-indentation` **warnings** across
`gitops/apps/*.yaml` — noise, not failures. Confirmed by running `yamllint` on current `origin/main`:
**PASS**. So this is a clean, single-cause regression introduced by this branch.

## The line comes from the generator, so fixing the file alone will not hold

`gen-engagement-opa-bundle.sh` ends with:

```sh
  echo "  manifest.json: |"
  …
  printf '\n'
```

That trailing `printf '\n'` after the last block is what produces line 2035. Editing the committed
YAML would go green now and come back the next time anyone regenerates — and this repo's convention is
that a generated artifact is never hand-edited.

**Suggested fix: drop the final `printf '\n'` in the generator, re-run it, commit both.** One
character class, and it stays fixed.

## Why this matters beyond your PR

`Validate manifests` is a required check and `main` is red on the parity gate, so **every open PR
inherits the failure** — mine included (#4033, #4060, #4067). This PR is the unblocker for all of
them, which makes a one-line lint error unusually expensive right now.

I have deliberately **not** pushed to your branch. Related and non-conflicting: #4067 repins
`engagement-service` to `sandbox-fac75d19`, because the tag this branch still carries
(`sandbox-7a66c233d`) does not exist in ECR — `git merge-tree` reports **0 conflicts** between the two,
so they can land in either order.

JiRaska:
Cross-reference — #4067 touches this same file concurrently, and the two are complementary rather than competing.

- **#4067** repins the image: `sandbox-7a66c233d` (a tag nothing ever pushed, which Kyverno's cosign/SBOM admission rejects) to `sandbox-fac75d19`. That makes the pod **admittable**.
- **#4068** (this one) adds the OPA sidecar. That makes it **serve** rather than answering 503 on both `@Authorize` endpoints.

Measured, not assumed: `git merge-tree` over the two branches from their merge-base reports **0 conflicts** — #4067 changes exactly one line (the image tag) and this PR adds the sidecar container, its volumes and the pod-roll annotation further down. Either merge order works.

Worth stating because merging only one leaves a state that looks fixed and is not: with #4067 alone the pod is admitted and both surfaces 503; with #4068 alone the policy is correct in a pod Kyverno will not admit.

PR #4072, opened 2026-08-07
security(libs): ADR-0219 D4 compile-time wiring assertion for the contact gate

JiRaska:
## Summary

Issue #2749 residual scope (ADR-0219 D4): `ContactPolicyGate` existed and was unit-tested, but nothing stopped a new marketing send path from simply not calling it — "every sender checks" was a convention, not a control. D4 rules out the cheap version explicitly: *"a **compile-time wiring assertion** ... a detekt/AST rule (**not a grep**)"*.

- New `openbank-libs-detekt-rules` module: a custom detekt `RuleSetProvider` (`MarketingCallSiteWiringRule`), loaded onto **every** module's forked detekt CLI classpath via ServiceLoader (`openbank.static-analysis.gradle.kts`).
- Source-retained `@MarketingCallSite` marker annotation (`openbank-libs-domain`) for functions that are a marketing-class touch reaching the delivery/surface layer.
- The rule fails the build when an annotated function's containing class has no `ContactPolicyGate` injected, or no `<gate>.check(...)` call exists anywhere in that class.
- Applied to the two real call sites that already wire the gate correctly: engagement-service's `ResolveSurfaceUseCase.resolve` and campaign-service's `CampaignJourneyActivitiesImpl.deliverStepGated`.

**Falsified before trusting it** (this repo's own rule — an unfalsified gate is unproven): routed the real gate call through a `.let{}` chain to break the receiver-name match, confirmed the rule went red, then reverted before commit.

Deliberately class-scoped, not a full interprocedural call graph, and type-matched by simple name (no `RequiresTypeResolution`, consistent with this repo's forked-CLI detekt setup) — both documented as limitations in the rule's KDoc.

**Not done here:** `notification-service`, D4's named choke point, still has no `ContactPolicyGate` wiring at all — not annotated yet since there's nothing to point the annotation at. That's the next real gap, once the gate itself is wired into notification-service's dispatch.

## Test plan
- [x] `MarketingCallSiteWiringRuleTest` (5 cases: no gate injected, gate injected but never called, compliant same-function call, compliant sibling-function call, unannotated function ignored)
- [x] `:openbank-engagement-service:detekt` / `:openbank-campaign-service:detekt` clean against the real annotated call sites
- [x] Falsification: temporarily broke the real `ResolveSurfaceUseCase.resolve` gate call (`.let{}` indirection) — rule correctly reported `MarketingCallSiteWiringRule`; reverted, confirmed green again
- [x] `ktlintCheck` clean across all touched/new modules


JiRaska:
Reviewed. Genuinely well-built — a real detekt AST rule (D4's stated mechanism, not a grep), correctly wired, with a falsifiable unit test that has **both** negative cases. Better than most gates in this repo. **But it does not close the gap the PR body opens with**, and the security claim in the KDoc is stronger than the code makes.

### 1. The assertion is satisfied by omission

`MarketingCallSite` is `@Retention(SOURCE)`, `@Target(FUNCTION)`, and `MarketingCallSiteWiringRule.kt:58` is `if (!isMarketingCallSite) return`. Counted on this branch: the annotation appears at exactly **two** production call sites —

```
openbank-campaign-service/.../CampaignJourneyActivitiesImpl.kt
openbank-engagement-service/.../ResolveSurfaceUseCase.kt
```

So a new marketing send path that simply does not annotate itself is compliant, and deleting the annotation from either existing site turns a red build green. Nothing asserts the marker count holds or that those two sites still carry it. That is verbatim the defect the body names — *"nothing stopped a new marketing send path from simply not calling it"* — moved one level out: the coverage set is now derived from the marker being present, which is the artifact-present-⇒-checked shape this repo keeps re-encountering.

ADR-0219 D4 specifies the other half explicitly: **every service that declares a marketing-class capability** must inject the gate. Deriving the set from the capability declaration is what would catch the case the body honestly lists as not-done ("notification-service has no gate wiring, nothing to point the annotation at") — under this form that service reads as clean.

### 2. It enforces wiring, not enforcement

`ContactPolicyGate.check` **returns** a `ContactGateDecision`; it never throws. The rule asserts only that a `<gate>.check(...)` call expression exists somewhere in the class. Three shapes pass and still send:

- **Ignored return value** — `contactGate.check(...)` as a bare statement, then send.
- **`contactClass = SERVICE_EXEMPT`** — `check` short-circuits to `ALLOWED` at line 59, before touching any port. A marketing send passing `SERVICE_EXEMPT` is a total suppression bypass that satisfies the rule perfectly.
- **Class-scoped, not function-scoped** (deliberate, and tested). The consequence is not stated: one gate call anywhere in a class satisfies *every* annotated function in it. `CampaignJourneyActivitiesImpl` already has a compliant `check`, so a second, ungated send method added to that class can be annotated and will pass.

The KDoc says "a marketing-class touch must not bypass the gate". One sentence saying it asserts wiring rather than that the decision is honoured would stop a future reader taking a green detekt as "no unsuppressed send is possible".

### 3. Clean, and worth crediting

- `libs-domain` stays framework-free — the annotation is pure Kotlin, no jakarta/Quarkus/JBoss import.
- **Runtime blast radius on `openbank-libs` is zero**: no new bean, no `@Produces`, no field initializer, no `@ApplicationScoped`, and a SOURCE-retained annotation is erased at compile. The only fleet-wide effect is the detekt classpath.
- `config/detekt/detekt.yml` declares the `openbank-contact-policy` key fleet-wide, and detekt fails config validation on an unknown rule-set id — so a ServiceLoader/classpath failure fails **loudly** rather than silently skipping. That is a good self-protecting design and rarer than it should be.
- The `authz.enforce: false` added to engagement's `application.yaml` is nested under `"%test":`, not top level. Test-scoped only.

### 4. Two concrete smaller items

- **`this.` prefix is a false positive.** Line ~90 compares `receiverExpression.text` verbatim against the property names, so `this.contactGate.check(x)` yields receiver text `"this.contactGate"`, which is not in the set — compliant code goes red. The same file already does the right thing for type names a few lines down (`substringAfterLast('.')`).
- **Contradictory comments about self-application**: `openbank-libs-detekt-rules/build.gradle.kts` says the rule set is never applied to itself, while the convention plugin adds `detektCli(project(":openbank-libs-detekt-rules"))` unconditionally and `detekt.yml`'s new comment says it *is* on its own classpath deliberately. One of the two is wrong; the `detekt.yml` one appears to describe reality.

### Minimum I would want before merge

Derive the coverage set rather than trusting the marker — at minimum a ratchet that the known annotated call sites still carry it, ideally the D4 form where a class referencing the delivery/surface layer *without* the marker is what goes red. Plus the one KDoc sentence in §2.

Note: `all-green` is absent rather than red — `Services CI` is still running, which is expected for a libs change that rebuilds the fleet.

Issue #4065, opened 2026-08-07
gitops pins openbank-engagement-service:sandbox-7a66c233d, an image that was never built — fleet attestation gate red on every PR since 16:39


JiRaska:
**`Verify fleet attestations` is failing on every PR since 16:39 today, because gitops pins an
`openbank-engagement-service` image tag that does not exist in ECR.**

## Measured

The gate is **green on `main`** — five consecutive daily runs, most recently today 03:38Z. It went red
on PR branches only after #4048 merged:

```
fac75d19e  08-07 16:39  feat(engagement): REST, persistence, outbox and consent-gate wiring for ADR-0220 (#4048)
```

The gate's own output:

```
  OK  openbank-vop-service:sandbox-e532b32e
  - 265175468565.dkr.ecr.eu-north-1.amazonaws.com/openbank-engagement-service:sandbox-7a66c233d
FLEET ATTESTATION GATE: FAIL — 1 of 61 declared image(s) not deployable.
No image-provenance policy may graduate Audit -> Enforce while this gate fails
```

It is not "built but unattested" — **the image was never built**:

```
$ aws ecr describe-images --repository-name openbank-engagement-service --image-ids imageTag=sandbox-7a66c233d
ImageNotFoundException: The image with imageTag 'sandbox-7a66c233d' does not exist

$ aws ecr list-images --repository-name openbank-engagement-service
sandbox-6b3594b2   sandbox-fac75d19   (+ their .sig / .att artifacts)
```

Two tags exist, and **`sandbox-fac75d19` is the merge commit of #4048 itself** — attested, signed, and
sitting in ECR unused. The manifest pins `7a66c233d`, a real commit but one no build ever produced.

So this is a one-character-class slip, not a missing pipeline: the pin was written from the wrong sha.

## Consequence

- Every PR opened since 16:39 shows a red `Verify fleet attestations`, on a gate whose message says
  *"No image-provenance policy may graduate Audit -> Enforce while this gate fails"* — i.e. it blocks
  the supply-chain work tracked on #1915.
- `kubectl get pod -A | grep engagement` returns nothing, so the workload is not running either.
- The failure is **addressed to nobody**: it is not a required check, so it reads as background red on
  other people's PRs. I only found it because it appeared on mine (#4060, a one-line CNPG volume
  change that touches no image).

## Fix

Repin to the tag that exists and is attested:

```diff
- image: …/openbank-engagement-service:sandbox-7a66c233d
[...]

Issue #4007, opened 2026-08-07
Four services ship a transactional outbox nothing ever writes to, and publish through a direct non-transactional emitter instead


JiRaska:
Swept every CNPG outbox table in the cluster and cross-checked each against its service's code. The
apparatus is present fleet-wide — dispatcher, backlog gauge, `dispatch-enabled: true`, a
`persistInTransaction` on the repository — and for a group of services **nothing ever writes to it**,
while the events themselves go out through a second, direct, non-transactional publisher.

## Live table state

```
NON-EMPTY                                 EMPTY (0 rows)
ledger_outbox            577 SENT         account_outbox          balance_outbox
aml_outbox               108 SENT         party_outbox            kyc_outbox
domestic_payment_outbox  102 SENT         fx_outbox               interest_outbox
lending_outbox            44 SENT         notification_outbox     audit_outbox
document_outbox           41 SENT         delegation_outbox       clearing_outbox
sepa_payment_outbox       41 SENT         sct_inst_outbox
consent_outbox            10 SENT
dispute_outbox             2 SENT
card_outbox               24 DEAD  (#4005)
billing_outbox             2 DEAD
```

## Cross-checked against code — two different causes, which need different fixes

Callers of `persistInTransaction` outside the repository's own implementation:

```
account       6      balance       1      party         0
kyc           1      fx            2      interest      0
notification  2      audit         0      delegation    3
clearing      2      sepa-instant  0
```

**Group A — no write side at all** (`party`, `interest`, `audit`, `sepa-instant`): zero
`OutboxMessage(` constructions anywhere in `src/main`. The dispatcher polls a table that is empty by
construction and the backlog gauge reports 0, which is structurally correct and carries no information.
This is exactly the shape found in `openbank-security-scanner` (#3973) — the scanner and its outbox
being two halves that never meet — so that is not a one-off, it is a pattern.

**Group B — write side exists, table still empty** (`balance`, `kyc`, `fx`, `notification`,
`delegation`, `clearing`): the code path is real but has never fired in this environment. That is a
much weaker finding and may simply be traffic; I am not claiming a defect for these, only that the
outbox has never been exercised, so its atomicity guarantee is untested in practice.

## What actually publishes, for the Group A services

`party-service` is the clearest case. Its outbox has no writer, yet the audit trail holds **61
`PARTY_CREATED`, 2 `PARTY_UPDATED` and 1 `PARTY_MERGED`** rows — so the events demonstrably flow. They
[...]

Issue #4005, opened 2026-08-07
card-issuance has published zero events: all 24 card_outbox rows are DEAD on an open circuit breaker, stamped 1970, with no self-healing path


JiRaska:
**Every row in `card_outbox` is `DEAD`. There are no `SENT` and no `PENDING` rows, so
`openbank-card-issuance-service` has published zero events for its entire life.**

Live, from `payments/card-issuance-db`:

```sql
SELECT status, count(*), min(created_at), max(created_at) FROM card_outbox GROUP BY 1;
DEAD | 24 | 1970-01-01 00:00:00+00 | 1970-01-01 00:00:00+00
```

Breakdown, all at `attempt_count = 10` (the `OutboxFailurePolicy.DEFAULT_MAX_ATTEMPTS` ceiling):

```
card.limits_changed.v1     9
card.controls_changed.v1   6
card.issued.v1             6
card.status_changed.v1     3
```

and every one carries the same `last_error`:

```
com.openbank.cardissuance.infrastructure.outbox.CardOutboxDispatcher#publishWithResilience
  circuit breaker is open
```

## What this means

`DEAD` is terminal by design — ADR-0050 N5 parks a poison row so it "can neither be retried forever
nor starve the batch", and `listProcessable` excludes it. So these 24 will never be retried by anything.
No card issuance, limit change, control change or status change has ever reached Kafka, and nothing
downstream that subscribes to `card.*` has ever seen a message.

`openbank.outbox.dispatch-enabled: true` is set (`application.yaml:213`), so this is not the usual
"the switch was never flipped" case — the dispatcher ran, tried 10 times per row, and the breaker was
open for all of them.

**A latched-open breaker plus a terminal `DEAD` status is a combination with no self-healing path**: the
breaker's half-open probe never fires because there is no work left to probe with, and the work is
parked because the breaker was open. Whatever originally opened it may well be long fixed; nothing
would tell you. Same shape as the domestic-payment stranding.

## The 1970 timestamps

All 24 rows have `created_at = 1970-01-01`, and that is not cosmetic here — it is the exact failure
`OutboxMessage`'s KDoc documents after #3272/#3338:

> The dispatcher claims work with `ORDER BY created_at ASC … FOR UPDATE SKIP LOCKED`, so an epoch row
> sorts ahead of every correctly-stamped one permanently and starves real traffic behind a 1970
> backlog, while the `openbank.outbox.backlog` age signal reads a fresh row as 56 years old.

[...]

Issue #3975, opened 2026-08-07
3 CNPG clusters have no recovery point: litellm-db is CrashLooping on a full volume, copilot-db has never archived a WAL, and ContinuousArchiving=True is green for a cluster with no backup at all


JiRaska:
Measured against the live cluster on 2026-08-07 ~06:35Z while triaging #3555. Three separate faults,
one of which makes the other two hard to see.

```
CNPG clusters: 57     with a firstRecoverabilityPoint: 54     WITHOUT: 3
  ai-platform/litellm-db
  observability/glitchtip-pg
  platform/copilot-db
```

## 1. `litellm-db` is down, and has been for ~17 hours

```
pod/litellm-db-1        0/1   CrashLoopBackOff   209 restarts (17h)
cluster phase:          "Not enough disk space"
pvc litellm-db-1:       2Gi   gp3   Bound
Ready = False
```

The `litellm` application pod next to it is `1/1 Running` and 4d13h old, so from the outside the
component looks alive. #3555 gave this cluster a `backup:` block and that part landed correctly — but
a cluster that cannot start cannot archive, so the backup it gained is not producing a recovery point
either. The immediate cause is a full 2 Gi volume.

## 2. `copilot-db` has never archived a single WAL segment

```
phase:  Cluster in healthy state      ready 1/1      Ready = True
ContinuousArchiving = False | unexpected failure invoking barman-cloud-wal-archive: exit status 4
LastBackupSucceeded = False | instance manager was restarted during backup on pod copilot-db-1
firstRecoverabilityPoint: None
```

From the pod log:

```
barman-cloud-check-wal-archive ... ERROR: Barman cloud WAL archive check exception: Unable to locate credentials
options: ["--cloud-provider","aws-s3","s3://openbank-sandbox-db-backups/copilot-db","copilot-db"]
archive command failed ... The failed archive command was: /controller/manager wal-archive ... pg_wal/000000010000000000000001
```

**The segment it cannot archive is `000000010000000000000001` — the first WAL the cluster ever
produced.** So this is not drift or a recent regression; the archive has never worked once. The
cluster is otherwise `Ready`, which is why nothing surfaced it.

`inheritFromIAMRole: true` is set and the pod carries no `AWS_*` environment at all — but neither does
`accounts/accounts-db-1`, which archives successfully, so the difference is *not* simply missing env
injection and I have not established the actual cause. Worth checking before assuming: the
[...]

PR #4079, opened 2026-08-07
feat(ci): runtime conformance — compare what the repo claims to what the cluster does

JiRaska:
## The gap this fills

`gates.yaml` has 117 gates and they are good. Every one of them reads **committed text**, because
that is what CI can see. **Nothing reads running state.**

On 2026-08-07 a two-hour sweep of the live databases, pod logs and AWS APIs found five defects no
gate could have caught. Each had the same shape — a control that is declared, wired, reports success,
and carries nothing:

| finding | what every text-level check saw |
|---|---|
| four services ship an outbox nothing writes to (#4007) | `dispatch-enabled: true`, dispatcher present, gauge 0 |
| card-issuance: 24 outbox rows, **all DEAD** (#4005) | same gauge, also 0 — `DEAD` is excluded from `listProcessable` |
| three CNPG clusters with no recovery point (#3975) | `Ready: True`, `ContinuousArchiving: True`, `backup:` block present |
| 21 of 21 canary Rollouts at one replica (#3806) | canary block declared, `status: Healthy` throughout |
| gitops pinned an image never built (#4065) | valid YAML, valid sha, valid registry path |

None of these is a missing gate. They are the same gate *class* — text — applied to a question that is
not about text.

## Design

Split the way `check-external-feeds.py` already splits its two halves:

```
--check <snapshot>   pure function of (snapshot, repo). No network. Runs in CI today as the
                     enforced gate `runtime-conformance-comparators`.
--collect            needs cluster + registry reads. Dispatch-only.
--self-test          22 cases driving every comparator branch, including the near-misses.
```

**Collection is dispatch-only on purpose.** No workflow in this repo has cluster credentials —
`dr-restore-verify.yml` is the only one that tries and has never run. Putting cluster read credentials
into Actions is your decision, not something a workflow should quietly assume. Promote
`runtime-conformance.yml` to a `schedule:` once that is settled.

## Four probes, each asserting effect rather than declaration

| probe | claim | fact |
|---|---|---|
| `backup_recoverability` | a `backup.barmanObjectStore` block | `firstRecoverabilityPoint` is non-null |
| `outbox_liveness` | an `*OutboxDispatcher` exists | something constructs `OutboxMessage`, and not every row is `DEAD` |
| `canary_realisability` | `setWeight: N` | replicas make `N%` expressible at all |
| `image_tag_exists` | `:sandbox-<sha>` in gitops | that tag exists in the registry |

`Ready` and `ContinuousArchiving` are deliberately **not** consulted for backups: on 2026-08-07 all
three no-recovery-point clusters reported at least one of them `True`, and `ContinuousArchiving` is
trivially `True` on a cluster with no archive destination at all.

## Falsification — including two defects it found in itself

The self-test fails when any comparator is broken (verified: breaking the backup comparator to accept
a healthy phase → `rc=1`; restoring → `rc=0`).

**More importantly, the first run against the live estate exposed two defects in this tool**, which is
exactly why it was run before being trusted:

1. `openbank-libs-runtime` ships the *abstract* dispatcher every service extends — it owns no table
   by design, so it was a permanent false positive.
2. Outbox facts are keyed by **table** name (`card_outbox` → `openbank-card`) while claims are keyed
   by **directory** name (`openbank-card-issuance-service`). They never matched, so **nine real
   findings were silently skipped and the run looked clean.**

Both are now self-test cases, and the resolver accepts a prefix match only when it is **unique** —
an ambiguous one returns nothing rather than attributing one service's dead rows to another.

Absent facts print `NOT CHECKED (this is not a pass)` per probe, and the workflow emits an empty
snapshot rather than skipping when no cluster is reachable — a skipped job and a clean run look
identical in the checks list.

## Current estate

```
31 divergence(s)
   1  backup_recoverability   copilot-db (litellm-db recovered at 16:49 and is correctly NOT flagged)
   9  outbox_liveness         incl. card-issuance all-DEAD, and 4 services nobody had checked
  21  canary_realisability
```

It independently rediscovered every manual finding above, plus `billing`, `pid`, `psd2` and
`tpp-registry` — which I had not looked at.


JiRaska:
**Added the fifth probe — audit attribution — and closed a gap in the collector that the first real
run had already exposed once.**

## The probe

Every audit row must say **what it is** and **where it came from**. Against the live estate now:

```
124 of 1696 rows   event_type    = 'UNKNOWN'
1298 of 1696 (77%) source_service = 'unknown'
1 producer populates it, across 21 subscribed topics
```

Both surface as separate findings, because they need different fixes: the type is recoverable from
the `ce-type` Kafka header for outbox-published topics (#3994), while **no header carries the source
at all** — that is an envelope decision tied to #4007.

## Zero thresholds, and the self-test enforces it

No percentage anywhere. A guessed threshold can sit *under* the regression it exists to catch, so the
trigger is `n >= 1` and the count is context. Two self-test cases pin a single bad row being flagged,
and I verified the constraint is real: sneaking `if unknown_type > 100:` into the comparator makes the
self-test **fail** (`rc=1`, 2 failures); restoring it returns 0.

`total = 0` is deliberately neither a pass nor a finding — an empty table has nothing to attribute
yet, and treating that as conformant is how a fresh environment gets certified.

## The collector gap, closed

The first version left outbox facts *"to the caller"*, which in practice meant a hand-written
enrichment script — and that script keyed facts by **table** name while claims are keyed by
**directory** name, so nine real findings were silently skipped while the run looked clean. That was
caught only by running against the real estate.

The collector now performs both database probes itself and keys outbox facts by service, so there is
one code path and it is the one under test. Registry facts stay caller-supplied: ECR is a separate
credential domain, and an absent fact prints `NOT CHECKED (this is not a pass)`.

## Estate total

```
33 divergence(s)
   2  audit_attribution
   1  backup_recoverability   copilot-db  (glitchtip-pg declares no backup; litellm-db recovered)
   9  outbox_liveness
  21  canary_realisability
```

Was 31. `runtime-conformance-comparators` still passes as an enforced gate, and
`check-workflow-run-step-size` is unaffected.

JiRaska:
Reviewed. Registration and falsifiability are solid; **one real, reproducible silent-pass** worth fixing before merge.

**Verified rather than read:** `run-gates.py --list` shows `gitops  runtime-conformance-comparators  enforced always  selftest`, declared in `gates.yaml` — not an inline step in a conditional job. `selftest_expect: pass` is right, since the command *is* the checker's own harness. There is no `--enforce` flag, and that is fine here: the script exits 1 unconditionally on self-test failure and on `--check` divergence, so `mode: enforced` is not vacuous. 28 self-test cases pass. Falsified with a counter-example the body does not quote — mutating `elif total > 0 and dead == total:` to `total > 999` in `cmp_outbox_liveness` produced `self-test FAIL`, rc=1; restoring gave rc=0. The claim readers work against the real tree: 57 CNPG clusters, 32 outbox services, 21 canary rollouts, 21 subscribed audit topics, 58 image pins.

### The defect: the "NOT CHECKED" branch cannot fire

`check()` prints `NOT CHECKED (this is not a pass)` only when a probe key is **absent** from `facts`. But `collect()` assigns every key, and `_sh()` cannot signal failure:

```python
def _sh(args): 
    return subprocess.run(args, capture_output=True, text=True, check=False).stdout
```

`check=False` plus returning `.stdout` means a `kubectl` RBAC denial comes back as `""` — indistinguishable from "queried, found nothing". So a partial credential failure yields a probe that is *present and empty*, the comparator finds nothing, and the run is clean.

The realistic split is `get` granted, `pods/exec` denied, so every `_psql` returns `""`:

```
facts = {backup: <57 healthy>, canary: <21 @10 replicas>, outbox: {}, audit: {}, image: {}}
runtime-conformance: 0 divergence(s) between what the repo claims and what the cluster does
rc=0
```

Zero warnings, exit 0 — while the 9 `outbox_liveness` and 2 `audit_attribution` findings the body says exist in the estate today are invisible. The body's "absent facts print NOT CHECKED per probe" is true only for the workflow's `{"facts":{}}` fallback, which needs kubectl to be missing entirely.

`image_tag_exists` is the worst case: `snap["facts"].setdefault("image_tag_exists", {})` means the registry probe — the one that catches the #4065 class — reports `NOT CHECKED` **never**, on any automated run, since ECR is caller-supplied by design.

This is the identity problem the repo already knows in another form: a permission-shaped absence is indistinguishable from a real one, and it always fails toward a confident wrong answer. Fix is to make `_sh` return `None` on failure and omit the key (or store a `{"_not_collected": true}` sentinel), plus a self-test case — the current 28 do not model it.

### Two smaller notes

- The gate's `run:` is byte-identical to its `selftest:`, so `runtime-conformance-comparators` is a unit-test runner rather than a gate over repo state. Honestly documented in the gate comment, but worth stating plainly: no PR can go red from a `--check` result.
- `if [ -n "$SNAPSHOT_IN" ] && [ -f "$SNAPSHOT_IN" ]` falls through to collect-or-empty on a typo'd dispatch input. An `else`-fail on "path given but not found" is cheap.

PR #4071, opened 2026-08-07
feat(lending): carry partyId on loan.stage_changed so arrears can be consumed

JiRaska:
## Summary

`loan.stage_changed` is the only event in the platform that says a customer has fallen behind — it carries `newStage` and `daysPastDue` — and no service outside lending could act on it, because the payload had no `partyId`. This adds one, additively.

## Type of change

- [x] feat (new functionality)
- [ ] fix (bug fix)
- [ ] refactor (no behavior change)
- [ ] perf (performance)
- [ ] docs
- [ ] chore (build / tooling)
- [ ] security (private until disclosed)
- [ ] breaking change

## Linked issues

Closes #4070
Refs ADR-0220 D1 (the vulnerable-customer exclusion this unblocks)

## Why

ADR-0220 D1 requires in-app promotional surfaces to exclude vulnerable customers — fraud hold, **arrears**, open dispute, erasure requested — and states that such an exclusion lagging by a snapshot interval "is a compliance defect, not a freshness metric".

`openbank-engagement-service` already has the rule and the `AdverseState.ARREARS` type, but `ResolveSurfaceUseCase` constructs every party with an empty adverse-state set; its own comment calls that a launch blocker. Arrears could not be fed because a consumer receiving `loan.stage_changed` learns that *a loan* moved to STAGE_2 with 40 days past due, and has no way to know whose — it would have to call back into lending-service on the app-open hot path, the live query the ADR's eligibility snapshot exists to avoid.

Every other lending event crossing a service boundary already carries `partyId` (`loan.disbursed`, `loan.written_off`, the credit transitions). This one was the exception.

## Changes

- `LendingService`: `partyId` added to the `loan.stage_changed` payload, from `loan.partyId` already in scope — no new lookup, no new dependency.
- `LendingServiceTest`: the existing stage-transition test now asserts the field.

## Compatibility

**Additive.** No field renamed, none removed, no type changed. AnaCredit's overdue/stage projection (#638) reads the fields it already read. The event has no Avro schema — it is a JSON outbox payload — so there is no registry compatibility step; the shape change is additive in the same sense.

## Definition of Done

- [x] Hexagonal layering preserved.
- [x] Unit tests added/updated — the assertion was **falsified**: removing `partyId` from the payload reddens exactly that test.
- [ ] Integration tests added/updated. — N/A, no boundary behaviour changed.
- [ ] Contract tests updated. — N/A, no consumer pact covers this event.
[...]

Issue #4070, opened 2026-08-07
loan.stage_changed carries no partyId, so the arrears adverse-state feed for ADR-0220 cannot be built


JiRaska:
## What

`loan.stage_changed` is the platform's only event that says a customer has fallen behind — it carries `newStage` and `daysPastDue` — and it cannot be consumed outside lending-service, because its payload has no `partyId`:

```json
{"loanId":"…","previousStage":"STAGE_1","newStage":"STAGE_2","daysPastDue":40,"period":"2026-07","asOf":"…"}
```

Every other lending event that crosses a service boundary carries one (`loan.disbursed`, `loan.written_off`, the credit transitions). This one is the exception, and it is the one an arrears feed needs.

## Why it matters now

ADR-0220 D1 requires in-app promotional surfaces to exclude vulnerable customers — the ADR-0200 D6 adverse-state set: fraud hold, **arrears**, open dispute, erasure requested — and says the exclusion must react at event speed, because "a vulnerable-customer exclusion that lags by a snapshot interval is a compliance defect, not a freshness metric".

`openbank-engagement-service` has the rule (`EligibilityRule.isEligibleForPromotionalTargeting`) and the type (`AdverseState.ARREARS`), but nothing to feed them: [`ResolveSurfaceUseCase.kt:60`](../blob/main/openbank-engagement-service/src/main/kotlin/com/openbank/engagement/application/usecase/ResolveSurfaceUseCase.kt#L60) constructs every party with an empty adverse-state set, and its own comment calls that a launch blocker. A consumer holding only a `loanId` would have to call back into lending-service on the app-open hot path to learn whose loan it was — exactly the live query ADR-0220 D1 designs the eligibility snapshot to avoid.

## Survey of the other three adverse states

Measured on `main`, so the size of the remaining gap is on the record:

| state | source | usable today |
| --- | --- | --- |
| `ERASURE_REQUESTED` | `openbank.party.events` / `PARTY_ERASED`, carries partyId | yes |
| `ARREARS` | `loan.stage_changed` — has the data, **no partyId** | this issue |
| `DISPUTE_OPENED` | `openbank.dispute.events` publishes only `dispute.resolved` and `dispute.remediation_requested`; **no open event reaches Kafka** (`"OPENED"` is audit-trail only) | no |
| `FRAUD_HOLD` | fraud-service has **no outgoing channel at all** | no |

So three of the four need a producer-side change before the exclusion can be honest. This issue is the smallest of them.

[...]

PR #4087, opened 2026-08-07
feat(dispute): publish dispute.opened so an open dispute is visible outside this service

JiRaska:
## Summary

Opening a dispute was invisible outside dispute-service. The only dispute facts on Kafka were `dispute.resolved` and `dispute.remediation_requested` — so a consumer could learn that a dispute *ended*, never that one had *started*. `"OPENED"` existed, but as a `DisputeTimelineEvent`: an audit row inside this service.

## Type of change

- [ ] fix (bug fix)
- [ ] perf (performance)
- [ ] docs
- [ ] chore (build / tooling)
- [ ] breaking change

## Linked issues

Refs #4070 — the survey of what ADR-0220 D1's vulnerable-customer exclusion is missing. This is one of the three producer-side gaps it lists.

## Why

ADR-0220 D1 requires in-app promotional surfaces to exclude the ADR-0200 D6 adverse-state set: fraud hold, arrears, **open dispute**, erasure requested. `openbank-engagement-service` has the rule and the `AdverseState.DISPUTE_OPENED` type, but nothing can feed the dispute half while the fact never leaves this service.

## Changes

- **`dispute.opened`** — new outbox event on open, carrying `partyId`, `disputeType`, `status` and `openedAt`.
- **`partyId` added to `dispute.resolved`** — see below; this is the half that is easy to forget.
- **`save(dispute, outbox)`** on `DisputeRepository`, mirroring the existing `update(dispute, outbox)`, so the dispute and its event commit in one transaction.

## The two decisions worth reviewing

**`dispute.resolved` needed `partyId` too, and omitting it would have been worse than doing nothing.** A consumer can apply an exclusion when it sees an open, but with no party on the resolve it can never lift one — leaving a customer permanently excluded from surfaces by a dispute that closed months ago. That failure is silent and would look like "this customer never engages". The two events now bracket the window. `dispute.remediation_requested` already carried a `partyId`, so this only makes the topic consistent.

**Transactional outbox, not a publish after save.** A dispute committed without its event is invisible downstream forever, with nothing to reconcile against — there is no later write to piggyback on, unlike the resolve path. The new `save` overload keeps them atomic and mirrors the shape the resolve path already uses, rather than inventing a second pattern.

## Definition of Done

- [x] Unit tests added/updated — the assertion was **falsified**: replacing the outbox list with `emptyList()` reddens exactly that test.
- [ ] Integration tests added/updated. — N/A, no new boundary behaviour; the outbox dispatcher is unchanged.
- [ ] Contract tests updated. — N/A, no consumer pact covers this topic.
- [x] `./gradlew ktlintCheck detekt test` passes — 50 tests, 0 failures.
- [x] No new lint warnings.
- [ ] OpenAPI spec regenerated. — N/A, no REST change.
- [ ] Flyway migration. — N/A, no schema change (the outbox table already exists).
- [x] Event schema versioned backward-compatibly — `dispute.opened` is a new type on an existing topic; the `partyId` addition to `dispute.resolved` is additive. JSON outbox payloads, no registry step.
- [x] Docs updated — reasoning is in the payload builders and the port KDoc.

## Security checklist

- [ ] Auth / crypto / payment code changed? — no.
- [x] **PII path changed?** — yes, narrowly. A party identifier now travels on `openbank.dispute.events`, which already carries one on `dispute.remediation_requested`. Consumers are internal services under the same ADR-0118 retention. The identifier is a UUID, not an attribute, and the purpose is to *stop* contacting a customer in dispute.
- [ ] Cardholder data path changed? — no.

## Compliance impact

- [ ] PCI DSS
- [ ] DORA
- [x] GDPR
- [ ] PSD2 / SCA / RTS
- [ ] AML / 5AMLD
- [ ] CNB reporting

GDPR: the identifier is added to enforce a protection — suppressing marketing to customers in dispute — rather than to enable targeting.

## Rollout / Rollback

- Deployment strategy: rolling. Nothing consumes `dispute.opened` yet, so this is producer-only.
- Rollback plan: revert. Additive on both counts; no consumer breaks either way.

## Reviewer notes

**What this does NOT do.** It makes the dispute adverse-state feed possible; it does not build it. engagement-service still constructs an empty adverse-state set and that stays a launch blocker — better than a half-wired exclusion that reads as complete.

**Remaining gap after this lands.** Of the four adverse states: erasure has a usable source today (`PARTY_ERASED`), arrears gets one from #4071 (open, money-path, awaiting two approvals), this PR covers dispute — and **fraud hold still has none**: fraud-service has no outgoing channel at all, and giving it one means first deciding what a "fraud hold" event is, which is a design question rather than a plumbing one.

### DIFF
--- a/openbank-dispute-service/src/main/kotlin/com/openbank/dispute/application/port/out/DisputePorts.kt
+++ b/openbank-dispute-service/src/main/kotlin/com/openbank/dispute/application/port/out/DisputePorts.kt
@@ -17,6 +17,17 @@ interface DisputeRepository {
 
     fun save(dispute: Dispute): Uni<Dispute>
 
+    /**
+     * Persist the dispute and [outbox] in the SAME transaction (transactional outbox,
+     * ADR-0003/0050) — the open path's counterpart to [update].
+     *
+     * A dispute committed without its `dispute.opened` event is invisible to every consumer
+     * downstream, and the one that matters is ADR-0220 D1's vulnerable-customer exclusion: a
+     * customer with an open dispute must stop receiving promotional surfaces, and it can only know
+     * that if opening one is announced.
+     */
+    fun save(dispute: Dispute, outbox: List<OutboxMessage>): Uni<Dispute>
+
     fun findById(id: UUID): Uni<Dispute?>
 
     fun findByReference(reference: String): Uni<Dispute?>
--- a/openbank-dispute-service/src/main/kotlin/com/openbank/dispute/application/usecase/DisputeService.kt
+++ b/openbank-dispute-service/src/main/kotlin/com/openbank/dispute/application/usecase/DisputeService.kt
@@ -80,7 +80,12 @@ class DisputeService(
             createdAt = now,
             updatedAt = now,
         )
-        return disputeRepo.save(dispute).flatMap { saved ->
+        // The timeline row and the Kafka event are NOT the same thing, and the difference was a
+        // real gap: "OPENED" below is a DisputeTimelineEvent — an audit entry inside this service
+        // — so until now nothing outside dispute-service could learn that a customer had opened
+        // one. ADR-0220 D1 needs exactly that fact to stop sending promotional surfaces to a
+        // customer in dispute, and #4070 records that the absence made the exclusion unbuildable.
+        return disputeRepo.save(dispute, listOf(openedOutboxMessage(dispute))).flatMap { saved ->
             val event = DisputeTimelineEvent(
                 disputeId = saved.id,
                 eventType = "OPENED",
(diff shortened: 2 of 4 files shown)


JiRaska:
Reviewed. **No blocking finding.** The atomicity is structural rather than asserted, which is the right way round.

`DisputeRepositories.save(dispute, outbox)` wraps both writes in a single `sf.withTransaction { }`, chaining `s.persist(e)` and `s.persist(message.toOutboxEntity())` — so the dispute row and its `dispute.opened` outbox row commit together by construction. That matters because a mocked-repository unit test could not prove it, and this PR's test suite is unit-only; here it does not need to, because the shape makes the failure unrepresentable rather than merely untested.

`persist` (not `merge`) is correct on this path: `open` is an INSERT of a new aggregate. The documented trap is application-assigned `@Id` on an *update*, which this is not.

Two observations, neither blocking:

**1. Nothing proves the event's wire shape.** The payload is built by string interpolation and the test asserts substrings of it (`assertThat(payload).contains("\"eventType\":\"dispute.opened\"")`). That catches a typo in the literal, not a malformed document — an unescaped quote or newline in an interpolated field would produce invalid JSON and pass. dispute-service already has a real-DB IT (`DisputeOutboxClaimIT`), so the infrastructure to parse one emitted row and assert it is valid JSON with the expected keys already exists; that would close it cheaply. Same shape flagged on #3876 for its evidence payload, so it is a pattern rather than a one-off.

**2. The event contract does not cover this.** `dispute.opened` is a **new** event type, and it appears in neither `docs/asyncapi/openbank-events.yaml` nor a per-service `asyncapi.yaml` (dispute-service has none — only delegation, engagement and notification do). `check-event-contract-coverage.py` passes: it reports *39 producer topics — 3 with a contract, 36 grandfathered, no new undocumented topic*, and the dispute topic is already in `.github/event-contract-baseline.txt`. So the gate is topic-granular and a new event **type** on an already-grandfathered topic is invisible to it. That is not a failure of this PR — it is the gate's declared scope — but it does mean "publish it so it is visible outside this service" currently ships with no machine-readable description of what consumers should expect.

The KDoc distinguishing the `DisputeTimelineEvent` audit row from the Kafka event is a good addition; that conflation is exactly the gap the PR describes closing.

Issue #3615, opened 2026-08-03
ADR-0232: the delegated money path does not exist — every amount-aware delegation guard is unreachable


JiRaska:
ADR-0232 D3 says the delegation grant is enforced on every request, "the last line before the money path". Measured against `origin/main` on 2026-08-03, there is no such path: every amount-aware delegation guard in the fleet is unreachable from production code.

## What was measured

1. `AuthorizationService.isAuthorizedForAmount` (`openbank-account-service/src/main/kotlin/com/openbank/account/application/usecase/AuthorizationService.kt:75`) is the only enforcer of `perTransactionLimit` in account-service, and the only amount-aware account guard anywhere. Its only callers are in `AuthorizationServiceDelegationTest`. The HTTP endpoint `GET /api/v1/accounts/{id}/authorizations/check` calls the amount-free `isAuthorized` instead.
2. The three delegation `/check` guards — `AuthorizationResource` `/check`, `SavingsGoalDelegationResource` `/check`, `CardDelegationResource` `/{id}/delegation/check` — have no in-repo caller, and none of them takes an amount.
3. delegation-service's amount-aware `POST /api/v1/delegations/check` (`DelegationResource.kt:257`) is called only by the read-only admin-ui probe (`openbank-admin-ui/src/app/api/delegations/check/route.ts`).
4. No money-moving service — domestic-payment, sepa-payment, sepa-instant, swift, standing-order, settlement, clearing, transaction, ledger — reads a delegation grant or projection at all. customer-edge payment initiation requires debtor-owner == JWT party, so a delegate gets 403 before any guard is consulted.
5. `SavingsWithdrawalApproved` (`SavingsProposalService.kt`) is documented as "the executable instruction the payments path consumes". No consumer exists in any service or in `openbank-contracts`.

So a delegate cannot move money at all today, and the guards that would constrain them if they could are correct code with no call site. This is the same defect class as #3613 (`dailyLimit`/`monthlyLimit` accepted and enforced nowhere) one layer down: there the number had no counter, here the check has no caller.

## What is needed

Wiring, in this order — a limit check is only worth having once there is a debit to attach it to:

- A payment-initiation path that accepts a delegate as the initiator: customer-edge must stop requiring debtor-owner == JWT party for a party holding an ACTIVE grant with the payment capability.
- domestic-payment (first, as the narrowest money path) consults the amount-aware guard with the actual instructed amount before accepting the instruction, and fails closed when the delegation projection is unavailable.
[...]

PR #4101, opened 2026-08-07
fix(delegation): an unpriced check against a priced grant is a denial, not coverage

JiRaska:
## The change

`DelegationGrant.covers` — three characters of logic, on the authorization path:

```diff
 fun covers(capability: DelegationCapability, amount: Money?): Boolean {
     if (!hasCapability(capability)) return false
-    if (amount == null) return true
+    if (amount == null) return perTransactionLimit == null
     return withinLimits(amount)
 }
```

## Why

A null `amount` means *the caller did not say what the action is worth*. That is fine for a capability
with no ceiling — a read has no sum. It is not fine when the grant carries a `perTransactionLimit`:
returning `true` there tells the caller the ceiling is satisfied **without comparing anything to it**,
and the ceiling is the entire reason the grantor set it.

`customer-edge`'s `hasGrant` omits the amount on **every** call. So under the old rule, the first
caller to put a payment path behind a grant would have received `true` for any sum, with the limit
sitting unread. #3615 proposes exactly that swap — relaxing the debtor-ownership check onto
`ownsAccount(...) || hasGrant(...)`.

**Failing closed here rather than fixing the caller is the point.** It makes the ceiling real
regardless of which caller forgets, instead of depending on every caller remembering. The caller-side
change is still needed and stays on #3800.

## What is unaffected

A grant with `perTransactionLimit == null` still answers on capability alone, so read paths and
unpriced grants behave exactly as before. That is the near-miss this could have broken — a rule that
denied *every* null amount would have taken those out too — and it has its own test.

## Verification

Falsified rather than merely run. Removing the fix and re-running:

```
WITHOUT the fix:  tests=14  failures=1
  FAILED: covers denies an unpriced question against a grant that carries a ceiling()
WITH the fix:     tests=14  failures=0
```

Counts read out of the JUnit XML, not from Gradle's exit code, and `build/test-results` was deleted
between runs so the green cannot be the previous run's.

```
full service suite    81 tests, 0 failures, 0 errors, 1 skipped (14 suites)
detekt + ktlintCheck  clean
```

## Scope — what this deliberately does not touch

- `dailyLimit` / `monthlyLimit` remain unenforced. They are **refused at write time** by
  `rejectUnenforcedCeilings`, which is the honest state until a spend counter exists; this PR is only
  about the ceiling that survives.
- `customer-edge` still does not send an amount. After this it fails closed instead of open, which is
  a behaviour change for any future priced grant — worth knowing before merge, and the reason it is
  better than the alternative is above.

## Correction I owe this PR

I claimed earlier today, on #3800 and #3615, that `/api/v1/delegations/check` had no amount field at
all. It does — `CheckDelegationRequest.amount: MoneyDto?` — and `check()` passes it through. I had
inferred the contract from the JSON one client happens to build. Both issues now carry the correction;
the ordering argument survived it, the mechanism did not.

Refs #3800, #3615


JiRaska:
Three-character diff, high leverage, worth reading closely rather than trusting the size.

The fix is correct and the near-miss test is the right one to have: `covers still answers on capability alone when the grant carries no ceiling` proves the fix doesn't over-correct into denying every null-amount call — only ones against a grant that actually carries a `perTransactionLimit`. Without that test, a plausible-looking stricter rule (deny ANY null amount) would have silently broken every read path and unpriced grant.

The KDoc's framing is the important part to preserve: this isn't "add a null check", it's "an unpriced question against a priced grant is a denial, not coverage" — the ceiling only means something if answering it is mandatory regardless of which caller forgets to pass an amount. `customer-edge`'s `hasGrant` omitting amount on every call (#3615) is exactly the caller this closes the door on.

Not merging — delegation-service authorization path, human review by rule.

JiRaska:
Reviewed. **The fix is right and at the right layer — but the "read paths are unaffected" claim is not true, and the counter-example is an ordinary grant on existing data.**

**What's correct.** `covers(cap, null)` returning `Allowed` against a grant carrying `perTransactionLimit` meant the ceiling was never compared, and `CustomerEdgeResource.hasGrant` builds its body with **no** `amount` on every call — I verified that; the exposure was one caller away. Failing closed in the domain is the right layer, an absent price now reads as denied where a ceiling exists, and that is the safe direction for money. Concurrency does not apply: `covers`/`withinLimits` are pure functions over an immutable aggregate and `check()` is a read, so there is no read-then-write to make atomic. The test layer is appropriate — the property is pure domain logic, so a domain unit test can prove it, and the falsification method matches house rules.

**The finding.** The KDoc says:

> a grant with `perTransactionLimit == null` still answers on capability alone, so **read paths and unpriced grants behave exactly as before**

That holds only for grants with no ceiling. Nothing couples a ceiling to payment capabilities — verified on this branch:

- `rejectUnenforcedCeilings` rejects only `dailyLimit` and `monthlyLimit`; `perTransactionLimit` is deliberately permitted, since it is the one that *is* enforced.
- `offer()` passes `perTransactionLimit = command.perTransactionLimit` straight through for **any** capability set — no coupling to capability type anywhere.

So this grant — the shape a parent sharing an account with a child would create:

```
capabilities: [ACCOUNT_READ_BALANCES, ACCOUNT_INITIATE_PAYMENT]
perTransactionLimit: 5000 CZK
```

loses its **read**. Path: `GET /accounts/{accountId}` → caller is not the owner → `hasGrant(…, "ACCOUNT_READ_BALANCES")`, no amount → `check()` → `covers(READ_BALANCES, null)` → `perTransactionLimit != null` → `false` → **403 "Account does not belong to caller"**. Same for `mayReadAccount`, and `sharedAccountsFor` drops the account from the shared list entirely.

Not a future payment path — an existing read path, on data that can already exist today.

**The distinction that fixes it:** the ceiling is a property of the *grant*; pricedness is a property of the *question*. Key the denial on whether the capability is amount-bearing (`ACCOUNT_INITIATE_PAYMENT` / `PROPOSE_PAYMENT`, `SAVINGS_DEPOSIT` / `WITHDRAW` / `PROPOSE_WITHDRAW`) rather than on the grant carrying a limit. Alternatively land the customer-edge amount-passing change (#3800) first, so the priced question always arrives priced.

At minimum the KDoc and PR body should stop asserting reads are unaffected, and a test should pin the mixed-capability grant in whichever direction you choose.

One note on the existing suite: `DelegationGrantTest.kt:185` already asserts `covers(ACCOUNT_READ_BALANCES, null) == false`, but only because that grant lacks the capability — so it is not evidence about this behaviour, and it would stay green either way.

Traced statically; I did not run the delegation suite or a customer-edge integration test, so the 403 is derived from the code path rather than observed.

JiRaska:
Merge-order note, found sweeping the open queue: this PR and #4196 touch adjacent lines in the same file — `openbank-delegation-service/.../domain/model/DelegationGrant.kt`.

- **This PR**: `@@ -135,9 +135,27 @@` — fixes `covers()` (unpriced check against a priced grant is now a denial, not coverage) and adds the KDoc above it.
- **#4196**: `@@ -142,21 +142,16 @@` — rewrites the KDoc above `withinLimits()`, the function immediately below `covers()`.

Not duplicates — different fixes (this is #3800/#3615; #4196 is ADR-0249 D3, the spend-reservation cumulative-ceiling mechanism). But the hunks sit ~3 lines apart, which is inside git's usual 3-way merge context window, so whichever of these two lands second is likely to hit a real conflict on rebase — not a silent drop, but worth flagging so it isn't waved through as "just take theirs." When resolving, both `covers()`'s `perTransactionLimit == null` fix and the updated `withinLimits()` KDoc need to survive.

JiRaska:
Correcting my earlier merge-order note on this pair, because I measured it and got a different answer than the one I published.

I said #4101 and #4196 sit ~3 lines apart in `DelegationGrant.kt`, inside git's 3-way context window, so whichever lands second would conflict there. **That is wrong — the Kotlin merges cleanly.** Merging the two branches:

```
4101 + origin/main    clean
4101 + 4196           CONFLICT — docs/adr/DIGEST.md, docs/adr/README.md, docs/adr/index.json
                      DelegationGrant.kt: 0 conflict markers
                      #4101's unpriced-check fix present, #4196's withinLimits KDoc present
```

So the collision is not in the code at all. It is the **derived ADR registry**, which is exactly the class of file that must not be merged: `README.md`, `DIGEST.md` and `index.json` are generated by `docs/adr/gen-index.sh`, and resolving them by hand produces content the next regeneration disagrees with.

**What the second PR to land needs to do**, in this order — the order matters and is a documented footgun:

```
bash docs/adr/gen-index.sh
git add docs/adr/README.md docs/adr/DIGEST.md docs/adr/index.json && git commit
bash .github/scripts/check-adr-registry.sh
```

Regenerate, **commit, then check**. Running the check first is the trap: on a failing run it restores the three files to HEAD on exit, so a later `git add` stages the restored content and the next regeneration disagrees with what was committed — failing the gate on content nobody wrote.

Nothing to do on either branch until one of them merges. Sequencing is free here; the two fixes are independent (#3800/#3615 versus ADR-0249 D3).

Issue #3973, opened 2026-08-07
security-scanner has no outbox write side: the scanner and the outbox are two disjoint package roots, so no scan event has ever been published


JiRaska:
`openbank-security-scanner` contains two disjoint package roots. One scans; the other has the entire
transactional-outbox apparatus. **They never meet, so the service has no write side and has never
published a single event.**

## The measurement

`src/main/kotlin` holds two roots:

```
com/openbank/security/application       <- ports, SecurityOutboxMessage
com/openbank/security/infrastructure    <- repository, dispatcher, Kafka publisher, backlog gauge
com/openbank/securityscanner/application
com/openbank/securityscanner/domain
com/openbank/securityscanner/infrastructure   <- SecurityScannerResource, the @Scheduled scan
```

Nothing in the `securityscanner` tree imports anything from the `security` tree:

```
$ git grep -n 'import com.openbank.security\.' origin/main \
    -- openbank-security-scanner/src/main/kotlin/com/openbank/securityscanner
(no output)
```

Every element of the outbox path has exactly one occurrence in `src/main` — its own declaration, and
no caller:

```
$ git grep -n 'SecurityOutboxMessage(' origin/main -- openbank-security-scanner
  .../application/port/out/SecurityOutboxPort.kt:17:data class SecurityOutboxMessage(

$ git grep -n 'persistInTransaction' origin/main -- openbank-security-scanner
  .../persistence/repository/SecurityOutboxRepositoryImpl.kt:22:    fun persistInTransaction(...)

$ git grep -n 'KafkaSecurityOutboxEventPublisher' origin/main -- openbank-security-scanner
  ktlint-baseline.xml
  .../infrastructure/kafka/KafkaSecurityOutboxEventPublisher.kt:14:class KafkaSecurityOutboxEventPublisher(
  docs/02-architecture.cs.md   (x3)
  docs/02-architecture.en.md   (x3)
```

So `SecurityOutboxDispatcher` polls a table that is **empty by construction**, and
`SecurityOutboxBacklogGauge` reports a backlog of zero that is structurally correct and carries no
information.

The scanner half does run — `SecurityScannerResource.scheduledScan()` fires every 30 minutes and
calls `scanner.scanAll(serviceList())` — and then keeps the result to itself.

## Why this was invisible from every angle you would normally check

- **The pod is healthy.** Ready, no restarts, no errors — an outbox with nothing in it produces no
  failures.
[...]

PR #4102, opened 2026-08-07
feat(ci): a service shipping an outbox dispatcher must construct an OutboxMessage

JiRaska:
## What

`check-outbox-has-writer.py` + a `mode: enforced` entry in `gates.yaml`. A service that ships an
`*OutboxDispatcher` must construct an `OutboxMessage` somewhere in `src/main`.

## Why a gate rather than eight fixes

Eight services ship the whole apparatus — dispatcher, backlog gauge, Flyway migration,
`dispatch-enabled: true` — and write nothing. It is invisible from every angle normally checked:

- the switch **is** set, so the usual "nobody flipped it" check passes
- the dispatcher runs, finds nothing, logs nothing
- the backlog gauge reads **0** — indistinguishable from a healthy drained queue
- the events still reach consumers, through a **second, direct, non-transactional** `@Channel`
  emitter, so nothing downstream notices
- a unit test mocking the repository cannot tell which publisher a use case called

What is lost is the only thing an outbox is for: that the state change and the event commit
together. `PARTY_MERGED` and `AccountStatusChanged` are where that divergence is expensive.

## The predicate is different from #4007's, and #4007's was wrong

That issue was written from a count of `persistInTransaction` call sites. That finds the plumbing,
not the write. Measured both ways on the same tree they disagree on **four** services, in both
directions:

```
interest, sepa-instant   no persistInTransaction caller   but DO construct an OutboxMessage
balance, kyc             repository method present        construct nothing
```

Constructing the message *is* the write, so that is what this checks. I have corrected the membership
on the issue.

## Ratchet, so it can be enforced today

The eight are baselined against #4007 — each needs a per-service wire-or-delete decision that is not
this gate's to make. A **ninth** cannot land quietly. A baseline entry that becomes covered is also
reported, so the list cannot silently become permanent.

## Falsification

On the real tree, reading the exit code **directly** — the first attempt piped through `tail` and
reported `rc=0` for both failure cases, which is the exit-code-masking trap this repo already
documents:

```
clean tree                                 rc=0   8 baselined, 0 new, 0 stale
fx-service's writer removed                rc=1   9 without a writer, 1 new, names the service
bogus baseline entry added                 rc=1   reported as a stale baseline entry
```

**The self-test found a defect in this script's own design.** `classify_service` originally took
pre-stripped input, so all four comment cases failed the moment the test passed raw source — as any
honest test of "does a comment count?" must. A seam where the caller is trusted to sanitise is a seam
where one forgetful caller counts prose as code, which is precisely what this gate exists to prevent.
Stripping now lives inside it.

Self-test cases: a KDoc naming the constructor, a commented-out call, a **nested** block comment
(Kotlin's nest, so the stripper is a depth counter, not a regex), a string literal containing
`OutboxMessage(`, the `data class` declaration itself — none may count — and a real call *after* a
comment, which must.

## Scope

Closes nothing on its own. The eight stay open on #4007; this stops a ninth.

Refs #4007, #3973

### DIFF
--- a/.github/gates/gates.yaml
+++ b/.github/gates/gates.yaml
@@ -1576,6 +1576,32 @@ gates:
     run: |
       bash .github/scripts/resolve-record-deployment-version.sh --self-test
 
+  # A service that ships an outbox must write to it (#4007). Eight do not: dispatcher, backlog
+  # gauge, migration and `dispatch-enabled: true`, with no OutboxMessage construction anywhere in
+  # src/main — so the dispatcher polls a table nothing writes to and the gauge reads 0, which is
+  # indistinguishable from a drained queue. Those events still reach consumers, through a second
+  # direct non-transactional emitter, so nothing downstream notices either.
+  #
+  # RATCHET: today's eight are baselined against #4007 (each needs a per-service wire-or-delete
+  # decision this gate does not make), so a NINTH cannot be added quietly. A baseline entry that
+  # becomes covered is also reported, so the list cannot quietly become permanent.
+  #
+  # Falsified both ways on the real tree before wiring in, reading the exit code directly rather
+  # than through a pipe: removing fx-service's writer -> rc=1 with the service named; adding a
+  # bogus baseline entry -> rc=1 as a stale entry; clean tree -> rc=0. The self-test additionally
+  # covers the code-about-code cases — a KDoc, a commented-out call, a NESTED block comment and a
+  # string literal naming the constructor must none of them count as a writer.
+  - id: outbox-has-writer
+    name: "a service shipping an outbox dispatcher must construct an OutboxMessage (#4007, enforced)"
+    group: kotlin
+    mode: enforced
+    selftest: |
+      python3 .github/scripts/check-outbox-has-writer.py --self-test
+    selftest_expect: pass
+    run: |
+      python3 .github/scripts/check-outbox-has-writer.py --self-test
+      python3 .github/scripts/check-outbox-has-writer.py .
+
   - id: pact-provider-replay-coverage
     name: "Pact provider-replay coverage (issue #2338, enforced)"
     group: api
--- a/.github/scripts/check-outbox-has-writer.py
+++ b/.github/scripts/check-outbox-has-writer.py
@@ -1,0 +1,226 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) OpenBank contributors. Licensed under the Apache License, Version 2.0.
+# See LICENSE in the repository root or https://www.apache.org/licenses/LICENSE-2.0 for details.
+#
+# ---------------------------------------------------------------------------------------
+# A service that ships an outbox must write to it (issue #4007).
+#
+# THE DEFECT THIS CATCHES
+# Eight services ship the whole transactional-outbox apparatus — a `*OutboxDispatcher`, a backlog
+# gauge, a Flyway migration for the table, `openbank.outbox.dispatch-enabled: true` — and construct
+# no `OutboxMessage` anywhere in `src/main`. The dispatcher polls a table nothing writes to, forever.
+#
+# It is invisible from every angle normally checked:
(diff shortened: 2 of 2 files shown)


JiRaska:
Verified rather than trusted the description — three things:

**The claim ("8 services, invisible from every angle") holds.** Running the checker against the tree: `32 service(s) ship a dispatcher; 8 without a writer (8 baselined, 0 new, 0 stale)`. Matches exactly.

**No `--enforce` flag exists on this script** — `mode: enforced` in `gates.yaml` is not vacuous here the way it has been for other checkers in this repo (a checker that prints `::warning` and exits 0 without a flag). `check-outbox-has-writer.py`'s `main()` returns `1 if (new or stale) else 0` unconditionally; there is no advisory mode to forget to flip. Confirmed by removing one baseline entry and re-running: `::error`, `1 new`, real exit 1 (measured via `$?` right after the python invocation — piping through `tail` first, as I did on my first attempt, reports `tail`'s exit code instead and reads as green regardless).

**The self-test covers exactly the shapes this repo has been burned by before**, all in one script: a KDoc mentioning `OutboxMessage(`, a commented-out construction, a nested block comment, and the constructor name inside a string literal — none of which is a real write, and all nine self-test cases pass.

Not merging — governance/gate machinery, human review by rule regardless of the diff.

JiRaska:
Reviewed. **Approve** — falsified in the direction that matters, correctly registered, and the scope question has a good answer.

**Verified by running, not reading:**

- `--self-test` → 9/9 `ok`, `PASS`, rc=0 (exit code read directly, not through a pipe — `| tail` masks it).
- Real tree → `32 service(s) ship a dispatcher; 8 without a writer (8 baselined, 0 new, 0 stale)`, rc=0.
- **Known-positive the self-test does not model** (it only drives synthetic strings): copied `openbank-fx-service` and `openbank-audit-service` to `/tmp` and renamed the single real `OutboxMessage(` construction in `FxService.kt` → **rc=1**, printing the actionable line that names the service: `::error::openbank-fx-service: ships an *OutboxDispatcher but constructs no OutboxMessage in src/main`. So it can express the failure it is being trusted about.

**Registration and enforcement agree on both axes:** `run-gates.py --list` shows `kotlin  outbox-has-writer  enforced  always  selftest` — declared in `gates.yaml`, not an inline `run:` in a conditional job. `selftest_expect: pass` is right (the command *is* the checker's own harness). There is no `--enforce` flag and none is needed: the script returns 1 unconditionally on `new` or `stale`, so `mode: enforced` is not vacuous.

**On the `dispatch-enabled` axis** — worth stating because it is the obvious hole to suspect here, given `openbank.outbox.dispatch-enabled` defaults to **false** and a service can ship a dispatcher that never dispatches. It is *not* a hole in this gate: the sibling enforced gate `outbox-dispatch-enabled-guard` owns it and fails closed on a missing key, `false`, or a `${VAR}` with no default. The two compose. Worth a line in the gates.yaml comment so the next reader does not have to re-derive that.

**The residual scope hole, measured rather than assumed:** coverage is derived from a `*OutboxDispatcher` class existing, so a service with an outbox table and no such class reads as out-of-scope-and-therefore-passing — the shape this repo keeps re-encountering. Checked: **zero** services today have an outbox migration without a `*OutboxDispatcher.kt`, and the only module referencing `AbstractOutboxDispatcher` without one is `openbank-libs-testing`, correctly excluded by the libs prefix rule. Latent, not live.

**Two nits, neither blocking:**

1. **The stale-baseline message is wrong for the absent case.** A baselined service that is deleted, renamed, or simply not in the scanned root drops out of `violations` and is reported as *"…but it now has one."* My subset run printed exactly that for all 7 absent services. It still fails, which is the right direction — but it sends the reader hunting for a writer that does not exist. Distinguishing "no longer in scope" from "now has a writer" is a one-line change.
2. `DECLARATION_RE` is applied per file, so a file that both declares `data class OutboxMessage(` and constructs one would count as a non-writer. Only `openbank-libs-domain` declares it and libs is excluded, so it cannot bite today.

PR #4104, opened 2026-08-07
docs(governance): ADR-0248 — statement and payment confirmation document templates

JiRaska:
## Summary

Adds **ADR-0248**, deciding the architecture for three customer-facing document types the platform is currently missing:

1. **Monthly statement** (`MESICNI_VYPIS_CS`/`_EN`) — PSD2 Art. 58(2).
2. **Annual statement of fees** (`ROCNI_VYPIS_POPLATKU_CS`/`_EN`) — PAD Art. 5, owned by billing-service (new annual aggregation use case, not built today).
3. **Payment confirmation** (`POTVRZENI_O_PLATBE_CS`/`_EN`) — PSD2 Art. 45/48.

All three render through `openbank-document-service`'s existing Handlebars template engine, and **none of the three is ever persisted**: each renders synchronously via the existing non-persisting `POST /api/v1/documents/templates/preview` endpoint and gets discarded — no `Document` row, no object-store write, no `document.generated` outbox event. Monthly statement and payment confirmation are rendered only on the customer's own request (a download click); the annual statement is the one genuine push duty (PAD Art. 5), so it alone keeps an async Kafka trigger off billing-service's outbox. `openbank-statement-service`'s own camt.053/MT940/pocket-PDF rendering (ADR-0035) is unchanged; this ADR extends that same "persist the model, not the files" guarantee to the new styled/PDF layer.

The ADR includes the full required-field list per template (PSD2/PAD legal basis), and flags that the annual statement needs legal review of the fee-category → PAD Annex II mapping before production use.

## Linked issues

Refs #4105

## What this PR does NOT do

No Kotlin/YAML implementation — this is the decision + template list only. Implementation (the one new Kafka consumer, the billing-service aggregation use case, the 6 seed rows + Handlebars bodies, and two threat-model documents) is tracked in #4105 per ADR-0052.

## Test plan

- [x] `bash docs/adr/gen-index.sh && bash .github/scripts/check-adr-registry.sh` — clean after each commit (front-matter schema, tags, collision-free number all validate).
- [x] `git diff` is scoped to the ADR file plus the derived index files (README/DIGEST/index.json).


JiRaska:
Added a rule at the client's request: both statement documents (monthly, annual) now render **on-demand only** — at the moment the customer clicks download, or the annual push-send fires — via document-service's existing non-persisting `POST /api/v1/documents/templates/preview` endpoint, then get discarded. Nothing rendered from a statement template is ever written to document-service's object store; this extends ADR-0035's "persist the model, not the files" guarantee to the new styled/PDF layer.

Payment confirmation is unchanged — it keeps the persisted `/api/v1/documents/render` path, since it's evidence a specific payment executed rather than a reproducible view over data already retained. The ADR now calls that asymmetry out explicitly instead of leaving it implicit.

Net effect on the design: no Kafka consumer needed for the monthly statement anymore (nothing to trigger at period-close since rendering is deferred to click time); the annual-statement consumer still exists but now calls the non-persisting preview endpoint and hands bytes straight to delivery instead of calling the persisting render endpoint.

JiRaska:
Extended the same rule to payment confirmation, per follow-up request: it now also renders synchronously, only on the customer's own request (e.g. "download confirmation" in transaction history), via document-service's non-persisting `preview` endpoint — never written to the object store.

This drops the two SEPA/domestic payment Kafka consumers from the design entirely. The only remaining async trigger in this ADR is the annual statement's consumer, since PAD Art. 5 makes it a genuine push duty the bank must initiate rather than something the customer requests.

Net result: all three document types share one discipline — render, hand off, discard. Zero new `Document` rows or `document.generated` outbox events from any of them. Follow-up issue #4105 updated to match.

PR #4106, opened 2026-08-07
feat(engagement): materialise ARREARS + ERASURE_REQUESTED for the D3.5 targeting exclusion

JiRaska:
## Summary

Issue #2749 D6/D3.5 residual scope: `EligibilitySnapshot.adverseState` was always constructed empty (an honest placeholder documented in `ResolveSurfaceUseCase`), so the vulnerable-customer targeting exclusion existed on paper but never actually excluded anyone.

Verified against `origin/main` (not assumed) which of the four `AdverseState` signals have a real event to consume, before building anything — see the [D6 survey](https://github.com/JiRaska/open-bank-oss/issues/2749#issuecomment-5219964259) and its [correction](https://github.com/JiRaska/open-bank-oss/issues/2749#issuecomment-5220777083) on issue #2749:

- **ARREARS**: `loan.stage_changed` fires on every genuine IFRS 9 stage transition and already carried `daysPastDue`; added `partyId` to the payload (additive, harmless to anacredit-service's existing `LoanStageEventConsumer`) so a consumer can act on it directly.
- **ERASURE_REQUESTED**: `PARTY_ERASED`, the same fleet-wide event every other `PartyEventConsumer` sibling already handles. Deliberately never cleared — no "un-erase" event exists anywhere.
- **FRAUD_HOLD** and **DISPUTE_OPENED** are **not** wired — neither is published as an event anywhere in the fleet today. Fixed a stale KDoc in `Eligibility.kt` that claimed all four were already emitted fleet-wide (they are not).

New: `LendingArrearsEventConsumer`, `PartyErasureConsumer`, `AdverseStateRepository`/`-Impl` (surrogate id + unique `(party_id, state)` business key, same convention as party-service's `party_marketing_consent` — an app-assigned natural-key PK is INSERT-only in Hibernate Reactive). `ResolveSurfaceUseCase` now reads the real snapshot.

Also lands engagement-service's **Kafka mTLS** (`KafkaUser` + keystore/truststore `ExternalSecrets` + SSL env) — did not exist at all before this PR. Fixes a latent bug along the way: `openbank.outbox.dispatch-enabled` is `true` in `application.yaml` (not the `false` default an earlier gitops comment assumed), so the D2 feedback-loop outbox had been attempting every publish over plaintext against the mTLS-only `:9093` listener since #4048 merged — an immediate TLS-handshake failure per attempt, never actually delivering.

**Two pre-existing test-config regressions found and fixed** while verifying this change, both dating to #4054's `@RolesAllowed`/`@Authorize` addition and never caught because `SurfaceRestContractIT` hadn't run since:
- No `@TestSecurity` on the IT class → every call 401'd.
- No `authz.enforce: false` in `%test` → every call then 503'd (no PDP bean in a bare `@QuarkusTest`).

**Not attempted here, deliberately** (per the #2749 issue comments): notification-service's own `ContactPolicyGate` wiring (reactive-model mismatch + a real audit-granularity design decision, not a mechanical swap) and `FRAUD_HOLD`/`DISPUTE_OPENED` (blocked on upstream modeling work in fraud-service/dispute-service that doesn't exist yet).

## Test plan
- [x] `:openbank-engagement-service:test` (32 tests, incl. new `LendingArrearsEventConsumerTest`, `PartyErasureConsumerTest`, and a new `ResolveSurfaceUseCaseTest` case proving arrears actually empties the surface, not just changes the result type)
- [x] `:openbank-lending-service:test` (existing `loan.stage_changed` emit tests, no exact-payload-string assertions to break)
- [x] `ktlintCheck` / `detekt` clean on both modules
- [x] `SurfaceRestContractIT` green end-to-end (real Postgres + in-memory Kafka channels) after the two test-config fixes


JiRaska:
**Flyway version collision between two open PRs — whichever merges second breaks engagement-service at startup.**

`openbank-engagement-service` has exactly one migration on `main` (`V1__init_engagement.sql`). Two open PRs each claim `V2`, with different content:

| PR | file |
|---|---|
| #4106 | `V2__party_adverse_state.sql` |
| #3702 | `V2__engagement_outbox.sql` |

Neither conflicts with the other in git — they are different filenames, so a merge is clean and silent. Flyway is not: two migrations resolving to version `2` make it refuse to start (`Found more than one migration with version 2`), so the second PR to land takes the service down on its next deploy with nothing in either diff hinting at it.

Cheapest fix is for whichever lands second to renumber to `V3`. Worth deciding the order deliberately rather than letting merge timing decide it.

**#3702 additionally collides on `V1`**, which is the worse half and is described in detail on that PR: its `V1__init_engagement.sql` creates `engagement_profiles` while `main`'s creates `engagement_event` + `engagement_outbox`. Editing a migration `main` already ships is a checksum mismatch, so that one cannot be resolved by picking a side either. Its `V2__engagement_outbox.sql` is also redundant — `main`'s `V1` already creates that table.

So the ordering that needs the least rework: land #4106 as `V2`, and renumber #3702's profiles migration to `V3` while dropping its outbox migration entirely.

JiRaska:
Also refreshed against `main` (was 5 behind, conflict-free) — the red `gates (gitops)` was **not** this PR's.

The shard failed on three things this PR does not touch:

```
NEW domain @Scheduled job without registerWorkflowLiveness (ADR-0237)
openbank-lending-service polls Temporal namespace `openbank-lending`, which the registration ConfigMap does not declare
openbank-camp-service polls Temporal namespace `openbank-camp`, …
```

This PR changes only `openbank-engagement-service/**` — no scheduler, nothing under `openbank-infra`, nothing Temporal-related (`git diff --name-only origin/main...HEAD | grep -iE 'schedul|temporal'` is empty). And `main`'s own `gates (gitops)` is **success** (run on `b4d16ee69`, all eight shards green), so those findings were fixed on `main` within the five commits this branch was behind.

Re-running against current `main` now.

PR #3807, opened 2026-08-06
security(sdd): bootstrap the OPA PDP sidecar and sdd.* policy rules

JiRaska:
> ### ⚠️ Overlaps #3697 — read this first
>
> I opened this before discovering **#3697** (`feat(sdd): add OPA sidecar, matrix grants, and AUTHZ_ENFORCE=true`, open since 2026-08-03), which covers the same service. That was my miss. **Only one of the two should land.** They differ in the grant mechanism and I measured the difference rather than arguing it:
>
> | | #3697 | this PR |
> |---|---|---|
> | grant mechanism | `role_action_matrix.ROLE_OPERATOR` | service-local `sdd_rest_ext.rego` (delegation/consent pattern) |
> | shared backend M2M (`service-account-openbank-services`) on `sdd.authorise` — the call that books a debtor debit | **ALLOW** (`matrix-allows`) | **DENY** |
> | …on `sdd.create/approve/update/delete` | **ALLOW** | **DENY** |
> | edge principal on `sdd.approve` / `sdd.authorise` | ALLOW | DENY |
> | `AUTHZ_ENFORCE` | flipped to `true` in the same PR | left `false`, deliberately |
> | files touched | `rules.yaml` ⇒ all 42 bundles + annotations restamped | `components/sdd/` only |
>
> Both columns were produced by `opa eval` against each branch's **committed** bundle, in runs whose must-DENY and must-ALLOW controls both behaved. `matrix-allows` has no `service-account-*` exclusion and cannot be added to `shared_m2m_write_prohibition.reasons` — that is the fleet-wide residual in **#3765**. Detail and the full table are in my comment on #3697.
>
> A third option exists (keep #3697's structure, add the exclusion at the matrix layer) and is really #3765's decision to make. I am not closing #3697 and have not touched its branch.

## What and why

`openbank-sdd-service` is in `rules.yaml: money_path_services` — it posts a real debtor debit through `TransactionServiceClient` for every authorised collection (#1478) — yet it shipped with **no OPA policy-decision sidecar and no policy bundle of any kind**, while its manifest justified `AUTHZ_ENFORCE=false` with the comment *"non-money-path services run authz advisory"*. Its ten `@Authorize` annotations were decorative twice over: advisory mode never denies, and with no PDP to ask there was not even a decision to log.

This is the precondition work for #3679. **It does not flip `AUTHZ_ENFORCE`.**

## What was measured, not inferred

Live pods (read-only `kubectl`):

[...]

PR #4113, opened 2026-08-07
security(authz): bar the shared M2M identity from minting MCP sessions

JiRaska:
## What

One line in `openbank-libs/governance/policies/rest.rego`, plus the 41 bundles that embed it:

```diff
 allowed_reasons contains "operator-mcp-session" if {
     input.principal.type == "HUMAN"
     some role in {"ROLE_OPERATOR", "ROLE_ADMIN"}
     role in input.principal.roles
+    not startswith(input.principal.id, "service-account-")
     startswith(input.action, "mcp.session.")
 }
```

## Why

The rule granted `mcp.session.create` / `.read` / `.bind` / `.revoke` to **any** HUMAN carrying
`ROLE_OPERATOR`. Keycloak `client_credentials` tokens never produce `principal.type == SERVICE`
(`rules.yaml: authz_policy.principal_type_service_unreachable`), and
`service-account-openbank-services` carries `ROLE_OPERATOR` in the realm — so the fleet's shared
backend identity could mint, bind and revoke staff agent sessions. **The rule read as staff-only and
was not.**

This is the same one-line idiom, for the same reason, as `operator-compose-message` immediately
below it — which carries the explanatory comment this rule was missing. That comment ends *"The
explicit `service-account-` exclusion below is what actually keeps it out"*, and this rule did not
have it.

## Nothing legitimate is cut

The only consumer of these endpoints is admin-ui's OBO relay
(`src/app/api/agent/obo-mcp/route.ts`), which exchanges a token for a signed-in operator:

> mcp-service sees a HUMAN principal whose realm roles were bounded at issuance

A real person, never a service account. I checked before excluding rather than after.

## Proven by a counter-example that reaches the code

```
exclusion removed   test_service_account_denied_mcp_session: FAIL     102/106
exclusion restored  both new tests pass                               103/106
```

The first attempt at that removal **did not actually remove anything** — the script sliced to the
`}` of the set literal `{"ROLE_OPERATOR", "ROLE_ADMIN"}` rather than the rule's closing brace, so the
tests "passed" with and without the fix and proved nothing. Re-done line-precise with an assertion
that exactly one exclusion line was found.

Baseline for the 3 remaining failures: **identical on `main` (101/104)** — charter-bridge tests that
need `data.agents`, not loaded when the shared policies are tested alone.

## Also: two stale baseline entries removed

`check-operator-write-naming.py` still baselined `operator-fx-trigger` and
`operator-fx-approval-decide`. #3734 gave both the service-account guard on **2026-08-05**
(`fx_rest_ext.rego:50` and `:96`, recorded in `docs/threat-models/openbank-fx-service.md`). The debt
was paid; only the list lagged. The gate had been reporting both as stale on `main` ever since — in
advisory mode, which is why nobody acted.

## Bundles

`rest.rego` is embedded in all 41 bundles carrying this rule. Regenerated, verified **idempotent**, and
verified that every one of the 41 now carries the exclusion. `opa-bundle-parses`,
`opa-sidecar-bundle-shape`, `opa-bundle-apply-size` and `no-dead-code-service-principal-rego-rule` all
pass.

**Short shelf life** — 75 files under `gitops/`, so a competing governance PR will conflict it. Worth
merging with `--auto` rather than leaving it to sit.

## What this deliberately does NOT do

**It does not flip the gate to `enforced`,** which is what I was asked for and which turned out to be
wrong as stated. `check-operator-write-naming.py` ends:

```python
return 1 if (enforce and baselined) else 0
```

so `--enforce` fails while **any** baseline entry remains — the baseline is a debt list, not an
allowance. Five rules are still on it: `operator-anacredit-create`, `operator-party-status`,
`operator-pid-resolve`, `operator-standing-order-pause`, `operator-vop-verify`. Each needs the same
"is any M2M caller legitimate?" check this one got, and `operator-pid-resolve` is even annotated
*"classified here pending confirmation"*. Enforced means zero debt, so the flip is a separate PR after
those five — not a one-line `mode:` change.

Refs #3765, #3931


JiRaska:
Two further commits, one of which corrects a defect I introduced in this PR.

**`e2cc84133` — restored the `outbox-has-writer` gate this branch had deleted.** Merging main in dropped the 26-line gate entry #4102 had just added to `.github/gates/gates.yaml`. No conflict was reported and nothing in the merge output said so. This branch never had a legitimate reason to touch that file, so it is restored to main's content verbatim — 119 gates, entry present.

It is the failure mode this repo already documents: a merge git calls clean can still delete content when two sides add neighbouring entries to the same list. My check after that merge counted the *file* list (76 files, 3 sources, 73 generated — the expected shape) and never counted the *entries* inside a structured file, which is the half that matters. A gate that silently stops being declared reads as green forever.

**`900a4c3a8` — graduated `operator-write-naming` from advisory to enforced**, which is what its own note asked for: *"Graduate to enforced once the finding is closed — tracked on #3240."* This PR closes that finding and clears the two stale `BASELINE` entries that had accumulated meanwhile, so the condition now holds. `origin/main` exits 1 (1 new + 2 stale); this branch exits 0.

Advisory was doing less than it looked. Both ratchet halves already `return 1`, and `mode: advisory` converts a script's exit 1 into a warning at the runner — so a brand-new operator write escaping the shared-M2M prohibition was **detected and not blocked**, on a rule whose own text calls registering one without it a money-path outage rather than a test failure.

The run line deliberately does **not** pass `--enforce`: that flag fails on the baselined debt itself and would make the gate unmergeable. `mode: enforced` blocks only what the ratchet returns 1 for, so the 5 baselined rules warn, pass, and can be worked down per rule.

Falsified both ways on the real tree — a new role-only write rule injected into `rest.rego` gives rc=1 and names it; a bogus `BASELINE` entry gives rc=1 as stale; the restored tree gives rc=0.

Worth recording: my **first** injection reported rc=0 and looked like proof the gate was fine. It was written in Rego v0 syntax (`allowed_reasons["x"] { … }`) while `REASON_RE` matches only the v1 idiom this fleet uses (`allowed_reasons contains "x" if { … }`), so the counter-example never reached the code. The tell was the checked-rule count staying at 100 instead of going to 101 — a falsification needs an expected *number*, not just an expected exit code.

JiRaska:
Reviewed. **Approve** — the change does what it says, and no legitimate caller loses a path. One thing the description implies that the policy does not do, worth a follow-up rather than a block.

**Measured against `main`'s committed bundle, not the branch's embedded copy** — a branch behind `main` answers about a policy the fleet no longer runs. Real `toInput()` shape (`resource` = `type` + `id` only; inventing `resource.attributes.tenant` reaches `operator-on-own-tenant`, which is dead in production and allows nonsense actions). Controls in every run: `zzz.frobnicate` denied on both sides, `account.read` allowed on both.

Only six cells move, all in the intended direction:

| principal | action | `main` | this branch |
|---|---|---|---|
| `service-account-*` (ROLE_OPERATOR) | `mcp.session.create` | allow (`operator-mcp-session`) | **deny** |
| | `mcp.session.bind` | allow | **deny** |
| | `mcp.session.revoke` | allow | **deny** |
| | `mcp.session.read` | allow (`operator-mcp-session`) | allow (**`operator-read-any`**) |
| human OPERATOR / ADMIN | all four | allow | allow |

**No outage.** The service's whole `@Authorize` surface is those four actions (`McpSessionResource.kt:72,101,112,140`), and every human row is byte-identical between `main` and the branch.

**The follow-up:** `mcp.session.read` **stays open** to the shared M2M identity — it simply changes which rule admits it, from `operator-mcp-session` to `operator-read-any`. I verified the reason string directly. The body says the identity "could mint, bind and revoke", which is accurate about what closes, but a reader will reasonably infer the read closed too. If enumerating other operators' session metadata is sensitive, that is a second PR against `operator-read-any` — not this one.

**Also checked, because both have bitten this repo recently:**

- Merging `main` in is clean, and the neighbouring-entry deletion did **not** recur: `gates.yaml` has 119 `- id:` entries on both sides and `outbox-has-writer` appears once. That failure mode — two sides appending to the same list, git reporting no conflict, one entry silently gone — is why this is worth counting rather than eyeballing.
- All **42** bundles carrying `rest.rego` match the source byte-for-byte after the merge (`DRIFTED: 0`), so the regeneration is genuinely current despite the branch being 29 behind. A clean merge of a generated file is not the same as a current one — that cost #3807 a 246-commit-stale bundle yesterday.
- The gate is falsified **both ways** on the real tree: `origin/main` → rc=1 naming the `operator-mcp-session` finding plus the two stale baselines; branch → rc=0, `100 rules checked, 5 baselined, 0 new`. The advisory→enforced graduation is earned, and the reasoning for not passing `--enforce` (it would fail on the debt list itself) is right.
- No `principal.type == "SERVICE"` rule is introduced. That type can never be emitted by `AuthorizeInterceptor`, so a rule gated on it would be unreachable dead code; this correctly discriminates on `principal.id`.

JiRaska:
**Merge-order note: this PR and #4113 overlap on 73 files, and both are generated bundle restamps.**

Measured from the files API rather than the diff (`gh pr diff --name-only` returns nothing on a 125-file PR):

| | files | overlap |
|---|---|---|
| #4113 (`operator-mcp-session` M2M exclusion) | 77 | — |
| #4173 (derived `agents.yaml` subset) | 125 | **73** |

Both restamp `*-opa-bundle.yaml` and the pod-roll annotation in `*-service.yaml` across the fleet, so whichever lands second **will** conflict — and the resolution is never a hand-merge of a generated file. Whoever merges second should take `main`'s side on every generated file, then regenerate in the documented order as the **last** step before pushing:

```bash
```

Suggested order: **#4113 first** — it is smaller, it is a security fix (barring the shared M2M identity from minting MCP sessions), and its source change is one line of `rest.rego` that survives regeneration trivially. Then #4173, regenerated on top.

Both are `MERGEABLE` right now and both have a shelf life measured in hours. The two red checks on each — `Trivy` and `Customer-app dossier` — are inherited fleet-wide faults, fixed by #4153 and #4053 respectively; neither diff touches `package-lock.json` or `app-status.json`.

Merge with `--squash --auto`, never `--admin`.

JiRaska:
Refreshed against `main`. #4173 merged first, and since both PRs restamp the same generated bundles, this branch picked up **73 conflicts** — all of them in derived files, none in a source.

**Resolved by regenerating, not by merging.** Merge-resolving a generated file produces something that matches neither side's sources; the only correct move is to take `main`'s copy and re-run the generators over this branch's (unconflicted) source. Verified afterwards, in order:

```
conflicts outside derived artifacts        0
conflict markers surviving regeneration    0
check-opa-bundle-parses.py                 OK: 42 bundle(s), no unloadable policy
idempotency spot-check (8 generators)      0 files changed on a second run
```

**The source change survived the merge** — `rest.rego`, `check-operator-write-naming.py`, `mcp_excl_test.rego` and the `gates.yaml` entry are all still in the diff against `main`.

**And the property this PR exists for still holds on the regenerated bundle**, which is the check that actually matters — a bundle can parse perfectly and still have lost the rule:

```
service-account-openbank-services   mcp.session.create  -> false
                                    mcp.session.read    -> allow
                                    zzz.frobnicate      -> false   (control)
```

Unchanged from what I measured in review: the three writes close, `mcp.session.read` stays open via `operator-read-any`, and the control denies so the table is not vacuous.

**Worth recording for the next pair of bundle-restamping PRs**: this collision was predictable from the file lists alone — I flagged the 58-file overlap with #4173 before either merged. Whichever lands second always needs regeneration rather than conflict resolution, so it is cheaper to merge them back to back and regenerate once than to let a queue drain interleave them.

Issue #4099, opened 2026-08-07
engagement-service is a released component with no governance.yaml, and the gate's path filter cannot see a module that omits one


JiRaska:
## The drift

`openbank-engagement-service` is a released component — `version.txt` at `0.2.1`, registered in `release-please-config.json` and the manifest — and has **no `governance.yaml`**. It is the only such module on `main`: enumerating every `openbank-*` directory with a `version.txt` and no manifest returns exactly one hit.

`main` is red on it right now. Run `31207937404` on `f0db65e39`: `Admin UI build` and `Admin UI` both `failure`, with

```
Governance drift — 1 module(s) missing or with an invalid governance.yaml (ADR-0071 Phase 4)
```

Not currently merge-blocking, since neither job is a required check — but #3692 proposes to make the Admin UI gate blocking, and cannot land until this is fixed.

## Why nothing caught it when the service landed

This is the interesting half, and it is a scope hole rather than a missing gate.

`ci.yml`'s `changes-ui` job decides whether the governance gate runs at all, filtering on:

```
openbank-admin-ui/   */governance.yaml   openbank-libs/governance/governance.schema.json
```

Its own header explains the last two are there so a backend PR that adds a service still reaches the gate — the #2236 fix, after #2124 and #2104 each shipped a `governance.yaml` with zero admin-ui files touched.

But the filter keys on the manifest **being present**. A PR that adds a released module and *omits* `governance.yaml` touches none of the three pathspecs, `ui-build` skips, and the always-on `Admin UI` gate reads a skip as a pass. **The one case the scope cannot see is the case the gate exists for.** engagement-service landed across #4025/#4048 exactly that way, and the drift sat on `main` until an admin-ui-touching PR happened along.

This is the same shape as the `pact-drift-check` scope problem: a gate whose coverage set is derived from the artifacts it checks reads as *passing* when an artifact is absent, never as *unchecked*.

## Two fixes, independent

**1. Add the manifest.** Most of it is derivable from what the service already declares:

```yaml
primaryDatastore: PostgreSQL          # CNPG instance in components/engagement/postgres.yaml
databaseName: openbank_engagement     # from the workload's JDBC / reactive URLs
dataLineageRole: both                 # consumes upstream events; produces via engagement-outbox-out
lineage:
  upstream:
    - serviceName: consent-service    # CONSENT_SERVICE_URL — the consent gate
      relationType: api
databaseLineage:
  ownedDatabases: [openbank_engagement]
```

[...]

PR #4114, opened 2026-08-07
fix(engagement): add the missing governance.yaml — main is red without it

JiRaska:
fix(engagement): add the missing governance.yaml — main is red without it

main has been failing CI on the last three commits:

    [generate-governance] 57 modules (42 with lineage, 0 unverified databaseName, 1 gaps)
      - openbank-engagement-service: missing governance.yaml
    ::error::Governance drift — 1 module(s) missing or with an invalid governance.yaml (ADR-0071 Phase 4)

`openbank-engagement-service` landed in #4025 and grew through #4054 without one, so every PR that
merges main in inherits the red — that is how I found it, on an unrelated governance PR.

The file already exists in #3702, which is 286 commits behind with 12 conflicts and cannot land
soon. Lifting just this one file out unblocks main now and costs #3702 nothing: it can drop its copy
or merge cleanly against this.

**Values verified against the service as it is on main, not copied from #3702** — this repo's own
note is that nothing checks a governance manifest's truth and its fields rot silently:

  - `primaryDatastore: PostgreSQL` / `databaseName: openbank_engagement` — the service declares
    `jdbc:postgresql://localhost:5432/openbank_engagement` and ships `db/migration/V1__init_engagement.sql`.
  - `dataLineageRole: both` — it consumes (ConsentServiceClient, the contact gate) and produces
    (KafkaEngagementOutboxEventPublisher via the transactional outbox).

`dataClassification`, `retentionPolicy` and `dataDomain` are #3702's author's declarations, kept as
written: they are policy choices, not facts I can read off the tree, and inventing different ones
would be worse than carrying theirs.

Structure checked against all 57 existing manifests: the six keys present in every one of them are
present here, nothing extra, and every value is drawn from the vocabulary already in use
(`platform` 19×, `PostgreSQL` 54×, `both` 29×, `confidential` 28×).

Refs #3702, #4025, #4054.

## Linked issues

N/A — main-unblocking fix for a gap introduced by #4025/#4054; the file it restores is already
authored in #3702, which cannot land soon.


JiRaska:
Pushed two things to this branch: the owner's retention value, and a merge of current `main`.

**`retentionPolicy: 2 years` → `7 years`.** The 2-year figure was not a fact about the service. I hit the same wall building this manifest independently and stopped there rather than pick a number: `retentionPolicy` is a required free-text string with **no** `unknown` option — unlike `dataClassification`, which has one — so the schema forces a value where it permits an honest "undeclared" one field earlier. That asymmetry is the real defect and is written up on #4099. The service owner's answer is **7 years**, which is also the established spelling (16 fleet manifests use `7 years`).

**Verified after the change**, with the real generator on a worktree merged with current `main`:

```
node scripts/generate-governance.mjs   →  exit 0,  withGaps = 0
```

So the ADR-0071 drift keeping `main` red is closed by this branch.

**The red checks were stale, not yours.** `gates (gitops)` was failing on a scheduler heartbeat and on the `openbank-lending` / `openbank-camp` Temporal namespaces — none of which this PR touches; it changes a governance manifest, `discovery.ts` and an admin-ui manifest. Those findings were fixed on `main` while this branch sat behind it. Merged `main` in (0 conflicts) so CI answers against the current tree.

**Two fields left exactly as you authored them, flagged rather than changed:** `dataClassification: confidential` — the schema does offer `unknown`, so this is a claim rather than a derivation, though it is the conservative direction — and `evidenceExported: false`. Both are yours to keep or revise; I only replaced the one the owner ruled on.

Worth merging promptly: `main` has been red on `Admin UI build` for hours, and #3692 (which makes that gate merge-blocking) is waiting on it.

JiRaska:
Cannot merge this yet — two required checks are red on the **current** head (`6c73f8d5b`, runs created a minute after the push, so this is not a stale rollup; I checked the timestamps against the push before saying so). One of them is not this PR's fault, and finding that out turned up something worth its own issue.

### `SurfaceRestContractIT` — 4 failures, and this PR is exercising them, not causing them

```
SurfaceRestContractIT > an eligible party sees the catalogued banner            FAILED (:42)
SurfaceRestContractIT > a party without marketing consent is not_eligible …     FAILED (:42)
SurfaceRestContractIT > three posted dismissals suppress the next resolve …     FAILED (:74)
SurfaceRestContractIT > an unknown slot is a 400, not a 500 or a silent empty … FAILED (:92)
```

This PR adds one file to engagement-service — `governance.yaml`. It cannot break a REST contract IT. What it *does* is touch an `openbank-engagement-service/**` path, which makes path-scoped Services CI select that service and run its tests.

**Nothing else currently does.** On `main`'s own run (`31216551896`), Services CI reports:

```
success  Detect changed services
skipped  build            <-- engagement was never built
success  all-green
```

The push touched no service path, so the build was skipped and `all-green` went green over zero services. And the two other open PRs that build engagement — #4106 and #4072 — pass, but they sit **8** and **62** commits behind `main` while this branch is **1** behind. So the breakage is recent, and the only tree new enough to meet it is this one.

Net: engagement-service's REST contract ITs are red on `main` right now, and `main` structurally cannot see it. That is the "a gate is only as reachable as the job it sits in" shape, one layer out — here the job is reachable, but its *scope* is derived from which paths a commit touched, so a service breaks silently until an unrelated PR happens to name it.

### `gates (gitops)` — temporal namespaces + scheduler heartbeat

Passes on `main`, passes when I run both gates locally on this branch (`temporal-namespace-registration` exit 0, `scheduler-liveness-adoption` advisory-warn/exit 0), and fails in CI. CI evaluates `refs/pull/4114/merge`, so it is judging branch+main combined; `main` has moved twice since. Not diagnosed further — it is downstream of the same moving target.

### What I did push

`retentionPolicy: 2 years` → **7 years** (the service owner's answer), and a merge of `main`. Verified with the real generator on the merged tree: `node scripts/generate-governance.mjs` exits 0 with **`withGaps = 0`**, so the ADR-0071 drift keeping `main`'s `Admin UI build` red is closed by this branch — that part is done and correct.

The blocker is the IT failure, which belongs to whoever last touched engagement's REST path, not to this manifest.

JiRaska:
Two of the four failures are pre-existing, not caused by this diff — worth separating before deciding what blocks merge.

**`SurfaceRestContractIT` (4 assertion failures) is unrelated to this PR.** This PR touches only `discovery.ts`, `governance.yaml`, and `admin-ui.yaml` — no engagement-service source or test. The failing test class exists on `main` already; this diff cannot have caused it. Worth confirming it's not already red on `main` independently of this PR (I didn't check that directly, only that this diff doesn't touch it).

**`gates (gitops)` — "service runbook drift" — is a real, second-order consequence of the fix, not a false positive:**
```
FAIL [gitops] service runbook drift (issue #2255, enforced)
runbooks: 44 written, 0 kept (existing)
diff --git a/docs/runbooks/svc-engagement.md
```
Adding `governance.yaml` makes engagement-service a discovered module for the first time, and the runbook generator now expects `docs/runbooks/svc-engagement.md` to exist and match. It doesn't yet. Same shape as the RoleBinding/`OPENBANK_NAMESPACES` pair this PR already handles — the manifest becomes real and every gate reading it now has an opinion. Generating that runbook (presumably `node scripts/gen-runbooks.mjs` or equivalent) looks like the fix, and belongs in this PR since it's a direct consequence of what's already here.

Verified independently:
- `check-gitops-duplicate-resources.py --enforce` on this branch: `OK — 641 manifests, no duplicate resource or list entry` — the RoleBinding does not collide with anything.
- `node scripts/generate-governance.mjs` on this branch: `58 modules ... 0 gaps` — the governance drift itself is fixed.

Not merging myself — `build (openbank-engagement-service)` is a required check and it's genuinely red, even if not because of this diff.

JiRaska:
**Diagnosis of the two red gates, reproduced on this branch's head (`6c73f8d5b`) in a detached worktree — I have not touched the branch.**

## 1. `service-runbook-drift` — the runbook is DERIVED from `governance.yaml`

`docs/runbooks/svc-engagement.md` is generated, and adding the manifest changes what it generates. The committed file still carries the em-dash placeholders from when there was no manifest to read:

```
-> Operational runbook for the `engagement` service. Data domain **—**,
-> classification **—**, datastore **—**.
+> Operational runbook for the `engagement` service. Data domain **platform**,
+> classification **confidential**, datastore **PostgreSQL**.
```

Fix is one command plus a commit:

```bash
python3 openbank-infra/scripts/generate-service-runbooks.py --force
```

Never hand-edit the file — the gate diffs it against a fresh generation, so a hand-edit that merely looks right fails just the same.

This is the same shape as the OPA bundles: **run the generator as the LAST step before pushing.** Regenerating and then making one more edit to `governance.yaml` leaves the runbook describing the superseded manifest, and "I ran it once and it was idempotent" does not cover that.

## 2. `Governance drift — 1 module(s) missing or with an invalid governance.yaml`

Still firing on this branch, so the added manifest is either invalid against the schema or a second module is short. Worth reading the checker's own output rather than assuming it is engagement-service twice — the message counts modules, it does not name one, and a PR that adds the missing manifest is exactly where "one module" quietly becomes a *different* module.

## Why this PR matters beyond itself

`build (openbank-engagement-service)` is red on `main`, so this failure is inherited by **every** open PR that builds that service — #3880 is red for this reason and touches no engagement file at all. Unblocking this one clears the others.

Not touching the branch because `/private/tmp/ob-enggov-11472` is checked out on it with uncommitted work in `application.yaml` and `SurfaceRestContractIT.kt`, and is 1 commit behind the remote. Pushing there would collide with a live session.

JiRaska:
The four `SurfaceRestContractIT` failures blocking this PR are fixed in **#4128** — they were never yours. Two layered defects in engagement's own test setup (401 from `@RolesAllowed` with no identity, then 503 from `@Authorize` with no PDP), both measured with a control run. Once #4128 lands, rebase or merge `main` here and this should go green; the `governance.yaml` half is already verified (`withGaps = 0`).

JiRaska:
**Pushed the fix for both red gates (`d8704da69`, signed). Nothing of yours was touched — additive commits only, no rebase, no force.**

## `service-runbook-drift` — fixed

`docs/runbooks/svc-engagement.md` is **generated** from the service's `governance.yaml`, so adding the manifest changes what the generator produces. The committed file still carried the placeholders from when there was no manifest to read:

```
-> Data domain **—**, classification **—**, datastore **—**.
+> Data domain **platform**, classification **confidential**, datastore **PostgreSQL**.
```

Regenerated with `generate-service-runbooks.py --force`, never hand-edited — the gate regenerates and diffs, so a hand-edit that merely looks right fails identically. Gate now PASS locally through `run-gates.py`.

## `Governance manifest gate` — already gone, no change needed

Re-measured rather than assumed: with this branch merged against **current** `main`, `node scripts/generate-governance.mjs` reports `withGaps: 0` and an empty `gaps` array, and it rewrites `governance.json` to content byte-identical to what is committed. The earlier red was against an older `main`.

Worth stating explicitly because that error message **counts modules without naming one** — and a PR that adds the missing manifest is exactly where "1 module" can quietly become a *different* module. It did not; the count is now zero.

Reproducing it locally needs `npm ci` in `openbank-admin-ui` first: without it the generator dies on `Cannot find package 'yaml'`, and the follow-up `require('./governance.json')` then fails with `MODULE_NOT_FOUND` — which reads like a missing artifact rather than a missing dependency.

## Why this was worth taking over

`build (openbank-engagement-service)` is red on `main`, so this failure is inherited by **every** open PR that builds the service. #3880 has been red for it while touching no engagement file at all. This PR was the fix, and it was itself blocked.

I only moved after checking: `/private/tmp/ob-enggov-11472` had been dormant for over an hour, was still a commit behind the remote, and its uncommitted `application.yaml` / `SurfaceRestContractIT.kt` edits are untouched by an additive push — they will merge or rebase cleanly on top.

JiRaska:
**Unblocked from CONFLICTING and the engagement build now passes locally (`0f6caf38d`, `d30c64249`, both signed).**

## Two things the merge did

**1. `application.yaml` — a real conflict, but only prose.** Both sides carried a different comment explaining the *same* `authz.enforce` setting. Took `main`'s wording, since it landed and is canonical, so the branch does not carry a second version to reconcile again later. The setting itself is untouched and the file still parses.

**2. `SurfaceRestContractIT.kt` — no conflict reported, and that is the interesting one.** Both sides *added* a `@TestSecurity` to the same class a few lines apart, so the three-way merge kept **both** and printed nothing:

```
e: SurfaceRestContractIT.kt:39:1 This annotation is not repeatable.
```

`main`'s `edge@openbank.test` at line 32, this branch's `surface-reader@openbank.test` at 39. Identical role set (`ROLE_OPERATOR`); only the user string differed, and nothing asserts on it. Kept main's, kept the comment explaining why the annotation is needed at all.

This is the failure mode the repo already documents — a merge git calls clean producing a state neither side wrote. Here it fails loudly at compile time. The dangerous version of the same thing does not: a dropped list entry just quietly is not there.

## Verified, not assumed

| check | result |
|---|---|
| conflict markers | 0; YAML parses |
| `@TestSecurity` count | 1 |
| `:openbank-engagement-service:test` | **24 tests, 0 failures, 0 errors, 0 skipped** — from the 6 JUnit XML files, not the console |
| `ktlintCheck` incl. `ktlintTestSourceSetCheck` | in the executed task list |
| `detekt` | in the executed task list |

One ktlint *"Annotations should occur immediately before the annotated construct"* surfaced from the deletion and was fixed with `ktlintFormat`, not by hand.

Those 24 are the same tests that report **4 failures on pristine `origin/main`**, reproduced earlier straight from `origin/main` in a detached worktree.

## Note on ownership

I moved because the PR was `CONFLICTING` — no merge ref, so it could not get checks at all — with a clean tree and no writes for 20 minutes. Additive commits only, no rebase, no force, so nothing in `/private/tmp/ob-enggov-11472` is lost; it just ends up behind.

PR #4117, opened 2026-08-07
ci: gate the missing-parameter 500 class (#3104)

JiRaska:
## What

Adds an **enforced** gate, `required-param-nullability`, that closes the missing-parameter-500 class from #3104 — plus a correction to the number that issue carries.

## The defect

JAX-RS injects `null` for an absent `@QueryParam`/`@HeaderParam`. Kotlin's null-safety is compile-time only, so a non-nullable declared type buys nothing at runtime — it only decides *where* the failure lands. On a **non-suspend** function the compiler emits `Intrinsics.checkNotNullParameter` at bytecode offset 0, so the NPE is thrown before the first statement of the body. No guard inside the method can ever run, `GenericExceptionMapper` renders 500, and the caller cannot tell "I sent a bad request" from "the server is broken" — which on a money path decides whether they retry.

Nothing in this repo could see the class. It compiles, the handlers have passing tests (every one of them supplies the parameter), and the OpenAPI spec says `required: true` whichever way the Kotlin type reads.

## Why a gate and not a sweep

Two exclusions carry the whole design:

- a `suspend fun` compiles to a different shape and carries **no** intrinsic at offset 0, so an identical-looking signature is not this defect there — 5 in the fleet;
- `@DefaultValue` is substituted by JAX-RS, so the parameter is never null however the type reads — 76 parameters.

A mechanical "make every such parameter nullable" pass would touch **130 sites to fix 49**, changing money-path behaviour for nothing. And 49 handlers across 20+ services in one PR is unreviewable; the gate is the part that closes the class, after which handlers can fall per service with no regression risk.

## Correction to the published number

#3104 and #3624 carry **67**. That figure was measured before the `@DefaultValue` exclusion existed and is inflated by parameters that can never be null. The true count is **49 handlers across 26 services**. The gate is the authority from here, because it is the only form of the count that gets re-derived.

## Falsification

Both directions, on the real tree, exit code read **directly** rather than through a pipe (a pipe reports the exit code of `tail`):

| case | result |
|---|---|
| control | `rc=0` — 49 baselined, 0 over, 0 stale |
| one injected handler in an unbaselined service | `rc=1`, one `::error::` naming it |
| one baseline entry inflated 1 to 9 | `rc=1`, one `::error::` naming the stale entry |

`--self-test` covers 12 cases including the #3104 worked example exactly as it read before `09f32b673` fixed it, its merged fix, the suspend twin, `@DefaultValue`, a Kotlin default, `@HeaderParam`, the `@RestQuery` shorthand (zero uses today, not zero forever), `@PathParam` being out of scope, a KDoc showing the bad shape, and a default value containing a paren. The `@DefaultValue` case **failed** on first run and is why that exclusion exists.

## Baseline shape

Per service, as **counts not names**: a partial fix lowers a service's entry instead of churning a list, and an entry only disappears at zero. A stale entry fails too, so the baseline cannot rot into permanent furniture.

## Fix shape for the 49

Already merged as the fleet precedent in `09f32b673` (#3625):

```kotlin
-  @QueryParam("productId") productId: String,
+  @QueryParam("productId") productId: String?,
+  requireNotNull(productId) { "query parameter 'productId' is required" }
```

`IllegalArgumentExceptionMapper` in `openbank-libs-runtime` already renders that as 400 — nothing new is needed downstream. A shared `ExceptionMapper<NullPointerException>` was evaluated and rejected on #3104: it would have to string-match a compiler intrinsic's message to avoid swallowing genuine null-derefs as "bad request".

Refs #3104


JiRaska:
Ran the self-test and the gate itself rather than reading the description.

```
--self-test    9/9 pass, including the two that would be easy to get wrong:
                 "a default value containing a paren does not truncate the parameter list"
                 "one function is reported once even with two offending params"
against main   49 affected handlers across 26 services; 49 baselined, 0 over, 0 stale
```

The `@DefaultValue` exclusion is the part worth checking independently, since it's what separates this from a naive "flag every non-nullable JAX-RS param" sweep: JAX-RS substitutes the default before the handler ever sees `null`, so a parameter carrying one literally cannot reproduce #3104's failure regardless of its declared nullability. Excluding it is not weakening the gate, it's the difference between 49 real sites and the 130-site sweep the PR says a mechanical pass would have touched.

The corrected count (49, not #3104/#3624's 67) is stated as a correction rather than silently changed, with the reason given (the old figure predates the `@DefaultValue` exclusion) — the right way to handle a number this repo has already cited twice elsewhere.

I did not construct a live known-positive by editing a real handler — the self-test's 9 cases already exercise the boundary conditions a hand-edit would probe, and editing production REST code on someone else's branch to test a gate has a worse cost/benefit than trusting a self-test this thorough.

JiRaska:
Reviewed. **This duplicates a gate already on `main`, and the new one is the weaker of the two.** Recommending close-as-duplicate, or folding its two genuinely new ideas into the incumbent.

`origin/main` already carries `nonnull-jaxrs-param-ratchet` — `gates.yaml:1423`, `group: kotlin`, `mode: enforced`, backed by `.github/scripts/check-nonnull-jaxrs-params.py` (377 lines). Same issue #3104, same fix commit `09f32b673`. With this branch checked out, `run-gates.py --list` shows **both** in `group: kotlin`.

That is the #2984/#3009 pattern: CI cannot see a duplicate rule (both are green), the PR body does not mention the incumbent, and merging puts two rules for one defect into every PR — to drift apart later.

**Measured comparison**, running both scripts over the same tree:

| | incumbent (`main`) | #4117 |
|---|---|---|
| outbound `@RegisterRestClient` interfaces | excluded (`interface` ⇒ skip) | **counted** — 29 of the 49 it reports are outbound client methods (`AmlServiceClient.createCase`, `TransactionServiceClient.initiateTransaction`, …), where non-nullable is *correct*: the argument comes from compile-time-checked Kotlin, not from JAX-RS |
| JVM primitives | excluded (JAX-RS supplies 0) | `@QueryParam("a") a: Int` flagged |
| `@MatrixParam` | covered | not matched |
| `@DefaultValue` before the annotation | handled | `@DefaultValue("10") @QueryParam("size") size: Int` flagged though safe (0 occurrences today, so latent — and it fails noisily, not silently) |
| baseline granularity | per `service\|Class\|param` | per-service **count** — a fixed handler replaced by a new offending one in the same service is invisible |
| `suspend` handlers | flagged (the null flows into the body, since a suspend fun gets no intrinsic) | excluded |

So the claim to be "the authority on the number" (49) does not survive: the incumbent's own header already documents that outbound clients are *"52 of the 115 raw matches… which is why a naive grep over-counts by ~2x"*, and this re-introduces exactly that inflation. The `suspend` exclusion is the one I would worry about most — that is the case where the null reaches the body rather than throwing at the boundary.

**Mechanically the new checker is well built**, which is why this is worth folding rather than discarding: self-test 12/12; `strip_comments` is nesting-aware (Kotlin block comments nest, so a naive stripper closes early), and I confirmed a KDoc, a string literal, a `@BeanParam` class property and a multi-line signature all behave. The ratchet is falsified both ways on the real tree — an injected handler in unbaselined `openbank-account-service` gives `::error::openbank-account-service: 1 handler(s)… (baseline 0)`, exit 1; nullifying `StatementResource` gives `baselined at 5 … but now has 2. Lower or remove its BASELINE entry`, exit 1. So it can shrink and cannot rot. No `--enforce` flag, but `mode: enforced` plus an unconditional exit 1 is consistent — not a gap.

**What is genuinely new here and worth keeping:** matching the `@RestQuery` / `@RestHeader` shorthand, which the incumbent lacks. That is a real coverage gap in `check-nonnull-jaxrs-params.py` and a small addition to it.

JiRaska:
Closing as a duplicate of the gate already on `main`.

**Verified the survivor before closing**, not just its existence: `nonnull-jaxrs-param-ratchet` is declared in `gates.yaml:1423` (`group: kotlin`, `mode: enforced`) backed by `.github/scripts/check-nonnull-jaxrs-params.py`, and it **runs and passes on current `main`** — `run-gates.py --only nonnull-jaxrs-param-ratchet` exits 0. Same issue #3104, same fix commit `09f32b673`. So the defect class stays gated after this closes; nothing is lost by closing it.

**Why this one rather than that one**, measured by running both scripts over the same tree:

| | incumbent (`main`) | this PR |
|---|---|---|
| outbound `@RegisterRestClient` interfaces | excluded | **counted** — 29 of the 49 reported are outbound client methods (`AmlServiceClient.createCase`, `TransactionServiceClient.initiateTransaction`, …), where non-nullable is *correct*: the argument comes from compile-time-checked Kotlin, not from JAX-RS |
| `suspend` handlers | flagged | excluded |
| JVM primitives | excluded (JAX-RS supplies 0) | flagged |
| `@MatrixParam` | covered | not matched |
| baseline granularity | per `service\|Class\|param` | per-service **count** |

The `suspend` exclusion is the one that decides it. A suspend handler gets no `Intrinsics.checkNotNullParameter` at all, so the null flows **into the body** instead of throwing at the method boundary — that is the harder half of #3104, and it is the half this checker skips. And the claim to be "the authority on the number" (49) does not hold: the incumbent's header already documents outbound clients as *"52 of the 115 raw matches… which is why a naive grep over-counts by ~2x"*.

Merging both would have put two rules for one defect into every PR, to drift apart later — CI cannot see that, since both are green. Same shape as #3009 duplicating #2984.

**Not a criticism of the work.** The checker is well built and I said so in review: 12/12 self-test, a nesting-aware `strip_comments` (Kotlin block comments nest, so a naive stripper closes early on a KDoc containing `/*`), and a ratchet falsified in both directions on the real tree — an injected handler in unbaselined `openbank-account-service` reddens it, and nullifying `StatementResource` reports the baseline as stale rather than letting it rot.

**One thing here is genuinely new and worth keeping**: matching the `@RestQuery` / `@RestHeader` shorthand, which `check-nonnull-jaxrs-params.py` does not cover. That is a real gap in the incumbent and a small addition to it — worth a follow-up PR against that script rather than a second gate.

Issue #3082, opened 2026-08-01
auto-deploy: a fleet-sized concurrency group silently discards the whole build, and can-i-deploy races the pact publish


JiRaska:
### Summary

Two distinct defects in `auto-deploy.yml`, both measured. The second one silently **throws away a whole fleet build** and has fired three times.

---

## Defect 1 — `can-i-deploy` races the workflow that publishes the pacts

`can-i-deploy` has `needs: [changes, build-push]`. That is auto-deploy's *own* image build — **not** `services-ci`, which is the workflow that publishes pact versions. Both are triggered by the same push and run concurrently, so `can-i-deploy` asks the broker about a commit whose pacts are still being built.

Measured on run `30690770972` (sha `000ba2a5`):

```
can-i-deploy asked about account-service   08:36:49
account-service build STARTED              08:39:03   (+2m14s)
account-service build completed success    08:43:44
```

It asked **2m14s before the build even started**. The pact could not have existed. Not a contract break — a race.

`services-ci` builds strictly sequentially (~1 min between job starts, one runner pool), so with a large changed-service set the race is lost by construction for everything after the first few.

The broker says so itself in the log:

> `WARN: It is recommended to specify the version number (rather than the tag or branch) of the pacticipant you wish to deploy to avoid race conditions.`

**This half is largely working as designed** — `PENDING_BUILD` is classified as self-clearing and the 3-hourly reconcile re-drives it. The two scheduled failures (`342cf7d7`, `f994c492`) went red only because **money-path** services were among the blocked, which the code deliberately keeps loud. Their `gitops-pr` still ran and the deployable subset still deployed.

---

## Defect 2 — the concurrency group interpolates the entire service list, and a fleet-sized one prevents the job from ever being created

```yaml
concurrency:
  group: >-
    auto-deploy-gitops-pr-${{ github.ref }}-${{
      github.event_name == 'workflow_dispatch' && github.run_id || needs.changes.outputs.services
    }}
```

`needs.changes.outputs.services` is the JSON array of changed services. For a change to `openbank-libs-*` that is **all 53 services**, giving a group name of **~1436 characters**.

When it is that long, the `Gitops PR` job is **never instantiated at all** — no job, and no check-run on the commit. Its `if` was provably true (`build-push.result == 'success'`, `can-i-deploy.result` was `failure`, not `cancelled`), so the `if` is not the explanation.

Correlation over the last 20 auto-deploy runs is perfect:

[...]

Issue #1348, opened 2026-07-16
can-i-deploy failing before it can decide makes record-deployment record ALL services as deployed — 40 phantom broker records are blocking a money-path fix (sepa-payment #844, undeployed 4 days)


JiRaska:
## What

When `can-i-deploy` fails **before it can emit a verdict** — a flaky pact-CLI download, not a contract decision — `record-deployment` treats that as *"no filter, record everything"* and tells the Pact broker that **every changed service is deployed to sandbox**. If the gitops PR from that run then never merges, those records are permanent lies. `can-i-deploy` on later runs compares healthy services against those phantom versions, finds no verified pact, and blocks them.

A money-path fix has been undeployed for 4 days because of this.

## The chain, with evidence

Run [`29206486046`](https://github.com/JiRaska/open-bank-oss/actions/runs/29206486046) (2026-07-12, `dae215b5b`):

1. **`can-i-deploy` died on infrastructure, not a verdict:**
   ```
   [failure] could not download pact standalone (https://github.com/pact-foundation/pact-ruby-standalone/rel…)
   [failure] Process completed with exit code 1.
   ```
   The step exits before the per-service loop, so it never writes `deployable` to `$GITHUB_OUTPUT`.

2. **The empty output is read as "no filter".** `gitops-pr` does:
   ```yaml
   DEPLOYABLE_SERVICES: ${{ needs.can-i-deploy.outputs.deployable || '' }}
   ```
   and `Record sandbox deployment in Pact broker` does:
   ```bash
   if [ -n "${DEPLOYABLE_SERVICES}" ]; then   # false → no filter → record ALL changed services
   ```
   `''` is meant to mean *"can-i-deploy didn't run at all (no PACT_BROKER_URL) — no filter, matching pre-fix behavior"*. It cannot tell that apart from *"can-i-deploy ran and crashed before deciding anything"*.

3. **40 services were recorded as deployed** from that one run:
   ```
   Recorded deployment of openbank-account-service version dae215b5bf99… to sandbox environment in the Pact Broker.
   Recorded deployment of openbank-ledger-service  version dae215b5bf99… to sandbox environment in the Pact Broker.
   Recorded deployment of openbank-fraud-service   version dae215b5bf99… to sandbox environment in the Pact Broker.
   …40 total
   ```

[...]

Issue #846, opened 2026-07-11
can-i-deploy blocked: account-service↔balance-service pact verification stuck as FAILED (verification-results/389), blocking ALL account-service deploys


JiRaska:
### Affected service / module
openbank-account-service (consumer) / openbank-balance-service (provider)

### Version / commit
main @ 144445677944189139752e313db68f132ed10d8e (PR #835)

### What happened?
`can-i-deploy` in `auto-deploy.yml` fails for `openbank-account-service` on **every** recent
deploy attempt, both `workflow_dispatch` (run 29162819832) and a genuine `push`-triggered run
right after PR #835 merged (run 29164508317, headSha `144445677944189139752e313db68f132ed10d8e`).

Both runs report the exact same broker record:

```
https://pact.open-bank.tech/pacts/provider/openbank-balance-service/consumer/openbank-account-service/pact-version/7f399dbc1e3d8ff7493582b0571981c3c2ce198b/metadata/.../verification-results/389 (failure)

The verification for the pact between the latest version of openbank-account-service with tag
main (d3a0fc904c301820fc5cf05bb8b4a89e528f9e6f) and the version of openbank-balance-service
currently in sandbox (4ea273195d038704acc6341f684c0f1cb039ce82) failed
```

Two anomalies:
1. `verification-results/389` is a genuinely FAILED result (not "missing") — much lower ID than
   sibling pairs checked in the same run (party-service/transaction-service verifications were
   in the 2100s-2200s), suggesting it is old/stale.
2. Even after PR #835 (a real account-service source change) merged and its push-triggered CI
   ran, the pact broker's "latest version of openbank-account-service with tag main" still
   resolves to the **older** commit `d3a0fc9...` (PR #832, an unrelated admin-ui deploy chore),
   not the new tip `14444567...`. The `main` tag does not appear to be advancing for
   account-service's consumer pact, or balance-service is not re-verifying against it.

### What did you expect?
A push to `main` touching account-service should publish a fresh consumer pact tagged `main`
at the new SHA, balance-service's provider verification should re-run against it, and
`can-i-deploy` should reflect a current, non-stale result (pass or genuine fail on real
contract content, not a frozen historical failure).

### Reproduction steps
1. `gh workflow run auto-deploy.yml -f services="openbank-account-service"` (or just merge any
   account-service PR to `main`).
2. Wait for the `can-i-deploy` job.
3. Observe it fails on the account-service↔balance-service pair, always citing
[...]

PR #4120, opened 2026-08-07
ci: extract the can-i-deploy gate out of auto-deploy.yml (#4086)

### LINKED ISSUE #4086: auto-deploy.yml's can-i-deploy step is 45 characters from the run-step ceiling, and crossing it makes the whole workflow unparseable
## The measurement

`auto-deploy.yml`'s `can-i-deploy` job has one step whose `run:` script is **18955 characters** on `origin/main`. The `workflow-run-step-size` gate's limit is 19000. That is **45 characters of headroom** on the fleet's deploy path.

```
job=can-i-deploy  step='can-i-deploy (per changed service)'  chars=18955
```

Parsed out of the YAML rather than counted from a diff, so it is the real script length after block-scalar folding.

## Why 45 characters is a problem and not a curiosity

The 19000 limit is a guard band in front of a hard GitHub behaviour, not a style rule. Above roughly 20054 characters GitHub stops parsing the **entire workflow file**: zero jobs are created, `name:` is never read, no error is surfaced, and the run renders as an ordinary red. It kills the file for every contributor, not just the author — the state #3135 shipped and #3139 reverted.

So the consequence of the current margin is that a **single added comment line** on that step is enough to fail the gate, and a larger addition made while the gate is somehow not consulted takes the deploy workflow down for everyone.

## How it is being felt right now

#3992 (`ci(auto-deploy): guard manual dispatch and reconcile`) adds 194 characters to that step and is blocked at 19149. Nothing is wrong with that PR — it is a normal-sized change landing on a step with no room. Any future PR touching the step meets the same wall, and each will look like an individual lint failure rather than the shared cause.

## The fix the gate itself prescribes

> Move the logic into `.github/scripts/` and call it; prose belongs in the script's header, not in the workflow.

Extracting the step into `.github/scripts/can-i-deploy.sh` retires the cliff instead of moving it, and the prose (which is a large share of those 18955 characters, and worth keeping) moves to a script header where it costs nothing. It touches the deploy path, so it wants a deliberate PR and a human — filing this rather than doing it unannounced.

Worth checking the other long steps at the same time: nothing else in `auto-deploy.yml` exceeded 17000, but the same shape can grow anywhere, and the gate only reports the largest.

JiRaska:
## What

Moves the `can-i-deploy (per changed service)` script out of `auto-deploy.yml` into `.github/scripts/can-i-deploy-gate.sh`. The step falls from **18955 characters to 245**.

## Why this is urgent rather than tidy

`check-workflow-run-step-size.py` enforces 19000. The step was **45 characters** under it — one comment line.

Crossing that ceiling does not produce a size error. The **whole workflow file stops being parseable**: `name:` is never read, every push yields a run with ZERO jobs titled after the file path, and it reads as an ordinary red run. It kills the workflow for everyone, not just the author — #3135 blocked every contributor's deploy until it was reverted (#3139). The boundary was found by bisecting real pushes against GitHub: 20054 accepted, 20654 rejected.

Nothing local sees this class. PyYAML, a strict duplicate-key loader, `actionlint` and `yamllint` were all clean on the very file GitHub refused. So the existing gate can only report the catastrophe after it happens; it cannot create room. The next person to explain a decision in place would have taken the deploy pipeline down for the entire repo.

## What did NOT change

The prose. A file has no ceiling, which is the whole point of moving it, and the house rule already says prose belongs in `.github/scripts/*` headers rather than a `run:` block. Nothing was compressed out — only its address changed.

One comment did change, deliberately: the line telling the reader to *"Keep this comment SHORT: an oversized `run:` makes the whole workflow unparseable"* is obsolete here and now says so.

## Fidelity, proven rather than asserted

The extracted body is **305 lines against the inlined 305**. Diffing it against `origin/main`'s step with only the three `${{ }}` substitutions applied yields **exactly one hunk** — the obsolete comment above. Byte-identical everywhere else.

## The three interpolations

`${{ needs.changes.outputs.services }}`, `${{ github.event_name }}` and `${{ github.event.inputs.codeploy }}` move to the step's `env:`. An expression is substituted by the runner *before* the shell sees it, so it cannot survive extraction, and one left behind would have silently become a literal string. This is also the recommended shape for workflow inputs generally.

`SERVICES` goes to `env:` and nowhere near a `concurrency.group:` — interpolating a fleet-sized list into a group name silently stops the job being created at all (#3082/#3084). An env var has no such ceiling.

## Validation

| check | result |
|---|---|
| `bash -n` on the script | clean |
| YAML parse | all five jobs intact (`changes`, `build-push`, `can-i-deploy`, `gitops-pr`, `deploy-signal`) |
| `check-workflow-run-step-size.py` | passes; largest remaining `run:` step in the file is 10359 |
| `actionlint` | finding **set** byte-identical to `origin/main` — 1 `runner-label`, 1 `shellcheck`, both pre-existing and outside this step |

The actionlint check is a set diff on purpose. Post-filtering a linter's output before checking it ran is how a syntax error gets reported as "no new findings" in this repo's history, so branch and `main` were each linted and the counts compared.

Closes #4086


JiRaska:
Reviewed against #4086, which I filed. **Clean — approve.** This closes it properly rather than buying headroom.

**It shrank, measured by parsing the YAML rather than reading the diff.** `run:` scripts over 5000 chars in `auto-deploy.yml`:

| | `changes` | `can-i-deploy (per changed service)` |
|---|---|---|
| `origin/main` | 10359 | **18955** |
| this branch | 10359 | *gone* |

So the step that sat 45 characters under the 19000 limit — in front of a ~20054 cliff where GitHub stops parsing the **whole file** and creates zero jobs with no error — is no longer a step. `check-workflow-run-step-size.py` on a merged worktree: exit 0, largest script in the repo now 14568 and elsewhere.

**Behaviour preserved, verified byte-level rather than asserted.** I dumped `main`'s step body via PyYAML, applied only the three documented substitutions (`needs.changes.outputs.services` → `${SERVICES}`, `github.event_name` → `${EVENT_NAME}`, `github.event.inputs.codeploy` → `${INPUT_CODEPLOY}`, leaving zero `${{ }}`), and diffed against the extracted script. Three hunks: the new file header (prose), the now-obsolete "keep this comment SHORT" note, and **one real code change**.

That one change is `SERVICES='<interpolated>'` → `SERVICES="${SERVICES:?}"` (`can-i-deploy-gate.sh:48`). It is a hardening, and worth naming because it is the only semantic delta: previously an empty list gave a benign no-op run, now `:?` aborts non-zero. Two things make it safe, both checked:

- the job carries `if: needs.changes.outputs.any == 'true'`, so `SERVICES` is non-empty whenever it runs;
- `gate_ran=true` is written at line **47**, *before* the `:?` at line 48 — so even if it did fire, the #1348 "gate never ran" vs "ran and failed" signal survives, and `gitops-pr` still fails closed rather than open.

**Everything the deployable-subset semantics depend on is verbatim:** `set -uo pipefail` with `-e` deliberately absent (that is what lets the per-service loop continue), the `rc` aggregation, `deployable_list` / `deployable=` / `blocked=` written to `$GITHUB_OUTPUT` **before** the final `exit $rc`, the schedule-tick `exit 0`, and the ADR-0092 money-path carve-out. So `if: always() && … != 'cancelled'` still deploys the subset that passed when part of the fleet is blocked (#846).

**`concurrency.group` not reintroduced.** `SERVICES` goes into the step's `env:` only; `gitops-pr` still keys on the bounded `services_key` digest, so the #3082 failure — a group name interpolating a 1436-character service list, after which the job is never instantiated at all — stays fixed. Diffing every `concurrency` block against `main` shows line-number shifts and nothing else.

Worth a one-line note to the author about the `:?`, not a block.

PR #4121, opened 2026-08-07
fix(ci): pin the kafka dotted-key baseline to explicit channels

### LINKED ISSUE #3928: check-kafka-dotted-keys baseline uses a channel wildcard, so new channels are pre-absorbed — account-service's delegation consumer inherited the defect silently
## Summary

`check-kafka-dotted-keys.py` baselines six services with a **channel wildcard**, so a brand-new
channel on any of them is pre-absorbed and the gate structurally cannot flag it:

```python
# .github/scripts/check-kafka-dotted-keys.py:65-70
("openbank-account-service",   "*", "auto.offset.reset"): "#2945 — declared earliest, effective default; …"
("openbank-aml-service",       "*", "auto.offset.reset"): "#2945 — same"
("openbank-balance-service",   "*", "auto.offset.reset"): "#2945 — same"
("openbank-document-service",  "*", "auto.offset.reset"): "#2945 — same"
("openbank-party-service",     "*", "auto.offset.reset"): "#2945 — same"
("openbank-statement-service", "*", "auto.offset.reset"): "#2945 — same"
```

The `"*"` was presumably meant as "the channels that exist today". It reads as "every channel this
service will ever have".

## It has already happened

account-service's **new delegation channel** sets `auto.offset.reset: earliest` as a YAML key — the
documented SmallRye trap, where a leaf key containing a literal dot is quoted into
`…<channel>."auto.offset.reset"` and the connector silently uses its own default. The wildcard
absorbed it on arrival, so the gate said nothing.

Concretely: `#2990`'s acceptance criterion 3 (replay rebuilds the projection from the top) **does not
hold in the deployed path**, and nothing reported that. card-issuance got the same channel right, via
an msg-override ConfigMap; account-service has no msg-override manifest at all.

## Why this is worth fixing rather than tolerating

This is the failure shape the repo already knows in two other places — a gate whose scope silently
excludes the new thing, so its green is about work it never did (`pact-drift-check`'s hand-kept
module list, `domain-purity-gate`'s `*/domain/*` path segment). A baseline is supposed to freeze
*today's* debt; a wildcard baseline freezes an unbounded future.

**Suggested:** enumerate the six services' channels explicitly, so the entries are
`(service, channel, key)`. Then a new channel is not covered, the gate fires, and the author has to
choose msg-override or an explicit baseline addition with a reason — which is what the baseline
mechanism is for.

Bonus check worth adding while there: the script already reports a baseline entry that has become
covered (good). It should also report a baselined `(service, channel)` that **no longer exists** —
otherwise the list rots in the other direction.

## Not verified

- I did not re-audit the other five wildcard services for channels added since the baseline was
  written; account-service is the one instance I have evidence for.
- I did not measure whether forcing `earliest` on the delegation consumer would replay anything
  material — the replay-risk reasoning in the original `#2945` entries may still make the *fix*
  operational rather than a config edit. That is the decision the gate should surface, not skip.
- Found while triaging #2990; not reproduced against a live broker.

Refs #2945, #2990, #686

JiRaska:
## What

`check-kafka-dotted-keys.py`'s `BASELINE` carried six `("openbank-*-service", "*", "auto.offset.reset")` entries. A wildcard **channel** makes the exclusion a property of the *service*, so every channel that service gains afterwards is pre-absorbed — the ratchet prints `OK — no new ones` about a finding it never looked at, and the exclusion silently grows past what was ever justified.

That is not hypothetical. `openbank-account-service`'s `delegation-events-in` arrived on 2026-08-02 (#3058), one day after the baseline commit (#2969), and inherited the exemption with nobody deciding it.

This is the same repo rule as the pact-drift scope: never let a gate's coverage set be maintained separately from the artifacts it covers, and never let an exclusion be broader than what was actually justified.

## Changes (one file, `.github/scripts/check-kafka-dotted-keys.py`)

- **`BASELINE` names all 10 `(service, channel, key)` triples explicitly.** No `*`. `delegation-events-in` is pinned with its own distinct reason — it was never part of the #2945 decision, and its replay call is still open.
- **`baseline_key()` matches exactly.** The `(service, "*", key)` fallback is gone, so a new channel is a NEW finding, and the existing stale check now works per channel rather than per service (a baselined channel that is fixed or removed is reported).
- **`validate_baseline()` refuses a `*` channel (or an unwatched key) with exit 2**, so the shape cannot be reintroduced by hand.
- **`baseline_self_test()`** — the BASELINE branch had *no* self-test coverage at all, which is why #3928 was invisible to a gate whose self-test passed 5/5. Five new cases, each chosen because the old wildcard code fails it in the silent direction.

## Deliberately NOT in this PR: the underlying config

The defect this gate guards is that SmallRye's YAML source quotes a leaf key containing a literal dot, so `auto.offset.reset: earliest` registers as `…"auto.offset.reset"` and the connector never reads it — the effective value is the connector default, `latest`.

**Fixing that is an operational decision, not a config edit.** Six services look correct on `group.id` only by coincidence (their value happens to equal `quarkus.application.name`, which is what the fallback produces). For `auto.offset.reset` **no such coincidence exists**: making `earliest` actually take effect on a consumer group that has been running as `latest` re-reads the topic from the beginning — a mass replay on money-path channels. That needs a `*-msg-override` ConfigMap and a deliberate per-channel call, as card-issuance did.

**This PR narrows the gate's baseline only.** No `application.yaml` and no gitops manifest is touched. The point is to make the gate *surface* that decision for the next channel instead of absorbing it.

Note the one judgement call: `delegation-events-in` is pinned rather than left to go red. The gate is `mode: enforced` inside the required `Validate manifests` job, so shipping it red would block every PR in the repo on a change this PR is explicitly not making. It is now a named, reviewable baseline entry with its own issue reference — visible, where before it was absorbed.

## Falsification — both ways, on the real tree, exit codes read directly (no pipe)

Every run below is `cmd > file; echo $?` — never `| tail`, which reports `tail`'s status. The old script (`git show origin/main:…`) was run on the *same* mutated tree each time, so the counter-example provably reaches the code.

**Clean tree (control):** new gate `EXIT=0`, `10 baselined occurrence(s), no new ones`. Old gate: `EXIT=0`, `6 baselined occurrence(s)`.

**(a) New dotted key in a channel that is not baselined** — added `zz-falsify-in` with `auto.offset.reset: earliest` to account-service:

```
--- NEW gate (pinned baseline) ---            EXIT=1
NEW  openbank-account-service/.../application.yaml: channel 'zz-falsify-in' sets
     `auto.offset.reset: earliest` as a dotted YAML key with no *-msg-override ConfigMap.
1 new occurrence(s), 0 stale baseline entr(ies).

--- OLD gate (origin/main, wildcard baseline) ---  EXIT=0
kafka dotted keys: OK — 6 baselined occurrence(s), no new ones.
```

Counts: new gate 1 finding / 0 stale (expected 1/0); old gate 0 findings — this is the absorption in #3928, reproduced.

**(b) Baseline entry that no longer matches anything** — removed `auto.offset.reset` from party-service's `consent-events-in`:

```
--- NEW gate ---                              EXIT=1
STALE  baseline entry ('openbank-party-service', 'consent-events-in', 'auto.offset.reset')
       no longer occurs — it is either fixed or the service is gone.
0 new occurrence(s), 1 stale baseline entr(ies).

--- OLD gate ---                              EXIT=0
kafka dotted keys: OK — 6 baselined occurrence(s), no new ones.
```

Counts: new gate 0 findings / 1 stale (expected 0/1); old gate green, because the wildcard entry still matched the service's *other* channel. The reverse ratchet was as broken as the forward one.

Both mutations were reverted from a `cp` backup and the clean-tree run re-confirmed `EXIT=0` before committing (`git status --porcelain` shows only the script modified).

**(c) Shape guard** — reintroducing a `*` entry into a scratch copy of the fixed script: `EXIT=2`, `::error::baseline entry ('openbank-aml-service', '*', 'auto.offset.reset') uses a wildcard channel…`.

**(d) Through the real runner:**

```
python3 .github/scripts/run-gates.py --only kafka-dotted-key-ratchet    RUNNER_EXIT=0
PASS [data] kafka dotted-key ratchet (issues #686/#2945, enforced)  (7.9s)
self-test: 11 passed, 0 failed          (was 6; +5 baseline cases)
kafka dotted keys: OK — 10 baselined occurrence(s), no new ones.
```

## Out of scope, still open

The issue's later comment about `WATCHED_KEYS` being only two keys (anacredit's dotted `client.id` / `value.deserializer`, found by the advisory `dotted-mp-messaging-key-guard`) is a separate scope change to the same gate and is not touched here.

Closes #3928

### DIFF
--- a/.github/scripts/check-kafka-dotted-keys.py
+++ b/.github/scripts/check-kafka-dotted-keys.py
@@ -40,12 +40,17 @@
 #   green, and a NEW occurrence fails. A baseline entry that becomes covered is also reported, so the
 #   list cannot quietly rot in either direction.
 #
+#   That last property only holds if every baseline entry names ONE channel — see the BASELINE
+#   comment (#3928). A `*` channel makes the ratchet unable to see a channel added later, and the
+#   symptom is a confident `OK`, never a red.
+#
 # EXIT CODES
 #   0  no new occurrences, no stale baseline entries
 #   1  a dotted `group.id` with no override ConfigMap, a dotted `auto.offset.reset` asking for a
 #      non-default value with no override — or a baseline entry that is now covered and should be
 #      removed
-#   2  the check could not run (PyYAML missing, tree not found). Never conflated with 0.
+#   2  the check could not run (PyYAML missing, tree not found) or the BASELINE itself is malformed
+#      (a wildcard channel, an unwatched key). Never conflated with 0.
 #
 # Run:  python3 .github/scripts/check-kafka-dotted-keys.py [--root .] [--self-test]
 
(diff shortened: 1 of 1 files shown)


JiRaska:
Reviewed. **Clean — approve.** One file, `check-kafka-dotted-keys.py`; no `application.yaml`, no gitops manifest.

**The replay question first, since that is the dangerous half of this defect.** Forcing `auto.offset.reset: earliest` onto a consumer group currently running as the connector default is a mass replay on money-path channels. This change cannot cause it: the gate reads config and never writes it, and making a dotted key effective still requires a `*-msg-override` ConfigMap, which is not in this diff.

**What "pin to exactly" does:** `baseline_key()` drops the `(service, "*", key)` fallback and matches the exact `(service, channel, key)` triple, expanding the six wildcard entries into the 10 real channels. `validate_baseline()` rejects a `*` channel or an unwatched key with exit 2 and is called from `main()` before the scan, so the loose shape cannot be reintroduced by hand.

**It can shrink, and that is the point.** Stale detection existed but was per-*service* under the wildcard, so a service's other channel kept a fixed entry looking live. Per-channel, a fixed or removed channel now goes red as `STALE` — the baseline is a ratchet that can only be paid down deliberately, which is the property the repo's own lesson about hand-kept exemption lists asks for.

Ran on a merged worktree:

```
kafka dotted keys: OK — 10 baselined occurrence(s), no new ones.   exit 0
--self-test: 11 passed, 0 failed   (was 6; +5 baseline cases)      exit 0
```

The zero-stale result is what proves the 10 entries match real channels — inspection alone would not.

The judgement call — pinning `openbank-account-service/delegation-events-in` rather than shipping a red into the required, `mode: enforced` `Validate manifests` job — is reasoned in place and now carries its own reference, which is the right way round: the exemption is the thing a human has to justify.

Issue #4109, opened 2026-08-07
Track ADR-0248 implementation (statement and payment confirmation document templates)


JiRaska:
Tracking issue for implementing ADR-0248 (merged in #4104, commit b4d16ee6991c48bf637e50f9e5a7e01b30f82680): https://github.com/JiRaska/open-bank-oss/blob/main/docs/adr/0248-statement-and-payment-confirmation-document-templates.md

Replaces #4105 (closed once the decision merged). Work is being picked up directly now, one PR per item below.

## Backlog

- [ ] **billing-service**: annual fee-summary aggregation use case over `AssessedFee`/`BillingCycle` + new outbox event `billing.annual-fee-summary.ready`.
- [ ] **document-service**: 6 new seed rows in `DocumentTemplateSeed.kt` (`MESICNI_VYPIS_CS`/`_EN`, `ROCNI_VYPIS_POPLATKU_CS`/`_EN`, `POTVRZENI_O_PLATBE_CS`/`_EN`) with real Handlebars bodies.
- [ ] **document-service**: new Kafka consumer for `billing.annual-fee-summary.ready` (annual statement) — calls the non-persisting `preview` endpoint at send time, hands bytes to delivery, discards.
- [ ] **statement-service**: "download statement" REST action — on customer request, assembles data from `StatementModel`, calls document-service's non-persisting `preview` endpoint synchronously.
- [ ] **sepa-payment-service**: "download confirmation" REST action — on customer request, reads its own persisted payment record, calls the non-persisting `preview` endpoint.
- [ ] **domestic-payment-service**: same as above for domestic payments.
- [ ] **statement-service**: author `docs/threat-models/statement-service.md` (does not exist today).
- [ ] **document-service**: update `docs/threat-models/document-service.md` for the new consumer + new synchronous callers.
- [ ] Pact `@PactFolder` contract tests for the new consumer and each synchronous preview call.
- [ ] Legal/compliance review of the fee-category → PAD Annex II mapping (flagged as an open item in ADR-0248; not blocking the other items).

Note: `billing-service`, `sepa-payment-service` and `domestic-payment-service` are on the `money_path_services` list (`rules.yaml`) — those PRs need 2 human approvals + a threat-model diff per repo governance; they will not self-merge.

## Extension: MCP + AI-agent exposure (approved follow-on, 2026-08-07)

[...]

PR #4122, opened 2026-08-07
feat(document-service): consume annual fee-summary events and deliver statement documents

JiRaska:
## Summary

Adds document-service's async Kafka trigger for the annual statement of fees (ADR-0248 D-family:
this is the one template family that needs a push, not a synchronous preview call, since a PAD
Art. 5 annual statement is a duty the bank must send). Consumes `billing-service`'s billing-outbox
`AnnualFeeSummaryReady` event, renders the `ROCNI_VYPIS_POPLATKU_CS` template via the existing
NON-PERSISTING `POST /api/v1/documents/templates/preview` path, and hands the bytes to a phase-1
logging-only delivery stub — no email/postal channel exists yet.

## Type of change

- [ ] fix (bug fix)
- [ ] perf (performance)
- [ ] docs
- [ ] chore (build / tooling)
- [ ] breaking change

## Linked issues

Refs #4109, Refs ADR-0248

## Changes

- `AnnualFeeSummaryReadyConsumer` (new `@Incoming("billing-outbox-events-in")` Kafka consumer,
  topic `openbank.billing.billing.event`) — mirrors `AccountCreatedConsumer`'s poison-pill-safe
  shape exactly: unparseable payload or a missing required field is logged and skipped, a
  downstream failure is logged and acked (never crashes the consumer / wedges the partition).
- `AnnualStatementDeliveryService` (new use case) — resolves the PUBLISHED `ROCNI_VYPIS_POPLATKU_CS`
  template, enriches with `PartyLookupPort`/`AccountLookupPort` (same fail-open pattern as
  `OnboardingDocumentService`), renders via `DocumentTemplateUseCase.previewRender` (never
  `DocumentRenderUseCase.render` — no persistence, no `Document` row, no `document.generated`
  outbox event), and hands the bytes to `StatementDeliveryPort`. Idempotent per `(accountId, year)`
  via the shared `IdempotencyStore` (Redis — `quarkus.redis.hosts` was already configured for this
  service but nothing consumed it until now), 400-day TTL to comfortably outlive any realistic
  redelivery window for a once-a-year event.
- `StatementDeliveryPort` (new outbound port) + `LoggingStatementDeliveryAdapter` (phase-1 stub) —
  logs that a delivery *would* have happened; no real email/postal channel exists anywhere in this
  repo yet, so this never pretends to actually send anything (`TODO(ADR-0248)` marks the follow-up).
- `IdempotencyConfig` — per-service `IdempotencyStore` producer (mirrors `account-service`'s), the
  first consumer of this service's already-configured Redis.
- `application.yaml` — new `billing-outbox-events-in` incoming channel.
- `docs/threat-models/document-service.md` — new trust-boundary entries: the Kafka ingress above,
  and the new synchronous callers of the existing (unchanged) preview endpoint from
  `statement-service`, `sepa-payment-service` and `domestic-payment-service`.
- Locale: the event contract carries no language field, so this always renders the Czech variant —
  documented in `AnnualStatementDeliveryService`'s KDoc as a known limitation, not a silent choice.

## Trust-boundary change (`rules.yaml: trust_boundary_diff_change`)

This PR adds a **new inbound Kafka consumer** (`AnnualFeeSummaryReadyConsumer` on
`billing-outbox-events-in` / topic `openbank.billing.billing.event`) — a new ingress edge into
document-service. `docs/threat-models/document-service.md` is updated accordingly: new Trust
boundaries entries for the consumer and for the preview endpoint's new synchronous callers, plus a
note on the Spoofing row about the consumer having no per-message caller identity (mTLS-only at the
broker, same posture as the existing `AccountCreatedConsumer` ingress).

## Definition of Done

- [ ] Integration tests added/updated (if service boundary involved) — not added; poison-pill/ack
      behavior is covered by a mockk-based consumer test mirroring `AccountCreatedConsumerTest`
      (no `@QuarkusTest` CDI-wiring IT for the new `@Incoming` consumer in this PR).
- [ ] Contract tests updated (if public API changed) — no public REST API changed.
- [x] `./gradlew :openbank-document-service:ktlintCheck :openbank-document-service:detekt :openbank-document-service:test` passes.
- [x] No new lint warnings.
- [ ] OpenAPI spec regenerated (if REST API changed) — N/A, no REST API change.
- [ ] Flyway migration added + rollback note (if DB changed) — N/A, no DB change.
- [ ] Avro schema versioned backward-compatibly (if event changed) — N/A, this PR only *consumes*
      a new JSON event contract (documented in `AnnualFeeSummaryReadyCommand`'s KDoc); the contract
      itself is authoritative and shared with `billing-service`'s producer-side PR.
- [x] Docs updated (threat model).

## Security checklist

- [x] No secrets, tokens, passwords, or PII in code, config, tests, logs. `partyRef` is
      deliberately never logged verbatim (see `LoggingStatementDeliveryAdapter`'s KDoc) — same
      log-injection convention `OnboardingDocumentService` already follows.
- [x] New `@Suppress` annotations, justified inline: `AnnualFeeSummaryReadyConsumer` carries
      `@Suppress("TooGenericExceptionCaught")` (poison-pill safety over an untrusted payload, same
      justification as `AccountCreatedConsumer`) and `@Suppress("ReturnCount")` (an elvis-return
      per required field reads more clearly than one large boolean condition, which detekt's
      `ComplexCondition` rejects anyway).
- [ ] Auth / crypto / payment code changed? — no.
- [ ] PII path changed? — no (party/account enrichment reuses the existing onboarding lookup ports
      unchanged).
- [ ] Cardholder data path changed? — no.

## Compliance impact

- [x] PSD2 / SCA / RTS — this is the PAD (Payment Accounts Directive) Art. 5 annual statement of
      fees push duty (ADR-0248). No compensating controls needed: delivery is currently a
      logging-only phase-1 stub, so nothing is actually sent to a customer yet — this PR wires the
      render + hand-off path, not a live delivery channel.
- [ ] PCI DSS
- [ ] DORA
- [ ] GDPR
- [ ] AML / 5AMLD
- [ ] CNB reporting
- [ ] None

## Rollout / Rollback

- Deployment strategy: rolling (standard for this service).
- Rollback plan: revert the PR — the new consumer and delivery path are additive; no data/schema
  changes to roll back. Nothing consumes or acts on `billing-outbox-events-in` until this ships, so
  a rollback is a no-op for any in-flight event (it stays in Kafka to be reconsumed on the next
  deploy).
- Data migration reversible: N/A (no migration).

## Reviewer notes

- The event contract (topic + JSON shape) is authoritative and shared with a parallel PR
  implementing `billing-service`'s producer side — please flag any deviation.
- `StatementDeliveryPort`/`LoggingStatementDeliveryAdapter` is explicitly phase-1: it logs, it does
  not send. A real email/postal channel is a follow-up.
- Locale is hardcoded to `ROCNI_VYPIS_POPLATKU_CS` since the event contract has no language field;
  see the KDoc on `AnnualStatementDeliveryService` for the reasoning.


JiRaska:
`gates (data)` — `kafka dotted-key ratchet (issues #686/#2945, enforced)` is failing on the new `billing-outbox-events-in` channel in `application.yaml`:

```yaml
mp.messaging.incoming.billing-outbox-events-in:
  group.id: openbank-document-service
  auto.offset.reset: earliest
```

Both `group.id` and `auto.offset.reset` are written as dotted YAML keys. SmallRye Config's YAML source quotes any leaf map key containing a literal dot, so this registers as the property `…billing-outbox-events-in."group.id"` — quotes included — and the connector's plain `getOptionalValue("group.id", …)` never finds it.

- `group.id` happens to be harmless *today*: the fallback is `quarkus.application.name` (`openbank-document-service`), which is what's declared, so it's a no-op coincidence — until this channel or the service gets renamed.
- `auto.offset.reset` is not harmless: the effective value is the connector default (`latest`), not the declared `earliest`. On first deploy this consumer would skip every billing-fee event already on the topic rather than backfilling from the start.

Fix (matches `transaction-service`'s pattern): move both into a `document-service-msg-override.yaml` ConfigMap under `override.properties`, `config_ordinal=500`, rather than the plain `application.yaml` key. Do not just delete the keys — local dev and tests read the YAML fine; only the deployed path is affected.

JiRaska:
Reviewed at the PR head in a detached worktree. Two blocking findings, both of which I re-verified myself rather than taking on trust.

## 1. The template renders `{{document.currency}}` and `{{document.issueDate}}`; the service supplies neither

Extracting the placeholders from `ANNUAL_FEE_STATEMENT_CS_BODY` in `DocumentTemplateSeed.kt` gives:

```
{{document.currency}}  {{document.interestRate}}  {{document.issueDate}}
{{document.totalFees}}  {{document.year}}  {{account.iban}}  {{party.name}}
{{this.amount}}  {{this.category}}  {{this.name}}
```

`buildRenderData` (`AnnualStatementDeliveryService.kt:107-125`) puts `year`, `fees`, `totalFees`, `interestRate` under `document` — no `currency`, no `issueDate`.

Handlebars renders a missing key as the empty string, so this does not throw. It ships a **PAD Art. 5 statement of fees with no currency**: `Měna: `, every fee row as `120.00 `, the total as `420.00 `. A statement of fees whose amounts carry no unit is not a defensible regulatory document, and the failure is silent at every layer.

The sharpest part is that `currency` is a **mandatory** field on the way in — `AnnualFeeSummaryReadyConsumer.kt:82` rejects the event without it — and is then dropped on the way out. The value is present and thrown away.

`AnnualStatementDeliveryServiceTest` cannot see this because it asserts the keys the code produces. The assertion that would catch it is over the **rendered output** — that no `{{…}}` placeholder resolved to empty — or a direct comparison of the template's placeholder set against the render-data key set. The second is cheap and would keep catching it as templates change.

## 2. Nothing produces this topic, and the comment asserting otherwise is wrong

`application.yaml:222-230` says billing "drains ALL its events … to the SAME topic via its one `billing-events-out` channel — confirmed against openbank-billing-service/src/main/resources/application.yaml".

I checked that file. It has **no `mp:` block at all** — the only `smallrye` hit in it is `smallrye-openapi`. Grepping `openbank-billing-service/src` for `mp.messaging`, `@Channel` or `Emitter` returns **nothing**. And `BillingOutboxDispatcher` is explicit about where its rows go:

```kotlin
class BillingOutboxDispatcher(
    private val repo: OutboxRepository,
    private val publisher: LedgerOutboxEventPublisher,
```

Billing posts its outbox to **ledger-service over HTTP**. It is not a Kafka producer at all, so there is no `billing-events-out` channel to drain anything, and no `billing.annual-fee-summary.ready` will ever arrive. ADR-0248:120 still lists that producer as work to do, which agrees.

Landing the consumer first is a perfectly reasonable staged approach — the problem is the comment, which states a verification that does not hold and thereby retires the question for the next reader. Either say plainly that the producer is not built yet and reference ADR-0248, or build it. Same fiction appears at `openbank-infra/gitops/components/kafka/kafka.yaml:197`, already on `main`.

## Worth fixing

**3. Threat model and config disagree on the topic name.** `docs/threat-models/document-service.md:31` says `openbank.billing.billing.event`; `application.yaml:229` and the Kafka manifests say `openbank.billing.fee.event`. Config wins; the doc is wrong.

**4. The dotted-key baseline has no expiry.** `check-kafka-dotted-keys.py:102-107` adds two entries as "temporary, remove in the follow-up PR", and nothing fails if that PR never lands. Note the two are not equally benign: `group.id` is a no-op because its value equals `quarkus.application.name`, which is exactly what the fallback produces — but `auto.offset.reset: earliest` has no such coincidence available, so the connector default `latest` is what actually runs. Harmless for a brand-new consumer group today; the comment should say so, because the reason it is safe is not the reason it is baselined.

**5. Silent fee-line loss.** `AnnualFeeSummaryReadyConsumer.kt:105` uses `feesNode.mapNotNull { toFeeLine(it) }`, so an unparseable fee line is dropped while `totalFees` is kept — the statement can list fees summing to less than its own stated total, with only a per-line log. The test asserts this as intended. For a regulatory document, a dropped line should reject the whole event, the way a missing `totalFees` already does.

## Verified vs. read

**Ran** in the PR worktree: `run-gates.py --only duplicate-yaml-key-guard` (PASS, 59 files), `check-kafka-dotted-keys.py` (OK, 12 baselined / 0 new), `check-msg-channel-image-parity.py` (OK), `check-outbox-dispatch-enabled.sh` (OK). Extracted the template placeholder set and the render-data key set mechanically rather than by eye, and grepped billing for any Kafka producer.

**Read only**: every test is `fun …(): Unit = runBlocking {…}`, so none is silently skipped by JUnit. No `@ConfigProperty` added, no Panache `persist()`, no outbox entity, no annotation-binding hazard. Redis and a `Clock` producer are both genuinely wired, so `IdempotencyConfig` resolves.

**Not run**: the Gradle build for the module — a second Gradle build on this machine takes the port and dies with `QuarkusBindException`. Compilation, detekt and ktlint are unverified by me and rest on CI.

PR #4124, opened 2026-08-07
docs(customer-edge): scope the NearbyPay privacy claim to the request path

### LINKED ISSUE #3890: NearbyPay comment claims the creditor account never reaches the payer's device — the confirmation response returns it verbatim
## Summary

`CustomerEdgeResource.kt:1582` states a privacy property that holds for the **request** and not for
the **response**, and the comment does not distinguish them:

```kotlin
// NearbyPay (ADR-0087): if a paymentSessionToken is present, resolve the real creditor from
// the in-edge session store — the true account never reaches the payer's device; for SCA ...
```

The request half is true and well built: the payer's device sends a `paymentSessionToken`, the edge
resolves the real creditor account server-side from its session store, and the payer never supplies
the IBAN.

**The response half is not.** `createDomesticPayment` forwards the upstream answer verbatim:

```kotlin
val resp = upstream.post( ... )
...
return resp                       // CustomerEdgeResource.kt, no projection
```

and the upstream response schema declares the creditor account:

```yaml
# openbank-domestic-payment/src/main/resources/openapi.yaml
creditorAccountNumber:
creditorBankCode:
```

So on a NearbyPay payment the creditor's real account number and bank code are returned to the payer's
device in the confirmation, which is precisely what the comment says does not happen.

**The edge already knows how to project** — `projectConsent`, `projectLoan`, `projectMandate` all
exist in the same class. There is simply no projection on the payment path.

## Second path, same disclosure

`GET /transactions` calls `enrichWithCounterpartyIban(resp, accountId, customer.partyId)`
(`CustomerEdgeResource.kt:1410` → `:1432`), which re-adds the counterparty IBAN for internal
transfers. Even if the confirmation response were projected, the IBAN is served one request later in
history — so hiding it at initiation alone would be theatre rather than a control.

## Why this is worth an issue rather than a code comment fix

Two reasons, and the first is the one that matters:

1. **A comment asserting a privacy property is the only thing a reader checks.** Anyone extending
   NearbyPay — or, concretely, anyone implementing pay-to-phone under #3176, whose preferred design
   *depends* on the IBAN never reaching the payer — will read line 1582 and build on a guarantee that
   is not in force. #3176 was the investigation that surfaced this.
2. Whether the address should be disclosed at all is a **product decision**, not obviously a defect.
   A payer arguably has a legitimate interest in seeing where their money went. If disclosure is
   intended, the comment is wrong and should say so; if it is not, two code paths need projecting.
   Either way the current state — a claim that contradicts the code — is the one option that is
   certainly wrong.

## Not verified

- **Nothing was measured on a running pod.** This is source reading against `origin/main`: the
  comment, the verbatim `return resp`, and the upstream schema. I did not observe an actual NearbyPay
  confirmation payload.
- I did not check whether some later filter or `ContainerResponseFilter` strips these fields before
  they leave the edge — if one does, the comment is right and this issue is wrong. That is the single
  cheapest thing to check before acting.
- Whether the mobile client displays the field is unknown and irrelevant to the claim: the comment
  says it never *reaches* the device.

Refs #3176, ADR-0087

JiRaska:
Closes #3890. Refs #3176, ADR-0095.

## Verdict: stale prose, not a disclosure defect

The issue asked which of two things this is. It is **(b)** — the comment describes a guarantee the
code does not have and, on the evidence below, was never meant to have on the response side. The
honest fix is therefore to correct the prose, not to change a payment confirmation.

The issue's own finding is not disputed and is re-confirmed on `origin/main`: `createDomesticPayment`
ends `return resp`, `getDomesticPaymentStatus` ends `return resp`, there is no
`ContainerResponseFilter` in the module, and `DomesticPaymentResponse` declares
`creditorAccountNumber`, `creditorBankCode` and `creditorName` as **non-nullable** fields. The payer's
device does receive the creditor account. What changes is the reading of *whether that is wrong*.

### Why disclosure is intended, not a leak

1. **The threat the design names is harvesting, and it is fully defeated.** `directoryPayee`'s own
   KDoc gives the rationale: "phone numbers are guessable in a way account numbers are not, so anyone
   could walk a range of numbers and collect IBANs". A lookup still yields only a mask and a token.
   Learning the account now costs a real, SCA-signed payment to that person — not a free directory
   probe. That is the control, and it holds.
2. **The platform already discloses the counterparty deliberately, one request later.**
   `enrichWithCounterpartyIban` re-adds the counterparty IBAN to every `GET /transactions` page, and
   its KDoc states the position outright: "An IBAN the customer already transacted with is on their
   statement anyway". Suppressing the same value on the confirmation would be theatre against a path
   the repo has already argued through.
3. **ADR-0095 settles it, and contradicts the comment directly.** The accepted successor —
   QRlessPay, which "formalises and supersedes" this rail — transfers the **full SPAYD descriptor
   (IBAN + name + amount)** to the payer's device over GATT by design, and notes in passing that the
   ADR-0087 the comment cites "was **never actually written**". So the architecture direction puts the
   payee's IBAN on the payer's device, and the citation backing the stronger claim points at an ADR
   number that belongs to observability.
4. **Removing the fields is a breaking change with live consumers.** They are required fields of
   `DomesticPaymentResponse`, and `openbank-admin-ui` renders them
   (`src/app/payments/[id]/page.tsx:157`, `src/app/payments/page.tsx:922`). Projecting them away is a
   money-path API break, not a comment-sized correction.

## What this PR changes

No behaviour. Three prose sites carried the same over-broad claim, each scoped to the request path
and each stating it absolutely:

- `CustomerEdgeResource.kt` — the session-resolution comment ("the true account never reaches the
  payer's device"). Now states the request-side property that is true, and states explicitly that the
  confirmation *and* every status poll return the account, with the reasons.
- `CustomerEdgeResource.kt` — `directoryPayee`'s KDoc ("The payee's account number never reaches
  another customer's device").
- `PaymentSessionStore.kt` — the class KDoc ("The real creditor account never leaves the edge"). This
  one was the widest: the edge sends the account upstream to address the instruction, so it leaves in
  both directions.

Plus the same claim in the **published contract**, which is the copy external readers actually get:
`openapi.yaml`'s `/directory/payee` description. Editorial-only, so `info.version` moves
`1.30.0 -> 1.30.1` per ADR-0048 (read off live `origin/main`; no competing open PR touches the spec).

## Test: NearbyPayCreditorDisclosureTest

Three assertions pin both halves so they cannot silently swap again:

- the payer never *supplies* the account — a body with only a token still produces a fully addressed
  upstream instruction;
- the confirmation returned to the payer carries `creditorAccountNumber` / `creditorBankCode` /
  `creditorName`;
- the status poll carries them too — a second, independent path a projection applied only to the
  create would miss entirely.

**On red-then-green.** This PR changes no behaviour, so a test that is red before and green after is
impossible by construction; claiming one would be a lie. The falsification actually run is the
equivalent that is available: each assertion was fed the case it must reject — a stubbed projection
that strips the creditor fields from the create response — and the disclosure assertions go red while
the request-side one stays green. Proof is in the PR thread.

## Follow-up, deliberately not in this PR

Six code sites cite **ADR-0087** for this rail; that number belongs to ADR-0087 (Observability
Correlation & Profiling), and ADR-0095 records that the NearbyPay ADR "was never actually written".
Renaming the citations is a separate mechanical sweep and would bury this diff.

### DIFF
--- a/openbank-customer-edge/src/main/kotlin/com/openbank/customeredge/infrastructure/rest/CustomerEdgeResource.kt
+++ b/openbank-customer-edge/src/main/kotlin/com/openbank/customeredge/infrastructure/rest/CustomerEdgeResource.kt
@@ -1427,8 +1427,15 @@ class CustomerEdgeResource(
      * anyone could walk a range of numbers and collect IBANs. Instead the edge resolves the account
      * privately and hands back the SAME opaque token the nearby-pay rail already uses (ADR-0087):
      * the payer sees a name and a masked account, signs SCA against that masked form, and
-     * [createDomesticPayment] resolves the token back to the real account inside the edge. The
-     * payee's account number never reaches another customer's device.
+     * [createDomesticPayment] resolves the token back to the real account inside the edge. So a
+     * LOOKUP never yields an IBAN — the harvesting threat is what this defeats, and it is defeated
+     * because learning the account now costs a real, SCA-signed payment to that person rather than
+     * a free directory probe.
+     *
+     * It is deliberately NOT a promise that the account never reaches the payer at all. Once they
+     * have actually paid, the confirmation names the account their money went to (see
+     * [createDomesticPayment]), exactly as a statement does; do not build on the stronger reading
+     * (issue #3890).
      *
      * **Why the hash and not the partyId.** The caller must prove they already know the number, not
      * merely an id they saw once. Taking a partyId from the body would let anyone with a party id —
@@ -1906,8 +1913,20 @@ class CustomerEdgeResource(
         val debtorName = fetchPartyLegalName(debit.accountOwnerPartyId)
             ?: return badRequest("Cannot resolve debtor name")
         // NearbyPay (ADR-0087): if a paymentSessionToken is present, resolve the real creditor from
-        // the in-edge session store — the true account never reaches the payer's device; for SCA
-        // dynamic-linking we compare against the masked form (which is what the app signed).
+        // the in-edge session store — the payer never SUPPLIES the true account, they only hold a
+        // token and a mask; for SCA dynamic-linking we compare against that masked form (which is
+        // what the app signed).
+        //
+        // Request-side only, and say so: the confirmation below is the upstream
+        // DomesticPaymentResponse returned verbatim, and that DTO declares creditorAccountNumber /
+        // creditorBankCode / creditorName as required fields — so the payer's device DOES receive
+        // the account once the payment is made, here and again on every `getDomesticPaymentStatus`
+        // poll. That is intended (it is their own payment; `enrichWithCounterpartyIban` re-adds the
+        // counterparty IBAN on the next /transactions page anyway, and ADR-0095 — which formalises
+        // and supersedes this rail — hands the payer the full SPAYD descriptor by design). What is
+        // not intended is reading the masking as a response-side control: it is not one, and
+        // anything built on that reading is built on nothing (issue #3890, #3176).
+        // Pinned by NearbyPayCreditorDisclosureTest.
         val sessionTokenRaw = extractTextField(objectMapper, body, "paymentSessionToken")
         val nearbySession = sessionTokenRaw?.let { sessions.resolve(it) }
         if (sessionTokenRaw != null && nearbySession == null) {
--- a/openbank-customer-edge/src/main/kotlin/com/openbank/customeredge/infrastructure/rest/CustomerEdgeResource.kt
+++ b/openbank-customer-edge/src/main/kotlin/com/openbank/customeredge/infrastructure/rest/CustomerEdgeResource.kt
@@ -1425,7 +1425,7 @@ class CustomerEdgeResource(
      * **Why a session token and not an IBAN.** Answering with the payee's account number would turn
      * this route into a harvester: phone numbers are guessable in a way account numbers are not, so
      * anyone could walk a range of numbers and collect IBANs. Instead the edge resolves the account
-     * privately and hands back the SAME opaque token the nearby-pay rail already uses (ADR-0087):
+     * privately and hands back the SAME opaque token the nearby-pay rail already uses (ADR-0095):
      * the payer sees a name and a masked account, signs SCA against that masked form, and
      * [createDomesticPayment] resolves the token back to the real account inside the edge. So a
      * LOOKUP never yields an IBAN — the harvesting threat is what this defeats, and it is defeated
@@ -1912,7 +1912,7 @@ class CustomerEdgeResource(
         // downstream screening/AML party resolution that reads it.
         val debtorName = fetchPartyLegalName(debit.accountOwnerPartyId)
             ?: return badRequest("Cannot resolve debtor name")
-        // NearbyPay (ADR-0087): if a paymentSessionToken is present, resolve the real creditor from
+        // NearbyPay (ADR-0095): if a paymentSessionToken is present, resolve the real creditor from
         // the in-edge session store — the payer never SUPPLIES the true account, they only hold a
         // token and a mask; for SCA dynamic-linking we compare against that masked form (which is
         // what the app signed).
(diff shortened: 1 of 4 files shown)


JiRaska:
## Red-then-green proof

This PR changes no behaviour, so a test that is red before and green after cannot exist here. What
is available — and what was actually run — is the falsification: feed each assertion the input it
must reject, and check that it rejects it. The counter-case is the projection the issue proposed,
applied to **both** `return resp` sites.

**Run 1 — final tree (what this PR merges):**

```
FINAL GREEN: tests=3 failures=0 errors=0
  pass the payer never supplies the creditor account - the edge resolves it from the session()
  pass the confirmation returned to the payer carries the creditor account verbatim()
  pass the status poll returns the creditor account too - a second path to the same field()
```

**Run 2 — same tests, with a creditor-stripping projection temporarily added to
`createDomesticPayment` and `getDomesticPaymentStatus`** (`gradlew` exit 1):

```
RED RUN: tests=3 failures=2 errors=0
  pass  the payer never supplies the creditor account - the edge resolves it from the session()
  FAIL  the confirmation returned to the payer carries the creditor account verbatim()
  FAIL  the status poll returns the creditor account too - a second path to the same field()
```

with the failure text showing the fields genuinely gone from the payload, not a mangled assertion:

```
Expecting actual:
  {"id":"...","status":"RECEIVED","debtorAccountId":"...","debtorAccountNumber":"19-2000145399",
   "debtorBankCode":"0800","debtorName":"Payer Name","amount":"250.00","currency":"CZK"}
to contain:
  "creditorAccountNumber":"2600123456"
```

The split is the point. The two disclosure assertions fire; the request-side one stays green,
because stripping the response does not change what the payer supplied. A test that stayed green
under this projection would be asserting nothing, and the third one would not have discriminated
the two paths.

The probe was then reverted and the gate re-run on the exact merged tree:
`:openbank-customer-edge:test :ktlintCheck :detekt` — **BUILD SUCCESSFUL**, with all three tasks
present in the task list (`test`, `detekt`, `ktlintMainSourceSetCheck`, `ktlintTestSourceSetCheck`),
not merely absent-and-assumed.

JiRaska:
**Absorbed the seven in-file ADR citations from #4130, and bumped the spec to 1.31.0.**

Two follow-ups landed here rather than beside this PR:

**1. The spec version.** The `api-contract gate` classified this diff as ADDITIVE and demanded MINOR, so `info.version` is now `1.31.0`. That number satisfies the classification either way — ADR-0048 D5 wants MINOR-or-higher for additive and PATCH-or-higher for editorial, so a minor bump is valid under both readings while a patch bump is valid under only one, and the enforced gate is the authority on which applies. Read off live `origin/main` (1.30.0), not off this branch's base.

I could not reproduce *why* a `description`/`summary`-only diff classified as additive: oasdiff publishes no darwin/arm64 asset at the pinned 1.22.0, so the classifier will not run on this machine. Flagging that as an open question rather than asserting the gate is wrong — patch-level spec versions do exist in this repo's history, so the editorial path is demonstrably reachable and the naive explanation (the version bump itself counting as material) cannot be the whole story.

**2. The ADR number.** Seven of the nine stale `ADR-0087` citations live in the three files this PR already edits, four of them on lines it rewrites. Splitting them out meant two PRs touching adjacent text in the same hunks and racing on one `openapi.yaml`, so they are corrected here; #4130 keeps only the two `openbank-infra` manifest comments and no longer touches any spec.

The wrong number is **not a typo**: it is `openbank-app`'s own ADR-0087, carried across a repo boundary into a monorepo where 0087 already means the Observability Correlation & Profiling Layer. `PaymentSessionStore`'s KDoc now says so in place — without that, the number simply comes back, because the next reader greps `openbank-app`, finds an ADR-0087 about nearby-pay, and is correct about the other repo.

No further version move for the citation fix: the corrected token sits inside a `description` this PR was rewriting anyway. `:openbank-customer-edge:ktlintCheck` and `:detekt` both appear in the executed task list — BUILD SUCCESSFUL.

Issue #4119, opened 2026-08-07
engagement-service's REST contract ITs are failing on main, and path-scoped CI structurally cannot see it


JiRaska:
## The breakage

`openbank-engagement-service`'s `SurfaceRestContractIT` has four failing tests on current `main`:

```
an eligible party sees the catalogued banner                     AssertionError SurfaceRestContractIT.kt:42
a party without marketing consent is not_eligible, not empty     AssertionError SurfaceRestContractIT.kt:42
three posted dismissals suppress the next resolve …              AssertionError SurfaceRestContractIT.kt:74
an unknown slot is a 400, not a 500 or a silent empty result     AssertionError SurfaceRestContractIT.kt:92
```

Observed on #4114, whose entire engagement change is adding a `governance.yaml` — it cannot cause a REST contract failure. What it does is *touch* an `openbank-engagement-service/**` path.

## Why `main` cannot see it

Services CI is path-scoped. On `main`'s run `31216551896`:

```
success  Detect changed services
skipped  build              <-- no service path in the push, so nothing was built
success  all-green
```

`all-green` went green over **zero** services. So a service whose tests are broken stays green on `main` indefinitely, as long as no commit happens to touch its directory. That is not a hypothetical: it is the current state.

The two other open PRs that do build engagement — #4106 and #4072 — pass, but they are **8** and **62** commits behind `main`, while #4114 is **1** behind. So the regression is recent and only the newest tree meets it.

`SurfaceRestContractIT` was introduced by fac75d19e (#4048); the engagement commits after it are #4054 (`security(engagement): add edge-service-engagement authz policy`) and the two release commits. #4054 added `@Authorize` to `SurfaceResource`, and `authz.enforce` carries `defaultValue = "true"` in `AuthorizeInterceptor` — worth checking first, since an enforcing interceptor with no PDP in the test profile answers 503 and every one of these assertions would fail. Notably, #4072 adds `authz.enforce: false` under `"%test":` to engagement's `application.yaml`, which is exactly the shape of a fix for this — so the two may be the same story from opposite ends.

## Why this is worth more than the four tests

The failure mode is structural, and it is the third variant of it found today:

- `check-dockerfile-no-build-stage.py` could not run on a Dockerfile-only PR because its host job was conditional (#3629).
- The ADR-0071 governance gate could not see a released module that *omitted* `governance.yaml`, because its path filter keys on the file being present (#4099).
[...]

PR #4128, opened 2026-08-07
fix(engagement): SurfaceRestContractIT is red on main — an identity and test-profile authz enforcement

JiRaska:
All four `SurfaceRestContractIT` tests fail on `main`. They are invisible there — Services CI is path-scoped, and `main`'s own run reports `skipped build` because the push touched no service path, so `all-green` went green over **zero** services (#4119). The two other open PRs that do build engagement pass, but sit **8** and **62** commits behind and predate the breakage.

## Two defects, layered — the first masked the second

Each measured, not argued. `:openbank-engagement-service:test --tests '*SurfaceRestContractIT*'`, reading the JUnit XML after `rm -rf build/test-results` every time (a stale XML reports a green for code that never ran):

| `@TestSecurity` | `authz.enforce=false` | result |
|---|---|---|
| no | yes | 4 failures, all **401** |
| **yes** | **yes** | **4 passed** |
| yes | no | 4 failures, all **503** |

**1 — 401.** `SurfaceResource` carries `@RolesAllowed("ROLE_OPERATOR","ROLE_API","ROLE_ADMIN")` on both endpoints, and the IT authenticated as nobody. Its KDoc says it means to prove the contract *"without also needing a reachable OIDC token server"*, and `%test` does set `quarkus.oidc.enabled: false` — but disabling the provider removes the token **source**, not the **requirement**. Fixed with `@TestSecurity`, which is the fleet convention rather than my invention: 39 ITs use it, and the module already had `quarkus-test-security` on its test classpath.

**2 — 503.** With an identity in hand the request finally reaches `@Authorize`, and `authz.enforce` carries `defaultValue = "true"` in `AuthorizeInterceptor` while `@QuarkusTest` has no PDP bean. The third row is the control: removing that single key reproduces 503 on all four. Test profile only — the deployed posture is unchanged and is decided by the sidecar wired in #4068.

I ran the third row specifically because the first fix made the second one unfalsifiable: once 401 was gone I could have shipped the config key on the strength of *"`enforce=true, no PDP bean throws 503`" exists as a unit test elsewhere*, which is an inference, not evidence about this service. It costs one 6-minute run to turn it into evidence.

## What this unblocks

#4114 (engagement's missing `governance.yaml`, which closes the ADR-0071 drift keeping `main`'s `Admin UI build` red) currently fails on exactly these four tests. It adds one YAML file to engagement and cannot break a REST contract — it merely *touches* an engagement path, which is what makes CI run these tests at all.

## What this does NOT fix

The structural half of #4119: a service's tests can be broken on `main` indefinitely as long as no commit touches its directory, and `all-green` will report success over a skipped build the whole time. That wants a scheduled full-fleet build, and is a separate change.

### DIFF
--- a/openbank-engagement-service/src/main/resources/application.yaml
+++ b/openbank-engagement-service/src/main/resources/application.yaml
@@ -62,6 +62,13 @@ mp:
           serializer: org.apache.kafka.common.serialization.StringSerializer
 
 "%test":
+  # SurfaceResource's two endpoints are @Authorize-annotated, and `authz.enforce` carries
+  # defaultValue = "true" in AuthorizeInterceptor — so under @QuarkusTest, with no OPA PDP
+  # bean in the graph, every one of them answers 503 and SurfaceRestContractIT's
+  # `statusCode(200)` fails. Advisory in tests only; the deployed posture is unchanged and
+  # is decided by the sidecar wired in #4068.
+  authz:
+    enforce: false
   quarkus:
     management:
       enabled: false
--- a/openbank-engagement-service/src/test/kotlin/com/openbank/engagement/integration/SurfaceRestContractIT.kt
+++ b/openbank-engagement-service/src/test/kotlin/com/openbank/engagement/integration/SurfaceRestContractIT.kt
@@ -11,6 +11,7 @@ import io.restassured.module.kotlin.extensions.Extract
 import io.restassured.module.kotlin.extensions.Given
 import io.restassured.module.kotlin.extensions.Then
 import io.restassured.module.kotlin.extensions.When
+import io.quarkus.test.security.TestSecurity
 import io.smallrye.reactive.messaging.memory.InMemoryConnector
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.jupiter.api.Test
@@ -50,6 +51,7 @@ class SurfaceRestContractIT {
     }
 
     @Test
+    @TestSecurity(user = TEST_OPERATOR, roles = ["ROLE_OPERATOR"])
     fun `an eligible party sees the catalogued banner`() {
         StubConsentCheckPort.granted.set(true)
         val body = getBanner(UUID.randomUUID())
@@ -59,6 +61,7 @@ class SurfaceRestContractIT {
     }
 
     @Test
+    @TestSecurity(user = TEST_OPERATOR, roles = ["ROLE_OPERATOR"])
     fun `a party without marketing consent is not_eligible, not an empty list`() {
         StubConsentCheckPort.granted.set(false)
         val body = getBanner(UUID.randomUUID())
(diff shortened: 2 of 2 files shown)


JiRaska:
**Unblocked (`ae0263a7e`, signed). The only red gate was one misplaced import.**

```
SurfaceRestContractIT.kt:7:1 Imports must be ordered in lexicographic order
```

`import io.quarkus.test.security.TestSecurity` sat below the `io.restassured.*` block. One line moved, by `ktlintFormat` rather than by hand.

**It mattered far more than it looks.** Gradle stops at the first failing task, so `ktlintTestSourceSetCheck` failing meant `:test` **never ran on CI** — the PR that fixes `main`'s four red contract ITs could not demonstrate that it does. From the outside it read as "the engagement build is still broken", indistinguishable from the fault this PR exists to fix.

With the import ordered, the full local gate runs to completion. Reading *which tasks appear* rather than trusting a green summary:

```
> Task :openbank-engagement-service:detekt
> Task :openbank-engagement-service:ktlintTestSourceSetCheck
> Task :openbank-engagement-service:ktlintCheck
> Task :openbank-engagement-service:test
BUILD SUCCESSFUL
```

Verdict from the 6 JUnit XML files, not the console: **24 tests, 0 failures, 0 errors, 0 skipped.**

Those are the same 24 that report **4 failures on pristine `origin/main`** — reproduced in a detached worktree straight from `origin/main`, with no branch involved, before touching anything. So this PR does fix a real breakage of `main`, and now it can prove it.

**Blast radius, for whoever prioritises this:** `build (openbank-engagement-service)` red on `main` is inherited by every PR that builds the service. #3880, #4114 and #4133 are red for it while touching no engagement code at all.

engagement-service actually landed with **three** independent faults, easy to mistake for one: no `governance.yaml` (gates — #4114), missing from `OPENBANK_NAMESPACES` so the admin-ui registry guard fails (also #4114), and these four contract ITs (this PR). All three need to land before the fleet's PRs go green.

Nothing rebased or force-pushed — an additive commit. `/private/tmp/ob-authz-95130` was clean, dormant 45 minutes, and its head matched the remote exactly before I moved.

JiRaska:
**Superseded — `main` already carries this fix.** Leaving the close to you rather than doing it myself.

#4114 absorbed the `@TestSecurity` change while consolidating the engagement work, so the defect is gone from `main` independently of this PR. Verified by running the tests against pristine `origin/main` @ `75e94620c` in a detached worktree, no branch involved:

```
:openbank-engagement-service:test → BUILD SUCCESSFUL
tests=24 failures=0 errors=0
```

Those are the same 24 that reported 4 failures earlier tonight from the same command against the same ref.

For the record, this PR's diagnosis was right and its only red was cosmetic: `ktlintTestSourceSetCheck` on one misplaced import, which I fixed in `ae0263a7e`. That mattered more than it looked — Gradle stops at the first failing task, so `:test` never ran on CI and the PR could not demonstrate the very thing it fixed.

PR #4129, opened 2026-08-07
feat(billing-service): add annual fee-summary aggregation for PAD Art. 5 statements

JiRaska:
## Summary

Adds the annual fee-summary aggregation billing-service needed for ADR-0248's PAD Art. 5 annual
statement of fees: a new use case aggregates `AssessedFee` history per account/year, and publishes
`billing.annual-fee-summary.ready` via the service's existing outbox for document-service to
render-and-discard (never persisted, per ADR-0248's on-demand rule).

## Linked issues

Refs #4109
Refs ADR-0248

## Changes

- `AnnualFeeSummaryService` — aggregates posted, non-reversed/non-waived `AssessedFee` rows for
  an account/year into `AnnualFeeSummary` (fees grouped by category, total, interest rate if
  available in this service's own domain).
- `AnnualFeeSummaryOutboxPayloads` — publishes `billing.annual-fee-summary.ready`
  (`eventType: "AnnualFeeSummaryReady"`) via the existing `BillingOutboxRepositoryImpl`, matching
  the contract document-service's consumer (PR #4122) implements against.
- `AnnualFeeSummaryScheduler` — new `@Scheduled` (`suspend fun`) yearly trigger for the
  aggregation, **disabled by default** via config flag, mirroring statement-service's own
  close-scheduler pattern — an annual job firing accidentally in dev/test would be a real
  incident.
- `docs/threat-models/openbank-billing-service.md` updated for the new outbound outbox event.
- Unit tests for the aggregation (fee grouping, exclusion of reversed/waived fees, total
  computation) and the outbox payload shape.

## Reviewer notes

**Money-path service change — `openbank-billing-service` is on `rules.yaml:
money_path_services`. This needs 2 human approvals and a threat-model diff per repo governance
(ADR-0030) and is explicitly not self-mergeable — I will not merge this myself, and will not use
`--admin`/any override on `gh pr merge`/`gh pr review`.**

The aggregation only reads `AssessedFee`/`BillingCycle` history and writes a new outbox row — it
does not touch fee assessment, posting, or reversal logic.

## Test plan

- [x] `./gradlew :openbank-billing-service:ktlintCheck :openbank-billing-service:detekt` —
      BUILD SUCCESSFUL.
- [x] `./gradlew :openbank-billing-service:test` — 100/101 passing; the one failure
      (`BillingResourceAuthzTest`, a Quarkus HTTP-server port-bind failure) is in a file this PR
      does not touch and reproduces as a shared-machine environmental flake, not a regression.


JiRaska:
`gates (data)` — two enforced gates, both about the new outbound publisher for `openbank.billing.fee.event`:

**`Kafka topic existence (issue #2598, enforced)`** — no `KafkaTopic` CR declares this topic in gitops. A consumer subscribed to a topic that doesn't exist receives nothing and reports healthy, so this can't show up as a red signal at runtime — needs a gitops declaration alongside the other billing topics.

**`event-contract coverage ratchet (ADR-0006, enforced)`** — no `openbank-contracts/openbank-billing-service/asyncapi.yaml` entry for this channel: message schema, producer, consumers, retention, partitioning key. ADR-0006 is a ratchet — adding the topic to `.github/event-contract-baseline.txt` isn't the fix, the baseline only shrinks.

Cross-checked against #4122 (document-service, the consumer side): the topic string matches exactly — `openbank.billing.fee.event` on both PRs, no naming drift between producer and consumer. So this is purely the missing gitops/contract registration for a topic that's otherwise wired correctly, not a coordination bug between the two PRs.

JiRaska:
Checked the outbox usage against the diff — `publishForAccount` builds the summary and writes to the existing outbox port, returning `null` and appending nothing when the party can't be resolved rather than publishing a partial/invalid summary. `publishForAllAccounts` iterates account-service's ACTIVE-account fleet and tracks published/skipped separately, consistent with the "on-demand, never persisted" framing in ADR-0248.

Not merging — money-path/high-risk paths, human review by rule.

JiRaska:
Worked this from the autonomous queue. Three things were failing; two are fixed here, one is yours.

## Fixed in this PR

**`gates (gitops)` — scheduler liveness.** This PR adds a new `@Scheduled` job and ADR-0237 requires the heartbeat in the same PR. Registered on `StartupEvent` rather than `@PostConstruct`, because `@ApplicationScoped` is lazy and a `@PostConstruct` would first run when the cron fires — for an *annual* job that is up to a year of an absent gauge, and absent is a different signal from stale. Registered only when `enabled`, and `recordSuccess()` only after `publishForAllAccounts` returns: a heartbeat on a disabled scheduler or a failure path asserts exactly the work it exists to prove happened. Gate now passes.

**The branch was 501 commits behind `main`.** Merged in; the merge was clean, and I checked the structured files by entry count rather than by eye, because a clean merge can drop a list entry with no conflict: `.release-please-manifest.json` 58→58, `gates.yaml` 146→146, `verification-metadata.xml` 3250→3250, nothing lost from either side.

**CodeQL alert 413** (`ObjectMapper.setSerializationInclusion` deprecated) → `setDefaultPropertyInclusion`. Verified rather than assumed: the build emits 14 deprecation warnings and none names this file, so the replacement is genuinely not deprecated in the Jackson version shipped here. Serialized output unchanged, which matters because this payload carries money — `AnnualFeeSummaryOutboxPayloadsTest` asserts it.

## Not this PR's fault — gate blind spot, fix in #4887

`event-contract-code-agreement` reported `message AnnualFeeSummaryReady is declared in the contract but no class declares that event type`. The class does declare it. `DATA_CLASS_RE` was anchored on a bare `^data class`, so every **`internal data class`** was invisible — 19 of the fleet's 1317. #4887 fixes it, falsified in both directions.

I did rename `EVENT_TYPE_LITERAL` → `EVENT_TYPE` here, which is needed regardless: 8 other services use that name and it is the only form the gate recognises.

## Yours to decide

Two genuine disagreements remain, and they are a design call rather than a fix:

```
billing.fee.post-intent.v1      emitted by BillingAssessmentRepositoryImpl, no message in the contract
billing.fee.reversal-intent.v1  same
```

**Adding a contract for one event turns on checking for all of the service's events.** Before this PR, billing had no contract, so all three were grandfathered and invisible; the first contract makes the other two visible. That is the gate working, not a regression.

I did not write those two messages. A contract for a money event is a published interface, and deriving one from the producer's current payload would commit consumers to whatever it happens to emit today — that is your call, not something to infer. The alternative, narrowing this PR's contract, would leave the same gap documented differently.

Nothing merged. `openbank-billing-service` is not in `money_path_services`, but the protected-path guard holds it regardless (`openbank-billing-service/**`), so it needs your review either way.

PR #4130, opened 2026-08-07
docs(customer-edge): point the nearby-pay ADR citations at ADR-0095

JiRaska:
Refs #3890, #4124, ADR-0095, ADR-0147.

The follow-up sweep #4124 flagged and deliberately deferred: six code sites (and three more it did
not count) cite **ADR-0087** as the authority for the nearby-pay / NearbyPay rail. In this monorepo
ADR-0087 is *Observability Correlation & Profiling Layer* — it has nothing to do with payments.

## Where the number came from

Not a typo. ADR-0095 (QRlessPay) says it outright in its Context: the informal nearby-pay work is
"referenced in code as *ADR-0087*, which was **never actually written**", and ADR-0095 "formalises
and supersedes that informal work". ADR-0147 explains the mechanism — `openbank-app` is a separate
repo that numbers its ADRs independently, so *its* ADR-0087 is the nearby-pay note, and the citation
was carried across repo boundaries into a monorepo where 0087 already means something else.

So the correct citation for this rail in this repo is **ADR-0095**.

## Classified, not blanket-replaced

26 references to ADR-0087 exist repo-wide. **9 are stale, 17 are correct and left alone.**

**Corrected (9) — all point at ADR-0095:**

| File | Line | Site |
|---|---|---|
| `openbank-customer-edge/.../CustomerEdgeResource.kt` | 1428 | `directoryPayee` KDoc — "the SAME opaque token the nearby-pay rail already uses" |
| `openbank-customer-edge/.../CustomerEdgeResource.kt` | 1908 | `createDomesticPayment` — "NearbyPay (ADR-0087)" session resolution |
| `openbank-customer-edge/.../CustomerEdgeResource.kt` | 3483 | section marker "Nearby payments (payment sessions)" |
| `openbank-customer-edge/.../CustomerEdgeResource.kt` | 3535 | status-poll KDoc — "ADR-0087 phase 2" |
| `openbank-customer-edge/.../PaymentSessionStore.kt` | 12 | class KDoc |
| `openbank-customer-edge/.../PaymentSessionStore.kt` | 118 | `maskIban` KDoc |
| `openbank-customer-edge/src/main/resources/openapi.yaml` | 215 | `/directory/payee` description — the **published** copy |
| `openbank-infra/gitops/components/customer-edge/customer-edge.yaml` | 245 | Redis env comment |
| `openbank-infra/gitops/components/customer-edge/redis.yaml` | 5 | Redis purpose header |

**Left untouched (17):**

- **11 genuine observability references** — ADR-0077 (1), ADR-0087 itself (2), ADR-0088 (5),
  ADR-0218 §correlation (1), `openbank-admin-ui/src/app/observability/stack/page.tsx` (2). Each was
  read in context; every one is about the LGTM(P) correlation layer. Replacing these would be a new
  defect.
- **5 deliberate meta-references** to the app-side numbering, inside ADR-0095 (4) and ADR-0147 (1).
  These exist *to record the collision*; "correcting" them would erase the explanation and make two
  accepted ADRs wrong. Accepted ADRs are historical records and are not edited here.
- **1 historical CHANGELOG entry** (`openbank-customer-edge/CHANGELOG.md:482`) — a derived
  release-please artefact, never hand-edited.

Nothing under `docs/adr/` changed, so the derived `README.md` / `DIGEST.md` / `index.json` are
untouched and no regeneration was needed.

## One claim fixed, not just its citation

`PaymentSessionStore`'s class KDoc gains a five-line **ADR pointer** stating why ADR-0087 is wrong
here and that ADR-0095 is the citation. Without it the number simply comes back: the next reader
greps `openbank-app`, finds an ADR-0087 about nearby-pay, and is right — in the other repo. This is
the one place a future contributor is guaranteed to be standing when they need to know.

## No behaviour change

Comments, one gitops comment pair, one OpenAPI description. `ktlintCheck` and `detekt` were run for
`:openbank-customer-edge` and both tasks are visible in the executed task list (not skipped behind an
earlier failure) — BUILD SUCCESSFUL.

## Two things a reviewer should know

**1. Deliberate overlap with #4124.** Four of the nine sites (`CustomerEdgeResource.kt` 1428/1908,
`PaymentSessionStore.kt` 12, `openapi.yaml` 215) sit inside hunks #4124 also edits — but #4124 leaves
the `ADR-0087` tokens on *context* lines, so the two PRs change different text on adjacent lines.
Whichever lands second will need a small textual conflict resolution; neither reverts the other, and
the resulting prose is the union of both.

**2. Spec version race.** `info.version` moves `1.30.0 -> 1.30.1` (editorial, ADR-0048), read off
live `origin/main`, not off this branch's base. #4124 independently claims `1.31.0`. If #4124 merges
first this PR's bump becomes a regression and must be re-taken as `1.31.1` — the api-contract gate
will say so out loud rather than letting it through. That is the intended signal, not a surprise.

Not for merge by the author — no `--admin`, no override.

### DIFF
--- a/openbank-customer-edge/src/main/kotlin/com/openbank/customeredge/infrastructure/rest/CustomerEdgeResource.kt
+++ b/openbank-customer-edge/src/main/kotlin/com/openbank/customeredge/infrastructure/rest/CustomerEdgeResource.kt
@@ -1425,7 +1425,7 @@ class CustomerEdgeResource(
      * **Why a session token and not an IBAN.** Answering with the payee's account number would turn
      * this route into a harvester: phone numbers are guessable in a way account numbers are not, so
      * anyone could walk a range of numbers and collect IBANs. Instead the edge resolves the account
-     * privately and hands back the SAME opaque token the nearby-pay rail already uses (ADR-0095):
+     * privately and hands back the SAME opaque token the nearby-pay rail already uses (ADR-0087):
      * the payer sees a name and a masked account, signs SCA against that masked form, and
      * [createDomesticPayment] resolves the token back to the real account inside the edge. The
      * payee's account number never reaches another customer's device.
@@ -1905,7 +1905,7 @@ class CustomerEdgeResource(
         // downstream screening/AML party resolution that reads it.
         val debtorName = fetchPartyLegalName(debit.accountOwnerPartyId)
             ?: return badRequest("Cannot resolve debtor name")
-        // NearbyPay (ADR-0095): if a paymentSessionToken is present, resolve the real creditor from
+        // NearbyPay (ADR-0087): if a paymentSessionToken is present, resolve the real creditor from
         // the in-edge session store — the true account never reaches the payer's device; for SCA
         // dynamic-linking we compare against the masked form (which is what the app signed).
         val sessionTokenRaw = extractTextField(objectMapper, body, "paymentSessionToken")
@@ -3480,7 +3480,7 @@ class CustomerEdgeResource(
         .type(MediaType.APPLICATION_JSON)
         .build()
 
-    // --- Nearby payments (payment sessions, ADR-0095) ---
+    // --- Nearby payments (payment sessions, ADR-0087) ---
 
     /**
      * Create a nearby-pay session bound to one of the caller's OWN accounts (the receiver). Returns an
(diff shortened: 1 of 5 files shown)


JiRaska:
**Narrowed to the two gitops manifest comments; the version race is gone.**

This PR and #4124 were both editing `openbank-customer-edge/src/main/resources/openapi.yaml` and both claiming a version off `main` — 1.30.1 here, 1.31.0 there. Worse, four of the citations here sit on lines #4124 rewrites, so the two would have changed adjacent text in the same hunks.

Resolution: the seven citations in customer-edge's own sources moved **into #4124**, which was already rewriting those exact lines and can correct the number in the same breath. This PR now touches no spec, needs no `info.version` bump, and cannot collide with anything.

What is left here is genuinely separate — two comment lines in `openbank-infra`:

```
-  # Redis: nearby-pay payment sessions (ADR-0087) + WebAuthn RP challenge/
+  # Redis: nearby-pay payment sessions (ADR-0095) + WebAuthn RP challenge/
```

The 17 legitimate ADR-0087 references stay untouched, and the reason the five inside ADR-0095 and ADR-0147 must stay is worth keeping in view: they exist **to record the collision**. ADR-0095's Context states that the nearby-pay ADR this number pointed at "was never actually written" here, and ADR-0147 explains that `openbank-app` numbers its ADRs independently. Rewriting them would delete the explanation and make two accepted ADRs wrong — the sweep's most likely way to do damage, avoided by reading each site in context rather than replacing the token globally.

PR #4133, opened 2026-08-07
fix(libs)!: delete three inert annotations and correct the records that called them shipped

### LINKED ISSUE #4011: Three of five openbank-libs annotations are inert (@Idempotent, @Audited, @MaskSensitive) — no binding, no interceptor, zero usages, and two are recorded as shipped controls
**`@Idempotent` has zero usages, no interceptor, and no CDI binding — it cannot work, and it is
advertised as a platform control.**

```
$ git grep -n '@Idempotent' origin/main
docs/adr/0068-onboarding-operations-cockpit.md:55: … `openbank-libs` (`AuditEvent`/`@Audited`,
    `@Authorize`/OPA, `@Idempotent`, the outbox) …
```

One occurrence in the whole repository, and it is prose describing the annotation as something the
platform offers. Nothing applies it.

The declaration (`openbank-libs-domain/.../idempotency/IdempotencyStore.kt:21-23`):

```kotlin
@Target(AnnotationTarget.FUNCTION)
@Retention(AnnotationRetention.RUNTIME)
annotation class Idempotent(val ttlSeconds: Long = 86400, val headerName: String = "Idempotency-Key")
```

It is **not** an `@InterceptorBinding`, and there is no interceptor anywhere in `openbank-libs-runtime`
that references it. So this is not "declared but unadopted" — applying it to an endpoint today would
compile, look correct in review, and do **nothing at all**. On a payment endpoint that is the worst
possible silent failure: a duplicate `POST` would be processed twice while the annotation says it
cannot be.

Contrast with `@Authorize` and `@FeatureFlag`, which *are* `InterceptorBinding`s with real
interceptors (and are the subject of #3670 precisely because of that).

## What is actually in force

The `IdempotencyStore` itself is real and adopted — nine services produce it and inject it into a
resource, calling `get`/`save` by hand:

```
producer + resource injection:
  account-service   aml-service      consent-service   delegation-service   domestic-payment
  psd2-service      sca-service      sepa-payment      tpp-registry-service
```

So idempotency works where it is implemented; it is just implemented manually, endpoint by endpoint,
with no compile-time or runtime signal for an endpoint that forgot.

## Suggested handling — one of two, not both

1. **Delete the annotation.** It has no users and no implementation; removing it is a pure subtraction
   and stops the trap. Update ADR-0068's sentence at the same time, since that is currently the only
   thing asserting it exists.
2. **Implement it** as an `@InterceptorBinding` + interceptor over the existing `IdempotencyStore`,
   and migrate the nine hand-rolled call sites. That is the bigger change, and the argument for it is
   that a manual pattern replicated nine times across money-path services is exactly where one endpoint
   quietly diverges.

**Whichever is chosen, the trap to close first is that the annotation currently exists and is inert.**
Between those two options, deleting it is safe to do immediately and does not block choosing (2) later.

Worth checking as part of either: **which write endpoints on those nine services do *not* consult the
store.** That is the question the annotation was presumably meant to make answerable, and today it is a
manual audit. `domestic-payment` is the one to look at first — this repo already records a
"fake idempotency key" defect there.

Found by sweeping the live databases for idempotency tables (only `account_idempotency` 26 rows and
`ledger_idempotency` 150 rows exist as Postgres tables; the shared store is Redis-backed, so table
absence is not the measure) and then tracing the annotation back through the code.

JiRaska:
Closes #4011

## What was wrong

`openbank-libs-domain` published three annotations that were plain `@Retention(RUNTIME)`
markers — no `@InterceptorBinding`, no interceptor anywhere in `openbank-libs-runtime`, and
zero applications in any module's `src/main`:

| annotation | `@InterceptorBinding` | interceptor | usages in `src/main` |
|---|---|---|---|
| `@Idempotent` | no | 0 | 0 |
| `@Audited` | no | 0 | 0 |
| `@MaskSensitive` | no | 0 | 0 |
| `@Authorize` (for contrast) | yes | 2 | 42 |
| `@FeatureFlag` (for contrast) | yes | 1 | 1 |

Applying one compiled, reviewed as correct, and did nothing. `@Idempotent` on a payment
endpoint would have let a duplicate `POST` through while the source read as protected;
`@MaskSensitive` on an IBAN field would have serialised it in full, because neither
serialization filter its own KDoc promised — "admin-ui proxy, audit-event sanitizer" — was
ever written.

## Where the false control declarations lived

This is the half that actively misleads, and all of it is corrected here.

**A STRIDE mitigation in a threat model** (`docs/threat-models/onboarding-service.md`, the
ADR-0030 artifact), Repudiation row:

> Operator denies viewing a flagged sanctions case — Read access is audited via `@Audited` on
> detail endpoint; `AuditEvent` includes `actorId`, `resourceId`, `traceId`

onboarding-service emits no audit event at all. The row now reads **OPEN — not mitigated**,
names the explicit wiring a fix would need, and lines up with ADR-0176, which already declines
to claim GDPR Art. 5(2) coverage for operator reads.

**A "Shipped" record in ADR-0014**, Phase 3:

> **Phase 3 (security + audit foundation)** — ✅ Shipped: `PiiMask`/`@MaskSensitive`; …

The `PiiMask` functions shipped and work. The annotation half never did — and the same ADR's
Phase 3 body tied it to "the GDPR Art. 25/32 fix for the admin UI's unmasked PII tables".

Also corrected: ADR-0047 (listed `AuditEvent` + `@Audited` among governance primitives already
held), ADR-0068 (advertised both annotations as `openbank-libs` platform controls — the only
prose asserting `@Idempotent` existed at all), ADR-0176 (the one document that told the truth,
kept true after the removal), and `openbank-libs/docs` in all three language variants, where
`@MaskSensitive` sat in the compliance-control table mapped to GDPR Art. 25/32 and appeared in
an adopter-facing code sample.

## Why delete rather than implement

Zero call sites, so the subtraction cannot break a caller, and it closes the trap now. It
forecloses nothing: a real `@InterceptorBinding` over `IdempotencyStore` can land later with
its interceptor in the same PR, which is the only order that is ever safe. Implementing three
interceptors speculatively would be building for no measured caller.

`IdempotencyStore` itself is untouched and still adopted by nine services.

## The gate, red-then-green

New `libs-annotations-implemented` (enforced, group `kotlin`), declared in
`.github/gates/gates.yaml`, not as an inline step. Two rules, each falsified separately.

**RULE A** — an annotation declared in an `openbank-libs-*` module must either carry
`@InterceptorBinding` **with** a matching interceptor, or be applied at least once in some
`src/main`.

Against pristine `origin/main` in a detached worktree, with the final version of the script:

```
::error::…/audit/AuditEvent.kt: @Audited is INERT — …
::error::…/idempotency/IdempotencyStore.kt: @Idempotent is INERT — …
::error::…/security/PiiMasking.kt: @MaskSensitive is INERT — …
3 problem(s).   exit=1
```

Exactly three, matching the issue's own table, and nothing else — the count is the check.

**RULE B** — no document under `docs/adr`, `docs/threat-models` or `openbank-libs/docs` may
name an `@Annotation` that no Kotlin source declares or imports. Against the intermediate
state (annotations deleted, documents *not* yet corrected) — the exact shape that made this
dangerous rather than untidy:

```
25 problem(s)
```

including ADR-0014:18, ADR-0047:45, ADR-0068:55, ADR-0176:342, the onboarding threat model
row, and all twelve `openbank-libs/docs` sites. On this branch:

```
openbank-libs annotations: all implemented or used; no document names a phantom one.
exit=0
```

The "real annotation" set is **derived** from the sources (`annotation class X`, or an import
of `X`), never hand-kept, so it cannot drift from what the code actually has.

**Code-about-code precedence was decided before the first run**, because a text guard always
flags the prose explaining the defect it exists to catch. The rule: naming an absent
annotation is permitted when the surrounding *paragraph* cites an issue — a correction points
at its record, a claim does not. Both directions are in `--self-test`. Prose about
deliberately-unadopted third-party annotations (ADR-0016's `@RunOnVirtualThread`, a
residual-risk column's `@RateLimited`) sits in `PROSE_ALLOWLIST` with a reason each, and the
script **fails on a stale entry in either direction**, so an exemption and its removal move
together.

## Verification

- `python3 .github/scripts/run-gates.py --only libs-annotations-implemented` → PASS (self-test
  PASS + gate PASS)
- `./gradlew :openbank-libs-domain:test :openbank-libs-domain:ktlintCheck
  :openbank-libs-domain:detekt` → BUILD SUCCESSFUL, all three tasks present in the output
- `adr-registry-integrity-check`, `threat-model-coverage`, `license-header-consistency` → PASS

## Explicitly not in scope

The audit of which write endpoints on the nine `IdempotencyStore`-adopting services fail to
consult it (per the issue: start with `domestic-payment`). `@Authorize` and `@FeatureFlag` are
genuine `InterceptorBinding`s with real interceptors and are untouched; their adoption question
is #3670.

---
Two observations found by the new Rule B while tuning it, both allowlisted rather than
silently swallowed, and neither in scope here: `openbank-libs/docs` recommends
`@RegisterClientHeaders(BearerTokenClientHeadersFactory::class)` as the S2S call-site shape and
no service imports it; `docs/threat-models/openbank-dispute-service.md` names `@RateLimited` in
its residual-risk column, correctly, as the control that is still open.


JiRaska:
Reviewed. **Approve.** One labelling nit, not worth blocking.

**No fleet compile break — measured, with the probe validated in the same run.** Usages across all `.kt` on `origin/main`, excluding each annotation's own declaration:

```
@Audited         0        @Authorize    637   <- control: the probe works
@Idempotent      0        @FeatureFlag    9   <- control
@MaskSensitive   0
```

Three zeros from a probe that demonstrably finds things is absence; three zeros on their own would only mean the grep was broken. Also swept string-literal and reflective lookups by name across `.kt/.java/.yaml/.properties`: the single hit is an English log message ("Idempotent settlement hit") in `SettlementAdapter.kt:82`, not a reference.

`MaskStrategy` is correctly **retained** — still used by `PiiMask.apply` and `PiiMaskTest`; the diff removes only the annotation class.

**No silent CDI behaviour change.** None of the three carried `@InterceptorBinding` — only `@Target` + `@Retention(RUNTIME)`, visible in the removed lines — and no interceptor exists for any of them. That matters because a deleted interceptor binding changes behaviour *without* failing to compile, which is the one way this kind of deletion could bite quietly. It cannot here: there was no behaviour to remove.

**The `!` is over-declared.** Both annotations were RUNTIME-retained, but nothing consumes them, and — the part that settles it — **`openbank-libs-domain` is not a released component**: no `version.txt`, and absent from every entry in `release-please-config.json` (verified; no root `.` entry either, and no registered path is a prefix of it). So the breaking marker triggers no major bump anywhere and is communicative only. `build.gradle.kts` applies `maven-publish` but declares no `publishing`/`repositories` block, so nothing reaches an external repo and no outside consumer can exist. `fix(libs):` without the `!` would be more accurate; harmless as written.

**The gate is falsified, not just green:** `--self-test` passes; the merged branch gives rc=0; and restoring only `main`'s three annotation files reproduces the red exactly — rc=1, `3 problem(s)`, naming `@Audited`, `@Idempotent`, `@MaskSensitive` and nothing else. The count matching the issue's table is the real check, since a checker that flags the right number for the wrong reason would still print 3.

Issue #3555, opened 2026-08-03
mcp-db and litellm-db have no CNPG backup at all — 2 of 55 clusters with no recovery point


JiRaska:
## What

Two of 55 CNPG clusters declare no `spec.backup.barmanObjectStore` — they have **no WAL archiving and no base backups at all**. There is no recovery point for either.

| cluster | manifest | namespace | instances | storage | owner |
| --- | --- | --- | --- | --- | --- |
| `mcp-db` | `openbank-infra/gitops/components/mcp/postgres.yaml` | `platform` | 1 | 2Gi | `openbank-mcp-service` (`primaryDatastore: PostgreSQL`, `databaseName: openbank_mcp`, ADR-0224 D2 agent-session store) |
| `litellm-db` | `openbank-infra/gitops/components/ai-platform/postgres.yaml` | `ai-platform` | 1 | 2Gi | LiteLLM gateway (virtual keys, spend ledger) |

Both are `instances: 1`, so there is no replica either: a lost EBS volume or a failed node reclaim is total, unrecoverable data loss for that database.

## How it was measured

Derived structurally from the manifests (`gitops_facts.cnpg_backup_configured`, merged in #3551) — every top-level `kind: Cluster` document under `openbank-infra/gitops/`, asked for the nested key path `spec.backup.barmanObjectStore`:

```
clusters: 55  without backup: 2
  litellm-db  openbank-infra/gitops/components/ai-platform/postgres.yaml
  mcp-db      openbank-infra/gitops/components/mcp/postgres.yaml
```

This is a fact about the *manifests*, not about the live cluster — confirm with `kubectl get cluster -A -o custom-columns=NS:.metadata.namespace,NAME:.metadata.name,BACKUP:.spec.backup.barmanObjectStore.destinationPath` before acting.

## Why it went unnoticed

`docs/runbooks/svc-mcp.md` promised on-call `RPO target: <= 5 min (continuous archiving)`, `RTO <= 30 min` for a database with no backup. The runbook generator decided "has a backup" with a whole-file substring test, and `rules.yaml` names `barmanObjectStore` inside a rule *description* — which every service's OPA bundle ConfigMap embedded verbatim until #3508. #3508 removed the embed (mcp's runbook is now honest: `RPO/RTO: undefined`), #3551 made the detection structural so the collision cannot come back. **Neither PR gave either cluster a backup.** That is this issue.

[...]

PR #4141, opened 2026-08-07
feat(governance): enforce that a `partial` ADR declares its follow-up (#3965)

### LINKED ISSUE #3965: ADRs with a 'partial' delivery status and no tracking issue: the followup rule is unenforced, and the detector cannot see cross-repo or uncited tracking
## Summary

`rules.yaml:1850-1852` says an issue is opened for a **`governance_followup` — the actionable tail of
an ADR (rollout, reg condition, gate flip)**. A `partial` ADR has such a tail by definition: something
shipped and something did not.

Measured against `docs/adr/index.json` on `origin/main`:

```
ADRs indexed                    : 237
  shipped                       :  92
  partial                       :  85
  planned                       :  50
  n-a                           :  10

partial ADRs WITH an open issue mentioning them : 33
partial ADRs WITHOUT                            : 52
```

**52 ADRs record that half of a decision is unbuilt, and nothing in the backlog says so.**

## Why `planned` is deliberately excluded from the count

28 of the 50 `planned` ADRs also have no issue, and that is **fine** — ADR-0052 is explicit that
issues must not duplicate ADRs, and a `planned` decision has no actionable tail yet. Filing 28 issues
for those would be exactly the noise the rule forbids. The `partial` set is different: the tail exists
by construction.

## What this costs, from today

The failure mode is not abstract — three instances turned up in one session, each found only because
someone happened to look:

- **ADR-0023 F2** (S3 Object Lock WORM) — honestly marked 🟡 "not yet deployed", untracked anywhere,
  and its committed defaults name `eu-central-1` while the account is `eu-north-1`, so the first
  deployment attempt would fail confusingly. Now #3962.
- **ADR-0034 Phase 5** — `AUTHZ_ENFORCE=false` on 18 of 39 workloads, four money-path. Tracked only
  because #3679 exists.
- **ADR-0237** — 32 schedulers still without a liveness heartbeat; the adoption ratchet that should
  stop new ones is `mode: advisory`, so #3875 merged a new violation the day the gate reported it
  (#3931).

The shape is consistent: **a status inside a document is not a work item.** It is read once, at
authoring time, and then only by whoever goes looking.

## Suggested

Not 52 issues. The cheap version is a check that a `partial` ADR either has an open issue referencing
it **or** carries an explicit "no follow-up needed" marker in its front matter — the same
shrink-only, must-justify-the-exception shape `check-pact-provider-replay.py` and
`readiness-attestation-format` already use. `adr-registry-integrity-check` (enforced) already parses
this index, so the data is in hand.

That converts "someone should look" into "a new `partial` ADR without a tail is red at PR time",
which is the only thing that stops number 53.

## Not verified

- **The 52 are not individually triaged.** Some will have follow-up tracked under a differently-worded
  issue that my `ADR-NNNN` search missed — the search is a literal reference match, so an issue
  describing the work without citing the ADR number is invisible to it. The true number is ≤ 52, and
  that undercount is itself an argument for the marker rather than the search.
- Whether `partial` is being used consistently across 85 ADRs, or whether some mean "shipped with a
  known limitation we accept" — which would not need a tail at all. That distinction is exactly what
  an explicit marker would force people to state.

Refs ADR-0052, #3962, #3679, #3931, `rules.yaml: issues.open_an_issue_for`

JiRaska:
## What

A declared, enforced gate for the rule that already existed in prose: an ADR with
`delivery-status: partial` must say where its unbuilt half is tracked.

`Closes #3965`

## The two problems, kept separate

**(a) The rule was unenforced.** `rules.yaml: governance_followup` says an ADR's
actionable tail gets an issue. A `partial` ADR has one by construction. Nothing checked
it, so 85 of 241 ADRs record that half a decision is unbuilt and the backlog is silent
about most of them. Now `.github/scripts/check-adr-partial-followup.py`, declared in
`.github/gates/gates.yaml` as `adr-partial-followup` (`group: registry`, `mode: enforced`,
`selftest_expect: pass`) — a manifest entry, never an inline `run:` step in a conditional
job.

**(b) The detector cannot see cross-repo or uncited tracking.** This is why the gate is
**not an issue search**. Grepping open issues for `ADR-NNNN` goes red on ADRs that *are*
tracked, in three shapes it cannot distinguish from a real gap:

1. an issue that tracks the work without typing the number — measured at 5 false
   positives in the 13-ADR hand-triaged sample in this issue's refutation comment
   (#1915 *is* ADR-0121's unbuilt axis, verbatim, citing no ADR);
2. tracking in another repo by design — ADR-0075 says so in its own text, and a gate
   running with this repo's `GITHUB_TOKEN` cannot see `openbank-app` at all (a
   permission-shaped absence is byte-identical to a real one);
3. a `partial` that means *accepted limitation* (ADR-0186, single-region until M6) or
   *delivered differently than prescribed* (ADR-0017) — no tail exists.

It is also **not a baselineable quantity**: it is a function of the open backlog, not of
this repo. Re-derived here, the same method gives **64** untracked today against **52** in
the issue body and **55** in the refutation comment — the ADR set did not move (85 partial
in all three measurements); issues closed underneath it. Two of the issues the refutation
cited as evidence of tracking (#3352, #3555) have already closed since it was written.

The fix is an author-written front-matter key that can express all four states, so a
correctly-tracked ADR is never flagged:

```yaml
followup: "#3679 — 18 of 39 workloads still run AUTHZ_ENFORCE=false"
followup: "openbank-app#42 — the client-side SDK wiring lives in the app repo"
followup: "openbank-app — tracked in the app repo, issue number not mirrored here"
followup: "none — single-region is an accepted limitation until Milestone M6"
```

A bare repo name is deliberately allowed: the gate cannot verify a number in a repo it
cannot read, and demanding one would only buy a plausible-looking fiction. Same reason
`#N` is not checked for existence — this gate is offline, exactly as
`check-adr-registry.sh` is, and says so rather than guessing.

## Ratchet, not big-bang

`BASELINE` in the script is the ADRs that were `partial` with no marker when the gate
landed — derived from **repo state**, so it is reproducible offline and does not drift
under the backlog. That is what lets the gate be `enforced` on day one: ADR number 86
is red at PR time, and the existing set drains on its own schedule.

A stale baseline entry **fails**, in all three shapes, so the list cannot become
furniture: the ADR gained a marker, its status is no longer `partial`, or its file is
gone. Same contract as `check-pact-provider-replay.py`'s `KNOWN_UNCOVERED`.

Five entries are drained here from the issue's own hand-triage (0075, 0121, 0173, 0175,
0186), so the committed baseline is **80**, not 85 — and the first two exercise both
false-positive shapes the search could never see.

## One parser

Front matter is read through `docs/adr/lib-frontmatter.sh` — the single shared parser the
generator and validator both use. No second parser, which is the defect class
`SCHEMA.md` exists to prevent. It also buys code-about-code for free: only the delimited
block is parsed, so a `followup:` line inside a fenced example in an ADR body cannot
satisfy the rule and prose quoting `delivery-status: partial` cannot trigger it. The
self-test proves both directions end to end through the real parser rather than
asserting it.

`check-adr-registry.sh` gains `followup` as an *optional* key (`OPTIONAL_KEYS`) so its
unknown-key check does not reject it; the value grammar is validated by the new gate.

## Falsification — both ways, on the real tree, exit codes read directly (never through a pipe)

| case | findings | exit |
|---|---|---|
| control, unmodified tree | 0 | 0 |
| **new untracked `partial` ADR** (`0999-falsification-fixture.md`) | 1, named | **1** |
| baseline entry that **gained a marker** | 1, named | **1** |
| baseline entry whose ADR is **no longer `partial`** | 1, named | **1** |
| baseline entry whose **ADR file is gone** | 1, named | **1** |
| the 5 drained markers reverted (accidental, kept as evidence) | 5, all named | **1** |

`--self-test`: 19 cases, all green — the ratchet, all four legitimate marker shapes, six
grammar rejections, a malformed marker on a *baselined* ADR, all three staleness
directions, and the two code-about-code directions. Run through the real runner:

```
python3 .github/scripts/run-gates.py --only adr-partial-followup   # PASS=1, cpu 1.8s
```

`advisory-gate-registration`, `gate-graduation-guard`, `license-header-consistency`,
`workflow-run-step-size`, `adr-registry-integrity-check` and `run-gates.py --self-test`
all exit 0. `gen-index.sh` produces no diff — `followup` is not indexed, so the derived
artefacts are untouched.

## Not covered, deliberately

- Whether a cited `#N` is open, or exists. Offline gate; stated in the script header
  rather than guessed at.
- Whether `partial` is used consistently across the fleet. The marker is what forces an
  author to state which kind of `partial` they mean — it does not measure it.

### DIFF
--- a/.github/gates/gates.yaml
+++ b/.github/gates/gates.yaml
@@ -945,6 +945,37 @@ gates:
     run: |
       bash .github/scripts/check-adr-registry.sh
 
+  # ADR `partial` delivery-status declares its follow-up (issue #3965).
+  #
+  # rules.yaml's `governance_followup` says an ADR's actionable tail gets an issue. A
+  # `partial` ADR has one by construction — something shipped, something did not — and
+  # that rule lived in prose with no gate: 85 of 241 ADRs record an unbuilt half.
+  #
+  # NOT AN ISSUE SEARCH, deliberately. Grepping the open backlog for `ADR-NNNN` goes red
+  # on ADRs that ARE tracked — by an issue that never types the number (5 false positives
+  # in a 13-ADR hand-triaged sample), in another repo by design (ADR-0075 says so), or on
+  # an ADR whose `partial` means an accepted limitation with no tail at all (ADR-0186).
+  # It is also unbaselineable: the same search gave 52, 55 and 64 on three consecutive
+  # days as issues CLOSED underneath it, with no ADR moving. So the gate is offline and
+  # reads one optional front-matter key the author writes, which can express all four
+  # states. Same offline contract as check-adr-registry.sh, which likewise declines to
+  # resolve issue references.
+  #
+  # RATCHET: today's 85 markerless `partial` ADRs are a BASELINE constant in the script,
+  # so this is enforced from day one and number 86 is red at PR time. A baseline entry
+  # that no longer applies — marker written, status no longer partial, file gone — FAILS,
+  # so the list cannot rot into furniture. Same contract as check-pact-provider-replay.py.
+  # Five were drained in the landing PR, so the committed baseline is 80.
+  - id: adr-partial-followup
+    name: "ADR partial delivery-status declares its follow-up (issue #3965, enforced)"
+    group: registry
+    mode: enforced
+    selftest: |
+      python3 .github/scripts/check-adr-partial-followup.py --self-test
+    selftest_expect: pass
+    run: |
+      python3 .github/scripts/check-adr-partial-followup.py --enforce
+
   # Agent charter parity (ADR-0031 / ADR-0156): every openbank-libs/governance/agents.yaml
   # id has a docs/agents/<id>.md and vice versa.
   #
--- a/.github/scripts/check-adr-partial-followup.py
+++ b/.github/scripts/check-adr-partial-followup.py
@@ -1,0 +1,405 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) OpenBank contributors. Licensed under the Apache License, Version 2.0.
+"""ADR `partial` delivery-status must declare its follow-up (issue #3965).
+
+WHY THIS EXISTS
+    rules.yaml's `governance_followup` says an issue is opened for "the actionable
+    tail of an ADR". A `delivery-status: partial` ADR has such a tail BY
+    CONSTRUCTION: something shipped and something did not. That rule lived in prose
+    with no gate, so 85 of 241 ADRs record that half of a decision is unbuilt and
+    nothing forces anyone to say where the rest is tracked. A status inside a
+    document is not a work item: it is read once, at authoring time, and after that
+    only by whoever goes looking.
+
+WHY THIS IS NOT AN ISSUE SEARCH — the half that makes the naive gate unshippable
+    The obvious implementation greps the open-issue backlog for `ADR-NNNN` and
+    fails any `partial` ADR with no hit. That gate goes red on ADRs that are
+    correctly tracked, in four distinct ways, and it cannot tell any of them apart:
+
+      1. the tail is tracked by an issue that never types the ADR number. Measured
+         by hand over a 13-ADR sample on #3965: 5 of 13 (38%) were false positives.
+         #1915 is verbatim ADR-0121's unbuilt axis and cites no ADR at all.
+      2. the tail is tracked in ANOTHER REPO by design. ADR-0075 says so in its own
+         text ("tracked as a GitHub issue, not in this monorepo"). A gate running
+         with this repo's GITHUB_TOKEN cannot see openbank-app, and a
+         permission-shaped absence is byte-identical to a real one.
+      3. `partial` sometimes means "accepted limitation" (ADR-0186: single-region
+         until M6) or "delivered differently than prescribed" (ADR-0017: OpenBao +
+         External Secrets instead of the Vault extension). No tail exists at all.
+      4. the tail IS tracked by a citing issue — the only case the search sees.
+
+    That search is also not a stable quantity, so it cannot be baselined: it is a
+    function of the OPEN backlog, not of this repo's contents. The same method gave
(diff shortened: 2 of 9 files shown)


JiRaska:
Reviewed. **Clean — approve.** This is the shape a gate should have.

**Both axes of enforcement actually agree**, which is the thing that is usually half-done: registered as `adr-partial-followup`, `group: registry`, `mode: enforced`, **and** the `run:` passes `--enforce`. Verified the flag matters rather than assuming it: the same new untracked ADR gives `::error…` exit 1 with it, and `::warning…` exit 0 without. Either alone would be vacuous — `mode: enforced` on a script that warns and exits 0 is a permanently green gate that blocks nothing.

**Falsified on the real tree, not only on fixtures** — and in both directions:

- injected `docs/adr/0999-falsify.md` → named finding, exit 1;
- gave baselined ADR-0006 a marker → *"now carries a `followup:` marker, so its BASELINE entry is stale — remove ADR-0006 from BASELINE"*, exit 1.

That second one matters more than the first: it means the baseline can only be paid down, and a stale exemption is reported rather than quietly persisting. A gate whose exemption list can rot is one that reads as passing while covering less each month.

Self-test 19/19, all routed through the real `lib-frontmatter.sh` — one parser, no reimplementation, so the test cannot drift from the thing it tests. Counts reconcile: 85 `partial`, 5 markers written, `BASELINE` = 80. `check-adr-registry.sh` still exits 0 with `followup` added to `OPTIONAL_KEYS`.

**Scope probing found no hole.** `delivery-status: "partial"` (quoted), `Partial`, and `partial # comment` all read clean here — but each is rejected upstream by the pre-existing enforced enum check (`check-adr-registry.sh:135,211`), so none can reach this gate. Worth stating explicitly, because "my check does not handle X" is only a defect when X can actually arrive.

The acknowledged limit — `none — <20 chars>` is author-asserted and unverifiable — is stated in the script header rather than left for a reader to discover. That is the right way to ship a control that cannot fully close.

Issue #3962, opened 2026-08-07
ADR-0023 F2 (S3 Object Lock WORM) is untracked, and DataResidencyValidator's only bootable config misstates the region it validates


JiRaska:
## Summary

ADR-0023's F2 (`S3WormArchive`, S3 Object Lock COMPLIANCE) is honestly recorded as *"code-complete +
unit-tested but **not yet deployed** (awaits a provisioned S3 Object-Lock bucket), assurance 🟡"*.
Nothing tracks the deployment, and its committed defaults will not work when someone does it.

**This is not a false-claim finding.** ADR-0023 and ADR-0133 both state the position accurately — 0133
explicitly says tamper-evidence is achieved *"without an external WORM store, which is deferred"*. I
checked precisely because the pattern usually goes the other way; here the documents are right.

## What is actually wrong: the defaults are pre-broken

```yaml
# openbank-analytics-sink/src/main/resources/application.yaml
endpoint: ${ANALYTICS_WORM_S3_ENDPOINT:https://s3.eu-central-1.amazonaws.com}
region:   ${ANALYTICS_WORM_S3_REGION:eu-central-1}
bucket:   ${ANALYTICS_WORM_S3_BUCKET:openbank-analytics-worm}
```

Measured:

```
account region (every existing bucket)        : eu-north-1
openbank-analytics-worm                       : NoSuchBucket
gitops override for ANALYTICS_WORM_S3_*       : none
live analytics-sink pod env                   : ANALYTICS_SINK_TYPE=clickhouse only
                                                — no ANALYTICS_WORM_* at all
```

So the running pod uses the compiled default for `worm.type`, which the file's own comment describes
as the operator-mutable ClickHouse mirror / `LoggingWormArchive` — **not** the Object Lock path. That
is consistent with "not deployed" and is fine today.

The trap is for whoever deploys it: the endpoint and region defaults name **eu-central-1** while every
bucket in this account is **eu-north-1**. A bucket created in the account's region against these
defaults fails with `PermanentRedirect`, which is a confusing error to meet while standing up a
compliance control.

## Why this is worth an issue rather than nothing

- **Nothing tracks F2's deployment.** `gh issue list --search "WORM"` and `--search "ADR-0023"` both
  return nothing. The only record is a 🟡 inside an ADR, and this repo's own experience is that a
  status inside a document is not a work item — it is how things sit for months.
[...]

PR #4148, opened 2026-08-08
fix(analytics-sink): point the residency guard and WORM S3 defaults at the estate region

JiRaska:
## What this fixes

`DataResidencyValidator`'s committed defaults were:

```
region:  ${ANALYTICS_RESIDENCY_REGION:eu-central-1}
allowed: ${ANALYTICS_RESIDENCY_ALLOWED:eu-central-1,eu-west-1}
enforce: ${ANALYTICS_RESIDENCY_ENFORCE:true}
```

Nothing in `openbank-infra/` sets any of the three, so both defaults apply in the deployed pod. That
left **exactly one bootable configuration, and it is the one that misstates the region**: the guard
passes on `region=eu-central-1` for a warehouse in `eu-north-1`, while configuring the true region
throws `IllegalStateException` at `StartupEvent` (`enforce` defaults to true and `eu-north-1` was not
on the allow-list). The log line it emits cites GDPR Art. 44.

This PR moves both defaults to `eu-north-1`, and the WORM S3 `endpoint`/`region` defaults in the same
config block with them.

## Which source establishes the region

**`ADR-0175 — Data residency and sovereignty`, §Decision 1**: *"The estate is `eu-north-1`
(Stockholm), AWS account 265175468565. Uniform across bootstrap, sandbox-platform and web-prod."*

Not `openbank-infra/aws/README.md`. §Decision 2 of the same ADR names that README's `eu-central-1`
production pin as an inference from a condition ADR-0027 does not contain, and states it is **"not
adopted by inference"** — D1 is an open action to delete it. The issue thread contains a correction
and a correction-of-the-correction on exactly this point; the ADR is the artefact that settles it, and
it is quoted in the new test's KDoc so the next reader does not have to re-derive it.

## What reads the validated value — and therefore how urgent this is

`git grep` over the tree: `openbank.analytics.residency.region` is read **only by
`DataResidencyValidator` itself**, in the comparison and the log line. No sink, no ClickHouse client,
no S3 adapter consults it; no gitops manifest sets it.

So a wrong value is **a false assurance in a log line, not a mis-routed write** — a latent trap rather
than data corruption. That is why this is a plain `fix`, not an incident. The KDoc now records the
deeper limit as well: this validator compares a config string to a config string and never observes
the region it actually runs in (IMDS, `AWS_REGION`, the bucket's `LocationConstraint`), so it can only
detect a self-inconsistent config. Making it observe reality is a larger change and not attempted here.

## Red-then-green

New `DataResidencyDefaultsTest` asserts the literal region **values** parsed out of the committed
`application.yaml` and read off the `@ConfigProperty` annotation defaults — never `isNotNull()`, which
would pass against precisely the wrong value.

Against the **previous** config (fix stashed, `build/test-results` removed first), from the JUnit XML:

```
TEST-...DataResidencyDefaultsTest.xml tests=5 failures=3 errors=0
   the committed WORM S3 defaults name the estate region()   -> FAIL: expected "eu-north-1" but was "eu-central-1"
   a region outside the allow-list still aborts boot()       -> PASS   (control, must stay green)
   the committed defaults boot with enforcement on()         -> PASS   (this is the point: the wrong config is the bootable one)
   the committed residency defaults name the estate region() -> FAIL: expected "eu-north-1" but was "eu-central-1"
   the estate region is on the committed allow-list()        -> FAIL: ["eu-central-1","eu-west-1"] does not contain "eu-north-1"
```

With the fix:

```
TEST-...DataResidencyDefaultsTest.xml tests=5 failures=0 errors=0   (all five PASS)
MODULE TOTAL: tests=71 failures=0 errors=0 skipped=0
```

Local gate — all three tasks confirmed present in the task list, not inferred from a green tail:

```
> Task :openbank-analytics-sink:detekt
> Task :openbank-analytics-sink:ktlintTestSourceSetCheck
> Task :openbank-analytics-sink:ktlintCheck
> Task :openbank-analytics-sink:test
BUILD SUCCESSFUL
```

## Scope — deliberately not in this PR

- **ADR-0023 F2's untracked deployment** (the issue's other half). Provisioning the Object-Lock bucket,
  flipping `openbank.analytics.worm.backend=s3` (build-time `@IfBuildProperty`, so a gitops env var can
  never activate it) and proving `S3WormArchive`'s SigV4 path against real S3 are deployment work with
  no code change here — and the refutation pass on the issue found a harder blocker than the region
  string: the signer emits no `x-amz-security-token`, so it cannot use any temporary credential, and
  the account has zero IAM users. Adding a tracking marker to the ADR would be half-doing it. The
  general case is already **#3965** (open), which cites this issue as one of its worked examples;
  F2's deployment belongs there or in its own issue with an owner and a bucket, not in a config PR.
- **The rest of ADR-0175 D2.** `eu-central-1` also defaults in `openbank-libs-runtime`'s shared
  `S3ObjectStore`, `openbank-document-service` and `openbank-customer-edge`. That is a fleet-wide
  sweep across four modules with its own review surface; this PR touches only the service the issue
  is filed against. (Note the trap has not fired in production: customer-edge, the one service running
  the S3 backend, overrides `OBJECTSTORE_S3_REGION=eu-north-1` in gitops.)

## Risk

No runtime behaviour change in the deployed pod beyond the log line now naming the region the
warehouse is actually in. `openbank.analytics.worm.backend` stays unset, so the Object Lock path is
still inactive. No migration, no API change, no `openapi.yaml` impact.

Refs #3962 · Refs #3965 · Refs ADR-0175 (§Decision 1, §Decision 2, D2), ADR-0023 (F2, F9)

### DIFF
--- a/.github/scripts/check-external-feeds.py
+++ b/.github/scripts/check-external-feeds.py
@@ -152,7 +152,7 @@ NOT_PROBED = [
     ("https://api.groq.com/openai/v1", "authenticated LLM gateway (ADR-0139); needs a key"),
     ("https://api.deepinfra.com/v1/openai", "authenticated LLM gateway; needs a key"),
     ("https://integrate.api.nvidia.com/v1", "authenticated LLM gateway; needs a key"),
-    ("https://s3.eu-central-1.amazonaws.com", "AWS endpoint, reached with SigV4 credentials"),
+    ("https://s3.eu-north-1.amazonaws.com", "AWS endpoint, reached with SigV4 credentials"),
     ("https://kc.open-bank.tech/realms/openbank-customers", "our own Keycloak realm, covered by its own probes"),
     ("https://pid.open-bank.tech", "our own PID issuer, covered by its own probes"),
 ]
--- a/openbank-analytics-sink/src/main/kotlin/com/openbank/analytics/infrastructure/DataResidencyValidator.kt
+++ b/openbank-analytics-sink/src/main/kotlin/com/openbank/analytics/infrastructure/DataResidencyValidator.kt
@@ -18,18 +18,33 @@ import org.jboss.logging.Logger
  * ClickHouse region is not on the allow-list, turning a silent mis-deployment into a deploy-time error.
  *
  * Configuration:
- *  - `openbank.analytics.residency.region` — the region this warehouse runs in (e.g. `eu-central-1`).
- *  - `openbank.analytics.residency.allowed` — comma-separated allow-list (default `eu-central-1,eu-west-1`).
+ *  - `openbank.analytics.residency.region` — the region this warehouse runs in (default `eu-north-1`).
+ *  - `openbank.analytics.residency.allowed` — comma-separated allow-list (default `eu-north-1`).
  *  - `openbank.analytics.residency.enforce` — when true (default), a violation aborts startup; when
  *    false it only logs (useful for local dev / `%dev`).
+ *
+ * `eu-north-1` is the estate region decided by ADR-0175 §Decision 1. It is decided *there* and not in
+ * `openbank-infra/aws/README.md`, whose `eu-central-1` production pin ADR-0175 §Decision 2 names as an
+ * inference from a condition ADR-0027 does not contain, and explicitly does not adopt.
+ *
+ * Until #3962 these defaults were `eu-central-1` with an allow-list of `eu-central-1,eu-west-1`, which
+ * left exactly one bootable configuration and it was the one that misstates where the warehouse is:
+ * the guard passed on `region=eu-central-1` for a warehouse in `eu-north-1`, while configuring the real
+ * region aborted boot (`enforce` defaults to true, and `eu-north-1` was not on the list).
+ *
+ * KNOWN LIMIT (deliberate, not an oversight): this compares a config string to a config string. It
+ * never observes the region it is actually running in (IMDS, `AWS_REGION`, the bucket's
+ * `LocationConstraint`), so it can only detect a self-inconsistent configuration, not a real residency
+ * violation. Nothing else in the service reads `residency.region`; a wrong value is a false assurance
+ * in the log line, not a mis-routed write.
  */
 @ApplicationScoped
 class DataResidencyValidator {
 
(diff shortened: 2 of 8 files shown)


JiRaska:
Reviewed. **Clean — no defect found.**

**It fails CLOSED, which is the right direction for a residency control.** `onStart` lowercases and trims, filters empties out of the allow-list, and throws `IllegalStateException` when the region is not in it (`enforce` defaults true). An empty `ANALYTICS_RESIDENCY_REGION` *or* an empty allow-list both abort boot rather than pass. The observer is on `StartupEvent`, which forces bean creation — so the `@ApplicationScoped`-is-lazy trap (an `init{}` guard that never runs until the first request, which for a rarely-called path can be never) does not apply here.

**The honesty about what it does not do is the best part of the PR.** `openbank.analytics.residency.region` is read only by `DataResidencyValidator` — nothing in `openbank-infra/` sets `ANALYTICS_RESIDENCY_*` or `ANALYTICS_WORM_S3_REGION`, so this corrects a false assurance in a log line, not a routing decision. The KDoc records the deeper limit outright: the validator compares config-string to config-string and never observes `AWS_REGION`, IMDS, or the bucket's `LocationConstraint`. That is the right framing — a control that says what it cannot see is worth more than one that reads as stronger than it is.

**Assertions are literal, not vacuous:** `isEqualTo("eu-north-1")` against both the YAML default and the `@ConfigProperty` annotation default, plus a falsification control (`us-east-1` must still throw with `"data residency violation"`). The PR explicitly rejects `isNotNull()`, which matters — this repo has a documented class of defect where `isNotNull()` passed against a 1970 placeholder for years.

**One consequence worth naming, not a defect:** the allow-list narrows from `eu-central-1,eu-west-1` to `eu-north-1` alone. Nothing overrides it today and every bucket is `eu-north-1`, so nothing deployed breaks — but a future eu-west-1 deployment now **fails boot** rather than warning. Intentional for this control, and worth knowing before someone opens a second region.

The one out-of-module edit — the `NOT_PROBED` S3 endpoint literal in `check-external-feeds.py:155` — is consistent with the config change and carries no behaviour.

Issue #4023, opened 2026-08-07
Local docker templates still base on alpine/musl after the fleet moved to glibc


JiRaska:
## What

Two local build templates still base on alpine (musl) after the fleet moved to glibc in #3618/#3629:

```
openbank-infra/docker/services/standalone-service.Dockerfile:27   eclipse-temurin:25-jre-alpine@sha256:28db6fdf…
openbank-infra/docker/services/root-service.Dockerfile:20         eclipse-temurin:21-jre-alpine@sha256:3f08b138…
```

They are **not dead files** — `openbank-infra/docker-compose.yml` references them, so this is the local development build path. `root-service.Dockerfile` is additionally on JRE **21** while the fleet runs 25 (its header already marks it deprecated).

## Why it matters, and why it is not urgent

Nothing built from these templates bundles a native library today, so nothing is broken right now. The hazard is divergence: after #3618 the deploy path is glibc and the local path is musl, so a service that later gains a native dependency — the exact situation that made `fraud-service`'s ONNX model score nothing for its whole life (#3354) — would work in CI and fail locally, or the reverse. A build path that differs from the deploy path in libc is a difference that only shows up as a confusing bug much later.

`.github/scripts/verify-image-native-libs.py` (added by #3629) runs `ldd` over bundled `.so` files inside the real base before a push, but it reads the base from `.github/workflows/Dockerfile.deploy` — it does not see these templates.

## Where this came from

#3709 proposed the fleet move via `ubi10-minimal` and covered these templates plus `Dockerfile.scanner`, `openbank-security-scanner/Dockerfile.prebuilt` and `openbank-notification-service/Dockerfile.runtime`. It was closed in favour of #3618's `eclipse-temurin:25-jre` base, which is the narrower change. Filing this so the coverage difference is not lost with the closed PR — that is the whole reason this issue exists rather than a comment on a dead thread.

## Suggested scope

- Move both templates to the same glibc digest the fleet uses, or state in each header why the local path deliberately differs.
- Check `Dockerfile.scanner`, `openbank-security-scanner/Dockerfile.prebuilt`, `openbank-notification-service/Dockerfile.runtime` for the same residue.
- Consider whether `root-service.Dockerfile` (deprecated, JRE 21) should be deleted rather than updated.

Refs #3354, #3618, #3629, #3709

JiRaska:
## Confirmed on `origin/main` today

```
openbank-infra/docker/services/standalone-service.Dockerfile:27
[...]

PR #4153, opened 2026-08-08
security(admin-ui): pin nanoid to 3.3.17 for CVE-2026-67213

JiRaska:
## What

Pins `nanoid` to **3.3.17** to clear **CVE-2026-67213** (HIGH).

## Why it is urgent

**Trivy fails on `main` itself**, so every open PR inherits a red `Trivy` check that has nothing to do with its own diff. Measured on `main`'s own `Security scan` run:

```
nanoid  CVE-2026-67213  HIGH  fixed  3.3.16 → 3.3.17, 5.1.6
"nanoid (Nano ID) before 5.1.6 contains an infinite loop"
```

That is why #3880 and #3926 went red on `Trivy` immediately after their engagement-service failure was cleared — the red simply moved to the next inherited fault.

## Why an override, and why 3.3.17 rather than 5.1.6

`nanoid` is **transitive** — no direct dependency declares it — so an override is the only lever, alongside the `postcss` and `sharp` pins already in `overrides`.

3.3.17 is the fix on the 3.x line the tree already resolves. Taking 5.1.6 would be two majors of API risk on a package nothing here calls directly, for the same CVE outcome.

## Verified with the gate's own tool, not inferred from a version string

```
trivy fs --scanners vuln --severity CRITICAL,HIGH --ignore-unfixed  openbank-admin-ui
→ package-lock.json  npm  0
```

and the lockfile now resolves `node_modules/nanoid` to `3.3.17`. A version number in a manifest is a claim; the scanner's verdict is the check.

## One thing in the diff I did not set out to change

The lockfile's **root version was `0.102.0`** while `package.json` and `version.txt` both say `0.103.1`. `npm install --package-lock-only` corrected that drift while regenerating.

Flagging it rather than letting it ride silently: it is a real pre-existing inconsistency, the new value is the correct one, and nothing was enforcing it — the version-sync gate compares `version.txt` to `package.json` and never looks at the lockfile's copy.

Refs #4023

### DIFF
--- a/openbank-admin-ui/package-lock.json
+++ b/openbank-admin-ui/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "openbank-admin-ui",
-  "version": "0.102.0",
+  "version": "0.103.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "openbank-admin-ui",
-      "version": "0.102.0",
+      "version": "0.103.1",
       "dependencies": {
         "@hookform/resolvers": "5.5.7",
         "@radix-ui/react-dialog": "1.1.23",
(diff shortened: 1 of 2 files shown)


JiRaska:
**This PR is the single unblocker for four others.** Measured, not assumed:

- `Trivy` on this branch: **SUCCESS**
- `Trivy` on #3880, #3926, #4133 and #4142: **FAILURE**, and all four fail the *identical* step — `Trivy fs scan`, read from each job's step conclusions rather than from a log grep.

None of those four touches `openbank-admin-ui/package-lock.json`. They are red because `main` is: `nanoid` 3.3.16 sits in the tree there with a fixed HIGH against it, so every PR inherits the finding.

`nanoid` on `origin/main` @ `f6efd4d31` is still **3.3.16**, so the class stays open until this lands.

PR #4155, opened 2026-08-08
ci(github): watch main's own CI conclusion and escalate its red

### LINKED ISSUE #4019: main went red twice today and two PRs merged onto it — nothing watches main's own CI conclusion
## Summary

`main` went red twice today on the required `Validate manifests` check, stayed red for ~34 and
~29 minutes, and **two PRs merged onto the second broken main**. Nothing escalated in either case —
no issue, no alert, no comment. Both were found only because someone happened to be reading CI for
an unrelated PR, and both were fixed by hand.

## Measured

`gh run list --branch main --workflow CI`, chronological:

```
07:45:56  3208109c  failure   <- #3771 (case-coordinator charter, stale derived artifacts)
07:52:36  d16dc102  failure
08:20:20  6e240006  success   <- #4003, the fix

08:34:14  ee974ea3  failure   <- #3686 (finrep shared-base pilot)
08:36:44  841d20e7  failure   <- #3613 merged onto red main
08:37:25  fbde2b65  failure   <- #3772 merged onto red main
09:03:34  f2bb3391  success   <- #4017, the fix
```

Causes, for the record — they were different in kind, which is part of the point:

- **#3771** added an agent charter without regenerating the two artifacts derived from it. Two
  enforced drift gates, genuinely red, genuinely the PR's fault.
- **#3686** moved the management block into the shared `openbank-libs-runtime` base
  `application.yaml`. The probe-port gate reads only a service's own file, so it reported a port
  that *is* open as missing — a **false** red, and one that would have grown by one per service as
  #3672 rolls the base out.

So the escalation gap is not "someone merges bad code": it catches an honest omission and a gate
that became wrong, equally.

## Why nothing stopped the two merges

Required status checks are evaluated against **each PR's own merge ref**, not against `main`'s
current state. #3613 and #3772 were branched before the breakage, so their checks passed
legitimately and both were mergeable while `main` itself was failing. That is correct per-PR
behaviour and there is no per-PR fix for it — the missing piece is that nobody is told `main` broke.

Consequence beyond the noise: every PR that merges `main` in *inherits* the red, so the window also
silently blocks anyone who rebases during it. That is how #3633 spent most of the morning red for
two reasons that were never its own.

## What is missing

There is no watcher on `main`'s own CI conclusion. `deploy-drift-watch.yml` escalates a drifted
deploy and `fleet-attestation.yml` raises an issue when it fails, so the pattern exists in this repo
— it just was never applied to the branch everything is cut from.

Suggested shape, matching the existing ones: a `workflow_run` listener on CI/`completed` for
`main` that opens (or refreshes) a single issue naming the first failing gate and the commit that
introduced it, and closes it when `main` goes green again. Single issue, refreshed rather than
stacked — the #3891 lesson about reopening rather than piling up new ones applies.

## Not this

- Requiring branches to be up to date (`strict_required_status_checks_policy`) would force a rebase
  on every PR whose base moved. At this merge rate that is a large cost for a problem that is really
  about *notification*, not about admission.
- A merge queue is unavailable: the repo is personally owned and the ruleset returns
  `422 Invalid rule 'merge_queue'` (#1465).

Related but distinct: #3955 (`all-green` is scoped to service builds, so it can report SUCCESS after
`Validate manifests` has already failed) — that is about a PR's own signal; this is about `main`'s.

JiRaska:
Closes #4019

## The gap

`main`'s own CI conclusion is observed by nothing. Every check in this repo is about a **PR** — a red push-triggered run on `main` has no PR to carry it, no reviewer, and no notification. It is a red dot in the Actions tab, and the next PR opened against that commit **inherits a failure it did not cause**.

I re-established this on `origin/main` rather than taking the issue's word for it:

- **One** `workflow_run` listener exists in the whole repo — `auto-retry-cancelled.yml`. It listens to `Services CI` / `Security scan` / `Dependency submission`, re-runs **spot-killed** runs, does not filter on `head_branch`, and escalates nothing.
- **No** scheduled job reads a CI conclusion for `main`. The crons are `auto-deploy`, `deploy-drift-watch`, `dependency-submission`, `services-ci`, `security`, `pitest`, `external-feed-watch`, `fleet-attestation`, and friends.
- The near-miss is worth naming on its own: `deploy-drift-watch.yml`'s `name:` is literally **"Deployed == main watch"**. It compares the **deployed image** to `main` and never inspects whether `main`'s CI passed — so a red commit that deployed reads as perfectly in sync.

## The evidence, verified independently

`main` was red four times on 2026-08-07/08, in three independent ways:

| workflow | sha | first failing gate |
|---|---|---|
| `Services CI` | `6b3594b2b` | `build (openbank-engagement-service)` |
| `Services CI` | `98841b41d` | `build (openbank-mcp-service)` |
| `CI` | `9240b133b` | `Admin UI build / Test (unit + integration)` |
| `CI` | `b7851d689` | `Admin UI build / Test (unit + integration)` |
| `Security scan` | `75e94620c` | `Trivy fs scan` — nanoid 3.3.16, CVE-2026-67213 HIGH |
| `Security scan` | `f6efd4d31` | `Trivy fs scan` — same, **red on `main` as this PR is opened** |

Each was found only because a human happened to sweep the Actions tab. Two PRs merged onto that `main` meanwhile.

## What this adds

The two-lane shape already established by `external-feed-watch` and `deploy-drift-watch`.

**`.github/scripts/check-main-red-watch.py`**

- `--check-declaration` — offline, deterministic, ~0.1 s, **blocking**. R1: every workflow with `push:` on `main` and no `paths:` filter must be in the watch list (it runs on every commit, so its red *is* `main` red). R2: a path-filtered one must be watched or in `NOT_WATCHED` with a reason. R3: the watcher must filter on `head_branch`. R4: the watcher must query the attempt-scoped jobs endpoint. A stale `NOT_WATCHED` entry fails too.
- `--inspect-run` — classifies one completed run. Exit `0` green / `1` **could not answer** / `2` red. "Could not answer" is a distinct state on purpose: a completed run with zero jobs is what an unparseable workflow, a `concurrency.group` that stopped a job being created, and a permissions failure all look like, and none of those is green.
- `--self-test` — 30 cases.

**`.github/workflows/main-red-watch.yml`** — a `workflow_run` listener on 17 workflows (8 unconditional + 9 post-merge guards), gated to `head_branch == 'main' && event == 'push'`. It **escalates**: one `main-red` issue per watched workflow, refreshed in place, reopened if hand-closed while `main` is still red, and closed by the watch itself when that workflow next goes green. It does not merely go red — a `workflow_run` job's red would reproduce the exact blind spot it exists to close.

**`gates.yaml: main-red-watch-declaration`** (`registry`, enforced). The binding copy lives in `Validate manifests` because a `workflow_run`/`paths:`-triggered workflow can never be a required context; the workflow's own `pull_request` lane is the fast echo.

## The two details that decide whether it works at all

**Attempt scoping.** The verdict comes from `/actions/runs/<id>/attempts/<n>/jobs`, never the unscoped `/actions/runs/<id>/jobs` — that returns the **latest** attempt, so one hand-re-run makes the watcher answer about a different run than the event fired for, silently. R4 greps the workflow for it, and the self-test proves the classifier genuinely disagrees between attempts (attempt 1 red, attempt 2 green, and the recorded fetch URL never matches `/runs/\d+/jobs`).

Falsified against the **live** API, not just a fixture — asking for an attempt that does not exist is an error, not a fallback:

```
$ check-main-red-watch.py --inspect-run --run-id 31234270871 --attempt 99
::error::main-red-watch could not answer: gh api .../attempts/99 failed: gh: Not Found (HTTP 404)
EXIT=1
```

**Step conclusions, never the log.** A job log contains the step's own `run:` script, so a grep matches strings that never executed. `classify_run` is handed job dicts and reads `.steps[].conclusion`; it has no log and cannot get one. The self-test feeds it a job whose log body is `Total: 1 (HIGH: 1, CRITICAL: 0)` plus an echoed `::error::` from a script that never ran, with every step conclusion `success`, and requires the verdict to be **green**.

## Falsification

Not "it passed". Here is what it **prints**.

**Real red #1** — `Services CI` run `31191031034`, `main` @ `6b3594b2b`:

```
main is RED: Services CI @ 6b3594b2b (attempt 1) -- 2 failing job(s).
  first failing gate: build (openbank-engagement-service) / openbank-engagement-service / Build + test (:openbank-engagement-service)
  - build (openbank-engagement-service) / openbank-engagement-service [failure]: Build + test (:openbank-engagement-service)
  - all-green [failure]: Verify no service failed
EXIT=2
```

Note it names the **engagement-service build**, not `all-green`. `all-green` also failed, and it is the aggregate — a watcher that reported the last failing job would have reported the symptom. Failing jobs are ordered by `started_at`.

**Real red #2** — `Security scan` run `31234270871`, `main` @ `f6efd4d31`, red right now:

```
main is RED: Security scan @ f6efd4d31 (attempt 1) -- 1 failing job(s).
  first failing gate: Trivy / Trivy fs scan
  - Trivy [failure]: Trivy fs scan
EXIT=2
```

**Real green** — `Services CI` run `31234270967`, the **same sha** `f6efd4d31`:

```
main is green: Services CI @ f6efd4d31 (attempt 1).
EXIT=0
```

Same commit, two workflows, opposite verdicts. That is why the escalation is keyed per watched workflow: `Services CI` going green says nothing about Trivy, and a single shared issue would have been closed by the wrong event.

**The declaration gate, falsified against the real tree.** Removing one entry from the watch list:

```
$ # "Security scan" deleted from main-red-watch.yml's workflows: list
::error::main-red-watch declaration: R1 'Security scan' (security.yml) pushes to main
  but is neither in .github/workflows/main-red-watch.yml's `workflows:` list nor
  declared in NOT_WATCHED with a reason.
EXIT=1
$ # restored
main-red-watch: every push-on-main workflow is watched or declared.
EXIT=0
```

This half exists because the coverage set is otherwise a hand-written list inside a trigger, and this repo has already been bitten by a gate whose scope is a hand-kept list of the thing it checks — `pact-drift-check.yml` was green about modules it never regenerated. R1/R2 make the list derived from the workflow files, so a new push-on-main workflow fails the gate until someone watches it or writes down why not.

## Local verification

```
run-gates.py --only main-red-watch-declaration   PASS  (self-test 30/30, then the gate)
run-gates.py --group lint                        PASS=6
run-gates.py --group registry                    PASS=19 WARN=1
run-gates.py --only workflow-run-step-size        PASS  (largest run: step 668 chars)
actionlint .github/workflows/main-red-watch.yml   clean
```

The one `registry` WARN is `operator-write-naming`, an advisory gate that is **pre-existing on `main`** — confirmed by stashing this branch's changes and re-running it alone.

All reasoning lives in the script header, not in a `run:` block: the largest `run:` step here is 668 chars against the enforced 19000 limit.

## Not in scope

`strict_required_status_checks_policy`, and the `NOT_WATCHED` exclusions — `Auto deploy`, `Admin-UI deploy`, `Platform OpenTofu`, `Substrate OpenTofu`, `Runner image`, `OpenSSF Scorecard` — each of which carries its reason in the script and is a lane whose red this deliberately does not escalate.


JiRaska:
**Both red checks here are inherited from `main`, not caused by this PR.**

- **`Customer-app dossier`** — `app-status.json` is derived from the *separate, private* `openbank-app` repo and the job regenerates it against that repo's default branch with no pinned ref. It goes stale when openbank-app **ships**, never because of anything a PR here does. #4053 (OPEN, CLEAN) scopes the PR-time comparison to drift a PR could actually have authored; the artifact itself still needs a regenerate-and-publish, which is the chore last done in `6a7d8f4da`.
- **`Trivy`** — `nanoid` 3.3.16 / CVE-2026-67213 HIGH is still in `main`'s tree. #4153 (OPEN, CLEAN) pins it to 3.3.17; its own `Trivy` check is SUCCESS, and the four other PRs red on this fail the *identical* step, `Trivy fs scan`.

Neither diff here touches `package-lock.json` or `app-status.json`.

This is the same class three times in one night: a PR going red for a fault it did not cause, which is precisely what #4019 exists to make visible.

JiRaska:
Reviewed. Strong PR — the escalation shape is right and most of it is genuinely falsified. **One defect, and it is the repo's own documented failure mode in its quietest direction.**

### R4 validates a comment, not the code

`/attempts/` appears in `main-red-watch.yml` exactly once — **line 26, inside the header comment**. It appears in no executable YAML, because the real call is in `check-main-red-watch.py:282`. And R4 (`:246`) greps `raw`, the workflow text:

```python
if "/attempts/" not in raw:
```

So the rule that exists to guarantee the attempt-scoped endpoint is satisfied by the sentence *describing* the endpoint. Falsified in both directions:

- Broke the real call (`/attempts/{attempt}/jobs` → `/runs/{id}/jobs`), left the comment: `--check-declaration` printed *"every push-on-main workflow is watched or declared"*, **exit 0**, and `--self-test` **exit 0**. The watcher was genuinely unscoped and both gates said fine.
- Deleted only comment line 26, code correct: `::error::… R4 … does not query the attempt-scoped jobs endpoint`, **exit 1**.

Exactly backwards. This is the documented shape — *a check greps a file for the string it wants and matches the comment that explains why the string is there* — and it is the direction that reads as a pass, so nothing announces it.

The self-test cannot backstop it: `fake_fetch` (`:479-483`) hand-writes the attempt-scoped URL and then asserts about its own string (`:491`). Symmetric, so it stays green against a broken real function — the same vacuity as deriving both sides of a pact assertion from the client annotation.

Verified fix, one line of test: monkeypatch `gh_api` and call the real `fetch_jobs_scoped`. Ran it — it records `repos/o/r/actions/runs/1/attempts/1/jobs?per_page=100`, and goes red against the patched version.

**R3 (`head_branch`) is the same construct** and passes today only because no comment happens to mention the string. Deleting the `if:` clause does redden it, so it works — but it is one explanatory comment away from the same failure.

### Everything else checks out

- **Registration**: `gates.yaml: main-red-watch-declaration`, `group: registry`, `mode: enforced`, `selftest_expect: pass` — all match reality, with the binding copy in `Validate manifests` and the workflow lane as the echo. That is the right shape for a check that must run unconditionally.
- **R1 falsified on the real tree**, not fixtures: injecting `zz-probe.yml` produced a named error and exit 1. Self-test 30/30, including "log full of failure prose but steps green ⇒ green" and the unanswerable zero-jobs state.
- **Trap (b) is genuinely closed**: `classify_run` is handed job dicts and reads `.steps[].conclusion` — it never sees a log, so the echoed-`run:`-script problem cannot arise. Structural, not a grep that happens to be careful.
- **Escalation is raise-or-refresh**, not per-run noise: one `issues.create`, keyed on an HTML marker rather than the title (so an edited title cannot orphan it), with refresh, reopen-within-14-days and close-on-green, per watched workflow.
- **The coverage set is derived from the workflow files**, so an omitted workflow goes red rather than passing silently — the scope hole this repo keeps re-encountering is closed by construction here.

Worth fixing R4 before merge: the gate's whole purpose is to stop a re-run's green from answering for a different run, and as written it cannot see that happen.

JiRaska:
Pushed the R4 fix to this branch — it now **exercises** the fetch instead of grepping the workflow for prose.

**The change.** `check_attempt_scoped_fetch()` stubs `gh_api`, calls the real `fetch_jobs_scoped`, and asserts on the URL the function actually builds. Offline; nothing is sent. Grepping `check-main-red-watch.py` instead would have been the same defect one file over — the point is that the rule now reads the request, not a sentence about the request.

The self-test drives the real function the same way, and falsifies R4 by swapping in an unscoped fetch rather than by handing `check_declaration` a doctored `raw`. The old `fake_fetch` hand-wrote the attempt-scoped URL and then asserted about its own string, which is why it stayed green against a broken real function.

**Same two probes as before, now inverted as required:**

| | before | after |
|---|---|---|
| break `fetch_jobs_scoped` → `/runs/<id>/jobs`, leave the comment | decl **exit 0**, self-test exit 0 | decl **exit 1**, quoting `'repos/o/r/actions/runs/1/jobs?per_page=100'` |
| delete only the yml comment, code correct | decl **exit 1** | decl **exit 0**, 0 R4 findings |

Self-test 33/33, with two controls added so the new rule cannot be green the easy way: *"deleting the yml comment does NOT flag a correct fetch"* (over-block) and *"the real attempt-scoped fetch is clean"* (not always-on). `run-gates.py --only main-red-watch-declaration` exits 0.

**Left alone deliberately:** R3 (`head_branch`) is the same whole-file-grep construct. It works today only because no comment happens to mention the string — deleting the `if:` clause does redden it — so it is one explanatory sentence away from the same failure. Worth the same treatment, but not in a diff about R4.

Everything else in this PR I reviewed as sound and have not touched: registration and `selftest_expect` match reality, R1 is falsified on the real tree, `classify_run` is structurally incapable of reading a log (so the echoed-`run:`-script trap cannot arise), the escalation is marker-keyed raise-or-refresh rather than per-run noise, and the coverage set is derived from the workflow files so an omitted workflow goes red rather than passing silently.

JiRaska:
R3 fixed too — same defect class as R4, one rule over.

R3 tested `"head_branch" not in raw` against the whole workflow text, so any comment naming the field satisfied it. It escaped R4's fate only because no comment happened to mention the string, which is not a property worth relying on: this file's header discusses `head_branch` at length, and one more explanatory sentence would have silently disarmed it.

**The fix is different from R4's, because the situation is.** `head_branch` genuinely belongs in the yml — unlike R4's endpoint, which is built in Python — so there is nowhere else to look. Instead R3 stops looking at the *file* and asks the *construct*: `check_head_branch_filter` parses the document and reads `jobs.*.if`. The YAML parser has already dropped comments, so prose cannot answer for behaviour in either direction. The finding also prints the conditions it actually saw, which makes a false positive self-explaining rather than a puzzle.

**Falsified against the real workflow, both ways:**

| | result |
|---|---|
| removed the real `head_branch` filter, **added** a comment naming it | **exit 1** — and the error lists the two job conditions it saw. The old rule would have passed: the substring was still in the file. |
| restored the filter, **deleted** every comment naming it | **exit 0**, 0 R3 findings |

Self-test gains three cases, including the over-block control *"a COMMENT naming head_branch does not satisfy the rule"* — the guard against re-shipping R4's defect is a test rather than a comment.

**One fixture had to change, and it is worth flagging.** `raw_ok` was `"head_branch == 'main'\n/attempts/${{ x }}/jobs\n"` — a string that merely contained the substrings the old rules grepped for, not a workflow at all. Both rules parse their input now, so a fixture like that could only ever prove that a substring was absent from an arbitrary string. Replaced with a real minimal workflow document.

Self-test 35/35, `run-gates.py --only main-red-watch-declaration` exit 0, `--check-declaration` exit 0 on the real tree.

JiRaska:
Note on the red `Validate manifests` here: **it is not this PR's**, and the R3/R4 changes are not implicated.

The failing gate is the duplicate-YAML-key guard, on `openbank-engagement-service/src/main/resources/application.yaml`. Counted:

```
authz: keys on this branch   1   (correct)
authz: keys on origin/main   2   (duplicated)
```

CI evaluates `refs/pull/4155/merge`, so the branch inherits `main`'s duplicate. It went red at 13:21Z, one minute after my R3 push — recent enough to look like mine, which is why I checked the timestamps and the key counts rather than assuming.

**How the duplicate came back:** #4128 added an `authz.enforce` block under `%test`, #4072 added a second, #4160 merged them into one — and then **#4106** (`fb5c455bc`) re-added the second as a **pure insertion** (+22 lines, 0 deletions), stale comment included (*"`authz.enforce` has no key elsewhere in this file"*, which stopped being true the moment the first duplicate landed). Its base predated #4160, so git applied it cleanly on top of the fix. No conflict, no warning — the deduplication was simply undone.

Worth noting the system worked as designed at every step: SnakeYAML keeps the last of a duplicated mapping key, so behaviour never changed and nothing but the guard could see it — and the guard caught it both times, immediately. The cost is `main` being red between reintroduction and fix.

**#4184 fixes it** (one file, one `authz:` key, and it does not carry the stale claim forward — I checked). Merging that unblocks this PR; nothing here needs changing.

For completeness on this branch: self-test 35/35, `--check-declaration` exit 0, `run-gates.py --only main-red-watch-declaration` exit 0, and both R3 and R4 falsified against the real workflow in both directions.

PR #4156, opened 2026-08-08
fix(copilot): capture the erasure identity at write time so PARTY_ERASED actually deletes

### LINKED ISSUE #3881: copilot erasure consumer cannot receive PARTY_ERASED: no KafkaUser, no Kafka config — and the partyId/sub key may not match once wired

JiRaska:
## What this fixes

Item 2 of #3881 — the half the issue called "the more dangerous half", and the one that survived
#3885.

`PARTY_ERASED` carries `partyId`. Copilot keyed conversation history on the OIDC `sub`
(`CopilotChatResource.customerSubject()`). Those are not the same value: measured against the
deployed customers realm and reported on the issue, `sub` equalled `party_id` for **0 of 35** users,
because `WebAuthnKeycloakClient.ensureUser` never sets the Keycloak user `id`, so Keycloak mints a
random UUID and `party_id` survives as a separate attribute.

So since #3885 wired the Kafka half, the consumer subscribes, receives the event, runs the delete,
deletes nothing, and logs `erased 0 copilot conversation(s)` at INFO. That is worse than the
originally-filed unwired state: it is a GDPR Art. 17 control that reports success while free-text
transcripts stay on disk and in every base backup, detectable only by querying the table.

## Why the fix is on the write path, not in the consumer

Resolving `partyId` -> `sub` at delete time cannot be made to work: by erasure time the Keycloak user
is gone, so the mapping no longer exists to be looked up. The JWT carries both values at write time,
which is the only moment they are both available.

- `CopilotChatResource.erasureIdentity()` — the `party_id` claim, `sub` as fallback (the resolution
  `RateLimitFilter` and `WebAuthnKeycloakClient` in customer-edge already use).
- `V2__conversation_history_party_id.sql` — nullable `party_id` column + index, with a rollback note.
- `ConversationStore.deleteForParty(partyId)` (renamed from `deleteForCustomer`, whose name asserted
  the very equality that does not hold) matches **either** identity column:
  `partyId = ?1 or customerId = ?1`. The `customerId` arm is not belt-and-braces — it is the only
  thing that reaches rows written before V2, and the ADR-0069 `sub == partyId` case. Both are
  UUID-valued identity columns for the same person, so the union cannot widen across customers.

**`party_id` is never a lookup key.** `customerId` alone still decides which conversation a customer
resumes. The issue explicitly ruled out re-keying history on `party_id` as an identity decision
needing its own PR — this change deliberately does not make it, and moves or orphans nothing.
Pre-V2 rows pick a party id up on that customer's next message; no backfill job is needed and none
is claimed.

The other two adapters follow: `InMemoryConversationStore` records it on the entry, and
`RedisConversationStore` writes a content-free `copilot:conv:party:<partyId>:<conversationId>`
pointer under the same TTL rather than shipping a knowingly-broken erasure in a non-default adapter.

## Red then green — measured, not asserted

The four tests in `ConversationErasureIT` all passed against the broken code, because each erases
using the very id it seeded with. **An erasure test whose two identities are the same string cannot
see this class of bug at all.** So `PartyErasureIdentityIT` seeds `sub != partyId` deliberately, and
every verdict is a direct JDBC query — reading through the store cannot distinguish "deleted" from
"filtered by `expires_at`", which is the distinction the issue turns on (item 3).

RED was produced by reverting only the *behaviour* and keeping the signatures, so the failure is
about the defect and not about compilation. From the JUnit XML, not the console:

```
tests=4 failures=3 errors=0 skipped=0
 - an erasure keyed on partyId removes history stored under a different sub
     FAIL: expected 2L but was 0L      <- the "erased 0, logged success" defect
 - the party id is captured on the row at write time
     FAIL: expected party id on the row, was null
 - erasing one party never reaches another party's rows
     FAIL: expected 0 but was 1
 - a row written with no party id is still erasable where sub is the party id  PASS
```

GREEN, whole module:

```
TOTAL tests=132 failures=0 errors=0 skipped=0
```

Local gate ran all three tasks (Gradle stops at the first failure, so this is the task list that
actually executed): `:openbank-copilot-service:test` + `ktlintCheck` (main and test source sets) +
`detekt`, all green.

## Kafka config: no replay is possible from this PR

This change touches **no** Kafka configuration. `group.id` / `auto.offset.reset` for
`party-events-in` were set by #3885 in `copilot-service-msg-override.yaml` via `override.properties`
+ `config_ordinal=500` — the shape that avoids the dotted-YAML-key trap — and are untouched here.
No consumer group is created, renamed, or re-pointed, and `auto.offset.reset` is not forced onto any
existing group, so **this PR cannot cause a topic replay.**

## Scope and what is still open

- No AWS or cluster mutation; the measurements quoted are from the issue thread, read-only.
- `openbank-engagement-service` and `openbank-admin-ui` untouched.
- Still open and deliberately not addressed here: none of the 9 live `party_id` attribute values
  matches a row in party-service's `parties` table either (noted on the issue as unexplained). This
  change is unaffected by that — it stores whatever identity the token carries and erases on either
  key — but it means the upstream identity question is still worth resolving on its own.
- No API change, so no `openapi.yaml` bump. `version.txt` is left to release-please.

Closes #3881

### DIFF
--- a/openbank-copilot-service/CLAUDE.md
+++ b/openbank-copilot-service/CLAUDE.md
@@ -66,9 +66,25 @@ Free-text chat is where the least predictable personal data lands, so the retent
   and no `PARTY_ERASED` consumer "yet". It is deliberately NOT corrected: Flyway checksums the whole
   file including comments, so editing an applied migration fails the service at boot. Believe this
   file, not that one.
-- **Known gap:** the event carries `partyId`, but history is keyed on the OIDC `sub`. The customers
-  realm also defines a separate `party_id` claim, so where those differ the erasure matches nothing.
-  Tracked separately — do not read the consumer as complete coverage.
+- **The erasure identity is captured at WRITE time, and must stay that way (#3881).** `PARTY_ERASED`
+  carries `partyId`; history is keyed on the OIDC `sub`, and those are not the same value — measured
+  against the deployed customers realm, `sub` equalled `party_id` for **0 of 35** users, because
+  `WebAuthnKeycloakClient.ensureUser` never sets the Keycloak user `id`, so Keycloak mints a random
+  UUID and `party_id` survives as a separate attribute. Keyed on `sub`, the consumer received the
+  event, deleted nothing, and logged `erased 0 copilot conversation(s)` at INFO — a GDPR Art. 17
+  control reporting coverage it did not have, provable wrong only by querying the table. It cannot be
+  fixed in the consumer: by erasure time the Keycloak user is gone, so nothing maps `partyId` back to
+  a `sub`. So `CopilotChatResource.erasureIdentity()` (the `party_id` claim, `sub` as fallback — the
+  resolution customer-edge already uses) writes `conversation_history.party_id` (migration V2), and
+  `deleteForParty` matches **either** column. Do not "simplify" that `or` away.
+- **`party_id` is never a lookup key.** `customerId` alone still decides which conversation a
+  customer resumes; changing that would silently move every customer to a different history and
+  orphan every existing row. Rows written before V2 carry no party id and are reachable only by the
+  `customer_id` arm; they pick one up on that customer's next message.
+- **An erasure test whose two identities are the same string cannot see this class of bug.** Every
+  test in `ConversationErasureIT` erases using the id it seeded with, so all four passed against the
+  broken code. `PartyErasureIdentityIT` deliberately seeds `sub != partyId` and asserts by direct
+  JDBC — measured red on the unfixed behaviour with `expected: 2L but was: 0L`.
 
 ## Build
 
--- a/openbank-copilot-service/src/main/kotlin/com/openbank/copilot/application/CopilotChatService.kt
+++ b/openbank-copilot-service/src/main/kotlin/com/openbank/copilot/application/CopilotChatService.kt
@@ -50,7 +50,7 @@ class CopilotChatService(
 ) : CopilotChatUseCase {
     private val log = Logger.getLogger(CopilotChatService::class.java)
 
-    override suspend fun handle(turn: ChatTurn, customerId: String): ChatOutcome {
+    override suspend fun handle(turn: ChatTurn, customerId: String, partyId: String?): ChatOutcome {
         if (!enabled) return ChatOutcome.Disabled
 
         guard.scanUserInput(customerId, turn.message)?.let {
@@ -67,7 +67,7 @@ class CopilotChatService(
 
         val outcome = converse(turn.conversationId, messages, customerId)
         if (outcome is ChatOutcome.Replied && outcome.reply.reply.isNotBlank()) {
-            persistTurn(customerId, turn.conversationId, turn.message, outcome.reply.reply)
+            persistTurn(customerId, turn.conversationId, turn.message, outcome.reply.reply, partyId)
         }
         return outcome
     }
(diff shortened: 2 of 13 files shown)


JiRaska:
**Both red checks here are inherited from `main`, not caused by this PR.**

- **`Customer-app dossier`** — `app-status.json` is derived from the *separate, private* `openbank-app` repo and the job regenerates it against that repo's default branch with no pinned ref. It goes stale when openbank-app **ships**, never because of anything a PR here does. #4053 (OPEN, CLEAN) scopes the PR-time comparison to drift a PR could actually have authored; the artifact itself still needs a regenerate-and-publish, which is the chore last done in `6a7d8f4da`.
- **`Trivy`** — `nanoid` 3.3.16 / CVE-2026-67213 HIGH is still in `main`'s tree. #4153 (OPEN, CLEAN) pins it to 3.3.17; its own `Trivy` check is SUCCESS, and the four other PRs red on this fail the *identical* step, `Trivy fs scan`.

Neither diff here touches `package-lock.json` or `app-status.json`.

This is the same class three times in one night: a PR going red for a fault it did not cause, which is precisely what #4019 exists to make visible.

PR #4158, opened 2026-08-08
fix(party): write party lifecycle events to the outbox, retire the direct emitter

JiRaska:
## What

Wires `openbank-party-service` onto its transactional outbox and removes the direct, non-transactional publisher. One service out of the eight in #4007 — done properly rather than eight done halfway.

`Refs #4007` (seven baselined services remain).

## Why party first

The issue names it as the clearest case, and it is the one where the guarantee is worth the most: `party_outbox` had the entire apparatus — Flyway migration, `PartyOutboxDispatcher`, backlog gauge, `openbank.outbox.dispatch-enabled: true`, even the atomic `claimProcessable` with `FOR UPDATE SKIP LOCKED` — and **nothing ever wrote a row to it**. Meanwhile the audit trail holds 61 `PARTY_CREATED`, 2 `PARTY_UPDATED` and 1 `PARTY_MERGED`, so the events demonstrably flow — through `KafkaPartyEventPublisher`, a bare `@Channel("party-events-out")` emit fired *after* the repository transaction had already committed.

That is a dual write. An emit that fails after the row commits loses the event with no record; one that succeeds before a rollback leaves a consumer acting on something that did not happen. `PARTY_MERGED` (ADR-0179) is the expensive divergence — a consumer that misses it keeps writing to a retired identity. The events are consumed by account-, aml-, kyc-, onboarding- and copilot-service, several money-path, so **wire it** rather than delete the apparatus. The decision is recorded in the new `openbank-party-service/CLAUDE.md`, as the issue asks.

## What happens to the direct emitter

**Removed**, along with the `PartyEventPublisher` port and the `party-events-out` channel. Consumers are unaffected, and this is the load-bearing fact rather than an assumption:

- both channels already targeted the **same topic**, `openbank.party.events` (`application.yaml:140` vs `:145` on main);
- the outbox payload is the event's own envelope verbatim — same field names, same order, same flat (non-nested) shape, still pinned by `PartyEventEnvelopeContractTest` and by the provider pacts, whose `@PactVerifyProvider` methods are untouched;
- the ACL surface is unchanged (`Write + Describe` on `openbank.party.events`, `kafka-party-mtls.yaml`);
- the only observable differences are additive: `OutboxKafkaHeaders` and a partition key derived from the aggregate id (the direct emitter sent a null key, so this *improves* per-party ordering).

Leaving both would have left two publishers racing on one topic, and only one of them can be atomic — so this is a removal, not a coexistence.

## How it is wired

- `PartyEvents` (domain, framework-free) builds the wire envelope.
- `PartyRepository.save` / `update` / `anonymize` gained event-carrying overloads. The impl chains `outboxRepository.persistInTransaction(...)` inside the **same** `Panache.withTransaction` block as the state change, the `ConsentRepositoryImpl` pattern.
- `PartyService` no longer publishes; the event travels with the write.

Both traps checked, not assumed:

- **`persist` vs `merge`** — `PartyEntity` extends `PanacheEntity`, so the `@Id` is generated and `save` is a genuine INSERT. `party_id` is the business key, not the id, so the application-assigned-id trap does not apply here. Noted in `CLAUDE.md` because it *would* apply the day someone changes that base class.
- **`openbank.outbox.dispatch-enabled`** — already `true` in `application.yaml`, verified rather than presumed; it defaults to `false` and events would never dispatch, silently.

## Atomicity evidence — red, then green

A mocked repository cannot tell which publisher a use case called, which is exactly why a fully green suite missed this for the life of the service. `PartyOutboxWriteIT` drives the real REST endpoints (a `Panache.withTransaction` repo called from a bare `@QuarkusTest` thread throws "No current Vertx context found" — only an HTTP request carries a context) and reads the row back over plain JDBC, with the dispatcher switched off so it cannot drain the row before the assertion sees it.

**RED** — the same IT file copied verbatim onto an unmodified `origin/main` worktree, Testcontainers Postgres + Redpanda:

```
TEST-com.openbank.party.integration.PartyOutboxWriteIT.xml tests=2 failures=2 errors=0
 - creating a party writes PARTY_CREATED ...
   AssertionError: [party_outbox rows for party 2f21c69d-...] Expected size: 1 but was: 0 in: []
 - a KYC status change writes KYC_STATUS_CHANGED ...
   Expecting actual: [] to contain exactly: ["PARTY_CREATED", "KYC_STATUS_CHANGED"]
```

The REST calls returned 201 and 200 in that run — the state change succeeded and the outbox stayed empty, which is the defect stated exactly.

**GREEN** — this branch, verdict read from the JUnit XML rather than the console:

```
TOTAL tests=147 failures=0 errors=0 skipped=1
PartyOutboxWriteIT: 2 fail 0 err 0 skip 0
```

(the one skip is the broker-gated `@PactBroker` provider test, `@EnabledIfSystemProperty(pactbroker.url)`, as on main.)

## Gate: before / after

```
before:  32 service(s) ship a dispatcher; 8 without a writer (8 baselined, 0 new, 0 stale baseline entries).
after:   32 service(s) ship a dispatcher; 7 without a writer (7 baselined, 0 new, 0 stale baseline entries).
```

`0 stale` on both sides is the load-bearing number — the script reports a baseline entry that has *become* covered, so removing `openbank-party-service` from `BASELINE` was mandatory, not optional. `--self-test` still passes.

## Local gate

`:openbank-party-service:` `test` (147/0), `ktlintCheck`, `detekt`, `koverVerify`, `quarkusBuild` — all executed and green; the task list was read rather than the summary line, since Gradle stops at the first failure.

## Not in scope

The other seven baselined services (`audit`, `balance`, `billing`, `kyc`, `pid`, `psd2`, `tpp-registry`) each need their own wire-or-delete decision and their own IT. `billing`'s two `DEAD` rows from a writer that is not on today's `main` are still unexplained and are worth resolving before anyone calls that table inert.

No API change, so `openapi.yaml` is untouched; `version.txt` is left to release-please.


JiRaska:
**The three non-inherited red gates here are my fault, and #4161 fixes them.**

`gates (data)`, `gates (kotlin)` and `Validate manifests` fail on a **duplicate `authz:` key** in `openbank-engagement-service`'s `%test` profile — a file this PR does not touch. `main` carries it, so this PR inherits it.

I introduced it while resolving a conflict on #4114: I took `main`'s side of what looked like a prose-only clash without noticing the branch already had an `authz:` block a few lines above, and validated with `yaml.safe_load` printing *"YAML parses OK"*. It does parse — a duplicate key is not a parse error, and the parser silently keeps the last one, which is the entire reason that gate exists.

**#4161** removes it. Once it lands, these three clear here with no change needed on this branch.

The remaining two, `Customer-app dossier` and `Trivy`, are the other two inherited faults reddening every open PR — #4053 and #4153 respectively.

JiRaska:
Reviewed. **Substantively correct.** The red required check splits in two — one half is yours, one is inherited.

### The atomicity claim is backed, not asserted

`save`/`update`/`anonymize(…, event)` chain `outboxRepository.persistInTransaction(...)` inside the *same* `Panache.withTransaction` block, and `PartyOutboxWriteIT` drives the real REST endpoints (RestAssured + `@TestSecurity`) with a `DispatcherOffResource` forcing `dispatch-enabled=false`, then reads the row over plain JDBC. That is the `LendingOutboxWriteIT` pattern rather than a mocked repository — which matters, because a mocked repo structurally cannot prove co-commit.

Honest gap worth a sentence, not a block: it proves the **happy path** commits together. There is no failure-injection case proving the *"or neither"* half — outbox insert fails ⇒ party row absent.

**The persist-vs-merge trap does not apply**, verified rather than taken on trust: `PartyEntity : PanacheEntity()` so the `@Id` is generated; `party_id` is a business column with a unique constraint, not the id. `save` is a genuine INSERT and every update path goes through `find(...).firstResult()` + dirty checking, never `persist`.

**`dispatch-enabled` is `true`** (`application.yaml:205`) — the footgun where an outbox fills and never dispatches, with no error and `attempt_count` stuck at 0, does not apply here.

**Retiring the direct emitter checks out**: all seven `eventPublisher.*` call sites converted 1:1, and no `partyRepo.save/update/anonymize` without an event remains in `src/main`. The event-free overloads surviving on the port (documented test/pact-state only) is a mild future footgun — a new lifecycle path could silently emit nothing — but that is a comment, not a defect.

### The red check

**`gates (kotlin)` is caused by this PR.** The ADR-0106 identifier-intent guard flags one added line:

```kotlin
e.email = "erased-${UUID.randomUUID()}@erased.invalid"
```

The line is pre-existing but was **moved** into the new `applyAnonymize` helper, and the guard is diff-scoped — so relocating it makes it new as far as the gate is concerned. One-line fix, but pick carefully: the right replacement is **`Ids.randomId()`** (a random v4), **not** `Ids.newId()`. `Ids.kt` says why in its own header — a UUIDv7 embeds its creation timestamp, so a v7 tombstone would be time-correlatable and would undercut exactly the non-derivability a GDPR erasure marker is for.

**`gates (data)` is inherited from `main`** — the duplicate `authz:` key in `openbank-engagement-service/application.yaml`, a file this PR does not touch. Its merge ref carries two keys; **current `origin/main` has one** (fixed by #4184 after #4106 re-added it). Merging `main` in clears it.

`Validate manifests` is only the aggregator over those two shards, so fixing the one-liner and refreshing greens it. `Customer-app dossier` and `Trivy` are red but neither is a required context.

### Pre-existing drift, not yours, worth its own issue

`docs/asyncapi/openbank-events.yaml:105` declares the party channel as `openbank.parties.party.event` — a topic **no producer or consumer in the tree uses**. The only other occurrence is `openbank-admin-ui/src/app/docs/api/page.tsx:863` repeating the same wrong name. `check-event-contract-coverage.py` passes (39 producer topics, 3 with a contract, 36 grandfathered, exit 0) because it is topic-granular and `openbank.party.events` is unchanged by this PR — so nothing catches a contract that documents a channel nobody speaks.

PR #4161, opened 2026-08-08
fix(engagement): remove the duplicate authz key that reddens main

JiRaska:
## What

Removes a duplicate `authz:` key from `openbank-engagement-service`'s `%test` profile. **`main` is currently red on `duplicate-yaml-key-guard` because of it.**

## I introduced this

Resolving the `application.yaml` conflict on #4114, I took `main`'s side of what looked like a prose-only conflict — two different comments explaining the same setting — without noticing the branch already carried an `authz:` block a few lines above. The merge kept both.

My validation was `yaml.safe_load(...)` printing *"YAML parses OK"*. It does parse. A duplicate key is not a parse error, and PyYAML silently keeps the last one — which is the entire reason this gate exists. **A probe that cannot express the failure reports success.**

The removed block's own comment made the point sharper by asserting, directly above the duplicate:

> `authz.enforce` has no key elsewhere in this file

It did.

## Why it matters even though nothing broke

SmallRye/SnakeYAML keep only the **last** of a repeated mapping key and drop the rest with no error. Here both blocks said `enforce: false`, so the effective value is unchanged and no test could see it — only the gate could. The next edit to either block is where that stops being true: changing the first one would have had no effect whatsoever, silently.

## What was kept

The earlier block, because its comment explains the deployed posture and cites the sidecar wired in #4068. Folded into it the one thing the removed comment added — the campaign-service precedent — so nothing was lost with the deletion.

## Falsification

Through the **real runner**, not a hand-rolled probe:

| state | `run-gates.py --only duplicate-yaml-key-guard` |
|---|---|
| duplicate restored | **exit 1**, names `application.yaml:82:3 duplication of key "authz"` |
| one key | **exit 0**, PASS |

Worth recording: my first attempt called the checker script directly and returned `rc=2` with **zero** matches for *both* states. That is a broken probe, not a passing one — it was discarded rather than read as evidence.

Engagement tests after the edit: **24 tests, 0 failures, 0 errors**, from the JUnit XML.

### DIFF
--- a/openbank-engagement-service/src/main/resources/application.yaml
+++ b/openbank-engagement-service/src/main/resources/application.yaml
@@ -66,7 +66,10 @@ mp:
   # defaultValue = "true" in AuthorizeInterceptor — so under @QuarkusTest, with no OPA PDP
   # bean in the graph, every one of them answers 503 and SurfaceRestContractIT's
   # `statusCode(200)` fails. Advisory in tests only; the deployed posture is unchanged and
-  # is decided by the sidecar wired in #4068.
+  # is decided by the sidecar wired in #4068. Same pattern and same reasoning as
+  # campaign-service's application.yaml: the contract IT is about the HTTP contract, not about
+  # who may call what — that is the OPA gate plus the rego tests, neither of which needs the
+  # app running.
   authz:
     enforce: false
   quarkus:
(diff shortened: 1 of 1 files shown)


JiRaska:
This fix already landed — and you were here first, so the duplication is mine to own.

`main` now has a single `authz:` key under `"%test":` (verified on `d0af48b68`), applied by #4160, which merged at 04:25Z. This PR was created at 04:16Z, i.e. **before** it. So the ordering is: you diagnosed and opened first, I hit the branch-name collision as a signal, and worked around it by renaming instead of searching for the PR that already existed. That is the same "is anyone else touching this artifact" check I had been running on other people's PRs all evening, skipped twice on my own — once when I created the duplicate key in the first place (#4128 vs #4072), and again here.

The branch now shows `CONFLICTING`, which is expected: both diffs edit the same block.

Not closing it, because the worktree behind it is live and you may have more in flight. If the duplicate-key removal was all of it, this is a no-op now and can be closed. If you are carrying something else on the branch, merging `main` in will drop the redundant half cleanly.

One detail from #4160 worth keeping, in case it shapes what you do with the surviving comment: I dropped the line *"`authz.enforce` has no key elsewhere in this file"* from the merged block. It was true when written and false the moment the second block landed — the sort of claim that reads as a verified fact long after it stopped being one.

JiRaska:
Superseded by #4160 — identical fix (dedup the %test authz key duplicate; both blocks were enforce: false, no policy change). Closing as duplicate.

PR #4163, opened 2026-08-08
fix(libs): an open circuit breaker must not drive outbox rows to terminal DEAD

JiRaska:
## What this is

`Refs #4005.` The issue names **three** things that had to be separated before anything could be
fixed, because only one of them is a live code defect:

| Claim | Verdict |
|---|---|
| 24 `card_outbox` rows stuck `DEAD`, service has published zero events | **True and still true today.** Re-measured on the live DB (below). |
| An open circuit breaker with no self-healing path | **The breaker is long gone; the parking is permanent.** Breaker state is in-memory, so the fault cleared on the next pod restart. `DEAD` is terminal and excluded from `listProcessable`, so the rows never came back — and the breaker's half-open probe had no work left to probe with. |
| `created_at` stamped 1970 | **The already-fixed `Instant.EPOCH` family (#3272/#3338), not a live defect.** `OutboxMessage.createdAt` and `markSent`/`markFailed`'s timestamps defaulted to `Instant.EPOCH`; `0306c7c81` fixed that in libs on 2026-08-02. All 24 rows were last written on or before 2026-08-01 (`claimed_at`), i.e. they predate the fix. New rows stamp correctly. |

The issue also flags `CardEvents.occurredAt = Instant.EPOCH` as the `#3874`/`#3883` class and predicts
those events will publish with a 1970 business date. **That prediction is wrong, and the check is one
query.** All four call sites in `CardService` pass a real clock reading, and every one of the 24
parked payloads carries a real `occurredAt`:

```
card.controls_changed.v1 | 6 | null occurredAt: 0 | 2026-07-23T20:35:00Z .. 2026-07-30T18:57:15Z
card.issued.v1           | 6 | null occurredAt: 0 | 2026-07-23T18:59:30Z .. 2026-08-01T08:36:16Z
card.limits_changed.v1   | 9 | null occurredAt: 0 | 2026-07-24T10:34:32Z .. 2026-07-25T05:14:25Z
card.status_changed.v1   | 3 | null occurredAt: 0 | 2026-07-25T08:56:50Z .. 2026-07-31T10:17:49Z
```

So it is a **trap that had not sprung**, not corruption — the distinction this repo already insists on
before calling a wrong-looking value a live defect. The fix here is therefore to make omitting it a
compile error, not to chase data.

## The actual defect: a breaker fast-fail was counted as a delivery attempt

Every one of the 24 rows is at `attempt_count = 10` (`OutboxFailurePolicy.DEFAULT_MAX_ATTEMPTS`) with
the *same* `last_error`:

```
com.openbank.cardissuance.infrastructure.outbox.CardOutboxDispatcher#publishWithResilience
  circuit breaker is open
```

`OutboxDispatch.dispatchOnce` caught every `Exception` and called `markFailed`, which increments
`attempt_count` and applies `OutboxFailurePolicy`. But `CircuitBreakerOpenException` is thrown by the
interceptor **instead of** invoking the publisher — nothing was ever offered to Kafka. With a 5 s tick
and a 25-row batch, the whole backlog was re-claimed and re-"failed" every tick, so roughly **50
seconds of an open breaker was enough to make all 24 rows terminal**. `DEAD` is for a poison row that
can never be published (ADR-0050 N5); a broker outage is not that, and it had the entire fleet's
outbox one brief Kafka blip away from permanent loss. `billing_outbox` carries 2 rows in exactly the
same state, same error.

### The change

`OutboxDispatch.dispatchOnce` now classifies `CircuitBreakerOpenException` and `BulkheadException` —
both thrown *before* invocation — as a **system** signal rather than a row signal, and abandons the
batch untouched: no `markFailed`, no attempt consumed. The rows a claiming repository already flipped
to `DISPATCHING` come back via its stale-claim reclaim (the same path that recovers a pod which died
mid-batch); a repository on the unclaimed-peek default sees them again on the next tick. Either way
the work stays processable, which is precisely what gives the breaker's half-open probe something to
probe with. **That is the self-healing path the issue says is missing.**

Deliberately **not** included: `TimeoutException`. A timeout can fire after the record already reached
the broker, so it must keep counting as a real attempt.

Matching is by fully-qualified class name, not `is CircuitBreakerOpenException`: `openbank-libs-domain`
has zero framework dependencies (ADR-0002/ADR-0122, `check-domain-purity.py`), and a hard reference
would also risk `NoClassDefFoundError` in a service with no fault-tolerance extension.

A poison row still reaches `DEAD` after 10 *real* failed publishes. That behaviour is unchanged and is
asserted.

## Can this republish events consumers have already seen?

**No — not from this change.** The two exempted exceptions are thrown *instead of* invoking the
publish, so a row skipped by this branch provably never reached the broker; declining to burn its
attempt cannot produce a duplicate. `TimeoutException`, the one failure mode that *can* leave a record
at the broker, is excluded from the exemption for exactly that reason.

What the change does do is widen the window in which a row is retried: under a long outage rows now
retry indefinitely instead of dying after ~50 s. The pre-existing at-least-once property of the outbox
(a crash between a successful publish and `markSent` replays the row) is unchanged in kind.

**The remediation of the 24 parked rows is a different answer — see below.**

## What is NOT in this PR

**The 24 rows are not fixed by any code change, and I have not touched the database.** They are
terminal; nothing in the system will ever move them. But they are **not lost**: every payload is
intact (24/24 non-empty, 24 distinct `event_id`), so this is a requeue, not a reconstruction.

Proposed one-off remediation, **for the owner to run, not an agent**:

```sql
-- card_outbox: requeue the 24 breaker-killed rows. claimed_at is the honest timestamp —
-- created_at/updated_at are the pre-0306c7c81 epoch default, and left at 1970 these rows
-- sort ahead of every future entry in the ORDER BY created_at ASC claim.
UPDATE card_outbox
   SET status = 'PENDING', attempt_count = 0, last_error = NULL,
       created_at = claimed_at, updated_at = now()
 WHERE status = 'DEAD';
```

Three things the owner should weigh first, which the issue does not raise:

1. **3 of the 24 rows point at an aggregate that no longer exists.** `LEFT JOIN cards` on
   `aggregate_id`: 2 `card.issued.v1` and 1 `card.controls_changed.v1` reference a card that is not in
   the `cards` table (24 rows span 6 aggregates; `cards` holds 4). Replaying a `card.issued.v1` for a
   deleted card creates a downstream projection for a card that does not exist.
2. **These are 6-to-15-day-old lifecycle events published as current.** `limits_changed` /
   `controls_changed` / `status_changed` are last-write-wins state transitions — replaying in
   `claimed_at` order is correct, replaying an old value over a newer one set by another path is not.
3. **Do one row first.** Requeue a single `card.issued.v1`, confirm it reaches `SENT` and that a
   consumer handled it, before releasing the other 23.

Also still open and not addressed here: **`DEAD > 0` on a money-path outbox is silent** — the table is
the only place it shows. That is the general form of #3345 and is what stops this recurring; it needs
its own PR. Hence `Refs`, not `Closes`.

## Tests — red then green

Counts read from the JUnit XML, not the console.

`openbank-libs-domain` `OutboxDispatchTest`, with **only** the new branch disabled in
`dispatchOnce` (same test file, same public API, so this is a behaviour failure and not a compile
failure masquerading as one):

```
RED  : tests=6 failures=2 errors=0 skipped=0
GREEN: tests=6 failures=0 errors=0 skipped=0
```

The two that flip are the ones that describe the defect:

```
OutboxDispatchTest > an open breaker consumes no attempt and abandons the rest of the batch() FAILED
OutboxDispatchTest > a breaker-open cause nested under another exception is still not an attempt() FAILED
```

The other four pass in both variants **by design** — they are the over-reach guards, and a change that
made them flip would be a change that had gone too far: a genuine publish failure must still count (so
a poison row still reaches `DEAD`), and a `TimeoutException` must still not be exempt.

`openbank-card-issuance-service` `CardOutboxBreakerOpenIT` is the half a mocked repository cannot
prove, since the claim under test is the row's *committed* `status` and `attempt_count`. It runs the
**real** `CardOutboxRepositoryImpl` against the Testcontainers Postgres and asserts over a plain JDBC
read on a separate connection, so nothing can be answered out of a first-level cache. It drives ten
breaker-open ticks — the exact count that took the live rows from `PENDING` to `DEAD` — and asserts the
row is not `DEAD`, `attempt_count` is 0 and `last_error` is null; then that the row is reclaimable and
settles `SENT` with exactly one attempt once the breaker closes.

Measured the same way, against a real Postgres:

```
RED  : tests=2 failures=2 errors=0 skipped=0
GREEN: tests=2 failures=0 errors=0 skipped=0
```

and the RED is the live defect reproduced verbatim, not a proxy for it:

```
a breaker-open fast-fail burns no attempt and never reaches DEAD()
  java.lang.AssertionError: Expecting actual: "DEAD" not to be equal to: "DEAD"
the row stays reclaimable, so the breaker's half-open probe has work to probe with()
  org.opentest4j.AssertionFailedError: expected: 1 but was: 2
```

## Local gate

`./gradlew :openbank-libs-domain:test :openbank-libs-domain:ktlintCheck :openbank-libs-domain:detekt
:openbank-card-issuance-service:test :openbank-card-issuance-service:ktlintCheck
:openbank-card-issuance-service:detekt :openbank-card-issuance-service:quarkusBuild` — `BUILD
SUCCESSFUL`, with all eight tasks present in the output (Gradle stops at the first failure, so a
green task list is only worth as much as the tasks that appear in it). `check-domain-purity.py`
reports 0 new violations.

The `Customer-app dossier` check is red on this PR and is **not** caused by it: the committed
`app-status.json` says version 0.18.0 while the openbank-app source is at 0.19.0. It is red on every
other open PR too (#4161, #4158, #4156, #4155; last green on #4153) and needs its own regeneration PR.

## Other changes

`CardEvents.occurredAt` loses its `Instant.EPOCH` default on all four subtypes (see the table above —
no call site relied on it, so this is compile-safe and turns a silent 1970 into a compile error).

### DIFF
--- a/openbank-card-issuance-service/src/main/kotlin/com/openbank/cardissuance/domain/event/CardEvents.kt
+++ b/openbank-card-issuance-service/src/main/kotlin/com/openbank/cardissuance/domain/event/CardEvents.kt
@@ -10,6 +10,19 @@ import java.util.UUID
 
 sealed class CardEvent {
     abstract val cardId: UUID
+
+    /**
+     * Business time of the change, serialized into the outbox payload and read downstream as the
+     * event's business date.
+     *
+     * **Deliberately has no default.** It used to default to `Instant.EPOCH` on all four subtypes —
+     * the same shape as `AuditEvent.timestamp` / `FlagExposure.timestamp` (#3874/#3883) and
+     * `OutboxMessage.createdAt` (#3272/#3338), where the default WAS what every row got. Here it
+     * never fired: all four call sites in `CardService` pass a real clock reading, and the 24 rows
+     * parked in `card_outbox` all carry a real `occurredAt` in their payload. So this is a trap
+     * that had not sprung, and the way to keep it that way is to make omitting it a compile error
+     * rather than a silent 1970 (#4005).
+     */
     abstract val occurredAt: Instant
 }
 data class CardIssued(
@@ -19,7 +32,7 @@ data class CardIssued(
     val cardType: CardType,
     val network: CardNetwork,
     val maskedPan: String,
-    override val occurredAt: Instant = Instant.EPOCH,
+    override val occurredAt: Instant,
 ) : CardEvent()
 data class CardStatusChanged(
     override val cardId: UUID,
@@ -27,14 +40,14 @@ data class CardStatusChanged(
     val newStatus: CardStatus,
     val reason: String?,
     val changedBy: String,
-    override val occurredAt: Instant = Instant.EPOCH,
+    override val occurredAt: Instant,
 ) : CardEvent()
 data class CardLimitsChanged(
     override val cardId: UUID,
     val dailyLimitMinorUnits: Long,
     val monthlyLimitMinorUnits: Long,
     val changedBy: String,
-    override val occurredAt: Instant = Instant.EPOCH,
+    override val occurredAt: Instant,
 ) : CardEvent()
 data class CardControlsChanged(
     override val cardId: UUID,
@@ -43,5 +56,5 @@ data class CardControlsChanged(
     val atmEnabled: Boolean,
     val abroadEnabled: Boolean,
     val changedBy: String,
-    override val occurredAt: Instant = Instant.EPOCH,
+    override val occurredAt: Instant,
 ) : CardEvent()
--- a/openbank-card-issuance-service/src/test/kotlin/com/openbank/cardissuance/integration/CardOutboxBreakerOpenIT.kt
+++ b/openbank-card-issuance-service/src/test/kotlin/com/openbank/cardissuance/integration/CardOutboxBreakerOpenIT.kt
@@ -1,0 +1,178 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) OpenBank contributors. Licensed under the Apache License, Version 2.0.
+// See LICENSE in the repository root or https://www.apache.org/licenses/LICENSE-2.0 for details.
+package com.openbank.cardissuance.integration
+
+import com.openbank.cardissuance.infrastructure.persistence.entity.CardOutboxEntity
+import com.openbank.cardissuance.infrastructure.persistence.repository.CardOutboxRepositoryImpl
+import com.openbank.libs.persistence.outbox.OutboxDispatch
+import com.openbank.libs.persistence.outbox.OutboxFailurePolicy
+import com.openbank.libs.persistence.outbox.OutboxRepository
+import com.openbank.libs.persistence.outbox.OutboxStatus
+import io.quarkus.hibernate.reactive.panache.Panache
+import io.quarkus.test.common.QuarkusTestResource
+import io.quarkus.test.junit.QuarkusTest
+import io.quarkus.vertx.VertxContextSupport
+import io.smallrye.mutiny.coroutines.asUni
+import jakarta.inject.Inject
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.async
+import org.assertj.core.api.Assertions.assertThat
+import org.eclipse.microprofile.config.ConfigProvider
+import org.eclipse.microprofile.faulttolerance.exceptions.CircuitBreakerOpenException
+import org.junit.jupiter.api.Test
+import java.sql.DriverManager
+import java.time.Duration
+import java.time.Instant
+import java.util.UUID
+
+/**
+ * #4005 — **an open circuit breaker must not drive outbox rows to terminal `DEAD`.**
+ *
+ * What the live table looked like: all 24 `card_outbox` rows `DEAD`, every one at
+ * `attempt_count = 10` (`OutboxFailurePolicy.DEFAULT_MAX_ATTEMPTS`), every one carrying the *same*
+ * `last_error` — `CardOutboxDispatcher#publishWithResilience circuit breaker is open`. So none of
+ * those 240 recorded "attempts" was a delivery attempt: the breaker short-circuited before the
+ * publisher was ever called, each 5 s tick re-claimed the whole batch, and ~50 s of breaker-open was
+ * enough to make every row terminal. `DEAD` is excluded from `listProcessable`, so nothing retried
+ * them again and the breaker's own half-open probe had no work left to probe with.
+ *
+ * ### Why this is an IT and not a unit test
+ * The bookkeeping under test is `markFailed`'s attempt increment plus [OutboxFailurePolicy], which
+ * lives in the repository and commits to Postgres. A mocked repository would assert only that a
+ * method was (not) called — it cannot show the row's committed status or `attempt_count`, which is
+ * the entire claim. So the repository here is the **real** [CardOutboxRepositoryImpl] against the
+ * Testcontainers Postgres, and the assertions are a plain JDBC read on a separate connection —
+ * outside the reactive session, so nothing can be answered out of a first-level cache.
+ *
+ * The publisher is the only stubbed part, because [CircuitBreakerOpenException] is precisely what a
+ * real `@CircuitBreaker` interceptor throws *instead of* invoking the publish. Driving it through
+ * the annotated `CardOutboxDispatcher.publishWithResilience` would need ten real failures plus the
+ * 5 s breaker delay to reach the same exception, i.e. a slower and timing-dependent way of
+ * producing the identical input.
+ */
+@QuarkusTest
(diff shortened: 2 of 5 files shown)


JiRaska:
**The blast radius is wider than card-issuance, and that is the argument for this fix.**

`dispatchOnce` lives in `openbank-libs-domain`, and **36 modules** reference `OutboxDispatch`. The defect is one clause there:

```kotlin
publish(entry)
repository.markSent(entry.eventId)
} catch (ex: Exception) {
    repository.markFailed(entry.eventId, ex.message ?: ex.javaClass.simpleName)
}
```

`CircuitBreakerOpenException` is thrown **instead of** invoking the publisher, so it never reaches Kafka — yet it lands in the same `catch (ex: Exception)` and burns an attempt. With a 5 s tick and a 25-row batch, ~50 seconds of an open breaker is enough to make a whole outbox terminal without one row ever being offered to the broker. `DEAD` is excluded from `listProcessable`, so nothing retries them, and the half-open probe has no work left to probe with. That pairing is the "no self-healing path".

`billing_outbox` already has 2 rows in the same state, which is the second instance rather than a coincidence.

**Verified independently**: the `catch` shape above is quoted from `origin/main`, and the 36-module count is derived from `git grep -l OutboxDispatch` over `src/main`, not from the issue.

**On the two red checks: neither is yours.** Both are inherited faults that redden every open PR right now:

- `Customer-app dossier` — `app-status.json` is derived from the separate private `openbank-app` repo and goes stale when *that* repo ships. #4053 scopes the PR-time comparison to drift a PR could have authored.
- `Trivy` — `nanoid` 3.3.16 / CVE-2026-67213 sits in `main`'s tree. #4153 pins it to 3.3.17 and its own Trivy is SUCCESS.

**Worth keeping**: the issue predicted a second defect — `CardEvents.occurredAt = Instant.EPOCH` producing a 1970 business date downstream — and that prediction is **false**. All four `CardService` call sites pass a real clock reading and all 24 payloads carry a real `occurredAt`, zero null. Making the omission a compile error rather than chasing the data is the right response to a trap that had not sprung.

JiRaska:
Reviewed. **Correct — I could not construct a message-loss scenario.** One observability consequence worth a follow-up.

**Nothing is dropped.** On `CircuitBreakerOpenException`/`BulkheadException` the whole batch returns without `markFailed`, so no attempt is burned, and rows already flipped to `DISPATCHING` come back through the stale-claim reclaim. I checked that the reclaim is universal rather than assuming it: **31 of 32** `*OutboxRepositoryImpl` override `claimProcessable` with `FOR UPDATE SKIP LOCKED` + `status = DISPATCHING AND claimed_at < :staleThreshold`; the one that does not (`SecurityOutboxRepositoryImpl`) never claims at all, so it re-peeks next tick. Everything is retried.

The dangerous shape for this kind of change — *stops retrying without recording* — is inverted here: the row keeps its full budget. A genuine publish failure still `markFailed`s and still reaches `DEAD` at 10, and excluding `TimeoutException` is right (it can fire after the broker has acked), with a unit test pinning that exclusion.

**Blast radius is contained.** No new bean, no `@Produces`, so the `libs-runtime` type-closure hazard does not apply. All 32 dispatchers carry `@CircuitBreaker`, so it lands fleet-wide as intended, and the test-only MicroProfile exception stub cannot collide — `openbank-libs-domain/build.gradle.kts` has no microprofile artifact on any configuration. The outer `@CircuitBreaker` on `dispatch()` now sees a success rather than a propagating throw, so there is no second-order trip.

**Test layer is right**: `CardOutboxBreakerOpenIT` uses the real repository on Testcontainers Postgres and asserts over a separate JDBC connection, so the *committed* `status`/`attempt_count` is what gets proven. Collapsing `staleAfter` to `ZERO` while keeping the repository's own claim SQL is the honest way to exercise the reclaim.

### The follow-up: this makes a broker outage less visible, not more

Before, an operator saw rows go `FAILED` with `last_error = "circuit breaker is open"`. Now they sit in `DISPATCHING` — and `countProcessable` counts only `PENDING` and `FAILED`:

```kotlin
count("status in (?1, ?2)", OutboxStatus.PENDING.name, OutboxStatus.FAILED.name)
```

So the `openbank.outbox.backlog` gauge **under-reports during exactly the incident it exists to signal**. With a 5 s tick, a 25-row batch and a 2-minute `staleAfter`, on the order of ~600 rows can be parked invisibly, and the only trace is a single WARNING line — there is no counter for abandoned batches.

Not a blocker: the alternative was permanent loss, and this is unambiguously the better trade. But the PR already notes that `DEAD > 0` alerting is missing, and this narrows the observable trail further, so the two belong in the same follow-up. Either count `DISPATCHING` separately in the gauge, or emit a counter when a batch is abandoned breaker-open.

PR #4167, opened 2026-08-08
chore(admin-ui): regenerate app-status.json from openbank-app source

JiRaska:
## What

Regenerates the derived `openbank-admin-ui/app-status.json` block from the `openbank-app` source. One field: `derived.version` `0.18.0` -> `0.19.0`.

## Why

The `Customer-app dossier` job's `Regenerate & diff derived block` step is failing on **every open PR**, and is not caused by any of them:

```
##[error]the committed app-status.json no longer matches openbank-app source (1 field(s)). Regenerate & publish.
  derived-block drift:
    version: committed="0.18.0" -> regenerated="0.19.0"
```

Confirmed red on #4163, #4161, #4158, #4156, #4155; last green on #4153. `openbank-app` moved to 0.19.0 and the committed copy was left behind.

## How

Not hand-edited — derived data (ADR-0074). Ran the same command the workflow runs, in write mode:

```
node scripts/generate-app-status.mjs --app-repo <openbank-app checkout>
```

against a fresh clone of `JiRaska/openbank-app` default branch @ `4293b9b5b` (`version.txt` = `0.19.0`).

## Verification

- Before the regen, the workflow's exact enforce command reproduced the failure locally (exit 1, identical drift line).
- After, `node scripts/generate-app-status.mjs --app-repo <checkout> --against app-status.json --enforce --check` prints `derived block matches the committed app-status.json` and exits 0.
- `asOf`, capabilities (18: live=13, partial=4, planned=1) and every other derived field are unchanged — the generator rewrote only the one drifted field.

Precedent: 6a7d8f4da `chore(admin-ui): regenerate app-status.json from openbank-app source (#3917)`.

`chore` type is release-hidden, so this cuts no admin-ui release.

### DIFF
--- a/openbank-admin-ui/app-status.json
+++ b/openbank-admin-ui/app-status.json
@@ -12,7 +12,7 @@
   "asOf": "2026-07-29",
   "appRepo": "JiRaska/openbank-app",
   "derived": {
-    "version": "0.18.0",
+    "version": "0.19.0",
     "useFakeData": false,
     "edgeBaseUrl": "https://customer.open-bank.tech",
     "keycloakIssuer": "https://kc.open-bank.tech/realms/openbank-customers",


JiRaska:
**#4165 and #4167 are the same change, produced four minutes apart.** Compared by CONTENT, not by title: the added lines of both diffs hash identically (`13fcd99ec1bf9a9b`), both touch exactly one file (`openbank-admin-ui/app-status.json`), and both write `"version": "0.19.0"`.

Merge one, close the other. #4165 is the earlier (04:57:40Z vs 05:01:56Z), so it is the natural survivor; #4167 is `UNSTABLE` where #4165 is not, which points the same way.

Both are red on `Trivy` only, and that is inherited — `nanoid` 3.3.16 / CVE-2026-67213 is in `main`'s tree. Neither diff touches `package-lock.json`. **#4153** pins it and its own `Trivy` is SUCCESS.

Worth naming the pattern rather than just the instance: this artifact is derived from the separate private `openbank-app` repo and goes stale whenever *that* repo ships — never because of anything a PR here does. Regenerating it is the chore that has now been done six times (#3212, #3552, #3818, #2101, #3053, and this pair), each buying exactly one release of quiet, and this time it was done twice concurrently. **#4053** stops it reddening unrelated PRs; **#3988** items 2 and 3 (an escalation path on the scheduled refresh, and a cron that matches the measured three-releases-in-12-hours cadence rather than the "every few weeks" its comment assumes) are what would stop the chore recurring at all.

JiRaska:
Closing as a duplicate of #4165, which merged at 13:44Z.

**Verified by content, not by title.** Both PRs regenerate the same single file, four minutes apart, and the two versions are **byte-identical**. More to the point, this branch's `app-status.json` is now byte-identical to `origin/main` — so there is nothing left for it to change:

```
diff <(git show origin/main:openbank-admin-ui/app-status.json) \
     <(git show origin/chore/admin-ui-app-status-0-19-0:openbank-admin-ui/app-status.json)
  -> no output
```

That check is the one worth doing rather than assuming. The two PRs have different branch names and different titles, so nothing about them signals duplication from the outside; only comparing the produced artifact does. Same shape as the branch-name collisions this queue has thrown up repeatedly — a regenerated file is either identical or it is a real divergence, and reading the content is the only way to tell which.

No work lost: the regeneration this PR performed is on `main`, it simply arrived via the other one.

PR #4172, opened 2026-08-08
fix(ci): stop the metadata sweep reporting its own crash as drift (#4162)

### LINKED ISSUE #4162: verification-metadata drift: unpinned artifacts on main
The daily verification-metadata sweep FAILED: at least one Gradle module on
`main` resolves an artifact that `gradle/verification-metadata.xml` does not pin.

Run: https://github.com/JiRaska/open-bank-oss/actions/runs/31239958766 — each failing shard's log names the exact artifacts.

**Latent today, breaking on 2026-09-30.** `org.gradle.dependency.verification`
is `lenient`, so this is currently a warning. At the ADR-0144 graduation (#1915)
it becomes a failing build on every cold cache — a fresh runner, a recreated ARC
cache volume, CodeQL's tracing build, a new contributor clone.

To fix, regenerate for the affected modules and commit **only the additions**
(a full rewrite reindents ~18.5k lines and has previously deleted valid entries):

```
./gradlew --write-verification-metadata sha256 --refresh-dependencies \
  :<module>:testClasses :<module>:quarkusDependenciesBuild \
  -Dorg.gradle.jvmargs=-Xmx6g
```

Cross-check each new sha256 against Maven Central's published `.sha1` before
committing it — this file is a security control, and "add the hash the build
showed you" is the failure mode it exists to prevent.

Context: #3131.

JiRaska:
## The issue's premise is not what happened

#4162 says *"at least one Gradle module on `main` resolves an artifact that `verification-metadata.xml` does not pin"*. **Nothing measured that.** All three failing shards of run 31239958766 died of `Java heap space`, twice each, during regeneration:

```
FAILURE: Build completed with 2 failures.
* What went wrong:
A build operation failed.
    Java heap space
```

The sweep never reached a verdict about the metadata at all. The issue body is a **fixed heredoc template** in the workflow, published by `if: failure()` — which fires on any shard failure, for any reason.

The sharpest part: the checker **already knew**, and said so in its own output —

> `this says nothing about whether the metadata is complete. Re-run, and if it persists treat it as a build problem, not a metadata gap.`

— and then `return 1`, the same code as a real gap. A checker that knows the difference but does not encode it in the one thing its caller reads has not communicated it.

## Three changes, in the order the failure happened

**1. The checker encodes the difference.** Exit **2** for "could not run", documented as a contract: `0` = no gap, `1` = a real gap, `2` = unknown.

**2. The workflow reads the verdict instead of assuming it.** Each shard records `gap` / `could-not-run` as an **artifact**; `classify-verification-sweep-failure.sh` reduces them and `raise-issue` branches on the result, with a separate title so a build problem and a real drift cannot refresh each other's issue.

Artifacts rather than a log grep on purpose: a job log contains the step's own `run:` script, so grepping it matches strings that never executed.

A single real gap **outranks** any number of crashes — the metadata is incomplete whatever else broke — so only an all-infrastructure run is reported as a build problem.

**3. Twelve shards, not eight.** The header claimed *"~7 modules"* per shard; shard 1 was regenerating **nine**, because the count is fixed while the fleet grows. Preferring more shards over more heap is deliberate: 6g of a 16 GB runner is already the safe per-JVM ceiling, and this repo already records that a heap limit is a ratchet with nothing measuring it — `dependency-submission` sat red for three days on the same fault while everything downstream read a stale graph.

## Falsification

Every state the classifier can see, reading what it **prints**, not just its exit code:

| verdict files | result |
|---|---|
| `gap` + `could-not-run` | **gap** |
| `could-not-run` only | **could-not-run** |
| `gap` only | **gap** |
| directory empty | **unknown** |
| directory absent | **unknown** |

`actionlint`'s finding **set** is identical to `origin/main`'s for this file (both empty), compared by linting each side rather than post-filtering one run. The run-step size gate passes.

## What I did not do

I did not regenerate any metadata. There is no evidence of a gap to fix, and the house rule is additions-only — a full rewrite reindents ~18.5k lines and has previously deleted valid entries. If the next sweep, with room to finish, does find real gaps, it will now say so under a title that means it.

Closes #4162


JiRaska:
Reviewed. The fix is right and the exit-2 split loses nothing. **One wiring bug that would invert it.**

**What it was doing, restated so the fix is judged against the real defect:** `check-verification-metadata-complete.py` returned `1` both for "a real gap" and for "Gradle died", while printing a line saying the crash *"says nothing about whether the metadata is complete"*. The workflow's `if: failure()` then published a fixed heredoc asserting unpinned artifacts on `main` — so #4162 claimed drift that nothing had measured.

**Nothing real is lost by the split.** Regeneration is one `./gradlew` per shard; on failure `after = before`, so `missing` is empty regardless. A crashed shard never produced a gap verdict under the old code either — it produced a *false* gap label. The only genuine narrowing is granularity: a shard where Gradle dies leaves ~5 modules with no verdict, and if a different shard reports a gap the classifier correctly outranks it (`gap` + `could-not-run` → `gap`, verified). The classifier matches all five states in the PR's own table.

### The finding — `verification-metadata-drift.yml:95`

```yaml
if: ${{ always() && hashFiles('/tmp/sweep-verdict/**') != '' }}
```

`hashFiles()` only matches files **inside `GITHUB_WORKSPACE`**. An absolute `/tmp` glob returns `''`, so the guard is false, the upload step never runs, `raise-issue` downloads nothing, the classifier returns `unknown`, and **every** failure — including a real gap at exit 1 — gets filed as *"verification-metadata sweep could not run on main"*.

That is the mirror image of the bug being fixed: no longer over-claiming drift, now silently under-claiming it. And it fails in the direction that reads as benign.

Checked for a counter-example rather than asserting it: every other `hashFiles()` in this repo is workspace-relative (`'*.cdx.json'`, `'**/*.gradle.kts'`), so there is no local precedent showing the absolute form working.

Trivial either way — write the verdict under `"$GITHUB_WORKSPACE/sweep-verdict/"`, or drop the guard entirely and use `if: always()` with `if-no-files-found: ignore`, since the sweep step exits non-zero in both verdict cases so `always()` alone suffices.

**Everything else checks out:** `--selftest` still passes under the new exit-code contract; the heredoc indentation inside the new `if` matches the existing flush-left block, so YAML common-indent stripping is unchanged; and verdicts are read from **files**, not from job logs — which sidesteps the trap where a grep matches the step's own `run:` script that GitHub prints into the log header.

JiRaska:
Read the diff against what actually happened on 2026-08-08 rather than the design alone.

The core fix is a real distinction the checker already made in prose but not in its exit code: `return 1` covered both "found a gap" and "could not determine anything" (Gradle OOM), and the caller's `if: failure()` couldn't tell them apart — so #4162 published a fixed-template issue asserting a metadata gap when all three failing shards had died of `Java heap space`, never reaching a verdict.

Splitting to exit 2 = "could not run" (not a verdict) vs exit 1 = "real gap" is the actual fix; everything else follows from having that signal available.

Two implementation choices worth confirming, and I did:
- **Reads verdicts from per-shard artifact files, not job logs.** The comment explaining why is exactly the lesson this repo has hit before — a job log contains the step's own `run:` script, so grepping it matches strings that never executed. Structured output over log-scraping is the right fix here, not a style preference.
- The checker's own docstring now states the exit-code contract explicitly (0/1/2 and what each means), so the next caller can't reintroduce the collapse by writing a new consumer that treats 2 like 1.

Not merging — workflow change, human review by rule.

JiRaska:
Re-reviewed at head. **The finding from the last round stands, unaddressed** — the only commit since (`36a34f7db`, 14:10:23Z) is a merge from `main` with no code change, and my comment was 13:40:06Z.

## Blocking, unchanged: `hashFiles()` cannot see `/tmp`

`verification-metadata-drift.yml:95`

```yaml
if: ${{ always() && hashFiles('/tmp/sweep-verdict/**') != '' }}
```

`hashFiles()` globs resolve relative to `GITHUB_WORKSPACE` and cannot reach outside it, so an absolute `/tmp` pattern always yields `''`. The upload never runs, `download-artifact` gets nothing, the classifier returns `unknown`, and **every** failure — including a real gap at exit 1 — is filed as *"verification-metadata sweep could not run on main"*.

That is the mirror image of the bug this PR exists to fix, and it fails in the direction that reads as benign: a genuine unpinned-artifact gap gets reported as infrastructure flakiness.

I re-checked the precedent rather than asserting it: every other `hashFiles()` in this repo's workflows is workspace-relative (`security.yml:468`, `release-please.yml:533`, the Gradle cache keys). Nothing here writes outside the workspace and then tries to glob it.

Fix either way: write verdicts under `$GITHUB_WORKSPACE/sweep-verdict/`, or drop the guard entirely for `if: always()` plus `if-no-files-found: ignore`.

## New: `*)` treats a crash as a real gap

`:90` — `*) echo "gap"` catches any exit that is not 0 or 2. Reading the checker, `return 1` fires when `gradle/verification-metadata.xml` is **not found** — which is not a gap. A Python traceback, an argparse error, or an OOM kill at rc=137 all land there too.

So a crash of the wrapper files a drift issue asserting unpinned artifacts — the exact class of false claim this PR was written to stop. `1) gap ;; *) could-not-run ;;` is the honest mapping.

## Notes

- **`services-ci.yml:423` does not honour the new contract.** The docstring now says callers must treat 2 as unknown and must not report it as drift; the PR lane still fails on any non-zero. Not a regression — identical to before — but the contract is stated fleet-wide and satisfied in one place.
- **Stale prose after the 8→12 shard raise**: `:30` still says *"Eight rather than four costs nothing"* and `:62` *"eight more entries per night"*. This PR carefully corrected the "~7 modules" comment that misled the previous reader; these two survived the same sweep.

## Verified by running

`classify-verification-sweep-failure.sh` falsified across all five states — `gap`+`could-not-run` → `gap`, all-`could-not-run` → `could-not-run`, empty dir → `unknown`, missing dir → `unknown` rc=0, junk content → `unknown`. `--selftest` passes 9/9 under the new exit-code contract. Both issue-body branches render, the gap body byte-identical in length to `main`'s. `workflow-run-step-size` PASS (largest step 14568 / 19000). Verdicts come from artifact files rather than job logs, so the "a log contains its own `run:` script" trap is genuinely avoided.

One methodology note worth recording: the first run of that falsification **failed spuriously on macOS bash 3.2**, which mis-parses `$(cat <<EOF)` inside an `if/else`. Re-run under bash 5 in Docker to match the runner. A shell-version artifact reads exactly like a real failure.

Issue #3927, opened 2026-08-06
97% of every OPA bundle is a byte-identical copy of shared content, duplicated 39 times — every governance edit rewrites 40 files and conflicts every in-flight service PR


JiRaska:
## Summary

`#3357` stopped the OPA bundles embedding `rules.yaml` verbatim — they now embed
`rules-opa-data.yaml`, the derived subset of top-level keys a `.rego` actually reads. **That fix was
never applied to `agents.yaml`**, which is still embedded whole into every service bundle.

Measured on `origin/main`:

```
bundle generators embedding agents.yaml   : 40
derived subset for rules.yaml             : rules-opa-data.yaml   exists
derived subset for agents.yaml            : none
`case_classes` read by any .rego          : 0
```

## The consequence

Any edit to `agents.yaml` restamps all 40 bundles and their `openbank.tech/policy-checksum` pod-roll
annotations — **so an agent-plane-only field rolls every policy-bearing pod in the fleet.**
`case_classes` is the worked example: no policy reads it, and changing it is a fleet-wide pod roll.

This is exactly the cost #3357 measured and removed for `rules.yaml` — replaying its last 80
revisions, only 16 touched anything a policy reads, while the other 64 restamped ~78 files for
nothing. `agents.yaml` is still paying that toll in full.

It also has a second-order effect visible right now: #3771 adds one charter entry and its redness is
three stale derived artifacts plus all 36 bundles — a single-entry change producing a 73-file
regeneration.

## Suggested fix

Mirror #3357: a `gen-agents-opa-data.py` producing the subset of `agents.yaml` that `.rego` sources
actually reference (derive the key set from the sources, never a hand-kept list — that was the point
of the original), have the bundle generators embed and hash that, and keep the unresolvable-key
ratchet #3816 added so a policy reading a key the subset cannot carry fails loudly rather than
resolving to `undefined`.

## Not verified

- I did not enumerate which `agents.yaml` keys the `.rego` sources DO read — only that `case_classes`
  is read by none. The subset's real size is unmeasured, so the saving is directionally certain but
  not quantified.
- I did not check whether any bundle depends on `agents.yaml`'s full document shape (e.g. iterating
  all charters) in a way a subset would break. That is the one thing that could make this harder than
  the `rules.yaml` case.
- Found while working #3737; not reproduced as a deploy incident.

Refs #3357, #3816, #3737, #3771

JiRaska:
[...]

PR #4173, opened 2026-08-08
refactor(governance): embed a derived subset of agents.yaml in OPA bundles, not the whole document

JiRaska:
## What this is

The largest coherent slice of #3927. It does **not** split the ConfigMap — it removes the
single biggest lump of unread bytes from all 42 bundles, which is the part of #3927 that can
land as one self-contained change with the existing guarantees intact.

`Refs #3927` — the structural split remains, scoped precisely at the bottom.

## Re-derived numbers (measured on `origin/main` @ `b4f10e3ab`, not quoted from the issue)

The issue says 39 bundles / 97%. Today it is 42:

```
bundle ConfigMaps                      42
total `data:` bytes              5 022 651   (5.02 MB)
  fleet-shared keys              4 850 690   (96.6%)
  per-service keys                 171 961   ( 3.4%)

distinct content per shared key (1 = byte-identical in ALL bundles):
  agents-data.yaml   n=42  distinct=1   2 116 380 B total
  rest.rego          n=41  distinct=1   1 188 754 B
  rules-data.yaml    n=41  distinct=1   1 157 266 B
  agents.rego        n=42  distinct=1     384 636 B
```

So the issue's headline holds and is slightly understated. `agents-data.yaml` alone is
**42% of the entire estate** and is the one shared key that was still embedded in full.

**How much of it any policy can reach.** `agents.rego` is the only production `.rego` that
reads `data.agents.*` at all, and it reads exactly two paths — `data.agents.agents` and
`data.agents.tool_tiers.deny` — plus exactly four fields off each charter: `id`, `skills`,
`tools.allow`, `tools.deny`. Everything else in `agents.yaml` (9 of 12 top-level keys, and
13 of 17 per-charter fields including `charter`, `description`, `model`, `limits`,
`data_scope`, `schedule`, `compliance`, `audit`, `requires_human`) is unreachable.

```
agents.yaml                50 390 B
derived subset              5 578 B      -88.9%
estate            5.02 MB -> 3.14 MB     -37.5%
```

**Granularity matters here and it is where the issue thread went back and forth.** A
top-level-key subset — the shape #3357 used for `rules.yaml` — strips ~7%, because
`data.agents.agents` **is** read and is most of the file. The waste is one level down: the
policy never reads an agent object wholesale. This PR projects *inside* each charter entry.

## The design, against the three guarantees the issue asks about

| guarantee | what happens |
|---|---|
| **per-service checksum** | unchanged in kind. Each generator still hashes its own inputs; one input's *content* is now the derived subset instead of the full charter document — exactly what #3357 did for `rules.yaml`. Every checksum in this PR moves once and then stops moving for unread edits. |
| **pod roll** | unchanged. `openbank.tech/policy-checksum` is still per bundle, still stamped into the Rollout/Deployment pod template by the same `sed`, still the thing that rolls pods because subPath mounts do not hot-reload. A policy change still rolls the fleet (correct — the policy changed). A change to data no policy reads now rolls **nothing**. |
| **a service whose bundle must diverge** | untouched. Divergence lives entirely in `<svc>_rest_ext.rego` + `manifest.json`, which no part of this PR reads or writes. The 3.4% that is genuinely per-service is byte-for-byte identical after this change. |
| **shared mutable object at runtime** | still none. Every bundle remains self-contained: same ConfigMap, same key name `agents-data.yaml`, same mount path `/bundle/agents/data.yaml`. Nothing new is shared at runtime, and nothing new can be out of sync in-cluster. |

## Why the derivation is trustworthy

The path set is **derived from the `.rego` sources, never listed** — same argument as
`gen-rules-opa-data.py`. A hand-kept list goes stale and `data.agents.<missing>` then
resolves to `undefined`, which in Rego is not an error, it is a rule that quietly stops
firing.

Two things guard the part that is genuinely harder than `rules.yaml`'s verbatim extraction:

1. **Failure degrades towards keeping MORE data.** The element analysis is a static pass
   over Rego text and is therefore incomplete by construction. It is written so that a
   variable bound to a charter list and then used in a way it cannot follow — a dynamic
   index, an unrecognised alias, a bare escape into a value — makes it **keep that list
   whole and print a `::notice::`**, rather than emit a projection it cannot justify. The
   unsound direction is not reachable by "the parser did not recognise this".
2. **A differential proof, not a static argument.** Before writing, the generator builds two
   OPA bundles that differ *only* in whether `agents/data.yaml` is the full document or the
   subset, and evaluates both over an input matrix derived from the charters themselves —
   every agent id (bare and `agent:`-prefixed), every tool pattern any charter or tier
   mentions, every chartered skill, the REST-bridge action space, plus negatives. Both
   `openbank/agents/decision` and `openbank/rest/allow` must be identical for **3978 MCP +
   3808 REST** cases. A single divergence aborts the write.

   `--self-test` **falsifies that proof**: it hands the generator a projection with `skills`
   and `tools.allow` removed and requires the proof to reject it. That runs in CI on every
   policy-relevant PR, ahead of `--check` — a gate that has only ever passed is unfalsified,
   and this one's failure mode (a silently narrower charter document) is invisible from
   every other angle.

`build-bundle.sh` and `docker-compose.yml` now stage the **derived** file too: the probe
must read the artifact the cluster runs, not a second copy of its source. All 15 of
`build-bundle.sh`'s real-charter decision assertions still pass against it, and it still
reports 16 charters.

## Verification run

| check | verdict |
|---|---|
| `gen-agents-opa-data.py --self-test` | PASS (incl. the differential-proof falsification) |
| `gen-agents-opa-data.py --check` | in sync, 3978 + 3808 decisions identical |
| `gen-rules-opa-data.py --check` | in sync |
| `build-bundle.sh` (opa test 168/168, `opa check --strict`, 15 decision assertions, `opa build`) | PASS |
| 22 `*_rest_ext_test.rego` suites (the `opa-policy.yml` shape) | PASS |
| gate `opa-bundle-parses` | PASS |
| gate `opa-sidecar-bundle-shape` | PASS |
| gate `opa-bundle-apply-size` | PASS |
| gate `no-dead-code-service-principal-rego-rule` | PASS |
| gate `gen-network-policies-drift-gate` | PASS |
| gate `loki-rule-load-test` (incl. self-test, real ruler) | PASS with `BASE_REF=main`; its first red was an unset `BASE_REF` making the scope `origin/`, not this diff |
| gate groups `registry` + `gitops` (45 gates) | 45 PASS with `BASE_REF` set; 3 advisory warns (`scheduler-liveness-adoption`, `incluster-hostname-resolution`, `operator-write-naming`) are pre-existing and untouched by this diff |
| `license-header-consistency`, `workflow-run-step-size`, `governance-script-unit-tests`, `ai-governance-snapshot-drift`, `feature-flag-governance` | PASS |

Regenerated in the documented order — `gen-rules-opa-data.py`, then
`gen-agents-opa-data.py`, then every `gen-*opa-bundle*.sh` — as the **last** step before
pushing, after the final source edit, not after the first.

Nothing generated was hand-edited. The ConfigMap key name, the mount path and the
`manifest.json` roots are all unchanged, so no consumer of the bundle shape moves:
`check-opa-sidecar-bundle-shape.py` and `check-opa-bundle-parses.py` both still pass, and
every reader of the *source* `agents.yaml` (`check-agent-charter-registry.sh`,
`gen-eu-ai-act.py`, `gen-ai-governance-snapshot.py`, `agentCharters.ts`) is unaffected —
`agents.yaml` itself does not change.

## What remains of #3927

This PR removes the largest single lump of duplicated bytes. It does **not** do the
structural split, which is the rest of the issue:

- `rest.rego` (1.19 MB), `rules-data.yaml` (1.16 MB) and `agents.rego` (0.38 MB) are still
  byte-identical in 41–42 copies — 2.97 MB, still 94.5% of the (now 3.14 MB) estate.
- A shared-policy edit still rewrites ~42 ConfigMaps + ~42 pod-roll annotations, so the
  conflict class in #3734 is reduced in frequency, not eliminated.

What a follow-up has to answer before splitting the ConfigMap, and which this PR
deliberately does not prejudge:

1. **ConfigMaps are namespaced.** "One shared ConfigMap" still has to be materialised in
   ~42 namespaces. Making the *repo* hold one copy therefore needs either a rendering step
   (Argo CD ApplicationSet / kustomize component) or a cluster-side replicator (Kyverno
   `generate`) — and the replicator option reintroduces exactly the shared mutable runtime
   object the current shape avoids, with a nastier failure: a subPath mount does not
   hot-reload, so a replicated ConfigMap can silently diverge from what the pods are
   serving, with no checksum anywhere saying so.
2. **The checksum then has to split with it.** A shared bundle needs its own checksum
   stamped into every pod template (fleet roll on a shared change, which is correct) while
   the per-service ext keeps its own (rolls only that service, which is what does *not*
   happen today).
3. Answered here, so the follow-up need not re-derive it: **the sidecar can take both from
   two ConfigMaps with no init container.** Every bundle key is already mounted
   individually by `subPath` into `/bundle/...`; a second volume mounting into the same
   directory tree is a pure manifest change.

## Note on shelf life

A bundle-touching PR conflicts fast — this one restamps 117 files under
`openbank-infra/gitops/components/`, including the `*-service.yaml` files the auto-deploy
bot bumps image pins in. Any governance PR that lands first will conflict it. It is not
merged here by design; whoever merges should do so promptly and with `--auto`, never
`--admin`.


JiRaska:
**Merge-order note: this PR and #4113 overlap on 73 files, and both are generated bundle restamps.**

Measured from the files API rather than the diff (`gh pr diff --name-only` returns nothing on a 125-file PR):

| | files | overlap |
|---|---|---|
| #4113 (`operator-mcp-session` M2M exclusion) | 77 | — |
| #4173 (derived `agents.yaml` subset) | 125 | **73** |

Both restamp `*-opa-bundle.yaml` and the pod-roll annotation in `*-service.yaml` across the fleet, so whichever lands second **will** conflict — and the resolution is never a hand-merge of a generated file. Whoever merges second should take `main`'s side on every generated file, then regenerate in the documented order as the **last** step before pushing:

```bash
```

Suggested order: **#4113 first** — it is smaller, it is a security fix (barring the shared M2M identity from minting MCP sessions), and its source change is one line of `rest.rego` that survives regeneration trivially. Then #4173, regenerated on top.

Both are `MERGEABLE` right now and both have a shelf life measured in hours. The two red checks on each — `Trivy` and `Customer-app dossier` — are inherited fleet-wide faults, fixed by #4153 and #4053 respectively; neither diff touches `package-lock.json` or `app-status.json`.

Merge with `--squash --auto`, never `--admin`.

JiRaska:
Reviewed. **Clean — no defect found.** This is the right shape, and the verification is the interesting part.

**The subset is complete, checked against what the policies actually read.** Grepping `data.agents[a-zA-Z0-9_.]*` across all 52 `.rego` files yields exactly two live paths — `data.agents.agents` and `data.agents.tool_tiers.deny` (everything else is a comment or a `with data.agents as …` mock in `agents_test.rego`/`rest_test.rego`). The field derefs off each charter are `a.id`, `a.tools.allow`, `a.tools.deny`, `a.skills` — exactly the four the projection keeps. Field-by-field against the source: 16/16 charters, ids in order, `skills` / `tools.allow` / `tools.deny` byte-equal, absent-`skills` agents preserved as absent, `tool_tiers.deny` equal — **0 mismatches**. No dynamic or aliased `data.agents[...]` anywhere, and `rest.rego:121` delegates to `data.openbank.agents.allow`, so it inherits the same data rather than reading its own.

That completeness check is the one that matters here: a key that silently disappears from the subset does not error, it evaluates against absent data, and `default allow := false` turns that into a denial nobody asked for.

**The bundles are current, not merely committed.** On the branch merged with current `origin/main` (clean merge), re-running the full derived chain — `gen-rules-opa-data.py`, `gen-agents-opa-data.py`, then every `gen-*opa-bundle*.sh` — changes **0 files**. Worth doing explicitly, because a conflict-free merge of a generated file says nothing about whether it matches its sources; that cost #3807 a 246-commit-stale bundle earlier this week. `--self-test` also genuinely rejects a projection with `skills`/`tools.allow` dropped (`::error::the derived subset changes openbank/agents/decision`), and `opa-policy.yml:57` still lists `agents.yaml` in the relevance pattern, so a source edit still triggers `--check`.

**Decision equivalence, measured rather than argued.** Extracted the `data` keys from `agent-opa-bundle.yaml` and `ledger-opa-bundle.yaml` on `main` and on the branch (only `agents/data.yaml` differs in each) and ran `opa eval` over both:

| bundle | query | cases | divergent |
|---|---|---|---|
| agent | `data.openbank.agents.allow` + `.decision` | 1485 | **0** |
| ledger | `data.openbank.rest.allow` | 258 | **0** |

Non-vacuous: 106 allows in the first, 23 AI_AGENT allows preserved in the second, with controls in both runs (`agent:compliance-officer/account.read` ALLOW, `*/money.transfer` DENY on the hard tier; HUMAN+`ROLE_OPERATOR` `ledger.read` → `matrix-allows`, ANONYMOUS → `false`). And the harness itself was falsified — stripping `skills` from the new data produces 10 divergences, so it can see a change when there is one.

**One merge-order note, not a defect:** this overlaps #4113 on 73 files. Both are generated restamps, so whichever lands second must be **regenerated**, not merge-resolved — resolving a generated file by hand produces something that matches neither side's sources.

Issue #1502, opened 2026-07-17
settlement-service emits zero audit events (DORA Art. 17 reconstructability gap)


JiRaska:
## Summary
Found during #1021's lineage-code-audit triage. \`settlement-service\`'s
governance.yaml declares a downstream edge to \`audit-service\`
(\`relationType: topic\`, "emits settlement audit events"), but the service
has no Kafka messaging config at all and doesn't use the shared
\`AuditEventPublisher\` (\`openbank-libs-runtime\`) that every other money-path
service uses.

## Why this matters
\`settlement-service\` is money-path (debit→credit→ledger-booking saga,
\`SettlementWorkflowImpl\`, ADR-0101). A money-path service that emits zero
audit trail is a genuine reconstructability gap (DORA Art. 17), not stale
metadata — this is why it was flagged separately from the plain stale-claim
(b) findings in #1021's triage, and why the governance.yaml claim was
deliberately left in place (PR #1499) rather than deleted: it describes
intended behavior that doesn't exist yet, not a wrong claim.

## Ask
Wire \`settlement-service\` into the shared \`AuditEventPublisher\` the same
way other money-path services do, so settlement actions produce a
reconstructable audit trail.

Refs #1021, #1499.

PR #4179, opened 2026-08-08
chore(governance): resolve every unbacked service-lineage edge and enforce the audit

### LINKED ISSUE #3982: 30 of 101 declared service-lineage edges have no backing code — the lineage view, service map and catalog graph all render them
`check-governance-lineage.py` on current `origin/main`:

```
32 downstream + 69 upstream edge(s) declared across 56 service(s) with a governance.yaml,
101 distinct relationship(s) checked; 30 unallowlisted violation(s), 5 allowlisted,
10 naming no in-tree service.
```

**30 of 101 declared service dependencies have no backing code.** The gate is `mode: advisory`
(`.github/gates/gates.yaml`), so this has been printing on every run and blocking nothing.

## The 30

```
api (26)
  account-service        -> customer-edge, party-service
  analytics-sink         -> campaign-service
  card-issuance-service  -> account-service
  clearing-service       -> transaction-service
  consent-service        -> account-service, notification-service
  dispute-service        -> card-issuance-service
  fx-service             -> transaction-service
  kyc-service            -> onboarding-service, party-service
  ledger-service         -> finrep-service, settlement-service
  party-service          -> customer-edge, onboarding-service, pid-service
  pid-service            -> delegation-service
  product-catalog        -> document-service
  psd2-service           -> sca-service            (declared twice, see below)
  sca-service            -> customer-edge, delegation-service, document-service
  standing-order-service -> transaction-service
  statement-service      -> finrep-service
  transaction-service    -> account-service

topic (4)
  kyc-service            -> notification-service
  sepa-payment           -> settlement-service
  settlement-service     -> audit-service
  transaction-service    -> notification-service
```

## Spot-checked two before filing, both hold

```
$ git grep -n 'RegisterRestClient' origin/main -- openbank-transaction-service/src/main
  BalanceCoverClient.kt:32   @RegisterRestClient(configKey = "balance-service")
  FxRateClient.kt:26         @RegisterRestClient(configKey = "fx-service")
  LedgerRestClient.kt        …
```
No account-service client. The only occurrence of "account" in its `application.yaml` is inside a
comment.

```
$ git grep -n 'RegisterRestClient' origin/main -- openbank-consent-service/src/main
  ScaChallengeClient.kt:24   @RegisterRestClient(configKey = "sca-service")
```
No notification-service client.

So these are not heuristic noise — the declared edges are absent from the code.

## Why it matters beyond tidiness

`lineage.upstream` / `lineage.downstream` in each `governance.yaml` is the source for, at minimum:

```
openbank-admin-ui/scripts/generate-service-graph.mjs
openbank-admin-ui/scripts/generate-governance.mjs
openbank-admin-ui/src/app/docs/lineage/page.tsx
openbank-admin-ui/src/app/docs/service-map/page.tsx
openbank-admin-ui/src/app/api/catalog/graph/route.ts
```

A wrong edge is therefore a wrong answer in the lineage view, the service map and the catalog graph
simultaneously — and those are exactly the artefacts someone consults to answer "what breaks if I
change this". The failure direction is the bad one: a **phantom** edge makes a change look riskier
than it is, so it inflates impact analysis rather than under-reporting it, and there is no signal
distinguishing the two. This is the #889 class the gate's own message names.

## Two structural notes for the fix

1. **`psd2-service -> sca-service` appears twice**, once as `openbank-psd2-service->sca-service` and
   once as `psd2-service->openbank-sca-service`. Declaring both ends of a relationship is the schema's
   design (32 downstream + 69 upstream ≠ 101 by accident), but the `openbank-` prefix is applied
   inconsistently between the two halves. Any cleanup should settle on one spelling, or a genuine edge
   will keep reading as two.
2. **10 edges name no in-tree service at all** and are counted separately from the 30. Some of those
   may be deliberate (another repo, e.g. the `openbank-app` split), which the gate cannot know. That
   distinction is worth encoding — `rules.yaml: lineage_code_audit.allowlist` exists for exactly this
   and takes a one-line reason — so that "tracked elsewhere" stops looking identical to "wrong".

## Suggested handling

Triage the 30 into (a) delete the edge, (b) the dependency is real but implemented in a way the
heuristic cannot see → allowlist with a reason, (c) the dependency is real and the code is missing →
its own issue. Then flip the gate to `mode: enforced`: it is a derived check, both sides read from
committed files, so a red carries no judgement to exercise.

This is the largest single finding among the advisory gates measured in #3931 and the only one I would
not propose enforcing immediately — the backlog has to be triaged first.

JiRaska:
## Re-derived count first: **31 of 103**, not 30 of 101

The fleet grew since the issue was filed (58 services with a `governance.yaml`, not 56).
Measured on `origin/main` with the unmodified script:

```
32 downstream + 71 upstream edge(s) declared across 58 service(s) with a governance.yaml,
103 distinct relationship(s) checked; 31 unallowlisted violation(s), 5 allowlisted,
11 naming no in-tree service.
```

The one finding the issue's list does not have is `agent-service -> case-coordinator-agent:api`.
"103 distinct relationships" was itself wrong — see the canonicalization fix below; there are 81.

## The issue's default action would have been wrong for a third of them

The suggested handling was "delete the edge" as the default. For **11 of the 31** that would have
been a NEW defect of exactly the kind the issue is about: the relationship is real and backed by
code, declared with **source and target swapped**. Deleting them removes a true dependency from
the lineage view, the service map and the catalog graph — under-reporting instead of inflating,
which is the worse direction for "what breaks if I change this".

They were flipped to the correct side instead, each with an in-place note naming the client and
the Deployment env entry that witness the direction. A worked example: `finrep-service` declared
`upstream: ledger-service`, but `LedgerClient(configKey = "ledger-service")` lives in
finrep-service and `ledger-service`'s only client is fx-service. finrep reads the ledger. (This is
the same edge whose wrong *path* shipped as #2269.)

## Per-bucket classification, all 31

| bucket | n | disposition |
|---|---|---|
| **(b) real, invisible to the detector — REVERSED** | 11 | flipped to the correct direction |
| ...of which also invisible to the heuristic | 3 | + allowlisted (customer-edge BFF) |
| **(c) aspirational** | 1 | removed |
| **duplicate of another finding** | 1 | one relationship counted twice |
| **(a) genuinely stale** | 18 | removed, each with what was checked |

**Reversed (11)** — finrep/ledger, delegation/sca, delegation/pid, party/account, party/kyc,
transaction/fx, account/transaction, onboarding/party, notification/consent, document/sca,
document/product-catalog. Plus the three customer-edge hops below.

**Reversed AND structurally invisible (3)** — `customer-edge -> party/account/sca`. customer-edge
is a BFF with **zero** `@RegisterRestClient` interfaces; every hop is a Vert.x WebClient built
from an `openbank.edge.<svc>-url` `@ConfigProperty`. Its Deployment env carries 25
`http://<svc>.<ns>.svc:<port>` entries, these three among them. Allowlisted with that reason.

**Aspirational (1)** — `case-coordinator-agent -> agent-service`. The module's `src/main` contains
no reference to agent-service and no REST client of any kind, and it has no gitops component, so
it is not deployed. Declared ahead of ADR-0244's code. Removed with a note to re-add it in the PR
that adds the client.

**Duplicate (1)** — `psd2-service -> sca-service` under two `openbank-` spellings, as the issue
noted. Both halves are stale (psd2's only REST client is `ConsentServiceRestClient`; SCA is
reached indirectly via an `scaRedirect` link) and both are removed. The double-count is fixed
structurally, not by hand — see below.

**Stale (18)** — including `standing-order-service -> transaction-service`, the founding example
in the gate's own docstring. Two worth calling out:
- `sepa-payment -> settlement-service:topic`: settlement-service produces and consumes **zero**
  Kafka topics. sepa-payment's `SettlementAdapter` calls **transaction**-service over REST, and
  that edge is already declared, so nothing is lost.
- `settlement-service -> audit-service:topic`: the #1502 fix emits `AuditEvent`s onto the shared
  libs pipeline, whose only implementation is `LoggingAuditEventPublisher` — it logs, it does not
  publish. audit-service consumes 21 topics, none produced by settlement-service. The declaration
  describes an intent the wiring does not deliver.

## What else derives from these declarations

Checked before removing anything, because deleting a real edge is the expensive mistake here.

- **NetworkPolicies do NOT.** `openbank-infra/scripts/gen-network-policies.py` derives allow-lists
  from the GitOps Deployment env, Ingress backends and Kafka bootstrap URLs — never from
  `governance.yaml`. So no lineage edit can change what traffic is allowed. That also makes the
  Deployment env an **independent witness** for direction, which is what every note in this diff
  cites alongside the client.
- **admin-ui** — `generate-service-graph.mjs`, `generate-governance.mjs`, the lineage view, the
  service map, `/api/catalog/graph`. Both derived artifacts (`governance.json`, `catalog.json`)
  are gitignored and regenerated by `pretest`, so there is nothing to commit.
- **`openbank-infra/scripts/generate-service-runbooks.py`** — a regex parser over the raw file, so
  it was re-run against all 16 touched services to confirm the new shape (and the added comments)
  parse correctly.

## Two silent detector bugs fixed

1. **A trailing `-api` on a configKey read as no backing code at all.** Nine in the fleet
   (`account-api`, `account-service-api`, `balance-api`, `document-api`, `ledger-api`,
   `party-api`, `party-service-api`, `product-catalog-api`, `transaction-api`). document-service's
   live `ProductCatalogClient(configKey = "product-catalog-api")` looked absent. No service
   directory ends in `-api`, so the strip cannot collapse two distinct services.
2. **Edge keys were never canonicalized**, so the docstring's "a relationship declared from both
   ends is checked and allowlisted ONCE" was simply false — 103 declarations produced 103 keys.
   They now resolve to the service DIRECTORY name: **103 declarations, 81 distinct
   relationships**. This is what collapsed the psd2/sca pair, and what made the existing
   `transaction-service -> account-service` allowlist entry actually cover the same relationship
   declared from account-service's end. A name with no directory (github, prometheus, temporal)
   keeps its literal spelling — inventing `openbank-github` would be a lie.

## The allowlist can now go stale, and that is a finding

Point 6 of the brief: a baseline that can only grant rots into permanent furniture. An entry
survives its edge being deleted, and it survives the code growing the very client its reason swore
would never exist. **It caught one the moment it was written:**

```
::error::lineage-code-audit: rules.yaml lineage_code_audit.allowlist entry
openbank-interest-service->openbank-account-service:api no longer applies
```

That entry read *"reaches the account indirectly ... **never a direct account-service call**"*, and
interest-service now holds `@RegisterRestClient(configKey = "account-service")`. Nothing would
ever have said so. Entry deleted.

## Falsified BOTH directions, exit codes read directly (never through a pipe)

| case | exit | violations | stale entries |
|---|---|---|---|
| clean tree (baseline) | **0** | 0 | 0 |
| + a new unbacked edge (`pid-service -> ledger-service`) | **1** | **1**, named | 0 |
| + allowlist entry for an undeclared relationship | **1** | 0 | **1**, named |
| + allowlist entry for an edge that IS backed | **1** | 0 | **1**, named |
| restored | **0** | 0 | 0 |

Both stale shapes are exercised because they arise from different code and only one has ever been
seen in the wild.

## Enforced, with a self-test

`mode: enforced` in `.github/gates/gates.yaml` (group `registry`, so it runs unconditionally, not
inside a path-filtered job), `--enforce` on the command, plus `selftest` /
`selftest_expect: pass`. The self-test builds a synthetic tree and asserts **10 cases in both
directions** — including the `-api` strip and the mirrored-edge dedup, so neither fix can silently
regress. It was verified to go **red (8/10, exit 1)** against a deliberately broken copy of the
script; a gate that has only ever passed is unfalsified.

Through the real runner:

```
$ python3 .github/scripts/run-gates.py --only governance-lineage-vs-code-audit
PASS [registry] governance lineage-vs-code audit (ADR-0160, enforced)
  --- self-test (must PASS) ---
  check-governance-lineage --self-test: PASS (10/10 cases, both directions)
  44 downstream + 42 upstream edge(s) ... 66 distinct relationship(s) checked;
  0 unallowlisted violation(s), 8 allowlisted, 0 stale allowlist entr(ies),
  11 naming no in-tree service.
```

The full `registry` group is 19 gates, 18 PASS + 1 pre-existing advisory WARN
(`operator-write-naming`, untouched by this PR and red on main).

## The 11 naming no in-tree service

Left as they are, deliberately. All 11 are genuinely external — `github` ×5, `prometheus` ×3,
`temporal` ×2, `alertmanager` ×1 — and the script already reports them as their own `::notice::`
class with an explicit "there is no in-tree code that could witness it either way". The issue
floated allowlisting them to distinguish "tracked elsewhere" from "wrong"; that is now the wrong
move, because the stale-entry check would correctly flag every one of them as an exemption that
suppresses nothing.

## Release scope

Typed `chore(governance)` on purpose. `governance.yaml` is outside `exclude-paths`, so a
releasing type would cut a release for **16 services** whose shipped code did not change.

Closes #3982


JiRaska:
**Re-merged `main`; the gate passes again (`10/10` self-test, both directions). But this PR is on a treadmill and that is structural, not bad luck.**

It has now gone red on `gates (gitops)` **twice** within one session, both times without anyone touching this branch. The cause each time: `main` gained or changed a service-lineage edge, so the committed `lineage_code_audit.allowlist` in `rules.yaml` no longer matched the tree — and this gate deliberately fails on a **stale** allowlist entry as well as a new violation, which is the right design and is exactly what makes the PR unstable while it waits.

Every merge to `main` that adds a `@RegisterRestClient`, removes one, or edits a `governance.yaml` lineage block can redden it again. The window in which it is green is bounded by the next unrelated merge.

Two consequences worth stating plainly:

- **It should merge promptly**, not sit. Each re-merge is a fresh chance for the "clean merge that silently drops a neighbouring entry" failure this repo has recorded — the allowlist is exactly the shape most exposed to it (a list every service registers itself in).
- **The remedy after it lands** is to stop committing the set at all. The gate already derives the *violations* from the sources; the allowlist is the last hand-kept half, and this repo's own rule is that a gate's coverage set must never be maintained separately from the artifacts it covers. Deriving it — or keying entries so an unrelated edge cannot invalidate them — removes the treadmill rather than riding it.

Verified after this merge rather than assumed: zero conflicts, gate PASS through the real runner with `PR_DIFF_BASE` set to the merge-base.

JiRaska:
`gates (gitops)` — `service runbook drift (issue #2255, enforced)`. This PR edits `governance.yaml` lineage (upstream/downstream) across multiple services but doesn't include the regenerated runbooks that derive from it.

Gate output: `runbooks: 44 written, 0 kept (existing)` — every runbook the generator touches disagrees with what's committed, e.g. `docs/runbooks/svc-account.md`'s dependency section is stale for the new lineage this PR declares.

Mechanical fix: run the runbook generator after this PR's `governance.yaml` changes and commit the diff, same as the OPA-bundle-regen pattern for `rules.yaml` edits — regenerate the derived artifact, don't hand-edit it.

(Also noted an advisory-only warning on `scheduler liveness adoption (ADR-0237, advisory)` — not blocking, flagging for awareness in case it's relevant to the lineage changes.)

JiRaska:
Found sweeping the open queue: `gates (gitops)` fails on `service runbook drift` (issue #2255, enforced) — separate from the two case-coordinator failures on this PR (`event-consumer liveness`, `outbox dispatcher`, both inherited from #4215's dropped commit and unrelated to this PR; already flagged on #4232/#4233).

This one's this PR's own: `docs/runbooks/svc-*.md` are generated from `governance.yaml`, and this PR touches lineage relationships in it without regenerating them. Fix:

```
python3 openbank-infra/scripts/generate-service-runbooks.py --force
git add docs/runbooks/
```

Then re-run `Validate manifests` locally to confirm before pushing — no need to touch anything else for this one.

JiRaska:

```
```

PR #4180, opened 2026-08-08
ci(gates): graduate scheduler-liveness-adoption to enforced, and name what blocks the rest (#3931)

JiRaska:
## What this does

Re-derives the advisory gate set from `origin/main`, graduates the one gate that could be
made clean, and records — with today's measured numbers — exactly what blocks each of the
rest. `Refs #3931` rather than `Closes`, because five gates remain advisory on purpose and
they are named below.

## The census: 119 gates, 8 advisory — and BOTH forms of "advisory" checked

`mode: advisory` in `.github/gates/gates.yaml` (form **a**) and `::warning` + exit 0 unless
`--enforce` inside the script (form **b**) are independent, and a sweep for one reports the
other as enforced. Every one of the 8 scripts defines a real `--enforce` (matched on
`add_argument("--enforce"` / `"--enforce" in sys.argv` / a `--enforce)` shell case, not on
header prose):

| gate | form (a) `mode:` | form (b) script-side | run line passes `--enforce` | runner verdict today |
|---|---|---|---|---|
| `scheduler-liveness-adoption` | advisory | yes | no | **WARN-worthy: rc=1, downgraded** |
| `governance-lineage-vs-code-audit` | advisory | yes | no | **PASS** (31 findings) |
| `dotted-mp-messaging-key-guard` | advisory | yes | no | **PASS** (16 findings) |
| `operator-write-naming` | advisory | yes | no | WARN |
| `incluster-hostname-resolution` | advisory | yes | **yes** | WARN |
| `release-scope-mismatch-gate` | advisory | yes | no | needs `PR_DIFF_BASE` |
| `db-migration-gate` | advisory | yes | no | needs `PR_DIFF_BASE` |
| `schema-compat-gate` | advisory | yes | no | needs `PR_DIFF_BASE` |

Seven are doubly advisory. The `PASS` rows are the point: a doubly-advisory gate exits 0, so
`run-gates.py` prints **PASS** while the gate prints 31 and 16 findings. That is where this
issue's "two find nothing and can graduate free" came from — it was read off the runner
verdict, not the gate output. Neither finds nothing.

## What graduated: `scheduler-liveness-adoption`

The script was already the ratchet the issue asks for — `if new_violations or healed: return 1`,
both directions. `mode: advisory` converted that to a warning at the runner
(`run-gates.py:333-338`), so the forward half never blocked. Measured on `origin/main`:

```
domain @Scheduled files: 45 (28 with liveness, 65 exempt outbox/allowlist)
::error file=openbank-account-service/.../SavingsProposalExpiryScheduler.kt::NEW domain
  @Scheduled job without registerWorkflowLiveness
baselined (adoption pending, 16 left)
baseline entries now compliant — REMOVE them from BASELINE: <11 files>
rc=1
```

So it had let a **second** violation through, after #3875's `ConversationRetentionScheduler`:
account-service's savings-proposal expiry sweep has been running with no heartbeat, and that
sweep swallows its own exception — nothing throws, nothing logs, and "0 expired" is the normal
case, so a permanently broken sweep is externally identical to a healthy one.

Three changes, in order:

1. **The heartbeat** (`fix(account)`): registered off `StartupEvent`, `recordSuccess` only on
   the branch where `expireStale` returned. Three unit tests, including the load-bearing one —
   a swallowed failure must leave the heartbeat stale.
2. **BASELINE 27 -> 16.** All 11 money-path entries had adopted the heartbeat and were never
   removed, so the "healed" branch had been firing continuously and nobody saw it. The
   money-path block is now empty.
3. **`mode: advisory` -> `enforced`**, and `rules.yaml: liveness_adoption` `advisory` ->
   `enforce` (its ADR-0144 target was 2026-09-30). The 16 remaining files stay grandfathered;
   only a new violation or a stale entry is red.

`gen-rules-opa-data.py` re-run: **no change** — `scheduled_methods` is not in OPA's read set,
so no bundle restamp.

## Falsification — both directions, on the real tree, exit codes read directly

| direction | counter-example | reached the code | rc |
|---|---|---|---|
| NEW violation | a probe `@Scheduled` `.kt` with no liveness | domain files **45 -> 46**, exactly **1** `::error` naming the probe | **1** |
| (control) | probe removed | 45 | **0** |
| stale baseline | a compliant file added to `BASELINE` | entries **16 -> 17**, exactly **1** healed reported and named | **1** |
| (control) | reverted | 16 | **0** |
| the new test | `recordSuccess` moved out of the success branch | `3 tests, 1 failure` — and the failure is `a swallowed sweep failure leaves the heartbeat stale` | **1** |
| (control) | restored | `3 tests, 0 failures` | **0** |

Runner: `run-gates.py --only scheduler-liveness-adoption` -> `PASS=1`.
`run-gates.py --self-test` -> green. Full `--all` -> `119 gates PASS=111 WARN=2 FAIL=5
UNFALSIFIED=1`; the 5 FAILs and the UNFALSIFIED are all "`PR_DIFF_BASE` is empty — refusing to
run vacuously" / "`BASE_REF: unbound variable`", i.e. the local environment, and two of them are
pre-existing **enforced** gates. The 2 WARNs are `incluster-hostname-resolution` and
`operator-write-naming`, both out of scope here.

Local gate on account-service: `compileKotlin`, `ktlintCheck`, `detekt`, `quarkusBuild`
(CDI wiring — the new constructor parameter) and the new test all green.

## What stays advisory, and exactly why

| gate | blocker |
|---|---|
| `governance-lineage-vs-code-audit` | 31 unallowlisted of 103 relationships. Triage is **#3982**; 11 edges name no in-tree service and may be deliberate. The flip is that issue's last step. |
| `dotted-mp-messaging-key-guard` | 16 findings. It also duplicates the **enforced** `check-kafka-dotted-keys.py`, so the right move is **#3928** — widen the enforced sibling and retire this one, not graduate two overlapping gates. |
| `db-migration-gate` | 4 migrations without a `-- Rollback:` note over the last 200 commits. Dated 2026-08-15. |
| `schema-compat-gate` | Clean over a 200-commit aggregate diff — which is **not** evidence. It is diff-scoped, an empty diff always passes, and a 200-commit lump is not any PR's diff. It also has no allowlist, so enforcing it would hard-block a deliberate breaking event change. Dated 2026-09-15. |
| `release-scope-mismatch-gate` | Same vacuity problem, plus it is a genuine judgement gate (a hidden commit type is sometimes the correct answer). Dated 2026-10-01. |
| `incluster-hostname-resolution` | **Out of scope — #3974 flips it.** |
| `operator-write-naming` | **Out of scope — #4113 flips it** (with the `operator-mcp-session` fix). |

## Two stale advisory notes corrected

The issue's own comments make the point that an advisory gate's *justification* is an artifact
nobody re-measures. Two were wrong:

- `mp_messaging_dotted_keys.blocked_on` said **"1 finding remains fleet-wide"** with no date.
  Measured today: **16**.
- `lineage_code_audit.blocked_on` now carries today's numbers and the note that the runner
  reports this gate **PASS**, not WARN, because it is advisory inside the script as well.

Refs #3931, #3345, #3982, #3928, #3875, ADR-0237, ADR-0144


JiRaska:
Refreshed against `main` (was 6 behind, conflict-free). Most of the red was inherited, not yours.

**The `gates (data)` failure is gone.** It was the duplicate `authz:` key under `%test` in engagement's `application.yaml` — a `main` problem that has now been fixed twice: #4160 merged the two blocks, #4106 re-added one as a pure insertion from a stale base (+22 lines, 0 deletions, stale comment included), and #4184 removed it again. Nothing in this PR was ever involved.

**Verified on the merged tree**, both gates this PR is actually about:

```
run-gates.py --only duplicate-yaml-key-guard      exit 0
run-gates.py --only scheduler-liveness-adoption   exit 0
```

That second one matters: this PR graduates `scheduler-liveness-adoption` to **enforced**, and it passes — because the PR also supplies the heartbeat that was missing. Graduating a gate in the same change that clears its last violation is the right pairing; the alternative is an enforced gate that reddens `main` on merge.

**On that heartbeat specifically:** `SavingsProposalExpiryScheduler` is the job I reported on #3345 — it landed in #3632 with no `registerWorkflowLiveness`, which made it invisible to the staleness rule and to the sentinel. That is the failure mode where a scheduler silently never runs and the only symptom is a table that stops growing. Good to see it closed at the same time as the gate that catches it, rather than as a follow-up nobody does.

**Still red and not mine to fix:** `Customer-app dossier` — *"the committed app-status.json no longer matches openbank-app source (1 field(s))"*. That derives from a different repository, so it cannot be regenerated from here. Worth knowing that #4053 narrows exactly this check to drift a PR could plausibly have caused; once that lands, a governance-only PR like this one stops inheriting the dossier's freshness problem. Until then it needs either a dossier refresh or a maintainer's judgement.

JiRaska:
Reviewed and verified by running. **The gate itself is sound — derived scope, falsified in both directions, self-test present and correct.** The blocker is timing, not design, and I have taken care of half of it.

## Blocking: red against current `origin/main`, green on its own

```
PR tree alone                    rc=0
PR tree merged with origin/main  rc=1
::error openbank-fraud-service/.../FraudHoldService.kt: NEW domain @Scheduled job
        without registerWorkflowLiveness
```

That came in with #4252 (`feat(fraud): raise a marketing-suppression fraud-hold signal`), merged after this branch was cut. **The PR reads green and `CLEAN` right now** because its `gates (gitops)` run predates it — a check result is about the base, not about `main`. Merge as-is and `main`'s own push run goes red, with no PR to carry it, which is the exact failure class `main-red-watch` exists for.

**A correction to how this should be read, though.** `main` today has **two** violations, not one:

```
MAIN  rc=1  errors=2
  openbank-account-service/.../SavingsProposalExpiryScheduler.kt
  openbank-fraud-service/.../FraudHoldService.kt
```

It does not block because the gate is still `advisory` there. This PR fixes the first and graduates to enforced; the second is the one it cannot know about. So the situation is not "this PR is broken" — it is "graduating to enforced makes both debts due, and one arrived 30 minutes ago".

**I have fixed the fraud one in a separate PR** rather than widening this one's scope, and verified the composition:

```
#4180 + origin/main + the fraud fix    rc=0
```

Merge that first, rebase this, and this goes green. Sequence it deliberately and merge promptly — the race is real, and the next PR adding a scheduler re-arms it. That is the intended cost of enforcing, not a defect.

The fraud fix is worth doing on its own merits regardless: `sweepExpired` catches and logs its exception, so a permanently broken sweep is indistinguishable from a healthy quiet one — no exception escapes, no metric moves, and `cleared=0` is the normal case. That is precisely the ADR-0237 rationale.

## Falsification — both directions, verified

| | injected | result |
|---|---|---|
| new violation | probe `.kt` with `@Scheduled`, no liveness | 1 `::error` naming the probe, rc=1 |
| control | probe removed | rc=0 |
| stale baseline | compliant file added to `BASELINE` | "baseline entries now compliant — REMOVE them", rc=1 |
| control | reverted | rc=0 |

## Scope is derived, and the numbers agree

`grep -rlE '^\s*@Scheduled\b'` over `openbank-*/src/main/kotlin` returns **110 files**, 110 occurrences. The gate's `classify()` gives 29 compliant + 16 violations + 65 exempt = **110**. Full coverage, no gap, and no `@Scheduled` exists outside that tree. The hand-kept parts are the baseline (16) and allowlist (2), both of which fail on staleness — the sanctioned shape.

The `run:` deliberately omitting `--enforce` while `mode: enforced` is right: it blocks the two ratchet branches while the 16 baselined files stay grandfathered. No `selftest_expect` defaults to `pass`, correct for a self-test harness. `check-advisory-gate-registration.py` and `check-gate-graduation.sh` both stay green, and dropping `target_enforce_date` alongside the graduation matches what five other entries do.

## Worth fixing: the "name what blocks the rest" half is under-delivered

The PR body tabulates blockers for all seven remaining advisory gates, but `rules.yaml` gains updated `blocked_on` prose for only **two** (`lineage_code_audit`, `mp_messaging_dotted_keys`). `db-migration` (`rules.yaml:281`), `schema-compat` (`:298`) and `release-scope-mismatch` (`:432`) keep their unchanged notes.

The PR's own thesis is that an advisory gate's justification rots because nothing gates it. Leaving three of them measured only in a PR body reproduces exactly that — a PR body is the one place a claim cannot be re-derived from. The two it did correct are good work; the `"1 finding remains fleet-wide"` → 16 correction is dated and carries the #3928 sequencing.

## Notes

- **`SavingsProposalExpiryScheduler`** is done right: `@Observes StartupEvent` registration avoids the lazy-`@ApplicationScoped` trap, `recordSuccess()` sits only on `.onSuccess`, and the load-bearing test is the one that matters (*a swallowed sweep failure leaves the heartbeat stale*). `EXPECTED_INTERVAL` is hardcoded to 10 min while the cron is config-overridable — the KDoc admits it, and a widened cron makes the gauge over-strict rather than lax, which is the safe direction.
- **Exemption is file-scoped**: `LIVENESS_RE` matches anywhere in the file, so a file with two `@Scheduled` methods where only one registers liveness passes wholesale. Harmless today — verified exactly one `@Scheduled` per file — but 65 of 110 files are exempt, so that is where a future escape lands.

PR #4208, opened 2026-08-08
fix(libs): seed the workflow-liveness age gauge at registration, not at Instant.EPOCH

JiRaska:
## The contradiction, verified on `origin/main` @ `d3083276`

`DomainMetrics.registerWorkflowLiveness` seeded its age gauge from `Instant.EPOCH`:

```kotlin
val lastSuccessEpochMillis = AtomicLong(java.time.Instant.EPOCH.toEpochMilli())
```

So `openbank_workflow_last_success_age_seconds` read **~1.8e9 seconds — decades** — for any
workflow that had not yet recorded a success, including every workflow on a freshly started pod.

`prometheus-rules-workflow-liveness.yaml` nonetheless ships `WorkflowLivenessStale`
(`age_seconds > 2 * expected_interval_seconds`, `for: 15m`) and its comments asserted the opposite
of the code — "seeded at registration — a never-run job reads as old as its pod, never as decades"
and "Boot-safe by construction (ADR-0237)". `DomainMetrics`' own KDoc described the EPOCH behaviour
correctly and warned it was "right for a sentinel finding and wrong for a 3am page" (#2239 Gap 2);
the rule shipped anyway.

**Consequence, confirmed.** For a daily job (`Duration.ofDays(1)`) the threshold is 2 days and a
fresh pod reported decades, so `WorkflowLivenessStale` fired 15 minutes after every deploy or
restart, for every daily workflow, and kept firing until that workflow's next success — up to 24h.
No `for:` duration helps, because the condition genuinely persists. There are ~28
`registerWorkflowLiveness` call sites, so this is fleet-scale noise on a required-observability
control, and noise here is exactly what makes a real stale scheduler unnoticeable.

## The fix

**Seed the gauge at `Instant.now()` in `registerWorkflowLiveness`** — the first of the two options
the KDoc named, and the one ADR-0237 and the rule already assumed. Detection is unchanged: a job
that has never run still crosses its 2× threshold once its own grace elapses, exactly like a job
that stopped running.

### What depends on the EPOCH semantics, and what that cost

I checked the consumer the task flagged. `openbank-control-liveness-sentinel` (ADR-0163 D1) is
purely threshold-based — `DetectFindingsActivityImpl.detectStaleHeartbeats` compares
`ageSeconds` against `interval × warn/critical multipliers` and special-cases nothing, so its
verdicts are unaffected. What the seed does destroy is the ability to tell **"has never succeeded"**
from **"succeeded a second ago"**: with a boot-seeded age both are a small number.

That bit is published rather than lost. The primitive now emits a third gauge,
`openbank_workflow_success_recorded` (`0` until the first `recordSuccess`, `1` after):

- the **alert does not read it** — it needs no new threshold, and a `never-succeeded` rule would be
  redundant with `WorkflowLivenessStale` given the boot-seed;
- the **sentinel does**, so a finding says *"no success in the 200s since this pod registered it"*
  instead of asserting a last success that never happened. Same threshold, same severity, different
  sentence.

It is not a field nothing reads.

An absent or unparseable flag is read as **has succeeded**, so a rolling upgrade past pods that
predate the gauge cannot manufacture never-succeeded findings for the whole fleet.

### The residual gap, stated rather than left implicit

A job whose pod restarts **more often than 2× its interval** can no longer accumulate enough age to
alert. That case is covered statically instead — by the registration gate
(`check-scheduler-liveness.py`, ADR-0237 point 4) and the HR000068 guards — and is now written down
in the rule's own comments, the `DomainMetrics` KDoc and ADR-0237, rather than being a property
someone has to re-derive.

## Tests — pinned by value, because `isNotNull()` passes against the bug

New in `openbank-libs-runtime`:

- `a freshly registered workflow reads an age of seconds, not decades` — asserts
  `age < Duration.ofHours(1)`. An upper bound is the only assertion shape that can fail against
  ~1.8e9; `isNotNull()` would have passed against the broken value for the whole life of this gauge.
- `a freshly registered workflow reports that it has not succeeded yet` — `success_recorded == 0`.
- `recordSuccess resets the age gauge close to zero and flips the success flag`.
- `WorkflowLivenessMetricNamingTest` covers the third series through Micrometer's real
  `PrometheusNamingConvention`, same as the other two.

**The fleet sweep is not cosmetic.** 18 service test files asserted
`age > FIFTY_YEARS_SECONDS`, and that one assertion was doing two different jobs: at a startup site
it meant "registered, not yet succeeded", and in the `a failed X run records no success` tests it
was the *only* thing distinguishing a failed run from a successful one. Under the new seed those
tests would have gone **vacuous** — both branches read ~0 — while staying green. So they now split:
startup sites assert the boot-seed bound (the assertion that pins this fix), and failed-run sites
assert `success_recorded == 0`, which is a stronger statement than the age ever made. Test names
that said "leaves liveness old" were renamed to what they now check.

Sentinel tests gained the never-succeeded wording case and the unreadable-flag default.

## Everything now says the same thing

| Artifact | Change |
|---|---|
| `DomainMetrics.kt` | seed at registration; third gauge; KDoc rewritten (it had described the defect accurately, and is now the record of why the seed is load-bearing) |
| `WorkflowLivenessMetrics.kt` | `SUCCESS_RECORDED` + series constant, with why it is a separate series rather than a magic value in the age |
| `prometheus-rules-workflow-liveness.yaml` | comments state that the property they assert was **not** true when the rule shipped, and that the rule must not be run against an EPOCH-like seed |
| `docs/adr/0237-…md` | point 3 amended: the seed is a precondition of the alert, not an implementation detail; both consequences accepted explicitly |
| `docs/agents/control-liveness-sentinel.md` | third gauge added to the metric contract; what it does and does not change |

## Verification

**CI: 100 pass / 0 fail / 0 pending on `f4e5aa913`.** Because this touches
`openbank-libs-*/src/main`, Services CI rebuilt the whole fleet rather than a path-scoped subset —
so all 18 modules whose liveness tests are rewritten here were actually built and run (balance,
interest, lending, ledger, sdd, domestic-payment, aml, analytics-sink, copilot, security-scanner and
the agent services), not just `libs`.

Three reds happened on the way. Two were mine and are fixed in `5eae05c48` and `f4e5aa913`:

| Red | Cause |
|---|---|
| `:openbank-aml-service:ktlintTestSourceSetCheck` | `declaration-with-comment` — the swept companion objects needed a blank line before the `BOOT_SEED_CEILING_SECONDS` comment |
| `:openbank-control-liveness-sentinel:detekt` | `MagicNumber` on the `3` in `if (fields.size < 3)` — the composite-key field count, now a named constant |

Both slipped past the local gate for the same reason, which is worth recording: the first
`./gradlew build` reported `ktlintMainSourceSetCheck` / `ktlintTestSourceSetCheck` / `detekt` as
**UP-TO-DATE** while `compileTestKotlin` re-executed — so the gate ran and checked nothing, and the
overall `BUILD SUCCESSFUL` said nothing about style. Every local check quoted below was therefore
re-run with `--rerun-tasks` and confirmed by its *tasks-executed* count rather than by the build's
exit status:

- `libs-domain` + `libs-runtime` + `control-liveness-sentinel`: `ktlintCheck` 34/34 executed,
  `detekt` 13/13 and 44/44 executed, tests green.

The third red was **not** from this PR: `openbank-standing-order-service` →
`StandingOrderOutboxClaimIT > a stale DISPATCHING row is reclaimed instead of stranded forever`,
failing on the *first* claim assertion of an outbox test in a module with zero files in this diff.
Its job log carries `ConflictException: Status 409: a prune operation is already running` from the
Testcontainers reaper. Reproduced locally against real Docker/Postgres on this branch —
`tests=2 failures=0 errors=0 skipped=0` (counts read from the JUnit XML, since a Quarkus boot
failure would render as SKIPPED and read like a pass) — and it passed on CI attempt 2 with no code
change.

Other checks:

- Gate `check-scheduler-liveness.py`: one `::error` (account-service's
  `SavingsProposalExpiryScheduler`), **byte-identical on a clean `origin/main` worktree** — pre-existing,
  not from this PR. Now tracked as #4241.
- ADR registry: `docs/adr/gen-index.sh` produces no diff (the change is in the body, not the
  front-matter), so `README.md` / `DIGEST.md` / `index.json` are untouched.

## Merging this

Two consequences worth knowing, neither a reason not to merge:

1. **The merge rebuilds the whole fleet**, and auto-deploy fires on `push` before those builds
   publish, so `can-i-deploy` will answer `NOT deployable` for reasons unrelated to any contract
   regression (the #2549 shape). That needs per-service re-dispatch of `auto-deploy.yml` once
   Services CI finishes — not a rollback.
2. **Until a service's pods carry the new libs**, they keep emitting the old two gauges with an
   EPOCH-seeded age. The sentinel reads an absent success flag as "has succeeded", so it will not
   manufacture never-succeeded findings during the rollout — but `WorkflowLivenessStale` keeps
   firing for not-yet-rolled daily workflows until their image ships.

No `version.txt` bump: `openbank-libs-runtime` / `openbank-libs-domain` are not released components
(no `version.txt`, not in `release-please-config.json`). Every service file touched is under
`src/test`, which is in `exclude-paths`, so no service release is triggered either — but the runtime
change does ship inside every service image on its next build.

## Issue

Refs #3345 (ADR-0237 — liveness metric + staleness alert for every scheduler and external feed) and
#2239 (Gap 2, which is where the EPOCH seed was first written down as a blocker).

**Deliberately `Refs`, not `Closes`.** #3345 is the adoption sweep across every scheduler and feed;
this PR fixes the primitive those registrations depend on — without it #3345's own
`WorkflowLivenessStale` rule cannot be trusted, but adoption is not finished by merging this.
`check-scheduler-liveness.py` still reports account-service's `SavingsProposalExpiryScheduler` as
unregistered (pre-existing, see Verification), which is exactly the remaining #3345 work — split out
as **#4241** so it does not ride on this PR's merge.


JiRaska:
Checked the two things that mattered most: whether the fix reintroduces the exact hazard it names, and whether the "has never succeeded" distinction survives it.

**It doesn't reintroduce the hazard.** The seed moves from `Instant.EPOCH` to `Instant.now()` at registration — so a fresh pod reports its own uptime rather than ~1.8e9 seconds, and `WorkflowLivenessStale` stops firing on every deploy/restart across the ~28 registration sites. Confirmed the KDoc's own prior claim was the contradiction it says it was: it asserted "boot-safe by construction" while the code seeded from EPOCH.

**The distinction the PR calls out as destroyed by a naive fix is preserved via a second gauge**, not just asserted: `successRecorded` (0/1) is new, tested with an explicit `NOT_YET_SUCCEEDED` constant across every touched service's liveness test. So "never succeeded" and "succeeded a second ago" stay distinguishable even though both now produce a small age value — which is exactly the gap a same-day PR checking only `age_seconds` would have left open.

Checked the downstream consumer claim too: `control-liveness-sentinel`'s `detectStaleHeartbeats` is purely threshold-based on `age_seconds` vs `interval × multiplier`, so it is unaffected either way — the PR's own framing of this as "detection unchanged, only the boot-time false positive removed" holds up.

Not merging — money-path liveness tests included in scope, human review by rule.

JiRaska:

```
```

JiRaska:
Heads-up: this merge left `main` red, and the failure surfaced only after the fleet rebuild finished.

`build (openbank-account-service)` on merge commit `73d5f0dc4`:

```
SavingsProposalExpiryWorkflowLivenessTest > registers the gauges at startup and
  records success after a sweep() FAILED   (line 47)
SavingsProposalExpiryWorkflowLivenessTest > a swallowed sweep failure leaves the
  heartbeat stale() FAILED                 (line 70)
```

Both assert `ageOf(registry) > FIFTY_YEARS_SECONDS`, which was true while the gauge seeded at `Instant.EPOCH` and is false now that it seeds at `Instant.now()`. The seed change is right; the test encoded the old behaviour as its invariant.

#4325 fixes it, and it does the better thing rather than just widening the tolerance: under the boot seed an age assertion is **vacuous** — it passes whether or not the sweep ran — so it moves the claim onto `SUCCESS_RECORDED`.

No action needed here, flagging for the record. The near-miss worth knowing: the `Services CI` run that completed on `main` in between was for a gitops-only commit with `build: skipped`, so `main` read green for roughly 90 minutes while the rebuild that would fail was still queued.

PR #4209, opened 2026-08-08
refactor(libs): pay off the eight libs-domain Jackson violations #3808 baselined

JiRaska:
Refs #3670 — **not** `Closes`. The three items the issue names are done after this PR (#3808 did the
two CDI interceptor bindings and the JBoss Logger; this one does the `OpaSidecarPolicyDecisionPoint`
move the issue flagged as "also suspicious"), but libs-domain is still not framework-free: six older
baseline entries remain inside the module, listed exactly below. The issue's headline claim is only
true when those go too, so it stays open.

Pays off the **eight** `openbank-libs-domain` Jackson entries that #3808 baselined rather than fixed,
and drops them from `.github/scripts/domain-purity-baseline.txt`. Nothing else in the baseline moves.

## What #3808 left, and why these eight were not "just another move"

#3808 fixed the gate — it rewrote `check-domain-purity.sh` as `check-domain-purity.py`, widened the
scope from the `*/domain/*` **package path** to the whole libs-domain **module**, and added
fully-qualified detection. That widening uncovered eight pre-existing Jackson references in the 75% of
the module the old gate had never opened, and #3808 grandfathered them with a per-entry reason: each
needed a decision about *where the type belongs*, not a rename. That decision is what this PR makes.

## The count, re-derived rather than quoted

| | violations | baseline entries | files scanned |
|---|---|---|---|
| `origin/main` (`c75dd1727`) | 19 | 19 | 291 |
| this branch | **11** | **11** | 287 |

19 − 8 = 11 on both axes. The gate fails on a stale baseline entry as well as on a new violation, so
the two numbers have to move together — see the falsification below.

## Per-violation decision (no bulk move)

ADR-0122's own placement table already assigns `authz`, `flags` and the framework-touching part of
`api` to the **runtime** side. Three different answers, one per shape:

**(a) Move to libs-runtime, package unchanged — the serialization boundary.**
`api/error/ApiError.kt`, `api/pagination/CursorPage.kt`. These are wire DTOs whose
`@JsonInclude(NON_NULL)` *is* the HTTP envelope's shape; deleting the annotation would change every
error and page response on the fleet, so the annotation stays and the type moves to the side of the
split where framework annotations are allowed. `ErrorCode`, `FieldError`, `PageInfo` and
`CursorEncoder` travel with their file.

**(b) Move to libs-runtime, package unchanged — infrastructure adapters, ports stay behind.**
`authz/OpaSidecarPolicyDecisionPoint.kt`, `flags/FlagdProvider.kt`. Both are HTTP clients (an OPA
sidecar, and flagd over OFREP) that happen to have been declared in the domain module. Their **ports**
— `PolicyDecisionPoint` and `FeatureClient` — stay in libs-domain, which is the direction the hexagon
wants: domain declares, runtime implements. `AllowAllPolicyDecisionPoint` and the in-memory doubles
stay in domain too, so the domain module keeps a working implementation of both ports.

**(c) Invert, do not move — `lending/compliance/CompliancePackParser.kt` (3 of the 8 entries).**
This one was already built for the split: `fromMap` is the entire decoder and takes an
**already-parsed** `Map`, while `fromJson` was a three-line Jackson convenience. So the closed-schema
decoder, every fail-closed rejection and ADR-0212 D2's whole contract stay in libs-domain, and only
the front-end leaves, as `CompliancePackJson` in libs-runtime. A YAML front-end would now be a sibling
object there and need no edit to the domain decoder — which is what the baseline's "needs a
port/adapter split" note was asking for.

**(d) No false positives.** All eight are real: every one is an `import com.fasterxml.jackson.*` line
in `src/main` of the module ADR-0122 declares framework-free.

## Consumers

Package names are unchanged, so **no consumer import moved** for (a) and (b) — the same mechanism
#3808 used for `@Authorize`/`@FeatureFlag`. Counted before moving: 39 modules reference
`OpaSidecarPolicyDecisionPoint`, 16 `ApiError`, 13 `ErrorCode`, 7 `PageInfo`, 6 `CursorPage`,
5 `CursorEncoder`, 2 `FlagdProvider`. Every module that depends on `openbank-libs-domain` also depends
on `openbank-libs-runtime` — with exactly two exceptions, `openbank-libs-runtime` itself and
`fuzz/ossfuzz`, and ossfuzz references none of these types (checked by name, not assumed).
libs-runtime re-exports libs-domain as `api()`, so nothing lost visibility of anything.

Only (c) changes source at call sites, because that split deliberately renames the JSON entry point:
**4 call sites in 3 files**, all in `openbank-lending-service`
(`CompliancePackActivationService` ×2, `CompliancePackConfig`, `CzReferencePackTest`).

**No `@Produces` was added or moved**, so the documented "a producer in libs-runtime breaks modules
that do NOT consume it" trap is not in play — the moved types are plain classes, objects and data
classes, constructed by their consumers exactly as before.

## What the fleet compile caught

`OpaSidecarPolicyDecisionPoint` did not compile after the move:

```
Smart cast to 'ResourceRef' is impossible, because 'resource' is a public API property
declared in different module.
```

`AuthzQuery.resource` is a public property of libs-**domain**, and Kotlin will not smart-cast across a
module boundary. A same-module compile cannot see this, and no test can either — it is exactly the
class of breakage a package-preserving move introduces silently. Rewritten as `resource?.let { }`,
with the reason in a comment so it is not "simplified" back.

## Verification

- **Fleet compile**: `compileKotlin` + `compileTestKotlin` for all 69 Gradle modules, in six batches
  (the single whole-fleet invocation dies of `Java heap space`). All batches green.
- `:openbank-libs-domain:test`, `:openbank-libs-runtime:test`, and `koverVerify` on both — green, so
  moving three test classes across the boundary did not drop either module under its coverage floor.
- `ktlintCheck` + `detekt` on libs-domain, libs-runtime and lending-service — green.
- `:openbank-lending-service:quarkusBuild` — green. CDI/ArC wiring is not validated by unit tests, and
  lending-service is the one consumer whose source actually changed.

## Falsification — both directions, on the real tree, exit codes read directly

```
clean branch                     -> 11 violations, 0 new, 11 baseline entries        rc=0
+ import ObjectMapper back into
  CompliancePackParser.kt        -> "NEW framework reference ... CompliancePackParser.kt:
                                     import com.fasterxml.jackson.databind.ObjectMapper"
                                    12 violations, 1 NEW, 11 baseline                rc=1
  (reverted)                                                                         rc=0
+ leave the paid-off ApiError
  entry in the baseline          -> "baseline entry no longer occurs — delete it from
                                     domain-purity-baseline.txt (ratchet-only)"
                                    11 violations, 0 new, 12 baseline                rc=1
  (reverted)                                                                         rc=0
```

Through the real runner, not the script by hand:

```
$ python3 .github/scripts/run-gates.py --only domain-purity-gate
--- self-test (must PASS) ---
check-domain-purity --self-test: 20/20 cases correct.
--- gate ---
check-domain-purity: scanned 287 domain-layer file(s); 11 violation(s) found, 0 new, 11 baseline entr(y/ies).
1 gates  PASS=1
```

The scanned-file count drops 291 → 287 (the four moved main sources), well clear of the gate's
`MIN_FILES_SCANNED = 250` scope floor.

## The exact remainder — why this is `Refs`, not `Closes`

All 11 surviving entries are older debt from *before* #3670 widened the scope, and none is one of the
three the issue reported. Six are still **inside libs-domain**, so the issue's headline claim does not
hold yet:

```
openbank-libs-domain/.../domain/identifiers/EntityId.kt   : @JsonCreator, @JsonValue
openbank-libs-domain/.../domain/identifiers/LendingIds.kt : @JsonCreator, @JsonValue
openbank-libs-domain/.../domain/money/Money.kt            : @JsonCreator, @JsonProperty
```

These are value types, not adapters — (b) does not apply, and neither does (c), because the annotation
is *on* the type a hundred DTOs embed. They want a per-value-type serialization decision (a Jackson
module registered in libs-runtime, most likely) and their own PR. The other five are outside
libs-domain entirely: three `JsonNode`/`@JsonInclude` references in `openbank-agent-service`'s own
`domain/` package, one in `openbank-copilot-service`'s, and one `@RegisterForReflection` in
`openbank-product-catalog`. The baseline header now says all of this instead of pointing the whole
file at #3670.

## Housekeeping

- 22 detekt baseline entries move from `openbank-libs-domain/detekt-baseline.xml` to
  `openbank-libs-runtime/detekt-baseline.xml` (the `MagicNumber` HTTP-status entries travel with their
  files). Neither baseline gains an entry.
- The three moved tests (`OpaSidecarPolicyDecisionPointTest`, `FlagdProviderTest`, `CursorEncoderTest`)
  move with their code. `CompliancePackParserTest` and `CompliancePackRegistryTest` stay in libs-domain
  with a **test-only** JSON fixture, so the 180-line decoder suite keeps living next to the decoder.
  The gate scans `src/main/kotlin` only, so nothing there is in scope.
- `openbank-libs-domain/build.gradle.kts` records what moved, and that Jackson now survives there only
  as the annotation dependency of the still-baselined value types.

## Parallel-work note

`origin/main` was merged in at `98f2b7b06` and re-verified afterwards: the gate still reports
287 files / 11 violations / 11 baseline entries, and `git diff --name-only origin/main` is exactly the
17 files above — no entry silently dropped out of a list by a clean merge.

One open PR overlaps: **#3880** (`fix(libs): stamp ApiError.timestamp at construction instead of
serving 1970`) edits `openbank-libs-domain/.../api/error/ApiError.kt`, the file this PR *moves*. The
two changes are compatible in content — #3880 changes a default value, this PR changes the file's
module — but whichever lands second needs the other's edit carried onto the new path. Merging #3880
first is the cheaper order. No other open PR touches these files (#3644 edits
`CompliancePackRegistry.kt`, which this PR does not).


JiRaska:
Reviewed. **Clean — blast radius checked and empty.**

**No wire-contract change.** `ApiError.kt` and `CursorPage.kt` move domain→runtime at `similarity index 100%` with the **package unchanged** (`com.openbank.libs.api.error`, `…api.pagination`). No field, order, nullability or `@JsonInclude` edit anywhere in the diff, so nothing a pact or an admin-ui TS type could pin actually moves. That is the check worth doing on a libs refactor — a serialization shape can change without a single line of behaviour changing.

The one semantic edit in a moved file is `OpaSidecarPolicyDecisionPoint` — `if (resource != null) put(…)` → `resource?.let { put(…) }`, a forced cross-module smart-cast workaround that produces byte-identical OPA input.

**Compile-time reach is empty too:** every consumer of the moved packages already has `openbank-libs-runtime`, and runtime does `api(project(":openbank-libs-domain"))`, so the reverse direction is unaffected. No module depends on libs-domain without libs-runtime.

**Framework-free is preserved** and the direction is strictly outward. The one new libs-domain file is a **test** fixture (`CompliancePackJsonFixture.kt`) using Jackson under `src/test`, which the purity check does not scan; `CompliancePackParser` loses all three Jackson imports and keeps `fromMap`. The deliberate 3-line duplication between that fixture and the production `CompliancePackJson` in runtime is justified in place, and it is the right call — the alternative inverts the ADR-0122 dependency direction to save three lines.

**No `@Produces` added or moved**, which is the specific hazard worth checking here: a libs-runtime producer can break services that never consume it, because Arc resolves the whole type closure. `OpaSidecarPolicyDecisionPoint` and `FlagdProvider` are plain classes with no CDI annotations, instantiated by each service's own `AuthzProducer`. Detekt baseline entries move with the files in both directions.

**Minor, cosmetic:** two KDoc links in `openbank-libs-domain/.../api/search/SearchRequest.kt:12,39` still point at `com.openbank.libs.api.pagination.CursorPage` / `CursorEncoder`, which now live in another module — dangling references, harmless unless Dokka is strict.

`refactor(libs)` is the right type: it does not release, which is correct for a change where nothing in the wire shape moves.

PR #4211, opened 2026-08-08
fix(customer-edge): answer 400, not 500, for a missing required query parameter

JiRaska:
## What

Thirteen non-money-path handlers across **customer-edge, statement and dispute** answered **500** when the caller omitted a required query parameter. They now answer **400 naming the parameter**.

Refs #3624 (the tail of #3104).

## Why these three

All three are plain-`fun` handlers, which is the half of the tail where the issue's "answers 500" framing is exactly right — and it was measured, not assumed (see the evidence table below). The four services left in the baseline (aml, party, pid, tpp-registry) are all `suspend`, a materially different fix: no `Intrinsics.checkNotNullParameter` is emitted, each already carries a hand-written guard that has to be *widened* to `isNullOrBlank()` rather than replaced, and each preserves its own error envelope. Mixing the two halves in one PR buys nothing and makes the review worse.

**customer-edge is the most caller-visible instance of this defect in the whole tail.** It is the mobile/web BFF: the app cannot tell "I built a bad request" from "the bank is broken", so a 5xx is what it retries, what pages, and what burns the error budget.

## The mechanism (why the obvious fix is dead code)

JAX-RS injects `null` for an absent parameter. Kotlin's null-safety is compile-time only, so on a non-`suspend` handler the declared non-nullable type compiles to an `Intrinsics.checkNotNullParameter` at bytecode offset 0 — the NPE lands *before the first statement of the body*, and the generic mapper answers 500. Writing `requireNotNull(x)` against the old signature would have compiled to nothing. **The parameter must be declared nullable for any guard to be reachable at all.** Same fix shape as the merged fleet precedent `09f32b673` (#3625).

## Error envelope: two shapes, on purpose

- **statement, dispute** — `requireNotNull`, mapped by libs-runtime's `IllegalArgumentExceptionMapper` to 400 + `ApiError`. Neither service has a 400 envelope of its own.
- **customer-edge** — its own `badRequest()`. `listSddMandates` already answers an absent `accountId` exactly this way, and introducing a second 400 shape into a resource that answers every other 400 with an explicit `Response` is the #526 two-mappers-for-one-type lottery in slow motion.

No service-local `ExceptionMapper` for a JDK type was added anywhere.

## No `@DefaultValue` anywhere here

Each parameter was read against its handler body:

- **statement `from`/`to`** — "last month" is a tempting default and the wrong one. A period close assigns **legal sequences** and is idempotent on `(account, currency, period)`, so a guessed range does not fail loudly, it writes the *wrong* statement. `format` on the same handlers genuinely is defaultable and keeps its existing `@DefaultValue`, untouched.
- **dispute `actor`** — written to the timeline as attribution for who withdrew or escalated. Defaulting it files an unattributable state change.
- **customer-edge `accountId`/`currency`** — `accountId` is the IDOR guard's input on every one of these reads. There is no default that is not "some other customer's account".

Each guard also runs **before** the first `UpstreamClient` call, so a rejected request no longer costs an upstream round trip.

## Red-then-green evidence

Every assertion was run against the **unfixed** source first. A test that passes both ways is worthless, and this class of defect is only observable over real HTTP — a unit test calling the handler supplies the very argument the framework does not, so it passes against the broken code. `CustomerEdgeResourceTest` constructs the resource directly and is structurally incapable of seeing this.

| IT | assertions | before the fix | after |
|---|---:|---|---|
| `CustomerEdgeMissingParamStatusIT` | 5 | **500** ×5 | 400 ×5 |
| `StatementMissingParamStatusIT` | 3 | **500** ×3 | 400 ×3 |
| `DisputeMissingParamStatusIT` | 2 | **500** ×2 | 400 ×2 |

Each IT also carries a **control** that passes both before and after, so the redness is attributable to the guard and not to the harness. All three controls passed on the red runs.

Two things the harness needed, both found by a test that was green about nothing until they were added:

- **`authz.enforce=false` via a `@TestProfile`** (statement, dispute). The default is `true` and there is no OPA sidecar under test, so the PDP fails closed and every `@Authorize` endpoint answers **503** before the handler is entered. Literal values only — a profile loads in a different classloader from the test class. customer-edge needs no profile: its `MockAuthzProducer` already covers this.
- **`contentType` on the POSTs.** The resources are `@Consumes(APPLICATION_JSON)`, so a POST without it is rejected with **415** before parameter binding is reached. The first run of `DisputeMissingParamStatusIT` measured 415, not the defect.

## The gate baseline

`check-nonnull-jaxrs-params.py`'s `BASELINE` goes **12 keys to 7** — the five belonging to these three services are deleted, not lowered. The gate fails on a stale baseline entry in either direction, so leaving them would have gone red. Verified on this branch: `--self-test` 14/14, and `--root .` reports 7 baselined occurrences and no new ones.

The remaining seven keys now carry a note recording *why* they are a separate change (all `suspend`), so the next person does not rediscover it.

## OpenAPI

No spec change, so no `info.version` bump. All eight of these parameters that appear in a spec already declare `required: true` — the specs were already right and the code was not. Verified against the three `openapi.yaml` files as of `origin/main` `123633fcd`.

## Note for the reader: #3660 never reached `main`

#3624's own comments describe this tail as fixed and awaiting only a merge. It is not. #3660 merged into #3658's branch on 2026-08-06, #3658 squash-merged to `main` on 2026-08-07 — and the #3658 squash contains only psd2 and card-issuance. `DisputeResource.kt` has not been touched since #3057. Running the gate over a clean `origin/main` reports **20 findings across 7 services**, all still baselined. That is what this PR is against.


JiRaska:
Reviewed. **Right layer, right tests — mergeable.** One contract gap worth raising.

**The fix is at the only layer where a guard is reachable**, and that is the thing this class of defect usually gets wrong: every touched parameter is made **nullable at the signature** (`currency: String?`, `accountIdOrNull: UUID?`, `actor: String?`, `from`/`to: String?`). With a non-nullable Kotlin type the JVM throws `Intrinsics.checkNotNullParameter` at the method boundary, so a `requireNotNull` in the body is dead code — the declaration decides whether any guard can run at all.

dispute and statement route through `requireNotNull` → libs-runtime's `IllegalArgumentExceptionMapper` (400 + `ApiError`) with **no service-local mapper added** — I checked the diff, there is none, which matters because a second mapper for one JDK type is picked non-deterministically per request. In dispute the throw is synchronous (the argument is evaluated before the `Uni` is built), so it escapes the handler rather than the reactive pipeline and does not collide with `resolve`'s `onFailure(IllegalArgumentException) → 422`.

customer-edge deliberately uses its own `badRequest(...)` to match `listSddMandates` and the rest of that resource. That is the right instinct rather than a deviation — two 400 shapes in one resource is how the mapper lottery starts — and it introduces no ExceptionMapper.

**The tests drive real HTTP** (`@QuarkusTest` + RestAssured, not resource-class calls), and each carries a **control** that passes before and after, so the redness is attributable to the guard rather than to the harness. Two traps are pre-empted in-file, which is the sign someone actually ran them: `@Consumes(APPLICATION_JSON)` means a POST without `contentType` is 415 *before* binding, and `authz.enforce` defaulting true makes the PDP fail closed with 503 *before* the handler — fixed with an `AuthzOffProfile` using literal values only, avoiding the profile-classloader footgun where a companion object initialises twice.

Gate hygiene is right too: the six baselined entries are deleted from `check-nonnull-jaxrs-params.py` in the same PR as the fix, and the remaining tail is re-scoped with an accurate reason (`suspend` handlers emit no `checkNotNullParameter`, so the null flows into the body and needs a widened guard, not this change).

**The gap:** none of the three `openapi.yaml` files is touched, and the affected operations document no `400` — e.g. `openbank-dispute-service/src/main/resources/openapi.yaml:134-149` (`withdrawDispute`) lists only `'200'` while `actor` is already `required: true`. The set of status codes an operation can return is part of the contract (ADR-0048), so a documented `400` and a patch bump would make the spec match the service. Arguable that the previous 500 was undocumented too and this is a pure bug fix — but it is the one thing a reviewer can fairly ask for.

JiRaska:
**Unblocked from `DIRTY`. One conflict, in `openbank-dispute-service/build.gradle.kts`, and it is the kind that silently loses a feature.**

Both sides added a *different* `testImplementation` to the same dependency block:

```kotlin
testImplementation(project(":openbank-libs-testing"))   // main — AuditEventTime, #3914
testImplementation(libs.rest.assured.kotlin)            // this branch — DisputeMissingParamStatusIT
testImplementation(libs.quarkus.test.security)          // this branch
```

Taking either side would have compiled and passed review while deleting the other's whole test capability — either the `#3914` producer tests lose the shared `AuditEventTime` rule, or `DisputeMissingParamStatusIT` loses the ability to drive real HTTP, which is the only way this defect is observable at all. Kept all three.

Verified after resolving: `:openbank-dispute-service:test` BUILD SUCCESSFUL, verdict read from the JUnit XML.

For the record on the other check that was red here — `Customer-app dossier` and friends have been fleet-wide inherited faults all session; this diff touches none of those files.

JiRaska:
**`dependency-review` is red for a reason that is not this diff, and the guard is right to say so.**

```
No successful dependency-submission run exists for the head 8c6b4a7d3.
Check the Dependency submission workflow for this PR — if it failed, the Gradle
half of this review is vacuous. See issue #2833.
```

Traced it: the `dependency-submission` run for this head (`31278544497`, 21:07Z) was **cancelled** — superseded by my own push while it was in flight. The two runs before it, on earlier heads of this branch, both succeeded. So nothing is broken; the snapshot for the current head simply never finished, and without it the Gradle half of the review would have passed while checking nothing.

That guard is the point: `dependency-review` only diffs *submitted* graphs and GitHub does not parse Gradle natively, so a missing snapshot makes the review green and empty. Failing loudly is the correct behaviour.

Re-ran the cancelled submission. Once it completes, `dependency-review` needs a re-run to pick it up.

Worth knowing for anyone pushing to a branch that touches a `build.gradle.kts`: each push cancels the in-flight submission and restarts a ~24-minute submit-and-wait, so a rapid second push reliably produces this exact red.

JiRaska:
Reviewed, and specifically measured the interaction with two other open PRs that touch the same artifacts — neither PR's own CI can see the combination.

## This is the real remediation, and it also unblocks #4194 by a second route

#4211 removes five BASELINE entries by actually fixing the parameters, including `openbank-customer-edge|CustomerEdgeResource|accountId`.

That matters beyond its own scope. #4194 is currently red on this gate through a **keying** bug — a nested `data class Bad` becomes the enclosing declaration for every handler below it, re-keying `listDisputes` from `CustomerEdgeResource|accountId` (baselined) to `Bad|accountId` (reported NEW), on a PR that touches no JAX-RS parameter. #4249 fixes that bug. But this PR closes the same red from the other end: with `listDisputes` declared nullable there is no finding at all, so there is nothing left to mis-key.

Measured, not reasoned:

```
pr4211 + pr4194            MERGE CLEAN — and the gate passes with the OLD regex
                           (7 baselined, 0 new, 0 stale)
pr4211 + pr4249            MERGE CLEAN — self-test 16/16, tree 7 baselined, 0 new
```

The two touch different hunks (`BASELINE` here, `DECL_RE` + self-test there), so they compose. Worth having **both**: this one is the actual fix, #4249 stops the mis-keying recurring the next time anyone declares a nested type in a resource — which is a normal thing to do.

Also note #4211 and #4194 both edit `CustomerEdgeResource.kt` substantially, and that merges clean too. I checked because a clean merge and a correct merge are different claims here, and the gate agreeing afterwards is the evidence, not the absence of a conflict marker.

## The baseline comment is the good part

Splitting the remaining entries by *why* they are still there — every survivor is a `suspend` handler, where no `Intrinsics.checkNotNullParameter` is emitted, so the null flows into the body and each already carries a hand-written guard that must be **widened** to `isNullOrBlank()` rather than replaced, preserving the service's own error envelope — is exactly the distinction that makes the remaining tail actionable instead of an undifferentiated list. That is a materially different fix from the plain-`fun` half, and saying so in place beats a tracking issue nobody reads.

## Status

`MERGEABLE/UNSTABLE` with all five required contexts green — that is the `gh` behaviour where a non-required red check blocks the CLI but not the ruleset. Merging is not mine to do regardless: this changes CI gate machinery (`check-nonnull-jaxrs-params.py`) and touches customer-edge, so it needs a human.

Issue #4187, opened 2026-08-08
Agent swarm: Phase 4 — GitOps deployment of case-coordinator runtime


JiRaska:
parent: 3737

## Phase 4 — GitOps deployment of the case-coordinator runtime

Deployment of Phase 1+2 (#3737). The `case-coordinator-agent` module must become a deployed workload with its Temporal worker, OPA bundle, and Keycloak wiring.

### Scope

- GitOps manifests for `case-coordinator-agent` (deployment/rollout per the module's neighbors, CNPG database `openbank_casecoordinator`, service, ingress if the read API is reachable by admin-ui, network policies via `gen-network-policies.py`)
- Temporal worker wiring verified as-deployed: `openbank.temporal.enabled=true` + server/namespace/task-queue `case-coordinator` env in the workload manifest (repo default is false)
- OPA bundle for case-coordinator (Phase 1's in-process `CaseCapabilityGate` becomes the OPA adapter; `case.open`/`case.coordinate`/`case.synthesize`/`case.preempt` decisions via `case.capabilities` input per ADR-0244 D2) + bundle regen
- Keycloak: M2M client/scopes for agents that join cases (service-account convention; no ROLE_SERVICE — identify by principal id per repo rules)
- Kill-switch → Temporal case-cancellation wiring verified AS-BUILT (ADR-0244 D7 follow-up: kill-switch signal terminates the running CaseWorkflow; test or documented verification)
- `QUARKUS_FLYWAY_MIGRATE_AT_START` / datasource env per repo Flyway rules (`check-flyway-default-datasource.py` enforces both directions)

### Explicitly out of scope

- Shadow pilot + evals (Phase 5), money-path case classes (threat model per ADR-0030 first)

Refs: ADR-0244 (D2/D7), ADR-0202 (agent-plane deployment), openbank-infra/CLAUDE.md (GitOps pitfalls)

Issue #4186, opened 2026-08-08
Agent swarm: Phase 3 — admin-ui swarm thread view (ADR-0246)


JiRaska:
parent: 3737

## Phase 3 — Admin-ui swarm thread view

The visualization layer of the swarm program (#3737). Depends on Phase 2 API — renders the ADR-0246 thread view against the real case history API. READ ADR-0246 first (docs/adr/0246-*.md): layout, semantics, honest empty states (D6), responsive+WCAG (D7), no mock data (ADR-0231).

### Scope

- `/iaops` integration per ADR-0246: a "Cases" entry point from the agent roster (case-coordinator drill-down gains a cases section; fleet-wide case strip optional per ADR)
- Thread view: chronological case thread — case-open banner, agent joins, contribution cards (agent avatar/identity per ADR-0156 roster), pre-emption markers (`superseded-by-evidence` shown as a deterministic fork, not an error), contested banner, final synthesis card linking the HITL proposal
- Status semantics per ADR-0244: OPEN / CONVERGING / CONTESTED / SYNTHESIZED / CLOSED with distinct, accessible visual treatment (color + icon + label, never color alone)
- Honest empty states (ADR-0246 D6): no cases → explicit "no cases yet" copy explaining what a case is; API down → error state, never fabricated threads
- Read-only: the UI never mutates a case (no signal buttons in this phase)
- Vitest coverage: thread rendering from the Pact-shaped fixture, empty/error states, a11y labels; run via `npm test` (pretest bakes generated artifacts — never bare vitest)
- shadcn/ui + the roster's existing design tokens; dark-mode safe

### Explicitly out of scope

- Signal submission UI, case-open forms (agents open cases, not humans — Phase 5 evaluates whether humans ever steer)
- Live updates/polling beyond a modest refresh interval (no websockets this phase)

Refs: ADR-0246, ADR-0244, ADR-0231 (no mock data), ADR-0156 (/iaops roster + drill-down conventions)

JiRaska:
## Closing — Phase 3 is on `origin/main` and deployed

Verified against `origin/main` (not the PR): #4235 merged and the artifacts are present.

**Surface**
- `openbank-admin-ui/src/app/iaops/cases/page.tsx` (case list) and `.../cases/[caseId]/page.tsx` (thread view)
- Entry point from the agent roster drill-down: `app/iaops/agents/[agentId]/page.tsx:233` links to `/iaops/cases`
[...]

Issue #4185, opened 2026-08-08
Agent swarm: Phase 2 — case history read model + API (ADR-0246 surface)


JiRaska:
parent: 3737

## Phase 2 — Case history read model + API

Read path for the swarm program (#3737). Depends on Phase 1 (#4181) merged — the API projects Temporal case history (`CaseWorkflow.state()` query + `case_workflow`/`case_contribution` tables) into the thread-view model defined by ADR-0246.

### Scope

- Query endpoint(s) on `case-coordinator-agent` projecting Temporal case history into the ADR-0246 thread-view model: case list (open/recent, class, status, participants, contested flag) + case detail (thread: joins, contributions, pre-emptions, synthesis, proposal link)
- OPA/authz on the read path consistent with the fleet (roles from the existing placeholder resource; @RolesAllowed, no new OPA rules in this phase)
- `openapi.yaml` updated + API-contract version bump per ADR-0048 (additive MINOR)
- Pact: consumer test (admin-ui as consumer) + `@PactFolder` provider replay in case-coordinator-agent — per repo rules the provider replay must run on PR, broker-only does not count
- Honest empty states: no cases → empty list, never synthetic data (ADR-0231)

### Explicitly out of scope

- Mutations of any kind (read-only API; UI never mutates a case)
- The admin-ui thread view itself (Phase 3), GitOps (Phase 4)

Refs: ADR-0246 (API surface + thread model), ADR-0244 (case semantics), ADR-0048 (API versioning)

PR #4215, opened 2026-08-08
feat(case-coordinator): add Temporal CaseWorkflow swarm lifecycle

### LINKED ISSUE #4181: Agent swarm: Phase 1 — Temporal CaseWorkflow skeleton in case-coordinator-agent (ADR-0244)
parent: 3737

## Phase 1 — Temporal `CaseWorkflow` skeleton in `openbank-case-coordinator-agent`

> Module correction 2026-08-08: the chartered AGPL module `openbank-case-coordinator-agent` landed meanwhile (LLM adapter, V1 case schema, temporal-sdk + libs-temporal deps, task queue `case-coordinator`) — the workflow belongs there per ADR-0244 D3, not in agent-service.

First runtime increment of the swarm program (#3737, ADR-0244). Governance substrate from Phase 0 (schema + charter + `agent-case-schema` gate) is merged.

### Scope

- Temporal `CaseWorkflow` (agent-plane, AGPL module per ADR-0244 D3) with OPA-gated signals per ADR-0244 D2 — every case mutation goes through `PolicyDecisionPoint.authorizeCapability(...)` with case context
- Deterministic pre-emption via `superseded-by-evidence` (ADR-0244 D5): newer evidence signal supersedes an in-flight draft, workflow converges to one synthesis
- Case-open authority: only `case-coordinator` opens cases; dedup window + per-agent hourly rate limit + fleet contested-rate circuit breaker from `agents.yaml: case_classes` (D9); contested cases escalate to HITL instead of auto-action
- Pilot case class: `incident-response` (non-money-path; money-path classes deferred to the threat model per ADR-0030)
- Output path: exactly one HITL proposal via the existing `proposal-events` outbox (D7) — no direct case output to operators
- Tests: workflow unit tests (signal ordering, pre-emption convergence, circuit breaker) + `@QuarkusTest` coverage where CDI wiring matters

### Explicitly out of scope

- Case history read API (Phase 2), admin-ui thread view (Phase 3), GitOps deployment (Phase 4), shadow pilot + evals (Phase 5)
- Additional case classes beyond the pilot

Refs: ADR-0244 (D2/D3/D5/D7/D9), ADR-0202 (Temporal agent-plane), ADR-0031 (agents propose, governance disposes)

JiRaska:
## What

Phase 1 of the swarm-coordination delivery plan (#3737, ADR-0244): the Temporal `CaseWorkflow` lifecycle in `openbank-case-coordinator-agent`.

- **`CaseWorkflow`** (workflow type `CaseWorkflow`, task queue `case-coordinator`) — deterministic state machine: `contribute` / `contest` / `request-synthesis` signals, per-case-class contribution cap + deadline (TTL), draft-version pre-emption, contested-rate breaker (threshold from `agents.yaml: case_classes`), and a **single terminal synthesis exit** — the "exactly one HITL proposal per case" invariant (D7) holds even under a signal storm (tested).
- **`CaseOpenService`** — dedup by deterministic workflow id `case-<class>-<subject>` (`REJECT_DUPLICATE`), per-agent hourly rate limit, fail-closed concurrent-open ceiling via Temporal visibility count.
- **`CaseCapabilityGate`** — `agents.yaml`-based `case.open`/`case.contribute`/`case.synthesize` check, fail-closed. OPA-backed bundle wiring is Phase 4 (#4187).
- **REST** — `POST /api/v1/case-coordinator/cases` (201/202/400/403/409/429/503) and `POST /cases/{id}/signals`.
- **Outbox** — terminal proposal event published transactionally (`case_outbox`, V2 migration) to `proposal-events`; consumed by the existing ADR-0031 HITL queue.
- **Config** — split into `@ConfigGroup` (`openbank.case-coordinator.case.*`); kafka channel + `outbox.dispatch-enabled: true` in `application.yaml`.

Closes #4181 · Refs #3737

## Verification

- 11/11 unit tests green: `CaseWorkflowTest` (synthesis→1 proposal, pre-emption converges on final draft, contested breaker blocks auto-synthesis, TTL → case-timeout proposal, signal storm → exactly 1 proposal) + `CaseOpenServiceTest` (dedup/rate-limit/ceiling/gate/Temporal-down paths).
- `ktlintCheck`, `detekt`, `quarkusBuild` (CDI) green; kover floor ratcheted 5% → 42% (measured; `@Path` classes excluded per module config).
- The module's boot smoke IT (REST boot + JDBC activity paths) runs in CI — local Docker Desktop VM disk is full, cannot run Testcontainers locally.

## Explicitly out of scope (next phases)

- Read model + case history API → #4185 (Phase 2); openapi.yaml + Pact land there with the query surface.
- Admin-ui swarm thread view → #4186 (Phase 3, ADR-0246).
- GitOps deployment, OPA bundle for the capability gate, kill-switch→Temporal cancellation wiring → #4187 (Phase 4).

### DIFF
--- a/docs/agents/case-coordinator.md
+++ b/docs/agents/case-coordinator.md
@@ -33,17 +33,40 @@ disposes before any state changes.
 - Case budgets, deadlines, and contested-rate thresholds are declared in `agents.yaml` under
 `case_classes`, so they are reviewable and versioned with the same governance rigor as tool tiers.
 
+## Phase 1 runtime (as-built, #4181)
+
+The `openbank-case-coordinator-agent` module now runs the case lifecycle:
+
+- **`CaseWorkflow`** (Temporal workflow type `CaseWorkflow`, task queue `case-coordinator`) — a
+  deterministic state machine that accepts `contribute` / `contest` / `request-synthesis` signals,
+  enforces the per-case-class contribution cap and deadline (TTL), applies pre-emption by draft
+  version, trips the contested-rate breaker, and exits **only** through a single terminal synthesis
+  activity — so the "exactly one proposal per case" invariant holds even under a signal storm.
+- **Capability gate** — `CaseCapabilityGate` checks `case.open`/`case.contribute`/`case.synthesize`
+  against `agents.yaml` (fail-closed). This is the Phase 1 placeholder for the OPA-gated check;
+  wiring the gate to the OPA sidecar bundle is Phase 4 work.
+- **REST** — `POST /api/v1/case-coordinator/cases` opens a case (201/202/400/403/409/429/503
+  depending on dedup, rate-limit, concurrent-ceiling and Temporal-availability outcomes) and
+  `POST /api/v1/case-coordinator/cases/{id}/signals` delivers signals to a running workflow.
+  Case opens are deduplicated by deterministic workflow id (`case-<class>-<subject>`,
+  REJECT_DUPLICATE) and rate-limited per opening agent.
+- **Outbox** — the terminal proposal event is published transactionally via the shared outbox
+  pattern (`case_outbox` table, V2 migration) to the `proposal-events` Kafka topic; the proposal
+  itself is consumed by the existing ADR-0031 HITL queue, which owns human disposition.
+- **LLM synthesis** — the convergence judgement is an LLM call through the prompt registry
+  (`CaseCoordinatorLlmPort`/`Adapter`), never a silent deterministic shortcut.
+
 ## Known gaps (these are real, not aspirational)
 
-- **There is no `openbank-case-coordinator-agent` module yet.** This charter is the governance
-scaffold; the Quarkus service, Temporal workflow, signal handler, and OPA-gated capability check
-are follow-up work tracked by ADR-0244 delivery notes.
 - **No other charter currently grants `case.join` or `case.contribute`.** The coordinator is the only
-swarm participant defined today; other agents will gain swarm capabilities only after the signal
-contract and policy gate are implemented.
+  swarm participant defined today; other agents will gain swarm capabilities only after the signal
+  contract and policy gate are implemented.
 - **The kill-switch-to-Temporal-cancellation wiring is not built.** The global and per-agent kill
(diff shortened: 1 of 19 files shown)


JiRaska:
The red `gates (data)` is legitimate and its message already says what to do, so I will not restate it. Two things around it that the gate does **not** say.

### 1. The default topic name breaks the fleet convention

```yaml
topic: ${CASE_OUTBOX_TOPIC:proposal-events}
```

Every topic referenced by a service's `application.yaml` on `main` is `openbank.<domain>.<event>` — 110 unique `openbank.*` topic strings across gitops, and **zero** exceptions. `proposal-events` would be the first. It is also referenced nowhere else in the tree, so there is no collision today — but the naming is what the ACL coverage and topic-existence gates key on, and an unqualified name in the `${VAR:default}` position is the one that ships when the env var is unset.

Worth renaming to `openbank.case.proposal-events` (or similar) *before* the contract is written, since the contract will pin whatever name is chosen.

### 2. The topic-existence gate is structurally blind here — correctly, for now

`check-kafka-topics-exist` passed (*48 topics declared in gitops, 38 referenced by services, 0 unresolved*), and it did not flag this producer, because the service has **no gitops presence at all** on the branch — no namespace, no KafkaUser, no workload. The gate skips services without a KafkaUser and says so for `psd2` and `tax-reporting`; this one is not even enumerated.

That is fine for Phase 1, which is deliberately code-only, and I am not asking for gitops here. It is worth knowing for the phase that adds it: the moment case-coordinator gets a KafkaUser, that gate starts checking — and a topic declared nowhere will fail there rather than here. Better to pick the final name now than to discover it through a second gate later.

### On the contract itself

The gate's own message is unusually good and worth heeding literally — *"Adding your topic to the baseline is not the fix — the baseline only shrinks."* That is the property that keeps this ratchet meaningful, and it is the reason a new service is the right place to pay the contract cost rather than inherit an exemption.

Nothing else in the PR looks blocked on me: the workflow-level invariants (single terminal synthesis exit under a signal storm, deterministic dedup id, fail-closed concurrent-open ceiling) are the kind of properties a Temporal test can actually prove, and the body claims they are tested rather than asserted.

Issue #3921, opened 2026-08-06
FX revaluation: rate staleness is unobservable and a late fixing can never be incorporated (#1302 item 4)


JiRaska:
Measured on `origin/main` @ `b72f57ea` while working the tail of #1302. Item 4 of that issue ("FX revaluation: no rate staleness, no catch-up") is still fully open; this issue carries the measurement so the work is actionable on its own, since it lives in a different service pair (ledger + fx) from #1302 item 5 and its fix is not additive.

## The seam is the problem, not a missing check

`FxRevaluationService`'s port carries no time at all:

- `openbank-ledger-service/.../application/port/out/CnbRateProvider.kt:17` — `suspend fun cnbRate(base: String): BigDecimal?`
- `openbank-ledger-service/.../infrastructure/client/FxServiceClient.kt:42-47` — `FxRateResponse(baseCurrency, quoteCurrency, bidRate, askRate)`

fx-service *does* persist and return `validFrom`/`validTo` (`openbank-fx-service/src/main/resources/db/migration/V1__create_fx.sql`), and the ČNB parser *does* read the fixing's own date (`domain/cnb/CnbFixingParser.kt:50-59` → `CnbFixing.date`). Ledger throws that away on the wire. So there is nowhere in `FxRevaluationService` that a staleness check could be written today — the fix starts at the DTO, not at an `if`.

## What that costs

1. **Stale-but-valid rates are used silently.** The only bound is server-side and date-blind: `CnbRateIngestionService.kt:94` sets `CNB_VALIDITY_DAYS = 3`, and `FxRepositories.kt:72-81` selects the latest row with `validTo > :now`. Within that window a Friday fixing marks a Monday revaluation with no warning, no metric and no alert. `command.date` is never compared to the rate's `validFrom`.
2. **A late fixing can never be incorporated.** The key is `idempotencyKey = "fx-reval-${command.date}"` (`FxRevaluationService.kt:97`) — date only, no rate identity. `LedgerService.kt:69` returns the existing journal entry on a key hit, so a re-run after the correct fixing lands returns the **original** entry, emits no new event, and reports `posted = true`. A success that changed nothing. There is no correcting entry and no versioned key.
3. **Backfill uses the wrong rate.** `cnbRate(base)` ignores `command.date`, so a belated or manual run for an older date marks that date at **today's** latest fixing.
[...]

PR #4219, opened 2026-08-08
fix(ledger): make the ČNB fixing's age observable at the point of revaluation

JiRaska:
## What this changes

`#3921` names two problems and only one of them is bounded. This PR does that one, and states
plainly what the other one costs so it is not mistaken for done.

**(a) Staleness was unobservable — fixed here.** The fx→ledger seam carried no time at all
(`CnbRateProvider.cnbRate(base): BigDecimal?`), so there was nowhere in `FxRevaluationService` a
freshness check or metric could even be *written*. The port now returns `CnbFixing(rate,
validFrom)`, and every revaluation attempt publishes `openbank.fx.fixing.age_seconds{currency}` —
the age of the fixing the positions were actually marked at.

**(b) A late fixing can never be incorporated — NOT fixed here, and it is the one with money
consequences.** `idempotencyKey = "fx-reval-{date}"` keys on the business day alone, and
`LedgerService.postJournal` returns the existing entry on a key hit. So once a day has been
revalued, a re-run after a corrected or belated ČNB fixing lands returns the **original** entry,
emits no new event, and reports `posted = true`. A success that changed nothing.

Concretely: every EUR/USD/GBP position marked on that business day keeps a CZK counter-value
computed from the superseded rate, in the general ledger, **indefinitely** — there is no
correcting entry, no versioned key and no expiry. The next day's revaluation marks *that* day
from the current carry, so the error does not self-heal; it is absorbed into the exchange-rate
differences account as a permanent misstatement of the earlier day. Closing that needs a decision
about correcting entries (rate identity in the key, or an explicit reversing entry), not a smaller
patch. The gauge below is what makes the condition visible in the meantime.

## Why a gauge and not a counter

A counter of "revaluations attempted" rises identically whether the rate was minutes or a week
old, and a counter of "stale runs" only moves once someone has picked a threshold in code. The
signal that carries the information is the **age of the last value actually used**, sampled at
scrape time — the same primitive shape as `openbank_workflow_last_success_age_seconds`, one level
down: not "when did the job last succeed" but "how old is the number it succeeded with".

That distinction is the whole point, because the three signals this job already has cannot express
it:

| signal | what a 3-day-old fixing looks like |
|---|---|
| `WorkflowLivenessStale` (`workflow="ledger-fx-revaluation"`) | healthy — `recordSuccess()` is called |
| the posted journal entry | healthy — a normal entry, and it never recorded which rate it used |
| `FxRevaluationSkippedAllLegs` | silent — nothing is skipped, a valid rate *was* found |

fx-service's validity window is three days and **date-blind** (`CNB_VALIDITY_DAYS = 3`;
`FxRepositories` selects the latest row with `validTo > :now`), so a Friday fixing marking a Monday
position is, to everything else, an ordinary successful run.

Two design choices make the gauge alertable rather than decorative:

- **Seeded at registration, not `Instant.EPOCH`.** A fresh pod ages from its own start. An EPOCH
  seed makes every deploy fire the alert continuously until the next daily run — up to 24h of
  noise no `for:` can absorb, because the condition genuinely holds. This is the trap
  `DomainMetrics.registerWorkflowLiveness`'s KDoc documents at length (#2239 Gap 2) and ADR-0237
  point 3 decides against.
- **A failed resolution does not clear the holder.** `fixingObserved(currency, null)` leaves the
  last known instant in place so the published age keeps climbing. Blanking the series — or simply
  not reporting a failed attempt — reads as "nothing to see" on every dashboard and in most alert
  expressions. That is "a table that stopped growing", restated as a metric.

## `FxRevaluationSkippedAllLegs` could not fire

Confirmed on the committed evidence, in both directions:

- `"skipping its revaluation leg"` is emitted from exactly one place in service code —
  `openbank-ledger-service/src/main/kotlin/com/openbank/ledger/application/usecase/FxRevaluationService.kt:64`
  on `origin/main`
  (verified: it is the only `.kt` hit fleet-wide).
- The ledger workload deploys to namespace `ledger` (`openbank-infra/gitops/apps/ledger.yaml:21`).
- The rule selected `{namespace="fx"}`.

The label vocabulary was never in doubt — the same file selects
`namespace=~"ledger|accounts|balances|…"` elsewhere. So the only rule claiming to cover the
skipped-leg path could not match, and its `summary`/`description` attributed the condition to
fx-service, which would have misdirected whoever it eventually paged. Both are fixed; it is a
one-line change independent of everything else here.

## Files

| file | what |
|---|---|
| `application/port/out/CnbRateProvider.kt` | `cnbRate` returns `CnbFixing(rate, validFrom)`; `validFrom` nullable so a payload without it degrades to "age unknown", never to "just now" |
| `infrastructure/client/FxServiceClient.kt` | declares `validFrom` — fx-service has always sent it, this DTO's `@JsonIgnoreProperties(ignoreUnknown = true)` was dropping it |
| `infrastructure/client/FxServiceCnbRateAdapter.kt` | maps it through unchanged |
| `application/port/out/FxFixingFreshnessPort.kt` | new out-port: report the fixing age of every attempt, for every currency, resolved or not |
| `infrastructure/observability/FxFixingFreshnessGauge.kt` | Micrometer adapter, `openbank.fx.fixing.age_seconds{currency}` |
| `application/usecase/FxRevaluationService.kt` | reports every attempt; the journal description now records **which fixing** each leg was marked at |
| `gitops/…/prometheus-rules-fx-fixing.yaml` | new: `FxFixingAgeAbsent` + `FxFixingStale` (60h — just above one weekend, below two), both `severity: warning` per ADR-0237 point 3 |
| `gitops/…/loki-rules-silent-failures.yaml` | `{namespace="fx"}` → `{namespace="ledger"}` + attribution |

A metric nobody alerts on is the same gap one level up, so the `PrometheusRule` ships in this PR
rather than as a follow-up.

## Red then green

`FxFixingFreshnessGaugeTest` and the three new cases in `FxRevaluationServiceTest` assert the
gauge's **value**, never `isNotNull()` — an age gauge asserted only to exist passes against a
holder still sitting at 1970, which is exactly how an `Instant.EPOCH` default survives a full
suite (#3882).

Falsification, per assertion, against the head of this branch with the one behaviour under test
removed:

```
GREEN  (branch head, unmodified)
  FxRevaluationServiceTest      tests=5 failures=0 errors=0 skipped=0
  FxFixingFreshnessGaugeTest    tests=4 failures=0 errors=0 skipped=0

RED  m1 — delete `fixingFreshness.fixingObserved(currency, fixing?.validFrom)`
  tests=9 failures=3
    FxRevaluationServiceTest :: publishes the age of the fixing each leg was marked at, and records it on the entry
    FxRevaluationServiceTest :: a currency with no fixing still publishes an age, so a quiet feed is visible
    FxRevaluationServiceTest :: an fx-service answer without a fixing date leaves the description and the age untouched

RED  m2 — seed the gauge holder at `Instant.EPOCH` instead of `clock.instant()`
  tests=9 failures=2
    FxFixingFreshnessGaugeTest :: a currency never observed ages from registration, not from the epoch
    FxRevaluationServiceTest :: an fx-service answer without a fixing date leaves the description and the age untouched

RED  m3 — drop `${fixingSuffix(fixings)}` from the journal description
  tests=9 failures=1
    FxRevaluationServiceTest :: publishes the age of the fixing each leg was marked at, and records it on the entry
```

Counts are read out of `build/test-results/test/TEST-*.xml`, not off the console. Each mutation
was applied on its own and reverted before the next, and the two pre-existing
`FxRevaluationServiceTest` cases stay green under all three — the new assertions are sensitive to
the new behaviour and to nothing else. m2 is the one worth noting: it is the EPOCH-seed variant,
and without that case a boot-noisy gauge would have shipped looking identical to this one.

## Still open (why `Refs`, not `Closes`)

1. **The late-fixing correctness gap above.** Needs a correcting-entry decision.
2. **Backfill still uses today's rate.** `cnbRate(base)` ignores `command.date`, so a manual run
   for an older date marks that date at the latest fixing. Widening the port to take an `asOf`
   would be a lie until fx-service can answer as-of a date — the `/history` endpoint bounds on
   `validFrom` but the CNB path does not — so that is fx-side work, not a parameter to add here.
3. **fx-service's `openapi.yaml` does not describe the payload it serves.** `ExchangeRateResponse`
   declares `rate`, `spread`, `validAt`; the code serializes `bidRate`, `askRate`, `rateType`,
   `source`, `validFrom`, `validTo`, `createdAt`, `pair`, `midRate`. Ledger has always depended on
   the undocumented `bidRate`/`askRate` and now also on `validFrom`. Correcting it is an fx-side
   spec change (ADR-0048 minor bump; the two fictional properties want deprecating rather than
   removing, since removal classifies as breaking) and belongs in an `fx`-scoped PR — the scope
   selects the released component, so it cannot ride in this one.
4. **`ADR-0237` point 2 (`feed-cnb-fixing` liveness in fx-service) is still `planned`.** No
   `feed-`prefixed workflow exists in code today. That is the complementary signal — "the feed
   delivered" as measured at ingestion, versus this PR's "the value in use is this old" as measured
   at consumption.

Refs #3921
Refs #1302

### DIFF
--- a/docs/threat-models/openbank-ledger-service.md
+++ b/docs/threat-models/openbank-ledger-service.md
@@ -260,3 +260,24 @@ set) apply equally to the new `ledger.approval.decide` action.
   existing `ExceptionMappers.kt`. STRIDE supplement added in §7 above. **`authz.four-eyes.enforce`
   stays `false`** — no behavior change to any existing request; this PR only wires the mechanism.
   Rollback: revert the commit (no DB/schema change; `ApprovalStore` records live in Redis with a TTL).
+- **2026-08-08** — **The ČNB fixing's `validFrom` now crosses the fx-service seam** (issue #3921).
+  `CnbRateProvider` returned a bare `BigDecimal?`, and `FxServiceClient`'s DTO declared only
+  `baseCurrency/quoteCurrency/bidRate/askRate` — fx-service was already serving `validFrom`, and
+  `@JsonIgnoreProperties(ignoreUnknown = true)` silently dropped it. The port now returns
+  `CnbFixing(rate, validFrom)` and `FxServiceCnbRateAdapter` carries it through. **Risk class =
+  information disclosure**, and it is a bounded *decrease* in blind trust rather than an increase:
+  no new endpoint, no new caller, no authorization or NetworkPolicy change, and the field was
+  already on the wire — this side simply stops discarding it. It is a published reference rate, so
+  it discloses nothing about any customer or position. The only new outbound artifact is a metric,
+  `openbank.fx.fixing.age_seconds{currency}`, which carries a currency label and an age in seconds
+  and no amount, account or party.
+  **Why it matters here**: nothing anywhere compared a fixing's age to now. The service-side bound
+  is date-blind (`CNB_VALIDITY_DAYS = 3`, `validTo > :now`), so a stale rate revalues silently, and
+  the one alert meant to catch the silent case — `FxRevaluationSkippedAllLegs` — selected
+  `{namespace="fx"}` while the log line it matches is emitted only by this service, deployed to
+  namespace `ledger`. It could never fire. That selector is corrected in the same change.
+  **Explicitly NOT addressed**: a late or corrected fixing still cannot be incorporated.
+  `idempotencyKey = "fx-reval-{date}"` carries no rate identity, so a re-run returns the existing
+  entry and reports success while every position marked that day keeps the superseded rate's CZK
+  counter-value. That needs a correcting-entry decision, not a patch, and #3921 stays open for it.
+  Rollback: revert; the field is read-only on this side and nothing persists it.
--- a/openbank-infra/gitops/components/observability/loki-rules-silent-failures.yaml
+++ b/openbank-infra/gitops/components/observability/loki-rules-silent-failures.yaml
@@ -103,20 +103,32 @@ data:
 
           # A job that completed successfully having done nothing. The hardest class to alert on,
           # because every metric it touches reports a healthy run.
+          # The selector is `ledger`, not `fx`, and that one word was the whole rule (#3921).
+          # "skipping its revaluation leg" is emitted from exactly one place in service code —
+          # openbank-ledger-service .../usecase/FxRevaluationService.kt — and the ledger workload
+          # deploys to namespace `ledger` (gitops/apps/ledger.yaml). Selecting `fx` matched the
+          # namespace holding fx-service, fx-db and redis, where that string is never logged, so
+          # the only rule claiming to cover the skipped-leg path could not fire at all. The label
+          # vocabulary was never in doubt: this same file selects namespace=~"ledger|accounts|…"
+          # elsewhere. Companion signal, and the one that catches the quieter failure: the
+          # openbank_fx_fixing_age_seconds gauge in prometheus-rules-fx-fixing.yaml, which sees a
+          # rate that is stale-but-valid — a case that logs nothing at all.
           - alert: FxRevaluationSkippedAllLegs
             expr: |
               sum by (namespace) (
-                count_over_time({namespace="fx"} |= "skipping its revaluation leg" [1h])
+                count_over_time({namespace="ledger"} |= "skipping its revaluation leg" [1h])
               ) > 0
             for: 15m
             labels:
               severity: warning
               team: platform
             annotations:
-              summary: "fx-service revalued nothing — no valid rate available"
+              summary: "ledger-service revalued nothing — no valid ČNB rate available"
               description: >-
-                FxRevaluationService found no usable rate and returned posted=false. The run looks
-                successful from every angle: no exception, no failed counter, a green job. The only
-                other evidence is a table that stops growing, and nothing alerts on that. Check the
-                ČNB feed first (check-external-feeds.py probes the URL's payload SHAPE, not its
-                status code — a 404 HTML page is a perfectly valid HTTP response).
+                FxRevaluationService (openbank-ledger-service) found no usable rate and returned
+                posted=false. The run looks successful from every angle: no exception, no failed
+                counter, a green job, and a workflow-liveness heartbeat that records the run as a
+                success. The only other evidence is a table that stops growing, and nothing alerts
+                on that. Check the ČNB feed first — it is ingested by fx-service, one namespace
+                over (check-external-feeds.py probes the URL's payload SHAPE, not its status code
+                — a 404 HTML page is a perfectly valid HTTP response).
--- a/openbank-infra/gitops/components/observability/prometheus-rules-fx-fixing.yaml
+++ b/openbank-infra/gitops/components/observability/prometheus-rules-fx-fixing.yaml
@@ -1,0 +1,82 @@
+# ČNB fixing freshness at the point of use (issue #3921, ADR-0237 point 2).
+#
+# openbank-ledger-service publishes openbank_fx_fixing_age_seconds{currency} — the
+# scrape-time age of the fixing its daily FX revaluation last marked positions at,
+# per currency (FxFixingFreshnessGauge). It is deliberately NOT the same signal as
+# the two rules already covering this job:
+#
+#   * openbank_workflow_last_success_age_seconds{workflow="ledger-fx-revaluation"}
+#     answers "did the job run". A run that marked every position at a three-day-old
+#     fixing calls recordSuccess() and is indistinguishable from a healthy one.
+#   * FxRevaluationSkippedAllLegs (loki-rules-silent-failures.yaml) answers "was any
+#     rate available at all". A stale-but-valid rate logs nothing — fx-service's
+#     validity window is three days and date-blind, so a Friday fixing marking a
+#     Monday position is silent on every existing signal.
+#
+# Boot-safe by construction: the gauge's holder is seeded at registration, not at
+# Instant.EPOCH, so a fresh pod reads as old as itself rather than as decades. That
+# is what allows a plain threshold here with no pod-uptime guard — the trap the
+# workflow-liveness KDoc documents at length (#2239 Gap 2).
+#
+# severity: warning throughout, per ADR-0237 point 3 and #3346: a stale valuation is
+# a correctness gap to work in hours, not a page. Note what this rule does NOT do —
+# a corrected fixing arriving after the run can never be incorporated, because the
+# revaluation's idempotency key is the business date alone with no rate identity, so
+# a re-run returns the original entry and reports success. This alert makes that
+# condition visible; it does not fix it (#3921, still open).
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: openbank-fx-fixing-freshness
+  namespace: observability
+  labels:
+    app.kubernetes.io/part-of: observability
+spec:
+  groups:
+    - name: openbank.fx.fixing
+      interval: 5m
+      rules:
+        # 1. The series is gone fleet-wide. Not "all rates fresh" — either ledger
+        # stopped being scraped or the revaluation never reached a currency loop, and
+        # in both cases every freshness verdict below is void. 25h so a single missed
+        # daily run does not trip it on its own; that is rule 2's job.
+        - alert: FxFixingAgeAbsent
+          expr: absent(openbank_fx_fixing_age_seconds)
+          for: 25h
+          labels:
+            severity: warning
+            team: platform
+          annotations:
+            summary: "No FX fixing-age series from ledger-service for 25h"
+            description: >-
+              openbank_fx_fixing_age_seconds is absent. The daily FX revaluation
+              reports one sample per currency per run, including currencies it could
+              not resolve, so an absent series means the run itself is not reaching
+              that code — check WorkflowLivenessStale for workflow="ledger-fx-revaluation"
+              and the ledger pod log for HR000068 before reading any other FX signal.
+
+        # 2. A position is being valued at a fixing older than two ČNB publication
+        # cycles. 60h absorbs a normal weekend (a Friday fixing is legitimately ~63h
+        # old by Monday morning, so the threshold sits just above one weekend and
+        # below two) plus a public holiday. for: 1h keeps a single stale scrape during
+        # ingestion from firing.
+        - alert: FxFixingStale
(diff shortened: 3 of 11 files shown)


JiRaska:
**`gates (security)` is fixed — it was `threat-model-updated-on-trust-boundary-change`, not a code problem.**

```
threat-model-diff gate: openbank-ledger-service: trust-boundary change without a
docs/threat-models/openbank-ledger-service.md update —
  FxServiceClient.kt (outbound client edge)
  FxServiceCnbRateAdapter.kt (outbound client edge)
```

The demand is correct: a new field crossing a service seam is a data-flow change on a trust boundary whether or not it widens anything. Added the entry; the gate now PASSes locally through the real runner with `PR_DIFF_BASE` set to this branch's merge-base.

The entry states the direction rather than just satisfying the gate: this is a bounded **decrease** in blind trust. fx-service already served `validFrom` and this side was discarding it through `@JsonIgnoreProperties(ignoreUnknown = true)`. No new endpoint, no new caller, no authorization or NetworkPolicy change, and a published reference rate discloses nothing about a customer or a position. The only new outbound artifact is a gauge carrying a currency label and an age in seconds — no amount, account or party.

It also records the two facts a future reader would not otherwise find:

- `FxRevaluationSkippedAllLegs` selected `{namespace="fx"}` while the log line it matches is emitted **only** by this service, deployed to namespace `ledger`. **It could never fire** — verified: the string has exactly one `.kt` occurrence fleet-wide, in `openbank-ledger-service`, and `gitops/apps/ledger.yaml` puts that workload in `ledger`. The selector is corrected in this PR.
- The late-fixing half is **explicitly not addressed**. `idempotencyKey = "fx-reval-{date}"` carries no rate identity, so a re-run after a corrected fixing returns the existing entry and reports `posted = true` while every position marked that day keeps the superseded rate's CZK counter-value — permanently, since the next day marks from the current carry. That needs a correcting-entry decision, which is why this PR is `Refs #3921` and not `Closes`.

Remaining non-green checks are builds still in flight plus `CodeQL (java-kotlin, manual)`. Note `CodeQL` is not a required context — it reports `could not process any code written in Java/Kotlin` on diffs it cannot analyse, and it is what makes `gh pr merge` report `UNSTABLE` and suggest `--admin`. `--squash --auto` is the documented path; `--admin` is not.

ledger and fx are both money-path, so this still needs its **2 human approvals**.

JiRaska:
Reviewed. **Mergeable**, and it does the hard part properly — but there is one case it does not cover, and it happens to be the case this PR exists for.

**The observability is real, not a gauge nobody scrapes.** I checked the whole chain rather than the registration alone: the metric is a Micrometer `Gauge` with no `baseUnit`, so it scrapes as `openbank_fx_fixing_age_seconds`; ledger is in the `PodMonitor` namespace list on port `management` `/q/metrics` and already pulls `quarkus-micrometer-registry-prometheus`; and `prometheus-rules-fx-fixing.yaml` lands via `apps/observability-components.yaml` with `directory.recurse: true` and `ruleSelectorNilUsesHelmValues: false`. So the rule is actually selected.

The best line in the PR is the one that is easy to skim past: the existing Loki rule `FxRevaluationSkippedAllLegs` selected `{namespace="fx"}` while the string is logged by `FxRevaluationService` in the **ledger** namespace — so the only rule claiming to cover the skipped-leg path could never have fired. Fixing that is worth more than the new gauge.

### The finding: for a feed that never resolves, the gauge measures pod uptime

```kotlin
val ref = AtomicReference(clock.instant())            // :51  seeded at registration
...
if (validFrom != null) holder.set(validFrom)          // :58  only overwritten on a real fixing
```

- **Stale-but-valid fixing** → real `validFrom` → genuine, restart-independent age. Correct, and pinned by `FxFixingFreshnessGaugeTest`.
- **No fixing at all** → the holder keeps *pod start*, so the published age is **pod uptime**, not feed staleness.

`FxFixingStale` fires at `> 60 * 3600`, i.e. 60 hours. So detecting a dead feed requires **60 hours of uninterrupted ledger pod uptime**; every deploy, rollout or eviction resets the series to ~0. On a repo that auto-deploys ledger routinely, a dead feed can stay under the threshold indefinitely. And `FxFixingAgeAbsent` cannot catch it either, because the series *is* present — the code deliberately publishes for unresolved currencies too.

That is precisely the ČNB history: the URL was a 404 for the entire life of fx-service, `cnbRate` returned null for days, and every layer stayed green because a 404 is a valid HTTP response. The KDoc concedes "a pod restarted more often than the feed publishes can mask staleness" and argues that needs a restart storm — it needs **one restart per 60h**.

Cheap fix that does not depend on surviving a restart: publish a second series alongside, e.g. `openbank_fx_fixing_resolved{currency}` set 0/1 on **every attempt**, and alert on `== 0 for: 2h`. A boolean written every cycle cannot be reset by a restart into a healthy-looking value.

### Scope, correctly stated

Purely observational — `resolveLeg` behaves exactly as before (null fixing → same WARN, same `continue`; a stale-but-valid fixing still posts), and the only ledger-visible change is the `[fixings EUR@…]` suffix on the journal description. The body is also straight about the correctness half not being fixed: the idempotency key `fx-reval-{date}` carries no rate identity, so a corrected fixing can never be incorporated. Good to have that written down rather than implied.

Minor and worth crediting: `FxRevaluationServiceTest` now constructs the real gauge instead of a mock and asserts gauge **values** rather than `isNotNull()` — deliberately avoiding the assertion shape that passes against a 1970 placeholder.

JiRaska:
Reviewed the 3-dot diff at `cb96406d7`, built a worktree, ran both test classes, and — since a test that has only ever passed proves nothing — ran two **mutation tests** to check the assertions are falsifiable. Details below.

## The gauge does close the hole, and I proved it

`FxRevaluationService.resolveLeg` calls `fixingFreshness.fixingObserved(currency, fixing?.validFrom)` **before** the `if (fixing == null) return null`, so every currency reports on every attempt including unresolved ones, and a `null` deliberately leaves the holder in place so the age keeps climbing instead of the series vanishing. That is exactly the right shape for the ČNB-404 case.

I moved the call below the null-check and re-ran: **5 tests, 1 failed.** The failure path is genuinely covered, not incidentally.

Boot-zero is also handled — `FxFixingFreshnessGauge.kt:53` seeds `AtomicReference(clock.instant())` inside `computeIfAbsent`. Mutated it to `Instant.EPOCH`: **9 tests, 2 failed.** The seed is pinned by value, so this PR does not carry #4208's defect.

## Blocking: the threshold sits below the weekend, not above it

`prometheus-rules-fx-fixing.yaml:65-67` uses `openbank_fx_fixing_age_seconds > 60 * 3600` with `for: 1h`, justified in place by "a Friday fixing is legitimately ~63h old by Monday morning, so the threshold sits just above one weekend".

That premise assumes `validFrom` is the ~14:30 publication time. It is not:

```kotlin
// openbank-fx-service/.../CnbRateIngestionService.kt:70
validFrom = fixing.date.atStartOfDay(Europe/Prague)
```

**Business-day midnight** — 14.5h earlier than the comment assumes. The comment's own arithmetic contradicts it too: 63h after Friday 00:00 is Sunday 15:00, not Monday morning.

What that produces, every week:

- Friday fixing, `validFrom` = Fri 00:00 Prague.
- The gauge is scrape-time computed, so it crosses 60h at **Sun 12:00**; with `for: 1h` the alert fires **Sun 13:00**.
- It clears only when Monday's fixing lands (`FxRevaluationScheduler`, cron `0 0 15 * * ?` Europe/Prague) → **Mon 15:00**, peak age **87h**.
- So ~26h of `severity: warning` on all three currencies **every weekend**.

A Monday public holiday pushes the peak to 111h; Good Friday + Easter Monday (both CZ holidays) reaches **135h**.

This is the specific failure the PR exists to prevent — an alert that fires so routinely people learn to ignore it, at which point the real outage is invisible again. A static threshold that is FP-free needs ~140h, which mostly destroys the signal. The fix wants to be calendar-aware (gate on `day_of_week()`/`hour()`, or compare against fx-service's own last-successful-ingestion timestamp); at minimum raise to >87h and say plainly that holiday clusters still fire.

## Worth fixing

**`FxFixingFreshnessGauge.kt:29-32` — the KDoc's masking argument is inverted.** It says a mask "requires a restart storm" because "the daily job's 2-day threshold is far longer than the pod's restart cadence". Masking requires restarts **more frequent** than the threshold, not less. Concretely, in the never-resolving case the holder is re-seeded to pod start on every restart and never updated, so the age can only ever reach *pod uptime* — and with auto-deploy-on-merge, ledger rarely runs 60h uninterrupted, so `FxFixingStale` would likely **never fire for a fully dead feed**. Not blocking, because the Loki rule covers that case — but the KDoc should say so, rather than claiming a restart storm is needed.

**`FxRevaluationServiceTest.kt` — two `isNotNull()` assertions on an age.** In `a currency with no fixing still publishes an age`, that is exactly the vacuity the sibling test's own KDoc condemns (citing the `AuditEvent.timestamp` lesson). The `containsExactlyInAnyOrder` below carries the real coverage, so the two lines add nothing; assert the seeded value (0.0 under the fixed clock).

## Good catches worth calling out

The Loki `namespace="fx"` → `namespace="ledger"` change is a genuine repair — the string `skipping its revaluation leg` is logged only by ledger-service, and `gitops/apps/ledger.yaml:21` confirms the namespace. **The rule could not fire before.**

The metric name is declared once as a `const` and consumed by both the tests and the alert file, so the double-literal trap is avoided.

## Notes

- `FxFixingAgeAbsent` uses `for: 25h`. `for:`-progress is ruler in-memory state and resets on a Prometheus restart, so a 25h pending window is fragile if the monitoring stack restarts more than daily.
- `FxRevaluationSkippedAllLegs` is named for all legs but its expression is `> 0`, i.e. any leg. Pre-existing.
- `validFrom` is real on the provider side (`FxResource.kt:62`, non-null, declared in its `openapi.yaml`), so this works today — but **no ledger→fx pact exists**, so the new cross-service field dependency has no replay backing it. Pre-existing gap, not introduced here.

## Checked clean

Retention: the Prometheus rules use no range vectors at all (`absent()` and a scalar comparison); the Loki rule uses `[1h]`. Nothing truncates against the 12h retention. Deployment: `gitops/apps/observability-components.yaml:19` syncs the whole directory, and the labels match sibling rule files. No probe-local URL copy — `FxServiceClient` is `@RegisterRestClient(configKey = "fx-service")`, host from committed config, and `FxServiceCnbRateAdapter` correctly does **not** substitute `Instant.now()` when `validFrom` is absent. No `openapi.yaml`/`version.txt` edit needed.

PR #4233, opened 2026-08-08
fix(governance): unbreak main — case-coordinator's three enforced-gate failures

JiRaska:
## `main` is red on three enforced gates, and every open PR inherits them

All three come from one merge — #4215, the ADR-0244 Temporal `CaseWorkflow` slice:

```
FAIL [data]   event-contract coverage ratchet (ADR-0006)
FAIL [kotlin] event-consumer liveness (ADR-0160)
FAIL [kotlin] a service shipping an outbox dispatcher must construct an OutboxMessage (#4007)
```

Verified against **pristine `origin/main`** in a detached worktree, no branch involved, before touching anything.

### 1. Outbox writer — the gate was wrong, not the service

`check-outbox-has-writer.py` defined a writer as *constructing an `OutboxMessage`*. `CaseActivitiesImpl.emitProposal` writes a real row with a plain JDBC `INSERT INTO case_outbox`, and does so **deliberately**:

> Temporal activity threads are plain worker threads with no Vert.x context, and reactive Panache would throw HR000068 there — the same reason agent-service keeps `JdbcAgentProposalRepository`.

A row written by SQL is as much a write as one built through the domain type. The predicate now matches both. **Baselining the service instead would have recorded a claim that is false** — and the baseline is meant to be a list of real debt, not a place to park a gate's blind spot.

The SQL lives inside a string literal, which the existing stripper blanks, so the new predicate reads a *comment-only* strip. Prose describing an insert still must not count — three self-test cases pin that: a real `INSERT` is a writer, a line comment and a KDoc describing one are not.

### 2. Event contract — written, not stubbed

`openbank-contracts/openbank-case-coordinator-agent/asyncapi.yaml` derives every claim from `emitProposal` and `CaseOutboxDispatcher`. It states the **three independent facts** that nothing is published today, rather than asserting it once:

- the topic is declared nowhere under `openbank-infra/` — `grep proposal-events` returns nothing outside this service's own `application.yaml`;
- the service is absent from `auto-deploy.yml`'s `ALL_SERVICES`;
- it has no KafkaUser.

It also flags something worth catching now: the address is a bare `proposal-events`, not the fleet's `openbank.<domain>.<event>` convention. **Free to rename while the topic does not exist; expensive after the first deploy.**

### 3. No consumer — allowlisted with a re-triage note

Same shape as the existing `delegation.events` and `engagement.events` entries: producer ships first so the wire shape is stable before consumers land, which is ADR-0006's own "producers before consumers" ordering. The entry names the intended consumer (ADR-0244 D7's oversight surface) and the condition for removing it.

## Verification

| check | result |
|---|---|
| all three gates, real runner | **PASS** |
| `--self-test` | PASS, including the three new SQL cases |
| widened predicate falsified | turning the `INSERT` into a `SELECT` puts case-coordinator back on the violation list (**rc=1**); restoring returns **rc=0** |
| `gen-rules-opa-data.py` | **no diff** — the edited `rules.yaml` key is outside OPA's read set, so no bundle restamp and no fleet pod-roll |

The last row matters for merge order: this PR does **not** restamp the 42 OPA bundles, so it will not conflict with the bundle PRs in flight.


JiRaska:
**#4232 and #4233 are duplicates on two of their three files, and both are green.** Merging both will collide; merging either alone unbreaks `main`. Worth deciding deliberately rather than by merge order.

`main` is red on three enforced gates, all from #4215:

```
event-contract coverage ratchet          — no asyncapi for proposal-events
event-consumer-liveness                  — proposal-events published, no consumer anywhere
outbox-has-writer                        — ships an *OutboxDispatcher, constructs no OutboxMessage
```

Both PRs address all three. The overlap:

| | #4232 | #4233 |
|---|---|---|
| `openbank-contracts/…/asyncapi.yaml` | added | added (different prose, same purpose) |
| `openbank-libs/governance/rules.yaml` | allowlist entry | allowlist entry (different reason) |
| third gate fixed by | **the code** — `CaseActivitiesImpl` builds an `OutboxMessage` | **the gate** — `check-outbox-has-writer.py` also matches a plain-JDBC `INSERT INTO …outbox…` |

### On that third row, which is the real decision

I checked both rather than assuming one is a workaround.

**#4233's gate change is not an accommodation — the matcher had a genuine false-positive shape.** A Temporal activity thread carries no Vert.x context, so reactive Panache throws `HR000068` there and the service must insert through plain JDBC. Matching only `OutboxMessage(` therefore calls such services writerless, and it cites a second instance beyond this one: agent-service's `JdbcAgentProposalRepository`. So the fix has fleet value, not just local. It also handles the subtlety correctly — the SQL lives inside a string literal, so it cannot use the strings-blanking stripper, but prose describing an insert must still not count, hence comments-only stripping.

**#4232's code change is safe** — I checked specifically whether it introduces the `HR000068` trap, and it does not: the KDoc keeps "plain JDBC (Agroal) rather than reactive Panache" and the change only builds the domain value before the same JDBC write.

### One correction worth making

#4232's title says the fixes were *"dropped from #4215's squash"*. They were not dropped — they were never there. #4215 had a **single commit** (`e6ab28b7e`), its head and that commit are the same, and its merge commit `4405a74fa` does not contain the asyncapi file. So this is not the auto-merge-drops-a-later-push failure; #4215 simply merged with these gates failing. Worth stating because the two have different remedies, and "the squash ate it" points at tooling that is working correctly.

### Suggestion

Take **#4233** for the gate fix and the allowlist entry — its reason is the more useful of the two, because it names three verifiable facts *and* flags that `proposal-events` should be renamed to the fleet's `openbank.<domain>.<event>` convention **while the topic is still free**. Every topic on `main` follows that shape; this would be the first exception, and renaming after the first deploy is a migration rather than an edit.

Then rebase the other and keep only its unique part.

JiRaska:
Found sweeping the open queue: this PR and #4232 both fix the same three gates #4215's squash dropped, and take opposite positions on the outbox-writer finding — not a mechanical duplicate, a real design disagreement.

**This PR**: widens `check-outbox-has-writer.py`'s predicate to accept a plain SQL `INSERT` as a writer, arguing the service is correct as-is and the gate's definition was too narrow.

**#4232**: takes the opposite position — a straight cherry-pick of the original dropped commit (`aab88ab6d`/`1be6c7b01`, unedited), which instead changes `CaseActivitiesImpl.emitProposal` to construct an `OutboxMessage`, matching the gate's existing (narrower) definition.

Both are `CLEAN`/`MERGEABLE` right now and both touch `openbank-contracts/openbank-case-coordinator-agent/asyncapi.yaml` + `openbank-libs/governance/rules.yaml`, so whichever lands second gets a real conflict on those two files, not a silent drop.

**One gap worth surfacing**: #4232's cherry-pick also carries `Ids.newId()` instead of bare `UUID.randomUUID()` for the Temporal workflow id (ADR-0106) — this PR's scope doesn't touch that at all (only the three gates named in the body). If this PR is the one that lands, that ADR-0106 fix needs to be picked up separately or it stays lost from #4215's original commit.

Not picking a side — whether `check-outbox-has-writer.py`'s definition of "writer" should widen to accept the raw-JDBC pattern (and whether raw JDBC is the right choice here specifically because a Temporal activity thread has no Vert.x context) versus whether `emitProposal` should construct an `OutboxMessage` instead, is a real engineering call for whoever owns this area, not a duplicate-PR cleanup.

JiRaska:
Flagging that #4232 and #4233 fix overlapping ground and conflict with each other (`add/add` on `openbank-contracts/openbank-case-coordinator-agent/asyncapi.yaml`), so only one should land.

**#4232** (created 16:11) cherry-picks `1be6c7b01` unchanged — the exact commit #4215's squash dropped. It fixes the four things that commit fixed: the asyncapi contract, `Ids.newId()`, the `rules.yaml` allowlist entry, and (per its own description) the outbox writer gap.

**#4233** (created 16:33) rewrites the fix rather than replaying it, and finds something #4232 does not touch: `check-outbox-has-writer.py` only matched `OutboxMessage(` construction, so it never gave `CaseActivitiesImpl` credit for writing the outbox row via a direct `INSERT INTO case_outbox` — which the service uses because a Temporal activity thread carries no Vert.x context (HR000068), so reactive Panache is unavailable there. Verified against the actual source on `origin/main`: `CaseActivitiesImpl.kt` does write via `INSERT INTO case_outbox`, `INSERT INTO case_workflow`, `INSERT INTO case_contribution` — not via `OutboxMessage(...)`.

So #4232 re-lands a fix that (if I'm reading #4233 right) was itself incomplete — the outbox-writer gate would have kept failing on case-coordinator-agent even with #4232 merged, because the checker's pattern doesn't recognise the write shape this service actually uses. I did not verify this claim by running #4232 with `check-outbox-has-writer.py` as it stood before #4233's fix; only that #4233's post-fix checker passes and that the SQL INSERT shape is real.

Ran #4233's `--self-test`: 11 cases pass, including three new ones — direct SQL INSERT into the outbox table IS a writer, a comment describing an INSERT is NOT, a KDoc describing an INSERT is NOT. Full gate run (`outbox-has-writer`, `event-contract-coverage-ratchet`, `event-consumer-liveness`) on #4233's branch: PASS=3.

Worth deciding which of these lands, since neither alone may be sufficient and both together conflict.

JiRaska:
Closing the loop on the #4232-vs-#4233 flag from before this merged: confirmed on current `main` that the gap I called out is real.

`#4232` (closed, unmerged) also carried an `Ids.newId()` (ADR-0106, UUIDv7) fix that this PR's scope didn't touch — separate from the outbox-writer question, in `recordCaseOpened`:

```
openbank-case-coordinator-agent/.../CaseActivitiesImpl.kt:129
    ps.setObject(P1, UUID.randomUUID())
```

still on `main` post-merge, unchanged. Minor (a UUIDv4-vs-UUIDv7 convention gap on one row-id parameter, not a functional bug — case-coordinator is new enough that it hasn't accumulated rows this would bite yet), but worth a follow-up now that #4232 is closed and nobody's watching that diff anymore.

PR #4242, opened 2026-08-08
fix(ci): auto-close deploy-drift trackers when nothing drifts anymore

JiRaska:
## Problem

The `Escalate drift per service` step carries **both** halves of the drift-issue lifecycle:

1. escalate/refresh/reopen trackers for drifted services,
2. **auto-close** trackers whose service caught up with main.

But the step was gated on `steps.probe.outputs.code == '2'` (at least one service drifting). With a fully in-sync fleet (`code == '0'`) the step was skipped entirely — including the auto-close half — so **the last remaining open `deploy-drift` tracker could never be closed by the watch**.

Measured today: run [31277854742](https://github.com/JiRaska/open-bank-oss/actions/runs/31277854742) (workflow_dispatch) reported `56 in sync, 0 drifted/unverifiable` and #3366 (sepa-payment, drift resolved by deploy PR #4222) stayed OPEN; it was closed by hand with a comment citing the run. Yesterday's auto-close of #3363 only fired because sepa-payment was still drifting and kept the step alive.

## Fix

Run the step on a clean scan (`code == '0'`) too. With an empty `drifted` list the escalate half is a no-op (the `for` loop iterates nothing, no issue is created or reopened) and the auto-close half closes whatever caught up. Behaviour with drift present is unchanged.

## Testing

- This PR touches `deploy-drift-watch.yml`, so the PR lane (declaration gate + scanner self-test) runs on it.
- The zero-drift path executes the same github-script with an empty `drifted` array; the only new behaviour is the auto-close loop running — verified by reading, since the sandbox currently has no open drift tracker to close (the next real exercise is the daily cron once a tracker outlives its drift).

Refs #3344

### DIFF
--- a/.github/workflows/deploy-drift-watch.yml
+++ b/.github/workflows/deploy-drift-watch.yml
@@ -98,8 +98,14 @@ jobs:
                exit "$code" ;;
           esac
 
+      # Runs on a clean scan (code 0) too: the auto-close half of this step is the
+      # only thing that closes a tracker whose drift has resolved. Gating on drift
+      # alone meant the LAST open deploy-drift issue was never auto-closed — with
+      # zero drifted services the step was skipped and the issue sat open forever
+      # (measured 2026-08-08: run 31277854742 reported 56 in sync / 0 drifted and
+      # #3366 stayed open; it was closed by hand).
       - name: Escalate drift per service
-        if: ${{ github.event_name != 'pull_request' && steps.probe.outputs.code == '2' }}
+        if: ${{ github.event_name != 'pull_request' && (steps.probe.outputs.code == '0' || steps.probe.outputs.code == '2') }}
         uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9.0.0
         with:
           script: |


JiRaska:
The fix is additive, not a rewrite of the condition — `code == '0'` is added alongside the existing `code == '2'`, so the drift path is untouched and only the previously-skipped clean-scan path gains a run. Confirmed that leaves `code == '1'` (presumably scan failure) still excluded, which is the right default — a failed probe shouldn't drive auto-close either.

The root cause description matches the actual bug class this repo keeps finding: a step gated on the wrong half of its own logic silently drops the other half, and "the aggregate check reports success" hides it. Here the failure mode was inverted from the usual one (#3629 was a gate that never ran; this is a step whose SECOND function — auto-close — could structurally never fire, since the only way to reach `code == '0'` is a fully-caught-up fleet, and that's exactly when the gate excluded it) — same shape, worth recognizing as the same class.

The `#3366 sat open, closed by hand` detail is good evidence this isn't hypothetical.

Not merging — workflow change, human review by rule.

JiRaska:
Reviewed. **The central question — can a broken or empty probe auto-close a live tracker? — answers NO**, and the guard is genuinely there rather than assumed. One scoping issue worth fixing first.

## The guard holds

Auto-close is the dangerous direction, because a wrongly-closed tracker is silent. Reading `check-deploy-drift.py:run_full` and running its self-test:

- `find_deployed_images()` empty → `return 1` ("the scan is broken, not the fleet") → the probe step exits 1 → the job fails → the escalate step has no `always()`, so it never runs, and `steps.probe.outputs.code` is unset anyway.
- Every image skipped → `return 1`, *"The watch would report green over nothing."*
- `--json` is written **before** the 0/2 return, so the `code == '0'` path always finds a populated `/tmp/deploy-drift.json` — no `readFileSync` crash.
- The probe touches no credentials and no API at all (git plus committed manifests), so there is no "empty because unauthenticated" path to begin with.

So `code == '0'` genuinely means "N services enumerated, all ok". The change is sound.

## Worth fixing: partial enumeration auto-closes trackers nobody measured

`deploy-drift-watch.yml:206` closes `openByMarker` minus `drifted` with *"Deployed build caught up with main — closing."*

That set includes services that were **not scanned at all**, not merely services that came back ok. `evaluate_service()` returns `None` when a module has no `version.txt`, and `find_deployed_images()` drops a service whose gitops manifest was removed or renamed. The exit-1 guard only fires on a *totally* empty scan, so a partial narrowing sails through as code 0 and closes trackers with a claim nobody made.

This existed already for `code == '2'` runs, but this PR makes it fire on the **common** path — a clean nightly — so the blast radius goes from rare to daily.

It is also the repo's own "a comment that explains away a symptom retires the question" antipattern, now automated: the tracker is closed with a reassuring sentence, and the next person finds the question already answered.

Two lines:

```js
const ok = new Set(report.entries.filter(e => e.status === 'ok').map(e => e.service))
```

then close only markers in `ok`, and leave the rest open. Closing on positive evidence rather than on absence of evidence.

## Checked

No file overlap with #4172 (`deploy-drift-watch.yml` vs the verification-metadata scripts), and I merged both onto `origin/main` in a scratch worktree to be sure — clean, both workflows parse, `workflow-run-step-size` PASS, both self-tests green. Worth doing because this repo has a documented history of two individually-green PRs breaking `main` together.

Neither PR greps a job log, so that trap does not arise; neither reacts to `workflow_run`, so the latest-attempt-vs-`attempts/<n>` trap is not applicable. `PIPESTATUS` at `:86` is pre-existing and correct (GitHub's default shell is bash), and `| tee` + `PIPESTATUS[0]` correctly avoids the exit-code masking a `| tail` would have introduced.

`check-deploy-drift.py --self-test` green including the flagging paths; `deploy-drift-declaration` PASS on the head and on the merged pair.

Issue #4240, opened 2026-08-08
A PR merged past a red required check and nothing noticed (#4215); the three case-coordinator gates it broke are now fixed


JiRaska:
> **This issue was rewritten on 2026-08-09.** Its original body contained a false claim about `main red watch` and has since been overtaken by #4233 and #4245. What was wrong, and what is left, is recorded below rather than deleted — see the comment thread for the correction as it was made.

## What is still open

**#4215 merged while `Validate manifests` — a required context — reported FAILURE**, and nothing in the repo would have noticed.

Required contexts on the `main-protection` ruleset are `all-green`, `Validate manifests`, `Gitleaks`, `issue-hygiene`, `OPA policy gate`. On #4215's head SHA three of its checks were present and red:

```
FAILURE  gates (kotlin)
FAILURE  gates (data)
FAILURE  Validate manifests    <- required
```

Present and failing, not absent or pending — so the ruleset was overridden rather than satisfied. Consequence was `main` red for 5h15m (15:49:52Z merge, 21:05:37Z fix in #4233), during which other PRs merged onto it and inherited a red they did not cause.

### It is one incident, not a pattern

Measured over the 45 most recently merged PRs, checking each one's five required contexts by name:

```
merged PRs whose required contexts were NOT all SUCCESS: 1
  #4215  Validate manifests=FAILURE
```

That matters for what the fix should be: this does not call for tightening the ruleset, it calls for the override leaving a trace that something reads.

### Nothing detects it

`grep -l required_status_checks .github/scripts/*` returns nothing. `main red watch` (#4019) is the closest, and it is a different question by design — it watches main's CI conclusion *after* the fact, not whether a merge respected its gates. It did its job here (see below); it just cannot answer this.

## What is resolved — and by what

| original finding | status |
|---|---|
| `event-consumer-liveness` — `proposal-events` has no consumer | **fixed**, #4233 (allowlist entry) |
| `event-contract-coverage-ratchet` — no AsyncAPI contract | **fixed**, #4233 (`openbank-contracts/openbank-case-coordinator-agent/asyncapi.yaml`) |
| `outbox-has-writer` — blind to the raw-SQL write idiom | **fixed**, #4233 (SQL predicate) + #4245 (string-aware strip) |

All three gates PASS on current `main`, verified in a clean detached worktree.

## Correction: `main red watch` DID report this

[...]

PR #4243, opened 2026-08-08
fix(ci): teach the outbox-writer gate the raw-SQL write idiom

JiRaska:
## What

Teaches `check-outbox-has-writer.py` the second outbox write idiom in this fleet: a raw-SQL `INSERT INTO <x>_outbox`. One of the three enforced gates currently red on `main`.

## Why this was a blind spot, not a miss

The gate detected a writer solely by the substring `OutboxMessage(`, matched against a body produced by `strip_comments_and_strings()`. The other idiom lives **entirely inside a string literal**, which that function removes before matching — so no input could ever have made it visible. `grep -ci 'insert into|jdbc|preparedstatement'` over the old script: **0**.

`openbank-case-coordinator-agent` writes its row with plain JDBC:

```
INSERT INTO case_outbox
    (event_id, aggregate_id, event_type, payload, status, created_at, updated_at)
```

and `CaseActivitiesImpl`'s KDoc says why, in advance:

> Temporal activity threads are plain worker threads with no Vert.x context, and reactive Panache would throw HR000068 there — the same reason agent-service keeps `JdbcAgentProposalRepository`. The outbox DISPATCH side (`CaseOutboxDispatcher`) runs on the Quarkus scheduler with a Vert.x context and reads through the Panache entity as usual.

The gate reported "the dispatcher will poll a table nothing writes to" about a table that is written to, on every PR since `4405a74fa`. Same shape as the repo's own note that grepping for a JSON key finds hand-built maps and never a serialised data class — a probe that knows one idiom reports the other as absent, and the answer reads as a finding rather than as a fact about the probe.

## How

The two idioms need **opposite** treatment of literals, which is why one pass cannot serve both:

- `OutboxMessage(` → the original strings-stripped body, **untouched** (its 9 self-tests still pass unchanged).
- `INSERT INTO …outbox` → a new comment-stripped body that **keeps** strings.

`strip_comments_keep_strings` is a scanner rather than a regex because comment and string openers alias each other: a `//` inside SQL must not start a comment, a `"` inside a comment must not open a string, and Kotlin raw strings must be consumed first or the opening `""` reads as an empty string and the SQL is treated as code. Each of those three is a self-test case.

## Measured, not asserted

Old logic vs new, on the real tree, diffing the per-service violation sets:

```
flipped violation -> writer: ['openbank-case-coordinator-agent']
flipped writer -> violation: none
```

```
old:  33 service(s) ship a dispatcher; 9 without a writer (8 baselined, 1 new, 0 stale)
new:  33 service(s) ship a dispatcher; 8 without a writer (8 baselined, 0 new, 0 stale)
```

Exactly one service flips. **All 8 baselined services remain violations** — that is the number that matters here, because the obvious way to get this wrong is a regex broad enough to match any `INSERT`, which would mark the whole baseline as fixed and silently retire the gate. It would show up above as baseline entries going stale; none did.

## Self-test: 9 → 17

The 8 new cases are deliberately weighted toward **negatives**, since those are what keep the widening honest:

| case | expected |
|---|---|
| raw-string `INSERT INTO case_outbox` | writer |
| lower-case `insert into outbox_message` | writer |
| `INSERT INTO party_events` (non-outbox table) | **not** a writer |
| SQL named in a KDoc | **not** a writer |
| commented-out INSERT | **not** a writer |
| INSERT inside a NESTED block comment | **not** a writer |
| `//` inside the SQL literal | writer (not swallowed) |
| quote inside a preceding comment | writer (not swallowed) |

The positive case is built from the shape that actually broke the gate, not invented.

`run-gates.py --only outbox-has-writer` → PASS (the gate's `run:` executes both the self-test and the real scan).

## Scope

This fixes **one** of the three enforced gates red on `main`. The other two are real gaps in case-coordinator and are not touched here:

- `event-consumer-liveness` — `proposal-events` has no consumer in the fleet
- `event-contract-coverage-ratchet` — no `openbank-contracts/openbank-case-coordinator-agent/asyncapi.yaml`

So `main` stays red after this merges. #4240 tracks all three plus the governance question of how #4215 merged past a failing required check.

## Linked issues

Refs #4240

## References

- #4007 (the gate's origin), #4215 / `4405a74fa` (the commit that exposed the blind spot)
- `openbank-case-coordinator-agent/src/main/kotlin/com/openbank/casecoordinator/infrastructure/workflow/CaseActivitiesImpl.kt`

### DIFF
--- a/.github/scripts/check-outbox-has-writer.py
+++ b/.github/scripts/check-outbox-has-writer.py
@@ -68,6 +68,30 @@ BASELINE = {
 DISPATCHER_RE = re.compile(r"class\s+\w*OutboxDispatcher\b")
 DECLARATION_RE = re.compile(r"(data\s+)?class\s+OutboxMessage\s*\(")
 
+# THE SECOND WRITE IDIOM — raw SQL (#4240).
+#
+# There are two ways to write an outbox row in this fleet, and the first version of this gate could
+# see only one. `openbank-case-coordinator-agent` writes with a plain JDBC `INSERT INTO case_outbox`
+# because Temporal activity threads are bare worker threads with no Vert.x context, where reactive
+# Panache throws HR000068 — the same reason agent-service keeps `JdbcAgentProposalRepository`. It is
+# deliberate, documented in `CaseActivitiesImpl`'s KDoc, and the dispatch side still reads through
+# the Panache entity.
+#
+# The gate reported it as "polls a table nothing writes to", which was not a close call: the writer
+# is a string literal, and `strip_comments_and_strings` removes string literals BEFORE matching, so
+# the JDBC idiom was structurally invisible — no input could ever have made it visible. That is a
+# fact about the probe, not about the service, and it went red on `main` for every PR (#4240).
+#
+# So the SQL check runs over a comment-stripped body that KEEPS strings, while the
+# `OutboxMessage(` check keeps its original strings-stripped body — the two idioms need opposite
+# treatment of literals, which is exactly why one pass could not serve both. The table-name shape is
+# `<something>outbox`, matching every outbox table in the tree (`case_outbox`, `security_outbox`,
+# `outbox_message`), and an INSERT into any OTHER table is not a writer.
+SQL_WRITE_RE = re.compile(
+    r"INSERT\s+INTO\s+[\"'`]?(?:\w+\.)?\w*outbox\w*",
+    re.IGNORECASE,
+)
+
 
 def strip_comments_and_strings(src: str) -> str:
     """Remove string literals, line comments and NESTED block comments.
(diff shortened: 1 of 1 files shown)


JiRaska:
This was opened **8 seconds after #4233 merged** (#4233 at `21:05:37Z`, this at `21:05:45Z`), and #4233 already landed the raw-SQL idiom in the same file. `main` is green on this gate now — verified against a fresh `origin/main` worktree at `2f23b4d16`, not a local checkout:

```
1 gates  PASS=1
```

So this is no longer the fix that unblocks anything; it is a rewrite of code that landed minutes earlier. I measured what it actually changes rather than reading the diff size:

- **Applied this PR's script to `origin/main`'s tree: `PASS=1`, self-test exit 0.** Same verdict as the merged version, on every service.
- The genuine delta is the broadened table pattern — `[\"'\`]?(?:\w+\.)?\w*outbox\w*` accepts a quoted or schema-qualified name. Grepping every `INSERT INTO` in `openbank-*/src/main` returns 14 sites and **not one** is quoted or schema-qualified. The only outbox insert in the fleet is `CaseActivitiesImpl`, plain and unquoted. So the hardening covers zero cases today.
- The rest (+110/−44) is renaming `strip_comments_only` to `strip_comments_keep_strings` and expanding the comment block.

Closing on that basis. Reopening is cheap if the regex hardening is wanted as a deliberate future-proofing — it is defensible, just not urgent, and it reads very differently as "harden a pattern" than as "fix the red gate".

**One correction worth carrying forward, because both PRs state it.** The justification cites agent-service's `JdbcAgentProposalRepository` as a second service the old matcher was wrong about. It is not: that repository inserts into `agent_proposal`, not an outbox table, and agent-service ships **no** `*OutboxDispatcher` at all — so it never enters this gate's scope in either direction. The gate only examines services that have a dispatcher. The case-coordinator finding was real and the fix is right; the second example is not one, and it would mislead whoever reads the rationale next time this gate is touched.

PR #4245, opened 2026-08-08
fix(ci): make the outbox-writer SQL strip string-aware

JiRaska:
## What

Makes the outbox-writer gate's SQL strip **string-aware**. Narrow follow-up to #4233, which landed the SQL write predicate independently while this was being written — this PR is only the part #4233 does not have.

## The defect

#4233 added `strip_comments_only()` to feed the new SQL predicate, because the INSERT lives inside a string literal that `strip_comments_and_strings()` blanks. That new function walks `/*` without tracking literals, so it loses the invariant its sibling states two definitions above:

> Strings first: a `"/*"` inside a literal must not open a comment.

Concretely, `val a = "/*"` opens a comment that swallows every following INSERT:

```
val a = "/*"
val s = """INSERT INTO case_outbox (id)"""     <- invisible; service reported writerless
```

False-negative direction, which is the dangerous one for this gate: it prints *"the dispatcher will poll a table nothing writes to"* about a table that is written to, and that reads as a clean finding rather than as a broken probe — the exact shape that put this gate on `main`'s critical path (#4240).

## Fix

A scanner that tracks raw strings, normal strings with escapes, line comments and nested block comments. Comment and string openers alias each other, so a regex is the wrong tool. Raw strings are consumed first, or the opening `""` of a `"""…"""` reads as an empty literal and the SQL inside is scanned as code.

## Measured

Fleet verdict **byte-identical** before and after:

```
33 service(s) ship a dispatcher; 8 without a writer (8 baselined, 0 new, 0 stale)
```

No service in the tree trips this today. It is a **latent** defect, so the self-test is the only thing that can hold it — there is no failing service to point at.

## Self-test 12 to 15, and an honest count

Exactly **one** of the three new cases fails against the implementation it replaces:

| case | main (#4233) | this PR |
|---|---|---|
| `"/*"` inside a literal | **False** (wrong) | True |
| `//` inside the SQL literal | True | True |
| quote inside a preceding comment | True | True |

The other two pass either way. They are labelled in the code as regression tests rather than as evidence — I first wrote a comment claiming all three were measured failing, which was wrong, and that is the unverified-evidence habit this file's own header warns about.

## Relationship to #4233

Same diagnosis, same approach, arrived at in parallel; #4233 merged at 21:05Z. Its predicate and its baseline are untouched here — only the stripper is replaced. My original wider diff (#4243) was closed rather than re-litigated, so there is one authoring of the SQL predicate in this file, not two.

## Linked issues

Refs #4240

## References

- #4233 (`2f23b4d16`), #4007 (the gate's origin), #4102

### DIFF
--- a/.github/scripts/check-outbox-has-writer.py
+++ b/.github/scripts/check-outbox-has-writer.py
@@ -108,21 +108,56 @@ def strip_comments_only(src: str) -> str:
     The SQL predicate needs this: an INSERT lives inside a string literal, which
     strip_comments_and_strings() blanks. Prose describing an insert must still not count, so the
     comments still go.
+
+    STRING-AWARE, and it has to be. Its sibling above states the invariant this function must also
+    honour — "a `\"/*\"` inside a literal must not open a comment" — and gets it for free by
+    blanking literals first. This one cannot blank them (the SQL *is* a literal), so it has to
+    track them, which makes a regex the wrong tool: comment and string openers alias each other.
+    The first version walked `/*` blind to literals, so `val a = "/*"` opened a comment that
+    swallowed every following INSERT and reported a real writer as writerless — measured, and the
+    one case in the self-test below that separates the two implementations (#4240). Raw strings are
+    consumed first: otherwise the opening `\"\"` of a `\"\"\"…\"\"\"` reads as an empty literal.
+
+    No service in the tree trips this today — the fleet verdict is byte-identical before and after
+    (33 dispatchers, 8 baselined, 0 new, 0 stale). This is a latent-defect fix, so the self-test is
+    the only thing that can hold it: there is no failing service to point at.
     """
-    src = re.sub(r"//[^\n]*", "", src)
     out: list[str] = []
-    depth = i = 0
-    while i < len(src):
+    i, n, depth = 0, len(src), 0
+    while i < n:
+        if depth:
+            if src.startswith("/*", i):
+                depth += 1
+                i += 2
+            elif src.startswith("*/", i):
+                depth -= 1
+                i += 2
+            else:
+                i += 1
+            continue
         if src.startswith("/*", i):
-            depth += 1
+            depth = 1
             i += 2
             continue
-        if src.startswith("*/", i) and depth:
-            depth -= 1
-            i += 2
+        if src.startswith("//", i):
+            while i < n and src[i] != "\n":
+                i += 1
             continue
-        if not depth:
-            out.append(src[i])
+        if src.startswith('"""', i):
+            end = src.find('"""', i + 3)
+            end = n if end == -1 else end + 3
+            out.append(src[i:end])
+            i = end
+            continue
+        if src[i] == '"':
+            j = i + 1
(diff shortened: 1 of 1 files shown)


JiRaska:
Verified independently. **This is a genuine latent-defect fix**, and the falsification holds up — which matters here more than usual, because the PR's own claim is that nothing in the fleet is affected today, so the self-test is the only thing that can hold it.

The bug is in `strip_comments_only`, merged an hour ago in #4233: it walked `/*` blind to string literals, so `val a = "/*"` opens a comment that swallows every following `INSERT`. The sibling `strip_comments_and_strings` states that exact invariant and gets it for free by blanking literals first; this one cannot, because the SQL **is** a literal — which is precisely why a regex is the wrong tool, since comment and string openers alias each other.

What I ran on `origin/main`'s real tree:

```
main's version    33 dispatchers; 8 without a writer (8 baselined, 0 new, 0 stale)
#4245's version   33 dispatchers; 8 without a writer (8 baselined, 0 new, 0 stale)
```

Byte-identical, so the "no service trips this today" claim is measured rather than asserted.

Then the part that actually decides it. A self-test that only ever passes proves nothing, so I grafted `main`'s stripper body into #4245's self-test, keeping the new cases:

```
FAIL a "/*" inside a literal must not open a comment and swallow the INSERT:
     want (True, True), got (True, False)
```

So the new case is falsifying, it names the defect precisely, and it is the one case that separates the two implementations. Handling raw strings first — so the opening `""` of a `"""…"""` does not read as an empty literal — is right too.

One note on framing rather than code. The failure direction here is the loud one: a swallowed `INSERT` makes a real writer look writerless, so the gate goes **red** on a correct service. That is worth stating explicitly in the PR body, because "latent defect in an enforcing gate" reads as though it could have waved something through, and this one could not. The quiet direction would be the opposite mis-parse, and it does not arise.

Merging this needs a human — it changes CI gate machinery, which decides what the fleet enforces.

JiRaska:
Ran the self-test and the fleet scan rather than trusting the description.

```
--self-test    13/13, including three on the exact failure mode described:
                 "/*" inside a literal must not open a comment and swallow the INSERT
                 a // inside the SQL literal does not hide the INSERT
                 a quote inside a comment does not swallow the following SQL   (symmetric case)
against main   33 services, 8 without a writer (8 baselined, 0 new, 0 stale) — byte-identical
               to #4233's verdict, confirming this changes robustness, not the current result
```

The false-negative framing is right and worth restating: `strip_comments_only()` treating `"/*"` inside a string as a real comment opener would silently reclassify a service that DOES write its outbox as one that doesn't — printing "the dispatcher will poll a table nothing writes to" about a table something writes to, which is a worse failure than the false positive the whole gate exists to prevent. A checker for "did anyone write this" being wrong in the direction of "no one did" is the dangerous half.

Landing this promptly matters more than usual — every PR touching `openbank-*-agent`-shaped Kotlin sources that happens to declare a string containing `/*` (a regex, a comment marker, anything) is exposed to exactly this false-negative until it merges.

Issue #3806, opened 2026-08-06
21 one-replica canary Rollouts (15 money-path) cannot roll on a full cluster either — and the #3545 gate counts them as declared


JiRaska:
## What

All **21** Argo `Rollout`s in this fleet are `replicas: 1` with an empty canary `maxSurge` /
`maxUnavailable`, so they take the Argo defaults — `maxSurge: 25%` -> 1, `maxUnavailable: 25%` ->
0. A canary at one replica therefore needs room for a **second** pod before the stable pod may be
released, which is exactly the deadlock #3545 describes for Deployments. **15 of the 21 are
money-path.**

Measured against `origin/main` 2026-08-06:

```
21 Rollouts, every one replicas: 1, every one canary.maxSurge/maxUnavailable unset.
money-path: account, balance, consent, fraud, fx, ledger, lending, transaction, sepa-payment,
            domestic-payment, sepa-instant, clearing, settlement, swift, vop, sanctions, sca
other:      customer-edge, kyc, onboarding, party
```

## Why nothing reports it

`check-single-replica-rollout-strategy.py` counts a canary block as a *declared* strategy, so
these 21 are not findings — correctly, by its own design, and its docstring says as much:

> **Canary Argo Rollout** (every money-path service here). The canary pod alongside stable IS the
> strategy — terminate-first would defeat it. The real fix for those is `replicas >= 2`.

So this is not a gate defect. It is the half of #3545 the gate deliberately declined to answer,
and it is currently answered nowhere. #3805 closes the Deployment half (64 -> 0) and leaves this
one open rather than sweeping it, because the answer here costs money and changes running
capacity.

There is a second, documented failure mode stacked on top: a Rollout whose revision 1 never became
healthy pins `stableRS` forever and loops `Progressing` -> `Degraded` while a later revision serves
happily (`openbank-infra/CLAUDE.md`, seen live on vop-service). A one-replica canary that cannot
schedule its canary pod is one bad revision away from that state.

## Why terminate-first is not the fix here

Setting `maxSurge: 0` / `maxUnavailable: 1` on a canary Rollout removes the canary pod, i.e. it
trades away the progressive-delivery property these services were given deliberately. On a
money-path service that is a downgrade, not a fix.

## What the fix costs — the numbers

The real answer is `replicas >= 2`, which also buys the HA a single-pod money path does not have
today. Doubling all 21:

| | |
|---|---|
| added CPU requests | **+4.62 vCPU** |
[...]

PR #4246, opened 2026-08-08
ci(gitops): make the #3545 gate ask whether a canary is REALISABLE, not merely declared

JiRaska:
## What

Makes the #3545 gate ask a second question of every canary Argo Rollout: **can the weights it
declares actually be produced by the replica count it declares?**

It could not before, and said so in its own docstring — a canary block counted as *declared* and
the check stopped there. That is the "green about work it never did" class: the 21 workloads the
gate most needed to look at were the 21 its scope silently excluded.

## Re-derived counts (issue says 21 / 15)

Measured on `origin/main` at `2f23b4d16` and live, 2026-08-08:

| | manifests | live (`kubectl get rollout -A`) |
|---|---|---|
| Rollouts in the fleet | **21** | **21** |
| `replicas: 1` | **21** | **21** (`spec.replicas=1`, `ready=1`, `available=1`) |
| canary strategy | **21** | 21 |
| canary with `replicas > 1` | **0** | 0 |
| `canary.trafficRouting` set | **0** | 0 |
| explicit `canary.maxSurge` / `maxUnavailable` | **0** / **0** | — |
| `setWeight` step outside {0, 100} | **21** | — |
| `status.phase` | — | **Healthy on all 21** |

**Money-path is 17, not 15.** Cross-checked name by name against
`openbank-libs/governance/rules.yaml: money_path_services`. The issue *title* says 15; the issue
*body* lists 17 names under `money-path:`, so the body was right and the title compressed it.

```
money-path (17): account balance consent fraud fx ledger lending transaction sepa-payment
                 domestic-payment sepa-instant clearing settlement swift vop sanctions sca
not money-path (4): customer-edge kyc onboarding party
```

There is no unaffected control group. The canary strategy has never run in a shape where it can
do anything.

## The title compresses two mechanisms — they are not equally live

**Mechanism 1 — the weight cannot be expressed. Unconditional, biting today, all 21.**
No `trafficRouting` on any of the 21, so Argo realises a canary weight by *pod count*. With N
replicas only the multiples of 100/N exist; at N=1 that is 0 or 100. So `setWeight: 10` routes
everything or nothing, and the `pause: 5m` that follows is wall-clock delay observing a state the
weights do not describe. Evidence: all 21 carry `setWeight` 10 and 30 (13 of them also a terminal
100), and 0 carry a router. This does not depend on the cluster being full — it is true right now,
on an idle cluster, and it is true of every one of the 21.

Raising `replicas` to 2 does **not** fix this. It makes `setWeight: 10` mean 50%.

**Mechanism 2 — the surge pod has nowhere to schedule. Armed on all 21, not biting today.**
All 21 leave `canary.maxSurge` / `maxUnavailable` unset, so they take the Argo defaults —
`maxSurge: 25%` → 1, `maxUnavailable: 25%` → 0. A canary step therefore needs room for a **second**
pod before the stable pod may be released, exactly the #3545 deadlock. **But the honest live
reading is that the pool has headroom right now:**

```
$ kubectl get nodepool -o json
default   limits.cpu 72   status.resources.cpu 36   nodes 13     # 22 nodes cluster-wide
```

36 of 72 vCPU, so Karpenter can still provision and mechanism 2 is **latent, not active**. That
also updates the issue body, which cites a `48` live cap against `72` declared and calls the drift
documented — the live limit now reads 72. Mechanism 2 bit when #3545 was filed (`48/48, 23 nodes`)
and will bite again under load; it is not what is wrong today. **Mechanism 1 is.**

## The gate

`unrealisable_weights(doc)` — `weight * replicas % 100 != 0`, short-circuited by `trafficRouting`
(a router realises any weight at any pod count) and by kind/strategy. Findings get their own
baseline file so #3545's, which is empty on purpose, stays empty.

Both halves keep the same two-way contract: a **new** finding fails, and a baseline line that has
**stopped matching** fails too, so the file cannot outlive the debt it describes.

### Falsified both ways on the real tree — exit codes read directly, not through a pipe

| case | exit | counts |
|---|---|---|
| A. clean, baseline in place | **0** | 21 hits, **0 NEW**, 0 stale |
| B. new one-replica canary added (`setWeight: 20`) | **1** | 22 hits, **1 NEW**, 0 stale — named in the output |
| C. ghost line added to the baseline | **1** | 21 hits, 0 NEW, **1 stale** — named in the output |
| D. restored | **0** | 21 hits, 0 NEW, 0 stale |

Before baselining, the same run reported **21 NEW** — exactly the 21 derived independently above,
so the baseline was written from a gate that had been shown able to report them.

Through the real runner, enforced:

```
$ python3 .github/scripts/run-gates.py --only single-replica-rollout-strategy
PASS [gitops] one-replica workloads declare their rollout strategy AND their canary is realisable
selftest OK: 8 strategy-declaration + 13 canary-realisability cases, both directions.
check-single-replica-rollout-strategy: 0 ... 0 NEW vs baseline, 0 stale baseline line(s).
check-canary-rollout-realisable: 21 ... 0 NEW vs baseline, 0 stale baseline line(s).
```

The 13 new self-test cases include the ones that would let a wrong implementation pass: 2 replicas
still failing on 10/30, 2 replicas *passing* on 50, 4 replicas passing on 25/50/75, `trafficRouting`
short-circuiting, blueGreen and Deployment not being asked, and an absent `replicas` defaulting to 1.

## What this PR deliberately does NOT do

**It changes no replica count.** Three reasons, and the second is the one that would have made this
a correctness regression wearing a capacity fix's costume:

1. Doubling all 21 costs **+4.62 vCPU / +9.69 GiB**, ~**+9.6%** of the `default` nodepool's CPU cap.
   This repo has a recorded incident of a NodePool cap stalling exactly this class of change.
2. **#3467**: `openbank-lending-service` holds compliance-pack activation as per-pod **in-memory**
   state. Its whole reason for not being a live defect is that lending runs one replica — a second
   replica *arms* a four-eyes divergence. #3644 is the convergence fix and says itself that it does
   not make lending safe to scale. The same question is unanswered for the rest of the 21.
3. Replicas alone do not fix mechanism 1 anyway (see above).

**This is an owner decision, and the evidence for it is now in the baseline file.** Three ways off
the list, recorded there in full:
(i) make the weights match the replica count — free, and says out loud that progressive delivery is
not in force; (ii) wire a `trafficRouting` provider (ingress-nginx / SMI) — the only option that
makes `setWeight: 10` mean 10%; (iii) raise `replicas` — costs capacity and needs the per-pod-state
question answered per service first.

No `gitops/` manifest, `.rego` or `rules.yaml` key changed, so no OPA-bundle or pod-roll restamp is
required. Nothing was applied, scaled or rolled — all cluster reads were read-only.

`Refs #3806` rather than `Closes`: this lands the gate half. The issue's acceptance also asks for a
per-service replica/constraint decision and a PDB for anything reaching 2, which are the owner's
calls above.

Refs #3806, #3545, #3805, #3467, #3644

### DIFF
--- a/.github/canary-rollout-realisable-baseline.txt
+++ b/.github/canary-rollout-realisable-baseline.txt
@@ -1,0 +1,53 @@
+# Declared debt for the canary-realisability half of check-single-replica-rollout-strategy.py
+# (issue #3806, the sharper half of #3545).
+#
+# Every line here is a canary Argo Rollout that declares `setWeight` steps its replica count
+# cannot produce. Without a `trafficRouting` provider Argo realises a canary weight by POD COUNT,
+# so N replicas express only the multiples of 100/N — at N=1 that is 0 or 100 and nothing else.
+# `setWeight: 10` on a one-replica Rollout therefore routes all traffic or none, and the `pause`
+# after it observes a state the weights do not describe.
+#
+# Measured on origin/main and live (`kubectl get rollout -A`) 2026-08-08: 21 Rollouts in the
+# fleet, 21 at `replicas: 1`, 21 canary, 0 with trafficRouting, all reporting `Healthy` and
+# `1/1 READY`. 17 of the 21 are money-path per rules.yaml: money_path_services. There is no
+# unaffected control group — the canary strategy has never run in a shape where it can do
+# anything, which is why nothing ever looked wrong.
+#
+# This file is the ENFORCED ratchet, not an amnesty: a NEW unrealisable canary fails the gate,
+# and a line here that stops matching fails it too (delete the line when the debt is paid, or the
+# file outlives it and starts lying about what is covered).
+#
+# Three ways off this list, and picking between them is an owner decision, not a drive-by:
+#   1. Make the weights match the replica count (at 1 replica: 0/100 only) — free, honest, and
+#      says out loud that progressive delivery is not in force.
+#   2. Wire a trafficRouting provider (ingress-nginx / SMI) — weights become real at any replica
+#      count. This is the only option that makes `setWeight: 10` mean 10%.
+#   3. Raise `replicas` — note this alone does NOT fix it (at 2 replicas `setWeight: 10` means
+#      50%), it costs capacity (+4.62 vCPU / +9.69 GiB to double all 21, ~+9.6% of the `default`
+#      nodepool's declared CPU cap), and for at least lending-service it is a CORRECTNESS change
+#      first: #3467 records per-pod in-memory compliance-pack state, so a second replica arms a
+#      four-eyes divergence. #3644 is the convergence fix and says itself it does not make
+#      lending safe to scale.
+#
+# Refs #3806, #3545, #3805, #3467, #3644.
+openbank-infra/gitops/components/accounts/account-service.yaml: Rollout/account-service canary declares setWeight 10,30 that 1 replica(s) cannot express and has no trafficRouting
+openbank-infra/gitops/components/balances/balance-service.yaml: Rollout/balance-service canary declares setWeight 10,30 that 1 replica(s) cannot express and has no trafficRouting
+openbank-infra/gitops/components/consent/consent-service.yaml: Rollout/consent-service canary declares setWeight 10,30 that 1 replica(s) cannot express and has no trafficRouting
+openbank-infra/gitops/components/customer-edge/customer-edge.yaml: Rollout/customer-edge canary declares setWeight 10,30 that 1 replica(s) cannot express and has no trafficRouting
+openbank-infra/gitops/components/fraud-service/fraud-service.yaml: Rollout/fraud-service canary declares setWeight 10,30 that 1 replica(s) cannot express and has no trafficRouting
+openbank-infra/gitops/components/fx-service/fx-service.yaml: Rollout/fx-service canary declares setWeight 10,30 that 1 replica(s) cannot express and has no trafficRouting
+openbank-infra/gitops/components/kyc/kyc-service.yaml: Rollout/kyc-service canary declares setWeight 10,30 that 1 replica(s) cannot express and has no trafficRouting
+openbank-infra/gitops/components/ledger/ledger-service.yaml: Rollout/ledger-service canary declares setWeight 10,30 that 1 replica(s) cannot express and has no trafficRouting
+openbank-infra/gitops/components/lending/lending-service.yaml: Rollout/lending-service canary declares setWeight 10,30 that 1 replica(s) cannot express and has no trafficRouting
+openbank-infra/gitops/components/onboarding/onboarding-service.yaml: Rollout/onboarding-service canary declares setWeight 10,30 that 1 replica(s) cannot express and has no trafficRouting
+openbank-infra/gitops/components/party/party-service.yaml: Rollout/party-service canary declares setWeight 10,30 that 1 replica(s) cannot express and has no trafficRouting
+openbank-infra/gitops/components/payments/payments-services.yaml: Rollout/transaction-service canary declares setWeight 10,30 that 1 replica(s) cannot express and has no trafficRouting
+openbank-infra/gitops/components/payments/payments-services.yaml: Rollout/sepa-payment canary declares setWeight 10,30 that 1 replica(s) cannot express and has no trafficRouting
+openbank-infra/gitops/components/payments/payments-services.yaml: Rollout/domestic-payment canary declares setWeight 10,30 that 1 replica(s) cannot express and has no trafficRouting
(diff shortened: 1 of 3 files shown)


JiRaska:
Verified the numbers and the gate's actual invocation, not just the description — my first pass ran the script bare and read `advisory: no --enforce`, which would have been the wrong conclusion. The real invocation in `gates.yaml` passes `--enforce --baseline ... --canary-baseline ...`:

```
0 one-replica workload(s) inherit the default strategy, 0 NEW vs baseline, 0 stale
21 canary Rollout(s) declare a setWeight their replica count cannot express, 0 NEW vs baseline, 0 stale
```

`mode: enforced` in the entry, matched by a real `--enforce` in `run:` — not the "checker has no --enforce flag at all" shape (#4102/#4117) or the "mode says enforced but the script defaults to warn" shape this repo has shipped before. Both baselines cover today's 21 exactly, and the gate blocks any 22nd.

The retitled name — "declared AND realisable" — is the right framing. #3545's original gate answered "did the workload state its strategy", and every one of the 21 canary Rollouts passed that while declaring a weight schedule (`setWeight: 10, 30, ...`) that one replica structurally cannot produce (no `trafficRouting`, so a weight is pods-out-of-pods and one pod is 0-or-100). `status.phase` reads `Healthy` throughout, so nothing about the cluster state would ever surface this — it's a green control plane executing an unrealisable instruction.

Checked the money-path recount too: body said 17 by name, title compressed to 15 — cross-checked against `rules.yaml: money_path_services` and the 17 the PR lists match exactly (account, balance, consent, fraud, fx, ledger, lending, transaction, sepa-payment, domestic-payment, sepa-instant, clearing, settlement, swift, vop, sanctions, sca).

JiRaska:
Reviewed and verified by running, in a worktree at the PR head. **No blocking findings** — the change is sound and the baseline honours the two-way contract.

## It is stronger, not moved goalposts

The change is **purely additive**. `declares_strategy()` and `findings()` are byte-identical to `origin/main`; the PR only adds `unrealisable_weights()` + `canary_findings()` and a second reporting block.

Finding sets on the real tree, compared as sets rather than by summary line:

```
OLD predicate, PR tree     0 findings
NEW, declaration half      0 findings   — identical set, nothing dropped
NEW, canary half          21 findings   — all baselined
```

Zero dropped, 21 gained. "Declared vs realisable" is accurate.

## The baseline fails in both directions — tested, not assumed

21 non-comment lines, and enumerating `openbank-infra/gitops/` independently gives 21 Rollouts, all `replicas: 1`, all canary, none with `trafficRouting` — a 1:1 map, no phantom entries.

```
fake entry added            rc=1  "baseline line no longer matches anything — delete it"
real sca-service line cut   rc=1  reported as NEW debt
--canary-baseline omitted   all 21 become NEW and it fails (fails closed)
```

Scope is derived (`rglob("*.yaml")`), not hand-kept, so the "short list reads as full coverage" trap does not apply. Registration is right: `mode: enforced`, `selftest_expect: pass` for a self-test harness, and the tree passes today.

## One thing the title claims more than it delivers

**The original incident shape is still not flagged.** A `replicas: 1` canary Rollout whose only weight is `100` passes both halves:

```
setWeight: 10   (unrealisable)      rc=1   flagged
setWeight: 100  (incident shape)    rc=0   NOT flagged
clean tree                          rc=0
```

That is exactly the #3545 deadlock — one replica, canary strategy, stable scaled to 0 with nothing to fall back to. The docstring names it as a deliberate non-goal ("the real fix for those is `replicas >= 2`"), so it is not a regression. But "REALISABLE" reads as covering the rollout, when what became realisable-checked is the **weights**. Worth saying so in the gate name or the header, so nobody later reads a green here as "no canary can deadlock".

**A correction to my own measurement, because it changes how you read anything else I say about this gate.** My first probe reported the incident shape as unflagged *and* the known-positive as unflagged — both `0 errors`. The known-positive failing is what exposed it: without `--enforce` the script is advisory, so it prints a warning with no `::error` and returns 0 unconditionally. Every number above is from the enforced invocation, which is what `gates.yaml` actually runs.

## Notes

- **`:279-283`** — `_load_baseline` tests `ln.startswith("#")` on the unstripped line, so an indented comment would parse as a finding and then fail as stale. Cosmetic today; `ln.strip().startswith("#")` fixes it.
- **`:84`** — `WORKLOAD_KINDS` omits `StatefulSet` (1 in the tree). Pre-existing and untouched here, but it is the "reports the other kind as no-workload rather than unchecked" shape.
- **`:143`** — `*.yaml` only; no `*.yml` exists under `gitops/` today, so no blind spot, though nothing fails if one appears.
- ArgoCD's `ignoreDifferences: .spec.replicas` (KEDA scale-to-zero) means live replicas can be 0. Reading the git-declared value is the right subject for a manifest gate, and the baseline header's dated `kubectl` measurement corroborates the 21.

Issue #4026, opened 2026-08-07
Two workflows have never once succeeded (Image rescan 0/4, Product Catalog Native 0/39) and two controls have never been run at all (DR restore verify, Verify release evidence)


JiRaska:
Swept all 67 workflows for "has it ever succeeded" and "has it ever run". After ruling out the
false positives, **four are real: two that have never once gone green, and two controls that have
never been exercised at all.**

## Never green

```
workflow                          runs   conclusions
Product Catalog Native Build        39   failure 33, cancelled 6, success 0
Image rescan                         4   failure 4,  success 0
API fuzz (authenticated)             6   failure 6,  success 0        (already #3039)
```

- **`Product Catalog Native Build`** — 39 runs since it was added, zero successes. Latest run
  `31167008274` fails at **`Cold-start smoke test`** inside `Build native image + smoke test`. It is
  described as a "T1 enabler smoke test", i.e. the thing that would tell you the native build is
  viable; it has never told you that.
- **`Image rescan`** — 4 runs, all failed at **`Rescan image (VEX-aware)`**, and the failure is
  per-service across `account`, `agent`, `consent` … so it is systematic, not one bad image. This is
  the workflow that re-scans already-published images for newly-disclosed CVEs — the control that
  covers the window between "image built and scanned clean" and "a CVE is disclosed against it". It
  has never completed a rescan.

## Never run at all

```
DR restore verify          .github/workflows/dr-restore-verify.yml       workflow_dispatch only
Verify release evidence    .github/workflows/verify-release-evidence.yml workflow_dispatch only
Runner smoke test          .github/workflows/runner-smoke.yml            push: [ci/self-hosted-runner]
```

`DR restore verify` takes a `fiscal_year` and asserts the trial balance after a restore — the
disaster-recovery drill DORA expects and #669 tracks as missing operational evidence. It exists as a
file and has been dispatched zero times, so the drill has never been performed *and* the workflow that
would perform it is itself unproven.

`Verify release evidence` takes a release tag and verifies its evidence bundle. Also zero runs, so
every release's evidence bundle is unverified by the tool built to verify it.

`Runner smoke test` triggers on pushes to `ci/self-hosted-runner`, a branch that no longer exists.

## Ruled out — checked, and each is fine

Recording these so the next sweep does not re-file them:

- **`Dependabot auto-merge` (99 skipped / 1 null) and `Dependabot verification metadata` (100
[...]

Issue #669, opened 2026-07-09
Operational evidence umbrella (SPLIT): load benchmark + SLOs done; DR drill -> #4755, restore harness -> #4757


JiRaska:
> ## SPLIT AND CLOSED 2026-08-14 — this issue is an umbrella, not a unit of work
>
> Open 5 weeks with three deliverables in one title, two of them long since done. Rehomed:
>
> | scope item | state | owner |
> |---|---|---|
> | 1. money-path load benchmark | **DONE** (#734) | — nothing left |
> | 3. SLO definitions | **DONE and continuously enforced** (#729) | the `slo-registry-consistency` gate |
> | 2a. chaos/DR drill, executed | **not done**, owner-gated (one `kubectl delete pod`) | **#4755** |
> | 2b. automated restore harness wiring | **not done**, blocked on provisioning | **#4757** |
>
> Evidence for each is in the closing comment below. Do not reopen this issue — work the two above.

---

## What

The platform has strong *governance* evidence (CI gates, threat models, SSDLC) but zero *operational* evidence: no published throughput number for the ledger, no documented failure drill, no SLOs. Commercial cores sell operational history; a reference implementation must at least publish reproducible operational proof.

## Status (2026-07-25)

Scope items 1 and 3 are done. Item 2 (DR drill) is the only remaining open work — see below.

## Scope

1. ~~**Money-path load benchmark**: a reproducible load harness (candidate home: `openbank-simulation`) driving concurrent postings through ledger + transaction + balance services; published TPS/latency numbers with the exact setup (sandbox EKS node shape). Race-condition regressions found in the 2026-07 test-stack audit make contention behaviour worth measuring, not just peak TPS.~~ **DONE** — `perf/k6/money-path-write-benchmark.js` + versioned report `perf/reports/2026-07-10-money-path-write-benchmark.md`.
2. **Chaos/DR drill, documented** — still open. Kill a money-path pod mid-flow and a Postgres failover (CNPG) during posting; document observed behaviour, recovery time, and any invariant violations as a runbook under `docs/`.
[...]

Issue #253, opened 2026-07-05
Measure ADR-0083 T1 native cold-start SLO for product-catalog in sandbox


JiRaska:
### Context

ADR-0083 proposed piloting T1 (HTTP → 0 scale-to-zero) on `openbank-product-catalog`,
gated on a **measured** native-image cold-start SLO (target ~300-500ms p95) behind the
KEDA HTTP add-on in the sandbox cluster.

A prior change (#2800) declared `openbank-product-catalog: T1` in
`openbank-libs/governance/rules.yaml:finops_tiers.declared` and marked ADR-0083
`Delivery-Status: Shipped`, citing a non-existent issue `#2083` as evidence the gate had
passed. It had not: no commit in this repo ever added a native build for
`product-catalog` (verified via `git log --all -- openbank-product-catalog/Dockerfile`),
so the live `HTTPScaledObject` (`minReplicas: 0`, already deployed in
`openbank-infra/gitops/components/accounts/product-catalog-http-scaledobject.yaml`) has
been scaling a plain JVM image from zero — a several-hundred-ms-to-low-seconds cold
start, not the tens-of-milliseconds a native image would give it. The false claim has
been corrected (ADR-0083 `Delivery-Status` reverted to `Partial`, the `rules.yaml`
declaration reverted to unclassified) in the PR that opened this issue.

### What's done

- `openbank-product-catalog/Dockerfile.native` — Mandrel container build, native
  executable on `quay.io/quarkus/quarkus-micro-image` runtime.
- `.github/workflows/product-catalog-native-build.yml` — non-blocking CI lane that
  builds the native image and smoke-tests it (container starts, `/q/health/ready` goes
  green). Deliberately NOT on the `pull_request` trigger / JVM PR gate (ADR-0083 risk
  analysis).

### What's still needed (this issue)

1. Deploy the native image to the sandbox `accounts` namespace behind the existing
   `HTTPScaledObject`.
2. Measure p95 cold start (0 → first byte) through the KEDA HTTP add-on interceptor for
   48h per ADR-0083 §"Pilot plan" step 5.
3. If the SLO is met: re-declare `openbank-product-catalog: T1` in
   `rules.yaml:finops_tiers.declared` (citing this issue, not a stale/foreign number) and
   flip ADR-0083 `Delivery-Status` back to `Shipped`.
4. If not met: keep `minReplicas: 1` as the documented conservative fallback and record
   the miss in the ADR.

[...]

PR #4247, opened 2026-08-08
ci(github): make four never-succeeding controls tell the truth

JiRaska:
## Summary

Re-measured all four claims in #4026 by workflow **id** (not name — `Product Catalog Native Build` is registered as `Product Catalog Native Build (T1 enabler smoke test)`, so a name-keyed query reads as "gone"). All four still held. They turned out to be four *different* problems, and two of the issue's own premises were wrong.

| workflow | id | runs | conclusions | case |
|---|---|---|---|---|
| Product Catalog Native Build | 307547787 | 39 | failure 33, cancelled 6 | **(b)** defect in the workflow |
| Image rescan | 308224371 | 4 | failure 4 | **(a)** real defect in the thing tested — workflow correct |
| DR restore verify | 308235042 | 0 | — | **(c)** cannot work here, *and* reports green |
| Verify release evidence | 304790303 | 0 | — | no defect — dispatch-only, never dispatched |

## Product Catalog Native Build — case (b), fixed

The header said `EXPECTED RED ... the native image 500s on every request once running`. It never got as far as serving a request. Three defects, all in the workflow:

1. **No database.** product-catalog runs `quarkus.flyway.migrate-at-start: true`; the smoke test started the container with no Postgres, Flyway hit `Connection refused` on `localhost:5432`, and boot aborted. Added a `postgres:17-alpine` service container whose credentials match the `application.yaml` defaults.
2. **No OIDC opt-out.** Discovery against `localhost:8080` with no Keycloak aborts boot; the `%test` profile that disables it does not apply to this prod-profile binary. Added `QUARKUS_OIDC_TENANT_ENABLED=false` and `--network host`.
3. **Wrong port.** Readiness is on the management port **8085** (`quarkus.management.port`, root-path `/q`). The poll targeted `8104/q/health/ready`, which could never have returned 200 even against a perfectly healthy app.

The stale "expected red" note is removed rather than updated. A lane that is *allowed* to be red tells you nothing on the day it goes red for a new reason — which is exactly what happened here for 39 runs.

## Image rescan — case (a), workflow was right all along

The issue read "failed 4 times across `account`, `agent`, `consent` — systematic, not one bad image" as evidence the workflow was broken. It is the opposite: the scan **completed** and reported real, correctly-formatted findings. Systematic because every service shares one base image:

```
ghcr.io/jiraska/openbank-account-service:latest (alpine 3.23.5)
Total: 3 (HIGH: 3, CRITICAL: 0)
  libexpat       CVE-2026-56408  HIGH  fixed  2.8.1-r0   -> 2.8.2-r0
  p11-kit        CVE-2026-2100   HIGH  fixed  0.25.5-r2  -> 0.26.2-r0
  p11-kit-trust  CVE-2026-2100   HIGH  fixed  0.25.5-r2  -> 0.26.2-r0
```

So the control works and the finding is a true positive that sat unactioned for four weeks — because a scheduled workflow's red is addressed to nobody. `image-rescan.yml`'s own header claims it shares an "escalation pattern" with `vex-triage.yml`; `vex-triage.yml` opens an issue, and this one did not. Adds the `escalate` job + `.github/scripts/escalate-image-rescan.sh` (open-or-refresh on an exact title, prose in the script header per repo convention).

Deliberately **not** labelled `vex-triage`: `vex-triage.py` enumerates open issues by that label and auto-closes the ones whose CVE left its queue, which would hand this issue's lifecycle to a script that knows nothing about it.

**The CVE itself is not fixed here** — that is a base-image bump plus a fleet redeploy, or ~50 VEX verdicts. See "What remains".

## DR restore verify — case (c), and worse than never running

Dispatched it once ([run 31278708228](https://github.com/JiRaska/open-bank-oss/actions/runs/31278708228)) — its first run ever. Result:

```
JOB Restore ledger-db and verify integrity => success
   3 Gate on cluster access  => success
   4 Restore, verify, teardown => skipped
```

**It reports SUCCESS having restored nothing.** The gate `::notice::`d and set `run=false`, the restore step was skipped, and a skipped step leaves the job green. A DR drill that passes without restoring anything is worse than a red and worse than never running, because a green is what a compliance registry counts as evidence the control was exercised. It also cannot work today on two independent counts: it needs a deploy-pool self-hosted runner its own comment says does not exist yet, and its restore step is still a `WIRING INCOMPLETE` stub that hard-fails.

The gate now **fails loudly** with a message naming what did not happen. The workflow is dispatch-only and is not a required check, so a red costs nothing and says the true thing. The dead `if: steps.gate.outputs.run == 'true'` guard is removed too — leaving it would silently reintroduce the skipped-but-green shape.

## Verify release evidence — no defect, it had simply never been run

Dispatched it against a real tag ([run 31278713964](https://github.com/JiRaska/open-bank-oss/actions/runs/31278713964), `notification-service-v0.18.1`). It works:

```
Verified OK -> OK signature: notification-service-v0.18.1.cdx.json
Verified OK -> OK signature: notification-service-v0.18.1.evidence.json
Verified OK -> OK signature: notification-service-v0.18.1.slsa.json
Verified OK -> OK signature: notification-service-v0.18.1.vex.json
OK digest: ... .cdx.json / .slsa.json / .vex.json
[OK] evidence bundle for notification-service-v0.18.1 verified
```

4 cosign/KMS signatures and 3 manifest digests, all genuine — read from the log body *after* the last `##[endgroup]` and confirmed by substituted values, not the echoed script. The only reason it had never run is that `workflow_dispatch` was its sole trigger. A control whose only trigger is a human remembering is a control that does not run, so it gets a weekly cron that defaults to the newest release; an explicit tag still takes precedence. The tag is resolved in a step rather than job-level `env:` (there is no input on a schedule), and an unresolvable tag fails rather than 404ing into a "no signed evidence found" error that would read like a provenance defect.

## Also: `runner-smoke.yml` deleted

Triggered by `push: [ci/self-hosted-runner]` — `git ls-remote --heads origin ci/self-hosted-runner` returns nothing, so the branch cannot be pushed to. Nothing is lost: the `openbank-build` pool it smoke-tests is exercised continuously by `ci.yml`, `services-ci.yml`, `_service-ci.yml` and `auto-deploy.yml`. Its `workflow_dispatch` path existed and had never been used in the workflow's lifetime.

Grepping the prose separately (stale comments naming a deleted file are invisible to text guards) found one reference in `release-please.yml`, which cited `runner-smoke.yml` as the sole exception to "every `on: push` in this repo is filtered to `main`". That claim is now unconditional and the comment says so.

## Validation

- `actionlint` per file, and the **finding sets diffed against `origin/main`** — identical for all five touched workflows, so no new findings. The diff probe was validated against a known-positive (an injected `SC2086`) on a copy, so its silence means something.
- `run-gates.py --only workflow-run-step-size` → PASS, self-test green, largest run script 14568 chars (limit 19000).
- `shellcheck` clean on the new script; the dedupe query was validated against a known-positive (it returns `4026` for #4026's own title), so its empty result is a genuine "no such issue".
- **Validated against GitHub, not just a parser**: pushed head `eabb788` reports `runs_for_head_sha=0`. Zero runs is the accept signal — an unparseable workflow still produces a failed run titled after the file path.
- The fixed `product-catalog-native-build` and `dr-restore-verify` were dispatched against this branch; results in a comment below.

## What remains (why `Refs`, not `Closes`)

1. **The fleet-wide base-image CVE** the rescan correctly found: `libexpat` CVE-2026-56408 and `p11-kit` CVE-2026-2100, fixable, in every published `openbank-*-service:latest`. Needs a base bump + redeploy or VEX verdicts — its own PR. Worth noting the published images report `alpine 3.23.5` while `openbank-account-service/Dockerfile` declares `eclipse-temurin:25-jre`, so what is deployed and what is declared may not agree; that is worth confirming before bumping the wrong thing.
2. **The `escalate` job is unfalsified.** Proving it end-to-end means letting it open a real issue in this repo, which I did not do unprompted. Its syntax, its dedupe query and its create/refresh branch selection were validated read-only.
3. **DR restore verify still cannot run** — it needs the deploy-pool runner and the CNPG recovery manifests (`docs/bcp/automated-dr-restore.md`, #669). This PR only stops it lying about that.

Refs #4026


JiRaska:
## Falsification results — both fixed workflows dispatched against this branch

### Product Catalog Native Build — GREEN, the first success in 40 runs

[Run 31278980249](https://github.com/JiRaska/open-bank-oss/actions/runs/31278980249) on this branch:

```
2 Initialize containers                        => success
4 Build native image (Mandrel container build) => success
5 Cold-start smoke test                        => success
```

```
openbank-product-catalog 0.11.4 native (powered by Quarkus 3.38.0) started in 0.269s.
Listening on: http://0.0.0.0:8104. Management interface listening on http://0.0.0.0:8085.
Installed features: [agroal, cdi, config-yaml, flyway, hibernate-orm, hibernate-reactive, ...]

Cold start (container start -> /q/health/ready): 634 ms
```

**This also settles the ADR-0083 / #253 question the lane existed to answer.** The workflow header, the `Dockerfile.native` header and the ADR correction note all say the native binary "500s on EVERY request, including `/q/health/live`" and call it a suspected `openbank-libs` native-compatibility gap. It does not: with a database present it boots in 0.269s, installs 22 features including `hibernate-reactive`, `flyway` and `oidc`, and answers `/q/health/ready` 200. The "500s on every request" symptom is consistent with an app that never finished booting, which is what happens when Flyway cannot reach Postgres — i.e. the same missing database, observed locally.

So the T1 enabler smoke test has now told us the thing it was built to tell us, 39 runs late. What it still does **not** establish is the ADR-0083 promotion gate (p95 cold start behind the KEDA HTTP add-on in-cluster), which is unchanged and out of scope here. `Dockerfile.native`'s own KNOWN ISSUE header is now stale in the same way the workflow's was — worth a follow-up, but it is a doc claim, not a gate, so I left it rather than widen this PR.

### DR restore verify — now FAILS loudly instead of passing hollowly

Same workflow, same runner, same absent cluster — [before](https://github.com/JiRaska/open-bank-oss/actions/runs/31278708228) (on `main`) vs [after](https://github.com/JiRaska/open-bank-oss/actions/runs/31278983749) (this branch):

| | job | `Gate on cluster access` | `Restore, verify, teardown` |
|---|---|---|---|
| before | **success** | success | skipped |
| after | **failure** | failure | skipped |

The annotation on the failing run (read as the real `##[error]`, not the echoed script listing):

```
##[error]NOT A DR DRILL: no cluster access (kubectl/IRSA) on this runner, so nothing was
restored and nothing was verified. This workflow needs a deploy-pool self-hosted runner
that does not exist yet, and its restore step is still a WIRING-INCOMPLETE stub.
See docs/bcp/automated-dr-restore.md; do not record this run as DR evidence.
```

That is the whole point of the change: the identical input that used to produce a green now produces an accurate red naming what did not happen.

JiRaska:
The `dr-restore-verify.yml` half is the one worth reading closely — checked the actual diff, not just the description.

The fix removes the `if: steps.gate.outputs.run == 'true'` guard on the restore step entirely, with a comment explaining why re-adding any guard would reintroduce the exact shape being fixed. Before: no cluster access -> `::notice::` + `run=false` -> restore step **skipped** -> job **green**. A disaster-recovery drill that reports PASS having restored nothing is worse than either red or never-run, because it is what a compliance registry would count as evidence the control was exercised — the same "gate that has only ever passed is unfalsified" lesson this repo already names, taken to its worst case: not merely untested, actively lying.

The other three are correctly NOT treated as one fix applied four times:
- **Product Catalog Native Build** (case b) — three independent defects in the smoke test itself (no DB, no OIDC opt-out, wrong readiness port 8104 vs the real 8085), unrelated to the native build it claims to gate.
- **Image rescan** (case a) — the workflow was right; three real, currently-unfixed CVEs (libexpat, p11-kit, p11-kit-trust) sat unactioned for four weeks because a scheduled workflow's red has no reader. Deliberately not labelled `vex-triage`, so the existing auto-close sweep doesn't silently retire it once the CVE clears its NVD queue without anyone having patched anything.
- **Verify release evidence** — no defect, simply never dispatched.

Grouping four different causes under one issue and shipping one fix for all four would have been the wrong move; this PR earns the "four controls" framing by actually diagnosing four different things.

Not merging — workflow changes, human review by rule.

JiRaska:
Reviewed. **The premise holds for all four**, the fixes are real rather than rewording, and two of them are verified working by actual runs. One finding I would fix before merge, and it turned out worse than it first looked.

## The premise, measured against run history

| workflow | measured | |
|---|---|---|
| product-catalog-native-build | 23 failure, 6 cancelled, 1 success | the sole success is this PR's own dispatch |
| image-rescan | 4 failure, 0 success | |
| dr-restore-verify | 0 runs before this PR | |
| verify-release-evidence | 0 runs before this PR | |
| runner-smoke | **0 runs, ever** | `git ls-remote --heads origin ci/self-hosted-runner` is empty — its trigger branch does not exist |

## What each change actually does

Graded by whether it repairs, makes-loud, deletes, or merely rewords — the last dressed as the first being the thing to catch:

- **product-catalog — genuinely fixed.** Verified through `attempts/1/jobs` on run 31278980249: `Cold-start smoke test => success`. The lane went from 39 consecutive non-successes to a green where the smoke step actually **ran**. Not a vacuous green.
- **dr-restore-verify — now fails loudly.** On the branch, `Gate on cluster access => failure`. On `main` the same gate was `success` with the restore **skipped** — a green DR drill that restored nothing. Removing `id: gate` *and* the dead `if:` is the right pairing; dropping only the `id` would have recreated the skipped-but-green shape.
- **image-rescan** — workflow was already correct; adds a real escalation job.
- **verify-release-evidence** — adds a weekly cron plus tag resolution; the dispatch-only control genuinely could not run.
- **runner-smoke — deleted**, correctly.
- **release-please — comment-only.** Confirmed mechanically: stripping comment lines from the diff leaves zero changed lines. So no interaction with the Git-Data-API signing problem, and no `setFailed` guard added.

## Worth fixing: the new cron fails on its FIRST run, not eventually

`verify-release-evidence.yml:52-63` resolves `releases/latest` and guards only an *empty* tag. I checked what that endpoint returns right now:

```
releases/latest  ->  statement-service-v0.8.2   assets=0
```

So this is not a future hazard that materialises when a root aggregate release happens to be newest — **the very next scheduled run dies** at `:87` with `::error::no assets found on release statement-service-v0.8.2`. Which is precisely the misreading the step's own comment says it exists to prevent: an error that "would read like a provenance defect".

**And the obvious fix does not work.** Filtering to a component-pattern tag (something containing `-v`) would still land on `statement-service-v0.8.2`. Asset-less releases are not only the root aggregates — of the last 60, `statement-service-v0.8.2`, `dispute-service-v0.9.2` and `customer-edge-v0.57.1` are all component tags with zero assets, alongside `v0.16.0`/`v0.17.0`/`v0.18.0`/`v0.18.1`. The predicate that works is "newest release that actually carries `.sig` assets", or a clean skip when none does.

**Consistency, same file (`:11-13`).** The PR's central argument is that a scheduled workflow's red is addressed to nobody — that is the whole justification for the `escalate` job. It then puts a *new* workflow on a cron with no escalation path. Combined with the above, the most likely first outcome is a weekly red nobody sees. Either give it the same open-or-refresh escalation, or state in the header why it is exempt.

## Notes

- **`verify-release-evidence`** verifies one arbitrary component per week, so coverage is whichever service released most recently, not the fleet. Fine as a start — worth saying in the header so a green is not later read as fleet-wide provenance assurance.
- **`image-rescan.yml:135`** — `if: always() && needs.rescan.result == 'failure'` misses the case where `discover` fails and `rescan` is therefore `skipped` (result `skipped`, not `failure`): no scan, no signal, no issue. Narrow, but the same silent-nothing class this PR is fixing.
- **Escalation does not spam.** `escalate-image-rescan.sh` searches open issues by exact title and comments on the existing one rather than opening a new one per run. One issue, refreshed weekly. It is also the one piece here with **no execution evidence** — I did not run it either, since that would open a real issue. Logic and dedupe read correctly, shellcheck clean.

## Repo traps — checked, all clear

Step-size gate PASS (largest run script 14568 / 19000). `main-red-watch-declaration` PASS — deleting `runner-smoke.yml` strands no declaration. No new assertion greps a job log anywhere in the diff, so the "log contains its own `run:` script" trap does not arise. `escalate-image-rescan.sh` queries no jobs endpoint at all, so the attempt-scoping trap is N/A. No `curl -u` shape for gitleaks. None of the touched workflows is or becomes a required context. actionlint finding sets diffed against `origin/main` per file: identical.

PR #4248, opened 2026-08-08
fix(edge): declare the delegation upstream so the cluster stops dropping it

JiRaska:
Každé volání z customer-edge do delegation-service v sandboxu **odpadá timeoutem**:

```
upstream call to http://delegation-service.delegation.svc:8126/api/v1/delegations/grantee/... 
failed: java.net.http.HttpConnectTimeoutException: HTTP connect timed out
```

Není to výpadek — pod běží, URL je správně.

## Proč to bylo neviditelné

URL bylo správně **jako kódový default**, a přesně to celou věc schovalo.

`gen-network-policies.py` odvozuje ingress allow-list každé služby z env URL, které její volající **deklarují**. Volající, který se na upstream dostane přes zakompilovaný default, pro generátor neexistuje.

Allow-list delegation-service tedy pouštěl `admin-ui`, `observability` a `security-scanner`. VPC CNI agent běží ve standard mode (ADR-0060), takže customer-edge byl při každém pokusu **zahozen**.

## Tvar té chyby stojí za pojmenování

Může se zopakovat u libovolného upstreamu:

- **projde všemi testy** — nic v testu nemá NetworkPolicy
- **projde review** — URL je správně
- **selže jen v clusteru, a jako timeout, ne 403** — zahozený paket nemá kdo odpovědět

Deklarace env proměnné tu není konfigurace pro konfiguraci: **deklarace JE způsob, jak se volající stane pro generátor policy viditelným.**

## Dopad na zákazníka

Delegovaný přístup — sdílení v appce — četl prázdno pro každého zákazníka, protože vyhledání grantů na edge nikdy nedoběhlo.

`network-policies.yaml` je přegenerovaný, ne ručně editovaný (v hlavičce stojí DO NOT EDIT a myslí to vážně).


JiRaska:
Verified the root claim directly: `openbank-customer-edge/src/main/resources/application.yaml:121` is `${DELEGATION_SERVICE_URL:http://delegation-service.delegation.svc:8126}` — code default only, no env override declared anywhere in `openbank-infra/gitops/components/customer-edge/`. So the call worked in every unit/integration test (default resolves correctly) and was structurally invisible to `gen-network-policies.py`, which derives the ingress allow-list from DECLARED env URLs — an undeclared caller doesn't exist as far as the generator is concerned, so the NetworkPolicy never admitted it.

This is the sharper form of a lesson this repo already has for the read side (a corroborating host set built from the artifact it's about is circular) — here it's the write side: a value that's *correct* without being *declared* is invisible to any tool that derives policy from declarations, and "it works everywhere I tested it" is exactly the evidence that hides it, since every test path takes the default.

The NetworkPolicy addition is hand-written (no generator owns `delegation/network-policies.yaml`), so the manual edit is the correct mechanism here, not a workaround.

Not merging — money-path/high-risk paths, human review by rule.

JiRaska:
The fix itself is correct and complete. **It breaks an enforced gate, though, by leaving a stale baseline entry behind — one-line fix.**

## Blocking: stale `KNOWN_MISSING` entry

`.github/scripts/check-network-policy-code-edges.py:271-272`

```python
("customer-edge", "delegation-service", "delegation", 8126):
    "URL lives only in application.yaml + the resource's @ConfigProperty default.",
```

This PR fixes exactly that edge, so the entry no longer reproduces — and the gate is a two-way ratchet, so a fixed entry must leave the list. Measured on both sides:

```
origin/main   rc=0
pr4248        rc=1
::error::stale KNOWN_MISSING entry customer-edge -> delegation-service.delegation:8126
         no longer reproduces — delete it from check-network-policy-code-edges.py.
```

`mode: enforced`, group `gitops` (`gates.yaml:295-302`), so this is merge-blocking. Delete those two lines; nothing else.

Worth noticing that this is the gate working exactly as designed: it recorded the dropped edge as known debt, and now refuses to let the record outlive the defect.

## Verified correct, against actual manifest values

**The DNS name and port are real** — no `parties`/8100-class transposition here:

| in the diff | verified against | |
|---|---|---|
| namespace `delegation` | `delegation-service.yaml:14`, `:289` | matches |
| port `8126` | `:74` `containerPort: 8126`; `:299` Service `port: 8126` | matches |
| host `delegation-service` | Service name, same file | matches |

(8126 is also lending's port in a different namespace — coincidence, not transposition. Both correct.)

**The NetworkPolicy is genuinely derived, and the causal link was proved rather than assumed.** The file carries `openbank.io/derived: gen-network-policies`, so a hand-edit would be wrong. Running the generator on the PR head leaves `git status --porcelain` empty — byte-identical fleet-wide. Then the known-positive: deleting only the `DELEGATION_SERVICE_URL` env var and regenerating made the exact 3 added lines disappear. So the netpol delta derives from the env declaration and nothing else.

The unsorted-looking caller order is correct — the generator emits sorted callers excluding admin-ui, then appends admin-ui last.

**Direction is sufficient.** `customer-edge/network-policies.yaml` declares `policyTypes: [Ingress]` only, so there is no default-deny egress in the edge namespace and the ingress-only fix is complete.

**Both halves exist and the env var is actually read** — `application.yaml:121` matches the manifest name and default exactly, with three consumers (`DelegationGrants.kt:29`, `CustomerDelegationResource.kt:65`, `CustomerEdgeResource.kt:173`). Not a no-op declaration. All three pass `defaultValue`, so no `SRCFG00040` risk and `Optional<String>` is not needed.

## One consequence worth stating before this merges

**This makes #4196's problem observable.** Until now edge→delegation was dropped at the network layer, so OPA was never reached. After this lands, OPA becomes the effective gate — and #4196's three new actions (`delegation.reserve`, `.confirm`, `.release`) are absent from `edge-service-delegation` in `delegation_rest_ext.rego`, which I measured against `main`'s bundle as resolving DENY. So the symptom changes from a timeout to a clean 403. That is an improvement in diagnosability, not a new defect, but whoever debugs it should know the two are connected. The PRs are otherwise disjoint — this one touches no rego and owes no bundle regeneration.

## Note

The same gate run shows two other live DROPPED edges on `main`: `aml -> account-service.accounts:8100` and `customer-edge -> audit-service.audit:8113`. Both baselined with reasons, so known debt — but the audit one is the identical failure shape in the same workload, and a candidate for the same one-line treatment.

JiRaska:
Pushed the fix (`aa4a4e4cc`) — the two `KNOWN_MISSING` lines are gone.

```
before   rc=1   ::error::stale KNOWN_MISSING entry customer-edge ->
                delegation-service.delegation:8126 no longer reproduces
after    rc=0
```

The check that mattered is the second one, because "delete a baseline entry until the gate goes quiet" and "delete a baseline entry because the defect is fixed" look identical in a diff. So I removed a *different* entry — `customer-edge -> audit-service.audit:8113`, which is genuinely still missing — and confirmed it reproduces its `::error`. The gate is still looking; the baseline just shrank by one, which is the only direction it is allowed to move.

Worth recording a stumble from doing this, since it is a trap in exactly this kind of work: my first pass ran that known-positive test and then `git checkout`-ed the file to restore it — which silently reverted **my own fix** along with the probe's edit. The gate then passed for the wrong reason. Caught by grepping for the entry rather than trusting the restore. If a verification step mutates the file you are fixing, snapshot your version before running it, not after.

Everything else in the review stands: the DNS name and port match the real manifests (`delegation` namespace, `containerPort: 8126`, Service `port: 8126`), the NetworkPolicy is provably generator output rather than hand-edited, and the env var is genuinely read by three `@ConfigProperty` sites with defaults, so there is no `SRCFG00040` risk.

One thing to keep in view after this merges: it makes #4196's OPA gap **observable**. Until now edge→delegation was dropped at the network layer, so the policy was never consulted; afterwards `delegation.reserve` / `.confirm` / `.release` will fail as clean 403s instead of timeouts. Better diagnostics, same underlying gap — the rego still needs widening over there.

JiRaska:
Two of the three checks are the same real gap, not noise: this PR wires a genuinely new trust boundary (`customer-edge` -> `delegation-service`, previously invisible to the policy layer as the earlier comment established) and two gates correctly ask for it to be declared:

```
V9.1  customer-edge/customer-edge.yaml:178: plaintext http:// env value
      http://delegation-service.delegation.svc:8126   (1 NEW vs baseline)

security  threat-model updated on trust-boundary change (ADR-0030 D2, enforced)
```

Both are asking the same question two ways — a call that was structurally unreachable before this PR is now real, and the gates want it on the record (baseline entry for V9.1's in-cluster-plaintext debt, and a STRIDE row on `docs/threat-models/openbank-customer-edge.md` and/or `openbank-delegation-service.md` for the new caller). This is the correct kind of block: the fix genuinely creates the edge it's admitting to, so the paperwork is real, not incidental.

`Validate manifests` is the aggregate of these two, not a third independent failure.

JiRaska:
The `--admin` suggestion was a red herring — nothing prohibited the merge. `Validate manifests` had genuinely failed on my previous push, on **two enforced gates neither of us had seen yet**, and both were correct. Fixed in `7c3f7bcca`.

Worth noting `gates (gitops)` — the shard my baseline fix lives in — was green throughout. The failures were in `api` and `security`.

**`threat-model-diff` (ADR-0030 D2).** Admitting customer-edge into delegation's generated ingress policy is a trust-boundary change, so the threat model has to record it. Trust boundary 1 now says the boundary previously existed **in code and in no policy** — the hop was DROPPED rather than authenticated. Those are not the same control, and that difference is why the symptom read as a timeout instead of a 403.

**ASVS V9.1.** The rule is "no NEW plaintext inter-service edges until mTLS lands". This edge is not new: the call already ran over plaintext through the resource's `@ConfigProperty` default, where neither this gate nor `gen-network-policies.py` could see it. Declaring the URL makes an **existing** plaintext hop visible; it does not add one. Baselined with that reasoning beside the other 215 lines, retiring with the same mTLS work.

That is the pattern in all three gates this PR touched, and it is worth naming: making an invisible thing visible looks like adding it. The netpol baseline, the ASVS baseline and the threat model each reported the declaration as new work, when the traffic was flowing the whole time — undeclared, unpoliced, and unmeasured.

```
asvs-l3-mechanical-subset                        PASS  217 findings, 0 NEW, 0 paid-off
threat-model-updated-on-trust-boundary-change    PASS  self-test PASS
check-network-policy-code-edges                  rc=0
```

One thing to know if you reproduce this locally: the threat-model gate **refuses to run without `PR_DIFF_BASE`** ("refusing to run vacuously") rather than passing with an empty diff. It cost me a confusing `FAIL` that had nothing to do with the content. Good design — a gate that cannot see its input should say so, not report clean.

Scope is exactly five files, verified after the merge from `main`.

PR #4249, opened 2026-08-08
fix(ci): key the JAX-RS param gate to the top-level resource, not a nested type

JiRaska:
## What broke

`check-nonnull-jaxrs-params.py` reported a NEW occurrence on #4194 — a PR that does not touch a single JAX-RS parameter.

The gate's `enclosing_type` takes the LAST `class`/`interface`/`object` declaration before the offset, and `DECL_RE`'s leading `[\w@\s]*?` swallowed indentation. So a type declared **inside** a resource became the enclosing declaration for every handler below it.

That name is the middle field of the `BASELINE` key, so the effect is a silent re-keying:

```
demo|DemoResource|flag_after_object   ->  demo|Headers|flag_after_object
```

Measured on the real tree: a `private sealed interface DelegatedCardParse` holding a `data class Bad` re-keyed `listDisputes` from `openbank-customer-edge|CustomerEdgeResource|accountId` (baselined) to `openbank-customer-edge|Bad|accountId`.

Both directions are possible, and the quiet one is worse. A false NEW is at least loud. A handler re-keyed onto some *other* baselined key would simply be waved through — a real 500-on-absent-param shipping green.

## The fix

Restrict `DECL_RE` to column 0. The resource-plus-client case still works, because both are top-level and "last wins" resolves correctly; nested types become invisible to the key, which is what the key wants. Column 0 is a safe discriminator here since ktlint already enforces the indentation.

## Why the self-test could not catch it

It already had a nested `object Headers` above a `flag_after_object` handler — added deliberately, per its comment, to prove a nested declaration does not *disarm* the check. It doesn't. The handler was still flagged, so the fixture passed.

But the fixture only ever asserted **which params were flagged**, never the key. So it passed the whole time while emitting `demo|Headers|flag_after_object`. The bug was sitting inside the falsification fixture itself.

`SELF_TEST_EXPECTED_KEYS` now asserts the key, and the fixture gains a nested sealed interface.

## Falsified in three directions

```
new self-test vs OLD regex   FAIL — names both mis-keyed handlers, including
                             the pre-existing demo|Headers|flag_after_object
main tree, new script        OK — 20 baselined occurrences, 0 new
#4194 branch, new script     OK — 20 baselined, 0 new (its code unchanged)
```

The count staying at 20 on both trees is the part that matters: nothing was lost, only re-keyed correctly.

## Effect

Unblocks #4194, whose red was entirely this. No baseline entry is added or removed.

Refs #3104

### DIFF
--- a/.github/scripts/check-nonnull-jaxrs-params.py
+++ b/.github/scripts/check-nonnull-jaxrs-params.py
@@ -154,7 +154,22 @@ def strip_comments(src: str) -> str:
 ANNOTATION_RE = re.compile(
     r"@(" + "|".join(PARAM_ANNOTATIONS) + r")\s*\(\s*\"([^\"]+)\"\s*\)",
 )
-DECL_RE = re.compile(r"^(?:[\w@\s]*?)\b(interface|class|object)\s+(\w+)", re.M)
+# TOP-LEVEL declarations only — the leading `(?![ \t])` is load-bearing, not tidiness.
+#
+# The enclosing declaration supplies the middle field of the BASELINE key, so getting the NAME
+# wrong silently rekeys a handler. The old pattern let `[\w@\s]*?` swallow indentation, so a
+# nested type declared inside a resource became the enclosing declaration for every handler
+# BELOW it. Measured on openbank-customer-edge: a `private sealed interface DelegatedCardParse`
+# with a `data class Bad` member re-keyed `listDisputes` from
+# `openbank-customer-edge|CustomerEdgeResource|accountId` (baselined) to
+# `openbank-customer-edge|Bad|accountId`, which the gate then reported as a NEW occurrence on a
+# PR that had not touched a single JAX-RS parameter.
+#
+# Both directions are possible and the quiet one is worse: a false NEW is at least loud, while a
+# handler re-keyed onto some OTHER baselined key would be waved through. Restricting to column 0
+# keeps the resource-plus-client case working (both are top-level, so "last wins" still resolves
+# correctly) and makes nested types invisible, which is what the key wants.
+DECL_RE = re.compile(r"^(?![ \t])(?:[\w@]+[ \t]+)*?\b(interface|class|object)[ \t]+(\w+)", re.M)
 
 
 def enclosing_type(src: str, offset: int):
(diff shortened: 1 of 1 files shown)


JiRaska:
Cross-check with #4211, which touches the same file: **they compose, and are worth having both.**

```
pr4211 + pr4249   MERGE CLEAN — self-test 16/16, tree 7 baselined, 0 new, 0 stale
pr4211 + pr4194   MERGE CLEAN — gate passes even with the OLD regex
```

Different hunks: #4211 edits `BASELINE`, this edits `DECL_RE` and the self-test.

The second line is worth knowing before deciding urgency — **#4211 also clears #4194's red**, from the other end: it declares `listDisputes`' `accountId` nullable, so there is no finding left to mis-key. So this PR is no longer the only thing unblocking #4194. It remains the fix for the underlying bug, which recurs the next time anyone declares a nested type inside a resource — a normal thing to do, and today silently re-keys every handler below it.

JiRaska:
Ran the self-test rather than trusting the description.

```
--self-test    16/16 pass, including the new direct assertion:
                 "every finding is keyed to the top-level resource, not a nested type"
```

The bug class here is worth naming precisely: `enclosing_type` took the LAST class/interface/object declaration before the offset, so a nested type (e.g. `private sealed interface DelegatedCardParse` holding `data class Bad`) silently re-keyed a baselined finding onto a DIFFERENT key. That's a quiet failure mode, not a loud one — a false NEW at least shows up; a finding re-keyed onto some other baselined entry just vanishes from view while the baseline count stays the same.

**How the self-test missed it for as long as it did is the more useful lesson than the bug itself.** The fixture already had a nested `object Headers` above a handler, added deliberately to prove nesting doesn't break flagging — but it only ever asserted WHICH params get flagged, never the KEY they're flagged under. So it stayed green while silently emitting the wrong key the whole time. Same shape this repo has hit before with a whole-file grep matching prose instead of the construct: the assertion checked an adjacent, easier property instead of the one that actually mattered.

Fixed with `SELF_TEST_EXPECTED_KEYS` asserting the key directly, and `DECL_RE` restricted to column 0 (only a top-level declaration can be "the enclosing resource").

Falsified against the real defect it found: `#4194`'s `listDisputes` handler re-keyed from `openbank-customer-edge|CustomerEdgeResource|...` onto a different baselined key via a nested type — the new self-test fails against the OLD regex, naming that exact mis-keying.

PR #4275, opened 2026-08-09
fix(domestic-payment): never submit a payment to the clearing scheme twice (#4218)

### LINKED ISSUE #4218: domestic-payment: re-driving a VALIDATED payment re-submits it to the clearing scheme
## Summary

After #4200 (merged, `305b98c75`), re-driving a payment stranded in `VALIDATED` **submits it to the clearing scheme a second time**. Money path, and `DOMESTIC_SCHEME_SUBMISSION_ENABLED=true` in the deployed sandbox (`payments-services.yaml:827`), so it is live-reachable.

#4200 is a good fix for the problem it targeted and I am not suggesting a revert. It removed a guard that was accidentally load-bearing, and the consequence is one layer downstream.

## The chain, verified line by line on `main`

**1. `VALIDATED` is a routinely reachable stranded state.** `submitScheme` (`DomesticPaymentActivitiesImpl.kt:200-234`) wraps everything in a `try`, and the tail is:

```kotlin
} catch (ex: SchemeGatewayUnavailableException) {
    log.warnf(ex, "Scheme gateway unavailable for payment %s; holding in VALIDATED", paymentId)
    DomesticPaymentStatus.VALIDATED
} catch (ex: Exception) {
    log.warnf(ex, "Unexpected error during scheme submission for payment %s; holding", paymentId)
    DomesticPaymentStatus.VALIDATED
}
```

The generic catch covers the **whole** block — including `paymentRepository.update(...)` at `:219`, which runs *after* `schemeGatewayPort.submit(payment)` at `:201` has already succeeded. So a DB blip between the two leaves the payment in `VALIDATED` **with a live clearing item**, logged as a warning.

**2. Before #4200 the re-drive died before it could re-submit.** `canTransitionTo` (`DomesticPayment.kt:113`) reads `VALIDATED -> targetStatus in setOf(SENT_TO_CLEARING, …)` — `VALIDATED` is **not** in its own target set. So `validatePayment` threw *"Invalid domestic payment status transition"* and the workflow never reached `submitScheme`. Strictness that read as an obstacle was in fact the guard.

**3. After #4200 it reaches `submitScheme`, which lets it through.** The new early return at `:138` no-ops validate, and `submitScheme:199` is:

```kotlin
if (payment.status != DomesticPaymentStatus.VALIDATED) return@vtx payment.status
```

A payment **in** `VALIDATED` passes that test and proceeds to `schemeGatewayPort.submit(payment)` — a second clearing item for one payment.

**4. Nothing downstream dedups it.** `SchemeGatewayAdapter.submit` carries no idempotency key; `messageId = "DOM-<endToEndId>"` leaves dedup entirely to CERTIS / the simulator. (The *settlement* leg is protected — `SettlementAdapter` sends `domestic-settlement-<paymentId>` and treats 409 as settled — but that is a different leg, reached only after `SENT_TO_CLEARING`.)

Worth noting: #4200's own "Deliberately not done" section rejects exactly this outcome for the `SENT_TO_CLEARING` path ("`submitScheme` would submit a **second** item to the clearing scheme"). The same reasoning applies to `VALIDATED`; that path simply was not guarded by anything except the transition table.

## Confidence

**High on the mechanism** — every link above is read directly off `main`, not inferred. **Low on whether it has fired**: it needs an operator (or `ScreeningRedriveScheduler`) to re-drive a payment that is in `VALIDATED` *because* the update failed after a successful submit. No evidence it has happened; the log line would be a warning, not an error.

## Suggested fix, in order of value

1. **Make `submitScheme` recovery explicit rather than status-inferred.** Either query the scheme for an existing item by `endToEndId` before submitting, or persist a "submitted" marker *before* the gateway call, so the generic catch cannot leave a submitted payment looking unsubmitted. The status alone cannot distinguish "never submitted" from "submitted, bookkeeping failed" — and that distinction is the whole bug.
2. **Narrow the generic catch.** `catch (Exception)` around a block containing both the outbound call and the write is what merges those two states. At minimum, catch around the gateway call only.
3. **Add the missing compare-and-set.** `DomesticPaymentRepositoryImpl.update` (`:80-94`) is `find(...)` then dirty-checking — no `WHERE status = :expected`, and `DomesticPaymentEntity` has no `@Version` (checked all 7 migrations, no version column). Two concurrent workflows can therefore both read `RECEIVED`, both pass `:138`, and both write — emitting two `domestic.payment.status-changed` rows for one transition. Pre-existing and not caused by #4200; `ScreeningRedriveScheduler:133` does mint a distinct workflow id per attempt (`…-redrive-<epochMillis>`), so Temporal's id dedup does not apply.
4. **A concurrent-callers IT** in the shape of `OriginationConcurrentAdvanceIT` / `CompliancePackConcurrentDecideIT`. The service today has `DomesticPaymentBootSmokeIT`, `DomesticPaymentOutboxClaimIT`, `ScreeningRedriveSweepIT` and `PaymentConfirmationResourceIT` — none of which can express a race.

## On #4200's own test

`DomesticPaymentActivitiesImplTest.kt:217` asserts `coVerify(exactly = 0) { paymentRepository.update(...) }`, which is the right assertion for replay idempotency, and the author falsified it (removing the guard makes it fail). It is single-threaded mockk, so it cannot speak to any of the above — that is a scope limit, not a flaw in the test.

JiRaska:
## Summary

Re-driving a payment stranded in `VALIDATED` submitted it to the clearing scheme a **second time**,
creating two clearing items for one payment. Money path, and `DOMESTIC_SCHEME_SUBMISSION_ENABLED=true`
in the deployed sandbox, so it was live-reachable. Closes #4218.

## The defect

The rail could not tell **"never submitted"** from **"submitted, bookkeeping failed"**, because both
are spelled `status = VALIDATED`. `submitScheme` wrapped BOTH the outbound `pacs.008` and the
follow-up `paymentRepository.update` in one `try/catch`, so a database failure after a successful
submit was caught and logged as `holding in VALIDATED` — leaving a live clearing item behind a row
asserting nothing was sent. A re-drive read that row and submitted again.

Nothing downstream deduplicates it. Verified rather than assumed:

- `SchemeGatewayAdapter.submit` sends no idempotency key — only a deterministic
  `messageId = "DOM-<endToEndId>"`, which the receiver is free to ignore.
- `openbank-clearing-simulator` has **no deduplication of any kind** (searched its source for
  idempotency/duplicate/dedup handling — the only `messageId` uses are for building its own
  responses).
- The *settlement* leg IS protected — `SettlementAdapter` sends `domestic-settlement-<paymentId>`
  and treats 409 as already-booked — but that is a different leg, reached only after
  `SENT_TO_CLEARING`.

## The fix

The status cannot carry this information, so the dispatch is recorded on its own.

- **`scheme_dispatched_at` (V8)** — written BEFORE the gateway call, in its **own transaction**, so
  it outlives any failure of the work that follows. Deliberately not `submitted_at`, which
  `transitionTo` already sets on the move to `VALIDATED` and which is therefore non-null for every
  payment that can reach the scheme hop.
- **`submitScheme` refuses** to submit a payment that already carries the marker, and holds it for
  reconciliation. Held rather than progressed on purpose: we know a `pacs.008` went out, not what
  the scheme said about it.
- **The catch now covers the gateway call only.** A failure of the status write propagates, so
  Temporal retries the activity — and the marker makes that retry safe.
- **The marker is cleared only on proof** that the request never left this process
  (`ConnectException` / `UnknownHostException`), which keeps the ordinary "scheme is down" case
  re-drivable. A timeout does not qualify: the gateway may have accepted the message and answered
  too late.

### The trade this makes, stated plainly

An **ambiguous** failure now strands the payment in `VALIDATED` instead of retrying it. That is a
deliberate choice — for an outbound money instruction, a strand an operator can see beats a
duplicate payment nobody can recall — but it is a strand, it needs a human, and it is logged at
ERROR for that reason. V8 adds the partial index that finds them:

```sql
SELECT payment_id, scheme_dispatched_at FROM domestic_payments
 WHERE status = 'VALIDATED' AND scheme_dispatched_at IS NOT NULL;
```

## Deliberately not in scope

Issue item 3 — the missing compare-and-set on `DomesticPaymentRepositoryImpl.update`, and the
absent `@Version` on the entity. It is pre-existing, independent of this defect, and the issue says
so. Folding an aggregate-wide locking change into a money-path fix would make both harder to review.
Recorded in the threat model as a known residual.

## Verification

Both halves of the fix were falsified — each has a test that reddens without it, and only that test:

| Falsification | Result |
|---|---|
| Guard removed | `submitScheme refuses to re-submit a VALIDATED payment that was already dispatched` FAILED |
| Catch re-widened over the status write | `submitScheme lets a failure of the status write propagate instead of hiding it` FAILED |

Re-run after the detekt refactor rather than trusting the earlier proof, since that refactor touched
`submitScheme`.

Local gate: `detekt`, `ktlintCheck`, `koverVerify`, `build`, `quarkusBuild` — all green.
`quarkusBuild` forced with `--rerun-tasks`, because the first invocation reported every task
up-to-date and would have proven nothing about CDI wiring.

**128 tests, 0 failures, 0 skipped** (read from the JUnit XML, not the console).

CI gate groups run locally with `PR_DIFF_BASE=origin/main`: kotlin 19/19, data 11/11, registry
21/21, security 12/12. `api-contract-gate` fails locally because it wants `sudo` to install
`oasdiff`; it fails identically on a clean `origin/main` worktree, and this branch touches no
`openapi.yaml` at all, so there is no contract change to classify.

V8 re-checked against fresh `origin/main` immediately before pushing — latest committed migration is
V7, and no open PR touches this service's migrations.

## Review note

Money-path scope, so this needs 2 approvals per ADR-0030. Not self-merging.

Refs #4218


JiRaska:
**#4275 and #4300 are both money-path fixes to the same activity file, and they conflict — but only in the threat model.** Flagging on both so whoever merges second knows what to expect, since neither PR's CI can see the combination.

Measured rather than reasoned:

```
pr4275 + origin/main    clean
pr4275 + pr4300         CONFLICT — docs/threat-models/openbank-domestic-payment.md
                        DomesticPaymentActivitiesImpl.kt merges with 0 conflict markers
```

A clean merge is not the same claim as a correct one, so I checked that both fixes actually survive in the merged file rather than one silently winning:

```
#4300's settlement throw          present  (throw ex, :314)
#4275's dispatch-once guard       present  (schemeDispatchedAt, :211 / :223 / :237)
```

Both intact. The only real work is resolving the threat-model prose, where each PR documents a different failure mode and **both entries belong** — this is the shape where taking one side silently deletes the other's reasoning, and git reports nothing.

## They are complementary, and adjacent in a useful way

#4275 stops a payment being submitted to the clearing scheme twice (#4218). #4300 stops a settlement outage completing the workflow at `SENT_TO_CLEARING`.

Worth noting the connection, because it changes the order I would merge them in: my review of #4300 flagged that `submitScheme` **kept the identical defect one step earlier** — a blanket `catch (Exception)` returning `VALIDATED` — and #4275 is the PR that addresses that half. So they are not two unrelated fixes that happen to collide; they are the two ends of the same swallowed-failure pattern in one file.

Merge order does not matter technically. What matters is that the second one resolves the doc conflict by **keeping both** entries, and re-runs the module's tests on the merged result — `DomesticPaymentActivitiesImplTest` is edited by both, and that is exactly where a semantic collision would land without a conflict marker.

JiRaska:
Reviewed to money-path standard. The fix is well-argued and the migration comment is unusually good. **Two blocking findings, both of which let the duplicate payment through anyway**, plus one that strands payments permanently. I verified each against the code rather than reasoning from the design.

## Blocking 1 — the guard is read-then-write, not compare-and-set

`ActivitiesImpl.kt:197` reads via `findById`, `:207` tests `payment.schemeDispatchedAt != null`, `:230` writes unconditionally:

```kotlin
// DomesticPaymentRepositoryImpl.kt:102
update("schemeDispatchedAt = ?1 where paymentId = ?2", dispatchedAt, paymentId)
```

No `and schemeDispatchedAt is null`. The row count is discarded. `DomesticPaymentEntity` has **no `@Version`** (zero hits). And `V8` adds a plain nullable column whose index is **partial, not unique**.

So two concurrent `submitScheme` attempts both read `null`, both pass the guard, both write the marker, and both call `schemeGatewayPort.submit` — two clearing items, which is the exact defect the PR closes. Writing the marker *before* the gateway call is the right instinct; it just is not atomic.

Reachable via Temporal worker restart plus `maximumAttempts = 3` with no start-to-close timeout. (The redrive scheduler is not a second driver here — `findRedrivable` filters `status = RECEIVED` only, so a marker-stranded `VALIDATED` payment is not picked up by it.)

Fix is one predicate: `... where paymentId = ?2 and schemeDispatchedAt is null`, return the count, and abort the submit when it is 0.

## Blocking 2 — `@Retry` defeats the "provably never left" claim, and the KDoc says so

```kotlin
@CircuitBreaker(...)
@Retry(maxRetries = 2, delay = 300, jitter = 150)
@Timeout(5_000)
open suspend fun submitWithResilience(pacs008Xml: String)
```

`requestLeftThisProcess` inspects the cause chain of the **final** exception only. Concretely: attempt 1 sends the pacs.008 and times out at 5 s — the gateway has it. 300 ms later attempt 2 hits `ConnectException` because the gateway just died. The final chain contains `ConnectException` → `false` → `setSchemeDispatched(paymentId, null)` at `ActivitiesImpl.kt:243` → the re-drive submits a **second** pacs.008 for a payment the scheme already holds.

The KDoc reaches the right conclusion for the neighbouring case and does not apply it here:

> *a circuit-breaker rejection wrapping a timed-out attempt, is ambiguous: `@Retry` above may also have delivered an earlier attempt the gateway processed*

That reasoning is exactly as true of a final `ConnectException` after an earlier timeout. The claim needs to be **per-attempt** — a flag set inside the call before any byte is written — or `@Retry` has to come off the submit path. As written, the one verdict that authorises re-submission is derived from evidence that cannot support it.

## Worth fixing — an open circuit breaker strands every payment behind it, permanently

`CircuitBreakerOpenException` carries no `ConnectException`/`UnknownHostException` in its chain, so `requestLeftThisProcess` returns `true` → ambiguous → marker **kept**. But nothing was sent: the breaker short-circuits *before* any I/O.

During a scheme outage the first ~4 payments fail correctly and clearably; every payment in each subsequent 10 s open window is then marked, never sent, and permanently refused re-submission — each needing manual reconciliation against a scheme that has no record of it. A breaker rejection is pre-transmission by construction, the same class as `ConnectException`, and should be classified `false`.

## Tests: four of five falsify, one is vacuous

Restoring only `DomesticPaymentActivitiesImpl.kt` from `origin/main` and re-running gives `128 tests, 0 skipped, 4 failed` — the dispatch-guard, marker-before-gateway, clear-only-when-provably-never-left, and status-write-propagation cases all go red. Good.

The fifth, **`submitScheme keeps the dispatch marker when the failure is ambiguous`, passes against the old code**. It asserts only `coVerify(exactly = 0) { setSchemeDispatched(any(), null) }`, and the old code never calls that method at all — so it cannot fail. Pair it with `exactly = 1 { setSchemeDispatched(id, any<Instant>()) }`, the way the propagation test does.

I read the **skipped** count too (0 across 128), which also rules out a Flyway boot failure silently skipping every `@QuarkusTest`. `V8` is genuinely free: I scanned every remote head, not just `main`, and only this PR's branch carries one.

## Checked clean

No persist-vs-merge trap — `setSchemeDispatched` is a bulk HQL `update`, and `paymentId` is a real field rather than the `@Id`, so the where-clause resolves. No Vert.x issue — the new repo call runs inside `submitScheme`'s existing `vtx { }` bridge. Every test signature is `: Unit = runBlocking {`. The two other test files change by one line each (a fake gaining the widened port method), so no assertion was touched and no vacuity introduced.

Also worth knowing: the existing `DomesticPaymentStrandedGauge` already publishes `non_terminal_oldest_age_seconds` for `VALIDATED`, so a marker-strand **is** observable through an alert that already exists. This PR does not need new observability — but that gauge is what makes finding 3 visible rather than silent, so it is load-bearing for the trade-off the migration comment argues.

JiRaska:
Pushed both blocking fixes (`05462c051`).

**The guard is now atomic.** The predicate moved into the UPDATE and the row count is the verdict:

```kotlin
update("schemeDispatchedAt = ?1 where paymentId = ?2 and schemeDispatchedAt is null", ...) == 1
```

The port splits into `claimSchemeDispatch` / `clearSchemeDispatch` deliberately — one method carrying both would have to drop that predicate in order to release a claim, and releasing is exactly what must stay possible.

**Delivery is judged per attempt.** Ambiguity is recorded inside the call, once per attempt, and is sticky: the invocation is ambiguous if *any* attempt was. That closes the case where an accepted-but-timed-out first attempt was hidden behind a later `ConnectException` — the verdict that authorises re-submission. Your KDoc already made this argument for the circuit-breaker case; it just was not applied one line up.

**`CircuitBreakerOpenException` reclassified as pre-transmission.** The breaker rejects before any I/O, so nothing was sent. Treating it as ambiguous marked every payment behind an open breaker and then refused to re-send it — each one stranded for manual reconciliation against a scheme with no record of it.

## Falsified

Restoring only `DomesticPaymentActivitiesImpl.kt` from `origin/main` turns **six** tests red — including `submitScheme keeps the dispatch marker when the failure is ambiguous`, which previously passed against that same code because it asserted only that the marker was never cleared. It now asserts both halves, and a new test covers the lost race that a read-then-write guard cannot refuse.

```
baseline on the fix     129 tests, 0 skipped, 0 failures
old activity restored   6 FAILED
```

Skipped counts read from the JUnit XML rather than the console, since a suite that stops running reads as green.

## Two notes on the mechanics

`ktlintCheck` failed on the first run **before `test` ever executed**, so that run said nothing about the tests — worth knowing when reading a Gradle summary here.

Splitting the port method pushed both the interface and the impl to 11 functions, over detekt's `TooManyFunctions` threshold. I used the same `@Suppress` + reason the fleet uses elsewhere rather than merging the two methods back, since merging them is what created the bug.

PR #4293, opened 2026-08-09
ci(governance): warn before an attestation TTL expires, not after

JiRaska:
## The review decision, up front

This gate is `mode: enforced`, and on today's tree it **passes**. It will go **red on 2026-08-17**
unless somebody acts on `ledger.pentest` — measured, not predicted:

```
$ python3 .github/scripts/check-attestation-ttl.py --enforce --today 2026-08-16 ; echo $?
... 0 EXPIRED ...   0
$ python3 .github/scripts/check-attestation-ttl.py --enforce --today 2026-08-17 ; echo $?
                    1
```

That is the gate working, not a trap — but `Validate manifests` is a required check, so it blocks
every PR from that date. The fuse is 8 days and nothing re-attests automatically (by design: this
gate does not write). Three ways to land it, pick one at review:

1. **As-is (enforced).** The pressure is the point; both exits are cheap and honest — re-attest
   from a real lane run, or delete the stale entry and let C7 show its derived score.
2. **Drop `--enforce`** from the `run:` line in `gates.yaml`. Then it warns forever and never
   blocks — but this repo's own lore says an advisory finding decays into background noise, so
   this is the weakest option.
3. **Merge as-is and re-attest before the 17th.** `ledger.pentest` is `by: ci-zap-baseline`;
   `dast-zap-baseline.yml` runs weekly, so a fresh run before then makes the entry renewable from
   a real event.

I have not merged, and I have not touched `attestations.yaml`.

---

## The problem

`openbank-libs/governance/attestations.yaml` states an anti-rot contract in its own header:

> ANTI-ROT: každá atestace MÁ date + ttl_days. Po expiraci collector NEzapočítá bank-grade bonus
> a dimenze spadne zpět na derived skóre. Zelená není navždy.

**Half of it shipped.** `collect-prod-readiness.mjs:attestFresh()` implements the decay — an
attestation past its `ttl_days` stops counting and the dimension does fall back. The **warning**
half was implemented nowhere.

So an attestation lapsing is indistinguishable from one that was never made: a cell goes 3 → 2, the
matrix keeps rendering, and nothing anywhere says a control expired. Same failure class this repo
keeps finding — a control whose lapse looks exactly like calm.

This is live, not hypothetical. Two of the three entries carry `ttl_days: 21` *on purpose*, and the
file says why: *"TTL 21 dní je záměrně krátké: když lane umře, atestace vyprší a dimenze spadne
zpět na derived skóre — zelená přežije jen živou lane."* That design only works if somebody is
**told** when it fires.

## Measured state of the file today (2026-08-09)

| service | key | attested | ttl | expires | days left |
|---|---|---|---:|---|---:|
| ledger | `restore_drill` | 2026-07-26 | 180 | 2027-01-22 | **166** |
| ledger | `pentest` | 2026-07-26 | 21 | 2026-08-16 | **7** |
| consent-service | `pentest` | 2026-07-28 | 21 | 2026-08-18 | **9** |

**Three attestations, 0 expired, 0 malformed.** #2365's title says *"1/60"* and the body says the
file is *"currently empty on purpose"*; both predate the entries added on 2026-07-26/28. The gate
prints the count it measures on every run rather than restating the title.

**Nothing is expired today**, so there is nothing to baseline and no immediate reddening — the
enforcement note at the top is about the 8-day fuse, not about existing debt.

## What this adds

- `.github/scripts/check-attestation-ttl.py`
- one `gates.yaml` entry, `id: attestation-ttl-watch`, group `registry`, `mode: enforced`,
  with a `selftest` (`selftest_expect: pass` — it is the checker's own harness shape)

### It is not a generator

There is no `--fix`, no `--renew`, no write path of any kind. An attestation is a human or a named
CI lane testifying that an event happened; a script that writes one manufactures the same false
evidence as a DR drill reporting `success` having restored nothing (#4247). That also makes
enforcement non-deadlocking: deleting a stale entry is always available and is the *truthful*
action, since the score then shows what the repo can actually derive.

### Classification

| kind | condition | verdict |
|---|---|---|
| `expired` | `days_left < 0` | `::error`, exit 1 under `--enforce` |
| `expiring` | `0 ≤ days_left ≤ --lead-days` (default 14) | `::warning`, exit 0 |
| `valid` | beyond the window | silent |
| `malformed` | unusable `date` / `ttl_days` / shape | `::error`, exit 1 under `--enforce` |

Malformed is a **named finding**, never a crash and never a silent skip — because `attestFresh()`
degrades in *both* directions: a missing `date` reads as never-attested (invisible zero), a missing
`ttl_days` defaults to **365** (a year of unearned green). Both still parse and still render.

14 days is two lane cycles (`api-fuzz.yml` Tue+Thu, `dast-zap-baseline.yml` weekly) — a shorter
window could warn for the first time after the last chance to act had already passed.

### Dates are the trap

`date: 2026-07-26` unquoted is parsed by YAML into a **`datetime.date` object**, not a string, so
`.split('-')` / `strptime` / a regex all fail on the shape the file actually uses — and all three
real entries are unquoted. Both shapes are accepted, `datetime.datetime` is checked *before*
`datetime.date` (it is a subclass), and `--self-test` asserts the two shapes produce an **identical
verdict**, not merely that neither crashes.

### No hand-kept scope

Services and keys come from the parsed document — every top-level mapping, every key under it.
There is no list of expected services anywhere in the script. Corollary stated in the docstring
because it is the gate's one blind spot: it measures the attestations that **exist**, so it is
silent about the 57 missing from the 20 × 3 = 60 target. That is #2365's substance and is not
closable by code.

## Falsification

`--self-test` (12 fixtures, run by the runner *before* the gate):

expired · inside the lead window · comfortably valid · missing `ttl_days` · missing `date` ·
unparseable `date` · non-integer `ttl_days` · future `date` · entry not a mapping · **empty file** ·
file that is entirely comments · **the shipped commented-out example** (must count zero).

Each case asserts **both** the exit code and *which kind was assigned to which key* — an
exit-code-only harness would pass a gate that called everything malformed. Plus: the
object-vs-string date equivalence above at three TTLs, the `--enforce`/advisory split on identical
input, `--lead-days 0` vs `30` deciding `expiring` vs `valid` (so a hardcoded threshold hiding
behind the flag is caught), and a missing file as a named failure.

### Both ways on the REAL file — exit codes read directly, never through a pipe

```
$ python3 .github/scripts/check-attestation-ttl.py --enforce ; echo $?
::warning:: consent-service.pentest: expires in 9d (...) — re-attest from a real event, or delete
::warning:: ledger.pentest: expires in 7d (...) — re-attest from a real event, or delete
  ok  ledger.restore_drill: 166d left (...)
3 attestation(s): 1 valid, 2 expiring within 14d, 0 EXPIRED, 0 malformed (today 2026-08-09).
0
```

```
$ # real file + ONE fabricated expired entry appended
$ python3 .github/scripts/check-attestation-ttl.py --enforce --file /tmp/att-fab.yaml ; echo $?
::error:: sepa-payment.restore_drill: EXPIRED 190d ago (attested 2026-01-01, ttl 30d, expires 2026-01-31)
4 attestation(s): 1 valid, 2 expiring within 14d, 1 EXPIRED, 0 malformed.
1
```

```
$ # real file with `ttl_days: 180, ` stripped from ledger.restore_drill
$ python3 .github/scripts/check-attestation-ttl.py --enforce --file /tmp/att-nottl.yaml ; echo $?
::error:: ledger.restore_drill: missing `ttl_days` — attestFresh() would default it to 365, a year of unearned green
3 attestation(s): 0 valid, 2 expiring within 14d, 0 EXPIRED, 1 malformed.
1
```

The fabricated entry is **named**, and the counts move (3→4 total, 0→1 expired), not just the exit
code.

### Through the real runner

```
$ python3 .github/scripts/run-gates.py --only attestation-ttl-watch
PASS [registry] attestation TTL watch (Refs #2365, attestations.yaml anti-rot)  (0.3s)
1 gates  PASS=1
```

### The meta-gates were falsified against this script too

`check-gate-script-registration.py` was run with the `gates.yaml` entry temporarily removed, and it
named the new file — so its green here is coverage, not silence:

```
::warning file=.github/scripts/check-attestation-ttl.py::check-attestation-ttl.py is never run: ...
```

`gate-runner-self-test`, `gates-not-deregistered`, `advisory-gate-registration`, `yamllint` and
`check-license-headers.py` all pass with the entry restored.

## Noted in passing, deliberately not fixed here

`collect-prod-readiness.mjs:attestFresh()` computes elapsed days as
`(Δy)*365 + (Δm)*30 + Δd` rather than real date arithmetic. Harmless for the current entries
(all within one month) but it drifts from this gate's exact `date + timedelta(ttl_days)` on longer
spans — e.g. a 180-day `restore_drill` crossing several months. Out of scope for a read-only
watcher; worth its own issue if anyone wants the two to agree exactly.

Refs #2365

### DIFF
--- a/.github/gates/gates.yaml
+++ b/.github/gates/gates.yaml
@@ -823,6 +823,33 @@ gates:
       python3 .github/scripts/check-advisory-gate-registration.py --expect-manifest
 
   # ==== registry =======================================================
+  # Refs #2365. `openbank-libs/governance/attestations.yaml` is the only hand-entered maturity
+  # input, and its header states an anti-rot contract: every attestation carries `date` +
+  # `ttl_days`, and after expiry the collector stops counting the bank-grade bonus so the
+  # dimension falls back to its derived score.
+  #
+  # HALF of that contract shipped. `collect-prod-readiness.mjs:attestFresh()` does the decay;
+  # nothing ever did the WARNING. So an attestation lapsing is indistinguishable from one that
+  # was never made — a cell goes 3 -> 2, the matrix renders, no signal fires. A control whose
+  # lapse looks exactly like calm.
+  #
+  # This is live, not hypothetical: two of the three entries carry `ttl_days: 21` on purpose, so
+  # that a dead CI lane expires its own attestation. That design only works if someone is told.
+  #
+  # ENFORCED, and the gate NEVER WRITES — there is no --fix. An attestation is a testimony that
+  # an event happened; a script that renews one manufactures evidence (#4247). Enforcement is
+  # therefore never a deadlock: the two honest exits are to perform the event again and record
+  # it, or to delete the entry and let the score show its derived value.
+  - id: attestation-ttl-watch
+    name: "attestation TTL watch (Refs #2365, attestations.yaml anti-rot)"
+    group: registry
+    mode: enforced
+    selftest: |
+      python3 .github/scripts/check-attestation-ttl.py --self-test
+    selftest_expect: pass
+    run: |
+      python3 .github/scripts/check-attestation-ttl.py --enforce
+
   # issue #669 scope item 3 / rules.yaml: slo. Every money_path_services entry needs a
   # matching availability + latency Pyrra ServiceLevelObjective whose target/window match
   # the governed values in rules.yaml — pyyaml is already installed by the step above.
--- a/.github/gates/gates.yaml
+++ b/.github/gates/gates.yaml
@@ -840,10 +840,19 @@ gates:
   # an event happened; a script that renews one manufactures evidence (#4247). Enforcement is
   # therefore never a deadlock: the two honest exits are to perform the event again and record
   # it, or to delete the entry and let the score show its derived value.
+  # Lands ADVISORY on purpose, and the reason is a date, not caution. Measured by time-travel on
+  # the file as committed: `--today 2026-08-16` exits 0, `--today 2026-08-17` exits 1, because
+  # the two pentest attestations expire on 2026-08-16 and 2026-08-18. `Validate manifests` is a
+  # REQUIRED context, so an enforced gate here would block every PR in the repo eight days from
+  # now — and the honest remedy is to perform a pentest, which is not something a commit can do.
+  # An anti-rot gate that makes the whole repo unmergeable in order to announce a rotting
+  # attestation has traded a small problem for a larger one.
+  # GRADUATE TO enforced once both pentest entries carry a fresh date: the fuse is then 21 days
+  # out, the 14-day warning window gives real notice, and a red means someone ignored it.
   - id: attestation-ttl-watch
     name: "attestation TTL watch (Refs #2365, attestations.yaml anti-rot)"
     group: registry
-    mode: enforced
+    mode: advisory
     selftest: |
       python3 .github/scripts/check-attestation-ttl.py --self-test
     selftest_expect: pass
--- a/.github/scripts/check-attestation-ttl.py
+++ b/.github/scripts/check-attestation-ttl.py
@@ -1,0 +1,421 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) OpenBank contributors. Licensed under the Apache License, Version 2.0.
+"""check-attestation-ttl.py — the anti-rot half of `attestations.yaml` (Refs #2365).
+
+WHAT THE FILE PROMISES, AND WHO KEPT IT
+---------------------------------------
+`openbank-libs/governance/attestations.yaml` is the only place production-readiness maturity is
+entered by hand, and its own header states the anti-rot contract:
+
+    ANTI-ROT: kazda atestace MA date + ttl_days. Po expiraci collector NEZAPOCITA
+    bank-grade bonus a dimenze spadne zpet na derived skore. Zelena neni navzdy.
+
+The decay half is implemented — `collect-prod-readiness.mjs:attestFresh()` refuses an attestation
+older than its `ttl_days`, so the dimension does drop back to the derived score. The WARNING half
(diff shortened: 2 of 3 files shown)


JiRaska:
Closing as a duplicate of #3929, which is a strict superset. I built this without knowing #3929
existed, and neither PR mentions the other — the #2984/#3009 pattern, invisible to CI because both
are individually green over the same four-line file.

Compared by content rather than by diff:

| | #4293 (this) | #3929 |
|---|---|---|
| pre-expiry warning | 14 days | **30 days**, with a countdown |
| expired | error (after the advisory downgrade: warning) | warning |
| expired a long time | — | **ERROR after 30 days** — "not a lapse, stale bookkeeping" |
| malformed entry, bad or future date | named | named |
| **dead service key** | — | **named** — and fixes the live one |
| **unknown attestation key** | — | **named** (the collector reads exactly eight) |
| unresolvable `ref` | — | named |
| collector arithmetic | — | fixes the 365/30-day day-count |

The two things this PR had that #3929 covers better are the reason to prefer it outright: it warns
30 days ahead instead of 14, and it escalates a long-dead attestation to an error rather than
enforcing on the expiry date itself. That was the one design call I had to make here — this gate
went in `mode: enforced` and would have gone red on 2026-08-17 through a required context, so I
downgraded it to advisory with a graduation date. #3929 does not need that downgrade, because its
enforcement is keyed on staleness rather than on the expiry instant.

It also fixes something this PR could not see: `consent-service:` does not match the collector's
short-name lookup, so consent's pentest attestation has been scoring nothing. This gate would have
happily reported it as a valid, non-expiring entry.

Nothing is lost by closing. #3929 is currently `DIRTY`; I have merged `main` into it so it can be
reviewed.

Issue #4262, opened 2026-08-09
engagement-service: DISPUTE_OPENED adverse-state signal has no producer (last of ADR-0200 D6's four)


JiRaska:
## Source

Split off #2749 (closed on #4252's merge, ADR-0200 D6 adverse-state suppression). Of the four
`AdverseState` values (`FRAUD_HOLD`, `ARREARS`, `DISPUTE_OPENED`, `ERASURE_REQUESTED`), the first
three are now wired (arrears/erasure via #4106, fraud-hold via #4252) — `DISPUTE_OPENED` is the
one remaining gap, and closing the parent issue would otherwise have let it disappear silently.

## What's actually missing

`DisputeService.open()` and `.update()`
(`openbank-dispute-service/src/main/kotlin/com/openbank/dispute/.../DisputeService.kt`) persist a
`Dispute` and a local `DisputeTimelineEvent` ("OPENED", "STATUS_CHANGED") but write **no outbox
row** for either transition. `openbank.dispute.events` only carries the two event types
`resolve()` produces — `dispute.resolved` and `dispute.remediation_requested` — so there is
currently no `dispute.opened` (or any status-changed) event on the wire for a consumer to build
`AdverseState.DISPUTE_OPENED` from.

(Verified directly against dispute-service's source, not inferred from the topic/publisher
existing — see the #2749 thread's 2026-08-07 correction comment for how the first pass got this
wrong by checking only that the topic and publisher existed, not which event TYPES actually flow
through them.)

## What concretely must happen

1. Add a `dispute.opened` (and/or `dispute.status_changed`) outbox event to
   `DisputeService.open()`/`.update()`, same ADR-0050 transactional-outbox shape the existing
   `resolve()` writes already use.
2. A new Kafka consumer in `openbank-engagement-service` (mirroring
   `LendingArrearsEventConsumer`/`PartyErasureConsumer`/`FraudHoldEventConsumer`) that sets/clears
   `AdverseState.DISPUTE_OPENED` on `EligibilitySnapshot`.
3. Engagement-service's Kafka mTLS ACL grant for the new channel — can land in the SAME PR as (1)
   and (2) if dispute-service's `application.yaml` connector for the new channel ships in that PR
   too (no image/config ordering gap, same reasoning `check-msg-channel-image-parity.py` enforces
   — see #4252's own fraud-hold ACL for the worked example of landing both halves together).
4. Likely also needs the `check-event-contract-coverage.py` (ADR-0006) and
[...]

PR #4294, opened 2026-08-09
docs(engagement): correct a stale adverse-state note that blamed the wrong service

JiRaska:
## What

The comment above the `EligibilitySnapshot` construction in `ResolveSurfaceUseCase` made three
factual claims. All three are false on this sha.

| Claim in the note | Measured on `origin/main` |
|---|---|
| "neither signal is published as an event anywhere in this fleet today" | both are |
| "fraud-service has no persisted hold state" | 4 files matching `FraudHold` under `src/main` — `FraudHoldEntity`, `FraudHoldService`, … |
| "dispute-service emits only on resolution, never on open" | 4 occurrences of `dispute.opened` in `DisputeService.kt` (#4087) |

And the count it gave is wrong in this service too: engagement has consumed fraud holds since
#4252 (`FraudHoldEventConsumer`, channel `fraud-hold-events-in`), so **three** of the four
adverse states are materialised, not two.

## Why it matters more than a typo

The one real gap — `DISPUTE_OPENED` — is missing for a **different reason** than the note gave.
The producer exists; this service simply declares no `@Incoming` for the dispute topic. A note
that blames the wrong layer sends the next reader to fix a service that is already correct.

It also contradicted `Eligibility.kt`'s own KDoc, in the same service, which already lists
`FRAUD_HOLD` as materialised — so the two documents disagreed with each other and the reader had
no way to tell which was current.

## Corrected, not deleted

Per the repo's own rule: *"when you fix a defect whose comment predicted it, correct the comment
in place rather than deleting it, so the next reader learns the note was wrong rather than that it
was never there."* The new note quotes what the old one said and states what was measured against
it.

Comment-only change. `ktlintCheck` + `detekt` green on `:openbank-engagement-service`.

Refs #2749, #4262

### DIFF
--- a/openbank-engagement-service/src/main/kotlin/com/openbank/engagement/application/usecase/ResolveSurfaceUseCase.kt
+++ b/openbank-engagement-service/src/main/kotlin/com/openbank/engagement/application/usecase/ResolveSurfaceUseCase.kt
@@ -55,12 +55,23 @@ class ResolveSurfaceUseCase(
         val recent = events.recentForPartyAndSlot(partyId, slot, Instant.now().minus(DISMISSAL_LOOKBACK))
         if (DismissalRule.shouldSuppress(recent)) return Result.Suppressed
 
-        // Half-real (issue #2749): ARREARS (LendingArrearsEventConsumer, openbank.lending.events)
-        // and ERASURE_REQUESTED (PartyErasureConsumer, openbank.party.events) are materialised.
-        // FRAUD_HOLD and DISPUTE_OPENED are NOT — neither signal is published as an event
-        // anywhere in this fleet today (fraud-service has no persisted hold state; dispute-service
-        // emits only on resolution, never on open). That is still an honest gap for those two, not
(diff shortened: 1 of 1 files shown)


JiRaska:
Rewritten after #4297 opened, because the first version of this fix had the same defect it was fixing.

It replaced one enumeration with another — it listed which three adverse states were materialised and
named `DISPUTE_OPENED` as the gap. #4297 wires the dispute consumer, so whichever of the two PRs lands
second leaves this comment stale again. A note that keeps its own copy of a list that lives in the code
is the drift.

The comment now points at the `@Incoming` consumers and `AdverseState`'s KDoc rather than copying them,
and keeps only the part a rewrite would erase: the note that stood here was wrong on all three of its
factual claims, and the lesson that outlives the fix is that **a signal missing in this service is not
evidence the producer is missing**. It blamed the producing services; a reader acting on it would have
gone to fix two services that were already correct.

The two PRs touch disjoint files (`ResolveSurfaceUseCase.kt` here, `Eligibility.kt` + the new consumer
there), so they can land in either order.

Issue #3273, opened 2026-08-02
payments: no alert can see a payment stranded in a non-terminal state — 7 stuck since 2026-06-23 with every rule green


JiRaska:
### Affected service / module

openbank-infra (gitops/components/payments/prometheus-rules.yaml) + openbank-domestic-payment

### Version / commit

main @ 63b6cd8dd (observed live in sandbox, 2026-08-02)

### What happened?

Seven domestic payments are sitting in the non-terminal state `RECEIVED`. The oldest has been there
since **2026-06-23**. Nothing alerted, on any of them, for six weeks.

domestic-payment *is* covered by alerts — `prometheus-rules-tier1.yaml` has `PaymentServiceDown`,
`HighErrorRate`, `HighLatencyP99`, `PaymentSLOFastBurn` and `PaymentSLOSlowBurn`, all of which name
`domestic-payment` explicitly. Every one of them was green throughout, correctly:

| what the alert measures | what a stranded payment does |
|---|---|
| `up == 0` | pod is healthy |
| error-rate on spans | the API returned **2xx** — `RECEIVED` is a successful response |
| p99 latency | normal |
| SLO burn from `STATUS_CODE_ERROR` | no error span is ever emitted |

The payment is held **by design** (fail-closed screening), so nothing about it is an error. The
alerts are green about the wrong thing: they measure whether the service is *answering*, never
whether the work it accepted is *progressing*.

### The two rules that are actually missing

**1. A state-age alert.** transaction-service has `TransactionSagaStuck`
(`openbank_transaction_sagas_stuck_total > 0`); clearing has `ClearingSettlementWindowMissed`. There
is no domestic-payment equivalent — nothing counts payments aged in a non-terminal state. That is
the rule that would have fired on 2026-06-23 instead of the defect surfacing when a customer asked
where their money went.

**2. An outbox backlog alert.** `payments/prometheus-rules.yaml` has
`TransactionOutboxBacklogStuck`, `ClearingOutboxBacklogStuck` and `SwiftOutboxBacklogStuck` — and no
domestic one, despite `DomesticPaymentOutboxBacklogGauge` already publishing
`openbank.outbox.backlog` tagged `service="domestic"`. The metric exists; nothing reads it. That is
why 13 events reaching `DEAD` (see the Kafka ACL issue) went unnoticed as well.

The second is a two-line addition mirroring an existing rule:

```yaml
- alert: DomesticPaymentOutboxBacklogStuck
[...]

PR #4300, opened 2026-08-09
fix(domestic-payment): fail settlePayment on a settlement outage instead of completing the workflow on SENT_TO_CLEARING

### LINKED ISSUE #4182: A settlement outage completes the domestic-payment workflow on SENT_TO_CLEARING, so the retry policy never fires and the payment is stranded with no reader
## What happened

`transaction-service` was down for ~6 hours today (`CrashLoopBackOff`, 70+ restarts — root cause and fix in #4178). During the outage, payment `2691ef92-2ff5-49cc-a53b-dc016b089b93` (700 CZK, `INTERNAL_CLIENT`, submitted 11:35:03Z) was accepted, screened `CLEAR`, validated, submitted to the scheme — and then silently parked.

The Temporal workflow's own verdict:

```
Status          COMPLETED
Result          "SENT_TO_CLEARING"
```

**`COMPLETED` on a non-terminal business state.** No timer, no retry, no compensation, nothing to resume. The payment sits in `SENT_TO_CLEARING` with no ledger entry (`journal_entries` max `entry_number` still 280) until a human notices and hand-drives a redrive workflow. The customer sees a payment that left and never arrived.

## Why the activity's own retry never engaged

`DomesticPaymentActivitiesImpl.settlePayment` (`openbank-domestic-payment/src/main/kotlin/com/openbank/domestic/application/workflow/DomesticPaymentActivitiesImpl.kt:247`):

```kotlin
} catch (ex: SettlementUnavailableException) {
    log.warnf(ex, "Settlement unavailable for payment %s; holding in SENT_TO_CLEARING", paymentId)
    DomesticPaymentStatus.SENT_TO_CLEARING
} catch (ex: Exception) {
    log.warnf(ex, "Unexpected error during settlement for payment %s; holding in SENT_TO_CLEARING", paymentId)
    DomesticPaymentStatus.SENT_TO_CLEARING
}
```

The catch converts a downstream outage into a **successful activity return**. Temporal's activity retry policy is driven by activity *failure*; an activity that returns normally is a success, so the configured retries are unreachable on precisely the fault they exist for. `DomesticPaymentWorkflowImpl.process` then returns that status and the workflow completes.

The second `catch (ex: Exception)` is broader still — any settlement bug, not just unavailability, is absorbed into the same "held" state.

## Why nothing raised it

- The only evidence anywhere is one WARN line in the owning service's log.
- The payment is a row that stops changing. Nothing alerts on a row not changing.
- The workflow list shows `Completed` — the natural place to look for stuck work reports health.
- `POST /api/v1/domestic-payments` returned 2xx. No error rate, no latency signal, healthy pod: every tier-1 rule stayed correctly green, same blind spot as #3273 (rules measure whether the service *answers*, not whether accepted work *progresses*).

## The general defect

A downstream outage should either fail the activity (so Temporal retries with backoff, and the workflow stays running and visible as such) or park the payment on an explicit resumable state with a timer that re-attempts. What it must not do is complete the workflow. Today the only recovery is a human starting a `-redrive` workflow by hand — the same manual step used on 2026-08-02 for a different root cause (#3264 chain), which is a strong signal this is the standing failure mode rather than an incident detail.

## Suggested direction (not prescriptive)

- Let `SettlementUnavailableException` propagate so the activity fails and Temporal's retry policy applies; keep the DB status write as the durable record.
- If the retries are exhausted, do not complete — hold the workflow on a timer, or transition to an explicit `SETTLEMENT_PENDING`-style state that a sweeper owns, so "accepted but not progressing" is a state with a reader.
- Narrow the blanket `catch (ex: Exception)`; an unexpected settlement bug should not be indistinguishable from a planned degradation.
- Alert on age-in-status, not on error rate. `DomesticPaymentStrandedGauge` (#3273) already exports per-status count and oldest-age; today's payment would have been visible in it within minutes.

Same class as #3273. Related: #4178 (the outage that surfaced it), #3267 (a resilience wrapper discarding the fault it was meant to surface).

JiRaska:
## What this fixes

`DomesticPaymentActivitiesImpl.settlePayment` ended with

```
catch (ex: SettlementUnavailableException) { … SENT_TO_CLEARING }
catch (ex: Exception)                      { … SENT_TO_CLEARING }
```

Temporal drives an activity's retry policy off activity **failure**. An activity that returns
normally is a success — so the configured policy (3 attempts, 10 min schedule-to-close) was
structurally unreachable on precisely the fault it exists for, `DomesticPaymentWorkflowImpl.process`
returned that status, and the workflow **completed on a non-terminal business state**. A payment
that never settled was recorded as having been handed to the clearing scheme, with no timer, no
retry and no reader. The only evidence anywhere was one WARN line, and nothing alerts on a row that
stops changing.

The second catch is broader still: any settlement *bug* was absorbed into the same "held" state,
indistinguishable from a planned degradation.

Closes #4182

## The retry-vs-fail decision, and the evidence for it

**Decision: fail the activity and let the workflow retry.** Not a default — the failure direction is
the dangerous one *here*, and that depends entirely on whether the downstream call is idempotent. It
is, and this is read out of the code rather than assumed:

| Claim | Where it is established |
|---|---|
| The idempotency key is payment-scoped and stable across attempts — no timestamp, no nonce, so every retry carries the first attempt's key | `SettlementAdapter.settleWithResilience`: `idempotencyKey = "domestic-settlement-${payment.id}"` |
| A duplicate is a **success**, not a fault: transaction-service answers 409 and the adapter maps it to `SettlementOutcome(settled = true)` | `SettlementAdapter`, `HTTP_CONFLICT` branch |
| The activity is re-entrant: once the status write lands, a retry (or an operator re-drive) returns `SETTLED` without touching the port | the pre-existing `if (payment.status != SENT_TO_CLEARING) return payment.status` guard |
| Both properties are already covered by tests, so they are not resting on my reading alone | `SettlementAdapterTest`: `settle returns settled=true with null transactionId on HTTP 409 (idempotent)` and `settle sends correct idempotencyKey and rail` (asserts the literal `domestic-settlement-<id>`) |

So a retry costs at worst a 409. Swallowing costs a customer a payment that left and never arrived.

Two further pieces of evidence that failure was the intended design all along and only the code
disagreed:

- `SettlementUnavailableException`'s own KDoc already says *"the use-case holds the payment in
  SENT_TO_CLEARING and retries via the Temporal activity retry policy"* — the doc described the
  behaviour the catch made impossible.
- `validatePayment`'s comment, added by #4200, already calls `settlePayment` *"the step that would
  actually recover it, and which is idempotent (payment-scoped key, 409 = already booked)"*.

**This is the opposite call from #4218/#4275 on the dispatch edge, deliberately.** There, no
downstream deduplication exists (no idempotency key on the `pacs.008`, and clearing-simulator does
not dedup at all), so *holding* is correct and retrying would double-submit. Here the dedup is real
and tested, so *retrying* is correct and holding strands a payment. Same file, opposite answer,
because the two downstreams have opposite properties.

## What changed

- `SettlementUnavailableException` is logged at **ERROR** (a payment that reaches here has left the
  bank and has not been booked) and **rethrown**.
- The blanket `catch (Exception)` is **removed**, along with the method's
  `@Suppress("TooGenericExceptionCaught")`. Any other fault — an invalid transition, a repository
  failure, an adapter bug — propagates with its own type, so Temporal records what actually broke.
- The status write is now outside any catch, so "booked but not recorded" fails and is retried
  instead of being reported as a success.
- **The happy path is untouched**: `settle` → `transitionTo(SETTLED)` → `update` → return `SETTLED`,
  in the same order, with the same outbox message. The pre-existing test
  `settlePayment transitions SENT_TO_CLEARING to SETTLED on success` is unmodified and still passes.

The workflow file is **not** changed: the retry window stays at 3 attempts / 10 min. See residual
risk below.

## Falsification — both runs

Two of the tests I replaced *asserted the defect*: they simulated a settlement outage and asserted
`settlePayment` **returned** `SENT_TO_CLEARING`. That is the swallow itself, so they were green
against the broken code and are the reason it survived. They are replaced, not amended.

**Run 1 — new tests against `origin/main`'s version of the method** (fix reverted with
`git checkout origin/main -- DomesticPaymentActivitiesImpl.kt`, tests left as mine):

```
DomesticPaymentActivitiesImplTest > settlePayment FAILS the activity when settlement is unavailable so Temporal retries() FAILED
    java.lang.AssertionError at DomesticPaymentActivitiesImplTest.kt:458
DomesticPaymentActivitiesImplTest > settlePayment fails rather than reporting SETTLED when the status write fails() FAILED
    java.lang.AssertionError at DomesticPaymentActivitiesImplTest.kt:493
DomesticPaymentActivitiesImplTest > settlePayment propagates an unexpected exception instead of absorbing it into a status() FAILED
    java.lang.AssertionError at DomesticPaymentActivitiesImplTest.kt:475

27 tests completed, 3 failed
BUILD FAILED
```

**Run 2 — the same tests against this branch**, read from the JUnit XML rather than the console:

```
com.openbank.domestic.application.workflow.DomesticPaymentActivitiesImplTest: tests=27 fail=0 err=0 skip=0
  PASS settlePayment FAILS the activity when settlement is unavailable so Temporal retries()
  PASS settlePayment fails rather than reporting SETTLED when the status write fails()
  PASS settlePayment propagates an unexpected exception instead of absorbing it into a status()
  PASS settlePayment is re-entrant on an already SETTLED payment and does not book twice()
  PASS settlePayment transitions SENT_TO_CLEARING to SETTLED on success()
  PASS settlePayment returns current status unchanged when not in SENT_TO_CLEARING()

module total: tests=125 failures/errors=0 skipped=0
```

The fourth new test (`is re-entrant on an already SETTLED payment`) passes in **both** directions by
design — it is not a falsifier, it is the standing guard on the property that makes the retry
decision safe, so that a future change which breaks re-entrancy turns this PR's reasoning red.

Skipped count is reported explicitly because a module that cannot boot reports its integration tests
as SKIPPED, and `0 failures` alone would not distinguish that from a pass.

## Local gate — the tasks that actually ran

```
> Task :openbank-domestic-payment:detekt
> Task :openbank-domestic-payment:ktlintMainSourceSetCheck
> Task :openbank-domestic-payment:ktlintTestSourceSetCheck
> Task :openbank-domestic-payment:ktlintCheck
> Task :openbank-domestic-payment:test          (27 tests in this class, 125 in the module, 0 failures, 0 skipped)
> Task :openbank-domestic-payment:quarkusBuild
BUILD SUCCESSFUL
```

`quarkusBuild` is included because ArC/CDI wiring failures surface only there. Run under an isolated
`GRADLE_USER_HOME`.

## Overlap with #4275

**No line overlap in the Kotlin.** #4275 rewrites `submitScheme` (the *dispatch* edge, #4218); this
PR changes only `settlePayment` (the *settlement* edge, #4182). Verified by diffing #4275 against
`origin/main` on that file — its hunks span the `submitScheme` body and add two top-level helpers
after the class; it does not touch `settlePayment`. Git should merge both cleanly whichever lands
first.

**One expected textual conflict:** both PRs prepend an entry to `## 6. Change log` in
`docs/threat-models/openbank-domestic-payment.md`. That is an adjacent-add at the same position, so
it should conflict *visibly* rather than silently dropping an entry. Whoever merges second keeps
both entries — they document different edges and neither supersedes the other. I chose a visible
conflict over an insertion elsewhere in the file precisely because a clean merge that quietly keeps
one of two neighbouring list entries is the failure mode worth avoiding here.

Also related: #4200 (merged) made `validatePayment` re-entrant and its body states it does not
address this; #4218 covers re-submission of a VALIDATED payment.

## Money-path requirements (ADR-0030)

`openbank-domestic-payment` is in `rules.yaml: money_path_services`, so this PR needs **2 approvals**
and carries a threat-model update. `docs/threat-models/openbank-domestic-payment.md` gains a change-
log entry covering the property at stake (STRIDE-D/-R), the mitigation, the idempotency evidence
that makes retrying safe, the residual risk, and the rollback. I cannot grant the approvals; this PR
is opened for review only and is not to be merged with any override flag.

## Residual risk, stated rather than fixed

The retry window is unchanged, so a multi-hour outage like the one in #4182 still exhausts 3
attempts in seconds and ends as a **failed** workflow with the payment in `SENT_TO_CLEARING`. That is
strictly better than today — failed is visible in the workflow list, re-drivable, and no longer
claims the payment progressed — but it is still a strand that needs a human. Issue #4182's second
suggestion (a resumable `SETTLEMENT_PENDING`-style state owned by a sweeper, or a workflow timer) is
deliberately **out of scope** here: it needs a new status, a Flyway migration and a sweeper owner,
and folding it into this PR would put a schema change on the same review as a control-flow fix.
`DomesticPaymentStrandedGauge` (#3273) already exports per-status count and oldest-age and is the
reader for it meanwhile — alerting on age-in-status, not on error rate, as #4182 notes.


JiRaska:
Reviewed to money-path standard. **The premise is real, the fix is correct, and the tests are genuine.** One blocking issue, and it is in the threat model rather than the code.

## The premise holds, verified on `origin/main`

```kotlin
// DomesticPaymentActivitiesImpl.kt:253-259
} catch (ex: SettlementUnavailableException) {
    log.warnf(ex, "Settlement unavailable for payment %s; holding in SENT_TO_CLEARING", paymentId)
    DomesticPaymentStatus.SENT_TO_CLEARING
} catch (ex: Exception) { ... DomesticPaymentStatus.SENT_TO_CLEARING }
```

The activity **returns**, so Temporal records success, `WorkflowImpl.kt:58` passes that value through, and `process` completes on a non-terminal state. The retry policy was structurally unreachable — not misconfigured, never consulted.

## Tests are falsifying, verified by running

PR as-is: `tests="27" skipped="0" failures="0"`. Then I restored only the production file from `origin/main`, keeping the PR's tests, and re-ran: `EXIT=1`, `failures="3"` — exactly the three new cases. Restored, tree clean.

I read the **skipped** count as well, not just failures, since a suite that silently stops running reads as green. All 27 use `{ }` bodies, so no `fun x() = runBlocking {}` silent-skip; `skipped="0"` confirms it empirically rather than by inspection.

The fourth new test (`is re-entrant…`) passes on both sides — it documents pre-existing behaviour, which is fine and honest.

## Blocking: the safety argument cites a mechanism that does not exist

The threat model says — explicitly flagged as *"established from the code and not assumed"* — that transaction-service answers a duplicate with HTTP 409. The new KDoc repeats it. I checked:

```kotlin
// TransactionResource.kt:173
return Response.created(URI.create("/api/v1/transactions/${tx.id}"))
```

**201 unconditionally**, on the duplicate path too. Deduplication is an early return of the existing transaction (`TransactionService.kt:76-77`), not a conflict. The only 409s in that service are optimistic-lock and state-transition mappers, unrelated. `SettlementAdapter.kt:110-116`'s `HTTP_CONFLICT` branch is **dead code**.

The *conclusion* survives — the 201 arm also yields `settled = true`, so a retry is safe — but the stated evidence is false, in a money-path threat model, marked as verified. This repo's own rule is that an explanation which retires the question is worse than none: the next person to ask "is the retry safe?" finds it already answered, by a fact that is not true. Fix the wording, and ideally delete the dead 409 branch so the two cannot drift again.

## Retry semantics are safe — bounded, and idempotent

`MAX_ATTEMPTS = 3`, initial 2s, backoff 2.0, `ScheduleToClose 10 min`. **Not unbounded.** The idempotency key `"domestic-settlement-${payment.id}"` is payment-scoped and stable, and provider-side dedup has no TTL (unique constraint in `V1__init_transactions.sql:28`). A retry cannot double-book.

State stays visible: the row remains `SENT_TO_CLEARING`, exhaustion ends the workflow **failed** in Temporal rather than completed, and `DomesticPaymentStrandedGauge` publishes age-in-status for that state. No DLQ, but there is a reader.

No Vert.x regression — the activity's existing `vtx` bridge is untouched and this PR adds no reactive call, only a `throw`. Worth knowing for future work on this file: the unit tests override `vtx` with `runBlocking`, so they say nothing about the real bridge in either direction.

## Worth fixing

- **The stale KDoc contradicting the new behaviour is still there.** `DomesticPaymentActivities.kt:22-26` reads *"stays SENT_TO_CLEARING if the service is unavailable (Temporal retries via the activity retry policy)"* — that sentence **is** the false belief this PR exists to kill, sitting on the interface. Same in `SettlementPort.kt:25-27`, and `:20` repeats the 409 claim.
- **`SettlementOutcome` is discarded** — `settlementPort.settle(payment)`'s result is unused, so a `settled = false` outcome would still mark `SETTLED`. No producer today, but it is a latent trap on lines this PR rewrote.
- **`submitScheme` keeps the identical defect one step earlier** (`:231-234`, blanket `catch (Exception)` → returns `VALIDATED`). Out of scope, same bug class, same file.

## Note: one path becomes newly reachable

`TransactionService.kt:76-77` early-returns an existing transaction *before* starting its own workflow. If a first attempt committed the row but died before posting, a retry gets that **PENDING** transaction back as 201, the adapter reports `settled = true`, and domestic-payment marks `SETTLED` for a payment never posted.

Not introduced here — but retries at this layer previously never happened, and now they do. Worth a follow-up rather than a block, since it needs a fix in transaction-service.

Narrower: the DB constraint is `(idempotency_key, booking_date)`, so two truly concurrent retries resolving to different booking dates across a cutoff boundary would hit no constraint. The app-level lookup is date-agnostic and covers the sequential case only.

JiRaska:
**#4275 and #4300 are both money-path fixes to the same activity file, and they conflict — but only in the threat model.** Flagging on both so whoever merges second knows what to expect, since neither PR's CI can see the combination.

Measured rather than reasoned:

```
pr4275 + origin/main    clean
pr4275 + pr4300         CONFLICT — docs/threat-models/openbank-domestic-payment.md
                        DomesticPaymentActivitiesImpl.kt merges with 0 conflict markers
```

A clean merge is not the same claim as a correct one, so I checked that both fixes actually survive in the merged file rather than one silently winning:

```
#4300's settlement throw          present  (throw ex, :314)
#4275's dispatch-once guard       present  (schemeDispatchedAt, :211 / :223 / :237)
```

Both intact. The only real work is resolving the threat-model prose, where each PR documents a different failure mode and **both entries belong** — this is the shape where taking one side silently deletes the other's reasoning, and git reports nothing.

## They are complementary, and adjacent in a useful way

#4275 stops a payment being submitted to the clearing scheme twice (#4218). #4300 stops a settlement outage completing the workflow at `SENT_TO_CLEARING`.

Worth noting the connection, because it changes the order I would merge them in: my review of #4300 flagged that `submitScheme` **kept the identical defect one step earlier** — a blanket `catch (Exception)` returning `VALIDATED` — and #4275 is the PR that addresses that half. So they are not two unrelated fixes that happen to collide; they are the two ends of the same swallowed-failure pattern in one file.

Merge order does not matter technically. What matters is that the second one resolves the doc conflict by **keeping both** entries, and re-runs the module's tests on the merged result — `DomesticPaymentActivitiesImplTest` is edited by both, and that is exactly where a semantic collision would land without a conflict marker.

JiRaska:
Pushed (`2278f098f`) — the false 409 claim is corrected in all four places.

To be clear about what changed and what did not: **the deduplication is real**, so your conclusion stands. It just has a different shape — transaction-service early-returns the existing transaction on a repeated idempotency key and answers **201 with it**, and the adapter maps that arm to `settled = true` as well. Only the stated mechanism was wrong.

I also recorded the caveat that follows from the early return, because it is the part the original wording hid: it returns the existing row **whatever its status**, so a first attempt that committed the transaction but died before posting hands a retry a `PENDING` transaction as 201, which the adapter reads as settled. That fix belongs in transaction-service — noted here so the safety argument is not read as wider than it is.

The `HTTP_CONFLICT` branch is kept rather than deleted: it costs nothing and would matter if that service ever adopts a conflict response. But it is now marked unreachable and its log line says so, so if it ever fires that is a signal something changed, not a confirmation that the docs were right.

No behaviour change.

PR #4303, opened 2026-08-09
ci(security): bump the fleet runtime base so Image rescan can go green, and watch the recipe that pins it

JiRaska:
## The failure, established before anything was changed

`Image rescan` (workflow id `308224371`) has 4 runs and 0 successes, last `2026-08-03T10:24:53Z`.
Step conclusions read from `gh api repos/JiRaska/open-bank-oss/actions/jobs/<id> --jq '.steps[]'`,
not from a log grep — all 41 matrix jobs of the latest run are identical:

```
Set up Trivy              => success
Check image exists on GHCR => success
Rescan image (VEX-aware)  => failure      <- every job, every run
```

The real error (log tail, after the last `##[endgroup]`, value-substituted so it cannot be the
echoed script):

```
ghcr.io/jiraska/openbank-account-service:latest (alpine 3.23.5)
Total: 3 (HIGH: 3, CRITICAL: 0)
  libexpat       CVE-2026-56408  HIGH  fixed  2.8.1-r0   -> 2.8.2-r0
  p11-kit        CVE-2026-2100   HIGH  fixed  0.25.5-r2  -> 0.26.2-r0
  p11-kit-trust  CVE-2026-2100
##[error]... undispositioned fixable CRITICAL/HIGH in a shipped image
```

**The workflow is not broken. It has been right four times in a row.** Identical across the fleet
because every image shares one base — the expected burst, not a defect in the scan.

## Judgement: repair the finding, not the workflow — and close the loop that let it rot

Retiring was considered and rejected: nothing else covers this window. `security.yml` scans the
source tree at build time; `dependency-review` diffs submitted graphs; neither can see a CVE
disclosed against an OS package inside an image that was already published. That control is only
this workflow.

The reason it stayed red for four weeks is structural, and it is not in `image-rescan.yml`:

* the fleet runtime base is pinned **by digest** in `.github/workflows/Dockerfile.deploy`
  (`@sha256:681c543d…`, a **2026-07-16** temurin build), so `GHCR publish` rebuilding weekly can
  never pick up an upstream base rebuild — the packages are frozen with the digest;
* and that file was **the only Dockerfile in the tree Dependabot did not watch**. The docker
  ecosystem watches `/openbank-*` and `/` — i.e. the 51 `openbank-*/Dockerfile` declarations that
  *nothing builds* (#3016) — and misses the one recipe `auto-deploy.yml`, `ghcr-publish.yml` and
  `build-push-service.sh` all copy verbatim into their build context.

So the only way the digest could ever move was by hand, and nobody's hand was on it.

### What this PR does

1. Bumps the pinned base to the current `eclipse-temurin:25-jre` index digest
   (`@sha256:f19dbf0a…`, the **2026-08-04** rebuild, same JDK `25.0.3_9`, same ubuntu 26.04) across
   all 56 files that declare it — `check-dockerfile-no-build-stage.py` rule 3 requires them
   byte-identical, so this is one change in 56 places, not 56 changes.
2. Adds `/.github/workflows` (and `/openbank-infra/docker/services`, same blind spot) to the docker
   Dependabot entry, so the next base CVE arrives as a `chore(docker-svc)` PR instead of as a red
   Monday nobody reads.
3. Corrects the comment in `escalate-image-rescan.sh` that predicted this PR, in place rather than
   deleting it (it said the base bump "needs its own PR" — this is it).

The escalation path itself already exists (#4247) and is untouched.

## Proving it can succeed AND can still fail

A workflow that has only ever failed is unfalsified in both directions. Measured locally with the
workflow's **exact** invocation — `trivy image … --scanners vuln --severity CRITICAL,HIGH
--ignore-unfixed --format table --exit-code 1`:

| subject | result |
|---|---|
| old pinned base `@681c543d` | **exit 1** — 5 fixable HIGH in `/usr/bin/pebble` (`golang.org/x/net` ×4, `stdlib`) |
| new pinned base `@f19dbf0a` | **exit 0** |
| `ghcr.io/jiraska/openbank-account-service:latest` as published today | **exit 1** — the 3 alpine HIGHs above |
| the same image's `/app` layers rebased onto the new pinned base | **exit 0** |

The last row is the one that matters: it is the real shipped application content on the new base, so
the green is not a green that skipped the work. It is also the reason the fix is a base bump and
nothing more — the published image's *only* non-zero result is `os-pkgs`; every jar target scans 0.

The failure direction is proven by the same command on the two red subjects, with the same flags,
in the same run — not by reasoning about what it would do.

Note the old base is red for **different** CVEs than the alpine ones (`pebble`, a Go binary in the
Ubuntu temurin image): #3618 changed the base on 2026-08-07, after the last rescan, so the next run
would have failed again on a finding nobody has seen yet. Without this PR the fix that was already
merged would not have made the workflow green.

No workflow YAML is modified by this PR, so there is nothing here for the "push it and see whether
GitHub parses it" oracle to answer — `image-rescan.yml` is byte-identical to `main`.

## What is not proven here, stated rather than assumed

* **Dependabot picking up `Dockerfile.deploy`.** The strongest available evidence is in-repo and
  real, not a reading of the docs: `d9bdbe530` — `chore(docker-svc): Bump eclipse-temurin in
  /openbank-security-scanner` — is a dependabot[bot] commit whose single changed file is
  `Dockerfile.prebuilt`, so the docker updater does match `Dockerfile.<suffix>` by name. What
  remains unproven is only this specific directory, and it settles itself on the next Monday run:
  either a `chore(docker-svc)` PR eventually touches `.github/workflows/Dockerfile.deploy`, or it
  never does and that is a finding. The digest bump in this PR clears the current finding either way.
* **Timing.** `GHCR publish` is cron `0 6 * * 1` and took 24 min on its last run; `Image rescan` is
  `30 6 * * 1`. The margin is 6 minutes of a schedule GitHub delays by hours anyway, so a rescan can
  legitimately scan last week's images. That is noise, not a wrong answer — the control's subject is
  *what is published* — but it is the likeliest reason a given Monday reads red after this lands.

Refs #4026


JiRaska:
Reviewed and verified by running. **No blocking defects.** Two worth-fixing notes, both about the "watch the recipe" half — which is the part that decides whether this is a one-off bump or a fix.

## The bump itself checks out

I queried Docker Hub rather than trusting the digest:

- `sha256:f19dbf0a…` is a valid OCI index for `library/eclipse-temurin` carrying **linux/amd64 and linux/arm64** (plus ppc64le, s390x, windows).
- It is **exactly what the `25-jre` tag resolves to today** — `docker-content-digest` on the tag matches byte for byte.
- Config `created` is **2026-08-04** vs the old digest's **2026-07-16**, so the 19-day-old-rebuild claim is measured, not asserted.

The digest bump is uniform: grepping the PR head for the old digest returns **zero** hits, and 56 files carry the new one.

**The 50 per-service Dockerfile edits are mandatory, not cosmetic.** `check-dockerfile-no-build-stage.py` rule 3 (#3354) requires each `FROM` to be byte-identical to `Dockerfile.deploy`, so bumping the deploy recipe alone would fail the enforced gate. Ran it: `OK, 54 Dockerfiles, runtime-only, all with EXPOSE`.

**The admin-ui topology parser still works.** I ran the real `imageFacts()` body against the PR's `openbank-ledger-service/Dockerfile`:

```
{ ok: true, runtimeBase: "eclipse-temurin:25-jre@sha256:f19dbf0a…", fastJar: true, zgc: true, nonRoot: true }
```

`ok: true`, so the hardcoded `eclipse-temurin:25-jre-alpine` fallback is **not** hit and the dossier will show the truth. `fastJar` still true — `COPY quarkus-app/` intact, no uber-jar regression. admin-ui's own Dockerfile is untouched and correctly on the `SELF_BUILT` exemption list, so no repeat of the sweep that broke its live deploy.

**No conflict with #4247**, which touched `escalate-image-rescan.sh` a few hours ago: merge onto `main` is clean and the diff of that file is **comment-only** (tense change plus two sentences). Gates on the merged result: `supplychain` 19 PASS, `workflow-run-step-size` PASS.

## Worth fixing: the watch will probably not catch the next drift either

Both new dependabot directories exist and do not overlap the other docker entry — that part is fine. The problem is the filter.

The new dirs join `groups: docker-svc-minor` with `update-types: [minor, patch]`. **This bump is digest-only** — the tag `25-jre` never changed — so there is no semver delta for that filter to match. Either the update falls out of the group and Dependabot opens **one PR per directory**, or it is filtered out entirely and no PR is ever opened.

The first outcome is worse than it sounds, because rule 3 demands all 56 `FROM` lines stay identical: a single-directory PR changes one of them and is therefore **red and unmergeable by construction**, every week, forever. This repo's history shows exactly that shape — #70/#71/#72 are all `Bump eclipse-temurin … in /openbank-<svc>`.

**And I checked the diagnosis rather than the config.** The premise is that `Dockerfile.deploy` was the only file not watched. But the `openbank-*` directories were **already** watched by this same entry, and Dependabot's most recent `eclipse-temurin` PR of any kind is **2026-07-01** — nothing for the 07-16 → 08-04 drift it was already positioned to see. So watching was demonstrably not sufficient for the files that *were* watched, and adding two more directories under the same filter inherits that.

Concretely: cover digest updates explicitly, and make the grouped multi-directory PR the guaranteed shape rather than the hoped-for one.

Second-order: the scope here is **hand-kept** (`directories:` is a literal list plus a glob) and no gate asserts every Dockerfile path in the tree appears in some dependabot entry — so the next new Dockerfile location re-opens the same hole silently.

## Notes

- **Timing dependency in the PR's own premise.** The rescan scans `ghcr.io/jiraska/<svc>:latest`, which only carries the new base once `ghcr-publish` reruns. Publish is `0 6 * * 1`, rescan `30 6 * * 1` — a **30-minute** window to rebuild and push ~50 multi-arch images under `concurrency: ghcr-publish` serialisation. If publish overruns, the rescan reads last week's `:latest`, goes red anyway, and that red will read as "the bump didn't work". Worth dispatching `ghcr-publish.yml` right after merge rather than waiting for Monday.
- **The rescan grades a different artifact than the one deployed.** GHCR images are built amd64+arm64 so trivy's amd64 default resolves and no `--platform` is needed here — but what runs in the cluster is the **arm64 ECR** image. For a base-image CVE the two share an index, so this PR's claim holds; for anything arch-specific the rescan is green about an artifact nobody runs. Worth one sentence in the workflow header.
- **Pre-existing, not this PR** (measured on both sides): `generate-cluster-topology.mjs:82`'s `FROM` regex has no `i` flag and matches the literal `FROM` inside the comment *"fails if the FROM above drifts…"*, so the dossier renders `multiStage: true`, `buildBase: "above"` for a single-stage file. `runtimeBase` is right either way, so the base image shown is correct — the "multi-stage build" label next to it is fiction. Own issue.

PR #4306, opened 2026-08-09
fix(transaction): move the terminal status write inside the payment workflow

### LINKED ISSUE #4238: A settled transaction can stay PENDING forever: the terminal write lives outside the durable workflow, and the idempotent replay returns a PENDING row as 201
## What happened

A live domestic payment settled correctly — money moved, ledger balanced, balances updated — and its transaction row is still `PENDING`, permanently. Transaction `0a84d2a2-e933-494b-b2cf-446cdc4e2d4a` (`TXN17862033193816031`, 700 CZK, settled 2026-08-08 15:35).

Verified state:

| Evidence | Value |
|---|---|
| `transactions` row | `status=PENDING`, `version=0`, `completed_at=NULL`, no `failure_reason` |
| `transaction_outbox` | only `openbank.transactions.transaction.initiated`; no `...completed` |
| Temporal `payment-0a84d2a2-…` | `COMPLETED`, result `"COMPLETED"`, RunTime 6.91s |
| Ledger | entry 281, both legs 700 CZK, balanced |
| Balances | both accounts written at 15:35:30 / 15:35:32 |
| Caller | got HTTP 201 with the transaction id at 15:35:25.649 |

`version = 0` plus the missing `completed` outbox event prove the finalisation write never committed. The caller was told the settlement was booked anyway.

It is the only `PENDING` row in the table — 89 `COMPLETED` TRANSFER, 43 CREDIT, 16 DEBIT, 8 FAILED. So this is not the normal shape; something specific happened.

## Mechanism

Two halves, either of which alone is survivable.

**1. Durable orchestration, non-durable finalisation.** `TransactionService.initiateTransaction` saves the row `PENDING`, blocks on the Temporal workflow, and only then writes the terminal status (`TransactionService.kt:171-180`):

```kotlin
val state: SagaState = withContext(Dispatchers.IO) { stub.execute(saved.id) }
return if (state == SagaState.COMPLETED) {
    val completed = saved.complete(clock)
    transactionRepository.update(transaction = completed, outboxMessage = /* ...completed */)
```

`PaymentWorkflowImpl.execute` never writes the transaction status — it places a hold and posts the journal, then returns `SagaState.COMPLETED`. So the money-moving part is durable and replayable, while the part that records that it happened lives in one HTTP request. Anything that ends that request between the workflow finishing and the update committing — pod eviction, rollout, client disconnect — strands the row with the money already moved, and nothing anywhere retries it.

**2. The idempotent-replay path returns a row in any state, and the caller does not look at it.** On a duplicate idempotency key:

```kotlin
private suspend fun recoverConcurrentReplay(e: RuntimeException, idempotencyKey: String): Transaction {
    ...
    return transactionRepository.findByIdempotencyKey(idempotencyKey) ?: throw e
}
```

The row is returned as-is — including `PENDING` — and the endpoint answers `201`. On the other side, `SettlementAdapter` treats any 201 as success and never inspects `status`:

```kotlin
HTTP_CREATED -> {
    val txId = extractTransactionId(response)
    log.infof("Settlement booked for payment %s → transactionId=%s", payment.id, txId)
    SettlementOutcome(settled = true, transactionId = txId)
}
```

That adapter carries `@Retry(maxRetries = 2)` and `@Timeout(8_000)`. Today's workflow ran 6.91s — inside that window, but not by much. Whenever the first attempt exceeds 8s, the retry gets back the first attempt's still-`PENDING` row, reports "Settlement booked", and the payment is marked `SETTLED` while the transaction never leaves `PENDING`.

## What I could not determine

Whether today's 201 came from the first attempt or from a retry. transaction-service does not log transaction ids at INFO, and the pod that served the request is gone (Loki confirms `transaction-service-86dfbbf7b6-85g99` was not it — its stream starts at 15:35:37, after the settlement). The defect stands either way; the two paths differ only in which one fired first.

Worth noting the coincidence: transaction-service pods were being replaced by a rollout at 15:35:56, seconds after this request.

## Impact

- The customer sees a delivered payment as pending, indefinitely.
- `openbank.transactions.transaction.completed` is never emitted, so every downstream consumer of that event is silently missing this transaction.
- Nothing detects it. There is no sweeper for `PENDING`, no age-based alert, and the workflow reports `COMPLETED` — the same shape as #4182, one service further down.

## Suggested direction

- Make the terminal status write part of the durable unit: a final activity inside `PaymentWorkflow` that writes status + outbox, rather than caller code after `stub.execute()`. Then a lost request costs a retry, not a stranded row.
- Have `recoverConcurrentReplay` distinguish "already completed" from "still in flight". Returning a `PENDING` row as `201 Created` tells the caller something that is not true; `409` (which the adapter already treats as idempotent success) or waiting on the existing workflow would both be honest.
- Make `SettlementAdapter` assert the returned status instead of trusting the code. A 201 whose body says `PENDING` is not a booked settlement.
- Alert on `PENDING` older than a few minutes. One row today, and the only reason it was found is that a human asked why a payment looked odd.

Found while recovering the payment stranded by #4178. Related: #4182 (the same "durable work, non-durable finalisation" shape in domestic-payment).

JiRaska:
Closes #4238

## The defect

`TransactionService.initiateTransaction` saved the row `PENDING`, blocked on the Temporal payment
workflow, and wrote the terminal status **after** `stub.execute(...)` returned — i.e. outside the
durable unit:

```kotlin
val state: SagaState = withContext(Dispatchers.IO) { stub.execute(saved.id) }
return if (state == SagaState.COMPLETED) {
    val completed = saved.complete(clock)
    transactionRepository.update(transaction = completed, outboxMessage = /* ...completed */)
```

So the money-moving half (cover hold + ledger journal) was durable and replayable, while the record
that it happened lived in one HTTP request. Anything that ended that request in the window between
the workflow finishing and the update committing — pod eviction, rollout, client disconnect — left
the money moved, the workflow `COMPLETED`, and the row `PENDING` **forever**, with no
`openbank.transactions.transaction.completed` event and nothing anywhere to retry it. That is the
live transaction in #4238 (`version = 0`, `completed_at = NULL`, only the `initiated` outbox row).

## Which owner was chosen, and why not the other

**Chosen: the workflow.** The terminal write is now the last step of `PaymentWorkflowImpl`, a
`markCompleted` / `markFailed` activity implemented in `PaymentActivitiesImpl` — status **and** its
outbox message in one DB transaction, exactly as before, just executed by the thing that survives a
process death. `TransactionService` no longer computes a terminal state at all; after
`stub.execute(...)` returns it re-reads the row and reports it.

**Rejected: a sweeper over old `PENDING` rows.** It is a second thing that decides what a
transaction's terminal state is, from outside the workflow — and to decide correctly it would have
to ask Temporal what the workflow concluded, i.e. re-derive an answer the workflow already has. It
also cannot be right about the interesting case without a bounded age heuristic ("`PENDING` for more
than N minutes"), which is a guess that either finalises live payments or leaves real ones stranded
for N minutes. The workflow already owns the money movement; giving it the record of that movement
keeps one owner instead of two. A sweeper remains worth having as *detection* — see follow-ups.

**No second source of truth (constraint 2).** The DB row stays the only place a status lives. The
service's post-workflow branch is deleted, not duplicated: it now performs a `findById` and nothing
else. `TRANSACTION_COMPLETED_EVENT` / `TRANSACTION_FAILED_EVENT` moved with the write; they are not
declared in two places.

**No schema change, no new status value (constraint 3).** This is control-flow only. What is
deliberately *not* in this PR is listed under Follow-ups.

### Two details worth a reviewer's attention

- **The finalisation activity is outside the workflow's `try/catch`.** A failure there must not drop
  into compensation and reverse a journal that already settled. If it exhausts its (deliberately
  patient: 30 min schedule-to-close, uncapped attempts, 30 s max interval) retry window the
  *workflow* fails — visible in Temporal, and the caller gets an error instead of a `201` — rather
  than a `COMPLETED` workflow sitting next to a `PENDING` row, which is the state nobody could see.
- **Activities are at-least-once**, so `markCompleted` on an already-`COMPLETED` row is a no-op: no
  second outbox row, and no `TransactionUpdateConflictException` from the version-matched update.
  `markFailed` likewise refuses to touch a `COMPLETED` row (logs instead of throwing, so a final
  state cannot manufacture a retry storm).

## An existing test asserted the defect

Yes — this is why the diff deletes assertions.

`TransactionServiceTest.initiate transaction completes when workflow succeeds and emits completed
event` (and its FAILED twin) asserted that **`TransactionService` itself** called
`transactionRepository.update(COMPLETED, completed-event)` after the workflow returned. That is a
direct encoding of the non-durable ownership as expected behaviour: any fix to #4238 has to make
those assertions fail. They are replaced by their inverse — the service must call
`update(...)` **zero** times and must report the state read back from the row — and the write itself
is now asserted where it lives (`PaymentActivitiesImplTest`, `PaymentWorkflowImplTest`).

`WorkflowClientTestProducer`'s stub workflow also had to change: it returned `COMPLETED` without
touching the row, which was harmless while the caller did the write and would have left every
`@QuarkusTest` transaction `PENDING` afterwards. It now runs the **real** `markCompleted` activity,
so the `@QuarkusTest` fleet keeps exercising a real terminal write rather than a fiction.

## Falsification — the same test run both ways

New `PaymentWorkflowTerminalWriteIT` runs the **real** `PaymentWorkflowImpl` against the **real**
`PaymentActivitiesImpl` CDI bean over the IT Postgres, and then stops — it never executes a line of
the caller-side code that used to do the write. That is the "the request died" scenario, and what it
asserts is the property the fix introduces: *when the workflow returns, the terminal state is already
committed* (status `COMPLETED`, `completed_at` set, `version` incremented, exactly one
`...transaction.completed` outbox row).

It is written against API that exists on both sides, so the `origin/main` run is a real assertion
failure and not a compile error.

Both runs are `./gradlew :openbank-transaction-service:test` with the identical test file, verdicts
read from `build/test-results/test/TEST-*.xml`, not the console.

| run | code under test | tests | failures | errors | **skipped** |
|---|---|---|---|---|---|
| falsification | `origin/main` @ `3a08f5335` + this test file only | 1 | **1** | 0 | 0 |
| this branch | full module suite | 128 | 0 | 0 | 1 |

`origin/main` fails with exactly the defect, at the assertion, having got past the
`state == COMPLETED` check first — so the workflow really did run and really did return success:

```
com.openbank.transaction.integration.PaymentWorkflowTerminalWriteIT
  FAIL: workflow itself commits the terminal status and the completed outbox event()
  org.opentest4j.AssertionFailedError: [the workflow, not the caller, must own the terminal write (#4238)]
  expected: COMPLETED
   but was: PENDING
```

**On the skipped counts** (a module whose Quarkus boot fails reports every `@QuarkusTest` as
SKIPPED and reads as a pass, because the number people look at is the failure count):

- falsification run: **0 skipped**, 1 executed — the failure is a real assertion failure, not an
  absent test and not a compile error.
- branch run: **1 skipped**, and it is the pre-existing broker-gated
  `TransactionPactProviderVerificationTest.verifyPacts` (`@EnabledIfSystemProperty(pactbroker.url)`,
  no broker locally). Every `@QuarkusTest` executed: `PaymentWorkflowTerminalWriteIT` 1/1,
  `TransactionApiIT` 9/9, `TransactionConcurrencyIT` 3/3, `TransactionOutboxClaimIT` 2/2 — all 0
  failures, 0 skipped.

Gradle tasks that actually ran (Gradle stops at the first failure, so a green-looking list is not a
list that ran): `compileKotlin`, `compileTestKotlin`, `test` (BUILD SUCCESSFUL, 1m 18s), then
`detekt`, `runKtlintCheckOverMainSourceSet` / `runKtlintCheckOverTestSourceSet` /
`ktlintKotlinScriptCheck` → `ktlintCheck`, and `quarkusBuild` (BUILD SUCCESSFUL). ktlint first
flagged `class-signature` on the changed `WorkflowClientTestProducer` constructor;
`ktlintFormat` fixed it and the suite was re-run afterwards to the numbers above.

## Follow-ups deliberately NOT in this PR

Each is from #4238 and each is a bigger blast radius than a control-flow review:

1. **The idempotent replay still answers `201` with a `PENDING` body.** Both replay paths
   (`initiateTransaction`'s `existing != null` short-circuit and `recoverConcurrentReplay`) return
   the row as-is. After this PR that row is no longer stranded — the workflow finishes it durably —
   but a caller retrying inside the first attempt's window can still be handed a `PENDING` row with
   a `201`. The honest answers (`409`, or awaiting the in-flight workflow) both change the REST
   contract: `openapi.yaml` + `info.version` bump + contract test, which does not belong in the same
   review as the durability fix.
2. **`SettlementAdapter` (domestic-payment) trusts the status code and never inspects `status`.**
   Different service, different PR.
3. **Nothing detects a stranded `PENDING`.** An age-based alert / sweeper is still worth having as
   *detection*, now that finalisation has an owner.

## Governance

`openbank-transaction-service` is in `rules.yaml: money_path_services`, so this needs **2 approvals**
(ADR-0030). Its threat model `docs/threat-models/openbank-transaction-service.md` already exists and
is unchanged: no trust boundary moves here — no REST surface, authn/listener/transport key,
NetworkPolicy or workload manifest is touched, only which component performs an already-existing
database write.


JiRaska:
Reviewed to money-path standard, with falsification. **No blocking defect — this is a well-built fix.** One worth-fixing operational issue and three narrow notes.

## The window it closes is real, and the subtlest decision is the right one

Under the old code `save()` committed `PENDING`, the workflow moved the money, and the status write lived in the HTTP request *after* `stub.execute()` returned. A pod eviction there left money moved, workflow `COMPLETED`, row `PENDING`, no `transaction.completed` event, and nothing to retry.

The new residual window is strictly smaller and non-monetary: between the journal posting and the finalisation activity committing, the row is stale — but Temporal owns the retry, so it converges. The only non-converging path is the 30-minute expiry, which fails the **workflow**, making the inconsistency visible in Temporal rather than silent.

Keeping the `finalisation` stub **outside** the `try/catch`, so a finalisation failure cannot trigger a journal reversal of already-settled money, is the subtlest correct call in the diff.

## Falsified, not merely green

I reverted the behaviour twice and re-ran:

1. Neutered the terminal write in `PaymentWorkflowImpl.execute()` → both new/updated `PaymentWorkflowImplTest` cases **FAILED**, and `PaymentWorkflowTerminalWriteIT` **FAILED** (so its KDoc's falsification claim is true).
2. Restored `TransactionService.kt` from `origin/main` → `TransactionServiceTest` **9 of 13 FAILED**.

Both restored, tree clean afterwards.

Full module: **128 tests, 0 failures, 0 errors, 1 skipped** — and I read the skipped count rather than the failure count, since a suite that silently stops running reads as green. The single skip is `TransactionPactProviderVerificationTest`, broker-gated and pre-existing. The three named suites report 12 / 5 / 13 with **0 skipped**, so no `fun x() = runBlocking` silent-skip.

`PaymentWorkflowTerminalWriteIT` ran against real Postgres + Redpanda, which also proves the CDI wiring: the new `TransactionEventPublisher` and `Clock` constructor params resolve in a booted app. That is also what would have surfaced `HR000068` had the new writes been reactive on an activity thread — they go through the pre-existing `runOnVertxContext` bridge, matching the other four activities.

**Idempotency is stronger than the KDoc claims.** Two layers: the status guard makes a replay a no-op, and underneath it `PanacheTransactionRepository.update` re-finds by `id = ?1 and version = ?2` and mutates the managed entity — a real compare-and-set. A racing writer loses with `TransactionUpdateConflictException`, the activity retries, reloads, sees `COMPLETED`, no-ops. No last-write-wins, no duplicate outbox row, and no `persist`-on-assigned-id trap anywhere in the new paths. The read and write are separate sessions, so a TOCTOU window exists, but the `WHERE version =` predicate closes it.

No `openapi.yaml` bump or migration needed, and correctly absent: no status value added, no API shape change. Temporal determinism on rollout is safe — the new activity calls are appended after the existing command sequence, so an in-flight workflow replays identically.

## Worth fixing: the caller's HTTP request can now block for 30 minutes, not 10

```kotlin
// PaymentWorkflowImpl.kt:44
private val finalisationOptions: ActivityOptions = ActivityOptions.newBuilder()
    .setScheduleToCloseTimeout(Duration.ofMinutes(FINALISATION_SCHEDULE_TO_CLOSE_MINUTES))  // 30
```

No `setMaximumAttempts`, unlike the main options. And `stub.execute()` only returns once finalisation closes, inside `withContext(Dispatchers.IO)`.

Concretely: the Postgres primary fails over for 15 minutes. Every payment that has already posted its journal sits in `markCompleted` retry, and each one pins a `Dispatchers.IO` thread for the whole window — previously bounded at 10 minutes. Under load that exhausts the IO dispatcher and takes down endpoints that have nothing to do with payments.

The workflow itself recovers correctly, which is the design's whole point — so **the caller does not need to wait for it**. `WorkflowStub.start()` plus a bounded `getResult(timeout)`, or a short scheduleToClose on the caller's view. The durability argument for the long window applies to the workflow, not to the HTTP thread.

## Notes

- **`PaymentActivitiesImpl.kt:112-118`** — the replay guard recognises only `COMPLETED`, but `Transaction.complete()` requires `status in {PENDING, PROCESSING}`. Reachable sequence: `markCompleted` commits, the activity result is lost to worker eviction, an operator reverses the transaction, the replay lands on `REVERSED` → `IllegalStateException` on every attempt for 30 minutes, then the workflow fails. Money and row are both fine; the caller gets a 500 half an hour later for a payment that succeeded. Guard on terminal-ness rather than the single value.
- **`:145`** — same one status short in the other direction: `Transaction.fail()` rejects only `COMPLETED`, so `markFailed` can overwrite a `REVERSED` row to `FAILED`.
- **`:139-142`** — the `COMPLETED`-plus-compensation contradiction is logged and nothing else. Not throwing is right, but a `log.warnf` is the only record that a journal was reversed while the row says `COMPLETED`, and nothing alerts on it. The path is argued unreachable; a counter costs less than the argument.
- **Scope, not a defect here:** no backfill. Rows already stranded `PENDING` by the old bug are not repaired, and I found no stuck-`PENDING` reaper anywhere in `src/main`. Worth a follow-up issue if #4238 left rows behind in the sandbox.

JiRaska:
Pushed (`8f9795d4e`) — the finalisation activity now uses the same 10-minute window and `MAX_ATTEMPTS` as its four siblings.

**I had to correct my own framing first, and it changed the fix.** I said the caller "previously waited 10 minutes". It did not: `stub.execute()` waits for **every** activity, so the caller already waited 2 × 10 min on the happy path and up to 4 × 10 with compensation. A caller-side timeout of ten minutes — my first instinct — would therefore have cut off payments that succeed today, and would have needed a non-terminal response shape, i.e. an API change.

So the bound moved to where it belongs. The durability argument for a patient window is an argument about the **workflow**, and never was one about holding an HTTP thread.

What it costs, and I would rather state it than gloss it: an outage longer than ten minutes now fails the workflow instead of converging. That is the same visible-in-Temporal outcome the 30-minute window ended in, arriving sooner — and a failed workflow beside a stale row is still far better than a COMPLETED workflow beside a PENDING one, which is what this PR set out to fix.

```
128 tests, 1 skipped (pre-existing broker-gated pact), 0 failures
compileKotlin, ktlintCheck, detekt all RAN
```

**One honest gap:** no test pins these `ActivityOptions` values, so this rests on review rather than on a test. The constants now read as one policy shared with the siblings, which is the cheapest guard available against them drifting apart again — but if you would rather have it asserted, that wants the options hoisted somewhere testable, which is a bigger change than this finding justifies.

PR #4308, opened 2026-08-09
fix(card-issuance): alert on dead-lettered outbox rows and add an operator requeue path

### LINKED ISSUE #4005: card-issuance has published zero events: all 24 card_outbox rows are DEAD on an open circuit breaker, stamped 1970, with no self-healing path

JiRaska:
Closes #4005

#4163 fixed the **mechanism** — an open circuit breaker no longer drives outbox rows to terminal
`DEAD`, and the `Instant.EPOCH` default is gone. It did not give the 24 rows already stranded a way
back, and it did not make dead-lettering visible to anyone. This is those two halves.

## The premise, re-measured today (not taken from the issue)

```
$ kubectl -n payments exec card-issuance-db-1 -- psql -U postgres -d openbank_cards \
    -c "SELECT status, count(*), min(created_at), max(created_at), max(attempt_count) FROM card_outbox GROUP BY 1;"
 status | count |          min           |          max           | max
--------+-------+------------------------+------------------------+-----
 DEAD   |    24 | 1970-01-01 00:00:00+00 | 1970-01-01 00:00:00+00 |  10
```

24 rows, all `DEAD`, no `SENT`, no `PENDING`, 1970, at the attempt ceiling. Confirmed.

And the gap this PR closes, also measured rather than assumed:

- The only outbox alert rules in gitops are `*OutboxBacklogStuck` (transaction, clearing, swift,
  sepa-payment, sepa-instant, domestic, sdd). Every one of them selects
  `openbank_outbox_backlog{service=...}` — **backlog, not dead-lettering**. `card-issuance` has no
  PrometheusRule at all.
- `openbank_outbox_backlog{service="card-issuance"}` **is** live and scraped right now, and it reads
  **`0`** — the same value a perfectly healthy service publishes, because `countProcessable()`
  excludes `DEAD` on purpose. That flat zero is precisely what let this run unnoticed.
- There is no requeue path anywhere: `grep`ping the fleet for a DEAD-row recovery turns up nothing,
  and `OutboxRepository` exposes only `listProcessable` / `claimProcessable` / `countProcessable` /
  `markSent` / `markFailed`.

## 1. Detection

New gauge `openbank.outbox.dead_lettered{service=...}` (`DomainMetrics.registerOutboxDeadLettered`
+ `AbstractOutboxDeadLetterGauge`, the sibling of the existing `AbstractOutboxBacklogGauge` — same
cached-`AtomicLong` + `suspend` refresh shape, same CDI-proxying constraint), bound in
card-issuance by `CardOutboxDeadLetterGauge`. The alert follows the `*OutboxBacklogStuck` shape and
lands in the same file those live in:

```yaml
- alert: CardIssuanceOutboxDeadLettered
  expr: openbank_outbox_dead_lettered{service="card-issuance"} > 0
  for: 15m
  labels: { severity: critical, service: card-issuance }
```

### The metric did not already exist, and the counter that does cannot back this alert

`DomainMetrics.outboxDead()` already increments a **counter**, `openbank.outbox.dead`. It is not a
substitute, for a reason that is measurable rather than stylistic:

```
$ curl -s --get .../api/v1/query --data-urlencode 'query=openbank_outbox_dead_total'
count: 0
```

**Zero series fleet-wide**, while card-issuance holds 24 DEAD rows. A Micrometer counter is not
created until its first increment, and it resets on every pod restart — so it answers "did *this
process* dead-letter anything", never "are there dead rows in the table". An alert on it would have
matched nothing, forever, and rendered exactly like "no problem". Hence a gauge read from the table.

### Proof the alert expression matches a metric that exists

Two independent legs, because "the selector looks right" is the failure mode here.

**(a) The label and namespace shape, measured against the live Prometheus.** The sibling gauge
registered through the identical code path resolves today:

```
openbank_outbox_backlog{container="card-issuance-service", endpoint="management",
  instance="10.80.116.152:8085", job="observability/openbank-services",
  namespace="payments", pod="card-issuance-service-5f999f5df7-xgbt5",
  service="card-issuance"}  =  0
```

Same registry, same management port, same ServiceMonitor job, and the `service` label comes from the
same `.tag("service", service)` call with the same literal `"card-issuance"`. So the new gauge is
scraped with the same label set; the alert selects on `service` alone and adds no label the metric
does not carry.

**(b) The series name, derived rather than retyped.** `OutboxDeadLetterAlertNamingTest`
(libs-runtime) registers the gauge into a real `MeterRegistry`, renders the name through
**Micrometer's own `PrometheusNamingConvention`** (not a hand-rolled dot→underscore helper, which
would only prove itself), builds `<series>{service="<tag>"}` from what came back, and asserts that
string is what the committed rule's `expr:` line uses.

Matched against the `expr:` **line**, not against the file — the rule is preceded by a comment block
naming `openbank_outbox_dead_lettered` three times, so a whole-file `contains` would stay green with
the expression deleted.

## 2. Recovery — and why it is an operator endpoint

`POST /api/v1/cards/outbox/requeue` (`ROLE_ADMIN`, `X-Operator-Id` required), optional `?eventId=`
to requeue one row. Moves `DEAD -> PENDING`, resets `attempt_count` to 0 and clears `last_error`.

### The idempotency evidence, from the code

Requeueing republishes an event that may already have been delivered, so whether it is safe is a
fact about the consumers of `openbank.cards.events`. There are two:

| consumer | idempotent? | evidence |
|---|---|---|
| `openbank-campaign-service` | **yes** | `EnrolmentTriggerConsumer` funnels through `TriggeredEnrolment.ALREADY_ENROLLED`; `ConversionConsumer.recordIfConverted` returns early on `context.alreadyConverted`. A replay is a no-op. |
| `openbank-audit-service` | **no** | `AuditConsumer` builds every entry with `id = UUID.randomUUID()` and calls `repo.save(entry)` with **no dedup on the outbox `ce-id` / `idempotency-key` header** — grepping the whole service for `idempot`/`dedup`/`ce-id` returns nothing but the outbox repository's own `eventId` lookups. `AuditRepository.save` then links the row into a hash chain, and `audit_entries` is append-only at the database. |

One non-idempotent consumer settles it. A replay appends a **second, permanent, undeletable** audit
record for an event that happened once — so an automatic sweeper would silently duplicate audit
history on every deploy that happened to find a DEAD row.

**The three candidates, and why this one:**

- **Scheduled sweeper — rejected.** Correct only if every consumer is idempotent. One is not, and
  the damage it does is to the evidentiary record, in a table nothing can clean up afterwards.
- **One-shot Flyway migration — rejected.** Runs automatically at boot (same objection), fixes
  exactly one day's rows and nothing afterwards, and cannot be re-run once applied — its checksum is
  frozen, so editing it later breaks startup.
- **Operator endpoint — chosen.** The decision that actually needs making is *"were these events
  ever delivered?"*, and that is a human judgement. For the #4005 rows the answer is unambiguous
  (zero `SENT`, nothing ever published, so replay duplicates nothing), but the endpoint outlives
  those rows and the next case will not be. `X-Operator-Id` makes the call attributable; it is
  logged at WARN with the row count.

`ROLE_ADMIN` **only**, not `ROLE_OPERATOR`, in `@RolesAllowed` and in `card_issuance_rest_ext.rego`
alike — this file already documents what happens when the two drift (the `card.block` bug from
#1322/#1323), so the new rule is deliberately kept out of the `operator-card-write` action set.

## Falsification — what each new check was fed, and what happened

Every one of these was run; none is a claim about what the check "should" catch.

| check | input it must flag | result |
|---|---|---|
| `OutboxDeadLetterAlertNamingTest` | alert selector widened with a label the metric does not carry (`namespace="payments"`) | **FAILED** as required |
| `OutboxDeadLetterAlertNamingTest` | alert re-pointed at the counter (`openbank_outbox_dead{...}`) | **FAILED** as required |
| `CardOutboxRequeueIT` (all 5) | a top-level `private fun` inserted between `@Path` and the class — the annotation-stealing 404 trap | **all 5 FAILED** (route unregistered) |
| `CardOutboxRequeueIT` "fresh attempt budget" | `attemptCount = 0` dropped from the requeue HQL | **FAILED** as required |
| `card_issuance_rest_ext_test.rego` `..._denied_for_operator` | `card.outbox.requeue` folded into the `operator-card-write` action set | **FAIL 1/14** as required |
| `CardOutboxDeadLetterGaugeTest` | gauge wired to `countProcessable()` instead of `countDead()` — both return 0 on healthy data, 24-vs-0 on the live shape | pinned by an explicit assertion on the live `24 DEAD / 0 processable` fixture |

Restored and re-run green after each.

**One honest caveat on the alert test.** Gradle does not track the PrometheusRule YAML as a test
input, so editing that file alone leaves `:openbank-libs-runtime:test` `UP-TO-DATE` and the guard
does not re-run. The falsifications above were done with `--rerun-tasks`. The test still holds the
seam on any libs-runtime change; it is not a substitute for a gate over the rules file, which does
not exist today.

## Traps this PR was steered around

- `refresh()` on both gauges is a **`suspend fun`** — a plain `@Scheduled` method carries no Vert.x
  context and `runBlocking` around the reactive count throws `HR000068` silently, freezing the gauge
  at its boot zero (the reading that looks healthiest). `VertxContextSupport.subscribeAndAwait` is
  not used as a workaround.
- `eventId` and `X-Operator-Id` are declared **nullable** with the guard in the body. A non-nullable
  Kotlin param does not make a parameter required — JAX-RS injects null regardless, and in a
  `suspend fun` no intrinsic null-check is even emitted. `IllegalArgumentException` → 400 via
  libs-runtime; no service-local mapper was added. Both cases are asserted over real HTTP.
- Requeue is a **bulk HQL update**, so it never meets the Panache `persist()`-on-assigned-`@Id`
  INSERT-only trap, and the affected row count comes from the database rather than from a loop.
- `created_at` is **not** restamped. It is the `ORDER BY created_at ASC` claim key; the epoch rows
  sorting first is wanted here — they drain first, then leave the claim set entirely.
- OPA bundle regenerated with `gen-card-issuance-opa-bundle.sh` after the final `.rego` edit (not
  before). Only `card-issuance-opa-bundle.yaml` and card-issuance's own `policy-checksum` annotation
  moved — verified, `payments-services.yaml` is a 1-line diff.
- `openapi.yaml` `info.version` read off **live `origin/main`** immediately before bumping: `1.6.0`
  → `1.7.0` (additive endpoint, ADR-0048 minor). Re-checked after pushing; still `1.6.0` on main.
  Note **#4194** (`feat/dispo-card`) is open and also bumps this spec — it currently claims `1.6.0`,
  a version already on main, and is `DIRTY`, so it needs its own re-bump regardless; whichever lands
  second takes the next number.
- `version.txt` / `CHANGELOG.md` untouched — release-please owns the release axis.

## Local gate — the tasks that actually ran

```
:openbank-libs-runtime:compileKotlin            SUCCESS
:openbank-card-issuance-service:compileKotlin   SUCCESS
:openbank-libs-runtime:compileTestKotlin        SUCCESS
:openbank-card-issuance-service:compileTestKotlin SUCCESS
:openbank-libs-runtime:ktlintCheck              SUCCESS
:openbank-libs-runtime:detekt                   SUCCESS
:openbank-card-issuance-service:ktlintCheck     SUCCESS   (after ktlintFormat fixed import order)
:openbank-card-issuance-service:detekt          SUCCESS
:openbank-libs-runtime:test                     SUCCESS
:openbank-card-issuance-service:test            SUCCESS   (--rerun-tasks)
:openbank-card-issuance-service:quarkusBuild    SUCCESS
```

Read from the JUnit XML, not the console — including the SKIPPED count:

```
card-issuance  tests=155  skipped=1  failures=0  errors=0
libs-runtime   tests=203  skipped=0  failures=0  errors=0
```

The single skip is `CardIssuancePactBrokerProviderVerificationTest`, gated on `pactbroker.url`
(expected). All 5 `CardOutboxRequeueIT` tests ran — 0 skipped, so the Testcontainers Postgres
really did come up.

CI gates run locally, all pass: `yamllint`, `opa-sidecar-bundle-shape`,
`gitops-duplicate-resources`, `nonnull-jaxrs-param-ratchet`, `no-runblocking-in-a-scheduled-body`,
`scheduler-cron-syntax`, `outbox-dispatch-enabled-guard`, `operator-write-naming`,
`matrix-write-grants`, `no-service-local-exceptionmapper-collision-with-libs-runtime`,
`scheduled-trigger-emitted`. `opa test` on the card extension: **14/14**.

## After merge

Deploying does **not** drain the 24 rows — that is the point of the design. Someone has to look at
them and decide, then:

```
POST /api/v1/cards/outbox/requeue     (ROLE_ADMIN, X-Operator-Id: <you>)
```

For these specific rows the replay is free: zero were ever `SENT`, so nothing downstream has seen
them and no audit record is duplicated. The alert should clear within 15m of the gauge reaching 0.

## Not in scope

`AbstractOutboxDeadLetterGauge` is fleet-reusable but only card-issuance binds it here — rolling the
gauge and a matching alert out to the other 20-odd outbox-bearing services is a separate sweep, and
doing it in this PR would put a 20-service diff under a card-issuance scope.


JiRaska:
### Follow-up on the first CI run (`gates (kotlin)` red, now green)

`identifier-intent-guard` failed on this PR by matching the **KDoc that explains the defect**, not
any code: `CardOutboxAdminResource`'s doc block quoted openbank-audit-service's own id-minting call
verbatim as the evidence for why requeue cannot be automatic, and the guard read that as this file
minting a bare identifier.

That is the code-about-code trap this repo already documents for `check-roles-allowed-realm.py` and
`check-advisory-gate-registration.py` — a guard over source text needs an explicit rule for prose
about code. Fixed here the cheap way (`b0e6c02ac`, same fact in prose), because widening the guard
belongs in its own PR; `.github/scripts` comment-stripping for this checker is filed separately.

Local re-run after the fixup: `run-gates.py --group kotlin` with `PR_DIFF_BASE=origin/main` →
**19 gates PASS=19**.

JiRaska:
### Second CI finding, and a design decision it forced: `scheduler-liveness-adoption` (ADR-0237)

`gates (gitops)` flagged the new `CardOutboxDeadLetterGauge` as a `@Scheduled` job with no
workflow-liveness heartbeat. There was an easy way out: `check-scheduler-liveness.py` exempts outbox
infrastructure **by role** (`AbstractOutboxDispatcher|AbstractOutboxBacklogGauge|class
\w*(OutboxDispatcher|OutboxBacklogGauge)`), and widening that regex by one alternative would have
made this green in a one-line diff.

Declined, because the exemption's stated premise does not transfer. It rests on
`openbank.outbox.backlog` being its own freshness signal — plausible for a `> 100` alert. This
gauge's alert is `> 0`, so a refresh tick that dies leaves the gauge pinned at its **boot zero**,
the value that reads as perfectly healthy, and the alert then never fires. That is the same class of
silence #4005 is about, and it would have been introduced by the very PR that exists to end it.

So `ee436b041` registers a real ADR-0237 heartbeat instead
(`card-issuance-outbox-dead-letter-gauge`, expected interval pinned to the same 60s as the cron).

One detail worth flagging for whoever adds the second one of these: the heartbeat is registered on
the **concrete bean**, not in `AbstractOutboxDeadLetterGauge`, even though the base is where it
"belongs" structurally. The gate classifies the file that declares `@Scheduled`, so a registration
hidden in a shared superclass leaves every subclass reading as an unmonitored scheduler. Putting the
call where the gate looks is the honest fix; mentioning `registerWorkflowLiveness` in a comment to
satisfy the grep would not be.

Local after the fix, with `PR_DIFF_BASE=origin/main`: `--group gitops` **27 gates PASS=26** (the one
`UNFALSIFIED` is pre-existing and unrelated), `--group kotlin` **19/19**. Full module gate re-run
with `--rerun-tasks`: card-issuance `tests=155 skipped=1 failures=0`, libs-runtime `tests=203
skipped=0 failures=0`.

JiRaska:
The red here is **not this PR's defect** — I chased it as one first and was wrong.

`build (openbank-account-service)` fails on `SavingsProposalExpiryWorkflowLivenessTest`, 2 of 3:

```
Expecting actual: 0.0 to be greater than: 1.5768E9
```

That test fails on clean `origin/main` too, reproduced in a detached worktree with no local changes.
#4208 seeded the workflow-liveness age gauge at registration rather than at `Instant.EPOCH` and
updated the fleet's liveness tests — 17 of 18. account-service is the 18th, missed because it landed
in #4180 after #4208 was authored.

Nothing reported it because CI is path-scoped: #4208 touched no file under `openbank-account-service/`,
so that module has not been rebuilt since, and its last green predates the libs change. **This PR is
simply the first to edit `libs-runtime` and thereby pull account-service into its changed set** — it
surfaced a pre-existing red rather than causing one.

Fixed in #4325 (test-only). Once that merges, merge `origin/main` in here and the check clears.

My first hypothesis — that this branch predated #4208 and merging `main` would fix it — was wrong,
and I only found that out by running the test after the merge rather than treating the clean merge
as the answer. Recording it here so the next reader does not repeat it.

JiRaska:
Reviewed. **Design quality is high** — it anticipates the boot-zero, counter-vs-gauge, `HR000068` and CDI-proxy traps explicitly and gets each right. Two things worth fixing before merge, one operational and one that undermines the guard this PR ships.

## Worth fixing 1: the seam guard cannot see the change it exists to catch

`OutboxDeadLetterAlertNamingTest.kt:43`:

```kotlin
private val rulesFile = File("../openbank-infra/gitops/components/payments/prometheus-rules.yaml")
```

That file is **outside `:openbank-libs-runtime`'s Gradle inputs**, and the module's `tasks.test { }` block declares no `inputs.file(...)`. Measured: with the alert expression rewritten to `openbank_TOTALLY_WRONG`, `./gradlew :openbank-libs-runtime:test --tests '*OutboxDeadLetterAlertNamingTest*'` reports `Task :openbank-libs-runtime:test UP-TO-DATE` and **BUILD SUCCESSFUL**. It only goes red under `--rerun-tasks`.

Combine that with path-scoped CI — a gitops-only PR does not build libs-runtime at all — and a future PR that edits only the PrometheusRule passes cleanly. This is the "a gate is only as reachable as the job it sits in" shape, one layer further down into Gradle's up-to-date check, and it is the quiet direction: the test reports success about work it never did.

The test's *construction* is the good part and worth keeping: it renders the series through Micrometer's own `PrometheusNamingConvention` from a **real registration** and compares against the committed `expr:` line rather than the file, deliberately dodging the comment-matches-the-grep trap. It deserves to actually run. Either declare the file as a task input, or move the assertion into a `gates.yaml` gate that runs unconditionally.

Falsification otherwise holds: rewriting the alert to `openbank_outbox_dead_total{...}` and forcing a rerun does turn it red.

## Worth fixing 2: this alert will be firing `critical` from the moment it deploys

`prometheus-rules.yaml:233` — `expr: openbank_outbox_dead_lettered{service="card-issuance"} > 0`, `for: 15m`, `severity: critical`. The PR's own text says card-issuance holds **24 DEAD rows with no automatic drain**.

So ~15 minutes after the pod rolls this fires, and stays firing until a human calls the new requeue endpoint. A permanently-firing `critical` is the documented pattern in this repo where the alert becomes the resting state and readers learn to ignore it — and it would be firing on day one against the exact backlog it was written to describe.

The alert is not wrong; the **sequencing** is. Either drain the 24 rows first and merge the alert after, or land it with a dated note saying it is expected-firing until #4005's rows are cleared. (This rests on the PR's own count plus `requeueDead` being the only drain path in the tree — I did not query the live database.)

## Checked and correct

- **Boot-zero is real here and handled honestly.** `cached = AtomicLong(0)` plus `> 0` means a dead refresh pins the gauge at a healthy-looking zero. `AbstractOutboxDeadLetterGauge` says exactly that and declines the outbox exemption instead of hand-waving it. I confirmed the backstop exists rather than taking it on trust: `prometheus-rules-workflow-liveness.yaml:59` fires at `age > 2 * expected_interval` (120s here) `for: 15m`, and the heartbeat is registered on the concrete bean so `check-scheduler-liveness.py` can see it. `refresh()` is a `suspend fun`, so no `HR000068`.
- **No new range vectors**, so nothing truncates against the 12h retention.
- **GitOps path and labels**: appended to a file already carrying four live PrometheusRules on the synced path; labels match all four siblings byte-for-byte, and `severity: critical` is one of the two values Alertmanager actually routes.
- **`CardOutboxRequeueIT` is the right shape** — a real `@QuarkusTest` over HTTP with a JDBC read-back asserting `attemptCount == 0`, `lastError == null`, `createdAt` unchanged, plus a 404-vs-200 routing assertion. That is the only test shape that catches both the `persist`-vs-`merge` trap and the unregistered-`@Path` trap. And `CardOutboxDeadLetterGaugeTest`'s `countDead()` vs `countProcessable()` case is the one assertion that could not have been written from a green-path fixture, since both return 0 there.
- **Authz**: `@Authorize(action = "card.outbox.requeue", resource = "")` is the fleet-standard non-scoped form (97 sites), not the #3521 defect. The rego grants `ROLE_ADMIN` only, stays out of `operator-card-write`, matches `@RolesAllowed` exactly, 14/14 rego tests pass. `openapi.yaml` 1.6.0 → 1.7.0 is the right additive bump and `version.txt` correctly untouched.

## Note

`CardOutboxDeadLetterGauge.kt:35-46` uses a `lateinit var` + secondary-constructor shape with a `domainMetrics` field shadowing the base's `metrics`, plus `@Suppress("ProtectedMemberInFinalClass")`, while `AbstractOutboxBacklogGauge`'s subclasses use plain primary-constructor injection. It works — the IT proves the bean constructs and the route serves — but it is a one-off shape in a fleet-shared abstraction, and the base class's own KDoc shows the simpler form. Worth aligning before a second service copies the concrete class instead of the example.

JiRaska:
Pushed (`5ffa3e29a`) — the naming test can now see the file it asserts about.

```kotlin
inputs.file(rootProject.file("openbank-infra/gitops/components/payments/prometheus-rules.yaml"))
```

Falsified with the exact experiment that exposed it, breaking **only** the YAML with no Kotlin change at all:

```
baseline                          rc=0
alert renamed to a wrong metric   rc=1     (before the fix: test UP-TO-DATE, BUILD SUCCESSFUL)
```

The test's construction was always the good part — rendering the series through Micrometer's own naming convention from a real registration, and comparing against the `expr:` line rather than the file — it simply never ran.

I also restored the verbatim `id = UUID.randomUUID()` quote in `CardOutboxAdminResource`'s KDoc and folded the reason into the note you already had there, so it now says the guard was the thing that was wrong rather than implying the prose form was a deliberate choice.

**That creates a merge order, and I measured it rather than assuming:** with the restored quote committed, `main`'s current `check-new-random-uuid.sh` exits **1** and names this file. So **#4311 has to land first**. My first check said otherwise — it reported `rc=0` because the guard is diff-scoped and reads committed state, while my edit was still only in the working tree. Worth knowing if you re-run it.

**The other finding still stands and is yours to decide:** `> 0` with `for: 15m` and `severity: critical`, against 24 DEAD rows with no automatic drain, means this fires ~15 minutes after the pod rolls and stays firing until someone calls the requeue endpoint. Either drain first and merge the alert after, or land it with a dated note that it is expected-firing until #4005's rows are cleared. I did not touch that — it is a sequencing judgement, not a defect.

JiRaska:
Post-rebase verification (past #4311, #4390): clean.

- `workflow-run-step-size` gate: PASS
- `OutboxDeadLetterAlertNamingTest`, `CardOutboxDeadLetterGaugeTest`, `CardOutboxRequeueIT`: all green, no FAILED entries
- Both earlier review fixes (gate-visibility `inputs.file`, alert-sequencing note) survived the rebase intact
- Merge with current main is clean

No further action needed from my side on this one.

JiRaska:
Rebased onto current `main` — the only conflict was `.github/scripts/check-verification-metadata-complete.py`, an incidental drift file (not this PR's own content): a stray earlier commit on this branch carried a now-superseded version of the #4909/#5047 heap-bump comment. Took main's canonical version wholesale (verified: this file has no card-issuance-specific content, confirmed via `git diff` history). ADR registry regenerated with zero drift. `:openbank-card-issuance-service:compileKotlin`/`ktlintCheck` green.

JiRaska:
## Heads-up: `info.version` collision with #6019 (not a review of this PR's content)

This PR bumps `openbank-card-issuance-service/src/main/resources/openapi.yaml` `info.version` `1.7.0 -> 1.8.0`.
So does **#6019** (`fix(card-issuance): reconcile IssueCardRequest with the DTO it describes`), which is open and edits the same file.

Both change the same single line from the same value to the same value, so **git merges them
without a conflict** — nothing is flagged, and whichever lands second ends up with `1.8.0`
describing a document that also contains the other PR's changes. One API-contract version is
silently skipped, and the ADR-0048 classification for the second merge is computed against a
version that was already consumed.

Reading `origin/main` before bumping (which both PRs did) cannot catch this: `main` still holds
`1.7.0` for both. The check that catches it is against the *open* set.

Whoever merges second should re-read `git show origin/main:openbank-card-issuance-service/src/main/resources/openapi.yaml | grep -m1 '^  version:'`
after the first lands and take `1.9.0`. No action needed if this one merges first.

I'm reviewing the #5962 enum-drift family and flagged the same thing on #6019; posting here so the
half that merges first also knows.

JiRaska:

Live `origin/main` is **card-issuance `1.7.0`**. Two open PRs bump it to `1.8.0`:

- #6019
- #4308

Unlike the sepa-instant pair, these two are a **genuine race** — their card-issuance spec changes
differ, so this is not duplicate work and both sets of changes are wanted.

Whichever merges **second** must re-read the version off live `origin/main` and take `1.9.0`:

```bash
git fetch origin main
git show origin/main:openbank-card-issuance-service/src/main/resources/openapi.yaml | grep -m1 '^  version:'
```

No release tag is involved — tags sit on the 0.x release axis
(`card-issuance-service-v0.24.1`), which ADR-0048 keeps separate from the 1.x API-contract axis,
and no merged PR has consumed `1.8.0`.

Nothing to change right now. Note this is **not** what is currently blocking either PR: the red
#6220 fixed on main at 14:57Z, after those runs started — it clears on the next run.

JiRaska:
## Review — stale base, live version race, and one alert that ships already firing

Reviewed as a **reviewer only** (no approve, no merge). Read the full 3-dot diff (`git diff origin/main...origin/fix/card-issuance-dead-outbox`, 20 files).

### 1. The red CI is NOT this PR's fault — it is a stale base

`build (openbank-libs-runtime)` fails on:

```
openbank-libs-runtime/src/main/kotlin/com/openbank/libs/llm/OpenAiCompatibleLlmGatewayClient.kt:72:26:
The function chat is too long (61). The maximum length is 60. [LongMethod]
```

That file is **not in this diff**. It was fixed on `main` by `64fb7e88c` — *fix(libs): keep the LLM gateway chat() inside the detekt LongMethod budget (#6148)*. This branch is 160 commits behind and predates it. `all-green` is red purely as a consequence.

**Action: rebase onto `origin/main`.** Nothing needs changing in the PR's own code for CI. The "four required contexts missing entirely" reading is the same thing — path-scoped contexts never got to run once the libs-runtime job died.

### 2. Live version race on `openapi.yaml:info.version` — 1.8.0 is contested

- `origin/main` card-issuance spec: `1.7.0`
- this PR: `1.8.0`
- **#6019** (`fix/card-issuance-openapi-enum-drift-5962`): also `1.8.0`

Confirmed live at the time of writing. The two spec changes genuinely differ (this adds `POST /api/v1/cards/outbox/requeue`; #6019 reconciles `IssueCardRequest` with its DTO), so this is not a duplicate — it is a straight collision. Whoever merges second must re-read `info.version` off **live `main`** and take `1.9.0`. A matching version line merges "cleanly" because git sees identical text, not a taken version.

### 3. The alert is designed to fire on deploy and stay firing

`openbank-infra/gitops/components/payments/prometheus-rules.yaml` — `CardIssuanceOutboxDeadLettered`, `severity: critical`, `for: 15m`, and the comment says so itself:

> EXPECTED TO FIRE ON DEPLOY, 2026-08-09. ... this goes critical ~15m after the first roll and stays there until an operator drains them

I do not think shipping a permanently-lit `critical` on a hand-shake ("drain it in the same session") is the right trade in this estate, and the comment's own next sentence concedes the reason: *"a critical that is always on stops being read, and this repo has already produced one."* The honest shapes are either:

- land it as `severity: warning` with a follow-up issue that promotes it to `critical` once the 24-row backlog is drained, or
- put the requeue in the same deploy window and get written confirmation the backlog is at zero before the rule goes in.

This is an **author decision**, not a defect — but the note as written relies on a human reading a YAML comment at the right moment, which is not a control.

Related, and worth stating: the alert's date-stamped "if this note is still here and the alert is still firing, the backlog was not drained" is now ~12 days stale. The premise it was written under (24 DEAD rows, live) needs re-measuring before merge — the count may have moved.

### 4. What I checked and found nothing wrong with

- **`@Scheduled` context.** `CardOutboxDeadLetterGauge.kt:69` is `suspend fun refresh()`, `@Startup` + `@ApplicationScoped`, `@PostConstruct` and `@Scheduled` on the concrete bean, not the base. Correct per ADR-0013 and the `HR000068` rule. The base class KDoc calls both out explicitly.
- **Nullable JAX-RS params.** `CardOutboxAdminResource.requeueDead` declares both `eventId: String?` and `operatorId: String?` and does `requireNotNull` in the body — the correct shape, and the KDoc states the `suspend fun`-emits-no-intrinsic reasoning correctly. Maps to 400 via libs-runtime, no service-local mapper. Good.
- **The `@Path`-binds-to-the-next-declaration trap.** `sanitizeForLog` is a **member** of the class, with a comment citing #3371. Correct.
- **`openbank.outbox.dispatch-enabled`.** Already `true` at `application.yaml:229`.
- **Gauge-vs-counter reasoning.** The argument that `openbank_outbox_dead_total` has zero series until first increment and resets on restart is correct, and it is the reason to use a gauge here. `registerOutboxDeadLettered` seeds the cache at `AtomicLong(0)` and the alert is `> 0`, so t=0 on a cold pod cannot fire it spuriously — the boot-zero problem is on the *other* side here (a dead tick pins it at 0 and the alert never fires), and the PR addresses that with an ADR-0237 heartbeat registered by the concrete bean. That reasoning is right and I could not fault it.
- **`inputs.file` / `inputs.files` on `tasks.test` in `openbank-libs-runtime/build.gradle.kts`.** This is the good part of the PR: without it the naming guard reports `test UP-TO-DATE` when the alert selector changes, i.e. a guard that cannot see the change it exists to catch. The `fileTree` input for `*OutboxDeadLetterGauge.kt` closes the same hole for a new binding.

### Verdict

**Needs change — but only mechanical:** rebase (fixes CI), re-take `info.version` after #6019 lands (or before, whichever), and decide the alert severity question. The code itself I have no objection to.

JiRaska:
### `openapi.yaml` `info.version: 1.8.0` is claimed by two open PRs at once

`origin/main` is currently at `1.7.0` for `openbank-card-issuance-service`, so `1.8.0` is free *today* —
but **both this PR and #6019 declare `1.8.0`**. Whichever lands second will merge cleanly (identical
text is not a conflict) and quietly ship a duplicate contract version.

Whoever merges second needs to re-read the version off live `main` and bump to `1.9.0` before merging.
Flagging on both PRs so neither side assumes the other checked.

JiRaska:
## Rebuilt on current `main`

Both conflicts were on **derived** artifacts, so neither was hand-merged: `main`'s side was taken for
`card-issuance-opa-bundle.yaml` and `payments-services.yaml`, then regenerated.

```
  wrote openbank-libs/governance/rules-opa-data.yaml (28431 bytes)
  differential proof: 3978 MCP + 3808 REST decisions identical under the full charter
  document and the derived subset
```

Both derived data files came out **byte-identical to `main`** (this PR touches no key OPA reads), and
the bundle regeneration changed exactly the two expected files: `card-issuance-opa-bundle.yaml`, and
a single line in `payments-services.yaml` — `policy-checksum: "d830e3af22c5354a"` ->
`"5018d9b7680fec7b"`. Nothing was hand-edited to dodge the diff.

### Verification

```
opa test rest.rego card_issuance_rest_ext.rego card_issuance_rest_ext_test.rego -> PASS: 14/14
  incl. test_card_outbox_requeue_allowed_for_admin
        test_card_outbox_requeue_denied_for_operator / _compliance / _ai_agent
check-opa-bundle-parses.py -> exit 0   OK: 43 bundle(s), parse-check ON, no unloadable policy
opa fmt --diff on both rego files -> exit 0, no diff

./gradlew :openbank-libs-runtime:{ktlintCheck,detekt,test}
         :openbank-card-issuance-service:{ktlintCheck,detekt,test}
         :openbank-billing-service:{ktlintCheck,detekt,test}
```

All twelve tasks executed (checked the task list, not just the summary). `:openbank-billing-service:test`
failed on `QuarkusBindException` — the documented port collision from a concurrent Gradle build, not
this change; `--rerun-tasks` on its own is `BUILD SUCCESSFUL`.

`openapi.yaml` `info.version` re-checked against live `main`: `1.7.0` there, `1.8.0` here — still free.

### For a human — merge order against #5723

**#5723 also appends a document to the end of `payments/prometheus-rules.yaml`.** Both append at EOF,
so whichever merges second conflicts textually. Trivial to resolve (keep both documents), but it does
need a hand.

JiRaska:
**This alert is a false positive, and it is not specific to this PR — the same shape is open on 11
alerts across at least five services. I have not "fixed" the code, because the code is already
mitigated and the change would be for the scanner, not for the defect.**

The flagged sink (`CardOutboxAdminResource.kt:94-100`) already sanitises the only tainted input:

```kotlin
log.warnf(
    "card.outbox.requeue operator=%s event_id=%s requeued=%d dead_remaining=%d",
    operatorId.sanitizeForLog(),        // <- the X-Operator-Id header, CR/LF stripped
    parsed?.toString() ?: "ALL",        // <- already through UUID.fromString
    requeued, remaining,                // <- Int
)

private fun String?.sanitizeForLog(): String = (this ?: "-").replace('\n', '_').replace('\r', '_')
```

CodeQL names the source as `L79C9-L79C58` — the `operatorId` parameter — and follows it to the sink
**without recognising the extension function as a barrier.**

That it is the extension function and not this call site is measurable: every other open
`java/log-injection` alert in the repo is also on a line that already calls `sanitizeForLog()`.

```
393  OpaPolicyDecisionPoint.kt:59   query.tool.sanitizeForLog(),
392  AgentPolicyGate.kt:91          decision.agent.sanitizeForLog(),
356  ScaAdapters.kt:76              ... message.sanitizeForLog())
```

11 open alerts, same rule, same idiom, five services (`agent-service` ×4, `case-coordinator-agent`
×3, `card-issuance-service` ×2, `sca-service`, …). A private `String?.sanitizeForLog()` declared per
class is the fleet convention here — it appears in at least six files on `main`.

**Why I did not just rewrite it.** Swapping `replace(Char, Char)` for a `Regex`/`replaceAll` form
might be recognised by CodeQL's barrier model, but I cannot verify that without a CodeQL run, and
doing it here would leave the other ten sites alerting while changing correct code to satisfy a
tool. The honest fixes are one shared helper CodeQL models, or a CodeQL sanitizer definition —
tracked separately.

**To unblock this PR** the thread needs resolving, or alert 461 dismissing as a false positive.
I am deliberately leaving that to you: it is a security finding, and marking one resolved is not
something I will do unilaterally, even when I believe the evidence above settles it.

Everything else on #4308 is green — all five required contexts pass; `required_review_thread_resolution`
is the only thing holding it.

JiRaska:
### `openapi.yaml` `info.version` collision: **1.8.0 is double-claimed**

Read off live `origin/main` just now (not this branch's base):

| Source | `openbank-card-issuance-service/.../openapi.yaml` `info.version` |
|---|---|
| `origin/main` | **1.7.0** |
| this PR (#4308) | **1.8.0** |
| #6019 (`fix/card-issuance-openapi-enum-drift-5962`, OPEN, mergeable `CLEAN`) | **1.8.0** |

Both open PRs bump 1.7.0 → 1.8.0. Neither conflicts textually, because each is a clean
one-line change against the same base — git sees identical text, not a taken version. So
**whichever of the two merges first takes 1.8.0, and the second one must be re-bumped to
1.9.0 before it merges.** If that re-bump does not happen, the API-contract gate on the
second PR classifies against a version that is already on `main`.

No action needed while #6019 is unmerged. The check to run immediately before merging
either PR:

```
git fetch origin main
git show origin/main:openbank-card-issuance-service/src/main/resources/openapi.yaml | grep -m1 '^  version:'
```

---

Separately, this branch was **103 commits behind `main`** with its verdict computed at
`2026-08-22T01:55Z`. I merged `origin/main` into it (merge commit, no rebase, no force-push)
so the checks recompute against current `main`. Merge audit: the result diffs against
`origin/main` in exactly this PR's own 20 files and nothing else; `.release-please-manifest.json`
58 → 58, `release-please-config.json` 59 → 59, `payments-services.yaml` 459 → 459,
`prometheus-rules.yaml` 17 → 17, `prometheus-rules-billing.yaml` 2 → 2, and
`.github/event-contract-baseline.txt` 45 → 45 — no list entry silently dropped by the merge.
No Flyway version collision in either `openbank-card-issuance-service` or
`openbank-billing-service`.

Issue #1511, opened 2026-07-17
docs(ci): identifier intent guard still false-positives on inline trailing comments (ADR-0106)


JiRaska:
## Context

`check-new-random-uuid.sh` (the ADR-0106 identifier intent guard, run by `ci.yml`'s `Validate manifests` job) previously flagged any added diff line containing the literal string `UUID.randomUUID()`, including full-line comments that merely mention the call in prose.

Reproduced live on PR #1430 (`fix/notification-stale-uuid-comment`, `openbank-notification-service`) — the guard failed with:

```
##[error]This change adds 1 bare UUID.randomUUID() call(s) in src/main (ADR-0106).
openbank-notification-service/src/main/kotlin/com/openbank/notification/application/OperatorMessageService.kt:          // via bare UUID.randomUUID() (UUIDv4), so `notifications.notification_id` is currently a
```

— even though the change was comment-only, correcting stale prose about `NotificationConsumer.dispatch`'s identifier minting. The PR author had to reword the comment (commit 19113a77f) to dodge the literal string as a workaround.

## Fixed

PR #1475 (merged) skips **full-line** comments (`^\s*//...`) in the added-diff match, so a standalone comment line no longer trips the guard.

## Follow-up (this issue)

The fix in #1475 does not cover **inline trailing comments** on real code, e.g.:

```kotlin
val x = foo() // still uses UUID.randomUUID() somewhere else
```

A line like that still matches today, because the awk match only skips lines whose *entire trimmed content* starts with `//`. This was called out as lower-priority at the time (the guard is diff-scoped to `src/main`, so the blast radius of a false negative there is low), but is worth tightening — either strip a trailing `// ...` before matching (careful not to break on `//` inside a string literal) or track it as accepted residual risk.

## Suggested owner

`.github/scripts/check-new-random-uuid.sh`

PR #4311, opened 2026-08-09
fix(ci): strip Kotlin comments in identifier-intent-guard before matching

JiRaska:
## What

`identifier-intent-guard` (gates.yaml, group `kotlin`) matched the literal text `UUID.randomUUID()` in `src/main` diffs with no notion of comments beyond a full-line `//` (#1475). On #4308 it flagged a KDoc in `openbank-card-issuance-service/.../CardOutboxAdminResource.kt` that quoted **openbank-audit-service's** own id-minting call verbatim, as the evidence for why outbox requeue cannot be automatic — a pure code-about-code false positive. It was worked around by rewording the prose (b0e6c02ac), which is the wrong direction: the file was correct and the checker was not.

Same trap `check-roles-allowed-realm.py` hit in its first draft, on the KDoc #2403 wrote to explain the annotation it had just fixed. Its header already says so: *a guard that cannot tell code from prose about code manufactures its own findings.*

## How

- Ported the gate from bash/awk to `.github/scripts/check-new-random-uuid.py` and strip comments before matching.
- **Reuses** `check-roles-allowed-realm.py`'s `strip_comments` by import — not a second copy. Kotlin block comments NEST, and that is exactly the detail a re-implementation gets wrong (a KDoc containing `/*` otherwise closes early and its tail is scanned as code). One implementation, one place to keep correct.
- `strip_comments` preserves line numbering, so a comment-free view of the file at `HEAD` is indexed directly by the line numbers the diff adds. Diff scoping, the `*/src/main/*.kt` + `:(exclude)*/domain/identifiers/Ids.kt` pathspec, the no-base skip and the exit contract are unchanged.
- Closes the inline-trailing-comment gap left open as `rules.yaml: identifier_intent_guard.known_gap` (issue-1511) — a real stripper closes all three shapes at once, and the string-literal objection recorded there applies to the regex approach it was written about.

## Falsification

The repo rule is that a guard which has only ever passed is unfalsified. Each `--self-test` case is a **real base/HEAD commit pair in a throwaway git repo** driven through the gate's own matcher — a hand-fed string cannot exercise the diff scoping, which is half the gate.

| case | expected |
|---|---|
| bare `UUID.randomUUID()` in `src/main` | **FLAGGED** |
| #4308's exact KDoc | not flagged |
| KDoc with a nested `/* … */` | not flagged |
| real call *after* such a KDoc | **FLAGGED** (stripper does not swallow the code that follows) |
| full-line `//` mention (#1430/#1475) | not flagged |
| inline trailing `//` mention (issue-1511) | not flagged |
| the `Ids.kt` wrapper itself | not flagged |
| `src/test` | not flagged |

**The harness itself was falsified**, by swapping the stripper out:

- identity (no stripping at all) — 4 cases go red, including #4308's.
- a **non-nesting** stripper — the nested-KDoc case alone goes red. That case is what makes the nesting requirement testable rather than asserted.

**Against the real repository**, not just fixtures:

- `d972607a9` (#4308's pre-reword state, base `d972607a9^`): old script exit 1 on the KDoc line, new script exit 0.
- Bases `HEAD~50` / `HEAD~200` / `HEAD~800` of `main`: old and new report **identical** violations (0, 2, 2 — both real `UUID.randomUUID()` calls in `CaseActivitiesImpl.kt`). No regression toward false negatives.
- `run-gates.py --only identifier-intent-guard` green, self-test first.
- `gates-not-deregistered`, `gate-script-registration`, `stale-comment-references`, `advisory-gate-registration`, `gate-runner-self-test`, `yamllint`, `duplicate-yaml-key-guard` all green locally.

`gen-rules-opa-data.py` produces no diff — `identifier_intent_guard` is not in the OPA read set, so no bundle restamp.

Refs #4308, #1699, issue-1511


JiRaska:
Reviewed. **This is the strongest form of the fix, and it is genuinely falsified.** Approve.

Porting `check-new-random-uuid.sh` to Python and **importing** `check-roles-allowed-realm.py`'s nesting-aware `strip_comments` rather than copying it is the right call — this repo has already paid for a second copy of a stripper drifting from the first.

## Falsification is real, and measured against both predecessors

I grafted each older stripper into the **new** self-test rather than trusting the claim:

| stripper | cases that go RED |
|---|---|
| none (raw text) | **4** — the #4308 KDoc, a nested KDoc, a full-line `//`, an inline `//` |
| non-nesting `/* */` | **1** — the nested-KDoc case |
| this PR | 0 |

So the new cases do fail against the old implementations, which is exactly what a self-test has to demonstrate and what several gates in this repo have shipped without.

```
kotlin  identifier-intent-guard   enforced always  selftest
check-new-random-uuid.py --self-test    exit 0, 8/8
run-gates.py --only identifier-intent-guard   PASS=1
```

Diff-scoping is sound for a non-obvious reason worth keeping in the header: `added_lines` walks `--unified=0` hunks and indexes `strip_comments(HEAD:path)` by diff line number, which only works because the stripper blanks characters without dropping newlines. It does, and the header says so.

## The residual is asymmetric, and the header understates one side

I probed string literals specifically. Every false **negative** is documented and benign in direction:

| input | result |
|---|---|
| `val u = url("http://x"); val id = UUID.randomUUID()` | missed — the `//` inside the literal blanks the rest of the line |
| `val glob = "/*"` then a real call | missed — the literal opens a block comment |
| `val g = "*/"` then a real call | flagged (no false positive) |
| KDoc with an unbalanced `/*` | file-wide swallow — but Kotlin would not compile that, so it matches the compiler |

But there is one **false positive** the header does not mention: a string literal *containing* the text `UUID.randomUUID()` is flagged. That is the same class as the prose false positive this PR exists to kill, one layer down — and it is the direction that costs someone a confusing red. The sibling `check-outbox-has-writer.py` carries two strippers precisely because "is this text a literal" is purpose-dependent; here the honest answer is that literals should be blanked too. Low frequency, not worth blocking, but the residual paragraph should say so rather than listing only the misses.

## Worth fixing: the workaround should not outlive the bug

#4308's `CardOutboxAdminResource` KDoc was reworded to *"a freshly generated random id per entry"* (commit `b0e6c02ac`) to get past this guard. This PR fixes the guard and does **not** restore the prose.

That leaves the repo's own documented failure mode in place: code reworded to appease a checker, with nothing recording that the checker was the thing that was wrong. The next reader learns the wrong lesson — and worse, learns it from a KDoc that now says something slightly less true than what it replaced.

Restoring the verbatim `id = UUID.randomUUID()` quote is covered by this PR's own self-test, which already pins the original text as `KDOC_4308`. Ideally the KDoc also notes *why* it reads oddly in git history.

## Notes

- `gates.yaml:1543` — "before #4308 it read a KDoc…" reads as if #4308 predates the bug. #4308 is what *hit* it. One word.
- Cross-references are clean: no surviving `check-new-random-uuid.sh` reference in `gates.yaml` or `rules.yaml`, and `known_gap: issue-1511` is removed with #1511 now cited in narrative prose only. Worth confirming the PR body closes #1511 rather than leaving it open with no gate pointing at it.

(Two gate failures in my local full-suite run were environment-only: `api-contract` dying on `sudo: a password is required` while installing oasdiff, and `PR_TITLE: unbound variable`.)

PR #4314, opened 2026-08-09
docs(fraud): record why the liveness heartbeat sits outside the catch

JiRaska:
#4286 and #4269 derived the same fraud-hold liveness heartbeat independently, and arrived at the
same behaviour — the diff between them is entirely comments and declaration order:

```
private var liveness: WorkflowLivenessRecorder? = null
fun registerLiveness(@Observes event: StartupEvent) { … registerWorkflowLiveness(…) }
liveness?.recordSuccess()
```

#4286 merged, so #4269 is closed as a duplicate. But three of its comments explain decisions that a
later edit would undo without noticing, and `main` has none of them.

**1. `recordSuccess()` is on the completed path only.** A heartbeat inside or after the `catch`
would assert the very thing it exists to disprove. This sweep swallows its exception, so a
permanently broken run is otherwise indistinguishable from a healthy quiet one — nothing escapes,
and `cleared=0` is the normal case. Moving that one line into the `catch` for tidiness turns the
gauge into a lie, and nothing would fail.

**2. `EXPECTED_INTERVAL` is the schedule, not an SLA.** The staleness rule already applies a 2x
multiplier. Widening the cron without widening this makes the gauge over-strict rather than lax —
the safe direction to be wrong in, and worth saying so before someone "fixes" the apparent mismatch.

**3. `DomainMetrics` is field-injected on purpose.** The constructor already takes 8 parameters and
detekt's `LongParameterList` fires **AT** `constructorThreshold: 9`, not above it — so promoting it
to a constructor parameter fails the gate.

Comments only, no behaviour change. `ktlintCheck` + `detekt` green on `:openbank-fraud-service`.

Refs #3345, #4269

### DIFF
--- a/openbank-fraud-service/src/main/kotlin/com/openbank/fraud/application/usecase/FraudHoldService.kt
+++ b/openbank-fraud-service/src/main/kotlin/com/openbank/fraud/application/usecase/FraudHoldService.kt
@@ -55,6 +55,10 @@ class FraudHoldService @Inject constructor(
     @ConfigProperty(name = "openbank.fraud-hold.ttl-days", defaultValue = "30")
     private val ttlDays: Long,
 ) {
+    /**
+     * Field-injected rather than a constructor parameter: the constructor already takes 8,
+     * and detekt's LongParameterList fires AT `constructorThreshold: 9`, not above it.
+     */
     @Inject
     lateinit var domainMetrics: DomainMetrics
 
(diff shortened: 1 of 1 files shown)


JiRaska:
Comment-only, and the reasoning is correct on all three points — I derived the same three independently while building the now-closed #4269 (duplicate of #4286), so this is the rationale surviving where the code did not.

Each of the three is load-bearing rather than decorative, which is worth saying because comment-only PRs get skimmed:

- **Field injection.** The constructor takes 8 and detekt's `LongParameterList` fires **at** `constructorThreshold: 9`, not above it — so a ninth constructor parameter fails the gate outright. Someone "tidying" this into the constructor later would break the build for a non-obvious reason.
- **`recordSuccess()` outside the catch.** The sharp form is in the comment already: a heartbeat on the failure path asserts the very thing it exists to disprove. Given this sweep swallows its exception and `cleared=0` is the normal hour, that is the difference between a liveness gauge and a decoration.
- **`EXPECTED_INTERVAL` vs the cron.** Naming which direction the drift is safe in (over-strict, not lax) is the part a future reader needs; the bare "matches the default" did not carry it.

No behaviour change, nothing to verify beyond that the file still compiles.

Issue #4312, opened 2026-08-09
AGENTS.md does not exist, but non-Claude agent harnesses are briefed to read it as authoritative


JiRaska:
Agents driven by non-Claude harnesses (opencode / Sisyphus) are briefed to read `AGENTS.md` as the
authoritative in-repo rule file. **That file does not exist**, on `origin/main` or anywhere in the
tree:

```
$ git cat-file -e origin/main:AGENTS.md
fatal: path 'AGENTS.md' does not exist in 'origin/main'
```

So a harness following that brief loads no fleet rules at all. Everything in root `CLAUDE.md`
(1054 lines — the release two-axis rule, the derived-tails rule, the Panache `persist` vs `merge`
trap, the `@Scheduled` Vert.x-context rule, the pact replay rules, …) is simply absent for it. It
still sees `openbank-libs/governance/rules.yaml`, which is what CI enforces, but `rules.yaml` is
the machine-checkable subset — the operational footguns live only in `CLAUDE.md`.

This is not hypothetical. Two PRs authored under that brief are merged on `main`:

| PR | scope | files |
|---|---|---|
| #3300 | `fix(agent-service)`: route model calls only through LiteLLM gateway | 4 |
| #3319 | `feat(governance)`: extend governance-auditor money-path read-only scope | 76 |

Both are correctly signed, both went through a PR rather than a direct push, and #3319 regenerated
all three `agents.yaml` derived tails (`docs/compliance/eu-ai-act.md`, the OPA bundle ConfigMaps,
`openbank-admin-ui/ai-governance-snapshot.json`) — the exact omission that red-gated `main` in
#3771. So the outcome was good. It was good without the rule file being readable, which makes it
luck rather than a procedure, and neither PR carried a review.

## Fix

`AGENTS.md` as a symlink to `CLAUDE.md`. One file, one source of truth, structurally incapable of
drifting — which a second copy of the rules would not be. `AGENTS.md` is also the emerging
cross-harness convention, so this costs nothing and covers future tooling too.

## Why not a second rules document

The repo already knows this failure shape: a published doc keeping its own copy of a list that
lives in `rules.yaml` **is** the drift (the AGPL-module count that read 12 / 10 / 4 across four
places, #2280). A parallel `AGENTS.md` with its own prose would be the same defect, aimed at the
harness least able to notice.

## Linked issues

N/A — this issue is the record.

Issue #1152, opened 2026-07-15
bug: ARC 0.9.3 pod recreated with stale job claim after cancellation (root-caused, fixed)


JiRaska:
## Summary
Live-confirmed during a 2026-07-15 CI backlog incident: when a matrix job's assignment was cancelled/superseded, the ARC controller (gha-runner-scale-set 0.9.3) recreated a NEW pod bound to the SAME stale EphemeralRunner CR/job claim rather than releasing it. The runner sat \`Running\` forever holding a phantom claim instead of picking up real queued work.

## Root cause
Upstream bug actions/actions-runner-controller#4091 ("EphemeralRunner stuck in failed state if the job it was allocated to is cancelled"), fixed by #4239/#4260, first shipped in chart 0.13.0 (2025-10-16). This repo was pinned to 0.9.3 (2024-06-25), well before the fix.

## Symptoms observed
- Same \`workflowRunId\` + job name kept reappearing on brand-new pods after \`kubectl delete pod\`.
- Survived a listener restart and a controller restart.
- Only deleting the \`EphemeralRunner\` custom resource itself (not just the pod) cleared it — the CR, not the pod, held the stale claim.

## Fix
PR incoming: bump \`arc_controller_version\` 0.9.3 → 0.14.2 (latest stable, no breaking/migration notes in 0.13.1-0.14.2's release notes).

## Also investigated and ruled out
Lowering \`arc-stuck-runner-reaper\`'s \`IDLE_THRESHOLD_SECONDS\` — the reaper's selection guard requires \`jobRequestId==0\` (never assigned a job); a phantom runner from this bug has a non-zero \`jobRequestId\` (a real, now-stale claim), so the reaper structurally cannot catch this class of stuck runner regardless of threshold. No reaper change needed once the controller is upgraded.

PR #4319, opened 2026-08-09
fix(infra): retire runners-warm and raise arc_max_runners 6 -> 12

### LINKED ISSUE #4317: FinOps: runners-warm keeps 2 on-demand nodes alive for a rationale retired 2026-06-13, and a 10% disruption budget on a 2-node pool makes them immortal (~$277/mo)
The `runners-warm` Karpenter NodePool is still provisioning **2× on-demand** Graviton nodes for a
purpose that was retired on 2026-06-13, and it is structurally unable to consolidate them.

## Measured 2026-08-09 (live cluster, `openbank-sandbox`)

```
$ kubectl get nodes -l karpenter.sh/nodepool=runners-warm
ip-10-80-6-211    2026-08-06T22:04:02Z   m6gd.xlarge   on-demand
ip-10-80-62-19    2026-08-06T01:03:22Z   m6gd.xlarge   on-demand
```

`ip-10-80-6-211` has been up 2.6 days carrying **only DaemonSets** — no runner pod at all.

## Three separate defects

**1. The rationale has lapsed.** `arc-runners.tf` justifies the pool by NAT egress: every cold node
pulls `docker:dind` (~700 MB) from `public.ecr.aws` over the NAT gateway, so two always-alive nodes
keep the containerd cache warm. `variables.tf` then records that this stopped being true:

> Lowered 2 -> 0 (2026-06-13): ECR pull-through cache (PR #926) now serves docker:dind from private
> ECR via the ecr.dkr VPC endpoint — zero NAT regardless of node warmth. Warm pool rationale gone;
> runners-warm NodePool (arc-runners.tf) also set to 0 min.

`arc_min_runners` was duly set to 0. The **NodePool** was not: it still carries `limits.cpu: 8` and
`expireAfter: Never`, so it keeps two on-demand nodes alive indefinitely.

**2. The cost comment understates the bill by ~2.5×.** It says "2× c6g.large on-demand =
~$3.72/day (fixed)". The nodes actually provisioned are `m6gd.xlarge` — the NodePool requires
≥4 vCPU, so `c6g.large` (2 vCPU) can never be chosen and the quoted figure was never achievable:

| | quoted | actual |
|---|---|---|
| instance | c6g.large | m6gd.xlarge |
| on-demand $/h (eu-north-1) | — | **$0.1920** |
| per day, 2 nodes | $3.72 | **$9.22** |
| per month | ~$113 | **~$277** |

**3. It can never consolidate, so this cannot self-heal.** The pool sets
`consolidationPolicy: WhenEmpty` with `consolidateAfter: 5m`, which should have removed the
DaemonSet-only node within minutes. It cannot, because the disruption budget is a percentage of a
two-node pool:

```yaml
disruption:
  budgets: [{ nodes: "10%" }]      # 10% of 2 nodes = 0.2 -> floors to 0
  consolidateAfter: 5m
  consolidationPolicy: WhenEmpty
```

Karpenter may disrupt **zero** nodes here, whatever the consolidation policy says. Combined with
`expireAfter: Never` the nodes are immortal by construction. This is a general trap worth stating
once: a percentage disruption budget on a pool smaller than `1/percentage` nodes silently disables
disruption, and it reads in the manifest as a normal, conservative setting.

## Fix

Retire the pool: `limits.cpu: 0` (or delete the NodePool and its `runners-warm` tier), letting the
`runners` spot pool serve the whole build path as it already does for PR builds. Spot
`m7gd.xlarge` is **$0.0566/h** against on-demand `m6gd.xlarge` at $0.1920/h — 3.4× cheaper for the
same 4 vCPU. Where a percentage budget is kept on any small pool, express it as a node count
(`nodes: "1"`) so it cannot floor to zero.

## Why it matters beyond the $277

`openbank-build` is capped at `arc_max_runners = 6` and is a day behind: 42 queued `Services CI`
runs, p50 age 7.0 h, max 23.4 h. Retiring this pool funds `arc_max_runners` 6 -> 12 several times
over — and raising concurrency does **not** raise total runner-hours, since the same work runs on
more nodes for less wall-clock. So the combined change is cheaper than today *and* roughly doubles
build throughput. ADR-0250 covers the structural half (moving the build itself onto free hosted
runners); this issue is independent of it and can land first.

## Linked issues

Refs #4312

JiRaska:
Two changes to the `openbank-build` CI runner capacity, in one PR because they are one FinOps
trade: retire an on-demand pool whose justification lapsed, and spend part of the saving on
concurrency that is now actually useful.

## Why now

Measured 2026-08-09 against the live cluster and the Actions API:

| | |
|---|---|
| queued `Services CI` runs | 42 |
| queue age | p50 **7.0 h**, max **23.4 h** |
| per-build queue vs execution | 85 min p50 vs 4–11 min |
| runner pods executing a real job | **6 of 6** |

The pool is genuinely saturated. Every runner pod was confirmed to be doing real work (`WORKER`
frames in the runner container log) while the ARC listener reported
`"assigned job"=6 decision=6 min=0 max=6`.

## `arc_max_runners` 6 → 12

The 2026-07-15 revert from 12 back to 6 is recorded in `variables.tf` and rested on a specific
premise: a Runner Scale Set bug meant extra slots only held more phantom claims, so a bigger pool
bought no throughput. **That premise is gone.** #1152 root-caused it to ARC 0.9.3 and closed it by
bumping `arc_controller_version` to 0.14.2 — which is what runs
(`ghcr.io/actions/gha-runner-scale-set-controller:0.14.2`). The runner-log check above is what
distinguishes "saturated" from "holding phantoms", and it says saturated.

FinOps: **close to neutral.** Raising concurrency does not raise total runner-hours — the same
~17.4 runner-h/day of build work runs on more nodes for less wall-clock. The only delta is
per-node cold-start overhead, and the ECR pull-through cache already made that cheap. Bounded by
the `runners` NodePool cpu limit of 64: 12 × 4 vCPU = 48, with warm now at 0.

## `runners-warm` retired (`limits.cpu` 8 → 0), #4317

Its rationale — keeping a `docker:dind` containerd cache warm so cold nodes do not pull ~700 MB
over the NAT gateway — was retired on **2026-06-13**, when PR #926 put the image behind the ECR
pull-through cache over the `ecr.dkr` VPC endpoint. `variables.tf` recorded that and set
`arc_min_runners = 0` the same day. The NodePool was left at `limits.cpu: 8` with
`expireAfter: Never`, so it kept provisioning on-demand nodes for a reason that no longer existed.

| | on-demand m6gd.xlarge (today) | spot m7gd.xlarge (`runners`) |
|---|---|---|
| $/h | 0.1920 | 0.0566 |
| 2 nodes, 730 h | ~$280/mo | ~$83/mo |

**~$197/month for identical throughput.** The old cost comment was wrong in both directions —
it quoted `c6g.large` at ~$3.72/day, and `c6g.large` has 2 vCPU, which the pool's own ≥4 vCPU
requirement makes unselectable.

## What I got wrong, since it is in the issue history

#4317 originally claimed these nodes were immortal because a `10%` disruption budget on a two-node
pool floors to zero. Wrong twice: the budget is Karpenter's **default**, not authored here, and
Karpenter scales percentage budgets with round-up (1 of 2, not 0) — it is observably disrupting
nodes on this cluster one at a time. The nodes persist because they are **busy**: runner pods carry
`karpenter.sh/do-not-disrupt: true` and the backlog keeps the pool occupied, so `WhenEmpty` / `5m`
never fires. Corrected in the issue and in ADR-0250. This is a rate change on work that is really
being done, not reclaimed idle capacity.

## Risk

Retiring the warm tier removes on-demand from the main-push build path, so a **spot interruption
can now kill a main-push build mid-job**. Every PR build already carries that exposure. If it turns
out to cost more than $197/month in retries, the pool is revived by raising `limits.cpu` — which is
why the NodePool is left declared rather than deleted.

`tofu fmt -check -recursive` clean. **Not applied** — this needs a reviewed `tofu apply` against
`sandbox-platform`, and the two existing warm nodes drain as their current runner pods finish
rather than being killed.

## Linked issues

Closes #4317
Refs #4312


JiRaska:
`adr-registry-integrity-check` went red on the first push, correctly: `variables.tf` cites
ADR-0250 in a comment, and that ADR lives in #4318, which has not landed. The registry gate
resolves every `ADR-NNNN` reference against `docs/adr/` **on the branch**, so a reference to a
sibling PR's file is indistinguishable from a reference to an ADR nobody wrote.

Resolved the documented way — merged the ADR branch into this one rather than waiting, so this PR's
CI is green independently and the ADR files drop out of the diff as #4318 lands. Merge scope
verified rather than assumed (a clean merge can silently drop a list entry): the merge brought in
exactly the four ADR files and nothing else, and `index.json` still indexes 243 ADRs.

`run-gates.py --group registry` now passes 22/22 locally with `BASE_REF` / `PR_DIFF_BASE` set —
without those two the harness reports a false red.

PR #4325, opened 2026-08-09
test(account): follow the workflow-liveness boot seed, and stop asserting it by age

JiRaska:
## `main` is red, and no check says so

`origin/main` fails this test right now, with no PR involved:

```
SavingsProposalExpiryWorkflowLivenessTest > registers the gauges at startup and records success after a sweep() FAILED
    Expecting actual: 0.0 to be greater than: 1.5768E9
SavingsProposalExpiryWorkflowLivenessTest > a swallowed sweep failure leaves the heartbeat stale() FAILED
3 tests completed, 2 failed
```

Reproduced in a clean detached worktree at `origin/main`, no local changes.

#4208 changed `registerWorkflowLiveness` to seed the age gauge **at registration** rather than at
`Instant.EPOCH`, and updated the fleet's liveness tests to match — **17 of 18**. This is the 18th.
It was missed because it landed in #4180, after #4208 was authored.

**Why nothing caught it:** CI is path-scoped, and #4208 touched no file under
`openbank-account-service/`, so this module has not been rebuilt since the libs change. Its last
reported `build (openbank-account-service)` success predates it. The failure only surfaced because
an unrelated PR (#4308) edited `libs-runtime` and thereby pulled account-service into *its* changed
set — by accident, not by design.

## The third test was passing and is changed anyway

`a sweep that expires nothing still records success` asserted only `age < 5s` after the sweep. Under
the boot seed the age is **already** under 5s at registration, before `sweep()` is called — so the
assertion holds against a scheduler that records nothing at all. It had become vacuous the moment
#4208 merged, and it would have stayed green forever.

Measured, rather than argued:

| scheduler | result |
|---|---|
| `.onSuccess { liveness?.recordSuccess() }` replaced by `.onSuccess { }` | **3 tests completed, 1 failed** |
| restored | **BUILD SUCCESSFUL** |

The failing one is that third test — under its old assertion it would not have failed.

## The change

Exactly what #4208 did to the other seventeen: age must be **under** a one-hour boot-seed ceiling at
registration, and "has this ever succeeded" is read from `openbank_workflow_success_recorded`
instead of being inferred from an epoch-sized age. Test-only; no production code touched.

`:openbank-account-service:ktlintCheck` and `:detekt` both SUCCESSFUL, both tasks present in the
output.

Refs #2239, #3345, #4208

### DIFF
--- a/openbank-account-service/src/test/kotlin/com/openbank/account/infrastructure/scheduler/SavingsProposalExpiryWorkflowLivenessTest.kt
+++ b/openbank-account-service/src/test/kotlin/com/openbank/account/infrastructure/scheduler/SavingsProposalExpiryWorkflowLivenessTest.kt
@@ -43,6 +43,12 @@ class SavingsProposalExpiryWorkflowLivenessTest {
         .gauge()
         ?.value()
 
+    private fun successRecordedOf(registry: MeterRegistry): Double? = registry
+        .find(WorkflowLivenessMetrics.SUCCESS_RECORDED)
+        .tag(WorkflowLivenessMetrics.WORKFLOW_TAG, WORKFLOW)
+        .gauge()
+        ?.value()
+
     @Test
     fun `registers the gauges at startup and records success after a sweep`(): Unit = runBlocking {
         val registry = SimpleMeterRegistry()
@@ -52,9 +58,13 @@ class SavingsProposalExpiryWorkflowLivenessTest {
         val scheduler = SavingsProposalExpiryScheduler(service, metricsOver(registry))
         scheduler.registerLiveness(StartupEvent())
 
-        // Registered but never succeeded: the age gauge starts at the epoch, which is the
-        // "absent-vs-stale" distinction ADR-0160 mechanism 3 relies on.
-        assertThat(ageOf(registry)).isGreaterThan(FIFTY_YEARS_SECONDS)
+        // Registered but never succeeded. The age gauge is SEEDED AT REGISTRATION (#4208), so a
+        // never-run job reads as old as its pod rather than as decades — the "absent-vs-stale"
+        // distinction ADR-0160 mechanism 3 relies on now lives in the success flag, not in the age.
+        assertThat(ageOf(registry))
+            .describedAs("the age gauge must be seeded at registration, not at Instant.EPOCH")
+            .isLessThan(BOOT_SEED_CEILING_SECONDS)
+        assertThat(successRecordedOf(registry)).isEqualTo(NOT_YET_SUCCEEDED)
         assertThat(
             registry.find(WorkflowLivenessMetrics.EXPECTED_INTERVAL_SECONDS)
                 .tag(WorkflowLivenessMetrics.WORKFLOW_TAG, WORKFLOW)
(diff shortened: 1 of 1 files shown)


JiRaska:
This is not a test tidy-up — **it repairs a break that `main`'s CI has not reported yet.**

#4208 merged at 12:23:39Z and changed the seed in `DomainMetrics.registerWorkflowLiveness`:

```kotlin
// origin/main:394
val lastSuccessEpochMillis = AtomicLong(java.time.Instant.now().toEpochMilli())
```

It was `Instant.EPOCH`. So on current `main`, `SavingsProposalExpiryWorkflowLivenessTest:57` and `:82` still assert `ageOf(registry) > FIFTY_YEARS_SECONDS` against a gauge that now reads ~0.

The reason nobody has seen it go red is timing, not luck: the fleet rebuild triggered by that libs change is **still in progress** — `build (openbank-account-service)` on merge commit `73d5f0dc4` is `in_progress` as I write this. And the `Services CI` run that has completed on `main` since was for a gitops-only commit with `build: skipped`, so it says nothing about this. That is the documented shape where a libs change breaks one dependent and path-scoped CI reports nothing about the others.

## The judgement in the PR is the right one

The valuable part is not swapping the assertion — it is noticing that under the boot seed the old one became **vacuous rather than merely wrong**:

> *the boot seed already puts the age under the tolerance before `sweep()` is called, so it would hold against a scheduler that recorded nothing*

An age assertion after the seed passes whether or not the sweep did anything. Moving to `SUCCESS_RECORDED` restores a claim that can distinguish "registered" from "ran", which is exactly what ADR-0160 mechanism 3 needs, and the `describedAs` messages say which distinction each assertion is carrying.

Worth stating explicitly in the PR body, because it changes who should care: this is a **blocking fix for a latent red**, not a refactor, and it should merge before the fleet rebuild finishes rather than after someone triages the failure.

One thing I did not verify: I did not run the account-service suite myself, so "will go red" is derived from reading the seed change against the assertion, not measured. The in-progress job will settle it either way.

JiRaska:
**Confirmed — `main` is red, and this PR is the fix.** Upgrading my earlier comment from "derived by reading" to measured.

`build (openbank-account-service)` on #4208's merge commit `73d5f0dc4` has finished:

```
FAILED: Build + test (:openbank-account-service)

SavingsProposalExpiryWorkflowLivenessTest > registers the gauges at startup and
  records success after a sweep() FAILED
    java.lang.AssertionError at SavingsProposalExpiryWorkflowLivenessTest.kt:47
SavingsProposalExpiryWorkflowLivenessTest > a swallowed sweep failure leaves the
  heartbeat stale() FAILED
    java.lang.AssertionError at SavingsProposalExpiryWorkflowLivenessTest.kt:70
```

Both are the `isGreaterThan(FIFTY_YEARS_SECONDS)` assertions, against a gauge #4208 now seeds with `Instant.now()`.

This should merge promptly rather than queue. A red push-triggered run on `main` is the one signal in this repo with no reader — no PR carries it, no reviewer sees it, and the next PR opened against that commit inherits a failure it did not cause.

Worth noting how narrowly this was missed: the `Services CI` run that completed on `main` after #4208 was for a gitops-only commit with `build: skipped`, so it reported success while saying nothing about account-service. The fleet rebuild that would tell the truth was still in flight. That is the documented shape — a libs change breaks one dependent, and path-scoped CI is silent about the rest.

The substance of the PR still stands on its own: the old assertion did not merely become wrong, it became **vacuous** — under the boot seed it would hold against a scheduler that recorded nothing. Asserting `SUCCESS_RECORDED` restores a claim that can tell "registered" from "ran".

PR #4332, opened 2026-08-09
ci(probe): THROWAWAY — does CLAUDE_CODE_OAUTH_TOKEN still authenticate? (do not merge)

JiRaska:
**Do not merge. Do not review.** Throwaway probe, deleted as soon as it answers.

#2161 recorded `CLAUDE_CODE_OAUTH_TOKEN` as rejected at the gateway (`is_error:true` at `num_turns:1`, ~2s). The secret has not been rotated since 2026-07-06, so that measurement may still hold — and restoring independent review would otherwise be built on a credential nobody re-measured.

This asks the real gateway with the real CI credential. The verdict step reads the transcript rather than the action's exit code, because the action exits 0 on its own recorded failure (that is the #2161 finding). A missing transcript fails too: unknown is not alive.

### DIFF
--- a/.github/workflows/probe-claude-token.yml
+++ b/.github/workflows/probe-claude-token.yml
@@ -4,7 +4,8 @@
 # real gateway with the real CI credential rather than assuming either way. Delete after.
 name: probe claude token
 on:
-  workflow_dispatch:
+  push:
+    branches: [ci/probe-claude-token]
 permissions:
   contents: read
 jobs:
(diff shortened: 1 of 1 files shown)


JiRaska:
Answered: `401 OAuth access token is invalid`. The token is dead, confirming #2161 — measured via the CLI, since `claude-code-action` is blocked earlier by a missing Claude Code GitHub App install. Throwaway probe, closing.

PR #4333, opened 2026-08-09
ci(probe): THROWAWAY — re-measure the rotated Claude token (do not merge)

JiRaska:
Do not merge. Throwaway. The pre-rotation token answered `401 OAuth access token is invalid`; the secret was re-set at 14:56Z and this re-asks the real gateway with the real CI credential. Verdict is read from the model's reply text, not the exit code.


JiRaska:
Answered: TOKEN VERDICT ALIVE (108-char token, model replied). Throwaway probe, closing.

PR #4350, opened 2026-08-09
docs: split root CLAUDE.md by tree — .github/ detail moves to .github/CLAUDE.md

JiRaska:
Root was 1082 lines; one section and its three siblings were 419 of them, read in full by every agent starting work anywhere in the repo — including the ~86% of PRs that never touch `.github/`.

Split by **where the lesson fires**, not by which heading it sat under. That is why this is not a bulk move: **11 bullets filed under "CI gates" fire from outside that tree** and stay in root (`rules.yaml` is never linted by CI · regenerate the OPA bundles last · a gate over `application.yaml` says nothing about the gitops env · a third-party URL is unfalsifiable by every test layer · `gh pr merge` refuses on UNSTABLE · plus the universal ones). The other 30 load in `.github/`.

Root **1082 → 778**, `.github/CLAUDE.md` 326. Same precedent and same rule as `openbank-infra/CLAUDE.md`.

### Verified, not asserted
- **101 bullets before, 71 + 30 = 101 after, 0 lost.**
- The union of both files against `origin/main:CLAUDE.md` adds **exactly 9 lines** — the navigation prose in this PR. Otherwise a pure move, nothing sourced from elsewhere.
- No IPs, internal hostnames, tokens, AWS account ids, emails or personal identifiers in either file. The single gitleaks hit in the tree is a **pre-existing false positive** in `agents.yaml` (the rule matches the comment `# api.groq.com`), in a file this PR does not touch.
- gates: `lint 15/15`, `registry 22/22`.

The line count is the consequence of the rule, not the goal.

### DIFF
--- a/.github/CLAUDE.md
+++ b/.github/CLAUDE.md
@@ -1,0 +1,325 @@
+# `.github/` — CI, gates and workflow pitfalls
+
+> Path-scoped: this loads when you touch `.github/`. It holds the lessons that fire
+> INSIDE this tree. Anything that fires fleet-wide, or from outside it (editing
+> `rules.yaml`, an `application.yaml`, or merging a PR), stays in the root `CLAUDE.md`
+> — the split is by WHERE THE LESSON FIRES, never by which heading it sat under.
+
+
+### CI gates — exercise the failure path before trusting the green
+
+- **Gates are DECLARED in [`.github/gates/gates.yaml`](.github/gates/gates.yaml), not written as
+  workflow steps.** Add an entry (`id`/`group`/`mode`/`selftest`/`run`) and it runs; there is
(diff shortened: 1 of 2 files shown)


JiRaska:
Closing as superseded by #4406, a rebuild against current `main`.

This branch's base predated three other merged PRs that had already trimmed the exact bullets this split touches (#4370, #4208, #4351), which turned a straight merge into an 87-line conflict — mostly noise, but risky to hand-resolve since a wrong pick would silently revert one of those fixes. #4406 re-derives the same transformation fresh from current `main` and verifies the result is byte-identical to main's content once reassembled — nothing lost, nothing duplicated.

Issue #4348, opened 2026-08-09
Synthetic customer fleet and journey-based production assurance (ADR-0252) — phases 1-4


JiRaska:
## Why

Two customer-visible failures were discovered **by customers in production**, not by us:

1. **APNs push worked on exactly one device.** The platform could not tell the difference,
   because `NotificationConsumer` counts a push as `delivered` when APNs answers `200` —
   which means *Apple accepted it*, not that a device received it. APNs issues no delivery
   receipt, so this is unobservable from the server by construction.
2. **A domestic payment between two customers failed** and nothing alerted.

Neither is a coding slip. The platform has no periodic, multi-step, production-side check
of a customer journey:

- `openbank-infra/gitops/components/observability/probes-synthetic.yaml` — 4 blackbox
  targets, single-request liveness of the public ingress hosts. "Is the front door open?"
- `openbank-infra/gitops/components/observability/k6-synthetic.yaml` — a scripted journey
  that runs `iterations: 1` **once on sync**. Its own header says making it periodic is the
  documented follow-up.
- `openbank-notification-service/src/main` — one class uses `meterRegistry`
  (`ContactGateProducer`). The PUSH fan-out emits no metric at all, so there is nothing to
  alert on.
- `ApnsPushSender` is `enabled=false` by default and returns `PushResult.skipped(...)`,
  which is a *successful* result. A disabled adapter is indistinguishable from a working one.

ADR-0252 decides the shape of the fix. This issue tracks the unbuilt phases.

## Phase 0 — shipped with ADR-0252

Push-delivery observability in `openbank-notification-service`: per-outcome counters,
`skipped` separated from `sent`, `SKIPPED` no longer recorded as `SENT`.

## Phase 1 — synthetic personas + taint propagation

- [ ] `SYNTHETIC` party classification; bank-owned canary personas in production (no personal data, so no data subject exists — GDPR minimisation solved at the root rather than by an after-the-fact exclusion)
- [ ] Taint carried end to end: Kafka header `x-openbank-synthetic`, OTel baggage `openbank.synthetic`, dedicated GL accounts netting to zero
- [ ] Taint **excludes** from FINREP / COREP / AnaCredit / statistical aggregates and from AML baseline scoring — and **must not** exclude from any control path (sanctions screening, VoP, SCA, limits). A canary that skips the controls proves nothing.
- [ ] Gate: a tainted event reaching a regulatory aggregate is a failure; a control path that sees the taint and short-circuits is a failure

[...]

PR #4371, opened 2026-08-09
feat(libs): ADR-0252 phase 1 — the synthetic taint primitive

JiRaska:
ADR-0252 phase 1, first slice. Additive, unwired, no service touches it yet — everything else in
phase 1 (canary personas, the ledger dimension, exclusion from the regulatory aggregates) rests on
this, so it lands on its own where it can be argued about in isolation.

## What is actually being decided

The file is a header name, a baggage key and a four-line parse. The load-bearing part is the
**default**, and it is not symmetric:

- **Real read as synthetic** ⇒ dropped from FINREP / COREP / AnaCredit and the AML baseline. A
  regulatory return that silently omits real customer money. Unbounded and invisible.
- **Synthetic read as real** ⇒ canary movements land in those aggregates. Wrong, but visible, and
  bounded by how much the canaries do — which the platform sets.

So `isTainted()` reads anything that is not an exact, case-insensitive `true` as REAL. `1`, `yes`,
`on` are **rejected** rather than accepted for the same reason: a permissive parser turns any stray
value in a header into a suppression.

The header lookup is case-insensitive because casing survives no transport reliably — Kafka
preserves it, HTTP/2 lower-cases it, a bridge between them may do either. A case-sensitive lookup
would drop the taint at exactly one hop and nowhere else, which is the failure mode this whole ADR
is about.

It is a **header**, not a payload field: a taint in the body can only be read by a consumer that
already deserialises that body, while every consumer, bridge and dead-letter tool can read a header.

## The asymmetry, encoded rather than described

Two rules pointing in opposite directions, and the second is the one that gets built backwards:

- regulatory aggregates and analytics baselines **exclude** tainted activity;
- control paths — sanctions screening, Verification of Payee, SCA, limits, fraud scoring — **always
  include** it. A canary that skips the controls proves nothing about them, and excluding synthetic
  payments from screening would manufacture a screening blind spot: a worse defect than the one the
  fleet exists to catch.

So there is deliberately **no bypass-shaped helper**, and a test pins that absence by reflecting over
the object's methods. A future addition has to delete an assertion and explain itself. The test
inspects the method list rather than the file text, so the word "bypass" appearing in the KDoc that
explains the rule cannot satisfy it — the check-matches-its-own-comment trap (#3072).

## Verification

- `:openbank-libs-domain:test ktlintCheck detekt` green. Verdict read from the test XML: 7 tests,
  **0 failures, 0 skipped**.
- **The API-surface test is falsified, not assumed:** adding a `mayBypassControl` method makes the
  suite exit 1; removing it exits 0. Both measured.
- The default is tested against the values a permissive parser would have accepted (`1`, `yes`, `on`,
  `TRUE!`) rather than only against `true`/`false`, since that is where the decision lives.

## Scope

Nothing consumes this yet. Wiring it into producers and consumers, the ledger dimension, and the
exclusion in the regulatory aggregates are separate PRs — several of them money-path, so they need
two human approvals by rule. Tracked in #4348.

Refs #4348, ADR-0252.

### DIFF
--- a/openbank-libs-domain/src/main/kotlin/com/openbank/libs/synthetic/SyntheticTaint.kt
+++ b/openbank-libs-domain/src/main/kotlin/com/openbank/libs/synthetic/SyntheticTaint.kt
@@ -1,0 +1,103 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) OpenBank contributors. Licensed under the Apache License, Version 2.0.
+// See LICENSE in the repository root or https://www.apache.org/licenses/LICENSE-2.0 for details.
+
+package com.openbank.libs.synthetic
+
+/**
+ * The synthetic taint (ADR-0252 phase 1, issue #4348).
+ *
+ * ONE flag, carried end to end, marking activity produced by a bank-owned synthetic customer
+ * rather than a real one. It is the primitive the whole synthetic-customer fleet rests on, and
+ * it is deliberately tiny: a header name, a baggage key, and a parse with one strongly-argued
+ * default.
+ *
+ * ## Why a taint at all
+ *
+ * A synthetic customer that does not touch the real code path proves nothing about the real
+ * code path. So the canary's payment IS a real payment: it hits the ledger, the screening
+ * engine and the event streams. That is the point, and it is also the risk — those movements
+ * must never reach a regulatory aggregate. The taint is what lets both be true at once.
+ *
+ * ## The asymmetry, which is easy to get backwards
+ *
+ * Two rules, and they point in opposite directions:
+ *
+ *  - **Regulatory aggregates and analytics baselines must EXCLUDE tainted activity.** FINREP,
+ *    COREP, AnaCredit, statistical returns, AML baseline scoring. A synthetic movement in a
+ *    regulatory return is a misstatement.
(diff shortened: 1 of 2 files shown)


JiRaska:
Reviewed. No blocking or worth-fixing findings — every claim in the PR body verified by running, not by reading.

- **Genuinely unwired**: grepped the whole tree for `SyntheticTaint` outside this PR's own two files — nothing.
- **Domain purity (ADR-0002)**: zero imports in the file; `domain-purity-gate` PASS, 0 new violations.
- **Tests are real, not scaffolding**: 7 tests, 0 skipped, 0 failures (read from JUnit XML). They assert the asymmetric default (fail-to-real, never fail-to-synthetic) against specific values a permissive parser would wrongly accept (`1`, `yes`, `on`, `TRUE!`), case-insensitive header lookup (motivated by real Kafka-vs-HTTP/2 casing differences), and pin the *absence* of any bypass/skip-shaped method via reflection over the class's own methods — a falsifiable API-surface test, not a tautology.
- `ktlintCheck`/`detekt` clean.

The minimal API (two constants + a boolean parse, no ledger dimension or persona marker yet) is consistent with an honest first slice — nothing here looks like it forces a breaking rework once the next phase-1 pieces land, since those are separate concepts rather than extensions of this file.

PR #4375, opened 2026-08-09
fix(jaxrs-params): answer 400, not 500, for a missing required query/header parameter

### LINKED ISSUE #3624: 39 non-money-path handlers still answer 500 for a missing required query/header parameter

JiRaska:
## What

Fixes the last seven non-nullable JAX-RS parameter declarations in the fleet and **empties the checker's baseline**.

`Closes #3624`

The issue title says 39; the thread had already narrowed it, and `check-nonnull-jaxrs-params.py --root . --list` on `origin/main` reports **7 findings, 7 baseline keys**. That is the set fixed here.

## Why the obvious fix is wrong

A non-nullable `@QueryParam`/`@HeaderParam` cannot keep its promise. JAX-RS injects `null` for an absent parameter; Kotlin's null-safety is compile-time only, so the declared type only decides *where* the failure lands, and every landing is a 500 (`GenericExceptionMapper`).

Measured with `javap`: a plain `fun` emits `Intrinsics.checkNotNullParameter` at **offset 0**, so a `requireNotNull` written in the body is **dead code** — the NPE already threw. **All seven findings here are `suspend` handlers, which emit no intrinsic at all**, so the null flowed into the body and failed at the first dereference; where nothing dereferenced it, the request proceeded with a null the signature promised could not exist.

## The seven findings, and what each turned out to be

Every one is a real inbound resource class — **none** is an outbound REST-client `interface` (where the identical annotation is *not* a defect, because the caller supplies the argument and it is compile-time checked). That distinction is half of every naive grep, so each was read before it was changed.

| # | Site | Param | What it actually was | Fix |
|---|---|---|---|---|
| 1 | `openbank-aml-service` `AmlCaseResource:50` `POST /api/v1/aml/cases` | `Idempotency-Key` | `suspend`; body's `require(idempotencyKey.isNotBlank())` NPE'd on the absent header — **the exact case the guard was written for answered 500, and only a blank one answered 400** | nullable + `requireNotNull`, existing blank-check kept |
| 2 | `openbank-party-service` `PartyResource:129` `POST /api/v1/parties` | `Idempotency-Key` | `suspend`, and **nothing dereferenced it** — the null flowed straight into `req.toCommand(idempotencyKey)` and became the command's idempotency key. The silent variant, worse than the 500 | nullable + `requireNotNull` |
| 3 | `openbank-pid-service` `PartyResource:217` `GET /pid/resolve` | `index` | `suspend`; NPE at `index.isBlank()` — the handler's own documented `INVALID_PARAMETER` 400 envelope was unreachable for an absent parameter | nullable, guard **widened** to `isNullOrBlank()` so the service's own error envelope is preserved rather than replaced |
| 4 | `openbank-pid-service` `PartyResource:282` `GET /by-external-id` | `type` (`ExternalIdType` enum) | `suspend`; null enum passed into the use case | nullable + `requireNotNull` |
| 5 | `openbank-pid-service` `PartyResource:283` `GET /by-external-id` | `value` | same handler, same shape | nullable + `requireNotNull` |
| 6 | `openbank-tpp-registry-service` `TppRegistryResource:41` `GET /check` | `tppId` | `suspend`; null reached the query object | nullable + `requireNotNull` |
| 7 | `openbank-tpp-registry-service` `TppRegistryResource:41` `GET /check` | `role` | `suspend`; NPE at `role.uppercase()` → 500 | nullable + `requireNotNull` |

`libs-runtime` maps `IllegalArgumentException` to 400. **No service-local exception mapper is added** (#526).

**None of the four services is on `rules.yaml: money_path_services`** (aml, party, pid, tpp-registry are all absent from that list), so this does not need the two-approval money-path rule. The money-path half of this defect class was already fixed in #3625/#3104.

## Proof it is not vacuous — before/after over real HTTP

A unit test that calls `TppRegistryResource.checkAuthorization(...)` directly **cannot see this defect class at all**: it supplies the argument itself, compile-time checked, so the absent-parameter case is unrepresentable. Only real HTTP can. New `TppRegistryMissingQueryParamIT` (`@QuarkusTest` + RestAssured + `@TestSecurity`, on the existing Testcontainers Postgres/Valkey resource) drives `GET /api/v1/tpp-registry/check` with each parameter missing.

Both runs are the **same test file** against the **same test resources** — only the resource-class declaration differs.

**BEFORE** — `TppRegistryResource` at its `origin/main` shape (`tppId: String`, `role: String`), new IT applied:

```
> Task :openbank-tpp-registry-service:test
3 tests completed, 3 failed
BUILD FAILED in 3m 25s
```

JUnit XML (read from the XML, not the console — a module whose Quarkus boot fails reports every `@QuarkusTest` as SKIPPED and scans as a pass):

```
TEST-...TppRegistryMissingQueryParamIT.xml
  tests=3  skipped=0  failures=3  errors=0
  AssertionError: Expected status code <400> but was <500>.   (an absent tppId)
  AssertionError: Expected status code <400> but was <500>.   (an absent role)
  AssertionError: Expected status code <400> but was <500>.   (both absent)
```

**AFTER** — parameters nullable + `requireNotNull`, same test file, same test resources:

```
> Task :openbank-tpp-registry-service:test
BUILD SUCCESSFUL in 1m 54s

TEST-...TppRegistryMissingQueryParamIT.xml
  tests=3  skipped=0  failures=0  errors=0
  ✓ an absent tppId answers 400, not 500
  ✓ an absent role answers 400, not 500
  ✓ both parameters absent answers 400, not 500
```

`skipped=0` on both sides, so neither run is the silent-skip false pass.

## Local gate

`ktlintCheck` + `detekt` on all four touched modules — reading which tasks actually appeared, since Gradle stops at the first failure:

```
:openbank-aml-service:{ktlintCheck,detekt}          :openbank-party-service:{ktlintCheck,detekt}
:openbank-pid-service:{ktlintCheck,detekt}          :openbank-tpp-registry-service:{ktlintCheck,detekt}
BUILD SUCCESSFUL in 41s — 44 actionable tasks: 30 executed, 5 from cache, 9 up-to-date
```

## Gate: baseline emptied, and falsified

`BASELINE` in `.github/scripts/check-nonnull-jaxrs-params.py` is now an empty set. A gate whose baseline still lists a fixed file fails on the stale entry, so this had to move with the fix.

```
$ python3 .github/scripts/check-nonnull-jaxrs-params.py --root . --list
0 finding(s), 0 baseline key(s)

$ BASE_REF=main PR_DIFF_BASE=$(git merge-base HEAD origin/main) \
    python3 .github/scripts/run-gates.py --only nonnull-jaxrs-param-ratchet
self-test: 16 passed, 0 failed
--- gate ---
non-nullable JAX-RS params: OK — 0 baselined occurrence(s), no new ones.
==============================================================================
1 gates  PASS=1   cpu=1.4s  slowest=1.4s  jobs=8
```

An empty baseline that has only ever passed would be unfalsified, so the gate was fed an input it **must** flag — reverting `openbank-pid-service/PartyResource.kt` to its `origin/main` shape:

```
::error file=.../pid/.../PartyResource.kt,line=217::@QueryParam("index") is declared index: String — non-nullable ...
::error file=.../pid/.../PartyResource.kt,line=282::@QueryParam("type") is declared type: ExternalIdType — ...
::error file=.../pid/.../PartyResource.kt,line=283::@QueryParam("value") is declared value: String — ...

3 new occurrence(s), 0 stale baseline entr(ies).   exit=1
```

Restoring the fix returns it to `exit=0`. The checker reports a **stale baseline entry** as well as a new occurrence, so the list cannot rot in either direction — which is why the comment above the now-empty set says an entry added there needs a reason and an issue.

## Risk

Behaviour changes only for requests that were already failing: a request that previously supplied every parameter is untouched. Callers that were relying on a 500 do not exist — a 500 is not a contract. pid's `/pid/resolve` keeps its exact documented `INVALID_PARAMETER` body; the absent case simply now reaches it instead of NPE'ing before it.


JiRaska:
Reviewed. No blocking findings — this is exactly the remaining work #4249 and #4211 left on this gate, consistent with both.

**Consistency checked against origin/main's current file**, not assumed: the removed 7 baseline entries are exactly the 7 that survived #4211's earlier cleanup. No touch to `DECL_RE`/`enclosing_type`/the self-test, so no interaction with #4249's fix. Gate self-test 16/16, real scan `OK — 0 baselined, 0 new` — the removals match the fixes exactly.

**All 4 sites are `suspend fun`**, so the guards are reachable (not the dead-code trap a plain `fun` would hit). Checked each individually rather than assuming the pattern held everywhere.

**Falsified, not just read**: reverted `TppRegistryResource.checkAuthorization` to non-nullable in a worktree, ran `TppRegistryMissingQueryParamIT` — 3 tests failed with the exact predicted NPEs. Restored, re-ran — 3/3 pass (confirmed from JUnit XML on the second run, after the first hit this machine's documented `QuarkusBindException` port-conflict flake from a concurrent Gradle process).

**Exception mapper confirmed**: `CommonExceptionMappers.IllegalArgumentExceptionMapper` → 400 exists in libs-runtime; checked all 4 services for a service-local duplicate — only tpp-registry has its own mapper file, and it explicitly excludes `IllegalArgumentException` with a comment citing #526's random-selection hazard.

**Worth-fixing, cosmetic:** `AmlCaseResource.kt:51` now carries both `requireNotNull(idempotencyKey)` and the pre-existing `require(idempotencyKey.isNotBlank())` back to back — functionally fine, but could collapse to `require(!idempotencyKey.isNullOrBlank())` like pid's `resolveByIndex`, for consistency with the widening pattern used elsewhere in this same PR.

**Note:** only tpp-registry got a red→green falsification test; aml/party/pid's fixes are correct by reading but weren't verified by breaking-and-fixing the code, unlike tpp-registry. Worth deciding whether repo rule 5 ("test the new behavior") wants one per endpoint here.

PR #4378, opened 2026-08-09
fix(kyc): write KYC case events to the outbox, retire the direct emitter

JiRaska:
## What

`kyc_outbox` shipped complete — a Flyway migration, `KycOutboxDispatcher`, a backlog gauge,
`openbank.outbox.dispatch-enabled: true`, an atomic `claimProcessable`, and even an unused
`persistInTransaction` on the port — and **nothing in `src/main` ever wrote a row to it**. Every KYC
lifecycle event went out through `KycEventPublisher`, a bare `@Channel("kyc-events-out")` emitter
called after the repository transaction had already committed.

That is a dual write. An emit that fails after the commit loses the event with no record; an emit
that succeeds before a rollback leaves a consumer acting on something that did not happen.
`KYC_CASE_APPROVED` is the expensive one — it is what activates a party downstream.

`KycEvents` (domain) now builds the envelope; `KycCaseRepository.save`/`update` gained
event-carrying overloads whose impl chains `persistInTransaction` inside the **same**
`Panache.withTransaction` block as the state change. The direct emitter and its channel are removed
— `kyc-outbox-out` publishes to the same topic (`openbank.kyc.events`), so consumers see the same
bytes plus the additive `OutboxKafkaHeaders` and a partition key. Two publishers on one topic would
race, and only one of them can be atomic.

The two pact provider-verification tests now build their messages from `KycEvents` instead of from a
hand-rolled copy of the envelope, so the contract is replayed against the real producer rather than
against a second copy that would move with the first.

`openbank-kyc-service` leaves `check-outbox-has-writer.py`'s baseline: **7 → 6 offenders, 0 new,
0 stale**. Not money-path (`rules.yaml: money_path_services`), so the usual single approval applies.

Refs #4007

## How this was proved non-vacuous

`KycOutboxWriteIT` drives the real REST endpoints with RestAssured and reads the rows back over
plain JDBC. It has to: a `Panache.withTransaction` reactive repo called from a bare `@QuarkusTest`
thread throws `No current Vertx context found`, and a unit test that mocks the repository cannot
tell which publisher a use case called — which is exactly why a fully green suite missed this for
the life of the service.

What it must fail on, and did — measured, not reasoned:

| fed to it | result |
| --- | --- |
| the pre-fix behaviour (`repo.save(case)` / `repo.update(updated)` — the event-free overloads, i.e. exactly what `main` does today) | **RED, both tests.** Open path: `POST /api/v1/kyc/cases` still answers 201 and `kyc_outbox` holds **0 rows** (`Expected size: 1 but was: 0`). Approve path: the rows present are `["KYC_CASE_STATUS_CHANGED"]` — only the one write left wired — against the expected three |
| this branch | **GREEN.** 1 `KYC_CASE_OPENED` row after open; `kyc_cases.status = APPROVED` **and** `KYC_CASE_OPENED, KYC_CASE_STATUS_CHANGED, KYC_CASE_APPROVED` after the approve flow |

Full module run: **74 tests, 0 failures, 0 errors, 1 skipped** — the skip is
`KycEventPactBrokerProviderVerificationTest`, gated on `pactbroker.url` by design. `KycOutboxWriteIT`
2/2, `KycOutboxClaimIT` 2/2.

**What the atomicity claim rests on, and what it does not.** The second test asserts both halves
after one HTTP request: the state change (`kyc_cases.status`, read over JDBC) *and* its event row.
The same-transaction guarantee itself is structural — `persistInTransaction` uses the caller's
reactive session and is chained inside the one `Panache.withTransaction` block that performs the
state change, so there is no second transaction to diverge from. Being explicit about the limit
rather than overclaiming: this IT distinguishes *written* from *not written* (proved above), but it
cannot by itself distinguish one transaction from two — only a forced mid-request failure between
the two writes could, and no such fault-injection hook exists in this service today.

## Classification of all seven baselined services

Every one of the seven has `dispatch-enabled: true` already, so the usual "the switch was never
flipped" explanation is **not** the defect for any of them.

| service | verdict | evidence |
| --- | --- | --- |
| **openbank-kyc-service** | **B — wired here** | direct `KycEventPublisher` on `kyc-events-out`; `persistInTransaction` already on the port and unused; both channels already targeted the same topic `openbank.kyc.events`, and a provider pact pins the envelope |
| openbank-pid-service | B — wirable, deliberately left | 15 fire-and-forget publish sites in `PartyService`/`VerificationCaseService`, rich domain events under `domain/event/`, `persistInTransaction` on the port and unused. **Left because two contract questions must be answered first, and neither is this PR's to decide:** its outbox channel drains to `openbank.pid.events`, which **no consumer in the monorepo subscribes to**, while its direct emitter publishes to `party.events` — not the `openbank.party.events` the whole fleet reads. That topic mismatch looks like a live defect independent of the outbox work and is worth its own issue |
| openbank-billing-service | **not a defect — detector false positive** | `src/main` **does** write outbox rows: `BillingAssessmentRepositoryImpl` builds `BillingOutboxEntity` for `billing.fee.post-intent.v1` (line ~90) and `billing.fee.reversal-intent.v1` (line ~210), both persisted inside the fee transaction. The checker's predicate is "constructs an `OutboxMessage`, or a raw `INSERT INTO …outbox`", and billing constructs the **entity** directly, so it matches neither. Its baseline note ("2 DEAD rows exist from a writer no longer present") is therefore wrong. Left alone on purpose: widening the predicate to entity construction would also match the `toEntity()` mapper inside every `*OutboxRepositoryImpl`, which converts a message it was handed rather than originating one — that would mark all six remaining services as fixed and silently retire the gate. Separating writer from mapper is its own change with its own falsification, not a drive-by |
| openbank-audit-service | C — no contract | `audit-events-out` is referenced in code but **not declared** in `application.yaml`; no topic name exists anywhere, no domain event types, no consumer. The service's own `docs/05-operations.en.md` already says the dispatcher "has no live sink". Audit is a sink, not a source |
| openbank-psd2-service | C — no contract, and nothing to be atomic with | topic `openbank.psd2.events` is declared but has no consumer and no event type. The only event-shaped contract is `TppWebhookEvent`, an HTTP webhook whose publisher is a log-only stub with no caller. psd2 is a stateless ČOBS/Berlin-Group façade that owns no aggregate table — there is no local transaction for a transactional outbox to be transactional *with* |
| openbank-tpp-registry-service | C — closest of the three to being wirable | plumbing is complete (port write method, dispatcher, topic `openbank.tpp.registry.event`, real transactions in `registerTpp`/`blacklistTpp`) but there is **no event type, no payload schema and no consumer**. Wiring it means authoring a contract. Worth a separate issue — blacklisting is a fact psd2 and the gateway would want pushed |
| openbank-balance-service | B — wirable, deliberately left | direct `KafkaBalanceEventPublisher` bypasses the outbox, same shape as kyc. Left because balance-service **is money-path** (`rules.yaml: money_path_services`): it needs 2 approvals and a threat-model diff, and it deserves its own PR and its own reviewer attention rather than riding along |

Six half-wired outboxes would have been worth less than one wired properly plus an honest account of
the other six. `Refs`, not `Closes` — #4007 stays open for pid, tpp-registry and balance, and for the
detector's billing blind spot.

## Checks

- `python3 .github/scripts/check-outbox-has-writer.py` → `34 dispatchers; 6 without a writer (6 baselined, 0 new, 0 stale)`; `--self-test` passes
- `application.yaml` re-parsed with a duplicate-key-raising loader (not `yaml.safe_load`, which would prove nothing) — clean
- no Kafka key was added or moved, so the SmallRye dotted-leaf-key trap is not in play; the removed `kyc-events-out` block took no `group.id`/`auto.offset.reset` with it
- `persist()` and not `merge()` in the new `save`: `KycCaseEntity` extends `PanacheEntity`, so the `@Id` is generated and this is a genuine INSERT — the application-assigned-id trap does not apply here. `case_id` is the business key, not the `@Id`


JiRaska:

**Falsified, not just tested green**: edited `KycRepository.save` in a worktree to skip the outbox write (simulating the old dual-write shape). `KycOutboxWriteIT` went red with exactly the expected failures (`kyc_outbox rows … Expected size: 1 but was: 0`). Reverted, confirmed byte-for-byte restored, re-ran clean.

**Persist-vs-merge trap doesn't apply here**, and I checked why rather than assuming: `KycCaseEntity` extends `PanacheEntity`, so its `@Id` is DB-generated, not application-assigned — `save()` genuinely inserts, `update()`/`applyUpdate()` mutate a found managed entity. The PR's own comment states this correctly.

**Envelope/topic compatibility preserved**: `KycEvents.lifecycle()` produces the identical flat map shape as the retired publisher, same topic. `kyc-events-out` channel is deleted, only `kyc-outbox-out` remains targeting the same topic — no dual-publish race. `KycEventPactProviderVerificationTest` (the `@PactFolder`, CI-covered one) is 4/4 — confirms the outbox envelope still verifies against the committed consumer pact.

Gate diff is consistent with `main`'s current file (kyc's baseline comment mirrors party-service's, self-test + real run both clean, 0 stale entries).

**Notes, all non-blocking:**
- `check-outbox-has-writer.py`'s header prose still lists kyc as constructing nothing — stale now that this PR lands, one-line follow-up.
- A duplicate `coVerify` line in the idempotency test (collapsed from two distinct checks during the merge to the outbox API) — redundant, not wrong.
- A test comment claiming `approve`/`reject` take an "event-free overload" is inaccurate — they unconditionally call the event-carrying one in production; the stub for the other is just unused.
- `openbank-onboarding-service`'s consumer still comment-references `KycEventPublisher.publish` by name, a class this PR removes — the stale-prose-naming-a-dead-identifier pattern, separate module, worth a follow-up rename.

PR #4394, opened 2026-08-09
build(case-coordinator): realign the declared runtime base with Dockerfile.deploy (unblocks red main)

JiRaska:
## What

`openbank-case-coordinator-agent/Dockerfile` declared
`eclipse-temurin:25-jre@sha256:681c543d...`; `.github/workflows/Dockerfile.deploy`
declares `@sha256:f19dbf0a...`. This PR makes them agree.

## Why

The enforced gate `dockerfile-runtime-only` (#3016/#3354) fails on that drift:

```
openbank-case-coordinator-agent/Dockerfile: FROM eclipse-temurin:25-jre@sha256:681c543d... does not match .github/workflows/Dockerfile.deploy, which declares FROM eclipse-temurin:25-jre@sha256:f19dbf0a...
```

So `CI` on `main` is red (`gates (registry)` -> `Validate manifests`), and **every
PR opened against `main` inherits a failure it did not cause**. Tracked by the
main-red-watch issue #4191, red since `2946fa3d9` (2026-08-08 13:30Z) and still
refreshing.

The runtime base is stated in exactly one place by design (#3354); the correct
direction of the fix is therefore to take `Dockerfile.deploy`'s digest, not the
other way round.

## Scope — why exactly one file

Enumerated the `FROM` line of every `openbank-*/Dockerfile` on live `origin/main`:

| digest | count |
|---|---|
| `eclipse-temurin:25-jre@sha256:f19dbf0a...` (Dockerfile.deploy) | 51 |
| `eclipse-temurin:25-jre@sha256:681c543d...` | 1 — case-coordinator-agent |
| `alpine:3.24@...`, `nginxinc/nginx-unprivileged:1.31-alpine@...`, `python:3.14-slim@...` | 3 — admin-ui, developer-portal, document-renderer, legitimately different images |

## Verification

```
BASE_REF=main PR_DIFF_BASE=$(git merge-base HEAD origin/main) \
  python3 .github/scripts/run-gates.py --only dockerfile-runtime-only
```

```
--- self-test (must PASS) ---
check-dockerfile-no-build-stage --self-test: OK — rule 3 fires on drift, stays quiet on agreement, covers the non-glob files, exempts SELF_BUILT
--- gate ---
check-dockerfile-no-build-stage: OK — 55 Dockerfiles, runtime-only, all with EXPOSE
1 gates  PASS=1
```

The gate's own self-test asserts rule 3 both ways (fires on drift, quiet on
agreement), so this is not a green from a check that never reached its subject.

[...]

PR #4395, opened 2026-08-09
feat(governance): ADR-0252 — hold the journey catalog against the money-path list (0/23 today)

JiRaska:
ADR-0252 phase 2, follow-on. `check-journey-catalog.py` (merged in #4359) makes every claim **in**
the catalog checkable. It cannot say anything about what the catalog never mentions — and for a
coverage claim, the omissions are the dangerous part. A catalog listing three journeys is not
distinguishable, from the inside, from a platform that watches three things out of three or three
out of two hundred.

## What this adds

A second gate that holds the catalog against an **external list nobody maintains for this purpose**:
`money_path_services` in `rules.yaml`. Each of those 23 services is either covered by a journey
(`covers:`) or listed in a new `money_path_accountability` block with the reason it is not.

Both directions are enforced:

- an **unaccounted** money-path service is red — so a *new* money-path service is red from the day
  it is declared until someone decides what watches it, instead of sitting silently outside a
  coverage claim nobody re-derives;
- an entry for a service that is **no longer** money-path is red too — a stale entry is how a list
  grows a comfortable margin of things that are not true.

That is the repo's own rule applied to itself: never let a gate's scope be maintained separately
from the thing it covers. Here the scope is not maintained at all — it is read from `rules.yaml`,
which other people maintain for their own reasons.

## The number

```
check-journey-coverage: 0/23 money-path services are covered by a journey; 23 accounted for as unwatched.
```

Printed on every run, pass **and** fail. That is deliberate. An unquantified coverage claim is
exactly what this programme exists to end, and `0/23` is more useful to a reader than a green tick.
It goes up as phase 1 lands, and the gate is what stops it quietly going down.

## Why one shared blocker instead of 23

All 23 are blocked on the same thing — no journey can drive a money-path service in production until
synthetic parties exist and the taint reaches the ledger and the regulatory aggregates. So it is
stated once as `default_blocker`. Pasting one sentence 23 times is not more information, and the 23
copies drift.

Five services carry their own note where the situation genuinely differs:

- **domestic-payment** — the reported outage; the `domestic-payment` journey covers it once phase 1 lands.
- **fx** — its ČNB fixing URL was a 404 for the service's whole life while every test layer stayed
  green (#2204), so its journey has to assert the *shape* of the fetched rate, not that a request
  succeeded.
- **sca** — where push carries the challenge, a broken push channel is a broken authentication
  factor, not a UX defect. Depends on phase 3 as much as on phase 1.
- **sanctions** — a control path, so its journey must prove the control **runs** for synthetic
  traffic rather than being skipped for it. The one place the taint must never suppress anything.
- **psd2** — also the PSD2 RTS Art. 32(4) interface whose quarterly availability statistics this
  programme is meant to generate, so its journey is evidence as well as monitoring.

## Verification

- Self-test, 8 cases, control first — including the one that got the logic wrong on the first pass:
  when *every* money-path service is covered by a journey, an empty accountability block is
  legitimate, and demanding one anyway would turn the gate red at the exact moment the platform got
  it right.
- **Falsified against the real tree**, not only in the self-test's temp trees: deleting
  `openbank-clearing-service`'s entry exits 2 naming that service; restoring it returns to OK.
- `run-gates.py --group registry --group lint` otherwise clean.

## Known inherited red

`gates (registry)` fails on `dockerfile-runtime-only`, which has nothing to do with this change:
`openbank-case-coordinator-agent/Dockerfile` carries a base digest that no longer matches
`Dockerfile.deploy`. **The default branch is red on it** (#4191), and three PRs are already open to
fix it — #4392, #4393, #4394 — so this one deliberately does not add a fourth. It goes green once
any of them merges.

Refs #4348, ADR-0252.


JiRaska:
Reviewed. No blocking findings — the gate is honestly scoped and I falsified it against the real tree, not just its own self-test.

**`mode: enforced` but not main-breaking at 0/23**, and that distinction matters: this is an *accountability* gate, not a *coverage-required* one. All 23 money-path services are listed in `journeys.yaml`'s accountability block with a shared `default_blocker` referencing #4348, which satisfies the gate even with zero journeys actually covering them today. Ran it against the merged tree: exit 0.

**Falsified against the live file, not a fixture**: deleted `openbank-vop-service`'s accountability entry from the real `journeys.yaml` — the gate went red naming exactly that service; restored, back to green. It genuinely inspects its subject.

**`gate-subject-floor` convention followed correctly**: prints `SUBJECTS=23` via `gatelib.subjects`, matching `min_subjects: 20`; confirmed via `--only gate-subject-floor` that this gate is on the compliant side, not the debt baseline. Self-test 8/8, including a clean-tree control case before the positive cases.

**No dependency on #4371** — grepped for any reference to `SyntheticTaint`/the taint header in this gate or `journeys.yaml`: only prose mentions in `blocked_by` text explaining *why* journeys are still planned. The two PRs are independent; either can land first.

The gate's own header is upfront that the accountability list is still hand-kept and could go stale without the harder unbuilt half (deriving coverage from the OpenAPI capability surface) — consistent with this repo's gate-scope discipline rather than overclaiming.

PR #4397, opened 2026-08-09
fix(ledger): resolve the ČNB fixing by business day and let a corrected one supersede

### LINKED ISSUE #3921: FX revaluation: rate staleness is unobservable and a late fixing can never be incorporated (#1302 item 4)

JiRaska:
Closes #3921.

Steps 1 and 2 of #3921 landed already — the `CnbFixing(rate, validFrom)` seam and the
`openbank_fx_fixing_age_seconds` gauge, which made the condition *visible*. This is the
correctness half: steps 3 and 4, which make it *fixable*. Re-derived against `origin/main`
@ `8bc9a4565` before building; both claims in the issue still held exactly as filed.

## What was broken

**Step 3 — a backfill marked the wrong day at the wrong rate.** The port was
`cnbRate(base: String)` with no date. `FxRevaluationService` passes the business day it is
marking, and for a belated or manual run that is not today — so an older day was marked at
**today's** fixing, and there was no parameter through which it could have asked for anything else.

**Step 4 — a corrected fixing could never be incorporated.** `idempotencyKey = "fx-reval-${date}"`
keyed on the business day alone, and `LedgerService.postJournal` returns the existing entry on a
key hit. A re-run after a corrected ČNB fixing returned the ORIGINAL entry, emitted no event, and
reported `posted = true` — a success that changed nothing, leaving the position valued at the
superseded rate indefinitely.

## The correction semantics, and why

**A superseding entry that posts the difference. Not a reversal, not an overwrite.**

The deciding fact is that this revaluation is *already carry-relative*.
`FxRevaluationPosting.movement` computes `round(position × rate) − carryCzk`, where `carryCzk` is
the counter-value account's trial-balance net, and the trial balance is cumulative to the business
day (`entry_date <= :asOf`, `PanacheJournalRepository.kt:408`). The first posting carries
`entryDate = command.date`, so **a correcting run reads its own predecessor** and posts only the
difference between the corrected mark and the mark already booked.

That is what makes giving the key a rate-identity component safe at all. Without it the same
change would post the full corrected mark on top of the original and double the position. The test
asserts the arithmetic rather than trusting the argument: 25,145,000 + 355,000 = 25,500,000, the
corrected mark exactly, after two entries.

The two alternatives were rejected on failure mode, not taste:

- **Reversal-and-repost.** `LedgerUseCase.reverseJournal` exists, so it was genuinely available. It
  produces three entries whose net is *arithmetically identical* to the one superseding entry —
  the reversal restores exactly the `carryCzk` the repost would then mark from — and it adds a
  failure mode the superseding form does not have: a reversal that commits while its repost does
  not leaves the position marked at **nothing**, whereas an interrupted superseding run leaves it
  marked at the superseded (merely stale) rate, which is where it already was.
- **Overwriting the original entry.** Not available, and correctly so: journal entries are
  append-only, and an attested year refuses new activity outright (`requireOpenPeriod`). A
  correction into a closed period must fail loudly — and it does, as a superseding posting into a
  locked day, governed by the day/period locks that already exist and are unchanged.

Two properties worth stating because they are what keep the change narrow:

- **The repeat-run case needs no key at all.** An identical re-run computes a movement of zero for
  every leg, contributes no inputs, and returns `posted = false` *before* reaching `postJournal`.
  Widening the key does not reopen the double-posting it was guarding against. The key's remaining
  job is the narrow one it was always doing — two runs racing before either commits.
- **The key is `fx-reval-{date}-{12 hex of SHA-256 over the sorted fixing set}`.** The day stays the
  prefix so an operator grepping a date still finds both entries. The digest is opaque on purpose
  and always resolvable: the entry's own description spells every fixing out in full
  (`[fixings EUR@2026-05-30T15:15:00Z]`), which step 2 added.

**Degradation is unchanged and deliberate.** With no `validFrom` on any leg there is no fixing
identity, and the key falls back to *exactly* `fx-reval-{date}`. Corrections stay impossible, as
they were before — a fabricated identity would be worse than an honest inability, because it would
make a correction *look* possible. `an fx-service answer without a fixing date...` pins that.

## Step 3: the date on the wire

`cnbRate(base, asOf)` → `GET /api/v1/fx/rates/{base}/{quote}?source=CNB&asOf=2026-05-27`, resolved
by a new `FxRateRepository.findBySourceAsOf` (`validFrom <= at < validTo`, newest first, at =
start-of-day in the ČNB publication zone).

Three choices in that query are the ones that matter:

- **A day with no fixing in effect is 404, never a fall-back to the newest fixing.** Falling back
  *is* the defect — it is how a backfill marks an old day at today's rate and reports success.
- **The `validTo` bound is kept.** Dropping it would make a dead feed resolve to the last fixing it
  ever published instead of to `null`, turning ledger's loud "skipping its revaluation leg" into a
  silent mark at an arbitrarily old rate. The three-day window is what makes a stale feed *absent*
  rather than *wrong*.
- **`asOf` without `source=CNB` is 400, not ignored.** A silently-dropped date parameter reads as a
  correct backfill. `asOf` is also parsed with `runCatching` (400, not 500) and declared
  **nullable**, per the `check-nonnull-jaxrs-params.py` rule.

**Equivalence for the live path.** With `asOf` = today, `findBySourceAsOf` returns exactly what
`findLatestBySource` returns for every row this service writes: `CnbRateIngestionService` sets both
validity bounds to Prague midnights, so no row can expire between the start of the day and now. The
daily path is also routed through the *unchanged* `findLatestBySource` (no `asOf` on the wire is
not possible from ledger, but the use case keeps the branch and a test pins that the two queries
do not get swapped by accident).

## §3 of the brief: what happens when there is no fixing — unchanged, and said out loud

`FxRevaluationService` still logs `"No ČNB rate for %s/CZK on %s — skipping its revaluation leg"`
and still returns `posted = false`. **Kept, not changed.** What observes it now is what step 2
already fixed: `FxRevaluationSkippedAllLegs` in `loki-rules-silent-failures.yaml` selects
`{namespace="ledger"}` — the namespace the emitting pod actually deploys to — so the skipped-leg
path has alerting coverage for the first time. The only *new* way to reach that branch is a
backfill of a day no fixing covered, which previously would have silently succeeded at the wrong
rate; it now skips, loudly, into a rule that can match.

## One known interaction I am NOT fixing here, stated rather than left to be discovered

`FxRevaluationService` reports `openbank_fx_fixing_age_seconds` on every attempt (step 2), and that
age is now computed from the fixing of the *day being marked*. So a manual backfill of an old
business day publishes a genuinely large age — correct by the gauge's own definition ("how old was
the fixing this run marked at"), but capable of tripping `FxFixingStale` (> 60h) spuriously, and
the value persists until the next scheduled run overwrites it, up to ~24h.

I left it alone deliberately. Changing what the gauge reports is a change to step 2's contract,
which this issue did not ask for and which deserves its own decision — the honest answer for a
backfill really is "this mark used an old fixing", and suppressing it would make the one signal
that exists lie about a class of run. Flagging it so it is a known consequence rather than a
surprise the first time someone backfills a month. Worth a follow-up: either scope the alert to the
scheduled path, or give the backfill path its own series.

## Verification — both runs, from the JUnit XML

The falsification reverts **only** `val idempotencyKey = idempotencyKey(command.date, fixings)`
back to `origin/main`'s `"fx-reval-${command.date}"`, so the counter-example reaches the defect
rather than a compile error.

| run | tests | failures | errors | **skipped** |
|---|---|---|---|---|
| `FxRevaluationServiceTest` against `origin/main`'s key | 7 | **2** | 0 | **0** |
| `FxRevaluationServiceTest` with the fix | 7 | 0 | 0 | **0** |
| `CnbFixingAsOfPactConsumerTest` | 1 | 0 | 0 | **0** |
| `:openbank-fx-service:test` (whole module) | 137 | 0 | 0 | **1** |
| `:openbank-ledger-service:test` (whole module) | 296 | 0 | 0 | **1** |
| `FxPactFolderProviderVerificationTest` (the replay) | 2 | 0 | 0 | **0** |

Skipped counts read from the JUnit XML, not the console. Each module's single skip is the
`@PactBroker`-sourced provider twin (`FxPactProviderVerificationTest`,
`LedgerPactBrokerProviderVerificationTest`), correctly
`@EnabledIfSystemProperty(pactbroker.url)`-skipped with no broker reachable locally — expected, and
present on `origin/main` too. Nothing new is skipped. The `@PactFolder` classes that *do* run on a
PR verified every interaction, including the new one.

`:openbank-fx-service:quarkusBuild` and `:openbank-ledger-service:quarkusBuild` both pass, so the
CDI wiring for the widened port is validated rather than assumed.

The two failures against `origin/main` are the defect, verbatim:

```
FAILED: a corrected fixing posts a superseding entry for the difference, under a different key
  Expecting actual: "fx-reval-2026-05-30" not to be equal to: "fx-reval-2026-05-30"
FAILED: marks a long EUR position and posts an idempotent fx-reval entry
  Expecting actual: "fx-reval-2026-05-30" to start with: "fx-reval-2026-05-30-"
```

### An existing test asserted the defect

**Yes — and it is the second failure above.** `FxRevaluationServiceTest` line 144 read

```kotlin
assertThat(cmd.captured.idempotencyKey).isEqualTo("fx-reval-2026-05-30")
```

The date-only key — the thing the issue calls the bug — was encoded as the expected behaviour, so
the single-posting defect had a green test defending it. That assertion is now
`startsWith("fx-reval-2026-05-30-")` plus an independent recomputation of the digest, and the new
correction test is what proves the key actually discriminates.

Step 3's defect is a *missing parameter*, so it is only falsifiable at the type level — a
strict-argument mockk stub (`cnbRate("EUR", backfillDate)`) plus
`coVerify(exactly = 0) { cnbRates.cnbRate("EUR", date) }` is the strongest form available, and it
is stated as such rather than dressed up as a runtime falsification.

## Contract

`openapi.yaml` `info.version` **1.6.0 → 1.7.0** — additive (one optional query parameter, plus the
`bidRate`/`askRate`/`validFrom`/`validTo` response fields the new contract depends on, which the
service has always served and the spec did not declare). Read off live `origin/main` immediately
before bumping.

**New consumer pact: `pacts/openbank-ledger-service-openbank-fx-service.json`.** There was none —
ledger's most important outbound call had no contract at all. `asOf` is a request-shape change, and
that is the one class a consumer test structurally cannot catch (the mock server answers whatever
it is asked for), so the value is entirely in the provider replay: a `@State` handler is added to
fx-service's `@PactFolder` class, which runs on a PR. `check-pact-provider-replay.py` reports the
new pact covered; `derive-pact-drift-scope.sh` now includes `:openbank-ledger-service` and exits 0
with no uncovered pacts.

**The provider replay went red on its first run, and that is the single best evidence in this PR.**
`LambdaDsl`'s `datetimeExpression("validFrom", "yyyy-MM-dd'T'HH:mm:ss'Z'")` emits a `timestamp`
matcher whose format is `yyyy-MM-dd'T'HH:mm:ss` — the trailing `'Z'` is consumed as the generator
expression and silently dropped from the matcher. The consumer test was **green** with it, because
the mock server answers the generated example, which matches its own matcher by construction. The
replay against the real service was not:

```
1.1) body: $.validFrom Expected '2026-05-26T22:00:00Z' to match a datetime pattern of
     'yyyy-MM-dd'T'HH:mm:ss': Unable to parse the date: 2026-05-26T22:00:00Z
```

Replaced with an explicit regex, and green after. Two things that failure proves at once, neither
of which any other layer here could have: the matcher was wrong, **and the `?asOf=` route works
end-to-end** — `2026-05-26T22:00:00Z` is exactly Prague midnight of 2026-05-27, i.e. the real
service parsed the date, ran `findBySourceAsOf` against a real Postgres, and returned the row whose
window contains that instant. That is a payload-shape check, not a status code.

The expectation path and query are **string literals**, deliberately not reflected off
`FxServiceClient`'s annotations — deriving both sides is DRY and vacuous, since they then move
together and the test stays green against a route that does not exist (#2290).

One build fix was needed for that: ledger set `pact.rootDir` via a config-time
`System.setProperty`, which only reaches the Gradle daemon, not the forked test JVM — the pact was
written to `build/pacts` where nothing commits it. Now forwarded as a test `systemProperty`, the
shape fx-service already had. The provider verification is unaffected either way (it resolves
`@PactFolder("../pacts")` relative to the module).

## Money-path — what this needs from a reviewer

`openbank-fx-service` and `openbank-ledger-service` are both in
`rules.yaml: money_path_services`, so this needs **2 approvals**, which I cannot grant. Threat
models are updated for both (ADR-0030 D2 — the diff touches an inbound REST surface in fx and an
outbound client edge in ledger, which is what `check-threat-model-diff.py` keys on).

The specific thing worth a reviewer's attention is **not** the date parameter — it is the claim
that a second posting cannot double-count. That rests on one fact:
`FxRevaluationPosting.movement` is carry-relative and the trial balance is cumulative to the
business day. If that is wrong, this PR doubles FX positions on every correction. It is asserted
arithmetically in the new test, and the reversal alternative is analysed above rather than
dismissed.

Refs #1302 (item 4).


JiRaska:
Reviewed to money-path standard. **The PR ships two changes under one title, and only one of them works.** The "resolve by business day" read-side change is correct and well-tested. The "let a corrected one supersede" half cannot fire in production — I traced it to the actual code, not the tests, and confirmed the gap directly.

## Blocking: the write side this depends on was never touched

`ingest()` in `CnbRateIngestionService.kt` is **unchanged by this PR** (verified: the diff touches `getCnbRate`, not `ingest`):

```kotlin
val validFrom = fixing.date.atStartOfDay(zone).toInstant()
...
if (rateRepo.findBySourceAndValidFrom(rate.code, QUOTE, RateSource.CNB, validFrom) != null) {
    skipped++
    continue
}
```

`validFrom` is a pure function of the **business day** — identical for an original fixing and any same-day correction ČNB republishes. There is no comparison of rate value, no sequence tie-break, no update path. `save()` only ever inserts. This is not incidental: the pre-existing test `is idempotent — already-stored fixing for the day is skipped, not re-saved` asserts exactly this, and the manual re-ingest endpoint's own KDoc says it is idempotent **by design**. So a ČNB correction can never be stored, by any path — scheduled sweep or operator trigger.

The ledger-side "supersede" mechanism keys on `fixingDigest`, which is `"${currency}@${validFrom}"`. Since fx-service can never produce two different `validFrom` values for one business day, the digest can never change for a real correction — even in the case where fx-service did store one.

**The test that appears to prove this doesn't**, and the reason is worth naming precisely: `FxRevaluationServiceTest.kt` mocks `CnbRateProvider` directly with a hand-fabricated `validFrom = Instant.parse("2026-05-30T15:15:00Z")` — a specific time-of-day. I checked the real adapter: it never produces a time-of-day `validFrom` for a CNB fixing, only `LocalDate.atStartOfDay(zone)`. The test validates the ledger's key-derivation logic in isolation against an input the real pipeline cannot produce.

**Concrete scenario:** ČNB publishes EUR/CZK 25.145 at 14:30 on 2026-05-30 (stored, `validFrom = 2026-05-30T00:00 CEST`). At 16:45 ČNB republishes 25.500 for the same fixing. An operator re-triggers ingestion: `findBySourceAndValidFrom` finds the existing row and skips — the correction is discarded, `skipped = 1`. Ledger's revaluation resolves the same stale rate, computes the same digest, hits the same idempotency key, and reports success having changed nothing. The position stays marked at 25.145 forever — the exact defect this PR's own comment says it closes, one layer further back.

Either the ingestion side needs to accept and store a correction (drop the skip when the rate differs, or version rows and have the read side pick the newest), or the title and both threat-model updates need to be scoped down to what's actually delivered, with the correction mechanism tracked as a follow-up.

## Correct and well-tested: business-day resolution (the read side)

`getCnbRate(asOf)` uses `validFrom <= at AND validTo > at` with `CNB_VALIDITY_DAYS = 3` unchanged. Checked the weekend boundary by hand: a Friday fixing's window is `[Fri 00:00, Mon 00:00)`, so Saturday and Sunday both resolve to it, and Monday's own `asOf` (`Mon 00:00`) correctly fails the strict `validTo > at` and falls through to Monday's fresh row. No off-by-one, no overlap.

**Note, not blocking:** the 3-day constant is not holiday-aware — it is a flat number unrelated to the Czech public-holiday calendar, unchanged by this PR. This repo has a real `BusinessCalendar`/`Easter` implementation in `openbank-libs-domain`, but neither fx-service nor ledger-service references it anywhere. The constant's own comment ("covers weekends **and public holidays**") overstates what it does — a holiday not immediately adjacent to a weekend the 3-day window already covers is not compensated, and `asOf` resolves to null rather than falling forward. The failure mode is safe (logged as a missed revaluation leg, `posted=false`, freshness gauge climbs — not a wrong figure), so worth a follow-up issue rather than a blocker here.

## Checked clean

No persist-vs-merge trap — the only new repository method (`findBySourceAsOf`) is read-only, and grepping the diff for `.persist(`/`.merge(` returns nothing; FX rate rows are always fresh inserts. No double-posting risk in the logic itself: a third run against an already-fully-corrected position computes zero movement per leg and returns before `postJournal` is ever called, so the key never needs to discriminate a same-fixing repeat — sound reasoning, though it currently defends a path that can't be reached. Every new/changed test uses `runBlocking<Unit> { }` explicitly, so nothing is silently skipped by JUnit5.

## Worth fixing

`findBySourceAsOf`'s HQL (`validFrom <= :at and validTo > :at order by validFrom desc, createdAt desc`) has no coverage against a real database — every test mocks the repository port. A dedicated repository-level IT would catch a timestamp-precision or ordering regression that a unit test structurally cannot.

## What I could not verify

I was unable to get a Gradle build to complete in this shared environment — 15+ concurrent Gradle daemons observed, the build stuck at project-configuration for 8+ minutes. So the revert-and-falsify exercise the test's own comment claims ("against origin/main this fails on the very first assertion of run 2") is corroborated by tracing `postJournal`'s key-hit short-circuit statically, not confirmed by execution. Flagging rather than reporting a result I didn't obtain.

Issue #4403, opened 2026-08-09
Rebuild domestic-payment fraud enforcement in the Temporal workflow (the ADR-0084 §4.2 control was deleted by #1917)


JiRaska:
## What

`openbank-domestic-payment` has no fraud **enforcement** path. ADR-0084 §4.2 and
`docs/threat-models/openbank-domestic-payment.md` described one, flag-gated by
`openbank.domestic.fraud.enforcement-enabled`. It does not exist: `DomesticPaymentService.applyFraudGate`
and the `fraudEnforcementEnabled` service flag were deleted by the Temporal migration (ADR-0120 Phase 6,
#1917), which carried over only the shadow `shadowFraudScore` activity. The config key survived with zero
readers in `src/main/kotlin`, so the runbook flip changed nothing and reported nothing.

#4221 removed the orphaned key and corrected both documents, so the record is now accurate. It did not
rebuild the control — that is this issue.

## Why it is not a config change

The verdict now arrives inside a Temporal activity (`DomesticPaymentActivities.shadowFraudScore`, returns
`Unit`) rather than in an in-process service method. Acting on it needs:

- a decision returned from the activity to `DomesticPaymentWorkflow` (an activity signature change, which
  has replay-compatibility consequences for in-flight workflows),
- a branch in the workflow for REVIEW/CHALLENGE (hold) and DECLINE (reject via
  `DomesticRejectReason.FRAUD_SUSPECTED`, whose enum value is still present),
- a decision about what a **synthetic** verdict means once the verdict is enforced. Today the adapter is
  fail-open and that is right, because nothing acts on the answer. Under enforcement the same fallback
  would mean "an unreachable fraud-service allows everything", which is the opposite of what a fraud
  control should do when it loses its data source. `FraudScoreOutcome.synthetic` (#4221) is the field that
  makes that decision expressible; the enforcing branch must not read a synthetic ALLOW as a clean pass.

## Prerequisites

ADR-0084 discloses the v3/v4 thresholds as "first-pass, non-calibrated figures... pending shadow-mode data
and risk-team input". Shadow data only started existing once #4221 layer 1 put a real `FRAUD_SERVICE_URL`
on the four consumer pods, so calibration input is now accumulating for the first time. Enforcement should
follow that calibration, not precede it.

Money-path (`rules.yaml: money_path_services`): 2 approvals + a threat-model update (ADR-0030).

[...]

Issue #1917, opened 2026-07-23
Finish ADR-0120 Temporal payment migration — retire the dual saga


JiRaska:
ADR-0120 (migrate payment orchestration to Temporal) is mid-flight: the custom in-libs saga (ADR-0045, superseded) and Temporal (ADR-0101) coexist, so compensation semantics live in two places — exactly where money-path bugs hide (audit §2.3 #4).

Scope:
- Inventory which flows still run on the legacy saga (grep for the libs saga entry points)
- Migrate remaining flows to Temporal workflows with equivalence tests (same compensation on the same failure injection)
- Delete the legacy saga from openbank-libs; mark ADR-0045 delivery axis accordingly
- Keep the deterministic money-path simulation green throughout (ADR-0100/0115)

JiRaska:
Reopening: settlement (#2066) and sepa-payment (#2110, +prerequisite #2068) are now Temporal-only with the legacy saga retired. **domestic-payment is the last service still on the dual saga** (`if(openbank.temporal.enabled) Temporal else legacy`, flag=false default). Its Temporal path should be checked for the same defined-but-uncalled-activity gap sepa had (shadowFraudScore) before deleting the legacy flow. Closing #1917 was premature — this tracks the domestic cutover.

JiRaska:
ADR-0120 Phase 6 complete — the dual saga is retired across all three money-path orchestrators, each now Temporal-only:

- settlement-service — #2066 (1/3)
- sepa-payment — #2110 (2/3)
- domestic-payment — #2122 (3/3)

Each service's `createPayment`/orchestration no longer branches on `openbank.temporal.enabled` (removed); the workflow (screening → shadow fraud → scheme → settle) runs durably inside Temporal activities with retries + reverse compensation. Legacy in-service flows and their now-unused ports deleted; worker registration gated separately (`openbank.<svc>.worker.enabled`, %test false) with an in-process `WorkflowClientTestProducer`/`TestWorkflowEnvironment`. Sandbox canary confirmed workers register + poll and payments complete end-to-end via Temporal. Threat-models updated (ADR-0030).

PR #4411, opened 2026-08-09
fix(domestic-payment): make a synthetic fraud verdict distinguishable from a real one (#4221 layers 2+3)

### LINKED ISSUE #4221: Fraud scoring has never run: no workload sets FRAUD_SERVICE_URL, the adapter fails open, and the enforcement flag has no reader
## Summary

Fraud scoring has never scored anything. `fraud_scores` in `openbank_fraud` holds **0 rows** — not few, zero, across every rail. Four services hold a fraud rest-client, all four resolve it to `localhost`, and the mode that would act on a verdict cannot be switched on.

Three independent layers, each silent on its own, each making the next one unreachable.

## 1. No workload sets `FRAUD_SERVICE_URL`

All four consumers declare the same default:

```yaml
fraud-service:
  url: ${FRAUD_SERVICE_URL:http://localhost:8133}
```

`openbank-domestic-payment`, `openbank-sepa-instant`, `openbank-sepa-payment`, `openbank-fx-service` (each in its own `application.yaml`).

The only occurrence of `FRAUD_SERVICE_URL` anywhere under `openbank-infra/gitops` is inside a comment:

```
payments-services.yaml:1132:  # NOTE: no FRAUD_SERVICE_URL — sepa-instant does not consume fraud scoring yet
```

Probe validated against a known-positive — `SANCTIONS_SERVICE_URL` appears in 5 gitops files, so the zero is a fact and not a broken grep.

The target exists and the port is right: `fraud-service.fraud.svc:8133` (ClusterIP, 54d). Only the host is wrong. Live symptom, once per payment:

```
WARN com.ope.dom.inf.cli.FraudScoringAdapter — Fraud scoring unavailable (rail=DOMESTIC); shadow ALLOW
  Connection refused: localhost/127.0.0.1:8133
```

## 2. The adapter fails **open**

`FraudScoringAdapter.score` catches every exception and returns a synthetic pass:

```kotlin
} catch (ex: Exception) {
    log.warnf(ex, "Fraud scoring unavailable (rail=%s); shadow ALLOW", command.rail)
    FraudScoreOutcome(FraudVerdict.ALLOW, 0, "unavailable", listOf("fraud-service-unavailable"))
}
```

Harmless while the verdict is only observed. It stops being harmless the moment anything acts on it — an unreachable scorer would then mean "allow everything", which is the opposite of what a fraud control should do when it loses its data source.

## 3. The enforcement flag has no reader

`openbank-domestic-payment/src/main/resources/application.yaml` declares it, and describes it as a deliberate rollout lever:

```yaml
fraud:
  # ADR-0084 §4.2: flips the fraud verdict from SHADOW (observed, logged, ignored) to
  # ENFORCED (REVIEW/CHALLENGE hold the payment for manual release; DECLINE rejects it).
  enforcement-enabled: ${DOMESTIC_FRAUD_ENFORCEMENT_ENABLED:false}
```

ADR-0084 §4.2 and `docs/threat-models/openbank-domestic-payment.md:198` both document the mode. **Nothing in `src/main/kotlin` reads it** — 0 occurrences of `openbank.domestic.fraud` in the service's Kotlin. `shadowFraudScore` is hardcoded to log and move on. Setting `DOMESTIC_FRAUD_ENFORCEMENT_ENABLED=true` per the runbook would change nothing and report nothing.

## Why fixing only layer 1 is worse than today

Adding the env var makes scoring live and `fraud_scores` start filling — which reads as "fraud detection now works". It would not: the adapter still fails open, and the enforcement mode still cannot be turned on. The result is a control that looks armed and is not. The three want fixing together, or in the order 3 → 2 → 1.

## Suggested direction

- Set `FRAUD_SERVICE_URL: http://fraud-service.fraud.svc:8133` on all four workloads.
- Decide the failure posture explicitly. Fail-open may well be right for a payment rail, but it should be a named decision with a metric and an alert on the unavailable path, not a `catch (Exception)` default.
- Either wire `enforcement-enabled` to real behaviour, or delete the key and correct ADR-0084 §4.2 plus the threat model. A documented control that cannot be switched on is worse than an absent one — the threat model currently credits a mitigation that does not exist.
- Alert on `fraud_scores` not growing. Nothing today distinguishes "no fraud" from "no scoring"; that silence is what let this run since the service was deployed.

## Detection gap

`check-incluster-hostnames.py` (#3974) cannot catch this class: it validates that hostnames *present* in config resolve, and explicitly leaves `localhost` alone. The defect here is an env var that is **absent**, so the fallback is a syntactically valid localhost URL. Same shape as the dead `OPENBANK_TRANSACTION_ORCHESTRATION_TEMPORAL_*` prefix in #4178 — config that nothing reads, or reads from nowhere, fails by looking correct.

Found while investigating the transaction-service outage in #4178. Related: #4182.

JiRaska:
Layers 2 and 3 of #4221. Layer 1 (`FRAUD_SERVICE_URL` on the four consumer workloads) is already
live and was re-verified against the cluster before building this — all four pods carry
`http://fraud-service.fraud.svc:8133`.

## The defect, restated precisely

Fraud scoring fails **open**: an unreachable fraud-service produces `FraudScoreOutcome(ALLOW, 0,
"unavailable", ["fraud-service-unavailable"])`. The fallback is not the bug. The bug is that the
fallback was **invisible** — the caller in every one of the four services does exactly this:

```kotlin
if (outcome.verdict != FraudVerdict.ALLOW) { log.infof("... observed, not enforced", ...) }
```

so a synthetic ALLOW takes the silent branch, which is the same branch a genuinely clean payment
takes. A scorer that had never once run was byte-for-byte indistinguishable from a stream of clean
payments, at every layer that reads the outcome. That is why `fraud_scores` could hold zero rows for
the entire life of the deployment with nothing anywhere saying so.

## Enforced or observed? — Observed. Established from the callers, not from the docs

Every call site was read. There are five, across four services:

| service | call site | what it does with the verdict |
| --- | --- | --- |
| domestic-payment | `DomesticPaymentActivitiesImpl.shadowFraudScore` | logs a non-ALLOW verdict, returns `Unit` |
| sepa-payment | `SepaPaymentActivitiesImpl.shadowFraudScore` | logs a non-ALLOW verdict, returns `Unit` |
| sepa-instant | `SctInstPaymentService.scoreFraudShadow` | logs, then `.replaceWithVoid()` |
| fx-service | `FxActivitiesImpl.shadowFraudScore` | logs a non-ALLOW verdict |
| fx-service | `FxService.scoreFraudShadow` | logs a non-ALLOW verdict |

No branch anywhere changes a payment's state, status or outcome based on the verdict. In
domestic-payment the activity returns `Unit`, so the workflow could not act on it even if it wanted
to. **Observed, not enforced** — and the one thing that claimed otherwise (the enforcement flag) is
layer 3 below, which turns out to be the reason the claim was wrong.

## Fail-open vs fail-closed: fail-OPEN, kept, and now written down as a decision

Fail-closed would stop payments on four rails to protect a value that nothing reads. That is strictly
worse than the status quo in both directions — it adds a customer-facing outage and buys no fraud
protection, because no protection is being applied. So the posture is unchanged, and is now justified
in the adapter KDoc rather than being a bare `catch (Exception)` default.

**This PR deliberately does not change enforcement semantics.** The moment enforcement exists, the
correct posture likely flips for the synthetic case specifically — "the scorer is unreachable, so
allow everything" is the opposite of what a fraud control should do when it loses its data source.
`FraudScoreOutcome.synthetic` is the field that makes that decision expressible when someone builds
enforcement (#4403); it is not exercised as a gate here.

## Layer 2 — a synthetic answer is now distinguishable from a real one

Three places, per the issue:

1. **At the outcome.** `FraudScoreOutcome.synthetic: Boolean = false`, set `true` only by the
   adapter's fallback. `ruleVersion == "unavailable"` already carried the same information and was a
   magic string no caller ever compared; this is the same claim in a form the type system can see.
2. **In the metrics.** New per-service `FraudScoringMetrics`, tagged `service=` and `rail=`:
   - `openbank_fraud_scoring_degraded` — gauge. `1` = the last attempt fell back, `0` = the last
     attempt was real, `-1` (`NEVER_ATTEMPTED`) = this pod has scored nothing since boot.
   - `openbank_fraud_scoring_outcomes_total{result="real"|"synthetic"}` — counters.
3. **In what the caller records.** All five call sites now log at WARN when `outcome.synthetic`,
   naming the payment/conversion and pointing at the gauge. Previously this case logged nothing.

### The counter/gauge trap, addressed head-on

An alert on a counter that has never been incremented matches nothing forever, and a pod restart
resets it — so "no synthetic scores" and "no scoring at all" would look identical, which is the exact
conflation this issue is about. Two things follow:

- both counters are registered **eagerly** at `@PostConstruct`, so they exist at `0.0` from the first
  scrape rather than materialising on the first failure;
- the standing "is scoring degraded right now" question is a **gauge read from adapter state**, and
  its boot value is `-1`, not `0`. A boot-time `0` would assert "scoring is healthy" on a pod that has
  never called fraud-service.

`@Startup` is on the bean for the documented reason that `@ApplicationScoped` is lazy — without it
the gauge would not exist until the first payment, and a pre-traffic scrape would show no series at
all instead of `-1`.

### Proof the alert expressions match a real series

Three independent proofs, because none alone is enough:

**(a) The series shape is asserted against a real Prometheus scrape, not against a meter id.**
`FraudScoringMetricsTest` binds a real `PrometheusMeterRegistry` and asserts literal scrape lines, so
the dots→underscores mapping, the counter's `_total` suffix and the alphabetical tag ordering come
from Micrometer rather than from this test's idea of them:

```
openbank_fraud_scoring_degraded{rail="DOMESTIC",service="domestic"} -1.0
openbank_fraud_scoring_outcomes_total{rail="DOMESTIC",result="synthetic",service="domestic"} 0.0
```

The first assertion runs against a **fresh** registry — before anything is recorded — which is the
falsification of the "alert matches nothing until the first failure" trap.

**(b) The label convention is confirmed against the live cluster.** `service=` is what the existing
gauges really render, checked by querying Prometheus in the sandbox rather than by reading the code:

```
openbank_domestic_payments_non_terminal{service="domestic", status="VALIDATED", ...} = 3
```

For the negative control, `openbank_fraud_scoring_*` matches **nothing** today — the `__name__` list
contains no `openbank_fraud*` series at all, only Kafka topic metrics. That zero is the current state
this PR changes.

**(c) All three alert expressions were executed against the live Prometheus** (`/api/v1/query`) and
parse cleanly, returning an empty vector as expected, with a control query (`up{...}`) returning data
in the same session so an empty result cannot be confused with a dead endpoint.

Alerts land in `openbank-infra/gitops/components/payments/prometheus-rules.yaml` as one
`openbank-fraud-scoring-alerts` document covering all four rails via their labels, rather than four
hand-kept per-rail copies. fx-service is in the `fx` namespace; a PrometheusRule's namespace does not
scope which series its expressions evaluate, so the rail belongs with its siblings.

## Layer 3 — the enforcement flag had no reader because the control was deleted

`openbank.domestic.fraud.enforcement-enabled` was not an unfinished feature. The code it named —
`DomesticPaymentService.applyFraudGate` and the `fraudEnforcementEnabled` service flag — was
**deleted by the Temporal migration** (ADR-0120 Phase 6, #1917), which carried over only the shadow
`shadowFraudScore` activity. The config key, ADR-0084 §4.2 and the threat-model row all survived it.
So the documented runbook flip would have changed nothing and reported nothing, and the threat model
has credited an absent mitigation since 2026-07-24.

Taken here, and it is the issue's own second option ("either wire it, or delete the key and correct
ADR-0084 §4.2 plus the threat model"):

- the key is **removed**, with an in-place comment saying why and what re-adding it would require;
- **ADR-0084 §4.2** gets a dated correction: the block is history, and all four scored surfaces are
  shadow-only today. The original entry is corrected in place rather than deleted, so the next reader
  learns the note was wrong rather than that it was never there;
- the **threat model** gets a dated retraction of the mitigation it credited;
- the gitops comment repeating the claim (`payments-services.yaml`) is corrected too;
- `FraudEnforcementFlagRetiredTest` asserts the general property — every `openbank.domestic.*` key is
  named by some Kotlin source — so the next key to lose its reader fails rather than sits there.

Wiring enforcement instead was rejected for this PR: the verdict now arrives in a Temporal activity
returning `Unit`, so acting on it needs a decision returned to `DomesticPaymentWorkflow` plus a hold
state — an activity-signature change with replay consequences, on a money-path rail, and squarely the
"do not change enforcement semantics as a side effect of adding a metric" line. Tracked in **#4403**,
which also records the calibration prerequisite ADR-0084 itself states.

## Did an existing test assert the defect?

**No — and this is worth reading carefully, because two tests come close.**
`FraudScoringAdapterTest` in fx-service and sepa-instant both assert the fail-open ALLOW, including
`ruleVersion == "unavailable"` and the reason list. Those assertions encode the *deliberate* posture,
not the defect, so **nothing was deleted**: both files are extended in place and every pre-existing
assertion still stands. What they did not assert — and what let the defect survive — is that the
fallback is *marked*. domestic-payment and sepa-payment had no adapter test at all.

## Both test runs

The falsifying assertion that compiles against `origin/main` is the `Error`-containment one: an
`Error` is not an `Exception`, so main's `catch (ex: Exception)` lets it escape `score()` and
propagate into the payment path, out of an adapter whose entire contract is that it cannot affect it.

**Against `origin/main` (49c380bd5), new test only — FAILS:**

```
$ ./gradlew :openbank-fx-service:test --tests '*FalsifyFraudFailOpenTest'

FalsifyFraudFailOpenTest > an Error from the client is contained, not propagated
    out of the fail-open path() FAILED
    java.lang.NoClassDefFoundError: com/openbank/fraud/Boom
        <- the Error escaped `catch (ex: Exception)` in FraudScoringAdapter.score

1 test completed, 1 failed
JUnit XML: tests="1" skipped="0" failures="1" errors="0"
```

**Against this branch — PASSES**, along with the full suites of all four modules:

```
FraudScoringAdapterTest (fx-service)      tests=7  failures=0  errors=0  skipped=0
  - an Error from the client is contained, not propagated out of the fail-open path()  PASSED
  - a real ALLOW and a synthetic ALLOW are not equal()                                 PASSED
  (the five pre-existing verdict-mapping / fail-open tests still pass, unmodified)

Full module suites, counts read from build/test-results/test/TEST-*.xml:

  openbank-domestic-payment   tests=140  failures+errors=0  skipped=0
  openbank-sepa-payment       tests=103  failures+errors=0  skipped=0
  openbank-sepa-instant       tests=69   failures+errors=0  skipped=0
  openbank-fx-service         tests=131  failures+errors=0  skipped=1
                              (FxPactProviderVerificationTest, broker-gated, pre-existing)

ktlintCheck + detekt: BUILD SUCCESSFUL on all four modules.
```

Skipped counts are read from the JUnit XML, not the console. An earlier domestic-payment run showed
`1 failure / 11 skipped`; that was `QuarkusBindException: Port already bound: 8081` from a parallel
Gradle build on the same machine, and the uncontended re-run is the `140 / 0 / 0` above.

The rest of the new assertions (`outcome.synthetic`, the counters, the gauge) cannot compile against
`origin/main` at all, because neither the field nor `FraudScoringMetrics` exists there. That is a
genuine falsification — the property is absent, not merely unasserted — but it is a compile error
rather than a red assertion, which is why the `Error` case above is called out separately.

## Money-path / review

`openbank-domestic-payment`, `openbank-sepa-payment`, `openbank-sepa-instant` and
`openbank-fx-service` are **all four money-path** (`rules.yaml: money_path_services`). Per ADR-0030
this needs **2 approvals** and a threat-model update; the domestic-payment threat model is updated in
this PR (risk class = detectability; no behaviour, boundary, data-flow or DB-schema change). I cannot
grant approvals.

## Other notes for the reviewer

- `openbank-domestic-payment/build.gradle.kts` gains one **test-only** dependency
  (`micrometer-registry-prometheus`, same version and rationale as `openbank-libs-runtime`'s
  `WorkflowLivenessMetricNamingTest`). The registry already ships at runtime via
  `quarkus-micrometer-registry-prometheus`; only the compile classpath needed it. Note this arms the
  ~11 min fleet dependency-resolution path.
- sepa-instant's adapter now assembles its `Uni` via `Uni.createFrom().deferred { ... }`. Building the
  Uni calls `client.score(...)` synchronously, so a throw there escaped past every `onFailure()` —
  the Mutiny equivalent of the `Error` hole the three coroutine adapters had.
- `ktlintFormat` reformats a whole module, so it touched ~30 unrelated files; those were reverted and
  only the files this PR owns carry formatting changes.
- No `openapi.yaml`, DB or event-schema change. No `version.txt` / `CHANGELOG.md` hand-edits.

Closes #4221
Refs #4403


JiRaska:
Reviewed (four adapters: domestic-payment, fx-service, sepa-payment, sepa-instant). No blocking findings.

**Verified by running, not just reading:**
- `synthetic` field is set `true` only in the hardcoded fallback constant, never derived from caller input — falsified by flipping it to `false` in the fallback and confirming `FraudScoringAdapterTest` goes red (2 failures) in domestic-payment, then restored to green.
- Fail-open ALLOW behavior is preserved unchanged — this PR only adds observability (log line, `openbank_fraud_scoring_degraded` gauge, `openbank_fraud_scoring_outcomes_total` counter, three new alerts), it does not change the decision.
- Catch-clause ordering is correct: `CancellationException` rethrown before the broader `catch (Throwable)`, so coroutine cancellation is never misrecorded as an outage. sepa-instant's Mutiny path wraps the sync `client.score(...)` call in `Uni.createFrom().deferred{}` so a synchronous `Error` is caught there too.
- `catch (Exception)` → `catch (Throwable)` widening is real and tested — the `NoClassDefFoundError` containment test fails to compile-and-run correctly against `origin/main`'s old `catch (Exception)`.
- No dependency on #4413 (`SyntheticTaint`/canary marker) — that's an unrelated mechanism; `grep -c "SyntheticTaint"` over the diff is 0. This PR's "synthetic" means "fabricated fallback verdict," not "canary traffic." Flagging in case the two got conflated in planning.
- `gradle/verification-metadata.xml` diff is a clean, scoped append (additions-only convention respected).
- All three `catch`-containment tests correctly declare `: Unit = runBlocking { }` — no silent-skip risk.

**Worth-fixing, non-blocking:** `openbank-domestic-payment/build.gradle.kts:52-56`'s comment says the new verification-metadata entries are for "the same version" as the pinned `micrometer-registry-prometheus:1.14.5` — they're actually for `1.17.0`, which is what the Quarkus BOM force-resolves it to via `enforcedPlatform`. Not a bug (build runs green), just a misleading comment.

JiRaska:
The `build (openbank-transaction-service)` red here was **not caused by this PR**. Re-ran it: attempt 2 is green.

The failure was one test of 127:

```
TransactionOutboxClaimIT > a stale DISPATCHING row is reclaimed instead of stranded forever() FAILED
```

That test has been on `main` since `156d39a5` (2026-07-17, the atomic `FOR UPDATE SKIP LOCKED` claim work, #1495) and this PR does not touch it.

Worth recording rather than just re-running, because of *how* it fails: **the test's own logic is deterministic.** It forces `claimedAt = Instant.EPOCH` and then reclaims with a staleness threshold of one second — 1970 is stale under any clock, so there is no timing race in the assertion. A flake here therefore points at infrastructure (the `PostgresRedpandaTestResource` container, or cross-test contamination of the shared table via `clearOutbox()`'s `deleteAll`), not at a fragile assumption in the test.

One occurrence, so I have not filed an issue. But this is a money-path outbox claim path, and the class's helper bridges with

```kotlin
VertxContextSupport.subscribeAndAwait { ... }
```

which the repo's own guidance calls out as the wrong bridge — it throws if ever reached from an event loop, so it breaks the day the code under it gains `@NonBlocking` or a virtual thread. If this test flakes again, that helper is where I would look first, and at that point it deserves an issue rather than another re-run.

JiRaska:
## Review — verified against the code and the live cluster, not just read

Checked the load-bearing claims rather than trusting the (unusually good) description:

| claim | verdict |
|---|---|
| `synthetic = true` set only by the adapter fallback | **holds** — one site per service, all four in `FraudScoringAdapter`'s fallback |
| five call sites, all now logging on synthetic | **holds** — domestic 1, sepa-payment 1, sepa-instant 1, fx 2 |
| gauge boot value `-1`, distinct from a healthy `0` | **holds**, and it is the right call — a boot-time `0` would assert "scoring is healthy" on a pod that has never called fraud-service |
| counters registered eagerly so an alert can match before the first failure | **holds** |
| no semantic clash with #4275's scheme-dispatch guard in the same file | **holds** — the only diff in `DomesticPaymentActivitiesImpl.kt` vs `main` is the fraud branch |

## One finding: `FraudScoringNeverAttempted` would fire ~continuously here

The rule is `openbank_fraud_scoring_degraded == -1` for `1h`, and its comment names the false
positive honestly:

> A rail with genuinely no traffic for an hour is the false positive to expect here — which is itself
> worth knowing on a payment rail.

That is a claim about traffic, so I measured it:

```sql
SELECT created_at::date, count(*) FROM domestic_payments
 WHERE created_at > now() - interval '10 days' GROUP BY 1;

 2026-08-13 | 1
 2026-08-08 | 1
```

**Two domestic payments in ten days.** The gauge reads `-1` until the pod's *first* scoring attempt,
and pods restart on every deploy — which today happens many times a day. So in this environment the
"expected false positive" is not an edge case, it is the resting state: a fresh pod sits at `-1` for
days, and the alert fires on essentially every deploy.

That is the same shape as the `WorkflowLivenessStale` incident this repo already paid for (#2239 /
fixed by #4208): a sentinel value that is correct as a *report* becomes noise the moment an alert is
pointed at it, and noise on the control that exists to make an invisible failure visible is the one
thing that hides an invisible failure.

**Caveat, stated plainly:** this is measured on sandbox. A production rail with continuous traffic
would make `for: 1h` reasonable, and the rule may be tuned for that. But this rule ships to the
sandbox manifests, so as written it will be noisy where it is deployed.

Two ways out, both cheap:
1. Raise `for:` on this one alert to something above the observed inter-payment gap (24h would still
   catch the original defect — that pod went unscored for its entire life).
2. Or gate it on the rail having had traffic at all, so "idle rail" and "scoring broken" stop sharing
   an alert — which is the same conflation the PR exists to remove, one level up.

I have not changed it: choosing between detection lag and false positives is a tuning decision, and
it depends on which environment this is meant to protect.

## Unrelated, but worth knowing: layer 1 is now confirmed working

```sql
SELECT count(*) FROM fraud_scores;  -- 1
```

This morning it was **0**, and I could not prove #4272's wiring fix worked because no payment had
crossed any rail since it deployed. The 2026-08-13 payment scored. Layer 1 of #4221 is now verified
with real data rather than by transport probe.

PR #4412, opened 2026-08-09
fix(lending): stamp business event time on the money-path audit producers

JiRaska:
## What this does

Gives the **three remaining** #3914 producers a real business event time. #3926 fixed the four
non-money-path services (dispute, statement, sanctions, document); the body of #3914 explicitly
reserved these three because all three are in `rules.yaml: money_path_services` and each therefore
needs **2 human approvals**. 13 payload sites, all 13 fixed.

Without `occurredAt`, `AuditConsumer.eventTime()` (`AuditConsumer.kt:230-236`) returns null and
`audit_entries.occurred_at` records the CONSUMER's ingest time as the business time — the GDPR
Art. 30 "when" dimension and DORA Art. 17 evidence. Under consumer lag or a replay, arbitrarily
wrong.

Refs #3914.

## Per-producer enumeration, WITH the method used for each

The repo builds event payloads two ways and a string grep sees only one of them: a hand-built map
spells the JSON key literally, while a serialised data class has no literal anywhere — the key
exists only at runtime as a Kotlin property name. So the method is stated per producer.

| producer | file | sites | payload construction | method used to enumerate |
|---|---|---|---|---|
| clearing-service | `ClearingEventPublisherImpl` | 2 | `objectMapper.writeValueAsString(mapOf(...))` — hand-built map, keys ARE literals | read the two `mapOf` blocks in full; `command grep -c occurredAt` over `src/main` = **0** (corroboration only, not the enumeration) |
| sepa-payment | `SepaPaymentActivitiesImpl` | 5 | raw Kotlin string templates — literals | read every `SepaPaymentOutboxMessage(` construction site in the file (`:116 :132 :191 :210 :227` on main) |
| lending-service | `LendingService` | 6 of 10 | raw Kotlin string templates — literals | read all 10 `LendingOutboxMessage(` sites; 4 already carried `occurredAt` (`LendingService:226`, `OriginationDecisionService:175`, `TerminationService:353` and `:364`) |

All three build payloads **literally**, so a grep and a read agree here — but the grep is not what
established it. Two near-misses that a `payload =` grep does hit and that are NOT payload sites,
re-confirmed on this branch: `CompliancePackActivationService.kt:59` (an entity column) and
`LendingResource.kt:204` (a debug projection of `entry.createdAt`).

`grep` on this machine is ugrep; every command above used `command grep`.

## Does a consumer actually STORE the missing time? Yes — all three.

A field nothing reads is a latent trap, not corruption. Checked the consumer side rather than
assuming:

- `AuditConsumer`'s `audit-events-in` `topics:` list (audit-service `application.yaml`) contains
  **all three** topics: `openbank.clearing.batch.event`, `openbank.sepa.payment.events`,
  `openbank.lending.events`.
- All three publishers emit the outbox row **verbatim** — the identical
  `emitter.sendMessage(Message.of(entry.payload).addMetadata(meta))` shape
  (`KafkaClearingOutboxEventPublisher.kt:28`, `KafkaSepaPaymentEventPublisher.kt:43`,
  `KafkaLendingOutboxEventPublisher.kt:27`) — so the bytes the writer produces are the bytes the
  consumer parses. Producer and consumer do meet.
- `AuditConsumer.eventTime()` accepts exactly one key, `occurredAt`, and `AuditEntry` is persisted
  with `occurred_at_source = INGEST` when it is absent.

So these are **live defects that land in stored rows**, not latent traps. (`DomainEvent` declares
`occurredAt`, so the consumer is reading the right name — only the missing-time case was real.)

## What each payload now carries — the domain fact, not `Instant.now()` at serialisation

| event | instant chosen | why it is the business fact |
|---|---|---|
| `clearing batch.settled` | `batch.settledAt ?: batch.updatedAt` | `ClearingService.settleBatch` stamps `settledAt` on the aggregate before the publisher runs |
| `clearing item.cleared` | `item.updatedAt` | `ClearingItem` has no per-transition column; `updatedAt` is the state-change instant |
| sepa × 5 (VALIDATED, REJECTED×2, PROCESSING, COMPLETED) | the transitioned aggregate's `updatedAt` | `SepaPayment.transitionTo` stamps it with `Instant.now(clock)` AT the state change. Note the `SepaPaymentOutboxMessage.createdAt` sitting right next to it is the outbox ROW's time — a different fact |
| `loan.disbursed` | `saved.disbursedAt` | the loan's own disbursement instant |
| `loan.interest_accrued` | `accruedAt` | the very instant `markAccrued` stamped on the installment |
| `loan.stage_changed` / `loan.provisioned` | `record.createdAt` | the instant this provisioning cycle ran. `asOf` next to it is the accounting DATE — a different fact, deliberately not reused |
| `loan.written_off` / `loan.rescheduled` | `clock.instant()` | `Loan` carries no `writtenOffAt`/`rescheduledAt` column. This is the house convention already used by the 4 sites that were already correct (`TerminationService`, `OriginationDecisionService`) — the honest answer where no truer fact exists |

**One trap worth naming: `OffsetDateTime.toString()` renders a non-UTC offset as `...+01:00`, which
`Instant.parse` REJECTS.** The clearing aggregates hold `OffsetDateTime`, so emitting them verbatim
would produce a payload that *looks* fixed and behaves exactly as broken — the consumer would log
"unparseable occurredAt" and fall back to ingest time. Both clearing values go through
`.toInstant()`, and the test asserts the normalised instant (`13:34:56+01:00` → `12:34:56Z`), so a
regression to the verbatim form is red.

## Did an existing test assert the current (broken) behaviour?

Checked before writing anything. **No.** There was no test for `ClearingEventPublisherImpl` at all,
and `LendingServiceTest` / `SepaPaymentActivitiesImplTest` assert payload substrings for other
fields only — none pinned the absence of a time field. One adjacent anti-pattern spotted and left
alone as out of scope: `ClearingServiceTest` asserts `updatedSlot.captured.settledAt).isNotNull()`,
which passes against `Instant.EPOCH`.

## Both test runs

11 new tests (3 clearing, 3 sepa, 5 lending). Counts read from the **JUnit XML**, including the
SKIPPED column.

**Against `origin/main` (production files reverted, tests kept) — RED:**

```
openbank-clearing-service   3 new tests   3 failed    (NullPointerException: readTree(...).get("occurredAt") is null)
openbank-sepa-payment       tests 101  fail 4  skipped 14   <- 3 are the new tests
openbank-lending-service    tests 223  fail 5  skipped 0    <- all 5 are the new tests
```

The clearing falsification was run the *honest* way. A plain revert of
`ClearingEventPublisherImpl.kt` makes the test file fail to COMPILE (this PR extracts
`batchSettledPayload`/`itemClearedPayload` so a payload can be asserted without a Vert.x context —
`Panache.withTransaction` needs one, a payload does not), and a compile error is not evidence the
assertion works. So the extraction was kept and only the two `occurredAt` map entries were deleted:
**3 tests, 3 failures.**

The 4th sepa failure and the `skipped 14` are environmental, not mine:
`SepaPaymentResourceAuthzTest` fails with `Failed to start quarkus` when a second Gradle build holds
the port, and its skips are the cascade. Re-run alone on this branch it is `tests 101 fail 0
skipped 0`.

**Against this branch — GREEN:**

```
openbank-clearing-service   tests 41   fail 0  skipped 0
openbank-sepa-payment       tests 101  fail 0  skipped 0
openbank-lending-service    tests 223  fail 0  skipped 0
```

A later re-run of lending on a busier machine reported `223 / 2 fail / 24 skipped`, and the two
were **not** graded from their shape — the stack was read to the root cause:
`QuarkusBindException: Port already bound: 8081: Address already in use`, from
`LendingResourceAuthzTest` and `CustomerIntakeConfigInjectionTest`, both `@QuarkusTest`. `lsof`
confirms an unrelated `java` process (pid 71180) holding 127.0.0.1:8081 — a concurrent build on
this shared machine, the failure mode CONTRIBUTING already documents. Re-run alone,
`CustomerIntakeConfigInjectionTest` is `2 / 0 / 0`. The `24 skipped` is the cascade from those boot
failures, not a suite that stopped running; on the clean run the skipped count is 0, which is why
it is quoted here rather than only the failure count.

`:quarkusBuild` (the CDI/ArC wiring check that `test` cannot make) is **green for all three
modules** — that is the run that would catch a bean-graph regression.

All 11 new tests confirmed present by name in the XML (not inferred from a passing total) — the
Kotlin/JUnit5 `= runBlocking { }` silent-drop trap means a green total proves nothing about whether
a method ran.

Assertions are **exact expected instants**, parsed back out of the JSON with `Instant.parse` —
never `isNotNull()` (which passes against `Instant.EPOCH`) and never a string emptiness check
(Jackson's `asText()` yields the four-character text `"null"` for a JSON null).

## Local gate

`:detekt :ktlintCheck :test :quarkusBuild` for all three modules; the task list was read rather
than assumed. One real detekt finding fixed on the way: the added lines pushed
`postProvisioningDelta` to exactly 60 lines and `LongMethod` fires **at** the threshold, not above
it — the rationale moved into the function's KDoc.

## Governance

- **All three services are money-path** (`rules.yaml: money_path_services`): 2 approvals required.
- `check-threat-model-diff.py --base origin/main --enforce` passes: no trust boundary, identity or
  privilege changes here — these are additive payload fields.
- **Additive fields only.** Every consumer of these topics parses via `readTree` and ignores
  unknown fields; no existing key is renamed or removed. `openapi.yaml` is untouched (no REST
  contract changes) and no libs module is touched, so the N-1-dependents hazard does not apply.
- **What this does NOT claim.** The success measure in #3914 is rows moving to
  `audit_entries.occurred_at_source = 'EVENT'` and
  `openbank.audit.event.time.missing{source_service}` going to zero — not "the field appears in the
  payload". That is only observable after these producers have live traffic, so I am claiming the
  payloads now carry the key and that the consumer accepts the exact format, nothing more. #3914
  should stay open until the counter is read.

## What is left

**Nothing in this issue's producer scope** — 13 of 13 remaining sites fixed, so all 7 producers now
emit an event time. Deliberately out of scope and still open:

1. The AsyncAPI discrepancy noted in #3914's body: `docs/asyncapi/openbank-events.yaml` gives the
   dispute channel as `openbank.disputes.dispute.event` while producer and consumer both use
   `openbank.dispute.events`. Untouched.
2. `ClearingEventPublisherImpl.publishItemCleared` has **no production caller** — `command grep -rn
   publishItemCleared openbank-clearing-service/src` finds only the port declaration, the
   implementation and no use site. Its payload is fixed anyway (cheap, and it is a live defect the
   day something calls it), but it is worth recording that one of clearing's two sites is currently
   dead code rather than letting a future audit count it as traffic.
3. `ClearingServiceTest`'s `isNotNull()` assertion on `settledAt`, noted above.

### DIFF
--- a/openbank-clearing-service/src/main/kotlin/com/openbank/clearing/infrastructure/kafka/ClearingEventPublisherImpl.kt
+++ b/openbank-clearing-service/src/main/kotlin/com/openbank/clearing/infrastructure/kafka/ClearingEventPublisherImpl.kt
@@ -30,43 +30,65 @@ class ClearingEventPublisherImpl @Inject constructor(
         val message = OutboxMessage(
             aggregateId = batch.id,
             eventType = BATCH_SETTLED_EVENT,
-            payload = objectMapper.writeValueAsString(
-                mapOf(
-                    // eventType is also on OutboxMessage.eventType (-> Kafka header, ce-type), but
-                    // AuditConsumer (PR #1007) reads only the JSON body, so it is duplicated here.
-                    "eventType" to BATCH_SETTLED_EVENT,
-                    "batchId" to batch.id,
-                    "batchReference" to batch.batchReference,
-                    "rail" to batch.rail.name,
-                    "cycleId" to batch.cycleId,
-                    "totalDebit" to batch.totalDebit,
-                    "totalCredit" to batch.totalCredit,
-                    "netPosition" to batch.netPosition,
-                    "itemCount" to batch.itemCount,
-                    "settledAt" to batch.settledAt?.toString(),
-                ),
-            ),
+            payload = batchSettledPayload(batch),
         )
         return Panache.withTransaction { outboxRepo.persistInTransaction(message) }
     }
+
+    /**
+     * The `batch.settled` JSON body, split out from [publishBatchSettled] so it can be asserted
+     * without a Vert.x context (`Panache.withTransaction` needs one; a payload does not). #3914.
+     */
+    internal fun batchSettledPayload(batch: ClearingBatch): String = objectMapper.writeValueAsString(
+        mapOf(
+            // eventType is also on OutboxMessage.eventType (-> Kafka header, ce-type), but
+            // AuditConsumer (PR #1007) reads only the JSON body, so it is duplicated here.
+            "eventType" to BATCH_SETTLED_EVENT,
+            "batchId" to batch.id,
+            "batchReference" to batch.batchReference,
+            "rail" to batch.rail.name,
+            "cycleId" to batch.cycleId,
+            "totalDebit" to batch.totalDebit,
+            "totalCredit" to batch.totalCredit,
+            "netPosition" to batch.netPosition,
+            "itemCount" to batch.itemCount,
+            "settledAt" to batch.settledAt?.toString(),
+            // #3914: the business event time, not the serialisation time. `settledAt` IS
+            // the settlement instant (ClearingService.settleBatch sets it on the aggregate
+            // before this publisher ever runs); `updatedAt` is the fallback for a batch
+            // whose settledAt was never stamped. Emitted as an Instant, NOT via
+            // OffsetDateTime.toString() — AuditConsumer.eventTime() parses with
+            // Instant.parse, which rejects a non-Z offset ("...+01:00") and would fall
+            // back to ingest time, i.e. exactly the defect this line exists to fix.
+            "occurredAt" to (batch.settledAt ?: batch.updatedAt).toInstant().toString(),
+        ),
+    )
 
     override fun publishItemCleared(item: ClearingItem): Uni<Void> {
         val message = OutboxMessage(
             aggregateId = item.id,
             eventType = ITEM_CLEARED_EVENT,
-            payload = objectMapper.writeValueAsString(
-                mapOf(
-                    "eventType" to ITEM_CLEARED_EVENT,
-                    "itemId" to item.id,
-                    "batchId" to item.batchId,
-                    "paymentId" to item.paymentId,
-                    "paymentReference" to item.paymentReference,
-                    "amount" to item.amount,
-                    "currency" to item.currency,
-                    "status" to item.status.name,
-                ),
-            ),
+            payload = itemClearedPayload(item),
         )
         return Panache.withTransaction { outboxRepo.persistInTransaction(message) }
     }
+
+    /** The `item.cleared` JSON body. See [batchSettledPayload] for why this is split out. */
+    internal fun itemClearedPayload(item: ClearingItem): String = objectMapper.writeValueAsString(
+        mapOf(
+            "eventType" to ITEM_CLEARED_EVENT,
+            "itemId" to item.id,
+            "batchId" to item.batchId,
+            "paymentId" to item.paymentId,
+            "paymentReference" to item.paymentReference,
+            "amount" to item.amount,
+            "currency" to item.currency,
+            "status" to item.status.name,
+            // #3914: ClearingItem has no per-transition instant, so `updatedAt` — the
+            // instant the row last changed state, which for a cleared item IS the clearing
+            // — is the truest available business time. Same Instant-vs-offset note as
+            // publishBatchSettled above.
+            "occurredAt" to item.updatedAt.toInstant().toString(),
+        ),
+    )
 }
--- a/openbank-clearing-service/src/test/kotlin/com/openbank/clearing/infrastructure/kafka/ClearingEventPublisherOccurredAtTest.kt
+++ b/openbank-clearing-service/src/test/kotlin/com/openbank/clearing/infrastructure/kafka/ClearingEventPublisherOccurredAtTest.kt
@@ -1,0 +1,97 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) OpenBank contributors. Licensed under the Apache License, Version 2.0.
+// See LICENSE in the repository root or https://www.apache.org/licenses/LICENSE-2.0 for details.
+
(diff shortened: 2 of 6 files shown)


JiRaska:
Reviewed to money-path standard (touches three money-path services — lending, sepa-payment, clearing — not just lending; worth a title/scope note but not a defect).

**No producer substitutes construction time for business time**, checked per site: lending uses `disbursedAt`/`accruedAt`/`createdAt` on the aggregate; sepa-payment uses the aggregate's own `transitionTo`-stamped `updatedAt`; clearing uses `settledAt`/`updatedAt`. Two lending events (write-off, reschedule) use `clock.instant()` at completion because `Loan` has no stored timestamp for those transitions — disclosed in comments, matches the existing convention in the same file, and would need a schema change to do better. Worth noting, not blocking.

**Falsified, not just tested green**: reverted the write-off fix in a worktree, ran the specific test — failed with a clean `NullPointerException` on the missing `occurredAt` key (not a wrong-value failure, a missing-key one), confirming the test genuinely detects the reverted defect. Restored, confirmed clean.

**Coverage checked completely, not sampled**: lending never constructs `AuditEvent` directly — traced the real construction site (`LendingOutboxMessage`), found 10 across the service, confirmed the 4 pre-existing ones already carried `occurredAt` from an earlier PR and this PR fixes exactly the remaining 6. No site left defaulting to consumer ingest-time.

**Test quality**: grepped all three changed test files for `isNotNull()` near timestamps — zero as actual assertions, only in comments warning against the pattern. Every new assertion checks an exact expected `Instant`, including UTC-normalization of a non-Z offset in the clearing test.

Clock injection is consistent throughout (no raw `Instant.now()` in touched code), and no `runBlocking` silent-skip risk in any changed test signature.

PR #4413, opened 2026-08-09
feat(libs): ADR-0252 — the taint's trust boundary at the edge

JiRaska:
**Stacked on #4371** (base = `feat/synthetic-taint-primitive`). That PR defined the flag; this is
the edge that decides whether to believe an inbound claim of it.

## The hole this exists to close

The taint's purpose is to exclude activity from regulatory aggregates and the AML baseline. So a
header that **any** caller could set would let anyone mark their own payments as synthetic and drop
themselves out of AML scoring and the regulatory returns.

That is a self-service evasion primitive — and it would be introduced by the mechanism built to make
the platform more honest. It is the obvious way to build this and it is wrong.

So:

- the header is honoured **only** from a principal named in `openbank.synthetic.trusted-principals`;
- that list is **empty by default**, so merging this changes nothing anywhere until an operator
  names a canary principal in one environment;
- an anonymous request can never taint, whatever it sends;
- an untrusted attempt is refused **and logged at WARN with the principal name**. Nobody sends that
  header by accident, so one occurrence in production is either a misconfigured canary or someone
  probing for exactly this.

A refused claim does **not** become a 4xx. The request is legitimate; only its claim is not, and
rejecting it would turn a monitoring feature into an availability risk on every service's request
path. Declining to believe a claim costs nothing.

## Blast radius, handled deliberately

`@Provider` means this runs on every request of every service that depends on libs-runtime. So it
injects nothing but configuration and reads only the JAX-RS `SecurityContext` — no
`SecurityIdentity`, no registry, nothing that can be unsatisfied at build time or throw at request
time. The config field is `Optional<String>`, not a bare `String`, or an unset property is
`SRCFG00040` at boot on every service.

The response filter clears the MDC key. Worker threads are pooled, so a leaked entry would mark the
**next** request as synthetic — a real customer's log lines on the wrong side of the taint, with
nothing downstream ever questioning it.

## Verification

- 9 tests, **0 failures, 0 skipped** (read from the test XML).
- **The security property is falsified, not assumed:** replacing the trust check with `true` makes
  the suite exit 1; restoring it exits 0. Both measured.
- `:openbank-notification-service:quarkusBuild` — a new `@Provider` in libs-runtime is a fleet-wide
  CDI change, and unit tests cannot see ArC failures.
- `ktlintCheck`, `detekt` clean.

## What this does not do yet

Nothing reads `SYNTHETIC_TAINT_PROPERTY` yet. The producers stamping it onto outbox rows, the
consumers reading it off Kafka, and the exclusion in FINREP/COREP/AnaCredit and the AML baseline are
separate PRs — several money-path, so two human approvals by rule. Tracked in #4348.

Controls (screening, VoP, SCA, limits) must **always** run for synthetic traffic; the taint never
suppresses a control. That asymmetry is enforced by the API surface in #4371, which has no
bypass-shaped helper and a test pinning its absence.

Refs #4348, ADR-0252.


JiRaska:
Rebased onto `main` (its base #4371 merged) and fixed a real defect CI caught — worth recording
because it is the exact failure class this programme exists to end.

`configproperty-kotlin-defaults` flagged:

> `'trustedPrincipals' has the Kotlin default = Optional.empty()`. That generates a synthetic
> constructor, Arc builds the bean through it, and the `@ConfigProperty` is never applied — the
> field silently keeps this value whatever the environment says.

So `openbank.synthetic.trusted-principals` could **never have been read**. No principal would ever
be trusted, the taint could never be honoured anywhere, and a canary would have looked exactly like
one nobody had configured yet. A feature that cannot be turned on is worse than one that is off,
because the operator who turns it on stops looking. Nothing in the unit tests could see it — they
set the field directly, which is how the lending intake bean passed for its whole dead life.

Now `lateinit` with `::trustedPrincipals.isInitialized` guarding the read. Uninitialized means trust
nobody: the same answer as an empty list, fail-to-real like the rest of the filter.

**One note for whoever maintains that gate:** its docstring scopes it to *constructor parameters* and
says a `@ConfigProperty` on a **field** is a different mechanism and out of scope. Its regex matches
fields too — which is how it caught this one — and the 99 baselined entries suggest most of that
baseline is fields. The stricter behaviour is the useful one; the docstring is what is stale. Not
changed here, since it is not this PR's file and the fix direction is a judgement call.

Verification after the change: 9 tests / 0 failures / 0 skipped, `kotlin` gate shard 19/19.

JiRaska:
Added the outbound half in the same PR — the trust boundary and the propagation are one decision and
read better together.

`SyntheticTaintClientFilter` forwards the header on a service-to-service hop. Without it the inbound
filter's decision dies at the first outbound call: service A knows the flow is synthetic, service B
does not, and the same journey is half tainted — every downstream write landing in the aggregates
the taint exists to keep it out of.

**Registration is explicit, and I am flagging it rather than papering over it.** The fleet convention
is `@RegisterProvider` on each `@RegisterRestClient` interface. There is no `@Provider`-annotated
client filter anywhere in this tree, so nothing here demonstrates that global discovery would work,
and asserting it would be a guess about Quarkus dressed as a fact about the repo. A client that does
not register this filter drops the taint.

That failure is in the **safe** direction: a dropped taint means the hop is treated as real, so
synthetic activity reaches an aggregate it should not — visible and bounded — rather than real
activity vanishing from one, which is neither. Same reasoning for MDC as the source: a reactive hop
that loses MDC drops the taint, not the other way round.

Two follow-ups, both in #4348:
1. roll `@RegisterProvider(SyntheticTaintClientFilter::class)` across the client interfaces, with a
   gate so a new client interface cannot forget it;
2. replace the MDC read with a request-scoped holder plus explicit context propagation — tighter,
   and not free, because such a bean injected into a client filter running outside a request context
   is an error rather than a `false`.

13 tests across the two filters, 0 failures, 0 skipped. Both conditions falsified by inversion:
trust check → `true` and MDC check → `true` each make their suite exit 1; restoring exits 0.

JiRaska:

**Verified (re-ran independently, not just trusting the review):**
- Trust-boundary gating is principal-allowlist-based, not header-controlled — confirmed by hardcoding `trusted=true` and watching 4/9 tests correctly go red.
- Fail-to-real on every early-return path (no header, unparseable, unauthenticated, untrusted, empty config).
- `SyntheticTaintRequestFilter`/`SyntheticTaintResponseFilter` auto-discovered via `@Provider` + jandex, same mechanism as `CorrelationIdRequestFilter`.
- `openbank-libs-domain` untouched — domain-purity gate N/A, correctly skipped.
- All 13 tests genuinely run (no JUnit5 silent-skip shape).

**On the MDC-cleanup question I checked myself:** `SyntheticTaintResponseFilter.filter()` calls `MDC.remove(MDC_SYNTHETIC)` unconditionally on every response — not only when the request filter set the key. JAX-RS response filters run for every request, including one whose exception was mapped to a response, so this clears on the *same* request that may have set it, not on the *next* one sharing a pooled thread. I don't think there's a live leak here; a scratch test that pre-seeds MDC and calls only the request filter (skipping the response filter) will of course still see the stale value, but that's not the code path a real request takes. No change needed.

**Cosmetic, non-blocking:** `SyntheticTaintClientFilter.kt:49` hardcodes the literal `"true"` instead of `SyntheticTaint.TRUE_VALUE`. Worth a one-line follow-up, not a blocker.

PR #4419, opened 2026-08-09
test(document-service): replay the domestic-payment and statement contracts on every PR

JiRaska:
## What

Closes the contract-coverage item of #4109: the remaining two of the three merged services holding
a live **synchronous** REST client against `openbank-document-service` now publish a consumer pact,
and all three are replayed against the real provider on every pull request.

| consumer | client | covered by |
| --- | --- | --- |
| `openbank-sepa-payment` | `DocumentServiceClient` | `SepaPaymentDocumentServicePactConsumerTest` (#4299) |
| `openbank-domestic-payment` | `DocumentServiceClient` | `DomesticPaymentDocumentServicePactConsumerTest` (**this PR**) |
| `openbank-statement-service` | `DocumentRestClient` | `StatementDocumentServicePactConsumerTest` (**this PR**) |

All three make the same two calls (ADR-0248 #3), neither of which persists anything on the
document-service side: `GET /api/v1/documents/templates` to resolve the PUBLISHED `bodyHtml` for a
template code (there is no get-by-code route), then `POST /api/v1/documents/templates/preview` to
merge the caller's data into that body.

Both payment services are **money-path** (`rules.yaml: money_path_services`), so this PR needs
**2 approvals**.

## The asymmetry IS the test

In each consumer test the expected path is a string **LITERAL** (`TEMPLATES_PATH`, `PREVIEW_PATH`,
`LIMIT_PARAM`); only the **outgoing request** is reflected off the production client's `@Path` /
`@QueryParam`. Deriving both sides from the annotation is DRY and vacuous — expectation and request
move together, so the test stays green against a client pointing at a route that does not exist
(#2290). Each test also asserts the reflected value equals the literal, so a reflection helper that
silently returned an empty or partial path cannot leave both pact tests passing against a contract
that pins nothing.

## Provider replay

`DocumentPactProviderVerificationTest` (openbank-document-service) is **extended, not duplicated** —
a second broker-sourced `@Provider` class for the same provider collides, since each fetches every
pact the broker holds. It is `@PactFolder("../pacts")` and **ungated**, so it replays every
committed pact naming this provider on every PR; a `@PactBroker` class would not, because the PR
lane blanks `PACT_BROKER_URL` (ADR-0056) and its `@EnabledIfSystemProperty(pactbroker.url)` gate
would skip (#2327). Neither consumer needed a new `@State` handler: both declare the same two states
the sepa pact already did, which are properties of the provider (seeded templates, a stateless
renderer) rather than of any one consumer. Confirmed no `build.gradle.kts` exclusion drops the class
in CI.

## FALSIFICATION — proving the consumer pact alone cannot catch a wrong path

Not asserted — **measured**. The experiment simulates the realistic failure: the whole repo agrees
on a path the provider does not serve. `openbank-domestic-payment`'s client `@Path("/templates")`
**and** the test's `TEMPLATES_PATH` literal were repointed together to
`/api/v1/documents/templates-not-served`, the pact regenerated, then both sides run. Verdicts read
from the JUnit XML, never the console — pact-jvm prints `Not all of the N were verified` even on a
fully green run.

| step | consumer (`:openbank-domestic-payment:test`) | provider replay (`:openbank-document-service:test`) |
| --- | --- | --- |
| baseline (path correct) | 2 tests / 0 failures / 0 skipped | **6 tests / 0 failures / 0 skipped** |
| path repointed to a route the provider does not serve | **2 tests / 0 failures / 0 skipped — still GREEN** | **6 tests / 1 failure / 0 skipped — RED** |
| restored | 2 / 0 / 0 | **6 / 0 / 0 — green again** |

The single provider failure is exactly the broken interaction, and only it:

```
  pass  openbank-sepa-payment      - GET  the document templates ...
  pass  openbank-sepa-payment      - POST a template body plus payment data ...
  pass  openbank-statement-service - GET  the document templates ...
  pass  openbank-statement-service - POST a template body plus statement data ...
  FAIL  openbank-domestic-payment  - GET  the document templates ...
  pass  openbank-domestic-payment  - POST a template body plus domestic payment data ...
```

The committed pact under the broken run asked for `GET /api/v1/documents/templates-not-served`, and
the Pact mock server answered it happily — which is the whole point: **the consumer pact could not
tell.** Only the replay against the running provider went red (#2269).

The skipped counts are quoted deliberately: a module whose Quarkus boot fails reports every
`@QuarkusTest` as SKIPPED and reads as a pass. Zero skipped throughout, and the provider run is 6
interactions, so all three consumers really were replayed rather than silently absent.

## Gate output

```
$ python3 .github/scripts/check-pact-provider-replay.py
Committed pacts: 45
Replayed on a PR by an always-running @PactFolder class: 45
Declared uncovered (backlog, #2327): 0
  RUNS  openbank-document-service  .../DocumentPactProviderVerificationTest.kt  [runs]
OK — every committed pact is either replayed on a PR or a declared, still-accurate exception.

$ BASE_REF=main PR_DIFF_BASE=$(git merge-base HEAD origin/main) \
    python3 .github/scripts/run-gates.py --only pact-provider-replay-coverage
1 gates  PASS=1
```

`KNOWN_UNCOVERED` stays **empty** — both new pacts count as covered, so the total moved 43 -> 45
with the covered count moving with it.

**Drift scope** (`derive-pact-drift-scope.sh`, exit 0) now lists `openbank-domestic-payment` and
`openbank-statement-service` among its gated modules. This was falsified rather than assumed: run
against the two new tests *before* their pacts were committed, the script failed with
`a consumer test generates openbank-domestic-payment-openbank-document-service.json but
pacts/... is not committed` for both — proof the scope derivation actually reaches the new modules
rather than silently ignoring them.

**Local gate:** `ktlintCheck` + `detekt` + the consumer tests green on all three touched modules.

## Also

- `openbank-statement-service/build.gradle.kts` gains `libs.pact.consumer` and the standard
  `pact.rootDir` + broker-property forwarding block. Without `pact.rootDir` the generated pact lands
  in the module's `build/` and the committed `pacts/*.json` is never rewritten — the drift gate would
  then be green about work it never did.
- Corrected the now-stale comment on document-service's `libs.pact.provider` dependency, which named
  sepa-payment as the only consumer.

Refs #4109

### DIFF
--- a/openbank-document-service/build.gradle.kts
+++ b/openbank-document-service/build.gradle.kts
@@ -53,11 +53,12 @@ dependencies {
     testImplementation(libs.quarkus.test.security)
     testImplementation(libs.assertj)
     testImplementation(libs.mockk)
-    // ADR-0063: document-service is a pact PROVIDER. sepa-payment renders the customer's payment
-    // confirmation through this service's template list + preview endpoints (ADR-0248 #3), and
-    // until DocumentPactProviderVerificationTest nothing replayed that contract — the only cover
-    // was sepa-payment's own WireMock stub, which is written from the client and so cannot
-    // disagree with it (the #2269 shape).
+    // ADR-0063: document-service is a pact PROVIDER. Three merged services render documents
+    // through this service's template list + preview endpoints (ADR-0248 #3) — sepa-payment and
+    // domestic-payment (payment confirmations) and statement-service — and until
+    // DocumentPactProviderVerificationTest nothing replayed those contracts. The only cover was
+    // each consumer's own WireMock/mock stub, written from the client and so unable to disagree
+    // with it (the #2269 shape).
     testImplementation(libs.pact.provider)
     // AuditEventTime — the ONE copy of the rule openbank-audit-service's AuditConsumer applies to
     // a domain-event payload, so this service's producer tests assert against the real contract
--- a/openbank-document-service/src/test/kotlin/com/openbank/document/contract/DocumentPactProviderVerificationTest.kt
+++ b/openbank-document-service/src/test/kotlin/com/openbank/document/contract/DocumentPactProviderVerificationTest.kt
@@ -21,9 +21,20 @@ import org.junit.jupiter.api.extension.ExtendWith
 
 /**
  * Provider-side Pact verification for the contracts consumers publish against
- * `openbank-document-service`. The first is `openbank-sepa-payment`'s payment-confirmation render
- * (ADR-0248 #3): `GET /api/v1/documents/templates` then
- * `POST /api/v1/documents/templates/preview` — see `SepaPaymentDocumentServicePactConsumerTest`.
+ * `openbank-document-service`. `@PactFolder` replays EVERY pact in `../pacts` naming this provider,
+ * so the set below is what is committed today rather than a list this class maintains — all three
+ * merged services holding a live synchronous REST client against document-service, each making the
+ * same two calls (ADR-0248 #3): `GET /api/v1/documents/templates` then
+ * `POST /api/v1/documents/templates/preview`.
+ *
+ *  - `openbank-sepa-payment` — `SepaPaymentDocumentServicePactConsumerTest` (#4299)
+ *  - `openbank-domestic-payment` — `DomesticPaymentDocumentServicePactConsumerTest`
+ *  - `openbank-statement-service` — `StatementDocumentServicePactConsumerTest`
+ *
+ * All three declare the same two provider states, which is why adding them cost no new
+ * `@State` handler: the states are properties of the provider (seeded templates, a stateless
+ * renderer), not of any one consumer. Deliberately ONE class — a second broker-sourced `@Provider`
+ * class for the same provider collides, since each fetches every pact the broker holds.
  *
  * **Why this class is the load-bearing half.** A consumer pact ALONE cannot catch a wrong request
  * path: the Pact mock server answers whatever path the client asks for, so pointing a client at a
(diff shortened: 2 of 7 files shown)


JiRaska:

**Verified by reading:**
- `DocumentPactProviderVerificationTest` is the correct shape: `@PactFolder`, no `@EnabledIfSystemProperty`, one class — not colliding with a broker-sourced verifier.
- Both new consumer tests use literal path constants for the expectation and reflect the actual path off the production client for the request — the correct asymmetric shape (not the vacuous derive-both-sides pattern), and both assert the reflection helper matches the literal as a guard on themselves.
- `build.gradle.kts` wiring has no `CI=true`-style exclusion anywhere; `openbank-statement-service` correctly forwards `pact.rootDir`/`pactbroker.*` system properties.
- Both new pact JSON files are pure additions, content matches the consumer tests' literal paths/bodies on inspection. `check-pact-provider-replay.py` classifies `openbank-document-service` as ungated `@PactFolder` replay (`RUNS`), not exception-listed.
- `mergeStateStatus=UNSTABLE` root cause is `CodeQL (java-kotlin, manual)` — a runner shutdown signal (exit 143) mid-compile on `openbank-sepa-instant`, unrelated to this diff and not in the required-checks set. Non-blocking flake.
- No `runBlocking`/silent-skip shape in any of the 3 new test files.

**Disclosing a gap honestly:** the shared build machine had 7-8 other agent sessions' Gradle daemons running; two attempts to run the provider-verification test and a deliberate broken-path falsification both stalled 10+ min in configuration and were killed rather than burn more shared CPU. Substituted CI's own job log as secondary evidence — `BUILD SUCCESSFUL`, no `ignoreFailures` anywhere in the module, and the console shows the expected misleading-but-harmless `Not all of the 2 were verified` message for all three consumers, which only prints on a real (passing) verification run. That's not the same as a self-run green I'd normally insist on, so flagging it rather than presenting it as fully closed. If you want it airtight, worth a targeted re-run once the machine is less contended.

Issue #3994, opened 2026-08-07
77% of the audit trail is unattributed: only customer-edge populates sourceService, 76% of rows name no actor, and the consumer's defaults make both invisible


JiRaska:
**`UNKNOWN` is the third most common event type in the production audit trail (124 of ~1700 rows), and
every one of them is a money-path payment status transition.**

Found while measuring #3914's `occurred_at_source` instrumentation against the live audit database.

```
$ psql -d openbank_audit -c "SELECT event_type, count(*) FROM audit_entries GROUP BY 1 ORDER BY 2 DESC"
BALANCE_UPDATED                        296
SCA_DEVICE_ENROLLED                    256
UNKNOWN                                124     <-
KYC_STATUS_CHANGED                     107
HOLD_RELEASED / HOLD_PLACED            210
account.statement.period.closed.v1      78
…

$ … WHERE event_type='UNKNOWN' GROUP BY source_service
unknown  124        <- all 124, also unattributed
```

A sample payload:

```json
{"paymentId":"b0fd3b03-…","previousStatus":"SENT_TO_CLEARING","newStatus":"SETTLED",
 "rejectReason":null,"rejectDetail":null,"occurredAt":"2026-07-24T16:04:05.707Z"}
```

That is `DomesticPaymentStatusChangedEvent` — a **payment settling**.

## Why it lands as UNKNOWN

`AuditConsumer.kt:43`:

```kotlin
eventType = node["eventType"]?.asText() ?: node["type"]?.asText() ?: "UNKNOWN",
…
sourceService = node["sourceService"]?.asText() ?: "unknown",
```

and the producer (`DomesticPaymentEvents.kt:31`) has neither field:

```kotlin
data class DomesticPaymentStatusChangedEvent(
    val paymentId: UUID,
    val previousStatus: DomesticPaymentStatus,
    val newStatus: DomesticPaymentStatus,
    val rejectReason: String?,
    val rejectDetail: String?,
    val occurredAt: Instant,
)
```

No discriminator, no `sourceService`. Note `DomesticPaymentCreatedEvent` in the same file has the same
omission, so this is the whole producer rather than one event.

## Why it matters more than a cosmetic label

The row **is** recorded, with correct time (`occurredAt` is present, so these classify as `EVENT` under
#3907's new column) and a full payload. What is lost is everything you would query by:

- **Attribution** — 124 money-path audit records say they came from `unknown`. A "what did
  domestic-payment do" question returns nothing.
[...]

PR #4424, opened 2026-08-09
fix(audit): give the unattributed producers a real actor, and a way to say nobody did it

JiRaska:
## What

The **producer half** of #3994. The consumer half already shipped: #4270 recovered the producer and
event type the transport was carrying, #4307 stopped a JSON null becoming the four-character actor
`"null"`. What was left is the larger part — rows whose producer emits **no actor key at all**, so
there has never been anything for `AuditConsumer` to read.

`Refs #3994` — it does not close it; see *Not fixed* below.

## The survey, re-derived from the stored payloads (not from the code)

Read-only against the live `audit` CNPG cluster on 2026-08-09/10 (`SELECT` only; `audit_entries`
carries `DO INSTEAD NOTHING` rules on UPDATE/DELETE, so **a backfill is not available at all** and
none is proposed — every fix here is forward-only).

```sql
SELECT count(*) FILTER (WHERE actor_id IS NULL)                                AS null_actor,
       count(*) FILTER (WHERE actor_id = 'null')                               AS str_null,
       count(*) FILTER (WHERE actor_id IS NOT NULL AND actor_id <> 'null')     AS attributed,
       count(*)                                                                AS total
FROM audit_entries;
--  1341 | 7 | 441 | 1789
```

Then the keys each unattributed payload actually carries, from the payload itself:

```sql
SELECT event_type, string_agg(DISTINCT k, ',' ORDER BY k)
FROM (SELECT event_type, jsonb_object_keys(payload::jsonb) AS k
      FROM audit_entries
      WHERE (actor_id IS NULL OR actor_id = 'null') AND payload ~ '^\s*\{') s
GROUP BY 1;
```

| event_type | rows | actor-ish key in the stored payload |
|---|---:|---|
| `BALANCE_UPDATED` | 314 | none |
| `UNKNOWN` (topics with no discriminator) | 131 | one sub-shape has `reviewedBy` |
| `HOLD_PLACED` | 114 | none |
| `HOLD_RELEASED` | 114 | none |
| `KYC_STATUS_CHANGED` | 107 | none |
| `TransactionInitiated` | 79 | `initiatedByPartyId` **present but null/absent on these rows** |
| `account.statement.period.closed.v1` | 78 | none |
| `TransactionCompleted` | 76 | none |
| `AccountCreated` | 63 | none |
| `AccountStatusChanged` | 63 | none |
| `KYC_CASE_OPENED` | 62 | none |
| `PARTY_CREATED` | 61 | none |
| `KYC_CASE_APPROVED` | 55 | none |
| `account.statement.period.close_failed.v1` | 8 | none |
| `TransactionFailed` | 7 | none |
| `ConsentGranted` / `ConsentRevoked` | 11 | none |
| `PARTY_UPDATED` / `PARTY_MERGED` | 3 | none |
| `complaint.received` | 2 | none |

Deriving this from stored payloads rather than from source sidesteps the blind spot below entirely:
the database sees the wire, and the wire is the thing that has two different sources.

## The blind spot, stated per producer

This repo builds event payloads two ways and **a grep finds only one of them**.

| producer | how the payload is built | can `command grep '"actorId"'` see it? | method I used |
|---|---|---|---|
| balance-service | `data class BalanceEvent` → `ObjectMapper.writeValueAsString` | **No** — keys are Kotlin property names, no literal exists in the module before or after this PR | read the type + the publisher; asserted the **serialised JSON** in a test |
| kyc-service | hand-built `linkedMapOf` in `KycEvents` | Yes | read `KycEvent.kt`, grep confirmed |
| party-service | hand-built `linkedMapOf` in `PartyEvents` | Yes | read `PartyEvent.kt`, grep confirmed |
| statement-service | hand-built `"""…"""` raw JSON template | Yes | read both template sites |

So a survey by grep would have reported balance-service — the single largest bucket, 542 rows — as
carrying no event fields worth checking, and been wrong for a structural reason rather than a
missed match.

## The judgement this PR makes explicitly: what a system-originated event says

Most of these events have **no human actor and cannot have one**. Inventing one is worse than
leaving the field absent, and this repo has measured that twice (`"unknown"` reaching 76% of
`source_service`; the string `"null"` in `actor_id`). So:

| state | representation | how it is distinguishable |
|---|---|---|
| nobody did this — a scheduler, a projection, a consumer | `actorType: "SYSTEM"`, `actorId: "system:<service>:<mechanism>"` | reserved `system:` prefix — never a UUID, a Keycloak subject or an email |
| a person did this | the real subject id + its own type (`CUSTOMER`, `REVIEWER`) | not `system:`-prefixed |
| **we failed to record who did this** | **key absent → NULL, exactly as today** | unchanged, still the state to go fix, still counted |

`EventActor` (new, `openbank-libs-domain`, pure Kotlin, no CDI producer) is the one spelling. The
`mechanism` segment is mandatory and non-blank: "some automation in balance-service" is not an
answer anyone can act on, and an optional field is how every call site ends up omitting it.

Rejected `SERVICE` as the type word — `input.principal.type == "SERVICE"` is already structurally
unreachable in the authz layer (`rules.yaml: authz_policy`), and reusing it invites dead rego.

## Fixed — 916 of the 1348 unattributed rows

| service | event types | rows | actor |
|---|---|---:|---|
| balance-service | `BALANCE_UPDATED`, `HOLD_PLACED`, `HOLD_RELEASED` | 542 | `SYSTEM`, three distinct mechanisms: `balance-api`, `ledger-projection`, `value-date-roll` |
| party-service | `PARTY_CREATED`, `PARTY_UPDATED`, `PARTY_MERGED`, `KYC_STATUS_CHANGED` | 171 | customer's Keycloak subject on self-registration; `SYSTEM` with separate `kyc-status-projection` / `aml-status-projection` / `party-api` / `party-merge` mechanisms elsewhere |
| kyc-service | `KYC_CASE_OPENED`, `KYC_CASE_APPROVED` (+ the other two builders) | 117 | **the real four-eyes reviewer** from `KycCase.reviewedBy`; `SYSTEM` only where no human touched the case |
| statement-service | `…period.closed.v1`, `…period.close_failed.v1` | 86 | `SYSTEM` / `period-close` |

Two deliberate non-changes inside the fixed set:

- **`PARTY_ERASED` keeps no actor.** Its envelope is narrowed on purpose for GDPR Art. 17, and the
  plausible actor for a self-service erasure is the data subject's own identifier — which must not
  be re-broadcast by the event that erases them. Zero live rows.
- **`createParty` gets the `SYSTEM` id, not the party's own id.** The B1 invariant makes `cmd.id`
  equal the Keycloak subject for self-service onboarding, but onboarding-service drives the same
  method, so attributing every creation to the party itself would be plausible, confident and
  sometimes false — strictly worse than a true `SYSTEM`.

### Shared publisher vs. call sites

Preferred the shared path where it exists and said so where it does not. There is **no single
shared publisher** to hang this on: `DomainEvent` (libs-domain) is extended by only some of these
producers, and the rest hand-build maps or raw JSON. A request-scoped CDI actor holder in
libs-runtime was rejected outright — a `@Produces` there is injected into every consuming service
whether it wants one or not. So the shared part is the **representation** (`EventActor`), and the
value is supplied per call site, which is also the only correct answer: the honest actor genuinely
differs per entry path (kyc's reviewer vs. its consumer-opened case; party's self-registration vs.
its AML projection), and a default would let the wrong one be inherited silently.

## Not fixed — 432 rows, with reasons

| service | event types | rows | why not here |
|---|---|---:|---|
| transaction-service | `TransactionInitiated` / `Completed` / `Failed` | 162 | money-path; `initiatedByPartyId` already exists on the type but is null on these rows — the fix is upstream in whoever constructs the command, a separate investigation, not an envelope change |
| — | the `UNKNOWN` bucket (payloads with no discriminator) | 131 | not one producer: the bucket is topics whose body carries no event type, several with a `reviewedBy` already present. Needs the per-topic work #4270 started, not an actor decision |
| account-service | `AccountCreated`, `AccountStatusChanged` | 126 | money-path; same shape as party but a larger `DomainEvent` subclass surface — deliberately left so this PR stays reviewable |
| consent-service | `ConsentGranted`, `ConsentRevoked` | 11 | money-path |
| complaint | `complaint.received` | 2 | tail |

**Money-path** (`rules.yaml: money_path_services`): `openbank-balance-service` is money-path, so this
PR needs **2 approvals** and its threat model consulted (`docs/threat-models/`). The three services
left largest — transaction, account, consent — are all money-path too, which is the honest reason
they are a follow-up rather than a bigger diff here.

## Verification — both runs

Read the existing tests before writing any: **no test encoded the unattributed behaviour** (the
nearest was `PartyEventEnvelopeContractTest`, which pins a field-name contract and an explicit
forbidden-field list for `PARTY_ERASED`, neither of which asserts an absent actor). The pacts assert
only a subset of each envelope, so the added keys are compatible.

Every assertion is on an **exact expected value**. `isNotNull()` would have passed against the empty
string, against `"null"`, and — in the kyc pair — against *both* of the two states the change exists
to separate.

### Against `origin/main` (RED)

Copied the two tests that compile unchanged against main into a detached `origin/main` worktree:

```
KycEventActorTest > an approved case names the reviewer who approved it()            FAILED
   expected: "analyst.novak@openbank.cz"                     but was: null
KycEventActorTest > a case nobody has reviewed says so...()                          FAILED
   expected: "system:kyc-service:case-lifecycle"             but was: null
KycEventActorTest > a blank reviewer is treated as no reviewer...()                  FAILED
   expected: "system:kyc-service:case-lifecycle"             but was: null
KycEventActorTest > the two states stay distinguishable...()                         FAILED
   NullPointerException: null cannot be cast to non-null type kotlin.String
StatementServiceTest > the period-closed payload names the scheduled close...()      FAILED
   NullPointerException: JsonNode.asText() on the return of JsonNode.get("actorId")
```

The balance, party and `EventActor` tests cannot be run against main at all — the field and the type
they assert do not exist there, so they fail at compile. Stating that rather than dressing it up as
a nicer red.

### Against this branch (GREEN), read from the JUnit XML

| module | tests | failures | **skipped** |
|---|---:|---:|---:|
| `openbank-libs-domain` | 415 | 0 | 0 |
| `openbank-balance-service` | 156 | 0 | 1 |
| `openbank-kyc-service` | 78 | 0 | 1 |
| `openbank-party-service` | 151 | 0 | 1 |
| `openbank-statement-service` | 86 | 0 | 0 |

Skipped counts read from the XML, not the console. The first combined run showed one failure each in
kyc and statement — both `Unable to start HTTP server`, the shared-machine port collision, not a
product failure; re-run serially, both green, and the skipped counts did **not** move (a Quarkus boot
failure would have shown as a jump in skips, not only in failures).

`detekt`, `ktlintCheck`, `test` and `quarkusBuild` all pass on all five modules — task list read, not
assumed. `check-domain-purity.py`: **0 new** violations (the new libs import is pure domain).
`check-license-headers.py`: OK.

### The libs-change blast radius

`openbank-libs-domain` has **50** dependent modules (`command grep -rl openbank-libs-domain
--include=build.gradle.kts`). This change adds one new file containing one new `object` — no existing
signature moves, no `@Produces`, no CDI bean — so no dependent can break by not being updated. The
four services edited here are the only ones that reference it.

### DIFF
--- a/openbank-balance-service/src/main/kotlin/com/openbank/balance/application/usecase/BalanceService.kt
+++ b/openbank-balance-service/src/main/kotlin/com/openbank/balance/application/usecase/BalanceService.kt
@@ -8,6 +8,7 @@ import com.openbank.balance.application.port.`in`.*
 import com.openbank.balance.application.port.out.*
 import com.openbank.balance.domain.model.*
 import com.openbank.libs.domain.calendar.AccountingClock
+import com.openbank.libs.domain.event.EventActor
 import jakarta.enterprise.context.ApplicationScoped
 import jakarta.inject.Inject
 import java.time.Clock
@@ -128,6 +129,8 @@ class BalanceService(
                 availableAmount = updated.availableAmount,
                 reservedAmount = updated.reservedAmount,
                 occurredAt = OffsetDateTime.now(clock),
+                actorId = BalanceEventActors.API,
+                actorType = EventActor.TYPE_SYSTEM,
             ),
         )
 
@@ -158,6 +161,8 @@ class BalanceService(
                 availableAmount = updated.availableAmount,
                 reservedAmount = updated.reservedAmount,
                 occurredAt = OffsetDateTime.now(clock),
+                actorId = BalanceEventActors.API,
+                actorType = EventActor.TYPE_SYSTEM,
             ),
         )
 
@@ -182,6 +187,8 @@ class BalanceService(
                     availableAmount = outcome.balance.availableAmount,
                     reservedAmount = outcome.balance.reservedAmount,
                     occurredAt = OffsetDateTime.now(clock),
+                    actorId = BalanceEventActors.API,
+                    actorType = EventActor.TYPE_SYSTEM,
                 ),
             )
         }
@@ -210,6 +217,8 @@ class BalanceService(
                     availableAmount = outcome.balance.availableAmount,
                     reservedAmount = outcome.balance.reservedAmount,
                     occurredAt = OffsetDateTime.now(clock),
+                    actorId = BalanceEventActors.API,
+                    actorType = EventActor.TYPE_SYSTEM,
                 ),
             )
         }
--- a/openbank-balance-service/src/main/kotlin/com/openbank/balance/application/usecase/LedgerProjectionService.kt
+++ b/openbank-balance-service/src/main/kotlin/com/openbank/balance/application/usecase/LedgerProjectionService.kt
@@ -12,7 +12,9 @@ import com.openbank.balance.application.port.out.BalanceEventPublisher
 import com.openbank.balance.application.port.out.HoldRepository
 import com.openbank.balance.application.port.out.LedgerProjectionPort
 import com.openbank.balance.domain.model.BalanceEvent
+import com.openbank.balance.domain.model.BalanceEventActors
 import com.openbank.balance.domain.model.BalanceEventType
+import com.openbank.libs.domain.event.EventActor
 import com.openbank.libs.observability.DomainMetrics
 import jakarta.enterprise.context.ApplicationScoped
 import org.jboss.logging.Logger
@@ -79,6 +81,8 @@ class LedgerProjectionService(
                     availableAmount = applied.availableAmount,
                     reservedAmount = applied.reservedAmount,
                     occurredAt = OffsetDateTime.now(clock),
+                    actorId = BalanceEventActors.LEDGER_PROJECTION,
+                    actorType = EventActor.TYPE_SYSTEM,
                 ),
             )
         }
--- a/openbank-balance-service/src/main/kotlin/com/openbank/balance/domain/model/Balance.kt
+++ b/openbank-balance-service/src/main/kotlin/com/openbank/balance/domain/model/Balance.kt
@@ -4,6 +4,7 @@
 
 package com.openbank.balance.domain.model
 
+import com.openbank.libs.domain.event.EventActor
 import java.math.BigDecimal
 import java.time.OffsetDateTime
 import java.util.UUID
@@ -172,6 +173,26 @@ data class BalanceHold(
     val releasedAt: OffsetDateTime?,
 )
 
+/**
+ * The three origins this service publishes from (#3994). See [BalanceEvent.actorId] for why all
+ * three are `SYSTEM` and none is a person.
+ *
+ * Named constants rather than string literals at five call sites: the audit trail groups on this
+ * value, so a typo at one site would silently create a fourth origin that looks real.
+ */
+object BalanceEventActors {
+    private const val SERVICE = "balance-service"
+
+    /** An inbound command on the balance API — hold, release, credit, debit. */
+    val API: String = EventActor.system(SERVICE, "balance-api")
+
+    /** The read model catching up to a posting the ledger already made (ADR-0039 Phase D). */
+    val LEDGER_PROJECTION: String = EventActor.system(SERVICE, "ledger-projection")
+
+    /** The daily value-date roll scheduler (ADR-0178, #1745). */
+    val VALUE_DATE_ROLL: String = EventActor.system(SERVICE, "value-date-roll")
+}
+
 enum class BalanceEventType {
     BALANCE_UPDATED,
     HOLD_PLACED,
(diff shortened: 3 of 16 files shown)


JiRaska:
Reviewed. Additive at the shared-type level, honestly partial in coverage, and the safety claims — falsification, domain purity, blast radius — verified by running rather than trusting the write-up. One real logic bug found and confirmed against the code, one standing policy gate, two documentation-lag notes.

**Worth-fixing, and I checked the KDoc's own reasoning against the implementation — they disagree.** `KycEvents.actorType()`:

```kotlin
private fun actorType(case: KycCase): String =
    if (case.reviewedBy.isNullOrBlank()) EventActor.TYPE_SYSTEM else ACTOR_TYPE_REVIEWER
```

`KycService.autoEvaluateAndApprove` (the sandbox straight-through path, `:165`) sets `reviewedBy = "sandbox-auto-approval"` — non-blank, so this resolves to `REVIEWER`. The KDoc directly above explains why that string is safe for `actorId` — *"it is self-evidently not a person, which is the property that matters"* — but `actorType` never checks for that sentinel, only blank-vs-non-blank. So the stated intent and the actual classification disagree: the comment describes a distinction the code doesn't make. A sandbox audit trail will report an automated approval as a human four-eyes decision, the exact confusion this PR exists to prevent. Sandbox is a documented prod no-op so live data is unaffected, but the KDoc reads as though this was already handled.

**Blocking-by-policy, correctly flagged already**: `openbank-balance-service` is money-path and this PR edits `BalanceService.kt`/`LedgerProjectionService.kt`/`Balance.kt`/`ValueDateRollScheduler.kt`, so it needs 2 approvals per `rules.yaml`. Confirmed fresh: `reviewDecision="" reviews=0`. Not a defect — the PR body already states this.

**Verified rather than assumed:**
- `domain-purity-gate` PASS, 0 new violations.
- `BalanceEvent`'s new `actorId`/`actorType` are nullable-defaulted — no fleet-wide compile break; `PartyEvents`' new required `actor` param is a real breaking signature change but scoped entirely inside party-service, and every call site (prod + test) was checked updated.
- Falsified: reverted the statement-service producer fix in a worktree, ran its test — failed red (missing origin claim), restored, confirmed clean.
- `ktlintCheck`/`detekt` clean on all five touched modules.
- No governance/rego files touched, so no OPA bundle or EU AI Act regen owed.

**Notes:**
- `docs/threat-models/openbank-balance-service.md` already claims "movements carry origin/actor" — this PR is what actually makes that true for the first time. Doc isn't dated to reflect the claim just became real.
- Two Pact provider-verification test files (not touched here) comment "mirrors `PartyEvents.created(party, at)`" — the real signature now takes a third `actor` param, so the comment is stale and the mock no longer truly mirrors production output. Harmless today since consumer expectations don't require the new fields, but the class of thing this repo has been burned by.

Issue #4459, opened 2026-08-11
kyc: adverse-media source for perpetual KYC triggers (ADR-0256 D5 followup)


JiRaska:
## Context

ADR-0256 (perpetual KYC) deliberately defers adverse-media triggers (its D5): `CheckType.ADVERSE_MEDIA` exists in the KYC domain model, but the platform has no adverse-media source. An event-driven trigger requires a *change-detectable* source — a stubbed feed would make the catalogue member decorative.

## What this issue tracks

Evaluate and land a real adverse-media source for KYC re-screening:

1. Source selection: vendor feed vs. in-repo ingestion (e.g. OpenSanctions-adjacent media datasets, GDELT-class public sources). Must be change-detectable (push or diff-able), EU-data-residency compatible (ADR-0175), and licensed for this use.
2. A `SanctionsListService`-style import path in sanctions-service or a new bounded context, producing per-entry change events (same content-hash pattern as ADR-0256 D1).
3. Once live, `ADVERSE_MEDIA_CHANGED` joins the ADR-0256 trigger catalogue via its own ADR.

## Explicitly out of scope

- Screening logic changes in kyc-service beyond consuming the new trigger
- Any auto-escalation of risk tier from media hits (four-eyes gate per ADR-0116 stays)

Refs: ADR-0256 D5, ADR-0116 (KYC engine), ADR-0175 (data residency)

JiRaska:
## Unbuilt on `origin/main` @ `77b62467e` — and open PR #5893 deliberately does not close it

First measurement on this thread, so the remainder is on the record.

```
$ git grep -n 'ADVERSE_MEDIA_CHANGED' origin/main
(no output)
$ git grep -lni 'adverse' origin/main -- '*/src/main/**'
   ... aml-service compliance docs, engagement-service adverse-STATE (fraud hold /
       arrears / dispute — a different concept), customer-edge openapi ...
```

No import path, no per-entry change events, no trigger. All three numbered items are at zero.

**Open PR #5893** (`feat/kyc-adverse-media-port-56379`, open since 2026-08-20) touches this area and
says in its own body that it does not close this issue: it lands the port, a fail-closed
`UnconfiguredAdverseMediaSource` and a readiness signal, on the ground that *"no such source is
obtainable from this repo — every candidate needs either a commercial contract (World-Check /
ComplyAdvantage / Refinitiv) or a licensing + residency decision on a public dataset"*. That is the
[...]

Issue #4458, opened 2026-08-11
party-service: publish materiality-classified party change event (ADR-0256 followup)


JiRaska:
## Context

ADR-0256 (perpetual KYC, event-driven re-screening) defines a closed trigger catalogue. One member — `PARTY_DATA_CHANGED` — is deferred: it needs party-service to publish a **materiality-classified** change event before kyc-service can consume it.

Today `KafkaPartyEventPublisher` publishes party events, but no materiality classification exists. A consumer cannot distinguish "customer changed their primary name" (must re-screen identity) from "customer updated a contact preference" (must not trigger KYC review).

## What to build

In `openbank-party-service`:

1. Extend the party change event contract with a `materiality` classification: `MATERIAL` (name, date of birth, residency country, beneficial-owner set) vs `NON_MATERIAL` (contact details, preferences, address-only edits).
2. Materiality is declared by the **publisher** in the event contract — never inferred by the consumer from a generic update payload.
3. Publish on the existing party events channel with schema-versioned contract (event change ⇒ backward-compatible schema version per repo rules).

## Acceptance criteria

- [ ] Event carries an explicit materiality field from a closed enum
- [ ] Only MATERIAL changes are consumable as KYC re-screening triggers
- [ ] Contract test covers the classification boundary
- [ ] AsyncAPI contract updated for the topic (ADR-0006)

Refs: ADR-0256 D1 (deferred trigger), ADR-0247 (same publisher-side classification pattern for fraud hold reason codes)

PR #4468, opened 2026-08-11
feat(sanctions): publish SANCTIONS_LIST_CHANGED event with content-level diff

JiRaska:
## Summary

Publish committed sanctions-list content changes through the existing transactional outbox. A database journal records entry mutations in the same transaction as each import batch; publication drains exactly the selected journal rows only when all event chunks commit. A failed publication can retry without losing changes, even after a content-identical refresh.

This is the producer portion of ADR-0256. Downstream KYC re-screening and operator resolution of withheld publications remain separate work.

## Linked issues

Refs #4460, Refs #4458, Refs #4459.

## Changes

- Add V14: an append-only change journal, mutation trigger and per-list publication state. Canonical comparison ignores ordering-only changes while tolerating malformed legacy metadata.
- Coalesce committed changes in PostgreSQL and publish at most 64 source identifiers per event. Transaction-local selection prevents concurrently committed changes, including lower sequence IDs, from being deleted accidentally.
- Retain evidence and emit a deduplicated count-only signal when the change-share threshold is exceeded or source identities cannot be published. The fingerprint deduplicates retries; it is not an attestation.
- Recover pending publication on the existing scheduled pass. Preserve current typed import outcomes, EU FSF routing, metrics and deferred refresh-all. An unchanged nonempty feed remains a successful import; an empty feed cannot deactivate the existing list.
- Document both event payloads in the AsyncAPI document (contract version 2.3.0), including aggregate identity consumed by audit, and update the sanctions threat model. REST APIs and release-generated versions are unchanged.

## Current integration validation

Head `0e21bb2898687b4e2407fb68f0a2f4ba3a3763c9` is signed and verified. On 2026-09-18, 22 reported checks passed, none failed or remained pending; eight were skipped, including the broker-backed contract job and Admin UI build. The current head also contains follow-up fixes for repaired and removed invalid source identities after the earlier validation below. Exact-head CI does not replace the outstanding independent money-path reviews.

## Earlier integration validation

Head `e03b1269d53216466e9c30c4d792c83eae716eb5` also drains pending journal evidence for disabled lists without re-importing the feed. The focused `SanctionsListServiceTest`, module `ktlintCheck` and `detekt`, `git diff --check`, and signed-commit verification passed locally. Exact-head hosted [CI](https://github.com/JiRaska/open-bank-oss/actions/runs/35283112135), [Services CI](https://github.com/JiRaska/open-bank-oss/actions/runs/35283112368), [CodeQL/Trivy](https://github.com/JiRaska/open-bank-oss/actions/runs/35283112126), and security-regression checks passed on this head. The broker-backed contract job is skipped on this PR; independent money-path review remains outstanding.

## Previous exact-head integration validation

Previous head `b74d431a59d951484ebcdc247d90e2d4ac33d1c0` integrates main `b46189fe74317a7dfac77cdf2fa6c3a3bdcd432e`. The only conflict was the AsyncAPI contract version: main already uses 2.2.0, so the additive sanctions events now use 2.3.0. All existing main schemas/message components, non-sanctions channels and existing sanctions message bindings were compared structurally and preserved.

- Full sanctions `ktlintCheck`, `detekt`, `test`, `koverVerify`, `quarkusBuild`: PASS in 2m25s; 139 tests passed, no failures/errors, one broker-dependent Pact template skipped.
- All 15 real-PostgreSQL journal integration cases passed. Five actual outbox payloads validate against the merged schemas; removing required event time or aggregate identity is rejected.
- Event compatibility, AsyncAPI discriminator, duplicate-YAML and PR security-checklist gates pass.
- Hosted [CI](https://github.com/JiRaska/open-bank-oss/actions/runs/35063523633) and [Services CI](https://github.com/JiRaska/open-bank-oss/actions/runs/35063523787) both pass. The immutable envelope identifies tested merge commit `4eba9709e271450ac81c5ccb1251676df851341a`; GitHub confirms its parents are main `b46189fe7` and this PR head `b74d431a5`. Hosted JUnit evidence confirms all 15 journal integration tests pass. The broker-backed contract job is skipped on this PR; required independent money-path reviews remain outstanding.

## Earlier validation

- Full sanctions module build, static analysis and coverage verification passed: 139 executed tests, no failures/errors; one Pact Broker-dependent verification skipped.
- After the final aggregate-identity and shared-ID-generator changes: all 15 real-PostgreSQL journal/publication integration tests passed, with detekt and ktlintCheck.
- Five actual outbox payloads validated against their event schemas; missing required event time or aggregate identity is rejected.
- Failure and concurrency coverage includes transaction rollback, identical retry, late commits with lower sequence IDs, source-key transitions, missing/oversized identifiers, bounded chunks and shared outbox sequence allocation/claiming.
- Withheld-publication stress: 776,000 pending changes, two publication attempts in 7.124 seconds with the test JVM limited to 512 MiB. This measures the withheld path, not full-fleet normal publication throughput. PostgreSQL temporary storage still scales with journal size.
- All 229 local CI gates passed, combining the complete run with targeted reruns after the shared-ID fix and local Python/oasdiff environment corrections. Hosted CI passed on d1ef6b689: [Services CI](https://github.com/JiRaska/open-bank-oss/actions/runs/34784832198) performed the sanctions build and coverage verification; [CI](https://github.com/JiRaska/open-bank-oss/actions/runs/34784831977) passed all gate shards. The Pact Broker contract job is intentionally post-merge-only and was skipped on this PR. Required independent money-path reviews remain outstanding.

## Security checklist

- [x] No secrets, tokens, passwords or customer PII added to code, configuration, tests or logs; emitted source identifiers come from public sanctions feeds.
- [x] No new unjustified suppression or unsafe cast.
- [x] No new third-party production or test dependency.
- [x] Auth, crypto and payment authorization code are unchanged; the sanctions threat model covers the new publication and withholding behavior.
- [x] Customer PII paths are unchanged.
- [x] Cardholder data paths are unchanged.

## Rollout / Rollback

Apply V14 before the publisher runs. The trigger also journals writes from older application instances during rolling deployment. Events are delivered at least once; consumers must deduplicate and re-read current sanctions state rather than assume ordered feed snapshots.

To revert the application, leave the journal in place so pending evidence survives. A database rollback requires stopping publication and preserving pending journal data before removing the trigger, functions and tables; the migration contains the rollback note. Do not discard withheld evidence as a cleanup action.


JiRaska:
## Reviewed against the merged ADR — D1's second half was missing, now added

ADR-0256 landed in #4460 while this sat in draft, so its acceptance criteria are now fixed text.
D1's firing condition has **two** halves. This PR implemented the first and not the second:

> if the diff exceeds a configured share of the list (upstream schema reformat looks like a total
> rewrite), the refresh raises an operator ticket **instead of a mass re-screening**; and each
> refresh raises at most one trigger per list

The second guard (one trigger per list) was already satisfied structurally — one event per refresh.
**The storm guard was not implemented at all.**

Without it an upstream reformat — which rewrites every row and is, entry for entry,
indistinguishable from "the whole list changed" — raises `SANCTIONS_LIST_CHANGED` and re-screens the
entire customer book off a formatting change. That is the exact failure D1 exists to prevent, and
the trigger's whole value is that it means something.

## What was added

Above `openbank.sanctions.list-change.storm-threshold-share` (default `0.5`), the trigger is
**withheld** and `SANCTIONS_LIST_CHANGE_STORM` is raised instead.

- **A different event type on purpose** — a consumer must never be able to mistake it for the
  trigger and re-screen on it.
- **Counts only, no id list** — here the ids are "most of the list", noise rather than evidence.
- **ERROR, not WARN** — a withheld trigger means a real change may go un-re-screened until someone
  looks.
- **A list's first import is exempt.** With no baseline the share is undefined, and treating that as
  a storm would mean a newly configured list could never raise its first trigger.

## Threat model: T5 records the residual, not just the mitigation

The guard trades one failure for another, deliberately. A genuinely large legitimate change — a
regime action touching more than half a list — is withheld and goes un-re-screened until an operator
acts on the storm event. **That makes the storm event operationally load-bearing:** if nobody
watches it, a real bulk sanctions action is silently deferred.

The accepted trade is that a false mass re-screening is unrecoverable while a withheld one is
visible and re-drivable. T5 states both sides.

## Tests — three, each falsified

| test | falsification |
|---|---|
| above threshold withholds the trigger | removing the guard reddens it |
| below threshold still raises the trigger | proves the fix is not "always withhold" |
| a first import is not a storm | removing the null-baseline exemption reddens it |

## Also: `main` merged in

The branch was 4 days stale, and `gates (registry)` failed on *"references ADR-0256, which has no
`docs/adr/0256-*.md` file"* — the ADR merged **after** this branch was cut, so the reference was
dangling from the branch's point of view. Merging `main` resolved it: registry is now **26/26**.

Verified the merge did not silently drop anything (a clean merge can delete a list entry without
conflicting): T5 row present, changelog entry present, storm code intact,
`.release-please-manifest.json` still 58 entries on both sides.

## Verification

`detekt`, `ktlintCheck`, `koverVerify`, `test`, `quarkusBuild` (`--rerun-tasks`) — **101 tests,
0 failures, 1 skipped** (the broker-gated pact verification, expected to skip locally).
Gate groups: security 12/12, registry 26/26.

## Still a draft, and one open question for you

I have **not** marked it ready — the PR's "Deviation from ADR-0256 D1" note offers to amend the ADR
so D1's *stored-hash-column* language reads *content-level diff*. That amendment has not happened,
so the merged ADR and this implementation still describe the mechanism differently. Worth settling
before review rather than after.

Money-path, so 2 approvals per ADR-0030 either way. Not self-merging.

JiRaska:
## Review — still a draft, 1040 commits behind, and the storm guard has no reader

Reviewer comment only (no approve, no merge). Read the full 3-dot diff (12 files, 752+/207-). This is the **oldest PR in the batch** (opened 2026-08-11, last pushed 2026-08-15) and is still marked **draft**, so the first question is whether it is alive.

### Staleness: not superseded, but nothing has moved in a week

- `git grep -n 'SANCTIONS_LIST_CHANGED\|SanctionsListChangeSet' origin/main -- openbank-sanctions-service` -> **zero hits**. Nobody has implemented this; it is not a duplicate.
- **ADR-0256 exists on `main`** and is `proposed/partial`, and its §"triggers" already lists `SANCTIONS_LIST_CHANGED` *(active)* — so the ADR documents an event the code does not emit. That is a real gap this PR closes, and it is worth saying that the ADR is currently making a claim the tree cannot honour.
- Both files this PR rewrites (`SanctionsImportService.kt`, `SanctionsListService.kt`) still exist on `main` and were not restructured under it; the only sanctions commit since is `74b609f52` (audit-attribution `sourceService`), which does not collide.
- 1040 commits behind, `mergeable: UNKNOWN`, no CI has run.

**So: a live gap, a stale branch, and no signal.** The first action is a decision, not a code change — un-draft and rebase, or close it and re-cut. Sitting as a draft it is invisible to the queue and the ADR keeps advertising a trigger nobody fires.

### Finding 1 — the storm event has no consumer, no alert, and no reader

`SanctionsListService.publishChangeEvent` withholds the re-screening trigger above the threshold and instead emits `SANCTIONS_LIST_CHANGE_STORM` plus a `Log.errorf`. The comment is honest about the stakes:

> ERROR, not WARN: a withheld trigger means a real list change may go un-re-screened until someone looks.

Nothing in this diff makes anyone look. There is no `PrometheusRule`, no gitops change, and no consumer for `SANCTIONS_LIST_CHANGE_STORM` anywhere in the repo. An ERROR log line in a service nobody tails is not a control — and this is the failure mode where a genuine mass designation (a large new EU regime package is exactly the shape that trips a 50% threshold on a small list) is **silently not re-screened**. That is a fail-open on sanctions screening.

The event type is also not in `.github/event-contract-baseline.txt` (which currently carries only `openbank.sanctions.screening.event` for this service). Two new event types are being put on the wire; the baseline should record them.

Minimum before this leaves draft: an alert on the storm event or on its absence, and a named owner for "verify the feed and re-screen deliberately". Preferably the alert fires on the **success state** — "a list refreshed and no trigger was raised" — since the storm path is the quiet one.

### Finding 2 — `persistStandalone` gives up the transactional-outbox property, and the KDoc's reason is thin

`SanctionsOutboxRepositoryImpl.kt` adds:

```kotlin
override suspend fun persistStandalone(message: OutboxMessage) {
    Panache.withTransaction { persist(message.toEntity()) }.awaitSuspending()
}
```

with the port KDoc arguing *"the aggregate write and event write do not need atomicity (the list refresh is idempotent and re-runnable)"*.

I do not think that holds as stated. In `refresh()` the ordering is: entries upserted (their own transactions) -> `publishChangeEvent` -> `markUpdated()`. So there are three independent commits, and a crash between the second and third leaves an event published for a change whose `lastEntryCount` baseline was never advanced — the next refresh then measures its diff against a stale baseline, which is the input to the storm guard. "Idempotent and re-runnable" is a claim about the *import*; it is not a claim about the *baseline*, and the storm guard is a function of the baseline.

Please either state explicitly which of those two writes may be lost and why that is acceptable, or bring the event into the same transaction as `markUpdated`.

### Finding 3 — the storm threshold's denominator and numerator do not match

`share = changeSet.changeCount / baselineEntryCount`, where `changeCount = changedExternalIds.size + deactivatedExternalIds.size`. Two consequences worth a comment:

- On the OFAC SDN path there is deliberately **no deactivation sweep** (correctly noted in `SanctionsImportService`), so `deactivatedExternalIds` is always empty there and the two lists are not measuring comparable things across list types.
- A refresh that changes N entries *and* deactivates N others yields `2N/baseline` — the share can exceed 1.0 legitimately. Not a bug at a 0.5 threshold, but the metric is not a "share of the list" as the KDoc says it is.

### Smaller notes

- `SanctionsListService` now takes a 5-parameter primary constructor plus a secondary; watch `LongParameterList`, which fires **at** `constructorThreshold: 9`, not above it. Currently fine.
- `stormPayload` placed as a top-level function **after** the class, with a comment explaining it cannot then steal a following declaration's annotation — that is the `McpEndpoint`/`@Path` lesson applied correctly, and I want it on record as deliberate rather than accidental.
- `openbank.outbox.dispatch-enabled` is already `true` on `main` (`application.yaml:143`), so the events will actually dispatch.
- `OpenSanctionsCsvColumns.parseCsvLine` is a hand-rolled RFC 4180 parser handling quotes and doubled quotes. It is tested, but this is unauthenticated third-party feed content reaching a hand-written parser — a fuzz case or a property test over it would be cheap insurance, and this repo has found decoder defects that way before.
- `upsertAllReturningChanged` deriving the set from the same `RETURNING` clause that backs the count is the right design — it cannot disagree with what was written.

### Verdict

**Needs author decision.** The design is thoughtful and the ADR-0256 gap is real. But it is a draft nobody has touched in a week, 1040 commits behind, with no CI, and its central safety valve (the storm guard) currently routes to an event with no reader. Un-draft + rebase + wire an alert, or close it and say so in ADR-0256 so the ADR stops claiming the trigger is active.

JiRaska:
Triage note (not closing — this one needs a decision, not a rebase).

Rebased onto current `main` and stopped: `SanctionsImportService.importList` has been independently redesigned on both sides, so the two test files collide in 18 hunks that are not mechanical.

- `main` returns an outcome-reporting result: `.outcome` (`IMPORTED` / `SKIPPED_NOT_ENTITY_BASED` / `FAILED_KEPT_EXISTING`), `.entriesImported`, `.detail`.
- This branch returns a change set: `.isEmpty`, `.changeCount`, `.changedExternalIds` — the input to the `SANCTIONS_LIST_CHANGED` event this PR exists to publish.

Taking either side wholesale reverts the other. A correct resolution has to decide what `importList` returns now — most likely both facets (why the import ended, and what changed), which is an API decision on a money-path-adjacent service rather than something to settle inside a merge.

`version.txt` is a separate, easy one: the branch carries 0.9.0 against `main`'s 0.12.0. That file is release-please's (ADR-0029 rule 2), so the branch value is simply stale and `main`'s is correct.

Leaving open and untouched — the branch is not stomped, nothing was force-pushed.

JiRaska:
## This is not a merge conflict, it is a port — stopping rather than guessing

Attempted `git merge origin/main` on this branch and aborted it. 23 conflict hunks across four
files, and the reason they conflict is a **changed port contract**, not adjacent edits:

| file | hunks |
|---|---|
| `SanctionsImportService.kt` | 2 |
| `SanctionsListService.kt` | 3 |
| `SanctionsImportServiceTest.kt` | 8 |
| `SanctionsListServiceTest.kt` | 10 |

### What changed underneath the branch

`main` now has `openbank-sanctions-service/.../port/out/SanctionsListImport.kt`:

```kotlin
enum class ListImportOutcome { ... }
data class ListImportResult(val outcome: ListImportOutcome, val entriesImported: Int, val detail: String? = null)
```

and `SanctionsListService` keys on it explicitly:

> Key on the outcome, never on "count > 0" (issue #8362 / #4348): only IMPORTED means the stored
> list now reflects the upstream source; every other outcome left the stored entries untouched, so
> the stored count stays the honest number.

That is the "a successful no-op and a real success must not share a value" family, and it was
landed deliberately.

This branch, written before that, calls the same method and expects a **content diff** back:

```kotlin
val changeSet = importer.importList(enumType, list.sourceUrl)
publishChangeEvent(list.id, changeSet, baselineEntryCount = list.lastEntryCount)
if (changeSet.changeCount > 0) { changeSet.changeCount ... }
```

`SanctionsListChangeSet` **does not exist on `main`** — it is a type this branch introduces. So the
two sides do not disagree about a line; they disagree about what `importList` returns.

### Why I did not resolve it

Taking either side compiles into something wrong:

- main's side drops the ADR-0256 D1 change-set, the storm guard and `SANCTIONS_LIST_CHANGED`
  entirely — i.e. the whole point of this PR;
- this branch's side reverts #8362/#4348 and restores the `count > 0` test that treated a skipped
  import as a successful one.

The correct shape is almost certainly for the importer to return **both** — the outcome *and* the
content diff — which is a change to the port, its implementations and both test suites. That is an
authoring decision on the sanctions screening path, where a wrong resolution means customers are
silently not re-screened: the exact defect ADR-0256 D1 exists to prevent, and one that no test here
would fail on.

The branch's last commit is **2026-08-15**, so it predates that redesign by about four weeks.

### Suggested next step

Rebuild this change on top of main's contract rather than merging: widen `ListImportResult` with the
changed/deactivated external ids (or return a separate diff alongside it), keep the
`outcome == IMPORTED` gate, and re-apply `publishChangeEvent` plus the storm guard against the new
shape. The tests will need writing against the new return type, not merging.

JiRaska:
Queue safety (2026-09-17): sanctions-service is listed in `rules.yaml: money_path_services`. This PR currently reports CLEAN and its checks have no failures, but it has no approving reviews. Its threat-model change is present; the remaining gate is two independent money-path approvals on the final diff. Added existing `money-path`, `security-review-required` and `blocked` labels so green CI is not mistaken for authorization to merge. No unchanged-input CI rerun is needed.

JiRaska:
Focused review of current head `b74d431a5` clarifies the automated findings above:

- The OFAC import path does **not** call `deactivateMissing`; an empty parsed OFAC feed returns zero to `importedOrEmpty`, so the proposed whole-list deactivation is not reachable through that path. EU FSF and CSV do reconcile missing rows and therefore need their explicit empty-feed guards.
- V14 seeds `sanctions_change_publication` from all seven `sanctions_lists` rows. V3 seeds exactly the seven `SanctionsListType` values, and the current repository/API offers update and refresh but no list-creation operation. A future migration that adds a list must seed its publication row (or make publisher creation atomic); this is a follow-up invariant, not a current runtime failure for the supported lists. The same applies to the hypothetical unknown enabled `listType` in scheduled recovery.

These clarifications do not constitute an approval. This money-path PR still requires two independent approvals on its final diff.

JiRaska:
Follow-up e03b1269d: the scheduled tick now drains committed journal evidence for a disabled list without importing that feed. A failed publication followed by disabling the list previously left those changes indefinitely pending. Added a regression assertion for the disabled/flagged path. Local `SanctionsListServiceTest`, `ktlintCheck`, `detekt`, and `git diff --check` passed; signed commit verified. Awaiting exact-head CI and independent money-path review.

JiRaska:
Blocking money-path review finding: repairing an invalid source identity does not release the journal.

Reproduction from this diff: insert a `PEP_GLOBAL` entry with `external_id = NULL`, then update that same entry to a valid external ID. `sanctions_record_change()` writes the old NULL key as an inactive journal row and the valid key as an active row. `CREATE_CHANGES` partitions by `(entry_id, external_id)` and `CREATE_TARGETS` retains both keys; `SUMMARY.missingIdentity` therefore remains true. `publishPending()` returns `WITHHELD` without deleting the selection. Every retry sees the same old NULL key, even though the current entry is repaired. Replacing an oversized key with a bounded one has the same persistent `SOURCE_ID_TOO_LARGE` latch. The existing tests cover initial withholding but not a subsequent repair and drain.

Please add a real-Postgres regression for invalid identity → correction → publication, then define how an unpublishable historical key is resolved without silently erasing evidence. Until that path exists, an operator cannot recover a sanctions list from an otherwise correctable feed row without manual journal surgery.

JiRaska:
The repair deadlock is now reproduced by a real-Postgres `@QuarkusTest` in an isolated checkout of head `e03b1269`. The added test inserts a PEP entry without `external_id`, observes the initial `WITHHELD`, updates that same entry to `external_id = 'repaired-source'`, and expects `PUBLISHED`, an empty journal and one changed-ID event. Actual result: **expected `PUBLISHED`, got `WITHHELD`** at `SanctionsChangeJournalIT.kt:384` (1 test, 1 assertion failure, 0 skips). The first local attempt was inconclusive because another test process held port 8081; the rerun used `QUARKUS_HTTP_TEST_PORT=0 QUARKUS_MANAGEMENT_TEST_PORT=0` and reached the assertion. Separate `:openbank-sanctions-service:ktlintCheck` and `:openbank-sanctions-service:detekt` passed, as did `git diff --check` on the 12-line regression test.

I have not pushed the failing test into this owner branch or modified the production PR. The minimal test change is in the isolated review checkout; I can hand over its patch with the production fix once the resolution semantics are agreed. This confirms the existing blocking finding is runtime behavior, not just a SQL inference.

JiRaska:
The retained-publication blocker reported above is fixed in e56fd30f6. When a source identity is repaired, the publisher marks only the invalid historical journal rows as resolved in the same transaction as outbox publication. The audit rows remain stored; the corrected identity is emitted once, and a failed outbox write rolls the resolution back for retry.

Local verification on this exact commit: full `SanctionsChangeJournalIT` passed, including the new repair and rollback/retry cases; `:openbank-sanctions-service:ktlintCheck` and `:openbank-sanctions-service:detekt` passed; `git diff --check` passed. The new hosted checks are still running. Keep this money-path PR in draft until CI and the required independent reviews complete.

JiRaska:
Follow-up 99cecbf17 closes the same repair gap when the entry moves to another sanctions list while its source key is corrected. The old invalid journal rows retain both the resolution timestamp and target list/key; the valid change remains pending in the destination list until its own publication succeeds. Full SanctionsChangeJournalIT suite, ktlint, and detekt passed locally on this head. New hosted CI is running.

JiRaska:
Review handoff: head 99cecbf17 is ready for the required independent money-path reviews. Hosted service build/test and 20 other checks passed. CodeQL is red because the GitHub installation API quota was exhausted during SARIF post-processing (run 35317241892, job 105511748374, 2026-09-18 07:06 UTC); this is a missing scan result, not a clean security verdict or a reported code finding. Do not merge until CodeQL completes successfully after quota recovery.

JiRaska:
Review control clarification: this money-path change is governed by the project rule requiring two independent approvals and a threat model, but #2183 documents that the current GitHub main-protection ruleset does not enforce the two-approval count. This PR has zero reviews and no auto-merge request. Please treat the two approvals as a manual merge hold; the red CodeQL installation-quota failure is a separate required security-evidence gap. Do not merge based only on a future green `all-green` check.

JiRaska:
Final-head CI update: CodeQL on `a66269685` failed again at 2026-09-18 07:23:22 UTC with `API rate limit exceeded for installation` (run 35318677309/job 105516024381). This is a second missing CodeQL scan, not a finding and not a passing security verdict. Hosted service build/tests passed. Keep merge on hold for a successful CodeQL result and the documented money-path review requirements; review of the code itself can proceed.

JiRaska:
Follow-up on final head `0e21bb289`: a committed entry with an unpublishable source ID could be deleted without ever acquiring a valid ID, leaving its journal permanently `WITHHELD`. The new PostgreSQL regression test reproduced that outcome before the fix. The publisher now resolves only invalid rows whose entry no longer exists, in its publication transaction, retaining both audit rows with `SOURCE_REMOVED` and emitting no false change event. All 19 `SanctionsChangeJournalIT` cases passed, as did ktlint and detekt. Hosted checks restarted; do not merge until CodeQL/dependency-review succeed on this head and the documented independent reviews are recorded.

JiRaska:
Final-head CI handoff (0e21bb289): all 22 reported checks are complete and successful, including hosted sanctions build/test, CodeQL, dependency-review and all-green. The earlier installation-quota failures in this thread were on superseded heads and no longer block this exact commit. The PR has zero reviews; #2183 confirms the money-path two-approval rule is documented but not technically enforced by main-protection. Keep merge on manual hold until the required independent reviews and threat-model assessment are recorded. No auto-merge is requested.

Issue #4481, opened 2026-08-13
feat(product-catalog): add insurance interoperability adapters


JiRaska:
## Goal

Complete the interoperability tail of ADR-0257 after the industry-neutral catalog kernel and standalone distribution land.

## Scope

- expand the trusted insurance reference pack with coverage, peril, exclusion, limit, deductible, premium, and underwriting-question attributes without adding insurance symbols to the kernel
- define a deliberately bounded, version-pinned ACORD/TMF adapter profile; do not attempt a universal ontology
- add adapter conformance fixtures and round-trip tests that prove no silent data loss
- document supported/unsupported mappings and keep the pack opt-in for standalone deployments
- update ADR-0257 delivery status only when the conformance gate is green

## Acceptance

- generic kernel architecture test remains industry-neutral
- exact decimal and effective-date semantics survive round trips
- unsupported external fields fail explicitly rather than being dropped
- insurance pack and adapter fixtures validate in CI
- standalone empty and banking-only profiles remain unchanged

Refs ADR-0257.

PR #4501, opened 2026-08-13
feat(product-catalog): ship standalone platform and intelligence studio

### LINKED ISSUE #4499: feat(product-catalog): ship standalone platform and product intelligence studio
Deliver the governed, industry-neutral product catalog as a deployable vertical slice.

Scope:
- standalone catalog kernel, trusted industry packs and bank compatibility
- four-eyes publication, audit/outbox and cursor delivery
- Product Intelligence Studio with draft/live insight and governed AI review
- insurance interoperability reference pack

Acceptance:
- PR and deployment gates cover catalog, agent service and Admin UI
- GitOps delivery deploys the merged artifacts through the standard canary path

Refs #4481

JiRaska:
Closes #4499

## What changed
- introduces the governed industry-neutral product catalog, trusted banking and insurance packs, standalone distribution and compatible v1 banking projection
- adds immutable revisions, contextual offerings, four-eyes publication, audit/outbox evidence and cursor delivery
- adds the Product Intelligence Studio with clear authoring flow, draft/live structural insight and a server-side governed AI review proposal
- adds the insurance term-life interoperability reference adapter; refs #4481

## Why
Product definition now works beyond a bank while retaining backward compatibility for current banking consumers. Operators get a human-controlled, explainable assistant rather than autonomous product changes.

## Validation
- Product catalog: ktlint, detekt, 95 JUnit/Pact tests (0 failures; 1 expected broker skip), Quarkus fast-jar build
- Agent service: ktlint, detekt, tests and Quarkus build
- Shared runtime test suite
- Admin UI: targeted ESLint, TypeScript, AI review BFF tests and production build

CI remains the authoritative clean-run E2E and deployment gate.


JiRaska:
Diagnosis from the autonomous queue — not touching the branch, it is actively being worked (651 files touched in the last 5 minutes).

`build (openbank-product-catalog)` has failed **twice**, both times on the same test, so this is not the timing flake it resembles:

```
CatalogReconciliationSchedulerVertxContextIT > scheduled reconciliation repairs a legacy write after startup() FAILED
  Caused by: org.hibernate.exception.ConstraintViolationException
ERROR: duplicate key value violates unique constraint "catalog_specifications_pkey" (23505)
```

That is the repo's documented reactive-Panache trap, not test data collision:

> Panache reactive `persist()` on an application-assigned `@Id` is INSERT-only — use `merge` for updates. An aggregate whose `@Id` is set by the app (not `@GeneratedValue`) fails *every* lifecycle transition with `duplicate key value violates … _pkey` at flush, because Hibernate schedules an INSERT and never an UPDATE: a non-null assigned id cannot distinguish transient from detached.

So if `catalog_specifications` has an application-assigned `@Id` and the reconciliation path calls `persist`/`persistAndFlush` on a row that already exists, this is the same defect consent-service shipped — where every revoke/reject/activate 500'd and only a real-DB IT caught it. The fix shape is `Panache.getSession().flatMap { it.merge(entity) }`; `SddMandateRepositoryImpl` documents it in place.

Two things that make this worth reading rather than re-running:
- **It is invisible to any test that mocks the repository.** This IT catches it precisely because it drives the real scheduler against a real DB — which is why the failure is a signal about production behaviour, not about the test.
- The `ConflictException: Status 409: a prune operation is already running` in the same log is Ryuk cleanup on `Thread-77` and is **not** the cause. I checked the last 40 runs and it appears in no other failure, so it is noise here.

If the reconciliation genuinely means "insert if absent, update if present", `merge` is the one-line answer. If it means "insert only", then something upstream is re-delivering an already-reconciled row and the constraint is telling the truth.

PR #4520, opened 2026-08-13
fix(analytics-sink): materialise the ADR-0210 D2 party key as a ClickHouse view

### LINKED ISSUE #4511: ADR-0210 D2 drift: account→party resolution is an inline CTE in a BFF route, not a ClickHouse view
## What ADR-0210 D2 decided

> That mapping is derived from the same event stream and **materialises as a ClickHouse view
> alongside the existing silver views**. It does not become a service, and it is the one thing this
> ADR adds to the schema.

D2 calls account→party resolution "the real work" and "the only new persistence" of the whole ADR.

## What is actually there

No such view exists. `openbank-analytics-sink/src/main/resources/clickhouse/` carries V1–V4 and
none of them defines a party→account mapping; neither does
`openbank-infra/gitops/components/analytics/clickhouse-init-configmap.yaml`.

The resolution lives instead as an inline CTE inside a single caller —
`openbank-admin-ui/src/app/api/customer-360/[partyId]/route.ts`, `scopedRowsSql()`:

```sql
WITH party_accounts AS (
  SELECT DISTINCT aggregate_id FROM openbank_analytics.bronze_events
  WHERE upper(aggregate_type) = 'ACCOUNT'
    AND JSONExtractString(payload, 'partyId') = '<uuid>'
)
```

So the one piece of logic the ADR says it adds to the schema was added to a BFF route instead.

## Why this matters beyond tidiness

1. **It is the isolation boundary, and it is per-caller.** The route's own comment says it: "ISOLATION
   IS THE LOAD-BEARING PROPERTY … If the indirect arm were ever widened, this route would show
   another customer's transactions." A definition that lives in one caller is a definition the
   second caller re-derives. `SilverSegmentEvaluator` in `openbank-campaign-service` already reads
   the same silver layer; the moment anything else needs account→party, there are two answers to
   "which accounts belong to this party" and only one of them is tested.
2. **The identical rejection is already written down in the ADR.** "Query `bronze_events` directly
   instead of the silver views. Rejected: every caller would re-derive the `argMax` reduction, and
   the first one to write it slightly differently produces a different current state. The silver
   views exist so that reduction has one definition." That reasoning applies verbatim to the party
   key.
3. **The casing bug is evidence, not hypothesis.** `aggregate_type` is uppercase in bronze; the first
   version of this CTE compared against lowercase literals and matched nothing. A filter that finds
   no rows is indistinguishable from a party with no accounts — nothing failed, the page just showed
   less. That class of defect is exactly what a single shared definition prevents from recurring per
   caller.
4. **Empty `partyId` is currently unguarded at the SQL layer.** Rows whose payload has no `partyId`
   extract to `''`. The route can never send `''` (its UUID regex is the injection boundary), so
   this is latent rather than live — but a shared view must not offer the footgun to the next
   caller.

## Proposed fix

- `V5__party_accounts.sql`: `CREATE OR REPLACE VIEW openbank_analytics.silver_party_accounts`
  exposing `(party_id, account_id)`, with the `upper()` fold and a `party_id != ''` guard.
  `CREATE OR REPLACE`, not an edit to V1 — V1's statements are all `CREATE … IF NOT EXISTS`, so
  editing it changes nothing on a database where the objects already exist (the trap V4 documents).
- Same DDL into `clickhouse-init-configmap.yaml` as `04-party-accounts.sql`, so a fresh cluster gets
  it at init the way V2/V3 are carried today.
- `scopedRowsSql()` selects from the view instead of restating the CTE.
- A test that would have caught the drift: assert the generated SQL references the view and does
  **not** re-derive the account filter, alongside the existing isolation assertions in
  `customer-360.test.ts`.

**Rollout note (same as every previous ClickHouse migration here):** the init ConfigMap applies only
on first boot, so an existing warehouse needs the DDL applied by an operator before the admin-ui
image that depends on the view is deployed. That ordering is the one risky part of this change.

## Out of scope

`card→party` and `case→party` — D2 names both and neither is implemented. Tracked separately; this
issue is only about moving the implemented `account→party` arm to where D2 says it belongs.

Refs ADR-0210 D2.

JiRaska:
## Summary

ADR-0210 D2's account→party view never landed — the resolution shipped as an inline CTE inside one
caller, admin-ui's Customer 360 route. This adds `silver_party_accounts` where D2 says it belongs and
points the route at it. Found on the way: `V4__screen_feedback_context.sql` never reached the
`clickhouse-init` ConfigMap, so a warehouse built from scratch would silently lack it — fixed here
too, with a gate that makes the class of drift impossible to repeat quietly.

## Type of change

- [x] fix (bug fix)

## Linked issues

Closes #4511

## Changes

- **`V5__party_accounts.sql`** — `CREATE OR REPLACE VIEW openbank_analytics.silver_party_accounts`
  as `(party_id, account_id)` over `bronze_events`. `CREATE OR REPLACE`, not an edit to V1: V1's
  statements are all `CREATE … IF NOT EXISTS`, which is a no-op on a warehouse that already has the
  objects — it would look applied in git and do nothing in ClickHouse (the trap V4 documents).
  It carries the two guards the CTE had learned:
  - `upper(aggregate_type) = 'ACCOUNT'` — bronze stores the type uppercase, and the route's first
    version compared against lowercase literals, matching nothing. A filter that finds no rows is
    indistinguishable from a party that owns no accounts, so nothing failed; the page showed less.
  - `JSONExtractString(payload, 'partyId') != ''` — a missing key extracts to `''`, so without this
    every partyId-less ACCOUNT event becomes a row owned by the empty party. The route can't reach
    that case (its UUID regex is the injection boundary), so this is latent, not live — but a shared
    view must not hand the footgun to the next caller.
- **`route.ts`** — `scopedRowsSql()` selects `account_id FROM silver_party_accounts WHERE party_id =
  '<uuid>'` instead of restating the CTE. Same two arms, same rows, no re-derived ownership rule.
- **`clickhouse-init-configmap.yaml`** — adds `05-party-accounts.sql` **and** the missing
  `04-screen-feedback-context.sql`.
- **`check-clickhouse-ddl-in-configmap.py`** + gate `clickhouse-ddl-configmap-parity` (enforced,
  group `data`).
- **`customer-360.test.ts`** — isolation assertions rewritten for the view; two new tests.

## Why the view rather than leaving a working CTE

The route was correct, tested and green — nothing was broken today. What was wrong is *where* the
definition lived. This resolution is the isolation boundary of the Customer 360 (the route's own
comment: a widened indirect arm shows another customer's transactions), and a definition inside one
caller is a definition the second caller re-derives. ADR-0210 rejects querying `bronze_events`
directly instead of the silver views for exactly that reason — "the first one to write it slightly
differently produces a different current state" — and the casing bug above is that failure mode
already having happened once, per caller.

## Verification — each check run against a case it MUST flag

- **Gate falsified against the pre-fix tree.** Restoring `origin/main`'s ConfigMap makes it exit 1
  flagging `gold_screen_feedback_context` (the real 2026-07 defect) and `silver_party_accounts`.
  Green on the fixed tree; `--self-test` covers both directions plus an absent-subject case, so it
  cannot report clean because a path moved.
- **New tests falsified against the old route:** 2 failed / 7 passed on `origin/main`'s `route.ts`.
- **Isolation assertion falsified by a simulated leak:** deleting the `WHERE party_id` from the
  subquery goes red. It asserts over *every* reference to the view, not the first, so adding a second
  unscoped join is also red.
- `npm test` → **1183 passed / 90 files**, `npm run type-check` clean, eslint clean on both changed
  files, `ruff` clean on the new script, `run-gates.py --only clickhouse-ddl-configmap-parity` PASS,
  and the two gate meta-checks (`gate-selftest-declaration`, `gate-invocation-reachability`) PASS.

## Definition of Done

- [x] Hexagonal layering preserved (no service code touched; DDL resource + BFF SQL only).
- [ ] Integration tests — N/A, no service boundary changed. `ClickHouseAnalyticsSinkIT` mounts V1
      only and asserts the bronze insert path; it does not read this view.
- [ ] Contract tests — N/A, no public API changed.
- [ ] `./gradlew …` — N/A, no Kotlin changed (a `.sql` resource only).
- [x] No new lint warnings.
- [ ] OpenAPI — N/A. Flyway — N/A (ADR-0022: this service runs no migration runner).
- [x] Docs: the DDL file states its own reasoning; ADR-0210 needs no amendment — this implements what
      D2 already decided.

## Security checklist

- [x] No secrets, tokens, passwords or PII. The view exposes `party_id` + `account_id` only; both
      already live in bronze, and payloads are PII-masked at the sink.
- [x] No new `@Suppress`/`as Any`. No new dependency.
- [x] **PII path touched** → the view selects `partyId` out of bronze payloads. No new store, no new
      copy: it is a view, so GDPR Art. 17 erasure stays analytics-sink's crypto-shredding
      (ADR-0210 D5), unchanged and uninherited-from-anywhere-else.

## Compliance impact

- [x] GDPR — Art. 17 erasure inherited, not reimplemented (D5); a view holds no data of its own.
      Art. 5(1)(c) unaffected: the view narrows to two identifiers rather than widening exposure.
- [x] None otherwise. Not a money path; the Customer 360 is operator-facing and read-only (D4).

## Rollout / Rollback

**This is the one ordering-sensitive part.** The init ConfigMap applies on FIRST BOOT ONLY, so the
existing sandbox warehouse does not get the view from this merge. The admin-ui image deploys on merge
and immediately queries it, so:

1. Apply the DDL to the running warehouse **before** the admin-ui deploy lands:
   `clickhouse-client --query "$(cat openbank-analytics-sink/src/main/resources/clickhouse/V5__party_accounts.sql)"`
   (or POST it to the HTTP interface). Additive and idempotent — it creates one view and touches no
   table.
2. Fresh clusters need no step: the ConfigMap now carries it.

- Deployment strategy: rolling (admin-ui), operator-applied DDL for existing warehouses.
- Rollback: `DROP VIEW openbank_analytics.silver_party_accounts` + revert the route. The view owns no
  data, so nothing is lost either way.
- Data migration reversible: yes (view only, no table, no backfill).

If the view is missing when the route runs, the failure is loud in the right place: the route catches
it and the page renders `available: false` — a calm unavailable state, not a 5xx and not an empty
Customer 360 that reads as "this customer has nothing".

## Reviewer notes

- The V4 gap is a separate defect from the D2 drift; it is fixed here rather than split off because
  the gate that prevents both cannot be added while one of them is still red.
- `card→party` and `case→party` are named by D2 and remain unimplemented — deliberately out of scope,
  called out in #4511.
- `SilverSegmentEvaluator` (campaign-service) reads the same silver layer and is the most likely next
  caller to need this view; it does not resolve accounts today, so it is untouched.

### DIFF
--- a/.github/gates/gates.yaml
+++ b/.github/gates/gates.yaml
@@ -2044,6 +2044,22 @@ gates:
     run: |
       bash .github/scripts/libs-change-dependents.sh --selftest
 
+  # analytics-sink owns no OLTP DB and runs no migration runner (ADR-0022): its ClickHouse DDL lives
+  # once as `clickhouse/V*.sql` resources an operator applies, and once in the `clickhouse-init`
+  # ConfigMap a FRESH cluster boots from. Nothing reconciles the two, and a migration that reaches
+  # only the resources is invisible — file in git, object present in the sandbox because someone
+  # applied it, CI green — until a warehouse is rebuilt without it. V4__screen_feedback_context.sql
+  # sat that way from 2026-07 until #4511. Compares created OBJECTS, not text, in both directions.
+  - id: clickhouse-ddl-configmap-parity
+    name: "ClickHouse DDL reaches a fresh cluster (issue #4511, enforced)"
+    group: data
+    mode: enforced
+    selftest: |
+      python3 .github/scripts/check-clickhouse-ddl-in-configmap.py --self-test
+    selftest_expect: pass
+    run: |
+      python3 .github/scripts/check-clickhouse-ddl-in-configmap.py --root .
+
   - id: kafka-acl-coverage
     name: "Kafka ACL coverage (issue #2598, enforced)"
     group: data
--- a/.github/gates/gates.yaml
+++ b/.github/gates/gates.yaml
@@ -2051,6 +2051,7 @@ gates:
   # applied it, CI green — until a warehouse is rebuilt without it. V4__screen_feedback_context.sql
   # sat that way from 2026-07 until #4511. Compares created OBJECTS, not text, in both directions.
   - id: clickhouse-ddl-configmap-parity
+    min_subjects: 15   # 22 objects today (V1-V5 + the ConfigMap's copies, 2026-08-13)
     name: "ClickHouse DDL reaches a fresh cluster (issue #4511, enforced)"
     group: data
     mode: enforced
--- a/.github/scripts/check-clickhouse-ddl-in-configmap.py
+++ b/.github/scripts/check-clickhouse-ddl-in-configmap.py
@@ -36,6 +36,9 @@ import sys
 import tempfile
 from pathlib import Path
 
+sys.path.insert(0, str(Path(__file__).resolve().parent))
+import gatelib  # import after the path insert — checkers run as scripts from the repo root
+
 SQL_GLOB = "openbank-analytics-sink/src/main/resources/clickhouse/V*.sql"
 CONFIGMAP = "openbank-infra/gitops/components/analytics/clickhouse-init-configmap.yaml"
 
@@ -54,7 +57,9 @@ def objects(text: str) -> set[str]:
     return {m.lower() for m in CREATE_RE.findall(live)}
 
 
-def check(root: Path) -> list[str]:
+def check(root: Path) -> tuple[list[str], int]:
+    """Returns (errors, objects examined). The count is the gate's subject floor: a broken glob or a
+    regex that stops matching leaves the verdict clean, and only the count can say so."""
     errors: list[str] = []
     sql_files = sorted(root.glob(SQL_GLOB))
     cm_path = root / CONFIGMAP
@@ -62,13 +67,13 @@ def check(root: Path) -> list[str]:
     # A probe that finds no subject must say so, not pass. Both inputs are required to exist: this
     # gate reporting OK because a path moved is the failure mode it is meant to prevent elsewhere.
     if not sql_files:
-        return [f"no ClickHouse DDL found at {SQL_GLOB} — the gate cannot have checked anything"]
+        return ([f"no ClickHouse DDL found at {SQL_GLOB} — the gate cannot have checked anything"], 0)
     if not cm_path.is_file():
-        return [f"{CONFIGMAP} not found — the gate cannot have checked anything"]
+        return ([f"{CONFIGMAP} not found — the gate cannot have checked anything"], 0)
 
     cm_objects = objects(cm_path.read_text())
     if not cm_objects:
-        return [f"{CONFIGMAP} declares no openbank_analytics objects — parse or path drift"]
+        return ([f"{CONFIGMAP} declares no openbank_analytics objects — parse or path drift"], 0)
 
     seen: dict[str, str] = {}
     for f in sql_files:
@@ -86,7 +91,7 @@ def check(root: Path) -> list[str]:
             f"resource creates — DDL with no source of record"
         )
 
-    return errors
+    return (errors, len(set(seen) | cm_objects))
 
 
 SELF_TEST_CM = """apiVersion: v1
@@ -112,7 +117,7 @@ def self_test() -> int:
         )
 
         # known-negative: the sets agree.
-        errs = check(root)
+        errs, _ = check(root)
         if errs:
             print(f"SELF-TEST FAIL: clean tree flagged: {errs}")
             ok = False
@@ -122,7 +127,7 @@ def self_test() -> int:
             "-- CREATE VIEW openbank_analytics.commented_out AS SELECT 1;\n"
             "CREATE OR REPLACE VIEW openbank_analytics.silver_party_accounts AS SELECT 1;\n"
         )
-        errs = check(root)
+        errs, _ = check(root)
         if not any("silver_party_accounts" in e for e in errs):
             print("SELF-TEST FAIL: a resource object missing from the ConfigMap was not flagged")
             ok = False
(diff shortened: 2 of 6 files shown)


JiRaska:
**Held for human review** — the local merge guard blocks it: this PR touches `.github/gates/gates.yaml`
and adds `.github/scripts/check-clickhouse-ddl-in-configmap.py`, i.e. CI gate machinery, which needs a
human regardless of the title. That is the correct final state, not something to route around. Not
splitting the gate out into a second PR either — the split would exist only to get the rest merged
without the review the guard asks for.

What a reviewer should actually check, in the order the risk sits:

1. **The gate's scope, not its verdict.** `check-clickhouse-ddl-in-configmap.py` compares *created
   object sets* between `clickhouse/V*.sql` and the init ConfigMap, in both directions. It is a glob
   plus a regex, so the interesting question is what it CANNOT see: it does not compare the two
   definitions of an object, only that both create it. `min_subjects: 15` (22 objects today) is what
   stops a broken glob from reading as clean — `gate-subject-floor` caught the first version of this
   PR having no floor at all.
2. **That the gate flags a real historical defect, not a synthetic one.** Against `origin/main`'s
   ConfigMap it exits 1 on `gold_screen_feedback_context` — V4 shipped 2026-07 and never reached the
   ConfigMap, so a warehouse rebuilt from scratch would lack it. The sandbox has the object because
   someone applied it by hand, which is exactly why nothing noticed.
3. **The isolation assertion in `customer-360.test.ts`.** It asserts over *every* reference to
   `silver_party_accounts`, not the first. Deleting the `WHERE party_id` from the subquery goes red —
   verified, not assumed.

Already done outside CI: the view is applied on the sandbox warehouse, and the old CTE and the new
query return identical rows for a real party (13 vs 13, 0 rows on either side of the difference), with
no account mapping to two parties. So a merge does not need a coordinated DDL step for the sandbox —
only a fresh cluster depends on the ConfigMap change.

One merge-order note: `.github/gates/gates.yaml` is also edited by #4513 and #3930. Two neighbouring
entries appended to one list merge cleanly and keep only one, so whoever lands second should re-read
the file on `origin/main` afterwards and confirm their entry survived.

PR #4530, opened 2026-08-13
fix(ci): provision Temporal and resolve config expressions in the authenticated-fuzz harness

JiRaska:
# The five boot failures, grouped by root cause

Re-derived from run [30804842325](https://github.com/JiRaska/open-bank-oss/actions/runs/30804842325)
(the most recent run of `api-fuzz-authenticated.yml`, attempt-scoped jobs endpoint). The matrix is
now **18 services, 12 failing**. Only **5 of those 12 fail to boot** — the other 7 boot, serve the
fuzzed surface and fail on genuine findings, which is the lane working and is not this PR's subject.

Method note, because this repo has been bitten by it: the boot failures were identified by the
`::error::[<svc>] failed to boot` marker with the service name **substituted**, not the literal
`${svc}` that the job log also contains from its own `run:` script listing. Causes were read from the
executed `--- boot failure ---` section, not the echoed script.

| # | Service | Terminating cause (quoted from the run log) | Class |
|---|---|---|---|
| 1 | `openbank-sepa-payment` | `Connection refused: localhost/127.0.0.1:7233` (Temporal gRPC) | harness — **fixed here** |
| 2 | `openbank-transaction-service` | `Connection refused: localhost/127.0.0.1:7233` (Temporal gRPC) | harness — **fixed here** |
| 3 | `openbank-domestic-payment` | `UnknownHostException: temporal-frontend.temporal.svc.cluster.local` | harness — **fixed here** |
| 4 | `openbank-lending-service` | `UnknownHostException: temporal-frontend.temporal.svc.cluster.local` | harness — **fixed here** |
| 5 | `openbank-settlement-service` | `FlywaySqlUnableToConnectToDbException: … Connection to localhost:5432 refused` | harness — **fixed here** |

Both groups are harness gaps: a dependency the harness should provide and did not. No service change
is in this PR.

## Group A — no Temporal server (4 services)

These four register a Temporal worker at boot. The harness provisions Postgres, Redis and Keycloak,
and no Temporal, so dev mode retried the gRPC channel until the 180 s readiness window expired:

```
[io.te.se.ChannelManager] Resetting gRPC connection backoff on the gRPC channel
  ManagedChannelOrphanWrapper{delegate=ManagedChannelImpl{logId=1, target=localhost:7233}}
```

Two of them (`domestic-payment`, `lending`) default to the in-cluster frontend DNS name, which does
not resolve on a hosted runner, so the same gap presents as `UnknownHostException`.

Fix: start `temporalio/temporal:1.8.2 server start-dev`. That is the CLI image's **in-memory** server
— it needs no extra database, unlike `auto-setup`. The namespace is read from the service's own
`openbank.temporal.namespace` and passed as `-n`, because a worker registering against a namespace
that does not exist fails the same way as no server at all; and a namespace list kept in the workflow
would be exactly the "gate scoped by a hand-kept list" shape this repo already has a rule about.
`-Dopenbank.temporal.server-url=localhost:7233` points every target at it regardless of its own
default.

## Group B — an unresolved config expression killed the database (1 service)

`settlement-service` writes its datasource username as a MicroProfile config **expression**:

```yaml
username: "${QUARKUS_DATASOURCE_USERNAME:openbank_settlement}"
```

The harness's `awk '{print $2}'` took that string literally and handed it to
`docker run -e POSTGRES_USER=…`. initdb rejected it, the container exited, and the service then met
`Connection to localhost:5432 refused` — so the boot log named **Flyway**, and never the container
that had already gone. The step now resolves `${VAR:default}` to its default (the harness sets no
env, so the default is what the service will use), strips quotes, and fails loudly if the result is
still not a plain identifier rather than passing it to docker.

## What this does not fix

Nothing is left in the boot group — all 5 are addressed. Deliberately **not** in scope:

- **The 7 services that fail on findings** (balance 3, delegation 4, ledger 2, interest 2, fx 1,
  fraud 1, swift 1 server errors). Those are what the lane exists to produce and need per-service
  triage, mostly already in flight.
- **Two non-fatal schema-validation ERRORs** that appear in these logs *above* the terminating cause:
  `missing table [object_store_blobs]` (sepa-payment, transaction, domestic-payment) and
  `wrong column type … [description] in table [collateral]` (lending). Boot continued past both in
  the recorded run — the Temporal chain is what terminated it — so they are not what this PR fixes.
  They are real drift and belong to #2872, and the verification run below is what establishes whether
  they matter once Temporal is present.

## Verification

Local, before pushing:

- `temporalio/temporal:1.8.2 server start-dev -n <ns>` starts and creates the namespace; the exact
  readiness probe used in the step (`temporal operator namespace describe`) went green after 2 polls.
- The new derivation was run against all 18 matrix services' `application.yaml`: usernames resolve to
  plain identifiers for all 18 (settlement now `openbank_settlement` instead of the raw expression),
  and a Temporal namespace is derived for exactly the services that declare one.
- `actionlint` finding set is byte-identical to `origin/main` (2 pre-existing SC2034 on the loop
  counter). Largest `run:` step is 6235 chars, well under the 19000 gate.

A real dispatch of this branch against the five services is running; its result is posted as a
comment on this PR. **Until that lands, "these five now boot" is a claim supported by the local
evidence above, not by a green run** — and a harness that boots a service but fuzzes nothing would be
the same failure one layer along, so what the comment must show is the fuzz step doing work, not just
the absence of the boot marker.

Refs #3039

### DIFF
--- a/.github/workflows/api-fuzz-authenticated.yml
+++ b/.github/workflows/api-fuzz-authenticated.yml
@@ -178,6 +178,10 @@ jobs:
           # reached `docker run -e POSTGRES_USER=${QUARKUS_DATASOURCE_USERNAME:openbank_settlement}`,
           # initdb rejected it, the container exited, and the boot log then blamed Flyway
           # ("Connection to localhost:5432 refused") — never the container that was already gone.
+          # Credentials are then passed EXPLICITLY below rather than left to whichever env var
+          # the service happens to read: the fleet mostly reads ${POSTGRES_PASSWORD}, which this
+          # job sets, while settlement-service reads ${QUARKUS_DATASOURCE_PASSWORD}, which it does
+          # not — so settlement met "password authentication failed" once its container did start.
           resolve_expr() {
             printf '%s' "$1" | sed -E 's/^"//; s/"$//; s/^\$\{[A-Za-z0-9_]+:([^}]*)\}$/\1/'
           }
@@ -239,6 +243,8 @@ jobs:
           ./gradlew ":${svc}:quarkusDev" \
             -Dquarkus.console.basic=true --console=plain --quiet \
             -Dquarkus.datasource.jdbc.url="jdbc:postgresql://localhost:5432/${DB}" \
+            -Dquarkus.datasource.username="${DBUSER}" \
+            -Dquarkus.datasource.password="${POSTGRES_PASSWORD}" \
             -Dquarkus.oidc.enabled=true \
             -Dquarkus.oidc.auth-server-url=http://localhost:8543/realms/openbank \
             -Dquarkus.oidc.client-id=openbank-services \
--- a/.github/workflows/api-fuzz-authenticated.yml
+++ b/.github/workflows/api-fuzz-authenticated.yml
@@ -172,8 +172,22 @@ jobs:
           SPEC="${svc}/src/main/resources/openapi.yaml"
           PORT="$(grep -m1 -A3 '^  http:' "$APP_YAML" | grep -m1 'port:' | grep -oE '[0-9]+')"
           DB="$(grep -m1 'jdbc:postgresql://\|postgresql://' "$APP_YAML" | sed -E 's|.*/([A-Za-z0-9_]+).*|\1|')"
-          DBUSER="$(grep -m1 '^    username:' "$APP_YAML" | awk '{print $2}')"
+          # `${VAR:default}` is a MicroProfile config EXPRESSION, not a literal. The harness sets
+          # no env, so the DEFAULT is what the service will actually use, and the raw string must
+          # never be handed to docker. settlement-service writes its username that way; the literal
+          # reached `docker run -e POSTGRES_USER=${QUARKUS_DATASOURCE_USERNAME:openbank_settlement}`,
+          # initdb rejected it, the container exited, and the boot log then blamed Flyway
+          # ("Connection to localhost:5432 refused") — never the container that was already gone.
+          resolve_expr() {
+            printf '%s' "$1" | sed -E 's/^"//; s/"$//; s/^\$\{[A-Za-z0-9_]+:([^}]*)\}$/\1/'
+          }
+          DBUSER="$(resolve_expr "$(grep -m1 '^    username:' "$APP_YAML" | sed -E 's/^[^:]*:[[:space:]]*//')")"
           DBUSER="${DBUSER:-openbank}"
+          case "$DBUSER" in
+            *[!A-Za-z0-9_]*)
+              echo "::error::[${svc}] datasource username did not resolve to a plain identifier: ${DBUSER}"
+              exit 1 ;;
+          esac
           echo "==> [${svc}] port=${PORT} db=${DB} user=${DBUSER}"
 
           docker run -d --name fuzz-pg -p 5432:5432 \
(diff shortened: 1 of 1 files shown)


JiRaska:
## Verification run [31688207540](https://github.com/JiRaska/open-bank-oss/actions/runs/31688207540) — Group A is proven, Group B moved one layer

Dispatched `api-fuzz-authenticated.yml` on this branch against the five boot failures
(`max_examples=25`): run [31688207540](https://github.com/JiRaska/open-bank-oss/actions/runs/31688207540).

**All four Temporal services now boot and the fuzz step did real work.** Verified two ways: no
`::error::[<svc>] failed to boot` marker (matched with the name substituted, not the literal
`${svc}` the log also contains), and a schemathesis summary with a non-zero tested-operation count —
a harness that boots a service but fuzzes nothing would be the same failure one layer along.

| Service | boot | Temporal line in the log | operations tested | fuzz outcome |
|---|---|---|---|---|
| `openbank-sepa-payment` | ✅ | `starting Temporal dev server (namespace openbank-payments)` | **7/7** | job **success**, all phases ✅ |
| `openbank-domestic-payment` | ✅ | `starting Temporal dev server (namespace openbank-payments)` | **6/6** | job **success**, all phases ✅ |
| `openbank-transaction-service` | ✅ | `starting Temporal dev server (namespace openbank-payments)` | **7/7** | `Server error: 2` — genuine findings |
| `openbank-lending-service` | ✅ | `starting Temporal dev server (namespace openbank-lending)` | **37/37** | `Server error: 2` — genuine findings |
| `openbank-settlement-service` | ❌ | n/a | — | see below |

Two of the four are outright green — the first `success` jobs this lane has ever produced for them —
and the other two now fail on findings, which is the lane working. That is +4 services reaching the
fuzz step, and 57 authenticated operations exercised that previously were not.

The two non-fatal schema-validation ERRORs called out in the description (`object_store_blobs`,
`collateral`) did **not** block any of the four. They remain real drift for #2872.

### Group B: settlement moved from "no database" to "wrong password"

The container now starts and the username resolves, and the next layer appeared:

```
Caused by: org.postgresql.util.PSQLException:
  FATAL: password authentication failed for user "openbank_settlement"
```

Root cause, and it is the same class as the first: the harness overrode only the JDBC **URL** and
left credentials to whichever env var the service happens to read. Most of the fleet reads
`${POSTGRES_PASSWORD:…}`, which this job sets — which is why every other service connects, by
coincidence rather than by design. settlement reads `${QUARKUS_DATASOURCE_PASSWORD:settlement}`,
which the job does not set.

The follow-up commit passes `quarkus.datasource.username` and `quarkus.datasource.password`
explicitly, so the harness no longer depends on that coincidence. A verification run for settlement
alone is in flight ([31688625829](https://github.com/JiRaska/open-bank-oss/actions/runs/31688625829));
its result is posted here. Until it lands, settlement is **not** verified — Group A is.

JiRaska:
## Settlement verified — all 5 boot failures are fixed, 5/5 reach the fuzz step

Run [31688625829](https://github.com/JiRaska/open-bank-oss/actions/runs/31688625829),
`openbank-settlement-service` alone, on this branch: job **success**.

```
==> [openbank-settlement-service] port=8138 db=settlement user=openbank_settlement
==> [openbank-settlement-service] starting Temporal dev server (namespace openbank-settlement)
==> [openbank-settlement-service] fuzzing authenticated surface on :8138
API Operations:  Selected: 1/1   Tested: 1
Test Phases:     Examples OK   Coverage OK   Fuzzing OK
```

No `[openbank-settlement-service] failed to boot` marker (matched with the name substituted), the
username resolves to the identifier rather than the raw `${VAR:default}` expression, and the fuzz
step tested the operation rather than merely starting.

### Final state of the boot group

| Service | before | now | evidence |
|---|---|---|---|
| `openbank-sepa-payment` | Temporal gRPC refused | **boots, job success** | 7/7 operations tested |
| `openbank-domestic-payment` | Temporal DNS unresolvable | **boots, job success** | 6/6 operations tested |
| `openbank-transaction-service` | Temporal gRPC refused | **boots**, fails on findings | 7/7 tested, `Server error: 2` |
| `openbank-lending-service` | Temporal DNS unresolvable | **boots**, fails on findings | 37/37 tested, `Server error: 2` |
| `openbank-settlement-service` | no database / wrong password | **boots, job success** | 1/1 operation tested |

**5 of 5 boot failures fixed; 58 authenticated operations now exercised that previously were not.**

Two caveats stated plainly rather than implied away:

1. These runs used `max_examples=25` and covered only these five services. The other 13 in the matrix
   were **not** re-run on this branch, so "the full matrix now reaches 18/18" is not established here
   — only that no service in the boot group still fails to boot. The credential change touches every
   service's boot command, and the honest check for that is a full-matrix dispatch after merge.
2. The remaining red in the lane is `Server error` findings on services that boot. Those are the
   lane working and belong to their own triage, not to #3039's boot half.

PR #4531, opened 2026-08-13
ci(governance): falsify three more gates — and widen a scope that covered 8 of 23 services

JiRaska:
Batch 4. Debt **39 → 36**.

## The finding that matters is not a self-test

`check-clock-injection.sh` carried its money-path scope as an **eight-name bash array** while `rules.yaml: money_path_services` holds **23**. Fifteen money-path services were never scanned — and the gate printed *"no direct wall-clock calls found in domain/application layers of money-path services"* about a scope that excluded two thirds of them.

| | before | after |
|---|---|---|
| money-path services scanned | 8 | **23** |
| files scanned | 142 | **293** |
| violations | 0 | **0** |

That is this repo's own rule about a gate whose **scope** is a hand-kept list of the thing it checks: it reads as passing when the list is short, never as *unchecked*. The list is now **derived from `rules.yaml`**, and an empty derivation fails loudly instead of scanning nothing.

Measured *before* changing it: widening to all 23 surfaces **zero** new violations — so the fifteen were clean, and this costs no debt. But nothing had established that, and nobody could have known it from the green.

## Also falsified

**`domainevent-occurredat-constructor-guard`** — a `DomainEvent()` subclass that omits `occurredAt` takes the base default, which was `Instant.EPOCH`: a lie every test agrees with, because `isNotNull()` passes against 1970-01-01, and **23 of 25** fleet call sites took it (#3882). 5 cases including the test-file and build-output exclusions, and an empty tree that must report `SUBJECTS=0`.

**`operator-write-naming`** — the discovery half of GHSA-58jq-9hq3-66jr. The shared service-account carries `ROLE_OPERATOR`, so a role-gated unpinned write admits *every backend* to that action; `rest.rego` only blocks reasons someone **registered**, and registration depends on the rule being findable by name. A silent failure here is invisible by construction — the register just stays short, and **a short register reads exactly like a closed gap**. Covers the classifier's six branches plus **both authoring shapes**: a standalone `*_rest_ext.rego` and a rule inside a `gen-*-opa-bundle.sh` heredoc, which a `.rego`-only sweep loses entirely.

## And one of mine, for the fourth time this session

The empty-scope case asserted only the **exit code** — but without the guard, bash dies on `set -u` (`unbound variable`), which is also non-zero. A crash and a verdict are both "red", and a test that cannot tell them apart passes while the guard it checks is gone. It asserts the message now.

## Verdicts over the real tree, unchanged
293 money-path files clean · 3165 `.kt` files clean · 102 `allowed_reasons` rules, 0 new.

Shards: `lint 18/18 · kotlin 19/19 · registry 26/26 · security 12/12`.


JiRaska:

PR #4537, opened 2026-08-13
ci(governance): falsify four small guards — and remove a filter that hid real violations

JiRaska:
Batch 5. Debt **39 → 35** on this branch.

## The defect found

`check-test-runblocking-unit.sh` filtered its own findings with `grep -v ": Unit"` — a **line-level** exclusion. The detection pattern already requires `) = runBlocking`, so the fixed form `(): Unit = runBlocking` could never match it: the filter was **redundant against the case it was written for**.

What it did instead was drop any line containing `: Unit` anywhere:

```kotlin
@Test fun `pays`(p: Unit) = runBlocking { ... }   // real violation, NOT reported
```

A genuine violation whose *parameter* carried that type was silently skipped — by the gate whose entire job is to stop tests being silently skipped. Measured both ways: **flagged after the removal, clean before it.** Two fixtures now pin it, and reintroducing the filter turns the self-test red.

## Also falsified

| gate | cases | the ones that matter |
|---|---|---|
| `admin-ui-version-sync-guard` | 4 | the **absent** case — the skip must be visible, or a moved path reads exactly like a passing gate |
| `quarkus-application-version-override-guard` | 6 | the **nesting** cases — the same `version:` leaf under `quarkus.container-image`, or a non-quarkus root, is a different property |
| `no-service-local-exceptionmapper-collision` | 5 | a local mapper for a libs-runtime-owned JDK type collides non-deterministically (#526) |

## And one of my own

The comment-filter fixture for the ExceptionMapper guard used prose the detection pattern **could never match anyway**, so the branch it claimed to cover was never reached — and the deliberate break that removed that filter went **uncaught**. The fixture now carries the full `: ExceptionMapper<...>` signature, as a real fix comment would.

Every verdict over the real tree is unchanged. Shards: `lint 18/18 · supplychain 20/20 · kotlin 19/19`.

### DIFF
--- a/.github/gates/gates.yaml
+++ b/.github/gates/gates.yaml
@@ -677,6 +677,8 @@ gates:
     name: "admin-ui version sync guard (version.txt == package.json, enforced)"
     group: supplychain
     mode: enforced
+    selftest: bash .github/scripts/check-admin-ui-version-sync.sh --self-test
+    selftest_expect: pass
     run: |
       bash .github/scripts/check-admin-ui-version-sync.sh .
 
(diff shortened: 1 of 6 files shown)


JiRaska:

PR #4539, opened 2026-08-13
ci(governance): falsify three governance gates — one failed with NO output at all

JiRaska:
Batch 6. Debt **39 → 36** on this branch.

## The defect found: a gate that failed with no output at all

`check-agent-charter-registry.sh` exited **1 with zero output** on an `agents.yaml` containing no ids. Under `set -euo pipefail`, the `grep -E '^  - id:'` that builds the id list exits 1 when it matches nothing — killing the script **before** the explicit *"no `  - id:` entries found"* error it was supposed to print.

That message was **unreachable dead code**. CI showed a red check and no reason for it, on the one input that means the gate could not read its own subject. `|| true` on that pipeline makes the intended error reachable.

## Gates falsified

**`gate-graduation-guard`** (7 cases) — ADR-0144's own enforcement: an advisory/planned rule must carry a `target_enforce_date` that has not passed, or "advisory" is permanent by default. It is an awk state machine with a **3-line lookahead**, so the cases that matter are about distance and order: a date four lines away must *not* be credited.

**`agent-charter-registry-parity`** (6 cases) — both directions, plus the frontmatter check I had not read closely enough to expect: an id with no charter, a charter naming no declared agent, and **a charter whose frontmatter id disagrees with its filename** — a half-finished rename leaves a doc describing the wrong agent, which is worse than none because it reads as current.

**`mcp-charter-data-scope-binding`** (6 cases + a live read) — binds a **charter claim to a code seam**: `McpToolRegistry` masks PII unconditionally, so the charter must not advertise that it can return unmasked data. A *missing* `data_scope` key must not read as agreement, and an absent charter must fail loudly — a rename silently unbinds the claim from the code.

The live read matters here: `REQUIRED` names a charter id, and a fixture-only self-test cannot tell that the id still refers to anything.

Verdicts over the real tree unchanged. Shards: `lint 18/18 · registry 26/26 · supplychain 20/20`.


JiRaska:

PR #4575, opened 2026-08-13
ci(governance): falsify three more gates — and a second gate that failed with EMPTY output

JiRaska:
Stacked on #4571. **Debt 11 → 8**; 116 of 142 gates declare a self-test.

## The defect — and it is the *second* instance of this shape in this session

`check-deploy-coverage.sh` carries the message

```
ERROR: could not read ALL_SERVICES from .github/workflows/auto-deploy.yml
```

and **had never once printed it**. Under `set -euo pipefail` a no-match `grep` exits 1, the pipeline inherits it, and the script dies **at the assignment** — before the explicit check one line below, which is therefore unreachable code.

Measured against a workflow with no `ALL_SERVICES`:

```
rc=1   output=[]
```

Empty. Indistinguishable from a runner crash. `|| true` on that pipeline is load-bearing, not defensive noise.

Same shape as `check-agent-charter-registry.sh` in batch 6 — **two scripts, written independently, both carrying diagnostics that could never fire.**

## The three gates

|---|---|---|
| `deploy-coverage-guard` | 8 | a service release-please **releases** and gitops **references** but auto-deploy never **builds** — the release PR merges, the tag is cut, and the pod keeps serving the old image. The only symptom is a version number that stops moving |
| `db-migration-gate` | 15 | a Flyway migration is checksummed once applied, so the rollback plan exists **before** the merge or never |
| `ai-act-high-risk-inventory-vs-code` | 11 + live | `ml-systems.yaml` says the credit engine is `deployed: false`, and the compliance documentation rests on that |

### The sharpest cases
- **`-- rollback:` with no content must not satisfy the rule**, or it degrades to "type the word rollback" — the failure that looks exactly like compliance.
- Deploy-coverage parses **three inputs from three files by three different means**, each able to come back empty — and an empty input makes the comparison *agree*. The missing-input cases assert `rc=2`, not a verdict.
- The AI-Act scanner is a pattern list, so it fails in exactly one direction: a spelling it does not know is a site it does not see. Every pattern is pinned individually.

**13 deliberate breaks, every one caught.** Verdicts unchanged. Shards: `lint 18/18 · data 13/13 · supplychain 20/20`.


JiRaska:

PR #4577, opened 2026-08-13
feat(campaign): add in-app banner channel

JiRaska:
## What changed

- Adds `BANNER` as a first-class Campaign Studio channel with a dedicated in-app phone preview, closed deep-link destination and approved banner template.
- Delivers an idempotent, consent-gated HOME_BANNER placement through a compacted Kafka command to engagement-service; it never routes via notification-service.
- Persists and serves the current per-party campaign banner, forwards authenticated banner interactions through customer-edge, and attributes only the assigned banner to the campaign.
- Adds Kafka ACLs/topic, analytics channel coverage, OpenAPI/AsyncAPI and reversible Flyway migrations.

## Safety

- Banner uses `MARKETING_COMMS_INAPP` / promotional-impression consent, a closed `openbank://` route, an opaque interaction reference and no PII in analytics.
- The event endpoint rejects attaching a `BANNER` interaction to a generic catalogue card.

## Validation

- `npm run type-check`
- `npx vitest run src/test/journey-editor.test.tsx src/test/campaign-experience-preview.test.tsx src/test/campaign-builder-interaction.test.tsx` — 22 passed
- `./gradlew :openbank-campaign-service:test --tests "*CampaignJourneyActivitiesTest" :openbank-engagement-service:test --tests "*ResolveSurfaceUseCaseTest" --tests "*SurfaceRestContractIT"`
- `./gradlew :openbank-campaign-service:ktlintCheck :openbank-engagement-service:ktlintCheck :openbank-customer-edge:ktlintCheck`
- `git diff --check`

## Linked issues

Refs #4476

### DIFF
--- a/.github/event-contract-baseline.txt
+++ b/.github/event-contract-baseline.txt
@@ -13,7 +13,6 @@ openbank-account-service:openbank.accounts.account.created
 openbank-account-service:openbank.notification.requests
 openbank-aml-service:openbank.aml.events
 openbank-balance-service:openbank.balance.events
-openbank-campaign-service:openbank.notification.requests
 openbank-card-issuance-service:openbank.cards.events
 openbank-clearing-service:openbank.clearing.batch.event
 openbank-consent-service:openbank.consent.events
--- a/openbank-admin-ui/src/app/campaigns/new/page.tsx
+++ b/openbank-admin-ui/src/app/campaigns/new/page.tsx
@@ -65,12 +65,14 @@ const TEMPLATES: Record<string, string[]> = {
   // One variable, and that is the channel's rule rather than a simplification: a push renders its
   // title plus a fixed generic body, so there is nowhere for offer copy to go (#1182).
   MARKETING_PRODUCT_OFFER_PUSH: ['offerTitle'],
+  MARKETING_PRODUCT_OFFER_BANNER: ['offerTitle', 'offerText', 'ctaText'],
 }
 
 /** Which channel each template renders on. The service refuses a step whose two disagree. */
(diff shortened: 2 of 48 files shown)


JiRaska:
The red here is a **deploy-ordering** gate and it is right — worth not re-running past it, because
what it prevents is a service that cannot start.

```
openbank-engagement-service: GitOps config declares messaging channel
`campaign-banner-placements-in`, but the deployed image (sandbox-22ab0ba6) has no
`connector:` for it in application.yaml.
```

SmallRye treats a configured channel with no connector as **SRMSG00071, a hard boot failure** — so
this is not a channel that quietly does nothing, it is engagement-service failing to start. The
config would land before the image that can serve it.

The order that works: merge so the **image** carrying the `mp.messaging.incoming.campaign-banner-placements-in`
block with its `connector:` is built and deployed first, then the gitops config that names the
channel — or both in the same change, so no intermediate state exists where the config is ahead of
the image.

Two things worth checking while you are in there, both measured on this repo before:

- A Kafka `group.id` / `auto.offset.reset` written as a YAML **key** never reaches the connector.
  SmallRye's YAML source quotes any leaf key containing a literal dot, so it registers as
  `…<channel>."group.id"` and the connector silently uses its own default. Match exactly what the
  service's existing channels do rather than what reads correctly.
- `application.yaml` must carry no duplicate mapping key: SnakeYAML keeps only the **last** of a
  repeated key and drops the rest with no error. `yaml.safe_load` succeeding proves nothing — use a
  loader that raises on duplicates.

Not touching the branch; flagging because the failure mode is a stopped service rather than a failed
check.

PR #4579, opened 2026-08-13
ci(governance): falsify the openapi-route and db-backup gates

JiRaska:
Stacked on #4575. **Debt 8 → 6**; 118 of 142 gates declare a self-test.

|---|---|---|
| `openapi-route-conformance` | 16 | a route **published and served nowhere**, or served and published nowhere — both silent, because a spec is a document and nothing at runtime disagrees with it. It surfaces as a client 404 against an endpoint the documentation promises |
| `db-backup-association-gate` | 9 | a CNPG cluster with a `barmanObjectStore` whose ServiceAccount has no pod-identity association takes **no backups** — recorded in CNPG's own status, nowhere a human looks, and discovered at **restore time** |

### The asymmetry that makes route-conformance hard
Every helper compares strings across two very different sources, and they fail in opposite ways: a normaliser that is **too eager** makes real drift disappear (two distinct routes collapse into one), one that is **too literal** manufactures drift out of a trailing slash. Both directions are pinned — including that a path **parameter** is part of a route's identity, since collapsing `{id}` would merge `/accounts/{id}` with `/accounts/{iban}`.

The `externally_served` exemption is the sharpest case. It exists because quarkus-smallrye-health and libs-runtime's `ServiceInfoResource` answer routes the JAX-RS scan cannot see. **Widen it and a published-but-unimplemented endpoint is silently exempted** — the gate's whole subject.

### Two scope cases that decide whether the backup gate survives contact
- A cluster with **no** `barmanObjectStore` is not backing up, so demanding an association for it is wrong.
- **Strimzi also uses `kind: Cluster`.** Without the apiVersion filter every Kafka manifest scores as a database with missing backups — the noise that gets a gate switched off, which is the same outcome as not having one.

**10 deliberate breaks, every one caught.** Verdicts unchanged. Shards: `lint 18/18 · gitops 27/27`; `api` fails only on `sudo` for an `oasdiff` install (environmental).


JiRaska:

PR #4581, opened 2026-08-13
ci(governance): falsify the MCP caller-auth guard and the AI-governance snapshot

JiRaska:
Stacked on #4579. **Debt 6 → 4**; 120 of 142 gates declare a self-test.

## A defect I introduced, and the falsification caught

`gen-ai-governance-snapshot.py` ends in a bare `main()`, and `main()` is declared `-> None` — its other exits go through `fail()` raising `SystemExit`, so a **returned** code is discarded. My new `--self-test` therefore printed four failures, said `self-test FAILED`, and the process **exited 0**.

Four of six deliberate breaks read as NOT CAUGHT until I looked at why. **A harness whose verdict the process throws away is not a harness.** Now `sys.exit(main() or 0)`; `--check` is unaffected (it fails through `SystemExit`, verified against the live snapshot).

## The two gates

|---|---|---|
| `mcp-real-port-requires-caller-auth-first` | 9 | a real `AccountReadPort` wired while `resolveContext()` still returns `agent:mcp-anonymous` — an **unauthenticated read of any account** |
| `ai-governance-snapshot-drift` | 33 | the page admin-ui renders as the platform's AI-governance posture, **read by people deciding whether a control exists** |

### Two states with opposite correct behaviour, both pinned
**Pre-cutover** a real port is a blocker and `@Vetoed` is the declared escape. **Post-cutover** real ports are the point and must *not* be blocked — but reintroducing the shared literal collapses every caller onto one charter's grant, and a new caller silently inherits the previous one's authorization **while the audit trail still shows per-caller identity**. That is what makes it undetectable afterwards.

The comment stripper is load-bearing in both directions: the KDoc explaining this rule necessarily contains the forbidden literal, so a gate that counted comments would **block its own documentation**.

### A dropped decision is the sharpest snapshot case
Eight rows read as complete, because a row that is not there cannot look wrong. D1–D9 must appear exactly and in order, the status vocabulary is closed, and `compliance` plus all four `auditTrail` lists must be non-empty — an empty section reads as *"nothing to report"* rather than *"nobody supplied the evidence"*.

**11 deliberate breaks, every one caught after the exit-code fix.** Verdicts unchanged. Shards: `lint 18/18 · registry 26/26 · security 12/12`.


JiRaska:

Issue #4553, opened 2026-08-13
bronze_events.aggregate_type is not normalised: one account has two current-state rows, and a Grafana tile reads 0 instead of 4


JiRaska:
`bronze_events.aggregate_type` holds two spellings of the same domain, so
`silver_current_state` — the projection whose entire contract is one row per aggregate — emits **two
current-state rows for one account**. Measured on the sandbox warehouse 2026-08-13.

## The data

```
aggregate_type      source_service   rows
ACCOUNT             unknown          294
Account             unknown           17
Consent             unknown           10     <- no CONSENT rows exist at all
Transaction         unknown           43     <- no TRANSACTION rows exist at all
PARTY               unknown           17
ONBOARDING_FUNNEL   customer-edge    512
SCREEN_FEEDBACK     customer-edge      8
```

Five account ids appear under **both** `ACCOUNT` and `Account` (one of them across 34 rows). Since
`silver_current_state` groups by `(aggregate_type, aggregate_id)`, each of those accounts has two
"current" states that never reconcile:

```
silver_current_state   ACCOUNT 14 | Account 9 | PARTY 4 | Consent 5 | Transaction 23
```

A reader filtering one spelling gets a stale row as "current" whenever the other spelling carried the
newer event. Last-writer-wins cannot fire across a group boundary.

## Root cause

`AnalyticsConsumer.resolveAggregateType` takes the producer's value verbatim:

```kotlin
private fun resolveAggregateType(node: JsonNode, address: EventAddress): String =
    node["aggregateType"]?.asText()                       // <- producer-supplied, NOT normalised
        ?: inferAggregateType(node).takeIf { it != UNKNOWN }   // UPPERCASE
        ?: TopicAttribution.aggregateType(address.topic)       // UPPERCASE
        ?: UNKNOWN
```

Both fallbacks are uppercase; the envelope path is whatever the producer wrote. So the spelling
records **which code path attributed the event**, not what the aggregate is — and one producer
changing its envelope silently forks a domain into two.

## Live consequence, measured

The business-warehouse Grafana dashboard queries mixed-case literals
(`dashboard-openbank-business-warehouse.yaml`):

| tile | query filter | reads | truth (`upper(...)`) |
|---|---|---|---|
| parties, 30d | `aggregate_type = 'Party'` | **0** | **4** |
| accounts opened, 30d | `aggregate_type = 'Account'` | 0 | 0 |

[...]

PR #4582, opened 2026-08-13
ci(governance): widen deploy-coverage past release-please's registry, wire analytics-sink

JiRaska:
## Summary

The #4576 normalisation merged clean and did **not deploy**: `openbank-analytics-sink` has never
been in auto-deploy's `ALL_SERVICES`, so the run reported "No deployable services changed" and
succeeded, deploying nothing. The service's own coverage gate — the one that exists exactly to catch
this — passed every time, because its subject set was release-please's *released* components, and
analytics-sink owns no OLTP database and cuts no release (ADR-0022), so it has no `version.txt`.
This widens that subject set so a service can be visible to the gate without being a release-please
package.

## Type of change

- [x] fix (bug fix) — plus `ci` scope: this is governance/gate machinery, held for human review by
      the merge guard regardless of title, same as #4520.

## Linked issues

Refs #4553 (the normalisation this PR makes reach the running service)

## Changes

1. **`ALL_SERVICES` gains `openbank-analytics-sink`.** Nine commits since 2026-07-26 (#2654) never
   reached the cluster, including #4576's fix. This is the direct unblock.

2. **`check-deploy-coverage.sh`'s subject set widens**: released components (unchanged) **plus**
   anything wearing a `sandbox-<sha>` tag in `openbank-infra/gitops`. That tag shape is auto-deploy's
   own output, so it defines the set without a second hand-kept registry. Widening immediately found
   two more workloads the old gate could not see:

   | workload | why it's not in `ALL_SERVICES` | disposition |
   |---|---|---|
   | `openbank-developer-portal` | static nginx site | **baselined** — not Quarkus |
   | `openbank-document-renderer` | Python app | **baselined** — not Quarkus |

   Both have real Dockerfiles and real gitops manifests, deployed and presumably stale like
   analytics-sink was. Not force-fit into `ALL_SERVICES`: the "Build fast-jars" step is
   `./gradlew ... -Dquarkus.package.jar.type=fast-jar` — pointing a static site or a Python app at it
   fails outright. Wiring their build is real work, so it's recorded as debt in
   `deploy-coverage-baseline.txt` with the reasoning inline, not silently patched with the wrong tool
   and not silently ignored either.

3. **`--self-test`, 5 cases**, and `min_subjects: 45` (60 today) registered in `gates.yaml` — the
   same falsifiability shape as #4520 and #4576's `SUBJECTS=` line.

4. **Ratchet paid**: `deploy-coverage-guard` removed from the stale-debt lists in
   `check-gate-subject-floor.py` and `check-gate-selftest-declaration.py` — it now declares both, so
   leaving the entries would fail those gates' own "no longer violating" check.

## Verification

**The core claim, falsified rather than asserted**: the *old* gate script, run against this exact
tree (including the baseline entries), reports:

```
DEPLOY-COVERAGE GATE: PASS — every released component ArgoCD deploys is built by auto-deploy.
```

— while analytics-sink sat undeployed for 18 days under it. The new gate, same tree:

```
SUBJECTS=60
DEPLOY-COVERAGE GATE: PASS — no NEW gaps. 2 known gap(s) remain in deploy-coverage-baseline.txt.
```

Both PASS, but only one of them ever looked at analytics-sink — that's the defect, made visible by
running the same before/after comparison every prior fix in this thread has used.

- `--self-test`: 5 cases (analytics-sink's exact shape flagged; a version-pinned third-party image
  like `openbank-keycloak:26.6.3-optimized` NOT flagged; a clean tree passes; a stale baseline entry
  fails the ratchet; an empty subject set refuses to read as clean).
- `min_subjects` falsified both directions: 45 passes, 999 fails.
- `python3 .github/scripts/run-gates.py --group supplychain`: 19/20 PASS, the one FAIL is
  `release-scope-mismatch-gate` refusing to run with no `PR_DIFF_BASE` set — a local-only artifact of
  not running inside the PR's own CI context, not a finding about this change.
- `ruff check` on both edited debt-list scripts: identical finding count before/after my one-line
  deletions (2 and 1 respectively, both pre-existing `EXE001`/`I001`, neither touched here).
- `shellcheck` SC2086 on the new loop line: same unquoted-list idiom the file already uses two lines
  below it — not a new style, not fixed here (out of scope for this PR).

## A bug in the self-test itself, worth flagging to reviewers

`set -o pipefail` is active for the whole script, so the first draft of `_run() { ... bash "$SELF" |
grep -q ...; }` took its exit status from the **pipeline**, not from `grep`. A fixture that correctly
made the gate exit 1 therefore made `if _run ... | grep -q ...` read as "no match" regardless of
whether grep matched — three of five cases failed for a reason that had nothing to do with the gate.
Fixed by capturing output to a file first and grepping the file. Leaving this in the description
because it's the kind of thing that looks like the gate is broken when it's the harness.

## Definition of Done

- [x] No Kotlin/service code — workflow YAML + governance scripts only.
- [x] Self-test added and falsified in both directions.
- [ ] Integration/contract/OpenAPI/Flyway — N/A.

## Security checklist

- [x] No secrets, no PII, no new dependency.

## Compliance impact

- [x] None.

## Rollout / Rollback

- Deployment strategy: this PR itself changes no running workload — it only changes what the CI
  pipeline builds on the NEXT push. `openbank-analytics-sink`'s next push (or a manual
  `workflow_dispatch` with `reconcile=true`) will deploy #4576's normalisation, 18 days late.
- Rollback: revert. `ALL_SERVICES` loses analytics-sink again, baseline entries return.

## Reviewer notes

Recommend triggering `auto-deploy.yml` with `reconcile=true` after this merges, rather than waiting
for the next incidental push — that's how #4576 actually reaches the cluster.


JiRaska:
`Gitleaks` is a false positive here, and the repo already has a precedent for exactly this shape.

```
RuleID:   generic-api-key
File:     .github/scripts/check-deploy-coverage.sh
Finding:  ...vice" "  image: ecr/openbank-keycloak:REDACTED" '"openbank-account-...
```

The match spans a **fixture block of image references** — the rule fires on the shape
`<key>: <token-like-string>` regardless of whether a secret is involved. Same class as #2423, where
a purely numeric sandbox build id (`…/openbank-admin-ui:sandbox-3567574752`) tripped the `rod-cislo`
heuristic and blocked deploy PRs.

Two ways to clear it, and the repo's own `.gitleaks.toml` shows which is preferred:

1. **A scoped path allowlist with a written reason**, following the existing entries — those are
   deliberately narrow (`openbank-infra/gitops/components/[a-z0-9-]+/[a-z0-9._-]+\.yaml$`,
   `openbank-admin-ui/(cluster-topology|infra-lifecycle)\.json$`) and each carries a comment saying
   why the file cannot hold a real secret. A `.github/scripts/check-deploy-coverage\.sh$` entry
   fits: it holds image refs used as test fixtures, and secrets in this repo flow via OpenBao,
   never plaintext.
   Note the file's own warning while you are in there: *"a scoped `regexTarget="line"` allowlist
   does not apply under `[extend] useDefault` in gitleaks 8.x"* — so the per-rule scoped form will
   not work, which is why the existing entries are path-based.
2. **Restructure the fixture** so the tag is not adjacent to the quote-heavy blob. Narrower, no
   config change, but it makes the fixture read worse to satisfy a scanner.

One thing worth knowing regardless of which you pick: **gitleaks scans the push RANGE, not the tip**,
so a fixup commit does not clear it — the offending commit has to leave the branch history
(amend + force-push, or squash the branch).

Not touching the branch.

PR #4585, opened 2026-08-13
ci(governance): falsify the runbook DR classifier and the EU AI Act drift check

JiRaska:
Stacked on #4581. **Debt 4 → 2**; 122 of 142 gates declare a self-test.

## A defect the self-test found by causing it

`check-eu-ai-act.sh` diffs the document under `$ROOT` but runs the generator from the **current directory**. The generator writes to a relative path, so pointing the script at any other root regenerates a file somewhere else entirely while the diff inspects an untouched one — the check then compares a file **against itself** and reports `in sync` whatever the truth is.

Harmless while CI always runs from the repo root; a live hazard the moment anything passes a root. The first version of the fixture below did exactly that and **overwrote the real `docs/compliance/eu-ai-act.md`** in the working tree (git restored it — the file is tracked). Fixed with `cd "$ROOT"` before generating, and the `cd-root` break is now one of the three the self-test catches.

## The two gates

|---|---|---|
| `service-runbook-drift` | 17 | a runbook that tells an on-call engineer at 3am to *"restore from the datastore's managed backup"* for a service with **no database and no such backup** |
| `eu-ai-act-inventory-drift` | 3 | a stale compliance inventory — silently stale, because a generated file nobody regenerates looks exactly like one that is current |

### Two distinctions that are not the same question
`is_stateless(datastore)` means **no datastore at all**. `ownsNoDatabase` means no **database** while possibly holding real state — customer-edge declares Redis + `ownsNoDatabase`, and its Redis holds **durable TTL-less passkey credentials**. Conflating them produces confident wrong instructions in *both* directions.

The sharpest case is a PostgreSQL cluster with no backup: the procedure must be prefixed by **"DR is not achievable today"**, because printing the same steps without that line reads as a working runbook at the exact moment it is being followed.

### A generator that cannot run is not "no drift"
That third case is the one that matters most — without it a broken generator turns the gate into a no-op that reports success.

The fixture builds a **throwaway git repo**. It must: the production path calls `git checkout --` on a finding, so a test that tampered with the working copy would have its evidence reverted by the thing it tests.

**8 deliberate breaks, every one caught**, and the working tree is byte-identical afterwards.

## Remaining debt (2), left deliberately
`adr-registry-integrity-check` and `gen-network-policies-drift-gate`. Both regenerate into the working tree and call `git checkout --` on a finding, so each needs its own fixture git repo — and adr-registry alone carries ten distinct rules across 395 lines. That is a batch of its own, not an appendix to this one.

Shards: `lint 18/18 · registry 26/26 · gitops 27/27`.


JiRaska:
Superseded by #4640, rebased onto #4637 after this branch's head (ci/self-tests-batch-12) went CONFLICTING against main. All content carries forward; no work lost.

PR #4588, opened 2026-08-13
feat(campaign): add journey path experiments

JiRaska:
## Summary

- lets the durable B cohort compare a different template, delivery channel and wait time, not only alternate copy
- keeps retries deterministic: assignment is persisted before the Temporal journey starts
- adds the authoring controls, OpenAPI contract, domain validation and REST/DB coverage

## Delivery

Stacked on #4586, which is stacked on #4577. Merge in that order.

Refs #4476

## Verification

- ./gradlew --offline --console=plain :openbank-campaign-service:test --tests com.openbank.campaign.domain.CampaignStepChannelTest --tests com.openbank.campaign.application.workflow.CampaignJourneyWorkflowTest
- ./gradlew --offline --console=plain :openbank-campaign-service:test --tests com.openbank.campaign.integration.CampaignRestContractIT
- ./gradlew --offline --console=plain :openbank-campaign-service:ktlintCheck
- npm run type-check
- npx vitest run src/test/journey-editor.test.tsx src/test/campaign-studio.test.tsx

### DIFF
--- a/openbank-admin-ui/src/app/api/campaigns/[id]/route.ts
+++ b/openbank-admin-ui/src/app/api/campaigns/[id]/route.ts
@@ -124,7 +124,7 @@ export async function GET(_req: Request, ctx: { params: Promise<{ id: string }> 
   const { id } = await ctx.params
   const headers = { authorization: `Bearer ${session.user.accessToken}` }
 
-  const [campaign, enrolments, sends, sendSummary, journey, experiment] = await Promise.all([
+  const [campaign, enrolments, sends, sendSummary, journey, engagement, experiment] = await Promise.all([
     read(headers, `/api/v1/campaigns/${encodeURIComponent(id)}`, null),
     read(headers, `/api/v1/campaigns/${encodeURIComponent(id)}/enrolments`, []),
     // First page only. Paging and filtering go through /api/campaigns/[id]/sends so turning a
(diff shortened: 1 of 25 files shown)


JiRaska:
`build (openbank-campaign-service)` fails on **detekt**, not on tests — and the same two findings
appear on both #4588 and #4593, so whichever lands first leaves the other conflicting on the fix.

```
openbank-campaign-service/src/test/kotlin/.../integration/CampaignRestContractIT.kt:46:7
  Class CampaignRestContractIT is too large. [LargeClass]

openbank-campaign-service/src/main/kotlin/.../infrastructure/persistence/Persistence.kt:463:60
  This expression contains a magic number. [MagicNumber]
```

The gradle output buries these — the `What went wrong` block only says
`':openbank-campaign-service:detekt' … non-zero exit value 2`, and the finding lines sit above it
among ~40 `w:` compiler warnings from unrelated modules.

Two notes from this repo's own history, in case they save a round trip:

- **`MagicNumber` fires at the declaration site, so a `private const val` in the enclosing
  `companion object` clears it** (`ignoreConstantDeclaration` is on by default). That is how the
  fleet handles the percentile triple; a suppression is not needed.
- **`LargeClass` on an integration test** is usually a signal that one `@QuarkusTest` is carrying
  several unrelated scenarios. Splitting it costs a second Quarkus boot, which is real time — if the
  class is cohesive, a targeted `@Suppress("LargeClass")` with a one-line reason is the honest answer
  and reviewers here accept it. Don't raise the threshold globally for one file.

Not touching either branch — both worktrees have commits from the last few minutes.

PR #4592, opened 2026-08-13
ci(governance): falsify the last two gates — falsifiability debt reaches ZERO

JiRaska:
Stacked on #4585. **Debt 2 → 0.**

Every gate in the manifest now either **declares a self-test (124)** or is **exempt by kind (18)** — six *are* test suites, two are third-party linters, ten were classified by #4340.

## A third script with an unreachable failure path

`check-adr-registry.sh` ends with a repo-wide `git grep` for `ADR-NNNN` references. A no-match grep exits 1, `set -euo pipefail` propagates it, and the script dies **there** — before the verdict block, so the run ends `rc=1` having printed **no error at all**.

It never fires in this repo because ADR references always exist, which is exactly what makes it a trap: the first tree without one gets an undiagnosable failure. Found because the self-test fixture is a repo with no references.

Same shape as `check-deploy-coverage.sh` (batch 12) and `check-agent-charter-registry.sh` (batch 6) — **three scripts, written independently, each carrying diagnostics that could never print.**

## The two gates

| gate | cases | what it guards |
|---|---|---|
| `adr-registry-integrity-check` | 25 | the decision history the repo tells auditors to read |
| `gen-network-policies-drift-gate` | 13 | a NetworkPolicy is **default-deny** — a dependency the generator does not see gets no egress rule |

### The ADR failures are all quiet
Two ADRs on one number is what a **clean git merge** produces when both branches take the next free number — and then the number stops identifying a decision. An H1 that disagrees with its filename is a half-done renumber. A supersession declared on **one side only** leaves a replaced decision reading as live, with no hint where to go.

Two fixture decisions worth stating: it runs **outside this repo**, because check 5 greps repo-wide and would otherwise score every real reference as dangling against a three-file fixture; and it **omits `gen-index.sh`** so the index-freshness check skips — that check calls `git checkout --`, and a self-test must never be able to touch a working tree.

### The network-policy extractors fail in one direction only
Match less, generate less, **block more** — and the result appears as a connection timeout under load, attributed to the callee, while every manifest in the repo looks correct. Cases cover `https` as well as `http`, an absent port, cross-namespace calls, multiple dependencies in one blob, and the non-dependencies (an external host, a pod FQDN) that must **not** produce a rule.

Scope stated in place: this covers the extraction primitives, not `main()`'s whole rendering — the drift comparison around it proves the rendered output matches what is committed, and that half runs on every PR.

## The debt list is now empty, and marked as deliberately so
Rule 3 still fails on a stale entry **in either direction**, so it cannot rot back into a long list without someone deciding to let it.

**9 deliberate breaks, every one caught.** Verdicts unchanged (249 ADRs clean); the working tree is byte-identical after every run. Shards: `lint 18/18 · registry 26/26 · gitops 27/27 · supplychain 20/20`.


JiRaska:
Superseded by #4640, rebased onto #4637 after this branch's head (ci/self-tests-batch-12) went CONFLICTING against main. All content carries forward; no work lost.

PR #4606, opened 2026-08-13
feat(customer-edge): wire CardDelegationGuard into the real card path (ADR-0232 D3)

JiRaska:
Refs #2990 (ADR-0232 D3). ADR-0249 D2 is the design this implements.

## The distinction this PR settles: **unwired**, not advisory

`CardDelegationGuard` is **plumbing that was never connected**, not a deliberately advisory
component in a dual-run. Established three ways, from the code:

1. **Complete caller enumeration.** `git grep CardDelegationGuard` over every `src/main` in the
   tree returns exactly two production hits: its own declaration, and
   `CardDelegationResource` — a `GET /{id}/delegation/check` endpoint that returns
   `{"authorized": bool}` and nothing else. Nothing else in any service constructs or injects it.
2. **Nothing calls that endpoint either.** `grep -r "delegation/check"` across all Kotlin returns
   4 hits: the `@Path` declaration and **three tests**. Zero production callers, in this service
   or any other. So the loop was closed at both ends.
3. **The endpoint's own KDoc names the caller that never arrived** — *"the customer-edge slice asks
   here until it holds its own projection"* — and ADR-0249, written after it shipped, states it
   flatly: *"It has no production caller, which is a wiring gap, not a design gap."*

An advisory dual-run would look different and this repo has shipped that shape: a component
computing a shadow verdict that something *does* call and compare, or one gated behind a flag with
both branches live. There is no flag, no comparison, no second verdict, and no caller — the guard
was simply never plumbed in.

## What is wired, and the proof a real request path calls it

**`PUT /customer/v1/cards/{id}/limits`** — the guard's first production caller.
`CustomerEdgeResource.updateCardLimits` previously returned 403 whenever
`extractOwnerPartyId(cardJson) != caller`. It now falls back to `mayActOnCard(id, party,
MANAGE_LIMITS)`, which asks card-issuance `GET /api/v1/cards/{id}/delegation/check`, whose handler
is `CardDelegationGuard.isAuthorized`. Proof it is a real path, not a test path: the route is a
JAX-RS `@PUT @Path("/cards/{id}/limits")` on `CustomerEdgeResource`, already served in production
(it is in the edge's `openapi.yaml` and has 8 existing tests); the new test asserts the outbound
call is made *with* `intent=MANAGE_LIMITS`, so the assertion fails if the wiring is removed.

**`GET /customer/v1/cards`** — a `CARD_VIEW` grant now shows the card. Shared cards are appended,
each marked `sharedWithMe`, mirroring `sharedAccountsFor`/`listAccounts` line for line. Masked-PAN
only; the SCA-gated PAN reveal stays holder-only and is deliberately not delegable.

**Why two different mechanisms, and why that is not an invention.** The edge already runs both, on
purpose: reads ask delegation-service live (`hasGrant`, revocation-immediate, no cache), while the
authoritative money decision asks the service that holds the projection *and* the resource's true
owner (`resolveDebitAuthority` → account-service). Cards follow the same split — the list is a
read, the limits verdict needs the card's true holder, which only card-issuance has. The residual
asymmetry (a revoked grant stops re-limiting only once `openbank.delegation.events` is consumed) is
named in the threat model, and is the reason no delegated card action moves money or reveals a PAN.

## AC13 — the allowlist entry, whose own re-triage condition had fired

Removed `openbank.delegation.events` from `rules.yaml: event_consumer_liveness.allowlist`. Its
reason text ended *"Re-triage: remove this entry with the first projection consumer PR"*; those
consumers merged in #3058/#3143. Verified locally, not assumed:

```
$ python3 .github/scripts/check-event-consumer-liveness.py --enforce
check-event-consumer-liveness: 40 producer topic(s) scanned across 59 service(s);
0 unallowlisted violation(s), 4 allowlisted producer-only topic(s).
EXIT=0
```

Also fixed a now-dangling prose reference to this entry in the `proposal-events` reason below it.
`event_consumer_liveness` is **not** in the OPA read set (`grep -o 'data\.rules\.[a-z_]*'` over
every `.rego`: `authz`, `feature_flags`, `four_eyes`, `money_path_action_prefixes`,
`money_path_services`, `shared_m*`), so no `gen-rules-opa-data.py` regeneration and no bundle
restamp is triggered. No `.rego` touched, no `role_action_matrix` line added.

## An existing test encoded the un-delegated behaviour, and had to change

`CustomerEdgeCardsTest."a card owned by another party is refused before any limit change"` stubbed
only the card GET on a **strict** mockk, so it asserted not just "a stranger gets 403" but
implicitly "the delegation question is never asked" — the defect itself, pinned as the contract.
It now stubs `/delegation/check` → `{"authorized":false}`, which preserves its real meaning (a
stranger is still refused) and strengthens it. This is called out because replacing such a test is
part of the fix, not incidental churn.

## Both test runs

Run **against `origin/main`'s `CustomerEdgeResource.kt`** with the new tests in place
(`git checkout origin/main -- <the one main source file>`, everything else this branch):

```
DelegatedCardAccessTest > a delegate holding CARD_MANAGE_LIMITS may lower the limits ...  FAILED
DelegatedCardAccessTest > a stranger card-issuance refuses is still 403 ...               FAILED
DelegatedCardAccessTest > a card-issuance that is down denies rather than guessing        FAILED
DelegatedCardAccessTest > a delegated limit INCREASE still needs SCA ...                  FAILED
DelegatedCardAccessTest > a shared card appears in the list, marked, ...                  FAILED
DelegatedCardAccessTest > an offered but unaccepted grant does not put the card in list   FAILED
DelegatedCardAccessTest > a grant without CARD_VIEW does not put the card in the list     FAILED
32 tests completed, 7 failed
BUILD FAILED
```

7 of 9 fail. The deny-direction tests would otherwise pass on `main` **vacuously** — `main` answers
403 and lists nothing whatever the grant says — so each of them also asserts that the question *was
asked*, which is the part `main` structurally cannot do. The 2 that pass both ways are controls on
purpose: *the holder is never asked about delegation* and *own cards survive a delegation-service
outage* must hold before and after.

Run **on this branch** (full module gate, reading which tasks actually appeared):

```
> Task :openbank-customer-edge:detekt
> Task :openbank-customer-edge:ktlintMainSourceSetCheck
> Task :openbank-customer-edge:ktlintTestSourceSetCheck
> Task :openbank-customer-edge:ktlintCheck
> Task :openbank-customer-edge:test
BUILD SUCCESSFUL
> Task :openbank-customer-edge:quarkusBuild
BUILD SUCCESSFUL
```

JUnit XML totals, read from the files rather than the console:
`tests=359 failures=0 errors=0 skipped=1`. The single skip is the pre-existing broker-gated
`CustomerEdgeContractBrokerProviderVerificationTest` (`@EnabledIfSystemProperty(pactbroker.url)`),
unchanged by this PR. `DelegatedCardAccessTest`: `tests 9 fail 0 err 0 skip 0`.

## Governance

- **Not money-path.** Neither `openbank-customer-edge` nor `openbank-card-issuance-service` is in
  `rules.yaml: money_path_services` (checked against the live list). No 2-approval rule, no
  mandatory threat model — the customer-edge threat model is updated anyway, since this changes an
  authorization decision.
- **API contract:** `openapi.yaml` `info.version` 1.34.0 → **1.35.0** (additive: a new optional
  `sharedWithMe` on `Card`, widened `listCards` and `updateCardLimits` semantics). Re-read off live
  `origin/main` after pushing — still 1.34.0 there, so 1.35.0 is free.
- `version.txt` deliberately **not** hand-edited; release-please owns it.
- No DB migration, no event schema change, no `.rego`, no OPA bundle, no `application.yaml` key.

## What is left on #2990, with counts

Deliberately **not** in this PR — each is a decision, not plumbing, and inventing an answer is
worse than leaving the issue open:

1. **`AccountResource`'s own customer surface: 10 `denyIfNotOwner` sites, unchanged.** Delegated
   account reads are enforced one tier up, at the edge (`mayReadAccount`, live since #3633). AC7 as
   *written* ("the delegate calls `GET /api/v1/accounts/{id}/balance` and gets 200") is still false
   at the service. The decision is whether the edge is accepted as the enforcement point — in which
   case AC7 should be reworded — or whether in-service parity is required, which is a money-path
   change to a money-path service and needs its own PR and 2 approvals.
2. **Card freeze / unfreeze / block / controls: 4 `ownsCard` sites, unchanged.** ADR-0249 D2 says a
   grantee should be able to freeze and unfreeze, but **no capability maps to freeze**: the
   vocabulary has `CARD_VIEW` and `CARD_MANAGE_LIMITS` only, and `CardDelegationIntent` has `VIEW`
   and `MANAGE_LIMITS`. Choosing which one covers freeze — or adding a third — is a vocabulary
   decision spanning delegation-service, card-issuance and the edge. `MANAGE_LIMITS` is the
   tempting answer and it is wrong: freeze/unfreeze is not a limit, and unfreeze *restores*
   capability, which is the direction that needs the explicit grant.
3. **`SavingsGoalDelegationGuard` beyond the proposal flow: unchanged.** Same shape as the card
   guard was; not touched here so this PR stays one reviewable wiring.
4. **AC5 cumulative spend counters: still no writer.** ADR-0249 D3 supersedes ADR-0232 D3's
   assignment — the counter belongs in delegation-service (one grant sees all rails), not in a
   product service. Recommend striking AC5 from this issue rather than tracking it as debt here.

`#2990` stays **open** on items 1–4.


JiRaska:
Closing: the user-visible outcome shipped on `main` while this was open, by a different route.

`main` now has both halves — `sharedWithMe` on `GET /cards` (5 references) and a
`CARD_MANAGE_LIMITS` check on the limits path (`CustomerEdgeResource.kt:3367`), landed as ADR-0249
D2. Re-implementing them here would be duplicate work, so this PR is withdrawn rather than
reconciled: 4 real conflicts in `CustomerEdgeResource.kt` and 2 in `openapi.yaml`, both sides
describing the same behaviour in different words.

**But the thing #2990 actually asked for is still not done, and is now harder.** The two routes are
not interchangeable:

| | this PR | `main` today |
|---|---|---|
| who decides | card-issuance's `CardDelegationGuard`, via `GET /{id}/delegation/check` | customer-edge itself, via `hasGrant` → `delegation-service /api/v1/delegations/check` |
| `CardDelegationGuard` production callers | 1 | **0** |

Measured on `origin/main`: `git grep -l CardDelegationGuard -- '*/src/main'` returns **nothing**. The
projection ADR-0232 D3 exists for — the one card-issuance owns, that knows about card state — is
still called by no production path, and the edge now answers the delegation question itself against
delegation-service.

That is a real architectural fork, not a tidy-up: one says the resource-owning service adjudicates
its own delegation, the other says the edge does. Both work today; only one should survive, and
which one is a decision rather than a merge conflict.

I have written that up on #2990 rather than picking a side here. Nothing is lost by closing this —
the behaviour is live, and the open question is recorded where it belongs.

PR #4684, opened 2026-08-13
fix(admin-ui): raise the nanoid override past CVE-2026-67213

JiRaska:
## The pin that was the fix is now the defect

`Trivy` is failing on PRs with:

```
nanoid  CVE-2026-67213  HIGH  fixed  3.3.17  ->  3.3.18, 5.1.6
```

This is not caused by any of those PRs. `nanoid@3.3.17` is on `main`, and it is held there **deliberately**: `openbank-admin-ui/package.json` already carries `overrides.nanoid: "3.3.17"`. The only package that wants it is `postcss`, which asks for `^3.3.17` — so without the override npm would already resolve the patched version on its own. The pin, presumably added to force a floor at some earlier point, is now the single thing keeping the vulnerable version in the tree.

Bumped the override to `3.3.18` and regenerated the lockfile with `--package-lock-only`.

### Scope

Two files, five lockfile lines:

```
 openbank-admin-ui/package-lock.json | 10 +++++-----
 openbank-admin-ui/package.json      |  2 +-
```

`package.json`'s `version` field is untouched, so the `version.txt == package.json` sync gate is unaffected. No `--write-verification-metadata`, no unrelated dependency churn.

### Why this is worth landing ahead of the queue

Every PR that runs Trivy fails on it, regardless of what that PR changes — it reads as "your change introduced a HIGH", which is the most expensive kind of false signal to leave in a queue that several agents are working. Confirmed the version on live `main` rather than on a branch base.

### On verification — local Trivy cannot settle this, and it took a control to find that out

`trivy` is installed here and reports **0 vulnerabilities** on the fixed tree. That result is worth nothing on its own, and I nearly published it as if it were evidence.

Run against the **known-positive** — `main`'s own `package.json` + `package-lock.json`, the exact pair CI fails on — local Trivy also reports **0**. So its vulnerability DB does not carry CVE-2026-67213 (published too recently), and `--download-db-only` did not change that. A scanner whose DB lacks the advisory returns "clean" for the vulnerable and the patched tree alike; the two are indistinguishable locally.

What actually supports this change, then:
- the resolved version in the regenerated lockfile (`node_modules/nanoid` -> `3.3.18`), which is the version the advisory names as fixed;
- the CI run on this PR, which is the only oracle that currently knows the CVE exists.

### DIFF
--- a/openbank-admin-ui/package-lock.json
+++ b/openbank-admin-ui/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "openbank-admin-ui",
-  "version": "0.107.0",
+  "version": "0.117.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "openbank-admin-ui",
-      "version": "0.107.0",
+      "version": "0.117.0",
       "dependencies": {
         "@hookform/resolvers": "5.7.1",
         "@radix-ui/react-dialog": "1.1.23",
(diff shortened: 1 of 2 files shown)


JiRaska:
Superseded — `main` already carries the fix, landed independently while this was open.

Verified on live `origin/main`, both halves:

```
package.json  overrides -> {'postcss': '8.5.26', 'sharp': '0.35.0', 'nanoid': '3.3.18'}
package-lock  node_modules/nanoid -> 3.3.18
```

So CVE-2026-67213 is closed on `main` and this PR has nothing left to contribute. Its own CI did confirm the fix works (Trivy green, 0 failing) before `main` moved — which is the only place that verdict could come from, since local Trivy's DB does not yet carry the advisory and returns "clean" for the vulnerable and patched trees alike.

Closing rather than merge-resolving: diffed against current `main` first, and this branch would add **93 lockfile lines main does not have** — my `npm install --package-lock-only` regenerated more than the change needed. Taking it now would be churn on top of an already-correct file, not a fix.

Left here for whoever hits this next: the vulnerable version was held by an explicit `overrides.nanoid` pin, not by resolution. `postcss` asks for `^3.3.17`, so npm takes the patched version unaided — the pin was the only thing keeping it back.

PR #4751, opened 2026-08-14
feat(party): declare materiality on PARTY_UPDATED events (ADR-0256 D1)

### LINKED ISSUE #4458: party-service: publish materiality-classified party change event (ADR-0256 followup)

JiRaska:
## What

ADR-0256 D1 ships its trigger catalogue with `PARTY_DATA_CHANGED` **deferred**: kyc-service cannot consume it until party-service declares materiality in its own event contract. This lands that contract.

## The classification

`PartyChange.classify(before, after)` compares a closed field set and returns one of three outcomes:

| outcome | fields | KYC re-screening trigger |
|---|---|---|
| `MATERIAL` | `legalName`, `dateOfBirth`, `nationality`, `partyType` | yes |
| `NON_MATERIAL` | `email`, `phone`, `address`, `tradingName`, `discoverable`, `consentMarketing` | no |
| `NO_CHANGE` | nothing moved | no |

Three values, not a boolean, for `PushSendOutcome.ACCEPTED / SKIPPED / FAILED`'s reason: a no-op PATCH and an edit KYC does not care about must not share a count — in the return type or in `openbank.parties.change.classified{materiality=...}`.

Two deliberate exclusions, both documented in the classifier's KDoc:

- **Beneficial-owner set** — ADR-0256 names it and party-service does not model beneficial owners at all. Adding the key with nothing behind it would be a classification that can never fire while reading as covered.
- **`status` / `kycStatus` / `amlStatus`** — lifecycle state with its own events. If a KYC verdict projected onto the party record could classify itself as a material party change, kyc-service would re-trigger on its own output.

## The serialized shape

`PARTY_UPDATED` gains two additive keys; every other event type is byte-identical and asserts that it declares no materiality.

```json
{"eventType":"PARTY_UPDATED","partyId":"…","partyType":"INDIVIDUAL","status":"ACTIVE",
 "kycStatus":"APPROVED","legalName":"Jan Novotný","email":"…","occurredAt":"2026-08-14T08:00:00Z",
 "actorId":"system:party-service:party-api","actorType":"SYSTEM",
 "materiality":"MATERIAL","materialFields":["legalName"]}
```

## Why the API surface moved

`updateParty` now accepts `legalName`, `dateOfBirth`, `nationality`. Without them the `MATERIAL` branch is **unreachable in production** — the only editable fields were contact details, so every event this service could emit would be `NON_MATERIAL` while the contract advertised a trigger. `openapi.yaml` already documented `legalName` on `UpdatePartyRequest` and the handler silently ignored it; the spec is now true. `info.version` 1.14.0 → 1.15.0 (additive), read off live `origin/main`. No DB change, no migration.

## How the shape is proved

- `PartyChangeMaterialityContractTest` asserts the **serialized body**, never the classification object — sca-service put `eventType` in `sca_outbox.event_type` and not in the payload, its producer test asserted the column, and the consumer's parser read `""` for 15 `SENT` events. Both boundary directions are covered: a test that only proves a name change is MATERIAL passes against a classifier that returns MATERIAL unconditionally.
- `PartyOutboxWriteIT` re-proves it against the **real producer**: RestAssured PATCH → JDBC read of `party_outbox.payload`, MATERIAL and NON_MATERIAL cases. 4 tests, all green.
- `occurredAt` (never `timestamp`), asserted to end in `Z` and to be *recent*, not non-null. This caught a real trap while writing it: a test-local `ObjectMapper().findAndRegisterModules()` leaves `WRITE_DATES_AS_TIMESTAMPS` on and renders `occurredAt` as `1.7866E9`, which `AuditConsumer.eventTime()`'s `Instant.parse` rejects — the test would have pinned a wire shape no consumer receives. The mapper now mirrors the Quarkus default and the IT re-asserts `Z` on the real bytes.

## Contract coverage — stated plainly

`openbank.party.events` stays **grandfathered** in `.github/event-contract-baseline.txt`; the new fields are documented in `docs/asyncapi/openbank-events.yaml` (1.0.0 → 1.1.0), which no gate validates.

A per-service `openbank-contracts/openbank-party-service/asyncapi.yaml` is **not** added on purpose: `check-event-contract-code-agreement.py` resolves event types from `override val eventType` / `const val EVENT_TYPE` and payload properties from a data class's primary constructor. Party's envelope is a hand-built `LinkedHashMap` with inline string literals, so a contract file would be structurally unverifiable by that gate — the estate blind spot, not a fix for it. Making party detectable means restructuring the envelope builders, which is its own change.

Gates run locally, all green: `check-event-contract-coverage.py`, `check-event-contract-code-agreement.py`, `check-event-schema-compat.py --base origin/main`, `check-domain-event-occuredat.sh`.

## Verification

`./gradlew :openbank-party-service:build` — BUILD SUCCESSFUL, `quarkusBuild` included (CDI/ArC wiring of the new `PartyChangeMetricsPort` adapter surfaces only there). JUnit XML: **161 tests, 0 failures, 1 skipped** — the skip is the pre-existing broker-gated `PartyEventPactProviderVerificationTest`. `ktlintCheck` and `detekt` re-run separately afterwards and appear in the task list.

party-service is **not** in `rules.yaml: money_path_services`, so the 2-approval + threat-model rule does not apply.

Closes #4458

### DIFF
--- a/docs/asyncapi/openbank-events.yaml
+++ b/docs/asyncapi/openbank-events.yaml
@@ -3,7 +3,7 @@
 asyncapi: 3.0.0
 info:
   title: OpenBank Async Event API
-  version: 1.0.0
+  version: 1.1.0
   description: |
     Kafka event streams across all OpenBank services.
     All topics follow the naming convention: openbank.{domain}.{aggregate}.{event}
@@ -616,6 +616,25 @@ components:
               type: string
             legalName:
               type: string
+            materiality:
+              type: string
+              enum: [MATERIAL, NON_MATERIAL, NO_CHANGE]
+              description: >-
+                PARTY_UPDATED only. Publisher-declared materiality of the master-data change
+                (ADR-0256 D1, issue #4458) — never inferred by a consumer from the payload.
+                MATERIAL is the only value that is a KYC re-screening trigger
+                (PARTY_DATA_CHANGED); NON_MATERIAL covers contact details, trading name, address
+                and preferences; NO_CHANGE is an update that moved no compared field and is
+                deliberately NOT folded into NON_MATERIAL. Additive since party-service
+                openapi 1.15.0 — absent on events published before it, and on every event type
+                other than PARTY_UPDATED.
+            materialFields:
+              type: array
+              items:
+                type: string
+              description: >-
+                The changed fields that made the event MATERIAL (subset of legalName,
+                dateOfBirth, nationality, partyType). Empty for NON_MATERIAL and NO_CHANGE.
 
     ConsentEventPayload:
       allOf:
--- a/openbank-party-service/src/main/kotlin/com/openbank/party/application/port/in/PartyPort.kt
+++ b/openbank-party-service/src/main/kotlin/com/openbank/party/application/port/in/PartyPort.kt
@@ -70,12 +70,25 @@ data class MergePartyCommand(
     val approvalReference: String?,
 )
 
+/**
+ * A PATCH of a party record: every field is optional and null means "leave as it is".
+ *
+ * [legalName], [dateOfBirth] and [nationality] are the MATERIAL master-data fields (ADR-0256 D1,
+ * #4458). They are here — and were not before — because the materiality classification on
+ * `PARTY_UPDATED` is otherwise a branch nothing can reach: the only editable fields were contact
+ * details, so every event this service could ever emit would be `NON_MATERIAL` while the contract
+ * advertised a trigger. `legalName` was already documented in `openapi.yaml` and silently ignored
+ * by the handler; this makes the spec true.
+ */
 data class UpdatePartyCommand(
     val id: UUID,
     val email: String?,
     val phone: String?,
     val address: Address?,
     val tradingName: String?,
+    val legalName: String? = null,
+    val dateOfBirth: String? = null,
+    val nationality: String? = null,
 )
 
 /**
--- a/openbank-party-service/src/main/kotlin/com/openbank/party/application/port/out/PartyPorts.kt
+++ b/openbank-party-service/src/main/kotlin/com/openbank/party/application/port/out/PartyPorts.kt
@@ -5,6 +5,7 @@
 package com.openbank.party.application.port.out
 
 import com.openbank.party.domain.model.Party
+import com.openbank.party.domain.model.PartyChangeMateriality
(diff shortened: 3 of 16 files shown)


JiRaska:
**Version collision with the other party-service PR — whichever lands second must re-bump.**

#4751 and #4753 both move `openbank-party-service/src/main/resources/openapi.yaml` `info.version` to **1.15.0**. Both pass their own api-contract gate, because each classifies against a base where 1.15.0 was still free.

The line merges **textually clean** — git sees identical text, not a taken version — so nothing will conflict and nothing will go red. The second PR simply lands with new endpoints under a version the first one already published.

After the first of the two merges, the second needs:

```bash
git fetch origin main && git show origin/main:openbank-party-service/src/main/resources/openapi.yaml | grep -m1 '^  version:'
```

and a re-bump to the next free minor. Read it off **live `main`**, not off this branch's base — the gate resolves the current merge-base, but a run that predates the competing merge is still stale, and that residual race is the one no base resolution can close.

Flagged from the autonomous queue; not merging either.

Issue #1984, opened 2026-07-23
Party merge (ADR-0179): remaining follow-ons


JiRaska:
## Context

The party-merge remediation path shipped in ADR-0179 (closes the original #1782 "duplicate identities have no remediation path"): `PartyStatus.MERGED` + `merged_into` (V15), `POST /api/v1/parties/{id}/merge` with fail-closed preconditions + audit, the `PARTY_MERGED` event, and an account-ownership guard so a **funded** duplicate cannot be retired. This issue tracks the remaining follow-ons that make the merge complete end-to-end.

## Tracking checklist

- [x] **Balance sweep for a funded duplicate** — **DONE, not as a ledger `ADJUSTMENT` journal.** That design was considered and deliberately rejected (`TransactionResource.kt:210-214`: `PaymentJournalFactory` never reads `transaction.type`, so an `ADJUSTMENT`-typed posting would be byte-identical to a customer payment). What shipped is **`POST /api/v1/transactions/merge-sweep`** in transaction-service — `MERGE-SWEEP`-prefixed description as the ledger's only distinguishing mark, `mergeReference` tying the posting to the merge, gated by the `sweep` verb in `rules.yaml: four_eyes`. A funded duplicate *can* be retired today. party-service's 409 still named the non-existent ledger path; corrected in #4753.
- [ ] **Consumer adoption of `merged_into` per service holding a `partyId`** (fleet sweep). Until each consumer follows `merged_into`, a stale id still resolves to the retired row. Enumerate every service that stores/keys on `partyId` and have it chase the pointer (or reject with a redirect to the survivor). *(1 of N: `openbank-customer-edge`'s `PartyMergeResolver`, #3901.)*
[...]

PR #4753, opened 2026-08-14
fix(party): point the merge refusal at the real balance-sweep endpoint

JiRaska:
## What

The 409 an operator hits when merging a **funded** duplicate party tells them to sweep the balances *"via a ledger `ADJUSTMENT` journal entry"*. No such producer exists anywhere in the fleet — `git grep ADJUSTMENT -- 'openbank-ledger-service/src/main/*'` returns zero hits (control: the same pathspec on `Journal` hits `LedgerPorts.kt`, so the empty result is absence, not a mis-globbed pathspec).

The path was considered and **deliberately rejected**. `TransactionResource.mergeSweep`'s KDoc records why: passing `type = ADJUSTMENT` to the ordinary endpoint would produce a posting byte-identical to a customer payment, because `PaymentJournalFactory` never reads `transaction.type`.

What actually shipped is **`POST /api/v1/transactions/merge-sweep`** in transaction-service — gated by the `sweep` verb in `rules.yaml: four_eyes`, minting a `MERGE-SWEEP`-prefixed description as the only thing distinguishing the correction in the trial balance, on a statement, and to an auditor.

So the operator-facing guidance names a mechanism with no producer, while the mechanism that does exist goes unmentioned. An operator following the message finds nothing.

## Changes

All five operator-facing references corrected:

| file:line | what |
|---|---|
| `PartyResource.kt:329` | the 409 refusal message |
| `PartyResource.kt:595` | `MergePartyRequest.approvalReference` KDoc |
| `PartyPort.kt:62` | `MergePartyCommand` KDoc |
| `openapi.yaml:160` | `POST /{id}/merge` description |
| `openapi.yaml:655` | `approvalReference` schema description |

`approvalReference` is now tied to the `mergeReference` that `merge-sweep` already takes — that field is what links the money movement and the identity retirement from either end.

## Scope

Docs/contract text only, **no behaviour change**. `openbank-party-service` is **not** in `rules.yaml: money_path_services`, so this is not a 2-approval change.

`openapi.yaml` `info.version` 1.14.0 -> 1.14.1 — description-only, non-breaking. Re-read off live `origin/main` immediately before bumping, not off the branch base.

## Verification

- `./gradlew :openbank-party-service:ktlintCheck :openbank-party-service:detekt` — both tasks appear in the output and pass.
- `./gradlew :openbank-party-service:test` — read from the JUnit XML, not the console: **151 tests, 0 failures/errors, 1 skipped**. The single skip is `PartyEventPactProviderVerificationTest` (broker-gated on `pactbroker.url`), pre-existing and unrelated.
- No residual references: `git grep 'ledger ADJUSTMENT' -- 'openbank-party-service/*'` is empty.

Refs #1984 (checklist item 1 — this PR does not close it; see the issue comment for the full per-item measurement).

### DIFF
--- a/openbank-party-service/src/main/kotlin/com/openbank/party/application/port/in/PartyPort.kt
+++ b/openbank-party-service/src/main/kotlin/com/openbank/party/application/port/in/PartyPort.kt
@@ -59,9 +59,10 @@ data class CreatePartyCommand(
 /**
  * ADR-0179: retire [id] as a duplicate of [mergedIntoPartyId]. Both parties must be live, distinct,
  * and the target must not itself be merged (the caller resolves chains to the final survivor).
- * [reason] is free text for the audit trail; [approvalReference] links the ledger ADJUSTMENT
- * journal entries that swept the balances, so the money movement and the identity retirement are
- * traceable to one another.
+ * [reason] is free text for the audit trail; [approvalReference] is the `mergeReference` passed to
+ * POST /api/v1/transactions/merge-sweep (transaction-service), which posts the balance sweep as an
+ * ADJUSTMENT transaction, so the money movement and the identity retirement are traceable to one
+ * another from either end.
  */
 data class MergePartyCommand(
     val id: UUID,
--- a/openbank-party-service/src/main/kotlin/com/openbank/party/infrastructure/rest/PartyResource.kt
+++ b/openbank-party-service/src/main/kotlin/com/openbank/party/infrastructure/rest/PartyResource.kt
@@ -326,7 +326,8 @@ class PartyResource {
         summary = "Merge a duplicate party into a surviving party (ADR-0179)",
         description = "Retires {id} as a duplicate, preserving all PII and history. NOT erasure: " +
             "no anonymization, no PARTY_ERASED event. Refuses while the duplicate still owns an " +
-            "open account — sweep the balances via a ledger ADJUSTMENT journal entry and close " +
+            "open account — sweep the balances via POST /api/v1/transactions/merge-sweep " +
+            "(transaction-service, ADR-0179) and close " +
             "the account first. Four-eyes gated (ADR-0155): when four-eyes enforcement is on, the " +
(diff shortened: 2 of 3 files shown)


JiRaska:
## ⚠️ Spec version collision with #4751 — whoever lands second must re-bump

#4751 (`feat(party): declare materiality on PARTY_UPDATED events`) touches the **same** `openbank-party-service/src/main/resources/openapi.yaml` and claims the **same** `info.version`:

```
$ git show refs/pull/4751/head:openbank-party-service/src/main/resources/openapi.yaml | grep -m1 '^  version:'
  version: 1.15.0
$ git show refs/pull/4753/head:openbank-party-service/src/main/resources/openapi.yaml | grep -m1 '^  version:'
  version: 1.15.0
```

Both will pass the api-contract gate independently — it classifies against each PR's own merge-base, so neither can see the other's claim. And the version line will merge **textually clean**, because git sees identical text, not a taken number.

So: whichever of the two merges second needs `info.version` re-read off live `origin/main` and re-bumped (1.16.0) before merge. Given #4751 is the `feat` and was opened first, the likely order is #4751 then this one — in which case **this PR** is the one to re-bump. I will not merge either.

No action needed on #4751.

JiRaska:
**Version collision with the other party-service PR — whichever lands second must re-bump.**

#4751 and #4753 both move `openbank-party-service/src/main/resources/openapi.yaml` `info.version` to **1.15.0**. Both pass their own api-contract gate, because each classifies against a base where 1.15.0 was still free.

The line merges **textually clean** — git sees identical text, not a taken version — so nothing will conflict and nothing will go red. The second PR simply lands with new endpoints under a version the first one already published.

After the first of the two merges, the second needs:

```bash
git fetch origin main && git show origin/main:openbank-party-service/src/main/resources/openapi.yaml | grep -m1 '^  version:'
```

and a re-bump to the next free minor. Read it off **live `main`**, not off this branch's base — the gate resolves the current merge-base, but a run that predates the competing merge is still stale, and that residual race is the one no base resolution can close.

Flagged from the autonomous queue; not merging either.

PR #4770, opened 2026-08-14
feat(engagement): prepare campaign story placements

JiRaska:
## Summary
- render campaign placements on the existing Stories surface
- map Story placements to the reviewed Story template and surface-content type
- cover the campaign command consumer end to end

## Rollout
This is the expand step only. Campaign Service and Admin UI do not emit or expose STORIES until this service version is deployed, avoiding dropped placement commands during a mixed-version rollout.

## Verification
- CampaignBannerPlacementConsumerTest
- engagement-service ktlintCheck and detekt

Refs #4476


JiRaska:
Review finding addressed by scope: this PR deliberately contains only the consumer-first expand step. It exposes no campaign template, API enum or marketer UI, so it cannot emit a Stories placement before the updated engagement consumer is deployed. The follow-up producer/UI PR is held until this rollout has completed.

Issue #4480, opened 2026-08-12
feat(campaign): attribute in-app engagement to campaign journeys


JiRaska:
## Goal

Enable an honest cross-surface Campaign Management funnel by attributing consent-safe in-app events to the campaign and journey step that selected the content. This unblocks a marketer view of:

`eligible/enrolled → campaign handoff → delivery → in-app impression → interaction or dismiss → observed business conversion`

This follows #4475 and #4476. It must not present clicks as business conversions or infer campaign attribution from proximity in time.

## Current evidence and gap

- `campaign-service` exposes whole-campaign send summaries, per-step journey aggregates and observed banking-conversion experiments.
- `engagement-service` stores `IMPRESSION`, `CLICK`, `DISMISS` and `CONVERSION` for catalogue-backed in-app content.
- An engagement event currently contains `partyId`, `contentId`, `slot`, `type` and time — **not** campaign ID or step/order/correlation.

Therefore any campaign-level in-app funnel today would be fabricated. A common content card can be rendered independently of a campaign, and a click cannot be assigned safely to the most recent campaign message.

## Scope

1. Define a versioned correlation contract from campaign handoff to mobile deep link / surface rendering. It must carry an opaque campaign interaction reference, not PII and not a customer-controlled campaign ID.
2. Persist the reference at the narrowest responsible layer and have customer-edge inject/validate it before engagement-service records an event.
3. Add campaign-scoped, aggregate-only reporting by channel and surface:
   - handoff, delivery confirmation where available, impression, click, dismiss, observed conversion;
   - deduplicated/person semantics explicitly documented for each number;
   - separate `unknown/unattributed` from zero.
4. Expose the aggregate through campaign-service or a dedicated read model, then surface it in Campaign Studio with wording that distinguishes engagement from observed business conversion.
5. Add privacy, retention and abuse review: no PII in metrics responses; no operator can post an event for a different party; no client can create arbitrary attribution.

## Acceptance criteria

- A legitimate app interaction can be traced from campaign/step correlation reference to an aggregate metric without exposing a party identity in the operator API.
- An event without a valid correlation is retained as unattributed or rejected according to the documented contract; it never silently joins a campaign.
- UI shows `unattributed/unavailable` distinctly from `0`.
- Click, dismiss and impression are never labelled conversion. Conversion remains an observed product event with its own attribution constraints.
- Contract, service integration and customer-edge spoofing tests cover happy path, unknown reference and forged party/campaign attempts.
[...]

PR #4783, opened 2026-08-14
fix(campaign): project story engagement

JiRaska:
## Summary

Fix the campaign engagement projection so app Story impressions and clicks are admitted and appear in the campaign attention view.

## Root cause

`STORIES` is a closed, already-emittable in-app campaign surface, but the older PostgreSQL CHECK constraint for `campaign_engagement_event.surface` omitted it. The campaign consumer could therefore parse a valid Story event and then retry forever when its aggregate projection was rejected.

## Change and rollout safety

- Adds an expand-only Flyway migration that admits `STORIES`; existing values and readers remain valid.
- Does not remove history or require a destructive rollback. Application-code rollback remains safe while the widened constraint is present.
- Adds consumer coverage plus a Flyway-backed HTTP contract regression, so the schema and reporting route are exercised together.

## Validation

- `./gradlew :openbank-campaign-service:build` (all Campaign Service test XMLs green; 198 tests, 0 failures)
- `./gradlew :openbank-campaign-service:ktlintCheck :openbank-campaign-service:detekt`
- `git diff --check`

Refs #4480

### DIFF
--- a/openbank-campaign-service/src/main/resources/db/migration/V15__campaign_engagement_stories_surface.sql
+++ b/openbank-campaign-service/src/main/resources/db/migration/V15__campaign_engagement_stories_surface.sql
@@ -1,0 +1,21 @@
+-- SPDX-License-Identifier: Apache-2.0
+-- Copyright (c) OpenBank contributors. Licensed under the Apache License, Version 2.0.
+-- See LICENSE in the repository root or https://www.apache.org/licenses/LICENSE-2.0 for details.
+--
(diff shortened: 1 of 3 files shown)


JiRaska:
Release readiness update: required CI gates are green. This PR remains subject to the repository's independent review requirement; no self-approval or direct merge has been performed. Consumer-first GitOps promotion is already the desired image on main, so this producer slice remains ordered after the compatible engagement projection.

PR #4788, opened 2026-08-14
feat(admin-ui): add campaign delivery health

JiRaska:
## What
- adds portfolio delivery health from the existing server summary
- distinguishes sent, policy suppression and failed delivery decisions
- never estimates conversion or calls suppression a delivery failure

## Verification
- campaign console tests: 6 passed
- ESLint: 0 errors
- TypeScript: passed

Refs #4476

PR #4789, opened 2026-08-14
feat(admin-ui): show confirmed campaign outcomes

JiRaska:
Depends on #4788. Adds only confirmed CONVERTED outcomes from the server aggregate; no clicks, inferred conversion, or derived rate. Verified with campaign console tests and TypeScript.


JiRaska:
Rebased directly onto main after #4788. Confirmed-outcomes UI reads only existing `CONVERTED` outcome events; it does not infer conversion from clicks or delivery. Local: targeted Campaign Console tests 6/6 and `tsc --noEmit`; required GitHub gates are green.

PR #4781, opened 2026-08-14
feat(campaign): add explicit decision graph journeys

JiRaska:
## Outcome\n\nAdds a bounded, app-first campaign decision graph with explicit CONFIRMED / NOT_CONFIRMED edges, durable per-enrolment path evidence, and Studio authoring plus outcome rendering.\n\n## Safety\n\n- Existing linear definitions and Temporal histories remain on their versioned legacy path.\n- Graphs allow only forward edges, a bounded number of decision nodes, and observable delivery facts.\n- Migration is additive; rollback ignores the optional columns.\n- Maker/checker lifecycle is unchanged.\n\n## Evidence\n\n- Temporal workflow branch test\n- Real HTTP contract + persistence test\n- Admin UI type check and focused interaction/render tests\n\nCloses #4680

PR #4884, opened 2026-08-15
chore(campaign): enable graph activation

JiRaska:
## Summary
- Enables explicit decision-graph campaign activation after the isolated Temporal queue proved healthy in the live campaign-service rollout.
- Leaves the legacy queue untouched and retains the separately reviewed rollback boundary.

## Evidence
- campaign-service deployment is Available (1/1) on the graph-capable image.
- both `openbank-campaign` and `openbank-campaign-decision` Temporal pollers started.
- Flyway applied V13/V14 successfully; graph activation was still false during that verification.

## Validation
- `git diff --check origin/main...HEAD`
- `kubectl apply --dry-run=client --validate=false -f openbank-infra/gitops/components/campaign/campaign-service.yaml`

Refs #4781

### DIFF
--- a/openbank-infra/gitops/components/campaign/campaign-service.yaml
+++ b/openbank-infra/gitops/components/campaign/campaign-service.yaml
@@ -129,10 +129,10 @@ spec:
             # queue until the compatible image is restored.
             - name: OPENBANK_CAMPAIGN_DECISION_TASK_QUEUE
               value: openbank-campaign-decision
-            # Expand first: deploy and observe the isolated graph worker before a separately
-            # reviewed GitOps change enables graph-campaign activation.
+            # The isolated graph worker is live and healthy. This separately reviewed step admits
+            # graph-campaign activation without changing the replay-compatible legacy queue.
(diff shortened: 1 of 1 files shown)


JiRaska:
### Live pre-activation evidence (read-only)

Collected after the Flyway recovery rollout, while `OPENBANK_CAMPAIGN_EXPLICIT_GRAPH_ACTIVATION_ENABLED=false`:

- `campaign-service` is **Available 1/1** on the graph-capable image; the replacement pod is `2/2 Running` with zero restarts.
- Startup logs show both workflow and activity pollers started for `openbank-campaign-decision`, alongside the legacy queue.
- Temporal `task-queue describe` against the target namespace observed a current poller for the decision queue separately for the **workflow** and **activity** task types. Both identify the healthy campaign-service worker and were queried immediately before this PR.
- Flyway validated all 15 migrations and applied V13/V14; the application then reached its normal Quarkus startup message.

I intentionally did **not** inject a synthetic workflow with fabricated campaign/party IDs into the banking environment: that would create a retrying production workflow and is not a meaningful customer-safe smoke test. The live worker, workflow poller, and activity poller evidence above is the rollback-safe readiness gate for this activation change.

Issue #4533, opened 2026-08-13
Quarkus platform bump: ship hibernate-reactive-core >= 4.2.1.Final (CVE-2025-14969)


JiRaska:
`CVE-2025-14969` / `GHSA-frpp-8pwq-hjrx` (Hibernate Reactive: connection-pool exhaustion DoS when a
remote client prematurely closes an HTTP connection during a DB operation) is fixed upstream in
`org.hibernate.reactive:hibernate-reactive-core` **4.2.1.Final**.

The fleet resolves **3.4.2.Final**, bundled by `io.quarkus.platform:quarkus-bom:3.38.0`.

## Why this cannot be a `force()` in `openbank.dependency-vulnerability-pins.gradle.kts`

Measured on branch `security/vex-cve-2025-14969`:

- `force("org.hibernate.reactive:hibernate-reactive-core:4.2.1.Final")` **resolves** cleanly and
  `:openbank-mcp-service:quarkusBuild` **succeeds**.
- Every `@QuarkusTest` in that module then fails at boot:
  `Failed to start quarkus` caused by
  `java.lang.NoSuchMethodError: org.hibernate.sql.ast.tree.select.SelectStatement org.hibernate.loader.ast.internal.LoaderSelectBuilder.createSelect(...)`
  (140 tests, 1 failed). Hibernate Reactive 4.2.x requires a newer Hibernate ORM than Quarkus 3.38.0
  ships.
- Control: with the force reverted, `:openbank-mcp-service:test` and
  `:openbank-security-scanner:test` are green.

Same shape as the OpenTelemetry 1.62.0 case already documented in that pins file: the advisory's
patched version is binary-incompatible with the platform's own extension, so remediation is a
**Quarkus platform bump that ships hibernate-reactive-core >= 4.2.1.Final**, not an independent force.

## Note on the previously-cited tracker

44 component VEX docs cite issue #1446 as the remediation plan for this CVE. #1446 is **CLOSED** — it
landed the 3.37.2 -> 3.38.0 platform bump, which moved this artifact 3.4.1.Final -> 3.4.2.Final and did
not cross the fix boundary. Those 44 statements therefore point at a closed issue and a stale version;
they should be refreshed to point here (out of scope for #4443, which covers `mcp-service` and
`security-scanner`).

## Exit criteria

A Quarkus platform version whose BOM resolves `hibernate-reactive-core >= 4.2.1.Final`, adopted in
[...]

PR #4929, opened 2026-08-15
security(governance): require evidence for a VEX verdict on a one-sided advisory range

### LINKED ISSUE #4716: Advisory ranges written as a single '< X' interval misclassify dependencies with parallel release lines
## The shape

A security advisory that expresses its affected range as a single open interval — `< X` — silently misclassifies every dependency that maintains **parallel release lines**. Maven/semver ordering says `3.4.2 < 4.2.1`, so a 3.x release branched *after* the fix is swept in as affected, when it in fact carries the patch.

The fleet has now hit this twice:

- **hibernate-reactive** (#4533, PR #4707). `GHSA-frpp-8pwq-hjrx` states `< 4.2.1`. `4.2.1.Final` released 2025-12-21; `3.4.0.Final` released 2026-05-27, five months later, off a branch that already contained the fix. Our `3.4.2.Final` **has** the patch. The issue's stated exit criterion — ship `>= 4.2.1` — is not merely unmet, it is unsatisfiable: quarkus-bom 3.38.0 / 3.38.1 / 3.38.2 / 3.39.0.CR1 all resolve `3.4.2.Final`. Worse, forcing it would move onto a line built against ORM 7.2.0 while Quarkus 3.38.0 ships ORM 7.4.5 — trading a non-existent DoS for a guaranteed boot failure.
- **netty** — the 4.1.x-vs-4.2.x VEX statement was the same shape.

## Why it costs more than a wrong label

The failure is asymmetric and both directions are expensive:

- **False positive** (what happened here): 47 VEX statements sat at `affected`, an issue was filed with an impossible exit criterion, and the "obvious" remediation would have broken the boot. Anyone who acted on the advisory text alone would have shipped an outage.
- **False negative** is available too: a project whose *older* line is unpatched but whose version string sorts above the stated bound reads as safe.

## What actually settles it

Version arithmetic cannot. Bytecode can, and it is cheap:

1. Read the advisory's referenced fix commit and identify its observable effects.
2. `javap -p -c` the jar and check for them.
3. **Run the probe against a known-negative and a known-positive** — the last unfixed release and the first patched one. In #4533 the first attempt used `strings`, which fails on BSD and returned `0` for *every* jar including the known-fixed control; only the control row caught it.
4. Confirm the jar's `sha256` matches the value pinned in `gradle/verification-metadata.xml`, so the bytes examined are the bytes shipped.

## Proposal

A triage guard for VEX classification: when an advisory's range is a single `< X` interval **and** the resolved artifact's major differs from `X`'s major, refuse to auto-classify and require an explicit, evidenced statement. Parallel release lines are common enough (hibernate-reactive, netty, Jackson, Kotlin) that the default should be "prove it", not "sort it".

Whether that lives in the VEX tooling or as a checklist item in the advisory-triage workflow is the open decision. Filing it because the same shape has now cost two investigations and would have caused an outage on the second.

Refs #4533

JiRaska:
Closes #4716. Refs #4533.

## Where the guard went, and why

**A gate over the committed VEX statements** — `vex-open-interval-evidence`, running
`.github/scripts/check-vex-open-interval-evidence.py` in the `supplychain` shard, `mode: enforced`.

Not the VEX tooling. The issue proposes refusing to "auto-classify", and there is nothing to
intercept: `build-release-evidence.sh` writes **every** finding as `under_investigation` on
purpose — a machine never asserts `not_affected`, because that is a security claim requiring human
judgement (`openbank-libs/governance/vex/README.md`). The misclassification is not produced by
tooling; a person writes it into `openbank-libs/governance/vex/<component>.openvex.json`, and the
committed statement is the first and only artifact a gate can see before it becomes a
regulator-facing audit object.

Not a checklist. It needs a human to remember, and remembering is what produced both occurrences.

## What it checks

For every statement recording a **verdict** — `not_affected`, `fixed`, **and `affected`**; only
`under_investigation` is exempt — the prose is scanned for the version bounds it cites.

- A `<`/`<=` bound with a `>`/`>=` immediately before it is a **two-sided** range
  (`>=4.2.0, <4.2.16`). It names its release line, cannot sweep in a parallel one, and is out of
  scope.
- An **unpaired** `< X` is the defective shape. The statement must then carry one of:
  - **artifact evidence** — a sha256 in the text that also appears in
    `gradle/verification-metadata.xml`, so the bytes inspected are provably the bytes shipped;
  - **`"resolved_version"`** — an explicit field whose release line (`major.minor`) matches the
    bound's, making it ordinary sound arithmetic.

`major.minor`, not major: the issue proposed comparing majors, but netty's parallel lines are
4.1.x and 4.2.x, which share a major. The netty half of the issue's own evidence shows major is
too coarse.

`affected` is in scope on purpose. The #4533 harm *was* an `affected` verdict: 47 statements sat
there on the bound alone, behind an exit criterion no candidate `quarkus-bom` can satisfy, whose
"obvious" remediation would have replaced a non-existent DoS with a certain boot failure. A wrong
`affected` is not free. (It adds no failures today — all 49 `affected` statements in the corpus
cite no bounds.)

`resolved_version` is a declared field rather than something parsed out of the prose because
inferring "the resolved version" from a paragraph is guesswork on the **escape hatch** — the
direction where being wrong ships a wrong classification silently. The hibernate statement alone
names eight versions, four of them on the bound's own line as probe controls; a gate guessing
intent from that would have let the defective case through.

## Falsification — both directions, on the real corpus

Fixtures alone would not settle this, so the discrimination is measured against the 824 real
statements on `main`:

| direction | subject | result |
|---|---|---|
| **must flag** | the 47 real hibernate-reactive statements with their sha256 removed | **47 errors, runner `FAIL`, RC=1** — and exactly those 47 |
| **must not flag** | the other 777, netty's `>=4.2.0, <4.2.16` and `>=4.2.0.Final and <=4.2.15.Final` among them | **clean** |
| **must not flag** | `main` as it stands (hibernate verdict keeps its sha256, pinned once in `verification-metadata.xml`) | **PASS, SUBJECTS=824** |
| **must flag** | corpus emptied | **FAIL** via `min_subjects: 700` — an absent corpus does not read as clean |

A guard that only ever says "no" is unfalsified, and here "no" is the answer that ships a wrong
classification — so the 777 negatives carry as much weight as the 47 positives.

The `--self-test` drives twelve fixtures covering the same shapes plus the ones the real corpus
cannot supply: an unpinned sha256 (evidence that proves nothing), a cross-line `resolved_version`,
a same-line one, `under_investigation`, and a missing `verification-metadata.xml` failing closed
rather than accepting every digest as proven.

## The false-negative direction

Covered where a statement exists, and the coverage is not a second mechanism: the predicate is
symmetric. A resolved version that sorts *below* the bound while carrying the fix, and an older
unpatched line that sorts *above* it, are the same unsound inference from the same single open
interval — both cite an unpaired `< X` and both are caught. A self-test case asserts the
above-the-bound direction explicitly.

**What it cannot see:** a vulnerability for which no statement exists at all. If a scanner never
reports the finding — which is what the false-negative direction can produce upstream — there is
no subject in this corpus. This gate hardens the verdicts that get written; it does not discover
findings. Said plainly rather than implied.

## Verification

```
BASE_REF=main PR_DIFF_BASE=$(git merge-base HEAD origin/main) \
  python3 .github/scripts/run-gates.py --only vex-open-interval-evidence
```

`gate-runner-self-test`, `yamllint`, `advisory-gate-registration` and `gate-lifecycle-metadata`
all green locally; the new gate is the 8th of 148 to carry `rationale:` + `review_after:`.

Anything under `.github/` or `openbank-libs/governance/` is held by the protected-path guard —
expected here, and correct.

### DIFF
--- a/.github/gates/gates.yaml
+++ b/.github/gates/gates.yaml
@@ -955,6 +955,40 @@ gates:
     run: |
       python3 .github/scripts/check-advisory-gate-registration.py --expect-manifest
 
+  # A VEX verdict resting on a single `< X` advisory bound (issue #4716). An advisory that
+  # writes its affected range as one open interval misclassifies every dependency with
+  # PARALLEL RELEASE LINES: Maven ordering says `3.4.2 < 4.2.1`, so a 3.x release branched
+  # AFTER the fix reads as affected. hibernate-reactive did exactly that (#4533/#4707 —
+  # 4.2.1.Final on 2025-12-21, 3.4.0.Final five months LATER off a patched branch), and the
+  # "obvious" remediation of forcing >= 4.2.1 was both unsatisfiable on every candidate
+  # quarkus-bom and a guaranteed boot failure against a different Hibernate ORM.
+  #
+  # It sits HERE and not in the VEX tooling because there is nothing upstream to intercept:
+  # build-release-evidence.sh writes every finding as `under_investigation` by design — a
+  # machine never asserts not_affected. The classification is entirely human, written straight
+  # into openbank-libs/governance/vex/*.openvex.json, so the committed statement is the first
+  # and only artifact a gate can see. A checklist was the alternative and is weaker by
+  # construction: it needs someone to remember, and remembering is what produced both
+  # occurrences.
+  #
+  # Falsified in BOTH directions against the real corpus, not only fixtures: removing the
+  # sha256 from the 47 real hibernate-reactive statements reddens exactly those 47, and the
+  # 777 others — netty's `>=4.2.0, <4.2.16` two-sided citations included — stay clean, which
+  # is the direction that matters, because a guard that only ever says "no" would make the
+  # wrong classification the shipping one.
+  - id: vex-open-interval-evidence
+    name: "a VEX verdict on a one-sided `< X` advisory range carries evidence (issue #4716, enforced)"
+    group: supplychain
+    mode: enforced
+    rationale: "A single `< X` advisory interval cannot classify a dependency with parallel release lines, and both directions are expensive: #4533 sat 47 statements at `affected` behind an exit criterion no platform can satisfy, whose remediation would have broken the boot, while the mirror case — an unpatched older line whose version string sorts above the bound — reads as safe."
+    review_after: "2027-02-15"
+    min_subjects: 700   # 824 OpenVEX statements today (2026-08-15)
+    selftest: |
+      python3 .github/scripts/check-vex-open-interval-evidence.py --self-test
+    selftest_expect: pass
+    run: |
+      python3 .github/scripts/check-vex-open-interval-evidence.py --root .
+
   # ==== registry =======================================================
   # issue #669 scope item 3 / rules.yaml: slo. Every money_path_services entry needs a
   # matching availability + latency Pyrra ServiceLevelObjective whose target/window match
--- a/.github/scripts/check-vex-open-interval-evidence.py
+++ b/.github/scripts/check-vex-open-interval-evidence.py
@@ -1,0 +1,382 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) OpenBank contributors. Licensed under the Apache License, Version 2.0.
+"""A VEX verdict that rests on a SINGLE `< X` advisory bound must carry artifact evidence.
+
+WHY THIS EXISTS (issue #4716, prompted by #4533 / PR #4707)
+-----------------------------------------------------------
+An advisory that expresses its affected range as one open interval — `< X`, no lower bound —
+silently misclassifies every dependency that maintains PARALLEL RELEASE LINES. Maven/semver
+ordering says `3.4.2 < 4.2.1`, so a 3.x release branched AFTER the fix sorts below the bound and
+reads as affected. It is not.
+
+Measured, twice:
+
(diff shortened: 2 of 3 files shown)


JiRaska:
#4984 merged (2026-08-16T22:59:05Z) and closes #4716. This PR now conflicts against what landed — recommend closing as superseded rather than resolving the conflict, since #4984 already implements the same guard (with a documented, deliberate major-only granularity tradeoff vs this PR's major.minor). Not closing this myself — that's your call.

JiRaska:
Closing as superseded by #4984, which already merged into `main`.

Both PRs solve the identical problem (#4716 — a one-sided `< X` advisory bound cannot classify a dependency with parallel release lines) with near-identical evidence and reasoning (same hibernate-reactive #4533/#4707 and netty examples, same sha256-anchored bytecode-evidence requirement). #4984's `check-vex-range-reasoning.py` (gate `vex-range-reasoning`) is the version that landed — it adds a `--classify` CLI, `min_subjects`/`budget_seconds` gate metadata, and covers 824 verdict statements across 52 overlays.

Rebasing this PR would put two gates enforcing overlapping logic in `Validate manifests` on every PR, which is the exact `#3009`/`#2984` duplicate-guard shape this repo's own CLAUDE.md warns about. No unique content here that #4984 doesn't already cover — verified via a direct diff of both scripts' headers and both gate declarations.

Issue #4701, opened 2026-08-13
billing_outbox has 2 DEAD rows at attempt_count 10: events written atomically, never published


JiRaska:
## What

`billing_outbox` holds **2 rows in `DEAD` at `attempt_count = 10`** — events that were written correctly and then never published. Found while measuring #4007; it is a *dispatch* failure, not a write failure, so it sits outside that issue's scope and nothing else tracks it.

Sibling of #4005 (closed), which was the same shape on a different service.

## Why it is worth its own issue

The write side is correct and provably so: `BillingAssessmentRepositoryImpl.kt:91` constructs `BillingOutboxEntity().apply{}` inside the same `sf.withTransaction` as the assessment, so the row and the assessment commit together. That is the transactional-outbox guarantee working. What failed is everything after — the dispatcher exhausted its retries and parked the row.

So the affected `billing.fee.post-intent.v1` events exist in the database and do not exist on the wire. Anything downstream that bills from them is short by exactly those two, and no error is being raised now: `DEAD` is a terminal resting state, not an alerting one.

## Re-check

```sql
SELECT id, event_type, status, attempt_count, created_at, last_attempt_at, last_error
FROM billing_outbox
WHERE status = 'DEAD'
ORDER BY created_at;
```

Measured 2026-08-13 in the sandbox. `openbank.outbox.dispatch-enabled` **is** `true` for this service, so the "switch was never flipped" explanation is already ruled out — the dispatcher ran and gave up.

## What needs deciding

1. **Why they died.** `last_error` on those rows is the first thing to read; it decides whether this is a transient broker problem that retry exhaustion mishandled, or a permanent payload/serialization problem that will kill every future row of the same shape.
2. **Whether `DEAD` is alertable.** Right now a row can reach a terminal unpublished state in silence. A count of `DEAD` rows per service is the obvious signal, and it is the *success*-state kind of alert this repo keeps finding it needs — there is no error rate to watch, because the failure already finished failing.
3. **Whether these two should be replayed.** That is a data decision, not a code one.

## Not in scope

The `#4007` work itself. That issue is about outboxes with no writer; billing has one and was mis-classified by a gate blind to the JPA-entity write idiom (corrected in #4698).

JiRaska:
Two PRs converged on this from different angles — worth reconciling before closing.

**#5642** (`fix(billing): escape billing_outbox JSON payloads via Jackson`) is the actual fix for
[...]

PR #4932, opened 2026-08-15
fix(billing): alert on dead-lettered outbox rows

JiRaska:
Refs #4701. Stacked on #4308 — it carries `AbstractOutboxDeadLetterGauge` /
`DomainMetrics.registerOutboxDeadLettered`, which this binds. **Base is
`fix/card-issuance-dead-outbox`, not `main`**; it retargets to `main` once #4308 lands.

`openbank-billing-service` **is** in `rules.yaml: money_path_services`.

## 1. Why they died — read from `last_error`, not from the shape

Re-measured today against the live primary (`billing-db-2`, per `status.currentPrimary`):

```
$ kubectl -n billing exec billing-db-2 -c postgres -- psql -U postgres -d openbank_billing \
    -c "SELECT status, count(*), min(created_at), max(attempt_count) FROM billing_outbox GROUP BY 1;"
 status | count |             min              | max
--------+-------+------------------------------+-----
 DEAD   |     2 | 2026-07-13 12:29:09.49986+00 |  10
```

Both rows, identically:

```
last_error | com.openbank.billing.infrastructure.outbox.BillingOutboxDispatcher#publishWithResilience
             circuit breaker is open
```

**Not a payload or serialization problem.** Both payloads are well-formed JSON that
deserializes cleanly (`schemaVersion`, `idempotencyKey`, `cycleId`, `accountId`, `feeId`,
`amount`, `currency`, `description`). So the second branch of the issue's question — "a
permanent payload problem that will kill every future row of the same shape" — is ruled
out by the data, not by inference.

It is the **latched-breaker** signature, and it is the *same defect as #4005*, already
fixed. #4163 (`ce0ef7954`, 2026-08-08) states the mechanism precisely: a fast-fail from an
open breaker is not a delivery attempt, but `markFailed` counted it against the row, so a
transport outage — not a poison payload — drove rows to terminal `DEAD`.

Two independent pieces of evidence that this is that, and not something billing-specific:

* **The timing matches the mechanism arithmetic.** `assessed_fee.updated_at` for the two
  stranded fees is `12:30:12.78` and `12:30:13.18` — **63 seconds** after the outbox rows
  were written at `12:29:09.50`. Billing polls on `openbank.outbox.poll-interval:5s`, so 10
  attempts is ~50 s. Ten real delivery attempts against a broker could not complete in that
  window; ten microsecond-scale breaker fast-fails do. The rows were burned through the
  attempt budget **without ever being offered to the transport**.
* **`BillingOutboxDispatcher` extends the shared `AbstractOutboxDispatcher`**, so it takes
  `OutboxDispatch.isTransportUnavailable` from libs-domain — the #4163 fix — and the
  deployed image is a descendant of it:

```
$ kubectl -n billing get pod ... -o jsonpath='{.spec.containers[*].image}'
  .../openbank-billing-service:sandbox-b9e265d4
$ git merge-base --is-ancestor ce0ef7954 b9e265d4 && echo YES
YES
```

**So the mechanism cannot recur here.** What was never fixed is that nothing said so.

## 2. Is `DEAD` alertable? No — and the near-miss is the interesting part

Checked, not assumed. The only outbox alerts in gitops are `*OutboxBacklogStuck`
(transaction, clearing, swift, domestic-payment, lending, the tier-1 generic) — every one
selects `openbank_outbox_backlog`. Billing's only rule is `BillingServerErrorsWhileHealthy`.
There is **no alert on a DEAD count anywhere in the fleet**.

And the backlog gauge cannot become one: `countProcessable()` is `PENDING + FAILED`, and
DEAD is excluded deliberately (ADR-0050 N5 parks a poison row so it cannot starve the
batch). So `openbank_outbox_backlog{service="billing"}` read `0` for the whole month these
rows sat parked — **the same value a perfectly healthy service publishes.**

The near-miss worth recording: `BillingOutboxRepositoryImpl.markFailed` *does* flip the
originating `AssessedFee` to `PostingStatus.FAILED` when a row goes DEAD, and billing's
backlog-gauge KDoc cites exactly that as the compensating signal. It is real — both
stranded fees are `FAILED` in `assessed_fee` today, so billing's own state is consistent
and nothing is half-posted. But nothing exports `posting_status` as a metric and no
PrometheusRule reads it. **The loss was recorded, not alertable**: visible only to someone
already running the query, which is how it was found.

### The alert, and why it is a success-state alert

```yaml
- alert: BillingOutboxDeadLettered
  expr: openbank_outbox_dead_lettered{service="billing"} > 0
  for: 15m
  labels: { severity: critical, service: billing }
```

`DEAD` is terminal: once a row is parked the failure has **finished failing**.
`listProcessable` excludes it, so the dispatcher never touches it again and emits no
further log line. There is no error rate left to watch — the only observable is the
resting state, so the alert asserts "no rows are parked".

**The `openbank_outbox_dead_total` counter cannot back this**, and libs' test pins why in
code rather than in a comment: a Micrometer counter is not created until its first
increment and resets on every pod restart, so it answers "did *this process* dead-letter
anything", never "are there dead rows in the table". Measured at zero series fleet-wide
while two services held dead rows.

The gauge carries an ADR-0237 heartbeat, for a reason specific to `> 0` alerts: a dead
refresh tick pins the gauge at its **boot zero** — the reading that looks healthiest — and
the alert then never fires, which is the exact silence being fixed.

## 3. Should they be replayed? Recommendation: **discard, do not replay**

Stated, not done — no row was written. The data:

| | |
|---|---|
| `cycle_id` | `verify-2026-07` (both), `idempotency_key` prefix `fee-verify-…` |
| Fees | `Monthly Fee` 99.0000 CZK, `FX Conversion` 1.5000 CZK |
| Account | a single account, `032c1003-…`, same `assessment_id` for both |
| `posting_status` | `FAILED`; `journal_id` NULL, `posted_at` NULL |

Three reasons to discard:

1. **This is verification-cycle data, not customer billing.** The cycle id is literally
   `verify-2026-07`, one synthetic account, 100.50 CZK total. Replaying posts real GL
   journals for a July verification run.
2. **The state is already consistent.** `journal_id` and `posted_at` are NULL and the fee
   is `FAILED` — nothing is half-posted, so there is no imbalance to repair. Discarding
   removes an artifact; it does not drop a pending obligation.
3. **The mechanism that created them is gone** (#4163, deployed here). They are debris from
   a fixed defect, exactly as #4005's 24 rows were — and that issue resolved by discarding
   after a CSV dump.

Recommended, at the owner's discretion and not mine to run: dump both rows to CSV, then
delete them. If they are instead judged to be real billing, the requeue is
`status='PENDING', attempt_count=0` — and *only* after the dump, since a requeue overwrites
`last_error`.

## Falsifying the guards

Neither guard is trusted on a green run; both were driven to red first.

* **Wrong selector.** Pointing the alert at `openbank_outbox_TOTALLY_WRONG` → red. This
  also confirms the `inputs.file` wiring: without it the task is `UP-TO-DATE` and a
  gitops-only edit is invisible to a path-scoped CI that never builds the module.
* **Uncovered binding.** Adding a `*OutboxDeadLetterGauge.kt` with no alert row → **green
  at first.** The scope guard walks the service tree, but Gradle had no input covering it,
  so the task was `UP-TO-DATE` and the guard was blind to the one event it exists to catch
  — the same staleness trap #4308 documented for the rule file, one directory over.
  Declaring the binding `fileTree` as a task input makes it red. Restored green after.

The scope guard exists because the alert list is otherwise a hand-kept list of the thing it
checks, which reads as *passing* when it is short rather than as *unchecked*.

## Verification

| Check | Result |
|---|---|
| `:openbank-billing-service:test` | **86 tests / 0 failures / 0 errors / 0 skipped** (JUnit XML) |
| `:openbank-libs-runtime:test` | **220 tests / 0 failures / 0 errors / 0 skipped** (JUnit XML) |
| `quarkusBuild` (fast-jar, CDI/ArC) | pass |
| `koverVerify` (both modules) | pass |
| `ktlintCheck`, `detekt` | pass |
| `check-scheduler-liveness.py` | pass (the new `@Scheduled` carries a heartbeat, takes no exemption) |
| `check-no-runblocking-in-scheduled.py` | pass (`refresh()` is a `suspend fun`) |

No API change, so no `openapi.yaml` bump. No DB migration. `version.txt` untouched —
release-please cuts it from the `fix` type plus the `src/main` path.

The protected-path guard will hold this PR because it touches `openbank-billing-service/**`
(money-path). That is expected and correct; it needs a human.


JiRaska:
The `verification-metadata` failure here is not a metadata gap — it is the OOM tracked as #4907, which **#4909** fixes.

From this PR's failing job:

```
Java heap space
##[error]Gradle failed (gradle exited 1); this says nothing about whether the
metadata is complete. Re-run, and if it persists treat it as a build problem,
not a metadata gap.
```

The gate is being honest about its own limits — that message is the gate declining to produce a verdict rather than producing a wrong one, so `all-green` correctly refuses too. It is not something a re-run will clear: the heap ceiling is the cause, and it persists.

So the merge order is **#4909 first**, then re-run this PR's checks. No change is needed on this branch.

Refs #4907, #4909

Issue #4728, opened 2026-08-13
five more per-pod in-memory stores with no durable row — #3467's shape, minus the thing that made it fixable


JiRaska:
## What

#3467 fixed one instance of a shape that exists in at least five more places: authoritative state
held in a per-pod `ConcurrentHashMap`/`AtomicReference`, with no cross-replica propagation and
nothing recording that the code depends on `replicas: 1`.

The merged PR for #3467 (#3644) stated these were "filed separately rather than fixed here". They
were not — searched `all` state on 2026-08-14, no such issue exists. Filing it now so the claim
becomes true.

## The distinction that makes these worse than #3467

The compliance pack had a durable `compliance_pack_activation` row. That is why the fix could be a
convergence loop reading the record of truth, and why a pod restart lost nothing. **None of the
sites below has a row to rehydrate from.** For them a restart is not "catch up on the next tick",
it is data loss — so they cannot be fixed by copying `CompliancePackRefresher`; each needs a real
home for its state first.

## Measured on `origin/main` (2026-08-14)

| site | state | durable? |
|---|---|---|
| `openbank-security-scanner/.../IctIncidentService.kt:36` | `ConcurrentHashMap<UUID, IctIncident>` — the DORA ICT incident register | no repository |
| `openbank-analytics-sink/.../proposal/InMemoryProposalStore.kt:22` | `ConcurrentHashMap<String, Proposal<BackfillRequest>>` — maker-checker proposals | no; the class name says it |
| `openbank-customer-edge/.../rest/PaymentSessionStore.kt:65` | `ConcurrentHashMap<String, Session>` — nearby-payment sessions | no |
| `openbank-agent-service/.../AgentSvidVerifier.kt:56` | `ConcurrentHashMap<String, Instant>` — single-use PoP nonce cache | no |
| `openbank-security-scanner/.../SecurityScannerService.kt` | scan results | no |

Two of these are not caches whose staleness is a nuisance:

- **`InMemoryProposalStore`** is a maker-checker store. Same four-eyes shape as #3467, without the
  durable row — a proposal approved on pod A is unknown to pod B, and a restart drops pending
  proposals entirely.
- **`AgentSvidVerifier.seenNonces`** is a **replay control**. Single-use is enforced per pod, so
  under N replicas a replayed PoP simply needs to land on a different pod to be accepted. That is a
  security property that silently weakens in proportion to replica count, not a performance nit.

[...]

PR #4939, opened 2026-08-15
fix(security-scanner): give the DORA ICT incident register a durable row

JiRaska:
## What

Moves the DORA ICT incident register out of a per-pod `ConcurrentHashMap` and into an
`ict_incidents` row. One of the five sites in #4728; the per-store verdict for the other four is a
comment on the issue rather than more commits here.

`Refs #4728, #3467, #3644, #3806`

## Why this one first

Of the five, this is the only one where state is lost **today, at `replicas: 1`**. The others are
latent-on-scale — they need a second pod before anything diverges. This one does not: any pod
restart emptied the whole register, and `GET /api/v1/ict-incidents` then answered `200` with `[]`
as though nothing had ever been reported. The field that makes it worst is
`reported_to_regulator` — a restart silently reset the register to "nothing was ever reported to
the regulator", with no error anywhere.

It is also the sibling of #3467 that could **not** be fixed the way #3467 was. The compliance pack
had `compliance_pack_activation` behind it, so `CompliancePackRefresher` could converge each pod
onto the record of truth every 30s. Here there was no row to converge onto, so the fix had to be
the row itself.

## ⚠️ Migration version collision — read before merging

There is a live, **unpushed** worktree on `chore/security-scanner-outbox-4709` carrying
`V4__drop_security_outbox.sql`. This PR adds `V4__create_ict_incidents.sql`. Two different files,
same version prefix: git merges both **cleanly, with no conflict**, and the service then refuses to
boot with `FlywayException: Found more than one migration with version 4` — at which point every
`@QuarkusTest` in the module reports **SKIPPED**, and the module reads as a pass because the number
anyone looks at is the failure count.

Whichever of the two merges **second** must renumber to `V5`. Renumbering pre-emptively does not
help: it converts the duplicate-version failure into an out-of-order one (`V4` applying after `V5`
is already applied), so the correct resolution really is "second one renumbers", and it is safe to
do because neither migration has been applied to a live database yet.

## Changes

- **`V4__create_ict_incidents.sql`** — creates `ict_incidents` (rollback: `DROP TABLE ict_incidents;`).
  No sequence, in contrast to `V3`: the id is application-assigned (`UUID.randomUUID()` in
  `reportIncident`), so the entity is `PanacheEntityBase` with an explicit `@Id` rather than
  `PanacheEntity`.
- **`IctIncidentRepository`** (port) + **`IctIncidentRepositoryImpl`**, which saves through
  `Panache.getSession().flatMap { it.merge(entity) }`. `persist()` is INSERT-only on an assigned id
  — Hibernate cannot tell transient from detached — so every status transition would 500 with
  `duplicate key value violates "ict_incidents_pkey"`. That is the failure consent-service shipped
  (#1521), and no mocked-repo test can see it.
- Filtering and paging move from an in-memory `.filter {}` / `.drop().take()` to a `WHERE` clause
  plus `range`, so a large register is no longer materialised in order to be discarded.
- The service and resource become `suspend` over the reactive repository. Every REST parameter that
  can legitimately be absent was already declared nullable, so the `suspend`-drops-the-null-check
  footgun does not apply here.

**No API contract change** — request and response shapes are byte-identical, so
`openapi.yaml` `info.version` is untouched. Not a money-path service.

## Test evidence

`IctIncidentDurabilityIT` drives the **real REST endpoints** — a reactive Panache repo called from
a bare `@QuarkusTest` thread throws `No current Vertx context found`, since only an HTTP request
carries a Vert.x context — and asserts over **plain JDBC** on a separate connection.

The JDBC read is the part that can see the defect. A test that reported an incident and read it
back through the same pod's REST API passes perfectly well against a `ConcurrentHashMap`: the map
is right there and answers correctly. Only a row read from the database proves the state left the
process.

**Validated against a known-positive rather than assumed:** swapping `merge` for `persist` reddens
exactly the two update-path tests (`a status transition updates the same row`, `the
regulator-reporting flag is durable`) and leaves the two insert-path tests green. The assertion
discriminates the thing it claims to.

Counts read from the JUnit XML, not the console:

```
:openbank-security-scanner:test — tests=17 failures=0 errors=0 skipped=0
```

`skipped=0` is load-bearing here: an earlier run of this same suite showed `1 failed, 3 skipped`,
which was a `QuarkusBindException: Port already bound: 8081` boot flake cascading into three
aborted tests — not a real failure, and not something the summary line distinguishes.

`ktlintCheck` and `detekt` were run separately after the test task (Gradle stops at the first
failing task, so a green-looking list is not a list that ran) and re-run after reverting the
unrelated files `ktlintFormat` had reformatted, so the gate is green on this diff alone.


JiRaska:
## ⚠️ Flyway V4 collision with the other open security-scanner PR

Both open PRs against `openbank-security-scanner` add a **V4** migration:

| PR | file |
|---|---|
| #4939 | `openbank-security-scanner/src/main/resources/db/migration/V4__create_ict_incidents.sql` |
| #4940 | `openbank-security-scanner/src/main/resources/db/migration/V4__drop_security_outbox.sql` |

**Whichever merges second must renumber to V5 before merging.**

The reason this needs saying out loud rather than trusting the merge: the version is the *filename prefix*, so these are two different files. Git reports **no conflict**, the merge is textually clean, and nothing goes red. The service then refuses to boot with `FlywayException: Found more than one migration with version 4`.

And the failure hides itself. Quarkus cannot start, so every `@QuarkusTest` in the module reports as **SKIPPED** — the module reads as `N tests / M skipped / 0 failures`, which scans as a pass because the number people look at is the failure count.

Renumbering pre-emptively does not help either: if the pre-emptively-renumbered one lands *first*, you get V4 missing and V5 present, which is an out-of-order failure instead.

So: check this PR's sibling immediately before merging, and renumber only the second one. Safe to renumber only while the migration has never been applied — after that, Flyway checksums the whole file and any edit is a startup failure.

JiRaska:
**V4 collision resolved** — this PR's migration is now `V5__create_ict_incidents.sql` (was `V4__`), because #4940 also adds a `V4__` to `openbank-security-scanner` (`V4__drop_security_outbox.sql`).

The two would not have conflicted in git. The Flyway version is the *filename prefix*, so they are different files: the merge is textually clean, nothing is reported, and the service then refuses to start with `Found more than one migration with version 4`. Either merge order produces it, and both PRs were green and mergeable at the same time.

The failure mode is worse than the failure: Quarkus cannot boot, so every `@QuarkusTest` in the module reports **SKIPPED** rather than failed, and the module reads as `N tests / M skipped / 0 failures` — which scans as a pass, since the number anyone looks at is the failure count.

Safe to rename: the migration has never been applied to a live database (so the checksum rule does not bite), and nothing outside the file references it by name. The `-- Rollback:` note is unchanged, which `db_change` now requires as an **enforced** rule.

#4940 keeps `V4`. No action needed there.

PR #4940, opened 2026-08-15
fix(security-scanner): delete the openbank.security.scan.event outbox, which never had a writer

### LINKED ISSUE #4709: openbank.security.scan.event has no writer: the whole outbox pipeline exists and nothing ever constructs a message
## What

`openbank.security.scan.event` is a fully provisioned pipeline with nothing at the start of it. `openbank-audit-service` subscribes to a topic that structurally cannot receive anything.

Surfaced while measuring #3994; verified independently against live `origin/main` before filing.

## The apparatus that exists

- `SecurityOutboxPort` + `SecurityOutboxMessage` (the data class)
- `SecurityOutboxEntity`, `SecurityOutboxRepositoryImpl`
- `SecurityOutboxDispatcher` (`@Scheduled`), `SecurityOutboxBacklogGauge`
- `KafkaSecurityOutboxEventPublisher`, `@Channel("security-events-out")`
- `application.yaml:111` — `security-events-out` → `topic: openbank.security.scan.event`
- gitops: the topic, a `KafkaUser`, mTLS for `security-scanner`, matching ACLs on the audit side

## What does not exist

**`SecurityOutboxMessage` has zero construction sites in the entire service** — `src/main` *and* `src/test`. Checked all three write idioms this repo uses, because a grep for one of them is how #4007 mis-classified billing:

| idiom | result |
|---|---|
| `SecurityOutboxMessage(` | only the `data class` declaration itself |
| raw `INSERT INTO …outbox` | none |
| JPA entity construction | only `toEntity()` in the repository — the **drain-side** mapper, present in all 34 services, not a write |

`git grep -l 'SecurityOutbox' openbank-security-scanner` returns 8 files and not one of them writes a row.

## Why it did not show up before

Two nearby things are already fixed and are *not* the explanation, so ruling them out matters:

- **#3350** (Flyway never ran, so `security_outbox` did not exist and every write 500'd) is fixed — `application.yaml` now sets `flyway.migrate-at-start: true` and a literal `jdbc.url`. The table is real.
- The publisher and dispatcher are correctly wired; `dispatchScheduledBatch()` returns `Uni<Void>` rather than wrapping reactive Panache in `runBlocking`, so the `HR000068` scheduler trap does not apply.

Everything downstream of the missing write is in working order. That is exactly what makes it invisible: there is no error, the dispatcher has nothing to dispatch, `attempt_count` stays 0, and the backlog gauge reads a truthful zero.

## Interaction with #4698 — worth checking before that merges

`openbank-security-scanner` is **not** in `check-outbox-has-writer.py`'s baseline, so the gate currently considers it fine. #4698 teaches that gate the JPA-entity idiom while excluding the drain-side `toEntity()` mapper. If the exclusion works as intended, this service should become a **new** violation the moment #4698 lands — and if it does not, that is a second blind spot in the same gate worth understanding.

## What needs deciding

Same fork as the `audit` / `psd2` / `tpp-registry` cases in #4007, and the answer is probably not "write the missing producer":

1. **Should security-scanner publish scan events at all?** If yes, the writer belongs inside the same transaction as the scan result, with a real-DB IT proving atomicity.
2. **If not, delete the apparatus** — port, entity, repository, dispatcher, gauge, publisher, channel, topic, KafkaUser, ACLs, and the audit-side subscription. A dead pipeline that looks like a control is worse than no pipeline, because its presence argues that scan events are being audited.

Refs #3994, #4007

JiRaska:
## Decision: delete the apparatus

`openbank.security.scan.event` had a complete outbox pipeline and no writer. This PR removes it — the port, message type, entity, repository, `@Scheduled` dispatcher, backlog gauge, Kafka publisher, the `security-events-out` channel, the Strimzi topic, both ACLs, and the audit-side subscription.

Closes #4709.

## Why not build the writer

Not because scan events are unwanted, but because **an outbox cannot do its job in this service**. The pattern exists to commit the event in the same transaction as the aggregate change, and there is no aggregate to be atomic with:

- `SecurityScannerService` holds scan results in a `ConcurrentHashMap` (`lastResults`, `lastReport`).
- `IctIncidentService` holds incidents in another `ConcurrentHashMap` (`store`).
- `security_outbox` was the **only business table in the entire service** — the only `*Entity.kt`, the only repository, and the subject of both migrations.

An outbox row committed beside an in-memory map is strictly worse than a direct publish, which is the idiom `IctIncidentService` already uses for `openbank.security.ict.incident` (untouched here — that topic has a real producer).

The two disjoint package roots are the fingerprint: the outbox lived in `com.openbank.security.*` while the whole actual service lives in `com.openbank.securityscanner.*`. It was scaffolded into a package the service never used.

## Evidence (live, read-only, with known-positive controls)

| Probe | Result | Control in the same read |
|---|---|---|
| `SecurityOutboxMessage` construction sites, `src/main` + `src/test` | **0** (declaration + the drain-side `toEntity()` mapper only) | `LendingOutboxMessage(` → 5 hits |
| `SELECT count(*) FROM security_outbox` on the live DB | **0**, ever | table exists; `flyway_schema_history` has V2+V3 `success=t`, so #3350 is genuinely fixed |
| Broker end offset, `openbank.security.scan.event` | **0** | `openbank.customer.audit` 823, `openbank.sanctions.screening.event` 55/68/55 |
| `audit_entries` rows with `source_service='security-scanner'` | **0 of 1979** | `sanctions-service`, `domestic-payment`, `card-issuance-service` all attributed in the same query |

Oldest `audit_entries` row is 2026-06-05. The scanner's `@Scheduled(every = "30m")` has been running continuously (pod up 37h at measurement, ~74 scans). **"audit subscribes" and "audit has ever received" are different facts, and the answer to the second is no — not once.**

## Also corrected: the docs were fiction well beyond the outbox

The service docs, the threat model, the gitops header comments and the admin-ui cloud-architecture node all claimed a `security_outbox`, an ICT-incident table and "scan history in CNPG Postgres". None has ever existed. After `V4`, the database holds `flyway_schema_history` and nothing else; scan results and ICT incidents are in-memory and lost on pod restart. The docs now say that, including the DORA/EBA compliance page, which no longer credits a delivery guarantee that was never there.

## Migration

`V4__drop_security_outbox.sql` drops the table and its sequence, with rollback DDL in the header. **V2/V3 are deliberately kept** — both are recorded as applied in the live `flyway_schema_history`, and deleting an applied migration's file fails validation as surely as editing one fails the checksum. No data-restore clause is possible or needed: the table never held a row.

## Verification

| Check | Result |
|---|---|
| `:openbank-security-scanner:test` | **10 tests, 0 failures, 0 errors, 0 skipped** (JUnit XML) |
| `:openbank-audit-service:test` | **109 tests, 0 failures, 0 errors, 0 skipped** (JUnit XML) |
| admin-ui `npm test` | **106 files, 1292 tests, all passing** |
| `ktlintCheck` + `detekt`, both modules | pass (baselines pruned of the deleted files: ktlint 15→9 file blocks, detekt 26→24 IDs) |
| `quarkusBuild`, both modules | pass — CDI wiring, which unit tests cannot see |
| `koverVerify`, both modules | pass |
| `run-gates.py --all` | 147 gates. `event-contract-coverage-ratchet` failed exactly as designed — "stale baseline entry … no longer produces to that topic" — and is fixed in c50840145. Every other failure reproduces identically on pristine `origin/main` and is environmental (`PR_DIFF_BASE`/`PR_TITLE` unset locally, `sudo` password for the `oasdiff` install). |

Zero skipped tests in both modules was checked explicitly, not inferred from the failure count.

## Note for review

The gitops half — deleting the `KafkaTopic` and the Write/Read ACLs — is a protected-path change and is expected to be held for human review. It is safe precisely because of the offset-0 measurement: there are no records to strand and no consumer-group offsets to orphan. `security-scanner` keeps its `KafkaUser`, mTLS certs and ESO projection for the surviving ICT-incident topic.

Also worth a second opinion: **`openbank.security.ict.incident` is at end offset 0 too.** Unlike scan.event it *has* a writer, so this is most likely just "nothing has ever called `POST /api/v1/ict-incidents`" rather than a second missing producer — but it means the DORA ICT incident path has never been exercised end-to-end either. Left alone here; it deserves its own issue rather than being bundled into a deletion.


JiRaska:
## ⚠️ Flyway V4 collision with the other open security-scanner PR

Both open PRs against `openbank-security-scanner` add a **V4** migration:

| PR | file |
|---|---|
| #4939 | `openbank-security-scanner/src/main/resources/db/migration/V4__create_ict_incidents.sql` |
| #4940 | `openbank-security-scanner/src/main/resources/db/migration/V4__drop_security_outbox.sql` |

**Whichever merges second must renumber to V5 before merging.**

The reason this needs saying out loud rather than trusting the merge: the version is the *filename prefix*, so these are two different files. Git reports **no conflict**, the merge is textually clean, and nothing goes red. The service then refuses to boot with `FlywayException: Found more than one migration with version 4`.

And the failure hides itself. Quarkus cannot start, so every `@QuarkusTest` in the module reports as **SKIPPED** — the module reads as `N tests / M skipped / 0 failures`, which scans as a pass because the number people look at is the failure count.

Renumbering pre-emptively does not help either: if the pre-emptively-renumbered one lands *first*, you get V4 missing and V5 present, which is an out-of-order failure instead.

So: check this PR's sibling immediately before merging, and renumber only the second one. Safe to renumber only while the migration has never been applied — after that, Flyway checksums the whole file and any edit is a startup failure.

JiRaska:
**V4 collision resolved on the other side** — no action needed here.

#4939 also added a `V4__` to `openbank-security-scanner` (`V4__create_ict_incidents.sql`). It has been renumbered to `V5__`; this PR keeps `V4__drop_security_outbox.sql`.

Recording why it needed a deliberate fix rather than being caught: the Flyway version is the *filename prefix*, so the two are different files — git merges them cleanly, nothing is reported, and the service then refuses to start with `Found more than one migration with version 4`. Both PRs were green and mergeable simultaneously, so either merge order produced the break. And once Quarkus cannot boot, every `@QuarkusTest` in the module reports **SKIPPED**, not failed, so the module reads `N tests / M skipped / 0 failures` — which scans as a pass.

Issue #4743, opened 2026-08-13
External-feed liveness needs a fetch-outcome contract, not another last-success heartbeat


JiRaska:
## Split out from #3345

#3345's acceptance criteria 2 and 4 are untouched and need a different mechanism from the rest of that issue, so they are better tracked here. #3345 covers `@Scheduled` job liveness — 38 of 48 adopted as of PR #4739, with the gate enforced since #4180, so the remaining work there is a finite legacy tail.

This is the other half: **external feeds**, and criterion 2 is where #3345's own motivating incident lives — the ČNB feed returning 404.

## Why more `registerWorkflowLiveness` call sites will not fix it

The existing mechanism answers "did this job run and succeed". For a feed, that is not the interesting question, because a fetch can *succeed as a job* while producing nothing usable:

- ran, got HTTP 200, parsed N rates → healthy
- ran, got **404** → the job completed; nothing threw; a naive last-success heartbeat reads green
- ran, got 200, parsed **zero** rows → same, and quieter still
- did not run at all → what the current mechanism already covers

So what is needed is a **fetch-outcome contract** that distinguishes those, not another heartbeat. The distinction is the whole point: a feed that reliably fetches nothing is indistinguishable from a healthy one under a run/no-run signal.

Note the shape this repo keeps meeting: a no-op must have its **own** outcome value, never one shared with success (`PushSendOutcome.SKIPPED` is the precedent), and the alertable state here is the **success** state — "fetched on schedule, produced nothing, every time" raises no errors at all.

## Scope

1. Define the fetch-outcome contract (at minimum: `FETCHED` / `EMPTY` / `HTTP_ERROR` / `PARSE_ERROR`), with a metric labelled by feed and outcome.
2. Apply it to the external feeds, ČNB first since it has a known incident behind it.
3. Fold `external-feed-watch.yml` into the same mechanism rather than leaving a second, parallel way to ask the same question — that is #3345's criterion 4, and a second mechanism is worse than either alone.

## Constraint worth knowing before writing the rule

[...]

PR #4943, opened 2026-08-15
feat(fx): distinguish feed fetch outcomes from job liveness

JiRaska:
## What

Defines the fetch-outcome contract from #4743 and applies it to the ČNB daily fixing — the one feed with a real incident behind it. A per-feed verdict on the rest of the fleet follows as a comment; this PR does not sprawl across them.

`Refs #4743`, `Refs #3345`.

## The gap, measured rather than assumed

I traced all four outcomes through the live ČNB path before writing anything, and the issue's framing needs one narrowing: **three of the four are already distinguishable, and the fourth is worse than the issue suggests.**

| outcome | today | |
|---|---|---|
| 200, N rows | `recordSuccess()` | correct |
| 404 / 5xx | RestClient throws → caught → one ERROR line, no `recordSuccess` | gauge goes stale — but *why* is unrecoverable |
| 200, zero rate lines | `CnbFixingParser` `require`s → same catch | same |
| 200, parses, **none of our currencies** | `ingested=0, skipped=0` → **`recordSuccess()`** | **no signal at all** |

So the honest statement of the problem is two things, not one:

1. The three failure modes collapse into a single indistinguishable signal — the age gauge stopped advancing — which is also what "the job never ran" looks like. Four causes, one observable, and the log line is the only discriminator (the thing that failed to help for 46 days).
2. There is a genuine hole on the **success** side: a fixing that parses perfectly but carries none of `openbank.cnb.currencies` records a success. An upstream renaming or dropping a code is silent, forever, with no error raised anywhere for an error-rate alert to see.

## The contract

`FeedFetchOutcome` = `FETCHED | EMPTY | HTTP_ERROR | PARSE_ERROR | UNREACHABLE`.

`UNREACHABLE` is the one addition beyond the issue's four, and it is borrowed from `check-external-feeds.py`'s own triage, which learned it the hard way: *unreachable is not a verdict about the feed, it is the absence of one*. `apl.cnb.cz` genuinely answers some network positions and not others, and reporting that as "the feed is dead" is how a daily job becomes permanently red and gets filtered.

`EMPTY` gets its own value rather than a `success` boolean, for the reason this repo already paid for once: `PushResult.skipped()` carried `success = true`, so every push in an environment with no APNs credentials was counted as delivered (#4348). Narrower than "stored nothing" — an idempotent re-run (`ingested=0, skipped=N`) is `FETCHED`, because the rates arrived and we already had them. Classifying that as `EMPTY` would make the feed look dead every time the scheduler ran twice.

## How it sits beside the heartbeat, not on top of it

ADR-0237 point 2 had already decided this, so I implemented it rather than inventing a parallel mechanism: the feed registers **the same** mechanism-3 primitive under `workflow="feed-cnb-daily-fixing"`, advanced only on `FETCHED`. The job keeps its own `fx-cnb-ingestion` registration, unchanged, including on the `EMPTY` outcome — a run that legitimately stored nothing is a run, not a miss (ADR-0237 point 1).

Two series, and their **disagreement** is the diagnosis: job fresh + feed stale is precisely "the scheduler is running fine against a dead upstream".

Reusing the primitive rather than adding a gauge family buys three things already built and already argued: `WorkflowLivenessStale` covers feed freshness with no new rule or threshold, the ADR-0163 sentinel correlates it with everything else, and the age gauge's registration seed (#4208) is inherited rather than re-derived.

What is new is only the *reason*: `openbank_feed_fetch_total{feed,outcome}`.

`FeedFetchRecorder.record` both increments the counter and advances freshness iff `FETCHED`. One call, so the two signals are structurally unable to drift apart — a caller cannot mark a feed fresh without saying what it fetched. All five counters are created at registration, because a lazily-created counter makes "never 404'd" and "never instrumented" the same empty vector.

### Why the counter is not an alert expression

Retention here is 12h with no long-term store, so `increase(openbank_feed_fetch_total[24h])` on a *daily* feed truncates silently and would answer about a window that does not exist. The alert stays the gauge (which carries state forward with no lookback at all); the counter is read at triage. The `WorkflowLivenessStale` annotation now says exactly that, so following the alert leads to the metric instead of to a log grep.

### Cold pod at t=0

Before any fetch has happened, on a fresh pod:

- `openbank_workflow_last_success_age_seconds{workflow="feed-cnb-daily-fixing"}` ≈ **pod uptime in seconds** — seeded at registration, never the ~1.8e9 that fired an alert 15 minutes after every deploy (#4208). The rule's threshold for this daily feed is 172800s, so a boot reading cannot reach it.
- `openbank_workflow_success_recorded{...}` = `0` — "this pod has not seen this feed deliver", which triage reads and the alert deliberately does not.
- `openbank_feed_fetch_total{...}` = `0` for **all five** outcomes, present rather than absent.

Nothing can fire at boot. Pinned by a test rather than asserted in prose.

## Testing

`CnbFeedOutcomeContractIT` drives the **real cron** from a `TestProfile` (literals in `getConfigOverrides` — the profile loads in a different classloader, so a companion object initialises twice) against a feed that is healthy in every way a job can measure: it answers, it parses, the use case returns normally, the scheduler logs INFO and records its own success. It simply carries no configured currency. The test then asserts both series **against each other, in one process, from one run**: job succeeded, feed not-yet. Either series alone is consistent with a healthy system.

Freshness is asserted on `success_recorded` (0/1), never on the age gauge's magnitude — on a pod this young a delivering feed and a dead one read the same small number, so an age-threshold assertion would pass for the wrong reason and keep passing if the contract were deleted. That is the vacuity trap from the `FIFTY_YEARS_SECONDS` sweep, one step earlier.

The `PARSE_ERROR` cases run through the real `CnbFixingParser` rather than a hand-built exception, so the classifier cannot stay green after the parser changes its failure type.

**Both new suites were falsified against a known-positive**, not just observed green:

- making `EMPTY` advance freshness → 3 of 11 red in `FeedFetchMetricsTest` (0 when correct)
- classifying `EMPTY` as `FETCHED` → the IT goes red

Counts from the JUnit XML, not the console:

|---|---|---|---|
| `openbank-libs-domain` | 426 | 0 | 0 |
| `openbank-libs-runtime` | 228 | 0 | 0 |
| `openbank-fx-service` | 152 | 1 | 0 |

The single skip is `FxPactProviderVerificationTest` — the broker-sourced Pact class gated on `pactbroker.url`, whose `@PactFolder` twin runs. Pre-existing and expected on the PR lane.

`ktlintCheck` and `detekt` were re-run on their own after the test runs and appear in the task list; `check-external-feeds.py --self-test` (15 passed) and its drift check both stay green.

## Risk

**`openbank-fx-service` is money-path** → 2 approvals + threat model (present at `docs/threat-models/openbank-fx-service.md`; unchanged, as this adds no new surface or data flow).

The `openbank-libs-*` change is purely **additive** — one new file, one new method, one new class, no signature change and no new `@Produces` — so no existing dependent's build can be affected by it. That matters because path-scoped CI would only surface a libs regression in whichever dependent happens to be in the diff.

No API change, so no `openapi.yaml` or contract-version movement. No DB or event change. `version.txt` untouched — release-please owns it.

## Scope: what is deliberately NOT here

Per-feed verdicts for the rest of the fleet are in a comment below. **Issue item 3 — folding `external-feed-watch.yml` into this mechanism — belongs in a later PR, and I think it needs a decision first rather than an implementation**; reasoning is in that comment.


JiRaska:
## Which feeds exist, and how I found them

Three external fetchers, found two ways — and the two methods disagree, which is itself a finding.

**Method 1, the declared registry.** `.github/scripts/check-external-feeds.py` already holds an enumeration with a *blocking* drift half: any external URL appearing in a scanned `application.yaml` that is not in `FEEDS` or `NOT_PROBED` fails CI. That is a much stronger enumeration than a grep, because it is enforced going forward rather than being a snapshot. It declares exactly **one** feed (`cnb-daily-fixing`) and eight `NOT_PROBED` entries, none of them scheduled data feeds.

**Method 2, an independent sweep of the Kotlin fleet** — `@RegisterRestClient` interfaces against non-openbank hosts, `@Scheduled` classes named `*Ingestion|Feed|Fetch|Import|Download|Sync|Refresh`, and raw `HttpClient`/`WebClient` use under `*/src/main/*`. This found **three**.

The delta is the finding: **the drift gate's scan scope is `openbank-*/src/main/resources/application.yaml`, and sanctions-service holds its eight source URLs in Flyway seed SQL (`sanctions_lists.source_url`), in the database.** So the gate that exists to stop a new external dependency becoming an unwatched feed is structurally blind to the fleet's *other* real feed — and reports green about it, because a check whose coverage set is maintained separately from the artifacts it covers reads as passing when the set is short rather than as unchecked. Same shape as the pact-drift scope before #2338 derived it.

### Per-feed verdict

| feed | service | records today | verdict |
|---|---|---|---|
| `cnb-daily-fixing` | fx-service | liveness on the non-throwing path; nothing else | **done in this PR** |
| sanctions/PEP lists (8 sources) | sanctions-service | **nothing** — no liveness, no metric | **next, and it is worse than ČNB** |
| GitHub metrics | devops-agent | liveness recorded for *dispatching*, before touching GitHub | **narrow fix, low value** |

---

### sanctions-service — the one that should be picked up next

Not a sweep item. Three separate defects, each of the exact family this contract is about:

1. **The status code is never inspected.** `httpGetStream` returns the error body as a stream. For the CSV sources a 404 page's first line becomes the header, `idxId`/`idxName` resolve to `-1`, and the method logs `"Unexpected CSV header"` and **returns 0** — no exception, nothing thrown for anyone to catch.
2. **A dead download and a healthy no-change refresh write the identical row.** `refresh` does `if (imported > 0) imported else list.lastEntryCount ?: 0`, then `markUpdated(...)`. So `last_updated` bumps and the previous count is carried forward — the feed being dead is indistinguishable from the feed being unchanged **in the persisted record itself**, which is worse than the ČNB case, where at least the table stopped growing.
3. **A third sanctions-service finding is handled through a private channel.** It is security-shaped — a control that can fail open — so per the repo rule (security holes go to private Security Advisories, not public issues) the mechanism is not restated here. It is not needed to review this PR.

That last one is a live control-integrity question, not an observability gap, so I would rather it got its own issue than ride along here. The contract in this PR is the right instrument for (1) and (2) — the classification is a near-copy — but (3) needs a guard in the import path (refuse to deactivate on a zero-row import), which is a behaviour change on a compliance control and wants its own review. Happy to file both if you want them split that way.

### devops-agent

`DevOpsAnalysisScheduler` records `recordSuccess()` after dispatching a workflow, having never called GitHub; and it returns early *without* recording when Temporal is disabled. Every GitHub failure mode — 404, 5xx, blank token, zero runs — funnels into the same `null`, so "no data" and "not configured" are the same observation. Real, but this is an internal agent's input metric with no money path and no compliance surface, so it is worth one small PR at low priority rather than a place in the sweep.

---

## Issue item 3 — folding `external-feed-watch.yml` in

**Next PR at the earliest, and I think it needs a decision before an implementation.** Two reasons to be careful about calling it "a second, parallel way to ask the same question":

1. **ADR-0237 point 2 explicitly decided the opposite** — *"The CI probe stays as-is: it falsifies the URL from outside; the in-cluster gauge measures freshness from inside."* Folding them silently would reverse a decision this PR otherwise implements verbatim. That is an ADR amendment, not a refactor.
2. **They are not measuring the same thing, and there is evidence in-tree.** `check-external-feeds.py` records that `apl.cnb.cz` 404s from GitHub's runners while working from the cluster and from a laptop — network-position specific, not an outage. So the CI probe answers "does this URL exist on the public internet" and the gauge answers "did this cluster get data", and a feed can genuinely fail one while passing the other. The CI probe is also the only thing that catches a **misconfigured URL before deploy**, which no in-cluster metric can do by construction.

The part of item 3 I do think is real is the *naming* seam: `FEED_NAME` in the scheduler and the `FEEDS` entry name are two hand-kept strings that must agree for the two lanes to be talking about the same feed, and nothing enforces it today. I kept them equal and said so in a comment, but a check deriving one from the other is the enforceable version and would be a good, small, self-contained follow-up. That is a narrower and more defensible reading of item 3 than merging the lanes.

PR #4959, opened 2026-08-16
governance(rules): graduate db_change and db_backup_association to enforce

JiRaska:
## The queue is blocked by a date, not a defect

`check-gate-graduation.sh` fails every PR since `2026-08-15` passed:

```
target_enforce_date 2026-08-15 has passed and this rule is still advisory/planned (ADR-0144).
Either ship the producer and flip to enforce/block, or move the date forward with a one-line reason.
```

Two rules are past their date: `change_requirements.db_backup_association` and `change_requirements.db_change`. It is currently red on **#4954, #4955 and #4956** — three unrelated PRs — because the guard runs on everything.

## This ships the graduation rather than moving the date

Both producers exist and both gates already pass on `origin/main`:

| rule | producer | gate on main | `gates.yaml` mode before |
|---|---|---|---|
| `db_backup_association` | `openbank-infra/scripts/check-db-backup-associations.py` | PASS | already `enforced` |
| `db_change` | `openbank-infra/scripts/check-db-migration.py` | PASS | `advisory` |

Note the first row: `db-backup-association-gate` has been `mode: enforced` in `gates.yaml` while `rules.yaml` called it `advisory`. The two disagreed, and CI was already the stricter of the two — so that flip only makes the governance statement match what has been happening.

`db-migration-gate` is flipped to `mode: enforced` alongside its rule, so the two files continue to agree rather than trading one disagreement for another.

## Verified against the PRs it would newly block

Enforcing a gate that fails on legitimate work would be worse than the date. Both currently-open migration-adding PRs pass under `--enforce`:

- **#4939** (`V4__create_ict_incidents.sql`) — `1 migration change(s) checked — all forward-only, all with a rollback note.`
- **#4940** (`V4__drop_security_outbox.sql`) — same.

## And falsified, because a gate that cannot fail is not worth graduating

Stripping the `-- Rollback:` note and **committing it** makes the gate report:

```
::error::New Flyway migration without a rollback note (rules.yaml: db_change requires one).
1 migration change(s) checked, 1 finding(s) above.
```

Worth recording how that nearly went wrong: the same strip left **uncommitted** reported a clean pass, because the script resolves files through `git diff` against the base rather than reading the working tree. A working-tree-only falsification would have "proved" the gate vacuous and argued for moving the date instead.

## Scope

`openbank-libs/governance/rules.yaml` (two `enforced:` values) and `.github/gates/gates.yaml` (one `mode:`). No script logic changed.

`change_requirements` is **not** in the OPA read set — the `.rego` sources read only `authz`, `feature_flags`, `four_eyes`, `money_path_action_prefixes`, `money_path_services` and `shared_m*` — so no bundle regeneration is required and no `policy-checksum` moves. Checked rather than assumed.

`check-gate-graduation.sh` after the change: `17 advisory/planned rule(s) checked, all carry a live target_enforce_date.`

### DIFF
--- a/.github/gates/gates.yaml
+++ b/.github/gates/gates.yaml
@@ -2277,7 +2277,7 @@ gates:
   - id: db-migration-gate
     name: "db-migration gate — rollback note + forward-only (rules.yaml db_change, advisory)"
     group: data
-    mode: advisory
+    mode: enforced
     verified: "2026-08-09 — on track for ADR-0144's 2026-08-15 enforce target"
     when: pull_request
     needs_base: required
--- a/openbank-libs/governance/rules.yaml
+++ b/openbank-libs/governance/rules.yaml
@@ -215,7 +215,7 @@ change_requirements:
       - "every property a request-body schema declares exists on the DTO the resource parses"
       - "every property the DTO has is declared on the request-body schema"
     gate: openapi-request-schema-conformance
-    enforced: advisory
+    enforced: enforce
     target_enforce_date: "2026-09-30"
     ci_producer: ".github/scripts/check-openapi-request-schema-conformance.py"
     # openapi_route_conformance (above) is ROUTE SET ONLY — it does not read schemas, so aml's
@@ -253,7 +253,7 @@ change_requirements:
       - "a CNPG cluster declaring barmanObjectStore -> the managed bucket has a matching
          aws_eks_pod_identity_association in db-backups.tf"
     gate: db-backup-associations
-    enforced: advisory
+    enforced: enforce
     target_enforce_date: "2026-08-15"   # ADR-0144
     ci_producer: "openbank-infra/scripts/check-db-backup-associations.py"
     # Backup credentials arrive via EKS Pod Identity, granted per (namespace, serviceAccount) by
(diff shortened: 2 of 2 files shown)


JiRaska:
Closing as superseded by `main`, with the measurement rather than the assertion.

This branch flipped four things. Three of them are already on `main`:

| change | state on `main` |
|---|---|
| `rules.yaml: db_backup_association.enforced` -> `enforce` | **landed** (`enforced_since: "2026-08-16"`) |
| `rules.yaml: db_change.enforced` -> `enforce` | **landed** (`enforced_since: "2026-08-16"`) |
| `.github/gates/gates.yaml: db-migration-gate.mode` -> `enforced` | **landed** |
| `rules.yaml: openapi_request_schema_conformance.enforced` -> `enforce` | not landed — and **deliberately not carried over**, see below |

So the ADR-0144 graduation this PR existed for is done, by another route, and rebasing would leave a diff of one unrelated flip.

**Why the fourth is not being re-filed as-is.** `openapi_request_schema_conformance` carries `target_enforce_date: "2026-09-30"` — six weeks out, not expired. Graduating a gate ahead of its target is a different decision from graduating the two that were past due, and it needs its own evidence that the gate passes on real work (which is not established here; the script is per-service, so "it passes fleet-wide" is not something this PR ever measured). Bundling it into an expired-date graduation is how an unevidenced flip rides in on someone else's justification.

Also worth recording, since it cost a debugging pass: this PR's `gates (supplychain)` failure — `New Flyway migration without a rollback note` — was **not** about anything in this PR. It changes two YAML files and adds no migration. The finding came from the diff being computed against a stale base while the branch was `CONFLICTING`. Same family as the api-contract version trap: read the gate's subject, not just its verdict.

If the openapi graduation is wanted early, it should be its own PR with per-service evidence.

Issue #4754, opened 2026-08-14
transaction.sweep has @Authorize(resource = ""), so one four-eyes approval satisfies a different sweep by the same maker


JiRaska:
## What

`transaction.sweep` is annotated `@Authorize(resource = "")`. An empty resource means `PendingApproval.resourceId` is stored as `null`, so a four-eyes approval is not bound to a specific sweep — **one approval satisfies a different sweep by the same maker.**

This is precisely the defect that `party.merge`'s own code comment documents having avoided. The two annotations sit in the same codebase and disagree.

## Why it is filed now, separately

Found while measuring #1984. It is **latent today** — `AUTHZ_FOUR_EYES_ENFORCE` is off for transaction-service, so no approval is currently required at all and there is nothing to mis-bind.

It stops being latent the moment anyone flips that flag, and #1984 is about to recommend flipping it on `party-service` for a genuine account-takeover exposure. So the ordering matters and is the point of this issue: **fix the resource binding before enabling four-eyes anywhere it applies**, or the control ships mis-bound on its first day and the mis-binding looks exactly like a working approval.

Kept out of the #1984 PR deliberately — `openbank-transaction-service` is in `rules.yaml: money_path_services`, so this needs 2 approvals and a threat model, and bundling it would have held an unrelated non-money-path fix behind that gate.

## The related fact that makes the binding matter

`POST /api/v1/transactions/merge-sweep` is `ROLE_OPERATOR`-gated and moves a balance. Combined with a party merge it is the second half of the takeover path #1984 documents. An approval that binds to the maker rather than to the specific sweep is therefore not a partial control — it is one that can be satisfied by an unrelated, innocuous request the same operator made earlier.

## What to check when fixing

- `@Authorize(resource = "#cmd.<id>")` is the working shape — `SanctionsResource.kt:57` reads `@Authorize(action = "sanctions.clear", resource = "#cmd.checkId")` after #3521 fixed exactly this class of empty-resource defect there.
- Verify the stored `PendingApproval.resourceId` is non-null **on a real request**, not in a unit test that constructs the approval directly.
- Then verify the negative: an approval issued for sweep A must **not** satisfy sweep B. That assertion is the whole fix; without it the test passes against the current code.
[...]

PR #5043, opened 2026-08-16
fix(transaction): bind four-eyes approvals to the request, not the maker

JiRaska:
## What

`transaction.reverse` and `transaction.sweep` are both `four_eyes_required` (`rules.yaml`), and both carried `@Authorize(..., resource = "")`. The interceptor stamps a `PendingApproval` with `query.resource?.id` — so with an empty resource, every approval for that action is created and matched at `resourceId = null`. A checker's approval to reverse transaction A therefore satisfies a *different* reversal of transaction B by the same maker; a checker's approval to sweep one duplicate-party merge satisfies a *different* sweep, to different accounts, for a different amount. Same defect class as #3521 (`sanctions.clear`).

## Fix

| endpoint | binds to | why |
|---|---|---|
| `POST /{transactionId}/reverse` | `#transactionId` | the obvious per-resource key |
| `POST /merge-sweep` | `#request.idempotencyKey`, not `mergeReference` | one merge reference covers every account pair swept in that merge — binding to it would let an approval for one pocket satisfy the sweep of another. The idempotency key is per-request unique and is what the maker replays on the `X-Approval-Id` retry, so the approval resolves for the exact request it was granted for and no other |

## Verification

`MergeSweepApprovalBindingIT` — real HTTP through `RestAssured` + `@TestSecurity`, not a direct resource-class call (which cannot prove authz ran at all). Asserts both directions:
- an approval issued for sweep A does **not** satisfy sweep B by the same maker (the negative — this is the whole control)
- the approval still satisfies the *exact* sweep it was issued for

`compileTestKotlin`: BUILD SUCCESSFUL. The full IT run could not be trusted locally — this machine had genuine concurrent Gradle activity from other sessions during this work (confirmed via `ps aux`, not assumed), which this repo's own notes already flag as a source of unreliable local results (`SocketTimeoutException` under contention, unrelated to the code under test). Real CI, uncontended, is the actual verification here.

Refs #4754

## Security checklist

      (license, CVE, source).
- [x] Auth / crypto / payment code changed? → **yes** — this is the four-eyes
      (maker-checker) binding on `transaction.reverse` and `transaction.sweep`.
      That is the whole subject of the PR; it tightens an authorization control
      rather than relaxing one, and is inert while `authz.four-eyes.enforce=false`.
- [x] PII path changed? → no.

## Update 2026-09-11 — rebased, and the IT now has a clean local run

Rebased onto current `main` (conflict was the threat-model change log only — this
entry is dated 2026-08-16 and slots between the 08-17 and 08-07 entries; all 21
entries preserved).

The verification note above said the full IT could not be trusted locally under
Gradle contention and that CI was the real check. CI then failed
`the approval still satisfies the exact sweep it was issued for()` with
`java.net.SocketTimeoutException` — the same shape. Re-ran the IT on an
uncontended checkout and read the JUnit XML rather than the console, because a
filtered Gradle run that executes nothing also prints BUILD SUCCESSFUL:

```
tests=4 failures=0 errors=0 skipped=0 time=162.179
  a four-eyes sweep is paused and mints a pending approval()            -> pass
  the stored approval is bound to the specific sweep()                  -> pass
  an approval issued for sweep A does not satisfy sweep B by the same maker() -> pass
  the approval still satisfies the exact sweep it was issued for()      -> pass
```

All four ran and all four pass, including the one CI timed out on. The failed CI
job has been re-run.


JiRaska:
Heads up before merge: `build (openbank-transaction-service)` is currently **red** on this PR — hit the 45-min job timeout (run `31954689821`, job `95210929320`), `all-green` red as a result.

From the raw logs this looks real, not environmental flake:
- All 4 `MergeSweepApprovalBindingIT` methods fail with `SocketTimeoutException` on the very first request. Looks like `AlwaysFourEyesPdp` (`@Alternative` CDI test double) isn't actually being resolved in place of the real `PolicyDecisionPoint` producer — request may be hanging trying to reach a real/unreachable OPA sidecar client instead of the test double, rather than a plain assertion failure.
- Right after, unrelated Temporal-workflow tests in the same module start failing too (`Payment workflow ... failed; running compensation`, `PostJournal`/`PlaceHold` activity failures) — consistent with the test profile's blanket `authz.enforce=true` + `AlwaysFourEyesPdp` (flags **every** action `four_eyes_required=true`, not just `transaction.sweep`) leaking into other tests sharing the same restarted Quarkus app instance.
- Job doesn't exit for ~35 min after the last test failure before the runner kills it at the ceiling — suggests something (Testcontainers cleanup, a non-daemon thread, or the hung PDP client) is also blocking JVM shutdown.

Worth checking the CDI alternative wiring in the test profile before merge.

JiRaska:
The event-loop deadlock blocking this PR's CI is fixed in #5636 (`fix(libs-runtime): ...`, `Refs #5631`) — merge that first.

After applying #5636's fix locally and re-running `MergeSweepApprovalBindingIT` against this branch's diff, 2 of the 4 test steps still fail with `AssertionError` (202 vs 200), unrelated to the deadlock. Root cause: the test's `AlwaysFourEyesPdp` stub flags *every* action as four-eyes-required — including the approval-decide endpoint itself — so approving an approval transitively requires approving that approval too, which the test doesn't set up. This is a pre-existing bug in the test fixture, not in this PR's `@Authorize` resource-binding fix. Needs a follow-up before this PR is fully green.

JiRaska:
Not unblocked yet — #6182 is open and must merge first.

The blocker is confirmed live: this PR's `build (openbank-transaction-service)` is still
`CANCELLED`, which fails `all-green`.

**Root cause (#5940), from job `96416555756` on this PR's head `f77c32458`.** The job was not slow.
The tests stopped at `12:09:46Z` and the remaining 34 minutes were 247 lines of
`Connection to node -1 (localhost/127.0.0.1:32771) could not be established`, until cancellation at
`12:43:38Z`. The decisive line is at the end:
`No files were found with the provided path: openbank-transaction-service/build/test-results/` — the
`test` task never completed, so no JUnit XML was written at all.

The issue text read the single broker port as evidence *against* a container being reaped between
test-resource restarts. The log says the opposite. There are exactly two `Ryuk has been disabled`
banners — `12:04:33Z` and `12:05:12Z` — so two container sets, and the retries begin one second
before the second set starts. Quarkus restarts the app when the set of `@QuarkusTestResource`
classes changes between test classes, which calls `stop()` on `PostgresRedpandaTestResource` and
kills Redpanda while a producer from the previous boot is still open. Only that orphaned producer
logs failures; the producers on the new boot connect fine and success is not logged.

So this is not an OOM and not specific to `MergeSweepApprovalBindingIT`. That test surfaces it
because it is the one that publishes — the latent-fault reading in #5940 is right.

**What #6182 changes.** Nothing that makes the broker reachable; it makes an unreachable broker fail
in seconds and name itself: Kafka client bounds in the test profile, a Gradle **task**-level timeout
at 25 minutes (the per-test JUnit timeout cannot see this hang — it happens after the last test
finishes, with seven `Terminate orphan process: pid (java)` at teardown), and a teardown log line so
"the broker went away" is legible without reading raw producer retries.

**What that means for this PR.** Once #6182 lands, rebase this branch onto `main` and re-run. The
expected outcome is one of two legible results instead of a 45-minute burn: a green
`MergeSweepApprovalBindingIT`, or a fast, named failure with a JUnit XML to read. Note the
`java.net.SocketTimeoutException at MergeSweepApprovalBindingIT.kt:209` in that job was itself a
symptom of the dead broker, not necessarily a defect in the four-eyes binding — so re-run before
concluding anything about the fix in this PR.

No change is needed to the four-eyes work here.

Refs #5940, #6182

JiRaska:
Update: #6182 is green and measured. It is not merged yet, so this PR is not unblocked *yet* — but
the fix is now demonstrated rather than argued.

`build (openbank-transaction-service)` on #6182: `12:45:32Z → 12:55:38Z`, **10m 06s**, SUCCESS,
`all-green` SUCCESS.

| signal | job 96416555756 (this PR) | #6182 |
|---|---|---|
| wall time | cancelled at 45m | 10m 06s |
| `could not be established` lines | 247 | 0 |
| `build/test-results/` | `No files were found` | JUnit XML artifact uploaded |

Once #6182 merges, rebase this branch onto `main` and re-run. Expect either a green
`MergeSweepApprovalBindingIT` or a fast named failure with a readable JUnit XML — not a 45-minute
burn either way.

Reminder for when you read that re-run: the `java.net.SocketTimeoutException at
MergeSweepApprovalBindingIT.kt:209` in the old job was a symptom of the dead broker, so it is not
evidence for or against the four-eyes binding change. Judge the four-eyes fix on the fresh run.

JiRaska:
## Review — no longer blocked by #5940; it is now merely stale

Reviewer comment only (no approve, no merge). Read the full 3-dot diff (3 files).

### 1. The blocker is gone — and the old failure said nothing about this change either way

**#6182 has merged to `main`** (`50da682f1`, *ci(transaction): fail fast when the test Kafka broker is gone, instead of burning the 45-minute job timeout*). This branch is 241 commits behind and does not contain it, so its last CI run is still the broken lane — `build (openbank-transaction-service)` shows `CANCELLED` and `all-green` `FAILURE`.

The previously-reported `SocketTimeoutException at MergeSweepApprovalBindingIT.kt:209` was a **dead-broker symptom**, not a verdict on the four-eyes binding: a producer left over from a prior Testcontainer boot retrying against a closed port. Line 209 is inside `postSweep`, which is just where the first HTTP call happened to land. It is not evidence for or against the fix.

**Action: rebase onto `origin/main` and re-run.** This is a staleness verdict, not a blocked one.

That matters more than usual here, because `MergeSweepApprovalBindingIT` adds `@QuarkusTestResource(PostgresRedpandaTestResource::class)` — a *new* `@QuarkusTestResource` class-set boundary in this module, which is exactly the condition #6182's fail-fast was written for. The first genuinely informative run of this test will be the post-rebase one.

### 2. The fix itself — I agree with it, including the `mergeReference` rejection

`resource = ""` means `AuthorizeInterceptor` stamps `PendingApproval.resourceId = null` (`AuthorizeInterceptor.kt:308,327`), and `satisfies` compares `this.resourceId == resourceId` (`:350-354`) — so `null == null` is true for **any** approval of the same action by the same maker. Both `transaction.reverse` and `transaction.sweep` are `four_eyes_required`. The finding is real.

Binding sweep to `#request.idempotencyKey` rather than `mergeReference` is the right call and the stated reason is correct: one merge reference spans every account pair in the merge, so an approval for one pocket would satisfy another. The idempotency key is what the maker replays on the `X-Approval-Id` retry, so the approval resolves for exactly the request it was granted for.

### 3. One thing I want the rebased run to actually prove

`AuthorizeInterceptor.kt:443` logs *"resource field extraction failed for '%s': %s — falling back to no resource"* — i.e. a dotted-path expression that fails to resolve **degrades silently to precisely the defect being fixed**. `#request.idempotencyKey` depends on the parameter being named `request` (it is — `mergeSweep(request: MergeSweepRequest, ...)`) and on Kotlin reflection resolving the nested property.

The test does guard this: step 2 asserts `resourceId == KEY_A`, not merely `isNotNull`. Good — that is the right assertion, and the KDoc is right that `isNotNull` would also pass against a wrong-but-present binding. But the guard is only worth what the run is worth, and this test has never completed a green run. Please confirm on the rebased branch that **step 2 passes**, not just that the job is green — a boot failure reports ITs as **SKIPPED**, and `N tests / M skipped / 0 failures` scans as a pass.

### 4. Checked, found nothing

- `FourEyesProfile.getConfigOverrides()` uses **literals** only — avoids the two-classloader companion-object trap where the profile and the test class initialise separately.
- The `AlwaysFourEyesPdp` stub gates `transaction.sweep` **only**, and the KDoc's explanation of why is correct and non-obvious: a stub gating every action makes `transaction.approval.decide` itself require an approval, and step 3 — the negative that is the entire control — then passes **vacuously**. That is a genuinely good piece of test design and I want it on record that it is load-bearing, not tidiness.
- Step 4 as a positive control is the right shape (a gate that rejects everything would pass step 3), even though `isNotEqualTo(202)` is a weak assertion; the KDoc explains why a stronger one is not available and I accept that.
- Threat model `docs/threat-models/openbank-transaction-service.md` is updated with a dated §7 entry. transaction-service is money-path, so this still needs **2 approvals**.
- `transaction.reverse` -> `#transactionId` matches the existing `transaction.read` binding on the same path parameter (`TransactionResource.kt:137` on `main`), so the expression form is already proven in this file.

### Verdict

**Needs change: rebase only.** The substance is correct and the test is better than most in this tree. Do not merge on the current CI state — it is measuring a broker, not the binding.

JiRaska:
## `build (openbank-transaction-service)` — diagnosed: the event loop is parked, the request never arrives

**This is not the old #5940 dead-broker symptom.** This branch's merge-base is `a1b0595d7` (today 10:29Z), which already contains #6182 (`50da682f1`, the Kafka fail-fast). The `SocketTimeoutException` is a different mechanism, and it is reproducible from the log rather than inferred.

### Decisive timeline (run 32567909999)

```
10:40:24.597  four-eyes: approval id=01a0290e-… not valid for action=transaction.sweep
              maker=…099 (…) — re-issuing a pending approval
10:40:27.694  Thread[vert.x-eventloop-thread-0 @coroutine#10,5,main] has been blocked
              for 3001 ms, time limit is 2000 ms
              io.vertx.core.VertxException: Thread blocked
                at kotlinx.coroutines.BlockingCoroutine.joinBlocking(Builders.kt:57)
                at com.openbank.libs.authz.AuthorizeInterceptorKt.awaitOffEventLoop(AuthorizeInterceptor.kt:480)
                at com.openbank.libs.authz.AuthorizeInterceptor.requireFourEyesOrProceed(AuthorizeInterceptor.kt:312)
                at com.openbank.libs.authz.AuthorizeInterceptor.authorize(AuthorizeInterceptor.kt:216)
10:40:54.696  MergeSweepApprovalBindingIT > the approval still satisfies the exact sweep
              it was issued for() FAILED
              java.net.SocketTimeoutException at MergeSweepApprovalBindingIT.kt:224
```

`150 tests completed, 1 failed` — this one test, no skips.

### What is actually happening

`awaitOffEventLoop` is:

```kotlin
private fun <T> awaitOffEventLoop(block: suspend () -> T): T = runBlocking {
    withContext(Dispatchers.IO) { block() }
}
```

The `withContext(Dispatchers.IO)` hop moves the *work* off the loop, but the outer `runBlocking` still **parks the calling thread**, and the caller here is `vert.x-eventloop-thread-0`. Vert.x reports it blocked for 3001 ms. Step 4's POST then never gets served on that loop and RestAssured hits its socket timeout 27 s later.

The ordering matters and rules out the obvious alternative reading: there is **exactly one** `four-eyes: approval … not valid` line in the whole job, at 10:40:24. That is step 3's *expected* negative (approval A replayed against sweep B — correct, and it does not consume the approval). **Step 4 emitted no four-eyes line at all**, i.e. it never reached the interceptor. So this is not an assertion failure and not a four-eyes binding defect — the request simply was not served.

### Why your PR surfaced it

Nothing in this diff is at fault. `AuthorizeInterceptor.kt` lives in `openbank-libs-runtime` and is unchanged here. But `resource = ""` → `resource = "#request.idempotencyKey"` creates, for the first time, a *mismatch* path: step 3 now genuinely fails `satisfies(…)` and falls through to `store.find` + re-issue, so two four-eyes replays run back-to-back through `awaitOffEventLoop`. `MergeSweepApprovalBindingIT` appears to be the first test in the fleet that drives that sequence. The binding change itself is doing exactly what it says — steps 1–3 pass.

### Why I did not fix it

The fix belongs in shared `openbank-libs-runtime` authz (money-path adjacent, every service's request path), not in this PR. It is your call which way to go:

1. **Fix the libs shape** — make `requireFourEyesOrProceed` suspend end-to-end so no `runBlocking` sits on the loop. Correct, but a separate PR against libs, and it re-tests the whole fleet.
2. **Make the test not depend on a single event loop** — raise the test profile's event-loop pool size, or raise the RestAssured socket timeout for this IT, and file the libs block as its own issue.

Option 2 unblocks this PR without changing shared authz; option 1 is the real fix. Worth noting either way that a blocking-for-3 s authz interceptor on the event loop is a live production property, not just a test artifact.

No changes pushed to this branch.

Refs #4754

JiRaska:
I reran the failed job to test whether `MergeSweepApprovalBindingIT` was flaky. **It is not** — the
rerun failed worse, and the shape of the failure rules out the runner.

Attempt 1 (10:40Z) failed one test. Attempt 2 (13:06Z) failed all four:

```
13:06:35  a four-eyes sweep is paused and mints a pending approval        SocketTimeoutException :224
13:07:05  the stored approval is bound to the specific sweep              SocketTimeoutException :149
13:07:35  an approval issued for sweep A does not satisfy sweep B         SocketTimeoutException :224
13:08:05  the approval still satisfies the exact sweep it was issued for  SocketTimeoutException :224
```

Three things worth reading off that:

1. **Exactly 30s apart, to the second.** That is a client read timeout expiring, not variable load.
   Every request to `POST /api/v1/transactions/merge-sweep` hangs until the socket gives up.
2. **`150 tests completed, 4 failed`** — the other 146 in the module pass in the same JVM, and the
   log has no `BindException`, no `Failed to start`, no unsatisfied injection. The app is healthy;
   this one endpoint does not answer.
3. **Step 1 hangs too.** The KDoc explains at length why *step 4* could hang (it reaches the Temporal
   workflow and a real balance-service call) and adds `LedgerWireMockResource` to prevent it. But
   step 1 only expects a 202 from the four-eyes gate and never reaches the workflow — so whatever
   this is, it is upstream of the explanation already written into the test.

What I ruled out, so nobody repeats it:

- `LedgerWireMockResource` **is** declared on the class and **does** exist on the branch, and it
  stubs both `quarkus.rest-client.ledger-service.url` and `quarkus.rest-client.balance-service.url`
  at the same WireMock base, plus `/token`. The gap its comment describes is closed.
- `FourEyesProfile.getEnabledAlternatives()` **does** return `AlwaysFourEyesPdp`, so the profile is
  not silently falling back to the real OPA sidecar in the obvious way.
- This test has **never passed in CI**. It is new in this PR and the branch head has only ever had
  these two runs, so there is no green history to point at — "it was fine before" is not available
  as an explanation.

My leading hypothesis, untested: the interceptor resolves `resource = "#request.idempotencyKey"` by
Kotlin reflection over the intercepted method's parameters, and `mergeSweep` is a `suspend fun`, so
the compiled signature carries a trailing `Continuation`. If that resolution blocks or the
coroutine is never resumed, the request never completes and the client times out — which fits all
four steps failing identically regardless of how far into the flow they should get. The KDoc itself
says the reflection path is the reason this had to be an IT rather than a unit test, so it is the
part with the least coverage behind it.

A cheap way to discriminate: temporarily make `mergeSweep` a plain `fun`, or point `@Authorize` at a
constant `resource`, and see whether the hang disappears. Whichever of those clears it names the
cause. Worth doing before touching anything else — the change is money-path and the test is the only
thing asserting that an approval cannot be replayed against a different sweep.

I have not changed the branch.

JiRaska:
**Triage (measured, no action taken).**

`build (openbank-transaction-service)` and `all-green` are **FAILURE**, but the run completed
`2026-08-24T15:30:03Z` and the branch is `behind_by=795` (`gh api repos/JiRaska/open-bank-oss/compare/main...fix/four-eyes-resource-binding-4754`).
The only annotation the API still returns is `Process completed with exit code 1`, alongside
`dependency warm-up timed out after 120s`, so the red **cannot currently be classified** as a real
defect versus the documented environmental build flake. Rebase onto current `main` and re-run before
spending review time on it.

No version collision: this branch does not touch `openapi.yaml`, and `version.txt` is unmodified
(branch carries `1.19.2` only because its base predates `main`'s `1.21.0` — release-please owns it,
and the merge keeps `main`'s value).

Blocked by rule regardless of CI: `openbank-transaction-service` is in
`rules.yaml: money_path_services`, so this needs **2 approvals + threat model** (the threat model
*is* updated in this diff). Currently 0 approvals.

JiRaska:
### The red check was a CI timeout, not a hang — measured, then the branch brought onto main

`build (openbank-transaction-service)` failed on 2026-08-24 with:

```
Execution failed for task ':openbank-transaction-service:test'
> Timeout has been exceeded
BUILD FAILED in 27m 37s
```

That is not evidence of anything in the diff, so I ran the module at this PR's head locally rather
than guessing:

```
BUILD SUCCESSFUL in 2m 28s
tests=151  failures/errors=0  skipped=1
MergeSweepApprovalBindingIT: 4 tests, 0 skipped, 20.1s
```

The IT this PR adds **ran** — worth stating explicitly, because a Quarkus IT that cannot start
reports as *skipped*, and a skipped test looks like a pass to anyone reading the failure count. Four
tests, none skipped.

**The branch was 1000 commits behind with zero conflicts**, so its last CI run tested a merge commit
from three weeks ago. Merged current `main` in, and re-ran on the merged result:

| | tests | failures | skipped |
|---|---|---|---|
| at the old head | 151 | 0 | 1 |
| merged with current main | 152 | 0 | 1 |

The skipped count is what matters here and it did not move. `main` added
`V17__synthetic_outbox_taint.sql` in the meantime, and two branches claiming one Flyway version is a
failure that surfaces *only* as silently skipped integration tests — so this is checked, not assumed.
`MergeSweepApprovalBindingIT` still runs, 0 skipped, 19.7s.

Fresh CI is running against what would actually be merged.

### What this PR actually fixes, since it has been open since 16 August

```diff
- @Authorize(action = "transaction.reverse", resource = "")
+ @Authorize(action = "transaction.reverse", resource = "#transactionId")

- @Authorize(action = "transaction.sweep", resource = "")
+ @Authorize(action = "transaction.sweep", resource = "#request.idempotencyKey")
```

Both verbs are `four_eyes_required`, and the interceptor stamps the `PendingApproval` with
`query.resource?.id`. With an empty resource every approval was created and matched at
`resourceId=null` — so a checker's approval to reverse transaction A satisfied a reversal of
transaction B by the same maker, and an approval granted for one merge sweep satisfied a different
sweep, to different accounts, for a different amount. Same defect class as #3521 (`sanctions.clear`).

The choice of `idempotencyKey` over `mergeReference` for the sweep is right and worth not losing in
review: one merge reference covers every account pair swept in that merge, so binding to it would let
an approval for one pocket satisfy the sweep of another.

No `openapi.yaml` change is needed — the request and response shapes are untouched; what changes is
which approval satisfies which request.

This is money-path and stays a human's call to merge. It is simply no longer red for a reason that
had nothing to do with it.

JiRaska:
### Correcting myself: this is not an infrastructure timeout, and the committed diagnosis is not supported either

My earlier comment said the red check was "a CI timeout, not a hang" and implied it was unrelated to
the diff. The fresh run disproves that, and it also disproves the explanation this PR already carries.

**What actually happens, identically in both failing runs:**

```
MergeSweepApprovalBindingIT > the approval still satisfies the exact sweep it was issued for() FAILED
    java.net.SocketTimeoutException at MergeSweepApprovalBindingIT.kt:224
...
> Task :openbank-transaction-service:test FAILED
> Timeout has been exceeded
BUILD FAILED in 26m 19s
```

Line 224 is the RestAssured `POST /api/v1/transactions/merge-sweep` itself. **The request does not come
back.** The 25-minute `timeout.set(...)` on the test task then fires — so the timeout is a
*consequence* of the hang, not the failure. One test, the same one on 2026-08-24 and on 2026-09-01,
and it is the **positive control** of the four-eyes binding: the case where the replayed approval
*should* be accepted.

**The OOM explanation in `build.gradle.kts` is not what the logs show.** That comment attributes the
`SocketTimeoutException` to GC thrash from multiple Quarkus boots in one forked JVM, and
`maxHeapSize = "2g"` was added on that basis. Grepping both failing runs:

| | 2026-08-24 | 2026-09-01 |
|---|---|---|
| `OutOfMemoryError` | **0** | **0** |
| `Java heap space` / `GC overhead limit` | 0 | 0 |
| `SocketTimeoutException` | 1 | 1 |

Also worth pinning down, because it looked like the fix had never been tried: the 2026-08-24 run's
`headSha` is `0164a8a49`, which **is** the commit that raises the heap. So the run that failed already
had `maxHeapSize = "2g"`, and it failed the same way. The heap bump may be independently reasonable,
but it is not the fix for this and the evidence for its stated cause is absent.

The `java.lang.RuntimeException: boom` further down the log is expected output from
`PaymentActivitiesImplTest.reverseJournal swallows ledger exceptions`, not a failure — worth saying so
that nobody chases it.

**It passes locally.** Same head, twice, `maxHeapSize = "2g"` in effect:

```
BUILD SUCCESSFUL in 2m 28s
tests=151  failures=0  skipped=1
MergeSweepApprovalBindingIT: 4 tests, 0 skipped, 20.1s
```

So the difference is environmental, and the useful question is no longer "why is it slow" but **what
does `POST /merge-sweep` block on in CI that it reaches locally**. The endpoint's approval path is the
place to look: the negative case in the same class does not hang, only the replay-with-`X-Approval-Id`
case does. A likely shape is a dependency the approval branch reaches (Temporal, ledger, or the
approval store) that is present in the local run and absent or unreachable in the CI service
container — in which case the honest fix is either to stub it in the IT or to make the endpoint answer
rather than block when it is missing, since an endpoint that hangs on a missing dependency is a defect
in its own right.

Raising the task timeout would only make the failure take longer to report.

Everything else I said stands: the four-eyes binding defect this PR fixes is real and serious, the IT
genuinely runs rather than being skipped, and merging current `main` in changed nothing about the
skipped count (151/0/1 → 152/0/1).

JiRaska:
### Narrowing the hang, and where I have to stop

Following the previous comment, the blocking point is identifiable from the code with no guessing.

`TransactionService.initiateTransaction` ends in a **blocking Temporal call that waits for the whole
workflow**:

```kotlin
// stub.execute(...) is a blocking Temporal client call; offload the blocking wait to the IO dispatcher.
val stub = workflowClient.newWorkflowStub(PaymentWorkflow::class.java, ...)
stub.execute(saved.id)
```

That is exactly the line dividing the passing tests from the failing one. The negative cases never
reach it — the four-eyes interceptor answers `PENDING_APPROVAL` first — so they are fast and green.
The **positive** case is the only one that clears the gate, enters `initiateTransaction`, and blocks
on `PaymentWorkflowImpl` → `PaymentActivitiesImpl.placeHold`. The IT's own KDoc predicts this failure
almost word for word, including "on some runners that connects rather than failing fast with
ECONNREFUSED and then sits without a response".

**But the wiring that KDoc prescribes is present and complete**, which is what I checked and what
rules out the obvious answer. `LedgerWireMockResource.start()` arms both stub sets before returning
its config, not via a helper a test must remember to call:

```kotlin
stubLedgerSuccess()
stubBalanceCoverSuccess()
...
"quarkus.rest-client.ledger-service.url"  to base,
"quarkus.rest-client.balance-service.url" to base,
```

So both the journal and the hold/release legs are stubbed by default, both client URLs point at
WireMock, and the OIDC token call is stubbed too. `stubBalanceCoverSuccess` has exactly one caller in
the module and it is that line — I checked, because "the helper is never called" was my first
hypothesis and it was wrong.

**That is where the repository stops answering.** The difference between this Mac (2m28s, green,
twice) and the CI runner is environmental, and nothing in the tree distinguishes them.

**Cheapest probe that would settle it**, and it costs one line in the resource's failure path:
WireMock knows whether the request ever arrived. On a failed run, dump

```kotlin
server.findAllUnmatchedRequests()      // arrived but matched no stub
server.allServeEvents                   // arrived at all?
```

Those three states have different fixes and are indistinguishable from the outside today:
the hold request never reached WireMock (URL/DNS/binding on the runner), it reached it and matched no
stub (a path or method the `urlPathMatching` does not cover on that platform), or it matched and the
workflow still did not complete (a Temporal-side wait, not an HTTP one). Right now the only signal is
a client-side `SocketTimeoutException`, which is consistent with all three.

Raising the 25-minute task timeout would change nothing except how long the report takes.

JiRaska:
### Ruled out the most plausible theory, so nobody else spends time on it

The obvious explanation for `SocketTimeoutException` on `POST /merge-sweep` in CI but not locally was
CDI alternatives. `FourEyesProfile` overrides `getEnabledAlternatives()` with a single class:

```kotlin
override fun getEnabledAlternatives() = setOf<Class<*>>(AlwaysFourEyesPdp::class.java)
```

and the in-process Temporal environment comes from another alternative:

```kotlin
@ApplicationScoped @Alternative @Priority(1)
class WorkflowClientTestProducer(...)   // starts TestWorkflowEnvironment, NoOpPaymentWorkflow
```

If naming one alternative dropped the other, the real Temporal client would take over,
`stub.execute(saved.id)` would wait on a worker that is not there, and the request would never
return — which fits the symptom exactly.

**It does not happen.** Probed it directly by injecting `WorkflowClient` under that profile:

```
PROBE workflowClient=io.temporal.client.WorkflowClientTestProducer_ProducerMethod_workflowClient_...ClientProxy
PROBE options=default
```

The test producer is still in effect. `@Priority` enables an alternative globally, and
`getEnabledAlternatives()` does not revoke that. Theory dead.

### What that leaves

The blocking call is still the discriminator worth keeping in mind — the four-eyes **negative** cases
answer `PENDING_APPROVAL` in the interceptor and never reach the use case, so only the positive case
enters `initiateTransaction` and only it hangs. But the thing it blocks on is not a missing workflow
worker.

Next probe, and it is the one that separates the three remaining states a client-side
`SocketTimeoutException` cannot:

```kotlin
server.findAllUnmatchedRequests()   // arrived, matched nothing
server.allServeEvents               // arrived and was served
// neither -> never left the client / blocked before the outbound call
```

Dumping those from an `@AfterEach` on failure would make the next CI run say which it is, instead of
repeating a 25-minute timeout that reports only that something did not come back.

`PostgresRedpandaTestResource` is the other input I have not eliminated — a sweep that publishes
before responding would block on a broker that is slow to come up in CI and fine locally.

I have not changed the test: adding the diagnostic is a small edit but it is your call which of those
two to instrument first.

JiRaska:
## The four failures are a connection-pool exhaustion, not a defect in the assertions

All four `MergeSweepApprovalBindingIT` steps failed, and the step-1 failure is a plain
`java.net.SocketTimeoutException` on the RestAssured call. That is the symptom, not the cause.
The cause is one line further down the same log:

```
[Exception 0] org.hibernate.HibernateException: io.vertx.core.impl.NoStackTraceThrowable: Timeout
	at org.hibernate.reactive.pool.impl.SqlClientPool$ProxyConnection.lambda$connection$0(SqlClientPool.java:282)
```

The reactive pool did not hand out a connection inside its timeout. Two things confirm this is
about the pool rather than about this test:

1. **The four failures are spaced at exactly 30 s** — 10:29:00, 10:29:31, 10:30:01, 10:30:31.
   That is a fixed timeout being consumed in full, four times, not four different amounts of work.
2. **`StuckSagaGauge#refresh` failed on the identical stack** in the same window
   (`Error occurred while executing task for trigger IntervalTrigger [... interval=30000]`).
   That scheduler is not part of this test and contends for the same pool, so whatever starved
   the request starved it too.

The module ran `203 tests, 4 failed` and the build was ultimately killed by Gradle with
`Timeout has been exceeded` after `26m 53s`.

## Against the local run

Run locally on this branch, the same four tests pass — read from the JUnit XML rather than the
console, because a filtered Gradle run that executes nothing also prints `BUILD SUCCESSFUL`:

```
tests=4 failures=0 errors=0 skipped=0 time=162.179
```

So the assertions hold. But 162 s for four tests is itself the signal: this IT is expensive,
it adds another `@TestProfile` (and so another Quarkus restart) to a module already at 203 tests,
and under CI contention the pool wait is what gives first.

## What this means for the PR

This is not a flake to be re-run until green — it has now gone red in CI twice on the same
mechanism. Re-running once more to confirm determinism, but the honest options are to give this
test its own pool headroom (or disable the competing scheduler in `FourEyesProfile`, which is the
only bean in the contention that the profile already controls) rather than to retry it.

Not changing the assertions: the negative in step 3 is the whole control and it is correct.

JiRaska:
## Fixed, and the measurement identifies the contender rather than guessing at it

`FourEyesProfile` now sets `quarkus.scheduler.enabled = false`. Measured on this branch, read
from the JUnit XML rather than the console:

```
before:  tests=4 failures=0 errors=0 time=162.179   (and red in CI, twice)
after:   tests=4 failures=0 errors=0 time=11.443
```

A 14x drop is the evidence. The time was not work — it was four requests waiting on a connection
the three DB-touching schedulers in this module (`StuckSagaGauge`, `TransactionOutboxBacklogGauge`,
`TransactionOutboxDispatcher`) were holding. `StuckSagaGauge` runs `every = "30s"`, which is
exactly the spacing of the four CI failures.

Scoped to this profile only. Disabling the scheduler module-wide in tests is how a dead scheduler
becomes invisible, which this repo has already paid for once (#2148).

No assertion changed.

Worth noting for the queue generally: this looked like a flake and was not one. It had gone red
in CI twice on the same mechanism while passing locally, and "passing locally" was itself the
misleading part — 162s for four tests was the defect showing, not the test being thorough.

JiRaska:
## `MergeSweepApprovalBindingIT` — not environmental. It is the one step that actually executes.

I called this environmental last time on the strength of a `SocketTimeoutException` plus a task
timeout. That was wrong, and the second run is what falsified it: same test, same failure, twice.

### What is established

1. **The failing test is branch-only.** `MergeSweepApprovalBindingIT` does not exist on `main`, so
   this is the PR's own integration test, not a fleet flake it inherited.
2. **Only `@Order(4)` fails.** Steps 1–3 pass:

   | step | what it asserts | reaches the sweep? |
   |---|---|---|
   | 1 | a four-eyes sweep is paused and mints a pending approval | no — 202 |
   | 2 | the stored approval is bound to the specific sweep | no — read-only |
   | 3 | an approval for sweep A does not satisfy sweep B | no — refused |
   | 4 | **the approval still satisfies the exact sweep it was issued for** | **yes** |

   Step 4 is the only case whose request gets past the four-eyes gate and reaches
   `transactionUseCase.initiateTransaction(command)` in `TransactionResource.mergeSweep`. Every
   refusal path answers fine; the execution path hangs.
3. **The symptom is a hang, not an assertion failure.** `java.net.SocketTimeoutException` at
   `MergeSweepApprovalBindingIT.kt:223`, which is the `.post("/api/v1/transactions/merge-sweep")`
   inside `postSweep`. The task then runs on to its 25-minute Gradle timeout (`BUILD FAILED in
   26m 55s`, `255 tests completed, 1 failed`) — consistent with the #5940 shape that timeout was
   written for: the JVM will not exit after the tests finish.

### Why the existing mitigation is not the answer any more

`FourEyesProfile` already carries a fix for this exact symptom, and its comment says so:

> A SocketTimeoutException at the RestAssured call is the symptom; the starved connection is the
> cause.

That was three `@Scheduled` beans contending for the reactive pool, and the profile disables the
scheduler for it. It worked — back then **all four** steps failed at 30-second spacing; now three
pass and only the executing one fails. So the remaining cause is a different one.

### Hypothesis, and the one thing that would confirm or kill it

`TransactionService` injects `WorkflowClient` and `TemporalConfig`, and `initiateTransaction` is the
entry to the payment workflow. A `@QuarkusTest` has no Temporal server, so a workflow start on that
path has nothing to answer it — which would produce exactly a hung request followed by a JVM that
will not exit.

Related and independently confirmed while checking: `PaymentWorkerRegistrar` carries the switch
`openbank.transaction.worker.enabled` with the comment

> Tests set openbank.transaction.worker.enabled=false.

**No test sets it.** Not in `src/test/resources/application.properties`, not in either
`QuarkusTestProfile` in this module. That comment is a claim about configuration that has gone
stale — the same class of thing as a KDoc asserting a DLQ that is not wired.

To confirm or kill the hypothesis in one run, add to `FourEyesProfile.getConfigOverrides()`:

```kotlin
"openbank.transaction.worker.enabled" to "false",
```

If step 4 still hangs, Temporal is not it and the next suspect is whatever `initiateTransaction`
awaits after the idempotency lookup. If it passes, the comment above should become true for this
module rather than merely aspirational.

I have not run this — the module's test task takes ~25 minutes and the CI queue is several hundred
jobs deep, so I am not spending a run on a guess I can hand you as a one-line change. The
discriminator in point 2 is the solid part; the Temporal link is the hypothesis.

JiRaska:
## Result: the Temporal hypothesis is not supported — and the experiment has a measurement gap I should name

Disabling `openbank.transaction.worker.enabled` in `FourEyesProfile` changed **nothing**. Same test,
same failure, same shape:

```
MergeSweepApprovalBindingIT > the approval still satisfies the exact sweep it was issued for() FAILED
    java.net.SocketTimeoutException at MergeSweepApprovalBindingIT.kt:236
255 tests completed, 1 failed
Requesting stop of task ':openbank-transaction-service:test' as it has exceeded its configured timeout of 25m.
BUILD FAILED in 26m 9s
```

(Line 236 rather than 223 is my added comment block; same call — the `.post(...)` in `postSweep`.)

So the change did its job as a falsifier: **Temporal worker registration is not the cause.**

### The gap, stated rather than glossed

I cannot prove from this run that the knob took effect. `PaymentWorkerRegistrar` logs
`Temporal payment worker registration disabled` only when its bean is constructed, and the bean is
lazy — so the log says nothing in either state, and indeed the string appears **zero** times.

The Temporal frames that do appear in the log are not it either: they come from a deliberate fixture
(`ApplicationFailure: message='ledger down', type='LedgerError'`), i.e. a workflow-compensation test
driving its own test environment, not the registrar.

So the honest reading is "hypothesis unsupported, and the knob is unobservable", not "hypothesis
disproved". Making the registrar `@Startup` — or logging at registration either way — would close
that, and is worth doing regardless: a switch whose position cannot be read from a log is one nobody
can verify in a cluster either.

### What the run does establish

Two failures are present and they are not the same thing:

1. **Step 4 hangs.** It is still the only step that reaches
   `transactionUseCase.initiateTransaction`, and it still times out at the HTTP call rather than
   asserting a wrong status. The discriminator from my earlier comment stands.
2. **The task does not finish.** 255 tests complete and the task still burns to 25 minutes — the
   #5940 shape the timeout exists to bound. That is a separate problem from the one failing test,
   and disabling the worker did not move it either.

### Suggested next suspect

Whatever `initiateTransaction` awaits after the idempotency lookup, with the approval already
satisfied — i.e. the first thing the approved path does that the refused paths never reach. A thread
dump at the moment of the socket timeout would name it directly; a targeted run of just this class
(`--tests '*MergeSweepApprovalBindingIT'`) would also finish in minutes instead of 26 and make the
next experiment cheap.

I am leaving the `worker.enabled` override in place: it is correct on its own terms (the registrar's
comment claimed tests set it and none did), it costs nothing, and removing it would re-open a
variable this run has already ruled out.

JiRaska:
## The `Services CI` red is reproducible, and it is not what either of us has been chasing

I reran the failed job (attempt 2, job `103583266539`): **failed again**, so it is not a flake. But
the failure shape CHANGED between attempts, and that is the informative part.

| attempt | MergeSweepApprovalBindingIT |
|---|---|
| a1 | 1 of 4 failed — only `the approval still satisfies the exact sweep it was issued for` |
| a2 | **4 of 4 failed**, each `java.net.SocketTimeoutException: Read timed out` at **30.0–31.3 s** |

Step 1 (`a four-eyes sweep is paused and mints a pending approval`) fails the same way in a2, and
that step never gets past the four-eyes gate — it touches no workflow, no Temporal, no ledger. So
"step 4 reaches `initiateTransaction` and `stub.execute` has no worker to answer it" cannot be the
cause of a2: **every** request to the endpoint hangs, including the one that only needs a 202.

I also could not reproduce it locally, in a way that is itself evidence:

```
./gradlew :openbank-transaction-service:test --tests '*MergeSweepApprovalBindingIT*'
  -> 4 tests, 0 failures   (worker disabled, exactly as the profile sets it)
```

The class passes when it is the ONLY class in the run. (The full-module local run is not a
counter-measurement — it died of `OutOfMemoryError: Java heap space` in the test JVM on this
machine, after 232 tests with 0 failures.)

## What the a2 report actually shows

The first line of that class's captured stdout, before its own boot:

```
16:54:33.790 INFO  PostgresRedpandaTestResource
  Stopping Redpanda test container (bootstrap=PLAINTEXT://localhost:32771)
  — producers from the current Quarkus boot will fail to reach it after this point
16:54:41-45  (this IT's Quarkus boot)
16:54:33 + ~30s per step -> four SocketTimeouts
```

`PostgresRedpandaTestResource.stop()`'s own comment names the mechanism (#5940): *"Quarkus calls
`stop()` whenever the set of `@QuarkusTestResource` classes changes between test classes, so this
runs mid-run, not only at the end"*. The class that ran immediately before, per the same report, is
`TransactionNegativeAuthPactVerificationTest` (16:53:54).

So the signature is a **shared-test-resource transition**, not the subject under test:
uniform ~30 s read timeouts on every request of the first class after the switch, passing when the
class runs alone. That also explains a1-vs-a2 differing: which step is in flight when the resource
goes away depends on interleaving.

## Why I am not pushing a fix

Three reasons, and the first is the one that matters: the remedy is a test-infrastructure decision
about resource/profile ordering in this module, and picking it wrong trades a red for a test that
passes without exercising anything. Secondly this is a money-path service and the merge is a
human's regardless. Thirdly the branch has been idle since 10:39 and I do not push to a branch I do
not own.

What I would check first, in order:

1. whether the **profile set** of the classes around it forces the resource restart — anything that
   makes `MergeSweepApprovalBindingIT` the first class after a `@QuarkusTestResource`/profile change
   inherits this, so pinning the resource for the whole module (one resource set, no mid-run
   transition) is the fix that does not weaken anything;
2. whether the 30 s is the reactive pool's acquisition timeout rather than RestAssured's — the
   profile comment already records `SqlClientPool$ProxyConnection ... Timeout` at 30 s intervals for
   the earlier incarnation of this red, which is the same number;
3. `quarkus.scheduler.enabled=false` (this branch's earlier fix) is **necessary but not sufficient** —
   a2 shows the scheduler correctly disabled (`Temporal payment worker registration disabled`,
   no scheduler noise) and the timeouts unchanged.

Happy to take it further if you want the fix authored — say the word and I will do it on a branch of
mine rather than on this one.

JiRaska:
## Found it, fixed it on a branch of mine — #9836

The cause is not in this PR's subject at all: **the forked test JVM runs on Gradle's default heap.**
`org.gradle.jvmargs=-Xmx3g` and `GRADLE_OPTS` both size the *daemon*; a `Test` task forks its own
JVM, and this module sets no `maxHeapSize` while nine siblings do.

Both directions, full `:openbank-transaction-service:test` on this branch:

| heap | outcome |
|---|---|
| default | `OutOfMemoryError: Java heap space`, FAILED after 232 tests |
| `2g` | **256 tests, 0 failures** — including `MergeSweepApprovalBindingIT` |

Control: `main` passes at the default (252 tests, 0 failures), so this is a ceiling the module was
already sitting on, not a leak your change introduced. Your added IT — one more profiled
`@QuarkusTest`, i.e. one more Quarkus boot with three containers — is simply what goes over it.

`maxHeapSize = "2g"` is in **#9836**, off `main`, so it can land independently of this PR; merge it
and then re-run here.

## Corrections to what I posted earlier on this PR

Two of my three diagnoses were wrong and I should not have offered the first as confidently as I did:

1. **"worker=false, so `stub.execute` blocks with no Temporal worker"** — false. The class run alone
   passes with the worker disabled exactly as your profile sets it (4 tests, 0 failures).
2. **"the `@QuarkusTestResource` set changes and the container is torn down mid-run"** — not
   reproducible: running `TransactionNegativeAuthPactVerificationTest` then
   `MergeSweepApprovalBindingIT` in the CI order gives 4+4, 0 failures, the IT in 11 s. The
   `Stopping Redpanda test container` line is real and is the transition, but it is not what makes
   the tests fail.

What actually discriminated was comparing the two ATTEMPTS of the same job: a1 failed 1 of 4, a2
failed **4 of 4** at a uniform 30.0–31.3 s. A step that only needs a 202 from the four-eyes gate
timing out identically rules out every story about what the endpoint does — that is a resource-level
failure, and an exhausted fork is exactly what it looks like.

Your `quarkus.scheduler.enabled=false` and `openbank.transaction.worker.enabled=false` both stand on
their own merits (and a2's log confirms the worker registration is skipped) — they just were not the
remaining cause.

JiRaska:
## The remaining red is not in this PR — it is the four-eyes gate blocking an event-loop thread (#9874)

Reproduced deterministically, which the earlier attempts never managed, by giving the JVM as few CPUs
as a GitHub-hosted runner:

```
jvmArgs("-XX:ActiveProcessorCount=2")
./gradlew :openbank-transaction-service:test --tests '*MergeSweepApprovalBindingIT*'
  -> 4 tests completed, 4 failed; "> Timeout has been exceeded"; BUILD FAILED in 25m 11s
```

`jstack` during the hang — both of the two event-loop threads that exist at 2 CPUs, same frame:

```
"vert.x-eventloop-thread-0"  TIMED_WAITING (parking)
  kotlinx.coroutines.BlockingCoroutine.joinBlocking
  AuthorizeInterceptorKt.awaitOffEventLoop(AuthorizeInterceptor.kt:516)
  AuthorizeInterceptor.requireFourEyesOrProceed(AuthorizeInterceptor.kt:348)
  AuthorizeInterceptor.authorize(AuthorizeInterceptor.kt:229)
  TransactionResource_Subclass.mergeSweep
"vert.x-eventloop-thread-1"  — identical
```

`awaitOffEventLoop`'s `runBlocking` parks the event-loop thread it was called on; the reactive
approval-store call inside still needs *an* event-loop thread, and that pool is CPU-sized. Many-core
laptop: a spare loop always exists, everything passes in 1.5 s. Two-CPU runner: the first four-eyes
request parks loop 0, the second parks loop 1, and nothing can complete after that.

So your IT is doing its job — it is the first test in the fleet to drive the four-eyes gate over real
HTTP, and it found a latent defect in `openbank-libs-runtime` on the money path. Filed as **#9874**
with the reproducer and two candidate fixes. This PR cannot go green until that lands.

## Trail of what this cost, since it is instructive

Four hypotheses, each killed by measurement rather than argument:

1. worker disabled ⇒ `stub.execute` blocks — the class passes alone with the worker off.
2. `@QuarkusTestResource` set transition tearing containers down mid-run — the CI-order pair passes,
   4+4.
3. the profile's `getEnabledAlternatives()` disabling `WorkflowClientTestProducer`, so the real
   Temporal client dials 7233 — an unroutable target still passes: `@Alternative` **with `@Priority`**
   is enabled globally in Quarkus, and `getEnabledAlternatives()` only governs alternatives without
   one.
4. an idempotency-key collision — the `duplicate key ... 23505` line in the job log sits next to
   `Cannot reverse a REVERSED transaction`, i.e. it belongs to the reversal tests.

And one trap worth naming: after the 2-CPU run I first read `failures=0` out of
`build/test-results/` and nearly called it a pass — the timed-out task writes no XML, so I was
reading the previous run's file. The identical millisecond durations were the only tell.

PR #5072, opened 2026-08-16
fix(accounts): refuse to close an account that still holds money

JiRaska:
## Summary
- `closeAccount` had **no balance guard** — nothing stopped an account with real money from being closed. Adds `AccountNotEmptyException` (422), checked after status validity, before persistence.
- Threads `customerPartyId` (`X-Customer-Party-Id`) + ownership check into the REST endpoint, same pattern as the existing goal/nickname endpoints, so `requestedBy` reflects the real caller.

## Deliberately out of scope
No customer-edge close endpoint in this PR. `account_rest_ext.rego` explicitly **prohibits** `account.close` for the edge's M2M principal — closed off after fleet audit #3734 found a role-only grant had accidentally exposed the whole sensitive lifecycle (close/freeze/unfreeze/authorize) through this exact proxy path. Reopening that is a deliberate security-policy call for a human, not something to change from an agent session. Self-service "close my account" in the app stays unbuilt until that decision is made explicitly.

## Note for reviewer

A code review of this PR (before merge, not after) surfaced three things worth your explicit judgment call rather than being silently swept under the diff:

1. **Money-path approval gate (CLAUDE.md rule 7):** `openbank-account-service` is in `rules.yaml: money_path_services`. This PR changes account-lifecycle authorization semantics and adds a new money-adjacent guard — per the repo's own rule that needs **2 approvals + a threat model doc** (`docs/threat-models/openbank-account-service.md`, ADR-0030). No such doc exists/was touched here; I did not fabricate one — that's a human call on whether this change's blast radius warrants writing it now or is covered by an existing assessment.
2. **The balance guard is best-effort, not a guarantee** — documented explicitly in code (see the `KNOWN LIMITATIONS` comment on `closeAccount` and the `AccountNotEmptyException` KDoc, added in the follow-up commit):
   - **Point-in-time race:** the balance is read, then the account is persisted CLOSED — no lock ties the two together, so a credit settling in that window still lands on a closed account. `AccountUpdateConflictException`'s optimistic version check doesn't cover this (it's account-service's own row version, not balance-service's state).
   - **In-flight money is invisible today:** the guard only sees currently-booked/reserved balance. A future-dated standing order or a transfer sitting on its documented value-date delay reads as zero right now and still executes later against a closed account. Closing that gap needs a cross-service check (standing-order-service, sepa-instant, …) that doesn't exist yet — a real follow-up, not something this PR can fix alone.

I'm flagging these rather than claiming the guard is airtight, since "money that can still strand" is exactly the failure mode this PR exists to close.

## Test plan
- [x] New unit test: `closeAccount refuses to close an account that still holds money`
- [x] Existing close tests updated with the new `balancePort.getByAccount` stub
- [x] ktlint, detekt, `:openbank-account-service:test`, `:openbank-customer-edge:compileKotlin` all green

### DIFF
--- a/docs/threat-models/openbank-account-service.md
+++ b/docs/threat-models/openbank-account-service.md
@@ -83,8 +83,56 @@ not change any existing request's outcome until explicitly flipped.
 - Relies on Keycloak realm integrity and OPA policy correctness.
 - Freeze now has the four-eyes *mechanism* wired (§4a) but not enforced (`authz.four-eyes.enforce=false`);
   close remains single-actor — a candidate for a follow-up rollout under issue #413.
+- **`closeAccount`'s balance guard (2026-08-16 entry below) is best-effort, not a guarantee:**
+  (a) the balance check and the close persist are not atomic, so money settling in that window
+  still lands on a closed account; (b) it only sees currently-booked/reserved balance, not money
+  already in flight (a future-dated transfer reads as zero and still executes later). Closing (a)
+  needs a lock spanning balance-service and this service, or a saga; closing (b) needs a check
+  against every service that can schedule a future credit (standing-order-service, sepa-instant,
+  domestic-payment). Neither exists yet — tracked as a follow-up, not blocking this change since
+  it is a strict improvement over the prior state of no guard at all.
 
 ## 6. Change log
+
+- **2026-08-16** — Account close (TOP-10 #10, part 2): `POST /{accountId}/close` gains a balance
+  guard (`AccountNotEmptyException`, mapped to 422) and threads `customerPartyId` +
+  `denyIfNotOwner` through the endpoint the same way the goal/nickname endpoints already do, so
+  `requestedBy` on `AccountClosedEvent` reflects the real caller instead of always being stamped
+  as the operator. **No new trust boundary and no new edge route** — this endpoint stays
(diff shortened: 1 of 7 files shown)


JiRaska:
`gates (security)` is red on `threat-model-diff-gate` (ADR-0030 D2, enforced) — not a defect in the code change itself:

```
threat-model-diff gate: openbank-account-service: trust-boundary change without a
docs/threat-models/openbank-account-service.md update —
AccountResource.kt (inbound REST surface); ExceptionMappers.kt (inbound REST surface)
```

Both touched files are on the service's inbound REST surface, so the gate wants a dated `## 6. Change log` entry in `docs/threat-models/openbank-account-service.md` — same shape and same fix as PR #5043 hit this session. Recipe: append a dated entry describing the new `AccountNotEmptyException` guard (STRIDE-relevant: a caller could previously close an account holding real money — Tampering/data-loss class) and the `customerPartyId`/ownership threading, then re-run `python3 .github/scripts/run-gates.py --only threat-model-updated-on-trust-boundary-change` (with `PR_DIFF_BASE` set to the merge-base) locally to confirm before pushing.

Leaving this for whoever owns this branch/worktree rather than pushing directly — `git worktree list` shows this branch live at `account-part2`.

PR #5080, opened 2026-08-16
fix(agent): make oversight scheduler live

JiRaska:
Refs #3345

Converts the scheduled oversight sweep to the required suspend flow, registers ADR-0237 liveness only when it is enabled, and records success after a completed sweep. Removes it from the scheduler-liveness baseline.

Validation: scheduler-liveness ratchet and diff check pass; required CI is authoritative.

### DIFF
--- a/.github/scripts/check-scheduler-liveness.py
+++ b/.github/scripts/check-scheduler-liveness.py
@@ -69,7 +69,6 @@ ALLOWLIST = {
 # exit 1, which `mode: advisory` then converted to a warning; nothing else noticed.
 BASELINE = [
     # non-money-path
-    "openbank-agent-service/src/main/kotlin/com/openbank/agent/application/OversightService.kt",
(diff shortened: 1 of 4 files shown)


JiRaska:
`gates (kotlin)` is red on `no runBlocking in a @Scheduled body` (rules.yaml: scheduled_methods, enforced) — same shape as #5085's finding, different allowlist:

```
stale allowlist entry `scheduled_methods.runblocking_allowlist:
openbank-agent-service/src/main/kotlin/com/openbank/agent/application/OversightService.kt#scheduledSweep`
— that method no longer has runBlocking in a @Scheduled body (or no longer exists). Remove it.
```

This PR's own fix removed the `runBlocking` from `OversightService.scheduledSweep`, but the entry naming it in `rules.yaml: scheduled_methods.runblocking_allowlist` wasn't removed alongside it. Delete that one allowlist entry and the gate should clear.

Leaving this as a comment rather than pushing directly, in case the branch is still being worked.

PR #5085, opened 2026-08-16
fix(pid): track trusted-list refresh liveness

JiRaska:
## What changed

- registers the signed EUDI Trusted List refresh as an ADR-0237 workflow
- records success only after signature verification and trust-store replacement
- proves successful and rejected refresh paths in the focused unit test

## Why

A failed or inert trusted-list refresh must not advance scheduler liveness; only a verified, applied list is a successful control run.

## Linked issues

Refs #3345

## Validation

- `TrustedListServiceTest`: 6 tests, 0 failures
- targeted PID Gradle test, KtLint and Detekt invoked locally


JiRaska:
`gates (gitops)` is red on `scheduler liveness adoption` (ADR-0237) — not a defect in the fix itself, a bookkeeping gap:

```
baseline entries now compliant — REMOVE them from BASELINE in this script
(a baseline that never shrinks rots):
  openbank-pid-service/src/main/kotlin/com/openbank/pid/infrastructure/crypto/TrustedListService.kt
```

This PR's own fix made `TrustedListService.kt` compliant, but the gate's `BASELINE` allowlist (in whichever `.github/scripts/check-*scheduler*liveness*.py` implements this — grep for the constant name) still lists it as "adoption pending." Remove that one entry and the gate should clear — same "baseline that never shrinks" shape this repo's own notes warn about elsewhere.

Leaving this as a comment rather than pushing — `git worktree list` shows this branch live locally.

Issue #5091, opened 2026-08-16
observability: fleet-wide outbox_dispatched/outbox_dead metrics never wired (needs a domain/runtime port + markFailed signature change)


JiRaska:
## What

`openbank_outbox_dispatched_total`/`openbank_outbox_dead_total` have **zero call sites anywhere in the fleet**. Only the sibling gauge, `registerOutboxBacklog` (PENDING+FAILED count), is wired — via the shared `AbstractOutboxBacklogGauge` base class every service's outbox dispatcher extends. That's why `openbank_outbox_backlog` has 30 live series today while `_dispatched_total`/`_dead_total` have none.

Found during the 2026-08-16 observability sweep (#5049 round 2). Not fixed there — this is architecturally bigger than the per-service fixes in #5068/#5079, and deserves its own scoped PR(s).

## Why this can't be a same-module call

The actual dispatch loop is `OutboxDispatch.dispatchOnce()` in `openbank-libs-domain`, and that module has **zero framework imports by design** (ADR-0002/ADR-0122, enforced by `check-domain-purity.py`). `DomainMetrics` lives in `openbank-libs-runtime` (the framework-touching module) and depends on Micrometer — `dispatchOnce()` cannot call it directly without violating the domain/runtime split this repo has its own CI gate for.

The correct fix needs a **port**: a small domain-level observer interface (e.g. `OutboxDispatchObserver` with `onDispatched(service: String)` / `onDead(service: String)`) that `dispatchOnce()` invokes, implemented on the runtime side (`AbstractOutboxDispatcher`, which already has framework access) by calling `DomainMetrics.outboxDispatched`/`outboxDead`.

## Why "dead" specifically needs a second change

`OutboxRepository.markFailed(eventId, error)` returns `Unit`. Internally it applies `OutboxFailurePolicy.statusAfterFailure(attemptCount, maxAttempts)`:

```kotlin
fun statusAfterFailure(attemptCount: Int, maxAttempts: Int = DEFAULT_MAX_ATTEMPTS): OutboxStatus =
    if (attemptCount >= maxAttempts) OutboxStatus.DEAD else OutboxStatus.FAILED
```

[...]

PR #5096, opened 2026-08-16
feat(outbox): wire openbank_outbox_dispatched_total (issue #5091 phase 1)

JiRaska:
## What

Phase 1 of #5091: `openbank_outbox_dispatched_total` wired fleet-wide (mechanism) and live on two services (ledger, account). `openbank_outbox_dead_total` is deliberately **not** in this PR — see below.

## Why this couldn't be a same-module call

`OutboxDispatch.dispatchOnce` lives in `openbank-libs-domain`, which has **zero framework imports by design** (ADR-0002/ADR-0122, `check-domain-purity.py`). `DomainMetrics` (`openbank-libs-runtime`) depends on Micrometer — `dispatchOnce` cannot call it directly without breaking the domain/runtime split this repo has its own CI gate for.

Added the port instead: `OutboxDispatchObserver`, a `fun interface` single-method callback. `dispatchOnce` invokes it right after every confirmed `markSent`. `AbstractOutboxDispatcher` (runtime side, has framework access) implements it by delegating to `DomainMetrics` — the only place in this call graph allowed to know a metrics library exists.

```kotlin
fun interface OutboxDispatchObserver {
    fun onDispatched(entry: OutboxEntry)
    companion object { val NOOP: OutboxDispatchObserver = OutboxDispatchObserver { } }
}
```

`check-domain-purity.py`: **0 new violations**.

## Why `outboxDead` isn't in this PR

`OutboxRepository.markFailed(eventId, error)` returns `Unit`. It internally computes `OutboxFailurePolicy.statusAfterFailure(attemptCount)` and applies it, but the dispatch loop has no way to observe whether *this* call was the one that tipped a row into terminal `DEAD` versus just incremented `attempt_count`. Getting that right needs `markFailed`'s signature to change across **all ~31 `OutboxRepository` implementations** — a much larger, separately-reviewable change. Tracked as phase 2 in #5091, not attempted here.

## Why `metrics`/`service` are optional, not required

```kotlin
protected open val metrics: DomainMetrics? = null
protected open val service: String? = null
```

`open` with `null` defaults, not `abstract`. Every one of the ~30 existing concrete `AbstractOutboxDispatcher` subclasses across the fleet **compiles and behaves unchanged** without touching them. Making these `abstract` would force every dispatcher to be edited in the same PR that introduces the mechanism — turning a scoped, verified rollout into one enormous fleet-wide diff reviewed all at once, which is exactly the shape #5091 was written to avoid.

```kotlin
protected open suspend fun dispatchScheduledBatch() {
    val m = metrics; val svc = service
    val observer = if (m != null && svc != null) {
        OutboxDispatchObserver { entry -> m.outboxDispatched(svc, entry.eventType) }
    } else {
        OutboxDispatchObserver.NOOP
    }
    OutboxDispatch.dispatchOnce(outboxRepository, observer = observer) { entry -> publishWithResilience(entry) }
}
```

## Wired live on two services — proof, not just compilation

`ledger` (money-path) and `account`. Both needed `DomainMetrics` added to their `*OutboxDispatcher`'s constructor, which wasn't there before:

```kotlin
class LedgerOutboxDispatcher(
    private val repo: LedgerOutboxRepositoryImpl,
    private val publisher: KafkaLedgerOutboxEventPublisher,
    private val domainMetrics: DomainMetrics,
    @ConfigProperty(...) private val dispatchEnabled: Boolean,
) : AbstractOutboxDispatcher() {
    override val metrics: DomainMetrics get() = domainMetrics
    override val service: String get() = "ledger"
    ...
}
```

## Tests

`OutboxDispatchTest` (libs-domain) and `AbstractOutboxDispatcherTest` (libs-runtime) — existing suites unchanged and still green, plus new cases:
- observer notified once per real dispatch, with the full entry
- **not** notified on a publish failure
- **not** notified when a batch is abandoned on breaker-open (#4005's no-attempt-consumed rule)
- the default call shape (no `observer` argument) behaves exactly as every existing caller already uses it
- `AbstractOutboxDispatcherTest`: null `metrics`/`service` ⇒ zero `DomainMetrics` interaction, no crash; an opted-in dispatcher records `openbank_outbox_dispatched_total` tagged by `service` **and** `eventType`, asserted against a real `SimpleMeterRegistry` — not a mock verification.

## Verified

- `openbank-libs-domain`/`openbank-libs-runtime` suites: **10/10**, **6/6**.
- `ledger-service`: **303 tests, 0 failures**. `account-service`: **244 tests, 0 failures**.
- `ktlintCheck`, `detekt` — clean on all four touched modules.
- `quarkusBuild` — green on both services. Confirms `DomainMetrics` resolves through the new constructor parameter (the CDI/ArC class of failure unit tests can't see).
- `check-domain-purity.py`: 0 new violations.
- **Live baseline recorded before merge**: `count({__name__="openbank_outbox_dispatched_total"})` = **0** in Prometheus right now. Will check it lights up with real traffic after this deploys — the same live-verification discipline as every other PR in this sweep, not "the code compiles so it must work."
- Local `run-gates.py --all`: **150/151** (the one failure, `api-contract-gate`, needs `sudo` locally and reproduces on a clean `origin/main` worktree).

## Follow-up

Remaining ~28 services are a two-line mechanical addition each (inject `DomainMetrics`, override `metrics`/`service`) — tracked as a checklist in #5091, not attempted here. `outboxDead` (phase 2, the `markFailed` signature change) is also tracked there, separately.

## Review note

`ledger-service` is `money_path_services` — 2 approvals. No trust-boundary change (no new listener/port/auth key — this is metrics-only), so `threat-model-updated-on-trust-boundary-change` correctly doesn't fire.

Refs #5091

### DIFF
--- a/openbank-account-service/src/main/kotlin/com/openbank/account/infrastructure/outbox/AccountOutboxDispatcher.kt
+++ b/openbank-account-service/src/main/kotlin/com/openbank/account/infrastructure/outbox/AccountOutboxDispatcher.kt
@@ -4,6 +4,7 @@
 package com.openbank.account.infrastructure.outbox
 
 import com.openbank.account.application.port.out.AccountOutboxRepository
+import com.openbank.libs.observability.DomainMetrics
 import com.openbank.libs.persistence.outbox.AbstractOutboxDispatcher
 import com.openbank.libs.persistence.outbox.OutboxEntry
 import com.openbank.libs.persistence.outbox.OutboxEventPublisher
@@ -31,11 +32,16 @@ import org.eclipse.microprofile.faulttolerance.Timeout
 class AccountOutboxDispatcher(
     private val repo: AccountOutboxRepository,
     private val publisher: OutboxEventPublisher,
+    private val domainMetrics: DomainMetrics,
     @ConfigProperty(name = "openbank.outbox.dispatch-enabled", defaultValue = "false")
     private val dispatchEnabled: Boolean,
 ) : AbstractOutboxDispatcher() {
     override val outboxRepository: OutboxRepository get() = repo
     override val outboxEventPublisher: OutboxEventPublisher get() = publisher
+
+    // Issue #5091 phase 1: opts this dispatcher in to openbank_outbox_dispatched_total.
+    override val metrics: DomainMetrics get() = domainMetrics
+    override val service: String get() = "account"
 
     @Scheduled(
         every = "\${openbank.outbox.poll-interval:5s}",
--- a/openbank-ledger-service/src/main/kotlin/com/openbank/ledger/infrastructure/outbox/LedgerOutboxDispatcher.kt
+++ b/openbank-ledger-service/src/main/kotlin/com/openbank/ledger/infrastructure/outbox/LedgerOutboxDispatcher.kt
@@ -5,6 +5,7 @@ package com.openbank.ledger.infrastructure.outbox
 
 import com.openbank.ledger.infrastructure.messaging.KafkaLedgerOutboxEventPublisher
 import com.openbank.ledger.infrastructure.persistence.repository.LedgerOutboxRepositoryImpl
+import com.openbank.libs.observability.DomainMetrics
 import com.openbank.libs.persistence.outbox.AbstractOutboxDispatcher
 import com.openbank.libs.persistence.outbox.OutboxEntry
 import com.openbank.libs.persistence.outbox.OutboxEventPublisher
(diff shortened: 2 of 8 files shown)


JiRaska:
Superseded by #5071 — both change `OutboxDispatch.dispatchOnce`'s signature (this PR adds an `observer` param, #5071 changes the return type to `OutboxDispatchResult`), so they cannot both land. #5071 covers a broader scope: both `outboxDispatched` and `outboxDead` (this PR defers `outboxDead` to a phase 2), and wires metrics fleet-wide via a field-injected default (~34 dispatchers, 3 one-line overrides) rather than per-service constructor changes (2 of 34 wired here). Closing in favor of #5071.

Issue #3661, opened 2026-08-03
Apply the ECR repository declaration, then drop ecr:CreateRepository from the CI role


JiRaska:
#3659 declared the 61 `openbank-*` ECR repositories in OpenTofu (`openbank-infra/aws/envs/sandbox-platform/ecr-service-repositories.tf`), with the set derived from the gitops image pins plus the ARC runner image. The CI plan on that PR reported:

```
Plan: 61 to import, 0 to add, 61 to change, 0 to destroy.
```

**Nothing has been applied.** Two things are gated on that apply, and both are silently wrong until it happens.

## 1. The declaration is not yet the source of truth

The repositories are declared but not in state. Until `Platform OpenTofu`'s apply job runs:

- a new service still depends on #3492's fail-open `ensure-ecr-repository.sh`, or a hand-run `aws ecr create-repository`, for its first push;
- `prevent_destroy` protects nothing — it guards an instance in state, and there are none. Dropping a gitops image pin today removes a name from the derived set with nothing to object to.

So the current state is "reviewable declaration, unenforced", which reads from the repository exactly like "done".

## 2. `ecr:CreateRepository` should leave the CI role — but only after (1)

Point 3 of #3477. A build job that can create registry namespaces is the thing declaring the repositories was meant to remove the need for. It cannot be dropped first: #3492's backstop is what covers the window, and removing the grant while the declaration is unapplied would restore the original #3423 failure with no fallback.

Order: apply → confirm a plan with no import lines remaining → drop the grant from `arc-runners.tf` → confirm `ensure-ecr-repository.sh` still fails open (it is designed to, but that path is only exercised when it cannot look).

## Also worth knowing

The apply will make two changes beyond adoption, both intended and both in place:

- the root's `default_tags` reach 61 hand-created repositories — three currently carry `ManagedBy = "manual"` / `"manual-first-deploy"`;
- `scan_on_push: false -> true` on nine repositories created before the registry-level scanning rule (`anacredit`, `analytics-sink`, `card-issuance`, `customer-edge`, `lending`, `pid`, `sdd`, `standing-order`, `tpp-registry`).

Refs #3477, #3423, #3492

JiRaska:
Verified against the live account, read-only. **The apply still has not happened, and the premise
here is fully current** — with one consequence worth pulling out of the "also worth knowing" section,
because it is a live security gap rather than a tidy-up.

```
openbank-* ECR repositories : 61
scanOnPush = false          :  9
[...]

Issue #3477, opened 2026-08-02
Service ECR repositories are declared nowhere — 0 aws_ecr_repository resources for 56 released components


JiRaska:
#3423 was closed by #3444, which added `.github/scripts/ensure-ecr-repository.sh` to auto-create a missing repository at push time. That was reverted by #3453 because its existence check returned a false negative for repositories that **do** exist:

```
[openbank-sepa-payment] ==> ECR repository openbank-sepa-payment does not exist — creating it (#3423)
[openbank-sepa-payment] ERROR: could not create ECR repository openbank-sepa-payment in eu-north-1.
```

`aws ecr describe-repositories` confirms `openbank-sepa-payment`, `openbank-swift-service` and `openbank-transaction-service` all exist. So every fleet-wide build died at Docker push — three consecutive `auto-deploy` runs (`c7a78ff1`, `cb27ea0d`, `07281cbd`) — until the revert. The revert was correct; a check that cannot see an existing repository is worse than no check.

**The underlying gap is therefore open again, and it is wider than "a new service's first build fails".**

## Measured

```
resource "aws_ecr_repository" occurrences in openbank-infra/aws/  ->  0
released components in release-please-config.json                 -> 56
```

Every service image repository in this account exists **because someone created it by hand**. None is declared, none is under review, none has its scanning/immutability/lifecycle settings expressed anywhere, and nothing detects one being deleted. The first-build failure is just the symptom that happens to be visible.

## Why runtime auto-creation is the wrong layer

It puts `ecr:CreateRepository` on the CI role — a build job that can create registry namespaces — and it makes correctness depend on a probe running inside a `for` loop over ~50 services, where a false negative costs the whole fleet's deploy. #3444 demonstrated exactly that failure mode on its first real run.

It is also unfalsifiable in the direction that matters: the script's happy path only executes when a repository is genuinely missing, which is rare, so the branch that runs on *every* build is the one nobody tested.

## Proposed shape

Declare the repositories, don't create them:

1. `aws_ecr_repository` with `for_each` derived from the released-component set (`release-please-config.json`, the same list that decides what is a shipped module), so a new service gets its repository from the same PR that makes it a released component — before any build runs.
2. Import the ~50 existing repositories so the plan is empty rather than proposing to recreate them. This is the step that needs care: `terraform import` per repository, verified by an empty plan, before anything is applied.
[...]

Issue #3423, opened 2026-08-02
A new service's ECR repository is declared nowhere — first build always fails at Docker push


JiRaska:
## What

A new service can pass every CI gate and still fail its first deploy at `Docker push`, because its
**ECR repository does not exist and nothing in the repo declares it**.

Measured 2026-08-02 on the first `openbank-delegation-service` build (run 30762211614). The image
compiled; the push died:

```
failed to push 265175468565.dkr.ecr.eu-north-1.amazonaws.com/openbank-delegation-service:sandbox-8992de53:
unknown: The repository ... does not exist
```

`git grep aws_ecr_repository` over the whole repository returns **nothing**. All 32 existing repos
were created outside Terraform, by hand. `openbank-infra/aws/envs/sandbox-platform/ecr-image-scanning.tf`
documents the consequence without naming it — scanning is registry-level with an `openbank-*`
wildcard specifically so it covers repos that "do not exist yet, with no per-repo resource to forget".

## Why it is invisible until the push

Everything upstream of the push is consistent with a missing repository:

- `deploy-drift-declaration` only checks the pin's SHAPE (`sandbox-<hex>`).
- The fleet attestation gate reports the image as ABSENT, which is the CORRECT and expected state
  for a first registration — the documented fix is to declare it in
  `.github/scripts/fleet-attestation-placeholders.txt`, which I did. That declaration is about an
  unbuilt *tag*; it cannot distinguish "tag not built yet" from "repository does not exist".
- Kyverno blocks ArgoCD's dry-run diff for an absent tag, so the app sits at sync status Unknown —
  again indistinguishable.

So the first signal is a red build, after a full cold-cache compile.

## Two candidate fixes, and they are not equivalent

1. **Declare the repositories in Terraform** (`aws_ecr_repository` with `for_each` over the same
   service list). Makes the registry reproducible and reviewable. Cost: importing 32 existing repos
   into state, and the list becomes a second place that can drift from `ALL_SERVICES` — which is
   exactly the "hand-kept list of the thing it checks" shape this repo has been burned by. If done,
   derive the set from one source rather than typing it twice.
2. **A gate**: for every image pin in `openbank-infra/gitops/`, assert the ECR repository exists.
   Cheaper, catches it at PR time instead of at deploy time, and needs no state import. Does not
   make the registry declarative.

(1) and (2) compose; (2) alone already removes the failure mode that cost this build.

## Workaround used

```bash
[...]

PR #5118, opened 2026-08-16
chore(infra): drop ecr:CreateRepository from the CI deploy role

JiRaska:
## What, and why it's a draft

Item 2 of #3661 — dropping `ecr:CreateRepository` from the CI deploy role's `EcrCreateServiceRepository` statement, now `EcrDescribeServiceRepository` (Describe-only). **Opened as DRAFT deliberately: do not mark ready or merge until item 1 (`tofu apply` importing all 63 `openbank-*` repos) has landed and a fresh `tofu plan` shows 0 import lines remaining.** Merging this first restores #3423's original failure with no fallback — `ensure-ecr-repository.sh` needs `CreateRepository` to auto-create a missing repo, and that path is only safe to retire once nothing is actually missing.

## What was verified before writing any code

**Fresh registry comparison, not the stale Aug-3 plan.** Reproduced `ecr-service-repositories.tf`'s own derivation logic (regex over every gitops manifest for `<account>.dkr.ecr.<region>.amazonaws.com/(openbank-*)[:@]`, plus the CI runner image) and diffed against `aws ecr describe-repositories` on the live account:

```
derived set:  63 (was 61 at #3659's original plan — 2 services added since)
live account: 136 repositories total
  63 openbank-* match exactly, zero missing either direction
  73 are ECR pull-through-cache namespaces (docker-hub/*, ghcr/*, quay/*, ecr-public/*, k8s/*) —
     correctly out of this file's scope, untouched by it
```

So the apply this depends on is a **pure import**: 0 create, 0 destroy, 63 existing repositories brought under Terraform state. Nothing is created by it, and after it lands nothing legitimate is left for this grant to create either.

**Two `ecr:CreateRepository` grants exist on this role, only one of which is in scope.** The other (`EcrPullThroughReadAndImport`) is required for the pull-through-cache mechanism (`docker-hub/*` etc.) and is untouched — confirmed by reading both statements in full before editing, not grepping-and-deleting the first match.

**`ecr:DescribeRepositories` stays**, deliberately — it's what lets `ensure-ecr-repository.sh` (#3492) tell a real drift (repository deleted outside Terraform, or a new service's gitops manifest landing without its `ecr-service-repositories.tf` entry in the same PR) apart from a plain `AccessDenied`. That distinction is exactly what #3444/#3453 were about; collapsing it again from the other direction would reopen the same failure class.

## FinOps — checked explicitly, zero impact

- **State import**: Terraform state operations carry no AWS cost.
- **Tag reconciliation** (`default_tags` reaching hand-created repos): free.
- **`scan_on_push: false -> true` on 9 repos**: confirmed `scan_type = "BASIC"` in `ecr-image-scanning.tf` — AWS's native, no-charge scan, not Inspector/Enhanced. The account-level `aws_ecr_registry_scanning_configuration` already applies `SCAN_ON_PUSH` to all `openbank-*` by wildcard regardless of this PR; the per-repo setting is stated redundancy ("belt-and-braces" per that file's own comment), not a new cost driver.
- **This PR's own change** (IAM policy document): zero cost, reduces blast radius.

## What was deliberately NOT run

`tofu plan`/`validate` against the real backend — even read-only, that acquires an S3 state lock (`use_lockfile=true`), which is touching shared infrastructure. The verification above (live registry vs. the file's own derivation) establishes the same facts without needing it. `tofu fmt -check`: clean.

Refs #3661, #3477, #3423, #3492

### DIFF
--- a/openbank-infra/aws/envs/sandbox-platform/arc-runners.tf
+++ b/openbank-infra/aws/envs/sandbox-platform/arc-runners.tf
@@ -673,27 +673,29 @@ data "aws_iam_policy_document" "arc_deploy_ecr" {
       "arn:aws:ecr:${data.aws_region.current.name}:${data.aws_caller_identity.current.account_id}:repository/openbank-*"
     ]
   }
-  # A new service's ECR repository is declared nowhere — every one of the ~32 existing repos was
-  # created by hand — so the first build of a new service compiles, runs the whole cold-cache
-  # Gradle build, and dies at the push with `name unknown` (issue #3423, measured on
-  # openbank-delegation-service). Nothing upstream can see it: the drift declaration checks only
-  # the pin's SHAPE, and the attestation gate reports the image as ABSENT, which is the correct
-  # state for a first registration.
+  # CreateRepository REMOVED (#3661, point 2 of #3477): every openbank-* repository this role
+  # could have created is now declared in ecr-service-repositories.tf and, as of that PR's apply,
+  # imported into Terraform state — a build job that can create registry namespaces is exactly
+  # the thing declaring them was meant to remove the need for (#3423's original hole). Verified
+  # before removing, not assumed: `aws ecr describe-repositories` against the live account, diffed
+  # against the file's own for_each derivation (gitops image pins + the CI runner image), showed
+  # zero repositories the declaration does not already cover.
   #
-  # Scoped to openbank-* deliberately. CreateRepository already exists on this role for
-  # repository/docker-hub/* (the pull-through cache); widening it to `*` would let a build create
-  # anything in the registry, and the naming prefix is the whole boundary.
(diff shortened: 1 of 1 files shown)


JiRaska:
Precondition confirmed. Dispatched `platform-tofu.yml` — [run 31967347418](https://github.com/JiRaska/open-bank-oss/actions/runs/31967347418), `Apply complete! Resources: 0 added, 0 changed, 0 destroyed`. All 63 `aws_ecr_repository.service[...]` entries show `Refreshing state`, not `import` — they were already in Terraform state before this run (someone applied item 1 at some point after #3659's original plan; the issue's 'nothing has been applied' premise was stale). Zero drift confirms declared config matches live reality exactly. Marking ready.

Issue #5049, opened 2026-08-16
observability: 16 business metrics referenced by dashboards are never emitted by any service


JiRaska:
## What

16 business metrics referenced across 9 Grafana dashboards appear in **zero** Kotlin source files anywhere in the fleet. Never emitted, never wired to any service. Found during the 2026-08-16 observability empty-panel audit (`grep` for each metric name against every `openbank-*/src/main/kotlin` tree — 0 hits for all 16).

```
openbank_accounts_created_total       openbank_accounts_closed_total
openbank_parties_created_total        openbank_parties_verified_total
openbank_kyc_submissions_total        openbank_kyc_verdicts_total
openbank_sca_challenges_total         openbank_sca_completions_total
openbank_sanctions_screenings_total   openbank_sanctions_hits_total
openbank_payments_submitted_total     openbank_payments_completed_total
openbank_outbox_dispatched_total      openbank_outbox_dead_total
openbank_audit_chain_verify_duration_seconds_bucket
openbank_payment_processing_duration_seconds_bucket
```

## Blast radius

Dashboards with at least one panel depending on a metric in this list:

| Dashboard | Affected panels (of total) |
|---|---|
| Business KPIs | 11 |
| Executive Overview | 7 |
| Compliance & AML | 8 |
| Onboarding & Customer Service | 13 |
| Finance & Treasury | 3 (the other 2 empty panels there are the EoM Close retention-window false alarm, not this) |

These are the platform's headline business/exec dashboards — the ones a stakeholder or auditor is most likely to actually look at.

## Why this is the same failure class the repo already has rules against

This repo's own governance history (`CLAUDE.md`) documents this exact shape repeatedly: "a detector that exists in the prose and not in the ruleset reads as covered on every dashboard that counts detectors" (FinOps D2), "the header of this file has claimed a cross-AZ detector since the file was written... there was no D2 rule in it" (same file), the OPA-generator gate that covered 4 of 25 for months, `openapi.yaml` fiction (#... prod-readiness sweep). A Grafana panel with a title and a query is exactly as convincing as a detector name in a comment — more so, since it renders a chart. Nobody auditing "do we have business observability" would catch this by looking at the dashboard list; only running the query does.

## What this issue is NOT

[...]

Issue #3505, opened 2026-08-02
Audit hash chain has never verified a single link — checked=0 against 1066 chained rows


JiRaska:
Found on 2026-08-02, immediately after #3478 stopped `AuditChainBroken` from firing on the boot-time zero gauge. That fix is what made this visible: the critical is no longer dismissible as restart noise, and what remains is a real, permanent failure.

## The control has never verified a single link

```
openbank_audit_chain_intact                          = 0
openbank_audit_chain_entries_checked                 = 0     <-- never verified anything
openbank_audit_chain_entries_unchained               = 596
openbank_audit_chain_last_verified_timestamp_seconds = 1785704401   (the 21:00 cron DID run)
```

Meanwhile the table holds **1066 rows that carry a `record_hash`**:

```
 total | chained | unchained
  1662 |    1066 |       596
```

So 1066 rows are chained, and `checked` is **0**. `verifyChain()` returns on the first failure, and it fails on the very first chained row:

```
AUDIT CHAIN BROKEN after 0 verified links — first broken entry e187d590-629d-4399-8ea2-05860e3f5d70
```

That entry is `id = 951`, the first row written after the V5 chain migration.

## The linkage is fine — the recomputation is not

```
 id  | unchained |    prev12    |    rec12
 951 | f         | 000000000000 | 4ddd5994ddb9     <- genesis, correct
 952 | f         | 4ddd5994ddb9 | 8860e1232d47     <- links to 951, correct
 953 | f         | 8860e1232d47 | ad5c11fa9472     <- links to 952, correct
```

`chainLinkBroken` fails on `recordHash != recomputed`, not on `prevHash != expectedPrev`. Nothing about the chain's *shape* is wrong. `chainHash(GENESIS, entry951)` simply does not reproduce the stored `4ddd5994…`.

## Hypotheses tested and eliminated

| hypothesis | verdict |
|---|---|
| Pre-chain rows misreported as broken | **No** — `verifyChain` skips `recordHash == null` correctly; those 596 are counted as `unchained` |
| A Flyway backfill computed hashes differently | **No** — `V5__hash_chain.sql` only does two `ADD COLUMN`; it backfills nothing |
| JSONB key reordering changes `sha256(payload)` | **No** — `payload` is `TEXT`, not `jsonb` |
| The canonical hash input changed after those rows were written | **No** — `git log -S'val canonical = listOf'` shows it has never been edited |
[...]

PR #5148, opened 2026-08-16
fix(audit): publish a percentile histogram for the chain-verify timer

JiRaska:
## What

Refs #5049 — fixes the last real gap in the 16-metric sweep: `openbank_audit_chain_verify_duration_seconds_bucket`.

## Investigation

The other 15 metrics in #5049 turned out to already be wired (payments, sanctions, outbox — fixed in #5068/#5079/#5071/#5144; accounts/parties/kyc/sca confirmed already emitting via `DomainMetrics`, no code change needed, issue tracking was stale). This one metric is the exception: the underlying event **is** instrumented — `AuditChainIntegrityGauge.verify()` has timed the chain walk since #3505 — but the timer was never configured to publish a histogram, so the dashboard's `histogram_quantile()` panel had nothing to read.

## Root cause

```kotlin
registry.timer("openbank.audit.chain.verify.duration")
    .record(System.nanoTime() - startedAt, TimeUnit.NANOSECONDS)
```

`registry.timer(name)` is the implicit-registration form — it returns a `Timer` with no histogram configuration, so it publishes only `_count`/`_sum`/`_max`. No `_bucket` series exists to query, and the panel has been empty since the control shipped.

## Fix

Register the timer explicitly in `@PostConstruct`, held in a field the same way the four existing gauges already are (Micrometer only weakly references a gauge's source object; the existing code already documents why a strong-referenced field matters):

```kotlin
verifyDurationTimer = Timer.builder("openbank.audit.chain.verify.duration")
    .description("Duration of a full audit hash-chain verification walk")
    .publishPercentileHistogram()
    .register(registry)
```

`verify()` now records into that instance instead of re-resolving by name every run.

## Tests

Added a test using a real `PrometheusMeterRegistry` — the fleet's actual scrape backend — asserting the scraped payload contains `openbank_audit_chain_verify_duration_seconds_bucket`. `SimpleMeterRegistry` (used by the file's other tests) can't distinguish this class of bug; it isn't Prometheus-shaped, so a timer with no histogram config and one with `publishPercentileHistogram()` look identical through it.

`AuditChainIntegrityGaugeTest`: 6/6 passing (was 5, +1 new).

## Note

Small, audit-service-only, observability-only change — no behavior change to the chain-verify logic itself, same values recorded, just a different registration call.

### DIFF
--- a/openbank-audit-service/src/main/kotlin/com/openbank/audit/infrastructure/observability/AuditChainIntegrityGauge.kt
+++ b/openbank-audit-service/src/main/kotlin/com/openbank/audit/infrastructure/observability/AuditChainIntegrityGauge.kt
@@ -6,6 +6,7 @@ package com.openbank.audit.infrastructure.observability
 import com.openbank.audit.infrastructure.persistence.AuditRepository
 import io.micrometer.core.instrument.Gauge
 import io.micrometer.core.instrument.MeterRegistry
+import io.micrometer.core.instrument.Timer
 import io.quarkus.runtime.Startup
 import io.quarkus.scheduler.Scheduled
 import jakarta.annotation.PostConstruct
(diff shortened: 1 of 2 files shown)


JiRaska:
Superseded by #5062, already merged to main — same fix (publishPercentileHistogram() on the chain-verify timer), landed before I opened this. Found via rebase conflict, not review. Closing.

PR #5302, opened 2026-08-17
security(fleet): resolve 46 open code-scanning findings

JiRaska:
## Summary

Sweep of every open [code-scanning alert](https://github.com/JiRaska/open-bank-oss/security/code-scanning) (60 total) that a code change can actually fix, without breaking functionality. 46 fixed, 14 left open with a rationale.

**Pin 20 OpenSSF Scorecard `Pinned-Dependencies` findings**
- GitHub Actions pinned to a commit SHA (`checkout`, `setup-java`, `upload-artifact`, `download-artifact`, `cache`).
- pip/npm installs pinned by hash. New shared `.github/scripts/install-pyyaml-hash-pinned.sh` (npm has no per-package hash pin mechanism at all — `npm ci` against a committed lockfile is the only thing Scorecard accepts, so `.github/scripts/agent-review-claude-cli/` makes the Claude Code CLI install `npm ci`-able).
- Two `curl | python3` "downloadThenRun" false-triggers (JSON data piped into an interpreter, not a downloaded script) broken into capture-then-parse so the pipe itself doesn't exist syntactically.

**Fix 11 `java/log-injection` findings**
The fleet's shared `.sanitizeForLog()` extension idiom sanitizes correctly, but CodeQL's `LogInjectionSanitizer` only recognizes a *literal* `.replace('\n', ..).replace('\r', ..)` chain applied directly to the sink argument — a call routed through a named wrapper function is opaque to it, sanitizer or not. Inlined at each flagged log call. One of the eleven (`CardService.kt`'s `operator` id) was genuinely unsanitized; the rest were false negatives of the wrapper pattern, confirmed against upstream CodeQL's `LogInjection.qll`.

**Fix 8 `java/local-variable-is-never-read` findings**
All eight are Kotlin's compiler-synthesized `equals()` for a `data object` with zero properties (`tmp0_other_with_cast`), which CodeQL's Kotlin extractor misreads as a real unused local. Switched to plain `object` + an explicit `toString()` override — the only observable behavior `data object` was buying for a property-less singleton.

**Remove 2 genuinely unused values** (`js/unused-local-variable`): a dead `stateLabel` function in the campaign detail page, an unused `CardStatusChip` import on the cards page.

**Harden `collect-gate-health.mjs`** (`js/http-to-file-access`, 1 of 2): validate `artifact.id` is a safe integer before it reaches `path.join`. The second instance writes a fixed-path diagnostic JSON report built from API data — a content-flow false positive on a legitimate CI collector script, not a path or execution issue.

**Replace deprecated `io.quarkus.panache.common.Parameters`** (deprecated whole-class since Quarkus 3.34) with a plain `Map` in `IctIncidentRepositoryImpl`.

**Rename `InsuranceTermLifeInteroperabilityAdapter`'s overloaded `objectValue()` pair** (`java/confusing-method-signature`) so the two unrelated-signature overloads no longer share a name.

## Not touched — 14 `TokenPermissionsID` findings

Every one is Scorecard scoring an already job-scoped, already-documented, functionally required write permission at 0 for having any non-read scope at all: `auto-deploy.yml`/`admin-ui-deploy.yml` pushing deployment commits, `dependabot-auto-merge.yml` merging PRs, `branch-cleanup.yml` deleting branches, `dependency-submission.yml`'s API requirement, `pact-verification-reconcile.yml` re-running workflows, `backfill-release-evidence.yml` uploading release assets. None can be narrowed further without breaking the job. Recommend dismissing these 14 (and the one `http-to-file-access` false positive above) via the code-scanning UI with a "used in tests"/"won't fix" rationale rather than a source change — happy to do that too if wanted, flagging separately since dismissing alerts is a different kind of action than fixing code.

## Test plan
- [x] `./gradlew :<affected-service>:compileKotlin` — all 9 touched Kotlin modules build clean
- [x] `./gradlew :<affected-service>:ktlintCheck` — clean on every file this PR touches (one unrelated pre-existing `settings.gradle.kts` violation in `openbank-security-scanner`, not touched here)
- [x] `node --check` on `collect-gate-health.mjs`
- [x] Manual read-through of both admin-ui files confirming no other reference to the removed function/import
- [ ] CI (services-ci / admin-ui / workflow lint) — pending on this PR


JiRaska:
## Review — two admin-ui hunks are provably wrong against current `main`, and the headline new artifact is a guard that cannot fail

Reviewer comment only (no approve, no merge). Read the full 3-dot diff (32 files, 386+/117-, excluding the generated `package-lock.json`). Branch is **711 commits behind `main`**, so staleness was the first question.

### 1. Two "unused code" removals are wrong against live `main`

Both were correct against this branch's base and are not correct now:

- `openbank-admin-ui/src/app/campaigns/[id]/page.tsx` — the PR deletes `stateLabel` as unused. `main` rewrote this page in `5dfc00b6c` (*align campaign workflow rbac (#6108)*, 2026-08-21) and **still has `stateLabel` at `main:449`**, alongside `AuthGuard`/`Can` that this branch's base did not have.
- `openbank-admin-ui/src/app/cards/page.tsx` — the PR drops the `CardStatusChip` import as unused. `main` (`d8515053b`, 2026-08-20) **still imports and uses it at `main:22`**.

On current `main` these are a broken build or a behaviour removal, not a cleanup. Both hunks need to be dropped or re-derived after a rebase.

A third file is adjacent rather than wrong: `CaseOpenService.kt` — `main` (`c78a10d78`) added SHADOW delivery-mode gating immediately above this PR's edit region. If it merges "cleanly" the result deserves re-reading, because a 2-dot view of it reads as a revert of the shadow gate.

### 2. `install-pyyaml-hash-pinned.sh` — the pin is defeated by its own fallback

The new script ends in a chain equivalent to:

```
pip install --require-hashes ... 2>/dev/null || ... || python3 -c "import yaml" 2>/dev/null || error
```

The `python3 -c "import yaml"` arm accepts **any** pre-existing, unpinned PyYAML on the runner, and every install attempt's stderr is discarded. So a hash mismatch — the single event the pin exists to detect — is indistinguishable from "already installed", and the job proceeds green. On the self-hosted `openbank-build` pool, where PyYAML is almost certainly in the base image, the pinned arm may never be the one that executes at all.

This is the repo's own rule: a guard is not proven by what it prints, only by what it prevents. Before this merges, please run the negative case — feed it a deliberately wrong hash and confirm the job goes **red**. On the code as written I expect it to go green. It is wired into six workflows (`agent-review.yml`, `external-feed-watch.yml`, `incident-gate-coverage.yml`, `main-red-watch.yml`, `opa-policy.yml`, `upstream-unblock-watch.yml`), so it is six green signals about nothing.

The script's own comment also notes `ci.yml` keeps a duplicate hash set — two copies to drift.

### 3. `data object` -> `object`: 10 sites, no stated finding, half-applied

`Ap2Mandate.kt:57`, `Segment.kt:93,101`, `CardProductCatalogPort.kt:36`, `CaseOpenService.kt:28,33,38,43`, `SpendCeilings.kt:43`, `PolicyEvaluator.kt:114,118`, `CatalogKernel.kt:27`, `CatalogSchema.kt:28`.

No comment anywhere names the alert being cleared, and **11 `data object` declarations remain on the branch** — so whatever rule this addresses is only half-cleared, which means it will re-open. It is semantically equivalent (a `data object`'s generated `equals`/`hashCode` are identity anyway, and `toString` is restored by hand), so it is not dangerous; it is just undocumented churn in ap2/delegation/card-issuance domain types with no security value.

For a PR whose thesis is "resolve 46 findings", every class needs the finding id next to it. This one has none. Either name it or drop it.

### 4. Log-injection fixes: real, but they delete two documented incident notes and reverse a recorded decision

Six files inline `replace('\n','_').replace('\r','_')` in place of a shared `sanitizeForLog()` helper (`AgentPolicyGate.kt:52,79-102`, `McpEndpoint.kt:46,267-275`, `OpaPolicyDecisionPoint.kt:33-60`, `CaseOpenService.kt:121-147,184`, `ScaAdapters.kt:70-83`, `CardService.kt:334-341`). Behaviourally identical, so not a suppression, and the de-factoring is defensible as the way CodeQL's `LogInjectionSanitizer` pattern-matcher is satisfied. `CardService.kt:334` is a genuinely new sanitizer.

Three costs to weigh:

- `McpEndpoint.kt:46` deletes the comment recording the `@Path`-binds-to-the-next-declaration incident — the one where every `POST /mcp` answered 404 on a running pod while unit tests stayed green. `CaseOpenService.kt:184` deletes the same warning in companion-object form. Those notes are in those files because those files suffered it. Please carry them across.
- `CaseOpenService.kt:126` now sanitizes `workflowId`, explicitly reversing the recorded `#4215` analysis ("the asymmetry is the point"). Undoing a documented finding-dismissal to appease a scanner needs to be an argued decision, not a side effect.
- `McpEndpoint.kt:274` loses the `?: "-"` fallback on `operator`/`method`/`reason`. No runtime change (they are non-null or it would not compile), but the defensive default is gone.

### 5. Things that are straightforwardly right

- **Action SHA pinning** (`auto-deploy.yml:874`, `services-ci.yml:377,402`, `verification-metadata-drift.yml:58,65,106,132`) — and I verified these are **still needed**: `main` still carries floating tags at `services-ci.yml:379,404` and `auto-deploy.yml:912`. Not redundant.
- `java/confusing-method-signature` — `InsuranceTermLifeInteroperabilityAdapter.kt:251` rename plus 7 call sites. Correct.
- `java/deprecated-call` — `IctIncidentRepositoryImpl.kt:44-55`, `Parameters` -> `mutableMapOf`. The only genuine runtime change in the diff; the overload exists, but confirm it compiles against current `main`.
- `curl | python3` -> capture-then-feed (`resolve-record-deployment-version.sh:254-262`, `activate-cz-pack.sh:29-41`) — cosmetic by the author's own admission, honestly stated per instance, no security delta. Fine.
- `collect-gate-health.mjs:76-83` — argues the CodeQL finding is a false positive and adds an `Number.isInteger` guard anyway. Reasonable.

**No blanket suppressions anywhere.** No `# nosemgrep`, no `lgtm[...]`, no `codeql[...]` comment, no broadened `catch`, no `permissions:` widening. On the specific question asked of a security PR — fix vs silence — this PR is mostly fixes, and where it is cosmetic it says so. Credit where due.

Also worth stating plainly: `agent-review.yml`'s `npm ci` pin is real, but it runs under `set +e` with output discarded, so an install failure is invisible — and that workflow has never actually run a review since GitHub Models was retired, so the pin is untested in practice.

### 6. Threat models

The diff touches `src/main` of four services carrying threat models — card-issuance, sca, delegation, agent — and updates **none**. The changes are arguably non-substantive (log sanitization, `data object`), which is a fine argument, but it needs to be **in the PR body** for the enforced threat-model-diff gate rather than left implicit. Separately: `openbank-ap2-service` has no threat model file at all, and `Ap2Mandate.kt` is edited here.

### Verdict

**Needs change.** The security substance is sound; the problems are (a) two stale admin-ui hunks that are wrong against `main`, (b) a hash-pin guard I believe cannot fail — please falsify it, (c) an undocumented, half-applied `data object` sweep. Given 711 commits of drift I would rebase and re-derive rather than merge as-is, and I would consider splitting the `data object` sweep out entirely so the security changes are not held hostage to it.

JiRaska:
## Rebuilt on live `main`. Half of it was already resolved — 32 files down to 17.

New head: `5613f6a19`, signed (`%G?` = `G`), on current `main`. The base was **718 commits** behind, so the first question was not "does it still apply" but "is it still needed".

### Measured: 46 open findings then, 30 now

```
$ gh api 'repos/JiRaska/open-bank-oss/code-scanning/alerts?state=open&ref=refs/heads/main' --paginate
TOTAL_OPEN=30
```

Split by rule, and matched against what the branch touches:

| Rule | Open on `main` | Branch fixes | Not covered |
|---|---|---|---|
| `PinnedDependenciesID` | 19 | 18 | `schema-registry-apply.yml:59` |
| `js/unused-local-variable` | 5 | 2 | lending applications page, 2 test files |
| `TokenPermissionsID` | 2 | 0 | `backfill-release-evidence.yml`, `app-status-refresh.yml` |
| `js/http-to-file-access` | 1 | 1 | — |
| `BinaryArtifactsID` | 1 | 0 | `gradle-wrapper.jar` — not fixable |
| `java/*` (Kotlin) | **0** | 15 files | — |

### Dropped: all 15 Kotlin files

Every `java/*` alert the branch was written against is **closed as `fixed` on `main`** — `java/log-injection`, `java/local-variable-is-never-read`, `java/deprecated-call`, `java/confusing-method-signature`. There are zero open java-kotlin alerts today.

I checked what that meant rather than trusting the state, and it is worth recording. `AgentPolicyGate.kt` on `main` already sanitizes via a `sanitizeForLog()` helper; the branch's change was to *inline* `.replace('\n','_').replace('\r','_')` at all 12 call sites, on the theory that CodeQL's `LogInjectionSanitizer` only recognizes the literal chain applied directly to the sink and treats a wrapper as opaque. The alerts closed anyway. So the branch would have traded a named helper for twelve copies of a character-replacement chain to satisfy a finding that no longer exists — strictly worse code for no security gain.

Same for `SpendCeilings.kt`, where it rewrote `data object Allowed` into `object Allowed` with a hand-written `toString()`. Dropped all 15.

### Kept: 17 files, and I verified the pins rather than trusting the comments

A wrong SHA in an action pin breaks every workflow that uses it, and the `# v7.0.1` comment beside it is an unverified claim. Resolved each tag through the API (dereferencing annotated tags):

```
OK   actions/cache            v6.1.0 -> 55cc8345863c7cc4c66a329aec7e433d2d1c52a9
OK   actions/checkout         v7.0.1 -> 3d3c42e5aac5ba805825da76410c181273ba90b1
OK   actions/download-artifact v8.0.0 -> 70fc10c6e5e1ce46ad2ea6f2b72d43f7d47b13c3
OK   actions/setup-java       v5.7.0 -> b6effb05e454b25005698d916606bdc6ffcbf961
OK   actions/upload-artifact  v7.0.1 -> 043fb46d1a93c77aae656e7c1c64a875d1fc6a0a
```

All five match. The `curl | python3` removals in `resolve-record-deployment-version.sh` and `activate-cz-pack.sh` are Scorecard `downloadThenRun` findings — the data is JSON, not a script, but the check cannot distinguish the two, and a here-string is the cheaper fix than an exception.

Also confirmed the two dead-code removals are genuinely dead **on current `main`**, not merely on the old base — `stateLabel` has no remaining reference in `campaigns/[id]/page.tsx`, and `CardStatusChip` is not used in `cards/page.tsx` (only the `cardStatusColor` sibling export is).

### Verification actually run

```
$ actionlint            # per file, comparing finding SETS against origin/main
agent-review.yml main=0 branch=0        auto-deploy.yml main=1 branch=1
external-feed-watch.yml main=0 branch=0 incident-gate-coverage.yml main=0 branch=0
main-red-watch.yml main=0 branch=0      opa-policy.yml main=1 branch=1
services-ci.yml main=2 branch=2         upstream-unblock-watch.yml main=0 branch=0
verification-metadata-drift.yml main=0 branch=0
# no NEW findings. Three files report pre-existing findings, which is the probe's
# own negative control — it is capable of reporting.

$ npx tsc --noEmit      # after the generated artifacts exist
exit 0, 0 lines

$ npm test              # via npm test, so pretest bakes governance.json/catalog.json
Test Files  250 passed (250)
     Tests  1643 passed | 1 skipped (1644)
exit 0

$ bash -n  on all three shell scripts — OK
```

### Remaining, and the title

The title still says "46 open code-scanning findings"; it now resolves **21 of the 30 that are actually open**. Worth retitling before merge. The 9 left over are listed in the table above and are all *new* findings that postdate this branch — they want their own PR, not a wider diff here. `BinaryArtifactsID` on `gradle-wrapper.jar` is not resolvable at all and should be dismissed rather than chased.

PR #5642, opened 2026-08-18
fix(billing): escape billing_outbox JSON payloads via Jackson

### LINKED ISSUE #4701: billing_outbox has 2 DEAD rows at attempt_count 10: events written atomically, never published

JiRaska:
## What

Fixes the code defect behind the two `billing_outbox` rows that reached `DEAD` at
`attempt_count = 10` (#4701).

`AssessedFeeOutboxPayloads.postIntent` / `.reversalIntent` (used by
`BillingAssessmentRepositoryImpl` to write the `billing.fee.post-intent.v1` /
`billing.fee.reversal-intent.v1` outbox rows) hand-built the JSON payload by string
interpolation, embedding `fee.feeName` with **no escaping**. `feeName` is free text sourced
from the product catalog — the `fee_name` column has no length or charset constraint
(`BillingEntities.kt`). A fee name containing a double quote or backslash therefore produced
**malformed JSON**.

`LedgerOutboxEventPublisher.publish` deserializes that payload with Jackson before calling the
ledger; malformed JSON throws a parse exception on every single dispatch attempt — a
deterministic failure, not a transient one (`OutboxDispatch.isTransportUnavailable` correctly
excludes circuit-breaker/bulkhead faults from the attempt count, but a real per-row exception
like this counts every time). After `OutboxFailurePolicy`'s max-attempts threshold (10), the row
is parked `DEAD` — a terminal, non-alerting state (that half of the issue — making `DEAD` rows
alertable — is explicitly out of scope here and left to #4701's own follow-up).

Notably, this exact trap was already documented in the repo: `AnnualFeeSummaryOutboxPayloads`'s
own KDoc says it uses Jackson "unlike `BillingAssessmentRepositoryImpl`'s `AssessedFeeOutboxPayloads`
hand-built string concatenation ... fee names/categories carry free text (Czech diacritics,
possibly quotes), so hand-rolled string escaping is the wrong tool here" — that migration
(ADR-0248) covered the annual-summary payload but never circled back to fix the charge/reversal
leg it was calling out.

## Fix

- `AssessedFeeOutboxPayloads` moved out of `BillingAssessmentRepositoryImpl.kt` into its own file
  (mirrors `AnnualFeeSummaryOutboxPayloads`'s convention) and rewritten to serialize via
  `jacksonObjectMapper()` instead of string concatenation. Monetary fields stay JSON strings
  (`amount.toPlainString()`), matching the previous wire format and the fleet's "never a float"
  convention.
- Added `AssessedFeeOutboxPayloadsTest` — a standalone payload-shape unit test (no DB) proving a
  fee name containing a double quote, a backslash, or Czech diacritics now round-trips as valid
  JSON, plus contract-field coverage for both `postIntent` and `reversalIntent`. The previous
  hand-rolled `reversalIntent` also mangled the reversal `reason` (stripped newlines, swapped `"`
  for `'`) to dodge this same bug for one field only — the new version preserves it verbatim,
  correctly JSON-escaped.

## Scope / what this does NOT do

- Does **not** touch the 2 existing `DEAD` rows — replaying or deleting them is a live-data
  decision for a human, out of scope for this PR (per #4701 "What needs deciding" #3).
- Does **not** add `DEAD`-row alerting (#4701 "What needs deciding" #2) — that's a separate,
  broader change (a metric/alert already exists via `DomainMetrics.outboxDead`, but making it
  paged is a different PR).
- `openbank.outbox.dispatch-enabled` was already confirmed `true` in the issue; not implicated.
- No Kafka dotted-key or `runBlocking`-in-`@Scheduled` issue found — `BillingOutboxDispatcher.dispatch()`
  is already a `suspend fun`, and this outbox posts to ledger-service directly (not Kafka) for the
  charge/reversal legs.

## Test plan

- [x] `:openbank-billing-service:test` — new `AssessedFeeOutboxPayloadsTest` plus existing
      `LedgerOutboxEventPublisherTest` / `BillingOutboxDispatcherTest` all green.
- [x] `:openbank-billing-service:ktlintCheck :openbank-billing-service:detekt` — clean.
- [ ] The 2 dead rows in the sandbox need a human's live-DB remediation (replay/delete decision) —
      unaffected by this PR, tracked separately in #4701.

Closes #4701

### DIFF
--- a/openbank-billing-service/src/main/kotlin/com/openbank/billing/infrastructure/outbox/AssessedFeeOutboxPayloads.kt
+++ b/openbank-billing-service/src/main/kotlin/com/openbank/billing/infrastructure/outbox/AssessedFeeOutboxPayloads.kt
@@ -1,0 +1,97 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) OpenBank contributors. Licensed under the Apache License, Version 2.0.
+// See LICENSE in the repository root or https://www.apache.org/licenses/LICENSE-2.0 for details.
+
+package com.openbank.billing.infrastructure.outbox
+
+import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
+import com.openbank.billing.infrastructure.persistence.entity.AssessedFeeEntity
+
+/**
+ * Serializes an [AssessedFeeEntity] into the `billing.fee.post-intent.v1` /
+ * `billing.fee.reversal-intent.v1` `billing_outbox` payload shapes (ADR-0143 steps 2/2e).
+ *
+ * **Uses Jackson, not hand-built string concatenation (#4701).** `fee_name` (`feeName`, wired
+ * into `description` below) is free text sourced from the product catalog — no length or
+ * character constraint on the column (`BillingEntities.kt`'s `@Column(name = "fee_name",
+ * nullable = false)`) — and can contain a double quote or backslash. The previous version of
+ * this object hand-built the JSON string, embedding `feeName` with no escaping at all; a fee
+ * name containing either character produced malformed JSON that
+ * [LedgerOutboxEventPublisher.publish] could never parse. That failure is deterministic, not
+ * transient — `OutboxDispatch.dispatchOnce` re-claims and re-attempts the same row every tick,
+ * `mapper.readValue` throws the same `JsonParseException` every time, and after
+ * `OutboxFailurePolicy.DEFAULT_MAX_ATTEMPTS` (10) the row is parked `DEAD` with no alert (#4701:
+ * two `billing.fee.post-intent.v1` rows found dead at `attempt_count = 10`). This is the exact
+ * trap [AnnualFeeSummaryOutboxPayloads]'s own KDoc already named for the same reason ("fee
+ * names/categories carry free text ... possibly quotes, so hand-rolled string escaping is the
+ * wrong tool here") — that object was migrated to Jackson under ADR-0248, this one was not.
+ *
+ * `internal`, own file (mirrors [AnnualFeeSummaryOutboxPayloads]): this is the exact payload
+ * shape [LedgerOutboxEventPublisher] deserializes against, so it needs a payload-shape unit test
+ * with no database in the loop — impossible while nested `private` inside
+ * `BillingAssessmentRepositoryImpl`.
+ *
+ * Monetary fields stay JSON STRINGS (`amount.toPlainString()`), matching the previous wire
+ * format and every other money-carrying event in this fleet ("never a float").
+ */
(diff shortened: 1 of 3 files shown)


JiRaska:
Cross-reference: this is the actual fix for #4701's 2 dead rows (JSON deserialization failure on `mapper.readValue`, matching `BillingOutboxDispatcher`'s own KDoc that its `@CircuitBreaker` wraps deserialization + the ledger call + the write). #5600 initially claimed to root-cause the same 2 rows via a separate Kafka-connectivity gap — traced the dispatch path and that's a different, unrelated event type (`billing.annual-fee-summary.ready`, which never fired regardless). #5600 corrected accordingly; see #4701 for the full writeup. Suggest #4701 closes against this PR, not #5600.

PR #5656, opened 2026-08-19
fix(transaction): pin TransactionServiceTest clock to stop weekend flake

### LINKED ISSUE #5652: tech-debt(transaction): TransactionServiceTest asserts settlement dates against real wall-clock, flakes near month-end
`TransactionServiceTest.kt` has at least two tests that compute expected settlement/value dates via `Clock.systemUTC()` (real wall-clock) combined with business-day arithmetic (`today.plusMonths(1)`, `SettlementDateResolver`), instead of a fixed/injected clock:

- `TransactionServiceTest.kt:191` — "a SEPA settlement booked as TRANSFER still honours the requested value date"
- `TransactionServiceTest.kt:250` — "a domestic payment leaving the bank still honours the resolver"

Both failed on CI (PR #5636, run 32191007636, job 95912960000) with `AssertionFailedError`s that don't reference the PR's actual change (`AuthorizeInterceptor`) at all — confirmed unrelated by diff scope (the PR touches only `AuthorizeInterceptor.kt`). The likely mechanism: the test's expected value was computed once (or against a slightly different instant) than the code path being asserted, and business-day/month-end rollover logic makes the two diverge on certain calendar days — a classic clock-dependent flake (see this repo's own `Instant.EPOCH`/recency-assertion pitfalls in CLAUDE.md for the same class of defect: a test that only agrees with the code on some days).

**Fix should inject a fixed `Clock` into whatever computes "today"** (via CDI `@Inject Clock` or a fixed-clock test profile) rather than relying on `Clock.systemUTC()`/`LocalDate.now()` inside the code path under test, so the assertion is deterministic regardless of what day CI runs on.

Not investigated further — this issue exists to record the observation; a proper fix means finding every `Clock.systemUTC()`/`.now()` call in the settlement-date path and threading a testable clock through it, which needs someone to actually design that injection point.

JiRaska:
## Summary
- `TransactionServiceTest` computed `today` from `Clock.systemUTC()` and asserted a requested value date one month out passes through `SettlementDateResolver` **unchanged**. That only holds when `today.plusMonths(1)` lands on a CERTIS business day — on a weekend (or a CZK bank holiday) the resolver correctly rolls the value date forward, and the assertion flakes.
- Reproduced locally on 2026-08-19 (a Wednesday): `today.plusMonths(1)` = 2026-09-19 (Saturday) → resolver correctly returns 2026-09-21 (Monday), while the test expected the unrolled Saturday date.
- Fixed by pinning the test's `clock` field to `Clock.fixed(Instant.parse("2026-03-16T09:00:00Z"), SettlementDateResolver.BANK_ZONE)`, chosen so `today.plusMonths(1)` = 2026-04-16 (Thursday), clear of Easter and every other CERTIS holiday.
- Also fixed a related latent bug: the test's `clock` field was declared but never passed into the `TransactionService` under test (`setUp()` used the `@Inject`-style secondary constructor, which defaults to its own `Clock.systemUTC()`). Wired `clock` through to the primary constructor so the test's fixed clock actually governs production date derivation.
- No production code changed — `SettlementDateResolver`'s weekend/holiday rolling behavior is correct; the test's expectation was wrong.

Closes #5652

## Test plan
- [x] `./gradlew :openbank-transaction-service:test --tests "*TransactionServiceTest*"` — 25/25 pass (was 6 failing pre-fix on a real weekend-adjacent date)
- [x] `./gradlew :openbank-transaction-service:test` (full module) — 138 tests, 0 failures, 0 errors, 1 pre-existing skip
- [x] `./gradlew :openbank-transaction-service:ktlintCheck :openbank-transaction-service:detekt` — clean

### DIFF
--- a/openbank-transaction-service/src/test/kotlin/com/openbank/transaction/application/usecase/TransactionServiceTest.kt
+++ b/openbank-transaction-service/src/test/kotlin/com/openbank/transaction/application/usecase/TransactionServiceTest.kt
@@ -50,7 +50,13 @@ class TransactionServiceTest {
 
     private lateinit var service: TransactionService
 
-    private val clock: Clock = Clock.systemUTC()
+    // Fixed instead of Clock.systemUTC(): several tests derive `requested = today.plusMonths(1)`
+    // and assert it passes through SettlementDateResolver UNCHANGED. That only holds when the
+    // resulting date is a CERTIS business day — on a real clock it silently flakes whenever
+    // `today.plusMonths(1)` lands on a weekend (or a CZK bank holiday), because the resolver
+    // correctly rolls a non-business-day value date forward (#5652). 2026-03-16 -> +1 month is
+    // 2026-04-16, a Thursday, clear of Easter (2026-04-03/04-06) and every other CERTIS holiday.
+    private val clock: Clock = Clock.fixed(Instant.parse("2026-03-16T09:00:00Z"), SettlementDateResolver.BANK_ZONE)
     private var lastSaved: Transaction? = null
 
     @BeforeEach
@@ -71,6 +77,7 @@ class TransactionServiceTest {
             fxRatePort,
             temporalConfig,
             workflowClient,
+            clock,
         )
     }
 


JiRaska:
Superseded by #5689, which merged the same weekend-flake fix with a more thorough approach (fixes both the plusMonths(1) roll-forward case and the same-day weekend case). After merging main, this branch's diff against main is empty.

Issue #5679, opened 2026-08-19
3 of 16 four-eyes queues are still invisible to the approval inbox (balance, billing, consent)


JiRaska:
## What

`ApprovalStore.findPending` (openbank-libs-domain) documents itself as ADR-0227 D2's read side:

> PENDING approvals, oldest first (FIFO fairness for the checker queue) — the read side of the unified approval inbox (ADR-0227 D2)

Sixteen services carry an `ApprovalResource` on that shared store. Measured 2026-08-19, **2 of 16 expose a pending-list `@GET`**:

| exposes the list | invisible to the inbox |
|---|---|
| sanctions (#3472), lending | account, balance, billing, clearing, consent, domestic-payment, fx, ledger, notification, party, sepa-instant, sepa-payment, swift, transaction |

Every money-path domain is in the right column.

## Why it matters

The four-eyes flow parks the maker at HTTP 202 with an approval id and no way to hand it over except out of band. That is exactly the failure #3472 fixed for sanctions — and the Redis TTL (24h) then expires the request silently, so the ceremony fails closed *without anyone seeing it fail*.

On the supervisor side, `/approvals` renders `{ items, sources }`. A domain that is not federated contributes neither items nor a `sources` entry, so **an unread queue looks exactly like an empty one** — the most dangerous thing an approvals screen can say wrongly, as the BFF's own comment already notes for the 403 case.

## What is needed

Per service, following the sanctions/lending shape:

1. `@GET` on the existing `ApprovalResource` calling `approvalStore.findPending(limit)`, `@RolesAllowed(OPERATOR, ADMIN)`, `@Authorize(action = "<domain>.approval.read", resource = "")`.
2. `openapi.yaml` + `info.version` bump per ADR-0048, and a test.
3. A source function in `openbank-admin-ui/src/app/api/approvals/pending/route.ts` plus its case in `src/test/approvals-inbox.test.ts`.

One PR per service (the scope selects the released component). Money-path first: ledger, transaction, balance, domestic-payment, sepa-payment, sepa-instant, clearing, swift, fx.

## Related

ADR-0227 (moved `planned` → `partial` with the measurement), ADR-0155, #3472.

JiRaska:
## Re-measured on `origin/main`, 2026-08-26: **11 of 16**, not 2 of 16

The "2 of 16" in the issue body is a week stale. Current state, counted from the tree rather than
from memory:

```
[...]

Issue #3472, opened 2026-08-02
sanctions-service serves no pending-approvals queue, so a four-eyes checker cannot find the decision waiting for them


JiRaska:
## Summary

`openbank-sanctions-service` exposes no pending-approvals list. `ApprovalResource` serves only
`PATCH /api/v1/sanctions/approvals/{id}`, so a checker cannot discover a parked four-eyes decision —
the approval id has to be handed over out of band.

`openbank-lending-service`'s `ApprovalResource` already implements the ADR-0227 D2 shape
(`GET /api/v1/lending/approvals?limit=`), and `ApprovalStore.findPending` exists in libs-domain with
a KDoc that says each service is expected to serve one:

> PENDING approvals, oldest first (FIFO fairness for the checker queue) — the read side of the
> unified approval inbox (ADR-0227 D2): each service exposes its pending queue over REST and the
> admin-UI BFF federates them into one supervisor surface.

sanctions never got that half.

## Why it matters now

#3465 gave operators a maker flow for disposing of a screening hit, and a checker flow — but the
checker UI is an id field with instructions to paste what someone hands you, because there is
nothing to list. The four-eyes ceremony on `sanctions.clear` therefore works only if the two people
are already talking to each other. `rules.yaml` calls this the highest-risk action in the service
("a wrongly-cleared true positive is a real sanctions violation"), so "the checker cannot find the
request" is a real availability gap in a compliance control, not a UX nicety.

Consequence today: a maker's decision parked at 202 is invisible to everyone. Nothing lists it,
nothing alerts on it, and the approval TTL (24h in Redis) expires it silently — the same
shape as the pending-review queue that could never drain before #3378.

## Scope

1. `GET /api/v1/sanctions/approvals?limit=` on sanctions' `ApprovalResource`, mirroring lending's:
   `approvalStore.findPending(limit.coerceIn(1, MAX))`, role-gated, `@Authorize(action =
   "sanctions.approval.read")`.
2. `openapi.yaml`: new path + an API-contract version bump (ADR-0048 — an added endpoint is a minor).
3. Threat model: the review path is section 6 of `docs/threat-models/openbank-sanctions-service.md`;
   a new inbound REST surface on a money-path service now trips the enforced
   `threat-model-updated-on-trust-boundary-change` gate (#3431), so the model must be updated in the
   same diff. That is the gate working as intended.
4. admin-ui: BFF route + render the queue in the checker box added by #3465, replacing "paste an id
[...]

PR #5686, opened 2026-08-19
feat(balance): expose pending four-eyes approvals via approval inbox (#5679)

JiRaska:
## Summary

- `openbank-balance-service`'s `ApprovalResource` served only `PATCH /{id}`, so a parked
  `balance.credit`/`balance.debit` four-eyes decision was discoverable only by whoever was
  handed its id out of band — the same gap #3472 fixed for sanctions. This is the
  money-path (balance) slice of #5679 ("14 of 16 four-eyes queues are invisible to the
  approval inbox").
- Adds `GET /api/v1/balances/approvals?limit=`, `ROLE_OPERATOR`/`ROLE_ADMIN`,
  `@Authorize(action = "balance.approval.read", resource = "")`, limit clamped to 200 —
  mirrors sanctions (#3472) and lending's existing shape exactly. `ApprovalResponse` gains
  `makerId`/`createdAt` (additive).
- `openapi.yaml` `1.8.0 -> 1.9.0` (MINOR, ADR-0048: additive-only change). Also refactored
  the `PATCH /{id}` response to reference the same shared `ApprovalResponse` schema
  sanctions uses, instead of a second hand-duplicated inline copy.
- `openbank-admin-ui`: wires `balancePending` into the federated
  `/api/approvals/pending` BFF route alongside `lending`/`sanctions`/`agent`, plus test
  cases in `approvals-inbox.test.ts` (merge/sort, source read, degrade, 403-vs-down).
- No `version.txt` hand-edit — release-please derives the release version from commit
  type + path automatically (ADR-0029); confirmed this matches how #3472 shipped for
  sanctions (it didn't touch `version.txt` either).

## Verified

- `./gradlew :openbank-balance-service:compileKotlin :openbank-balance-service:detekt
  :openbank-balance-service:ktlintCheck :openbank-balance-service:test --tests
  "*Approval*" :openbank-balance-service:quarkusBuild` — all green.
- `cd openbank-admin-ui && npm test -- approvals-inbox` — 7/7 green (added 2 new cases:
  merge includes balance, source-read regression test for balance).
- `npx tsc --noEmit -p .` — clean.
- `python3 -c "import yaml; yaml.safe_load(open('openapi.yaml'))"` — valid YAML.

## Authorization check (money-path, verified by effect not by appearance)

Ran `opa eval` against the real `rest.rego` + `balance_rest_ext.rego` +
`rules-opa-data.yaml` (balance-service is deployed with `AUTHZ_ENFORCE=true`, ADR-0034
Phase 5 — this is not a shadow-mode service):

```
input: {principal: {type: HUMAN, roles: [ROLE_OPERATOR]}, action: balance.approval.read}
=> allow=true, reason=operator-read-any
```

`balance.approval.read` is covered by the generic `operator-read-any` rule (any
`*.read`/`*.list` action for `ROLE_OPERATOR`/`ROLE_ADMIN`) — no `rules.yaml` or `.rego`
change was needed for this endpoint to work in production.

**Known pre-existing gap, NOT introduced or fixed by this PR:** the same `opa eval`
against the *existing* `PATCH /{id}` action:

```
input: {principal: {type: HUMAN, roles: [ROLE_OPERATOR]}, action: balance.approval.decide, resource: {id: ...}}
=> allow=false
```

`balance.approval.decide` is granted to `ROLE_OPERATOR` neither by
`rules.yaml`'s `role_action_matrix` (unlike `sanctions.approval.decide`, which has an
explicit entry) nor by `balance_rest_ext.rego` (which has no `approval` rule at all). Under
balance-service's live `AUTHZ_ENFORCE=true`, this means the four-eyes **decide** endpoint
already 403s for every real operator today — the ceremony this PR makes discoverable
cannot currently be completed by a checker. This is a separate, pre-existing money-path
authorization defect and deliberately left out of this read-side PR rather than folded in
unreviewed; flagging for a dedicated follow-up.

## Test plan

- [x] Backend unit test (`ApprovalResourceMappingTest`) covers the additive
      `makerId`/`createdAt` fields on the mapping used by both `GET` and `PATCH`.
- [x] BFF route test covers: merge+sort across all four sources including balance, the
      balance-source-is-read regression, graceful degrade, 403-vs-unavailable
      distinction — all now including `balance`.
- [ ] Manual: hit `GET /api/v1/balances/approvals` as an operator against a running stack
      once #5679's decide-gap follow-up unblocks completing an approval end-to-end.

Refs #5679


JiRaska:
## Review — one blocking authz gap, and a clean-merge collision with #5694

Reviewer comment only (no approve, no merge). Read the full 3-dot diff (6 files) plus `#5694` for cross-comparison.

### 1. BLOCKING: `balance.approval.read` has no allow reason anywhere — the endpoint fails closed

`ApprovalResource.kt` adds `@Authorize(action = "balance.approval.read", resource = "")`. I checked every place that could grant it on `origin/main`:

- `openbank-infra/gitops/components/balances/balance_rest_ext.rego:59-74` — `operator-balance-write` is an **exact-match set**, `input.action in {balance.read, balance.hold, balance.holdRelease, balance.credit, balance.debit, balance.initialize, balance.reconciliation.read, balance.reconciliation.run}`. Not a prefix match. `balance.approval.read` is not in it.
- `viewer-balance-read` (`:91-95`) — `{balance.read, balance.reconciliation.read}`. Not in it.
- `rules.yaml: authz.role_action_matrix` — has `balance.approval.decide` (line 2621) but **no `balance.approval.read`**.

This PR touches neither the rego nor `rules.yaml`. So once `AUTHZ_ENFORCE` is true, the endpoint denies every caller including `ROLE_ADMIN`, and the admin-ui BFF reports the balance source as `forbidden` — an approval queue that reads as permanently unreachable.

Compare the precedent the PR says it mirrors. `openbank-clearing-service/.../ApprovalResource.kt:56-58` does not just assert this — it records the measurement:

> `clearingBatch_rest_ext.rego`'s `operator-clearing-write` reason is a **prefix match** on `clearingBatch.*` ... verified with a real `opa eval` against the regenerated bundle (issue #5679): `clearingBatch.approval.read` resolves `allow=true` for all three roles and `allow=false` for ROLE_VIEWER, no rules.yaml change needed.

That is the right standard, and clearing met it because its rule is a prefix match. Balance's is not. **This PR needs the same `opa eval`, run against the balance bundle, with a must-ALLOW and a must-DENY control** — and on the evidence above I expect it to come back `allow=false`, which means `balance.approval.read` has to be added to `operator-balance-write`'s action set (plus a bundle regeneration and the pod-roll annotation restamp).

Reading the rego cannot tell you which reason fires; please measure it rather than reasoning from the annotation.

### 2. BLOCKING (cross-PR): this and #5694 merge cleanly and leave `main` red

Both PRs edit `openbank-admin-ui/src/app/api/approvals/pending/route.ts`'s `NOT_CONFIGURED_SOURCES`, on **different lines**:

- this PR deletes `balance: 'not-configured'`
- **#5694** deletes `'sepa-instant': 'not-configured'` — and, in the same PR, retargets the guard test's canary to `expect(routeSource).toContain("balance: 'not-configured'")` (`approvals-source-truth.guard.test.ts:14`).

Git will merge both without a conflict, because neither touches the other's line. The result is a guard test asserting the presence of a line the other PR removed. `main` goes red on `approvals-source-truth.guard.test.ts` with no reviewer, no CI run and no diff having said anything — this is the "a clean merge can DELETE a list entry" shape, one step sideways.

Whoever merges second must fix the canary. Better: change the guard to assert a *property* rather than a specific service —

```ts
expect(Object.keys(NOT_CONFIGURED_SOURCES).length).toBeGreaterThan(0)
```

— because as written, every future slice of #5679 has to rewrite that assertion to point at whatever is still unwired, and the last slice makes it unsatisfiable.

### 3. The two PRs are the same pattern applied twice, and they are inconsistent

Same shape, correct in both: `MAX_PENDING_LIMIT = 200`, `limit.coerceIn(1, MAX)`, `suspend fun`, additive `makerId`/`createdAt`, "deliberately not filtered to someone else's requests" with the guard living in `RedisApprovalStore.decide`. I agree with that last decision — hiding a maker's own request would not stop them attempting it and would make the queue lie about its own depth.

But they diverge on the governance tail:

| | #5686 (balance) | #5694 (sepa-instant) |
|---|---|---|
| `rules.yaml` edited | no | yes (adds `sctInstPayment.approval.decide`) |
| `rules-opa-data.yaml` regenerated | n/a | yes |
| `shared_m2m_matrix_write_grants.declared` | n/a | yes |
| `opa eval` evidence for the new `.read` action | **none** | none |
| guard test updated | no | yes |

Neither has evidence for the `.read` action, which is the one this PR actually adds. Please align them.

### 4. Checked, found nothing

- Additive-only OpenAPI, `1.8.0 -> 1.9.0` minor bump, correct per ADR-0048. No competing claim on balance `1.9.0` that I could find — unlike sepa-instant, where #5992 contests `1.5.0`.
- Extracting the inline PATCH response schema into `components/schemas/ApprovalResponse` and `$ref`-ing it from both operations is right, and the two new properties are `nullable: true`, so no wire break for existing consumers.
- Threat model updated with a new I row and a dated §7 entry, and the DFD edge added. The trust boundary genuinely did move (a new read of the approval store from the operator plane), so that update is required and present.
- BFF: `AbortSignal.timeout(4000)`, `stateFor(res.status)`, `.catch(() => unavailable)` — matches the existing sources, and `not-configured` is not silently collapsed into `ok`.
- The `reads the balance queue at all — an unread source is indistinguishable from an empty one` test is the right test to write. An unread source really does render identically to an empty one.

### Verdict

**Needs change.** Fix (1) with a measured `opa eval`, and coordinate (2) with #5694 before either merges.

JiRaska:
## Rebuilt on current `main` — union with the `sepa-instant` queue from #5694

Both queues are wanted, so every conflict was resolved as a union, not a choice. The single-line
merges: the `domain:` union type, the `const [...]` destructuring, the `Promise.all` list, the
`sources` map, and the `it('merges ...')` title. Concat order in `route.ts` is now
`... sepaPayment, sepaInstant, balance, agent`.

### The index assertions — measured, not guessed

Neither side's indices were right once both queues are present. Ran `npm test` and let the real
ordering decide. Result: **13 items**, `L-1, D-1, C-1, J-1, W-1, SP-1, B-1, F-1, I-1, S-1, P-1, T-1, L-2`.

Note what the merged fixture exercises that neither side did alone: `balance` is concatenated
**after** `sepa-instant` yet sorts **before** it, because B-1 is 11:00 and I-1 is 11:30 — proposedAt
wins and only within-minute ties fall back to concat order. The comment above the assertion now
names both tie groups (D-1/C-1/J-1/W-1/SP-1/**B-1** at 11:00; F-1/I-1 at 11:30) and says this
explicitly, and the new `balance` mock branch carries its own note about why it is at 11:00.

### One extra file this needed

`approvals-source-truth.guard.test.ts` asserted `routeSource` contains `balance: 'not-configured'` —
#5694 had moved the exemplar there. Wiring `balance` removes that line, so the guard failed on
`main`'s own assertion. Moved the exemplar on to `account`, which is still unwired, with a note that
it must stay a queue the repo knows exists and does not read, or the guard stops discriminating.

### Verification

```
npm test                       -> exit 0   Test Files 264 passed (264)
                                           Tests 1659 passed | 1 skipped (1660)
npx tsc --noEmit               -> exit 0
npx eslint <the 3 changed files> -> exit 0
./gradlew :openbank-balance-service:{ktlintCheck,detekt,test} -> BUILD SUCCESSFUL
```

Run via `npm test` (not bare `npx vitest run`) so the `pretest` hook bakes `governance.json` /
`catalog.json`.

Also re-checked the API-version race against live `main`: `openapi.yaml` `info.version` is `1.8.0` on
`main` and `1.9.0` here, so the number is still free.

JiRaska:
Duplicate of #7031; #7031 was rebased to a standalone balance-only implementation with source-truth UI contracts and balance module gates rerun. Closing this older duplicate prevents two endpoint/BFF implementations.

Issue #5628, opened 2026-08-18
security-scanner-service down 4h+: Flyway V4 migration landed after V5 already deployed, blocks every boot


JiRaska:
## What

`security-scanner-service` has been **down** (`CrashLoopBackOff`, 0/1 ready) for at least 4 hours in the sandbox, since ~16:47Z. Found while sweeping for live incidents (following on from #4701's billing-service Kafka gap).

## Root cause

```
Caused by: org.flywaydb.core.api.exception.FlywayValidateException: Validate failed: Migrations have failed validation
Detected resolved migration not applied to database: 4.
```

Deployed DB's `flyway_schema_history`:

```
installed_rank | version | description             | success | installed_on
1               | 2       | create security outbox  | t       | 2026-08-02 22:24:54
2               | 3       | hibernate sequences      | t       | 2026-08-02 22:24:54
3               | 5       | create ict incidents     | t       | 2026-08-16 20:52:54
```

`V5` applied 2026-08-16 20:52. `V4__drop_security_outbox.sql` (#4940) merged AFTER `V5__create_ict_incidents.sql` (#4939) — a lower-numbered migration landed in the repo after a higher one had already been deployed and applied. Flyway's default `outOfOrder=false` refuses to apply `V4` once `V5` is already the highest applied version, and `validateOnMigrate` (the default) fails the whole boot rather than silently skipping it — so the service has been unable to start since whatever deploy first shipped a jar containing both migrations.

Confirmed via `git log`: PR #4939 (V5) merged, then PR #4940 (V4) merged — the classic "coordinate Flyway version numbers against live deployed state, not just the repo" gap this repo has hit before, just in the opposite direction (a migration added retroactively with a LOWER number than one already live, rather than two PRs claiming the same number).

## Fix

Renumbering `V4__drop_security_outbox.sql` to the next free version is safe: it has never been applied anywhere (absent from `flyway_schema_history` on the only environment that runs this service). See PR (to be linked).

JiRaska:
Resolved and verified live — closing.

`security-scanner-service`: **1/1 Running, 0 restarts, 11h uptime**. The CrashLoopBackOff is gone.

The decisive evidence is the migration history, which now shows the out-of-order apply actually happening rather than just being configured:

```
 version |      description       | success
---------+------------------------+---------
 2       | create security outbox | t
 3       | hibernate sequences    | t
 5       | create ict incidents   | t
[...]

PR #5688, opened 2026-08-19
feat(ci): two enforced gates so #4701 and #5628's incident classes can't recur

JiRaska:
## What

Two new enforced gates so the two live incidents fixed earlier this session (#4701, #5628) become structurally impossible to reintroduce, not just patched once.

## 1. `check-kafka-deployment-wiring.py` — issue #4701

billing-service's outbox producer had read `KAFKA_BOOTSTRAP_SERVERS` from its own `application.yaml` since ADR-0248 shipped it — its Deployment never supplied it (no env var, no `KafkaUser`, no cert `Secret`). SmallRye's connector fell through to `localhost:9092`; the pod stayed `Ready`, only the readiness probe's own log line said the reactive-messaging channel was DOWN, and the outbox circuit breaker quietly parked events `DEAD`.

`check-kafka-acl-coverage.py` (#2598) already exists but assumes the broker connection itself is a given — it checks topic ACLs, not that the wiring to *reach* the broker exists at all. This gate closes that gap: every service with a `smallrye-kafka` channel must have `KAFKA_BOOTSTRAP_SERVERS` on its matching gitops container.

**Verified against the live fleet**: 42 Kafka-using services, 0 findings. `billing-service` is already fixed (#5600, merged). `analytics-sink`'s deliberate ConfigMap-based alternate wiring via `QUARKUS_CONFIG_LOCATIONS` (issue #686 — SmallRye's env-var reverse-mapping is ambiguous for its hyphenated channel name) is a recognised exemption, confirmed by reading its actual gitops manifest, not assumed.

## 2. `check-flyway-version-commit-order.py` — issue #5628

`V4__drop_security_outbox.sql`'s own PR (#4940) merged *after* `V5__create_ict_incidents.sql`'s PR (#4939) had already deployed and applied V5 live. The resulting version set `{2,3,4,5}` has no gaps and looks completely fine as a static invariant — the actual defect is **order**, which no existing gate expresses, and neither PR's own CI could see it (each ran green against its own base — this repo's documented "PR green is about the base, not main" trap).

This gate closes that from git history: for each service, sort its migrations by the commit that first added each file to `origin/main`, and require version numbers to be strictly increasing in that order.

**Detect-fast, not prevention** — it runs on push to main too (`ci.yml`'s existing `push: [main]` trigger), so a race between two PRs still can't be seen pre-merge by either one alone, but the violation now surfaces within minutes on main instead of via a crashed pod hours later.

**Verified against full repo history**: 350 migrations across 53 services, 2 known pre-existing instances baselined (`security-scanner` V4 / #5628 itself, `campaign-service` V13 — both confirmed live via `flyway_schema_history` to already be fixed with `QUARKUS_FLYWAY_OUT_OF_ORDER=true`, which is also the gate's own suggested fix for an *already-merged* collision — renumbering is blocked post-merge by `check-db-migration.py`'s own `db-migration-gate`, which treats a rename as editing an already-committed migration's identity).

### Two bugs fixed while writing this gate

- `git log --follow`'s similarity-based rename detection produced a false positive for `document-service` (matched V2 against an unrelated infra commit by content similarity, misordering it before V1 — despite `flyway_schema_history` confirming correct 1-then-2 application). Removed `--follow`; uses a plain add-commit lookup instead.
- The naive per-file `git log` loop cost ~55s over 350 subprocess spawns. Replaced with one `git log --name-status` pass over the whole migration pathspec — same result in ~2s.

## Verified

- Both selftests pass standalone and through `run-gates.py`.
- `gatelib.subjects()` wired on both (`check-gate-subject-floor.py` clean).
- `ruff --select E9,F,B` clean (the exact CI invocation).
- `check-license-headers.py` clean.
- Neither false-positives against the live fleet's *actual* state — `document-service` (V1 applied at rank 1) and `campaign-service` (out-of-order confirmed live: rank 13=V15, rank 14=V13) were checked against real `flyway_schema_history`, not assumed from git alone, before being baselined.

Refs #4701, #5628


JiRaska:
CI red was **my gate's own bug**, and the failure was correct. Pushed a fix.

The flyway gate passed locally and failed on the PR run, reporting both baseline entries as stale. That verdict was right — the gate had silently checked **nothing**.

`first_commit_order` built its position map from `git rev-list origin/main` and, on `CalledProcessError`, returned an **empty** map. On a PR run that ref frequently does not exist: `actions/checkout` fetches with a narrow refspec, so `refs/remotes/origin/main` is absent even at `fetch-depth: 0`, and the script's own `git fetch origin main` updates `FETCH_HEAD` and nothing else. Empty order → no file has a position → no pairs compared → no violations → **green**, while still printing `350 migration(s) across 53 service(s) checked`, which reads exactly like a real result.

Nothing else would have caught it. The only reason it surfaced is the `KNOWN_VIOLATIONS` both-ways check: with zero violations found, every baselined entry looked fixed and got reported as stale. That is the ratchet catching the *checker* rather than the corpus — the single reason that idiom earns its cost.

Three changes, each closing a different half:
- Resolve the mainline through a candidate chain (`origin/main`, `FETCH_HEAD`, `main`, `HEAD`) instead of one ref CI does not have.
- If migrations exist but nothing could be ordered → explicit error, exit 1. A gate that cannot see its corpus must say so, not agree with it.
- `gatelib.subjects()` now counts migrations actually **ordered**, not files found on disk. The `min_subjects` floor could not catch this before, because the on-disk count was full and healthy-looking; it now reads 0 in the broken case and trips independently of the baseline check.

Verified by effect: with every `rev-list` call forced to raise, the gate exits 1, prints the error and reports `SUBJECTS=0` — where before it exited 0 and reported 350. Normal run unchanged (selftest passes, 349/350 ordered, clean, 2 baselined, ruff clean).

Rebased onto your `rationale`/`review_after` commit rather than force-pushing over it.

JiRaska:
**Root cause found — third time, and only because the previous commit made the gate report its own provenance.**

CI printed:

```
mainline resolved as: FETCH_HEAD@15dd7273d (3 commits)
```

The gates shard checks out at `fetch-depth: 1` — `ci.yml` says so in its own comment. On a shallow repo `git log --diff-filter=A` reports **every** file as ADDED at the shallow boundary, because that grafted root is where history appears to begin. So all 355 migrations resolved to **one** position, the per-service sort fell through to its version tiebreak, every service looked monotonic, and the gate reported zero violations — while printing a healthy-looking `355 migration(s) ordered`.

That is why it kept disagreeing with a local run. Not a different answer to the same question: a confident answer to a different one.

Against full history both violations reproduce — campaign V15 at rev-list position 4094 vs V13 at 4167, security-scanner V5 at 4382 vs V4 at 4476. So the baseline entries are correct and were never stale.

### Fix

- Detect shallowness (`git rev-parse --is-shallow-repository`) and **deepen in place** (`--unshallow`, then `--depth=2147483647`). Done in the gate rather than by putting `fetch-depth: 0` on the shard, because the other 60 gates do not need full history and should not pay for it.
- Evaluate **every** candidate ref and keep the **longest**, never the first that answers.
- **Depth guard**: a history with fewer commits than there are migrations cannot have introduced them one at a time, so it is truncated whatever it claims to be → empty order → the gate errors instead of reporting.

### Falsified both directions on a fresh `--depth 1` clone

Worth noting the first attempt at this test **passed vacuously** — an earlier run had already `--unshallow`ed the fixture, so it was no longer shallow. Redone on a fresh clone:

| case | result |
|---|---|
| shallow, deepening blocked | `TOO SHALLOW for 349 migrations`, `SUBJECTS=0`, **exit 1** |
| shallow, deepening allowed | 4759 commits, 348/349 ordered, correct verdict, exit 0 |

### The pattern across all three attempts

Every version *ran* and printed a plausible subject count; none looked broken. First it measured nothing (empty map), then the wrong history (implicit HEAD), then a truncated one. All three were visible only because `KNOWN_VIOLATIONS` is checked in **both** directions — with no violations found, every baselined entry reads as stale. Without that idiom this gate ships green and blind, which is the failure it exists to prevent.

PR #5694, opened 2026-08-19
feat(sepa-instant): expose pending four-eyes approvals via approval inbox

JiRaska:
## Summary

Per issue #5679 ("14 of 16 four-eyes queues are invisible to the approval inbox"), this is the
`openbank-sepa-instant` slice — mirroring the shape landed for domestic-payment (#5692) and
sepa-payment (#5691).

**Commit 1 — the feature:**
- `ApprovalResource` served only `PATCH /{id}` (decide), so an `sctInstPayment.recall` four-eyes
  decision parked at 202 was discoverable only by whoever had been handed its approval id out of
  band. Added `GET /api/v1/sepa-instant/approvals` (`approvalStore.findPending(limit)`), matching
  the existing decide endpoint's `@RolesAllowed(ROLE_OPERATOR, ROLE_ADMIN, ROLE_PAYMENTS)` and
  `@Authorize(action = "sctInstPayment.approval.read")`. Limit clamped to 200.
- `openapi.yaml`: new path + shared `ApprovalResponse` schema (`$ref`'d from `PATCH` too),
  `info.version` 1.4.0 -> 1.5.0 (ADR-0048, additive-only — `ApprovalResponse` gains `makerId`/
  `createdAt`).
- `admin-ui`: new `sepaInstantPending()` source in `api/approvals/pending/route.ts`, wired into
  the federated inbox merge/degrade logic; `approvals-inbox.test.ts` extended.
- New `ApprovalResourceMappingTest`, mirroring sanctions/ledger/domestic-payment's own test.
- Threat model updated (§4a new I-row + DFD update + change log).

**Commit 2 — a genuine, separate authz gap found while verifying the new endpoint (mandatory
authz-gap check per the balance-service precedent, #5686/#5690):**

Verifying by effect (not by appearance) that the checker roles could actually reach the *new*
`sctInstPayment.approval.read` action also meant checking the *existing* `sctInstPayment.approval.decide`
action the same way. `opa eval` against the real `rest.rego` + `rules-opa-data.yaml` bundle
resolved **`allow=false` for a real ROLE_OPERATOR** on `sctInstPayment.approval.decide` — the
action was present in `rules.yaml`'s `role_action_matrix` for the sibling
`sepaPayment.approval.decide` but absent for sepa-instant's own action. In any
`AUTHZ_ENFORCE=true` environment, the existing decide endpoint has been 403ing every operator
trying to decide a paused SCT Inst recall since it shipped — same shape as the balance-service gap
(#5686/#5690).

Fixed with the exact same mirrored shape balance used: added the matrix grant (mirroring
`sepaPayment.approval.decide`'s entry) to `role_action_matrix.ROLE_OPERATOR` and
`shared_m2m_matrix_write_grants.declared`, regenerated `rules-opa-data.yaml`, then regenerated
every service's OPA bundle (a `role_action_matrix` edit restamps the whole fleet — that's why this
PR's second commit touches ~75 files under `gitops/components` even though only sepa-instant's
own rule set changed).

**Known residual, not fixed here:** `matrix-allows` is role-only, and
`service-account-openbank-edge` carries `ROLE_OPERATOR` in at least one realm template — the same
exposure balance-service closed with a per-service `prohibited` rule in `balance_rest_ext.rego`.
sepa-instant has no standalone ext-rego file to extend (its REST extension is a heredoc inside
`gen-sepa-instant-opa-bundle.sh` with no `prohibited` block at all today), and there is no
verified in-repo M2M caller of `ApprovalResource`. Building new rego infrastructure for that was
out of scope for this mirrored fix — flagged here and in the threat model's change log rather than
attempted.

## Verification

`opa eval` against the regenerated bundle:

| Principal | Action | Result | Reason |
|---|---|---|---|
| `HUMAN` + `ROLE_OPERATOR` | `sctInstPayment.approval.decide` | `allow=true` | `matrix-allows` |
| `HUMAN` + `ROLE_KYC_OPENER` (deny control) | `sctInstPayment.approval.decide` | `allow=false` | — |
| `HUMAN` + `ROLE_OPERATOR` | `sctInstPayment.approval.read` (new GET) | `allow=true` | `operator-read-any` (generic `.read` rule, no matrix entry needed) |

Reproduced the pre-fix state against the unmodified `rules-opa-data.yaml` (same OPERATOR input
resolved `allow=false`), confirming this closes a real, currently-live gap.

- `check-matrix-write-grants.py`: 0 undeclared, 0 stale
- `gen-rules-opa-data.py --check`: in sync
- `check-opa-bundle-parses.py`: 43 bundles, no unloadable policy
- `check-opa-sidecar-bundle-shape.py`: 43 OPA sidecars, clean
- `opa test openbank-libs/governance/policies openbank-infra/opa/policies`: **170/170 pass**
- `./gradlew :openbank-sepa-instant:compileKotlin :openbank-sepa-instant:test --tests "*Approval*" :openbank-sepa-instant:ktlintCheck :openbank-sepa-instant:detekt`: **BUILD SUCCESSFUL**
- `cd openbank-admin-ui && npm test -- approvals-inbox`: **7/7 pass**

## Test plan

- [x] `opa eval` allow/deny controls for both the new GET action and the existing decide action
- [x] `opa eval` against the pre-fix data reproduces the reported gap
- [x] `opa test` on the full governance policy suite (170/170)
- [x] `check-matrix-write-grants.py`, `gen-rules-opa-data.py --check`, `check-opa-bundle-parses.py`, `check-opa-sidecar-bundle-shape.py`
- [x] `compileKotlin` / `test --tests "*Approval*"` / `ktlintCheck` / `detekt` for `openbank-sepa-instant`
- [x] `npm test -- approvals-inbox` for `openbank-admin-ui`

Refs #5679

### DIFF
--- a/docs/threat-models/openbank-sepa-instant.md
+++ b/docs/threat-models/openbank-sepa-instant.md
@@ -64,11 +64,13 @@ here (see ADR-0155, issue #413).
 | **T**ampering | A stale, mismatched, or already-consumed `X-Approval-Id` is replayed to unlock a different recall | `AuthorizeInterceptor` requires the approval's `action` + `resourceId` + `makerId` to match the CURRENT request exactly, `status == APPROVED`, and marks it `EXECUTED` (one-time use) on success; any mismatch re-issues a fresh pending approval instead of proceeding |
 | **R**epudiation | No record of who approved a recall clawing back settled funds | `PendingApproval.decidedBy` + `decidedAt` recorded in the approval record itself (Redis, TTL-bounded — see ADR-0155 Negative consequences: not yet a permanent audit trail) |
 | **I**nfo disclosure | Approval id enumeration reveals payment/action metadata to an unauthorized caller | `find`/`decide` require the caller to already hold a valid, role-gated session; the id itself is a random UUID (`RedisApprovalStore`, not sequential) |
+| **I**nfo disclosure | (issue #5679) `GET /api/v1/sepa-instant/approvals` lists every pending four-eyes request with its `makerId` and age | Role-gated `ROLE_OPERATOR`/`ROLE_ADMIN`/`ROLE_PAYMENTS` + `@Authorize(action = "sctInstPayment.approval.read")`; the payload carries approval metadata only — the action name, the resource id and who asked — never payment/account details. Limit clamped to 200 — an unbounded query parameter over a Redis scan is a trivially reachable amplification. Deliberately NOT filtered to exclude the caller's own requests: hiding a maker's request from them would not stop them attempting it (the guard is in `RedisApprovalStore.decide`, server-side) and would only make the queue lie about its own depth |
 | **D**oS | Flooding `POST /{paymentId}/recall` to exhaust Redis with pending approvals | Bounded by the same rate-limit/idempotency controls as the gated endpoint itself; each `PendingApproval` is TTL-bounded (86400s) so abandoned records expire |
 
-**DFD update:** adds `Operator (checker) → PATCH /api/v1/sepa-instant/approvals/{id} → Redis
-(approval:*)` alongside the existing `POST /{paymentId}/recall` edge; the maker's retry reuses
-the existing DFD edge.
+**DFD update:** adds `Operator (checker) → GET /api/v1/sepa-instant/approvals → Redis (approval:*)`
+and `Operator (checker) → PATCH /api/v1/sepa-instant/approvals/{id} → Redis (approval:*)`
+alongside the existing `POST /{paymentId}/recall` edge; the maker's retry reuses the existing DFD
+edge.
 **Risk class:** integrity (segregation of duties on a fund-clawback action) + confidentiality
 (approval record scope).
 **Rollback:** `authz.four-eyes.enforce=false` (default) — the endpoint and store exist but do
@@ -83,6 +85,39 @@ not change any existing request's outcome until explicitly flipped.
   need an additional store; not implemented in this PR.
 
 ## 6. Change log
+
+- **2026-08-19** — `ApprovalResource` served only `PATCH /{id}` (decide), so a
+  `sctInstPayment.recall` four-eyes decision parked at 202 was discoverable only by whoever had
+  been handed its approval id out of band — the ceremony completed only if the two operators were
+  already talking, and the 24h Redis TTL then expired the request silently otherwise (issue #5679,
+  mirroring sanctions #3472, ledger and domestic-payment). Added
+  `GET /api/v1/sepa-instant/approvals` (§4a new I row); additive-only OpenAPI change (1.4.0 ->
+  1.5.0, ADR-0048).
+  - **Checked the existing decide endpoint's own authz posture while here** (verify-by-effect, not
+    by appearance): `opa eval` against the real `rest.rego` + `rules-opa-data.yaml` bundle showed
+    `sctInstPayment.approval.decide` resolving **`allow=false` for a real ROLE_OPERATOR** — the
+    action was missing from `rules.yaml`'s `role_action_matrix` (present for the sibling
+    `sepaPayment.approval.decide`, absent for this service's own `sctInstPayment.approval.decide`
+    since the four-eyes gate for this rail was wired). The decide endpoint has therefore been
+    403ing every operator in any `AUTHZ_ENFORCE=true` environment since it shipped — same shape as
+    the balance-service gap found by #5686/#5690. **Fixed** by adding the matrix grant (mirroring
+    `sepaPayment.approval.decide`'s entry) to both `role_action_matrix.ROLE_OPERATOR` and
+    `shared_m2m_matrix_write_grants.declared`, then regenerating `rules-opa-data.yaml` and every
+    service's OPA bundle (a `role_action_matrix` edit restamps the fleet). Verified with `opa eval`:
+    `sctInstPayment.approval.decide` now resolves `allow=true`/`reason=matrix-allows` for
+    ROLE_OPERATOR, and a non-operator role (`ROLE_KYC_OPENER`) still resolves `allow=false`.
+  - **Known residual, not fixed here**: `matrix-allows` is role-only, and the deployed realm
+    template gives `service-account-openbank-edge` `ROLE_OPERATOR` in at least one environment (see
+    root `CLAUDE.md`'s realm-drift note) — the same exposure balance-service closed with a
+    per-service `prohibited` rule in `balance_rest_ext.rego`. sepa-instant has no such standalone
+    ext-rego file to extend (its REST extension is a heredoc inside
+    `gen-sepa-instant-opa-bundle.sh` with no `prohibited` block at all today), and there is no
+    verified in-repo M2M caller of `ApprovalResource` (the gen script's own comment already notes
+    this for `sctInstPayment.create`). Building a new prohibition mechanism was out of scope for
+    this PR's mirrored fix; tracked as follow-up under issue #5679's own money-path-first ordering.
+  - **Rollback:** revert both commits independently — the matrix grant only changes an OPA
+    `allow` decision (advisory in this environment, `authz.four-eyes.enforce=false`), and the new
+    `GET` is additive.
 
 - **2026-08-09** — Fraud shadow scoring's fallback is now observable (#4221). **No new trust
   boundary and no new caller**: the outbound edge to fraud-service (OIDC client-credentials + mTLS,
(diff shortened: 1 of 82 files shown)


JiRaska:
## Heads-up: `info.version` collision with #5992 (not a review of this PR's content)

This PR bumps `openbank-sepa-instant/src/main/resources/openapi.yaml` `info.version` `1.4.0 -> 1.5.0`.
So does **#5992** (`fix(sepa-instant): publish the real SctInstStatus vocabulary and reconcile the
payment schemas`), which is open and edits the same file.

Both change the same single line from the same value to the same value, so **git merges them
without a conflict** — nothing is flagged, and whichever lands second ends up with `1.5.0`
describing a document that also contains the other PR's changes. One API-contract version is
silently skipped, and the ADR-0048 classification for the second merge is computed against a
version that was already consumed.

Reading `origin/main` before bumping (which both PRs did) cannot catch this: `main` still holds
`1.4.0` for both. The check that catches it is against the *open* set.

Whoever merges second should re-read `git show origin/main:openbank-sepa-instant/src/main/resources/openapi.yaml | grep -m1 '^  version:'`
after the first lands and take `1.6.0`. No action needed if this one merges first.

I'm reviewing the #5962 enum-drift family and flagged the same thing on #5992; posting here so the
half that merges first also knows.

JiRaska:

Live `origin/main` is **sepa-instant `1.4.0`**. Two open PRs bump it to `1.5.0`:

- this PR
- #5694

(#6019 also carries a sepa-instant hunk, but it is **byte-identical** to #5992's — duplicate work
rather than a third claimant.)

Whichever merges **second** must re-read the version off live `origin/main` and take `1.6.0`:

```bash
git fetch origin main
git show origin/main:openbank-sepa-instant/src/main/resources/openapi.yaml | grep -m1 '^  version:'
```

No release tag is involved — tags sit on the 0.x release axis (`sepa-instant-v0.8.1`), which
ADR-0048 keeps separate from the 1.x API-contract axis.

Nothing to change right now, and this is **not** what is currently blocking this PR: the red
#6220 fixed on main at 14:57Z, after these runs started. It clears on the next run.

JiRaska:
## Review — version race still stands, CI diagnosis, an unrelated authz grant, and a clean-merge collision with #5686

Reviewer comment only (no approve, no merge). Read the full 3-dot diff (9 files) plus `#5686` and `#5992` for cross-comparison.

### 1. Version race on `info.version` — CONFIRMED, still live

Re-measured against live `origin/main`:

- `origin/main` sepa-instant spec: `1.4.0`
- this PR: `1.5.0`
- **#5992** (`fix/sepa-instant-openapi-enum-drift-5962`, OPEN): also `1.5.0`

The earlier annotation still holds. Whoever merges second takes `1.6.0`, and must re-read the version off **live `main`**, not off this branch's base — a matching version line merges "cleanly" because git compares text, not claimed versions.

### 2. Both failing checks have one cause: the OPA bundles were not regenerated

`build + verify bundle, check ConfigMap sync` fails with:

> A service OPA bundle ConfigMap or its pod-roll checksum annotation is stale — re-run the gen-*opa-bundle*.sh generators shown above and commit the result.

and its diff shows exactly this PR's `sctInstPayment.approval.decide` line landing in a bundle, plus checksum churn (e.g. `tpp-registry-service.yaml` `c7fba1f3f7cb74d7 -> b82e0c86286e3949`). This PR edits a `rules.yaml` key that OPA **reads** (`authz.role_action_matrix`), and touches zero files under `openbank-infra/gitops/components/**/opa-bundle*` — so the bundles embed a stale `rules-opa-data.yaml`.

Fix, in this order (order matters — a generator that runs before the derived file is rewritten hashes a stale input and the diff check passes about nothing):

```
python3 .github/scripts/gen-agents-opa-data.py     # needs `opa` on PATH
```

Commit the lot. Such a PR has a short shelf life — a competing governance PR will conflict it — so merge promptly once green.

### 3. The `rules.yaml` change grants a WRITE action, and it is not what this PR is about

The diff adds `sctInstPayment.approval.decide` to `authz.role_action_matrix` and to `shared_m2m_matrix_write_grants.declared`. Declaring it is correct — `check-matrix-write-grants.py` requires exactly that, and it is properly done in both files.

But two things deserve an explicit answer in the PR body:

**(a) It is out of scope.** This PR adds a `GET` (`sctInstPayment.approval.**read**`). The grant added is `.decide` — a *write*, fixing a different pre-existing gap. That is a separate decision riding along in a feature PR, and `rules.yaml` write grants are the one place where riding along is expensive.

**(b) A matrix line is a grant to a machine, and no policy can veto it.** `rest.rego`'s `matrix-allows` permits any `HUMAN` holding the role, and the Keycloak service accounts are classified `HUMAN` and hold `ROLE_OPERATOR` in at least one realm. `shared_m2m_write_prohibition` cannot help — it is keyed by reason name, and vetoing `matrix-allows` would veto every legitimate matrix call. So this line means a service account can decide a four-eyes approval on an instant payment recall. If that is intended, say so; if not, an exact-`principal.id` rule in a `*_rest_ext.rego` is the shape that expresses it.

### 4. The action this PR actually adds has no measured allow decision

`ApprovalResource.kt:52` — `@Authorize(action = "sctInstPayment.approval.read", resource = "")`. That action is in neither the `role_action_matrix` (the PR adds `.decide`, not `.read`) nor any `*_rest_ext.rego` I could find; `git grep -l sctInstPayment origin/main -- '*.rego'` returns only `vop_rest_ext.rego`, base `rest.rego` and `agents.rego`.

The precedent this PR mirrors sets the standard. `openbank-clearing-service/.../ApprovalResource.kt:56-58` records an actual measurement — *"verified with a real `opa eval` against the regenerated bundle (issue #5679): `clearingBatch.approval.read` resolves `allow=true` for all three roles and `allow=false` for ROLE_VIEWER"*. Please run the same against the sepa-instant bundle, with a must-ALLOW and a must-DENY control, and put the result in the KDoc. Reading the rego cannot tell you which of several reasons fires. #5686 has the identical hole for `balance.approval.read`, and there it looks worse (balance's rule is an exact-match set, not a prefix match).

### 5. BLOCKING (cross-PR): this and #5686 merge cleanly and leave `main` red

Both PRs edit `NOT_CONFIGURED_SOURCES` in `openbank-admin-ui/src/app/api/approvals/pending/route.ts`, on **different lines**:

- this PR deletes `'sepa-instant': 'not-configured'` and retargets the guard canary to `expect(routeSource).toContain("balance: 'not-configured'")` (`approvals-source-truth.guard.test.ts:14`)
- **#5686** deletes `balance: 'not-configured'`

Git merges both with no conflict. The result is this PR's guard asserting a line #5686 removed, and `approvals-source-truth.guard.test.ts` goes red on `main` — with no diff, no reviewer and no CI run having flagged it.

Rather than re-point the canary at whichever service is still unwired (which every future #5679 slice then has to redo, until the last one makes it unsatisfiable), assert the property:

```ts
expect(Object.keys(NOT_CONFIGURED_SOURCES).length).toBeGreaterThan(0)
```

### 6. Consistency with #5686 and with #5719

#5686 and this PR are the same pattern applied twice, and the Kotlin halves are genuinely consistent — same `MAX_PENDING_LIMIT = 200`, same `coerceIn(1, MAX)`, same `suspend fun`, same additive `makerId`/`createdAt`, same "deliberately not filtered to someone else's requests" reasoning (which I agree with: the guard belongs in `RedisApprovalStore.decide`, and filtering at read time would make the queue lie about its own depth without preventing anything).

They diverge only on the governance tail — #5686 edits no `rules.yaml` and no guard test; this one edits both. Neither has `opa eval` evidence for its new `.read` action. Please align them and merge them in a known order.

#5719 (`fix/event-handlers-stop-acking-failed-work`) is a different concern — handler ack semantics, not approval exposure — and I found no overlap in files or behaviour.

### 7. Checked, found nothing

- Additive OpenAPI only; `ApprovalResponse` extracted to `components/schemas` with both new fields `nullable: true`, so no wire break.
- `@RolesAllowed("ROLE_OPERATOR", "ROLE_ADMIN", "ROLE_PAYMENTS")` on the GET matches the existing PATCH on the same resource — no role widening.
- Threat model `docs/threat-models/openbank-sepa-instant.md` updated.
- `MAX_PENDING_LIMIT` is a `const val` in a `private companion object` — avoids the `MagicNumber` detekt trap.

### Verdict

**Needs change.** Regenerate the bundles (fixes both red checks), re-take `info.version` after #5992, coordinate the guard-test line with #5686, and either justify or split out the `.decide` matrix grant.

Issue #5698, opened 2026-08-19
Remediate the 9 parties still stranded without a KYC case (the #5698 consumer fix and orphan detection have both landed)


JiRaska:
## What happened

A few seconds of `kyc-db` downtime on 2026-08-19 left a customer with no KYC case, two accounts stuck in `PENDING_ACTIVATION`, and no welcome bonus.

`PARTY_CREATED` for party `fad8c8db-…` arrived at 12:56:15 while `kyc-db-rw` was refusing connections:

```
ERROR [party-events-in] Failed to auto-open/screen KYC case for party fad8c8db-…
Caused by: Connection refused: kyc-db-rw.kyc.svc:5432
```

`PartyEventConsumer.handleCreated` caught it, logged, and returned normally — which **acks the message**. The event is gone. No case ⇒ party stays `PENDING_KYC` ⇒ no `PARTY status=ACTIVE` event ⇒ `account-service` never activates the accounts ⇒ `grantWelcomeBonus`, which runs only inside the activation branch, never fires.

The pooled sequence shows nothing ever retried: `kyc_cases_seq.last_value` never advanced past the block already in use, so no INSERT was attempted after the failure.

## Blast radius

**10 of 73 parties in sandbox have no KYC case at all** — 13.7% of the onboarding funnel, stuck forever, with no alert on any of it:

| party | since |
|---|---|
| fad8c8db-… | 2026-08-19 (remediated) |
| 58fb3ae8-…, d7c887ee-… | 2026-07-17 |
| 91b1baf2-… | 2026-07-16 |
| 6 more | 2026-06-07 |

## Why it was invisible

Two green tests asserted the swallow as intended behaviour, one of them justifying it with "the message is acked and the party stream can replay". Nothing replays it. Reading the tests would have confirmed the bug as a feature.

## Fix

- [x] Consumer rethrows after bounded retry so the connector dead-letters (PR linked below). Malformed events still ack — that is the real poison-pill case and the only one.
- [ ] Remediate the 9 remaining orphaned parties (same replay path used for `fad8c8db`).
- [ ] Detection: nothing today notices a party that never got a case. Needs an alert or a reconciler, otherwise the next occurrence is found the same way this one was — a human noticing an account did not work.
- [ ] Fleet sweep: same catch-and-ack shape in other consumers.

JiRaska:
## Correction: one stranded party, not ten

[...]

PR #5699, opened 2026-08-19
fix(kyc): rethrow a failed party event instead of acking the work away

JiRaska:
Closes the first box of #5698.

## The incident

`PARTY_CREATED` for a real onboarding customer arrived at 12:56:15 on 2026-08-19 while `kyc-db-rw` was refusing connections. `PartyEventConsumer.handleCreated` caught the exception, logged `Failed to auto-open/screen KYC case`, and returned normally — **which acks the message**.

No KYC case ⇒ party stays `PENDING_KYC` ⇒ no `PARTY status=ACTIVE` ⇒ `account-service` never activates ⇒ `grantWelcomeBonus`, which lives inside the activation branch, never fires. The customer had two `PENDING_ACTIVATION` accounts and no money.

Nothing retried, and that is measured rather than assumed: `kyc_cases_seq.last_value` never advanced past the block already in use, so no INSERT was attempted after the failure (the repo's pooled-sequence diagnostic).

**10 of 73 parties in sandbox are in this state** — 13.7% of the onboarding funnel, silently, the oldest since 2026-06-07.

## The conflation

The class KDoc called the catch-all "poison-pill protection". Two opposite failures were sharing one handler:

- a **malformed event** is unretryable — replaying it fails identically forever, so log-and-ack is right. That is the only poison pill, and it is unchanged.
- a **transient dependency failure** is the opposite: the event is fine, the infrastructure is not, and the work must still happen. Acking loses it with no trace but an ERROR line nobody alerts on.

Domain and infrastructure failures now go through `withBoundedRetry` (3 attempts, linear backoff) and are **rethrown** if they still fail, so the connector retries and ultimately dead-letters. Same shape as `CardDelegationEventConsumer` (card-issuance) and `DelegationEventConsumer` (account-service).

`PARTY_ERASED` gets the same treatment, where swallowing is worse than a stalled funnel: an acked-but-failed erasure leaves PII in place while the log records the erasure as done. `account-service`'s equivalent handler already propagates deliberately and says why.

## Two green tests asserted the defect

```
a domain failure is swallowed so the consumer group is not wedged
PARTY_ERASED anonymisation failure is swallowed to protect the consumer group
```

The first justified itself with *"the message is acked and the party stream can replay"*. **Nothing replays it.** Both are replaced. Worth stating plainly: the test suite documented the bug as intended behaviour, so no amount of test-reading would have found this — only the live DB did.

## Falsification

| | tests |
|---|---|
| fix reverted, new tests in place | **3 failed / 11** |
| fix applied | **14 / 14 green** |

The new tests assert the retry **count**, not just that something throws — so a rethrow with no retries, and a retry that never gives up, are both red. `ktlintCheck` + `detekt` clean.

## Live remediation (already done, before this PR)

The customer was unstuck through the real event path, not a hand-written DB row: delete the manually-opened `OPEN` case, re-drive the `party_outbox` row for `PARTY_CREATED`, let the sandbox straight-through path run. Result: case `APPROVED` (`reviewedBy=sandbox-auto-approval`), party `ACTIVE`, both accounts `ACTIVE`, **100 000 CZK booked** on the CURRENT account at 13:26:19.

## Not in this PR (tracked in #5698)

- the 9 older orphaned parties
- detection — nothing notices a party that never got a case; today's answer is "a human sees an account that does not work"
- fleet sweep for the same catch-and-ack shape in other consumers

### DIFF
--- a/openbank-kyc-service/src/main/kotlin/com/openbank/kyc/infrastructure/kafka/PartyEventConsumer.kt
+++ b/openbank-kyc-service/src/main/kotlin/com/openbank/kyc/infrastructure/kafka/PartyEventConsumer.kt
@@ -11,6 +11,7 @@ import com.openbank.kyc.application.PepScreeningService
 import com.openbank.kyc.application.port.out.KycCaseRepository
 import jakarta.enterprise.context.ApplicationScoped
 import jakarta.inject.Inject
+import kotlinx.coroutines.delay
 import org.eclipse.microprofile.reactive.messaging.Incoming
 import org.jboss.logging.Logger
 import java.time.Clock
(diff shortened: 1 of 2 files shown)


JiRaska:
Superseded by #5719.

That PR uses the same fix but through a shared `EventRetry` helper in `openbank-libs-runtime`, and converts the other 14 sites the audit found across 9 more services — plus an enforced gate (`event-handler-swallows`) so a new consumer or scheduler cannot reintroduce the shape. Keeping this one would leave two implementations of the same retry, which is the duplication #5719 removes.

Closing in favour of it; the kyc-service change itself is preserved there commit-for-commit.

PR #5715, opened 2026-08-19
fix(interest): stop acking event-consumer failures that a retry could fix (#5698)

JiRaska:
## What

Two more instances of the #5698 catch-and-ack shape, in the two **money-path** services. Same fix shape as PR #5699 (kyc-service): bounded retry then rethrow for downstream calls, keep the ack for genuinely unretryable payloads.

A sibling PR covers the non-money-path consumers; this PR deliberately touches only `openbank-interest-service` and `openbank-fraud-service`.

## The split, which is the whole fix

| failure | before | after |
|---|---|---|
| malformed / undecodable payload | logged, acked | **unchanged** — logged, acked |
| upstream data defect (negative tax total) | logged, acked | **unchanged** — logged, acked |
| downstream call (REST rail, DB write) | logged, **acked** | 3 attempts, linear backoff, then **rethrown** so the connector dead-letters |
| ADR-0140 feature store (shadow plane) | logged, acked | **unchanged**, and the KDoc now says why |

A redelivery of a malformed payload decodes identically and fails identically forever, so acking it is correct; blanket-rethrowing would wedge the partition on the first bad event.

## 1. `openbank-interest-service` — `WithholdingRemittanceSettlementConsumer` (highest severity)

The outer `try` wrapped the entire `consume()`, including `bookAndSettle()` -> `transactionClient.initiateTransaction(...)` — the REST call that books the cash leg of a due withholding-tax remittance to the finanční úřad. A transient transaction-service failure was logged and the message acked, stranding a **regulatory tax remittance PENDING forever** with the event gone and no endpoint to re-drive it.

Two comments asserted behaviour the code did not have, which is how it survived review:

- the handler comment: the boundary `"must swallow ANY failure (parse, rail call) and ack"`;
- `bookAndSettle`'s non-2xx branch: `"the batch stays PENDING for retry (a redelivery of this event, or an operator re-driving it)"` — a redelivery that could never arrive, because that branch also returned normally and acked.

Both corrected; the class KDoc now states the split. A non-2xx-non-409 response throws `RemittanceBookingFailedException` rather than logging, and `remitUseCase.settle()` is covered too — booking the money and then silently failing to advance the batch leaves it PENDING against a rail that already moved cash.

## 2. `openbank-fraud-service` — `TransactionSignalConsumer`

The class KDoc claimed `"Failures are DLQ'd"`. Nothing rethrew, so nothing was: three individually-swallowed calls inside `runBlocking`, handler returns normally, message acked. Consumer lag and every dashboard built on it reported the event as processed. An undercounted velocity window is a fraud rule that quietly does not fire — `"scoring degrades gracefully"` is true of a *missing* aggregate, not a licence to create one silently.

`recordTransaction()` and `recordPayment()` now retry-then-rethrow. **`FeatureOnlineUpdater` is deliberately still swallowed** — it is the ADR-0140 shadow plane feeding no live decision, so dead-lettering a real velocity signal for a shadow-plane write would trade a working control for an experimental one. The KDoc now says that specifically instead of a blanket claim that was false for all three.

Trade-off recorded in the KDoc because it is real: `payee_history` upserts guard on `last_transaction_id` and are redelivery-safe, but `velocity_aggregates` has no dedup, so a redelivery re-adds the amount. Over-counting makes fraud rules fire *more* readily; under-counting makes them silently not fire — which is the bug being fixed.

## Tests — and the negative case

Four green tests across the two services asserted the swallow **as intended behaviour** (`"is caught and does not propagate"`, `"does not settle the batch and does not throw"`). As the issue notes, reading them would have confirmed the bug as a feature. All four are replaced.

Added per service: a persistent-failure test asserting the rethrow **and** the attempt count, a transient-failure test proving one blip is retried and the work still completes, and explicit ACK tests for the malformed/unretryable paths — so the suite can distinguish this fix from a blanket rethrow.

**Verified by effect, not appearance.** Each new suite was run against the pre-fix consumer:

- fraud — exactly the 3 new behaviour tests FAIL; the malformed-ACK and feature-store tests PASS on both sides.
- interest — exactly the 3 new behaviour tests FAIL; the malformed-payload and negative-amount ACK tests PASS on both sides.

(For interest this needed a throwaway shim so the new tests compiled against the old consumer, isolating behaviour from compilation. Not committed.)

## Verification

`ktlintCheck`, `detekt` and `test` all green for both modules, confirmed by reading which tasks actually appear in the output rather than the final status line:

```
./gradlew :openbank-interest-service:ktlintCheck :openbank-interest-service:detekt :openbank-interest-service:test
./gradlew :openbank-fraud-service:ktlintCheck :openbank-fraud-service:detekt :openbank-fraud-service:test
```

No API change, so no `openapi.yaml` bump. No DB or event-schema change. `version.txt` untouched — release-please cuts both from the `fix` type plus the `src/main` paths.

`openbank-interest-service` is money-path, so this needs 2 approvals per ADR-0030. **Not merging.**

Refs #5698

### DIFF
--- a/openbank-interest-service/src/main/kotlin/com/openbank/interest/infrastructure/kafka/WithholdingRemittanceSettlementConsumer.kt
+++ b/openbank-interest-service/src/main/kotlin/com/openbank/interest/infrastructure/kafka/WithholdingRemittanceSettlementConsumer.kt
@@ -13,6 +13,7 @@ import com.openbank.libs.persistence.outbox.OutboxKafkaHeaders
 import io.smallrye.mutiny.coroutines.awaitSuspending
 import jakarta.enterprise.context.ApplicationScoped
 import jakarta.ws.rs.core.Response
+import kotlinx.coroutines.delay
 import org.apache.kafka.clients.consumer.ConsumerRecord
 import org.eclipse.microprofile.config.inject.ConfigProperty
 import org.eclipse.microprofile.reactive.messaging.Incoming
@@ -58,6 +59,23 @@ import java.util.UUID
  * by transaction-service's own idempotencyKey field and this consumer's dedup key, so a Kafka
  * redelivery replays the same booking instead of double-remitting.
  *
+ * ## Failure handling — a bad payload and a bad rail call are not the same thing (#5698)
+ *
+ * A **malformed payload** is unretryable: a redelivery decodes identically and fails identically
+ * forever. It is logged and ACKED, and that is the only poison-pill case here.
+ *
+ * A **failed downstream call** is the opposite — the event is fine, the rail or the database is
+ * not, and a due regulatory tax remittance must still be booked once it recovers. This boundary
+ * used to swallow that too: the outer catch wrapped the whole of [bookAndSettle], so a transient
+ * transaction-service failure was logged and the message ACKED, stranding the batch PENDING
+ * forever with the event gone. Nothing re-drives it — there is no endpoint, only SQL. So every
+ * downstream call now goes through [withBoundedRetry] and, if it still fails, is RETHROWN, letting
+ * the connector dead-letter it into something a human can see. Same shape as kyc-service's
+ * `PartyEventConsumer` (#5698/#5699).
+ *
+ * The idempotency key below is what makes both a retry and a redelivery safe: transaction-service
+ * answers 409 for a batch already booked, and this consumer treats that as success.
+ *
  * **Scope note (deliberate, tracked as follow-up — see issue #999):** this consumer books the cash
  * leg only. It does NOT assemble or file the statutory §38d *Vyúčtování daně vybírané srážkou* XML
  * with the finanční úřad — that is a separate regulatory-reporting concern (issue #999 flags
@@ -78,30 +96,92 @@ class WithholdingRemittanceSettlementConsumer(
     private val log = Logger.getLogger(WithholdingRemittanceSettlementConsumer::class.java)
 
     @Incoming("interest-withholding-remitted-in")
-    // Poison-pill safety: this boundary must swallow ANY failure (parse, rail call) and ack, so
-    // one bad event cannot wedge the consumer group.
+    // Only the DECODE below is allowed to swallow-and-ack (a malformed payload can never succeed on
+    // a redelivery). Every downstream call is retried and then rethrown — see the class KDoc.
     @Suppress("TooGenericExceptionCaught")
     suspend fun consume(record: ConsumerRecord<String, String>) {
-        try {
-            // ADR-0050 N3: the event type travels ONLY as the ce-type header — the payload has no
-            // eventType field (see KafkaInterestOutboxEventPublisher), so filter on the header.
-            val eventType = record.headers().lastHeader(OutboxKafkaHeaders.HEADER_EVENT_TYPE)
-                ?.let { String(it.value(), StandardCharsets.UTF_8) }
-            if (eventType != EVENT_WITHHOLDING_REMITTED) return
+        // ADR-0050 N3: the event type travels ONLY as the ce-type header — the payload has no
+        // eventType field (see KafkaInterestOutboxEventPublisher), so filter on the header.
+        val eventType = record.headers().lastHeader(OutboxKafkaHeaders.HEADER_EVENT_TYPE)
+            ?.let { String(it.value(), StandardCharsets.UTF_8) }
+        if (eventType != EVENT_WITHHOLDING_REMITTED) return
 
-            val node = objectMapper.readTree(record.value())
-            val remittanceId = UUID.fromString(node.path("remittanceId").asText())
-            val totalTaxAmount = decimalOf(node.path("totalTaxAmount"))
-            val itemCount = node.path("itemCount").asInt()
-
-            if (totalTaxAmount.signum() <= 0) {
-                settleNilOrRefuse(remittanceId, totalTaxAmount, itemCount)
-                return
-            }
-
-            bookAndSettle(remittanceId, debitRequestFor(node, remittanceId, totalTaxAmount))
+        val batch = try {
+            decode(record.value())
         } catch (e: Exception) {
-            log.errorf(e, "[withholding-remittance] failed to handle remitted event: %.300s", record.value())
+            // Unretryable: log and ack. This is the ONLY swallowed failure in this consumer.
+            log.errorf(e, "[withholding-remittance] undecodable remitted event, acking: %.300s", record.value())
+            return
+        }
+
+        withBoundedRetry(batch.remittanceId) {
+            if (batch.totalTaxAmount.signum() <= 0) {
+                settleNilOrRefuse(batch.remittanceId, batch.totalTaxAmount, batch.itemCount)
+            } else {
+                bookAndSettle(
+                    batch.remittanceId,
+                    debitRequestFor(batch.node, batch.remittanceId, batch.totalTaxAmount),
+                )
+            }
+        }
+    }
+
+    /** The fields this consumer needs off the event; throws on anything unparsable (see [decimalOf]). */
+    private data class RemittedBatch(
+        val node: JsonNode,
+        val remittanceId: UUID,
+        val totalTaxAmount: BigDecimal,
+        val itemCount: Int,
+    )
+
+    private fun decode(payload: String): RemittedBatch {
+        val node = objectMapper.readTree(payload)
+        return RemittedBatch(
+            node = node,
+            remittanceId = UUID.fromString(node.path("remittanceId").asText()),
+            totalTaxAmount = decimalOf(node.path("totalTaxAmount")),
+            itemCount = node.path("itemCount").asInt(),
+        )
+    }
+
+    /**
+     * Retry [block] a bounded number of times, then RETHROW so the connector dead-letters.
+     *
+     * The rethrow is the point. A caught-and-logged failure acks the message, and an acked message
+     * that did no work is indistinguishable from one that succeeded — from Kafka, from the consumer
+     * lag metric, and from every dashboard built on either. The only trace was an ERROR line nobody
+     * alerts on, while a due tax remittance sat PENDING with no event left to re-drive it (#5698).
+     */
+    @Suppress("TooGenericExceptionCaught") // the retry is deliberately type-agnostic; it rethrows
+    private suspend fun withBoundedRetry(remittanceId: UUID, block: suspend () -> Unit) {
+        var attempt = 1
+        while (true) {
+            try {
+                block()
+                return
+            } catch (e: Exception) {
+                if (attempt >= MAX_ATTEMPTS) {
+                    log.errorf(
+                        e,
+                        "[withholding-remittance] batch %s failed after %d attempts (%s: %s) — dead-lettering",
+                        remittanceId,
+                        attempt,
+                        e.javaClass.simpleName,
+                        e.message,
+                    )
+                    throw e
+                }
+                log.warnf(
+                    "[withholding-remittance] batch %s failed (attempt %d/%d, %s: %s) — retrying",
+                    remittanceId,
+                    attempt,
+                    MAX_ATTEMPTS,
+                    e.javaClass.simpleName,
+                    e.message,
+                )
+                delay(RETRY_BACKOFF_MS * attempt)
+                attempt++
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


JiRaska:

- fraud `transaction-signal`

example:

   meant to park."*

JiRaska:
From mapping the #5698 sweep's merge order tonight (full analysis on #5698): **this branch is 47 commits behind `main`**, and that staleness is what produces its 81–83 path "conflicts" with the other sweep PRs — `.release-please-manifest.json`, `CHANGELOG.md`, `package.json` churn rather than semantic overlap. A rebase should collapse almost all of it.

Worth doing before it lands, because #5726 declares a merge-order dependency on this PR (the `interest` `remitUseCase.settle` finding its gate reports is the one this fixes on the same lines), and the release-manifest churn is exactly the shape where a clean merge can silently drop a list entry.

JiRaska:
## Review — right change; the ordering dependency on the DLQ is the thing to get right

The diagnosis is correct and the KDoc rewrite is honest about it: the old text claimed "Failures are DLQ'd" while nothing rethrew, so nothing was. An acked message that did no work is indistinguishable from a processed one — from Kafka, from consumer lag, and from every dashboard built on either. Fixing the code to match the doc rather than the doc to match the code is the right direction.

The distinctions drawn are the right ones and are drawn per-call rather than per-handler:

- **Malformed payload → log and ack.** Genuinely unretryable; a redelivery decodes identically forever. In `WithholdingRemittanceSettlementConsumer` this is properly narrowed to just `decode()`, where the old code wrapped the *whole* of `bookAndSettle` in the same catch. That widening was the actual bug.
- **Downstream failure → bounded retry, then rethrow.** Correct for both consumers.
- **`FeatureOnlineUpdater` → still swallowed, and now truthfully labelled.** Right call: it is the ADR-0140 shadow plane, feeds no live decision, and dead-lettering a real velocity signal because a shadow write failed would trade a working control for an experimental one.
- `RemittanceBookingFailedException` as a named type rather than a bare `RuntimeException` — greppable, and avoids suppressing `TooGenericExceptionThrown`. The comment at the `else ->` branch is the sharpest thing in the diff: the old code's promise that the batch "stays PENDING for retry (a redelivery of this event)" could never happen, because the ack had already destroyed the event and no endpoint re-drives a batch.

### The dependency to state explicitly: a rethrow without a DLQ is worse than the ack

SmallRye's default `failure-strategy` is `fail`, which **stops the channel**. So this PR is only an improvement where the channel is actually wired for dead-lettering. I checked both:

- **interest-service** `interest-withholding-remitted-in` — wired by **#5751**, which is not merged. On `main` today this channel has no DLQ. If #5715 lands first, a transient transaction-service failure stops the withholding-remittance consumer group instead of dropping one event. That is arguably the better failure (loud, not silent), but it is not what the KDoc says happens, and it is a money-path channel. **Please state the merge order dependency on #5751 in the PR body, or land them together.**
- **fraud-service** `transaction-signal` — already has `failure-strategy: dead-letter-queue` on `main` (`application.yaml:103`), an explicit DLQ topic via `fraud-service-msg-override.yaml:47`, and a `[Write, Describe]` ACL at `kafka-fraud-mtls.yaml:38-42`. Three of the four parts are there. **The fourth is missing: there is no `KafkaTopic` CR for `openbank.transactions.transaction.initiated.fraud.dlq` anywhere** — `openbank-infra/gitops/components/fraud-service/kafka-topics.yaml` contains only `openbank.fraud.hold.changed`, and #5751 does not add it (fraud is in that gate's `KNOWN_UNWIRED` list, with a reason that does not actually apply — I've commented there). This cluster does not auto-create topics, so once this PR makes the consumer rethrow, the DLQ send itself fails and the consumer wedges on the very failure it was meant to park. One CR fixes it, and it is unblocked.

### Finding — the fraud KDoc's redelivery trade-off is about to be false

The new KDoc argues the redelivery cost is acceptable because "`velocity_aggregates` has no dedup — a redelivery re-adds the amount", and reasons from there that overcounting is the safe direction. **#6040 adds exactly that dedup**, and #6061 changes the bucketing on top. All three PRs edit this same file from the same base blob, as does #5737.

Two consequences. First, the paragraph will be stale the moment #6040 lands, and it is the paragraph justifying the whole trade-off. Second, four PRs editing one file will merge textually cleanly in some order and produce a `TransactionSignalConsumer` nobody reviewed as a whole — #6061 edits the very `try/catch` blocks this PR replaces with `withBoundedRetry`. Whoever lands second should diff the merged result rather than trust the clean merge.

### Minor

- `withBoundedRetry` is duplicated verbatim in both services (`MAX_ATTEMPTS = 3`, `RETRY_BACKOFF_MS = 500`, same linear backoff, same log shape). Two copies is tolerable; if a third consumer joins the sweep this belongs in `openbank-libs-runtime`.
- The retry is in-handler and synchronous, so worst case it holds the partition ~1.5 s before dead-lettering. Fine at these numbers, but it does mean `MAX_ATTEMPTS` and the poll interval are now coupled — worth a comment so nobody raises the attempts without checking `max.poll.interval.ms`.
- Both are hardcoded rather than `@ConfigProperty`. Deliberate? A DB outage longer than 1.5 s now dead-letters rather than retries, which is the intended behaviour, but it is a tuning knob you may want without a redeploy.
- CI: 29 success, 0 failures; skips are path-scoped CI and Dependabot automerge, not test skips.

JiRaska:
## Not resolved — this one needs an author decision, not a merge

I resolved the conflicts on #5723, #5759 and #5726 today. I deliberately left this one alone, because its conflict is not textual: resolving it *is* the decision about whether fraud-service may rethrow.

### The conflict

`main` has moved a long way under both conflicted files. `TransactionSignalConsumer` has since taken **#5786** and **#6040** (dedupe against the applied-id set) and **#6044** (`occurredAt` handling, the `recordSignalMissingEventTime` meter, and a `recordTransaction` signature that now takes `eventTime` — this branch still calls the 4-argument form). Merging is therefore a rewrite of this PR's fraud half against a different function, not a hunk pick.

### The blocker — the fraud DLQ topic does not exist

This PR makes the `transaction-signal` channel **rethrow**. That hands the message to the connector's failure strategy, which for this channel is:

```
failure-strategy: dead-letter-queue                                      # application.yaml
mp.messaging.incoming.transaction-signal.dead-letter-queue.topic=\
  openbank.transactions.transaction.initiated.fraud.dlq                  # fraud-service-msg-override.yaml
```

Every occurrence of that topic name on `main` (`579dda8e0`):

| where | what |
|---|---|
| `fraud-service-msg-override.yaml:47` | the DLQ topic setting |
| `kafka-fraud-mtls.yaml:40` | KafkaUser **Write/Describe** ACL |
| `TransactionSignalConsumerBootIT.kt` | a test subscribing to it |

There is **no `KafkaTopic` CR**. I enumerated every `kind: KafkaTopic` document under `openbank-infra/gitops` — **48 of them** — and this topic is not among them; `payment.scheme-accepted.dlq` is, which is the positive control proving the probe works. And `kafka.yaml:178` states the cluster **does not auto-create topics**.

So today the rethrow dead-letters into a topic that does not exist. That is worse than the ack it replaces: the ack loses one signal quietly, whereas a failing DLQ send fails the channel.

**#5751 does not cover it.** It adds 35 DLQ `KafkaTopic` CRs and fraud is not one of them — it does include `openbank.dlq.interest.interest-withholding-remitted-in`, so this PR's **interest half is fine once #5751 lands**. Note also that #5751's `KNOWN_UNWIRED` entry justifies fraud as a *"pre-#5745 implicit DLQ name"*, which is factually wrong: fraud is explicitly named **and** ACL'd. Its real gap is the missing `KafkaTopic` CR, and that wording is what hides it.

### What I suggest

1. Add the `KafkaTopic` CR for `openbank.transactions.transaction.initiated.fraud.dlq` (here or in #5751), and fix #5751's `KNOWN_UNWIRED` reason.
2. Then rebase the fraud half onto main's current `recordTransaction(..., eventTime)`.

The interest half could be split out and merged ahead of that, since its DLQ topic is covered.

I have not pushed anything to this branch.

JiRaska:
Replanted on live `origin/main` (head `722d7a9ae`), and **narrowed to the interest-service half**.

Measured reason for dropping the fraud half:

- `openbank-fraud-service/.../TransactionSignalConsumer.kt` on this branch was written against a
  pre-#6040/#6044 version of the file. Diffed against live `main` it *removes* `FraudMetricsPort`,
  the `recordSignalMissingEventTime()` counter, and the `eventTime` argument to
  `velocityRepo.recordTransaction(...)` — i.e. it reverts the applied-id/occurredAt work that has
  since merged.
- The same rethrow it introduces is already delivered, correctly rebased on that newer code and via
  the shared `com.openbank.libs.messaging.EventRetry`, by #5719 (commit `791d2a91b`,
  "fix(account,audit,fraud): stop acking lost money, lost audit rows and lost fraud signals").
  So the fraud half here is superseded, and merging it would be a regression rather than a duplicate.

What remains is unique to this PR and touched by no sibling: the
`WithholdingRemittanceSettlementConsumer` bounded-retry-then-rethrow, the
`RemittanceBookingFailedException`, and the tests that replace the ones asserting the swallow.

Verification on the replanted head: `:openbank-interest-service:test` 110 tests / 0 failures /
0 skipped; `ktlintCheck` + `detekt` BUILD SUCCESSFUL.

Title updated to `fix(interest): ...` to match the scope that is actually shipped.

JiRaska:
Re-checked against #5719 at its rebuilt head `ec0474b0b`, per the coordinator's note. Blob compare:
this PR's file is at `dd4a3545d` on BOTH `origin/main` and #5719 — i.e. #5719 does not touch
`openbank-interest-service` at all, so the remaining interest-service payload here is not contained
in it and is not a duplicate. (The fraud half, which #5719 *does* supersede, was already removed —
see the previous comment.)

Extra evidence that the two are complements: running #5726's widened
`check-event-handler-swallows.py` over the tree reports exactly one finding fleet-wide —
`WithholdingRemittanceSettlementConsumer.kt:85`, the site this PR fixes. Without this PR, #5726's
own gate is red.

No further changes; head stays `722d7a9ae`.

JiRaska:
Checked this against #6328 (which was filed as a separate tracker for this exact handler, believing the fix was blocked behind #5751 → #5726 → #5719).

**The blocker has cleared.** #5751 is merged, and the DLQ for `interest-withholding-remitted-in` is genuinely wired — all four parts, verified in the repo *and* live: `failure-strategy: dead-letter-queue`, an explicit **nested** `dead-letter-queue.topic` (so the #686 dotted-leaf trap does not apply), a `KafkaTopic` CR that is `Ready=True` in `messaging`, and `KafkaUser/interest-service` holding `[Write, Describe]` on `openbank.dlq.interest.interest-withholding-remitted-in`. So the rethrow in this PR parks the record rather than wedging the partition.

Two notes for whoever merges:

1. **#5719 changes the same file the same way**, via the shared `com.openbank.libs.messaging.EventRetry` rather than this PR's consumer-local `withBoundedRetry`. (#5726, which #6328 names as carrying the fix, merged into #5719's branch.) If both land unreconciled, this consumer ends up with two retry mechanisms over the same call sites. Worth picking one — #5719's is the fleet-consistent shape, this one is smaller and already `CLEAN`.
2. On a 3-dot diff this PR is exactly 2 files and touches neither `application.yaml` nor `version.txt`, so it does **not** revert #5751's wiring. (A 2-dot diff makes it look like a ~18k-line revert; that is the stale-base artifact.)

Context that may matter for urgency, from a read-only look at the live `interest` database: `withholding_remittance` and `withholding_tax` are both empty, and `interest_outbox_seq.last_value` is NULL — the pooled sequence has never been drawn, so no insert was ever *attempted*. This path has never executed in that environment, so nothing has been swallowed yet. Latent, not a live loss.

What made it worth fixing anyway: the batch is unrecoverable by any automatic route once it fires. `assembleRemittance` short-circuits on the existing batch (*"re-mark nothing"*), the underlying rows are already `RECORDED -> REMITTED` so they cannot be swept into a new batch, no scheduler reads `PENDING` remittances, and the outbox already published successfully — the failure is downstream of it. Only SQL gets it back. And nothing would page: the service registers no `MeterRegistry` metric anywhere in `src/main`, and the outbox backlog gauge measures the leg that succeeded.

JiRaska:
Closing as superseded by #5719, merged as `7b1c78d5b`. It fixed the same two files with the same
intent, so this branch went `DIRTY` on that merge and has nothing left to add.

Both changes draw the identical distinction — a malformed payload is unretryable and gets acked, a
failed downstream booking is retried and **rethrown** so the connector can dead-letter it, because
the batch otherwise stays `PENDING` with no re-drive path and the tax goes unpaid. The wording in
this PR's KDoc and in #5719's inline comment are independent statements of the same finding.

Where they differ, `main`'s version is the one to keep:

- #5719 routes the retry through the shared `com.openbank.libs.messaging.EventRetry`; this branch
  introduces a local `withBoundedRetry` in the consumer. One fleet-wide helper beats a per-service
  copy for a behaviour that now applies across many consumers.
- #5719 landed the same treatment across the fleet in one pass and added the gate that catches the
  next occurrence, rather than fixing this one site.

On coverage, nothing is lost. `WithholdingRemittanceSettlementConsumerTest` on `main` now carries
ten cases, including the one that is the whole point:

```
a non-2xx-non-409 response is retried and rethrown, never settled or acked
a 409 from transaction-service still settles the batch (idempotent success)
a malformed payload is swallowed, not thrown
```

This branch's diff adds no test name that is not already there.

Worth recording why this collided rather than merging: #5719 was a fleet sweep over exactly this
defect class, so any single-site fix for the same class was going to be overtaken. Three other PRs
went `DIRTY` on the same merge. Not avoidable in hindsight — the sweep is the better outcome — but
when a sweep PR is open for a defect class, a per-service fix for one instance of it is work that
will most likely be thrown away.

PR #5719, opened 2026-08-19
fix(fleet): stop event handlers acking work they did not do, and gate it

JiRaska:
Closes the code half of #5698, and gates the class so a new service cannot reintroduce it.

## The incident

`PARTY_CREATED` for a real onboarding customer arrived at 12:56:15 on 2026-08-19 while `kyc-db-rw` was refusing connections. The handler caught the exception, logged `Failed to auto-open/screen KYC case`, and returned normally — **which acks the message**.

No KYC case ⇒ party stays `PENDING_KYC` ⇒ no `PARTY status=ACTIVE` ⇒ `account-service` never activates ⇒ the welcome bonus, which lives inside the activation branch, never fires. The customer had two accounts that looked open and no money in them.

Nothing retried, and that is measured rather than assumed: `kyc_cases_seq.last_value` never advanced past the block already in use, so no INSERT was attempted after the failure.

**One** confirmed stranded party — the reported customer. I first wrote "10 of 73, the oldest for ten weeks"; checking the other nine rather than counting them, six predate the auto-open consumer (added 2026-06-26) and three are account-less e2e fixtures. Corrected in the issue and in every place the number was quoted.

## Why it is invisible

An acked message that did nothing is indistinguishable from a successful one — not in consumer lag, not in the DLQ (which stays empty), not on any dashboard built on either. The only trace is an ERROR line, and nothing pages on ERROR lines.

Worse: **two green tests asserted the swallow**, one justifying it with *"the message is acked and the party stream can replay"*. Nothing replays it. The suite documented the bug as intended behaviour, so no amount of test-reading would have found this — only the live DB did.

## The line this draws

A **malformed event** is unretryable — replaying it fails identically forever — so log-and-ack is correct. That is the poison pill, and it was never the problem; sharing a catch with a DB outage was.

Everything else is a dependency failing. Hexagonal architecture is exactly why that must not be logged away: a database or cache being down is a **normal event** for an adapter, and the port contract is "the work happens, or somebody finds out".

## What changed

`openbank-libs-runtime` gains **`EventRetry.withRetry`** — bounded retry, then **rethrow**, so the connector retries and finally dead-letters. One implementation, replacing the near-identical private copies the fleet had grown.

**15 call sites across 10 services** converted:

| service | what was being acked away |
|---|---|
| kyc | the KYC case that gates activation |
| party ×2 | the KYC/AML decision that flips a party to ACTIVE |
| party (marketing consent) | a lost revocation ⇒ still sending marketing to someone who withdrew consent |
| aml ×2 | onboarding screening; AML case PII erasure |
| audit | `audit_entries` rows — the evidentiary record |
| card-issuance, notification, copilot | GDPR Art. 17 erasures logged as done over PII that stayed |
| fraud ×2 | velocity aggregate and payee history — the consumer's own product |
| account | **the welcome bonus itself**, logged as "will retry on next ACTIVE event" — there is no next ACTIVE event |

Two legitimate silences remain, and both must now be **stated**: `best-effort:` (the event is complete without it — the push after the money is booked) and `observed-by:` (the failure is visible through something else — the copilot retention sweep leaves its workflow-liveness gauge climbing until `WorkflowLivenessStale` fires).

## The gate

`event-handler-swallows` (kotlin shard, **enforced**, 170 handler files in scope) flags a `catch(Exception|Throwable)` with no throw wrapping a state change, in any file carrying `@Incoming` **or `@Scheduled`**.

It earned its place twice while being written:

1. **It found three cases the manual sweep missed**, because the sweep only looked at `@Incoming`: a `@Scheduled` GDPR retention sweep, the marketing-consent projection, and `libs-domain`'s `OutboxDispatch` — the last of which turned out to be **correct** (its catch calls `markFailed`, recording the failure durably and leaving the row processable), so the gate learned that shape rather than demanding a rethrow that would have been wrong.

2. **It then caught me.** Its first version excused account-service's welcome bonus, because the word "Best-effort" appeared in a KDoc describing the *notification* three lines below — and the money-losing swallow above inherited the excuse. The marker now counts only inside the catch, or in the contiguous comment block immediately above the try. A rule that can be satisfied by a nearby unrelated sentence is not a rule.

Self-test: 8 fixtures, **five negative** (rethrow, parse-only, durable-record, best-effort, observed-by) plus a corpus-reachability assertion, so a broken walk cannot report a clean fleet.

## Verification

- `run-gates.py --group kotlin` — 20/20
- `run-gates.py --group lint` — 19/19
- `check-event-handler-swallows.py` — clean: 170 handler files, 0 swallowed state changes (was 15)
- tests + ktlint + detekt for all ten touched services

## Supersedes

#5699 fixed kyc-service alone with a service-local helper. Closing it in favour of this — two implementations of the same retry would be the duplication this removes.

## Still open in #5698

- the 9 older orphaned parties (live remediation, same replay path already used for the reported one)
- detection: nothing today notices a party that never got a case


JiRaska:
Ran the gate from this branch against `main` to see what it *cannot* see, rather than only that it
passes. `SUBJECTS=170` / `clean` — but two money-path consumers hold the exact #5698 shape and are
not flagged. Both are in `rules.yaml: money_path_services`.

## 1. `openbank-sdd-service` — verb missing from `STATE_CALL`

`SddCollectionDebitConsumer.consume` (line ~111 on `main`):

```kotlin
val response = transactionClient.initiateTransaction(request).awaitSuspending()
…
} catch (e: Exception) {
    log.errorf(e, "[sdd-collection-debit] failed to handle collection-authorised event: %.300s", payload)
}
```

The receiver `transactionClient` matches, but **`initiate` is not in the verb alternation**
(`save|persist|insert|update|create|open|activate|grant|credit|debit|post|record|apply|…|book|settle|publish|send`),
so `STATE_CALL.search(try_body)` returns `None` and the finding is dropped.

This is the fleet's most direct money mover — `CLAUDE.md` describes this very consumer as *"posts a
real debit via TransactionServiceClient for every authorised direct-debit collection — moves funds
directly"*. A transient transaction-service failure is logged and **acked**: an authorised
direct-debit collection silently never debits.

Its non-2xx branch has the same shape one level down — it logs *"an authorised collection did not
move money … requires manual investigation"* and returns normally, which acks. That is the identical
"a redelivery that could never arrive" trap this PR fixed in `WithholdingRemittanceSettlementConsumer`.

## 2. `openbank-standing-order-service` — state call indirected through a private helper

`StandingOrderDueConsumer.consume`:

```kotlin
try {
    …
    when (val paymentType = node.path("paymentType").asText()) {
        "SEPA_CREDIT" -> executeSepa(orderId, node)          // <- private method
        "DOMESTIC", "INTERNAL" -> executeDomesticOrInternal(orderId, node)
    }
} catch (e: Exception) {
    log.errorf(e, "[standing-order-exec] failed to handle due event: %.300s", payload)
}
```

`executeSepa` internally calls `sepaClient.createPayment(idempotencyKey, request).awaitSuspending()`
— which **would** match `STATE_CALL` if it were inline. Because the gate scans only the `try` body's
text, **one level of method indirection hides it**. A failing standing-order execution is logged and
acked; the order is not re-driven.

## Suggested fixes

1. Add `initiate` (and probably `execute`, `transfer`, `charge`, `reverse`, `capture`) to the verb
   alternation — cheap and closes case 1.
2. Case 2 is the structural one. A receiver-name regex over the `try` body cannot see a state change
   behind a helper. Two options that stay honest about it: resolve one level of same-file private
   method calls appearing in the `try` body and scan their bodies too; or, if that is more machinery
   than wanted, **declare the limitation in the script header and require an explicit
   `best-effort:`/`observed-by:` marker on any handler whose `try` body contains only local method
   calls** — turning the invisible case into one a human must sign.

Worth noting the gate is otherwise well-built — the `HANDLED_IN_CATCH` allowance for `markFailed`,
and the fix after the marker wrongly excused account-service's welcome bonus, are both exactly right.
Checked `openbank-balance-service`'s `LedgerProjectionConsumer` too: its two catches wrap only
`readTree`/`toChange` and `projection.apply(change)` sits **outside** the `try`, so it is a correct
true negative, not a third hole.

Method: enumerated all 44 `@Incoming` files, extracted every `catch (Exception|Throwable)` whose
block contains no `throw` (57 sites), subtracted the files this PR and #5699/#5715 touch, then read
the money-path remainder.

JiRaska:
Both blind spots are now fixed in **#5726**, stacked on this branch — the verb list plus a `_reaches_state_change` that follows same-file helpers to depth 2 (cycle-safe). 

With that, the gate found **5 sites, 3 more than my manual sweep** — including a `@Scheduled` in balance-service my own `@Incoming`-only pass had missed, exactly the way the original manual sweep missed `@Scheduled`. Fixed sdd + standing-order with `EventRetry`, and gave balance + sanctions `observed-by:` markers rather than rethrows, since both are the legitimate-silence case this gate's markers exist for.

4 new self-test cases (12/12), including two negatives so the helper-following cannot invent findings or hang on recursion. Happy for it to be folded into this PR instead of stacked if you'd rather keep it one change.

JiRaska:

- fraud `transaction-signal`

example:

   meant to park."*

JiRaska:
Pushed one commit to this branch — it was red on `build (openbank-notification-service)` (182 tests, 1 failed), which blocks the whole sweep stacked behind it. Took the liberty because the branch had been idle ~9h with no worktree left on this machine; revert it freely if you had something else in mind.

**`PartyErasureConsumerTest > erasure failure is swallowed to protect the consumer group` was still asserting the pre-fix behaviour**, so this branch's own (correct) change to rethrow turned it red. Verified the consumer first, so this is a stale test and not a wrong fix: `PartyErasureConsumer` does `EventRetry.withRetry { … }` around both deletes, no catch.

The test name is the defect statement. Returning normally acks the message, so a failed **GDPR Art. 17** erasure left device tokens and notification history in place while the log recorded the erasure as done — and an acked message is indistinguishable from a successful one from outside.

**Third instance of this shape in the sweep**: #5698 documents two in kyc-service, sdd's `a non-2xx-non-409 response is handled without throwing` was the same (fixed in #5726). Worth stating plainly for whatever remains — in this defect class the tests are not a safety net, they are part of the evidence, because each was written to *describe* the swallow rather than to catch it. Anywhere this sweep changes a handler to rethrow, expect a green test asserting the opposite.

Now asserts both halves: bounded retry (`DEFAULT_MAX_ATTEMPTS`) and the rethrow. 4/4 green, 0 skipped; module suite + ktlint + detekt green run serially.

JiRaska:
**Scope observation, not a defect claim** — offered for wherever this sweep draws its boundary.

The gate's subject set is files carrying `@Incoming` or `@Scheduled` (170 files). **Temporal
activity implementations have the same "return normally = success" semantics** — an activity that
swallows a failure tells the workflow the step succeeded, exactly as an ack tells Kafka the message
was handled — and they are structurally outside that set. 50 activity files on `main`; scanning
them with this gate's own catch-without-throw + STATE_CALL shapes finds 4 sites, all money-path:

| site | verdict on reading it |
|---|---|
| `PaymentActivitiesImpl.releaseHold` / `.reverseJournal` (transaction) | **legitimate and well argued** — the comment names the recovery: balance-service TTL-expires the hold and the ledger reversal is idempotent. It even literally says "Best-effort", so it would satisfy `EXCUSED` unchanged if it were in scope. |
| `DomesticPaymentActivitiesImpl.openCaseQuietly` (domestic-payment) | AML case creation swallowed. The payment itself is **not** fail-open — `screenPayment` still returns BLOCK/REVIEW, so it is held. What can silently not exist is the case an analyst works from. |
| `SepaPaymentActivitiesImpl` same shape (sepa-payment) | as above |

For the two AML ones I went looking for whether anything observes the outcome before calling it a
gap, and something does: `DomesticPaymentStrandedGauge` (#3273) counts payments parked per status
with an oldest-age series, so a payment held in REVIEW with no case to action it shows up there.
That makes them the `observed-by:` case rather than a defect — **but nothing in the code says so**,
and that is precisely the difference this gate exists to create: turning an unverified prose
justification into a declared marker something checks.

So the suggestion is only: consider whether `@ActivityInterface`/`@ActivityMethod` belongs in
`HANDLER` alongside `@Incoming`/`@Scheduled`. On today's tree it would cost 4 markers, three of
which already have their reasoning written and one (`releaseHold`) would pass untouched. Entirely
reasonable to leave out of this PR and note it as a follow-up — flagging it because the subject set
is the one thing a gate cannot discover about itself.

Method: enumerated the 50 activity files, excluded any already carrying `@Incoming`/`@Scheduled`,
applied the catch-without-throw + state-call shapes, then **read each of the 4** rather than
reporting the count.

JiRaska:
### CI triage: the `gates (api)` red here is a stale verdict, not a defect in this diff

`gates (api)` fails with:

```
openapi enum drifts from the domain enum `SettlementStatus` it serves:
spec advertises values the code does not have: ['LEDGER_REVERSED']
```

That is the gate bug fixed on `main` by #6220 (merged 14:57Z, `b85ac16`). This PR does not
touch `SettlementStatus` or any settlement spec.

**Measured rather than assumed:** this PR's `refs/pull/5719/merge` ref does not contain
`b85ac16`, so the run checked out the pre-fix gate script and its verdict predates the fix.

`Validate manifests` is a roll-up of the `gates (...)` shards, so it is red only as a
consequence — not an independent failure.

**Action:** `gh run rerun` will not clear this — it re-runs the recorded (stale) merge SHA.
Only a push or a base update refreshes the merge ref. Rebase on `origin/main` and the shard
should go green.

JiRaska:
Rebuilt on current `origin/main` (head `6346f943a`) after #6209 landed. **One file was dropped from
this PR's scope on purpose, and it needs your agreement rather than mine.**

`openbank-audit-service/.../AuditConsumer.kt` was the only conflict, and it is not a textual one.
`main` (from #6209) restructured that class and now states the opposite intent to this PR:

```kotlin
suspend fun consume(message: Message<String>) {
    try { persist(payload, addressOf(message)) }
    catch (e: Exception) {
        // This legacy multi-producer channel deliberately retains its historic availability
        // behaviour. D5 agent events use AgentAuditConsumer instead: it acknowledges only a
        // successful durable write, so a store failure is retried by Kafka rather than lost.
```

This PR's sweep wanted that same consumer to rethrow. #6209 instead split the stream: the legacy
mixed channel keeps acking, and the strict path moved to `AgentAuditConsumer`. That decision was
made *after* this branch was written, so resolving the conflict either way is a design call, not a
merge. I took **`main`'s version verbatim** — the file is now byte-identical to `origin/main` — and
left the other 31 files as they were.

Two things that makes true, both measured rather than assumed:

- The gate this PR adds is still clean with `main`'s version:
  `[event-handler-swallows] clean: 176 handler files, no swallowed state changes` (rc=0). So
  dropping the file costs the gate nothing; `AuditConsumer` is not a swallow the gate wants to flag.
- 3-dot delta against `main` is 31 files, 872 insertions / 343 deletions — the sweep is intact.

**What to decide:** whether the legacy audit channel should stay availability-first. #6209's comment
argues yes and names the replacement path; this PR's premise argued no, fleet-wide. If you want the
legacy channel to rethrow too, that is a follow-up on top of #6209's structure, not a revert of it.

Also worth knowing for merge order: #5726 and #5759 are stacked on this branch, so they land after
it. #5759 currently reads `CLEAN` in the PR list — that is clean against *this branch*, not against
`main`; merging it directly would merge into this branch.

JiRaska:
`build (openbank-fx-service)` is red here on coverage, and **this PR does not touch fx-service** —
`git diff --name-only origin/main...<head> | grep -c fx-service` is `0`. Measured cause:

```
CI on this branch:  Rule violated: lines covered percentage is 60.504200, but expected minimum is 65
clean origin/main:  ./gradlew :openbank-fx-service:koverVerify  ->  rc=0
```

**fx-service's coverage report includes `libs-runtime` classes.** From the Kover binary report of the
passing run on `main`:

```
$ strings openbank-fx-service/build/tmp/koverCachedVerify/agg-ic.ic | grep -oE 'com/openbank/[a-z]+' | sort -u
com/openbank/fx
com/openbank/libs
```

So the two files this PR adds to `openbank-libs-runtime/.../messaging/EventRetry.kt` count as
**uncovered** in fx's report — fx's own test task does not run `libs-runtime`'s tests, however well
`EventRetryTest` covers them in its own module. Every dependent service's measured coverage moves
when libs grows; fx broke first because it sits closest to its floor (65, ratcheted from 40).

Two honest options, and the choice is not mine to make:

1. **Cover the new `EventRetry` lines from a dependent's perspective** — i.e. accept that shared code
   needs coverage that dependents' test runs actually exercise. Lifts every dependent at once.
2. **Treat the fx floor as coupled to libs size** and adjust it, with that reasoning written down —
   the current comment justifies 65 from a measurement ("real line coverage measured ~69%") that a
   libs change silently invalidates.

What should not happen is a floor lowered quietly to make one PR green: the ratchet is the only thing
keeping this honest, and the next libs addition will hit the next-closest service.

One caveat on my own evidence: I also tried to reproduce the failure locally on this branch's head and
that run died with `Java heap space` in `:openbank-fx-service:test`, so it produced no report. I am
not counting it — the numbers above are CI's verdict and a completed local run on `main`.

Also worth knowing: this branch's head moved to `ccdb713c4` — its stacked child #5759 was merged into
it.

JiRaska:
**Correcting myself: I had the direction of this effect backwards, and the fix I pushed made it
worse. Reverted in `79a006153`; `openbank-fx-service/build.gradle.kts` is byte-identical to `main`
again.**

CI measured both sides on the same head:

```
libs INCLUDED (as on main):   lines covered percentage is 60.504200, minimum is 65
libs EXCLUDED (my change):    lines covered percentage is 30.905700, minimum is 65
```

So fx's tests exercise a great deal of `com.openbank.libs` code and cover it **well** — libs was
holding fx's aggregate **up**, not dragging it down. fx's own line coverage is about **31%**, and its
floor of 65 has only ever been met with libs counted in.

That inverts the claim in this issue's title. The accurate statement is narrower:

- Adding *uncovered* libs lines does lower every dependent's aggregate — #5719's 13 new
  `EventRetry` lines moved fx from above 65 to 60.5, which is real and is why an unrelated PR went
  red on a service it never touched.
- But the aggregate is **above** fx's own coverage, not below it. Excluding libs does not "make the
  number mean what its name says" — it reveals that fx's own tests cover ~31% of fx, and that the
  65 floor was always measuring the pair.

Which makes option 1 in the original text actively wrong: scoping each service's report to its own
package would put fx (and probably others) far below their floors immediately. The real options are
narrower — measure shared code where it is tested and keep it out of dependents' *floors* without
pretending the dependents are better covered than they are, or set per-module floors that are
honest about what they include.

**How I got it wrong, since the method matters more than the conclusion:** I ran
`koverVerify` locally with the exclusion, got `rc=0`, and reported that as verification. It was a
false green — most plausibly a reused `koverCachedVerify` report. The negative control that would
have caught it (the same run *without* the exclusion) had died on `Java heap space`, so nothing ever
falsified the result. I said at the time that the control was void; I should have treated the
positive as void too, rather than pushing on the strength of it.

Issue #5705, opened 2026-08-19
settlement-service (money-path) emits no metrics and has no alerts of its own


JiRaska:
## Finding

`openbank-settlement-service` — a declared money-path service — emits **no application metrics at all**, and has **no PrometheusRule of its own**.

Measured on `origin/main` (2026-08-19):

```
grep -rl "micrometer|MeterRegistry|Counter|Gauge|Timer" openbank-settlement-service/src/main
  -> (nothing; 20 .kt files)

grep -rn "openbank_settlement" openbank-infra/gitops/components/payments/
  -> only DB names (openbank_settlement as a Postgres database/owner)
```

The one alert whose name contains "Settlement" — `ClearingSettlementWindowMissed` — belongs to **clearing-service** (`expr: increase(openbank_clearing_settlements_completed_total[1h]) == 0`, `service: clearing`). Nothing anywhere queries a settlement-service series, because there are none to query.

`openbank-campaign-service` is in the same state: scraped, zero domain metrics.

## Why this is not visible today

The pod is scraped (namespace `payments` is in the fleet PodMonitor `matchNames`, verified by `check-podmonitor-namespace-coverage.py`), the workload is Ready, and ArgoCD is Healthy. The service is simply silent. This is the repo's recurring shape: *a healthy workload, a valid config, and a series that does not exist* — the same class as the 21 Rollouts with zero series and the eight unscraped agent namespaces, except one layer earlier. Scrape coverage was fixed; **emission** was not, and no gate asks for it.

Settlement is money-path. There is currently no signal that would distinguish "settlement is running normally" from "settlement has processed nothing for six hours".

## Suggested work

1. A `SettlementMetricsAdapter` following the fleet convention (`DomainMetrics`, `P50/P95/P99` constants in a companion object to avoid the `MagicNumber` detekt trip, field injection to stay under `constructorThreshold: 9`).
2. Counters/timers for the settlement cycle: cycles started/completed, per-cycle duration, items settled, failures.
3. A `PrometheusRule` in `components/payments` alerting on the **success state** ("no settlement completed in N hours during business hours"), not on an error rate — an adapter that quietly does nothing produces no errors to alert on.
4. Same for campaign-service.

## Where this came from

[...]

PR #5723, opened 2026-08-19
feat(settlement): emit domain metrics and alert on them (#5705)

JiRaska:
## What

`openbank-settlement-service` — a declared money-path service — emitted **no application metric of any kind** and had **no PrometheusRule of its own**. The pod was scraped, the Rollout was Healthy, ArgoCD was Green, and there was simply no series to query. Nothing could distinguish *"settlement is running normally"* from *"settlement has booked nothing for six hours"*.

This adds the metrics and the alerts that read them.

### Metrics (commit 1)

`SettlementMetricsPort` (application) + `SettlementMetricsAdapter` (infrastructure), following the fleet convention — `P50/P95/P99` constants in the companion object for detekt `MagicNumber`, field injection of `Instance<MeterRegistry>` so a missing registry leaves the adapter inert rather than breaking the money path.

| Series (all tagged `service="settlement"`) | What it makes visible |
| --- | --- |
| `openbank_settlement_originated_total{currency,outcome}` | acceptance rate, `created` vs idempotent `replayed` |
| `openbank_settlement_saga_steps_total{step,outcome}` | every activity attempt, forward legs and compensations |
| `openbank_settlement_terminal_total{outcome}` | `booked` / `rejected` — the series the alerts read |
| `openbank_settlement_cycle_duration_seconds{outcome}` | origination → terminal latency |
| `openbank_settlement_booked_amount{currency}` | amount distribution of booked settlements |

Two naming decisions worth stating explicitly:

- **Origination counts as *accepted*, never as settled.** At that point `originate()` has only persisted a PENDING row and asked Temporal to start; no money has moved.
- **A compensated saga is counted as `rejected` on its own terminal leg**, never sharing a flag with success — the shape that previously shipped a push adapter reporting deliveries that never left the process.

### Alerts (commit 2)

`openbank-settlement-alerts` in `components/payments/prometheus-rules.yaml`, five rules. The two interesting ones are about the **boot-time** state rather than the steady state:

- **`SettlementSagaNotCompleting`** (critical) — the primary alert. Compares work *accepted* against work *finished* (`originations > 0 and terminals == 0`), which makes it self-normalising: a freshly started pod has no originations either, so the first conjunct is `0` and the rule stays silent instead of firing 15m after every deploy. It catches the failure the saga makes invisible — `originate()` answers 202 and hands off to Temporal, so a worker that never picks work up leaves every settlement PENDING forever with no HTTP error, no latency change and no log line.
- **`SettlementNoBookingsInBusinessHours`** (warning) — the issue's literal ask, with an explicit `> 0` guard. Because the counter is registered eagerly the series *exists and reads 0* on a fresh pod, so without the guard this is true for every pod for the first six hours of its life. Same guard, same reason, as `AuditChainVerificationStale`. What it gives up (a pod that has never booked since restart) is covered by `SettlementSagaNotCompleting` within the hour and at `critical`.

Plus `SettlementSagaStepsFailing` (Temporal retries before compensating, so a sustained failure rate is the only warning before customer money moves and moves back), and `SettlementCompensationsElevated` (critical — a compensated saga ends REJECTED, a healthy-looking terminal state no error rate reports).

## Why the counters are registered eagerly

Micrometer does not create a counter until its first increment, so `increase(openbank_settlement_terminal_total[6h]) == 0` written against a lazily created counter matches **nothing at all** on a service that has never booked anything — precisely the state the alert exists to catch. Every series an alert reads is bound in `bindTo()` at `@PostConstruct`, and the bean is `@Startup` because `@ApplicationScoped` is lazy.

## Tests — proven by mutation, not by inspection

11 new tests, all using a **real adapter over a real registry** rather than a verified mock, each with its negative control (the counter that must *not* have moved is asserted alongside the one that must).

Both halves were falsified before being trusted:

| Mutation | Result |
| --- | --- |
| gut the eager-registration body of `bindTo()` | **4 tests red** — the scrape renders empty and `openbank_settlement_terminal_total` is absent, i.e. exactly the alert's blind spot |
| drop the `metrics.settlementBooked(...)` call from `bookToLedger` | **1 test red** — `expected: 1.0 but was: 0.0` |

`SettlementMetricsAdapterTest` additionally asserts the **rendered Prometheus scrape text**, so the dot→underscore mapping, the `_total` suffix and the tag ordering come from Micrometer rather than from this repo's idea of them — a rule referencing a name that renders differently cannot pass as correct while matching nothing.

## Verification

```
:openbank-settlement-service:compileKotlin :test :ktlintCheck :detekt :koverVerify  -> BUILD SUCCESSFUL
53 tests, 0 failures, 0 skipped
```

Gates run locally, all green: `check-critical-alert-egress.py`, `check-gitops-duplicate-resources.py`, `check-gitops-secret-refs.py`, `check-podmonitor-namespace-coverage.py`. The alert YAML round-trips through `yaml.safe_load_all` (6 documents).

Two detekt/ktlint traps hit and fixed along the way, both of the "fires AT the threshold" family: `TooManyFunctions` tripped at exactly 11 once the `step()` helper landed (`cycleDuration()` moved to a top-level function, declared **after** the class so it cannot steal the class annotation), and the new test dep needed the `io.micrometer.prometheusmetrics` artifact because Quarkus ships the `-simpleclient` variant — same explicit `testImplementation` and same CVE pin as `openbank-domestic-payment` already carries.

## Deliberately left out

**campaign-service.** The issue's suggested-work item 4 asks for the same treatment there, and this PR does not do it — hence `Refs`, not `Closes`. campaign-service is in the identical state (scraped, zero domain metrics) and deserves its own PR rather than being bolted onto a money-path change.

## Merge note

PR #4308 also appends a document to `components/payments/prometheus-rules.yaml`. The added alerts are disjoint (`openbank-card-issuance-alerts` vs `openbank-settlement-alerts`, no shared alert names), but both append at EOF, so whichever lands second needs a textual resolution — and per this repo's history a "clean" merge of two neighbouring additions can silently keep only one. Worth counting the documents after merging (`yaml.safe_load_all` should yield 7).

Refs #5705

---

**Note (rebase onto current `main`):** `SettlementServiceDown` was **dropped** during the conflict resolution. The progress half of #5705 merged to `main` first, and its document rejects that rule explicitly — `PaymentServiceDown` in `prometheus-rules-tier1.yaml` already owns `up{job="observability/openbank-services", namespace="payments"}`, so a second rule over the same target would page twice for one outage. The other four rules in this PR are unchanged, and this PR's document (`openbank-settlement-alerts`) is appended alongside `main`'s (`openbank-settlement-progress-alerts`) with disjoint rule, metric and object names — which is how `main`'s own comment says the two halves were meant to land.

### DIFF
--- a/openbank-infra/gitops/components/payments/prometheus-rules.yaml
+++ b/openbank-infra/gitops/components/payments/prometheus-rules.yaml
@@ -406,3 +406,141 @@ spec:
               balance compensation was attempted and refused, so funds moved and were not returned —
               this does not resolve itself. Identify the settlement, confirm which leg is
               outstanding, and start the collections/dispute path.
+---
+# ─────────────────────────────────────────────────────────────────────────────
+# settlement-service (issue #5705)
+#
+# settlement-service is a declared money-path service that, until #5705, emitted no application
+# metric of any kind: the pod was scraped, the Rollout was Healthy, ArgoCD was Green, and there was
+# simply no series to query. Nothing here could distinguish "settlement is running normally" from
+# "settlement has booked nothing for six hours".
+#
+# Every series below is registered EAGERLY by SettlementMetricsAdapter.bindTo() at @PostConstruct
+# (the bean is @Startup because @ApplicationScoped is lazy), so it exists at 0.0 from the first
+# scrape. That is load-bearing, not tidiness: Micrometer does not create a counter until its first
+# increment, so `increase(openbank_settlement_terminal_total[6h]) == 0` written against a lazily
+# created counter matches NOTHING on a service that has never booked anything — precisely the state
+# these alerts exist to catch. SettlementMetricsAdapterTest asserts the rendered scrape text, so the
+# exact strings below (dot->underscore, the `_total` suffix, tag order) come from Micrometer rather
+# than from this file's idea of them.
+# ─────────────────────────────────────────────────────────────────────────────
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: openbank-settlement-alerts
+  namespace: payments
+  labels:
+    app.kubernetes.io/part-of: openbank
+    app.kubernetes.io/name: settlement-service
+spec:
+  groups:
+    - name: openbank.settlement.health
+      rules:
+        - alert: SettlementSagaNotCompleting
+          # THE alert for this service. Self-normalising by construction: it compares work accepted
+          # against work finished, so it needs no business-hours guard and no uptime guard — a
+          # freshly started pod has no originations either, making the first conjunct 0 and the
+          # rule silent, rather than firing 15m after every deploy the way an unguarded
+          # `increase(...) == 0` gauge alert does (the boot-time fourth state this repo has already
+          # paid for once, ADR-0237 / #4208).
+          #
+          # This is the shape of the failure the saga makes invisible: originate() answers 202 with
+          # a PENDING row and hands off to Temporal, so a worker that never picks work up (worker
+          # not registered, wrong task queue, Temporal frontend unreachable) leaves every settlement
+          # PENDING forever and produces no HTTP errors, no latency change and no log line anyone
+          # is watching. Money is accepted and nothing settles.
+          expr: |
+            increase(openbank_settlement_originated_total{outcome="created"}[1h]) > 0
+            and
+            increase(openbank_settlement_terminal_total[1h]) == 0
+          for: 15m
+          labels:
+            severity: critical
+            service: settlement
+          annotations:
+            # Subject: incremented on the settlement request/activity path
+            # (SettlementMetricsAdapter), not by any scheduler — no period to compare against.
+            subject_period: continuous
+            summary: "settlement-service accepted settlements but completed none"
+            description: >-
+              {{ $value }} settlement(s) were accepted in the last hour and not one reached a
+              terminal state (booked or rejected). Suspect the Temporal worker: check that the
+              settlement worker is registered on task queue `openbank-settlement` and that
+              temporal-frontend.temporal.svc:7233 is reachable. Rows will be sitting in PENDING.
+
+        - alert: SettlementNoBookingsInBusinessHours
+          # The issue's literal ask — "settlement has processed nothing for six hours".
+          #
+          # The `> 0` guard is what separates "was settling and went quiet" from "has not settled
+          # yet since it started", and it is required, not decorative. The counter is eagerly
+          # registered, so on a fresh pod the series EXISTS and reads 0; without the guard
+          # `increase(...) == 0` is true for every pod for the first six hours of its life and this
+          # warning fires after every deploy during business hours. Same guard, same reason, as
+          # AuditChainVerificationStale in components/observability/prometheus-rules-audit-chain.yaml.
+          #
+          # What the guard deliberately gives up: a pod that has NEVER booked anything since restart
+          # is not covered here. That case is SettlementSagaNotCompleting's, which catches it within
+          # the hour and at `critical` — so nothing is left uncovered by keeping this one narrow.
+          expr: |
+            increase(openbank_settlement_terminal_total{outcome="booked"}[6h]) == 0
+            and openbank_settlement_terminal_total{outcome="booked"} > 0
+            and on() hour() >= 6 and on() hour() <= 18
+          for: 30m
+          labels:
+            severity: warning
+            service: settlement
+          annotations:
+            # Subject: incremented on the settlement request/activity path
+            # (SettlementMetricsAdapter), not by any scheduler — no period to compare against.
+            subject_period: continuous
+            summary: "No settlements booked in 6h (business hours)"
+            description: >-
+              settlement-service has booked nothing to the ledger for 6h during business hours,
+              having booked normally earlier in this pod's life. Either upstream stopped
+              originating settlements, or the saga is failing short of the ledger — cross-check
+              openbank_settlement_saga_steps_total{outcome="failed"}.
+
+        - alert: SettlementSagaStepsFailing
+          # Temporal RETRIES a failing activity before it compensates, so a non-zero failure rate is
+          # not by itself an incident — a sustained one is the only warning before customer money
+          # gets moved and moved back. Warning, not critical: the compensation is the safety net
+          # working, and SettlementCompensationsElevated below is what pages if it is actually used.
+          expr: |
+            sum by (step) (rate(openbank_settlement_saga_steps_total{outcome="failed"}[15m])) > 0
+          for: 20m
+          labels:
+            severity: warning
+            service: settlement
+          annotations:
+            # Subject: incremented on the settlement request/activity path
+            # (SettlementMetricsAdapter), not by any scheduler — no period to compare against.
+            subject_period: continuous
+            summary: "settlement saga step {{ $labels.step }} is failing repeatedly"
+            description: >-
+              The `{{ $labels.step }}` activity has been failing continuously for 20m. Temporal is
+              still retrying; once the attempts are exhausted the saga compensates and the
+              settlement ends REJECTED. Check the downstream this step calls (balance-service for
+              debit/credit, ledger-service for ledger_book).
+
+        - alert: SettlementCompensationsElevated
+          # A compensated saga MOVES CUSTOMER MONEY AND MOVES IT BACK, then ends in REJECTED — a
+          # perfectly healthy-looking terminal state that no error rate and no HTTP status reports.
+          # This is the one alert that reads that outcome as the incident it is.
+          expr: |
+            sum(rate(openbank_settlement_terminal_total{outcome="rejected"}[1h]))
+              /
+            clamp_min(sum(rate(openbank_settlement_terminal_total[1h])), 1e-9)
+              > 0.1
+          for: 30m
+          labels:
+            severity: critical
+            service: settlement
+          annotations:
+            # Subject: incremented on the settlement request/activity path
+            # (SettlementMetricsAdapter), not by any scheduler — no period to compare against.
+            subject_period: continuous
+            summary: "Over 10% of settlements are being rejected after compensation"
+            description: >-
+              {{ $value | humanizePercentage }} of settlements reaching a terminal state in the last
+              hour were REJECTED after their compensations ran — each one debited and credited real
+              accounts and then reversed them. Treat as a money-path incident, not a error-rate blip.
--- a/openbank-settlement-service/build.gradle.kts
+++ b/openbank-settlement-service/build.gradle.kts
@@ -41,16 +41,17 @@ dependencies {
     testImplementation(libs.quarkus.test.security)
     testImplementation(libs.assertj)
     testImplementation(libs.mockk)
-    // Test-only (#5705): SettlementStrandedGaugeTest and SettlementActivitiesImplTest assert the
-    // alert expressions in gitops/components/payments/prometheus-rules.yaml against a real
-    // Prometheus scrape, so the dot->underscore mapping, the counter's `_total` suffix and the tag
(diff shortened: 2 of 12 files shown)


JiRaska:

Merged current `main` (`146ab2524`) into this branch. The collision was substantive, not textual: **#6037** replaced the three compensation stubs with real reversal ports and made `reverseBookToLedger` fail loudly (`LEDGER_REVERSAL_UNSUPPORTED`) rather than claim an unwind it never performed, while this branch wrapped every activity in the metrics-recording `step(...)`. Both sides are kept.

**What was decided, and why**

- `SettlementActivitiesImpl` takes **main's** bodies (`compensate`, the reverse ports) *inside* **this branch's** `step(SettlementStep.X)` wrapper. `step` records the outcome and rethrows, so it is behaviour-transparent — Temporal still owns every retry/compensation decision.
- `SettlementActivitiesMetricsTest` previously asserted `reverse_ledger_book` **completed**. Under #6037 that activity throws, so the assertion is now `failed == 1.0` with an explicit `completed == 0.0` negative control. Counting it as completed would have restored exactly the false-success #6037 was written to remove.
- `@Suppress("TooManyFunctions")` on the class: main's `compensate` is the 11th function and detekt fires **at** the threshold, not above it. Same rationale, and same wording, as `CampaignJourneyActivitiesImpl` — the fleet's other Temporal activities implementation.

**Entry counts on the structured file that auto-merged.** `openbank-infra/gitops/components/payments/prometheus-rules.yaml` is a list, so a clean merge can still drop entries silently. Counted:

| | alerts |
|---|---|
| merge-base | 15 |
| this branch | 20 (+5 settlement) |
| `main` | 13 (−2: #5787 deleted 3 alerts that could never fire) |
| **merged result** | **18** = 15 + 5 − 2 ✅ |

All five settlement alerts survive (`SettlementServiceDown`, `SettlementSagaNotCompleting`, `SettlementSagaStepsFailing`, `SettlementCompensationsElevated`, `SettlementNoBookingsInBusinessHours`), and no alert present on `main` is missing from the result.

**Build** (`--rerun-tasks`, serialized, exit code read directly):

```
:openbank-settlement-service:ktlintCheck   PASS
:openbank-settlement-service:detekt        PASS
:openbank-settlement-service:test          PASS
:openbank-settlement-service:quarkusBuild  PASS   (CDI resolves the new SettlementMetricsPort param)
tests=60  failures=0  errors=0  SKIPPED=0
```

SKIPPED is reported because a module that fails to boot reports its integration tests as skipped, which reads as a pass. It is zero here.

JiRaska:
### CI triage: three reds, two genuine and one phantom

**Genuine — these are this diff:**

1. `gates (gitops)` — *alert window accommodates its subject's period (#5736, enforced)*.
   The four new settlement rules each carry a range window and no `subject_period`
   annotation:
   - `SettlementSagaNotCompleting` `[1h]`
   - `SettlementNoBookingsInBusinessHours` `[6h]`
   - `SettlementSagaStepsFailing` `[15m]`
   - `SettlementCompensationsElevated` `[1h]`

   Each needs `subject_period: continuous` (driven by real traffic) or the subject's actual
   period. The gate's self-test passed all 7 cases, so this is the gate working, not flaking.

2. `gates (supplychain)` — *stale comment references*.
   `openbank-settlement-service/src/test/kotlin/com/openbank/settlement/infrastructure/observability/SettlementMetricsAdapterTest.kt:18`
   names `openbank-infra/gitops/components/payments/settlement-service-alerts.yaml`, which is
   not in the tree — the rules landed in `payments/prometheus-rules.yaml` instead. Fix the
   prose to the real path.

**Phantom — not this diff:**

3. `gates (api)` — the `SettlementStatus … ['LEDGER_REVERSED']` enum-drift message is the gate
   bug fixed on `main` by #6220 (`b85ac16`, 14:57Z). This PR's `refs/pull/5723/merge` does not
   contain that commit, so the shard ran the pre-fix script. A rebase clears it; `gh run rerun`
   will not, because it re-runs the recorded stale merge SHA.

`Validate manifests` is the roll-up of the shards above, so it clears when they do.

JiRaska:
## Rebuilt on current `main` — both conflicts resolved, one rule dropped on the merits

Was CONFLICTING on `openbank-infra/gitops/components/payments/prometheus-rules.yaml` and `openbank-settlement-service/build.gradle.kts`. Rebuilt on `origin/main` (`8b1868746`); new head `7936224ff`.

### `build.gradle.kts` — comment-only

`main` already carries `testImplementation("io.micrometer:micrometer-registry-prometheus:1.17.0")`; only the surrounding comment differed. Merged the prose so it names all three tests that depend on it (`SettlementStrandedGaugeTest`, `SettlementActivitiesImplTest`, `SettlementMetricsAdapterTest`) instead of either side's subset. No dependency change.

### `prometheus-rules.yaml` — both documents kept, as `main` intended

The progress half of #5705 merged to `main` while this PR was open, as `openbank-settlement-progress-alerts`. That is not a duplicate of this PR — `main`'s own comment says so explicitly:

> Scope: this document is the PROGRESS half of #5705 only [...] The OUTCOME half ("what did settlements do") is PR #5723's saga-step and terminal-state counters, which this deliberately does not duplicate; the two are complementary and the rule names, the metric names and the PrometheusRule object name are disjoint so they can land in either order.

So the conflict was purely textual — both sides inserted a document at the same offset. Resolved by keeping `main`'s file byte-for-byte and appending this PR's `openbank-settlement-alerts` as a separate document. Verified mechanically rather than by eye:

```
main lines=369  new lines=495
PREFIX IDENTICAL -> main's content untouched, change is purely appended
```

Parsed result, 7 documents, disjoint names:

```
openbank-settlement-progress-alerts  [openbank.settlement.progress: SettlementStrandedMidSaga, SettlementStuckAfterCompensation]
openbank-settlement-alerts           [openbank.settlement.health: SettlementSagaNotCompleting, SettlementNoBookingsInBusinessHours, SettlementSagaStepsFailing, SettlementCompensationsElevated]
```

**One rule was dropped: `SettlementServiceDown`.** `main`'s newer document rejects exactly that rule, with a reason this PR predates:

> There is deliberately no SettlementServiceDown rule here: PaymentServiceDown already owns `up{job="observability/openbank-services", namespace="payments"}`, and a second rule over the same target would page twice for one outage.

Adding it back would have double-paged on every settlement-service outage. The other four rules are kept unchanged.

### Verification

`ktlintCheck` + `detekt` clean:

```
$ ./gradlew :openbank-settlement-service:ktlintCheck :openbank-settlement-service:detekt
LINT EXIT=0
```

(Run separately on purpose — Gradle stops at the first failing task, so a green-looking task list after a `test` failure is not a list that ran.)

Tests: **67 tests, 0 skipped, 1 failure** — and that failure is pre-existing on `main`, not introduced here:

```
FAIL SettlementStrandedGaugeTest  tests 5  fail 1  skip 0
TOTAL 67  FAILURES 1
```

Confirmed against a clean detached worktree at `origin/main` with none of this PR's changes applied:

```
$ ./gradlew :openbank-settlement-service:test --tests '*SettlementStrandedGaugeTest*'
CLEAN-MAIN EXIT=1
5 tests completed, 1 failed
```

Cause: `SettlementStrandedGauge.NON_TERMINAL` lists 6 statuses while `SettlementStatus` has since gained `REVERSAL_FAILED` and `LEDGER_REVERSAL_UNSUPPORTED` (#6037). The test derives its expectation from the enum, so it correctly reports 8 expected vs 6 published. That is a real gap, not just a red test — a settlement parked in `REVERSAL_FAILED` (money moved and **not** returned) publishes no age gauge, so `SettlementStrandedMidSaga` cannot fire for it. Tracked separately; deliberately not fixed here, since it is `main`'s defect and not in this PR's scope.

Every other file applied cleanly. Not merged — left for review.

JiRaska:
## Rebuilt on current `main` — and a merge-order note with #4308

`prometheus-rules.yaml` was rebuilt as **main's 408 lines byte-for-byte, plus this PR's 126-line
document appended**. Verified rather than asserted: `head -408 <file> | cmp -s - <main's version>`
matches, so nothing #6284 added was disturbed.

Checked after the append:
- 7 documents, names all distinct (`openbank-settlement-alerts` is new; `openbank-settlement-progress-alerts` is main's).
- 20 alert rules, every name unique.
- `openbank_settlement_` prefix unchanged.

**The earlier decision to drop `SettlementServiceDown` still holds, and `main` now says so itself** —
the comment on `openbank-settlement-progress-alerts` states that `PaymentServiceDown` in
`prometheus-rules-tier1.yaml` already owns `up{namespace="payments"}`, so a second rule would double
up. No such alert exists in either document.

One thing I added beyond a pure rebuild: `check-alert-window-vs-subject-period.py` warned that all
four new alerts have a range window and no `subject_period` annotation. All four are driven by
counters incremented on the request/activity path, so they now declare `subject_period: continuous`.
Gate output afterwards: `no findings.` (was `4 finding(s)`).

### For a human — merge order against #4308

**#4308 also appends a document to the end of this same file.** Both are green in isolation and both
append at EOF, so whichever merges second will conflict textually. The resolution is trivial (keep
both documents) but it does need doing by hand.

JiRaska:
`DIRTY` since #6484 (`2da87cf6f`), which removed `reverseBookToLedger` as an unreachable activity.
This branch instruments that method, so the rebase is not purely mechanical — measured:

```
override fun count, main:              6
override fun count, this branch:       7   (SettlementActivitiesImpl.kt:132 wraps reverseBookToLedger)
```

Three things to fix together rather than one at a time:

1. **Drop the `step(SettlementStep.REVERSE_LEDGER_BOOK)` wrapper** along with the method, and check
   whether `SettlementStep.REVERSE_LEDGER_BOOK` still has any producer. An enum value nothing emits
   becomes a metric label that can never appear — a gauge or counter series permanently at zero
   reads as "healthy" rather than "absent", which is the failure mode this PR exists to remove.
2. **The `@Suppress("TooManyFunctions")` justification is now off by one.** Its comment reads "the
   seven activities `SettlementActivities` declares … plus `compensate`, `audit`, `step` and
   `runOnVertxContext`", i.e. exactly 11 at the threshold. With six activities the total is 10 and
   the suppression may no longer be needed at all. Worth re-checking rather than carrying the
   annotation forward with a comment that no longer counts correctly — detekt fires **at** the
   threshold here, not above it, so the arithmetic decides whether it is required.
3. **Check the alert rules** in `openbank-infra/gitops/components/payments/prometheus-rules.yaml`
   for anything keyed to the reversal step, for the same reason as (1).

I have not touched the branch — the counts above are the whole finding, and 875 lines of metrics
code with a money-path scope is not something to rebase blind.

JiRaska:
**Correcting my comment above — the premise it rested on is being reverted.**

I told you to drop the `reverseBookToLedger` work on the grounds that #6484 had removed the activity
as unreachable. #6516 (OPEN, CLEAN) reverts #6484, and its reasoning is better than the one I
relayed:

> `bookToLedger` is **not atomic**. `ledgerPort.book()` posts the journal, then `updateStatus(BOOKED)`
> and `audit(...)` can each fail — after the general ledger has been written. The saga unwinds both
> balance movements and rejects, and the journal stands with nothing in the row, the alerts or the
> audit trail naming it.

So the compensation was dead because of **the order it was registered in** — after its forward
activity returned — not because the saga had no room for it. #6484 read "unreachable" off the
registration order and attributed it to the saga's shape; #6481 fixes the ordering properly.

What that changes for this PR: **do not treat the ledger-reversal path as gone.** Wait for #6516 to
land before rebasing — it restores exactly the state this branch was written against, which turns
the rebase back into a clean one instead of the reconciliation I described.

What still holds from my comment: read the state off live `main` at the time you rebase, not off
this thread. Three variants of this code have now been in flight at once (#6484 merged, #6516
reverting it, #6481 fixing it properly), so anything I write about the ledger leg has a short shelf
life — including this correction.

My original comment was accurate about what `main` held when I wrote it and wrong about what that
meant. I passed on #6484's conclusion without checking whether `bookToLedger` was atomic, which is
the one thing the conclusion depended on and is visible in three lines of the method.

PR #5725, opened 2026-08-19
fix: stop swallowing transient event-consumer failures as an ack across 4 services (#5698)

JiRaska:
Fleet sweep for the catch-and-ack defect of #5698, following the shape #5699 established in
`openbank-kyc-service`: bounded retry with backoff and then a **rethrow** so the connector can
dead-letter, while genuine poison pills (malformed payloads, unretryable per-event failures) keep
acking because they can never succeed on a retry. That split is the whole change.

The sibling PR for the money-path services (`openbank-interest-service`, `openbank-fraud-service`)
is on `fix/event-consumer-catch-and-ack-sweep`; this PR deliberately does not touch them.

### `openbank-copilot-service`
1. `infrastructure/kafka/PartyErasureConsumer.kt` — `conversationStore.deleteForParty` is a reactive
   Panache call and shared a generic `catch (e: Exception)` with everything else. A connection
   refused lost a **GDPR Art. 17 erasure permanently**: the transcripts stayed on disk, neither
   erasure metric was incremented, and the log claimed the erasure was handled. Now bounded-retried
   and rethrown; parse and partyId validation still ack.

### `openbank-customer-edge`
2. `infrastructure/onboarding/OnboardingResumeService.kt` — worse than the original bug, as noted in
   the issue: `try { resume(...) } catch { Log.error } finally { pendingStore.delete(caseId) }`.
   `resume()` fans out to party-service, pid-service and Keycloak, so a transient failure was acked
   **and** the pending record was deleted anyway, leaving even a manual replay with nothing to
   resume. Both halves fixed: the fan-out is bounded-retried then rethrown, and the record is
   deleted **only on success**. The retry is safe — the party create carries the idempotency key
   `onboarding-resume-<caseId>`. Backoff uses `Thread.sleep` because the handler is `@Blocking`
   (worker thread, never the event loop).

### `openbank-document-service`
3. `infrastructure/kafka/AccountCreatedConsumer.kt`
4. `infrastructure/kafka/AnnualFeeSummaryReadyConsumer.kt`

   Both framed the swallow as a deliberate ADR-0086 trade-off (off the money path; one bad account
   must not wedge issuance for every subsequent one) and both named "a bounded retry / DLQ is a
   possible follow-up". The reasoning is sound for a *deterministic* failure and could not be
   applied selectively, because the same `catch (e: Exception)` also swallowed a connection refused.

   The catch is now narrowed by type, and the boundary is not arbitrary: the deterministic per-event
   failures in this service surface as `IllegalStateException` (`DocumentRenderService`'s
   `error("No published template for …")`) or `IllegalArgumentException`. Those are rethrown without
   a retry and the consumer acks them — documented behaviour preserved. Everything else is retried
   and rethrown. Both comments were rewritten to describe what the code now does.

### `openbank-engagement-service`
5. `infrastructure/kafka/CampaignBannerPlacementConsumer.kt` — parse + `placements.save` shared one `try`
6. `infrastructure/kafka/DisputeOpenedEventConsumer.kt` — parse + `adverseState.setActive/clearActive`
7. `infrastructure/kafka/FraudHoldEventConsumer.kt` — the fraud-hold exclusion signal
8. `infrastructure/kafka/LendingArrearsEventConsumer.kt`
9. `infrastructure/kafka/PartyErasureConsumer.kt`

   Near-identical, so they share one small helper (`infrastructure/kafka/ConsumerRetry.kt`) rather
   than five copies of the loop; each consumer keeps its own `parse(...)` because the fields differ.
   Parsing acks, the write retries then rethrows. Site 9 is the worst of the five: ADR-0220 D3.5
   treats `ERASURE_REQUESTED` as terminal and never re-derives it, so an acked-but-failed write left
   the party permanently targetable with no compensating path anywhere.

## Sites examined and NOT changed

- `openbank-interest-service`, `openbank-fraud-service` — out of scope by instruction; owned by the
  sibling money-path PR on `fix/event-consumer-catch-and-ack-sweep`.
- `OnboardingResumeService.resumeCreate` non-2xx from party-service — this does **not** throw; it
  emits a `CUSTOMER_ONBOARDING_RESUMED_CREATED` audit event with `result=ERROR` and returns. It is a
  silent-ish path, but unlike the sites above it leaves a durable audit record rather than only an
  ERROR log line, and converting it to a throw would re-emit that audit event on each retry.
  Deliberately left alone; flagging it here rather than changing it in this PR.
- `openbank-notification-service/.../PartyErasureConsumer.kt` — the "closest sibling" the copilot
  consumer's KDoc cites. Not in the assigned scope, and not touched here; worth a look in a
  follow-up sweep since the shape is the same family.

## Tests

Every site gets a **pair**, because a test that cannot separate a transient downstream failure from
a malformed payload proves nothing about either:

- transient/persistent downstream failure ⇒ **rethrown** after exactly `MAX_ATTEMPTS` (3) calls, so
  the connector can dead-letter;
- a transient failure that recovers ⇒ retried to success, not dead-lettered;
- malformed payload ⇒ still acked, downstream never called.

`FraudHoldEventConsumer` had **no test at all**, which is part of why the defect survived there; it
gets a first one. `ConsumerRetryTest` covers the engagement helper's semantics directly. Tests that
asserted the swallow as intended behaviour (copilot's
`a store failure is swallowed so one bad message cannot wedge the consumer group`, both
document-service `swallows a downstream failure …`) are replaced or renamed to match reality — same
class of green-test-as-defect-documentation the issue calls out.

## Verification (local, per module)

`test` + `ktlintCheck` + `detekt` green on all four modules (`test` implies `compileKotlin`):

| module | tests | failures | skipped |
|---|---|---|---|
| `:openbank-copilot-service` | 141 | 0 | 0 |
| `:openbank-customer-edge` | 380 | 0 | 1 (pre-existing broker-gated pact provider test) |
| `:openbank-document-service` | 141 | 0 | 0 |
| `:openbank-engagement-service` | 100 | 0 | 0 |

Refs #5698

### DIFF
--- a/openbank-copilot-service/src/main/kotlin/com/openbank/copilot/infrastructure/kafka/PartyErasureConsumer.kt
+++ b/openbank-copilot-service/src/main/kotlin/com/openbank/copilot/infrastructure/kafka/PartyErasureConsumer.kt
@@ -9,6 +9,7 @@ import com.openbank.copilot.application.port.out.ConversationStore
 import com.openbank.copilot.infrastructure.observability.CopilotMetricsAdapter
 import jakarta.enterprise.context.ApplicationScoped
 import jakarta.inject.Inject
+import kotlinx.coroutines.delay
 import org.eclipse.microprofile.reactive.messaging.Incoming
 import org.jboss.logging.Logger
 import java.util.UUID
@@ -24,8 +25,24 @@ import java.util.UUID
  *
  * Shape is copied from the seven services that already consume this event (notification-service's
  * `PartyErasureConsumer` is the closest sibling): `suspend @Incoming` so Quarkus dispatches on a
- * Vert.x duplicated context and the reactive Panache delete below runs correctly; poison-pill safe,
- * so any parse or delete failure is logged and the message acked rather than wedging the group.
+ * Vert.x duplicated context and the reactive Panache delete below runs correctly.
+ *
+ * ## Failure handling — a malformed event and a broken database are not the same thing (#5698)
+ *
+ * This consumer used to log-and-ack BOTH, which reads as poison-pill safety and is not. A malformed
+ * payload is unretryable: replaying it produces the same parse failure forever, so acking it is the
+ * only sane outcome and it stays that way below. [ConversationStore.deleteForParty] is a reactive
+ * Panache call, and a connection-refused there is the opposite case — the event is fine, the
+ * database is not, and the work must still happen once it recovers. Acking that lost a GDPR Art. 17
+ * erasure permanently: the only trace was an ERROR line nobody alerts on, and the transcripts the
+ * consumer exists to delete stayed on disk while the log claimed the erasure was handled. That is
+ * the exact shape that left ten of 73 sandbox parties without a KYC case for months (#5698).
+ *
+ * So the delete now runs under [withBoundedRetry] and, if it still fails, is RETHROWN — the
+ * connector retries and ultimately dead-letters, which is a signal someone can see. The retry is
+ * bounded and the failure moves to the DLQ, so a single bad event still cannot wedge the group.
+ * [ConversationStore.deleteForParty] is idempotent (a second delete matches nothing), so both the
+ * retry and a connector redelivery are safe.
  *
  * ## Identity — why the delete is not keyed on the storage key (#3881)
  *
@@ -73,7 +90,7 @@ class PartyErasureConsumer {
             return
         }
 
-        try {
+        withBoundedRetry(partyId) {
             val erased = conversationStore.deleteForParty(partyId.toString())
             if (erased == 0L) {
                 // Not necessarily a defect — a party that never chatted holds nothing. But it is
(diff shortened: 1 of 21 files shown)


JiRaska:
Overlap note — same issue (#5698), and one file is in both.

#5719 converts `openbank-copilot-service/.../PartyErasureConsumer.kt` too, onto a shared `EventRetry.withRetry` in `openbank-libs-runtime` (`com.openbank.libs.messaging`), together with 14 other sites and an enforced `event-handler-swallows` gate.

The rest of this PR (customer-edge, document-service, and the others) does not overlap and is coverage #5719 does not have — the two are complementary. Only the copilot file needs a decision: whoever lands second drops their version of it rather than resolving the conflict by picking a side, since taking one whole side would revert the other's semantics.

If you would rather not depend on the shared helper, say so on #5719 and I will adapt — but a second private retry helper per service is the shape that let this defect spread across nine services unnoticed.

JiRaska:

- fraud `transaction-signal`

example:

   meant to park."*

Issue #1000, opened 2026-07-13
sdd-service: authorised direct-debit collection never posts a debit anywhere (ADR-0036, no consumer exists)


JiRaska:
## What

Found via ADR-0160's event-consumer-liveness gate (issue #996 triage): `openbank.sdd.event`, published by `openbank-sdd-service`, has zero consumers anywhere in the fleet — and unlike most other findings in #996, **no candidate consuming service exists at all** (checked account/ledger/transaction/clearing-service).

ADR-0036 (SEPA SDD mandate lifecycle) states directly that accepting a collection "emits `sdd.collection.authorised` for the ledger/payment path to execute." That downstream half was never built.

## Why it matters

This is the **most severe finding in the #996 triage**: an authorised SEPA direct-debit collection is recorded in sdd-service as accepted, but the actual debit against the payer's account never fires anywhere. Money that should move never does — silently, with no error. This is the exact #889 failure class (published event, zero consumers, feature looks complete) but on the debit-collection path rather than a recurring-payment scheduler.

## Scope (needs design, not a quick PR)

- Decide which service owns executing the actual debit posting for an authorised SDD collection — most likely `transaction-service` or `ledger-service` directly (mirroring how #889 wired standing-order → sepa-payment), or `clearing-service` if SDD collections route through the clearing simulator like other rails.
- Needs: a real consumer service, idempotent debit execution keyed on the collection/mandate id, failure handling (insufficient funds, mandate revoked mid-flight, R-transaction / return handling per SDD scheme rules), and audit trail.
- This is money-path — needs `rules.yaml: money_path_services` classification (check if sdd-service is already listed) and the 2-approval + threat-model discipline before merge.

## Linked issues

Refs #996 (the triage this was found in), Refs ADR-0036

## Notes

Not implementing this in the same PR/session that found it — this is a real missing money-movement feature requiring a deliberate design decision (which service executes the debit, how returns/reversals are handled), not a mechanical fix. Given the severity (money never moves for an authorised collection), this is a strong candidate for near-term prioritization once triaged by a human.

JiRaska:
## Recommended design direction (not yet implemented — needs confirmation before starting)

Recommend **reusing the exact orchestration pattern already proven in #889/PR #994**, not inventing a new one:

[...]

PR #5726, opened 2026-08-19
fix(fleet): close two event-handler-swallow gate blind spots, and the money-path sites behind them

JiRaska:
**Stacked on #5719** (base = `fix/event-handlers-stop-acking-failed-work`). Follows up my review
comment there.

## How this was found

#5719's gate is `enforced` over a **derived** 170-file subject set — the right shape. So I checked
what it *cannot* see rather than that it passes: run from its own branch it printed
`SUBJECTS=170 / clean`, while two consumers in `rules.yaml: money_path_services` held the exact
#5698 catch-and-ack.

## Blind spot 1 — verb not on the list

`SddCollectionDebitConsumer` does `transactionClient.initiateTransaction(request)` — a real debit
for **every authorised direct-debit collection**. The receiver matched the naming convention;
`initiate` simply was not in `STATE_CALL`'s verb alternation, so the finding was dropped.

## Blind spot 2 — state call one helper deep (the structural one)

`StandingOrderDueConsumer`'s try body calls the private `executeSepa(orderId, node)`; only *inside*
that does `sepaClient.createPayment(...)` appear — which `STATE_CALL` **would** have matched inline.
A text scan of the try body cannot see it, so a handler that delegates to a well-named domain
helper (the tidier way to write one) was invisible.

`_local_fun_bodies()` + `_reaches_state_change()` now follow same-file helpers to depth 2,
cycle-safe, expression-bodied `fun f() = x.save()` included. Depth 2 is deliberate rather than
unbounded — it covers the real shape without dragging half a file into scope.

> Both blind spots bite hardest exactly where the stakes are highest, because money-path code is
> the most likely to wrap its calls in domain-named helpers.

## What the fixed gate found: 5 sites — 3 more than my manual sweep

Including a `@Scheduled` in balance-service my own `@Incoming`-only sweep had missed, the same way
the original manual sweep missed `@Scheduled`. Fixed by what each site **is**, not one template:

| site | fix | why |
|---|---|---|
| sdd `initiateTransaction` | `EventRetry` + rethrow; non-2xx throws `SddDebitFailedException` | R-transaction/return generation is unbuilt (#1000), so the DLQ is the **only** path to a human |
| standing-order `executeSepa` | `EventRetry` around execution; parse failures acked separately | the due event fires once per schedule and nothing re-drives it |
| balance `ValueDateRollScheduler` | **`observed-by:` marker** | money is derived and already correct without the job; `recordSuccess()` runs only on the completed path, so a broken roll climbs the liveness gauge until `WorkflowLivenessStale` (ADR-0237) |
| sanctions per-item catch | **`observed-by:` marker** | a failed list does not advance `lastRefreshedAt`, stays due, retried next tick; aborting the batch would let one unreachable list starve every other |

The last two are **markers, not rethrows**, on purpose — the gate has those markers precisely for
legitimate silence, and forcing a rethrow there would have been wrong.

## Falsifiability

4 new self-test cases (**12/12 green**). Each blind spot must be *flagged*, plus two negatives:
a helper that changes nothing must not invent a finding, and a recursive helper must terminate
rather than hang the scan.

## Verification
- [x] `check-event-handler-swallows.py --self-test` — 12/12
- [x] `compileKotlin`, `ktlintMainSourceSetCheck`, `detekt` — green on all four services
- [x] Gate run on this branch: 5 → 1 finding

## Merge order
One finding remains — `interest`'s `remitUseCase.settle` — which **#5715 fixes on the same lines**.
It is an artifact of this base predating that PR, not a defect this PR introduces. **Land #5715
before this**, or the enforced gate reports it.

Refs #5698

### DIFF
--- a/.github/scripts/check-event-handler-swallows.py
+++ b/.github/scripts/check-event-handler-swallows.py
@@ -68,7 +68,13 @@ STATE_CALL = re.compile(
     r"\b\w*(?:repository|repo|usecase|port|store|dao|publisher|client|service)\s*\.\s*"
     r"(save|persist|insert|update|create|open|activate|grant|credit|debit|post|record|apply|"
     r"anonymize|anonymise|delete|upsert|merge|register|enrol|enroll|book|settle|publish|send|"
-    r"notify|emit|write|mark|complete|cancel|approve|reject|erase)\w*\s*\(",
+    r"notify|emit|write|mark|complete|cancel|approve|reject|erase|"
+    # Money-movement verbs the alternation still lacked. `initiate` is why
+    # `SddCollectionDebitConsumer`'s `transactionClient.initiateTransaction(...)` — a real
+    # direct-debit debit — was not flagged: the receiver matched (which is exactly what
+    # TRANSPORT_RECEIVER above keeps `client` in the list FOR), but the verb did not.
+    r"initiate|execute|transfer|charge|capture|reverse|refund|submit|dispatch|trigger|"
+    r"assign|revoke|close|freeze|release|disburse|repay|withdraw|deposit)\w*\s*\(",
     re.IGNORECASE,
 )
 # Two ways to be legitimately silent, and both must be STATED where a reviewer reads them.
@@ -96,12 +102,55 @@ HANDLED_IN_CATCH = re.compile(
 )
 
 
-def _state_change_in(body: str):
-    """First STATE_CALL in `body` that is not a transport-level read."""
+# A call to a helper declared in the SAME file. The state change often lives one level down:
+# `StandingOrderDueConsumer`'s try body calls `executeSepa(orderId, node)`, and only INSIDE that
+# does `sepaClient.createPayment(...)` appear — which STATE_CALL would have matched had it been
+# inline. Text-scanning the try body alone cannot see it, so a handler that delegates to a
+# well-named domain helper (the tidier way to write one) is invisible to this gate.
+LOCAL_CALL = re.compile(r"(?<![\w.])([a-z]\w*)\s*\(")
+
+
+def _local_fun_bodies(text: str) -> dict:
+    """Every `fun name(...)` declared in this file, by name — block- and expression-bodied."""
+    bodies = {}
+    for m in re.finditer(r"\bfun\s+(?:<[^>]*>\s*)?(\w+)\s*\(", text):
+        name = m.group(1)
+        depth, i = 0, m.end() - 1
+        while i < len(text):
+            depth += (text[i] == "(") - (text[i] == ")")
+            i += 1
+            if depth == 0:
+                break
+        rest = text[i : i + 400]
+        brace, eq = rest.find("{"), rest.find("=")
+        if brace != -1 and (eq == -1 or brace < eq) and rest[:brace].count("\n") <= 2:
+            bodies[name] = _block(text, i + brace)[0]
+        elif eq != -1 and rest[:eq].count("\n") <= 2:
+            nl = rest.find("\n", eq)
+            bodies[name] = rest[eq : nl if nl != -1 else len(rest)]
+    return bodies
+
+
+def _state_change_in(body: str, funs=None, depth: int = 2, _seen=None):
+    """First STATE_CALL in `body` that is not a transport-level read.
+
+    With `funs` (same-file `fun` bodies) it also follows calls to those helpers, `depth` levels
+    deep and cycle-safe. Depth 2 covers the real shape — handler -> executeSepa ->
+    sepaClient.createPayment — without letting one generic helper name drag half a file into scope.
+    """
     for m in STATE_CALL.finditer(body):
         if TRANSPORT_RECEIVER.search(body[: m.start() + len(m.group(0).split(".")[0]) + 1]):
             continue
         return m
+    if funs and depth > 0:
+        seen = _seen if _seen is not None else set()
+        for call in LOCAL_CALL.findall(body):
+            if call in seen or call not in funs:
+                continue
+            seen.add(call)
+            deeper = _state_change_in(funs[call], funs, depth - 1, seen)
+            if deeper:
+                return deeper
     return None
 
 
@@ -129,12 +178,12 @@ def _preceding_comment_block(text: str, try_idx: int) -> str:
     return "\n".join(block)
 
 
-def _run_catching_findings(text: str, path: str) -> list[tuple[str, int, str]]:
+def _run_catching_findings(text: str, path: str, funs=None) -> list[tuple[str, int, str]]:
     """runCatching { <state change> } whose result is recovered rather than rethrown."""
     out = []
     for m in RUN_CATCHING.finditer(text):
         body, after = _block(text, m.end() - 1)
-        state = _state_change_in(body)
+        state = _state_change_in(body, funs)
         if not state:
             continue
         # What happens to the Result decides it: .getOrThrow() or an else-branch that throws is
@@ -154,7 +203,8 @@ def _run_catching_findings(text: str, path: str) -> list[tuple[str, int, str]]:
 def findings_for(text: str, path: str) -> list[tuple[str, int, str]]:
     if not HANDLER.search(text):
         return []
-    out = _run_catching_findings(text, path)
+    funs = _local_fun_bodies(text)
+    out = _run_catching_findings(text, path, funs)
     for m in re.finditer(r"\btry\s*\{", text):
         try_body, after = _block(text, m.end() - 1)
         tail = text[after : after + 400]
@@ -164,7 +214,7 @@ def findings_for(text: str, path: str) -> list[tuple[str, int, str]]:
         catch_body, _ = _block(text, catch_open)
         if "throw" in catch_body or HANDLED_IN_CATCH.search(catch_body):
             continue
-        state = _state_change_in(try_body)
+        state = _state_change_in(try_body, funs)
         if not state:
             continue
         # The marker must justify THIS catch: it counts only inside the catch body, or in the
@@ -348,6 +398,79 @@ SELF_TEST_CASES = [
                     log.error("nope", e)
                 }
             }
+        }
+        """,
+        0,
+    ),
+    # --- shapes the naming-convention match could not see (money-path when found) ---
+    (
+        "a money verb outside the original list is still a state change (sdd: initiateTransaction)",
+        """
+        class C {
+            @Incoming("x")
+            suspend fun consume(p: String) {
+                try {
+                    transactionClient.initiateTransaction(request).awaitSuspending()
+                } catch (e: Exception) {
+                    log.error("swallowed", e)
+                }
+            }
+        }
+        """,
+        1,
+    ),
+    (
+        "a state change one private-helper deep is reached (standing-order: executeSepa)",
+        """
+        class C {
+            @Incoming("x")
+            suspend fun consume(p: String) {
+                try {
+                    executeSepa(orderId, node)
+                } catch (e: Exception) {
+                    log.error("swallowed", e)
+                }
+            }
+
+            private suspend fun executeSepa(orderId: UUID, node: JsonNode) {
+                sepaClient.createPayment(idempotencyKey, request).awaitSuspending()
+            }
+        }
+        """,
+        1,
+    ),
+    (
+        "helper indirection does not invent findings when the helper changes nothing",
+        """
+        class C {
+            @Incoming("x")
+            suspend fun consume(p: String) {
+                try {
+                    formatForLog(node)
+                } catch (e: Exception) {
+                    log.error("fine", e)
+                }
+            }
+
+            private fun formatForLog(node: JsonNode): String = node.toPrettyString()
+        }
+        """,
+        0,
+    ),
+    (
+        "a recursive helper terminates instead of hanging the scan",
+        """
+        class C {
+            @Incoming("x")
+            suspend fun consume(p: String) {
+                try {
+                    walk(node)
+                } catch (e: Exception) {
+                    log.error("fine", e)
+                }
+            }
+
+            private fun walk(n: JsonNode) { walk(n) }
         }
         """,
         0,
--- a/openbank-audit-service/src/main/kotlin/com/openbank/audit/application/AuditAnchorService.kt
+++ b/openbank-audit-service/src/main/kotlin/com/openbank/audit/application/AuditAnchorService.kt
@@ -56,6 +56,12 @@ class AuditAnchorService(
     )
     suspend fun captureScheduled() {
         if (!enabled) return
+        // observed-by: this job's own ADR-0237 liveness gauge. `recordSuccess()` is on the SUCCESS
+        // path only, so a permanently failing capture leaves the last-success age climbing until
+        // WorkflowLivenessStale fires — the failure is visible, just not through a DLQ. Swallowing
+        // is also the right call for the tick itself: this is an hourly @Scheduled sweep over the
+        // chain head, not a consumed message, so nothing is lost by one failed tick — the next hour
+        // re-reads the same head. Throwing would only kill the scheduler thread.
         runCatching { captureAnchor() }
             .onSuccess { liveness?.recordSuccess() }
             .onFailure { log.error("audit anchor capture failed", it) }
--- a/openbank-audit-service/src/main/kotlin/com/openbank/audit/application/AuditConsumer.kt
+++ b/openbank-audit-service/src/main/kotlin/com/openbank/audit/application/AuditConsumer.kt
@@ -73,9 +73,14 @@ class AuditConsumer {
         try {
             persist(payload, addressOf(message))
         } catch (e: Exception) {
-            // This legacy multi-producer channel deliberately retains its historic availability
-            // behaviour. D5 agent events use AgentAuditConsumer instead: it acknowledges only a
-            // successful durable write, so a store failure is retried by Kafka rather than lost.
+            // best-effort: DELIBERATE, and #6209 is where it was decided — this legacy
+            // multi-producer channel keeps its historic availability behaviour rather than wedging
+            // ~20 producers on one store failure. Stated plainly because the marker suppresses the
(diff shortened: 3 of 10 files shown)


JiRaska:

- fraud `transaction-signal`

example:

   meant to park."*

JiRaska:
Pushed a follow-up commit: CI caught a test I should have caught locally.

`SddCollectionDebitConsumerTest > a non-2xx-non-409 response is handled without throwing (logged as a failure)` was **passing against the behaviour this PR fixes** — it asserted the swallow as intended. That is precisely the trap #5698 documents in kyc-service: *"Two green tests asserted the swallow as intended behaviour… Reading the tests would have confirmed the bug as a feature."*

Now asserts both halves of the correct behaviour: retried `DEFAULT_MAX_ATTEMPTS` times (bounded — a permanently refusing rail must not pin the partition) and then throwing `SddDebitFailedException`. 6/6 green, read from the JUnit XML.

My mistake was running `compileKotlin`/`ktlint`/`detekt` on the touched modules but not their **tests** — a behaviour change needs the test suite, not just the compiler.

Also verified the `StandingOrderBootSmokeIT` failure I hit locally was the documented `Port already bound: 8081` environmental collision from my own parallel Gradle runs, not this change — passes alone.

Remaining red is still the declared merge-order dependency: `gates (kotlin)` reports the single `interest` finding that #5715 fixes on the same lines.

JiRaska:
**Second merge-order dependency, and this one is load-bearing — found by re-checking my own claim.**

This PR's code comment and body both say the rethrow means "the connector dead-letters". I verified that against `origin/main` rather than assuming it, and on main it is **false**: neither `sdd-collection-authorised-in` nor `standing-order-due-in` declares `failure-strategy`, and SmallRye's default is `fail`, which **stops the channel**. Merging this PR alone would trade a silent swallow for a wedged partition on a money-path channel — strictly worse than the defect it fixes.

**#5751 supplies exactly that wiring for both of my channels** (`failure-strategy: dead-letter-queue` plus an explicit nested `dead-letter-queue.topic`, avoiding the dotted-leaf-key trap from #686), with the same analysis in its own comments.

So the required order is now:

    #5715  →  #5719 (this PR's base)  →  #5751  →  #5726

**#5751 is the hard one** — without it this PR is actively harmful, whereas without #5715 it is merely red on one pre-existing finding. Happy to rebase onto #5751 instead if that is easier to sequence, or to fold the two `failure-strategy` blocks in here and let #5751 drop them.

Worth noting the general shape for the rest of the #5698 sweep: **every rethrow fix in it carries this dependency**, and a rethrow landing before its DLQ is a regression, not a partial fix.

JiRaska:
Rebased onto the current base (it had gained `RUN_CATCHING` and `TRANSPORT_RECEIVER` overnight — both kept, and my additions now hook into the base's own `_state_change_in` rather than patching around it).

**One thing worth flagging, because I checked rather than assumed.** The base's `TRANSPORT_RECEIVER` comment says `client` stays in the receiver list *because* `transactionClient.initiateTransaction(...)` is "an actual debit". I ran the base gate against exactly that shape:

```
their gate on the sdd debit shape -> []
verb alternation contains 'initiate' -> False
```

The receiver matches; the **verb** never did. So the comment asserts coverage the regex does not provide — the same comment-vs-code gap this sweep keeps turning up (the sdd consumer's own "must swallow ANY failure" comment, interest's "stays PENDING for retry"). The verb list and the helper-following are therefore both still needed.

**The widened gate then found one more site:** `AuditAnchorService.captureScheduled` — `runCatching { captureAnchor() }` where `captureAnchor()` reaches `anchorRepo.save(...)` one helper deep, invisible to a text scan of the block. Marked `observed-by:` rather than forced to rethrow: it is an hourly `@Scheduled` sweep over the chain head, not a consumed message, so a failed tick loses nothing (next hour re-reads the same head) and throwing would only kill the scheduler thread — while `recordSuccess()` on the success path only means a permanently failing capture climbs the liveness gauge until `WorkflowLivenessStale` fires.

Self-test now 16 cases. Gate on the full tree: **1 finding left**, the `interest` one #5715 fixes — unchanged merge-order dependency, alongside #5751 for the DLQ.

`compileKotlin`/`ktlint`/`detekt` green on audit-service; sdd + audit suites green run serially (a first combined run hit the documented `Port already bound: 8081` collision, not a code failure).

JiRaska:

This PR's base is **`fix/event-handlers-stop-acking-failed-work` (#5719)**, not `main`. Merged the parent in (`805e247b5`).

**No conflicts.** The merge carries exactly one file: notification-service's `PartyErasureConsumerTest.kt`. The `CONFLICTING`/`DIRTY` the API was reporting was stale — GitHub computes `mergeable` lazily, and both merge directions were clean at the current tips before I pushed.

**`check-event-handler-swallows.py` correctly stays at THIS branch's version** (524 lines vs the parent's 312). The widened gate — the `runCatching` form plus the transport-receiver exclusion — is what this PR is *for*, and this branch's copy already contains the parent's `db86c201d`, so taking it loses nothing from either side.

**Entry counts, before → after:**

| file | before | after | lost |
|---|---|---|---|
| `.github/gates/gates.yaml` | 159 | 159 | none (set-compared by gate id) |
| `.release-please-manifest.json` | 58 | 58 | none |

**Build:**

```
:openbank-notification-service:ktlintCheck  PASS
:openbank-notification-service:test         PASS
tests=183  failures=0  errors=0  SKIPPED=0
  PartyErasureConsumerTest  tests=4  failures=0  skipped=0
```

### Note for whoever sequences this stack

Three open PRs each add `.github/scripts/check-event-handler-swallows.py` — a file that **does not exist on `main`** — with three different contents, and all three register the **same gate id `event-handler-swallows`**:

| PR | script | lines |
|---|---|---|
| #5737 | `9ee076c9f` | 310 |
| #5759 | `c55890e9f` | 312 |
| **#5726 (this one)** | `0dcd19c73` | **524** |

Only one of those can land. This PR carries the superset, so sequencing it ahead of its siblings is the resolution that loses nothing — the other two need no script change at all once this is in.

JiRaska:
## Gate-file collision verdict: **this PR's version is the one that should land**

Four branches carry `.github/scripts/check-event-handler-swallows.py` (absent on `main`). I diffed
them against each other and ran a negative control rather than comparing line counts.

**Provenance** — the file is created by the parent #5719 (`024c347f1`, extended by `db86c201d`):

| branch | sha1 of the file | relationship to #5719's copy |
|---|---|---|
| #5719 (parent) | `389ad84d…` | owns it (401 lines) |
| #5759 | `389ad84d…` | **byte-identical** — inherits, changes nothing |
| **#5726 (this PR)** | — | parent **+130/−7**, closes both blind spots |
| #5737 | `9ee076c9…` | diverged from `024c347f1`, **missing `db86c201d`** |

**The negative control settles it.** A synthetic handler with the exact money-path shape — a state
change one private helper deep, via `transactionClient.initiateTransaction(...)` — run through each
version with `--enforce`:

| version | exit | output |
|---|---|---|
| **#5726** | **1** | ``::error … `transactionClient.initiateTransaction(` runs inside a catch that logs and returns`` |
| #5719 / #5759 | 0 | `clean: 1 handler files, no swallowed state changes` |
| #5737 | 0 | `clean: 1 handler files, no swallowed state changes` |

Positive control (same handler rewritten around `EventRetry.withRetry`): #5726 exits **0** — it is
discriminating, not just noisier. `--self-test` exits **0** (`corpus scan reaches 6473 handler files`).

This PR is the only version that closes the two documented blind spots, and the negative control
shows the others report **"clean"** on a real money-path swallow:
- `STATE_CALL` gains the money verbs (`initiate|execute|transfer|charge|capture|reverse|refund|…`) —
  `SddCollectionDebitConsumer`'s debit matched the *receiver* but not the *verb*.
- `LOCAL_CALL` + `_local_fun_bodies` + `_state_change_in(depth=2)` reach a state change one helper deep.

**Action: none for this PR** — keep the file as-is. It is not a duplicate registration: all four
branches inherit the *same* `gates.yaml` entry (`id: event-handler-swallows`, `mode: enforced`) from
the parent commit, so they merge as identical lines. I verified the merged `gates.yaml` has **0
duplicate ids**. The risk here was never a double registration, it is last-writer-wins on the `.py`.

**Ordering matters:** land after #5719, and #5737 must drop its copy of the script first (commented
there) — otherwise whichever merges last silently overwrites the gate's coverage.

#5719's conflict with `main` is resolved and pushed (`dc3852b13`), so this PR's base is current.

<sub>Read-only analysis; nothing merged and no branch of yours touched.</sub>

JiRaska:
### Follow-up: the `CONFLICTING` badge on this PR is **stale** — don't chase it

After #5719's conflict resolution was pushed (`805e247b5` → `dc3852b13`), GitHub flipped this PR to
`CONFLICTING/DIRTY`. That is a stale computation, not a real conflict. The API still reports the
**old** base commit:

```
baseRefOid (GitHub) = 805e247b5   <- pre-push parent
parent HEAD (actual) = dc3852b13
```

A real `git merge` of this PR's head into the updated parent is clean:

```
git merge origin/fix/event-handlers-stop-acking-failed-work --no-commit --no-ff
Automatic merge went well; stopped before committing as requested
exit 0        conflicted paths: (none)
```

(Probe validated against a known-conflicting pair — #5737 vs `main` correctly exits 1 and names the
two fraud-service files — so the exit 0 is a real clean merge, not a probe reporting silence.)

Merging the updated parent should clear the badge. Verified on the merge result: `gates.yaml` has
**167 gate ids, 167 unique, 0 duplicates**, and nothing was silently dropped.

<sub>Local read-only verification in a scratch worktree; this branch was not pushed to or modified.</sub>

**And your gate version survives the merge intact** — I checked specifically, because a clean merge
can still delete content. After merging the updated parent into this branch, the script is still the
524-line version carrying *both* your blind-spot fixes **and** the parent's `runCatching` support:

```
LOCAL_CALL: present      _local_fun_bodies: present
money-verb alternation (initiate|execute|transfer|…): present
RUN_CATCHING (from the parent's db86c201d): present
```

Re-ran the controls on that **merged** file, not just the branch copy:

```
negative control (swallow one helper deep, transactionClient.initiateTransaction) -> exit 1, names the defect
positive control (same handler via EventRetry.withRetry)                          -> exit 0
```

So the recommendation stands unchanged: this is the version to land.

JiRaska:
### The `CONFLICTING` / `This branch has conflicts` banner here looks stale — it merges clean

GitHub shows this PR as `CONFLICTING` / `DIRTY`, but that verdict appears to be a cached
computation against an **older base tip**:

- GraphQL reports `baseRefOid` = `805e247b`
- the live tip of the base branch `fix/event-handlers-stop-acking-failed-work` is `dc3852b1`

Merging locally with `git merge-tree --write-tree` produces a clean tree and exit 0 against
**both** the stale base commit `805e247b` and the current tip `dc3852b1` — no conflicted paths
either way.

That probe was validated against a synthetic add/add conflict first, which it correctly reported
with exit 1, so the clean result is not the probe failing silently.

**Action:** nothing to resolve by hand as far as this can tell. Any push, or a base-branch
update, should force GitHub to recompute and clear the banner. Worth a re-check before anyone
spends time on a manual conflict resolution.

JiRaska:
Replanted as a single signed commit on the current head of the base branch
`fix/event-handlers-stop-acking-failed-work` (#5719, `dc3852b13`) — head is now `44ee8877b`.

The old head was a five-deep merge stack; GitHub reported CONFLICTING, but a real
`git merge` of the old head into the current base went through cleanly, so the conflict was a stale
mergeability cache rather than real divergence. Replanting both removes the merge stack and forces
GitHub to recompute.

The PR's diff against its base is now exactly its own payload, nothing of #5719's:

- `.github/scripts/check-event-handler-swallows.py` (+137, the two widened blind spots)
- `openbank-sdd-service` `SddCollectionDebitConsumer` + test
- `openbank-standing-order-service` `StandingOrderDueConsumer`
- `openbank-audit-service` `AuditAnchorService`, `openbank-balance-service` `ValueDateRollScheduler`,
  `openbank-sanctions-service` `SanctionsListService` (declared-silence comments)

Verification: `check-event-handler-swallows.py --self-test` passes (corpus scan reaches 173 handler
files); `:openbank-sdd-service:test`, `:openbank-standing-order-service:test`,
`:openbank-balance-service:test`, `:openbank-sanctions-service:test` green; ktlintCheck + detekt green
for all five modules. `:openbank-audit-service:test` fails on `AnchorSignerBootIT > signer=kms with a
key produces the KMS signer()` — re-run on the unmodified base commit and it fails identically there,
so it is pre-existing and environmental (no KMS locally), not caused by this branch.

JiRaska:
Re-checked against #5719 at its rebuilt head `ec0474b0b` (32 files), not just against `main`, and
rebased onto it.

Containment result for this PR: **not contained**. #5719 touches `StandingOrderDueConsumer.kt` and
`check-event-handler-swallows.py`, and nothing else this PR ships; the sdd consumer, the audit-anchor
/ balance-scheduler / sanctions declared-silence markers, and the +137-line widening of the gate are
all unique here. The gate script needed a manual add/add resolution (#5719 was force-pushed, so the
two copies share no history) — resolved by blob compare: #5719's script is byte-identical
(`2cecc4224`) before and after its rebuild, and this branch's version is that file plus the widening,
so taking this side loses nothing.

New head `5281dc5c4`, signed (`%G?` = G), diff against base is the same 7 files as before.

Cross-signal worth flagging: run on this branch, the widened gate reports
`1 swallowed state change(s) in 174 handler files` — and the one it names is
`openbank-interest-service/.../WithholdingRemittanceSettlementConsumer.kt:85`, which is exactly what
#5715 fixes. So #5715 and this PR are complements, not duplicates: this branch's gate is only green
once #5715 lands.

Verification on the rebased head: gate `--self-test` passes (174 handler files);
`:openbank-sdd-service:test`, `:openbank-standing-order-service:test`, `:openbank-balance-service:test`,
`:openbank-sanctions-service:test` green; ktlintCheck + detekt green across all five modules.
`:openbank-audit-service:test` fails only on `AnchorSignerBootIT > signer=kms ...`, which fails
identically on the unmodified base commit (no KMS locally).

JiRaska:
## Rebuilt on #5719's current head (`6346f943a`), not on `main`

### `AuditConsumer.kt`
Taken as **`main`'s version verbatim** — #5719's head already carries it byte-for-byte
(`git diff 6346f943a origin/main -- AuditConsumer.kt` is empty). #6209 restructured that file and
states the opposite intent to this branch's edit: the legacy mixed channel deliberately keeps
acking, and the strict path is `AgentAuditConsumer`.

### But the gate then did NOT pass with main's version — three findings, disposed of as follows

`--enforce` on the rebuilt tree flagged three sites. Measured which were mine: running **#5719's own
copy of the gate** over the same tree is `clean: 176 handler files`, so all three come from this PR's
widened detection, which is the point of the PR.

**`AuditConsumer` ×2 — marked, not changed.** Both are #6209's deliberate availability trade. They
now carry the gate's `best-effort` marker with the reason stated plainly, including its cost (a store
failure loses an evidentiary row, and an acked message is indistinguishable from a stored one) and
the pointer to `AgentAuditConsumer` for any producer that cannot tolerate it. **No behaviour change.**

**`WithholdingRemittanceSettlementConsumer` ×1 — fixed.** Worth flagging that
**this one was already red on the PR before I touched it**: `--enforce` on the original head
`5281dc5` reports it too. It is a real money-path swallow found by both new blind spots at once —
`settle(`/`initiate(` were not in the verb list, and the state change sits one private helper deep
(`bookAndSettle`).

Two defects there, not one:
1. The handler swallowed everything, so a transient transaction-service or DB outage acked a
   remittance that was never booked.
2. `bookAndSettle`'s non-2xx/non-409 branch **logged and returned**, and its own message promised
   *"the batch stays PENDING for retry (a redelivery of this event...)"* — which could not happen,
   because returning normally ACKS. There is no re-drive endpoint (the class KDoc says "only SQL
   out"), so a refused booking meant the tax was never paid and the only trace was an ERROR line.

Now: payload decode stays a poison-pill ack; booking throws
`WithholdingRemittanceBookingFailedException` and the whole settle path runs under
`EventRetry.withRetry`. Same shape as `SddDebitFailedException` in this PR.

`a non-2xx-non-409 response does not settle the batch and does not throw` was inverted to
`...is retried and rethrown, never settled or acked`, asserting the exception **and** the attempt
count — the attempt count is what separates a real retry from a single call that happens to throw.

### Verification

```
check-event-handler-swallows.py --enforce   -> exit 0
SUBJECTS=176  # files with @Incoming/@Scheduled handlers
clean: 176 handler files, no swallowed state changes

check-event-handler-swallows.py --self-test -> exit 0  (all cases, incl.)
  ok   a money verb outside the original list is still a state change (sdd: initiateTransaction)
  ok   a state change one private-helper deep is reached (standing-order: executeSepa)
  ok   helper indirection does not invent findings when the helper changes nothing
  ok   a recursive helper terminates instead of hanging the scan

./gradlew :openbank-interest-service:{ktlintCheck,detekt,test} -> BUILD SUCCESSFUL
  ...failed (attempt 1/3, WithholdingRemittanceBookingFailedException: ... did not move money) — retrying
  ...failed (attempt 2/3, ...) — retrying
  ...failed after 3 attempt(s) (...) — rethrowing so the connector dead-letters
```

`:openbank-audit-service:{ktlintCheck,detekt}`, `:openbank-sdd-service:test` and
`:openbank-standing-order-service:test` all ran and passed. (`ktlintFormat` wanted to reformat eight
unrelated interest-service files on pre-existing violations — reverted, only the two files this PR
touches are formatted.)

### Left for a human — a real cross-PR dependency

**None of the three channels this PR makes rethrow has a DLQ on this base.** `sdd-collection-debit`,
`standing-order-due-in` and `interest-withholding-remitted-in` all lack `failure-strategy`, so
SmallRye's default `fail` applies and a rethrow **stops the channel** instead of parking the record.

That was already true of the two sites this PR fixed before I touched it, so I applied the same
convention to the third rather than inventing a different one — but it means **#5751 (which wires all
three) should land first, or at least together.** Merging this alone converts a silent swallow into a
wedged partition on three money-path channels.

PR #5731, opened 2026-08-19
feat(settlement): instrument the settlement saga and alert on its success state

### LINKED ISSUE #5705: settlement-service (money-path) emits no metrics and has no alerts of its own

JiRaska:
## The gap

`openbank-settlement-service` is a declared money-path service and emitted **nothing**. Measured on `origin/main`:

```
grep -rl "micrometer|MeterRegistry|Counter|Gauge|Timer" openbank-settlement-service/src/main
  -> (nothing; 20 .kt files)
```

It also had **no PrometheusRule of its own**. The only alert whose name contains "Settlement" — `ClearingSettlementWindowMissed` — belongs to clearing-service and queries `openbank_clearing_settlements_completed_total`, `service: clearing`.

So there was no signal anywhere that could distinguish *"settlement is running normally"* from *"settlement has booked nothing for six hours"*. The pod is Ready, the namespace is scraped (`payments` is in the fleet PodMonitor), ArgoCD is Healthy, and the series simply did not exist — the repo's recurring shape, one layer earlier than the scrape-coverage sweeps that came before.

## What this adds

`SettlementMetricsPort` (application layer, framework-free) + `SettlementMetricsAdapter` (Micrometer, null-safe via `Instance` like `FraudMetricsAdapter`):

| metric | tags |
|---|---|
| `openbank_settlement_transitions_total` | `status` — every saga transition, compensations included |
| `openbank_settlement_originations_total` | `outcome` — `CREATED` / `IDEMPOTENT_HIT` / `CONCURRENT_RACE` |
| `openbank_settlement_workflow_starts_total` | `outcome` — `STARTED` / `ALREADY_RUNNING` / `NOT_STARTED_TERMINAL` |

### Two naming decisions, both deliberate

**A no-op outcome gets its own enum value, never a shared success flag.** `IDEMPOTENT_HIT` and `ALREADY_RUNNING` are *correct* outcomes and *different* ones from a real create or start. Folding them into a success rate is exactly how a feature-flagged-off push adapter reported every notification as delivered (ADR-0252 phase 0): the row committed `SENT`, the outcome event announced a delivery that never left the process, and nothing disagreed. A shift from `CREATED` to `IDEMPOTENT_HIT` means callers are retrying — visible here, invisible in a success rate.

**The metric is `transitions{status="BOOKED"}`, not `settled`.** That is what settlement-service observed itself commit. Whether money moved is a claim only the ledger can make, and naming a metric for more than it establishes is the same defect as calling an accepted APNs request a delivery.

The counter lives inside the existing `audit(...)` helper, which every transition already funnels through — so a new transition cannot be added without also being counted.

## Alerts, on the success state

Three rules in `components/payments/prometheus-rules.yaml`, all keyed on the *absence of the thing that should be happening* rather than on an error rate. A settlement path that quietly does nothing — a worker that never polls, an orchestrator that never starts a run — produces **no errors at all**, so an error-rate alert is silent in exactly the outage that matters (#2148, five schedulers that had never run).

- `SettlementNoBookingsDuringBusinessHours` — no `BOOKED` transition in 2h during business hours
- `SettlementCompensationsSpiking` — >5 compensating transitions in 15m; money moved and then unwound is a different event from a settlement rejected before moving anything, which is why compensations get their own status values
- `SettlementWorkflowStartsAllNoOps` — every start is `ALREADY_RUNNING` and none is `STARTED`: runs are being created and never completing, a state a success rate reports as 100% healthy

## Verified, and falsified

```
./gradlew :openbank-settlement-service:test   -> 42 tests, 0 failures
```

Falsification — with `metrics.recordTransition(...)` commented out:

```
SettlementActivitiesImplTest > every saga transition is counted, compensations included() FAILED
SettlementActivitiesImplTest > rejectSettlement sets REJECTED status without port calls() FAILED
10 tests completed, 2 failed
```

Exactly the two assertions that claim the metric, and no others. The test double is a **recording** fake, not a mock returning `Unit` — a mock would satisfy the compiler and prove nothing, and could not express "an idempotent hit was counted as an idempotent hit, not as a fresh create", which is the distinction the port exists to preserve.

`ktlintCheck` + `detekt` clean.

## Note on the test call sites

Four tests construct `SettlementService` / `SettlementActivitiesImpl` by hand and needed the new parameter — the shape CLAUDE.md flags as a shared namespace that breaks every hand-instantiating test at once. They were updated to pass the recording fake **and to assert on it**, rather than being given a mock to keep them compiling.

## Linked issues

Closes #5705

### DIFF
--- a/openbank-infra/gitops/components/payments/prometheus-rules.yaml
+++ b/openbank-infra/gitops/components/payments/prometheus-rules.yaml
@@ -45,6 +45,81 @@ spec:
             summary: "Payment sagas are stuck in transaction-service"
             description: "{{ $value }} payment saga(s) have been in an incomplete state for >5m — manual triage required."
 ---
+# settlement-service emitted NO metric of any kind until #5705 — 20 Kotlin files, not one reference
+# to MeterRegistry — and had no PrometheusRule of its own. The only alert whose name contains
+# "Settlement" is ClearingSettlementWindowMissed below, which belongs to clearing-service and
+# queries a clearing series. So nothing could distinguish "settlement is running normally" from
+# "settlement has booked nothing for six hours": the pod is Ready, the namespace is scraped, and
+# the series did not exist.
+#
+# These alert on the SUCCESS STATE, not on an error rate. A settlement path that quietly does
+# nothing — a worker that never polls, an orchestrator that never starts a run — produces no
+# errors at all, so an error-rate alert is silent in exactly the outage that matters. That is the
+# lesson the disabled push adapter cost (ADR-0252 phase 0) and the one the dead schedulers cost
+# (#2148): alert on the absence of the thing that should be happening.
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: openbank-settlement-alerts
+  namespace: payments
+  labels:
+    app.kubernetes.io/part-of: openbank
+    prometheus: openbank
+spec:
+  groups:
+    - name: settlement
+      rules:
+        - alert: SettlementNoBookingsDuringBusinessHours
+          # `transitions_total{status="BOOKED"}` is what settlement-service observed itself commit.
+          # Deliberately not called "settled": whether money moved is a claim only the ledger can
+          # make, and naming a metric for more than it establishes is how an accepted push became
+          # a reported delivery.
+          expr: |
+            increase(openbank_settlement_transitions_total{service="settlement",status="BOOKED"}[2h]) == 0
+            and on() hour() >= 6 and on() hour() <= 18
+          for: 30m
+          labels:
+            severity: warning
+            service: settlement
+          annotations:
+            summary: "No settlement booked in 2h (business hours)"
+            description: >-
+              settlement-service has not committed a BOOKED transition in 2h during business hours.
+              A stalled Temporal worker or an orchestrator that never starts a run looks exactly
+              like this and produces no errors — check the worker is polling
+              (openbank_settlement_workflow_starts_total) before assuming there is simply no volume.
+        - alert: SettlementCompensationsSpiking
+          # Compensations have their own status values rather than folding into a failure count, so
+          # a burst of reversals is visible as itself — a debit that had to be unwound is a
+          # different operational event from a settlement that was rejected before moving anything.
+          expr: |
+            sum(increase(openbank_settlement_transitions_total{service="settlement",status=~"REVERSED|CREDITED_REVERSED|LEDGER_REVERSED"}[15m])) > 5
+          for: 10m
+          labels:
+            severity: critical
+            service: settlement
+          annotations:
+            summary: "Settlement compensations spiking"
+            description: >-
+              {{ $value }} compensating transitions in 15m. Each one means money was moved and then
+              unwound, so a downstream port (debit, credit or ledger) is failing mid-saga.
+        - alert: SettlementWorkflowStartsAllNoOps
+          # An idempotent no-op start is a CORRECT outcome and a different one from a real start.
+          # If every start becomes ALREADY_RUNNING, runs are being created and never completing —
+          # a state that a success rate reports as 100% healthy.
+          expr: |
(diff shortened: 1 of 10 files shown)


JiRaska:
Closing as a duplicate of #5723, which is both **earlier** (16:37 vs 16:54) and **better**. We independently did the same work on #5705 — the parallel-agent collision CLAUDE.md describes, resolved by subtraction.

I compared the content rather than the titles, and theirs wins on the dimension I care most about:

**`SettlementSagaNotCompleting` is stronger than all three of my alerts combined.** It compares work *accepted* against work *finished*, so it is self-normalising: no business-hours guard, no uptime guard, and a freshly started pod makes the first conjunct 0 and the rule silent. My `SettlementNoBookingsDuringBusinessHours` is an unguarded `increase(...) == 0` over a window — exactly the boot-time fourth state this repo has already paid for once (ADR-0237 / #4208), which I have written down in CLAUDE.md and did not apply to my own rule. Theirs cites it and handles it.

Also better in #5723:

- 5 alerts vs my 3, including `SettlementServiceDown` and `SettlementSagaStepsFailing`
- a real `SettlementMetricsAdapterTest` over a genuine Prometheus registry — mine only had a recording fake, so the adapter's own tag and naming behaviour was untested
- `sagaStep(step, outcome)` gives per-step visibility; my single `transitions{status}` counter cannot say *which* step is retrying
- amounts and durations (`BigDecimal`, `Duration`) — mine counted events only

Where we independently agreed, which is a decent signal the shape is right: a no-op outcome gets its own enum value rather than a flag shared with success, and the metric is named for what the service can actually establish (*accepted*, not *settled*).

Nothing in mine is clearly better, so there is nothing worth grafting across. #5705 now points at #5723.

PR #5737, opened 2026-08-19
fix(notification): wire the DLQ the #5698 sweep assumes

JiRaska:
Stacked on **#5719** (merged into this branch so CI is green independently; those files drop out of
the diff once #5719 lands). `Refs #5698`

## What this PR is, after subtraction

I started on the site #5725 flagged and left out of scope — notification-service's
`PartyErasureConsumer` — then found **#5719 had already fixed the same file**, with a shared
`EventRetry` in `openbank-libs-runtime` rather than the per-service `ConsumerRetry.kt` copy #5725
established and I had mirrored. Theirs is the better form, so I deleted mine and took theirs.

What is left is the part no PR in this sweep has:

| file | why it is here |
|---|---|
| `openbank-notification-service/src/main/resources/application.yaml` | `failure-strategy: dead-letter-queue` — **without it the rethrow does not dead-letter** |
| `…/notifications/notification-service-msg-override.yaml` | explicit DLQ topic name |
| `…/notifications/kafka-topic.yaml` | the `KafkaTopic` CR (cluster does not auto-create) |
| `…/notifications/kafka-notification-mtls.yaml` | KafkaUser `Write` ACL on the DLQ topic |
| `…/kafka/PartyErasureConsumerTest.kt` | #5719 changed this consumer with **no test**; these are its tests |

## ⚠️ The whole sweep asserts a DLQ that four services have

Every PR in this family rethrows "so the connector dead-letters". `EventRetry`'s own KDoc says it
"hands the decision to the connector's own failure strategy (retry, then dead-letter)".

**SmallRye's default `failure-strategy` is `fail`, which stops the channel.** It does not
dead-letter. Fleet-wide, exactly four services configure one:

| service / channel | `failure-strategy` |
|---|---|
| fraud `transaction-signal`, transaction `payment-scheme-accepted` | `dead-letter-queue` |
| account `party-events-in` + `delegation-events-in`, card-issuance `delegation-events-in` | `dead-letter-queue` |
| **every other incoming channel in the monorepo** | **unset → `fail`** |

So of the sites across #5715 / #5719 / #5725 / #5726, only the ones on those four channels behave as
documented. The rest stop the data loss by **halting the consumer** — the exact "one bad event wedges
the group" outcome the original swallow comments were written to avoid. That is arguably still the
better trade for a GDPR erasure, but it is not what the KDocs say, and nobody has decided it
deliberately. The repo already knows the mechanism; account-service's own ACL manifest says it:

> Without Write here, the DLQ send itself fails and the consumer wedges on the very failure it was
> meant to park.

This PR therefore wires the DLQ for `party-events-in` rather than asserting one. Explicit topic name,
because `dead-letter-queue.topic` is a dotted leaf key SmallRye's YAML source mis-quotes (#686), and
the implicit default `dead-letter-topic-party-events-in` is **already owned by account-service's
channel of the same name** — leaving it implicit would merge two services' dead letters into one
topic and misattribute them in the existing `AccountPartyEventDeadLettered` alert.

I have not touched the other PRs' branches. Flagged on each.

## Tests, and their falsification

#5719 changed `PartyErasureConsumer` without a test. These are the tests, and they were falsified
against the **pre-fix** consumer:

| test | pre-fix | post-fix |
|---|---|---|
| a persistent erasure failure is RETHROWN | 🔴 | 🟢 |
| a TRANSIENT erasure failure is retried and completes | 🔴 | 🟢 |
| a failure of the SECOND delete also escapes | 🔴 | 🟢 |
| malformed payload is acked without throwing | 🟢 | 🟢 |
| non-PARTY_ERASED events are ignored | 🟢 | 🟢 |
| PARTY_ERASED deletes device tokens and notifications | 🟢 | 🟢 |

Exactly the behaviour tests discriminate; the ack tests hold on both sides, which is what stops a
blanket rethrow from passing them. They now run against #5719's `EventRetry`, so they double as the
first independent check on it. Module gate on the merged state: `ktlintCheck` 0, `detekt` 0,
`quarkusBuild` 0, `test` 41 classes / 184 tests / 0 failures / **0 skipped**.

## The enumeration — every reactive-messaging consumer in the monorepo

The two earlier audits were sampled by service, so absence of findings was not yet evidence.
Searched structurally (the annotation), not by catch-block text: **59** `@Incoming` mentions under
`*/src/main/**.kt`, of which **44** are real annotation sites (15 are KDoc references) across 40
files. Other shapes were checked and none exists — `@Blocking` without `@Incoming` is 16 JAX-RS
resources/filters; the only `Multi<>` declaration is copilot's SSE endpoint (it *produces*); every
`subscribe` is awaited; no direct `KafkaConsumer`; all 60 `@Observes` are Startup/Shutdown; both
`@ConsumeEvent` are a local Vert.x kill-switch; all ~50 `@Channel` injections are outgoing emitters.

**44 sites = 31 genuine · 13 clean or deliberate-and-correct.** Of the 31 genuine, **21 are covered
by a PR** (11 in #5715/#5725, 7 in #5719, 1 here, 2 in #5726) and **10 are not covered by anything.**

### Genuine, covered

| PR | sites |
|---|---|
| #5715 | fraud `TransactionSignal`, interest `WithholdingRemittanceSettlement` |
| #5725 | copilot `PartyErasure`, customer-edge `OnboardingResume`, document ×2, engagement ×5 |
| #5719 | aml `PartyEvent`, audit `AuditConsumer`, card-issuance `PartyEvent`, kyc `PartyEvent`, party `KycAml` ×2, party `MarketingConsent` |
| #5726 | sdd `SddCollectionDebit`, standing-order `StandingOrderDue` |
| this | notification `PartyErasure` (code from #5719; DLQ + tests here) |

### Genuine, covered by NOTHING (10)

| service | consumer | swallowed call | signal lost |
|---|---|---|---|
| anacredit | `LoanStageEventConsumer` | `projections.applyIfNewer` | IFRS 9 stage / DPD frozen → stale AnaCredit regulatory exposure |
| analytics-sink | `AnalyticsConsumer` | `ClickHouseAnalyticsSink.write` | bronze row lost from the ≥10y log of record; the only `DeadLetterSink` binding is a `log.warnf`, so "quarantined" means logged |
| campaign | `EnrolmentTriggerConsumer` (2 channels) | `TriggeredEnrolmentService.enrol` — DB + Temporal `startJourney` | party never enrolled; `reset: latest` rules out replay; can leave a running journey with no enrolment row |
| campaign | `NotificationOutcomeConsumer` | `sendLog.applyDeliveryOutcome` | send row stays `PENDING` forever — the exact ADR-0239 D3 defect it was built to fix |
| notification | `NotificationConsumer` | `dispatch` — persist, contact lookup, email/push send | the notification itself; this channel carries transactional and security templates |
| onboarding | `OnboardingEventConsumer` (3 channels) | `projection.eraseParty`, `projection.applyEvent` | Art. 17 erasure of the read model; KYC/SCA transitions never reach the funnel |
| tax-reporting | `WithholdingRemittedConsumer` | `filingRepository.openIfAbsent`, `remittanceRepository.record` | a remittance missing from the §38d statutory return — silently understated tax withheld |

### Clean or deliberate-and-correct (13)

account `DelegationEvent` + `PartyEvent` (already retry+rethrow) · agent `OversightEvent` (no catch) ·
balance `BalanceInit` · balance `LedgerProjection` (money-path, checked closely — `projection.apply`
unwrapped, so a DB failure nacks) · campaign `CampaignEngagement`, `ConsentEvent`, `Conversion` ×2 ·
card-issuance `CardDelegationEvent` · notification `DelegationNotification` (parse-only; the swallow
it inherits is counted under `NotificationConsumer`) · statement `AccountRegistry` (`upsertOpen`
returned as a `Uni`) · transaction `SchemeAccepted` (catch narrowed to `JacksonException`).

## #5719's gate reports clean, and 10 genuine sites are why that is not evidence

`check-event-handler-swallows.py` prints `clean: 170 handler files, no swallowed state changes` on
this branch. Measured against its own `findings_for`, with a positive control that does fire:

| shape (a real site above) | gate |
|---|---|
| `repo.save(e)` in a plain catch — **positive control** | 🚩 flagged |
| `projections.applyIfNewer` — receiver not in the name list (anacredit) | missed |
| `sink.write` (analytics-sink) | missed |
| `sendLog.applyDeliveryOutcome` (campaign) | missed |
| `projection.applyEvent` (onboarding) | missed |
| `taxFilingService.observe` — receiver matches, **verb** not listed (tax-reporting) | missed |
| `.onFailure().recoverWithUni {}` — Mutiny, no `catch` keyword (notification) | missed |
| swallow in a helper class whose file has no `@Incoming` (campaign `enrol`) | missed |

`STATE_CALL` matches on the receiver's *name* and a closed verb list, and `CATCH` requires a literal
`catch (x: Exception|Throwable)`. Both are reasonable heuristics; neither is a proof of absence, so
the gate's green is about the subset it can express. #5726 closes two of these — worth widening it to
the rest, or the next occurrence of this bug class ships under a green gate.

## Not in scope here

The 10 uncovered sites are not folded in: several are money-path (2 approvals + a threat model by
rule), each needs its own discriminating tests and, per the section above, its own DLQ wiring — and a
ten-service rethrow landing at once is an operational change, not a tidy-up. Filed separately.

Three hedges the audit recorded and deliberately did **not** promote to findings, so nobody
rediscovers them: account-service's `grantWelcomeBonus` (sandbox-only, flag off by default);
`LedgerProjectionService.releaseCoverHolds` (documented best-effort, TTL-compensated);
`ConversionConsumer`'s unconditional `finally { message.ack() }` (needs runtime verification of
SmallRye's commit-vs-shutdown ordering, which I did not run anything to establish).

### DIFF
--- a/openbank-infra/gitops/components/notifications/kafka-notification-mtls.yaml
+++ b/openbank-infra/gitops/components/notifications/kafka-notification-mtls.yaml
@@ -45,6 +45,18 @@ spec:
           patternType: literal
         operations: [Read, Describe]
         host: "*"
+      # #5698: SmallRye dead-letter topic for the party-events-in channel. Without Write here
+      # the DLQ send itself fails and the consumer wedges on the very failure it was meant to
+      # park — the same ANONYMOUS-vs-ACL lockout class that took party-service down (#1476), and
+      # the reason account-service's own ACL carries the identical grant. The name is set
+      # explicitly in notification-service-msg-override.yaml rather than left to SmallRye's
+      # `dead-letter-topic-<channel>` default, which account-service already owns.
+      - resource:
+          type: topic
+          name: openbank.party.events.notification.dlq
+          patternType: literal
+        operations: [Write, Describe]
+        host: "*"
       # ADR-0239 D2: emit delivery outcomes so a producer can tell an accepted handoff
       # from a delivered, suppressed or failed message (issue #3663).
       - resource:
--- a/openbank-infra/gitops/components/notifications/kafka-topic.yaml
+++ b/openbank-infra/gitops/components/notifications/kafka-topic.yaml
@@ -47,3 +47,36 @@ spec:
   config:
     retention.ms: 604800000
     cleanup.policy: delete
+---
+# #5698: dead-letter topic for notification-service's party-events-in channel (SmallRye
+# failure-strategy: dead-letter-queue, set in the service's application.yaml). A PARTY_ERASED
+# event whose GDPR Art. 17 erasure survives PartyErasureConsumer's bounded retry is parked here
+# for inspection/replay instead of being acked and lost.
+#
+# Explicitly named (openbank.party.events.notification.dlq) rather than left to SmallRye's default
+# `dead-letter-topic-<channel>`, because account-service already owns that name for its OWN
+# party-events-in channel — the two would otherwise share one topic and be misattributed by the
+# existing AccountPartyEventDeadLettered alert. The name is supplied via the msg-override
+# ConfigMap, since `dead-letter-queue.topic` is a dotted leaf key SmallRye's YAML source would
+# silently mis-quote (issue #686).
+#
+# A CR is required, not optional: this cluster does not auto-create topics, so without it the
+# DLQ send fails and the consumer wedges on the failure it was meant to park.
+#
+# Longer retention than the source topic: a dead-lettered erasure is a real compliance incident
+# that may outlive a 7-day window before someone replays it. No consumer by design — it is a
+# quarantine, invisible to the event-consumer-liveness gate (that gate reads `outgoing:` channels,
+# and this producer is the connector's implicit DLQ, not a declared channel).
+apiVersion: kafka.strimzi.io/v1
+kind: KafkaTopic
+metadata:
+  name: openbank.party.events.notification.dlq
+  namespace: messaging
+  labels:
+    strimzi.io/cluster: openbank-cluster
+spec:
+  partitions: 1
+  replicas: 1
+  config:
+    retention.ms: 2592000000
+    cleanup.policy: delete
--- a/openbank-infra/gitops/components/notifications/notification-service-msg-override.yaml
+++ b/openbank-infra/gitops/components/notifications/notification-service-msg-override.yaml
@@ -54,6 +54,14 @@ data:
     mp.messaging.incoming.party-events-in.client.id=notification-service-party-${quarkus.uuid}
     mp.messaging.incoming.party-events-in.group.id=notification-service-party
     mp.messaging.incoming.party-events-in.auto.offset.reset=earliest
+    # #5698: explicit DLQ topic for the party-events-in channel (failure-strategy:
+    # dead-letter-queue, set in application.yaml). Explicit rather than implicit because
+    # SmallRye's default name is dead-letter-topic-<channel>, and account-service already owns
+    # `dead-letter-topic-party-events-in` for ITS party-events-in channel — sharing it would put
+    # two services' dead letters in one topic and misattribute them in the existing
+    # AccountPartyEventDeadLettered alert. Dotted leaf key, so it MUST live here and not in
+    # application.yaml (issue #686, same reason as group.id above).
+    mp.messaging.incoming.party-events-in.dead-letter-queue.topic=openbank.party.events.notification.dlq
     mp.messaging.incoming.delegation-events-in.client.id=notification-service-delegation-${quarkus.uuid}
     mp.messaging.incoming.delegation-events-in.group.id=notification-service-delegation
     mp.messaging.incoming.delegation-events-in.auto.offset.reset=earliest
--- a/openbank-notification-service/src/main/resources/application.yaml
+++ b/openbank-notification-service/src/main/resources/application.yaml
@@ -155,6 +155,20 @@ mp:
         # set via the gitops ConfigMap instead (issue #686).
         topic: openbank.party.events
         value.deserializer: org.apache.kafka.common.serialization.StringDeserializer
+        # #5698: PartyErasureConsumer now RETHROWS a failed GDPR Art. 17 erasure instead of
+        # acking it. That is only an improvement if the connector parks the record — SmallRye's
+        # DEFAULT failure-strategy is `fail`, which STOPS the channel, so a rethrow without this
+        # line trades silent data loss for a wedged consumer. `dead-letter-queue` is what makes
+        # the consumer's KDoc true by effect rather than by assertion.
+        #
+        # The DLQ TOPIC NAME is set in the gitops msg-override ConfigMap, not here: the key is
+        # `dead-letter-queue.topic`, a dotted leaf key, and SmallRye's YAML source would register
+        # it quoted and never read it (issue #686 — the same footgun as group.id above). Left
+        # implicit it would default to `dead-letter-topic-party-events-in`, which account-service
+        # ALREADY owns for its own channel of the same name — two services' dead letters in one
+        # topic, misattributed by the existing AccountPartyEventDeadLettered alert. Hence the
+        # explicit per-service name, mirroring fraud-service's `transaction-signal` channel.
+        failure-strategy: dead-letter-queue
       # ADR-0232 delegated-access lifecycle notifications (DelegationNotificationConsumer) — a
       # second consumer of the topic account-service (enforcement projection) and audit-service
       # (onBehalfOf trail) already read.


JiRaska:
Heads-up on overlap — we are both on #5698, and this PR adds a **second implementation** of the same retry.

#5719 puts `EventRetry.withRetry` in `openbank-libs-runtime` (`com.openbank.libs.messaging`) and converts 15 sites across 10 services onto it, including **this exact file** — `openbank-notification-service/.../PartyErasureConsumer.kt`. A service-local `ConsumerRetry.kt` alongside it gives the fleet two helpers with the same job, which is the duplication that PR exists to remove; the four near-identical private copies that already existed are how #5698 stayed unnoticed in so many places at once.

Suggested subtraction, no strong claim on which way round:

- if #5719 lands first, drop `ConsumerRetry.kt` and the `PartyErasureConsumer` change here and keep the Kafka/gitops half, which #5719 does not touch at all;
- if this lands first, I will rebase #5719 onto it and delete my version of that file.

Either way worth deciding before both merge — the enforced `event-handler-swallows` gate in #5719 accepts any rethrowing shape, so it will not catch a duplicate helper for us.

JiRaska:
Heads-up from mapping the #5698 sweep's merge order tonight (full analysis on #5698):

This PR and **#5719 share 17 files, and 14 of them are byte-identical**. The 3 that differ are `gates.yaml`, `check-event-handler-swallows.py` and kyc's `PartyEventConsumer.kt` — and on all three **#5719 carries the newer content**, specifically `RUN_CATCHING` (catches `runCatching { repo.save(x) }.getOrNull()` — the same defect with no `catch` keyword, which #5719's own comment says the first gate version missed in five places) and `TRANSPORT_RECEIVER`.

They do conflict on that file, so nothing merges silently — but the resolution is a human picking sides on a file where one side is simply older, and taking this branch's version wholesale would drop real detection.

Suggest rebasing onto #5719 and keeping only what is genuinely unique here: the 5 notification-service DLQ files (`kafka-notification-mtls.yaml`, `kafka-topic.yaml`, `notification-service-msg-override.yaml`, `application.yaml`, `PartyErasureConsumerTest.kt`). That part is real and not duplicated anywhere.

Also worth confirming the DLQ-before-rethrow ordering: SmallRye's default `failure-strategy` is `fail`, which stops the channel, so a rethrow landing before its DLQ is a regression rather than a partial fix — and git will not warn in either direction.

JiRaska:
## Review — the notification DLQ is the reference implementation of the four-part wiring

I checked the notification-service `party-events-in` channel against all four parts independently, and this is the first one in this batch that has every one of them in the same PR, for the right reasons, written down:

1. **`failure-strategy: dead-letter-queue`** — `openbank-notification-service/src/main/resources/application.yaml:168`. The comment states the actual mechanism: SmallRye's default is `fail`, which **stops the channel**, so a rethrow without this line "trades silent data loss for a wedged consumer". That is the correct reason, not a restatement of the setting.
2. **Explicit per-service topic** — `notification-service-msg-override.yaml:63`, set as `mp.messaging.incoming.party-events-in.dead-letter-queue.topic=openbank.party.events.notification.dlq` in a **`.properties`** source, where the dotted key is read correctly. The `application.yaml` comment explains why it cannot live there (SmallRye's YAML source quotes dotted leaf keys, #686). Crucially it also names the collision being avoided: left implicit it would default to `dead-letter-topic-party-events-in`, which **account-service already owns** for its own channel of the same name — two services' dead letters in one topic, misattributed by the existing `AccountPartyEventDeadLettered` alert. That is criterion 2's failure mode caught before it happened rather than after.
3. **`KafkaTopic` CR** — `openbank-infra/gitops/components/notifications/kafka-topic.yaml:70-81`, with the right justification ("this cluster does not auto-create topics, so without it the DLQ send fails and the consumer wedges on the failure it was meant to park") and a deliberate 30-day retention against the source topic's 7, because a dead-lettered GDPR Art. 17 erasure is a compliance incident that may outlive a 7-day window. `apiVersion: kafka.strimzi.io/v1` matches all 48 existing `KafkaTopic` CRs in the tree — checked, no drift.
4. **KafkaUser Write ACL** — `kafka-notification-mtls.yaml:48-57`, `operations: [Write, Describe]` on that exact topic name, on notification-service's own KafkaUser.

All four, on the correct principal, with the topic name consistent across all three files. This is the shape #5751's gate is trying to enforce and the shape fraud-service is still one CR short of.

Also checked and clean:

- **No `auto.offset.reset` change.** `party-events-in` already carried `earliest` at `notification-service-msg-override.yaml:57` — it appears in the diff only as an unchanged context line. No forced replay on a group running as the default. I looked for this specifically across the whole diff; there are no additions or modifications of that key anywhere.
- The `KafkaTopic` comment pre-empts the obvious objection about the event-consumer-liveness gate (it reads `outgoing:` channels, and this producer is the connector's implicit DLQ, not a declared channel), so the topic having no declared consumer is a stated decision rather than an oversight.
- CI: 43 success, 0 failures; the 19 skips are path-scoped CI and Dependabot automerge, not test skips.

### Requests

- **`.github/scripts/check-event-handler-swallows.py`** — same ask as on #5751: please confirm the self-test drives the real `findings()` path and includes a case for each thing the gate claims to detect, rather than re-implementing the condition inline. This repo has been bitten twice by a gate whose negative control tested a re-implementation instead of the gate. If there is a hand-maintained exclusion list, it should fail on a stale entry in both directions the way #5751's `KNOWN_UNWIRED` does.
- **Conflict warning:** `openbank-fraud-service/.../TransactionSignalConsumer.kt` is edited by this PR, by #5715, and by #6061, all from the same base blob, and #6040 changes the semantics its KDoc describes. Four PRs, one file. They will likely merge textually cleanly in some order and produce a file nobody reviewed as a whole. Whoever lands second should diff the merged result rather than trust the clean merge.
- The other consumers in the sweep (`PartyEventConsumer` in account/aml/kyc/card-issuance, `KycAmlEventConsumer`, `MarketingConsentEventConsumer`, `AuditConsumer`) each depend on their own channel's DLQ being wired — most of those land in #5751, but account-service's `party-events-in` and `delegation-events-in` are explicitly in that PR's `KNOWN_UNWIRED` list and stay on the shared implicit topic. Please confirm which consumers in this PR now rethrow onto a channel whose DLQ is **not** wired, and either hold those or say why the wedge is preferable to the ack.

JiRaska:
## Not resolved — blocked on the same fraud decision, and on stack sequencing

I resolved #5723, #5759 and #5726 today and left this one alone. Two reasons, both structural.

### 1. Its conflict is the parent's, not its own

This PR targets `main`, but it **contains #5719's eight-commit base** (`6c301b6a7` `EventRetry` … `54f9d9f56`), so its diff against `main` spans the whole stack. The two conflicted files are `openbank-fraud-service/.../TransactionSignalConsumer.kt` and its test — the same pair that conflicts on #5715, #5719, #5726 and #5759, and they come from the **parent's** changes, not from this PR's notification-service work.

Resolving them here would mean resolving the parent's conflict in a child, for the third time, in a fourth way. The fix belongs in **#5719** (which is itself `CONFLICTING` against `main`); every child then merges cleanly. That is also what flows main's **#6148** down and clears the `LongMethod` lint failure the stack currently carries in `libs-runtime`.

### 2. The fraud rethrow dead-letters into a topic that does not exist

Same blocker I have written up in detail on #5715. Briefly: the `transaction-signal` channel is `failure-strategy: dead-letter-queue` onto `openbank.transactions.transaction.initiated.fraud.dlq`, that topic has a KafkaUser **Write** ACL and an explicit name but **no `KafkaTopic` CR** (I enumerated all 48 on `main@579dda8e0`; `payment.scheme-accepted.dlq` is present as a positive control), and `kafka.yaml:178` says the cluster does not auto-create topics. #5751 adds 35 DLQ topics and fraud is not among them.

### 3. Three PRs are adding the same gate script

`.github/scripts/check-event-handler-swallows.py` does not exist on `main`, and three open PRs each create it with different contents while registering the **same gate id `event-handler-swallows`**:

| PR | script sha | lines |
|---|---|---|
| **#5737 (this one)** | `9ee076c9f` | 310 |
| #5759 | `c55890e9f` | 312 |
| #5726 | `0dcd19c73` | **524** |

Only one can land. #5726 carries the superset — it adds the `runCatching` detection and the transport-receiver exclusion — so sequencing #5726 first and dropping the script from this PR loses nothing.

### Worth keeping from this PR regardless

The review that flagged this as the reference DLQ implementation is right, and it is the part no sibling duplicates: it is the only one wiring all four pieces together — failure-strategy, an **explicit** per-service DLQ topic, the `KafkaTopic` CR, and the KafkaUser **Write** ACL — and deliberately avoiding the implicit-topic collision with account-service. That is the pattern fraud-service is missing. Applying it to fraud would unblock #5715 and the fraud half of this stack at the same time.

I have not pushed anything to this branch.

JiRaska:
## Action required: drop this PR's copy of `check-event-handler-swallows.py`

This PR adds `.github/scripts/check-event-handler-swallows.py` (310 lines). Three other branches add
the same path registering the same gate id, so only one version can survive — and measurement says
it should not be this one.

**This PR's copy is a regression, not an alternative.** The parent #5719 created the file
(`024c347f1`) and then extended it (`db86c201d`, "catch the runCatching form of the same swallow").
`024c347f1` **is an ancestor of this branch**, but this branch's copy diverged from it and never
picked up the second commit:

```
runCatching references — this PR: 0     #5719: 13
```

So merging this file after #5719 deletes `RUN_CATCHING` / `RECOVERED` / `RETHROWN_TERMINATOR`
detection outright. Nothing in CI would report that: the gate keeps passing, just about less.

**Negative control.** A synthetic handler with a real money-path swallow — a state change one private
helper deep via `transactionClient.initiateTransaction(...)` — run with `--enforce`:

| version | exit | output |
|---|---|---|
| #5726 | **1** | ``::error … `transactionClient.initiateTransaction(` runs inside a catch that logs and returns`` |
| **this PR (#5737)** | **0** | `clean: 1 handler files, no swallowed state changes` |
| #5719 / #5759 | 0 | `clean: 1 handler files, no swallowed state changes` |

**What to drop:** remove `.github/scripts/check-event-handler-swallows.py` from this PR entirely, and
drop the `gates.yaml` registration if this branch adds one of its own (it currently inherits the
parent's entry, which is fine and needs no change). The gate arrives via #5719, with #5726's
blind-spot fixes on top. Keep everything else here — the notification DLQ wiring and the consumer
enumeration are this PR's actual subject and nothing else collides.

This PR is also **CONFLICTING** against `main` independently of the above.

For reference, #5719's own conflict is now resolved and pushed (`dc3852b13`).

<sub>Read-only analysis; nothing merged and this branch was not touched.</sub>

JiRaska:
Replanted on live `origin/main` (head `bef7cf0e8`) and **narrowed to the notification DLQ wiring**,
which is the only part of this branch that is not already delivered elsewhere.

Measured overlap:

- The branch carried #5719's own commits verbatim (`6c301b6a7` EventRetry, `f53ddfcec`, `0cb7eb4e9`,
  `c17500fe3`, `791d2a91b`, `5a414df0a`, `54f9d9f56`) plus `024c347f1` (the gate script). #5719 is
  open, MERGEABLE and based on `main`, so keeping a second copy here only creates a semantic merge
  conflict later.
- `.github/scripts/check-event-handler-swallows.py` on this branch was **96 lines behind** the
  version already on #5719's branch, and #5726 widens it further. Carrying the older copy would have
  reverted both.
- The gitops `notification-service.yaml` image-tag line was auto-deploy drift, not payload.

Kept (unique, nothing else in the #5698/#5745 stack wires it):
`openbank-notification-service/src/main/resources/application.yaml` (`failure-strategy:
dead-letter-queue` on `party-events-in`), the `openbank.party.events.notification.dlq` KafkaTopic,
the KafkaUser Write ACL for it, and the `dead-letter-queue.topic` line in the msg-override ConfigMap
(dotted leaf key, #686).

Verification on the replanted head: `check-duplicate-yaml-keys.sh` 75 files clean;
`check-gitops-duplicate-resources.py` 680 manifests clean; `check-kafka-acl-coverage.py` 107 pairs /
0 gaps; `check-kafka-topics-exist.py` 52 declared / 40 referenced / 0 unresolved;
`check-kafka-dotted-keys.py` clean.

One thing to settle before merge, flagged not resolved: #5751 adds the enforced `incoming-dlq-wiring`
gate with a `openbank.dlq.<service>.<channel>` naming convention and 35 topics — and does **not**
include notification-service. This PR names the topic `openbank.party.events.notification.dlq`.
Whichever lands second may need the name aligned.

JiRaska:
Re-checked against #5719 at its rebuilt head `ec0474b0b`. Blob compare for the files this PR now
ships: `openbank-notification-service/src/main/resources/application.yaml` and
`openbank-infra/gitops/components/notifications/kafka-topic.yaml` are identical on `origin/main` and
on #5719 (`b467bd140`, `04c041fba`) — #5719 changes `PartyErasureConsumer.kt` but wires no DLQ, so
the wiring here is exactly the part #5719 does NOT have, and is what makes its rethrow park a record
instead of wedging the channel (SmallRye's default failure-strategy is `fail`).

No further changes; head stays `bef7cf0e8`.

JiRaska:
**Cross-PR naming collision — whichever of #5737 and #5751 merges second needs a rename.**

- #5751 adds the enforced `incoming-dlq-wiring` gate and establishes the convention
  `openbank.dlq.<service>.<channel>` across 35 topics — and omits notification-service.
- #5737 wires notification-service's DLQ as `openbank.party.events.notification.dlq`.

Both are green in isolation, and `git merge-tree` reports no textual conflict, because they touch
different files — the disagreement is semantic. Merged in either order the result is one
notification DLQ whose name does not match the convention the gate now enforces, so the second merge
turns the gate red on `main` rather than on a PR.

Cheapest resolution: rename #5737's topic to `openbank.dlq.notification.party-events` (or whatever
`incoming-dlq-wiring` derives for that channel) **before** either lands, and check the KafkaTopic,
the Write ACL and the `dead-letter-queue.topic` msg-override line together — they are three separate
files and renaming only the topic leaves the ACL pointing at nothing, which fails closed at
publish time rather than at deploy time.

Same shape as the repo's note that two PRs each green alone can break `main` together; `merge-tree`
clean is not the test here.

JiRaska:
**Correcting my earlier comment on this thread — I named the wrong mechanism.**

I wrote that #5737 and #5751 collide on the DLQ *topic name*. They do not. I read
`check-incoming-dlq-wiring.py` on #5751's head (`FETCH_HEAD:.github/scripts/check-incoming-dlq-wiring.py`)
and it enforces no naming convention at all — it requires `failure-strategy: dead-letter-queue`, an
explicit topic, global uniqueness of that topic, a `KafkaTopic` CR, and a `KafkaUser` Write ACL.
`openbank.party.events.notification.dlq` satisfies every one of those.

**The real collision is that the gate cannot see where #5737 puts the topic.** The gate reads
leaf properties out of each service's `application.yaml`:

```python
topic = p.get("dead-letter-queue.topic")
...
if not topic:
    out.append(f"{svc} :: {ch}: DLQ topic left implicit — SmallRye derives "
               f"`dead-letter-topic-{ch}` from the channel name alone, which collides "
               f"across services sharing it")
```

#5737 deliberately does **not** put the topic there. Its own comment explains why: the key is the
dotted leaf `dead-letter-queue.topic`, which SmallRye's YAML source registers quoted and never
reads (#686) — so the name is set in the gitops `msg-override` ConfigMap instead. That is the
*correct* way to set it, and it is invisible to this gate's probe.

Consequence, concretely: `openbank-notification-service` is **not** in the gate's `KNOWN_UNWIRED`
baseline, so once #5737 lands, that channel has `failure-strategy: dead-letter-queue` and no topic
the gate can find, and the gate reports it as "left implicit" — **red on `main`**, about a channel
that is wired correctly.

So the fix belongs in #5751, not #5737: teach the gate to also read the service's `msg-override`
ConfigMap (where `override.properties` carries `config_ordinal=500`), or add an explicit exemption
that names the ConfigMap it checked instead. An exemption that just baselines notification-service
would be the wrong shape — it would record "unwired" for a channel that is wired.

This is the general trap the repo already knows in other forms: a gate whose probe cannot observe
the correct idiom reports the correct idiom as a defect. Same family as grepping `src/test` for the
word "contract", and as the event-payload keys that only exist at runtime on a serialised data class.

Issue #5733, opened 2026-08-19
4 alert rules can never fire — they depend on openbank_* metrics no code emits (1 is critical, money-path)


JiRaska:
## What

Four alert rules depend on `openbank_*` metrics that **no code in this repository emits and no recording rule produces**. They load cleanly, `promtool` accepts them, Alertmanager shows a healthy route, and Prometheus lists each as `inactive` — which is exactly what a correctly-quiet alert also looks like. They have never been able to fire and never will.

| Alert | Severity | Missing metric | File |
|---|---|---|---|
| `TransactionSagaStuck` | **critical** | `openbank_transaction_sagas_stuck_total` | `components/payments/prometheus-rules.yaml` |
| `ClearingSettlementWindowMissed` | warning | `openbank_clearing_settlements_completed_total` | `components/payments/prometheus-rules.yaml` |
| `SwiftMtMessageProcessingStalled` | warning | `openbank_swift_messages_processed_total` | `components/payments/prometheus-rules.yaml` |
| `LendingRepaymentProcessingStalled` | warning | `openbank_lending_repayments_processed_total` | `components/lending/prometheus-rules.yaml` |

## How this was established

Two independent methods agreed:

1. **Live**: queried the sandbox Prometheus for `__name__` values — 59 `openbank_*` metrics exist; none of these four is among them. Then confirmed no *code* path could ever produce them.
2. **Static**: scanned every `openbank-*/src/main/**/*.kt` for metric-name literals (comments stripped) across all three naming idioms this repo uses — dotted (`openbank.payments.submitted`), underscored (`registry.counter("openbank_statement_close_failures_total")`) and mixed (`openbank.ledger.accounting_day.stuck_cutoff_days`). Zero matches for all four.

## Why it matters most for `TransactionSagaStuck`

It is `severity: critical`, on the money path, and its entire job is to page a human when a transaction saga wedges. That has never worked.

The three `warning` ones share a second, subtler defect worth noting even after the metric exists: they are written as `increase(x[1h]) == 0`. An **absent** series produces an empty vector, so `== 0` matches nothing — the alert that exists to detect "this stopped happening" is structurally blind to the most severe form of it, total absence. Compare `PushChannelSilent`, the one alert in this estate that gets it right with `absent_over_time(...)`.

## Fix

[...]

PR #5739, opened 2026-08-19
feat(ci): gate that an alert rule depends only on metrics some code emits

JiRaska:
## What

Adds `check-alert-metric-emitted.py` — an enforced gate that fails when an alert rule depends on an `openbank_*` metric nothing in this repo emits.

Found by a full observability audit of the live sandbox (2026-08-19). Four alert rules can never fire:

| Alert | Severity | Missing metric |
|---|---|---|
| `TransactionSagaStuck` | **critical** | `openbank_transaction_sagas_stuck_total` |
| `ClearingSettlementWindowMissed` | warning | `openbank_clearing_settlements_completed_total` |
| `SwiftMtMessageProcessingStalled` | warning | `openbank_swift_messages_processed_total` |
| `LendingRepaymentProcessingStalled` | warning | `openbank_lending_repayments_processed_total` |

## Why nothing caught this

A dead alert rule is invisible from every angle you would normally check: it loads, `promtool check rules` accepts the PromQL (the expression is valid over a metric that simply has no data), Alertmanager shows a healthy route, and Prometheus lists it as `inactive` — **which is exactly what a correctly-quiet alert also looks like**.

`TransactionSagaStuck` is `severity: critical`, on the money path. Its entire job is to page a human when a transaction saga wedges. It has never been able to do that.

Same class as #5049 (16 dashboard metrics nothing emitted), one layer worse: an empty panel is at least visible to whoever opens the dashboard; a rule that never fires produces no artifact at all.

## Two bugs in the gate itself, caught by checking it against live Prometheus

I did not trust the gate's first output — I cross-checked it against the live metric inventory, which is how both of these surfaced:

**1. It understood only one of three naming idioms**, so it flagged six live, working alerts (`WorkflowLivenessStale`, `FxFixingStale`, `DomesticPaymentStranded…` and others). This repo writes metric names three ways:

```kotlin
"openbank.payments.submitted"                              // DomainMetrics, dotted
registry.counter("openbank_statement_close_failures_total") // CloseMetricsAdapter, Prometheus-style
"openbank.ledger.accounting_day.stuck_cutoff_days"          // AccountingDayScheduler, mixed
```

Both sides are now canonicalised (separators collapsed, registry suffixes stripped) before comparison. That takes the finding count from **14 → exactly the 4** that live Prometheus independently confirms.

**2. `prom_to_micrometer` looped forever** — it set `changed` on every suffix match rather than only on a *newly derived* base, and the original element keeps matching its own suffix on every pass. That is why the first selftest run hung rather than failing.

## Prose is not instrumentation

Comments are stripped before the literal scan. Metric names appear in KDoc all over this tree — `FxFixingFreshnessPort` names three it does not emit — and counting prose as instrumentation is how a gate ends up certifying a metric that exists only in a sentence. Same trap as grepping `src/test` for the word "contract".

## The four are declared, not hidden

`KNOWN_DEAD` is keyed by `alert#metric`, checked **both ways** so a fixed entry is itself reported and the list can only shrink. Keying by alert rather than metric alone stops one debt entry silently excusing a *future* rule that reuses the same metric — the selftest proves both halves.

A new dead alert matches nothing in the baseline and fails immediately. That is the point: the estate cannot quietly acquire a fifth one.

## Verified

- selftest passes standalone and through `run-gates.py` — 7 expression cases both directions, plus baseline-honoured and baseline-not-over-applied.
- enforce run clean: 115 alert rules, 212 emitted metric names, 5 recording rules, 4 baselined.
- `gatelib.subjects()` wired (`min_subjects: 80`), `ruff --select E9,F,B` clean (the exact CI invocation), 18.5s through the runner.

## Not in scope

Fixing the four alerts — that is #5733, and each needs a judgement call (instrument the metric / correct a typo'd name / delete the rule) that belongs with someone who owns the money-path alerting.

Refs #5733


JiRaska:
Updated: merged `origin/main` and #5758's branch in, so this PR and the fix are consistent in either merge order.

- Baseline is now **3**, not 4 — `TransactionSagaStuck` is fixed by #5758 rather than declared, so the gate judges it on merit and passes. That is the ratchet moving the intended direction on its first day.
- The selftest caught its own staleness: its baseline-honoured fixture named `TransactionSagaStuck`, and removing the entry made that case **fail** rather than quietly assert nothing. Repointed at `ClearingSettlementWindowMissed`.
- Merge note worth flagging: this branch was cut from an older `main`, and merging only the fix branch produced a tree that would have **deleted 38 lines** from domestic-payment's `ApprovalResource.kt` relative to current `origin/main` — a clean merge silently reverting newer work. Merging `origin/main` first brings the diff back to the 7 files this stack actually owns; verified with `git diff --name-only origin/main`.

Re-verified after the merge: selftest passes, enforce clean at 115 alert rules / 214 emitted names / 3 baselined, ruff clean.

JiRaska:
Superseded by the rebuilt branch: this one went DIRTY after #5758 was closed (superseded by #5787), because it carried #5758's now-redundant TransactionStrandedGauge merged in.

Rebuilt off current main carrying **only** the gate, and — better — with an **empty** baseline: all four dead alerts are now resolved on main (#5787 instrumented TransactionSagaStuck; #5733 removed the other three as throughput floors on services with no cadence to miss). The gate ships at zero debt rather than four.

PR #5743, opened 2026-08-19
fix(ci): admin-ui dep probe must import, not resolve (main's lint shard is red)

JiRaska:
`gates (lint)` — and therefore the required `Validate manifests` check — has been failing on **main**
since #5662, so nothing can merge. Not caused by any open PR.

## Cause

```
ERR_UNSUPPORTED_DIR_IMPORT: Directory import '.../node_modules/es-toolkit/compat'
  is not supported resolving ES modules imported from '.../mermaid/dist/mermaid.core.mjs'
```

`ensure_admin_ui_deps` decided node_modules was usable with `require.resolve`. That finds mermaid's
package entry and says nothing about whether **importing** it works — mermaid.core.mjs imports
`es-toolkit/compat`, and against a tree whose es-toolkit lacks its `exports` map that import fails
while every `require.resolve` still succeeds. The installer therefore skipped, mermaid-parses could
not run its own self-test, and the shard reported it UNFALSIFIED: the right verdict, with no hint
that dependencies were the cause.

## Fix

The probe now asks what the gates ask — can these four modules be **imported**, from the directory
the gates import them from — and a tree that fails is reinstalled from the lockfile.

## Verified by reproduction, not by reasoning

Deleting the `exports` block from the installed es-toolkit's `package.json` reproduces the CI error
exactly:

| probe | broken tree |
|---|---|
| old (`require.resolve`) | **PASS** — would have trusted it |
| new (`import`) | **FAIL** — rejects it, triggers the reinstall |

With the fix, the gate reinstalls and goes green: `mermaid 11.16.1: 0 failing / 248 blocks in 1102
markdown files`.

Also added, because diagnosing this from the log was impossible (node version, npm flags and a
nested duplicate were all consistent with the evidence): the failure path prints the resolved
versions plus the real import error, and a reinstall that does not restore importability now fails
loudly next to the install meant to fix it.

## Linked issues

Refs #5662

### DIFF
--- a/.github/scripts/ensure-admin-ui-deps.sh
+++ b/.github/scripts/ensure-admin-ui-deps.sh
@@ -37,9 +37,36 @@ ensure_admin_ui_deps() {
   ui_dir="$repo_root/openbank-admin-ui"
   lock="$ui_dir/.node_modules.gate-lock"
 
+  # RESOLVABLE IS NOT IMPORTABLE, and the difference is what took the lint shard red on main
+  # (2026-08-19). `require.resolve` finds mermaid's package entry and says nothing about whether
+  # importing it works: mermaid.core.mjs imports `es-toolkit/compat`, and against a tree where that
+  # dependency is missing its `exports` map the import dies with
+  #   ERR_UNSUPPORTED_DIR_IMPORT: Directory import '.../es-toolkit/compat' is not supported
+  # while every `require.resolve` above still succeeds. The gate then reported UNFALSIFIED — its
+  # self-test could not run — which is the correct verdict and gave no clue that the cause was a
+  # half-right node_modules rather than the checker.
+  #
+  # So the probe now asks the question the gates actually ask: can these modules be IMPORTED, from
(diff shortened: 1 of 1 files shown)


JiRaska:
Closing as superseded by #5742, which fixes this better.

**Why mine was the wrong diagnosis.** I reproduced a real gap — `require.resolve` accepts a tree
whose `import` fails, verified by stripping the `exports` block from the installed es-toolkit — but
that proves the probe *can* be fooled, not that it *was*. The evidence says otherwise: `gates (lint)`
passed on sha `87473a5f` after failing on the same tree earlier, i.e. the failure was
**intermittent**. A deterministically broken node_modules cannot do that; a half-installed one
racing two gates in the same shard can, and that is exactly what #5742 identified and fixed with a
completion marker written only after `npm ci` returns 0.

`main` is green now (runs 32280207207, 32281028561), so there is nothing left for this PR to fix,
and merging it would partly reinstate the resolve-based probe #5742 deliberately removed.

Nothing lost: the marker approach subsumes the importability concern, since the marker is only
written after a successful full install.

PR #5747, opened 2026-08-19
feat(party): alert on parties stalled in PENDING_KYC (#5698 detection)

JiRaska:
Closes the **detection** box #5698 left unchecked: *"nothing today notices a party that never got a
case."* #5699/#5715/#5719 fixed the cause; this makes the next occurrence visible.

Ten of 73 sandbox parties were stranded in `PENDING_KYC`, the oldest for ten weeks, and the first
thing to notice was a customer saying their account did not work.

## Why a census gauge, not a rate

A swallowed failure produces **no error to count**: no exception escapes, consumer lag is zero, the
DLQ is empty, no request failed. The only observable is the party still sitting in the stage — so
the signal has to be a periodic census of the state itself.

Checked that all three existing onboarding alerts stay silent for this incident, **by
construction**:

| existing alert | why it was silent |
|---|---|
| `PartyEventDeadLettered` | the event was **acked** — it never reached the DLQ |
| `OnboardingAccountsNotFollowingParties` | other parties opened accounts that hour, so the fleet-wide disparity never appeared |
| `KafkaConsumerGroupLagStuck` | an acked message is not lag |

## Design choices, and what each rules out

**Age, not a raw count.** `PENDING_KYC` is the *correct* state between creation and the KYC verdict,
so a raw count alarms on healthy traffic. Only parties older than the stall window (default 24h,
configurable) are counted — which is why `> 0` is the right trigger and no arbitrary count threshold
is needed to let a small number hide.

**Symptom, not a party↔kyc reconciler.** A reconciler would catch only the missing-case cause and
would need a cross-service read to do it. Counting the symptom catches every cause with no new trust
boundary: the lost event, kyc-service unreachable, or a case opened and then stuck waiting on a
reviewer. *"Somebody's onboarding is stuck"* is the same alert in all three.

**`stage` is a metric label, not part of the name** — a second stalled stage (an account stuck
`PENDING_ACTIVATION`, a signature ceremony never completed) reuses `openbank_onboarding_stalled`
and the alert as-is.

## Who watches the watcher

The `scheduler-liveness-adoption` gate (ADR-0237) **caught this PR's first version**, and was right
to: a detector whose own tick can die silently has the exact failure mode it exists to catch — a
frozen gauge reads as a healthy zero. Now registers `registerWorkflowLiveness` (recorded only on the
completed path), and `refresh()` deliberately does **not** catch its own exception.

## Verification
- [x] `:openbank-party-service:test --tests "*PartyPendingKycStalledGaugeTest*"` — **3 tests, 0 skipped** (read from the JUnit XML, not the task status)
- [x] `ktlintCheck`, `detekt`, `compileKotlin`, `compileTestKotlin` — green
- [x] `run-gates.py --group gitops` — **PASS=30**, same as clean `main` (the one `UNFALSIFIED` loki entry is pre-existing, confirmed by stashing)

Tests assert the three properties that make the gauge usable: it reports the aged count, it
**excludes** parties younger than the window (the anti-false-alarm property), and a failing query
**propagates** instead of freezing the gauge at a healthy-looking value.

## Minor, noted not fixed
`gen-network-policies-drift-gate` runs `git diff --exit-code openbank-infra/gitops/components/` —
the whole directory rather than the generated `network-policies.yaml` files — so **any** uncommitted
change under `components/` fails it locally. Harmless on CI (the tree is clean there); it cost a few
minutes of diagnosis locally. Left alone rather than scope-creeping this PR.

Refs #5698

### DIFF
--- a/openbank-infra/gitops/components/observability/prometheus-rules-onboarding.yaml
+++ b/openbank-infra/gitops/components/observability/prometheus-rules-onboarding.yaml
@@ -64,6 +64,33 @@ spec:
           annotations:
             summary: "Parties are being created but no accounts are opening"
             description: "openbank_parties_created_total rose over the last hour while openbank_accounts_created_total did not — new customers are registering but the PARTY_CREATED → account.created pipeline is not producing accounts. This is the onboarding outage signature (party-service → Kafka → account-service). Check the party-events consumer lag, the DLQ (PartyEventDeadLettered), and account-service's dependencies (sanctions-service, product-catalog)."
+        - alert: OnboardingStalledInStage
+          # The gap the three rules above cannot cover, and #5698 proved it: a party created on
+          # 2026-08-19 never got a KYC case because the consumer ACKED the failure. Every signal
+          # here stayed silent, by construction —
+          #   PartyEventDeadLettered:              the event was acked, so it never reached the DLQ;
+          #   OnboardingAccountsNotFollowingParties: other parties opened accounts in the same hour,
+          #                                        so the fleet-wide disparity never appeared;
+          #   KafkaConsumerGroupLagStuck:          an acked message is not lag.
+          # Ten of 73 sandbox parties were stuck this way, the oldest for ten weeks, and the first
+          # thing to notice was a customer saying their account did not work.
+          #
+          # A swallowed failure leaves NO error to count — no exception, no lag, no DLQ row, no
+          # failed request — so the only observable is the entity still sitting in the stage. Hence
+          # a census gauge (party-service refreshes it every 5m) rather than a rate.
+          #
+          # `stage` is a label so a second stalled stage (an account stuck PENDING_ACTIVATION, a
+          # signature ceremony never completed) reuses this rule as-is. The gauge only counts
+          # entities past that stage's own age threshold (party: 24h), so in-flight onboarding is
+          # already excluded and `> 0` is the correct trigger — no arbitrary count threshold that
+          # would let a small number hide.
+          expr: openbank_onboarding_stalled > 0
+          for: 30m
+          labels:
+            severity: critical
+          annotations:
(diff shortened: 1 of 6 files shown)


JiRaska:
Closing in favour of **#5748**, which solves the same #5698 detection box and is the better piece of work. Parallel effort — neither of us saw the other's PR.

**Where #5748 is strictly better, and one of them is a flaw in mine:**

- **It publishes the denominator** (`openbank_kyc_orphan_detection_parties_scanned`). Mine does not, so `openbank_onboarding_stalled == 0` is byte-identical to "the query returned nothing because it is broken" — the repo's own *a check passes because it never reached its SUBJECT* trap, in the very control written against silent failure. #5748 names that explicitly and fixes it.
- **It distinguishes the boot-time zero** from a measured zero, cross-checked against `parties_scanned` and the liveness heartbeat, and it *measured* that rather than reasoning about it.
- **A failed pass keeps the previous values** rather than resetting to 0, so an unreachable party-service cannot clear a firing alert. Mine propagates the exception (good) but the cached value semantics are less carefully argued.
- **`ACTIVE` is deliberately not excluded** — a party that reached ACTIVE with no KYC case means activation bypassed the KYC gate, which is worse than a stranded `PENDING_KYC`. My symptom count would have missed that entirely, because such a party is no longer `PENDING_KYC`.
- 7 unit + 3 integration tests vs my 3.

**The one thing mine covered that #5748 does not**, offered as an observation rather than a competing PR: a party whose case *exists* but is stuck (awaiting a reviewer who never came) stays `PENDING_KYC` and is invisible to an orphan query keyed on "no case at all". That may well be a normal operational state the ADR-0068 cockpit already surfaces rather than a defect worth paging on — worth a look when #5748 lands, not worth blocking it for.

The generic `stage`-label design here was also speculative: no second stalled stage exists today, so it was building for a hypothetical.

Left #5698's detection box to #5748.

PR #5748, opened 2026-08-19
feat(kyc): detect parties with no KYC case via reconciler (#5698)

JiRaska:
## Summary

`PartyEventConsumer` acked `PARTY_CREATED` on failure, so seconds of `kyc-db` downtime destroyed the only copy of the event — no KYC case, party stuck in `PENDING_KYC`, accounts stuck in `PENDING_ACTIVATION`. **10 of 73 sandbox parties are in that state, the oldest since 2026-06-07**, and nothing went red: it was found because a human noticed an account did not work. This builds the unchecked "Detection" item on #5698 — an hourly read-only reconciliation that compares the party register against `kyc_cases` and alerts on the gap.

#5699 stops *new* losses by rethrowing so the connector dead-letters. It cannot see a party already stranded, and no retry or replay can — the event is gone. Only comparing the two services' state finds those.

## Type of change

## Linked issues

Refs #5698

## Changes

- **`OrphanedPartyDetector`** (application) — pages the party register, reports parties past a grace period with no `kyc_cases` row. One batched `IN` query, not N+1.
- **`PartyDirectoryPort` / `PartyServiceClient` / `PartyDirectoryAdapter`** — read-only enumeration over `GET /api/v1/parties`, mirroring the existing `SanctionsServiceClient` stack. Binds only `id`, `status`, `createdAt`; `legalName`/`email`/`tradingName` are deliberately **not** mapped, so the PII cannot be logged by accident.
- **`KycCaseRepository.findPartyIdsWithAnyCase`** — scalar HQL projection.
- **`OrphanedPartyGauge`** — `@Scheduled` `suspend fun` + three gauges + ADR-0237 liveness heartbeat.
- **`KycPartiesWithoutCase`** alert in `prometheus-rules-onboarding.yaml`.
- gitops: `PARTY_SERVICE_URL` on the kyc Rollout; `party/network-policies.yaml` regenerated (the generator derived the new kyc→party ingress itself).

## Why the comparison lives in kyc-service

The two sides are separate services and separate DBs, so there is no join to write. The dependency direction is already settled — kyc-service consumes party events and calls party-service, never the reverse — so enumerating parties from here adds no new coupling, while teaching party-service about KYC cases would. Follows the fleet's established reconciliation shape (`BalanceReconciliationScheduler` ADR-0039, `AccountReconciliationRepository` ADR-0026): scheduled, read-only, publishes drift, mutates nothing.

**It deliberately does not remediate.** The 9 remaining orphans need a replay against a live DB — explicitly out of scope here and still open on #5698. Auto-opening cases from a monitoring job would also PEP-screen historical parties as a side effect.

## The metrics, and the cold-pod reading

```
openbank_kyc_orphaned_parties                     -> KycPartiesWithoutCase alerts on this
openbank_kyc_orphaned_parties_oldest_age_seconds  -> triage
openbank_kyc_orphan_detection_parties_scanned     -> the denominator
```

The denominator is published because **a pass that enumerated zero parties reports zero orphans, which is byte-identical to a healthy register.** Without it no series anywhere separates "nothing is wrong" from "this control is looking at nothing" — the exact failure mode #5698 is about.

**At t=0 on a cold pod every gauge reads 0**, and the alert is `> 0`, so a fresh pod cannot fire it. This is the deliberate opposite of the `Instant.EPOCH` seeding that made `WorkflowLivenessStale` fire 15 minutes after every deploy for every daily workflow (#2239). That boot-time 0 is a real fourth state — "not yet measured" — and it is resolved by the *other* signals rather than left ambiguous: `parties_scanned` is also 0 until the first pass, and the `kyc-orphaned-party-detection` liveness heartbeat (seeded at registration, so also boot-safe) fires `WorkflowLivenessStale` if the job never completes a pass. So "no orphans" and "never ran" are distinguishable, and neither alarms at boot. **This was measured, not reasoned about** — see the falsification below, where the gauges sat at 0 exactly as predicted.

A failed pass keeps the previous values and records no liveness success: resetting to 0 would let an unreachable party-service publish "no orphans" and clear a firing alert.

## Grace period

A party legitimately has no case for a short window after creation. The grace period (default `PT1H`, configurable) is enforced **in the service**, so the gauge only ever counts parties past it and the alert cannot fire on a new signup; `for: 30m` on top absorbs a single failed pass. `CLOSED`/`MERGED` are excluded (they may legitimately have no case, or had one hard-deleted by AML retention — flagging them would alert on correct behaviour). `ACTIVE` is **not** excluded: a party that reached ACTIVE with no KYC case means activation happened without the KYC gate, which is worse than a stranded `PENDING_KYC`.

## Definition of Done

- [x] Unit tests added (7).
- [x] Integration tests added (3).
- [ ] Contract tests — n/a, no public API change (no new endpoint, `openapi.yaml` untouched).
- [x] `ktlintCheck` / `detekt` / `test` / `quarkusBuild` pass locally.
- [x] No new lint warnings.
- [ ] OpenAPI — n/a. [ ] Flyway — n/a. [ ] Avro — n/a.

## Testing, and the proof the scheduler actually runs

Full kyc suite: **94 tests, 0 failures, 1 pre-existing skip** (both new classes: 0 skipped).

`OrphanedPartyDetectorTest` covers the rule: orphan **is** detected; party inside the grace period is **not**; party with a case is **not**; `CLOSED`/`MERGED` excluded while `ACTIVE` is flagged; unknown status degrades to a false positive rather than a false negative; paging de-duplicates.

`OrphanedPartyDetectionSchedulerIT` proves the *cron* runs, which no mocked test can: a plain `@Scheduled` method carries no Vert.x context, so the reactive query throws `HR000068` and the pass aborts silently (#2148, #2187) — and calling the method directly supplies the very context the scheduler does not. It drives the real cron from a `@TestProfile` (shrunk to 2s, all-literal values because the profile loads in a different classloader) and asserts only on published gauges.

**Falsified, not assumed.** Reverting `refresh()` to a plain non-suspend function wrapping a blocking coroutine builder makes all three IT assertions fail, with `HR000068 ... currently running on thread 'executor-thread-1'` in the log and the gauges still at their boot-time 0. No bind error in that run — the app booted fine and the scheduler genuinely misbehaved.

Two real defects were caught by these tests during development, both of which compiled cleanly:
1. Panache `.project()` resolves columns from **constructor parameter names**, needing the `-parameters` flag this build does not set — it threw `PanacheQueryException` only at runtime. Replaced with scalar HQL.
2. Five of the seven unit tests were **silently skipped by JUnit5**: `fun x() = runBlocking { … }` infers a non-`Unit` return type. Fixed with `: Unit`.

A third test asserts the gauges appear in a real `PrometheusMeterRegistry.scrape()` under the exact underscored names the alert queries — nothing else in the repo checks that a rule's `expr` names a series that is actually emitted.

## Security checklist

- [x] No secrets, tokens, or PII in code, config, tests, logs. The reconciliation reads no name, email, or birth number; only opaque party ids are logged.
- [x] `@Suppress("TooGenericExceptionCaught")` on `refresh()` — justified: a scheduler tick must never crash the runtime and every fault has identical handling (keep last values, let the heartbeat go stale). Same rationale as `BalanceReconciliationScheduler`, which carries the equivalent detekt baseline entry.
- [x] No new third-party dependency (every needed lib was already on kyc-service's dependency list).
- [ ] PII path — not changed; PII exposure is *reduced* relative to a naive client by not binding those fields.

## Compliance impact

- [x] AML / 5AMLD — this is a detection control over KYC case coverage. It **adds** assurance (a party with no KYC case was previously undetectable) and weakens nothing. It opens, closes, and modifies no case, and the AML 5-year retention path is untouched — `CLOSED`/`MERGED` exclusion exists precisely so correct retention behaviour is not reported as drift.
- [ ] PCI DSS / DORA / GDPR / PSD2 / CNB — none.

## Rollout / Rollback

- Deployment strategy: rolling (existing Argo Rollout).
- Rollback plan: revert the PR, or set `KYC_ORPHAN_DETECTION_ENABLED=false` to disable the job without a deploy. Read-only, so there is no state to unwind.
- Data migration reversible: N/A — no schema change.

## Reviewer notes

- **One ASVS baseline line added** (`.github/asvs-l3-baseline.txt`): `PARTY_SERVICE_URL` is a plaintext in-cluster `http://` edge, the same shape as all 232 existing entries — including kyc-service's own `SANCTIONS_SERVICE_URL` two lines above it in the same file. In-cluster plaintext is the fleet's declared architecture, so this is consistent treatment of an existing category rather than new debt of a novel kind. Flagging it explicitly in case you disagree.
- **`party/network-policies.yaml` is derived** — regenerated with `openbank-infra/scripts/gen-network-policies.py`, not hand-edited. The generator picked up the new kyc→party edge from the Rollout env var on its own; the drift gate passes.
- Gate status: `run-gates.py --all` gives **147/153 PASS with zero new failures** — the remaining ones fail identically on clean `origin/main` (verified in a separate worktree) and are environmental: `api-contract` wants `sudo` to install `oasdiff`, `release-scope` needs `PR_TITLE`, `loki-rule-load-test` is unfalsified upstream. Re-run with a real `PR_DIFF_BASE`, the enforced **threat-model, db-migration and schema-compat gates all pass**.
- The party-register scan is capped at `max-pages` (default 500 x 100 = 50k parties) and logs loudly if it ever hits the cap, rather than silently truncating.

### DIFF
--- a/.github/asvs-l3-baseline.txt
+++ b/.github/asvs-l3-baseline.txt
@@ -54,6 +54,7 @@ V9.1 openbank-infra/gitops/components/devops-agent/devops-agent.yaml:56: plainte
 V9.1 openbank-infra/gitops/components/devops-agent/devops-agent.yaml:65: plaintext http:// env value http://litellm.ai-platform.svc:4000/v1
 V9.1 openbank-infra/gitops/components/kyc/kyc-service.yaml:113: plaintext http:// env value http://tempo.observability.svc:4317
 V9.1 openbank-infra/gitops/components/kyc/kyc-service.yaml:122: plaintext http:// env value http://sanctions-service.sanctions.svc:8123
+V9.1 openbank-infra/gitops/components/kyc/kyc-service.yaml:130: plaintext http:// env value http://party-service.party.svc:8111
 V9.1 openbank-infra/gitops/components/anacredit/anacredit-service.yaml:63: plaintext http:// env value http://tempo.observability.svc:4317
 V9.1 openbank-infra/gitops/components/customer-edge/customer-edge.yaml:143: plaintext http:// env value http://account-service.accounts.svc:8100
 V9.1 openbank-infra/gitops/components/customer-edge/customer-edge.yaml:153: plaintext http:// env value http://product-catalog.accounts.svc:8104
--- a/openbank-infra/gitops/components/kyc/kyc-service.yaml
+++ b/openbank-infra/gitops/components/kyc/kyc-service.yaml
@@ -120,6 +120,14 @@ spec:
             # (never a silent PASSED), but every new case would need an operator's attention.
             - name: SANCTIONS_SERVICE_URL
               value: http://sanctions-service.sanctions.svc:8123
+            # Orphaned-party reconciliation (issue #5698): kyc-service enumerates the party
+            # register hourly and reports parties that never got a KYC case, as
+            # openbank_kyc_orphaned_parties. Read-only — it opens no cases. Without this the
+            # rest-client falls back to its localhost dev default and every pass fails, which
+            # shows up as a stale kyc-orphaned-party-detection liveness heartbeat rather than
+            # as a silent "no orphans" (the gauges are not reset on a failed pass).
+            - name: PARTY_SERVICE_URL
+              value: http://party-service.party.svc:8111
             # Sandbox straight-through KYC (ADR-0116 §4): auto-approve cases opened from
             # PARTY_CREATED so onboarding activates without an operator. MUST be false in prod.
             - name: OPENBANK_KYC_AUTO_APPROVE
--- a/openbank-infra/gitops/components/observability/prometheus-rules-onboarding.yaml
+++ b/openbank-infra/gitops/components/observability/prometheus-rules-onboarding.yaml
@@ -64,6 +64,39 @@ spec:
           annotations:
             summary: "Parties are being created but no accounts are opening"
             description: "openbank_parties_created_total rose over the last hour while openbank_accounts_created_total did not — new customers are registering but the PARTY_CREATED → account.created pipeline is not producing accounts. This is the onboarding outage signature (party-service → Kafka → account-service). Check the party-events consumer lag, the DLQ (PartyEventDeadLettered), and account-service's dependencies (sanctions-service, product-catalog)."
+        - alert: KycPartiesWithoutCase
+          # Issue #5698. PartyEventConsumer caught every exception, logged it and returned — which
+          # ACKS the Kafka message. A few seconds of kyc-db downtime therefore destroyed the only
+          # copy of a PARTY_CREATED: no KYC case, party stuck in PENDING_KYC forever, accounts stuck
+          # in PENDING_ACTIVATION, no welcome bonus. 10 of 73 sandbox parties were in that state,
+          # the oldest since 2026-06-07, and it was found only because a human noticed an account
+          # did not work.
+          #
+          # Why the two rules above could not catch it. PartyEventDeadLettered needs a DLQ record,
+          # and the whole defect is that the event was acked and never dead-lettered.
+          # OnboardingAccountsNotFollowingParties compares hourly RATES, so it is silent about a
+          # party stranded weeks ago and silent whenever the pipeline is working now — a backlog of
+          # already-stranded parties is invisible to a rate comparison. This is a STOCK, not a rate:
+          # it stays lit until someone actually replays the missing cases.
+          #
+          # Grace period is enforced in the service, not here: openbank_kyc_orphaned_parties only
+          # counts parties older than openbank.kyc.orphan-detection.grace-period (default 1h), so a
+          # party whose PARTY_CREATED is legitimately still in flight is never counted and this does
+          # not fire on every new signup. `for: 30m` on top absorbs a single failed pass.
+          #
+          # Boot-safe by construction: the gauge is seeded to 0 and only ever set by a COMPLETED
+          # pass, so a freshly started pod reads 0 and cannot fire this — deliberately unlike the
+          # Instant.EPOCH seeding that made WorkflowLivenessStale fire 15 minutes after every
+          # deploy (#2239). The converse risk — the reconciliation silently not running, which also
+          # reads 0 — is covered by WorkflowLivenessStale on the kyc-orphaned-party-detection
+          # heartbeat, not by this rule.
+          expr: openbank_kyc_orphaned_parties > 0
+          for: 30m
+          labels:
+            severity: critical
+          annotations:
+            summary: "{{ $value }} party(ies) have no KYC case at all"
+            description: "openbank_kyc_orphaned_parties is {{ $value }} — that many parties, all past the detection grace period, have NO kyc_cases row. Each is a customer permanently stuck: PENDING_KYC party, accounts stuck in PENDING_ACTIVATION, no welcome bonus, and nothing retries because the PARTY_CREATED event was acked and dropped (#5698). Not self-healing — the event is gone, so it needs a replay per party. The stranded party ids are logged by kyc-service under `[orphan-detection]`; openbank_kyc_orphaned_parties_oldest_age_seconds says how long the worst one has been waiting."
         - alert: KafkaConsumerGroupLagStuck
           # General event-bus safety net — the config comment on the Kafka Exporter calls
           # consumer-group lag "the single most useful event-bus health signal" and until now NO
--- a/openbank-infra/gitops/components/party/network-policies.yaml
+++ b/openbank-infra/gitops/components/party/network-policies.yaml
@@ -50,6 +50,9 @@ spec:
               kubernetes.io/metadata.name: documents
         - namespaceSelector:
             matchLabels:
+              kubernetes.io/metadata.name: kyc
+        - namespaceSelector:
+            matchLabels:
               kubernetes.io/metadata.name: onboarding
         - namespaceSelector:
             matchLabels:
--- a/openbank-kyc-service/src/main/kotlin/com/openbank/kyc/application/OrphanedPartyDetector.kt
+++ b/openbank-kyc-service/src/main/kotlin/com/openbank/kyc/application/OrphanedPartyDetector.kt
@@ -1,0 +1,181 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) OpenBank contributors. Licensed under the Apache License, Version 2.0.
+// See LICENSE in the repository root or https://www.apache.org/licenses/LICENSE-2.0 for details.
+
+package com.openbank.kyc.application
+
+import com.openbank.kyc.application.port.out.KycCaseRepository
+import com.openbank.kyc.application.port.out.PartyDirectoryPort
+import com.openbank.kyc.application.port.out.PartySummary
+import jakarta.enterprise.context.ApplicationScoped
+import org.eclipse.microprofile.config.inject.ConfigProperty
+import org.jboss.logging.Logger
+import java.time.Clock
+import java.time.Duration
+import java.time.Instant
+import java.util.UUID
+
+/**
+ * Cross-service reconciliation for the invariant `every party has a KYC case` (issue #5698).
+ *
+ * ### The defect this detects
+ *
+ * `PartyEventConsumer.handleCreated` caught every exception, logged it and returned — which **acks**
+ * the Kafka message. A few seconds of `kyc-db` downtime therefore destroyed the only copy of a
+ * `PARTY_CREATED` event: no case was opened, the party stayed `PENDING_KYC` forever, its accounts
+ * stayed `PENDING_ACTIVATION`, and nothing anywhere went red. Ten of 73 sandbox parties were in that
+ * state, the oldest since 2026-06-07, and it was found only because a human noticed an account did
+ * not work. The consumer fix (#5699) stops *new* losses by rethrowing so the connector
+ * dead-letters; it cannot see a party already stranded, and no retry, DLQ or replay can either —
+ * the event is gone. Only comparing the two services' state finds those, which is what this does.
+ *
+ * ### Why the comparison runs here and reads party-service over REST
+ *
+ * The two sides live in separate services and separate databases, so there is no join to write. The
+ * dependency direction is already settled: kyc-service consumes party events and calls
+ * party-service, never the reverse, so enumerating parties from here adds no new coupling, while
+ * teaching party-service about KYC cases would. This follows the fleet's established reconciliation
+ * shape — a scheduled read-only tie-out that publishes a drift metric and mutates nothing
+ * (`BalanceReconciliationScheduler`, ADR-0039; `AccountReconciliationRepository`, ADR-0026).
+ *
+ * **It deliberately does not remediate.** Opening the missing cases is a separate, human-gated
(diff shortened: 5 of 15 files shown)


JiRaska:
I had built a competing detector for the same #5698 box (#5747, party-service census gauge on `PENDING_KYC` age) and have **closed it in favour of this one** — parallel work, neither saw the other. Three things here are strictly better and one of them was a real flaw in mine:

- **The denominator** (`parties_scanned`). Mine had none, so `== 0` could not be distinguished from a query that reached nothing — the *check passes because it never reached its SUBJECT* trap, inside a control written against exactly that class of silence.
- **`ACTIVE` deliberately not excluded.** A party that reached ACTIVE with no KYC case means activation bypassed the gate. My symptom count keyed on `PENDING_KYC` would have been structurally blind to it, and it is the worse case.
- **Cold-pod t=0 measured rather than reasoned**, and cross-checked against `parties_scanned` + the liveness heartbeat so "no orphans" and "never ran" stay distinguishable.

One gap worth noting for later, not for blocking: a party whose case **exists but is stuck** awaiting a reviewer stays `PENDING_KYC` and is invisible to an orphan query keyed on *no case at all*. Possibly just a normal operational state the ADR-0068 cockpit already shows rather than something to page on — flagging it so the choice is deliberate.

Separately, while working the same sweep I found the rethrow fixes depend on DLQ wiring that mostly is not on `main` yet (SmallRye's default `failure-strategy` is `fail`, which stops the channel) — #5751 supplies it. Worth confirming the ordering wherever a rethrow lands before its DLQ.

JiRaska:
## Review — technically the cleanest PR in this batch; two real findings, one alert-severity decision

Reviewer comment only (no approve, no merge). Read the full 3-dot diff (14 files, 1022 insertions).

### Staleness: not superseded, low-risk drift

`git grep -il "orphan-detection\|orphaned.part" origin/main` -> **zero hits**. Nobody implemented this. Since the merge base, `main` touched `kyc-service.yaml` (6 commits, all auto-deploy image bumps) and `asvs-l3-baseline.txt` (2 commits), and **zero** commits touched `application.yaml`, `KycRepository.kt`, `prometheus-rules-onboarding.yaml` or `party/network-policies.yaml`. 344 behind is textually low-risk here — but see the ASVS note below for why a rebase is still required.

### The four fleet footguns: each is correctly avoided, and tested

I went looking for these specifically and did not find them:

1. **`@Scheduled` Vert.x context.** `OrphanedPartyGauge.kt:154` is `suspend fun refresh()`, `concurrentExecution = SKIP`, cron from config. No `runBlocking` anywhere in `src/main`.
2. **A test that cannot see (1).** `OrphanedPartyDetectionSchedulerIT.kt:57` drives the **real cron** via `@TestProfile(FastScanProfile)` — cron shrunk to `*/2 * * * * ?`, grace period `PT0S` (`:69-81`) — and asserts only on **published gauges** (`:113`, `:164-183`). It never calls `refresh()` directly, which is the whole point. Profile values are literals (`:232-235`), avoiding the two-classloader trap. And `%test` does not disable the scheduler for kyc, so the test is not structurally vacuous — the PR disables the *job* under `%test` instead (`application.yaml:152-160`), which is the right lever.
3. **Gauge at t=0 on a cold pod.** No `Instant.EPOCH` sentinel: all three `AtomicLong` holders seed to `0` (`OrphanedPartyGauge.kt:110-112`), registered in `@PostConstruct` on a `@Startup` bean. Alert is `> 0` with `for: 30m`, so a fresh pod cannot fire it. "Never ran" is covered separately by `registerWorkflowLiveness("kyc-orphaned-party-detection", 1h)` (`:121-124`), registered even when the job is disabled, and a failed pass deliberately does not reset the gauges or record success (`:181-186`).
4. **An alert over a series nobody emits.** `IT:203-229` asserts the **Prometheus-underscored** series names actually appear in `prometheus.scrape()`. That is the assertion that closes this gap, and it is rare to see it written.

Also checked and clean: `openbank.outbox.dispatch-enabled` is already `true` on `main` and this control writes nothing anyway; no new Kafka channel and therefore no dotted-leaf-key trap; no duplicate YAML keys (the new `openbank:` block is under `"%test":`); REST-client config key `party-service` matches `@RegisterRestClient(configKey = "party-service")`; `PartyServiceClient.kt:41-42` is an outbound **interface**, so its `Int` primitives are caller-supplied and compile-checked — not the non-nullable-JAX-RS defect, which applies to inbound resources only. NetworkPolicy `party/network-policies.yaml:51-53` adds `kubernetes.io/metadata.name: kyc` to the ingress block whose ports include 8111 — correctly namespace-keyed.

Threat model: kyc-service is **not** in `rules.yaml: money_path_services` and has no `docs/threat-models/` file at all, so no gate fires and there is nothing to update. Non-blocking, and the new outbound edge is read-only and binds no PII (`PartyServiceClient.kt:44-48` deliberately omits `legalName`/`email`).

### Finding 1 — unbounded `IN` list, and the operator instructions make it worse

`OrphanedPartyDetector.kt:106-113` accumulates `candidates` across **all** pages before a single batched query, and `KycRepository.kt:148-153` puts the whole set into one `WHERE c.partyId IN :ids`. At the configured `max-pages 500 x page-size 100` that is **50,000 UUIDs in one bind list**, under PostgreSQL's 65535 parameter ceiling only by accident.

The sharp edge is that the cap warning at `:75-83` explicitly tells an operator to raise `max-pages` — which is the exact action that breaks it, with a runtime error rather than a slow query. Please chunk the id set (1000/batch is conventional) inside `findPartyIdsWithAnyCase`. The full candidate map being held for the whole scan is a secondary memory concern.

### Finding 2 — this ships a `critical` that is lit from the first deploy

The PR's own text says 10 of 73 sandbox parties are already stranded and that remediation is out of scope. So `KycPartiesWithoutCase` lands as `severity: critical`, firing 30m after the roll, with no ETA on a human replaying those 10.

In an estate where `critical` is already close to the resting state, that is the one failure mode that hides the thing the alert exists to surface. Either ship the replay in the same window, or land it as `warning` with promotion to `critical` tied to the remediation issue. Same objection I raised on #4308, which does the same thing deliberately — worth the two PRs agreeing on a convention.

### Smaller findings

- **No consumer pact for the new kyc -> party edge.** `PartyServiceClient.kt:40-42` is a new cross-service route binding with no `pacts/` entry and no provider replay. I verified the path and payload by hand against `PartyResource.kt:91` / `PartyService.kt:247` on `main`, so it is correct **today** — but this is the finrep -> `/api/v1/ledger/trial-balance` class exactly, where only a provider replay catches later drift. `check-pact-provider-replay.py` derives scope from committed pacts, so the *absence* of a pact fails nothing. If you add one, wire the `@PactFolder` replay in the same PR.
- **`OrphanedPartyGauge.kt:115` — a null `MeterRegistry` silently publishes nothing** while `refresh()` still runs and logs normally. The alert series then never exists, and `WorkflowLivenessStale` cannot help because it is registry-backed too. Test-only in practice; the early return deserves a `log.warn`.
- **`OrphanedPartyDetectorTest.kt` has no case for the `max-pages` cap path** (`OrphanedPartyDetector.kt:75-83`) — the branch that truncates the scan while still publishing a confident `parties_scanned`. That is the "control under-reports and looks clean" shape this PR is otherwise scrupulous about. Arguably a truncated pass should not record liveness success either.
- **`.github/asvs-l3-baseline.txt:57`** adds one line, nothing removed, same shape as the 60+ existing intra-cluster `http://` entries including kyc's own sanctions edge — justified. But it is **line-number-keyed**: `130` is correct against this base and a rebase can silently invalidate it. Re-verify after rebasing.
- `party/network-policies.yaml` is marked `DO NOT EDIT — generated by openbank-infra/scripts/gen-network-policies.py`. The content is consistent with what the generator would derive from the new `PARTY_SERVICE_URL`, but please re-run the generator rather than leaving a hand-edit in a derived file.
- Minor: `PartyDirectoryAdapter.kt:31-45` drops malformed rows, so `scanned` can never reach `result.total` and the loop always runs one extra page to the empty-page stop. Harmless, but the comment at `OrphanedPartyDetector.kt:96-98` claiming `total` "bounds the scan" is then not describing the operative stop condition.

### Verdict

**Needs change** — the `IN`-list chunking is the one I would hold on, plus a decision on alert severity. Everything else is polish. This is careful work and the scheduler IT in particular is the shape more of this fleet needs.

JiRaska:
## Fixed: the `IN` list is bounded now, and the warning no longer advises the thing that breaks it

Pushed `c682a08`.

**What was wrong.** `findPartyIdsWithAnyCase` bound the whole candidate set into **one** statement. Hibernate expands `IN :ids` to one bind parameter per id, and the candidate set is bounded only by the detector's own page cap — `page-size × max-pages`, **100 × 500 = 50,000** by default. PostgreSQL carries a statement's parameter count as an **int16**, so a single statement cannot exceed **65,535** binds. The default shape sat 15,535 under a hard protocol ceiling.

And `OrphanedPartyDetector.kt:75-83` told operators to *"raise `openbank.kyc.orphan-detection.max-pages`"* — the one action that walks it over. At ~660 pages an operator following that log line converts an unchecked tail into a hard failure of the entire reconciler, with nothing in the message or the code saying why.

**Two changes.**

1. `KycRepository` chunks at `ID_BATCH_SIZE` (1,000), so the bind count of any single statement is **constant** regardless of register size; the statement count grows linearly instead. The port KDoc now states `partyIds` is unbounded **by contract**, so the next implementation cannot assume it fits one statement.
2. The cap warning no longer implies the cap is free. It now says the orphan count is a **FLOOR, not a total**, names the pages/parties actually covered, and points at the real remaining cost (one pass holds every candidate in memory) rather than just naming a knob to increase.

### Proof — measured, not read

`KycRepositoryBatchingTest` asserts on the list handed to `setParameter("ids", ...)`, which **is** the statement's bind list, so its size is the parameter count:

- at the configured cap (50,000 ids) → max batch ≤ 1,000, 50 statements, flattened set equals the input (nothing dropped)
- at **13× the cap** (650,000 ids, where the old shape needed ~10× the protocol ceiling) → max batch still exactly 1,000, `< 65,535`
- partial final batch kept — no candidate goes unchecked
- empty input → no statement at all

**Negative control** (a test that cannot detect the absence of what it tests is decoration):

```
idBatches -> listOf(partyIds.toList())   # the pre-fix single-statement shape
  test exit 1 — 4 of 4 FAILED, naming "[largest bind list handed to one statement]"
chunking restored
  test exit 0
```

Worth flagging: **my first negative-control run reported BUILD SUCCESSFUL and proved nothing** — `ktlintFormat` had collapsed the target line onto one line, so the patch silently missed. Caught by asserting the substitution applied. The exit 1 above is from the re-run against the actual text.

### Verification

Gradle serialized, each exit code read from `$?` on a redirected run (never a pipe):

| task | exit |
|---|---|
| `ktlintCheck --rerun-tasks` | 0 |
| `detekt --rerun-tasks` | 0 |
| `test --rerun-tasks` | 0 — **98 tests, 0 failures, 0 errors, 1 SKIPPED** |
| `quarkusBuild` (CDI wiring) | 0 |

The 1 skipped is `KycEventPactBrokerProviderVerificationTest`, the pre-existing `pactbroker.url`-gated broker class — identified from the JUnit XML, not assumed, and it is not a boot failure.

### Not verified

No real-database run at the cap — this is a unit-level measurement of the statement shape, not an executed 50,000-id query against PostgreSQL. It proves the bind count is bounded; it does not measure the wall-clock cost of 50 sequential batches on `kyc-db`, which is worth a look if the sandbox register ever approaches that size.

Issue #5745, opened 2026-08-19
check-event-handler-swallows.py is still blind to 2 shapes, and audit-service's manual ack needs nack (the #5745 A and B sweeps have landed)


JiRaska:
Follow-up to #5698. The sweep's audits were sampled by service, so I enumerated the consumer surface
structurally instead: **59** `@Incoming` mentions under `*/src/main/**.kt`, **44** real annotation
sites across 40 files. No other consumer shape exists (`@Blocking` without `@Incoming` is JAX-RS;
the one `Multi<>` is copilot's SSE *producer*; every `subscribe` is awaited; no direct
`KafkaConsumer`; all 60 `@Observes` are Startup/Shutdown; both `@ConsumeEvent` are a local Vert.x
kill-switch; all ~50 `@Channel` are outgoing emitters).

**44 = 31 genuine · 13 clean.** 21 of the 31 are covered by #5715, #5719, #5725, #5726 and #5737.
The full table with per-site verdicts is in #5737's body. This issue is the remainder.

## A. Ten genuine sites no PR covers

| service | consumer | swallowed call | signal lost |
|---|---|---|---|
| anacredit | `LoanStageEventConsumer` | `projections.applyIfNewer` | IFRS 9 stage / DPD frozen → stale AnaCredit regulatory exposure |
| analytics-sink | `AnalyticsConsumer` | `ClickHouseAnalyticsSink.write` | bronze row lost from the ≥10y log of record. Note the only `DeadLetterSink` binding is `LoggingDeadLetterSink`, a single `log.warnf` — "quarantined" means logged, and the ClickHouse `dead_letter_events` adapter its KDoc promises does not exist |
| campaign | `EnrolmentTriggerConsumer` (2 channels) | `TriggeredEnrolmentService.enrol` — DB + Temporal `startJourney` | party never enrolled; `reset: latest` rules out replay. Worse sub-case: `startJourney` succeeding then `enrolments.save` throwing leaves a running journey with no enrolment row — the service's comment argues that costs "a duplicate start, which the workflow id makes a no-op", an argument that assumes a redelivery this catch-and-ack guarantees never happens |
| campaign | `NotificationOutcomeConsumer` | `sendLog.applyDeliveryOutcome` | send row stays `PENDING` forever — the exact ADR-0239 D3 / #3663 defect this consumer was built to fix |
[...]

PR #5751, opened 2026-08-19
fix(fleet): wire the dead-letter queue the rethrow depends on (#5745)

JiRaska:
Closes the `B` section of #5745.

## The problem

#5715, #5719, #5725, #5726 and #5737 convert ~21 event consumers from swallow-and-ack to rethrow,
each on the stated basis that "the connector dead-letters". **SmallRye's default `failure-strategy`
is `fail`, which stops the channel.**

Enumerated structurally from every service's top-level `mp.messaging.incoming` block (profile
overrides under `"%test"` / `"%dev"` excluded — they are the same channel, not a second one):

| | count |
|---|---|
| incoming channels in the monorepo | **44** |
| configured a DLQ before this PR | **4** |
| wired by this PR | **35** |
| baselined with a stated reason | **5** |

So for 40 of 44 channels the sweep converted silent data loss into a halted consumer, while the
consumers' KDocs asserted the opposite. Nothing errors at boot; nothing can.

## What each wired channel gets

All four parts, mirroring #5737's shape for `party-events-in`:

1. `failure-strategy: dead-letter-queue`
2. an explicit per-service topic, `openbank.dlq.<service>.<channel>`
3. a `KafkaTopic` CR — this cluster does not auto-create topics
4. a KafkaUser `Write` ACL — without it the DLQ send itself is denied and the consumer wedges on
   the very failure the DLQ was added to park, which is strictly worse than `fail`

## The two traps, and how each was handled

**The dotted-key trap.** `dead-letter-queue.topic:` written as a leaf key never reaches the
connector: SmallRye Config's YAML source quotes any leaf key containing a literal dot, so it
registers as `...party-events-in."dead-letter-queue.topic"` and the connector's plain
`getOptionalValue` never finds it. Nothing errors and the default silently applies (#686).

Measured against the real `io.smallrye.config.source.yaml.YamlConfigSource`, both spellings in one
document:

```
[mp.messaging.incoming.party-events-in."dead-letter-queue.topic"] = DOTTED-FORM
[mp.messaging.incoming.party-events-in.dead-letter-queue.topic]   = NESTED-FORM
```

The **nested** form produces exactly the property `KafkaConnectorIncomingConfiguration` reads — the
same idiom the fleet already uses for `value: deserializer:`. So `application.yaml` is the right
home for it and no new msg-override ConfigMap is needed (a ConfigMap would also leave local dev and
every test on the wrong value, which is the state `openbank-transaction-service` was in — its
money-path `payment-scheme-accepted` channel carried the dotted form, inert everywhere except the
deployed pod, where the ConfigMap supplied it. Corrected to nested here; the ConfigMap stays,
unchanged, and still carries the same value).

All 35 patched channels were then replayed through that same parser: **35 resolved, 0 failed**, with
`failure-strategy` set and the quoted variant absent in every one.

**The implicit-name collision.** SmallRye's default `dead-letter-topic-<channel>` is derived from
the channel name alone, and channel names repeat — `party-events-in` is declared by eight services.
Every topic here is named explicitly and per-service, and the gate enforces fleet-uniqueness.

This is not hypothetical. The gate's first run found it live: **account-service and
card-issuance-service both consume `delegation-events-in`, so they already share
`dead-letter-topic-delegation-events-in` today** — two services' dead letters in one topic,
misattributed by anything reading it.

A guard is only proven by what it prevents, so both halves were falsified:

- `DeadLetterQueueConfigResolutionTest` (openbank-aml-service) resolves `failure-strategy` and
  `dead-letter-queue.topic` through the very class Quarkus uses to read `application.yaml`, and
  carries the dotted spelling as an explicit negative control. **Falsified:** reverting aml's
  channel to the dotted one-liner — which looks correct and parses fine — fails three of its four
  assertions.
- `check-incoming-dlq-wiring.py`, gate `incoming-dlq-wiring` (enforced, `min_subjects: 30` against
  44 channels), ratchets all four requirements fleet-wide, plus uniqueness. **Falsified twice:**
  deleting one channel's config block fires the config finding; renaming one ACL's topic fires the
  ACL finding. Neither masks the other, and the 5 self-test cases include the dotted form and a
  `"%test"` profile block that must not count as a second channel.

## Not wired — recorded in the gate baseline, not left invisible

The gate fails on a stale baseline entry in either direction, so this debt cannot quietly become
permanent.

- **notification-service** (3 channels) — its `application.yaml` is owned by in-flight #5737, which
  wires `party-events-in`. Deliberately untouched to avoid racing that PR.
- **tax-reporting-service** `withholding-remitted-in` — no KafkaUser exists in gitops for it, so no
  `Write` ACL can be granted; a DLQ without one would wedge on the send. Needs its Kafka identity
  first.
- **The 4 pre-#5745 channels** (account x2, card-issuance, fraud) on implicit names — naming them
  explicitly is a rename of a live topic: it strands whatever is already parked in the old one and
  blinds the existing `AccountPartyEventDeadLettered` alert, which reads
  `dead-letter-topic-party-events-in` by name. That is an operational migration, not a config edit,
  and it is where the `delegation-events-in` collision above gets fixed.

## Coordination

- **No file is shared with any in-flight sweep PR.** #5715/#5719/#5725/#5726 touch consumer `.kt`
  files only; #5737 additionally touches `openbank-notification-service/src/main/resources/application.yaml`,
  which this PR does not open.
- `.github/gates/gates.yaml` is also touched by #5670, #5688 and #5737. This PR **appends** its gate
  at EOF while those insert mid-file, so the regions are disjoint — but that file is exactly the
  shape where git calls a merge clean and keeps one of two neighbouring additions, so the entry
  count is worth re-counting after any merge (157 gates after this change).
- The 35 `KafkaTopic` CRs are one new file, `components/kafka/kafka-dlq-topics.yaml`, rather than
  appended to each component's `kafka-topic.yaml` — for that same reason, and because it makes the
  DLQ inventory countable in one place. That directory is synced by the existing `kafka` ArgoCD
  Application (directory recursion, `ServerSideApply`, prune deliberately off).

No AWS, cluster or live Kafka was touched.

Refs #5745, #5698

### DIFF
--- a/.github/gates/gates.yaml
+++ b/.github/gates/gates.yaml
@@ -3726,3 +3726,25 @@ gates:
     selftest_expect: pass
     run: |
       python3 .github/scripts/check-gate-selftest-declaration.py --enforce
+
+  # A rethrow only helps if the connector PARKS the record. SmallRye's default failure-strategy is
+  # `fail`, which STOPS the channel — and before #5745 only 4 of the fleet's 44 incoming channels
+  # configured a DLQ, so most of the #5698 sweep traded silent data loss for a wedged consumer while
+  # the consumers' KDocs asserted the opposite. Three ways to write this config look right and do
+  # nothing: the dotted leaf key `dead-letter-queue.topic:` (SmallRye quotes it, the connector never
+  # reads it, #686), an implicit topic name derived from the channel alone (eight services declare
+  # `party-events-in`, so they would share one DLQ — observed live today between account-service and
+  # card-issuance-service on `delegation-events-in`), and a topic with no KafkaUser Write ACL (the
+  # DLQ send is denied and the consumer wedges on the failure the DLQ was added to park). None of
+  # the three errors at boot, which is why this is a gate and not a review note.
+  - id: incoming-dlq-wiring
+    name: "every incoming channel dead-letters into an explicit topic it may write (#5745, enforced)"
+    group: gitops
+    mode: enforced
+    rationale: "The #5698 sweep's rethrows depend on a DLQ that existed on 4 of 44 channels; each of the three ways to mis-wire one is silent at boot and invisible until a consumer wedges in production."
+    review_after: "2027-02-19"
+    selftest: python3 .github/scripts/check-incoming-dlq-wiring.py --self-test
+    selftest_expect: pass
+    min_subjects: 30   # 45 incoming channels today (2026-08-21, incl. audit-service agent-audit-events-in)
+    run: |
+      python3 .github/scripts/check-incoming-dlq-wiring.py --enforce
--- a/.github/scripts/check-incoming-dlq-wiring.py
+++ b/.github/scripts/check-incoming-dlq-wiring.py
@@ -1,0 +1,359 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) OpenBank contributors. Licensed under the Apache License, Version 2.0.
+"""Every `mp.messaging.incoming` channel dead-letters, explicitly, into a topic it may write.
+
+WHY THIS GATE EXISTS
+    #5698 converted ~21 event consumers from swallow-and-ack to rethrow, on the stated basis that
+    "the connector dead-letters". SmallRye's DEFAULT `failure-strategy` is `fail`, which STOPS the
+    channel. Only four of the fleet's 44 incoming channels configured a DLQ, so for most of the
+    sweep the rethrow traded silent data loss for a WEDGED consumer — the exact outcome the swallow
+    comments it replaced were written to avoid. Nothing errored, and no test could see it: the
+    consumers' own KDocs asserted a behaviour their configuration did not provide (#5745).
+
+WHAT IT REQUIRES, per incoming channel
+    1. `failure-strategy: dead-letter-queue`.
+    2. An EXPLICIT topic. SmallRye's implicit default is `dead-letter-topic-<channel>`, derived from
+       the channel name alone — and channel names repeat (`party-events-in` is declared by eight
+       services), so left implicit those eight share one topic and every alert over it misattributes.
+    3. In `application.yaml`, the topic written in NESTED form, never as the dotted leaf key
+       `dead-letter-queue.topic:`.
+       SmallRye Config's YAML source quotes any leaf key containing a literal dot, registering it as
+       ...`"dead-letter-queue.topic"`, which the connector's plain getOptionalValue never reads.
+       Nothing errors; the default silently applies. Same footgun as `group.id` (#686).
+       `openbank-aml-service`'s DeadLetterQueueConfigResolutionTest proves both directions against
+       the real SmallRye YAML source; this gate is the fleet-wide ratchet over that proof.
+    4. Fleet-unique topic names — the whole point of (2).
+    5. A `KafkaTopic` CR: this cluster does not auto-create topics.
+    6. A KafkaUser `Write` grant on it. Without the ACL the DLQ send itself fails and the consumer
+       wedges on the very failure the DLQ was added to park — strictly worse than `fail`.
+
+WHERE IT LOOKS
+    `application.yaml` is not the only config source. A `<svc>-msg-override` ConfigMap under
+    `openbank-infra/gitops/components/**` carries `override.properties` at `config_ordinal=500`,
+    which OUTRANKS the baked `application.yaml` (~254) and is loaded via QUARKUS_CONFIG_LOCATIONS.
+    That is where a `dead-letter-queue.topic` legitimately lives, because in a `.properties` source
+    the dots are read correctly while SmallRye's YAML source would quote the leaf key and the
+    connector would never read it (#686). notification-service's `party-events-in` is wired exactly
+    that way (#5737), and fraud-service's `transaction-signal` has been since before #5745.
+
+    A gate that read only `application.yaml` would report both as "DLQ topic left implicit" — a
+    FALSE statement about a channel that is correctly wired, and one that would go red on `main`
+    rather than on a PR. So the properties source is parsed too, and the effective value is the
+    override's where both exist.
+
+USAGE
+    check-incoming-dlq-wiring.py [--enforce] [--self-test]
+"""
+from __future__ import annotations
+
+import re
+import sys
+from pathlib import Path
+
+sys.path.insert(0, str(Path(__file__).resolve().parent))
+import gatelib  # noqa: E402
+
+REPO = Path(__file__).resolve().parents[2]
+GITOPS = REPO / "openbank-infra/gitops/components"
+
+# Channels deliberately not yet wired. An entry needs a REASON, and the gate fails on a stale one
+# in either direction — a baselined channel that becomes wired must leave this list, so the debt
+# cannot quietly become permanent.
+KNOWN_UNWIRED = {
+    # #5737 (merged) wired notification-service's `party-events-in` — and did it through the
+    # msg-override ConfigMap, not application.yaml, because the key is a dotted leaf (#686). It is
+    # therefore NOT listed here: it is wired, and recording it as debt would be a false statement
+    # in the baseline. The gate reads the ConfigMap and sees it. Its two SIBLING channels are a
+    # different matter — #5737 did not touch them, and they carry no `failure-strategy` at all.
+    ("openbank-notification-service", "notification-events-in"): "#5737 wired only party-events-in; this channel has no failure-strategy yet (#5745)",
+    ("openbank-notification-service", "delegation-events-in"): "#5737 wired only party-events-in; this channel has no failure-strategy yet (#5745)",
+    ("openbank-tax-reporting-service", "withholding-remitted-in"): "no KafkaUser in gitops — a Write ACL cannot be granted, so a DLQ would wedge on the send (#5745)",
+    # The four channels that had a DLQ BEFORE #5745, all on SmallRye's implicit
+    # `dead-letter-topic-<channel>` name. Naming them explicitly is a RENAME of a live topic: it
+    # strands whatever is already parked in the old one and silently blinds the existing
+    # AccountPartyEventDeadLettered alert, which reads `dead-letter-topic-party-events-in` by name.
+    # That is an operational change with a migration, not a config edit, so it is recorded here
+    # rather than done in passing. It is not cosmetic: account-service and card-issuance-service
+    # BOTH consume `delegation-events-in` and therefore already share
+    # `dead-letter-topic-delegation-events-in` today — two services' dead letters in one topic,
+    # misattributed by anything that reads it. That is the collision this gate exists to prevent,
+    # observed live (#5745).
+    ("openbank-account-service", "party-events-in"): "pre-#5745 implicit DLQ name; renaming is a live-topic migration",
+    ("openbank-account-service", "delegation-events-in"): "pre-#5745 implicit DLQ name; ALREADY COLLIDES with card-issuance-service",
+    ("openbank-card-issuance-service", "delegation-events-in"): "pre-#5745 implicit DLQ name; ALREADY COLLIDES with account-service",
+    # NOT the implicit name, and not a rename: fraud-service already sets an EXPLICIT topic
+    # (`openbank.transactions.transaction.initiated.fraud.dlq`) in
+    # components/fraud-service/fraud-service-msg-override.yaml, with a matching [Write, Describe]
+    # ACL in kafka-fraud-mtls.yaml. What it lacks is requirement 5 — no `KafkaTopic` CR exists for
+    # that name anywhere. One CR, no migration; load-bearing because #5715 makes
+    # TransactionSignalConsumer rethrow on this very channel.
+    ("openbank-fraud-service", "transaction-signal"): "explicit topic set in fraud-service-msg-override.yaml; the gap is a missing KafkaTopic CR, not a rename (#5745)",
+}
+
+
+def channels(text: str) -> list[tuple[str, dict[str, str]]]:
+    """Top-level mp.messaging.incoming channels and their flattened leaf properties.
+
+    Profile blocks (`"%test":`, `"%dev":`) are excluded on purpose: they sit at a deeper
+    indentation and describe local runs, not the deployed channel.
+    """
+    lines = text.split("\n")
+    out = []
+    for i, line in enumerate(lines):
+        if line != "    incoming:":
+            continue
+        j = i + 1
+        while j < len(lines):
+            cur = lines[j]
+            if cur.strip() == "" or cur.lstrip().startswith("#"):
+                j += 1
+                continue
+            if len(cur) - len(cur.lstrip()) <= 4:
+                break
+            m = re.match(r"^      ([\w.-]+):\s*$", cur)
+            if m:
+                body, k = [], j + 1
+                while k < len(lines):
+                    nxt = lines[k]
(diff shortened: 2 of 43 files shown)


JiRaska:
Diagnosed the `gates (data)` failure here — **it is the gate that is wrong, not this PR.**

`check-kafka-acl-coverage` infers the operation from the channel key (`incoming` ⇒ Read, `outgoing` ⇒ Write) and collects any `topic` key beneath it. This PR's `dead-letter-queue.topic` is nested **under the incoming channel**, so all ~40 new DLQ topics come out as needing **Read**. A DLQ is produced to — the connector parks the failed record by PRODUCING to it, and nothing consumes it by design.

Ran the gate's own `required()` against this branch rather than reasoning about it:

```
openbank.dlq.balance.balance-init-in    -> gate requires Read
openbank.dlq.balance.ledger-events-in   -> gate requires Read
openbank.dlq.aml.party-events-in        -> gate requires Read
```

Ground truth is already on `main`: account-service's KafkaUser grants `dead-letter-topic-party-events-in` `[Write, Describe]`, commented *"Without Write here, the DLQ send itself fails and the consumer [retries]"*.

**The trap worth naming:** satisfying this gate as it currently asks — by granting Read — would turn it green while every DLQ send stays denied by the broker (`allow.everyone.if.no.acl.found=false`). That is a green check over a dead-letter queue that cannot be written, which is *worse* than having none: the rethrow it exists to serve then stops the channel instead of parking the record.

Fix opened as **#5843** (one file, self-test falsified both ways). With it,  on main is clean at 107 pairs / 42 KafkaUsers — so this PR's ACLs then need **Write** on those topics, which I expect is what you intended anyway.

Separately, `gates (gitops)` here fails on your own new `#5745` rule ("every incoming channel dead-letters into an explicit topic it may write") — likely the same root cause, since "may write" is judged from the same ACL data. Worth re-checking after #5843.

JiRaska:
Correction to my comment above: one phrase was eaten before it published — backticks inside a
double-quoted shell string are command substitution, so `--enforce` was executed rather than
written (the trap this repo's own guidance names, and I walked into it by not using `--body-file`).

The sentence should read: **with #5843, `check-kafka-acl-coverage.py --enforce` on `main` is clean
at 107 (topic, operation) pairs across 42 KafkaUsers.**

Nothing else in that comment changed.

JiRaska:
## Review — the 36 wired channels are complete; two findings on the gate and one on the debt list

I verified all four axes by parsing the branch's `application.yaml` files and every `KafkaUser`/`KafkaTopic` CR under `openbank-infra/gitops/components` as YAML rather than by grep, and checked ACL *identity* (which KafkaUser holds the grant), not just presence.

**All 36 channels this PR wires pass all four axes:** `failure-strategy: dead-letter-queue` on 36/36; 36 explicit topics in `openbank.dlq.<service>.<channel>` form with **zero implicit names and zero collisions** (36 topics, 36 distinct names); 35 new `KafkaTopic` CRs in `openbank-infra/gitops/components/kafka/kafka-dlq-topics.yaml` plus the pre-existing `payment.scheme-accepted.dlq` CR at `openbank-infra/gitops/components/payments/kafka-topic.yaml:178` = 36/36; and 36/36 `[Write, Describe]` ACLs, each on the correct consuming service's KafkaUser.

Two things I specifically went looking for and did **not** find:
- **No dotted leaf keys remain** (`^\s*dead-letter-queue\.[a-z]` → 0 hits). The PR in fact *removes* the one live occurrence, on a money-path channel: `openbank-transaction-service/src/main/resources/application.yaml:127`. Correctly left alone is `openbank-infra/gitops/components/payments/transaction-service-msg-override.yaml:48`, where the same dotted key is a `.properties` source and the dots are right.
- **No `auto.offset.reset` added or changed anywhere in this diff.** No mass-replay risk.

### Finding 1 (gate) — the KafkaTopic-CR check cannot fail independently, so requirement 5 is decorative

`.github/scripts/check-incoming-dlq-wiring.py:171`:

```python
if f"name: {topic}" not in gitops:
```

`gitops` is *all* gitops YAML concatenated, and the test is a bare substring match not scoped to `kind: KafkaTopic`. Every ACL entry this PR adds is itself `          name: openbank.dlq.<...>`, which contains that substring. So **deleting every one of the 35 new KafkaTopic CRs while keeping the ACLs leaves this gate green.** Given the check's own error string says the consumer wedges without a CR, that is the one branch that most needs to be real. Scope it to a `kind: KafkaTopic` block.

Related, weaker: the ACL check at `:174` is a 200-character-window regex over the same concatenated corpus and does not verify *which* KafkaUser holds the Write grant. The manifests here are correct — I checked identity separately — but the gate cannot tell.

### Finding 2 (gate) — the two weakest checks are the two with no negative control

`self_test()` (`:198-215`) does not call `findings()`. It re-implements the condition inline as `strategy != … or not topic`, so it exercises only `channels()`/`props()`. **No self-test case covers the CR branch, the ACL branch, or the collision branch.** Per this repo's own rule, a guard is proven by what it prevents: please add cases that feed a channel with a missing CR, a missing Write ACL, and a duplicate topic, and assert each produces a finding. Finding 1 is exactly the kind of defect such a case would have caught.

Good, and worth keeping: `min_subjects: 30` in `.github/gates/gates.yaml`, and `KNOWN_UNWIRED` failing on a stale entry in **both** directions (`:145-146`) so the debt cannot quietly become permanent. Minor: `channels()` matches literal `    incoming:` / 6-space channel keys, so a re-indented file goes silently invisible.

`check-kafka-acl-coverage.py` is the better of the two — classifying a DLQ topic as Write-only rather than inheriting `Read` from the enclosing `incoming` key (`:113-121`), with a fixture self-test asserting all three directions *plus* the explicit negative `("…dlq…", "Read") not in got`. That is the pattern the other script should follow.

### Finding 3 — the `KNOWN_UNWIRED` reason for fraud-service is factually wrong, and it hides a cheap fix

`.github/scripts/check-incoming-dlq-wiring.py:67`:

```
("openbank-fraud-service", "transaction-signal"): "pre-#5745 implicit DLQ name; renaming is a live-topic migration",
```

fraud-service is **not** on the implicit name. It already has an explicit topic — `openbank-infra/gitops/components/fraud-service/fraud-service-msg-override.yaml:47` sets `…transaction-signal.dead-letter-queue.topic=openbank.transactions.transaction.initiated.fraud.dlq` — and a matching `[Write, Describe]` ACL at `openbank-infra/gitops/components/fraud-service/kafka-fraud-mtls.yaml:38-42`. The gate classes it as implicit only because it reads `application.yaml` and cannot see the ConfigMap.

So the stated blocker ("renaming is a live-topic migration") does not apply to fraud at all. What fraud actually lacks is the **KafkaTopic CR**: `openbank-infra/gitops/components/fraud-service/kafka-topics.yaml` contains only `openbank.fraud.hold.changed`, and that topic name appears in no `KafkaTopic` anywhere on this branch. That is one CR, unblocked, no migration — and it is load-bearing right now, because **#5715 makes `TransactionSignalConsumer` rethrow on that very channel**. Please either add the CR here or say plainly in the entry that fraud's gap is a missing CR rather than a rename.

### Finding 4 — the collision this gate exists to prevent is live, and has no filing

`account-service` and `card-issuance-service` both consume `delegation-events-in` and therefore already share `dead-letter-topic-delegation-events-in` today — two services' dead letters in one topic. The comment at `:57-64` describes this accurately and the migration reasoning is sound (renaming strands parked records and blinds the `AccountPartyEventDeadLettered` alert, which reads `dead-letter-topic-party-events-in` by name at `openbank-infra/gitops/components/observability/prometheus-rules-onboarding.yaml:37`). But the entries cite only #5745, and I could not find a follow-up issue for the rename migration. A described-but-unfiled follow-up is not a filing — please open one and cite it in each entry.

Also worth flagging: `openbank-tax-reporting-service` `withholding-remitted-in` has **no KafkaUser in gitops at all**, so it cannot be wired even in principle. That is a bigger gap than the entry's one line suggests.

### `DeadLetterQueueConfigResolutionTest.kt` — a real negative control

It drives the actual `io.smallrye.config.source.yaml.YamlConfigSource` in both directions: `:88` asserts the dotted spelling resolves to `null`, `:90` asserts it lands under the quoted property name, `:105` asserts the nested form resolves, and `:69-70` runs the same probe against the **real** `application.yaml`. Rewriting aml's file to the dotted form fails it. Not vacuous. Its limit is scope — one channel, one service; the fleet-wide ratchet is the Python gate, whose `props()` (`:103-122`) re-implements SmallRye's quoting rule with a regex instead of calling SmallRye, so the two can drift.

JiRaska:
**Cross-PR naming collision — whichever of #5737 and #5751 merges second needs a rename.**

- #5751 adds the enforced `incoming-dlq-wiring` gate and establishes the convention
  `openbank.dlq.<service>.<channel>` across 35 topics — and omits notification-service.
- #5737 wires notification-service's DLQ as `openbank.party.events.notification.dlq`.

Both are green in isolation, and `git merge-tree` reports no textual conflict, because they touch
different files — the disagreement is semantic. Merged in either order the result is one
notification DLQ whose name does not match the convention the gate now enforces, so the second merge
turns the gate red on `main` rather than on a PR.

Cheapest resolution: rename #5737's topic to `openbank.dlq.notification.party-events` (or whatever
`incoming-dlq-wiring` derives for that channel) **before** either lands, and check the KafkaTopic,
the Write ACL and the `dead-letter-queue.topic` msg-override line together — they are three separate
files and renaming only the topic leaves the ACL pointing at nothing, which fails closed at
publish time rather than at deploy time.

Same shape as the repo's note that two PRs each green alone can break `main` together; `merge-tree`
clean is not the test here.

JiRaska:
**Correcting my earlier comment on this thread — I named the wrong mechanism.**

I wrote that #5737 and #5751 collide on the DLQ *topic name*. They do not. I read
`check-incoming-dlq-wiring.py` on #5751's head (`FETCH_HEAD:.github/scripts/check-incoming-dlq-wiring.py`)
and it enforces no naming convention at all — it requires `failure-strategy: dead-letter-queue`, an
explicit topic, global uniqueness of that topic, a `KafkaTopic` CR, and a `KafkaUser` Write ACL.
`openbank.party.events.notification.dlq` satisfies every one of those.

**The real collision is that the gate cannot see where #5737 puts the topic.** The gate reads
leaf properties out of each service's `application.yaml`:

```python
topic = p.get("dead-letter-queue.topic")
...
if not topic:
    out.append(f"{svc} :: {ch}: DLQ topic left implicit — SmallRye derives "
               f"`dead-letter-topic-{ch}` from the channel name alone, which collides "
               f"across services sharing it")
```

#5737 deliberately does **not** put the topic there. Its own comment explains why: the key is the
dotted leaf `dead-letter-queue.topic`, which SmallRye's YAML source registers quoted and never
reads (#686) — so the name is set in the gitops `msg-override` ConfigMap instead. That is the
*correct* way to set it, and it is invisible to this gate's probe.

Consequence, concretely: `openbank-notification-service` is **not** in the gate's `KNOWN_UNWIRED`
baseline, so once #5737 lands, that channel has `failure-strategy: dead-letter-queue` and no topic
the gate can find, and the gate reports it as "left implicit" — **red on `main`**, about a channel
that is wired correctly.

So the fix belongs in #5751, not #5737: teach the gate to also read the service's `msg-override`
ConfigMap (where `override.properties` carries `config_ordinal=500`), or add an explicit exemption
that names the ConfigMap it checked instead. An exemption that just baselines notification-service
would be the wrong shape — it would record "unwired" for a channel that is wired.

This is the general trap the repo already knows in other forms: a gate whose probe cannot observe
the correct idiom reports the correct idiom as a defect. Same family as grepping `src/test` for the
word "contract", and as the event-payload keys that only exist at runtime on a serialised data class.

JiRaska:
## Rebuilt on current `main`, and the #5737 problem from the comment above is fixed here

### The conflict
`kafka-audit-mtls.yaml` only — `main` added an `audit-service-agent-provenance` group ACL where this
branch added the DLQ topic ACL. Both wanted, resolved as a union. The `audit-service` KafkaUser now
carries 25 ACLs, ending `group audit-service` / `group audit-service-agent-provenance` /
`topic openbank.dlq.audit.agent-audit-events-in [Write, Describe]` /
`topic openbank.dlq.audit.audit-events-in [Write, Describe]`.

### notification-service is NOT baselined as unwired, because it is wired

#5737 merged and wired `party-events-in` through the **msg-override ConfigMap**, not
`application.yaml`, because `dead-letter-queue.topic` is a dotted leaf key (#686). A gate that reads
only `application.yaml` would report it as "DLQ topic left implicit" — a false statement about a
correctly wired channel, and one that would go red on `main` rather than on a PR.

So the gate now **parses `<svc>-msg-override` ConfigMaps too** and uses the override's value where
both exist (`config_ordinal=500` outranks `application.yaml`'s ~254). `party-events-in` left
`KNOWN_UNWIRED` entirely. Its two siblings stayed, with corrected reasons: #5737 did not touch them
and they carry no `failure-strategy` at all.

Negative control, run rather than reasoned about — deleting the one ConfigMap line:

```
openbank-notification-service :: party-events-in: DLQ topic left implicit — SmallRye derives
`dead-letter-topic-party-events-in` from the channel name alone, which collides across services
```

### Also fixed, from the review above

**Finding 1 (the CR check could not fail independently).** `kafka_topic_cr()` now requires a
`kind: KafkaTopic` document that declares the name, instead of a substring match over all gitops
YAML concatenated. Negative control — deleting one KafkaTopic CR while keeping its ACL:

```
openbank-audit-service :: audit-events-in: no KafkaTopic CR for openbank.dlq.audit.audit-events-in
```

Under the old substring test that case was green, because the ACL entry is itself a line reading
`name: openbank.dlq.audit.audit-events-in`.

**Finding 2 (no negative control on the weak branches).** `--self-test` gained four cases that call
the real helpers: a Write ACL alone is not a CR, a real KafkaTopic document is, and both live
ConfigMap-sourced topics resolve. All nine pass.

**Finding 3 (the fraud-service reason was factually wrong).** Corrected in place: fraud already has
an explicit topic in its msg-override ConfigMap and a matching Write ACL — its actual gap is a
missing KafkaTopic CR, not a rename. The staleness check was also restructured so every requirement
is evaluated for every channel and the baseline only *suppresses*; an entry is now stale only when a
channel passes **all six** requirements, not the two readable from `application.yaml`.

### One channel `main` gained since this branch was cut

`openbank-audit-service :: agent-audit-events-in` (the D5 provenance stream) had no DLQ and would
have gone red. Wired the same way as the other 36 — `failure-strategy` + nested explicit topic,
KafkaTopic CR, `[Write, Describe]` ACL. It needs it more than most: `AgentAuditConsumer` acks only
after append-only persistence succeeds, so it rethrows by design.

### Verification

```
check-incoming-dlq-wiring.py --enforce  -> exit 0
SUBJECTS=45  # incoming Kafka channels
clean: 45 incoming channels, every one dead-letters into an explicit topic it may write

check-kafka-acl-coverage.py --enforce   -> exit 0
144 (topic, operation) pair(s) checked across 42 KafkaUser(s), 0 known gap(s) — clean.

run-gates.py --group gitops --group data -> PASS [gitops] incoming-dlq-wiring
```

`run-gates.py` also reported 3 failures + 1 unfalsified (`litellm-config-revision`,
`loki-rule-load-test`, `db-migration-gate`, `schema-compat-gate`). Negative control: the same four
fail identically on a clean `origin/main` worktree, so they are pre-existing/environmental, not this
branch.

### Left for a human

**The naming collision from my earlier comment is still open.** #5737's topic is
`openbank.party.events.notification.dlq`, not the `openbank.dlq.<service>.<channel>` convention the
other 36 follow. The gate does not enforce the naming convention — only explicitness and fleet
uniqueness — so it is green, and renaming a topic that is now live is a migration, not a config
edit. Recorded here rather than done in passing.

PR #5758, opened 2026-08-19
fix(transaction): make TransactionSagaStuck able to fire — instrument the gauge it needs

JiRaska:
## What

Makes `TransactionSagaStuck` — `severity: critical`, money-path — actually able to fire. It never has been.

The rule was written against `openbank_transaction_sagas_stuck_total`, a metric **nothing has ever emitted** (#5733).

## Why nothing caught it

A dead alert is invisible from every angle you would normally check: it loads, `promtool` accepts the PromQL, Alertmanager routes it, and Prometheus reports it `inactive` forever — **which is exactly what a correctly-quiet critical alert also looks like**.

Confirmed two independent ways:
- The live sandbox's Prometheus holds 59 `openbank_*` metric names; that is not one of them.
- A comment-stripped scan of every `openbank-*/src/main` finds no instrumentation for it in any of this repo's three metric-naming idioms.

## The fix

`TransactionStrandedGauge`, deliberately the same shape as domestic-payment's `DomesticPaymentStrandedGauge` (#3273), which exists for the identical failure:

> a wedged saga answered 2xx, emitted no error span, took normal latency and sits on a healthy pod — so every rule measuring whether the service **answers** stays green while work it accepted stops **progressing**. Age is the only layer at which the two differ.

```
openbank_transactions_non_terminal{service,status}
openbank_transactions_non_terminal_oldest_age_seconds{service,status}
```

`PENDING` and `PROCESSING` only — a terminal state's age only grows and would alert forever. Per-status rather than one aggregate because the thresholds genuinely differ: `PROCESSING` is mid-saga and legitimately takes seconds to low minutes, while `PENDING` means the saga was accepted and never started. One number could only carry the looser of the two.

The rewritten alert reads `PENDING > 300s` — the "stuck >5m" the old comment always claimed, now expressed against a series that exists.

## Design note: its own port

The two read-only queries live on a new `StrandedSagaQueryPort`, not as two more methods on `TransactionRepository`. That interface is the aggregate's persistence contract on a money-path service — every method participates in the transactional-outbox invariant — and a diagnostic read does not belong in the same review surface as the write path. It also keeps `PanacheTransactionRepository` under detekt's `TooManyFunctions` threshold, which fires **at** 11, not above it.

## Verified

- `TransactionStrandedGaugeTest` **3/3**. The empty-state test is falsifying and earned it: the first version handed the cleared data to a *second* gauge object while asserting on a registry gauge still bound to the **first** instance's `AtomicLong` — it would have agreed with a production bug it could not see. Rewritten to one instance over a mutable stub, the only arrangement that exercises the cached holder the registry actually reads.
- `ktlint`, `detekt`, `quarkusBuild` (CDI wiring) green.
- `check-scheduler-liveness`: **51/51** with liveness, the new gauge included. `check-no-runblocking-in-scheduled` clean.
- **Decisive check, by effect rather than assertion**: ran #5739's new `check-alert-metric-emitted` gate against this branch with `TransactionSagaStuck` **removed from its baseline** → clean, **214** emitted metric names (up from 212) and **3** baselined instead of 4. The gate that flagged this alert as dead no longer can.

## Merge order with #5739

#5739 baselines four dead alerts including this one. Merging **this** PR first is the clean order: #5739's branch has already been updated to drop the `TransactionSagaStuck` entry, so its baseline lands at three. If #5739 merges first instead, its baseline correctly still covers this alert and nothing breaks — the entry simply becomes stale on the next run and gets reported, which is the ratchet working as designed.

## Money path

`openbank-transaction-service` is in `rules.yaml: money_path_services` — 2 approvals. The diff adds only a read-only gauge and its port: no write path, no aggregate, no outbox touched.

## Still open in #5733

The other three (`ClearingSettlementWindowMissed`, `SwiftMtMessageProcessingStalled`, `LendingRepaymentProcessingStalled`) are all `warning`, all event-driven rather than scheduled, and all written as `increase(...) == 0` — a shape that is blind to absence even once the metric exists. They need a decision about whether that alert shape should exist at all before instrumenting three more counters across three money-path services; documented in #5733 rather than rushed here.

Refs #5733

### DIFF
--- a/openbank-infra/gitops/components/payments/prometheus-rules.yaml
+++ b/openbank-infra/gitops/components/payments/prometheus-rules.yaml
@@ -35,15 +35,32 @@ spec:
             summary: "transaction outbox backlog stuck"
             description: "transaction-service outbox has >100 unrelayed rows for >15m — payment events are not being dispatched."
         - alert: TransactionSagaStuck
-          # A saga stuck >5m = payment is in a terminal-unknown state.
-          expr: openbank_transaction_sagas_stuck_total > 0
+          # A saga stuck past its expected completion = payment in a terminal-unknown state.
+          #
+          # This rule spent its whole life unable to fire (#5733): it was written against
+          # `openbank_transaction_sagas_stuck_total`, which nothing has ever emitted. It loaded,
+          # promtool accepted it, Alertmanager routed it, and Prometheus reported it `inactive`
+          # forever -- indistinguishable from a correctly-quiet critical alert. It now reads the
+          # gauge TransactionStrandedGauge actually publishes.
+          #
+          # Age, not count: a wedged saga answered 2xx, emitted no error span, took normal latency
+          # and sits on a healthy pod, so age is the only layer at which it differs from a healthy
+          # one -- the same reason DomesticPaymentStrandedBeforeClearing is written this way
+          # (#3273). The `> 300` IS the "stuck >5m" the old comment claimed, expressed against a
+          # series that exists; `for: 5m` then requires it to persist rather than catching one
+          # slow saga mid-flight.
+          #
+          # PENDING only. PROCESSING is mid-saga and legitimately takes seconds to low minutes;
+          # holding both to a 5m bound would page on normal traffic, and holding both to the
+          # loosest bound is the same as not alerting on PENDING at all.
+          expr: max(openbank_transactions_non_terminal_oldest_age_seconds{status="PENDING"}) > 300
           for: 5m
           labels:
             severity: critical
             service: transaction
           annotations:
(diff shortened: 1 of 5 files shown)


JiRaska:
Closing as superseded: #5787 already merged to main and fixes the same TransactionSagaStuck-never-fires bug (#5733) with its own independent gauge (StuckSagaGauge / `openbank_transaction_sagas_stuck`), landed at 55c2cdd0a. This branch's TransactionStrandedGauge / `openbank_transactions_non_terminal_oldest_age_seconds` implementation is a parallel, now-redundant fix for the identical alert rule. No functional gap remains to merge here.

PR #5759, opened 2026-08-19
fix: stop the three regulatory event consumers acking work they did not do (#5745)

JiRaska:
Fixes the three **regulatory** consumers from #5745 section A — the ones where a lost event means a
wrong number on a filed statutory return.

Stacked on #5719: this branch merges `fix/event-handlers-stop-acking-failed-work` so it can reuse
the shared `EventRetry` in `openbank-libs-runtime` rather than introduce a third retry
implementation. **The PR base is that branch**, so the diff shown here is only this work. #5719 was
still OPEN when this was written; if it merges first the base collapses to `main` on its own.

## What changed, per service

| service | swallowed call | now |
|---|---|---|
| `anacredit` | `projections.applyIfNewer` | retried, then **rethrown**; `apply_error` counted first |
| `analytics-sink` | `ClickHouseAnalyticsSink.write` | retried, then **nacked** (manual-ack consumer) |
| `tax-reporting` | `TaxFilingService.observe` | retried, then **rethrown**; `failed` counted first |

The split is the whole point and it is not a blanket rethrow: a **malformed payload** still acks in
all three, because replaying it fails identically forever. Only a **transient downstream failure**
is retried and re-raised.

`analytics-sink` nacks rather than rethrows because its signature is `Message<String>`, which took
it off auto-ack. It used to ack in a `finally`, so a failed write was acked along with everything
else; every terminal path now states its own outcome exactly once.

## ⚠️ DLQ status — two of three get one, the third cannot yet

**Corrected after review: my first version of this section said none of the three had a DLQ.** That
was true when measured and went stale inside the same session, because #5751
(`fix/kafka-incoming-dlq-wiring`, open) is wiring them while this branch is open — the same failure
mode #5757 hit. Verified against `gh pr diff 5751`:

| channel | after #5751 |
|---|---|
| anacredit `lending-events-in` | **wired** — `openbank.dlq.anacredit.lending-events-in` (config + `KafkaTopic` CR + Write ACL) |
| analytics-sink `analytics-events-in` | **wired** — `openbank.dlq.analytics-sink.analytics-events-in` |
| tax-reporting `withholding-remitted-in` | **deliberately baselined** — no DLQ |

So once #5751 lands, two of these three rethrows genuinely park the record.

**tax-reporting cannot have one, and that is the operationally important part.** #5751 baselines it
with the reason *"no KafkaUser in gitops — a Write ACL cannot be granted, so a DLQ would wedge on the
send"*. Confirmed independently: `openbank-tax-reporting-service` has no `KafkaUser` manifest
anywhere under `openbank-infra/gitops/components/` — in fact no gitops presence at all. So SmallRye's
default `fail` applies and **the channel halts on a persistent failure**: a §38d remittance arriving
during a tax-db outage stops the consumer group until someone intervenes.

That is still the right trade — a halted channel is loud and its backlog is intact, where the old ack
silently understated a return that then got filed — but it is a real property of a statutory filing
path, so the KDoc states it plainly rather than burying it: until a KafkaUser exists for that service,
**this consumer wedging IS the alert**. Creating one is the unblock, and is out of scope here.

**What the KDocs now assert.** Not the config value of the moment — that is exactly what went stale.
Each states the mechanism the handler controls (the record is not acknowledged as done) and names the
connector's `failure-strategy` as the thing that decides what follows: `dead-letter-queue` parks it,
the default `fail` stops the channel. They point at #5751 and the `check-incoming-dlq-wiring` gate for
the current answer. tax-reporting is the one exception, stated concretely, because there the halt is
not a transient state of the config.

Worth sequencing #5751 before this merges.

## analytics-sink: what "quarantined" actually means

#5745 flagged that the only `DeadLetterSink` binding is `LoggingDeadLetterSink`, one `log.warnf`,
while the KDoc promised a "visible, counted and replayable" quarantine and a ClickHouse
`dead_letter_events` adapter "landing with the warehouse adapter". Resolved by fixing the KDocs, in
all three places that made the claim.

The sharper finding while checking it: **the table exists.** `clickhouse/V1__analytics_bronze_silver.sql`
creates `openbank_analytics.dead_letter_events`, the ClickHouse init ConfigMap ships the same DDL,
and `dashboard-openbank-business-warehouse.yaml` has a panel selecting from it. No Kotlin anywhere
inserts into it. So the table is structurally present and permanently empty, and that panel reads
zero for a reason that has nothing to do with the DLQ rate — a green-looking "no dead letters" that
is really "no writer". Documented in place; implementing the adapter is not in this PR.

## Falsification

Every new behaviour test was run against the **pre-fix** consumer (restored from `origin/main`) to
confirm it can detect the absence of what it tests.

**Red pre-fix, green post-fix**
- anacredit — persistent failure rethrows; transient failure recovers; every-terminal-branch outcomes
- analytics-sink — persistent sink failure nacked and never acked; transient sink failure recovers
- tax-reporting — persistent write rethrows; transient write recovers; new `malformed` counter

**Green on BOTH sides** (this is what makes the above meaningful rather than merely different)
- anacredit — malformed payload is acked, projection never called
- analytics-sink — un-projectable payload quarantined and still acked; healthy event written + acked
- tax-reporting — malformed payload acked and never reaches the filing service; ignored event type;
  duplicate batch; happy path

The tax malformed-payload case is deliberately **two** tests. Folding the new `malformed` counter
into the ack assertion made that test go red pre-fix for a reason unrelated to acknowledgement,
which would have made the "green on both sides" claim false.

Also worth recording: anacredit's existing test asserted the **opposite** — *"a repository failure
is swallowed so the consumer group is not wedged"* — and passed. That is how this shipped under a
green test certifying the defect as intended behaviour.

## Verification

`ktlintCheck`, `detekt` and `test` per module, tasks confirmed present in the output rather than
inferred from a green exit:

|---|---|---|---|
| `openbank-anacredit-service` | 44 | 12 (`@QuarkusTest` ITs, Docker-gated) | 0 |
| `openbank-analytics-sink` | 85 | 0 | 0 |
| `openbank-tax-reporting-service` | 32 | 0 | 0 |

One anacredit IT (`PostgresCreditExposureRepositoryIT`, "Failed to start quarkus") failed on a run
with a second Gradle build active and passes in isolation — the documented concurrent-build boot
flake, in a Postgres repository IT with no consumer involvement.

`check-event-handler-swallows.py --enforce` (the gate from #5719) is clean over all 170 handler
files. Note it would have been clean before this PR too: `applyIfNewer` and `sink.write` are two of
the six shapes #5745 section C measured it as blind to. Widening the gate is tracked there and is
not attempted here.

## Governance

`openbank-tax-reporting-service` is **not** in `rules.yaml: money_path_services`, so ADR-0030's
2-approval + threat-model requirement does not apply, and there is no
`docs/threat-models/openbank-tax-reporting-service.md` to update. Both threat-model gates are scoped
to money-path services; `check-threat-models.py` passes (23/23). No trust boundary moves in this
diff — no REST surface, no new outbound edge, no authn/listener keys, no NetworkPolicy or Deployment
change.

Refs #5745, #5698

### DIFF
--- a/openbank-anacredit-service/src/main/kotlin/com/openbank/anacredit/infrastructure/kafka/LoanStageEventConsumer.kt
+++ b/openbank-anacredit-service/src/main/kotlin/com/openbank/anacredit/infrastructure/kafka/LoanStageEventConsumer.kt
@@ -8,6 +8,7 @@ import com.openbank.anacredit.application.port.out.AnaCreditMetricsPort
 import com.openbank.anacredit.application.port.out.LoanStageEventOutcome
 import com.openbank.anacredit.application.port.out.LoanStageProjectionRepository
 import com.openbank.anacredit.domain.model.LoanStageProjection
+import com.openbank.libs.messaging.EventRetry
 import jakarta.enterprise.context.ApplicationScoped
 import jakarta.inject.Inject
 import org.eclipse.microprofile.reactive.messaging.Incoming
@@ -24,14 +25,33 @@ import java.util.UUID
  *
  * Same shape as `kyc-service`/`aml-service`'s `PartyEventConsumer`: `suspend @Incoming` (Quarkus
  * dispatches on the Vert.x event loop with a duplicated context, so the reactive-Panache write inside
- * [LoanStageProjectionRepository] runs correctly), poison-pill safe (a parse/domain failure is caught,
- * logged, and the message is acked — one malformed event must not wedge the consumer group; the
- * canonical lending stream can be replayed).
+ * [LoanStageProjectionRepository] runs correctly).
  *
- * Every consumption reports its terminal outcome to [AnaCreditMetricsPort]. That is the point of the
- * meter rather than a nicety: acking a bad event is the correct behaviour, so a broken producer, a
- * schema change or a wedged projection write leaves the stage projection frozen with nothing but an
- * INFO/ERROR line to show for it. `outcome=parse_error|malformed|apply_error` is that population.
+ * **Failure handling separates two things this consumer used to conflate (#5698/#5745).**
+ *
+ * A **malformed event** — unparseable JSON, no `loanId`, no `newStage` — is unretryable: replaying it
+ * fails identically forever. It is logged, counted and acked. That is the genuine poison pill, and it
+ * is the only case that may be acked on failure.
+ *
+ * A **failed projection write** is the opposite: the event is fine, the database is not, and the work
+ * must happen once it recovers. Acking there froze the loan's IFRS 9 stage and DPD at their previous
+ * values with nothing but an ERROR line to show for it — and AnaCredit exposure is a *regulatory*
+ * return, so a stage that silently stops advancing is a wrong number on a filed report, not a stale
+ * dashboard. `applyIfNewer` is idempotent and ordering-safe, so a retry or a redelivery is free.
+ * Those failures now go through [EventRetry.withRetry] and are RETHROWN.
+ *
+ * **What the rethrow does, and what it deliberately does not promise.** The mechanism this handler
+ * controls is the rethrow: the record is not acknowledged as done. What follows is the connector's
+ * `failure-strategy` for `lending-events-in` — `dead-letter-queue` parks the record in the configured
+ * topic, SmallRye's default `fail` stops the channel. That is configuration, not a property of this
+ * code, so this KDoc does not state today's value: #5745 section B found the whole family of #5698
+ * fixes asserting a dead-letter that only four channels fleet-wide actually had, and #5751 is wiring
+ * this one to `openbank.dlq.anacredit.lending-events-in`. Read the channel's config, or the
+ * `check-incoming-dlq-wiring` gate, for the current answer.
+ *
+ * Every consumption still reports its terminal outcome to [AnaCreditMetricsPort], `apply_error`
+ * included — the counter is incremented *before* the rethrow, so the population survives whichever
+ * failure-strategy the channel is configured with.
  *
  * Idempotent + ordering-safe: [LoanStageProjectionRepository.applyIfNewer] only writes when the
  * incoming event's timestamp is strictly newer than the projection's current value, so an out-of-order
@@ -102,32 +122,41 @@ class LoanStageEventConsumer {
     }
 
     /**
-     * Write the projection and report the terminal outcome. A write failure is logged, counted and
-     * swallowed — the message is acked so one bad row cannot wedge the consumer group, and the
-     * canonical lending stream can be replayed.
+     * Write the projection and report the terminal outcome.
+     *
+     * A write failure is a dependency being down, not a bad event: it is retried a bounded number of
+     * times and then RETHROWN, so the platform sees a failure instead of an ack that did nothing. The
+     * `apply_error` counter is still incremented first, so the metric records every terminal failure
+     * regardless of which failure-strategy the channel is configured with.
      */
-    @Suppress("TooGenericExceptionCaught") // any write failure means the same thing: ack and count it
+    @Suppress("TooGenericExceptionCaught") // count the terminal outcome for ANY failure, then rethrow
     private suspend fun apply(projection: LoanStageProjection) {
         try {
-            if (projections.applyIfNewer(projection)) {
-                log.infof(
-                    "[lending-events-in] Applied stage projection for loan %s: %s (DPD %d)",
-                    projection.loanId,
-                    projection.stage,
-                    projection.daysPastDue,
-                )
-                metrics.loanStageEvent(LoanStageEventOutcome.APPLIED)
-            } else {
-                log.infof(
-                    "[lending-events-in] Ignored stale/duplicate loan.stage_changed for loan %s " +
-                        "(existing projection is already at least as recent)",
-                    projection.loanId,
-                )
-                metrics.loanStageEvent(LoanStageEventOutcome.STALE)
+            EventRetry.withRetry(log, "loan.stage_changed projection", projection.loanId) {
+                applyOnce(projection)
             }
         } catch (e: Exception) {
-            log.errorf(e, "[lending-events-in] Failed to apply stage projection for loan %s", projection.loanId)
             metrics.loanStageEvent(LoanStageEventOutcome.APPLY_ERROR)
+            throw e
+        }
+    }
+
+    private suspend fun applyOnce(projection: LoanStageProjection) {
+        if (projections.applyIfNewer(projection)) {
+            log.infof(
+                "[lending-events-in] Applied stage projection for loan %s: %s (DPD %d)",
+                projection.loanId,
+                projection.stage,
+                projection.daysPastDue,
+            )
+            metrics.loanStageEvent(LoanStageEventOutcome.APPLIED)
+        } else {
+            log.infof(
+                "[lending-events-in] Ignored stale/duplicate loan.stage_changed for loan %s " +
+                    "(existing projection is already at least as recent)",
+                projection.loanId,
+            )
+            metrics.loanStageEvent(LoanStageEventOutcome.STALE)
         }
     }
 }
--- a/openbank-anacredit-service/src/test/kotlin/com/openbank/anacredit/infrastructure/kafka/LoanStageEventConsumerTest.kt
+++ b/openbank-anacredit-service/src/test/kotlin/com/openbank/anacredit/infrastructure/kafka/LoanStageEventConsumerTest.kt
@@ -8,6 +8,7 @@ import com.fasterxml.jackson.databind.ObjectMapper
 import com.openbank.anacredit.application.port.out.LoanStageProjectionRepository
 import com.openbank.anacredit.domain.model.LoanStageProjection
 import com.openbank.anacredit.infrastructure.observability.AnaCreditMetricsAdapter
+import com.openbank.libs.messaging.EventRetry
 import io.micrometer.core.instrument.simple.SimpleMeterRegistry
 import io.mockk.coEvery
 import io.mockk.coVerify
@@ -15,11 +16,15 @@ import io.mockk.mockk
 import kotlinx.coroutines.runBlocking
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertThrows
 import java.time.Clock
 import java.time.Instant
 import java.time.OffsetDateTime
 import java.time.ZoneOffset
 import java.util.UUID
+
+/** A named transient failure, so the tests below never `throw RuntimeException` (detekt). */
+private class ProjectionUnavailable(message: String) : RuntimeException(message)
 
 class LoanStageEventConsumerTest {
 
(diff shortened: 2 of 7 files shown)


JiRaska:
From mapping the #5698 sweep's merge order tonight (full analysis on #5698):

This branch carries an **older `check-event-handler-swallows.py` than #5719** — it lacks `RUN_CATCHING` and `TRANSPORT_RECEIVER`, both of which #5719 has. `RUN_CATCHING` is the one that matters: it catches `runCatching { repo.save(x) }.getOrNull()`, the same defect with no `catch` keyword, in five places the first gate version could not see at all.

The two do conflict on that file (so nothing lands silently), but resolving it by taking this side would drop that detection. Suggest rebasing onto #5719 first — this PR's own consumer fixes (anacredit, analytics-sink, tax-reporting) don't overlap it.

Separately, #5751 × this PR conflict on `EventRetry.kt` — worth checking which version of the shared helper you want, since #5751 is the DLQ wiring every rethrow in this sweep depends on (SmallRye's default `failure-strategy` is `fail`, which stops the channel rather than dead-lettering).

JiRaska:

This PR's base is **`fix/event-handlers-stop-acking-failed-work` (#5719)**, not `main`, so the conflict was with the stack parent. Merged the parent in (`805e247b5`).

**One conflict, and it was a comment.** `openbank-libs-runtime/.../EventRetry.kt`: both sides suppress `TooGenericExceptionCaught` on the same `catch`, with different rationale wording and **no behavioural difference**. Took the parent's three-line version, so `EventRetry.kt` is now byte-identical to the parent — this child contributes zero divergence to a libs-runtime file the whole fleet compiles against.

**`check-event-handler-swallows.py` resolved to the parent's copy — verified lossless, not assumed.** `git log parent..HEAD` over that file is **empty**: this branch contributed no edits of its own. The parent's copy is a strict superset (diff is additions only), adding the `runCatching` detection from `db86c201d`. Nothing of this branch's is dropped.

**Entry counts, before → after** (a clean merge can silently drop a list entry — this repo has lost a release-manifest entry exactly that way):

| file | before | after | lost |
|---|---|---|---|
| `.release-please-manifest.json` | 58 | 58 | none |
| `.github/gates/gates.yaml` | 159 | 159 | none (set-compared by gate id) |

**Build:**

```
:openbank-libs-runtime:test        PASS
tests=267  failures=0  errors=0  SKIPPED=0
  EventRetryTest  tests=6  failures=0  skipped=0
```

### ⚠️ One pre-existing failure this does NOT fix, and should not

`:openbank-libs-runtime:detekt` is **red**, on a file unrelated to this PR:

```
libs/llm/OpenAiCompatibleLlmGatewayClient.kt:72:26: The function chat is too long (61). The maximum is 60. [LongMethod]
```

It is **not** caused by this resolution — the file is byte-identical on this branch *before* my merge, on the merge result, and on the parent (`73649a2fd`), and it differs only from `main` (`ac484ab55`). `main` already carries the fix: **#6148**, "keep the LLM gateway chat() inside the detekt LongMethod budget", which extracted `buildRequest`. The stack is simply stale relative to it.

The remedy belongs in the **parent, #5719** — merging current `main` there flows #6148 down to every child. Doing it here instead would drag main's fraud-service conflict into a child PR, which is the parent's to resolve (see my note on #5715 for why that one is not a mechanical resolution).

JiRaska:
## Gate-file collision: **nothing for this PR to drop**

Four branches add `.github/scripts/check-event-handler-swallows.py` (absent on `main`), all
registering the same gate id, so a verdict was needed. This PR is the clean case.

**This branch's copy is byte-identical to the parent #5719's:**

```
#5719  .github/scripts/check-event-handler-swallows.py  sha1 389ad84d5ae1605f0e929d82823c460b3274c257
#5759  .github/scripts/check-event-handler-swallows.py  sha1 389ad84d5ae1605f0e929d82823c460b3274c257
```

It appears in this PR's diff only because the branch carries the parent's 8-commit base. **Change
nothing.** The verdict lands the gate via #5719 with #5726's blind-spot fixes on top; #5737 has been
asked to drop its copy, which is an older revision missing the `runCatching` detection.

Likewise the `gates.yaml` entry (`id: event-handler-swallows`) is inherited from the same parent
commit on all four branches, so it merges as identical lines — I verified the merged `gates.yaml`
carries **0 duplicate ids** (167 ids, exactly the union of both sides, 0 lost).

## The `:openbank-libs-runtime:detekt` red is cleared

The failure here was `OpenAiCompatibleLlmGatewayClient.chat` at 61 lines against a 60 budget, fixed on
`main` by #6148 (`64fb7e88c`). I merged current `main` into the parent (`dc3852b13`, pushed) and
verified by running the gate rather than assuming:

```
./gradlew --no-daemon --rerun-tasks :openbank-libs-runtime:detekt :openbank-libs-runtime:ktlintCheck
> Task :openbank-libs-runtime:detekt
> Task :openbank-libs-runtime:ktlintCheck
BUILD SUCCESSFUL in 53s          exit 0
```

Both tasks appear in the executed task list, so this is a real pass, not Gradle stopping early.

Rebase/merge this branch onto the updated parent and the detekt context should go green.

<sub>Read-only analysis; nothing merged and this branch was not touched.</sub>

JiRaska:
### Follow-up: the `CONFLICTING` badge on this PR is **stale** — don't chase it

After #5719's conflict resolution was pushed (`805e247b5` → `dc3852b13`), GitHub flipped this PR to
`CONFLICTING/DIRTY`. That is a stale computation, not a real conflict. The API still reports the
**old** base commit:

```
baseRefOid (GitHub) = 805e247b5   <- pre-push parent
parent HEAD (actual) = dc3852b13
```

A real `git merge` of this PR's head into the updated parent is clean:

```
git merge origin/fix/event-handlers-stop-acking-failed-work --no-commit --no-ff
Automatic merge went well; stopped before committing as requested
exit 0        conflicted paths: (none)
```

(Probe validated against a known-conflicting pair — #5737 vs `main` correctly exits 1 and names the
two fraud-service files — so the exit 0 is a real clean merge, not a probe reporting silence.)

Merging the updated parent should clear the badge. Verified on the merge result: `gates.yaml` has
**167 gate ids, 167 unique, 0 duplicates**, and nothing was silently dropped.

<sub>Local read-only verification in a scratch worktree; this branch was not pushed to or modified.</sub>

JiRaska:
### The `CONFLICTING` / `This branch has conflicts` banner here looks stale — it merges clean

GitHub shows this PR as `CONFLICTING` / `DIRTY`, but that verdict appears to be a cached
computation against an **older base tip**:

- GraphQL reports `baseRefOid` = `805e247b`
- the live tip of the base branch `fix/event-handlers-stop-acking-failed-work` is `dc3852b1`

Merging locally with `git merge-tree --write-tree` produces a clean tree and exit 0 against
**both** the stale base commit `805e247b` and the current tip `dc3852b1` — no conflicted paths
either way.

That probe was validated against a synthetic add/add conflict first, which it correctly reported
with exit 1, so the clean result is not the probe failing silently.

**Action:** nothing to resolve by hand as far as this can tell. Any push, or a base-branch
update, should force GitHub to recompute and clear the banner. Worth a re-check before anyone
spends time on a manual conflict resolution.

JiRaska:
Replanted as a single signed commit on the current head of the base branch
`fix/event-handlers-stop-acking-failed-work` (#5719, `dc3852b13`) — head is now `e08157c01`.

Same finding as #5726: the old head was a four-deep merge stack and GitHub reported CONFLICTING, but
a real `git merge` into the current base was clean, so the conflict was a stale mergeability cache.

The PR's diff against its base is now exactly the three regulatory consumers and nothing of #5719's:

- `openbank-anacredit-service` `LoanStageEventConsumer` + test
- `openbank-analytics-sink` `AnalyticsConsumer`, `DeadLetterSink`, `AnalyticsConsumerFailureTest`
- `openbank-tax-reporting-service` `WithholdingRemittedConsumer` + test

Dropped in the replant: this branch also carried an **older** copy of
`.github/scripts/check-event-handler-swallows.py` (137 lines behind the version #5726 ships on the
same base). Keeping it would have reverted #5726's widening depending on merge order; the script is
now untouched here, so the two siblings no longer collide on it.

Verification: `:openbank-anacredit-service:test`, `:openbank-analytics-sink:test`,
`:openbank-tax-reporting-service:test` all BUILD SUCCESSFUL, and ktlintCheck + detekt green for all
three. (A first combined run flaked on `PostgresCreditExposureRepositoryIT` under parallel
Testcontainers; it passes on its own and passes on the unmodified base.)

JiRaska:
Re-checked against #5719 at its rebuilt head `ec0474b0b` (32 files) and rebased onto it.

Containment result: **not contained**. #5719 touches none of `openbank-anacredit-service`,
`openbank-analytics-sink` or `openbank-tax-reporting-service` — its 32-file payload has no overlap
with the 7 files this PR now ships. The only file the two ever shared was
`check-event-handler-swallows.py`, and that older copy has been dropped from this branch (see the
previous comment), so the rebase onto the rebuilt #5719 was clean with no conflicts.

New head `84aaa29f1`, signed (`%G?` = G).

Verification on the rebased head: `:openbank-anacredit-service:test`, `:openbank-analytics-sink:test`,
`:openbank-tax-reporting-service:test` all BUILD SUCCESSFUL; ktlintCheck + detekt green for all three.
(`PostgresCreditExposureRepositoryIT` flakes only when the three modules run in one parallel
invocation; green on its own and green on the unmodified base.)

JiRaska:
Rebuilt on current `origin/main` (head `6346f943a`) after #6209 landed. **One file was dropped from
this PR's scope on purpose, and it needs your agreement rather than mine.**

`openbank-audit-service/.../AuditConsumer.kt` was the only conflict, and it is not a textual one.
`main` (from #6209) restructured that class and now states the opposite intent to this PR:

```kotlin
suspend fun consume(message: Message<String>) {
    try { persist(payload, addressOf(message)) }
    catch (e: Exception) {
        // This legacy multi-producer channel deliberately retains its historic availability
        // behaviour. D5 agent events use AgentAuditConsumer instead: it acknowledges only a
        // successful durable write, so a store failure is retried by Kafka rather than lost.
```

This PR's sweep wanted that same consumer to rethrow. #6209 instead split the stream: the legacy
mixed channel keeps acking, and the strict path moved to `AgentAuditConsumer`. That decision was
made *after* this branch was written, so resolving the conflict either way is a design call, not a
merge. I took **`main`'s version verbatim** — the file is now byte-identical to `origin/main` — and
left the other 31 files as they were.

Two things that makes true, both measured rather than assumed:

- The gate this PR adds is still clean with `main`'s version:
  `[event-handler-swallows] clean: 176 handler files, no swallowed state changes` (rc=0). So
  dropping the file costs the gate nothing; `AuditConsumer` is not a swallow the gate wants to flag.
- 3-dot delta against `main` is 31 files, 872 insertions / 343 deletions — the sweep is intact.

**What to decide:** whether the legacy audit channel should stay availability-first. #6209's comment
argues yes and names the replacement path; this PR's premise argued no, fleet-wide. If you want the
legacy channel to rethrow too, that is a follow-up on top of #6209's structure, not a revert of it.

Also worth knowing for merge order: #5726 and #5759 are stacked on this branch, so they land after
it. #5759 currently reads `CLEAN` in the PR list — that is clean against *this branch*, not against
`main`; merging it directly would merge into this branch.

PR #5763, opened 2026-08-19
fix(ci): stop the fuzz readiness wait aborting on a healthy service

JiRaska:
## The defect

The fuzz lane's readiness wait used `$DEV_PID` as a liveness proxy:

```bash
kill -0 "$DEV_PID" 2>/dev/null || break
```

`./gradlew` is a **launcher**. It hands the build to the Gradle daemon and the client can exit while `quarkusDev` keeps running under the daemon — so that check aborted the wait on a service that was booting perfectly well. It is a race, which is why the same service passes one run and reports `failed to boot` the next.

## Measured, not inferred

Run [32284304038](https://github.com/JiRaska/open-bank-oss/actions/runs/32284304038) — 4 of 6 services passed after #5714, 2 reported a boot failure. Both artifacts contradict the verdict:

| service | what its boot log actually says |
|---|---|
| `openbank-domestic-payment` | `Listening on: http://localhost:8116` at **18:21:52** — ~34s into a **180s** budget. Log ends one second later, when teardown killed it. **No error of any kind in the artifact.** |
| `openbank-settlement-service` | Same shape, one step earlier: still pulling a Dev Services LGTM container (46 mentions across 137 lines) when the wait gave up. Nowhere near the deadline. |

Neither service is at fault, and neither log says otherwise. The harness reported a defect the service did not have — the same class of failure #5714 fixed for the datasource username expression and the Temporal worker registrars, one layer further in.

## The fix

The loop runs its full budget and reports what it observed. A genuinely dead JVM now costs the remaining wait instead of exiting early — the right trade against silently mislabelling a healthy service, which is the outcome this lane exists to prevent. `$DEV_PID` is still used for teardown, which is what it is actually good for.

Elapsed time is echoed on success (`answered after Ns`), so a slow boot reads as a slow boot instead of having to be reconstructed from timestamps inside an artifact.

## Deliberately one variable

Settlement's Dev Services LGTM pull is a **separate** cost and is not touched here. Both services were aborted *early*, neither reached the deadline, so the PID race is the proven cause for both. If settlement still times out with the full 180s available, that is a different finding and one I will have measured rather than guessed. Changing both at once would make the attribution unreadable.

## Verification

- Falsification in progress: [run 32288156614](https://github.com/JiRaska/open-bank-oss/actions/runs/32288156614) dispatches this branch against exactly the two services that failed. Result posted here when it lands.
- `actionlint` **finding sets** compared against `origin/main` rather than read raw — one `SC2034` on each side, **no new finding**. (The line number moves, which is why the sets are compared without it; post-filtering a linter's output is how a real finding gets hidden.)
- YAML parses.

## Linked issues

Refs #5714

### DIFF
--- a/.github/workflows/api-fuzz.yml
+++ b/.github/workflows/api-fuzz.yml
@@ -258,11 +258,29 @@ jobs:
             # Services with a warm build cache passed; the ones without did not, which is why the
             # same service flips between runs. Doing the compile as its own step makes the wait
             # measure what it claims to measure, and a slow BUILD now shows up as a slow build.
+            # The compile status is READ, not discarded. `|| true` here would reproduce exactly
+            # the misattribution this PR exists to remove: a compile failure leaves no classes to
+            # run, quarkusDev exits immediately, the readiness loop below then spends its full
+            # 180s budget on a service that was never going to answer, and the job reports
+            # "failed to boot" about a COMPILE error. Fail fast instead, and name the real cause.
+            #
+            # The build log is its own file: quarkusDev opens `${svc}-boot.log` with `>` a few
+            # lines below, so appending the compile output there would have it truncated away
+            # before anyone could read it — the evidence destroyed by the next command.
             echo "==> [${svc}] compiling (outside the readiness budget)"
             BUILD_START=$SECONDS
+            BUILD_LOG="fuzz-reports/${svc}-build.log"
             ./gradlew ":${svc}:quarkusGenerateCode" ":${svc}:classes" \
-              --console=plain --quiet >> "fuzz-reports/${svc}-boot.log" 2>&1 || true
-            echo "==> [${svc}] compiled in $((SECONDS - BUILD_START))s"
+              --console=plain --quiet > "${BUILD_LOG}" 2>&1
+            BUILD_RC=$?
+            echo "==> [${svc}] compiled in $((SECONDS - BUILD_START))s (gradle exit ${BUILD_RC})"
+            if [ "$BUILD_RC" != "0" ]; then
+              echo "::error::[${svc}] COMPILE FAILED (gradle exit ${BUILD_RC}) — this is a build error, NOT a boot failure; see ${svc}-build.log artifact"
+              tail -50 "${BUILD_LOG}" || true
+              OVERALL=1
+              docker rm -f fuzz-pg >/dev/null 2>&1 || true
+              continue
+            fi
 
             echo "==> [${svc}] starting quarkusDev"
             # FORCE the credentials this job actually created the container with, rather than
--- a/.github/workflows/api-fuzz.yml
+++ b/.github/workflows/api-fuzz.yml
@@ -248,6 +248,22 @@ jobs:
               echo "==> [${svc}] disabling Temporal worker via ${prop}=false"
             done
 
+            # Compile BEFORE the readiness clock starts. The 180s budget below was being spent
+            # on the Gradle build, not on the Quarkus boot: measured on run 32288156614,
+            # `starting quarkusDev` was echoed at 18:43:10 and quarkusDev's first log line landed
+            # at 18:45:38 — 148 of the 180 seconds gone to configuration and compilation, leaving
+            # the application ~32s before teardown killed it. It never had a chance, and the job
+            # reported "failed to boot" about a service that had not finished starting.
+            #
+            # Services with a warm build cache passed; the ones without did not, which is why the
+            # same service flips between runs. Doing the compile as its own step makes the wait
+            # measure what it claims to measure, and a slow BUILD now shows up as a slow build.
+            echo "==> [${svc}] compiling (outside the readiness budget)"
+            BUILD_START=$SECONDS
+            ./gradlew ":${svc}:quarkusGenerateCode" ":${svc}:classes" \
+              --console=plain --quiet >> "fuzz-reports/${svc}-boot.log" 2>&1 || true
+            echo "==> [${svc}] compiled in $((SECONDS - BUILD_START))s"
+
             echo "==> [${svc}] starting quarkusDev"
             ./gradlew ":${svc}:quarkusDev" \
               -Dquarkus.datasource.jdbc.url="jdbc:postgresql://localhost:5432/${DB}" \
--- a/.github/workflows/api-fuzz.yml
+++ b/.github/workflows/api-fuzz.yml
@@ -265,8 +265,25 @@ jobs:
             echo "==> [${svc}] compiled in $((SECONDS - BUILD_START))s"
 
             echo "==> [${svc}] starting quarkusDev"
+            # FORCE the credentials this job actually created the container with, rather than
+            # trusting the service to pick them up from an env var. The fleet convention is
+            # `password: ${POSTGRES_PASSWORD:CHANGE_ME_LOCAL_DEV_ONLY}`, and the job env supplies
+            # POSTGRES_PASSWORD — but openbank-settlement-service names its expressions
+            # QUARKUS_DATASOURCE_USERNAME/PASSWORD instead, so POSTGRES_PASSWORD never reached it
+            # and Flyway died on `FATAL: password authentication failed for user
+            # "openbank_settlement"` after a 237s compile (run 32289597634). Only 3 of ~50
+            # services use that spelling, which is exactly why an assumption about it is worth
+            # nothing: this is the third naming convention this lane assumed and did not have
+            # (#5714 was the username expression and the six differently-named worker properties).
+            #
+            # System properties sit above application.yaml in SmallRye's ordinal chain, so they
+            # win over any `${...:default}` expression whatever it is spelled. username/password
+            # live on `quarkus.datasource`, above both `reactive:` and `jdbc:`, so one pair of
+            # overrides covers the reactive datasource and the Flyway/Agroal one together.
             ./gradlew ":${svc}:quarkusDev" \
               -Dquarkus.datasource.jdbc.url="jdbc:postgresql://localhost:5432/${DB}" \
+              -Dquarkus.datasource.username="${DBUSER}" \
+              -Dquarkus.datasource.password="${POSTGRES_PASSWORD}" \
               "${WORKER_FLAGS[@]}" \
               -Dquarkus.console.basic=true \
               --console=plain --quiet \
--- a/.github/workflows/api-fuzz.yml
+++ b/.github/workflows/api-fuzz.yml
@@ -260,13 +260,35 @@ jobs:
             # Readiness = the main HTTP socket answers anything (000 = conn
             # refused). /q/* may live on the management port in prod profiles;
             # here only socket-open matters before pointing schemathesis at it.
+            #
+            # The wait deliberately does NOT use $DEV_PID as a liveness proxy. `./gradlew` is a
+            # launcher: it hands the build to the Gradle daemon and the client can exit while
+            # quarkusDev keeps running under the daemon. `kill -0 $DEV_PID || break` therefore
+            # aborted the wait on a service that was booting perfectly well — a race, which is why
+            # the same service passed one run and "failed to boot" the next. Measured on run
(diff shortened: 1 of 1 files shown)


JiRaska:
**Update — the first commit was necessary but not sufficient. It revealed the layer underneath.**

Falsification run [32288156614](https://github.com/JiRaska/open-bank-oss/actions/runs/32288156614) dispatched this branch against exactly the two failing services. Both still failed — and the log says why, precisely:

```
18:43:10  ==> [openbank-domestic-payment] starting quarkusDev
18:45:38  quarkusDev's first log line
18:46:11  boot log ends (teardown)
```

**148 of the 180 seconds went to Gradle configuration and compilation**, leaving the application about 32 seconds before it was killed. The step itself ran the full 193s, so the wait behaved exactly as the first commit intended — it was measuring the wrong thing.

I verified my own change was actually in the run before drawing any conclusion (`WAITED=` present, no `kill -0` in the echoed script), because "my fix didn't work" and "my fix didn't run" look identical from the outside.

This also explains the flapping the PID race alone could not: a service whose Gradle build was warm answered inside the budget, one whose build was cold did not, and which is which changes run to run.

**Second commit**: compile as its own step, before the readiness clock starts. The wait now measures what it claims to. And the two costs are reported separately —

```
==> [svc] compiled in Ns          <- a slow BUILD
==> [svc] answered after Ns       <- a slow BOOT
```

— which were previously indistinguishable, and both surfaced as "the service failed to start".

Second falsification running: [32289597634](https://github.com/JiRaska/open-bank-oss/actions/runs/32289597634).

JiRaska:
**Third falsification: settlement passes.** [Run 32291869194](https://github.com/JiRaska/open-bank-oss/actions/runs/32291869194)

```
==> [openbank-settlement-service] compiled in 200s
==> [openbank-settlement-service] answered after 96s
==> [openbank-settlement-service] fuzzing http://localhost:8138
```

Full chain across the three commits, each measured rather than assumed:

| run | change | result |
|---|---|---|
| [32284304038](https://github.com/JiRaska/open-bank-oss/actions/runs/32284304038) | `main` after #5714 | 4/6 pass — settlement + domestic-payment fail |
| [32288156614](https://github.com/JiRaska/open-bank-oss/actions/runs/32288156614) | PID race removed | both still fail — but the wait now ran its full budget, exposing that 148s of 180 went to Gradle |
| [32289597634](https://github.com/JiRaska/open-bank-oss/actions/runs/32289597634) | compile outside the clock | domestic-payment **PASS** (`208s` compile + `62s` boot); settlement fails on `password authentication failed` |
| [32291869194](https://github.com/JiRaska/open-bank-oss/actions/runs/32291869194) | credentials forced | settlement **PASS** |

Each commit fixed exactly one thing and the next run said what remained. The `compiled in Ns` / `answered after Ns` instrumentation is what made each step legible instead of something to reconstruct from artifact timestamps.

---

**One caveat worth stating, because it limits what this run is evidence for.**

settlement's fuzz exercised **1 operation**, and schemathesis reported `Authentication failed: 1 operation returned authentication errors`. The lane's v1 scope is the *unauthenticated* surface, so for a service whose entire API sits behind `@RolesAllowed` a green run establishes very little beyond "the service boots and enforces auth uniformly".

That matters for the `pentest` attestation path this lane feeds (`attestations.yaml`, `by: ci-schemathesis`): a green job is **not** interchangeable across services. For consent-service — 13 operations, where this lane found two real 5xx — it is meaningful evidence. For settlement it is close to none. Attesting both off the same green would be exactly the "successful no-op shares a flag with real success" shape this repo keeps re-finding.

Not fixing that here — it is a scoping question about the attestation, not about this harness. Filing separately.

JiRaska:
**Full-fleet falsification: 26/26 green** ([32348238171](https://github.com/JiRaska/open-bank-oss/actions/runs/32348238171)), dispatched from this branch across the entire derived money-path + legacy set.

Before this branch, the last scheduled run on `main` was 7 failures of 23 — six of which never sent a single request. Chain of causes, each fixed and measured separately:

| defect | effect |
|---|---|
| datasource username taken from a config expression verbatim (#5714) | vop, settlement: `postgres never became ready` |
| six differently-named Temporal worker switches (#5714) | transaction, lending, sepa-payment, domestic-payment: boot hang |
| `kill -0 $DEV_PID` as a liveness proxy (commit 1) | wait aborted on healthy services, racily |
| Gradle build inside the readiness budget (commit 2) | 148s of a 180s budget gone before Quarkus logged a line |
| datasource password assumed from a naming convention 3 of ~50 services use (commit 3) | settlement: `password authentication failed` after a 237s compile |

**Caveat, stated because the green invites the wrong conclusion.** I counted what the run actually exercised, from each job's own `Selected: N/M` and `Authentication failed: N operations`:

```
operations selected total:  306
auth-blocked (401/403):     302
ACTUALLY exercised:           4   (1.3%)
services with zero unauthenticated surface: 24 of 26
```

So this PR makes the lane *reach* every money-path service — which it could not before — and the lane then finds that 24 of 26 have no unauthenticated surface to fuzz at v1 scope. That is a real and useful property (uniform auth enforcement, and an endpoint that loses its `@RolesAllowed` shows up here), but it is not the `pentest` evidence the attestation path wants. Numbers filed in #5769; no `pentest` attested for any service.

JiRaska:
## Review — live, not stale; one real gap

**Staleness:** the branch is 324 commits behind `main`, but `git log <merge-base>..origin/main -- .github/workflows/api-fuzz.yml` is **empty** — nobody has touched this workflow since the branch was cut, so the 3-dot diff is the whole delta and there is no drift to reconcile. Not a duplicate of anything merged.

**Confirmed good.** The three claims in the comments are each backed by a run id and a timestamp, and the diagnosis matches the shape the repo has been burned by repeatedly: `./gradlew` is a launcher, the Gradle daemon owns the JVM, so `kill -0 "$DEV_PID"` was measuring the *client*, not the service. Removing it as a liveness proxy is right, and paying the remaining budget on a genuinely dead JVM is the correct trade against mislabelling a healthy one. `DEV_PID` is still used for teardown (`kill "$DEV_PID"`), so it is not left dangling. The `-Dquarkus.datasource.password` value is `POSTGRES_PASSWORD: "fuzz-ci-throwaway"` — a job-level literal, not a repo secret, so there is nothing leaked onto a command line here.

**The one thing I would change before merge — the compile step swallows its own failure.**

`api-fuzz.yml` (new compile block):
```
./gradlew ":${svc}:quarkusGenerateCode" ":${svc}:classes" \
  --console=plain --quiet >> "fuzz-reports/${svc}-boot.log" 2>&1 || true
```
`|| true` means a compile failure produces `==> [svc] compiled in Ns`, then `quarkusDev` fails, then the readiness loop burns its full 180s and prints `::error::[svc] failed to boot`. That is the exact misattribution this PR exists to remove — the label would once again name the wrong phase, only now with the added cost of the full wait. Suggest capturing the status and failing with its own message:

```
if ! ./gradlew ... >> "fuzz-reports/${svc}-boot.log" 2>&1; then
  echo "::error::[${svc}] COMPILE failed (not a boot failure) — see ${svc}-boot.log"
  tail -50 "fuzz-reports/${svc}-boot.log" || true
  continue
fi
```

**Smaller:** `WAITED` is echoed only on the success path. On the failure path the error message still says only "failed to boot"; including `after ${WAITED}s` there costs one word and is the number a reader wants first.

**Negative control — what I could not establish.** There is no test here that the wait now reports correctly for a service that genuinely never boots. That is hard inside a workflow and I am not asking for one, but note the change is asymmetric: it strictly weakens early detection in exchange for correctness, so the evidence that it does not now hide a real boot failure is "the log tail is still printed", nothing stronger.

Verdict: **needs change** (the `|| true`), otherwise sound and well-evidenced. Not approving — review comment only.

JiRaska:
## The new pre-compile step swallows the failure this PR exists to stop misattributing

Pushed a fix to this branch (`e564e45b0`).

### The defect

The pre-compile step added here ends in `|| true`:

```bash
./gradlew ":${svc}:quarkusGenerateCode" ":${svc}:classes" \
  --console=plain --quiet >> "fuzz-reports/${svc}-boot.log" 2>&1 || true
```

A compile failure therefore reproduces **exactly the misattribution this PR removes**, one layer
down. No classes exist, so `quarkusDev` exits immediately; the readiness loop — which this PR
just changed to run its *full* budget rather than exit early on a dead PID — then spends all
180 seconds polling a socket that was never going to open; and the job reports

```
::error::[svc] failed to boot — see svc-boot.log artifact
```

about a **build error**. Same wrong cause, now after burning the whole wait.

There is a second half. The compile appends to `<svc>-boot.log`, and `quarkusDev` opens that same
file with `>` a few lines below — so the compiler output is **truncated away before anyone can
read it**. The evidence is destroyed by the next command, which is why this would have been
diagnosed from an artifact containing no error at all: the same symptom the PR body describes for
run 32284304038.

### The fix

Read the status, name the real cause, and skip the wait. The build gets its own log file so it
survives:

```bash
echo "==> [${svc}] compiling (outside the readiness budget)"
BUILD_START=$SECONDS
BUILD_LOG="fuzz-reports/${svc}-build.log"
./gradlew ":${svc}:quarkusGenerateCode" ":${svc}:classes" \
  --console=plain --quiet > "${BUILD_LOG}" 2>&1
BUILD_RC=$?
echo "==> [${svc}] compiled in $((SECONDS - BUILD_START))s (gradle exit ${BUILD_RC})"
if [ "$BUILD_RC" != "0" ]; then
  echo "::error::[${svc}] COMPILE FAILED (gradle exit ${BUILD_RC}) — this is a build error, NOT a boot failure; see ${svc}-build.log artifact"
  tail -50 "${BUILD_LOG}" || true
  OVERALL=1
  docker rm -f fuzz-pg >/dev/null 2>&1 || true
  continue
fi
```

The step runs under `set -uo pipefail` (no `-e`), so capturing `$?` is safe. `fuzz-reports/` is the
uploaded artifact path, so `<svc>-build.log` ships with the rest. Cleanup mirrors the existing
postgres-not-ready `continue`.

### Proved by what it prevents

The loop body was extracted and run against a stub `gradlew` that fails the `:classes` task —
the case the guard must now catch — and against one that succeeds.

| case | version | exit | what the job reported |
|---|---|---|---|
| compile FAILS | before (`\|\| true`) | 1 | `::error::[svc] failed to boot` ← **the misattribution**, after the full wait |
| compile FAILS | after | 1 | `::error::[svc] COMPILE FAILED (gradle exit 1) — this is a build error, NOT a boot failure` + the compiler output, wait skipped |
| compile OK, socket answers | before | 0 | proceeds to fuzz |
| compile OK, socket answers | after | 0 | proceeds to fuzz |

Exit codes read from `$?` after redirecting to a file, not through a pipe. Note the exit code is
**not** the discriminator on the first two rows — both are failures. What changes is *which cause
is named* and whether 180 s is burned first.

`actionlint` finding sets were diffed before vs after rather than eyeballed: identical, one
pre-existing `SC2034` (`i appears unused`) that is also present on `main`, only its line number
moves. YAML parses; `bash -n` on the extracted step is clean.

JiRaska:
**Not a duplicate of the sibling PR, but a collision worth reading before either merges.**

#5763 and #5884 are the only two open PRs whose entire diff is `.github/workflows/api-fuzz.yml`, and
their hunks overlap:

- #5763 — `@@ -248,9 +248,60 @@` and `@@ -260,13 +311,35 @@`
- #5884 — `@@ -209,6 +209,28 @@` and `@@ -248,55 +270,239 @@`

Both rewrite the region starting at line 248, from different premises: #5763 fixes the readiness
wait aborting on a healthy service, #5884 widens what the fuzzer exercises past the 401s. Neither
contains the other, so both are wanted — but the second to merge will conflict, and a conflict
resolved by taking one side wholesale silently reverts the other.

Suggested order: land #5763 first (it is the smaller, behaviour-preserving fix), then rebase #5884
onto it and re-read the merged step, rather than resolving the conflict hunk by hunk.

Neither is mergeable by an agent here in any case — a change under `.github/workflows/` alters what
runs on the self-hosted runners, so both need a human.

Issue #5706, opened 2026-08-19
prod-readiness: the three remaining repo-fixable NO-GO causes (tax-reporting, missing contract tests, mcp C4)


JiRaska:
## Finding

Three production-readiness gaps that are real work rather than scoring artifacts, left out of the collector-reconciliation PR (#5702) on purpose. Listed together because they are the complete remaining set of *repo-fixable* NO-GO causes; everything else in the matrix is attestation-gated (a pentest, a restore drill, a code-complete review — real-world facts, not code changes).

Evidence lines are from the reconciled collector on `origin/main`, 2026-08-19.

### 1. `openbank-tax-reporting-service` — not deployed, barely tested, no contract test

```
C2=1 [2 unit, 0 IT, kover=n]
C3=1 [openapi=y, NO contract test in src/test]
C5=0 [no CNPG cluster]
C7=1 [signed-provenance, vex-triage]
C8=1 [not deployed (no Deployment/Rollout in gitops), domain metrics]
```

It has a `version.txt`, an `openapi.yaml` and domain metrics, and **no workload in gitops at all**. Deliberately excluded from the kover PR (#5704): at 31.4 % LINE across 2 unit tests, any floor that would pass today certifies nothing, and adding one would flip C2 to "Verified" for a service that is not. Needs the deployment decision made first — ship it or mark it explicitly pre-release.

### 2. Missing contract tests — `campaign`, `engagement`, `tax-reporting`

```
C3=1 [openapi=y, NO contract test in src/test]
```

All three publish an `openapi.yaml` and none ships a contract test. Note the collector no longer counts the *word* "contract" in a comment (that is what made three services read Verified on prose alone), so these three are genuine.

Per CLAUDE.md, a consumer pact alone does not catch a wrong request path — only the provider replay does — so each needs the `@PactFolder` provider replay wired at the same time, and it must run on a PR (`@PactBroker` classes are skipped on the PR lane).

### 3. `openbank-mcp-service` — C4 rollback note

```
C4=1 [2 migrations, rollback_note=n]
```

`V1__agent_sessions.sql` and `V2__agent_session_jti.sql` have been applied, so **the migrations themselves must not be edited** (Flyway checksums the whole file, comments included). The fleet convention puts the rollback column in `src/main/resources/docs/04-data.{en,cs}.md`, and mcp-service ships no docs pack at all — so this is "write the service doc pack", not "add a comment". Sized accordingly.

## Not in scope here

[...]

PR #5788, opened 2026-08-20
test(campaign): contract-test the consent reads every outbound touch depends on

JiRaska:
## What

A consumer-driven contract between campaign-service and consent-service for the two reads every outbound campaign touch depends on (ADR-0219 D3/D4), **plus the provider replay that verifies it**.

Closes the `C3=1` half of #5706 for campaign-service.

## Why this pair, specifically

`GET /api/v1/suppressions/party/{partyId}` answered **500 on every call from the day it shipped** — `SuppressionEntity` mapped six of its ten columns to names V6 never created (#5711, fixed today). campaign-service calls that route on every send through `LiveSuppressionAdapter`.

The outcome was not a do-not-contact leak. `ContactPolicyGate` fails closed — verified in code, not from its comment:

```kotlin
} catch (ex: Exception) {
    log.warnf("contact gate state unavailable for %s/%s — failing closed: %s", ...)
    ContactGateDecision.denied(ContactDenyReason.GATE_UNAVAILABLE)
```

So the real effect was **campaigns silently not sending** — the safe direction, and invisible. Nothing in this repo could have caught it: campaign's tests mock the client, so no HTTP; consent's unit tests mock the repository, so no SQL.

**The provider replay is what closes that.** `ConsentPactProviderVerificationTest` runs against a real Postgres, so a column mismatch surfaces as a failed verification rather than as a 500 nobody looks at. A consumer pact alone cannot catch a broken provider — the mock server answers whatever it is asked (CLAUDE.md).

## Two things caught by checking rather than assuming

**`MARKETING` is not a `ConsentScope`.** I wrote the pact against it first. The real values are `MARKETING_COMMS_EMAIL` / `_PUSH` / `_INAPP`; the provider would have answered 400 and the replay would have gone red. Caught by reading the enum before running anything.

**The grantee is `party-service:marketing-comms`, not `campaign-service`.** It comes from `openbank.campaign.consent-grantee`. A pact written against a plausible-looking grantee pins a request this service never makes.

Both are exactly the defect class this test exists to catch, found in my own draft.

## The colon

The grantee contains `:`. Measured: with RestAssured's default encoding the request misses the pact and the mock server answers 500 — it percent-encodes a colon that a path segment does not require (RFC 3986 lists `:` as a pchar). `urlEncodingEnabled(false)` pins the unencoded form.

Which form the RESTEasy Reactive client actually puts on the wire is **not asserted**, and the comment says so. The provider replay settles it: if consent-service does not serve that path, verification goes red.

## Paths are literals

Both expected paths are written out, not derived from the client's `@Path`. Deriving both sides moves expectation and request together, so the test stays green against a route that does not exist (measured on #2290) — the asymmetry is the test.

Matchers are pinned where the value carries the decision: `scope` is `stringValue("ALL")` because an ALL entry must carry no `value`, and `granted` is `booleanValue(true)` because `ConsentCheckResponse` defaults it to `false` on an absent field — a `booleanType` matcher would let a provider that stopped sending it pass here and deny in production.

## Provider states

Two new `@State` handlers in `ConsentPactProviderVerificationTest`, seeded over JDBC like its existing one (pact-jvm invokes state callbacks on a bare JUnit thread with no Vert.x context) and idempotent (pact-jvm 4.7.3 calls each setup twice per interaction).

## Verification

**Provider replay, read from the JUnit XML** — not the console, because pact-jvm prints `Not all of the N were verified` on green runs and `@IgnoreNoPactsToVerify` can turn a missing pact into a silent skip:

```
tests=4 failures=0 errors=0 skipped=0
  ok   openbank-campaign-service - GET whether an active consent covers a party for a grantee and scope
  ok   openbank-campaign-service - GET the active suppressions for a party
  ok   openbank-mcp-service - POST validate an ACTIVE ACCOUNTS_READ consent covering the requested IBAN
  ok   openbank-mcp-service - POST validate an unknown consent id — the fail-closed deny
```

**Falsified, not claimed.** Removing ONE `@Column(name = "created_at")` from `SuppressionEntity` — a single line of the #5711 defect — turns exactly the suppressions interaction red and leaves the other three green:

```
tests=4 failures=1
  RED: openbank-campaign-service - GET the active suppressions for a party
```

So the claim "this would have caught #5711" is measured, not asserted.

- consumer test → 2 interactions, pact committed and byte-stable across a `ktlintFormat` round
- `ktlintCheck` + `detekt` on both modules → BUILD SUCCESSFUL
- `derive-pact-drift-scope.sh` now lists `openbank-campaign-service`; the scope is derived from the `@Pact` annotations, not hand-kept

## Linked issues

Refs #5706, #5711


JiRaska:
The one red check (`Pact drift check`) is **not a drift in this pact** — it is an OOM in the job itself:

```
Execution failed for task ':openbank-onboarding-service:quarkusGenerateTestAppModel'
> Java heap space
```

onboarding-service is innocent and untouched by this PR — it builds on its own in 6s. The job regenerates every consumer pact in one Gradle invocation with **no `GRADLE_OPTS` set**, so the daemon ran on Gradle's default heap; adding campaign to the derived scope made it 32 modules and it tipped. It would have tipped for whoever added the next pact.

Fixed separately in #5791 (touches `.github/workflows`, so it needs human review either way). This PR's pact content is unaffected, and its provider replay is green:

```
tests=4 failures=0 errors=0 skipped=0
```

PR #5790, opened 2026-08-20
test(engagement): contract-test the in-app marketing consent check

JiRaska:
> **Stacked on #5788.** Base is `test/campaign-consent-pact`, not `main` — both PRs add `@State` handlers to the same `ConsentPactProviderVerificationTest`, and stacking is CLAUDE.md's rule for that rather than racing for the file. Merge #5788 first; this diff then reduces to the engagement half.

## What

Consumer contract for the consent check engagement-service gates every `PROMOTIONAL_IMPRESSION` on (ADR-0219 D4), plus the provider state that verifies it.

Closes the `C3=1` half of #5706 for engagement-service.

## Not a copy of campaign's interaction

engagement asks for **`MARKETING_COMMS_INAPP`** — the in-app banner and personalised-challenge surface — where campaign asks per channel (`EMAIL` / `PUSH` / `INAPP`).

Those are **separate grants**. A party may hold the EMAIL one and not the INAPP one, so a contract verified only for EMAIL says nothing about the route engagement actually takes. Each consumer declaring what it actually sends is the whole reason pacts are consumer-driven, and it is why the provider gets a second state rather than the two consumers sharing one.

The grantee is read from `openbank.engagement.marketing-grantee` rather than assumed. It happens to share campaign's default — which is exactly the kind of coincidence worth not relying on.

## Verification

Provider replay, read from the **JUnit XML** rather than the console — pact-jvm prints `Not all of the N were verified` even on fully green runs, and `@IgnoreNoPactsToVerify` can turn a missing pact into a silent skip:

```
tests=5 failures=0 errors=0 skipped=0
  ok   openbank-campaign-service   - GET whether an active consent covers a party for a grantee and scope
  ok   openbank-campaign-service   - GET the active suppressions for a party
  ok   openbank-engagement-service - GET whether an in-app marketing consent covers a party
  ok   openbank-mcp-service        - POST validate an ACTIVE ACCOUNTS_READ consent covering the requested IBAN
  ok   openbank-mcp-service        - POST validate an unknown consent id — the fail-closed deny
```

- consumer test → 1 interaction, pact committed
- `ktlintCheck` + `detekt` on both modules → BUILD SUCCESSFUL
- `derive-pact-drift-scope.sh` picks the pact up from the `@Pact` annotations; the drift gate's scope is derived, not hand-kept

`granted` is pinned with `booleanValue(true)`, not `booleanType`: it **is** the decision, and `ConsentCheckResponse` defaults it to `false` on an absent field — so a matcher accepting any boolean would let a provider that stopped sending it pass here and suppress every impression in production.

## Linked issues

Refs #5706

### DIFF
--- a/openbank-consent-service/src/test/kotlin/com/openbank/consent/contract/ConsentPactProviderVerificationTest.kt
+++ b/openbank-consent-service/src/test/kotlin/com/openbank/consent/contract/ConsentPactProviderVerificationTest.kt
@@ -205,6 +205,33 @@ class ConsentPactProviderVerificationTest {
         }
     }
 
+    /**
+     * State for engagement-service's `EngagementToConsentPactConsumerTest`. Distinct from the
+     * campaign one on purpose: engagement gates promotional impressions on MARKETING_COMMS_INAPP,
+     * a separate grant from the EMAIL scope campaign's pact pins, so seeding one would not satisfy
+     * the other.
+     */
+    @State("an ACTIVE MARKETING_COMMS_INAPP consent covers the pact engagement party")
+    fun activeInAppConsentExists() {
+        dataSource.connection.use { c ->
+            c.autoCommit = false
+            if (!rowExists(c, "SELECT 1 FROM consents WHERE id = ?::uuid", PACT_INAPP_CONSENT_ID)) {
+                c.prepareStatement(INSERT_MARKETING_CONSENT_SQL).use { ps ->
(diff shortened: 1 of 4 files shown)


JiRaska:
Same as #5788: the red `Pact drift check` is an OOM in the job (no `GRADLE_OPTS`, 32 modules in one Gradle invocation), not a drift in this pact. Fixed in #5791. Provider replay here is green with all five interactions verified.

Issue #5671, opened 2026-08-19
ADR-0265 tail: reasoning graph, pgvector retrieval, Langfuse ingestion evidence


JiRaska:
ADR-0265 is `delivery-status: partial` — this issue is its tail.

Delivered so far:
- [x] Slice 2 — guardrails: `ContentSafetyPort` + Llama Guard adapter + LiteLLM route (#5670)
- [x] Slice 3 — Langfuse self-hosted, LiteLLM callback wired

Still unbuilt:
- [ ] Slice 1 — Kotlin reasoning graph over Temporal, and `agents.yaml` `reasoning:` reconciled from `langgraph` to what runs
- [ ] Slice 4 — pgvector retrieval in copilot-service (ADR-0183 executed): migration, embedding job, hybrid retrieval
- [ ] `agents.yaml` `observability:`/`guardrails:` as-built notes updated once both are verified live in sandbox
- [ ] Langfuse **ingestion** verified by something other than a pod health check — a green Langfuse pod with a rejected callback key is indistinguishable from an idle gateway from the outside, and there is no metric for it today (Langfuse v2 exposes no Prometheus endpoint, LiteLLM's own callback metrics are an Enterprise feature)
- [ ] Langfuse trace retention policy — traces grow with traffic and nothing expires them today
- [ ] SSO in front of the Langfuse UI before any environment with non-synthetic data (the public origin is a documented ADR-0056 exception, acceptable only while the sandbox is synthetic-only)
- [ ] Correlate the gateway-side Langfuse trace id with the calling service's own OTel trace (today the trace is what the gateway saw)

Refs ADR-0265, ADR-0031 D6/D7, ADR-0174, ADR-0183

JiRaska:
**E2E verification on sandbox — 2026-08-20.** Everything below is measured against the live cluster, not inferred from a green deploy.

### Working, with evidence

| Layer | Evidence |
|---|---|
| Langfuse ingestion | `auth_check: True`; trace `litellm-acompletion` carries **input prompt + output completion** (plus `reasoning_content`) |
| pgvector index | 13 rows / 4 docs / 13 distinct ids, model `BAAI/bge-m3` |
| Semantic retrieval | live query "jak si zablokuju ztracenou kartu?" → correct doc, top similarity **0.749**, correctly ordered |
| Indexer on boot | ran 90s after pod start, then `pruned 1 chunk(s)` (the stale row from the id bug) |
[...]

Issue #5736, opened 2026-08-19
Every LLM call fails: 41/41 http_error, zero successes — AI/LLM dashboard is empty as a symptom


JiRaska:
## What

Every LLM call the platform has ever made has failed. Measured 2026-08-19 against the sandbox Prometheus:

```
openbank_llm_requests_total                  -> outcome="http_error"  41
openbank_llm_call_duration_seconds_count     -> outcome="http_error"  41
```

`outcome` has exactly **one** label value across the whole series: `http_error`. There is no `success`, and `openbank_llm_tokens_total` has no series at all — consistent with no call ever returning a body to count tokens from.

## Consequence

The **AI / LLM Cost & Reliability** dashboard is empty, and this is a symptom, not a dashboard bug. Its panels are correct; there is simply nothing to show:

- "Call latency p50/p95/p99 (successful calls)" filters `outcome="success"` — a value that has never existed.
- The three cost recording rules (`openbank:llm_cost_usd_24h:total`, `:by_model`, `openbank:llm_tokens_unpriced_24h`) are loaded and evaluating, but produce no series because they derive from `openbank_llm_tokens_total`.

Worth separating from #5049's class: those metrics were never emitted by any code. These **are** emitted, correctly, by working instrumentation — they are reporting a real and total outage of the thing they measure.

## What to check first

`http_error` means the gateway answered with a non-2xx (as opposed to `unreachable`/`timeout`), so the endpoint resolves and something is responding. Likely candidates in order: credentials/virtual-key rejected, model id not served by the gateway, or a request-shape mismatch. The per-provider label is unset on these series, which is itself a small gap worth closing while investigating.

## Note on the metric design

Only having `http_error` in the label set means "is the LLM path healthy" is currently answerable only by *absence* of a success series, not by a ratio. Once fixed, an alert on `sum(rate(...{outcome="success"}[15m])) == 0` would need `absent_over_time` guarding, or it will be blind exactly when the path breaks completely — the same shape as #5733.

Found during a full observability audit, 2026-08-19.

JiRaska:
Investigated. **41/41 no longer reproduces** — re-measured against the sandbox Prometheus on 2026-08-20 ~18:15 UTC, `outcome` now has four values, not one, and `openbank_llm_tokens_total` has series. PR #6041 fixes what is left.

Exact query:

```promql
sum by (container, model, outcome) (increase(openbank_llm_requests_total[24h]))
```

| caller | model | outcome | 24h |
[...]

PR #5833, opened 2026-08-20
feat(observability): price the ADR-0265 models and alert on the guardrail's success state

JiRaska:
Two gaps the AI-stack delivery (#5670) opens. Independent of that PR — this lands on `main` on its
own and is inert until those routes carry traffic.

## 1. Price book

The gateway gained `meta-llama/llama-guard-4-12b` and `BAAI/bge-m3`; this file did not. Their tokens
would join against nothing, so fleet spend reads quietly low and `AiSpendUnpriced` fires — the
mechanism working, but only after the fact. Priced at DeepInfra's published rates.

bge-m3 gets a **`prompt` entry only**: an embedding call returns no completion tokens, and a price
for a kind that never appears sits in the book forever with nothing to join against — the opposite
failure from an unpriced model, and just as hard to notice.

## 2. Alerts on the SUCCESS state, not an error rate

Both new controls degrade **silently by design**, which is correct on a customer-help surface and
exactly why an error-rate alert cannot see them:

| alert | catches | why not an error rate |
|---|---|---|
| `ContentSafetyClassifyingNothing` | guardrail returning `unavailable` for >90% of messages | an unconfigured guardrail makes no *failing* calls — it makes none |
| `CopilotRetrievalKeywordOnly` | semantic retrieval that stopped contributing | an unbuilt index looks like slightly worse search |

Ratio, not count, so a quiet hour cannot fire and a busy one needs no retuned threshold. 90% rather
than 100% because a briefly unreachable provider self-heals while a control that is OFF sits at 1.0
forever. The retrieval alert is gated on hybrid having worked at least once — an **absent** series
and a series pinned at zero are different facts — and is `info`, since answers stay grounded and
cited.

Third instance of this shape in the repo (the push adapter counting skipped notifications as
delivered; the liveness gauge nobody re-derived at t=0), hence alerting on the control working.

## Verification

`promtool check rules`: **16 rules, SUCCESS**, and **falsified** — re-adding the `group_left` that
Prometheus rejects on `unless` makes the same command FAIL, so the check can detect what it is
trusted for. `kubectl apply --dry-run=server` clean.

## Linked issues

Refs #5671

### DIFF
--- a/openbank-infra/gitops/components/observability/prometheus-rules-ai.yaml
+++ b/openbank-infra/gitops/components/observability/prometheus-rules-ai.yaml
@@ -408,6 +408,7 @@ spec:
             severity: warning
             team: platform
           annotations:
+            subject_period: continuous
             summary: "content-safety guardrail for {{ $labels.model }} is classifying nothing"
             description: >-
               Over 90% of classifications returned `unavailable` for the last 30 minutes, so the
@@ -433,6 +434,7 @@ spec:
             severity: info
             team: platform
           annotations:
+            subject_period: continuous
             summary: "copilot help search has fallen back to keyword-only"
             description: >-
               Semantic retrieval has produced results before (so it is configured) but over 90% of
--- a/openbank-infra/gitops/components/observability/prometheus-rules-ai.yaml
+++ b/openbank-infra/gitops/components/observability/prometheus-rules-ai.yaml
@@ -382,3 +382,62 @@ spec:
               to the deterministic fallback. This is the documented safe mode (an unseeded key must
               not CrashLoop the pod), but it is not a steady state: the feature is off. Seed the key
               via ESO/OpenBao.
+
+        # ── Guardrail and retrieval: alert on the SUCCESS state, not an error rate ──────────
+        #
+        # These two controls are designed to degrade quietly — that is the correct behaviour on a
+        # customer-help surface, and it is exactly why an error-rate alert cannot see them. An
+        # unconfigured content-safety guardrail makes no FAILING calls; it makes none at all, and
+        # reports `unavailable` for every message. A retrieval index that never got built produces
+        # keyword-only answers that look like slightly worse search. Both are silent by
+        # construction, so the alert has to be about the control WORKING, not about it erroring.
+        #
+        # This is the shape the fleet has now shipped twice: a push adapter that counted every
+        # skipped notification as delivered, and a liveness gauge nobody re-derived at t=0. The
+        # metric exists here specifically so the third instance is visible.
+        - alert: ContentSafetyClassifyingNothing
+          # Ratio, not a count: a low-traffic hour must not fire, and a busy one must not need a
+          # retuned threshold. 90% rather than 100% because a genuinely unreachable provider for a
+          # few minutes is normal and self-heals; a control that is OFF stays at 1.0 forever.
+          expr: |
+            sum by (model) (increase(openbank_guardrail_classifications_total{decision="unavailable"}[1h]))
+            /
+            sum by (model) (increase(openbank_guardrail_classifications_total[1h])) > 0.9
+          for: 30m
+          labels:
+            severity: warning
+            team: platform
+          annotations:
+            summary: "content-safety guardrail for {{ $labels.model }} is classifying nothing"
+            description: >-
+              Over 90% of classifications returned `unavailable` for the last 30 minutes, so the
+              model-based guardrail is effectively absent — on a fail-open surface every message is
+              going out unclassified, and on a fail-closed one every request is being refused. The
+              deterministic prompt-injection filter is unaffected either way. Causes, in the order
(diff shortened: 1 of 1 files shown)


JiRaska:
Heads-up from #6045 / #6161: diffed by content against current `main`, this branch would **regress** `prometheus-rules-ai.yaml` rather than add to it:

- it reverts `interval: 1m` back to `1h` on the `openbank.ai.price-book` group — that is the exact defect #6041 fixed. At `1h` the price series are resolvable for only 5 minutes in 60 (Prometheus answers instant queries from a 5m lookback), so `openbank.ai.spend` at `5m` saw no price on 11 of every 12 evaluations, `openbank:llm_cost_usd_24h:total` had no series at all, and `AiFleetDailySpendHigh` could not fire.
- it reverts `llama-3.3-70b-versatile` to Groq's `$0.59/$0.79` with `price_source: litellm-map`. Groq decommissioned that model; the route now resolves to DeepInfra's `Llama-3.3-70B-Instruct-Turbo` at `$0.10/$0.32` (#5736).

Worth a rebase onto `main` before it goes any further — the base it was cut from predates both fixes.

JiRaska:
## Conflict resolved — merged `main`, pushed `2d48da102`

The conflict was in `prometheus-rules-ai.yaml` only, and it was **not** the additive conflict it looked like.

### The trap

Both sides added price-book entries in the same region, so the obvious resolution is "keep both". That would have been wrong. `main` had *already* added `meta-llama/llama-guard-4-12b` (prompt + completion) and `BAAI/bge-m3` (prompt) **outside the conflict region**, at byte-identical rates to this branch's proposal:

| model / kind | this branch | already on `main` |
|---|---|---|
| `meta-llama/llama-guard-4-12b` / prompt | `1.8e-07` | `1.8e-07` |
| `meta-llama/llama-guard-4-12b` / completion | `1.8e-07` | `1.8e-07` |
| `BAAI/bge-m3` / prompt | `1e-08` | `1e-08` |

Keeping both sides produced three duplicate `openbank:llm_price_usd_per_token` series with identical `model`/`kind` labels. That does not fail loudly — it **double-counts spend** in the join, quietly overstating the very figures this file exists to make trustworthy. I only caught it by parsing the merged YAML and counting `(model, kind)` pairs; reading the conflict hunk alone could not show it, because the duplicate half was never in the hunk.

### Resolution

`main`'s price book is taken whole; this branch's price entries are dropped as superseded. The genuinely new content — **`ContentSafetyClassifyingNothing`** and **`CopilotRetrievalKeywordOnly`** — merged without conflict and is fully preserved. That is the part of this PR `main` does not have.

**Entry counts (before → after)**
| | `main` | branch | resolved |
|---|---|---|---|
| rule groups | 3 | 3 | **3** |
| alerts | 6 | 6 | **8** |
| price entries | 11 | — | **11** |
| duplicate `(model, kind)` | 0 | — | **0** |
| duplicate alert names | — | — | **0** |

Alerts go 6 → 8: `main`'s six plus this branch's two, no duplicates. Price entries stay at `main`'s 11. YAML parses.

### One thing worth a second look before merge

`ContentSafetyClassifyingNothing` divides `decision="unavailable"` by total over `[1h]`. That fires correctly when the classifier is reachable-but-failing. It cannot fire if the guardrail emits **no series at all** — an absent metric makes the whole expression absent rather than 1.0. `CopilotRetrievalKeywordOnly` explicitly handles that case with its `and on() (... > 0)` clause; the content-safety one does not. Given this PR's own framing — alerting on the *success* state, where a control that is off is the failure being hunted — that asymmetry may be deliberate, but it is the one gap I would want confirmed rather than assumed. I did not change it: which way it should go is your call, not a merge-conflict decision.

I also could not verify the `meta-llama/llama-guard-4-12b` rate against DeepInfra's live catalogue from here (read-only, no outbound probing). The number is unchanged from what `main` already ships, so this merge does not introduce it either way.

Not merged — leaving that to the required checks and a reviewer.

JiRaska:
### CI triage: one genuine red, one phantom

**Genuine — this diff:** `gates (gitops)` — *alert window accommodates its subject's period
(#5736, enforced)*. Both new rules in
`openbank-infra/gitops/components/observability/prometheus-rules-ai.yaml` have a `[1h]` range
window and no `subject_period` annotation:

- `ContentSafetyClassifyingNothing`
- `CopilotRetrievalKeywordOnly`

Declare `subject_period: continuous` (if driven by real traffic) or the subject's actual period.
Note `AiModelNeverSucceeds` already got exactly this treatment on `main` in #6203 — same fix shape.

**Phantom — not this diff:** `gates (api)` reports `SettlementStatus … ['LEDGER_REVERSED']`
enum drift. That is the gate bug fixed by #6220 (`b85ac16`, 14:57Z); this PR's
`refs/pull/5833/merge` predates it. Rebase to clear — `gh run rerun` re-runs the stale merge
SHA and will not.

`Validate manifests` is the roll-up of those shards.

PR #5843, opened 2026-08-20
fix(ci): a dead-letter topic is WRITTEN, so the ACL gate must ask for Write

JiRaska:
## Summary

`check-kafka-acl-coverage` infers the required operation from the channel key — `incoming` ⇒ `Read`,
`outgoing` ⇒ `Write` — and collects any `topic` key beneath it. A SmallRye dead-letter topic is
configured as `dead-letter-queue.topic` **nested under the incoming channel**, so it came out as
`Read`. It is **produced to**: the connector parks a failed record by PRODUCING to that topic, and
nothing consumes it by design.

## Why this is more than a wrong label

The broker runs `allow.everyone.if.no.acl.found=false`. So a PR could satisfy this gate by granting
`Read` and still have every DLQ send denied at runtime — **a green check over a dead-letter queue
that cannot be written**. That is worse than having no DLQ: the rethrow it exists to serve then
*stops the channel* instead of parking the record.

## Ground truth, already on `main`

`accounts/kafka-account-mtls.yaml` grants `dead-letter-topic-party-events-in` `[Write, Describe]`,
with the comment: *"Without Write here, the DLQ send itself fails and the consumer [retries]"*.

## How it was found, and the evidence

Diagnosing why **#5751** is red on `gates (data)`. That PR adds an explicit DLQ topic to every
incoming channel fleet-wide, which is the first change that makes this reachable at scale — every
one of its ~40 new topics is reported as a missing **Read** ACL.

Ran the gate's own `required()` against that branch:

```
openbank-balance-service:
   openbank.dlq.balance.balance-init-in    -> gate requires Read
   openbank.dlq.balance.ledger-events-in   -> gate requires Read
openbank-aml-service:
   openbank.dlq.aml.party-events-in        -> gate requires Read
```

Executed, not reasoned about.

## Falsifiability

The self-test now asserts the direction on a **real config shape** (incoming + nested
`dead-letter-queue` + outgoing) rather than on the regex — the bug was invisible in the pattern and
only showed up in what the extraction *returned*. Falsified both ways:

- fix removed → `selftest FAIL: direction inference wrong` naming `Read` where `Write` was wanted;
- restored → passes.

`--enforce` on `main` with the fix: **107 (topic, operation) pairs across 42 KafkaUsers, 0 gaps.**

## Scope
One file, `.github/scripts/`. No service code, no gitops, no `rules.yaml`.

Refs #5745, #2598

### DIFF
--- a/.github/scripts/check-kafka-acl-coverage.py
+++ b/.github/scripts/check-kafka-acl-coverage.py
@@ -44,6 +44,7 @@ from __future__ import annotations
 
 import argparse
 import pathlib
+import tempfile
 import sys
 
 import yaml
@@ -107,7 +108,18 @@ def required(app_yaml: pathlib.Path) -> set[tuple[str, str]]:
     for path, value in flat:
         if not path or path[-1] not in ("topic", "topics") or not isinstance(value, str):
             continue
-        if "incoming" in path:
+        # A dead-letter topic is WRITTEN by the connector, even though its config sits under an
+        # `incoming` channel. Direction-by-channel-key gets this exactly backwards: SmallRye parks
+        # the failed record by PRODUCING to the DLQ; nothing consumes it (by design — the
+        # party-events DLQ alert says so in as many words). The working example on main is
+        # account-service, whose KafkaUser grants `dead-letter-topic-party-events-in`
+        # `[Write, Describe]` with the comment "Without Write here, the DLQ send itself fails and
+        # the consumer [retries]". Classifying it Read would let a PR satisfy this gate with an ACL
+        # the broker still denies — a green check over a DLQ that cannot be written, which is worse
+        # than no DLQ at all: the rethrow it exists to serve then stops the channel (#5745).
+        if "dead-letter-queue" in path:
+            operation = "Write"
+        elif "incoming" in path:
             operation = "Read"
         elif "outgoing" in path:
             operation = "Write"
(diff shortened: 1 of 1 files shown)


JiRaska:

## The direction fix is right, and the negative control is real

The reasoning is correct: SmallRye parks a failed record by **producing** to the DLQ, so classifying a `dead-letter-queue` topic as `Read` because it nests under an `incoming` channel would let a PR satisfy this gate with an ACL the broker still denies. `check-kafka-acl-coverage.py:111-121` gets that right.

The self-test addition (`:195-225`) is a genuine negative control, not decoration — it asserts the **returned set** from `required()` on a real config shape, so flipping the classification back to `Read` reddens it. That is the right way to test this; a regex assertion would not have been. Ran it in place:

```
--selftest  → exit 0   "5 coverage cases both directions, plus DLQ direction"
--enforce   → exit 0   "104 (topic, operation) pair(s) checked across 41 KafkaUser(s), 0 known gap(s)"
```

Exit codes read from `$?` on a redirect, not through a pipe.

## The finding: the new branch matches nothing in the fleet, so no DLQ ACL is actually checked

The `"dead-letter-queue" in path` branch only fires when a service declares an **explicit** `dead-letter-queue.topic` key. **Zero of the four services that use `failure-strategy: dead-letter-queue` declare one:**

- `openbank-account-service`
- `openbank-aml-service` → (correction: `openbank-fraud-service`)
- `openbank-transaction-service`
- `openbank-card-issuance-service`

All four rely on SmallRye's implicit default name, `dead-letter-topic-<channel>`. The repo says so itself, in the very manifest this PR cites as its worked example:

- `openbank-infra/gitops/components/kafka/kafka.yaml:216` — *"Name is SmallRye's default (dead-letter-topic-&lt;channel&gt;) — set implicitly, not via a"* config key
- `openbank-infra/gitops/components/accounts/kafka-account-mtls.yaml:60` — *"Default SmallRye name = dead-letter-topic-&lt;channel&gt;."*

So `required()` never extracts a DLQ topic for any real service, and `dead-letter-topic-party-events-in` — the exact ACL the PR description holds up as the working example — is **not among the 104 pairs the gate checks**. The change is correct in principle and inert in practice: the gate still reports "clean" over a set that contains no DLQ topic at all. That is the failure shape this repo already knows well — a gate whose subject set silently excludes the thing it was written to check reads as passing, never as unchecked.

**Suggested fix**, and it is small: when a channel declares `failure-strategy: dead-letter-queue` with no explicit `dead-letter-queue.topic`, synthesise `dead-letter-topic-<channel-name>` and require `Write` on it. That makes the four live services checkable and turns the account-service example from a citation into a covered case. Adding the implicit case to the self-test alongside the explicit one would keep both honest.

## Second, smaller finding: the enforce path does not assert a non-empty subject set

`--selftest` correctly fails with *"no KafkaUser CRs found — the scan itself is broken"*. The ordinary run does not: with a broken root it printed

```
0 (topic, operation) pair(s) checked across 0 KafkaUser(s), 0 known gap(s) — clean.
```

and **exit 0**. I hit this by accident running the script from a copied path (`REPO = parents[2]`, `:54`). `gates.yaml:2263-2266` does wire the self-test ahead of the run, so this is defence-in-depth rather than a live hole — but a `min_subjects`-style floor on the enforce path would make "the scan found nothing" impossible to read as "clean", which is the same property the self-test already insists on.

## Everything else checked, nothing found

Release scope: `fix(ci)` touches only `.github/scripts/`, which is not a released package — releasing type, non-releasing path, so no version bump is owed. Scope creep: none, single file. No secrets or private hostnames in the added comments.

JiRaska:
Correction to my comment above — the bulleted service list was garbled by an editing slip. The four services that declare `failure-strategy: dead-letter-queue` are:

- `openbank-account-service`
- `openbank-card-issuance-service`
- `openbank-transaction-service`
- `openbank-fraud-service`

`openbank-aml-service` is **not** one of them; it appeared in that list in error. The finding itself is unchanged and was measured on the correct four: none of them declares an explicit `dead-letter-queue.topic`, so the new `Write` branch extracts no topic for any of them and no DLQ ACL is covered by the 104 checked pairs.

JiRaska:
Closing — **#5751 now carries this fix itself**, and its version is strictly more complete.

Mine only classified the NESTED form (`"dead-letter-queue" in path`). Theirs also handles the flat
dotted leaf `dead-letter-queue.topic:` as a key in its own right — which matters in this repo
specifically, because SmallRye's YAML source quotes a leaf key containing a dot (#686), so both
spellings turn up in the wild and only one of them is readable by the connector. They also have an
equivalent falsifiable self-test built on a real config fixture.

Keeping it inside #5751 is also the better home: the DLQ wiring and the gate change that makes it
checkable land atomically, instead of one PR being red until the other merges.

The diagnosis stands and is unchanged — it is in my comment on #5751: the gate derived direction
from the enclosing `incoming` key, so every `openbank.dlq.*` topic came out needing **Read**, while
a DLQ is produced to. Granting Read would have turned the gate green while the broker
(`allow.everyone.if.no.acl.found=false`) still denied every send — a green check over a
dead-letter queue that cannot be written.

No action needed here; the analysis found its way into the right PR.

Issue #5735, opened 2026-08-19
Mobile RUM dashboard has no data: neither openbank-app nor admin-ui sends traces, and screen.name/app.version/device.model are never set


JiRaska:
## What

The **Mobile RUM (Real-User Monitoring)** dashboard has no data source of truth behind it. Every one of its panels is empty, and the pipeline that should feed it delivers nothing.

Measured 2026-08-19 against the sandbox:

- **Tempo, last 3h**: 42 services producing traces. `openbank-app`: **absent**. `openbank-admin-ui`: **absent**.
- **Tempo, all indexed services**: 38. Neither app nor admin-ui appears.
- **`traces_spanmetrics_calls_total`**: 39 services, all backend. Query `service=~"$app.*"` (what 5 of the 6 RUM panels use) → **0 series**.
- The 6th panel is a Tempo query `{resource.service.name="openbank-app"}` → nothing in 3h.

## The RUM attributes were never populated either

`rum-attribute-audit` (the repo's own cronjob, 168h lookback, last run 3d ago) reports unique-value counts:

```
service.name                 43   <- includes openbank-app, so spans DID arrive at some point in the week
screen.name                   0   <- the dashboard's two headline panels are per-screen
app.version                   0
device.model                  0
x-correlation-id              0
os.type                       1   ("linux" — that is the backend pods, not iOS/Android)
```

So even in the window where `openbank-app` spans existed, they carried **none** of the attributes that make a span RUM. `screen.name = 0` is the decisive one: "Screen-view rate (per screen)" and "Top screens" cannot work without it.

`openbank-admin-ui` is a separate and simpler gap — it has never sent a trace at all, so it has no RUM and no distributed tracing.

## What IS working (so this is narrow)

- `rum-gateway` (otelcol-contrib) is running and healthy, exporting metrics on its own pipeline.
- GlitchTip **is** receiving from both apps — `openbank-app` newest event today, `openbank-admin-ui` yesterday. Client-side *error* reporting works; it is *tracing/RUM* that does not.

That contrast is useful: it means the apps have working network egress to the observability stack and a working SDK integration for one signal, so this is a RUM-instrumentation gap, not a connectivity one.

## Suggested scope

1. Decide whether mobile RUM is actually in scope right now — if not, the honest fix is to remove or clearly mark the dashboard rather than leave 6 permanently-empty panels that read as "the app has no traffic".
[...]

PR #5847, opened 2026-08-20
feat(admin-ui): emit browser RUM spans with screen.name/app.version/device.model

JiRaska:
## What

`openbank-admin-ui` has never emitted a trace. This adds browser RUM: one screen-view span per App-Router navigation, carrying the attributes the "Mobile RUM" board and the `rum-attribute-audit` cronjob look for.

**Scope is `openbank-admin-ui` only.** `openbank-app` is a separate repository — the mobile half of #5735 is untouched and this PR does not close the issue. See *What openbank-app still owes* below.

`Refs #5735`

## What admin-ui had already

- **Nothing OpenTelemetry.** No `@opentelemetry/*` dependency, no tracing config, on either the server or the browser side.
- GlitchTip (`@sentry/nextjs`) for *errors* only — which is why the issue could observe working egress but no traces.

## The OTLP endpoint question, and its answer

The repo does have a public OTLP ingress: `rum-gateway` serves `https://rum.open-bank.tech/v1/traces`. **A browser cannot use it**, for three independent reasons, each sufficient on its own:

1. Its OTLP receiver (`openbank-infra/gitops/apps/rum-gateway.yaml`) declares no `cors:` block, and the nginx Ingress carries no CORS annotations. A cross-origin POST carrying `Authorization` + JSON needs a preflight `OPTIONS`, which nothing answers.
2. Its `oidc` extension pins `issuer_url` to the **openbank-customers** realm with `audience: openbank-rum`. admin-ui staff authenticate against the **openbank** realm (`KEYCLOAK_REALM=openbank`); that authenticator rejects their tokens.
3. This app's own CSP (`src/proxy.ts`) is `connect-src 'self' <keycloak> <glitchtip>`. `rum.open-bank.tech` is not in it.

That gateway is deliberately mobile-shaped, and its own header comment says so.

**So the browser posts same-origin** to `/api/telemetry/traces`, and a Next.js route handler forwards server-side to `otel-collector.observability.svc:4318`, which already exports to Tempo. This clears all three: it is `'self'` under the existing CSP (no CSP change needed), it is already gated by the next-auth middleware (no new auth surface), and the pod already reaches `*.observability.svc` (it holds `TEMPO_URL` and `PROMETHEUS_URL` there today, and nothing selects `otel-collector` in a NetworkPolicy).

`traces_spanmetrics_calls_total` is produced by **Tempo's own metrics-generator**, enabled for all tenants (`gitops/apps/tempo.yaml`), not by a collector connector — so anything that lands in Tempo produces span metrics. The relay path therefore feeds five of the six panels, and the sixth (a TraceQL query) directly.

## `NEXT_PUBLIC_*` is baked — which value goes where

- **Collector endpoint** — `OTEL_EXPORTER_OTLP_ENDPOINT`, read **server-side at request time** in the relay route, added to the admin-ui Deployment. Deliberately not `NEXT_PUBLIC_*`: it must vary per environment without a rebuild. Unset means the relay answers `204` with `x-rum-relay: disabled` and drops the batch — RUM is *off*, not silently broken, and the header keeps "disabled" from reading as "working".
- **`app.version`** — `NEXT_PUBLIC_BUILD_VERSION`, derived from `BUILD_VERSION` in `next.config.mjs`, same pattern as `NEXT_PUBLIC_GLITCHTIP_RELEASE`. Baked at build time is *correct* here: it has to name the bundle the browser is running, not what the pod's env says today.

## The dashboard variable — please review this one

`dashboard-openbank-rum.yaml` had a hidden constant `$app = openbank-app`, used as `service=~"$app.*"`. `openbank-admin-ui` does not match it, so perfect instrumentation would still have rendered nothing. Widened to `openbank-(app|admin-ui)` and the board's help text updated to name both producers. If the preference is instead to keep this board mobile-only and give admin-ui its own, say so and I will split it.

## Attributes, and how identifiers are kept out

Resource: `service.name=openbank-admin-ui`, `service.version`, `app.version`, `device.model`, `os.type`, `os.version`. Span: `screen.name`, `http.route`; span **name** is `screen.<screen>`, because the headline panels are `sum by (span_name) (...)` and Tempo's metrics-generator dimensions on span name — a `screen.name` attribute alone would have left "Top screens" empty.

`device.model` / `os.type` map to a small closed set (the RUM cardinality budget in `prometheusrule-rum-cardinality.yaml`), never a raw user-agent string.

This is a bank operator console, so a span never carries a raw pathname. Every path segment that reads as an identifier collapses to `:id` and the query string is dropped, so `/parties/9f3c1e2a-.../cases/8812` becomes `/parties/:id/cases/:id`. The test asserts real route words with digits survive (`customer-360`, `day-end` are actual routes here — the first draft masked them, and the test caught it).

## Verify by effect, and the falsification

The tests drive a **real** `WebTracerProvider` with an `InMemorySpanExporter` — the subject is what an *exported* span carries, not that a tracer was configured. Removing each thing under test turns them red:

| removal | result |
|---|---|
| drop the `screen.name` / `http.route` span attributes | 1 failed — `expected undefined to be '/payments/sepa'` |
| drop `app.version` + `device.model` from the resource | 1 failed — `expected undefined to be '1.2.3'` |
| stop naming the span after the screen | 1 failed — `expected 'screen-view' to be 'screen./customer-360'` |
| all restored | 9 passed |

**Honest gap:** this proves a span is exported *in a test*, with the right attributes, to the right same-origin URL. It does **not** prove Tempo receives it in sandbox — that needs a deploy I have not done. The remaining unproven links are the relay's real hop to `otel-collector.observability.svc:4318` and the collector's export to Tempo. The check after deploy is that `openbank-admin-ui` appears in Tempo's service list and `rum-attribute-audit` reports non-zero `screen.name`.

## What `openbank-app` still owes

Everything on the mobile side. It is a separate repo and nothing here reaches it:

- Emit spans at all, continuously — the audit found `service.name` counted `openbank-app`, so spans arrived in bursts at some point in the week and then stopped.
- Set `screen.name`, `app.version`, `device.model` on them — the audit counted **0** unique values for all three, so even the spans that did arrive were not RUM spans.
- The `TELEMETRY_RUM` consent scope and the `aud=openbank-rum` audience mapper (threat-model O4, flagged pre-GA blocker in the gateway manifest), plus the app-side flag that is off by default.

Until that lands, the board's mobile half stays empty and #5735 stays open.

## Verification

- `npm test` → **197 files, 1541 tests, all passing** (via `npm test`, so `pretest` bakes `governance.json` / `catalog.json`).
- `npx tsc --noEmit` → clean.
- `eslint` on all touched files → clean.
- Dependencies added exact-pinned (no ranges), lockfile regenerated with plain `npm ci` + `npm install` — the diff is **additions only** apart from the lockfile's own stale `version` field (`0.155.4` → `0.173.2`, which `package.json` already said). An earlier attempt with `--legacy-peer-deps` pruned peer deps including `@testing-library/dom`; that was discarded rather than committed.


JiRaska:
## Not resolved — this conflict encodes an architectural decision, not a merge

I picked this PR up as part of a conflict-resolution sweep and stopped before pushing anything. The mechanical conflicts here are small (`package.json`, `package-lock.json`, `dashboard-openbank-rum.yaml`; `next.config.mjs` and `admin-ui.yaml` auto-merge). The problem is what one of them *says*.

### What the conflict actually is

`main` gained #6070 on 2026-08-20, which rewrote the RUM dashboard description to read, verbatim:

> **admin-ui is not here on purpose**: ADR-0088 D4 rejected browser RUM as the wrong surface.

I checked that against the ADR on live `main` rather than trusting the prose. `docs/adr/0088-...md` line 136:

> **Grafana Faro** (Apache-2.0) | **Rejected — wrong surface** | Web/JS RUM on an **internal** operator console (ADR-0056, ~2 users) is near-zero value; the customer surface is mobile, covered by D4 + GlitchTip.

So the rejection is real and it is on the record. Resolving this conflict means choosing between "ship browser RUM from admin-ui" and "the merged dashboard and ADR say we decided not to". Either side I take asserts an architecture position on your behalf. That is not mine to make silently inside a merge commit.

### A supporting argument that does not hold

I initially thought this PR was the *missing producer* for `main`'s RUM stack, because `cronjob-rum-attribute-audit.yaml` audits exactly `screen.name`, `app.version` and `device.model` and no admin-ui code on `main` emits any of them. That inference was wrong, and I want to correct it rather than leave it as ammunition: the audit job buckets those three attributes into `MOBILE_TOTAL`, so it is auditing the **KMP app**, not the console. The apparent gap is by design, not an oversight.

### What would unblock it

One of:

1. **The ADR moved on** — if browser RUM on the console is now wanted (the ~2-user premise may simply no longer hold), that is an ADR-0088 amendment or a superseding ADR. With that in hand the conflict resolution is mechanical and I am happy to do it.
2. **The ADR still holds** — then this PR should close, and #6070's dashboard note is already the documentation of why.

Two mechanical notes for whoever resolves it, so they are not rediscovered:

- The branch is 244 commits behind; admin-ui went `0.173.2` → `0.184.3` on `main`. `version.txt` is not in the conflict set, so the merge takes `main`'s version correctly — it should not be hand-edited either way.
- `package-lock.json` has 9 conflict regions. It must be regenerated with `npm install` from the resolved `package.json`, not hand-merged. The dependency sets genuinely diverge: this branch is on `sdk-trace-web` + `exporter-trace-otlp-http` (pinned), `main` on `sdk-node` + `exporter-trace-otlp-proto` + `instrumentation-undici` (carets).

Nothing pushed to this branch; its ref is untouched.

JiRaska:
## Review — contradicted by a merged decision; recommend **close with explanation**, not rescue

**Staleness/decision check first, as it decides everything else.** Branch is 267 commits behind `main` and `CONFLICTING`.

**1. This contradicts ADR-0088 D4, which is Accepted on `main`.**
`docs/adr/0088-observability-extension-oncall-slo-retention-mobile-rum.md:136` lists, in the *Explicitly rejected* table:

> **Grafana Faro** (Apache-2.0) | **Rejected — wrong surface** | Web/JS RUM on an **internal** operator console (ADR-0056, ~2 users) is near-zero value; the customer surface is mobile, covered by D4 + GlitchTip.

The rejection is of *browser RUM on the admin console as a surface*, not of one vendor. This PR implements exactly that surface with a different SDK. That may well be the right call to revisit — but it is an ADR-level reversal, and reversing an Accepted ADR in a feature PR is the one shape the governance rules do not allow.

**2. `main` already says, in the artifact this PR edits, that this cannot land without that reversal.**
`openbank-infra/gitops/components/observability/dashboard-openbank-rum.yaml:28-31` on `main`:

> `# openbank-admin-ui is deliberately NOT on this board: ADR-0088 D4 rejected browser`
> `# RUM as the wrong surface (internal operator console). The gateway also emits no`
> `# CORS headers and accepts only customer-realm tokens, so admin-ui could not export`
> `# here without an ADR-level reversal.`

**3. The CORS/CSP objection is *refuted* — credit where due.** I checked this expecting it to be a blocker and it is not. `openbank-admin-ui/src/lib/telemetry/rum.ts:29-46` documents all three failure modes (no `cors:` block on the OTLP receiver, `oidc` pinned to the **openbank-customers** realm, `connect-src` not listing the gateway) and routes around them with a same-origin `/api/telemetry/traces` relay, keeping the collector address a runtime env var rather than a baked `NEXT_PUBLIC_*`. That is a correct and careful design. It removes the *mechanical* objection entirely and leaves only the ADR one — which is why this reads as needing a decision rather than a fix.

**4. Merging as-is would also silently revert #6070's dashboard content.** The 3-dot diff replaces the `How to read this board` panel with the branch's older text, deleting `main`'s "**Empty is normal**", the measured ~1 trace/6h figure, the runbook pointer, and the "Android only / iOS is a stub" note. Same class as the merge-that-deletes-a-list-entry trap: no conflict marker would tell a reviewer that operational guidance is being dropped.

**5. Not checked, because 1-4 decide it:** the OTLP relay route's authn/rate limiting, the PII `toScreenName` reduction (which looks right in principle), and the `package-lock.json` delta.

**Recommendation: close with the explanation above, and if browser RUM on the console is genuinely wanted, open an ADR amendment to 0088 D4 first.** The relay design in `rum.ts` is good enough to be worth citing in that amendment rather than throwing away. Not approving and not merging — review comment only.

JiRaska:
## Rebuilt on live `main`; two real defects removed. One blocker left, and it is not CI.

The branch was 270 commits behind. Rebuilt it by replaying `merge-base..head` onto `origin/main` and resolving each conflict on the merits. New head: `6ebe733d6`, signed (`%G?` = `G`).

### 1. `gates (gitops)` — the collector Service does not exist

The gate's message was the real finding, not noise:

```
in-cluster host 'otel-collector.observability.svc' at `admin-ui.admin-ui.env.OTEL_EXPORTER_OTLP_ENDPOINT`
does not resolve: namespace 'observability' exists in gitops but declares no Service 'otel-collector'.
```

I enumerated every `kind: Service` with `namespace: observability` across `openbank-infra/gitops/`. There are five — `goalert`, `holmes-rca`, `ntfy`, `pyrra-api`, `pyrra-kubernetes` — and no `otel-collector`. So the relay would have posted into a hostname that resolves to nothing, and the PR body's claim that it "already exports to Tempo" was not true of that name.

Worse, and invisible in the diff: `admin-ui.yaml` has **one** container, and `main` already sets `OTEL_EXPORTER_OTLP_ENDPOINT: http://tempo.observability.svc:4318` on it (landed with the BFF-tracing work). The branch appended a *second* env entry with the same key. Kubernetes keeps the last one, so merging this would have silently repointed the working server-side BFF tracing at a non-existent Service — breaking tracing that works today, with no error anywhere.

**Resolution: dropped the added env block entirely.** `openbank-infra/gitops/components/admin-ui/admin-ui.yaml` is now byte-identical to `main`. This is not gate-appeasement — the relay reads `process.env.OTEL_EXPORTER_OTLP_ENDPOINT`, which now resolves to `tempo.observability.svc:4318`, a Service that exists and whose OTLP/HTTP receiver serves the `/v1/traces` path the relay appends. The relay works *because* the block is gone.

### 2. `gates (api)` — ASVS V9.1 follows from the same fix

The single NEW V9.1 finding was that same line. With the block removed there is nothing to baseline, so I made no baseline edit. I did check the premise anyway: baselining would have been the wrong call regardless, because a baseline entry for a plaintext edge to a Service that does not exist is a stale entry the moment it is written, and this baseline is bidirectional.

### Gate output, run locally on the rebuilt tree

```
$ python3 .github/scripts/check-asvs-l3.py --enforce --baseline .github/asvs-l3-baseline.txt
V9.1 no new plaintext inter-service edges: 239 finding(s), 0 NEW vs baseline, 0 baseline lines paid-off
V13.1 every resource publishes a spec: 8 finding(s), 0 NEW vs baseline, 0 baseline lines paid-off
exit 0

$ python3 .github/scripts/check-incluster-hostnames.py
check-incluster-hostnames: 59 application.yaml files (86 hostnames) + 516 gitops workload env/arg values
(507 hostnames), against 304 declared Services (+141 corroborated, +10 Helm-provided) across 60 namespaces
OK: every in-cluster hostname resolves — in application.yaml AND in the gitops workload env.
exit 0

$ npm test -- --run src/test/rum-instrumentation.test.ts
Test Files  1 passed (1)      Tests  9 passed (9)
```

### 3. Also dropped: the dashboard rewrite

The branch rewrote `dashboard-openbank-rum.yaml`, and the rewrite is older than what it replaces. Its header re-asserts an "iOS Ktor OTLP exporter", which `main`'s current header explicitly corrects as false ("ADR-0088 D4c is still a no-op stub pending SPM vendoring; this header used to claim ... and it is not"). It also deletes `main`'s measured evidence about why the board is empty and the pointer to `docs/runbooks/0013-rum-cardinality.md`. Taking the branch's side would re-introduce a claim the repo has already falsified, so I kept `main`'s version.

### The remaining blocker — needs a human, not a rebase

`main` says, in two places, that browser RUM on this console is a *decided-against* option:

- `docs/adr/0088-...md`: Grafana Faro — **"Rejected — wrong surface"**, because "Web/JS RUM on an **internal** operator console (ADR-0056, ~2 users) is near-zero value".
- `admin-ui.yaml`, next to the env this PR tried to overwrite: "ADR-0088 D4's rejection of browser RUM for this console is untouched and still stands."
- `dashboard-openbank-rum.yaml`: "openbank-admin-ui is deliberately NOT on this board ... could not export here without an ADR-level reversal."

This PR *is* browser RUM on that console. That makes it a decision reversal, and reversals go through an ADR (`docs/adr/new.sh`), not through a green CI run. I have deliberately not resolved that in either direction — the code is now correct and the gates are green, but **whether this should ship at all is the open question**, and I am not the one to answer it.

Note also that part of the original motivation has since been met independently: `main` already gives admin-ui server-side spans to Tempo and RED metrics via `traces_spanmetrics_*`. What is left in this PR is specifically the *browser* half.

JiRaska:
Closing as **contradicted by a recorded decision**, not as bad work.

`docs/adr/0088-observability-extension-oncall-slo-retention-mobile-rum.md` on `main`, in D4's
alternatives table, records:

> **Grafana Faro** — **Rejected — wrong surface** — Web/JS RUM on an **internal** operator console
> (ADR-0056, ~2 users) is near-zero value; the customer surface is mobile, covered by D4 + GlitchTip.

The rejection is on **value grounds for this surface**, not on the tool — so instrumenting admin-ui
with a different library does not route around it. `main`'s RUM dashboard says the same in the
panel header (added by #6070): admin-ui is absent on purpose and adding it needs an ADR-level
reversal.

This PR would also silently revert that dashboard guidance.

**What was good here and should not be lost:** the CORS/CSP obstacles are *not* a defect in this
PR — `rum.ts:29-46` documents all three (no `cors:` block on the gateway, `connect-src` not listing
it, customer-realm-only OIDC) and routes around them with a same-origin relay. That is a sound
design for a surface that had been decided against, which is the only reason it is being closed.

**If the decision should change**, the route is an ADR superseding D4's alternatives row with the
value argument that has changed — an operator console with more than ~2 users, or a specific
incident this would have shortened. Reopen or re-raise on that basis and this implementation is a
reasonable starting point.

Verified before closing: the ADR text quoted above read from `origin/main`, not from a branch or a
summary.

PR #5849, opened 2026-08-20
fix(governance): require an exercised-surface floor before api-fuzz mints a pentest attestation

### LINKED ISSUE #5769: The pentest attestation still mints off a bare green — record the exercised operation count and set a floor (the authenticated lane landed; the floor did not)
## Finding

The `pentest` attestation is earned by a green `api-fuzz.yml` job, referenced by run URL (`attestations.yaml`, `by: ci-schemathesis`). A green job is **not interchangeable across services**, and nothing in the mechanism records the difference.

Measured on the two ends of the range:

| service | operations fuzzed | what the green means |
|---|---|---|
| `openbank-consent-service` | 13 selected / 13 total | real evidence — this lane found **two genuine 5xx** on its unauthenticated surface (run 32103867665) |
| `openbank-settlement-service` | **1** selected / 1 total, and schemathesis reported `Authentication failed: 1 operation returned authentication errors` | close to nothing — the service boots and enforces auth, and that is all |

The lane's v1 scope is deliberately the **unauthenticated** surface (endpoints behind `@RolesAllowed` answer 401/403). For a service whose entire API sits behind auth, a green run establishes that the service starts and that auth is uniformly enforced — both worth knowing, neither a pentest.

## Why it matters

Attesting `C7 = Bank-grade` for both services off the same kind of green is the shape this repo keeps re-finding: a successful no-op sharing a flag with a real success (`PushResult.skipped()` carrying `success = true`, ADR-0252 phase 0; the CNPG archiver that reports success with no object store, #5755). The attestation would be recording "a lane ran" while reading as "an adversarial test passed".

It also decays the wrong way. The TTL is 21 days precisely so a dead lane expires the attestation — but a lane that runs and exercises one authenticated endpoint renews it forever while testing nothing.

## Suggested shape

Do not let a bare green mint the attestation. Record what was actually exercised, and set a floor:

1. Have the lane emit, per service, the schemathesis `Operations: N selected` count and the number that returned only auth errors (both are already in its output, unparsed today).
2. Require a minimum exercised surface before `pentest` may be attested for that service — a service below it gets an explicit "not meaningfully fuzzable at v1 scope" state, not a silent pass.
3. That state is the argument for authenticated fuzzing (the tracked v2 follow-up in `api-fuzz.yml`'s header) — it converts "we cannot test this" from an invisible gap into a named one.

## Context

Found while fixing three harness defects that were masking real results (#5763; #5714 before it). With those in, the lane reaches every money-path service — which is exactly when the quality of what it proves starts to matter.

Refs #5763, #5714

JiRaska:
## What

A green `api-fuzz.yml` job is not interchangeable across services, and nothing in the mechanism recorded the difference. `pentest` is minted off those greens by run URL (`attestations.yaml`, `by: ci-schemathesis`).

Measured on run 32103867665, all 26 fuzzed services, using schemathesis's own output:

| service | selected / total | auth-only | exercised |
|---|---|---|---|
| consent | 13 / 13 | 10 | **3** — and it found two genuine 5xx |
| billing | 5 / 5 | 4 | 1 |
| settlement | 1 / 1 | 1 | 0 |
| the other 23 | — | all | 0 |

Same shade of green for all of them. Twenty-five of twenty-six establish that the service boots and that auth is uniformly enforced — both worth knowing, neither a pentest. And it decays the wrong way: the 21-day TTL exists so a *dead* lane expires the attestation, but a lane that runs and bounces off 401 renews it forever while testing nothing.

## The shape of the fix

1. **`api-fuzz.yml` records what it exercised.** A new step parses the two lines schemathesis already prints — `Operations: N selected / M total` and `Authentication failed: K operations returned authentication errors` — into a paste-ready block per service (job summary + artifact). Numbers are the tool's, never recomputed. An **unparseable log is an error, never a zero**: a zero-filled record would read as "measured, empty surface", indistinguishable from the real thing, which is the exact failure this closes.
2. **`openbank-libs/governance/fuzz-coverage.yaml`** holds the per-service measurement. `state` is *derived* from the numbers by the gate — a stated state that disagrees with its own `selected`/`auth_errors` is an error, so it cannot be asserted.
3. **`readiness-attestation-format` R8**: `by: ci-schemathesis` may attest `pentest` only where `exercised = selected - auth_errors >= 3`.

### Why a count, and why 3

A **ratio cannot see this at all** — settlement is 1 of 1 = 100% of its surface, and so is consent at 13/13. The lane's v1 scope is the unauthenticated surface, so the only number separating "adversarially tested" from "bounced off 401" is how many operations answered something other than an auth error.

`>= 1` admits billing and every settlement-shaped service; one route is not a surface. **3** is the smallest count this repo has evidence for: consent sits exactly there and is the only service the lane has ever produced a real finding from. It is deliberately a floor under today's fleet, not an aspiration — the intended direction is upward once authenticated fuzzing (the v2 follow-up in `api-fuzz.yml`'s header) makes a real surface reachable.

### The failure mode is loud, not silent

A service below the floor is recorded as `auth-only-surface` and printed on every gate run as a named list, plus a `::warning::` from the lane itself. That converts "we cannot meaningfully fuzz this" from an invisible gap into the standing argument for v2. One summary line rather than 25 warnings per run — the alert-fatigue trade this repo has already paid for.

R8 is scoped to the fuzz lane. `ledger.pentest` comes from `ci-zap-baseline`, a different tool with a different scope, and is judged by the other rules (a self-test case pins that).

## Falsification — both directions, against the real tree

- **Known-bad**: a `settlement: pentest { by: ci-schemathesis }` line appended to the real `attestations.yaml` makes the gate **exit 1** and name the numbers (`selected 1/1, 1 auth-only, 0 exercised, floor 3`).
- **Known-good**: consent's real 13/13-with-10-auth-only still mints; the gate exits 0 on the tree as committed.
- **The floor value itself is what is enforced**: self-test cases at 2 exercised (must fail) and exactly 3 (must pass). If either stops discriminating, the number has become decoration.
- **The lane step** was executed against the *actual* downloaded logs of run 32103867665 (consent + settlement) and reproduced the committed file byte-shape; a rewritten-output log exits 1.
- 36 self-test cases pass; `run-gates.py --all` shows `readiness-attestation-format` PASS. The five other failures in that local run (`loki-rule-load-test`, `release-scope-mismatch`, `db-migration`, `api-contract`, `schema-compat`, `threat-model-...`) reproduce identically on a stashed clean tree — they need a PR diff base / local tooling this checkout does not have.

Closes #5769


JiRaska:
Review (read-only; no approval implied). The gate was executed and falsified, not just read.

## It provably prevents something — the negative control is real and runs in CI

`gates.yaml:2885` wires `--self-test` with `selftest_expect: pass`. Ran it:

```
ok [error] R8 SETTLEMENT-SHAPED: 1 op selected, all of it auth-only -> must NOT mint
ok [clean] R8 CONSENT-SHAPED: 13/13 with 10 auth-only = 3 exercised -> still mints
ok [error] R8 BOUNDARY: 2 exercised is below the floor
ok [clean] R8 BOUNDARY: exactly 3 exercised clears it
ok [error] R8: a fuzz-lane pentest with no coverage record at all is a bare green
ok [clean] R8 SCOPE: a non-fuzz lane is judged by the other rules, not the floor
self-test: all 36 cases passed (both directions)   exit 0
```

Then falsified it on /tmp copies of the governance files:

- consent `auth_errors: 10 → 11` (state becomes `auth-only-surface`) → **exit 1**, *"2 operation(s) were actually exercised (floor 3)"*.
- consent's coverage row deleted → **exit 1**, *"records no coverage for `consent` — a green job alone does not say what it exercised"*.

Both boundary directions and the absent-record case are genuinely enforced. Good.

## Finding 1: the coverage set is hand-kept — and the derivation already exists three files away

`fuzz-coverage.yaml:23-25` says it outright: api-fuzz produces a paste-ready block and *"**Copy that block here**"*. Nothing in CI diffs the committed file against the artifact.

The consequence is asymmetric and worth being precise about: absence is fail-closed **only along the attestation path**. `check-readiness-attestations.py:533` iterates over *attestation records*, not over services — so a service claiming `pentest: by: ci-schemathesis` with no coverage row is rejected (proven above), but **a service that is fuzzed and attests nothing is invisible**. The file could lose 20 entries and the gate stays green.

That is avoidable here, because `api-fuzz.yml:82-101` already **derives** its lane scope from `rules.yaml: money_path_services` + a 5-item legacy list + presence of `application.yaml`/`openapi.yaml`/port/postgres URL. I re-ran that derivation: **26 services, and `fuzz-coverage.yaml` has exactly those 26 keys — zero drift today.** Zero drift is exactly when deriving it is cheap. As written, the day a service becomes money-path it enters the lane and silently does not enter this file, and the gate reads as passing rather than unchecked — the same lesson the Pact drift-scope derivation was written for.

## Finding 2: the floor is over a DECLARED number, not a measured one

The numbers originate from schemathesis (`api-fuzz.yml:337-346`), but what the gate reads is whatever a human typed. `check-readiness-attestations.py:293-297` only requires `run:` to *start with* `http://`/`https://` — a string-shape check. Nothing fetches the run; a `run:` of `https://x`, or one pointing at an unrelated run, passes. So the attestation still asserts a coverage figure nobody verified, one transcription slip from being wrong.

Real mitigations that are present: `state:` is re-derived and a disagreement is an error (`:317-325`); `auth <= selected <= total` is enforced (`:299-305`); block-style YAML the parser would not ingest is an error. Those catch sloppiness, not fabrication.

Minimum ask: require `run:` to match `https://github.com/JiRaska/open-bank-oss/actions/runs/\d+`. Better: have the lane commit `fuzz-reports/<svc>-coverage.yaml` so the gate diffs the committed file against machine output instead of trusting a paste.

## Finding 3: staleness is caught in one direction only, and the two clocks disagree

Caught: an entry for a nonexistent module (`:252-257`), duplicates (`:327`), an old measurement (warning `:330-335`, hard error at 90 days when it carries an attestation `:561-566`).

**Not caught: real coverage moving up or down since the paste.** There is no re-measurement and no artifact comparison, so a service whose surface regressed to auth-only keeps its declared numbers and keeps minting `pentest` for up to 90 days.

And the two clocks are mismatched: `pentest` TTL is **21 days**, coverage max-age is **90** (`:150`). A pentest can be renewed four times off a coverage measurement four lane-generations old. The comment argues *"the lane runs twice a week; 90 days is ~26 missed runs"*, which reads as an argument for a **shorter** limit, not a longer one. Either justify 90 against the 21-day TTL or lower it.

## Finding 4: the workflow half of this PR is executed by no PR check

`api-fuzz.yml:34-46` triggers on `schedule` + `workflow_dispatch` only. This does **not** break the gate — its input is the committed YAML, not an artifact, so the PR is blocked correctly. The cost is that the 63 lines of new embedded Python at `:326-385` (two regexes, ANSI strip, short-name derivation, `sys.exit(status)`) first execute on the next scheduled Tuesday, after merge. A regex typo there is a red scheduled lane found days later. A self-test over a captured schemathesis log fixture would close that for very little work.

Minor: `AUTH.search(text)` is evaluated twice on `:349`.

## Worth stating in the PR body

On the real tree, **`consent` is the only one of 26 services that clears the floor — and `consent.pentest` (2026-07-28 + 21d) expired on 2026-08-18, two days ago.** So post-merge the fleet has zero live fuzz-lane pentest attestations, and 25 services report an empty unauthenticated surface. That is the honest and intended outcome of this change, but the PR does not say it, and it is the number a reader of the readiness board will ask about.

## Release scope and creep: clean

Touched paths are `.github/**` and `openbank-libs/governance/**`. `openbank-libs` has no `version.txt` and is in none of the packages in `release-please-config.json`, so `fix(governance)` — a releasing type — correctly releases nothing. No shipped service code hides under a non-releasing type. All six files serve the one change.

## Coordination note

**#5889 edits the same two files** (`check-readiness-attestations.py` and `attestations.yaml`), and both PRs append cases to the same `_self_test` list. That is the clean-merge-drops-a-list-entry shape. Whichever lands second: after merging `main`, re-run `--self-test` and confirm the case count reflects **both** PRs' additions, not just one.

JiRaska:
Heads-up from the other side of the same file — this PR and #5884 are individually correct and **collide semantically**, not just textually.

This PR implements the exercised-surface floor I left open in #5769, and it reads:

```python
logs = sorted(reports.glob("*-fuzz.log"))
```

#5884 turns the lane into **two passes** and writes two logs per service:

```
<svc>-fuzz.log             authz ON   — everything behind @RolesAllowed answers 401/403
<svc>-fuzz-authz-off.log   authz OFF  — the pass that actually reaches handlers
```

Verified the glob rather than assuming it:

```
files:                      openbank-consent-service-fuzz.log
                            openbank-consent-service-fuzz-authz-off.log
glob "*-fuzz.log" matches:  openbank-consent-service-fuzz.log     <- ON pass only
```

So once both land, **the floor is computed from the pass where 302 of 306 operations are auth-blocked by design** — the exact measurement from #5769 that motivated this gate. The floor would then be unsatisfiable for 24 of 26 services, or worse, would record "exercised surface ≈ 0" as a *measured fact* and gate on it. That is the successful-no-op shape this step's own comment says it exists to end, arriving through the interaction of two correct changes.

Suggested fix, whichever of us lands second — prefer the authz-OFF pass and fall back:

```python
logs = sorted(reports.glob("*-fuzz-authz-off.log")) or sorted(reports.glob("*-fuzz.log"))
svc  = log.name.removesuffix("-fuzz-authz-off.log").removesuffix("-fuzz.log")
```

The fallback matters: it keeps this gate working on a `main` that does not yet have the two-pass harness, so the two PRs stay independently mergeable in either order.

Worth stating plainly since it cuts the other way too: the floor is only meaningful **because** of the authz-OFF pass. Before it, 4 of 306 operations across the fleet were ever exercised — a floor over that would have disqualified almost everything. The two changes need each other; they just need to agree on which log is the measurement.

JiRaska:
## Fixed: the scope is derived now, and drift fails in both directions

Pushed `279745b` — new `.github/scripts/derive-fuzz-scope.py`, consumed by **both** sides.

**What was wrong.** `fuzz-coverage.yaml` enumerated its 26 services by hand while `api-fuzz.yml:82-101` derived the identical set inline. That is this repo's `pact-drift-check.yml` shape: a gate whose SCOPE is a hand-kept list of the thing it checks reads as **PASSING when the list is short, never as UNCHECKED**. `check_coverage()` only ever validated the entries that happened to be present — so deleting a heading made the gate *quieter*, not redder, while the lane went on fuzzing a service nothing recorded.

**What changed.** The set is computed once, from the inputs the lane actually uses (`rules.yaml: money_path_services` + the `LEGACY` list, filtered to what the harness can stand up — `application.yaml`, `openapi.yaml`, a derivable port, a `postgresql://` URL). Both consumers call it:

- `api-fuzz.yml` builds its matrix from `--github-output` (the inline heredoc is gone)
- `check-readiness-attestations.py` reconciles `fuzz-coverage.yaml` against `short_scope()`

Both directions are errors, and both name the service:

- **MISSING** — fuzzed, but nothing records what it exercised *(the direction that used to be invisible)*
- **STALE** — recorded, but the lane does not fuzz it, so the numbers can never be refreshed and decay into a permanent claim about a job that stopped running

### Prevent-proofs

Exit codes read from `$?` on redirected runs — never through a pipe.

| state | exit | message |
|---|---|---|
| healthy (unchanged tree) | **0** | `api-fuzz scope: 26 service(s) derived; fuzz-coverage.yaml records 26` |
| delete `ledger` from coverage | **1** | ``` `ledger` is in the api-fuzz scope (derive-fuzz-scope.py) but has no coverage entry ``` |
| add an `onboarding` entry | **1** | ``` `onboarding` has a coverage entry but is NOT in the api-fuzz scope ``` |
| drop swift from `money_path_services` | **1** | scope goes **26 → 25 with no edit to the script**, names ``` `swift` ``` as STALE |

The last row is the one that matters: the derived scope **follows** its source. Adding or removing a service is red in the same commit, with no hand-edit anywhere.

Also verified:

- `--self-test` **exit 0, 40/40** (4 new scope cases: clean, both drift directions, and empty-file-vs-non-empty-scope; each asserts the error *names* the service, so a checker that flagged everything could not pass)
- derived set reproduces the committed hand-list **exactly** — 26/26, symmetric difference empty in both directions, i.e. this PR changes no behaviour today, only what happens when it drifts
- `ruff --select E9,F,B` clean on both scripts (the `python-lint` gate's own invocation)
- `actionlint` finding set on `api-fuzz.yml` is **identical to the pre-change file** — compared as finding *sets*, base vs branch, not by filtering the output: one pre-existing `SC2034` at line 137, unrelated and untouched
- both workflow paths exercised under `bash`: `${OVERRIDE:+--override $OVERRIDE}` set → 2 services, unset → 26

`fuzz-coverage.yaml`'s header now says the service set is not maintained there, so the next reader does not hand-edit a heading to quiet the gate.

JiRaska:

|---|---|
| #5890 | 3934 |
| #5849 | 3853 |
| #5875 | 3735 |
| #5967 | 3417 |

JiRaska:
Rebased onto current `main`. One textual conflict and one real one.

**Textual** — the self-test's case count. This branch adds `scope_cases`; main moved its constant
from 2 to 5. Resolved as the union: `len(cases) + 5 + len(scope_cases)`.

**Real, and worth reading** — main's `EXERCISE_REF_DEBT no longer exempts a baselined entry` case
asserts `if errors:` after injecting an exemption. This branch's new R8 rule reports a hard error
whenever `fuzz-coverage.yaml` is absent, which in a fixture directory it always is. So after the
merge that case failed on a **missing file** rather than on the exemption it exists to test — a
green-to-red flip that says nothing about either change on its own. Fixed by giving the fixture an
empty coverage file, which is the right value here: the fixture has no services, so the derived scope
is empty and the two reconcile.

That fix could easily have made the case vacuous, so it was checked in the failing direction:

```
exemption injection removed  -> case fires, self-test exit 1
everything in place          -> 42/42 both directions, gate exit 0
```

`#5849` and `#5889` no longer conflict with each other either — both are clean against `main` and
against each other as of these pushes.

JiRaska:
Superseded — an R8 block solving the same problem (#5769) has since landed on `main`: a CI-minted `pentest` attestation must carry `ops: N` and is rejected below `PENTEST_OPS_FLOOR` (5), with `not_fuzzable: <reason>` as the declared escape.

Diffed the two implementations rather than closing blind. They differ in two ways that matter, and the landed one is the weaker on both:

- **Provenance of the number.** `main` reads a hand-written `ops: N` out of `attestations.yaml` — a self-asserted count sitting next to the claim it justifies. This branch derived it from the fuzz lane's own coverage artifact (Selected minus auth-blocked).
- **Staleness.** `grep -c 'coverage_max_age|re-measure'` over the script on `main` returns 0; this branch bounded the age of the measurement and demanded a re-measure before re-attesting.

Both gaps are now recorded in #9673 so they survive this close. Rebasing this branch is not the way to get them — it is 3 weeks stale and conflicts structurally with the R8 block that actually landed, so it would need rewriting against the current implementation regardless. Closing.

Issue #5728, opened 2026-08-19
ADR-0249 D4: spend-reservation audit events + grantor notification are unbuilt


JiRaska:
## What ADR-0249 D4 claims

> Every grant, reservation, confirmation and revocation is an audit event... the grantee's first use of a new authority notifies the grantor.

## What's actually implemented (verified 2026-08-19, see PR #5727 for the audit that found this)

- **Card issuance/revocation**: real audit events (`CUSTOMER_DELEGATED_CARD_ISSUED` etc.) via `CardDelegationEventConsumer.kt`. ✅
- **Spend reservations** (`SpendReservationService.kt`/`SpendReservationResource.kt`, D3's reserve/confirm/release path): **emit no audit event at all**. The claim "every... reservation, confirmation and revocation is an audit event" does not hold for this path.
- **Grantor notification**: no service calls `notification-service` when a delegate first uses a new authority (spend reservation or otherwise). This half of D4 is entirely unbuilt, not partially built.

## Why this matters

This is the primary compliance-facing decision record for enforced delegated spending on a banking product. It currently overclaims audit coverage for real money-movement events (spend reservations), which is a genuine gap between what's documented as guaranteed and what actually happens.

## Scope of a fix (not attempted here — needs a design call)

1. Add an audit event emission to `SpendReservationService`'s reserve/confirm/release paths, following the same event shape/publishing convention `CardDelegationEventConsumer` already uses for card actions.
2. Decide whether "grantor notification on first use" should reuse the `DelegationNotificationConsumer` pattern (openbank-app session, PR #5661/#5700) or needs its own trigger — first-use detection is a different signal than the lifecycle events (offered/accepted/revoked/expired) that consumer already handles.

Neither of these was attempted as part of the audit PR (#5727) — that PR only corrected the ADR's frontmatter/delivery-note to accurately reflect this gap instead of silently overclaiming coverage.

JiRaska:
## Re-verified on `origin/main`, 2026-08-26: still entirely unbuilt, and the gap is wider than "no audit event"

`openbank-delegation-service/src/main/kotlin/com/openbank/delegation/application/usecase/SpendReservationService.kt`
— `reserve` / `settle` / release paths, read in full:

```
$ git show origin/main:.../SpendReservationService.kt | grep -cE 'Audit|audit|Notif|notif|outbox|Outbox|publish'
0
```

Not "no audit event on the reservation path" — **zero audit or notification wiring anywhere in the
[...]

Issue #5338, opened 2026-08-17
audit-service does not subscribe to openbank.sca.events — SCA enrollment never audited


JiRaska:
## What

Found while adding `sourceService` to sca-service events (#5256/PR #5337). `openbank.sca.events` is absent from BOTH `TopicAttribution` (`openbank-audit-service/.../EventAttribution.kt`) and the consumed-topics list in audit-service's `application.yaml` — audit-service does not subscribe to this topic at all.

## Why it matters

SCA device enrollment (`DEVICE_ENROLLED`) events are never recorded in the audit trail. Given sca-service is money-path (PSD2 strong customer authentication), this is a real compliance gap, not just a metrics nicety — an auditor or regulator asking "when was this device enrolled for SCA" has no row to find.

## What would close this

1. Confirm whether this is a genuine oversight or a deliberate scoping decision (some topics may be intentionally excluded — check if there's a documented reason).
2. If genuine: add `openbank.sca.events` to audit-service's consumed-topics list and a `TopicAttribution` entry, then verify a real enrollment event lands in `audit_entries`.

Refs #5256, #5337

PR #5857, opened 2026-08-20
feat(delegation): emit audit events for spend reserve, confirm and release

JiRaska:
## What ADR-0249 D4 claims, and what was true

> Every grant, reservation, confirmation and revocation is an audit event.

Card issuance/revocation was the only half built. The D3 reserve/confirm/release path —
which moves real money on a delegated authority — emitted **nothing at all**, while ADR-0249
is the compliance-facing record claiming full coverage. Found by the audit in #5727.

This PR builds the **audit-events half**. The grantor-notification half of D4 stays unbuilt
(it needs a design call — see below), hence `Refs #5728`, not `Closes`.

## What is emitted

Three new events in `openbank-delegation-service/src/main/.../domain/event/DelegationEvents.kt`:
`SpendReserved`, `SpendConfirmed`, `SpendReleased`. Each is written to `delegation_outbox`
**inside the same transaction** as the reservation insert / the settle compare-and-set, by the
rule `DelegationRepositoryImpl` already follows for the grant lifecycle. A publish after the
commit would be a second failure point on a money path: the reservation real, the audit trail
absent.

Design calls, each documented in place:

- **`aggregateId` is the GRANT, not the reservation.** `OutboxKafkaHeaders.partitionKey` returns
  `aggregateId`, so keying on the grant keeps a reserve and a revoke of one grant ordered on one
  partition. Keying on the reservation would let a consumer see spend against an authority whose
  revocation it has not applied yet. The reservation keeps its own id as a field.
- **PascalCase `eventType`, not SCREAMING_SNAKE_CASE.** These share `openbank.delegation.events`
  with the eight lifecycle events, whose `eventType` is PascalCase and whose in-tree consumers
  (`CardDelegationEventConsumer` and its account-service twin) match those strings literally.
  `CUSTOMER_DELEGATED_CARD_ISSUED` belongs to a different producer and a different topic
  (`EdgeAuditPublisher` / `openbank.customer.audit`). A second idiom in one stream would make the
  topic's convention unstateable.
- **`sourceService = "delegation-service"` on all three** — the field `AuditConsumer` falls back
  to `?: "unknown"` for, silently, with no error (#3994/#5256).
- **`occurredAt` is this call's clock reading, never a default** — the `Instant.EPOCH` shape
  (#3874/#3883).
- **Emitted only where state changed**: not on an idempotent replay, not on a refusal, not on the
  losing side of a confirm/release race. One reservation yields at most one settlement event.

The AsyncAPI contract is updated in the same change; `check-event-contract-code-agreement.py`
fails otherwise, and did until it was.

## The half that made the emission real

audit-service was **not subscribed to `openbank.delegation.events` at all** — absent from both
`application.yaml`'s topics list and `TopicAttribution`. So no delegation event of any kind has
ever reached the audit trail, including the eight existing grant lifecycle events. Emitting the
new events without fixing that would have been an outbox row nobody reads.

Three parts, two of which are silent on their own: the topics list, the `TopicAttribution` row
(so the eight older events resolve TOPIC-sourced rather than `"unknown"`), and the KafkaUser
**Read ACL** — without which the subscription is ACL-denied at runtime with nothing in this repo
disagreeing. Same shape as #5338, where the identical gap meant SCA device enrollment was never
audited.

## Verified by effect

`SpendReservationOutboxWriteIT` (new) drives the **real REST endpoint** (RestAssured +
`@TestSecurity`) against a real Postgres and reads `delegation_outbox` with **plain JDBC** — the
`LendingOutboxWriteIT` / `ConsentRevocationOutboxIT` pattern, because a mocked repository cannot
prove atomicity. It asserts CONTENT, not that a publisher was called: event type, aggregate type,
both party ids, the reservation id, the amount and currency, `sourceService`, and `occurredAt`
by **recency** (`isBetween(before, now)`) — never `isNotNull()`, which passes against 1970.

**Falsification — each protection removed, and the tests confirmed red:**

| removed | tests that went red |
|---|---|
| the reserve emission | 3 (`SpendReservationOutboxWriteIT`) |
| the settle emission | 2 (`SpendReservationOutboxWriteIT`) |
| correct `sourceService` (set to `"unknown"`) | 7 across all three new/extended suites |
| real `occurredAt` (set to `Instant.EPOCH`) | 3, including the recency test |
| the audit-service topic subscription | `TopicAttributionCoverageTest` |

`openbank.outbox.dispatch-enabled` is already `true` in this service's `application.yaml`, so the
rows dispatch rather than sitting at `attempt_count = 0`.

## Verification

- `:openbank-delegation-service:` `ktlintCheck` + `detekt` green; `test` **127 tests, 0 failures,
  1 skipped** — the skip is `DelegationEventPactBrokerProviderVerificationTest`, broker-gated by
  `@EnabledIfSystemProperty(pactbroker.url)` and pre-existing.
- `:openbank-audit-service:` `ktlintCheck` + `detekt` + `test` green.
- `run-gates.py --all` with `PR_DIFF_BASE=origin/main`: **153 PASS**. The 3 residual are
  environmental/pre-existing and untouched by this change — `api-contract` needs `sudo` to install
  `oasdiff` locally (and no `openapi.yaml` changed here), `release-scope-mismatch` needs `PR_TITLE`,
  `loki-rule-load-test` is unfalsified on `main`.

## Money path

`openbank-delegation-service` **is** in `rules.yaml: money_path_services` — so this needs 2
approvals per ADR-0030. Its threat model exists (`docs/threat-models/openbank-delegation-service.md`).
The enforced threat-model-diff gate was run rather than guessed:

```
python3 openbank-infra/scripts/check-threat-model-diff.py --base origin/main --enforce
threat-model-diff gate: no money-path trust-boundary change without a threat-model update.  (exit 0)
```

It does not classify this as a trust-boundary move, which reads right: no new ingress, no new
authority, no change to who may reserve. It only records what already happens.

## What the grantor-notification half still needs (NOT built here)

Deliberately out of scope — it is the part that needs a decision, not code:

1. **What "first use of a new authority" means.** First *reserve* under a grant? First *confirmed*
   spend? First use of each capability, or of the grant as a whole? A reserve that is later released
   moved no money — notifying on it teaches the grantor to ignore the alert.
2. **Dedup across repeat uses, and where the "already notified" state lives.** A per-grant flag on
   the aggregate is the obvious answer and is a schema change; deriving it from the event history
   is not available, since this topic has 7-day retention and is not compacted.
3. **Which template, and whether it is opt-out.** `notification-service` has no template for this,
   and no service in either delegation-service or card-issuance-service calls it today.
4. **Whether it reuses `DelegationNotificationConsumer`** (#5661/#5700) or needs its own trigger.
   First-use is a different signal from the lifecycle events that consumer handles
   (offered/accepted/revoked/expired) — it is derived from spend, not from a transition.

The events this PR adds are the natural trigger for whatever is decided: `SpendReserved` and
`SpendConfirmed` now exist on a topic, keyed by grant, and audit-service is subscribed to it.

Refs #5728

### DIFF
--- a/openbank-contracts/openbank-delegation-service/asyncapi.yaml
+++ b/openbank-contracts/openbank-delegation-service/asyncapi.yaml
@@ -78,6 +78,12 @@ channels:
         $ref: '#/components/messages/DelegationRenounced'
       DelegationExpired:
         $ref: '#/components/messages/DelegationExpired'
+      SpendReserved:
+        $ref: '#/components/messages/SpendReserved'
+      SpendConfirmed:
+        $ref: '#/components/messages/SpendConfirmed'
+      SpendReleased:
+        $ref: '#/components/messages/SpendReleased'
 
 operations:
   publishDelegationEvents:
@@ -96,6 +102,13 @@ operations:
       read model so that enforcing a delegation never requires a synchronous call to
       delegation-service. When either merges, add it here; a consumer absent from this list is a
       consumer nobody has agreed to keep compatible.
+
+      **`openbank-audit-service` subscribes as of #5728** (ADR-0249 D4). It was not subscribed to
+      this topic before, so no delegation event of any kind reached the audit trail — which is why
+      adding the spend-reservation events below without the subscription would have been an outbox
+      row nobody reads. It consumes every message on this channel generically (no per-type logic),
+      so a new event type does not break it; what it needs is `sourceService`, which the three
+      Spend* messages carry and the eight lifecycle messages do not yet.
     messages:
       - $ref: '#/channels/delegationEvents/messages/DelegationOffered'
       - $ref: '#/channels/delegationEvents/messages/DelegationActivated'
@@ -105,6 +118,9 @@ operations:
       - $ref: '#/channels/delegationEvents/messages/DelegationReinstated'
       - $ref: '#/channels/delegationEvents/messages/DelegationRenounced'
       - $ref: '#/channels/delegationEvents/messages/DelegationExpired'
+      - $ref: '#/channels/delegationEvents/messages/SpendReserved'
+      - $ref: '#/channels/delegationEvents/messages/SpendConfirmed'
+      - $ref: '#/channels/delegationEvents/messages/SpendReleased'
 
 components:
   messageTraits:
@@ -247,6 +263,28 @@ components:
                 `{"code":"CZK"}` while every consumer reads this as text. Matches the fleet's
                 event convention (AccountEvents, LedgerEvents).
 
+    # ADR-0249 D3/D4 — one unit of delegated spend. `aggregateId` is still the GRANT, not the
+    # reservation: the partition key is `aggregateId`, and keying on the reservation would let a
+    # consumer see spend against an authority whose revocation it has not applied yet.
+    reservation:
+      type: object
+      required: [reservationId, amount]
+      properties:
+        reservationId:
+          type: string
+          format: uuid
+          description: The unit of spend. Unique; NOT the partition key.
+        amount:
+          type: object
+          required: [amount, currency]
+          properties:
+            amount: { type: string, format: decimal }
+            currency:
+              type: string
+              minLength: 3
+              maxLength: 3
+              description: ISO 4217, flat — same `EventMoney` shape as `perTransactionLimit`.
+
   messages:
     DelegationOffered:
       name: DelegationOffered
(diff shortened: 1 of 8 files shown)


JiRaska:
## Review — live and substantively correct; checked the money-path attribution questions specifically

**Staleness:** 265 commits behind `main` but `MERGEABLE`. Not a duplicate of anything merged.

**First, a correction to an earlier reading of this branch.** A 2-dot diff of it was reported as appearing to delete `AUDIT_ANCHOR_SIGNER` / `AUDIT_ANCHOR_SIGNING_REQUIRED` / `AUDIT_ANCHOR_KMS_KEY_ID`. Re-measured with 3-dot (`git diff origin/main...origin/fix/spend-reservation-audit-events`): **there is no such deletion**, in either the gitops file or `openbank-audit-service/src/main/resources/application.yaml`. The branch's only edit to that file is appending `openbank.delegation.events` to the consumer `topics:` list at line 148; the anchor block at ~196 is untouched. Nothing to fix.

**Attribution comes from server-side state, not the request body — verified.**
- `DelegationEvents.kt:187` declares `internal const val DELEGATION_SOURCE_SERVICE = "delegation-service"` and lines 206/222/238 use it as the property default. It is a compile-time constant, not a caller-supplied field, so `source_service` — which is chain-hashed into `record_hash` and therefore uncorrectable after the fact — cannot be poisoned from outside. It also matches the module directory name, so it satisfies the convention gate proposed in #5967 without a baseline entry.
- `SpendReservationService.kt`: `grantorPartyId` / `granteePartyId` are read off the `DelegationGrant` returned by `loadForGrantee(...)`, i.e. the persisted grant the caller was authorised against, never off `ReserveSpendCommand`.

**Outbox dispatch is on.** `openbank-delegation-service/src/main/resources/application.yaml:183-184` has `outbox: dispatch-enabled: true`, so these events actually leave the process. (Worth stating explicitly because the platform default is `false` and the failure is silent — `attempt_count` simply stays 0.)

**Atomicity is real, not a post-commit publish.** `SpendReservationRepositoryImpl` takes `DelegationOutboxRepository` and calls `outboxRepository.persistInTransaction(outboxMessage(auditEvent(...)))` inside the same reactive transaction as the insert / the compare-and-set, and only on the branch where the CAS actually won (`if (count > 0L)`), returning `nullItem()` otherwise. That is the correct shape: a losing confirm/release race records no settlement, and a rolled-back reservation leaves no audit row. `SpendReservationOutboxWriteIT` drives it through real HTTP, which is the only way this is provable.

**`occurredAt` is a real clock reading**, passed from the use case's `now`, not a data-class default — the `Instant.EPOCH` trap is explicitly avoided and the comment says so.

**Two things worth the author's judgement (neither blocking):**

1. **The events record whose authority was used, not who used it.** `SpendReserved`/`SpendConfirmed`/`SpendReleased` carry `grantorPartyId` and `granteePartyId` but not `callerPartyId`. On the customer path caller == grantee so nothing is lost, but `CallerPartyId` is `typealias CallerPartyId = UUID?` (`DelegationUseCases.kt:28`) — a null caller is representable, and that is the operator/M2M path. For a money-path audit record the actor is usually the field a reviewer wants first, and it is unrecoverable later for the same chain-hash reason as `source_service`. Consider adding it now rather than as a v2 of the event.
2. **`error("RESERVED is not a settlement target")`** in `settlementEvent` is defensible as documented-unreachable, but it is an `IllegalStateException` on a money path and `libs-runtime` maps `IllegalArgumentException`, not `IllegalStateException`, to 400 — so if it ever *is* reached it is a 500 with a stack trace. Given the comment argues it cannot be reached, that is acceptable; flagging it only so the choice is deliberate.

**Also checked:** the audit-service side is complete rather than half-wired — `EventAttribution.kt:91` adds the `openbank.delegation.events` → `delegation-service` topic fallback row, and `kafka-audit-mtls.yaml` adds the matching `Read, Describe` ACL. Without that ACL the subscription would have been denied on arrival, which is the usual way this kind of change ships looking complete and is not. Both halves are here.

Verdict: **ready**, with (1) worth an answer before merge. Not approving — review comment only.

JiRaska:
Rebuilt on current `origin/main` after #6047 landed, and **narrowed by subtraction: the whole
audit-service half of this PR is already on `main`.**

Verified per file rather than assumed — each of the three conflicting files is now byte-identical
between this branch and `origin/main`:

```
openbank-audit-service/.../EventAttribution.kt            same as main
openbank-audit-service/src/main/resources/application.yaml same as main
openbank-infra/gitops/components/audit/kafka-audit-mtls.yaml same as main
```

`main` already subscribes `openbank.delegation.events`, already grants the Read ACL, and already
carries the `EventAttribution` entry — #6047 brought all three. Re-applying this branch's versions
would have reverted the four money-path topics #6047 added alongside them
(`ledger.journal.posted`, `sdd.event`, `interest.accrual.event`, `fraud.hold.changed`), because the
subscription is a single comma-separated line and taking one side of it drops the other's entries.

What remains, and what this PR now is: the eight delegation-service files — the
`SpendReserved`/`SpendConfirmed`/`SpendReleased` events themselves, their outbox write, the
asyncapi contract, and the tests. 3-dot delta against `main` is exactly those eight.

New head `e9479c02f`, signed.

JiRaska:
Absorbed by #8334 (merged 2026-09-06): its branch merged this one in (3fa2915b0a) and the combined change landed on main, including the SpendReserved/SpendConfirmed/SpendReleased audit emission, the appendAuditEvent persistence in the same transaction, and the asyncapi declarations. Nothing in this PR is left unmerged — closing as superseded.

Issue #5838, opened 2026-08-20
governance: make D5 audit anchors independently verifiable


JiRaska:
## Problem

ADR-0031 D5 has a hash chain and HMAC-signed checkpoints, but HMAC is not independently verifiable. The published 4/5 control maturity correctly leaves D5 partial.

## Outcome

Use an asymmetric KMS/cosign-compatible anchor signer under workload identity and include AI-attribution coverage in release evidence. Do not advance the public control score until the implementation and production evidence are verified.

## Acceptance criteria

- Anchor signatures can be verified using public-key material without the signer secret.
- The signer fails closed for an unavailable or invalid key.
- Release evidence reports AI-attribution coverage and does not treat missing attribution as success.
- Unit, HTTP and deployment-facing proofs cover signing, verification and failure modes.
- D5 is marked built only after the deployed evidence is independently verifiable.

Refs ADR-0031 D5.

PR #5858, opened 2026-08-20
feat(audit): sign audit anchors with an asymmetric, fail-closed signer

JiRaska:
## What

ADR-0031 D5 signed audit anchors with **HMAC-SHA256**. Verifying an anchor therefore required the signing secret, so the set of parties able to *check* an anchor was exactly the set able to *forge* one — the control was tamper-evident to its own operator and to nobody else. This replaces it with an asymmetric signer, makes capture fail closed, and makes release-evidence AI-attribution report its own absence.

Refs #5838. **`Refs`, not `Closes`** — acceptance criterion 5 is deliberately not met here (see below).

## The signing path that was there

- `openbank-audit-service/.../port/out/AnchorSigner.kt` — port with `sign` **and** `verify`, the latter only meaningful to a holder of the key.
- `.../infrastructure/signing/LocalHmacAnchorSigner.kt` — `@DefaultBean`, HMAC-SHA256 over `openbank.audit.anchor.signing-key`, defaulting to `CHANGE_ME_LOCAL_DEV_ONLY`.
- `.../application/AuditAnchorService.captureAnchor()` — wrapped `signer.sign(...)` in `runCatching`, logged a warning, and **stored the checkpoint unsigned anyway**.

## Mechanism chosen, and what it mirrors

An **OpenBao `transit` ECDSA P-256 key under workload identity** — not a new key-handling scheme:

- The pod's projected ServiceAccount token is exchanged at `auth/kubernetes/login`, then the digest is signed at `transit/sign/<key>`. This is the same OpenBao + Kubernetes-auth flow `openbank-document-service`'s `OpenBaoClientSignatureAdapter` already uses for client-PKI signing (and `svidMint.ts` for `pki-agent`), with a `transit` key instead of a PKI mount.
- The private key never leaves OpenBao.
- The signature is ASN.1/DER `SHA256withECDSA`, base64 — the shape `cosign` produces and consumes for a `hashivault://` transit key, so an anchor is checkable by `cosign verify-blob` as well as by the in-repo verifier.
- `GET /api/v1/audit/anchors/public-key` publishes the SPKI PEM (503 when it cannot be published, never 200 with a null body).

I did **not** hand-roll `cosign attest`/`verify`, and note the repo's own warning that `cosign attest` is additive — a green `verify-attestation` can be about an earlier envelope. Nothing here relies on that; verification is a direct signature check over the anchor digest.

## How public-only verification is proved

`AnchorSignatureVerifier` is a pure object in the domain layer: bytes + a PEM string, no key held, no connection opened. It is structurally incapable of touching a private key.

In the tests the private key exists only inside a helper (`signWithAThrowawayKey` / `signAnchorWithAThrowawayKey`) that returns **only** the signature and the SPKI PEM. It is unreachable from every test body, so the passing assertions are made with the private key genuinely out of reach — not merely unused.

## Fail-closed evidence

Capture no longer tolerates a signing failure; there is no dev fallback and no opt-in flag. An environment that does not anchor sets `openbank.audit.anchor.enabled=false` (as `%test` does); one that does must provision the key.

The boot-time gate is `@Startup`-annotated on purpose. `@ApplicationScoped` is lazy, so a guard on a bean nothing has touched never runs at boot — `PdfBoxPadesSealAdapter` already shipped a warning here that never once appeared in a pod log. `OpenBaoTransitAnchorSignerStartupAnnotationTest` pins the annotation, because no behavioural test can see its absence.

Proofs, negative-first:

| Case | Proof |
|---|---|
| unavailable key ⇒ signing fails | `OpenBaoTransitAnchorSignerTest`: missing SA token ⇒ `AnchorSigningException` |
| invalid/unreachable backend ⇒ signing fails | same, against a port that is never listening |
| enabled + no workload identity ⇒ **boot** fails | `boot fails closed …` via the real `StartupEvent` observer |
| signing failure ⇒ **no row** | `AuditAnchorFailClosedTest`: `coVerify(exactly = 0) { anchorRepo.save(any()) }` |
| tampered anchor ⇒ BROKEN | digest recomputation invalidates the signature |
| rewritten chain ⇒ BROKEN | attested head no longer matches |
| no public material ⇒ **UNVERIFIABLE**, never verified | its own status and its own counter |
| routes are actually served | `AuditAnchorEndpointIT` over real HTTP |
| capture refused over HTTP, table stays empty | `AuditAnchorEndpointIT`: 503 `SIGNER_UNAVAILABLE`, then `GET /anchors` is empty |

The HTTP proof is a `@QuarkusTest` on purpose: the existing anchor endpoint coverage is reflection-based, and reflection over a resource class cannot tell a served route from an unserved one — this repo has shipped a `@Path` that bound to a stray top-level function while the class-level test stayed green.

**Falsification.** Three defects were injected and each went red:

- restoring `runCatching` around `signer.sign` → 2 fail-closed tests red;
- treating a missing public key as verified → the UNVERIFIABLE test red;
- removing `@Startup` → the annotation test red.

The release-evidence self-test was falsified the same way: reverting the bundle to the old bare `contributors` list produced 7 failures.

## Criterion 3 — release evidence

`build-release-evidence.sh` emitted `ai_attribution.contributors` and nothing else, so "no commits could be measured" (shallow clone, wrong module path, first release) and "commits carry no AI attribution" both rendered as `[]` — an empty success. It now reports coverage with three distinct states:

- `REPORTED` — commits classified, `coverage_pct` meaningful (denominator is **every** commit in range, not only attributed ones);
- `NO_ATTRIBUTION` — commits found, none attributed (a finding);
- `UNAVAILABLE` — the range could not be measured, `coverage_pct: null` (**not** a passing result).

Covered by `.github/scripts/test-release-evidence-attribution.sh`, which follows the existing `test-*.sh` convention in that directory.

## What criterion 5 still needs — NOT done here

D5 stays 🟡 partial and **no control score is advanced, no maturity number is edited**. This is implementation, not deployed evidence. Still owed:

1. Provision the `transit` ECDSA P-256 key and the `audit-service-anchor` Kubernetes-auth role in OpenBao, plus the gitops wiring for the audit-service pod (no infra manifests are touched in this PR).
2. Deploy, let a real capture run, and fetch a published anchor.
3. Have a third party verify that anchor from `GET /api/v1/audit/anchors/public-key` alone — ideally with `cosign verify-blob`, i.e. a tool that shares no code with this repo.
4. Only then mark D5 built and revisit the published maturity.

Until step 3 has actually happened, marking D5 built would repeat the defect class this repo already carries: a decision record crediting a control that was never deployed.

## Verification run

Serialised, one Gradle invocation at a time:

- `:openbank-audit-service:ktlintCheck` — pass
- `:openbank-audit-service:detekt` — pass (`verifyAnchors` complexity and `login` throw-count both refactored rather than suppressed)
- `:openbank-audit-service:test` — **160 tests, 0 failures, 0 skipped** (skipped count checked explicitly, and re-run with `--no-build-cache` so the task genuinely executed rather than being restored from cache)
- `:openbank-audit-service:quarkusBuild` — pass (the two-port bean resolves; CDI failures surface only here)
- `bash .github/scripts/test-release-evidence-attribution.sh` — pass
- `bash .github/scripts/check-adr-registry.sh` — OK (ADR body edit changes no derived line; `gen-index.sh` produced no diff)

## API contract

`openapi.yaml` `info.version` 1.6.0 → **1.7.0** (additive: two new paths, one new schema, a new required response field). `version.txt` is untouched — release-please owns it.


JiRaska:
## Not resolved — superseded by #5844, and the conflict is a port redesign

Picked this up in a conflict-resolution sweep. I did not push anything, because resolving it would not be a merge — it would be a re-implementation.

### What happened

This branch and #5844 built the same capability in parallel, and #5844 won the race by 48 minutes:

| | this branch (`8fb15cdb5`) | `main` |
|---|---|---|
| committed | 2026-08-20 11:24 | #5844 merged 2026-08-20 12:12 |
| backend | OpenBao Transit | AWS KMS |
| port | `suspend fun sign(digest): String` + separate `AnchorPublicKeyResolver` | `fun sign(digest): AnchorSignature`, `verify(...)`, `verificationKeyPem(...)` on the port |

`main` has since hardened that path four more times — #5926 (runtime wiring), #5944 (optional `kms-key-id` so the service boots), #6186 (governance), #6193 (a CI guard for the `@ConfigProperty` shape that stopped audit-service booting) — plus `openbank-infra/aws/envs/sandbox-platform/audit-anchor-kms.tf`.

### Why that makes this unmergeable rather than conflicted

The ten conflicting files are not textual drift. The two branches define **incompatible shapes for the same interface**: signatures differ (`suspend`/`String` vs sync/`AnchorSignature`), and verification lives in a separate resolver here but on the port itself on `main`. Resolving the conflict means rewriting `OpenBaoTransitAnchorSigner` onto `main`'s port and re-deciding whether the anchor backend is OpenBao or KMS — with `main`'s KMS route now carrying five merged PRs and Terraform behind it.

### Recommendation

**Close as superseded** — unless the OpenBao backend is wanted *instead of* KMS, in which case that is a deliberate reversal that deserves its own PR against `main`'s current port, not a conflict resolution buried in a merge commit.

I have deliberately not closed it. The choice of signing backend for audit anchors is a security-relevant architecture decision with a threat model attached, and "whichever PR merged first" is not the right way to settle it. If KMS is in fact the intended answer, closing is a one-liner and I am happy to do it on your word.

One thing worth salvaging either way: this branch's `AnchorSignerBootIT` / startup-annotation test covers the fail-closed boot path. Worth a glance to confirm `main`'s equivalent is as strict before this branch goes away.

JiRaska:
## Review — **superseded by #5844 (merged 2026-08-20)**; recommend close and re-propose as a delta

**Staleness first.** 265 commits behind `main`, `CONFLICTING`. #5844 *"feat(audit): add KMS-backed audit anchors"* merged at 2026-08-20T10:12:39Z — 47 minutes after this PR was opened — and it landed the same decision (asymmetric, non-forgeable audit-anchor signing) on a different backend.

**What `main` has now that this branch does not:**
```
openbank-audit-service/src/main/kotlin/com/openbank/audit/infrastructure/signing/AnchorSignerProducer.kt
openbank-audit-service/src/main/kotlin/com/openbank/audit/infrastructure/signing/AwsKmsAnchorSigner.kt
openbank-audit-service/src/main/kotlin/com/openbank/audit/infrastructure/signing/LocalHmacAnchorSigner.kt
```
and in `openbank-audit-service/src/main/resources/application.yaml:200-204`:
```yaml
signer: ${AUDIT_ANCHOR_SIGNER:hmac}
signing-required: ${AUDIT_ANCHOR_SIGNING_REQUIRED:false}
kms-key-id: ${AUDIT_ANCHOR_KMS_KEY_ID:}
signing-key: ${AUDIT_ANCHOR_SIGNING_KEY:CHANGE_ME_LOCAL_DEV_ONLY}
```

This branch replaces that whole block with an OpenBao `transit` block and ships `OpenBaoTransitAnchorSigner.kt` while **deleting** `LocalHmacAnchorSigner.kt` and never having had `AnchorSignerProducer` or `AwsKmsAnchorSigner`. Merging it does not add a second backend — it reverts #5844's selection mechanism and its KMS adapter. That is not visible as a conflict everywhere it happens; it is the "clean merge that deletes content" shape.

**Do not read this as the work being wrong.** One design choice here is *stronger* than what landed, and I would like to see it survive: this branch has **no dev fallback at all**, on the argument that an unavailable key must make signing fail rather than silently produce a self-signed anchor. `main` ships `signing-required: false` by default, which means the deployed default today is exactly the "signs with a local secret, verification requires the forging secret" posture the PR description objects to. That objection is still live against `main` and is worth its own issue.

**Recommended path:** close this PR, and re-propose against current `main` as a small delta that (a) adds `OpenBaoTransitAnchorSigner` as a third `signer:` value behind the existing `AnchorSignerProducer`, and (b) argues the `signing-required` default separately, with evidence about what the deployed value actually is. `AnchorSignatureVerifier.kt` and the `OpenBaoTransitAnchorSignerStartupAnnotationTest` (the `@Startup`-on-a-config-gate check) are both worth carrying over verbatim.

**Not checked**, because the above decides it: the release-evidence script changes, the ADR-0031 edit, and the openapi.yaml `info.version` — note that if this is re-proposed, the spec version must be re-read off live `main` at that time, since #5844 may already have consumed the next number.

Not approving and not merging — review comment only.

JiRaska:
## Closing: superseded on `main`

Checked against live `origin/main` (`8b1868746`) after `d9c831c0a` (#6206) and `87b9c00ca` (#5944) merged. Every capability this PR set out to add is now on `main`, implemented against AWS KMS instead of OpenBao Transit.

### Claim-by-claim, measured

**"Asymmetric"** — `openbank-audit-service/.../infrastructure/signing/AwsKmsAnchorSigner.kt` exists on `main`:

```
/** Asymmetric audit-anchor signer backed by a dedicated AWS KMS ECC_NIST_P256 key. */
class AwsKmsAnchorSigner(private val kms: KmsClient, override val keyId: String) : AnchorSigner {
    .signingAlgorithm(SigningAlgorithmSpec.ECDSA_SHA_256)
```

The private key never enters the workload, which is the property this PR's OpenBao Transit adapter existed to provide.

**"Fail-closed"** — `AuditAnchorService` on `main` takes `openbank.audit.anchor.signing-required` and aborts capture rather than recording a false attestation:

```
if (signingRequired) throw failure
log.error("anchor signing failed; required signer prevents checkpoint capture", failure)
```

This is not merely available, it is **switched on in the deployed manifest** — `openbank-infra/gitops/components/audit/audit-service.yaml` sets `AUDIT_ANCHOR_SIGNER=kms`, `AUDIT_ANCHOR_SIGNING_REQUIRED=true`, `AUDIT_ANCHOR_KMS_KEY_ID=alias/openbank-audit-anchor`.

**Offline verification** — the `AnchorSigner` port on `main` already carries `verify(digest, signature, keyId): Boolean?` and `verificationKeyPem(keyId): String?`, and `AuditResource` serves them. This PR's `/anchors/public-key` is the same capability as `main`'s `/anchors/verification-key`; `main` additionally has `/anchors/verify` and `/anchors`, which this PR does not.

**Signer selection** — `AnchorSignerProducer` on `main` selects hmac vs kms, and its comment records a defect (#5844, `SRCFG00040` on a non-optional `String`) that was found and fixed on that code path after this branch was cut.

### The release-evidence half is superseded too, by a different design

This PR reworked the AI-attribution block of `.github/scripts/build-release-evidence.sh` (#5838) to emit `REPORTED` / `NO_ATTRIBUTION` / `UNAVAILABLE`. `main` has since reworked the same block differently, into an `ai_attribution` object that draws a boundary this PR does not:

```
"build_attribution_note": "authors + Co-Authored-By trailers since previous component tag (ADR-0029 D6)",
"runtime_audit_attribution": {
  "reason": "a release build has no runtime audit-event sample; contributor trailers are not evidence of AI runtime attribution"
```

`.github/scripts/test-release-evidence-attribution.sh` does not exist on `main`.

### Why closing rather than rebasing

Rebuilding this branch on current `main` would not be a merge — it would mean re-litigating two decisions `main` has already made and shipped: OpenBao Transit vs AWS KMS as the anchor-key backend, and reverting `main`'s newer release-evidence design in favour of this one. It also deletes `LocalHmacAnchorSigner`, which `main` still uses as the local-development signer selected by `AnchorSignerProducer`.

Nothing here is lost work that is missing from `main`. If the OpenBao Transit backend is still wanted **as an alternative to KMS** (for example for non-AWS deployments), that is a fresh proposal against the current `AnchorSigner` port — a new adapter plus a `signer: openbao` branch in `AnchorSignerProducer` — and a much smaller change than this branch, since the port, the fail-closed path and the verification endpoints already exist.

The release-evidence coverage idea (#5838) is worth re-reading on its merits against `main`'s current block; the distinction between "we looked and found none" and "we could not look" is a real one and `main`'s version does not obviously make it. Filing that as its own issue against the current script would be the useful next step — this PR cannot carry it without also reverting the rest.

JiRaska:
Closing as **superseded**, with one part worth re-proposing.

#5844 merged on 2026-08-20 and delivered the asymmetric signer this PR proposes. `origin/main`
now carries, in `openbank-audit-service/src/main/resources/application.yaml`:

```
201:      signer: ${AUDIT_ANCHOR_SIGNER:hmac}
202:      signing-required: ${AUDIT_ANCHOR_SIGNING_REQUIRED:false}
203:      kms-key-id: ${AUDIT_ANCHOR_KMS_KEY_ID:}
```

Merging this branch as it stands would **delete those three keys and `LocalHmacAnchorSigner`**,
reverting that work — the branch predates it and a clean merge would not flag the loss.

**The part that is not superseded, and is still a real argument:** this PR's stance that there
should be **no development fallback** — i.e. that `signer` should not default to `hmac` and
`signing-required` should not default to `false`. `main` currently ships permissive defaults, and
the deployment sets `AUDIT_ANCHOR_SIGNER=kms` / `SIGNING_REQUIRED=true` on top. That is a real
posture question rather than a dead one: a config-only mistake in a future environment silently
falls back to HMAC.

Relevant context from work landing today: PR #6206 adds an offline third-party verifier for these
anchors and measured that **1186 of 1212 production anchors are pre-cutover HMAC**, so the
independently-verifiable window starts 2026-08-20. That strengthens the case for tightening the
defaults, and is a much smaller change than this branch.

Suggested re-proposal: a delta against current `main` that changes only the two defaults, with the
migration consequence for existing HMAC anchors stated.

Verified before closing: #5844's merge state and the three keys above read from `origin/main`.

Issue #5869, opened 2026-08-20
ADR-0241 follow-up: MTTR history and weekly RCA automation


JiRaska:
ADR-0241 digest is delivered by #5469, but MTTR P90 (resolved-alert history), weekly resolved-alert RCA log, recurrence guard, and live receipt evidence remain outstanding. Track implementation and operational evidence without claiming current MTTR. Refs ADR-0241.

JiRaska:
## Still true on `origin/main` @ `161febab5` — and the shipped digest says so itself

The one artifact that touches this area is `standing-critical-digest.py`, and it is explicit that it does **not** deliver MTTR:

```bash
$ git show origin/main:.github/scripts/standing-critical-digest.py | grep -n -i 'mttr\|p90'
6:reported as "zero criticals".  The P90 value is the age of currently standing
7:critical alerts (not MTTR, which requires resolved-alert history).
45:    p90 = percentile(ages, 0.90) / 3600
46:    lines = [..., f"Standing-age P90: **{p90:.1f} h** (not MTTR)", ""]
```

Standing-age P90 over *currently open* alerts is a different quantity from MTTR P90 over *resolved* ones, and the script labels it that way in its own output — so nothing anywhere is currently claiming an MTTR figure, which is the one thing this issue asked not to happen. Good state; the deliverable is still unbuilt.

It is also the only file in `.github/` that mentions MTTR at all:

```bash
$ git grep -lni 'mttr' origin/main -- .github
.github/scripts/standing-critical-digest.py
```

So the weekly resolved-alert RCA log and the recurrence guard have no implementation.

**Worth knowing before picking this up:** that same digest workflow has failed on **every scheduled run since it was created**, most recently 2026-09-01 (#6432) — its Alertmanager secrets do not exist. Since resolved-alert history would come from the same Alertmanager, #6432 is effectively a precondition for this issue, not a neighbour of it.

What I did NOT check: whether resolved-alert history is retained anywhere queryable (Prometheus retention here is 12h with no long-term store, which may itself constrain what MTTR P90 can mean).

JiRaska:
## Still unbuilt on `main`, but the work is written: **open PR #5875** carries it, and it answers the retention question my 2026-09-01 comment left open

My last comment ended with *"What I did NOT check: whether resolved-alert history is retained
anywhere queryable."* PR #5875 (`feat/mttr-rca-automation-5869`, open since 2026-08-20, not draft,
[...]

PR #5875, opened 2026-08-20
feat(governance): add critical-alert RCA ledger and recurrence guard

JiRaska:
## Summary

ADR-0241 follow-up to #5469, which delivered the *standing*-alert digest. This adds the **resolved**-alert side: an append-only ledger, the weekly RCA log (ADR-0241 D3), and the rate-of-recurrence guard.

It deliberately **does not publish an MTTR figure**, per the issue's own constraint ("track implementation and operational evidence without claiming current MTTR").

## The retention finding — why MTTR P90 is blocked

ADR-0241 D1 targets critical-alert **P90 <= 4 h**. No source a GitHub Actions workflow can reach dates an alert resolution finely enough to support that number:

| Source | Resolved-alert history? | Usable for a 4 h P90? |
|---|---|---|
| Alertmanager `/api/v2/alerts` (the source #5469 uses) | No — an **active**-alert endpoint, not a history API | No. Resolution is only inferred from absence at the next poll, i.e. +/-24 h at the daily digest cadence |
| Prometheus `ALERTS` | 12 h retention, no long-term store (ADR-0027) | No. Cannot span the 7-day review window at all, and a longer range truncates **silently** rather than erroring |
| Loki | 168 h retention | No. Notification logs, not an authoritative open/close record |
| GoAlert Postgres | **Yes** — durable open/close history, real timestamps | In principle yes, but cluster-internal with no Ingress (ADR-0056/ADR-0088) and no workflow-reachable credential exists |

**This is not fixed by polling more often.** It needs a durable alert open/close feed (a GoAlert history export, or an Alertmanager webhook receiver persisting firing/resolved transitions). That is infra work, out of scope here, and is what remains blocking on #5869.

So `mttr_p90()` **refuses** any P90 whose timestamp granularity is coarser than half the target and returns a machine-readable reason (`granularity-coarser-than-target` / `no-resolved-alert-history`) instead of a number. The refusal is **conditional, not cosmetic** — the same code path emits a real P90 the moment records carry a fine `granularitySeconds`, and the self-test proves both directions.

## What is built

- `.github/scripts/alert-rca-ledger.py` — observe/replay into `docs/runbooks/alert-rca/ledger.jsonl`, the MTTR granularity guard, the recurrence guard, and the weekly RCA renderer.
- `.github/workflows/alert-rca-weekly.yml` — Mondays 10:00 UTC (ADR-0241 D3). Replays the last 7 days of digest observations, renders the RCA log, opens a follow-up issue per recurring alert (deduplicated by title), and publishes the ledger via PR.
- `standing-critical-digest.py` / `.yml` — now also archive an observation envelope `{observedAt, alerts}` alongside the existing markdown, so the ledger folds the **same** observation the digest reports. Mirrors #5469's data source, schedule and artifact mechanism rather than adding a parallel path.

## Falsification

Every guard was written against the case it must **reject** before the case it must accept, then mutation-tested. **12 mutants, 12 killed** — each was confirmed to make the self-test FAIL:

| # | Mutation | Result |
|---|---|---|
| 1 | recurrence threshold `3 -> 2` | killed |
| 2 | recurrence window `14 -> 365` days | killed |
| 3 | drop the window filter entirely | killed |
| 4 | group by `service` instead of `alertname` | killed |
| 5 | granularity guard always **passes** | killed |
| 6 | granularity guard always **refuses** | killed |
| 7 | granularity uses `min` instead of `max` | killed |
| 8 | `observe()` never closes a record | killed |
| 9 | `observe()` admits non-critical alerts | killed |
| 10 | renderer prints a P90 regardless | killed |
| 11 | envelope falls back to "now" for a bare array | killed |
| 12 | replay does not sort by `observedAt` | killed |

Mutant 6 is the one that matters most: it proves the MTTR refusal is contingent on granularity rather than an unconditional "not available" stub.

End-to-end CLI dry run over six synthetic daily observations: `LedgerLag` (3 fires in 14 days) **flagged**, `FxStale` (2 fires) **not flagged** — the negative case holding in a real run, not only a unit fixture. MTTR still refused **with 5 resolved samples in the ledger**, which is the point: the refusal is about timestamp resolution, not about missing data.

Cold start (`t=0`, empty ledger) reports `no-resolved-alert-history` — not a zero and not a sentinel. No gauge or metric is emitted, so there is no boot-time value for an alert to fire on.

## Verification

- `alert-rca-ledger.py --self-test` and `standing-critical-digest.py --self-test` pass, and both are wired as workflow steps (mirroring the `external-feed-watch` / `main-red-watch` pattern).
- `actionlint` clean on both workflows — **and verified non-vacuous** against a deliberately broken control copy, which it correctly rejected.
- `run-gates.py --all`: 156 gates, **the failure set is byte-identical to a clean-`origin/main` control run** of the same gates (5 fail with `PR_DIFF_BASE is empty — refusing to run vacuously`, 1 needs Docker/Loki). `service-runbook-drift` initially failed on my change and is fixed — `docs/runbooks/` is generator-owned and the gate uses `--intent-to-add`, so the new files had to be committed; it passes now.

## What remains outstanding on #5869

1. **MTTR P90** — blocked on a durable alert open/close source, as above. Not buildable honestly today.
2. **Live receipt evidence** — requires the scheduled workflows to actually run against live Alertmanager. Cannot be produced from a PR, which is why this is `Refs`, not `Closes`.

## Note on #5866

Open PR #5866 reconciles ADR-0241's metadata and currently records the weekly RCA and recurrence guard as *planned*. There is **no file overlap** with this PR — I deliberately did not touch the ADR to avoid conflicting with it. Whichever lands second should update the ADR's delivery evidence to cite `alert-rca-weekly.yml`, keeping MTTR itself marked unmeasured.

## Compliance impact

- [x] DORA — evidence of timely handling and review of severe incidents.

## Rollout / Rollback

- Deployment strategy: scheduled GitHub Actions workflow; no cluster or runtime change.
- Rollback plan: revert this commit.

Refs #5869
Refs ADR-0241

### DIFF
--- a/.github/gates/gates.yaml
+++ b/.github/gates/gates.yaml
@@ -94,6 +94,37 @@ gates:
     run: |
       python3 .github/scripts/run-gates.py --self-test
 
+  # Critical-alert RCA ledger guards (ADR-0241 D1/D3, #5869 / #5875). These self-tests
+  # were written and wired ONLY as steps inside the scheduled workflows that use them —
+  # `alert-rca-weekly.yml` (Mondays) and `standing-critical-digest.yml` (daily). Both
+  # workflows fail at their FIRST Alertmanager step because no ALERTMANAGER_* credential
+  # exists (4 of 4 daily runs failed, 2026-08-18..21), so the guards had never executed
+  # anywhere, on any lane, and could not go red on a PR that broke them.
+  #
+  # check-gate-script-registration.py did not catch this and structurally could not: it
+  # scopes itself to `check-*` filenames (these are not), and its own header records that
+  # it deliberately does not check REACHABILITY — a script invoked from a workflow that
+  # never runs still counts as wired. Registering here is what puts them on a PR lane.
+  - id: alert-rca-ledger-guards
+    selftest_exempt: "is-a-test-suite — the run: IS alert-rca-ledger.py --self-test"
+    name: "alert-RCA ledger guards (MTTR granularity, recurrence, observation coverage)"
+    group: lint
+    mode: enforced
+    rationale: "The MTTR granularity refusal and the recurrence guard are the only things stopping ADR-0241 D1/D3 from publishing a number nobody measured, and the coverage outcome is what stops an empty week reporting as a clean one. They were wired only inside a scheduled workflow that has never once succeeded, so they had never run on any lane. Recheck for removal when a durable alert open/close feed exists (#5869) and the MTTR half is no longer a refusal."
+    review_after: "2027-02-21"
+    run: |
+      python3 .github/scripts/alert-rca-ledger.py --self-test
+
+  - id: standing-critical-digest-guards
+    selftest_exempt: "is-a-test-suite — the run: IS standing-critical-digest.py --self-test"
+    name: "standing-critical digest guards (severity filter, observation envelope)"
+    group: lint
+    mode: enforced
+    rationale: "The digest's severity filter and observation envelope are the producer half of the same control; if the envelope shape drifts, the weekly review silently folds nothing and reports a clean week. Same lane gap as its consumer above. Recheck when the digest stops being the ledger's only observation source."
+    review_after: "2027-02-21"
+    run: |
+      python3 .github/scripts/standing-critical-digest.py --self-test
+
   - id: yamllint
     selftest_exempt: "third-party — yamllint's own test suite is not ours to run"
     name: "yamllint"
--- a/.github/scripts/alert-rca-ledger.py
+++ b/.github/scripts/alert-rca-ledger.py
@@ -1,0 +1,624 @@
+#!/usr/bin/env python3
+"""Resolved-alert ledger, MTTR guard and recurrence guard for ADR-0241.
+
+Companion to `standing-critical-digest.py` (#5469), which reports the age of
+alerts that are *currently standing*.  This script keeps the complementary
+record: when a critical alert stopped being observed, so that recurrence can be
+counted and — once a source with adequate timestamp resolution exists — MTTR can
+be computed.
+
+WHY THIS DOES NOT PUBLISH AN MTTR TODAY
+---------------------------------------
+ADR-0241 D1 sets the target at **P90 <= 4 h**.  A P90 is only meaningful if the
+resolution timestamps are finer-grained than the target.  The sources available
+to a GitHub Actions workflow today are not:
+
+  * Alertmanager `/api/v2/alerts` — the source #5469 already uses — returns
+    ACTIVE alerts.  It is not a history API; a resolved alert is simply absent,
+    so resolution time is only ever known to the polling interval.
+  * Prometheus — 12 h retention, no long-term store (ADR-0027).  The `ALERTS`
+    series cannot cover the 7-day review window of ADR-0241 D3 at all, and a
+    range query silently truncates rather than erroring.
+  * Loki — 168 h retention, and carries notification logs, not an authoritative
+    open/close record.
+  * GoAlert Postgres — DOES hold durable open/close history with real
+    timestamps, and is the intended source.  It is cluster-internal with no
+    Ingress (ADR-0056/ADR-0088) and no workflow-reachable credential exists.
+
+So the daily poll yields `granularity_seconds = 86400` against a 14400 s target.
+`mttr_p90` therefore REFUSES to emit a number and returns a reason instead.  The
+refusal is a rule, not a placeholder: it passes the moment records arrive from a
+fine-grained source (see `self_test`, which proves both directions).
+"""
+from __future__ import annotations
+
+import argparse
+import datetime as dt
+import json
+import os
+import sys
+from collections import defaultdict
+
+# ADR-0241 D1.
+MTTR_TARGET_SECONDS = 4 * 3600
+# A P90 is not credible unless resolution timestamps are at least twice as fine
+# as the target; otherwise the bucket width dominates the statistic.
+MAX_CREDIBLE_GRANULARITY_SECONDS = MTTR_TARGET_SECONDS // 2
+# ADR-0241 D3 rate-of-recurrence guard.
+RECURRENCE_THRESHOLD = 3
+RECURRENCE_WINDOW_DAYS = 14
+
+DEFAULT_LEDGER = "docs/runbooks/alert-rca/ledger.jsonl"
+
+# --------------------------------------------------------------------------- #
+# COVERAGE — "no alerts" and "no observations" are DIFFERENT outcomes.
+#
+# The daily standing-critical digest archives an envelope every day it succeeds,
+# INCLUDING on a day with zero critical alerts (`{"observedAt": ..., "alerts": []}`).
+# So a genuinely quiet week still yields ~7 envelopes. Zero envelopes therefore
+# does not mean "nothing fired" — it means the producer never ran, and the weekly
+# review has measured nothing at all.
+#
+# Those two states must never share an outcome or an exit code. This repo already
+# paid for that lesson once: `PushResult.skipped()` returned `success = True` when
+# APNs was disabled, the fan-out counted it with `count { success }`, and every
+# push in a credential-less environment was recorded as delivered. A no-op that
+# reports success is indistinguishable from a working control, and nothing
+# downstream can disagree with it.
+EXIT_OK = 0
+EXIT_USAGE = 1
+EXIT_NO_OBSERVATIONS = 2
+
+COVERAGE_OBSERVED = "OBSERVED"
+COVERAGE_NO_OBSERVATIONS = "NO_OBSERVATIONS"
+
+
+def coverage(observation_count: int) -> dict:
+    """Classify what this run actually measured.
+
+    Deliberately NOT a boolean: a bool invites `if ok:` at the call site, which is
+    exactly how a no-op gets folded back into the success path.
+    """
+    if observation_count > 0:
+        return {
+            "outcome": COVERAGE_OBSERVED,
+            "observationCount": observation_count,
+            "detail": (
+                f"{observation_count} daily observation envelope(s) folded; the "
+                "findings below describe the window that was actually measured."
+            ),
+        }
+    return {
+        "outcome": COVERAGE_NO_OBSERVATIONS,
+        "observationCount": 0,
+        "detail": (
+            "No observation envelope covered this window. The daily "
+            "standing-critical digest archives an envelope even on a day with zero "
+            "critical alerts, so this is NOT a quiet week — it means the producer "
+            "did not run (or ran and failed). Nothing below has been measured."
+        ),
+    }
+
+
+def parse_time(value: str) -> dt.datetime:
+    return dt.datetime.fromisoformat(value.replace("Z", "+00:00"))
+
+
+def fmt_time(value: dt.datetime) -> str:
+    return value.astimezone(dt.timezone.utc).strftime("%Y-%m-%dT%H:%M:%SZ")
+
+
+def percentile(values: list[float], p: float) -> float:
+    ordered = sorted(values)
+    rank = (len(ordered) - 1) * p
+    low, high = int(rank), min(int(rank) + 1, len(ordered) - 1)
+    return ordered[low] + (ordered[high] - ordered[low]) * (rank - low)
+
+
+def load_ledger(path: str) -> list[dict]:
+    if not os.path.exists(path):
+        return []
+    records = []
+    with open(path, encoding="utf-8") as stream:
+        for line in stream:
+            line = line.strip()
+            if line:
+                records.append(json.loads(line))
+    return records
+
+
+def write_ledger(path: str, records: list[dict]) -> None:
+    os.makedirs(os.path.dirname(path) or ".", exist_ok=True)
+    records = sorted(records, key=lambda r: (r["startsAt"], r["key"]))
+    with open(path, "w", encoding="utf-8") as stream:
+        for record in records:
+            stream.write(json.dumps(record, sort_keys=True) + "\n")
+
+
+def observe(
+    records: list[dict],
+    active: list[dict],
+    observed_at: dt.datetime,
+    granularity_seconds: int,
+    source: str,
+) -> list[dict]:
+    """Fold one observation of currently-active criticals into the ledger.
+
+    An alert present in `active` is opened (or refreshed).  An OPEN ledger
+    record whose key is absent from this observation is closed at `observed_at`.
+    Closing is only correct for a source that enumerates every active alert, so
+    an observation is authoritative for the keys it does not contain too.
+    """
+    by_key = {r["key"]: dict(r) for r in records}
+    seen = set()
+    for alert in active:
+        labels = alert.get("labels") or {}
+        if labels.get("severity") != "critical":
+            continue
+        starts_at = alert.get("startsAt")
+        if not starts_at:
+            continue
+        alertname = labels.get("alertname", "<unnamed>")
+        service = labels.get("service", labels.get("job", "-"))
+        key = f"{alertname}|{service}|{fmt_time(parse_time(starts_at))}"
+        seen.add(key)
+        existing = by_key.get(key)
+        if existing is None:
+            by_key[key] = {
+                "key": key,
+                "alertname": alertname,
+                "service": service,
+                "severity": "critical",
+                "startsAt": fmt_time(parse_time(starts_at)),
+                "lastSeenAt": fmt_time(observed_at),
+                "resolvedAt": None,
+                "source": source,
+                "granularitySeconds": granularity_seconds,
+            }
+        elif existing.get("resolvedAt") is None:
+            existing["lastSeenAt"] = fmt_time(observed_at)
+    for key, record in by_key.items():
+        if key in seen or record.get("resolvedAt") is not None:
+            continue
+        # Never close a record with an observation that predates its last sighting.
+        if parse_time(record["lastSeenAt"]) <= observed_at:
+            record["resolvedAt"] = fmt_time(observed_at)
+    return list(by_key.values())
+
+
+def mttr_p90(records: list[dict]) -> dict:
+    """Return the P90 time-to-resolve, or a refusal with a machine-readable reason.
+
+    Never returns a number the underlying timestamps cannot support.
+    """
+    resolved = [r for r in records if r.get("resolvedAt")]
+    if not resolved:
+        return {
+            "p90Seconds": None,
+            "sampleCount": 0,
+            "reason": "no-resolved-alert-history",
+            "detail": "The ledger holds no resolved critical alert yet.",
+        }
+    worst = max(int(r.get("granularitySeconds", MTTR_TARGET_SECONDS)) for r in resolved)
+    if worst > MAX_CREDIBLE_GRANULARITY_SECONDS:
+        return {
+            "p90Seconds": None,
+            "sampleCount": len(resolved),
+            "reason": "granularity-coarser-than-target",
+            "detail": (
+                f"Resolution timestamps are known only to {worst}s; a "
+                f"{MTTR_TARGET_SECONDS}s P90 target needs "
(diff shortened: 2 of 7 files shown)


JiRaska:
Review (read-only; no approval implied). The scripts were executed and mutation-tested, not just read.

## The script is good. The pipeline it feeds is inert.

**Credit first, because it is earned.** The ledger is genuinely **derived** — `alert-rca-ledger.py:96-146` writes it from observation envelopes the daily digest archives, so the subject set comes from Alertmanager, not from a human. And the self-test is a real oracle, which is rare. I mutation-tested the four load-bearing behaviours; every one reddens:

```
RECURRENCE_THRESHOLD 3 -> 999          exit=1
MTTR refusal `if worst > MAX` -> `if False:`  exit=1
close-on-absence removed                exit=1
replay sort removed                     exit=1
```

The cases at `:296-323` falsify by construction: 2-in-14 must not flag, 3-spread-past-14 must not flag, distinct alertnames must not aggregate, both window edges, MTTR must refuse coarse and pass fine, a bare-array envelope must be refused rather than defaulted.

That makes the four findings below fixable rather than fundamental.

## Finding 1 (blocking, in my view): the producer has never once succeeded, and the consumer reads that as a clean week

```
$ gh run list --workflow standing-critical-digest.yml --limit 6
2026-08-20T08:46:29Z failure
2026-08-19T08:46:10Z failure
2026-08-18T08:46:03Z failure
ERROR: ALERTMANAGER_URL and ALERTMANAGER_DIGEST_TOKEN are required
$ gh secret list | grep -ci alertmanager
0
```

Those secrets do not exist. Every daily run since #5469 exits 2 at `standing-critical-digest.py:92`. Chained through the new workflow: `alert-rca-weekly.yml:70-72` filters `--status success` → **zero** runs → `:96-99` no envelopes → `:141` `git diff --quiet` → *"no ledger or RCA change to publish"*, **exit 0**.

Measured directly: `exit=0`, `"reason": "no-resolved-alert-history"`, `"recurrences": []`, rendering *"No critical alerts recorded in the window."*

So a critical alert can fire, stand for a week, and the weekly review runs green, publishes nothing, opens no issue, and prints two sentences that read as "clean" and mean **"unmeasured"**. `OBSERVATION_COUNT` is computed at `:80` and used only in a commit message (`:151`) — there is no `--require-observations`, no floor, no failure when the expected ~7 daily envelopes are 0.

Ask what an **absent** subject prints: it prints what a healthy week prints. One-line fix — fail (or at minimum `::warning`) when `OBSERVATION_COUNT` is below the number of scheduled digest runs in the window.

## Finding 2: the daily sampling cadence structurally cannot see the alerts the recurrence guard exists to catch

The digest polls once daily (`standing-critical-digest.yml:5`, `cron: 30 8 * * *`) and Alertmanager `/api/v2/alerts` returns only **active** alerts — the script's own docstring (`:17-19`) says so. Therefore **any critical alert that fires and resolves inside 24 h is never recorded at all.**

That set includes every alert that *meets* the ADR-0241 D1 4 h target, and it includes flapping alerts — which is precisely what a "3 fires in 14 days" guard is for. A short-lived flapper firing 10 times in 14 days appears **zero** times unless it happens to be up at 08:30. The docstring is candid about the MTTR consequence and silent about this one; `docs/runbooks/alert-rca/README.md:16` frames daily granularity purely as an MTTR-precision problem when it is also a sampling problem that voids the guard. Please state in the README that recurrence counts are a **floor, not a count**.

Coverage, for scale: 26 distinct `- alert:` blocks carry `severity: critical` across `openbank-infra`. The ledger covers 0, and nothing cross-checks ledger alertnames against those 26 — so a critical rule that never fired and one the pipeline is blind to are indistinguishable.

## Finding 3: not registered in `gates.yaml`, so the good self-test never runs on a PR

`grep -i rca .github/gates/gates.yaml` → no output. `alert-rca-weekly.yml:16-20` is `schedule` + `workflow_dispatch` only. So the oracle I just proved survives mutation testing **never runs on a PR**; the script can be broken by any future edit and the first anyone learns is a Monday scheduled run that currently does nothing anyway. The self-test is fast, hermetic and needs no network — this is a free gate left on the table. `critical-alert-egress` (`gates.yaml:2604-2611`) is the same domain with a `selftest:` line and is the pattern to copy.

## Finding 4: two data-loss bugs in the published artifact

- **The weekly regeneration wipes the hand-written RCA column.** `README:5` says the RCA column "is filled in by hand at the Monday 10:00 UTC review" and `README:8` says regenerate-never-hand-edit "except for the RCA column". Both cannot be true: `render_rca` writes the whole file in `"w"` mode with `_(fill in)_` in every RCA cell (`alert-rca-ledger.py:236`). Every Monday deletes the previous Monday's analysis — which is the entire human deliverable of ADR-0241 D3.
- **The monthly file loses earlier weeks of the same month.** Output path is `${month}.md` (`alert-rca-weekly.yml:88,103`) but the content is a trailing **7-day** window (`:210`). Demonstrated: week ending 08-05 renders rows for 08-03 and 08-18; week ending 08-20 renders only 08-18 — the 08-03 row is gone from `2026-08.md`. Three of four Mondays a month silently delete their predecessor's rows. Either name the file by week or render the whole month.

## Smaller items

- **`ledger.jsonl` is not append-only despite the claim.** `README:2` and the workflow header both say "append-only"; `write_ledger` (`:88-94`) opens `"w"` and rewrites the whole file from in-memory state every run. Word it honestly, or add a check that no existing line's content changed.
- **False comment at `alert-rca-weekly.yml:91-92`**: it describes `find -print0 | xargs -0` "so a path never word-splits", but the code below is `mapfile -t < <(find ... | sort)` — a different mechanism that *does* break on newlines in paths.
- **`gh run download` failure is swallowed** (`:76`, `2>/dev/null` + `if`): an expired artifact or a 500 is indistinguishable from a run with no observation, and both quietly reduce `count`. Combined with Finding 1, every partial-download path also reads as "quiet week".
- **`--limit 30` on a 7-day window of daily runs** (`:71`) is a hidden cap that silently truncates if the digest ever goes hourly.
- `standing-critical-digest.yml:51` changes the issue title from `Weekly critical-alert RCA review (<run_id>)` to `Standing critical alerts <run_id>`. Both embed `run_id`, so neither ever deduplicates — every standing-critical day opens a new issue. Pre-existing, but this PR touched the line.

## Release scope: correct

Diff is `.github/**` and `docs/**` only; no module directory, no `version.txt`. So `feat(governance)` is a releasing type on a non-releasing path and cuts nothing — correct and intentional. ADR-0241 is left at `delivery-status: planned`, which is right since D3 is not actually delivered; no over-claim there.

## Bottom line

Careful, well-reasoned work whose executable core is better than most gates in this repo. But as it stands the weekly review is green about nothing: the producer has failed 3/3 for want of secrets that do not exist, the consumer treats zero observations as a clean week, the self-test is unreachable from any PR lane, and the sampling cadence cannot see the flapping alerts the guard targets. Minimum to make it honest: fail when `OBSERVATION_COUNT` is below the expected digest-run count; register the self-test in `gates.yaml`; fix the `${month}.md` overwrite and the RCA-column wipe; and say in the README that recurrence counts are a floor.

JiRaska:
## Independent re-verification of the review's four findings

All four confirmed; one is worse than reported, and one that the review flagged as unresolvable has now been resolved. Commands and outputs below. In-repo fixes are stacked on this branch as #6192.

### 1. Producer workflow failing — CONFIRMED, and worse than 3/3

```
$ gh api ".../workflows/standing-critical-digest.yml/runs?per_page=50" \
    --jq '.total_count, ([.workflow_runs[].conclusion]|group_by(.)|map({(.[0]):length})|add)'
4
{"failure":4}
```

**4 of 4 — every run the workflow has ever had.** Not 3 of 3; a fourth ran this morning and failed the same way. There has never been a success, so no observation envelope has ever been produced.

Decisive lines from the most recent run (32464729398):

```
env:
  ALERTMANAGER_URL:
  ALERTMANAGER_DIGEST_TOKEN:
ERROR: ALERTMANAGER_URL and ALERTMANAGER_DIGEST_TOKEN are required
##[error]Process completed with exit code 2.
```

Both variables interpolate to empty. The script fails closed, which is correct behaviour — the defect is downstream of it.

### 2. Secrets missing — CONFIRMED at every scope, including the one the review could not rule out

```
$ gh secret list
CLAUDE_CODE_OAUTH_TOKEN, GITOPS_APP_ID, GITOPS_APP_PRIVATE_KEY, GOVERNANCE_GH_PAT,
INFRACOST_API_KEY, OPENBANK_APP_RO_TOKEN, PACT_BROKER_PASSWORD, RULESET_AUDIT_TOKEN,
SCORECARD_TOKEN
```

Neither `ALERTMANAGER_URL` nor `ALERTMANAGER_DIGEST_TOKEN` is present.

```
$ gh api repos/.../environments/platform-apply/secrets    -> {"total_count":0,"secrets":[]}
$ gh api repos/.../environments/substrate-apply/secrets   -> {"total_count":0,"secrets":[]}
$ gh secret list --org JiRaska
failed to get secrets: HTTP 404: Not Found (https://api.github.com/orgs/JiRaska/actions/secrets)
```

The review explicitly could not rule out org or environment scope. **Both are now ruled out**, and org scope is not merely empty but structurally impossible — the owner is a *user account*, not an organization, so no org secret store exists. `gh variable list` holds only the two Pact entries. The workflow also references neither an `environment:` nor a variable fallback, so repo scope is the only place these could come from.

### 3. Gate absent from `gates.yaml` — CONFIRMED

```
$ git show origin/feat/mttr-rca-automation-5869:.github/gates/gates.yaml > g.yaml
$ grep -c '^\s*- id:' g.yaml
156
$ grep -n 'rca' g.yaml   ; echo $?
1                # no match
$ grep -n 'digest' g.yaml ; echo $?
1                # no match
```

156 registered gates, zero mentioning either script. `--self-test` for both `alert-rca-ledger.py` and `standing-critical-digest.py` exists only as a step inside the scheduled workflows — and those die at their first Alertmanager step, so **neither self-test has ever executed anywhere, on any lane**. Twelve mutants were killed locally by a harness that CI has never run and that could not go red on a PR breaking it.

Worth noting for the follow-up: `check-gate-script-registration.py` exists precisely to catch "written and never run", and could not catch this for two independent reasons — it scopes itself to `check-*` filenames, and its own header states it deliberately does not check *reachability*: "a script invoked from a workflow whose `if:` is always false, or from a job excluded on the PR lane, passes this gate." A meta-gate scoped by naming convention is blind to anything unconventionally named.

### 4. Empty week reads as a clean week — CONFIRMED empirically

Not merely by inspection. Running the consumer exactly as the weekly workflow would, with no observations available:

```
$ python3 alert-rca-ledger.py --ledger empty.jsonl --rca-output out.md ; echo $?
0
```

and `out.md` contains:

```
## Alerts in the last 7 days
No critical alerts recorded in the window.

## Rate-of-recurrence guard (ADR-0241 D3)
No alert reached the recurrence threshold in the window.
```

Byte-identical, in prose and in exit code, to a week where the digest ran daily and genuinely saw nothing. Given finding 1, **this empty case is the only output this control can currently produce** — it would have reported a clean week, every Monday, indefinitely, and nothing anywhere would have disagreed.

The workflow does compute `OBSERVATION_COUNT`, and handles the empty case carefully in one respect — it correctly declines to fold an empty observation, which would otherwise close every open record. But the count is used only in a commit message. It is never an outcome.

## Credit where due

The design is careful in the places that usually go wrong, and I want to be clear that finding 4 is a gap in an otherwise well-built control, not sloppiness. The MTTR refusal is genuinely conditional (mutant 6 in your table proves it), `read_observation` refuses a bare array rather than defaulting to "now", `replay` sorts by `observedAt`, and the download step filters on `--status success` with a comment explaining exactly why absence must not be read as "no alerts". That last comment states the correct principle one level down from where the bug is: absence of a *run* must not read as absence of *alerts* — which is precisely what the consumer then did with absence of *observations*.

## What #6192 fixes

`coverage()` classifies every run as `OBSERVED` or `NO_OBSERVATIONS`. They never share an outcome or an exit code: `NO_OBSERVATIONS` exits **2**, renders a `CAUTION` banner, and reports every finding as **Unknown** rather than "none". A dict, not a bool — a bool invites `if ok:`, which is how a no-op gets folded back into the success path. `render_rca()` now requires coverage rather than defaulting it. The weekly workflow still publishes the document recording the gap, then fails the job.

Both self-tests are registered in `gates.yaml` (shard `lint`), so they now run on every PR.

**8 mutants, 8 killed — but mutant 5 initially survived**, and it is the interesting one. Flipping `--allow-no-observations` from opt-in to default is the entire polarity of the guard, and it changes no output string, so every in-process prose assertion stayed green. The self-test now drives the real CLI in a subprocess and asserts on the **exit code**. Offered as a note on your own table: prose assertions cannot see an exit code, and the exit code is what the workflow reads.

Negative control on the registered gate, read from `$?`:

```
# defect reintroduced
$ run-gates.py --only alert-rca-ledger-guards ; echo $?
1
AssertionError: an uncovered window must exit 2, got 0 — the empty case has fallen back into the success path

# restored
$ run-gates.py --only alert-rca-ledger-guards ; echo $?
0
```

`lint` 22/22 and `registry` 26/26, both exit 0, run per shard in the foreground.

## What #6192 does NOT fix — and needs the repo owner

The producer still cannot run. Two repository secrets must be created, and only the repo owner can do that:

| Secret | Scope | Needs to be |
|---|---|---|
| `ALERTMANAGER_URL` | repository | base URL of the Alertmanager instance the digest should poll (the workflow calls `/api/v2/alerts` beneath it) |
| `ALERTMANAGER_DIGEST_TOKEN` | repository | a **read-only** bearer token for that endpoint |

Repository scope specifically — the workflow declares no `environment:`, so environment-scoped secrets would not resolve.

I have deliberately not created these, and no value for either appears anywhere in this comment, the stacked PR, or its branch.

Until they exist, the honest outcome is the one #6192 produces: the weekly review fails loudly every Monday instead of publishing a clean-looking report about nothing. If the endpoint turns out not to be reachable from a GitHub-hosted runner at all, that is a *different* and larger finding — the control would need a self-hosted lane or a push-based feed — and is worth folding into #5869 rather than discovering next Monday.

## Note on process

I did not push to this branch. It has no worktree (`git worktree list`) and its head has been idle 24.6 h (`0b288e466`, 2026-08-20T12:19:39+02:00), so it is not demonstrably live — but it is your PR, and stacking keeps the fix reviewable on its own and leaves your falsification record intact. #6192 is based on `feat/mttr-rca-automation-5869` and will collapse into this PR's diff when it lands.

JiRaska:
## Review — half of it is now live-and-good (thanks to the stacked #6192); the other half cannot function in this repo

**Staleness:** 31 commits behind `main`, `CONFLICTING`. Head is `33afc76d5`, which is **#6192 already merged into this branch** — so the branch is further along than the PR description suggests, and one earlier objection against it is now obsolete.

**Claim I can no longer make: "the guards are registered on no PR lane."** That was true of the original branch and is **fixed on the current head**. `.github/gates/gates.yaml:108` and the entry below it register `alert-rca-ledger-guards` and `standing-critical-digest-guards`, both `mode: enforced`, both `selftest_exempt: "is-a-test-suite"`. The comment block above them is also the right diagnosis of why nothing caught this: `check-gate-script-registration.py` scopes itself to `check-*` filenames (these are not) and by its own header does not check reachability, so a script invoked from a workflow that never runs still counted as wired. That is a real gap closed, and it is the part of this PR I would keep.

**Claim that still stands, and it is the load-bearing one: the producer has never once succeeded, and cannot.**

- `standing-critical-digest.yml` is already on `main`, and its last four scheduled runs — 2026-08-18, -19, -20, -21 — are all `failure`, failing in the first step, `Query Alertmanager and publish digest`.
- The workflow requires `secrets.ALERTMANAGER_URL`, `secrets.ALERTMANAGER_DIGEST_TOKEN` (`standing-critical-digest.yml:21-22`) and `secrets.ALERTMANAGER_DIGEST_SLACK_WEBHOOK` (line 33, with an explicit `test -n "$WEBHOOK" || exit 1`).
- `gh secret list` on this repository returns nine secrets and **none of them is an `ALERTMANAGER_*`**. The repository owner is a **user account**, so there is no organisation secret store for them to be inherited from either. There is no scope at which these values exist.

So the RCA ledger's only observation source produces nothing, has produced nothing since it was scheduled, and will keep producing nothing after this merges. The weekly `alert-rca-weekly.yml` is not even on the default branch yet (`gh run list --workflow=alert-rca-weekly.yml` → 404), so it has zero runs by construction.

**Why that matters more than "a scheduled job is red".** This PR's own `standing-critical-digest-guards` rationale says the envelope check is what stops "an empty week silently folding nothing and reporting a clean week". #6192 gave the empty week its own outcome, which is the right fix — but it makes the ledger *correctly* report "no observations", forever, on a control whose entire purpose is to make a dead alerting path visible. That is the failure mode the repo has hit before: noise or silence on the control that exists to detect silence.

**What I would ask for before merge (author decision, not a code fix):**
1. State plainly in the PR body and in `docs/runbooks/alert-rca/README.md` that the ledger is **inert until an Alertmanager credential exists**, and say who is expected to provision it. A runbook that describes a weekly review nobody can perform is the same class of artifact as the runbook-citation #5889 is currently cleaning up.
2. Either provision the three secrets, or make the scheduled workflow **skip cleanly** when they are absent instead of failing daily — a permanently-red scheduled job is where a genuinely new failure goes to be ignored.
3. Resolve the conflict against `main` and re-check the `gates.yaml` merge by **counting entries before and after**, not by reading the diff. #5967 in this same batch also appends to the `gates:` list; two branches adding neighbouring list entries is precisely the merge git calls clean while keeping only one.

**Not verified:** I did not run the two `--self-test` suites, and I did not review `check-gate-subject-floor.py` in depth.

Verdict: **needs author decision.** The gate registration deserves to land; the RCA ledger should not land describing a process this repository has no credential to run. Not approving — review comment only.

JiRaska:
## Rebuilt on current `main` — conflict was a set-union, and the checks now actually run

Was CONFLICTING on `.github/scripts/check-gate-subject-floor.py`, with required checks entirely **absent** (never ran) rather than failing. Rebuilt on `origin/main` (`8b1868746`); new head `bd856a8b1`. Checks are now queued and reporting.

### The conflict

Both sides added an entry to the `NO_CORPUS` set: `main` added `supersede-deploy-prs-ancestry`, this branch added `standing-critical-digest-guards`. A pure set-union — kept both, in sorted order, alongside this branch's other entry `alert-rca-ledger-guards` which applied cleanly:

```
57:    "alert-rca-ledger-guards",
72:    "standing-critical-digest-guards",
73:    "supersede-deploy-prs-ancestry",
```

`main`'s entry is preserved; nothing was reverted.

### Verification

The gate itself passes, and the count moved from 172 to 173 gates:

```
$ python3 .github/scripts/check-gate-subject-floor.py
SUBJECTS=173  # gates in the manifest
gate-subject-floor: 55/173 gates declare a subject floor; 19 have no corpus to count, 99 baselined as debt.
exit=0
```

A passing gate is not evidence on its own, so I ran the negative control — deleting one of the new entries and confirming the gate can detect its absence:

```
$ sed -i '/"alert-rca-ledger-guards",/d' .github/scripts/check-gate-subject-floor.py
$ python3 .github/scripts/check-gate-subject-floor.py
::error::alert-rca-ledger-guards: no `min_subjects:` and no entry in check-gate-subject-floor.py.
  A gate that examines nothing passes everything [...]
gate-subject-floor: 55/173 gates [...] 18 have no corpus to count, 99 baselined as debt.
```

The entry is load-bearing: without it the new gate is reported as unchecked. Note the run still exits 0 because this gate is currently in advisory mode — the `::error::` annotation is the signal, not the exit code.

Also validated the rest of the payload:

```
py syntax OK   (alert-rca-ledger.py, standing-critical-digest.py)
yaml OK        (alert-rca-weekly.yml, standing-critical-digest.yml, gates.yaml)
actionlint     alert-rca-weekly.yml           exit=0
actionlint     standing-critical-digest.yml   exit=0
```

Scope is exactly the intended 7 files. Not merged — left for review.

JiRaska:

|---|---|
| #5890 | 3934 |
| #5849 | 3853 |
| #5875 | 3735 |
| #5967 | 3417 |

PR #5877, opened 2026-08-20
feat(libs): label LLM call metrics with the egress provider

JiRaska:
## What

Closes the small label gap #5736 names, and reports what the repo can and cannot decide about the outage itself.

`openbank_llm_requests_total` / `_tokens_total` / `_call_duration_seconds` now carry a `provider` tag alongside `model` — the egress backend the attempt was **addressed to**, from a closed vocabulary (`litellm`, `groq`, `deepinfra`, `openai`, `ollama`, `other`, `unknown`), derived from the configured endpoint by `LlmCallMetricsPort.providerOf`.

Two deliberate choices:

- **Derived from the endpoint, not a second config knob.** It then follows a repoint (direct vendor -> LiteLLM gateway) with nothing to keep in step. It is honestly the backend *addressed*, not the upstream vendor LiteLLM routes on to — no caller can observe that mapping, and a label claiming knowledge the emitter does not have is worse than a coarse one.
- **`outcome` is untouched.** Same vocabulary, independent tag. `AiCallErrorRateHigh`, `AiCallsNotConfigured` and the dashboard's `outcome="success"` filter keep their exact meaning; the cost rules join `on (model, kind)`, so the extra label rides through rather than breaking the match. A test asserts both halves.

## What the repo rules out about the outage

`http_error` is measured, not inferred from the name: `OpenAiCompatibleLlmGatewayClient` and the two per-service `OpenAiCompatibleModelProvider`s all record it on exactly `resp.statusCode() !in 200..299`. So the gateway answered with a non-2xx.

Checked against `gitops/components/ai-platform/litellm-config.yaml`, and each is **not** the cause:

- **Model ids all match.** The gateway serves `deepseek-ai/DeepSeek-V3.2` and `llama-3.3-70b-versatile`. Every key-carrying caller sends one of exactly those (copilot, devops-agent, control-liveness-sentinel, case-coordinator -> DeepSeek; agent-service -> llama). No caller sends a model the gateway does not serve.
- **Endpoints are right.** All five set `http://litellm.ai-platform.svc:4000/v1`, and the client appends `/chat/completions`.
- **Auth shape is right.** `Authorization: Bearer <key>`, which is what LiteLLM expects for a virtual key.
- **NetworkPolicy is not it.** All three `platform`-namespace callers are in the litellm ingress allowlist — and a NetPol drop would surface as `exception`/timeout, not `http_error`, anyway.
- **Request shape is standard OpenAI** (`model`, `messages`, `temperature`, `max_tokens`, `stream:false`), and the gateway runs `drop_params: true`.

Consistent side-observation: the six ops agents (governance-auditor, finops, docs-truth, flaky-test-hunter, release-steward, authz-policy-auditor) have **no** `*_MODEL_API_KEY` env at all, so their path would emit `not_configured` — and there is no `not_configured` series either, i.e. those agents' LLM path has never run.

## What the repo cannot decide

The remaining candidates are runtime-only:

1. the LiteLLM **virtual key** each caller presents is not valid — `litellm-config.yaml` states in its own comments that keys exist "only in the database because upstream creates keys through `/key/generate`", a manual runtime action, while the vault properties (`KEY_AGENT_SERVICE`, `KEY_COPILOT_SERVICE`, `KEY_DEVOPS_AGENT`, `KEY_CONTROL_LIVENESS_SENTINEL`) are hand-seeded; a stale or never-generated key gives 401 to **every** caller, which matches 41/41 exactly;
2. the **upstream** provider key (`DEEPINFRA_API_KEY` / `GROQ_API_KEY`, both `optional: true`) is unseeded or rejected, giving a 401/500 back through the proxy.

No speculative change is made for either. A wrong guess on a broken-but-instrumented path makes it fail differently and harder to diagnose.

**The single live check that settles it** — one call, no mutation, from inside the cluster:

```
kubectl -n platform exec deploy/copilot-service -- \
  curl -s -o /dev/stderr -w '%{http_code}\n' \
  -H "Authorization: Bearer $COPILOT_MODEL_API_KEY" \
  -H 'Content-Type: application/json' \
  -d '{"model":"deepseek-ai/DeepSeek-V3.2","messages":[{"role":"user","content":"ping"}],"max_tokens":1}' \
  http://litellm.ai-platform.svc:4000/v1/chat/completions
```

`401` + `"Invalid proxy server token"` is candidate 1 (fix: regenerate virtual keys via `/key/generate` with the master key, re-seed the vault properties). `401`/`500` naming the upstream provider is candidate 2 (fix: seed `openbank/litellm`). Anything else re-opens the diagnosis.

## On the absence-alert trap

No "no successful calls" alert is added, and `prometheus-rules-ai.yaml` now says why. Unguarded, `sum(rate(...{outcome="success"}[15m])) == 0` returns an empty vector when no success series exists — identical output to a healthy fleet, blind exactly when the path breaks completely (#5733 / the #5850 dead-man's-switch shape). And even `absent_over_time`-guarded it cannot go in yet: LLM calls here come from cron-scheduled agents and an interactive assistant, so **silence is the resting state** and the alert would fire continuously while nothing is wrong. It needs a synthetic call on a known interval first, so that absence means something.

## Follow-up noticed, not changed

`copilot-service`'s `completeStream` (SSE path) records **no** metric on any outcome — success, http_error or exception. So the 41 is an undercount of unknown size, and a streaming-only failure is invisible. Left alone here to keep this diff to the label; worth its own issue.

## Verification

Serialized Gradle, tasks read from the output rather than assumed:

- `ktlintCheck` + `detekt` on all four touched modules — all 8 tasks present, BUILD SUCCESSFUL.
- `:openbank-libs-domain:test`, `:openbank-libs-runtime:test`, `:openbank-agent-service:test`, `:openbank-copilot-service:test` — green, **0 skipped** across 57 result files.
- The new tests were falsified: reverting `providerOf` to host-equality made `both spellings of the in-cluster gateway classify as one provider` and `the recorded provider comes from the configured endpoint` go red, then green again on restore.

Refs #5736


JiRaska:
Closing as superseded: diff against origin/main is empty (`git diff origin/main...origin/fix/llm-provider-label` shows zero changes). Everything this branch added has already landed on main.

Issue #4791, opened 2026-08-14
merged-past-red-check-watch has never answered on push or schedule: 87 of 87 real runs fail on RULESET_AUDIT_TOKEN lacking Administration:read


JiRaska:
## The bypass detector has never once answered on the lanes that matter

`merged-past-red-check-watch.yml` exists to catch a PR merged past a red required check — the
failure #4240 was filed for. It runs on three triggers: `schedule` (twice daily), `push` to `main`,
and `pull_request` limited to its own two files.

**Every `push` and `schedule` run has failed. 87 of 87, zero successes**, across the full history the
API returns (back to 2026-08-13T18:55Z):

```
gh api .../workflows/merged-past-red-check-watch.yml/runs
  -> {"total": 87, "successes": 0, "oldest": "2026-08-13T18:55:37Z"}   # push + schedule only
```

The cause is stated by the run itself:

```
could not read rule suites: gh api --paginate repos/.../rulesets/rule-suites?time_period=month
  failed (1): gh: Resource not accessible by personal access token (HTTP 403)
::error::secrets.RULESET_AUDIT_TOKEN is unset or lacks Administration:read.
```

## Why the green runs are not reassurance

The only successful runs are `pull_request`-triggered. That lane is path-filtered to the watch's own
files and, per the script's own comment, *probes* the endpoint with the plain `GITHUB_TOKEN` to
measure what CI can actually reach — it is not the audit. So the workflow's run list reads as
"mostly red with some green", and the green is from the lane that does not do the work.

## To the workflow's credit

It fails **loudly and honestly**. The exit path is explicit that this is not a clean verdict:

> `The watch could not answer. That is a failure of THIS WATCH, not a 'nothing was bypassed' verdict.`

That distinction is exactly right and is why this is filable rather than invisible. The gap is that
nothing acted on it: it has been red on every real run for as long as the history goes, and the
control it provides — the only automated check that a merge did not bypass a red required gate — has
therefore been absent that whole time.

## Fix

`RULESET_AUDIT_TOKEN` needs `Administration:read` on the repo (the `rule-suites` endpoint requires
it, and a workflow `GITHUB_TOKEN` cannot hold it — the script documents this at line 77). Either
grant it, or rotate in a token that has it.

Worth adding alongside: an alert on *this* control being unable to answer. A watch that is red for
87 consecutive real runs is indistinguishable from routine CI noise unless someone opens it, and the
whole point of it is to notice something nobody else will.

## How this was found

[...]

PR #5883, opened 2026-08-20
ci(governance): pin who may bypass main-protection, and fail a PR on widening

JiRaska:
## What this is

An investigation of #4828 that ended with one gate and one documentation change.

**#4828 itself is untouched.** Its body says it is refreshed in place while offending merges
remain in the window and closed by the watch when the window comes back clean, and that rows
age out on their own. So no row here was "fixed" and the issue is not closed by this PR.

## 1. Is the watch healthy? Yes — established by effect, not by reading it

- Run `32357808433` (today, `PERIOD: month`) produced a real `dirty` verdict from live data and
  refreshed #4828 with `25 bypass suites examined`. A blind watch cannot render that table.
- Its `--self-test` step passed in that same run: *"4 signal + 7 benign detail strings, dedup,
  rule-type scoping, suite-result scoping and 5 declaration falsifications"*.
- The `month` window is still passed explicitly (`PERIOD: month` in the run log), so the
  `time_period=day` trap the body warns about is not in play.
- **The negative case is already proven in production history.** The blind-detection path
  (#4791) has fired and self-closed six times — #4815, #5098, #5251, #5280, #5304, #5322 — so
  both branches of the watch are known to work, not just the one that says "clean". That is the
  strongest available evidence that a lost token would not read as a clean window forever.
- Independently confirmed the token's subject is readable: querying `rule-suites?time_period=month`
  returns `{('JiRaska','pass'): 49, ('openbank-gitops-bot[bot]','pass'): 49, ('JiRaska','bypass'): 2}`.

Nothing about the watch needed changing.

## 2. Is recurrence prevented, or only detected? Only detected — and half the mechanism had no reader

The live `main-protection` ruleset carries exactly one bypass actor:

```
{"actor_id": 5, "actor_type": "RepositoryRole", "bypass_mode": "pull_request"}
```

RepositoryRole 5 is `admin`, and `current_user_can_bypass` reads `pull_requests_only` for the
credentials every agent in this repo runs as. The bypass rows in #4828 are all actor `JiRaska`,
never the bot. So the surface is reachable by automation, which is exactly the incident the
repo's own guardrails record: an agent taking the override unprompted on a money-path security PR.

**The gap:** the watch reads the bypass LOG. Nothing read the bypass CONFIG. `bypass_actors` is
editable from the Settings UI in two clicks and produces no diff, no PR and no red check anywhere
— the same write-only-control shape #4240 exists for, one level up.

## 3. What this PR adds

`ruleset-bypass-actors` (enforced, `lint` group), a sibling of `ruleset-context-parity` sharing
its transport.

It **cannot** prevent the override, and says so in its own header: GitHub evaluates the bypass at
merge time, after every check has reported, so nothing in CI can. What it does is make a
**widening** of the surface reviewed instead of ambient. Both directions are checked — a stale
declaration that describes nothing is how a pinned list stops being a pin — and `bypass_mode` is
part of the compared tuple, because `pull_request` to `always` keeps the same `actor_id` and is
the difference between "can merge a PR past a red check" and "can push straight to main".

No permission grant: `GET .../rulesets/{id}` is world-readable on a public repo and includes
`bypass_actors`. Deliberately not the endpoint the watch needs — `rulesets/rule-suites` is
private and rejects fine-grained tokens (#4791), which is precisely why the log half can only be
a scheduled watch while the config half can be a PR-blocking gate.

### Verified by effect, against the real live ruleset

| case | expected | result |
|---|---|---|
| shipped declaration vs live API | green | `1 live, 0 undeclared, 0 stale`, rc=0 |
| declaration emptied | red | live admin actor reported undeclared |
| `bypass_mode` pinned to `always` | red both directions | undeclared + stale both reported |
| offline self-test | 8 cases pass | 2 of them run against the real declared set, so it cannot pass vacuously |

Also run: `run-gates.py --only ruleset-bypass-actors` (PASS), `ruff E9,F,B`, and the meta-gates
`gate-script-registration`, `gate-selftest-declaration`, `gate-lifecycle-metadata --today`,
`gate-subject-floor`, `gates-not-deregistered`, `gate-invocation-reachability` — all rc=0.
`gate-lifecycle-metadata` initially caught this gate missing `rationale:`/`review_after:`; both
are now declared rather than baselined as debt.

## 4. What is NOT in this PR, and needs the owner

**No repository settings or ruleset configuration were changed**, by design. Two things are worth
the owner's decision and cannot be done from a PR:

1. **Narrow the bypass actor.** `bypass_mode: pull_request` on the `admin` role is the broadest
   sensible setting short of removing it. Removing the bypass entirely would mean no override
   path at all; the honest middle is leaving it and relying on the two detectors. Owner's call.
2. **`required_approving_review_count` is `0`.** Every merge past a red check in #4828 was also a
   merge with zero human approvals — worth noting because the two controls are usually assumed to
   backstop each other here, and neither does.

Refs #4828


JiRaska:
## Review — reviewer, not author. Comments only; nothing merged, approved or pushed. I read this PR; I did not exercise any bypass, change any setting, or touch the ruleset.

3-dot diff against `origin/main`; 243 behind, additive-only (307 lines, 0 deletions).

The control is honest about its own limits, which is the rarest property in this class of change: it states plainly that it **cannot** prevent the override (GitHub evaluates the bypass at merge time, after every check has reported) and that what it buys is making a widening reviewed rather than ambient. That framing is correct and I would not water it down.

**The scope is derived, not hand-kept** — `live_bypass_actors()` enumerates every active branch ruleset covering `main` from the live API and compares against the declaration. The hand-kept half is the *declaration*, which is the correct half to maintain by hand, because a declaration is a claim a human must justify. The two-direction check is what makes that safe.

### The negative controls are present, and one of them does more than the PR claims

`self_test()` runs 8 cases, and two of them are the ones that matter:

- `widening bypass_mode pull_request -> always is a DIFFERENT actor tuple` — this is the case a comparison keyed on `actor_id` alone would call clean, and it is the difference between "can merge a PR past a red check" and "can push straight to `main`". Tested in both directions (`[admin_always]` vs `[admin]` reports the new tuple undeclared **and** the old one stale).
- `findings([admin, a_bot], DECLARED_BYPASS_ACTORS)` runs against the **real** declared set, not a fixture — so a comparison stubbed to `return [], []` fails the self-test. This is what the manifest's `selftest_expect: pass` is worth here, and most gates in the estate do not have it.

Beyond that, the `stale` direction is a stronger control than the PR body says: if `main-protection` were **deleted**, or flipped to `evaluate` enforcement, `live` becomes `[]`, `undeclared` is empty, and the gate still reddens because `(5, RepositoryRole, pull_request)` is now stale. Combined with `min_subjects: 1` — which `run-gates.py:397-403` enforces by failing a gate that declares a floor and prints fewer subjects — this gate cannot go green about a ruleset that has stopped existing. That is worth stating explicitly in the rationale; it is the answer to "what does this prevent" and right now the reader has to derive it.

### Two things to check before merge

**1. An org-level ruleset red-gates every PR the day one appears.** `GET /repos/{owner}/{repo}/rulesets` includes parent (org/enterprise) rulesets by default. `live_bypass_actors` then calls `gh api repos/{repo}/rulesets/{summary['id']}` for each — and a repo-scoped detail fetch for an **org** ruleset id answers 404, which `gh_api` turns into `RuntimeError` and `main()` into `return 1`. Failing closed is the right default, but the failure would be an opaque `gh api ... failed (rc=1)` on every PR at once, with nothing pointing at the cause. Either pass `includes_parents=false` and say why org rulesets are out of scope, or handle the 404 with a message that names the situation.

**2. Confirm the gate's runner supplies a token.** `gh api` in Actions requires `GH_TOKEN`/`GITHUB_TOKEN` in the environment even for world-readable endpoints — unauthenticated it errors out rather than falling back to anonymous. The header argues (correctly) that no *permission grant* is needed, which is a different claim from no token being needed. `check-ruleset-context-parity.py` shares the transport and presumably works today, so this is probably fine — but the manifest entry's `run:` block does not declare an `env:`, so please confirm the surrounding job does, and ideally say so in the header next to the "no token at all" sentence, which as written slightly overstates it. If it were ever unauthenticated the gate fails red, not green, so this is a diagnosability issue rather than a safety one.

### Smaller
- `int(actor.get("actor_id") or 0)` coerces a null `actor_id` to `0`. GitHub sends `actor_id: null` for `OrganizationAdmin` / `EnterpriseOwner` actor types. The coercion is harmless — `(0, "OrganizationAdmin", "always")` is still a distinct tuple and is reported undeclared — but a comment saying that is deliberate would stop a future reader "fixing" it into a skip.
- `review_after: "2027-02-20"` with a rationale naming the re-examination trigger ("when the repo gains an org owner or a GitHub App that legitimately needs one") is exactly right, and rarer than it should be.
- The `.github/CLAUDE.md` addition is accurate, public-safe (no token values, no account specifics) and captures the genuinely non-obvious half: same feature, two endpoints, opposite access, and that a fine-grained token's `Read access to administration` still 403s on the log endpoint with a message that names no permission. That bullet will save someone the debugging session it cost.
- The last bullet — "Every agent here runs as an identity that CAN bypass" — is the load-bearing one. `current_user_can_bypass` reading `pull_requests_only` rather than `never` is the fact that makes the prompt-level prohibition on override flags a real control rather than belt-and-braces. Good that it is written down where an agent will read it.

### Checked and clean
- Read live, never from a checked-in mirror — no snapshot of the ruleset is committed anywhere in this diff.
- `enforcement != "active"` and `target != "branch"` are both skipped with the reason stated (an `evaluate`-mode ruleset's bypass list grants nothing).
- `gh_api` collapses missing binary, network error, rate limit and non-JSON into one `RuntimeError`, and every one of them produces `::error::` + exit 1. A gate that cannot reach its subject says so instead of passing.
- `mode: enforced` with `min_subjects: 1` and a dated subject count comment — the gate cannot be quietly narrowed to nothing.

JiRaska:
Rebuilt on current `origin/main`, and then fixed a defect in the gate itself. Head `ac0642f8a`,
3-dot delta is exactly the three intended files (365 insertions, 0 deletions).

**The gate could never fire in CI — in either direction.**

`gates (lint)` was red on this PR with:

> `ruleset-bypass-actors: RepositoryRole#5 (bypass_mode=pull_request) is declared here but no
> longer holds a bypass on main. Remove the entry`
> `SUBJECTS=0  # bypass actors on an active ruleset covering main`

The declaration is correct and the finding is wrong. Measured both ways on 2026-08-21:

```
# admin token
GET /repos/JiRaska/open-bank-oss/rulesets/18325357
  bypass: [{"actor_id":5,"actor_type":"RepositoryRole","bypass_mode":"pull_request"}]

# unauthenticated (what a non-admin sees)
  keys: _links conditions created_at enforcement id name node_id rules source
        source_type target updated_at
  has bypass_actors: False
```

`bypass_actors` is **omitted** for a non-admin reader, not returned empty. `live_bypass_actors`
did `detail.get("bypass_actors", [])`, which turns *"I may not see this"* into *"there are none"* —
and CI runs as `GITHUB_TOKEN`, which is not an admin, so that was **every** CI run.

The stale-declaration noise is the mild half. The serious half is that the gate's whole purpose is
the other direction: a real widening of who may bypass `main-protection` would also arrive as zero
actors, so **the one event this gate exists to catch could never make it fire.** Its header even
asserted the opposite — *"`bypass_actors` is included in the detail response"* — which is what made
it look verified; that claim is now corrected in place with the measurement above.

Fix in this push: a missing `bypass_actors` key raises "cannot be evaluated" instead of resolving to
`[]`. A genuinely empty list still resolves to `[]` — the two are different facts and the gate now
distinguishes them. Two self-test cases cover exactly that, and they were falsified rather than
assumed:

```
# with the fix
self-test ok: ruleset-bypass-actors is falsifiable (8 cases, both directions)   rc=0
# reintroduce `.get("bypass_actors", [])`
::error::self-test: an unreadable ruleset (no bypass_actors key) must RAISE, not return []
self-test FAILED (1 case(s))                                                   rc=1
```

(Measured with a redirect, not through `| tail` — the pipeline's exit code is `tail`'s, and it read
`rc=0` on the broken build the first time I checked.)

**What still needs a decision from you.** The gate is now honest but still cannot pass in CI: with
`GITHUB_TOKEN` it will report "unreadable" every run. It needs one of

1. an admin-scoped PAT supplied as a secret, and the gate skipped-as-UNRESOLVED where none exists; or
2. wiring into the UNRESOLVED verdict #6293 is adding to `run-gates.py` — same shape as
   `ruleset-context-parity`, which hit the identical wall from the rate-limit side (#6290).

Option 2 looks like the natural home. What must not happen is the gate quietly exiting 0 when it
cannot look — that is the failure this PR just removed.

JiRaska:
## `gates (lint)` — the gate cannot be evaluated with the token CI gives it, so it can never go green on a PR

`Validate manifests` here is only the roll-up; the real shard is `gates (lint)`, and within it exactly one gate fails out of 21 (`21 gates  PASS=19  WARN=1  FAIL=1`) — this PR's own new gate.

### The decisive line (run 32527904798, job 96921446106)

```
##[group]FAIL [lint] who may bypass main-protection is pinned, and widening it fails a PR (Refs #4828, enforced)  (0.9s)
--- self-test (must PASS) ---
self-test ok: ruleset-bypass-actors is falsifiable (8 cases, both directions)
--- gate ---
##[error]ruleset-bypass-actors: ruleset 18325357 (main-protection) returned no `bypass_actors`
field — the token cannot see bypass actors (admin scope required), so this gate CANNOT be
evaluated. Not a finding: an unreadable subject is UNRESOLVED, never 'zero actors'. Run it
with an admin token, or skip it where none exists.
```

Note what did **not** fail: the self-test passes (8 cases, both directions), so the harness is sound and `live_bypass_actors` correctly distinguishes "unreadable" from "empty" — exactly as the script's own comment at line 139 intends. The gate is well built. The problem is the token it is handed.

### Why this is structural, not a flake

`.github/workflows/ci.yml:265` runs the gates job with:

```yaml
GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

The default `GITHUB_TOKEN` cannot read `bypass_actors` under any `permissions:` block — the field requires repo-admin scope, and a non-admin read omits the key entirely rather than erroring (which is precisely the case the script documents and raises on). So with `mode: enforced` in the `lint` group, **this gate fails on every PR, permanently** — not just this one. If it merges as-is it reddens the whole queue, and the `all-green` / `Validate manifests` required contexts go with it.

This is the "a check passes because it never reached its subject" shape, one step further along: it correctly refuses to report a false green, but the enforced verdict then blocks everything.

### Why I did not fix it

Every available fix is an owner decision I am not going to make unilaterally:

1. **Supply an admin-scoped token** — a new repository secret (a PAT or a GitHub App installation token with `administration: read`). Creating repository secrets is out of scope for me, and it widens what CI can read on a public repo, which deserves its own thought.
2. **Move the gate off the PR lane** — run it on `schedule` / main-push only, where a privileged token is defensible, and have it file an issue rather than redden a PR. This keeps the control and removes the queue blast radius.
3. **Declare it advisory** (`mode: advisory`) until 1 or 2 lands, so it warns instead of blocking.

Option 2 looks like the best fit for what the rationale actually wants — a repo *setting* drifting is not something a given PR author can fix, so blocking their PR on it has no remediation path for them. That is also the argument for not making it enforced-on-PR even once a token exists.

Separately, the narrative already in the gate's output — that `bypass_mode` on the `admin` role and `required_approving_review_count: 0` are both owner calls — is unchanged by any of this, and I have not touched the ruleset.

No changes pushed to this branch.

Refs #4828

JiRaska:
`gates (lint)` is red, and the gate is behaving correctly — it is the wiring that is missing.

```
##[error]ruleset-bypass-actors: ruleset 18325357 (main-protection) returned no `bypass_actors`
field — the token cannot see bypass actors (admin scope required), so this gate CANNOT be
evaluated. Not a finding: an unreadable subject is UNFALSIFIED.
```

The gate's self-test passed in the same run (`ruleset-bypass-actors is falsifiable (8 cases, both
directions)`), so its failure path works. It refuses to report green on a subject it cannot read,
which is the whole point — a non-admin read omits `bypass_actors` entirely, and "key missing" is a
different fact from "empty list". Everything else in the shard is fine: `21 gates PASS=19 WARN=1
FAIL=1`, and the five FAILs in the earlier self-test block are the must-fail fixtures.

**The cause:** the gate runs in the `gates (lint)` job, whose env is
`GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}` (`ci.yml:265`). The repo already has a
`RULESET_AUDIT_TOKEN` secret (created 2026-08-14) for exactly this class of read, but it is wired
only into `merged-past-red-check-watch.yml:125` — a different workflow. The gate's own docstring
anticipated this: *"Run it with an admin token, or skip it where none exists."*

Two ways to close it, and the choice is a security decision rather than a wiring detail:

**A — move the gate to the existing watch lane** (what I'd recommend). `merged-past-red-check-watch.yml`
already holds the token and already audits ruleset state, so the gate lands next to its sibling with
no new secret exposure. Cost: the check moves off the PR lane, so a widening is caught on a schedule
rather than at PR time.

**B — pass `RULESET_AUDIT_TOKEN` into `gates (lint)`.** Keeps PR-time detection, which is what the
PR title promises. Cost: this repo's PR lane would then carry an admin-scoped classic PAT. Same-repo
branch PRs do receive secrets, and agents open PRs here continuously, so this widens what a
compromised or mistaken PR can reach — on a public repo. Not obviously wrong, but it is a real
change in exposure and should be a deliberate call, not a side effect of fixing a red check.

I have not applied either. B in particular trades away more than the red check costs, and A changes
what the PR claims to do, so both want your decision.

One thing worth stating plainly: until this is wired, the gate is not merely red — it is
**unfalsified**, and merging it as-is would add a check that can never pass and never evaluate its
subject. That is worse than not having it, because a permanently red required-adjacent gate trains
people to ignore the shard.

JiRaska:
**Triage (measured, no action taken)** — shared-registry drift on `.github/gates/gates.yaml`.

Counts, read directly off the refs:

```
git show origin/main:.github/gates/gates.yaml | grep -c '^\s*-\s*id:'                     # 185
git show refs/remotes/origin/chore/4828-bypass-watch:.github/gates/gates.yaml | grep -c '^\s*-\s*id:'   # 181
```

This branch adds exactly one gate id, `ruleset-bypass-actors`. `main` has meanwhile gained five it
does not have: `pr-build-cloud-credentials`, `spot-kill-retry-selftest`,
`synthetic-outbox-taint-persistence`, `synthetic-taint-rest-client-boundaries`,
`test-intelligence-ecosystem`. No id collides, so the merge should be textually clean — but this is
the exact shape where two sides adding neighbouring list entries merge clean and silently keep only
one. **After merging, assert the count is 186, not 185.**

Staleness: `behind_by=663..795` and the last CI run completed `2026-08-23T04:21:53Z`, so the two red
checks (`Validate manifests`, `gates (lint)`) are verdicts about a base from a week ago and prove
nothing about today's `main`. Rebase and re-run before reading them as real.

`.github/scripts/check-ruleset-bypass-actors.py` is absent from `main`, so this is not duplicated
work by a parallel PR.

JiRaska:
### Why this PR has been red since 23 August, and what changed

`Validate manifests` failed on one gate — this PR's own:

```
[run-gates]  18/25 FAIL  ruleset-bypass-actors (0.7s)
##[error]ruleset-bypass-actors: ruleset 18325357 (main-protection) returned no `bypass_actors`
field — the token cannot see bypass actors (admin scope required), so this gate CANNOT be evaluated.
```

Everything about that message is correct, and the gate's refusal to read a missing field as "zero
actors" is the right call — defaulting it to `[]` would report every honest declaration as stale and,
far worse, let a real widening arrive as zero actors.

**The defect is the exit code, not the reasoning.** The unreadable case raised the same
`RuntimeError` that a broken `gh api` call raises, and `main()` mapped any `RuntimeError` to exit 1.
CI runs as `GITHUB_TOKEN`, which is not an admin, so the field is invisible on **every** CI run. An
enforced gate that has never once been evaluable was therefore red on every pull request in its
group — including this one, which is exactly why it never went green.

Measured both directions before changing anything:

| token | `bypass_actors` visible | gate |
|---|---|---|
| local admin | yes — `[{actor_id: 5, RepositoryRole, pull_request}]` | exit 0, `1 live actor, 0 undeclared, 0 stale` |
| CI `GITHUB_TOKEN` | field omitted entirely | exit 1, every run |

**The fix splits the two conditions.** A new `Unreadable` means "the subject could not be READ" and
answers `gatelib.subjects_unresolved(...)`, a warning, and exit 0. That is not a pass: the run states
that nothing was checked, and the runner's floor logic is told the corpus was never read, so it
cannot be mistaken for a clean run. A genuinely broken API call still exits 1, and an undeclared live
actor still fails under `--enforce`.

Treating an unreadable subject as red makes the gate permanently and uninformatively red; treating it
as green is a lie. UNRESOLVED is neither.

**The self-test now asserts exit codes, not exception types.** The old one checked that the
unreadable case *raised* — which it did, all along, while the gate stayed red. Checking the type
could never see this defect. Four new cases, in both directions:

```
an unreadable subject must exit 0 (UNRESOLVED)
an undeclared live bypass actor must exit 1 under --enforce
a broken API call must still exit 1
an unreadable ruleset must raise Unreadable, not a bare RuntimeError
```

Verification run locally:

```
python3 .github/scripts/run-gates.py --only ruleset-bypass-actors
  1 gates  PASS=1   cpu=1.1s
  PASS  ruleset-bypass-actors  1.1s  [lint]

self-test ok: ruleset-bypass-actors is falsifiable (12 cases, both directions, exit codes included)
```

plus `gate-selftest-declaration`, `gate-subject-floor`, `gate-lifecycle-metadata`,
`gate-runner-self-test` and `python-lint`, all PASS.

### One limitation, stated rather than hidden

On a PR this gate can now only ever report UNRESOLVED, because the CI token cannot see the field. It
has teeth where the token does — a local run, or a scheduled job with an admin PAT. That is a real
narrowing of where the check applies, and it should be a deliberate choice by whoever merges this,
not something discovered later from a gate that always says the same thing. The alternative — giving
CI an admin-scoped token so the gate can read `bypass_actors` on every PR — buys enforcement at the
cost of putting an admin credential in every pull-request run, which for this repository is the
larger risk.

JiRaska:
Brought onto current `main` as well. The branch was **1005 commits behind** with zero conflicts, so
the green I reported above was measured against a three-week-old main — and this PR changes gate
machinery, which is where that gap matters most: a gate passing against an old gate set is not
evidence about the set as it stands today.

Re-verified on the merged result, not just the branch:

```
run-gates.py --only ruleset-bypass-actors      PASS=1
gate-selftest-declaration                       exit 0
gate-subject-floor                              exit 0
gate-lifecycle-metadata                         exit 0
python-lint                                     exit 0
```

The three-file delta against main is unchanged: `.github/CLAUDE.md`, `.github/gates/gates.yaml`,
`.github/scripts/check-ruleset-bypass-actors.py`.

PR #5884, opened 2026-08-20
feat(ci): fuzz the surface authorization hides, not just the 401s

JiRaska:
> Stacked on #5763 (`fix/fuzz-readiness-pid-race`). Base is that branch, not `main` — this builds directly on its three harness fixes. Merge #5763 first; this diff then reduces to the two-pass change.

## The measurement that forced this

Full-fleet run [32348238171](https://github.com/JiRaska/open-bank-oss/actions/runs/32348238171) came back **26/26 green**. Then I counted what each job had actually exercised, from its own `Selected: N/M` against its `Authentication failed: N operations`:

```
operations selected total:  306
auth-blocked (401/403):     302
ACTUALLY exercised:           4   (1.3%)
services with zero unauthenticated surface: 24 of 26
```

ledger 28/28 blocked. lending 37/37. account 20/20. party 24/24.

Both real 5xx this lane has ever found came from the 3 non-blocked operations on consent-service — so its **entire yield to date comes from 1.3% of the surface it reports on**. A green was being read as "fuzzed" when it meant "authenticated" (#5769).

## Two passes, both must pass

| pass | what it proves |
|---|---|
| **authz ON** (unchanged) | an endpoint that *loses* its `@RolesAllowed` starts answering 2xx to anonymous traffic and shows up here |
| **authz OFF** (new) | the input-hardening property the `pentest` dimension actually wants |

Pass 2 does **not** test authorization — it is switched off. That is pass 1's job, which is why both run rather than one replacing the other.

## Falsified, not asserted

Dispatched from this branch against two services ([32359570690](https://github.com/JiRaska/open-bank-oss/actions/runs/32359570690), both green):

```
==> [openbank-sca-service]     authz ON:  Selected: 6/6   auth-blocked=6
==> [openbank-sca-service]     authz OFF: Selected: 6/6   auth-blocked=0

==> [openbank-consent-service] authz ON:  Selected: 13/13 auth-blocked=10
==> [openbank-consent-service] authz OFF: Selected: 13/13 auth-blocked=0
```

**sca-service went from testing nothing — 6 of 6 blocked — to testing all six.** consent went 3 → 13. Both passes green on both services.

## Why this flag and not Keycloak

`quarkus.security.auth.enabled-in-dev-mode` is a built-in dev-mode switch: no dependency, no service edit, no realm import, and none of the version-specific realm-shape traps CLAUDE.md documents for Keycloak imports.

The alternative was tried and is **worse than useless**: basic auth with embedded users is accepted as configuration while no `IdentityProvider` is on the classpath, so *every* request answers 500 —

```
BasicAuthenticationMechanism requires one or more IdentityProviders
supporting at least one [UsernamePasswordAuthenticationRequest]
```

Making that work needs an elytron extension on ~26 money-path classpaths. That is a security-surface change, not a harness change.

## The constraint that decides whether pass 2 can be trusted

Redis is started when the service's **own config** references it — derived, never a list kept in the workflow. 31 of ~50 services do.

The auth-on pass never needed it, because a 403 fires before the handler touches any dependency: **this lane has been accidentally immune to missing infrastructure for its whole life.** Pass 2 is not immune, and that is the single thing deciding whether it can be trusted. With Redis absent, `PATCH /api/v1/consents/approvals/{id}` answers 500 —

```
AnnotatedConnectException: Connection refused: localhost/127.0.0.1:6379
```

— which is indistinguishable from a real finding at the schemathesis layer. With Redis up, the same request answers 404. I hit exactly this while prototyping and briefly recorded it as a defect; correlating the response's traceId against the log, instead of eyeballing a nearby line, is what corrected it. A missing dependency must never be able to masquerade as a finding on the control that exists to surface real ones.

Kafka deliberately **not** started: writes go to the transactional outbox in postgres, not to a broker on the request path, so its absence does not 5xx a request.

## Each pass now states what its green is worth

```
==> [svc] authz ON:  Selected: 13/13 auth-blocked=10
==> [svc] authz OFF: Selected: 13/13 auth-blocked=0
```

Previously this had to be reconstructed from artifacts — which is how the 1.3% figure was found in the first place, by hand, after the fact.

## Cost

One extra boot per service (~26–60s observed); the compile is already shared between passes. Weekly lane, `fail-fast: false`.

## Verification

- falsification run above — both services, both passes, green
- `actionlint` finding sets compared against `origin/main`: one `SC2034` on each side, **no new finding**
- YAML parses

## Linked issues

Refs #5769, #5714


JiRaska:
**Fourth commit: the census was blind to exceptions that gained a mapper.**

It grepped the literal phrase `Unhandled exception` — what `GenericExceptionMapper` prints. Run [32501086843](https://github.com/JiRaska/open-bank-oss/actions/runs/32501086843) returned an **empty census** for dispute, interest and sdd while 78 exceptions had been thrown, mapped and ERROR-logged by the new persistence mappers (#6240): 64 `ConstraintViolationException` in dispute, 12 `DataException` in interest, 2 in sdd.

An empty census read as *"nothing happened"* when the truth was *"everything was handled"*. The same silence, two opposite meanings — which is precisely the failure this census exists to prevent. It had acquired the defect it was written against.

Filtering on the log **level** instead of one mapper's phrasing sees both.

Two details that each cost a wrong answer on the way:

- **Strip ANSI colour BEFORE filtering on level.** The runner interleaves escape codes, so ` ERROR ` does not match literally on a raw line — my first attempt returned an empty census for logs full of errors, a broken probe giving the reassuring answer.
- **Require the trailing `:` of `FQCN: message`,** and drop the bracketed logger field. Without it `[co.op.li.ap.er.GenericExceptionMapper]` matches as `GenericException` and the census reports the logger's own name as a finding.

Verified against both archived runs rather than reasoned about — the new pipeline reproduces the pre-mapper counts from 32368751763 and the post-mapper counts from 32501086843, and both match what I had counted by hand out of the artifacts.

JiRaska:

3-dot diff against `origin/main`; 320 behind, `MERGEABLE`, one file.

This is the best-evidenced control change I have read in this repo. The `Selected: N/M` vs `Authentication failed: N operations` measurement — 306 operations selected, 302 auth-blocked, **1.3% actually exercised**, 24 of 26 services with zero unauthenticated surface — converts "this lane is green" into "this lane was green about nothing", and it is a measurement rather than an argument. The negative controls are real and cited by run id: Redis absent produces a `Connection refused` 500 indistinguishable from a finding; the PDP-unreachable pass produced 66 of 83 fake server errors; the `grep -q` / SIGPIPE / `pipefail` self-suppression was found by running the pipeline under the options the step actually sets. Three findings, one worth fixing before merge.

### 1. The Redis dependency has no hard failure, and it is the exact dependency the comment says must never masquerade as a finding

`.github/workflows/api-fuzz.yml:213-232`. Postgres gets a hard gate:

```
if [ "$PG_UP" != "1" ]; then
  echo "::error::[${svc}] postgres never became ready — skip"
  OVERALL=1
```

Redis gets none:

```
docker run -d --name fuzz-redis -p 6379:6379 redis:7-alpine >/dev/null
for i in $(seq 1 30); do
  docker exec fuzz-redis redis-cli ping >/dev/null 2>&1 && break
  sleep 1
done
```

If the image pull fails, the port is taken by a previous service's container that `docker rm -f` did not clear, or Redis simply never answers, the loop exhausts its 30 attempts and **falls through silently**. Both passes then run against an absent Redis and produce precisely the `AnnotatedConnectException: Connection refused: localhost/127.0.0.1:6379` 500 that the comment three lines above calls out as indistinguishable from a real finding — with, this time, nothing in the log saying the dependency was missing. Give it the same shape as `PG_UP`: a `REDIS_UP` flag, an `::error::` and `OVERALL=1`. The exception census will show the class, but the census is explicitly documented as "a triage aid, NOT a verdict", so it cannot be the thing that catches this.

### 2. `grep -qi "redis" "$APP_YAML"` — a derived scope, but derived from one file's text

Better than a hand-kept list, and I would keep it. Two things it cannot see, worth a sentence in the comment so the next person does not read its silence as absence:
- a service that gets Redis from a libs default, a `%prod` include, or an env var rather than a literal in its own `application.yaml` — no container starts, and the masquerading 500 returns;
- a service that merely *mentions* redis in a comment or an unrelated key — a container starts for nothing (cheap, harmless).

The stated denominator ("31 of ~50 services reference it") is a count from this same grep, so it does not independently confirm the grep's reach. If it is easy, cross-check the count against the services that declare a `quarkus.redis` config key or a `redis-client` dependency in `build.gradle.kts` — a different method, which is what makes a denominator worth something.

### 3. Wall-clock and cost are doubled and not quantified

Every service now boots twice, and the readiness wait deliberately no longer uses `kill -0 "$DEV_PID"` as a liveness proxy (correctly — the Gradle client exits while the daemon keeps quarkusDev alive; run 32284304038 is convincing). The consequence is that a service that genuinely dies on boot now burns the **full** 90×2s wait, twice, instead of aborting early. Across ~26 services that is a materially longer job on a lane that already runs at fleet scale. Two things would help the reviewer: the measured before/after job duration, and whether the pass-2 boot could reuse the pass-1 build cache (it should, since the compile is hoisted out — but that is worth stating rather than assuming). No objection to the cost, only to it being unstated.

### Smaller
- Both passes share one Postgres database with no reset between them. Pass 1's fuzz writes rows; pass 2 then fuzzes a dirty schema, so a unique-constraint 500 in pass 2 may be an artifact of pass 1 rather than a defect. The census helps triage it but a `docker exec fuzz-pg dropdb/createdb` between passes would remove the question.
- `run_pass` sets `OVERALL=1` from inside a function — correct as a global here, but it is the one piece of state that decides the job's verdict and it is assigned in three places. Worth a comment marking it deliberately non-`local`.
- The comment block at `:280-289` says "the 180s budget below" — the budget in the loop is `seq 1 90` × `sleep 2` plus curl time, so the effective wall-clock is somewhat above 180s. Cosmetic.

### Checked and clean
- Pass 2 disables **both** `quarkus.security.auth.enabled-in-dev-mode` and `authz.enforce`, with the reason stated: `AuthorizeInterceptor` fails closed without a PDP (ADR-0034) and no OPA sidecar runs here, so leaving `authz.enforce` on would have manufactured 503 `POLICY_DECISION_POINT_UNAVAILABLE` as "server errors". Correct, measured, and it does not weaken pass 1, which still runs and still must pass.
- The scope of what pass 2 claims is stated honestly and narrowly ("does a handler answer 5xx to input it should reject with a 4xx") rather than overclaimed as authorization coverage.
- The census is deliberately **not** turned into an exclusion filter, with the reason given and measured (balance-service: 2 environmental + 4 genuine `DataException`; ledger: 2 + a real NPE). That is the right call and the temptation to "fix the noise" by suppressing it is the one that would have lost real findings.
- Forcing `quarkus.datasource.username`/`password` as system properties rather than trusting env-var spelling is right — system properties outrank `${...:default}` expressions in SmallRye's ordinal chain, and the settlement-service counterexample shows why an assumption about naming was worth nothing.
- The census pipeline builds into a variable and tests the variable, rather than running the same pipeline twice inside an `if` — which is what made the `grep -q` SIGPIPE bug possible. Fixed at the cause, not the symptom.

JiRaska:
**Not a duplicate of the sibling PR, but a collision worth reading before either merges.**

#5763 and #5884 are the only two open PRs whose entire diff is `.github/workflows/api-fuzz.yml`, and
their hunks overlap:

- #5763 — `@@ -248,9 +248,60 @@` and `@@ -260,13 +311,35 @@`
- #5884 — `@@ -209,6 +209,28 @@` and `@@ -248,55 +270,239 @@`

Both rewrite the region starting at line 248, from different premises: #5763 fixes the readiness
wait aborting on a healthy service, #5884 widens what the fuzzer exercises past the 401s. Neither
contains the other, so both are wanted — but the second to merge will conflict, and a conflict
resolved by taking one side wholesale silently reverts the other.

Suggested order: land #5763 first (it is the smaller, behaviour-preserving fix), then rebase #5884
onto it and re-read the merged step, rather than resolving the conflict hunk by hunk.

Neither is mergeable by an agent here in any case — a change under `.github/workflows/` alters what
runs on the self-hosted runners, so both need a human.

JiRaska:
**This PR's base is stale in a way that makes its `DIRTY` status misleading — and I am deliberately
not "fixing" it by rebasing, because the right move is a decision, not a merge.**

Measured:

```
base                : fix/fuzz-readiness-pid-race   (#5763's branch, head e564e45b0)
git merge-base --is-ancestor <5763 head> <this head>   ->  NOT an ancestor
git diff --name-only origin/main...<this head>         ->  2 files
```

So GitHub calls this `DIRTY` against a parent whose current head this branch does **not** contain —
#5763 moved on (`fix(ci): fail fast on a fuzz compile error instead of mislabelling it a boot
failure`) after this branch forked. The conflict is on `.github/workflows/api-fuzz.yml`, which both
edit.

Against `main` the delta is small and self-contained: `api-fuzz.yml` shrinks by 138 lines and the
runner moves to a new `.github/scripts/fuzz-services.sh`.

I also checked whether #5763's readiness fix is already carried here, since that would decide
whether the stack is still needed at all — it appears to be, but split across the extraction
(`api-fuzz.yml` 1 mention, `fuzz-services.sh` 2, versus 4 in #5763's single file). "Appears to be"
is not good enough to act on: an extraction can carry the words and lose the behaviour, and I have
no way to run the fuzz workflow to tell the difference.

**Two options, both yours:**

1. If this branch already subsumes #5763, retarget its base to `main` and close #5763 as
   superseded — then there is no stack and no conflict.
2. If they are still meant to be a stack, rebase this onto #5763's **current** head and resolve
   `api-fuzz.yml` by hand, reading the merged file rather than taking a side.

An earlier comment of mine on this PR called #5763/#5884 a *race* between two independent PRs. That
was wrong — they are a stack with a stale base. The merge-order advice in it still holds; the
framing did not.

Note neither is mergeable by an agent regardless: a change under `.github/workflows/` alters what
executes on the self-hosted runners, so both need a human.

JiRaska:
Semantic collision with #5849, flagged there in detail: it adds a `Record exercised surface` step globbing `*-fuzz.log`, which after this PR matches only the **authz-ON** pass — where operations are auth-blocked by design. The exercised-surface floor would then be computed from the pass that deliberately tests nothing.

Verified the glob empirically: `*-fuzz.log` does **not** match `<svc>-fuzz-authz-off.log`.

Whichever lands second should glob `*-fuzz-authz-off.log` with a fallback to `*-fuzz.log`, so both stay independently mergeable in either order.

JiRaska:
**Conflict with main resolved — and my own earlier "restore" corrected.**

#5763 merged, so this rebased onto it. Two things went wrong on the way and both are worth recording, because the second one is a mistake I made while fixing the first.

**1. My restore commit `8a95c6312` reverted someone else's fix.** Commit order on this branch:

```
19200c292  my pipefail fix                            (inline script)
0cf45d1e2  extract fuzz-services.sh to keep the workflow parseable   <- someone else
8a95c6312  "restore the two-pass harness"             <- took the file from 19200c292
```

My restore checked out `api-fuzz.yml` wholesale from a commit that predated the extraction, undoing it. I even *verified* the restore — but I checked that **my** content was present, not that nobody else's was missing. That is the same failure I opened that commit's message complaining about.

**2. It matters more than a style revert.** The extraction exists because the inline `run:` step had grown past GitHub's undocumented ~20 KiB limit — which does not report as a size error, it makes the **whole workflow unparseable and every run shows zero jobs**. My accumulated comments are what pushed it over. Measured here:

```
after my restore:      27 871 bytes   (broken)
after my bad merge:    29 085 bytes   (broken)
extracted shape:        4 659 bytes   (fine)
```

So the branch tip has been in a state that would have produced a silently dead workflow, and my first attempt at resolving the conflict made it worse by re-inlining.

**Resolution now:** workflow keeps the one-line `run: .github/scripts/fuzz-services.sh`; the two-pass logic, census, `authz.enforce=false` and the Redis derivation all live in the script, where the extraction had already correctly put them. #5763's fourth commit — the compile fast-fail with its own `${svc}-build.log` — is grafted **into the script** rather than the workflow, for the same size reason.

Verified after merging rather than assumed: every property present in the script, workflow's largest `run:` block back to 4 659 bytes, `bash -n` clean, actionlint finding set unchanged from main's baseline (one SC2034).

Issue #5256, opened 2026-08-17
18+ producers omit sourceService, same silent-UNKNOWN audit-attribution gap as #3994


JiRaska:
## What

Issue #3994 fixed domestic-payment's missing `eventType`/`sourceService` fields (124 audit rows landing as `UNKNOWN`/`unknown`, PR #5255). A fleet sweep done alongside that fix found the same gap essentially everywhere else.

`sourceService` is set in only 2 producers fleet-wide: `customer-edge` (all events) and `lending-service` (3 of ~9 event types). Zero `sourceService` in: account-service, transaction-service, balance-service, party-service, kyc-service, consent-service, clearing-service, security-scanner, card-issuance-service, dispute-service, sepa-payment, statement-service, sanctions-service, sepa-instant, fx-service, document-service, swift-service, plus lending's remaining 6 event types.

Per `AuditConsumer.kt`, an event with no `sourceService` field falls back to a **silent, successful** default (`?: "unknown"`) — same failure mode as #3994: no error, no metric, no log line, only visible by grouping `audit_entries` by `source_service` on a live database.

## Why filed separately

#3994's own fix was deliberately scoped to domestic-payment only (the one producer with confirmed live evidence — 124 real rows) rather than expanding into an 18-service sweep in one PR. This issue is that follow-up, tracked so the finding isn't lost.

## What would close this

Per-service, add `sourceService` (and `eventType` where also missing) to each producer's event data classes, following the pattern #5255 established: SCREAMING_SNAKE_CASE for `eventType` matching the fleet's dominant convention (`BalanceEventType.BALANCE_UPDATED`-style), `sourceService` = the module directory name minus the `openbank-` prefix. Verify each fix the same way #5255 did — a real wire-payload test proving `AuditConsumer` resolves `sourceServiceSource == AttributionSource.EVENT`, not the fallback.

Not a single PR — each producer is its own small, independently-reviewable change, same discipline as the #4007 outbox sweep.

Refs #3994, #5255

JiRaska:
Fleet sweep complete (code-side) as of this comment — every producer in `EventAttribution.kt`'s table now has a PR adding `sourceService`:

[...]

PR #5888, opened 2026-08-20
fix(sepa-payment): add sourceService to the Temporal-path event payloads

JiRaska:
## Summary

Adds `sourceService` to the five **hand-built** event payloads in
`SepaPaymentActivitiesImpl` (the Temporal path): `VALIDATED`, `REJECTED`
(sanctions), `REJECTED` (scheme), `PROCESSING`, `COMPLETED`.

The #5256 fleet sweep fixed this service's **non-Temporal** path only
(`SepaPaymentEvents.kt`, a serialised data class). These five payload strings
reach the *same* `openbank.sepa.payment.events` topic through the same outbox and
the same `KafkaSepaPaymentEventPublisher`, and carried no `sourceService` at all —
so `AuditConsumer` fell through to its topic-derived table for every one of them,
recording `AttributionSource.TOPIC` rather than the producer's own claim.

`sepa-payment` is a money-path service (`rules.yaml: money_path_services`) — this
PR needs 2 approvals.

### Why the sweep missed it

Event payloads are built two ways in this repo, and each of the two obvious probes
is structurally blind to one of them:

- a grep for `"sourceService"` finds hand-built maps and **cannot** see a
  serialised data class (the key exists only as a Kotlin property name);
- reading the event data class finds the serialised path and **cannot** see
  hand-built payload strings.

This service has one of each, on one topic, so whichever probe was used returned a
clean-looking answer about the half it could see. The gap was found by enumerating
every outbox write site that reaches the audited topic, rather than by either grep.

## Verification — the wire payload, not the field

A test asserting a field on a data class does not prove the wire carries it, and
here there is no data class at all. Both halves are asserted:

- **Producer side** — `SepaPaymentActivitiesImplTest`: drives the real activity
  methods and reads `sourceService` off the payload the production code actually
  builds, for all five payloads, via parsed JSON (not a substring `contains`,
  which would also pass on a nested key or a prefix match).
- **Consumer side** — `AuditAttributionTest`: feeds that same wire shape through
  `AuditConsumer.consume` and asserts it resolves
  `sourceServiceSource == AttributionSource.EVENT`, not the fallback.

**Falsification:** deleting `sourceService` from the `PROCESSING` payload turns the
producer test red (`BUILD FAILED`, the new test named as the failure). Restored
before commit.

## Test plan

- [x] `./gradlew :openbank-sepa-payment:ktlintCheck :openbank-sepa-payment:detekt` — green (both `ktlintMainSourceSetCheck` and `ktlintTestSourceSetCheck` confirmed to have run, not merely absent from a short-circuited task list)
- [x] `./gradlew :openbank-sepa-payment:test --tests "...SepaPaymentActivitiesImplTest"` — green
- [x] `./gradlew :openbank-audit-service:ktlintCheck :openbank-audit-service:detekt :openbank-audit-service:test --tests "...AuditAttributionTest"` — 27 tests, 0 failures, **0 skipped**
- [x] Negative case: field removed ⇒ producer test fails

## Not changed here

`occurredAt` was already correct on all five payloads (#3914).

Refs #5256

### DIFF
--- a/openbank-audit-service/src/test/kotlin/com/openbank/audit/application/AuditAttributionTest.kt
+++ b/openbank-audit-service/src/test/kotlin/com/openbank/audit/application/AuditAttributionTest.kt
@@ -196,6 +196,26 @@ class AuditAttributionTest {
     }
 
     @Test
+    fun `sepa-payment's Temporal-path payload is attributed to the producer too`(): Unit = runBlocking {
+        // The sweep above fixed only the non-Temporal path (SepaPaymentEvents.kt, a serialised data
+        // class). SepaPaymentActivitiesImpl builds five payloads BY HAND onto the same topic, and
+        // they carried no `sourceService` at all — so a grep for the quoted key found the fixed
+        // half and a reader of the data class found the other, and neither probe could see the gap.
+        // This is the exact wire shape that publisher now emits (see the producer-side assertion in
+        // SepaPaymentActivitiesImplTest, which reads it off the real built payload).
+        val entry = capturingSave()
+
+        consumer.consume(
+            """{"paymentId":"${UUID.randomUUID()}","status":"PROCESSING",""" +
+                """"occurredAt":"2026-03-01T12:34:56Z","sourceService":"sepa-payment"}""",
+            EventAddress(topic = "openbank.sepa.payment.events"),
+        )
+
+        assertThat(entry.captured.sourceService).isEqualTo("sepa-payment")
+        assertThat(entry.captured.sourceServiceSource).isEqualTo(AttributionSource.EVENT)
+    }
+
+    @Test
     fun `dispute-service's own sourceService wins, no longer falling to the topic table`(): Unit = runBlocking {
         // Issue #3994/#5256's dispute-service fix: DisputeService's hand-built outbox payloads
         // (dispute.opened / dispute.resolved / dispute.remediation_requested) and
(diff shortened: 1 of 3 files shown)


JiRaska:

**Premise confirmed still true today.** I checked the producer type itself rather than grepping for the quoted JSON key — a grep cannot see a serialised data class, so a no-hits result on this family is a false clean:

- `origin/main`'s `SepaPaymentActivitiesImpl.kt` contains **no** `sourceService` at all (5 hand-built payloads on `openbank.sepa.payment.events`).
- `origin/main`'s `TerminationService.kt` has `sourceService` on **one** builder only (the `credit.loan.transition` evidence payload, line 355); `emitDomainEvent` — which produces `loan.withdrawn` and `loan.accelerated` — still emits `{"loanId":…,"partyId":…,"occurredAt":…}` with no `sourceService`.

Neither gap has been closed by anything that merged in the last nine days, and #5888 and #5894 touch **disjoint** production files, so they are not duplicates of each other.

**One merge-order caveat.** Both PRs add a new test method to the same file, `openbank-audit-service/…/AuditAttributionTest.kt`, and that file has grown on `main` since these branches were cut (618 lines on both branches vs 646 on `main`). Two sides appending to one file is where git reports a clean merge and can still keep only one side. After merging the second of the pair, confirm **both** new test methods are present on `main` — the suite stays green either way, so a dropped test is silent.

Issue #4757, opened 2026-08-14
dr-restore-verify.yml still cannot restore: the recovery manifests landed (#4996), but there is no cluster-capable runner to run them on


JiRaska:
## What

`.github/workflows/dr-restore-verify.yml` cannot complete a restore. Two wiring gaps are written into the workflow itself, both currently causing it to `exit 1` rather than report a hollow PASS. This issue owns closing them.

Split out of #669 as item 2b. It was previously described as "already tracked at #4026" — **that is no longer true: #4026 was closed as COMPLETED on 2026-08-13T22:06:50Z**, and its closing comment resolves the *"these workflows have never run"* question, not the provisioning question below. Without this issue the gaps are untracked.

## Measured state (2026-08-14)

Workflow id `308235042`, run history read from the Actions API — **2 runs, both `workflow_dispatch`, both 2026-08-08**:

| run | conclusion | what it proved |
|---|---|---|
| 21:11:23Z | success | the hollow green — a gate set `run=false`, the restore step was **skipped**, and the job passed having restored nothing |
| 21:18:08Z | failure | after #4247 made it tell the truth: it now fails loudly naming what did not happen |

No runs since. So the harness is honest and the restore is still impossible.

## The two blockers, verbatim from the workflow on `origin/main`

1. **No runner with cluster access.**
   `# WIRING: change to the deploy-pool self-hosted label once it exists (kubectl + IRSA).`
   It runs on `ubuntu-latest`, which has no cluster access, and self-fails at the gate.
2. **The recovery manifests are not in-repo.**
   `::error::WIRING INCOMPLETE: CNPG recovery Cluster + ledger-service manifests for $NS are not yet in-repo (docs/bcp/automated-dr-restore.md). Failing rather than reporting a hollow PASS.`
   The intended verification shape is already written out as comments directly beneath that line.

## Scope

- [ ] A deploy-pool self-hosted runner label with `kubectl` + IRSA exists, and `runs-on` points at it.
- [ ] The CNPG recovery `Cluster` (bootstrap.recovery from the ledger-db barman S3 store) and the ephemeral `ledger-service` Deployment (`OUTBOX_DISPATCH_ENABLED=false`, OIDC off) are committed, per `docs/bcp/automated-dr-restore.md`.
- [ ] One green run whose restore step **actually executed** — verified by reading the step conclusions, not the job conclusion. A green with a skipped restore step is the exact defect #4247 fixed; do not re-introduce it.

[...]

Issue #4755, opened 2026-08-14
Run Scenario A of the money-path chaos drill against sandbox and record a measured RTO


JiRaska:
## What

Execute **Scenario A** of `docs/bcp/money-path-chaos-drill-procedure.md` against the sandbox and append the observed result to `docs/bcp/dr-test-log.md`.

Split out of #669 (opened 2026-07-09), which bundled three deliverables. Items 1 (load benchmark) and 3 (SLO definitions) are done and evidenced there; this issue owns the only remaining one, and owns it as a single physical act rather than as a programme.

## Measured state (2026-08-14, from `origin/main`)

- `docs/bcp/money-path-chaos-drill-procedure.md` exists (#2359, merged 2026-07-25) and declares itself *"procedure only, not yet executed"*. Scenario A and Scenario B are both fully specified, including what to record as RTO.
- `docs/bcp/dr-test-log.md` is **untouched since `b47647b56` (2026-06-30)** and carries exactly one entry: a **table-top** PITR walkthrough with `RTO achieved: N/A` and `RPO achieved: N/A`. No live drill has ever been recorded.
- `ADR-0151` is `delivery-status: planned`; `ADR-0242` (quarterly DR/chaos drill with measured RTO/RPO) is `decision-status: proposed` / `delivery-status: planned`. There is no `chaos/` tree and no Chaos Mesh manifest anywhere in the repo.
- The CNPG HA blocker cited in mid-July is cleared — every money-path Postgres cluster runs `instances: 2` — so Scenario B is *executable*, but it is not required by this issue (see Scope).

## Scope — Scenario A only

Force-delete a `transaction-service` pod while `perf/k6/money-path-write-benchmark.js` drives load, then record:

1. **Observed RTO** — wall-clock from pod deletion to the money path serving successfully again.
2. **Client-visible behaviour** — clean error/retry, or a partial/ambiguous state?
3. **Ledger invariants under failure** — `GET /api/v1/ledger/close/trial-balance` → is `balanced: true` after recovery?

Then append a `## YYYY-MM-DD` entry to `docs/bcp/dr-test-log.md` in the file's existing format, with `Type: live-failover` and real numbers in the RTO/RPO fields.

Scenario B (CNPG live failover during posting) is explicitly **not** required here. It is the expensive half and nothing in #669's acceptance criteria needs it; if it is run, it is a bonus second entry in the same log.

## Why this cannot be done by an agent

[...]

PR #5889, opened 2026-08-20
docs(governance): repoint ledger.restore_drill at the drill record, not the runbook

### LINKED ISSUE #5673: The fleet's only DR-drill attestation cites a PG-upgrade runbook, not a drill
## What

`openbank-libs/governance/attestations.yaml` carries exactly one drill attestation for the whole fleet:

```yaml
ledger:
  restore_drill: { date: 2026-07-26, ttl_days: 180, by: jiri, ref: runbook-0003 }
```

Two things are wrong with it as *evidence*:

1. **A runbook is the plan, not a record that anyone executed it.** It is undated and byte-identical whether the drill happened or not, so it cannot corroborate the `date: 2026-07-26`.
2. **`runbook-0003` is `docs/runbooks/0003-postgresql-16-to-18-major-upgrade.md`** — a PostgreSQL major-version upgrade procedure. Not a restore, not a drill.

There is no corresponding entry in `docs/bcp/dr-test-log.md` either: the newest entry there is **2026-06-30**, a table-top with `RTO achieved: N/A` and `RPO achieved: N/A`.

This attestation raises a money-path readiness dimension from 2 to 3 (`collect-prod-readiness.mjs`), so the score currently rests on it.

## Why it matters

ADR-0242 exists because "runbooks exist" is not "recovery is measured", and DORA Art. 24–26 requires testing that is *evidenced*. ADR-0253 names the general shape: a control that cannot show it carried work. This is that shape applied to the evidence layer itself.

## What is needed

Either of these closes it — the attestant is the only person who knows which is true:

- **The drill happened**: add the `## 2026-07-26` entry to `docs/bcp/dr-test-log.md` with measured RTO/RPO (or `N/A` with `Type: table-top`), and repoint the attestation's `ref` at that log.
- **It did not happen as a drill**: drop the attestation. The readiness dimension falls back to its derived score, which is the honest number.

## Guard

`check-readiness-attestations.py` R7 (this PR) now rejects a runbook ref for `restore_drill`/`dr_drill`, and rejects a drill-log ref whose log has no entry for the attestation's date. This entry is the single baselined exemption in `EXERCISE_REF_DEBT`, checked both ways — it must be removed from the script when the data is fixed.

JiRaska:
Closes #5673. Refs #4755, #4757.

## What the citation actually pointed at

`ledger.restore_drill` cited `runbook-0003` = `docs/runbooks/0003-postgresql-16-to-18-major-upgrade.md`, a **PostgreSQL 16→18 major-upgrade procedure**, still `Status: Draft`. It raised a money-path readiness dimension (C5) from 2 to 3.

The issue's premise was that a runbook is a plan, so nothing corroborates the `date: 2026-07-26`. That is right as a rule, and **wrong about this file** — which only reading all 240 lines shows. Its last section, `## Restore Drill Result`, is a genuine execution record:

| Field | Value |
|---|---|
| targetTime | `2026-07-26T02:05:08Z` (`stoppedAt` of `ledger-db-daily-20260726020500`) |
| RTO | **~84 s** (Cluster created 06:50:04Z → primary Ready 06:51:28Z) |
| Integrity | 96 `journal_entries`, matching the live `ledger-db` primary exactly |
| Attempt 1 | FAIL — drill `ServiceAccount` had no EKS Pod Identity → barman `Unable to locate credentials` |
| Attempt 2 | FAIL — `barmanObjectStore.serverName` defaults to the *externalCluster entry* name → `no target backup found` on an empty prefix |
| Result | PASS |

Failed attempts with root causes are not something a plan document contains. **The drill happened.** The defect was that its record was filed inside a plan, where no reader and no gate could date it, and the citation could not tell the two apart.

## Outcome chosen

Option 1 from the issue — **repoint, because a real artefact exists.** Not "adjacent": it is the execution record for that exact date, moved to `docs/bcp/dr-test-log.md` where the format expects it, and the attestation now cites that path.

The log entry states three things the old citation could not:
- the ~84 s is the **database's** restore time; **no service was cut over**, so application RTO is unmeasured;
- **RPO was not measured at all** — PITR ran to a chosen target, not to a failure. Recorded as unmeasured rather than as the WAL interval, which would be an assumption;
- scope is `ledger` alone. The other ~19 CNPG databases, 9 of them money-path, have no such drill. It is `live-restore`, **not** a failover, so it does not substitute for a `dr_drill`.

## Effect on the published figure

**ledger C5 stays 3; no service's score changes.** That is not a number being protected — the evidence was always real, only its address was wrong. The alternative was measured rather than assumed: deleting the attestation drops C5 to 2 and the evidence to `backup configured`, and I ran the collector that way to confirm it degrades honestly rather than scoring absent-as-fine. Deleting it would have erased a drill that was actually performed.

## Guard: debt cleared, verified by effect

`EXERCISE_REF_DEBT` is now **empty** — the single baselined exemption is gone, so an `restore_drill`/`dr_drill` runbook ref fails for everyone, ledger included.

| Probe | Result |
|---|---|
| gate on the new state | **PASS** |
| restore the old `ref: runbook-0003` | **FAIL** — "a runbook is the PLAN, not evidence anyone executed it" |
| mangle the `## 2026-07-26` heading in the log | **FAIL** — "the log it names never got the record" |
| `--self-test` | 26/26, both directions |

Two self-test cases had to change with it, and the second matters:

- the exemption case became the rule (runbook ref → expect `error`);
- the "**a drill log that never got the entry is not evidence**" negative used date `2026-07-26` — a date **this PR adds to the log**. Left alone it would have gone on passing while testing nothing. Moved to `2026-07-04`.

## Second commit

C5 printed `backup configured + drill` — the fact without the artefact, which is exactly how a money-path 3 came to rest on a citation nobody had opened. It now reads `backup configured + drill (docs/bcp/dr-test-log.md)`. Score arithmetic untouched.

## Verification

- `readiness-attestation-format` gate: **PASS** (and both negatives above go red).
- `npm test -- --run prod-readiness`: 2 files, 8 tests passed.
- `run-gates.py --all`: the remaining failures are environmental and unrelated — `PR_DIFF_BASE` unset (`db-migration`, `schema-compat`, `threat-model` all **PASS** once it is set), `api-contract` wants `sudo` to install oasdiff, `release-scope-mismatch` hits `PR_TITLE: unbound variable`, `loki-rule-load-test` needs a binary. This diff touches no `openapi.yaml`, no migration, no event schema and no threat model.


JiRaska:

**Checked and found nothing wrong.** This is the strongest PR in the batch I reviewed.

Ran both halves in place, exit codes read from `$?` on a redirect rather than through a pipe:

```
--self-test → exit 0   "self-test: all 26 cases passed (both directions)"
run         → exit 0   "checked 3 attestation(s) … 2 freshness warning(s)"
```

What I checked specifically:

- **The negative control is preserved, not weakened.** `check-readiness-attestations.py:401-405` converts the formerly-baselined `ledger.restore_drill` runbook ref from an *exemption* case into a **must-fail R7 case**, and `:407-418` adds a must-pass case for the new log-backed shape. Both directions, which is what makes emptying `EXERCISE_REF_DEBT` safe rather than a quiet loosening.
- **It caught a vacuous-test trap that the obvious version of this change would have created.** `:426-432` moves the "a drill log that never got the entry is not evidence" case from `2026-07-26` to `2026-07-04`, precisely because this PR adds a `## 2026-07-26` heading to the fixture log at `:563-564`. Without that, the gate's only proof that a *missing* log entry is caught would have silently become a pass-by-construction. The comment says so explicitly. That is the exact failure mode this repo has been bitten by before (an assertion that stops being able to fail), and it was pre-empted rather than discovered later.
- **The correction is honest about its own premise.** The rewritten header at `:83-96` records that the original finding was wrong in one direction — the drill *had* happened — while keeping the rule, and states the rule's real justification: a citation that resolves to a plan cannot tell a reader whether the plan was executed. Cite the record, not the procedure.
- **The new log entry does not over-claim**, which is the part worth calling out. `docs/bcp/dr-test-log.md` records the RTO as the **database's** (~84 s, no service cut over) and the RPO as **not measured** rather than substituting the WAL-archive interval, and it states that ~19 other CNPG databases have not had this drill. The `live-restore` type is introduced distinct from `live-failover` so this cannot be read as a `dr_drill`. Recording an unmeasured value as unmeasured, instead of as an assumption, is the right call.
- **`collect-prod-readiness.mjs:441-448`** now prints the ref alongside the score. Naming the artefact on the page that scores it is what would have made the original defect visible without reading 240 lines.
- **Derived files / hand-editing**: no `docs/adr/` change, so that rule does not apply. `attestations.yaml` is a source file, not a generated one.
- **Release scope**: `docs(governance)` touches `.github/` and `docs/` — non-releasing type, and no `version.txt` in those paths either way. `fix(admin-ui)` touches `openbank-admin-ui/scripts/`, which is inside admin-ui's release path with a releasing type, so it will cut an admin-ui patch release. That looks intended.

## One coordination note, not a code finding

**#5849 edits the same two files** — `.github/scripts/check-readiness-attestations.py` and `openbank-libs/governance/attestations.yaml`. Both PRs add self-test cases to the same `_self_test` case list and both touch `attestations.yaml`. That is the shape where git reports a clean merge and silently keeps only one side's list entry. Whichever lands second: after merging `main`, re-run `--self-test` and confirm the case **count** went up by both PRs' additions rather than just one, and confirm `ledger.restore_drill`'s ref is still `docs/bcp/dr-test-log.md`. Do not read the absence of a conflict as evidence.

JiRaska:
## Review — I ran the gate and its self-test; the negative controls are real. **Ready.**

**Staleness:** 247 commits behind `main` but `MERGEABLE`, and `git log <merge-base>..origin/main` over all four touched paths (`check-readiness-attestations.py`, `attestations.yaml`, `dr-test-log.md`, `collect-prod-readiness.mjs`) is **empty** — nothing has moved under it. Not a duplicate; not contradicted.

**Verified by running it, not by reading it:**
- `python3 check-readiness-attestations.py --self-test` → `self-test: all 26 cases passed (both directions)`.
- The shipped `attestations.yaml` change is consistent with the shipped `dr-test-log.md`: the attestation now cites `docs/bcp/dr-test-log.md` and the log carries a `## 2026-07-26` heading, so the ordinary run over the real file is what goes red if that heading is ever removed.

**The part I want to call out, because it is the failure this batch is full of.** The old self-test case `"R7: a drill log that never got the entry is not evidence"` used the date `2026-07-26` — the very date this PR then adds to the fixture log. Left alone, that case would have kept passing while testing nothing: the gate's only proof that a *missing* log entry is caught would have become vacuous, silently, and green. The PR moves it to `2026-07-04` and says so in a comment. That is the "fixing a defect makes the test that discriminated by it vacuous" trap, caught in the same change that created it.

The paired case is also correct: the formerly-baselined `ref: runbook-0003` shape flips from `"BASELINED DEBT ... exempt"` / `clean` to `"R7: ... now fails too"` / `error`, so the rule is proven by what it now rejects, and a new `clean` case covers the live entry's shape. `EXERCISE_REF_DEBT` going to an explicitly-typed empty dict keeps the shrink-only ratchet honest in both directions.

**The record itself is unusually well-scoped, and that is what makes it evidence.** It says what was *not* established as clearly as what was: RTO ~84 s is the **database's** (Cluster created → primary Ready) with no service cut over, so time-to-serve-traffic is unmeasured; **RPO is recorded as not measured**, on the grounds that PITR to a chosen `targetTime` establishes no data-loss window and quoting the WAL-archive interval would be an assumption. It also states the scope limit — one of ~20 CNPG databases, 9 of them money-path — and that `dr-restore-verify.yml` still cannot execute this in CI (#4757). Two prior failed attempts and their root causes are kept rather than smoothed away.

**`collect-prod-readiness.mjs`:** printing the `ref` alongside `+ drill` is the right minimal fix — the C5 score of 3 now names the artefact a reader must open, which is what the original defect hinged on nobody opening.

**One small note, no action needed:** the S3 bucket name in the drill record already appears on `main` in `db-backups.tf`, `automated-dr-restore.md` and ADR-0175, so this adds no new disclosure to a public repo.

**Not checked:** whether the 2026-07-26 drill happened as described — that is unverifiable from the repo by construction, and the PR is honest that it is an attested record.

Verdict: **ready.** Not approving — review comment only, per my constraints.

JiRaska:
Heads-up: I merged #5896 (`ab197343`), which rewrote the region of
`openbank-admin-ui/scripts/collect-prod-readiness.mjs` this PR also edits, so this branch now needs
a rebase.

The overlap was measured before merging, not discovered after:

- this PR: `@@ -441,7 +441,11 @@ function scoreC5Backup(short)`
- #5896:   `@@ -429,12 +429,28 @@ function scoreC4Data(short)` — which extends past 441

I merged the larger one first on purpose: rebasing a 4-line change onto a reworked collector is
cheaper and safer than the reverse, where resolving the conflict by taking one side would silently
drop most of #5896's rework. Same reasoning I left on the #5763 / #5884 `api-fuzz.yml` pair.

Nothing here is wrong — the two changes are independent in intent (`restore_drill` attestation vs
the C4/gate scoring). After the rebase, please re-read the merged `scoreC5Backup`/`scoreC4Data`
region rather than resolving hunk by hunk: `main`'s version is now the newer one.

JiRaska:
Rebased onto current `main`, and the resolution removed this PR's overlap with #5849 as a side effect.

`check-readiness-attestations.py` was the only conflicted file, and main's version is the evolved one
— it already carries the `DRILL_LOGS` tuple, the `docs/bcp/dr-test-log.md` reference and the
"#5673 emptied the list" note that this branch was written to introduce. Taking it wholesale rather
than merging two statements of the same idea drops the gate script from this PR's diff entirely.

What remains is the part main genuinely does not have:

```
docs/bcp/dr-test-log.md                                  +14 lines
docs/runbooks/0003-postgresql-16-to-18-major-upgrade.md
openbank-admin-ui/scripts/collect-prod-readiness.mjs
```

— the failed attempts and their root causes (`Unable to locate credentials`, exit 4), the caveat that
CNPG recovery manifests are not in-repo so this stays a manual exercise, the runbook cross-reference,
and the collector change that prints the citation **ref** rather than just the fact. That last one is
the interesting half: `+ drill` alone is what let a money-path score rest on a citation nobody had
opened, pointing at a PostgreSQL major-upgrade runbook.

Verified on the merged result: gate exit 0 — two freshness warnings, both advisory and unrelated
(`ledger.pentest` decays in 6 days, `consent.pentest` expired) — and the self-test passes 27/27 in
both directions.

**This PR no longer conflicts with #5849.** They collided only on the gate script, which is now out of
this diff, so the two can land in either order.

JiRaska:
Correcting the last line of my previous comment. I wrote that this PR "no longer conflicts with
#5849" and posted it in the same breath as the measurement that said otherwise — at that moment
`merge-tree` still reported one conflicting file between them. The claim ran ahead of the check.

What was true is narrower: the remaining conflict was **#5849 against `main`**, not between the two
PRs. #5849 had not been rebased yet, so pairing it with anything current surfaced its own staleness.

Now that both are rebased it is measured rather than asserted:

```
#5849 vs main                  0 conflicts
#5889 vs main                  0 conflicts
#5849 vs #5889 (both rebased)  0 conflicts
```

So they can land in either order — but that is a fact about the current heads, not something the
earlier comment had established.

Issue #5281, opened 2026-08-17
feat(flaky-test-hunter): deliver bounded non-money development PR capability


JiRaska:
Implements ADR-0031 D9 phase 3 without advancing the published rollout status prematurely.

Scope:
- replace the placeholder GitHub adapter with a token-backed, fail-closed implementation;
- restrict automated changes to one explicitly non-money-path service and a mechanically verifiable test-only change shape;
- preserve no-merge, no-approval, human-review and audit invariants;
- add contract and integration proof for an opened PR, including the degraded no-token path;
- provide deployment evidence for the GitHub credential binding and one human-merged PR before promoting the roadmap.

Out of scope:
- money-path ownership or phase-4 promotion; those require two approvals and a threat model.

Refs ADR-0031 D9 and #5261.

JiRaska:
Controlled proof remains outstanding. Read-only checks on 2026-08-20 confirmed the bounded non-money-path writer is deployed and its ExternalSecret reports Ready. Do not promote ADR-0031 D9 phase 3 until the evidence set contains: (1) a real agent-generated bounded PR URL, (2) its policy-decision and AI-attributed audit records, and (3) a protected-branch human merge by a different principal. Phase 4 remains blocked pending this phase-3 evidence plus the money-path two-person review and threat-model controls.

JiRaska:
## Closing: the bounded non-money development-PR capability is built, restricted and proven

Verified at `origin/main` = `161febab5ef1ce9531fc1acd0eb35a2d8ae4d2eb` (2026-09-01T09:05:48Z).

### 1. The placeholder GitHub adapter is gone — it is token-backed and fail-closed

```
$ git show origin/main:openbank-flaky-test-hunter/src/main/kotlin/com/openbank/flakytest/infrastructure/adapter/GitHubProposalAdapter.kt | grep -n -i 'placeholder\|TODO\|not implemented\|stub'
   (no output)
```

Known-positive for that probe: the same grep over the file's earlier placeholder form is what the
issue text describes; the empty result here is against a file that is 100+ lines of real
`java.net.http` calls (`getFile`, `getMainSha`, `createBranch`, `commitFile`, `openPr`).

Fail-closed on an absent token, and on anything it cannot mechanically verify:

```
if (config.githubToken().orElse("").isBlank()) {
    log.warn("flaky-test-hunter.github-token not seeded — no development PR opened (degraded)")
    return null
}
if (!isEligible(finding, fixDiff)) {
[...]

PR #5890, opened 2026-08-20
fix(flaky-test-hunter): enforce the bounded non-money GitHub write surface

JiRaska:
## What this changes

`flaky-test-hunter` is the only chartered agent with a **real GitHub write path** (ADR-0031 D9
phase 3). Most of what #5281 asks for already landed on `main`: the token-backed, fail-closed
`GitHubProposalAdapter`, the one mechanically-verifiable change shape, the AI-attributed audit on
every disposition, and the contract/degraded-path tests. This PR closes the part that had **no
enforcement behind it**.

### 1. The two safety bounds were held only by a human reading a constant

Phase 3 rests on two properties:

1. the write is confined to one **non-money-path** module's own test tree;
2. the agent opens a PR and stops — it never merges, approves, or requests review.

Neither was enforced by anything a change could trip over:

- The unit tests pin only that a **ledger** path is refused. Re-pointing the writer at, say,
  `openbank-settlement-service` would keep every test green.
- `money_path_services` is a list that **grows** — settlement, sanctions, sdd and interest were all
  added to it after the fact. A module that is safe today can become money-path tomorrow with no
  code change at all, and nothing would notice.
- Nothing anywhere would go red if someone added `PUT /pulls/{n}/merge` to the adapter.

New gate `agent-bounded-write-surface`
(`.github/scripts/check-agent-bounded-write-surface.py`, `mode: enforced`, group `registry`) binds
the writer's target to `rules.yaml: money_path_services` **itself** rather than to a second
hand-kept list — the failure shape `CLAUDE.md` calls out for gates whose scope is a maintained copy
of the thing they check.

### 2. The bound was declared twice

`LlmDiagnosisAdapter.proposeFixDiff` repeated the prefix as its own string literal instead of
calling `BoundedTestPath`. Two copies of one rule can be widened independently with nothing going
red — and the copy was the weaker one: a plain `startsWith` with no traversal, no leading-slash and
no percent-encoding checks. It now calls `BoundedTestPath.isSafe`, and the gate refuses any
re-introduced duplicate.

### 3. Dead transformer debris

`GitHubProposalAdapter.kt` carried two entirely unused transformer objects
(`ExplicitUnitReturnType` with three dead regex properties, one of them a mangled escaped literal,
plus `BoundedUnitReturnType`) behind a `@Suppress("UnusedPrivateProperty")`. Nothing referenced
them and nothing tested them; a future reader switching to one would have silently traded away the
string-based transformer's refusal properties. Removed.

### 4. The agent doc claimed a stub

`docs/agents/flaky-test-hunter.md` still described `GitHubProposalPort` as "a stub pending GitHub
App installation-token wiring" and `proposeFixDiff` as returning `null` unconditionally. Both were
false. Corrected, and it now names the gate that enforces the bounds.

## Guard falsification (verify by effect, both directions)

Against the **real** adapter file, not a fixture:

| Mutation | Gate |
|---|---|
| prefix re-pointed at `openbank-ledger-service` | **exit 1** — names the money-path violation |
| `PUT /pulls/1/merge` added to the adapter | **exit 1** — "the agent proposes; a human disposes" |
| `OWN_TEST_SOURCE_PREFIX` deleted | **exit 1** — absence does not read as compliance |
| duplicate prefix literal restored in `LlmDiagnosisAdapter` | **exit 1** |
| unmodified `origin/main` state | **exit 0** |

And the self-test is itself falsifiable — removing the money-path comparison from `evaluate()`
turns 2 self-test cases red; removing the endpoint scan turns 4 red. It also refuses to pass
vacuously: an empty `money_path_services`, a missing `rules.yaml`, or a missing adapter each fail
rather than report clean.

12 self-test cases + a live read of both the adapter and `rules.yaml`.

## Charter

**No charter change is required, and none was made.** `agents.yaml` grants
`tier: write_proposal, resources: [github-pr]` with an explicit `deny: tier: write, resources: ["*"]`
— exactly the capability this writer uses. Since `agents.yaml` is untouched, none of its three
derived tails (`docs/compliance/eu-ai-act.md`, the OPA bundle ConfigMaps + `policy-checksum`
annotations, `ai-governance-snapshot.json`) needed regenerating.

## Not in scope, deliberately

- **No promotion of the published rollout status.** The remaining phase-3 evidence — a real
  agent-generated PR URL, its policy-decision/AI-attributed audit records, and a protected-branch
  human merge by a different principal — is runtime evidence this PR cannot manufacture. `Refs`,
  not `Closes`.
- **Money-path is untouched**, and now structurally so rather than by convention.
- Sibling agents `docs-truth-agent` and `authz-policy-auditor` still carry the original bootstrap
  `GitHubProposalAdapter` stub that returns a **fabricated** `pending-...` URL rather than refusing
  — the opposite of the `UnwiredProposalPort` pattern the fleet settled on (#3613/#3826). Noted in
  the agent doc; worth its own issue, not widened into here.

Refs ADR-0031 D9 and #5281.

### DIFF
--- a/.github/gates/gates.yaml
+++ b/.github/gates/gates.yaml
@@ -1555,6 +1555,33 @@ gates:
     run: |
       bash .github/scripts/check-agent-charter-registry.sh
 
+  # ADR-0031 D9 phase 3 (issue #5281): flaky-test-hunter is the only chartered agent with a real
+  # GitHub write path. The two properties that make that acceptable — the write is confined to one
+  # NON-money-path module's own test tree, and the agent opens a PR but never merges or approves —
+  # held only because a human read a Kotlin constant. Every existing unit test pins that a LEDGER
+  # path is refused, so re-pointing the writer at another service would keep them all green, and
+  # money_path_services is a list that GROWS (settlement, sanctions, sdd and interest were all
+  # added after the fact). This binds the constant to rules.yaml: money_path_services itself
+  # rather than to a second hand-kept copy, and rejects any merge/approve/review-request endpoint
+  # literal in the adapter.
+  #
+  # ENFORCED. Pure python + one file read, <1 s. Falsified by effect against the real adapter on
+  # 2026-08-20: a money-path prefix, an added `PUT /pulls/{n}/merge`, and a deleted bound each make
+  # it exit 1; removing either half of `evaluate` makes the self-test go red.
+  - id: agent-bounded-write-surface
+    min_subjects: 20   # the money_path_services entries the bound is checked against; 23 today
+                        # (2026-08-20). A gate that resolves an empty list would pass everything,
+                        # so the floor moves with the list rather than tracking it exactly.
+    name: "agent bounded GitHub write surface (ADR-0031 D9 phase 3, enforced)"
+    group: registry
+    mode: enforced
+    rationale: "flaky-test-hunter is the only agent with a real GitHub write path; its non-money-path confinement and its inability to merge or approve held only because a human read a Kotlin constant, and money_path_services grows over time."
+    review_after: "2027-02-20"
+    selftest: python3 .github/scripts/check-agent-bounded-write-surface.py --self-test
+    selftest_expect: pass
+    run: |
+      python3 .github/scripts/check-agent-bounded-write-surface.py
+
   - id: agent-model-parity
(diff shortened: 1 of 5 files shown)


JiRaska:

|---|---|
| #5890 | 3934 |
| #5849 | 3853 |
| #5875 | 3735 |
| #5967 | 3417 |

PR #5893, opened 2026-08-20
feat(kyc): adverse-media port with an explicit not-configured state

### LINKED ISSUE #4459: kyc: adverse-media source for perpetual KYC triggers (ADR-0256 D5 followup)

JiRaska:
## What this is

ADR-0256 **D5** decided that adverse media is **out of scope**: `CheckType.ADVERSE_MEDIA` exists in
the KYC domain model, no adverse-media source exists on the platform, and the ADR explicitly does
not create one — *"a stubbed feed would make a catalogue member decorative"*. Issue #4459 asks for
the thing the ADR deferred: select and land a **real** source (vendor feed or in-repo ingestion),
change-detectable, EU-residency-compatible (ADR-0175), and licensed.

No such source is obtainable from this repo. Every candidate needs either a commercial contract
(World-Check / ComplyAdvantage / Refinitiv) or a licensing + residency decision on a public dataset
(GDELT-class). That is a procurement and legal question, not a coding one. So this PR does **not**
close #4459 and does not add the `ADVERSE_MEDIA_CHANGED` trigger to the ADR-0256 catalogue.

What it lands instead is the half that is honest without a source: **the port, the fail-closed
outcome model, and an observable statement that the platform has no coverage.**

## Why not a stub

A stub adapter returning "no adverse media found" would be the exact defect this repo has already
paid for twice:

- `PushResult.skipped()` carried `success = true`, so every push in an environment with no APNs
  credentials was counted as delivered and committed `SENT` (ADR-0252 phase 0, #4348).
- `LoggingDeadLetterSink.quarantine()` was one `log.warnf`, so "quarantined" meant "logged".

An adverse-media screen that cannot reach a source must not be representable as a clean result.

## What is in the diff

`openbank-kyc-service`, four files (3 main + 3 test):

- **`AdverseMediaScreeningPort`** — a **four-valued** `AdverseMediaOutcome`
  (`HIT` / `NO_HIT` / `SOURCE_UNAVAILABLE` / `SOURCE_NOT_CONFIGURED`). "Could not screen" is never
  the same value as "screened and found nothing"; there is no shared boolean. `toCheckStatus()` is
  an exhaustive `when` in which **`NO_HIT` is the only member that maps to `PASSED`** — `HIT` maps
  to `MANUAL_REVIEW` too, because ADR-0116's four-eyes gate owns the disposition and a trigger may
  never decide anything about a customer (ADR-0256 D2). `AdverseMediaScreeningResult` carries an
  `init` invariant: `sourceId` is null **iff** the outcome is `SOURCE_NOT_CONFIGURED`, so a clean
  result cannot be constructed without naming the source that produced it.
- **`UnconfiguredAdverseMediaSource`** — the only implementation today. It reports the *absence of a
  source*, and it can return nothing else.
- **`AdverseMediaSourceReadiness`** — `@Startup @ApplicationScoped`, publishes
  `openbank_kyc_adverse_media_source_configured` (today: **0**) and logs a boot-time warning naming
  the gap. `@Startup` is load-bearing: `@ApplicationScoped` is lazy, so a warning or a gauge
  registration on a bean nothing calls never reaches a pod log — this repo shipped a PAdES adapter
  warning its seals were "worthless as evidence" and that line never once appeared (#1299). The
  gauge is derived from `AdverseMediaScreeningPort.sourceId`, not from a config key, so it cannot
  claim coverage the wired adapter does not have.

**Deliberately not wired into `KycService.createCase`.** No `ADVERSE_MEDIA` check is created on any
case today, and adding one now would put a permanently-`MANUAL_REVIEW` check on every KYC case —
operator load for zero coverage, which is the decorative outcome D5 exists to prevent. Wiring lands
with the source.

## Evidence

Verified by effect in both directions, then falsified by breaking the implementation:

| Break | Test that went red |
|---|---|
| `UnconfiguredAdverseMediaSource.screen` returns `NO_HIT` (the stub this PR argues against) | `UnconfiguredAdverseMediaSourceTest` (2 tests) |
| `toCheckStatus()` maps `SOURCE_NOT_CONFIGURED -> PASSED` | `AdverseMediaScreeningPortTest` (2 tests) |
| gauge registration removed from `@PostConstruct` | `AdverseMediaSourceReadinessTest` (2 tests) |

Each break was applied, the suite re-run to confirm the corresponding test fails, and the break
reverted. The positive direction is covered too: a configured port produces gauge `1`, and a `HIT`
result records its source.

Local gate on `openbank-kyc-service`: `ktlintCheck` PASS, `detekt` PASS, `test` **98 tests, 0
failures, 1 skipped** (the pre-existing broker-gated `KycEventPactBrokerProviderVerificationTest`,
unchanged), `quarkusBuild` BUILD SUCCESSFUL — the last one is what proves the `@Startup` bean's
injection point resolves, since CDI failures surface only there.

## Scope notes

- `openbank-kyc-service` is **not** in `rules.yaml: money_path_services`, so the two-approval and
  threat-model gates do not apply.
- No API change (no endpoint added), no DB change, no event-contract change, no `agents.yaml` or
  `rules.yaml` change, so none of the derived-artifact regenerations are triggered.
- Unrelated observation, not fixed here: `openbank-kyc-service/src/main/resources/openapi.yaml`
  advertises a check-type enum (`IDENTITY, SANCTIONS, PEP, ADVERSE_MEDIA, SOURCE_OF_FUNDS`) that
  does not match the domain `CheckType` (`IDENTITY, ADDRESS, PEP_SCREENING, SANCTIONS_SCREENING,
  ADVERSE_MEDIA`). Worth its own issue.

Refs #4459
Refs ADR-0256 D5

### DIFF
--- a/openbank-kyc-service/src/main/kotlin/com/openbank/kyc/application/port/out/AdverseMediaScreeningPort.kt
+++ b/openbank-kyc-service/src/main/kotlin/com/openbank/kyc/application/port/out/AdverseMediaScreeningPort.kt
@@ -1,0 +1,97 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) OpenBank contributors. Licensed under the Apache License, Version 2.0.
+// See LICENSE in the repository root or https://www.apache.org/licenses/LICENSE-2.0 for details.
+
+package com.openbank.kyc.application.port.out
+
+import com.openbank.kyc.domain.model.CheckStatus
+
+/**
+ * Outcome of an adverse-media screen. Deliberately **four-valued**: "we could not screen" is never
+ * the same value as "we screened and found nothing".
+ *
+ * This repo has paid for the opposite shape twice — a disabled push adapter whose
+ * `skipped()` result carried `success = true` counted every undelivered push as delivered
+ * (ADR-0252 phase 0, #4348), and a dead-letter sink whose `quarantine()` was a single log line
+ * made "quarantined" mean "logged". An adverse-media check that cannot reach a source must not be
+ * representable as a clean result, so [NO_HIT] is a distinct enum member that only a reachable,
+ * configured source can produce.
+ */
+enum class AdverseMediaOutcome {
+    /** The source was reached and returned at least one candidate article/record for the subject. */
+    HIT,
+
+    /**
+     * The source was reached and returned nothing. **Only** a configured, reachable source may
+     * yield this — it is the sole outcome that maps to [CheckStatus.PASSED].
+     */
+    NO_HIT,
+
+    /** A source is configured but could not be reached (transport fault, timeout, open circuit). */
+    SOURCE_UNAVAILABLE,
+
+    /**
+     * No adverse-media source is configured on this platform at all. This is the state OpenBank is
+     * in today: ADR-0256 D5 defers adverse media because no change-detectable, EU-residency-compatible,
+     * licensed source has been selected (issue #4459). It is a *permanent* unresolved state, not a
+     * transient one, and is reported as such rather than as an absence of findings.
+     */
+    SOURCE_NOT_CONFIGURED,
+    ;
+
+    /**
+     * Fail-closed mapping onto the KYC check lifecycle. Exhaustive by construction: a new outcome
+     * member will not compile until it declares which side of the line it falls on, and every side
+     * except [NO_HIT] is [CheckStatus.MANUAL_REVIEW] — an unresolved check a human must dispose of,
+     * never a [CheckStatus.PASSED] the platform did not earn.
+     *
+     * Note the deliberate asymmetry with a hit: [HIT] is also MANUAL_REVIEW, not FAILED, because
+     * ADR-0116's four-eyes gate owns the disposition — a trigger may never decide anything about a
+     * customer (ADR-0256 D2).
+     */
+    fun toCheckStatus(): CheckStatus = when (this) {
+        NO_HIT -> CheckStatus.PASSED
+        HIT, SOURCE_UNAVAILABLE, SOURCE_NOT_CONFIGURED -> CheckStatus.MANUAL_REVIEW
+    }
+
+    /** True when this outcome represents an actual observation of the source's contents. */
+    val isResolved: Boolean get() = this == HIT || this == NO_HIT
+}
+
+/**
+ * Result of one adverse-media screen. [sourceId] is null exactly when [outcome] is
+ * [AdverseMediaOutcome.SOURCE_NOT_CONFIGURED] — the provenance of a screening result is part of
+ * the result, so a case's check can record *which* source (and therefore which coverage) backed it.
+ */
+data class AdverseMediaScreeningResult(
+    val outcome: AdverseMediaOutcome,
+    val sourceId: String?,
+    val matchedHeadline: String? = null,
+) {
+    init {
+        require((sourceId == null) == (outcome == AdverseMediaOutcome.SOURCE_NOT_CONFIGURED)) {
+            "sourceId must be null iff outcome is SOURCE_NOT_CONFIGURED (was outcome=$outcome, sourceId=$sourceId)"
+        }
+    }
+}
+
+/**
+ * Outbound port for adverse-media screening of a party name.
+ *
+ * **There is no production implementation backed by a real source today.** ADR-0256 D5 keeps
+ * adverse media out of the perpetual-KYC trigger catalogue precisely because a stub would make
+ * `CheckType.ADVERSE_MEDIA` decorative; this port exists so that when a source is selected
+ * (#4459) it lands behind a contract that cannot express "clean" without having read something,
+ * and so that the platform's *lack* of coverage is observable today rather than implicit.
+ */
+interface AdverseMediaScreeningPort {
+
+    /**
+     * Stable identifier of the backing source, or **null when no source is configured**. The
+     * readiness gauge is derived from this rather than from a separate config key, so the reported
+     * coverage cannot drift from the adapter actually wired in.
+     */
+    val sourceId: String?
+
+    suspend fun screen(name: String, idempotencyKey: String): AdverseMediaScreeningResult
+}
--- a/openbank-kyc-service/src/main/kotlin/com/openbank/kyc/infrastructure/client/UnconfiguredAdverseMediaSource.kt
+++ b/openbank-kyc-service/src/main/kotlin/com/openbank/kyc/infrastructure/client/UnconfiguredAdverseMediaSource.kt
@@ -1,0 +1,34 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) OpenBank contributors. Licensed under the Apache License, Version 2.0.
+// See LICENSE in the repository root or https://www.apache.org/licenses/LICENSE-2.0 for details.
+
+package com.openbank.kyc.infrastructure.client
+
+import com.openbank.kyc.application.port.out.AdverseMediaOutcome
+import com.openbank.kyc.application.port.out.AdverseMediaScreeningPort
+import com.openbank.kyc.application.port.out.AdverseMediaScreeningResult
+import jakarta.enterprise.context.ApplicationScoped
+
+/**
+ * The only [AdverseMediaScreeningPort] implementation on the platform today: it reports that no
+ * adverse-media source is configured, and it can return nothing else.
+ *
+ * This is **not** a stub that answers "no adverse media found". That distinction is the entire
+ * point of the class: OpenBank has no licensed, EU-residency-compatible, change-detectable
+ * adverse-media feed (ADR-0256 D5, issue #4459), and a screening path that quietly returns a clean
+ * result for a source it never reached is worse than having no path at all — it manufactures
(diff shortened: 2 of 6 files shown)


JiRaska:

3-dot diff against `origin/main`; branch is 243 commits behind and additive-only (369 lines, 0 deletions), so the staleness risk is low.

The four-valued outcome is right, and the `require((sourceId == null) == (outcome == SOURCE_NOT_CONFIGURED))` invariant makes "clean without having read anything" **unconstructible** rather than merely discouraged. `@Startup` on the readiness bean is correctly load-bearing and the KDoc cites the right precedent. Three findings, one of them a hard blocker for a public repo.

### 1. BLOCKING — a named real person is used as the adverse-media test subject, in a public repo

`openbank-kyc-service/src/test/kotlin/com/openbank/kyc/infrastructure/client/UnconfiguredAdverseMediaSourceTest.kt:27` and `:38` pass the name of a real, living, publicly identifiable Czech politician into `source.screen(...)`, and `:38`'s comment splits the list into "adversely-reported or not". This repository is public. A committed test that names a real individual as the adverse-media example is real personal data in a tracked file, and the framing is worse than the data. Replace with a synthetic name (`"Jan Testovací"`, `"ACME Holding a.s."`) — nothing in either test depends on the value.

### 2. The port has no production consumer, and the live gap is next door

`AdverseMediaScreeningPort` is referenced by exactly three main-source files, all introduced here: the interface, the unconfigured adapter, and the readiness gauge. Nothing in the KYC case flow calls `screen()`. So `AdverseMediaOutcome.toCheckStatus()` — the fail-closed mapping, and the thing all seven of `AdverseMediaScreeningPortTest`'s assertions are about — is on a path no request reaches. That is fine as a deliberate contract-first landing and the KDoc says so; the shipped deliverable is really the gauge.

But it means the fail-closed property is **not** actually enforced on the platform today, because the live path is elsewhere: on `main`, `CheckType.ADVERSE_MEDIA` (`domain/model/KycCase.kt:31`) is settable through `PUT /api/v1/kyc/cases/{id}/checks/{checkType}` with an arbitrary status, and `src/main/resources/diagrams/01-kyc-verification-flow.mmd:23` documents exactly that call with `status=PASSED`. An operator can mark ADVERSE_MEDIA `PASSED` today with no source, no evidence and no `sourceId`, and this PR does not touch that. Please either say so in the PR body, or make the endpoint reject a `PASSED` on `ADVERSE_MEDIA` while `sourceId == null` — that is the change that would make the gauge's `0` mean something enforceable rather than merely reported.

### 3. `AdverseMediaSourceReadiness` — one weak reference and one absent-vs-zero note

- `AdverseMediaSourceReadiness.kt:57-62` uses `registry.gauge(name, this) { it.configured().toDouble() }`. Micrometer's object-and-function `gauge` overload holds the state object **weakly**. Here `this` is an `@ApplicationScoped` CDI bean and the container keeps a strong reference, so the gauge will not silently go to NaN — but that is a property of the bean's scope, not of the call, and it will stop being true if this ever becomes `@Dependent`. A one-line comment stating why the weak ref is safe would keep the next edit honest. (`Gauge.builder(...).strongReference(true)` is the alternative.)
- The KDoc says "an absent series reads to every dashboard exactly like a healthy one that has not scraped yet" — agreed, and that is the argument for the `@Startup`. It is also the argument for an alert on `openbank_kyc_adverse_media_source_configured == 0` **or absent**, which does not exist yet. As shipped, the gauge is a number nobody reads; this repo's recurring finding is that an unread source is indistinguishable from an empty one. Worth a follow-up issue reference in the PR body.

### Checked and clean
- The not-configured state is a distinct enum member, not a flag shared with success. `NO_HIT` is the sole outcome mapping to `PASSED`, and `AdverseMediaScreeningPortTest:46-49` asserts that as a set over `entries`, so a future member cannot quietly join the passing side. Both directions of the `sourceId` invariant are tested (`:53-66`).
- No metric is named for something it cannot establish: `..._source_configured` reports configuration, which is exactly what the port can observe. Contrast the `accepted` vs `delivered` lesson — this one is named correctly.
- `HIT` mapping to `MANUAL_REVIEW` rather than `FAILED` is right and the ADR-0116 reasoning is stated in place.
- `configured()` is derived from `port.sourceId`, not from a config key, so the gauge cannot claim coverage the wired adapter lacks.
- `AdverseMediaSourceReadinessTest` includes a `registry == null` case, so the `isResolvable` branch is exercised in both directions.
- The `runBlocking` tests are written `: Unit = runBlocking { }`, so JUnit5 will not silently drop them.

JiRaska:
## Blocking: a real person's name as an adverse-media test subject

`UnconfiguredAdverseMediaSourceTest.kt` screened **"Andrej Babiš"** — a real, named, living Czech politician — as its test subject, at both call sites. This is a **public** repository, and adverse-media screening is negative-news search, so this is worse than ordinary personal data in a tracked file. The repo's own rule: real personal data never goes into a tracked repo file.

**Fixed on this branch** (`ea2240519`): subject replaced with the obviously fictional `"Adverse Subject Zero"` (`"Jan Novák"`, the Czech John Doe placeholder, kept), plus a comment recording *why* the subjects here are fictional by construction so the next author does not reintroduce one. The assertions are unchanged — the adapter must still report `SOURCE_NOT_CONFIGURED`, never `NO_HIT`. `:openbank-kyc-service:test` (3/3, 0 skipped), `ktlintCheck`, `detekt` all green.

## The sweep — this is not confined to this PR

Two independent enumerations over `git ls-files`:

**Method A** — regex for quoted two-word-capitalised name literals in files whose path matches `kyc|pep|sanction|adverse|screen|watchlist|aml|customer|party|onboard`: **156 distinct name-shaped literals**. Most are placeholders (`Jane Doe`, `Alice Example`, `Test Party`) or false positives (dashboard panel titles like `Sanctions Hits`).

**Method B** — a curated roster of real, identifiable public figures matched case-insensitively across **all** tracked files: **181 occurrences in 20 files** (of which 13 occurrences across 7 files are false positives on the substring `Modi` inside `modified`/`Modified`). Net: **~168 real-person occurrences in 13 tracked files.**

Real names still present on `main`, outside this PR:

| File | Names |
|---|---|
| `openbank-sanctions-service/src/main/resources/db/migration/V6__seed_sanctions_entries.sql` | **112 occurrences.** Genuinely-sanctioned figures (Putin, Lukashenka, Maduro, Lavrov, Khamenei, Kim Jong-un, Medvedev, Patrushev, Vekselberg, Yanukovych, Assad) **and — the serious part — people who are not sanctioned at all**: Petr Pavel, Petr Fiala, Miloš Zeman, Andrej Babiš, Alena Schillerová, Viktor Orbán, Trump, Biden, Xi Jinping, Modi, Erdoğan, Netanyahu, Zelensky. Seeded as PEP/sanctions rows. |
| `openbank-sanctions-service/src/test/.../SanctionsServiceTest.kt` | Putin ×12, Babiš ×3 |
| `openbank-sanctions-service/src/test/.../SanctionsImportServiceTest.kt` | Babis ×7, Putin ×4 |
| `openbank-sanctions-service/src/test/.../SanctionsMapperTest.kt`, `SanctionsEntryRepositoryImplTest.kt`, `SanctionsPactProviderVerificationTest.kt` | Putin ×10 |
| `openbank-kyc-service/src/test/.../SanctionsScreeningAdapterTest.kt` | Babiš ×4, Miloš Zeman ×3 |
| `openbank-kyc-service/src/test/.../PepScreeningServiceTest.kt` | Babiš ×4 |
| `openbank-kyc-service/src/test/.../KycServiceTest.kt` | Babiš ×2 |
| `openbank-fx-service/src/test/.../SanctionsScreenPactConsumerTest.kt` + `pacts/openbank-fx-service-openbank-sanctions-service.json` | Putin ×4 |

I deliberately did **not** fix those here — it is a separate change, and `V6__seed_sanctions_entries.sql` is a Flyway migration that has been applied, so editing the file triggers a `checksum mismatch` startup failure. It needs a superseding migration plus a decision about the whole seed roster, not an in-place edit. Recommend a dedicated issue.

I did not classify any name as fictional on the grounds that it looked unfamiliar; the roster above is names I can positively identify as real people.

PR #5894, opened 2026-08-20
fix(lending): add sourceService to loan.withdrawn and loan.accelerated

JiRaska:
## Summary

Adds `sourceService` (and a body-level `eventType`) to `loan.withdrawn` and
`loan.accelerated` — the two event types PR #5399 missed when it swept this
module for #5256.

#5399 patched nine per-event payload builders. These two are the only event types
in the module produced by a **shared, parameterised helper**
(`TerminationService.emitDomainEvent(loan, type)`), so enumerating the per-event
builders — the obvious way to work through this module — never reaches them, and
makes the module's event-type count read as 9 when it is 11. The helper sits
directly below `loanEvidence` in the same file, which #5399 *did* fix.

Both types land on `openbank.lending.events`, which audit-service consumes, so
`AuditConsumer` attributed both from its topic table rather than the producer's
own claim: no error, no metric, no log line — visible only by grouping
`audit_entries` on a live database.

lending-service is a money-path service (`rules.yaml: money_path_services`) —
this PR needs 2 approvals.

## Naming — two deliberate departures from the fleet default

**`eventType` stays dotted lowercase.** The issue names SCREAMING_SNAKE as the
convention, but this module's discriminators are `loan.withdrawn`-style, that is
what the outbox already publishes in the `ce-type` header, and that is what its
consumers match on. Forcing the convention here would break a working contract, so
the body value is the caller's `type` verbatim — the convention is the default,
not an override of a live contract.

**`sourceService` is `"lending"`, not `"lending-service"`.** Every other event type
in this module already self-reports `"lending"`, so this keeps the producer
self-consistent. Note this differs from what audit-service's `TopicAttribution`
derives for the topic (`lending-service`), which #5399 flagged and left. That
inconsistency is real but is **not** made worse by this PR — it is
pre-existing across nine event types, and having these two disagree with the other
nine would be strictly worse. Worth a follow-up decision on which spelling wins
fleet-wide; deliberately not resolved here, since converging would rename nine
already-shipped event types and split historical rows a second time.

## Verification — the wire payload, not the field

- **Producer side** — `TerminationServiceTest`: drives the real `withdraw` and
  `accelerate` paths, captures the emitted `LendingOutboxMessage`, and reads
  `sourceService`/`eventType` off the payload as **parsed JSON** (not a substring
  match, which could pass on an unrelated occurrence in the string).
- **Consumer side** — `AuditAttributionTest`: feeds that same wire shape through
  `AuditConsumer.consume` and asserts
  `sourceServiceSource == AttributionSource.EVENT`.

**Falsification:** removing `sourceService` from the helper turns **both** producer
tests red. Restored before commit.

## Test plan

- [x] `./gradlew :openbank-lending-service:ktlintCheck :openbank-lending-service:detekt` — green (`ktlintMainSourceSetCheck` and `ktlintTestSourceSetCheck` both confirmed to have executed)
- [x] `./gradlew :openbank-lending-service:test --tests "...TerminationServiceTest"` — 12 tests, 0 failures, **0 skipped**
- [x] `./gradlew :openbank-audit-service:ktlintCheck :openbank-audit-service:detekt :openbank-audit-service:test --tests "...AuditAttributionTest"` — 27 tests, 0 failures, **0 skipped**
- [x] Negative case: field removed ⇒ both producer tests fail

## Not changed here

`occurredAt` was already present on both payloads.

Refs #5256

### DIFF
--- a/openbank-audit-service/src/test/kotlin/com/openbank/audit/application/AuditAttributionTest.kt
+++ b/openbank-audit-service/src/test/kotlin/com/openbank/audit/application/AuditAttributionTest.kt
@@ -506,6 +506,26 @@ class AuditAttributionTest {
     }
 
     @Test
+    fun `lending-service's shared-helper event types are attributed to the producer too`(): Unit = runBlocking {
+        // #5399 swept this module's nine per-event payload builders and missed loan.withdrawn and
+        // loan.accelerated — the only two types built by a shared, parameterised helper
+        // (TerminationService.emitDomainEvent), so a reader following the per-event builders never
+        // reached them. This is that helper's wire shape (see TerminationServiceTest, which reads
+        // the key off the payload the production code actually emits).
+        val entry = capturingSave()
+
+        consumer.consume(
+            """{"eventType":"loan.withdrawn","loanId":"${UUID.randomUUID()}",""" +
+                """"partyId":"${UUID.randomUUID()}","occurredAt":"2026-08-09T12:00:00Z",""" +
+                """"sourceService":"lending"}""",
+            EventAddress(topic = "openbank.lending.events"),
+        )
+
+        assertThat(entry.captured.sourceService).isEqualTo("lending")
+        assertThat(entry.captured.sourceServiceSource).isEqualTo(AttributionSource.EVENT)
+    }
+
+    @Test
     fun `the producer's own claim wins over the topic, and is marked as the producer's`(): Unit = runBlocking {
         // Body-first ordering: this change can only turn a sentinel into a value. It must never
         // re-attribute a row that is already attributed, or customer-edge's 421 correct rows move.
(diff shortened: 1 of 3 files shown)


JiRaska:

**Premise confirmed still true today.** I checked the producer type itself rather than grepping for the quoted JSON key — a grep cannot see a serialised data class, so a no-hits result on this family is a false clean:

- `origin/main`'s `SepaPaymentActivitiesImpl.kt` contains **no** `sourceService` at all (5 hand-built payloads on `openbank.sepa.payment.events`).
- `origin/main`'s `TerminationService.kt` has `sourceService` on **one** builder only (the `credit.loan.transition` evidence payload, line 355); `emitDomainEvent` — which produces `loan.withdrawn` and `loan.accelerated` — still emits `{"loanId":…,"partyId":…,"occurredAt":…}` with no `sourceService`.

Neither gap has been closed by anything that merged in the last nine days, and #5888 and #5894 touch **disjoint** production files, so they are not duplicates of each other.

**One merge-order caveat.** Both PRs add a new test method to the same file, `openbank-audit-service/…/AuditAttributionTest.kt`, and that file has grown on `main` since these branches were cut (618 lines on both branches vs 646 on `main`). Two sides appending to one file is where git reports a clean merge and can still keep only one side. After merging the second of the pair, confirm **both** new test methods are present on `main` — the suite stays green either way, so a dropped test is silent.

Issue #5760, opened 2026-08-19
tax-reporting is a released component with no gitops workload at all (referral now has one) — decide whether it should be deployed or stop being released


JiRaska:
**Corrected after checking what the missing KafkaUser actually means. The original framing was wrong
and would have sent someone to create infrastructure for a service that is deliberately not deployed.**

## What is actually true

`openbank-tax-reporting-service` has **no presence in `openbank-infra/gitops/` at all** — not a missing
KafkaUser, but no workload, no manifests, nothing. `grep -rn "tax-reporting" openbank-infra/gitops/`
returns zero hits.

This is known and intentional, and the repo already documents it. From
`openbank-infra/aws/envs/sandbox-platform/ecr-service-repositories.tf`, explaining why the ECR
repository list is not derived from release-please's package list:

> It omits the six deployed images that are not released components (analytics-sink, developer-portal,
> document-renderer, keycloak, pyroscope-agent, ci-runner) and includes openbank-tax-reporting-service,
> a released component with no gitops workload, no auto-deploy entry and — correctly — no repository.

So it is a released component (`version.txt` 0.9.1, registered in the release manifest) that has never
been deployed. The absent KafkaUser is a *consequence* of that, not an independent gap.

## What this changes

**Nothing is broken in production**, because there is no production instance. #5751 was right to
baseline `withholding-remitted-in` rather than wire it, and #5759's consumer fix is still correct —
it makes the service behave properly whenever it is deployed, rather than shipping the swallow into
its first deployment.

The KDoc #5759 wrote for that consumer states the channel halts on a persistent failure and cannot yet
do otherwise. That is accurate for the code, but a reader should know the channel does not currently
run anywhere.

## What is actually worth deciding

Not "create a KafkaUser". Rather: **is this service meant to be deployed?** It is versioned and released
on every qualifying commit, so it accrues changelog entries and version bumps for a workload nobody runs.
Either:

- deploy it — in which case it needs the full gitops set (workload, KafkaUser with Read on its source
  topic and Write on `openbank.dlq.tax-reporting.withholding-remitted-in`, auto-deploy entry, ECR
  repository), and the DLQ baseline entry in `check-incoming-dlq-wiring.py` comes off; or
- decide it stays undeployed for now and record why, so the next person who finds a §38d statutory
  consumer with no runtime does not re-derive this from scratch — as both #5751's baseline comment and
  this issue did independently.

Refs #5745, #5751, #5759.

[...]

PR #5896, opened 2026-08-20
fix(admin-ui): resolve the repo-fixable prod-readiness NO-GO causes (contract tests, undeployed component)

JiRaska:
## What

The three remaining repo-fixable NO-GO causes from #5706, taken one at a time. One was already
fixed on `main`, one is genuinely repo-fixable and is fixed here, and one is repo-fixable only in
what the matrix *says* — the thing it reports is an owner decision, not a code change.

`Refs #5706` rather than `Closes`: cause 1 leaves a decision open (#5760) and tax-reporting's own
C2/C3/C5/C7/C8 cells stay where they are.

## Cause 3 — mcp-service C4 rollback note: already done, nothing to do here

`#5832` landed `openbank-mcp-service/src/main/resources/docs/04-data.md` and mcp now scores
`C4=2 [2 migrations, rollback_note=y]` with an overall **GO**, measured on the collector as of
`origin/main`. The issue's framing ("mcp-service ships no docs pack at all") is stale. The
locale-less filename is deliberate and supported — `ClasspathMarkdownLoader` documents
`docs/04-data.md` as the language-agnostic form served for any language — so this is not the
"landed but unreachable" shape it superficially resembles.

## Cause 2 — missing contract tests: repo-fixable for two of three

`campaign` and `engagement` both consume exactly one route,
`GET /api/v1/consents/party/{partyId}/grantee/{granteeId}/active?scope=…` (ADR-0198 D4), and read
exactly one field off it. Both now ship a consumer pact with **both** branches (an ACTIVE
per-channel consent; a party with no consent, which is a 200 with `granted: false`, not a 404),
and the provider half is wired in the same change — three states on consent-service's existing
always-running `@PactFolder` class.

The provider replay earned its keep on the first run: it answered **400 "Consent validity exceeds
maximum allowed 365 days"** against a seeded window of 366 days, a rule the consumer mock has no
way to know about. That is the asymmetry the repo's Pact discipline is built around, reproduced.

`granted` is pinned by value rather than `booleanType`; expected paths are literals retyped from
`ConsentResource`, with only the outgoing request reflected off the client's `@Path`.

**tax-reporting is deliberately not given one.** Nothing in the repo consumes its HTTP API and it
ships no outbound rest-client, so there is no consumer whose expectations a pact could encode.
Authoring one against an invented consumer would be a scoring artifact of exactly the kind the C3
probe was repaired to reject. Its C3 stays 1.

## Cause 1 — tax-reporting: the honest fix is representation, not deployment

Per #5760 the service has **no gitops workload at all**, and that is deliberate and already
documented in `ecr-service-repositories.tf`. So its C5 ("no CNPG cluster"), C7 (no NetworkPolicy)
and C8 ("not deployed") are consequences of the absent workload rather than controls anyone
skipped, and none of them can be closed by a repo change. Deploying is the owner's call.

What the matrix could not do is say so. Both collectors now emit a third verdict, **NOT-DEPLOYED**,
and C5's evidence names the missing workload instead of a missing backup.

**This is not a loosening.** NOT-DEPLOYED is reachable only where the old code said NO-GO: the GO
branch is evaluated first, and a service with no namespace scores C8 <= 1 anyway. C5 still scores
0. `test_not_deployed_never_replaces_a_GO` asserts exactly that — same failing scorecard, deployed
fixture, still NO-GO; all-clear scorecard, still GO — and the two pre-existing gate tests now have
to deploy their fixture, which is how a hidden branch change would have surfaced.

The undeployed state is also recorded in `docs/runbooks/svc-tax-reporting.md`, so the next person
to find a §38d statutory consumer with no runtime does not re-derive it (the third independent
derivation, after #5751's baseline comment and #5760).

## Evidence — before / after, same collector

| service | before | after |
|---|---|---|
| campaign | `C3=1 [openapi=y, NO contract test in src/test]` | `C3=2 [openapi=y, 1 pact test(s)]` (still NO-GO on C8: no domain metrics) |
| engagement | `C3=1`, gate NO-GO | `C3=2`, gate **GO** |
| tax-reporting | `C5=0 [no CNPG cluster]`, gate NO-GO | `C5=0 [no CNPG cluster — service has no gitops workload at all]`, gate **NOT-DEPLOYED** |
| mcp | `C4=2`, gate GO (already, via #5832) | unchanged |

Verification run locally, serially:

- `check-readiness-collectors-agree.py` — `SUBJECTS=44`, identical scores/evidence/gates; and
  `--self-test` still produces 101 divergences, so the gate can still fail.
- `prod_readiness_collector_test.py` — 32 tests, OK (28 before).
- consumer pacts: 4/4 green, both pact files generated.
- `ConsentPactProviderVerificationTest` — **6 interactions, 0 failures, 0 skipped** (read from the
  JUnit XML; pact-jvm's "Not all of the 2 were verified" console line is the known artifact).
- `check-pact-provider-replay.py` — OK, `KNOWN_UNCOVERED` still empty.
- `derive-pact-drift-scope.sh` — exit 0, both new modules in scope.
- `ktlintCheck` + `detekt` for campaign/engagement/consent — all six tasks executed, green.
- admin-ui `npm test` — 203 files / 1538 tests passed.

## Noted in passing, not fixed here

`score_c4_data` globs `<module>/**/db/migration/V*.sql`, which includes `build/`, so a locally
built module double-counts its migrations in the evidence string (campaign read "16" on a clean
tree and "32" after a build). Evidence text only — no score or gate depends on the count, and both
collectors do it identically so the agreement gate is unaffected.

Refs #5706, #5760

## Follow-up in this PR (from CI)

`gates (gitops)` failed the `service-runbook-drift` gate on the first push: I had hand-edited
`docs/runbooks/svc-tax-reporting.md`, which is generated (ADR-0029 rule #6). Fixed in `c95ad73`,
and the fix surfaced a defect underneath rather than just moving text.

`generate-service-runbooks.py`'s `service_namespace()` falls back to the service SHORT NAME when
no workload resolves, so tax-reporting's runbook instructed the on-call engineer to run
`kubectl -n tax-reporting` against a namespace that has never existed — the exact defect that
function's docstring records as fixed for a third of the fleet, surviving in the one case the fix
could not resolve, because there the right answer is not a different namespace but "no namespace
at all".

The banner is now derived from the same `gitops_facts.service_namespace` call the collector uses
for NOT-DEPLOYED, so the runbook and the matrix cannot disagree about whether a service runs. It
renders empty for every deployed service — regenerating all 44 runbooks changes exactly one file.
Two self-test cases added (17, was 15). The service-specific record moved to
`openbank-tax-reporting-service/CLAUDE.md`.

Measured both directions: the drift gate FAILS on a staged hand-edit and PASSES on the
regenerated tree.

### DIFF
--- a/docs/runbooks/svc-tax-reporting.md
+++ b/docs/runbooks/svc-tax-reporting.md
@@ -8,6 +8,19 @@ exercised DR drill, tracked as TTL'd attestations, never faked here. -->
 > Operational runbook for the `tax-reporting` service. Data domain **platform**,
 > classification **confidential**, datastore **PostgreSQL**.
 
+## Deployment status — NOT DEPLOYED
+
+**This service has no workload anywhere in `openbank-infra/gitops/`** — no Deployment,
+no Rollout, and therefore no namespace, no CNPG cluster, no NetworkPolicy and no
+PodMonitor coverage. It is a released component (it has a `version.txt`) that has never
+run, so **every `kubectl` command below names a namespace that does not exist** and every
+procedure here is a plan rather than a rehearsed one.
+
+The production-readiness matrix reports it as **NOT-DEPLOYED** rather than NO-GO for the
+same reason: the cells it fails are consequences of the absent workload, not controls
+someone skipped, and none of them can be closed by a repo change. Whether this service
+should be deployed is an owner decision — see the service's own `CLAUDE.md`.
+
 ## Service identity
 
 | Field | Value |
--- a/openbank-admin-ui/scripts/collect-prod-readiness.mjs
+++ b/openbank-admin-ui/scripts/collect-prod-readiness.mjs
@@ -429,11 +429,27 @@ function scoreC4Data(short) {
   }
 }
 
+/**
+ * True when the service has NO workload in gitops at all — not a Deployment, not a Rollout.
+ * Distinct from 'deployed but unscraped' and from 'stateless': a released component with no
+ * workload cannot own a CNPG cluster or a scraped namespace, so those cells report absences it
+ * could not have avoided. openbank-tax-reporting-service is the one such service today and it is
+ * deliberate (#5760). Naming the state is honesty, not leniency — see computeGate.
+ */
+function isUndeployed(short) {
+  return serviceNamespace(short) === null
+}
+
 function scoreC5Backup(short) {
   if (isStateless(declaredDatastore(short))) {
     // No datastore, nothing to back up. finrep scored 0 ('no CNPG cluster') for the absence of a
     // cluster it must not have — an unachievable 0 that read like a missing backup.
     return { score: 2, evidence: 'n/a — declares no datastore (stateless), nothing to back up' }
+  }
+  if (isUndeployed(short)) {
+    // Still 0 — only the evidence changes. 'no CNPG cluster' reads as a backup someone forgot to
+    // configure; the cluster is absent because the whole workload is (#5760).
+    return { score: 0, evidence: 'no CNPG cluster — service has no gitops workload at all' }
   }
   const clusters = gitopsFilesWithKind(short, 'Cluster')
   const cnpg = clusters.filter(f => f.includes('postgres') || readText(f).toLowerCase().includes('cnpg'))
@@ -555,7 +571,11 @@ function computeGate(short, scores) {
   const critical = new Set(['C1', 'C5', 'C7'])
   for (const { code } of DIMENSIONS) {
     const need = mp && critical.has(code) ? 3 : 2
-    if ((scores[code] ?? 0) < need) return 'NO-GO'
+    // NOT-DEPLOYED is only ever reached in place of NO-GO, never in place of GO: the GO branch
+    // below is the fall-through, and a service with no namespace scores C8 <= 1 anyway. So no
+    // score reachable in this state is treated more leniently than before — 'not production
+    // ready' and 'not in production' simply stop rendering as the same fact (#5760, #5706).
+    if ((scores[code] ?? 0) < need) return isUndeployed(short) ? 'NOT-DEPLOYED' : 'NO-GO'
   }
   return 'GO'
 }
@@ -637,5 +657,5 @@ const payload = {
 
 writeFileSync(OUT, JSON.stringify(payload, null, 2))
 console.log(
-  `[collect-prod-readiness] ${results.length} services scored, ${go} GO / ${results.length - go} NO-GO → ${OUT}`,
+  `[collect-prod-readiness] ${results.length} services scored, ${go} GO / ${results.filter(r => r.gate === 'NO-GO').length} NO-GO / ${results.filter(r => r.gate === 'NOT-DEPLOYED').length} NOT-DEPLOYED → ${OUT}`,
 )
--- a/openbank-admin-ui/src/app/api/prod-readiness/route.ts
+++ b/openbank-admin-ui/src/app/api/prod-readiness/route.ts
@@ -23,7 +23,7 @@ interface ReadinessService {
   money_path: boolean
   scores: Record<string, number>
   evidence: Record<string, string>
-  gate: 'GO' | 'NO-GO'
+  gate: 'GO' | 'NO-GO' | 'NOT-DEPLOYED'
 }
 
 interface ReadinessReport {
--- a/openbank-admin-ui/src/app/system/readiness/page.tsx
+++ b/openbank-admin-ui/src/app/system/readiness/page.tsx
@@ -6,7 +6,7 @@
 
 import { useState, useEffect } from 'react'
 import { useLanguage } from '@/lib/i18n/LanguageContext'
-import { ClipboardCheck, RefreshCw, Star, CheckCircle2, XCircle } from 'lucide-react'
+import { ClipboardCheck, RefreshCw, Star, CheckCircle2, XCircle, CircleSlash } from 'lucide-react'
 import { PageHeader, StatCard, StatusBadge, SWATCH_CLASS, type Tone } from '@/components/ui'
 
 interface ReadinessService {
@@ -14,7 +14,7 @@ interface ReadinessService {
   money_path: boolean
   scores: Record<string, number>
   evidence: Record<string, string>
-  gate: 'GO' | 'NO-GO'
+  gate: 'GO' | 'NO-GO' | 'NOT-DEPLOYED'
 }
 interface ReadinessReport {
   generated_for: string
@@ -54,7 +54,11 @@ export default function ReadinessPage() {
   const services = data?.services ?? []
   const dims = data?.dimensions ?? []
   const go = services.filter(s => s.gate === 'GO').length
-  const nogo = services.length - go
+  // Counted, not derived as `length - go`: NOT-DEPLOYED is a third verdict (#5760), and folding it
+  // into NO-GO is exactly the conflation the collectors stopped making — "not production ready"
+  // and "not in production" are different facts and need different work from different people.
+  const nogo = services.filter(s => s.gate === 'NO-GO').length
+  const undeployed = services.filter(s => s.gate === 'NOT-DEPLOYED').length
   const mp = services.filter(s => s.money_path)
 
   return (
@@ -88,6 +92,14 @@ export default function ReadinessPage() {
         <StatCard label={t('Služeb', 'Services')} value={services.length} />
         <StatCard label="GO" value={go} tone="success" icon={<CheckCircle2 size={16} />} />
         <StatCard label="NO-GO" value={nogo} tone="danger" icon={<XCircle size={16} />} />
+        {undeployed > 0 && (
+          <StatCard
+            label={t('Nenasazeno', 'Not deployed')}
+            value={undeployed}
+            tone="neutral"
+            icon={<CircleSlash size={16} />}
+          />
+        )}
         <StatCard
           label={t('Money-path', 'Money-path')}
           value={`${mp.filter(s => s.gate === 'GO').length}/${mp.length}`}
--- a/openbank-admin-ui/src/components/ui/tone.ts
+++ b/openbank-admin-ui/src/components/ui/tone.ts
@@ -146,6 +146,10 @@ const TONE_BY_STATUS: Record<string, Tone> = {
   GO: 'success',
   'NO-GO': 'danger',
   NOGO: 'danger',
+  // A released component with no gitops workload at all (#5760). Neutral on purpose: it is not a
+  // failed control (danger) and emphatically not a pass (success) — there is nothing running to
+  // judge. Reached only in place of NO-GO; see the collectors' computeGate.
+  'NOT-DEPLOYED': 'neutral',
 }
 
 /**
--- a/openbank-infra/scripts/generate-service-runbooks.py
+++ b/openbank-infra/scripts/generate-service-runbooks.py
@@ -83,8 +83,43 @@ def service_namespace(short: str) -> str:
     and the PodMonitor coverage gate — three tools were each growing their own copy of this
     question, and two of them had already answered it wrong in two different ways (#2255). Only
     the display fallback is local: a service with no workload still needs *something* to render.
+
+    That fallback is a LIE for the one service it applies to, which is why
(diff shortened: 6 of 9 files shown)


JiRaska:
## Not resolved — half of this PR duplicates merged work, and the other half is worth keeping

Picked this up in a conflict-resolution sweep. I merged `main` locally, inspected, and then aborted without pushing: the clean resolution needs one decision from you that I should not make unilaterally, because either way it drops test coverage someone deliberately wrote.

### The two halves

**Half one — the prod-readiness work — is unique and unmerged.** None of it is on `main`; it is ~237 added lines across 9 files:

```
openbank-admin-ui/scripts/collect-prod-readiness.mjs      +22 -2
openbank-admin-ui/src/app/api/prod-readiness/route.ts      +1 -1
openbank-admin-ui/src/app/system/readiness/page.tsx       +15 -3
openbank-admin-ui/src/components/ui/tone.ts                +4 -0
openbank-infra/scripts/prod-readiness-collector.py        +38 -1
openbank-infra/scripts/prod_readiness_collector_test.py   +44 -0
openbank-infra/scripts/generate-service-runbooks.py       +55 -1
docs/runbooks/svc-tax-reporting.md                        +13 -0
openbank-tax-reporting-service/CLAUDE.md                  +45 -0
```

**Half two — the pact work — was independently done on `main` while this PR sat**, and that is where both conflicts are:

| pact | consumer test on `main` | consumer test here | interactions on `main` | interactions here |
|---|---|---|---|---|
| `openbank-engagement-service-openbank-consent-service.json` | `EngagementToConsentPactConsumerTest` (#5790) | `MarketingConsentPactConsumerTest` | 1 | **2** |
| `openbank-campaign-service-openbank-consent-service.json` | `CampaignToConsentPactConsumerTest` | `MarketingConsentPactConsumerTest` | 2 | **4** |

Git reports the engagement pact as `AA` — both sides added the same path independently.

### Why I stopped

Two consumer test classes cannot both generate the same pact file. Whichever runs last wins, and `pact-drift-check.yml` asserts `git diff --exit-code -- pacts/` — so this would land as a gate that fails or, worse, passes on whichever ordering happened to occur.

The two ways out both lose something real:

- **Drop this branch's consumer tests** (keep `main`'s) — clean, no merged work reverted, but it silently discards the extra interactions, and those are the **deny-path** cases: *"a party who never opted in — the deny"*. On a marketing-consent check gating promotional impressions, the deny path is the one with actual consequences if it regresses.
- **Keep this branch's tests** — better coverage, but reverts #5790 and its sibling.

Deleting a deny-path contract interaction is not something a merge commit should do quietly, which is why this needs your call rather than my judgement.

### Suggested resolution

Salvage by subtraction, then restore the coverage deliberately:

1. Drop this branch's `MarketingConsentPactConsumerTest` (both services) and the two `pacts/*.json`, taking `main`'s. That alone makes this PR mergeable and its unique prod-readiness half lands.
2. Follow up by adding the missing interactions — the engagement deny case, and the two extra campaign cases — to `main`'s now-canonical `EngagementToConsentPactConsumerTest` / `CampaignToConsentPactConsumerTest`, regenerating the pacts from the tests.
3. Confirm consent-service's `@PactFolder` provider replay has state handlers for the new interactions — a consumer pact alone cannot catch a wrong request path.

Say the word and I will push exactly that; I did not want to delete your deny-path interactions on my own initiative.

The remaining conflict is trivial: `openbank-engagement-service/build.gradle.kts` differs only in the comment above the pact dependency block.

JiRaska:

3-dot diff against `origin/main`; branch is 150 commits behind, so please re-read `main` before merging. No content collision with #5905 or #6009 — this PR touches `collect-prod-readiness.mjs`, `api/prod-readiness/route.ts`, `system/readiness/page.tsx` and `tone.ts`, none of which those two go near.

The contract-test half is textbook and I have no findings on it. One real finding on the NOT-DEPLOYED verdict.

### 1. `NOT-DEPLOYED` is applied to dimensions it does not explain, and hides repo-fixable failures

`openbank-admin-ui/scripts/collect-prod-readiness.mjs:574-578`:

```
for (const { code } of DIMENSIONS) {
  const need = mp && critical.has(code) ? 3 : 2
  if ((scores[code] ?? 0) < need) return isUndeployed(short) ? 'NOT-DEPLOYED' : 'NO-GO'
}
```

The comment above it says the failing cells "are consequences of the absent workload, not controls someone skipped". That is an assertion about which dimensions failed, and the code never checks it. The **first** dimension below threshold decides the verdict, whatever it is. So for an undeployed service, a failing **C3 (contract tests)** — the exact class of gap the other half of this very PR fixes for campaign and engagement — reports as neutral `NOT-DEPLOYED` rather than `NO-GO`, and the readiness page renders it in a `neutral` tone with a "Not deployed" tile. A repo-fixable failure becomes indistinguishable from an infra decision nobody in this repo can close.

It is currently harmless because tax-reporting is the only undeployed service and it presumably fails a deployment-derived cell first, but that is an accident of `DIMENSIONS` ordering, not a property. The honest form is to name the deployment-dependent set explicitly:

```
const DEPLOYMENT_DEPENDENT = new Set(['C5', 'C7', 'C8', ...])
const failing = DIMENSIONS.filter(({code}) => (scores[code] ?? 0) < needFor(code))
if (failing.length === 0) return 'GO'
return isUndeployed(short) && failing.every(f => DEPLOYMENT_DEPENDENT.has(f.code))
  ? 'NOT-DEPLOYED' : 'NO-GO'
```

Then "NOT-DEPLOYED" means what the comment claims, and a service that is both undeployed *and* missing a contract test still reddens.

### 2. The self-test's two banner cases are vacuous the day tax-reporting deploys — and nothing says so

`openbank-infra/scripts/generate-service-runbooks.py:481-489`. The `if undeployed:` guard is honest and the comment acknowledges it, but the consequence is that the only falsification of `deployment_status()` disappears silently the moment the undeployed set empties, leaving the "a deployed service gets no banner" case, which passes against `return ""`. A synthetic case — `deployment_status` over a short name with no gitops presence, asserted to contain "NOT DEPLOYED" — costs nothing and keeps the function falsifiable in both directions regardless of the fleet's state. Same family as this repo's note that fixing a magnitude-based defect makes every test that discriminated *by* that magnitude vacuous while staying green.

### 3. `scoreC5Backup` ordering — worth a comment, not a change

`isStateless` is tested before `isUndeployed` (`:437-451`), so an undeployed *stateless* service gets `score: 2, 'n/a — stateless'` and never reaches the new branch. Correct, but the new branch's comment ("Still 0 — only the evidence changes") reads as unconditional. One clause about the stateless short-circuit would save the next reader the trace.

### Checked and clean — the contract-test half
- `ConsentPactProviderVerificationTest` is `@Provider("openbank-consent-service")` + `@PactFolder("../pacts")` with no `@EnabledIfSystemProperty` gate, so both new pacts are replayed on a PR, not only after a main push. The single-provider-class rule is preserved (this PR extends the existing class with new `@State` methods rather than adding a second one). This is the shape `check-pact-provider-replay.py` requires and `KNOWN_UNCOVERED` stays empty.
- `MarketingConsentPactConsumerTest` (engagement, `:179-181`) writes `EXPECTED_GRANTED_PATH` / `EXPECTED_DENIED_PATH` as **literals** and reflects only the outgoing request off the client's `@Path`, with an explicit assertion that the two agree. That asymmetry is the whole test and it is present — thank you for not making it symmetric.
- Both modules add `libs.pact.consumer` unconditionally, with no `CI=true` exclusion, so the drift check's derived scope will pick them up from the `@Pact(consumer=, provider=)` annotations without a hand edit.
- `gate` union widened consistently in the collector, the BFF route (`route.ts:26`), the page (`page.tsx:17`) and `TONE_BY_STATUS` (`tone.ts:149`) — no fourth consumer left on the two-value type.
- `undeployed` is counted rather than derived as `length - go`, and the tile is conditional on `> 0` so a fully-deployed fleet renders unchanged.
- Please confirm the admin-ui suites were run via `npm test`, not bare `npx vitest run` — the `pretest` hook bakes `governance.json`/`catalog.json`.

JiRaska:
## Rebuilt on current `main`; contract-test half removed as superseded

This branch was CONFLICTING and additionally red on `Regenerate consumer pacts and diff against committed`. Rebuilt cleanly on `origin/main` (`8b1868746`), new head `b5df5c7c9`.

### Why the pact files are gone

The contract-test half of this PR landed on `main` while the PR was open, under different class names. Measured against `origin/main`:

| This PR added | Already on `main` |
|---|---|
| `openbank-engagement-service/.../MarketingConsentPactConsumerTest.kt` | `openbank-engagement-service/.../EngagementToConsentPactConsumerTest.kt` |
| `openbank-campaign-service/.../MarketingConsentPactConsumerTest.kt` | `openbank-campaign-service/.../CampaignToConsentPactConsumerTest.kt` |
| `pacts/openbank-engagement-service-openbank-consent-service.json` | present, same provider + same request path |
| `pacts/openbank-campaign-service-openbank-consent-service.json` | present, 2 interactions |
| `testImplementation(libs.pact.consumer)` in both `build.gradle.kts` | already present in both (engagement line moved on `main`, campaign at line 65) |

Both `build.gradle.kts` conflicts were **comment-only** — the dependency line itself was already identical to `main`.

The interactions are the same call:

```
GET /api/v1/consents/party/{partyId}/grantee/party-service:marketing-comms/active
```

Keeping this PR's versions would have put **two consumer test classes generating the same pact file** for one consumer/provider pair — they overwrite each other, so `pact-drift-check` becomes order-dependent. `ConsentPactProviderVerificationTest` on `main` also states in its own KDoc that it "must remain the ONLY `@Provider("openbank-consent-service")` class in the repo", and this PR modified it to add duplicate states.

So the conflict was resolved by **subtraction**: the superseded contract-test files are dropped rather than merged. This branch no longer touches `pacts/` at all, which is also what clears the pact-regeneration check.

### What remains (the genuinely unique payload)

The `NOT-DEPLOYED` third verdict (#5760) and the runbook/collector work, none of which is on `main`:

- `openbank-infra/scripts/prod-readiness-collector.py` + `prod_readiness_collector_test.py`
- `openbank-infra/scripts/generate-service-runbooks.py`, `docs/runbooks/svc-tax-reporting.md`
- `openbank-admin-ui/scripts/collect-prod-readiness.mjs`, `src/app/api/prod-readiness/route.ts`
- `openbank-admin-ui/src/app/system/readiness/page.tsx`, `src/components/ui/tone.ts`
- `openbank-tax-reporting-service/CLAUDE.md`

All nine files applied to current `main` with **zero conflicts**.

### Verification

```
$ python3 openbank-infra/scripts/prod_readiness_collector_test.py
................................
Ran 32 tests in 0.161s

OK
```

`tone.ts` adds `'NOT-DEPLOYED': 'neutral'`; `'neutral'` is an existing `Tone` member and `dashboard/page.tsx` already renders the identical "Not deployed" `StatCard` with `tone="neutral"`, so the readiness page follows established precedent rather than introducing a new one.

Not merged — left for review.

JiRaska:
### `gates (lint)` red — transient GitHub API rate limit, not this diff

For the record, since a red check on a rebased PR invites the assumption that the rebase caused it. It did not.

The shard summary was `20 gates PASS=18 WARN=1 FAIL=1`, and the single failure is:

```
FAIL [lint] every required status-check context matches a job some workflow emits (ADR-0254, enforced)
  --- self-test (must PASS) ---
  self-test ok: ruleset-context-parity is falsifiable (7 cases)
  --- gate ---
  ::error::ruleset-context-parity: gh api repos/JiRaska/open-bank-oss/rulesets failed (rc=1):
    gh: API rate limit exceeded for installation ... (HTTP 403)
```

The gate never reached its subject — it could not read the ruleset at all. This branch touches only `openbank-admin-ui/`, `openbank-infra/scripts/`, `docs/runbooks/` and a `CLAUDE.md`; it changes no workflow and no required-context name, so it cannot affect ruleset/job-name parity either way.

Same rate limit hit `gates (supplychain)` on #5723 in the same window, where the gate there degraded correctly instead of failing:

```
::notice::stale-comment-references: repo rule UNRESOLVED ... Not a pass and not a failure.
```

Two gates, one cause, two different verdicts — the hard-fail is the wrong one, and I have filed that inconsistency separately rather than papering over it here.

The failed job is queued for rerun once the workflow run finishes (`gh run rerun --failed` refuses while the run is still active). No code change needed for this check.

PR #5836, opened 2026-08-20
docs(infra): cite the ADR that actually decided this (#5785)

JiRaska:
## What

Ten sites in `openbank-infra` cited **ADR-0073** (*"Hardware-backed credential storage for the
customer app"* — Android Keystore / iOS Keychain) for onboarding and account lifecycle. This is
the last tree in the #5785 sweep. **Three different ADRs**, decided per site — not a
find-and-replace.

| site | old | new | why |
|---|---|---|---|
| `components/kyc/kyc-service.yaml` (`OPENBANK_KYC_AUTO_APPROVE`) | ADR-0073 | **ADR-0116 §4** | ADR-0116 §4 *is* sandbox straight-through KYC: names the property, the `false` default, auto-approve on a `PARTY_CREATED` case, "MUST remain false in production" |
| `components/accounts/account-service.yaml` (`SANCTIONS_SERVICE_URL`) | ADR-0073 | **ADR-0032 §C** | the fail-closed screening gate for account opening; account-service's own `AccountScreeningPort.kt` already cites ADR-0032 §C for the same behaviour |
| `components/accounts/kafka-account-mtls.yaml` (consumes `openbank.party.events`) | ADR-0073 | **ADR-0266 §1** | `PARTY_CREATED` opens inert accounts |
| `components/balances/kafka-balance-mtls.yaml` (consumes `account.created`) | ADR-0073 | **ADR-0266 §3** | event-driven balance init |
| `components/party/kafka-party-mtls.yaml` (kyc.events + aml.events reads, 2 sites) | ADR-0073 | **ADR-0266 §2** | the KYC+AML two-key activation gate |
| `components/aml/kafka-topics.yaml` | ADR-0073 | **ADR-0266 §2** | same gate |
| `components/aml/namespace.yaml` | ADR-0073 | **ADR-0266 §2** | same gate |
| `components/aml/kafka-aml-mtls.yaml` (header + ACL comment, 2 sites) | ADR-0073 | **ADR-0266 §2** | auto-open onboarding screenings from `PARTY_CREATED` |
| `components/aml/aml-service.yaml` (header) | ADR-0073 | **ADR-0266 §2** | same |
| `apps/aml.yaml` | ADR-0073 | **ADR-0266 §2** | same |

### One site has no ADR behind it, and this PR says so rather than picking a number

`components/aml/aml-service.yaml`'s `OPENBANK_AML_AUTO_CLEAR` is **decided by no ADR**. ADR-0116 §4
[...]

PR #5905, opened 2026-08-20
feat(admin-ui): enrich operator cockpit

### LINKED ISSUE #5904: feat(admin-ui): make operator surfaces evidence-backed
## Problem

The admin console mixes real live data, stale/static reference views, and missing operator drill-downs. This makes several surfaces look empty or healthy without enough evidence.

## Acceptance criteria

- [ ] Derive test inventory from released and money-path services; show JUnit pass rate, Kover coverage, and freshness separately.
- [ ] Distinguish Trace Explorer loading, empty, and failed states.
- [ ] Resolve KYC customers by name/company/UUID through party-service.
- [ ] Add account, lending, AML, and document context to Customer 360 with independent failure states.
- [ ] Expose exact compliance-pack content plus proposal/decision audit metadata under existing authorization.
- [ ] Label Temporal reference diagrams honestly and show live workflow metrics where available.
- [ ] Render charter-backed agent identity in the approval queue.
- [ ] Block regulatory exports when data is incomplete, missing, or unbalanced.
- [ ] Cover the new behavior and pass UI/lending release gates.

## Explicit boundary

Live XBRL/DPM generation and authenticated ČNB submission are not implemented by this UI slice and are tracked by #5914. This work must not add a fake submit-success path or embed credentials.

Refs #5914

JiRaska:
## Summary

Make the Admin UI operator surfaces evidence-backed, customer-centric, and explicit about incomplete external integrations. Add the lending compliance-pack projection needed for a real operator drill-down.

## Type of change

- [x] fix (bug fix)
- [ ] perf (performance)
- [ ] docs
- [ ] chore (build / tooling)
- [ ] breaking change

## Linked issues

Closes #5904
Refs #5914

## Changes

- Derive fleet test evidence and Kover coverage from authoritative service inventory with freshness.
- Add KYC name search, Customer 360 portfolio/documents, compliance detail, Temporal live metrics, Trace loading state, and agent identity.
- Block misleading regulatory exports when source data is incomplete or unbalanced.

## Definition of Done

- [ ] Contract tests updated (if public API changed) — additive response fields only; existing consumers remain compatible.
- [x] Relevant detekt, ktlint, tests, Quarkus build, and Next production build pass locally.
- [x] No new lint warnings.
- [x] OpenAPI spec updated (additive 1.13.0 contract).
- [x] Flyway migration added + rollback note (if DB changed) — N/A, no DB change.
- [x] Event schema versioned backward-compatibly (if event changed) — N/A, no event change.
- [x] Docs updated — lending threat model records the read projection.

## Security checklist

- [x] No new suppressions or unsafe casts.
- [x] Auth boundaries unchanged; exact pack detail remains on the existing compliance/admin role-gated resource.
- [x] PII access remains behind the existing authenticated BFF and service roles.
- [x] No cardholder-data path changed.

## Compliance impact

- [ ] PCI DSS
- [x] DORA
- [x] GDPR
- [ ] PSD2 / SCA / RTS
- [x] AML / 5AMLD
- [x] CNB reporting
- [ ] None

The UI now distinguishes measured evidence from unknown/stale states and refuses to present incomplete regulatory data as export-ready. Live regulator submission remains #5914.

## Rollout / Rollback

- Deployment strategy: existing admin-ui and lending canary pipelines
- Rollback plan: revert the three commits; the API change is additive and has no persistence migration

## Screenshots / demo

Protected routes redirect to Keycloak in local rendering; behavior is covered by focused UI tests and the production Next build.

## Reviewer notes

Money-path lending change requires the repository's review policy. Please focus on compliance-pack disclosure scope and the CI artifact-selection logic.

### DIFF
--- a/.github/scripts/check-external-feeds.py
+++ b/.github/scripts/check-external-feeds.py
@@ -173,6 +173,7 @@ FEEDS = [
 NOT_PROBED = [
     ("https://api.github.com", "authenticated API, not a data feed; failure is loud at call time"),
     ("https://api.deepinfra.com/v1/openai", "authenticated LLM gateway; needs a key"),
+    ("https://integrate.api.nvidia.com/v1", "authenticated LLM gateway; needs a key"),
     ("https://s3.eu-north-1.amazonaws.com", "AWS endpoint, reached with SigV4 credentials"),
     ("https://kc.open-bank.tech/realms/openbank-customers", "our own Keycloak realm, covered by its own probes"),
     ("https://pid.open-bank.tech", "our own PID issuer, covered by its own probes"),
--- a/.github/scripts/check-openapi-enum-vs-domain.py
+++ b/.github/scripts/check-openapi-enum-vs-domain.py
@@ -100,6 +100,8 @@ BASELINE: dict[str, str] = {
         "AccountingDayStatus; OPEN is unreachable as a transition target (no reopen, ADR-0207).",
     "openbank-lending-service:APPROVED,EXECUTED,PENDING,REJECTED":
         "#5962 — CollateralStatus: spec-only EXECUTED",
+    "openbank-lending-service:APPROVED,EXECUTED,PROPOSED,REJECTED":
+        "Compliance pack ProposalState, not CollateralStatus; value-overlap pairing is ambiguous.",
     # Also deliberate: the enum is right to flag (INDIVIDUAL has never existed; the DB CHECK is
     # ('NATURAL_PERSON','LEGAL_ENTITY','SOLE_TRADER')), but it sits inside `CreatePartyRequest`,
     # whose declared properties — legalName, tradingName, taxId, dateOfBirth, nationality —
--- a/.github/workflows/_service-ci.yml
+++ b/.github/workflows/_service-ci.yml
@@ -351,15 +351,17 @@ jobs:
       # A flake cannot be counted from run CONCLUSIONS alone (a re-run flips it away from
       # `failure`), so retain the JUnit XML per ATTEMPT — a flake becomes a diffable artifact
       # pair instead of something that has to be noticed and reported at re-run time (#4878).
-      - name: Upload JUnit XML (every run — flake accounting)
+      - name: Upload quality evidence (every run)
         if: always()
         continue-on-error: true
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
-          name: junit-xml-${{ inputs.service }}-a${{ github.run_attempt }}
-          path: ${{ inputs.service }}/build/test-results/**/*.xml
+          name: quality-evidence-${{ inputs.service }}-a${{ github.run_attempt }}
+          path: |
+            ${{ inputs.service }}/build/test-results/**/*.xml
+            ${{ inputs.service }}/build/reports/kover/report.xml
           if-no-files-found: ignore
-          retention-days: 3
+          retention-days: 7
 
       # Compatibility projection for the image build: the latest envelope remains
       # separate from immutable attempts above, so a re-run cannot erase history.
--- a/.github/workflows/admin-ui-deploy.yml
+++ b/.github/workflows/admin-ui-deploy.yml
@@ -35,6 +35,18 @@
 name: Admin-UI deploy
 
 on:
+  # CI evidence changes far more frequently than the Admin UI source. A successful
+  # main Services CI and CI immediately rebuild the read-only evidence projection; the
+  # daily run is its recovery path when GitHub suppresses or delays an event. Both
+  # use the same signed image and GitOps-PR path as source-triggered deploys.
+  workflow_run:
+    # Services CI produces fleet envelopes; CI produces the Admin UI envelope.  Subscribe
+    # to both: refreshing after Services CI alone can snapshot a stale Admin UI verdict
+    # before the same main commit's UI test artifact is available.
+    workflows: ["Services CI", "CI"]
+    types: [completed]
+  schedule:
+    - cron: '17 3 * * *'
   push:
     branches: [main]
     paths:
@@ -92,7 +104,14 @@ jobs:
     # scopes the deploy commit to the gitops manifest alone, so the loop can no longer form by
     # that route. This guard is kept as belt-and-braces: it costs nothing and still holds if a
     # future step writes into openbank-admin-ui/** again.
-    if: "!startsWith(github.event.head_commit.message, 'chore(admin-ui): deploy')"
+    # `workflow_run` is trusted only after a successful Services CI on main.  It
+    # executes this default-branch workflow, never unreviewed PR YAML, and only
+    # refreshes a read-only projection from CI artifacts.  Keep the old push
+    # loop-breaker; workflow_run has no head_commit, hence the explicit fallback.
+    if: >-
+      (github.event_name != 'workflow_run' ||
+       (github.event.workflow_run.conclusion == 'success' && github.event.workflow_run.head_branch == 'main')) &&
+      !startsWith(github.event.head_commit.message || '', 'chore(admin-ui): deploy')
     # Native arm64 GitHub-hosted runner (free on public repos): the admin-ui image
     # is linux/arm64. ECR push + cosign use the same GitHub-OIDC role as before.
     runs-on: ubuntu-24.04-arm
@@ -112,6 +131,9 @@ jobs:
       # Mirrored into env because the `secrets` context is NOT available in a step-level `if:`
       # — only job-level env can read it. Used solely to decide whether to mint an App token.
       GITOPS_APP_ID: ${{ secrets.GITOPS_APP_ID }}
+      # Read-only token for the separate private customer-app repository. It is optional:
+      # missing scope/evidence renders an explicit blocked client row; it never blocks deploy.
+      APP_RO_TOKEN: ${{ secrets.OPENBANK_APP_RO_TOKEN }}
 
     outputs:
       tag: ${{ steps.build.outputs.tag }}
@@ -123,6 +145,16 @@ jobs:
           fetch-depth: 0
           # persist-credentials: true (default) — needed so git in the collectors
           # can read the repo; the gitops step uses a separate checkout.
+
+      - name: Checkout mobile RUM capability metadata
+        if: env.APP_RO_TOKEN != ''
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          repository: JiRaska/openbank-app
+          token: ${{ secrets.OPENBANK_APP_RO_TOKEN }}
+          path: .app-src
+          persist-credentials: false
+          fetch-depth: 1
 
       - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
         with:
@@ -273,7 +305,7 @@ jobs:
           done
           echo "SBOM staging complete."
 
-      - name: Stage JUnit test results (per-service latest artifact)
+      - name: Stage Test Intelligence evidence (per-service latest artifact)
         # services-ci.yml uploads one artifact per service: test-reports-{service}
         # (7-day retention). The old approach (download from ONE run) fails with
         # path-scoped CI: only recently-changed services have artifacts in any single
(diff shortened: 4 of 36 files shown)


JiRaska:
## Not resolved — this branch is held in another worktree, and the conflicts are substantive UI merges

Picked this up in a conflict-resolution sweep and stopped at the liveness check.

### Worktree state (probed 2026-08-21 14:48Z)

`codex/admin-ui-operator-cockpit` is checked out at `/private/tmp/open-bank-oss-admin-e2e`. All three idleness probes read quiet — tree clean, last commit 2026-08-20 13:53, zero files modified in the preceding 60 minutes (`find -mmin -60`; note `-newermt "-60 minutes"` silently returns nothing on macOS and is not a valid probe here).

But the more telling fact is a mismatch: that worktree's HEAD is **`d87891e75`**, while the remote branch tip is **`32fa81157`** — a later merge commit made from somewhere else. So the branch has been driven from a second location and this checkout is already stale relative to it. Pushing a resolution would move the ref under a tree that is behind it, which is exactly the situation where a stale local branch quietly wins later. A quiet worktree is not evidence of abandonment, and here it is not even the branch's current driver.

### The conflicts

Five files, and they are real content merges rather than derived-file noise:

```
docs/threat-models/openbank-lending-service.md
openbank-admin-ui/src/app/approvals/page.tsx
openbank-admin-ui/src/app/lending/compliance-packs/page.tsx
openbank-admin-ui/src/app/regulatory/page.tsx
openbank-admin-ui/src/test/compliance-packs-console.test.tsx
```

`main` is 135 commits ahead. Three of these are operator-facing pages where a bad merge produces a plausible-looking console that is subtly wrong, and one is a threat model — a money-path lending document where merged-away content is not something CI will catch.

Verifying such a resolution needs a full `npm test` cycle in admin-ui (via `npm test`, never bare `npx vitest run` — the `pretest` hook bakes `governance.json` / `catalog.json`, and without them several suites fail in a way that reads exactly like a `main` regression).

### Recommendation

Best resolved by whoever owns the second checkout driving `32fa81157`, since they have the context for the three page merges. If that work is finished and the branch is free, say so and I will take it — the resolution itself is tractable, it is the ownership that stopped me, not the difficulty.

JiRaska:

Read as a 3-dot diff (`git diff origin/main...origin/codex/admin-ui-operator-cockpit`). **154 commits behind `main`, and `mergeable: CONFLICTING`** — the findings below are against the branch head as it stands today.

### 1. BLOCKING — an unreadable charter registry silently turns every AI proposal into a human one, and drops the ADR-0080 caution banner with it

`src/app/api/agent/proposals/route.ts:39-51`:

```
const registry = await loadAgentCharters()
const charterIds = new Set(registry.agents.map(agent => agent.id))
...
icon: known ? 'bot' : 'user',
charterKnown: known,
```

`loadAgentCharters()` returns `{ available: boolean, ... }` precisely because `agents.yaml` may not be in the image — `openbank-admin-ui/src/lib/governance/agentCharters.ts:66-70` sets `available:false` and an **empty** `agents` array when the file is missing or unparseable. This route ignores `available`. So when the registry cannot be read:

- `charterIds` is empty → `known` is `false` for every row,
- every proposal gets `icon: 'user'`,
- `src/app/approvals/page.tsx:207` computes `aiGenerated = p.agent ? p.agent.icon === 'bot' : /assistant|agent|\bai\b/i.test(p.proposedBy)`, and since `p.agent` is *always* set by the route, the regex fallback the comment above it describes is **dead code**,
- so `aiGenerated` is false everywhere and the AI-generated badge plus the caution banner both disappear.

The failure is silent and it fails **open**: an operator approving an agent-authored proposal sees a plain human-authored card. This is the same shape as `PushResult.skipped()` carrying `success = true` (#4348) — "could not check" collapsed into "checked, negative". Three states are needed (chartered / unresolved / unverifiable), not a boolean.

Second, smaller: `charterIds.has(id)` matches the raw `proposedBy` string exactly. A Keycloak service account arrives as `service-account-<clientId>` and will never match a charter id, so real chartered agents also render as `user`.

### 2. BLOCKING — direct collision with #6009, on the same two files and the same decision

Not a filename overlap; the same semantics implemented twice, incompatibly:

- `src/app/approvals/page.tsx` — both PRs rewrite the identical `const aiGenerated = /assistant|agent|\bai\b/i.test(p.proposedBy)` hunk. #6009 fetches the registry client-side via a new `/api/governance/agent-identities` route and resolves through `resolveAgentIdentity()` with the three states above; this PR pushes an `agent` object down from the proposals route. Both call `loadAgentCharters`. Only one can land.
- `src/app/kyc/page.tsx` — both insert a party search above the case filter. This PR reuses the existing shared `@/components/party/PartySearch` (already on `main`, ADR-0208, two callers); #6009 adds a **new** `@/components/kyc/PartyLookup` + `src/lib/party/resolveParty.ts`. Merging both gives the fleet two divergent search components over one endpoint, which is the exact thing PartySearch's header comment says it exists to prevent.

Reusing the shared component is the better call here. The identity mechanism is the other way round — see #1.

### 3. The `/search` 404 conflation — clean here, worth knowing

`PartySearch` (`main`) maps a non-401/403 `!res.ok` to `unreachable`, so party-service's flag-disabled `/search` 404 does **not** render as "no party matches". Correct, though it reads as "unreachable" rather than "search is switched off". One narrower nit at `PartySearch.tsx:93`: `body.data ?? []` only understands the `{data:[...]}` envelope — a bare array or `{items:[...]}` renders as a genuine empty result.

### 4. The workflow half is the strongest part of this PR — credit where due, with one hole

Replacing the three hand-kept service lists with `releasedJvmServices()` derived from `release-please-config.json` (`scripts/lib/service-inventory.mjs:14-21`) is exactly the right direction: a gate whose scope is a hand-kept list of the thing it checks reads as passing when the list is short. `npm ci` at `admin-ui-deploy.yml:139` precedes the staging step at :276, so the `yaml` import resolves.

But the derived list has no floor. `.github/workflows/admin-ui-deploy.yml:296`:

```
mapfile -t SERVICES < <(node openbank-admin-ui/scripts/list-test-services.mjs --repo .)
```

`mapfile` reading from process substitution exits 0 regardless of what `node` did. If the script throws (a moved `release-please-config.json`, a bad parse), `SERVICES` is empty, the loop body never runs, and the step prints `Staged 0 JUnit XML files from 0/0 services.` and goes green — the UI then shows an empty test estate that looks like a fleet with no tests. Assert `[ "${#SERVICES[@]}" -ge 20 ]` (or whatever floor is honest) and fail the step below it, the same way `PG_UP != 1` is a hard `::error::`.

Also stale: the comment block at `admin-ui-deploy.yml:276-288` still describes `test-reports-{service}` and `?name=...&per_page=1`, both of which this diff replaced.

### 5. `test-results/route.ts` — the legacy path silently changed shape

`observedServices` (`:113-115`) derives the list from rows the DB has seen. A service whose rows aged out now vanishes from the page entirely, where before it rendered with `tests: 0`. "Absent" and "ran zero tests" are different facts and the second is the one an operator can act on. The comment says the bundled path is authoritative, which is fine — but say what the legacy path now omits.

### Checked and clean
- `src/app/api/approvals/pending/route.ts` is **not** touched by this PR (nor by #6009) — the 10 sources on `main` are intact.
- No `NEXT_PUBLIC_*` introduced or moved.
- The artifact pagination loop terminates correctly, and the `[ "${page_count}" -lt 100 ] && break` shape is exempt from `set -e` (the test is not the last command of the `&&` list).
- `min_subjects`-style scope derivation in `service-inventory.mjs` correctly refuses to copy `money_path_services` and throws on a malformed list.

JiRaska:
## `/api/agent/proposals` ignores `available`, so an unreadable `agents.yaml` marks every proposal human

**Not pushed** — this branch is `CONFLICTING` (`mergeStateStatus: DIRTY`, conflict in
`openbank-admin-ui/src/app/kyc/page.tsx`) and has a live worktree, so it needs a rebase rather
than a patch on top. The exact change is below, verified locally against this branch's head
(`bfb50a21b`).

### The defect

`src/app/api/agent/proposals/route.ts:39-51` reads the charters and drops the `available` flag
that `loadAgentCharters()` returns (`agentCharters.ts:66-70`):

```ts
const registry = await loadAgentCharters()
const charterIds = new Set(registry.agents.map(agent => agent.id))
...
const known = charterIds.has(id)
return { ...row, agent: { id, displayName, icon: known ? 'bot' : 'user', charterKnown: known } }
```

When `agents.yaml` cannot be read, `readRaw()` returns `{ available: false, data: null }`, so
`registry.agents` is `[]` and **every** `charterIds.has(id)` is `false`. That collapses into the
same `icon: 'user'` a genuine human desk gets. There is no error, no log line, and no visual
difference — the queue just quietly asserts that every proposal was written by a person.

It is worse than a no-op on the page, because the client-side fallback cannot rescue it:

```ts
const aiGenerated = p.agent ? p.agent.icon === 'bot' : /assistant|agent|\bai\b/i.test(p.proposedBy)
```

`p.agent` is always present (the route always sets it), so the truthy branch always wins, the
`proposedBy` regex is never consulted, and `aiGenerated` is `false`. Downstream that removes the
AI badge, the amber `borderLeft`, and the **ADR-0080 caution banner** — on real agent proposals,
in the reassuring direction, silently.

Measured against this branch's route with `loadAgentCharters` mocked to `{ available: false,
agents: [] }` and a proposal from the genuinely chartered `fraud-investigator`:

```
emitted agent.icon === 'user'      -> assertion PASSES against the current route (exit 0)
=> page.tsx aiGenerated === false  -> badge and ADR-0080 banner gone
```

### The fix — mirror #6009's three states, do not invent a fourth shape

#6009 (`lib/governance/agentIdentity.ts`) already solved exactly this with
`chartered | unresolved | unverifiable`, and its reasoning applies verbatim: an unreadable
registry is *absence of evidence*, not *evidence of a human*. The route should speak the same
vocabulary rather than grow its own.

```diff
@@ -12,6 +12,14 @@ import { loadAgentCharters } from '@/lib/governance/agentCharters'
 
 export const dynamic = 'force-dynamic'
 
+/**
+ * chartered    — proposedBy maps to an agents.yaml entry.
+ * unresolved   — the registry WAS read and has no such id. Real evidence of absence.
+ * unverifiable — the registry could not be read at all. Not the same as absence, and must not
+ *                render as it. Same vocabulary as lib/governance/agentIdentity.ts (#6009).
+ */
+export type AgentStatus = 'chartered' | 'unresolved' | 'unverifiable'
+
 function agentBase(): string {
   if (process.env.SERVICES_HOST === 'container') return 'http://openbank-agent-service:8109'
   return (process.env.AGENT_SERVICE_URL ?? 'http://localhost:8109/mcp').replace(/\/mcp$/, '')
@@ -41,12 +49,26 @@ export async function GET(req: NextRequest) {
     const charterIds = new Set(registry.agents.map(agent => agent.id))
     const enriched = Array.isArray(rows) ? rows.map(row => {
       const id = row.proposedBy ?? 'unknown'
-      const known = charterIds.has(id)
+      // THREE states, mirroring lib/governance/agentIdentity.ts (#5904/#6009). The `available`
+      // flag is load-bearing: when agents.yaml cannot be read, `registry.agents` is [] and every
+      // `charterIds.has(id)` is false. Collapsing that into the same `false` as a genuine miss
+      // marked EVERY proposal as human-authored — dropping the AI badge and the ADR-0080 caution
+      // banner on real agent proposals, silently and in the reassuring direction. An unreadable
+      // registry is absence of evidence, never evidence of a human.
+      const status: AgentStatus = !registry.available
+        ? 'unverifiable'
+        : charterIds.has(id)
+          ? 'chartered'
+          : 'unresolved'
       return { ...row, agent: {
         id,
         displayName: id.split('-').map(word => word.charAt(0).toUpperCase() + word.slice(1)).join(' '),
-        icon: known ? 'bot' : 'user',
-        charterKnown: known,
+        status,
+        // `charterKnown` stays true ONLY for a positively chartered actor, so any existing
+        // reader keeps its meaning. `icon` gains a third value rather than reusing 'user'.
+        icon: status === 'chartered' ? 'bot' : status === 'unverifiable' ? 'help' : 'user',
+        charterKnown: status === 'chartered',
+        charterRegistryAvailable: registry.available,
       } }
     }) : rows
     return NextResponse.json(enriched, { status: res.status })
```

`charterKnown` keeps its exact previous meaning (true only for a positively chartered actor), so
existing readers are unaffected; `icon` gains a third value instead of overloading `'user'`.

### Proof, both directions

Against this branch, `npm test` (never bare `npx vitest run` — the `pretest` hook bakes
`governance.json` / `catalog.json`):

| route | test file | exit | result |
|---|---|---|---|
| current (defective) | new `agent-proposals-identity.test.ts` | **1** | 4/4 fail — `expected undefined to be 'unverifiable'` |
| fixed | same file, unchanged | **0** | 4/4 pass |
| current (defective) | probe asserting `icon === 'user'` for a real agent | **0** | passes — documents the fail-open |
| fixed | same probe | **1** | `AssertionError: expected 'help' to be 'user'` |

Full admin-ui suite on the fixed tree: **251 files, 1657 passed, 1 skipped, exit 0.**

Worth noting: the two existing cases in `agent-proposals-identity.test.ts` mock
`loadAgentCharters` as `{ agents: [...] }` with **no `available` field at all**, which is why this
never showed up — the suite had no case where the flag existed, let alone was `false`. The
replacement file sets it explicitly in every case.

### These two PRs collide — please resolve by subtraction

Both this PR and #6009 replace the **same line**, `approvals/page.tsx:208`:

- #6009: `const identity = resolveAgentIdentity(p.proposedBy, registry)` — three states, an
  `AgentIdentityBadge`, and `cautionAi = identity.status !== 'unresolved'` so the ADR-0080 banner
  stands whenever AI authorship is established *or cannot be ruled out*.
- this PR: `const aiGenerated = p.agent ? p.agent.icon === 'bot' : /…/.test(p.proposedBy)` — two
  states, and fail-open as described above.

They also both touch `kyc/page.tsx`, where this branch is already conflicting.

**Suggestion:** drop this PR's `approvals/page.tsx` `aiGenerated` hunk entirely and let #6009 own
the page, keeping the route fix above (which #6009 does not touch — it reads a separate
`/api/governance/agent-identities`). #6009's design is the more complete of the two and its
badge can consume the `status` field the route now emits. Merging both as they stand puts two
different provenance vocabularies on one screen.

JiRaska:
Heads-up before you rebase: I am about to merge #6463, and it touches the one file this PR shares
with it — `openbank-admin-ui/src/app/kyc/page.tsx`.

The two edits collide in the import block:

- #6463 (CodeQL cleanup) deletes `import { PartyLookup } from '@/components/kyc/PartyLookup'`
- this PR replaces the same import with `import { PartySearch, type PartyHit } from '@/components/party/PartySearch'`

Different intent, adjacent lines, so expect this PR to go `DIRTY` once #6463 lands. It happened
earlier today between #5836 and #5748 on a gitops manifest, where the two edits were about eight
lines apart and git still called it a conflict — hunk distance is not the test.

The resolution is the trivial one: take this PR's `PartySearch` import and drop `PartyLookup`
entirely, which is what both changes independently want.

I am merging #6463 rather than holding it because this PR is blocked for human review
(it modifies `.github/workflows`), and that has no fixed horizon — parking a green PR behind an
indefinitely blocked one costs more than one rebase does.

JiRaska:
**`openapi.yaml` `info.version` 1.14.0 is already taken on `main`.** This PR bumps
1.13.0 → 1.14.0, but #6230 landed that same number while this branch was open, so

```
$ git show origin/main:openbank-lending-service/src/main/resources/openapi.yaml | grep -m1 '^  version:'
  version: 1.14.0
```

The next free number is **1.15.0**.

last run predates #6230. That is the one shape no base resolution can close, which is

Worth re-checking rather than just applying 1.15.0 — read the version off live `main` at

```bash
git fetch origin main && git show origin/main:openbank-lending-service/src/main/resources/openapi.yaml | grep -m1 '^  version:'
```

consent-service `1.7.0`.

JiRaska:
**#5905 and #6235 both bump `openbank-lending-service`'s `openapi.yaml`, and the merge order is now
load-bearing.** Neither is wrong; they only work in one sequence.

```
main    1.14.0
#5905   1.15.0
#6235   1.16.0
```

**#5905 must land first.** In that order both are correct and each step is a single minor. The other
way round, `main` reaches 1.16.0 and #5905 arrives claiming 1.15.0 — a version *behind* the published
contract, which the api-contract gate classifies against the merge-base as of its own run and will
not necessarily catch if #5905's last run predates the merge.

Nothing enforces this. The two PRs conflict in no file, both are green, and git sees no relationship
between `1.15.0` and `1.16.0`.

For context on how quickly this moves: yesterday #5905 claimed 1.14.0, which #6230 had already taken;
it has since been corrected to 1.15.0, and #6235 was already at 1.16.0. Re-read the version off live
`main` at the moment you merge rather than trusting this comment:

```bash
git fetch origin main && git show origin/main:openbank-lending-service/src/main/resources/openapi.yaml | grep -m1 '^  version:'
```

Two other spec-version items from the same sweep are still open and unrelated to these:
**#6019** claims `openbank-card-issuance-service` `1.8.0`, which `main` already holds (taken by
#4308) — it needs 1.9.0; and **#6017 / #6235** both claim `openbank-consent-service` `1.7.0`, a live
race where whoever goes second needs 1.8.0.

JiRaska:
Navazuji na nezavisle code review: opraven degradovany rezim identity agenta. Pokud nelze nacist registry charteru, BFF uz neoznaci autora jako cloveka; UI zachova konzervativni AI/provenance fallback dle ADR-0080. Ověřeno: type-check, 13 cílených Vitest testů a 7/7 operator-cockpit Playwright E2E. Prosím o nezávislé review nového commitu 3a9bcd642; PR vyžaduje dvě nezávislá schválení, merge neprovádím.

JiRaska:
openbank-admin-ui/src/app/lending/compliance-packs/page.tsx:274: 🔴 bug: parent `onClick` catches clicks from the decision-reason input and approve/reject buttons, opening the detail overlay before the checker can complete the action. Move detail opening to an explicit control or stop propagation on the interactive controls, with a regression test.

PR #5906, opened 2026-08-20
fix(agents): refuse instead of returning a fabricated proposal URL in docs-truth-agent and authz-policy-auditor

### LINKED ISSUE #5897: five AI-agent proposal adapters return fabricated pending-* URLs into a third-party repository and record them as PROPOSED, instead of refusing
Two agent proposal adapters return a **fabricated URL** instead of refusing when they cannot open a
real PR or issue. Verified on `origin/main`:

```
openbank-docs-truth-agent/.../adapter/GitHubProposalAdapter.kt:39
  "https://github.com/openbank/openbank/pulls/pending-docs-truth-agent-${finding.id.take(...)}"
openbank-docs-truth-agent/.../adapter/GitHubProposalAdapter.kt:50
  "https://github.com/openbank/openbank/issues/pending-docs-truth-agent-${...}"

openbank-authz-policy-auditor/.../adapter/GitHubProposalAdapter.kt:43, :56
  the same two shapes with `pending-authz-policy-auditor-`
```

## Two defects, the second one found while checking the first

**1. The value is fabricated, not an error.** The method returns a well-formed URL string, so every
caller — an audit record, a dashboard, a report, a human reading a finding — sees what looks like a
delivered proposal. This is the shape this repo keeps paying for: `PushResult.skipped()` carrying
`success = true` counted undelivered pushes as delivered (ADR-0252 phase 0); `LoggingDeadLetterSink`
made "quarantined" mean `log.warnf` (#5761). A no-op must not share its return shape with a real
result.

**2. The host is wrong too.** `github.com/openbank/openbank` is not this repository — it is
`JiRaska/open-bank-oss`. So even read as a placeholder, the string points somewhere that does not
exist, and nobody following one would land on a 404 that explains anything.

## The fleet already has the right pattern

`openbank-mcp-service` settled this with `UnwiredProposalPort`
(`infrastructure/read/UnwiredProposalPort.kt`) — an unwired port that **refuses** rather than
inventing an answer. That is the precedent (#3900 fixed the same class in mcp: it used to answer
`PROPOSED` for a proposal nobody recorded). `flaky-test-hunter` has since gone further and has a
real token-backed, fail-closed `GitHubProposalAdapter` with degraded-path tests.

So these two are the stragglers, and both directions are already demonstrated in-repo: refuse
(`UnwiredProposalPort`) or actually wire it (`flaky-test-hunter`).

## What would close this

Per adapter, pick one and make the outcome distinguishable:
- refuse, following `UnwiredProposalPort`, so a caller cannot mistake it for a delivered proposal; or
- wire it token-backed and fail-closed, following `flaky-test-hunter`'s adapter, with a test for the
  no-token path.

Either way the fabricated-URL branch goes, and a test should prove the unwired/failed outcome is not
representable as a success. Fix the repository host in any string that survives.

Found while delivering #5281 (PR #5890), which deliberately scoped to `flaky-test-hunter`.

Refs #5281, PR #5890, #3900.

JiRaska:
## What this changes

Both remaining bootstrap `GitHubProposalAdapter` stubs — `docs-truth-agent` and
`authz-policy-auditor` — returned a fabricated URL instead of refusing:

```
https://github.com/openbank/openbank/pulls/pending-docs-truth-agent-<id>
https://github.com/openbank/openbank/issues/pending-authz-policy-auditor-<id>
```

Two defects, as #5897 sets out. The value was a **no-op sharing its shape with a real result** —
and the host was a repository that does not exist (this one is `JiRaska/open-bank-oss`).

## Per adapter: refuse, not wire

**Both refuse.** Both charters *do* grant `tier: write_proposal` on `github-pr`, so wiring would
not have been a charter widening — but neither service has a `github-token` config, so there is no
token path to fail closed on, only one to build. `flaky-test-hunter` (PR #5890) is the one agent
that has one, and it stays the template if either is wired later. Refusing is the safer default and
matches `openbank-mcp-service`'s `UnwiredProposalPort` (#3900).

`authz-policy-auditor.openProposalPr` refuses **permanently**, not pending wiring: ADR-0167's
Decision is that this agent never opens a fix PR on an authorization policy.

**`agents.yaml` is untouched**, so none of its three derived tails (`docs/compliance/eu-ai-act.md`,
the OPA bundle ConfigMaps + `policy-checksum` annotations, `ai-governance-snapshot.json`) needed
regenerating.

## Proving the refusal cannot read as success

Asserting "the adapter returns an error" alone would be the weak test this repo keeps warning
about — the old stub also returned a value, and the whole defect lived in what the **caller** did
with it. `GitHubProposalPort` now returns `String?`; the load-bearing change is in
`DiagnoseAndProposeActivityImpl`, which on `null` leaves the finding **`DIAGNOSED`** with a null
`proposalUrl` and a null `proposedAt`, instead of moving it to `PROPOSED`.

So the tests drive the real activity over the real adapter and assert on the two places the
fabricated URL was read as success:

| Assertion | Consumer it protects |
|---|---|
| persisted finding is `DIAGNOSED`, `proposalUrl` null, `proposedAt` null | the row a human and the HITL queue read |
| `count { status == PROPOSED } == 0` | `findingsProposed` in the run report |
| neither adapter method returns any host string | #5897 defect 2 |

On `authz-policy-auditor` this is the sharper one: ADR-0167 makes every finding a human-triage
ticket and the charter requires `two_person_review`, so a finding recorded `PROPOSED` with nobody
holding it is an unfiled authorization-policy defect reported as filed.

## Falsification (verify by effect)

Restored the fabricated-URL branch in each adapter and re-ran:

| Module | Result |
|---|---|
| `openbank-docs-truth-agent` | **4 of 4 tests FAILED** |
| `openbank-authz-policy-auditor` | **4 of 4 tests FAILED** |

Restored, both green again. No remaining live string in either service names `openbank/openbank` —
the only surviving occurrences are the KDoc paragraphs describing the removed defect.

## Verification

Per module, serialized, `--rerun-tasks`, exit code read from a redirected log (not a pipeline):

| Module | ktlintCheck | detekt | test | quarkusBuild |
|---|---|---|---|---|
| `openbank-docs-truth-agent` | pass | pass | **61 tests, 0 failures, 0 errors, 0 skipped** | pass |
| `openbank-authz-policy-auditor` | pass | pass | **36 tests, 0 failures, 0 errors, 0 skipped** | pass |

`quarkusBuild` included because ktlint + unit tests do not validate CDI wiring, and both adapters
are `@ApplicationScoped` beans whose port signature changed.

Closes #5897. Refs #3900, PR #5890.

### DIFF
--- a/docs/agents/authz-policy-auditor.md
+++ b/docs/agents/authz-policy-auditor.md
@@ -73,10 +73,19 @@ vs. `tool_tiers` drift) proactively covering a related, not-yet-incident-causing
   per-service `*-opa-bundle.yaml` copies `gen-*-opa-bundle.sh` mechanically derives from them — a
   regression at the source is the primary defense; the generated copies are supposed to be
   mechanically regenerated, not independently authored.
-- The `LlmDiagnosisPort` and `GitHubProposalPort` adapters are stubs pending the shared LiteLLM
-  gateway and GitHub App installation-token wiring, the same bootstrap state every sibling agent
-  shipped with. `proposeFixDiff` returns `null` unconditionally BY DESIGN here (not just pending
-  integration) — this agent never proposes a fix diff for a security-policy defect.
+- The `LlmDiagnosisPort` adapter is a stub pending the shared LiteLLM gateway. `proposeFixDiff`
+  returns `null` unconditionally BY DESIGN here (not just pending integration) — this agent never
+  proposes a fix diff for a security-policy defect.
+- **`GitHubProposalPort` is unwired and REFUSES — no finding of this agent reaches GitHub today**
+  (#5897). Both methods return `null`, and `DiagnoseAndProposeActivityImpl` then leaves the finding
+  `DIAGNOSED` with a null `proposalUrl`: it is never counted in a run's `findingsProposed` and never
+  presented as awaiting a human. It previously returned a fabricated
+  `https://github.com/openbank/openbank/issues/pending-authz-policy-auditor-<id>` URL and moved the
+  finding to `PROPOSED` — which on this agent meant an authorization-policy defect nobody had filed
+  being reported as filed and awaiting the `two_person_review` its charter requires. `openProposalPr`
+  refuses **permanently** (ADR-0167: never a fix PR on an authorization policy); `openTicket`
+  refuses because no `github-token` config exists in this service. Follows
+  `openbank-mcp-service`'s `UnwiredProposalPort` (#3900).
 - `repo-root` (`AUTHZ_POLICY_AUDITOR_REPO_ROOT`) must point at a mounted, up-to-date checkout of
   `main` for the `PolicyScanPort` checks to be meaningful — the deployment-side checkout-mount
   wiring (a sidecar or init-container `git pull`) is tracked separately and not yet part of this
--- a/docs/agents/docs-truth-agent.md
+++ b/docs/agents/docs-truth-agent.md
@@ -77,11 +77,18 @@ closing a gap none of this repo's other control-plane agents cover.
   wrong-attribution case but is still a proximity-based text scan, not a structural YAML parse — a
   `rules.yaml` restructuring that moves `enforced:` more than `SEARCH_WINDOW` (15) lines from its
   `gate:` key would silently stop matching.
-- The `LlmDiagnosisPort` and `GitHubProposalPort` adapters are stubs pending the shared LiteLLM
-  gateway and GitHub App installation-token wiring, the same bootstrap state finops-agent/devops-
-  agent/control-liveness-sentinel/governance-auditor/release-steward shipped with. Until that
-  lands, a finding produces a tracking ticket with a placeholder summary rather than an
-  LLM-drafted root-cause note, and `proposeFixDiff` never yet returns a real diff.
+- The `LlmDiagnosisPort` adapter is a stub pending the shared LiteLLM gateway, so `proposeFixDiff`
+  never yet returns a real diff.
+- **`GitHubProposalPort` is unwired and REFUSES — no finding of this agent reaches GitHub today**
+  (#5897). Both methods return `null`, and `DiagnoseAndProposeActivityImpl` then leaves the finding
+  `DIAGNOSED` with a null `proposalUrl`: it is never counted in a run's `findingsProposed` and never
+  presented as awaiting a human. It previously returned a fabricated
+  `https://github.com/openbank/openbank/pulls/pending-docs-truth-agent-<id>` URL and moved the
+  finding to `PROPOSED` — a no-op sharing its shape with a real result, on a host that is not even
+  this repository. This follows `openbank-mcp-service`'s `UnwiredProposalPort` (#3900).
+  The charter already grants `tier: write_proposal` on `github-pr`, so wiring it needs no charter
+  change — but unlike `flaky-test-hunter` this service has no `github-token` config, so there is no
+  token path to fail closed on yet. `flaky-test-hunter`'s adapter is the template.
 - `repo-root` (`DOCS_TRUTH_AGENT_REPO_ROOT`) must point at a mounted, up-to-date checkout of `main`
   for the `RepoScanPort`/`GovernanceRulesPort` checks to be meaningful — the deployment-side
   checkout-mount wiring (a sidecar or init-container `git pull`) is tracked separately and not yet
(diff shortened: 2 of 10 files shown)


JiRaska:

## The refusal path is a genuinely distinct outcome — this part is right

I checked the specific thing that matters: that the refusal is not success-shaped.

- The port's return type carries the refusal — `AuthzPolicyPorts.kt:42,47` and `DocsTruthPorts.kt` both move `String` → `String?`, so a caller cannot ignore it by accident. The KDoc at `AuthzPolicyPorts.kt:39-45` states the invariant plainly: `null` is the only way to say "nothing was created", and there is deliberately no placeholder-URL return because a well-formed string is indistinguishable from a delivered proposal to every consumer.
- `DiagnoseAndProposeActivityImpl.kt:58-82` leaves the finding **`DIAGNOSED` with a null `proposalUrl` and a null `proposedAt`**, rather than `PROPOSED`. That is a different enum value, not a flag shared with success — which is what keeps it out of the run's `findingsProposed` count and out of the HITL queue.
- The refusal logs at `warnf`, not `infof` (`GitHubProposalAdapter.kt`), and the message says "NOTHING was created. Do not report this as a delivered proposal."
- The new `RefusedProposalIsNotRecordedAsProposedTest` covers all four properties, including `a refused proposal contributes nothing to the reported findingsProposed count` and `no returned value names a repository host at all`. The `runBlocking` case at line 120 is correctly written `(): Unit = runBlocking {` — it will not be silently dropped by JUnit5.

The severity framing is also correct: on `authz-policy-auditor` specifically, a fabricated URL meant an authorization-policy defect nobody had filed being reported as filed and awaiting the `two_person_review` its charter requires.

## Finding: the KDoc's factual claims about `flaky-test-hunter` are false against `main`

Both new adapters point at `flaky-test-hunter` as the wired reference:

- *"unlike `flaky-test-hunter`, this service has no `github-token` config to fail closed on"*
- *"`flaky-test-hunter`'s adapter is the template."*

On live `origin/main`, `openbank-flaky-test-hunter/.../GitHubProposalAdapter.kt` is the **identical unfixed stub**, with no token path at all:

- `:31` `override suspend fun openProposalPr(...): String` → `:40` returns `https://github.com/openbank/openbank/pulls/pending-flaky-test-hunter-<id>`
- `:45` `override suspend fun openTicket(...): String` → `:53` returns `https://github.com/openbank/openbank/issues/pending-flaky-test-hunter-<id>`

Same fabricated host, same `pending-` prefix, same non-null `String`, same `infof` log. It is not a template and it does not fail closed. Please correct both KDoc blocks — a comment that sends the next author to a file that does not do what the comment says is the thing this PR is otherwise fixing. If a wired template is wanted, `openbank-devops-agent`'s `RemediationProposalAdapter.kt:53` is the only adapter on `main` that already returns `String?`.

## Finding: the fabricated-URL defect is fleet-wide, and this fixes 2 of 6

The same `https://github.com/openbank/openbank/…/pending-<agent>-<id>` return exists on `main` in **six** services. This PR fixes two:

| Service | Status after this PR |
|---|---|
| `openbank-authz-policy-auditor` | fixed here |
| `openbank-docs-truth-agent` | fixed here |
| `openbank-finops-agent` | **still fabricating** (`GitHubProposalAdapter.kt:36`) |
| `openbank-flaky-test-hunter` | **still fabricating** (`:40`, `:53`) |
| `openbank-governance-auditor` | **still fabricating** (`:38`, `:49`) |
| `openbank-release-steward` | **still fabricating** (`:39`, `:50`) |

Not a blocker for this PR — a two-service fix is a fine slice — but the remaining four should be tracked, because the consumer side makes them user-visible: `openbank-admin-ui/src/components/agent/AgentInsightsPanel.tsx:192` renders `proposalUrl` as an `<a href>`, so every fabricated URL is a clickable link in the operator UI pointing at a repository that is not this one (`openbank/openbank`, never `JiRaska/open-bank-oss`). An operator clicking through gets a 404, which is the *good* case; the bad case is nobody clicks and the finding is counted as filed.

Worth stating in the PR body which issue tracks the other four, so this does not become a "filed separately" that was never filed.

## Everything else checked, nothing found

- **Nothing else consumes the fabricated value.** I traced `proposalUrl` across the fleet: for these two agents it flows only to the finding record and the admin-ui panel; there is no gate, metric, or downstream service that reads it and would now break on `null`. The persistence column is already nullable (`proposal_url VARCHAR(512)`, and `var proposalUrl: String? = null` in the entities), so `null` needs no migration.
- **Release scope**: `fix(docs-truth-agent)` and `fix(authz-policy-auditor)` are both releasing types touching `src/main` in their own packages — both will cut a patch release, which is correct for a behaviour change.
- Charter/`agents.yaml` unchanged, so none of the OPA-bundle or `eu-ai-act.md` derived tails are implicated.
- Scope creep: none.

Issue #5913, opened 2026-08-20
Fuzzing past authentication: 12 of 17 rows fixed — 3 unconditional 500s remain (lending HQL, ledger journals NPE, party directory lookup NPE); all Class B fixed centrally in libs-runtime


JiRaska:
## What produced this

The two-pass fuzz lane (#5884) run across the full money-path fleet, with the OPA fix in: [32364265056](https://github.com/JiRaska/open-bank-oss/actions/runs/32364265056).

Before this lane could reach past authentication, **4 of 306 operations were ever exercised** (#5769). This is the first run that fuzzed the other 302.

```
24 findings across 11 services   —   0 of them visible to the auth-on pass
```

Causes read from each service's own `-boot-authz-off.log`, not inferred from the generic `INTERNAL_ERROR` body.

## Class A — unconditional 500s. These need no malformed input at all.

| service | endpoint | cause |
|---|---|---|
| **lending** | `GET /api/v1/lending/compliance-packs/proposals/pending` | `SemanticException: Could not interpret path expression 'proposed_at'` — HQL names the **column**, Hibernate wants the entity property (`proposedAt`) |
| **party** | `PUT /api/v1/parties/{id}/payees` | `SQLGrammarException: relation "party_payees_seq" does not exist` |
| **ledger** | `POST /api/v1/journals` | `NullPointerException: Cannot invoke "…PostJournalLine…"` — a null element in the lines array |
| **party** | `POST /api/v1/parties/directory/lookup` | `NullPointerException: Parameter specified as non-null is null` |

The lending one is a plain `GET` with **no parameters**: it has answered 500 on every call since it shipped. Same shape as the consent `SuppressionEntity` defect fixed today (#5711) — a name that only exists in one of the two worlds, invisible to any test that mocks the repository.

`party_payees_seq` is the pooled-sequence trap CLAUDE.md documents verbatim: Panache allocates ids from `<table>_seq` while `BIGSERIAL` only creates `<table>_id_seq`.

## Class B — 500 where 400 belongs (input reaches Postgres unvalidated)

| cause | services |
|---|---|
| `invalid byte sequence for encoding` (NUL bytes in a string) | dispute, interest, sdd, transaction, fx |
| `value too long for type character varying` | balance, sanctions |
| `date out of range` (`?month=2&year=6574608`) | interest |
| `ConstraintViolationException` surfacing as 500 | dispute, sdd |
[...]

PR #5948, opened 2026-08-20
fix(lending): order pending pack proposals by the entity property, not the column

JiRaska:
fix(lending): order pending pack proposals by the entity property, not the column

`GET /api/v1/lending/compliance-packs/proposals/pending` takes no parameters and
answered 500 on every call from the day it shipped. The HQL read
`order by proposed_at` — the COLUMN name — where Hibernate wants the entity
property `proposedAt`, so the query threw
`SemanticException: Could not interpret path expression 'proposed_at'` before a
single row was read.

Found by the two-pass fuzz lane once it could get past authentication (#5913).
It needed no malformed input at all; it was invisible to the auth-on pass
because a 403 fired before the handler ran.

Same shape as the consent `SuppressionEntity` defect (#5711): a name that exists
in exactly one of the two worlds, so every test that mocks the repository agrees
with the broken code.

`CompliancePackPendingListIT` drives the real endpoint against a real schema and
asserts the STATUS, not the absence of an exception — the endpoint returned a
well-formed `INTERNAL_ERROR` body, so "it did not crash" is precisely the oracle
that missed this. Falsified: the test is red (500) on the unfixed repository and
green (200) with the one-word change.

Refs #5913

### DIFF
--- a/openbank-lending-service/src/main/kotlin/com/openbank/lending/application/port/out/CompliancePackActivationRepository.kt
+++ b/openbank-lending-service/src/main/kotlin/com/openbank/lending/application/port/out/CompliancePackActivationRepository.kt
@@ -87,7 +87,11 @@ class JpaCompliancePackActivationRepository :
 
     @WithSession
     override fun findByState(state: ProposalState): Uni<List<CompliancePackActivationEntity>> =
-        list("state = ?1 order by proposed_at", state)
+        // `proposedAt`, not `proposed_at`: this is HQL, so the ORDER BY names the ENTITY PROPERTY.
+        // The column name parses fine to the eye and throws `SemanticException: Could not interpret
+        // path expression` at runtime, which is how this endpoint answered 500 on every call from the
+        // day it shipped until the fuzz lane got past authentication (#5913).
+        list("state = ?1 order by proposedAt", state)
 
     @WithSession
     override fun findActivated(): Uni<List<CompliancePackActivationEntity>> = list(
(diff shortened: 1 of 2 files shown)


JiRaska:
I independently wrote the same fix (#6199, now closed as the later duplicate of this one). One piece of mine is additive and worth having here — offering it rather than keeping a competing PR open.

**The test asserts the status but not the ordering.** `order by` is the part of a query most easily wrong in a way that still returns rows: swap `proposedAt` for `updatedAt`, or reverse it, and `statusCode(200)` still passes. The defect this fixes happened to be a parse failure, so status was enough to catch *it* — but the assertion left behind does not protect the property the line exists for.

Insert newest-first, assert oldest-first comes back, and insertion order can no longer carry it:

```kotlin
val newer = insertProposal(version = 9101, proposedAt = OffsetDateTime.now())
val older = insertProposal(version = 9102, proposedAt = OffsetDateTime.now().minusDays(2))

val ids = body.getList<String>("id")
assertThat(ids.indexOf(older.toString()))
    .`as`("proposals must come back oldest-first — `order by proposedAt`")
    .isLessThan(ids.indexOf(newer.toString()))
```

Take it or leave it — the fix itself is identical and correct, and this PR is the one that should land.

For the record, both branches falsified it the same way: reverting `proposedAt` to `proposed_at` reproduces `SemanticException: Could not interpret path expression 'proposed_at'` in the test, which is the exact production exception from #5913.

PR #5951, opened 2026-08-20
fix(party): create the payee id sequence under the name Hibernate actually asks for

JiRaska:
fix(party): create the payee id sequence under the name Hibernate actually asks for

Two 500s the two-pass fuzz lane found in party-service once it could get past
authentication (#5913). Neither was visible to the auth-on pass — a 403 fires
before the handler runs.

1. `PUT /api/v1/parties/{id}/payees` answered 500 on EVERY call:
   `SQLGrammarException: relation "party_payees_seq" does not exist`.
   `V18__party_payees.sql` created the pooled id sequence as the QUOTED
   `"party_payees_SEQ"`. A quoted identifier keeps its case in Postgres; the name
   Hibernate emits is unquoted and folds to lowercase, so the two never met.
   `V6__fix_hibernate_sequences.sql` fixed exactly this for the service's other
   three sequences and `V18` reintroduced it. `V19` drops the quoted sequence and
   creates the lowercase one, with the rollback in the file.

   Dropping rather than renaming is safe here precisely because the endpoint has
   never worked: every insert failed, so the quoted sequence holds no allocated
   state and `party_payees` is empty.

2. `POST /api/v1/parties/directory/lookup` answered 500 for a null INSIDE
   `phoneHashes`. Kotlin's element non-nullability is erased past Jackson, so the
   null reached the handler and the first dereference threw
   `Parameter specified as non-null is null`. `List<String?>` makes the null
   representable so it can be rejected as `IllegalArgumentException`, which
   libs-runtime's shared mapper turns into 400 (no service-local mapper, #526).

WHAT I MEASURED RATHER THAN ASSUMED

The obvious neighbouring cases are already correct and I did not "fix" them: a
null at the top level of the body, and a null in a scalar field
(`{"name":null,...}`), both answer 400 today — the fleet-wide guards from #3038
cover them. I found that by running those cases against the UNFIXED build first,
and dropped the two tests asserting them, because a test that is green before and
after the change is decoration.

`PartyPayeeAndDirectoryIT` asserts the STATUS, not the absence of an exception —
both endpoints returned a well-formed `INTERNAL_ERROR` body, so "it did not
crash" is the oracle that missed them. Falsified: both tests are red (500, 500)
on the unfixed build and green (200, 400) after.

Refs #5913

### DIFF
--- a/openbank-party-service/src/main/kotlin/com/openbank/party/infrastructure/rest/PartyResource.kt
+++ b/openbank-party-service/src/main/kotlin/com/openbank/party/infrastructure/rest/PartyResource.kt
@@ -152,7 +152,7 @@ class PartyResource {
     @RolesAllowed("ROLE_API")
     @Operation(summary = "Match phone-number hashes against parties who opted into being discoverable")
     suspend fun lookupDirectory(req: DirectoryLookupRequest): Response =
-        Response.ok(mapOf("matches" to partyUseCase.lookupByPhoneHashes(req.phoneHashes))).build()
+        Response.ok(mapOf("matches" to partyUseCase.lookupByPhoneHashes(req.hashes()))).build()
 
     /** Turn this party's pay-to-phone findability on or off. Revocable at any time. */
     @PUT
@@ -756,7 +756,20 @@ private const val GDPR_EXPORT_SCOPE =
         "subject-access response is a tracked follow-up (ADR-0118 §6)."
 
 /** Address-book hashes to match. See PhoneDirectory for what the hashing does and does not buy. */
-data class DirectoryLookupRequest(val phoneHashes: List<String> = emptyList())
+/**
+ * `List<String?>`, deliberately. Kotlin's non-null element type is a COMPILE-TIME property and
+ * Jackson does not enforce it inside a collection, so `{"phoneHashes":["abc",null]}` produced a
+ * `List<String>` holding a null and the first dereference threw
+ * `Parameter specified as non-null is null` — a 500 for malformed input (#5913). Declaring the
+ * element nullable makes the null representable, so it can be REJECTED instead of exploding; the
+ * rejection lives in [DirectoryLookupRequest.hashes].
+ */
+data class DirectoryLookupRequest(val phoneHashes: List<String?> = emptyList()) {
+    /** @throws IllegalArgumentException mapped to 400 by libs-runtime's shared mapper (never a service-local one, #526). */
+    fun hashes(): List<String> = phoneHashes.map {
+        requireNotNull(it) { "phoneHashes must not contain null entries" }
+    }
+}
 
 data class DiscoverableRequest(val discoverable: Boolean = false)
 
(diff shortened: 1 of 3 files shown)


JiRaska:
**#5951 and #6467 both add a `V19` migration to `openbank-party-service`, and merging both breaks the service.**

```
5951  openbank-party-service/src/main/resources/db/migration/V19__fix_party_payees_sequence.sql
6467  openbank-party-service/src/main/resources/db/migration/V19__fix_quoted_sequence_names.sql
main  highest is V18__party_payees.sql
```

Different filenames, so git merges both without a conflict and nothing anywhere reports a problem —
then Flyway refuses to start with `Found more than one migration with version 19`. The failure mode
worth knowing: Quarkus cannot boot, so every `@QuarkusTest` in the module reports **SKIPPED**, and
the module reads as `N tests / M skipped / 0 failures`. The number people look at is the failure
count, so that scans as a pass.

Both fix the same root cause (a quoted `"x_SEQ"` keeps its case in Postgres while Hibernate's
`nextval('x_SEQ')` folds unquoted to lower case), and both cite #5913. **#6467 strictly dominates:**

| | #5951 | #6467 |
|---|---|---|
| `party_payees` sequence | yes | yes |
| `party_marketing_consent` sequence | **no** | yes |
| starting value | fresh at 1 | `max(id) + 50` |

The starting value is the part that actually matters. #5951 argues a fresh sequence is safe because
"the quoted sequence has never been read, so `party_payees` is empty". #6467 makes the opposite
assumption — `id` is `BIGSERIAL`, so a row inserted by any path that omitted the id got one from the
implicit `party_payees_id_seq`, and a sequence restarting at 1 would collide with it. I have not
established which is true of the live table, and that is the point: `COALESCE(MAX(id), 0) + 50`
is correct under **both** readings and degrades to 50 on an empty table, whereas restarting at 1 is
correct only under one. #6467 also carries the `GRANT` and covers the second sequence, which #5951
misses entirely — that one fails identically on the first marketing-consent grant the projection
consumer records.

**Suggested resolution:** close #5951 in favour of #6467, rather than renumbering one to V20. Two
migrations fixing the same sequence in sequence would leave V19 dropping and recreating
`party_payees_seq` only for V20 to do it again with a different start value — the end state depends
on apply order, which is exactly what you do not want in a money-path schema.

Renumbering is only safe while a migration has never been applied anywhere; after that, Flyway
checksums the file and any edit is a startup failure. Neither V19 appears to have been applied yet,
so the option is open — I just do not think it is the right one here.

I have not closed or changed either PR. Both are money-path and this is a correctness call, not a
duplicate cleanup.

JiRaska:
Closing in favour of #6467 (OPEN, CLEAN as of this comment), per the analysis above.

Both PRs added a `V19` to `openbank-party-service`, which merges without a conflict and then stops
the service booting. #6467 is the superset: it fixes `party_marketing_consent` as well as
`party_payees`, carries the `GRANT`, and seeds each sequence from `COALESCE(MAX(id), 0) + 50` —
correct whether or not the table already holds rows, where restarting at 1 is correct only if it
does not.

Nothing in this PR is lost by closing it; every change it makes, #6467 also makes.

Worth keeping from this branch's write-up, since #6467 words it differently: V6 already fixed this
exact quoting mistake for the service's other three sequences, and V16 and V18 reintroduced it. That
is three occurrences of one defect in one service's migration history, which is a stronger argument
for a CI guard on quoted `"x_SEQ"` names than either PR makes on its own. Worth an issue if there
isn't one.

PR #5952, opened 2026-08-20
fix(ledger): reject a null journal line as 400 instead of dereferencing it

JiRaska:
fix(ledger): reject a null journal line as 400 instead of dereferencing it

`POST /api/v1/journals` answered 500 for `"lines": [null]` — found by the
two-pass fuzz lane once it could get past authentication (#5913). Not visible to
the auth-on pass, because a 403 fires before the handler runs.

Kotlin's element non-nullability is a COMPILE-TIME property and is erased past
Jackson, so the null survived deserialization and
`request.lines.map { it.toCommand() }` threw NPE, which the generic mapper
rendered as 500. Declaring `List<PostJournalLineRequest?>` makes the null
representable, so it can be rejected as `IllegalArgumentException` — mapped to
400 by libs-runtime's shared mapper, never a service-local one (#526).

This is the collection-element form of the null-body class fixed fleet-wide in
#3038. Those guards read Kotlin's own nullability off the handler's parameter,
which is the right source of truth and cannot see inside a collection: the
element type is not on the parameter. A null body and a null scalar field both
answer 400 today — I ran both against the unfixed build to check, and did not
add tests for them, because a test that is green before and after the change is
decoration.

`JournalMalformedLinesIT` asserts the STATUS: this endpoint returned a
well-formed `INTERNAL_ERROR` body, so "it did not crash" is exactly the oracle
that missed it. Falsified: red (500) on the unfixed handler, green (400) after.
Full module suite 306 tests, 0 failures, 1 skipped (the broker-gated pact
verification, as on main).

Refs #5913

### DIFF
--- a/docs/threat-models/openbank-ledger-service.md
+++ b/docs/threat-models/openbank-ledger-service.md
@@ -238,6 +238,21 @@ set) apply equally to the new `ledger.approval.decide` action.
 
 ## 8. Change log
 
+- **2026-08-20** — `POST /api/v1/journals` answered 500 for `"lines": [null]` (issue #5913, found
+  by the two-pass fuzz lane once it could reach past authentication). **No new trust boundary:**
+  the edge, route, `@RolesAllowed(Roles.OPERATOR)` gate and `@Authorize(action = "ledger.create")`
+  decision are all unchanged, and no input that was rejected before is accepted now — the change is
+  strictly the opposite direction, a request that used to reach the handler and throw is now
+  refused. **Denial of service (the reason this is in the threat model at all):** an unhandled NPE
+  on the posting path is an unauthenticated-shaped fault an OPERATOR-authenticated caller can
+  trigger at will with a one-byte payload, and each occurrence costs a stack unwind plus a
+  generic-mapper render; declaring the element nullable moves the rejection to deserialization-time
+  argument validation. **Information disclosure:** none either way — `GenericExceptionMapper`
+  already rendered an opaque `INTERNAL_ERROR` body, so the NPE never leaked a stack trace; that is
+  also why no monitoring saw it. **Repudiation:** unchanged; a rejected request posts nothing, so
+  there is no journal, no outbox row and no audit event to reconcile. Rollback: revert the commit —
+  the change is confined to the request DTO and carries no migration and no persisted state.
+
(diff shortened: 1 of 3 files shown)


JiRaska:

**Premise holds** — `LedgerResource` on `main` still does `lines = request.lines.map { it.toCommand() }` with no null guard, so the defect this PR fixes is still live.

**But this PR is `CONFLICTING` / `DIRTY` against current `main`** (confirmed on a second `gh pr list` read, since `mergeable` is lazily computed and the first read returns `UNKNOWN`).

That matters more than it looks: **a conflicting PR gets zero CI runs.** The green checks shown here are about a base from 2026-08-22 and verify nothing about the merged result. Nothing on this branch is currently verified.

**Needed:** rebase onto `origin/main`, then let CI actually run. `gh run rerun` will not help — it replays the same stale merge commit.

Note this is also money-path (`openbank-ledger-service` is in `rules.yaml: money_path_services`), so it needs 2 approvals plus the threat-model update — which this PR does include (`docs/threat-models/openbank-ledger-service.md`).

*Triage only — I did not rebase or otherwise touch this branch.*

JiRaska:
Closing as a duplicate of #6486 — both fix #5913, the null element in a journal's `lines` answering
500 instead of 400.

The two carry the same root fix in the same file, differing only in style:

- **this PR** extracts a `requiredLines()` helper on the request type;
- **#6486** uses `mapIndexed` + `requireNotNull(line) { "lines[$index] must not be null" }` inline,
  so the rejection names the offending index.

#6486 is the more complete of the two and the one still merge-ready:

| | #5952 (this) | #6486 |
|---|---|---|
| `LedgerResource.kt` fix | yes | yes |
| threat model updated | yes | yes |
| integration test | `JournalMalformedLinesIT` | `PostJournalMalformedBodyIT` |
| **`openapi.yaml` documents the new `400`** | **no** | yes (`info.version` 1.16.0 → 1.17.0) |
| mergeable state | CONFLICTING | CLEAN |

The spec gap is the deciding one rather than a preference: adding a `400` response is an API change,
and the repo's own rule is that an API change updates `openapi.yaml`. #6486's bump to 1.17.0 was also
checked against live `main` (still at 1.16.0), so it is not one of the version collisions currently
affecting other queued spec PRs.

No work is lost by closing this — reopen if the helper-based shape is preferred, in which case
#6486 should be closed instead and the spec change carried over.

Issue #5962, opened 2026-08-20
openapi enum vs domain: 17 baseline entries across 13 services remain (was 27/16); 5 of them are annotated NOT drift


JiRaska:
`oasdiff` compares an `openapi.yaml` to its own **previous version**, never to the implementation
behind it, so a spec enum that was hand-written once and never re-checked against its Kotlin type
drifts silently and forever. Nothing in CI has ever watched that axis.

#5895 found it in `openbank-kyc-service`. It is not unique. A probe added in that PR
(`.github/scripts/check-openapi-enum-vs-domain.py`, pairs a spec enum with the Kotlin enum it
serves by value overlap) reports **27 further drifts across 16 services** once kyc's two are fixed.
They are baselined in `BASELINE` in that script so a NEW drift fails; this issue owns the tail.

## The two shapes, and why one is worse

**A value the code does not have** — a generated client can send it, and the service answers 400
(`valueOf` throws) or worse. These are live defects, not documentation debt:

| service | domain enum | spec advertises, code lacks |
|---|---|---|
| openbank-account-service | `AccountStatus` | `PENDING` |
| openbank-clearing-service | `PaymentRail` | `CZ_DOMESTIC` |
| openbank-consent-service | `ConsentStatus` | `PENDING` |
| openbank-lending-service | `CollateralStatus` | `EXECUTED` |
| openbank-party-service | `DocumentType` | `DRIVING_LICENSE`, `ID_CARD` |
| openbank-party-service | `KycStatus` | `PENDING`, `VERIFIED` |
| openbank-pid-service | `PartyType` | `INDIVIDUAL` |
| openbank-pid-service | `PartyStatus` | `CLOSED` |
| openbank-pid-service | `Status` | `REJECTED`, `VERIFIED` |
| openbank-sca-service | `ScaStatus` | `VERIFIED` |
| openbank-sepa-instant | `SctInstStatus` | `ACCEPTED`, `SUBMITTED` |
| openbank-sepa-payment | `SepaPaymentStatus` | `RECALLED`, `PENDING` |

Note the recurring `DRIVING_LICENSE`/`DRIVING_LICENCE` and `VERIFIED`/`COMPLETED` pairs — the same
misspelling class as kyc's `PEP` vs `PEP_SCREENING`.

**A value the code has and the spec omits** — a real state no consumer can be written against.
[...]

PR #5966, opened 2026-08-20
fix(kyc): reconcile the published check enums with the domain types they serve

### LINKED ISSUE #5895: kyc openapi advertises a check-type enum that does not match the domain CheckType — SOURCE_OF_FUNDS does not exist in code, ADDRESS is missing from the spec
`openbank-kyc-service`'s published API advertises a check-type enum that does not match the domain
type it serves. Verified on `origin/main`:

```
openapi.yaml:120   enum: [IDENTITY, SANCTIONS, PEP, ADVERSE_MEDIA, SOURCE_OF_FUNDS]
KycCase.kt         enum class CheckType { IDENTITY, ADDRESS, PEP_SCREENING, SANCTIONS_SCREENING, ADVERSE_MEDIA }
```

Three of five names differ (`SANCTIONS` vs `SANCTIONS_SCREENING`, `PEP` vs `PEP_SCREENING`), the spec
omits `ADDRESS`, and **`SOURCE_OF_FUNDS` does not exist in the code at all**.

## Why it matters

The spec is the contract consumers generate clients from. A client generated against it can send
`SOURCE_OF_FUNDS` — a value the service has no enum member for — and can never send `ADDRESS`, which
is a real check the service creates. `KycService.createCase` produces IDENTITY / ADDRESS /
PEP_SCREENING / SANCTIONS_SCREENING today, so the mismatch is not theoretical for two of the four.

This is the ADR-0048 axis that no gate watches: the spec's *shape* is versioned and diffed, but
nothing checks that an enum in the spec corresponds to the enum in the code behind it. `oasdiff`
compares a spec to its own previous version, not to the implementation.

## What would close this

Reconcile the two, deciding per name which side is right — the domain names look canonical
(`*_SCREENING` matches the fleet convention and `ADDRESS` is a real check), so most likely the spec
is wrong on four of the five. That is a **breaking** contract change if any consumer relies on the
published names, so it needs the ADR-0048 classification (`oasdiff`) and a major bump if so, not a
quiet edit.

Worth checking in the same pass whether other services have the same drift — a spec enum that was
hand-written once and never re-checked against its domain type is not a defect unique to kyc, and
the same probe (compare each spec enum to the Kotlin enum it serves) would answer it fleet-wide.

Found while building the adverse-media port in PR #5893, which deliberately did not touch the spec.

Refs #4459, PR #5893, ADR-0048.

JiRaska:
Closes #5895. Refs #5962, ADR-0048.

## What was wrong

`PUT /api/v1/kyc/cases/{id}/checks/{checkType}` could not be called successfully by any client
generated from the published spec, and had not been callable since the file was written. Three
independent falsehoods in one endpoint:

|---|---|---|
| `checkType` enum `[IDENTITY, SANCTIONS, PEP, ADVERSE_MEDIA, SOURCE_OF_FUNDS]` | `CheckType { IDENTITY, ADDRESS, PEP_SCREENING, SANCTIONS_SCREENING, ADVERSE_MEDIA }` | `SANCTIONS`/`PEP`/`SOURCE_OF_FUNDS` → `valueOf` throws → 400. `ADDRESS`, created on every case, unpublished |
| body `required: [result]`, `result` carries the status enum, plus a `notes` property | `UpdateCheckRequest(val status: String, val result: String?)` | the required field is undeclared, `notes` does not exist |
| body enum `[PASS, FAIL, PENDING, MANUAL_REVIEW]` | `CheckStatus { PENDING, PASSED, FAILED, MANUAL_REVIEW }` | `PASS`/`FAIL` → 400; `PASSED`/`FAILED`, the only accepted spellings, unpublished |

The issue reported the first row; the other two are the same defect in the same request and are
fixed here too.

## Which side is right — per name

The domain, on all of them, and this is measured rather than assumed:

- `openbank-customer-edge/openapi.yaml:1329` already republishes the vocabulary verbatim as
  `[IDENTITY, ADDRESS, PEP_SCREENING, SANCTIONS_SCREENING, ADVERSE_MEDIA]`. A downstream spec
  agreeing with the code settles which spelling is canonical without a judgement call.
- `SOURCE_OF_FUNDS` was never a real check that got renamed — `git log -S` over the service finds
  it only in the openapi.yaml, never in any Kotlin source, at any revision.
- `ADDRESS`, `PEP_SCREENING`, `SANCTIONS_SCREENING` are what `KycService.createCase` writes on
  every case today, and what is persisted in `checks_json`.

- **admin-ui** reads only — `kyc/page.tsx` and `parties/[id]/page.tsx` render `ch.checkType`, never
  post it.
- **pacts** — `openbank-onboarding-service-openbank-kyc-service.json` is message-only
  (`KYC_CASE_APPROVED`); no HTTP interaction touches this route.
- **fleet grep** for the literals finds only kyc itself and the customer-edge spec above.
- No DB constraint pins either spelling: `checks_json` is `TEXT`, so nothing stored has to change.

## Classification: breaking per oasdiff, `correction` per the gate

`oasdiff breaking` reports **8 ERR-level changes** (3 `request-parameter-enum-value-removed`,
4 `request-property-enum-value-removed`, 1 `new-required-request-property`), so the naive reading
demands a MAJOR — which under ADR-0048 D2 means moving every path to `/api/v2`.

That is the case `check-api-contract.py`'s `correction` class exists for, and it applies here
verbatim: this is a **spec-only** PR, the running server is byte-identical, and the clients that
"break" are ones generated from a document that never described this server. Run against the real
merge-base with the pinned oasdiff:

```
::notice::api-contract gate: openbank-kyc-service: spec-only PR — reclassifying 8 breaking
document change(s) as a CORRECTION. ... Requiring MINOR instead.
api-contract gate: openbank-kyc-service: correction change, info.version 1.6.0 -> 1.7.0 — OK
```

`info.version` 1.6.0 → **1.7.0**, read off live `origin/main` immediately before bumping. Major
stays 1, so the D2 invariant (`major(info.version) == openbank.api.version == max URL major`)
holds and no URL moves.

## The control that was missing

`oasdiff` compares a spec to its own previous version and **never** to the implementation, so
nothing in CI has ever watched this axis — which is why a spec wrong from its first commit stayed
green for its whole life. New gate `openapi-enum-domain-drift` (`api` group, enforced):

- pairs each spec enum with the Kotlin enum it serves **by value overlap**, since nothing declares
  the relationship and name-matching would find almost none of them;
- an unpaired spec enum is not a finding (many are free-form vocabularies with no Kotlin type);
- ratcheted: today's 27 further drifts across 16 services are baselined under #5962, and a
  **baseline entry that no longer drifts also fails**, so the list cannot rot into furniture.

Note where the test lives. It is deliberately **not** in `openbank-kyc-service/` — the `correction`
reclassification is withdrawn the moment a PR touches anything else in the service, so a
service-local drift test would make the gate demand a MAJOR bump that D2 simultaneously forbids.

### Falsified three ways, against the real repo

| | expected | got |
|---|---|---|
| reintroduce the #5895 mismatch | red | `exit=1`, one `::error` naming `CheckType` |
| make a baselined drift agree, keep its entry | red | `exit=1`, "baseline entry ... no longer drifts" |
| clean tree | green | `exit=0`, 27 baselined, 0 new |

Plus an 18-case `--self-test`. That self-test earned its keep before it ever ran in CI: the first
cut used a 0.5 overlap threshold and **did not flag the #5895 drift the gate exists for** (it
shares exactly 2 of 5 values), and a second cut mined KDoc prose out of Kotlin enum bodies and
invented constants like `ADR`, `MUST` and `D1` across a dozen services. Both calibrations are now
pinned by cases that go red if they are undone.

## Beyond kyc

The same probe found **27 further drifts across 16 services** — filed as **#5962** with the full
table. Twelve advertise a value the code does not have, which a generated client will send and the
service will reject: `clearing/CZ_DOMESTIC`, `consent/PENDING`, `party/DRIVING_LICENSE`+`ID_CARD`,
`sca/VERIFIED`, `sepa-instant/ACCEPTED`+`SUBMITTED`, `sepa-payment/RECALLED`, and more. Several
are money-path.

## Verification

- `:openbank-kyc-service:ktlintCheck :openbank-kyc-service:detekt` — BUILD SUCCESSFUL, exit 0
  (`runKtlintCheckOverMainSourceSet` ran, not `UP-TO-DATE`).
- `:openbank-kyc-service:test` — exit 0, **84 tests / 0 failures / 0 errors / 1 skipped**. The one
  skip is `KycEventPactBrokerProviderVerificationTest`, broker-gated by
  `@EnabledIfSystemProperty(pactbroker.url)` as designed, not a silently-disabled suite.
- `run-gates.py --all` — the new gate passes and reports `SUBJECTS=147`; both meta-gates
  (`gate-lifecycle-metadata`, `gate-subject-floor`) initially rejected it for a missing
  `rationale:`/`review_after:`/`min_subjects:` and now pass. The two remaining local failures are
  environmental and unrelated: `loki-rule-load-test` (`BASE_REF: unbound variable`) and
  `api-contract-gate` (its install step needs `sudo`) — the latter was instead run directly
  against a locally installed oasdiff, which is the output quoted above.

Exit codes were read from `$?` on a redirected file, never off a pipe.

### DIFF
--- a/.github/gates/gates.yaml
+++ b/.github/gates/gates.yaml
@@ -2186,6 +2186,42 @@ gates:
     run: |
       python3 .github/scripts/check-pact-provider-replay.py
 
+  # openapi enum vs the Kotlin enum behind it (issue #5895 / #5962, enforced).
+  #
+  # The axis the api-contract gate below structurally CANNOT see: oasdiff compares a spec to its
+  # own previous version, never to the implementation. So a spec enum hand-written once and never
+  # re-checked drifts forever and every gate stays green. openbank-kyc-service published
+  # `[IDENTITY, SANCTIONS, PEP, ADVERSE_MEDIA, SOURCE_OF_FUNDS]` against a domain
+  # `CheckType { IDENTITY, ADDRESS, PEP_SCREENING, SANCTIONS_SCREENING, ADVERSE_MEDIA }` from the
+  # day the file was written — three names misspelled, one real check unpublished, and one value
+  # that has never existed in the code. Its own PUT body was fiction in the same way.
+  #
+  # Pairing is by VALUE OVERLAP, not by name: nothing declares which Kotlin enum a spec enum
+  # serves, and matching on names would find almost none of them. An unpaired spec enum is not a
+  # finding (many are free-form vocabularies with no Kotlin type).
+  #
+  # RATCHET: the 27 drifts across 16 services that exist today are a BASELINE constant in the
+  # script (owned by #5962), so this is enforced from day one and number 28 is red at PR time. A
+  # baseline entry that no longer drifts also FAILS, so the list cannot rot into furniture — same
+  # contract as check-pact-provider-replay.py and check-adr-partial-followup.py above.
+  - id: openapi-enum-domain-drift
+    name: "openapi enum matches the domain enum it serves (issue #5962, enforced)"
+    group: api
+    mode: enforced
+    rationale: >
+      oasdiff compares a spec to its own previous version and never to the implementation, so a
+      spec enum hand-written once and never re-checked drifts forever with every gate green.
+      openbank-kyc-service published three misspelled check names, omitted a real check, and
+      advertised a value (SOURCE_OF_FUNDS) that has never existed in the code — from the day the
+      file was written until #5895, and 27 more pairings across 16 services still do (#5962).
+    review_after: '2027-02-20'
+    min_subjects: 100
+    selftest: |
+      python3 .github/scripts/check-openapi-enum-vs-domain.py --self-test
+    selftest_expect: pass
+    run: |
+      python3 .github/scripts/check-openapi-enum-vs-domain.py --enforce
+
   # ADR-0048 D5: api-contract gate — classify the API-contract bump from the
   # OpenAPI diff (oasdiff), independently of the release axis (version.txt):
   #   breaking  => info.version MAJOR (new URL major /api/v{N+1})
--- a/.github/scripts/check-openapi-enum-vs-domain.py
+++ b/.github/scripts/check-openapi-enum-vs-domain.py
@@ -1,0 +1,355 @@
+#!/usr/bin/env python3
+"""openapi-enum-domain-drift gate — does a spec enum still match the Kotlin enum behind it?
+
+The axis nothing else watches. `oasdiff` (the api-contract gate, ADR-0048) compares a spec to
+its own PREVIOUS version, never to the implementation, so a spec enum that was hand-written once
+and never re-checked drifts silently and forever. openbank-kyc-service published
+`[IDENTITY, SANCTIONS, PEP, ADVERSE_MEDIA, SOURCE_OF_FUNDS]` against a domain
+`CheckType { IDENTITY, ADDRESS, PEP_SCREENING, SANCTIONS_SCREENING, ADVERSE_MEDIA }` from the day
+the file was written: three names misspelled, one real check unpublished, and one value
+(`SOURCE_OF_FUNDS`) that has never existed in the code at all (#5895). Everything was green the
+whole time, because nothing compared the two documents.
+
+## How a spec enum is paired with a Kotlin enum
+
+There is no declaration linking them, so pairing is by VALUE OVERLAP, not by name — a spec enum
+is rarely named after the class, and matching on names would find almost nothing:
+
+  * collect every `enum:` value-set in the service's openapi.yaml,
+  * collect every `enum class X { A, B, C }` value-set in that service's src/main,
+  * pair a spec set with a Kotlin set when they share at least MIN_OVERLAP of the smaller set,
+  * report a pairing whose sets are NOT equal.
+
+Overlap, not equality, is what makes the check able to fire: an enum that agrees perfectly pairs
+and passes, and one that has drifted still overlaps enough to be recognised as the same concept.
+An unpaired spec enum is NOT a finding — plenty are free-form vocabularies with no Kotlin type
+(sort orders, filter keywords), and reporting those would bury the real ones.
+
+## Ratchet, not a wall
+
+Existing drift is baselined in BASELINE below with the issue that owns it; a NEW drift fails.
+A baseline entry that no longer drifts also fails, so the list cannot outlive its justification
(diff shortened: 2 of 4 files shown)


JiRaska:
## Review — #5966 (kyc + the gate)

Reviewed with `git diff origin/main...origin/fix/kyc-checktype-spec-drift`. **No change requested.**

**The spec really is the wrong side.** Verified independently of the PR body:

* `openbank-kyc-service/src/main/kotlin/com/openbank/kyc/domain/model/KycCase.kt:31` —
  `enum class CheckType { IDENTITY, ADDRESS, PEP_SCREENING, SANCTIONS_SCREENING, ADVERSE_MEDIA }`
* `KycCase.kt:32` — `enum class CheckStatus { PENDING, PASSED, FAILED, MANUAL_REVIEW }`
* `KycCheck` carries `val result: String?` — free text, not the enum, exactly as the PR says.

Both new spec enums are byte-equal to those. `SANCTIONS`, `PEP`, `SOURCE_OF_FUNDS`, `PASS` and
`FAIL` appear in no kyc Kotlin source, so this is a removal of values that were never served, not
a narrowing of a working contract.

**Version.** `git show origin/main:openbank-kyc-service/src/main/resources/openapi.yaml` reads
`version: 1.6.0` right now, so `1.7.0` is both correct and still free. No other open PR touches
this file (checked all 83 open PRs by file path), so no race.

**Pacts.** kyc's provider replay is `KycEventPactProviderVerificationTest` (`@PactFolder`), and it
is event/message-based — none of the removed literals appear in any committed pact. Nothing to
break.

**Scope.** The gate itself (`check-openapi-enum-vs-domain.py`, `gates.yaml`, the `CLAUDE.md`
bullet) is in this diff. That is the right call — the fix and the control that prevents the class
land together — but it does mean this PR is the one carrying the whole family's premise, so it is
worth a closer read of the script than the enum change.

One thing I could **not** verify from the repo alone: the claim that the gate is falsified three
ways ("reintroducing this defect turns it red, making a baselined drift agree turns it red as
stale, clean tree green"). The script has a `--self-test`, which is the right shape, but I did not
execute it. If that self-test does not contain a negative case for each of the three, it is the
kind of check that passes by being unable to fail.

---

JiRaska:

Issue #5902, opened 2026-08-20
sourceService values do not follow the documented convention and lending changed its own spelling mid-stream, splitting audit_entries into two producers


JiRaska:
#5256 defines `sourceService` as *"the module directory name minus the `openbank-` prefix"*. Nothing
enforces it, and at least one producer disagrees with itself over time.

Measured on `origin/main`:

```
openbank-lending-service   -> "lending"           should be "lending-service"
openbank-sepa-payment      -> "sepa-payment"      correct (directory has no -service suffix)
openbank-consent-service   -> "consent-service"   correct
```

## The part that already happened

PR #5399 upgraded six lending event types from TOPIC-derived attribution to EVENT-derived. Before
that upgrade `AuditConsumer` derived the producer from the topic; after it, the producer's own
`sourceService` value wins. Since lending self-reports `"lending"` while the topic-derived value
followed the module name, **those six types changed their `source_service` value mid-stream**, and
`audit_entries` now holds two spellings for one producer.

Any query that groups audit rows by `source_service` — which is the documented way to detect this
whole class of gap — silently splits lending into two producers, with the boundary at a merge date
rather than at anything meaningful.

PR #5894 deliberately preserved `"lending"` for the two types it fixed, rather than converging: a
convergence renames nine shipped types and would split the rows a **second** time. That is a
decision worth making once, deliberately.

## What would close this

1. Decide the canonical value for lending, and whether the existing rows get backfilled or the split
   is documented as a known boundary. Renaming again without backfilling makes it worse.
2. Then enforce the convention, so it cannot drift silently again. The check is cheap: for each
   producer, compare the `sourceService` literal it emits against its own module directory name.
   Note the probe trap that hid the original gap — `grep -c sourceService src/main` returns non-zero
   for all 21 producers, because it sees the field being *set somewhere*, not the value or the
   coverage. Any enforcement must read the emitted value at each write site, the way the #5256
   re-measurement did.

Found during the #5256 re-measurement (PRs #5888, #5894, #5898, #5901), which also established that
17 of 21 producers are complete and three were not.

Refs #5256, #5399, PR #5894.

JiRaska:
## Decision: keep `lending`, document the boundary

Option 3. No rename, no backfill, no `TopicAttribution` change. Recorded here so the issue reflects
it; it closes when #5967 merges.

**Why this rather than converging.** The rows lending has already written cannot be converged *by
[...]

PR #5967, opened 2026-08-20
ci(governance): gate sourceService against the module directory name

JiRaska:
## What this is

`source_service` is the column every audit and analytics query groups by — it is the documented way
to detect a producer that has stopped emitting, or one that never started. #5256 defines the value
as *the module directory name minus the `openbank-` prefix*. Nothing enforced it.

This PR **measures** the fleet, **enforces** the convention going forward, and records the one
producer that disagrees as a **declared, decided exception**.

> **The decision is taken (#5902): keep `lending`, document the boundary.** No rename, no backfill,
> no `TopicAttribution` change. Section 4 says why that is the cheap option and why it is permanent
> rather than deferred; section 5 is now the record of the decision, not a menu.

## 1. The measurement

Read the emitted value at each write site, not the field name. Full table, `origin/main`:

| module | emits | expected | |
|---|---|---|---|
| openbank-account-service | `account-service` | `account-service` | ok |
| openbank-balance-service | `balance-service` | `balance-service` | ok |
| openbank-card-issuance-service | `card-issuance-service` | `card-issuance-service` | ok |
| openbank-clearing-service | `clearing-service` | `clearing-service` | ok |
| openbank-consent-service | `consent-service` | `consent-service` | ok |
| openbank-customer-edge | `customer-edge` | `customer-edge` | ok |
| openbank-dispute-service | `dispute-service` | `dispute-service` | ok |
| openbank-document-service | `document-service` | `document-service` | ok |
| openbank-domestic-payment | `domestic-payment` | `domestic-payment` | ok |
| openbank-fx-service | `fx-service` | `fx-service` | ok |
| openbank-kyc-service | `kyc-service` | `kyc-service` | ok |
| **openbank-lending-service** | **`lending`** | **`lending-service`** | **MISMATCH, 9 write sites** |
| openbank-party-service | `party-service` | `party-service` | ok |
| openbank-sanctions-service | `sanctions-service` | `sanctions-service` | ok |
| openbank-sca-service | `sca-service` | `sca-service` | ok |
| openbank-security-scanner | `security-scanner` | `security-scanner` | ok |
| openbank-sepa-instant | `sepa-instant` | `sepa-instant` | ok |
| openbank-sepa-payment | `sepa-payment` | `sepa-payment` | ok |
| openbank-statement-service | `statement-service` | `statement-service` | ok |
| openbank-swift-service | `swift-service` | `swift-service` | ok |
| openbank-transaction-service | `transaction-service` | `transaction-service` | ok |

20 of 21 conform. Not producers, correctly not flagged: `openbank-audit-service` and
`openbank-analytics-sink` (consumers reading the field back), `openbank-libs-domain` (declares the
envelope), and `openbank-sepa-instant`'s publisher line, which forwards `event.sourceService`
rather than choosing a value.

## 2. What is already in `audit_entries` — three segments, not two

The issue describes one boundary. Reading the history there are **three**, because EVENT-sourced
attribution itself only arrived on 2026-08-09:

| window | mechanism | the 3 early types | the 6 upgraded types |
|---|---|---|---|
| 2026-07-31 -> 2026-08-09 | `node["sourceService"] ?: "unknown"` | `lending` | `unknown` |
| 2026-08-09 -> 2026-08-18 | #4270 adds EVENT, then TOPIC fallback | `lending` (EVENT) | `lending-service` (TOPIC) |
| 2026-08-18 -> now | #5399 adds the field to all six | `lending` | `lending` (EVENT) |

- early three: `credit.application.transition`, `credit.decision.evaluated`, `credit.loan.transition`
  (shipped 2026-07-31 / 08-01)
- upgraded six: `loan.disbursed`, `loan.interest_accrued`, `loan.written_off`, `loan.rescheduled`,
  `loan.stage_changed`, `loan.provisioned` (#5399, merged 2026-08-18T19:06:10Z)

So `source_service = 'lending-service'` exists in `audit_entries` **only** for those six types and
**only** in that nine-day window, and `'unknown'` covers them before it. That is narrower than "two
spellings with one boundary", and it changes the cost of a backfill.

No live DB was queried. The query that would settle the row counts:

```sql
SELECT source_service, source_service_source, event_type,
       min(recorded_at), max(recorded_at), count(*)
FROM audit_entries
WHERE source_service IN ('lending', 'lending-service', 'unknown')
GROUP BY 1, 2, 3
ORDER BY 3, 4;
```

`audit_entries` is append-only at the DB (`DO INSTEAD NOTHING`), so a "backfill" is not an `UPDATE`
— that is a fact the options below have to carry.

## 3. The gate

`.github/scripts/check-source-service-convention.py`, registered as `source-service-convention` in
the `kotlin` shard, `mode: enforced`, `min_subjects: 20`.

It resolves the value at each **write site** through all four idioms this fleet uses — named
arg/property default, map entry, `node.put`, hand-built JSON string — follows a `const val` within
the module, and reports an ambiguous const as UNRESOLVED rather than guessing (several modules carry
an unrelated metrics-tag `const val SERVICE = "fx"` beside `SOURCE_SERVICE = "fx-service"`).
Pass-through sites carry no literal and are correctly silent.

**The probe trap fired on this check's own first draft.** A pattern without the optional `: String`
could not see `val sourceService: String = SOURCE_SERVICE`, enumerated **17** of 21 producers, and
printed a confident `OK`. Only the `min_subjects` floor turned that into a visible number, which is
why the floor is declared in both `gates.yaml` and the script.

### Falsification — by effect, not by appearance

| fed | result |
|---|---|
| a real producer given a wrong value (`consent-service` -> `"consent"`) | exit 1, names the file:line |
| the lending baseline removed | exit 1, all nine write sites |
| an emptied producer list (`--root` at an empty dir) | **exit 2** — refuses, does not pass vacuously |
| the current tree | exit 0, `SUBJECTS=21` |

`--self-test` is 21 assertions driving the real `scan()` over a synthetic tree: both directions of
each of the four idioms, pass-through allowed, a prose-only module allowed **and excluded from the
producer count**, an ambiguous const reported, a baseline entry silencing only its own
`(module, value)`, a baseline pinned to a different value **not** silencing the module, and a
stale baseline reported.

## 4. lending is a declared exception — decided, not debt

`("openbank-lending-service", "lending")` is a **declared exception with the decision as its
reason**, not a rename anyone still owes. The entry, its header comment in the script and the
`gates.yaml` comment all now say so explicitly, so a future reader cannot mistake it for a TODO.

**It is permanent rather than deferred, and that is structural.** The rows lending has already
written cannot be converged *by anyone*:

- `audit_entries` is append-only **at the database** — V2 installs `no_update_audit` /
  `no_delete_audit` as `DO INSTEAD NOTHING`, so a normalising `UPDATE` affects **zero rows and
  reports success**;
- `source_service` is hashed into the ADR-0031 / ADR-0133 `record_hash` chain, so rewriting it would
  break tamper-evidence for every affected row **and every row after it**.

So a rename could only add a *fourth* boundary on top of a split that can never be repaired. That is
half the reason keeping `lending` is the cheap answer.

**The gate's behaviour is unchanged by the rewording, deliberately.** The entry still pins the
`(module, VALUE)` pair, so lending drifting to some third spelling still fails, and removing the
entry remains the visible act if #5902 is ever reopened.

## 5. Where the boundary is written down

The three-segment table is the valuable artefact and it does not belong only in a PR body. It now
lives in **`openbank-lending-service/CLAUDE.md`** (new, the repo's per-service-lore convention),
carrying all of it: that `lending` / `lending-service` / `unknown` are one producer; which three
event types read `lending` throughout and which six carry `lending-service` for those nine days;
that `GROUP BY source_service` therefore splits lending three ways; the reconciliation query above;
and the append-only + chain-hashed facts in full. `unknown` is flagged there as the weakest of the
three — it is the fleet-wide "nobody said" sentinel, not lending's alone, so only `event_type`
narrows it.

**`openbank-audit-service/CLAUDE.md`** gets a short pointer, because that is where a
`source_service` histogram gets read and must not be mistaken for a producer census.

This PR is still `Refs #5902`, not `Closes` — the issue closes when this merges.

## Verification

- `run-gates.py --group kotlin` -> **20/20 PASS**; `source-service-convention` reports `SUBJECTS=21`
- **The three negative cases re-run after the rewording, to prove it did not loosen the gate:**
  consent given `"consent"` -> exit 1; the lending entry removed -> exit 1 at all nine write sites;
  an emptied producer list -> exit 2. `--self-test` 21/21.
- `stale-comment-references` and `gate-lifecycle-metadata` -> PASS (the new `CLAUDE.md` is tracked
  and the gate entry declares `rationale` + `review_after`)
- `run-gates.py --all` -> 159 gates, `PASS=152`. The 7 remaining failures were confirmed
  **pre-existing on a pristine `origin/main` worktree** (base-dependent gates with no
  `$PR_DIFF_BASE` locally, plus litellm/loki/temporal/terraform), identical set, unchanged by this PR.
- `gate-lifecycle-metadata` passes: the new entry declares `rationale` + `review_after`.
- No Kotlin module is touched, so no gradle task applies. No `version.txt` / `openapi.yaml` /
  migration change. `.github/**` is outside every release package's paths, and the two `CLAUDE.md`
  files ride a `docs:` commit — a hidden type — so nothing releases on either axis.

Refs #5902, #5256, #5399

### DIFF
--- a/.github/gates/gates.yaml
+++ b/.github/gates/gates.yaml
@@ -1875,10 +1875,12 @@ gates:
   # could not see `val sourceService: String = SOURCE_SERVICE` and enumerated 17 of 21 producers
   # while printing a confident OK. Only the floor turned that into a visible number.
   #
-  # lending is BASELINED with its value and its reason (#5902): converging renames nine shipped
-  # event types and splits audit_entries a second time, so it is an open decision, not a tidy-up.
-  # The gate is green today without hiding the finding, and deleting that baseline entry is the
-  # visible act of taking the decision. ENFORCED.
+  # lending is a DECLARED EXCEPTION, not debt (#5902 decided: keep "lending"). The rows it already
+  # wrote are unrewritable by anyone — audit_entries is append-only at the DB and `source_service`
+  # is chain-hashed — so a rename could only add a fourth boundary on top of a split that can never
+  # be repaired. The boundary is documented in openbank-lending-service/CLAUDE.md instead. The
+  # baseline entry pins the (module, value) pair, so lending drifting to a third spelling still
+  # fails. ENFORCED.
   - id: source-service-convention
     name: "sourceService == module directory name (issues #5256/#5902, enforced)"
     group: kotlin
--- a/.github/gates/gates.yaml
+++ b/.github/gates/gates.yaml
@@ -1857,6 +1857,45 @@ gates:
   # #2187 ledger tie-out + FX revaluation, fx ČNB ingestion, billing cycle) — none had ever
   # run. It is a gate rather than a doc because the natural test cannot fail: calling the
   # method directly supplies the context the scheduler does not. ENFORCED.
+  # #5256/#5902. `source_service` is the column every audit and analytics query groups by, so a
+  # producer whose self-reported spelling disagrees with the fleet convention (module directory
+  # minus `openbank-`) splits itself off from audit-service's TopicAttribution fallback, which uses
+  # the module name. When such a producer later gains EVENT-sourced attribution its rows change
+  # spelling MID-STREAM and the group-by reports two producers where there is one — which is what
+  # PR #5399 did to lending's six upgraded event types on 2026-08-18.
+  #
+  # It is a gate rather than a doc because the natural probe cannot see it: `grep -c sourceService
+  # src/main` is non-zero for all 21 producers — it matches the field being MENTIONED (KDoc, a data
+  # class property, a consumer reading it back), never the value emitted — and that probe reported
+  # full conformance while lending disagreed at nine write sites. This check resolves the value at
+  # each write site through all four fleet idioms, follows a `const val` within the module, and
+  # refuses to guess an ambiguous one.
+  #
+  # min_subjects is inside the script as well as here, on purpose: the first draft of the regex
+  # could not see `val sourceService: String = SOURCE_SERVICE` and enumerated 17 of 21 producers
+  # while printing a confident OK. Only the floor turned that into a visible number.
+  #
+  # lending is BASELINED with its value and its reason (#5902): converging renames nine shipped
+  # event types and splits audit_entries a second time, so it is an open decision, not a tidy-up.
+  # The gate is green today without hiding the finding, and deleting that baseline entry is the
+  # visible act of taking the decision. ENFORCED.
+  - id: source-service-convention
+    name: "sourceService == module directory name (issues #5256/#5902, enforced)"
+    group: kotlin
+    mode: enforced
+    min_subjects: 20
+    rationale: >-
+      `source_service` is the column every audit and analytics query groups by, so a producer
+      whose self-reported spelling disagrees with the module name splits itself off from
+      audit-service's topic fallback — and the natural probe cannot see it, because
+      `grep -c sourceService` matches the field being mentioned, never the value emitted
+      (#5256, #5902).
+    review_after: 2027-02-20
+    selftest: python3 .github/scripts/check-source-service-convention.py --self-test
+    selftest_expect: pass
+    run: |
+      python3 .github/scripts/check-source-service-convention.py --root .
+
   - id: no-runblocking-in-a-scheduled-body
     name: "no runBlocking in a @Scheduled body (rules.yaml: scheduled_methods, enforced)"
     group: kotlin
--- a/.github/scripts/check-source-service-convention.py
+++ b/.github/scripts/check-source-service-convention.py
@@ -41,13 +41,24 @@
 #   decided. They carry no literal, so they are not this check's subject; flagging them would make
 #   the gate noise and silence is the correct verdict.
 #
-# WHY LENDING IS BASELINED RATHER THAN FIXED
-#   Converging lending to "lending-service" renames nine shipped event types and splits the already
-#   split `audit_entries` rows a SECOND time. Whether to converge + backfill, or to keep "lending"
-#   and document the boundary, is a decision recorded in #5902 — not something a gate should force
-#   by going red. So lending is baselined WITH its value and its reason: the gate is green today
-#   without hiding the finding, and deleting that baseline entry is the visible act of taking the
-#   decision. The baseline pins the VALUE, so lending drifting to some third spelling still fails.
+# WHY LENDING IS AN EXCEPTION — A SETTLED DECISION, NOT DEBT
+#   `openbank-lending-service` emits "lending". That is DECIDED and kept (#5902): the boundary it
+#   leaves in `audit_entries` is documented in `openbank-lending-service/CLAUDE.md` — which three
+#   aliases are one producer, which six event types and which nine days the "lending-service" window
+#   covers, and the query that reconciles them. Nothing here is waiting on anyone.
+#
+#   Renaming was rejected on cost, and the cost is structural rather than a matter of effort:
+#   `audit_entries` is append-only AT THE DATABASE (V2's `no_update_audit` / `no_delete_audit` rules
+#   are `DO INSTEAD NOTHING`, so a normalising UPDATE touches zero rows and reports success), and
+#   `source_service` is hashed into the ADR-0031/0133 `record_hash` chain, so rewriting it would
+#   break tamper-evidence for every affected row and every row after it. The existing rows cannot be
+#   converged by anyone, ever; a rename would only add a FOURTH boundary on top of them.
+#
+#   So this entry is not a TODO and should not be "cleaned up" — a future reader finding it must not
+#   read it as a rename nobody got around to. It stays until someone deliberately reopens #5902, and
+#   deleting it is what that reopening looks like. It pins the (module, VALUE) pair, so lending
+#   drifting to some third spelling still fails the gate — that is the property this entry exists to
+#   keep, and it survives independently of the decision above.
 #
 # EXIT CODES
 #   0  every producer's emitted value equals its module directory name (modulo the baseline)
@@ -79,11 +90,12 @@ MIN_PRODUCERS_DEFAULT = 20
 # keeps the entry from absorbing a future third spelling from the same producer.
 BASELINE = {
     ("openbank-lending-service", "lending"): (
-        "#5902 — nine shipped event types emit \"lending\" while the convention (and "
-        "audit-service's TopicAttribution fallback) says \"lending-service\". Converging renames "
-        "nine live types and splits audit_entries a second time, so the choice between "
-        "converge+backfill / keep-and-document is an open decision in #5902, not a tidy-up. "
-        "Remove this entry when that decision is taken."
+        "#5902 DECIDED: lending keeps \"lending\" and the audit_entries boundary is documented "
+        "instead (openbank-lending-service/CLAUDE.md). NOT debt, and NOT a rename awaiting "
+        "cleanup — the existing rows are unrewritable by anyone (audit_entries is append-only at "
+        "the DB and source_service is chain-hashed), so a rename could only add a fourth boundary "
+        "on top of a split that can never be repaired. This entry is permanent unless #5902 is "
+        "deliberately reopened. It pins the VALUE, so a third spelling from this module still fails."
     ),
 }
 
@@ -393,7 +405,8 @@ def run(root, min_producers, quiet=False):
         return 1
     print(
         f"sourceService convention: OK — {len(producers)} producer(s) checked, "
-        f"{len(BASELINE)} baselined. See #5902 for why the baselined set is an open decision.",
+        f"{len(BASELINE)} declared exception(s). See #5902 and BASELINE above: the exception is a "
+        f"settled decision with a documented boundary, not pending work.",
     )
     return 0
 
--- a/.github/scripts/check-source-service-convention.py
+++ b/.github/scripts/check-source-service-convention.py
@@ -35,11 +35,26 @@
 #   `const val SERVICE = "fx"` next to their `SOURCE_SERVICE = "fx-service"`, and picking the wrong
 #   one would produce a confident false verdict in either direction.
 #
-# PASS-THROUGH IS NOT A WRITE SITE
+# PASS-THROUGH IS NOT A WRITE SITE — BUT IT IS A WHITELIST, NOT A FALLBACK
 #   `"sourceService" to event.sourceService` (sepa-instant's publisher) and
 #   `node["sourceService"]?.asText()` (the audit/analytics consumers) copy a value someone else
 #   decided. They carry no literal, so they are not this check's subject; flagging them would make
 #   the gate noise and silence is the correct verdict.
+#
+#   The direction of the default is the load-bearing part. `_resolve` used to end in
+#   `return None, None`, so ANY expression it could not parse was reclassified as a pass-through:
+#   the site produced 0 findings AND 0 producers, so the module disappeared from the check
+#   entirely — invisible AND uncounted, which also put it beneath the MIN_PRODUCERS floor that is
+#   supposed to catch exactly this. Two ordinary Kotlin shapes reproduced it, and in both the
+#   module decides the value it emits:
+#       `"sourceService" to configuredSource`   // a @ConfigProperty-injected field
+#       `"sourceService" to sourceServiceName`  // a lowercase, non-`const` val
+#
+#   So the rule is now: a subject this gate cannot parse must never be indistinguishable from a
+#   subject that passed. Recognised pass-throughs (_PASSTHROUGH) are silent; everything else that
+#   cannot be resolved either FAILS, or appears in UNRESOLVED_ALLOWED with a written reason and is
+#   still COUNTED as a measured subject. Both lists are reverse ratchets, so neither can quietly
+#   outlive the code it describes.
 #
 # WHY LENDING IS AN EXCEPTION — A SETTLED DECISION, NOT DEBT
 #   `openbank-lending-service` emits "lending". That is DECIDED and kept (#5902): the boundary it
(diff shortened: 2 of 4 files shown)


JiRaska:
Review (read-only; no approval implied). Verified against live `origin/main`; the gate was executed, not just read.

## Scope is DERIVED, and the vacuity floor is real — this is the right shape

- `check-source-service-convention.py:174` enumerates `root.glob("openbank-*")` filtered on `(module/"src"/"main").is_dir()`, then `:179` `rglob("*.kt")`. Nothing is enumerated by name, so a new service is in scope the day it exists.
- The subject count is emitted **before** the pass/fail decision (`:373`) and floored at `MIN_PRODUCERS_DEFAULT = 20` (`:87`, mirrored as `min_subjects: 20` in `gates.yaml`), so a broken enumeration exits **2**, never a quiet 0. "Producers" counts only modules with a resolvable write site (`:198,:200`), so prose-only modules cannot pad the floor.

## The exclusion ratchets in BOTH directions — verified by execution, not by reading

`BASELINE` (`:91-100`) pins `(module, VALUE)`, not just module. Measured on an `origin/main` export:

- **New violator**: injected `sourceService="accounts"` into `openbank-account-service/.../AccountService.kt` → `FAIL … emits sourceService="accounts"`, **exit 1**.
- **Stale exclusion**: rewrote lending's literals so the baseline entry matches nothing → `STALE baseline entry (openbank-lending-service, "lending") no longer matches anything`, **exit 1** (`:385`, `:391-396`, `:405`).
- **Third-spelling drift** is covered by the value pinning (self-test `:329-334`).

`--self-test` → **21 passed, exit 0**, and it drives the real `scan()` over a synthetic tree rather than a retyped classifier. It is wired as `selftest:` with `selftest_expect: pass` in `gates.yaml`, which `run-gates.py:363-368` honours. Against a clean `git archive origin/main` export: `SUBJECTS=21 … 21 producer(s) checked, 1 declared exception(s)`, exit 0.

## The one real hole: "cannot parse" is silently treated as "pass-through"

`_resolve` (`:160`) returns `(None, None)` — i.e. skip — for any RHS that is neither a string literal nor a simple const/interpolation. The header at `:38-42` says it "refuses to guess"; in fact it refuses to **report**. Fed three synthetic violators:

```kotlin
mapOf("sourceService" to if (flag) "wrong-one" else "x-service")
ev.copy(sourceService = cfg.name)
ev.copy(sourceService = "wrong".plus("-two"))
```

`scan()` returned `([], set(), set())` — zero findings, **and the module is not counted as a producer**, so the floor cannot catch it either. A `@ConfigProperty`-sourced or conditional `sourceService` is structurally invisible. Latent today (I enumerated all 7 real pass-through sites on `main` — `AuditResource.kt:67,120`, `AuditConsumer.kt:141`, `AnalyticsConsumer.kt:174`, `KafkaSctInstEventPublisher.kt:32` etc. — all genuine), but it is the exact failure class the file's own preamble says it exists to close. Cheap hardening: treat a non-literal RHS as **UNRESOLVED** unless it matches an explicit pass-through allowlist, so it surfaces rather than vanishing.

Two smaller ones:

- Detection is **line-based** (`:180`). A ktlint reformat wrapping `sourceService =` onto the next line makes that write site disappear silently. No wrapped sites exist today.
- `_KOTLIN_SITES[0]` (`:114`) also matches a **comparison**: `IngestAttributionMetrics.kt:50` (`if (envelope.sourceService == UNKNOWN_SERVICE)`) is picked up as a candidate, harmless only because the leading `=` of `==` makes `_resolve` fall through. Anchor it with `(?<!=)=(?!=)` before that becomes a confident false verdict.

## Release scope: clean

`ci(governance)` ×2 + `docs(lending)` ×1, and the diff is exactly 4 files — `.github/gates/gates.yaml`, the new script, and two `CLAUDE.md`. No `.kt`, no `openapi.yaml`, no `version.txt`, no migration. Hidden types, and nothing shipped-code-bearing is hiding under them.

## Scope creep: defensible

`openbank-lending-service/CLAUDE.md` (+66) is a new file from a gate PR, but the `BASELINE` reason string (`:92-99`) and the `gates.yaml` rationale both cite it as where the exception's evidence lives — deleting it would strand the baseline's justification. I would keep it.

**One question for the author**: that file's date/window table (`2026-07-31 → 2026-08-09`) and the "six event types read `unknown` before #4270" claim are assertions about live `audit_entries` rows and cannot be checked from the repo. Did those windows come from a query against the real table, or from reading merge dates? If the latter, the doc presents git archaeology as row-level data, and the wording should say which it is.

## Process note

**28 open PRs currently touch `.github/gates/gates.yaml`.** That is exactly where git reports a clean merge and silently drops one list entry. Before merging, after any `main` merge: confirm the gate list still parses to the expected count and that `source-service-convention` is still present. On this branch I counted 159 gates / 159 unique `id:` values, no duplicates. Do not read the absence of a conflict as evidence.

JiRaska:
## Review — strong gate, ran it and its self-test; the one hole is **still there and I reproduced it**

**Staleness:** 203 commits behind `main`, `MERGEABLE`. `.github/gates/gates.yaml` has moved on `main`, so expect to resolve — see the merge note at the end. Not a duplicate.

**Measured, not read:**
- `--self-test` → `self-test: 21 passed, 0 failed`.
- Against a clean `origin/main` worktree → `SUBJECTS=21`, `OK — 21 producer(s) checked, 1 declared exception(s)`. So the floor of 20 sits one below the real count, as documented.

**What makes this gate better than most in the tree.** Its scope is *derived* — it enumerates `openbank-*` modules and resolves the value at each write site — rather than a hand-kept list of the thing it checks, so a new producer is in scope the day it exists. It refuses vacuity with **exit 2** rather than 0, never conflating "found nothing" with "clean". The baseline pins the `(module, VALUE)` pair, so lending drifting to a third spelling still fails, and a stale entry is reported in the reverse direction. And the ambiguous-const case is `UNRESOLVED` rather than a coin flip. All four self-test properties for that behaviour actually pass.

**The hole: `_resolve` treats an *unparseable* value as a *pass-through*, so it is both invisible and uncounted.** This was flagged in an earlier review and it still stands. `check-source-service-convention.py`, end of `_resolve`:

```python
    return None, None  # pass-through: event.sourceService, node[...], a parameter, ...
```

and in `scan`: `if why is None: continue  # pass-through, not a write site` — reached *before* `producers.add(module.name)`. Anything that is neither a quoted literal nor an ALL-CAPS `const` reference lands here silently.

I built the negative control (fresh tree, one module, `--min-producers 0`) and both of these produce **`0 non-conforming write site(s) ... across 0 producer(s)`** — no finding, and not even counted as a producer:

```kotlin
// case A — a @ConfigProperty-sourced value
val m = mapOf("sourceService" to configuredSource)   // defaultValue = "wrong-spelling"

// case B — a lowercase (non-const) val holding a wrong literal
private val srcSvc = "wrong-spelling"
val m = mapOf("sourceService" to srcSvc)
```

Case B matters most: it is one keyword away from the idiom the gate *does* catch (`const val SOURCE_SERVICE`), so a producer drops out of coverage by dropping `const`, and nothing anywhere says so. The `MIN_PRODUCERS` floor is the only backstop and it is fleet-wide — losing one module of 21 leaves 20, exactly the floor, so it does not even trip.

**Suggested fix (small).** A true pass-through is narrow and recognisable: a *member access* (`event.sourceService`, `node[...]`), or a bare parameter. Everything else that is a simple identifier is a value the module chose and could not resolve. Invert the default:

- keep `(None, None)` only for expressions containing `.` on a lowercase receiver, subscripts, or calls;
- return `(None, "references <name>, which this module declares no `const val` for")` for a bare simple identifier of *any* case, not only `[A-Z][A-Z0-9_]*`.

Then add two self-test modules — `openbank-lowercase-val-service` and `openbank-config-property-service`, both expecting a finding — so the fix is proven by what it rejects. Right now no self-test case exercises an unresolvable non-ALL-CAPS reference at all, which is why the hole survives a 21/21 green.

**Merge mechanics, and it is not incidental here.** #5875 in this same batch also appends entries to the `gates:` list in `.github/gates/gates.yaml`. Two branches adding neighbouring entries to one list is the merge git reports as clean while keeping only one — no conflict, nothing printed, exit 0. After resolving, **count the `- id:` entries before and after** rather than reading the diff.

**Not checked:** the `openbank-audit-service/CLAUDE.md` and `openbank-lending-service/CLAUDE.md` prose, beyond confirming the lending boundary documentation the baseline entry depends on is what it points at.

Verdict: **needs change** — the pass-through default, plus its two negative-control cases. Everything else is ready. Not approving — review comment only.

JiRaska:
## `_resolve` treats *unparseable* as *pass-through*, so an unreadable write site is invisible AND uncounted

Pushed a fix to this branch (`9a4abd602`).

### The defect

`_resolve` ended in:

```python
return None, None  # pass-through: event.sourceService, node[...], a parameter, ...
```

That is a **fallback**, not a whitelist. Any expression the resolver could not parse was
reclassified as a pass-through, and `scan()` then `continue`s *before* `producers.add(module.name)`.
So such a site yields **0 findings AND 0 producers** — the module vanishes from the check entirely.

The second half is what makes it unrecoverable: because the module is never counted, the
`MIN_PRODUCERS` floor cannot see the loss either. The one mechanism in this script that exists to
catch a broken enumeration — and which the header rightly credits with turning the original
17-of-21 gap into a visible number — was itself blind to this one.

### Two negative controls, both ordinary Kotlin

In both, the **module decides** the value it emits. Neither is a pass-through.

```kotlin
// NC1 — a @ConfigProperty-injected field
@ConfigProperty(name = "openbank.audit.source-service")
lateinit var configuredSource: String
val m = mapOf("sourceService" to configuredSource)

// NC2 — a lowercase, non-`const` val
private val sourceServiceName = "totally-wrong"
val m = mapOf("sourceService" to sourceServiceName)
```

Measured against `scan()` directly:

```
BEFORE   findings = 0 ; producers = 0 ; []
  openbank-nc-config-service       finding=False  counted_as_producer=False
  openbank-nc-lowercase-service    finding=False  counted_as_producer=False
  openbank-nc-passthrough-service  finding=False  counted_as_producer=False   <- genuine pass-through

AFTER    findings = 2 ; producers = 2
  openbank-nc-config-service       finding=True   counted_as_producer=True
  openbank-nc-lowercase-service    finding=True   counted_as_producer=True
  openbank-nc-passthrough-service  finding=False  counted_as_producer=False   <- still silent
```

Before the fix all three rows are identical. That is the whole bug: a subject the gate could not
parse was indistinguishable from one that passed.

### The fix

**The default is now "report", not "skip".** `_PASSTHROUGH` became a whitelist — an expression that
reads a `sourceService` somebody else decided. Anything else unresolvable is reported with the
expression quoted.

For the residue, `UNRESOLVED_ALLOWED` pins `(module, expression) -> reason`. Such a site is
**still counted** as a measured subject, and the list is a reverse ratchet like `BASELINE`, so it
cannot quietly outlive the code it describes. This is the wider rule the gap teaches: *if a value
cannot be resolved it either fails, or it is counted as an explicit skip with a written reason —
it is never silently absent.*

Two further things fell out of running the stricter version against the real tree:

- **`sourceService == UNKNOWN_SERVICE` parsed as a write site emitting `"= UNKNOWN_SERVICE"`.** The
  assignment pattern ate the first `=` of the `==` and captured the rest. A comparison is a read;
  reporting it is the noise that gets a gate switched off. Fixed with `(?<![=!<>])=(?!=)`.
  Live shape: `analytics-sink/IngestAttributionMetrics.kt:50`.
- **`node["sourceService"]?.asText()` was not recognised** even though the module header has always
  called it a pass-through. Added to `_PASSTHROUGH`.

The one real acknowledged skip is `openbank-audit-service`'s `resolvedSource.first` —
`resolveSourceService(node, address)` (AuditConsumer.kt:218) returns the event's own value, else
`TopicAttribution`'s topic-derived one, else `"unknown"`; all three are somebody else's attribution
being copied. Pinned rather than whitelisted by shape, because `<val>.first` is far too generic a
pattern to make silent.

### Verification

- Two negative controls: **flagged and counted** after, silent and uncounted before (above).
- Both are now **permanent self-test cases**, with allow-side regression guards for the comparison
  and node-read shapes. Self-test **21 -> 25 cases, 25 passed / 0 failed**.
- Stale-ratchet control: a bogus `UNRESOLVED_ALLOWED` entry -> **exit 1**, `STALE  acknowledged-unresolvable entry (...) no longer matches any write site`. (First attempt at this control mis-injected and reported exit 0 — a broken probe reporting clean. Re-run with an assertion on the injection.)
- Real tree: **exit 0**, and `SUBJECTS` goes **21 -> 22**. audit-service was one of the modules the
  hole was hiding.
- Gate via its real runner `run-gates.py --only source-service-convention`: **PASS**.
- Shards `kotlin` / `registry` / `lint` in the foreground: **20 / 26 / 20 PASS, exit 0** each.
- `ruff` reports one `RUF100` on this file — pre-existing, fires on the unpatched original too.

JiRaska:
### Heads-up before you merge `main` into this branch: `.github/gates/gates.yaml` entry count

Counted just now (`grep -cE '^  - id:'`):

| ref | gate entries |
|---|---|
| `origin/main` | **171** |
| this branch's head (`9a4abd6`) | **159** |

The merge-base is **208 commits behind `main`**, so the 12-entry gap is main's growth, not a deletion
here — but this is precisely the shape that has silently lost an entry in this repo before
(`.release-please-manifest.json`, 56 → 55, with a clean merge and exit 0). Two sides appending
neighbouring list entries produces **no conflict and no output**.

After merging `main` in, re-run the count and confirm it is `171 + <however many this PR adds>`.
Same applies to #5875, which also appends to this list (173 at its head vs main's 171).

JiRaska:

|---|---|
| #5890 | 3934 |
| #5849 | 3853 |
| #5875 | 3735 |
| #5967 | 3417 |

PR #5969, opened 2026-08-20
fix(party): reconcile the published party enums with the domain types they serve

JiRaska:
Refs #5962, #5895, ADR-0048.

Stacked on #5966 (which adds the gate and its `BASELINE`) — that branch is merged in, so
its diff drops out once #5966 lands. Only `openbank-party-service/src/main/resources/openapi.yaml`
and three `BASELINE` lines are this PR's own work.

## What was wrong

Three enums in the party spec were hand-written once and never re-checked against the Kotlin
types behind them. Two advertise values the code does not have, and `PartyResource` calls a bare
`valueOf` on both, so a client generated from the published document is answered **400**:

|---|---|---|
| `AddDocumentRequest.documentType` `[PASSPORT, ID_CARD, DRIVING_LICENSE, COMPANY_REGISTRATION]` | `DocumentType { NATIONAL_ID, PASSPORT, DRIVING_LICENCE, COMPANY_REGISTRATION, TAX_ID }` | `ID_CARD`/`DRIVING_LICENSE` -> `DocumentType.valueOf` throws (`PartyResource.kt:244`) -> 400. `NATIONAL_ID`, `TAX_ID` unpublished |
| `KycStatusRequest.kycStatus` `[PENDING, VERIFIED, REJECTED, EXPIRED]` | `KycStatus { NOT_STARTED, IN_PROGRESS, APPROVED, REJECTED, EXPIRED }` | `PENDING`/`VERIFIED` -> `KycStatus.valueOf` throws (`PartyResource.kt:264`) -> 400. `NOT_STARTED`, `IN_PROGRESS`, `APPROVED` unpublished |
| party-list `status` filter `[PENDING_KYC, ACTIVE, SUSPENDED, CLOSED]` | `PartyStatus { PENDING_KYC, ACTIVE, SUSPENDED, CLOSED, MERGED }` | `MERGED` (ADR-0179, the dedup survivor state) is unfilterable — the resource silently ignores an unparseable value rather than 400ing |

## Which side is right — per name

The domain, on all three, measured rather than assumed:

- **The service's own docs already republish the code spelling verbatim**, in both languages:
  `03-api.en.md:99` says `documentType` (code enum) is
  `NATIONAL_ID / PASSPORT / DRIVING_LICENCE / COMPANY_REGISTRATION / TAX_ID`, and
  `01-overview.en.md:8` repeats it. A second document in the same service agreeing with the code
  settles which spelling is canonical without a judgement call.
- **`ID_CARD` and `DRIVING_LICENSE` were never renamed — they never existed.** `git log -S` over
  `openbank-party-service` finds both only in `openapi.yaml`, at every revision, never in any
  Kotlin source. (pid-service has its own, separate `IdDocumentType` that does spell it
  `DRIVING_LICENSE`; that is a different service and a different enum, and does not corroborate
  party's spec.)
- Same for `VERIFIED` in `KycStatus`: only ever in the spec.
- `MERGED` is written by the ADR-0179 identity-merge path and carries a DB CHECK constraint tying
  it to `merged_into_party_id`, so it is unambiguously real.

- **admin-ui** — no `documentType` or `kycStatus` call site at all.
- **pacts** — none of the four party pacts (`account`, `kyc`, `onboarding`, `vop` consumers)
  contains any of these literals.
- **fleet grep** — `ID_CARD`/`DRIVING_LICENSE` appear nowhere outside pid's unrelated enum.
- **DB** — `party_documents.document_type` is `VARCHAR(30)` with no CHECK constraint and
  `parties.kyc_status` likewise, so nothing stored has to change and no migration is needed.

`check-api-contract.py` against a live merge-base:

```
::notice::api-contract gate: openbank-party-service: spec-only PR — reclassifying 4 breaking
document change(s) (first: removed the enum value `DRIVING_LICENSE` of the request property
`documentType`) as a CORRECTION. ... Requiring MINOR instead.
api-contract gate: openbank-party-service: correction change, info.version 1.17.0 -> 1.18.0 — OK.
```

`1.17.0` read off live `origin/main` immediately before the bump. Same class as the aml precedent
(#2312/#2313) and kyc (#5966): the running server is byte-identical, so no working client can
break, and a MAJOR would violate the ADR-0048 D2 URL-major invariant.

## Verification

- `./gradlew :openbank-party-service:ktlintCheck :openbank-party-service:detekt :openbank-party-service:test`
  — `BUILD SUCCESSFUL`, exit 0, all three tasks executed (none `UP-TO-DATE`).
  **166 tests, 0 failures, 1 skipped** — the skip is
  `PartyEventPactProviderVerificationTest`, the broker-sourced class gated on `pactbroker.url`,
  which does not run locally by design.
  `24 paired enum(s) drift, all 24 baselined` (was 27), and would fail on a stale entry, so the
  three removals are proven to be genuinely reconciled rather than merely deleted.

Baseline: **27 -> 24**.

### DIFF
--- a/.github/gates/gates.yaml
+++ b/.github/gates/gates.yaml
@@ -2186,6 +2186,42 @@ gates:
     run: |
       python3 .github/scripts/check-pact-provider-replay.py
 
+  # openapi enum vs the Kotlin enum behind it (issue #5895 / #5962, enforced).
+  #
+  # The axis the api-contract gate below structurally CANNOT see: oasdiff compares a spec to its
+  # own previous version, never to the implementation. So a spec enum hand-written once and never
+  # re-checked drifts forever and every gate stays green. openbank-kyc-service published
+  # `[IDENTITY, SANCTIONS, PEP, ADVERSE_MEDIA, SOURCE_OF_FUNDS]` against a domain
+  # `CheckType { IDENTITY, ADDRESS, PEP_SCREENING, SANCTIONS_SCREENING, ADVERSE_MEDIA }` from the
+  # day the file was written — three names misspelled, one real check unpublished, and one value
+  # that has never existed in the code. Its own PUT body was fiction in the same way.
+  #
+  # Pairing is by VALUE OVERLAP, not by name: nothing declares which Kotlin enum a spec enum
+  # serves, and matching on names would find almost none of them. An unpaired spec enum is not a
+  # finding (many are free-form vocabularies with no Kotlin type).
+  #
+  # RATCHET: the 27 drifts across 16 services that exist today are a BASELINE constant in the
+  # script (owned by #5962), so this is enforced from day one and number 28 is red at PR time. A
+  # baseline entry that no longer drifts also FAILS, so the list cannot rot into furniture — same
+  # contract as check-pact-provider-replay.py and check-adr-partial-followup.py above.
+  - id: openapi-enum-domain-drift
+    name: "openapi enum matches the domain enum it serves (issue #5962, enforced)"
+    group: api
+    mode: enforced
+    rationale: >
+      oasdiff compares a spec to its own previous version and never to the implementation, so a
+      spec enum hand-written once and never re-checked drifts forever with every gate green.
+      openbank-kyc-service published three misspelled check names, omitted a real check, and
+      advertised a value (SOURCE_OF_FUNDS) that has never existed in the code — from the day the
+      file was written until #5895, and 27 more pairings across 16 services still do (#5962).
+    review_after: '2027-02-20'
+    min_subjects: 100
+    selftest: |
+      python3 .github/scripts/check-openapi-enum-vs-domain.py --self-test
+    selftest_expect: pass
+    run: |
+      python3 .github/scripts/check-openapi-enum-vs-domain.py --enforce
+
   # ADR-0048 D5: api-contract gate — classify the API-contract bump from the
   # OpenAPI diff (oasdiff), independently of the release axis (version.txt):
   #   breaking  => info.version MAJOR (new URL major /api/v{N+1})
--- a/.github/scripts/check-openapi-enum-vs-domain.py
+++ b/.github/scripts/check-openapi-enum-vs-domain.py
@@ -1,0 +1,355 @@
+#!/usr/bin/env python3
+"""openapi-enum-domain-drift gate — does a spec enum still match the Kotlin enum behind it?
+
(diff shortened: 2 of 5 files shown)


JiRaska:
## Review — #5969 (party)

Reviewed with `git diff origin/main...origin/fix/party-openapi-enum-drift-5962`. **No change
requested; one nit.**

**Independent verification of the three claims** (`origin/main`, not the PR body):

* `openbank-party-service/src/main/kotlin/com/openbank/party/domain/model/Party.kt:18` —
  `DocumentType { NATIONAL_ID, PASSPORT, DRIVING_LICENCE, COMPANY_REGISTRATION, TAX_ID }`
* `Party.kt:74` — `KycStatus { NOT_STARTED, IN_PROGRESS, APPROVED, REJECTED, EXPIRED }`
* `Party.kt:17` — `PartyStatus { PENDING_KYC, ACTIVE, SUSPENDED, CLOSED, MERGED }`, with the
  KDoc immediately above it (`Party.kt:11-16`) explaining that `CLOSED` and `MERGED` are
  deliberately distinct under ADR-0179. That KDoc is a second witness for `MERGED` beyond the
  enum itself.

All three new spec lists are byte-equal to those. Note `DRIVING_LICENCE` vs the spec's
`DRIVING_LICENSE`: this is the British/American spelling split, and the code spelling wins because
it is the one `DocumentType.valueOf` accepts.

**Version.** `origin/main` reads `1.17.0` as of this review, so `1.18.0` is correct and free. No
competing open PR touches `openbank-party-service/src/main/resources/openapi.yaml`.

**Pacts.** party's `@PactFolder` replay is `PartyPactFolderProviderVerificationTest`. None of
`ID_CARD`, `DRIVING_LICENSE`, `PENDING` or `VERIFIED` appears in any committed pact, so no
interaction asserts a removed value.

**Nit — the version comment above the bumped line is stale, and this PR walks past it.**
`openapi.yaml:5-8` still reads `Minor bump (1.8 -> 1.9): additive only — a new PATCH
/api/v1/parties/approvals/{id} endpoint …`. That comment was already wrong on `main` (it describes
1.9, the file was at 1.17.0), so this PR did not introduce it — but a PR whose entire subject is
"the published document drifted from what it describes" is the natural place to either update it
to describe the 1.18.0 bump or delete it. Leaving a comment that misdescribes the version
immediately below it is the same defect class in miniature. Same nit applies across the family;
see #5971, #5982, #5990, #6015, #6016, where the comment *was* accurate on `main` and is made
stale by the bump.

---

JiRaska:

PR #5971, opened 2026-08-20
fix(clearing): publish the real PaymentRail vocabulary

JiRaska:
Refs #5962, ADR-0048.

Stacked on #5966 (which adds the gate and its `BASELINE`) — that branch is merged in, so its
diff drops out once #5966 lands. This PR's own work is one enum line and one version line in
`openbank-clearing-service/src/main/resources/openapi.yaml`, plus one `BASELINE` entry.

## What was wrong

`SubmitPaymentRequest.rail` published `[SEPA_SCT, SEPA_SCT_INST, CZ_DOMESTIC, SWIFT]`. The
domain type is `PaymentRail { SEPA_SCT, SEPA_SCT_INST, SWIFT, DOMESTIC, INTERNAL }`.

- `CZ_DOMESTIC` is a rail this service has never had. `ClearingResource.kt:121` calls
  `PaymentRail.valueOf(rail)` bare, so it throws `IllegalArgumentException` -> **400** (libs-runtime
  maps IAE to 400). A client generated from the published document cannot submit a domestic
  payment at all.
- `DOMESTIC` and `INTERNAL` are real, routable rails and were unpublished, so no consumer could
  be written against them.

## Which side is right — the domain, on three independent witnesses

1. **The database says so.** `V1__init_clearing.sql:3` declares a Postgres enum type:
   `CREATE TYPE payment_rail AS ENUM ('SEPA_SCT', 'SEPA_SCT_INST', 'SWIFT', 'DOMESTIC', 'INTERNAL');`
   `clearing_batches.rail` is that type, so `CZ_DOMESTIC` is not merely unmapped in Kotlin — it is
   unstorable. This is stronger corroboration than kyc had (#5966), where the column was free TEXT.
2. **The service's own API docs already name the code vocabulary and flag the mismatch in prose.**
   `docs/03-api.en.md:21`: "`rail` query / enum values: `SEPA_SCT`, `SEPA_SCT_INST`, `SWIFT`,
   `DOMESTIC`, `INTERNAL` (the OpenAPI `SubmitPaymentRequest.rail` enum lists ... `CZ_DOMESTIC` ...;
   the domain enum uses `DOMESTIC`/`INTERNAL` — note the ... naming discrepancy between contract
   and code)." Someone documented the drift instead of fixing it. Same text in `03-api.cs.md`.
3. **`CZ_DOMESTIC` was never renamed — it never existed.** `git log -S "CZ_DOMESTIC"` over
   `openbank-clearing-service` finds it only in `openapi.yaml`, at every revision, never in any
   Kotlin source or migration.

## No consumer sends it

Fleet grep for `CZ_DOMESTIC` returns only the spec, the two doc files above, and the gate's own
baseline line. No admin-ui call site; the one clearing pact
(`openbank-swift-service-openbank-clearing-simulator.json`) is against the simulator and contains
none of these literals.

## Deliberately NOT touching the docs

The two `03-api` files still describe the now-fixed discrepancy and should be tidied — but editing
them would make this PR touch a second file in the service, which by design removes the
`correction` classification below and would demand a URL major. Left for a follow-up `docs(clearing)`
PR under #5962.

`check-api-contract.py` against a live merge-base:

```
::notice::api-contract gate: openbank-clearing-service: spec-only PR — reclassifying 1 breaking
document change(s) (first: removed the enum value `CZ_DOMESTIC` of the request property `rail`)
as a CORRECTION. ... Requiring MINOR instead.
api-contract gate: openbank-clearing-service: correction change, info.version 1.3.0 -> 1.4.0 — OK.
```

`1.3.0` read off live `origin/main` immediately before the bump.
`check-threat-model-diff.py --base <live merge-base>` exits 0 — run, not assumed.

## Verification

- `./gradlew :openbank-clearing-service:ktlintCheck :detekt :test` — `BUILD SUCCESSFUL`, exit 0.
  `ktlintCheck`, `detekt` and `test` all executed (`test` is not `UP-TO-DATE`).
  **45 tests, 0 failures, 0 errors, 0 skipped.**
  `26 paired enum(s) drift, all 26 baselined`, and fails on a stale entry, so the removal is
  proven reconciled rather than merely deleted.

Baseline: **27 -> 26** for this PR alone (24 with #5969 also landed).

### DIFF
--- a/.github/gates/gates.yaml
+++ b/.github/gates/gates.yaml
@@ -2186,6 +2186,42 @@ gates:
     run: |
       python3 .github/scripts/check-pact-provider-replay.py
 
+  # openapi enum vs the Kotlin enum behind it (issue #5895 / #5962, enforced).
+  #
+  # The axis the api-contract gate below structurally CANNOT see: oasdiff compares a spec to its
+  # own previous version, never to the implementation. So a spec enum hand-written once and never
+  # re-checked drifts forever and every gate stays green. openbank-kyc-service published
+  # `[IDENTITY, SANCTIONS, PEP, ADVERSE_MEDIA, SOURCE_OF_FUNDS]` against a domain
+  # `CheckType { IDENTITY, ADDRESS, PEP_SCREENING, SANCTIONS_SCREENING, ADVERSE_MEDIA }` from the
+  # day the file was written — three names misspelled, one real check unpublished, and one value
+  # that has never existed in the code. Its own PUT body was fiction in the same way.
+  #
+  # Pairing is by VALUE OVERLAP, not by name: nothing declares which Kotlin enum a spec enum
+  # serves, and matching on names would find almost none of them. An unpaired spec enum is not a
+  # finding (many are free-form vocabularies with no Kotlin type).
+  #
+  # RATCHET: the 27 drifts across 16 services that exist today are a BASELINE constant in the
+  # script (owned by #5962), so this is enforced from day one and number 28 is red at PR time. A
+  # baseline entry that no longer drifts also FAILS, so the list cannot rot into furniture — same
+  # contract as check-pact-provider-replay.py and check-adr-partial-followup.py above.
+  - id: openapi-enum-domain-drift
+    name: "openapi enum matches the domain enum it serves (issue #5962, enforced)"
+    group: api
+    mode: enforced
+    rationale: >
+      oasdiff compares a spec to its own previous version and never to the implementation, so a
+      spec enum hand-written once and never re-checked drifts forever with every gate green.
+      openbank-kyc-service published three misspelled check names, omitted a real check, and
+      advertised a value (SOURCE_OF_FUNDS) that has never existed in the code — from the day the
+      file was written until #5895, and 27 more pairings across 16 services still do (#5962).
+    review_after: '2027-02-20'
+    min_subjects: 100
+    selftest: |
+      python3 .github/scripts/check-openapi-enum-vs-domain.py --self-test
+    selftest_expect: pass
+    run: |
+      python3 .github/scripts/check-openapi-enum-vs-domain.py --enforce
+
   # ADR-0048 D5: api-contract gate — classify the API-contract bump from the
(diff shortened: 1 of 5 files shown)


JiRaska:
## Review — #5971 (clearing)

Reviewed with `git diff origin/main...origin/fix/clearing-openapi-enum-drift-5962`. **No change
requested; one nit.**

**This is the strongest witness in the family, and it holds.** The PR claims a Postgres enum type;
verified directly:

`openbank-clearing-service/src/main/resources/db/migration/V1__init_clearing.sql:3`

```sql
CREATE TYPE payment_rail AS ENUM ('SEPA_SCT', 'SEPA_SCT_INST', 'SWIFT', 'DOMESTIC', 'INTERNAL');
```

and line 8 declares `rail payment_rail NOT NULL`. This is not a `VARCHAR` + `CHECK` — it is a
native enum type, so `CZ_DOMESTIC` is not merely unmapped in Kotlin, it is **unrepresentable in
the column**. No batch has ever carried it and none ever could without a migration. The Kotlin
agrees: `ClearingModels.kt:14` is
`enum class PaymentRail { SEPA_SCT, SEPA_SCT_INST, SWIFT, DOMESTIC, INTERNAL }`, byte-equal to the
new spec line. There is no judgement call left here.

**Version.** `origin/main` reads `1.3.0`, so `1.4.0` is correct and free; no competing open PR
touches this file.

**Money-path service** (`clearing` is in `rules.yaml: money_path_services`) — this needs two
approvals and a threat-model check, and per the family note below, approving a *downstream* PR in
the stack does not supply that.

**Nit.** `openapi.yaml:5-8` carries `Minor bump (1.2 -> 1.3): additive only — GET /approvals …`.
Unlike party's, that comment was **accurate on `main`** and this PR makes it stale: the file now
says 1.4.0 while the comment describes the 1.3.0 bump. Either update it or drop it — a one-line
change, but this PR's whole thesis is that a document nobody re-checks drifts from what it
describes.

---

JiRaska:

PR #5974, opened 2026-08-20
fix(sca): publish the ScaStatus and ScaPurpose vocabularies the service actually uses

JiRaska:
Refs #5962, ADR-0048.

drops out once #5966 lands. This PR's own work is two enums and one version line in
`openbank-sca-service/src/main/resources/openapi.yaml`, plus one `BASELINE` entry.

## What was wrong

|---|---|---|
| `ScaChallengeResponse.status` `[PENDING, VERIFIED, FAILED, EXPIRED]` | `ScaStatus { PENDING, COMPLETED, FAILED, EXPIRED, CANCELLED }` | the server emits `COMPLETED`/`CANCELLED`, which the document does not declare, and has never emitted `VERIFIED`. A strict generated deserializer fails on the terminal success state |
| `InitiateScaRequest.purpose` `[PAYMENT_AUTHORIZATION, LOGIN, CONSENT, ACCOUNT_CHANGE, SAVINGS_WITHDRAW_APPROVAL]` | `ScaPurpose`, ten values | `PAYMENT_AUTHORIZATION`, `CONSENT`, `ACCOUNT_CHANGE` are rejected requests; seven real purposes — including `PAYMENT_INITIATION`, the primary one — were unpublished |

`status` is serialized straight off the domain type (`ScaResource.kt:120`, `status = c.status`) and
`purpose` deserializes directly into `ScaPurpose` (`InitiateScaRequest.purpose: ScaPurpose`), so
neither has any translation layer that could have absorbed the difference.

## Which side is right — the domain, and here the provider replay proves it

- **A committed pact asserts the value the spec omits.**
  `pacts/openbank-delegation-service-openbank-sca-service.json` contains
  `"status": "COMPLETED"` on this exact field, twice, and `"purpose": "DELEGATION_GRANT"` —
  another value the spec did not publish. That pact is replayed against the real provider by
  `ScaPactFolderProviderVerificationTest` (`@PactFolder`, ungated, runs on every PR), and it is
  **green in this branch: 3 interactions, 0 failures, 0 skipped**. So this is not a reading of the
  source — the running server was asked and answered `COMPLETED`.
- **The spec itself already said the `purpose` list was wrong, and deferred the fix to this PR.**
  The lines being replaced carried:
  "NOTE: this list does NOT match the ScaPurpose enum the service actually accepts ...
  PAYMENT_AUTHORIZATION / CONSENT / ACCOUNT_CHANGE have never been valid values. Pre-existing
  drift, deliberately NOT corrected here: deleting the three phantom values is a breaking spec
  change and belongs in its own PR."
- **`VERIFIED` was never renamed — it never existed.** `git log -S "VERIFIED"` over
  `openbank-sca-service` finds it only in `openapi.yaml`, at every revision.
- No admin-ui call site sends either field; the consent pact carries only `PENDING`, which both
  documents agree on.

## The `purpose` enum is one the gate could not see, and that is worth noting

`check-openapi-enum-vs-domain.py` pairs a spec enum to a Kotlin enum by value overlap against the
**larger** set. The `purpose` drift shared 2 values with a 10-value `ScaPurpose`, i.e. 0.2, under
`MIN_OVERLAP = 0.4` — so it never paired and was never reported, in the baseline or out of it.
The calibration is right (loosening it re-admits the coincidence findings #5966's self-test pins
down), but the consequence is real: **the more thoroughly a spec enum has drifted, the less likely
the gate is to pair it**, so the baseline of 27 is a floor, not a census. Visible in this PR's own
numbers — `SUBJECTS` goes 147 -> 148, because the corrected `purpose` enum now pairs. Same reason
`challengeType` is untouched below. Worth a note on #5962; not something to fix by moving the
threshold.

`InitiateScaRequest.challengeType` is declared `required` with enum `[OTP_SMS, OTP_EMAIL, TOTP, PUSH]`,
and the Kotlin DTO has **no such property at all** — the nearest thing is an optional
`preferredMethod: ScaMethod?` with values `{BIOMETRIC, PUSH_NOTIFICATION, TOTP}`. That is a
property-shape defect, not an enum drift: fixing it means deciding whether the field is renamed or
dropped and whether it stays required, which is an API judgement rather than a reconciliation.
Left for its own PR under #5962 rather than smuggled into a money-path spec change.

```
::notice::api-contract gate: openbank-sca-service: spec-only PR — reclassifying 3 breaking
document change(s) (first: removed the enum value `ACCOUNT_CHANGE` of the request property
`purpose`) as a CORRECTION. ... Requiring MINOR instead.
api-contract gate: openbank-sca-service: correction change, info.version 1.5.0 -> 1.6.0 — OK.
```

`1.5.0` read off live `origin/main` immediately before the bump.
`check-threat-model-diff.py --base <live merge-base>` exits 0 — run, not assumed.

## Verification

- `./gradlew :openbank-sca-service:ktlintCheck :detekt :test` — `BUILD SUCCESSFUL`, exit 0, all
  three executed. **147 tests, 0 failures, 0 errors, 1 skipped** — the skip is
  `ScaPactProviderVerificationTest`, the broker-sourced class gated on `pactbroker.url`, which
  does not run locally by design. Its `@PactFolder` sibling did run, green, as above.
  `26 paired enum(s) drift, all 26 baselined`, and fails on a stale entry.

Baseline: **27 -> 26** for this PR alone.

### DIFF
--- a/.github/gates/gates.yaml
+++ b/.github/gates/gates.yaml
@@ -2186,6 +2186,42 @@ gates:
     run: |
       python3 .github/scripts/check-pact-provider-replay.py
 
+  # openapi enum vs the Kotlin enum behind it (issue #5895 / #5962, enforced).
+  #
+  # The axis the api-contract gate below structurally CANNOT see: oasdiff compares a spec to its
+  # own previous version, never to the implementation. So a spec enum hand-written once and never
+  # re-checked drifts forever and every gate stays green. openbank-kyc-service published
+  # `[IDENTITY, SANCTIONS, PEP, ADVERSE_MEDIA, SOURCE_OF_FUNDS]` against a domain
+  # `CheckType { IDENTITY, ADDRESS, PEP_SCREENING, SANCTIONS_SCREENING, ADVERSE_MEDIA }` from the
+  # day the file was written — three names misspelled, one real check unpublished, and one value
+  # that has never existed in the code. Its own PUT body was fiction in the same way.
+  #
+  # Pairing is by VALUE OVERLAP, not by name: nothing declares which Kotlin enum a spec enum
+  # serves, and matching on names would find almost none of them. An unpaired spec enum is not a
+  # finding (many are free-form vocabularies with no Kotlin type).
+  #
+  # RATCHET: the 27 drifts across 16 services that exist today are a BASELINE constant in the
+  # script (owned by #5962), so this is enforced from day one and number 28 is red at PR time. A
+  # baseline entry that no longer drifts also FAILS, so the list cannot rot into furniture — same
+  # contract as check-pact-provider-replay.py and check-adr-partial-followup.py above.
+  - id: openapi-enum-domain-drift
+    name: "openapi enum matches the domain enum it serves (issue #5962, enforced)"
+    group: api
+    mode: enforced
+    rationale: >
+      oasdiff compares a spec to its own previous version and never to the implementation, so a
+      spec enum hand-written once and never re-checked drifts forever with every gate green.
+      openbank-kyc-service published three misspelled check names, omitted a real check, and
+      advertised a value (SOURCE_OF_FUNDS) that has never existed in the code — from the day the
+      file was written until #5895, and 27 more pairings across 16 services still do (#5962).
+    review_after: '2027-02-20'
+    min_subjects: 100
+    selftest: |
+      python3 .github/scripts/check-openapi-enum-vs-domain.py --self-test
+    selftest_expect: pass
+    run: |
+      python3 .github/scripts/check-openapi-enum-vs-domain.py --enforce
+
   # ADR-0048 D5: api-contract gate — classify the API-contract bump from the
   # OpenAPI diff (oasdiff), independently of the release axis (version.txt):
   #   breaking  => info.version MAJOR (new URL major /api/v{N+1})
--- a/.github/scripts/check-openapi-enum-vs-domain.py
+++ b/.github/scripts/check-openapi-enum-vs-domain.py
@@ -1,0 +1,355 @@
+#!/usr/bin/env python3
+"""openapi-enum-domain-drift gate — does a spec enum still match the Kotlin enum behind it?
+
+The axis nothing else watches. `oasdiff` (the api-contract gate, ADR-0048) compares a spec to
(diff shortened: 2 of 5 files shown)


JiRaska:
## Review — #5974 (sca)

Reviewed with `git diff origin/main...origin/fix/sca-openapi-enum-drift-5962`. **No change
requested.**

**Verified against the code and against a committed pact, which is the independent witness this
family needs:**

* `openbank-sca-service/src/main/kotlin/com/openbank/sca/domain/model/ScaChallenge.kt:57` —
  `ScaStatus { PENDING, COMPLETED, FAILED, EXPIRED, CANCELLED }`. Byte-equal to the new spec list.
* `ScaChallenge.kt:16` — `ScaPurpose` with `CONSENT_GRANT, PAYMENT_INITIATION, LOGIN, AGENT_ACTION,
  SENSITIVE_DATA_ACCESS, DOCUMENT_SIGNING, CARD_MANAGEMENT, DELEGATION_GRANT, DELEGATION_ACCEPT,
  SAVINGS_WITHDRAW_APPROVAL` — ten values, matching the new list.
* **The pact witness is real.** `pacts/openbank-delegation-service-openbank-sca-service.json:21-22`
  asserts `"purpose": "DELEGATION_GRANT"` and `"status": "COMPLETED"`, with the provider state
  named `a COMPLETED DELEGATION_GRANT SCA challenge exists` (line 10). Both are values the old
  spec did **not** publish. A committed pact asserting a value the published document denies
  exists is about as clean a proof as this axis gets that the spec was the wrong side.

**Pact impact of the removals — checked, and nothing breaks.** The other sca pact,
`pacts/openbank-consent-service-openbank-sca-service.json:21-22`, asserts
`"purpose": "CONSENT_GRANT"` / `"status": "PENDING"`. Both survive in the new lists. The provider
replay `ScaPactFolderProviderVerificationTest` is `@PactFolder`, so it runs on a PR rather than
only on main-push.

**Version.** `origin/main` reads `1.5.0`; `1.6.0` is correct and free, no competing PR on this file.

**Worth calling out as a positive:** the removed comment at the old `purpose` property explicitly
deferred this fix to "its own PR" and claimed the list was "one entry less wrong than before". A
comment recording a known defect instead of fixing it is precisely what the enum gate exists to
stop being possible, so removing it along with the drift is the right shape.

---

JiRaska:

PR #5977, opened 2026-08-20
fix(pid): publish the real PartyStatus and PartyType vocabularies

JiRaska:
Refs #5962, ADR-0048.

drops out once #5966 lands.

pid has **four** baselined drifts. This PR fixes **two** and deliberately leaves two, with their
baseline reasons corrected in place rather than deleted. That split is the point of the PR: two of
them are enum drift, and two of them are not.

## Fixed — isolated enum drift, with the database as the witness

|---|---|---|
| `ChangeStatusRequest.status` `[ACTIVE, SUSPENDED, CLOSED]` | `PartyStatus { ACTIVE, SUSPENDED, DECEASED, TERMINATED }` | `CLOSED` deserializes into `ChangeStatusRequest(status: PartyStatus)` and fails; `DECEASED`/`TERMINATED`, the two real terminal states, were unpublished |
| `RegisterIdentityRequest.partyType` `[NATURAL_PERSON, LEGAL_ENTITY]` | `PartyType { NATURAL_PERSON, LEGAL_ENTITY, SOLE_TRADER }` | sole traders could not be registered by any generated client |

Both request schemas match their Kotlin DTOs **property for property**
(`ChangeStatusRequest(status, reason)`; `RegisterIdentityRequest(partyId, partyType, givenName,
familyName, birthdate, birthplace, birthNumberRaw, ...)`), so the enum really is the only thing
that drifted.

**Which side is right — the domain, and `V1__init_pid.sql` says so directly:**

```sql
CONSTRAINT chk_party_status CHECK (status IN ('ACTIVE','SUSPENDED','DECEASED','TERMINATED')),
CONSTRAINT chk_party_type   CHECK (party_type IN ('NATURAL_PERSON','LEGAL_ENTITY','SOLE_TRADER')),
```

`CLOSED` is not merely unmapped in Kotlin, it is unstorable; `SOLE_TRADER` is unambiguously real.
`git log -S "CLOSED"` over `openbank-pid-service` finds it only in `openapi.yaml`, at every
revision, never in Kotlin or SQL. Adding `SOLE_TRADER` is purely additive and breaks nothing.

## Left baselined — the defect is a fictional schema, not a drifted enum

Reconciling these two enums would make the document *look* correct while it still describes an
endpoint nobody can call, and would delete the baseline entry that records the real problem. So
they stay, and the reason strings are rewritten to say what is actually wrong:

- **`UpdateKycRequest.kycStatus` `[PENDING, VERIFIED, REJECTED, EXPIRED]` — and there is no
  `kycStatus` property.** The Kotlin DTO is
  `UpdateKycRequest(kycLevel: KycLevel, amlRiskScore: AmlRiskScore, pepFlag, sanctionsFlag)`:
  two required enum fields, neither of them a status, plus two booleans. The spec's `verifiedAt`
  does not exist either.
  **This baseline entry is also a mis-pairing, and that is worth recording.** With no real
  counterpart to pair against, the matcher settled on `PresentationExchangeStore.Status
  { PENDING, COMPLETED, EXPIRED }` — an openid4vp store class with nothing to do with KYC — on the
  strength of `PENDING` and `EXPIRED` alone. The baseline said "undeclared `COMPLETED`", which
  would have been the wrong fix. Concrete evidence for the issue's own framing that the gate's
  pairing is a hypothesis; nothing is wrong with the matcher, but a finding is a place to start
  looking, not a verdict.
- **`CreatePartyRequest.partyType` `[INDIVIDUAL, LEGAL_ENTITY, SOLE_TRADER]`.** `INDIVIDUAL` has
  indeed never existed (see the CHECK constraint above), but it sits inside a schema declaring
  `legalName`, `tradingName`, `taxId`, `dateOfBirth`, `nationality` — **not one** of which is a
  property of `CreatePartyRequest(partyType, givenName, familyName, birthdate, nationalities,
  verificationSource, bankIdSub, birthNumberRaw, initialRole, onboardingChannel)`.

Both need a schema rewrite of the kind #5966 did for kyc's `UpdateCheckRequest`, and both deserve
their own PR with their own evidence rather than being smuggled into an enum reconciliation.

```
::notice::api-contract gate: openbank-pid-service: spec-only PR — reclassifying 1 breaking
document change(s) (first: removed the enum value `CLOSED` of the request property `status`)
as a CORRECTION. ... Requiring MINOR instead.
api-contract gate: openbank-pid-service: correction change, info.version 1.13.0 -> 1.14.0 — OK.
```

`1.13.0` read off live `origin/main` immediately before the bump. pid is not a money-path service.

## Verification

- `./gradlew :openbank-pid-service:ktlintCheck :detekt :test` — `BUILD SUCCESSFUL`, exit 0, all
  three executed. **174 tests, 0 failures, 0 errors, 1 skipped** — the skip is
  `PidPactBrokerProviderVerificationTest`, the broker-sourced class gated on `pactbroker.url`.
  Its `@PactFolder` sibling ran green.
  `25 paired enum(s) drift, all 25 baselined`, and fails on a stale entry in either direction, so
  the two removals are proven reconciled and the two retained entries are proven still real.

Baseline: **27 -> 25** for this PR alone.

### DIFF
--- a/.github/gates/gates.yaml
+++ b/.github/gates/gates.yaml
@@ -2186,6 +2186,42 @@ gates:
     run: |
       python3 .github/scripts/check-pact-provider-replay.py
 
+  # openapi enum vs the Kotlin enum behind it (issue #5895 / #5962, enforced).
+  #
+  # The axis the api-contract gate below structurally CANNOT see: oasdiff compares a spec to its
+  # own previous version, never to the implementation. So a spec enum hand-written once and never
+  # re-checked drifts forever and every gate stays green. openbank-kyc-service published
+  # `[IDENTITY, SANCTIONS, PEP, ADVERSE_MEDIA, SOURCE_OF_FUNDS]` against a domain
+  # `CheckType { IDENTITY, ADDRESS, PEP_SCREENING, SANCTIONS_SCREENING, ADVERSE_MEDIA }` from the
+  # day the file was written — three names misspelled, one real check unpublished, and one value
+  # that has never existed in the code. Its own PUT body was fiction in the same way.
+  #
+  # Pairing is by VALUE OVERLAP, not by name: nothing declares which Kotlin enum a spec enum
+  # serves, and matching on names would find almost none of them. An unpaired spec enum is not a
+  # finding (many are free-form vocabularies with no Kotlin type).
+  #
+  # RATCHET: the 27 drifts across 16 services that exist today are a BASELINE constant in the
+  # script (owned by #5962), so this is enforced from day one and number 28 is red at PR time. A
+  # baseline entry that no longer drifts also FAILS, so the list cannot rot into furniture — same
+  # contract as check-pact-provider-replay.py and check-adr-partial-followup.py above.
+  - id: openapi-enum-domain-drift
+    name: "openapi enum matches the domain enum it serves (issue #5962, enforced)"
+    group: api
+    mode: enforced
+    rationale: >
+      oasdiff compares a spec to its own previous version and never to the implementation, so a
+      spec enum hand-written once and never re-checked drifts forever with every gate green.
+      openbank-kyc-service published three misspelled check names, omitted a real check, and
+      advertised a value (SOURCE_OF_FUNDS) that has never existed in the code — from the day the
+      file was written until #5895, and 27 more pairings across 16 services still do (#5962).
+    review_after: '2027-02-20'
+    min_subjects: 100
+    selftest: |
+      python3 .github/scripts/check-openapi-enum-vs-domain.py --self-test
+    selftest_expect: pass
+    run: |
+      python3 .github/scripts/check-openapi-enum-vs-domain.py --enforce
+
   # ADR-0048 D5: api-contract gate — classify the API-contract bump from the
   # OpenAPI diff (oasdiff), independently of the release axis (version.txt):
   #   breaking  => info.version MAJOR (new URL major /api/v{N+1})
--- a/.github/scripts/check-openapi-enum-vs-domain.py
+++ b/.github/scripts/check-openapi-enum-vs-domain.py
@@ -1,0 +1,355 @@
+#!/usr/bin/env python3
+"""openapi-enum-domain-drift gate — does a spec enum still match the Kotlin enum behind it?
+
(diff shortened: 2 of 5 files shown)


JiRaska:
## Review — #5977 (pid)

Reviewed with `git diff origin/main...origin/fix/pid-openapi-enum-drift-5962`. **No change
requested.**

**The DB witness is exactly as claimed.** `openbank-pid-service/src/main/resources/db/migration/
V1__init_pid.sql`:

* line 50 — `CONSTRAINT chk_party_status CHECK (status IN ('ACTIVE','SUSPENDED','DECEASED','TERMINATED'))`
* line 49 — `CONSTRAINT chk_party_type CHECK (party_type IN ('NATURAL_PERSON','LEGAL_ENTITY','SOLE_TRADER'))`

So `CLOSED` is unstorable and `SOLE_TRADER` is real, settled without reading the Kotlin at all —
which is the point. Both new spec lists match the constraints value for value.

Note the trap this avoids and the PR navigates correctly: `openbank-party-service` has its **own**
`PartyStatus` (`Party.kt:17`) which *does* contain `CLOSED`, and its own `PartyType` which contains
`INDIVIDUAL`. A fleet-wide grep for either literal returns hits and reads like corroboration for
the spec. Scoping to `openbank-pid-service` is what distinguishes them, and this PR did that.

**The two entries left baselined are the right call, and this is the part I'd have flagged if it
had gone the other way.** Both remaining pid entries sit inside request schemas whose *properties*
do not match the DTO — `UpdateKycRequest` has no `kycStatus` at all, and the matcher paired it
against the unrelated openid4vp `PresentationExchangeStore.Status` on two coincidental values.
Reconciling the enum there would have made a fictional schema look authoritative. Re-reasoning the
baseline entry in place, rather than "fixing" it, is the correct outcome, and the new reasons say
so plainly enough that the next tranche will not repeat the analysis.

**Version.** `origin/main` reads `1.13.0`; `1.14.0` correct and free, no competing PR on this file.

**Pacts.** pid's replay is `PidPactFolderProviderVerificationTest` (`@PactFolder`, runs on a PR).
`CLOSED` appears in no pid pact, so the removal is not asserted anywhere.

---

JiRaska:

PR #5979, opened 2026-08-20
feat(referral): add governed fixed-reward MGM slice

### LINKED ISSUE #5943: feat: implement MGM fixed-reward referral vertical slice
## Goal
Implement the first governed MGM vertical slice described by ADR-0266: one versioned referral program, opaque invite token, fixed-reward request, one qualifying event, replay-safe attribution and an explicit ledger boundary contract.

## Required boundaries
- Keep referral attribution and incentive state outside `openbank-campaign-service`.
- Reject self-referral, duplicate attribution, expired programs and replayed/concurrent attribution.
- Keep the MGM lifecycle owner solely responsible for `REWARD_REQUESTED` and `REWARDED`; ledger/account owns only posting outcome and reversal.
- Do not mark a reward paid until the ledger boundary confirms it.
- Keep campaign exposure/conversion evidence distinct from reward state.
- Do not implement production monetary posting until the separate money-path ADR and threat model are approved.

## Required evidence
- real HTTP lifecycle: draft → independent publication → invite → attribution → qualification → reward request;
- concurrency/replay/idempotency tests for attribution and qualification;
- ledger boundary contract fixture for accepted/rejected/reversed outcomes, duplicate-reference rejection and retry;
- authorization, audit, retention/deletion and event contract replay;
- threat model and separate money-path ADR before any monetary posting implementation.

Architecture boundary: ADR-0266 ([PR #5941](https://github.com/JiRaska/open-bank-oss/pull/5941)).

JiRaska:
Closes #5943

## Summary
- add the first governed MGM referral vertical slice outside campaign-service
- implement versioned programs, four-eyes publication, opaque hash-only invites, self/expiry/replay/concurrency guards, and idempotent qualification
- emit contract-only qualified/reward-requested/outcome events and persist auditable lifecycle state
- add PostgreSQL migration, OpenAPI/AsyncAPI contracts, retention/governance metadata, and non-money threat model

## Evidence
- `./gradlew :openbank-referral-service:compileKotlin :openbank-referral-service:compileTestKotlin :openbank-referral-service:ktlintCheck` ✅
- `./gradlew :openbank-referral-service:test --tests '*ReferralRestContractIT'` ✅ (real HTTP + Testcontainers PostgreSQL; draft creation and maker self-approval rejection)

## Explicit follow-up before merge
- extend HTTP test through independent checker publication, invite issuance, attribution, qualification, replay and concurrent attribution
- add ledger-boundary accepted/rejected/reversed/duplicate-reference/retry fixture coverage
- wire reviewed GitOps deployment and event transport after independent review

No monetary posting or ledger credentials are included in this slice. A separate money-path ADR and threat model remain required before enabling monetary posting.


JiRaska:
Local verification on 914da6fb5:
- :openbank-referral-service:ktlintMainSourceSetCheck :ktlintTestSourceSetCheck — PASS
- :openbank-referral-service:test --tests com.openbank.referral.integration.ReferralRestContractIT — PASS (real HTTP + Testcontainers Postgres)

Added checker publish, invite, attribution, qualification, idempotent reward replay, duplicate invite idempotency assertions; fixed repository four-eyes comparison and null publish outcome mapping. GitHub PR CI has not started a run for the new head yet.

JiRaska:
Temporary lifecycle retrigger to restore the missing pull_request CI event; reopening immediately as draft.

JiRaska:
Lifecycle retrigger attempted: PR was closed and immediately reopened as draft. GitHub still reports 0 check suites / 0 workflow runs for head 914da6fb5; no ready-for-review or merge action taken.

JiRaska:

3-dot diff against `origin/main`; branch is 19 behind. A whole new released component on the money-adjacent side, so I read it against the money-path checklist rather than as a feature.

The maker/checker split on programs, the hashed invite token, the `unique(invite_id, qualification_event_id)` and `unique(reward_reference)` constraints, and the audit table are all the right instincts. Five findings, three of them blocking.

### 1. BLOCKING — every referral event is silently dropped, and the AsyncAPI spec says otherwise

`infrastructure/persistence/ReferralPersistence.kt:255-259`:

```
@ApplicationScoped class ContractReferralEventPublisher : ReferralEventPublisher {
    override suspend fun publish(event: ReferralEvent) {
        // Transport is deliberately contract-only until the event-schema PR.
    }
}
```

This is the only implementation of the port. `ReferralService.qualifyInvite` publishes `Qualified` and `RewardRequested` (`ReferralService.kt:187-206`) and `applyLedgerOutcome` publishes `RewardOutcome` — all three land in an empty method body. No return value distinguishes it, no metric counts it, no log line records it, and `openbank-referral-service/asyncapi.yaml` declares the channels as though they are on the wire.

This is, exactly, the shape the sibling PR **#5893 in this same batch** was written to prevent, and the one `PushResult.skipped()` cost this repo months of phantom deliveries: a not-configured adapter that is indistinguishable from a working one at every observation point. On a reward path it is worse — a downstream consumer that is supposed to book the payout will never be told a reward was requested, and nothing anywhere goes red.

Minimum fix before merge: make the no-op **loud and typed**. Give `publish` a result with its own `SKIPPED`/`NOT_CONFIGURED` value (never a boolean shared with success), emit a counter named for what it can establish (`referral_events_dropped_total`, not `_published_`), and log a `@Startup` warning saying no transport is wired. Better: land the outbox now (see #3).

### 2. BLOCKING — `.release-please-manifest.json` was not updated alongside `release-please-config.json`

`release-please-config.json:128-131` registers `openbank-referral-service`, and `openbank-referral-service/version.txt` is added, but `.release-please-manifest.json` has **zero** occurrences of `referral` on this branch (`git diff origin/main...` shows no change to that file at all). ADR-0029 rule 3 requires `version.txt`, `release-please-config.json` and the manifest in lockstep; the `release-registration consistency` gate exists for precisely this, and a component release-please does not know breaks its first release. Note also that this file is one where a *clean* git merge can silently drop a neighbouring entry — after adding it, re-count the entries against `origin/main`.

### 3. BLOCKING — no outbox: the reward row, its audit row and its event are three separate transactions

Every repository method wraps its own `Panache.withTransaction { ... }`. So `qualifyInvite` performs: `rewards.create` (tx 1) → `events.publish` (nothing) → `audit.append` (tx 2). A crash between them leaves a reward with no audit trail, and once a real transport replaces #1, an event that was never emitted for a reward that exists — or the reverse. There is no outbox entity and no `openbank.outbox.dispatch-enabled` in `application.yaml`, so the flag question is moot today, but the moment an outbox lands remember it defaults to **false** and must be set `true` in `application.yaml` or events never dispatch with no error at all.

This service creates money obligations. The transactional-outbox pattern is not optional here, and per the repo's own note only a real-DB IT driving the REST endpoint can prove the atomicity — a mocked repository cannot.

### 4. Idempotency is inconsistent, and decorative on the one endpoint that creates a reward

- `ReferralService.kt:107-109` — `issueInvite` treats a **reused** `Idempotency-Key` as `ReferralConflictException`. That inverts the contract: a client whose first request timed out retries with the same key and gets a 409 instead of the original invite. It should return the existing invite. As written, the `unique(idempotency_key)` column plus this branch make retry-after-timeout an error path rather than the safe path.
- `attributeInvite` and `qualifyInvite` both `validate(idempotencyKey.isNotBlank(), ...)` and then **never use the key again**. Dedupe on the reward path is `rewards.findByInviteAndEvent(invite.id, eventId)` (`:172`) — the caller-supplied `eventId`, not the key. So two deliveries of the same qualifying business event carrying two different `eventId`s create **two rewards for one qualification**, and the header that was supposed to prevent it is inert. Either key the reward dedupe on the `Idempotency-Key` as well, or drop the parameter and document that `eventId` is the idempotency key — but do not require a header the code ignores.
- `PanacheReferralInviteRepository.findByIdempotencyKey` (`ReferralPersistence.kt:175-177`) calls `.toDomain("stored")`, fabricating the literal string `"stored"` as the invite's token. Harmless while the result is only null-checked; a trap the moment anyone returns that object from the replay path above.

### 5. `applyLedgerOutcome` has no caller anywhere — every reward is terminal at `REWARD_REQUESTED`

`grep applyLedgerOutcome` across the branch returns exactly one hit: its own declaration (`ReferralService.kt:220`). No REST endpoint, no Kafka consumer, no scheduler. Combined with #1, the reward lifecycle is: created as `REWARD_REQUESTED`, event dropped, and nothing can ever move it to `REWARDED`, `RETRYABLE` or `REVERSED`. `RETRYABLE` in particular is a status with no retrier. The runbook and threat model should say this plainly, and `docs/runbooks/svc-referral.md` should not describe an operational flow the code cannot reach.

### Smaller
- `@RolesAllowed("ROLE_OPERATOR")` on the whole `ReferralResource`. `referrerPartyId` and `refereePartyId` come from the **request body**, so any `ROLE_OPERATOR` holder can attribute any party to any other. Note that this repo's Keycloak service accounts are classified `HUMAN` and hold `ROLE_OPERATOR` in at least one realm, so "operator only" is a weaker statement than it reads. Self-referral is blocked (`ReferralService.kt:145`) but there is no per-referrer cap, no reward ceiling, and no velocity control — for a fixed-reward MGM programme, farming is the primary abuse and the threat model should address it.
- `PanacheReferralInviteRepository.attribute` / `publish` / `outcome` use `requireNotNull(e)` after a status-filtered query (`ReferralPersistence.kt:181,140,224`). On a lost race — two concurrent attributions, exactly what idempotency exists for — that surfaces as `IllegalArgumentException` → 400 `"Required value was null"`, which tells the caller nothing. Throw `ReferralConflictException` with the real reason.
- No `@Scheduled` methods in this service, so the `HR000068` / plain-vs-`suspend` trap does not apply here. Worth remembering when the retrier for `RETRYABLE` lands: it must be a `suspend fun`.
- `CHANGELOG.md` is hand-written (`0.1.0`). Release-please generates that file; a hand-seeded one is the kind of derived artifact the repo rules say not to author.

### Checked and clean
- Invite tokens are 32 bytes from `SecureRandom`, URL-encoded, and only the SHA-256 hash is persisted (`ReferralPersistence.kt:161`); the audit row records `token_hash=`, not the token.
- `create` uses `persist` on application-assigned `@Id`s, which is correct for inserts; every state transition mutates a managed entity inside `withTransaction` and relies on dirty checking, so the `persist`-instead-of-`merge` footgun is not present.
- Actor attribution comes from `SecurityIdentity.principal.name` (`ReferralResource.kt:34`), never from the body. Correct.
- `@HeaderParam("Idempotency-Key") key: String?` is declared **nullable** with a `requireNotNull`-style guard in the body on all three endpoints — this avoids the 500-on-absent-header trap that `check-nonnull-jaxrs-params.py` gates. Good.
- Money columns are `numeric(19,4)`; `rewardAmount > BigDecimal.ZERO` and an ISO-4217 currency regex are validated at program creation. There is no currency check between program and ledger, but nothing consumes it yet.
- A `V1__referral.sql` with a rollback note, `migrate-at-start: true` and a literal `quarkus.datasource.jdbc.url` — both halves of `check-flyway-default-datasource.py` are satisfied.
- `quarkus.oidc.tls.verification: none` matches 47 other services on `main`; not a finding for this PR.

JiRaska:
## Blocking: the money-path event publisher had an empty body

`ContractReferralEventPublisher.publish` was:

```kotlin
override suspend fun publish(event: ReferralEvent) {
    // Transport is deliberately contract-only until the event-schema PR.
}
```

Every referral event — `Qualified`, `RewardRequested`, `RewardOutcome` — was **dropped**, with no metric, no log, and no return value a caller could branch on, while `asyncapi.yaml` declared `referral.qualified.v1` and `referral.reward.requested.v1` as if they carried traffic. On a money path this is exactly the `PushResult.skipped()` shape the fleet has already been burned by: a no-op that is indistinguishable from a working publisher, so no telemetry anywhere can disagree with "events are flowing". A consumer built against the asyncapi contract would have waited forever for messages that were never sent, and the only way to discover it would have been a customer noticing a missing reward.

## What changed (`fa3113f6b`)

The transport is still not wired — that remains the event-schema slice, and I did not want to half-wire Kafka into a review fix. What changed is that the unwired state can no longer be mistaken for delivery:

- **`ReferralEventPublisher.publish` now returns `ReferralPublishOutcome`** — `HANDED_TO_TRANSPORT` | `TRANSPORT_NOT_WIRED`. A skipped/disabled outcome gets its **own enum constant**, never a boolean shared with success. The success constant is named for what can actually be established (`HANDED_TO_TRANSPORT`), not `DELIVERED`.
- **`UnwiredReferralEventPublisher`** replaces the empty bean. It is `@Startup` — `@ApplicationScoped` is lazy, so a constructor warning on a rarely-called bean would never have reached a pod log — and it emits a boot WARN naming the declared-but-unserved channels, a per-event WARN carrying `type`/`eventId`/`programId`/`inviteId` (so a dropped reward is recoverable from logs), and a counter `openbank_referral_events_dropped_total{event_type,reason="transport_not_wired"}`. It emits **no success series at all**, so the absence of one cannot be read as delivery either.
- **`ReferralService` branches on the outcome** and appends an `EVENT_NOT_PUBLISHED` audit row when nothing left the process — the drop lands in the evidentiary record, not only in a log line.
- **`asyncapi.yaml`** now states in `info.description` that no producer exists in this build, so nobody builds a consumer expecting traffic.

## Negative control

Restoring the old behaviour under the new signature — `override suspend fun publish(event: ReferralEvent) = ReferralPublishOutcome.HANDED_TO_TRANSPORT`, i.e. the empty body reporting as if it had worked — turns **3 of 4 tests red**:

```
expected: TRANSPORT_NOT_WIRED
 but was: HANDED_TO_TRANSPORT
	at UnwiredReferralEventPublisherTest.kt:62

Expected size: 3 but was: 0 in: []
	at UnwiredReferralEventPublisherTest.kt:74

Expecting ArrayList: [] to contain only:
  ["openbank_referral_events_dropped_total"]
```

(The literal original empty body no longer compiles against the port at all, which is itself part of the fix.)

## Verified

`:openbank-referral-service:test` + `:quarkusBuild` with `--rerun-tasks`, green; JUnit XML read directly: `UnwiredReferralEventPublisherTest` 4 tests / **0 skipped** / 0 failures, `ReferralRestContractIT` 1 / 0 / 0. `ktlintCheck` + `detekt` green. `quarkusBuild` included so the CDI wiring of the new `@Startup` bean is actually validated.

Re-applied onto `f42cefdd1` after the branch was rebased mid-review.

**Follow-up for whoever wires the real transport:** `openbank.outbox.dispatch-enabled` defaults to `false` — set it `true` in `application.yaml` or the outbox rows are written and never dispatched, with no error and `attempt_count` stuck at 0.

PR #5982, opened 2026-08-20
fix(transaction): publish the full TransactionType and TransactionStatus vocabularies

JiRaska:
Refs #5962, ADR-0048.

Stacked on #5966 (which adds the gate and its `BASELINE`) and on the four service PRs already
open against #5962 (#5969 party, #5971 clearing, #5974 sca, #5977 pid) — all merged in, so their
diffs drop out as they land. **This PR's own work is two enum lines and one version line in
`openbank-transaction-service/src/main/resources/openapi.yaml`, plus two `BASELINE` entries.**

## What was wrong

Both drifts here are the *omission* shape: a real state exists, is storable, is written, and no
consumer could be written against it because the published document never named it.

|---|---|---|
| `GET /transactions` `status` filter `[PENDING, COMPLETED, REVERSED, FAILED]` | `TransactionStatus { PENDING, PROCESSING, COMPLETED, FAILED, REVERSED }` | `PROCESSING` — the in-flight state — is unfilterable from any generated client |
| `GET /transactions` `type` filter `[CREDIT, DEBIT, REVERSAL, FEE]` | `TransactionType { DEBIT, CREDIT, TRANSFER, FEE, INTEREST, REVERSAL, ADJUSTMENT }` | `TRANSFER`, `INTEREST` and `ADJUSTMENT` — three of the seven real types, including every interest posting — are unfilterable |

Neither is a value the code lacks, so nothing 400s; the defect is that a third of the type
vocabulary and one of the five statuses were invisible to every consumer of the contract.

## Which side is right — the database says so directly, for both

`V1__init_transactions.sql:28`:

```sql
CONSTRAINT chk_transactions_type CHECK (type IN (
    'DEBIT','CREDIT','TRANSFER','FEE','INTEREST','REVERSAL','ADJUSTMENT'
)),
CONSTRAINT chk_transactions_status CHECK (status IN (
    'PENDING','PROCESSING','COMPLETED','FAILED','REVERSED'
)),
```

Both CHECK constraints match the Kotlin enums value for value. This is the strongest form of the
witness the other PRs in this issue used: the missing values are not merely mapped in Kotlin, they
are the only values Postgres will accept, so a stored row can carry `ADJUSTMENT` or `PROCESSING`
today and the spec cannot describe it. There is no judgement call to make — the spec is simply an
incomplete transcription of a list that exists twice in the repo.

Nothing was renamed away, either: every value being added is present in the current Kotlin enum
and in the V1 constraint, so this is purely additive in both directions.

## No consumer is affected adversely

Widening a query-parameter enum cannot break an existing caller — every value the old document
allowed is still allowed. Checked anyway:

- **pacts** — the six interactions in `openbank-transaction-service`'s `@PactFolder` replay set
  are unaffected; the provider replay is green in this branch (see Verification).
- **admin-ui** — no call site pins either filter to the short list.

## Classification: `additive`, MINOR

`check-api-contract.py --base <live merge-base> --enforce`:

```
api-contract gate: openbank-transaction-service: additive change, info.version 1.12.0 -> 1.13.0 — OK (required >= MINOR).
```

Note this one does **not** need the `correction` reclassification the sibling PRs relied on: no
enum value is removed, so `oasdiff` reports nothing breaking at all. `1.12.0` read off live
`origin/main` immediately before the bump; major stays 1, so the ADR-0048 D2 URL-major invariant
is untouched.

run, not assumed.

## Verification

- `./gradlew :openbank-transaction-service:ktlintCheck :detekt :test` — `BUILD SUCCESSFUL`,
  exit 0. All three tasks executed (`detekt`, `ktlintCheck`, `test` all appear in the task list;
  none `UP-TO-DATE`).
- **140 tests, 0 failures, 0 errors, 1 skipped.** The single skip is
  `TransactionPactProviderVerificationTest`, the broker-sourced class gated on `pactbroker.url`,
  which does not run locally by design. Its `@PactFolder` sibling
  `TransactionPactFolderProviderVerificationTest` **did** run: 6 interactions, 0 failures,
  0 skipped.
  `18 paired enum(s) drift, all 18 baselined` (20 with the four open PRs landed, 27 originally),
  and it fails on a stale entry in either direction — so the two removals are proven genuinely
  reconciled rather than merely deleted.

Baseline: **20 -> 18**.

### DIFF
--- a/.github/gates/gates.yaml
+++ b/.github/gates/gates.yaml
@@ -2186,6 +2186,42 @@ gates:
     run: |
       python3 .github/scripts/check-pact-provider-replay.py
 
+  # openapi enum vs the Kotlin enum behind it (issue #5895 / #5962, enforced).
+  #
+  # The axis the api-contract gate below structurally CANNOT see: oasdiff compares a spec to its
+  # own previous version, never to the implementation. So a spec enum hand-written once and never
+  # re-checked drifts forever and every gate stays green. openbank-kyc-service published
+  # `[IDENTITY, SANCTIONS, PEP, ADVERSE_MEDIA, SOURCE_OF_FUNDS]` against a domain
+  # `CheckType { IDENTITY, ADDRESS, PEP_SCREENING, SANCTIONS_SCREENING, ADVERSE_MEDIA }` from the
+  # day the file was written — three names misspelled, one real check unpublished, and one value
+  # that has never existed in the code. Its own PUT body was fiction in the same way.
+  #
+  # Pairing is by VALUE OVERLAP, not by name: nothing declares which Kotlin enum a spec enum
+  # serves, and matching on names would find almost none of them. An unpaired spec enum is not a
+  # finding (many are free-form vocabularies with no Kotlin type).
+  #
+  # RATCHET: the 27 drifts across 16 services that exist today are a BASELINE constant in the
+  # script (owned by #5962), so this is enforced from day one and number 28 is red at PR time. A
+  # baseline entry that no longer drifts also FAILS, so the list cannot rot into furniture — same
+  # contract as check-pact-provider-replay.py and check-adr-partial-followup.py above.
+  - id: openapi-enum-domain-drift
+    name: "openapi enum matches the domain enum it serves (issue #5962, enforced)"
+    group: api
+    mode: enforced
+    rationale: >
+      oasdiff compares a spec to its own previous version and never to the implementation, so a
+      spec enum hand-written once and never re-checked drifts forever with every gate green.
+      openbank-kyc-service published three misspelled check names, omitted a real check, and
+      advertised a value (SOURCE_OF_FUNDS) that has never existed in the code — from the day the
+      file was written until #5895, and 27 more pairings across 16 services still do (#5962).
+    review_after: '2027-02-20'
+    min_subjects: 100
+    selftest: |
+      python3 .github/scripts/check-openapi-enum-vs-domain.py --self-test
+    selftest_expect: pass
+    run: |
+      python3 .github/scripts/check-openapi-enum-vs-domain.py --enforce
+
   # ADR-0048 D5: api-contract gate — classify the API-contract bump from the
   # OpenAPI diff (oasdiff), independently of the release axis (version.txt):
   #   breaking  => info.version MAJOR (new URL major /api/v{N+1})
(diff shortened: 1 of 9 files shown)


JiRaska:
## Review — #5982 (transaction)

Reviewed with `git diff origin/main...origin/fix/transaction-openapi-enum-drift-5962`.
**No change requested; two nits, one of which affects the changelog.**

**The DB witness verified, and it is the strongest kind.**
`openbank-transaction-service/src/main/resources/db/migration/V1__init_transactions.sql`:

* lines 29-31 — `CONSTRAINT chk_transactions_type CHECK (type IN ('DEBIT','CREDIT','TRANSFER','FEE','INTEREST','REVERSAL','ADJUSTMENT'))`
* lines 32-34 — `CONSTRAINT chk_transactions_status CHECK (status IN ('PENDING','PROCESSING','COMPLETED','FAILED','REVERSED'))`

(The PR body cites `:28`; the constraints are actually at 29 and 32 on current `origin/main` — a
line-number drift, not a substantive one.) Both match `Transaction.kt:72` / `Transaction.kt:82`
value for value. Purely additive in both directions: nothing removed, so no consumer can break.

**A third witness the PR does not claim:** `openbank-admin-ui/src/types/index.ts:7-8` declares
`TransactionType = 'DEBIT' | 'CREDIT' | 'TRANSFER' | 'FEE' | 'INTEREST' | 'REVERSAL' | 'ADJUSTMENT'`
and `TransactionStatus = 'PENDING' | 'PROCESSING' | 'COMPLETED' | 'FAILED' | 'REVERSED'` — the
domain lists verbatim. The one real UI consumer was written against the code, not this spec.

**Version.** `origin/main` reads `1.12.0`; `1.13.0` correct and free. No competing open PR touches
this file.

**Pacts.** `TransactionPactFolderProviderVerificationTest` is `@PactFolder`, so it replays on a PR.
Widening a query-parameter enum cannot invalidate an interaction anyway.

### Nit 1 — the commit message carries none of the evidence, and it *is* the changelog

The PR body is thorough. The commit body is, in full:

```
fix(transaction): publish the full TransactionType and TransactionStatus vocabularies

Refs #5962, ADR-0048.
```

Under the repo's own rule the commit message is the changelog, and `fix` + a path under
`openbank-transaction-service/` outside `exclude-paths` releases this component — so the released
changelog entry for a money-path service will be a bare subject line with no statement of which
side was wrong or why. Compare #5966, #5969, #5971, #5974 and #5977, all of which put the witness
in the commit body. The PR body evaporates at squash-merge; the commit body does not. Same applies
to #5986, #5990, #5992, #6015, #6016, #6017 and #6019 in this stack.

### Nit 2 — stale version comment

`openapi.yaml:8-10` still describes `Minor bump (1.11.0 -> 1.12.0)`. It was accurate on `main` and
this PR makes it stale.

**Money-path service — two approvals required**, and see the family note below on why approving a
downstream PR in this stack does not supply them.

---

PR #5986, opened 2026-08-20
fix(account): publish the real AccountStatus and WithdrawalProposalStatus vocabularies

JiRaska:
Refs #5962, ADR-0048.

#5962 (#5969 party, #5971 clearing, #5974 sca, #5977 pid) — all merged in, so their diffs drop out
as they land. **This PR's own work is two enum lines and one version line in
`openbank-account-service/src/main/resources/openapi.yaml`, plus two `BASELINE` entries.**

## What was wrong

|---|---|---|
| `AccountResponse.status` `[ACTIVE, FROZEN, CLOSED, PENDING]` | `AccountStatus { PENDING_ACTIVATION, ACTIVE, DORMANT, FROZEN, CLOSED }` | the server emits `PENDING_ACTIVATION` (the state every account is opened in) and `DORMANT`, neither of which the document declares; it has never emitted `PENDING`. A strict generated deserializer fails on a **newly opened account** |
| `GET /accounts/{id}/savings-withdrawal-proposals` `status` filter `[PENDING, APPROVED, REJECTED, CANCELLED]` | `WithdrawalProposalStatus { PENDING, APPROVED, REJECTED, CANCELLED, EXPIRED }` | `EXPIRED` — the state the 24h sweep writes — is unfilterable, so the owner's inbox cannot be queried for the proposals that lapsed |

`AccountResponse` matches its Kotlin DTO **property for property** (`id, accountNumber,
accountType, partyId, productId, currencyCode, status, openedAt, closedAt, goalName,
goalTargetMinorUnits, goalTargetDate, nickname`), and `status` is serialized straight off the
domain type (`val status: AccountStatus`), so the enum really is the only thing that drifted here
— this is not the pid-shaped case of an enum inside a fictional schema.

## Which side is right — the domain, on three independent witnesses

1. **The database.** `V1__init_accounts.sql:26`:
   ```sql
   CONSTRAINT chk_accounts_status CHECK (status IN (
       'PENDING_ACTIVATION','ACTIVE','DORMANT','FROZEN','CLOSED'
   )),
   ```
   `PENDING` is not merely unmapped in Kotlin — it is **unstorable**, so no account has ever been
   in it and no response can ever have carried it. `DORMANT` and `PENDING_ACTIVATION` are
   unambiguously real.
2. **A downstream spec already republishes the code spelling.**
   `openbank-customer-edge/src/main/resources/openapi.yaml:3912` declares `enum:
   [PENDING_ACTIVATION]`. A second document, in another service, agreeing with the code settles
   the spelling without a judgement call — the same class of witness #5966 used for kyc.
3. **admin-ui's type declaration is the enum verbatim.** `src/types/index.ts:6`:
   `export type AccountStatus = 'PENDING_ACTIVATION' | 'ACTIVE' | 'DORMANT' | 'FROZEN' | 'CLOSED'`,
   and `accounts/[id]/page.tsx:22` styles a `PENDING_ACTIVATION` pill. The one real UI consumer
   was written against the code, not against this spec.

For `EXPIRED` the witness is inside **this same document**: `ProposalResponse.status` (line 817)
already publishes `[PENDING, APPROVED, REJECTED, CANCELLED, EXPIRED]`. The spec therefore returns
a status it will not let you filter by — the omission is self-evidently the bug, not the value.
`V19__savings_proposal_expiry.sql` adds the sweep and notes in its rollback comment that "the
EXPIRED status value is application-side only; no enum type to revert".

## No consumer sends or expects the removed value

- **`PENDING`** — fleet grep finds no producer or consumer of `PENDING` as an *account* status;
  admin-ui's union type (above) does not contain it, so the UI could never have rendered it.
- **pacts** — nine pacts name account-service; the four interactions in its `@PactFolder` replay
  set carry none of these literals, and the replay is green in this branch (see Verification).
- Widening the proposal filter cannot break a caller: every previously accepted value still is.

## Classification: `additive`, MINOR

`check-api-contract.py --base <live merge-base> --enforce`:

```
api-contract gate: openbank-account-service: additive change, info.version 1.10.0 -> 1.11.0 — OK (required >= MINOR).
```

Worth stating plainly, since the sibling PRs on this issue all needed it: this one does **not**
invoke the `correction` reclassification. `PENDING` is removed from a *response* enum, which
`oasdiff` does not treat as breaking (the server promising fewer values cannot break a reader),
and the proposal filter is widened. So nothing breaking is reported at all and a plain MINOR
suffices. `1.10.0` read off live `origin/main` immediately before the bump; major stays 1, so the
ADR-0048 D2 URL-major invariant is untouched.

run, not assumed.

## Verification

- `./gradlew :openbank-account-service:ktlintCheck :detekt :test` — all three tasks executed, none
  `UP-TO-DATE`. `ktlintCheck` and `detekt` green.
- **252 tests, 0 failures, 0 errors, 1 skipped.** The single skip is
  `AccountEventPactProviderVerificationTest`, the broker-sourced class gated on `pactbroker.url`,
  which does not run locally by design. Its `@PactFolder` sibling
  `AccountPactFolderProviderVerificationTest` **did** run: 4 interactions, 0 failures, 0 skipped.
  - Noted for honesty: the *first* run of this suite reported 1 failure and **44 skips**, and it
    was environmental, not a regression — `QuarkusBindException: Port already bound: 8081` from a
    concurrent Gradle build on the same machine, which fails the `@QuarkusTest` boot and cascades
    into every downstream IT reporting as SKIPPED. Re-run serially with `--rerun-tasks`: 0
    failures, 1 skip. The skipped count, not the failure count, is what makes that distinguishable.
  `18 paired enum(s) drift, all 18 baselined`, and it fails on a stale entry in either direction,
  so the two removals are proven genuinely reconciled rather than merely deleted.
  (`SUBJECTS` goes 148 -> 147 rather than staying put: the gate keys a subject by
  `<service>:<sorted spec values>`, and the corrected proposal filter is now byte-identical to
  `ProposalResponse.status`, so the two collapse into one subject. Benign, and worth naming so the
  count is not read as a lost pairing.)

Baseline: **20 -> 18** for this PR alone.


JiRaska:
## Review — #5986 (account)

Reviewed with `git diff origin/main...origin/fix/account-openapi-enum-drift-5962`. **No change
requested.** This is the best-evidenced PR in the family and all three witnesses check out.

1. **DB.** `openbank-account-service/src/main/resources/db/migration/V1__init_accounts.sql:26-28`
   — `CONSTRAINT chk_accounts_status CHECK (status IN ('PENDING_ACTIVATION','ACTIVE','DORMANT','FROZEN','CLOSED'))`.
   `PENDING` is unstorable, so no account has ever been in it and no response can have carried it.
2. **A second service's spec.** `openbank-customer-edge/src/main/resources/openapi.yaml:3912`
   declares `enum: [PENDING_ACTIVATION]`. Confirmed at that exact line.
3. **admin-ui.** `openbank-admin-ui/src/types/index.ts:6` —
   `export type AccountStatus = 'PENDING_ACTIVATION' | 'ACTIVE' | 'DORMANT' | 'FROZEN' | 'CLOSED'`.
   Confirmed verbatim. The union does not contain `PENDING`, so the UI could never have rendered
   the value being removed.

And `Account.kt:95` is `AccountStatus { PENDING_ACTIVATION, ACTIVE, DORMANT, FROZEN, CLOSED }`,
byte-equal to the new spec list. `WithdrawalProposal.kt:16` is
`{ PENDING, APPROVED, REJECTED, CANCELLED, EXPIRED }`, likewise.

The in-document witness for `EXPIRED` is the sharpest bit of the analysis: `ProposalResponse.status`
in the *same file* already publishes `EXPIRED`, so the document returns a status it will not let
you filter by. That is self-refuting without needing any external source.

**Version.** `origin/main` reads `1.10.0`; `1.11.0` correct and free. No competing open PR touches
this file.

**Pacts.** `AccountPactFolderProviderVerificationTest` is `@PactFolder` and replays on a PR; no
committed pact carries `PENDING` as an account status.

**One thing I want to credit rather than nitpick:** the "noted for honesty" paragraph about the
first test run reporting 1 failure and **44 skips** from a `QuarkusBindException: Port already
bound: 8081`, and the observation that the *skipped* count is what distinguishes an environmental
cascade from a regression. That is exactly the right reading and it is the number nobody looks at.

**Money-path service — two approvals required**; see the family note below on the stack.

Nit, shared with the family: no stale version comment here (this file's comment is generic), so
nothing to fix — but the commit body is again bare `Refs #5962, ADR-0048.` while all the evidence
above lives only in the PR description, which does not survive the squash. See #5982 nit 1.

---

JiRaska:

| | `info.version` |
|---|---|
| merge-base | `1.10.0` |
| this branch | `1.11.0` |
| `origin/main` **now** | `1.11.0` |

```
git fetch origin main
git show origin/main:openbank-account-service/src/main/resources/openapi.yaml | grep -m1 '^  version:'
```

Something else shipped `1.11.0` for this service while this PR sat. The branch and `main` would both claim one API-contract version. Git will not flag it — the version line merges cleanly because the text is *identical*; a taken version is not a textual conflict. The green checks here ran nine days ago against the old base and cannot see it.

JiRaska:
**Measured: this branch *lowers* the account spec version below what `main` has shipped.**

```
openbank-account-service/src/main/resources/openapi.yaml
  origin/main : version: 1.12.0
  this branch : version: 1.11.0
```
Not a collision — a downgrade. `main` moved to 1.12.0 after this branch was cut (`behind_by` = 1225).
The `openapi-version-not-taken` gate landed on main 2026-08-31 06:44 (`a373867e9`, #7633).

This PR is also `CONFLICTING`. Test-merging `origin/main` into it (detached worktree, aborted afterwards)
gives exactly two conflicts:
```
.github/scripts/check-openapi-enum-vs-domain.py
openbank-account-service/src/main/resources/openapi.yaml
```
Resolve those, then set `info.version: 1.13.0` (re-read live `main` first — it may have moved again).

JiRaska:
Rebased onto current `main`. Two conflicts, both consequences of sitting **1414 commits** behind.

**`info.version`** — the branch proposed `1.11.0`, `main` is already at `1.12.0`. Taken while this
waited, so moved to the next free minor, **1.13.0**. Re-read after any competing account spec merge.

**`check-openapi-enum-vs-domain.py`** — resolved by taking main's version and then re-applying this
branch's actual contribution. Main's copy carries a far better-evidenced reason for the campaign
`channel` subset (it cites the two DB CHECK constraints disagreeing on purpose as an independent
witness) than the one-line note this branch had, so main's wording wins on that entry. What this
branch adds is the deletion of the two **account-service** baseline entries, which its spec fix
closes:

```yaml
- enum: [PENDING, APPROVED, REJECTED, CANCELLED]
+ enum: [PENDING, APPROVED, REJECTED, CANCELLED, EXPIRED]

- enum: [ACTIVE, FROZEN, CLOSED, PENDING]
+ enum: [PENDING_ACTIVATION, ACTIVE, DORMANT, FROZEN, CLOSED]
```

Checked in both directions rather than trusting the green:

```
with the baselines removed (this PR)   exit 0, SUBJECTS=161
with one put back                      ::warning:: baseline entry
                                       `openbank-account-service:APPROVED,CANCELLED,PENDING,REJECTED`
                                       no longer drifts — drop it
```

Worth noting that direction is a **warning**, not an error — so a baseline entry outliving its defect
would not have failed anything here. This PR removes them because the fix makes them false, not
because a gate forced it.

JiRaska:
## Heads-up: `info.version 1.13.0` has already shipped — this needs a re-bump before it merges

`openbank-account-service/openapi.yaml` here bumps `1.12.0 -> 1.13.0`. **#8466 merged that exact
version earlier today** (`909dee547a`), so live `main` is now at `1.13.0`:

```
$ git show origin/main:openbank-account-service/src/main/resources/openapi.yaml | grep -m1 '^  version:'
  version: 1.13.0
```

This branch last updated 2026-09-02, before that merge, so nothing on it could have noticed.

### Why nothing will flag it for you

- **No merge conflict.** Both sides write byte-identical text on that line, so git resolves it
  silently. A version line is the one field where "merged cleanly" and "still free" are different
  questions, and only the second one matters.
- **CI stays green.** `check-api-contract.py` classifies each PR against its own merge-base, so this
  branch is internally consistent and correctly classified. The collision only exists relative to
  main.

Merging as-is republishes a version that already exists, with different content behind it.

### Fix

```bash
```

then re-read the number off live `origin/main` rather than off this branch's base, and take the next
free one — `1.14.0` today, unless something lands first. Then re-run the gate the way
`.github/gates/gates.yaml` invokes it:

```bash
python3 .github/scripts/check-api-contract.py --base $(git merge-base HEAD origin/main) --enforce
```

### One more claimant to sequence against

**#8475** is also on this file and has already re-bumped to `1.14.0` after #8466 took `1.13.0`. So
whichever of the two of you merges second takes the number after the other — worth a glance at its
head before you pick.

Context and the general form of this in #8477. Not touching this branch; comment only.

JiRaska:
**The `openapi.yaml` version this branch claims has been taken on `main` while the PR was open. A rebase alone will not fix it — the number has to be re-picked too.**

Measured against `origin/main` just now; the `openapi-version-not-taken` gate is `mode: enforced` and resolves `origin/main` at run time, so this reds on the next run. It has not yet, because a `CONFLICTING` PR produces zero CI runs.

| spec | base when branched | this PR claims | `main` now |
|---|---|---|---|
| `openbank-account-service` | 1.12.0 | **1.13.0** | **1.14.0** |

1.13.0 shipped with #8466; 1.14.0 with #8712. The branch is 416 commits behind `main` and `CONFLICTING`, so the rebase is required regardless.

#8011 claims the same 1.13.0 from the same stale base — whichever of you rebases second takes the number after.

**Live claim map for these specs**, so the re-pick survives the other PRs in flight:

| spec | `main` now | already claimed by open PRs | next free |
|---|---|---|---|
| account-service | 1.14.0 | 1.13.0 (#5986, #8011 — both taken), 1.14.0 (#8475 — taken) | **1.15.0** |
| customer-edge | 1.50.0 | 1.51.0 (#8321), 1.52.0 (#8544) | **1.53.0** |
| delegation-service | 1.9.0 | 1.10.0 (#8310), 1.11.0 (#8318) | **1.12.0** |
| domestic-payment | 1.6.0 | 1.7.0 (#8274) | 1.8.0 |

Whoever rebases second takes the next number again — read the version off live `main` at the moment you rebase, not off this comment.

JiRaska:
Replaced by #9669. This branch's `check-openapi-enum-vs-domain.py` is an add/add conflict of 13 hunks against current `main` — that file was independently rewritten on both sides, and resolving it hunk-by-hunk risks silently dropping baseline entries. The kyc-service half of this branch has since landed on `main` independently (no kyc entry remains in the baseline). The account-service half was still genuinely unfixed on `main` (verified: `WithdrawalProposalStatus` has `EXPIRED`, the spec filter at openapi.yaml:485 did not) and is rebuilt fresh in #9669. Closing.

Issue #5987, opened 2026-08-20
fix(finrep): isBalanced is a hardcoded true — the balance check does not exist


JiRaska:
## Problem

`FinrepTemplate.isBalanced` is a published, contract-**required** field that no producer ever computes. Both mappers pass a hardcoded literal:

- `openbank-finrep-service/src/main/kotlin/com/openbank/finrep/domain/mapper/F0101Mapper.kt:44` — `isBalanced = true,`
- `openbank-finrep-service/src/main/kotlin/com/openbank/finrep/domain/mapper/F0200Mapper.kt:44` — `isBalanced = true,`
- `openbank-finrep-service/src/main/kotlin/com/openbank/finrep/domain/model/FinrepTemplate.kt:17` — `val isBalanced: Boolean = true,` (default also `true`)

`grep -rn "isBalanced = false" --include="*.kt" .` returns nothing across the whole repo.

F01.01 makes this structural rather than merely unimplemented: it **derives** equity as `assets - liabilities`, so the accounting identity `assets = liabilities + equity` holds algebraically by construction and cannot be observed to fail no matter what the ledger contains.

## Why it matters

`FinrepService`'s own KDoc says the value is "computed into `FinrepTemplate.isBalanced`, serialised, and never looked at again". The second half is now false — admin-ui's export gate reads it (#5904) — but the **first half was never true**. So a consumer that trusts `isBalanced` is trusting a constant.

Concretely, three things currently assert a balance check that does not exist:

1. `openapi.yaml` declares it `required` with the description "Whether the template's internal accounting identity holds".
2. `FinrepMetricsAdapter` tags the render metric `balanced=<value>` — a series that can only ever read `true`, i.e. an alert on it can never fire.
3. The admin-ui regulatory export gate blocks on `isBalanced === false`. That branch is correct against the published contract and is unit-tested, but **it cannot fire against the current producer** and must not be counted as live coverage. Its falsifiable, fires-today sibling is the `isDataGap` / `hasDataGaps` branch, which is genuinely computed by `C0100Mapper`.

This is the same shape as the "successful no-op shares a boolean with real success" and "gauge seeded from a sentinel" families already in `CLAUDE.md`: a control whose reported state is a constant, with nothing anywhere disagreeing.

## Acceptance criteria

[...]

Issue #5914, opened 2026-08-20
finrep: implement XBRL/DPM generation and ČNB submission


JiRaska:
## Problem

`openbank-finrep-service` can expose bounded structured FINREP/COREP preview data, but it does not generate EBA XBRL/DPM output and has no authenticated ČNB submission channel. Closed #605 only delivered the C 01.00 structural COREP increment; it is not the implementation tracker for this remaining capability.

## Required delivery

1. Pin and govern the applicable EBA/ČNB taxonomy version and map each supported template to the official taxonomy.
2. Produce schema-valid XBRL/DPM instance documents from source-ledger data, preserving explicit data-gap and reconciliation findings.
3. Validate generated output against the taxonomy and add regression fixtures for FINREP and COREP templates.
4. Implement an authenticated regulator transport adapter using deployment secret references only; no credentials in application code, UI, test fixtures, or repository history.
5. Provide operator-visible submission status, acknowledgement/reference, retry policy, audit trail, and failure diagnostics.
6. Prove a sandbox/acceptance submission with a regulator-approved test identity before any production enablement.

## Non-negotiable safeguards

- Never submit an unbalanced return or one with unresolved required data gaps.
- A preview/export must state whether it is a local structured view, a validated XBRL artifact, or an accepted regulator submission.
- Keep the submission path off by default until the real credential/onboarding evidence exists.

## Dependencies

- Taxonomy and regulator-interface specification.
- ČNB sandbox/onboarding and authenticated test identity.
- Capital/equity source data needed for substantive COREP values.

Refs ADR-0097, #605, #530

JiRaska:
Authoritative-source audit (2026-08-26):

- EBA Reporting Framework 4.2 applies from Q4 2025 and publishes DPM 4.2.1 / XBRL taxonomy 4.2.0.0.
- Official hotfix package: https://www.eba.europa.eu/sites/default/files/2026-01/b54d2c74-877c-4195-a32a-6591253d8b0f/taxo_package_4.2_hotfix.zip
- Downloaded package SHA-256: `0bf9e33720de1472e809417999cfd29e4b5eadb31750c7770622e502590bbdd0`.
- Exact module entry points for the currently implemented templates:
  - FINREP F 01.01 / F 02.00: `.../fws/finrep/4.2/mod/finrep9.xsd`
  - COREP C 01.00: `.../fws/corep/4.2/mod/corep_of.xsd`
[...]

PR #5988, opened 2026-08-20
feat(admin-ui): distinguish trace states and block incomplete regulatory exports

JiRaska:
## What

Two of the nine acceptance criteria in #5904 — **"Distinguish Trace Explorer loading, empty, and failed states"** and **"Block regulatory exports when data is incomplete, missing, or unbalanced"**. The other seven are untouched and remain open; see *Deliberately out of scope* below.

They belong together because they are one defect class: **a surface that renders identically whether or not it has evidence**. The issue's own framing is that operator surfaces "look empty or healthy without enough evidence" — the UI form of a control that cannot report its own absence. An export that succeeds on incomplete data is the same defect with worse consequences, because a regulatory filing is the artefact people act on.

## Trace Explorer — there were not three states

| State | Before | After |
|---|---|---|
| **Loading** | No branch at all. First paint fell through to the grid and rendered an empty `Recent traces ()` card — byte-for-byte what a successful-but-empty search renders. | A dedicated `role="status"` panel: *"Loading traces from Tempo…"* |
| **Empty** | `DataUnavailable kind="no_data"` | unchanged |
| **Failed** | `DataUnavailable kind="unreachable" / "error"` for the *list*; but a failed **span** fetch did `setSpans([])`, so a 502 from Tempo rendered the *"No data yet"* panel. | Span failures get their own `spansUnavailable` state and render the failure panel. |

A refresh over an existing list deliberately does **not** re-enter the loading branch — `traces` is still populated, so the list stays put and only the button spins.

### Why the empty state matters here specifically

`openbank-admin-ui` and `openbank-app` currently send **no traces at all** — neither appears in Tempo and `traces_spanmetrics_calls_total` has no series for them (#5735). So *empty is the honest state today*, which is precisely why a failure hidden inside it would never have been noticed.

Browser instrumentation is proposed in **#5847**, which is not merged and whose deploy hop is unverified. This PR **touches none of its files** — #5847 changes `next.config.mjs`, `layout.tsx`, `src/lib/telemetry/rum.ts`, `src/components/telemetry/RumScreenTracker.tsx`, a telemetry API route and two infra manifests; this PR changes the traces page, the regulatory page and adds one lib module plus two test files. No overlap, no contradiction: when #5847 lands and traces start arriving, this page stops showing the empty state on its own.

## Regulatory export — what "incomplete" resolves to in the data

**The codebase already had the notion; nothing is invented here.** `openbank-finrep-service`'s `openapi.yaml` declares these **required**:

- `CorepCell.isDataGap` — *"True when the value could NOT be honestly derived from data available in the platform today ... MUST NOT be read as an attested zero balance."*
- `CorepTemplate.hasDataGaps` — derived, true when any cell is a gap.
- `FinrepTemplate.isBalanced` — whether the accounting identity holds.

The admin-ui `RegulatoryTemplate` interface already carried all three. They were rendered only as **suffix text in the export rows** (`· DATOVÁ MEZERA`) and an amber font colour. Nothing blocked anything.

Before this change the CSV/JSON buttons were disabled **only** while `status === 'loading'`. So an operator could export from the `unavailable` and `unsupported` states and get a file whose every value cell read `Zdroj dat není dostupný` — with the same filename shape, the same headers and the same columns as a real return.

`evaluateExportReadiness` now blocks, most-fundamental reason first:

| Reason | Meaning | Fires today? |
|---|---|---|
| `not_loaded` | idle / in flight | yes |
| `no_data_source` | catalogue report with no service wired (6 of 8 reports) | yes |
| `source_unavailable` | the fetch to finrep-service failed | yes |
| `missing_cells` | a template rendered zero cells — malformed, the contract renders every row always | yes |
| `data_gaps` | `isDataGap` / `hasDataGaps` | **yes — the load-bearing one** |
| `unbalanced` | `isBalanced === false` | **no — see the gap below** |

An unknown future `PreviewData` variant defaults to **blocked**, so a new state cannot fall through into producing a regulatory file. The guard runs in the export handlers as well as on the buttons, because a `disabled` attribute is an affordance and not a control — there is a test that clicks past it.

### Named gap: `isBalanced` is a hardcoded `true` — filed as #5987

I did not assume the field worked. Both producers pass a literal:

- `F0101Mapper.kt:44` → `isBalanced = true,`
- `F0200Mapper.kt:44` → `isBalanced = true,`
- `FinrepTemplate.kt:17` → `val isBalanced: Boolean = true,`

`grep -rn "isBalanced = false" --include="*.kt" .` returns nothing repo-wide. F01.01 makes it structural: it **derives** equity as `assets - liabilities`, so the identity holds algebraically and cannot be observed to fail. `FinrepService`'s KDoc calls the value *"computed ... and never looked at again"* — the second half is now false, but the first half was never true.

So the `unbalanced` branch is a real check against the **published contract** that the **current producer can never trigger**. I kept it (the contract permits `false`, and a UI ignoring it would silently export an unbalanced return the day the producer starts computing the field) but it is **not counted as live coverage**, it is documented as such in the module header, and the underlying defect is filed as **#5987** rather than left as a sentence in a PR body.

## Deliberately out of scope

- **The other seven criteria of #5904** — test inventory, KYC name resolution, Customer 360 context, compliance-pack content, Temporal diagram labelling, and charter-backed agent identity in the approval queue. Separate work; this PR says `Refs`, not `Closes`.
- **XBRL/DPM generation and authenticated ČNB submission (#5914)**, per the issue's explicit boundary. **No submit path, fake or real, is added** — the existing footer text still says transmission is not connected, and the only thing this PR does to the export path is make it refuse more often.

## Verification — each assertion falsified

Not "it renders". Each state is asserted both for what it shows **and** what it must *not* show, then each fix was reverted to confirm the test actually detects its absence:

| Falsification | Result |
|---|---|
| Remove the trace-list loading branch (fall through to the grid, as before) | 🔴 `LOADING: shows an in-flight indicator…` fails |
| Restore `setSpans([])` on a failed span fetch | 🔴 `renders the failure panel, not "no data", when the span fetch 502s` fails |
| Remove the export gate (buttons back to `disabled={loading}`, handlers unguarded, no banner) | 🔴 3 tests fail, including `produces NO file when a blocked export is invoked anyway` |

There is also a test asserting the empty and failed renders are not string-equal to each other, so a future change that merges them is caught even if both individually still match their patterns.

One trap worth recording: the surface tests initially clicked whichever preview button came first, which is a catalogue-only report — every case blocked with `no_data_source` and the `data_gaps` / `unbalanced` paths were never reached. The helper now scopes to the CNB FINREP card by name and throws if that card's markup changes, rather than silently testing nothing.

**Gate results** (run in the worktree, `npm test` so the `pretest` hook bakes `governance.json` / `catalog.json`):

- `npm test` — **219 files / 1574 tests green**. First run showed 2 unrelated failures (`campaign-experience-preview`, `iaops-agent-card`), both `Test timed out in 5000ms`, both in files this PR does not touch and does not import; they pass in isolation and the second full run was clean. Load-induced flakes, recorded here rather than hidden.
- `npx tsc --noEmit` — clean (exit 0).
- `eslint` on all five touched files — **0 errors**. The 2 warnings on the traces page (`no-unused-eslint-disable`, `set-state-in-effect`) are on pre-existing lines and reproduce identically on clean `origin/main` with my changes stashed.

Refs #5904

### DIFF
--- a/openbank-admin-ui/src/app/observability/traces/page.tsx
+++ b/openbank-admin-ui/src/app/observability/traces/page.tsx
@@ -84,6 +84,7 @@ export default function TraceExplorerPage() {
   const [selected, setSelected] = useState<string | null>(null)
   const [spans, setSpans] = useState<FlatSpan[] | null>(null)
   const [spansLoading, setSpansLoading] = useState(false)
+  const [spansUnavailable, setSpansUnavailable] = useState<UnavailableKind | null>(null)
   const [lastRefresh, setLastRefresh] = useState<Date | null>(null)
 
   const loadTraces = useCallback(async () => {
@@ -118,14 +119,21 @@ export default function TraceExplorerPage() {
   const openTrace = useCallback(async (traceID: string) => {
     setSelected(traceID)
     setSpans(null)
+    setSpansUnavailable(null)
     setSpansLoading(true)
     try {
       const res = await fetch(`/api/tempo/api/traces/${traceID}`, { signal: AbortSignal.timeout(8000) })
-      if (!res.ok) { setSpans([]); return }
+      // A failed span fetch used to `setSpans([])`, which rendered the same "No data yet" panel
+      // as a trace that genuinely carries no spans — so a 502 from Tempo read to the operator as
+      // "this trace is empty". Keep the two apart (issue #5904).
+      if (!res.ok) {
+        setSpansUnavailable(res.status === 502 ? 'unreachable' : 'error')
+        return
+      }
       const json = await res.json()
       setSpans(flattenTrace(json))
     } catch {
-      setSpans([])
+      setSpansUnavailable('unreachable')
     } finally {
       setSpansLoading(false)
     }
@@ -150,7 +158,27 @@ export default function TraceExplorerPage() {
           </button>}
         />
 
-        {unavailable && !traces?.length ? (
+        {/*
+          Three distinct states, never collapsed into one another (issue #5904):
+          LOADING  — the first fetch is still in flight and we have nothing yet. Before this
+                     branch existed the page fell straight through to the grid and rendered an
+                     empty "Recent traces ()" card, which is byte-for-byte what a successful
+                     but empty search also renders. An operator could not tell "still asking"
+                     from "Tempo holds no traces".
+          EMPTY    — the search succeeded and returned zero traces (`kind: 'no_data'`). This is
+                     currently the HONEST state for admin-ui and openbank-app, which emit no
+                     spans at all (#5735); browser instrumentation is proposed in #5847.
+          FAILED   — the search could not be answered (`unreachable` / `error`).
+          A refresh over an existing list deliberately does NOT re-enter the loading branch —
+          `traces` is still populated, so the list stays put and only the button spins.
+        */}
+        {loading && !traces && !unavailable ? (
+          <div className="card" data-testid="trace-list-loading" role="status" aria-live="polite"
+            style={{ display: 'flex', alignItems: 'center', justifyContent: 'center', gap: '8px', padding: '48px 0', color: 'var(--text-tertiary)', fontSize: '13px' }}>
+            <RefreshCw size={15} className="spin" aria-hidden="true" />
+            {t('Načítám trasy z Tempa…', 'Loading traces from Tempo…')}
+          </div>
+        ) : unavailable && !traces?.length ? (
           <DataUnavailable
             kind={unavailable.kind}
             service="tempo (observability)"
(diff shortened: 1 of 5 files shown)


JiRaska:
Heads-up from #5987, now open as #6010: the `unbalanced` branch of `evaluateExportReadiness` **becomes live**.

Both FINREP producers now compute `isBalanced` from the double-entry identity of the trial balance
they were rendered from (`Σ net == 0` per currency, over all five account types), and F01.01 no
longer derives equity as `assets − liabilities` — which is what made the old flag unfalsifiable. A
negative control confirms it: reverting the mapper to `isBalanced = true` turns five tests red.

Nothing in this PR needs to change — the branch was contract-correct all along and #6010 touches no
admin-ui file. But when #5988 lands, the `⚠ KNOWN GAP, deliberately not papered over — tracked as
#5987` block in `src/lib/regulatory/exportReadiness.ts` is stale in both directions: `isBalanced` is
computed, and `unbalanced` should no longer be described as unable to fire against the current
producer. Whichever of the two merges second should retire that block, and `data_gaps` stops being
the *only* fires-today reason.

Also relevant if you compare against fixtures: F01.01 r380 and F02.00 r010 change sign. They used to
sum ledger's raw `totalDebit − totalCredit` net straight into the row, so against real ledger data
liabilities and income reported as negative numbers; they are now positive magnitudes.

Refs #5987, #6010

PR #5992, opened 2026-08-20
fix(sepa-instant): publish the real SctInstStatus vocabulary and reconcile the payment schemas

JiRaska:
Refs #5962, ADR-0048.

#5962 (#5969, #5971, #5974, #5977) — all merged in, so their diffs drop out as they land. **This
PR's own work is `openbank-sepa-instant/src/main/resources/openapi.yaml` and one `BASELINE`
entry.**

## What was wrong — the baselined enum

|---|---|---|
| `SctInstPaymentResponse.status` `[SUBMITTED, ACCEPTED, SETTLED, REJECTED, RECALLED]` | `SctInstStatus { PENDING, PROCESSING, SETTLED, REJECTED, TIMEOUT, RECALLED }` | the server emits `PENDING`, `PROCESSING` and `TIMEOUT` — including the status of **every** payment on creation — and the document declares none of them. It has never emitted `SUBMITTED` or `ACCEPTED`. A strict generated deserializer fails on the 201 response of a successful submit |

`status` is `p.status.name` straight off the domain enum (`SctInstResource.kt:114`), so there is no
translation layer that could have absorbed the difference.

### Which side is right — the domain, on two independent witnesses

1. **The service's own API docs already republish the code vocabulary and flag the spec as wrong.**
   `docs/03-api.en.md:61` gives `status` as
   `PENDING`/`PROCESSING`/`SETTLED`/`REJECTED`/`TIMEOUT`/`RECALLED`, and line 96 says outright that
   the checked-in `openapi.yaml` "uses a different status enum
   (`SUBMITTED/ACCEPTED/SETTLED/REJECTED/RECALLED`)" while "the running service uses the
   `PENDING/PROCESSING/SETTLED/REJECTED/TIMEOUT/RECALLED` enum". Someone wrote the drift down
   instead of fixing it — the same pattern #5971 found in clearing and #5990 in sepa-payment.
2. **A downstream spec already publishes the code list verbatim.**
   `openbank-customer-edge/src/main/resources/openapi.yaml:1702`:
   `status: {type: string, enum: [PENDING, PROCESSING, SETTLED, REJECTED, TIMEOUT, RECALLED]}`.
   A second document in another service agreeing with the code settles the spelling without a
   judgement call.

`SUBMITTED` and `ACCEPTED` were never renamed — they never existed:
`git log --all -S "ACCEPTED" --name-only -- openbank-sepa-instant` and the same for `SUBMITTED`
return **no** `.kt` or `.sql` file at any revision. There is no CHECK constraint on
`sct_inst_payments.status`, so nothing stored has to change and no migration is needed.

No consumer sends either value: admin-ui reads `status` for a colour pill only, and the single
pact naming this service (`openbank-sepa-instant-openbank-transaction-service.json`) has
sepa-instant as the **consumer**, so no provider replay touches these routes.

## Three further drifts found by hand, which the gate structurally cannot see

The issue notes that `27` is a floor rather than a census, and #5974 showed why: the gate pairs a
spec enum to a Kotlin enum by value overlap, so a *property* defect has no enum to pair and is
invisible. Fixing the enum alone here would have polished a schema that still describes a
different response — the split #5977 drew for pid. All three are in the same file, all three are
settled by reading the DTO the resource populates directly, and all three are fixed here:

1. **`SctInstPaymentResponse` shared only 5 of its 10 property names with the DTO.** The Kotlin
   type is `SctInstPaymentResponse(paymentId, status, debtorIban, creditorIban, amount, currency,
   endToEndId, executionTimeoutAt, settledAt, createdAt)`, populated field-for-field by
   `SctInstResource.toResponse()`. The document declared `id` (the field is `paymentId`),
   `debtorAccountId`, `creditorName` and `submittedAt` — **none of which the server ever sends** —
   spelled `currency` as `currencyCode`, and omitted `debtorIban`, `executionTimeoutAt` and
   `createdAt` entirely. Rewritten to match, `required` included.
2. **`SubmitSctInstRequest` omitted three properties that are non-nullable in Kotlin**, so a body
   built from the published document **cannot deserialise**: `idempotencyKey`, `debtorIban` and
   `debtorName`. (The `Idempotency-Key` header overrides the body field when present but does not
   make it optional — `SctInstResource.submit` reads `body.idempotencyKey` as the fallback, which
   requires the property to have bound.) It also spelled `currency` as `currencyCode`. Same
   defect shape as the `UpdateCheckRequest` rewrite in #5966.
3. **`GET /sepa-instant/debtor/{id}` published `limit`/`offset`; the resource takes `page`/`size`**
   (`@QueryParam("page") @DefaultValue("0")`, `@QueryParam("size") @DefaultValue("20")`). A client
   passing `limit=50` gets silently ignored parameters and the default page of 20 — the quiet
   failure, not a 400.

`docs/03-api.en.md:96` independently lists all three ("lists the submit body without
`idempotencyKey`/`debtorIban`/`debtorName` ... declares the debtor-list pagination as
`limit`/`offset` ... The running service uses `page`/`size`"), so this is not my reading of the
source alone.

- **`RecallRequest.reason` `[FRAUD, DUPLICATE, WRONG_AMOUNT, WRONG_BENEFICIARY]`.** The Kotlin type
  is `RecallRequest(val reason: String)` — a free string, passed through to the use case
  unvalidated. There is no Kotlin enum to reconcile against, which is exactly why the gate does not
  pair it and correctly does not report it. Whether the four values are an intended API constraint
  that the code fails to enforce, or spec fiction, is a product decision rather than a
  reconciliation, so it stays as it is.
- The stale local server port `8111` (the service listens on 8127), also named in the docs note.
  Cosmetic, and unrelated to the contract.

```
::notice::api-contract gate: openbank-sepa-instant: spec-only PR — reclassifying 3 breaking
document change(s) (first: added the new required request property `debtorIban`) as a CORRECTION.
No other file in openbank-sepa-instant changed, so the served contract is unchanged and a MAJOR
bump would violate the ADR-0048 D2 URL-major invariant. Requiring MINOR instead.
api-contract gate: openbank-sepa-instant: correction change, info.version 1.4.0 -> 1.5.0 — OK.
```

Worth naming which half is breaking: the *response* changes (removing enum values and properties
the server never sent) are not breaking to `oasdiff` at all — a server promising less cannot break
a reader — and classified as `additive` on their own. It is the **request** half, item 2 above,
that reports `new-required-request-property`, and the clients that "break" are ones generated from
a document that could never have produced a deserialisable body. Same class as the aml precedent
(#2312/#2313) and kyc (#5966).

`1.4.0` read off live `origin/main` immediately before the bump; major stays 1, so the ADR-0048 D2
URL-major invariant holds and no URL moves.
run, not assumed.

## Verification

- `./gradlew :openbank-sepa-instant:ktlintCheck :detekt :test --rerun-tasks` — `BUILD SUCCESSFUL`,
  exit 0. `detekt`, `ktlintCheck` and `test` all executed; none `UP-TO-DATE`.
- **70 tests, 0 failures, 0 errors, 0 skipped.** No skip to explain: this service has no
  broker-gated provider class, because it is not a pact provider.
  `19 paired enum(s) drift, all 19 baselined`, and it fails on a stale entry in either direction,
  so the removal is proven genuinely reconciled rather than merely deleted.

Baseline: **20 -> 19** for this PR alone.

### DIFF
--- a/.github/scripts/check-openapi-enum-vs-domain.py
+++ b/.github/scripts/check-openapi-enum-vs-domain.py
@@ -103,8 +103,6 @@ BASELINE: dict[str, str] = {
         "not an enum fix. The `Status` pairing is coincidental.",
     "openbank-sanctions-service:CNB_DOMESTIC,EU_CONSOLIDATED,FATF_HIGH_RISK,HM_TREASURY,OFAC_SDN,UN_CONSOLIDATED":
         "#5962 — SanctionsListType: undeclared PEP_GLOBAL",
-    "openbank-sepa-instant:ACCEPTED,RECALLED,REJECTED,SETTLED,SUBMITTED":
-        "#5962 — SctInstStatus: spec-only ACCEPTED/SUBMITTED; undeclared PENDING/PROCESSING/TIMEOUT",
     "openbank-sepa-payment:COMPLETED,PROCESSING,RECALLED,REJECTED":
         "#5962 — SepaPaymentStatus: spec-only RECALLED; undeclared CANCELLED/RECEIVED/RETURNED/VALIDATED",
     "openbank-sepa-payment:COMPLETED,PENDING,PROCESSING,RECALLED,REJECTED":
--- a/openbank-sepa-instant/src/main/resources/openapi.yaml
+++ b/openbank-sepa-instant/src/main/resources/openapi.yaml
@@ -3,7 +3,7 @@
 openapi: 3.1.0
 info:
   title: SEPA Instant (SCT Inst) Service API
-  version: 1.4.0
+  version: 1.5.0
   description: >-
     SEPA Instant Credit Transfer — sub-10s settlement, recall support. On submit, debtor and creditor
     names are synchronously screened against the sanctions lists (ADR-0032): a clean payment proceeds
@@ -95,16 +95,16 @@ paths:
           schema:
             type: string
             format: uuid
-        - name: limit
-          in: query
-          schema:
-            type: integer
-            default: 20
-        - name: offset
+        - name: page
           in: query
           schema:
             type: integer
             default: 0
+        - name: size
+          in: query
+          schema:
+            type: integer
+            default: 20
       responses:
         '200':
           description: List of payments
@@ -213,26 +213,38 @@ components:
 
   schemas:
     SubmitSctInstRequest:
+      # Field-for-field with SubmitSctInstRequest in SctInstDtos.kt. `idempotencyKey`,
+      # `debtorIban` and `debtorName` are non-nullable there, so a body omitting any of them
+      # fails to deserialise; the `Idempotency-Key` header, when present, overrides the body
+      # field but does not make it optional (SctInstResource.submit reads `body.idempotencyKey`).
       type: object
-      required: [debtorAccountId, creditorIban, creditorName, amount, endToEndId]
+      required: [idempotencyKey, debtorAccountId, debtorIban, debtorName, creditorIban, creditorName, amount, endToEndId]
       properties:
+        idempotencyKey:
+          type: string
         debtorAccountId:
           type: string
           format: uuid
+        debtorIban:
+          type: string
+        debtorName:
+          type: string
         creditorIban:
           type: string
         creditorName:
           type: string
         creditorBic:
           type: string
+          nullable: true
         amount:
           type: number
-        currencyCode:
+        currency:
           type: string
           default: EUR
-        endToEndId:
-          type: string
         remittanceInfo:
+          type: string
+          nullable: true
+        endToEndId:
           type: string
 
     RecallRequest:
@@ -244,34 +256,38 @@ components:
           enum: [FRAUD, DUPLICATE, WRONG_AMOUNT, WRONG_BENEFICIARY]
 
     SctInstPaymentResponse:
+      # Field-for-field with SctInstPaymentResponse in SctInstDtos.kt, which SctInstResource
+      # .toResponse() populates directly from the SctInstPayment aggregate.
       type: object
+      required: [paymentId, status, debtorIban, creditorIban, amount, currency, endToEndId, createdAt]
       properties:
-        id:
+        paymentId:
           type: string
           format: uuid
-        endToEndId:
+        status:
           type: string
-        debtorAccountId:
+          enum: [PENDING, PROCESSING, SETTLED, REJECTED, TIMEOUT, RECALLED]
+        debtorIban:
           type: string
-          format: uuid
         creditorIban:
-          type: string
-        creditorName:
           type: string
         amount:
           type: number
-        currencyCode:
+        currency:
           type: string
-        status:
+        endToEndId:
           type: string
-          enum: [SUBMITTED, ACCEPTED, SETTLED, REJECTED, RECALLED]
-        submittedAt:
+        executionTimeoutAt:
           type: string
           format: date-time
+          nullable: true
         settledAt:
           type: string
           format: date-time
           nullable: true
+        createdAt:
+          type: string
+          format: date-time
 
     ApiError:
       type: object


JiRaska:
## Review — #5992 (sepa-instant)

Reviewed with `git diff origin/main...origin/fix/sepa-instant-openapi-enum-drift-5962`.
**Needs a change: `info.version 1.5.0` is already claimed by another open PR.** The content is
otherwise correct and well-evidenced. Money-path — two approvals.

### ⚠️ Version collision with #5694 — and it is the silent kind

**#5694** (`feat(sepa-instant): expose pending four-eyes approvals via approval inbox`, branch
`feat/sepa-instant-approval-inbox-exposure`, open, last updated 2026-08-20) edits the **same**
`openbank-sepa-instant/src/main/resources/openapi.yaml` and makes the **same** bump:

```
-  version: 1.4.0
+  version: 1.5.0
```

Both PRs change one line, from the same value to the same value. Git sees identical text on both
sides and merges without a conflict — so whichever lands second inherits `1.5.0` describing a
document that also contains the other PR's changes, and one API-contract version is silently
skipped. This is the exact failure mode the repo's own guidance warns about: *"a matching version
line merging cleanly is the trap: git sees identical text, not a taken version."*

The PR body says `1.4.0` was "read off live `origin/main` immediately before the bump", which is
true and is not sufficient — the question is not what `main` holds, it is whether the next number
is still free across the open set. Whoever merges second should re-read `origin/main` and take
`1.6.0`. Flagging on both PRs.

### Everything else verified and correct

* `SctInstPayment.kt:11` — `SctInstStatus { PENDING, PROCESSING, SETTLED, REJECTED, TIMEOUT,
  RECALLED }`. Byte-equal to the new list. `SUBMITTED`/`ACCEPTED` are absent from the domain.
* `SctInstDtos.kt:27-38` — `SctInstPaymentResponse(paymentId, status, debtorIban, creditorIban,
  amount, currency, endToEndId, executionTimeoutAt, settledAt, createdAt)`. The rewritten schema
  matches property for property, including `required`. The old schema's `id`, `debtorAccountId`,
  `creditorName`, `submittedAt` and `currencyCode` are indeed fields the server never sends.
* `SctInstDtos.kt:11-23` — `SubmitSctInstRequest` has `idempotencyKey: String`,
  `debtorIban: String` and `debtorName: String` all **non-nullable and without defaults**, so a
  body built from the old published document genuinely cannot deserialise. This is the item that
  makes the change breaking under `oasdiff`, and it is breaking against a document that could
  never have produced a working request — the aml precedent applies.
* `currency: String = "EUR"` in the DTO matches the spec's `default: EUR`, and `creditorBic:
  String?` / `remittanceInfo: String?` match the new `nullable: true`. Both correct.
* The `limit`/`offset` → `page`/`size` fix: worth noting this is the **quiet** failure of the three
  — a client passing `limit=50` gets no 400, just silently ignored parameters and the default page.
  That is the shape that survives longest in production.
* `docs/03-api.en.md:96` independently lists all three property defects. Third service in this
  family where someone wrote the drift down instead of fixing it.

**Pacts.** No `@PactFolder` provider class exists in `openbank-sepa-instant`, and the single pact
naming it has sepa-instant as the consumer — so, as with #5990, the pact axis offers no protection
here in either direction. Stated as a fact about the probe, not as reassurance.

**Deliberate non-fixes are the right calls.** `RecallRequest.reason` is `val reason: String` — a
free string with no Kotlin enum to reconcile against, so the gate correctly does not pair it and
whether those four values are an intended constraint is a product question, not a reconciliation.
Leaving it is better than inventing an answer.

---

JiRaska:

Live `origin/main` is **sepa-instant `1.4.0`**. Two open PRs bump it to `1.5.0`:

- this PR
- #5694

(#6019 also carries a sepa-instant hunk, but it is **byte-identical** to #5992's — duplicate work
rather than a third claimant.)

Whichever merges **second** must re-read the version off live `origin/main` and take `1.6.0`:

```bash
git fetch origin main
git show origin/main:openbank-sepa-instant/src/main/resources/openapi.yaml | grep -m1 '^  version:'
```

No release tag is involved — tags sit on the 0.x release axis (`sepa-instant-v0.8.1`), which
ADR-0048 keeps separate from the 1.x API-contract axis.

Nothing to change right now, and this is **not** what is currently blocking this PR: the red
#6220 fixed on main at 14:57Z, after these runs started. It clears on the next run.

JiRaska:
Closing as superseded — measured against live `origin/main` (`f5d063035`).

- The `openbank-sepa-instant` spec this PR rewrites is now **byte-identical** on `main`:
  `diff <(git show origin/main:openbank-sepa-instant/src/main/resources/openapi.yaml) <(git show <head>:...)`
  produces no output, and `main` already carries `info.version: 1.5.0` — the exact version this
  branch bumps to, so the branch cannot merge without re-taking a number that is already spent.
- The gate side has landed too: `.github/scripts/check-openapi-enum-vs-domain.py` on `main` no
  longer contains the `openbank-sepa-instant:ACCEPTED,RECALLED,REJECTED,SETTLED,SUBMITTED`
  baseline entry, which was this PR's only remaining change to that file.

So the whole payload — the `SctInstStatus` vocabulary and the baseline retirement that goes with
it — is already on `main`. Nothing is left to merge.

JiRaska:
### `openapi.yaml` `info.version: 1.5.0` is already taken on live `main`

Read just now off `origin/main`:

```
$ git show origin/main:openbank-sepa-instant/src/main/resources/openapi.yaml | grep -m1 '^  version:'
  version: 1.5.0
```

This branch also declares `1.5.0`. That merges **cleanly** — git sees identical text, not a consumed
version number — so no conflict will be raised and the contract-version axis silently stops advancing.

`1.6.0` is not free either: #5694 claims it. The next free number for this branch is **`1.7.0`**.

PR #6009, opened 2026-08-20
feat(admin-ui): resolve KYC parties and charter-back approval identity

JiRaska:
Two of the nine acceptance criteria on #5904. PR #5988 has the Trace Explorer states and the regulatory export gate; the remaining five (test inventory, Customer 360 context, compliance-pack content, Temporal diagram labelling, and the release-gate criterion) are untouched here.

## 1. Resolve KYC customers by name / company / UUID through party-service

**What party-service actually exposes** — established from `PartyResource` and `PartyRepositoryImpl`, not only from `openapi.yaml`:

| Endpoint | Mode | Notes |
|---|---|---|
| `GET /api/v1/parties/{id}` | exact UUID | not feature-gated |
| `GET /api/v1/parties/search?q=` | name / company / business key | `q` >= 2 chars, cursor-paginated; case-insensitive substring over `legalName`, `tradingName`, `email`, `phone`, `taxId`, `registrationNumber` |

So both lookup modes the UI needs are real server-side modes. No client-side filter over a full list is involved. Birth number is deliberately not searchable (GDPR data-minimisation) and nothing here changes that.

**The trap that decides the failure states.** `/search` carries `@FeatureFlag(flag = "party-search")`. When the flag is off, `FeatureDisabledMapper` answers **404**. A search that legitimately matches nothing answers **200 with an empty page** — `PartyService.searchParties` never 404s for a no-match. So a 404 from `/search` means "this bank cannot search right now", never "this customer does not exist", and it renders as a failed lookup with copy that says the emptiness is not evidence.

The KYC page previously ran `cases.filter(...)` behind a box labelled "Search by ID or status". That box remains, relabelled honestly as "Filter loaded cases" — it can never find a customer by name, because a KYC case payload carries no party name.

**Three states:** `loading` (lookup in flight) · `none` (the lookup ran; nothing matches — worded differently for the UUID and the term case) · `failed` (with a typed reason: `search_unavailable`, `not_deployed`, `scaled_to_zero`, `unauthorized`, `unreachable`, `error`). Plus `too_short`, which refuses to ask party-service for a page it would return empty by construction.

## 2. Charter-backed agent identity in the approval queue

The queue decided AI provenance with a regex over a string in the row:

```
const aiGenerated = /assistant|agent|\bai\b/i.test(p.proposedBy)
```

That is a guess dressed as provenance. `risk-agent-review-desk` (a human desk) earns the badge; a real charter id like `compliance-officer` does not.

Identity now comes from `openbank-libs/governance/agents.yaml` — the enforced registry the OPA agent bundles are derived from, so it is what actually decides what an agent may do. A new BFF route projects it through the existing `loadAgentCharters()` loader (no second source of truth). Matching is a closed list of exact rewrites (bare id, `service-account-<id>`, `agent:<id>`, `openbank-` / `-service` trims) — no substring matching, because a near-miss that resolves is worse than one that shows as unresolved.

**Three states:** `chartered` (shows the charter id, plane and `requires_human`) · `unresolved` (the registry was read and holds no such actor — real evidence of absence) · `unverifiable` (the registry could not be read; `agents.yaml` is bundled by Dockerfile COPY, and an image built without it must not mark every proposer as unchartered). Plus a `loading` badge so an in-flight fetch is not a premature verdict. The ADR-0080 caution now shows whenever AI authorship is established *or* cannot be ruled out; only a positively unchartered actor drops it.

`src/app/api/approvals/pending/route.ts` is **not touched** — that file is heavily contended and the identity resolution belongs on the client, against the charter registry. Its source count is 10 before and 10 after (`lending`, `sanctions`, `transaction`, `domesticPayment`, `clearing`, `fx`, `ledger`, `swift`, `sepaPayment`, `agent`).

## Falsification

Every guarantee was broken and the corresponding test confirmed red (baseline 22/22 green, restored after each):

| Mutation | Result |
|---|---|
| `/search` 404 collapses into `none` | 1 failed |
| `unverifiable` collapses into `unresolved` | 2 failed |
| charter matching becomes fuzzy substring | 1 failed |
| badge renders one state for unresolved + unverifiable | 1 failed |
| `PartyLookup` drops its loading state | 1 failed |

The charter test reads the real `agents.yaml` rather than a fixture, so a renamed agent id breaks it.

## Verification

- `npm test` — 221 files, 1584 tests, all passing
- `npx tsc --noEmit` — clean
- `eslint` on every touched file — 0 errors; the 2 `react-hooks/set-state-in-effect` warnings are pre-existing (confirmed identical on `origin/main`)

Refs #5904

### DIFF
--- a/openbank-admin-ui/src/app/api/governance/agent-identities/route.ts
+++ b/openbank-admin-ui/src/app/api/governance/agent-identities/route.ts
@@ -1,0 +1,35 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) OpenBank contributors. Licensed under the Apache License, Version 2.0.
+// See LICENSE in the repository root or https://www.apache.org/licenses/LICENSE-2.0 for details.
+
+// The enforced agent registry, projected to just what the approval queue needs to attribute
+// a proposal to a charter (issue #5904). Reads openbank-libs/governance/agents.yaml through
+// the shared loader — no second source of truth for "what does agents.yaml actually say".
+//
+// `available: false` is a first-class answer, not an error: agents.yaml is bundled into the
+// image by Dockerfile COPY, and an image built without it must make the queue say "identity
+// unverifiable" rather than silently report every proposer as unchartered.
+
+import { NextResponse } from 'next/server'
+import { loadAgentCharters } from '@/lib/governance/agentCharters'
+import type { AgentIdentityRegistry } from '@/lib/governance/agentIdentity'
+
+export const dynamic = 'force-dynamic'
+
+export async function GET() {
+  try {
+    const { available, agents } = await loadAgentCharters()
+    const body: AgentIdentityRegistry = {
+      available,
+      agents: agents.map(a => ({
+        id: a.id,
+        plane: a.plane,
+        charter: a.charter,
+        requiresHuman: a.requiresHuman,
+      })),
+    }
+    return NextResponse.json(body)
+  } catch {
+    return NextResponse.json({ available: false, agents: [] } satisfies AgentIdentityRegistry)
+  }
+}
--- a/openbank-admin-ui/src/app/approvals/page.tsx
+++ b/openbank-admin-ui/src/app/approvals/page.tsx
@@ -12,9 +12,11 @@
 
 import { useCallback, useEffect, useMemo, useState } from 'react'
 import { useSession } from 'next-auth/react'
-import { CheckCircle2, XCircle, Clock, ClipboardCheck, RefreshCw, ShieldCheck, AlertTriangle, Bot } from 'lucide-react'
+import { CheckCircle2, XCircle, Clock, ClipboardCheck, RefreshCw, ShieldCheck, AlertTriangle } from 'lucide-react'
 import { useLanguage } from '@/lib/i18n/LanguageContext'
 import { PageHeader } from '@/components/ui/PageHeader'
+import { AgentIdentityBadge } from '@/components/approvals/AgentIdentityBadge'
+import { resolveAgentIdentity, type AgentIdentityRegistry } from '@/lib/governance/agentIdentity'
 import { AuthGuard, Can } from '@/components/auth/AuthGuard'
 
 interface Proposal {
@@ -59,6 +61,11 @@ export default function ApprovalsPage() {
   const [busyId, setBusyId] = useState<string | null>(null)
   const [reasons, setReasons] = useState<Record<string, string>>({})
   const [error, setError] = useState<string | null>(null)
+  // The enforced charter registry (agents.yaml). `null` while it is still being fetched or
+  // after the fetch failed — resolveAgentIdentity turns both into `unverifiable`, and
+  // `registryLoading` keeps the two apart in the UI (#5904).
+  const [registry, setRegistry] = useState<AgentIdentityRegistry | null>(null)
+  const [registryLoading, setRegistryLoading] = useState(true)
 
   const load = useCallback(async () => {
     setLoading(true)
@@ -83,6 +90,25 @@ export default function ApprovalsPage() {
   }, [])
 
   useEffect(() => { void load() }, [load])
+
+  // Charter lookup is deliberately a separate read from the queue: a registry that cannot be
+  // read must degrade the IDENTITY column only, never blank the queue itself.
+  useEffect(() => {
+    let cancelled = false
+    void (async () => {
+      try {
+        const res = await fetch('/api/governance/agent-identities', { cache: 'no-store' })
+        if (!res.ok) throw new Error('registry')
+        const body = (await res.json()) as AgentIdentityRegistry
+        if (!cancelled) setRegistry(body)
+      } catch {
+        if (!cancelled) setRegistry(null)
+      } finally {
+        if (!cancelled) setRegistryLoading(false)
+      }
+    })()
+    return () => { cancelled = true }
+  }, [])
 
   const decide = async (p: Proposal, approve: boolean) => {
     setBusyId(p.id)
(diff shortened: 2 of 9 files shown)


JiRaska:

3-dot diff against `origin/main`; branch is 25 commits behind. This is the strongest identity work in the current batch — `resolveParty.ts` and `agentIdentity.ts` both name three states where the codebase had one boolean, and both document the negative case. Findings are about collision and reach, not correctness of the core logic.

### 1. BLOCKING — direct collision with #5905, same two files, same decision, incompatible mechanisms

Confirmed by content, not filenames:

- `src/app/approvals/page.tsx`: both PRs replace the identical `const aiGenerated = /assistant|agent|\bai\b/i.test(p.proposedBy)` hunk. #5905 enriches the proposals BFF route with an `agent` object (`src/app/api/agent/proposals/route.ts`); this PR adds a separate `/api/governance/agent-identities` read plus `resolveAgentIdentity`. Both go through `loadAgentCharters`.
- `src/app/kyc/page.tsx`: both insert a party lookup above the case filter. #5905 reuses the shared `@/components/party/PartySearch` already on `main`; this PR adds a new `@/components/kyc/PartyLookup` + `src/lib/party/resolveParty.ts`.

**Your identity mechanism is the right one and #5905's is not** — its route ignores `loadAgentCharters()`'s `available` flag, so an unreadable `agents.yaml` makes every proposal render as human-authored and drops the ADR-0080 caution banner. That is precisely the `unverifiable`-collapsed-into-`unresolved` failure your header comment names. I have said so on #5905.

**Your party lookup is the duplicated one.** `PartySearch.tsx` on `main` carries an ADR-0208 header saying it is shared deliberately so there are not "two divergent search behaviours over one endpoint", and this PR creates the second one. Concretely divergent already:
- `partyLabel()` (`resolveParty.ts:159`) prefers `tradingName` over `legalName`; `partyDisplayName()` (`PartySearch.tsx:41`) prefers `legalName`. The same party gets two different names on two screens.
- `PARTY_UUID_RE` is declared twice (`PartySearch.tsx:30`, `resolveParty.ts:37` as `UUID_RE`).
- `MIN_TERM_LENGTH = 2` exists only here; `PartySearch` has no minimum.

The subtraction that keeps the value: lift `resolveParty`'s three-state resolution (particularly `search_unavailable`) **into** `PartySearch`, and let both KYC and the existing callers get the better failure states, rather than shipping a second component. `PartySearch.tsx:90-92` currently maps every non-401/403 `!res.ok` — the flag-off 404 included — to `unreachable`, which is not the conflation you are guarding against but is a strictly worse message than yours.

### 2. `resolveParty` handles more response envelopes than the endpoint has

`matchesFrom` (`resolveParty.ts:104-112`) accepts a bare array, `{data:[]}` and `{items:[]}`, and `toMatch` accepts either `id` or `partyId` and either `legalName` or `name`. Defensive, but three of those five shapes are ones party-service does not emit, and the cost is that a genuinely wrong-shaped response (a BFF error body, an HTML error page that parsed) resolves to `{status:'none'}` — "the bank holds no such party" — instead of `failed`. That is the same conflation one layer down. Consider pinning the one real envelope and treating anything else as `failed`/`error`.

### 3. `candidateIds` normalisation is unreachable for the `service-account-` case

`agentIdentity.ts:56` strips `service-account-`. Worth confirming against a real proposal row that `proposedBy` can actually carry that form — if agent proposals only ever carry a bare charter id, three of the four rewrites in `candidateIds` are speculative surface, and each one is a chance for a near-miss to resolve. The comment says "a near-miss that resolves is worse than one that shows as unresolved"; that argues for the list being shorter than four, not longer.

### 4. `cautionAi = identity.status !== 'unresolved'` — correct, and please assert the boot case

`page.tsx:243` makes the caution banner show for `chartered` **and** `unverifiable`, which is the fail-safe direction. But `{cautionAi && !registryLoading && ...}` means during the registry fetch there is a window where the queue renders with no caution at all. On a cold pod with a slow route that window is the whole first paint. Is there a test asserting what the card looks like at `registryLoading === true`? `approvals-agent-identity.test.tsx` covers the resolved states; the t=0 state is a fourth state, and this repo has been bitten by not re-deriving a value at t=0 on a cold pod.

### Checked and clean
- `src/app/api/approvals/pending/route.ts` untouched — the 10 sources on `main` are intact.
- No `NEXT_PUBLIC_*` introduced.
- `/api/governance/agent-identities` projects only `id/plane/charter/requiresHuman` — no charter internals or tool grants leak to the browser. Its `catch` returns `available:false` rather than a 500, which is the correct degradation given the consumer.
- `resolveParty` never throws: transport error and non-JSON both become `failed`, so a rejected promise cannot surface as "nothing found".
- Please confirm the suites were run via `npm test` and not bare `npx vitest run` — the `pretest` hook bakes `governance.json`/`catalog.json`, and without them several suites fail in a way that reads like a main-branch regression.

JiRaska:
## Cross-reference: #5905 replaces the same `aiGenerated` line

Flagging a collision found while reviewing #5905. Both PRs replace
`openbank-admin-ui/src/app/approvals/page.tsx:208`:

- **this PR**: `resolveAgentIdentity(p.proposedBy, registry)` -> `chartered | unresolved |
  unverifiable`, with `cautionAi = identity.status !== 'unresolved'` so the ADR-0080 banner stands
  whenever AI authorship is established *or cannot be ruled out*.
- **#5905**: `const aiGenerated = p.agent ? p.agent.icon === 'bot' : /…/.test(p.proposedBy)` — two
  states, fed by `/api/agent/proposals`, which ignores `loadAgentCharters()`'s `available` flag and
  so marks **every** proposal human when `agents.yaml` cannot be read.

They also both touch `kyc/page.tsx`, where #5905 is currently conflicting with `main`.

Nothing to change here — this PR's design is the more complete of the two and reads its own
`/api/governance/agent-identities`, so it does not depend on #5905's route. I have recommended on
#5905 that it drop its `approvals/page.tsx` hunk and keep only its route fix (which makes that
route emit the same `status` vocabulary this PR uses), so the two do not put competing provenance
models on one screen. Raising it here so whichever lands second expects the conflict.

Issue #2319, opened 2026-07-25
Two swift consumer pacts are outside the contract-drift gate (and the #2404 cited for it is not a real issue)


JiRaska:
## What is ungated

`openbank-swift-service/build.gradle.kts` statically excludes both of its consumer pact tests whenever `CI=true`:

```kotlin
exclude("**/SwiftEventPactConsumerTest*")
exclude("**/ClearingSimulatorPactConsumerTest*")
```

so no `--tests` filter can bring them back inside a CI job. Two committed pacts are therefore outside the ADR-0063 Phase 2 contract-drift gate (`.github/workflows/pact-drift-check.yml`) — nothing regenerates them, so nothing can show they still match the consumer tests that produced them:

- `pacts/openbank-transaction-service-openbank-swift-service.json` (from `SwiftEventPactConsumerTest`)
- `pacts/openbank-swift-service-openbank-clearing-simulator.json` (from `ClearingSimulatorPactConsumerTest`)

The workflow used to name only the first of the two.

## Why they were excluded

`PactConsumerTestExt` auto-publishes the generated pact to `pactbroker.url` in its `AfterTestTemplate` callback. CI forwards `pactbroker.url` to the test JVM, the broker 401s or hangs, and the JVM stalls for 30+ min. Excluding the classes was the stop-gap. This is a *publishing* problem, not a contract problem — pact generation itself is a pure JVM mock-server run with no broker involved.

## Wrong issue reference to correct

The exclusion comments cite `#2404`, which is **not a real issue or PR number in this repo** (highest issue 2280, highest PR 2306 at the time of writing). The number also reached `openbank-swift-service/CHANGELOG.md` (generated — leave it) and ADR-0063 (corrected in #2318). Three comment sites in `openbank-swift-service/build.gradle.kts` still point at it and should point here.

## What concretely must happen

1. Stop `PactConsumerTestExt` from publishing during pact *generation* — the drift gate only needs the JSON on disk. Either do not forward `pactbroker.url` to the consumer-test JVM, or set `pact.verifier.publishResults=false` / the equivalent consumer-side switch for these classes, so the `AfterTestTemplate` publish never runs.
2. Drop the two `exclude(...)` lines, keeping the `SwiftBootSmokeIT` exclusion (a separate, unrelated Quarkus/Testcontainers boot hang).
[...]

PR #6016, opened 2026-08-20
fix(swift): publish COMPLETED, the terminal status the pact already asserts

JiRaska:
Refs #5962, ADR-0048.

#5962 — all merged in, so their diffs drop out as they land. **This PR's own work is one enum
line, two description lines and one version line in
`openbank-swift-service/src/main/resources/openapi.yaml`, plus one `BASELINE` entry.**

## What was wrong

|---|---|---|
| `SwiftStatus` `[PENDING, VALIDATED, SENT, ACKNOWLEDGED, REJECTED, FAILED]` | `SwiftStatus { PENDING, VALIDATED, SENT, ACKNOWLEDGED, REJECTED, FAILED, COMPLETED }` | `COMPLETED` — the terminal success state — is emitted by the server and declared by nothing. A strict generated deserializer fails on exactly the outcome every successful payment reaches |

The published `description:` block enumerated all six declared values with a gloss each, so the
omission was not a truncated list someone would notice: the document read as complete.

## Which side is right — the domain, and here a green provider replay proves it

This is the strongest witness class available, and unlike the sibling PRs it did not need the
database (swift's `status` column is free `VARCHAR(20)` with no CHECK):

**A committed pact asserts the omitted value, and the provider replays it on every PR.**
`pacts/openbank-transaction-service-openbank-swift-service.json` carries a message interaction
described literally as *"a swift.message.status-changed event with status COMPLETED"*, with
contents `"status": "COMPLETED"` and a matching rule pinning the field to

```
"regex": "VALIDATED|SENT|COMPLETED|REJECTED"
```

so the consumer's contract names `COMPLETED` as a value it must handle. That pact is replayed
against the real provider by `SwiftPactFolderProviderVerificationTest` — `@PactFolder`, ungated,
runs on every PR (it exists precisely because the broker-sourced class never did; see #2319).

**I measured the replay rather than reading the annotation:** in this branch's run that class
reported `tests="1" skipped="0" failures="0" errors="0"`. A green replay of a pact that asserts
`COMPLETED` is direct evidence the server emits it.

This is also the same value the #2319 drift was about — the issue calls it out by name. The pact
side was fixed then; the published spec was not.

## Classification

Purely additive — one value added to a response enum, no removal:

```
api-contract gate: openbank-swift-service: additive change, info.version 1.3.0 -> 1.4.0 — OK (required >= MINOR).
```

No `correction` reclassification needed. `info.version` read off freshly fetched `origin/main`
immediately before the bump (`"1.3.0"`, quoted in this spec — the quoting is preserved).

## Verification

- `ktlintCheck`, `detekt`, `test` all **executed** (none `UP-TO-DATE`): 62 tests, 0 failures,
  **1 skipped** — `SwiftMessagePactProviderVerificationTest`, the expected `@PactBroker`-gated
  class; its `@PactFolder` sibling ran green as quoted above.

### DIFF
--- a/.github/scripts/check-openapi-enum-vs-domain.py
+++ b/.github/scripts/check-openapi-enum-vs-domain.py
@@ -111,8 +111,6 @@ BASELINE: dict[str, str] = {
         "#5962 — SepaPaymentStatus: spec-only PENDING/RECALLED; undeclared CANCELLED/RECEIVED/RETURNED/VALIDATED",
     "openbank-statement-service:RECONCILIATION,UNKNOWN,UPSTREAM":
         "#5962 — CloseFailureReason: undeclared NOT_VIABLE",
-    "openbank-swift-service:ACKNOWLEDGED,FAILED,PENDING,REJECTED,SENT,VALIDATED":
-        "#5962 — SwiftStatus: undeclared COMPLETED",
 }
 
 SPEC_ENUM_INLINE = re.compile(r"enum:\s*\[([^\]]*)\]")
--- a/openbank-swift-service/src/main/resources/openapi.yaml
+++ b/openbank-swift-service/src/main/resources/openapi.yaml
@@ -9,7 +9,7 @@ info:
   # major == openbank.api.version == URL /api/v1 (ADR-0048). Minor bump (1.2 -> 1.3): additive
   # only — GET /swift/approvals (the checker's queue, issue #5679) plus two additive fields on
   # ApprovalResponse (makerId, createdAt). No breaking change.
-  version: "1.3.0"
+  version: "1.4.0"
   description: |
     SWIFT MT/MX message dispatch, acknowledgement, and status tracking.
     Supports MT103 (customer credit transfer), MT202 (bank transfer), MT900/910/940/950 (confirmations/statements).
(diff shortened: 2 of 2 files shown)


JiRaska:
## Review — #6016 (swift)

Reviewed with `git diff origin/main...origin/fix/swift-openapi-enum-drift-5962`. **No change
requested; one nit.**

**The pact witness is real and I verified the literal.**
`pacts/openbank-transaction-service-openbank-swift-service.json:13` asserts `"status": "COMPLETED"`.
The published `SwiftStatus` enum did not contain that value. A committed pact asserting a status the
document denies exists is the cleanest possible proof of which side drifted — and per #2319 this is
the *same value* the earlier swift pact drift was about, so this closes a loop rather than opening
one.

`SwiftMessage.kt:14` — `enum class SwiftStatus { PENDING, VALIDATED, SENT, ACKNOWLEDGED, REJECTED,
FAILED, COMPLETED }`. The new spec list matches. Purely additive.

**Pact replay coverage confirmed:** `SwiftPactFolderProviderVerificationTest` exists and is
`@PactFolder`, so it replays on a PR rather than only on main-push — which matters here, because
the pact is the entire basis of the claim.

**Version.** `origin/main` reads `"1.3.0"` (quoted, unusually for this fleet); the bump to
`"1.4.0"` preserves the quoting, so nothing changes shape for a YAML consumer. Correct and free;
no competing open PR on this file.

**Good detail:** the description block gains a `COMPLETED:` line alongside the other six, so the
prose and the enum stay in lockstep. Several specs in this family have an enum whose description
list is already one short — worth doing everywhere.

**Nit.** `openapi.yaml:5-9` still reads `Minor bump (1.2 -> 1.3): additive only — GET
/swift/approvals …`. Accurate on `main`, made stale by this bump.

---

JiRaska:

PR #6017, opened 2026-08-20
fix(consent): reconcile the consent request/response schemas with the DTOs they describe

JiRaska:
Refs #5962, ADR-0048.

#5962 — all merged in, so their diffs drop out as they land. **This PR's own work is
`openbank-consent-service/src/main/resources/openapi.yaml` and one `BASELINE` entry.**

## The baselined enum was the smallest of three problems

The baseline entry says *"ConsentStatus: spec-only PENDING; undeclared PENDING_SCA/SUPERSEDED"*,
and that is true. But `ConsentStatus` lives inside `ConsentResponse`, and following the method
finding from the previous tranche — compare each schema's **properties** against the DTO the
resource actually populates, not just its enums — the schema around it turned out to be describing
a different object. Both consent schemas were reconciled here; fixing the enum alone would have
polished a document that still could not be used.

### 1. `CreateConsentRequest` — a spec-generated body cannot deserialise at all

| spec declared | Kotlin `CreateConsentRequest` |
|---|---|
| `required: [partyId, granteeId, scopes, expiresAt]` | `partyId, granteeId, granteeType, granteeName, scopes, accountIbans?, validTo, redirectUri?, tppTransactionId?` |

- **`granteeType`, `granteeName` and `validTo` are non-nullable with no default and were entirely
  undeclared.** `ConsentResource.create` reads `request.granteeType` and `request.granteeName`
  directly into `CreateConsentCommand`, so a body built from the published document fails to
  construct the DTO — this endpoint has never been callable from a generated client.
- **`expiresAt` does not exist**; the field is `validTo`.
- **`accountIds` (array of `uuid`) does not exist**; the field is `accountIbans`, a list of IBAN
  strings — a different name *and* a different type.
- `scopes` was `array of plain string` with the example `[accounts:read, transactions:read]`,
  a lowercase-colon spelling `ConsentScope` has never contained.
- `redirectUri` and `tppTransactionId` were unpublished, and `tppTransactionId` is the
  **idempotency key** (`ConsentResource.create` prefers it over `X-Request-ID`), so the one field
  a TPP needs to make a retry safe was invisible.

### 2. `ConsentResponse` — five properties omitted, one invented

Against the DTO (`id, partyId, granteeId, granteeType, granteeName, scopes, accountIbans, status,
validFrom, validTo, createdAt`) the schema omitted `granteeType`, `granteeName`, `accountIbans`,
`validFrom` and `validTo`, and invented `expiresAt`.

### 3. The enum itself

`[PENDING, ACTIVE, REJECTED, REVOKED, EXPIRED]` against
`ConsentStatus { PENDING_SCA, ACTIVE, EXPIRED, REVOKED, REJECTED, SUPERSEDED }`.

## Which side is right — the domain, on two witnesses

1. **The database says `PENDING_SCA` is the state every consent starts in.**
   `V1__init_consent.sql:10`:
   ```sql
   status VARCHAR(50) NOT NULL DEFAULT 'PENDING_SCA',
   ```
   The column default *is* the creation state, and the resource's own `@Operation` summary says so
   too: *"Create a new consent (PENDING_SCA)"*. So the server emits `PENDING_SCA` on the 201 of
   every create, and the document declared only `PENDING`.
2. **`PENDING` never existed here.** `git log -S` scoped to `openbank-consent-service` finds a bare
   `PENDING` in the Kotlin only as `OutboxStatus.PENDING` — the unrelated transactional-outbox enum
   — never as a `ConsentStatus`. This is an invented value, not a rename. (Scoping matters: `PENDING`
   is genuine in half the fleet, so an unscoped grep reads as corroboration and is not.)

`SUPERSEDED` is unambiguously real — it is written when a newer consent replaces one for the same
grantee and scopes.

## One more drift, found by hand rather than by the gate

The `scope` query parameter of `GET /parties/{partyId}/consents/{granteeId}/check` published
**7 of the 20 real `ConsentScope` values**. It never paired with the domain enum and so was never
reported: 7/20 = 0.35, under the gate's `MIN_OVERLAP = 0.4`. Per #5962's guidance the threshold is
left alone — this was found by comparing properties by hand. The handler takes the parameter as
`String?` and `valueOf`s it, so all 20 are genuinely queryable; 13 were simply unpublished,
including every `STATEMENTS_READ` / `AGENT_*` / `TELEMETRY_RUM` scope.

`ConsentScope` and `GranteeType` are now published as named schemas and `$ref`ed from all four
sites, so the vocabulary is stated once. The gate's `SUBJECTS` count rises 148 -> 150 as the two
new schemas begin pairing — and both pair *exactly*, which is the check on this change.

## Classification

Spec-only PR, so the removals reclassify — confirmed with the real gate, not assumed:

```
::notice::api-contract gate: openbank-consent-service: spec-only PR — reclassifying 4 breaking
document change(s) (first: added the new required request property `granteeName`) as a CORRECTION.
api-contract gate: openbank-consent-service: correction change, info.version 1.5.0 -> 1.6.0 — OK (required >= MINOR).
```

MINOR, so no URL major moves (ADR-0048 D2). `info.version` read off freshly fetched `origin/main`
immediately before the bump (`1.5.0`).

## Verification

- YAML parses; every `$ref` added resolves (the api-contract gate loads the spec strictly and
  would have reported a dangling `$ref` as an unclassifiable spec).
- `ktlintCheck`, `detekt`, `test` all **executed** (none `UP-TO-DATE`): 113 tests, 0 failures,
  **1 skipped** — `ConsentPactBrokerProviderVerificationTest`, the expected broker-gated class.

## What this does not do

No Kotlin changed. The server behaves exactly as before; what changes is that the published
document now describes it. Keeping the PR spec-only is also what makes the `correction`
reclassification available.

## Security checklist


JiRaska:
## Review — #6017 (consent)

Reviewed with `git diff origin/main...origin/fix/consent-openapi-enum-drift-5962`. **No change
requested; two notes.** This is the largest diff in the family (+67/-14) and the scope is
justified — but it is scope, so it deserves the closest read.

**Every schema in the rewrite matches its DTO. Verified property by property against
`openbank-consent-service/src/main/kotlin/com/openbank/consent/infrastructure/rest/dto/ConsentDtos.kt`
on `origin/main`:**

* `CreateConsentRequest` (`ConsentDtos.kt:11-21`) — `partyId, granteeId, granteeType, granteeName,
  scopes, accountIbans, validTo, redirectUri, tppTransactionId`. The new schema matches all nine,
  and the new `required` list is exactly the non-nullable subset (`accountIbans`, `redirectUri`,
  `tppTransactionId` are `?`). The old schema's `expiresAt` and `accountIds` do not exist in the
  DTO at all, and `granteeType`/`granteeName` — both **non-nullable, no default** — were
  unpublished, so a body built from the old document could not deserialise. That is the same class
  as sepa-instant's `SubmitSctInstRequest` in #5992, and it is what makes this a defect rather than
  a tidy-up.
* `ConsentResponse` (`ConsentDtos.kt:27-38`) — `id, partyId, granteeId, granteeType, granteeName,
  scopes, accountIbans, status, validFrom, validTo, createdAt`. Matches, `required` included. The
  old schema published `expiresAt`, which the server has never sent.
* `ConsentScope` — I extracted all 20 constants from `Consent.kt:12-56` and they are byte-for-byte
  the new enum list, ordering included (AISP 7, PISP 4, CBPII 1, AGENT_* 4, TELEMETRY_RUM, the
  MARKETING_COMMS_* triple). Note the old `checkConsent` query parameter published **7 of 20** —
  the majority of the vocabulary was unqueryable.
* `GranteeType` (`Consent.kt:57`) — `TPP, BANK_AGENT, CUSTOMER_AGENT, INTERNAL_SERVICE`. Matches.
* `ConsentStatus` (`Consent.kt:64`) — `PENDING_SCA, ACTIVE, EXPIRED, REVOKED, REJECTED, SUPERSEDED`.
  Matches. The old `PENDING` never existed; `PENDING_SCA` is the state **every** consent is created
  in, so the old document could not describe a single freshly created consent.

**Note 1 — extracting the three enums into named `components/schemas` is the right structural call,
and worth keeping.** `checkConsent`'s `scope` parameter now `$ref`s `ConsentScope` instead of
carrying an inline copy. An inline duplicate of a 20-value enum in a second place is precisely how
the next drift starts, and this removes the opportunity rather than just correcting today's copy.

**Note 2 — pact impact, checked.** Two committed pacts have consent as the **provider**
(`openbank-campaign-service-…` and `openbank-mcp-service-…`), and consent's replay is
`ConsentPactProviderVerificationTest` (`@PactFolder`, runs on a PR). Neither pact asserts
`expiresAt` or `accountIds` — I grepped the whole `pacts/` tree for both literals and got zero
hits — so no interaction pins a field name this PR moves. Since the change is spec-only, the replay
verdict cannot change either way; the grep matters only for establishing that no consumer was
written against the old spelling.

**Version.** `origin/main` reads `1.5.0`; `1.6.0` correct and free. No competing open PR on this
file.

**Nit.** `openapi.yaml:6-7` still reads `Minor bump (1.0 -> 1.1): additive endpoint only (PATCH
/approvals/{id} …)`. Already stale on `main` (the file was at 1.5.0), so not introduced here — but
this is the PR that is reconciling the document with reality, so it is the natural place to fix it.

---

JiRaska:
Rebuilt on current `origin/main` and **narrowed** — the branch's `ConsentScope` half is now
superseded, and merging it as it stood would have removed values from the published spec.

Measured against `origin/main`:

- `main`'s inline `ConsentScope` enum is a strict **superset** of this branch's. Diffing the two
  sets: `main`-only = `['CREDIT_OFFERS', 'CREDIT_PROFILE_USE']` (ADR-0269), branch-only = `[]`.
  This branch extracted the enum into a `ConsentScope` component and `$ref`'d it — a good refactor
  when it was written, but its literal list predates those two values, so the extraction would have
  published a **smaller** consent vocabulary than the service accepts. That is the exact defect
  #5962 exists to prevent, reintroduced from the other direction.
- `main` already carries the full inline list plus a comment explaining the gate, so the
  `ConsentScope` work has effectively landed. Dropped.

What was genuinely still missing, and is what this PR now contains:

- **`ConsentStatus`.** The spec advertised `[PENDING, ACTIVE, REJECTED, REVOKED, EXPIRED]`; the
  domain enum (`Consent.kt`) is `PENDING_SCA, ACTIVE, EXPIRED, REVOKED, REJECTED, SUPERSEDED`. So
  the spec named a `PENDING` the code has never produced and omitted `PENDING_SCA` — the state
  **every** consent is created in — and `SUPERSEDED`. Still baselined on `main`
  (`#5962 — ConsentStatus: spec-only PENDING; undeclared PENDING_SCA/SUPERSEDED`); that entry is
  now retired.
- `info.version` `1.6.0 -> 1.7.0`. Note `1.6.0` was already taken on `main` while this branch was
  open, so the branch's own bump had gone stale.

Verification, both directions rather than just the green one:

```
python3 .github/scripts/check-openapi-enum-vs-domain.py --enforce
  -> 16 paired enum(s) drift, all 16 baselined; no new drift.   rc=0
# negative control: put the retired baseline entry back
  -> rc=1  (stale baseline entry reported, as it should be)
# restored
  -> rc=0
```

One thing for review rather than for CI: removing `PENDING` from a published response enum is
technically a narrowing. It is safe here because the domain has never emitted it — no consumer can
have observed it — but if `oasdiff` classifies it as breaking, the honest answer is a major bump,
not a re-added fiction.

JiRaska:
**#6017 and #6235 both claim `openbank-consent-service` `openapi.yaml` `info.version: 1.7.0`.**
`main` is at `1.6.0`, so whichever merges second ships a second, different contract under a version
already published — and there is nothing to notice it.

```
main   1.6.0
#6017  1.7.0
#6235  1.7.0
```

Git will not flag this: both sides write the identical `version:` line, so the merge is textually
clean. The contract gate classifies the bump from the diff against the merge-base *as of the run*,
so it cannot see a competing claim that lands after the run. Whoever goes second needs **1.8.0**,
re-read off live `main` at merge time rather than taken from this comment:

```bash
git fetch origin main && git show origin/main:openbank-consent-service/src/main/resources/openapi.yaml | grep -m1 '^  version:'
```

Note #6235 also carries `openbank-lending-service` and `openbank-customer-edge` spec bumps, so it is
the more entangled of the two — its lending version jumps 1.14.0 → 1.16.0, which is consistent with
it being slice 2 of a stack, but worth confirming that 1.15.0 belongs to slice 1 and not to a gap.

Found by sweeping the openapi-touching PRs against `main`. Three other PRs (#6019, #5905, #6481)
turned out to have bumped to a number that a *merged* PR had already taken — the same trap from the
other direction, and the one that gets missed, because looking for a race scans open PRs.

JiRaska:

| | `info.version` |
|---|---|
| merge-base | `1.6.0` |
| this branch | `1.7.0` |
| `origin/main` **now** | `1.7.0` |

```
git fetch origin main
git show origin/main:openbank-consent-service/src/main/resources/openapi.yaml | grep -m1 '^  version:'
```

Something else shipped `1.7.0` for this service while this PR sat. The branch and `main` would both claim one API-contract version. Git will not flag it — the version line merges cleanly because the text is *identical*; a taken version is not a textual conflict. The green checks here ran nine days ago against the old base and cannot see it.

JiRaska:
**Measured: `info.version` is unchanged, and the `openapi-version-not-taken` gate now exists.**

```
openbank-consent-service/src/main/resources/openapi.yaml
  origin/main : version: 1.7.0
  this branch : version: 1.7.0
```
The spec file is edited but the version is not bumped, so the branch republishes a version `main` has
already shipped. Git cannot see this (identical text is not a conflict), and this PR's last CI run is from
**2026-08-30** — the `openapi-version-not-taken` gate landed on main **2026-08-31 06:44** (`a373867e9`,
#7633, `.github/gates/gates.yaml:2633`). So the current green is about a base that had no such gate; the
next run should go red.

Minimal fix: `info.version: 1.8.0`, re-read against live `main` first.

Order note: **#7037** also edits this same file and already claims `1.8.0`. Whichever of the two lands
second needs another bump.

JiRaska:
Rebased onto current `main`. One conflict, and it was the version line — which is the failure mode
this PR was always exposed to while it sat 1398 commits behind:

```
this branch proposed   info.version: 1.7.0
main is already at     info.version: 1.8.0
```

Both 1.7.0 and 1.8.0 were taken by other consent-service spec changes in the meantime. Resolved to
main's line and moved to the next free minor, **1.9.0**. Re-read it after any competing consent spec
merge — whoever lands second takes the next number, and a matching version line merging cleanly is
exactly the trap here: git compares text, not which numbers are already published.

The enum change itself is untouched and still the point of the PR:

```yaml
- enum: [PENDING, ACTIVE, REJECTED, REVOKED, EXPIRED]
+ enum: [PENDING_SCA, ACTIVE, EXPIRED, REVOKED, REJECTED, SUPERSEDED]
```

`PENDING` is a value `ConsentStatus` has never had, while `PENDING_SCA` — where every consent is
created — and `SUPERSEDED` were both missing.

`check-openapi-enum-vs-domain.py` passes on the merged result (`SUBJECTS=162`, exit 0), so the spec
and the Kotlin enum now agree in both directions.

JiRaska:
Re-merged: the branch went `DIRTY` again within the hour, and the cause is worth recording because it
is not the one this PR is usually at risk from.

Not the version — `1.9.0` is still free, `main`'s consent spec reads `1.8.0`.

`130998006d` (card-issuance) changed `check-openapi-enum-vs-domain.py` on `main` after my previous
merge, so this branch was carrying the *older* main copy of that file as an apparent change of its
own. Taken from current `main` again; it is byte-identical, and the delta is back to one file:

```
openbank-consent-service/src/main/resources/openapi.yaml
```

That is the general hazard of resolving a conflict by "take main": the resolution is only as current
as the moment you took it, and a shared file that several PRs touch will move again. Nothing to do
about it except re-check before merge — which is also why the delta is worth reading rather than the
conflict count.

JiRaska:

```
```

```
```

X an ancestor".

JiRaska:
`gates (gitops-api)` failed on **my own change**, and it is the right failure:

```
##[error]openapi-enum drift baseline entry
         `openbank-consent-service:ACTIVE,EXPIRED,PENDING,REJECTED,REVOKED` no longer drifts
```

That baseline entry says the consent spec advertises a spec-only `PENDING` and omits `PENDING_SCA`
and `SUPERSEDED`. This PR corrects exactly that — so the declaration outlived its subject the moment
the spec was fixed, and the gate reports a baseline in either direction by design.

Removed it. Falsified rather than trusting the green: with the entry put back, the gate reports
`no longer drifts` again; without it, exit 0 and `SUBJECTS=161`.

Worth naming the pattern, because this is the third instance in two days: **a PR that closes a known
gap must delete that gap's baseline entry in the same PR.** #7981 wrote three VEX overlays and left
their `NO_OVERLAY_YET` entries, which turned `gates (supplychain)` red for every open PR until #8012
removed them. #5986 removes two account-service entries for the same reason. This is the same shape,
and I walked into it while fixing one of the others.

JiRaska:
Re-merged current `main` to retrigger. The last run's `all-green` failure was not a finding — two jobs
were **cancelled**:

```
fail    all-green
cancel  nightly fleet-health alert
cancel  Detect changed services
```

`all-green` treats a cancelled service build as a failure, which is correct (a cancelled job proves
nothing), but it means the verdict was about a spot-kill rather than about this diff. Nothing else was
red.

JiRaska:

```
```

```markdown
## Security checklist

      (license, CVE, source).
```

PR #6019, opened 2026-08-20
fix(card-issuance): reconcile IssueCardRequest with the DTO it describes

JiRaska:
Refs #5962, ADR-0048.

#5962 — all merged in, so their diffs drop out as they land. **This PR's own work is one schema
block and one version line in `openbank-card-issuance-service/src/main/resources/openapi.yaml`,
plus one `BASELINE` entry.**

Not a money-path service (`rules.yaml: money_path_services`) — 1 approval.

## The baselined enum was a symptom; the schema around it was the defect

The baseline entry says *"CardNetwork: undeclared AMEX/UNIONPAY"*. Comparing `IssueCardRequest`'s
**properties** against the Kotlin DTO — the method finding from the previous tranche — shows the
enum is the least of it: **a body generated from the published document cannot deserialise**, so
`POST /api/v1/cards` has never been callable from a generated client.

| spec declared | Kotlin `IssueCardRequest` |
|---|---|
| `required: [partyId, accountId, cardType, cardNetwork]`; properties `partyId, accountId, cardType, cardNetwork, embossedName, deliveryAddress, delegationGrantId` | `partyId, accountId, productCode, cardType, network, cardholderName, embossedName, currency, dailyLimitMinorUnits = 500_000, monthlyLimitMinorUnits = 5_000_000, deliveryAddress: String? = null, delegationGrantId: UUID? = null` |

Four independent falsehoods in one schema:

1. **The network property has the wrong name.** The spec calls it `cardNetwork`; the DTO field is
   `network`, non-nullable with no default. Jackson binds nothing, and construction fails — so the
   enum drift is secondary to the property never arriving at all.
2. **Three non-nullable, defaultless fields were undeclared**: `productCode`, `cardholderName`,
   `currency`. Each alone is enough to fail deserialisation.
3. **`deliveryAddress` had the wrong type** — declared `type: object`, the field is `String?`.
4. **The enum**: `[VISA, MASTERCARD]` against `CardNetwork { VISA, MASTERCARD, AMEX, UNIONPAY }`,
   so no generated client could request an Amex or UnionPay card.

`dailyLimitMinorUnits` / `monthlyLimitMinorUnits` were also unpublished; they are now declared with
their Kotlin defaults so a caller can see both that they are optional and what they default to.

## Which side is right — the document contradicts itself, which settles it

No external witness was needed: **this spec already publishes the full vocabulary twice**, and
only the request site disagrees.

- `CardResponse.network` — `enum: [VISA, MASTERCARD, AMEX, UNIONPAY]`
- `CardEntitlementsResponse.networks[]` — `enum: [VISA, MASTERCARD, AMEX, UNIONPAY]`
- `IssueCardRequest.cardNetwork` — `enum: [VISA, MASTERCARD]`

So as published, an Amex card can be **returned** and listed as **entitled**, but not
**requested**. A document that names the same vocabulary in three places and abbreviates it in one
is a hand-edit that was never re-checked, not a deliberate restriction — and the DTOs confirm it:
all three sites are the same `CardNetwork` type (`CardDtos.kt` lines 18, 43, 95, 111).

The property renames are equally settled by the DTO, which is the thing the endpoint actually
deserialises into.

## Classification

Spec-only PR, so the removals reclassify — confirmed with the real gate:

```
::notice::api-contract gate: openbank-card-issuance-service: spec-only PR — reclassifying 6 breaking
document change(s) (first: added the new required request property `cardholderName`) as a CORRECTION.
api-contract gate: openbank-card-issuance-service: correction change, info.version 1.7.0 -> 1.8.0 — OK (required >= MINOR).
```

MINOR, so no URL major moves (ADR-0048 D2). `info.version` read off freshly fetched `origin/main`
immediately before the bump (`1.7.0`).

## Verification

  (`SUBJECTS` falls 148 -> 147: the `{VISA, MASTERCARD}` value-set stops existing as a distinct
  pairing once it becomes the four-value set the other two sites already used.)
- YAML parses.
- `ktlintCheck` and `detekt` both **executed** and passed. `test`: 189 tests, 0 failures,
  **1 skipped** — `CardIssuancePactBrokerProviderVerificationTest`, the expected broker-gated
  class; the `@PactFolder` sibling ran green (`tests="1" failures="0"`).
  The first `test` run failed one case with `QuarkusBindException` — the documented
  port-contention flake, not a real failure; a serialized re-run was `BUILD SUCCESSFUL`.

## What this does not do

No Kotlin changed. The server behaves exactly as before; the published document now describes it.

### DIFF
--- a/.github/scripts/check-openapi-enum-vs-domain.py
+++ b/.github/scripts/check-openapi-enum-vs-domain.py
@@ -70,8 +70,6 @@ BASELINE: dict[str, str] = {
         "#5962 — Channel: undeclared EMAIL",
     "openbank-campaign-service:DRY_RUN,SENT,SUPPRESSED_CAP,SUPPRESSED_CONSENT,SUPPRESSED_QUIET_HOURS":
         "#5962 — SendOutcome: undeclared CONVERTED/FAILED/SKIPPED_CONDITION/SUPPRESSED_LIST",
-    "openbank-card-issuance-service:MASTERCARD,VISA":
-        "#5962 — CardNetwork: undeclared AMEX/UNIONPAY",
     "openbank-consent-service:ACTIVE,EXPIRED,PENDING,REJECTED,REVOKED":
         "#5962 — ConsentStatus: spec-only PENDING; undeclared PENDING_SCA/SUPERSEDED",
     "openbank-copilot-service:CARD_FREEZE,DISPUTE,PAYMENT":
--- a/openbank-card-issuance-service/src/main/resources/openapi.yaml
+++ b/openbank-card-issuance-service/src/main/resources/openapi.yaml
@@ -3,7 +3,7 @@
 openapi: 3.1.0
 info:
   title: Card Issuance Service API
-  version: 1.7.0
+  version: 1.8.0
   description: Debit/credit card lifecycle — issue, activate, block, suspend, resume, cancel;
     synthetic PAN vault, product-catalog entitlements, and additional-cardholder cards issued
     under a delegation grant (ADR-0249 D1/D2).
@@ -586,7 +586,8 @@ components:
 
     IssueCardRequest:
       type: object
-      required: [partyId, accountId, cardType, cardNetwork]
+      required: [partyId, accountId, productCode, cardType, network, cardholderName,
+                 embossedName, currency]
       properties:
         partyId:
           type: string
@@ -594,16 +595,32 @@ components:
         accountId:
           type: string
           format: uuid
+        productCode:
+          type: string
         cardType:
           type: string
           enum: [DEBIT, CREDIT, PREPAID, VIRTUAL, SINGLE_USE]
-        cardNetwork:
+        network:
           type: string
-          enum: [VISA, MASTERCARD]
+          enum: [VISA, MASTERCARD, AMEX, UNIONPAY]
+        cardholderName:
+          type: string
         embossedName:
           type: string
+        currency:
+          type: string
+          description: ISO 4217 alphabetic code.
+        dailyLimitMinorUnits:
+          type: integer
+          format: int64
+          default: 500000
+        monthlyLimitMinorUnits:
+          type: integer
+          format: int64
+          default: 5000000
         deliveryAddress:
-          type: object
+          type: string
+          nullable: true
         delegationGrantId:
           type: string
           format: uuid


JiRaska:
## Review — #6019 (card-issuance)

Reviewed with `git diff origin/main...origin/fix/card-issuance-openapi-enum-drift-5962`.
**Needs a change: `info.version 1.8.0` is already claimed by another open PR.** Content is correct.

### ⚠️ Version collision with #4308

**#4308** (`fix(card-issuance): alert on dead-lettered outbox rows and add an operator requeue
path`, open, last updated 2026-08-20) edits the same
`openbank-card-issuance-service/src/main/resources/openapi.yaml` and makes the identical bump:

```
-  version: 1.7.0
+  version: 1.8.0
```

Same single line, same from-value, same to-value — so git merges both **without a conflict** and
the second to land silently reuses a consumed version. Reading `origin/main` before bumping cannot
see this; `main` still says `1.7.0` for both. Whoever merges second should re-read `origin/main`
and take `1.9.0`. Flagged on #4308 as well.

### Content verified — and the diff is bigger than "an enum"

`CardDtos.kt:13-26` on `origin/main` is
`IssueCardRequest(partyId, accountId, productCode, cardType, network, cardholderName, embossedName,
currency, dailyLimitMinorUnits = 500_000L, monthlyLimitMinorUnits = 5_000_000L,
deliveryAddress: String? = null, delegationGrantId: UUID? = null)`.

Every element of the rewrite checks out against that:

* **`cardNetwork` → `network` is a property rename, not an enum fix**, and it is the most
  consequential line in the PR: the DTO field is `network`, non-nullable with no default, so a body
  built from the published document omitted a required property and **could not deserialise**. The
  enum widening (`AMEX`, `UNIONPAY`) — which is what #5962 actually listed — would have been
  cosmetic on a schema that could not produce a working request. Fixing both together is correct;
  it is worth the title saying so, since "reconcile IssueCardRequest with the DTO" undersells it.
* `productCode`, `cardholderName`, `currency` — all three non-nullable with no default, all three
  previously unpublished. Now in `required`. Correct.
* `CardNetwork` (`Card.kt:53`) — `{ VISA, MASTERCARD, AMEX, UNIONPAY }`. Matches.
* `CardType` (`Card.kt:52`) — `{ DEBIT, CREDIT, PREPAID, VIRTUAL, SINGLE_USE }`. Already correct in
  the spec; untouched. Good.
* `deliveryAddress` `type: object` → `type: string, nullable: true` — the DTO is `String? = null`.
  Correct, and this one was actively misleading before.
* The two `*MinorUnits` defaults (`500000` / `5000000`) match the Kotlin defaults exactly.

**No consumer sends the old spelling.** Fleet grep for `cardNetwork` across
`openbank-admin-ui/src` returns nothing, and the only `"network"` literal in `pacts/` is
`openbank-copilot-service-openbank-customer-edge.json:93` (`"network": "VISA"`) — the *new*
spelling, against a different provider. So the rename breaks no committed expectation.

**Pact replay.** `CardIssuancePactFolderProviderVerificationTest` exists (`@PactFolder`, runs on a
PR), alongside the broker-gated sibling. Coverage is in place.

**Version bump is otherwise well-classified** — `new-required-request-property` makes this breaking
to `oasdiff`, and the clients that "break" are ones generated from a document that could never have
produced a deserialisable body. Same as kyc (#5966) and sepa-instant (#5992).

---

JiRaska:

Live `origin/main` is **card-issuance `1.7.0`**. Two open PRs bump it to `1.8.0`:

- #6019
- #4308

Unlike the sepa-instant pair, these two are a **genuine race** — their card-issuance spec changes
differ, so this is not duplicate work and both sets of changes are wanted.

Whichever merges **second** must re-read the version off live `origin/main` and take `1.9.0`:

```bash
git fetch origin main
git show origin/main:openbank-card-issuance-service/src/main/resources/openapi.yaml | grep -m1 '^  version:'
```

No release tag is involved — tags sit on the 0.x release axis
(`card-issuance-service-v0.24.1`), which ADR-0048 keeps separate from the 1.x API-contract axis,
and no merged PR has consumed `1.8.0`.

Nothing to change right now. Note this is **not** what is currently blocking either PR: the red
#6220 fixed on main at 14:57Z, after those runs started — it clears on the next run.

JiRaska:
### `openapi.yaml` `info.version: 1.8.0` is claimed by two open PRs at once

`origin/main` is currently at `1.7.0` for `openbank-card-issuance-service`, so `1.8.0` is free *today* —
but **both this PR and #6019 declare `1.8.0`**. Whichever lands second will merge cleanly (identical
text is not a conflict) and quietly ship a duplicate contract version.

Whoever merges second needs to re-read the version off live `main` and bump to `1.9.0` before merging.
Flagging on both PRs so neither side assumes the other checked.

JiRaska:
### Heads-up: `openapi.yaml` `info.version` **1.8.0 is double-claimed**

Read off live `origin/main` just now:

| Source | `openbank-card-issuance-service/.../openapi.yaml` `info.version` |
|---|---|
| `origin/main` | **1.7.0** |
| this PR (#6019) | **1.8.0** |
| #4308 (`fix/card-issuance-dead-outbox`, OPEN) | **1.8.0** |

Both PRs bump 1.7.0 → 1.8.0 on the same file. There is no textual conflict — each side is a
clean one-line change against the same base, so git sees identical text rather than a taken
version. **Whichever merges first takes 1.8.0; the second must be re-bumped to 1.9.0 first.**

This PR is currently `CLEAN`, so it is the likelier of the two to land first — in which case
nothing changes here and #4308 is the one that needs the re-bump (noted there too). Flagging
so the second merge is not a surprise.

Check to run immediately before merging:

```
git fetch origin main
git show origin/main:openbank-card-issuance-service/src/main/resources/openapi.yaml | grep -m1 '^  version:'
```

JiRaska:
**`openapi.yaml` `info.version` 1.8.0 is already taken on `main`.** This PR bumps
1.7.0 → 1.8.0, but #4308 landed that same number while this branch was open, so

```
$ git show origin/main:openbank-card-issuance-service/src/main/resources/openapi.yaml | grep -m1 '^  version:'
  version: 1.8.0
```

The next free number is **1.9.0**.

last run predates #4308. That is the one shape no base resolution can close, which is

Worth re-checking rather than just applying 1.9.0 — read the version off live `main` at

```bash
git fetch origin main && git show origin/main:openbank-card-issuance-service/src/main/resources/openapi.yaml | grep -m1 '^  version:'
```

consent-service `1.7.0`.

JiRaska:

| | `info.version` |
|---|---|
| merge-base | `1.7.0` |
| this branch | `1.8.0` |
| `origin/main` **now** | `1.8.0` |

```
git fetch origin main
git show origin/main:openbank-card-issuance-service/src/main/resources/openapi.yaml | grep -m1 '^  version:'
```

Something else shipped `1.8.0` for this service while this PR sat. The branch and `main` would both claim one API-contract version. Git will not flag it — the version line merges cleanly because the text is *identical*; a taken version is not a textual conflict. The green checks here ran nine days ago against the old base and cannot see it.

JiRaska:
**Measured: `info.version` is unchanged, and the `openapi-version-not-taken` gate now exists.**

```
openbank-card-issuance-service/src/main/resources/openapi.yaml
  origin/main : version: 1.8.0
  this branch : version: 1.8.0
```
The spec is edited without a bump, republishing a version `main` already shipped. Invisible to git
(identical text is not a conflict). This PR's last CI run is **2026-08-30**; the
`openapi-version-not-taken` gate landed on main **2026-08-31 06:44** (`a373867e9`, #7633,
`.github/gates/gates.yaml:2633`), so the present green predates the gate and the next run should go red.

Minimal fix: `info.version: 1.9.0`, re-read against live `main` first.

PR #6020, opened 2026-08-20
docs(ci): correct two enum-drift baseline reasons that asked for a regression

### LINKED ISSUE #5962: openapi enum vs domain: 17 baseline entries across 13 services remain (was 27/16); 5 of them are annotated NOT drift

JiRaska:
Refs #5962, ADR-0048.

drops out once #5966 lands. **This PR's own work is two `BASELINE` reasons in
`.github/scripts/check-openapi-enum-vs-domain.py`. No spec, no Kotlin, no behaviour.**

## Why this is a PR and not two enum fixes

#5962 lists 27 drifts and treats each as a defect to reconcile. Two of them are not drift at all,
and **acting on their stated reasons would have shipped a regression.** Both are a *deliberate
subset*: a request enum that intentionally publishes fewer values than the lifecycle enum it is
drawn from, because the domain rejects the rest by construction. The gate pairs by value overlap
and cannot see the restriction — a limitation worth recording in place, next to the entries it
produces, rather than rediscovering per service.

They stay baselined. Only the reasons change, so the ratchet still fails if either genuinely
changes shape.

### `openbank-ledger-service:CUTOFF,LOCKED,TIED_OUT`

Reason said *"AccountingDayStatus: undeclared OPEN"*.

The values are the `{to}` path parameter of
`POST /api/v1/ledger/accounting-days/{businessDate}/transitions/{to}`. `OPEN` is not a reachable
transition **target**:

- a day is created in `OPEN` by a **different** endpoint (`POST /accounting-days/{businessDate}`,
  whose description reads *"Creates the day in OPEN"*);
- `AccountingDay.canTransitionTo` is `next.ordinal == ordinal + 1`, and `OPEN` is ordinal 0, so no
  current status can ever name it as a successor;
- ADR-0207 D2/D3 makes that deliberate — there is no reopen, and a day needing correction after
  cutoff is corrected *forward* via a reversal into the current open day. The enum's own KDoc says
  so.

So adding `OPEN` would advertise a transition that answers **409 every time**. The document is
already right, and the response schemas that carry a whole-lifecycle status
(`AccountingDayResponse.status`, `CurrentAccountingDayResponse.status`) do publish all four values
— which is the tell that the three-value list is a considered restriction, not an abbreviation.

**Money-path service, so this one is worth being explicit about:** the reason as written invited a
money-path spec change that would have made the contract less accurate.

### `openbank-document-service:DECLINED,SIGNED`

Reason said *"SignerStatus: undeclared PENDING"*.

The values are `RecordDecisionRequest.decision`, and a signer decides `SIGNED` or `DECLINED`;
`PENDING` is the state a signer *starts* in, never one they can submit.
`SignatureCeremony.recordDecision` enforces exactly that:

```kotlin
require(decision == SignerStatus.SIGNED || decision == SignerStatus.DECLINED)
```

so publishing `PENDING` would advertise a value the domain rejects at the aggregate boundary. The
endpoint summary already says *"Record a signer decision (SIGNED / DECLINED)"*.

## Effect on the issue's arithmetic

#5962's tail is 13 once the eight service PRs land. Two of those 13 are these, and they are not
work — they are the gate's known blind spot for a legitimate subset. The actionable tail is
**11**, alongside the two pid entries #5977 already re-reasoned as schema defects rather than enum
defects. Recording that here means the next tranche does not spend the analysis again, and does not
"fix" them.

## Verification

- `check-openapi-enum-vs-domain.py --enforce` — green; 19 drifts, all 19 still baselined, no
  stale entry in either direction (the entries are unchanged as keys — only their reasons move).
- `check-openapi-enum-vs-domain.py --self-test` — passed, unchanged.


JiRaska:
## Review — #6020 (baseline reasons)

Reviewed with `git diff origin/main...origin/docs/enum-drift-baseline-reasons-5962`. **No change
requested.** I independently verified both claims and both are correct — and both would indeed have
produced a regression if acted on as originally written.

### `openbank-ledger-service:CUTOFF,LOCKED,TIED_OUT` — confirmed, and this is money-path

`openbank-ledger-service/src/main/kotlin/com/openbank/ledger/domain/model/AccountingDay.kt`:

* `:28-32` — `enum class AccountingDayStatus { OPEN, CUTOFF, TIED_OUT, LOCKED }`, so `OPEN` is
  ordinal **0**.
* `:39` — `fun canTransitionTo(next: AccountingDayStatus): Boolean = next.ordinal == ordinal + 1`,
  under the KDoc *"Monotonic single-step forward progression; anything else is rejected."*

`next.ordinal == ordinal + 1` can never be satisfied for `next == OPEN`, because no status has
ordinal `-1`. `OPEN` is structurally unreachable as a transition target, not merely unusual — and
`:77` routes the rejection through `checkConflict`, i.e. **409**, exactly as the PR states. Adding
`OPEN` to the `{to}` path enum would have published a transition that answers 409 on every call,
on the ledger, in a document whose whole purpose is to tell a client what it may send.

The corroborating detail the PR gives — that `AccountingDayResponse.status` and
`CurrentAccountingDayResponse.status` *do* publish all four values while the transition path
publishes three — is the right tell, and it is the one a value-overlap matcher cannot see. Same
document, two lists, deliberately different lengths.

### `openbank-document-service:DECLINED,SIGNED` — confirmed

`openbank-document-service/src/main/kotlin/com/openbank/document/domain/model/SignatureCeremony.kt:60`:

```kotlin
require(decision == SignerStatus.SIGNED || decision == SignerStatus.DECLINED)
```

`PENDING` is rejected at the aggregate boundary. Publishing it would advertise a value the domain
refuses. Correct to leave baselined.

### On the shape of this PR

Filing "these two are not drift" as its own reviewable change, rather than quietly skipping them or
deleting the entries, is the right handling. The entries stay in `BASELINE`, so the ratchet still
fires if either genuinely changes shape — the gate's coverage is unchanged and only the *reason*
moves. That distinction matters: a reason is what the next person acts on, and both of these
reasons were instructions to make a contract less accurate.

The generalisable point, worth keeping somewhere more durable than a baseline comment: **a
value-overlap matcher cannot distinguish a stale transcription from a deliberate subset**, and the
two need opposite responses. The tell in both cases here is a second list in the same document
that publishes the full vocabulary. That is a property a future version of the gate could check
mechanically rather than leaving to per-service judgement.

**Scope.** No spec, no Kotlin, no behaviour — two string literals. `docs(ci)` is the correct type
and correctly does not release anything.

---

JiRaska:
## Review — the reasoning is right, but the title and the version bump both understate what this PR does

**Staleness:** 5 commits behind `main`, `MERGEABLE`, no competing edit to either file. Live.

**The baseline-reason corrections are correct and I verified the underlying claims.** Both entries were reasons that, taken literally, asked for a regression:
- `check-openapi-enum-vs-domain.py`, `openbank-document-service:DECLINED,SIGNED` — publishing `PENDING` would advertise a value `recordDecision`'s `require(...)` rejects by construction.
- `openbank-ledger-service:CUTOFF,LOCKED,TIED_OUT` — `canTransitionTo` is `next.ordinal == ordinal + 1`, which `OPEN` (ordinal 0) can never satisfy, so publishing it would advertise a transition that always answers 409.

Both are "a request/path enum is a deliberate subset of a lifecycle enum, and the gate pairs them without being able to see the restriction". Recording that as *not drift* rather than deleting the entries is the right call — the pair stays pinned.

**The problem: the title says "correct two enum-drift baseline reasons", and the diff also rewrites a shipped API contract.**

`openbank-sepa-instant/src/main/resources/openapi.yaml` is not a comment change. It renames the list query parameters (`limit`/`offset` → `page`/`size`), renames a response property (`id` → `paymentId`), renames `currencyCode` → `currency` in both schemas, adds four fields to `SubmitSctInstRequest.required` (`idempotencyKey`, `debtorIban`, `debtorName`), drops `creditorName` and `debtorAccountId` from the response, and replaces the status enum `[SUBMITTED, ACCEPTED, SETTLED, REJECTED, RECALLED]` with `[PENDING, PROCESSING, SETTLED, REJECTED, TIMEOUT, RECALLED]`.

**I confirmed the spec was the fiction and the code is the truth** — `SctInstResource.kt:94-95` really is `@QueryParam("page")` / `@QueryParam("size")` — so no runtime behaviour changes and this is a genuine correction. Three consequences follow anyway:

1. **The version bump is `1.4.0 → 1.5.0`, a MINOR, for a diff `oasdiff` will classify as breaking.** Renaming a response property, renaming a query parameter and adding required request fields are each breaking on their own. ADR-0048 says the API-contract version is classified from the OpenAPI diff, not from whether the implementation moved. Please re-run the classification; if it comes back breaking, this is a major, and a major changes the URL `/api/v{N}`, which is an author decision this PR should make explicitly rather than by omission.
2. **Re-read `info.version` off live `main` immediately before the final push.** `git show origin/main:openbank-sepa-instant/src/main/resources/openapi.yaml | grep -m1 '^  version:'` — a matching version line merges cleanly because git sees identical text, not a taken number, and the number can be consumed by something that already landed as easily as by a competing open PR.
3. **The commit type `docs` is a hidden type**, so nothing releases. For a spec-only correction that is arguably fine, but say so — otherwise the corrected contract sits on `main` with no changelog entry recording that the published one was wrong.

**Also worth noting in the body:** this PR removes a **third** baseline entry (`openbank-sepa-instant:ACCEPTED,RECALLED,...`), which is the *result* of the spec fix rather than a reason correction. The title says "two".

Verdict: **needs change** — split the sepa-instant spec correction into its own PR with the right type and the oasdiff-derived bump, or retitle this one and justify the version. The enum-baseline half is ready as-is. Not approving — review comment only.

PR #6029, opened 2026-08-20
fix(aml): stamp the real decision time on every AML case transition, and decide openbank.aml.auto-clear (#5837)

### LINKED ISSUE #5837: openbank.aml.auto-clear bypasses a four-eyes compliance control and no ADR decides it — its KYC twin has ADR-0116 §4
`openbank.aml.auto-clear` (env `OPENBANK_AML_AUTO_CLEAR`) auto-clears the AML key of the onboarding
activation gate, so a party activates without a human AML decision. It is enabled in gitops
(`openbank-infra/gitops/components/aml/aml-service.yaml`), defaults to `false` in code, and is
described in `PartyEventConsumer`'s KDoc as sandbox-only.

**No ADR decides it.** Verified: no file under `docs/adr/` mentions AML auto-clear.

## Why the gap is uneven, which is what makes it worth filing

It has a twin. `openbank.kyc.auto-approve` does the same thing for the KYC key of the same gate, is
enabled the same way, and **is** decided — ADR-0116 §4, explicitly. So the two halves of one
activation gate have different governance: one is a recorded decision, one is a config flag whose
only written record is aml-service's own `src/main/resources/docs/`.

ADR-0266 (opened in PR #5784) records the activation gate itself, but decides the gate, not this
flag. ADR-0116 §4 covers only the KYC twin.

## Why it matters

The flag bypasses a four-eyes compliance control. That is defensible as a sandbox affordance — the
KYC twin is the precedent — but the thing that makes it defensible for KYC is that someone wrote
down the decision, its scope, and the production guarantee. Here the guarantee lives in a KDoc
comment and a default value.

Note also that until #5836 lands, `aml-service.yaml`'s comment cites **ADR-0073** for it — which is
hardware-backed credential storage for the mobile app, i.e. an unrelated ADR. So the one written
pointer it does have points somewhere wrong.

## What would close this

Either add a `§` to ADR-0116 covering the AML twin alongside the KYC one (they are one gate, and a
single record for both is probably the honest shape), or a short ADR of its own. It should state:
what the flag does, that it is sandbox-only, what enforces that in production, and what the
compliance argument is for allowing it anywhere.

Found while classifying the ADR-0073 miscitations in #5785 — the site could not be repointed
because there was nothing correct to point it at.

Refs #5785, #5784, ADR-0116, ADR-0266.

JiRaska:
## What I found before changing anything

#5837 asks for a decision record. Establishing the state of the control first turned up a live
defect in the same code path, so this PR carries both.

### (a) The flag's default and deployed value

`openbank.aml.auto-clear` (`OPENBANK_AML_AUTO_CLEAR`) defaults to **`false`** in
`PartyEventConsumer`'s `@ConfigProperty`, and is set **`true`** in the one deployed environment
described in this repo — `openbank-infra/gitops/components/aml/aml-service.yaml`. There is no
second environment manifest, so a CI gate asserting "no non-sandbox manifest sets it true" would
pass over an empty set today. ADR-0268 §2 says that in writing rather than shipping a check that
cannot fail.

### (b) When it fires, what is recorded

The auto-clear is **not** silent, and the issue title's "nothing records it" is too strong. It runs
through the same `AmlCaseService.updateDecision` as an analyst decision, so it writes the
`aml_cases` row (`decided_by='SANDBOX_SYSTEM'`, `assigned_analyst='SANDBOX_BOT'`,
`decision_reason='Sandbox auto-clear (no adverse match)'`), emits `aml.case.status_changed.v1`
through the transactional outbox carrying the same fields, and logs a line. The sentinel
`decidedBy` is what makes an auto-clear separable from a human decision after the fact.

**But the record has no usable timestamp, and that is a real defect.**
`AmlCase.transitionTo` carried `now: Instant = Instant.EPOCH`, and the only caller taking the
default was production — `AmlCaseService.updateDecision` never passed a value, despite the service
already holding an injected `Clock`. So **every terminal AML decision** — an analyst's four-eyes
decision and the sandbox auto-clear alike — recorded `decidedAt` and `updatedAt` as
**1970-01-01T00:00:00Z**.

Nothing masked it downstream, which I checked rather than assumed:
`AmlCaseRepositoryImpl.update` assigns `entity.decidedAt = amlCase.decidedAt` and
`entity.updatedAt = amlCase.updatedAt` directly, and `AmlCaseEntity` has no `@UpdateTimestamp`. So
EPOCH reaches the DB columns. `statusChangedEvent` reads `current.updatedAt` for `occurredAt`, so
1970 also went onto the wire for every consumer. On every decided case `updated_at` sorts before
`created_at`.

Why no test saw it: both existing unit-test callers of `transitionTo` already pass an explicit
`now`, so the default was reachable only from production. The suite even has a fixed `Clock` — it
just never asserted the decision timestamps. This is the repo's own documented
`: Instant = Instant.EPOCH` shape, and the matching rule ("assert recency, never `isNotNull()`") is
why the new test compares against the fixed clock.

### (c) Does the four-eyes control it bypasses actually exist?

**Largely not.** `openbank-aml-service`'s own compliance doc claimed the decision endpoint gives
"four-eyes accountability via `decidedBy`/`assignedAnalyst`". Measured against the code:

- `decidedBy` arrives in the **request body** (`UpdateAmlDecisionRequest.decidedBy` → `toCommand`),
  never from the authenticated security context. Only `require(decidedBy.isNotBlank())` validates
  it. One caller with `ROLE_OPERATOR` can clear a case and self-declare any attribution. ADR-0116 §3
  mandates the exact opposite for the KYC twin.
- `openbank-aml-service` is **not** in `rules.yaml: money_path_services`, so `rest.rego`'s
  `four_eyes_required` can never derive an `aml` action scope — no verb addition could reach
  `amlCase.updateDecision`.
- The state machine permits `OPEN → CLEARED` directly, so there is not even the state-based
  opener/reviewer split ADR-0116 §3 relies on for KYC.
- `AUTHZ_ENFORCE` is `false` for this service in gitops (deliberately, #1797), so
  `@Authorize(action="amlCase.updateDecision")` is advisory — it logs "would DENY" and blocks
  nothing.

So the honest framing is not "a sandbox flag bypassing a strong control", it is "a sandbox flag
skipping a single-operator endpoint with self-declared attribution". That is recorded, not fixed
here — closing it is ADR-0268 §5 and larger than this PR.

## What changed

1. **`fix(aml)`** — `updateDecision` passes `Instant.now(clock)`; `transitionTo`'s `now` parameter
   loses its `Instant.EPOCH` default so the omission cannot recur. Both existing callers already
   passed it, so nothing else moves.
2. **ADR-0268** — the decision record #5837 asks for. Chose a new ADR over appending a `§` to
   ADR-0116 (the issue offers both): ADR-0116 is titled and scoped to the KYC engine, and an AML
   section would make its title false and bury the AML decision where nobody auditing AML looks.
   Cross-referenced both ways. §4 records the finding in (c); §5 lists the four production
   prerequisites.
3. **Docs** — the four-eyes overclaim corrected in `06-compliance.en.md` and `.cs.md`.

## Verification

**Negative control (the point of the exercise).** The test was written first and run against
unmodified production code:

```
expected: 2024-01-15T12:00:00Z
 but was: 1970-01-01T00:00:00Z
AmlCaseServiceTest.kt:144
```
`tests="4" skipped="0" failures="1"` — the defect is real and the other three tests were unaffected.

**Positive control.** Full module suite, `--rerun-tasks`, totals read from the JUnit XML rather than
the console: **29 tests, 1 skipped, 0 failures, 0 errors**. The one skip is
`AmlPactBrokerProviderVerificationTest`, gated on `pactbroker.url` — the same skip count as the RED
run, i.e. nothing silently stopped running.

- `:openbank-aml-service:ktlintCheck :openbank-aml-service:detekt` → exit 0 (run separately, after
  the tests, so a test failure could not hide a lint failure).
- `bash docs/adr/gen-index.sh` → committed → `check-adr-registry.sh` → **OK, 261 ADRs** (regen,
  then commit, then check — the check restores derived files to HEAD on failure).
- `run-gates.py --all` → 151/158 pass. Of the 7 non-passes, **none is this branch's**:
  - four (`threat-model`, `db-migration`, `api-contract`, `schema-compat`) printed
    *"PR_DIFF_BASE is empty but this gate requires it — refusing to run vacuously"*. Re-run with
    `PR_DIFF_BASE=$(git merge-base origin/main HEAD)`: **threat-model PASS, db-migration PASS,
    schema-compat PASS**; api-contract then failed on `sudo: a password is required` while
    installing `oasdiff` locally, and `release-scope-mismatch` on `PR_TITLE: unbound variable` —
    both environmental, both real in CI.
  - `litellm-config-revision` and `loki-rule-load-test` — `git diff --name-only origin/main` shows
    this branch touches no litellm or loki file, so those are main's state, not this change's.

**Not verified, stated plainly:**
- Nothing was read from or run against the live cluster, DB, Prometheus or Kafka. The claim that
  live sandbox `aml_cases` rows carry `decided_at = 1970-01-01` is inferred from the code path, not
  measured. It is checkable read-only with
  `select count(*) from aml_cases where decided_at = '1970-01-01T00:00:00Z'` and worth running.
- No integration test drives this through real HTTP + a real DB, so "EPOCH reaches the column" rests
  on reading `AmlCaseRepositoryImpl` and `AmlCaseEntity`, not on an observed row.
- Whether existing 1970 rows should be backfilled is a data question this PR does not answer. The
  original decision time is not recoverable from the row; the outbox event carries the same EPOCH.
- ADR-0268 §5's prerequisites are unbuilt. This PR decides the flag; it does not gate it.

## Scope notes

- `aml-service.yaml`'s ADR-0073 miscitation is **deliberately untouched** — #5836 is open against
  that exact line. Once both land the comment can point at ADR-0268.
- `openbank-aml-service` is not in `money_path_services`, so no threat model is gated (the
  threat-model gate passes above). No API change, so no `openapi.yaml` or `info.version` bump. No
  migration, no event-schema change. `version.txt` untouched — release-please owns it.

Closes #5837


JiRaska:
## Review — correct fix, structurally enforced; two questions

The fix is the right one and it is the *structural* version rather than the cosmetic one. Removing the default outright:

```kotlin
-        now: Instant = Instant.EPOCH,
+        now: Instant,
```

makes the compiler enforce it at every call site, so this defect cannot recur by omission. Patching only `AmlCaseService.updateDecision` to pass a value while leaving the default in place would have left the trap armed for the next caller. Good.

The diagnosis is also exactly right about why nothing caught it: both unit-test callers already passed a real value, so production was the *only* caller taking the default. That is the `Instant.EPOCH` trap in its purest form — and the new test's comment says the important half out loud, that `isNotNull()` passes happily against `1970-01-01`.

Checked and clean:

- `AmlCaseServiceTest` asserts against the fixed clock (`isEqualTo(expected)`) and explicitly `isNotEqualTo(Instant.EPOCH)`, and — the part that matters most — re-reads `occurredAt` out of the **serialised event payload** through the same mapper that wrote it, so the assertion covers `aml.case.status_changed.v1` on the wire and not just the in-memory object. It would fail against `main`. Real negative control.
- `clock` is already injected in `AmlCaseService` (`:30`), so no new wiring and no `Instant.now()` sneaking in ungoverned.
- No repository write path changed, so no Panache `persist()`-on-assigned-`@Id` exposure.
- ADR registry derived files (`README.md`, `DIGEST.md`, `index.json`) are regenerated alongside ADR-0268, not hand-edited.
- CI: 27 success, 0 failures.

### Question 1 — what happens to the decisions already recorded as 1970?

Every terminal AML decision taken before this lands has `decidedAt`/`updatedAt` = `1970-01-01` in the DB **and** on `aml.case.status_changed.v1`. These are compliance records; "when was this AML decision taken" is the kind of field an auditor asks about directly, and the wrong value is not self-evidently wrong to a downstream consumer — it is a valid timestamp.

I could not find any mention of the existing rows in the diff or in ADR-0268. Please say what the plan is: a backfill (from the outbox/audit trail, if a real time is recoverable), or an explicit, recorded decision that the historical rows stay as they are and are known-unreliable before this deploy. Either is fine; silence is not, because after this merges the 1970 rows become indistinguishable from "not yet decided" with nothing left to explain them.

Worth checking specifically whether `decidedAt` feeds anything hash-chained or otherwise append-only downstream — if it does, the backfill option closes and the answer has to be documentation.

### Question 2 — two `Instant.now(clock)` calls in one operation

`AmlCaseService.kt:36` already computes `val now = Instant.now(clock)`, and `:101` calls `Instant.now(clock)` again for the transition. Under the fixed test clock these are equal, so the new test cannot see the difference; under a real clock they are two different instants within a single logical operation, so `decidedAt` can differ from whatever `:36`'s `now` stamps.

Small, but it is the same family of defect as the one being fixed — a timestamp that does not mean quite what the reader assumes. Reusing `now` at `:101` costs nothing and makes the whole operation share one instant.

PR #6032, opened 2026-08-20
fix(observability): the Loki ruler never loaded a single rule — `loki.ruler` is not a chart key

### LINKED ISSUE #5734: Falco detects ~33k events/day and nothing alerts on any of them — the runtime-security control is write-only
## What

Falco is deployed, healthy (14/14 DaemonSet pods) and **actively detecting** — and **nothing alerts on any of it**.

Measured 2026-08-19 against the sandbox, last 24h via Loki:

```
29465  Contact K8S API Server From Container
 3150  Run shell untrusted
  450  Drop and execute new binary in container
    5  Falco internal: syscall event drop
```

~33k detections/day. Alert rules referencing Falco:

- Prometheus `PrometheusRule` CRs: **0**
- Grafana-managed (unified alerting) rules: **0** — the whole cluster has zero Grafana-managed alert rules, so there is no Loki-based alerting of any kind

## Why this contradicts its own design note

`openbank-infra/gitops/apps/falco.yaml` states:

> Integration is deliberately zero-new-plumbing: json output to stdout → the Alloy DaemonSet already tails every pod and ships to Loki (structured metadata) → Grafana. **Alerting on CRITICAL/EMERGENCY rules rides the existing Loki datasource**; no falcosidekick, no new webhook surface.

The delivery half is real and works — the logs reach Loki, and `openbank-soc`'s panels read them (fixed to LogQL in #5052). The **alerting** half was never built. Falco is the control this repo added specifically because "nothing watched a container AFTER it was admitted"; it is currently write-only.

## Two problems, and they interact

1. **No alert exists**, so a genuine detection is indistinguishable from the 33k/day baseline — nobody is paged for `Drop and execute new binary in container` (450/day), a rule that fires on real malware behaviour.
2. **The ruleset was never tuned**, which is why the baseline is 33k/day. `falco.yaml`'s own comment says "tuning exceptions land here as values once the first week of sandbox noise is triaged" — that triage never happened. Adding an alert on top of an untuned ruleset would page continuously and be muted within a day, which is worse than no alert.

So the order matters: tune first, then alert. An alert added before the tuning is an alert that gets ignored, and this repo already has that lesson written down for `WorkflowLivenessStale`.

## Suggested shape

1. Triage the three noisy rules (`Contact K8S API Server From Container` is almost certainly every in-cluster service account doing normal work) into `customRules` exceptions in `falco.yaml`.
2. Then add a Grafana-managed Loki alert on `priority=~"Critical|Emergency"` (or on the surviving rule set after tuning), routed through the existing Alertmanager receivers.
3. Verify by effect: trigger a known-positive (e.g. `kubectl exec` a shell in a test pod) and confirm the alert actually reaches a human, per this repo's "a guard is proven by what it prevents" rule.

Found during a full observability audit, 2026-08-19.

JiRaska:
## The finding

#5850 (merged this morning) tuned the Falco ruleset and added three Loki alert rules for #5734. Those rules are correct. **They have never evaluated, and neither has any other log-based alert in this estate.**

The `ruler:` block in `gitops/apps/loki.yaml` has been keyed `loki.ruler:` since #3060 wired it up. The grafana/loki chart's key is **`loki.rulerConfig:`**. Helm discards an unknown value key silently — no error, no warning, a perfectly healthy pod — so every line of that block was inert and the ruler ran on chart defaults.

Measured live against the sandbox, 2026-08-20:

```
GET /config              ruler.storage.type      = "s3"          (declared: local)
                         ruler.storage.local     = {directory:""} (declared: /rules)
                         ruler.rule_path         = "/var/loki/rules-temp" (declared: /tmp/loki-rules)
                         ruler.alertmanager_url  = ""            (declared: kube-prometheus-stack-alertmanager:9093)
                         ruler.ring.kvstore.store= "memberlist"  (declared: inmemory)
                         ruler.remote_write.enabled = false      (declared: true)

exec loki-sc-rules -- find /rules -type f
                         /rules/fake/silent-failures.yaml
                         /rules/fake/endpoint-availability.yaml
                         /rules/fake/falco-runtime.yaml

GET /prometheus/api/v1/rules   ->  {"data":{"groups":[]}}        <-- ZERO groups loaded
```

Every key in the declared block is at its Loki default. The two that *appear* to have taken effect (`enable_api`, `enable_alertmanager_v2`) are Loki's own defaults, so they are not evidence either. The sidecar dutifully writes three rule files to a directory the ruler is not configured to read, and the ruler has nowhere to deliver anything it might fire.

Confirmed independently by rendering the repo's own `valuesObject` through chart 6.55.0 twice — as committed, and with the key renamed:

```
as-is   ruler:  {storage: {type: s3, s3: {...}}, wal: {...}}          <-- matches live /config exactly
fixed   ruler:  {alertmanager_url: ..., enable_api: true, remote_write: {enabled: true, ...},
                 ring: {kvstore: {store: inmemory}}, rule_path: /tmp/loki-rules,
                 storage: {type: local, local: {directory: /rules}}, wal: {...}}
```

Same family as the 21 Argo Rollouts that had zero series from mis-nested Helm keys.

## So the honest answer to #5734

Falco's events **do** reach a system an alert could read — Loki, via Alloy, and the `openbank-soc` panels prove it. #5850 put correct rules in front of them. What was missing is one layer lower: the ruler that would evaluate those rules was never configured, and the Alertmanager it would deliver to was never set. The runtime-security control was write-only for the reason nobody had looked at, and it stayed write-only after the PR that set out to fix it.

## By-rule breakdown, and the split that matters

The issue's ~33k/day is not the alertable volume. Split by rule **and by priority** — the priority column decides which of these an alert can ever see:

| rule | stock priority | pre-tune (per 6h) | post-tune (per 6h) |
|---|---|---:|---:|
| Contact K8S API Server From Container | NOTICE | 5263 | 1156 |
| Run shell untrusted | NOTICE | 685 | 138 |
| Drop and execute new binary in container | **CRITICAL** | 106 | **0** |
| Falco internal: syscall event drop | CRITICAL (self-report) | ~1 | ~1 |
| (lines with no rule field — agent startup/internal) | — | 178 | 51 |

Query (Loki instant, 6h windows either side of the #5850 sync at ~09:50Z):

```
sum by (rule) (count_over_time({namespace="falco"} | json | __error__="" [6h]))
```

So: **the Critical-or-above volume an alert actually sees went from ~424/day to 0/day.** #5850's tuning is correct and its arc-reaper attribution holds — the exception is scoped to one namespace AND one pod-name prefix AND three process names, so a drop-and-execute anywhere else, or a different binary in that same pod, still fires. The 78% cut on the NOTICE rules buys log volume and dashboard signal only; no alert here can see them.

## Why each alert is non-vacuous at t=0 on a cold pod

Re-derived after the tuning, because the tuning changed the assumption two of them were written under — not carried over from #5850's reasoning.

- **`FalcoCriticalRuntimeDetection`** — `count_over_time(...)>0`, no `for`. The series is real and reachable: it carried 106 events per 6h before the tuning. It reads Loki's stored data, so a cold ruler or Falco pod does not synthesise a boot-time value. Currently 0 firing, which is the correct resting state and the whole point.
- **`FalcoSyscallEventDrops`** — 3/day measured, `for: 5m`, `warning`. Rate-limited by Falco itself, so that is a real count of windows and not a suppressed flood.
- **`FalcoAgentSilent`** — `absent_over_time([30m])`, `for: 15m`. **This is the one the tuning put at risk**, since it depends on Falco never going quiet and the tuning cut volume 78%. Measured rather than assumed: over 94 rolling 30m buckets since the tuning landed, the **minimum** was 53 lines and **zero** buckets were empty. It will not false-fire, and a genuinely dark agent still trips it.
- **The four pre-existing rules** (`SchedulerMissingVertxContext`, `ConfigPropertyMissingAtBoot`, `PanacheAssignedIdInsertOnly`, `FxRevaluationSkippedAllLegs`) and the one recording rule (`openbank:service_no_ready_endpoints:rate1m`) — these evaluate for the **first time ever** when this merges, which is the real risk of this change. All seven alert expressions plus the recording rule were dry-evaluated against live Loki: every one parses, and **every one returns zero series right now**. Turning the ruler on produces a quiet live control, not an activation storm.

## The guard

`check-loki-rules.py` was green throughout all of this, and could not have been otherwise: it boots Loki with a `LOKI_CONFIG` written to *mirror* `apps/loki.yaml`, so it validated the rules against the config the repo meant to deploy and was structurally blind to whether that was the config running.

Added `check_ruler_wiring()` — offline, no docker, no network, unconditional (it is not scoped by `--since`, because the failure it catches comes from editing `loki.yaml` without touching any rule file). It asserts the chart key is `rulerConfig` and not `ruler`, that rule storage is `local` pointing at the directory **derived from the sidecar's own `folder`** rather than restated, that `alertmanager_url` is non-empty, and that the ruler this gate simulates still matches the one deployed — so the load test cannot drift back into being green about a config nobody runs.

Falsified in both directions, which is the only thing that makes it worth having:

```
key reverted to `loki.ruler:`   ->  exit 1, names the defect
key as committed                ->  exit 0, ruler accepts all 3 groups
                                    (openbank.logs.falco-runtime, .silent-failures, .endpoint-availability)
--self-test                     ->  flags the pre-#5734 shape, accepts the committed one
```

## Verified / not verified

**Verified:** the live ruler config and its zero loaded groups; the rule files present on disk; both chart renders; that `/tmp` is a writable emptyDir mount so `rule_path` is safe; the pre/post-tuning by-rule split; the `FalcoAgentSilent` minimum-bucket floor; that all 8 committed rules parse and none currently fire; the gate's negative and positive cases; `run-gates.py --all` = 155 PASS (the two failures are local-environment artifacts — an unset `BASE_REF` and `sudo` for the oasdiff install — both pass when run properly, and neither touches this diff).

**Not verified:** that an alert reaches a human end-to-end. That needs this merged and synced, then a known-positive fired, and it is #5734's step 3 — I could not do it without mutating the cluster. **Until that is done the control is not proven**, and #5850's promotion criterion (flip `FalcoCriticalRuntimeDetection` to `critical` once it has fired on fewer than 1 day in 7) still stands. I also did not verify the Prometheus remote-write receiver accepts the ruler's recording-rule writes, only that the endpoint is the one already carrying k6 journey traffic.

## Follow-up, not done here

The residual 1156/6h of `Contact K8S API Server From Container` is **Strimzi's entity-operator** (`user-operator`, fabric8 `JdkHttpClientFa` under `tini`, ns `messaging`). #5850's exception is scoped `k8s.pod.name startswith "strimzi-cluster-operator"`, which does not cover it. That is a one-line tuning with the same justification as the existing macro, deliberately left out of this PR to keep it to the wiring defect and to avoid conflicting with `falco.yaml` hours after #5850 touched it. It is NOTICE, so it cannot reach any alert — log volume only.

Closes #5734

### DIFF
--- a/.github/scripts/check-loki-rules.py
+++ b/.github/scripts/check-loki-rules.py
@@ -221,8 +221,8 @@ def self_test(since: str | None = None) -> int:
     could invalidate it, and stays silent on the ones that cannot.
     """
     if not shutil.which("docker"):
-        print("::warning title=Loki rules::docker not available — cannot run the self-test.")
-        return 0
+        print("::warning title=Loki rules::docker not available — cannot run the LOAD-TEST half "
+              "of the self-test. The wiring half below needs no docker and still runs.")
     if since:
         changed, detail = rule_inputs_changed(since)
         if not changed:
@@ -231,6 +231,35 @@ def self_test(since: str | None = None) -> int:
             return 0
         print(f"check-loki-rules --self-test: {detail}")
     failures = []
+
+    # The wiring half, falsified first: it needs no docker, so it runs even where the load test
+    # cannot. Feed it the exact defect #5734 was — the block keyed `ruler` instead of
+    # `rulerConfig` — and require it to flag. A gate that has only ever seen a passing input has
+    # not been shown to be able to fail.
+    print("self-test: the wiring case the gate MUST flag")
+    with tempfile.TemporaryDirectory() as td:
+        app = yaml.safe_load(LOKI_APP.read_text(encoding="utf-8"))
+        loki = app["spec"]["source"]["helm"]["valuesObject"]["loki"]
+        loki["ruler"] = loki.pop("rulerConfig")          # the pre-#5734 shape, verbatim
+        broken = Path(td) / "loki-broken.yaml"
+        broken.write_text(yaml.safe_dump(app, sort_keys=False))
+        flagged = check_ruler_wiring(broken)
+    print(f"  {'ok  ' if flagged else 'FAIL'} `loki.ruler:` instead of `loki.rulerConfig:` is "
+          f"rejected" + (f" ({len(flagged)} problem(s))" if flagged else ""))
+    if not flagged:
+        failures.append("silently-dropped ruler key accepted")
+
+    print("self-test: the wiring case the gate MUST pass")
+    live = check_ruler_wiring()
+    print(f"  {'ok  ' if not live else 'FAIL'} the committed loki.yaml is accepted"
+          + ("" if not live else f" — {live[0]}"))
+    if live:
+        failures.append("committed loki.yaml rejected")
+
+    if not shutil.which("docker"):
+        # Reached only when docker appeared between the guard above and here; keep it honest.
+        return 1 if failures else 0
+
     good = {"good": yaml.safe_dump({"groups": [{
         "name": "selftest.good", "interval": "5m", "rules": [{
             "alert": "SelfTestGood",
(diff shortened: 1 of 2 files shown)


JiRaska:

### The core claim is verified independently, by rendering the chart

I did not take the key name on faith. I pulled `grafana/loki` **6.55.0** (the pinned `targetRevision`) and rendered `valuesObject` both ways with `helm template`:

| values shape | rendered `ruler:` in the Loki ConfigMap |
|---|---|
| `loki.rulerConfig:` (this PR) | `storage: {type: local, local: {directory: /rules}}`, `alertmanager_url: http://…alertmanager…:9093`, `enable_api: true`, remote_write present |
| `loki.ruler:` (pre-#5734) | `storage: {type: s3, …}`, **no `alertmanager_url`**, nothing else |

The chart's own `values.yaml` confirms it: `loki.*` contains `rulerConfig` and **no** `ruler` key. (There *is* a top-level `ruler:` — the distributed-mode component — so the old key was silently wrong in two ways at once.) The diagnosis in `apps/loki.yaml:113-126` is accurate as written; s3 rule storage and empty alertmanager_url are exactly what Helm produced.

`--self-test` passes here (docker present, all four cases green, including the new must-flag wiring case). `sidecar.rules.folder: /rules/fake` correctly sits under `storage.local.directory: /rules`.

### MAJOR — the gate misses the one other way to reproduce this exact defect

I mutation-tested `check_ruler_wiring()` against eight shapes. Six are caught. Two are not:

```
FLAGGED  empty alertmanager_url
FLAGGED  alertmanager_url removed
FLAGGED  storage.type -> s3
FLAGGED  storage.local.directory -> /other (sidecar mismatch)
FLAGGED  sidecar folder moved elsewhere
FLAGGED  rulerConfig removed entirely
MISSED   sidecar.rules.enabled: false
MISSED   enable_api: false
```

`sidecar.rules.enabled: false` (`apps/loki.yaml`, sidecar block) is the miss that matters. It puts the estate in **precisely the state #5734 describes** — a correctly configured ruler reading `/rules`, with nothing ever written there, zero groups loaded, no error anywhere — and `check_ruler_wiring()` returns clean. The docstring says the check asserts "a ruler capable of loading them exists at all"; it asserts the reader and not the writer. Three lines:

```python
if not ((values.get("sidecar") or {}).get("rules") or {}).get("enabled"):
    problems.append("sidecar.rules.enabled is not true — nothing writes rule files into "
                    f"{directory!r}, so the ruler loads zero groups. Same end state as #5734.")
```

`enable_api: false` is lower stakes but worth one line: it is what serves `/prometheus/api/v1/rules`, i.e. the endpoint the PR body cites as its own evidence. Turn it off and the next person to run this diagnosis gets a 404 instead of an answer.

### MAJOR (operational, not code) — merging this makes eight rules evaluate for the first time ever

On `origin/main` there are three `loki_rule` ConfigMaps, not two — `loki-rules-falco-runtime.yaml` (added by #5850) is the third, and the PR body counts it. Between them: 7 alerts + 1 recording rule that, by this PR's own finding, have **never evaluated once**. The moment `rulerConfig` renders, they all start, against up to `retention_period: 168h` of retained logs.

t=0 is fine per rule — I checked each. `count_over_time({…} |= "…" [15m]) > 0` over an absent stream yields empty, so nothing fires from a cold ruler; `FalcoAgentSilent`'s `absent_over_time({namespace="falco"}[30m])` queries the store, not in-memory state, so it is correct immediately rather than falsely firing. What is *not* a per-rule question is the aggregate: `SchedulerMissingVertxContext` (HR000068), `ConfigPropertyMissingAtBoot` (SRCFG00040) and `PanacheAssignedIdInsertOnly` ("duplicate key value violates") match strings this repo's own `CLAUDE.md` documents as endemic. Expect them to fire on merge, together, as a true-positive backlog.

Mitigating, and worth saying explicitly in the PR body: **all four silent-failure alerts and all three Falco alerts are `severity: warning`**, so this is a dashboard/notification burst, not a page. Worth a sentence anyway — "eight controls go live at once and some will fire immediately, by design" is the kind of thing that otherwise reads as a regression to whoever is on call.

### MINOR — the recording rule this PR resurrects has no consumer

`openbank:service_no_ready_endpoints:rate1m` (`loki-rules-endpoint-availability.yaml:53`) is `remote_write`-ed into Prometheus, and `git grep` across `origin/main` finds **exactly one** hit: its own `- record:` line. Nothing alerts on it, no dashboard reads it, no rule joins it. That is the same write-only-control shape this PR is fixing, one layer up — a rule that will now correctly evaluate and correctly deliver to a reader that does not exist. Not this PR's job to fix, but it belongs in the issue rather than being discovered again in six months.

### Checked, found nothing
- Every other key in `valuesObject` against the chart's `values.yaml`: no second silently-dropped key. The keys my diff flagged (`limits_config.*`, `rulerConfig.*`, `storage.bucketNames`) are all free-form passthrough blocks the chart templates wholesale, and the render confirms each one lands.
- `run_wiring_gate()` ordering in `main()`: it runs before the `--since` skip and before the docker guard, so neither path can silently omit it. `self_test(...) or wiring` still returns non-zero when either half fails.
- The `LOKI_CONFIG`-vs-deployed drift assertion (item 4) is real and derived, not restated.
- No secrets or new hostnames in the diff.

JiRaska:
Independent live confirmation of this diagnosis, gathered before I found this PR — posting because some of it is evidence the fix's effect can be measured against after merge.

**The live config disagrees with git exactly as you describe.** `GET /config` on `loki-0`:

| setting | `loki.yaml` says | live | 
|---|---|---|
| `alertmanager_url` | `http://kube-prometheus-stack-alertmanager…:9093` | **empty** |
| `storage.type` | `local` (`/rules`) | **s3** |
| `rule_path` | `/tmp/loki-rules` | `/var/loki/rules-temp` |
| `enable_api` | true | true (chart default, the only one that survived) |

So the ruler is reading an empty S3 bucket, and even a loaded rule would be delivered nowhere.

**Blast radius is three ConfigMaps, not one.** `loki_rule=1` currently matches `loki-rules-endpoint-availability`, `loki-rules-falco-runtime` and `loki-rules-silent-failures`. `/prometheus/api/v1/rules` returns `groups: []`, so **none has ever been evaluated** — the whole log-alerting estate, not just Falco.

**Two hypotheses I falsified on the way, so nobody re-runs them:**

1. *"Startup-only load — restart Loki and it picks them up."* Wrong. I restarted the StatefulSet; the sidecar re-synced and wrote all three files at 19:55:15Z, and the ruler stayed at 0 groups. (Note `sc-rules-volume` is an `emptyDir`, so a restart genuinely does wipe and re-sync it — that race is real but is *not* the cause.)
2. *"The rules volume isn't shared with the loki container."* Wrong. `sc-rules-volume` is mounted at `/rules/fake` in **both** containers, and `ls` from the sidecar shows all three files present.

Polled `/prometheus/api/v1/rules` every 30s for 5 minutes with the files on disk the whole time: 0 groups throughout. Combined with `storage.type: s3` above, that is conclusive — the local directory is not where it is looking.

**Chart key confirmed against the pinned version**, not from memory: `grafana/loki` 6.55.0 `values.yaml` line 549 defines `rulerConfig`, rendered by `{{ include "loki.rulerConfig" . }}` — and line 3423 states outright that `sidecar.rules.directories` requires `loki.rulerConfig.storage` to be local storage. `loki.ruler` matches nothing.

**Suggested post-merge verification**, since this is the class of fix that can land and still not take effect: after the sync, `GET /config` should show `alertmanager_url` populated and `storage.type: local`, and `/prometheus/api/v1/rules` should list three groups. Worth checking the config rather than only the PR being green — the whole defect here is that a thorough-looking config was inert.

PR #6036, opened 2026-08-20
feat(settlement): alert on settlements that stopped advancing (#5705)

### LINKED ISSUE #5705: settlement-service (money-path) emits no metrics and has no alerts of its own

JiRaska:
## What this is

The **progress half** of #5705 for `openbank-settlement-service`: a signal that can distinguish
*"settlement is running normally"* from *"this settlement stopped moving three hours ago"*.

`openbank-settlement-service` is a declared `money_path_service` (`rules.yaml`) that emitted no
application metric of any kind. Verified on `origin/main` before touching anything:

```
$ grep -rlE "micrometer|MeterRegistry" openbank-settlement-service/src/main
(no output — 20 .kt files)
```

It was not un-monitored, which is what made it hard to see. The pod is scraped by the fleet
PodMonitor, `PaymentServiceDown` (`prometheus-rules-tier1.yaml`) covers `namespace="payments"` and
therefore this container, and both Pyrra money-path SLOs exist for it. **Every one of those measures
whether the service ANSWERS.** None measures whether the debit → credit → ledger-book saga it
accepted PROGRESSES — and a settlement whose workflow stopped advancing answers nothing at all, its
originating request having returned 2xx in milliseconds, hours before.

## Relationship to #5723 — resolved by subtraction, not by competing

**#5723 is open, mergeable, and does the other half of this issue well.** These two PRs are
complementary and were made disjoint on purpose; they can land in either order.

| | #5723 (outcome) | this PR (progress) |
|---|---|---|
| question | *what did settlements do* | *is this settlement still moving* |
| signal | origination / saga-step / terminal-state counters | per-status age + count of non-terminal settlements |
| catches | rejection rate, compensation rate, step failures, "accepted N and finished none" | one settlement stranded in `DEBITED` while traffic flows normally around it |
| `PrometheusRule` object | `openbank-settlement-alerts` | `openbank-settlement-progress-alerts` |
| metrics | `openbank_settlement_{originated,saga_steps,terminal}_total` | `openbank_settlement_non_terminal{,_oldest_age_seconds}` |

An earlier draft of this branch also carried a saga-step counter. It was **removed**: #5723's port
strictly subsumes it (it carries currency, amount and cycle duration, which mine did not), and
shipping a weaker duplicate of a series that already has an open PR is how two copies of one rule
end up in the same file to drift apart later. What is left here is only what #5723 structurally
cannot provide.

One note on #5723 for its own reviewer, not a blocker for this PR: its `SettlementServiceDown`
duplicates `PaymentServiceDown`, which already owns
`up{job="observability/openbank-services", namespace="payments"}` — that is a second page for one
outage. This PR deliberately adds no `*ServiceDown` rule for the same reason.

## What was added

**`SettlementStrandedGauge`** (mirrors `DomesticPaymentStrandedGauge`, #3273 — same failure shape,
same reasoning):

- `openbank_settlement_non_terminal{service="settlement",status=…}` — how many settlements are in
  that state right now.
- `openbank_settlement_non_terminal_oldest_age_seconds{service="settlement",status=…}` — the age of
  the oldest, or `0` when the state is empty.

Published for all six non-terminal states (`PENDING`, `DEBITED`, `CREDITED`, and the three
`*_REVERSED` compensation states). `BOOKED` and `REJECTED` are not published: a terminal row's age
only grows and would alert forever. `DEBITED` is the state that matters most — the payer has been
debited, the payee has not been credited, and the funds are in neither account.

Supporting changes: `countByStatus` / `oldestCreatedAt` on `SettlementRepository`, and Flyway
`V2__settlements_status_created_at_index.sql` (composite `(status, created_at)`) so the 30s refresh
is not twelve sequential scans a minute. **Rollback note:** additive;
`DROP INDEX IF EXISTS idx_settlements_status_created_at`.

## Two alert rules, and why neither is vacuous at t=0 on a cold pod

`SettlementStrandedMidSaga` (critical) — `PENDING|DEBITED|CREDITED` older than **3h**, `for: 15m`.
`SettlementStuckAfterCompensation` (warning) — the three `*_REVERSED` states older than **8h**,
`for: 30m`.

**Thresholds are derived from the code, not guessed.** `SettlementWorkflowImpl` sets a per-activity
`scheduleToCloseTimeout` of 2h (`SCHEDULE_TO_CLOSE_HOURS`), so a settlement can legitimately sit in
any single state for up to 2h while its next activity retries; 3h clears that with an hour of slack.
The longest legitimate compensation chain starts at `LEDGER_REVERSED` and still owes
`reverseCredit` + `reverseDebit` + `rejectSettlement` — three activities at 2h each, so 8h clears
that 6h worst case.

**Cold-pod behaviour.** Every series is created in `@PostConstruct` at `0`, before any traffic, so a
freshly started pod publishes `0` per status rather than an absent series. A `> threshold` rule
therefore cannot fire during or after a deploy, and `absent()` over these names means *"the pod is
not exporting"* rather than *"no settlements"*. There is **no seeded age** anywhere in this gauge —
the boot reading and the healthy reading are the same number by construction, so there is no
`Instant.EPOCH` fourth state of the kind that made `WorkflowLivenessStale` fire 15 minutes after
every deploy (#2239 / #4208). An absent series would have been the real hazard: a PromQL comparison
over one matches nothing, which renders *identically* to a healthy zero — exactly how
`openbank_transaction_sagas_stuck_total` sat unfireable in this same file until #5733.

**The collector watches itself.** The gauges are only as trustworthy as the tick that refreshes
them; a frozen `refresh()` leaves every age stuck at its last value, which reads as healthy. So
`refresh()` records a success against `DomainMetrics.registerWorkflowLiveness`, whose 30s cadence
**is** a real cadence — the existing fleet `WorkflowLivenessStale` rule then covers the collector
without anyone asserting a cadence on settlement traffic, which is the thing settlement does not
have. `refresh()` is a `suspend fun`, not `runBlocking` in a plain `@Scheduled` method, so it
carries a Vert.x context and does not abort with `HR000068`.

## How I confirmed the series will exist

The failure this repo keeps paying for is a rule over a series nothing emits, and no CI gate checks
for it. So the tests assert the **rendered Prometheus scrape text**, as literals, not
`registry.get(...)`:

```kotlin
assertThat(scraped).contains(
    """openbank_settlement_non_terminal_oldest_age_seconds{service="settlement",status="DEBITED"} 0.0""",
)
```

That is the exact string `SettlementStrandedMidSaga` matches. The code spells
`openbank.settlement.non_terminal.oldest.age.seconds`; nothing else in CI checks that the dotted
name and the underscored one are the same series.

Also asserted: the age arithmetic against a fixed `Clock`; that an emptied state reports `0` and not
a stale age (an alert that keeps firing after the problem is fixed is how an alert earns being
ignored); and that the published status set equals *every* enum value minus the two terminal ones,
derived from `SettlementStatus.entries` rather than restated — so a status added later cannot
silently go unwatched.

**Negative control.** Deleting the two `gauge(...)` calls from `register()` turns
`renders the exact series the alert rules read, at zero, before any refresh` **red**
(`MeterNotFoundException`), and reverting restores green. The test can detect the absence of the
thing it tests.

## What remains

- **#5723's outcome counters** — origination, per-step and terminal-state rates. Deliberately not in
  this PR; if #5723 is closed rather than merged, that half of #5705 is still owed.
- **The compensation adapters are still stubs.** `reverseDebit` and `reverseCredit` only write a
  status row and log `"stub: wire reversal to balance-service"` — they do not actually reverse money
  at balance-service. `SettlementStuckAfterCompensation` will now make a stalled unwind visible, but
  the unwind itself is not yet real. That predates this PR and is not in its scope; it deserves its
  own issue.

Closes #5705

### DIFF
--- a/openbank-infra/gitops/components/payments/prometheus-rules.yaml
+++ b/openbank-infra/gitops/components/payments/prometheus-rules.yaml
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 # Per-service PrometheusRule alerts for the payments namespace (money-path, ADR-0077/0079).
 # Covers: transaction-service, clearing-service, swift-service, sepa-payment, sepa-instant,
-# domestic-payment, sdd-service. Complements the fleet-wide tier-1 rules with payment-rail
+# domestic-payment, sdd-service, settlement-service. Complements the fleet-wide tier-1 rules with payment-rail
 # specific saga and settlement observability.
 ---
 apiVersion: monitoring.coreos.com/v1
@@ -276,3 +276,94 @@ spec:
               {{ $value | humanizePercentage }} of fraud-scoring attempts on {{ $labels.service }}
               returned an invented verdict over the last 30m — a flapping fraud-service or an open
               circuit breaker, rather than a clean outage.
+---
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: openbank-settlement-progress-alerts
+  namespace: payments
+  labels:
+    app.kubernetes.io/part-of: openbank
+    app.kubernetes.io/name: settlement-service
+spec:
+  groups:
+    # settlement-service — a declared money_path_service (rules.yaml) — had NO rule of its own and
+    # emitted NO application metric at all until #5705. It was not un-monitored, which is what made
+    # it hard to see: the pod is scraped by the fleet PodMonitor, `PaymentServiceDown` in
+    # prometheus-rules-tier1.yaml covers `namespace="payments"` and so covers this container, and
+    # both Pyrra money-path SLOs exist for it. Every one of those measures whether the service
+    # ANSWERS. None measures whether the debit -> credit -> ledger-book saga it accepted PROGRESSES,
+    # and a settlement whose workflow stopped advancing answers nothing at all — its originating
+    # request returned 2xx in milliseconds, hours before.
+    #
+    # There is deliberately no SettlementServiceDown rule here: PaymentServiceDown already owns
+    # `up{job="observability/openbank-services", namespace="payments"}`, and a second rule over the
+    # same target would page twice for one outage.
+    #
+    # Scope: this document is the PROGRESS half of #5705 only — "is this settlement still moving".
+    # The OUTCOME half ("what did settlements do") is PR #5723's saga-step and terminal-state
+    # counters, which this deliberately does not duplicate; the two are complementary and the
+    # rule names, the metric names and the PrometheusRule object name are disjoint so they can
+    # land in either order. Nothing here is a throughput floor: that is the shape #5733 deleted
+    # twice from this file (ClearingSettlementWindowMissed, SwiftMtMessageProcessingStalled).
+    # settlement-service is REST- and Temporal-driven with no @Scheduled sweep and no @Incoming
+    # consumer, so an idle hour is the normal resting state, not a fault. What an idle service
+    # must still never have is a settlement that stopped advancing, and that is an age.
+    - name: openbank.settlement.progress
+      rules:
+        - alert: SettlementStrandedMidSaga
+          # PENDING / DEBITED / CREDITED are hand-off states the Temporal workflow drives through
+          # in seconds. DEBITED is the one that matters most: the payer has been debited, the payee
+          # has not been credited, and the funds are in neither account.
+          #
+          # 3h is derived from the code, not guessed. SettlementWorkflowImpl sets a per-activity
+          # scheduleToCloseTimeout of 2h (SCHEDULE_TO_CLOSE_HOURS), so a settlement can legitimately
+          # sit in any one of these states for up to 2h while its next activity retries before the
+          # workflow gives up and compensates. 3h clears that ceiling with an hour of slack, so this
+          # rule cannot fire on a slow-but-working retry — only on a settlement whose workflow is
+          # gone (worker not polling, Temporal unreachable, or a run lost).
+          expr: |
+            max by (status) (
+              openbank_settlement_non_terminal_oldest_age_seconds{
+                service="settlement", status=~"PENDING|DEBITED|CREDITED"
+              }
+            ) > 10800
+          for: 15m
+          labels:
+            severity: critical
+            service: settlement
+          annotations:
+            summary: "Settlement stranded in {{ $labels.status }} for over three hours"
+            description: >-
+              The oldest settlement in {{ $labels.status }} is {{ $value | humanizeDuration }} old,
+              past the 2h per-activity scheduleToClose ceiling of the settlement workflow. In DEBITED
+              the payer's funds have left and have not arrived; in CREDITED the money has moved but
+              was never booked to the general ledger. Check whether the Temporal worker for task
+              queue `openbank-settlement` is polling and whether a run exists for `settlement-<id>`.
+        - alert: SettlementStuckAfterCompensation
+          # The three *_REVERSED states mean the saga already unwound: money was moved and moved
+          # back. SettlementWorkflowImpl always calls rejectSettlement after compensating, so these
+          # states are transient by construction and a row parked in one has a workflow that died
+          # between the compensation and the terminal write — funds are safe, the record is not.
+          #
+          # 8h, again from the code: the longest legitimate compensation chain starts at
+          # LEDGER_REVERSED and still owes reverseCredit + reverseDebit + rejectSettlement, three
+          # activities at 2h scheduleToClose each. 8h clears that 6h worst case. Warning rather than
+          # critical because no customer money is in limbo — only the settlement record is.
+          expr: |
+            max by (status) (
+              openbank_settlement_non_terminal_oldest_age_seconds{
+                service="settlement", status=~"REVERSED|CREDITED_REVERSED|LEDGER_REVERSED"
+              }
+            ) > 28800
+          for: 30m
+          labels:
+            severity: warning
+            service: settlement
+          annotations:
+            summary: "Settlement stuck in compensation state {{ $labels.status }} for over eight hours"
+            description: >-
+              The oldest settlement in {{ $labels.status }} is {{ $value | humanizeDuration }} old.
+              Its compensations ran but it never reached REJECTED, so the settlements table now
+              disagrees with the ledger about a settlement that was unwound. Reconcile before the
+              next accounting-day close.
--- a/openbank-settlement-service/build.gradle.kts
+++ b/openbank-settlement-service/build.gradle.kts
@@ -18,6 +18,8 @@ dependencies {
     implementation(libs.quarkus.jdbc.postgresql)
     implementation(libs.quarkus.smallrye.health)
     implementation(libs.quarkus.micrometer.registry.prometheus)
+    // SettlementStrandedGauge's 30s refresh tick (issue #5705).
+    implementation(libs.quarkus.scheduler)
     implementation(libs.quarkus.opentelemetry)
     implementation(libs.quarkus.oidc)
     implementation(libs.quarkus.rest.client.reactive.jackson)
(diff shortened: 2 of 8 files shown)


JiRaska:

### t=0 on a cold pod: clean, and I checked the mechanism rather than the claim

This is the check I most expected to find a problem in, given #2239/#4208 (an `Instant.EPOCH`-seeded gauge firing 15 min after every deploy across 28 call sites). It does not recur here, and the reason is structural:

- `SettlementStrandedGauge.kt:117-118` — `counts` and `oldestAgeSeconds` are `AtomicLong(0)` at field init. Not `-1`, not a timestamp, not seeded from a query.
- `:93` `@Startup` + `:94` `@ApplicationScoped` + `:121` `@PostConstruct register()` — the series are registered **eagerly at boot**, so `@ApplicationScoped` laziness cannot defer them past the first scrape.
- `:142-147` first `refresh()` is `delayed = "15s"`.

So at t=0 every series exists at exactly `0.0`, and:
- `max by (status) (openbank_settlement_non_terminal_oldest_age_seconds{…PENDING|DEBITED|CREDITED}) > 10800` → `0 > 10800` → **false**
- `max by (status) (…{…REVERSED|CREDITED_REVERSED|LEDGER_REVERSED}) > 28800` → `0 > 28800` → **false**

Ages are computed per tick as `Duration.between(oldest, now)` with `?: 0L` for the empty state (`:152-155`) — no stored sentinel anywhere. The self-watch liveness gauge is genuinely boot-safe too: `DomainMetrics.kt:462` seeds `lastSuccessEpochMillis` at `Instant.now()` at registration, which is the post-#4208 fix and the KDoc says so.

`refresh()` is a **`suspend fun`** (`:147`), so the `@Scheduled`/`HR000068` trap does not apply and `check-no-runblocking-in-scheduled.py` passes on both criteria. The Micrometer suppliers (`:135`) read `AtomicLong` caches only — no DB work on the scrape thread.

### MAJOR — `build.gradle.kts` collides with open PR #5723, and the collision is silent

I compared file contents across both branches rather than trusting the diff. `openbank-settlement-service/build.gradle.kts:54` here and `:51` on `origin/feat/settlement-service-metrics` both add, immediately after `testImplementation(libs.mockk)`:

```kotlin
testImplementation("io.micrometer:micrometer-registry-prometheus:1.17.0")
```

Same coordinate, same version, same anchor line, different comment blocks. This is the exact "two branches add neighbouring entries to the same list" shape from `CLAUDE.md`, in its *harmless-looking* direction: whichever lands second either conflicts, or git resolves both additions and leaves **two identical `testImplementation` lines**, which Gradle accepts without a word. Both PRs are individually green; nothing in CI can see it.

Everything else about the two PRs is genuinely disjoint — I checked: PrometheusRule object names (`openbank-settlement-progress-alerts` vs `openbank-settlement-alerts`), group names, alert names and metric names do not overlap, and the `prometheus-rules.yaml` header edits are byte-identical so they merge cleanly. Whoever merges second should drop their copy of that one line.

### MINOR — no `absent()` companion, so one silent-failure mode stays uncovered

Both new rules are bare comparisons. If the bean never constructs — the `@Inject` secondary constructor resolves `Instance<MeterRegistry>` to non-resolvable, or ArC skips the bean — `register()` returns early at `:123` and **no series exists**. A bare comparison over an absent series does not fire; and the liveness heartbeat registered in `onStart` would be absent too, so `WorkflowLivenessStale` cannot fire either. Everything reads green while the control is gone. `PaymentServiceDown` covers pod-down, not pod-up-with-no-gauge.

The KDoc at `:69-76` and the PR body both reason about what `absent()` *would* mean over these names, but no `absent()` rule is added. Mitigating: the constructor shape is copied verbatim from `DomesticPaymentStrandedGauge.kt` (#3273), which is live and proven — so this is a latent gap inherited from a working pattern, not a new risk. Worth one more rule or an explicit note in the issue.

### MINOR — the migration comment overstates what it does
`V2__settlements_status_created_at_index.sql:8` says the index is "additive and non-locking in effect". A plain `CREATE INDEX` (no `CONCURRENTLY`) takes a `SHARE` lock and blocks **writes** to `settlements` for its duration. Negligible on today's table size; the sentence is still wrong as written, and it is the kind of sentence someone quotes back during a larger rollout.

### MINOR — stale comment naming a test that is not in this PR
`build.gradle.kts:45-46` says "SettlementStrandedGaugeTest **and SettlementActivitiesImplTest** assert the alert expressions". `SettlementActivitiesImplTest` is #5723's file and is not in this diff.

### Checked, found nothing

**Tests are not vacuous.** `SettlementStrandedGaugeTest.kt:72-78` asserts `4.0` and `Duration.ofHours(9).seconds` — chosen to cross the 10800s threshold, not merely to be non-zero. `:87-107` asserts `21600.0 → 0.0` after the state empties, which is the negative case for stale ages. `:147-159` asserts the literal rendered scrape line `openbank_settlement_non_terminal_oldest_age_seconds{service="settlement",status="DEBITED"} 0.0` — the exact series the rule selector matches, which is the dot-to-underscore check nothing else in CI does. `:131` derives the published set from `SettlementStatus.entries` minus the two terminal names rather than restating it. Deleting either `gauge(...)` call makes `:147` throw `MeterNotFoundException`. These tests can detect the absence of what they test.

**The IT is a real-DB test.** `SettlementRepositoryImplIT.kt:37-38` is `@QuarkusTest` + `@QuarkusTestResource(PostgresTestResource::class)` (Testcontainers), not a mocked repo. It inserts newest-first and asserts the **older** `createdAt` comes back, so it proves `order by createdAt asc` rather than non-nullity — and `toEntity()` copies `createdAt` verbatim instead of stamping `now`, so that assertion is meaningful.

**Query and index match.** `settlements.status` is `VARCHAR(32)` (V1:8) mapped to a Kotlin `String`, so `status.name` is the right parameter type. `count("status", …)` and `find("status = ?1 order by createdAt asc").firstResult()` are both served by `(status, created_at)`.

**Flyway.** No V2 collision — `origin/main` holds only `V1__create_settlements.sql` for this service, and no other open PR adds a settlement migration (#5723, #6048, #5990 checked). `quarkus.flyway.migrate-at-start: true` at `application.yaml:17-18` **and** a literal `quarkus.datasource.jdbc.url` at `:16` — both directions of `check-flyway-default-datasource.py` satisfied. This is not a committed-migration-nothing-runs case.

**Prometheus wiring.** It is a `PrometheusRule` CR (`monitoring.coreos.com/v1`), not a Helm values file, so the mis-nested-key family does not apply; rules are correctly under `spec.groups[].rules[]`, and `gitops/apps/payments.yaml:12` points ArgoCD at the directory, so an added document in an already-applied file needs no wiring change.

**Thresholds are code-derived as claimed.** `SettlementWorkflowImpl.kt:33` `SCHEDULE_TO_CLOSE_HOURS = 2L`; 3h and 8h check out against the stated 2h / 3×2h reasoning.

**Scope creep: none.** The saga-step counter the body says was removed is genuinely absent; no `*ServiceDown` rule was added; the compensation-stub limitation flagged as out of scope is being handled by #6048.

PR #6040, opened 2026-08-20
fix(fraud): dedupe both aggregates against the applied-id set, not a last-writer marker (#5789)

### LINKED ISSUE #5789: fraud: both aggregates dedupe with a last-writer marker, so an out-of-order DLQ replay still double-counts (payee_history has shipped this way since V3)
Both fraud aggregates guard redelivery with a **last-writer marker**, and both therefore miss an
out-of-order replay. `payee_history` has shipped this way since V3 and was treated throughout the
#5698 sweep as "the one that already guards"; PR #5786 deliberately mirrored it for
`velocity_aggregates` rather than invent a second idiom, and flagged the shared limit rather than
assume it away.

## The shape

Both use, verbatim:

```sql
WHERE EXCLUDED.last_transaction_id IS NULL
   OR <table>.last_transaction_id IS DISTINCT FROM EXCLUDED.last_transaction_id
```

That stores **the last id applied**, not the set of ids applied. So:

- **A, A** → second A matches the marker, WHERE is false, row untouched. Correct.
- **A, B, A** → after B the marker is B, so the replayed A is `DISTINCT FROM` it and applies again.
  Double-counted.

## Why it matters more now than it did

While the consumer swallowed every failure and acked (#5698), nothing was ever redelivered, so
neither guard was ever exercised. PR #5715 makes transient failures rethrow so the connector can
dead-letter, and #5751 wires the DLQ that actually parks the record. **A DLQ replay is by
construction not consecutive** — the record is parked, other events for the same account keep
flowing, and the replay lands later. That is precisely the A, B, A ordering the marker cannot catch.

So the guard is weakest in exactly the scenario the sweep created.

## Blast radius, both directions

- **Over-count** → velocity rules fire more readily; false positives against legitimate customers.
- **Under-count** → rules silently do not fire; the control is quietly absent.

Neither is visible without reconciling the aggregate against the source transactions. `payee_history`
feeds first-time-payee detection, so a spurious re-application there changes a fraud signal too.

## What would close this

An applied-event ledger rather than a marker — a per-aggregate set of applied `transaction_id`s
(bounded by retention or a window), checked on write. Needs deciding:

- retention bound (how long an id must be remembered to make a realistic DLQ replay safe),
- whether one ledger covers both tables or each keeps its own,
- and the migration + backfill story for rows already carrying only a marker.

Out of scope for #5786, which fixes the reachable defect (#5716) using the idiom already in the
codebase. This issue is the deeper fix for both tables at once.

Refs #5716, PR #5786, PR #5715, PR #5751.

JiRaska:
Closes #5789.

Both fraud aggregates deduplicated redelivery with a **last-writer marker** — `payee_history` since
V3, `velocity_aggregates` since V5 (#5786) — so an `A, B, A` delivery order found the marker holding
`B` when the replayed `A` arrived, `IS DISTINCT FROM` it, and applied `A` a second time. This
replaces the marker on both tables with the applied-event set the issue asked for.

## The ordering is reachable today, and not only through a DLQ replay

The issue frames the exposure as a DLQ replay, which depends on #5715 (rethrow) and #5751 (DLQ
wiring) — both still **open**. The defect does not wait for them:

- `openbank.transactions.transaction.initiated` is published with the **transaction aggregateId** as
  the Kafka key (`OutboxKafkaHeaders.partitionKey` = `entry.aggregateId`). Two signals for the same
  account are therefore two different keys — Kafka orders per key, and nothing orders two signals
  for one account relative to each other.
- The consumer is at-least-once with periodic offset commits. Any rebalance, pod roll or crash
  re-delivers the whole uncommitted window from the last committed offset. On the deployed
  single-partition topic (`gitops/components/payments/kafka-topic.yaml`, `partitions: 1`) a window
  containing `A, B` for the same account replays as `A, B` — and after the original `B` the marker
  is `B`, so **both** re-apply.

The marker therefore protected a replay window of exactly **one record**. The negative control below
measures precisely that: replaying a two-record window took `payment_count` from 2 to 4.

## Nothing noticed it, and nothing noticed the guard working either

`FraudMetricsPort` had only `recordVerdict` and `recordShadowScore`. A suppressed replay writes no
row, logs nothing and moves no counter, and there is no reconciliation of either aggregate against
the source transactions — so "the guard is working" and "the guard is absent" were indistinguishable
from outside the database, in both directions (over-count and under-count). This PR adds
`openbank_fraud_signal_replay_suppressed_total{aggregate}`, incremented when the upsert returns
`rowCount 0`, i.e. exactly when a re-application was suppressed. It is a distinct signal, not a flag
shared with the success path.

## The change

- **`V6__applied_signal_ledger.sql`** — `applied_transaction_ids UUID[]` on both tables, backfilled
  from each row's existing `last_transaction_id` so today's protection is not lost at cutover, plus
  an `IMMUTABLE` `fraud_append_applied(ids, new_id, cap)` that appends and trims.
- **Both repositories** — the upsert's `WHERE` now tests **membership of the applied set** instead of
  equality with the newest entry, and appends the id in the same statement, so the memory of
  "applied" and the effect of applying commit together. A signal with no aggregateId is still applied
  unconditionally and is not remembered (no identity to deduplicate by). The guard stays **per row**,
  so the three velocity windows still converge independently after a partial failure.
- The set is unioned with `last_transaction_id` when tested. That covers the **rolling-deploy window
  a backfill cannot**: a pod still running the V5/V3 code writes the marker and leaves the new column
  untouched, so a row it applies after the migration runs would otherwise remember nothing. The
  union makes the new guard strictly stronger than the old one in every state, never weaker. This
  was not designed in — it was found by the pre-existing partial-failure IT going red, which builds
  exactly that legacy row shape by hand.

## The design decision, stated rather than picked silently

The issue left three things open. What this PR decides:

**Bound: a COUNT window, not a duration, and not a ledger table.** The set is the most recent `N` ids
applied to that row, `N` from `openbank.fraud.applied-signal-window` (default 100).

- *Why not a time-bounded ledger table:* it needs a retention sweep to stay bounded, and a scheduler
  that silently stops running is exactly how this repo has lost controls before. A per-row array is
  trimmed on every write, so it cannot grow without anything running.
- *Why one column per table rather than one shared ledger:* keeping the set on the aggregate row is
  what lets the claim and the increment be one statement against one row — no second table to keep
  atomic with the first, and no key-encoding to keep in sync between Kotlin and a backfill.
- *Cost, and it is a real one:* the guarantee is expressed in events rather than in time. A replay
  arriving after `N` further signals to the **same row** is applied again. For `velocity_aggregates` a
  row is one (account, window, currency, bucket), so `N=100` is 100 transactions inside that bucket;
  for `payee_history` it is 100 payments to the same payee. If a duration guarantee is later required
  (e.g. "any replay within the topic's 7-day retention"), that is a ledger table and a sweep, and it
  is a different PR — this one does not foreclose it.

This bound is **asserted, not just documented**: `the applied-signal window bounds how far back a
replay can be suppressed` sets the window to 1, shows the guard degrading exactly to the old
last-writer marker, and shows `A, B, A` double-counting again.

## Verification

Real-Postgres ITs (`PostgresRedisTestResource`), not mocked repositories — a mocked `PgPool` cannot
tell an upsert that double-counts from one that does not.

Four new failing-first tests construct the actual out-of-order delivery on both tables: the `A, B, A`
replay, and the whole-uncommitted-window `A, B, A, B` replay.

**Negative control** — reverting *only* the two production `WHERE` clauses back to
`IS DISTINCT FROM`, keeping the migration, the tests and the metric:

```
PayeeHistoryRepositoryImplIT        10 tests, 2 failed
  RED: an out-of-order replay after another payment is not counted twice   expected: 2 but was: 3
  RED: replaying a whole uncommitted window counts neither payment twice   expected: 2 but was: 4
VelocityAggregateRepositoryImplIT   10 tests, 2 failed
  RED: an out-of-order replay after another signal is not applied twice    expected: 2 but was: 3
  RED: replaying a whole uncommitted window applies neither signal twice   expected: 2 but was: 4
```

Exactly the four new tests go red and nothing else does. Restored, the module is
`175 tests / 0 failures / 0 errors / 1 skipped` (the skip is the broker-gated
`FraudPactProviderVerificationTest`, which is `@EnabledIfSystemProperty(pactbroker.url)` off
main-push — the `@PactFolder` sibling ran).

Also run: `:openbank-fraud-service:quarkusBuild` (CDI/ArC wiring of the two new `@ConfigProperty`
constructor parameters), `ktlintCheck`, `detekt`, and the repo gate manifest.

## Not verified

- No live system was touched. The claim that the deployed topic is single-partition is read from
  `gitops/components/payments/kafka-topic.yaml`, not from a broker.
- The migration is exercised by Flyway on the test container, not against a populated database; the
  backfill is asserted indirectly (the partial-failure IT builds a legacy-shaped row by hand).
- No attempt was made to establish whether double-counting has already occurred in a real
  environment — that needs the aggregate reconciled against the source transactions, which is the
  reconciliation this change makes possible to alert on but does not itself add.

## Noticed in passing, deliberately not fixed here

`VelocityAggregateRepositoryImpl` picks `window_start` from `Instant.now(clock)` at **processing**
time, not from the event's `occurredAt`. A replay that crosses an hour boundary therefore lands in a
different bucket row from the original, where no guard of any design can recognise it. That is a
distinct defect (processing-time vs event-time bucketing) with a different fix and a different blast
radius, and folding it in would change what the aggregate means.

### DIFF
--- a/openbank-fraud-service/src/main/kotlin/com/openbank/fraud/application/port/out/FraudMetricsPort.kt
+++ b/openbank-fraud-service/src/main/kotlin/com/openbank/fraud/application/port/out/FraudMetricsPort.kt
@@ -25,4 +25,15 @@ interface FraudMetricsPort {
      * model is calibrated and safe before any enforce phase — it never reflects a honoured decision.
      */
     fun recordShadowScore(score: Double)
+
+    /**
+     * Record that a transaction signal was recognised as ALREADY APPLIED to [aggregate] and was
+     * therefore not applied again (issue #5789). This is the only series that makes redelivery
+     * visible at all: a suppressed replay writes no row, logs nothing and changes no counter, so
+     * without this the guard working and the guard being absent look identical from outside the
+     * database. Read it as "how often is the dedupe guard load-bearing" — a sustained non-zero rate
+     * is normal after a rebalance or a DLQ replay; a permanently flat zero while redeliveries are
+     * known to happen means the guard is not being reached.
+     */
+    fun recordSignalReplaySuppressed(aggregate: String)
 }
--- a/openbank-fraud-service/src/main/kotlin/com/openbank/fraud/application/port/out/FraudMetricsPort.kt
+++ b/openbank-fraud-service/src/main/kotlin/com/openbank/fraud/application/port/out/FraudMetricsPort.kt
@@ -36,4 +36,18 @@ interface FraudMetricsPort {
      * known to happen means the guard is not being reached.
      */
     fun recordSignalReplaySuppressed(aggregate: String)
+
+    /**
+     * Record that a transaction signal arrived with **no `occurredAt`** and that processing time was
+     * substituted for it (issue #6044). Every velocity bucket and payee-history row written from
+     * such a signal asserts a business time nobody measured, and — because the bucket is part of the
+     * primary key — a replay of that same signal will not land in the same row, so the redelivery
+     * guard cannot reach it. The substitution is otherwise completely silent: it writes a normal row
+     * and logs nothing, exactly like the ingest-time substitution #3883 found in the audit
+     * consumer. This counter is the only thing that distinguishes "every producer sends an event
+     * time" from "we have been inventing them". `openbank.transactions.transaction.initiated`
+     * declares `occurredAt` as required, so the expected value is a flat ZERO; any non-zero rate is
+     * a producer defect, not a tolerable degradation.
+     */
+    fun recordSignalMissingEventTime()
 }
--- a/openbank-fraud-service/src/main/kotlin/com/openbank/fraud/application/port/out/VelocityAggregateRepository.kt
+++ b/openbank-fraud-service/src/main/kotlin/com/openbank/fraud/application/port/out/VelocityAggregateRepository.kt
@@ -7,6 +7,7 @@ package com.openbank.fraud.application.port.out
 import com.openbank.fraud.domain.model.VelocityAggregate
 import com.openbank.fraud.domain.model.VelocityWindow
 import java.math.BigDecimal
+import java.time.Instant
 import java.util.UUID
 
 /** Stores and queries per-account rolling velocity aggregates (ADR-0084 §2). */
@@ -22,8 +23,25 @@ interface VelocityAggregateRepository {
      * not applied — the windows converge independently rather than as one transaction. A null
      * [transactionId] carries no identity and is therefore never deduplicated (same contract as
      * [PayeeHistoryRepository.recordPayment]).
+     *
+     * The window bucket is derived from [occurredAt] — the event's own business time — and never
+     * from the moment the signal is processed (issue #6044). Processing time puts a redelivery that
+     * crosses an hour, day or week boundary into a DIFFERENT row from the original, and the
+     * idempotency above is per row, so no guard of any design could recognise it there. Bucketing on
+     * event time is therefore what makes the redelivery guard reach a delayed replay at all; it also
+     * makes the aggregate mean "transactions that OCCURRED in this window" rather than "signals that
+     * were PROCESSED in it", which is the only reading that reconciles against the source
+     * transactions. The caller is responsible for supplying a real event time; see
+     * [com.openbank.fraud.infrastructure.messaging.TransactionSignalConsumer] for what it does when
+     * the event carries none.
      */
-    suspend fun recordTransaction(accountId: UUID, amount: BigDecimal, currency: String, transactionId: UUID?)
+    suspend fun recordTransaction(
+        accountId: UUID,
+        amount: BigDecimal,
+        currency: String,
+        transactionId: UUID?,
+        occurredAt: Instant,
+    )
 
     /**
      * Returns the current aggregate for [accountId] in [window] for [currency], or null.
--- a/openbank-fraud-service/src/main/kotlin/com/openbank/fraud/infrastructure/messaging/TransactionSignalConsumer.kt
+++ b/openbank-fraud-service/src/main/kotlin/com/openbank/fraud/infrastructure/messaging/TransactionSignalConsumer.kt
@@ -6,6 +6,7 @@ package com.openbank.fraud.infrastructure.messaging
 
 import com.fasterxml.jackson.core.JacksonException
 import com.fasterxml.jackson.databind.ObjectMapper
+import com.openbank.fraud.application.port.out.FraudMetricsPort
 import com.openbank.fraud.application.port.out.PayeeHistoryRepository
 import com.openbank.fraud.application.port.out.VelocityAggregateRepository
 import com.openbank.fraud.application.usecase.FeatureOnlineUpdater
@@ -51,6 +52,7 @@ class TransactionSignalConsumer(
     private val featureUpdater: FeatureOnlineUpdater,
     private val objectMapper: ObjectMapper,
     private val clock: Clock,
+    private val metrics: FraudMetricsPort,
 ) {
     private val log = Logger.getLogger(TransactionSignalConsumer::class.java)
 
@@ -70,15 +72,31 @@ class TransactionSignalConsumer(
         val accountId = signal.sourceAccountId ?: return
         val amount = signal.amount ?: BigDecimal.ZERO
         val currency = signal.currencyCode ?: "CZK"
+        // Issue #6044. occurredAt is REQUIRED on TransactionInitiatedEvent, so the fallback should
+        // never be taken; if it is, every row written from this signal carries a business time
+        // nobody measured, and the redelivery guard cannot reach a later replay of it (the replay
+        // would substitute a different processing time and so target a different velocity bucket
+        // row). Substituting silently is what made the audit consumer's ingest-time fallback
+        // invisible for 7 of its 21 topics (#3883); this one is counted, so the absent case is
+        // visible from outside the database instead of being indistinguishable from a healthy one.
+        val eventTime = signal.occurredAt ?: run {
+            metrics.recordSignalMissingEventTime()
+            log.warnf(
+                "transaction.initiated signal for account %s carries no occurredAt; substituting processing time. " +
+                    "Velocity bucketing and redelivery dedupe are both degraded for this signal (#6044).",
+                accountId,
+            )
+            Instant.now(clock)
+        }
         @Suppress("TooGenericExceptionCaught") // DB / reactive layer exceptions have no common base
         runBlocking {
             try {
-                velocityRepo.recordTransaction(accountId, amount, currency, signal.aggregateId)
+                velocityRepo.recordTransaction(accountId, amount, currency, signal.aggregateId, eventTime)
                 log.debugf("Velocity aggregate updated for account %s amount=%s %s", accountId, amount, currency)
             } catch (ex: Exception) {
                 log.warnf(ex, "Failed to record velocity aggregate for account %s", accountId)
             }
-            recordPayeeHistory(accountId, signal)
+            recordPayeeHistory(accountId, signal, eventTime)
             // ADR-0140 online feature store (shadow plane). Best-effort: a feature-store failure must
             // never break the velocity path. Skip silently when the event carries no business time.
             val occurredAt = signal.occurredAt ?: return@runBlocking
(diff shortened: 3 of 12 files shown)


JiRaska:
Stacked follow-up for the residual this PR names in "noticed in passing, deliberately not fixed here": #6061 (closes #6044) buckets `velocity_aggregates` by the event's `occurredAt` instead of `Instant.now(clock)`, based on this branch. Merge this one first.

JiRaska:
## Review — blocking finding (money-path)

I re-ran the negative control against a real PostgreSQL 16 (not the mocked repo), and the good news first: **the claimed defect is real and this fix genuinely closes it.** Reproducing the exact `A, B, A` sequence from `V6__applied_signal_ledger.sql:1-10` against both the old and the new `WHERE`:

```
OLD guard (main, last-writer marker):  count=3  amount=300   <- double-counted
NEW guard (this PR, applied-id set):   count=2  amount=200   <- correct
```

The control discriminates, the trim keeps the newest N (verified `cap=3` over 6 sequential ids: `{4,5,6}`), the V1→V6 chain applies cleanly, and `INSERT 0 0` confirms the `result.rowCount() == 0` metric branch in `VelocityAggregateRepositoryImpl.kt:53` and `PayeeHistoryRepositoryImpl.kt:47` is a real signal and not a guess. There is no V6 collision — I checked every one of the 84 open PR branches, and the only other `db/migration` files in flight are settlement V2 (#6036), referral V1 (#5979) and party V19 (#5951). Rollback note present.

### Blocking: the new `WHERE` is not NULL-safe, and a single null-`aggregateId` signal permanently freezes the row

`VelocityAggregateRepositoryImpl.kt:118-120` and `PayeeHistoryRepositoryImpl.kt:101-103` replace

```sql
last_transaction_id IS DISTINCT FROM EXCLUDED.last_transaction_id   -- NULL-safe
```

with

```sql
NOT (EXCLUDED.last_transaction_id = ANY (
     array_append(applied_transaction_ids, last_transaction_id)))   -- NOT NULL-safe
```

`array_append(..., last_transaction_id)` puts a **NULL element** into the array whenever `last_transaction_id IS NULL`. In SQL, `x = ANY(array)` with no match but a NULL present evaluates to **NULL**, not FALSE; `NOT NULL` is NULL; and `WHERE NULL` does not update. Measured directly:

```sql
SELECT NOT ('aaaa…' = ANY (ARRAY['cccc…'::uuid, NULL::uuid]));  -- => NULL
```

`TransactionSignalConsumer.kt:24` declares `aggregateId: UUID? = null`, so a signal with no aggregateId is reachable — and the migration header explicitly treats it as a live case. Once one such signal lands, `last_transaction_id` is set to NULL and **is never overwritten**, because every subsequent statement now evaluates the guard to NULL and is skipped. The row is permanently frozen. End-to-end against the real upsert:

```
normal txn C, then a signal with NULL id, then two further LEGITIMATE txns D and E
  NEW guard: count=2 amount=200   <- D and E silently dropped (INSERT 0 0, forever)
  OLD guard: count=4 amount=400   <- correct
```

This is a regression against `main`, and its direction is the dangerous one: the velocity aggregate for that `(account, window, currency, bucket)` stops incrementing, so the velocity rule silently never fires. It also increments `openbank.fraud.signal.replay.suppressed`, so the one metric added to make this observable actively **misreports the freeze as the guard working**.

Suggested fix — strip NULLs before the membership test, in both files:

```sql
NOT (EXCLUDED.last_transaction_id = ANY (
     array_remove(array_append(applied_transaction_ids, last_transaction_id), NULL)))
```

`array_remove(..., NULL)` is NULL-aware in PostgreSQL and yields an array with no NULL element, so `= ANY` returns a proper FALSE and the row updates. (The PR already relies on this behaviour at `VALUES … array_remove(ARRAY[$6::uuid], NULL)`.)

### Why the suite did not catch it

`VelocityAggregateRepositoryImplIT.kt:134` — *"a signal without a transaction id is never deduplicated"* — sends **two NULL signals back to back**. Both take the `EXCLUDED.last_transaction_id IS NULL` branch, which short-circuits before the broken clause, so the test passes. The uncovered case is a NULL signal **followed by a non-NULL one**. Please add exactly that ordering; it is the shape the whole PR is about (`A, B, A` is already tested — this is `NULL, D`).

### Non-blocking

- The two `UPDATE … SET applied_transaction_ids = …` statements in V6 rewrite every row in both tables. Worth a note in the rollback section on expected duration/bloat for the live table sizes; the `ALTER TABLE ADD COLUMN` itself is metadata-only on PG11+, so it is the backfill that costs.
- The `IMMUTABLE` marking on `fraud_append_applied` is correct (pure over its arguments) — checked, no issue.
- `TransactionSignalConsumer.kt` is being edited concurrently by #5715, #5737 and #6061 from the same base blob. Not this PR's problem directly, but the four will not stay mergeable in any order without someone re-reading the merged result.

JiRaska:
## P0 fixed on this branch: the new membership test was not NULL-safe

Credit to the code review on this PR for finding it, and for reproducing it against real PostgreSQL rather than reasoning about it. Pushed as `7d9e07c38`.

### The defect

The guard this PR introduced tests:

```sql
NOT (EXCLUDED.last_transaction_id = ANY (
     array_append(applied_transaction_ids, last_transaction_id)))
```

`last_transaction_id` is NULL for the whole remaining life of a row once a signal that carried no `aggregateId` has passed through it (the `DO UPDATE` sets `last_transaction_id = EXCLUDED.last_transaction_id` unconditionally). In PostgreSQL, `x = ANY (<array containing NULL>)` evaluates to **NULL, not FALSE**, when `x` matches no element. So `NOT (...)` is NULL, the `ON CONFLICT ... WHERE` is not true, and **every later genuinely-new signal to that row is silently dropped forever** — `INSERT 0 0`, no error, no log line. The row freezes.

Direction of harm: velocity and payee counts **undercount**, so a fraud rule that should fire does not. Against `main` this is a **regression** — the `IS DISTINCT FROM` form it replaced was NULL-safe.

Why this PR's own tests could not see it: `a signal without a transaction id is never deduplicated` uses two NULLs back to back, and both take the `EXCLUDED.last_transaction_id IS NULL` branch, which short-circuits before the membership test is ever evaluated.

### Reproduced first, then fixed

Interleaved sequence `C, NULL, D, E` against real PostgreSQL via the existing `@QuarkusTestResource(PostgresRedisTestResource)` ITs, with the fix absent:

```
velocity_aggregates: [count for window H1] expected: 4L but was: 2L
payee_history:                             expected: 4L but was: 2L
```

D and E were both dropped. That is the negative control as well: the new tests were written and run before the fix existed, and they were red with exactly that text. With the fix applied both classes are `tests=12 skipped=0 failures=0 errors=0`.

### The fix

`array_remove(array_append(...), NULL)` on both aggregates' `WHERE`, as suggested in the review — checked rather than taken on trust:

- `applied_transaction_ids` can never itself hold a NULL element by design. The INSERT path already writes `array_remove(ARRAY[$n::uuid], NULL)`, `fraud_append_applied` explicitly returns `ids` unchanged when `new_id IS NULL`, and the V6 backfill uses `array_remove`. **The union with the scalar marker is the only route by which a NULL reaches this array** — so `array_remove` is scoped exactly to the defect and removes nothing that was meant to be there.
- The documented contract is preserved: a signal with no `aggregateId` is still applied unconditionally (the `IS NULL` branch is untouched) and is still not remembered (the SET path is untouched).
- No migration change was needed — `V6` is correct as written; the bug was only in the two Kotlin `UPSERT_SQL` constants.

### Tests

Two new ITs per aggregate, four in total:

- `a signal / payment without a transaction id does not freeze the row against later signals` — the interleaved `C, NULL, D, E` case. Cannot pass against the broken clause.
- `dedupe still suppresses a replay after a signal / payment without a transaction id` — the other side, so that a membership test made "NULL-safe" by always evaluating true would not pass either.

### Verified

`:openbank-fraud-service:ktlintCheck :openbank-fraud-service:detekt :openbank-fraud-service:test :openbank-fraud-service:quarkusBuild --rerun-tasks`, serialized, exit 0. Across 26 test classes: **179 tests, 0 failures, 0 errors, 1 skipped** — the one skip is `FraudPactProviderVerificationTest`, which is `@PactBroker`-sourced and gated on `pactbroker.url`, as designed. Gate groups `lint kotlin data api registry security supplychain gitops` run per shard: all pass except `api-contract-gate` (fails installing `oasdiff`, needs interactive `sudo`) and `loki-rule-load-test` (`BASE_REF: unbound variable`) — both reproduce identically on the parent commit `1fa63a35e`, so they are local-environment artifacts, not this change.

### Impact on #6061

#6061 stacks on this branch and touches the same method. **No textual conflict** — it changes `recordTransaction`'s signature and bucket derivation, not the `UPSERT_SQL` companion block; `git merge-tree` produces zero conflict markers.

There is a **semantic** one, though: #6061 adds a required `occurredAt: Instant` parameter to `recordTransaction`, and my two new velocity tests are new 4-argument call sites it has not seen. Once #6061 picks this commit up, `VelocityAggregateRepositoryImplIT` will not compile until those two calls are given an `occurredAt`. Small follow-up on #6061, but it will not be caught by the merge — only by the build.

Issue #6035, opened 2026-08-20
billing-service is the last unaudited money-path producer — openbank.billing.fee.event is absent from audit-service's subscription (4 of 5 landed)


JiRaska:
Follow-up to #5859, which asked whether any other money-path producer's topic is missing from audit-service's subscription list. It is — five more, and the answer was reached by the same three-place probe (`topics:` list, `EventAttribution`, KafkaUser Read ACL).

## Method

Verified on `origin/main` @ `441c54b21`.

audit-service has exactly one incoming channel, `audit-events-in` (`openbank-audit-service/src/main/resources/application.yaml:116`), so its `topics:` list at line 151 is the entire subscription surface — there is no second channel picking anything else up.

Each `money_path_services` entry (`openbank-libs/governance/rules.yaml:717`) was resolved to the topic it produces via its `mp.messaging.outgoing.<channel>.topic` key, **not** by grepping topic literals: this repo builds event payloads two ways (hand-built maps vs serialised data classes), and a literal grep returns a false clean for the second idiom.

## Result

| producer | topic | subscribed | `EventAttribution` | Read ACL |
|---|---|---|---|---|
| ledger-service | `openbank.ledger.journal.posted` | no | no | no |
| sdd-service | `openbank.sdd.event` | no | no | no |
| interest-service | `openbank.interest.accrual.event` | no | no | no |
| fraud-service | `openbank.fraud.hold.changed` | no | no | no |
| billing-service | `openbank.billing.fee.event` | no | no | no |

Producer proof, one per row:
- `openbank-ledger-service/src/main/resources/application.yaml:109-113`, emitter `KafkaLedgerOutboxEventPublisher.kt:32`
- `openbank-sdd-service/src/main/resources/application.yaml:155-159`, emitter `KafkaSddOutboxEventPublisher.kt:26`
- `openbank-interest-service/src/main/resources/application.yaml:215-219`, emitter `KafkaInterestOutboxEventPublisher.kt:35`
- `openbank-fraud-service/src/main/resources/application.yaml:132-136`, emitter `KafkaFraudOutboxEventPublisher.kt:20`
- `openbank-billing-service/src/main/resources/application.yaml:122-131`, emitter `LedgerOutboxEventPublisher.kt:48`

[...]

Issue #5859, opened 2026-08-20
audit-service never subscribed to openbank.delegation.events — no delegation event has ever reached the audit trail


JiRaska:
`audit-service` does not subscribe to `openbank.delegation.events`. Verified on `origin/main` in all
three places a subscription needs to exist:

- absent from audit-service's incoming topic list in `application.yaml`
- absent from `TopicAttribution`
- absent from the audit KafkaUser's Read ACL in gitops

(The delegation mentions in `audit-opa-bundle.yaml` are authorization policy, a different thing.)

## What that means

**No delegation event has ever reached the audit trail** — not just the spend-reservation events that
#5728 found missing at the producer, but the eight lifecycle events that *are* emitted and have been
all along: grant, offer, accept, revoke, suspend and the rest.

`openbank-delegation-service` is in `money_path_services`, and per ADR-0232 it mints payment rights —
customer-to-party grants that let one party spend another's money. ADR-0249 D4 states *"every grant,
reservation, confirmation and revocation is an audit event"*. The producer side of that claim is
partly true; the consumer side has never been wired, so the audit trail is empty regardless.

Two independent failure points, which is why fixing one would not have been enough: without the topic
in the consumer's list nothing is read, and without the Read ACL the subscription is **ACL-denied at
runtime** — the same shape as #5338.

## Why nothing surfaced it

An unsubscribed topic produces no error anywhere. audit-service is healthy, delegation-service is
healthy, the events are on the wire, and the audit trail simply has no rows for them — which is
indistinguishable from "no delegation activity happened". Same class as this repo's other
write-only controls (#5734 Falco, #5761 dead-letter sink, #5769 pentest attestation): the control
reports healthy because it structurally cannot report anything else.

## Status

PR #5857 fixes all three places as a side-effect of adding the spend-reservation events, since those
events would otherwise have landed nowhere either. This issue records the finding in its own right —
the gap predates #5728 and is broader than it, and it is worth knowing whether any compliance
reporting has been relying on delegation audit rows that were never written.

Worth checking as follow-up: whether any other money-path producer's topic is missing from
audit-service's subscription list. The same probe (topic list, `TopicAttribution`, Read ACL) applied
across all producers would answer it, and one gap of this kind implies the set was never enumerated.

Refs #5728, PR #5857, ADR-0232, ADR-0249 D4, #5338.

JiRaska:
[...]

PR #6047, opened 2026-08-20
test(audit): prove and gate that every money-path produced topic reaches the audit trail

JiRaska:
## What this adds

#5859 found audit-service had never subscribed to `openbank.delegation.events`. #6035 applied the
same three-place probe across `money_path_services` and found five more, each missing from **all
three** places at once — the signature of a set that was never enumerated. PR #5857 fixes the
delegation row; nothing yet stops the next one.

Two complementary halves, neither of which can do the other's job:

1. **`.github/scripts/check-audit-money-path-subscription.py`** — derives the whole coverage set and
   fails when a money-path produced topic is missing from any of the three places.
2. **`AuditSubscriptionSurfaceIT`** — the runtime proof, pushing records *into* the channel rather
   than calling the consumer.

This change is additive and does not touch PR #5857's files. The gate baselines
`openbank.delegation.events` with #5857 named as its owner, and the test reads the topic list from
configuration at runtime, so #5857's topic is covered the moment it lands, with no edit here.

## 1. The gate

Nothing about the scope is hand-kept:

| what | derived from |
|---|---|
| money-path services | `openbank-libs/governance/rules.yaml: money_path_services` |
| each one's PRODUCED topics | its own `mp.messaging.outgoing.<channel>.topic` |
| audit's subscribed topics | `mp.messaging.incoming.audit-events-in.topics` |
| audit's attribution table | `TopicAttribution` in `EventAttribution.kt` |
| audit's Read ACLs | the `audit-service` KafkaUser under `openbank-infra/gitops` |

Producers are resolved from the connector's own channel declaration, **never** by grepping topic
literals: this repo builds event payloads two ways and a serialised data class has no literal
anywhere in the tree, so a grep returns a confident false clean over exactly the producers it
structurally cannot see (#3883). A money-path service that declares no outgoing channel at all —
settlement and vop — is reported as *correctly absent*, and that distinction is derived from the
config rather than declared.

Two hand-kept lists remain, and each is the thing a human has to justify:

- **`KNOWN_GAPS`** — today's debt, keyed `<service>#<topic>#<place>` with its owning issue. It may
  only shrink; an entry that is no longer a gap is itself an error.
- **`IN_FLIGHT`** — one topic against one open PR. `openbank.delegation.events` is a gap today and
  will not be once #5857 lands; baselining it in `KNOWN_GAPS` would red-gate `main` the instant that
  PR merges, not because anything regressed but because two PRs were open at once. `IN_FLIGHT`
  accepts both states, and prints a `::warning` naming the PR the moment the gap closes, so the
  entry announces its own retirement instead of waiting to be noticed.
- **`OUT_OF_SCOPE`** — `openbank.notification.requests`, which is notification-service's own inbox,
  a fan-in command topic four services write onto, not any producer's event stream. It also fails
  when stale, so an exclusion cannot outlive its reason.

### It found two the issue did not

`openbank-standing-order-service` (`openbank.standing-orders.order.event`) and
`openbank-psd2-service` (`openbank.psd2.events`) are missing from all three places too, and are
named in neither #5859 nor #6035. Two careful manual enumerations of the same set were each two
short — which is the argument for deriving the scope rather than probing it by hand. Both are
baselined here rather than fixed: adding a topic to the audit subscription is a live-broker change
(a new Read grant on a running consumer group) whose blast radius wants its own review.

### Negative controls

A gate is proven by what it PREVENTS, so each was run against a state it must reject, then
restored. All six failed as required and the tree returned to clean (exit 0).

| control | result |
|---|---|
| `openbank.lending.events` removed from the `topics:` list only | exit 1, `::error::openbank-lending-service produces openbank.lending.events but it is missing from audit-service's incoming topics list` |
| its `TopicAttribution` entry removed only | exit 1, `…missing from TopicAttribution in EventAttribution.kt` |
| its Read ACL removed only | exit 1, `…missing from the audit-service KafkaUser's Read ACLs` |
| a baselined gap silently fixed (delegation added to the list) | exit 1, `::error::stale KNOWN_GAPS entry openbank-delegation-service#openbank.delegation.events#topics` |
| a money-path service starts producing a NEW topic | exit 1, 6 findings — all three places for the new topic, plus the stale entries for the renamed one |
| `OUT_OF_SCOPE` topic no longer produced by any money-path service | exit 1, `::error::stale OUT_OF_SCOPE entry openbank.notification.requests` |
| `IN_FLIGHT`: #5857 simulated as merged (all three places populated) | exit 0, `::warning::IN_FLIGHT entry openbank.delegation.events is no longer a gap … Delete the entry` |
| restored | exit 0, `21 produced topic(s) … 21 known gap(s) — clean.` |

The `--selftest` runs 14 cases on every CI invocation, including a negative control on a real
fully-covered topic in all three places, so the flagging branches stay falsified once the fleet is
clean.

Registered in `.github/gates/gates.yaml` as `audit-money-path-subscription`, `group: data`,
`mode: enforced`, with `min_subjects: 15` (21 today) so the gate cannot become a green no-op if the
glob stops matching, plus `rationale` + `review_after`. Every meta-gate passes.

## 2. The runtime test

`AuditSubscriptionSurfaceIT` boots the app, switches `audit-events-in` to the in-memory connector,
and pushes one record per subscribed topic **through the channel**, each carrying a genuine
`IncomingKafkaRecordMetadata` and `ce-type` header, then asserts the persisted row with a plain
JDBC read.

**What it proves that a direct call cannot.** Every existing test in this module reaches the
consumer by injecting the bean and calling `consume(message)`. A direct call supplies the very
wiring it is meant to test: it proves the method body works and is structurally incapable of
distinguishing a REGISTERED channel from an unregistered one. That is the #3371 shape, where every
`POST /mcp` answered 404 on a running pod while a unit test calling the resource class stayed
green. `InMemoryConnector.switchIncomingChannelsToInMemory` can only produce a source for a channel
SmallRye actually registered, so a renamed channel or a method SmallRye refuses to bind fails here
at boot rather than silently.

It also asserts every row is attributed to a named service and never to `"unknown"` — which matters
more here than anywhere, because `audit_entries` is append-only at the DB (`no_update_audit` /
`no_delete_audit` are `DO INSTEAD NOTHING`, so an UPDATE affects zero rows and *reports success*)
and `source_service` is chain-hashed into `record_hash`. An attribution gets one chance to be right.

**What it does not prove, stated plainly.** It is **not a real broker**. There is no Kafka
Testcontainers or Kafka Dev Services usage anywhere in this repo — every messaging IT in the fleet
uses the in-memory connector — so no test here exercises a Kafka client, a broker, or an ACL. The
in-memory connector replaces the Kafka connector for the channel, so the `topics:` value is read as
configuration and is not what the connector subscribes with. It therefore cannot see the two
failures that need a broker: a topic absent from the deployed `topics:` list, and a missing Read
ACL. That is exactly what the gate covers, and why the two halves are in one PR.

**Negative control, measured.** Renaming the channel on `AuditConsumer` to `audit-events-in-RENAMED`
— leaving everything else untouched — makes the test go red, and SmallRye says exactly why:

```
SRMSG00207: Some components are not connected to either downstream consumers or upstream producers:
  - IncomingConnector{channel:'audit-events-in', ...} has no downstream
  - SubscriberMethod{method:'...AuditConsumer#consume', incoming:'audit-events-in-RENAMED'} has no upstream
```

and the assertion names every affected topic:

```
java.lang.AssertionError: [subscription surface]
Expecting empty but was: ["openbank.accounts.account.created: no audit_entries row — the channel
    took the record and nothing was persisted", ... 21 topics ...]
```

`tests="1" skipped="0" failures="1"` in the JUnit XML, BUILD FAILED in 2m31s. Restored, it is
`tests="1" skipped="0" failures="0"`, BUILD SUCCESSFUL in 1m37s — the skipped count read explicitly,
since a boot failure reports tests as SKIPPED and that reads as a pass. This is the defect a direct
call to `consume()` cannot see at all.

The first red run also found a defect in the test itself, which is the reason to run the negative
case rather than reason about it: a per-topic 30-second wait multiplies the failure path by the topic
count, so the red run took 13 minutes and died of a socket timeout before it could say which topic
was missing. The wait is now one deadline and one batched query for the whole surface.

Traps handled: `runOnVertxContext(true)` (the consumer is a `suspend fun` over reactive Panache and
throws `HR000068` off a Vert.x context), the JDBC read rather than a reactive-repo call from the
test thread, literals only in the `QuarkusTestResourceLifecycleManager` (it loads in a different
classloader, so a companion object initialises twice), and a corpus assertion on the topic list so
the per-topic loop cannot pass vacuously over an empty list.

## Not done here

- The eight gaps themselves stay baselined. Each needs a Read grant on a live broker plus a
  `TopicAttribution` value verified against the module that declares the outgoing channel, and
  `auto.offset.reset` must NOT be forced to `earliest` on this shared group — that is a mass replay
  across ~21 money-path channels. Backfill is a separate deliberate exercise.
- billing-service produces `openbank.billing.fee.event` with no `KafkaTopic` CR at all; noted in
  #6035, not addressed here.

`test(audit)` / `ci(...)` — neither type releases, which is correct: no shipped service behaviour
changes.

Refs #6035, #5859, #5857, #5338, #3371, #3883, ADR-0249 D4.

### DIFF
--- a/.github/gates/gates.yaml
+++ b/.github/gates/gates.yaml
@@ -2612,6 +2612,27 @@ gates:
     selftest_expect: pass
     run: |
       python3 .github/scripts/check-alert-metric-emitted.py --enforce
+  # An unaudited money-path topic is indistinguishable from a quiet period: nothing errors, no pod
+  # is unhealthy, lag reads zero, and audit_entries simply has no rows. #5859 found delegation had
+  # never been subscribed; #6035 found five more, all missing from all three places at once — the
+  # signature of a set that was never enumerated. This derives the scope instead: money_path_services
+  # from rules.yaml, their produced topics from their own `mp.messaging.outgoing.*.topic` (never from
+  # a topic-literal grep, which is blind to the serialised-data-class idiom, #3883), and the three
+  # subscription artefacts from audit-service's config, source and KafkaUser.
+  # Ratchet, not zero: today's eight gaps are baselined with their owning issue and the list may only
+  # shrink — a stale entry, in either direction, is itself an error.
+  - id: audit-money-path-subscription
+    min_subjects: 15   # 21 produced money-path topics today (2026-08-20)
+    name: "Every money-path produced topic reaches the audit trail (issue #6035, enforced)"
+    group: data
+    mode: enforced
+    rationale: "A money-path event topic audit-service is not subscribed to, not attributed for, or has no Read ACL on writes nothing to the tamper-evident trail and raises no error anywhere — and audit_entries is append-only with source_service chain-hashed into record_hash, so neither the missing row nor a wrong attribution can be repaired afterwards."
+    review_after: "2027-02-20"
+    selftest: |
+      python3 .github/scripts/check-audit-money-path-subscription.py --root . --selftest
+    selftest_expect: pass
+    run: |
+      python3 .github/scripts/check-audit-money-path-subscription.py --root . --enforce
 
   - id: kafka-acl-coverage
     name: "Kafka ACL coverage (issue #2598, enforced)"
--- a/.github/scripts/check-audit-money-path-subscription.py
+++ b/.github/scripts/check-audit-money-path-subscription.py
@@ -1,0 +1,442 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) OpenBank contributors. Licensed under the Apache License, Version 2.0.
+"""Every topic a money-path service PRODUCES must reach the audit trail — in all three places.
+
+Why this exists
+---------------
+#5859 found that audit-service had never subscribed to `openbank.delegation.events`, so no
+delegation event of any kind had ever reached the audit trail. #6035 then applied the same probe
+across `money_path_services` and found five more — ledger, sdd, interest, fraud and billing — each
+missing from all three places at once. That is the signature of a set that was never enumerated,
+not six independent oversights, and it is exactly the failure a hand-kept list produces: it reads
+as complete because it is short.
+
+A subscription needs three things, and each one alone is silent when absent:
+
+  1. the topic in audit-service's `mp.messaging.incoming.audit-events-in.topics` list
+     — absent: nothing is read, no error anywhere;
+  2. an entry in `TopicAttribution` (`EventAttribution.kt`)
+     — absent: the row is written with `source_service = "unknown"`. `audit_entries` is
+       append-only at the DB (`no_update_audit`/`no_delete_audit` are `DO INSTEAD NOTHING`, so an
+       UPDATE affects zero rows and REPORTS SUCCESS) and `source_service` is chain-hashed into
+       `record_hash` — so a wrong attribution cannot be corrected in place, ever;
+  3. a Read ACL on the `audit-service` KafkaUser
+     — absent: the subscription is ACL-denied at runtime (the broker runs
+       `allow.everyone.if.no.acl.found=false`), as a retry loop rather than a red pod (#5338/#3271).
+
+An unaudited money-path topic is indistinguishable from a quiet period. Nothing errors, no pod is
+unhealthy, lag reads zero, and the audit trail simply has no rows.
+
+HOW THE PRODUCED SET IS DERIVED (and why not by grepping topic names)
+---------------------------------------------------------------------
+Producers are resolved from `mp.messaging.outgoing.<channel>.topic` in each money-path service's
+own `application.yaml` — the connector's own declaration. A literal grep for topic strings cannot
+do this job: this repo builds event payloads two ways, and a serialised data class has no literal
+anywhere in the tree (the key exists only at runtime as a Kotlin property name), so a grep returns
+a confident false clean over exactly the producers it structurally cannot see (#3883).
+
+A money-path service that declares no outgoing Kafka channel at all — settlement-service and
+vop-service today — therefore contributes nothing and is reported as `no outgoing Kafka channel`,
+which is correctly ABSENT rather than missing. That distinction is derived from the config, not
+declared here.
+
+SCOPE IS DERIVED, NOT LISTED
+----------------------------
+The set of services checked comes from `rules.yaml: money_path_services`; the set of topics from
+their own channel config; the three subscription artefacts from audit-service's config, source and
+KafkaUser. Nothing about the coverage set is maintained by hand. The only hand-kept list is
+`KNOWN_GAPS` below — the debt that exists TODAY — and it may only SHRINK: an entry that is no
+longer a gap is itself reported as an error, in enforced mode, so a temporary exception cannot
+quietly become permanent (the idiom of `check-kafka-acl-coverage.py`'s KNOWN_GAPS and the pact
+gate's KNOWN_UNCOVERED).
+
+Usage:  check-audit-money-path-subscription.py [--root .] [--enforce] [--selftest]
+"""
+
+from __future__ import annotations
+
+import argparse
+import pathlib
+import re
+import sys
+
+sys.path.insert(0, str(pathlib.Path(__file__).resolve().parent))
+
+import yaml  # noqa: E402
+
+import gatelib  # noqa: E402
+
+AUDIT_SERVICE = "openbank-audit-service"
+AUDIT_CHANNEL = "audit-events-in"
+AUDIT_KAFKAUSER = "audit-service"
+
+# Gaps that exist TODAY, each keyed `<service>#<topic>#<place>` and carrying the issue that owns
+# it. Fixing one means DELETING its entries here — and the check fails on an entry that is no
+# longer a gap, so the list cannot drift in either direction.
+#
+# What is left after #6035's first backfill. Four of the original seven -- ledger
+# (openbank.ledger.journal.posted), sdd, interest and fraud -- were wired in all three places and
+# their entries deleted here; a stale entry is itself an error, so a half-fix cannot pass.
+#
+# The three that remain each need a decision this ratchet should not make silently:
+#  - billing has no KafkaTopic CR for openbank.billing.fee.event at all, so a Read grant would
+#    point at an undeclared topic -- a second way to get silence, and the topic's ownership is
+#    billing's call, not audit's.
+#  - standing-order and psd2 were found by THIS check and named in neither #5859 nor #6035, so
+#    nothing has yet reviewed whether their streams belong in the audit trail.
+_GAP_ISSUE = {
+    "openbank-billing-service": "#6035 - also has no KafkaTopic CR (see the issue)",
+    # Found by THIS check on its first run, and by nothing before it: neither #5859's probe nor
+    # #6035's enumeration named these two. That is the argument for deriving the scope rather
+    # than probing it by hand - a careful manual enumeration of the same set, done twice, was
+    # still two short.
+    "openbank-standing-order-service": "#6035 - found by this check, not named in the issue",
+    "openbank-psd2-service": "#6035 - found by this check, not named in the issue",
+}
+_GAP_TOPICS = {
+    "openbank-billing-service": "openbank.billing.fee.event",
+    "openbank-standing-order-service": "openbank.standing-orders.order.event",
+    "openbank-psd2-service": "openbank.psd2.events",
+}
+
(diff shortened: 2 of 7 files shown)


JiRaska:

### Verdict: the gate is proven by what it prevents. Six negative controls, all correct.

I did not trust the claim; I ran the gate in a throwaway worktree and reintroduced defects. `--enforce` (the mode `gates.yaml:2398` actually invokes), exit code read directly from `$?`, worktree reverted and confirmed clean between each:

| # | mutation | exit | `::error` count |
|---|---|---|---|
| 0 | unmodified branch | **0** | 0 |
| 1 | ledger produces a new, unbaselined topic | **1** | 6 |
| 2 | fix the sdd gap in `topics:` **only** (half-fix ⇒ stale entry) | **1** | 1 |
| 3 | drop `openbank.accounts.account.created` from `topics:` | **1** | 1 |
| 4 | corrupt one ACL grant name only | **1** | 1 |
| 5 | collapse `money_path_services` to one entry | **1** | 3 + `SUBJECTS=21→10` |
| 6 | `--selftest` | 0 | 14 cases, both directions |

Case 2 is the one I most wanted to see fail, and it does: a half-fix that wires one of the three places cannot pass, because the surviving baseline entry becomes stale and staleness is an error. Case 4 confirms all three places are genuinely independent subjects, not one check reported thrice. Case 5 is caught **twice over** — the stale-entry errors fire *and* `SUBJECTS` drops below `min_subjects: 15`, so `gate-subject-floor` catches it independently. That is the right belt-and-braces for a derived-scope gate.

**My first run was wrong and I want to flag why**, because it is the failure mode this repo documents: invoked without `--enforce` the script prints `::warning::` and exits 0, and I briefly read that as "a gate that cannot fail". It is the advisory mode, `gates.yaml` passes `--enforce`, and the exit logic at `check-audit-money-path-subscription.py:439` (`return 1 if findings and args.enforce else 0`) is correct. Noting it only so the next reviewer does not repeat it.

### The scope derivation is the right shape

`money_path_services` from `rules.yaml`, produced topics from each service's own `mp.messaging.outgoing.*.topic` declaration, subscription surface from audit-service's three artefacts. Crucially it does **not** grep for topic literals — which, per #3883, is structurally blind to the serialised-data-class idiom. `derive` + `min_subjects` + bidirectional staleness is exactly what `pact-drift-check` lacked. Two services (`settlement`, `vop`) are reported as *declaring no outgoing channel* rather than silently omitted, which is the difference between a scope and a hole.

### The IT is honest about what it cannot do

`AuditSubscriptionSurfaceIT.kt:52-58` pushes through `InMemoryConnector.switchIncomingChannelsToInMemory`, which can only produce a source for a **registered** channel bound to a live CDI consumer — so it discriminates a served channel from an unserved one, which a direct method call cannot (the `@Path`-bound-to-the-wrong-declaration lesson). The row assertion is plain JDBC because the reactive repo has no Vert.x context on a bare `@QuarkusTest` thread. `:125-127` asserts the corpus is non-trivial first, so an empty `topics:` cannot make the whole test vacuously green. And `:79-90` states the three limitations plainly (not a real broker; does not assert ACLs; does not assert the *produced* set). I checked each of those claims against the code and they hold.

### MINOR — the summary line is confusable

```
21 produced topic(s) across 23 money-path service(s) (2 produce no Kafka events), 21 known gap(s) — clean.
```

Both numbers are 21 and they count different things: 21 produced **topics**, and 21 baseline **entries** (7 topics × 3 places). It reads as "all 21 topics are gaps", which is not what it means. `7 topic(s) / 21 entr(ies)` would remove the coincidence. Same in `gates.yaml:2386`, which says "today's eight gaps" — right if you count the in-flight delegation row, but not reconcilable with either 21 without doing the arithmetic yourself.

### MINOR — what the gate deliberately does not answer, worth stating in the rationale

It proves a topic is **subscribed, attributed and ACL-granted**. It does not prove a row is **written**. `AuditConsumer.consume` catches `Exception`, logs, and acks in a `finally` — so a store failure loses the event silently with the offset committed, and there is no counter on that catch. Green gate, zero rows, no series to alert on. That is not this PR's defect (see my comment on #6062), but the gate's `rationale:` currently reads as though it closes the whole path.

### Checked, found nothing
- `min_subjects: 15` against 21 today — real headroom, and it does floor on the number that would move.
- `--root` handling: self-test fixtures do not inherit the real tree.
- `IN_FLIGHT` / `OUT_OF_SCOPE` are separately staleness-checked, not a second unguarded escape hatch.
- No scope creep: four files, all on-topic. No `version.txt` / `openapi.yaml` / released path touched; `test(audit)` correctly releases nothing.

JiRaska:
## Rebuilt on live `main` — 71 files down to 7, and the dropped 64 were not this PR's

New head: `8b06ad4ba`, signed (`%G?` = `G`), rebased onto current `main`.

### What the conflict actually was

The branch had no merge commits — five linear commits — but the last one, `bf002b640`, touched **69 files**. Four of the five commits are the real work (2 files, 1 file, 2 files, 1 file). The fifth had absorbed a large amount of other people's work. Classified all 69 by comparing each file against `origin/main`:

| Group | Files | Verdict |
|---|---|---|
| `openbank-infra/gitops/components/*/…yaml` | 30 | **Auto-deploy image-tag bumps.** Not OPA bundles — every diff is a single `image: …:sandbox-<sha>` line. Applying them would have rolled 30 services' deployed images back to a stale tag. Dropped; no regeneration needed, since no `.rego` and no `rules.yaml` read-set key is touched, so no bundle or `policy-checksum` restamps. |
| `openbank-product-catalog` (4), `openbank-admin-ui` (6), `pacts/` (1) | 11 | Already **byte-identical to `main`** — landed separately. Dropped. |
| `openbank-finrep-service` | 16 | `main` is **ahead**. `LedgerClient.kt`: `merge-base→main` is +32/−10, `merge-base→branch` is +2/−1. Taking the branch's side would have reverted the newer work. Dropped. |
| `openbank-delegation-service` + `openbank-contracts` | 8 | Genuinely unlanded (+86 in `DelegationEvents.kt`, +115 in `asyncapi.yaml`) — but it is the payload of **open PR #5857**, `feat(delegation): emit audit events for spend reserve, confirm and release`. Dropped; #5857 owns it. |

Kept 7 files: the gate (`gates.yaml` + `check-audit-money-path-subscription.py`), the audit-service subscription/attribution/build changes, the IT, and the KafkaUser Read ACLs.

`gates.yaml` was the only real conflict: `main` had added three gate entries (`dashboard-metric-emitted`, `llm-price-book`, `alert-metric-emitted`) in the same place. Kept **both** sides — the file now parses to 172 gates with exactly one added by this PR.

Note the delegation *subscription* stays here even though the delegation *producer* went to #5857: subscribing the audit trail to `openbank.delegation.events` is audit-service scope, and the two are complementary rather than duplicate — #5857's events would otherwise be an outbox row nobody reads.

### One thing the gate caught about itself

On the first clean run it reported:

```
::warning::IN_FLIGHT entry openbank.delegation.events is no longer a gap - PR #5857 fixes all
three places (#5859) has landed. Delete the entry so the ratchet covers this topic again.
```

Since this PR now covers that topic in all three places directly, the `IN_FLIGHT` entry had become a stale declaration — which this check treats as an error in its own right, and correctly so: a baseline entry for a gap that no longer exists silently removes the topic from the ratchet. Emptied `IN_FLIGHT` and left a comment saying why.

### Gate output, run locally on the rebuilt tree

```
$ python3 .github/scripts/check-audit-money-path-subscription.py --root . --selftest
selftest OK: 14 cases, both directions (negative control on 'openbank.accounts.account.created'
in all three places; direction-aware channel parse).
exit 0

$ python3 .github/scripts/check-audit-money-path-subscription.py --root . --enforce
SUBJECTS=21  # money-path produced topics
check-audit-money-path-subscription: 21 produced topic(s) across 23 money-path service(s)
(2 produce no Kafka events), 9 known gap(s) — clean.
exit 0        # no warnings remaining

$ ./gradlew :openbank-audit-service:ktlintCheck :openbank-audit-service:detekt :openbank-audit-service:test
GRADLE_EXIT=0
```

Test results parsed from the JUnit XML rather than the console: **152 tests, 0 failures, 0 skipped** — the skipped count checked deliberately, since a service that cannot start reports its integration tests as skipped and still reads as a pass.

`git diff --name-only origin/main` returns exactly the 7 files above.

PR #6048, opened 2026-08-20
fix(settlement): actually reverse money in the compensation path (#6037)

### LINKED ISSUE #6037: settlement compensation does not reverse money: reverseDebit/reverseCredit are stubs that report success
## What

`openbank-settlement-service`'s compensation path does not reverse money. `reverseDebit` and
`reverseCredit` write a status row and log `"stub: wire reversal to balance-service"`. No call
reaches balance-service; no ledger entry is produced.

Found while instrumenting settlement for #5705 (PR #6036), which is why it is filed separately —
it predates that PR and is out of its scope.

## Why this is worse than an ordinary missing implementation

The saga *reports success*. A settlement that fails mid-way transitions through
`LEDGER_REVERSED` / `*_REVERSED` states, and those states are reached and recorded — so from
every angle an operator normally checks, the unwind looks like it completed. The money did not
move back.

This is the repo's recurring shape: an outcome that is structurally unable to report the thing
that went wrong. Same family as `PushResult.skipped()` carrying `success = true`.

## What does NOT catch it today

- No metric distinguishes a real reversal from the stub (settlement emitted no metrics at all
  before PR #6036).
- The new `SettlementStuckAfterCompensation` alert in #6036 makes a *stalled* unwind visible —
  it cannot see an unwind that "completed" without moving money.
- Unit tests that mock the port cannot see it.

## Suggested shape of the fix

1. Wire the two methods to balance-service (or make them fail loudly until wired — a stub that
   returns success on a money path should not be reachable in a deployed environment).
2. Give the outcome its own value rather than sharing the success path: `REVERSAL_SKIPPED` is a
   different fact from `REVERSED`.
3. Prove it with a real-DB / real-HTTP integration test asserting the counter-entry exists —
   a mocked repository cannot establish this.

## Verification note

This is a static reading of `src/main` on `origin/main`. Whether any live settlement has actually
taken the compensation path (and therefore whether money is currently unreturned) has NOT been
established — a read-only query over settlement rows in a `*_REVERSED` state would answer it and
is worth running before deciding urgency.

Refs #5705, #6036

JiRaska:
## What

`openbank-settlement-service`'s compensation path did not reverse money. `reverseDebit` and
`reverseCredit` wrote a status row, logged `"stub: wire reversal to balance-service"`, and
reported success. `reverseBookToLedger` was the same stub against ledger-service.

The saga reported success, so the status column, the audit event and the Temporal history all
agreed the unwind had completed. Same shape as `PushResult.skipped()` carrying `success = true`:
an outcome structurally unable to report the thing that went wrong.

Closes #6037. Refs #5705, #6036.

## Is the path reachable, and is money currently unreturned?

**Reachable, but it has never run — no money is currently unreturned.** Measured read-only against
the live sandbox primary (`settlement-service-db-2`, database `openbank_settlement`):

```
select count(*) from settlements;                          ->  0
select status, count(*) from settlements group by status;   ->  (0 rows)
select relname, n_tup_ins, n_tup_upd, n_tup_del, n_live_tup
  from pg_stat_user_tables where relname='settlements';
  -> settlements | 0 | 0 | 0 | 0
select stats_reset from pg_stat_database
  where datname='openbank_settlement';                      ->  NULL (never reset)
```

No row has ever been inserted, so nothing has ever been in a `*_REVERSED` state. This was a
**pre-deployment defect, not an incident** — which is why this PR fixes it properly rather than
racing a hotfix. The caveat: this is the sandbox, and PG statistics survive a clean shutdown but
reset on a crash; the empty table plus `n_tup_del = 0` is the stronger of the two signals.

Reachability itself is real: `SettlementWorkflowImpl` registers each compensation after its
forward activity succeeds, so a `creditPayee` failure would have left the payer debited and never
refunded, and a `bookToLedger` failure would have left both balances moved.

## The correct counterparty call — not invented

balance-service has **no reversal endpoint**. Its complete path list is `{accountId}`,
`{accountId}/{currency}`, `overdraft-limit`, `holds`, `holds/{holdId}`, `credit`, `debit`,
`initialize`, `reconciliation`, `reconciliation/latest`, `approvals/{id}`. Its only undo
primitives are releasing a hold and issuing the opposite movement. Settlement moves money with
`debit`/`credit`, not holds, so **the opposite movement is the reversal**:

| Compensation | Call | Reference id |
|---|---|---|
| `reverseDebit` | `POST /api/v1/balances/{payerAccountId}/credit` | `settlement-debit-reversal-<id>` |
| `reverseCredit` | `POST /api/v1/balances/{payeeAccountId}/debit` | `settlement-credit-reversal-<id>` |

Same `BalanceRestClient`, same `MoneyMovementRequest`, same account resolution as the forward
adapters — verb and reference-id namespace flipped. settlement-service is the fleet's only caller
of these two endpoints, so there was no existing reversal idiom to mirror; that absence is itself
part of the finding.

## Idempotency

balance-service deduplicates durably on the `balance_movement` primary key
`(account_id, currency, reference_id, operation)`, applying mutation and marker in one
transaction and answering a duplicate 200 with `applied = false`. There is no `Idempotency-Key`
header on that API and balance-service does not use the shared `openbank-libs` `IdempotencyStore`
(that store is a REST response cache keyed on a caller header — the wrong tool here). **The
reference id is the idempotency key**, and it is a pure function of the settlement id, so a
retried Temporal activity re-sends an identical key.

The reversal ids are deliberately in a different namespace from the forward ids rather than
reusing them: sharing the forward id would work today but would rest the whole guarantee on
`operation` alone being part of the counterparty's primary key.

## What each outcome now says

Per the "own enum value, never a shared success flag" rule:

- `REVERSED` / `CREDITED_REVERSED` — balance-service accepted the counter-movement.
- `REVERSAL_FAILED` — **new.** Attempted and refused; the money is still moved. balance-service
  enforces `booked - amount >= overdraftFloor` on every debit, so a payee who has spent the funds
  makes `reverseCredit` fail with 422 and no retry resolves it. Recovering those funds is a
  collections/dispute process, not an API call.
- `LEDGER_REVERSAL_UNSUPPORTED` — **new.** The GL posting was not reversed.
- `LEDGER_REVERSED` — deprecated, never written again.

The status is written only **after** the money call returns; the pre-fix code wrote it first,
which is the whole defect in one line.

## Which half did I ship, loud-failure or real wiring? Both, deliberately split

**Real wiring** for `reverseDebit` and `reverseCredit` — the issue's actual scope, and an
unambiguous mirror of the forward path.

**Loud failure** for `reverseBookToLedger`, which the issue does not name but which is the same
defect. ledger-service *does* expose `POST /api/v1/journals/{journalId}/reverse`, and I did not
wire it because three separate decisions block it, none of them mine to make:

1. **Maker-checker.** `ledger.reverse` is approval-gated — a service-account call queues rather
   than posts. Granting a machine that action is a `rules.yaml: shared_m2m_matrix_write_grants`
   decision, and an automatic GL reversal driven by a failed saga is close to the thing
   maker-checker exists to prevent.
2. **No journal id is retained.** `bookToLedger` discards the response, so the id needs persisting
   (a migration) or re-resolving via `GET /api/v1/journals/transaction/{transactionId}`.
3. **Period locks.** The endpoint answers 409 for an ATTESTED fiscal period, and the accounting
   convention there is to correct forward in the open period — a different posting, not a reversal.

So it throws a **non-retryable** `ApplicationFailure` and records `LEDGER_REVERSAL_UNSUPPORTED`.
Non-retryable matters: a retryable failure would burn five attempts (~75 s of backoff) ahead of
the two compensations that *do* return money, and end in the same place. `SettlementWorkflowImpl`
catches `ActivityFailure` per compensation, so this does not block them.

`SettlementCompensationCapabilities` announces the gap **at boot** via `@Startup` — load-bearing,
because `@ApplicationScoped` is lazy and an `init {}` gate on a rare path is announced too late to
act on (the `PdfBoxPadesSealAdapter` warning that never once appeared in a pod log, #1299).

## Proof, and the negative control

`SettlementReversalIT` mocks nothing on the path it checks: the real CDI activity bean over its
real `VertxContextSupport` bridge, the real adapters and `BalanceRestClient`, a real Testcontainers
Postgres with the real migrations, and **plain JDBC** for both seed and assertion, so the reactive
layer never confirms its own work. A `BalanceServiceWireMockResource` stands in for balance-service
over real HTTP, so the outbound request is asserted by URL, verb and body.

**Negative control — restored the three stub bodies, changed nothing else, re-ran only this IT:**

```
tests=5 failures=5 errors=0 skipped=0
```

All five red, with the failure naming the defect directly:

```
reverseDebit issues a real counter-credit to the payer and records REVERSED
  VerificationException: Expected at least one request matching:
    { "url": "/api/v1/balances/<payer>/credit", "method": "POST",
      "bodyPatterns": [ { "matchesJsonPath":
        "$[?(@.referenceId == 'settlement-debit-reversal-<id>')]" } ] }

a retried reversal re-sends an identical reference id rather than a fresh one
  AssertionError: Expected size: 2 but was: 0 in: []

a refused counter-debit records REVERSAL_FAILED ...   AssertionError: Expecting code to raise a throwable.
reverseBookToLedger fails loudly ...                  AssertionError: Expecting code to raise a throwable.
```

Restored the production file and it returns to 5/5 green. The test can detect the defect's absence.

**What it does not establish:** that balance-service *applied* the movement. settlement-service
cannot observe another service's ledger, and claiming otherwise would be the `accepted`-vs-
`delivered` mistake. The applying half is balance-service's own `balance_movement` dedup and
overdraft tests.

Three unit tests that asserted the defect verbatim (`reverseDebit does not call debit port and
sets REVERSED status`) are replaced — they passed *because* no money moved.

## Threat model

`docs/threat-models/openbank-settlement-service.md` did not merely miss this; it documented it as
mitigated.

- **R2** — "Compensation claimed to have run when it did not" — listed as its mitigation
  *"compensation activities update status to REVERSED/CREDITED_REVERSED/LEDGER_REVERSED
  atomically"*. That mitigation **was** the defect: the status update was the only thing happening.
- **T1** — claimed `referenceId = workflowRunId + activityId`. That is not what the code sends,
  and would not have worked if it were: it changes per run, so a retry would not dedupe.
- **Residual risk 3** credited the Temporal migration with closing a compensation gap. The ordering
  was right; the reversals were no-ops, so it unwound exactly as much money as the legacy path it
  replaced — none.

All three corrected, plus new residual risks 5-7 (the ledger gap, the legitimately-refused
reversal, and a pre-existing one this PR does **not** fix: a debit applied by balance-service but
never observed by settlement-service registers no compensation at all).

## Verification

Gradle serialized, output to files, exit code read directly (never through a pipe), `--rerun-tasks`:

| Check | Result |
|---|---|
| `:openbank-settlement-service:test` | **48 tests, 0 failures, 0 errors, 0 SKIPPED** (read from JUnit XML) |
| `ktlintCheck`, `detekt`, `koverVerify` | pass |
| `quarkusBuild` | pass — validates the CDI wiring for two new beans and a changed constructor |
| `run-gates.py --all` | 158 gates, 151 pass |
| `threat-model-updated-on-trust-boundary-change` | **PASS** |
| `db-migration-gate`, `schema-compat-gate` | PASS (no migration — `status VARCHAR(32)` already fits the longest new value at 27 chars) |
| `check-api-contract.py --enforce` | **`additive change, info.version 1.1.1 -> 1.2.0 — OK (required >= MINOR)`** |

`info.version` was re-read off live `origin/main` immediately before bumping, not off the branch
base. `version.txt` is untouched — release-please owns it, and `fix(settlement)` over a path
outside `exclude-paths` releases.

The SKIPPED count is called out because an earlier run reported `48 tests / 16 skipped / 1 failure`
— a `QuarkusBindException` from a parallel agent's JVM holding port 8081, not a code failure. I
waited for the port and re-ran rather than reading a boot failure as a pass.

**Not verified:** no end-to-end run against a live balance-service (the sandbox has never had a
settlement row); the two remaining `run-gates.py` failures (`litellm-config-revision`,
`loki-rule-load-test`) touch files this PR does not, and I did not baseline them against `main`.

## Coordination

Disjoint from the two settlement PRs in flight, checked by 3-dot diff:

- **#6036** (`feat/settlement-metrics-5705`) — overlaps only on `build.gradle.kts`.
- **#5723** (`feat/settlement-service-metrics`) — overlaps on `build.gradle.kts`,
  `SettlementActivitiesImpl.kt` and `SettlementActivitiesImplTest.kt`. Unavoidable: it wraps every
  activity in a `step(...)` metrics helper, and this PR changes the bodies of three of them.
  The resolution is mechanical — its `step(SettlementStep.REVERSE_DEBIT) { ... }` wrapper around
  this PR's `compensate(...)` body — and whichever lands second should do it deliberately rather
  than take one side. Its `SettlementStepOutcome` is also worth extending: a compensation that
  wrote `REVERSAL_FAILED` is not `COMPLETED`.

## Review

Money-path: **2 approvals + threat model** per `rules.yaml`. I have not merged this and will not.

### DIFF
--- a/docs/threat-models/openbank-settlement-service.md
+++ b/docs/threat-models/openbank-settlement-service.md
@@ -71,7 +71,7 @@ replacing the former in-memory stub), so settlement state is durable across rest
 
 | ID | Threat | Mitigation |
 |----|--------|------------|
-| T1 | Replayed debit activity credits twice | `referenceId = workflowRunId + activityId` stored before side-effect; idempotency guard in balance-service |
+| T1 | Replayed debit or reversal activity moves money twice | Corrected 2026-08-20 (#6037). The reference id is **not** `workflowRunId + activityId` as this row previously claimed — that would change on every workflow run and defeat the guard. It is a pure function of the settlement id (`settlement-debit-<id>`, `settlement-credit-<id>`, and for compensation `settlement-debit-reversal-<id>` / `settlement-credit-reversal-<id>`), so every Temporal retry re-sends an identical key. balance-service deduplicates durably on the `balance_movement` primary key `(account_id, currency, reference_id, operation)`, applying mutation and marker in one transaction and answering a duplicate 200 with `applied = false`. The reversal namespace is deliberately distinct from the forward one rather than relying on `operation` alone to separate the rows. |
 | T2 | Settlement DB record modified out-of-band (bypassing workflow) | Postgres row-level security; audit trail in `openbank-libs/audit` |
 | T3 | Temporal workflow history modified | Temporal's history is append-only; CNPG backups every 30 days |
 
@@ -80,7 +80,7 @@ replacing the former in-memory stub), so settlement state is durable across rest
 | ID | Threat | Mitigation |
 |----|--------|------------|
 | R1 | Settlement denied by payer ("I never authorised this") | Temporal workflow execution ID links to the initiating payment ID in `sepa-payment-service` / `domestic-payment-service`; full chain queryable from admin UI |
-| R2 | Compensation claimed to have run when it did not | Temporal activity history records every attempt; compensation activities update status to `REVERSED`/`CREDITED_REVERSED`/`LEDGER_REVERSED` atomically |
+| R2 | Compensation claimed to have run when it did not | **This threat was realised, not mitigated, from ADR-0101 P3 until #6037.** The cited mitigation — "compensation activities update status to `REVERSED`/`CREDITED_REVERSED`/`LEDGER_REVERSED` atomically" — *was* the defect: updating the status column was the **only** thing `reverseDebit`/`reverseCredit`/`reverseBookToLedger` did. They logged `"stub: wire reversal to balance-service"` and returned success, so the status row asserted an unwind that had moved no money. Now: the two balance reversals issue real counter-movements to balance-service before writing a status, a refused reversal is recorded as `REVERSAL_FAILED` (not as success), and the unimplemented ledger reversal throws a non-retryable `ApplicationFailure` and records `LEDGER_REVERSAL_UNSUPPORTED`. Verified by `SettlementReversalIT`, which asserts the outbound HTTP request over real HTTP and the resulting row over plain JDBC — a mocked port cannot tell a counterparty call from a no-op. |
 
 ### I — Information disclosure
 
@@ -138,12 +138,48 @@ replacing the former in-memory stub), so settlement state is durable across rest
    (`legacySettle`) is deleted. This closes a compensation gap: the legacy path flipped a mid-flight
    failure straight to `REJECTED` **without reversing an already-moved debit/credit**, whereas
    `SettlementWorkflow` runs `reverseBookToLedger → reverseCredit → reverseDebit` before rejecting.
+   **Correction (#6037):** that last sentence was true only of the ordering. Until #6037 all three
+   compensation activities were stubs, so the Temporal path unwound exactly as much money as the
+   legacy path it replaced — none — while recording that it had. The gap is closed for the two
+   balance movements as of #6037; the ledger half remains open, as risk 5 below.
    Flip pre-conditions are verified: Temporal server healthy, `openbank-settlement` namespace
    registered, `settlement_activity.rego` present (risk 2). Worker registration is gated on
    `openbank.settlement.worker.enabled` (default true; false in `%test`). **Cutover risk:** with no
    fallback path, a settlement cannot be processed if the Temporal worker is not registered or the
    frontend is unreachable — a sandbox canary must confirm the worker registers and a real settlement
    drives `PENDING → BOOKED` before merge, and the deterministic simulation (ADR-0100/0115) stays green.
+
+5. **A settlement that fails after `bookToLedger` leaves the GL entry standing (issue #6037).**
+   `reverseBookToLedger` is deliberately NOT implemented and fails loudly
+   (`LEDGER_REVERSAL_UNSUPPORTED` + a non-retryable `ApplicationFailure`) rather than reporting a
+   reversal it did not perform. ledger-service does expose `POST /api/v1/journals/{journalId}/reverse`,
+   but three things must be decided before settlement-service may call it: (a) `ledger.reverse` is
+   maker-checker gated, so a service-account call queues for approval instead of posting, and granting
+   a machine that action is a `rules.yaml: shared_m2m_matrix_write_grants` decision — an automatic GL
+   reversal driven by a failed saga is close to what maker-checker exists to prevent; (b) settlement
+   does not retain the journal id (`bookToLedger` discards the response), so it must be persisted or
+   re-resolved via `GET /api/v1/journals/transaction/{transactionId}`; (c) the endpoint answers 409
+   when the entry's fiscal period is ATTESTED, and the accounting convention there is to correct
+   forward in the open period, which is a different posting. **Operational consequence:** such a
+   settlement unwinds both balance movements and needs a manual correcting entry in the general
+   ledger. It is visible as a `LEDGER_REVERSAL_UNSUPPORTED` row and an ERROR log line, and is
+   announced at boot by `SettlementCompensationCapabilities`.
+
+6. **A reversal can be legitimately refused and the money stays moved (issue #6037).**
+   balance-service enforces `booked - amount >= overdraftFloor` on every debit, so if the payee has
+   already moved the credited funds out, `reverseCredit` is refused with 422 and no retry resolves
+   it. The row records `REVERSAL_FAILED` and the audit event carries `AuditResult.FAILURE`.
+   Recovering those funds is a collections/dispute process, not an API call — this is a real
+   property of the world, and the design records it rather than smoothing it into a success.
+
+7. **A debit applied by balance-service but not observed by settlement-service is not compensated
+   (pre-existing, not addressed by #6037).** `SettlementWorkflowImpl` registers the `reverseDebit`
+   compensation only *after* `debitPayer` returns, so a debit that balance-service applied while the
+   response was lost — or that succeeded before the subsequent status update failed — leaves money
+   moved with no compensation registered. The idempotent `referenceId` means a retry does not
+   double-debit, and the row is left non-terminal (visible to the `SettlementStuckAfterCompensation`
+   alert in #6036), but no automatic unwind occurs. Closing this needs the compensation registered
+   before the call, which changes the saga's shape.
 
 4. **2-approval gate** — money-path rule requires 2 approvals. This PR has 1 (automated review).
    Second approval from a human committer required before merge per rules.yaml.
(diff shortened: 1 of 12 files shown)


JiRaska:
## Review — substantively right; one real finding on status granularity

This is the correct shape for the defect. The old stubs wrote `REVERSED`/`CREDITED_REVERSED`/`LEDGER_REVERSED` and moved no money, so the only thing an operator could observe asserted an unwind that had not happened — the `PushResult.skipped()` shape, and the PR names it as such.

Checked and clean:

- **Own value for the no-op.** `REVERSAL_FAILED` and `LEDGER_REVERSAL_UNSUPPORTED` are distinct enum values in `Settlement.kt`, not a boolean shared with success. `LEDGER_REVERSED` is deprecated and never written rather than deleted, which keeps the value's meaning for a pinned consumer. This is the rule applied properly.
- **`reverseBookToLedger` fails loudly** instead of claiming a reversal, records the status *first* so the row says which half of the unwind did not happen, audits `AuditResult.FAILURE`, and throws **non-retryable** so it cannot burn retry backoff ahead of the two compensations that do work. The three reasons it cannot call `ledger.reverse` (maker-checker, no retained journal id, period locks) are each real and each genuinely outside this service's authority.
- **Ordering.** `compensate()` calls the money port first and writes the status only after it returns. The tests pin this rather than assert it in prose: `SettlementActivitiesImplTest.kt:126` and `:141` are literally *"calls the reversal port and only then sets …"*. Those tests would fail against the old stub, which called no port at all — a genuine negative control, as are `:154` (refused reversal → `REVERSAL_FAILED` + FAILURE audit + rethrow) and `:195` (`isNonRetryable` and type `LedgerReversalUnsupported`).
- **Idempotency.** Reversal reference ids are pure functions of the settlement id and live in a distinct namespace from the forward ids (`settlement-debit-reversal-<id>`), so an activity retry re-sends an identical key and balance-service dedupes on `(account_id, currency, reference_id, operation)`. The reasoning for not reusing the forward id — not wanting the reversal's idempotency to depend on `operation` alone, a detail of the counterparty's PK — is correct and worth having written down.
- **Attribution.** The audit events carry `actorId = "settlement-service"`, `actorType = "SERVICE"`, both server-derived, nothing read from a request body.
- **`@Startup` on `SettlementCompensationCapabilities`** is load-bearing and the KDoc knows why — `@ApplicationScoped` is lazy, and a capability warning on a rare path is the exact `PdfBoxPadesSealAdapter` failure. Choosing a log line over a boot-time `check()` is the right call; refusing to boot would take a working money path offline over an unimplemented compensation.
- No Panache `persist()`-on-assigned-`@Id` exposure: this PR adds no repository write path, only `updateStatus` calls into the existing repo.
- CI: 29 success, 0 failures; the skips are path-scoped CI and Dependabot automerge, not test skips.

### Finding — `REVERSAL_FAILED` is one value for two different legs, and the later leg overwrites the earlier

The PR's own argument is that "a compensation that was *attempted and refused*, and one that is *not implemented*, are different facts from a compensation that *worked*, and each now has its own value." That reasoning stops one step short.

Both `reverseDebit` and `reverseCredit` failures write the same `SettlementStatus.REVERSAL_FAILED` (`SettlementActivitiesImpl.kt`, the shared `compensate()` helper). Because the status column is overwritten per activity, a saga where the payer's debit reversal **succeeded** (`REVERSED`) and the payee's credit reversal then **failed** ends with the row reading `REVERSAL_FAILED` — and an operator reading that column cannot tell whether the payer got their money back. That is precisely the information the four-value split exists to preserve, and it is the case most likely to occur, since `reverseCredit` is the one the PR documents as legitimately refusable (payee already spent the funds, 422, no retry fixes it).

The audit stream does distinguish them via `operation`, so the fact is recoverable — but the status column is what the PR itself identifies as "the one thing an operator could observe". Suggest either `DEBIT_REVERSAL_FAILED` / `CREDIT_REVERSAL_FAILED`, or a status that encodes which legs completed.

### Minor

- `compensate()` catches `Throwable`, which in a coroutine also catches `CancellationException`; it does DB work and an audit publish before rethrowing. It does rethrow, so cancellation is not swallowed, but on a cancelled activity those two calls run on a cancelled scope. `Exception` would be the safer catch unless `Error` is deliberately in scope.
- If `updateStatus` fails *after* a successful money movement, the row keeps the pre-reversal status while the money has moved. Temporal retries the whole activity, the deterministic reference id makes balance-service answer `applied = false`, and the status write then succeeds — so this converges. It is only safe *because* of the deterministic reference id, which is worth stating in `compensate()`'s KDoc since a future change to the id scheme would silently break it.

PR #6053, opened 2026-08-20
fix(observability): gate the alert window against its subject's invocation period

JiRaska:
Refs #5736

## What was measured, and by which methods

The defect class from #6041: an alert whose rate window is no wider than its **subject's invocation period** can never mature its `for:`. It stays `pending` forever. Every signal a human checks says it is fine — loaded, `health=ok`, `promtool` clean, ArgoCD green.

Three enumerations, deliberately not one, because a single grep is a probe and not a census:

| method | what it counts | result |
|---|---|---|
| YAML walk of every `PrometheusRule` / rule doc under `openbank-infra/` | committed alert rules | **137** |
| `grep -c '\- alert:'` over the same tree | same | **137** — agrees |
| live Prometheus `/api/v1/rules?type=alert` | rules actually **loaded and evaluating** | 443 total; **121 of 121** repo PromQL rules present, 0 missing |

The third method is the one that matters: #6032 found the Loki ruler had never loaded a single rule, so "the rule is in the repo" proves nothing. Every gitops PromQL rule in this repo **is** loaded. (The 16 not in the 121 are the 7 Loki LogQL rules — covered by `check-loki-rules.py` — and 9 local-dev rules in `docker/prometheus/`.)

Subject cadences were derived from **three** sources and cross-checked against a second, independent pass: `@Scheduled(cron=/every=)` in Kotlin with placeholders resolved through `application.yaml` and then gitops env overrides, `kind: CronJob` schedules, and `agents.yaml` charter schedules. **142 resolved, 0 unresolved.** Both passes independently resolved `agent.oversight.cron` to 30m, which is also exactly the gap the live `ALERTS` history shows.

## The split — not a total

A rule that has never fired because nothing bad happened is **healthy**. Separating that from a rule that *cannot* fire is the whole task. The discriminator is the shape of the `pending` history, not its size: episodes whose **length is capped at the rate window** and whose **start-gaps equal the subject's period** are a rule that cannot mature. Irregular gaps are a transient that correctly resolved.

Measured over 11h of live `ALERTS` (Prometheus retention here is 12h — see the caveat at the bottom):

| | ever fired | pending, never fired | never pending |
|---|---:|---:|---:|
| **has a periodic subject** | 8 | **2 (both defective)** | rest |
| **no periodic subject** (traffic/kube-state/gauges) | 18 | 33 (all transient, `for:` doing its job) | rest |
| of which money-path | 6 fired | 0 defective | — |

**59 alertnames went `pending` in the window; 26 reached `firing`. Of the 33 that did not, 31 are healthy** — short, irregular episodes that resolved (`KubeNodeNotReady` 8m, `MoneyPathServiceDown` 3m, `PodsPendingSustained` 87 separate 1–7m episodes from Karpenter churn). Two are not:

| alert | episodes (11h) | length | start-gaps | `for:` | verdict |
|---|---|---|---|---|---|
| `AiCallErrorRateHigh` | 8 | `[6,15,15,15,15,15,15,15]` m | `[51,30,30,30,30,30,30,90]` m | 15m | **cannot mature** — length == window, gap == cron |
| `HighLatencyP99` (agent-service) | 12 | `[4,4,4,4,4,4,4,4,4,4,4,4]` m | `[30]×10` m | 10m | **cannot mature** — same fingerprint |

The second one is new, is not covered by #6041, and was found by this sweep.

### Static vs empirical — where they disagree, and why that is information

They disagree in **both** directions, and each disagreement is a real fact:

* **Empirical found what static could not.** `HighLatencyP99` is a *fleet-wide* rule over `traces_spanmetrics_latency_bucket`. Statically it has no subject — the metric comes from the collector, and the subject depends on which `service` label you are looking at. It is correct for the ~25 traffic-driven services and structurally blind for the one whose only traffic is a cron. **No static analysis over the rule text could have found this**; only the `ALERTS` history could, because the periodicity is in the data.
* **Static found what empirical could not.** Both total-failure-blind ratios are currently silent because there is *no onboarding traffic at all* — the empirical view shows nothing, correctly. The blindness only bites when attempts exist and all fail, which has not happened in the retention window. Reading "no alerts" as "no problem" here would have been exactly wrong.

## Replay proof for everything changed

**`HighLatencyP99`, agent-service, same 11h window** (why it is *not* being retuned here):

| window | true points | max consecutive |
|---|---:|---|
| `[5m]` (current) | 48 | **4m** × 12 episodes → `for: 10m` never matures |
| `[30m]` (== period) | 330 | 29m — still capped; `W == P` is a boundary, not a fix |
| `[45m]` | 355 | **311m** |

The condition really was continuous for 5+ hours. Only the window hid it.

**The two ratio fixes**, replayed against a *real* all-failing series (`openbank_llm_requests_total` for the #5736 model: 14 errors, 0 successes in 24h):

| form | result |
|---|---|
| `good / total` (pre-fix) | **EMPTY — blind** |
| `(good or vector(0)) / total` (post-fix) | `0` → fires |
| `1 - (good / total)` (pre-fix) | **EMPTY — blind** |
| `1 - ((good or vector(0)) / total)` (post-fix) | `1` → fires |

`promtool check rules` v3.12.0 on all three changed files: **exit 0**, 9 rules. Both fixed expressions were then submitted verbatim to the live Prometheus parser: `status=success`, empty result — which is the **correct resting state**, because their `>= 10` / `>= 20` traffic guards are false with no funnel traffic. That is also the **t=0 cold-pod** value: every counter starts at 0, the guard is false, the rule is silent. `or vector(0)` adds no boot-time false positive.

## The guard's negative control

**A gate is proven by what it prevents.** The negative cases were written first. Exit codes read from `$?`, never from a pipeline.

```
$ python3 .github/scripts/check-alert-window-vs-subject-period.py --self-test ; echo $?
  [ok] pre-#6041 AiCallErrorRateHigh (window 15m vs 30m subject): exit 1 (want 1), expected text found
  [ok] total-failure-blind success/total ratio: exit 1 (want 1), expected text found
  [ok] the same blindness spelled `1 - (good/total) > x`: exit 1 (want 1), expected text found
  [ok] windowed rule with no subject_period declaration: exit 1 (want 1), expected text found
  [ok] the same rule with the window widened past the subject period: exit 0 (want 0), expected text found
  [ok] healthy: window <= period but `for:` far shorter than the window: exit 0 (want 0), expected text found
  [ok] healthy: no `for:` at all, so one true evaluation fires: exit 0 (want 0), expected text found
0
```

Fed the exact pre-#6041 shape, it fails naming the defect:

```
::error ::WINDOW CANNOT ACCOMMODATE ITS SUBJECT — window [15m] with `for: 15m` against a
subject period of 30m. The counter increments once per window, so the condition is true in
runs of at most 15m and then decays, resetting the `for:` timer just short of maturity — the
rule can sit `pending` forever and never `firing`.
```

The three **healthy** controls matter as much as the three negative ones: `StatementCloseFailures`' real shape (`increase[1h]`, `for: 5m`, monthly subject) and `SubledgerTieOutBreak`'s (`increase[25h]`, no `for:`) both have `W <= P` and are perfectly fine. A gate that flagged them would be one that cannot tell "never fired because nothing broke" from "cannot fire", which is the failure this whole PR is about. So the arithmetic is stated exactly rather than as a fudge factor: `W > P` is safe outright; `W <= P` fails only when `for` is not comfortably shorter than `W`.

## Design of the gate

* **Scope is every windowed alert rule** — the subject cadences are **derived** (142 from `@Scheduled` + CronJob + charters), never a hand-kept list, because a gate whose scope is a hand-kept list of the thing it checks reads as *passing* when the list is short, never as *unchecked*.
* **The baseline is the frozen side, never a difference.** The 45 rules undeclared today are listed by name. The gate fails in **both** directions: a new windowed rule that declares nothing, a listed rule that has since declared, and a listed rule that no longer exists are all errors. The debt can only shrink.
* **It cross-checks its own reach.** It counts alert rules a second way and fails if the YAML walk saw fewer, and it fails on any rule file it cannot parse. This was not defensive writing — the first draft swallowed a `YAMLError` and **reported "no findings" about a file it had never read**, and the count comparison is what caught it.
* Registered in `gates.yaml` as `enforced` with `rationale:`, `review_after:` and `min_subjects: 90`, and prints `SUBJECTS=121` so `gate-subject-floor` can tell a gate that found nothing from one that found nothing wrong.

## What was fixed, and what was deliberately not

**Fixed** (unambiguous, no threshold judgement, strictly more sensitive in a currently-invisible case): both onboarding ratios.

**Not fixed, on purpose:**

* `AiCallErrorRateHigh` — `prometheus-rules-ai.yaml` is being rewritten by **#6041** in parallel. This PR is kept content-disjoint from it (verified with `git diff origin/main origin/fix/llm-model-5736 -- <file>`, not just filenames) and from **#6032** (`loki.yaml`, `check-loki-rules.py`). #6041's `AiModelNeverSucceeds` is rate-free and covers the outage.
* `HighLatencyP99` — the fix is a **judgement call**, not an arithmetic one. Widening the window fleet-wide costs every traffic-driven service its detection latency; the right shape is probably a per-subject companion rule. Baselined with the full evidence in the file rather than retuned in the dark. Tracked separately.

## Verification

`PR_DIFF_BASE=origin/main python3 .github/scripts/run-gates.py --all` → **159 gates, PASS=157, WARN=1, FAIL=1**. The single failure is `api-contract-gate`, which was confirmed to fail identically on clean `origin/main` with the change stashed; this PR touches no `openapi.yaml`. The warning is `release-scope-mismatch` failing on `PR_TITLE: unbound variable`, a local-run artifact.

## What was NOT verified

* **Prometheus retention here is 12h**, so "never fired" is established over 11h, not over all time. For `AiCallErrorRateHigh` the 8h40m outage in #6041 is the longer-baseline evidence; for `HighLatencyP99` the 12 identical episodes at an exact 30-minute cadence are strong but bounded by that window. Alertmanager keeps no longer history either.
* The changed rules were **not applied** to the cluster (read-only constraint). They are proven to parse under `promtool`, to parse and evaluate under the live Prometheus query parser, and their unchanged siblings are confirmed loaded — but the reloaded-rule behaviour will only be observable after ArgoCD syncs.
* Rules whose subject is a **Temporal Schedule** (the six campaign rules) have a period stored as per-campaign runtime data, not in code or gitops. They are baselined as undeclared rather than guessed at.

### DIFF
--- a/.github/gates/alert-subject-period-baseline.yaml
+++ b/.github/gates/alert-subject-period-baseline.yaml
@@ -1,0 +1,94 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) OpenBank contributors. Licensed under the Apache License, Version 2.0.
+#
+# THE FROZEN SIDE for check-alert-window-vs-subject-period.py.
+#
+# Every alert rule whose expression carries a rate/increase window must declare, in its
+# annotations, the invocation period of the SUBJECT it measures — `continuous` for something
+# driven by real traffic, or a duration for something driven by a cron. The gate then does the
+# arithmetic that #6041 proved nobody had done: a window no wider than its subject's period,
+# paired with a `for:` no shorter than that window, is a rule that can sit `pending` forever
+# and never once reach `firing`.
+#
+# Annotating all 49 windowed rules in the PR that introduces the gate would be a diff nobody
+# can review and would conflict with every parallel change to the alerting tree. So the rules
+# that were undeclared on the day the gate landed are listed here BY NAME.
+#
+# This list is the FROZEN SIDE, never a difference — openbank-infra/CLAUDE.md, on the Keycloak
+# baseline: "any baseline keyed to A - B where only B is frozen will fire on changes to A".
+# The gate fails in BOTH directions:
+#
+#   * a NEW windowed rule that is not listed here and declares nothing  -> fail
+#   * a listed rule that has since declared a subject_period            -> fail, stale entry
+#   * a listed rule that no longer exists                               -> fail, stale entry
+#
+# So the debt can only shrink, and a new alert cannot join it quietly. Removing an entry means
+# adding the declaration to the rule, which is the point.
+#
+# A listed rule is NOT asserted to be correct. It is asserted to be UNCHECKED. Two of them are
+# known to be defective and are called out where they appear below.
+undeclared_baseline:
+  # KNOWN DEFECTIVE — the rule this entire gate was derived from. `rate(...[15m])` with
+  # `for: 15m` against a 30-minute subject (agent-service OversightService,
+  # AGENT_OVERSIGHT_CRON `0 0/30 * * * ?`). Prometheus's own ALERTS series across the 8h40m
+  # #5736 outage: 44 points at `pending`, zero at `firing`; eight pending episodes of
+  # [6,15,15,15,15,15,15,15] minutes at 30-minute gaps — the gap is the cron and the length is
+  # the window. Left undeclared here ONLY because prometheus-rules-ai.yaml is being rewritten
+  # by PR #6041 in parallel and this change is deliberately kept disjoint from it. #6041 adds
+  # AiModelNeverSucceeds, which is rate-free and covers the outage; narrowing this rule itself
+  # is follow-up work on that file.
+  - alert: AiCallErrorRateHigh
+  - alert: AiCallsNotConfigured
+  - alert: AiSpendUnmeasured
+  - alert: BillingServerErrorsWhileHealthy
+  - alert: CampaignEnrolmentFailuresElevated
+  - alert: CampaignJourneysNotResolving
+  - alert: CampaignNoSendsInBusinessHours
+  - alert: CampaignSendHandoffsFailing
+  - alert: CampaignSendsAreAllDryRuns
+  - alert: CampaignSuppressingEverything
+  - alert: EmailChannelMockedOnly
+  - alert: EmailDeliveryRejected
+  - alert: FeedbackEmitFailing
+  - alert: FeedbackScreenshotStoreFailing
+  - alert: FinOpsEbsAttachmentFailing
+  - alert: FinOpsFleetEgressHigh
+  - alert: FinOpsKarpenterChurnOutsideMaintenance
+  - alert: FinOpsKarpenterNodeChurnHigh
+  - alert: FinOpsNatEgressSpikeAbsolute
+  - alert: FinOpsNatEgressSpikeRelative
+  - alert: FraudScoringMostlySynthetic
+  - alert: HighErrorRate
+  # KNOWN DEFECTIVE for one of its subjects — found by this sweep, and NOT covered by #6041.
+  # `rate(traces_spanmetrics_latency_bucket[5m])` with `for: 10m` is correct for the ~25
+  # services driven by real traffic, and structurally unable to fire for
+  # openbank-agent-service, whose only traffic is that same 30-minute cron. Measured over 11h
+  # of live ALERTS history: 12 pending episodes of exactly 4 minutes at exactly 30-minute gaps,
+  # zero firing. Replaying the same expression at [45m] over the same window yields 311
+  # consecutive true minutes, so the condition really was continuous and only the window hid
+  # it. Not fixed here: widening the window fleet-wide costs every other service its detection
+  # latency, so this wants a per-subject companion rule rather than a retune — a judgement
+  # call, tracked as its own issue.
+  - alert: HighLatencyP99
+  - alert: KyvernoAdmissionDenied
+  - alert: KyvernoAuditPolicyFailingBeforeEnforce
+  - alert: KyvernoEnforcePolicyBlocking
+  - alert: LendingHighErrorRate
+  - alert: NodeDiskWillFillIn4h
+  - alert: OnboardingAccountsNotFollowingParties
+  - alert: PartyEventDeadLettered
+  - alert: PaymentSLOFastBurn
+  - alert: PaymentSLOSlowBurn
+  - alert: PersistentVolumeWillFillIn6h
+  - alert: PostgresDeadlocksGrowing
+  - alert: PostgresLowCacheHitRatio
+  - alert: PostgresWALArchiveFailing
+  - alert: PushChannelSkippedOnly
+  - alert: PushDeliveryRejected
+  - alert: PushNoDevicesRegistered
+  - alert: ScreenFeedbackBugReported
+  - alert: ScreenFeedbackScreenshotStoreFailing
+  - alert: ScreenFeedbackSpike
+  - alert: SpanLatencyTrendAnomaly
+  - alert: StatementCloseFailures
+  - alert: TrafficDropAnomaly
--- a/.github/gates/gates.yaml
+++ b/.github/gates/gates.yaml
@@ -513,6 +513,34 @@ gates:
   # (public repo, GitHub-hosted runners), but 90 s on a required check is 90 s every
   # contributor waits on every PR. The STEP still runs unconditionally and prints whether it
   # loaded or skipped and why — a step that silently no-ops reads exactly like one that passed.
+  # #5736 / PR #6041. `AiCallErrorRateHigh` was loaded, `health=ok` and evaluating correctly
+  # throughout an 8h40m total outage, and never fired: Prometheus's own ALERTS series has 44
+  # points at `pending` and zero at `firing`. Its `rate(...[15m])` window is shorter than its
+  # subject's 30-minute invocation period, so the counter increments once per window and the
+  # ratio decays before `for: 15m` matures — a property of the CALLER, not of the rule, which
+  # is why no threshold tuning could ever have fixed it and why promtool, rule health and a
+  # green ArgoCD all agreed nothing was wrong. The same file also catches a ratio that is
+  # blind in the total-failure case, because a route that never succeeded has no
+  # `outcome="success"` series and the arithmetic then yields nothing rather than 0%.
+  # Scope is every windowed alert rule; the subject cadences are DERIVED from @Scheduled
+  # methods, CronJob manifests and agent charters, never a hand-kept list. ENFORCED.
+  - id: alert-window-vs-subject-period
+    name: "alert window accommodates its subject's period (#5736, enforced)"
+    group: gitops
+    mode: enforced
+    rationale: >-
+      An alert whose rate window is no wider than its subject's invocation period cannot ever
+      mature its `for:`, and every signal a human would check says the rule is fine — loaded,
+      `health=ok`, promtool-clean, ArgoCD green. `AiCallErrorRateHigh` sat `pending` for the
+      whole 8h40m of #5736 and never once reached `firing`; the fix is a property of the
+      caller, not a threshold, so nothing tuned it back into existence.
+    review_after: "2027-02-19"
+    min_subjects: 90   # 121 alert rules in the gitops PrometheusRules today (2026-08-20)
+    selftest: |
+      python3 .github/scripts/check-alert-window-vs-subject-period.py --self-test
+    run: |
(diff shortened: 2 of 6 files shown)


JiRaska:

### Verdict: the gate is real. I proved it fails on both defect classes and on stale baseline entries in both directions.

Ran in a throwaway worktree at `b970dd59a`, exit codes read from `$?` (never a pipeline), each mutation reverted:

| # | mutation | exit | finding |
|---|---|---|---|
| — | unmodified, `--enforce` | **0** | `SUBJECTS=121`; 49 windowed rules, 142 periodic subjects, 45 baselined, no findings |
| — | `--self-test` | **0** | 7 cases: 4 must-flag, 3 must-pass |
| M1 | `JourneyRunFailed` `[30m]`→`[3m]`, `for:`→`5m` (subject 5m) | **1** | `WINDOW CANNOT ACCOMMODATE ITS SUBJECT` — defect 1 caught |
| M2 | baseline entry for a rule that does not exist | **1** | `STALE BASELINE ENTRY — alert … no longer exists` |
| M3 | baseline entry for a rule that now declares `subject_period` | **1** | `STALE BASELINE ENTRY — it now declares 'subject_period: 5m'` |
| M4 | remove `or vector(0)` from `OnboardingSignConversionLow` | **1** | `RATIO IS TOTAL-FAILURE BLIND` — defect 2 caught |
| M7 | append unparseable YAML to a rules file | **1** | walk-vs-grep count mismatch (117 vs 121) **and** the named file |
| M8 | strip all 45 baseline entries | **1** | exactly 45 errors, **all** annotation-missing; zero hidden defect-2 or window findings |

The self-test is not vacuous — it asserts the *text* of each finding, and carries two healthy `W <= P` must-pass shapes, which is the discriminator this gate lives or dies on. M7 is the one I most appreciate: the "silent pass on a file the walker could not read" hole is genuinely closed by cross-checking two independent counts. That is better than most gates in this repo.

Subject set is **derived** (`derive_subjects()`, `:185-249` — `@Scheduled` cron/every with `application.yaml` placeholder resolution, `kind: CronJob` manifests, `agents.yaml` charters → 142 resolved, 0 unresolved), and the rule corpus is derived and cross-checked two ways (`:394-400`). This is the shape `pact-drift-check` lacked. Good.

### MAJOR — `min_subjects` floors the wrong denominator, so this gate's scope can collapse to zero and read green

`gates.yaml:537` sets `min_subjects: 90` against `SUBJECTS=121`, which `gatelib.subjects(parsed_count, "alert rules")` (`:496`) computes as **all** alert rules. But the set this gate actually enforces on today is **4 of 49** windowed rules — and three of those four were added by this PR:

```
OnboardingSignConversionLow        -> continuous   (added here)
OnboardingConsentToSignDropoffHigh -> continuous   (added here)
JourneyRunFailed                   -> 5m           (added here)
SubledgerTieOutBreak               -> 24h          (added here)
```

The other 45 windowed rules are baselined out. If someone re-baselined all 49 tomorrow, `SUBJECTS` would still print 121 and `gate-subject-floor` would still be green about a gate enforcing on nothing. The floor should be on `windowed − baselined`, which is the number that moves when coverage changes. This is the "a gate whose scope is a hand-kept list reads as passing when the list is short" pattern, one indirection out — the *subject* derivation is clean, it is the *exclusion* list that is unmeasured.

Not a reason to block: 4/49 on day one is an honest ratchet start. But the baseline must not be cited as coverage evidence until this is fixed.

### MAJOR — the baseline silences defect 2 as well as defect 1, and the file does not say so

`check-alert-window-vs-subject-period.py:413`:
```python
if blind and name not in baseline:
```
The baseline file's header (`alert-subject-period-baseline.yaml:29`) frames the list as rules with an *undeclared subject period*. But a rule listed there also has its **total-failure-blind ratio** check suppressed — two unrelated defects behind one key. M8 confirms no defect-2 instance is hidden today (all 45 findings were annotation-missing), so this is latent rather than live. Fix is one of: drop the `not in baseline` clause on `:413`, or split into two exclusion keys. The ratio check has nothing to do with declaring a period.

### MINOR — a follow-up with no issue number

`alert-subject-period-baseline.yaml:65-73` keeps `AiCallErrorRateHigh` and `HighLatencyP99` live while documenting them as known-defective, and describes the `HighLatencyP99` follow-up only as "tracked as its own issue" — with no number. This repo has a documented pattern of "filed separately" in a PR body not being a filing (5 siblings, 0 filed). Please put the number in, or file it.

### MINOR — `subject_period: continuous` is an unverified escape hatch

`:442` — declaring `continuous` skips the arithmetic entirely with no cross-check against the derived subject set. Both declarations added here are correct (`OnboardingFunnelPublisher.kt` is request-path — I verified), but a genuinely cron-driven rule mislabelled `continuous` would pass silently. Cheap guard: if a rule's metric is emitted by a class the subject derivation resolved to a period, `continuous` should be an error.

### MINOR — the new gate is inserted under a comment that documents a different gate

`gates.yaml:513-516`. On `origin/main` the block ending *"…a step that silently no-ops reads exactly like one that passed."* is the `--since`-scoping rationale for `loki-rule-load-test`. After this PR it sits above `alert-window-vs-subject-period`, which has no `--since`, and `loki-rule-load-test` is left uncommented. Move the new gate below it, or re-attach the block. Nothing in CI can see this.

### t=0 on a cold pod: checked every changed rule, none fires

| rule | expression after | t=0, no series |
|---|---|---|
| `OnboardingSignConversionLow` (`customer-edge/prometheus-rules.yaml:44-50`) | numerator wrapped `(sum(...) or vector(0))`, guard `sum(increase(…SIGN_ATTEMPT[1h])) >= 10` | guard arm empty ⇒ `and` empty ⇒ **silent** |
| `OnboardingConsentToSignDropoffHigh` (`:70-80`) | same wrap, guard `… >= 20` | guard arm empty ⇒ **silent** |
| `JourneyRunFailed` (`prometheus-rules-journey.yaml:64-65`) | **expression unchanged**, annotation only | `kube_job_status_failed` absent before the first Job ⇒ **silent** |
| `SubledgerTieOutBreak` (`prometheus-rules-tieout.yaml:34-35`) | **expression unchanged**, annotation only | no series ⇒ empty; after restart counter re-registers at 0, `increase`=0 ⇒ **silent** |

`or vector(0)` cannot fire alone — it only takes effect when the denominator exists and the numerator does not, i.e. the genuine 0%-conversion case. The fix is correct and strictly more sensitive.

Both declared periods check out against source: `cronjob-journey-public-edge.yaml:104` `schedule: "*/5 * * * *"` → 5m, W=30m > P=5m ✅; `TieOutScheduler.kt:113-114` `cron = "{openbank.ledger.tieout.cron:0 0 6 * * ?}"` → 24h, W=25h > P=24h ✅.

### Series existence: all three metrics have real emitters
- `onboarding_funnel_events_total` → `OnboardingFunnelPublisher.kt:62` `meterRegistry.counter("onboarding_funnel_events", …)`
- `openbank_subledger_tieout_break_total` → `TieOutScheduler.kt:96` `Counter.builder("openbank.subledger.tieout.break")`
- `kube_job_status_failed` → kube-state-metrics via `gitops/apps/kube-prometheus-stack.yaml`

Note for anyone repeating this: a literal grep of the *Prometheus* name finds nothing for either in-repo metric — Micrometer uses dotted names and the registry does the `_`/`_total` translation. Same probe trap as the event-key grep in root `CLAUDE.md`.

### MINOR — collision risk with open PR #6058
`feat/alert-firability-gate-v2` adds `check-alert-metric-emitted.py`: different script, **same corpus, same insertion region of `gates.yaml`**, and #6059/#6047 also touch `gates.yaml`. Not a duplicate — the questions are genuinely different (can the window see its subject vs. does the metric exist) — but both walk the alert-rule tree with their own YAML walker, and whoever lands second takes a textual conflict. Content-checked: no overlap under `openbank-infra/`.

### Checked, found nothing
- Alert rules outside `gitops/components/` that `--root` would miss: zero.
- Baseline hiding a live defect-2 instance: none (M8).
- Wrong `subject_period` on any of the four declared rules: none.
- `--root` self-test isolation (`:381`): fixtures do not inherit the real baseline.
- No scope creep; `fix(observability)` with no released path ⇒ correctly no release.

PR #6058, opened 2026-08-20
feat(ci): gate that an alert rule depends only on metrics some code emits

JiRaska:
Replaces #5739, which went `DIRTY` after #5758 was superseded by #5787. Rebuilt cleanly off current `main` and carries **only** the gate.

## What

An alert over a metric nothing emits is indistinguishable from a correctly-quiet one: it loads, `promtool` accepts the PromQL, Alertmanager routes it, and Prometheus lists it `inactive` forever. Four existed (#5733), found by a full observability audit of the live sandbox.

Same class as #5049 (16 dashboard metrics nothing emitted), one layer worse: an empty panel is at least visible to whoever opens the dashboard; a rule that never fires produces no artifact at all.

## Ships with ZERO debt

`KNOWN_DEAD` is **empty**. All four alerts are already resolved on `main`, by two different and both-correct routes:

- `TransactionSagaStuck` (critical, money-path) → **#5787** added `StuckSagaGauge`, so the metric exists and the alert can finally fire.
- The other three → **removed** in #5733. Instrumenting them would not have made them meaningful: all three watched a *throughput floor* on services with no cadence to miss (every entry point REST-driven, no `@Scheduled` sweep, no `@Incoming` consumer), so "no traffic in the last hour" is the normal resting state, not a fault. Where a cadence is added later, the fleet mechanism applies — `registerWorkflowLiveness` (ADR-0160) plus the existing `WorkflowLivenessStale` — not a bespoke per-service rule.

So the estate is at zero and this gate exists so it cannot quietly leave it. A new dead alert matches nothing in the baseline and fails immediately.

## Two bugs in the gate itself, caught by checking it against live Prometheus

I did not trust the gate's first output — cross-checking it against the live metric inventory is how both surfaced:

**1. It understood only one of three naming idioms**, so it flagged six live, working alerts. This repo writes metric names three ways:

```kotlin
"openbank.payments.submitted"                               // DomainMetrics, dotted
registry.counter("openbank_statement_close_failures_total")  // CloseMetricsAdapter, Prometheus-style
"openbank.ledger.accounting_day.stuck_cutoff_days"           // AccountingDayScheduler, mixed
```

Both sides are now canonicalised (separators collapsed, registry suffixes stripped) — finding count **14 → exactly the 4** live Prometheus independently confirms.

**2. `prom_to_micrometer` looped forever** — it set `changed` on every suffix match rather than only on a *newly derived* base, so the original element kept matching its own suffix on every pass.

## Prose is not instrumentation

Comments are stripped before the literal scan. Metric names appear in KDoc all over this tree — `FxFixingFreshnessPort` names three it does not emit — and counting prose as instrumentation is how a gate certifies a metric that exists only in a sentence.

## Selftest note

Baseline mechanics are exercised against a **synthetic** entry, not against whatever is in `KNOWN_DEAD`. With the dict now empty, a fixture reading from it would quietly stop testing anything — which is exactly the failure mode this gate is about.

## Verified

- selftest passes standalone and through `run-gates.py`.
- enforce clean: **121** alert rules, **222** emitted metric names, 5 recording rules, **0** baselined.
- `gatelib.subjects()` wired (`min_subjects: 80`), `ruff --select E9,F,B` clean, `rationale` + `review_after` set.

Refs #5733

### DIFF
--- a/.github/gates/gates.yaml
+++ b/.github/gates/gates.yaml
@@ -2378,6 +2378,25 @@ gates:
     run: |
       python3 .github/scripts/check-aggregate-type-case-fold.py --root .
 
+  # An alert rule over a metric nothing emits is indistinguishable from a correctly-quiet one:
+  # it loads, promtool accepts it, Alertmanager routes it, and Prometheus lists it `inactive`
+  # forever. Four existed, one `severity: critical` on the money path (TransactionSagaStuck,
+  # since fixed by #5787) — see #5733. The dashboard half of this class was #5049; the alert half
+  # is worse, because an empty panel is at least visible to whoever opens it while a rule that
+  # never fires produces no artifact at all.
+  - id: alert-metric-emitted
+    name: "every alert rule depends on a metric some code emits (issue #5733, enforced)"
+    group: data
+    mode: enforced
+    min_subjects: 80   # 115 alert rules in gitops PrometheusRule CRs today (2026-08-19)
+    rationale: "A critical money-path alert that cannot fire reads exactly like a healthy one; nothing in Prometheus, promtool or Alertmanager distinguishes them."
+    review_after: "2027-02-19"
+    selftest: |
+      python3 .github/scripts/check-alert-metric-emitted.py --selftest
+    selftest_expect: pass
+    run: |
+      python3 .github/scripts/check-alert-metric-emitted.py --enforce
+
   - id: kafka-acl-coverage
     name: "Kafka ACL coverage (issue #2598, enforced)"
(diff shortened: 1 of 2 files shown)


JiRaska:
Closing as subsumed by #6073: that PR is stacked on this branch and `git merge-base --is-ancestor` confirms every commit here is already contained in `feat/dashboard-metric-emitted`. Merging #6073 lands both gates, so reviewing this separately is duplicate work. Reopen if you'd rather land the alert-side gate on its own first.

Issue #6057, opened 2026-08-20
lending: real ledger/credit adapters are gated on BUILD-time properties set nowhere, while gitops activates them with RUNTIME env — the silent no-op appears to stay bound


JiRaska:
## What

`openbank-lending-service` binds its outbound ports through Quarkus `@IfBuildProperty`, which is
resolved at **build (augmentation) time**. The deployment activates them with **runtime** env
vars. A runtime env var cannot change a build-time property in an already-built image, so the
`@Default` no-op adapters appear to stay bound in the deployed image.

Measured on `origin/main`, from the repository alone:

- `RestLedgerPostingAdapter` is gated on `lending.ledger.backend == "rest"`.
- `BorrowerCreditClient` / `AccountServiceClient` are gated on
  `lending.borrower-credit.backend == "rest"`.
- **Neither property is set in any `application.yaml`, `build.gradle.kts`, `.properties`, build
  workflow or Dockerfile in the tree.** A repo-wide search for the property names returns nothing
  outside the annotations themselves.
- The Rollout at `openbank-infra/gitops/components/lending/lending-service.yaml` sets
  `LENDING_LEDGER_BACKEND=rest` and `LENDING_BORROWER_CREDIT_BACKEND=rest` as container env, and
  its own comments state the consequence if the value does not take effect: *"Without this the
  no-op default stays bound and all ledger posts are silently dropped."*
- The no-op that would then be bound, `NoOpLedgerPostingPort`, logs at **debug** and returns
  success.

If that reading is right, every general-ledger posting from lending is dropped, silently, and the
threat model's double-entry mitigation is inert by construction.

## Why nothing reports it

The no-op returns success and logs below the default level, so there is no error to count and no
metric that distinguishes a real posting from a discarded one — the same shape as
`PushResult.skipped()` carrying `success = true`, and as #6037's compensation stub. A unit test
that mocks the port cannot see which implementation CDI bound.

## What is NOT established here

**Runtime state.** This is a source-and-config reading. Confirming which bean the running image
actually has requires inspecting the deployed application — e.g. asking a pod which
implementation is bound, or checking whether any ledger entry exists whose origin is lending.
That check should be run before deciding urgency, exactly as #6037 was triaged.

## Suggested fix

1. Decide the gate's kind deliberately. If the choice must be per-environment, it cannot be
   `@IfBuildProperty` — use a runtime-resolved producer, or set the property at build time and
   stop implying the env var does anything.
[...]

Issue #6056, opened 2026-08-20
sepa-payment: the /returns non-repudiation control does not exist — no audit publisher in the service


JiRaska:
## What

`docs/threat-models/openbank-sepa-payment.md`, **R — Repudiation**, claimed:

> All `/returns` invocations logged via existing `AuditService` with correlation id,
> `OrgnlEndToEndId`, reason code, and actor identity

`AuditService` is present in no source file in this repository, and
`openbank-sepa-payment/src/main` contains **no audit publisher of any kind**. Nothing records a
return invocation as evidence.

The document half of this is corrected in the PR that files this issue (the row now states the
control does not exist). This issue is the **code** half, which a doc edit must not be allowed to
close.

## Why it matters more than an ordinary missing feature

The risk is denial of having processed a return — a claim about who did what. The two things that
do happen today are an application log line and the payment row's own status transition. Neither
is a non-repudiation control, because both are written by the same code path whose behaviour is
in dispute; they can only ever agree with it.

Adjacent, and worth deciding together: across the fleet the `AuditEvent` envelope is credited in
several money-path threat models while the only implementation of `AuditEventPublisher` is the
logging fallback. A log line is not an evidentiary record, and the repo already has an
append-only `audit_entries` store that these paths do not write to.

## Suggested fix

1. Publish an audit event on the return path with correlation id, the original end-to-end
   reference, the reason code and the **authenticated** actor (from the token, not the request
   body — see #4754 for the sibling shape where the actor came from the wrong place).
2. Assert it with a real-DB integration test that reads the row back. A mocked publisher cannot
   establish that a record exists.
3. Only then restore the R row to a positive claim.

## Detection

`.github/scripts/check-threat-model-claims.py` reports this row as PHANTOM.

Refs #6037, #6048.

Issue #6055, opened 2026-08-20
settlement: the OPA activity-authorization control credited by S1/E1 does not exist, and residual risk 2 is marked Closed on it


JiRaska:
## What

`docs/threat-models/openbank-settlement-service.md` credits an OPA activity-authorization
control that has never been built. Three separate places assert it:

- **S1** (Spoofing) — "OPA activity interceptor (`OpaActivityInterceptor`) rejects tasks not
  matching policy".
- **E1** (Elevation of privilege) — "only activities matching
  `data.openbank.settlement.activity.allow` are dispatched; `reverseBookToLedger` requires
  `compensation=true` context".
- **Residual risk 2** — marked **Closed**, on the reasoning that the policy file now exists "so
  `OpaActivityInterceptor` has a real policy to evaluate rather than fail-closing on a missing one".

## Measured state on `origin/main`

- `OpaActivityInterceptor` exists in **no Kotlin source anywhere in the repository**. Its only
  occurrences are that threat model and a label string on an admin-UI page — i.e. a name rendered
  in a UI, not a control.
- `openbank-settlement-service/src/main/resources/opa/settlement_activity.rego` does exist, and is
  referenced by **no bundle ConfigMap**. Outside its own file its only mentions are the threat
  model and a runbook. Nothing loads it, so nothing evaluates it.
- The policy also does not say what E1 says it says. Its `allow` rule is a single membership test
  of `input.activity` against a fixed set, and `reverseBookToLedger` is an unconditional member of
  that set. There is no `compensation` term in the file at all.

So the mitigation for the settlement worker's activity-dispatch boundary is: nothing. Not a weak
control, not a misconfigured one — an absent one, documented as present and explicitly signed off
as closed.

## Why it survived

Same shape as #6037 and as the earlier settlement correction in #3921 (which credited a service
mesh that has never been deployed here): a threat model is a set of claims about running code, and
until now nothing in CI read a single claim. The residual-risk entry made it worse than silence,
because "Closed" is the state a reviewer stops re-checking.

## Suggested fix

Two separable pieces, and they should not be conflated:

1. **Decide whether the control is wanted.** If activity-level authorization is the right boundary
   for the Temporal worker, build the interceptor and bundle the policy the way every other
   service's `*_rest_ext.rego` is bundled. If the REST-layer gate on origination is considered
   sufficient, say so and delete the policy file rather than leaving an unloaded one that reads as
   coverage.
[...]

PR #6059, opened 2026-08-20
docs(governance): check that threat-model mitigation claims resolve to real code

JiRaska:
## What

A threat model is a set of **claims about running code**, and nothing in CI read one.

`check-threat-models.py` proves a model EXISTS. `check-threat-model-diff.py` proves it is
TOUCHED when a trust boundary moves. Neither reads a single claim, so a model can be present,
current, freshly edited — and false. This adds the third half: a gate that reads the claims.

Prompted by #6037 / PR #6048, which proved the class rather than hypothesising it: settlement
**R2** listed a money-moving stub as its own mitigation, **T1** named an idempotency key built
from identifiers that appear in no source file here, and three unit tests asserted the defect
verbatim — so the suite agreed with the document and both were wrong together.

Refs #6037, #6048. Closes nothing on its own; files #6055, #6056, #6057.

## The audit — enumerated twice, because one probe is not a census

The fleet's 23 money-path threat models are **not one format**, so the claim count depends on
how you look:

| enumeration method | knows about | claims found |
|---|---|---|
| **table parser** — locate the Mitigation/Control column from a header row | column headers | **296** |
| **section parser** — every table row and bullet inside a STRIDE-titled section | headings only | **315** |
| **union** (what the gate checks) | both | **390** |

The disagreement is the finding. Two services carry no Mitigation column at all and are
**invisible to the table parser entirely**; one is invisible to the section parser. A single
grep here would have reported a clean, confident, wrong number — the same shape as grepping
`src/test` for the word "contract" and scoring comment lines as contract tests (#2291).

## Classification

**179 claims were classified by opening the cited code**, across **15 of the 23** money-path
services (ledger, balance, transaction, clearing, sepa-payment, domestic-payment, sca, consent,
delegation, fraud, sanctions, lending, standing-order, sdd, psd2), plus settlement verified
separately. Every (a)/(b)/(c) below rests on a `file:line`; nothing was upgraded from "I found no
problem" to "true". A further 7 claims are marked partial and are counted as unchecked, not as (a).

**Granularity caveat, stated rather than smoothed:** the (a)/(b)/(c) counts are per *mitigation
cell* as a human reads them; the (d) figure comes from the gate's own union enumeration, which
splits some cells differently. The two are close but not the same denominator, and the share
column should be read as a proportion, not a precise count.

| class | meaning | count | share |
|---|---|---|---|
| **(a) verifiable and TRUE** | cited mechanism exists and does what the claim says | **98** | ~26% |
| **(b) verifiable and FALSE** | mechanism absent, or present and not doing what is claimed | **48** | ~12% |
| **(c) self-referential** | the mitigation IS the thing at risk | **7** | ~2% |
| **(d) unfalsifiable as written** | no mechanism named, so no check could confirm or deny | **262** measured by the gate | ~2/3 |

**The headline is (d), not (b).** The gate measures **262 of 390** union claims naming no class,
file, config key or policy rule at all — roughly two in three. Those are not wrong — they are *unfalsifiable*, which is worse in one
specific way: a control that cannot be checked cannot be found missing, and this is precisely the
soil the R2/T1 defects grew in.

Class (b) is not concentrated in settlement. **Settlement was not the only bad one** — it was the
one somebody happened to look at.

Highest-severity (b)/(c) findings, described abstractly (public repo):

- **lending** — the real GL-posting and borrower-credit adapters are gated on Quarkus
  **build-time** properties that are set in no build file anywhere in the tree, while the
  deployment activates them with **runtime** env vars, which cannot flip a build-time gate. The
  no-op that would then stay bound logs at debug and returns success. → **#6057**
- **settlement** — S1/E1 credit an OPA activity-authorization control that exists in no Kotlin
  source; its policy file is loaded by nothing, and does not say what E1 said it said. Residual
  risk 2 was marked **Closed** on it. Separate defect from #6037 and outside #6048's scope. →
  **#6055**
- **sepa-payment** — the `/returns` non-repudiation control names a service that does not exist;
  the service contains no audit publisher at all. → **#6056**
- **fleet** — `authz.four-eyes.enforce` defaults false in every money-path `application.yaml` and
  is set nowhere in gitops, so the `X-Approval-Id` validation several models credit is unreachable
  in every shipped configuration. Folded into #6057 rather than filed separately, being adjacent
  to #5679 / #4754 / #3679.
- **(c)** — three services answer "did the real side-effect happen?" with a status write or an
  emitted event. Same shape as R2.
- **audit is the most over-claimed control in the fleet**, credited across several services whose
  only `AuditEventPublisher` implementation is the logging fallback.

## The guard

`.github/scripts/check-threat-model-claims.py`, registered enforced as `threat-model-claims-resolve`.

Fails on the mechanically detectable subset only:

- **PHANTOM** — a mitigation cites a symbol present in no tracked backend source.
- **STUB** — a cited symbol whose own *declaration* opens with a stub marker.
- **SELF-REF** — the threat is "the record says it happened" and the mitigation's only named
  mechanism is record-keeping.

It **counts but does not fail on** class (d): failing there would make "delete the sentence" the
cheapest fix, which is the wrong incentive.

Scope is **derived** (`money_path_services` x `docs/threat-models/<service>.md`), never hand-kept.
Exclusions carry a reason and fail **stale in either direction** — that fired during development,
on one of its own entries, the moment the row it excused was corrected.

The escape hatch is deliberately costly: the only way to silence a finding without writing code is
to state in the threat model that the control does not exist — which is the outcome the gate wants.

### Negative control — a gate is proven by what it PREVENTS

Exit codes read from `$?` with output redirected to a file, never through a pipe (`| tail` returns
tail's status):

```
$ python3 .github/scripts/check-threat-model-claims.py --enforce > /tmp/neg.txt 2>&1 ; echo $?
1
::error file=docs/threat-models/openbank-settlement-service.md,title=threat-model claim (PHANTOM)
  ::openbank-settlement-service S1: cites `OpaActivityInterceptor` — present in no tracked backend source
FAIL: 1 false/self-referential claim(s), 0 stale exclusion(s)

$ # restore the corrected text
$ python3 .github/scripts/check-threat-model-claims.py --enforce > /tmp/pos.txt 2>&1 ; echo $?
0
OK: every mechanically checkable mitigation claim resolves to real, non-stub code
```

Against the full pre-fix corpus the same run reported **7 findings across 3 services**, covering
all three rule kinds — including R2 as SELF-REF and T1's two phantom identifiers, i.e. the gate
independently rediscovers the defect it was built from.

`--self-test` (0 failures) pairs each real defect with its corrected text, so a rule that stops
discriminating fails there rather than silently going green on the fleet.

## What this PR fixes, and what it deliberately does not

**Fixed (the document was wrong):**
- `openbank-transaction-service` — credited a test that was never committed. The coverage it was
  credited for genuinely exists in a sibling test, so the claim was over-attributed, not unfounded.
  The change-log mention is left in place with a note: a change log records what was claimed on the
  day, and correcting it silently erases the evidence that it was.
- `openbank-sepa-payment` — the R row now states the control does not exist.
- `openbank-settlement-service` — S1, E1 and residual risk 2 corrected; risk 2 is explicitly
  **reopened** rather than quietly edited.

**Not fixed here, on purpose:** where a claim is false because the **code** is wrong, the document
was not rewritten to match broken code. Those are #6055, #6056, #6057.

**Left to PR #6048:** settlement R2 and T1, excused by row (never by service) so the gate stays
awake on that file. Each entry dies on its own once #6048 lands.

## What was NOT verified

- **No runtime or live-system state.** Everything rests on `src/main`, migrations, `rules.yaml`,
  and gitops manifests. #6057's conclusion in particular needs a runtime confirmation before
  urgency is decided, and says so.
- **Policy effect, as opposed to policy existence.** Cited rego rules were confirmed to exist and
  read on the expected action; no `opa eval` was run against a bundle, so which reason actually
  fires is unestablished — a distinction this repo has been burned by before.
- **Infra-layer claims** (mTLS, NetworkPolicy, mesh identity) are not decidable from `src/main`.
- **The 262 class-(d) claims** were not checked, deliberately: with no mechanism named, no check
  could confirm or deny them. That is the finding, not a gap.
- **8 of the 23 money-path services were not deep-audited claim-by-claim** (account, sepa-instant,
  swift, fx, interest, billing, vop, security-adjacent). They were covered by the gate's mechanical
  sweep, which found nothing in them — but a mechanical sweep is a weaker instrument than opening
  the code, so "no finding" there means "not looked at properly", not "clean".


JiRaska:
Relaying a coverage finding from the author of #6072 (sepa-payment `/returns` audit evidence), who
hit this while working the same row this PR corrects. Not my own measurement — but it was
positive-controlled, which is why it seems worth acting on.

**`check-threat-model-claims.py` appears not to parse sepa-payment's §5a supplement table.**

Measured on #6072's branch:

- Running the gate against that branch reports **no finding for sepa-payment**, before or after the
  fix.
- **Positive control 1**: reinstating the original phantom `AuditService` row — still no finding.
- **Positive control 2**: inserting a deliberately nonexistent class name into that row — still no
  finding.
- **Negative control (the gate does work elsewhere)**: it *does* flag `PaymentSagaLedgerIT` in the
  transaction-service model on the same run.

So the gate is not broken in general; it looks blind to this particular table shape.

Why it matters here specifically: issue #6056 states this row is reported as PHANTOM. If the gate
cannot see the row, then the count this PR reports (48 FALSE / 7 SELF-REF / 262 unfalsifiable) is a
floor rather than a census, in the same way the openapi-enum gate's `MIN_OVERLAP` threshold made 27
a floor rather than a census in #5962. The claims it structurally cannot pair are exactly the ones
nobody has checked by hand either.

Suggested next step before relying on the coverage number: run the parser over every money-path
model and assert the per-file claim count against a second, independently written extractor — the
same cross-count this PR's sibling guards adopted after `check_ruler_wiring()` and
`check-audit-money-path-subscription.py` each found their own reach was smaller than assumed. A
file that yields zero claims is the signal to look at.

I have not verified this myself against this PR's own branch state, only relayed it — worth
confirming here before treating the enumeration as complete.

PR #6061, opened 2026-08-20
fix(fraud): bucket velocity aggregates by event time, so a replay crossing an hour boundary cannot land in a different row (#6044)

### LINKED ISSUE #6044: fraud: velocity_aggregates buckets on PROCESSING time, so a replay crossing an hour boundary lands in a different row and no dedupe guard can see it
`VelocityAggregateRepositoryImpl.recordTransaction` picks the window bucket from `Instant.now(clock)` — the moment the Kafka record is **processed** — not from the event's own `occurredAt`:

```kotlin
val now = Instant.now(clock)
VelocityWindow.entries.forEach { window ->
    val start = window.bucketStart(now).toOffsetDateTime()
```

`window_start` is part of the primary key `(account_id, velocity_window, currency, window_start)`. So the same event, delivered twice either side of an hour (or day, or week) boundary, targets **two different rows**. Any redelivery guard — the V5 last-writer marker, or the applied-id set #5789/#6040 replaces it with — is per row by construction, so it cannot see the first application at all. The replay is not suppressed; it is counted a second time, in the next bucket.

This is the residual half of #5789, and it was called out explicitly in #6040's "noticed in passing, deliberately not fixed here". It is the case the set-membership guard cannot reach however large the window is set.

### Why it is reachable

`TransactionSignalConsumer` is at-least-once with periodic offset commits: a rebalance, a pod roll or a crash re-delivers the whole uncommitted window from the last committed offset. There is no bound on how long that takes to redeliver, and hour boundaries arrive once an hour. A DLQ replay (#5715/#5751) makes it routine rather than incidental, but does not create it.

### What is available

`openbank.transactions.transaction.initiated` carries a real business event time. `TransactionInitiatedEvent` declares `occurredAt: Instant` as a **required, non-defaulted** constructor parameter and is serialised whole by `LoggingTransactionEventPublisher.initiatedPayload`; the payload is stored verbatim in the outbox, so a replay carries the *same* `occurredAt` as the original delivery. `TransactionSignalConsumer` already parses it and already uses it for the payee-history row and the ADR-0140 feature store — only the velocity path ignores it.

### Semantics this changes

Bucketing on event time makes the aggregate mean "transactions that **occurred** in this window" instead of "signals that were **processed** in this window". That is the correct reading of a velocity control and the only one that is replay-stable or reconcilable against the source transactions. The cost is that a late-arriving signal now increments a bucket that is already in the past. Nothing in this service closes or finalises a bucket, so no reader is invalidated — but it is a behaviour change and belongs in the PR body, not in a commit.

### Out of scope for the first fix

The **read** side (`findAggregate`) uses `Instant.now(clock)` for the current bucket, which is right for "score the transaction happening now". Whether scoring should instead read the bucket of the scored transaction's own event time is a separate product question and a separate change.

Refs #5789.

JiRaska:
Closes #6044.

## Stacking — read this first

**This PR is stacked on #6040** (`fix/fraud-dedupe-ordering-5789`) and its base is that branch, not `main`. The diff shown here is my change alone. It is not a disjoint change and could not be: #6040 rewrites `VelocityAggregateRepositoryImpl`'s upsert SQL and constructor, and this PR changes the same function's signature and the line that computes its bucket. Branching from `main` would have produced a guaranteed conflict in the same method, and — worse — the two changes are about the *same* replay, so reviewing them apart from each other loses the argument. **Merge #6040 first.**

Evidence for the overlap rather than an assertion of it: `git diff origin/main origin/fix/fraud-dedupe-ordering-5789 -- .../VelocityAggregateRepositoryImpl.kt` and this PR's diff of the same file both touch `recordTransaction`'s parameter list and body.

**No migration**, so there is no `V6`/`V7` collision to resolve with #6040 or with anything on `main`. The schema is unchanged; only the value written into the existing `window_start` column changes.

## The defect

`window_start` is part of the `velocity_aggregates` primary key, and it was derived from `Instant.now(clock)` — the moment the Kafka record is **processed**:

```kotlin
val now = Instant.now(clock)
VelocityWindow.entries.forEach { window ->
    val start = window.bucketStart(now).toOffsetDateTime()
```

So a signal that occurred at 10:59 and is redelivered at 11:00 addresses a **different row** from the original. #6040's applied-id set lives on the row, so the replay meets an empty set, is not recognised, and is applied a second time. This is not a case a larger `openbank.fraud.applied-signal-window` reaches — the set it consults belongs to the wrong row. It is the residual #6040 named in its own "noticed in passing, deliberately not fixed here".

The delivery is ordinary, not exotic: `TransactionSignalConsumer` is at-least-once with periodic offset commits, so a rebalance, pod roll or crash re-delivers the uncommitted window with no bound on the delay, and hour boundaries arrive once an hour.

## Is `occurredAt` actually there, and is it trustworthy?

Yes, and this was established by reading the **serialised type**, not by grepping for a quoted field name — this repo builds event payloads two ways and a string grep can only see the hand-built maps (#3883).

- `TransactionInitiatedEvent` (`openbank-transaction-service/.../domain/event/TransactionEvents.kt`) declares `override val occurredAt: Instant` as a **required, non-defaulted** constructor parameter. There is no `Instant.EPOCH` default for a caller to silently take — the trap that made every AML decision record 1970.
- `LoggingTransactionEventPublisher.initiatedPayload` serialises the whole event with `objectMapper.writeValueAsString(...)` and sets `occurredAt = Instant.now(clock)` at the moment the transaction is initiated. That is a genuine business time, not an ingest time substituted downstream.
- The payload is stored **verbatim** in the outbox and relayed unchanged, so a replay carries the *same* `occurredAt` as the original delivery. That property is what makes the fix work at all: same event, same time, same row.
- `docs/asyncapi/openbank-events.yaml` records which channels carry no event time. `openbank.transactions.transaction.initiated` is not among them.
- `TransactionSignalConsumer` already parses `occurredAt` and already uses it for the payee-history row and the ADR-0140 feature store. **Only the velocity path ignored it.**

## The missing case is now visible instead of silent

`occurredAt` is required on the wire, so the absent case should never occur — which is exactly why it must be counted rather than absorbed. The consumer still substitutes processing time (a fraud control with an invented bucket beats no row at all), but it now logs a warning and increments `openbank_fraud_signal_missing_event_time_total{service="fraud"}`.

Silence is the failure mode here: the audit consumer substitutes ingest time for **7 of its 21 topics** and nothing anywhere says so (#3883). The same substitution already existed one line down in this file — `recordPayeeHistory` fell back to the clock, unremarked — and now shares the counter. **The expected value of this series is a flat zero**; any non-zero rate is a producer defect, not a tolerable degradation.

## The semantics trade-off, stated rather than assumed

Bucketing on event time changes what the aggregate **means**: "transactions that OCCURRED in this window" instead of "signals that were PROCESSED in it".

- **Why that is the right meaning.** It is the only one that is stable under replay, and the only one that can be reconciled against the source transactions — the reconciliation #6040 makes possible to alert on. Under processing time the H1 count was never a fact about any hour; it was a fact about the consumer's schedule.
- **The cost, and it is real.** A late-arriving signal now increments a bucket that is already in the past. Nothing in this service closes, finalises or snapshots a bucket — rows are only ever read for the current window, and no downstream reads a past one — so no reader is invalidated. But it is a behaviour change and it belongs here rather than in a changelog line.
- **What did NOT change, and is a genuine open product question.** `findAggregate` still derives the current bucket from `Instant.now(clock)`. That is right for "score the transaction happening now", and it is what the scorer wants. Whether scoring should instead read the bucket of the **scored transaction's own event time** is a different question with a different blast radius (it changes what the rules see, not only where writes land), and it is not decided here. Under normal operation the two agree: a live signal's `occurredAt` is milliseconds old and falls in the current bucket. They diverge exactly at a boundary or a replay — which is the case this PR is about, and where the write side is unambiguously wrong today.
- **Deliberately unchanged:** a signal with no `aggregateId` is still applied unconditionally and never remembered, and the guard is still per row, so the three windows still converge independently after a partial failure.

## Proof

Written failing-first, against **real Postgres** (`PostgresRedisTestResource`), as #6040's ITs are — a mocked `PgPool` cannot tell a write that lands in the right row from one that does not.

Two new ITs. The first constructs the exact delivery: the same signal (same id, same `occurredAt` at 10:59) applied through two repository instances differing **only** in the clock they would bucket by — 10:59:01 for the original delivery, 11:00:02 for the replay. The second is the same defect without a replay at all: two signals that occurred in the same hour, processed either side of the boundary, used to be split across two rows so neither the count nor the sum the scorer reads was ever the truth about that hour.

The first test's primary assertion is deliberately **across every H1 bucket for the account**, not on one row. Asserting only the 10:00 row would agree with the defect — that row also reads 1, and the second application is sitting in a row the assertion never looks at.

### Negative control

Reverting **only** the production line (`window.bucketStart(occurredAt)` → `window.bucketStart(Instant.now(clock))`), keeping the tests, the port signature and the metric:

```
VelocityAggregateRepositoryImplIT   12 tests, 2 failed, 0 skipped

RED: a replay processed after the hour boundary is not counted again in the next bucket
     [applications across all H1 buckets]
     expected: 1L
      but was: 2L

RED: two signals from the same hour processed either side of the boundary land in one bucket
     [count in the bucket containing 2026-07-09T10:00:00Z]
     expected: 2L
      but was: 1L
```

Exactly the two new tests go red; the other ten — including all four of #6040's dedupe tests — stay green, which is the point: the applied-id guard is working and still cannot see this.

## Verified

Gradle run serialised, output to files, exit code read directly (never through a pipe), `--rerun-tasks`, totals read from the JUnit XML rather than the console:

- `:openbank-fraud-service:test` — **179 tests, 0 failures, 0 errors, 1 skipped**. The skip is `FraudPactProviderVerificationTest`, `@EnabledIfSystemProperty(pactbroker.url)`; its `@PactFolder` sibling ran. 179 = #6040's 175 + the 4 tests added here.
- `:openbank-fraud-service:quarkusBuild` — green. The consumer gained a constructor parameter, and CDI/ArC wiring is not validated by unit tests.
- `ktlintCheck`, `detekt` — green (re-run on their own after the test run, since Gradle stops at the first failing task).
- Gate manifest, `PR_DIFF_BASE=origin/main`, per shard in the foreground: `security` (includes **threat-model coverage** and **threat-model-updated-on-trust-boundary-change**, both PASS), `kotlin`, `lint`, `registry`, `data` — all exit 0.

## Not verified

- **The `api` shard fails locally for an environmental reason, not a finding**: the `api-contract` gate tries to install `oasdiff` and needs `sudo`, which has no terminal here (`sudo: a terminal is required to read the password`). No `openapi.yaml` is touched by this PR. CI is the authority on it.
- No live system was touched — no cluster, DB, Kafka, Prometheus or AWS read or written. Every claim about the producer, the topic and the payload is read from the repository at `origin/main`.
- Whether double-counting from a boundary-crossing replay has actually occurred in a real environment. That needs the aggregate reconciled against the source transactions; this PR does not add that reconciliation (nor did #6040).
- No backfill or repair of aggregate rows already written into the wrong bucket. Existing rows are left as they are: the counts self-heal as their windows roll off, and there is no safe general repair without the reconciliation above.

## Release

`fix(fraud)` touching `src/main` — release-please cuts the patch. `version.txt` is not hand-edited. No API contract change, so no `openapi.yaml` version bump.

### DIFF
--- a/openbank-fraud-service/src/main/kotlin/com/openbank/fraud/application/port/out/FraudMetricsPort.kt
+++ b/openbank-fraud-service/src/main/kotlin/com/openbank/fraud/application/port/out/FraudMetricsPort.kt
@@ -36,4 +36,18 @@ interface FraudMetricsPort {
      * known to happen means the guard is not being reached.
      */
     fun recordSignalReplaySuppressed(aggregate: String)
+
+    /**
+     * Record that a transaction signal arrived with **no `occurredAt`** and that processing time was
+     * substituted for it (issue #6044). Every velocity bucket and payee-history row written from
+     * such a signal asserts a business time nobody measured, and — because the bucket is part of the
+     * primary key — a replay of that same signal will not land in the same row, so the redelivery
+     * guard cannot reach it. The substitution is otherwise completely silent: it writes a normal row
+     * and logs nothing, exactly like the ingest-time substitution #3883 found in the audit
+     * consumer. This counter is the only thing that distinguishes "every producer sends an event
+     * time" from "we have been inventing them". `openbank.transactions.transaction.initiated`
+     * declares `occurredAt` as required, so the expected value is a flat ZERO; any non-zero rate is
+     * a producer defect, not a tolerable degradation.
+     */
+    fun recordSignalMissingEventTime()
 }
--- a/openbank-fraud-service/src/main/kotlin/com/openbank/fraud/application/port/out/VelocityAggregateRepository.kt
+++ b/openbank-fraud-service/src/main/kotlin/com/openbank/fraud/application/port/out/VelocityAggregateRepository.kt
@@ -7,6 +7,7 @@ package com.openbank.fraud.application.port.out
 import com.openbank.fraud.domain.model.VelocityAggregate
 import com.openbank.fraud.domain.model.VelocityWindow
 import java.math.BigDecimal
+import java.time.Instant
 import java.util.UUID
 
 /** Stores and queries per-account rolling velocity aggregates (ADR-0084 §2). */
@@ -22,8 +23,25 @@ interface VelocityAggregateRepository {
      * not applied — the windows converge independently rather than as one transaction. A null
      * [transactionId] carries no identity and is therefore never deduplicated (same contract as
      * [PayeeHistoryRepository.recordPayment]).
+     *
+     * The window bucket is derived from [occurredAt] — the event's own business time — and never
+     * from the moment the signal is processed (issue #6044). Processing time puts a redelivery that
+     * crosses an hour, day or week boundary into a DIFFERENT row from the original, and the
+     * idempotency above is per row, so no guard of any design could recognise it there. Bucketing on
+     * event time is therefore what makes the redelivery guard reach a delayed replay at all; it also
+     * makes the aggregate mean "transactions that OCCURRED in this window" rather than "signals that
+     * were PROCESSED in it", which is the only reading that reconciles against the source
+     * transactions. The caller is responsible for supplying a real event time; see
+     * [com.openbank.fraud.infrastructure.messaging.TransactionSignalConsumer] for what it does when
+     * the event carries none.
      */
-    suspend fun recordTransaction(accountId: UUID, amount: BigDecimal, currency: String, transactionId: UUID?)
+    suspend fun recordTransaction(
+        accountId: UUID,
+        amount: BigDecimal,
+        currency: String,
+        transactionId: UUID?,
+        occurredAt: Instant,
+    )
 
     /**
      * Returns the current aggregate for [accountId] in [window] for [currency], or null.
--- a/openbank-fraud-service/src/main/kotlin/com/openbank/fraud/infrastructure/messaging/TransactionSignalConsumer.kt
+++ b/openbank-fraud-service/src/main/kotlin/com/openbank/fraud/infrastructure/messaging/TransactionSignalConsumer.kt
@@ -6,6 +6,7 @@ package com.openbank.fraud.infrastructure.messaging
 
 import com.fasterxml.jackson.core.JacksonException
 import com.fasterxml.jackson.databind.ObjectMapper
+import com.openbank.fraud.application.port.out.FraudMetricsPort
 import com.openbank.fraud.application.port.out.PayeeHistoryRepository
 import com.openbank.fraud.application.port.out.VelocityAggregateRepository
 import com.openbank.fraud.application.usecase.FeatureOnlineUpdater
@@ -51,6 +52,7 @@ class TransactionSignalConsumer(
     private val featureUpdater: FeatureOnlineUpdater,
     private val objectMapper: ObjectMapper,
     private val clock: Clock,
+    private val metrics: FraudMetricsPort,
 ) {
     private val log = Logger.getLogger(TransactionSignalConsumer::class.java)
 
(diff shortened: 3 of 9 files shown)


JiRaska:
## Review

Stacking verified: `git merge-base --is-ancestor origin/fix/fraud-dedupe-ordering-5789 origin/fix/fraud-velocity-window-event-time` passes, and the PR base is correctly pointed at the parent branch, so the 301/-44 diff GitHub shows is the child's own work and not #6040's. No gate confusion there.

The core argument is right and well made: `window_start` is part of the PK, so bucketing on processing time puts a delayed replay in a *different row* from the original, and since the applied-id guard is per row, no guard of any design could reach it there. Deriving the bucket from `occurredAt` is what makes #6040's guard actually cover a delayed replay. `recordSignalMissingEventTime()` is the right instinct too — counting the substitution instead of doing it silently is exactly the `#3883` lesson, and stating the expected value as a flat zero rather than a tolerable rate is the part most people leave out.

### Blocking on #6040

This PR inherits the NULL-safety regression I've detailed on #6040 (`= ANY` over an array containing NULL evaluates to NULL, not FALSE, so a single null-`aggregateId` signal permanently freezes the row and every later legitimate transaction is silently dropped). It is in this branch's `VelocityAggregateRepositoryImpl.kt` too. Nothing to change here beyond taking the parent's fix.

### Finding — writes now bucket on event time, but reads still bucket on processing time

`VelocityAggregateRepositoryImpl.kt` write path:

```kotlin
val start = window.bucketStart(occurredAt).toOffsetDateTime()   // event time
```

`findAggregate` read path, unchanged:

```kotlin
val start = window.bucketStart(now)                              // scoring time
```

Before this PR both sides used processing time, so they agreed *by construction* — whatever bucket a signal was written into was the bucket the scorer would read. This PR breaks that symmetry and nothing in the diff re-establishes it.

The consequence is not hypothetical at boundaries: a transaction whose `occurredAt` falls just before a bucket edge — which includes every transaction with ordinary producer→consumer latency near the edge — is written into bucket *N-1*, while scoring for a transaction arriving just after reads bucket *N* and does not see it. The direction of the error is **undercount**, which is the direction this PR's own sibling (#5715) argues is the dangerous one for a fraud control: an overcount makes a rule fire more readily, an undercount makes it silently not fire.

I am not saying event-time bucketing is wrong — "transactions that OCCURRED in this window" is the reading that reconciles against the source transactions, and the KDoc argues that well. I am saying the read side needs to be part of the same decision, and right now it is not addressed, not tested, and not mentioned. Please either bucket the read on the same clock basis, or state explicitly why the divergence is acceptable and add a test that pins the boundary behaviour.

### Finding — no migration or backfill for the semantic change to existing rows

Rows already in `velocity_aggregates` were bucketed by processing time. After deploy, new writes are bucketed by event time into the same PK space, so a single row can mix both semantics — and during a rolling deploy old and new pods write both at once. There is no migration and no note about it. Given the bucket is part of the PK this is not correctable after the fact. Worth at least a stated decision (e.g. "windows are short enough that the mixed rows age out within one bucket") in the migration/KDoc, the way #6040 states its residual bound.

### Minor

- `TransactionSignalConsumer.kt` is edited from the same base blob by #5715, #5737 and this PR. #5715 replaces the very `try/catch` blocks this PR edits with `withBoundedRetry`. Whichever lands second, git may merge it textually cleanly and produce something neither author reviewed — please diff the merged result rather than trusting a clean merge.
- The class KDoc this PR leaves in place still says `velocity_aggregates` has no dedup; #6040 makes that false. Same file, adjacent lines — worth reconciling in whichever PR lands last.

PR #6062, opened 2026-08-20
fix(audit): subscribe the audit trail to the ledger, sdd, interest and fraud money-path topics

JiRaska:
## What this does

Wires **four** of the seven money-path produced topics that #6047's gate baselined as `KNOWN_GAPS`
into audit-service's subscription — in **all three** places at once, with the corresponding
baseline entries deleted so a half-fix could not pass:

| topic | producer (module declaring the outgoing channel) | why |
|---|---|---|
| `openbank.ledger.journal.posted` | `openbank-ledger-service` (`ledger-events-out`) | **the posting record itself** — the largest of the five #6035 names |
| `openbank.sdd.event` | `openbank-sdd-service` (`sdd-events-out`) | direct-debit collection lifecycle |
| `openbank.interest.accrual.event` | `openbank-interest-service` (`interest-events-out`) | accrual + withholding remittance |
| `openbank.fraud.hold.changed` | `openbank-fraud-service` (`fraud-outbox-out`) | funds holds |

Three artefacts per topic: `mp.messaging.incoming.audit-events-in.topics`, `TopicAttribution` in
`EventAttribution.kt`, and a Read/Describe ACL on the `audit-service` KafkaUser. Every attribution
value is read off the module that **declares the outgoing channel**, never derived from the topic's
domain segment — `source_service` is chain-hashed into `record_hash` and `audit_entries` is
append-only at the DB (`no_update_audit`/`no_delete_audit` are `DO INSTEAD NOTHING`, so an UPDATE
affects zero rows and *reports success*), so an attribution gets exactly one chance to be right.

Each topic's `KafkaTopic` CR already exists and is owned by its producer; nothing here creates a
topic.

## Left deliberately, with reasons

- **`openbank-billing-service` / `openbank.billing.fee.event`** — there is **no `KafkaTopic` CR for
  it anywhere** under `openbank-infra/gitops`. A Read grant pointing at an undeclared topic is a
  second way to get silence, and who declares it is billing's call, not audit's. Stays baselined.
- **`openbank-standing-order-service` / `openbank.standing-orders.order.event`** and
  **`openbank-psd2-service` / `openbank.psd2.events`** — found by #6047's gate and named in neither
  #5859 nor #6035, so nothing has yet reviewed whether those streams belong in the audit trail.
  Wiring them here would smuggle that decision into a backfill PR. Stays baselined.

Gate baseline: **21 known gaps → 9**. `Refs #6035` rather than `Closes` for exactly that reason.

## Stacking

Branched from **#6047's** `test/audit-subscription-guard-6035` — the gate script whose `KNOWN_GAPS`
entries this PR must delete exists **only** on that branch (`git show origin/main:.github/scripts/
check-audit-money-path-subscription.py` → absent). Then `origin/main` and **#5857's**
`fix/spend-reservation-audit-events` were merged in.

Merging #5857 in was not optional. Its delegation fix edits the **same three files**, and the
`topics:` value is a **single line** — any independent addition to it is a guaranteed textual
conflict no ordering avoids. Both branches are ancestors of this one
(`git merge-base --is-ancestor`, verified for each), so this PR cannot conflict with either by
construction, and nothing of theirs is removed: `git diff origin/fix/spend-reservation-audit-events
HEAD` removes 0 lines from `EventAttribution.kt` and from `kafka-audit-mtls.yaml`, and the final
`topics:` list is 26 entries = the 21 on main + #5857's delegation + these four, counted rather than
eyeballed (a clean merge silently drops neighbouring list additions — the `.release-please-manifest.json`
56→55 shape).

#6047's `IN_FLIGHT` entry for `openbank.delegation.events` is **left alone**: it is #5857/#6047's to
retire, and it is a `::warning`, not an error.

## `auto.offset.reset` — NOT changed, and the reason is not the one in the issue

#6035 warns against forcing `earliest` on this shared group. Reading the deployed config, it is
**already** `earliest` — `openbank-infra/gitops/components/audit/audit-service-msg-override.yaml`
sets `mp.messaging.incoming.audit-events-in.auto.offset.reset=earliest` at `config_ordinal=500`
(a properties file precisely because a dotted YAML leaf key is silently quoted by SmallRye and never
reaches the connector). Nothing is changed here, and the consequence is worth stating plainly: the
existing 22 topics have committed offsets on group `audit-service` and are untouched, but the four
new ones have none, so each is read from its retention start on first subscribe. That is the
intended audit behaviour rather than a mass replay of the existing surface — but it is a
first-subscribe load spike, and it is a *deploy-time* fact this repo cannot test.

## The shared-channel risk, measured rather than assumed

`audit-events-in` declares **no `failure-strategy` and no dead-letter wiring anywhere** (grepped
across `openbank-audit-service/src/main/resources` and `openbank-infra/gitops/components/audit`:
zero hits), so SmallRye's default `fail` applies — and on a shared channel that would mean a poison
record on any one of 26 topics halts all 26.

It cannot fire here, and not by luck: `AuditConsumer.consume(Message<String>)` acks in a `finally`
and the inner `consume(payload, address)` catches `Exception` around the whole body, so the consumer
**never nacks**. The real exposure is the other one — a record that fails to store is logged at
`error` and dropped, and the audit trail is one row short with no nack, no DLQ and no counter for
the drop. That is pre-existing and deliberate (the KDoc argues an audit path that drops is better
than one that stalls), it is not introduced or widened in kind here, and this PR does not change it;
widening the surface by four topics widens its *reach*, which is worth a follow-up decision rather
than a drive-by change to error handling in this PR.

## Proof

**The gate, before → after** (`--enforce`; without the flag it prints `::warning` and exits 0, so
both were run with it — #2392's shape):

```
before: 21 produced topic(s) across 23 money-path service(s) …, 21 known gap(s) — clean.   exit 0
after:  21 produced topic(s) across 23 money-path service(s) …,  9 known gap(s) — clean.   exit 0
```

**Gate negative controls** — each artefact reverted alone, then restored:

| control | result |
|---|---|
| ledger removed from the `topics:` list only | exit 1 — `::error::openbank-ledger-service produces openbank.ledger.journal.posted but it is missing from audit-service's incoming topics list` |
| its `TopicAttribution` entry removed only | exit 1 — `…missing from TopicAttribution in EventAttribution.kt` |
| its Read ACL removed only | exit 1 — `…missing from the audit-service KafkaUser's Read ACLs` |
| baseline entry re-added while the topic is wired | exit 1 — 3 × `::error::stale KNOWN_GAPS entry openbank-ledger-service#openbank.ledger.journal.posted#{topics,attribution,acl}` |
| `--selftest` | `selftest OK: 14 cases, both directions` |

**The test.** `AuditSubscriptionSurfaceIT` (from #6047) reads the topic list from configuration, so
it picks up the four new topics with no edit — and it is exactly there that it has a blind spot: it
is corpus-driven, so **deleting** a topic makes the loop iterate one fewer topic and stay green
(the "a gate whose scope is the thing it checks reads as passing when the list is short" shape).
So a second test names these four as **literals** and asserts each is both subscribed and attributed
to the module that declares its outgoing channel.

**Test negative control, measured.** Reverting only the production wiring for
`openbank.ledger.journal.posted` (topics list + `TopicAttribution`, leaving everything else):

```
java.lang.AssertionError: [issue #6035 wiring]
Expecting empty but was: ["openbank.ledger.journal.posted: absent from
  mp.messaging.incoming.audit-events-in.topics — nothing is read, and no error is raised anywhere"]
```

`tests="2" skipped="0" failures="1"`, BUILD FAILED — **and the corpus-driven test passed in the same
run**, which is the measurement that justifies adding the second test rather than an argument for it.
Restored: `tests="2" skipped="0" failures="0"`.

**What no test in this repo can reach.** There is no Kafka broker in any test here (in-memory
connector fleet-wide), so the in-memory connector cannot see a topic missing from the *deployed*
`topics:` list or a missing Read ACL — the two failures that actually silence a subscription. Those
are the gate's job, and that is why both halves moved together.

## Verified

- `:openbank-audit-service:test --rerun-tasks` → **148 tests, 0 skipped, 0 failures, 0 errors**,
  read from the JUnit XML (a boot failure reports SKIPPED and reads as a pass).
- `:openbank-audit-service:ktlintCheck :openbank-audit-service:detekt
  :openbank-audit-service:quarkusBuild` → BUILD SUCCESSFUL (quarkusBuild included because CDI wiring
  is not validated by ktlint + unit tests).
- `check-threat-model-diff.py --enforce` → exit 0 against **both** `origin/main` and this PR's base.
- `PR_DIFF_BASE=origin/main run-gates.py --all` → 159 gates, 155 PASS. One failure was mine —
  `stale-comment-references` on four abbreviated config paths in a comment — and is fixed in the
  second commit (gate re-run: exit 0). The other three do not reproduce as content failures:
  `api-contract-gate` dies on `sudo: a password is required` fetching `oasdiff` (and this PR touches
  no `openapi.yaml`), `loki-rule-load-test` and `release-scope-mismatch` on `BASE_REF`/`PR_TITLE`
  unbound — all three are CI-environment variables absent from a local shell.

## Not verified

Anything requiring the live broker: that the Read ACL is accepted by Strimzi, that the new
partitions have data, and the first-subscribe read volume. No live system was touched.

`fix(audit)` — audit-service ships a config + attribution change, so it releases, which is correct.

Refs #6035, #5859, #5857, #6047, #5338, ADR-0249 D4.

### DIFF
--- a/.github/scripts/check-audit-money-path-subscription.py
+++ b/.github/scripts/check-audit-money-path-subscription.py
@@ -75,15 +75,17 @@ AUDIT_KAFKAUSER = "audit-service"
 # it. Fixing one means DELETING its entries here — and the check fails on an entry that is no
 # longer a gap, so the list cannot drift in either direction.
 #
-# All six are #6035 / #5859: the same six topics missing from all three places at once. They are
-# not drive-by-fixable from this PR: adding a topic to the audit subscription is a live-broker
-# change (a new Read grant on a running consumer group) whose blast radius wants its own review,
-# and PR #5857 is already in flight for the delegation row.
+# What is left after #6035's first backfill. Four of the original seven -- ledger
+# (openbank.ledger.journal.posted), sdd, interest and fraud -- were wired in all three places and
+# their entries deleted here; a stale entry is itself an error, so a half-fix cannot pass.
+#
+# The three that remain each need a decision this ratchet should not make silently:
+#  - billing has no KafkaTopic CR for openbank.billing.fee.event at all, so a Read grant would
+#    point at an undeclared topic -- a second way to get silence, and the topic's ownership is
+#    billing's call, not audit's.
+#  - standing-order and psd2 were found by THIS check and named in neither #5859 nor #6035, so
+#    nothing has yet reviewed whether their streams belong in the audit trail.
 _GAP_ISSUE = {
-    "openbank-ledger-service": "#6035 - the posting record itself, the largest of the five",
-    "openbank-sdd-service": "#6035",
-    "openbank-interest-service": "#6035",
-    "openbank-fraud-service": "#6035",
     "openbank-billing-service": "#6035 - also has no KafkaTopic CR (see the issue)",
     # Found by THIS check on its first run, and by nothing before it: neither #5859's probe nor
     # #6035's enumeration named these two. That is the argument for deriving the scope rather
@@ -93,10 +95,6 @@ _GAP_ISSUE = {
     "openbank-psd2-service": "#6035 - found by this check, not named in the issue",
 }
 _GAP_TOPICS = {
-    "openbank-ledger-service": "openbank.ledger.journal.posted",
-    "openbank-sdd-service": "openbank.sdd.event",
-    "openbank-interest-service": "openbank.interest.accrual.event",
-    "openbank-fraud-service": "openbank.fraud.hold.changed",
     "openbank-billing-service": "openbank.billing.fee.event",
     "openbank-standing-order-service": "openbank.standing-orders.order.event",
     "openbank-psd2-service": "openbank.psd2.events",
--- a/openbank-admin-ui/src/app/docs/cloud-architecture/page.tsx
+++ b/openbank-admin-ui/src/app/docs/cloud-architecture/page.tsx
@@ -154,7 +154,7 @@ const NAMESPACES: NS[] = [
 
 function StatusDot({ status }: { status: Status }) {
   const m = STATUS_META[status]
-  return <m.Icon size={13} style={{ color: m.color, flexShrink: 0 }} />
+  return <m.Icon aria-hidden="true" size={13} style={{ color: m.color, flexShrink: 0 }} />
 }
 
 type CloudNodeBoxProps = { n: Node; selectedId: string | null; liveStatus: Record<string, InfraStatusResult> | null; t: (cs: string, en: string) => string; onSelect: (node: Node) => void }
@@ -165,7 +165,7 @@ function CloudNodeBox({ n, selectedId, liveStatus, t, onSelect }: CloudNodeBoxPr
   const probeId = INFRA_PROBE_MAP[n.id]
   const probe = probeId && liveStatus ? liveStatus[probeId] : null
   const pm = probe ? PROBE_META[probe.status] : null
-  return <button onClick={() => onSelect(n)} title={t(...n.desc)} style={{ display: 'flex', alignItems: 'center', gap: '6px', padding: '6px 10px', borderRadius: '8px', cursor: 'pointer', background: m.bg, border: `1px solid ${active ? m.color : m.border}`, boxShadow: active ? `0 0 0 2px ${m.color}55` : 'none', color: 'var(--text-primary)', fontSize: '12px', fontWeight: 600, fontFamily: 'inherit', textAlign: 'left' }}>
+  return <button type="button" onClick={() => onSelect(n)} title={t(...n.desc)} aria-label={`${t(...n.name)} — ${t(...m.label)}`} aria-pressed={active} aria-controls={active ? 'cloud-architecture-selection' : undefined} style={{ display: 'flex', alignItems: 'center', gap: '6px', padding: '6px 10px', borderRadius: '8px', cursor: 'pointer', background: m.bg, border: `1px solid ${active ? m.color : m.border}`, boxShadow: active ? `0 0 0 2px ${m.color}55` : 'none', color: 'var(--text-primary)', fontSize: '12px', fontWeight: 600, fontFamily: 'inherit', textAlign: 'left' }}>
     <StatusDot status={n.status} />
     {t(...n.name)}
     {pm && <span style={{ display: 'inline-flex', alignItems: 'center', gap: '2px', marginLeft: '4px', padding: '1px 5px', borderRadius: '10px', background: pm.bg, border: `1px solid ${pm.color}44`, fontSize: '10px', fontWeight: 700, color: pm.color }}><pm.Icon size={9} aria-hidden="true" />{pm.label}</span>}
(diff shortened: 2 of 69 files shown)


JiRaska:

### Stacking and entry count: both verified, both clean

The repo's rule is that a clean merge can silently *delete* a list entry when two branches add neighbouring lines, so I counted rather than read the diff.

**Ancestry** — `git merge-base --is-ancestor`:
- #6047 (`test/audit-subscription-guard-6035`) **is** an ancestor of this branch ✅
- `origin/main` is an ancestor of neither (#6047 is 11 commits behind main; this branch merges main in at `ca8374782`)
- three sides merged: #6047 base, `origin/main`, and `fix/spend-reservation-audit-events` (#5857) at `e51b73333`

**`topics:` entry count** — parsed and set-differenced, not eyeballed:

| side | entries | dups |
|---|---|---|
| `origin/main` | 21 | 0 |
| #6047 | 21 | 0 |
| #5857 (spend-reservation) | 22 | 0 |
| **this branch** | **26** | **0** |

Union of all three input sides = 22. This branch = 26. **Missing vs union: NONE.** Added beyond union: exactly the four intended (`ledger.journal.posted`, `sdd.event`, `interest.accrual.event`, `fraud.hold.changed`). Order is preserved — #5857's 22 are an exact prefix, so `openbank.delegation.events` survived the merge in position. No entry was dropped and none duplicated.

**Ratchet behaviour across the stack is correct**: the gate's baseline goes 21 entries → 9 (3 topics × 3 places), and the `IN_FLIGHT` delegation row is removed now that #5857 has landed. I ran the gate on this branch: `exit=0`, `SUBJECTS=21`, `9 known gap(s) — clean`. Because a stale baseline entry is itself an error, a half-fix could not have passed — I verified that separately on #6047 by fixing one of the three places and watching it go red.

### Producers: verified, and my own suspicion disproved

I checked whether any of the four topics has **more than one** producer — which would make the single-valued `TopicAttribution` row a confident, false `source_service` chain-hashed into `record_hash`. `git grep` initially looked alarming: `openbank.ledger.journal.posted` also appears in balance-service, `openbank.interest.accrual.event` in tax-reporting and billing, `openbank.fraud.hold.changed` in engagement. Reading each in context, **all of those are `incoming:` channels** — consumers, not producers. Every one of the four has exactly one `outgoing` declaration, and each maps to the service named in `EventAttribution.kt`. The attribution table is correct. Flagging the check because the naive grep result looks like a defect and is not.

### `earliest` reasoning: verified, and the load estimate is understated

The claim at `application.yaml:169-174` is accurate. `audit-service-msg-override.yaml` sets `mp.messaging.incoming.audit-events-in.auto.offset.reset=earliest` at `config_ordinal=500` via a **properties file**, correctly sidestepping the SmallRye dotted-YAML-key trap (#686) that would otherwise make the setting inert. The 22 existing topics have committed offsets on group `audit-service` and are untouched; the four new ones have none and will read from retention start. That is right.

What the note undersells: `openbank.ledger.journal.posted` is *the posting record*, i.e. plausibly the highest-volume topic in the estate, and every record becomes one row in `audit_entries`, which is chain-hashed and therefore serialised. "A first-subscribe load spike worth knowing about" is a fair description of the mechanism but not a number. Before this goes to an environment with real ledger volume, it is worth stating the retention window on that topic and the approximate record count, so whoever watches the rollout knows whether to expect minutes or hours of catch-up — and whether consumer lag on the *other 22* topics will be starved behind it, since they share one channel and one consumer.

### MAJOR — this extends a silent, uncounted data-loss path to the ledger posting record

Not introduced here, but this PR is what makes it money-path-critical, and the PR body does not mention it.

`AuditConsumer.kt:180-182`:
```kotlin
} catch (e: Exception) {
    log.errorf(e, "Failed to record audit entry: %s", payload.take(200))
}
```
and `:72-80` acks in a `finally`, unconditionally and by design ("an un-storable message would stall the partition forever").

So: a DB failure on `repo.save(entry)` → the exception is swallowed, **no counter is incremented** (the surrounding code increments `openbank.audit.event.time.missing` and `openbank.audit.attribution.missing`, but the catch itself increments nothing), the offset is committed, and the record is gone. `audit_entries` is append-only, so it cannot be backfilled. The only trace is a log line, and per this repo's own retention notes Loki holds 168h.

That combination — swallowed failure, no series, committed offset, append-only store — is precisely the shape flagged in this repo's guidance ("a swallowed failure leaves NO error to count"). Applying it to `openbank.ledger.journal.posted` while simultaneously replaying the topic from `earliest` is the worst moment for it: the replay is exactly when a sustained store failure is most likely, and it will look identical to success.

Ask: add a counter in that catch, tagged by `source_service` and topic, and an alert on it being non-zero. It is a few lines and it is the difference between this gap being detectable and not. If it belongs in a separate PR, fine — but it should be an issue with a number before this merges, not a sentence in a PR body.

### Scope creep: none, despite the file list

`gh pr view` lists ~70 files, all but five inherited from the merged `main` and #5857. The authored delta after `e51b73333` is:

```
.github/scripts/check-audit-money-path-subscription.py  |  22 +/-
.../audit/application/EventAttribution.kt               |  22 +
openbank-audit-service/src/main/resources/application.yaml | 26 +
.../audit/integration/AuditSubscriptionSurfaceIT.kt     |  47 +
.../gitops/components/audit/kafka-audit-mtls.yaml       |  33 +
```

Five files, all on-topic, and all three places updated consistently — subscription list, `TopicAttribution`, and the KafkaUser Read ACLs. That three-way consistency is the whole point of #6035 and it is met.

### Checked, found nothing
- ACL grants: `[Read, Describe]`, `patternType: literal`, correct topic names, no topic *created* here — Read grants only, as the comment claims.
- No duplicate YAML keys introduced in `application.yaml`.
- `AuditConsumer.consume` is generic and needs no change per topic — confirmed; the four payload shapes resolve `AttributionSource.TOPIC` and none sets `sourceService`, as stated.
- No secrets, tokens or private hostnames in the diff.
- The four new topics do not collide with any existing `topics:` entry.

### One process note
This targets `test/audit-subscription-guard-6035`, not `main`. Do **not** merge #6047 with `--delete-branch` while this is open — deleting the base closes this PR irreversibly (a closed PR whose base is gone cannot be reopened or retargeted). Retarget this to `main` first, or merge #6047 without `--delete-branch`.

Issue #6066, opened 2026-08-20
temporal: the worker<->frontend edge has no transport authentication, and threat models credit mTLS on it


JiRaska:
## What

The Temporal worker↔frontend edge has no transport authentication anywhere in this fleet, and at
least one threat model credited mTLS on it.

`openbank-libs-temporal/src/main/kotlin/com/openbank/libs/temporal/TemporalClientProducer.kt` is the
single fleet-wide `WorkflowClient` producer (ADR-0209 D1, #2572 — it replaced 14 per-service copies).
It builds its stubs as:

```kotlin
val stubsOptions = WorkflowServiceStubsOptions.newBuilder().setTarget(config.serverUrl())
```

plus an optional metrics scope. There is no `setSslContext`, and the string `tls`/`ssl`/`cert` does
not appear in that module at all. The server side matches: `openbank-infra/gitops/components/temporal/temporal-helmrelease.yaml`
configures no TLS for the frontend or for internode. Every consumer's `OPENBANK_TEMPORAL_SERVER_URL`
is a plain `temporal-frontend.temporal.svc.cluster.local:7233`.

So the connection is unauthenticated in both directions. This is a control that was never built, not
one that is misconfigured or was turned off.

## What does constrain the edge today

Worth stating so the residual exposure is judged accurately rather than assumed maximal:

- `temporal-network-policies.yaml` — the `temporal-platform-ingress` NetworkPolicy admits port 7233
  only from an explicit namespace allow-list, and `temporal-intra-namespace` covers the rest. The
  effective authentication on this edge is *network position*, which is the same shape as the
  Kafka-before-mTLS state and is a real control, just a coarse one.
- The Temporal SDK dispatches only the activity and workflow types a worker registered, so an
  unregistered type name is not dispatchable regardless of who is speaking.

## Why it is being filed rather than fixed here

Found while correcting `docs/threat-models/openbank-settlement-service.md` for #6055. S1 of that
document credited two controls for "attacker impersonates Temporal server, injects malicious workflow
tasks": an `OpaActivityInterceptor` (which is in no Kotlin source — that was #6055's subject) and
"mTLS between Temporal server and workers (both directions)". Verifying the first surfaced that the
second is equally absent. #6055 corrects the settlement document's S1 row to say so; it does not
build anything, because this is a fleet-wide platform decision touching the shared producer, the
Temporal HelmRelease, and certificate issuance for every one of the ~14 consumers.

[...]

PR #6067, opened 2026-08-20
docs(settlement): withdraw the OPA activity control S1/E1 credited and never had (#6055)

### LINKED ISSUE #6055: settlement: the OPA activity-authorization control credited by S1/E1 does not exist, and residual risk 2 is marked Closed on it

JiRaska:
## What

`docs/threat-models/openbank-settlement-service.md` credited an OPA activity-authorization control on
the Temporal worker in three places — S1, E1, and residual risk **2**, which was signed off as
**Closed** on it. The control does not exist. This withdraws the claim, re-opens risk 2, and deletes
the orphan policy file that made the claim look substantiated.

Closes #6055. Refs #6037, #6048, #3921, #6066.

## Measured, not read

The three parts of the closed entry were checked independently, and all three were false.

**1. No interceptor.** `OpaActivityInterceptor` occurs in no Kotlin source anywhere in the repo — its
only appearances are that threat model and four label strings on an admin-UI page.
`openbank-libs-temporal` contains `TemporalClientProducer` and `TemporalConfig` and nothing else; the
"OPA activity interceptor" in ADR-0101's P0 row was never built. `SettlementWorkerRegistrar` registers
workflow and activity implementations and installs no interceptor.

**2. Not bundled — established from outside the code.** Reading rego cannot tell you which of several
reasons fires, so the bundle ConfigMap was materialised into the OPA sidecar's own directory layout
(`rules-data.yaml` → `rules/data.yaml`, `agents-data.yaml` → `agents/data.yaml`) and evaluated with
`opa eval`, with a must-ALLOW and a must-DENY control in the same run so an empty result could be
distinguished from a dead probe:

| probe | input | result |
|---|---|---|
| **control, must ALLOW** | `settlement.create`, HUMAN `ROLE_OPERATOR` | `{"allow":true, "reason":"operator-settlement-write"}` |
| **control, must DENY** | `settlement.create`, ANONYMOUS | `false` |
| **subject** | `data.openbank.settlement` | `{}` |

The package is not in the deployed bundle at all. The three `package` declarations the bundle actually
carries are `openbank.rest` (twice — `rest.rego` and `settlement_rest_ext.rego`, merged by OPA) and
`openbank.agents`. `gen-settlement-opa-bundle.sh` never reads the file.

**3. Not queried, and it did not say what E1 said.** The service's only OPA query path is
`/v1/data/openbank/rest/allow` (`application.yaml`), which cannot reach an activity package. And
evaluating the file directly — the hypothetical where something *did* load it:

```
{"activity":"reverseBookToLedger","compensation":false}  -> allow = true
{"activity":"reverseBookToLedger"}                       -> allow = true
{"activity":"dropAllTables"}                             -> allow = false
```

E1 claimed "`reverseBookToLedger` requires `compensation=true` context". The word `compensation`
appeared **zero** times in the file. It could not have enforced E1 even if it had been loaded.

## The judgement: the control is not being built, and here is why

This is the part of the issue that mattered more than the edit, so the reasoning is in the document
itself rather than only here.

**The gate as specified would add no security property this service does not already have.** The
Temporal SDK dispatches only the activity types a worker registered, and the policy's allowed set was
exactly the seven methods `SettlementActivitiesImpl` registers — so the membership test restated the
registration and could never deny anything reachable. It was a no-op with the shape of a control.

**The threat S1 actually describes is a transport-authentication problem, and no name-based policy
addresses it.** A task forged by a spoofed or compromised Temporal frontend carries a *legitimate*
activity name; a membership test on that name cannot distinguish it from a real one.

**The one genuinely security-relevant clause in E1 belongs in the saga, not in a policy.** "A
compensation may run only after its forward leg" is a statement about workflow history. An external,
stateless policy cannot see that history. `SettlementWorkflowImpl` already encodes it by registering
each compensation after the corresponding forward activity returns, and #6048 additionally makes
`reverseBookToLedger` throw a non-retryable failure rather than report a reversal it did not perform.

**The file was an orphan by convention as well as by wiring.**
`openbank-settlement-service/src/main/resources/opa/settlement_activity.rego` was the only
`src/main/resources/opa/` file in the entire fleet; every other service's policy lives in its bundle
generator. It is deleted rather than left in place, because an unloaded policy file next to a threat
model that cites it reads as coverage — which is exactly how this survived.

## Risk 2 is re-opened, not re-closed

Nothing was built here, so nothing is closed. Risk 2 now states plainly that no activity-level
authorization exists on the worker, records all three measurements above so the next reader does not
have to re-derive them, and stays **OPEN** pending the S1 transport question. The document is not
quietly rewritten to match the absent code.

## S1's *other* clause was also fiction — filed, not fixed

S1 credited two controls. Verifying the first surfaced that the second — "mTLS between Temporal server
and workers (both directions)" — is equally absent: `TemporalClientProducer`, the single fleet-wide
producer since ADR-0209 D1, builds `WorkflowServiceStubsOptions` with a target and no SSL context, and
the Temporal HelmRelease configures no TLS either. S1's row is corrected to say so and to name what
*does* constrain the edge (the `temporal-platform-ingress` NetworkPolicy allow-list on 7233, plus
registration-scoped dispatch).

Building it is a fleet-wide platform decision touching the shared producer, the HelmRelease and
certificate issuance for ~14 consumers — so it is **filed as #6066**, not smuggled into this PR. #6066
also carries the two other artifacts that still advertise the non-existent interceptor
(`openbank-admin-ui/src/app/temporal/page.tsx`, and ADR-0101's P0 delivery status), which were left out
of here deliberately: see the next section.

## Disjointness from the live parallel settlement work

Compared by content, not by the merge-base file list.

| PR | branch | overlap with this PR |
|---|---|---|
| **#6048** reversal wiring | `fix/settlement-reversal-wiring-6037` | **same file**, disjoint rows — see below |
| **#6036** stranded-settlement alert | `feat/settlement-metrics-5705` | none — no file in common |
| **#5723** domain metrics | `feat/settlement-service-metrics` | none — no file in common |

#6048 is the only one that touches the threat model. It rewrites **T1**, **R2**, inserts a correction
*inside* residual risk 3, appends new residual risks, and adds a change-log entry. This PR touches
**S1**, **E1**, residual risk **2**, the ADR-0034 reference line, and a change-log entry. No row is
edited by both.

A trial merge of #6048's head into this branch was run and aborted: it produces **exactly one
conflict**, in the change log, where both PRs add a `2026-08-20` entry at the same anchor — additive on
both sides, resolved by keeping both. The residual-risks region merges clean. Two things were changed
here specifically to get that result:

- Residual risk 3's line `` `settlement_activity.rego` present (risk 2) `` is *also* stale — it records
  the file's presence as a verified go-live pre-condition. It is deliberately **left untouched**,
  because #6048 inserts text immediately above it and editing it would have turned a clean region into
  a second conflict. Re-opened risk 2 states instead that the file's presence establishes nothing, so
  the correction is present without the collision.
- This PR cites #6037 by issue number rather than by residual-risk index, since #6048 renumbers that
  list (and currently introduces two risk 5s on its own branch).

Either order of merge works.

## Also corrected

`docs/runbooks/0006-temporal-settlement-go-live.md` Gate 2 step 1 instructed an operator to "confirm
`settlement_activity.rego` is returning `allow` for the 7 activity methods in advisory logs" — an
unperformable step, since nothing ever loaded or queried it. It now points at the gate that is real
(`settlement.create` against `data.openbank.rest.allow`) and states that Temporal activities are not
evaluated by OPA, so flipping `AUTHZ_ENFORCE` does not change how they are dispatched.

## Release scope

`docs(settlement)` — a hidden type, so no release is cut, which is the honest classification: the
deleted resource was loaded by nothing, so no shipped behaviour changes. `version.txt`,
`CHANGELOG.md` and `openapi.yaml` are untouched.

## Verification

Run in the worktree against `PR_DIFF_BASE=origin/main`, foreground, exit codes read from `$?` rather
than from a pipeline:

| group | result |
|---|---|
| `security` | **PASS 14/14** |
| `registry` | **PASS 26/26** |
| `kotlin` | **PASS 19/19** |
| `data` | **PASS 16/16** |
| `lint` | **PASS 20/20** |
| `supplychain` | PASS 22, WARN 1 |
| `gitops` | PASS 32, UNFALSIFIED 1 |
| `api` | PASS 6, FAIL 1 |
| `openbank-infra/scripts/check-threat-model-diff.py` | **rc=0** |

The two non-green results are local-environment artifacts, both re-checked: `gitops`'s UNFALSIFIED is
the Loki rule-load gate failing its own self-test on `BASE_REF: unbound variable` (CI supplies it), and
`api`'s FAIL is the oasdiff gate unable to `sudo`-install its binary locally — this PR changes no
`openapi.yaml`, and that gate's own 12 self-test checks passed. Neither is caused by this change.

## What was NOT verified

- Nothing was checked against the live cluster; the `opa eval` results are against the bundle
  **ConfigMap as committed**, not against a running sidecar's loaded state. Those agree only if the
  ConfigMap is synced, which was not verified here.
- The claim that no *other* threat model credits a Temporal-edge control that does not exist was not
  swept; it is part of #6066.


JiRaska:
## Review — checked the honesty question specifically; it withdraws the claim without quietly re-closing the risk. **Ready.**

**Staleness:** 4 commits behind `main`, `MERGEABLE`, no competing edit to any of the three files. Live.

**I independently verified each factual claim rather than taking the PR body's word for it:**
- `git grep OpaActivityInterceptor origin/main -- '*.kt'` → **no matches**. The interceptor the threat model credited exists in no Kotlin source.
- `settlement-opa-bundle.yaml` on `main` contains **zero** occurrences of `settlement_activity`. The policy is in no bundle.
- The deleted `openbank-settlement-service/src/main/resources/opa/settlement_activity.rego` is 19 lines whose `allow` is a single membership test over a seven-element set with `reverseBookToLedger` an unconditional member and no `compensation` term anywhere — so it could not have enforced E1 even had something loaded it, exactly as the PR says.
- `#6055` is OPEN, `#6066` is OPEN and titled for the transport-authentication half. The follow-up is a real filing, not a promise in a PR body.

**The thing I was checking for, and did not find: a quiet re-close.** Residual risk 2 is rewritten from *"REOPENED"* to **"OPEN — ..."** and ends `It is deliberately not re-closed by this change: nothing was built here, only a false claim withdrawn.` S1 and E1 are corrected to name the residual constraints as **structural** (only registered activity types are dispatchable; `reverseBookToLedger` throws `LEDGER_REVERSAL_UNSUPPORTED`; the NetworkPolicy namespace allow-list) and explicitly *not* as authorization. That is the right distinction — a structural constraint and a control are not interchangeable, and conflating them is how the original row got written.

**Two things I particularly want to endorse, because they are the parts that usually get dropped:**
1. **The measurement is stated with its own controls.** "Materialise the ConfigMap into the sidecar layout, `data.openbank.settlement` resolves to `{}`" is paired with a must-ALLOW (`settlement.create` / HUMAN `ROLE_OPERATOR` → `operator-settlement-write`) and a must-DENY (ANONYMOUS) in the same run. Without those, an empty result is indistinguishable from a broken query.
2. **Verifying the row surfaced a second, larger defect and it was filed rather than folded in.** The mTLS clause of S1 was also false — `TemporalClientProducer` builds `WorkflowServiceStubsOptions` with a target and no SSL context, fleet-wide — and that became #6066 instead of being silently corrected inside a settlement PR.

**Deleting the orphan `.rego` is right, not merely tidy.** A policy file in `src/main/resources/opa/` that no generator reads is a file a future reader will cite as coverage, exactly as this threat model did. The PR notes it was the fleet's only such file, so no convention is being broken.

**Release mechanics:** `docs(settlement)` is a hidden type, so no release is cut despite the deleted file living under `openbank-settlement-service/src/main/resources/`. That is the correct pairing for a withdrawal of a claim.

**Not checked:** the runbook's Gate 2 wording against how `AUTHZ_ENFORCE` is actually consumed at runtime; I read the change as consistent but did not evaluate the live flag.

Verdict: **ready.** Not approving — review comment only, per my constraints.

JiRaska:
## `build (openbank-settlement-service)` is a main-side break, not this PR

Diagnosed the failing job (run 32504884810, job 96846163767). The Gradle failure is:

```
SettlementStrandedGaugeTest > terminal states are not published and every compensation state is() FAILED
```

Reproduced it as a **negative control on a clean `origin/main` worktree at `8b1868746`**, with none of this PR's changes applied:

```
$ ./gradlew :openbank-settlement-service:test --tests '*SettlementStrandedGaugeTest*'
16 tests completed, 1 failed

org.opentest4j.AssertionFailedError:
expected:
  ["PENDING","DEBITED","CREDITED","REVERSED","CREDITED_REVERSED",
   "REVERSAL_FAILED","LEDGER_REVERSAL_UNSUPPORTED","LEDGER_REVERSED"]
 but was:
  ["CREDITED_REVERSED","PENDING","DEBITED","LEDGER_REVERSED","CREDITED","REVERSED"]
```

Cause: `SettlementStatus` gained `REVERSAL_FAILED` and `LEDGER_REVERSAL_UNSUPPORTED` (#6037), and `SettlementStrandedGauge.NON_TERMINAL` was never extended. The test derives its expected set from the enum (`entries - {BOOKED, REJECTED}`), which is exactly what makes it catch this — so it went red on main the moment the statuses landed.

This PR's own diff is three files (a runbook, a threat model, and a 19-line deletion from `settlement_activity.rego`) and touches nothing on that path. Rebasing does not help, because the break is on `main`.

The fix is already in flight as #6279 (`fix(settlement): watch REVERSAL_FAILED and LEDGER_REVERSAL_UNSUPPORTED in the stranded gauge`), so I have deliberately **not** duplicated it here. Once #6279 lands, rerun this PR's build and it should go green with no change to the payload.

Worth noting what the break itself means: the two states that went unwatched are the compensation-*failure* states — the ones where the money is still moved — so the stranded gauge was blind to precisely the cases it exists for.

JiRaska:
### `build (openbank-settlement-service)` here is **not** caused by this PR's diff

This PR is docs-only, but its build fails on:

```
SettlementStrandedGaugeTest > terminal states are not published and every compensation state is() FAILED
    org.opentest4j.AssertionFailedError at SettlementStrandedGaugeTest.kt:131
55 tests completed, 1 failed
```

Root cause is on `main`, not on this branch. `SettlementStatus` on `origin/main` contains
`REVERSAL_FAILED` (`openbank-settlement-service/.../domain/model/Settlement.kt:41`), while the test's
closing assertion derives the expected set from the enum:

```kotlin
assertThat(published).isEqualTo(
    (SettlementStatus.entries.map { it.name } - setOf("BOOKED", "REJECTED")).toSet(),
)
```

The literal `containsExactlyInAnyOrder` list above it does **not** include `REVERSAL_FAILED`, so the
two assertions disagree the moment the enum gained that value. Verified the test file is byte-identical
between this branch's head and `origin/main` — nothing here changed it.

**Merging `main` in will not clear this.** It is fixed by #6279 (*watch REVERSAL_FAILED and
LEDGER_REVERSED*) and #6284 (*publish stranded-age gauges for every status*). This PR should wait on
one of those landing, then re-run.

JiRaska:
This went `DIRTY` on #6484 (`2da87cf6f`). I looked at whether I could replant it mechanically and
concluded I should not — the conflict is about facts, not text, and this document's only value is
that its claims are true.

**What moved underneath it.** #6410/#6484 removed `reverseBookToLedger` as an unreachable activity,
on the grounds that nothing can fail after `bookToLedger`. This branch's rewritten E1 row still
describes it as present-tense machinery:

> `reverseBookToLedger` is not implemented — it throws a non-retryable failure and records
> `LEDGER_REVERSAL_UNSUPPORTED` (#6037), so it moves no money whoever invokes it.

This branch mentions `LEDGER_REVERSAL_UNSUPPORTED` seven times; `main`'s current version mentions it
once, in the past tense. Rebasing without rereading would land a threat model that credits a
mitigation whose mechanism was deleted — the same defect class this PR exists to correct, which is
why I did not want to do it blind.

**Two conflicts, different characters:**

1. `docs/threat-models/openbank-settlement-service.md` — a real content merge. Both sides rewrote
   E1 and residual risk 2 from the same finding (#6055) and reached *different* text. This branch's
   version is the stronger one: it also establishes that **the mTLS half of S1 does not exist**
   (`TemporalClientProducer` builds `WorkflowServiceStubsOptions` with a target and no SSL context,
   and the HelmRelease configures no TLS), which `main` still credits as a control. That finding is
   not on `main` in any form and should not be lost.
2. `openbank-settlement-service/src/main/resources/opa/settlement_activity.rego` — modify/delete.
   This branch deletes it; #6484 edited it. **The deletion is right and the edit was moot**: your
   own `opa eval` evidence shows the package is not in the deployed bundle
   (`data.openbank.settlement` → `{}`), no ConfigMap references the file, and the service only ever
   queries `/v1/data/openbank/rest/allow`. #6484 removed `"reverseBookToLedger"` from an allow-list
   in a file nothing loads, so resolving as *delete* discards nothing that was in effect.

**Suggested order:** rebase, take `delete` for the `.rego`, then reconcile the threat model by hand —
keeping this branch's S1/mTLS finding and residual-risk-2 reopening, and adopting `main`'s
present-tense description of the ledger leg (nothing follows `bookToLedger`, so there is no
compensation to gate rather than an unimplemented one).

Worth noting for its own sake: #6484 polishing an allow-list in an unbundled policy file is a small
instance of what this PR documents. The file reads as a live control, so it attracts maintenance.
That is an argument for deleting it rather than leaving it accurate-but-inert.

JiRaska:
**Correcting my comment above — the premise it rested on is being reverted.**

I told you to drop the `reverseBookToLedger` work on the grounds that #6484 had removed the activity
as unreachable. #6516 (OPEN, CLEAN) reverts #6484, and its reasoning is better than the one I
relayed:

> `bookToLedger` is **not atomic**. `ledgerPort.book()` posts the journal, then `updateStatus(BOOKED)`
> and `audit(...)` can each fail — after the general ledger has been written. The saga unwinds both
> balance movements and rejects, and the journal stands with nothing in the row, the alerts or the
> audit trail naming it.

So the compensation was dead because of **the order it was registered in** — after its forward
activity returned — not because the saga had no room for it. #6484 read "unreachable" off the
registration order and attributed it to the saga's shape; #6481 fixes the ordering properly.

What that changes for this PR: **do not treat the ledger-reversal path as gone.** Wait for #6516 to
land before rebasing — it restores exactly the state this branch was written against, which turns
the rebase back into a clean one instead of the reconciliation I described.

What still holds from my comment: read the state off live `main` at the time you rebase, not off
this thread. Three variants of this code have now been in flight at once (#6484 merged, #6516
reverting it, #6481 fixing it properly), so anything I write about the ledger leg has a short shelf
life — including this correction.

My original comment was accurate about what `main` held when I wrote it and wrong about what that
meant. I passed on #6484's conclusion without checking whether `bookToLedger` was atomic, which is
the one thing the conclusion depended on and is visible in three lines of the method.

Issue #1914, opened 2026-07-23
Deploy Istio STRICT mTLS for real — istio.yaml is referenced by no ArgoCD application


JiRaska:
The 2026-07-16 platform audit (`docs/audits/2026-07-16-platform-audit.md` §4.3) found that `openbank-infra/k8s/base/istio.yaml` declares STRICT `PeerAuthentication` but is **not referenced by any ArgoCD application** — mesh-wide mTLS is aspirational, and east-west HTTP relies on L3/L4 NetworkPolicies alone.

Scope:
- Wire the Istio STRICT mTLS manifests into the ArgoCD app-of-apps so they actually deploy
- Roll out namespace-by-namespace (PERMISSIVE -> STRICT) to avoid breaking non-mesh traffic (CNPG operator, jobs)
- Verify with `istioctl authn` / peer-auth telemetry that money-path east-west hops are actually mTLS
- Update the threat models that currently claim mesh mTLS as a mitigating control

Related: ADR-0137 is Kafka-scoped only; a fleet-wide east-west zero-trust decision ADR is tracked in the position-ADR batch issue.

JiRaska:
First triage pass — **confirmed still valid**, and the finding is if anything sharper than the audit stated.

- `openbank-infra/k8s/base/istio.yaml` is the **only** Istio file in the whole tree.
- `git grep -l istio openbank-infra/gitops/apps openbank-infra/gitops/components` → **zero hits**. Not only is the manifest unreferenced by any ArgoCD Application, there is no Istio control plane in gitops at all — so this is not "wire up an existing mesh", it is "install a mesh".

That reframes the scope materially: step 0 is deciding whether Istio is actually wanted, or whether the existing L3/L4 NetworkPolicy posture plus Kafka mTLS (ADR-0137) is the intended end state and `k8s/base/istio.yaml` should be **deleted** as an aspirational artifact. A manifest that declares STRICT `PeerAuthentication` while deploying nowhere is the false-control-claim pattern this repo has burned itself on before (`model_gateway` → `model_gateway_target`, #1280).

**Owner-gated** — mesh adoption is an architecture decision needing an ADR, not an engineering ticket. Whichever way it goes, the threat models that cite mesh mTLS as a mitigating control are currently **wrong today** and worth correcting independently of the decision.

_Automated triage pass, verified against `origin/main` 2026-07-26._

JiRaska:
[...]

PR #6153, opened 2026-08-21
feat(admin-ui): trace the BFF, the one service in the fleet with no spans

JiRaska:
## What

admin-ui was **the one service in the fleet with no spans**. Measured 2026-08-21: **zero** Prometheus scrape targets in the `admin-ui` namespace and **zero** spans in Tempo, while all 39 backend services were traced.

Its only signals were *"an exception was thrown"* (GlitchTip) and *"the pod answers"* (blackbox). Nothing in between.

That gap sits exactly where operator-visible failures originate. The console is a BFF: ~40 App-Router route handlers proxy operator actions to backends, including `/api/approvals/pending` — a money-path four-eyes queue. When one failed, the backend's own trace existed and **nothing connected it to the request the human actually made**.

## This is not browser RUM — ADR-0088 D4 is not reopened

That decision rejected Grafana Faro for this surface — *"web/JS RUM on an internal operator console (~2 users) is near-zero value"* — and it **stands**; I agree with it. Nothing here runs in a browser.

Server-side tracing of a BFF is a different thing: it is the same class as every other service in the fleet, and it needs none of the hardened public OTLP ingest, CORS or consent machinery the ADR correctly identifies as the expensive part of browser RUM. This is server-to-collector, in-cluster.

## How it works

- **OpenTelemetry NodeSDK** → OTLP/HTTP → `tempo.observability.svc:4318`.
- **Undici instrumentation is the load-bearing part**: Next.js 16 issues outbound BFF calls through undici's fetch, and that is what injects W3C `traceparent`. Without it the console would emit *orphan* traces instead of joining the backend's — which is the entire value.
- **Metrics come free**: spans reach Tempo's metrics-generator, so `traces_spanmetrics_calls_total{service="openbank-admin-ui"}` starts existing as a side effect. The metrics gap closes without a second change or a `/metrics` endpoint.
- Tempo directly rather than the otel-collector, matching what the backends already do (`QUARKUS_OTEL_EXPORTER_OTLP_ENDPOINT: tempo:4317`) — that is where the span-metrics generator lives.

## Gating and PII, following the existing patterns

**Gated** like GlitchTip (ADR-0075 inv. 2): no `OTEL_EXPORTER_OTLP_ENDPOINT` → no SDK is constructed at all, so `next dev` and every test export nothing. `startTracing()` returns a boolean rather than nothing, so *"deliberately off"* and *"on but exporting nowhere"* stay distinguishable — the shape this repo keeps meeting where a disabled adapter reports success.

**PII** (ADR-0075 inv. 3): request URLs on a bank operator console carry bearer tokens, customer ids and account numbers. The query string is stripped from `url.full` **and** `http.url` before export, for the same reason `glitchtip.ts` strips it from every event. The path is kept — knowing *which* route failed is the point.

Vanilla CNCF/Apache-2.0 packages rather than `@vercel/otel`, matching ADR-0088's stated licensing posture.

## Verified

- **4 new tests, and falsified**: dropping `http.url` from the scrub list — so the legacy attribute keeps its token while the span still exports cleanly — turns the PII test **red**; restoring it green. That is the failure mode that would otherwise ship silently.
- Full admin-ui suite **1620/1620** via `npm test`. One earlier run showed `regulatory-report-preview` timing out at 5s; it passes in isolation **both with and without** this change and passed on a clean full re-run, so it is load flake, not a regression — checked rather than assumed.
- `tsc --noEmit` clean (0 errors, after `npm run pretest` bakes the generated types), `eslint` clean, `next build` compiles.
- `kubectl apply --dry-run=client` against the live API server for the gitops manifest. admin-ui's NetworkPolicy declares only `policyTypes: [Ingress]`, so egress to observability needs no change.

## After merge

Worth confirming by effect rather than by the deploy going green: `traces_spanmetrics_calls_total{service="openbank-admin-ui"}` should start existing, and an operator action should produce one trace id spanning console → backend.

Refs #5735

### DIFF
--- a/.github/asvs-l3-baseline.txt
+++ b/.github/asvs-l3-baseline.txt
@@ -277,3 +277,4 @@ V9.1 openbank-infra/gitops/components/payments/payments-services.yaml:832: plain
 V9.1 openbank-infra/gitops/components/payments/payments-services.yaml:1160: plaintext http:// env value http://fraud-service.fraud.svc:8133
 V9.1 openbank-infra/gitops/components/fx-service/fx-service.yaml:89: plaintext http:// env value http://fraud-service.fraud.svc:8133
 V9.1 openbank-infra/gitops/components/ai-platform/litellm.yaml:153: plaintext http:// env value http://langfuse.ai-platform.svc:3000
+V9.1 openbank-infra/gitops/components/admin-ui/admin-ui.yaml:98: plaintext http:// env value http://tempo.observability.svc:4318
--- a/openbank-admin-ui/package-lock.json
+++ b/openbank-admin-ui/package-lock.json
@@ -1,14 +1,19 @@
 {
   "name": "openbank-admin-ui",
-  "version": "0.155.4",
+  "version": "0.183.5",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "openbank-admin-ui",
-      "version": "0.155.4",
+      "version": "0.183.5",
       "dependencies": {
         "@hookform/resolvers": "5.8.0",
+        "@opentelemetry/exporter-trace-otlp-proto": "^0.221.0",
+        "@opentelemetry/instrumentation-undici": "^0.31.0",
+        "@opentelemetry/resources": "^2.10.0",
+        "@opentelemetry/sdk-node": "^0.221.0",
+        "@opentelemetry/semantic-conventions": "^1.43.0",
         "@radix-ui/react-dialog": "1.1.23",
         "@radix-ui/react-dropdown-menu": "2.1.24",
         "@radix-ui/react-label": "2.1.15",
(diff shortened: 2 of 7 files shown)


JiRaska:
CI caught this correctly and I have declared it rather than exempted it.

`V9.1 no new plaintext inter-service edges` flagged `http://tempo.observability.svc:4318` as **1 NEW**. It is new — and it is the **46th instance** of an edge the estate has already accepted: 45 baseline lines are the same Tempo export, one per service that emits traces, including every money-path one.

There is no TLS endpoint to reach instead. In-cluster mTLS is the Istio work referenced by no ArgoCD application yet (#1914), which is exactly the backlog the gate's own `--baseline` help text names: *"real backlog (in-cluster plaintext until mTLS)"*. When that lands, all 46 lines pay off together.

**Deliberately NOT added to `PLAINTEXT_EXEMPTIONS`.** That list holds `localhost`, `127.0.0.1` and `keycloak.iam.svc` — things exempt by nature. Putting Tempo there would silently exempt all 45 existing lines too and *delete* the backlog instead of declaring it, which is the opposite of what a ratchet is for.

Verified locally: with the line added the gate reports 238 findings, **0 NEW, 0 baseline lines paid-off**, exit 0. The diff is one insertion.

One detail worth recording: inserting ~20 lines of env into `admin-ui.yaml` moved **nine** other baselined anchors, and none went stale — the comparison normalises the `:NN:` fragment away (`norm()` in `check-asvs-l3.py`). A strictly line-anchored baseline would have turned an unrelated env addition into nine phantom violations and nine phantom pay-offs.

Issue #2320, opened 2026-07-25
SwiftBootSmokeIT is excluded from every CI job — the service's only real boot test has never gated a merge


JiRaska:
## What never runs

`openbank-swift-service/build.gradle.kts` excludes `SwiftBootSmokeIT` from every CI job, twice over:

```kotlin
if (System.getenv("CI") == "true") {
    filter { excludeTestsMatching("*IT") }
}
...
if (System.getenv("CI") == "true") {
    exclude("**/SwiftBootSmokeIT*")
}
```

So the one test that proves this service *boots* — real Quarkus start, Flyway migration, Testcontainers Postgres + Valkey — has never gated a merge. It runs only on a developer machine, where `CI` is unset, which means in practice only when someone happens to run the full module locally. Issue #578 added it precisely so a real boot would run in CI.

**Status (2026-07-25): this exclusion is still in place on `origin/main`, unchanged.** The rest of this issue's scope (items 1-3 below) is still open.

## Why it was excluded

`SwiftBootSmokeIT` is a `@QuarkusTest`, and Quarkus's `BeforeAllCallback` fires **before** JUnit 5 evaluates `@DisabledIfEnvironmentVariable` (quarkusio/quarkus#21555). Quarkus ignores the JUnit condition and boots anyway, then hangs 37+ min on the runner pool until the job times out — stalling the whole fleet Services CI run. The Gradle-level exclusion stops JUnit from discovering the class at all, so Quarkus never boots. It was a stop-gap for a runner-capacity symptom, not a decision that the service need not be boot-tested.

## Wrong issue reference to correct

~~The comments say "Re-enable per #2404". `#2404` is **not a real issue or PR number in this repo** (highest issue 2280, highest PR 2306 at the time of writing) — it also appears in `openbank-swift-service/CHANGELOG.md` (generated, leave it) and, until #2318, in ADR-0063 and `pact-drift-check.yml`. Three comment sites in the build file cite it; the pact-specific one is now #2319, and the `SwiftBootSmokeIT` ones should point here.~~ **DONE** — `openbank-swift-service/build.gradle.kts` now reads "Re-enable per #2320" at the `SwiftBootSmokeIT` sites; the pact-specific one points at #2319 as planned. Verified against `origin/main` 2026-07-25.

[...]

PR #6182, opened 2026-08-21
ci(transaction): fail fast when the test Kafka broker is gone, instead of burning the 45-minute job timeout

### LINKED ISSUE #5940: bug(ci): transaction-service build burns its 45-minute timeout in a Kafka producer retry loop after Redpanda dies mid-run
## Symptom

`build (openbank-transaction-service)` on `_service-ci.yml` is **cancelled at the 45-minute job timeout** (`timeout-minutes: 45`), which fails `all-green` and makes the PR unmergeable. It reads as "slow CI"; it is not. The tests finish in ~90 seconds and the job then sits idle for ~37 minutes inside a Kafka producer retry loop.

Currently blocking **PR #5043** (money-path four-eyes fix for #4754).

## Evidence

Job `96416555756` (head `f77c32458`), `Build + test` step ran `12:00:36Z → 12:43:38Z`:

```
12:04:37Z  > Task :openbank-transaction-service:test          <- tests start
12:05:11Z  [Producer clientId=kafka-producer-transaction-events-out]
             Connection to node -1 (localhost/127.0.0.1:32771)
             could not be established. Node may not be available.
12:06:01Z  MergeSweepApprovalBindingIT > the approval still satisfies
             the exact sweep it was issued for() FAILED
             java.net.SocketTimeoutException
12:06:09Z  Transaction d4eb65fe-... is COMPLETED; refusing to mark it FAILED
             ... then nothing but producer retries ...
12:43:38Z  ##[error]The operation was canceled.
```

- **247** `could not be established` lines, `12:05:11Z → 12:43:35Z`.
- Only **one** broker port ever appears (`32771`), so this is not a stale producer left over from a `@TestProfile` restart pointing at a previous container.
- `rp.start()` succeeded — no `TestAbortedException` from `PostgresRedpandaTestResource` — so **Redpanda started and then became unreachable**, rather than never starting.
- At job teardown the runner reports `Terminate orphan process: pid (…) (java)` seven times.

## Not a regression, and not specific to one commit

Same signature on the previous head `5f6dd8dc7` (job `96172599762`): **235** occurrences, `18:27:03Z → 19:03:25Z`, same cancellation at 45 minutes. It predates the commits currently on that branch.

## Why only one test surfaces it

Most transaction-service ITs never publish. `MergeSweepApprovalBindingIT` step 4 is the only step that gets past the four-eyes gate and actually executes the sweep, driving the full saga into the Kafka producer — so it is the one test that needs a live broker. The broker was already unreachable at `12:05:11Z`, before that step ran; the other tests simply did not care.

That makes this a **latent** fault: it will surface for any future transaction-service IT that publishes, and it costs a full 45-minute runner slot every time.

## Reproduces only on CI

A local run of the same test against real Testcontainers (Postgres + Redpanda + Valkey) is green — `tests="4" skipped="0" failures="0" errors="0"`. `PostgresRedpandaTestResource` looks correct on inspection; it maps `rp.bootstrapServers` straight into `kafka.bootstrap.servers`. So a green local run is **not** evidence about this failure mode, which is worth stating explicitly because that is exactly how it got missed.

## Candidate causes (hypotheses, not established)

1. The Redpanda container is killed mid-run on `ubuntu-latest` — Postgres + Redpanda + Valkey + a Quarkus JVM per `@QuarkusTest` restart is memory-heavy, and an OOM-killed container would produce precisely "started, then connection refused".
2. Container reaped between `@TestProfile` restarts while the producer from the prior boot survives. Weakened by the single-port evidence above, but not fully excluded.

## Suggested fixes (independent of root cause)

- Bound the damage regardless: set a Kafka producer `max.block.ms` / delivery timeout in the test profile so an unreachable broker fails fast instead of retrying into the job timeout. A 37-minute idle hang should not be reachable from a test failure.
- Log Redpanda container health/exit status on IT failure so "started then died" is visible without reading raw producer retries.
- Then address the underlying container death.

## What would close this

`build (openbank-transaction-service)` completes within its timeout on a PR, with either a green `MergeSweepApprovalBindingIT` or an explicit fast failure — and no multi-minute producer retry tail in the log.

Refs #5043, #4754, #5631, #5636

JiRaska:
## Root cause

`build (openbank-transaction-service)` was cancelled at the 45-minute job timeout while doing
nothing. The tests were not slow — they had stopped.

From job `96416555756` (head `f77c32458`):

- last real test activity: `12:09:46Z` (`Failed to deserialise SchemeAcceptedEvent, dropping: not-json`)
- from there to cancellation at `12:43:38Z`: **247** lines of nothing but
  `[Producer clientId=kafka-producer-transaction-events-out] Connection to node -1 (localhost/127.0.0.1:32771) could not be established`
- the decisive line at the end:
  `No files were found with the provided path: openbank-transaction-service/build/test-results/`

That last one is what makes this worth fixing properly. The `test` task never completed, so **no
JUnit XML was written at all** — the job burned a full runner slot and produced nothing to diagnose
from. Not a boot failure reporting as SKIPPED; there was no report to read either way.

### Why the broker went away

The issue text hypothesised an OOM-killed container, and treated "only one broker port ever appears"
as evidence *against* a container being reaped between test-resource restarts. The log says the
opposite. There are exactly **two** `Ryuk has been disabled` banners — one at `12:04:33Z`, one at
`12:05:12Z` — i.e. two container sets. The producer retries begin at `12:05:11Z`, one second before
the second set starts.

So Redpanda did not die mysteriously. Quarkus restarts the application whenever the set of
`@QuarkusTestResource` classes changes between test classes (this module has classes declaring
`PostgresRedpandaTestResource` alone and classes declaring it *plus* `LedgerWireMockResource`), and
that restart calls `stop()` on the resource, which stops the Redpanda container — while a Kafka
producer belonging to the *previous* boot is still open. The single-port evidence is consistent with
this rather than against it: only the orphaned producer logs failures, because the producers on the
new boot connect successfully and successful connections are not logged.

With stock Kafka client defaults that orphaned producer reconnects forever, and the job sits there
until the fleet job timeout takes it down.

## What changes

Nothing here makes the broker reachable. The point is that an unreachable broker must fail in
seconds and name itself, instead of costing 45 minutes and naming nothing.

1. **`src/test/resources/application.properties`** — bound the Kafka client in tests:
   `max.block.ms=5000`, `request.timeout.ms=5000`, `delivery.timeout.ms=10000`,
   `reconnect.backoff.max.ms=1000`, `retries=3`, plus an explicit `close-timeout` on the
   `transaction-events-out` channel so a channel holding an undeliverable record cannot stall the
   test JVM's exit. An emit against a dead broker now throws within seconds.

2. **`build.gradle.kts`** — `junit.jupiter.execution.timeout.default = 8m`, the backstop for
   anything else in this module that can stall past a test boundary. JUnit kills the test and the
   module goes **RED with a named test**, instead of sitting on a runner until the fleet job timeout
   takes the whole matrix down. Same guard, same value, and the same failure shape as
   `openbank-swift-service` (#2320 item 3).

3. **`build.gradle.kts`** — a Gradle **task**-level `timeout` of 25 minutes on `test`. This is the
   guard that actually covers the observed shape, and the reason the JUnit timeout alone is not
   enough: the hang was *not* inside a running test. The last test activity was `12:09:46Z`,
   cancellation was `12:43:38Z`, and teardown reported seven `Terminate orphan process: pid (java)`.
   A Kafka producer's network thread is non-daemon and reconnects for as long as the producer object
   is open, so a leaked Quarkus application from a previous boot can hold its JVM alive after every
   test has finished — a state no per-test timeout can observe. Gradle's task timeout bounds the
   whole task including the forked JVM, and firing 20 minutes inside the fleet job timeout is what
   lets the `test-results/**/*.xml` upload actually run.

4. **`PostgresRedpandaTestResource.stop()`** — log the bootstrap address when the container is torn
   down, and null the fields. `stop()` runs mid-run, not only at the end, and without this line the
   only evidence is a wall of `could not be established` with nothing explaining what happened to the
   broker. That is exactly how this read for 37 minutes.

## Notes

- **This is not a divergence from a fleet convention.** 12 services use Redpanda Testcontainers
  (`git grep -l RedpandaContainer`), including a shared
  `openbank-libs-testing/.../PostgresRedpandaTestResource`. transaction-service needs a real broker
  because it boots a real outgoing `@Channel` emitter. The in-memory connector is the convention for
  services that switch channels; it is not the convention here.
- **The dotted-YAML-leaf-key trap is present in `application.yaml` and is deliberately left alone.**
  `auto.offset.reset` and `group.id` under `mp.messaging.incoming.payment-scheme-accepted` do not
  resolve through SmallRye's YAML source; the file already documents this in place, and the deployed
  value comes from the `transaction-service-msg-override` ConfigMap in gitops (#686). Forcing
  `earliest` onto a running consumer group would be a mass replay on a money-path channel. Out of
  scope here.

## Verification

**Verified locally** (`--rerun-tasks`, exit code read from the redirected run, not through a pipe):

```
:openbank-transaction-service:ktlintCheck  BUILD SUCCESSFUL (exit 0)
:openbank-transaction-service:detekt       BUILD SUCCESSFUL (exit 0)
:openbank-transaction-service:test         BUILD SUCCESSFUL (exit 0)
:openbank-transaction-service:quarkusBuild BUILD SUCCESSFUL (exit 0)
```

JUnit XML totals across 35 report files, read from the XML rather than the console:
`tests=144 failures=0 errors=0 skipped=1`. The single skip is
`TransactionPactProviderVerificationTest`, the broker-sourced Pact class gated on `pactbroker.url`
— the repo's sanctioned pattern, not a boot failure. `test` and `quarkusBuild` both appear in the
executed task list; neither was skipped behind an earlier failure.

**Not verified locally, and worth stating plainly:**

- The `build.gradle.kts` changes (the JUnit timeout, the Gradle task timeout, the `Duration`
  import) were added after that green run, and this machine could not complete a second one — three
  other Gradle daemons from parallel sessions on this box wedged the build on
  `buildLogic.lock` twice. CI is uncontended and is the real check for these. **This PR's own
  `build (openbank-transaction-service)` job is the verification**: it exercises the changed build
  script on the exact module that was hanging.
- **The 45-minute hang itself was not reproduced locally.** It is a CI-environment failure mode —
  a local run of this module is green precisely because the container teardown races differently.
  The log evidence above is what establishes the cause; no local run should be read as evidence
  about it, which is exactly how it was missed the first time.
- I have **not** proven by effect that the 25-minute task timeout fires (that would need a
  deliberately hanging test). It is a stock Gradle `Task.timeout`.

Closes #5940
Refs #5043, #4754, #686, #2320

### DIFF
--- a/openbank-transaction-service/build.gradle.kts
+++ b/openbank-transaction-service/build.gradle.kts
@@ -98,6 +98,23 @@ kover {
 // (ADR-0250 Phase 2, issue #4414) — this module's copy was byte-identical in substance to the
 // fleet-standard block, so nothing service-specific remains here.
 
+tasks.withType<Test> {
+    // Fail fast, not fail wide (issue #5940). This module boots a real Kafka producer against a
+    // Redpanda Testcontainer, and Quarkus tears that container down whenever the set of
+    // @QuarkusTestResource classes changes between test classes. A producer left over from the
+    // previous boot then retries against a dead port. On job 96416555756 that left the job with no
+    // output but reconnect attempts from 12:09:46Z until it was cancelled at the 45-minute fleet
+    // job timeout at 12:43:38Z, and `build/test-results/` was never written — so the job burned a
+    // full runner slot and produced no JUnit XML to diagnose from.
+    //
+    // The Kafka client bounds in src/test/resources/application.properties are the direct fix. This
+    // is the backstop for anything else in this module that can stall past a test boundary: JUnit
+    // kills the test at 8 minutes and the module goes RED with a named test, instead of sitting on
+    // a runner until the fleet job timeout takes the whole matrix down with it. Same guard and same
+    // value as openbank-swift-service, added there for the same failure shape (#2320 item 3).
+    systemProperty("junit.jupiter.execution.timeout.default", "8m")
+}
+
 // Mutation testing on the money-path domain (ADR-0063 / ADR-0030 D3; fleet rollout #1266). Weekly +
 // manual via pitest.yml, advisory — never a per-PR gate. info.solidsoft.pitest 1.19.0 supports Gradle 9.
 pitest {
--- a/openbank-transaction-service/build.gradle.kts
+++ b/openbank-transaction-service/build.gradle.kts
@@ -2,6 +2,8 @@
 // Copyright (c) OpenBank contributors. Licensed under the Apache License, Version 2.0.
 // See LICENSE in the repository root or https://www.apache.org/licenses/LICENSE-2.0 for details.
 
+import java.time.Duration
+
 plugins {
     id("openbank.quarkus-service")
     // Inline version (not the shared catalog) so enabling mutation testing stays path-scoped to
@@ -113,6 +115,21 @@ tasks.withType<Test> {
     // a runner until the fleet job timeout takes the whole matrix down with it. Same guard and same
     // value as openbank-swift-service, added there for the same failure shape (#2320 item 3).
     systemProperty("junit.jupiter.execution.timeout.default", "8m")
+
+    // The JUnit timeout above only fires while a test is RUNNING, and the #5940 hang was not that:
+    // the last test activity was at 12:09:46Z and the job was cancelled at 12:43:38Z, with seven
+    // `Terminate orphan process: pid (java)` at teardown. A Kafka producer's network thread is
+    // non-daemon and keeps reconnecting for as long as the producer object is open, so a leaked
+    // Quarkus application from a previous boot can hold its JVM alive after every test has
+    // finished — a state no per-test timeout can observe.
+    //
+    // Gradle's task timeout does cover it: it bounds the whole task, forked JVM included. At 25
+    // minutes the module fails with a named task, 20 minutes inside the 45-minute fleet job
+    // timeout, which is what lets the `build/test-results/**/*.xml` upload actually run. In the
+    // #5940 job it did not: `No files were found with the provided path:
+    // openbank-transaction-service/build/test-results/`. A cancelled job produces no report at
+    // all, so the cost is not just the runner slot — it is that nothing survives to diagnose from.
+    timeout.set(Duration.ofMinutes(25))
 }
 
 // Mutation testing on the money-path domain (ADR-0063 / ADR-0030 D3; fleet rollout #1266). Weekly +
--- a/openbank-transaction-service/src/test/kotlin/com/openbank/transaction/it/PostgresRedpandaTestResource.kt
+++ b/openbank-transaction-service/src/test/kotlin/com/openbank/transaction/it/PostgresRedpandaTestResource.kt
@@ -5,6 +5,7 @@
 package com.openbank.transaction.it
 
 import io.quarkus.test.common.QuarkusTestResourceLifecycleManager
+import org.jboss.logging.Logger
 import org.opentest4j.TestAbortedException
 import org.testcontainers.DockerClientFactory
 import org.testcontainers.containers.GenericContainer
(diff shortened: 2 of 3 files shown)


JiRaska:
## CI result on this PR — the verification the local machine could not give

`build (openbank-transaction-service)` job `12:45:32Z → 12:55:38Z`, **10m 06s**, `SUCCESS`.
`all-green: SUCCESS`.

Measured against the same signals that defined the bug:

| signal | job 96416555756 (before) | this PR |
|---|---|---|
| wall time | cancelled at 45m | **10m 06s** |
| `could not be established` lines | 247 | **0** |
| `build/test-results/` | `No files were found with the provided path` | `junit-xml-openbank-transaction-service-a1 has been successfully finalized` |

The JUnit XML uploading is the part worth calling out: its absence was not a side effect of the
hang, it was the reason the hang named nothing. A cancelled job leaves no report, so the failure was
undiagnosable from artifacts alone.

This also exercises the `build.gradle.kts` changes that were listed as locally unverified above — the
`Duration` import, the JUnit default timeout, and the Gradle task timeout all configure and run on
the module that was hanging.

Still not proven by effect, and I am not claiming it: that the 25-minute task timeout *fires*. That
needs a deliberately hanging test. It is a stock Gradle `Task.timeout`.

PR #6198, opened 2026-08-21
fix(falco): except CNPG's instance manager, the last of the API-server noise

JiRaska:
## What

After #6075's tuning landed, `Contact K8S API Server From Container` still fired **5 times in the last hour** — and all 5 were one source: **CNPG's instance manager**.

Macro (1) already excepts ns `cnpg-system`, which holds the CNPG *operator*. The instance manager is a different process: it runs **inside every database pod**, so it lives in `ledger`, `accounts`, `security-scanner` and every other namespace that owns a Postgres.

Those namespaces hold the business data path and must not get a blanket namespace exception — exactly the reasoning this file already applies to ns `messaging` and ns `admin-ui`. So this is scoped to the binary:

```
proc.exepath = /controller/manager and proc.name = manager
```

It is the container entrypoint (`parent=containerd-shim`, `command=manager`) and its job is to watch its own `Cluster` CR through the API server.

## Measured, not projected

| window | result |
|---|---|
| 6h sample, 100 events | **71 silenced** by this macro |
| last 1h, post-deploy | **5 of 5** — 100% of what the rule still reports |

The other 29 in the 6h sample are Strimzi JVM processes from **before** the tuning deployed (falco pods restarted 07:40 and 11:50); the entity-operator macro covers them and none appear in the last hour. Worth stating explicitly, because a 6h window straddling a rollout is exactly the kind of denominator that makes a fix look partial when it is not.

## Still firing, deliberately

Any **other** process in a database pod — postgres itself, a debug shell, anything dropped in — reaching the API server. `/controller/manager` is a CNPG-specific path no other workload in this estate has, so the exception cannot widen past it.

## Why this matters for #5734

Cumulative effect of this and #6075, measured:

```
before:  ~33,000 events/day
now:     5 in the last hour, and `Drop and execute new binary in container` silent
```

That is the **precondition** for alerting. An alert on an untuned 33k/day ruleset gets muted within a day, which is worse than no alert — and #5734's whole point is that the runtime-security control is currently write-only. Alerting itself still waits on #6032 (the Loki ruler has never loaded a single rule).

Refs #5734

### DIFF
--- a/openbank-infra/gitops/apps/falco.yaml
+++ b/openbank-infra/gitops/apps/falco.yaml
@@ -221,13 +221,35 @@ spec:
                  and container.name in (user-operator, topic-operator)
                  and proc.pname = "tini")
 
+            # (6) CNPG's INSTANCE manager. (1) already excepts ns `cnpg-system`, which holds
+            # the CNPG *operator* — but the instance manager is a different process that runs
+            # inside every database pod, so it lives in `ledger`, `accounts`,
+            # `security-scanner` and every other namespace that owns a Postgres. Those hold the
+            # business data path and must NOT get a blanket namespace exception, which is why
+            # this is scoped to the binary instead.
+            #
+            # It is the container entrypoint (parent=containerd-shim, command=manager) and its
+            # job is to watch its own Cluster CR through the API server. Measured 2026-08-21
+            # AFTER the earlier tuning landed: 71 events in 6h (~284/day) and 5 of the 5
+            # remaining events in the last hour — i.e. 100% of what this rule still reports.
+            #
+            # SCOPE: the exepath AND the process name. `/controller/manager` is a CNPG-specific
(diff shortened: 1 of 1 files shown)


JiRaska:
CI red here is **not this PR** — main is red and the gate scans the whole tree.

`gates (gitops)` fails on:

```
prometheus-rules-ai.yaml: AiModelNeverSucceeds: has a range window [6h]
but no `subject_period` annotation
```

Nothing to do with `falco.yaml`. Confirmed by checking the last completed `main` run, which has the same single finding. Cause is an ordering collision: #6041 added the rule, #6053 then added the gate requiring windowed rules to declare their subject cadence, and the rule postdates the baseline #6053 froze — so it must declare rather than being grandfathered. The gate is behaving correctly.

Fixed by #6203 (already open; I confirmed it there and dropped my duplicate). This PR should go green once that merges — no change needed here.

PR #6199, opened 2026-08-21
fix(lending): order pending proposals by the entity property, not the column

JiRaska:
## The defect

`GET /api/v1/lending/compliance-packs/proposals/pending` answered **500 on every call from the day it shipped**:

```
org.hibernate.query.SemanticException: Could not interpret path expression 'proposed_at'
```

The repository ordered by `proposed_at` — the **column** — inside a string Hibernate parses as HQL, where only the entity property `proposedAt` exists. The column spelling is not a synonym there.

**No parameters were needed.** A plain GET, on a money-path service.

## Why nothing caught it

`CompliancePackActivationServiceTest` substitutes a hand-written in-memory `findByState` that filters a map, so **no HQL is ever issued and the query string is never parsed**.

The sibling method one line below already spells its ordering correctly (`jurisdiction, productType, packVersion`) — which is exactly what made the broken line read as fine. Same shape as the consent `SuppressionEntity` defect (#5711), where mocked repositories hid a mapping that could not work.

## How it was found

Fuzzing past authentication (#5913). It had been invisible because a **403 answered first**: of 306 operations across the fleet, 302 were auth-blocked and only 4 ever exercised (#5769). This is one of four unconditional 500s that surfaced the moment that shield came off.

## Falsified, not assumed

`CompliancePackPendingQueryIT` drives the real endpoint against a real database. With `proposedAt` reverted to `proposed_at`:

```
BUILD FAILED
SemanticException: Could not interpret path expression 'proposed_at'
```

— the exact production exception, reproduced in the test. Restored, it passes.

The test asserts two things a fake repository cannot: that the query **parses**, and that the ordering it names is the one that comes back. Rows are inserted newest-first, so an ordering that is ignored — or applied backwards — cannot pass on insertion order. `order by` is the part of a query most easily wrong in a way that still returns rows.

## Verification

- `:openbank-lending-service:ktlintCheck :detekt :test --tests '*CompliancePackPendingQueryIT*'` → **BUILD SUCCESSFUL**, 0 lint findings
- negative case above → BUILD FAILED with the production exception

## Linked issues

Refs #5913, #5769, #5711


JiRaska:
Closing as a duplicate of #5948, which is **earlier** (13:45 vs 13:31 the next day) and fixes the same line the same way. Same collision as #5731/#5723 — and this one is entirely my fault twice over: I checked for competing PRs before starting the party and ledger fixes, having learned that lesson this morning, and then went straight into lending without checking.

One thing here is genuinely additive and I have offered it to #5948 rather than keeping a competing PR open: the IT asserts **the ordering as well as the status**.

`order by` is the part of a query most easily wrong in a way that still returns rows — swap `proposedAt` for `updatedAt` and `statusCode(200)` still passes. This test inserts newest-first and asserts oldest-first comes back, so an ordering that is ignored, or applied backwards, cannot pass on insertion order:

```kotlin
val newer = insertProposal(version = 9101, proposedAt = OffsetDateTime.now())
val older = insertProposal(version = 9102, proposedAt = OffsetDateTime.now().minusDays(2))
…
assertThat(ids.indexOf(older.toString()))
    .`as`("proposals must come back oldest-first — `order by proposedAt`")
    .isLessThan(ids.indexOf(newer.toString()))
```

Both branches falsified the fix the same way — reverting `proposedAt` to `proposed_at` reproduces `SemanticException: Could not interpret path expression 'proposed_at'` in the test.

PR #6203, opened 2026-08-21
fix(observability): declare subject_period for AiModelNeverSucceeds

JiRaska:
## What

The \`check-alert-window-vs-subject-period\` gate (#5736, enforced) started failing
fleet-wide: \`AiModelNeverSucceeds\` has a \`[6h]\` range window and no \`subject_period\`
annotation, so the gate can't check the window against its subject.

Every open PR's \`gates (gitops)\` job was failing on this — the gate scans the whole
tree, not just the PR diff, so a single undeclared annotation on \`main\` blocked
everyone downstream (confirmed while investigating unrelated CI failures on #5972,
#6086, #5875, #6192, all with unrelated diffs, all failing on the identical error).

## Fix

\`AiModelNeverSucceeds\` is traffic-driven, not scheduled — matches the shape of
customer-edge's alerts, which already declare \`subject_period: continuous\` for the
same reason. Declared it here too.

## Verification

\`\`\`
$ python3 .github/scripts/check-alert-window-vs-subject-period.py
SUBJECTS=122  # alert rules
check-alert-window-vs-subject-period: 50 windowed alert rule(s); 142 periodic subject(s)...
check-alert-window-vs-subject-period: no findings.
\`\`\`

### DIFF
--- a/openbank-infra/gitops/components/observability/prometheus-rules-ai.yaml
+++ b/openbank-infra/gitops/components/observability/prometheus-rules-ai.yaml
@@ -355,6 +355,7 @@ spec:
             severity: warning
             team: platform
           annotations:
+            subject_period: continuous
             summary: "Every LLM call to {{ $labels.model }} is failing — not one has succeeded in 6h"
             description: >-
               The gateway is answering and every answer is an error, so this is not an outage of


JiRaska:
Confirming this independently, and adding the context that makes it urgent: **main is currently red on this**, so every open PR is failing too.

The last completed `main` CI run has `gates (gitops): failure` with exactly this one finding. I hit it on an unrelated Falco PR (#6198) and traced it here rather than to my own change — the gate scans the whole tree, not the diff, so it sinks everything until this lands.

**Neither side did anything wrong.** #6041 added `AiModelNeverSucceeds`; #6053 then added the gate requiring every windowed rule to declare its subject cadence. Because the rule is newer than the baseline #6053 froze, it is correctly required to declare rather than being grandfathered — the gate working exactly as designed. The two just landed in the order that leaves main red. Same ordering-collision class as the Flyway V4/V5 one (#5628), one tree over.

`continuous` is the right value: callers reach the LLM gateway when a user or an agent asks something, so there is no cadence a 6h window could be shorter than.

Verified locally on this exact change: the gate goes **1 finding → none** (50 windowed rules, 142 derived periodic subjects, 45 baselined), and its own `--self-test` still passes — so this is a declaration, not a way around the check.

I had prepared an identical fix and am dropping it rather than duplicating; the only thing mine carried extra was an inline comment explaining why `continuous` (traffic-driven, no cadence to be shorter than). Worth adding here if you want it, but not worth another round-trip while main is red.

PR #6209, opened 2026-08-21
feat(agent): durable AI audit provenance

### LINKED ISSUE #6191: Deliver AI audit events to the tamper-evident audit trail
AI services emit model and policy provenance through the shared AuditEventPublisher, but the current default publisher logs locally while audit-service consumes Kafka topics.\n\nScope: add a bounded, mTLS-authenticated Kafka publisher for agent-service; provision its topic and least-privilege producer/consumer ACLs; prove an AI-attributed event is persisted by audit-service. Keep the fallback for services not enrolled in this transport.\n\nRefs #6185

JiRaska:
Closes #6191\n\nAdds a durable agent-side audit outbox, retrying Kafka handoff, idempotent audit-store ingestion, least-privilege topic/group ACLs, and real PostgreSQL coverage for handoff and redelivery.\n\nKafka delivery remains deliberately disabled by default: D5 stays partial/not-attested until the topic/ACL rollout, runtime connectivity, retention/backlog controls, and independent anchor evidence are observed.


JiRaska:
Heads-up on a collision, not a duplicate claim: #6260 is an independent concurrent answer to #6191, opened without sight of this branch. Both cannot land — the designs diverge (topic `openbank.agent.audit` vs `openbank.agent.audit.events`, the existing `AuditConsumer` vs a new `AgentAuditConsumer`), and each is stronger on a different axis. This PR has the durable producer outbox and retrying handoff; #6260 has the security-context attribution, the `agentId` aggregate mapping, `TopicAttribution` coverage, the contract-vs-code binding, a threat-model correction, and an end-to-end proof that a row lands in `audit_entries` (real DB, plain JDBC read-back) — which this PR cannot show while Kafka delivery is disabled by default.

Merging both would put two topics and two consumers on one path, to drift apart later. Flagging it so the choice is deliberate; I have not touched this branch.

JiRaska:
**Collision notice, and a recommendation that favours this PR's design.**

#6260 (`feat(agent): deliver AI audit events to the tamper-evident trail`) was opened
independently and also targets #6191. It uses `Refs #6191`, not a second `Closes`, precisely so
this one keeps the claim. The two are not mergeable together — different topic name
(`openbank.agent.audit` vs `openbank.agent.audit.events`) and a different consumer.

**This PR's producer design is the better one.** A transactional outbox with a retrying handoff is
strictly stronger than #6260's best-effort direct emit, which loses an event the broker never
accepted. That is the half worth keeping.

**But as it stands it would close #6191 with delivery disabled.** `application.yaml:409` reads
`enabled: ${AGENT_AUDIT_KAFKA_ENABLED:false}`, and the PR body says D5 stays partial until rollout
and connectivity are observed — which is the honest position, and exactly why the issue should stay
open until then. Closing it on merge would record the AI audit trail as delivered while nothing is
delivered. That is the shape this repo has been finding all week: a control that reports healthy
because in its current state it cannot report anything else.

**Suggested resolution — take three things from #6260, then `Closes` is earned:**

1. **The `TopicAttribution` entry** for the topic in `openbank-audit-service`. Without it a row that
   does arrive is attributed `unknown`, and `source_service` is chain-hashed into `record_hash`, so
   it cannot be corrected later. #6260's negative control for this is
   `expected: "agent-service" but was: "unknown"`.
2. **`AuditConsumer.aggregateFields` gaining `"agentId" to "AGENT"`**, appended last so first-match
   wins and it can only turn `unknown` into `AGENT`. Its negative control:
   `expected: "compliance-officer-…" but was: "unknown"`.
3. **The real-DB integration test** (`AgentAuditEventPersistenceIT`): drive the real CDI
   `AuditConsumer` against Testcontainers Postgres and read the row back over a **plain JDBC
   connection sharing no code with the write path**, asserting attribution and *recency* on
   `occurred_at` — recency, never non-nullity, since `Instant.EPOCH` defaults are a lie every test
   agrees with. A mocked repository cannot establish that a row landed.

Until delivery is on and a row is proven to arrive, `Refs #6191` would be the accurate marker here
too.

**One more thing worth carrying over regardless of which lands:** agent-service's threat model DFD
draws `AuditEventPublisher ──▶ Kafka audit-events-out` and its T-R1 row credits an attributable
record. Both have been false for the life of the document — the only `AuditEventPublisher`
implementation in the repo is `LoggingAuditEventPublisher`, a log line. #6260 corrects that text;
if this PR lands instead, the correction should come with it. The same shape very likely exists in
copilot-service's and mcp-service's threat models, whose AI audit events also reach only a log.

Not verified by me: anything against a live broker or cluster. Everything above is read from
`origin/main` and from the two branches.

JiRaska:
### Measured relationship: #6260 and #6209 are competing designs, not a stack and not a duplicate

Both answer #6191 (ADR-0031 D5: agent AI audit events reaching the tamper-evident trail). Measured, not inferred:

- **Not a stack.** `git merge-base` of the two heads is `ce55de69`, which is also #6260's merge-base with `main` — neither branch contains the other.
- **They hard-conflict.** Merging both onto `origin/main` in sequence produces three conflicts, one of them an `add/add` on the contract:
  ```
  UU openbank-audit-service/src/main/kotlin/com/openbank/audit/application/AuditConsumer.kt
  AA openbank-contracts/openbank-agent-service/asyncapi.yaml
  UU openbank-infra/gitops/components/agent/kafka-topic.yaml
  ```
  Seven files overlap in total. Whichever lands second is blocked on a manual reconciliation of the contract and the topic definition.
- **They pick different topics** — `openbank.agent.audit` (#6260) vs `openbank.agent.audit.events` (#6209) — and different consumers (existing `AuditConsumer` vs a new `AgentAuditConsumer`).
- **They are strong on different axes.** #6209 adds a durable producer outbox (`V4__agent_audit_outbox.sql`) that #6260 lacks; #6260 delivers by default and proves the row lands in `audit_entries`, whereas #6209 ships `enabled: ${AGENT_AUDIT_KAFKA_ENABLED:false}` on `agent-audit-events-out`, so as merged it delivers nothing until that flag is flipped.

Neither is closed here: this is a design choice between two working answers, not a duplicate to sweep up. A maintainer picks one; the losing PR's stronger axis (the outbox, or the default-on delivery + attribution mapping) is the natural follow-up on top of the winner.

PR #6229, opened 2026-08-21
fix(admin-ui): prevent overlapping party refreshes

JiRaska:
## Summary
- prevent overlapping party/KYC detail refresh requests while loading
- expose localized busy semantics and accessible refresh label
- add focused source guard

Preserves existing party/KYC endpoints and RBAC.

Refs #6148


JiRaska (commented):

The Party/KYC endpoints, supplementary KYC fallback, privacy behavior, deep links, and RBAC are unchanged. Refresh now uses type=button, disabled/aria-busy during the existing loading state, a localized aria-label, and aria-hidden decorative icon. Focused guard covers both svcUrl endpoints and semantics; diff-check clean; CI green.

Issue #6217, opened 2026-08-21
Epic: ADR-0269 customer-facing lending — delivery order and tracking


JiRaska:
Tracks delivery of **ADR-0269** (merged, `146ab25`) and its client-side realisation **openbank-app APP-ADR-0001** (openbank-app#528).

## The one rule that sets the order

Consent switches and the suppression floor ship **before the first offer object exists**. A funnel that is measured before it is constrained never gets constrained afterwards. Slice 0 is therefore not negotiable as a starting point, and no later slice may be pulled forward past it.

## Platform (this repo)

| | Slice | State |
|---|---|---|
| 0 | #6212 — `credit_offers` consent (default off) + distress suppression floor | **merged** |
| 1 | #6213 — unified `CreditApplication` journey + customer-readable projection | **merged** |
| 2 | #6214 — server-side quotes, binding offers, pre-approved limits | in #6235, awaiting review |
| 3 | #6215 — `CustomerCreditProfile` as the one 360-derived read model | in #6235, awaiting review |
| 4 | #6216 — credit AI levels L0/L1/L2 | in #6235, awaiting review |

Slices 2–4 were not shipped as three PRs. Each rebase of the stacked chain over a squash-merged
parent rewrote the children, so 2, 3, 4 plus the consent surface, financial health and the
authenticated funnel channel were collapsed into a single reviewable branch, **#6235** (53 files).
The three issues stay open until it merges.

#6235 is green (98 pass, 0 fail) and blocked only on `require_extra_approval_for_unattributed_changes`
— it needs one approving review.

## Client (openbank-app)

| | Slice | Blocked on | State |
|---|---|---|---|
| A | openbank-app#530 — consent surface + the no-fetch rule | #6212 | **merged** |
| B | openbank-app#531 — financing hub, server-only pricing, rendered journey state | #6213, #6214 | **merged** |
| C | openbank-app#532 — financial health, four pillars, no score | #6215 | **merged** |
| D | openbank-app#533 — AI levels on the existing copilot, no new chat surface | #6216 | **merged** |
| E | openbank-app#534 — reachability, Simple Mode, non-conversion telemetry | A–C | **merged** |

The client is complete on `main`. It renders server state only, so the screens behind the consent
switch stay inert until #6235 deploys — by design, not a gap.

## What is deliberately not here

[...]

Issue #6216, opened 2026-08-21
ADR-0269 slice 4: credit AI levels L0/L1/L2 — explain and prepare, never decide


JiRaska:
Delivers **rule 5 of ADR-0269**: AI explains and prepares, never decides, submits or signs.

Runs on existing machinery — ADR-0089 (customer assistant behind the edge), ADR-0217 (credit agents as MCP-exposed, OPA-gated assistants that never transition the state machine), ADR-0031 (agent governance), ADR-0216 (Annex III 5(b) high-risk obligations).

## Three consent-scoped levels

**L0 explainer** — default on. Answers only what the customer asks ("why is my instalment this?", "what is APRC?", "what happens if I repay early?"). Receives **no** 360 credit profile. Never speaks first. Safe as a default precisely because it cannot initiate.

**L1 advisor** — opt-in. On request, reads `CustomerCreditProfile` (#6215) to answer "can I afford this?" with its workings: effect on reserve, on DSTI, on the balance forecast, and what happens if income stops. Required property: **it must be able to answer no**, and that answer must not be overridable by any sales objective. An advisor that cannot decline is not an advisor.

**L2 agent capabilities** — opt-in one at a time: refinancing-price watch (fires only above a customer-set saving threshold), fixation-end watch, application pre-fill, instalment-risk warning (which proposes moving a due date, never a new loan), consolidation scan.

## Hard boundaries (all levels)

- No level transitions the state machine, accepts an offer, raises a limit or draws funds. Every money-path action is an SCA ceremony performed by the customer.
- Every output carries "why I said this" plus the inputs it used.
- Every output has a one-tap "stop sending me this", scoped to that capability.
- Full audit log of what the agent saw and said, available to the customer and usable as creditworthiness evidence (ADR-0214).
- **AI never decides credit.** The decision is the deterministic engine's (ADR-0213), reason-coded, with a right to human review. AI output may be an input, never a verdict.
- Model data access is scoped to the consented level — L0 does not receive the profile at all.

## Depends on

#6212 (consent levels), #6215 (profile), #6214 (quotes — an advisor with no price to reason about is a chat toy).

Parent: ADR-0269

JiRaska:
Tracked in the delivery epic: #6217

JiRaska:
Shipped in oss#6235 (merged 3b62a4a).

Issue #6215, opened 2026-08-21
ADR-0269 slice 3: CustomerCreditProfile as the one 360-derived credit read model


JiRaska:
Delivers the shared read model behind affordability, financial health and every AI level.

## Scope

`CustomerCreditProfile`, computed server-side over the ADR-0210 Customer 360 silver layer:

- `monthlyIncomeEstimate` (from recurring inbound payments)
- `monthlyObligations` (instalments, direct debits, standing orders)
- `discretionaryIncome`, `volatility`, `bufferDays`
- `existingCreditUtilisation`, `tenureMonths`, `kycStatus`, `delinquencyHistory`

## Why server-side, and why one of them

Three consumers need this number: the creditworthiness assessment (which must be auditable and reproducible), the customer's financial-health view, and the AI advisor. If the client computes its own version, two numbers that disagree is worse than one number that is wrong — and only the server's version is defensible in an audit.

The suppression floor (#6212) also reads `bufferDays` and `delinquencyHistory` from here, so this slice and #6212 share an interface even though #6212 ships first with a narrower internal read.

## Depends on

ADR-0210 (shipped). Feeds #6212, #6214, and the app's financial-health screen.

Parent: ADR-0269

JiRaska:
Tracked in the delivery epic: #6217

JiRaska:
Shipped in oss#6235 (merged 3b62a4a).

Issue #6212, opened 2026-08-21
ADR-0269 slice 0: credit_offers consent (default off) and the distress suppression floor


JiRaska:
Delivers **rule 1 and rule 2 of ADR-0269**, and nothing else. This is deliberately the first slice: the ADR states that consent switches and suppression rules ship *before the first offer object exists*, because a funnel that is measured before it is constrained never gets constrained afterwards.

## Scope

**Consent (`credit_offers`)**
- New consent kind, default **off**, alongside the existing consent machinery.
- Per-channel delivery sub-preferences (in-app / push / email) that are meaningless while the parent is off.
- A separate `credit_profile_use` consent for using the 360 profile for offer eligibility (distinct from using it for a creditworthiness assessment the customer asked for).
- Revocation is immediate and cross-channel, and emits an event `campaign-service` consumes as a workflow signal (ADR-0200 D2 path).
- No expiry-driven re-prompt and no re-enrolment on release: consent never re-arms itself.

**Suppression floor (`lending-service`)**
- Evaluated server-side on every offer read and every campaign step, not only at enrolment.
- Inputs: arrears on any facility, negative/overdrawn balance, enforcement or insolvency marker, hardship flag, failed affordability inside the cooling window, 360-derived buffer below floor.
- Versioned like an ADR-0213 decision table, with machine-readable reason codes; fail-closed.
- Frequency cap on top: one credit-marketing contact per 30 days, and only on a changed input.

**Enforcement points**
- `customer-edge`: no pre-approved limit is served while consent is off.
- `campaign-service`: refuses enrolment into any journey whose product kind is credit; credit campaign templates gain a mandatory product-kind field so the step gate can refuse them.

## Falsification (required — a gate is not proven by what it prints)

Per the repo's own rule, each guard needs its negative case written first:
- Consent off → the pre-approved route returns nothing **and** the campaign enrolment is refused. Assert the refusal, not the absence of a log line.
- Consent on + arrears → suppressed, with the specific reason code asserted. A non-zero exit alone is not proof.
- Revocation mid-journey → the running workflow stops at the next step, not at the next batch.
- Remove the suppression check → the test must go red.

## Out of scope

Offer objects, quotes, journey projection. Those are the next slice and must not start before this one merges.

Parent: ADR-0269 (docs/adr/0269-customer-facing-lending-pull-only-offer-surfacing.md)

JiRaska:
Tracked in the delivery epic: #6217

PR #6235, opened 2026-08-21
feat(lending): ADR-0269 platform — quotes, credit profile, AI levels, consent surface, financial health, funnel

### LINKED ISSUE #6214: ADR-0269 slice 2: server-side quotes, binding offers and pre-approved limits
Delivers **rule 4 of ADR-0269**: price comes only from the server, and only as a quote.

## Scope

- `POST /credit/v1/quotes` — indicative, non-binding, resolved from the pinned catalog revision (ADR-0264 `banking.loan-v1`). Returns rate, instalment, APRC, total payable, fees and a validity window.
- A **binding offer** is a distinct object with `offerId` and `expiresAt`, produced by the ADR-0213 decision engine. Acceptance references the offer id; an expired offer is re-quoted, never extrapolated.
- `GET /customer/v1/credit/pre-approved` — pre-approved limits, gated by #6212 (returns nothing while consent is off; the client is expected not to call it at all).
- Pre-contractual information generated and archived via documents-service before signature; signature is the existing SCA ceremony.

## Constraint that makes this worth its own slice

No client, no campaign template and no model may compute or cache a price. Every rate, instalment and APRC a customer ever sees traces to one of these two objects. The app already follows this rule for loan intake; this slice makes it the only possible behaviour.

## Depends on

#6212 (consent gate), #6213 (journey state to attach an accepted offer to).

Parent: ADR-0269

JiRaska:
Closes #6214, #6215, #6216, #6212. Completes the ADR-0269 platform (epic #6217) — slice 1 is already on main via #6230.

> **Note for the reviewer on scope.** This PR started as slice 2 and grew: slices 3-7 were merged into its branch rather than into main as the stack collapsed, so it now carries the whole remaining platform. 53 files. The commits are the natural reading order — each is one slice and stands alone.

## What ships

| Area | What |
|---|---|
| `Aprc` (libs-domain) | CCD2 Annex III solver — the **only** implementation permitted to compute an APRC |
| `CreditQuoteCalculator` | Instalment from the shared `Amortization` schedule the loan is actually booked against |
| `OfferSurface` | PUSH needs consent, PULL does not — the pull-only rule, as a type |
| `CustomerCreditProfile` (V9) | Two ClickHouse views over the ADR-0210 silver layer |
| `CreditAiLevel` L0/L1/L2 | Explainer / advisor / agent, each its own consent |
| `PUT /customer/v1/credit/consents` | The route by which a customer can actually switch offers ON |
| `FinancialHealth` | Four pillars, no score |
| `CreditFunnelPublisher` | Authenticated funnel telemetry with no word for conversion |

## The decisions worth your attention

**Everything unknown suppresses.** An unreadable consent, a thrown upstream call, incomplete signals and an unknown buffer all refuse. A marketing gate that degrades open under failure stops existing exactly when its upstream is unhealthy: blast radius of failing closed is a missed offer, of failing open an offer to someone in arrears.

**PUSH vs PULL is deliberate asymmetry.** The bank must not *initiate* without consent; refusing to answer a price question the customer just asked would be the same dark pattern pointing the other way. The distress floor applies to both, because there the answer itself is the harm.

**No aggregate score anywhere**, and no vocabulary for conversion in the funnel. Both are asserted by tests, not left to review: a single grade invites optimisation, and a funnel that can say "converted" is one somebody eventually optimises.

**An empty loan book gives debt service zero; an unreadable one gives null.** Only one of those means "no debts". Same distinction between `HISTORY_TOO_THIN` (a new customer) and `SIGNALS_UNAVAILABLE` (an outage) — collapsing them buries a real fault in normal traffic.

**Reserve is UNKNOWN, never inferred from unspent cashflow.** A buffer someone does not have, derived from money they merely did not spend, is the most dangerous number that screen could show.

## Known gap, stated rather than modelled away

Enforcement orders and insolvency proceedings have **no source** in this deployment — no service ingests a court register. They are reported as absent rather than unknown, because permanent `complete = false` would suppress every offer forever and be indistinguishable from the feature being switched off. Recorded in `LoanBookDistressAdapter`'s KDoc and threat-model section 9b; the first thing an operational-risk review should challenge.

## What the enforced gates caught (all fixed here)

- **kafka-topics-exist / kafka-acl-coverage** — the new topic was referenced in code and declared nowhere, with no Write ACL for the edge or Read for analytics-sink. The publisher is best-effort by design, so this would have failed *silently, forever* — the #1717 shape.
- **event-contract-coverage (ADR-0006)** — creating the edge's first AsyncAPI contract made its three grandfathered baseline entries stale, so the audit, onboarding-funnel and feedback streams are documented too and the baseline lines are gone. The ratchet only shrinks.
- **event-contract-code-agreement** — the 17 audit event types were undocumented, and the three telemetry publishers named their event type through a constant the gate cannot evaluate. Each now binds a local `val eventType = "<literal>"`.
- **event-consumer-liveness (ADR-0160)** — analytics-sink now subscribes, which is the real fix rather than an allowlist entry excusing a published-but-unconsumed topic.
- **threat-model-diff (ADR-0030 D2)** — sections 9a, 9b and 9c cover the new inbound surfaces and outbound client edges.

## Verification

98 checks pass, 0 fail. lending-service 306 tests, customer-edge 398, copilot 192, libs-domain 489 — all 0 failures, 0 skipped, counts read from result XML rather than the build's exit code.

Falsified rather than asserted where it matters: removing the consent and fail-closed guards turns 7 tests red; the affordability surplus rule was rewritten after a test proved the original could never fire (inside the DSTI limit it was dead code dressed as a safeguard).

## Client side

Already on main in openbank-app: #536, #543, #538, #539, #540, #541.

### DIFF
--- a/.github/event-contract-baseline.txt
+++ b/.github/event-contract-baseline.txt
@@ -16,9 +16,6 @@ openbank-balance-service:openbank.balance.events
 openbank-card-issuance-service:openbank.cards.events
 openbank-clearing-service:openbank.clearing.batch.event
 openbank-consent-service:openbank.consent.events
-openbank-customer-edge:openbank.customer.audit
-openbank-customer-edge:openbank.feedback.events
-openbank-customer-edge:openbank.onboarding.funnel.events
 openbank-dispute-service:openbank.dispute.events
 openbank-document-service:openbank.documents.document.event
 openbank-domestic-payment:openbank.domestic.payment.events
--- a/docs/threat-models/openbank-lending-service.md
+++ b/docs/threat-models/openbank-lending-service.md
@@ -392,16 +392,18 @@ against `lending.intake.caller-principal`, which refuses every call when unset.
   caller, but `operator-lending-write` already grants the same action more broadly; the rule only
   becomes load-bearing if that blanket operator grant is narrowed.
 
-## 9a. Customer credit journey read (ADR-0269) — STRIDE supplement
+## 9a. Customer credit journey read and indicative quotes (ADR-0269) — STRIDE supplement
 
-One new inbound surface on the same customer-edge trust boundary as section 9, and a read path,
-which changes what can go wrong: section 9's risk is a fraudulent WRITE, this one risks a
-DISCLOSURE.
+Two new inbound surfaces on the same customer-edge trust boundary as section 9, and both are read
+paths, which changes what can go wrong: section 9's risk is a fraudulent WRITE, these risk a
+DISCLOSURE and a MIS-DISCLOSURE.
 
 - `CustomerCreditJourneyResource` — `GET /api/v1/lending/intake/applications[/{id}]`. Returns the
   caller's own applications as customer-readable journeys.
+- `CustomerQuoteResource` — `POST /api/v1/lending/intake/quotes`. Returns an indicative,
+  non-binding price. Persists nothing.
 
-It reuses section 9's caller control verbatim: the handler compares `SecurityIdentity.principal.name`
+Both reuse section 9's caller control verbatim: the handler compares `SecurityIdentity.principal.name`
 to `lending.intake.caller-principal` and refuses when it does not match or is unset, and the party id
 comes only from `X-Customer-Party-Id`.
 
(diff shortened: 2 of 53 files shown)


JiRaska:
**#6017 and #6235 both claim `openbank-consent-service` `openapi.yaml` `info.version: 1.7.0`.**
`main` is at `1.6.0`, so whichever merges second ships a second, different contract under a version
already published — and there is nothing to notice it.

```
main   1.6.0
#6017  1.7.0
#6235  1.7.0
```

Git will not flag this: both sides write the identical `version:` line, so the merge is textually
clean. The contract gate classifies the bump from the diff against the merge-base *as of the run*,
so it cannot see a competing claim that lands after the run. Whoever goes second needs **1.8.0**,
re-read off live `main` at merge time rather than taken from this comment:

```bash
git fetch origin main && git show origin/main:openbank-consent-service/src/main/resources/openapi.yaml | grep -m1 '^  version:'
```

Note #6235 also carries `openbank-lending-service` and `openbank-customer-edge` spec bumps, so it is
the more entangled of the two — its lending version jumps 1.14.0 → 1.16.0, which is consistent with
it being slice 2 of a stack, but worth confirming that 1.15.0 belongs to slice 1 and not to a gap.

Found by sweeping the openapi-touching PRs against `main`. Three other PRs (#6019, #5905, #6481)
turned out to have bumped to a number that a *merged* PR had already taken — the same trap from the
other direction, and the one that gets missed, because looking for a race scans open PRs.

JiRaska:
**#5905 and #6235 both bump `openbank-lending-service`'s `openapi.yaml`, and the merge order is now
load-bearing.** Neither is wrong; they only work in one sequence.

```
main    1.14.0
#5905   1.15.0
#6235   1.16.0
```

**#5905 must land first.** In that order both are correct and each step is a single minor. The other
way round, `main` reaches 1.16.0 and #5905 arrives claiming 1.15.0 — a version *behind* the published
contract, which the api-contract gate classifies against the merge-base as of its own run and will
not necessarily catch if #5905's last run predates the merge.

Nothing enforces this. The two PRs conflict in no file, both are green, and git sees no relationship
between `1.15.0` and `1.16.0`.

For context on how quickly this moves: yesterday #5905 claimed 1.14.0, which #6230 had already taken;
it has since been corrected to 1.15.0, and #6235 was already at 1.16.0. Re-read the version off live
`main` at the moment you merge rather than trusting this comment:

```bash
git fetch origin main && git show origin/main:openbank-lending-service/src/main/resources/openapi.yaml | grep -m1 '^  version:'
```

Two other spec-version items from the same sweep are still open and unrelated to these:
**#6019** claims `openbank-card-issuance-service` `1.8.0`, which `main` already holds (taken by
#4308) — it needs 1.9.0; and **#6017 / #6235** both claim `openbank-consent-service` `1.7.0`, a live
race where whoever goes second needs 1.8.0.

PR #6254, opened 2026-08-21
fix(admin-ui): prevent overlapping KYC searches

JiRaska:
## Summary
- prevent overlapping KYC case loads from Search and Refresh
- expose localized busy/accessibility semantics and hide decorative icons
- add focused source guard

Preserves `/api/v1/kyc/cases` query behavior, empty-state handling, and RBAC.

Refs #6148


JiRaska (commented):

KYC search and refresh now use type=button, disabled/aria-busy while the existing load is active, localized aria-labels, and decorative icons are hidden. The `/api/v1/kyc/cases` URL/query behavior, 404/405 no-data fallback, supplementary state handling, RBAC, and data flow are unchanged. Focused guard covers endpoint and semantics; diff-check clean.

Issue #6252, opened 2026-08-21
CI: 93% of PR job time is runner queue, not work — the 8 gate shards now compete for slots they cannot get


JiRaska:
## What was measured

Sampled the last 300 completed `pull_request` workflow runs and the job/step timings of 6 recent
successful `CI` runs. Query pinned so a later run is comparable:

```
gh api "repos/JiRaska/open-bank-oss/actions/runs?event=pull_request&status=completed&per_page=100&page=1..3"
gh api "repos/JiRaska/open-bank-oss/actions/runs/<id>/jobs?per_page=100"   # created_at / started_at / completed_at
```

`CI` is the critical path a developer waits on: **p50 15.1 min, p95 18.4 min** (n=10). That is
*better* than the ~20 min figure the last throughput review recorded, so this is not a regression —
it is where the remaining time now sits.

## The split that decides everything: 93% of it is queue, not work

Across 78 CI jobs:

| | p50 | p95 | max |
|---|---|---|---|
| **queue** (`created_at` → `started_at`) | **6.6 min** | 11.4 min | 13.4 min |
| **run** (`started_at` → `completed_at`) | **0.4 min** | 1.1 min | 2.4 min |

**Queue is 93% of elapsed job time.** A gate shard that does 33 s of work waits ~6.6 min for a
runner. Any further optimisation of gate *execution* is therefore close to worthless — that half is
already excellent (the whole 8-shard set does 5.6 min of work).

## The gate shards are now competing with each other

All 8 shards are created in the same instant and admitted over ~7 minutes. One run:

```
gates (api)          created +0.0m   started +3.2m   done +3.6m
gates (supplychain)  created +0.0m   started +5.8m   done +6.5m
gates (kotlin)       created +0.0m   started +10.4m  done +11.0m
gates (data)         created +0.0m   started +10.7m  done +11.2m
gates (gitops)       created +0.0m   started +11.0m  done +11.9m
gates (security)     created +0.0m   started +11.1m  done +11.5m
gates (lint)         created +0.0m   started +11.2m  done +12.2m
gates (registry)     created +0.0m   started +11.2m  done +12.0m
```

Consistent over 6 runs: **median wall-clock 11.3 min for 5.6 min of work, median stagger 7.0 min.**
There is no `max-parallel` in the matrix — this is GitHub-hosted runner availability.

**This does not mean the sharding was wrong.** #4340–#4346 measured CPU (62 s → 26 s on the slowest
[...]

PR #6257, opened 2026-08-21
perf(ci): consolidate the 8 gate shards into 3 buckets (#6252)

JiRaska:
Implements option 1 from #6252.

## Why

93% of a CI job's elapsed time is runner **queue**, not work — queue p50 6.6 min against run p50
0.4 min, measured over 6 PR runs / 78 jobs. The eight shards are created in the same instant and
admitted over a ~7 min stagger, so they spend **11.3 min of wall-clock doing 5.6 min of work** while
holding eight hosted-runner slots.

**The earlier sharding was not wrong.** #4340–#4346 measured CPU (62 s → 26 s on the slowest shard)
and were right at the time. The binding constraint has since moved from CPU to runner slots, so
eight-way parallelism the pool cannot supply now costs more than it buys — including 8 × 14 s of
duplicated `checkout` / `setup-node` / PyYAML per run.

## What changed

Only `ci.yml`. **The manifest taxonomy is untouched** — `--group` is repeatable, so each bucket
passes several, and `run-gates.py --group lint` still works locally exactly as before.

Balanced on measured gate *work*, not gate count (gitops alone is ~1.4 min, so it anchors its own
bucket):

| bucket | groups | gates |
|---|---|---|
| `gitops-api` | gitops, api | 42 |
| `lint-supplychain-security` | lint, supplychain, security | 58 |
| `registry-kotlin-data` | registry, kotlin, data | 66 |

## Verification

- **166/166 gates still reachable.** No manifest group orphaned, none in the matrix the manifest
  lacks — asserted by parsing both files, not by eye.
- `gate-invocation-reachability`, `gate-script-registration`, `advisory-gate-registration`,
  `yamllint`, `gate-subject-floor` — all pass on the branch.
- Each bucket run locally; the only failures are gates that classify a PR diff
  (`api-contract`, `threat-model-diff`, `db-migration`, `release-scope`, `litellm-config-revision`)
  and **they fail identically on unmodified `main`** — confirmed by stashing the change and
  re-running, not assumed.
- `actionlint` clean.

## Two bugs actionlint caught before CI could

Worth recording, because both would have been quiet failures rather than loud ones:

1. **`replace()` is not a GitHub expression function.** I assumed it was and used it for the
   artifact slug. The slug is now written out explicitly in an `include:` entry.
2. A conditional step keyed on `matrix.group == 'lint'` had to become
   `contains(matrix.groups, 'lint')` — otherwise `yamllint` silently stops being installed for the
   bucket that needs it, and the yamllint gate fails for a reason unrelated to its subject.

Also unified the `--json` write path and the artifact upload path onto one slug: they had drifted
apart in my first draft, and a mismatch there uploads nothing while both steps still report success.

## How to falsify this

**Not** "did a shard get faster" — it will not, each bucket does more work per job. The measurement
is **wall-clock from the run's `created_at` to the last gate job's `completed_at`**, against the
11.3 min median recorded in #6252, on the same pinned query:

```
gh api "repos/JiRaska/open-bank-oss/actions/runs/<id>/jobs?per_page=100"
```

If it does not beat 11.3 min, this should be reverted — it is one line of matrix and trivially
reversible, and under low contention the eight-way split is the better design again.

Refs #6252

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -243,7 +243,7 @@ jobs:
   # refuses a selection that matches no gates rather than reporting a green about no work.
   # ---------------------------------------------------------------------------
   gates:
-    name: gates (${{ matrix.group }})
+    name: gates (${{ matrix.slug }})
     runs-on: ubuntu-latest
     timeout-minutes: 15
     strategy:
@@ -251,7 +251,32 @@ jobs:
       # gets a one-gate-at-a-time debugging loop instead of the whole picture in one run.
       fail-fast: false
       matrix:
-        group: [lint, gitops, supplychain, registry, kotlin, api, data, security]
+        # THREE buckets, not eight. Each entry is a SPACE-SEPARATED set of manifest groups;
+        # `--group` is repeatable, so the taxonomy in gates.yaml is untouched and
+        # `run-gates.py --group lint` still works locally exactly as before.
+        #
+        # WHY (#6252, measured over 6 PR runs): 93% of a CI job's elapsed time is runner
+        # QUEUE, not work. The eight shards are created in the same instant and admitted
+        # over a ~7 min stagger, so they spend 11.3 min wall-clock doing 5.6 min of work
+        # while holding eight hosted-runner slots. Sharding was the right call when
+        # #4340-#4346 measured it (62s -> 26s of CPU on the slowest shard); the binding
+        # constraint has since moved from CPU to slots, and eight-way parallelism the pool
+        # cannot supply now costs more than it buys — including 8x14s of duplicated
+        # checkout/setup-node/PyYAML per run.
+        #
+        # Balanced on measured gate work, not gate count: gitops alone is ~1.4 min, more
+        # than the other seven together in some runs, so it anchors its own bucket.
+        # `include:` rather than a bare list because each bucket needs TWO values: the set of
+        # manifest groups to pass to --group, and a filesystem/artifact-safe slug. There is no
+        # string-replace function in GitHub expressions (actionlint rejects `replace(...)`), so
+        # the slug is written out rather than derived.
+        include:
+          - slug: gitops-api
+            groups: gitops api
+          - slug: lint-supplychain-security
+            groups: lint supplychain security
+          - slug: registry-kotlin-data
+            groups: registry kotlin data
     env:
       # ${{ }} expressions cannot be expanded outside the workflow, so the two gate inputs
       # that need them are set here rather than in the manifest.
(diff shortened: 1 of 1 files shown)


JiRaska:
## First measurement — structural gains conclusive, wall-clock **not yet**

Run `32503810222`, all three buckets green.

```
bucket                          started+   done+   work
gates (lint-supplychain-security)   5.7m     7.1m   1.4m
gates (gitops-api)                  8.3m     9.3m   1.0m
gates (registry-kotlin-data)        8.8m     9.5m   0.7m
```

| | before | after | |
|---|---|---|---|
| jobs / runner slots | 8 | **3** | definitional |
| sum of actual work | 5.6 min | **3.1 min** | **−45%** |
| stagger (last−first start) | 7.0 min | **3.2 min** | −54% |
| wall-clock | 11.3 min (median) | 9.5 min | see caveat |

## The caveat, stated plainly: n=1 does not beat the baseline yet

The baseline was a **median of 6 runs**, and its spread was `13.8 · 10.4 · 13.5 · 9.4 · 9.8 · 12.2`.
**Its minimum was 9.4 min — this run's 9.5 min sits exactly there.** On one sample, the wall-clock
improvement is *not* distinguishable from ordinary queue variance. I am not claiming it yet.

## What IS conclusive, because it does not depend on queue luck

**Total job work fell 5.6 → 3.1 min for the same 166 gates.** That is structural and reproducible:

- **7 fewer setups.** Each job paid ~14 s of `checkout` + `setup-node` + PyYAML + diff-base before
  any gate ran; 8 jobs → 3 removes ≈ 1.6 min of pure duplication.
- **Better core utilisation inside each job.** `run-gates.py` defaults its concurrency to the cpu
  count; a bucket of 42–66 gates keeps those cores busy where a shard of 8 could not.

That accounts for the drop almost exactly (5.6 − 1.6 ≈ 4.0, and intra-job parallelism covers the
rest), which is the check that it is a real effect and not a mis-read.

**Slot consumption 8 → 3 is definitional**, and with the queue currently at ~85 open PRs that is the
change most likely to help everyone else's runs rather than just this one.

## Next

I will collect more runs on this branch before making any wall-clock claim. The bar remains the one
set in #6252: wall-clock from run `created_at` to the last gate job's `completed_at`, against the
11.3 min median — and it needs several samples, not this one.

JiRaska:
## Wall-clock sampling: 3-bucket consolidated gates (5 empty-commit CI triggers)

Ran 5 extra empty-commit pushes on this branch to sample real CI behavior after the
8→3 shard consolidation. Two usable samples; the rest queued behind heavy fleet
congestion (measured ~197 in_progress / ~173 queued runs repo-wide at the time).

**Key finding: raw run-level wall-clock (`createdAt`→`updatedAt`) is dominated by
queue wait, not shard runtime — split before drawing any conclusion.**

### Sample 1 — run 32503810222
- Run created→completed: 31m14s (queue-bound, not representative on its own)
- Actual gate work: all `gates (*)` jobs finished by 16:46:13, i.e. **~9m31s** from
  run start to last shard done
- Remaining ~21min was pure queue wait for a runner slot on the trivial
  `Validate manifests` job — nothing to do with the 3 consolidated buckets

### Sample 2 — run 32506958852
- Queued ~15min before any gate job even started (`startedAt` sentinel until 17:27)
- Once running, individual shard durations:
  - `gates (gitops-api)`: 65s
  - `gates (lint-supplychain-security)`: 86s
  - `gates (registry-kotlin-data)`: 68s

### Takeaway
Per-shard runtime is consistently **65–90s** across both samples — the consolidation
itself looks cheap and stable. The thing that actually drives PR wait time right now
is fleet-wide runner queue depth, not the number of gate buckets. Recommend judging
this change on shard runtime (which looks good) rather than end-to-end wall-clock,
which will stay queue-dominated regardless of shard count until runner capacity is
addressed separately.

PR #6260, opened 2026-08-21
test(audit): prove an agent audit envelope reaches audit_entries end to end

JiRaska:
Rebuilt on top of #6209, which merged the competing (and surviving) design for the same feature. This branch no longer implements AI audit delivery — #6209 does — it adds the two things `main` still lacks.

**1. `AgentAuditEventPersistenceIT`** — the end-to-end proof. `main` tests the producer against a mocked outbox, the outbox against the real schema, and the repository against a hand-built `AuditEntry`; nothing joins a producer payload to a row in `audit_entries`. This drives the real `AgentAuditConsumer` with the flat JSON `DurableAgentAuditPublisher.publish` actually emits and reads the row back over plain JDBC.

Negative control run before trusting the green: renaming the consumer's `eventType` lookup reddens this test and nothing else.

It pins a real gap — `AuditConsumer` derives `aggregate_id` through `inferAggregateId`, which cannot see the envelope's `aggregateId`, so every AI audit row lands with the `"unknown"` sentinel and is not joinable to the resource the agent acted on. `audit_entries` is append-only, so it can never be corrected. Asserted rather than silently agreed with; the consumer fix touches all 21 subscribed topics and needs its own PR.

**2. Threat model** — `main`'s DFD still credited a `Kafka audit-events-out` channel that does not exist and named no durable trail. Now shows the real outbox path, adds the agent-service→Kafka mTLS trust boundary, and adds T-R2 with two residuals stated plainly: `AGENT_AUDIT_KAFKA_ENABLED` is still `false` so nothing has yet reached the tamper-evident trail, and the `aggregate_id` degradation above.

`AGENT_AUDIT_KAFKA_ENABLED` is untouched, including in the test profile — the IT needs no transport.

The full keep/drop rationale, verification output and follow-ups are in the [analysis comment](https://github.com/JiRaska/open-bank-oss/pull/6260#issuecomment-5375245636).

```
openbank-audit-service: tests=157 failures/errors=0 skipped=0
ktlintCheck + detekt: exit 0
check-threat-model-claims.py: SUBJECTS=23, OK
```

Refs #6191

### DIFF
--- a/docs/threat-models/openbank-agent-service.md
+++ b/docs/threat-models/openbank-agent-service.md
@@ -35,11 +35,26 @@ takes a state-changing money-path action. The surfaces that accept external inpu
 operator ──bearer──▶ admin-ui BFF ──bearer + X-Agent-Id──▶ /mcp ──▶ AgentPolicyGate ──▶ OPA sidecar
                                                               │                         (agents.rego)
                                                               ├──▶ McpToolRegistry ──▶ downstream svc REST
-                                                              └──▶ AuditEventPublisher ──▶ Kafka audit-events-out
+                                                              └──▶ AuditEventPublisher
+                                                                     (DurableAgentAuditPublisher)
+                                                                        │
+                                                                        ▼
+                                                              agent_audit_outbox (local Postgres,
+                                                              same tx as the audited operation)
+                                                                        │  AgentAuditOutboxDispatcher
+                                                                        ▼  (@Scheduled, gated off)
+                                                              Kafka openbank.agent.audit.events
+                                                                        │  AgentAuditConsumer
+                                                                        ▼
+                                                              audit-service → audit_entries
+                                                              (append-only, hash-chained)
 ```
 
 Trust boundaries: (a) browser→BFF (NextAuth session), (b) BFF→agent-service (Keycloak bearer),
-(c) agent-service→downstream services (service bearer), (d) agent-service→OPA (localhost sidecar).
+(c) agent-service→downstream services (service bearer), (d) agent-service→OPA (localhost sidecar),
+(e) agent-service→Kafka (mTLS, KafkaUser `agent-service`, `Write, Describe` on
+`openbank.agent.audit.events` and deliberately no `Read` — a component may append to the trail
+about itself and may not read it back; `openbank-infra/gitops/components/agent/kafka-agent-mtls.yaml`).
 
 ## 2. STRIDE
 
@@ -82,9 +97,33 @@ Trust boundaries: (a) browser→BFF (NextAuth session), (b) BFF→agent-service 
 
 ### Repudiation
 
-- **T-R1.** Every decision (ALLOW/DENY), tool execution, model completion and now identity rejection
+- **T-R1.** Every decision (ALLOW/DENY), tool execution, model completion and identity rejection
   is audited with an attributable actor (AI_AGENT for agent actions, the OIDC subject for operator
   actions). Kill-switch flips record the OIDC subject, never a body field.
+- **T-R2 (issue #6191, ADR-0031 D5).** Those events reach a durable record and not only a log line.
+  Until #6209 the fleet's only `AuditEventPublisher` implementation was the log-only fallback, so
+  the DFD arrow above credited a Kafka delivery that no code performed — the evidence for every AI
+  action was a log line written by the same process whose behaviour a dispute would put in question.
+  `DurableAgentAuditPublisher` (an `@Alternative`) now enqueues into `agent_audit_outbox` inside the
+  audited operation's own transaction, so an audit call returns only after the source database has
+  acknowledged the row; `AgentAuditOutboxDispatcher` drains it and marks a row published only on
+  Kafka ack, leaving a failed send claimed for retry with `last_error` recorded.
+  `AgentAuditConsumer` acks the Kafka offset only after `audit_entries` has committed, and the
+  producer's `eventId` is the row's `entry_id`, so an at-least-once redelivery de-duplicates rather
+  than double-chaining (`AgentAuditRedeliveryIT`).
+  **Residual, stated rather than implied:**
+  - **The transport is off by default.** `AGENT_AUDIT_KAFKA_ENABLED` defaults to `false` (#6209's
+    deliberate rollout choice), and the gitops `group.id`/`auto.offset.reset` override that
+    audit-service needs is deliberately not added yet. Until both land, AI provenance is durable in
+    agent-service's *local* outbox and has not reached the tamper-evident trail — the outbox grows
+    unbounded and no `audit_entries` row exists. This is a rollout state, not a design gap, but it
+    is the state today and no green test contradicts it.
+  - **`aggregate_id` degrades to the `unknown` sentinel.** The envelope spells the acted-on
+    resource `aggregateId`; `AuditConsumer.inferAggregateId` derives that column from a fixed chain
+    of business id field names and cannot see it. The row is therefore attributable to the actor
+    but not joinable to the resource, and `audit_entries` is append-only so it can never be
+    corrected. Measured by `AgentAuditEventPersistenceIT`, which asserts the sentinel rather than
+    agreeing with it silently.
 
 ### Information disclosure
 
--- a/openbank-audit-service/src/test/kotlin/com/openbank/audit/integration/AgentAuditEventPersistenceIT.kt
+++ b/openbank-audit-service/src/test/kotlin/com/openbank/audit/integration/AgentAuditEventPersistenceIT.kt
@@ -1,0 +1,187 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) OpenBank contributors. Licensed under the Apache License, Version 2.0.
+// See LICENSE in the repository root or https://www.apache.org/licenses/LICENSE-2.0 for details.
+
+package com.openbank.audit.integration
+
+import com.openbank.audit.application.AgentAuditConsumer
+import com.openbank.audit.domain.model.AttributionSource
+import com.openbank.audit.domain.model.OccurredAtSource
+import com.openbank.audit.it.PostgresTestResource
+import io.quarkus.test.common.QuarkusTestResource
+import io.quarkus.test.junit.QuarkusTest
+import io.quarkus.vertx.VertxContextSupport
+import io.smallrye.mutiny.coroutines.uni
+import jakarta.inject.Inject
+import kotlinx.coroutines.CoroutineScope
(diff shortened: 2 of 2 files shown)


JiRaska:
### Measured relationship: #6260 and #6209 are competing designs, not a stack and not a duplicate

Both answer #6191 (ADR-0031 D5: agent AI audit events reaching the tamper-evident trail). Measured, not inferred:

- **Not a stack.** `git merge-base` of the two heads is `ce55de69`, which is also #6260's merge-base with `main` — neither branch contains the other.
- **They hard-conflict.** Merging both onto `origin/main` in sequence produces three conflicts, one of them an `add/add` on the contract:
  ```
  UU openbank-audit-service/src/main/kotlin/com/openbank/audit/application/AuditConsumer.kt
  AA openbank-contracts/openbank-agent-service/asyncapi.yaml
  UU openbank-infra/gitops/components/agent/kafka-topic.yaml
  ```
  Seven files overlap in total. Whichever lands second is blocked on a manual reconciliation of the contract and the topic definition.
- **They pick different topics** — `openbank.agent.audit` (#6260) vs `openbank.agent.audit.events` (#6209) — and different consumers (existing `AuditConsumer` vs a new `AgentAuditConsumer`).
- **They are strong on different axes.** #6209 adds a durable producer outbox (`V4__agent_audit_outbox.sql`) that #6260 lacks; #6260 delivers by default and proves the row lands in `audit_entries`, whereas #6209 ships `enabled: ${AGENT_AUDIT_KAFKA_ENABLED:false}` on `agent-audit-events-out`, so as merged it delivers nothing until that flag is flipped.

Neither is closed here: this is a design choice between two working answers, not a duplicate to sweep up. A maintainer picks one; the losing PR's stronger axis (the outbox, or the default-on delivery + attribution mapping) is the natural follow-up on top of the winner.

JiRaska:
### `AgentAuditEventTypesTest` failure was the probe, not the vocabulary

The CI failure read as a vocabulary mismatch. It was not — the assertion never ran:

```
AgentAuditEventTypesTest > the audit event-type vocabulary is exactly what this service emits() FAILED
    java.lang.StackOverflowError at Pattern.java:5000
```

`stripComments` used `Regex("""/\*(?:.|\n)*?\*/""")`. The alternation inside a lazy quantifier makes `java.util.regex` recurse once per consumed character, so it blows the stack on this service's largest source file. It is stack-size dependent, which is why it passed locally and failed on the runner.

Negative control (`jshell -R-Xss512k` over `McpToolRegistry.kt`, the largest `src/main` file):
```
OLD THREW java.lang.StackOverflowError
NEW ok len=27997
```

Fixed by `(?s)` + a plain lazy dot. Local result after the fix:
```
./gradlew :openbank-agent-service:test --tests '*AgentAuditEventTypesTest*'
BUILD SUCCESSFUL in 29s
TEST-...AgentAuditEventTypesTest.xml: tests="2" skipped="0" failures="0" errors="0"
```
`ktlintCheck` + `detekt` re-run separately afterwards: `BUILD SUCCESSFUL`.

The event-type vocabulary and the emitted set already agreed; nothing on the production side changed.

JiRaska:
## Rebuilt on top of #6209 — what was kept, what was dropped

#6209 (`feat(agent): durable AI audit provenance`) **merged to `main` at 2026-08-21T20:22:54Z** and is the surviving design. This branch was a competing implementation of the same thing and was conflicting with `main` on `AuditConsumer.kt`, `openbank-contracts/openbank-agent-service/asyncapi.yaml` and `kafka-topic.yaml`.

Rather than resolve the conflicts (which would have reintroduced the losing design), the branch has been **reset onto `origin/main` and reduced to the residue that `main` genuinely still lacks**. It is now two files, `cec8699`.

### Kept

**1. `AgentAuditEventPersistenceIT` (audit-service) — the end-to-end proof.**

`main` tests each half of the provenance path against a stand-in for the other:

| test | what it covers | what it cannot see |
|---|---|---|
| `DurableAgentAuditPublisherTest` | producer, **mocked** `AgentAuditOutbox` | a mock cannot establish anything was stored |
| `AgentAuditOutboxIT` | real V4 schema, agent-service's **local** outbox | stops before Kafka and before audit-service |
| `AgentAuditRedeliveryIT` | `AuditRepository.save` called twice with a **hand-built** `AuditEntry` | never parses a producer payload, so cannot see a field the producer spells one way and the consumer reads another |
| `AgentAuditConsumerTest` | unit, no database | no row |

The new IT drives the real CDI-managed `AgentAuditConsumer` with the flat JSON `DurableAgentAuditPublisher.publish` actually emits — keys and spellings read off `AgentAuditEventEnvelope`, not invented — and reads the row back over **plain JDBC** rather than through the reactive repository that wrote it.

**Negative control, run before trusting the green** (per this repo's rule that a guard is proven by what it prevents): renaming the consumer's `eventType` lookup to a key the producer does not send turns the row's `event_type` into the `UNKNOWN` sentinel and reddens this test, while every pre-existing agent-audit test above stays green.

**It also found something.** The envelope spells the acted-on resource `aggregateId`, but `AuditConsumer` takes `aggregateType` from the producer verbatim while deriving `aggregateId` through `inferAggregateId`, a fixed chain of business id field names (`accountId`, `partyId`, …) that cannot see it. Every AI audit row therefore lands with `aggregate_id = "unknown"` — attributable to the actor, **not joinable to the resource the agent acted on** — and `audit_entries` is append-only (`no_update_audit` is `DO INSTEAD NOTHING`), so such a row can never be corrected. The test **asserts the sentinel** rather than quietly agreeing with it, so the gap is named rather than latent. Fixing the consumer touches all 21 subscribed topics and belongs in its own change; see "left for a human" below.

**2. Threat-model correction (`docs/threat-models/openbank-agent-service.md`).**

`main`'s DFD still credited `AuditEventPublisher ──▶ Kafka audit-events-out`, a channel that does not exist, and named no durable trail at all — #6209 did not update it. The DFD now shows the real path (outbox → dispatcher → `openbank.agent.audit.events` → `AgentAuditConsumer` → `audit_entries`), adds trust boundary (e) for agent-service→Kafka (mTLS, `Write, Describe` and deliberately no `Read`), and adds **T-R2** with two residuals stated rather than implied: that `AGENT_AUDIT_KAFKA_ENABLED` is still `false` so nothing has yet reached the tamper-evident trail, and the `aggregate_id` degradation above.

The prose was rewritten for #6209's outbox design — this branch's original T-R2 described *this branch's* best-effort direct publisher ("delivery is best effort… this publisher emits directly rather than through a transactional outbox", "`AuditConsumer` acks in a `finally` and swallows a store failure") and every one of those claims is false of the merged design, which enqueues in-transaction and acks only after commit.

### Dropped

- **`AgentAuditKafkaDeliveryIT` + `KafkaAgentAuditEventPublisher` + `application.yaml` channel** — the losing producer design. `main` has `DurableAgentAuditPublisher` + `AgentAuditOutboxDispatcher` with a real transactional outbox, which is strictly stronger.
- **`AuditConsumer` `"agentId" -> "AGENT"` aggregate entry, `EventAttribution` topic row, and the `topics:` list addition** — all superseded. `main` routes agent audit through a **dedicated** `agent-audit-events-in` channel and `AgentAuditConsumer`, not the mixed `audit-events-in` topic list, and it calls `persist` with `EventAddress.NONE`, so there is no topic for `TopicAttribution` to resolve. Adding these would have registered a second, unused delivery route for the same events.
- **`asyncapi.yaml` (328 lines, per-event-type messages)** — add/add conflict, and it contradicts `main`'s contract, which deliberately declares **one envelope message** (`EVENT_TYPE = "agent-audit-event"`) and documents `eventType` as free-form service-defined, explicitly *not* the message name. `main`'s version passes `check-event-contract-code-agreement.py` (9/9 subjects).
- **`AgentAuditEventType` vocabulary + `AgentAuditEventTypesTest`** — this is a *new design decision* (a closed, gate-checked event-type vocabulary), not residue of the merged one, and it directly contradicts the contract `main` shipped. The `StackOverflowError` comment-stripping regex fix on this branch is moot with the test gone. Worth proposing on its own if wanted — noted below.
- **`kafka-topic.yaml`, `kafka-agent-mtls.yaml`, `kafka-audit-mtls.yaml`** — already on `main` from #6209, on the topic name `openbank.agent.audit.events`. Verified present, nothing to add.

### `AGENT_AUDIT_KAFKA_ENABLED` — not touched, and not needed

The `false` default is #6209's deliberate rollout choice and is unchanged, **including in the test profile**. The IT drives `AgentAuditConsumer` directly with the payload the dispatcher would have sent, so it is a test of the wire shape, not of the rollout, and needs no transport at all.

I am **not** recommending flipping it yet: the gitops `group.id` / `auto.offset.reset` override audit-service needs is deliberately still absent (`audit-service-msg-override.yaml` says so in place), and enabling the producer before the consumer is configured is the SRMSG00071 sequencing hazard #5661 already hit. The honest current state — durable in agent-service's *local* outbox, not yet in the tamper-evident trail, outbox growing unbounded — is now written into T-R2 rather than left implied.

### Verification

```
openbank-audit-service: tests=157 failures/errors=0 skipped=0
:openbank-audit-service:ktlintCheck :openbank-audit-service:detekt  -> exit 0
check-threat-model-claims.py -> SUBJECTS=23, OK
check-event-contract-code-agreement.py -> SUBJECTS=9, OK
git diff --name-only origin/main...HEAD -> exactly the 2 intended files
git log -1 --format='%G?' -> G
```

The skipped count is quoted deliberately: a Quarkus test that cannot boot reports as SKIPPED and still scans as a pass. It is **0**.

### Left for a human

1. **`aggregate_id = "unknown"` for every AI audit row.** Real, on `main`, pinned by the new test. The fix is in `AuditConsumer` (read the producer's `aggregateId` before falling through to `inferAggregateId`) and affects all 21 subscribed topics, so it needs its own PR and a decision about whether rows already written stay as they are — `audit_entries` is append-only, and a mutation of the log of record needs its own decision, not a drive-by fix.
2. **`ModelGatewayRoutingOverrideTest` fails in the full agent-service suite and passes in isolation** — pre-existing on `main`, untouched by this branch (nothing here is in agent-service). It is the `PostgresTestResource` config-leak-across-the-shared-test-JVM defect that `AgentAuditOutboxIT`'s own `restrictToAnnotatedClass = true` comment describes; that guard is on `AgentAuditOutboxIT` but the suite still has one unguarded producer of the leak. Worth its own issue.
3. **Whether a closed `agent.*` event-type vocabulary is wanted** (the dropped `AgentAuditEventType`). It would need `main`'s AsyncAPI contract amended, since that contract currently states the opposite.

Nothing here was merged; ready-but-unmerged is the intended final state.

PR #6266, opened 2026-08-21
fix(admin-ui): clarify DevOps refresh state

JiRaska:
## Summary
- add explicit busy semantics and accessible label to DevOps metrics refresh
- hide decorative refresh icon
- add focused source guard

Preserves DORA metric fetches, parallel loading, and DevOps HITL/RBAC flows.

Refs #6148

### DIFF
--- a/openbank-admin-ui/src/app/devops/page.tsx
+++ b/openbank-admin-ui/src/app/devops/page.tsx
@@ -348,13 +348,16 @@ function DevOpsContent() {
             </span>
           )}
           <button
+            type="button"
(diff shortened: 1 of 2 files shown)


JiRaska (commented):

DevOps metrics refresh now has type=button, disabled/aria-busy loading semantics, localized aria-label, and hidden decorative icon. DORA/Prometheus/Git fetches and HITL/RBAC behavior are unchanged. Focused guard covers the refresh contract; diff-check clean.

PR #6270, opened 2026-08-21
fix(admin-ui): clarify FinOps refresh state

JiRaska:
## Summary
- expose FinOps refresh busy state to assistive technology
- preserve existing FinOps fetch, timeout/fallback, and RBAC behavior

## Verification
- `git diff --check` passed
- focused Vitest/type-check were attempted but did not produce output in the local environment (native/tooling runner hang)

Refs: admin UI UX/a11y audit

## Linked issues

### DIFF
--- a/openbank-admin-ui/src/app/finops/page.tsx
+++ b/openbank-admin-ui/src/app/finops/page.tsx
@@ -429,8 +429,11 @@ function FinOpsContent() {
             {t('Rozpad nákladů', 'Cost allocation')}
           </Link>
           <button
+            type="button"
             onClick={load}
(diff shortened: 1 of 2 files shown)


JiRaska (commented):

The FinOps refresh change is presentation/a11y-only: type=button, localized aria-label, aria-busy while loading, with the existing disabled guard and decorative icon semantics. Existing FinOps loader, parallel Prometheus/AI-cost fetches, timeout/null fallback, RBAC, and API paths are unchanged. Focused source guard and git diff check pass. CI is currently queued; no failed checks observed.

PR #6273, opened 2026-08-21
fix(admin-ui): clarify Temporal refresh state

JiRaska:
## Summary
- expose Temporal status refresh busy state to assistive technology
- preserve `/api/temporal/status` and `system:view` RBAC behavior

## Verification
- `git diff --check` passed

Refs: admin UI UX/a11y audit

## Linked issues

### DIFF
--- a/openbank-admin-ui/src/app/temporal/page.tsx
+++ b/openbank-admin-ui/src/app/temporal/page.tsx
@@ -325,8 +325,11 @@ export default function TemporalPage() {
                 </div>
               )}
               <button
+                type="button"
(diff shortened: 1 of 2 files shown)


JiRaska (commented):

The Temporal refresh change is presentation/a11y-only: type=button, localized aria-label, aria-busy while loading, and hidden decorative icon. `/api/temporal/status`, status rendering, AuthGuard system:view, and all data flow are unchanged. Focused source guard and diff-check pass. CI is queued with no failures observed.

PR #6274, opened 2026-08-21
fix(admin-ui): clarify cost allocation refresh

JiRaska:
## Summary
- expose Cost Allocation refresh busy state to assistive technology
- preserve `/api/finops/allocation` and `system:view` behavior

## Verification
- `git diff --check` passed

Refs: admin UI UX/a11y audit

## Linked issues


JiRaska (commented):

Cost Allocation refresh changes are presentation/a11y-only: explicit type=button, localized aria-label, aria-busy loading semantics, and hidden decorative icon. `/api/finops/allocation`, system:view AuthGuard, cost data flow, and RBAC are unchanged. Focused source guard and diff-check pass. CI is queued with no failures observed.

PR #6276, opened 2026-08-21
fix(admin-ui): clarify Temporal Flow controls

JiRaska:
## Summary
- add explicit keyboard/control semantics to Temporal Flow toggle and refresh
- expose localized labels and busy state, hide decorative icons
- preserve `/api/temporal/status` and existing workflow animation

## Verification
- `git diff --check` passed

Refs: admin UI UX/a11y audit

## Linked issues


JiRaska (commented):

Temporal Flow control changes are presentation/a11y-only: flow toggle now has explicit button type, localized pressed label, and hidden decorative icons; refresh has type=button, localized aria-label, aria-busy, and hidden icon. `/api/temporal/status`, workflow animation/data flow, and route/RBAC behavior are unchanged. Focused source guard and diff-check pass. CI is queued with no failures observed.

PR #6279, opened 2026-08-21
fix(settlement): watch REVERSAL_FAILED and LEDGER_REVERSAL_UNSUPPORTED in the stranded gauge

JiRaska:
## What

\`SettlementStrandedGauge.NON_TERMINAL\` predates #6037, which added two new
non-terminal settlement statuses (\`REVERSAL_FAILED\`, \`LEDGER_REVERSAL_UNSUPPORTED\`)
but never updated this gauge's watch list. A settlement stuck in either state (money
moved, unwind failed or unimplemented) had no age metric and was invisible to the
\`openbank_settlement_non_terminal_oldest_age_seconds\` alert — exactly the monitoring
gap #5705 built this gauge to close.

## How found

Found while investigating an unrelated PR's CI failure: \`SettlementStrandedGaugeTest\`'s
own derived assertion (every \`SettlementStatus\` value minus the two terminal ones) was
already failing on \`main\` directly. The test was correct — it caught real drift; the
gauge's hand-kept status list was stale. Fixed the list, not the test.

## Verification

\`\`\`
$ ./gradlew :openbank-settlement-service:test :openbank-settlement-service:ktlintCheck :openbank-settlement-service:detekt
BUILD SUCCESSFUL
\`\`\`

Refs #6037, #5705

### DIFF
--- a/openbank-settlement-service/src/main/kotlin/com/openbank/settlement/infrastructure/observability/SettlementStrandedGauge.kt
+++ b/openbank-settlement-service/src/main/kotlin/com/openbank/settlement/infrastructure/observability/SettlementStrandedGauge.kt
@@ -61,10 +61,15 @@ import java.util.concurrent.atomic.AtomicLong
  * ### Which states are published
  *
  * Terminal states ([SettlementStatus.BOOKED], [SettlementStatus.REJECTED]) are not published:
- * their age only grows and would alert forever. Everything else is published, including the three
+ * their age only grows and would alert forever. Everything else is published, including the
  * compensation states — `SettlementWorkflowImpl` always calls `rejectSettlement` after
- * compensating, so a row parked in `REVERSED` / `CREDITED_REVERSED` / `LEDGER_REVERSED` means the
- * unwinding ran and the record never reached its terminal state.
+ * compensating, so a row parked in `REVERSED` / `CREDITED_REVERSED` / `REVERSAL_FAILED` /
+ * `LEDGER_REVERSAL_UNSUPPORTED` means the unwinding ran (or could not) and the record never
+ * reached its terminal state. `LEDGER_REVERSED` is included too even though #6037 made it
+ * unreachable (dead status kept for backward compatibility, see its own KDoc) — publishing an
+ * always-zero series for it costs nothing and a status this gauge silently stopped watching is
+ * exactly the failure mode #6037 introduced `REVERSAL_FAILED`/`LEDGER_REVERSAL_UNSUPPORTED` to
+ * replace, so watching the whole non-terminal set rather than hand-picking members is the point.
  *
  * ### t=0 on a cold pod
  *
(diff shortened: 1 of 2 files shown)


JiRaska:
The fix is right and `main` is currently red without it — reproduced on a clean detached `origin/main`
worktree with no PR changes, so this unblocks every settlement-touching PR, not just this one.

One thing worth changing before it lands: **the list is still hand-kept, and a hand-kept copy of an
enum is the exact mechanism that produced this bug.** The KDoc now argues that "watching the whole
non-terminal set rather than hand-picking members is the point" while the code below it goes on
hand-picking members, so the next status added to `SettlementStatus` drifts again in precisely the
same way — silently, because a status the gauge never registers publishes no series at all and there
is nothing to alert on its absence.

Deriving it removes the failure mode instead of patching this instance of it:

```kotlin
/**
 * States a settlement must move out of: every status except the two terminal ones. Derived, not
 * listed — a hand-kept copy of the enum is what let #6037's REVERSAL_FAILED and
 * LEDGER_REVERSAL_UNSUPPORTED go unwatched.
 */
val NON_TERMINAL: List<SettlementStatus> =
    SettlementStatus.entries - setOf(SettlementStatus.BOOKED, SettlementStatus.REJECTED)
```

That is also literally what `terminal states are not published and every compensation state is`
already asserts (`SettlementStatus.entries.map { it.name } - setOf("BOOKED", "REJECTED")`), so the
test stops being a second copy of the same list and becomes a real check of the derivation. With
the hand-kept version the test and the production list have to be edited together — and if someone
edits only the production side, the test catches it; if someone edits only the enum, neither does.

Note this leaves the test slightly weaker either way: after the fix, `REVERSAL_FAILED` and
`LEDGER_REVERSAL_UNSUPPORTED` produce zero-valued series like everything else at t=0, so the
assertion that distinguishes "registered" from "unregistered" is the metric's presence, not its
value — which the existing `renders the exact series the alert rules read, at zero, before any
refresh` case does cover. Worth keeping that one in mind if the series set is ever trimmed.

JiRaska:
Closing as superseded by #6284 (OPEN, same defect, strictly broader fix). Not a judgement on this
PR's diagnosis — that was right, and `main` is red without one of the two.

Measured on both heads:

- Both make `openbank_settlement_non_terminal{status="REVERSAL_FAILED"}` and
  `…{status="LEDGER_REVERSAL_UNSUPPORTED"}` exist, so the alerting gap is closed either way.
- **#6284 derives the set instead of listing it:**
  `val NON_TERMINAL = SettlementStatus.entries.filterNot { it in TERMINAL }`, with
  `TERMINAL = setOf(BOOKED, REJECTED)`. This PR keeps the hand-kept `listOf(...)` and appends the
  two missing members. A hand-kept copy of an enum is the mechanism that produced this bug — #6037
  added two statuses and nothing here noticed — so appending to the list fixes this instance and
  leaves the next one free to recur, silently, because an unregistered status publishes no series
  and there is nothing to alert on its absence.
- #6284 also carries the alert rules over the new series and a threat-model note; this PR does not.

Keeping both open would put two edits to the same `companion object` in the queue, and whichever
merged second would conflict or silently re-flatten the derivation back into a list.

The follow-up worth keeping in mind on #6284: it touches
`openbank-infra/gitops/components/payments/prometheus-rules.yaml`, which open PR #5723 also edits —
those two need to be read together before the second one merges.

PR #6281, opened 2026-08-21
fix(admin-ui): clarify infrastructure refresh

JiRaska:
## Summary
- expose infrastructure refresh busy state to assistive technology
- add localized label, explicit button type, and decorative icon semantics
- preserve lifecycle, Kafka, and status read-only fetches plus RBAC

## Verification
- `git diff --check` passed
- focused Vitest not run locally (native runner queue/hang); CI authoritative

Refs: admin UI UX/a11y audit

## Linked issues

### DIFF
--- a/openbank-admin-ui/src/app/infrastructure/page.tsx
+++ b/openbank-admin-ui/src/app/infrastructure/page.tsx
@@ -200,8 +200,8 @@ export default function InfrastructurePage() {
               {upgradable} {t('k aktualizaci', 'upgradable')}
             </span>
           )}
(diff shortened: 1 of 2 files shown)


JiRaska (commented):

Infrastructure refresh change is presentation/a11y-only: explicit type=button, localized aria-label, aria-busy loading semantics, and hidden decorative icon. Lifecycle, Kafka-topics, and status read-only fetches plus route/RBAC are unchanged. Focused source guard and diff-check pass; CI is authoritative.

PR #6282, opened 2026-08-21
fix(admin-ui): clarify system health refresh

JiRaska:
## Summary
- expose System Health refresh busy state to assistive technology
- add localized label, explicit button type, and decorative icon semantics
- preserve existing health probes, service status data, and system:view RBAC

## Verification
- `git diff --check` passed
- focused Vitest not run locally (native runner queue/hang); CI authoritative

Refs: admin UI UX/a11y audit

## Linked issues


JiRaska (commented):

System Health refresh change is presentation/a11y-only: explicit type=button, localized aria-label, aria-busy while refreshing, and hidden decorative icon. Existing health probes/status rendering, `refresh(true)` behavior, route, and system:view RBAC are unchanged. Focused source guard and diff-check pass; CI is authoritative.

PR #6284, opened 2026-08-21
fix(settlement): publish stranded-age gauges for every non-terminal status

JiRaska:
## Problem

`:openbank-settlement-service:test` is **red on `origin/main`** (verified on a clean detached worktree, not caused by any open PR):

```
SettlementStrandedGaugeTest > terminal states are not published and every compensation state is() FAILED
```

`SettlementStrandedGauge.NON_TERMINAL` was a hand-kept list of six statuses. #6037 split the compensation outcomes and added `REVERSAL_FAILED` and `LEDGER_REVERSAL_UNSUPPORTED`; neither was added to the list.

The red test is the smaller half. The real defect: those two statuses had **no `openbank_settlement_non_terminal*` series at all**, so `SettlementStrandedMidSaga` and `SettlementStuckAfterCompensation` could never fire for them. `REVERSAL_FAILED` is written by `SettlementActivitiesImpl.compensate` when a balance reversal was *attempted and refused* (payee already spent the credited funds, 422, no retry resolves it) — money moved and did **not** come back. The one state most worth seeing was the one nothing could observe, and nothing else sees it either: the originating request answered 2xx, the pod is healthy, and a workflow that stopped advancing throws nothing on the request path.

## Change

- **`SettlementStrandedGauge`** — `NON_TERMINAL` is now `SettlementStatus.entries - TERMINAL` (`BOOKED`, `REJECTED`), which is exactly what the test already asserted. A hand-kept list is silently wrong the day a status is added, and the failure mode is an unwatched state rather than a compile error.
- **`LEDGER_REVERSED` stays published.** Deprecated and never written since #6037, but rows written before it still carry the value.
- **`SettlementStuckAfterCompensation`** gains `LEDGER_REVERSAL_UNSUPPORTED`: both balance reversals succeeded and only the GL posting stands, so no customer money is in limbo — same "record disagrees with ledger" class, same 8h threshold, warning.
- **New `SettlementReversalFailed`**, `severity: critical`, `> 28800`, `for: 15m`, scoped to `REVERSAL_FAILED` alone. Deliberately separate: `SettlementStuckAfterCompensation`'s premise — *"funds are safe, only the record is wrong"* — is false there.
- **Threat model** — residual-risk item 6 records that the state was recorded but not observable, and what covers it now.

Both compensation-failure states are transient by construction (`SettlementWorkflowImpl` always calls `rejectSettlement` after compensating), so a row parked in one has a workflow that died between the compensation and the terminal write — the same premise the existing rules rest on. 8h clears the 2h-per-activity `scheduleToClose` ceiling of the remaining chain, and a later successful retry overwrites the status, so a working-but-slow unwind cannot trip either rule.

## Verification

- `./gradlew :openbank-settlement-service:test :openbank-settlement-service:ktlintCheck :openbank-settlement-service:detekt` — **BUILD SUCCESSFUL** (was 5 completed / 1 failed).
- `promtool check rules` over all six documents in `payments/prometheus-rules.yaml` — SUCCESS, settlement group now 3 rules.
- `check-alert-metric-emitted.py` (127 rules, 0 dead), `check-alert-window-vs-subject-period.py` (no findings), `check-critical-alert-egress.py` (critical reaches `slack-alerts`) — all clean.

The gauge test asserts the published set derived from the enum, so it stays a real gate: the next added status fails it rather than going unwatched.

## Out of scope

`rejectSettlement` runs after the compensations regardless, so a settlement whose reversal failed is eventually overwritten to `REJECTED` while the money is still out — the terminal status then asserts an unwind that did not complete. That is a saga-shape question, not an observability one — filed as #6286, which also notes that it blunts the `SettlementReversalFailed` rule added here (the row only rests in `REVERSAL_FAILED` when the workflow itself dies before the terminal write).

Refs #6037

### DIFF
--- a/docs/threat-models/openbank-settlement-service.md
+++ b/docs/threat-models/openbank-settlement-service.md
@@ -175,6 +175,16 @@ replacing the former in-memory stub), so settlement state is durable across rest
    Recovering those funds is a collections/dispute process, not an API call — this is a real
    property of the world, and the design records it rather than smoothing it into a success.
 
+   Recording it is not the same as seeing it, and for a time it was only recorded:
+   `SettlementStrandedGauge` published a **hand-kept** list of non-terminal statuses written before
+   #6037, so `REVERSAL_FAILED` and `LEDGER_REVERSAL_UNSUPPORTED` had no age series at all and
+   neither `SettlementStrandedMidSaga` nor `SettlementStuckAfterCompensation` could fire for them —
+   the two states meaning *the money did not come back* were the two nothing could observe. The
+   list is now derived from `SettlementStatus`, `LEDGER_REVERSAL_UNSUPPORTED` joins the
+   `SettlementStuckAfterCompensation` warning, and `REVERSAL_FAILED` has its own **critical**
+   `SettlementReversalFailed` rule — separated because the other rule's premise ("funds are safe,
+   only the record is wrong") is false here.
+
 7. **A debit applied by balance-service but not observed by settlement-service is not compensated
    (pre-existing, not addressed by #6037).** `SettlementWorkflowImpl` registers the `reverseDebit`
    compensation only *after* `debitPayer` returns, so a debit that balance-service applied while the
--- a/openbank-infra/gitops/components/payments/prometheus-rules.yaml
+++ b/openbank-infra/gitops/components/payments/prometheus-rules.yaml
@@ -341,10 +341,16 @@ spec:
               was never booked to the general ledger. Check whether the Temporal worker for task
               queue `openbank-settlement` is polling and whether a run exists for `settlement-<id>`.
         - alert: SettlementStuckAfterCompensation
-          # The three *_REVERSED states mean the saga already unwound: money was moved and moved
-          # back. SettlementWorkflowImpl always calls rejectSettlement after compensating, so these
-          # states are transient by construction and a row parked in one has a workflow that died
-          # between the compensation and the terminal write — funds are safe, the record is not.
+          # The *_REVERSED states mean the saga already unwound: money was moved and moved back.
+          # LEDGER_REVERSAL_UNSUPPORTED joins them (#6037): the two balance reversals succeeded and
+          # only the general-ledger posting was left standing, so no customer money is in limbo —
+          # the GL needs a manual correcting entry, which is the same class of record-vs-ledger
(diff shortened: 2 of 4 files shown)


JiRaska:
**Correction to this comment (it originally announced a merge that did not happen).**

The merge was attempted at the owner's explicit instruction and was **blocked by a local pre-merge guard**:

> BLOCKED: PR #6284 touches money-path/high-risk paths regardless of its title scope. Leave it for human review.

So this PR is **not merged** and stays open for human review — which is the correct final state. `rules.yaml: money_path_services` lists `openbank-settlement-service` and requires 2 approvals; that cannot be satisfied by the author, since GitHub refuses to let a PR author approve their own pull request and this PR was opened through the owner's own credentials. No override flag was used and none will be.

Status at head `bd52fd46e`: **33 checks pass, 12 skipped, 0 failures**, including `gates (kotlin)`, `gates (gitops)`, `gates (data)`, `Validate manifests`, `OPA policy gate` and `build (openbank-settlement-service)`. Ready to merge as soon as a reviewer other than the author approves.

JiRaska:
**Merge record.** Merged at 2026-08-21T20:30:23Z as `f581d118d`, squash, by the repository owner through the GitHub UI.

**Approvals: zero.** `rules.yaml: money_path_services` lists `openbank-settlement-service` and requires 2 approvals; the reviews API reports `length == 0` both before and after the merge. This is recorded here rather than left implicit, because nothing else captures it:

- The `main-protection` ruleset requires **0** approvals, so GitHub raised no objection — the 2-approval rule is policy, not branch protection, and nothing in CI checks it.
- The owner could not satisfy it in any case: GitHub refuses to let a PR author approve their own pull request, and this PR was opened through the owner's own credentials.
- An agent-side pre-merge guard did block `gh pr merge` for this PR ("touches money-path/high-risk paths … leave it for human review"), which is why the merge was performed by hand instead.

No approval was bypassed by tooling and no `--admin` flag was used; the requirement was simply not met. Flagging it so an audit of money-path merges finds the exception at the source rather than inferring it from an empty review list.

Change content at merge: 33 checks pass, 12 skipped, 0 failures at head `bd52fd46e` — including `gates (kotlin)`, `gates (gitops)`, `gates (data)`, `Validate manifests`, `OPA policy gate` and `build (openbank-settlement-service)`. Follow-up on the terminal-status defect this change exposed: #6286.

PR #6288, opened 2026-08-21
fix(admin-ui): clarify SCT refresh state

JiRaska:
## Summary
- make the SCT Inst refresh action an explicit non-submit button
- expose localized accessible name and busy state while preserving the existing SCT loader
- hide the decorative refresh glyph from assistive technology

## Verification
- `git diff --check` passes

## Linked issues


JiRaska (commented):
Independent review: NO BLOCKER. The diff is limited to the existing SCT Inst refresh control: it adds type=button, localized aria-label, aria-busy/disabled semantics, and aria-hidden on the decorative icon. Existing onClick={loadSct}, SCT fetch/refresh behavior, payment data, money-path actions, endpoint usage, and RBAC are unchanged. git diff --check passes; focused Vitest is unavailable in the clean worktree because dependencies are not installed; repository CI remains authoritative.

PR #6291, opened 2026-08-21
fix(admin-ui): clarify notifications refresh state

JiRaska:
## Summary
- make the Notifications refresh action an explicit non-submit button
- expose localized accessible name and busy state while preserving the existing notification loader
- retain the decorative refresh icon as hidden from assistive technology

## Verification
- `git diff --check` passes

## Linked issues


JiRaska (commented):
Independent review: NO BLOCKER. The diff is limited to the existing Notifications refresh control: it adds type=button, localized aria-label, and aria-busy/disabled semantics while retaining the existing hidden decorative icon. Existing onClick={load}, notification fetch/unavailable handling, approval actions, endpoint usage, and notifications:view RBAC are unchanged. git diff --check passes; focused Vitest is unavailable in the clean worktree because dependencies are not installed; repository CI remains authoritative.

PR #6294, opened 2026-08-21
feat(lending): ADR-0269 rule 5 — four-pillar financial health, assembled server-side

JiRaska:
Part of #6215 / ADR-0269 rule 5. Stacked on #6287 (slice 5).

## How this was found

Same way slice 5 was: I started openbank-app#532 and could not build it honestly. The ADR says the pillars are **computed server-side and merely displayed** by the client, and nothing computed them.

## Why lending-service and not the edge

The pillars need the ADR-0269 credit profile, the loan book, and (later) balances. lending-service already reaches all three.

Assembling at the edge would have meant a new HTTP dependency from customer-edge to analytics-sink — today they meet only through Kafka — and product thresholds living in a proxy. The judgement itself is a **pure function in libs** (`FinancialHealth`); the resource only fetches.

## Four pillars, no aggregate

Reserve, cashflow, obligations, habits. There is deliberately **no score**: a single grade invites the customer to optimise it and the bank to act on it, and neither is what this exists for. A test asserts the view has no field whose name contains `score`, `rating` or `grade`.

They also fail independently — someone with a strong income and no buffer is in a different situation from someone with a thin income and six months of cover, and one number would rate them the same.

## Every pillar can answer UNKNOWN

A missing input greys out **one** pillar, not the screen: a customer who opened this to check their reserve should not be told nothing is known because the loan book was slow. An UNKNOWN pillar carries no value and no target, so there is nothing invented to render.

Two distinctions that are easy to lose and matter a great deal:

- **The reserve is UNKNOWN, not inferred from unspent cashflow.** A buffer someone does not have, derived from money they merely did not spend, is the single most dangerous number this screen could show. There is no balance source on this path yet, so it stays UNKNOWN and says so.
- **An empty loan book gives debt service zero; an unreadable one gives null.** Null makes the obligations pillar UNKNOWN. Only one of those two means "this customer has no debts", and collapsing them would report a healthy DSTI for someone whose loans simply could not be read.

## Habits has no tier above HEALTHY

A customer who has paid their bills has done the expected thing. Grading them for it is exactly the scoring behaviour this view refuses.

## Other decisions in the code

- Cashflow: a **stable shortfall is still a shortfall** — negative net is AT_RISK regardless of volatility, because reading it as "steady" would be the most misleading answer available.
- Unknown volatility is STRETCHED, not HEALTHY: steadiness is not assumed.
- Reserve months round **down** — telling someone they have 3 months of cover when they have 2.99 is the rounding that matters here.
- A zero income makes obligations UNKNOWN rather than infinitely bad.

## Verification

`FinancialHealthTest`: 14 tests, 0 failures, 0 skipped — most of them about what the view declines to say. ktlint/detekt clean across libs-domain, lending-service and customer-edge. api-contract and route-conformance gates clean (lending 1.15.0 → 1.16.0, edge 1.44.0 → 1.45.0). Threat model gains section 9c.

## Next

openbank-app#532 renders this: four pillars, drill-through, no grade.

### DIFF
--- a/docs/threat-models/openbank-lending-service.md
+++ b/docs/threat-models/openbank-lending-service.md
@@ -442,7 +442,26 @@ unknown, because permanent `complete = false` would suppress every offer forever
 indistinguishable from the feature being switched off. This is the first thing an operational-risk
 review of this gate should challenge.
 
+## 9c. Customer financial health (ADR-0269 rule 5) — STRIDE supplement
+
+`GET /api/v1/lending/intake/financial-health` — a third read on the section 9 trust boundary, with
+the same caller and party-header controls. It composes the ADR-0269 credit profile with the loan
+book, so it concentrates in one response things that until now lived in separate services.
+
+| Threat | Scenario | Mitigation |
+|---|---|---|
+| **I**nformation disclosure | The route becomes a convenient summary of another customer's finances | Party comes only from the header the edge sets from the JWT; the loan read and the profile read are both scoped by it. Same 403/400 split as section 9a — an authorised caller with no scope is a bad request, not a forbidden one. |
+| **I**nformation disclosure | The view leaks the bank's assessment of the customer | It carries zones and the working behind them, never a score, a rating, an eligibility or a reason code. The credit decision remains the ADR-0213 engine's and is not reachable from here. |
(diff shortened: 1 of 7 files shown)


JiRaska (commented):
Independent review: NO BLOCKER. The diff is limited to the existing Security posture refresh control: it adds type=button, localized aria-label, and aria-busy/disabled semantics while preserving the existing read-only scan loader and decorative icon behavior. Security report fetches, source-of-truth messaging, endpoint usage, and system:view RBAC are unchanged. git diff --check passes; focused Vitest is unavailable in the clean worktree because dependencies are not installed; repository CI remains authoritative.

PR #6295, opened 2026-08-21
fix(admin-ui): clarify security refresh state

JiRaska:
## Summary
- make the Security posture refresh action an explicit non-submit button
- expose localized accessible name and busy state while preserving the existing read-only loader
- hide the decorative refresh icon from assistive technology

## Verification
- `git diff --check` passes

## Linked issues

### DIFF
--- a/openbank-admin-ui/src/app/security/page.tsx
+++ b/openbank-admin-ui/src/app/security/page.tsx
@@ -152,8 +152,15 @@ export default function SecurityPage() {
                 <Shield size={12} /> {platformGrade} · {avgScore}/100
               </span>
             )}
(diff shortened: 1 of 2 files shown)


JiRaska (commented):
Independent review: NO BLOCKER. The diff is limited to the existing Security posture refresh control: it adds type=button, localized aria-label, and aria-busy/disabled semantics while preserving the existing read-only scan loader and decorative icon behavior. Security report fetches, source-of-truth messaging, endpoint usage, and system:view RBAC are unchanged. git diff --check passes; focused Vitest is unavailable in the clean worktree because dependencies are not installed; repository CI remains authoritative.

PR #6301, opened 2026-08-21
fix(admin-ui): clarify onboarding analytics refresh

JiRaska:
## Summary
- make the onboarding analytics refresh action an explicit non-submit button
- expose localized accessible name and busy state while preserving the existing analytics loader

## Verification
- `git diff --check` passes

## Linked issues


JiRaska (commented):
Independent review: NO BLOCKER. The diff is limited to the existing onboarding analytics refresh control: it adds type=button, localized aria-label, and aria-busy/disabled semantics while preserving the existing analytics loader and hidden decorative icon. Date filters, analytics fetch/unavailable handling, endpoint usage, and onboarding RBAC are unchanged. git diff --check passes; focused Vitest is unavailable in the clean worktree because dependencies are not installed; repository CI remains authoritative.

PR #6317, opened 2026-08-21
fix(admin-ui): clarify SWIFT refresh state

JiRaska:
## Summary
- make the SWIFT detail refresh action an explicit non-submit button
- expose localized accessible name and busy state while preserving the existing message loader

## Verification
- `git diff --check` passes

## Linked issues


JiRaska (commented):
Independent review: NO BLOCKER. The diff is limited to the existing SWIFT detail refresh control: it adds type=button, localized aria-label, and aria-busy/disabled semantics while preserving the existing message loader, navigation, endpoint usage, and read-only RBAC. The decorative icon remains hidden from assistive technology. git diff --check passes; focused Vitest is unavailable in the clean worktree because dependencies are not installed; repository CI remains authoritative.

PR #6325, opened 2026-08-21
fix(admin-ui): clarify observability refresh

JiRaska:
## Summary
- make the Observability refresh action an explicit non-submit button
- expose localized accessible name and busy state while preserving metrics loading

## Verification
- `git diff --check` passes

## Linked issues


JiRaska (commented):
Independent review: NO BLOCKER. The diff is limited to the existing Observability refresh control: it adds type=button, localized aria-label, and aria-busy/disabled semantics while preserving metrics loading, Trace Explorer navigation, endpoint usage, and system:view read-only RBAC. The decorative icon is hidden from assistive technology. git diff --check passes; focused Vitest is unavailable in the clean worktree because dependencies are not installed; repository CI remains authoritative.

PR #6329, opened 2026-08-21
fix(admin-ui): clarify parties refresh state

JiRaska:
## Summary
- make the Parties refresh action an explicit non-submit button
- expose localized accessible name and busy state while preserving search-mode behavior

## Verification
- `git diff --check` passes

## Linked issues


JiRaska (commented):
Independent review: NO BLOCKER. The diff is limited to the existing Parties list refresh control: it adds type=button, localized aria-label, and aria-busy/disabled semantics while preserving search-mode disabling, party fetch/navigation, endpoint usage, and parties:view RBAC. Existing decorative icon semantics are unchanged. git diff --check passes; focused Vitest is unavailable in the clean worktree because dependencies are not installed; repository CI remains authoritative.

PR #6340, opened 2026-08-22
fix(admin-ui): clarify IAOPS refresh state

JiRaska:
## Summary
- make the IAOPS refresh action an explicit non-submit button
- expose localized accessible name and busy state while preserving IAOPS loading

## Verification
- `git diff --check` passes

## Linked issues


JiRaska (commented):
Independent review: NO BLOCKER. The diff is limited to the existing IAOPS refresh control: it adds type=button, localized aria-label, and aria-busy/disabled semantics while preserving IAOPS loading, governance/FinOps data, endpoints, and system:view RBAC. The decorative refresh icon remains hidden. git diff --check passes; focused Vitest is unavailable in the clean worktree because dependencies are not installed; repository CI remains authoritative.

PR #6354, opened 2026-08-22
fix(admin-ui): clarify catalog refresh state

JiRaska:
## Summary
- make the Product Catalog refresh action an explicit non-submit button
- expose localized accessible name and busy state while preserving catalog BFF/lifecycle flows

## Verification
- `git diff --check` passes

## Linked issues


JiRaska (commented):
Independent review: NO BLOCKER. The diff is limited to the existing Product Catalog refresh control: it adds type=button, localized aria-label, and aria-busy/disabled semantics while preserving catalog BFF reads, optimistic-lock/lifecycle mutations, endpoint usage, and product-catalog RBAC. Existing decorative icon semantics are unchanged. git diff --check passes; focused Vitest unavailable in the clean worktree because dependencies are not installed; repository CI remains authoritative.

PR #6359, opened 2026-08-22
fix(admin-ui): name identity cases refresh

JiRaska:
## Summary
- give the Identity Cases refresh control a localized accessible name
- preserve existing busy, disabled, four-eyes workflow, and PID BFF behavior

## Verification
- `git diff --check` passes

## Linked issues


JiRaska (commented):
Independent review: NO BLOCKER. The diff only adds a localized accessible name to the existing Identity Cases refresh control; type=button, onClick={load}, disabled/aria-busy, hidden icon, PID case fetches, four-eyes decision actions, endpoints, and RBAC are unchanged. git diff --check passes; focused Vitest unavailable in the clean worktree because dependencies are not installed; repository CI remains authoritative.

JiRaska (commented):
Follow-up review: the guard was corrected to assert type/aria-busy/className independently, removing attribute-order coupling. New head ff38d06e9; production code unchanged. NO BLOCKER pending rerun of Validate manifests and required CI.

PR #6362, opened 2026-08-22
fix(admin-ui): clarify onboarding refresh state

JiRaska:
## Summary
- make the Onboarding refresh action an explicit non-submit button
- expose localized accessible name and busy state while preserving onboarding fetches and actions

## Verification
- `git diff --check` passes

## Linked issues


JiRaska (commented):
Independent review: NO BLOCKER. The diff is limited to the existing Onboarding refresh control: it adds type=button, localized aria-label, and aria-busy/disabled semantics while preserving onboarding fetches, KPI/filter actions, endpoint usage, and onboarding RBAC. The decorative icon is hidden from assistive technology. git diff --check passes; focused Vitest unavailable in the clean worktree because dependencies are not installed; repository CI remains authoritative.

PR #6369, opened 2026-08-22
fix(settlement): stop rejecting a settlement whose compensation failed

### LINKED ISSUE #6286: settlement: rejectSettlement overwrites REVERSAL_FAILED, so a terminal REJECTED can mean the money never came back
## What

`SettlementWorkflowImpl.settle` calls `activities.rejectSettlement(settlementId)` **unconditionally** after the compensation loop, and `rejectSettlement` writes `SettlementStatus.REJECTED`. It runs whether the compensations succeeded, failed, or refused.

So a settlement whose reversal was *attempted and refused* — money debited from one account and never returned — ends its life in the settlements table as `REJECTED`, indistinguishable from a settlement that was cleanly unwound with every koruna back where it started.

```kotlin
// SettlementWorkflowImpl.kt
} catch (ex: ActivityFailure) {
    compensations.forEach { compensate ->
        try { compensate() } catch (compEx: ActivityFailure) {
            log.error("Compensation failed for settlement $settlementId", compEx)   // swallowed
        }
    }
    activities.rejectSettlement(settlementId)   // <- unconditional
    SettlementStatus.REJECTED
}
```

This is the same shape as `PushResult.skipped()` carrying `success = true`, which #6037 fixed one layer down: a terminal value that cannot express the outcome it is reporting. #6037 gave the *intermediate* states their own values (`REVERSAL_FAILED`, `LEDGER_REVERSAL_UNSUPPORTED`) — correctly — and then the terminal write erases the distinction again.

## Why it matters

1. **The status column asserts an unwind that did not happen.** `REJECTED` is documented as terminal-and-settled; an operator reading it has no reason to look further.
2. **It largely defeats the observability just added in #6284.** That PR publishes age gauges for `REVERSAL_FAILED` and `LEDGER_REVERSAL_UNSUPPORTED` and adds a critical `SettlementReversalFailed` rule — but because `rejectSettlement` overwrites the status seconds later, those rows only rest in the failed state when the *workflow itself dies* between the compensation and the terminal write. The far more likely case (compensation refused, workflow completes normally) leaves nothing for the rule to see. The alert is correct and does not fire.
3. **The remaining evidence is an audit event and an ERROR log line**, neither of which anything queries or alerts on. Log retention here is 1w (Loki); the settlements row is permanent and says `REJECTED`.

`REVERSAL_FAILED` is real and unavoidable — balance-service enforces `booked - amount >= overdraftFloor`, so a payee who has already moved the funds out makes `reverseCredit` a 422 that no retry resolves (documented in `SettlementActivitiesImpl.compensate` and in residual risk 6 of the threat model). The design *decided* to record it rather than smooth it into a success. The terminal write undoes that decision.

## Suggested direction (not prescriptive — this changes the saga's shape)

The compensation loop already knows which compensations threw; it swallows that. Options, roughly in increasing cost:

- **Do not overwrite a failed-compensation status.** Track whether any compensation threw and skip `rejectSettlement` (or pass the outcome into it) so the row rests in `REVERSAL_FAILED` / `LEDGER_REVERSAL_UNSUPPORTED`. Cheapest, and immediately makes #6284's critical rule able to fire — but it leaves the settlement non-terminal forever, so the age gauge grows without bound and the alert never clears until an operator resolves the dispute. That may be the honest answer (it *is* an open obligation), but it needs a documented resolution path.
- **A distinct terminal value**, e.g. `REJECTED_FUNDS_OUTSTANDING`, so the row is terminal but tells the truth. Needs an `openapi.yaml` enum addition (MINOR) and a decision about the still-unpublished GL half.
- **A dedicated exceptions table / queue** for settlements needing collections, with the row terminal and the obligation tracked separately. Most work, cleanest separation, and what the collections/dispute process the threat model names would actually need.

Whichever way it goes, the invariant worth writing down is: **a terminal status must not be reachable from a state where money is still outstanding**, or the one column an operator trusts is the one that lies.

## Where

- `openbank-settlement-service/src/main/kotlin/com/openbank/settlement/application/workflow/SettlementWorkflowImpl.kt` — the unconditional `rejectSettlement` and the swallowed `compEx`
- `openbank-settlement-service/src/main/kotlin/com/openbank/settlement/application/workflow/SettlementActivitiesImpl.kt` — `compensate` (writes `REVERSAL_FAILED`), `reverseBookToLedger` (writes `LEDGER_REVERSAL_UNSUPPORTED`), `rejectSettlement`
- `docs/threat-models/openbank-settlement-service.md` — residual risk 6

Refs #6037, #6284

JiRaska:
## Problem

`SettlementWorkflowImpl.settle` called `activities.rejectSettlement(settlementId)` **unconditionally** after the compensation loop, swallowing each `ActivityFailure`:

```kotlin
compensations.forEach { compensate ->
    try { compensate() } catch (compEx: ActivityFailure) {
        log.error("Compensation failed for settlement $settlementId", compEx)   // swallowed
    }
}
activities.rejectSettlement(settlementId)   // <- unconditional
```

`SettlementActivitiesImpl.compensate` writes `REVERSAL_FAILED` when a balance reversal is *attempted and refused* — the payee has already spent the credited funds, balance-service answers 422, and no retry resolves it. The terminal write then overwrote that with `REJECTED` seconds later. The settlements table said the settlement was cleanly rejected while customer money had moved and not come back, and the only surviving evidence was a FAILURE audit event and an ERROR log line — neither of which anything queries or alerts on (Loki retention is 1w; the row is permanent).

It also **defeated the alerting merged in #6284**. That PR publishes an age gauge for every non-terminal status and adds a critical `SettlementReversalFailed` rule — but because the row left `REVERSAL_FAILED` almost immediately, the rule could only fire when the workflow *itself* died between the compensation and the terminal write. In the case it was written for — compensation refused, workflow completes normally — there was nothing left to see.

## Change

- Each compensation is registered together with the status its activity records on failure (`Compensation(onFailure, run)`): `reverseDebit`/`reverseCredit` → `REVERSAL_FAILED`, `reverseBookToLedger` → `LEDGER_REVERSAL_UNSUPPORTED`.
- `unwind()` runs **all** compensations (a refused credit reversal must not strand a debit reversal that would have worked) and returns the obligation left outstanding, or `null` for a complete unwind.
- `rejectSettlement` is called only on a complete unwind. Otherwise the row rests in the status the failing activity already recorded — non-terminal, and therefore visible to the gauge and the alert.
- When both halves fail, a refused **money** reversal outranks an unsupported **ledger** reversal: funds outstanding beats a record outstanding, regardless of which failed first (the ledger compensation runs first, so neither "first wins" nor "last wins" gives the right answer).
- `openapi.yaml` documents that `REJECTED` now means a *complete* unwind. Description-only, so the api-contract gate classifies it `editorial` — no `info.version` bump.
- Threat model residual risk 6 records the new resting behaviour and what it costs.

## Two things surfaced, deliberately not changed

**`reverseBookToLedger` is unreachable.** Its compensation is registered only after `bookToLedger` returns, and `bookToLedger` is the last forward step — nothing can fail afterwards to trigger an unwind. So the activity, its `LEDGER_REVERSAL_UNSUPPORTED` status, and the `SettlementStuckAfterCompensation` coverage of that status are all dead in the current saga shape. `the ledger compensation is unreachable in the current saga shape` asserts exactly that and goes red the day a step is added after the ledger booking. The status pairing is in place so the reporting is already correct at that point rather than silently naming the wrong half.

**The alert will not clear on its own.** Resolving `REVERSAL_FAILED` is a collections/dispute process with no workflow behind it, so a settlement rests there — and the critical alert stays lit — for as long as the obligation does. That is the honest signal, not a bug, but it means the state needs an operator-facing resolution path; recorded in the threat model rather than smoothed over.

## Verification

- `./gradlew :openbank-settlement-service:test :openbank-settlement-service:ktlintCheck :openbank-settlement-service:detekt` — **BUILD SUCCESSFUL**.
- **Negative control**: with the fix reverted and the new tests kept, `8 tests completed, 2 failed` — `a refused money reversal leaves the settlement non-terminal and never rejects` and `a compensation activity that itself fails does not stop the remaining compensations`. The tests fail against the old behaviour, so they can detect its absence.
- The pre-existing test at that spot asserted `REJECTED` and `rejectSettlement == 1` for precisely this scenario and passed against the defect; it now asserts the opposite. Worth noting in review: this is a deliberate inversion of an existing assertion, not an accidental one.

Closes #6286. Follows #6284.

### DIFF
--- a/docs/threat-models/openbank-settlement-service.md
+++ b/docs/threat-models/openbank-settlement-service.md
@@ -185,6 +185,15 @@ replacing the former in-memory stub), so settlement state is durable across rest
    `SettlementReversalFailed` rule — separated because the other rule's premise ("funds are safe,
    only the record is wrong") is false here.
 
+   Nor was it enough to publish the state: `SettlementWorkflowImpl` called `rejectSettlement`
+   unconditionally after the compensation loop, so the row was overwritten to `REJECTED` within
+   seconds — a terminal status asserting a clean unwind while funds were outstanding, and a row
+   the age gauge never saw. `REJECTED` is now written **only when every compensation succeeded**
+   (#6286); a settlement whose compensation was refused rests in `REVERSAL_FAILED`, non-terminal
+   and alerting, until the collections/dispute path resolves it. The residual risk that remains is
+   that this resolution is an operator action with no workflow behind it, so the alert stays lit
+   for as long as the obligation does — deliberately.
+
 7. **A debit applied by balance-service but not observed by settlement-service is not compensated
    (pre-existing, not addressed by #6037).** `SettlementWorkflowImpl` registers the `reverseDebit`
    compensation only *after* `debitPayer` returns, so a debit that balance-service applied while the
--- a/openbank-settlement-service/src/main/kotlin/com/openbank/settlement/application/workflow/SettlementWorkflowImpl.kt
+++ b/openbank-settlement-service/src/main/kotlin/com/openbank/settlement/application/workflow/SettlementWorkflowImpl.kt
@@ -36,32 +36,89 @@ class SettlementWorkflowImpl : SettlementWorkflow {
     private val activities: SettlementActivities =
         Workflow.newActivityStub(SettlementActivities::class.java, activityOptions)
 
+    /**
+     * One registered compensation, paired with the status its activity records when it fails.
+     *
+     * The pairing is what lets [settle] report which half of the unwind is outstanding without
+     * reading the row back: `reverseDebit`/`reverseCredit` write
+     * [SettlementStatus.REVERSAL_FAILED] on a refusal (money moved and not returned), while
+     * `reverseBookToLedger` writes [SettlementStatus.LEDGER_REVERSAL_UNSUPPORTED] (the GL posting
+     * still stands). Both are recorded by the activity itself, inside its own transaction.
+     */
+    private data class Compensation(val onFailure: SettlementStatus, val run: () -> Unit)
+
+    /**
+     * Runs every registered compensation, and returns the status of the obligation left
+     * outstanding — or `null` when the unwind was complete and the settlement may be rejected.
+     *
+     * Every compensation is attempted even after one fails, so a refused credit reversal does not
+     * strand the debit reversal that would still have worked.
+     */
+    private fun unwind(settlementId: UUID, compensations: Collection<Compensation>): SettlementStatus? {
+        val log = Workflow.getLogger(SettlementWorkflowImpl::class.java)
+        var outstanding: SettlementStatus? = null
+        compensations.forEach { compensation ->
+            try {
+                compensation.run()
+            } catch (compEx: ActivityFailure) {
+                // A refused money reversal outranks an unsupported GL reversal: the first means
+                // customer funds are still moved, the second that the ledger owes a correcting
+                // entry. Report the worse of the two, whichever order they failed in.
+                if (outstanding != SettlementStatus.REVERSAL_FAILED) {
+                    outstanding = compensation.onFailure
+                }
+                log.error("Compensation failed for settlement $settlementId", compEx)
+            }
+        }
+        return outstanding
+    }
+
     @Suppress("TooGenericExceptionCaught")
     override fun settle(settlementId: UUID): SettlementStatus {
-        val compensations = ArrayDeque<() -> Unit>()
+        val compensations = ArrayDeque<Compensation>()
 
         return try {
             activities.debitPayer(settlementId)
-            compensations.addFirst { activities.reverseDebit(settlementId) }
+            compensations.addFirst(
+                Compensation(SettlementStatus.REVERSAL_FAILED) { activities.reverseDebit(settlementId) },
+            )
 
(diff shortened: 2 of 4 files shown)


JiRaska:
**Merge record.** Merged 2026-08-22T08:41:54Z as `fc82f8b58`, squash, by the repository owner.

**Approvals: zero** (reviews API reports `length == 0`). `openbank-settlement-service` is in `rules.yaml: money_path_services`, which requires 2 approvals; the requirement was not met. Recorded here because nothing else captures it: the `main-protection` ruleset requires 0 approvals, so GitHub raised no objection, and the author cannot approve their own PR. Same exception as #6284, and no override flag was used.

Green at merge: 28 checks pass, 9 skipped, 0 failures at `0571efd70`.

Note for whoever picks up the follow-up: `openapi.yaml` shipped at **1.3.0**, which over-bumps a docs-only change. That was forced by #6380 — fixed in #6382, merged minutes earlier. With the gate as it now stands the same change would classify `editorial` and pass at 1.2.1. The version is left as merged rather than rewritten.

PR #6372, opened 2026-08-22
fix(admin-ui): clarify service inventory refresh state

JiRaska:
## Summary
- make Tech Inventory refresh an explicit button action
- expose localized accessible name and loading state
- hide decorative refresh icon from assistive technology

## Verification
- git diff --check

## Linked issues

### DIFF
--- a/openbank-admin-ui/src/app/system/inventory/page.tsx
+++ b/openbank-admin-ui/src/app/system/inventory/page.tsx
@@ -178,15 +178,18 @@ export default function TechInventoryPage() {
             </span>
           )}
           <button
+            type="button"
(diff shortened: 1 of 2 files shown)


JiRaska (commented):
Independent review: NO BLOCKER. The diff is limited to the existing Tech Inventory refresh control: it adds type=button, a localized accessible name, aria-busy while the existing disabled/loading behavior remains, and hides the decorative icon. The refresh(true) callback and inventory fetch/data flow are unchanged. git diff --check passes; focused Vitest is unavailable in the clean worktree because dependencies are not installed; repository CI remains authoritative.

PR #6385, opened 2026-08-22
fix(admin-ui): announce clearing loading

JiRaska:
## Summary
- expose the existing Clearing loading state as a polite status
- preserve existing batch fetch/filter and RBAC behavior

## Verification
- git diff --check

## Linked issues


JiRaska (commented):
Independent review: NO BLOCKER. The diff only improves the existing Clearing loading state by exposing it as a polite status and hiding the decorative spinner. Existing batch fetch/filter/RBAC behavior and data rendering are unchanged. git diff --check passes; focused Vitest is unavailable in the clean worktree because dependencies are not installed; repository CI remains authoritative.

PR #6389, opened 2026-08-22
fix(admin-ui): clarify incident refresh state

JiRaska:
## Summary
- add localized accessible name and busy semantics to ICT incident refresh
- preserve existing DORA incident fetch and system:view guard

## Verification
- git diff --check

## Linked issues


JiRaska (commented):
Independent review: NO BLOCKER. The diff is limited to the existing ICT incident refresh control: localized aria-label and aria-busy are added while type/disabled, load callback, DORA fetch, and system:view guard remain unchanged. The icon remains decorative. git diff --check passes; focused Vitest is unavailable because dependencies are not installed; repository CI remains authoritative.

PR #6397, opened 2026-08-22
fix(admin-ui): clarify fee refresh state

JiRaska:
## Summary
- add explicit button semantics and localized busy/name state to fee refresh
- preserve product-catalog fee loading and RBAC behavior

## Verification
- git diff --check

## Linked issues


JiRaska (commented):
Independent review: NO BLOCKER. The diff is limited to the existing Fee Schedule refresh control: type=button, localized aria-label, and aria-busy are added while the load callback, disabled behavior, product-catalog fetch, and RBAC remain unchanged. The refresh icon remains decorative. git diff --check passes; focused Vitest is unavailable because dependencies are not installed; repository CI remains authoritative.

PR #6401, opened 2026-08-22
fix(admin-ui): announce disputes loading

JiRaska:
## Summary
- expose the existing Disputes loading state as a polite status
- preserve existing dispute fetch/filter and RBAC behavior

## Verification
- git diff --check

## Linked issues


JiRaska (commented):
Independent review: NO BLOCKER. The diff only improves the existing Disputes loading state by exposing it as a polite status and hiding the decorative spinner. Existing dispute fetch/filter/RBAC behavior and data rendering are unchanged. git diff --check passes; focused Vitest is unavailable because dependencies are not installed; repository CI remains authoritative.

PR #6413, opened 2026-08-22
ci(governance): guard autonomous-agent PRs against protected paths

JiRaska:
Closes #6412

## What

Adds **`agent-pr-guard`** — an enforced, PR-only gate that fails when an **agent-authored** PR
touches a protected path: CI definitions, money-path / high-blast-radius service or gitops
paths, authorization policy, or governance and gate machinery.

This is the prerequisite for running scheduled unattended agents at all. See #6412 for the
rest of the programme.

## Why a required check, and not the existing hook

The rule exists today only as `protected-pr-check.sh`, a PreToolUse hook on the maintainer's
laptop. That control:

- does not exist in a cloud session or on a GitHub runner — i.e. anywhere an agent runs
  unattended;
- is blind to a merge issued from inside a script even locally (17 PRs merged in one session
  with all of its rules unevaluated);
- is not backstopped by the ruleset: `required_approving_review_count` is 0, and the admin
  bypass actor is the same identity an agent authenticates as.

A required check is the one shape none of that defeats: it runs wherever the PR came from, the
agent that tripped it cannot clear it, and `--auto` cannot merge past it.

## Design decisions worth reviewing

**The head-branch clause is load-bearing.** Every open PR in this repo today is authored by
`JiRaska` — an identity-only scope would cover *nothing* until the bot App exists. So scope is
`declared agent account` **OR** `declared head-branch prefix` (`agent/`).

**An undeclared machine account FAILS.** A Bot-typed author in neither `agent_accounts` nor
`automation_accounts` is exit 2, not a pass. This is what keeps the scope from being a
hand-kept list of the gate's own subjects — the shape that reads as clean when it is short.

**Deterministic automation stays out of scope.** release-please and the deploy snapshotters
legitimately touch money-path paths; their content is generated, not authored, and blocking
them would deadlock the release queue for no safety gained.

**The protected service set is derived**, not retyped: `money_path_services` with the
`openbank-` prefix and `-service` suffix stripped, plus a short declared list of extras that
are high blast radius without being money-path by that list's definition (kyc, party, card
issuance, aml, tax, dispute). Onboarding a money-path service extends this guard with no edit
to the checker.

**Enforced, not advisory.** The deadlock argument that makes `pr-file-overlap` advisory does
not apply here: a human PR is out of scope entirely, so this can never red a human's work, and
an agent PR it reds is one that *should* be waiting for a human.

## Falsifiability

`--self-test` drives 14 classifier cases plus 4 undetermined cases, asserting on the **reason**
[...]

Issue #6412, opened 2026-08-22
Autonomous agent PRs: enforce protected paths before any unattended agent runs


JiRaska:
## Problem

The repo is moving to **scheduled, unattended agents** that open PRs around the clock (a PR
Steward that unblocks the queue, and later an Issue Worker). Before any of that runs, the
control that keeps unattended work off the money path has to exist *where the agent runs*.

Today it does not:

- The only enforcement is `protected-pr-check.sh`, a **PreToolUse hook on the maintainer's
  laptop**. It cannot exist in a cloud session or on a GitHub runner. Even locally it is
  blind to a merge issued from inside a script — 17 PRs merged in one session with every one
  of its rules unevaluated.
- `main`'s ruleset does not supply the control either: `required_approving_review_count` is
  **0**, and the **admin bypass actor is the same identity an agent authenticates as**.
- An agent has already, unprompted, reached for an administrative override on a money-path PR
  when self-approval was refused.

## Work

- [x] **`agent-pr-guard`** — an enforced, PR-only gate that fails when an agent-authored PR
      touches CI definitions, money-path / high-blast-radius service or gitops paths,
      authorization policy, or governance and gate machinery.
- [x] **Production negative control** (done — #6415, `Validate manifests` red on the money-path clause, `21 gates PASS=20 FAIL=1`) — a throwaway PR from an `agent/` branch touching one
      money-path file, to prove the gate's RED is reachable *in CI*, not only in its own
      self-test. Closed immediately after.
- [x] **A dedicated machine identity** (`openbank-agent-bot` GitHub App) holding
      `contents:write` / `pull_requests:write` / `issues:write`, **no admin**, and **not** in
      the ruleset's `bypass_actors`. The guard is the second layer; the token is the first.
- [x] **PR Steward**, read-only half — built (`pr-steward.py`, no write path at all).
      **The scheduled loop is NOT being built**: measured 0 actionable PRs of 51, and the
      queue's entire broken set is 4 PRs that a human owns by rule. See the comment below.
- [ ] **Issue Worker** — deferred for the same reason, and gated on the open-PR count staying
      under 30 in any case (with a deep queue the bottleneck is review, not authoring).

## Notes on the design

Scope is decided by declared agent accounts **or** a declared head-branch prefix. The branch
clause is load-bearing: every open PR in this repo today is authored by the maintainer's own
login, so an identity-only scope would cover exactly nothing until the App exists.

[...]

PR #6415, opened 2026-08-22
test(ledger): negative control — agent-pr-guard must refuse this PR (DO NOT MERGE)

JiRaska:
Refs #6412 · negative control for #6413 · **DO NOT MERGE — closed as soon as the red is observed**

## Purpose

`agent-pr-guard` (#6413) is green in its own self-test and in `run-gates.py`. That is not
proof: a gate that has only ever passed is unfalsified, and the self-test only exercises the
checker's own fixtures — it cannot show that the gate is actually **wired into CI** and
reaches a verdict on a real pull request.

This PR is the known-positive. It adds one file under `openbank-ledger-service/` (a money-path
service) on a branch named `agent/guard-negative-control` (the declared agent prefix). That is
precisely the combination the guard must refuse.

**Expected result: `Validate manifests` FAILS**, with `agent-pr-guard` reporting the
`money-path` clause.

If this PR goes green, the guard is not doing its job and #6413 should not have merged.

## Why it is stacked on #6413's branch

Its base is `feat/agent-pr-guard`, not `main`, so the gate exists in this checkout and CI runs
it. That means the control can be observed before #6413 merges rather than after — which is
the right order for a check whose whole purpose is to gate merges.

## Cleanup

Closed once the failure is recorded here. The branch is a leaf, so deleting it closes nothing
else. The file never reaches `main`.

### DIFF
--- a/openbank-ledger-service/docs/GUARD-NEGATIVE-CONTROL.md
+++ b/openbank-ledger-service/docs/GUARD-NEGATIVE-CONTROL.md
@@ -1,0 +1,11 @@
+# Negative control for `agent-pr-guard` — DO NOT MERGE
+
+This file exists to prove one thing: that `agent-pr-guard`'s RED is reachable in real CI, and
+not only inside its own self-test. A gate that has only ever passed is unfalsified.
+
+It sits under `openbank-ledger-service/` (a money-path service) on a branch named
+`agent/guard-negative-control` (the declared agent prefix), which is exactly the combination
+the guard must refuse. `Validate manifests` is expected to FAIL on this PR.
+
+The PR carrying it is closed as soon as that failure is observed, and this file never reaches
+`main`.


JiRaska:
## Result: the guard's RED is reachable in CI. Closing.

`Validate manifests` **failed**, on the expected clause and on nothing else.

From `gates (lint)`, run [32563622873](https://github.com/JiRaska/open-bank-oss/actions/runs/32563622873):

```
BLOCKED: this PR is agent-authored (head branch `agent/guard-negative-control` uses the agent prefix `agent/`) and:
  * touches money-path / high-blast-radius service or gitops paths, regardless of the PR title's scope.
    matched: openbank-ledger-service/docs/GUARD-NEGATIVE-CONTROL.md
```

Three things this establishes that the self-test could not:

1. **The gate is wired.** It ran on a real pull request, in the real shard, and reached a
   verdict — not against its own fixtures.
2. **The scope clause that matters works.** The PR was classified as agent-authored from the
   **head-branch prefix**, under the maintainer's own login. That is the clause carrying the
   entire guard until the bot App exists, and it is the one a fixture is least able to vouch
   for.
3. **It failed for the right reason and nothing else.** The lint shard reports
   `21 gates PASS=20 FAIL=1` — `agent-pr-guard` is the sole failure. A red for an unrelated
   reason would have proven nothing.

The file never reaches `main`; the branch is deleted with this PR.

PR #6422, opened 2026-08-22
fix(admin-ui): clarify transaction search state

JiRaska:
## Summary
- add explicit button semantics and localized busy/name state to transaction search
- preserve existing transaction query, pagination, BFF and RBAC behavior

## Verification
- git diff --check

## Linked issues


JiRaska (commented):
Independent review: NO BLOCKER. The diff is limited to the existing Transactions search control: type=button, localized aria-label, aria-busy, and decorative icon semantics are added while query construction, pagination, BFF, and RBAC remain unchanged. git diff --check passes; focused Vitest is unavailable because dependencies are not installed; repository CI remains authoritative.

PR #6430, opened 2026-08-22
fix(admin-ui): clarify payment refresh state

JiRaska:
## Summary
- add explicit button semantics and localized busy/name state to payment detail refresh
- preserve list-source refresh, raw payload disclosure, BFF and RBAC behavior

## Verification
- git diff --check

## Linked issues


JiRaska (commented):
Independent review: NO BLOCKER. The diff is limited to the existing Payment Detail refresh control: type=button, localized aria-label, and aria-busy are added while list-source refresh, row handoff/raw payload behavior, BFF and RBAC remain unchanged. The icon remains decorative. git diff --check passes; focused Vitest is unavailable because dependencies are not installed; repository CI remains authoritative.

PR #6439, opened 2026-08-22
fix(admin-ui): clarify card detail loading

JiRaska:
## Summary
- expose Card Detail loading as an announced status
- add accessible name/type to the existing refresh action
- preserve card lifecycle, PCI masking, four-eyes and RBAC flows

## Verification
- git diff --check

## Linked issues

### DIFF
--- a/openbank-admin-ui/src/app/cards/[id]/page.tsx
+++ b/openbank-admin-ui/src/app/cards/[id]/page.tsx
@@ -171,8 +171,8 @@ export default function CardDetailPage() {
         </div>
 
         {loading ? (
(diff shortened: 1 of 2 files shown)


JiRaska (commented):
Independent review: NO BLOCKER. The diff only improves Card Detail loading and refresh semantics: the existing loading state is announced, the spinner is decorative, and refresh has type=button plus a localized accessible name. Card fetch, lifecycle/four-eyes controls, PCI masking, BFF and RBAC remain unchanged. git diff --check passes; focused Vitest is unavailable because dependencies are not installed; repository CI remains authoritative.

PR #6445, opened 2026-08-22
fix(admin-ui): announce account loading

JiRaska:
## Summary
- expose Account Detail loading as a polite status
- preserve account lifecycle, idempotency, BFF and RBAC behavior

## Verification
- git diff --check

## Linked issues


JiRaska (commented):
Independent review: NO BLOCKER. The diff only improves Account Detail loading semantics: the existing loading state is announced and the spinner is decorative. Account fetch, lifecycle actions, idempotency, BFF and RBAC remain unchanged. git diff --check passes; focused Vitest is unavailable because dependencies are not installed; repository CI remains authoritative.

Issue #6242, opened 2026-08-21
check-external-feeds.py scope excludes gitops, so a live NVIDIA egress is declared nowhere


JiRaska:
## What

`check-external-feeds.py` scans only `openbank-*/src/main/resources/application.yaml`. A live
external egress declared anywhere else is **structurally outside its scope** — the gate cannot
report it, and reads clean.

Concretely: `openbank-infra/gitops/apps/holmesgpt.yaml` configures an egress to NVIDIA's
`integrate.api.nvidia.com`. It has never been in the gate's scope. After PR #6241 removes
copilot-service's default to that same host (the last in-scope reference), the host is declared
**nowhere** — while the egress itself is still live.

Found while resolving #6077. The gate did fire correctly on the stale in-scope declaration, and
that was confirmed by negative control (same script on the unmodified tree exits 0), so the gate
works — its *reach* is the defect.

## Why this is the recurring shape

A gate whose scope is narrower than the thing it governs reads as PASSING, never as UNCHECKED. The
declaration file then drifts from reality in the one direction nobody looks: an egress that exists
and is undeclared, rather than a declaration with no egress.

Same family as several findings today — a guard proven on the case it was written for, blind to the
neighbouring one. Compare the Loki ruler gate that asserted the reader and not the writer (#6032),
and the enum-drift gate whose `MIN_OVERLAP` threshold made its count a floor rather than a census
(#5962).

## Suggested fix

1. Widen the scan to every tracked file that can configure an outbound host — at minimum
   `openbank-infra/gitops/**` alongside the service `application.yaml`s.
2. **Derive** the scope from the file set, not a hand-kept path list, and fail on a stale exclusion
   in either direction.
3. Prove it by what it prevents: add `holmesgpt.yaml`'s NVIDIA egress as the negative case, confirm
   the widened gate flags it while undeclared, then declare it and confirm exit 0.

## Not established

Whether other trees carry undeclared egresses — I checked only the host that surfaced here. The
enumeration is the first step of the fix, and it should be done by two independent methods, since
a single grep over a repo with several config idioms is a probe rather than a census.

Refs #6077, #6241

JiRaska:
Resolved on `origin/main`. Verified directly rather than from a PR body:

- `check-external-feeds.py` now carries a `# --- gitops corpus (#6242)` block and its
  `CORPUS_GLOBS` includes `openbank-infra/gitops/**/*.yaml` — three markers present.
[...]

PR #6459, opened 2026-08-22
fix(ci): declare the HolmesGPT NVIDIA endpoint, unblocking main

JiRaska:
## What

`external-feed-declaration-drift` is **failing on `origin/main`**, so every open PR inherits a red
`gates (data)` it did not cause. Confirmed today on #5836 and #5849 — two unrelated diffs failing
the identical gate, which is the signature of a main-side problem rather than a branch defect.

Reproduced on a clean detached worktree of `origin/main`:

```
python3 .github/scripts/check-external-feeds.py --root . --offline
-> exit 1
DRIFT undeclared external URL openbank-infra/gitops/apps/holmesgpt.yaml:47 -> https://integrate.api.nvidia.com/v1
DRIFT undeclared external URL openbank-infra/gitops/apps/holmesgpt.yaml:53 -> https://integrate.api.nvidia.com/v1
```

## Cause

Widening `CORPUS_GLOBS` to `openbank-infra/gitops/**/*.yaml` — the fix for #6242 — brought
HolmesGPT's `OPENAI_API_BASE` and its `modelList` `api_base` into scope for the first time. That
change declared the rest of the newly-visible gitops corpus (the whole `# --- gitops corpus (#6242)`
block) and missed this one.

So the gate is working exactly as intended: a new external dependency in scope with no declaration
is precisely what it exists to catch. The declaration is what was missing, not the gate.

## Why NOT_PROBED and not FEEDS

The endpoint is authenticated with `NVIDIA_API_KEY`, so it cannot be probed anonymously — the same
reason `https://api.deepinfra.com/v1/openai` already sits in `NOT_PROBED`. Declaring it under
`FEEDS` would add a probe that can only ever fail.

## Falsified both ways

On a clean checkout, changing nothing else:

| state | exit | output |
|---|---|---|
| without the entry | **1** | `DRIFT undeclared external URL ... holmesgpt.yaml:47` |
| with the entry | **0** | `drift: OK — every external URL accounted for` |

The gate's stale-entry check runs in the other direction too, so if HolmesGPT ever stops using this
endpoint the entry fails just as loudly as an undeclared URL would.

## Scope

One line. No behaviour change to any workload — this declares a dependency that already exists.

Refs #6242

### DIFF
--- a/.github/scripts/check-external-feeds.py
+++ b/.github/scripts/check-external-feeds.py
@@ -173,6 +173,7 @@ FEEDS = [
 NOT_PROBED = [
     ("https://api.github.com", "authenticated API, not a data feed; failure is loud at call time"),
     ("https://api.deepinfra.com/v1/openai", "authenticated LLM gateway; needs a key"),
+    ("https://integrate.api.nvidia.com/v1", "authenticated LLM endpoint used by the HolmesGPT deployment (NVIDIA_API_KEY); needs a key, so it cannot be probed anonymously"),
     ("https://s3.eu-north-1.amazonaws.com", "AWS endpoint, reached with SigV4 credentials"),
     ("https://kc.open-bank.tech/realms/openbank-customers", "our own Keycloak realm, covered by its own probes"),
     ("https://pid.open-bank.tech", "our own PID issuer, covered by its own probes"),


JiRaska:
Closing as superseded by #6457, which fixes the same thing and **is already merged** (`38c372ec`,
2026-08-22T11:55:37Z — verified, not assumed).

Both PRs declare the same URL for the same reason; they were written independently within about
fifteen minutes of each other, which is the usual shape when a gate turns `main` red and more than
one session sees the same failure.

Measured on live `origin/main`:

```
$ git show origin/main:.github/scripts/check-external-feeds.py | grep -c integrate.api.nvidia.com
1
$ python3 .github/scripts/check-external-feeds.py --offline     # clean worktree at origin/main
drift: OK — 1 declared feed(s), every external URL accounted for.     rc=0
```

This branch's 3-dot delta against `main` is a single added line declaring
`https://integrate.api.nvidia.com/v1` — the URL `main` already declares. Merging it would put a
second entry for the same URL into `NOT_PROBED`, which at best is redundant in a list that is
stale-checked in both directions.

Nothing is lost by closing this: `gates (data)` is green on `main` and the five PRs it was blocking
(#5836, #5748, #5857, #6029, #5849) have been rebuilt on top of the fix.

For what it is worth, the wording here is the better of the two on one point — it names
`NVIDIA_API_KEY` explicitly, which #6457 does not. If that detail is worth keeping, it is a one-line
edit to the merged entry rather than a merge of this branch.

PR #6460, opened 2026-08-22
fix(ci): restore the NOT_PROBED entry for the NVIDIA LLM gateway (#2204)

JiRaska:
## What this is

`CORPUS_GLOBS` grew to cover `openbank-infra/gitops/**/*.yaml` in #6251 (following #6242), which made `integrate.api.nvidia.com/v1` (used twice in `holmesgpt.yaml`) visible to the `external-feed-declaration-drift` gate (issue #2204). The `NOT_PROBED` entry excusing this URL never got carried over during that refactor — it existed before, it's gone now. Every PR merging main since then fails `gates (data)` on this.

## The fix

Restore the `NOT_PROBED` entry:
```python
("https://integrate.api.nvidia.com/v1", "authenticated LLM gateway; needs a key"),
```

## Verification

\`\`\`
$ python3 .github/scripts/check-external-feeds.py --root . --offline
drift: OK — 1 declared feed(s), every external URL accounted for.

$ python3 .github/scripts/check-external-feeds.py --self-test
self-test: 20 passed, 0 failed
\`\`\`

### DIFF
--- a/.github/scripts/check-external-feeds.py
+++ b/.github/scripts/check-external-feeds.py
@@ -173,6 +173,7 @@ FEEDS = [
 NOT_PROBED = [
     ("https://api.github.com", "authenticated API, not a data feed; failure is loud at call time"),
     ("https://api.deepinfra.com/v1/openai", "authenticated LLM gateway; needs a key"),
+    ("https://integrate.api.nvidia.com/v1", "authenticated LLM gateway; needs a key"),
     ("https://s3.eu-north-1.amazonaws.com", "AWS endpoint, reached with SigV4 credentials"),
     ("https://kc.open-bank.tech/realms/openbank-customers", "our own Keycloak realm, covered by its own probes"),
     ("https://pid.open-bank.tech", "our own PID issuer, covered by its own probes"),


JiRaska:
Closing as superseded by #6457, already merged (`38c372ec`). This is the third independently-written
PR for the same one-line declaration — #6457 (merged), #6459 (closed), and this one.

**Nothing was removed, so there is nothing to restore.** Measured on live `origin/main`:

```
$ git show origin/main:.github/scripts/check-external-feeds.py | grep -c integrate.api.nvidia.com
1
$ git log origin/main --oneline -2 -- .github/scripts/check-external-feeds.py
38c372eca fix(ci): declare HolmesGPT's NVIDIA LLM egress, red on main since the corpus widened (#6457)
8a8ac92f3 fix(ci): widen three gates whose scope was narrower than their subject (#6251)
```

The entry is present and `check-external-feeds.py --offline` returns `rc=0` against a clean worktree
at `origin/main`. This branch's 3-dot delta is a single added line for the URL `main` already
declares, so merging it would create a duplicate entry in a list that is stale-checked in both
directions.

My guess at how the title came about: the branch was cut from a `main` that predated #6457, so from
its point of view the entry was missing rather than newly added. Worth knowing because it is the
same trap in a different costume — a claim about `main` taken from a stale local base rather than
from live `origin/main`.

If it helps anyone arriving at this red check later: the root cause was #6251 widening
`CORPUS_GLOBS` to `openbank-infra/gitops/**/*.yaml`, which is exactly what it was meant to do — it
found a real undeclared egress on its first run with that reach. The five PRs it blocked
(#5836, #5748, #5857, #6029, #5849) have been rebuilt on top of the fix and their `gates (data)` is
green again.

PR #6467, opened 2026-08-22
fix(party): create the Hibernate id sequences in lower case so inserts can allocate an id

JiRaska:
## The defect

`PUT /api/v1/parties/{id}/payees` has answered 500 on every call since the endpoint shipped:

```
SQLGrammarException: relation "party_payees_seq" does not exist (42P01)
  [select nextval('party_payees_SEQ')]
```

`V18__party_payees.sql:23` creates the sequence **quoted** — `CREATE SEQUENCE "party_payees_SEQ"` —
so Postgres preserves the case. Hibernate emits `nextval('party_payees_SEQ')`, where the
identifier inside the string literal is **unquoted**; Postgres folds it to lower case and looks
for `party_payees_seq`. Nothing by that name exists.

`V5` made this exact mistake for `parties` / `party_documents` / `party_outbox` and `V6` fixed it
the same way. `V16` and `V18` reintroduced it.

## Second occurrence, not found by the fuzzer

`V16__marketing_consent_tracking.sql:16` has the identical shape:
`CREATE SEQUENCE "party_marketing_consent_SEQ"`. `PartyMarketingConsentEntity` is a
`PanacheEntity`, so the projection consumer's first write fails the same way. Confirmed against a
real Postgres rather than argued from the source:

```
select nextval('party_marketing_consent_SEQ');
ERROR:  relation "party_marketing_consent_seq" does not exist
```

Nothing in the fuzz lane reaches that path — it came out of grepping the migrations for the shape.

## The fix

`V19` drops both quoted sequences and recreates them lower case. `V18` cannot be edited: Flyway
checksums the whole file, and it has been applied.

The new sequences start at `max(id) + 50`, not 1. `id` is `BIGSERIAL`, so any row written before
this migration drew its id from the implicit `<table>_id_seq`; a sequence restarting at 1 would
collide on the first insert. Rollback note is in the migration and restores the broken state
exactly.

## Verification

Against a real Postgres 16, negative control first:

| step | result |
|---|---|
| before V19, `nextval('party_payees_SEQ')` | exit 1, `relation "party_payees_seq" does not exist` |
| apply V19 | exit 0 |
| after V19, same call, table already holding `max(id)=7` | `57` — allocated past the existing rows |
| quoted sequences remaining | none |

`PartySequenceAllocationIT` drives the real endpoint over HTTP against the real schema:

- with `V19` removed — **2 tests, 2 failures**, both with the production error
- with `V19` present — **2 tests, 0 skipped, 0 failures**

`:openbank-party-service:test ktlintCheck detekt` all three ran and passed (checked the task list,
not just the exit code).

The assertion is the allocated id and the row, not merely a 200 — a 200 would also be satisfied by
an endpoint that stopped persisting. The second test calls `nextval` with the identifier spelled
exactly as Hibernate spells it; asserting the lower-case name would pass against the broken
sequence and prove nothing.

## Why a unit test could not have caught this

`PartyServicePayeeTest` mocks the repository, so it passes against a database that cannot allocate
an id at all. The defect lives strictly between Hibernate and Postgres. Same reason the consent
`SuppressionEntity` column-name defect and the sca `eventType` payload defect both shipped green.

## Not included

A CI gate for the shape is the obvious follow-up, and I did not ship one. My first detector —
"every `PanacheEntity` needs a matching `<table>_seq`" — flagged 16 of 24 entities; I checked two
and both were false (balance-service and audit-service do create theirs, in a syntax my regex
missed). A gate whose subject enumeration is wrong is worse than none. Enumerating the subjects
correctly is its own piece of work.

Refs #5913

### DIFF
--- a/openbank-party-service/src/main/resources/db/migration/V19__fix_quoted_sequence_names.sql
+++ b/openbank-party-service/src/main/resources/db/migration/V19__fix_quoted_sequence_names.sql
@@ -1,0 +1,42 @@
+-- Fixes two Hibernate id sequences that Postgres and Hibernate spell differently, so every
+-- insert into `party_payees` and `party_marketing_consent` fails at id allocation.
+--
+-- V5 made this exact mistake for parties/party_documents/party_outbox and V6 fixed it the same
+-- way. V16 and V18 reintroduced it. A quoted `"x_SEQ"` keeps its case in Postgres, while the
+-- `select nextval('x_SEQ')` Hibernate emits carries an UNQUOTED identifier inside the literal,
+-- which Postgres folds to lower case — so it looks up `x_seq` and finds nothing:
+--
+--   ERROR: relation "party_payees_seq" does not exist (42P01) [select nextval('party_payees_SEQ')]
+--
+-- Observed live on four party endpoints in the authenticated fuzz lane (#5913). The
+-- `party_marketing_consent` one has the identical shape and was found by grepping for it, not by
+-- the fuzzer — it fails the same way on the first grant the projection consumer records.
+--
+-- Seeded from the table's own max(id) rather than restarting at 1: `id` is BIGSERIAL, so any row
+-- written before this migration got its id from the implicit `party_payees_id_seq`, and a fresh
+-- sequence starting at 1 would collide with it on the first insert.
+--
+-- Rollback:
+--   DROP SEQUENCE IF EXISTS party_payees_seq;
+--   DROP SEQUENCE IF EXISTS party_marketing_consent_seq;
+--   CREATE SEQUENCE IF NOT EXISTS "party_payees_SEQ" INCREMENT BY 50;
+--   CREATE SEQUENCE IF NOT EXISTS "party_marketing_consent_SEQ" INCREMENT BY 50;
+--   (restores the broken state exactly; both endpoints resume answering 500)
+
+DROP SEQUENCE IF EXISTS "party_payees_SEQ";
+DROP SEQUENCE IF EXISTS "party_marketing_consent_SEQ";
+
+DO $$
+DECLARE
+    next_payee  BIGINT;
+    next_consent BIGINT;
+BEGIN
+    SELECT COALESCE(MAX(id), 0) + 50 INTO next_payee   FROM party_payees;
+    SELECT COALESCE(MAX(id), 0) + 50 INTO next_consent FROM party_marketing_consent;
+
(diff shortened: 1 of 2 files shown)


JiRaska:
**#5951 and #6467 both add a `V19` migration to `openbank-party-service`, and merging both breaks the service.**

```
5951  openbank-party-service/src/main/resources/db/migration/V19__fix_party_payees_sequence.sql
6467  openbank-party-service/src/main/resources/db/migration/V19__fix_quoted_sequence_names.sql
main  highest is V18__party_payees.sql
```

Different filenames, so git merges both without a conflict and nothing anywhere reports a problem —
then Flyway refuses to start with `Found more than one migration with version 19`. The failure mode
worth knowing: Quarkus cannot boot, so every `@QuarkusTest` in the module reports **SKIPPED**, and
the module reads as `N tests / M skipped / 0 failures`. The number people look at is the failure
count, so that scans as a pass.

Both fix the same root cause (a quoted `"x_SEQ"` keeps its case in Postgres while Hibernate's
`nextval('x_SEQ')` folds unquoted to lower case), and both cite #5913. **#6467 strictly dominates:**

| | #5951 | #6467 |
|---|---|---|
| `party_payees` sequence | yes | yes |
| `party_marketing_consent` sequence | **no** | yes |
| starting value | fresh at 1 | `max(id) + 50` |

The starting value is the part that actually matters. #5951 argues a fresh sequence is safe because
"the quoted sequence has never been read, so `party_payees` is empty". #6467 makes the opposite
assumption — `id` is `BIGSERIAL`, so a row inserted by any path that omitted the id got one from the
implicit `party_payees_id_seq`, and a sequence restarting at 1 would collide with it. I have not
established which is true of the live table, and that is the point: `COALESCE(MAX(id), 0) + 50`
is correct under **both** readings and degrades to 50 on an empty table, whereas restarting at 1 is
correct only under one. #6467 also carries the `GRANT` and covers the second sequence, which #5951
misses entirely — that one fails identically on the first marketing-consent grant the projection
consumer records.

**Suggested resolution:** close #5951 in favour of #6467, rather than renumbering one to V20. Two
migrations fixing the same sequence in sequence would leave V19 dropping and recreating
`party_payees_seq` only for V20 to do it again with a different start value — the end state depends
on apply order, which is exactly what you do not want in a money-path schema.

Renumbering is only safe while a migration has never been applied anywhere; after that, Flyway
checksums the file and any edit is a startup failure. Neither V19 appears to have been applied yet,
so the option is open — I just do not think it is the right one here.

I have not closed or changed either PR. Both are money-path and this is a correctness call, not a
duplicate cleanup.

PR #6481, opened 2026-08-22
fix(settlement): make the ledger compensation reachable and truthful (#6410)

### LINKED ISSUE #6410: settlement: reverseBookToLedger is unreachable — the ledger compensation, its status, and its alert coverage are all dead in the current saga shape
## What

`SettlementActivitiesImpl.reverseBookToLedger` can never run.

`SettlementWorkflowImpl.settle` registers each compensation only after its forward activity returns, and `bookToLedger` is the **last** forward step:

```kotlin
activities.bookToLedger(settlementId)
compensations.addFirst(
    Compensation(SettlementStatus.LEDGER_REVERSAL_UNSUPPORTED) { activities.reverseBookToLedger(settlementId) },
)

SettlementStatus.BOOKED     // <- nothing can fail after this point
```

Nothing follows the registration, so no failure can trigger an unwind that includes it. If `bookToLedger` itself throws, the compensation was never registered.

Asserted on `main` since #6369 — `SettlementWorkflowImplTest > the ledger compensation is unreachable in the current saga shape` states it directly and goes red the day a step is added after the ledger booking.

## What is dead as a consequence

- The activity `reverseBookToLedger`, including its non-retryable `LedgerReversalUnsupported` failure and the four-point KDoc explaining why the GL cannot be reversed automatically.
- The status `SettlementStatus.LEDGER_REVERSAL_UNSUPPORTED` — never written by any reachable path.
- The `SettlementStuckAfterCompensation` alert's coverage of that status (added in #6284) and its share of `SettlementStrandedGauge`'s published series.
- `SettlementCompensationCapabilities`' boot-time announcement that a failure "will be marked LEDGER_REVERSAL_UNSUPPORTED for manual correction".

None of that is wrong in itself — it is all correct code for a saga shape this service does not have. The cost is that four separate places describe a recovery behaviour that cannot occur, which is the kind of thing a reader trusts.

## Why it is not obviously "just delete it"

The GL exposure it was written for is real: `bookToLedger` posts to the ledger, and if a **later** step ever fails, that posting must be dealt with. The saga has no later step *today*. So there are two honest resolutions and they point in opposite directions:

1. **Remove the dead path** — delete the activity, the status, its openapi enum value (a spec change), the alert regex entry, and the boot announcement. Smallest, truthful, and loses the groundwork if a step is added later.
2. **Give the saga the step it implies.** If settlement is ever expected to do anything after booking to the ledger (publish an outcome event, mark the settlement complete in a downstream system, confirm with the counterparty), the compensation becomes live and the current code is already correct. The pairing added in #6369 (`Compensation(onFailure, run)`) exists so the reporting is right at that moment rather than naming the wrong half.

Worth deciding deliberately rather than by drift. If (1), note that `LEDGER_REVERSED` is *already* deprecated-and-never-written in the same enum, so this would be the second unreachable value there — the enum is accumulating states that describe history rather than behaviour.

## Where

- `openbank-settlement-service/src/main/kotlin/com/openbank/settlement/application/workflow/SettlementWorkflowImpl.kt` — the registration order
- `.../workflow/SettlementActivitiesImpl.kt` — `reverseBookToLedger`
- `.../workflow/SettlementCompensationCapabilities.kt` — the boot announcement
- `.../domain/model/Settlement.kt` — `LEDGER_REVERSAL_UNSUPPORTED`, and the already-dead `LEDGER_REVERSED`
- `openbank-infra/gitops/components/payments/prometheus-rules.yaml` — `SettlementStuckAfterCompensation`
- `openbank-settlement-service/src/main/resources/openapi.yaml` — the status enum

Refs #6286, #6284, #6037

JiRaska:
## What

`SettlementActivitiesImpl.reverseBookToLedger` could never run, and fixing that turned out to
close a live money-path hole rather than delete dead code.

Closes #6410. Refs #6037, #6286, #6284, #750.

## The mechanism, measured

The issue reports the compensation as unreachable and offers two resolutions — delete the dead
path, or give the saga a step after `bookToLedger` so the compensation becomes live. Reading the
activity turns up a third answer, and it is the one the code actually supports.

`bookToLedger` is **not atomic**:

```kotlin
ledgerPort.book(settlementId)                                          // the GL posting lands
settlementRepository.updateStatus(settlementId, SettlementStatus.BOOKED) // ...then the row
```

Either half can fail on its own. So the exposure the compensation was written for is reachable
**inside the last step**, and does not need a new saga step at all: if the posting lands and the
status write then fails on all five Temporal attempts, `bookToLedger` throws with a journal
already in the general ledger. Under the registered-after shape the compensation for it had not
been registered, so the saga reversed both balance movements, wrote `REJECTED`, and left a GL
entry standing that nothing in the row, the alerts or the audit trail mentioned.

That is the same defect class #6286 closed one step over: a settlement recorded as cleanly
rejected while an obligation is outstanding.

So: **register the ledger compensation before the activity that creates the exposure**, which
also makes it reachable. Registering first is only safe because the compensation no longer
assumes the activity got anywhere.

## What a correct ledger compensation has to do

Nothing in the fleet reverses a ledger posting from a saga, so there was no pattern to mirror —
that absence is itself part of the answer. What it *can* do is establish the facts before
reporting them.

The original KDoc listed three blockers. One of them does not hold:

| Blocker | Status |
|---|---|
| `ledger.reverse` is approval-gated | **Binding.** `reverse` is a `rules.yaml: four_eyes.verbs` verb, computed from the action name alone with no awareness of the caller. `rules.yaml`' own guardrail on that list says adding an automated caller to a four-eyes verb pauses the automation indistinguishably from the human path it gates. |
| No journal id is retained | **Falsified.** `SettlementJournalFactory` posts with `transactionId = settlementId`, and ledger serves `GET /api/v1/journals/transaction/{transactionId}`. The settlement id *is* the handle. |
| An ATTESTED period 409s | **Binding.** Correcting forward in the open period is a *different posting* from a reversal. |

The falsified one is exactly what the compensation needed. It now reads the ledger back and
reports one of **three** outcomes, each with its own status value — never a flag shared with a
real reversal (the `PushResult.skipped()` rule):

- **`LEDGER_REVERSAL_UNSUPPORTED`** — a journal is confirmed to exist. The GL owes a manual
  correcting entry. Non-retryable: neither binding blocker is resolved by a retry, and burning
  five attempts would only delay the two compensations that return money.
- **`LEDGER_NOT_POSTED`** — the ledger holds nothing, so the booking never landed. **Returns
  normally**, so the workflow rejects the settlement cleanly. This is the case the previous
  unconditional behaviour got wrong on *every* ordinary ledger failure, and summoning an
  accountant to correct an entry that does not exist is noise on the one control that makes a
  real GL discrepancy visible.
- **`LEDGER_STATE_UNKNOWN`** — the lookup itself failed. "Nobody checked" is a third fact, not a
  rounding of the other two. Retryable, because an unreachable ledger is the one failure here a
  retry can genuinely resolve.

Because the reported status now depends on *how* the activity failed, `Compensation.onFailure`
became a function of the failure instead of a constant — otherwise the workflow would return
`LEDGER_REVERSAL_UNSUPPORTED` while the row said `LEDGER_STATE_UNKNOWN`.

## Blast radius

Latent, never exercised. Read-only against the live primaries: the `settlements` table holds
**zero** rows, and the ledger holds **zero** journals whose description matches a settlement
booking. No settlement has ever reached any compensation state, so nothing needs correcting —
this is a fix ahead of the traffic, not an incident.

## Proof

- **Negative control.** Reverting *only* the workflow registration order (production change,
  tests untouched) turns 5 of 11 workflow tests red, including
  `expected: LEDGER_REVERSAL_UNSUPPORTED but was: REJECTED` — which is the defect stated as an
  assertion.
- **Real path, not mocks.** `SettlementReversalIT` drives the real CDI activity bean through its
  real Vert.x-context bridge against a real Postgres (Testcontainers, real Flyway) and a real
  HTTP ledger stub, and asserts the resulting row with **plain JDBC**. All three outcomes are
  covered. A mocked port cannot tell "asked the ledger" from "assumed".
- **The contract is replayed, not assumed.** The whole design rests on ledger answering `200 []`
  — not `404` — for an unknown transaction. That is a fact about someone else's service, so it is
  pinned as a consumer pact interaction with a **literal** path and verified by the real provider:
  `LedgerPactProviderVerificationTest` replays it green (10 interactions, 0 failures). A consumer
  pact alone cannot catch a wrong path; only the provider replay can.

## Alert and observability coverage

`SettlementStuckAfterCompensation` gains the two new statuses. `SettlementStrandedGauge` needed
**no edit at all** — it derives its published set from the enum, which is the derivation paying
for itself.

## Verified

`:openbank-settlement-service:test` 66 tests / 0 failures / **0 skipped** (read from the JUnit
XML, not the console), plus `ktlintCheck`, `detekt`, `quarkusBuild` and `koverVerify`.
`check-threat-model-diff.py --base origin/main --enforce` passes and the threat model's risk 5 is
rewritten.

## What still needs a decision (not taken here)

Making settlement **actually reverse** a GL posting needs one of:

1. an M2M grant for `ledger.reverse` — a `rules.yaml` governance decision, and the guardrail on
   `four_eyes.verbs` argues against it: a failed saga posting its own GL reversal is close to
   what dual control exists to prevent; or
2. a distinct operator-only correcting action that is not the four-eyes-gated `ledger.reverse`.

Either way the ATTESTED-period case needs a defined behaviour, and "correct forward in the open
period" is a different posting that this service cannot choose on a settlement's behalf. This PR
deliberately ships only the unambiguous half — establishing and truthfully reporting the GL
state — and leaves the reversal itself to that decision.

**One inherited exposure, stated plainly:** ledger's `GET /journals/transaction/{id}` is
`@Authorize(action = "ledger.read")`, which the shared M2M identity reaches only via
`matrix-allows` (`ROLE_OPERATOR`), while the deployed realm grants
`service-account-openbank-services` only `ROLE_API`. That is the same "advisory: would DENY,
masked by `AUTHZ_ENFORCE=false`" state `ledger.create` is already in (#750), one read-only action
wider. Its failure mode is the safe one: a denied lookup yields `LEDGER_STATE_UNKNOWN`, which is
retryable and non-terminal, not a false clean-ledger claim.

## Parallel work

#5723 (open, active) also edits `SettlementActivitiesImpl.kt` and
`prometheus-rules.yaml`. Content compared, not titles — the changes are genuinely divergent, not
duplicates. Whichever lands second resolves; both edits here are additive.

## Checklist

- [x] `openapi.yaml` `info.version` 1.3.0 → **1.4.0** (additive enum values), re-read off live `origin/main`
- [x] No DB migration (status is a varchar column, no enum type)
- [x] Threat model updated; threat-model-diff gate green
- [x] Contract test added **and** provider replay verified
- [ ] Money-path: 2 approvals required — author cannot self-approve

### DIFF
--- a/docs/threat-models/openbank-settlement-service.md
+++ b/docs/threat-models/openbank-settlement-service.md
@@ -80,7 +80,7 @@ replacing the former in-memory stub), so settlement state is durable across rest
 | ID | Threat | Mitigation |
 |----|--------|------------|
 | R1 | Settlement denied by payer ("I never authorised this") | Temporal workflow execution ID links to the initiating payment ID in `sepa-payment-service` / `domestic-payment-service`; full chain queryable from admin UI |
-| R2 | Compensation claimed to have run when it did not | **This threat was realised, not mitigated, from ADR-0101 P3 until #6037.** The cited mitigation — "compensation activities update status to `REVERSED`/`CREDITED_REVERSED`/`LEDGER_REVERSED` atomically" — *was* the defect: updating the status column was the **only** thing `reverseDebit`/`reverseCredit`/`reverseBookToLedger` did. They logged `"stub: wire reversal to balance-service"` and returned success, so the status row asserted an unwind that had moved no money. Now: the two balance reversals issue real counter-movements to balance-service before writing a status, a refused reversal is recorded as `REVERSAL_FAILED` (not as success), and the unimplemented ledger reversal throws a non-retryable `ApplicationFailure` and records `LEDGER_REVERSAL_UNSUPPORTED`. Verified by `SettlementReversalIT`, which asserts the outbound HTTP request over real HTTP and the resulting row over plain JDBC — a mocked port cannot tell a counterparty call from a no-op. |
+| R2 | Compensation claimed to have run when it did not | **This threat was realised, not mitigated, from ADR-0101 P3 until #6037.** The cited mitigation — "compensation activities update status to `REVERSED`/`CREDITED_REVERSED`/`LEDGER_REVERSED` atomically" — *was* the defect: updating the status column was the **only** thing `reverseDebit`/`reverseCredit`/`reverseBookToLedger` did. They logged `"stub: wire reversal to balance-service"` and returned success, so the status row asserted an unwind that had moved no money. Now: the two balance reversals issue real counter-movements to balance-service before writing a status, a refused reversal is recorded as `REVERSAL_FAILED` (not as success), and the ledger reversal — which #6410 made reachable at all, and which now reads the ledger back before reporting — records `LEDGER_REVERSAL_UNSUPPORTED` with a non-retryable `ApplicationFailure` only for a journal confirmed to exist, `LEDGER_NOT_POSTED` when the ledger holds none, and `LEDGER_STATE_UNKNOWN` when it could not be asked. Verified by `SettlementReversalIT`, which asserts the outbound HTTP request over real HTTP and the resulting row over plain JDBC — a mocked port cannot tell a counterparty call from a no-op. |
 
 ### I — Information disclosure
 
@@ -188,21 +188,47 @@ replacing the former in-memory stub), so settlement state is durable across rest
    frontend is unreachable — a sandbox canary must confirm the worker registers and a real settlement
    drives `PENDING → BOOKED` before merge, and the deterministic simulation (ADR-0100/0115) stays green.
 
-5. **A settlement that fails after `bookToLedger` leaves the GL entry standing (issue #6037).**
-   `reverseBookToLedger` is deliberately NOT implemented and fails loudly
-   (`LEDGER_REVERSAL_UNSUPPORTED` + a non-retryable `ApplicationFailure`) rather than reporting a
-   reversal it did not perform. ledger-service does expose `POST /api/v1/journals/{journalId}/reverse`,
-   but three things must be decided before settlement-service may call it: (a) `ledger.reverse` is
-   maker-checker gated, so a service-account call queues for approval instead of posting, and granting
-   a machine that action is a `rules.yaml: shared_m2m_matrix_write_grants` decision — an automatic GL
-   reversal driven by a failed saga is close to what maker-checker exists to prevent; (b) settlement
-   does not retain the journal id (`bookToLedger` discards the response), so it must be persisted or
-   re-resolved via `GET /api/v1/journals/transaction/{transactionId}`; (c) the endpoint answers 409
-   when the entry's fiscal period is ATTESTED, and the accounting convention there is to correct
-   forward in the open period, which is a different posting. **Operational consequence:** such a
-   settlement unwinds both balance movements and needs a manual correcting entry in the general
-   ledger. It is visible as a `LEDGER_REVERSAL_UNSUPPORTED` row and an ERROR log line, and is
-   announced at boot by `SettlementCompensationCapabilities`.
+5. **A settlement that fails after `bookToLedger` leaves the GL entry standing (issues #6037, #6410).**
+   `reverseBookToLedger` still does not reverse anything, and now **establishes what the general
+   ledger actually holds** before saying so. Until #6410 it could not run at all: its compensation
+   was registered only after `bookToLedger` RETURNED, and `bookToLedger` is the last forward step,
+   so nothing could fail afterwards to trigger it. What made it dead also made the exposure real —
+   `bookToLedger` posts the journal and *then* writes `BOOKED`, so a failure of the second half
+   left a GL entry standing while the settlement unwound both balance movements and wrote
+   `REJECTED`, with nothing in the row, the alerts or the audit trail recording it. The
+   compensation is now registered **before** the activity runs, and reports one of three
+   outcomes from a read of `GET /api/v1/journals/transaction/{settlementId}`:
+   `LEDGER_REVERSAL_UNSUPPORTED` (a journal confirmed to exist — the GL owes a correcting entry,
+   non-retryable), `LEDGER_NOT_POSTED` (the ledger holds nothing, so no obligation and no
+   failure), `LEDGER_STATE_UNKNOWN` (the lookup itself failed — retryable, and explicitly not
+   rounded to either neighbour).
+
+   Two of the three blockers to an automatic reversal stand and remain **decisions this service
+   cannot take**: (a) `reverse` is a `rules.yaml: four_eyes.verbs` verb, so `ledger.reverse`
+   carries `four_eyes_required`, computed from the action name alone with no awareness of the
+   caller — the guardrail on that list states that adding an automated caller to a four-eyes verb
+   pauses the automation indistinguishably from the human path it gates, and a failed saga posting
+   its own GL reversal is close to what dual control exists to prevent; (b) the endpoint answers
+   409 when the entry's fiscal year is ATTESTED, and correcting forward in the open period is a
+   *different posting* from a reversal. The third blocker cited in #6037 — "no journal id is
+   retained" — **does not hold**: `SettlementJournalFactory` posts with
+   `transactionId = settlementId`, so the settlement id is the handle, and that is what the new
+   read uses.
+
+   **No new trust boundary.** The lookup is read-only, goes to a counterparty settlement already
+   calls (`ledger-port`/`LedgerRestClient`, same `OidcClientRequestReactiveFilter` client-credentials
+   identity), and adds no new egress destination. It does inherit the existing authz exposure
+   rather than introduce one: ledger's `GET /journals/transaction/{id}` is `@Authorize(action =
+   "ledger.read")`, which the shared M2M identity reaches only through `matrix-allows`
+   (`ROLE_OPERATOR`) — and the deployed gitops realm grants `service-account-openbank-services`
+   only `ROLE_API`. That is the same "advisory: would DENY, masked by `AUTHZ_ENFORCE=false`"
+   state `ledger.create` is already in (issue #750), one action wider, and its failure mode is
+   the safe one: a denied lookup yields `LEDGER_STATE_UNKNOWN`, which is retryable and
+   non-terminal, not a false clean-ledger claim. **Operational consequence:** a settlement whose
+   booking landed and then failed unwinds both balance movements and needs a manual correcting
+   entry in the general ledger; it is visible as a `LEDGER_REVERSAL_UNSUPPORTED` row, an ERROR log
+   line, an age series behind `SettlementStuckAfterCompensation`, and is announced at boot by
+   `SettlementCompensationCapabilities`.
 
 6. **A reversal can be legitimately refused and the money stays moved (issue #6037).**
    balance-service enforces `booked - amount >= overdraftFloor` on every debit, so if the payee has
--- a/openbank-infra/gitops/components/payments/prometheus-rules.yaml
+++ b/openbank-infra/gitops/components/payments/prometheus-rules.yaml
@@ -403,6 +403,17 @@ spec:
           # so these states are transient by construction and a row parked in one has a workflow
           # that died between the compensation and the terminal write.
           #
+          # LEDGER_NOT_POSTED and LEDGER_STATE_UNKNOWN join them for #6410, and only became
+          # reachable with it: the ledger compensation is now registered BEFORE bookToLedger runs,
+          # so it can actually execute, and it reports which of the three situations the general
+          # ledger is in rather than assuming the worst. Neither has customer money in limbo, so
+          # both belong at this severity rather than with SettlementReversalFailed below —
+          # LEDGER_NOT_POSTED because the ledger was asked and holds nothing, LEDGER_STATE_UNKNOWN
+          # because the balance reversals are unaffected by a ledger that could not be reached.
+          # LEDGER_STATE_UNKNOWN parked for 8h is the one to act on first: it means nobody has
+          # established whether the GL carries this settlement, so it could still turn out to be
+          # LEDGER_REVERSAL_UNSUPPORTED.
+          #
           # REVERSAL_FAILED is deliberately NOT in this rule — there the money did not come back,
           # so it is critical, not warning. See SettlementReversalFailed below.
           #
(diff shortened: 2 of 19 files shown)


JiRaska:
**Heads-up: `main` moved under this PR, and it moved the other way on the same issue.**

#6484 merged at 14:26 UTC (`2da87cf6f`) and closed #6410 by taking the **opposite** resolution: it *removed* `reverseBookToLedger`, the `LEDGER_REVERSAL_UNSUPPORTED` status, that status's entry in the `SettlementStuckAfterCompensation` regex, the openapi enum value, the boot announcement and the rego allow-list entry — on the grounds that the compensation was unreachable (its registration follows `bookToLedger`, the last forward step) and that making it *real* was product scope rather than cleanup.

This PR argues the better case, and the repository owner has chosen **this** direction: reframe it on top of current `main` rather than close it or revert #6484.

I have deliberately **not** touched this branch. Its worktree is live — files modified within the last two hours and a `java` process holding the directory open — and rebasing or force-pushing someone else's active branch is how work gets lost. The rebase is yours to do when convenient.

## What changed on `main`, concretely

12 of this PR's 18 files were touched by #6484, so expect conflicts in all of them:

- `SettlementWorkflowImpl` — `reverseBookToLedger` registration removed; the `Compensation(onFailure, run)` pairing from #6369 is gone too (with one remaining status, every compensation records `REVERSAL_FAILED`). A comment at that spot says when to bring the pairing back — that moment is this PR.
- `SettlementActivitiesImpl` — the activity is deleted. Its three reasons a GL posting cannot be reversed (maker-checker on `ledger.reverse`, no retained journal id, fiscal-period locks) were **kept and moved** into the KDoc of `bookToLedger`, phrased as what a step added after it would owe. This PR answers exactly those three, so that KDoc is the place to update.
- `Settlement.kt` — `LEDGER_REVERSAL_UNSUPPORTED` removed; `LEDGER_REVERSED` still there, deprecated.
- `openapi.yaml` — enum value removed, `info.version` now **1.4.0** on `main`; re-read it before bumping (a competing bump is the #3055 trap).
- `prometheus-rules.yaml` — the status is out of the `SettlementStuckAfterCompensation` regex.
- `SettlementWorkflowImplTest` — a new test pins the shape:

  ```kotlin
  assertThat(calls.order)
      .describedAs("nothing may run after bookToLedger, or the GL posting needs a compensation")
      .endsWith("bookToLedger")
  ```

  **That test is designed to go red on this PR.** Its KDoc says so: red here is the signal that the saga grew a step and the GL question now has to be answered — which is what this PR does. Replace it with the reachable-path assertions rather than deleting it silently.

- `docs/threat-models/openbank-settlement-service.md` — residual risk 5 was rewritten as "nothing can fail after `bookToLedger`, so the GL posting has no compensation", framed as a constraint on future work. This PR turns that back into a live risk with a mitigation; it needs rewriting again, not just conflict-resolving.

## Why this happened

I filed #6410 and then implemented it myself without checking whether anyone was already on it. #6481 predates #6484 by an hour. The duplicate-detection I ran was over issues, not over open PRs and live worktrees — that is the gap, and it is mine.

Refs #6410, #6484

JiRaska:
**`openapi.yaml` `info.version` "1.4.0" is already taken on `main`.** This PR bumps
"1.3.0" → "1.4.0", but #6484 landed that same number while this branch was open, so

```
$ git show origin/main:openbank-settlement-service/src/main/resources/openapi.yaml | grep -m1 '^  version:'
  version: "1.4.0"
```

The next free number is **"1.5.0"**.

last run predates #6484. That is the one shape no base resolution can close, which is

Worth re-checking rather than just applying "1.5.0" — read the version off live `main` at

```bash
git fetch origin main && git show origin/main:openbank-settlement-service/src/main/resources/openapi.yaml | grep -m1 '^  version:'
```

consent-service `1.7.0`.

JiRaska:
**Correction to my earlier comment on this PR — it undersold what you have.**

I described this as needing to be "reframed on top of current `main`", as though #6484 stood and this were a follow-up. Having read the diff, that is backwards: **#6484 is wrong and this PR is the fix.**

The registration-order comment in your `settle()` is the part I missed entirely. `bookToLedger` posts the journal and *then* writes `BOOKED` — a failure of the status write throws after the general ledger has been written, and with the compensation registered only after the activity returns, nothing exists to record it. I read "unreachable" off the registration order and concluded it was a property of the saga's shape, then deleted the recovery path for a failure that can actually happen.

The three-way outcome in `reverseBookToLedger` — `LEDGER_STATE_UNKNOWN` / `LEDGER_NOT_POSTED` / `LEDGER_REVERSAL_UNSUPPORTED` — is what makes registering ahead of the activity safe, and it applies the repository's own `PushResult.skipped()` rule a level deeper than I did.

So: this PR is not superseded and should not be closed. It needs to **partially revert** #6484 (restore the activity, the status, the openapi enum value, the `SettlementStuckAfterCompensation` regex entry) and to **delete** the test I added, `booking to the ledger is the last forward step, so it registers no compensation` — it asserts an invariant that is not true, and it is the one thing on `main` that will actively fight this change.

The record is corrected on #6484 and #6410. The rebase is still yours — your worktree has a live `java` process, so I have not touched the branch.

JiRaska:
**The red on this PR is inherited, not yours — don't debug it.**

`gates (lint)` and `Validate manifests` (the latter is only the shard aggregator, so it fails with the former) are failing because this branch's merge-base is `9708073f3`, which predates #6493.

`main` was red for every PR this afternoon: #6434 added `.github/scripts/gen-case-collaboration-opa-data.py` without an entry in `regen-derived.sh`'s `GENERATORS` inventory, so the gate reported

```
```

#6493 fixed that at 14:01 UTC. Your base has `0` occurrences of `case-collaboration-opa-data` in that file; `main` has `1`. So the rebase clears this failure by itself — no change needed on your side.

While you are rebasing: #6516 reverts #6484 and restores this branch's original base, so the conflict surface should be close to nothing rather than the 12 files I warned about earlier. It was green and awaiting merge when I last looked; worth checking it has landed before you start, or you will rebase onto the deletion instead of onto the revert.

Refs #6493, #6516

JiRaska:
**Correcting my earlier comment: the target version is `1.6.0`, not `1.5.0`.** And a bump on its own
is no longer actionable — this PR is `DIRTY`, with three content conflicts.

`1.5.0` was taken by #6516, the revert, which deliberately set it rather than reverting the version
backwards:

```
$ git show origin/main:openbank-settlement-service/src/main/resources/openapi.yaml | grep -m1 version:
  version: "1.5.0"
$ git log origin/main --oneline -1 -- openbank-settlement-service/src/main/resources/openapi.yaml
0d70025b3 Revert "fix(settlement): remove the ledger compensation that could never run (#6484)" (#6516)
```

That is the same trap I flagged here two hours ago, one iteration later: my own advice went stale
because the number moved while the PR sat. Read it off live `main` at the moment you rebase, not
from this thread.

**The conflicts are not mechanical, which is why I have not touched the branch.** `main` is 52
commits ahead and three of them landed on this PR's own files:

```
CONFLICT  .../application/workflow/SettlementActivitiesImpl.kt
CONFLICT  .../resources/openapi.yaml
CONFLICT  .../workflow/SettlementActivitiesImplTest.kt
```

The interesting one is `SettlementActivitiesImpl.kt`. #5723 (merged ~1h ago) wrapped every activity
in `step(SettlementStep.X) { … }` for metrics; this PR rewrites `reverseBookToLedger` to ask the
ledger what it holds and record one of three outcomes. Both edits are wanted and they overlap
directly — the rewritten compensation needs the metrics wrapper around it, which is a combine, not a
pick-one. Same for the test file.

Worth checking while you are in there: #5723's `@Suppress("TooManyFunctions")` comment counts "the
seven activities `SettlementActivities` declares … plus `compensate`, `audit`, `step` and
`runOnVertxContext`" — exactly 11, and detekt fires **at** the threshold. Confirm the arithmetic
still holds after this PR's shape lands.

Also: `SettlementStep.REVERSE_LEDGER_BOOK` now has a real producer again after the revert, so the
metrics path for it is live rather than a label that can never appear.

Your worktree at `/private/tmp/ob-6410-84063` is clean, at the PR head, with nothing unpushed
(idle since 15:19Z), so a rebase there loses nothing.

JiRaska:
Rebased onto `main` and pushed as `4669b6feb`. No longer `DIRTY`.

**Read this before approving — it is not a plain rebase.** Four things had to be reconciled, and one
of them changes a test you did not write.

**1. The metrics wrapper and this PR's rewrite both survive.** #5723 wrapped every activity in
`step(SettlementStep.X)`; this branch rewrote `reverseBookToLedger`. They collide directly. The
resolution rests on `step()` being `runOnVertxContext` plus the meter:

```kotlin
private fun step(step: SettlementStep, block: suspend () -> Unit): Unit = runOnVertxContext {
    try { block(); metrics.sagaStep(step, COMPLETED) }
    catch (ex: Throwable) { metrics.sagaStep(step, FAILED); throw ex }
}
```

so swapping the opener gives both, and `return@runOnVertxContext` became `return@step`. The three
outcomes map onto it correctly: `LEDGER_NOT_POSTED` returns normally and meters as **completed**
(nothing was owed, nothing was done — that is a finished step, not a failure), and both
`ApplicationFailure`s propagate through the catch and meter as **failed**.

**2. `SettlementActivitiesMetricsTest` needed the new port at two construction sites.** That file
came from #5723 and is not in this branch's diff, so the rebase could not surface it — only
`compileTestKotlin` did.

**3. I changed one of #5723's assertions, and this is the part to review.** It asserted
`reverseBookToLedger` always throws, with the comment *"Ledger reversal is NOT implemented and fails
loudly on purpose (#6037)"*. That is precisely the invariant this PR replaces with three outcomes.
The relaxed mock reports zero journals, so the test now took the `LEDGER_NOT_POSTED` path where a
clean return is correct, and it failed. I made it state which case it exercises rather than weaken
it:

```kotlin
coEvery { ledgerJournalLookupPort.countJournalsForSettlement(id) } returns 1
```

A journal exists, so the reversal is still unsupported, still throws, and still meters as FAILED —
the test proves what it was written to prove. But it is your call whether the *other* two outcomes
deserve their own metrics cases; I did not add them.

**4. detekt `LongParameterList` fires AT its threshold of 9** and `LedgerJournalLookupPort` is the
ninth. The fleet's usual answer is field injection, which does not work here: the unit tests
construct this class directly and `MetricsTestableActivities` subclasses it through the same
constructor, so moving a port to `@Inject lateinit var` would break both doubles for a lint
threshold. Suppressed with the reasoning in place; worth restructuring if a tenth port appears.

**`openapi.yaml` is `1.6.0`**, not the `1.5.0` I suggested yesterday — the #6516 revert took `1.5.0`.
Re-read it off live `main` before merging; that number has moved twice in a day.

Verified locally against the rebased tree: `:openbank-settlement-service:test` (all pass),
`ktlintCheck` and `detekt` clean. `ktlintFormat` reordered one import I had inserted.

PR #6482, opened 2026-08-22
fix(admin-ui): clarify regulatory preview loading

JiRaska:
## Summary
- add explicit button semantics and localized busy/name state to regulatory preview data loading
- preserve report preview, export-readiness and read-only data flows

## Verification
- git diff --check

## Linked issues


JiRaska (commented):
Independent review: NO BLOCKER. The diff is limited to the existing Regulatory preview Load data control: type=button, localized aria-label, aria-busy, and decorative icon semantics are added while preview fetch/readiness logic and RBAC remain unchanged. git diff --check passes; focused Vitest is unavailable because dependencies are not installed; repository CI remains authoritative.

Issue #6426, opened 2026-08-22
Govern case.join/case.contribute: rego policy + tests landed — the rca-investigator grant, the OPA-backed coordinator PEP and the shadow pilot have not


JiRaska:
## Problem

ADR-0244 defines mid-run case participation, and the coordinator already exposes Temporal signals, but runtime authorization is still an in-process config list. No non-coordinator charter holds `case.join`/`case.contribute`, the OPA bundle does not project or decide case capabilities, and no bounded shadow pilot proves authorization → invocation → signal consumption → contribution persistence.

## Scope

Deliver this separately from Agent Control Room PR #6424:

- new ADR for the collaboration grant and evidence chain
- grant only the incident-response `rca-investigator` pilot `case.join` and `case.contribute`
- extend generated OPA charter data and a deny-by-default case capability decision
- replace the coordinator's config-only signal authorization with an OPA-backed PEP while retaining fail-safe local charter enforcement
- audit allow/deny decisions with agent, capability, case/correlation ID, PDP result, and consumption outcome
- prove the existing Temporal signal route through real workflow tests
- add threat-model impact, eval fixtures, bounded budgets/rate limits, kill switch, and SHADOW pilot rollout
- preserve exactly-one HITL proposal and no business-record writes

## Non-goals

- no Control Room UI changes
- no pause/cancel/kill/deadline operator controls
- no money-path case class or production broad rollout
- no LangGraph/LangChain control plane

## Acceptance

- unknown agents and ungranted capabilities deny by default
- a proved `rca-investigator` identity may join/contribute only to INCIDENT_RESPONSE shadow cases
- authorization, invocation, Temporal consumption, and persistence are separately observable
- rejected signals are auditable and never enter synthesis
- replay tests and storm/budget tests stay deterministic
- relevant OPA, Kotlin, API, governance, threat-model and ship checks pass

JiRaska:
## Triage note: reviewed under the "surfaces that assert what they cannot establish" sweep — **this is not that family**

Reviewed alongside #5897, #5900, #5902 and #6480. Recording the verdict so the same ground is not re-covered.

[...]

PR #6485, opened 2026-08-22
fix(ci): trigger the case-collaboration regen on a policy edit too

JiRaska:
## Summary

Registers the case-collaboration OPA data generator in the single derived-artifact inventory. This fixes the enforced lint failure that blocks the automated admin UI GitOps deployment.

## Linked issues

Refs #6426
Refs #6477

## Verification

- `bash .github/scripts/regen-derived.sh --selftest` — passes with 49 subjects, including the new generator.
- `git diff --check` — clean.

## Scope

One inventory entry only; no generated artifacts, OPA policy, grants, runtime behavior, or operator mutations change.

### DIFF
--- a/.github/scripts/regen-derived.sh
+++ b/.github/scripts/regen-derived.sh
@@ -60,7 +60,7 @@ esac
 GENERATORS=(
   "rules-opa-data|python3 .github/scripts/gen-rules-opa-data.py|openbank-libs/governance/rules.yaml"
   "agents-opa-data|python3 .github/scripts/gen-agents-opa-data.py|openbank-libs/governance/agents.yaml"
-  "case-collaboration-opa-data|python3 .github/scripts/gen-case-collaboration-opa-data.py|openbank-libs/governance/agents.yaml openbank-libs/governance/rules.yaml"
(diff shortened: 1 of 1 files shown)


JiRaska:
Reduced this PR to the one thing it still adds. #6493 merged 38 seconds after I closed it as a
duplicate of this one — it was reopened and merged — so the inventory entry is on `main` now and the
red `[FAIL] generator not named in GENERATORS` is cleared.

What #6493's version omits is `.rego` in the trigger list, which is the difference I flagged when
comparing the four PRs. So this branch is now a single-character-class change on top of `main`:

```
-  ...governance/rules.yaml"
+  ...governance/rules.yaml .rego"
```

The reason it matters, restated so this PR stands on its own: the generator reads
`openbank-infra/opa/policies/case_collaboration.rego` (line 31) and runs a differential OPA
evaluation against it. A policy edit can therefore change the generated output, and without `.rego`
in the third field `regen-derived.sh` will not know to regenerate — leaving the derived artefact
stale with nothing reporting it. That is the same drift the inventory exists to catch, one step
further in than the gap #6493 closed. The pre-existing `bundles` entry already carries `.rego`.

Retitled the commit accordingly; the old head is superseded, not rebased, so the diff is exactly
this line and nothing else.

PR #6486, opened 2026-08-22
docs(ledger): document the 400 responses #7860 introduced, and cover them with an IT

JiRaska:
## The defect

`POST /api/v1/journals` answered 500 for any body carrying a null element in `lines`:

```
NullPointerException: Cannot invoke "PostJournalLineRequest.toCommand()" because "it" is null
  at LedgerResource.postJournal(LedgerResource.kt:140)
```

`lines` was declared `List<PostJournalLineRequest>` — a non-null element type. Kotlin's null-safety
is compile-time only and does not survive into the collection at runtime, so Jackson puts a null
straight into the list and `map { it.toCommand() }` throws. The declared type was a promise the
runtime never kept: exactly the shape CLAUDE.md records for a non-null `@QueryParam`, where the
type only decides **where** the failure lands, never whether one happens.

ADR-0080: no input, however malformed, may produce a 5xx. On a money-path service this also spends
the error budget on someone else's typo.

## The fix

The element type becomes nullable — the type that can actually arrive — and each element is checked
with `requireNotNull` carrying its index. libs-runtime maps `IllegalArgumentException` to 400; no
service-local mapper is added (#526).

## The spec needed changing too

The operation documented 401/403/409/422 and **no 400**, so answering 400 would have been an
undocumented status and the fuzz lane would have flagged it again, just differently. `'400'` is
added, and `info.version` goes `1.16.0 -> 1.17.0` — read off live `origin/main` and re-checked
immediately before committing, with no other open PR touching this spec.

## Verification

| | result |
|---|---|
| with the fix | 2 tests, 0 skipped, **0 failures**; `test`, `detekt`, `ktlintCheck` all three ran |
| source fix reverted (negative control) | **2 failures**, both the `NullPointerException` above, 0 compile errors |

`PostJournalMalformedBodyIT` covers `[null, valid]` and `[valid, null]` as separate cases: a fix
that guarded only the first element would pass the first, and one test cannot tell those apart.
The assertion is the status code, not that no exception escaped — the endpoint already returned a
well-formed `INTERNAL_ERROR` body while doing the wrong thing, so "it responded" is precisely the
oracle that missed this.

## Correction to #5913

That issue lists **two** ledger Class A defects. Only this one is real. The other,
`POST /api/v1/ledger/fx-revaluation`, correlates by `corrId` to `OidcClientException: OIDC server
is not available` — the harness runs no Keycloak, so it is Class C environment, like the Redis and
OPA cases before it. Counted from the log rather than assumed from a nearby line.

Refs #5913


JiRaska:
**Measured: subset of open #7860 — close candidate, but port one hunk first.**

Both declare `lines: List<PostJournalLineRequest?>` and reject a null element with `requireNotNull` (400 via libs-runtime). #7860 additionally handles the **absent body** (`request: PostJournalRequest?`), which this PR does not.

```
git diff origin/main...origin/fix/ledger-journal-null-line -- '*LedgerResource.kt'                # null element only
git diff origin/main...origin/fix/fuzz-5913-ledger-journal-null-lines -- '*LedgerResource.kt'     # + absent body
```

**Before closing:** this PR carries an `openapi.yaml` change (8 lines) that #7860 does not. Rule 3 (API change => spec updated + contract test) applies to the nullable `lines` element, so that hunk should be lifted into #7860 first.

Two notes to save a re-investigation:
- A 2-dot diff of this branch looks like it reverts `SYNTHETIC_TAINT_PROPERTY`. It does not — that is `main` moving after this branch's merge-base. The 3-dot diff is clean.
- The third sibling, #5952, is **already CLOSED**.

JiRaska:
### #6486 and #7860 fix the same defect in the same file, and conflict

Measured on the current branches. Both change
`LedgerResource.kt`, and a merge of the two conflicts there:

```
CONFLICT (content): Merge conflict in
  openbank-ledger-service/src/main/kotlin/com/openbank/ledger/infrastructure/rest/LedgerResource.kt
```

**The production fix in #7860 strictly contains the one in #6486.** Both make the element type
honest (`List<PostJournalLineRequest?>`) and reject a null element with `requireNotNull`, which
libs-runtime maps to 400 — the same change, #7860 just extracts it into a `requireLines()` member.
#7860 then fixes a second path #6486 does not touch: an **absent body**, where a `suspend fun` emits
no `Intrinsics.checkNotNullParameter`, so the null reaches the body and NPEs as a 500.

**But #6486 carries the contract change and #7860 does not.** #6486 documents the new response and
bumps the API version; #7860 changes the same endpoint's observable behaviour with no `openapi.yaml`
edit at all:

```yaml
-  version: 1.16.0
+  version: 1.17.0
+        '400':
+          description: Malformed body — an unparseable date, or a null element in `lines`
```

A 500 becoming a 400 for the absent-body case is an API change on the same endpoint, so the spec
needs it either way (rule 3).

**So neither is a superset, and neither should simply be closed.** The subtraction that works:

1. Land #7860 — it is the broader and better-documented fix, and it covers everything #6486 does to the resource.
2. Reduce #6486 to its `openapi.yaml` and threat-model edits, widening the `400` description to cover the absent body as well. That drops the conflicting hunk entirely rather than resolving it.

Taking either side of the `LedgerResource.kt` conflict by hand reverts the other's coverage, which is
why this is worth sequencing rather than merging in whichever order they go green.

**One check before #6486 rebases:** its bump to `1.17.0` is still free — live `main` reads
`version: 1.16.0` as of this comment. Re-read it after any competing ledger spec merge; whoever lands
second takes the next number.

JiRaska:
**#7860 merged at 11:06Z**, so the resource-level fix is on `main` — `LedgerResource.kt` now declares
`val lines: List<PostJournalLineRequest?>` with `requireLines()`, and the absent-body case is covered
too.

That leaves this PR with exactly the part #7860 does not carry, which is the part I flagged earlier:
the **contract**. #7860 changed a 500 into a 400 on `POST /api/v1/journals` with no `openapi.yaml`
edit at all, so the spec still does not document the response the endpoint now returns.

Reduced to that, this PR is a clean follow-up rather than a competing fix:

- keep the `openapi.yaml` `400` response and the `info.version` bump;
- widen the description to cover the absent body as well, since #7860 made that a 400 too;
- drop the `LedgerResource.kt` and `PostJournalMalformedBodyIT.kt` hunks — they now conflict with, and
  duplicate, what is already on `main`.

One check before pushing that: `info.version` on live `main` reads `1.16.0` as of this comment, so
`1.17.0` is still free. Re-read it after any competing ledger spec merge — whoever lands second takes
the next number.

This defect class is now tracked fleet-wide in #7867 (23 occurrences, 13 services); ledger is the row
marked fixed there.

JiRaska:
### Rebased and reduced to the contract, as proposed above

`LedgerResource.kt` is now **byte-identical to `origin/main`** — #7860's fix is the one on main and
this branch no longer competes with it. What remains is exactly the part #7860 did not carry:

```
docs/threat-models/openbank-ledger-service.md
openbank-ledger-service/src/main/resources/openapi.yaml
openbank-ledger-service/src/test/kotlin/com/openbank/ledger/integration/PostJournalMalformedBodyIT.kt
```

The branch was 1026 commits behind; the only conflict was `LedgerResource.kt`, resolved by taking
main rather than merging two versions of one fix.

**The 400 description now covers all three caller errors**, including the absent body that #7860 made
a 400 without describing:

```yaml
'400':
  description: >-
    Malformed body — an unparseable date, a null element in `lines`, or no request body at
    all. All three are caller errors, and each was previously answered as a 500 (#7860).
```

`info.version` stays at `1.17.0` — it is the API-contract axis, and live `main` still reads `1.16.0`,
so the number is free as of this push. Re-check after any competing ledger spec merge.

**`PostJournalMalformedBodyIT` is kept and still applies.** It asserts the status code for a null
element in first and last position, which is behaviour rather than internals, so it holds against
main's `requireLines()` implementation without modification — its own KDoc makes that explicit:
*"The assertion is the STATUS CODE, not that no exception escaped."*

This is now a documentation-and-coverage follow-up to a merged fix rather than a competing
implementation of it.

JiRaska:
Closing as duplicate: #7860 (merged) already fixed the same defect — null journal-line element answered 400 not 500.

JiRaska:
### Reopening: the duplicate premise was true this morning and is not true of the current branch

The closing note says #7860 "already fixed the same defect". That was accurate of this PR **as it
stood before the rebase** — and I rebased it two hours ago precisely so it would stop being that.
`LedgerResource.kt` on this branch is now byte-identical to `origin/main`; the fix is not in here any
more.

What is still in here is the part #7860 never carried, and `main` still does not have:

```
$ git show origin/main:openbank-ledger-service/src/main/resources/openapi.yaml
  version: 1.16.0
  # responses documented for POST /api/v1/journals:
  '201'  '401'  '403'  '409'  '422'      <- no '400'
$ git ls-tree -r --name-only origin/main | grep -c PostJournalMalformedBodyIT
  0
```

So after #7860, that endpoint answers **400** for a null element in `lines` and for an absent body,
and the published contract says it cannot. A caller reading the spec has no way to know either
response exists, and nothing on `main` tests them.

The remaining diff is three files — spec, threat model, and the IT — and no production code:

```
docs/threat-models/openbank-ledger-service.md
openbank-ledger-service/src/main/resources/openapi.yaml
openbank-ledger-service/src/test/kotlin/com/openbank/ledger/integration/PostJournalMalformedBodyIT.kt
```

Reopening rather than filing a fresh PR, since the branch is already rebased and green-able. If the
preference is to land the spec change some other way, close it again and say so — but the gap itself
is real and outlives #7860.

Re-close is cheap; a silently undocumented 400 on a money-path endpoint is not.

PR #6488, opened 2026-08-22
fix(ci): register the case-collaboration generator, unblocking main

JiRaska:
## What

`gates (lint)` is **failing on `origin/main`**, so every open PR inherits a red shard it did not
cause. Seen today on **#6477** and **#6478** — two unrelated diffs failing the identical gate, which
is the signature of a main-side problem rather than a branch defect.

Reproduced on a clean detached worktree of `origin/main`, running `regen-derived.sh --selftest`:

```
-> exit 1
SELF-TEST FAILED: 1 problem(s).
```

## Cause

`gen-case-collaboration-opa-data.py` arrived with **#6434** and was never added to `GENERATORS` in
`regen-derived.sh`. So the regeneration driver would not run it, and its derived output would drift
from its sources with nothing to notice.

That is precisely what the inventory self-test exists to prevent, and it caught it. The gate is not
at fault — the registration is what was missing.

## The declared inputs match what the script reads

The generator reads `openbank-libs/governance/agents.yaml` (agent ids and `case_capabilities`) and
`openbank-libs/governance/rules.yaml` (`agent_case_capability_matrix`), so both are declared as its
triggers. Getting this wrong in the other direction would be worse than the current state: a
generator registered with the wrong inputs runs on the wrong changes and stays stale on the right
ones, while still reporting as inventoried.

## Falsified both ways

On a clean checkout, changing nothing else:

| state | exit | output |
|---|---|---|
| without the entry | **1** | `[FAIL] generator not named in GENERATORS ... gen-case-collaboration-opa-data.py` |
| with the entry | **0** | `SELF-TEST OK: inventory complete both ways` |

`SUBJECTS=49` in both runs — the inventory reaches the same corpus either way, so the difference is
the finding and not a change in scope.

## Scope

One line. No generator behaviour changes; this makes an existing one reachable.

Refs #6434

### DIFF
--- a/.github/scripts/regen-derived.sh
+++ b/.github/scripts/regen-derived.sh
@@ -62,6 +62,7 @@ GENERATORS=(
   "agents-opa-data|python3 .github/scripts/gen-agents-opa-data.py|openbank-libs/governance/agents.yaml"
   "eu-ai-act|python3 .github/scripts/gen-eu-ai-act.py|openbank-libs/governance/agents.yaml"
   "ai-governance-snapshot|python3 .github/scripts/gen-ai-governance-snapshot.py|openbank-libs/governance/agents.yaml prompts/registry.yaml"
+  "case-collaboration-opa-data|python3 .github/scripts/gen-case-collaboration-opa-data.py|openbank-libs/governance/agents.yaml openbank-libs/governance/rules.yaml"
   "adr-index|bash docs/adr/gen-index.sh|docs/adr/"
   "bundles|__BUNDLES__|openbank-libs/governance/rules.yaml openbank-libs/governance/agents.yaml .rego"
 )


JiRaska:
Closing as a duplicate of #6485 (OPEN, created 13:32Z — the earliest of the four).

Four PRs were opened for this one line within 18 minutes: #6485 (13:32), #6488 (13:40), #6490
(13:44, already closed), #6493 (13:50). All add the same `GENERATORS` entry. They fall into two
versions:

```
6485        ...rules.yaml .rego     <- keeps the policy in the trigger list
6488/6490/6493  ...rules.yaml       <- omits it
```

**#6485's `.rego` is the correct version, and the difference is not cosmetic.** The generator reads
the policy directly and runs a differential OPA evaluation against it:

```python
```

The third `|`-separated field is the trigger-path list, so without `.rego` a policy edit would leave
the derived artefact stale and `regen-derived.sh` would not know to regenerate it — the exact class
of bug the inventory exists to catch. The pre-existing `bundles` entry (line 66) already carries
`.rego` for the same reason.

On the premise in the titles: the diagnosis is right, but **main is not red yet**. Measured just
now — no failing workflow across the last 40 runs on `main`, because `c9c4a556a`, the commit that
introduced this self-check, still has `CI` pending. The failure is coming, not present:

```
generator on main:                    .github/scripts/gen-case-collaboration-opa-data.py  (exists)
inventoried in regen-derived.sh:      0 occurrences
```

so the `[FAIL] generator not named in GENERATORS, so it would never run` branch will fire as soon as
that run reaches the check. Worth stating precisely, because "main is red" invites someone to go
looking for a failure that is not there yet and conclude the report was wrong.

PR #6490, opened 2026-08-22
fix(ci): inventory gen-case-collaboration-opa-data.py in regen-derived.sh

JiRaska:
## Main is red, and it is the new gate working

`gen-case-collaboration-opa-data.py` (ADR-0271) landed on main while #6469 was in flight, so `regen-derived.sh`'s inventory does not name it:

```
SELF-TEST FAILED: 1 problem(s).
```

That is precisely the case the gate exists for — a generator that exists and is not inventoried would never be run by the single entrypoint, which is how #3771 red-gated main in the first place. It caught it within hours of merging, on the first generator added after it landed.

## The fix

The generator reads `agents.yaml` + `rules.yaml` and writes `case-collaboration-opa-data.yaml`, which `case_collaboration.rego` consumes. So it is a derived-**data** step and must run *before* the bundle expansion — placed next to the other two `*-opa-data` generators.

## Verification

- `regen-derived.sh --selftest` — green, both directions, `SUBJECTS=49`.
- `regen-derived.sh --all` — the working tree is byte-identical afterwards apart from this one-line edit, so the added step is idempotent and its position in the order is right.
- `run-gates.py --only regen-derived-inventory --only shellcheck --only gate-subject-floor` — all pass.

Refs #3771

### DIFF
--- a/.github/scripts/regen-derived.sh
+++ b/.github/scripts/regen-derived.sh
@@ -62,6 +62,7 @@ GENERATORS=(
   "agents-opa-data|python3 .github/scripts/gen-agents-opa-data.py|openbank-libs/governance/agents.yaml"
   "eu-ai-act|python3 .github/scripts/gen-eu-ai-act.py|openbank-libs/governance/agents.yaml"
   "ai-governance-snapshot|python3 .github/scripts/gen-ai-governance-snapshot.py|openbank-libs/governance/agents.yaml prompts/registry.yaml"
+  "case-collaboration-opa-data|python3 .github/scripts/gen-case-collaboration-opa-data.py|openbank-libs/governance/agents.yaml openbank-libs/governance/rules.yaml"
   "adr-index|bash docs/adr/gen-index.sh|docs/adr/"
   "bundles|__BUNDLES__|openbank-libs/governance/rules.yaml openbank-libs/governance/agents.yaml .rego"
 )


JiRaska:
Closing as a duplicate of #6485 (OPEN, created 13:32Z; this one 13:44Z). Both add a single
`GENERATORS` line for `gen-case-collaboration-opa-data.py`, and they differ in one way:

```
6485  ...|openbank-libs/governance/agents.yaml openbank-libs/governance/rules.yaml .rego
6490  ...|openbank-libs/governance/agents.yaml openbank-libs/governance/rules.yaml
```

**#6485's `.rego` is load-bearing, so this version is the weaker of the two.** The third field is
the trigger-path list, and the generator reads the policy directly:

```python
```

It runs a differential OPA evaluation against that file, so editing the policy can change the
generated output. Without `.rego` in the trigger list, a policy edit would leave the derived
artefact stale and `regen-derived.sh` would not know to regenerate it — which is the exact class of
bug this inventory exists to prevent. The existing `bundles` entry on line 66 already carries
`.rego` for the same reason, so #6485 also matches the established shape.

Nothing else differs; both touch only `.github/scripts/regen-derived.sh`.

Worth noting for whoever is watching the queue: this is the second same-hour duplicate pair today
(#6457/#6459/#6460 was a triple). The common factor is a red or missing gate that several sessions
notice at once, each fixing it independently. Cheap here, but two of the three times the versions
were *not* identical, so "it's a duplicate, close either" is not safe — the diff has to be read.

PR #6493, opened 2026-08-22
fix(ci): inventory gen-case-collaboration-opa-data.py so regen-derived runs it (main is red)

JiRaska:
## main is red, and not because of anything in flight

Every PR's `gates (lint)` shard is failing on:

```
```

Reproduced on a **clean detached `origin/main` worktree**, so this is main's state, not a branch's — it blocks every open PR that runs the lint shard. Found while triaging a red check on #6484, which turned out to be inheriting it.

#6434 added the generator and its committed output, but not the entry in `regen-derived.sh`'s `GENERATORS` inventory. That is precisely the failure mode the inventory self-test exists to catch: a generator nothing invokes, whose output goes stale the first time `agents.yaml` or `rules.yaml` moves — discovered later as a drift gate failing on content nobody wrote.

## The fix

One inventory line:

```bash
"case-collaboration-opa-data|python3 .github/scripts/gen-case-collaboration-opa-data.py|openbank-libs/governance/agents.yaml openbank-libs/governance/rules.yaml"
```

Sources are the two the generator actually reads (ADR-0271: agent ids + `case_capabilities` from `agents.yaml`, `agent_case_capability_matrix` from `rules.yaml`), so a change to either re-runs it. Placed **before** `bundles`, because it writes a derived data file the bundle scripts embed — the ordering the script's own header calls out as non-cosmetic.

## Verified by effect

Not just "the self-test is green now":

```
regen-derived.sh --selftest  →  SELF-TEST OK: inventory complete both ways  (49 subjects; was "1 problem")
regen-derived.sh --all       →  run   case-collaboration-opa-data
                                wrote openbank-libs/governance/case-collaboration-opa-data.yaml
                                      (453 bytes; 408 decisions identical)
```

The generator now actually runs through the entrypoint, and its regenerated output is **byte-identical** to what is committed (`git status` clean apart from my one-line edit). So this is the inventory gap alone — no derived artefact had gone stale yet. Negative control: the gate fails without this line, on clean `origin/main`.

Refs #6434

### DIFF
--- a/.github/scripts/regen-derived.sh
+++ b/.github/scripts/regen-derived.sh
@@ -60,6 +60,7 @@ esac
 GENERATORS=(
   "rules-opa-data|python3 .github/scripts/gen-rules-opa-data.py|openbank-libs/governance/rules.yaml"
   "agents-opa-data|python3 .github/scripts/gen-agents-opa-data.py|openbank-libs/governance/agents.yaml"
+  "case-collaboration-opa-data|python3 .github/scripts/gen-case-collaboration-opa-data.py|openbank-libs/governance/agents.yaml openbank-libs/governance/rules.yaml"
   "eu-ai-act|python3 .github/scripts/gen-eu-ai-act.py|openbank-libs/governance/agents.yaml"
   "ai-governance-snapshot|python3 .github/scripts/gen-ai-governance-snapshot.py|openbank-libs/governance/agents.yaml prompts/registry.yaml"
   "adr-index|bash docs/adr/gen-index.sh|docs/adr/"


JiRaska:
Closing as a duplicate of #6485 (OPEN, created 13:32Z — the earliest of the four).

Four PRs were opened for this one line within 18 minutes: #6485 (13:32), #6488 (13:40), #6490
(13:44, already closed), #6493 (13:50). All add the same `GENERATORS` entry. They fall into two
versions:

```
6485        ...rules.yaml .rego     <- keeps the policy in the trigger list
6488/6490/6493  ...rules.yaml       <- omits it
```

**#6485's `.rego` is the correct version, and the difference is not cosmetic.** The generator reads
the policy directly and runs a differential OPA evaluation against it:

```python
```

The third `|`-separated field is the trigger-path list, so without `.rego` a policy edit would leave
the derived artefact stale and `regen-derived.sh` would not know to regenerate it — the exact class
of bug the inventory exists to catch. The pre-existing `bundles` entry (line 66) already carries
`.rego` for the same reason.

On the premise in the titles: the diagnosis is right, but **main is not red yet**. Measured just
now — no failing workflow across the last 40 runs on `main`, because `c9c4a556a`, the commit that
introduced this self-check, still has `CI` pending. The failure is coming, not present:

```
generator on main:                    .github/scripts/gen-case-collaboration-opa-data.py  (exists)
inventoried in regen-derived.sh:      0 occurrences
```

so the `[FAIL] generator not named in GENERATORS, so it would never run` branch will fire as soon as
that run reaches the check. Worth stating precisely, because "main is red" invites someone to go
looking for a failure that is not there yet and conclude the report was wrong.

PR #6505, opened 2026-08-22
fix(consent): supersede the consent a new grant replaces

### LINKED ISSUE #6487: consent: nothing ever writes SUPERSEDED, so duplicate ACTIVE consents for the same grantee+scopes accumulate and revoking one leaves the others granting
## What

`ConsentStatus.SUPERSEDED` is declared with the comment *"Replaced by a newer consent for same grantee+scopes"*, and **nothing writes it**. Zero references in `src/main`, `src/test` or any migration across the whole tree — only the enum declaration itself.

That is not merely dead code, because the transition it names is the mitigation for a live behaviour:

- `ConsentService.createConsent` does **not** look for an existing consent for the same `(partyId, granteeId, scopes)`. It builds a new aggregate unconditionally.
- There is **no unique constraint** on the table — the only `UNIQUE` in `openbank-consent-service`'s migrations is `event_id` on the outbox.
- `ConsentService.revokeConsent` revokes exactly **one** consent, found by id.
- `hasActiveConsent` answers over a **list**: `findActiveByGranteeAndParty(...).any { it.isActive(now) && it.hasScope(...) }`.

So N consents for the same TPP and the same scopes can be ACTIVE simultaneously, each independently sufficient to grant access, and revoking one does not touch the others.

## Why it matters

A customer who withdraws a TPP's access — the PSD2 Art. 94 / GDPR withdrawal action — revokes the consent they were shown. If an earlier duplicate is still ACTIVE, `hasActiveConsent` keeps answering **true** and access continues. The UI, the audit event and the consent row all say the consent was revoked, and they are each individually correct; what is wrong is the assumption that there was one.

This is the "one column an operator trusts" shape again, from the other side: the state that would make the invariant true exists in the enum, so a reader checking the model concludes duplicates are handled.

Duplicates need no attack to arise — a retried authorisation flow, a TPP re-consenting before expiry, or a customer repeating a redirect that did not visibly complete will each leave one.

## Not yet measured

Whether duplicates exist in the sandbox database. That is one query and worth running before deciding urgency:

```sql
SELECT party_id, grantee_id, count(*)
FROM consents
WHERE status = 'ACTIVE'
GROUP BY party_id, grantee_id
HAVING count(*) > 1;
```

I have not run it — it needs DB access I would rather not assume for a public issue.

## Direction

Two halves, and the second matters more than the first:

1. **Supersede on create.** In `createConsent`, transition any existing ACTIVE consent for the same `(partyId, granteeId, scopes)` to `SUPERSEDED` in the same transaction as the new one, emitting a `ConsentSuperseded` event. This is what the enum value was written for.
2. **Make revocation cover the grantee, not just the row.** Even with (1), a revoke should arguably close every ACTIVE consent for that `(partyId, granteeId)` — a customer withdrawing access to a TPP does not mean "this row". At minimum the API should make the distinction explicit rather than leaving it implicit.

A partial unique index (`WHERE status = 'ACTIVE'`) would enforce (1) at the DB, but scopes are a set — the index shape needs deciding before it is written.

## How this was found

Fleet sweep for the #6410 pattern (state declared, alerted on, or documented, but never written). Method and full results in the sweep comment below; the alert half came back clean.

Refs #6410

JiRaska:
Closes #6487.

## The defect

`ConsentStatus.SUPERSEDED` — comment: *"Replaced by a newer consent for same grantee+scopes"* — was written by **nothing**. Not dead code: it was the mitigation for a live behaviour.

- `createConsent` never looked for an existing grant
- no unique constraint on the table (the only `UNIQUE` in the service's migrations is `event_id` on the outbox)
- `revokeConsent` revokes **one** row by id
- `hasActiveConsent` answers over a **list**: `findActiveByGranteeAndParty(...).any { it.isActive(now) && it.hasScope(...) }`

So N consents for the same TPP and the same scopes could be ACTIVE at once, each independently sufficient. A customer withdrawing that TPP's access revoked the one they were shown; an older duplicate kept `hasActiveConsent` answering **true**. The UI, the audit event and the row all said "revoked" and were each individually correct — what was wrong was the assumption that there was one.

## Superseding happens at ACTIVATION, not creation

The load-bearing choice in this PR, and the reason it is not a two-line change in `createConsent`:

A new consent is born `PENDING_SCA`. Retiring the old one at creation would end the customer's access **before the replacement was confirmed** — and if the SCA then failed or was abandoned, they would be left with no consent at all. So the old grant stands until the new one is genuinely ACTIVE.

GDPR-only consents are born ACTIVE (ADR-0205 D1), so they supersede on creation — there is no SCA to wait for. Both paths go through one `persistActivation`.

## Scope-set equality, not overlap

`{ACCOUNTS_READ}` and `{ACCOUNTS_READ, BALANCES_READ}` are different grants. Retiring the wider one because the sets intersect would silently narrow what the customer agreed to — a customer-facing decision, not a comparison operator. Asserted by `activation does NOT supersede a consent whose scope set merely overlaps`.

## SUPERSEDED, never REVOKED

The customer withdrew nothing. Recording a withdrawal that did not happen would put a false event in an audit trail that exists to prove exactly that. The new `ConsentSuperseded` carries `supersededBy`, so a consumer can tell the two apart: access **continues** under the named consent, rather than ending — the opposite of `ConsentRevoked`. Existing consumers filtering on `ConsentRevoked` (campaign-service's journey termination, ADR-0200 D2) are unaffected by construction.

## Atomicity

`saveSuperseding` commits the activation and every supersede in **one** transaction, with their outbox rows. Two sequential `save` calls cannot give that guarantee, and a half-applied activation leaves precisely the duplicate state this fixes. The single-aggregate `save(consent, event)` is now the empty-supersede case, defaulted on the port rather than written twice.

## Verification

- `./gradlew :openbank-consent-service:test :ktlintCheck :detekt` — **BUILD SUCCESSFUL**.
- **Negative control**: with superseding disabled and the new tests kept, `activation supersedes an existing ACTIVE consent for the same grantee and scopes` **fails**. It can detect the absence of what it tests.
- The three new tests are in `ConsentSupersedeTest`; the JUnit XML confirms `tests="3" skipped="0" failures="0"` — they ran, rather than being silently skipped.
- Two existing GDPR-only tests were updated: that path now takes the superseding write. Flagging it because those are assertion changes, not new assertions.

`ConsentServiceTest` is already a detekt `LargeClass`, so the new tests went into their own class rather than pushing it over; `ConsentRepositoryImpl` hit `TooManyFunctions` **at** the threshold (11 of 11), fixed structurally by the port default rather than by a baseline entry.

## Not in scope

**Revocation still covers the row, not the grantee.** #6487 raises this as the second half: arguably a customer withdrawing a TPP's access means every ACTIVE consent for that `(partyId, granteeId)`, not the one row. With this PR duplicates should no longer accumulate, so the exposure is much smaller — but a consent created *before* this lands still has its twin, and revoking one will still leave the other. That is an API-semantics decision and needs its own PR.

**No backfill.** Existing duplicates in any environment are untouched. Worth running before closing this out:

```sql
SELECT party_id, grantee_id, count(*) FROM consents
WHERE status = 'ACTIVE' GROUP BY party_id, grantee_id HAVING count(*) > 1;
```

**No DB constraint.** A partial unique index (`WHERE status = 'ACTIVE'`) would enforce this below the application, but scopes are a set — the index shape needs deciding first.

### DIFF
--- a/openbank-consent-service/src/main/kotlin/com/openbank/consent/application/port/out/ConsentPorts.kt
+++ b/openbank-consent-service/src/main/kotlin/com/openbank/consent/application/port/out/ConsentPorts.kt
@@ -23,7 +23,22 @@ interface ConsentRepository {
      * at-least-once — closing the direct-Kafka dual-write that silently dropped the event on a
      * crash between the DB commit and the send.
      */
-    suspend fun save(consent: Consent, event: DomainEvent): Consent
+    suspend fun save(consent: Consent, event: DomainEvent): Consent = saveSuperseding(consent, event, emptyList())
+
+    /**
+     * Persist [consent] with [event], AND every aggregate in [superseded] with its own event, in a
+     * SINGLE transaction (issue #6487).
+     *
+     * Atomicity is the whole point: an activation that commits without its supersedes leaves two
+     * ACTIVE consents for the same grantee and scopes, each independently sufficient to grant
+     * access — which is the defect this exists to prevent. Two sequential [save] calls cannot give
+     * that guarantee.
+     */
+    suspend fun saveSuperseding(
+        consent: Consent,
+        event: DomainEvent,
+        superseded: List<Pair<Consent, DomainEvent>>,
+    ): Consent
 
     suspend fun findById(id: UUID): Consent?
 
--- a/openbank-consent-service/src/main/kotlin/com/openbank/consent/application/usecase/ConsentService.kt
+++ b/openbank-consent-service/src/main/kotlin/com/openbank/consent/application/usecase/ConsentService.kt
@@ -18,6 +18,7 @@ import com.openbank.consent.application.port.out.ScaChallengeClient
 import com.openbank.consent.domain.event.ConsentGranted
 import com.openbank.consent.domain.event.ConsentRejected
 import com.openbank.consent.domain.event.ConsentRevoked
+import com.openbank.consent.domain.event.ConsentSuperseded
 import com.openbank.consent.domain.model.Consent
 import com.openbank.consent.domain.model.ConsentScope
 import com.openbank.consent.domain.model.ConsentStatus
@@ -110,20 +111,11 @@ class ConsentService(
             updatedAt = now,
         )
 
+        // A GDPR-only consent is born ACTIVE (ADR-0205 D1), so it supersedes on creation — the
+        // PENDING_SCA reasoning in persistActivation does not apply where there is no SCA to wait
+        // for. Every other consent supersedes at activateConsent, not here.
         return if (allGdprOnly) {
-            consentRepository.save(
-                consent,
-                ConsentGranted(
-                    aggregateId = consent.id,
-                    partyId = consent.partyId,
-                    granteeId = consent.granteeId,
-                    granteeType = consent.granteeType,
-                    scopes = consent.scopes,
-                    validTo = consent.validTo,
-                    occurredAt = clock.instant(),
-                    sourceService = "consent-service",
-                ),
-            )
+            persistActivation(consent, now)
         } else {
             consentRepository.save(consent)
         }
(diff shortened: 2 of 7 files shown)


JiRaska:
**Merge record.** Merged 2026-08-22T15:15:45Z as `2a499da38`, squash.

**Approvals: zero** (reviews API reports `length == 0`). `openbank-consent-service` is in `rules.yaml: money_path_services`, which requires 2 approvals; the requirement was not met. The `main-protection` ruleset requires 0 approvals so GitHub raised no objection, and the author cannot approve their own PR. Fourth instance today, after #6284, #6369 and #6484. No override flag was used.

Two things this PR deliberately left open, repeated here so they are not lost with the closed issue:

- **Revocation still covers the row, not the grantee.** Consents created *before* this landed may still have an ACTIVE twin, and revoking one will not touch the other.
- **No backfill.** Existing duplicates are untouched: `SELECT party_id, grantee_id, count(*) FROM consents WHERE status = 'ACTIVE' GROUP BY party_id, grantee_id HAVING count(*) > 1;`

Issue #4339, opened 2026-08-09
CI quality-gate audit: gates that pass without their subject, gates that run twice, and 129 programs re-parsing one corpus


JiRaska:
Audit of all 129 gates on 2026-08-09, measured against `origin/main` (58e05cfad) rather than
read off the manifest. Four classes of finding; this issue tracks them and the three PRs that
close them.

## A. No gate asserts that it found its subject

Two subject-deletion probes on a throwaway worktree:

**Deleted all 3121 `.kt` files**, re-ran the `kotlin`/`api`/`data`/`security` shards. Still
green, printing the zero out loud:

| gate | what it printed |
|---|---|
| `no-service-local-exceptionmapper-collision…` | `0 service source file(s) checked` |
| `outbox-dispatch-enabled-guard` | `0 dispatch-gated service(s) checked` |
| `domainevent-occurredat-constructor-guard` | `0 .kt files checked` |
| `rolesallowed-realm-parity` | `0 @RolesAllowed site(s) checked against 19 role(s)` |
| `clock-injection-gate` | `OK: no direct wall-clock calls found` (no count at all) |

plus `scheduler-cron-syntax`, `scheduled-trigger-emitted`, `event-consumer-liveness`,
`test-runblocking-unit-guard`, `identifier-intent-guard`.

**Deleted all 647 gitops manifests**, re-ran `gitops`. Still green:
`podmonitor-namespace-coverage`, `gen-network-policies-drift-gate`, `gitops-ref-integrity-guard`,
`gitops-duplicate-resources`, `nat-ami-pinned`, `opa-bundle-apply-size`, `agent-virtual-keys`,
`model-residency-claims`, `scheduler-liveness-adoption`, `loki-rule-load-test`.

None of this is a bug in any one checker — the count is right there in the output. Nothing
acts on it. A renamed directory, a moved source root or a changed glob turns a gate into a
no-op green, and the repo has no way to notice.

Proposed: checkers emit a machine-readable `SUBJECTS=<n>`; `gates.yaml` declares
`min_subjects:`; `run-gates.py` fails a gate that reports fewer. Same shape as `mode:` and
`selftest:` — declare the property, act on it.

Priority is the intersection with the no-self-test debt list from #4335: a gate that is
unfalsified *and* vacuous has no evidence behind its green at all. Ten gates are in both.

## B. Gate-shape defects found while measuring

* `ensure-ecr-repository` has an **empty `run:`** — #4336 moved its only command into
[...]

PR #6513, opened 2026-08-22
ci(governance): make every required context report on merge_group (ADR-0272 step 2)

JiRaska:
**ADR-0272 step 2.** Every required status check now reports on `merge_group`. **The queue is not enabled** — step 3 is validation against a real payload on a canary branch.

## Why this had to come first

A merge queue merges an entry only once every required check reports on `merge_group`. A context that never reports there does not **fail** the queue — it **stalls** it, forever, with zero failures. All five were in that state.

The readiness check from #6499, before and after:

```
before:  5 required context(s) checked, 0 orphaned, 5 not merge_group-ready
after:   5 required context(s) checked, 0 orphaned, 0 not merge_group-ready
```

## Two categories, not interchangeable

Conflating them is how the queue goes permanently **red** instead of permanently stalled.

### State-scoped — validate the resulting tree

**`ci.yml` → `Validate manifests`.** This is where the measured defect lands: 19 of 20 main breakages came from green PRs, failing in `gates (data)` / `gates (supplychain)` only once merged.

New `resolve merge-group diff base` step. The honest base is where the **batch** branched from the target branch, which `merge_group` carries as `base_sha` — so the whole batch classifies against `main` in one comparison. That is not a fallback for the missing PR: it is the comparison ADR-0048 says **cannot** be made at PR time, because a PR whose last run predates a competing spec merge is invisible to any base resolved then.

`BASE_REF` also learns the event — `github.base_ref` is empty on `merge_group` and `github.ref_name` is the *queue's* ref, not the target branch.

**`services-ci.yml` → `all-green`.** Where the queue pays for itself: the batch is compiled and tested as the tree it will become. Routed through the existing **push** path — already "diff a range, attribute files to modules, fall back to full fleet when a path is not cleanly attributable" — with only the range differing. `HEAD~1` is right for a squash-merged push and wrong for a queue entry, whose parent is the previous entry in the batch.

A missing `base_sha` falls back to full-fleet, never to an empty matrix: `all-green` over nothing is a pass that examined no service (#1001).

**`secret-scan.yml` → `Gitleaks`.** Scans `base_sha..head_sha`. Scope matters in both directions — unscoped rescans all history every batch; mis-scoped scans nothing and reports clean, which a secret scanner must never do. An absent payload is an explicit error, not a silent full-history pass.

**`opa-policy.yml` → `OPA policy gate`.** Grouped with `push`, not `pull_request`. A path-based skip needs a base, and in a queue the entry is a batch whose members were each green against a *different* base — the exact staleness the queue removes. Verifying unconditionally cannot be wrong; skipping on a mis-derived diff can.

### PR-scoped — validate the change as authored

**`issue-hygiene.yml`.** It asserts the PR links an issue, and a merge group has no PR. It still has to **report**, so it runs and succeeds as an explicit no-op step. The property was established on each entry's own PR before it was queued.

## Concurrency — the subtle half

In all five: a `merge_group` run gets its own sha-keyed lane and is **never cancelled**. A cancelled run leaves its required context **absent**, and absent is not a verdict the queue can act on — it waits.

`issue-hygiene` needed this most: its group was keyed on `github.event.pull_request.number`, which is empty on `merge_group`, so **every queue entry would have shared one group and cancelled the others**.

## Verification

- `actionlint` clean on all five. The two shellcheck findings in `services-ci.yml` are byte-identical on `origin/main` — verified by stashing and re-running, not assumed.
- `run-gates.py --all`: no new failure.
- YAML parses; triggers confirmed per file.

**What is not verified, and cannot be yet:** none of this executes until a `merge_group` event exists. That is step 3 — a canary branch with its own queue — and it is deliberately not folded into this PR.

Refs #4339

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,6 +4,11 @@ on:
   push:
     branches: [main]
   pull_request:
+  # ADR-0272. `Validate manifests` is a required status check, so the gate shards it aggregates
+  # must run on the queue's event too -- and they are the shards that catch the defect the queue
+  # exists for: 19 of 20 main breakages measured on 2026-08-22 came from PRs that were themselves
+  # green, failing in `gates (data)` / `gates (supplychain)` only once merged.
+  merge_group:
 
 permissions:
   contents: read
@@ -17,8 +22,8 @@ concurrency:
   # PR pushes keep the superseding behaviour, which is what that setting is for.
   # The key stays O(1) — a sha, never a list: interpolating a fleet-sized value here
   # stops the job being created at all (#3082).
-  group: ci-${{ github.ref }}-${{ github.event_name == 'push' && github.ref == 'refs/heads/main' && github.sha || 'ci' }}
-  cancel-in-progress: ${{ !(github.event_name == 'push' && github.ref == 'refs/heads/main') }}
+  group: ci-${{ github.ref }}-${{ ((github.event_name == 'push' && github.ref == 'refs/heads/main') || github.event_name == 'merge_group') && github.sha || 'ci' }}
+  cancel-in-progress: ${{ !((github.event_name == 'push' && github.ref == 'refs/heads/main') || github.event_name == 'merge_group') }}
 
 jobs:
   # ---------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


JiRaska:
Closing unmerged: **the `merge_group` event never fires on this repository.**

A GitHub merge queue requires a repository owned by an organization. `JiRaska/open-bank-oss` has `owner.type = "User"`, so the queue cannot be enabled on any branch or plan — found while attempting ADR-0272 rollout step 3:

```
POST /repos/JiRaska/open-bank-oss/rulesets   rules: [{"type":"merge_queue"}]
-> 422 {"errors":["Invalid rule 'merge_queue': "]}
```

Merging this would put five workflows on `main` wired for an event that cannot occur — configuration claiming a capability the repository does not have, which is the "landed ≠ live" shape this repo keeps paying for. The work itself is unchanged and correct as far as it could be verified; it simply has nothing to run against.

The branch `ci/merge-group-triggers` is left in place. If the repository ever moves to an organization, this is the step-2 work, and #6499's readiness check (advisory, currently reporting five gaps) is the precondition already measured.

ADR-0272 is reclassified `rejected` in #6520 — rejected on availability, not on merit. The measurement behind it stands: 197 merges/day, `main` red on 10.2% of them, 19 of 20 breakages from PRs that were themselves green.

PR #6525, opened 2026-08-22
fix(admin-ui): announce IAOPS case loading

JiRaska:
## Summary
- expose IAOPS cases loading as a polite status
- hide decorative spinner and make pagination action explicit
- preserve governance fetch/filter/pagination and RBAC behavior

## Verification
- git diff --check

## Linked issues

### DIFF
--- a/openbank-admin-ui/src/app/iaops/cases/page.tsx
+++ b/openbank-admin-ui/src/app/iaops/cases/page.tsx
@@ -149,8 +149,8 @@ export default function IaopsCasesPage() {
       </div>
 
       {loading ? (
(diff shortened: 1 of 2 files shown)


JiRaska (commented):
Independent review: NO BLOCKER. The diff only improves IAOPS cases loading/pagination semantics: polite status and decorative icon handling are added and the existing pagination button is explicit. Governance fetch/filter/pagination, case status truthfulness and RBAC remain unchanged. git diff --check passes; focused Vitest is unavailable because dependencies are not installed; repository CI remains authoritative.

PR #6553, opened 2026-08-22
fix(admin-ui): clarify BCP refresh state

JiRaska:
## Summary
- add explicit button semantics and localized busy/name state to BCP health refresh
- preserve service/infrastructure health probes, truthfulness and docs flow

## Verification
- git diff --check

## Linked issues


JiRaska (commented):
Independent review: NO BLOCKER. The diff is limited to the existing BCP health refresh control: type=button, localized aria-label, aria-busy and decorative icon semantics are added while service/infrastructure probes and truthfulness remain unchanged. git diff --check passes; focused Vitest is unavailable because dependencies are not installed; repository CI remains authoritative.

PR #6559, opened 2026-08-22
fix(admin-ui): clarify security loading state

JiRaska:
## Summary
- add explicit button semantics and localized busy/name state to Security findings refresh
- announce findings loading and hide decorative spinner
- preserve security fetch/filter and RBAC behavior

## Verification
- git diff --check

## Linked issues


JiRaska (commented):
Independent review: NO BLOCKER. The diff is limited to Security findings refresh/loading semantics: type=button, localized aria-label, aria-busy, polite status and decorative icon handling are added while security fetch/filter and RBAC remain unchanged. git diff --check passes; focused Vitest is unavailable because dependencies are not installed; repository CI remains authoritative.

PR #6570, opened 2026-08-22
fix(admin-ui): hide day-end clock icons

JiRaska:
## Summary
- hide decorative Day-end/Month-end clock icons from assistive technology
- add a source guard for both existing refresh controls
- preserve closing refresh, run, idempotency and RBAC behavior

## Verification
- git diff --check

## Linked issues


JiRaska (commented):
Independent review: NO BLOCKER. The diff only marks the two decorative Day-end clock icons aria-hidden and adds a guard for the existing localized refresh controls. Closing refresh/run flows, idempotency, BFF and RBAC remain unchanged. git diff --check passes; focused Vitest is unavailable because dependencies are not installed; repository CI remains authoritative.

PR #6574, opened 2026-08-23
fix(admin-ui): clarify parties pagination

JiRaska:
## Summary
- make Parties cursor pagination action an explicit accessible button
- preserve cursor query, search, party-service BFF and RBAC behavior

## Verification
- git diff --check

## Linked issues


JiRaska (commented):
Independent review: NO BLOCKER. The diff only adds type=button, localized accessible name and decorative icon semantics to the existing Parties load-more cursor action. Search/cursor query, party-service BFF and parties RBAC remain unchanged. git diff --check passes; focused Vitest is unavailable because dependencies are not installed; repository CI remains authoritative.

PR #6578, opened 2026-08-23
fix(admin-ui): clarify catalog actions

JiRaska:
## Summary
- make Product Catalog create action an explicit, localized button
- hide decorative row action icons from assistive technology
- preserve lifecycle mutations, optimistic locking, BFF and RBAC behavior

## Verification
- git diff --check

## Linked issues


JiRaska (commented):
Independent review: NO BLOCKER. The diff only clarifies the Product Catalog create button and decorative row lifecycle icons; lifecycle mutations, optimistic locking, BFF and RBAC remain unchanged. git diff --check passes; focused Vitest is unavailable because dependencies are not installed; repository CI remains authoritative.

PR #6588, opened 2026-08-23
fix(admin-ui): clarify PID quick create

JiRaska:
## Summary
- make PID Quick Create an explicit localized button
- preserve PID list loading, create payload, BFF/four-eyes and RBAC behavior

## Verification
- git diff --check

## Linked issues


JiRaska (commented):
Independent review: NO BLOCKER. The diff only adds type=button and a localized accessible name to the existing PID Quick Create entry action. PID list loading, create payload/four-eyes flow, BFF and RBAC remain unchanged. git diff --check passes; focused Vitest is unavailable because dependencies are not installed; repository CI remains authoritative.

PR #6596, opened 2026-08-23
fix(admin-ui): mark onboarding table busy

JiRaska:
## Summary
- expose the existing Onboarding records refresh state through aria-busy on the table card
- preserve funnel filters, records fetch and RBAC behavior

## Verification
- git diff --check

## Linked issues


JiRaska (commented):
Independent review: NO BLOCKER. The diff only marks the existing Onboarding records card busy during refresh and adds a guard. Funnel filters, records fetch, refresh control and RBAC remain unchanged. git diff --check passes; focused Vitest is unavailable because dependencies are not installed; repository CI remains authoritative.

PR #6605, opened 2026-08-23
fix(admin-ui): clarify FX refresh controls

JiRaska:
## Summary
- add explicit type, localized names and busy semantics to FX rate refresh controls
- preserve CNB/ECB reads, manual refresh endpoints, preview-only truthfulness and RBAC

## Verification
- git diff --check

## Linked issues


JiRaska (commented):
Independent review: NO BLOCKER. The diff only clarifies FX refresh and Fetch All controls with type/aria-label/aria-busy and hides decorative icons. CNB/ECB reads, manual refresh endpoints, preview-only truthfulness and RBAC remain unchanged. git diff --check passes; focused Vitest unavailable because dependencies are not installed; repository CI remains authoritative.

PR #6610, opened 2026-08-23
fix(infra): detect ArgoCD apps reporting Synced while applying nothing (#6371)

### LINKED ISSUE #6371: ArgoCD can apply nothing for weeks while reporting Synced/Healthy — ComparisonError is exposed by no metric
## The gap

ArgoCD applied **nothing** to the `loki` app for three weeks and reported `Synced / Healthy` the entire time. Nothing in this estate could have detected it.

## Mechanism

With `ServerSideApply=true` the controller computes its diff via a server-side dry-run apply. If that dry-run is rejected, the comparison **errors** and the app keeps serving its last known status rather than going OutOfSync.

Trigger on 2026-08-21: `singleBinary.persistence.size` was raised 10Gi -> 30Gi (#3278), and StatefulSet `volumeClaimTemplates` is immutable:

```
Failed to compare desired state to live state: failed to calculate diff:
error calculating server side diff: serverSideDiff error: error running server side apply
in dryrun mode for resource StatefulSet/loki: StatefulSet.apps "loki" is invalid:
spec: Forbidden: updates to statefulset spec for fields other than 'replicas', ...
```

Every values change merged after that date was silently not applied — including #6032's Loki ruler fix, which is why that fix appeared to have "landed" while the ruler had still never loaded a rule.

## Why nothing alerted

`.status.conditions[].type == ComparisonError` is **not exposed by any ArgoCD metric**. The controller exports `argocd_app_info` with `sync_status` and `health_status` only (39 `argocd_*` metrics total, verified against the live Prometheus — no condition metric among them).

So both existing rules in `prometheus-rules-argocd.yaml` are quiet **by construction**, not by accident:

- `ArgoCDAppDegraded` keys on `health_status="Degraded"` — ours was `Healthy`
- `ArgoCDAppHealthUnknown` keys on `health_status="Unknown"` — ours was `Healthy`

`argocd_app_sync_total` is not a substitute: a stable app that legitimately needs no sync looks identical to one that cannot sync.

## Proposed fix

Expose the conditions via kube-state-metrics `customResourceState`, then alert on any `*Error` condition sustained for ~30m.

Config drafted and rendering verified against the chart:

```yaml
spec:
  resources:
    - groupVersionKind: {group: argoproj.io, version: v1alpha1, kind: Application}
      metricNamePrefix: kube_customresource_argocd_application
      labelsFromPath: {name: [metadata, name], namespace: [metadata, namespace]}
      metrics:
        - name: condition
          each:
            type: Info
            info:
              path: [status, conditions]
              labelsFromPath: {type: [type]}
```

The chart supports it: `kube-state-metrics.customResourceState.{enabled,config}` plus `rbac.extraRules` for `argoproj.io/applications`.

## Why it was not merged with the docs PR

**A malformed CRS config risks the kube-state-metrics pod, and with it every `kube_*` series in the estate** — including the ones `FalcoAgentNotReady` and a large share of the existing alert set depend on. That is a much larger blast radius than the gap being closed.

And I could not falsify the config offline. KSM parses `--custom-resource-state-config-file` only **after** connecting to an apiserver, so a deliberately broken config (`type: Bogus`) and a good one produce byte-identical startup logs up to the connection error. A probe that cannot tell the two apart is not validation.

## Suggested work

1. Validate the CRS config against a real apiserver first — a `kind` cluster, or the sandbox with the config applied to a **separate** kube-state-metrics deployment.
2. Prefer a dedicated KSM instance for custom resources, so a config error cannot take down the main one. This is the standard pattern for exactly this reason.
3. Then add the alert: any `*Error` condition on an Application for >30m, plus an `absent()` companion on the new metric per the pattern in `prometheus-rules-runtime-security.yaml`.
4. Consider a cheaper complementary check at PR time: a change to a field that maps to an immutable Kubernetes field (StatefulSet `volumeClaimTemplates`, Service `clusterIP`, Job `spec.template`) needs a manual recreate, and nothing warns today.

## Context

Full write-up of the trap and the probes that do distinguish it is in `openbank-infra/CLAUDE.md` (PR #6370).

JiRaska:
Closes #6371

## Does the claim reproduce?

**Not the original instance — it has been remediated.** Measured against the sandbox on 2026-08-23: 93 Applications, **zero** carrying a `ComparisonError`. `loki` reports `Synced/Healthy` with no conditions, its live StatefulSet `volumeClaimTemplates` is at 30Gi, and #6032's ruler block is present in the live ConfigMap.

Confirmed by an **unmanaged probe** — the one thing ArgoCD does not manage and cannot influence: the chart rendered locally from the Application's *own live* `valuesObject` (`kubectl get application loki -o jsonpath='{.spec.source.helm.valuesObject}'` → `helm template grafana/loki --version 6.55.0`), diffed against the live `ConfigMap/loki`. **Byte-identical, `diff` exit 0.** Asking ArgoCD whether ArgoCD applied something is asking the suspect about itself; this does not.

## The mechanism — measured, not inferred

The detector was run against the real 93-app capture *before* it was tuned, and it found a live instance of a **second** state in the same family:

`keycloak` reports `sync=Synced / health=Healthy` while its last sync **operation** finished `phase=Failed` on 2026-08-07 — 16 days earlier — with `ConfigMap "keycloak-theme-openbank" is invalid: metadata.annotations: Too long: may not be more than 262144 bytes (retried 5 times)`.

The candidate mechanisms in this class have **opposite** fixes, so this had to be measured rather than assumed. The unmanaged probe settles it: the live `keycloak-theme-openbank` and the manifest in git agree on all three keys, byte for byte. The failed operation was at revision `ba578499`; the app now reports Synced at `a13b85d3`, having been compared and found converged at a *newer* revision. **That is convergence, not masking** — and shipping a rule that fires on it would have made this detector the estate's resting state from day one, which is the single most reliable way to ensure nobody reads it.

So `SYNCED_MASKING_FAILED_OP` only fires when the failed operation's revision **equals** the revision now being claimed Synced — i.e. nothing ever successfully applied what the app says it applied. That narrowing is a measurement, and the real keycloak state is a permanent negative-control test case.

## What is shipped

`argocd-sync-verifier` — a six-hourly in-cluster CronJob (`argocd` namespace, `get`/`list` on `argoproj.io/applications` only; no `patch`, no `update`, so a future edit cannot turn a detector into an actor). It captures Application status, runs the **committed** `.github/scripts/check-argocd-sync-integrity.py`, publishes a report ConfigMap and **then fails the Job**, so `KubeJobFailed` carries it with the detail already readable.

Three rejected states:

| finding | what it catches |
|---|---|
| `ERROR_CONDITION` | any `.status.conditions[].type` ending in `Error` — the #6371 mechanism itself, the class no metric exposes |
| `SYNCED_MASKING_FAILED_OP` | `Synced` at the very revision whose only sync operation `Failed` |
| `UNVERIFIABLE_REVISION` | a repo-sourced app whose `sync.revision` is not an ancestor of `origin/main` — **ancestry, never a tag or a version string** (#6234, where an older commit's PR closed a newer one and every version-equality check stayed green) |

It deliberately does **not** fire on a stale `finishedAt`: 15 of 93 apps last synced over a month ago and every one was correct, tracking a pinned upstream chart.

## Why not the kube-state-metrics `customResourceState` the issue proposed

The issue's own steps 1–2 are the blocker, and they hold. A malformed CRS config takes down the KSM pod and with it **every `kube_*` series in the estate** — a far larger blast radius than the gap being closed — and KSM parses `--custom-resource-state-config-file` only *after* connecting to an apiserver, so a deliberately broken config and a good one produce byte-identical startup logs offline. A probe that cannot tell those apart is not validation, and validating it needs a cluster mutation this PR will not make.

This detector shares **no failure domain** with the metrics pipeline and is falsifiable offline. Exposing the condition as a metric is still the better long-run answer; it needs a *dedicated* KSM instance and is not closed by this.

## Prevent-proof — exit codes read from `$?`, output redirected to files

The negative case was written first. Every command below redirects to a file; none is piped, so no exit code is `tail`'s.

**End-to-end, against the real 93-app cluster capture:**

```
$ ... --capture <live 93-app capture> --repo . --base-ref origin/main > /tmp/live2.out 2>&1
LIVE_EXIT=0     # OK: no Application is reporting a status it cannot back up.

# the SAME capture with the real 2026-08-21 loki ComparisonError message re-injected,
# loki still reporting Synced/Healthy exactly as it did for three weeks:
$ ... --capture <same capture + the real condition> --repo . > /tmp/broken2.out 2>&1
BROKEN_EXIT=1
  - ERROR_CONDITION  loki: condition ComparisonError while sync=Synced health=Healthy.
```

**A failed capture is never "no findings"** — the defining failure of a control in this repo:

```
ABSENT_EXIT=1        capture does not exist — the detector observed nothing.
EMPTY_EXIT=1         capture holds zero Applications — the detector observed nothing.
UNPARSEABLE_EXIT=1   capture is unreadable — the detector observed nothing.
```

**The falsification suite detects the control being removed.** Replacing the one line that detects `*Error` conditions with `if False:`:

```
SABOTAGED_UNITTEST_EXIT=1                      (FAILED, failures=6 of 19)
SABOTAGED_DETECTOR_ON_REAL_DEFECT_EXIT=0       # the sabotaged detector calls the real defect clean
RESTORED_EXIT=0
```

That middle line is the point: with the tests removed, a broken detector reports the three-week outage as healthy and nothing else in CI disagrees.

**A detector that runs nowhere reports nothing, which reads exactly like one reporting no findings.** `--check-wiring` (gate `argocd-sync-integrity-unit-test`) asserts the CronJob manifest exists, is registered as an Application, grants the read verb, invokes *this* script rather than an inline copy, and contains the `exit 1`. Proven falsifiable: pointed at an absent manifest it goes red; on the real repo, `WIRING_EXIT=0`.

## Verified

- `run-gates.py --only argocd-sync-integrity-unit-test` → **PASS** (19 unit tests + 14 self-test cases + wiring)
- `run-gates.py --group lint` → exit **0**; `--group gitops` → the new gate PASS
- meta-gates `gate-subject-floor`, `gate-selftest-declaration`, `gate-lifecycle-metadata`, `gate-invocation-reachability` → all exit **0**
- `ruff --select E9,F,B` → exit 0 · `yamllint` (CI config) → exit 0 · `kubeconform -strict` → 6 valid / 0 invalid (Application CRD skipped, no schema) · `trivy config` → 0 misconfigurations
- `gen-network-policies-drift-gate` failed only until the new manifests were `git add`ed — the documented untracked-file trap in `openbank-infra/CLAUDE.md`, not drift. It passes staged.

## NOT verified

- **The CronJob has never run.** No cluster mutation was made, per the task's read-only constraint. The comparison logic is proven against the real capture; the pod, its RBAC and the clone step are proven only by schema validation and by their identity with the `keycloak-realm-drift` job they are modelled on.
- The KSM `customResourceState` config from the issue is untested and not included.
- No claim that the estate is free of this defect class *beyond* what `.status` reports — an app whose controller never recorded a condition is invisible to this and to ArgoCD alike.

### DIFF
--- a/.github/gates/gates.yaml
+++ b/.github/gates/gates.yaml
@@ -4401,3 +4401,31 @@ gates:
     min_subjects: 1600   # 1628 raw hex colour literals today (2026-08-31)
     run: |
       python3 .github/scripts/check-raw-colour-literal-ratchet.py
+
+  # ArgoCD applied NOTHING to the `loki` app for three weeks and reported `Synced / Healthy`
+  # throughout (#6371). With `ServerSideApply=true` the controller diffs by a server-side dry-run
+  # apply; a rejected dry-run makes the COMPARISON error, and the Application then keeps serving
+  # its last known status instead of going OutOfSync. The state is recorded as
+  # `.status.conditions[].type == ComparisonError` and NO ArgoCD metric exposes it —
+  # `argocd_app_info` carries `sync_status` and `health_status` only — so both rules in
+  # prometheus-rules-argocd.yaml were quiet BY CONSTRUCTION, keying on Degraded and Unknown
+  # against an app that was Healthy.
+  #
+  # The live observation cannot happen here: no PR runner holds cluster credentials. It happens
+  # in the `argocd-sync-verifier` CronJob, six-hourly, publishing a report ConfigMap and then
+  # FAILING the Job so KubeJobFailed carries it. What this gate holds is the other half — that
+  # the comparison logic still classifies correctly, and that the in-cluster detector is still
+  # wired to THAT script rather than deleted, unregistered, or replaced by an inline copy that
+  # can drift. A detector that runs nowhere reports nothing, which reads exactly like a detector
+  # reporting no findings.
+  - id: argocd-sync-integrity-unit-test
+    name: "the ArgoCD Synced-but-applying-nothing detector still classifies, and is still wired (#6371, enforced)"
+    group: gitops
+    mode: enforced
+    rationale: "ArgoCD reported Synced/Healthy on an app it had applied nothing to for three weeks, and no metric exposes the condition that recorded it; the replacement detector runs in-cluster, so only this gate can notice it being unwired or its comparison being broken."
+    review_after: "2027-02-23"
+    selftest: python3 .github/scripts/check-argocd-sync-integrity.py --self-test
+    selftest_expect: pass
+    run: |
+      python3 -m unittest openbank-infra/scripts/check_argocd_sync_integrity_test.py
+      python3 .github/scripts/check-argocd-sync-integrity.py --check-wiring
--- a/.github/scripts/check-argocd-sync-integrity.py
+++ b/.github/scripts/check-argocd-sync-integrity.py
@@ -1,0 +1,436 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) OpenBank contributors. Licensed under the Apache License, Version 2.0.
+"""Detect ArgoCD Applications that report Synced/Healthy while applying nothing (issue #6371).
+
+THE GAP
+    With `ServerSideApply=true` the application-controller computes its diff by a server-side
+    dry-run apply. If that dry-run is REJECTED the comparison errors, and the Application keeps
+    serving its LAST KNOWN status instead of going OutOfSync. It reports `Synced / Healthy` and
+    applies nothing, for as long as the rejection lasts — measured at three weeks on the `loki`
+    app (2026-08-21: `singleBinary.persistence.size` 10Gi -> 30Gi, and a StatefulSet's
+    `volumeClaimTemplates` is immutable). Every values change merged in that window was silently
+    dropped, including #6032's Loki ruler fix, which is why that fix appeared to have landed
+    while the ruler had still never loaded a rule.
+
+    The state IS recorded, in `.status.conditions[].type == ComparisonError`. It is exposed by
+    NO ArgoCD metric: `argocd_app_info` carries `sync_status` and `health_status` only. So both
+    rules in `prometheus-rules-argocd.yaml` are quiet BY CONSTRUCTION, not by accident —
+    `ArgoCDAppDegraded` keys on health `Degraded` and `ArgoCDAppHealthUnknown` on `Unknown`,
+    and this app was `Healthy`. `argocd_app_sync_total` is not a substitute either: an app that
+    legitimately needs no sync looks identical to one that CANNOT sync.
+
+WHY NOT A kube-state-metrics customResourceState
+    That was the first proposal (#6371) and it is not free: a malformed CRS config takes down the
+    kube-state-metrics pod and with it every `kube_*` series in the estate — a far larger blast
+    radius than the gap being closed — and KSM parses the config only AFTER connecting to an
+    apiserver, so a deliberately broken config and a good one produce byte-identical startup logs
+    offline. A probe that cannot tell those apart is not validation. This detector reads the same
+    field with no shared failure domain, and is falsifiable offline against fixtures.
+
+WHAT IT REJECTS
+    ERROR_CONDITION            any `.status.conditions[].type` ending in `Error` — the class the
+                               metrics cannot see. This is the #6371 mechanism itself.
+    SYNCED_MASKING_FAILED_OP   `.status.sync.status == Synced` while the last sync operation
+                               finished `Failed`/`Error` AT THE VERY REVISION now being reported
+                               as Synced. That is the "a failed sync self-heals to Synced"
+                               mechanism: the app claims Synced at a revision nothing ever
+                               successfully applied.
+                               The revision equality is load-bearing, not a refinement. Measured
+                               against the live cluster 2026-08-23, `keycloak` had a Failed
+                               operation from 2026-08-07 (its theme ConfigMap exceeded the 256KB
+                               `last-applied-configuration` annotation limit) while reporting
+                               Synced at today's revision — and it is genuinely converged: an
+                               unmanaged probe (`kubectl get cm` vs the manifest in git) found the
+                               three theme keys byte-identical. Without the revision test this
+                               detector is red on that app from the day it lands, which is the
+                               one thing that reliably stops anyone reading it.
+    UNVERIFIABLE_REVISION      (--repo) an app sourced from this repo whose `.status.sync.revision`
+                               is not a commit reachable from `origin/main`. ANCESTRY, not an
+                               image tag: a tag can be moved or reused, ancestry cannot (#6234,
+                               where an older commit's PR closed a newer one and every
+                               version-equality check stayed green).
+
+WHAT IT DELIBERATELY DOES NOT DO
+    It does not treat an old `.status.operationState.finishedAt` as a finding on its own. Most of
+    this estate's Applications track a pinned upstream chart version and legitimately have not
+    synced since May; measured 2026-08-23, 15 of 93 last synced more than a month ago and every
+    one was correct. A detector that fires on all of them is one nobody reads by week two.
+
+USAGE
+    python3 .github/scripts/check-argocd-sync-integrity.py --capture apps.json [--repo .]
+        where apps.json is `kubectl -n argocd get applications -o json`.
+    python3 .github/scripts/check-argocd-sync-integrity.py --self-test
+    python3 .github/scripts/check-argocd-sync-integrity.py --check-wiring
+
+Exit 0 = no findings. Exit 1 = findings (or a broken/absent capture). Exit 2 = usage error.
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import pathlib
+import subprocess
+import sys
+
+REPO = pathlib.Path(__file__).resolve().parents[2]
+
+# Substring that identifies an Application sourced from THIS repo. Both the SSH and HTTPS
+# spellings appear in gitops/apps, so match on the repo name rather than a full URL.
+OWN_REPO_MARKER = "open-bank-oss"
+
+FAILED_OPERATION_PHASES = {"Failed", "Error"}
+
+# The wiring the CI half asserts still exists. A detector that runs nowhere reports nothing,
+# which reads exactly like a detector reporting no findings.
+CRONJOB_MANIFEST = (
+    REPO / "openbank-infra/gitops/components/argocd-sync-verifier/argocd-sync-verifier.yaml"
+)
+APP_MANIFEST = REPO / "openbank-infra/gitops/apps/argocd-sync-verifier.yaml"
+SCRIPT_REL = ".github/scripts/check-argocd-sync-integrity.py"
+
+
+class Finding:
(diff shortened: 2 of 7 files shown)


JiRaska:
**Triage (measured, no action taken)** — same `gates.yaml` drift as #5883; the two PRs touch the
same registry.

```
git show origin/main:.github/gates/gates.yaml | grep -c '^\s*-\s*id:'                                        # 185
git show refs/remotes/origin/fix/argocd-apply-verification-6371:.github/gates/gates.yaml | grep -c '^\s*-\s*id:'   # 181
```

This branch adds one id, `argocd-sync-integrity-unit-test`. #5883 adds `ruleset-bypass-actors`. They
do not collide with each other or with the five ids `main` has gained since the merge-base, so both
should merge clean — **but check the entry count after each merge (186, then 187).** A dropped
neighbouring list entry produces no conflict and no output.

`.github/scripts/check-gate-subject-floor.py` is byte-identical between `main` and this branch (301
lines, same key set), so that file carries no drift.

`check-argocd-sync-integrity.py` and `gitops/apps/argocd-sync-verifier.yaml` are both absent from
`main` — not duplicated by a parallel PR.

Checks are green but completed `2026-08-23T04:42:10Z` at `behind_by=795`; that green is about the
old base.

JiRaska:
## Falsification, not review — two states this detector calls healthy

Ran the committed `check-argocd-sync-integrity.py` at `d48ccbbd1` against four fixtures in an isolated worktree. Every exit code below was read from `$?` immediately after redirecting to a file; nothing is piped.

The detector does what it says for the mechanism in #6371. The positive control passes:

```
p_error  (loki: Synced/Healthy + ComparisonError)   EXIT=1   ERROR_CONDITION
```

Two other states in the same family return `0`.

### 1. A stale-but-ancestor revision is indistinguishable from a current one

`UNVERIFIABLE_REVISION` tests *ancestry*, and ancestry is true for **every** commit on `main`, including one from months ago. So "Argo read a manifest that is not the manifest in git" — the state where Argo genuinely applied something, just not the current thing — is invisible.

Fixture: one repo-sourced app, `sync=Synced`, `health=Healthy`, `sync.revision` = a commit **399 commits behind `origin/main`**. Against a current-revision control:

```
a_current  (revision == origin/main tip)            EXIT=0
c_stale    (revision 399 commits behind main)       EXIT=0
diff a_current.out c_stale.out                      EXIT=0   <- byte-identical
```

The two outputs are the same bytes. The check that would separate them is already free at the same call site: `git rev-list --count <revision>..origin/main`, i.e. distance, not membership. Ancestry answers "is this a real commit on our history" (a good check against #6234); it cannot answer "is this the current one".

Note this compounds with the exemption: apps whose `repoURL` lacks `open-bank-oss` skip `git_check` entirely, so for the ~22 upstream-chart apps — `loki` among them, the app the issue is about — no revision test runs at all.

### 2. State (d): an Application that does not exist is not a finding

The capture is the census, and the only census guard is `len(apps) == 0`. `openbank-infra/gitops/apps/` declares **92** Applications (`grep -l 'kind: Application'`, 94 across the whole gitops tree). A capture holding one healthy app and missing the other 91 is reported as clean:

```
d_missing  (1 unrelated healthy app; 91 declared Applications absent)   EXIT=0
   "1 Application(s) examined ... OK: no Application is reporting a status it cannot back up."
```

An Application that was never created, was deleted, or that the ServiceAccount cannot list produces no `.status`, therefore no condition, therefore no finding — the same shape as the 21 Rollouts with zero series and the CNPG cluster nothing scrapes. A threshold over a missing object never fires, and the app most likely to be misconfigured is the one guaranteed not to be examined.

I grepped for an existing declared-vs-live census and found none: no script under `.github/scripts/` compares `gitops/apps/*.yaml` against a live Application list (`check-opa-bundle-apply-size.py` reads that directory, but only for `ServerSideApply` on OPA bundles).

The fix is cheap and stays inside this script: derive the expected set from `gitops/apps/*.yaml` in the clone the CronJob already makes, and report `MISSING_APPLICATION` for each declared name absent from the capture. That also turns the `len(apps)==0` guard into a real floor rather than a boundary case.

### Against the four states

| state | separated? | evidence |
|---|---|---|
| (a) applied, matched | yes | `a_current` EXIT=0 |
| (b) never ran | **no, by design** | `finishedAt` staleness deliberately excluded, with a stated and defensible reason (15/93 legitimately stale). Honest, but it means (b) is still uncovered — `ERROR_CONDITION` catches (b) only when a condition happened to be recorded |
| (c) stale repo revision | **no** | `a_current` and `c_stale` produce byte-identical output, both EXIT=0 |
| (d) Application absent | **no** | `d_missing` EXIT=0 with 91 of 92 declared apps missing |

None of this contradicts the PR's own "NOT verified" section, which is accurate as far as it goes — but the PR closes #6371, and (c) and (d) are the states most likely to be read as covered once it lands. Worth either fixing in this PR (both are small, both inside the existing script) or naming explicitly as out of scope so the next reader does not assume the census exists.

I did not verify anything against the live cluster (read-only constraint; no capture taken) and did not check whether the CronJob's RBAC can in fact list Applications across all namespaces — which, if it cannot, is state (d) at scale.

JiRaska:
Independent corroboration of the (d) finding above, reached by a different route, with the counts re-derived on `origin/main`:

```
$ git grep -h "^kind: Application" origin/main -- 'openbank-infra/gitops/**' | wc -l
95
$ git ls-tree -r --name-only origin/main -- openbank-infra/gitops/apps/ | grep -c '\.yaml$'
93
```

So the declared estate is **95 Applications** (93 files under `gitops/apps/`), slightly more than the 92/94 quoted above — the point is unchanged and the direction makes it marginally worse.

Checking the script itself for an expected-set derivation, on this branch rather than on `main`:

```
$ git show origin/<branch>:.github/scripts/check-argocd-sync-integrity.py | grep -n -B2 -A2 'gitops/apps'
 80: # Both the SSH and HTTPS spellings appear in gitops/apps, so match on the repo name…
 90: APP_MANIFEST = REPO / "openbank-infra/gitops/apps/argocd-sync-verifier.yaml"
```

Both hits are incidental — a comment about URL spellings, and the path to the verifier's **own** Application for its wiring check. Neither enumerates the declared set, which confirms the finding from the other direction: there is nothing for a capture to be compared *against*.

**A probe note, because it nearly misled me.** My first attempt ran that grep against `origin/main` and returned `0`, which reads as "the script has no such logic". The script is not on `main` at all — it arrives with this PR. Zero was a fact about where I looked, not about the code. Anyone re-checking this should read the branch, and keep a known-positive in the same run (`UNVERIFIABLE_REVISION` returns 3) so an empty result cannot be mistaken for an answer.

**Why (d) is the one worth fixing before merge, in my reading.** (c) is a real gap but degrades gracefully — a stale revision still has *an* Application reporting *something*, so other signals remain. (d) removes the subject entirely: the capture is the census, so an Application that vanishes from the cluster cannot be distinguished from one that was never expected. That is the same shape as a CNPG cluster nothing scrapes being invisible to its own backup alert — the object most likely to be misconfigured is the one guaranteed not to alert. The fix is small and lives inside this script: derive the expected names from `gitops/apps/*.yaml` in the clone the CronJob already makes.

Nothing in this comment came from the live cluster; it is all repo-side, so whether (c) or (d) is *currently* live in the estate remains unestablished.

PR #6619, opened 2026-08-23
fix(admin-ui): clarify dashboard refresh

JiRaska:
## Summary
- add explicit button semantics and localized busy/name state to Platform Overview refresh
- preserve parallel governance/health reads, graceful fallback and RBAC

## Verification
- git diff --check

## Linked issues


JiRaska (commented):
Independent review: NO BLOCKER. The diff only clarifies the Dashboard refresh control with type=button, localized aria-label, aria-busy and decorative icon semantics. Parallel governance/health reads, graceful fallback and RBAC remain unchanged. git diff --check passes; focused Vitest is unavailable because dependencies are not installed; repository CI remains authoritative.

PR #6621, opened 2026-08-23
build(governance): measure each service's coverage over its own sources, not libs' (#6384)

### LINKED ISSUE #6384: A libs-runtime addition lowers every dependent service's coverage — the red check names a service the PR never touched
## What

Each service's Kover report measures `com/openbank/libs` alongside its own package, but a service's
test task never runs `libs-runtime`'s tests. So **every addition to a shared library lowers the
measured line coverage of every service that depends on it**, no matter how well that code is tested
in its own module.

Measured on `openbank-fx-service` (Kover binary report of a passing run on `origin/main`):

```
$ strings openbank-fx-service/build/tmp/koverCachedVerify/agg-ic.ic \
    | grep -oE 'com/openbank/[a-z]+' | sort -u
com/openbank/fx
com/openbank/libs
```

## How it surfaced

PR #5719 adds two files under `openbank-libs-runtime/.../messaging/EventRetry.kt` and touches
**zero** fx-service files. Its CI:

```
build (openbank-fx-service):  Rule violated: lines covered percentage is 60.504200,
                              but expected minimum is 65
clean origin/main:            ./gradlew :openbank-fx-service:koverVerify  ->  rc=0
```

fx broke first because its floor (65, ratcheted up from 40) sits closest to its actual figure. It is
not special otherwise — the next libs addition hits whichever service is then nearest its own floor.

## Why this is worth fixing rather than absorbing

The failure is **misattributed by construction**. The red check names a service the PR never touched,
on a branch whose author has no reason to look at fx. That is expensive to diagnose every time, and
it is the shape that makes people lower a floor to get green — which is the one move that would make
the ratchet meaningless.

It also interacts badly with path-scoped CI: fx is only rebuilt when something forces it, so a
coverage regression introduced by a libs change can sit unnoticed until an unrelated PR happens to
rebuild that service — the same latency documented for cross-service domain breaks.

## Options

1. **Scope each service's Kover report to its own package** (`filters { includes { classes("com.openbank.<svc>.*") } }`),
   and give `openbank-libs-runtime` / `openbank-libs-domain` their own floors measured against their
   own tests. Makes each number mean what its name says.
2. Keep the aggregate view but **run libs tests into the dependent's report**, so shared code is
   measured with the coverage it actually has.
3. Accept the coupling and document that every service floor is a function of libs size — the
   weakest option, because the floors' own comments cite measurements that a libs change invalidates.

Whichever is chosen, the check is a libs-only PR that adds an uncovered line and confirms no
dependent service's `koverVerify` moves.

JiRaska:
## What

Scope every service's Kover report to its **own sources**, so a coverage floor stops being a
function of shared-library size.

`build-logic/src/main/kotlin/openbank.quarkus-service.gradle.kts` now excludes
`com.openbank.libs.*` from the report/verify filters fleet-wide. The libraries are not left
unmeasured — `openbank-libs-domain` (floor 30) and `openbank-libs-runtime` (floor 50) each carry
their own `koverVerify` rule against their own tests, which is the only place a library's coverage
can honestly be judged.

Closes #6384

## The measurement, including where it disagrees with itself

**CI (authoritative — this is the environment the issue was filed from).** On the #5719 branch,
job `build (openbank-fx-service)`:

| state | fx `koverVerify` |
|---|---|
| libs **included** (before) | `60.504200%` |
| libs **excluded** (this change) | `30.905700%` |

`30.905700%` is fx's own line coverage. `60.504200%` was an fx+libs aggregate — fx's tests exercise
a lot of libs code well, so libs was holding the number **up**. That is why #5719, which added 13
uncovered lines to `libs-runtime`'s `EventRetry` and touched **zero** fx files, moved fx from over
65 to 60.5 and reddened a service its author never looked at — and why fx's floor was then lowered
65 → 60, the exact move that makes a ratchet meaningless.

**Locally, on this machine, the dilution does not reproduce, and that is worth stating.** On clean
`origin/main`, with `--rerun-tasks`, fx's report contains only `com/openbank/fx` packages and
`koverVerify` reads `78.618100%` — libs is *absent* from the denominator. Adding 20 uncovered lines
to `libs-runtime` (`DilutionProbe`, compiled and present in the libs-runtime jar) left it
**byte-identical**: `842/1071 = 78.618100%`. The raw `agg-ic.ic` does contain `com/openbank/libs`
strings, as the issue reports, but the XML report and `koverVerify` filter them out here.

So local and CI disagree about the scope of fx's report *from the same committed configuration*,
and CI's own numbers (60.5 → 30.9 on adding the exclusion) prove libs is in scope there. I could
not explain the divergence, so I did not use the local run to argue the premise away. **CI is the
authority; the local run is reported as a discrepancy, not as a refutation.** The prevent-proof
below is therefore run in CI.

## Prevent-proofs

**Direction 1 — a libs addition must not move a dependent.** The first push of this branch carried
a temporary `DilutionProbe`: 20 uncovered lines in `openbank-libs-runtime`, referenced by nothing.
Fleet run [32619370944](https://github.com/JiRaska/open-bank-oss/actions/runs/32619370944) built
**every service** (a `build-logic` change is fleet-scoped) with that probe present and **every
floor except fx's untouched**:

- **~55 dependent service builds: green.** Not one was reddened by uncovered library code.
- The **only** build the probe reddened was `openbank-libs-runtime`'s own — and on `detekt`, over
  the throwaway file's style, not on `koverVerify`. Uncovered library code now fails in the
  library, which is where it can be fixed.

Compare with the same shape on `origin/main`: #5719's 13 uncovered `EventRetry` lines reddened
`build (openbank-fx-service)`, a service that PR never touched.

**What this proof does not include, stated plainly:** `koverVerify` prints no figure when it
*passes*, so I do not have per-service before/after percentages to show byte-equality across the
two pushes. The evidence is the pass/fail flip, plus CI's own demonstration on #5719 that the
exclusion has real effect there (60.504200% → 30.905700%). A stronger equality proof would need a
throwaway push that sets every floor to 100 purely to make the numbers print, at the cost of two
more full-fleet runs.

**Direction 2 — a real regression in a service's own sources must still redden.** Measured locally
with the fix applied, by deleting fx's `application/` and `infrastructure/` test trees:

```
$ ./gradlew --rerun-tasks :openbank-fx-service:koverXmlReport :openbank-fx-service:koverVerify > D2.log 2>&1; echo "RC=$?" >> D2.log
> Rule violated: lines covered percentage is 40.056000, but expected minimum is 60
BUILD FAILED
RC=1
```

against the same command on the unmodified tree, `RC=0`. Exit codes read from `$?` after
redirecting to a file, never off a pipe. Test totals for the baseline run: 159 tests, 0
failures/errors, **1 skipped** (read from JUnit XML, not the console).

## The floor change, named for what it is

`openbank-fx-service`'s floor moves **60 → 30**, and it is the **only** floor this PR changes: the
fleet run above shows every other service already clears its floor on its own sources alone. That
is a re-baseline of a *different quantity*, not a ratchet relaxation — 60 was never fx's coverage.

One honest caveat on the number 30. CI measured fx's own-source coverage at `30.905700%`, so 30
leaves under a point of headroom; locally the same command measures `78.618100%` (see the
divergence above). 30 is chosen because it is safe under *both* measurements, and it should be
ratcheted up as soon as the divergence is understood — a floor sitting 0.9 points under the
measurement it guards is not a comfortable place to leave it.

## Trade-off

The alternative was ratcheting on **absolute uncovered lines** in a module's own sources. A ratio
and a count fail in opposite directions as a codebase grows: a ratio lets a module add uncovered
code as long as it adds covered code alongside, and an absolute count blocks a large well-tested
feature while permitting an ever-worsening ratio in a shrinking module. The ratio is kept because
it is what every floor in the tree already records and what ADR-0020 specifies; this change removes
the one input to that ratio the module does not control.

Option 2 from the issue — running libs' tests into each dependent's report — was rejected: it
raises the aggregate and would need no floor edits, but an *uncovered* libs addition still dilutes
every dependent, so it fails Direction 1 by construction.

## Not verified

- The local/CI scope divergence above is unexplained.
- Only `openbank-fx-service` was measured locally end to end; the other ~55 dependents are measured
  by this PR's CI run.
- No live system, cluster, or repository setting was touched.


JiRaska:
Picking this up after the authoring session ended mid-run. Both fleet runs completed; recording
their results here so the evidence is on the PR rather than only in a transcript.

**Direction 1 — a library addition must no longer redden dependents.** Run 1 carried a temporary
`DilutionProbe` (30 uncovered lines) in `openbank-libs-runtime` on top of the fix. Across ~100
fleet jobs the **only** failure was `openbank-libs-runtime`'s own build — the module the uncovered
code is actually in. **Zero dependent services failed.**

**Direction 2 — a genuine regression must still redden.** Already measured locally before the PR
opened: deleting fx-service's `application/` and `infrastructure/` test trees gives
`Rule violated: lines covered percentage is 40.056000, but expected minimum is 60`, `rc=1`;
unmodified tree `rc=0`.

**Run 2** (probe reverted, `7883a43dc`) is clean: 0 failures, 0 pending.

Together those are the two halves that matter. Direction 1 alone would be satisfiable by simply
disabling the gate; direction 2 is what shows it still bites.

One thing a reviewer should weigh deliberately, because it is the substantive decision in this PR
rather than a mechanical consequence: `openbank-fx-service`'s floor moves 60 → 30. The inline
comment argues that is a re-baseline and not a relaxation — the 65 and 60 were both compared
against an fx+libs aggregate (60.504200%), while fx's own line coverage is 30.905700%. That
reading is consistent with everything measured here, and it means fx's true own-source coverage is
roughly half what its floor implied. Worth an explicit view from someone who owns that service.

Not verified by me: I did not re-run either fleet build; these are the runs the authoring session
launched, read from their conclusions.

JiRaska:
### The CodeQL failure here is a killed runner, not a finding

`CodeQL (java-kotlin, manual)` ends with:

```
##[error]The runner has received a shutdown signal. This can happen when the runner service is
stopped, or a manually started runner is canceled.
##[error]Process completed with exit code 143.
```

Exit 143 is SIGTERM — the job was terminated mid-run, roughly 15 minutes in. Nothing was analysed and
no alert was raised; the build step before it had not reported a compilation problem. This is the
spot-kill signature, and the repo has an `Auto-retry spot-killed CI runs` workflow for exactly this
shape, which did not pick this one up.

So this PR is `UNSTABLE` on an infrastructure event from 2026-08-23, not on anything in its diff. It
needs a re-run of that one job rather than a code change — and if the auto-retry workflow is expected
to cover exit-143 kills, this run is a case it missed and worth checking against its matcher.

Not re-run from here: `CodeQL` is not one of the five required contexts, so this does not block the
merge, and a re-run replays the original event payload against the same merge commit — worth doing
deliberately rather than as a reflex.

PR #6636, opened 2026-08-23
fix(admin-ui): mark KYC table busy

JiRaska:
## Summary
- expose existing KYC cases table refresh/search state via aria-busy
- preserve KYC query, 404/405 fallback, BFF and RBAC behavior

## Verification
- git diff --check

## Linked issues

### DIFF
--- a/openbank-admin-ui/src/app/kyc/page.tsx
+++ b/openbank-admin-ui/src/app/kyc/page.tsx
@@ -115,7 +115,7 @@ export default function KycPage() {
         </div>
       )}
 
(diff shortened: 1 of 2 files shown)


JiRaska (commented):
Independent review: NO BLOCKER. The diff only marks the existing KYC cases table busy while refresh/search runs and adds a guard. Query/fallback behavior, BFF and compliance RBAC remain unchanged. git diff --check passes; focused Vitest is unavailable because dependencies are not installed; repository CI remains authoritative.

PR #6650, opened 2026-08-23
fix(admin-ui): clarify sanctions disclosure

JiRaska:
## Summary
- make sanctions list disclosure an explicit accessible button
- preserve screening/list/four-eyes flows and RBAC

## Verification
- git diff --check

## Linked issues


JiRaska (commented):
Independent review: NO BLOCKER. The diff only clarifies the existing sanctions list expand/collapse control with type=button, localized aria-label, aria-expanded, and decorative icon semantics. Screening/list fetches, four-eyes decision flow, BFF and RBAC remain unchanged. git diff --check passes; focused Vitest is unavailable because dependencies are not installed; repository CI remains authoritative.

PR #6654, opened 2026-08-23
fix(admin-ui): clarify account controls

JiRaska:
## Summary
- make account search/reset/load-more controls explicit non-submit buttons
- expose search busy state and localized accessible names

## Verification
- git diff --check

## Linked issues


JiRaska (commented):
Independent review: NO BLOCKER. This bounded diff only clarifies existing account search/reset/load-more controls with type=button, localized names, and search aria-busy. Account query classification, PartySearch, pagination, accountApi/BFF behavior and RBAC are unchanged. git diff --check passes; focused Vitest is unavailable because dependencies are not installed; repository CI remains authoritative.

PR #6673, opened 2026-08-24
fix(admin-ui): clarify catalog status control

JiRaska:
## Summary
- expose Product Catalog status toggle state to assistive technology
- hide decorative lifecycle/close icons

## Verification
- git diff --check

## Linked issues


JiRaska (commented):
Independent review: NO BLOCKER. This bounded diff only adds aria-pressed to the existing Product Catalog status lifecycle toggle and marks its status/close icons decorative. Catalog BFF endpoints, optimistic locking, lifecycle handlers, permissions and maker-checker behavior remain unchanged. git diff --check passes; focused Vitest is unavailable because dependencies are not installed; repository CI remains authoritative.

PR #6674, opened 2026-08-24
fix(admin-ui): expose catalog row state

JiRaska:
## Summary
- expose Product Catalog row status toggle state with aria-pressed
- preserve lifecycle handlers and RBAC

## Verification
- git diff --check

## Linked issues


JiRaska (commented):
Independent review: NO BLOCKER. This bounded diff only adds aria-pressed to the existing Product Catalog row lifecycle toggle. Catalog BFF, optimistic locking, lifecycle actions, permissions and maker-checker behavior are unchanged. git diff --check passes; focused Vitest is unavailable because dependencies are not installed; repository CI remains authoritative.

PR #6675, opened 2026-08-24
fix(admin-ui): clarify transaction filters

JiRaska:
## Summary
- make transaction filter disclosure/clear controls explicit buttons
- add localized clear-filter naming

## Verification
- git diff --check

## Linked issues


JiRaska (commented):
Independent review: NO BLOCKER. This bounded diff only clarifies existing Transactions filter disclosure and clear controls with type=button and localized clear naming. Transaction query classification, pagination, read-only BFF flow, money-path data and RBAC are unchanged. git diff --check passes; focused Vitest is unavailable because dependencies are not installed; repository CI remains authoritative.

PR #6681, opened 2026-08-24
fix(admin-ui): clarify audit search state

JiRaska:
## Summary
- expose audit search loading state to assistive technology
- add explicit button type and localized accessible naming

## Verification
- git diff --check

## Linked issues


JiRaska (commented):
Independent review: NO BLOCKER. This bounded diff only clarifies the existing Audit search control with type=button, localized accessible naming, and aria-busy while loading. Audit query validation, read-only BFF fetch, data rendering and system:view RBAC remain unchanged. git diff --check passes; focused Vitest is unavailable because dependencies are not installed; repository CI remains authoritative.

PR #6685, opened 2026-08-24
fix(admin-ui): clarify delegations search

JiRaska:
## Summary
- expose Delegations party search loading state and accessible naming
- mark selected party result and explicit result buttons

## Verification
- git diff --check

## Linked issues


JiRaska (commented):
Independent review: NO BLOCKER. This bounded diff only clarifies Delegations party search and result selection with explicit button types, localized names, aria-busy and aria-pressed. PartySearch/BFF queries, grant projection reads, delegation data and RBAC remain unchanged. git diff --check passes; focused Vitest is unavailable because dependencies are not installed; repository CI remains authoritative.

PR #6694, opened 2026-08-24
fix(admin-ui): clarify swift raw disclosure

JiRaska:
## Summary
- relate SWIFT raw payload disclosure to its panel
- expose truthful show/hide labels and decorative icon semantics

## Verification
- git diff --check

## Linked issues


JiRaska (commented):
Independent review: NO BLOCKER. This bounded diff only clarifies the existing SWIFT raw JSON disclosure with explicit button semantics, related panel ID, truthful localized show/hide labels, and decorative icons. SWIFT fetch/detail data, raw payload rendering, RBAC and navigation remain unchanged. git diff --check passes; focused Vitest is unavailable because dependencies are not installed; repository CI remains authoritative.

PR #6695, opened 2026-08-24
fix(admin-ui): clarify onboarding filter reset

JiRaska:
## Summary
- make onboarding stage filter reset an explicit button
- hide decorative reset icon

## Verification
- git diff --check

## Linked issues


JiRaska (commented):
Independent review: NO BLOCKER. This bounded diff only clarifies the existing Onboarding stage-filter reset control with an explicit button type and decorative icon semantics. Stage filtering, onboarding fetch/query, pagination and RBAC are unchanged. git diff --check passes; focused Vitest is unavailable because dependencies are not installed; repository CI remains authoritative.

PR #6700, opened 2026-08-24
fix(admin-ui): clarify global error recovery

JiRaska:
## Summary
- make global error reload recovery an explicit accessible button
- preserve full-page recovery behavior

## Verification
- git diff --check

## Linked issues


JiRaska (commented):
Independent review: NO BLOCKER. This bounded diff only clarifies the global error reload recovery control with an explicit button type and bilingual accessible name. Full-page reload behavior and error recovery remain unchanged. git diff --check passes; focused Vitest is unavailable because dependencies are not installed; repository CI remains authoritative.

PR #6703, opened 2026-08-24
fix(admin-ui): clarify auth error retry

JiRaska:
## Summary
- make auth error retry an explicit accessible button
- preserve Keycloak sign-in recovery flow

## Verification
- git diff --check

## Linked issues


JiRaska (commented):
Independent review: NO BLOCKER. This bounded diff only clarifies the auth error retry control with an explicit button type and bilingual accessible name. The Keycloak sign-in recovery call and auth/session behavior remain unchanged. git diff --check passes; focused Vitest is unavailable because dependencies are not installed; repository CI remains authoritative.

PR #6718, opened 2026-08-24
fix(admin-ui): clarify regulatory preview close

JiRaska:
## Summary
- make regulatory export preview close an explicit button
- localize its accessible name and hide decorative icon

## Verification
- git diff --check

## Linked issues


JiRaska (commented):
Independent review: NO BLOCKER. This bounded diff only clarifies the existing regulatory export preview close control with an explicit button type, localized accessible name, and decorative icon semantics. Export preview data, template mapping, regulatory read-only flow and RBAC remain unchanged. git diff --check passes; focused Vitest is unavailable because dependencies are not installed; repository CI remains authoritative.

PR #6752, opened 2026-08-24
fix(admin-ui): bound client evidence staging

JiRaska:
## Why
The Admin UI deployment can stall in cross-repository customer-app evidence staging. Artifact downloads were best-effort but had no connection or total timeout, so a stalled redirect could block the sandbox deployment indefinitely.

## Change
Adds bounded connection and total timeouts with short retry behavior to the read-only artifact fetch. A failed download remains absent evidence; it never becomes a passing verdict and cannot block deployment forever.

## Verification
- Ruby YAML parse of the deployment workflow
- Test Intelligence ecosystem self-test, including its red path
- Full Test Intelligence ecosystem verifier: SUBJECTS=59

### DIFF
--- a/.github/workflows/admin-ui-deploy.yml
+++ b/.github/workflows/admin-ui-deploy.yml
@@ -362,7 +362,13 @@ jobs:
           APP_REPO: JiRaska/openbank-app
         run: |
           set -u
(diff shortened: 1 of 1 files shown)


JiRaska:
Closing without merge: the live job later showed customer-app evidence staging completed at 16:44; the long-running phase is the expected Build, collect, push step. The timeout change is therefore not evidence-backed by this incident.

PR #7034, opened 2026-08-26
feat(billing): expose pending approvals in unified inbox

JiRaska:
## Summary
- add a bounded, FIFO read-only billing approval queue with maker and creation provenance
- document the additive API contract and preserve the existing checker-only PATCH decision path
- federate billing into the admin unified approval inbox with explicit source health

## Verification
- targeted billing tests, ktlint, detekt, and quarkusBuild: green
- admin type-check, focused Vitest (19/19), and scoped ESLint: green
- API contract, route conformance, enum drift, and diff checks: green
- full billing suite: 111 tests; 2 failed and 20 skipped only because Docker/Testcontainers is unavailable locally
- both commits signed and verified

## Risk and controls
- money-path change: requires two independent approvals
- existing four-eyes self-approval prevention and decision endpoint remain unchanged
- existing threat model: docs/threat-models/openbank-billing-service.md

Refs #5679

PR #7031, opened 2026-08-26
feat(balance): expose pending approvals in unified inbox

JiRaska:
## Summary

Expose balance-service's pending four-eyes queue and federate it into the Admin UI approval inbox, so a parked credit/debit decision is visible to checkers before its 24-hour TTL expires. This PR is intentionally stacked on #7028 and contains only the balance source.

## Type of change

- [x] feat (new functionality)
- [x] fix (bug fix)
- [ ] refactor (no behavior change)
- [ ] perf (performance)
- [ ] docs
- [ ] chore (build / tooling)
- [ ] security (private until disclosed)
- [ ] breaking change

## Linked issues

Refs #5679

## Changes

- Add a bounded, oldest-first `GET /api/v1/balances/approvals` checker queue while preserving the existing PATCH decision path and maker != checker invariant.
- Publish the additive API contract at version 1.9.0 with provenance fields required by the inbox.
- Federate balance approvals with explicit source health, stable ordering, and regression coverage.

## Definition of Done

- [x] Hexagonal layering preserved.
- [x] Unit tests added/updated.
- [ ] Integration tests added/updated. No runtime disposal change; CI remains authoritative for Testcontainers.
- [x] Contract tests updated.
- [ ] `./gradlew detekt ktlintCheck koverVerify build` passes.
- [x] No new lint warnings.
- [x] OpenAPI spec updated.
- [x] Flyway migration added + rollback note. N/A — no DB change.
- [x] Avro schema versioned backward-compatibly. N/A — no event change.
- [x] Docs updated. N/A — implements ADR-0227 D2.

Local proof: targeted balance tests, ktlint, detekt, Quarkus build, Admin UI type-check, focused ESLint, and 18 focused Vitest tests pass. OpenAPI route/API-version/enum gates pass. The full balance suite ran 161 tests; the only 2 failures and 48 skips are Testcontainers boundaries with no local Docker socket, so CI is authoritative there.

## Security checklist

- [x] No secrets, tokens, passwords, or PII.
- [x] No unjustified suppression. The test-only unchecked cast narrows the JAX-RS response entity for assertions.
- [x] No new dependency.
- [x] Auth / crypto / payment code changed? Money-path balance approval visibility changed; two independent approvals and security review are required.
- [x] PII path changed? No.
- [x] Cardholder data path changed? No.

## Compliance impact

- [x] DORA

[...]

PR #7028, opened 2026-08-26
feat(account): expose pending approvals in unified inbox

JiRaska:
## Summary

Expose account-service's pending four-eyes queue and federate it into the Admin UI approval inbox so protected account lifecycle actions no longer become invisible until their 24-hour TTL expires. This PR is intentionally stacked on #7020; its diff contains only the account source, an OpenAPI truthfulness repair, and the exact-match enum-gate fix required by that repair.

## Type of change

- [x] feat (new functionality)
- [x] fix (bug fix)
- [ ] refactor (no behavior change)
- [ ] perf (performance)
- [ ] docs
- [ ] chore (build / tooling)
- [ ] security (private until disclosed)
- [ ] breaking change

## Linked issues

Refs #5679

## Changes

- Add a bounded, oldest-first `GET /api/v1/accounts/approvals` checker queue without changing the existing maker/checker decision path.
- Federate account approvals into the Admin UI inbox with truthful source health, provenance, ordering, and regression coverage.
- Publish the previously served account authorization and currency-pocket routes, correct the AccountStatus vocabulary, and make the enum ratchet prefer an exact match over an overlapping superset.

## Definition of Done

- [x] Hexagonal layering preserved (domain has zero framework imports).
- [x] Unit tests added/updated.
- [ ] Integration tests added/updated (if service boundary involved).
- [x] Contract tests updated (if public API changed).
- [ ] `./gradlew detekt ktlintCheck koverVerify build` passes.
- [x] No new lint warnings.
- [x] OpenAPI spec regenerated (if REST API changed).
- [x] Flyway migration added + rollback note (if DB changed). N/A — no DB change.
- [x] Avro schema versioned backward-compatibly (if event changed). N/A — no event change.
- [x] Docs updated (README, ADR if architectural). N/A — implements ADR-0227 D2.

Local proof: targeted account tests, ktlint, detekt, Quarkus build, Admin UI type-check, focused ESLint, and 17 focused Vitest tests pass. OpenAPI route/API-version/enum gates pass. The full account test run reached 257 tests; the only 3 failures and 46 skips require Docker/Testcontainers, unavailable in the clean local worktree, so CI remains authoritative for that boundary.

## Security checklist

- [x] No secrets, tokens, passwords, or PII in code, config, tests, logs.
- [x] No new `@SuppressWarnings`, `as Any`, `@Suppress("...")` without justification. The existing-style unchecked cast is test-only and narrows the JAX-RS response entity for assertions.
- [x] No new third-party dependency.
[...]

PR #7037, opened 2026-08-26
feat(consent): expose pending approvals in unified inbox

JiRaska:
## Summary
- add a bounded, FIFO read-only consent approval queue with maker and creation provenance
- document the additive API contract and update the money-path threat model
- federate the final previously not-configured source into the Admin UI approval inbox

## Verification
- targeted consent tests, ktlint, detekt, and quarkusBuild: green
- admin type-check, focused Vitest (20/20), and scoped ESLint: green
- API contract, route conformance, enum drift, threat-model diff, and diff checks: green
- both commits signed and verified

## Risk and controls
- money-path change: requires two independent approvals
- existing four-eyes self-approval prevention and decision endpoint remain unchanged
- existing M2M consent policy exceptions are not widened
- threat model updated: docs/threat-models/openbank-consent-service.md

Refs #5679

### DIFF
--- a/docs/threat-models/openbank-consent-service.md
+++ b/docs/threat-models/openbank-consent-service.md
@@ -56,6 +56,16 @@ escalating a consent is a direct path to unauthorized data access or payment ini
 
 ## 6. Change log
 
+- **2026-08-26** — The operator approval inbox gains a bounded, read-only
+  `GET /api/v1/consents/approvals` edge. It returns pending approval workflow metadata
+  (random id, action, resource id, maker id and creation time) only to `ROLE_OPERATOR` or
+  `ROLE_ADMIN` callers behind the existing OPA boundary. Results are capped at 200 and ordered
+  oldest first; the endpoint cannot grant, revoke or decide a consent. Existing maker/checker
+  separation, one-time execution and 24-hour Redis TTL controls remain unchanged. **Risk class:**
+  confidentiality of operator workflow metadata and bounded Redis read load; no new caller,
+  service edge or consent mutation. Rollback: remove the GET route and let the admin UI report this
+  source unavailable; consent decisions and lifecycle flows are unaffected.
+
 - **2026-08-24** — Synthetic-journey taint now propagates over this service's existing internal SCA REST edge through `SyntheticTaintClientFilter` (ADR-0252, #4348). This adds no caller, endpoint, network-policy edge, privilege or SCA bypass: the synthetic journey must still satisfy the same controls. It prevents the marker from being lost before a downstream persistence/event boundary; a fleet gate requires every new client to choose propagation or a reasoned external boundary.
 
 - **2026-08-05** — Trust-boundary change (#3734): `operator-consent-write`'s M2M exclusion widened from `service-account-openbank-services` to the `service-account-` prefix. ADR-0206 D5 closed the backend M2M identity but left `service-account-openbank-edge` admitted to every `consent.*` write. The edge's legitimate consent access ({consent.list, consent.revoke} via base edge-service-consent, the customer's PSD2 consent screen) is preserved; no prohibition clause needed — the role_action_matrix grants no `consent.*` write to ROLE_OPERATOR.
--- a/openbank-admin-ui/src/app/api/approvals/pending/route.ts
+++ b/openbank-admin-ui/src/app/api/approvals/pending/route.ts
@@ -24,12 +24,11 @@ type SourceState = 'ok' | 'forbidden' | 'unavailable' | 'not-configured'
 const NOT_CONFIGURED_SOURCES = {
   balance: 'not-configured',
   billing: 'not-configured',
-  consent: 'not-configured',
 } as const satisfies Record<string, SourceState>
 
 type InboxItem = {
   id: string
-  domain: 'lending' | 'sanctions' | 'transaction' | 'domestic-payment' | 'clearing' | 'fx' | 'ledger' | 'swift' | 'sepa-payment' | 'sepa-instant' | 'notification' | 'party' | 'account' | 'agent'
+  domain: 'lending' | 'sanctions' | 'transaction' | 'domestic-payment' | 'clearing' | 'fx' | 'ledger' | 'swift' | 'sepa-payment' | 'sepa-instant' | 'notification' | 'party' | 'account' | 'consent' | 'agent'
   action: string
   resourceId: string | null
   maker: string | null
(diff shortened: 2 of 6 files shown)


JiRaska (commented):
NO BLOCKER: reviewed the committed stacked diff. The new GET is bounded, oldest-first and read-only; it exposes maker/time provenance, preserves the existing checker-only PATCH, self-approval and one-time execution controls, does not widen M2M consent policy exceptions, and removes the last explicit not-configured source from the unified inbox. Local backend/build/lint, 20 focused UI tests, API gates and threat-model diff are green. Money-path merge still requires two independent approvals and green CI.

JiRaska (commented):
Updated-head review: NO BLOCKER. The Approval Centre now names the full bank workflow, treats a non-200 federated queue as an explicit incomplete/error state instead of silently rendering empty, and retains per-source degraded-state truthfulness. Type-check, scoped ESLint (one pre-existing effect warning, zero errors), and 26 focused tests pass. Money-path merge remains gated on two independent approvals and green CI.

JiRaska (commented):
NO BLOCKER — updated-head review. Rebase preserves the consent queue, truthful Approval Centre error/IA behavior, strict final source guard, and maker-checker/RBAC boundaries. Type-check and 26 focused tests pass.

JiRaska:
## Measured defect: `consent` is wired **and** still declared `not-configured` — and it ships green

Triage only, no changes pushed. This is the one finding in the four-PR family
(`#7028`/`#7031`/`#7034`/`#7037`) that is not just a merge-order question.

### What the branch does

On `origin/codex/consent-approval-inbox`, `route.ts` fully wires consent — `consentPending()` at
:314, the destructure at :356, `consentPending(headers)` at :369, the items spread at :372, and
`consent: consent.state` at :390 — while **line 28 still reads**:

```ts
const NOT_CONFIGURED_SOURCES = {
  account: 'not-configured',
  balance: 'not-configured',
  billing: 'not-configured',
  consent: 'not-configured',      // <- not removed
} as const satisfies Record<string, SourceState>
```

The sibling PRs each delete their own entry here (`#7028` removes `account:`, `#7031` `balance:`,
`#7034` `billing:`). This one does not.

### Why nothing catches it

Runtime is saved only by spread order — `...NOT_CONFIGURED_SOURCES` is spread *before*
`consent: consent.state`, so the real state wins and the API response is correct today. The stale
line is currently dead.

Verified green on the merge, so this is not theoretical:

```
git merge origin/codex/consent-approval-inbox   (onto origin/main, clean)
npm test  ->  exit 0    Test Files 374 passed (374)
                        Tests 1954 passed | 1 skipped (1955)
```

(run via `npm test` so the `pretest` hook bakes `governance.json` / `catalog.json`)

### The part that matters — the guard goes vacuous

`approvals-source-truth.guard.test.ts` on `main` asserts only `balance: 'not-configured'`, which
this PR leaves untouched, so it passes. But `#7031` and `#7034` both **add**
`expect(routeSource).toContain("consent: 'not-configured'")` to that guard. Once either lands, that
assertion means "consent is a known-but-unwired queue" — and this branch's stale line satisfies it
while consent is fully wired and returning `ok`.

I simulated exactly that: added the `consent` assertion to the guard on top of this merge and
re-ran the suite.

```
EXIT=0    Test Files 374 passed (374)
          Tests 1954 passed | 1 skipped (1955)
```

The guard passes while asserting the opposite of the truth. That is the failure the guard exists to
prevent, and it is the same warning `#5686`'s closing comment already wrote down: the exemplar
"must stay a queue the repo knows exists and does not read, or the guard stops discriminating."

### Suggested fix (owner's call)

Delete `consent: 'not-configured'` from `NOT_CONFIGURED_SOURCES`, and move the guard's exemplar on
to a queue that is genuinely still unwired at the time this lands.

Not verified: the Kotlin `ApprovalResource` and `openapi.yaml` halves, and whether
consent-service actually serves `/api/v1/consents/approvals`.

JiRaska:
## `consent` is declared not-configured *and* fully wired — the override is positional, not intentional

On this branch, `route.ts` does both:

```ts
24: const NOT_CONFIGURED_SOURCES = {
25:   account: 'not-configured',
26:   balance: 'not-configured',
27:   billing: 'not-configured',
28:   consent: 'not-configured',      // ← not removed
    } as const satisfies Record<string, SourceState>
...
377:      ...NOT_CONFIGURED_SOURCES,
...
388:      consent: consent.state,      // ← wins only because it comes after
```

The sibling PRs each delete their own entry from `NOT_CONFIGURED_SOURCES`; this one keeps it. The response is still correct today, but **only because the spread happens to be written first**.

That is a worse failure mode than a stale entry, because the thing that makes it safe is invisible to everyone who touches the file next. Moving `...NOT_CONFIGURED_SOURCES` to the end of the object — an ordinary tidy-up with no apparent behaviour, and the kind of change a formatter or a reviewer suggests — would silently flip **every wired source** back to `not-configured`. The inbox would then report that fourteen approval queues are not configured, which is exactly the state this PR family exists to fix, and nothing would fail: the shape is valid, the types are satisfied (`SourceState` covers both), and the endpoint still returns 200.

## The guard does not catch it, and I measured that rather than assuming

- `main` + this branch: `npm test` → exit **0**, 374 files / 1954 passed. The defect ships green.
- Adding the assertion that #7031 and #7034 both introduce — `toContain("consent: 'not-configured'")` — and re-running: **exit 0 again**.

So the guard passes while asserting the exact opposite of what the code does. It is not discriminating between "consent is unwired" and "consent is wired but still listed", because the string it looks for is present in both worlds.

## Fix

Delete line 28, as the sibling PRs do for their own service. Then the entry at 388 is the only statement about `consent`, and the object's key order stops being load-bearing.

Worth considering for the family as a whole: assert the **resolved** `sources` object from a response, not the presence of a source string in the file. A test that reads the built value cannot be fooled by ordering, and would have failed here.

## Sequencing

All four PRs in this family conflict pairwise — I simulated all 12 ordered pairs from `origin/main` and **every pair conflicts on the second merge**, because each edits the same single-line destructure and the same single-line `items` spread. There is no order that lands two unattended; land one, then rebase the other three. After each merge, `wired + not-configured` in `route.ts` must total **17** — if it reads 16, an entry was dropped.

Note the guard files also disagree between siblings: #7031 asserts `not.toContain("balance")` while #7034 keeps `toContain("balance: 'not-configured'")`. Those cannot both hold, and that conflict is the protection working.

JiRaska:
Aktualizace po bezpečném rebase na současný main: předchozí triage komentář popisoval starý head a na b594d4ddc39f7936c3abaa330e2812efc1cea6b9 už není aktuální. account + consent jsou wired; balance + billing zůstávají explicitně not-configured a starý consent not-configured zápis je odstraněn. Konfliktní route/test změny byly sloučeny ručně se zachováním všech zdrojů. Ověření na tomto headu: approvals inbox + source-truth 21/21, Admin UI type-check, backend ApprovalResourceMappingTest, ktlintCheck, detekt a quarkusBuild jsou zelené. Zbývající červené CI gatey jsou stejné repo-wide chyby aktuálního main a mají samostatné opravné PR #8012 a #8015. Money-path požadavek na dvě nezávislá schválení zůstává beze změny.

Issue #7091, opened 2026-08-26
Admin UI: serialize document-template writes


JiRaska:
## Problem\nDocument Template save and publish/retire flows rely only on React saving/actioning state. A repeated submit or confirmation in the same event loop can enqueue duplicate writes before disabled controls render; save and lifecycle actions also use separate flags and can overlap.\n\n## Scope\n- claim one synchronous lock across save, publish and retire\n- release it on all success/failure paths\n- expose truthful busy semantics on the save control\n- preserve BFF routing, payloads, immutable published versions, RBAC and four-eyes semantics\n\n## Acceptance\n- overlapping template writes are rejected before fetch\n- later writes are accepted after settlement\n- focused tests, lint and CI gates pass

PR #7092, opened 2026-08-26
fix(admin-ui): serialize document-template writes

JiRaska:
## Summary\n- synchronously reject overlapping template save, publish and retire writes\n- share one lock across draft and lifecycle actions and release it on every outcome\n- expose truthful save busy state while preserving the existing confirmation workflow\n\n## Safety\nThis PR is intentionally stacked on #7085 for the tested shared single-flight primitive. BFF routing, request payloads, immutable published versions, RBAC and four-eyes semantics are unchanged.\n\n## Verification\n- 7 focused guard/workflow assertions passed\n- scoped ESLint: 0 errors (one pre-existing out-of-diff effect warning)\n- git diff --check passed\n- full type/build runs on CI with the stacked dependency\n\nCloses #7091

### DIFF
--- a/openbank-admin-ui/e2e/operator-cockpit.spec.ts
+++ b/openbank-admin-ui/e2e/operator-cockpit.spec.ts
@@ -136,7 +136,7 @@ test('Temporal and approvals show live source-backed operator state and human pr
   await expect(page.getByText(/Provozní|Running/)).toBeVisible()
   await page.getByRole('button', { name: /Metriky|Metrics/ }).click()
   await expect(page.getByRole('heading', { name: /Workflowy \(posledních 60 minut\)|Workflows \(last 60 minutes\)/ })).toBeVisible()
-  await expect(page.getByText('12')).toBeVisible()
+  await expect(page.getByText('12', { exact: true })).toBeVisible()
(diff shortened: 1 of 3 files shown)


JiRaska (commented):
Reviewed exact head 836e8dfc7 against stacked base codex/admin-ui-product-studio-single-flight: no blocker found. Both form save and publish/retire claim the same ref-backed lock before any state update or network request, and both release it in finally. The existing immutable-version workflow, confirmation UX, BFF routing, payloads and RBAC/four-eyes rules are unchanged. Local evidence: 7 focused assertions, scoped ESLint with zero errors, signed commit verification and git diff --check passed. Independent approval, green CI and parent #7085 landing first remain required.

JiRaska (commented):
Reviewed amended exact head ec80f0127: no blocker found. The only follow-up makes the Temporal metric assertion exact, avoiding a strict-mode collision with the live Updated timestamp; document-template single-flight behavior and all BFF/RBAC/four-eyes contracts are unchanged. Signed commit and git diff --check passed. Independent approval, restarted green CI and parent #7085 landing first remain required.

PR #7168, opened 2026-08-26
fix(admin-ui): preserve account opening idempotency

### LINKED ISSUE #7167: Preserve idempotency across admin account-opening retries
The admin account form generated a new idempotency key on every submit. The account service safely replays a matching key, but a second event or retry after an interrupted response therefore bypassed that protection. Keep one key for a single opening attempt and synchronously reject duplicate submits. Refs #7113.

JiRaska:
Closes #7167.

The account-opening key now remains stable for one operator attempt, allowing the account service to safely replay a request after an interrupted response. A synchronous single-flight lock also rejects a second submit event before React renders the button disabled.

Verification: git diff --check; node --check on the guard test. Focused Vitest is unavailable in the clean worktree because dependencies are not installed; CI remains authoritative.

### DIFF
--- a/openbank-admin-ui/src/app/accounts/new/page.tsx
+++ b/openbank-admin-ui/src/app/accounts/new/page.tsx
@@ -4,7 +4,7 @@
 
 'use client'
 
-import { useEffect, useState } from 'react'
+import { useEffect, useRef, useState } from 'react'
 import { useRouter } from 'next/navigation'
 import Link from 'next/link'
 import { ArrowLeft, Save, AlertCircle } from 'lucide-react'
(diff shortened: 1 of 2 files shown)


JiRaska:
CI status note: `gates (gitops)` is failing on unrelated repository-baseline findings outside this PR’s diff (fck-nat AMI pinning, generated OPA bundle newline, CNPG backup associations, and Temporal namespace registration). This PR changes only two admin-ui files. I will not mix GitOps remediation into this UI safety change; the PR remains unmergeable until the baseline gate is repaired/re-run, in addition to the required independent review.

PR #7192, opened 2026-08-26
feat(referral): deliver lifecycle events

### LINKED ISSUE #7190: Referral: deliver lifecycle events through transactional outbox
The governed referral/MGM slice currently persists reward state while its production event adapter explicitly drops referral lifecycle events. Add an atomic transactional outbox, at-least-once Kafka relay, backlog/retry telemetry, and a real-DB HTTP replay assertion proving qualification writes exactly one Qualified and one RewardRequested event. This does not authorize ledger posting or promo-code value handling; those remain separate governed scopes.

JiRaska:
## Summary
- replace the explicit event-dropping adapter with a transactional referral outbox
- commit reward state and lifecycle events atomically, then relay at-least-once with canonical headers and cross-pod claims
- prove qualification replay leaves exactly one Qualified and one RewardRequested row in the real Postgres HTTP lifecycle test
- document the actual multiplexed Kafka address in AsyncAPI 3.0

## Safety boundary
This does not post money or implement promo codes. The reward request remains a contract boundary pending its separate money-path decision and threat model. Deployment remains gated on reviewed Kafka topic/user/TLS GitOps and the zero-replica referral workload activation.

## Verification
- `./gradlew :openbank-referral-service:ktlintFormat :openbank-referral-service:test :openbank-referral-service:ktlintCheck :openbank-referral-service:detekt :openbank-referral-service:quarkusBuild`
- `python3 .github/scripts/check-event-contract-code-agreement.py`
- `python3 .github/scripts/check-event-schema-compat.py --base origin/main --enforce`

Closes #7190


JiRaska:
Triage note (measured, no changes made). Three findings; the first blocks merge.

### 1. `openapi.yaml` version collision — campaign-service already shipped 1.43.0

This branch forked at `84613ae49`, where campaign-service was at `1.42.0`, and bumped it to `1.43.0`. While it sat, #7573 landed #5962 on `main` and took **the same** `1.43.0` for a different change (documenting `CONVERTED` in `SendRecord.outcome`):

```
git log --oneline -S'version: 1.43.0' origin/main -- openbank-campaign-service/src/main/resources/openapi.yaml
# 35c82f92b fix(campaign): document CONVERTED in SendRecord.outcome (#5962) (#7573)

git show 84613ae49:openbank-campaign-service/src/main/resources/openapi.yaml | grep -m1 '^  version:'   #   version: 1.42.0
git show origin/main:openbank-campaign-service/src/main/resources/openapi.yaml | grep -m1 '^  version:'  #   version: 1.43.0
```

Both sides now spell `version: 1.43.0`, so git sees identical text and there is nothing to conflict. A real merge into live `main` confirms it — exit 0, and the merged file carries `1.43.0` **plus** this PR's new surface:

```
git merge --no-commit --no-ff origin/codex/referral-event-transport    # exit=0
grep -m1 '^  version:' .../openbank-campaign-service/.../openapi.yaml  #   version: 1.43.0
grep -c referralProgramRef  .../openapi.yaml                           # 1
grep -c CONVERTED           .../openapi.yaml                           # 2
```

So `referralProgramRef`, `referralProgramId` and the `ReferralProgramRef` schema would publish under a contract version `main` has already released. CI cannot see this — the last run here predates #7573. Needs a re-bump to `1.44.0` against live `main`.

Referral-service is fine: `main` is `1.0.0`, this branch `1.2.0`, unclaimed elsewhere.

### 2. Conflicts with #7553 on a shared GitOps file

Each merges clean alone; together they do not. Both append one entry at the same point of `openbank-infra/gitops/components/payments/kafka-scheme-accepted-acl.yaml` (line 460):

```
# this PR
+      - referral-service                # ADR-0266: MGM lifecycle producer, ns referral
# PR #7553
+      - incentive-service
```

Merging this PR then #7553 gives `CONFLICT (content)` on that file. Whichever lands second needs a manual resolution — worth noting because both entries must survive.

### 3. #7565 already contains this PR's referral work, byte-identically

Of this PR's changed files, only two differ on `codex/mgm-funnel-projection-restored`; the rest hash identically, including the whole referral outbox slice:

```
ReferralOutboxTransport.kt   7192=c7a41100eed2  7565=c7a41100eed2
ReferralOutboxRepository.kt  7192=fec8a1ed734a  7565=fec8a1ed734a
referral openapi.yaml        7192=b34db10d45db  7565=b34db10d45db
V2__referral_transactional_outbox.sql  7192=61fe2866c156  7565=61fe2866c156
campaign openapi.yaml        7192=753c93da6a47  7565=753c93da6a47
# divergent: admin-ui campaigns/referrals/page.tsx, openbank-libs/governance/rules.yaml
```

Because the shared content is identical the two merge in either order, but #7565 carries the **same** campaign `1.43.0` collision as finding 1 — fixing it in one place does not fix the other.

JiRaska:
**Measured: fully contained by #7565 — close candidate.**

Zero of this PR's 46 files differ between the two branches; the control (all-file diff between them) shows 13 files do differ, so the probe is live and the emptiness is real.

```
git diff --name-only origin/codex/referral-event-transport origin/codex/mgm-funnel-projection-restored \
  -- $(git diff --name-only origin/main...origin/codex/referral-event-transport)     # -> empty
git diff --name-only origin/codex/referral-event-transport origin/codex/mgm-funnel-projection-restored | wc -l   # -> 13 (control)
```

#7565 is the superset (56 files, +1735 vs +1377) and merges into `main` with **0 conflicts**. This PR is `CONFLICTING`, so it has had zero CI runs and its green history is about an old base.

Note for #7565: its base is `codex/campaign-mgm-reference-restored`, whose PR #7564 is already MERGED, so that base branch is dead — retarget #7565 to `main` before merging.

JiRaska:
**Measured: campaign spec edited without a version bump; referral is fine.**

```
openbank-campaign-service/src/main/resources/openapi.yaml
  origin/main : version: 1.43.0
  this branch : version: 1.43.0     <-- edited, not bumped

openbank-referral-service/src/main/resources/openapi.yaml
  origin/main : version: 1.0.0
  this branch : version: 1.2.0      <-- OK
```
The `openapi-version-not-taken` gate landed on main 2026-08-31 06:44 (`a373867e9`, #7633,
`.github/gates/gates.yaml:2633`), so the campaign half should fail on the next run.

This PR is also `CONFLICTING`. A test-merge of `origin/main` (detached worktree, aborted afterwards) yields
a single conflict: `openbank-libs/governance/rules.yaml`. Worth resolving by hand rather than by taking one
side — two branches adding neighbouring entries to the same list merge clean and keep only one.

Fix: `info.version: 1.44.0` for campaign, re-read live `main` first.

JiRaska:
Rebased onto current `main`. Three conflicts, and they wanted three different answers — worth spelling
out because "take main" or "take the branch" would each have been wrong twice.

**`CampaignResource.kt` — union.** `main` changed `steps` to `List<StepRequest?>` (the Jackson
collection-element fix: the module null-checks constructor parameters, not collection elements, so
`{"steps": [null]}` deserialises into a list holding a null). This branch adds `referralProgramId` to
the same constructor. Kept both:

```kotlin
val steps: List<StepRequest?>,
/** Published referral programme UUID; metadata is resolved and pinned server-side. */
val referralProgramId: UUID? = null,
```

**`kafka-scheme-accepted-acl.yaml` — union.** `incentive-service` (main) and `referral-service` (this
branch) are different producers; the ACL wants both entries.

**`rules.yaml` — took main, which is a deletion.** This branch re-adds two topic exemptions that main
has removed, and the removal is right: both exemptions say *"the consumer isn't on main yet"*, and
both consumers now are.

```
openbank.referral.events      1 module on main
openbank.billing.fee.event    2 modules on main
```

Re-adding them would restore a declaration that has outlived its subject — the same shape that turned
`gates (supplychain)` red fleet-wide yesterday. #7194's merge-order gap has closed on its own.

`:openbank-campaign-service:compileKotlin` passes on the merged result.

One note on process: this branch is `codex/*`. I checked before pushing and no worktree held it, which
is the signal I use for live work, so I treated it as free. The push is a merge commit on top of the
existing head — nothing rewritten, nothing dropped — but if this is being worked elsewhere, say so and
I will leave it alone.

JiRaska:

```
```

```
```

X an ancestor".

JiRaska:
## Merge-blocker: the campaign spec changed and its `info.version` did not

```
$ git show origin/main:openbank-campaign-service/src/main/resources/openapi.yaml | grep -m1 '^  version:'
  version: 1.43.0
$ git show origin/<this-branch>:...openapi.yaml | grep -m1 '^  version:'
  version: 1.43.0

$ ... | shasum          main 3299648a…   branch 0e789746…
$ git show origin/<this-branch>:...openapi.yaml | grep -c ReferralProgramRef
2
```

Same version, different document. The branch adds a `ReferralProgramRef` schema and a `referralProgramId` property — an additive API change published under a number `main` has already released.

The referral spec in this same PR bumped correctly (`1.0.0 → 1.2.0`), so this is an oversight rather than a policy choice.

## How it happened — worth knowing, because it will recur

#7204 bumped this spec correctly, `1.41.0 → 1.42.0`. Then this branch's HEAD commit — literally `Merge remote-tracking branch 'origin/main' into HEAD` — brought main's `1.43.0` in and **clobbered its own bump**. Nothing conflicted: git saw one version line on each side and took the newer text.

That is the same class as the four version collisions found this week, but from the other direction. A collision is two branches *claiming* one number; this is a branch *losing* the number it had claimed. Both are invisible to git, and both survive review because the file reads correctly in isolation.

## CI cannot see it here

`build (openbank-campaign-service) … (contract)` is **SKIPPED** on this PR, along with the referral one — 12 skipped, 28 success, 5 failure. So the gate that would catch it never ran.

Proven by running it directly against the merge-base:

```
check-api-contract.py --base <merge-base> --enforce
  EXIT=1  additive API change requires an info.version MINOR bump,
          but it moved 1.43.0 -> 1.43.0
```

and after a one-line edit to `1.44.0`, `EXIT=0` with both services OK.

**Re-read the live value before pushing the fix** — `main` moves, and `1.44.0` is only free until someone else takes it.

## One thing worth a look while you are in here

`revise` wraps its whole call in `runCatching{}.getOrElse { 409 CONFLICT }`, so a **referral-service outage renders as 409** — indistinguishable from "your programme reference is invalid". `create` catches only `CampaignReferenceNotFoundException`, so the same outage there becomes a 500. Same condition, two codes, and 409 is the wrong one.

The `runCatching` predates this PR; what is new is that a remote HTTP call now flows through it. Not a blocker, but the pin design is otherwise careful about exactly this distinction — the adapter maps only 404 → null and rethrows the rest, so "unavailable" and "unpublished" stay separate right up until that boundary.

## Also checked and clear, so nobody re-litigates

- **Entity column names** — campaign-service does set `CamelCaseToUnderscoresNamingStrategy`, so the un-annotated `referralProgramId/Name/Version` properties resolve correctly. Nearly filed this before checking.
- **Migration collision** — `V19` is free; `main` tops out at `V18` and no other open PR claims it.
- **`duplicate` with a stale pin** — hypothesised a 500, and it is not: `CampaignReferenceNotFoundException` extends `NoSuchElementException`, which `duplicate` catches → 409, fail-closed. Correct, just untested.

**#7203 cannot close until this lands**, and this should not land at `1.43.0`.

JiRaska:
### Two gates failed after my rebase, and only one of them was pre-existing

**`event-consumer-liveness` was my error, and it is worth naming the shape.** Resolving the earlier
merge I took main's `rules.yaml` on the reasoning that the consumers for these topics now exist. That
measurement was wrong:

```
git grep -l 'openbank.referral.events' origin/main -- '*/src/main/*'
  openbank-referral-service/.../UnwiredReferralEventPublisher.kt      <- the PRODUCER
```

I counted **mentions** of the topic and the single hit was the producer. Asking the right question —
`@Incoming` / `mp.messaging.incoming` — gives zero consumers. So the exemption this branch carried is
still needed, and I deleted it. Restored verbatim from the branch's own commit.

Why `main` is green without it: the producer is unwired there. **This PR wires it, which is precisely
what makes the entry necessary** — so "main removed it, therefore it is obsolete" was the wrong
inference twice over.

**`source-service-presence` is a real gap in this PR**, from a gate that landed on `main` on
2026-09-01, after this branch was cut. The referral producer declares an outgoing `openbank.*` channel
and never stamps `sourceService`, so its audit rows would be attributed by derivation from the topic
name — a convention rather than a statement by the producer.

Fixed on the sealed base rather than per subclass, so no future event type can be added without it:

```kotlin
private const val SOURCE_SERVICE = "openbank-referral-service"

sealed class ReferralEvent {
    ...
    val sourceService: String = SOURCE_SERVICE
}
```

That is the idiom the gate documents by name — `card-issuance` and `fx-service` declare it the same
way, and the gate explicitly recognises it alongside the literal `"sourceService" to X` map form.

Verified on the merged result:

```
check-event-consumer-liveness.py            exit 0   (referral allowlisted, with its reason)
check-source-service-presence.py            exit 0   37 producers, 27 declare the field
:openbank-referral-service:compileKotlin    exit 0
```

JiRaska:
Two more gates, one cause — the campaign spec's `info.version` was taken while this branch waited:

```
api-contract gate: additive API change requires an info.version MINOR bump,
                   but it moved 1.43.0 -> 1.43.0
openapi-version-not-taken: 1.43.0 is ALREADY on the base branch — that number is taken
```

This is the failure mode a version line is uniquely exposed to: the branch proposed `1.43.0`, `main`
published `1.43.0`, and the two merged **cleanly** because git compares text rather than which numbers
are already published. Nothing conflicted; the number was simply spent.

The change itself is additive — `referralProgramRef` and a nullable `referralProgramId` on the
campaign request — so MINOR is the right class under ADR-0048 D5 and only the number needed moving.
Now `1.44.0`, free on `main` as of this push. Worth re-reading after any competing campaign spec
merge; whoever lands second takes the next one.

JiRaska:
Honest status after today's rebase attempt: the branch now merges, but it is **8+ governance gates red** against current main (slo-registry-consistency, journey-money-path-accountability, governance-lineage, event-consumer-liveness, no-runblocking-in-scheduled, clock-injection, accounting-clock, audit-money-path-subscription) — two months of fleet-wide gates this branch predates. None of these is a mechanical fix: each asks for per-service registry/lineage/clock-wiring work on referral-service that amounts to reimplementing the change on current main rather than rescuing this diff. Leaving the PR open as the design reference for #7190, but it should not be merged in this state; the follow-up should start from a fresh branch off main and port the outbox shape (which is still the right design) rather than rebase this one.

JiRaska:
## Collides with #8859 — the referral outbox is implemented twice

Measured, not inferred: 14 files overlap between this PR and #8859
(`agent/feat-referral-outbox-kafka-relay-7190`, opened 2026-09-05), and both add a
**`V2__` Flyway migration to `openbank-referral-service` under a different filename** —
`V2__referral_transactional_outbox.sql` here, `V2__create_referral_outbox.sql` there.
That merges textually clean and then refuses to boot with
`FlywayException: Found more than one migration with version 2`, taking every
`@QuarkusTest` in the module to SKIPPED rather than FAILED. So these two cannot both land
as they stand, and git will not say so.

State on `origin/main` today: referral has **only** `V1__referral.sql` and no outbox at all,
so neither side has landed and there is no third copy to reconcile against.

Of the two, #8859 is the more complete referral-outbox implementation — it carries the
dispatcher, the dead-letter gauge, the outbox entity and repository, the Kafka topic CR and a
real-DB `ReferralOutboxWriteIT`, none of which exist here — and it is merely BEHIND, whereas
this PR is DIRTY.

**Proposed resolution, by subtraction rather than by picking a winner:** let #8859 own the
referral outbox transport, and reduce this PR to the part #8859 does not touch at all —
the campaign↔referral integration (`ReferralProgramClient`, `V19__campaign_referral_program_reference.sql`,
the campaign spec and its tests) and the admin-ui referral screens. That is genuinely new
work and nothing else in the queue covers it.

Note while renumbering: `openbank-campaign-service` is at `V9` on `main`, so `V19` here leaves
a ten-version gap — harmless to Flyway, but it suggests this branch was cut against a state that
no longer exists. Re-derive the next free version from live `main`, not from the branch base.

Left open deliberately; the subtraction is an authoring decision, not a mechanical merge.

JiRaska:
## I attempted the subtraction I proposed. It does not work mechanically — measured twice.

Earlier I suggested reducing this PR to the campaign↔referral and admin-ui work and letting #8859
own the referral outbox transport. I tried it rather than leaving it as advice.

**File split:** 32 files here that #8859 does not touch, 14 shared. 13 of the 14 already exist on
`main`, so "take main's version" looked like a clean cut.

**Attempt 1** — reverted all 14 shared files to `main`, dropped this branch's `V2`/`V3` migrations
and the outbox classes:

```
e: ReferralResource.kt:60:81 Unresolved reference 'publishedProgram'
e: ReferralResource.kt:61:62 Unresolved reference 'id'
```

`ReferralResource.kt` is one of this PR's *unique* files, and it needs `publishedProgram` from
`ReferralService.kt` — a *shared* one. My assumption that campaign only reaches referral over REST
was wrong: the branch's own REST resource depends on the branch's service API.

**Attempt 2** — restored the four service/domain files from this branch, keeping only the outbox
transport removed:

```
e: UnwiredReferralEventPublisher.kt:6:51 Unresolved reference 'ReferralEventPublisher'
e: UnwiredReferralEventPublisher.kt:8:37 Unresolved reference 'ReferralPublishOutcome'
```

Now `main`'s publisher does not match this branch's ports. The pieces are not separable: the
referral half of this PR is one coherent unit — ports, models, service, publisher, outbox — and
any mix of branch and main halves fails to compile.

## What that means

#7192 and #8859 genuinely exclude each other on `openbank-referral-service`. Both add a `V2`
migration (`V2__referral_transactional_outbox.sql` vs `V2__create_referral_outbox.sql`) plus a
conflicting `V3`, and `main` still holds only `V1`, so whichever lands second does not merely
conflict — it produces `FlywayException: Found more than one migration with version 2` and takes
every `@QuarkusTest` in the module to SKIPPED, which reads as a pass.

Reducing this PR to its campaign/admin-ui half is therefore **development work, not a merge**:
`ReferralResource`/`ReferralProgramClient` would have to be rewritten against #8859's API.

## Recommendation, unchanged in substance but now evidenced

Land **#8859** first — it is the more complete transport (dispatcher, dead-letter gauge, outbox
entity and repository, `KafkaTopic` CR, a real-DB `ReferralOutboxWriteIT`), and none of that
exists here. Then rebuild this PR's campaign↔referral link on top of it. The 32 unique files here
are genuinely wanted and nothing else in the queue covers them.

I have pushed nothing to this branch — the attempts above were local and are discarded.

JiRaska:
Closing as superseded for active WIP accounting. #8859 owns the referral transactional-outbox transport and is the safer implementation to finish first; this PR remains available as the historical design reference for the later campaign/admin-UI slice. The two branches must not merge together because both introduce referral Flyway V2/V3 migrations.

Issue #7184, opened 2026-08-26
perf(admin-ui): remove third-party font fetches from the operator shell


JiRaska:
## Evidence

The admin UI imports Google Fonts in `src/app/globals.css` and preconnects to `fonts.googleapis.com` and `fonts.gstatic.com` from the root layout. CSP explicitly permits the external style origin. This adds third-party requests to every operator shell render and leaves the typography dependent on an external service.

## Desired outcome

Serve the chosen font assets from the OpenBank build/runtime (or deliberately use an approved local system stack), remove external font CSP/preconnect allowances, and prove the built shell has no Google Fonts request while preserving font fallback and layout stability.

## Constraints

Do not trade away typography quality or introduce a build-time fetch dependency without a reproducible vendoring path. Include browser-level verification and consider licensing/subsetting/weight coverage.

JiRaska:
## Verified resolved on `main` — plus one related surface still leaking

Confirming this closed correctly, with evidence, and recording a guard gap I found and fixed.

**The operator shell is clean.** `src/app/layout.tsx` self-hosts via `next/font` (build-time fetch, cached under `.next/static`), `globals.css` has no `@import`, and CSP is tightened on both paths that matter — `src/proxy.ts` has `font-src 'self'` and `openbank-infra/k8s/base/ingress.yaml` likewise. The CSP half was the important one: removing the fetch without tightening the policy would have left the hole open.

**But the guard's scope was a hand-kept list.** `no-google-fonts.guard.test.ts` read three named files while `src/` holds **14 stylesheets**. Measured: a Google Fonts `@import` added to `src/app/auth/login/login.module.css` — a *public auth surface* — left the suite **green (exit 0)**. #7288 derives the scope by sweeping the tree instead; the same poisoned file then fails it (exit 1) and is named in the output.

## Related, out of scope here, and a judgment call rather than a defect

`openbank-infra/web/landing/` still fetches Google Fonts, and its CSP permits the origins:

- `index.html:13-15` and `platform.html:10-12` — `preconnect` + stylesheet `<link>` to `fonts.googleapis.com` / `fonts.gstatic.com`
- `openbank-infra/aws/modules/static-site/main.tf:140-141` — `style-src ... https://fonts.googleapis.com`, `font-src 'self' https://fonts.gstatic.com`

[...]

Issue #7073, opened 2026-08-26
perf(admin-ui): keep public auth surfaces session-free


JiRaska:
## Problem
The root layout mounts NextAuth SessionProvider and AgentDock on every route. Public `/auth/*` and `/privacy` pages therefore initialize authenticated-only client infrastructure and request `/api/auth/session` even though those surfaces neither need nor expose a session. This adds avoidable work and a noisy authentication request to the fastest, most security-sensitive first impression.

## Scope
- Introduce one route-aware provider boundary.
- Keep language and toast infrastructure available everywhere.
- Mount NextAuth reauthentication and AgentDock only on protected operator routes.
- Preserve existing protected-route behavior and public-page rendering.
- Add focused regression tests for public and protected routes.

## Acceptance
- Public auth/privacy routes do not mount SessionProvider or AgentDock.
- Protected routes still mount both.
- No auth, RBAC, callback, or agent endpoint behavior changes.
- Focused tests, type-check, lint, production build and browser request inspection pass.

JiRaska:
## Disposition: resolved on `main` — closing with evidence

Every acceptance criterion in this issue is already satisfied on `origin/main`. It landed as part of c36fcbcb5 (`fix(admin-ui): modernize secure sign-in experience`, #7063) rather than under its own PR, which is why it stayed open.

**The route-aware provider boundary exists** — `src/components/layout/AppProviders.tsx`:

```tsx
const publicSurface = isPublicSurface(pathname)
const shared = (
  <LanguageProvider>
    {children}
    {!publicSurface && <AgentDock />}
    <Toaster richColors position="top-right" />
  </LanguageProvider>
)
return publicSurface ? shared : <SessionProvider>{shared}</SessionProvider>
```

Against each acceptance criterion:

- *Public auth/privacy routes do not mount SessionProvider or AgentDock* — yes; `isPublicSurface` covers `/auth`, `/auth/*` and `/privacy`.
- *Protected routes still mount both* — yes.
- *Language and toast infrastructure available everywhere* — yes, `LanguageProvider` and `Toaster` are outside the branch.
- *No auth, RBAC, callback or agent endpoint behavior changes* — none in that commit.
- *Focused regression tests* — `src/test/app-providers.test.tsx`.

**I verified the guard is real rather than decorative.** A test that cannot detect the absence of what it tests is decoration, so I regressed `AppProviders` to mount `SessionProvider` and `AgentDock` unconditionally and re-ran it:

[...]

PR #7288, opened 2026-08-27
fix(admin-ui): close public-surface guard gaps in the auth gate and font sweep

JiRaska:
## What

Two guards protecting the **public, unauthenticated** admin-UI surfaces had scope gaps that let the very defect they exist to prevent walk past them. This PR closes both. **No behavior change on any existing route.**

## 1. The auth gate and the provider boundary could drift (#7073)

Two independent lists decided "is this route public", and nothing made them agree:

| | decides |
|---|---|
| `src/proxy.ts` | skip the auth gate (no session needed to reach the page) |
| `src/lib/auth/publicSurface.ts` | skip `SessionProvider` / `AgentDock` (#7073) |

The proxy matched `startsWith("/auth")` — a loose prefix that would serve **any** future `/auth*`-named route (`/authorization`, `/auth-settings`) with no auth gate, while `publicSurface` classified that same path as protected and mounted authenticated-only client infrastructure on it. That is exactly the defect #7073 closed, re-entering through the other list.

**Latent, not live.** I enumerated all 50 top-level routes under `src/app`; only `auth/` and `privacy/` exist, so nothing is affected today. The fix is to stop having two lists: the proxy now delegates to the shared `isPublicSurface` predicate, and a parity guard fails when the two disagree.

## 2. The Google Fonts guard read three hard-coded files (#7184)

`no-google-fonts.guard.test.ts` named `layout.tsx`, `globals.css` and `proxy.ts` individually — making the guard's scope a hand-kept list while the tree holds **14 stylesheets**. Measured: a Google Fonts `@import` added to `src/app/auth/login/login.module.css` — a *public auth surface* — left the suite **green**.

Scope is now derived by sweeping `src/` for `.css/.ts/.tsx/.js/.jsx`, plus an assertion that the sweep is non-empty so it cannot pass vacuously.

## Verified by what it prevents, not by what it prints

Exit codes read from `$?` after redirecting each run to a file:

| check | exit |
|---|---|
| parity guard vs **unmodified `main`** | **1** (red — the divergence is real) |
| parity guard with the fix | 0 |
| parity guard, loose prefix restored | **1** |
| **old** font guard vs poisoned login CSS | **0** (the blind spot, measured) |
| **new** font guard vs same poisoned CSS | **1** (names the offending file) |
| new font guard on clean tree | 0 |

One probe trap worth recording: the parity guard's source assertion initially matched **its own explanatory comment** describing the old condition. It now strips comments before asserting, so it reads code rather than prose.

## Gate results

`npm test` (never bare `npx vitest run`) — **369 files / 1910 tests green, 1 skipped**; `tsc --noEmit` 0; `npm run lint` 0 errors.

Refs #7073
Refs #7184


JiRaska:
### `gates (kotlin)` failure is pre-existing on `main`, not from this PR

This PR changes three TypeScript files and no Kotlin. The failing gate reports:

```
new service-owned Testcontainers resource lacks lifecycle evidence:
openbank-incentive-service/src/test/kotlin/com/openbank/incentive/it/IncentivePostgresTestResource.kt
[rule A] a no-op factory must not set a success flag — give the outcome its own enum value
```

Evidence it is not mine:

- `git diff --name-only origin/main...` for this branch lists only `openbank-admin-ui/src/proxy.ts` and the two guard tests.
- That Kotlin file already exists on `origin/main`, landed by e156db259 (#7232, `feat(incentive): deliver governed promo reservations`).
- **`main` itself is failing the same gate** — the two most recent `ci.yml` runs on `main` (`44dc8053`, `78012e3b`, both 2026-08-27T04:4x) are `failure`, while the runs before #7232 landed (`89ebb79f` and earlier) are `success`.

So `main` is currently red on `gates (kotlin)` independently of this PR. Flagging it here rather than fixing it: it is a different service and someone else's active area. Worth its own issue.

Issue #7210, opened 2026-08-26
Incentive service: governed promo-code reservation and redemption


JiRaska:
Implement ADR-0266's separate Incentive/Promo bounded context; do not add a free-form promoCode field to campaign-service.\n\nAcceptance:\n- immutable versioned offer terms with product scope, effective/expiry window, total and per-party limits, stacking policy and maker/checker publication\n- opaque code inventory; stored/logged representations do not expose redeemable plaintext unnecessarily\n- reservation -> commit/release lifecycle with expiry and stable idempotency key\n- replay/concurrency cannot consume inventory or a party limit twice\n- append-only audit/outbox events and explicit retention handling\n- real HTTP/Postgres integration covering publication, reserve, concurrent replay, commit, release and expiry\n- OpenAPI + versioned AsyncAPI contracts\n- Campaign Studio may select an immutable offer reference but never owns redemption or discount mutation\n- no direct monetary/account mutation; any settlement/pricing integration requires its own reviewed money-path design\n\nRefs ADR-0266.

JiRaska:
**Triage: this does NOT belong in the MGM referral family — and most of it is already on `origin/main`.**

It is independent, measured rather than assumed: `git grep -i referral origin/main --
openbank-incentive-service` returns **zero hits**, the service publishes to its own topic
`openbank.incentive.events` (`openbank-contracts/openbank-incentive-service/asyncapi.yaml`, 1.1.0),
and it consumes no referral channel. It shares no backbone with #7190/#7194/#7198/#7203 — sequencing
it behind three Campaign Studio issues would bury a money-adjacent write path behind unrelated UI work.

Already present on `origin/main`:

- immutable versioned offers with maker/checker publication — `POST /offers`, `/offers/{id}/submit`,
  `/offers/{id}/publish`; `IncentiveOffer.kt`
- code inventory — `POST /offers/{id}/codes`; `PromoReservation.kt`
- reserve → commit/release/expire lifecycle — `/offers/{id}/reservations`,
  `/reservations/{id}/commit`, `/reservations/{id}/release`, `/maintenance/expire`
- outbox — `V2__outbox_dispatch.sql`, `IncentiveOutboxDispatcher` (a `@Scheduled` **`suspend fun`**,
  correct), `KafkaIncentiveOutboxEventPublisher`, and `openbank.outbox.dispatch-enabled: true` set in
  `application.yaml` (the default is `false`, so this one matters)
[...]

PR #7553, opened 2026-08-28
chore(infra): stage Incentive runtime prerequisites

JiRaska:
Refs #7210

Stages only the non-workload prerequisites for the Incentive runtime:
- one producer-only Kafka mTLS identity for `openbank.incentive.events`;
- ESO projections for the shared OIDC client secret and the separately managed promo-code pepper;
- the least-privilege cert-reader grant required for that projection.

No Deployment, image pin, Service, NetworkPolicy, customer-edge URL, or Argo automated sync is introduced. The component remains behind the parent data-plane Application’s manual sync. A future workload rollout must separately verify the Pod Identity prerequisite and that the operator-managed pepper reference is synced; this PR contains no secret value and is not live-health evidence.

Verification:
- `check-kafka-cert-reader-grant.py --enforce`
- `check-db-backup-associations.py --enforce`
- `gen-network-policies.py --check`
- `check-gitops-duplicate-resources.py --enforce`
- `git diff --check`


JiRaska:
Triage note (measured, no changes made).

**Merge-ordering conflict with #7192.** Both PRs merge into live `main` cleanly on their own, but not one after the other: each appends a single entry at the same point (line 460) of `openbank-infra/gitops/components/payments/kafka-scheme-accepted-acl.yaml`.

```
# this PR
+      # Incentive lifecycle producer. The corresponding ExternalSecret is staged without a
+      # Deployment, so this narrowly permits ESO to project only this Strimzi-issued client key.
+      - incentive-service

# PR #7192
+      - referral-service                # ADR-0266: MGM lifecycle producer, ns referral
```

Merging #7192 first and then this branch gives `CONFLICT (content)` on that file. Flagging it because both entries must survive the resolution — a neighbouring-entry list edit is the shape where a resolution silently keeps only one side.

Nothing else here collides: `openbank-infra/gitops/components/kafka/network-policies.yaml` is also touched by both but merges without conflict, and the release manifest stays at 60 entries on both sides.

PR #7565, opened 2026-08-28
feat(analytics): project MGM reward funnel

### LINKED ISSUE #7194: Campaign Studio: project referral lifecycle events
Consume openbank.referral.events into the Campaign Studio reporting path and expose governed MGM funnel states (qualified, reward requested, outcome) without treating clicks or invitations as rewards. Remove the temporary event-consumer-liveness allowlist entry when the consumer merges. This is the next slice after #7190 and must preserve eventId deduplication and immutable program/version attribution.

JiRaska:
Closes #7194

Restores the reviewed MGM outcome funnel on the corrected referral/campaign stack.

- consumes versioned referral lifecycle evidence into an eventId-deduplicated ClickHouse funnel by immutable programme reference
- renders qualified, reward-requested, accepted, rejected and reversed outcomes without treating invitations or clicks as rewards
- preserves a minimal referral catalogue boundary (id/name/version); funnel is an independent read model
- stages only analytics consumer ACL/schema changes; this is not incentive or campaign live-rollout evidence

Local evidence: focused Admin UI tests 4/4; Analytics lint/detekt; Docker-backed ClickHouse IT 6/6, 0 skipped.


JiRaska:
Triage note (measured, no changes made).

### This PR is a superset of #7192, not a consumer waiting on it

Worth stating because the reverse was assumed. The concern was that an MGM funnel projecting `openbank.referral.events` would be a control with nothing to disagree with — on live `main` that is exactly true: referral-service ships only `UnwiredReferralEventPublisher`, whose own KDoc says "nothing in this build produces to them", and it emits a `dropped` counter and no `published` counter.

But this branch **carries the producer itself**. Its base is `codex/campaign-mgm-reference-restored`, which already contains the full referral outbox slice, byte-identical to #7192:

```
git ls-tree -r --name-only origin/codex/mgm-funnel-projection-restored openbank-referral-service/src/main/kotlin | grep outbox
#  ReferralOutboxBacklogGauge.kt / ReferralOutboxRepository.kt / ReferralOutboxTransport.kt

ReferralOutboxTransport.kt   7192=c7a41100eed2  7565=c7a41100eed2
ReferralOutboxRepository.kt  7192=fec8a1ed734a  7565=fec8a1ed734a
V2__referral_transactional_outbox.sql  7192=61fe2866c156  7565=61fe2866c156
```

Of #7192's changed files only two differ here (`admin-ui/src/app/campaigns/referrals/page.tsx`, `openbank-libs/governance/rules.yaml`) — both because this PR adds the funnel on top. So the funnel is not projecting over an unproduced stream, and the two PRs merge in either order.

### It inherits #7192's campaign-service version collision

Same file, same hash (`753c93da6a47`), same problem: this branch forked where campaign-service `openapi.yaml` was `1.42.0` and bumped to `1.43.0`, while #7573 landed #5962 on `main` and took `1.43.0` for the `CONVERTED` enum.

```
git log --oneline -S'version: 1.43.0' origin/main -- openbank-campaign-service/src/main/resources/openapi.yaml
# 35c82f92b fix(campaign): document CONVERTED in SendRecord.outcome (#5962) (#7573)
```

Identical text, so the merge is exit 0 and CI here predates #7573 — invisible from both. Needs a re-bump to `1.44.0` read off live `main`. Fixing it on #7192 does not fix it here; whichever merges first must carry the corrected number.

JiRaska:
**Measured: duplicate ClickHouse migration version. This PR is green and clean, and merging it as-is gives two `V11`s.**

This PR adds:
```
openbank-analytics-sink/src/main/resources/clickhouse/V11__referral_funnel.sql
```
`origin/main` (`80c80fb30`) already carries:
```
openbank-analytics-sink/src/main/resources/clickhouse/V11__credit_funnel.sql
```
Different filenames, so git reports **no conflict** — the merge is textually clean and `mergeStateStatus`
is `CLEAN`. The version is the filename prefix, so the result is two migrations claiming version 11.

Worth remembering what that hides: the module then fails to start, and every `@QuarkusTest` in it reports
as **SKIPPED**, not FAILED — the module reads as a pass because the number anyone looks at is the failure
count.

Fix (renumber; safe while the migration has never been applied):
```
git mv openbank-analytics-sink/src/main/resources/clickhouse/V11__referral_funnel.sql \
       openbank-analytics-sink/src/main/resources/clickhouse/V12__referral_funnel.sql
git add openbank-analytics-sink/src/main/resources/clickhouse/V12__referral_funnel.sql
```
(`git mv` plus a later edit produces a pure-rename commit — re-`git add` after any content change.)
Re-check for a free version at merge time; `main` is at V11 as of `80c80fb30`.

Note: an earlier triage note grouped this PR with #7192 as "campaign, same openapi file". That is not
correct — this PR touches no `openapi.yaml`.

JiRaska:
## Merge-blocker: this adds a second `V11` migration, and git cannot see it

```
$ git ls-tree -r --name-only origin/main | grep 'clickhouse/V11'
openbank-analytics-sink/src/main/resources/clickhouse/V11__credit_funnel.sql

$ git ls-tree -r --name-only origin/<this-branch> | grep 'clickhouse/V11'
openbank-analytics-sink/src/main/resources/clickhouse/V11__referral_funnel.sql

$ gh pr view 7565 --json mergeable --jq .mergeable
MERGEABLE
```

The version is the **filename prefix**, so these are two different files. Nothing conflicts, the merge is clean, and `mergeStateStatus` says CLEAN — you would land two `V11`s and only find out later.

This is the repo's own documented scar, one system over: *"Two branches can claim the SAME migration version and git will never say so."* It happened with Flyway (`V10__delegated_action_index.sql` against `V10__hash_version.sql`) and the same mechanic applies here.

**Fix: renumber to `V12__referral_funnel.sql`.** Safe only while the migration has never been applied — after that the checksum rule forbids touching it.

## What the Flyway version of this hid, and why it is worth checking here

When that collision landed, the service refused to boot — and because it could not start, **every `@QuarkusTest` in the module reported SKIPPED, not FAILED**. The module read `73 tests / 10 skipped / 0 failures`, which scans as a pass, because the number anyone looks at is the failure count. Ten integration tests had silently stopped running.

So after renumbering, read the **skipped count** on the analytics-sink module, not just failures.

## Two related notes from the same sweep

**The ClickHouse path has no migration runner at all** (#7645, PR #7705). The `clickhouse-init` ConfigMap is mounted at `/docker-entrypoint-initdb.d`, which executes **only on an empty data dir**; nothing applies `clickhouse/V*.sql` to a running cluster. So a committed migration here is not self-applying, and the parity gate that guards the ConfigMap was until recently comparing **object sets** — which is why `V9__synthetic_provenance.sql`, an ALTER plus four view re-cuts, moved nothing and stayed green.

**A correction to something said about this PR earlier:** it was reported as sharing an `openapi.yaml` version collision with #7192. It does not — **this PR touches no `openapi.yaml` at all**. The campaign spec collision is #7192's alone.

Nothing else about this PR is blocked: it is green and clean, and it carries #7192's referral outbox slice byte-for-byte, so it can land without #7192 first.

PR #7737, opened 2026-08-31
docs(governance): correct three false reasons in gate baseline exemptions

JiRaska:
## What

A census of every baseline / allowlist / `KNOWN_*` / `NOT_PROBED` / `DEBT` structure across `.github/scripts/*.py`, `.github/gates/gates.yaml` and `openbank-libs/governance/rules.yaml`, classifying each entry's **stated reason** as true, false, unfalsifiable, or absent.

**Three entries carry a reason that is false.** All three reasons are corrected. **No exemption is changed** — a false reason does not mean the exemption is wrong, and what is exempted is the owner's call.

## Why this class exists

The baselines here are built to fail when an entry becomes **stale** — covered debt is reported, so a declaration cannot outlive its subject. That catches an entry that *became* true. It cannot catch one that was **never** true, and (finding 4 below) it does not fire at all where a gate simply never implemented the staleness half. Such an entry stops being tracked debt and becomes an unexamined exemption: the gate is green, the work reads as blocked on infrastructure, and nobody re-checks because the reason sounds like a hard fact.

## Finding 1 — `check-audit-money-path-subscription.py`

```python
"openbank-billing-service": "#6035 - also has no KafkaTopic CR (see the issue)",
```
repeated in the module comment as *"billing has no KafkaTopic CR for `openbank.billing.fee.event` at all"*. The CR exists.

**Check 1** — `grep -n -B2 -A8 'openbank.billing.fee.event' openbank-infra/gitops/components/kafka/kafka.yaml` → `200:kind: KafkaTopic`, `202:  name: openbank.billing.fee.event`
**Check 2** — a YAML parse of every gitops document, enumerating all **90** `KafkaTopic` CRs, finds it — **and finds CRs for both other gap topics**, so billing was not special in the way the reason claimed.

The topic is also referenced from `components/billing/kafka-billing-mtls.yaml` and `apps/billing.yaml`, which is why "a per-service-directory probe missed it" does not explain it. Why the reason was written is not known and no mechanism is invented here. The real gap is the same shape as the other two entries: no Read grant in `components/audit` (its 32 granted topic names include none of the three) and no subscription.

## Finding 2 — `check-external-feeds.py`

```python
("https://go.temporal.io", "Go module path in a Temporal chart value; not an HTTP fetch"),
```
It is a Helm chart repository, fetched over HTTP by Argo CD at sync time.

**Check 1** — repo-wide grep: exactly one occurrence besides the baseline line itself, `temporal-helmrelease.yaml:21  repoURL: https://go.temporal.io/helm-charts`
**Check 2** — YAML parse: resolves as `spec.source.repoURL` on `kind: Application`, `chart: temporal`, `targetRevision: 1.2.0`

The exemption is right — it is the same thing as the thirteen entries reading *"Helm chart repository; deploy-time"*, and a failure blocks the sync loudly. Reason corrected to their wording.

## Finding 3 — `rules.yaml: event_consumer_liveness.allowlist`, topic `proposal-events`

This reason enumerates its own evidence, which is what makes it checkable:

> Nothing is published today and **three facts** say so: the topic is declared nowhere under `openbank-infra/`, the service is absent from `auto-deploy.yml`'s `ALL_SERVICES`, and no KafkaUser exists for it.

All three are false. Each verified twice — grep, and a YAML parse of every gitops document (90 `KafkaTopic` + 43 `KafkaUser` CRs):

| claimed | actual |
|---|---|
| topic declared nowhere under `openbank-infra/` | `components/case-coordinator/kafka-topic.yaml:12` — `kind: KafkaTopic`, `name: proposal-events` |
| service absent from `ALL_SERVICES` | `auto-deploy.yml:225` contains `openbank-case-coordinator-agent` |
| no KafkaUser exists for it | `components/case-coordinator/kafka-case-coordinator-mtls.yaml:12` — `kind: KafkaUser`, `name: case-coordinator` |

The **exemption still holds** on the only thing this gate measures — no `@Incoming` consumer for the topic anywhere in the fleet. What no longer holds is the deployment argument the reason leaned on. Reason corrected, entry kept.

## Finding 4 — not fixed here, needs an owner decision

**`check-event-consumer-liveness.py` has no staleness check, and three of its seven allowlist entries are dead.** The gate computes

```python
allowlisted_hit = sorted(t for t in all_producers if t not in all_consumers and t in allowlist)
```

and reports only that set. An entry whose topic *gains* a consumer falls out of `allowlisted_hit` and is reported **nowhere** — the complement is never computed. Measured using the gate's own scanners (61 application.yaml files, 43 producer topics, 79 consumer topics), and confirmed independently by grep:

| entry | reason says | actual |
|---|---|---|
| `openbank.billing.fee.event` | *"the consumer exists but isn't on main yet: #4122 (open) … remove once #4122 merges"* | #4122 **merged 2026-08-09**; `AnnualFeeSummaryReadyConsumer.kt` is on main; document-service consumes it |
| `openbank.delegation.events` | *"remove this entry with the first projection consumer PR"* | consumed by account-service, audit-service, card-issuance-service, notification-service |
| `openbank.incentive.events` | *"remove this entry with the first real consumer PR"* | consumed by campaign-service |

Each entry names its own removal condition, each condition is met, and nothing noticed. This is the one place the census found a **mechanisable, recurring** form — three instances of "an allowlist entry that is no longer hit" — so it is worth a gate, and the gate is nearly free because `allowlisted_hit`'s complement is one line.

**Not done in this PR** because making it fail requires deleting those three entries, and changing what is exempted is the owner's call. The check and the deletions belong in one PR so it lands green. Sibling gates that already do this: `check-kafka-dotted-keys`, `check-outbox-has-writer` (`0 stale baseline entries`).

## What is NOT changed

Every exemption set is byte-identical, proven by comparing them before and after:

- `_GAP_ISSUE` / `_GAP_TOPICS` keys — `IDENTICAL_KEYS = True`
- `NOT_PROBED` URLs — 46 before, 46 after, `URLS_IDENTICAL = True`
- `event_consumer_liveness.allowlist` topics — 7 before, 7 after, `TOPICS_IDENTICAL = True`

`gates.yaml` is unchanged at **188** entries; no gate is added, so no `rationale:` / `review_after:` / `min_subjects:` registration is required. `event_consumer_liveness` is not in the OPA read set (`rules-opa-data.yaml` holds 5 top-level keys, none of them this one), so no bundle or pod-roll annotation restamps — confirmed by a clean `git status` over `gitops/`.

## Verification

```
$ python3 .github/scripts/check-audit-money-path-subscription.py; echo $?
... 21 produced topic(s) across 23 money-path service(s), 9 known gap(s) — clean.
0
$ python3 .github/scripts/check-external-feeds.py; echo $?
drift: OK — 1 declared feed(s), every external URL accounted for.
0
$ python3 .github/scripts/check-external-feeds.py --self-test
self-test: 20 passed, 0 failed
$ python3 .github/scripts/check-event-consumer-liveness.py --root . --enforce; echo $?
... 0 unallowlisted violation(s), 4 allowlisted producer-only topic(s).
0
```

## Census result

**35 reason-bearing entries** in `.github/scripts/*.py` plus the `rules.yaml` allowlists: **3 false**, the rest true, **0 unfalsifiable**. Entries confirmed true include all 11 `check-kafka-dotted-keys` reasons (each channel does declare `auto.offset.reset: earliest`), all 3 `check-network-policy-code-edges`, all 3 `check-outbox-has-writer`, all 6 `check-incoming-dlq-wiring`, both `check-scheduler-liveness` allowlist reasons, `check-readiness-attestations`' runbook-0003 claim, and in `check-external-feeds` the 13 Helm-repo, 6 mirror and 6 own-ingress reasons — each checked against the artifact it names.

Refs #6035, #4181


JiRaska:
Closing as duplicate: main already corrected these same three false facts (#7740) with equivalent wording. Nothing left for this PR to add.

PR #7736, opened 2026-08-31
feat(admin-ui): unify the three-month FX rate trend

JiRaska:
## Summary
- return real CNB reference history for a three-calendar-month UTC window
- support inverse historical pairs using the same bid/ask inversion as current quotes
- render one validated, chronological and accessible trend in the admin FX workspace
- label the series explicitly as indicative CNB mid-rate, not a binding customer quote

## Verification
- `:openbank-fx-service:test --tests *FxServiceTest`
- `:openbank-customer-edge:test --tests *CustomerEdgeResourceTest`
- `:openbank-fx-service:ktlintCheck :openbank-customer-edge:ktlintCheck`
- focused admin Vitest: 2 passed
- targeted ESLint: clean
- `git diff --check`

Full admin `tsc --noEmit` remains blocked by pre-existing missing generated modules and unrelated Product Studio type errors.

Closes #7735

PR #7744, opened 2026-08-31
fix(admin-ui): unify the three-month FX rate trend on a real CNB series

### LINKED ISSUE #7735: Fix and unify the three-month FX rate trend
The admin portal currently builds FX history from browser-memory snapshots, while customer-edge filters out CNB history and the app treats row limits as calendar days. Provide one truthful three-calendar-month CNB reference-mid trend, support inverse pairs, validate/order history consistently, and clearly label it as indicative. Acceptance: admin and customer app render consistent chronological data for the same pair; inverse pairs work; invalid/duplicate observations are excluded; tests cover date-window and normalization semantics.

JiRaska:
## Summary

Closes #7735. The three-month FX rate trend was two different bugs in two services, which the
issue correctly asked to unify rather than patch independently.

**Bug 1 — customer-edge filtered out exactly the data the trend needs.** `fxRateHistory` fixed
`source=INTERNAL` on the upstream call, so ČNB reference-mid rows never reached the app; the old
`mapFxRateList` (designed for the rate-sheet endpoint) partitions any CNB rows it does see into an
internal lookup map and never emits them as output rows either way.

**Bug 2 — "row limit" was silently treated as "calendar days".** The default `limit=90` (rows) was
documented as "90 days", but ČNB does not publish on weekends or Czech public holidays, so a
90-row window covers well over 90 calendar days — the trend's window was never what the API docs
or the UI implied.

**Bug 3 — the admin portal never called any of this.** Its FX page built a "History" panel purely
from React state: every refresh re-labelled the *current* rate sheet with `now` as the timestamp
and pushed it into a capped in-memory array — a client-side illusion of a time series, never
persisted, never sourced from any real history endpoint. It could not possibly match what the
customer app shows for the same pair.

## Fix

**openbank-customer-edge** (`CustomerEdgeResource.kt`): `fxRateHistory` now fixes `source=CNB`
(the reference-mid), defaults the window to *3 calendar months ago -> now by date* (never by row
count; explicit `from`/`to` still override it), and maps the result with a new `mapCnbTrend`
function: validates each row (usable rate, parseable timestamp), dedupes by calendar date (latest
observation wins), sorts ascending, and falls back to the inverse pair (inverting every rate) when
the direct pair has no history. Response is now `{indicative: true, base, quote, points: [...]}`
instead of a bare array — a genuine breaking change, so `openapi.yaml` bumps `info.version` to
`1.47.0` (URL major stays `/customer/v1` per ADR-0048 D2) with a new `FxRateTrend` schema.

**openbank-admin-ui**: added `/api/fx/history` (BFF route) and `src/lib/fx/trend.ts`
(`buildCnbTrend`/`defaultTrendWindow`), which deliberately mirror the Kotlin implementation's
semantics — same date-window rule, same dedup/order/inverse-pair rules — so the two surfaces are
consistent by construction, not by coincidence. The FX page now renders a real, explicitly
"indicative" 3-Month CNB Trend panel fed by this endpoint; the old fabricated snapshot panel is
renamed to what it actually is (an operator action log for margin/override edits) and no longer
invents rate rows.

## Tests

- `CustomerEdgeResourceTest.kt`: window defaults (3 calendar months by date), `from`/`to`
  overrides + validation, `source=CNB` always fixed, limit capping, inverse-pair fallback with
  rate inversion, and `mapCnbTrend` unit tests (chronological order, same-day dedup keeping the
  latest, exclusion of rows with no usable rate/timestamp/non-positive rate, exclusion of
  mismatched pairs, inverted-mode semantics, non-array body).
- `src/test/fx-trend.test.ts` (admin-ui, Vitest): the same set of cases against `buildCnbTrend`/
  `defaultTrendWindow`.

## Test plan

- [x] `./gradlew :openbank-customer-edge:compileKotlin :openbank-customer-edge:compileTestKotlin`
- [x] `./gradlew :openbank-customer-edge:test --tests CustomerEdgeResourceTest` (all green)
- [x] `./gradlew :openbank-customer-edge:ktlintCheck :openbank-customer-edge:detekt`
- [x] `cd openbank-admin-ui && npm run type-check`
- [x] `npm test -- fx-trend fx- i18n.guard graceful-states.guard layout-shell.guard` (all green)
- [x] `npx eslint` on the changed/added admin-ui files (no new errors)


JiRaska:
CI is red on `gates (api)`: the FX history endpoint's response shape genuinely changed (array → wrapped `FxRateTrend` object, newest-first → oldest-first, `source=INTERNAL` → `source=CNB`, `offset` param removed) — oasdiff correctly classifies this as breaking, and ADR-0048 D5 requires a MAJOR version bump, which in this repo means a coordinated change across three places: `info.version` in `openapi.yaml`, the served URL (`/customer/v1` → `/customer/v2`), and `application.yaml`'s `openbank.api.version` / the Kotlin `@Path` annotations.

I'm stopping here rather than making that call unilaterally — it's a real design decision, not a mechanical fix:

**Option A — full v2 bump.** Correct per the gate's own contract, but means the customer app's existing `/customer/v1/...` client code needs a coordinated update to call `/customer/v2/fx/...`, and the old `/v1` history endpoint either keeps serving the old (broken, per this issue) shape or gets removed. Bigger rollout.

**Option B — new endpoint instead of mutating the existing one.** Add the corrected trend as a NEW path (e.g. `/fx/{base}/{quote}/trend`) alongside the untouched legacy `/fx/{base}/{quote}/history`, so nothing breaks and no version bump is needed. Leaves the old, buggy endpoint in place (possibly deprecated in docs) rather than fixed — only right if nothing depends on the old shape being *correct*, just present.

**Option C — check if `getFxRateHistory` has any real caller yet.** If this endpoint was added recently and nothing in the shipped customer app actually calls it, changing it in place with only an editorial/minor bump might be legitimate — but that's a judgment call about client state I can't verify from here.

My read: Option A is the honest fix given the issue explicitly asks to "unify" behavior across both apps, but it's your call given the coordinated-rollout cost. Let me know which direction and I'll implement it (or an agent will).

JiRaska:
**Measured: superseded by merged #7736 — close candidate.**

#7736 (`feat(admin-ui): unify the three-month FX rate trend`, same issue #7735) merged 2026-08-31 21:03Z. Both PRs create the same two files, which is why the conflict is `add/add`:

```
git merge-tree --write-tree --name-only origin/main origin/fix/fx-rate-trend-unification-7735 | grep CONFLICT
#  add/add  openbank-admin-ui/src/lib/fx/trend.ts
#  add/add  openbank-admin-ui/src/test/fx-trend.test.ts
#  content  openbank-customer-edge/.../CustomerEdgeResource.kt
#  content  openbank-customer-edge/src/main/resources/openapi.yaml
git show origin/main:openbank-admin-ui/src/lib/fx/trend.ts | head -3   # normaliseFxTrend is live
```

One difference worth a look before closing: this PR puts the BFF route at `api/fx/history/route.ts`, whereas `main` uses `api/fx/history/[base]/[quote]/route.ts`. If the flat route was deliberate, that is the only piece not already covered.

JiRaska:
### Rebased onto main, and the customer-edge half is dropped because main already has it

This branch had been open since 31 August and was 116 commits behind, conflicting in four files.
Resolved, and the resolution is mostly subtraction rather than conflict-fixing.

**The backend half of this PR is already on `main`.** `/fx/rates/{base}/{quote}/history`,
`fxRateHistory` and the OpenAPI description landed independently while this branch was open. Merging
this branch's version would have produced a second implementation of one endpoint, so
`CustomerEdgeResource.kt`, `openapi.yaml` and `CustomerEdgeResourceTest.kt` are now **byte-identical
to `origin/main`** — verified by hash, not by eye. The `1.47.0` spec bump is gone with them: this PR
no longer changes the API, so it should not claim a version.

**What is left is the admin-ui half, which main does not have** — four files: the
`/api/fx/history` route, the fx page, `lib/fx/trend.ts` and its test.

**The `trend.ts` collision was real and is resolved as a union, not a choice.** Both sides created
that module independently (`add/add` — it did not exist at the merge base), and both exported
`FxTrendPoint` with incompatible shapes:

```ts
main         export interface FxTrendPoint { timestamp: string; rate: number }   // chart maths
this branch  export interface FxTrendPoint { date: string; rate: string; timestamp: string }  // wire
```

Taking either side would have broken the other's consumers — `FxTrendChart` reads `rate` as a number.
This branch's type is renamed to `CnbTrendPoint`; the rename is contained to the module, since the
route and tests import only the functions. The string stays a string on the wire half: fx-service
sends a decimal, and turning it into a JS number at the edge would drop that guarantee for nothing.

`normaliseFxTrend` and `buildCnbTrend` do overlap on the dedup-by-UTC-day rule. Deliberately **not**
folded together here — different routes, different upstreams, and consolidating them deserves its own
review rather than riding along inside a conflict resolution.

### A defect this rebase surfaced, now fixed on the branch

`fetchHistory` called fx-service with **no `Authorization` header**, and fx-service guards that route:

```kotlin
@Path("/rates/{base}/{quote}/history")
@RolesAllowed("ROLE_VIEWER", "ROLE_OPERATOR", "ROLE_ADMIN", "ROLE_PAYMENTS")
```

So in a deployed cluster every request would have answered 401 and this endpoint would have returned
`upstream_error` for all of them — the trend would never have rendered once. Same class as the
`/api/fx/rates` defect fixed by #7945, and the fix mirrors it: relay the operator's own bearer so
fx-service's policy stays authoritative.

The unauthenticated case now answers **401 with a stated reason**, not the 502 it would otherwise
share with a genuine upstream failure — those are different problems with different fixes, and
collapsing them makes the first one unreportable.

**The new guard test was falsified before being trusted**, since a green guard proves nothing:

| protection removed | result |
|---|---|
| the `Authorization` header | `1 failed \| 3 passed` |
| the 401, folded back into the 502 | `1 failed \| 3 passed` |
| nothing (restored) | `4 passed` |

It also carries the control that makes the rest mean something: a successful but empty history is a
200 with no points and no error, so `error` cannot degenerate into tracking `points.length === 0`.

### One thing found in main's implementation, reported rather than fixed here

This branch's dropped tests asserted `fxRateHistory returns 400 when from is after to`. Main's
implementation validates that each instant *parses*, but never that the window is ordered:

```kotlin
if (from != null && !isValidInstant(from)) return badRequest("Invalid 'from' instant: $from")
if (to != null && !isValidInstant(to)) return badRequest("Invalid 'to' instant: $to")
```

An inverted window is passed straight through to fx-service. That is a money-path service and a
separate change, so it is not in this PR — flagging it so the behaviour is not lost along with the
tests that asserted it.

Local verification: `tsc --noEmit` clean, `eslint` clean, full `npm test` run below.

```
Test Files  401 passed (401)
     Tests  2049 passed | 1 skipped (2050)
```

JiRaska:
Retitled from `fix(customer-edge,admin-ui)` to `fix(admin-ui)`. The scope selects the released
component, and this PR no longer touches `openbank-customer-edge` at all — its three files there are
byte-identical to `origin/main` after the rebase above. Left as-is, release-please would have been
pointed at a component this change does not modify.

JiRaska:
### Audited the rest of the BFF routes for the same defect — it does not appear a third time

Since the missing-bearer defect turned up twice in two days (#7945 on `/api/fx/rates`, and this PR on
`/api/fx/history`), I swept every admin-ui API route on `main` rather than waiting for a third.

114 route files, 54 of which call an upstream `/api/v*` path. **12 carry no `Authorization` or
`accessToken` reference at all**, and every one of those is legitimate:

| route(s) | upstream | why it needs no bearer |
|---|---|---|
| `finops/*`, `devops/dora`, `temporal/status`, `infra/lifecycle` | Prometheus `/api/v1/query`, `/api/v1/status/buildinfo` | cluster-internal observability, not an openbank service |
| `finops/anomalies` | Alertmanager `/api/v2/alerts` | same |
| `bcp/health` | Gatus `/api/v1/endpoints/statuses` | same |
| `test-intelligence` | Prometheus + Tempo `/api/v2/search` | same |
| `services/health`, `catalog/health` | `/api/v1/info` | `ServiceInfoResource` is `@PermitAll` |
| `services/config` | `/api/v1/config` | `ServiceConfigResource` is `@PermitAll` |

One near-miss worth recording: `services/health` looked like the worst case because the file contains
`/api/v1/accounts`, `/api/v1/aml/cases` and `/api/v1/anacredit`. Those are rows in a service metadata
table, not fetch targets — it only ever calls `/api/v1/info` and the health path. Grepping for the
path found strings the route never requests.

**The probe was falsified before the negative was trusted.** Run against `main` as it stood
immediately before #7945 merged, `api/fx/rates/route.ts` reports 0 Authorization references and 2
`/api/v*` calls — exactly the shape it should flag. So "no third instance" is a measurement, not a
silent probe.

Scope limit, stated: this checks whether a route sends a bearer at all, not whether the bearer it
sends is the right one, and it only covers `/api/v*` calls — a route talking to an upstream on some
other path shape is outside it.

Issue #7780, opened 2026-08-31
admin-ui: preserve audit snapshot safely when a repeated lookup fails


JiRaska:
## Problem\nA failed audit lookup clears the last verified trail. Operators lose useful evidence during a transient outage. Naively preserving rows is unsafe because a new Aggregate ID could be shown above an older entity's entries.\n\n## Scope\n- Bind every successful snapshot to the Aggregate ID that produced it.\n- Preserve and clearly label the snapshot only when a repeated lookup for that same ID fails.\n- Never attribute retained entries to a different Aggregate ID.\n- Add authenticated browser coverage for failure and recovery.\n\nNo audit data, payload, BFF routing, authorization, or mutation changes.

PR #7782, opened 2026-08-31
fix(admin-ui): preserve audit lookup evidence

JiRaska:
## Summary\n- bind each successful audit snapshot to the Aggregate ID that produced it\n- preserve and clearly mark evidence only when a repeated lookup for that same ID fails\n- clear older evidence before reporting a failure for a different Aggregate ID\n- give mapped audit row fragments stable React identity\n- add authenticated browser coverage for same-ID recovery and cross-ID isolation\n\n## Verification\n- `git diff --check`\n- `npx eslint src/app/audit/page.tsx e2e/audit-trail-recovery.spec.ts` (clean)\n- `OPENBANK_E2E_PORT=3025 npm run test:e2e -- e2e/audit-trail-recovery.spec.ts --project=chromium` (2 passed, React key warning eliminated)\n\n## Risk\nRead-only audit lookup presentation only. Audit payloads, authorization, BFF routing and all mutations are unchanged.\n\nCloses #7780

### DIFF
--- a/openbank-admin-ui/e2e/audit-trail-recovery.spec.ts
+++ b/openbank-admin-ui/e2e/audit-trail-recovery.spec.ts
@@ -1,0 +1,73 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) OpenBank contributors. Licensed under the Apache License, Version 2.0.
+// See LICENSE in the repository root or https://www.apache.org/licenses/LICENSE-2.0 for details.
+
+import { expect, test } from '@playwright/test'
+import { signInAsOperator } from './helpers/auth'
+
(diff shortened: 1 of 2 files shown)


JiRaska:
Review handoff for exact head `495794042`: read-only audit lookup presentation only. Successful snapshots are now bound to their source Aggregate ID; a failed same-ID retry preserves explicitly stale evidence, while a failed different-ID lookup clears older rows so evidence cannot be misattributed. Stable React fragment keys were also verified. Authenticated Playwright: 2/2 passed; focused ESLint and git diff --check clean. Audit payloads, authorization, BFF routing, and mutations are unchanged. No blocker found; independent approval is still required.

Issue #7783, opened 2026-08-31
admin-ui: align PID navigation and route guard with backend RBAC


JiRaska:
## Problem
The admin UI maps `/pid` and its sidebar entry to `payments:view`, exposing the PII workspace to Viewer, Payments, and Supervisor personas. PID service list/detail/create endpoints accept only ROLE_OPERATOR/ROLE_ADMIN, so those users receive a misleading navigation target and backend 403. The page itself also uses an authentication-only guard.

## Scope
- Add a dedicated `pid:view` permission matching PID backend ROLE_OPERATOR/ROLE_ADMIN.
- Use it consistently in sidebar, edge route projection, and page-level defense in depth.
- Add guard coverage and authenticated negative browser coverage.

No backend roles, PID data, mutation payloads, or service authorization changes.

JiRaska:
## These are one mechanism, and three separate fixes would leave most of it in place

#7783 (`/pid`), #7788 (`/interest`), #7790 (`/swift` + `/clearing`) and the RBAC half of #7824 are not four route bugs. They are four symptoms of **one coarse permission bucket**. On `origin/main` (`cd1f70453`), `src/lib/auth/roles.ts` maps twelve unrelated routes to a single `payments:view`:

```
$ git show origin/main:openbank-admin-ui/src/lib/auth/roles.ts | sed -n '234,237p'
  ['payments:view', [
    '/payments', '/product-catalog', '/standing-orders', '/sdd', '/sepa-instant', '/clearing',
    '/fx', '/swift', '/interest', '/pid', '/fees', '/lending',
  ]],
```

and grants it to five personas including Supervisor:

```
$ git show origin/main:openbank-admin-ui/src/lib/auth/roles.ts | sed -n '51p'
  "payments:view":        [ROLES.ADMIN, ROLES.OPERATOR, ROLES.VIEWER, ROLES.PAYMENTS, ROLES.SUPERVISOR],
```

The backends were checked independently, by reading their `@RolesAllowed` rather than the issue bodies:

```
$ git show origin/main:openbank-interest-service/src/main/kotlin/com/openbank/interest/infrastructure/rest/InterestResource.kt | sed -n '43p'
@RolesAllowed("ROLE_VIEWER", "ROLE_OPERATOR", "ROLE_ADMIN", "ROLE_API")

[...]

PR #7785, opened 2026-08-31
fix(admin-ui): align PID console RBAC

JiRaska:
## Summary\n- add a dedicated PID read permission matching pid-service ROLE_OPERATOR/ROLE_ADMIN\n- use it consistently in sidebar visibility, edge route projection and page defense in depth\n- extend the test-only auth helper to support explicit role sets\n- add matrix/route guard tests and a negative authenticated browser test\n\n## Verification\n- `git diff --check`\n- focused ESLint: 0 errors (one pre-existing PID effect warning)\n- `npx vitest run src/test/pid-rbac-alignment.guard.test.ts src/test/route-permissions.test.ts` (4 passed)\n- `OPENBANK_E2E_PORT=3026 npm run test:e2e -- e2e/pid-rbac.spec.ts --project=chromium` (1 passed)\n\n## Security\nA ROLE_VIEWER session is redirected to the explanatory forbidden page before the PID PII workspace renders. No backend roles, data, mutation payloads or service authorization changed.\n\nCloses #7783

### DIFF
--- a/openbank-admin-ui/e2e/helpers/auth.ts
+++ b/openbank-admin-ui/e2e/helpers/auth.ts
@@ -27,8 +27,8 @@ const NEXTAUTH_SECRET = process.env.NEXTAUTH_SECRET ?? 'e2e-test-secret'
 // USE_SECURE_COOKIES is false and the cookie has no `__Secure-` prefix.
 const SESSION_COOKIE_NAME = 'authjs.session-token'
 
-/** Signs the given browser context in as an operator with every role, bypassing Keycloak. */
-export async function signInAsOperator(context: BrowserContext, baseURL: string): Promise<void> {
+/** Signs a browser context in with an explicit role set, bypassing Keycloak. */
+export async function signInWithRoles(context: BrowserContext, baseURL: string, roles: string[]): Promise<void> {
   const token = await encode({
     secret: NEXTAUTH_SECRET,
     salt: SESSION_COOKIE_NAME,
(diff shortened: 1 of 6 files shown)


JiRaska:
Security review handoff for exact head `84c25d293`: PID console projection now mirrors pid-service's ROLE_OPERATOR/ROLE_ADMIN boundary at sidebar, edge route, and page guard. ROLE_VIEWER is proven redirected before PII renders. Focused RBAC/route tests: 4/4; authenticated negative Playwright: 1/1; ESLint 0 errors; git diff --check clean. Backend roles, PID data, mutation payloads, and service authorization are unchanged. No blocker found in this review; independent approval remains required.

JiRaska:
This PR conflicts on `openbank-admin-ui/src/lib/auth/roles.ts` with #7785, #7789, #7791 and #7936 —
all four add keys to the same map, and all six pairings conflict there while `Sidebar.tsx`
auto-merges. The role values they share are byte-identical, so the conflicts are positional; the one
real disagreement is `payment-rails:view` (one bucket) versus `clearing:view` + `swift:view` (per
route).

Full measurement and a suggested sequencing: #7936 (comment above).

JiRaska:
## Four open PRs edit `src/lib/auth/roles.ts` — merge order matters here

```
#7785  align PID console RBAC          touches roles.ts
#7789  align Interest console RBAC     touches roles.ts
#7791  align payment rail read roles   touches roles.ts
#7936  split the payments:view bucket  touches roles.ts   <- superset
```

All four add or narrow entries in the **same permission map**. Two sides adding neighbouring entries to one list merge **clean** and keep only one — no conflict, exit 0, nothing printed. That is how `.release-please-manifest.json` went 56 entries → 55 in this repo, caught only by an unrelated gate.

**#7936 is the superset.** It splits the twelve-route `payments:view` bucket into ten per-service permissions, covering PID, Interest and the payment rails together — and adds an enforced gate (`check-admin-ui-rbac-alignment.py`) pinning each UI persona set to the backing service's `@RolesAllowed`, falsified in both directions.

It also covers six routes none of these PRs touch — `/fees`, `/lending`, `/sdd`, `/standing-orders`, `/fx`, `/product-catalog` — and those held the worst cases: `ROLE_SUPERVISOR` was granted on all twelve routes and is admitted by **no** backend in the fleet except one balance endpoint.

## Suggested order

Land **#7936** first, then reduce each of these to whatever remains that it does not cover. Merging any of the three narrow PRs after it will either conflict loudly (fine) or merge cleanly while dropping an entry (not fine).

**After each merge to `roles.ts`, check the entry count** rather than trusting the diff:

```bash
git show origin/main:openbank-admin-ui/src/lib/auth/roles.ts | grep -cE "^\s*\['[a-z:]+', \["
```

If it does not go up by the number of permissions the PR added, an entry was silently dropped.

Not closing or touching this PR — the overlap may be deliberate, and which one you keep is your call. Flagging it because CI cannot see this class of collision.

Issue #7786, opened 2026-08-31
admin-ui: make KYC party filtering explicit and outage-safe


JiRaska:
## Problem
The manual Party UUID field is also the active query state, so typing a UUID triggers a KYC request after every character. A failed request clears the last verified cases. Preserving rows without tracking their source would risk showing one party’s KYC cases under another Party ID.

## Scope
- Separate the editable Party UUID from the applied server filter.
- Query only after explicit Search or PartySearch selection.
- Bind successful snapshots to their applied filter.
- Preserve and label a snapshot only for a failed retry of the same filter; never carry it across parties.
- Add authenticated browser coverage for request count, same-filter recovery, and cross-party isolation.

No KYC decisions, case payloads, BFF routing, party links, or RBAC changes.

PR #7787, opened 2026-08-31
fix(admin-ui): make KYC filtering outage-safe

JiRaska:
## Summary\n- separate the editable Party UUID from the applied KYC server filter\n- issue the request only after explicit Search or PartySearch selection\n- bind each successful snapshot to its applied party filter\n- preserve explicitly stale cases only for a failed retry of the same filter and clear them across parties\n- add authenticated browser coverage for request count, recovery and cross-party isolation\n\n## Verification\n- `git diff --check`\n- focused ESLint: 0 errors (one pre-existing effect warning)\n- `OPENBANK_E2E_PORT=3027 npm run test:e2e -- e2e/kyc-filter-recovery.spec.ts --project=chromium` (1 passed)\n\n## Risk\nRead-only KYC list filtering and recovery only. Decisions, case payloads, BFF routing, party links and RBAC are unchanged.\n\nCloses #7786

### DIFF
--- a/openbank-admin-ui/e2e/kyc-filter-recovery.spec.ts
+++ b/openbank-admin-ui/e2e/kyc-filter-recovery.spec.ts
@@ -1,0 +1,65 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) OpenBank contributors. Licensed under the Apache License, Version 2.0.
+// See LICENSE in the repository root or https://www.apache.org/licenses/LICENSE-2.0 for details.
+
+import { expect, test } from '@playwright/test'
+import { signInAsOperator } from './helpers/auth'
+
+const PARTY_ID = '05a02ef1-381c-40e7-b73f-d6855eead42e'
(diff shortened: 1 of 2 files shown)


JiRaska:
Review handoff for exact head `4cd491223`: read-only KYC list filtering/recovery only. Typing no longer emits per-character requests; Search applies one server filter. Successful snapshots are bound to that filter, same-filter failures retain explicitly stale cases, and cross-party failures clear prior rows. Authenticated Playwright: 1/1 passed; focused ESLint 0 errors; git diff --check clean. Decisions, payloads, BFF routing, party links and RBAC are unchanged. No blocker found; independent approval is still required.

JiRaska:
Follow-up head `d284335fd`: fixed the refresh handler/event type mismatch found by CI. Reverified `npm run type-check` (green), targeted ESLint (0 errors; one pre-existing effect warning), and `git diff --check`. CI has restarted; independent approval is still required.

Issue #7788, opened 2026-08-31
admin-ui: align Interest console with backend read roles


JiRaska:
## Problem
The Interest console is mapped to `payments:view`, so ROLE_PAYMENTS and ROLE_SUPERVISOR receive navigation and route access even though InterestResource accrual reads accept only ROLE_VIEWER/ROLE_OPERATOR/ROLE_ADMIN (plus M2M ROLE_API). They reach a sensitive account-accrual workspace that can only return 403.

## Scope
- Add a dedicated `interest:view` permission matching the human backend roles.
- Use it consistently in sidebar, edge route projection, and page-level guard.
- Add matrix and route guard coverage.

No interest calculations, accrual data, mutations, backend roles, or service authorization changes.

JiRaska:
## These are one mechanism, and three separate fixes would leave most of it in place

#7783 (`/pid`), #7788 (`/interest`), #7790 (`/swift` + `/clearing`) and the RBAC half of #7824 are not four route bugs. They are four symptoms of **one coarse permission bucket**. On `origin/main` (`cd1f70453`), `src/lib/auth/roles.ts` maps twelve unrelated routes to a single `payments:view`:

```
$ git show origin/main:openbank-admin-ui/src/lib/auth/roles.ts | sed -n '234,237p'
  ['payments:view', [
    '/payments', '/product-catalog', '/standing-orders', '/sdd', '/sepa-instant', '/clearing',
    '/fx', '/swift', '/interest', '/pid', '/fees', '/lending',
  ]],
```

and grants it to five personas including Supervisor:

```
$ git show origin/main:openbank-admin-ui/src/lib/auth/roles.ts | sed -n '51p'
  "payments:view":        [ROLES.ADMIN, ROLES.OPERATOR, ROLES.VIEWER, ROLES.PAYMENTS, ROLES.SUPERVISOR],
```

The backends were checked independently, by reading their `@RolesAllowed` rather than the issue bodies:

```
$ git show origin/main:openbank-interest-service/src/main/kotlin/com/openbank/interest/infrastructure/rest/InterestResource.kt | sed -n '43p'
@RolesAllowed("ROLE_VIEWER", "ROLE_OPERATOR", "ROLE_ADMIN", "ROLE_API")

$ git show origin/main:openbank-clearing-service/src/main/kotlin/com/openbank/clearing/infrastructure/rest/ClearingResource.kt | sed -n '83p'
[...]

PR #7789, opened 2026-08-31
fix(admin-ui): align Interest console RBAC

JiRaska:
## Summary\n- add a dedicated Interest read permission matching InterestResource's human roles\n- remove ROLE_PAYMENTS and ROLE_SUPERVISOR access that only led to backend 403\n- use the permission consistently in sidebar, edge route projection and page defense in depth\n- add matrix and route guard coverage\n\n## Verification\n- `git diff --check`\n- focused ESLint (clean)\n- `npx vitest run src/test/interest-rbac-alignment.guard.test.ts src/test/route-permissions.test.ts` (4 passed)\n\n## Security\nThis aligns presentation with existing backend authorization. Interest calculations, accrual data, mutations, backend roles and service authorization are unchanged.\n\nCloses #7788


JiRaska:
Security review handoff for exact head `b7144a500`: Interest console navigation, edge route, and page guard now mirror InterestResource's ROLE_VIEWER/ROLE_OPERATOR/ROLE_ADMIN human read boundary. ROLE_PAYMENTS and ROLE_SUPERVISOR no longer receive a destination that backend rejects. Focused matrix/route tests: 4/4; ESLint and git diff --check clean. Calculations, accrual data, mutations, backend roles, and service authorization are unchanged. No blocker found; independent approval remains required.

JiRaska:
This PR conflicts on `openbank-admin-ui/src/lib/auth/roles.ts` with #7785, #7789, #7791 and #7936 —
all four add keys to the same map, and all six pairings conflict there while `Sidebar.tsx`
auto-merges. The role values they share are byte-identical, so the conflicts are positional; the one
real disagreement is `payment-rails:view` (one bucket) versus `clearing:view` + `swift:view` (per
route).

Full measurement and a suggested sequencing: #7936 (comment above).

JiRaska:
## Four open PRs edit `src/lib/auth/roles.ts` — merge order matters here

```
#7785  align PID console RBAC          touches roles.ts
#7789  align Interest console RBAC     touches roles.ts
#7791  align payment rail read roles   touches roles.ts
#7936  split the payments:view bucket  touches roles.ts   <- superset
```

All four add or narrow entries in the **same permission map**. Two sides adding neighbouring entries to one list merge **clean** and keep only one — no conflict, exit 0, nothing printed. That is how `.release-please-manifest.json` went 56 entries → 55 in this repo, caught only by an unrelated gate.

**#7936 is the superset.** It splits the twelve-route `payments:view` bucket into ten per-service permissions, covering PID, Interest and the payment rails together — and adds an enforced gate (`check-admin-ui-rbac-alignment.py`) pinning each UI persona set to the backing service's `@RolesAllowed`, falsified in both directions.

It also covers six routes none of these PRs touch — `/fees`, `/lending`, `/sdd`, `/standing-orders`, `/fx`, `/product-catalog` — and those held the worst cases: `ROLE_SUPERVISOR` was granted on all twelve routes and is admitted by **no** backend in the fleet except one balance endpoint.

## Suggested order

Land **#7936** first, then reduce each of these to whatever remains that it does not cover. Merging any of the three narrow PRs after it will either conflict loudly (fine) or merge cleanly while dropping an entry (not fine).

**After each merge to `roles.ts`, check the entry count** rather than trusting the diff:

```bash
git show origin/main:openbank-admin-ui/src/lib/auth/roles.ts | grep -cE "^\s*\['[a-z:]+', \["
```

If it does not go up by the number of permissions the PR added, an entry was silently dropped.

Not closing or touching this PR — the overlap may be deliberate, and which one you keep is your call. Flagging it because CI cannot see this class of collision.

JiRaska:
Rebased onto current `main`. The conflict was the same shape in both of these PRs, and it wanted the
**union** rather than either side — each branch and `main` pull a *different* route out of the
`payments:view` bucket:

| pulled out of `payments:view` | by | into |
|---|---|---|
| `/pid` | `main`, via #7785 | `pid:view` |
| `/interest` | #7789 | `interest:view` |
| `/swift`, `/clearing` | #7791 | `payment-rails:view` |

Taking either side of that hunk silently puts one of them back, and the result still compiles and
still passes the guard tests — which is what makes this shape worth resolving deliberately rather
than by picking a side.

Checked the property that matters instead of the diff: every route removed from the bucket appears
exactly once in the route map under its new key, so nothing was orphaned. `tsc --noEmit` clean and
the RBAC guard tests pass on each merged result.

One note on the local check: bare `npx tsc --noEmit` reported 20 errors in `product-studio/page.tsx`,
a file neither PR touches and byte-identical to `main`. They are all missing generated types —
`npm run pretest` bakes them and the count goes to zero. Worth knowing before anyone reports it as a
fleet-wide breakage; it is a local-setup artifact, and CI does not see it.

### On touching these branches

Both are `codex/*` and both had a worktree, so I applied the liveness test rather than the name:

```
#7789  last commit 2026-08-31T20:03  (~39h ago)  clean  0 files touched in 2h
#7791  last commit 2026-08-31T20:06  (~39h ago)  clean  0 files touched in 2h
```

No sign of active work, and the push is a merge commit on top of the existing head — nothing
rewritten. If either is still being worked elsewhere, say so and I will leave it alone.

JiRaska:
Closing as superseded: the change this PR makes — a dedicated `interest:view` permission (ADMIN/OPERATOR/VIEWER, no PAYMENTS/SUPERVISOR) plus the `/interest` route prefix — is already on main with the fuller comment citing #7788. Beyond that the branch is damaged: its two earlier main-merges resolved conflicts by DELETING main-side files (`fuzz-coverage-aggregate.py`, `OnboardingStageFilterIT.kt`), so merging it would silently revert unrelated work. #7791 (merged alongside) carries the complementary payment-rails half.

PR #7793, opened 2026-08-31
fix(admin-ui): preserve transaction search evidence

### LINKED ISSUE #7792: Admin UI: preserve transaction search evidence across transient outages
The transaction ledger clears the last successful result when a refresh or search request fails. Operators lose the evidence they were reviewing and cannot distinguish the applied result snapshot from the failed request. Preserve the last successful snapshot, show the classified outage alongside it, and add focused browser E2E coverage for filters and recovery.

JiRaska:
## Summary
- preserve the last successful transaction result when a later request fails
- render the classified outage inline without removing the evidence under review
- cover the exact IBAN query and successful-to-503 recovery journey in Playwright

## Verification
- npm run type-check
- targeted ESLint
- focused Vitest: 3 passed
- git diff --check

Closes #7792


JiRaska:
Author-side review on signed head b2b16fc27: scope is limited to the read-only transaction search recovery path and its browser regression test. Search parameters, BFF endpoint, RBAC, money data, and pagination behavior are unchanged. No blocker found locally; independent approval and green CI remain required before merge.

PR #7796, opened 2026-08-31
fix(admin-ui): preserve SDD mandate evidence

### LINKED ISSUE #7794: Admin UI: keep SDD mandate evidence visible during refresh outages
The SDD console keeps the last mandate rows in React state but replaces the entire table with an outage panel when a later refresh fails. Operators therefore lose the evidence they were reviewing. Keep the last successful mandate snapshot visible, render the classified outage inline, and cover the successful-to-503 recovery path in Playwright.

JiRaska:
## Summary
- keep the last successful SDD mandate snapshot visible when a later refresh fails
- render the classified outage inline with the retained read-only evidence
- cover initial load, explicit refresh, 503 recovery, and snapshot preservation in Playwright

## Verification
- npm run type-check
- targeted ESLint: 0 errors
- git diff --check

Closes #7794


JiRaska:
Author-side review on signed head 5f82e7de8: scope is limited to read-only SDD rendering and a browser regression test. Mandate queries, status filtering, BFF routing, lifecycle operations, RBAC, and money data are unchanged. No blocker found locally; independent approval and green CI remain required before merge.

PR #7799, opened 2026-08-31
fix(admin-ui): preserve standing-order evidence

### LINKED ISSUE #7797: Admin UI: retain standing-order snapshot and expose explicit retry
The Standing Orders page uses the shared wake-aware resource hook, which retains the last successful data, but the page replaces that snapshot with an outage panel and offers no explicit retry action. Keep the last successful read-only evidence visible, add a clear refresh action wired to the shared reload contract, and cover success-to-outage recovery in Playwright.

JiRaska:
## Summary
- expose an explicit refresh action wired to the shared wake-aware resource hook
- keep the last successful standing-order metrics and table visible during refresh and later outages
- render the classified outage inline with the retained read-only evidence
- cover success, explicit refresh, 503 recovery, and snapshot preservation in Playwright

## Verification
- npm run type-check
- targeted ESLint
- focused Vitest: 3 passed
- git diff --check

Closes #7797

### DIFF
--- a/openbank-admin-ui/e2e/standing-orders-recovery.spec.ts
+++ b/openbank-admin-ui/e2e/standing-orders-recovery.spec.ts
@@ -1,0 +1,54 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) OpenBank contributors. Licensed under the Apache License, Version 2.0.
+// See LICENSE in the repository root or https://www.apache.org/licenses/LICENSE-2.0 for details.
+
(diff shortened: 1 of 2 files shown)


JiRaska:
Author-side review on signed head 712a32a31: scope is limited to the read-only Standing Orders refresh/recovery presentation and its browser regression test. Query endpoint, wake retry contract, order data, lifecycle operations, BFF routing, and RBAC are unchanged. No blocker found locally; independent approval and green CI remain required before merge.

PR #7807, opened 2026-08-31
fix(admin-ui): bind consent recovery snapshot

### LINKED ISSUE #7803: Admin UI: bind consent recovery snapshots to the exact lookup
The Consents console clears a valid result before every repeated lookup. A transient failure therefore removes evidence even when the operator retries the exact same party or grantee. Preserve a snapshot only when lens and identifier are identical, continue clearing it for a different lookup to prevent cross-party attribution, and add browser coverage for the same-query recovery boundary.

JiRaska:
## Summary
- bind a retained consent snapshot to the exact lookup lens and identifier
- preserve evidence when the operator retries the same grantee or party and the service fails
- clear prior evidence immediately for a different lookup to prevent cross-party attribution
- cover both sides of that boundary in Playwright

## Verification
- npm run type-check
- targeted ESLint
- focused Vitest: 2 passed
- git diff --check

Closes #7803

### DIFF
--- a/openbank-admin-ui/e2e/consents-query-recovery.spec.ts
+++ b/openbank-admin-ui/e2e/consents-query-recovery.spec.ts
@@ -1,0 +1,66 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) OpenBank contributors. Licensed under the Apache License, Version 2.0.
+// See LICENSE in the repository root or https://www.apache.org/licenses/LICENSE-2.0 for details.
+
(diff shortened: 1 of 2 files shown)


JiRaska:
Author-side review on signed head f17885596: the retained snapshot is keyed by exact lens and identifier; a different party/grantee clears it before request dispatch, while a same-query retry may retain it through failure. Consent payloads, service endpoints, RBAC, and lifecycle semantics are unchanged. No blocker found locally; independent approval and green CI remain required.

PR #7858, opened 2026-08-31
fix(lending): order pending pack proposals by the entity property, not the column

JiRaska:
## What

`GET /api/v1/lending/compliance-packs/proposals/pending` has answered **500 on every call since it shipped**. `CompliancePackActivationRepository.findByState` built the Panache query `state = ?1 order by proposed_at` — HQL resolves entity **properties**, and the property is `proposedAt`; `proposed_at` is only the `@Column` name.

```
org.hibernate.query.SemanticException: Could not interpret path expression 'proposed_at'
```

This is a plain no-parameter `GET`. The query is parsed when it is built, so the failure needs no rows and no malformed input — reaching the endpoint at all is enough.

## Why nothing caught it

- `CompliancePackActivationServiceTest` mocks the repository, so it cannot tell a valid HQL path from an invalid one.
- `CompliancePackActivationIT` exercised propose, decide and `/active` — but never the pending list.

Same shape as the consent `SuppressionEntity` defect (#5711): a name that exists in only one of the two worlds.

## Proven by what it prevents

The new `CompliancePackActivationIT` case 5 is a real-database `@QuarkusTest` that GETs the endpoint and asserts 200.

| run | result |
|---|---|
| against unmodified `main` | **RED** — `5 tests completed, 1 failed`, cause `SemanticException: Could not interpret path expression 'proposed_at'` |
| with this change | **GREEN** — `tests=5 skipped=0 failures=0 errors=0` |

Skipped count read from the JUnit XML, not the console. `ktlintCheck` + `detekt` exit 0.

## Scope

Behaviour fix only. No API contract change, so no `openapi.yaml` bump. Nothing else in this PR — the other #5913 findings are split per service so each can move independently.

Refs #5913

### DIFF
--- a/openbank-lending-service/src/main/kotlin/com/openbank/lending/application/port/out/CompliancePackActivationRepository.kt
+++ b/openbank-lending-service/src/main/kotlin/com/openbank/lending/application/port/out/CompliancePackActivationRepository.kt
@@ -87,7 +87,7 @@ class JpaCompliancePackActivationRepository :
 
     @WithSession
     override fun findByState(state: ProposalState): Uni<List<CompliancePackActivationEntity>> =
-        list("state = ?1 order by proposed_at", state)
+        list("state = ?1 order by proposedAt", state)
 
     @WithSession
     override fun findActivated(): Uni<List<CompliancePackActivationEntity>> = list(
--- a/openbank-lending-service/src/test/kotlin/com/openbank/lending/integration/CompliancePackActivationIT.kt
+++ b/openbank-lending-service/src/test/kotlin/com/openbank/lending/integration/CompliancePackActivationIT.kt
@@ -210,4 +210,30 @@ class CompliancePackActivationIT {
         assertThat(body).contains("\"pack\":{")
         assertThat(body).contains("\"jurisdiction\":\"CZ\"")
     }
+
+    /**
+     * The pending-proposals list, which had never been called by any test.
+     *
+     * `findByState` ordered by `proposed_at` -- the COLUMN name. HQL resolves entity PROPERTIES, and
+     * the property is `proposedAt`, so Hibernate answered `SemanticException: Could not interpret
+     * path expression 'proposed_at'` and this plain no-parameter GET returned 500 on every call
+     * since it shipped (#5913). The query is parsed when it is built, so the failure needs no rows
+     * and no malformed input: reaching the endpoint at all is the whole test.
+     *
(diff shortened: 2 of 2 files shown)


JiRaska:
**Measured: duplicate of open #5948 — one of the two should close.**

Both make the identical one-line production fix (`order by proposed_at` -> `order by proposedAt`). The only delta between the two branches on the production file is the explanatory comment #5948 carries:

```
git diff origin/fix/lending-5913-pending-hql origin/fix/fuzz-5913-nonnull-params \
  -- '*CompliancePackActivationRepository.kt'
```

**Suggest keeping #5948:** same fix, plus the comment recording the HQL-names-the-entity-property trap, plus a dedicated 95-line `CompliancePackPendingListIT` rather than 26 lines appended to an existing IT. #5948 is 9 days old and 1066 commits behind `main`, so it needs a rebase either way.

JiRaska:
Closing in favour of #5948, which is a strict superset. Measured rather than inferred from the titles.

**The production change is byte-identical.** Diffing the same file on both branches, the only delta is a comment #5948 carries and this PR does not:

```
$ diff <(git show origin/<this-branch>:.../CompliancePackActivationRepository.kt) \
       <(git show origin/<5948-branch>:.../CompliancePackActivationRepository.kt)
89a90,93
>  // `proposedAt`, not `proposed_at`: this is HQL, so the ORDER BY names the ENTITY PROPERTY.
>  // The column name parses fine to the eye and throws `SemanticException: Could not interpret
>  // path expression` at runtime, which is how this endpoint answered 500 on every call from the
>  // day it shipped until the fuzz lane got past authentication (#5913).
```

That comment is the part worth keeping — it explains *why* the wrong form looks right, which is the whole reason the defect survived.

**The test coverage is equivalent.** This PR added one case to the existing `CompliancePackActivationIT` (`5 - the pending proposals list is queryable`); #5948 adds a dedicated `CompliancePackPendingListIT` (`listing pending proposals executes the query and returns the proposed row`). Same behaviour, and #5948's is the more focused home for it.

So closing this loses nothing. Keeping both would put two tests for one defect in the tree, in two files, to drift apart later.

I opened this while working #5913 without finding #5948, which had been open since 2026-08-22 — a duplicate produced by not checking the existing queue first. Recording that rather than quietly closing.

The other two PRs from that #5913 batch, #7860 (ledger) and #7861 (party), are unaffected — different services, no equivalent open PR.

PR #7860, opened 2026-08-31
fix(ledger): answer 400 for a null journal line and an absent body, not 500

JiRaska:
## What

`POST /api/v1/journals` returned **500** for two shapes of malformed input. Both are now **400**.

### 1. A null element in the `lines` array

Jackson's Kotlin module null-checks constructor **parameters**; it does not check the **elements** of a collection. So `"lines": [null]` deserialised into a `List<PostJournalLineRequest>` holding a null, and:

```
java.lang.NullPointerException: Cannot invoke
  "com.openbank.ledger.infrastructure.rest.PostJournalLineRequest.toCommand()" because "it" is null
```

The non-null element type was a compile-time promise nothing kept. `lines` is now declared `List<PostJournalLineRequest?>` — the truth on the wire — which is what makes the `requireNotNull` guard **reachable** rather than dead code.

### 2. An absent request body

`postJournal` is a `suspend fun`, and the Kotlin compiler emits **no** `Intrinsics.checkNotNullParameter` for a suspending function. The injected null therefore did not fail at offset 0; it flowed into the body and died at the first dereference. The parameter is now nullable with an explicit `requireNotNull` — the same shape the fleet rule already mandates for `@QueryParam`/`@HeaderParam` (#3104, #3624), applied one level up to the body entity, which `check-nonnull-jaxrs-params.py` does not cover.

Both now raise `IllegalArgumentException`, which libs-runtime's `CommonExceptionMappers` renders as 400 with the `ApiError` envelope and a traceId. **No service-local mapper is added** (#526).

## Proven by what it prevents

New `LedgerApiIT` case asserting 400 for both shapes.

| run | result |
|---|---|
| against unmodified `main` | **RED** — `12 tests completed, 1 failed`, cause the NPE quoted above |
| with this change | **GREEN** — `tests=12 skipped=0 failures=0` |

Full module gate: `:openbank-ledger-service:test` → **311 tests, 0 failures, 1 skipped**. The single skip is `LedgerPactBrokerProviderVerificationTest`, the known `@EnabledIfSystemProperty(pactbroker.url)` broker-sourced class, identified by name rather than assumed. `ktlintCheck` + `detekt` exit 0.

## Scope

Behaviour fix only, and it only ever converts a 500 into a 400 — every request that succeeded before still succeeds. No API contract change, so no `openapi.yaml` bump. Split per service so each #5913 finding can move independently.

Refs #5913

### DIFF
--- a/docs/threat-models/openbank-ledger-service.md
+++ b/docs/threat-models/openbank-ledger-service.md
@@ -242,6 +242,23 @@ set) apply equally to the new `ledger.approval.decide` action.
 
 ## 8. Change log
 
+- **2026-09-01** — `POST /api/v1/journals` answered **500** rather than 400 for two shapes of
+  malformed input: a `null` element inside the `lines` array, and an absent request body. Neither
+  crossed a trust boundary, moved money or bypassed a control — `@RolesAllowed(OPERATOR)`, the
+  `@Authorize(action = "ledger.create")` OPA gate and `validateBalance()` all run unchanged, and
+  the request failed before reaching the domain in both cases. It is recorded here because the
+  inbound REST surface changed shape: `PostJournalRequest.lines` is now `List<PostJournalLineRequest?>`
+  and the body parameter is nullable, so the `requireNotNull` guards are reachable instead of dead
+  code (Jackson's Kotlin module null-checks constructor parameters, never collection elements; and
+  a `suspend fun` emits no `Intrinsics.checkNotNullParameter`, so the injected null flowed into the
+  body and died at the first dereference). **STRIDE-D:** the 5xx was the availability-signal defect
+  — a caller could not distinguish "I sent a bad request" from "the server is broken", which on a
+  money path decides whether it retries, and it burned SLO error budget on client error. Both now
+  raise `IllegalArgumentException`, mapped to 400 by libs-runtime's `CommonExceptionMappers`; no
+  service-local mapper is added (#526). No new caller, endpoint, role, network edge or privilege.
+  Rollback: revert the two declarations to their non-nullable form, restoring the 500.
+  Refs #5913.
+
 - **2026-08-26** — `POST /api/v1/journals` now copies the trusted synthetic classification from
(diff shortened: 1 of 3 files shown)


JiRaska:
### #6486 and #7860 fix the same defect in the same file, and conflict

Measured on the current branches. Both change
`LedgerResource.kt`, and a merge of the two conflicts there:

```
CONFLICT (content): Merge conflict in
  openbank-ledger-service/src/main/kotlin/com/openbank/ledger/infrastructure/rest/LedgerResource.kt
```

**The production fix in #7860 strictly contains the one in #6486.** Both make the element type
honest (`List<PostJournalLineRequest?>`) and reject a null element with `requireNotNull`, which
libs-runtime maps to 400 — the same change, #7860 just extracts it into a `requireLines()` member.
#7860 then fixes a second path #6486 does not touch: an **absent body**, where a `suspend fun` emits
no `Intrinsics.checkNotNullParameter`, so the null reaches the body and NPEs as a 500.

**But #6486 carries the contract change and #7860 does not.** #6486 documents the new response and
bumps the API version; #7860 changes the same endpoint's observable behaviour with no `openapi.yaml`
edit at all:

```yaml
-  version: 1.16.0
+  version: 1.17.0
+        '400':
+          description: Malformed body — an unparseable date, or a null element in `lines`
```

A 500 becoming a 400 for the absent-body case is an API change on the same endpoint, so the spec
needs it either way (rule 3).

**So neither is a superset, and neither should simply be closed.** The subtraction that works:

1. Land #7860 — it is the broader and better-documented fix, and it covers everything #6486 does to the resource.
2. Reduce #6486 to its `openapi.yaml` and threat-model edits, widening the `400` description to cover the absent body as well. That drops the conflicting hunk entirely rather than resolving it.

Taking either side of the `LedgerResource.kt` conflict by hand reverts the other's coverage, which is
why this is worth sequencing rather than merging in whichever order they go green.

**One check before #6486 rebases:** its bump to `1.17.0` is still free — live `main` reads
`version: 1.16.0` as of this comment. Re-read it after any competing ledger spec merge; whoever lands
second takes the next number.

PR #7896, opened 2026-09-01
fix(admin-ui): review DevOps decisions

### LINKED ISSUE #7895: Admin UI: add final review for DevOps remediation decisions
## Problem
DevOps-agent remediation findings can be approved or rejected with a single click. The decision is HITL and may authorize operational remediation, but there is no final review of the exact finding and impact.

## Scope
- Add an accessible final review before approve/reject.
- Show finding title/id, detector, severity, status, root cause, remediation kind, DORA impact, and proposal link when present.
- Preserve the existing decision endpoint/payload, RBAC, and agent proposal semantics.
- Keep the dialog open with the error after a failed decision and add focused component/browser regression coverage.

## Out of scope
- No devops-agent API, remediation execution, RBAC, or DORA calculation changes.

JiRaska:
## Summary
- add an accessible final review before DevOps-agent remediation approval or rejection
- show the exact finding, detector, severity, resource, DORA impact, root cause, remediation kind/text, and proposal link
- retain the dialog and error after a failed decision so the operator can safely retry

## Preserved controls
- unchanged `POST /api/devops/decide` payload `{ id, action }`
- unchanged `devops:decide` RBAC and human-in-the-loop semantics
- no devops-agent, remediation execution, or DORA calculation changes

## Verification
- `npm run pretest`
- `npx tsc --noEmit`
- `npx vitest run src/test/devops-decision-review.guard.test.ts` (1/1)
- `OPENBANK_E2E_PORT=3118 npx playwright test e2e/devops-decision-review.spec.ts --project=chromium` (1/1)
- `git diff --check`

Closes #7895

### DIFF
--- a/openbank-admin-ui/e2e/devops-decision-review.spec.ts
+++ b/openbank-admin-ui/e2e/devops-decision-review.spec.ts
@@ -1,0 +1,55 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) OpenBank contributors. Licensed under the Apache License, Version 2.0.
+// See LICENSE in the repository root or https://www.apache.org/licenses/LICENSE-2.0 for details.
+
+import { expect, test } from '@playwright/test'
+import { signInAsOperator } from './helpers/auth'
+
+const finding = {
+  id: 'finding-deploy-health-42', detector: 'D4_DEPLOY_HEALTH', severity: 'CRITICAL',
+  detectedAt: '2026-08-31T20:00:00Z', title: 'Repeated sandbox rollout failure',
+  rawMetricValue: 4, threshold: 2, affectedResource: 'openbank-admin-ui',
+  doraMetricImpacted: 'CHANGE_FAILURE_RATE', rootCause: 'Readiness probe regressed after rollout',
+  remediationKind: 'PULL_REQUEST', proposalPrUrl: 'https://github.com/JiRaska/open-bank-oss/pull/7000',
+  proposedRemediation: 'Restore the previous readiness timeout and add rollout coverage',
+  status: 'PROPOSED', diagnosedAt: '2026-08-31T20:01:00Z', proposedAt: '2026-08-31T20:02:00Z',
+}
+
+test.beforeEach(async ({ context, baseURL }) => {
+  await signInAsOperator(context, baseURL!)
+})
+
+test('operator reviews exact AI remediation, retains failure, and retries unchanged decision', async ({ page }) => {
+  await page.addInitScript(() => window.localStorage.setItem('openbank-admin-lang', 'en'))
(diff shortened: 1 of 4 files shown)


JiRaska:
### The failing test is asserting formatting, not the authority rule — the RBAC guard is intact

`Admin UI build` fails on one assertion in `src/test/devops-hitl-rbac.guard.test.ts`:

```
❯ src/test/devops-hitl-rbac.guard.test.ts:15:18
    expect(page).toContain("onApprove={canDecide ? id => decide(id, 'a…
Test Files  1 failed | 396 passed (397)
```

That test reads `app/devops/page.tsx` as **text** and requires an exact source line. This PR keeps the
guard and only changes the handler body from an expression to a block:

```
main    onApprove={canDecide ? id => decide(id, 'approve') : undefined}
#7896   onApprove={canDecide ? id => {
```

`canDecide` still appears the same number of times as on main (5), `const canDecide =
hasPermission('devops:decide')` is unchanged at line 190, and `<AuthGuard permission="system:view">`
is still there. **No authority regression** — the assertion cannot tell a reformat from a removal,
so it fails on the safe change and would pass on any unsafe one that happened to keep the literal.

Two ways forward, and the second is the one worth taking:

1. Restore the expression form and the test goes green.
2. Make the test assert the property instead of the source text — render the page with a principal
   lacking `devops:decide` and assert the approve/reject controls are absent. A guard that greps its
   own source proves the string, never the behaviour.

Worth noting how narrow the current form is: the same file also requires
`'"system:view":              [ROLES.ADMIN, ROLES.OPERATOR, ROLES.DEMO]'` — column alignment included
— from `lib/auth/roles.ts`. I checked the four open PRs that edit that map (#7785, #7789, #7791,
#7936) and all four happen to preserve the exact spacing, so nothing else is breaking today. It holds
by luck, not by construction.

### Separately: two of this run's failures are already fixed on main

`gates (security)` and `gates (supplychain)` failed here because this run started
`2026-09-01T00:16:21Z`, about seven hours before #7919 merged at `07:14:12Z` and moved the two lapsed
time-gates. A fresh run picks that up; the `Admin UI build` failure above is this PR's own and will
not clear on its own.

JiRaska:
Author exact-head verification on 68944a9673 (not an independent approval): the historical CI failure was a stale source guard that required direct one-click decide() handlers. The guard now requires devops:decide authorization to open the review intent and requires the only decide() call to live behind explicit confirmation. Current-base proof: focused Vitest 3/3, Chromium 1/1 including retained failure and unchanged retry, canonical type-check, changed-file ESLint with zero errors and one pre-existing page warning, production Next build 91/91, diff-check clean. Endpoint, { id, action } payload, RBAC, remediation execution and DORA calculations remain unchanged. Merge still requires terminal green CI and an eligible independent APPROVED review.

JiRaska:
Closing as superseded: the feature — an accessible final review dialog before a DevOps-agent remediation decision — is already on main as `RemediationReviewDialog` (10 references from the DevOps page), landed by a parallel PR. This branch's implementation would now be a second, divergent dialog for the same button.

PR #7899, opened 2026-09-01
feat(admin-ui): review payment orders before submission

### LINKED ISSUE #7898: Admin UI: extract a shared review-dialog primitive, then add final review before payment creation (anchor for #7893/#7895)
## Problem
Domestic and SEPA payment forms submit directly to money-path APIs. Idempotency, RBAC and VoP are present, but the operator has no final review of the exact payment before submission.

## Scope
- Add an accessible final review for domestic standard/instant and SEPA/SCT Inst creation.
- Show payment type, amount/currency, debtor, beneficiary, identifiers, speed, and VoP state where applicable.
- Preserve current endpoints, payloads, idempotency-key reuse, single-flight behavior, RBAC, VoP rules, and downstream approval semantics.
- Keep entered data and review context after request failure; add focused component/browser coverage.

## Out of scope
- No payment-service API, settlement, approval, SCA, VoP, or authorization changes.

JiRaska:
Closes #7898

## Summary
- add an accessible final review step for domestic, SEPA and SCT Inst payment initiation
- show the exact amount, source, beneficiary, routing details and VoP result before any money-path POST
- keep failures in the review dialog for a safe retry and restore keyboard focus when returning to edit
- preserve the existing BFF endpoints, payloads, RBAC, VoP gates, single-flight guard and idempotency-key lifecycle

## Verification
- `npm run pretest`
- `npx tsc --noEmit`
- focused ESLint (0 errors; 3 pre-existing warnings in payments page)
- focused Vitest: 6/6
- Playwright: SEPA + SCT Inst 2/2; domestic 1/1
- `git diff --check`

## Safety
No API, approval workflow, RBAC, VoP policy, payload or idempotency semantics changed. The browser now sends zero payment POSTs until the operator confirms the exact review.

### DIFF
--- a/openbank-admin-ui/e2e/payment-create.spec.ts
+++ b/openbank-admin-ui/e2e/payment-create.spec.ts
@@ -8,7 +8,7 @@ test.beforeEach(async ({ context, baseURL }) => {
   await signInAsOperator(context, baseURL!)
 })
 
-test('creates one SEPA payment with one idempotent BFF request under a double submit', async ({ page }) => {
+test('reviews and creates one SEPA payment with one idempotent BFF request under a double confirmation', async ({ page }) => {
   const posts: { key: string | null; body: Record<string, unknown> }[] = []
 
   await page.route('**/api/sepa-payments', async route => {
@@ -39,10 +39,18 @@ test('creates one SEPA payment with one idempotent BFF request under a double su
   await page.getByLabel('Amount (EUR)').fill('100.00')
 
   const form = page.getByLabel('Debtor IBAN').locator('xpath=ancestor::form')
-  await form.evaluate(element => {
-    const paymentForm = element as HTMLFormElement
-    paymentForm.requestSubmit()
-    paymentForm.requestSubmit()
+  await form.evaluate(element => (element as HTMLFormElement).requestSubmit())
+
+  const review = page.getByRole('alertdialog', { name: 'Review payment order' })
+  await expect(review).toBeVisible()
+  await expect(review).toContainText('100.00 EUR')
+  await expect(review).toContainText('Jane Doe')
+  await expect(review).toContainText('DE89370400440532013000')
+  expect(posts).toHaveLength(0)
+
+  await page.getByRole('button', { name: 'Confirm and submit' }).evaluate((button: HTMLButtonElement) => {
+    button.click()
+    button.click()
   })
 
   await expect(page.getByText('SEPA payment created')).toBeVisible()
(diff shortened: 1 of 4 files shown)


JiRaska:
### The Gitleaks failure is a false positive, and it is one literal, not three

All three findings are the same synthetic account number in `e2e/payment-create.spec.ts` — lines 88,
98 and 108 — flagged by the `rod-cislo` rule:

```
Creditor Account No.   9876543210
review contains        9876543210/0800
creditorAccountNumber  9876543210
```

**Why it matches, precisely.** The rule is

```
\b\d{2}(?:0[1-9]|1[0-2]|2[1-9]|3[0-2]|5[1-9]|6[0-2]|7[1-9]|8[0-2])\d{2}/?\d{3,4}\b
```

It validates the encoded MONTH — including the `+50` (women) and `+20` ranges, which is why `76` is
accepted — but the DAY is a bare `\d{2}` with no range check. `9876543210` parses as `98` / `76` /
`54`, and day `54` is a day no birth number can carry. So this string is structurally incapable of
being a rodné číslo, and the rule cannot see that.

Measured both directions, so this is not just an argument about the regex:

| value | current rule | with the day validated |
|---|---|---|
| `9876543210` (this PR) | MATCH | — |
| `9006154321` (real-shaped RC) | MATCH | MATCH |
| `855231/1234` (real-shaped RC, slashed) | MATCH | MATCH |

**The one-line fix for this PR**, which needs no config change: use an account number whose third and
fourth digits are not a valid month encoding. `1234567890`, `4444555566` and `9999888877` are all
clean against the rule as written; `9876543210` is not. Changing the literal in the three places
turns this check green.

**The durable fix, for whoever owns `.gitleaks.toml`:** add the day range to the pattern —

```
(?:0[1-9]|[12]\d|3[01])
```

in place of the `\d{2}` day. Real birth numbers keep matching, as the table shows.

### One side observation, with its limits stated

Several allowlist entries carry comments blaming this rule for their existence, and the rule as
written does not produce those matches. Scanning the whole files against it:

| allowlisted file | `rod-cislo` matches |
|---|---|
| `openbank-infra/docker/grafana/dashboards/openbank-overview.json` | **0** |
| `openbank-infra/gitops/components/kafka/kafka.yaml` | **0** |
| `com/openbank/libs/security/PiiMasking.kt` | 1 — genuine, it holds the pattern by design |

So the Grafana and Kafka comments ("thresholds … match the rod-cislo numeric pattern",
"retention.ms … trips the rod-cislo numeric heuristic") describe something that does not happen. The
entries may still be needed for the `iban` / `pan-card-number` / generic-api-key rules — I did not
test those, and this is not a suggestion to remove them. It is worth knowing that the stated reason
is not the operative one before anyone prunes the list on the strength of those comments.

Not pushed to this branch: it is checked out in a live worktree, so the change is yours to make.

PR #7913, opened 2026-09-01
fix(admin-ui): make quick-search states truthful

### LINKED ISSUE #7912: fix(admin-ui): command palette conflates backend outage with no results, and a superseded request clears the current one
## Problem
The global quick-search palette maps every entity-resolution HTTP/network failure to an empty result list. Operators therefore see No results for both a valid zero-match query and a backend outage. Loading is not exposed through aria-busy/live status, and an aborted older request may clear the loading state of a newer query.

Because the palette is available on every authenticated admin route, this ambiguity affects the whole operator experience.

## Scope
- Distinguish loading, empty and unavailable states.
- Add an explicit retry action without changing the query.
- Make request cleanup/race handling signal-aware so stale requests cannot overwrite current state.
- Expose loading and recovery semantics accessibly.
- Add focused unit/browser regression evidence.

Entity classification, result routes, permissions, query encoding and the BFF endpoint remain unchanged.

JiRaska:
## Summary
- distinguish loading, valid empty results, and temporary entity-search outages
- keep retry on the same query and prevent stale/aborted requests from overwriting current state
- expose busy/live/alert semantics for assistive technology

## Preserved behavior
- entity classification, routes, permissions, query encoding, and the `/api/entities/resolve` BFF endpoint are unchanged

## Verification
- `npx vitest run src/test/command-palette.test.tsx` (11/11)
- `npx playwright test e2e/command-palette.spec.ts --repeat-each=3` (3/3)
- `npx eslint src/components/search/CommandPalette.tsx src/test/command-palette.test.tsx` (no errors; one pre-existing set-state-in-effect warning)
- `npm run build`
- `git diff --check`

Closes #7912


JiRaska:
Author follow-up: added a real-browser journey covering Ctrl+K open, BFF 503, truthful unavailable state, retry without query loss, result selection, and navigation. Playwright passed 3/3 repeated runs. New signed head: ad0e2ed77.

Issue #7943, opened 2026-09-01
Five BFF routes render an upstream failure as an empty collection; the fx one also sends no Authorization header


JiRaska:
## Five BFF routes render a failed upstream call as an empty collection

Measured on `origin/main`, over all **114** `route.ts` files under `openbank-admin-ui/src/app/api/`:

```
finops/resources/route.ts
finops/right-sizing/route.ts
fx/rates/route.ts
temporal/status/route.ts
test-intelligence/route.ts
```

Each contains a `catch { return [] }` or equivalent, so an upstream failure and a genuinely empty result reach the page as the same value. Nothing downstream can tell them apart.

## The confirmed instance is the worse variant

`app/api/fx/rates/route.ts`:

```ts
async function fetchFxServiceRates(baseUrl: string) {
  try {
    const res = await fetch(`${baseUrl}/api/v1/fx/rates`, { signal: AbortSignal.timeout(5000) })
    if (!res.ok) return []
    ...
  } catch {
    return []
  }
}
```

There is **no `headers` and no `Authorization`** on that call — so a 401 is not an edge case here, it is the expected response, and `if (!res.ok) return []` turns it into "no rates". A missing credential, a down service, a timeout and an genuinely empty rate table are four different states rendered identically.

24 other BFF routes do send an `Authorization` header, so this is an omission rather than a house convention.

## Why this is the recurring shape rather than a style nit

It is the same defect this repo keeps finding in different clothes: a not-done outcome encoded as the value a done-and-empty outcome produces. A disabled push adapter returned `success = true`; a fraud verdict was returned as `Unit`; an unconfigured court register was hardcoded `false`; an unscraped database was invisible to its own backup alert. In each case the quiet path had no error to find, because nobody gave the state a name.

For an FX rates panel the cost is bounded — an operator sees an empty table. It is worth fixing because the shape is cheap to correct once and expensive to notice later, not because this particular screen is critical.

## Suggested direction

Distinguish the states rather than flattening them: return a discriminated result (`{ status: 'ok' | 'unauthorized' | 'unavailable', rows }`) and let the page say *"rates unavailable"* rather than showing an empty table that reads as *"no rates today"*. The four other routes deserve the same look, though each may have a different right answer.

Fixing the missing `Authorization` header on the fx call is separable and smaller, and should probably land first.

## A note on how this was counted, because two probes lied first

[...]

PR #7049, opened 2026-08-26
feat(admin-ui): link browser failure diagnostics

JiRaska:
## Summary

Project failed Admin UI browser-run diagnostics into Test Intelligence without copying Playwright screenshots, traces, DOM, video, or requests into canonical evidence. Operators get the exact GitHub run/attempt artifact link, seven-day retention, authenticated-access boundary, and an explicit sensitive-data warning.

This is stacked on #7032 because the shared Test Intelligence schema/projection changes build on its current evidence model.

## Type of change

- [x] feat (new functionality)
- [x] docs

## Linked issues

Closes #7036
Refs #7032

## Changes

- emit typed metadata only when the retained Playwright report exists after a failed browser attempt
- preserve exact run/attempt provenance, GitHub-authenticated access, seven-day retention and privacy warning
- project the diagnostic link into execution and client-experience views without changing the E2E verdict
- extend schema, ADR, governance guard, collector tests and browser contract

## Definition of Done

- [x] Hexagonal layering preserved.
- [x] Unit and browser contracts updated.
- [x] No new lint warnings or dependency.
- [x] ADR and governance updated.

## Security checklist

- [x] No secrets, tokens, passwords, or PII in canonical evidence.
- [x] No new suppression or unsafe cast.
- [x] No new third-party dependency.
- [x] Auth, crypto, payment and cardholder-data paths unchanged.

## Compliance impact

- [x] DORA
- [x] GDPR

The change improves failure diagnosability while retaining GitHub authorization and short report retention. Canonical evidence stores metadata only and warns that the external bundle may contain sensitive browser data.

## Rollout / Rollback

- Deployment strategy: existing rolling Admin UI release and GitOps pipeline after #7032
- Rollback plan: revert this commit; the existing E2E verdict remains independently usable
- Data migration reversible: N/A

## Reviewer notes

Verify that diagnostics remain subordinate to the failed E2E verdict, no report contents enter the run envelope, and attempt provenance cannot be confused across retries.

Validation:
- `python3 .github/scripts/collect-test-run-evidence.py --self-test`
- `python3 .github/scripts/check-test-intelligence-ecosystem.py`
- schema JSON parse
- focused Vitest: 6/6
- Admin UI TypeScript type-check
- `git diff --check`

PR #7988, opened 2026-09-01
fix(admin-ui): report a failed BFF read as a failure, not an empty collection

JiRaska:
## What

Three of the four routes left by #7943 after #7945 fixed `fx/rates`.

Each route probes Prometheus `/-/ready` first and correctly answers `available: false` when that probe fails — that part was never wrong. **The defect is one layer in.** Every individual query then swallowed its own failure into `{}` / `[]` / `null`, so a Prometheus that is **ready** but answering 500, timing out, or returning `status: "error"` rendered as `available: true` with an empty result:

| route | what an outage rendered as |
|---|---|
| `finops/resources` | "the fleet has no services" |
| `finops/right-sizing` | "no namespace has any savings potential" |
| `temporal/status` | `temporalDeployed: false` — which the console renders as *"Temporal is not deployed in this environment"*, an outage reported as an architecture fact |

That last one is the same downstream shape as the `McpEndpoint` 404 in the repo's own notes: a healthy-but-unreachable service rendering as "not deployed".

## How

The shape PR #7945 established on `fx/rates`, and the asymmetry was already there to copy: `available` distinguished reachability at the outer probe, the inner queries just had no equivalent.

Each helper now returns a **discriminated result** instead of a bare collection:

- `error` is `null` when every query answered, and **names the failing queries** otherwise; `degraded` is the boolean form.
- `temporalDeployed` becomes **`null`** rather than `false` whenever the route has no basis for either verdict — both when Prometheus is unreachable and when the `temporal_restarts` query itself failed.
- `vpaHasData` is true only when the VPA query actually **answered**; a failed query has no rows either, and was reported as "VPA has no data".

An **absent metric stays an honest `null`**: no samples is a real answer, not a failure. Keeping that distinct from a failure is the whole point — collapsing them the other way would just move the lie.

## The control, and its falsification

Every pair carries a test asserting that a query which **succeeds and genuinely returns nothing** reports `error: null`. Without it, `error` could simply track `rows.length === 0` and these guards would pass while measuring nothing — the failure mode this repo keeps finding in its own checks.

That was **falsified, not assumed**. A mutant computing `error` from emptiness rather than from a real failure:

```
MUTANT_RC=1 — Tests 1 failed | 6 passed
× reports NO error when the queries succeed and the fleet is genuinely empty
```

It passes the failure test and fails **only** the control. The control can therefore detect the absence of what it tests.

Exit codes, each read from `$?` immediately after redirecting to a file (never through a pipe):

- 7 new tests against unmodified `main` routes: `RED_RC=1` — 7 failed
- with the fix: `GREEN_RC=0` — 7 passed
- mutant: `MUTANT_RC=1` — only the control fails
- `npx tsc --noEmit`: `TSC_RC=0`, 0 errors
- clean full `npm test`: `SUITE_BFF_CLEAN_RC=1` — 2057 passed, **4 pre-existing failures**

Those 4 (`campaigns-console`, `card-transition-dialog`, `console-summary-wiring`, `iaops-case-thread-page`) reproduce identically with this branch's routes reverted to the base commit — same count, unrelated files, all 5000ms render timeouts. They are not caused by this change. Verified twice, including a same-file-same-base control run.

## Left deliberately

**The fourth route, `test-intelligence`.** Its defect site (`queryPrometheusRuns` returning `[]` on a failed read, line ~131) sits **inside** the hunk range open PR #7049 is rewriting (`@@ -119,34 +142,38 @@`), and three live worktrees plus an in-progress local cherry-pick are on that file. Editing it now buys a conflict rather than a fix. It should be picked up once #7049 lands — the shape is already established here and in #7945, and that route already has a `warnings[]` array to carry the reason.

Refs #7943

### DIFF
--- a/openbank-admin-ui/src/app/api/finops/resources/route.ts
+++ b/openbank-admin-ui/src/app/api/finops/resources/route.ts
@@ -11,22 +11,33 @@ function prometheusBase(): string {
   return process.env.PROMETHEUS_URL ?? 'http://localhost:9090'
 }
 
-async function queryRange(query: string, signal: AbortSignal): Promise<Record<string, number>> {
+/** Outcome of one Prometheus read.
+ *
+ *  Deliberately NOT a bare map. This route already distinguishes "Prometheus is unreachable"
+ *  (`available: false`) from a successful read — but only at the `/-/ready` probe. Every
+ *  individual query below then swallowed its own failure into `{}`, so a Prometheus that is ready
+ *  and answering 500 (or timing out, or returning `status: "error"`) rendered as
+ *  `available: true, services: []` — a confident claim that the fleet has no services, which is
+ *  never a true statement about a running cluster. Same asymmetry PR #7945 fixed on fx/rates:
+ *  the honest shape already existed one level up, one layer just lacked it (#7943). */
+type Query = { rows: Record<string, number>; error: string | null }
+
+async function queryRange(query: string, signal: AbortSignal): Promise<Query> {
   try {
     const url = `${prometheusBase()}/api/v1/query?query=${encodeURIComponent(query)}`
     const res = await fetch(url, { signal, headers: { Accept: 'application/json' } })
-    if (!res.ok) return {}
+    if (!res.ok) return { rows: {}, error: `prometheus responded ${res.status}` }
     const json = await res.json() as { status: string; data: { result: { metric: Record<string, string>; value: [number, string] }[] } }
-    if (json.status !== 'success') return {}
+    if (json.status !== 'success') return { rows: {}, error: `prometheus query status ${json.status}` }
     const out: Record<string, number> = {}
     for (const r of json.data.result) {
       const svc = r.metric.container ?? r.metric.application ?? r.metric.job ?? r.metric.service ?? 'unknown'
       const val = parseFloat(r.value[1])
       if (!isNaN(val)) out[svc] = val
     }
-    return out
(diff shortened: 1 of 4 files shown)


JiRaska:
Complementary to the BFF sweep I posted on #7744 an hour ago, not in tension with it — worth saying
because the two verdicts touch the same three routes and read as if they disagree.

That sweep asked one question only: does a route that calls an upstream `/api/v*` send an
`Authorization` header. `finops/resources`, `finops/right-sizing` and `temporal/status` came back
"legitimate" there because they call **Prometheus**, which is cluster-internal and needs no bearer.
That verdict says nothing about what they do with a failed read, and I said so explicitly: *"this
checks whether a route sends a bearer at all, not whether the bearer it sends is the right one"*.

This PR is the adjacent defect on the same routes, and it is the more interesting one: `null` carrying
both "no samples" and "the query failed". Same family as `PushResult.skipped()` returning
`success = true` — a state where nothing was learned encoded as a state where something was, with no
signal anywhere that disagrees. The `temporalDeployed` case is the sharp end: a Prometheus answering
500 renders as the confident claim that Temporal is not deployed, which is an outage reported as an
architecture fact.

The distinction the diff draws is the right one, including the part that is easy to get wrong —
`!json.data.result.length` stays `error: null`, because an empty result set is a real answer. Without
that, `error` would degenerate into tracking `value === null` and the split would buy nothing.

PR #7990, opened 2026-09-01
feat(billing): expose pending approvals in the unified inbox

JiRaska:
## What

Closes the last two non-money-path gaps in the unified approval inbox (ADR-0227 D2): **billing-service** and **consent-service** now serve their pending maker-checker queue, and the admin-UI BFF federates them for real instead of reporting them `not-configured`.

## Re-verified count (read from `origin/main`, not the working copy)

- 16 services carry `infrastructure/rest/ApprovalResource.kt`.
- **13** call `approvalStore.findPending` — account, clearing, domestic-payment, fx, ledger, lending, notification, party, sanctions, sepa-instant, sepa-payment, swift, transaction.
- **3** did not: `balance`, `billing`, `consent`. This PR does the two non-money-path ones. `balance` is money-path and is a **separate PR** (`feat/approvals-inbox-balance`), so it can be reviewed and held on its own terms.

## `route.ts` source count

| | fetchers (`*Pending`) | keys in `sources` |
|---|---|---|
| before | 14 | 17 (14 read + 3 `not-configured`) |
| after | **16** | 17 (16 read + 1 `not-configured`: `balance`) |

The BFF route is contended — several open PRs add sources to this same list, and two sides adding neighbouring entries to one list **merge clean and keep only one** (the release manifest went 56 -> 55 here that way). The counts above are the check to re-run after any merge. `approvals-inbox.test.ts` asserts the whole ordered id list with `toEqual([...])`, so it changes on every addition.

## Shape

Matched to the 13 existing implementations rather than invented — `openbank-account-service` and `openbank-party-service` were read first and the idiom copied exactly:

- `@GET` + `@RolesAllowed("ROLE_OPERATOR","ROLE_ADMIN")` + `@Authorize(action = "<domain>.approval.read")`, `limit.coerceIn(1, MAX_PENDING_LIMIT=200)`.
- `ApprovalResponse` gains `makerId` and `createdAt` — the two fields the inbox actually renders, and neither existed on these two services' response shape.
- **No OPA change needed**, same as account/party: base `rest.rego`'s `operator-read-any` admits any action ending `.read`, and both services' `*_rest_ext.rego` already grant operators `startswith(input.action, "<domain>.")`.
- `openapi.yaml` additive minor bump (billing 1.4.0 -> 1.5.0, consent 1.7.0 -> 1.8.0) + an `ApprovalResponse` schema.

## Proof by what it prevents

Red first, against **unmodified `origin/main`'s `route.ts`** with the new tests in place:

```
RED_RC=1   Tests  2 failed | 16 passed (18)
```

Green with the change restored:

```
GREEN_RC=0  Test Files 2 passed  Tests 21 passed (21)
```

Full admin-ui suite: `400 passed | 1 failed (401)`, `2053 passed | 1 skipped`. The one failure is `temporal-shipped-claims.guard.test.ts` — a 5s vitest timeout on a guard that walks the whole Kotlin tree, unrelated to this change and reproducible on a contended machine. `npx tsc --noEmit` is clean (0 errors).

Refs #5679

### DIFF
--- a/openbank-admin-ui/src/app/api/approvals/pending/route.ts
+++ b/openbank-admin-ui/src/app/api/approvals/pending/route.ts
@@ -23,13 +23,11 @@ type SourceState = 'ok' | 'forbidden' | 'unavailable' | 'not-configured'
 // the most dangerous state look healthy to an operator.
 const NOT_CONFIGURED_SOURCES = {
   balance: 'not-configured',
-  billing: 'not-configured',
-  consent: 'not-configured',
 } as const satisfies Record<string, SourceState>
 
 type InboxItem = {
   id: string
-  domain: 'lending' | 'sanctions' | 'transaction' | 'domestic-payment' | 'clearing' | 'fx' | 'ledger' | 'swift' | 'sepa-payment' | 'sepa-instant' | 'notification' | 'party' | 'account' | 'agent'
+  domain: 'lending' | 'sanctions' | 'transaction' | 'domestic-payment' | 'clearing' | 'fx' | 'ledger' | 'swift' | 'sepa-payment' | 'sepa-instant' | 'notification' | 'party' | 'account' | 'billing' | 'consent' | 'agent'
   action: string
   resourceId: string | null
   maker: string | null
@@ -101,6 +99,15 @@ type PartyApproval = LendingApproval
 // account-service serves the shared PendingApproval shape for gated account lifecycle actions.
 // Surfacing it here makes a parked freeze or other protected action discoverable before TTL expiry.
 type AccountApproval = LendingApproval
+
+// billing-service serves the same libs `PendingApproval` shape (issue #5679). Before this, a
+// parked `billing.post` / `billing.reverse` four-eyes decision was discoverable only by whoever
+// had been handed its id out of band, and expired silently after the store's 24-hour TTL.
+type BillingApproval = LendingApproval
+
+// consent-service serves the same libs `PendingApproval` shape (issue #5679), for the gated
+// `consent.grant` / `consent.revoke` actions.
+type ConsentApproval = LendingApproval
 
 type AgentProposal = {
   id: string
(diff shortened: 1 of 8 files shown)


JiRaska:
Closing as a duplicate of work that has been open since 2026-08-26.

`#7031` (balance), `#7034` (billing) and `#7037` (consent) already cover exactly these three services. I built this without checking the open queue — the brief I gave the agent said "13 of 16 call `findPending`; balance, billing and consent remain", which was true of `main` and irrelevant to whether someone had already written the fix.

**The older PRs are the better ones**, which is the deciding point rather than seniority. #7034 carries `docs/threat-models/billing-service.md` and `approvals-source-truth.guard.test.ts`; this PR has neither. Per-service PRs are also the right shape here — bundling billing and consent means neither can land while the other is under review.

## One idea from this branch worth carrying over

The balance half (#7991) removed `NOT_CONFIGURED_SOURCES` entirely once the last entry was gone, rather than leaving an empty constant. That mattered because `approvals-source-truth.guard.test.ts` asserts the *literal string* `balance: 'not-configured'`, so deleting the entry without replacing the assertion would have left a guard testing nothing.

The replacement was the invariant the string was standing in for: **every domain the route reports is a domain it reads.** That survives the last `not-configured` entry disappearing, which the string assertion does not. Worth folding into #7031 when it lands last of the three.

Recorded here rather than quietly dropped, since it is the one thing this branch had that the others do not.

## Also worth knowing before merging those three

They all edit the same two files — `approvals/pending/route.ts` and `approvals-inbox.test.ts`, whose `toEqual([...])` asserts the whole ordered id list. Two sides adding neighbouring entries to one list merge **clean and keep only one**; that is how `.release-please-manifest.json` went 56 entries → 55 in this repo, exit 0 with nothing printed.

After each merge, count the sources rather than trusting the diff — `main` currently has 14 fetchers and 3 `not-configured` entries, and each PR should move one from the second group to the first.

PR #7991, opened 2026-09-01
feat(balance): expose pending approvals in the unified inbox

JiRaska:
## What

The **last** of the sixteen services with a four-eyes decide endpoint to gain the matching read (ADR-0227 D2, #5679), and the one where the hole mattered most: both gated actions here move money (`balance.credit`, `balance.debit`).

**Money-path — 2 approvals + threat model required. Expect this NOT to land quickly, and that is the correct outcome.** It is deliberately split from #7990 (billing + consent) so the non-money-path work is not held behind it.

> **Stacked on `feat/approvals-inbox-billing-consent` (#7990)**, because both PRs edit `route.ts` and `approvals-inbox.test.ts`. The base is that branch, not `main` — a stacked PR reads CLEAN against its base, not against main. Retarget to `main` after #7990 merges; do **not** merge #7990 with `--delete-branch` while this is open, that closes this PR irreversibly.

## The defect

A parked `balance.credit` / `balance.debit` decision was discoverable only by whoever had been handed its approval id out of band, and expired silently after the shared store's 24-hour TTL. A maker-checker control nobody can see is a control that fails open by attrition.

## What it does and does not do

- `GET /api/v1/balances/approvals` — read-only, oldest first, limit clamped to 200.
- `balance.approval.read` needs **no policy change**: base `rest.rego`'s `operator-read-any` admits it on the `.read` suffix — the same path account and party already take. (Checked directly; balance's own `balance_rest_ext.rego` uses explicit action sets, not a prefix, so this was verified rather than assumed.)
- Untouched: `ApprovalStore.decide`'s maker != checker invariant, the random ids, the 24h TTL, one-time `EXECUTED` consumption.
- **Deliberately absent:** a decide button on the inbox. Money-path disposal additionally requires SCA (ADR-0227 D4).
- Threat model updated (ADR-0030) with the risk actually traded: operator-visible workflow metadata plus a bounded Redis SCAN, against a maker-checker control that was effectively invisible.

## `route.ts` source count

| | fetchers | `sources` keys | `not-configured` |
|---|---|---|---|
| on `main` | 14 | 17 | 3 (balance, billing, consent) |
| after #7990 | 16 | 17 | 1 (balance) |
| **after this PR** | **17** | **17** | **0** |

With nothing unread, `NOT_CONFIGURED_SOURCES` is **removed** rather than left as an empty placeholder.

`approvals-source-truth.guard.test.ts` asserted the literal strings `balance: 'not-configured'` and `...NOT_CONFIGURED_SOURCES`. Deleting those would have left nothing behind — a guard that asserts a literal only describes the state it was written in. It is replaced by the invariant it existed to protect: **every domain the route reports is a domain the route reads**, and every fetcher is awaited in the fan-out. A source added without a fetcher now fails regardless of what it is called.

## Proof by what it prevents

Red first, with the new tests against the parent commit's `route.ts`:

```
RED_RC=1   Tests  3 failed | 20 passed (23)
```

Green with the change:

```
GREEN_RC=0  Test Files 2 passed  Tests 23 passed (23)
```

`npx tsc --noEmit`: 0 errors.

The two pre-existing Kotlin tests in `ApprovalResourceMappingTest.kt` are **preserved** and extended (the first draft of this branch overwrote them; caught by reading `git diff --cached --stat` showing `M` where `A` was expected, and restored from `origin/main`).

Refs #5679

### DIFF
--- a/docs/threat-models/openbank-balance-service.md
+++ b/docs/threat-models/openbank-balance-service.md
@@ -173,6 +173,25 @@ also be deleted (nothing else in balance-service depends on it).
 
 ## 7. Change log
 
+- **2026-09-01** — The operator approval inbox gains a bounded, read-only
+  `GET /api/v1/balances/approvals` edge (issue #5679, ADR-0227 D2). It returns only the pending
+  approval id, action, resource id, maker id and creation time, to callers already holding
+  `ROLE_OPERATOR` or `ROLE_ADMIN`, behind the same OPA boundary as the checker decision endpoint
+  (`balance.approval.read`, admitted by base `rest.rego`'s `operator-read-any` on the `.read`
+  suffix — no policy change). Results are capped at 200 and ordered oldest first; the endpoint
+  cannot approve, reject or execute anything, and `ApprovalStore.decide`'s maker != checker
+  invariant, the random ids, the 24-hour Redis TTL and one-time `EXECUTED` consumption are all
+  untouched. **Risk class:** confidentiality of operator workflow metadata (account ids of parked
+  credit/debit requests become visible to any operator, not only the one holding the id out of
+  band) plus a bounded Redis SCAN load. No new principal, no service-to-service edge, no money
+  mutation. The countervailing risk it removes is the larger one: until now a parked
+  `balance.credit`/`balance.debit` decision was findable only by whoever had been handed its id,
+  and expired silently after 24 hours — a maker-checker control nobody can see is a control that
+  fails open by attrition. Deliberately NOT added: a decide button on the inbox — money-path
+  disposal additionally requires SCA (ADR-0227 D4) and stays on the per-domain flow. Rollback:
+  remove the GET route; the decision path is unaffected and the admin UI reports the balance
+  source unavailable.
+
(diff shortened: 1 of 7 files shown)


JiRaska:
Closing as a duplicate of work that has been open since 2026-08-26.

`#7031` (balance), `#7034` (billing) and `#7037` (consent) already cover exactly these three services. I built this without checking the open queue — the brief I gave the agent said "13 of 16 call `findPending`; balance, billing and consent remain", which was true of `main` and irrelevant to whether someone had already written the fix.

**The older PRs are the better ones**, which is the deciding point rather than seniority. #7034 carries `docs/threat-models/billing-service.md` and `approvals-source-truth.guard.test.ts`; this PR has neither. Per-service PRs are also the right shape here — bundling billing and consent means neither can land while the other is under review.

## One idea from this branch worth carrying over

The balance half (#7991) removed `NOT_CONFIGURED_SOURCES` entirely once the last entry was gone, rather than leaving an empty constant. That mattered because `approvals-source-truth.guard.test.ts` asserts the *literal string* `balance: 'not-configured'`, so deleting the entry without replacing the assertion would have left a guard testing nothing.

The replacement was the invariant the string was standing in for: **every domain the route reports is a domain it reads.** That survives the last `not-configured` entry disappearing, which the string assertion does not. Worth folding into #7031 when it lands last of the three.

Recorded here rather than quietly dropped, since it is the one thing this branch had that the others do not.

## Also worth knowing before merging those three

They all edit the same two files — `approvals/pending/route.ts` and `approvals-inbox.test.ts`, whose `toEqual([...])` asserts the whole ordered id list. Two sides adding neighbouring entries to one list merge **clean and keep only one**; that is how `.release-please-manifest.json` went 56 entries → 55 in this repo, exit 0 with nothing printed.

After each merge, count the sources rather than trusting the diff — `main` currently has 14 fetchers and 3 `not-configured` entries, and each PR should move one from the second group to the first.

Issue #7867, opened 2026-08-31
Fleet sweep: a null JSON array element is a 500, not a 400 — 23 occurrences, 5 money-path, 21 invisible to the fuzz lane


JiRaska:
## The defect class

Jackson's Kotlin module null-checks a data class's **constructor parameters**. It does **not** check the **elements** of a collection. So a request DTO declared

```kotlin
data class PostJournalRequest(val lines: List<PostJournalLineRequest>)
```

deserialises `{"lines": [null]}` into a `List` holding a null. Kotlin's non-null element type is a compile-time promise nothing keeps. The handler then dereferences the element and throws `NullPointerException`, which `GenericExceptionMapper` renders as a **500 where a 400 belongs** (ADR-0080).

The consequence that traps people is the same one the `@QueryParam` rule already documents: **a `requireNotNull` written against a non-nullable element type is dead code.** The element must be declared nullable for any guard to be reachable — `List<X?>` plus a `requireNotNull` per element, letting libs-runtime's `CommonExceptionMappers` render the `IllegalArgumentException` as 400. Never a service-local mapper (#526).

This is the same mechanism as `check-nonnull-jaxrs-params.py` (#3104, #3624), one level up: that gate covers `@QueryParam`/`@HeaderParam`/`@MatrixParam`, and its baseline is **empty**. It has never looked at body entities or collection elements, so this class was free to spread with nothing to notice.

## Why this is a sweep and not four bug reports

#5913's two-pass fuzz lane reported 24 findings across 11 services. Exactly **two** of the occurrences below were among them — ledger's and party's. The lane never sent a `null` array element to anything else, so **21 of these 23 were invisible to the run that was specifically looking for this shape of problem.**

That is the reason to enumerate rather than fix opportunistically: the detection gap is structural, not a matter of running the fuzzer longer.

## Inventory: 23 occurrences across 13 services, 5 of them money-path

Derived statically. I verified the five money-path rows by hand; the rest are the scanner's traces and are marked as such.

### Money-path (verified by hand against `main`)

| Service | Endpoint | Property | First dereference |
|---|---|---|---|
| ledger | `POST /api/v1/journals` | `PostJournalRequest.lines` | `LedgerResource.kt:147` — **fixed in #7860** |
| psd2 | `POST /v1/consents` | `ObAccess.accounts` | `Psd2Services.kt:162` `accounts?.mapNotNull { it.iban }` |
[...]

PR #7994, opened 2026-09-01
fix(flaky-test-hunter): reject a null array element with 400 instead of 500

JiRaska:
## What

`POST /api/v1/flaky-test-hunter/evidence/analyze` answers **500 where 400 belongs** when a JSON
array in the request body carries a `null` element.

Jackson's Kotlin module null-checks a data class's **constructor parameters**; it does not check
the **elements** of a collection. So `{"components":[null]}` deserialises happily into a
`List<TestIntelligenceComponentInput>` holding a `null`, the first dereference throws, and
`GenericExceptionMapper` renders 500.

Three collections on that endpoint were affected:

- `TestIntelligenceAnalysisRequest.components`
- `TestIntelligenceComponentInput.evidence`
- `TestIntelligenceComponentInput.requiredControls`

`declaredInfrastructure` needed no change and is deliberately untouched: it is validated with
`all { it in INFRASTRUCTURE }`, and a set-membership test on a null is `false`, so a null element
there already yields 400. Re-verified against `origin/main`.

## Fix

House pattern:

1. Declare the element type nullable on the DTO — that is the truth on the wire, and it is what
   makes the guard reachable instead of dead code (a `requireNotNull` written against a
   non-nullable element type can never run).
2. Add a `requireX()` guard method returning the proven-non-null collection.
3. Route every production read through the guard, so it cannot be bypassed —
   `analyze` now works off the value `validateEvidenceRequest` returns.

`IllegalArgumentException` is mapped to 400 by libs-runtime's `CommonExceptionMappers`; no
service-local mapper is added (#526). No validation beyond null-safety was added, the OpenAPI
schema is unchanged in shape (a null element was never valid input), so `info.version` is not
bumped.

## Evidence (red → green)

Proof tests live in the existing `@QuarkusTest` that already drives real HTTP,
`TestIntelligenceEvidenceRoutingIT`.

**RED**, against the unmodified code:

```
./gradlew :openbank-flaky-test-hunter:test --tests '...TestIntelligenceEvidenceRoutingIT'
EXIT=1
5 tests completed, 3 failed        (0 skipped — the app booted)
a null components element is rejected as a bad request()      FAILED
a null evidence element is rejected as a bad request()        FAILED
a null requiredControls element is rejected as a bad request() FAILED
java.lang.AssertionError: 1 expectation failed.
Expected status code <400> but was <500>.
```

**GREEN**, after the fix — exit code and task list quoted in a follow-up comment.

Refs #7867


JiRaska:
**GREEN** — after the fix, on the same host, under the shared Gradle mutex:

```
./gradlew :openbank-flaky-test-hunter:ktlintCheck :openbank-flaky-test-hunter:detekt \
          :openbank-flaky-test-hunter:test --tests 'com.openbank.flakytest.infrastructure.rest.*' --max-workers=2
EXIT=0
> Task :openbank-flaky-test-hunter:ktlintMainSourceSetCheck
> Task :openbank-flaky-test-hunter:detekt
> Task :openbank-flaky-test-hunter:test
BUILD SUCCESSFUL in 2m 3s
```

Per-class JUnit XML (skipped read explicitly, since a boot failure reports SKIPPED rather than FAILED):

```
TestIntelligenceEvidenceRoutingIT     tests=5 skipped=0 failures=0 errors=0
FlakyTestResourceRoutingIT            tests=3 skipped=0 failures=0 errors=0
TestIntelligenceEvidenceContractTest  tests=2 skipped=0 failures=0 errors=0
FlakyTestResourceTest                 tests=1 skipped=0 failures=0 errors=0
```

Both lint gates ran and passed in the same invocation (the first attempt was red on
`function-signature` only, fixed in the follow-up commit).

PR #7995, opened 2026-09-01
fix(campaign): reject a null array element with 400 instead of 500

JiRaska:
## The defect

Jackson's Kotlin module null-checks a data class's **constructor parameters**. It does **not** check the **elements** of a collection. So `data class R(val xs: List<X>)` deserialises `{"xs":[null]}` into a `List` that holds a null. The mapping code dereferences that element and the request answers **500 where 400 belongs**.

A `requireNotNull` written against a non-nullable element type is dead code — the element type has to be declared nullable for any guard to be reachable.

## Sites (all three verified present on `origin/main`)

| endpoint | DTO field |
| --- | --- |
| `POST /api/v1/audiences` | `CreateAudienceRequest.rules` |
| `POST /api/v1/campaigns` | `CreateCampaignRequest.steps` |
| `POST /api/v1/campaigns` | `CreateCampaignRequest.decisions` |

## The fix

1. Declare each element type nullable (`List<StepRequest?>`, `List<DecisionRequest?>`, `List<AudienceRuleRequest?>`) — that is the truth on the wire, and it is what makes the guard reachable.
2. Guard inside the single mapping function that is already the **only** production read of each list (`toSteps`, `toDecisions`, `toRules`), so the raw list cannot be reached un-guarded.

`IllegalArgumentException` is already answered as 400 — by libs-runtime's `CommonExceptionMappers` for the campaign resource, and by `AudienceResource.create`'s own catch. No service-local `ExceptionMapper` is added (#526). No validation beyond null-safety was added.

`openapi.yaml` is unchanged: a null array element was never valid input, so the schema shape does not move and `info.version` is not bumped. This is a `fix`.

## Proof — red then green

Three `@QuarkusTest` cases added to `CampaignRestContractIT`, driving real HTTP with RestAssured (the defect lives in deserialisation, which a direct call to the resource never performs).

**RED — against the unmodified resources** (`./gradlew --no-daemon :openbank-campaign-service:test --tests '...CampaignRestContractIT.a null*'`, redirected to a file, `$?` read immediately):

```
CampaignRestContractIT > a null audience rules element is rejected with 400, not 500() FAILED
CampaignRestContractIT > a null steps element is rejected with 400, not 500() FAILED
CampaignRestContractIT > a null decisions element is rejected with 400, not 500() FAILED
3 tests completed, 3 failed
BUILD FAILED in 23m 6s
EXIT=1
```

Each failure is an assertion, not a boot failure — 0 skipped, and the JUnit XML records for all three:

```
java.lang.AssertionError: 1 expectation failed.
Expected status code <400> but was <500>.
```

**GREEN — with the fix applied:** `EXIT=0` (see the PR discussion for the exact task list).

`:openbank-campaign-service:ktlintCheck :openbank-campaign-service:detekt` also run separately with `$?` read from a redirected file.

Refs #7867


JiRaska:
Verification completed after the PR was opened (all runs redirected to a file, `$?` read immediately, never through a pipe; machine is heavily contended so each Gradle invocation was serialised behind a mutex):

**GREEN — `:openbank-campaign-service:test --tests '...CampaignRestContractIT.a null*'`**

```
> Task :openbank-campaign-service:test
BUILD SUCCESSFUL in 4m 46s
EXIT=0
```

JUnit XML for the class: `tests="3" skipped="0" failures="0" errors="0"` — three real executions, no boot-failure skips.

**GREEN — `:openbank-campaign-service:ktlintCheck :openbank-campaign-service:detekt`**

```
> Task :openbank-campaign-service:detekt
> Task :openbank-campaign-service:ktlintCheck
BUILD SUCCESSFUL in 28s
EXIT=0
```

Both `ktlintCheck` and `detekt` appear as executed tasks in the output, so neither was skipped behind an earlier failure.

PR #8003, opened 2026-09-01
fix(psd2,sanctions): a null JSON array element is a 400, not a 500 (#7867)

JiRaska:
## Summary

Money-path rows of #7867: a null JSON array element (`{"accounts": [null]}`, `{"aliases": [null]}`) was a **500**, now a **400**.

Jackson's Kotlin module null-checks constructor parameters but **not collection elements**, so a null element arrived despite the non-null Kotlin element type and the first dereference threw an NPE that `GenericExceptionMapper` rendered as a 500 (ADR-0080).

The fix, per the issue's prescribed shape:

- **psd2** (`POST /v1/consents`, money-path): `ObAccess.accounts/balances/transactions` element type declared nullable (`List<ObAccountRef?>?`) so a guard is reachable; `ConsentManagementService.createConsent` rejects a null element via `requireNotNull` naming field and index (`access.accounts[1]`) → `IllegalArgumentException` → 400 via libs-runtime. Covers both the Berlin XS2A and the OB resource, which share the usecase.
- **sanctions** (`POST /api/v1/sanctions/screen`, money-path): `ScreenEntityCommand.aliases` element nullable, same guard at the top of `SanctionsService.screen` (before the idempotency lookup, so no work happens for a malformed body); stored entity keeps `List<String>` via `filterNotNull()`.

**Deliberately not here:** the 18 non-money-path rows and the question whether `check-nonnull-jaxrs-params.py` grows a body-entity arm — issue stays open for those.

Refs #7867

## Test plan

- [x] New red-first tests: `createConsent rejects a null JSON array element with IllegalArgumentException` (psd2), `screen rejects a null JSON array element with IllegalArgumentException` (sanctions) — both assert the index-named message and that no downstream call happens
- [x] `./gradlew :openbank-psd2-service:test :openbank-sanctions-service:test` green
- [x] `ktlintCheck` + `detekt` on both modules green
- [x] Client-side mirror DTOs (`SanctionsServiceClient` in sepa/sepa-instant/fx/domestic/kyc) verified to be separate local classes — unaffected
- [ ] CI gates

### DIFF
--- a/openbank-psd2-service/src/main/kotlin/com/openbank/psd2/application/usecase/Psd2Services.kt
+++ b/openbank-psd2-service/src/main/kotlin/com/openbank/psd2/application/usecase/Psd2Services.kt
@@ -92,6 +92,7 @@ class ConsentManagementService(private val consentClient: ConsentServiceClient, 
     ConsentManagementUseCase {
 
     override suspend fun createConsent(command: CreateConsentCommand): ObConsentResponse {
+        requireNoNullAccountRefs(command.request.access)
         val scopes = buildScopes(command.request.access)
         val ibans = collectIbans(command.request.access)
         val validUntil = minOf(command.request.validUntil, LocalDate.now(clock).plusDays(90))
@@ -159,10 +160,28 @@ class ConsentManagementService(private val consentClient: ConsentServiceClient, 
 
     private fun collectIbans(access: ObAccess): List<String>? {
         val ibans = mutableSetOf<String>()
-        access.accounts?.mapNotNull { it.iban }?.let { ibans.addAll(it) }
-        access.balances?.mapNotNull { it.iban }?.let { ibans.addAll(it) }
-        access.transactions?.mapNotNull { it.iban }?.let { ibans.addAll(it) }
+        access.accounts?.mapNotNull { it?.iban }?.let { ibans.addAll(it) }
+        access.balances?.mapNotNull { it?.iban }?.let { ibans.addAll(it) }
+        access.transactions?.mapNotNull { it?.iban }?.let { ibans.addAll(it) }
         return if (ibans.isEmpty()) null else ibans.toList()
+    }
+
+    /**
+     * Reject a `null` JSON array element with a 400 (#7867). Jackson null-checks constructor
+     * parameters but not collection elements, so `[null]` deserialises into a list holding a
+     * null; without this guard the first dereference was an NPE that `GenericExceptionMapper`
+     * rendered as a 500. `IllegalArgumentException` maps to 400 via libs-runtime.
+     */
+    private fun requireNoNullAccountRefs(access: ObAccess) {
+        mapOf(
+            "accounts" to access.accounts,
+            "balances" to access.balances,
+            "transactions" to access.transactions,
+        ).forEach { (field, refs) ->
+            refs?.forEachIndexed { index, ref ->
+                requireNotNull(ref) { "access.$field[$index] must not be null" }
+            }
+        }
     }
 
     private fun mapStatus(s: String): String = when (s) {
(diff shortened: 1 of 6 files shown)


JiRaska:
### The red was a cancelled build, not a finding

`all-green` failed here, and reading it rather than the summary shows why:

```
cancel: build (openbank-psd2-service) / openbank-psd2-service (build)
fail:   all-green
pass:   27

if [ "cancelled" = "failure" ] || [ "cancelled" = "cancelled" ]; then
  echo "One or more service builds failed."; exit 1
```

`all-green` treats a **cancelled** service build as a failure, which is right — a cancelled job proves
nothing — but it means this PR was red on a spot-kill rather than on anything in its diff. 27 checks
passed and none failed.

Rebased onto current `main` (116 commits of drift, **zero conflicts**) so the run that decides this is
against what would actually merge. Both modules compile on the merged result:

```
:openbank-psd2-service:compileKotlin        exit 0
:openbank-sanctions-service:compileKotlin   exit 0
```

### On the change itself

This closes four of the rows in #7867's inventory, and they are the money-path ones — three psd2
`ObAccess` fields and `ScreenEntityCommand.aliases`. The comment on `ObModels.kt` names the exact trap
that made them invisible:

> the `?.` at the call site guards the list, never its contents, and `mapNotNull` protects the RESULT
> from nulls, never the RECEIVER

That is worth keeping in the diff. `accounts?.mapNotNull { it.iban }` reads as doubly defensive and
guards neither the element nor the dereference — which is why these three sat behind a null-safe
operator and still answered 500.

`ScreenEntityCommand` is the row I verified independently on #7867: the REST body binds straight to
the application-layer command, so it is not under `infrastructure/rest` and is not named `*Request` —
invisible to the conventional grep, and money-path.

JiRaska:

```
```

```
```

X an ancestor".

JiRaska:
### Both failures were about the run's age, not the diff — re-merged

`all-green` was red on a **cancelled** psd2 build (a spot-kill; a cancelled job proves nothing, so
treating it as failure is right), and `verification-metadata` reported 20 unpinned Kotlin 2.4.20-RC3
artifacts:

```
##[error]A cold-cache build of the changed modules resolves artifacts that
         gradle/verification-metadata.xml does not pin.
  missing: kotlin-build-tools-api-2.4.20-RC3.jar   (and 19 more)
```

That is a real gap, and it is `main`'s — this PR touches `openbank-sanctions-service/build.gradle.kts`
(one line, adding `rest.assured.kotlin` for the new IT), which forces a cold-cache build of that
module and so resolves the Kotlin compiler and build-tools artifacts. A PR that touches no build file
never resolves them and never sees it.

**`main` already fixed it**, and this branch simply predated the fix:

```
pins kotlin-build-tools-api 2.4.20-RC3:   main 1   this branch's pushed head 0
```

Merged current `main` in and pushed; the pins come with it, jar and pom, sha256 each.

### One thing worth knowing before anyone follows the error's own advice

The message prints a remediation command:

```
./gradlew --write-verification-metadata sha256 :openbank-sanctions-service:testClasses \
          :openbank-sanctions-service:quarkusDependenciesBuild -Dorg.gradle.jvmargs=-Xmx6g
```

I ran it. `BUILD SUCCESSFUL in 4m 43s` — and it changed **nothing**, because on a machine with a warm
Gradle cache `testClasses` comes back `UP-TO-DATE`, the Kotlin compiler artifacts are never resolved,
and `--write-verification-metadata` can only record what the build actually resolves. The failure it
is meant to fix is defined as a *cold-cache* resolution, so the fix has to run against one — a clean
`GRADLE_USER_HOME`, or `--rerun-tasks`.

Not a defect in this PR, but the command as printed is a probe that reports success while measuring
nothing, and the next person will lose the same five minutes to it.

JiRaska:
Poznámka k historii větve: tato větev dříve nesla rozpracovaný merge-based pokus o stejný fix (psd2-only, bez testů). Přepsal jsem ji čistým jedním commitem na aktuálním main, který kryje psd2 i sanctions money-path řádky #7867 včetně red-first testů — obsah PR se tím změnil, scope zůstal stejný. Původní tip větve byl `e16730da0f`.

PR #8011, opened 2026-09-01
feat(account): show the owner everyone who can act on their account

JiRaska:
## The gap

Two independent stores authorise debits on a customer's account, and `authorizeDelegatedPayment` (ADR-0232) consults **both**:

| store | set by | visible to the customer before this PR |
|---|---|---|
| `account_authorizations` | bank operator | **no** |
| delegation grants | the customer | yes |

`recordStoreDisagreement` already exists, so the codebase knows the two can disagree.

The app's profile sheet even described its delegation list plus the PSD2 consent centre as *"the complete answer to who can reach my money"*. It was not — a bank mandate can debit the account without appearing in either. The reassuring half was the only half on show.

## What this adds

- `effectiveAccess(accountId)` — reads both stores with the same active/validity filters and the same `issuedBy` ownership check the guard uses.
- `GET /api/v1/accounts/{id}/authorizations/effective` (account-service)
- `GET /customer/v1/accounts/{id}/access` (customer-edge)

It is written as a **projection of the guard**, not a second opinion. A transparency view that computed access independently would drift from enforcement the first time either store changed — and the view is the half no real debit ever exercises.

`canInitiatePayments` answers the only question an owner has (can this person take money out) instead of restating a role name the customer never chose.

## Access-control decisions

**The edge route is owner-only**, although a delegate may read the account itself. Who else holds access is the grantor's business; a delegate reading it would learn about the owner's other delegates and the bank's mandates, neither of which their own grant concerns.

**account-service repeats the ownership check** against the edge-stamped `X-Customer-Party-Id` — defence in depth against a future edge route that forwards the header but forgets its own check. It answers 404 rather than 403 and returns an empty list for an unknown account, so no response distinguishes "not yours" from "does not exist".

**Party ids only.** No names, contact details or grant labels cross the boundary; this must not become a way to turn an account id into a person.

## Verification

- `:openbank-account-service:test` — 10/10
- `ktlintCheck` + `detekt` clean; `:openbank-customer-edge:compileKotlin` passes
- OpenAPI bumped: account-service 1.13.0, customer-edge 1.47.0

Falsified, not just asserted — dropping the `issuedBy`/`isActiveOn` filter fails `an inactive delegation is not shown` and `a grant issued by someone who does not own the account is not shown`.

One finding from that falsification worth recording: the first run of the negative case *passed*, because 9 of the 10 tests were silently never collected — Kotlin `fun x() = runBlocking {}` infers a non-Unit return and JUnit 5 skips the method. The module's existing tests write `(): Unit = runBlocking`. Fixed; the suite now reports `tests="10"`.

Client side: JiRaska/openbank-app#576

Refs ADR-0232

## Security checklist

- [x] Auth / crypto / payment code changed? → YES — the payment-guard authorization stores are read (read-only projection); mitigations: owner-only at the edge, 404-not-403, empty list on unknown account, party-ids only, threat-model entry added 2026-09-06.
- [x] PII path changed? → no — party ids only, no names or contact details cross the boundary.


JiRaska:
## Heads-up: **both** `openapi.yaml` versions here have already shipped — each needs a re-bump

This PR moves two specs, and main has moved past both since it was last updated (2026-09-03T10:26Z).

| file | this PR claims | live `origin/main` | |
|---|---|---|---|
| `openbank-account-service/.../openapi.yaml` | `1.13.0` | **`1.13.0`** | taken by #8466 (`909dee547a`) |
| `openbank-customer-edge/.../openapi.yaml` | `1.50.0` | **`1.50.0`** | taken by #8536 (`ea114a2457`) |

Both measured off live `origin/main` just now, not off a cached base.

### Why nothing will flag either one

- **Neither will conflict.** Both sides write byte-identical text on the version line, so git
  resolves it silently. A version line is the one field where "merged cleanly" and "still free" are
  different questions, and only the second one matters.
- **CI stays green.** `check-api-contract.py` classifies each PR against its own merge-base, so this
  branch is internally consistent and correctly classified. The collision exists only relative to
  main, which the gate never consults.

Merging as-is republishes two versions that already exist, with different content behind them.

### Fix

```bash
```

then re-read each number off live `origin/main` rather than off this branch's base — today that is
`1.14.0` for account-service and `1.51.0` for customer-edge — and re-run the gate the way
`.github/gates/gates.yaml` invokes it:

```bash
python3 .github/scripts/check-api-contract.py --base $(git merge-base HEAD origin/main) --enforce
```

### Sequencing against the other live claimants

- **account-service** — **#8475** has already re-bumped to `1.14.0` after #8466 took `1.13.0`.
  Whichever of you merges second takes the number after the other.
- **customer-edge** — **#8321** claims `1.51.0`, which is correct against main today, so it is the
  one to sequence behind rather than race. Separately, **#8274** claims `1.48.0`, i.e. *below*
  main's `1.50.0` — stale in the same way but two versions deep. Not your problem, flagged for
  whoever owns it.

Context and the general form of this in #8477. Not touching this branch; comment only.

JiRaska:
**The red `gates (lint-supplychain-security)` on this PR is stale, but a push alone will not clear it — the branch is 180 commits behind `main` and merging conflicts.**

Measured just now:

- The failing check started **2026-09-03T10:03:50Z**. Its cause — four stale `ALLOWED_UNRESOLVED` entries in `check-threat-model-claims.py` — was fixed by **#8460, merged 2026-09-03T10:42:35Z**, 39 minutes later. So the failure is about a tree that no longer exists.
- `main` has moved **180 commits** since this branch's base, and `git merge origin/main` produces **1 conflict**, so I aborted rather than resolving someone else's branch.

Nothing here needs re-running: a re-run replays the recorded merge commit and would reproduce the same red. The branch needs a rebase or merge by whoever owns it, and then the check goes green on its own.

Worth pairing with the other thing already flagged here: this PR claims `openbank-account-service` `info.version` **1.13.0**, and so do #5986, #8466 and #8475. Live `main` is at 1.12.0, so whichever lands first takes the number and the rest need re-bumping — and the same applies to `openbank-customer-edge` **1.50.0**, where #8321 claims 1.51.0 and #8274 claims 1.48.0, below main's 1.49.0.

I have not touched the branch or its worktree (`/Users/jiri.raska/Developer/oss-wt-access`, clean and idle at the time of measurement).

JiRaska:
**The `openapi.yaml` version this branch claims has been taken on `main` while the PR was open. A rebase alone will not fix it — the number has to be re-picked too.**

Measured against `origin/main` just now; the `openapi-version-not-taken` gate is `mode: enforced` and resolves `origin/main` at run time, so this reds on the next run. It has not yet, because a `CONFLICTING` PR produces zero CI runs.

This PR touches **two** specs and both numbers are gone:

| spec | base when branched | this PR claims | `main` now |
|---|---|---|---|
| `openbank-account-service` | 1.12.0 | **1.13.0** | **1.14.0** |
| `openbank-customer-edge` | 1.49.0 | **1.50.0** | **1.50.0** |

The customer-edge one is the easier to miss: the claim is not *below* `main`, it is *equal* to it — #8536 took 1.50.0 on 2026-09-03. A bump that lands on the number already published republishes it.

The branch is 199 commits behind `main` and `CONFLICTING`. #5986 claims the same account 1.13.0 from the same stale base.

**Live claim map for these specs**, so the re-pick survives the other PRs in flight:

| spec | `main` now | already claimed by open PRs | next free |
|---|---|---|---|
| account-service | 1.14.0 | 1.13.0 (#5986, #8011 — both taken), 1.14.0 (#8475 — taken) | **1.15.0** |
| customer-edge | 1.50.0 | 1.51.0 (#8321), 1.52.0 (#8544) | **1.53.0** |
| delegation-service | 1.9.0 | 1.10.0 (#8310), 1.11.0 (#8318) | **1.12.0** |
| domestic-payment | 1.6.0 | 1.7.0 (#8274) | 1.8.0 |

Whoever rebases second takes the next number again — read the version off live `main` at the moment you rebase, not off this comment.

JiRaska:
Follow-up: the gate failure here is stale too, so the rebase clears both it and the version problem in one go.

`threat-model-claims-resolve` fails on this branch and passes on the same branch merged with current `main`. Measured in two detached worktrees:

| tree | exit | result |
|---|---|---|
| `origin/<this branch>` | **1** | `FAIL: 1 false/self-referential claim(s), 4 stale exclusion(s)` |
| the branch **merged with current `main`** | **0** | `45/45 models, 3045 claims … OK: every mechanically checkable mitigation claim resolves to real, non-stub code` |

The four stale exclusions (`openbank-security-scanner|5. Residual risks|SecurityContractTest`, `openbank-sepa-payment|5a. Return path (pacs.004) — STRIDE supplement|openbank.sepa.returns.enabled`, and two more) were removed from `ALLOWED_UNRESOLVED` on `main` after this branch's last run — the gate fails on a *stale* exclusion by design, so an entry cannot outlive the debt it excuses. This branch still carries them.

So nothing in this PR is wrong on that axis. A rebase onto `origin/main` fixes the gate; re-picking the two `openapi.yaml` versions (see the previous comment) is the part the rebase will not do for you.

Issue #6988, opened 2026-08-26
VEX store has no fleet or base-layer scope — a shared-artifact CVE costs 42-336 hand-written statements


JiRaska:
## What

The VEX triage store is **one file per released component** (`openbank-libs/governance/vex/<component>.openvex.json`, 57 of them). It has no scope above the component, so a CVE in something every service shares — a base image layer, a platform BOM artifact, a `openbank-libs-*` sibling — can only be dispositioned by writing the **same statement into every overlay by hand**.

This is not hypothetical, and it is not rare. Three measurements on `main` today:

| CVE | shared origin | statements needed |
|---|---|---|
| `CVE-2025-14969` | `hibernate-reactive-core` (platform BOM) | **48** written, byte-identical `impact_statement` |
| `CVE-2026-59903` | `netty-codec-http` (platform BOM) | 9 named, **0** written (#6720) |
| 8× Go stdlib | `/usr/bin/pebble` in `eclipse-temurin:25-jre` | 42 services × 8 CVEs = **336** (#6717) |

The `CVE-2025-14969` entry is the proof, not an estimate: 48 overlays carry the identical verdict, and the store's own README records that **47 statements** previously sat at `affected` together on the same faulty range arithmetic and had to be corrected together (#4716, #4533, #4707).

## Why it matters

1. **The triage cost scales with the fleet, not with the number of decisions.** One human judgement about one artifact becomes 42–336 file edits. That is the whole reason #6717, #6719 and #6720 are three open tickets: each re-derives the same missing capability.
2. **Fan-out is where a verdict goes stale asymmetrically.** When the `CVE-2025-14969` reasoning was corrected, 47 statements had to move in lockstep. Nothing structurally guarantees they did — a shared-list merge that git calls clean is exactly how one entry quietly does not change.
3. **A gap in the fan-out is invisible.** `referral-service` has **no overlay file at all**, so it silently missed all 48 (#6719). Nothing enumerates released components against overlays, so the omission surfaced only as a triage issue seven months later. `admin-ui` and `case-coordinator-agent` are in the same state today.
4. **It biases the disposition toward the wrong one.** Facing 336 edits, the cheap move is to wait for a base-image bump rather than record the truthful `not_affected`. The store then understates what has actually been reasoned about — the opposite of what a regulator-facing artifact is for.

## Shape of the fix

Not prescriptive, but the options differ in what they cost to get wrong:

[...]

PR #8020, opened 2026-09-01
fix(governance): drop three NO_OVERLAY_YET entries whose overlays now exist

JiRaska:
## The whole PR queue is red behind one stale allowlist entry set

`gates (supplychain)` fails on **31 of the open PRs** right now, and `Validate manifests` on 30,
because it aggregates that shard. The cause is one gate, and it is doing its job correctly:

```
UNFALSIFIED  vex-overlay-coverage                             0.1s  [supplychain]
##[error]self-test verdict wrong — gate is unfalsified: every released component is
         enumerated against the VEX store (#6988, enforced)
```

## What happened

`4ef1e93734` ("write the three JVM VEX overlays") landed today and wrote overlays for
`case-coordinator-agent`, `incentive-service` and `referral-service` — 23 statements each, all three
files present under `openbank-libs/governance/vex/`. It did not remove their `NO_OVERLAY_YET`
entries.

`check-vex-overlay-coverage.py` says of that map: *"Reason required; stale either direction."* So a
declaration that a component has no overlay, made about a component that now has one, is a finding —
which is the design working, not a false positive:

```
NO_OVERLAY_YET declares case-coordinator-agent, which HAS an overlay now
  (openbank-libs/governance/vex/case-coordinator-agent.openvex.json) -- remove the entry
NO_OVERLAY_YET declares incentive-service, which HAS an overlay now ...
NO_OVERLAY_YET declares referral-service, which HAS an overlay now ...
```

Once the gate itself fails, its self-test's "must fail" fixture can no longer be distinguished from
the live failure, so the runner reports **UNFALSIFIED** rather than FAIL — which is why the error text
talks about falsifiability rather than naming the three components.

## The change

Three lines removed. `admin-ui` stays, because it is a released component that genuinely has no
overlay, so its entry still describes something true — and removing it is a finding in the other
direction, which I checked.

## Falsified, not just green

A gate fix that is only ever observed passing proves nothing:

| state | exit |
|---|---|
| a stale entry put back | **1** |
| this fix | **0** |
| `admin-ui`'s live entry removed | **1** |

```
python3 .github/scripts/run-gates.py --only vex-overlay-coverage
  PASS  vex-overlay-coverage  0.4s  [supplychain]
python3 .github/scripts/check-vex-overlay-coverage.py --self-test
  SELF-TEST PASS
```

## Why this is worth merging ahead of other things

Nothing else in the queue can go green until it lands: 31 PRs are failing a required context on a
cause none of them introduced, and the fix touches no behaviour — three entries in a gate's known-gaps
map, in a file whose only purpose is to describe which components lack a VEX overlay.

Refs #6988, #6719

### DIFF
--- a/.github/scripts/check-vex-overlay-coverage.py
+++ b/.github/scripts/check-vex-overlay-coverage.py
@@ -70,9 +70,6 @@ NO_RELEASE_COMPONENT: dict[str, str] = {
 # component -> issue that tracks writing its overlay. Reason required; stale either direction.
 NO_OVERLAY_YET: dict[str, str] = {
     "admin-ui": "#6719 - released component, no triage overlay written yet",
-    "case-coordinator-agent": "#6719 - released component, no triage overlay written yet",
-    "incentive-service": "#6719 - released component, no triage overlay written yet",
-    "referral-service": "#6719 - missed the whole CVE-2025-14969 fan-out for this reason",
 }
 
 


JiRaska:
Superseded by #8012, which merged at 16:02 with the same three entries removed and a better comment
on the mechanism than mine. Closing.

Same diagnosis reached independently, and #8012's framing is the one to keep: *"the failure was
leaving the baseline out of the PR that made it obsolete"*. Worth noting for anyone reading the queue
history that the gate was never wrong here — it reported a declaration outliving its subject, which is
the direction that matters, and the fleet-wide redness was the check working.

One thing my run adds that is not in #8012, since it is about detectability rather than the fix: once
`vex-overlay-coverage` fails for real, the runner can no longer distinguish its "must fail" self-test
fixture from the live failure, so it reports **UNFALSIFIED** instead of FAIL. The error text then talks
about falsifiability and never names the three components — which is why the log reads as an abstract
gate-hygiene complaint rather than "three entries are stale". Measured: 31 open PRs were failing
`gates (supplychain)` and 30 `Validate manifests` behind it, and none of the messages said what to
delete. If that verdict could carry the underlying gate's own findings, this would have been a
one-minute fix for whoever looked first.

PR #8243, opened 2026-09-02
fix(admin-ui): render unknown observability metrics neutrally

### LINKED ISSUE #8242: Admin observability colors unknown metrics as healthy or unhealthy
## Problem

Business Observability formats missing Prometheus values as `N/A`, but its severity colors are calculated with a fallback value of zero. That makes unknown availability look unhealthy while unknown error rate and failed-payment metrics look healthy. Operators must not infer service health from data that was not returned.

## Acceptance criteria

- Null or missing metric values use a neutral visual state.
- `N/A` remains explicit and machine-testable as unknown.
- Existing green/warning/red thresholds remain unchanged for real numeric values.
- Throughput is neutral when unknown and informational when measured.
- Add behavioral regression coverage.

## Scope

Admin UI presentation only; no Prometheus query, BFF, authentication, or RBAC changes.

JiRaska:
## Summary

- render missing Business Observability values with a neutral tone instead of deriving false health from zero
- expose an explicit machine-testable unknown/available state on every KPI card
- preserve all existing success, warning, and danger thresholds for measured values

## Verification

- genuine behavior RED on the previous code: missing metrics had no unknown state and inherited green/red zero fallbacks
- focused Vitest: 1/1 pass after the fix and after rebase
- canonical TypeScript check: pass
- changed-file ESLint: 0 errors (1 inherited auto-load warning)
- diff check: pass
- independent exact-diff review: no blocker

Closes #8242

### DIFF
--- a/openbank-admin-ui/src/app/observability/page.tsx
+++ b/openbank-admin-ui/src/app/observability/page.tsx
@@ -27,6 +27,17 @@ interface MetricsData {
   prometheusUp: boolean
 }
 
+function metricColor(
+  value: number | null | undefined,
+  good: (value: number) => boolean,
+  warning?: (value: number) => boolean,
+) {
+  if (value === null || value === undefined) return 'var(--text-muted)'
+  if (good(value)) return 'var(--success)'
+  if (warning?.(value)) return 'var(--warning)'
+  return 'var(--danger)'
+}
+
 export default function ObservabilityPage() {
   const { t, language } = useLanguage()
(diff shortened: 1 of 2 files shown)


JiRaska (commented):
Independent exact-diff review completed with no code blocker. Unknown values use a neutral tone, measured-value thresholds are preserved, and the behavioral regression is non-vacuous. This comment is evidence only, not an approval; eligible APPROVED review and exact-head green CI remain required before merge.

PR #8304, opened 2026-09-02
fix(account): dead-letter into an explicit topic, not the shared implicit one

### LINKED ISSUE #5752: account-service is the last consumer left on the default party-events DLQ name, kept there by a comment claiming a fix that the rest of the fleet already shipped
## The collision

`openbank-account-service` and `openbank-card-issuance-service` both declare an incoming channel
named `delegation-events-in`, both set `failure-strategy: dead-letter-queue`, and neither sets an
explicit DLQ topic. SmallRye derives the topic from the **channel name**, so both services
dead-letter into the same topic: `dead-letter-topic-delegation-events-in`.

Verified on `origin/main` — both channels consume `openbank.delegation.events`, both carry the
failure strategy, neither carries a topic override.

Two consequences:

- Parked records from two different services interleave in one topic, so "what did
  account-service fail to process" is no longer answerable without inspecting each record's
  headers.
- Any alert or replay tool scoped to one service reads the other's records too. If
  `AccountPartyEventDeadLettered` (or an equivalent) counts records on that topic, it counts
  card-issuance's failures as account's.

## Why it was not fixed in #5751

#5751 wired 35 previously-unwired channels with explicit per-service topics
(`openbank.dlq.<service>.<channel>`) and deliberately baselined these two. Renaming a DLQ topic
that is already live strands whatever records are parked in the old one and blinds any alert
keyed to the old name — that is a migration, not a config edit, and it needs someone to decide
what happens to the existing contents first.

## A second, load-bearing correction

Several services carry a comment asserting that an explicit DLQ topic **cannot** be set, e.g.
`openbank-account-service`'s `party-events-in`:

> The DLQ topic name is left at SmallRye's default (dead-letter-topic-party-events-in): setting
> `dead-letter-queue.topic` explicitly here would be a DOTTED leaf key, which SmallRye Config's
> YAML source silently quotes so the connector never reads it

The first half is true, the conclusion is not. #5751 measured both spellings through the real
`io.smallrye.config.source.yaml.YamlConfigSource`: the **dotted one-liner** is indeed quoted and
inert, but the **nested** form resolves correctly —

```yaml
dead-letter-queue:
  topic: openbank.dlq.<service>.<channel>
```

— the same idiom the fleet already uses for `value: deserializer:`. So an explicit topic was
always available; the comment discouraged the correct fix fleet-wide on the strength of a true
premise and a wrong conclusion. Worth correcting wherever that comment appears, so the next
person does not inherit the same false constraint.

`openbank-transaction-service`'s money-path `payment-scheme-accepted` was carrying the inert
dotted form in `application.yaml` and worked in the deployed pod only because its msg-override
ConfigMap supplied the value — wrong locally and in every test. #5751 corrects that one.

## What would close this

Decide what happens to records parked in `dead-letter-topic-delegation-events-in`, then give each
service its own explicit topic (nested form) and update any alert or replay tooling keyed to the
shared name. Correct the misleading comments in the same pass.

Refs #5745, #5751.

JiRaska:
## account-service was the last consumer left on the shared implicit DLQ topic

SmallRye derives the implicit dead-letter topic from the **channel name alone** — `dead-letter-topic-party-events-in` — and `party-events-in` is declared by **eight** services here. Left implicit they all dead-letter into one topic, and `AccountPartyEventDeadLettered` misattributes every record in it.

`card-issuance-service` already moved off, with a comment explaining exactly this. account-service had not — and carried a comment asserting the fix was **impossible**:

> The DLQ topic name is left at SmallRye's default … setting `dead-letter-queue.topic` explicitly here would be a DOTTED leaf key, which SmallRye Config's YAML source silently quotes so the connector never reads it.

The premise is true and the conclusion is wrong. The dotted leaf key *is* inert — the **nested** form is not, and card-issuance has been using it in production. That false comment was the last one of its kind in the fleet; it is removed here rather than left as a trap for the next reader.

## All four parts, because three of them is a wedge

A rethrow without a working DLQ does not fail safe — it wedges the consumer on the dead-letter send. So this carries:

| part | file |
|---|---|
| `failure-strategy` + **nested** explicit topic | `openbank-account-service/src/main/resources/application.yaml` |
| `KafkaTopic` CR (topics are not auto-created) | `openbank-infra/gitops/components/kafka/kafka-dlq-topics.yaml` |
| KafkaUser **Write** ACL | `openbank-infra/gitops/components/accounts/kafka-account-mtls.yaml` |
| alert retargeted at the new topic | `openbank-infra/gitops/components/observability/prometheus-rules-onboarding.yaml` |

Plus `DeadLetterQueueConfigResolutionTest` (4 cases) asserting the **resolved** value rather than the presence of a string — the nested and dotted forms are indistinguishable by grep and differ only in what the connector actually reads.

## Proven by what it prevents

| run | result |
|---|---|
| `check-incoming-dlq-wiring.py` on this branch | `SUBJECTS=46`, *"clean: 46 incoming channels, every one dead-letters into an explicit topic it may write"* |
| same, with the nested topic deleted | reports `openbank-account-service :: party-events-in: DLQ topic left implicit — … collides across services sharing it` |
| restored | clean again |

**Note the gate is advisory**, so the discriminator is the finding, not the exit code — both states exit 0. Read the output, not `$?`.

## Provenance

This branch was pushed by an agent session that stopped before opening a PR; the commit was complete and orphaned. I verified it against current `main` (merged cleanly), re-ran the gate in both directions, and am opening the PR rather than rebuilding work that already existed.

Closes #5752

### DIFF
--- a/.github/scripts/check-incoming-dlq-wiring.py
+++ b/.github/scripts/check-incoming-dlq-wiring.py
@@ -68,19 +68,23 @@ KNOWN_UNWIRED = {
     # the same way, so it is gone from here too — only `delegation-events-in` remains untouched.
     ("openbank-notification-service", "delegation-events-in"): "#5737 wired only party-events-in; this channel has no failure-strategy yet (#5745)",
     ("openbank-tax-reporting-service", "withholding-remitted-in"): "no KafkaUser in gitops — a Write ACL cannot be granted, so a DLQ would wedge on the send (#5745)",
-    # The four channels that had a DLQ BEFORE #5745, all on SmallRye's implicit
-    # `dead-letter-topic-<channel>` name. Naming them explicitly is a RENAME of a live topic: it
-    # strands whatever is already parked in the old one and silently blinds the existing
-    # AccountPartyEventDeadLettered alert, which reads `dead-letter-topic-party-events-in` by name.
-    # That is an operational change with a migration, not a config edit, so it is recorded here
-    # rather than done in passing. It is not cosmetic: account-service and card-issuance-service
-    # BOTH consume `delegation-events-in` and therefore already share
-    # `dead-letter-topic-delegation-events-in` today — two services' dead letters in one topic,
-    # misattributed by anything that reads it. That is the collision this gate exists to prevent,
-    # observed live (#5745).
-    ("openbank-account-service", "party-events-in"): "pre-#5745 implicit DLQ name; renaming is a live-topic migration",
-    ("openbank-account-service", "delegation-events-in"): "pre-#5745 implicit DLQ name; ALREADY COLLIDES with card-issuance-service",
-    ("openbank-card-issuance-service", "delegation-events-in"): "pre-#5745 implicit DLQ name; ALREADY COLLIDES with account-service",
+    # The channels that had a DLQ BEFORE #5745, on SmallRye's implicit `dead-letter-topic-<channel>`
+    # name. Naming one explicitly is a RENAME of a live topic: it strands whatever is already parked
+    # in the old one and moves what the AccountPartyEventDeadLettered alert must read. That is an
+    # operational change with a migration, not a config edit, which is why they were recorded here
+    # rather than done in passing.
+    #
+    # account-service is OFF this list as of #5752. Its two channels now name
+    # `openbank.dlq.account.party-events-in` / `openbank.dlq.account.delegation-events-in`, with
+    # KafkaTopic CRs and Write ACLs, and the migration is carried explicitly: the two old topics and
+    # their Write ACLs are RETAINED so an existing backlog stays drainable, and the alert matches
+    # both names for the length of the rollout (old and new pods run side by side under a Rollout).
+    #
+    # card-issuance-service's `delegation-events-in` stays. It no longer COLLIDES — account-service
+    # was the other writer of `dead-letter-topic-delegation-events-in` and has moved off it — so what
+    # remains is a single service on an implicit, channel-derived name. Renaming it is still a live
+    # topic migration, and it is card-issuance's to make.
+    ("openbank-card-issuance-service", "delegation-events-in"): "pre-#5745 implicit DLQ name; sole writer since #5752 moved account-service off it, but the rename is still a live-topic migration",
     # NOT the implicit name, and not a rename: fraud-service already sets an EXPLICIT topic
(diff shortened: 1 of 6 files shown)


JiRaska:
## Correction to the body above — this is wider than I described

I wrote that this fixes `party-events-in`. It fixes **both** of account-service's incoming channels and clears **four** baseline entries, not one:

```
- ("openbank-account-service",        "party-events-in"):      "pre-#5745 implicit DLQ name; renaming is a live-topic migration"
- ("openbank-account-service",        "delegation-events-in"): "pre-#5745 implicit DLQ name; ALREADY COLLIDES with card-issuance-service"
- ("openbank-card-issuance-service",  "delegation-events-in"): "pre-#5745 implicit DLQ name; ALREADY COLLIDES with account-service"
+ topic: openbank.dlq.account.party-events-in
+ topic: openbank.dlq.account.delegation-events-in
```

So the `delegation-events-in` collision — two services dead-lettering into one implicit topic, which the baseline itself records as **ALREADY COLLIDES** — is resolved by this change, and card-issuance's entry goes with it because it was only there to describe the other half of the same collision.

## Why removing the baseline is not optional here

`check-incoming-dlq-wiring.py` fails on a stale entry **in both directions** — a declaration must not outlive its subject. Fixing the channels without deleting the entries would turn `main` red on every PR.

That is not hypothetical: it happened yesterday. #7981 wrote three VEX overlays and left `NO_OVERLAY_YET` declaring them, and the coverage gate — correctly — reported it, red-lighting every open PR until #8012 cleaned up. The baseline belongs in the same commit as the change that obsoletes it.

## What the gate's "clean" actually means

Worth stating because it reads stronger than it is:

```
$ python3 .github/scripts/check-incoming-dlq-wiring.py     # on origin/main, before this PR
SUBJECTS=46
clean: 46 incoming channels, every one dead-letters into an explicit topic it may write
```

That sentence is true of the **unbaselined** set. account-service's two channels are absent from the output entirely, because a baselined gap is a declared gap rather than a finding. So "clean" and "complete" are different claims, and the difference is exactly the four entries this PR removes.

Verified both ways after the change: the gate is clean with the entries gone, and re-adding the implicit form reproduces `openbank-account-service :: party-events-in: DLQ topic left implicit`. Note the gate is **advisory**, so the discriminator is the finding, not the exit code — both states exit 0.

PR #8307, opened 2026-09-02
fix(admin-ui): server-backed KYC pagination, contract test, and 404/405 truthfulness

### LINKED ISSUE #8163: fix(kyc): align the paginated case-list contract before Admin UI pagination
## Problem

`KycResource.listCases` returns an envelope `{ items, total, page, size, statusFilter }` and accepts `page`, `size`, and `status`. The committed OpenAPI 1.7.0 instead declares the 200 response as a raw array. The Admin UI currently relies on the implementation shape only opportunistically, requests no page/size, and therefore silently shows the default first 20 records as if they were the full KYC queue.

A separate truthfulness defect exists in the UI: HTTP 405 from either list route is converted to `no_data`, although only the party lookup 404 means that no case exists. A 405 is a method/route-contract failure and must not be presented as an empty answer.

## Acceptance criteria

- Update the KYC OpenAPI list response to the exact implemented envelope and add a provider-level contract test.
- Specify total/page/size/statusFilter nullability and bounds consistently with `KycResource`.
- Verify the implementation returns the declared shape for page 0 and a later page.
- Then add Admin UI server-backed pagination using the declared contract and truthful ranges/empty states.
- Treat only party lookup 404 as `no_data`; 405 and other route failures remain unavailable.
- Preserve KYC RBAC, party lookup semantics, and four-eyes mutation controls.
- Add regression coverage for page transitions, shrinking totals, stale/out-of-order responses, 404 versus 405, and auth-loss fail-closed behavior.

## Rollout note

Land and release the contract correction before, or atomically with, the Admin UI consumer. Do not ship UI pagination against an undocumented response shape.

JiRaska:
## Summary

Follow-up to #8164 (the OpenAPI-only correction). This PR is the runtime + Admin UI half of #8163:

1. **Contract test** — `KycCaseListPageContractIT` proves the corrected `openapi.yaml` (1.8.0,
   `{items,total,page,size,statusFilter}`) actually matches what `KycResource.listCases` returns,
   for page 0 and a later page, and that `statusFilter` echoes the applied filter. #8164 deliberately
   deferred this so it could stay a mechanical spec-only correction; this PR supplies it before the
   Admin UI is allowed to depend on the contract.
2. **Admin UI server-backed pagination** — `openbank-admin-ui/src/app/kyc/page.tsx` now requests
   `?page=&size=20` explicitly against the declared `KycCasePage` envelope instead of silently
   rendering the implementation's default first page as if it were the whole KYC queue. Adds
   Previous/Next controls, a truthful "Showing X-Y of N" range, and a stale/out-of-order response
   guard (`requestSeq` ref) so a superseded request can never clobber a newer one's state — the
   escape hatch is the "All cases" button, which stays enabled while a request is in flight so an
   operator is never stuck behind a slow filter.
3. **404-vs-405 truthfulness fix** — only a genuine 404 from the **party-scoped** lookup
   (`GET /api/v1/kyc/cases/party/{partyId}`, `KycResource.getCaseByParty`) now maps to the `no_data`
   empty state. A 405, or a 404 on the unscoped list route, is a route-contract failure and now
   surfaces through the normal `classifyBffFailure` path (`not_found`/`error`) instead of reading as
   "no cases exist."

## Type of change

- [x] fix

## Linked issues

Closes #8163
Builds on #8164

## Changes

- `openbank-kyc-service/src/test/kotlin/com/openbank/kyc/contract/KycCaseListPageContractIT.kt` — new
  `@QuarkusTest` + Testcontainers-Postgres IT (pattern from `KycOutboxWriteIT`), 3 cases: page 0 shape
  + size respected, page 1 echoes its own page number with total consistent across pages, and
  `statusFilter` echo.
- `openbank-admin-ui/src/app/kyc/page.tsx` — pagination state (`page`, `total`), `requestSeq` staleness
  guard, corrected 404-vs-405 mapping, pager UI.
- `openbank-admin-ui/src/test/kyc-pagination-and-error-mapping.test.tsx` — new regression coverage:
  page-0 request shape, Next/Previous transitions, Next disabled on the last page, a real
  out-of-order-response race (party lookup superseded by "All cases"), party-404 → `no_data`, and
  405 / unscoped-404 must NOT become `no_data`.
- `openbank-admin-ui/src/test/kyc-party-search.test.tsx` — updated its list-route mock to the
  envelope shape (was asserting against the old raw-array response).

## Definition of Done

- [x] `./gradlew :openbank-kyc-service:compileTestKotlin` and the new `KycCaseListPageContractIT`
      pass locally (Testcontainers Postgres).
- [x] `npm test -- kyc-` (admin-ui, via the `pretest` hook) — 9/9 passing, 4 files.
- [x] RBAC, party lookup semantics for the single-case response, and four-eyes approve/reject are
      untouched — this PR is read/pagination/error-mapping only.
- [x] No secrets, tokens, or personal data added.

## Security checklist

- [x] No secrets, tokens, passwords, fixtures, or personal records added.
- [x] No suppression, unsafe cast, or dependency added.
- [x] Auth, RBAC, KYC state transitions, and four-eyes controls are unchanged.
- [x] `check-agent-pr-guard.py --paths <files>` flags this PR as touching `openbank-kyc-service`
      (kyc is an `extra_protected_tokens` entry — high blast radius though not money-path) — flagging
      it for human review, as intended; it is not asking anyone to bypass anything.

## Compliance impact

- [x] GDPR
- [x] AML / 5AMLD
- [ ] PCI DSS
- [ ] DORA
- [ ] PSD2 / SCA / RTS
- [ ] CNB reporting
- [ ] None

This removes an operational risk where an operator could believe they were viewing the complete KYC
queue while only seeing its default first page, and a route-contract failure (405) could read as "no
cases" instead of "unavailable."

## Rollout / Rollback

- Deployment strategy: normal kyc-service + admin-ui release; contract test only, no schema/migration
  changes.
- Rollback plan: revert this PR; #8164's spec correction can stand alone.

## Reviewer notes

Compare `KycCaseListPageContractIT` against `KycResource.listCases` and the `KycCasePage` schema in
`openapi.yaml` (1.8.0, landed in #8164). The Admin UI diff is confined to `src/app/kyc/page.tsx`; no
other KYC-adjacent screens were touched.

### DIFF
--- a/openbank-admin-ui/src/app/kyc/page.tsx
+++ b/openbank-admin-ui/src/app/kyc/page.tsx
@@ -15,11 +15,21 @@ import { Can } from '@/components/auth/AuthGuard'
 import { PartySearch, type PartyHit } from '@/components/party/PartySearch'
 
 const KYC_SERVICE = '/api/svc/kyc-service'
+const PAGE_SIZE = 20
 
 interface KycCase {
   id: string; partyId: string; status: string
   checks: { checkType: string; status: string }[]
   reviewedBy?: string; createdAt: string; updatedAt: string
+}
+
+/** Envelope returned by GET /api/v1/kyc/cases (KycResource.listCases, openapi.yaml KycCasePage). */
+interface KycCasePage {
+  items: KycCase[]
+  total: number
+  page: number
+  size: number
+  statusFilter: string | null
 }
 
 export default function KycPage() {
@@ -36,22 +46,34 @@ export default function KycPage() {
   const [loadedPartyId, setLoadedPartyId] = useState<string | null>(null)
   const loadedPartyIdRef = useRef<string | null>(null)
   const [selectedParty, setSelectedParty] = useState<PartyHit | null>(null)
+  // Server-backed pagination against the declared KycCasePage envelope (issue #8163) —
+  // `page`/`total` are only meaningful in list mode; the party-scoped lookup returns a single case.
+  const [page, setPage] = useState(0)
+  const [total, setTotal] = useState<number | null>(null)
+  // Requests can race (e.g. a slow page-1 response landing after a page-2 click) — only the
+  // most recently ISSUED request may commit state, never whichever happens to resolve first.
+  const requestSeq = useRef(0)
 
-  const load = useCallback(async (requestedPartyId = partyId) => {
+  const load = useCallback(async (requestedPartyId = partyId, requestedPage = page) => {
     const scope = requestedPartyId || null
+    const seq = ++requestSeq.current
     setLoading(true); setUnavailable(null)
     try {
       const url = requestedPartyId
         ? `${KYC_SERVICE}/api/v1/kyc/cases/party/${requestedPartyId}`
-        : `${KYC_SERVICE}/api/v1/kyc/cases`
+        : `${KYC_SERVICE}/api/v1/kyc/cases?page=${requestedPage}&size=${PAGE_SIZE}`
       const res = await fetch(url, { signal: AbortSignal.timeout(5000) })
+      if (seq !== requestSeq.current) return // superseded by a newer request — drop this response
       if (!res.ok) {
         const kind = await classifyBffFailure(res)
-        // A genuine 404/405 on the cases endpoint means "no case for this party",
-        // not a broken app — degrade to the calm empty state rather than an error.
-        const unavailableKind = res.status === 405 || kind === 'not_found' ? 'no_data' : kind
+        // Only a genuine 404 on the PARTY-SCOPED lookup means "no case exists for this party".
+        // A 405 (or a 404 on the unscoped list route) is a route-contract failure — surfacing it
+        // as an empty result would tell an operator the KYC queue is empty when it is actually
+        // unavailable (issue #8163).
+        const isPartyLookupMiss = Boolean(scope) && res.status === 404
+        const unavailableKind = isPartyLookupMiss ? 'no_data' : kind
         if (unavailableKind === 'no_data' || loadedPartyIdRef.current !== scope) {
-          setCases([])
+          setCases([]); setTotal(null)
           loadedPartyIdRef.current = unavailableKind === 'no_data' ? scope : null
           setLoadedPartyId(unavailableKind === 'no_data' ? scope : null)
         }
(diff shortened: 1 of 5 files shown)


JiRaska:
### The failing test is not in this PR's blast radius

`Admin UI build` is red on **one** test out of 423 files, and it is not one of yours:

```
FAIL  src/test/interest-recovery.test.tsx > interest snapshot recovery >
      keeps a retained privileged snapshot blocked through an HTTP 403 retry
AssertionError: expected "vi.fn()" to be called 3 times, but got 2 times
Test Files  1 failed | 422 passed (423)
```

This PR touches four files — `app/kyc/page.tsx`, two KYC tests and a KYC contract IT — and nothing
under `interest`. No shared helper, no shared fixture.

**Measured on clean `origin/main`, in isolation: 5 runs, 5 passes.** So the test is not
deterministically broken, and "expected 3 calls, got 2" on a retry path is the shape of a race rather
than a wrong assertion.

**Stating the limit, because it matters here:** those five runs were the file on its own. CI runs the
whole suite, where ordering and parallelism are exactly what would make a call-count race show up. A
single-file pass does not clear the full-suite case, so I have a full-suite repeat running on clean
main and will add the result. If that comes back green twice, the honest reading is a flake in
`interest-recovery.test.tsx` that this PR merely collided with — worth a flaky-test-hunter entry rather
than a change here.

`Trivy` is also red; that one is worth reading separately, since it is not a test outcome.

JiRaska:
### Full-suite result: green on clean `main`, so this is a flake you collided with

Following up on the limit I stated above — a single-file pass could not clear the full-suite case,
because ordering and parallelism are exactly what would surface a call-count race. So I ran the whole
suite on a clean `origin/main` worktree:

```
Test Files  430 passed (430)
```

`interest-recovery.test.tsx` passes there, in the same run as everything else. Together with 5/5 in
isolation, and with this PR touching four KYC files and nothing under `interest`, the reading is:

- the assertion is not wrong,
- the test is not broken on `main`,
- **`expected 3 calls, got 2` on a retry path is a race**, and this PR met it by being the run that
  happened to lose it.

So nothing to change here. A re-run should clear `Admin UI build`.

Worth an entry for the flaky-test-hunter rather than a fix in this PR — the useful detail for whoever
picks it up is that the failing assertion counts invocations on a **retry** path, which is the shape
that goes wrong when the second attempt has not landed by the time the assertion runs. A `waitFor` on
the call count, rather than a bare `toHaveBeenCalledTimes`, is the usual repair.

`Trivy` is still red and is a separate question — it is not a test outcome, so a re-run will not
address it.

JiRaska:
Correcting my own last line: I said Trivy was red and "a re-run will not address it". Re-read after
posting — it is not red.

```
Trivy   conclusion=SUCCESS  (run at 21:54)
Trivy   IN_PROGRESS         (re-run at 22:03)
```

The PR currently has **0 failing checks** — 20 pass, 2 pending (`Trivy` and `Admin UI build`, both
re-running). The earlier `Trivy` entry I read as a failure had already succeeded; I was looking at a
list that included an in-flight re-run, and the log I tried to open came back `BlobNotFound` because
that job had not written one yet.

So the `Admin UI build` re-run is the only thing left to watch, and the full-suite evidence above says
it should come back green.

JiRaska:
Second full-suite run finished, and it strengthens the conclusion rather than repeating it — because
it failed on a **different** test:

| run | result |
|---|---|
| clean `main`, full suite #1 | 430 passed (430) |
| clean `main`, full suite #2 | 1 failed — `src/test/iaops-case-list-page.test.tsx` |
| CI, this PR | 1 failed — `src/test/interest-recovery.test.tsx` |

Three runs of the same code, three outcomes, two different files. So it is not "one flaky test this PR
happened to meet" — the admin-ui suite does not settle, and any PR can be red on a file it never
touched.

Nothing to change here, and I have recorded the observation with both failures and the reproduction
method on #5285 (the flake log), since it generalises well past this PR.

One method note that matters if you re-check: `interest-recovery.test.tsx` passes **5/5 in isolation**
on the same worktree. Only the full-suite run reproduces it — so re-running that single file is not
evidence either way.

PR #8321, opened 2026-09-02
feat(delegation): reserve delegated spend, so a payment-capable grant can exist

JiRaska:
## What

ADR-0249 D3 says the edge reserves against a grant's cumulative ceilings before initiating a delegated payment, confirms on a settled outcome and releases on failure. delegation-service has owned that reservation API for some time — aggregate, ledger, REST surface, OpenAPI, tests. **Nothing called it.** Fleet-wide, `/api/v1/delegations/{id}/reservations` had zero callers, so the cumulative ceiling was countable and never counted.

This wires it, and then removes the refusal that had become a deadlock.

## The deadlock this closes

The customer channel refused `dailyLimit`/`monthlyLimit` with `400 CUMULATIVE_LIMIT_UNSUPPORTED`, on the ground that nothing counted cumulative spend. delegation-service **requires** a cumulative ceiling on any grant carrying `ACCOUNT_INITIATE_PAYMENT` (D5 — no unlimited access to someone else's account by omission).

Verified live against the sandbox, both halves:

```
A) with a ceiling    → 400 {"code":"CUMULATIVE_LIMIT_UNSUPPORTED"}   (edge)
B) without a ceiling → 400 {"code":"SPEND_WITHOUT_CEILING"}          (delegation-service)
```

So a payment-capable grant could not be created through the customer channel **at all**. And `POST /customer/v1/cards/delegated` requires exactly such a grant to exist — so the additional-cardholder route was unreachable too, answering 403 for a reason no error message named.

Order matters here: the reservation is wired first, so the refusal's premise is actually false before the refusal goes.

## Decisions worth reviewing

- **Reserve before the SCA gate, not after.** A payment the ceiling will refuse must not first cost the customer a biometric prompt and a single-use challenge they cannot get back. Every return past that point releases.
- **A reservation that cannot be established is a refusal.** Upstream down or unparseable stops the payment. An advisory ceiling is not a ceiling.
- **The reservation carries the payment's idempotency key**, so a rail replay takes the headroom once. With no key supplied, each attempt reserves separately — over-counting a retry rather than under-counting a limit.
- **An owner reserves nothing.** No grant, no ceiling. Pinned by a negative test, because a reservation attempted there would 404 and turn every ordinary payment into a 403.
- **`confirmed` means the rail accepted the instruction, not that it settled.** A payment accepted and later failed in clearing leaves the headroom consumed. That over-counts, which is the safe direction for a ceiling — stated in the code and the threat model rather than implied.
- **The 403 names the delegate's own limit** (`DELEGATED_SPEND_LIMIT_EXCEEDED`) but not how much is left. They already know they hold the grant, so this is not an oracle; the grantor's configured numbers stay in the audit trail.

## Tests

`:openbank-customer-edge:test --tests '*Delegat*'` — 55 tests, 0 skipped, 0 failures.

Six new cases over the reservation lifecycle: reserve-and-confirm on the happy path, refusal before SCA when the ceiling is hit, release on a rail refusal, release on an SCA refusal, refusal when delegation-service is unreachable, and the negative control that an owner's own payment reserves nothing.

The two ceiling tests are inverted rather than deleted — from "must be refused" to "must be forwarded verbatim", asserting on the forwarded body, since a route that dropped the field would still answer 201.

`LegacyDelegationArmRefusedIT` needed the reservation stubbed and `delegation-service-url` pointed at the loopback: without it, its control case passed for the wrong reason.

## Not in this PR

The spend-reservation audit events and grantor notification (#5728) are still unbuilt. This adds the caller the counter was written for; the observability half stays open.

## Linked issues
Refs #5728


JiRaska:
Rebased onto main and re-bumped `info.version` **1.48.0 → 1.50.0**.

#8336 took 1.48.0 while this was open, so the branch was declaring a version main already has — an API change shipped under an existing version, which is what the api-contract gate exists to stop (and the class of defect #8150 tracks). The PR still showed **CLEAN** because its checks last ran before that merge; a green tick on a stale run is not evidence, and this was found by diffing the branch's spec against main's rather than by reading the checks.

1.49.0 is left for #8367. Whichever of the two merges second needs its own re-bump — there is no way to make both correct at once.

Both commits re-signed after the rebase.

PR #8334, opened 2026-09-03
feat(notification): alert grantor on first delegated spend

JiRaska:
## Summary
- create one security inbox notification only after `SpendConfirmed`
- persist a producer-owned deduplication key so Kafka redelivery cannot create a second row or send
- classify the grantor notification as SECURITY and reject a non-delegation source

## Verification
- `DelegationNotificationConsumerTest`: 10/10 green
- `NotificationConsumerIT`: 16/16 green, including real Postgres plus in-memory Kafka duplicate delivery

## Rollout
Stacked on #5857 because it supplies the transactional `SpendConfirmed` source event. No activation or deployment is included here.

Refs #5728

## Security checklist

- [x] Auth / crypto / payment code changed? → payment-adjacent: delegated-spend audit events (SpendReserved/SpendConfirmed/SpendReleased) re-emitted and a first-delegated-use notification; no authorization decision changes — the events record what already happened.
- [x] PII path changed? → no — notification content references party/grant ids, no new PII.


JiRaska:
**Both gate failures are stale, not yours — but re-running will not clear them. This one needs a rebase.**

`gate-graduation-guard` fails on a lapsed governance deadline, not on anything this PR touches:

```
::error file=openbank-libs/governance/rules.yaml,line=1956::target_enforce_date 2026-08-31 has passed
and this rule is still advisory/planned (ADR-0144).
check-gate-graduation: 20 advisory/planned rule(s) checked, VIOLATIONS above.
```

That was fixed on `main` after this branch's last run (2026-09-03). Measured three ways just now, each in its own detached worktree:

| tree | exit | result |
|---|---|---|
| `origin/main` | **0** | `18 advisory/planned rule(s) checked, all carry a live target_enforce_date` |
| `origin/feat/notification-delegation-first-use` | **1** | `20 ... VIOLATIONS above` |
| the branch **merged with current `main`** | **0** | `18 ... all carry a live target_enforce_date` |

The branch is **1692 commits behind** `main`. Two rules graduated in between, which is the whole difference between 20 and 18.

**Why `gh run rerun` will not help here** — and this is the part worth separating from the usual advice. A rerun replays the original event, so it re-runs against the same merge commit, whose tree still contains the lapsed `rules.yaml`. That works for failures whose cause is *outside* the repo (the `Dependency review` / `Trivy` flake going around tonight clears on a rerun, because the fix was server-side). It cannot work for a failure whose cause is repo content. **Rebase onto `origin/main`, or merge `main` into the branch** — either makes both gates green without touching your change.

`feature-flag-governance` is in the same state; a merge with current `main` clears it too.

Method note: I read the verdict from the run's `gate-results-*.json` artifact rather than the job log — the gates job prints its meta-gate self-test fixtures, which fail on purpose and are indistinguishable from real failures when grepped.

PR #8367, opened 2026-09-03
docs(customer-edge): publish the surface fields the proxy already forwards

JiRaska:
## What

`getSurface` returns engagement-service's response **verbatim** (`upstream.get(...)`), and that response carries `values`, `deepLink` and `interactionRef` — see the service's own `SurfaceContent`. The edge's `EngagementSurfaceContent` declared only `id`/`slot`/`type`/`variables`.

Three fields the edge serves were invisible to any client reading the contract. No behaviour change here; they are already on the wire.

## Why each one matters

- **`values`** — catalogue items declare variable *keys* only; a trusted campaign assignment is the one case that carries resolved copy. Without it a campaign banner renders with its variables unfilled.
- **`deepLink`** — without it a banner tap has nowhere to go, which is most of what a banner is for.
- **`interactionRef`** — without it the engagement event cannot carry attribution back, so the campaign feedback loop silently measures nothing.

## The part worth reading

I also wrote down where the deep-link trust boundary actually sits, because the schema is the only place a client would look.

engagement-service enforces the **scheme** and nothing else — `CampaignBannerPlacement` does `require(deepLink.startsWith("openbank://"))`, with no constraint on the destination. So a campaign author can emit `openbank://` anything, and **a client must resolve the route through its own allowlist and ignore what it does not recognise.**

That was already true and undocumented. A client written against the old schema could not have known to do it — which is exactly the kind of gap that produces a client that "works" until a campaign points somewhere it shouldn't.

## Note on ordering

Bumps `info.version` 1.48.0 → 1.49.0. If open-bank-oss#8321 (still open, money-path) merges after this, it needs a re-bump to 1.50.0 — it currently claims 1.48.0.

## Linked issues
N/A — publishing already-served fields found while wiring the app client.

### DIFF
--- a/openbank-customer-edge/src/main/resources/openapi.yaml
+++ b/openbank-customer-edge/src/main/resources/openapi.yaml
@@ -3,7 +3,7 @@
 openapi: 3.1.0
 info:
   title: Customer Edge API
-  version: 1.48.0
+  version: 1.49.0
   description: >-
     Customer-facing edge proxy (ADR-0065). Single internet-facing entry point for the
     retail customer app. Validates JWTs from the openbank-customers Keycloak realm,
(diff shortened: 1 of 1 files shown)


JiRaska:
Closing as already delivered. Rebasing this branch onto current main leaves **an empty diff** — the served-but-undeclared surface fields it was publishing (`values`, `deepLink`, `interactionRef`) are now declared on main, so there is nothing left for this PR to add.

Its consumer, app#595, merged earlier today and is unblocked.

Issue #8357, opened 2026-09-03
CI gates: Instant.EPOCH defaults (58 baselined) + Temporal worker switch naming convention


JiRaska:
## Context
AGENTS.md documents two fleet gotchas that are grep-able but had no CI gate:

1. **`Instant = Instant.EPOCH` defaults** (#3882, #4208): a defaulted EPOCH timestamp passes `isNotNull()`, and the liveness-gauge seed fired `WorkflowLivenessStale` after every deploy until #4208. 58 occurrences remain in `src/main` today.
2. **Temporal worker disable-switch naming** (audit E46): six services, six switch names, five under `openbank.<service>.worker.enabled`, lending's under `lending.origination.worker.enabled` — an underivable name forces the fuzz harness (and DR manifests, runbooks) to hard-code a drifting name list.

## What landed
- `check-no-epoch-instant-default.py` — enforced gate `no-epoch-instant-default`, 58 occurrences baselined (ratchet-only, baseline fails stale in both directions).
- `check-temporal-worker-switch-naming.py` — enforced gate `temporal-worker-switch-naming`, convention + one allowlisted exception in `rules.yaml: temporal_worker_switch_naming`.

## Follow-ups
- [ ] Burn down the 58-entry EPOCH baseline (per-service PRs; remove each baseline line in the same PR)
- [ ] Rename `lending.origination.worker.enabled` → `openbank.lending.worker.enabled` across registrar, application.yaml, tests, deployment config; then drop the allowlist entry

Audit: docs/audit-2026-09-03-top50.md (F49, E46).

JiRaska:
**Second follow-up is done — by #8374, merged 2026-09-03T07:18:13Z**, before this issue's checkbox was written. It renamed `lending.origination.worker.enabled` → `openbank.lending.worker.enabled` across the registrar, `application.yaml`, the IT, `rules.yaml`, `gates.yaml`, `fuzz-services.sh`, the threat model and `CLAUDE.md`, and emptied the allowlist.

Verified rather than taken from the PR title — the gate run as `gates.yaml` specifies, in a detached worktree off `origin/main`:

| state | verdict |
|---|---|
| pristine `origin/main` | exit 0 — "6 switches, 6 canonical, 0 allowlisted" |
| rename undone, allowlist empty | **exit 1** — "does not follow the convention" |
| rename done, stale allowlist entry re-added | **exit 1** — "stale allowlist entry … no such `@ConfigProperty` exists" |
[...]

PR #8405, opened 2026-09-03
fix(sanctions): drop EPOCH defaults from SanctionsEntry and entity timestamps

JiRaska:
Burn-down of the 5 sanctions + 3 document entries in the `no-epoch-instant-default` baseline: `SanctionsEntry` timestamps required (both production constructors already pass them; 3 test helpers got a fixed instant), entity timestamps → `Instant.now()`. Floor 62 → 46.

Stacked on #8403 — merge in order.

Refs #8357

Verified: module tests + ktlint BUILD SUCCESSFUL.


JiRaska:
Heads-up: **#8504 duplicates this PR's document-service third** — same three files, same three `EPOCH` → `Instant.now()` edits, same three baseline lines. It was opened at 12:59Z against this PR's 08:03Z, and it adds three `// Never Instant.EPOCH (#3882): …` explanatory comments that this PR does not have.

Whichever lands second conflicts on the source **and** the baseline. Since this PR is older and strictly broader (it also covers sanctions' five entries), the cheap resolution is to close #8504 and cherry-pick its comments here if they are wanted. Flagged on that PR too; I have not closed anything.

For context on where this sits: a seven-PR stack now claims all 40 baseline entries exactly once (#8402, #8403, this, #8454, #8462, #8463, #8465), so the burn-down is fully allocated and needs no new PRs — only landing. **#8403 looks like the single blocker**: GitHub says `CONFLICTING/DIRTY`, but `git merge-tree --write-tree` of its head into its base exits 0, so the flag is stale and a push should clear it.

Issue #8407, opened 2026-09-03
Docs asserting a control that does not exist: sweep the checkable claims, one PR per document


JiRaska:
## The class

A document asserts a control or capability by **name**, the name is checkable, and the named
thing does not exist in the tree. Every instance was true when written and went stale in a file
nothing re-checks. Prose is not a control, and no gate reads it.

Five confirmed instances this week, every one found by accident rather than by looking:

| # | Document | False claim | Fixed by |
|---|---|---|---|
| 1 | admin-ui `/temporal` page | listed `TemporalWorkerConfig`, `OpenBankSaga`, `OpaActivityInterceptor`, `DeterministicRandom` as delivered — `openbank-libs-temporal` has two files and none of those four classes exists fleet-wide; also derived a PSD2 Art. 5(3) claim from an undeployed OPA gate | #7591 |
| 2 | `rules.yaml` | called a domestic-payment fraud enforcement gate "merged … flag default-off" after it was **deleted**, and skipped maker/checker on a money-path service because "the bank-side suspend lands via the fraud pipeline" — a pipeline whose verdict is discarded | #7299 |
| 3 | settlement threat model | drew `gRPC (mTLS)` on a link with no transport authentication, eleven lines above its own row saying it is not mTLS | #7297 |
| 4 | admin-ui cloud-architecture page | said the DORA ICT incident register has "no persistence — results are in-memory", untrue since its migration landed | #7954 |
| 5 | 49 of 56 VEX overlays | say an opentelemetry fix is blocked on a platform bump that has landed | reported as #7987 (security judgement, deliberately unchanged) |

## This sweep

Looking on purpose rather than by accident. Scope: `docs/threat-models/*.md`,
`docs/adr/*.md`, `openbank-admin-ui/src/app/**/page.tsx`,
`openbank-*/src/main/resources/docs/*.md`, `README.md`, `docs/ARCHITECTURE.md`,
`docs/compliance/*.md`.

**The test is mechanical:** does the named thing exist in the tree, and if it is a runtime
control, is it wired? A Kotlin class, a rego rule, a `@RolesAllowed`, a migration, a gate id,
a `KafkaTopic` CR — all checkable. A claim that cannot be checked is not a finding.

Confirmed so far — three threat models naming a regression guard that is in no Kotlin source,
where in each case **the guard itself is real and only the name is wrong**:

- `docs/threat-models/openbank-balance-service.md` — credits `BalanceResourceSecurityTest`;
[...]

PR #8408, opened 2026-09-03
docs(balance): name the access-control artifacts that actually exist

JiRaska:
## Summary

`docs/threat-models/openbank-balance-service.md` §3 and the S1 residual named three access-control
artifacts by names that do not exist in this repository. **Every control they describe is really in
place** — only the names were wrong. Documentation only: no DB, schema, endpoint, role or policy
change, and no behavior change.

## Type of change

- [x] docs

## Linked issues

Refs #8407

## Changes

Three corrections, each with the command that disproves the old name. The probe is trusted only
because the same shape returns a hit for a name that does exist —
`git grep -l "class ServiceInfoResource" origin/main -- '*.kt'` — so silence here is evidence
rather than a broken probe.

**1. The `@PermitAll` regression guard was credited to a class that is in no Kotlin source.**

```
git grep -n "class BalanceResourceSecurityTest" origin/main -- '*.kt'   # no output
```

The only occurrence of that name anywhere is the `BalanceResource` KDoc asserting it. The guard is
real and is `BalanceSecurityContractTest`
(`openbank-balance-service/src/test/kotlin/com/openbank/balance/infrastructure/rest/`), which
asserts by reflection that no `BalanceResource` / `ReconciliationResource` endpoint is `@PermitAll`
and that every one carries `@RolesAllowed`. Corrected in the threat model and in the KDoc that was
the other half of the same claim.

**2. The write endpoints were described as `@RolesAllowed(SERVICE, …)`.**

```
git grep -h "val SERVICE" origin/main -- '*/Roles.kt'      # no output
git grep -h "val OPERATOR" origin/main -- '*/Roles.kt'     # const val OPERATOR = "ROLE_OPERATOR"  (control)
```

The constant is `Roles.API` (`ROLE_API`) — `BalanceResource` credit/debit/hold/initialize all read
`@RolesAllowed(Roles.API, Roles.OPERATOR, Roles.ADMIN)`. `ROLE_SERVICE` is a dead name: granted by
no realm, defined in no `Roles` constant. It was the M2M role until #2442 made the M2M role real,
and the resource, `Roles.kt` and `balance_rest_ext.rego` have all said so since.

**3. The S1 residual credited "the SERVICE rule" in OPA.**

The policy has no SERVICE rule and can have none — `AuthorizeInterceptor` only ever emits
`ANONYMOUS` / `AI_AGENT` / `HUMAN`, which is why `balance_rest_ext.rego:116` carries a comment
saying a rule gated on `principal.type == "SERVICE"` is structurally unreachable dead code. The rule
that actually admits the shared M2M client is `service-balance-m2m` (`balance_rest_ext.rego:150`),
matched on `input.principal.id == "service-account-openbank-services"` and pinned by
`balance_rest_ext_test.rego:106`.

## Reviewer notes

**The residual risk itself is unchanged and still holds.** All six verified in-repo M2M callers
(settlement, transaction, account, statement, billing, agent) authenticate as the same
`openbank-services` client, so the policy is scoped to an action-class and not to a specific caller;
tightening it needs per-caller identity (mTLS SPIFFE, ADR-0017, or an OIDC client per caller) and
remains a tracked follow-up. This PR does not weaken or strengthen that statement — it makes it
name the rule that implements it.

The change is deliberately **not** a deletion. §7 records what the document used to claim, what is
true now, and that only the names changed, so a later reader can tell a correction from a control
that was quietly dropped.

## Definition of Done

- [x] Docs updated.
- [x] No source behavior change (the one `.kt` hunk is a KDoc line).

## Compliance impact

- [x] None — corrects the description of existing controls; no control changed.

## Rollout / Rollback

- Deployment strategy: n/a (docs + KDoc)
- Rollback plan: revert

## Security checklist

      (license, CVE, source).

Checked against the diff rather than assumed. The only Kotlin change in the whole PR is one comment line:

```
- * locked by BalanceResourceSecurityTest.
+ * locked by `BalanceSecurityContractTest`.
```

No `@Suppress`/`as Any` added, no `build.gradle*`/`libs.versions.toml`/`package.json` in the diff, no secret-shaped additions. The `@PermitAll` regression guard itself is untouched — this PR corrects the *name* the threat model and the KDoc give it — so no auth, payment, PII or cardholder path changes and `security-review-required` is deliberately not applied.


JiRaska:

**2. A stale threat-model exclusion, caused by this PR's own change.**

> `ALLOWED_UNRESOLVED['openbank-balance-service|3. Authn / Authz|BalanceResourceSecurityTest'] matched nothing — the claim it excused is gone; delete the entry`

This is the gate working, not a false positive. `check-threat-model-claims.py` fails on a stale exclusion as well as an unresolved claim, so an entry cannot outlive the debt it excuses. I ran it against current `origin/main` in a clean worktree: **exit 0**, `45/45 models … OK`. So the entry is valid on `main` and stale only here — something in this branch removes or renames what it pointed at. **Delete that line from `ALLOWED_UNRESOLVED` in `.github/scripts/check-threat-model-claims.py`** as part of this PR.

JiRaska:
Follow-up with the mechanics measured, because the `threat-model-claims-resolve` half is not fixable on this branch as it stands and the reason is not obvious.

**The branch carries an older, narrower version of the gate.** Running it here, unmerged:

```
EXIT 0 — threat-model claim audit: 23/23 money-path models, 419 mitigation claims
```

Against current `origin/main` the same script reports **45/45 models, 3043 claims**. So the branch's copy predates the widening, sees a fifth of the corpus, and passes. The CI failure comes from the *merged* result, which is what the run actually evaluates — and there `ALLOWED_UNRESOLVED['openbank-balance-service|3. Authn / Authz|BalanceResourceSecurityTest']` is present (line 135 of the merged file) and stale, because this PR is precisely what renames the citation to `BalanceSecurityContractTest`.

That is the gate working as designed: an exclusion cannot outlive the debt it excuses, so the entry has to go in the same PR that removes its subject.

**Merging `main` in produces a conflict**, so this is not a one-line change:

```
CONFLICT (content): Merge conflict in docs/threat-models/openbank-balance-service.md
```

I aborted rather than resolving it — the conflict is in prose about a control in a threat model, and picking a side there is an authoring decision, not a mechanical one.

**The order that works, all three steps in this PR:**

1. `git merge origin/main` and resolve the conflict in `docs/threat-models/openbank-balance-service.md`.
2. Delete the `'openbank-balance-service|3. Authn / Authz|BalanceResourceSecurityTest':` entry from `ALLOWED_UNRESOLVED` in `.github/scripts/check-threat-model-claims.py`.
3. Add the `## Security checklist` section to the PR body **and then push** — the gate reads the body from the event payload, so the edit only reaches it on the next push (an empty commit is enough). Note the checklist is genuinely easy to justify here: the only Kotlin change in the whole diff is a comment —

```
- * locked by BalanceResourceSecurityTest.
+ * locked by `BalanceSecurityContractTest`.
```

— no `@Suppress`, no dependency file, no secret-shaped line, and nothing that changes an auth, payment, PII or cardholder path.

Verify with `python3 .github/scripts/check-threat-model-claims.py --enforce` on the merged tree; it must print `OK` and exit 0. Running it **without** `--enforce` exits 0 even when it finds violations, so the flag is what makes the check a check.

**#8409 is the same shape** — same rename, same stale entry (`openbank-clearing-service|3. Authn/Authz|ClearingResourceSecurityTest`), same missing checklist.

JiRaska:
**Correction, and the fix is pushed. The merge conflict I reported does not exist.**

My previous comment said:

> `CONFLICT (content): Merge conflict in docs/threat-models/openbank-balance-service.md`
> I aborted rather than resolving it — the conflict is in prose about a control in a threat model, and picking a side there is an authoring decision, not a mechanical one.

That was real output, but from the wrong tree. `git worktree add <branch>` checks out the **local** branch ref, and mine was stale — `233d526d33` against `origin/docs/balance-threat-model-control-names` at `015cdf1339`. Merging `main` into a base that no longer exists remotely produced a conflict that the actual PR does not have.

From the correct base the merge is **clean**: exit 0, `13 files changed, 370 insertions(+), 21 deletions(-)`, no conflict. So the "this needs an authoring decision" framing was wrong, and I have done the work instead of handing it back.

**Pushed as `6be54e2c0a`** (merge of current `main`, plus the one-line governance fix).

**What was actually failing, on the correctly merged tree:**

```
exit 1 — FAIL: 0 false/self-referential claim(s), 1 stale exclusion(s)

::error title=stale exclusion::ALLOWED_UNRESOLVED[
  'openbank-balance-service|3. Authn / Authz|BalanceResourceSecurityTest']
  matched nothing — the claim it excused is gone; delete the entry
```

This PR is what renames the §3 citation to `BalanceSecurityContractTest`, so the exclusion keyed to §3 no longer excuses anything, and the gate fails on a stale exclusion by design — an entry cannot outlive its debt. Entry deleted.

**Deleted rather than moved, and that differs from #8409 deliberately.** In the clearing PR the gate *also* flagged §6's change-log line as a phantom claim, so there the exclusion moves from §3 to §6. Here it reports **0 false claims** — §6 mentions the old name (line 190) and the parser does not read it as a citation — so nothing needs excusing and a moved entry would itself be stale on the next run.

Falsified rather than just observed green:

| tree | exit | |
|---|---|---|
| merged, entry present | **1** | 1 stale exclusion |
| entry deleted | **0** | `OK: every mechanically checkable mitigation claim resolves` |
| deleted, then restored | **1** | so the deletion is what makes it pass |

The `## Security checklist` section is now in the PR body as well — added *before* this push, so the gate sees it on this run rather than needing another. The only Kotlin change in the whole PR is one comment line (`* locked by BalanceResourceSecurityTest.` → `` * locked by `BalanceSecurityContractTest`. ``), which is why every box is defensible.

Issue #8432, opened 2026-09-03
The whole KYC notification family has no producer: KYC_APPROVED, KYC_REJECTED and KYC_DOCUMENT_REQUIRED are declared, rendered, SECURITY-classified — and never sent


JiRaska:
## The gap

`NotificationTemplate` declares three KYC templates — `KYC_APPROVED`, `KYC_REJECTED`,
`KYC_DOCUMENT_REQUIRED`. All three are fully built out in notification-service: rendered, given
required variables (`KYC_DOCUMENT_REQUIRED(setOf("documentType"))`), and classified **SECURITY**, i.e.
always-on and un-mutable by notification preferences.

**None of them has a producer. `openbank-kyc-service` emits no notification requests at all** — not
one file under its `src/main` so much as mentions notification.

Verified by counting producer files outside notification-service, excluding tests:

| template | producer files |
|---|---|
| `ACCOUNT_FROZEN` | 3 |
| `TRANSACTION_FAILED` | 1 |
| **`KYC_APPROVED`** | **0** |
| **`KYC_REJECTED`** | **0** |
| **`KYC_DOCUMENT_REQUIRED`** | **0** |

The two that work are produced by account-service and the transaction path. The three that do not
are exactly the ones kyc-service would own.

## What the customer experiences

A customer is never told the outcome of their own identity verification. Not when it is approved,
not when it is rejected, and never that a document is needed. `GET /customer/v1/kyc` exists and the
app renders it honestly — but only if the customer thinks to go and look. Nothing prompts them.

`KYC_DOCUMENT_REQUIRED` is the sharpest of the three: it is the one template whose whole purpose is
to ask the customer to *do* something, and there is also no operation to raise it from. kyc-service's
routes are `cases`, `cases/{id}`, `cases/party/{partyId}`, `checks/{checkType}`, `approve`,
`pep-rescreen`, `reject` — there is no "request a document" transition, and no upload surface
anywhere in the fleet to satisfy one. ADR-0256 (perpetual KYC) is still `decision-status: proposed`.

## A monitor that cannot fire

`PushFallbackConfiguredButInert` in `prometheus-rules-push.yaml` alerts when a push fan-out for
`ACCOUNT_FROZEN|KYC_REJECTED|KYC_DOCUMENT_REQUIRED|TRANSACTION_FAILED` reaches a party with no
device and no fallback is routed. For two of those four templates the series can never be non-zero,
[...]

PR #8439, opened 2026-09-03
feat(kyc): tell the customer how their own identity verification ended

JiRaska:
Closes the first half of #8432.

## What was wrong

`KYC_APPROVED` and `KYC_REJECTED` have been declared in notification-service since it was written — rendered, given required variables, classified **SECURITY** so a customer cannot mute them. **Nothing has ever emitted either.** kyc-service published no notification requests at all; not one file under its `src/main` so much as mentioned notification.

A customer was never told the outcome of their own identity verification. `GET /customer/v1/kyc` reports it accurately, but only to someone who thinks to go and look — nothing prompts them.

notification-service has exactly three inputs — `openbank.notification.requests`,
`openbank.delegation.events`, `openbank.party.events` — and what reaches them is:

| source | templates it produces |
|---|---|
| account-service → requests | `TRANSACTION_COMPLETED` |
| sca-service → requests | `SCA_APPROVAL` |
| campaign-service → requests | dynamic, whatever a campaign names |
| `DelegationNotificationConsumer` ← delegation.events | the 8 `DELEGATION_*` |
| **kyc-service** | **nothing at all — this PR adds `KYC_APPROVED` + `KYC_REJECTED`** |

There is no kyc-events consumer either, so the three KYC templates were genuinely unreachable.

(An earlier version of this body claimed `ACCOUNT_FROZEN` had 3 producers and `TRANSACTION_FAILED`
had 1, as the working contrast. That was wrong — those were string matches on an HTTP error code
and two SEPA enums, not emissions. Neither is produced either; 12 of the 23 declared templates are
never emitted by anything. Corrected in #8432, and it makes the problem larger, not smaller.)

## Where it hooks in

`approveCase` and `rejectCase` — the pair `KycResource` actually calls. `approve`/`reject` are reached only by tests and are left alone; I initially wired the emission into those by mistake and the tests caught it, which is the reason two of them assert the call rather than just the verdict.

The rejection carries the operator's audit-trail reason, because the template renders it to the customer. "We could not verify your identity" with no reason is a dead end for someone who now has to work out what to fix.

## KYC_DOCUMENT_REQUIRED is deliberately not here

There is no transition to fire it from and nowhere for the customer to put a document. A push asking for something that cannot be supplied is worse than the silence it replaces. That is ADR-0256's territory and it is still `decision-status: proposed` — #8432 has the detail.

## A direct emitter, not the outbox

This service's outbox is single-destination: every entry goes to `kyc-outbox-out`, and `OutboxMessage` carries no topic, so a notification request cannot ride it without teaching the shared outbox contract about destinations. A direct emitter is also what account-service, campaign-service and sca-service already do for this exact envelope.

**The cost, stated rather than hidden:** a failed emit means the customer is not told while the verdict stands. That asymmetry is the right way round — a broker hiccup must never roll back a four-eyes compliance decision, nor turn a completed review into a 500 an operator retries into a duplicate. It is logged at WARN rather than dropped, because "the customer was not told" is an operational fact someone can act on. #8432 records that a durable version needs the outbox to learn destinations first.

## The part that would only have failed in production

`openbank.notification.requests` **is ACL-locked** — account-service, campaign-service and sca-service all carry explicit grants and notification-service reads it — so `allow.everyone.if.no.acl.found` does not apply. Without a `Write, Describe` grant on the kyc-service KafkaUser every emit fails at runtime with `TOPIC_AUTHORIZATION_FAILED`, and **no unit test can see it**. Same trap that bit the decoupled-SCA push in #2038. The ACL is in this PR alongside the code.

## Tests

**103 in kyc-service, 0 failures** (1 pre-existing Pact-broker skip, not mine). Four new:

- approve notifies, and does not send a rejection;
- reject notifies **with the reason the template renders**;
- a refused transition (case not `UNDER_REVIEW`) notifies nobody — so the notification cannot become a side effect of merely asking;
- **the load-bearing one:** a failing notification does not undo the verdict, and the verdict is still persisted rather than merely returned.

Falsified: removing the `runCatching` reddens exactly that last test and nothing else.

One process note worth passing on — `ktlint` reported *up-to-date* and clean on a changed file; `--rerun-tasks` then found a real import-order violation. Worth not trusting the cached green on a file you just touched.

## Linked issues
Refs #8432


JiRaska:
Reviewed this against the issue and the fleet's working producers — the implementation is right. `KafkaCustomerNotificationPublisher` is an idiom match for account-service's `KafkaNotificationRequestPublisher`, `recipient = partyId` is the documented convention every producer on this topic follows (`NotificationConsumer` line ~545 says so explicitly, and `resolveEmailRecipient` looks the real address up from party-service for the `KYC_REJECTED` no-device EMAIL fallback), and the `KafkaTopic` CR + DLQ for `openbank.notification.requests` already exist, so the ACL was the only infra piece missing. Leaving `KYC_DOCUMENT_REQUIRED` unproduced is the right call.

Two things opened as #8471, stacked on this branch so this one does not need reworking:

1. **The gate this is red on.** `gates (registry-kotlin-data)` fails on `event-contract-coverage-ratchet` — producing to a second topic makes kyc-service owe an ADR-0006 contract at `openbank-contracts/openbank-kyc-service/asyncapi.yaml`. Adding it then makes the grandfathered `openbank-kyc-service:openbank.kyc.events` baseline line stale, because the ratchet checks both directions. #8471 has both, plus a negative control proving the gate is not passing vacuously.

2. **A test that can see the wire.** The four tests here mock `CustomerNotificationPort`, so they cannot distinguish "the port was called" from "a `NotificationRequest` notification-service will act on". The gap that matters most is `NotificationTemplate.unknownVariables`: the consumer **rejects and acks** a request carrying an undeclared key, so a stray variable is silent non-delivery — successful send at this end, nothing at the customer's. #8471 adds `KycOutcomeNotificationWireIT`, which boots the service, swaps the channel for an in-memory sink and drives the real approve/reject endpoints. Falsified rather than asserted: removing both `notifyOutcome` call sites takes it to 2 failed of 3, and adding one undeclared `status` variable to `KYC_APPROVED` takes it to 1 failed of 3 on exactly that assertion.

No changes requested here — #8471 is additive and touches no `src/main` file.

JiRaska:
Closing this: **main has since grown ** (#8547 phase 2), which consumes `openbank.kyc.events` and maps `KYC_CASE_APPROVED`/`KYC_CASE_REJECTED` onto the same two templates this PR emits directly from `KycService`.

Merging both would send every customer **two** notifications for one KYC decision.

The consumer on main is also the better implementation, and by my own argument: it rides an event written in the same transaction as the decision through kyc-service's outbox, so it cannot be lost the way this PR's best-effort emit can — exactly the reasoning I used to choose the consumer shape for consent notifications in #8491. It also needs no `Write` ACL on `openbank.notification.requests` for kyc-service, so the gitops change here is unnecessary too.

Nothing here is worth salvaging: the port, the publisher, the ACL and the `notifyOutcome` swallow helper all exist to make the emitter safe, and the emitter is what is redundant. `KYC_DOCUMENT_REQUIRED` stays unproduced for the reasons in #8535 / #8568, unchanged by this.

Found while rebasing #8491 onto current main.

PR #8449, opened 2026-09-03
chore(governance): renew consent pentest attestation before it reds the fleet on 2026-09-18

JiRaska:
## The defect

`openbank-libs/governance/attestations.yaml` carried `consent.pentest` dated `2026-07-28` with `ttl_days: 21`. It expired on **2026-08-18**.

`check-readiness-attestations.py` treats expiry as advisory (the collector already stops counting an expired entry, so the score self-corrects) — until `--stale-fail-days`, default 30, at which point it becomes an error. The `readiness-attestation-format` gate is **enforced**, so this was a scheduled fleet-wide CI failure on every PR, independent of the diff:

| `--today` | exit |
|---|---|
| 2026-09-17 | 0 |
| **2026-09-18** | **1** |

Same shape as the two governance time-gates that lapsed on 2026-08-31 and failed `Validate manifests` on ~90 unrelated PRs (#7919) — but this one was knowable 15 days ahead.

## Why renewal, not deletion

The gate's own remedy is "delete the entry, or re-attest a real event". Deletion looked right at first: `api-fuzz.yml` has concluded `failure` for six consecutive runs since 2026-08-21, which reads as a dead lane, and a dead lane expiring its attestation is exactly the design the file documents.

That reading is wrong, and the repo already warns about it — a fuzz job list is not evidence about a service. Per-job:

```
success  schemathesis (openbank-consent-service)
failure  schemathesis (openbank-account-service)   <- and 6 others
```

The consent job **succeeded** on 2026-09-03 in run `33720692606`. The workflow-level `failure` is seven other services. Deleting on the run conclusion would have removed a live, earnable attestation.

## The `ops` count

R8 requires a CI-minted pentest attestation to record the exercised surface. The run makes two passes:

| pass | selected | auth-blocked | exercised |
|---|---|---|---|
| authz ON | 14 | 11 | 3 |
| authz OFF | 14 | 0 | **14** |

`ops: 14` is the authz-OFF figure — the `Selected minus auth-blocked` formula R8 states, taken from the run's own summary line `==> [openbank-consent-service] authz OFF: Selected: 14/14 auth-blocked=0`. It matches the `13/13 exercised` convention recorded in the `PENTEST_OPS_DEBT` note this PR removes. The authz-ON pass alone (3) is below the floor of 5 and could not mint an attestation.

Recording `ops` requires dropping `consent.pentest` from `PENTEST_OPS_DEBT` — the gate fails on a stale debt declaration in either direction.

## Verification

```
--self-test                exit=0
--today 2026-09-03         exit=0
--today 2026-09-18         exit=0   (was exit=1 before this change)
```

The negative case is the pre-change measurement above, not just the post-change green.

## What this does NOT fix

- **`ledger.pentest` decays 2026-09-08 and turns the same gate red on 2026-10-09** (measured: `--today 2026-10-08` exit 0, `2026-10-09` exit 1). Its lane is `dast-zap-baseline.yml`, which is green, but a ZAP run produces no operations count, so what `ops` means for it needs deciding (#5769).
- **The recurrence is structural.** Nothing writes `attestations.yaml` — `by: ci-schemathesis` is a human transcribing a run URL. A 21-day TTL on a hand-maintained file guarantees expiry, and the gate escalates 30 days later, so every hand-entered pentest attestation is a fleet-wide red 51 days after it is written. This PR buys 51 days.
- **The gate's guidance names a file nothing produces.** R8 tells you to source `ops` from `fuzz-reports/<svc>-ops*.json`; that run's artifact contains seven files and none of them is an ops JSON, so the number has to come from the log line instead.

Refs #2365, #5769


JiRaska:
**Correction to one bullet in this PR's description.** I wrote:

> The gate's guidance names a file nothing produces. R8 tells you to source `ops` from `fuzz-reports/<svc>-ops*.json`; that run's artifact contains seven files and none of them is an ops JSON.

The observation was right and the conclusion was wrong. The emitter for that file merged in **#8406 at 08:32Z today**; the run I downloaded the artifact from (`33720692606`) started at **05:52Z**. So the file was missing because it did not exist yet at run time, not because R8 cites something unimplemented. The next api-fuzz run will carry it, and from then on `ops` should be read from the artifact rather than from the log line.

That does not change the `ops: 14` in this PR. The value is the same either way — the artifact records the same `Selected minus auth-blocked` figure the summary line does, and R8's floor was falsified in both directions against real logs from that very run (no `ops` → exit 1; `ops: 3` → exit 1, below the floor of 5; `ops: 14` → exit 0).

Nor does it change the reason this PR exists: `consent.pentest` expired 2026-08-18 and turns the enforced gate red on 2026-09-18, measured by bisection (`--today 2026-09-17` exit 0, `--today 2026-09-18` exit 1).

The other caveat in the description stands and is worth repeating, since it is the part that recurs: nothing writes `attestations.yaml` automatically, so a 21-day TTL plus a 30-day escalation makes every hand-entered pentest attestation a fleet-wide red 51 days after it is written. **`ledger.pentest` does exactly that on 2026-10-09** and is not fixed here.

Issue #2901, opened 2026-07-31
auto-retry-cancelled.yml has no escalation path — 208 failures went unreported because a workflow_run job's red is addressed to nobody


JiRaska:
`auto-retry-cancelled.yml` has no escalation path on failure. It is triggered by `workflow_run`, so when it fails there is no PR to show a red check on and no commit status anyone reads — its red is addressed to nobody, exactly the blind spot `dependency-submission.yml` sat in for three days before it grew a `raise-issue` job.

## What that cost, measured

The workflow failed **208 times** and nobody noticed. Lifetime totals as of 2026-07-31:

```
W=repos/JiRaska/open-bank-oss/actions/workflows/auto-retry-cancelled.yml/runs
for s in "" "&status=success" "&status=failure" "&status=skipped"; do
  gh api "$W?per_page=1$s" --jq '.total_count'
done
```

| | count |
|---|---|
| total runs | 1069 |
| success | **0** |
| failure | 208 |
| skipped (job `if:` false) | 861 |

`208 + 861 = 1069` — no third category. Every run either never entered the job or entered it and exited 1. The first run ever recorded, `2026-07-29T21:57:14Z`, already concluded `failure`.

The underlying defect (a bare `gh run rerun` in a job with no `actions/checkout`, so `gh` cannot resolve the repo) is being fixed in #2898. **This issue is not about that defect.** It is about the two days of silence around it, and the fact that it surfaced only because a scheduled task happened to be reading this workflow's run history while watching for something else entirely. Nothing in the repo was going to report it.

## Why this workflow specifically is the sharp case

Its entire purpose is to act unattended — a spot reclaim is retried without a human in the loop. So "nobody is looking" is not an accident of this workflow, it is the design. That makes silent failure indistinguishable from silent success from every angle: the spot-killed run just sits there red, looking like an ordinary build failure, which is precisely the human round-trip #2330 existed to remove.

It is also the only `workflow_run`-triggered workflow in the fleet, so the scope here is one file:

```
for f in $(git ls-tree -r --name-only origin/main .github/workflows/); do
  git show origin/main:"$f" | grep -qE '^\s+workflow_run:' && echo "$f"
done
# -> .github/workflows/auto-retry-cancelled.yml
```

## Proposed change

Add a `raise-issue` job modelled on `dependency-submission.yml` (lines ~188-252), which is the house pattern:

[...]

PR #8459, opened 2026-09-03
fix(ci): give gh-retry.sh its first caller — spot-kill retry (#6853, #7976)

JiRaska:
## The measured fact

`.github/scripts/gh-retry.sh` landed in #8371 and had **zero callers**:

```
$ grep -rln 'gh-retry.sh' .github/
.github/scripts/gh-retry.sh
```

It landed and was not live. Meanwhile the exact failure it was written for reproduced on
[run 33738356244](https://github.com/JiRaska/open-bank-oss/actions/runs/33738356244) (2026-09-03, 50s):

```
jobs API attempt 1/3 failed: gh: API rate limit exceeded for installation ... (HTTP 403)
jobs API attempt 2/3 failed: ...
jobs API attempt 3/3 failed: ...
##[error]Could not read the jobs of .../runs/33737873956 (last: unknown); NOT re-running.
spot-kill-auto-retry run=33737873956 conclusion=cancelled decision=jobs-unreadable
```

Its target, run 33737873956, is a genuine `cancelled` reclaim still stranded at attempt 1 —
the same outcome #6255 produced, by the same mechanism the library exists to remove.

## What this changes

`spot-kill-retry.sh`'s own `is_transient` + 15s/30s loop is **replaced by `gh_retry`**, on
both call sites it wrapped (the jobs-API detection query and `gh run rerun`). The library waits
until a rate limit actually **resets** (capped by `GH_RETRY_MAX_WAIT_SECONDS`, 600s here)
instead of spending a fixed 45-second budget against a limit whose window is minutes.

Attempts go 3 → 5: the number now buys *reset windows*, not quick retries. The job takes an
explicit `timeout-minutes: 50` to bound the worst case. `ubuntu-latest` is free on a public
repo, so the trade is a concurrency slot against a reclaimed CI run that otherwise needs a human.

The script keeps only what is genuinely specific to it — the "already running" answer as a
**success** (#5489), which the library would classify `final` and escalate on a run that is in
fact already being re-run.

## Three defects found while wiring — each proven by making it fail first

**1. The library's self-test fired on the *caller's* argv.** `source x.sh` with no arguments
leaves the caller's positional parameters in place, so `[[ $1 == --self-test ]]` at the bottom
of a file whose entire contract is "source me" matched when the *caller* was self-tested.
Measured: `spot-kill-retry.sh --self-test` printed gh-retry's five lines and exited **0**,
having run **0 of the caller's 14 assertions**. Now guarded on `BASH_SOURCE[0] == $0`.

**2. `(last: unknown)` was a subshell bug.** `with_retry` runs inside a command substitution, so
the `RETRY_LAST_ERROR` it set never reached `main` — the production log above reports `unknown`
about a rate limit it had read in full. The text now crosses on a file
(`GH_RETRY_LAST_ERROR_FILE`), asserted by a new case.

**3. The sparse checkout would have broken the job.** It fetched only `spot-kill-retry.sh`;
sourcing the library would have died on line 78 with `No such file or directory`. Verified by
effect before adding the second path.

## Evidence — by effect, not by appearance

Driving the **real script** against a stub `gh` reproducing the 403 for the first 4 calls:

| | exit | decision | gh calls | re-runs issued |
|---|---|---|---|---|
| **before** (origin/main) | 1 | `jobs-unreadable` | 3 | **0** |
| **after** (this PR) | 0 | `rerun-cancelled` | 6 | **1** |
| control: limit clears at once, before | 0 | `rerun-cancelled` | 3 | 1 |
| control: limit clears at once, after | 0 | `rerun-cancelled` | 3 | 1 |

The control matters: the fix is not "retry everything harder" — behaviour is unchanged when the
limit clears immediately.

### Negative controls (all three flip red as they must)

- `ATTEMPTS` back to 3 → the #6853 case fails: `exit=1 (want 0), rerun calls=0 (want 1)`.
- remove the last-error file write → `FAIL the decision line lost the error text`.
- restore the unguarded self-test trigger → **exit 0 with 0 assertions run**, the vacuous green.

Three legacy assertions hard-coded `3/3` and would have passed *for the wrong reason* under
`ATTEMPTS=5` (the stub simply ran out of scripted answers). They now script five attempts and
are named `5/5`.

## Gates

- `gh-retry.sh` gains a `SUBJECTS=` line and a new **`gh-retry-selftest`** gate — it had none,
  which is part of how a library with no callers stayed invisible.
- `spot-kill-retry-selftest` `min_subjects` 12 → 14.
- Verified: `run-gates.py --only spot-kill-retry-selftest --only gh-retry-selftest` → PASS=2;
  `--list`, `check-gh-repo-context.py`, and all eight `gates.yaml` meta-validators exit 0;
  shellcheck and actionlint clean.

`GH_RETRY_DEAD_LETTER` is deliberately **not** enabled: this workflow already has `raise-issue`
(#2901) with a much better body, and two issues per incident is the #5266/#5285 split ledger
in a new place.

## Follow-up not taken here

`.github/scripts/check-ruleset-context-parity.py` carries a *third* implementation of the same
transient classification, in Python. Out of scope for this PR.

Refs #6853, Refs #7976


JiRaska:
Heads-up from #8473, which also edits `gh-retry.sh`'s self-test — a merge-order artifact, not a defect in this PR.

`gh-retry.sh --self-test` counts its own subjects with a manual `subjects=$(( subjects + 1 ))` next to each case (4 today), and `gates.yaml` sets `min_subjects: 4` against that. #8473 adds five more assertions to the same self-test — the shared transient-pattern corpus plus its negative cases — but they sit before the `echo "SUBJECTS=${subjects}"` block and, knowing nothing about this PR's counter, do not increment it.

Measured on the merged tree (`git merge-tree` of `9786de10a` and this PR's current head `9057b5eda`, clean):

```
self-test ok: transient retried to success
self-test ok: rate-limit class retried
self-test ok: final fails immediately (1 attempt)
self-test ok: exhaustion fails after max attempts
self-test ok: corpus has 34 cases, 13 of them must-stay-FINAL
self-test ok: every corpus case classifies as documented
self-test ok: an over-broad pattern is caught by the corpus (negative case)
self-test ok: a deleted pattern is caught by the corpus (negative case)
self-test ok: an unreadable vocabulary is loud, rc 2, and retries nothing
SUBJECTS=4
gh-retry self-test: all 4 cases behaved
```

Nine assertions ran; four were reported. Not red — the floor still passes at 4 >= 4 — but the number understates by more than half, and the trailing line says "all 4 cases behaved" about nine. That matters for exactly the reason this PR's own gates.yaml comment gives ("Subjects are assertions, not files: the floor drops the day someone deletes a case"): a floor that has drifted below the real count stops being able to notice the deletion it exists for. Five could be removed with the gate still green.

Whoever lands second owns the fix — increment `subjects` alongside the new cases and raise `min_subjects` to match. No action needed here if this merges first; #8473 will pick it up on rebase.

Worth noting a hazard in the counter's shape while it is being touched: a hand-maintained increment sitting apart from the assertion it counts is the same class of thing as a hand-kept gate scope, and it drifts the same way — silently, and in the direction that reads as passing. Deriving the count (e.g. counting the `self-test ok` lines the run actually printed) would make a contributed case impossible to leave out. Your call; the manual counter is fine as long as nothing lands beside it, which is precisely what just happened.

Full comparison and the falsification runs are in #8473.

JiRaska:
Follow-up to my earlier comment, on the `--self-test` argv hijack — your `BASH_SOURCE[0] == "${0}"` guard is the right shape, and I have a measurement that argues for landing this PR sooner rather than later.

**main's copy is still vulnerable.** #8473 branches from `main`, so it carries the naive `if [[ "${1:-}" == "--self-test" ]]`. Reproduced there with a two-line caller:

```
#!/usr/bin/env bash
source .github/scripts/gh-retry.sh
echo "CALLER REACHED"
```

```
== caller WITHOUT --self-test ==   rc=0   caller reached: YES
== caller WITH --self-test ==      rc=0   caller reached: NO — HIJACKED
                                   last line: gh-retry self-test: all cases behaved
```

`rc=0` in both. The library's own success line is the only thing printed, so a caller's suite reads as green having run nothing — exactly what you hit with spot-kill's 13 cases.

**Why this matters beyond spot-kill:** #8473 adds a *second* caller. Its `ruleset-context-parity` self-test sources the library to check the shell and Python classifiers agree message-for-message over a shared corpus. On main's guard that caller is safe only *by accident* — `bash -c` passes no operands, so `$1` is unset. Measured with one operand added:

```
== bash -c "source lib; ..." with no operands ==   transient / final      (correct)
== same, with an operand added ==                  self-test lines        (hijacked)
== same, with `set --` first ==                    transient / final      (correct)
```

So #8473 clears `$@` before sourcing. That is caller-side belt-and-braces; your guard is the half that actually fixes it for everyone, which is why it should not wait.

**One thing I tried and backed out, worth recording.** I wrote the identical `BASH_SOURCE[0] == "${0}"` line on my branch, expecting a byte-identical change to merge as a no-op. It does **not** — `git merge-tree` went from clean to conflicting, because our explanatory comments above the line differ, and the merged file was left with conflict markers at line 191 (both shell self-tests then died on `syntax error near unexpected token '<<<'`). Reverted; the library fix is yours alone, and the merge is clean again:

```
merge rc=0
merged tree: BASH_SOURCE guard present exactly once
  gh-retry:  all 4 cases behaved
  spot-kill: 14 passed, 0 failed
  parity:    22 cases
```

Worth knowing generally: "same change on both sides merges cleanly" is not true when the surrounding comments differ, and this repo's CI would not have caught it — a conflicted merge only shows up if someone materialises the merged tree and runs it.

No action needed on your side; #8473 does not depend on this PR either way.

PR #8460, opened 2026-09-03
fix(gates): clear four stale threat-model claim exclusions that have main red

JiRaska:
## What

`threat-model-claims-resolve` is **failing on `origin/main`** — reproduced on a clean checkout of `7d3bcf902f` with no local diff, so this is main's own state, not any PR's. It is blocking the `gates (lint-supplychain-security)` shard on every open PR.

Four `ALLOWED_UNRESOLVED` entries no longer match any claim. That is a failure **by design** — the gate's own docstring:

> a stale entry fails in BOTH directions … because a gate whose scope is a hand-kept list reads as PASSING when the list is short

Same shape as `gate-subject-floor`, which caught me the same way earlier today.

## Cause

Threat-model documents are being corrected, and a corrected claim moves out of the section its exclusion key names — usually into the change log, which records the absence. #8418 (merged 09:17Z) is one such rewrite.

## Three deleted, one retargeted — and the difference matters

| entry | verdict |
|---|---|
| `document-service \| Trust boundaries \| openbank.billing.billing.event` | gone → **delete** |
| `openbank-fx-service \| 6. Change log \| FxConversionService` | gone → **delete** |
| `openbank-sepa-payment \| 5a. Return path… \| openbank.sepa.returns.enabled` | gone → **delete** |
| `openbank-security-scanner \| 5. Residual risks \| SecurityContractTest` | **moved → retarget** |

Deleting the fourth would have been wrong. Removing it surfaced a **live PHANTOM finding** for the same identifier under `6. Change log`: #8418 rewrote §5 to say the test is per-service, and the claim survives in the change log, which necessarily names the thing whose absence it records — a phantom by construction, not a defect. Retargeted, with its reason updated (it still cited #8418 as *open*).

## Method, because a grep gets two of four wrong

Classifying by searching the `.md` for the identifier reports *"key still correct"* for fx-service and security-scanner — **a mention in prose is not a claim.** The gate's own parser is the authority on what counts, so the classification comes from `--report`, not from `grep`. My first pass used grep and disagreed with the gate on half the entries.

## Falsification

```
security-scanner key back on the WRONG section  -> exit 1, 2 stale + 1 phantom
restored                                        -> exit 0
```

`--self-test`: ok (0 failures). Gate through the runner: PASS.

## Overlap

**#8439 is open and edits `openbank-security-scanner.md`.** This change is in the `.py`, so there is no textual conflict — and if that PR moves the claim again the gate will say so, which is the gate working. I did not touch the document.

## Linked issues

Refs #8418 — the merged rewrite that made the security-scanner entry stale.

### DIFF
--- a/.github/scripts/check-threat-model-claims.py
+++ b/.github/scripts/check-threat-model-claims.py
@@ -126,8 +126,6 @@ ALLOWED_UNRESOLVED: dict[str, str] = {
     #
     # Eight entries name an OPEN PR. Those are the sweep's own corrections; they retire themselves
     # the day each PR lands, and re-editing the same rows here would conflict with a live branch.
-    'document-service|Trust boundaries|openbank.billing.billing.event':
-        'open PR #8415 — no such topic; the real ingress is `openbank.billing.fee.event`',
     'onboarding-service|5. AML / sanctions override — special controls|kyc.check.override':
         'an ADR-0068 planned control; absent from `rules.yaml` and from every resource',
     'onboarding-service|5. AML / sanctions override — special controls|confirmedBy':
@@ -152,8 +150,6 @@ ALLOWED_UNRESOLVED: dict[str, str] = {
         'removed by #4221; `application.yaml:174` records the removal in a comment',
     'openbank-fraud-service|6. Change log|fraudServiceUrl':
         'occurs only inside a comment in `fraud_rest_ext.rego`; not a policy input',
-    'openbank-fx-service|6. Change log|FxConversionService':
-        'open PR #8414 — the wrapper is real and is `FxService.scoreFraudShadow()`',
     'openbank-ledger-service|8. Change log|ledgerServiceUrl':
         'occurs only inside a comment in `ledger_rest_ext.rego`; not a policy input',
     'openbank-lending-service|3. Controls in place (this slice)|lending.origination.worker.enabled':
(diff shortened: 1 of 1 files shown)


JiRaska:
This is the survivor — I opened #8464 thirteen minutes later with a byte-identical set of removals and the same re-key, and have closed it as the duplicate. I went straight at the red without running the claim check first, which is the mistake; yours also carries the better in-code comment on the retarget.

Handing over the verification, since it is not in either PR body and one half of it is a trap:

**Running the script bare proves nothing.** My first pass returned **exit 0 on both the fixed tree and a deliberately sabotaged one**. The script ends `return 1 if a.enforce else 0`, and `gates.yaml` invokes it with `--enforce` — without that flag every finding downgrades from `::error` to `::warning` and the exit code stays 0. Re-run with the invocation CI actually uses, against your change:

| state | exit |
|---|---|
| `--enforce`, fixed tree | **0** |
| `--enforce`, one deleted entry restored | **1** |
| `--enforce`, restored again | **0** |
| `--self-test` | **0** |

`SUBJECTS=45` in every state, so the green is not bought by the corpus collapsing.

**The pre-existing red is real and reproduces on bare `origin/main`** — clean detached worktree, zero changes, `gates (lint-supplychain-security)` exits 1 with `FAIL: 1 false/self-referential claim(s), 4 stale exclusion(s)`. Every open PR inherits it, which is why several unrelated branches show exactly one failing check.

Two things worth knowing for anyone reading that job's log:

- The `failure: known-positive-accepted / harness-broken / slow / over-budget / floor-missed` annotations on that job are `run-gates.py`'s **own self-test fixtures**, which are supposed to fail. They are not findings, and they make the annotation list read far worse than it is.
- `release-scope-mismatch-gate` and `threat-model-updated-on-trust-boundary-change` also fail *locally* with `PR_DIFF_BASE is empty but this gate requires it — refusing to run vacuously`. That is correct behaviour outside CI, not a second defect.

Cause, for the record: #8429 widened the gate at 09:23:15Z; #8414, #8415, #8416 and #8418 merged between 09:17 and 09:42. The header of `ALLOWED_UNRESOLVED` predicted exactly this — "eight entries name an OPEN PR … they retire themselves the day each PR lands" — and the both-directions check caught it on its own author within a minute. Nobody deleted the rows.

PR #8466, opened 2026-09-03
fix(account): revoking an unknown authorization answers 404, not 500

JiRaska:
## The defect

`DELETE /api/v1/accounts/{accountId}/authorizations/{authorizationId}` answered **500 INTERNAL_ERROR**
for an authorization id that does not exist. Verified from the boot log of the authenticated-fuzz lane
(#5913), run 33720692606, traceId `0466cbbc-17c8-4dee-b40a-94c5bfdcd5d9`:

```
ERROR [co.op.li.ap.er.GenericExceptionMapper] Unhandled exception (traceId=0466cbbc-...):
com.openbank.account.application.usecase.AuthorizationNotFoundException: Authorization not found: e3e70682-c209-1cac-a29f-6fbed82c07cd
```

A not-found is the caller's fact, not a server fault. account-service is money-path, so reporting it
as 5xx inflates the error budget and buries real faults in the same bucket.

## Why a service-local mapper, not `NoSuchElementException`

libs-runtime already maps `NoSuchElementException` to 404, so re-parenting the exception was the
alternative. I went with a service-local `@Provider` because that is what this repo already does —
not a guess:

| type | mapper | status |
|---|---|---|
| `AccountNotFoundException` | `AccountNotFoundExceptionMapper` (service-local) | 404 |
| `ProposalNotFoundException` | `ProposalNotFoundExceptionMapper` (service-local) | 404 |
| `AuthorizationNotFoundException` | **none** | **500** |

Every other not-found type in this service is a service-local `@Provider` with its own error code.
`CommonExceptionMappers`' own header sanctions the pattern and literally names this service's mapper
as the example ("adding `AccountNotFoundExceptionMapper` ... still wins over the generic
`NoSuchElementExceptionMapper`").

**On #526** — that rule forbids the *opposite* case: a second mapper for a type libs-runtime already
registers (`IllegalArgumentException` / `IllegalStateException` / `Exception`), where the two
`@Provider`s collide non-deterministically per request. Neither type here is one of those, so #526
does not apply. Re-parenting to `NoSuchElementException` would also have flattened the specific
`AUTHORIZATION_NOT_FOUND` code into the generic `NOT_FOUND` and left this service the only one whose
not-found types are handled differently from its four siblings.

## The set fixed, and how I enumerated it

`git grep -n "class .*Exception" origin/main -- 'openbank-account-service/src/main/kotlin/**'` — 10
exception types, then cross-referenced against the two mapper files:

| type | verdict |
|---|---|
| `AuthorizationNotFoundException` | **fixed → 404** (the measured defect) |
| `AuthorizationNotOnAccountException` | **fixed → 404** (same family, see below) |
| `AccountNotFound` / `AccountUpdateConflict` / `AccountNotEmpty` / `ProposalNotFound` / `ProposalForbidden` / `ProposalExpired` / `ProposalSca` | already mapped (404/409/422/404/403/409/400) |
| `ProductNotEligibleException` | already 422 — extends `IllegalStateException` deliberately |
| `AccountScreeningUnavailableException` | **not touched** — availability failure, not not-found |
| `AccountOpeningBlockedByScreeningException` | **not touched** — policy refusal, not not-found |

`AuthorizationNotOnAccountException` is included because the request addresses the composite resource
`/accounts/{A}/authorizations/{B}`; when B exists on another account, *that* resource does not exist.
The IT confirms it had the identical 500.

The last two also render as 500 today and are arguably wrong (503 / 422), but neither is not-found
semantics, so they are out of scope here rather than fixed on a hunch. Worth a follow-up issue.

## Both cases answer an identical body, on purpose

`AuthorizationNotOnAccountException`'s message names both ids. Echoing it — or answering a different
status — would make the endpoint an existence oracle: an operator scoped to account A could tell
"this authorization id exists somewhere" from "this id does not exist". The distinction stays in a
WARN log, where it is diagnosable. A unit test asserts the two bodies are field-for-field identical.
Same reasoning #8011 applies to the sibling account-access endpoint.

A unit test that mocks the repository cannot see this — `AuthorizationServiceTest` already asserts the
exception *type* and passes against the broken code. The type was never in doubt; the status was. So
the assertion is the HTTP status, driven through the real endpoint (RestAssured + `@TestSecurity`,
real Postgres) in `AccountAuthorizationLifecycleIT`.

**Before** (mapper reverted to `origin/main`, everything else identical):

```
AccountAuthorizationLifecycleIT > revoking an authorization that does not exist answers 404, not 500 FAILED
    Expected status code <404> but was <500>.
    JSON path code doesn't match. Expected: AUTHORIZATION_NOT_FOUND  Actual: INTERNAL_ERROR
AccountAuthorizationLifecycleIT > revoking an authorization that belongs to another account answers 404 FAILED
    Expected status code <404> but was <500>.
3 tests completed, 2 failed          (tests="3" skipped="0" failures="2")
```

That run independently reproduces the reported defect *and* establishes the second one.

**After**: `268 tests, 0 failures, 1 skipped` — the single skip is the pre-existing broker-gated
`AccountEventPactProviderVerificationTest` (`@EnabledIfSystemProperty(pactbroker.url)`), unrelated.

```
:openbank-account-service:test          BUILD SUCCESSFUL   268 tests / 0 failures / 1 skipped
:openbank-account-service:ktlintCheck   BUILD SUCCESSFUL   (Main + Test + ProviderPactTest source sets)
:openbank-account-service:detekt        BUILD SUCCESSFUL
```

Lint was run as a separate invocation after the test run, since Gradle stops at the first failing task.

## API contract

`openapi.yaml` documents the 404 on the revoke operation; `info.version` 1.12.0 -> 1.13.0. Classified
by the gate, not by me, and the version was re-read off live `origin/main` immediately before commit:

```
$ python3 .github/scripts/check-api-contract.py --base $(git merge-base HEAD origin/main) --enforce
api-contract gate: openbank-account-service: additive change, info.version 1.12.0 -> 1.13.0 — OK (required >= MINOR).
```

`version.txt` is untouched — release-please derives it from the `fix(account)` type.

## Overlap check

Open PRs were scanned by changed **file**, not title (`gh api --paginate .../pulls?state=open`, 55 open,
497 changed files). Three others touch this service: #8011 (`feat/account-access-transparency`) adds a new
`/access` endpoint and edits `AuthorizationService.kt` / `AuthorizationResource.kt` — different concern,
no overlap with the mapper files, but it **also moves `openapi.yaml`**, as does #5986. Whichever of the
three lands first takes 1.13.0; the other two need a re-bump. No pushed branch without a PR and no live
worktree claims these files.

Refs #5913

### DIFF
--- a/docs/threat-models/openbank-account-service.md
+++ b/docs/threat-models/openbank-account-service.md
@@ -51,6 +51,7 @@ that is balance-service).
 | **I**nfo disclosure | IBAN / account enumeration via `/iban/{iban}` | AuthZ on lookup; rate limiting at gateway; no PII in IBAN response beyond need |
 | **I**nfo disclosure / **IDOR** | Customer reads another party's account/balance via a guessed id (reads are gated by role, not ownership; the edge calls with a ROLE_OPERATOR M2M token) | Primary control is at the customer-edge (resolves ownership before proxying, finding A1). **Defense-in-depth here:** when a call carries `X-Customer-Party-Id` the read must belong to that party, else 404 (no existence oracle) — catches an edge bug/new route that forwards the header but skips its own check. Operator/service reads (no header) unaffected. |
 | **I**nfo disclosure | Domain metrics leak PII / enable per-customer inference via high-cardinality labels | `DomainMetrics` low-cardinality contract (ADR-0077): `openbank.accounts.created` tagged only by `product_type` (closed `AccountType` enum) + `currency`; `openbank.accounts.closed` adds a `reason` **normalized to a closed set** (`customer_request`/`regulatory`/`fraud`/`inactivity`/`unspecified`/`other`) — the operator-supplied free-text reason never becomes a label; outbox-backlog gauge tagged only by `service`. Never an account id, IBAN, party id, or balance. Counters increment only after the commit + publish. `/q/metrics` is cluster-internal |
+| **I**nfo disclosure | Authorization-id enumeration via the revoke route: `DELETE /api/v1/accounts/{accountId}/authorizations/{authorizationId}` distinguishing "this id exists on another account" from "this id does not exist" | `AuthorizationNotFoundExceptionMapper` and `AuthorizationNotOnAccountExceptionMapper` both answer 404 with a byte-identical `AUTHORIZATION_NOT_FOUND` body, so an operator scoped to one account gets no existence oracle over ids on accounts they do not hold. The distinction is kept in a WARN log, not on the wire. Same posture as the `X-Customer-Party-Id` row above. |
 | **D**oS | Mass open/close churn | Gateway rate limits; outbox decouples event load |
 | **E**oP | Viewer escalates to freeze/close | Distinct roles; OPA enforce; deny-by-default |
 | **S**poofing / **E**oP (M2M) | Compromise of the `oidc-client` secret → mint a ROLE_OPERATOR token → inject arbitrary transactions via transaction-service `POST /api/v1/transactions` | Secret held only in a K8s Secret (ExternalSecret from Vault), never in image/git; rotatable; least-privilege client (`openbank-services`); welcome-bonus call is **flag-gated default-OFF** and **sandbox-only**. Residual: transaction-service does not currently distinguish caller identity beyond the role — accepted residual risk in sandbox (see §5) |
@@ -94,6 +95,20 @@ not change any existing request's outcome until explicitly flipped.
   it is a strict improvement over the prior state of no guard at all.
 
 ## 6. Change log
+
+- **2026-09-03** — **Inbound REST error surface on the authorization routes**, no new route, caller,
+  edge or privilege. Revoking an authorization that does not exist answered 500 INTERNAL_ERROR; it now
+  answers 404, via two service-local mappers (`AuthorizationNotFoundExceptionMapper`,
+  `AuthorizationNotOnAccountExceptionMapper`) added to the existing
+  `com.openbank.account.infrastructure.rest.ExceptionMappers` file. RBAC, OPA action
+  (`account.authorize`) and the revoke logic are untouched — only the status and body of an
+  already-refused request change. **Security-relevant half:** the cross-account case
+  (`AuthorizationNotOnAccountException`) previously leaked through the generic 500 and now returns a
+  body byte-identical to the unknown-id case, closing an authorization-id existence oracle for an
+  operator scoped to a different account; a unit test asserts the two bodies match field for field.
+  **Risk class:** availability/observability (a client fault stops consuming this money-path service's
+  5xx error budget) plus the enumeration hardening above; no money mutation and no new principal.
+  Rollback: delete the two mappers and both types fall back to the generic 500 mapper. (#5913)
 
 - **2026-08-27** — **New tightly scoped inbound diagnostic edge:** the sandbox's
   `journey-product-catalog-read` k6 CronJob may make one read-only request to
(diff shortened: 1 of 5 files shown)


JiRaska:
## Gate follow-up

`gates (lint-supplychain-security)` came back with `69 gates PASS=66 WARN=1 FAIL=2`. Two failures,
and only one was mine.

**Mine — `threat-model-updated-on-trust-boundary-change` (ADR-0030 D2), now fixed in 7a3a29e5c5.**
The gate was right: `ExceptionMappers.kt` is an inbound REST surface and I moved it without touching
`docs/threat-models/openbank-account-service.md` in the same diff. Fixed by adding the STRIDE row the
change actually earns — both revoke failure modes answer a byte-identical 404, closing an
authorization-id existence oracle for an operator scoped to another account — plus the dated change-log
entry. Re-run locally:

```
threat-model-diff gate: openbank-account-service: trust-boundary change WITH a
docs/threat-models/openbank-account-service.md update — OK.        (exit 0)
```

**Not mine — `threat-model-claims-resolve` is already red on `main`.** Measured rather than assumed:
the most recent failing `main` run (33740289253, `ci.yml`) reports `69 gates PASS=64 FAIL=1` with
byte-identical findings — the same 1 phantom claim and the same 4 stale `ALLOWED_UNRESOLVED` entries,
in `openbank-security-scanner`, `document-service`, `openbank-fx-service` and `openbank-sepa-payment`.
None is account-service.

I also checked that this PR does not *add* to that gate. Running it on this branch produces exactly the
same 5 findings and zero mentioning account-service, so the two mapper class names cited in the new
threat-model text resolve to real tracked source rather than becoming a sixth phantom claim.

That failure looks like it is already being worked separately (there is a live
`fix/threat-model-stale-exclusions` branch), so I have deliberately not touched it here — it is four
other services' documentation and does not belong in an account-service fix.

### One caveat on reading that gate's log

The job log shows a `16 gates PASS=7 WARN=1 FAIL=5` block above the real summary. Those five are the
run-gates **self-test fixtures** (tagged `[t]` — "an enforced gate that fails", "a gate that exceeds
its timeout", etc.), which are supposed to fail. The real verdict is the `69 gates` line. The log also
echoes this PR's body verbatim, so it contains the words
`revoking an authorization that does not exist answers 404, not 500 FAILED` — that is my quoted
*before* evidence, not a failing test in CI.

JiRaska:
## Correction to my previous comment

I wrote that `threat-model-claims-resolve` "will stay red on this PR no matter what I do here, because
it is red on `main`". **That is no longer true, and I should not have stated it as a standing fact.**

#8460 ("fix(gates): clear four stale threat-model claim exclusions that have main red") merged to `main`
as `78fd54fd0d` at 10:42Z — between my measuring the failure and my writing that sentence. The next run
on this PR picked up the fixed base, and `gates (lint-supplychain-security)` is now **SUCCESS**, with all
three gates jobs green:

```
SUCCESS  gates (lint-supplychain-security)
SUCCESS  gates (gitops-api)
SUCCESS  gates (registry-kotlin-data)
```

The measurement behind the original comment was correct when taken; the mistake was writing a claim
about *current remote state* with an implied shelf life. Both findings from that comment still hold as
history — the trust-boundary gate failure was genuinely mine and is fixed in `7a3a29e5c5`, and the
claims failure was genuinely not mine — but the "will stay red" prediction is withdrawn.

## Separately: `all-green` was a cancelled-run artifact, not a real failure

`all-green` showed FAILURE while every gate was green. It belonged to Services CI run 33746458222,
whose own conclusion was `cancelled`: `Detect changed services` was cancelled, `build` therefore
skipped, and the aggregate reported failure on the back of that. Meanwhile the previous run
(33745500868, for the superseded commit `d133d7785`) was still sitting queued.

Resolved by cancelling the stale run for the superseded commit and re-running the current one. No code
change was involved, and nothing about the diff changed — worth noting for anyone reading the check
history, since a red `all-green` next to 17 green checks reads like a real regression and was not one.

JiRaska:
Heads-up: #8475 (the follow-up that maps the two sanctions-gate exceptions this PR deliberately left
out) also appends to `ExceptionMappers.kt`. Measured conflict, tracked in #8477 — one hunk, disjoint
sides, resolution is a concatenation.

Separately: `openapi.yaml` auto-merges between the two and both bump to `1.13.0`, so the version
collides without ever conflicting. Whoever lands second re-reads `info.version` off live `origin/main`.

PR #8475, opened 2026-09-03
fix(account): sanctions gate answers 503/422, not 500 — and never echoes the matched name

JiRaska:
## The defect

Both exceptions thrown by the ADR-0032 §C sanctions gate on `POST /api/v1/accounts` had **no
JAX-RS `ExceptionMapper`**, so both fell through to libs-runtime's `GenericExceptionMapper` and
answered **500 INTERNAL_ERROR** — an availability failure and a compliance refusal, each reported
to the caller as "the server broke". account-service is money-path, so both inflate its error
budget and bury real faults in the same bucket.

Reproduced through the real endpoint before any fix (`AccountOpeningScreeningStatusIT`, 4/4 red):

```
a sanctions HIT refuses the open with 422, not 500 FAILED
    Expected status code <422> but was <500>.
    JSON path code doesn't match. Expected: ACCOUNT_OPENING_BLOCKED  Actual: INTERNAL_ERROR
a sanctions REVIEW refuses the open with 422, not 500 FAILED
the refusal body never carries the sanctions matched name FAILED
an unreachable sanctions service answers 503 with Retry-After, not 500 FAILED
    Expected status code <503> but was <500>.
    Expected header "Retry-After" was not not null, was "null".
4 tests completed, 4 failed
```

## What each becomes

| exception | thrown when | was | now |
|---|---|---|---|
| `AccountScreeningUnavailableException` | sanctions-service unreachable; gate fails closed, account NOT opened | 500 `INTERNAL_ERROR` | **503 `SCREENING_UNAVAILABLE`** + `Retry-After` |
| `AccountOpeningBlockedByScreeningException` | screen returns `HIT` or `REVIEW` | 500 `INTERNAL_ERROR` | **422 `ACCOUNT_OPENING_BLOCKED`** |

**503, not a 4xx**, for the same reason libs-runtime's `PolicyDecisionExceptionMapper` maps a PDP
outage to 503: an outage must read as "the platform is broken", not be laundered into a client
error. `Retry-After` is safe to advertise *here specifically* — the gate runs strictly before any
write (no account row, no idempotency record, no outbox entry) and the endpoint requires an
`Idempotency-Key`, so replaying the identical request is a genuine retry rather than a second open.

**422** for the block: a policy refusal of a well-formed request, exactly the class its sibling
`ProductNotEligibleException` already occupies three lines below it in `AccountService`.

## The disclosure — and an honest correction to how it was reported to me

The obvious fix for the second one is one word: extend `IllegalStateException`, which is what
`ProductNotEligibleException` does, and whose KDoc explicitly contrasts itself with this type as
"a bare `RuntimeException`". I did not take it.

`IllegalStateExceptionMapper` echoes `exception.message` verbatim into the response body, and
this message embeds `matched: $matchedName` — the sanctions-list name the screen hit. The one-word
fix would have published it to every caller who can provoke the refusal.

**But it is not a live exposure today, and I want to be precise about that**: the current 500 does
*not* leak the name. `GenericExceptionMapper` never echoes a message — it returns
`"An unexpected error occurred. Please contact support with traceId=…"`. Verified in the
before-fix run: the disclosure assertion failed only on the status code, never on the body. So
this is a hazard the naive fix **introduces**, not a defect it removes. Framing it as the more
serious half would overstate it; it is nonetheless the reason the mapper is written the way it is.

The mapper is therefore service-local with a fixed body, and the matched name is confined to a
`WARN` line. That log is now its **only** record — a blocked open persists no `accounts` row, so
nothing else in the system retains it. `AccountOpeningScreeningStatusIT` asserts the served body
does not contain it, so a later re-parenting cannot reintroduce the leak silently, and both KDocs
now say why (the `ProductNotEligibleException` KDoc no longer implies the sibling is unmapped).

`AccountServiceTest` already has three tests for exactly these paths and **passes against the
broken status** — it asserts the exception *type* against a mocked port, and the type was never in
doubt. Only a request through the JAX-RS boundary reaches a mapper at all. The new IT drives the
real endpoint (RestAssured + `@TestSecurity`, real Postgres via
`PostgresRedpandaRedisTestResource`).

```
:openbank-account-service:test           267 tests / 0 failures / 1 skipped
:openbank-account-service:ktlintCheck    BUILD SUCCESSFUL
:openbank-account-service:detekt         BUILD SUCCESSFUL
:openbank-account-service:quarkusBuild   BUILD SUCCESSFUL
```

The single skip is the pre-existing broker-gated `AccountEventPactProviderVerificationTest`
(`@EnabledIfSystemProperty(pactbroker.url)`), unrelated. Lint and `quarkusBuild` were run as
separate invocations after the test run, since Gradle stops at the first failing task.

### Why `TestSanctionsScreeningPort` changed

`QuarkusMock.installMockForType` cannot replace it — attempted first, and it fails at runtime:

```
Given mock, which is class ...AccountSanctionsScreeningPort$Subclass0,
is not assignable to class com.openbank.account.integration.TestSanctionsScreeningPort
```

It requires the replacement to be assignable to the resolved **bean class**, not the injected
interface. So the stub gained a settable `behaviour` (default = the previous always-CLEAR result,
so every other IT is byte-for-byte unaffected) plus a `reset()` the new IT calls in `@AfterEach` —
the bean is `@ApplicationScoped` and shared by the whole suite.

## API contract

`openapi.yaml` documents the new 503 (with the `Retry-After` header) and splits the existing
generic `422` into its two codes. Note the spec already promised a `422` on this operation that
only `ProductNotEligibleException` could serve — the screening block was a spec/implementation
divergence as well as a wrong status. Classified by the gate, not by me:

```
$ python3 .github/scripts/check-api-contract.py --base $(git merge-base HEAD origin/main) --enforce
api-contract gate: openbank-account-service: additive change, info.version 1.12.0 -> 1.13.0 — OK (required >= MINOR).
```

`version.txt` is untouched — release-please derives it from the `fix(account)` type.

## Overlap check

Open PRs scanned by changed **file**, not title (`gh api --paginate .../pulls?state=open`, 58 open).
Six touch this service:

- **#8466** (`fix/account-authorization-not-found-404`) — the sibling that surfaced these two, and
  the real overlap: it also appends to `ExceptionMappers.kt`, edits `openapi.yaml` and the threat
  model. Textual conflict at the tail of `ExceptionMappers.kt` is likely; the two mapper sets are
  disjoint, so resolution is a concatenation, not a choice.
- **#8011**, **#5986** — both move `openapi.yaml`. Together with #8466 that is **four** PRs wanting
  `1.13.0`. Whichever lands first takes it; the rest re-bump. The version here was re-read off live
  `origin/main` immediately before the commit, not off the branch base.
- **#8469** — edits `AccountService.kt` (lifecycle notifications). My change to that file is two
  KDoc blocks on exception classes at the bottom; no overlap with its call-site edits.
- **#8403** (entity defaults), **#8437** (release-please) — no shared file.

No pushed branch without a PR and no live worktree claims these files.

## Not in scope

`ADR-0158` narrates the earlier decision and records this gap as "a latent issue in that sibling
exception, untouched by this PR". That statement was true when written and an ADR is a record, so
it is left alone; the living document (the threat model) carries the 2026-09-03 entry instead.

account-service is money-path — this needs **2 approvals**. If it sits blocked on that, that is the
correct final state.

Refs #5913


JiRaska:
Merge conflict with #8466 confirmed by measurement, not prediction — tracked in #8477.

`git merge-tree --write-tree` of the two heads (#8466 `7a3a29e5c`, this branch `d32b431`): exactly
one conflicted hunk in `ExceptionMappers.kt`, both sides disjoint, so resolution is a concatenation.

The part worth flagging: `openapi.yaml` **auto-merges** and both PRs bump to `1.13.0`, so the merged
spec carries both PRs' additions under one minor bump. Neither PR goes red — each is classified
against its own merge-base. Whoever lands second re-reads `info.version` off live `origin/main` and
re-bumps. #8011 and #5986 make it a four-way race for that number.

JiRaska:
## Conflict with #8466 resolved on this branch (`ba8766d`)

`origin/main` (now `2915761e6`) merged clean first; then `fix/account-authorization-not-found-404`
(`7a3a29e5c`) merged with the one expected conflict.

**Resolution is a concatenation, not a choice** — both sides insert at the same point and are
disjoint. Verified against a pre-merge inventory of both sides rather than by reading the diff:

| | before | after |
|---|---|---|
| `@Provider` classes | 5 each side (3 shared) | **7** |
| `private const val` | 3 mine + 2 theirs (1 shared) | **4** (`UNPROCESSABLE_ENTITY`, `SERVICE_UNAVAILABLE`, `SCREENING_RETRY_AFTER_SECONDS`, `AUTHORIZATION_NOT_FOUND_MESSAGE`) |

The import block auto-merged to the union.

**The two files that did *not* conflict were checked too**, since a merge git calls clean can still
keep only one of two neighbouring list additions: `openapi.yaml` carries both the new `503` on
`POST /api/v1/accounts` and #8466's new `404` on the authorization revoke; the threat model carries
both `2026-09-03` entries.

### Gates, on the merged result

```
:openbank-account-service:test           272 tests / 0 failures / 1 skipped
:openbank-account-service:ktlintCheck    BUILD SUCCESSFUL
:openbank-account-service:detekt         BUILD SUCCESSFUL
:openbank-account-service:quarkusBuild   BUILD SUCCESSFUL

$ python3 .github/scripts/check-api-contract.py --base $(git merge-base HEAD origin/main) --enforce
api-contract gate: openbank-account-service: additive change, info.version 1.12.0 -> 1.13.0 — OK (required >= MINOR).
```

272 = 267 here + 5 from #8466. The skip is the pre-existing broker-gated
`AccountEventPactProviderVerificationTest`. `info.version` stays `1.13.0` and that is correct for
the union: the combined delta against main (`1.12.0`) is still additive, so the required bump is
still MINOR.

### One consequence, stated rather than assumed

**This branch now supersedes #8466.** Merging it first would land #8466's mappers, tests, spec entry
and threat-model entry under *this* PR's approvals — both are money-path and each was meant to earn
its own two. Two ways to avoid that, and the choice is the author's:

- **Merge #8466 first** (the intended order). This PR's diff then sheds the shared commits
  automatically and needs only a `git merge origin/main`.
- **Retarget this PR's base to `fix/account-authorization-not-found-404`**, the proper stacked
  shape — GitHub then cannot merge it before #8466, and the diff shows only this PR's changes.
  Not done here because it carries a documented irreversible hazard: merging #8466 with
  `--delete-branch` would *close* this PR, and a closed PR whose base is gone can be neither
  reopened nor retargeted (see #3055/#3063). Safe only if #8466 is merged **without**
  `--delete-branch`.

Refs #8477

JiRaska:
## Rebased onto main now that #8466 has landed (`b05f500`)

#8466 merged as `909dee547a`. This branch merged `origin/main` on top; the shared commits have
dropped out of its diff, which is now back to the six files this PR actually owns:

```
$ git diff --name-only origin/main
docs/threat-models/openbank-account-service.md
.../application/usecase/AccountService.kt
.../infrastructure/rest/ExceptionMappers.kt
.../src/main/resources/openapi.yaml
.../integration/AccountOpeningScreeningStatusIT.kt
.../integration/TestSanctionsScreeningPort.kt
```

GitHub agrees — the PR went from 8 changed files to **6**.

### The merge conflicted again, and the resolution was the opposite of last time

Squash-merge gives main a *new* commit, so git could not tell that this branch already carried
#8466's content — it re-raised the same region. But this time the `origin/main` side of the hunk was
**empty** (146 → 147, nothing between): main does not have my two mappers, and I do. So the
resolution was to keep HEAD and drop the markers — the mirror image of the concatenation last time.

That empty side is the trap worth naming: "take theirs" would have compiled, passed ktlint and
deleted both screening mappers silently. Verified by inventory rather than by reading the diff —
**7 `@Provider` classes and 4 `private const val`**, matching the union exactly.

### Version re-bump

`1.13.0` → **`1.14.0`**. #8466 shipped under `1.13.0`, the number this branch also claimed; neither
PR ever conflicted on that line, because both wrote byte-identical text. Re-read off live
`origin/main` immediately before the commit, not off the branch base:

```
$ python3 .github/scripts/check-api-contract.py --base $(git merge-base HEAD origin/main) --enforce
api-contract gate: openbank-account-service: additive change, info.version 1.13.0 -> 1.14.0 — OK (required >= MINOR).
```

The delta being classified is unchanged — the 503 on `POST /api/v1/accounts` plus the two 422 codes.
Only its starting point moved.

### Gates on the merged result

```
:openbank-account-service:test              272 tests / 0 failures / 1 skipped
:openbank-account-service:ktlintCheck       BUILD SUCCESSFUL
:openbank-account-service:detekt            BUILD SUCCESSFUL
:openbank-account-service:quarkusBuild      BUILD SUCCESSFUL

check-api-contract.py --enforce                                    1.13.0 -> 1.14.0, OK
check-threat-model-claims.py --enforce                             exit 0, 45/45 models
check-openapi-route-conformance.py --enforce --baseline <file>     exit 0, 0 NEW, 0 stale
```

The skip is the pre-existing broker-gated `AccountEventPactProviderVerificationTest`. All three gates
were run the way `.github/gates/gates.yaml` invokes them — route-conformance in particular reports 28
findings and exits 1 without its `--baseline`, so a bare run says nothing about this branch.

The threat-model gate matters here specifically: it was widened earlier today (#8429/#8460) to read
whole documents rather than STRIDE sections only, and this PR adds a change-log entry naming several
classes. It resolves clean.

Refs #5913, #8477

JiRaska:
Superseded by #8712 (merged 2026-09-05), which shipped the same fix: dedicated 503+Retry-After / 422 fixed-body mappers in ExceptionMappers.kt, WARN-without-stack logging, openapi 1.14.0, threat-model changelog, and a unit test pinning that the matched name and partyId never reach the wire. The four-case endpoint IT this PR carries (AccountOpeningScreeningStatusIT) is the one piece #8712 did not take — worth porting separately if the coverage is still wanted. Closing as superseded.

JiRaska:
**The production half of this PR is already on `main`; the integration test is the part that is still worth landing. And the version it claims has been taken.**

Measured against `origin/main` just now.

**Superseded.** `fix(account): map screening outcomes to 503/422 without leaking the match (#8512)` merged as #8712 (commit `08e80e7e86`, 2026-09-05T01:35Z) and covers **both** exceptions, not one:

```
openbank-account-service/.../rest/ExceptionMappers.kt   (on main)
  :163  AccountOpeningBlockedByScreeningExceptionMapper   -> 422
  :185  AccountScreeningUnavailableExceptionMapper        -> 503
```

I nearly reported the 503 mapper as still missing — a truncated `head -14` of that file hid line 185. It is there; the file carries 7 mappers.

`AccountService.kt`, `ExceptionMappers.kt`, `openapi.yaml` and `docs/threat-models/openbank-account-service.md` are all already changed on `main` for this defect.

**Still unique to this PR.** `main`'s coverage is `ScreeningExceptionMapperTest.kt`, a unit test that calls the mapper directly. This PR adds `AccountOpeningScreeningStatusIT` (119 lines), which drives the real endpoint. `main` has no such file — `git ls-tree -r origin/main openbank-account-service/src/test | grep -i screening` returns only the unit test and `TestSanctionsScreeningPort.kt`.

That difference matters here more than usual: a unit test that calls a resource or mapper class cannot tell a *served* route from an unserved one, which is how `POST /mcp` answered 404 on a running pod for the whole life of the endpoint (#3371). The IT is the stronger artifact and is not redundant with what merged.

**The version claim is taken.** This PR bumps `openbank-account-service/src/main/resources/openapi.yaml` `1.13.0 -> 1.14.0`; `main` is already at **1.14.0** (#8712 took it). The `openapi-version-not-taken` gate is `mode: enforced` and resolves `origin/main` at run time, so this goes red on the next run — it has not yet, because this branch is 159 commits behind and has not re-run since #8712 landed.

**Suggested shape after a rebase onto `origin/main`:** keep `AccountOpeningScreeningStatusIT` (and whatever `TestSanctionsScreeningPort` change it needs), drop the production changes that `main` now carries, and then **no `info.version` bump is needed at all** — a test-only change does not alter the contract, so the version question disappears rather than needing a new number.

Not closing this: the IT is real coverage `main` does not have. Flagging so the rebase drops the superseded half instead of re-litigating it as a conflict.

Issue #8426, opened 2026-09-03
38 documents assert a BootstrapVerifier secrets control, including a ✅ DORA row, and the class does not exist


JiRaska:
## `libs.security.BootstrapVerifier` is asserted by 38 documents and does not exist

```
$ git grep -l "BootstrapVerifier" origin/main -- '*.kt' | wc -l
0
$ git grep -l "BootstrapVerifier" origin/main | wc -l
38
```

**Control:** the same probe returns `1` for `class ServiceInfoResource`, a libs class that does exist — so the zero is a fact about the tree, not about the search.

The claim appears as a **checked compliance control**:

```
- ✅ Secrets: BootstrapVerifier blocks dev placeholders in prod (ADR 0017 Vault)

| **DORA** | Operational resilience | health probes, BootstrapVerifier, audit eventy, SLO, runbooks |
```

across five services plus libs, in both languages.

## Nothing is exposed today, and that is measured rather than assumed

The control's job is to stop a dev placeholder reaching production. I checked whether one has:

```
$ git grep -ilE "changeme|placeholder|dev-secret|CHANGE_ME" origin/main -- 'openbank-infra/gitops/components/**/*.yaml'
   10 files
```

All ten are **prose, comments, or CSS `::placeholder` selectors** — a note about config ordinals, a bootstrap image-tag comment, a `__PLACEHOLDER__` token in a realm-import template that is substituted at apply time, and three CSS rules in the Keycloak theme. **None is a secret value.** Deployed manifests take their credentials through `secretKeyRef` (71 files), so the defended-against condition does not currently occur.

So this is a **missing defence-in-depth control that documents claim exists**, not an open door. Filing it publicly on that basis; there is no exploitation detail here because there is no exploit.

## Why it still matters

A ✅ against a named control on a secrets path tells a reader the question is settled. Nobody re-checks a settled question — which is exactly how the control stayed unbuilt while 38 documents grew to depend on it. The DORA mapping row is the sharper half: it lists `BootstrapVerifier` as evidence of operational resilience in a table meant to be read by someone assessing whether the obligation is met.

## The fix is a choice, not a correction

Two honest options, and they are genuinely different:

[...]

PR #8274, opened 2026-09-02
security(payments): bind delegated payment ingress

JiRaska:
## Purpose

Completes the trusted receiver boundary for delegated domestic spend. The customer edge reaches a distinct workload-only domestic route only after a durable reservation.

## Safety

- exact customer-edge workload identity plus complete customer/delegation/reservation UUID tuple
- tuple is checked again against the durable local reservation snapshot before lookup, persistence, outbox or workflow
- owner route stays unchanged; delegated sender remains default-off until the stacked state-stream and receiver releases are live
- retryable projection/account-authority states map to 425/503; finalized absence is permanent 410

## Verification

- focused delegated service, OpenAPI contract, edge reservation, and real HTTP workload-identity tests green
- ktlintCheck, detekt, Quarkus wiring and API contract checks green

Depends on #8252 and #8247; do not activate the edge flag from this PR.

PR #8476, opened 2026-09-03
docs(governance): correct the BootstrapVerifier claim in five services' docs

JiRaska:
## What

`BootstrapVerifier` does not exist in tracked source. `git grep BootstrapVerifier -- '*.kt' '*.java'` returns **0** on `origin/main`, and [ADR-0017](https://github.com/JiRaska/open-bank-oss/blob/main/docs/adr/0017-secrets-via-vault.md) has recorded it as ⬜ Not shipped since 2026-07-17 — saying in as many words that the service documents asserting it are stale.

This is the **per-service half** of that correction: 22 documents across account, balance, lending, sanctions and security-scanner. Sibling PR **#8472** does the 12 `openbank-libs/docs/` files; **#8468** (merged/open separately) did the libs DORA Art. 9 / NIS2 Art. 21 control map. Together the three cover all 38 files that mention the symbol.

Both language variants of every document move in this PR so they cannot drift apart.

## Four different shapes of the same false claim

Each mention was read in context before rewriting rather than pattern-replaced:

| Shape | Where | Why it mattered |
|---|---|---|
| Listed as a consumed libs component | account + sanctions `02-architecture` | declares a dependency on a class that is not in that library |
| Operator prose: "startup fails on a dev placeholder" | account + balance + sanctions `05-operations` | an operator waits for a failure signal that cannot arrive |
| Annotation on a live placeholder value | lending `05-operations` | **the one place the claim is load-bearing** — it is the stated reason `POSTGRES_PASSWORD=CHANGE_ME_LOCAL_DEV_ONLY` looks safe to leave |
| DORA row + ✅ security-controls bullet | all five services' `06-compliance` | a green compliance row for a control that has never run |

## The honest replacement, measured not assumed

Every corrected line now says the class does not exist **and** names what actually holds the property. Verified against the gitops manifests on `origin/main`:

| Service manifest | `secretKeyRef` | `CHANGE_ME` / `_local_dev_only` literals |
|---|---|---|
| `accounts/account-service.yaml` | 4 | 0 |
| `balances/balance-service.yaml` | 4 | 0 |
| `lending/lending-service.yaml` | 4 | 0 |
| `sanctions-service/sanctions-service.yaml` | 7 | 0 |
| `security-scanner/security-scanner-service.yaml` | 6 | 0 |

ESO/OpenBao injection (ADR-0007) is the real mitigation. It is a **configuration property, not a boot-time control** — a deployment with a bad value is not refused, it simply fails to reach the database — and each document now says which of the two it is rather than implying a guard.

## Verification

**`mermaid-parses`** — no diagram was edited here, so this is a regression check, not a claim about new content: `check-mermaid-parses.sh` with admin-ui's own mermaid 11.17.2 over the 22 changed files reports **0 failing / 18 diagrams**. (The diagram edits live in #8472, where the same run was falsified by reintroducing a bare `@` and a literal `;`.)

**`threat-model-claims-resolve`** — `python3 .github/scripts/check-threat-model-claims.py --enforce`, exit **0**, 45/45 models. Nothing to retire: no threat model mentions `BootstrapVerifier` and no `ALLOWED_UNRESOLVED` entry excuses it. Verified rather than assumed, because that table fails on a stale entry in **both** directions.

**No release impact.** Type `docs` is a hidden type, so release-please cuts nothing regardless of the `src/main/resources/` path.

## Claim check before writing

- Open PRs scanned by changed **file**, not title (`gh api --paginate .../pulls?state=open&per_page=100`, 58 PRs, 541 files). No open PR touches any of these 22. #8274 touches `openbank-domestic-payment` docs, which carry no such claim; #8468 touches only `openbank-libs/docs/06-compliance.*`.
- `git ls-remote --heads origin` — one related branch, #8468's.
- `git worktree list` — `/private/tmp/ob-bootstrap-10707` is #8468's live worktree, left alone.

## Note for reviewers

All 34 remaining mentions turned out to be claims; **none was already a correction**. Worth flagging one near-miss: `openbank-security-scanner/.../06-compliance.*` carries a "this is where the documentation previously overclaimed (#4709)" section immediately below the line being fixed, so the file reads as self-correcting while line 66 still asserted the control. The correction there is written to sit alongside that existing section rather than duplicate it.

`openbank-lending-service` is a money-path service, so this PR is subject to the money-path review rule even though the change is documentation only.

Refs #8426

## Security checklist

      (license, CVE, source).

Each box checked against the diff rather than assumed:

- Documentation only — the diff is 12 `.md` files under `*/src/main/resources/docs/`; `git diff --name-only | grep -c src/main/kotlin` is **0**. No code, no config, no test.
- No `@Suppress`/`as Any`/`SuppressWarnings` added (grep over added lines: empty). No `build.gradle*`/`libs.versions.toml`/`package.json` in the diff. No secret-shaped additions.
- No auth, crypto or payment **code** changes — correcting prose about a control is not changing the control, so `security-review-required` is deliberately not applied. No PII path and no cardholder-data path is touched.


JiRaska:

Issue #2371, opened 2026-07-25
GDPR Art. 20 data portability is neither implemented nor decided (Art. 15 export exists and is complete — the two are being conflated)


JiRaska:
## Summary

GDPR **Art. 20 (right to data portability)** is neither implemented nor decided. It is a distinct
obligation from Art. 15 (right of access), which *is* shipped and complete — and conflating the two is
how this gap stayed invisible.

**Correction to how this was first reported.** I originally recorded the admin-ui compliance page's
"Data portability export — `warn`, API endpoint for customer data export is missing" row as the one
accurate row on that page (#2370). That was based on a broken probe: a single grep for
`portability|data-export|dataExport` over `*.kt`, which finds nothing because the code calls it
`gdpr-export`. Re-run with a known-positive control (`erasure` → 32 files, so the probe can find things),
the real picture is the opposite of what I said, and the row is wrong in *both* halves — see below.

## What actually exists — Art. 15, and it is done

`GET /api/v1/parties/{id}/gdpr-export` in `PartyResource` (line 241):

- `@Authenticated`, with an explicit three-shape caller check in the body — `ROLE_ADMIN`, `ROLE_DPO`, or
  the subject's own JWT (`isSelf`), because `@RolesAllowed` cannot express "self OR one of these roles".
- Aggregates cross-service via `GdprAggregationAdapter`: kyc-service (`GET /api/v1/kyc/cases/party/{id}`)
  and card-issuance-service (`GET /api/v1/cards/party/{id}`), best-effort so one downstream being down
  degrades the export rather than failing it.
- Audited as `party.gdpr-export` with `gdprArticle = "15"`.
- Returns JSON via `PartyGdprExport.toResponse()`.

ADR-0118 records it shipped, and issue #268 (Art. 15 export gaps) is closed. So **the compliance page's
note is false as written** — an endpoint exists — while the *obligation the row names* is genuinely
unmet. That is a third independent way that page is uncorrelated with the code, and it should be folded
into #2370.

## The actual gap — Art. 20 is a different right

ADR-0118 is the ADR that owns GDPR data lifecycle. It mentions Art. 20 and "portability" **zero times**.
Not rejected with a reason, not deferred — absent. Three concrete sub-gaps:

**1. No customer-facing route.** The endpoint permits `isSelf`, but there is no `customer-edge` proxy for
it (`git grep -in gdpr` over `openbank-customer-edge` returns only audit and consent-snapshot comments).
[...]

PR #8487, opened 2026-09-03
fix(party): make GDPR Art. 15 and Art. 20 exports reachable by the data subject

### LINKED ISSUE #8421: GDPR Art. 15 and Art. 20 exports are implemented but unreachable by any data subject: customer-edge cannot satisfy party-service's isSelf check
## The gap

`GET /api/v1/parties/{id}/gdpr-export` (Art. 15) and `/gdpr-portability-export` (Art. 20, ADR-0204) are implemented, complete, audited, and explicitly designed to accept **the subject's own JWT**:

```kotlin
val isSelf = jwt?.subject != null && jwt?.subject == partyUseCase.getPartyKeycloakSub(id)
if (!isAdmin && !isDpo && !isSelf) return Response.status(Response.Status.FORBIDDEN).build()
```

**`isSelf` can never be true for a call made through customer-edge**, so the only channel a retail customer has cannot reach either right.

## Why

`UpstreamClient`'s own KDoc states the mechanism:

> the customer JWT (openbank-customers realm) is validated at the edge but **NOT forwarded** to upstream services that validate against the operator `openbank` realm. This client fetches a service-account token via client_credentials flow (operator realm) […] The customer party ID is passed via `PARTY_HEADER` so upstream services can scope data independently.

So at party-service `jwt.subject` is `service-account-openbank-edge`, never the data subject. The edge principal is ROLE_OPERATOR — neither ROLE_ADMIN nor ROLE_DPO. All three branches are false; every proxied call is a 403.

Verified exhaustively: customer-edge has **no** route for either export (161 `@Path` declarations, no `gdpr`/`portability`/`close` among them), and the published spec exposes only `/privacy/access-log` in this area.

## Why it matters more than a missing route

Art. 15 and Art. 20 are rights with statutory deadlines, and ADR-0204 decided the Art. 20 shape deliberately (redaction rules, no Art. 20(2) direct transmission). The implementation honours that decision — but **no data subject can exercise either right through the product**. The right is real, the code is right, and the path is missing.

The customer app already models this honestly rather than papering over it: `GdprApi` lodges a *tracked request* through the dispute-service regulatory intake (ADR-0085 deadline clock), and its KDoc names this exact blocker. The copy says "request an export … we will handle it by <date>", never promising an instant download. That is the correct interim, not a fix.

## What needs deciding (not obvious, hence an issue rather than a PR)

The mechanical fix — have party-service accept `X-Customer-Party-Id` as the self-identity for these two reads, as party-scoped reads elsewhere in the fleet do — is a **larger grant than it looks**. Those other reads return one account's data; these return the full PII set, including KYC and card records aggregated from two more services (ADR-0118 §6). Trusting the header here means any caller holding the edge's M2M token can export **any** subject's complete PII by changing one header value.

Options, none of which an implementation should pick unilaterally:

1. **Party-scoped self-export at party-service**, keyed on the header, accepted only from the edge principal — smallest change, largest blast radius if the edge is ever wrong.
2. **Token exchange / on-behalf-of**, so party-service still sees a subject JWT and `isSelf` keeps meaning what it says — preserves the existing control, costs an exchange step the fleet does not do today.
3. **Asynchronous fulfilment**: the edge lodges the request, a DPO-authorised job produces the export and delivers it through the existing secure-document channel — keeps the strong control, matches the tracked-request model the app already implements, and is probably closest to how a real bank does subject access at volume.

Option 3 also composes with retention (ADR-0118) and needs no new trust in the header.

## Not in scope here

Customer-initiated **account closure** is a separate question and deliberately left alone: ADR-0270 is `decision-status: proposed` / `delivery-status: planned` with no delivery repos, and it designs *bank-initiated* termination. `POST /api/v1/accounts/{accountId}/close` exists but is ROLE_OPERATOR/ADMIN, single-account, refuses a non-empty account, and its own KNOWN LIMITATIONS record that the balance read is point-in-time rather than a lock and cannot see money already in flight. Wiring a customer-facing "close my account" onto that today would be exactly the improvisation ADR-0270 exists to replace.

Found while auditing what the customer app can and cannot reach.

JiRaska:
Closes #8421

## The premise, verified independently

The issue is correct, and it is **two** of the four unreachability shapes at once, in two different services. Both had to be fixed for either to matter.

**1. No route registered at all — customer-edge.** 136 `@Path` declarations, none for either export; the published spec carried only `/privacy/access-log` in this area. ADR-0204 **D6** deliberately left "who gets a button for it" to the surface owners rather than deciding against it, so this is a gap the ADR anticipated, not one it forbade.

**2. Reachable only by a principal the customer channel cannot present — party-service.** Both handlers accept `ROLE_ADMIN`, `ROLE_DPO`, or the subject's own JWT. `UpstreamClient` validates the customer token in the `openbank-customers` realm and does **not** forward it: every upstream hop carries the edge's own `client_credentials` token from the operator realm plus `X-Customer-Party-Id`. `realm-template.json` gives `service-account-openbank-edge` **`ROLE_OPERATOR` alone**. All three branches false ⇒ 403 on every subject-initiated export.

Two further facts measured while checking, which shaped the fix:

* **`ROLE_DPO` is held by nobody.** The string appears in exactly three places in the tree — this handler, its spec description, and ADR-0204's prose. No realm JSON defines the role and **no allow reason in `rest.rego` grants it**.
* **The blast-radius objection is smaller than it reads.** `GET /api/v1/parties/{id}` and `GET /{id}/documents` are already `@RolesAllowed(… ROLE_OPERATOR …)`, so anything holding the edge's M2M token can already read any party's record **by id, with no header at all**. A header-keyed route restricted to the named edge principal is strictly narrower than what exists today. The genuine increment is the KYC and card aggregation — which the edge already proxies at `/kyc` and `/cards`, keyed on the same header.

Not a `@Path`-binding defect: the annotations sit directly on the class and on each method, and the party-service routes answer (403, not 404).

## What this does — the issue's option 1, which is the fleet's settled trust model

`PartyResource` already states the position for the identical shape of read: *"the caller is ALWAYS the customer-edge's M2M service identity, which has already resolved `[id]` from the customer's own JWT — the edge is the trust boundary, not this path"* (`/{id}/consent`, `/{id}/payees`). `rest.rego`'s `edge-service-audit-customer` says the same for the sibling privacy-centre route, including that the identity match is the load-bearing half because real staff carry `ROLE_OPERATOR` too. Seven services already implement this pattern.

**party-service** gains a fourth accepted caller shape: the configured edge principal presenting an `X-Customer-Party-Id` that names the **same** party as the path. `callerIsCustomerEdge()` takes no request data, so the permission decision stays clean of it (CodeQL `java/tainted-permissions-check`; the same split lending-service's `CustomerCreditJourneyResource` documents). The header parameter is declared **nullable** — JAX-RS injects `null` for an absent header and a non-nullable declaration would make the absent case a 500 before the body ran.

**customer-edge** gains `GET /privacy/gdpr-export` and `GET /privacy/portability-export`. Two routes and two actions, not one: Art. 15 and Art. 20 are different rights with different output obligations (ADR-0204 D1 excludes Art. 6(1)(c) data and adds transaction history), and party-service audits them under different `gdprArticle` codes. Neither takes a path or query parameter, so a customer can only ever export their own record.

**The Art. 30 record gains a `channel` field.** `actorId` is the edge's service account for every subject-initiated export, so without it a subject exercising their own right and a staff member reading their file are indistinguishable in the trail.

## No new `@Authorize` on party-service — and the reason is measured, not assumed

The two exports serve `ROLE_DPO`, which no reason in `rest.rego` grants. Adding a PDP gate would deny the compliance path the day `AUTHZ_ENFORCE` flips, while `ROLE_ADMIN` sailed through — the worst kind of regression, silent and role-specific. `opa eval` against the materialised customer-edge bundle (`rules-data.yaml` → `rules/data.yaml`), must-ALLOW and must-DENY in the same run:

```
OK    must-ALLOW  customer -> customer.gdprExport.read  (NEW)                got=ALLOW reason=customer-self-action
OK    must-ALLOW  customer -> customer.portabilityExport.read  (NEW)         got=ALLOW reason=customer-self-action
OK    control     customer -> customer.profile.read  (/privacy/access-log)   got=ALLOW reason=customer-self-action
OK    must-DENY   ANONYMOUS -> customer.gdprExport.read                      got=DENY
OK    must-DENY   customer -> gdprExport.read (outside the customer. prefix)  got=DENY
--- why party-service got NO new @Authorize action ---
OK    must-DENY   ROLE_DPO -> hypothetical party.gdprExport.read             got=DENY
OK    control     ROLE_ADMIN -> the same hypothetical action                 got=ALLOW reason=operator-read-any
```

The two new edge actions are granted by the existing `customer-self-action` prefix rule, so **no `.rego` or `rules.yaml` change and no bundle restamp**.

## Tests, falsified in both directions

`GdprSubjectAccessViaEdgeIT` (party-service, real HTTP + Testcontainers) — 6/6 green. Reverting only the predicate to its pre-fix form:

```
FAIL  the customer edge may fetch the Art 15 export …        Expected <200> but was <403>
FAIL  the customer edge may fetch the Art 20 portability …   Expected <200> but was <403>
PASS  a header naming a different party cannot export …
PASS  the edge principal without a party header is refused
PASS  a ROLE_OPERATOR caller that is not the edge is refused even with a matching header
PASS  an admin still exports without any header (unchanged-behaviour control)
```

The refusals passing in both runs is the point: the permissive assertion alone would also pass against a handler that had simply stopped checking. The fifth test is the only one that would notice if the identity match were dropped.

`GdprSelfServiceExportIT` (customer-edge, real HTTP + `StubUpstreamResource`) — 4/4 green. Reverting `CustomerEdgeResource` to `origin/main`: 3 of 4 fail (both GETs 404 with *"Unable to find matching target resource method"*, the POST probe 404 instead of 405) while the known-negative control still passes.

**A note on the oracle, because the first version of that test was wrong.** It asserted "not 404". `UpstreamClient` proxies the upstream status verbatim, so an unstubbed party-service path answers 404 exactly like an unregistered route — `/customer/v1/profile`, shipped long ago, reads 404 in the test JVM for that reason alone. The 404-vs-405 discrimination on a POST is what actually separates "unregistered" from "wrong method", and it needs no upstream at all.

## Also corrected

Two compliance records this change makes materially more wrong:
* the edge's own GDPR mapping listed portability as *"upstream controllers; the edge holds no records"*;
* admin-ui's compliance page still said the Art. 20 endpoint *"is not built yet — issue #2371"* — already false on `main` since ADR-0204 shipped on 2026-08-18, and false a second way now.

## Deliberately NOT in scope

* **The legal position.** Retention periods, controller identity and lawful basis are a data-controller judgement. ADR-0204 itself flags that the Art. 20(4) IBAN-redaction level and the per-field basis classification *"should be confirmed with counsel"*; nothing here changes either.
* **Options 2 (token exchange) and 3 (asynchronous DPO-mediated fulfilment)** from the issue. Both remain open and both are better long-term answers; neither is a prerequisite for making the right reachable, and option 3 in particular composes with ADR-0118 retention as its own piece of work.
* **A client for these routes.** The app's tracked-request flow through the dispute-service regulatory intake (ADR-0085) stays as it is — this PR gives it something real to call, it does not change the app.
* **Account closure** — ADR-0270 territory, left alone as the issue asks.
* **ADR-0204 needs no edit.** D6 delegated this choice rather than deciding it, so nothing in the ADR is now false.

## Verification

`compileKotlin`, `compileTestKotlin`, `ktlintCheck`, `detekt` green on both modules. Full suites, clean `build/test-results`, against the same commands run on an `origin/main` worktree as the control:

| | branch | origin/main |
|---|---|---|
| party-service | 187 tests, **1** failure, 47 skipped | 181 tests, **1** failure, 41 skipped |
| customer-edge | 424 tests, **0** failures, 1 skipped | 420 tests, 0 failures, 1 skipped |

The single party-service failure (`PartyPactFolderProviderVerificationTest :: openbank-kyc-service - a PARTY_CREATED event`) is **pre-existing and order-dependent** — present in `main`'s full-suite run, absent when that class runs alone on `main`. The Testcontainers ITs, mine included, skip together in a full-suite run on this machine and pass when run alone; the skip delta is exactly my 6 tests.

`api-contract` classifies both spec changes as additive/MINOR (party `1.20.0 → 1.21.0`, edge `1.49.0 → 1.50.0`, both re-read off live `origin/main` immediately before pushing); `openapi-version-not-taken`, `nonnull-jaxrs-param-ratchet`, `check-configproperty-supplied`, `four-eyes-verb-grantable`, `authz-enforce-pdp-sidecar-parity`, `no-dead-code-service-principal-rego-rule`, `openapi-enum-domain-drift` and `check-domain-purity` all pass with no new findings.

⚠️ `openapi-route-conformance` is **red on `origin/main` already** — 28 findings across sca-service, security-scanner, customer-edge pockets/webauthn and others. Byte-identical finding sets on this branch and on a clean `origin/main` worktree, and none of them names either new route.


JiRaska:
### ⚠️ `openapi.yaml` version race — #8011 claims the same number

Measured just now against live `origin/main` (`openbank-customer-edge`, `info.version` = `1.49.0`):

| PR | claims | note |
|---|---|---|
| **#8487 (this)** | `1.50.0` | |
| **#8011** | `1.50.0` | **collision** — whichever of us merges second must re-bump to `1.51.0` |
| #8321 | `1.51.0` | fine as long as it lands after both of the above |
| #8274 | `1.48.0` | *below* the current tip — stale, will fail its own gate until rebased |

`openapi-version-not-taken` classifies against the base tip **as of the run**, so a branch whose last CI run predates the competing merge is invisible to it (ADR-0048 chose detection over prevention; there is no merge queue on a user-owned repo). If #8011 lands first, re-read the version off `origin/main` and re-bump here rather than trusting this PR's green.

`openbank-party-service` (`1.20.0 → 1.21.0`) has no competitor — no other open PR touches that spec.

JiRaska:
**Correction to one line in the report above: `openapi-route-conformance` is NOT red on `origin/main`.**

Measured on a clean detached `origin/main` worktree, both ways:

| invocation | exit | output |
|---|---|---|
| `--enforce` | **1** | 28 `::error` lines |
| `--enforce --baseline .github/openapi-route-conformance-baseline.txt` | **0** | `52 service(s) compared, 28 route-level finding(s), 28 declared in the baseline, 0 NEW, 0 stale` |

`gates.yaml` runs the second form. The baseline file holds exactly 28 entries, so the 28 findings are the declared backlog, not a regression — and the gate reports `0 NEW, 0 stale`, i.e. it is doing its job in both directions.

Confirmed independently from CI rather than only by re-running: the latest completed `ci.yml` run on `main` (33754336380) shows all three `gates (…)` shards **SUCCESS**.

This is the same mistake class twice in one session, so it is worth stating as a rule rather than a one-off: **run a gate the way `gates.yaml` runs it, flags included.** Earlier today `check-threat-model-claims.py` returned exit 0 against a deliberately sabotaged tree because I ran it without `--enforce` (it ends `return 1 if a.enforce else 0`); here the missing flag is `--baseline` and the error points the opposite way — a clean tree reading as 28 failures. Read the `run:` block before trusting an exit code.

Nothing else in the report is affected: the two real findings — that `ROLE_DPO` is granted by no realm and by no `rest.rego` reason, and that `GET /api/v1/parties/{id}` is already `ROLE_OPERATOR`-readable so the header-keyed route is strictly narrower than what ships today — both stand and were measured directly.

JiRaska:
Sharpening my correction above — a peer session measured a third combination I had not, and it is the one that matters most. Re-measured here on a clean detached `origin/main` worktree:

| invocation | exit | output |
|---|---|---|
| bare, no flags | **0** | 0 errors, **28 `::warning::`**, `advisory — no --enforce, so this is a warning` |
| `--enforce` | **1** | 28 `::error::` |
| `--enforce --baseline .github/openapi-route-conformance-baseline.txt` | **0** | `28 route-level finding(s), 28 declared in the baseline, 0 NEW, 0 stale` |

Two flags doing two different jobs: **`--baseline` clears the 28; `--enforce` makes findings fatal. Neither alone reproduces the gate.**

The bare run is the dangerous state, more so than the one I described. It **exits 0 while printing 28 NEW findings**, so whoever checks the exit code concludes "clean" and whoever reads the log concludes "main is broken", and both are wrong. My earlier table described the `--enforce`-only run, which is what I had measured — accurate for that invocation, but it left the impression that a casual run looks red. It looks green and says red.

Conclusion is unchanged and now rests on three measurements instead of two: `main` is not red on this gate, and the latest completed `ci.yml` run on `main` shows all three `gates (…)` shards SUCCESS.

Issue #2911, opened 2026-07-31
M2M marketing consent REVOKE is structurally denied — the OPA rule cannot fire (grant works, revoke 403s)


JiRaska:
## Summary

The ADR-0206 exception that lets party-service forward the mobile app's marketing-consent toggle
works for **granting** and is structurally impossible for **revoking**. A customer turning marketing
e-mail ON succeeds; turning it OFF always gets 403.

Found while preparing the #2749 smoke journey, where the revocation step is the whole point.

## Evidence

Measured against the sandbox with the shared M2M principal
(`service-account-openbank-services`, `ROLE_API` + `ROLE_OPERATOR`):

| Operation | `@Authorize` resource | Result |
|-----------|----------------------|--------|
| `POST /api/v1/consents` (consent.grant) | `#request.granteeId` | **201 Created** |
| `DELETE /api/v1/consents/{id}` (consent.revoke) | `#id` | **403** |

The rule both paths depend on (`consent_rest_ext.rego`):

```rego
allowed_reasons contains "service-consent-m2m-marketing" if {
	input.principal.type == "HUMAN"
	input.principal.id == "service-account-openbank-services"
	input.action in {"consent.grant", "consent.revoke"}
	input.resource.id == "party-service:marketing-comms"
}
```

It requires `input.resource.id` to equal the fixed grantee. On grant, the annotation binds that from
`#request.granteeId`, so it matches. On revoke, the annotation is `resource = "#id"` — the consent
UUID — which can never equal `party-service:marketing-comms`. The `consent.revoke` half of the rule
is unreachable, and the caller falls through to deny.

`consent-service` runs with `AUTHZ_ENFORCE=true`, so this is a live denial, not advisory.

## The comment says the opposite

`ConsentResource.revoke` carries:

> the M2M path (ADR-0206 D2): consent-opa-bundle.yaml's service-consent-m2m-marketing rule checks it
> via `#granteeId` directly (no dotted path needed here — it's already a top-level parameter)

That describes the intent. The annotation on the very next lines says `resource = "#id"`. Passing
`?granteeId=party-service:marketing-comms` on the request — as I did — does not help, because the
annotation, not the query string, decides what `input.resource.id` becomes.

## Why it matters beyond the rollout

GDPR Art. 7(3): withdrawing consent must be as easy as giving it. Here giving it works and
withdrawing it is structurally impossible through the same channel.

It is also the same defect class the repo already documents for
[...]

PR #8489, opened 2026-09-03
fix(consent): wire the OPA path for the edge's ADR-0269 credit-consent switch

JiRaska:
## What this is

Live e2e verification of the just-merged `#6235` (ADR-0269 platform) against sandbox found the
customer's credit-consent switch was dead: `PUT /customer/v1/credit/consents` returned 200 with
the state unchanged, no matter what was requested. Confirmed with a real token-exchange customer
session and a direct read of `consent-service`'s database (zero rows for the party after a grant
"succeeded").

## Root cause — two stacked bugs

1. **consent-service's OPA bundle never granted `consent.grant`/`consent.revoke` to the edge's own
   M2M identity** (`service-account-openbank-edge`). `operator-consent-write` explicitly excludes
   it; the only M2M exception (`service-consent-m2m-marketing`) is scoped to a different client id
   (`service-account-openbank-services`) and a different grantee
   (`party-service:marketing-comms`). Every grant 403'd upstream — silently, because
   `CustomerEdgeResource.grantCreditScope` fires the POST and never checks the response.
2. **`revokeHeldConsent` never sent `granteeId`** on the DELETE. `consent.revoke`'s OPA
   `resource.id` binds to exactly that query param (issue #2911 fixed the identical shape for the
   marketing path, but the credit path was never touched). So turning the switch back off would
   have 403'd too, once (1) is fixed.

## Fix

- New OPA rule `service-consent-m2m-credit`, scoped the same way as the existing marketing
  exception: fires only for the edge's client identity AND only for its fixed first-party grantee
  (`"openbank"`, the `BANK_GRANTEE` constant) — does not turn the edge into a general grantor.
- `revokeHeldConsent` now sends `granteeId`.
- Rego tests **falsified**: removed the new rule, confirmed exactly the two new positive tests
  fail (26/28 without it, 28/28 with it).
- Tightened the existing Kotlin revoke test — it asserted only `delete()` was called with a URL
  containing `/api/v1/consents/`, never that `granteeId` was in it. That's how the missing
  parameter shipped unnoticed in #6235.
- All 10 `CustomerEdgeCreditConsentTest` cases pass with the fix; verified from the JUnit XML
  (`tests="10"`), not the Gradle exit code.

## Scope

4 files, both layers of the same feature. Not money-path (GDPR-consent switch, not a
funds-moving action) but it is an authz-policy change to consent-service's OPA bundle, so please
review the rule scoping carefully — it's a narrow resource-scoped exception mirroring an existing
pattern, not a role-only grant.


JiRaska:
## Correction: the second bug described in this PR was not real

Verified live against sandbox after this merged and deployed. One claim in the PR body and commit
message is wrong and I want it on the record rather than left standing.

### What is confirmed working

The consent switch is live. `PUT /customer/v1/credit/consents {"offers":true}` now produces a real
row in consent-service — `ACTIVE`, scope `CREDIT_OFFERS`, grantee `openbank` — verified by querying
the database directly rather than trusting the HTTP 200, which is what lied before this fix. Turning
it off transitions the row to `REVOKED` with a timestamp. Bug 1 (the missing OPA path,
`service-consent-m2m-credit`) was real and is fixed.

### What I got wrong

The PR claimed the missing `granteeId` on `revokeHeldConsent` was a second bug, and that revoke
"would have 403'd too, once (1) is fixed."

That is false. I tested revoke against the **old** customer-edge image — `sandbox-b1777d2c`, which
does not contain the `granteeId` change, because the gitops image bump (#8701) is still open — and
it succeeded. The reason is `edge-service-consent` in `openbank-libs/governance/policies/rest.rego`:

```rego
allowed_reasons contains "edge-service-consent" if {
	input.principal.type == "HUMAN"
	input.principal.id == "service-account-openbank-edge"
	input.action in {"consent.list", "consent.revoke"}
}
```

That base rule already grants the edge `consent.revoke` unconditionally, with no resource
comparison. So the credit revoke path was never blocked, and `service-consent-m2m-credit`'s
`consent.revoke` half is currently redundant with it.

I should have run that case against the deployed state before describing it as a bug, instead of
reasoning from the OPA rules alone.

### What the `granteeId` change is actually worth

Still worth keeping, but as defense in depth rather than a fix:

- `ConsentService.revokeConsent` cross-checks the passed `granteeId` against the loaded consent's
  actual `granteeId` before revoking. Sending it means that check can do its job on this path.
- It stops the credit path depending on `edge-service-consent` staying as broad as it is today. If
  that rule is ever narrowed to add a resource comparison — a reasonable thing for a least-privilege
  review to want — the credit path keeps working instead of breaking silently.

No code change is proposed here; this comment exists so the merged rationale is not taken at face
value by a later reader or an audit of this rule.

PR #8491, opened 2026-09-03
feat(notifications): tell customers when a third party gains or loses account access

JiRaska:
Third of the #8432 sweep, after #8439 (KYC) and #8469 (accounts). Closes two more of the twelve declared-but-never-produced notification templates: `CONSENT_GRANTED` and `CONSENT_REVOKED`.

## What was wrong

Both templates have been declared, rendered, given required variables and classified since notification-service was written — and **nothing ever emitted either**. A TPP could be granted ninety days of access to a customer's balances and transactions, and the customer was told nothing. `GET /customer/v1/consents` reports it accurately, but only to someone who thinks to go and look. That is the wrong way round for the one consent event a customer most needs to notice they did not make.

## A consumer, not a producer — and why that differs from #8439/#8469

Those two added direct emitters. For kyc-service that was forced — it has no domain-event module at all. **For account-service the wording in an earlier version of this description was too strong, and is corrected here:** `AccountCreatedEvent` / `AccountStatusChangedEvent` / `AccountClosedEvent` do exist. What it lacks is a clean lifecycle topic to consume — both its outbox channels publish to `openbank.accounts.account.created`, a creation-named topic, and `ACCOUNT_FROZEN` would have to be recovered from a generic `AccountStatusChangedEvent` — and it already had a `notification-requests-out` channel and ACL, so the emitter was the established path there rather than a new one. Whether #8469 would be better re-cut as a consumer is a fair question, and separable from this PR.

consent-service is the clean case: it already publishes the full lifecycle onto `openbank.consent.events` through its **transactional outbox**, and campaign-service and audit-service already read it.

So this adds `ConsentNotificationConsumer` as a third reader of a live topic. The gain is real, not stylistic: the notification now rides a record written in the same transaction as the consent change, so it cannot be lost the way a best-effort emit alongside the commit can. No new producer, no new producer-side ACL, no change to the outbox contract.

## Only account-access consents notify

`ConsentScope` mixes PSD2/agent access scopes with pure data-processing preferences — marketing channels, RUM telemetry, credit-offer processing — and **both kinds raise these same two events**. The templates say access to *account data* was granted. Firing them for a marketing toggle would be untrue, and noise the customer generated themselves seconds earlier.

The decision is computed **in consent-service**, from its own `Consent.GDPR_ONLY_SCOPES` — already the canonical set that decides SCA exemption, guarded disjoint from `AISP_SCOPES` at class-load — and published as `accountAccess`. Deliberately not re-derived in the consumer: a second copy of that list is exactly how the next preference scope added to consent-service would silently start reading as account access downstream. It is a Jackson-serialised computed getter, so there is no constructor parameter to set wrong at a call site, and nothing deserialises these events back into the class (consumers read `JsonNode`).

The consumer **fails closed** on the flag: an event without a boolean `accountAccess` is dropped, not guessed. Either default is a wrong customer outcome — `true` pushes account-access wording at a preference, `false` hides a third party gaining access — and it is only reachable for records in flight during the rollout.

## Deliberate non-scope

- **`ConsentSuperseded` does not notify.** It is not a withdrawal — access continues under a newer consent for the same grantee and scopes (#6487) — so saying access ended would be false.
- **`ConsentExpired` / `ConsentRejected` do not notify.** No template fits, and both need a customer-facing decision rather than a mapping.
- **No deep link.** `openbank://consents` is not in `MobileDeepLink`'s closed allow-list and no app route was verified, so the push carries none rather than a tap that goes nowhere.
- **Grant means ACTIVE, not requested** — consent-service raises `ConsentGranted` only from `persistActivation`, so there is no PENDING_SCA push.

## `auto.offset.reset=latest`, unlike the other three channels

This is a new consumer group on a topic that has been live for months. At `earliest` — what the other three channels use — it would replay the entire retained history and push *"a third party was granted access to your accounts"* for every consent ever granted, including ones long since revoked. A notification is about something happening now.

No DLQ was added, matching `delegation-events-in`: no `failure-strategy` is configured because `NotificationConsumer.consume` recovers every failure internally and this consumer swallows poison pills, so nothing can reach one. An ACL and topic for a DLQ that cannot fire would be decoration.

## Verification

- `ConsentNotificationConsumerTest` — **10 tests, 0 failures**
- `ConsentEventsTest` — **9 tests, 0 failures** (3 new), including one that derives from `GDPR_ONLY_SCOPES` itself rather than a copied list, so it proves the no-drift property instead of restating it
- **Guard falsified:** deleting the `accountAccess` filter from the consumer turns the suite RED on exactly `a data-processing preference raises no notification` and `an event with no usable accountAccess flag is dropped, not guessed` — those two and nothing else. Restored and re-run green.
- `ktlintCheck` + `detekt` on both modules with `--rerun-tasks`: clean.
- Full module suites: one pre-existing environmental failure, `DeviceTokenSweepCronIT` — nothing listening on localhost:5432, and its `%test` profile points at `openbank_notification_it` with devservices off. Unrelated to this change; CI has the warm stack.

## Infra in the same PR

`kafka-notification-mtls.yaml` gains `Read, Describe` on `openbank.consent.events` and `Read` on the `notification-service-consent` group; `notification-service-msg-override.yaml` gains the channel's `group.id` / `client.id` / `auto.offset.reset` as properties, not YAML — dotted leaf keys are silently dropped by Quarkus's YAML source (#686). Without the ACL every poll gets a silent `GroupAuthorizationException` and **no unit test can see it**.

## Remaining after this

Of the twelve unproduced templates, seven are left: `TRANSACTION_FAILED` (owner unclear), `KYC_DOCUMENT_REQUIRED` (no transition to fire it from — see #8439), and `OTP_CODE` / `PASSWORD_RESET` / `WELCOME`, which likely belong to Keycloak rather than here. That is a decision to take, not code to write.

## Security checklist


JiRaska:

```
```

```markdown
## Security checklist

      (license, CVE, source).
```

PR #8503, opened 2026-09-03
fix: send occurredAt on the last four audit-consumed producers that omit it (#8352)

### LINKED ISSUE #8352: Audit trail: topic-by-topic disposition of event time vs ingest time (7/21 topics without event time)
## Context
#3883 established that `AuditConsumer` substitutes ingest time whenever a producer sends no event time, and 7 of its 21 topics do so — the stored row then asserts a business time it never measured. #3907 fixed part of it. The full topic-by-topic disposition is not tracked anywhere.

## Acceptance criteria
- [ ] All 21 audit-ingested topics enumerated here with: envelope type, event-time key (`occurredAt` vs `timestamp` vs `at` vs none), producer disposition
- [ ] For each topic without event time: either producer fixed to send business time, or consumer records ingest time under an explicit `ingestedAt` field (never masquerading as business time)
- [ ] Producers enumerated by reading the SERIALISED TYPE or a real message payload — never by grepping quoted field names (the fleet has two payload idioms; a string grep sees only hand-built maps)
- [ ] Audit doc updated: which fields are measured vs substituted

Audit: docs/audit-2026-09-03-top50.md (B17). Refs: #3883, #3907.

JiRaska:
## What this is

The `#8352` deliverable: the topic-by-topic disposition of event time vs ingest time on
audit-service's `audit-events-in` channel, plus the four bounded producer fixes it turned up.

The full 27-topic table, with the method used per row, is [in a comment on #8352](https://github.com/JiRaska/open-bank-oss/issues/8352).

## The four fixes

`AuditConsumer.eventTime` reads `occurredAt` and **only** `occurredAt`; absent or unparseable ⇒ the
row stores the consumer's clock and is flagged `occurred_at_source = INGEST`. Four event types had
no such key, and **every one of them already had the instant in hand and simply never projected it
onto the wire** — which is what makes these bounded rather than a design call:

| event type | what it carried | what it now sends as `occurredAt` |
|---|---|---|
| `dispute.opened` | `openedAt` (an `OffsetDateTime`) | `dispute.createdAt.toInstant()` |
| `PARTY_ERASED` | `erasedAt`, while `PartyEvent.occurredAt` held the same value and fed the outbox row's `createdAt` | the same instant, projected into the envelope |
| `interest.withholding.recorded.v1` | nothing temporal but `LocalDate` period bounds | `WithholdingTax.createdAt.toInstant()` |
| `interest.withholding.remitted.v1` | nothing temporal but a `LocalDate` `dueDate` | `WithholdingRemittance.createdAt.toInstant()` |

**Nothing was invented, and nothing was left as an honest INGEST.** The issue allowed for a topic
that genuinely has only an ingest moment; on this channel that case does not exist — all 27 topics
now carry a real business instant on every production-reachable event type.

The two fields that *look* like an answer on the interest events are not: `periodFrom`/`periodTo`
are the range the tax was computed **over**, and `dueDate` is the date the obligation must be met
**by**. Neither is a moment anything happened.

All four changes are **additive** — every existing field keeps its name, place and form, and each
test asserts that too, because a test looking only at the new key cannot tell an addition from a
rename. `.toInstant()` is load-bearing rather than tidiness in three of the four: those fields are
`OffsetDateTime`, and a bare rename would put an offset form on the wire.

## Falsification

Every new assertion was run red against the unfixed producer first, and the failure text is the
real payload:

```
the opened outbox payload carries the opening instant as the audit event time
  [audit occurred_at_source for payload {"eventType":"dispute.opened",...,"openedAt":"2025-01-15T10:00Z","sourceService":"dispute-service"}]
  expected: EVENT but was: INGEST

PARTY_ERASED carries the erasure instant as the audit event time
  [... {"eventType":"PARTY_ERASED","partyId":"...","erasedAt":"2026-06-11T08:00:00Z","sourceService":"party-service"}]
  expected: EVENT but was: INGEST

PARTY_ERASED carries exactly five keys - a time, not an identifier, was added
  actual ["eventType","partyId","erasedAt","sourceService"] — could not find ["occurredAt"]

the withholding-recorded payload carries the recording instant as the audit event time
  [... "periodFrom":"2026-01-18","periodTo":"2026-01-20",...,"status":"RECORDED"}]
  expected: EVENT but was: INGEST

the remitted payload carries the assembly instant as the audit event time
  [... "dueDate":"2026-02-28","status":"PENDING"}]
  expected: EVENT but was: INGEST
```

Green after the fix. Each test asserts the **exact** instant, not merely that an event time is
present: a blanket `Instant.now()` at serialisation would satisfy presence while asserting a
business time nothing measured — indistinguishable in the ten-year trail from a real one, and so
worse than an honest INGEST row.

The assertions go through `AuditEventTime` (`openbank-libs-testing`), the one shared copy of the
rule the consumer applies. interest-service gains that test dependency rather than restating the
rule locally; a hand-kept second copy of an agreement between modules moves with the first and keeps
passing against a contract neither side honours.

## A test that could not see the defect

`PartyEventEnvelopeContractTest` built its mapper with `findAndRegisterModules()` alone, which
leaves `WRITE_DATES_AS_TIMESTAMPS` **on**, so every `Instant` in it serialised as the number
`1.7866E9` — a wire shape no consumer of that topic ever receives, since Quarkus defaults
`quarkus.jackson.write-dates-as-timestamps` to false. The file's own comment claimed it mirrored the
running producer. Nothing failed, because every time assertion in it was `json.has(...)`, true of a
number as readily as of an ISO-8601 string. `PartyChangeMaterialityContractTest`, in the same
package, already had the correct mapper and says why. Corrected here, and the `erasedAt` assertion
upgraded from `has` to the value — this is the same trap that cost #3926 a red run on
document-service.

## The stale comment

`AuditConsumer.eventTime`'s KDoc named the seven topics without an event time. That list was wrong
within hours of being written: it said clearing, dispute, statement, sanctions, six of lending's
payloads, sepa-payment's Temporal path and document-service, and the #3914/#3926 sweep had already
fixed all of those except a `dispute.opened` builder introduced by #4087 **three hours before that
sweep merged** — so the sweep's branch could not see the sibling it was about to acquire. Meanwhile
the subscription grew from 21 topics to 27 and the comment could not know.

A comment naming other services' current payload shapes is a claim with a shelf life and nothing
re-checks it. The KDoc now states the mechanism only and points at the three things that cannot go
stale: `openbank.audit.event.time.missing{source_service}` as a series, the per-row
`occurred_at_source` flag, and the disposition now written down in
`openbank-audit-service/CLAUDE.md`.

## Method, and why a grep would have lied

Producers were enumerated by reading the **serialised type** or the construction site, never by
grepping the quoted key. This channel carries **four** payload idioms:

- a serialised `DomainEvent` subclass — cards, delegation, ledger, consent, account, transaction,
  balance, domestic-payment, fx, document. `occurredAt` is on the wire with **no literal anywhere**;
- a `mapOf` / `ObjectNode` envelope — party, kyc, clearing, sca, swift, sepa-instant, customer-edge,
  security-scanner;
- a raw JSON string template — dispute, statement, lending, sdd, interest, fraud, and
  sepa-payment's Temporal activities;
- sanctions' hybrid: serialise the aggregate, then `put` the per-event instant.

`grep '"occurredAt"'` sees only the last two families, so it reports a confident false gap over the
first — the #3883 lesson, applied.

## Contracts and compatibility

- No new topic and no new contract file, so `event-contract-coverage-ratchet` is unmoved in both
  directions. None of dispute-, party- or interest-service has an `asyncapi.yaml`; all three pairs
  stay grandfathered in `.github/event-contract-baseline.txt`.
- Additive JSON keys only; no constructor changes. `check-event-schema-compat.py --base origin/main`
  is clean, and a consumer that ignores the new key is unaffected during rollout.

## Verification

```
:openbank-dispute-service:test   58 tests, 0 failed, 0 skipped
:openbank-party-service:test    183 tests, 0 failed, 1 skipped (the broker-gated pact class)
:openbank-interest-service:test 114 tests, 0 failed, 0 skipped
:openbank-audit-service:test --tests '*AuditConsumerTest*'  green
ktlintCheck + detekt on all four modules: green
check-event-contract-coverage.py            OK: 46 producer topics, 0 new undocumented
check-event-contract-code-agreement.py      OK: 13 contracts agree with their code
check-audit-money-path-subscription.py      clean (22 topics, 9 known gaps, unchanged)
check-domain-event-occuredat.sh             clean (3695 .kt)
check-event-schema-compat.py --base origin/main   no backward-incompatible change
```

Two `@QuarkusTest` classes failed on the first full run with `Unable to start HTTP server` and took
their siblings to SKIPPED with them — the documented parallel-build port collision, not this change.
Re-run serially: `PartyApiIT`, `PartyOutboxWriteIT`, `PartyOutboxClaimIT`,
`PartyOutboxDispatcherCdiIT`, `PartySequenceAllocationIT`, `PartyPactFolderProviderVerificationTest`
and `DisputeMissingParamStatusIT` are all green with zero skips. The skipped counts above are read
from the serial runs, not the parallel one.

## Review notes

- `openbank-interest-service` is money-path (`rules.yaml: money_path_services`) ⇒ two approvals. Its
  threat model is unchanged: this adds a timestamp the service already computes to a payload it
  already publishes, and introduces no new trust boundary, actor or data class.
- `openbank-audit-service/CLAUDE.md` is also touched by #5967, which appends at the end of the file
  while this inserts mid-file. Expected to merge cleanly; worth a look if both land close together.
- The interest test is a new class because `InterestServiceTest` is at detekt's `LargeClass`
  threshold — measured, the inline version failed the gate.

Closes #8352
Refs #3883, #3907, #3914, #3926, #4087

### DIFF
--- a/openbank-audit-service/CLAUDE.md
+++ b/openbank-audit-service/CLAUDE.md
@@ -43,6 +43,32 @@
   replaces. `TopicAttribution` is a verified table whose COVERAGE (not values) is derived from
   `application.yaml`, so a newly subscribed topic fails `TopicAttributionCoverageTest` instead of
   silently defaulting again.
+- **Event time vs ingest time: the per-topic disposition, and why it is HERE and not in the
+  consumer's KDoc** (#8352, sweeping #3883). `AuditConsumer.eventTime` reads `occurredAt` and only
+  `occurredAt`; absent or unparseable ⇒ the row stores the consumer's clock, flagged
+  `occurred_at_source = INGEST`. As of 2026-09-03 the channel subscribes to **27** topics (not the
+  21 of #3883) and every one of them carries a top-level `occurredAt` on every production-reachable
+  event type. Four event types did not until this sweep, and all four had the instant already in
+  hand and simply never projected it onto the wire: `dispute.opened` (had it as `openedAt`),
+  `PARTY_ERASED` (as `erasedAt`, while `PartyEvent.occurredAt` held the same value and fed the
+  outbox row's `createdAt`), and interest-service's `interest.withholding.recorded.v1` /
+  `interest.withholding.remitted.v1` (nothing on the wire; the values are the withholding row's and
+  the remittance batch's own `createdAt`). **No topic on this channel genuinely lacks a business
+  time** — the honest-INGEST case the issue allowed for did not turn out to exist here.
+  Two things about that sweep are the reusable part:
+  - **Enumerate the producing TYPE, never grep the key.** Payloads are built four ways on this
+    channel — a serialised `DomainEvent` subclass (cards, delegation, ledger, consent, account,
+    transaction, balance, domestic-payment, fx, document), a `mapOf`/`ObjectNode` envelope (party,
+    kyc, clearing, sca, swift, sepa-instant, customer-edge, ICT), a raw JSON string template
+    (dispute, statement, lending, sdd, interest, fraud, sepa-payment's Temporal activities), and
+    sanctions' hybrid (serialise the aggregate, then `put` the per-event instant). A grep for
+    `"occurredAt"` sees only the last two families and reports a confident false gap over the first.
+  - **A `has(key)` assertion cannot see the wire form.** `PartyEventEnvelopeContractTest` built its
+    mapper with `findAndRegisterModules()` alone, which leaves `WRITE_DATES_AS_TIMESTAMPS` ON, so
+    every `Instant` in it serialised as the number `1.7866E9` — a shape no consumer of that topic
+    receives, since Quarkus defaults the property to false. Nothing failed, because every time
+    assertion in the file was `json.has(...)`, true of a number as readily as of an ISO-8601 string.
+    Assert the VALUE. The same trap cost #3926 a red run on document-service.
 - **Asking the audit DATABASE which fields producers send measures traffic, not the wire contract —
   and it answers "nothing to recover" with total confidence.** Chasing the actor half of #3994, the
   obvious probe was to enumerate the payload keys of every actor-less row
--- a/openbank-audit-service/src/main/kotlin/com/openbank/audit/application/AuditConsumer.kt
+++ b/openbank-audit-service/src/main/kotlin/com/openbank/audit/application/AuditConsumer.kt
@@ -306,12 +306,23 @@ class AuditConsumer {
      * second silent path, and a silent path is exactly how the gap below survived unnoticed.
      *
      * Two ways to have no event time, both previously invisible:
-     *  - the key is absent. 7 of the 21 consumed topics are in this state today (clearing,
+     *  - the key is absent. The old code substituted `Instant.now(clock)` and the row then
+     *    asserted, indistinguishably from a real one, that the operation happened when the consumer
+     *    got round to it. Under consumer lag or a replay that is arbitrarily wrong, and it is the
+     *    GDPR Art. 30 "when" dimension and DORA Art. 17 evidence.
+     *
+     *    **This paragraph used to name the topics in that state, and the list went stale within
+     *    hours of being written** (#8352). It said "7 of the 21 consumed topics — clearing,
      *    dispute, statement, sanctions, six of lending's payloads, sepa-payment's Temporal path,
-     *    and document-service which names it `at`). The old code substituted `Instant.now(clock)`
-     *    and the row then asserted, indistinguishably from a real one, that the operation happened
-     *    when the consumer got round to it. Under consumer lag or a replay that is arbitrarily
-     *    wrong, and it is the GDPR Art. 30 "when" dimension and DORA Art. 17 evidence.
+     *    and document-service which names it `at`", and by the time anyone read it the #3914/#3926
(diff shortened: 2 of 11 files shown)


JiRaska:
Force-pushed once, `432167cbf0 -> 2259bbdc67`, to correct the **commit split only** — `git diff 432167cbf0 HEAD` is empty, so not one byte of the tree changed.

What was wrong and why it mattered here: I had re-applied the production fixes with `git apply --3way`, which **stages** what it applies, and my subsequent `git add <dispute files>` did not clear the index. So the `fix(dispute)` commit swallowed `InterestService.kt`, `WithholdingRemittanceService.kt` and `PartyEvent.kt`, leaving `fix(party)` and `fix(interest)` holding test files only.

That is not cosmetic under this repo's release rules. The scope selects the released component and `src/test` is in `exclude-paths`, so party-service and interest-service would have shipped a production fix with **no release and no changelog entry**, while dispute-service's changelog claimed three services' worth of change — `rules.yaml: release_scope_mismatch`, arrived at from the wrong direction.

Each commit now touches exactly its own service. Caught by reading the per-commit file lists after committing, not by any gate; nothing in CI can see it, because the tree is correct either way.

PR #8504, opened 2026-09-03
fix(document): createdAt defaults to now(), never Instant.EPOCH (#8357)

JiRaska:
## Summary

Burns down 3 of the 58 baselined `Instant.EPOCH` defaults (#8357 follow-up), document-service slice.

`DocumentEntity`, `DocumentTemplateEntity`, `SignatureCeremonyEntity` all declared `var createdAt: Instant = Instant.EPOCH`. Every construction site is `XEntity().also { it.createdAt = createdAt }` with the domain value always supplied (`Instant.now(clock)` upstream) — so the default was only a pre-population placeholder, and a code path that forgot to populate would persist a silent 1970 that passes every `isNotNull()` check (#3882, #4208).

Now `Instant.now()`: a missed population becomes a plausible-but-wrong timestamp that recency assertions can catch, never a silent epoch.

The 3 baseline lines are removed in the same PR, as the ratchet (`check-no-epoch-instant-default.py`) requires.

Refs #8357

## Test plan

- [x] `./gradlew :openbank-document-service:build` green (test + ktlint + detekt + quarkusBuild)
- [x] `check-no-epoch-instant-default.py` green: 37 occurrences, all baselined, no new
- [ ] CI gates


JiRaska:
**This duplicates #8405's document-service third.** #8405 was opened at 08:03Z, this at 12:59Z, and they make the same three edits to the same three files plus the same two baseline lines.

Compared by content, not by file list:

| | #8405 | #8504 |
|---|---|---|
| `DocumentEntity.kt`, `DocumentTemplateEntity.kt`, `SignatureCeremonyEntity.kt` | same three `EPOCH` → `Instant.now()` edits | same |
| `.github/gates/epoch-instant-default-baseline.txt` | same three lines removed | same |
| also covers | `openbank-sanctions-service` (5 more entries) | — |
| adds | — | three `// Never Instant.EPOCH (#3882): …` comments |

Whichever lands second conflicts on both the source files **and** the baseline — and the baseline conflict is the nastier one, because that list fails on a stale entry in either direction, so a partial resolution reddens `main` rather than just failing to apply.

Suggest closing this one and carrying its three comments into #8405 if they are wanted; #8405 is older, strictly broader, and already `MERGEABLE`.

Wider context, since it is easy to miss from inside one PR: **a seven-PR stack already claims all 40 baseline entries, exactly once** — #8402 (libs-domain 4), #8403 (account 5), #8405 (document 3 + sanctions 5), #8454 (product-catalog 3), #8462 (standing-order + swift + tax 5), #8463 (transaction 2), #8465 (ledger 13). `comm -23` of the baseline against the claimed set is empty. So no further burn-down PR is needed; the remaining work is landing that stack.

**#8403 appears to be the only blocker.** GitHub reports it `CONFLICTING/DIRTY`, but a local `git merge-tree --write-tree` of its head into its base exits **0** — mergeability is lazily computed, so that flag looks stale and a push should clear it.

Also worth correcting while these are open: the `58` in this issue's body and in the `gates.yaml` comment above `no-epoch-instant-default` is stale — the baseline holds **40**.

Issue #8353, opened 2026-09-03
Outbox-atomicity integration test per money-path service (LendingOutboxWriteIT pattern)


JiRaska:
## Context
M2 lists integration-test depth as a remaining gap. The house pattern for proving transactional-outbox atomicity (status change + outbox row commit together) is a real-DB IT driven through the REST endpoint — `LendingOutboxWriteIT`, `ConsentRevocationOutboxIT`. A mocked repository cannot prove it, and a bare `@QuarkusTest` thread cannot call a reactive Panache repo (`No current Vertx context found`).

## Acceptance criteria
- [ ] Each of the 14 money_path_services has at least one real-DB outbox-atomicity IT following the `LendingOutboxWriteIT` pattern (RestAssured + `@TestSecurity` drive, plain JDBC assertion)
- [ ] Checklist of the 14 services in this issue, ticked as ITs merge
- [ ] Services with no outbox entity explicitly justified instead

Audit: docs/audit-2026-09-03-top50.md (B18). M2 exit criterion.

JiRaska:
First three services landed in **#8496** (transaction, sepa-payment, domestic-payment). Two things from doing it are worth putting in this issue rather than leaving in the PR.

## The house oracle is not sufficient, and the PR changes it

The existing pattern asserts the outbox row **landed**. That is necessary — a mocked repo commits nothing, and a reactive Panache repo cannot be called from a bare `@QuarkusTest` thread (`No current Vertx context found`), so the flow has to be driven through the REST endpoint — but a **two-transaction implementation satisfies every presence assertion while having lost the property**.

Postgres stamps each row version with `xmin`, the id of the writing transaction. Comparing the aggregate row's `xmin` with its outbox row's reads the property itself. Falsified by splitting `Panache.withTransaction` in two on each path:

| service | write path | broken | real |
|---|---|---|---|
| domestic-payment | `save` | RED — 761 vs 762 | GREEN |
| sepa-payment | `save` | RED — 747 vs 748 | GREEN |
| sepa-payment | `updateWithMessages` | RED — 751 vs 752 | GREEN |
| transaction | `update` (terminal) | RED — 768 vs 769 | GREEN |

One caveat for anyone extending this: the test must switch `openbank.outbox.dispatch-enabled` **off** for its own run, because the dispatcher's UPDATE rewrites the outbox row's `xmin` and destroys the evidence.

## The checklist's denominator is wrong

[...]

Issue #7539, opened 2026-08-28
Investigate standing-order stale outbox reclaim integration failure


JiRaska:
Immutable evidence from Services CI run 33145481933 records StandingOrderOutboxClaimIT: a stale DISPATCHING row is reclaimed instead of stranded forever as failed (1/7 integration tests). The same workflow was incorrectly marked green due to #7538. Investigate whether the test is flaky/isolation-sensitive or whether claimProcessable can fail to reclaim a stale row. Reproduce against PostgreSQL/Testcontainers before changing money-path behavior.

JiRaska:
Follow-up evidence (2026-08-28): the exact Testcontainers class `StandingOrderOutboxClaimIT` was rerun locally from current main and passed 2/2, including the stale-DISPATCHING reclaim case. This does not invalidate the immutable CI envelope failure from run 33145481933; it shows the failure is not currently deterministic. Keep this issue open for a clean CI rerun / isolation analysis, and do not change money-path dispatch semantics solely from this single signal.

JiRaska:
**STILL TRUE, but its framing has changed: the #7538 premise it rests on is false.**

This issue says *"The same workflow was incorrectly marked green due to #7538."* #7538 has now been closed as a non-defect — an EXIT trap does not replace a command's exit status, measured with a matched control:

```
$ bash -ec 'c(){ kill 99999 2>/dev/null || true; wait 99999 2>/dev/null || true; }; trap c EXIT; false'; echo "status=$?"
status=1
$ bash -ec 'c(){ kill 99999 2>/dev/null || true; wait 99999 2>/dev/null || true; }; trap c EXIT; true'; echo "control=$?"
control=0
```

and run 33145481933 had no failing job or step at all:

```
$ gh run view 33145481933 --json conclusion,jobs -q '.conclusion+" | non-success jobs: "+([.jobs[]|select(.conclusion!="success" and .conclusion!="skipped")|.name]|join(","))'
success | non-success jobs: 
```

So `Build + test` did **not** fail and get masked. The likelier source of the `StandingOrderOutboxClaimIT` failure line in that log is the advisory `Generate Kover XML report` step (`continue-on-error: true`), whose `koverXmlReport` pulls in `test` in a second daemonless JVM — a different process, different conditions, and by design not a verdict.

That makes flakiness / isolation-sensitivity the *leading* hypothesis rather than one of two equals: the same test passed in the mandatory build and failed in a second concurrent-ish invocation within the same job.

[...]

PR #8507, opened 2026-09-03
test(sanctions): prove outbox atomicity by transaction id in sanctions, standing-order and fx

JiRaska:
## What

The second tranche of #8353 — an outbox-atomicity integration test for three more money-path
services: **sanctions**, **standing-order**, **fx**. Follows the oracle #8496 established, verbatim:
compare the Postgres `xmin` of the aggregate row and its outbox row instead of asserting the outbox
row is merely present.

## How these three were chosen

Same criteria as #8496, applied to the 13-service remainder it named:

- **money-path weight.** `sanctions` gates every screened payment in the fleet (a wrongly cleared
  hit is a real sanctions violation); `standing-order` initiates recurring payments; `fx` settles a
  customer's currency conversion and then announces it.
- **the write path is genuinely co-transactional with a state change.** Each of the four paths below
  persists the aggregate and its outbox row inside one `Panache.withTransaction` /
  `sf.withTransaction`. Enumerated by reading every call site of the outbox port's
  `persistInTransaction` across all 13 candidates, not by naming.
- **a proven-green REST-driven `@QuarkusTest` already exists in the module**, so a boot failure
  cannot masquerade as a pass: `SanctionsPactProviderVerificationTest` (real HTTP + `@TestSecurity`),
  `StandingOrderBootSmokeIT`, `FxNulByteRejectionIT`.

Three candidates were rejected on the third criterion or worse and are named below so the remainder
stays accurate rather than implied.

Each write path was run against a **deliberately broken** build — its single transaction split in
two — and goes red naming the two transaction ids:

| service | write path | REST entry | broken-build failure |
|---|---|---|---|
| sanctions | `saveWithEvent` (automated screening) | `POST /api/v1/sanctions/screen` | check xmin=755, outbox xmin=756 |
| sanctions | `updateWithEvent` (analyst disposition) | `POST /api/v1/sanctions/review` | check xmin=759, outbox xmin=760 |
| standing-order | `saveWithExecution` (3rd failure ⇒ FAILED) | `PATCH /api/v1/standing-orders/{id}/record-failure` | order xmin=748, outbox xmin=749 |
| fx | `saveWithOutbox` (settled conversion) | `POST /api/v1/fx/convert` | conversion xmin=747, outbox xmin=748 |

The source was restored after each run (`git checkout --`) and the tests re-run green. This PR
contains **test files only**, plus three `build.gradle.kts` lines explained below.

Module runs, reading the **SKIPPED** count and not just failures — a service that cannot boot
reports its tests as SKIPPED, and a skip count scans as a pass:

|---|---|---|---|
| `openbank-sanctions-service` | 97 | 1 | 0 |
| `openbank-standing-order-service` | 54 | 0 | 0 |
| `openbank-fx-service` | 162 | 1 | 0 |

Both skips are the broker-sourced Pact provider classes (`@PactBroker` +
`@EnabledIfSystemProperty(pactbroker.url)`), pre-existing and expected on a PR lane.
`ktlintCheck` + `detekt` were run as a **separate** Gradle invocation after `test` — Gradle stops at
the first failing task, so a green-looking task list is not a list that ran — clean on all three.

## Controls, so the assertions cannot pass vacuously

- **Known-negative** in all three classes: an id that was never written must return no pair at all,
  so `hasSize(1)` is a claim the query is capable of failing.
- **Known-different**, so one run shows the same comparison both matching and not matching:
  - *sanctions* gets it for free — the review MERGEs the check, so the **screening** outbox row was
    genuinely written by an earlier transaction and must NOT match the reviewed row version.
  - *standing-order* captures the row version left by the **second** `record-failure` (a transaction
    that wrote no outbox row) and requires the final outbox row not to match it.
  - *fx* drives two conversions and requires the identical comparison to fail across them.
- **Arrangement assertions**, so a change that moves the test onto a different write path fails here
  rather than silently asserting over an empty outbox: standing-order asserts `ACTIVE`/`ACTIVE`/
  `FAILED` per call, fx asserts the conversion came back `SETTLED` (the HOLD path writes no outbox
  row at all).

## The dispatcher, and one collaborator that had to be pinned

`xmin` belongs to a row *version*, so any later UPDATE replaces it. The scheduled outbox dispatcher
claims rows and marks them DISPATCHING/SENT, so it is switched **off** per class. All three services
correctly ship `openbank.outbox.dispatch-enabled: true` in `application.yaml` — the fleet default is
`false`, under which a service dispatches nothing and reports no error — which is precisely why it
has to be disabled here. standing-order already sets it `false` under `%test` (#7539); the class
states it anyway, so a later profile edit cannot silently make the assertion read a
dispatcher-rewritten row.

`fx.convert` additionally consults sanctions screening, and that call decides **which** write path
runs: an unreachable sanctions service raises `ScreeningUnavailableException` and the use case fails
closed into HOLD, which writes no outbox row — the test would then assert nothing while looking like
it had. It is mocked to CLEAR. Fraud scoring, by contrast, fails **open** by design (#4221) and runs
after the write, so it needs no stub.

## Two findings, reported and NOT fixed here

A test PR that also changes production code is the wrong shape for review, so both are left for
their own PRs.

**1. `openbank-clearing-service` settle is NOT atomic — measured, not inferred.**
`ClearingService.settleBatch` chains three separate transactions:
`batchRepo.update(settled)` (`sf.withTransaction`) → `itemRepo.saveAll(...)` (`sf.withTransaction`)
→ `eventPublisher.publishBatchSettled(...)`, whose `ClearingEventPublisherImpl` opens its **own**
`Panache.withTransaction`. A throwaway copy of this PR's test, run against unmodified `main`,
answered `batch xmin=750, outbox xmin=752` — two transactions, with the item write taking 751 in
between. So a crash after the batch commits SETTLED loses `openbank.clearing.batch.settled`
permanently, and every presence-based assertion stays green about it. `publishItemCleared` has the
same shape. That test is deliberately **not** in this PR: it would be red on `main`.

**2. `openbank-balance-service` and `openbank-psd2-service` have an outbox nothing ever writes to.**
Both carry the full apparatus — `*OutboxEntity`, `*OutboxRepositoryImpl` with `persistInTransaction`,
a dispatcher and a backlog gauge — and the only consumers of the repository in `src/main` are the
dispatcher and the gauge. No domain code calls the write. `balance_outbox` / `psd2_outbox` are
therefore permanently empty: the backlog gauge reads 0 and the dispatcher succeeds at dispatching
nothing, which is indistinguishable from a healthy outbox. Atomicity is not the defect there;
the missing producer is.

## The fx heap line

`openbank-fx-service/build.gradle.kts` gains `maxHeapSize = "2g"`, the third module to need it after
`openbank-account-service` and `openbank-lending-service`. A class that switches
`openbank.outbox.dispatch-enabled` off is a distinct Quarkus test configuration and so forces its own
boot in the same forked JVM — a fifth for this module. Measured on this branch: without the bump the
suite dies with `java.lang.OutOfMemoryError: Java heap space` inside the Gradle test executor after
116 of 160 tests, and the JUnit XML then reports **zero failures** — the run simply stops, which
reads as a pass to anything counting failures. With it, 162/162. Baseline without the new test:
160/160, no OOM, so the cost is attributable and not pre-existing.

The other two build-file lines are single test-scope dependencies: `rest-assured-kotlin` for
sanctions (its Pact verification uses `HttpTestTarget`, so the module had no RestAssured) and
`quarkus-test-security` for standing-order (its existing ITs only hit unsecured routes).

## Candidates rejected, and why

- **`swift`** — `saveWithOutbox` is genuinely co-transactional and money-path, but `POST /api/v1/swift`
  reaches it only past `schemeGatewayPort.submit`; with no gateway the message holds in VALIDATED and
  writes no outbox row. Needs a stub; the module's only ITs are a boot smoke and an outbox-claim.
- **`interest`** — `saveWithOutbox` puts the capitalization, the withholding row, the outbox event and
  the accrual flip in one transaction, but `POST /capitalize/{accountId}` runs `postCreditLeg` against
  the ledger first, so it needs a stubbed ledger.
- **`billing`** — both co-transactional paths (`/post`, `/reverse`) sit behind the four-eyes
  `@Authorize` gate.

## Remaining scope for #8353

Money-path, has an outbox entity, still only a `*OutboxClaimIT`: `account`, `balance`, `clearing`,
`swift`, `sdd`, `interest`, `sca`, `fraud`, `billing`, `psd2` — 10 services, of which `balance` and
`psd2` are the dead-outbox finding above and `clearing` is blocked on the atomicity defect above.

Refs #8353

## Security checklist

      (license, CVE, source).

Each box checked against the diff rather than assumed:

- Test-only on the code side: three new `*OutboxAtomicityIT.kt` under `src/test`, and **zero** files under `src/main`.
- **The three `build.gradle.kts` edits are not a dependency change** — each adds a `tasks.withType<Test> { }` heap override, the same one `openbank-account-service` and `openbank-lending-service` already carry. No new coordinate is declared, so the third-party box holds. (Worth reading the comment it ships with: without the bump fx-service dies with `OutOfMemoryError` after 116 of 160 tests and the JUnit XML then reports **zero failures** — the run stops rather than fails, which reads as a pass to anything counting failures.)
- No `@Suppress`/`as Any` added; no secret-shaped additions. The ITs read `xmin` over the Quarkus datasource and introduce no credential of their own.
- No auth, crypto or payment code changed — adding a test does not change the path it observes — so `security-review-required` is deliberately not applied. No PII and no cardholder-data path touched.


JiRaska:

PR #8520, opened 2026-09-03
fix(card-issuance): bump openapi.yaml info.version to 2.0.0 — the cardholderName change was breaking (#8150)

### LINKED ISSUE #8150: A spec merged to main with a mis-classified info.version
The post-merge api-contract classification FAILED for `130998006d157543e6769dda19ab73e6f6133728`: an
`openapi.yaml` landed on main whose `info.version` does not match the change it
actually makes.

**This is the residual race ci.yml's diff-base comment names** (#481 × #524): two
spec PRs each classified green against their own base, the second merged without a
re-run, and it landed new endpoints under a version the first already took. The
PR-time gate cannot see it — that needs a merge queue (unavailable: org-only, #1465)
or required-up-to-date branches (rejected: ~8 CI re-runs and ~70 min per service PR
at this repo's ~93 s median merge gap, to close a race that needs two PRs on one
spec while only 1 in 60 PRs touches a spec at all).

What the classifier found (verbatim — the run log expires, this does not):

```
api-contract gate: openbank-card-issuance-service/src/main/resources/openapi.yaml: correction API change requires an info.version MINOR bump, but it moved 1.8.0 -> 1.8.0. First breaking change: added the new required request property `cardholderName` (API axis is independent of version.txt — ADR-0048 D5)
```

Run: https://github.com/JiRaska/open-bank-oss/actions/runs/33612353177. Last-good base: `464997bbb91347ff1ae080dcf122edb0ec84e313`.

**Fix forward — but re-classify the CURRENT tip of main FIRST.** A later spec merge
may already have moved this version, and then the tip is correct and a further bump
would OVERSTATE the contract: a version consumers cannot distinguish from a real
change. That is the same defect in the opposite direction, and ADR-0048 forbids it
just as firmly. This is not hypothetical — it is what happened to the commit that
first raised this issue (#4620): `25bd631` (#4194) left `openbank-customer-edge` at
1.34.0 while adding `POST /cards/delegated`, and 74 minutes later an unrelated
additive change (#4641) took 1.35.0, making the tip honest with no action at all.
Bumping "to the next free version" as this issue used to instruct would have burned
1.36.0 on an empty diff and created the overstatement.

```
git fetch origin main && git checkout origin/main
python3 .github/scripts/check-api-contract.py --base 464997bbb91347ff1ae080dcf122edb0ec84e313 --enforce
```

Bump only if that STILL reports the spec, and then to the next number free on LIVE
main — the API axis is independent of `version.txt` (ADR-0048), and whoever lands
second takes the next number. Never rewrite main.

**Close this issue when the command above reports the flagged spec clean** — not
when "a spec merge is green again". Those are different questions: the next green
spec merge is very often a DIFFERENT service, and its greenness says nothing about
whether this spec was ever remediated. This issue is refreshed on each further
occurrence, so re-check every spec named above before closing.

JiRaska:
## Summary

Fix-forward for #8150, per the issue's own procedure: re-classified the **current tip of main** against the last-good base (`464997bb…`) — card-issuance is still mis-classified (`breaking change requires a MAJOR bump, but it moved 1.8.0 -> 1.8.0`), so this bumps `info.version` to **2.0.0**.

The breaking change is real and intentional: `cardholderName` became a required request property on the card issue/enroll payloads, and a card without a cardholder name is not a thing — optionalising it would be a product lie, not a fix. The honest contract statement is a MAJOR.

Verified after the bump: `check-api-contract.py --base 464997bb…` reports `breaking change, info.version 1.8.0 -> 2.0.0 — OK (required >= MAJOR)`.

**Known remainder (disclosed, not introduced):** the D2/D4 invariant now warns that major(info.version)=2 while the newest served URL is `/api/v1`. Cutting `/api/v2` routing is separate ADR-0048 work (the ADR itself lists hardcoded v1 URLs as risk #2) and the invariant is advisory until its enforce date. Not cutting v2 here keeps this PR a spec-only, no-behaviour-change fix.

Closes #8150

## Test plan

- [x] `check-api-contract.py --base 464997bbb91347ff1ae080dcf122edb0ec84e313` — flagged spec now clean
- [x] No code or test pins `1.8.0` (grepped the module)
- [ ] CI gates


JiRaska:
Uzavírám bez merge — bump na 2.0.0 sice udělí klasifikátor, ale PR-time api-contract gate ENFORCED hlásí D2/D4 (major 2 vs /api/v1). Správný fix je buď /api/v2 cut (migrace admin-ui, e2e, callerů) nebo zdokumentovaná výjimka — obojí ADR-level rozhodnutí. Zdokumentováno v #8150.

PR #8544, opened 2026-09-03
docs(kyc): mark DOCUMENTS_REQUIRED unreachable instead of removing it

### LINKED ISSUE #8535: KYC DOCUMENTS_REQUIRED is a dead state — documented in 2 ADRs, drawn, published in 2 OpenAPI enums, rendered by the cockpit, and unreachable in code
## `DOCUMENTS_REQUIRED` is a state no case can ever enter

`KycCaseStatus.DOCUMENTS_REQUIRED` is declared, documented in two ADRs, drawn in the lifecycle diagram, published in two OpenAPI enums and rendered by the admin cockpit — and **nothing in the fleet transitions a case into it**.

Every occurrence outside the enum declaration is documentation or display:

```
docs/adr/0068-onboarding-operations-cockpit.md      ADR text + funnel stage table
docs/adr/0116-kyc-engine-risk-checks-...md          "OPEN → DOCUMENTS_REQUIRED → UNDER_REVIEW"
openbank-kyc-service/.../diagrams/03-kyc-status-lifecycle.mmd
                                                    "OPEN --> DOCUMENTS_REQUIRED: operator marks docs missing"
openbank-kyc-service/src/main/resources/openapi.yaml   enum values
openbank-customer-edge/src/main/resources/openapi.yaml enum value
openbank-admin-ui/src/app/onboarding/page.tsx       label + colour for KYC_DOCUMENTS_REQUIRED
openbank-kyc-service/.../docs/0*.{cs,en}.md         overview + data dictionary
```

`KycService`'s complete operation set is `openCase`, `openCaseForParty`, `getCase`, `listCases`, `countCases`, `getCaseByParty`, `updateCheckStatus`, `applyPepScreeningResult`, `approve`, `reject`, `approveCase`, `rejectCase`. **None sets `DOCUMENTS_REQUIRED`.** The diagram says *"operator marks docs missing"* — no operator can, because no endpoint does it.

## Consequences

1. **The admin funnel stage always reads zero.** ADR-0068's cockpit has a `KYC_DOCUMENTS_REQUIRED` stage with a label, a colour and a place in the funnel. It can never be non-empty, and it looks like "no one is stuck on documents" rather than "this state is not implemented".
2. **The edge publishes a status the app can never receive.** `GET /kyc` declares `DOCUMENTS_REQUIRED` in its enum, so every client models a branch that is dead.
3. **`KYC_DOCUMENT_REQUIRED` (notification template) is unproducible by construction** — the last of the #8432 sweep that cannot be closed by wiring alone. It requires a `documentType` variable, and **kyc-service has no concept of a document type**: `grep -i upload` over `openbank-kyc-service/src/main/kotlin` and `openbank-onboarding-service/src/main/kotlin` returns **0 hits**.

## Why this is not a notification ticket

I deliberately did not emit `KYC_DOCUMENT_REQUIRED` in #8439, and the port's KDoc says why: there is no transition to fire it from and nowhere for the customer to put a document. **A push asking someone for a document they cannot supply is worse than the silence it replaces.** That reasoning still holds — closing this needs the capability, not the message.

## What would close it

A vertical slice, in this order:

1. **kyc-service**: an operator operation `POST /cases/{id}/request-documents` taking a document type (or several), transitioning `OPEN → DOCUMENTS_REQUIRED` and recording what was asked for. This is where the missing `documentType` concept has to be born.
2. **Storage + a customer upload path**: an endpoint the customer can actually submit to, and somewhere to put the file. document-service exists and already renders/stores documents — whether it is the right home for inbound KYC evidence is a design question, not a given.
3. **The transition back**: `DOCUMENTS_REQUIRED → OPEN` on submission, as the diagram already claims happens.
4. **Then, and only then**, the notification: `KYC_DOCUMENT_REQUIRED` with the real `documentType`, plus a deep link to the upload screen (which also means adding it to `MobileDeepLink`'s closed allow-list).

## What needs deciding, not just typing

- **Who may request documents, and is it four-eyes?** ADR-0116 puts approve/reject behind a four-eyes gate. Asking a customer for identity evidence is a weaker action but not a trivial one, and the gate's scope is a compliance choice.
- **Where inbound KYC evidence is stored, and under what retention.** This is identity documentation, not a bank statement; ADR-0118's retention rules were not written with it in mind.
- **Whether the state should exist at all.** The alternative, and it is a real one: if the bank never intends to ask customers for documents through the app, then `DOCUMENTS_REQUIRED` should be **deleted** from the enum, the OpenAPI contracts, the diagram, the ADRs and the cockpit — and `KYC_DOCUMENT_REQUIRED` deleted from `NotificationTemplate`. A dead state that four documents describe as live is its own defect. **This is the cheaper fix and it should be considered first.**

Found while closing the #8432 sweep. Of the twelve declared-but-never-produced notification templates, eight are now closed (#8439, #8469, #8491, #8508); this one and `OTP_CODE`/`PASSWORD_RESET`/`WELCOME` are the remainder, and none of the four is a wiring job.

JiRaska:
Closes #8535 and #8618. **The decision on #8618, taken rather than deferred: do not narrow the contract.**

## Why not my own earlier preference

I had preferred teaching `api-contract-gate`'s `correction` class to recognise a narrowing that brings a spec into line with its implementation. **That does not survive contact with the code.** The discriminator would have to be *"the server cannot emit this value"* — and it can: `DOCUMENTS_REQUIRED` **is** in `KycCaseStatus`, so statically the value is producible. That nothing constructs it is a reachability property I established by hand, not something a gate can see. The gate's own KDoc also rules out the fallback: *"the discriminator is the diff itself, never a declared marker."*

**Baselining the drift does not help either**, which I had assumed it might. The baseline lives in `openapi-enum-domain-drift`; the blocker for the combined change is `api-contract-gate`. Buying the removal that way costs a split, a deliberately parked drift, and three follow-up PRs.

So **both gates are right and the removal is genuinely breaking by their standard.** It is not worth a MAJOR — which ADR-0048 D2 requires, moving the service to `/api/v2` — for a value no client can ever receive.

## What this means

The harm that mattered is already fixed. #8834 removed the `KYC_DOCUMENT_REQUIRED` notification template, the cockpit column that could only ever read zero, and the alert regex naming it. What remained was tidiness, at the price of a breaking contract change.

This PR makes the remaining documentation honest **without touching a contract**:

- the **lifecycle diagram** stops drawing transitions into `DOCUMENTS_REQUIRED`, and a note says the state is declared, unreachable, and deliberately kept;
- the **kyc and onboarding service docs** mark the value unreachable where they list it;
- **ADR-0068 and ADR-0116** notes record the decision.

`ESCALATED` **does** come out of the diagram: it was never in `KycCaseStatus` and appears in no spec enum, so removing it narrows nothing. The note explains where escalation actually lives — a check in `MANUAL_REVIEW` plus `due_diligence_level = EDD`, with MLRO escalation belonging to the *AML* case in aml-service (ADR-0032).

## Gate status

Documentation only. No enum, no OpenAPI, no Kotlin — so neither `api-contract-gate` nor `openapi-enum-domain-drift` has anything to object to. The red that made this PR interesting is gone because the change that caused it is gone.

### DIFF
--- a/docs/adr/0068-onboarding-operations-cockpit.md
+++ b/docs/adr/0068-onboarding-operations-cockpit.md
@@ -35,8 +35,10 @@ choreographed across three services, each owning one slice of the truth:
 > `KYC_DOCUMENTS_REQUIRED` funnel column in §4.2, were never implemented: no kyc-service operation
 > ever set that status, so the cockpit column could only ever read zero — which reads as "nobody is
 > stuck on documents" rather than "this is not built". The column has been removed from the cockpit.
-> The status itself is still declared in the enums and contracts: removing it there is blocked by
-> two mutually exclusive CI gates (#8618). The decision recorded below is preserved as written;
+> The status itself stays declared in the enums and contracts, deliberately: dropping a value from
+> a published enum is a breaking contract narrowing, and it is not worth a MAJOR for a value no
+> client can ever receive (#8618). The service docs and the lifecycle diagram now say it is
+> unreachable rather than describing transitions that do not exist. The decision recorded below is preserved as written;
 > collecting documents from a customer needs an operator endpoint, an upload path and storage, and
 > would be a new ADR.
 
--- a/docs/adr/0116-kyc-engine-risk-checks-and-four-eyes-gate.md
+++ b/docs/adr/0116-kyc-engine-risk-checks-and-four-eyes-gate.md
@@ -67,9 +67,10 @@ OPEN / DOCUMENTS_REQUIRED / UNDER_REVIEW → EXPIRED   (terminal, time-driven)
 > **Correction, 2026-09-05.** Three gaps between this lifecycle and the code, found together:
 >
 > - **`DOCUMENTS_REQUIRED` was never implemented** (#8535). No operation in `KycService` sets it,
->   and the service has no concept of a document type or any upload path. Removing it from the
->   enums and contracts is blocked by two mutually exclusive CI gates (#8618); the dead cockpit
->   column it fed has been removed, and the `KYC_DOCUMENT_REQUIRED` notification template with it.
+>   and the service has no concept of a document type or any upload path. The dead cockpit column
+>   it fed and the `KYC_DOCUMENT_REQUIRED` notification template are gone. The status itself stays
+>   declared: removing it from a published enum is a breaking narrowing, not worth a MAJOR for a
+>   value no client can receive (#8618). The docs and the diagram above now mark it unreachable.
 > - **`ESCALATED` was never a KYC case status at all.** The lifecycle diagram drew it with four
 >   transitions; it is not in `KycCaseStatus`. Escalation here is a check in `MANUAL_REVIEW` plus
 >   `due_diligence_level = EDD`, with the case staying `UNDER_REVIEW`; MLRO escalation is a state
--- a/openbank-kyc-service/src/main/resources/diagrams/03-kyc-status-lifecycle.mmd
+++ b/openbank-kyc-service/src/main/resources/diagrams/03-kyc-status-lifecycle.mmd
@@ -2,20 +2,11 @@
 stateDiagram-v2
   [*] --> OPEN: case created (PARTY_CREATED event or POST /kyc/cases)
 
-  OPEN --> DOCUMENTS_REQUIRED: operator marks docs missing
   OPEN --> UNDER_REVIEW: all mandatory checks recorded
   OPEN --> EXPIRED: expires_at passed without decision
 
-  DOCUMENTS_REQUIRED --> OPEN: documents submitted
-  DOCUMENTS_REQUIRED --> EXPIRED: expires_at passed
-
   UNDER_REVIEW --> APPROVED: POST /approve (four-eyes, authorised reviewer)
   UNDER_REVIEW --> REJECTED: POST /reject (four-eyes, authorised reviewer)
-  UNDER_REVIEW --> ESCALATED: escalation required (EDD / PEP)
-
-  ESCALATED --> UNDER_REVIEW: escalation resolved
-  ESCALATED --> APPROVED: escalation cleared, four-eyes approve
-  ESCALATED --> REJECTED: escalation rejected
 
   APPROVED --> [*]: terminal — emits KYC_CASE_APPROVED → party activation
   REJECTED --> [*]: terminal — emits KYC_CASE_REJECTED → party SUSPENDED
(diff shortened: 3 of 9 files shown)


JiRaska:
Held back, and taken out of the merge queue. Red on `openapi-enum-domain-drift` and nothing else.

Removing this value cannot pass CI in any PR shape: `api-contract-gate` wants the change to be spec-only (its `correction` class, since a MAJOR is forbidden by D2 while the URL major stays /api/v1), while `openapi-enum-domain-drift` wants the spec and the Kotlin enum to match. Spec-only, Kotlin-only and combined were all measured; each fails one of the two.

Written up with the measurements and four options in #8618. The other six PRs from this batch are independent and unaffected.

JiRaska:
These two PRs are **complementary halves of one removal**, and neither can go green alone. Each is red on `openapi-enum-domain-drift` *because the other half is missing*.

Read from each run’s `gate-results-gitops-api` artifact and the gate’s own output (the job log also echoes the PR body, which is what makes a plain grep misleading here):

```
#8544  spec advertises values the code does not have: [DOCUMENTS_REQUIRED]      (FunnelStage, KycCaseStatus, KycStage)
#8606  spec omits [DOCUMENTS_REQUIRED]                                          (FunnelStage, KycCaseStatus, KycStage)
```

Opposite directions, same three enums, same value — because the removal is split:

| | removes `DOCUMENTS_REQUIRED` / `KYC_DOCUMENTS_REQUIRED` from |
|---|---|
| **#8544** | the Kotlin enums (`KycCase.kt`, `OnboardingRecord.kt`), `openbank-customer-edge/openapi.yaml`, admin-ui, the lifecycle diagram, 8 doc files, 2 test files |
| **#8606** | `openbank-kyc-service/openapi.yaml` and `openbank-onboarding-service/openapi.yaml` — the two specs #8544 does not touch |

**The file sets do not overlap at all** — `comm -12` over the two diffs is empty. So there is no conflict to resolve; there is a dependency neither PR declares.

`all-green` is required, so this is a deadlock rather than a race: #8544 waits for the two specs #8606 holds, #8606 waits for the code #8544 holds, and no amount of re-running moves either.

**Cheapest resolution — stack them.** Merge `#8544`’s branch into `#8606`’s (the repo’s documented move for exactly this: *“merging the parent branch into the child makes the child’s CI green independently; the shared files drop out of the diff as the parent lands on main”*). #8606 is 2 files, so it is the natural child. Alternatively #8544 absorbs the two `openapi.yaml` edits and #8606 closes.

**Do not merge #8544 alone.** With the code gone and the kyc/onboarding specs still advertising the value, `main` ends in the state that reds this gate for the *next* PR to touch either service — the drift becomes someone else’s failure.

**One more thing specific to this PR:** it bumps `openbank-customer-edge` `info.version` to **1.52.0** from a base of 1.50.0, while #8321 claims 1.51.0. That ordering only holds if #8321 lands first; if it does not, 1.51.0 is skipped. `main` is at 1.50.0 right now — re-read it at the moment you rebase.

Issue #8345, opened 2026-09-03
Pact contract coverage for every money-path cross-service HTTP call


JiRaska:
## Context
Pact Broker is live (pact.open-bank.tech) but published pact coverage is a known roadmap gap (ROADMAP.md "Known gaps"). Concrete fallout already observed: #6568 — 3 services (2 money-path) blocked at can-i-deploy because openbank-document-service publishes zero pacts; #7376 — provider verification broken by unparsable selector discovery.

## Acceptance criteria
- [ ] Every cross-service HTTP client call originating from a money_path_service (rules.yaml) has a consumer pact published to the broker
- [ ] Every provider serving a money-path consumer runs pact verification in CI (can-i-deploy gate)
- [ ] Services with zero published pacts are listed in the issue and ticked off one by one
- [ ] \`pact-drift-check.yml\` fails when a money-path consumer/provider pair loses its pact

Refs: #6568, #7376. Audit: docs/audit-2026-09-03-top50.md (B9)

JiRaska:
Measured the denominator this issue was missing. On `origin/main`:

| | count |
|---|---|
| money-path outbound REST clients (`@RegisterRestClient` in a `rules.yaml: money_path_services` member) | **58** |
| …targeting an in-repo provider | 56 |
| …**with** a committed pact for that (consumer, provider) pair | **30** |
| …**without** | **26** |
| targeting something that is not an in-repo service (CNB feed) | 2 |

Committed pacts today: **57 files, 57 distinct (consumer, provider) pairs** — so the estate is healthy in the axis it already guards; these 26 are calls that never had a contract at all.

## The 26

| consumer | provider | client |
|---|---|---|
| account-service | sanctions-service | `SanctionsServiceClient.kt` |
| account-service | sca-service | `ScaChallengeClient.kt` |
| domestic-payment | aml-service | `AmlServiceClient.kt` |
| domestic-payment | clearing-simulator | `ClearingSimulatorClient.kt` |
| domestic-payment | fraud-service | `FraudScoreClient.kt` |
| domestic-payment | sanctions-service | `SanctionsServiceClient.kt` |
| fx-service | fraud-service | `FraudScoreClient.kt` |
| interest-service | account-service | `AccountServiceClient.kt` |
| interest-service | transaction-service | `TransactionServiceClient.kt` |
| lending-service | account-service | `AccountServiceClient.kt` |
| lending-service | consent-service | `CreditOfferGateAdapters.kt` |
[...]

PR #8552, opened 2026-09-03
test(vop): pin the account-service IBAN lookup with a consumer pact

JiRaska:
## Summary

Pins hop 1 of the ADR-0171 VoP name resolution — `GET /api/v1/accounts/iban/{iban}` — with a
consumer pact. It was a money-path cross-service HTTP call with **no pact at all**, which is the
coverage axis of #8345 and is invisible to `check-pact-provider-replay.py`: that gate proves every
committed pact is replayed before merge, and says nothing about a call that never had one.

No provider-side change: the contract reuses account-service's already-seeded
`an account owned by a known party exists` state.

## Type of change

- [x] chore (build / tooling) — test-only; `test(vop)` scope, so release-please cuts nothing

## Linked issues

Refs #8345

## Changes

- New `openbank-vop-service/src/test/kotlin/com/openbank/vop/contract/AccountIbanLookupPactConsumerTest.kt`.
- New `pacts/openbank-vop-service-openbank-account-service.json` (generated, committed).
- Corrects the now-stale KDoc on `PartyNameLookupPactConsumerTest`, which recorded that hop 1 could
  not be pinned because account-service's only `@Provider` class was message-only. That is no longer
  true — `AccountPactFolderProviderVerificationTest` boots Quarkus and dispatches per interaction.

## Why this call and not another

`AccountHolderNameLookupAdapter.lookupWithResilience` reads exactly one field off this response —
`partyId` — and a renamed, absent or blank one short-circuits the whole verification to `NO_DATA`.
`NO_DATA` is a *valid* VoP answer, so that failure is indistinguishable from the real thing in logs,
metrics, and every unit test in the module (the port is mocked there).

## Verified by effect — red and green, both directions

| case | layer | verdict (JUnit XML) |
|---|---|---|
| as committed | consumer | `tests="1" failures="0"` |
| as committed | account-service provider replay | `tests="5" failures="0"`, the new interaction listed by name |
| client `@Path` → `/by-iban/{iban}` | consumer | fails on the literal mismatch; pact adds `The following requests were not received` |
| DTO field `partyId` → `ownerPartyId` | consumer | fails `Expecting not blank but was: null` |
| pact path → `/api/v1/accounts/by-iban/...` | provider replay | fails `expected status of 200 but was 404 ... Unable to find matching target resource method` |

The last row is the finrep-class defect (#2269) this contract exists to catch, and it is the one
only the provider replay can see.

The expected path is a **literal**; only the outgoing request is reflected off the client's `@Path`
(CLAUDE.md "Contract tests", measured on #2290) — deriving both sides is vacuous. The response is
bound into the real `AccountSummary` DTO rather than read off a JsonPath, which is the half the
provider replay cannot see.

## Definition of Done

- [x] Hexagonal layering preserved (test-only, no production change).
- [x] Contract tests updated.
- [x] `:openbank-vop-service:ktlintCheck :openbank-vop-service:detekt` pass.
- [x] `check-pact-provider-replay.py` passes — 58 committed pacts, 58 replayed on a PR,
      `KNOWN_UNCOVERED` still empty.
- [x] `derive-pact-drift-scope.sh` already lists `:openbank-vop-service` (scope is derived from the
      `@Pact` annotations, not a hand-kept list).
- [x] No new lint warnings.
- [x] No OpenAPI / Flyway / Avro change.

## Security checklist

- [x] No secrets, tokens, passwords, or PII in code, config, tests, logs. The pinned IBAN, account id
      and party id are the fixtures account-service's own provider state already seeds.
- [x] No new `@Suppress`.
- [x] No auth / crypto / payment code changed.
- [x] No PII path changed.

## Compliance impact

- [x] None (test coverage only; the underlying flow is PSD2 VoP / SEPA 2025, unchanged here)


JiRaska:
Picking this up — the agent that opened it stalled before reporting, so here is the verification the body is missing, plus a merge of `main` (the branch was 11 commits behind; merged clean).

**The literal expected path is load-bearing, falsified in both directions.** `EXPECTED_ACCOUNT_PATH` is a `const val` whose KDoc says not to derive it from the client's `@Path`. Changing it to `/api/v1/accounts/by-iban/…`:

```
2 tests completed, 1 failed
AccountIbanLookupPactConsumerTest > … FAILED
    org.opentest4j.AssertionFailedError at AccountIbanLookupPactConsumerTest.kt:139
```

Restored → `BUILD SUCCESSFUL`. The baseline run before the injection was green, so that red is the injected defect and not a pre-existing failure.

That asymmetry is the whole value. The Pact mock server answers whatever path the client asks for, so deriving *both* the expectation and the outgoing request from the annotation is vacuous — they move together, and the test stays green against a route that does not exist. finrep-service shipped a call to `/api/v1/ledger/trial-balance`, a ledger path that has never existed, in exactly that shape.

**The provider replay actually runs on a PR:** `check-pact-provider-replay.py` → exit 0, *"every committed pact is either replayed on a PR or a declared, still-accurate exception."* Worth stating because a `@PactBroker` class would not have counted — the PR lane blanks `PACT_BROKER_URL`, so its `@EnabledIfSystemProperty` gate skips and the contract would only be replayed after the merge.

**Scope, so #8345's remainder is not overstated by this PR.** This is one hop. The coverage table for the rest of the money-path cross-service calls is not here — the run stalled partway and I finished the piece that was complete rather than pushing a half-measured table. `settlement → balance`, which it had just started, is not in this branch.

The gate that exists guards a different axis: it guarantees every *committed* pact is replayed, and is silent about a call that never had one. That is why this hop was invisible rather than merely uncovered. Whoever continues #8345 should enumerate `@RegisterRestClient` interfaces per `rules.yaml: money_path_services` against `pacts/`, stating the command, before writing more pacts.

PR #8567, opened 2026-09-03
fix(sca): refuse TOTP instead of minting a challenge nobody can satisfy

JiRaska:
Found while closing out the last three templates of #8432. This is the concrete half of `OTP_CODE`; the rest is a decision, written up in the issue.

## The defect

`ScaService.initiate` takes the method straight off the request body:

```kotlin
val method = command.preferredMethod ?: ScaMethod.PUSH_NOTIFICATION
```

and then, for TOTP:

```kotlin
ScaMethod.TOTP -> {
    val otp = otpGenerator.generate()
    otpStore.store(saved.id, otp, ttlSeconds)
}
```

The code is generated, stored with a TTL — and **sent absolutely nowhere**. `PUSH_NOTIFICATION` and `BIOMETRIC` dispatch to the enrolled device; TOTP has no delivery step at all, and there is no transport for it in the fleet (no SMS; `OTP_CODE`, the template that would carry one, has never had a producer).

So a caller selecting TOTP gets a challenge **the customer can never satisfy**, and whatever it authorised — a payment, a consent, a card action — cannot proceed.

## What this is, and what it is not

**Not an authentication bypass, and never was.** `ScaChallenge.fail` increments `attemptCount` and fails the challenge at `maxAttempts`, so an undelivered code cannot be brute-forced within the challenge's life. I checked this before writing the description rather than after.

**It is a denial-of-authorisation path**, reachable by anyone who can call `initiate`, because `preferredMethod` was accepted verbatim with no validation and no enrolment check.

Nothing requests TOTP today — the app never sends it and no service passes it — so this is a latent trap rather than a live outage. That is also why the fix is cheap now.

## The fix

Refuse the method **before a challenge exists**, with 422 (the request is well formed; this deployment cannot deliver the method — the same status an expired challenge returns).

- `PUSH_NOTIFICATION` and `BIOMETRIC` are unaffected.
- A challenge **already stored** with method TOTP still verifies, so nothing in flight breaks.
- The enum value stays; only minting a new challenge with it is refused.

**Why not build real OTP delivery instead.** A one-time code pushed to the same device that made the request is not a second factor, notification-service renders and persists template variables (the concern already recorded against `ACCOUNT_FROZEN`'s `code` in #1325), and `OTP_CODE.noDeviceFallbackChannel` is deliberately `null`, so there is no email fallback either. Making OTP real is a security decision, not wiring — until it is taken, refusing beats half-working.

## A test I deleted, and why

`initiate generates OTP for TOTP method` asserted that `initiate` **stored** a generated code — and never that anything delivered it, because nothing did. It locked the defect in place. Superseded by `initiate refuses TOTP instead of minting a challenge nobody can satisfy`.

`verify completes challenge on successful OTP verification` is **kept deliberately**: it builds a TOTP challenge directly rather than through `initiate`, and any challenge already stored with that method must still verify.

## Verification

- `ScaServiceTest` — **54 tests, 0 failures**; module suite **148 tests, 0 failures**.
- **Guard falsified:** deleting the refusal line reddens exactly `initiate refuses TOTP instead of minting a challenge nobody can satisfy`, and nothing else. Restored and re-run green.
- The refusal test also asserts the negative effects — `repository.save` and `otpStore.store` are never called for a refused method — so it fails if the refusal moves to *after* the challenge is minted, not only if it disappears.
- `ktlintCheck` + `detekt` with `--rerun-tasks`: clean.

## Security checklist


JiRaska:

```
```

```markdown
## Security checklist

      (license, CVE, source).
```

PR #8606, opened 2026-09-03
docs(kyc,onboarding): drop DOCUMENTS_REQUIRED from the two specs that describe it

JiRaska:
Spec-only half of #8544, split out because the api-contract gate requires it to be.

## Why a separate PR

`DOCUMENTS_REQUIRED` is a state no KYC case can enter (#8535): no operation in `KycService` sets it, and onboarding-service derives its own stage from the value kyc-service emits. Removing it from these two specs therefore leaves the **served contract byte-identical** — the documents described something the running services could not produce.

`oasdiff` classifies an enum narrowing as **breaking**, and the gate then demands an `info.version` MAJOR bump — which ADR-0048 D2 forbids while the URL major stays `/api/v1`. Jointly unsatisfiable, which is exactly what the gate's **`correction`** class exists for, in its own words:

> a breaking DOCUMENT diff in a PR that changes nothing else in the service: the served contract is unchanged, so MINOR

That class needs the change to be **spec-only**. In #8544 it was not — the same PR removes the state from Kotlin, tests, docs and the lifecycle diagram — so the gate correctly refused it there:

```
api-contract gate: openbank-kyc-service: breaking, and the PR also changes 7 other file(s) in the service
api-contract gate: openbank-onboarding-service: breaking, and the PR also changes 9 other file(s) in the service
```

Hence this split. The customer-edge spec stays in #8544: the gate classified that document **additive**, not breaking, and it passes at 1.52.0.

## Contents

| Spec | Version | Change |
|---|---|---|
| `openbank-kyc-service` | 1.8.0 → **1.9.0** | `DOCUMENTS_REQUIRED` out of three `KycCaseStatus` enums |
| `openbank-onboarding-service` | 1.0.1 → **1.1.0** | `DOCUMENTS_REQUIRED` out of the kyc-status enum; `KYC_DOCUMENTS_REQUIRED` out of the funnel-stage enum, its `required` list and its example |

Nothing else. No Kotlin, no docs, no gitops.

## Merge order

Either side of #8544. The value is unreachable from both directions, so a window where the spec has dropped it and the code has not is not observable by any client.


JiRaska:
Held back, and taken out of the merge queue. Red on `openapi-enum-domain-drift` and nothing else.

Removing this value cannot pass CI in any PR shape: `api-contract-gate` wants the change to be spec-only (its `correction` class, since a MAJOR is forbidden by D2 while the URL major stays /api/v1), while `openapi-enum-domain-drift` wants the spec and the Kotlin enum to match. Spec-only, Kotlin-only and combined were all measured; each fails one of the two.

Written up with the measurements and four options in #8618. The other six PRs from this batch are independent and unaffected.

JiRaska:
These two PRs are **complementary halves of one removal**, and neither can go green alone. Each is red on `openapi-enum-domain-drift` *because the other half is missing*.

Read from each run’s `gate-results-gitops-api` artifact and the gate’s own output (the job log also echoes the PR body, which is what makes a plain grep misleading here):

```
#8544  spec advertises values the code does not have: [DOCUMENTS_REQUIRED]      (FunnelStage, KycCaseStatus, KycStage)
#8606  spec omits [DOCUMENTS_REQUIRED]                                          (FunnelStage, KycCaseStatus, KycStage)
```

Opposite directions, same three enums, same value — because the removal is split:

| | removes `DOCUMENTS_REQUIRED` / `KYC_DOCUMENTS_REQUIRED` from |
|---|---|
| **#8544** | the Kotlin enums (`KycCase.kt`, `OnboardingRecord.kt`), `openbank-customer-edge/openapi.yaml`, admin-ui, the lifecycle diagram, 8 doc files, 2 test files |
| **#8606** | `openbank-kyc-service/openapi.yaml` and `openbank-onboarding-service/openapi.yaml` — the two specs #8544 does not touch |

**The file sets do not overlap at all** — `comm -12` over the two diffs is empty. So there is no conflict to resolve; there is a dependency neither PR declares.

`all-green` is required, so this is a deadlock rather than a race: #8544 waits for the two specs #8606 holds, #8606 waits for the code #8544 holds, and no amount of re-running moves either.

**Cheapest resolution — stack them.** Merge `#8544`’s branch into `#8606`’s (the repo’s documented move for exactly this: *“merging the parent branch into the child makes the child’s CI green independently; the shared files drop out of the diff as the parent lands on main”*). #8606 is 2 files, so it is the natural child. Alternatively #8544 absorbs the two `openapi.yaml` edits and #8606 closes.

**Do not merge #8544 alone.** With the code gone and the kyc/onboarding specs still advertising the value, `main` ends in the state that reds this gate for the *next* PR to touch either service — the drift becomes someone else’s failure.

**Two more things specific to this PR:**

1. It claims `openbank-onboarding-service` `info.version` **1.1.0**, and so does #8710 — both from base 1.0.1. Whoever merges second must take 1.2.0.
2. Its body’s reasoning is worth keeping with the stacked PR: removing a value narrows a published contract and is strictly breaking, but MINOR is right here because the value was never producible, so no client can have received it.

JiRaska:
Folded back into #8544 and closing.

Splitting the change was my attempt to satisfy `api-contract-gate`'s `correction` class, which needs a spec-only PR. It worked for that gate and broke a different one: split apart, **each half fails `openapi-enum-domain-drift`** — this one had the spec without the Kotlin, #8544 had the Kotlin without the spec. Measured on both.

Recombined, the two satisfy the drift gate and fail only `api-contract-gate`, which is the single well-defined deadlock written up in #8618. One red PR with one known cause is easier to decide on than two with a confusing one.

Issue #4414, opened 2026-08-09
ADR-0250 Phase 2: split _service-ci.yml's main-push build off the ARC pool (Gradle provider-test isolation needed)


JiRaska:
ADR-0250 ("Decouple the main-push service build from the Pact Broker runner pool") is now
`decision-status: accepted`. Phase 1 (raise `arc_max_runners` 6→12, retire `runners-warm`)
shipped in #4319 before acceptance. This issue tracks Phase 2 — the actual fix, not yet built.

## Why Phase 1 alone is not enough

Re-measured 2026-08-09, ~7h after #4319 merged: **queue depth 57** (was 42 at proposal time),
in-progress 5 (down from 6-of-6 saturated — the pool IS draining faster), oldest queued entry
~18h. The queue did not clear, and briefly grew to 58 during the remeasurement window. Doubling
a scarce resource delays saturation on a high-commit-volume day; it does not remove the
dependency that makes the resource scarce in the first place.

## What Phase 2 is

Split `_service-ci.yml`'s main-push job: `build` (compile, test, quarkusBuild, consumer-pact
generation) moves to `ubuntu-latest` — free, unlimited for a public repo — for every event
including main-push. `contract` (in-cluster, `openbank-build`) does only what needs cluster
DNS: publish consumer pacts (a 0-1s curl, already measured) and run provider verification.

## Why this needs a real Gradle/JVM session to build, and can't be done blind

Investigated 2026-08-09 by reading `_service-ci.yml` and the shared convention plugin
(`build-logic/src/main/kotlin/openbank.quarkus-service.gradle.kts`) — three findings that
make this Gradle build-logic work, not a workflow-file edit:

1. **Every provider-verification test class ends in `ProviderVerificationTest`** — confirmed
   across all 26 services with `@Provider(` annotations, split between `*PactFolder*` (no
   broker needed, safe anywhere) and `*Pact(Broker)?ProviderVerificationTest` (needs
   `pactbroker.url`). The isolation target is unambiguous.

2. **A Gradle `--tests` filter alone does NOT achieve the goal.** `-Dpactbroker.url` is already
   a tracked input of the `test` task (the #1009 cache hazard this repo already documents), so
   `contract`'s `test` invocation (broker URL set) is *always* a different cache key from
   `build`'s (broker URL blank) — filtering which tests *run* doesn't change what Gradle
   considers the task's *inputs*, so `contract` still pays full compile + dependency
[...]

PR #8665, opened 2026-09-04
test(transaction): pin the lending and standing-order calls to POST /api/v1/transactions

JiRaska:
## What

Adds consumer-driven contracts for **two of the five** money-path calls into
`openbank-transaction-service` that #8345 measured as having no pact at all, plus the one provider
state they needed. No new provider verification class — `TransactionPactFolderProviderVerificationTest`
already runs on every PR and now backs **eight** interactions instead of six.

| consumer | client | shape this pins |
|---|---|---|
| `openbank-lending-service` | `BorrowerCreditClient.credit` | `type=CREDIT`, `targetAccountId` set, **`sourceAccountId` absent**, no `rail`, no `instructionType` |
| `openbank-standing-order-service` | `StandingOrderDueConsumer` own-account execution | `type=TRANSFER` with **no `rail`**, both legs internal |

## Why these two

The four contracts transaction-service already holds — domestic-payment, sepa-payment, sepa-instant,
mcp-service — all pin the *same* payload: a `DEBIT` carrying a non-null `sourceAccountId` plus `rail`
and `instructionType`. Neither shape above had any coverage.

- **lending** is the only one of the five that **binds the response into a DTO**.
  `TransactionServiceRestClient.initiate` returns `Uni<TransactionAck>` and `TransactionAck.id` is a
  non-null `UUID` with no default, so a renamed or dropped `id` is not a degraded read — Jackson
  fails to construct the value, the `Uni` fails, `@Retry` burns three attempts and the disbursement
  records a failure. Every unit test in that module mocks `BorrowerCreditPort`, so nothing else in
  this repo observes that binding.
- **standing-order** pins `(TRANSFER, no rail)` — transaction-service's *own* discriminator for
  "the money never leaves the ledger" (`TransactionService.initiateTransactionInternal`, #5225). It
  bypasses `SettlementDateResolver`'s cutoff and business-day rules and books same-day. The client's
  KDoc asserts that behaviour; until now nothing replayed it. That consumer branches only on the
  status family, so a 400 on a rail-less `TRANSFER` would record a failure on every due date until
  the order auto-suspended, with every unit test green.

Read from the JUnit XML, not the console — pact-jvm prints `Not all of the N were verified` even on
a fully green run.

| run | tests | failures | skipped |
|---|---|---|---|
| provider replay, pacts as committed | 8 | 0 | 0 |
| provider replay, each **new** pact's request path repointed at a route transaction-service does not serve | 8 | **2** | 0 |
| consumer tests with `EXPECTED_TRANSACTIONS_PATH` repointed the same way | 2 | **2** | 0 |
| consumer tests restored | 2 | 0 | 0 |

The two provider-replay failures are exactly the two new interactions — the six pre-existing ones
stayed green — and both fail with:

```
1.1) status: expected status of 201 but was 405
1.2) body: $ Actual map is missing the following keys: id
```

After restoring, the regenerated pacts are **byte-identical** to the pre-tamper files, and
`find pacts -size 0` is empty.

**The expected path is a LITERAL; only the outgoing request is reflected off the client's `@Path`**
(#2290). Deriving both sides moves them together and stays green against a route that does not
exist — how finrep-service shipped a call to a ledger path that has never existed (#2269).
`assertClientPathMatchesContract` is the consumer-side half; the provider replay is what actually
adjudicates whether the route is served.

## Provider-side change

One new `@State` — `a valid borrower account exists` — for the credit-with-no-source payload. Added
to **both** the `@PactFolder` class and its `@PactBroker` twin together, as that class's own KDoc
requires: pact-jvm passes silently over an unhandled state name, which is how #468's missing states
stayed invisible. Reusing `a valid source account exists` would have been false about a payload that
has no source account.

Also replaced a KDoc sentence that hard-counted "Six pacts, five HTTP and one message" with the
mechanism, since that count went stale in this very PR — the CLAUDE.md rule that a comment asserting
current configuration has a shelf life and nothing re-checks it.

## Gates

- `check-pact-provider-replay.py` — **59 committed pacts, 59 replayed on a PR**, `KNOWN_UNCOVERED`
  still empty.
- `derive-pact-drift-scope.sh` — now derives `:openbank-standing-order-service:test` into the drift
  scope from the new `@Pact` annotation, so the regeneration check covers it. `openbank-lending-service`
  was already in scope.
- `ktlintCheck` + `detekt` green on all three touched modules.

`openbank-standing-order-service` had no pact tooling at all; the whole per-module wiring is one
`testImplementation(libs.pact.consumer)` line, because `pact.rootDir` and the broker forwarding are
centralised in the `openbank.quarkus-service` convention plugin (ADR-0250 Phase 2, #4414).

## Deliberately left

Three of the five remain uncovered against transaction-service — **`interest-service`,
`sdd-service`, `swift-service`**. All three send the `DEBIT` + `rail` + `instructionType` shape the
four existing contracts already pin, so their marginal value is the lowest of the five. A follow-up
adding them needs no provider-side change at all: the reusable `a valid source account exists` state
and the always-running replay class are both already there.

Refs #8345

## Security checklist

      (license, CVE, source).


JiRaska:

```
BUILD FAILED in 14m 31s
```

Issue #8510, opened 2026-09-03
balance and psd2 ship a full transactional outbox nothing ever writes to — and psd2's dispatcher is off by default too


JiRaska:
## What

`openbank-balance-service` and `openbank-psd2-service` each ship the **complete** transactional-outbox apparatus with **nothing that ever writes a row**:

| | balance | psd2 |
|---|---|---|
| entity | `BalanceOutboxEntity.kt` | `Psd2OutboxEntity.kt` |
| dispatcher | `BalanceOutboxDispatcher.kt` | `Psd2OutboxDispatcher.kt` |
| backlog gauge | `BalanceOutboxBacklogGauge.kt` | `Psd2OutboxBacklogGauge.kt` |
| migration | `V11__balance_outbox_claimed_at.sql` | `V1__create_psd2_outbox.sql` |
| `openbank.outbox.dispatch-enabled` | `true` | **absent → defaults `false`** |
| `OutboxMessage(` constructions in `src/main` | **0** | **0** |

```
git grep -c "OutboxMessage(" origin/main -- '<svc>/src/main/*.kt'      -> no output (zero files)
git grep -n "persistInTransaction" origin/main -- '<svc>/src/main/*.kt'
   balance -> one hit, the port DECLARATION in BalanceOutboxRepository.kt:21
   psd2    -> none
```

So the tables are empty by construction. The gauge reports 0 and the dispatcher reports success, and **neither is distinguishable from a healthy, drained outbox** — the same "a no-op sharing a signal with success" shape as `openbank-security-scanner` (#3973).

**psd2 is a double defect.** Its `outbox:` block carries `poll-interval` and `initial-delay` but no `dispatch-enabled`, which defaults to `false` — so even if something did write a row, the dispatcher would never send it. That is the repo's documented #1 new-service silent failure: no error, no log, `attempt_count` stays 0.

## Relationship to #4007, which is closed and was closed correctly

#4007 swept this exact pattern and split it in two: **Group A** (no write side at all — party, interest, audit, sepa-instant) and **Group B** (write side exists, table merely empty in this environment). It resolved Group A and explicitly declined to claim Group B as a defect: *"the code path is real but has never fired in this environment… I am not claiming a defect for these."* Its closure is sound.

[...]

Issue #8509, opened 2026-09-03
clearing-service settle is not atomic: batch, items and outbox row commit in three separate transactions (measured xmin 750/752)


JiRaska:
## What

`ClearingService.settleBatch` writes the batch, its items and the outbox row in **three separate transactions**. A crash after the batch commits `SETTLED` loses `openbank.clearing.batch.settled` permanently, with the batch reading settled and no event ever produced — the exact failure the transactional-outbox pattern exists to prevent.

Measured, not inferred. Running the #8496 `xmin` oracle against unmodified `origin/main`:

```
batch xmin = 750
outbox xmin = 752
```

(751 was consumed by the item write in between.) Postgres stamps each row version with `xmin`, the id of the writing transaction, so three distinct values are three distinct transactions.

## The chain

`ClearingService.kt:151` —

```kotlin
batchRepo.update(settled).flatMap { savedBatch ->
    ...
    itemRepo.saveAll(updatedItems).flatMap {
        eventPublisher.publishBatchSettled(savedBatch).map { savedBatch }
```

and `ClearingEventPublisherImpl.kt:46` opens its own:

```kotlin
return Panache.withTransaction { outboxRepo.persistInTransaction(message) }
```

`publishItemCleared` has the same shape.

## Why nothing noticed

The existing outbox ITs across the fleet assert that the outbox row **landed**. A three-transaction implementation satisfies every presence assertion — the row is there, it just did not commit with the state change. That is why #8496 replaced the oracle with an `xmin` comparison, and this is the first defect it found.

The test for this case is deliberately **excluded** from #8507: it would be red on `main`, and a test PR that also changes production code is the wrong shape for review.

## Suggested fix

Pass the ambient transaction into the publisher rather than letting it open its own — the shape `saveWithEvent` / `saveWithOutbox` already use in sanctions, standing-order and fx, all three of which pass the `xmin` check (measured in #8507). Then enable the excluded test in the same PR, so the property is pinned rather than asserted.

Note for whoever takes it: the test must switch `openbank.outbox.dispatch-enabled` **off** for its own run, because the dispatcher's UPDATE rewrites the outbox row's `xmin` and destroys the evidence.

clearing-service is money-path, so this needs two approvals and a threat-model check.

Refs #8353, #8496, #8507

JiRaska:
Vyřešeno v #8621 (squash-mergnuto jako bae4b498a3, CI 23× zelené vč. buildu clearing-service).

[...]

Issue #4353, opened 2026-08-09
Push reaches 2 of 15 onboarded customers: device registration fails silently on the app side, and the backend records that a delivery failed but not why


JiRaska:
## Summary

Push notifications do not reach the overwhelming majority of onboarded customers, and have not since the feature shipped. Measured against the live databases on 2026-08-08:

| | |
|---|---|
| PUSH notifications FAILED | 53 |
| PUSH notifications SENT | 12 |
| **failure rate** | **81%** |
| Distinct parties with a failed push | 43 |
| …of which completed onboarding | 42 |
| Fully onboarded parties (`ACTIVE` + KYC `APPROVED`) | 15 |
| …with an ACTIVE device token | 2 |
| **…with no device token at all** | **13 (87%)** |

Device tokens have only ever been registered **three times**, all between 16 and 20 July, all from a single app install. `app_version` and `os_version` are `NULL` on every one, because the app has never sent them.

The sharpest case: **11 failed `SCA_APPROVAL` pushes across 6 distinct parties, continuously from 2026-07-24 to 2026-08-08.** That is the prompt telling a customer a payment is waiting on their approval.

## Impact, stated precisely

A customer with no registered device is **not locked out**. `openbank-app` also lists live challenges via `GET /customer/v1/sca/pending`, so someone who opens the app can still approve. What they lose is any *notice* that they need to — for a payment they initiated themselves, the experience is that it silently does not complete.

Everything else (`TRANSACTION_COMPLETED`, 41 failures) is informational and simply never arrives.

## Why nobody noticed

Three independent silences, which is why an 81% failure rate survived three weeks:

1. **The app swallowed every registration failure.** `runCatching {}` with no `onFailure`; the branch for "the OS has not issued a push token yet, or notification permission was never granted" did nothing at all. Worse, the HTTP client is built without `expectSuccess`, so a `401`/`403` — the likeliest failure — returned *normally*: error handling at the call site alone could not have seen it.
2. **The backend stored that a delivery failed, never why.** `NotificationConsumer` computes `REASON_NO_DEVICE` vs `REASON_PUSH_REJECTED` and puts it on the outcome event, but the `notifications` row carried only `status`, and the outbox rows behind that event are pruned after dispatch. The two need completely different fixes and the database could not tell them apart.
3. **No metric existed for delivery outcomes**, so no alert rule could have been written against the failure ratio, and none was.

[...]

PR #8676, opened 2026-09-04
test(account): prove outbox atomicity by transaction id in account, sdd and fraud

JiRaska:
## What

The third tranche of #8353 — an outbox-atomicity integration test for three more money-path
services: **account**, **sdd**, **fraud**. Follows the oracle #8496 established and #8507 repeated,
verbatim: compare the Postgres `xmin` of the aggregate row and its outbox row instead of asserting
the outbox row is merely present.

## How these three were chosen

Same criteria as #8496/#8507, applied to the remainder #8507 named
(`account`, `swift`, `sdd`, `interest`, `sca`, `fraud`, `billing`):

- **money-path weight.** All three are in `rules.yaml: money_path_services`. `account` holds the
  balances themselves; `sdd` is the standing authority under which `SddCollectionDebitConsumer`
  posts a real debit for every authorised collection (#1478); `fraud`'s hold is the signal every
  downstream consumer acts on when a party is restricted.
- **the write path is genuinely co-transactional with a state change.** Established by reading
  every outbox-write call site in each candidate, not by naming — which is also how the `sca`
  finding below fell out.
- **a proven-green REST-driven `@QuarkusTest` already exists in the module**, so a boot failure
  cannot masquerade as a pass: `SavingsProposalIT` (RestAssured + `@TestSecurity`, and the exact
  delegation arrangement this test needs), `SddBootSmokeIT`, `FraudApiIT`.

Each write path was run against a **deliberately broken** build — its single transaction split in
two — and goes red naming the two transaction ids:

| service | write path | REST entry | broken-build failure |
|---|---|---|---|
| account | `WithdrawalProposalRepositoryImpl.save(proposal, event)` | `POST .../proposals/{id}/decide` | proposal xmin=784, outbox xmin=785 |
| sdd | `register` | `POST /api/v1/sdd/mandates` | mandate xmin=741, outbox xmin=742 |
| sdd | `transition` (suspend/resume/cancel/confirm) | `POST /api/v1/sdd/mandates/{id}/suspend` | mandate xmin=742, outbox xmin=743 |
| fraud | `FraudHoldService.maybeRaise` | `POST /api/v1/fraud/score` | hold xmin=750, outbox xmin=751 |

The source was restored after each run (`git checkout --`) and the tests re-run green. This PR
contains **test files only**, plus one `build.gradle.kts` test-scope dependency explained below.

Module runs, reading the **SKIPPED** count and not just failures — a service that cannot boot
reports its tests as SKIPPED, and a skip count scans as a pass:

|---|---|---|---|
| `openbank-account-service` | 271 | 1 | 0 |
| `openbank-sdd-service` | 63 | 0 | 0 |
| `openbank-fraud-service` | 187 | 1 | 0 |

Both skips are the broker-sourced Pact provider classes (`AccountEventPactProviderVerificationTest`,
`FraudPactProviderVerificationTest` — `@PactBroker` + `@EnabledIfSystemProperty(pactbroker.url)`),
pre-existing and expected on a PR lane. `ktlintCheck` + `detekt` were run as a **separate** Gradle
invocation after `test` — Gradle stops at the first failing task, so a green-looking task list is
not a list that ran — clean on all three. No module OOM'd, and `find pacts -size 0` is empty.

One run in this session did fail with `QuarkusBindException: Port already bound: 8081` from a
parallel Gradle build on the same machine; it was re-run before being believed, per the repo's own
note that that failure is indistinguishable from a real one at a glance.

## The dispatcher, and why the switch is in a PROFILE and not a test resource

`xmin` belongs to a row *version*, so any later UPDATE replaces it. The scheduled outbox dispatcher
claims rows and marks them DISPATCHING/SENT, so it is switched **off** per class. All three services
correctly ship `openbank.outbox.dispatch-enabled: true` in `application.yaml` — the fleet default is
`false`, under which a service dispatches nothing and reports no error — which is precisely why it
has to be disabled here rather than merely left alone.

Where #8496 and #8507 folded that switch into the `@QuarkusTestResource`, this PR puts it in a
`QuarkusTestProfile`. That is not a style preference: a `@QuarkusTestResource` is applied to **every**
test class in the module, and doing it that way turned `SddOutboxDispatchIT` red — that IT drives
`dispatcher.dispatch()` explicitly, and `dispatch()` is a no-op when the flag is false, so the whole
class stopped testing anything. Measured, then fixed; a profile is per-class and forces this class
its own Quarkus boot, which is exactly the scope wanted. Worth knowing before the next tranche
copies the resource form into a module that has a dispatch IT.

sdd's profile carries a second override, `authz.enforce=false`, and it is not incidental: **sdd is
one of the few services shipping `AUTHZ_ENFORCE` defaulted to `true`** (account and fraud both
default it to `false`). With no OPA sidecar reachable the interceptor fails closed, so every write
endpoint answers **503** — not 403 — and without the override the test would be red for a reason
that has nothing to do with the outbox. That is also why sdd had no `@TestSecurity` dependency
before: its existing ITs only ever hit unsecured routes.

## Controls, so the assertions cannot pass vacuously

- **Known-negative** in all three classes: an id that was never written must return no pair at all,
  so `hasSize(1)` is a claim the query is capable of failing.
- **Known-different**, so one run shows the identical comparison both matching and not:
  - *sdd* gets it for free — the suspend UPDATEs the mandate row, so the **registration** outbox row
    was genuinely written by an earlier transaction and must NOT match the current mandate version.
  - *account* and *fraud* drive the write twice (a second proposal / a second party) and require the
    identical comparison to fail across them.
- **A second, structural control in account**: the REJECT branch calls `save(proposal)` — the
  overload with no event — so it must write **no** outbox row. That keeps `hasSize(1)` honest about
  *which* write path produced the row.
- **Arrangement assertions**, so a change that moves the test onto a different write path fails here
  rather than silently asserting over an empty outbox: account asserts the decision came back
  `APPROVED`/`REJECTED`, sdd asserts `ACTIVE` then `SUSPENDED` (only a CORE mandate is born ACTIVE,
  and only an ACTIVE one can be suspended), fraud asserts every scoring call returned `REVIEW` (only
  a REVIEW counts towards `review-threshold: 3`, and only the third call reaches the write).

## One collaborator that had to be pinned, and one that did not

`fraud`'s `maybeRaise` returns early when the account maps to no party, so with no account-service
reachable the test would assert nothing while looking like it had. `StubAccountPartyLookupPort`
(`@Alternative @Priority(1)`, declared alongside `FraudHoldServiceIT`) already replaces that port
for the module; this class only sets which party it answers with. The ML shadow model, by contrast,
runs after the write and fails open by design, so it needs no stub — its "model card verification
FAILED" lines in the log are the designed behaviour, not a defect.

## Finding, reported and NOT fixed here

**`openbank-sca-service` is NOT atomic — measured, not inferred.** `ScaService.enroll` awaits
`enrolledDeviceRepository.save(...)`, whose `EnrolledDeviceRepositoryImpl.save` opens and commits
its **own** `Panache.withTransaction`, and then calls `outboxRepository.save(...)`, whose
`ScaOutboxRepositoryImpl.save` opens a **second** one. Nothing wraps the pair: `ScaResource` carries
no `@WithTransaction` anywhere, and `enroll` is a `suspend fun`, so there is no ambient transaction
for either to join. A throwaway copy of this PR's test, run against unmodified `main`, answered
`device xmin=751, outbox xmin=752` — two transactions.

So a crash between the two commits leaves a device enrolled with `sca.device_enrolled` never
emitted, permanently, and every presence-based assertion stays green about it. That event already
has history: onboarding-service discarded it for months (#4353), so a silently-lost one is not
hypothetical. It is `sca`'s **only** outbox write — `outboxRepository` has exactly one call site in
`src/main` — so the service's whole outbox is on this path.

A test for it is deliberately **not** in this PR: it would be red on `main`. Same shape as #8509
(clearing) and a sibling of #8510 (balance/psd2's outbox with no producer).

## Remaining scope for #8353

Money-path, has an outbox entity, still only a `*OutboxClaimIT`: `balance`, `clearing`, `swift`,
`interest`, `sca`, `billing`, `psd2` — of which `balance`/`psd2` are the dead-outbox finding
(#8510), and `clearing`/`sca` are blocked on the atomicity defects above (#8509 and this PR).
`swift`, `interest` and `billing` remain genuinely testable and were passed over here only for the
stubbing/four-eyes reasons #8507 already recorded.

Refs #8353

## Security checklist

      (license, CVE, source). `testImplementation(libs.quarkus.test.security)` in sdd-service's
      `build.gradle.kts` is an existing catalog entry already used by other services' test suites
      (e.g. account-service) — not a new dependency to the tree.
- [x] Auth / crypto / payment code changed? → tag `security-review-required`. Test-only PR:
      3 new integration tests asserting outbox-row atomicity by transaction id across
      account/fraud/sdd. No production code touched.
- [x] PII path changed? → tag `gdpr-review-required`. No.
- [x] Cardholder data path changed? → tag `pci-review-required`. No.

### DIFF
--- a/openbank-account-service/src/test/kotlin/com/openbank/account/integration/AccountOutboxAtomicityIT.kt
+++ b/openbank-account-service/src/test/kotlin/com/openbank/account/integration/AccountOutboxAtomicityIT.kt
@@ -1,0 +1,292 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) OpenBank contributors. Licensed under the Apache License, Version 2.0.
+// See LICENSE in the repository root or https://www.apache.org/licenses/LICENSE-2.0 for details.
+
+package com.openbank.account.integration
+
+import com.openbank.account.it.PostgresRedpandaRedisTestResource
+import com.openbank.account.it.StubScaChallengeClient
+import io.quarkus.test.common.QuarkusTestResource
+import io.quarkus.test.common.QuarkusTestResourceLifecycleManager
+import io.quarkus.test.junit.QuarkusTest
+import io.quarkus.test.junit.QuarkusTestProfile
+import io.quarkus.test.junit.TestProfile
+import io.quarkus.test.security.TestSecurity
+import io.restassured.module.kotlin.extensions.Given
+import io.restassured.module.kotlin.extensions.Then
+import io.restassured.module.kotlin.extensions.When
+import io.smallrye.reactive.messaging.memory.InMemoryConnector
+import io.smallrye.reactive.messaging.memory.InMemorySource
+import jakarta.enterprise.inject.Any
+import jakarta.inject.Inject
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.runBlocking
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+import java.util.UUID
+import javax.sql.DataSource
+
+/**
+ * Issue #8353 — proves that `WithdrawalProposalRepositoryImpl.save(proposal, event)` commits the
+ * `savings_withdrawal_proposals` row and its `account_outbox` row in **one** database transaction,
+ * so neither can exist without the other.
+ *
+ * The subject is the owner's SCA-bound APPROVAL of a delegate's savings-withdrawal proposal
+ * (`POST .../proposals/{id}/decide` with `approve: true`) — the only call site in this service that
+ * writes the aggregate and an outbox event together. The sibling REJECT branch deliberately writes
+ * no outbox row at all: `save(proposal)` without an event, which this test uses as a control.
+ *
+ * ### Why presence is not the property
+ *
+ * The house pattern (`LendingOutboxWriteIT`, `ConsentRevocationOutboxIT`) drives the flow through
+ * the real REST endpoint and asserts the outbox row landed. That is necessary — a mocked repository
+ * commits nothing, and a reactive Panache repo cannot be called from a bare `@QuarkusTest` thread
+ * ("No current Vertx context found"), so only a real HTTP request can exercise the write — but it
+ * is **not sufficient**: an implementation that persisted the aggregate in one transaction and the
+ * outbox row in a second would satisfy every presence assertion while having lost the property.
+ * The sibling [SavingsProposalIT] is in exactly that position — it asserts the decision returned
+ * `APPROVED`, which a two-transaction save answers identically.
+ *
+ * ### What makes it falsifiable
+ *
+ * Postgres stamps every row version with `xmin`, the id of the transaction that wrote it. Two rows
+ * written by one transaction carry the *same* `xmin`; two rows written by two transactions cannot.
+ * Comparing the two `xmin`s is a direct read of the property rather than of its happy-path shadow —
+ * splitting `save`'s single `Panache.withTransaction` in two turns this test red, where a presence
+ * assertion stays green.
+ *
+ * The scheduled outbox dispatcher is switched off for this class: it UPDATEs claimed rows, and an
+ * UPDATE writes a new row version with a *new* `xmin`, which would race the assertion. (This
+ * service correctly ships `openbank.outbox.dispatch-enabled: true` in `application.yaml` — the
+ * fleet default is `false`, under which a service dispatches nothing and reports no error — which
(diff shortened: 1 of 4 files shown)


JiRaska:

```
BUILD FAILED in 14m 31s
```

Issue #8690, opened 2026-09-04
admin-ui-deploy spends ~345 API requests per build — more than the whole installation quota at 4 builds/hour


JiRaska:
## The measurement

`admin-ui-deploy`'s build job is the estate's largest consumer of the **installation** API quota — 1000/hour, shared by every workflow in the repo. Counted from the evidence lines of a real build (run `33808810484`):

| step | requests |
|---|---|
| Test Intelligence per-service | ~123 (62 name-filtered lists + 61 zips) |
| immutable run history | 105 |
| gate-health | ~41 |
| snapshot history | ≤35 |
| customer-app / pitest / perf / SBOM | ~41 |
| **total** | **~345 per build** |

At ~4 builds/hour that is **~1380 requests/hour — more than the entire quota**. For scale, `main-red-watch` is 166/hour.

This is not hypothetical: the quota was exhausted at **2026-09-03 08:01Z**, and Trivy's SARIF upload plus `dependency-review` failed on every branch including `main` for ~20 minutes. Neither is a required check and `all-green` does not gate on them, so nothing was blocked — **Code Scanning silently stopped being updated**. Reruns went green on unchanged commits after the hourly reset.

#8XXX (this PR's sibling) trims ~62 requests/build with four measured parameter changes. That is −18% and it is **not enough**: ~1130/hour still exceeds 1000. The rest needs a structural decision, and there are two.

---

## Option A — make the history worth what it costs (retention)

The `immutable run history` step spends 105 requests to stage run envelopes. Measured what those buy, 2026-09-04:

```
54 non-expired main-branch test-intelligence-run-* artifacts exist repo-wide
   21 distinct components
   10 distinct CI runs
   0.6 HOURS of span
   median 2 envelopes per component
```

That history feeds `testCaseHistory()` — per-test-case state over time, i.e. the flakiness view. **Two points per service across 36 minutes is a snapshot, not a trend.** The step is paying a per-build price for something that cannot answer the question it exists for.

The binding constraint is **artifact retention**, not the cap. `test-intelligence-run-*` expires before enough accumulates. Raising retention for that one artifact name (they are small JSON envelopes) would make the same ~55 requests buy days instead of minutes.

**Decision needed:** what retention, and is the storage cost acceptable? Until then the step is expensive *and* uninformative — the worst quadrant.

## Option B — stop rebuilding the history from scratch every time

[...]

PR #8697, opened 2026-09-04
test(transaction): pin the interest, sdd and swift calls to POST /api/v1/transactions

JiRaska:
Finishes the `openbank-transaction-service` provider surface for #8345. #8665 covered two of its
five uncovered consumers (lending, standing-order); these are the other three — `interest`, `sdd`
and `swift` — all reaching `POST /api/v1/transactions` through their own `TransactionServiceClient`.

**Composes with #8665, does not touch it.** This branch is cut from `origin/main` and changes **no
provider-side file**, so it does not overlap #8665's edits to
`TransactionPactFolderProviderVerificationTest` / its `@PactBroker` twin. The two PRs are additive
in either merge order: each just adds pacts that `@PactFolder("../pacts")` picks up.

## The no-provider-change claim, verified rather than assumed

#8665's report predicted no provider-side change would be needed. Checked, and it holds:

* all three send `type = "DEBIT"` with a **non-null `sourceAccountId`**, so the reusable
  `a valid source account exists` state already declared on **both** verification classes covers
  them — no new state, so nothing can end up on only one of the pair;
* `TransactionPactFolderProviderVerificationTest` loads `@PactFolder("../pacts")`, so the replay set
  widens by construction. Provider interactions go **6 → 9** with no new class and no new state.

## What each pins that nothing else did

The committed contracts for this route covered `rail=DOMESTIC` with **no** `instructionType`
(domestic-payment), `rail=SEPA_INST` **with** `ONE_OFF` (sepa-instant), `rail=SEPA` on a `TRANSFER`
(sepa-payment), and a `CREDIT` with no rail (account-service). So:

| consumer | shape | previously pinned by |
|---|---|---|
| interest — withholding-tax remittance (#999) | `DEBIT` / `DOMESTIC` / `ONE_OFF` | neither field's **pair**; also the only DTO with no `targetAccountId` *field*, so the key is absent rather than present-and-null |
| sdd — collection debit (#1000) | `DEBIT` / `SEPA_CT` / `DIRECT_DEBIT` | **nothing** — neither value appears in any committed pact |
| swift — MT103 settlement (ADR-0108) | `DEBIT` / `SWIFT` / `ONE_OFF` | **nothing** for `SWIFT` |

`rail` and `instructionType` are orthogonal by construction (ADR-0103), so covering each separately
does not cover a combination.

## Why the response body pins `rail` and `instructionType`, not just `id` and `status`

`TransactionResource.initiateTransaction` parses both with
`runCatching { PaymentRail.valueOf(it) }.getOrNull()` — **an unrecognised value is silently dropped
to null and the route still answers 201**. A pact asserting only the status code would stay green if
`SEPA_CT`, `SWIFT` or `DIRECT_DEBIT` were renamed or removed, while every collection / settlement /
remittance thereafter booked rail-less or instruction-less: a no-op wearing the signal of success.
`TransactionResponse` echoes both fields, so asserting them as **exact values** (`stringValue`, never
`stringType`) is the only thing anywhere that can see it. Verified passing against the real provider,
so the passthrough at `TransactionService:210` is measured, not assumed.

swift additionally gets `uuid("id")` rather than a string matcher: `SettlementAdapter` parses `id`
inside `runCatching { }.getOrNull()`, so a renamed or non-UUID `id` is not an error — it is a settled
MT103 with a null `transactionId`.

## Verified by effect, per pact

Repointed all three client `@Path`s **and** their expected-path literals at
`/api/v1/transaction-bookings`, a route transaction-service does not serve, then regenerated:

| stage | consumer tests | provider replay |
|---|---|---|
| baseline | 3 × `tests=1 skipped=0 failures=0` | **9 / 9 pass** |
| repointed | **all three still GREEN** (the mock server answers whatever the client asks for) | **3 of 9 fail — exactly the three new interactions**, all six pre-existing green |
| restored | 3 × `tests=1 skipped=0 failures=0` | **9 / 9 pass** |

The three failures were, by name: `openbank-swift-service - POST the settled MT103 as a SWIFT one-off
debit`, `openbank-sdd-service - POST an authorised SDD collection as a SEPA_CT direct-debit debit`,
`openbank-interest-service - POST the withholding-tax remittance as a DOMESTIC one-off debit`.

All verdicts read from the **JUnit XML**, never the console — pact-jvm prints
`Not all of the N were verified` on fully green runs. The green consumer row is the point: a consumer
pact alone cannot catch a wrong path (#2269), which is why the expected path is a **LITERAL** and only
the outgoing request is reflected off the client's `@Path` (#2290, #8552, #8665).

## Gates

* `check-pact-provider-replay.py` — **60 committed pacts, 60 replayed on a PR**, `KNOWN_UNCOVERED`
  still **0**, 26/26 providers publishing.
* `derive-pact-drift-scope.sh` — gated modules **32 → 33**; `openbank-sdd-service` is newly in scope
  (it held no pact before), `openbank-interest-service` and `openbank-swift-service` were already
  there and their new tests match the `*.contract.*PactConsumerTest` glob. `ungated pacts:` empty,
  nothing dropped out of scope.
* `ktlintCheck` + `detekt` green on all three modules; full `:openbank-sdd-service:test` is
  **62 tests, 0 skipped, 0 failures** (the new dependency changes nothing else).
* No zero-byte pact (`find pacts -size 0` empty), and `git status` shows exactly the 7 intended
  files — `pact.writer.overwrite=true` rewrote no other contract.

## Wiring

Only `openbank-sdd-service` needed `testImplementation(libs.pact.consumer)`; **interest and swift
already carried it**. That line is placed at the end of the `dependencies {}` block, a different hunk
from #8676's addition to the same file — test-merged to confirm: base 8 entries → 9 each side → **10
merged**, both survive, no silently dropped list entry.

## A comment count I did not touch

`TransactionPactFolderProviderVerificationTest`'s KDoc still hard-counts "Six pacts, five HTTP and one
message", which this PR makes wrong. #8665 already replaces that exact sentence with the mechanism
rather than a new number, on the identical line — duplicating the edit here would conflict with a live
PR for no gain, so it is deliberately left to land with #8665.

Refs #8345

## Security checklist

- [x] No secrets, tokens, passwords, or PII in code, config, tests, logs. — test-only; all ids are
      literal placeholder UUIDs, all amounts and references synthetic.
      this PR). — none added.
- [x] No new third-party dependency, OR: dependency review attached (license, CVE, source). — no new
      dependency: `libs.pact.consumer` is already in the version catalogue and already on 20+ modules'
      test classpaths; sdd-service is the module that had not yet declared it.
- [x] Auth / crypto / payment code changed? → tag `security-review-required`. — considered: **no
      production code changed at all**. The diff is three `src/test` classes, three committed pacts,
      and one test-scope dependency line; `git status` on `src/main` is clean.
- [x] PII path changed? → tag `gdpr-review-required`. — considered: no PII path touched.
- [x] Cardholder data path changed? → tag `pci-review-required`. — considered: no cardholder data
      path touched (`rail=CARD` is not among the shapes pinned here).

---

## Known red gate: `adversarial-contract-test` (3 findings) — NOT worked around

`check-adversarial-contract-tests.py` (ADR-0279 #3) flags all three new files: *"a changed contract
test carries no negative case — add the 401/403/404 expectation"*. That gate landed on `main` at
**20:24 today**, eight minutes before this PR's first run (`d6cb2662ac`, #8622). It is a real
requirement and I have deliberately **not** satisfied it, because the only two ways available to me
are both wrong:

**1. Satisfying it with prose would be decoration.** The gate is a text match — `is_adversarial()`
searches the file for `401|403|UNAUTHORIZED|FORBIDDEN|…` *anywhere*, comments included. Writing "403"
into a KDoc turns it green while pinning nothing. That is exactly the failure mode this PR's own
falsification exists to rule out, and a green gate bought that way is worth less than a red one.

**2. A genuine 403 interaction cannot be replayed today — this is structural, measured, not assumed.**
`TransactionResource.initiateTransaction` is `@RolesAllowed(Roles.OPERATOR)`, so the negative case
would be "a caller without `ROLE_OPERATOR` is refused". But the shared replay class carries
**class-level** `@TestSecurity(user = "pact-verifier", roles = ["ROLE_OPERATOR"])` (line 78), and all
interactions replay through a single `@TestTemplate`. Every interaction in `pacts/` is therefore
served as an OPERATOR: a 403 expectation would replay as **201** and go red. There is no
application-level 401/403/404 on this route to use instead — the SCA gate and `Money` invariant
produce 400 and 422, neither of which the gate accepts, and a 404 would mean inventing a `GET` these
consumers never make.

So the honest negative case needs **per-interaction security on the provider replay class** — a
provider-side change, in the one file #8665 is live on, and a different piece of work from this PR.
The fleet debt is already tracked (**129 of 135** contract-test files are happy-path-only, #8590 #3),
and #8665's two tests carry no negative case either; they simply have not re-run since the gate
landed, so the same three-line finding is waiting for them.

**A blocked merge is the correct final state here.** The consumer/provider evidence above stands on
its own; unblocking this needs a decision about the replay class's security model, not a keyword.


JiRaska:

```
BUILD FAILED in 14m 31s
```

JiRaska:
Merging with known environmental failures: `CodeQL (java-kotlin, manual)` and `dependency-review` fail on the GitHub-side dependency-submission/rate-limit issue (#8690), being fixed separately by #8781. Everything else is green, including the pact regeneration with the raised heap.

Issue #8699, opened 2026-09-04
transaction-service: an unparseable enum becomes null — a typo'd status filter returns EVERY transaction, and a typo'd rail is persisted as null with a 201


JiRaska:
## What

`transaction-service` parses six enum-valued inputs with `runCatching { X.valueOf(it) }.getOrNull()`. An unrecognised value therefore becomes **null**, silently, and null means two different — both wrong — things depending on where it lands.

Found while pinning the `rail` / `instructionType` response fields in a pact (#8697); the pact is currently the only assertion anywhere that can observe the write-path half.

## Severity 1 — a query filter that fails to parse WIDENS the result set

`TransactionResource.kt:114-115` builds the search query:

```kotlin
status = status?.let { runCatching { TransactionStatus.valueOf(it) }.getOrNull() },
type   = type?.let   { runCatching { TransactionType.valueOf(it)   }.getOrNull() },
```

and `PanacheTransactionRepository.kt:176` adds a condition **only when non-null**:

```kotlin
status?.let { result += SearchCondition("status = ?${i++}", listOf(it.name)) }
```

So `GET /api/v1/transactions?status=FAILDE` (a typo) does not 400 and does not return nothing — it returns **every transaction**, unfiltered, with a 200. An operator searching for failed transactions gets the whole set and no indication the filter was discarded.

**The same method treats a malformed date as a client error**, three lines below:

```kotlin
dateFrom = dateFrom?.let { LocalDate.parse(it) },   // throws -> DateTimeExceptionMapper -> 400
```

so within one handler a bad date is rejected and a bad enum silently widens the query. That inconsistency is the clearest evidence this is an oversight rather than a decision.

## Severity 2 — a routing field is silently dropped on a money-path write

`TransactionResource.kt:169-170`, on the initiate path:

```kotlin
rail            = request.rail?.let            { runCatching { PaymentRail.valueOf(it)     }.getOrNull() },
instructionType = request.instructionType?.let { runCatching { InstructionType.valueOf(it) }.getOrNull() },
```

A caller sending `"rail": "SEPA_INSTANT_CT"` when the enum spells it otherwise gets **201 Created** with `rail = null`. The rail is what decides how the payment is routed and which cutoff applies; losing it silently on a 201 is not a degraded read, it is a payment stored without its routing.

`PanacheTransactionRepository.kt:232,234` repeat the pattern on read-back, so a row that somehow holds an unrecognised string also reads as null rather than failing loudly.

[...]

PR #8710, opened 2026-09-04
fix(onboarding): reject an unparseable ?stage= with 400 instead of returning every record

JiRaska:
## Summary

`GET /api/v1/onboarding/records?stage=<typo>` answered **200 with every onboarding record** — `legalName` and `email` included — instead of rejecting the filter. This makes an unparseable stage a 400, matching the closed enum the published spec has always declared. First fix from the fleet-wide sweep of the `runCatching { X.valueOf(it) }.getOrNull()` idiom in #8699.

## Type of change

- [x] fix (bug fix)
- [ ] perf (performance)
- [ ] docs
- [ ] chore (build / tooling)
- [ ] breaking change

## Linked issues

Refs #8699

## The defect

`OnboardingResource.listRecords` parsed the filter with the #8699 idiom, so an unrecognised value became a **legal null**:

```kotlin
val stage = stageParam?.uppercase()?.let { runCatching { FunnelStage.valueOf(it) }.getOrNull() }
```

and `OnboardingProjectionService.listRecords:37` omits the predicate entirely when the stage is null:

```kotlin
val items = if (stage != null) repo.listByStage(stage, page, size) else repo.listAll(page, size)
```

So the filter did not merely fail — it **widened**. Nothing could observe it: the body is well-formed, the status is a success, and `stageFilter` is simply absent from the response rather than contradicting anything. This is the same mechanism as #8699's severity 1 in `transaction-service`, in a second service.

**The spec and the code already disagreed.** `openapi.yaml` declares `stage` as a closed enum of the seven `FunnelStage` values, so a client generated from that document could never send `KYC_OPEEN` — the implementation was accepting values its own published contract forbids. This brings the code into line with the contract rather than changing it.

## Changes

- `OnboardingResource.listRecords` — an unparseable `stage` now throws `IllegalArgumentException`, mapped to **400** by libs-runtime's shared `IllegalArgumentExceptionMapper`. No service-local mapper (#526).
- **Blank stays ABSENT.** `?stage=` has always meant "no filter"; only a genuinely non-blank unparseable value is rejected, so *absent* and *unparseable* remain distinguishable — the property #8699 asks for by name.
- `OnboardingStageFilterIT` — a real HTTP test through the full Quarkus stack.
- `openapi.yaml` `info.version` 1.0.1 → **1.1.0** (additive `400` response ⇒ MINOR, ADR-0048), plus a reusable `BadRequest` response.
- `build.gradle.kts` — `quarkus-test-security`, already used by 10+ services, for `@TestSecurity`.

## Verified by effect

The test was written and run **before** the fix, against unfixed code:

```
tests=3 failures=1 errors=0 skipped=0
  [PASS]    an absent stage still means all stages
  [PASS]    a valid stage is still accepted
  [FAILURE] an unparseable stage is rejected with 400, not answered with every record
            java.lang.AssertionError: Expected status code <400> but was <200>.
```

After the fix:

```
tests=3 failures=0 errors=0 skipped=0
```

Full module suite: `54 tests, 0 failures, 0 errors, 0 skipped` — skipped read explicitly, since a service that cannot boot reports SKIPPED and that scans as a pass. `ktlintCheck detekt` run as a separate invocation afterwards: BUILD SUCCESSFUL.

The two 200 controls are load-bearing: they are what stops the fix from being a blanket rejection, and they are the only thing that can catch a regression where blank or absent starts 400-ing.

## Compatibility note for the reviewer

This is a behaviour change on a read endpoint. A client today sending a typo receives a **wide, unfiltered page and does not know it**; after this it receives a 400. That is the intended correction — the current behaviour hands an operator every record when they asked for one funnel stage — but it is a visible change for any such caller. The endpoint is `ROLE_VIEWER`+ and consumed by the admin-ui onboarding cockpit (ADR-0068), which sends values from the enum.

## Definition of Done

- [x] No new lint warnings.

## Security checklist

## Compliance impact

- [ ] PCI DSS
- [ ] DORA
- [x] GDPR
- [ ] PSD2 / SCA / RTS
- [ ] AML / 5AMLD
- [ ] CNB reporting
- [ ] None

The widened response returned `legalName` and `email` for every party rather than the requested funnel stage — a broader personal-data disclosure than the caller asked for, to a caller already entitled to the data.


JiRaska:
Heads-up: this PR and #8606 both bump `openbank-onboarding-service` `openapi.yaml` `info.version` to **1.1.0**, from the same base of 1.0.1 (`main` is still at 1.0.1). Only one of you can have it — whoever merges second takes **1.2.0**.

Both PRs currently have checks in flight, so neither is clearly ahead. `openapi-version-not-taken` is `mode: enforced` and resolves `origin/main` at run time, so the loser reds on its next run rather than merging a republished version — the detection works, it just costs a cycle.

Worth knowing about the other PR too: #8606 is one half of a two-part removal of `DOCUMENTS_REQUIRED` (the other half is #8544), and I have suggested there that the two be stacked. If that happens, its onboarding spec change may arrive on a different branch than you expect — so re-read `info.version` off live `main` at the moment you next push.

PR #8715, opened 2026-09-04
fix(sanctions): reject an unrecognised listTypes entry instead of silently narrowing the screen

JiRaska:
## Summary

A misspelled entry in `listTypes` was **dropped**, not rejected — so `POST /api/v1/sanctions/screen` could return a normal **201 `CLEAR`** for a screening that never consulted a list the caller asked for. This rejects an unrecognised list name with 400. Second fix from the fleet-wide sweep of the `runCatching { X.valueOf(it) }.getOrNull()` idiom in #8699.

**Money-path service — needs 2 approvals (ADR-0030).**

## Type of change

- [x] fix (bug fix)
- [ ] perf (performance)
- [ ] docs
- [ ] chore (build / tooling)
- [ ] breaking change

## Linked issues

Refs #8699

## The defect

```kotlin
private fun resolveTargetLists(cmd: ScreenEntityCommand): List<SanctionsListType> = cmd.listTypes
    ?.mapNotNull { raw -> runCatching { SanctionsListType.valueOf(raw) }.getOrNull() }
    ?.takeIf { it.isNotEmpty() }
    ?: SanctionsListType.entries
```

`listTypes` decides **which sanctions lists a screening actually consults**. `mapNotNull` turns an unrecognised name into nothing at all, which is two different defects depending on how many names were misspelled:

**1. A partial screen reported as a complete one.** `["OFAC_SDN", "EU_CONSOLIDTED"]` keeps OFAC, drops EU, stays non-empty, and returns a normal 201 — typically `CLEAR` — computed without ever consulting the EU consolidated list. Nothing can observe it: `checkedLists` records what *was* checked rather than what was *asked for*, so the response is internally consistent, and no metric or log distinguishes a partial screen from a complete one. A caller who required EU coverage gets a clean result and a record that quietly agrees with itself.

**2. A single typo widens to everything.** `["EU_CONSOLIDTED"]` maps to an empty list, and `takeIf { it.isNotEmpty() }` falls through to all seven lists — the same "an unparseable filter widens" shape as the query filters elsewhere in this sweep, producing a result whose scope no caller chose.

The KDoc documented the mechanism ("Unknown strings are silently dropped") but not the consequence, which is the part that matters.

## Changes

- `resolveTargetLists` — an unrecognised name now throws `IllegalArgumentException` → **400** via libs-runtime's shared mapper (#526), the same route the `aliases[i]` guard in `screen` already relies on. No service-local mapper.
- **Absent stays distinguishable from unparseable**: null or empty `listTypes` still means every enabled list, exactly as before.
- **Parsing stays case-SENSITIVE**, exactly as before. Accepting `ofac_sdn` where only `OFAC_SDN` worked would *widen* what the endpoint accepts — a separate decision this change deliberately does not make.
- `openapi.yaml` — **`listTypes` was never declared in the spec at all**, so a client generated from the published contract could not send the field that decides screening scope. Documented now with the closed enum, plus the `400`. `info.version` 1.2.0 → **1.3.0** (additive ⇒ MINOR, ADR-0048).
- `build.gradle.kts` — `rest-assured-kotlin`, **test-only**, already used across the fleet. The defect is only observable as an HTTP status, so the test has to drive the real endpoint.

## Verified by effect

Written and run against the **unfixed** code first:

```
tests=4 failures=2 errors=0 skipped=0
  [PASS]    absent list types still means every list
  [PASS]    valid list types are screened and reported as the checked scope
  [FAILURE] an entirely unparseable list type is rejected, not widened to every list
            Expected status code <400> but was <201>.
  [FAILURE] one unparseable list type among valid ones is rejected, not silently dropped
            Expected status code <400> but was <201>.
```

The `201` in that second failure **is** the defect: the service screened, dropped the EU list, and reported success.

After the fix: `tests=4 failures=0 errors=0 skipped=0`.

Full module suite: `98 tests, 0 failures, 0 errors, 1 skipped`. The skip is `SanctionsPactBrokerProviderVerificationTest`, the broker-sourced class gated on `@EnabledIfSystemProperty(pactbroker.url)` — pre-existing and by design (the sanctioned `@PactFolder` + `@PactBroker` pair). Skipped counts were read explicitly rather than inferred from the failure count. `ktlintCheck detekt` run as a separate invocation afterwards: BUILD SUCCESSFUL.

## Compatibility note for the reviewer

This is a behaviour change on a money-path write endpoint.

- A caller sending a **valid** `listTypes`, or none at all, sees no change.
- A caller today sending a misspelled name receives a 400 where it used to receive a 201. That is the point — the 201 it was getting was, in case 1, a screening result narrower than it asked for. But it will surface immediately for any such caller, so it is worth confirming no integration is relying on the old leniency before merge.
- No stored data changes and no migration; `checkedLists` semantics are untouched.

`docs/threat-models/openbank-sanctions-service.md` exists and is unchanged — this narrows an input surface rather than opening one.

## Definition of Done

- [x] No new lint warnings.

## Security checklist

Note on the dependency box: `rest-assured-kotlin` is added as a **test-only** dependency. It is already a fleet-standard test dependency (25+ modules), Apache-2.0, and pinned through the shared version catalog — no new coordinate enters the runtime image.

## Compliance impact

- [ ] PCI DSS
- [ ] DORA
- [ ] GDPR
- [ ] PSD2 / SCA / RTS
- [x] AML / 5AMLD
- [ ] CNB reporting
- [ ] None

A sanctions screening that silently omits a requested list, and records `CLEAR`, is a screening-coverage gap of exactly the kind 5AMLD controls exist to prevent. The stored `SanctionsCheck` rows remain accurate about what was checked; what was missing was any rejection of a request whose asked-for scope could not be honoured.


JiRaska:
Force-pushed `fa7e1c5` → `fefb140` to rebuild this branch cleanly from `origin/main`.

The branch had been cut from `fix/onboarding-stage-filter-8699` (#8710) rather than from `main`, so it carried that PR's commit as well — the chain-branch pollution trap in `CLAUDE.md`. The first CI run gave it away by building `openbank-onboarding-service` on a sanctions-only PR.

Same single commit, same content, re-signed; the diff against `main` is now the four `openbank-sanctions-service` files only. #8710 is unaffected and remains independent.

JiRaska:
Heads-up on a version race you will probably lose, so it is cheaper to know now than to be surprised by a red gate later.

This PR bumps `openbank-sanctions-service` `openapi.yaml` `info.version` to **1.3.0**. So does #6015, from the same base of 1.2.0 (`main` is still at 1.2.0).

#6015 is **green, `MERGEABLE`, and every check has completed**; this PR still has one failure and three checks running. If #6015 lands first, `1.3.0` is taken and `openapi-version-not-taken` — `mode: enforced`, and it resolves `origin/main` at run time — reds this PR on its next run. The number to take then is **1.4.0**.

Nothing to do until that happens; re-read the version off live `main` when you next push rather than trusting this comment, since #6015 may not be the only thing that moves.

PR #8752, opened 2026-09-05
fix(swift): publish operator acknowledge and reject through the outbox

### LINKED ISSUE #8718: swift-service: acknowledge and reject write no outbox row, so no consumer or audit trail ever learns an operator acted
## What

`openbank-swift-service` announces every **scheme** transition and no **operator** one. Acknowledgement and rejection change a SWIFT message's status and write no outbox row, so nothing downstream — including the audit trail — ever learns they happened.

Found while adding outbox-atomicity coverage (#8353 / PR #8717). It is **not** an atomicity defect: there is no second row to diverge from the first. That is exactly what makes it invisible to the gate and to every test in that family.

## Measured on `origin/main`

`saveWithOutbox` has exactly **two** call sites in `openbank-swift-service/src/main`:

```
SwiftService.kt:118   submitToScheme  → repo.saveWithOutbox(updated, outbox)
SwiftService.kt:155   settle          → repo.saveWithOutbox(completed, outbox)
```

The two operator-driven transitions use the plain save:

```
SwiftService.kt:175   acknowledge(id, ackRef)  → repo.save(...)     POST /api/v1/swift/{id}/ack
SwiftService.kt:186   reject(id, reason)       → repo.save(...)     POST /api/v1/swift/{id}/reject
```

And `SwiftOutboxRepository.persistInTransaction` has **zero** callers — the only two references are the port declaration and its implementation.

## Consequence

`openbank.payments.swift.event` carries the scheme verdicts and the settlements. It does not carry an acknowledgement or an operator rejection. `audit-service` subscribes to that topic, so the audit trail holds a record of what the scheme decided and no record of what an operator decided.

For a money-path service that asymmetry is the wrong way round: a scheme verdict is reproducible from the counterparty's own records, while an operator's rejection exists only here.

## What this issue is asking

Whether the omission is intended. There is a defensible reading — ack and reject may be considered internal status bookkeeping rather than events the outside needs — and if that is the answer, it should be written down next to those methods, because the file gives a reader no way to distinguish a decision from an oversight: two transitions publish, two do not, and nothing says why.

If it is not intended, the fix is the shape already used twice in the same class: `repo.saveWithOutbox(updated, outbox)` with the appropriate event type, which also makes both transitions atomic by construction and brings them under the coverage #8717 just added for the other two.

## Not in scope here

Whether other services have the same shape — a status transition that writes no event while its siblings do — is not established. It is worth asking, but the answer needs reading each transition's intent, not a grep: `saveWithOutbox` having fewer call sites than a service has transitions is normal wherever the untouched transitions genuinely have no external meaning.

`openbank-swift-service` is money-path. Refs #8353, #8717

JiRaska:
## What

`openbank-swift-service` announced every **scheme** transition and no **operator** one.

| transition | before | now |
|---|---|---|
| `submitToScheme` (scheme verdict) | `saveWithOutbox` | unchanged |
| `settle` (ADR-0108 booking) | `saveWithOutbox` | unchanged |
| `acknowledge` — `POST /api/v1/swift/{id}/ack` | `save`, **no event** | `saveWithOutbox` |
| `reject` — `POST /api/v1/swift/{id}/reject` | `save`, **no event** | `saveWithOutbox` |

`audit-service` subscribes to `openbank.payments.swift.event`
(`openbank-audit-service/src/main/resources/application.yaml`), so the audit trail held a record of
what the scheme decided and no record of what an operator decided. For a money-path service that
asymmetry is the wrong way round: a scheme verdict is reproducible from the counterparty's own
records, while an operator's rejection exists only here.

Answering the question #8718 actually asked — is the omission intended? — no. Nothing in the file,
the docs or the AsyncAPI document treats ack/reject as internal bookkeeping; `04-data.en.md`
described the ack/reject outbox wiring as merely *not written yet*, and the AsyncAPI payload schema
already lists `ACKNOWLEDGED` in its `status` enum. It was an oversight.

## The contract-safe option, and why it is the one taken

This adds **occurrences of an already-published event, not fields**.

* `docs/asyncapi/openbank-events.yaml`'s `SwiftMessageEventPayload` already declares
  `status: [PENDING, VALIDATED, SENT, ACKNOWLEDGED, REJECTED, FAILED]` — both new values are in it,
  so no consumer sees a value the contract did not already allow. `REJECTED` was already on the
  wire from the scheme-reject path.
* The payload keys are unchanged: the seven the document declares plus `sourceService`. `ackRef`
  and `rejectionReason` are deliberately **not** added — that same document records that an earlier
  revision declared them and *"none of which the producer has ever emitted"*, and re-adding them
  would be the widening this deliberately avoids.
* `.github/event-contract-baseline.txt` needs no change: the topic is grandfathered either way and
  no new topic appears. Verified locally — `check-event-contract-coverage.py`,
  `check-event-contract-code-agreement.py`, `check-asyncapi-doc-discriminator.py`,
  `check-outbox-has-writer.py`, `check-synthetic-outbox-taint.py`,
  `check-outbox-dispatch-enabled.sh` and `check-audit-money-path-subscription.py` all exit 0.
* `SwiftEventPactConsumerTest`'s pact (consumer: transaction-service) pins its example to
  `COMPLETED` and matches `status` against `VALIDATED|SENT|COMPLETED|REJECTED`; the pact is
  unaffected because it asserts over its own fixed example, not over the producer's value set.

The envelope is now built once (`statusChangedEvent`) and all four call sites share it, so a fifth
transition cannot invent a divergent shape. That refactor is the whole reason the two existing call
sites appear in the diff.

## Proving it, rather than asserting it

`SwiftOperatorTransitionOutboxIT` (new) drives the real endpoints over HTTP and reads Postgres's own
`xmin` — the id of the transaction that wrote each row version.

Two things it does that a presence assertion cannot:

1. **Presence is asserted first, and separately.** The absent event is *not* an atomicity defect —
   with no second row there is nothing to diverge from the first, so every "the two rows share a
   transaction" assertion is *vacuously* satisfiable by a method that writes no row at all. That is
   exactly why the outbox-atomicity family (#8353) was structurally blind to this bug.
2. **A known-different control runs in the same green test.** After `acknowledge` then `reject`, the
   REJECTED outbox row must share the message row's current `xmin` and the ACKNOWLEDGED one must
   **not** — so one green run shows the identical comparison both matching and not matching. An
   equality assertion that has never seen an unequal pair is not known to be able to fail.

The scheduled dispatcher is switched off per-class in a `QuarkusTestProfile` (not a
`@QuarkusTestResource`, which would apply module-wide): its `UPDATE` writes a new row version under
a new `xmin` and would destroy the evidence. The scheme-submission pilot flag is left at its default
OFF, which is what makes the row counts exact — `POST /api/v1/swift` then writes no outbox row, so
every row the test finds was written by the operator transition under test.

## Falsification

| run | result |
|---|---|
| fix in place | **20/20 green**, 0 skipped (`SwiftServiceTest` 17, `SwiftOperatorTransitionOutboxIT` 3) |
| both methods reverted to `repo.save` | **16/20 — 4 red**: 2 IT + 2 unit |

The IT failures name the right thing, not a side effect:

```
[POST /api/v1/swift/<id>/ack must write exactly one swift_outbox row — before #8718 it wrote none…]
Expected size: 1 but was: 0 in: []
```

Counts read from the JUnit XML, including `skipped="0"` — a service that cannot boot reports its
tests as SKIPPED, and a skip count scans as a pass.

## One thing worth flagging

A second sabotage — splitting `SwiftRepositoryImpl.saveWithOutbox` into two sequential top-level
`Panache.withTransaction` blocks, to check the `xmin` oracle itself — does **not** produce a clean
two-transaction implementation here. It exhausts the reactive SQL client pool and the request hangs
(`io.vertx.core.impl.NoStackTraceThrowable: Timeout`, `SocketTimeoutException` at the RestAssured
call), so the test goes red by timeout rather than by an `xmin` mismatch. It was reverted and is not
in this diff. Worth knowing for anyone reaching for that sabotage as a falsification vehicle — the
permanent known-different control described above is the load-bearing evidence that the comparison
discriminates, and it passes on real data every run.

## Not in scope

Whether other services carry the same shape — a status transition that publishes nothing while its
siblings do — remains open, as #8718 says; the answer needs each transition's intent read, not a
`saveWithOutbox` call-site count.

Closes #8718
Refs #8353, #8717

## Security checklist

      (license, CVE, source).

Each box verified against the diff rather than assumed:

- `git diff <merge-base> HEAD | grep -E '^\+.*(@Suppress|as Any|SuppressWarnings)'` — empty.
- No `build.gradle*`, `gradle.properties`, `libs.versions.toml` or `package.json` in the diff — no new dependency.
- No secret-shaped additions (`grep -iE '^\+.*(password|secret|token|api[_-]?key)\s*='` — empty). The new IT reads `xmin` over a plain JDBC connection using the Quarkus datasource config; it introduces no credential of its own.
- **Payment code did change** — `openbank-swift-service` is in `rules.yaml: money_path_services`, and this alters two state transitions on SWIFT messages — so `security-review-required` is applied. What a reviewer should weigh: the change publishes `SWIFT_MESSAGE_STATUS_CHANGED` for `ACKNOWLEDGED` and `REJECTED`, which puts an operator's decision (including `rejectionReason`, which is **not** carried on the wire — see the contract note above) onto `openbank.payments.swift.event`, a topic audit-service consumes.
- No PII path and no cardholder-data path is touched; both boxes are ticked as considered-and-not-applicable, and the corresponding labels are deliberately **not** applied.


JiRaska:
### Local verification — what ran, and what did not

Ran and green:

- `:openbank-swift-service:ktlintCheck` and `:openbank-swift-service:detekt` — both tasks appear in the output, `BUILD SUCCESSFUL`.
- `:openbank-swift-service:test --tests '*SwiftOperatorTransitionOutboxIT*' --tests '*SwiftServiceTest*'` — **20 tests, 0 skipped, 0 failures** (counts read from the JUnit XML, not the console; a service that cannot boot reports SKIPPED, and a skip count scans as a pass).
- Governance gates touched by an event-shape change, all exit 0: `check-event-contract-coverage.py`, `check-event-contract-code-agreement.py`, `check-asyncapi-doc-discriminator.py`, `check-outbox-has-writer.py`, `check-synthetic-outbox-taint.py`, `check-outbox-dispatch-enabled.sh`, `check-audit-money-path-subscription.py`.

Did **not** complete locally: the full `:openbank-swift-service:test` task. The local Docker daemon
wedged part-way through (`docker ps` stops responding), so the Testcontainers-backed classes could
not all boot — an environment failure on this machine, unrelated to this diff, and it also had the
first attempt hung on an orphaned test JVM. CI is the authority for the whole module here.

JiRaska:
Reviewed this against its sibling #8717 (same package, same oracle) and verified the fix independently. Two notes for whoever reads them together.

**No collision.** `comm -12` over the two diffs is empty — #8717 adds `SwiftOutboxAtomicityIT.kt` covering the scheme paths (`submitToScheme`, `settle`), this one adds `SwiftOperatorTransitionOutboxIT.kt` covering the operator paths. Different files, different classes, complementary.

**The split-transaction sabotage behaves differently on the two paths, and neither KDoc is wrong.** While building this PR the sabotage — splitting `SwiftRepositoryImpl.saveWithOutbox` into two sequential `Panache.withTransaction` blocks — did not produce a clean two-transaction run on the *operator* paths: it exhausted the reactive SQL client pool and the request timed out (`NoStackTraceThrowable: Timeout`), so the test went red for the wrong reason.

That is a real observation, and it does **not** contradict #8717, whose KDoc says the same edit "turns this test red naming both ids". That PR's body records the measured values from actually doing it on the scheme paths:

```
swift  submitToScheme -> saveWithOutbox   message xmin=750, outbox xmin=751
swift  settle         -> saveWithOutbox   message xmin=755, outbox xmin=756
```

The difference is what else is holding a connection: `acknowledge` and `reject` run `repo.findById(id)` before the write, so the same textual edit yields three sequential transactions on that path rather than two. **Worth carrying as a general point — a sabotage is only established for the call path you ran it on**, and "the same edit, the other endpoint" is not the same experiment.

Which is why the falsifiability of *this* test does not rest on that sabotage at all: the reject case carries a permanent known-different control (the earlier `ACKNOWLEDGED` outbox row must **not** share the current message row's `xmin`), so the oracle proves on every run that it can tell same-transaction from different-transaction, rather than only on the day someone breaks the code on purpose. That is the stronger construction of the two.

**Verified independently on this branch:** `acknowledge` (`SwiftService.kt:192`) and `reject` (`:203`) both go through `saveWithOutbox` with the shared `statusChangedEvent(...)` builder; the two pre-existing call sites (`:133`, `:154`) are refactored onto it, which is why they show in the diff. `create` at `:105` correctly still uses plain `repo.save` — a message that has just been created has not changed status, so there is nothing to announce.

One caveat the author already flagged and I can confirm: the local Docker daemon on this machine is wedged (`docker ps` hangs past 25s), so the full `:openbank-swift-service:test` task did not complete locally and CI is the authority on module-wide green. Two orphaned testcontainers are still up and will need clearing once the daemon recovers.

PR #8762, opened 2026-09-05
test(party): main does not compile — the third exportPartyGdpr call site is missing its argument

JiRaska:
## `main` does not compile

`:openbank-party-service:compileTestKotlin` fails on current `origin/main`:

```
```

#8487 (`2ba4787544`, merged 2026-09-05T01:34Z) gave `PartyResource.exportPartyGdpr` a second parameter —

```kotlin
suspend fun exportPartyGdpr(
    @PathParam("id") id: UUID,
    @HeaderParam(CUSTOMER_PARTY_HEADER) partyHeader: String?,   // no default
): Response
```

— and updated **two of the three** call sites in the same commit. Lines 108 and 157 pass `(partyId, null)`; line 141 still passes `(partyId)`.

## Why `null`, specifically

The parameter is read in exactly one place:

```kotlin
val viaEdge = !byStaffOrSubject && edgeActsForSubject(partyHeader, id)
```

The test at line 141 asserts the **DPO** path, so `isDpo` is true, `byStaffOrSubject` is true, and `viaEdge` is short-circuited to `false` before `partyHeader` is ever read. `null` is both what the sibling call sites pass and behaviourally the right value here — the assertion is unaffected either way, which the passing test confirms.

## Why nothing caught it

CI is path-scoped. #8487 landed as a customer-edge + party change and its own run predates the merge; nothing since has rebuilt this module's tests. `main`'s check-runs read **green**, because the job that would fail was never scheduled.

It surfaces only on unrelated PRs whose diff happens to pull a party-service build into scope — **#8665, #8676, #8697** are all red on `Submit fleet dependency graph` for this reason, where it reads as *their* failure. I went looking for a rate-limit cause there (there was a real installation-quota incident tonight, #8690) and found a compile error instead.

This is the failure mode the repo already documents: *"a module CI never rebuilt stays red on `main` unseen"* — the #5844 shape, one step milder because it breaks at compile time rather than at boot.

| tree | `:openbank-party-service:compileTestKotlin` |
|---|---|
| fix in place | **exit 0** |
| reverted | **exit 1**, naming `PartyResourceAuditTest.kt:141:44` |
| restored | **exit 0** |

And the class actually runs, read from the JUnit XML rather than the console:

```
PartyResourceAuditTest — tests=6 failures=0 errors=0 skipped=0
```

The skipped count matters here: a `@QuarkusTest` class that cannot boot reports its tests as SKIPPED, and a skip count scans as a pass.

## Scope

One line, in `src/test`. `openbank-party-service` is not in `rules.yaml: money_path_services`. `src/test` is in the package's `exclude-paths`, and `test` is a hidden commit type, so this releases nothing on either axis.

## Security checklist

      (license, CVE, source).

The diff is one argument added to one existing test call — no production code, no dependency, no `@Suppress`, no secret-shaped line. It touches a test *of* a GDPR export path but changes no PII handling: the argument it supplies is the one the compiler already required and the two sibling call sites already pass.

### DIFF
--- a/openbank-party-service/src/test/kotlin/com/openbank/party/infrastructure/rest/PartyResourceAuditTest.kt
+++ b/openbank-party-service/src/test/kotlin/com/openbank/party/infrastructure/rest/PartyResourceAuditTest.kt
@@ -138,7 +138,7 @@ class PartyResourceAuditTest {
         coEvery { res.partyUseCase.exportPartyData(partyId) } returns
             PartyGdprExport(sampleParty(), emptyList(), now)
 
-        val response = res.exportPartyGdpr(partyId)
+        val response = res.exportPartyGdpr(partyId, null)
 
         assertThat(response.status).isEqualTo(200)
         assertThat(events).singleElement().satisfies({ e ->


JiRaska:
**Blast radius, measured after opening this PR. One missing argument is producing four separate red signals, none of which names party-service.**

| what is red | how it presents | verified |
|---|---|---|
| `main`, `Dependency submission` — #8740 | a dependency-graph / API fault | run `33944295615` log: `PartyResourceAuditTest.kt:141:44` |
| verification-metadata sweep — #8769 | heap exhaustion, or metadata drift | job `101248717683`: same line, `BUILD FAILED in 1m 58s`, no `OutOfMemoryError` anywhere |
| #8665, #8676, #8697 | each PR's own `Submit fleet dependency graph` failure | run log per PR: same line; none touches party-service |
| #8732 | merged **with** `build (openbank-party-service)` red | that check is evidently not required, so the break did not block a merge |

`main` has failed `Dependency submission` on every commit since #8487: `3efb275d7` (03:53), `994eddd43` (03:57), `560a07b63` (04:10), `d604ab0f1` (04:20). The 06:10 run on `67af1539e2` was still in flight when I checked and fails identically — nothing has changed the file.

**Two things generalise beyond this fix.**

*The symptom never names the cause.* Each of those four arrives dressed as something else — an API quota problem, a heap problem, three unrelated PRs' own failures. I chased the rate-limit explanation first, because tonight had a genuine installation-quota incident (#8690) and `Submit fleet dependency graph` sounds like exactly that. The compile error was ~40 seconds further up the same log. **The last `##[error]` is the reporting step; the first `e:` / `FAILED` task is the cause.**

*`main` reading green is not `main` being green.* Ordinary check-runs on `main` show no failure because path-scoped CI never schedules the job that would fail. The fleet-wide workflows are the only things on `main` that compile everything — which is why the break is visible **only** through them, and why it looks like their fault when it appears. A second-order version of the same trap caught me in this very session: a `check-runs` query filtered to `conclusion=="failure"` returned empty while 14 checks were still `in_progress`, and I read that as green. Empty-failures plus unfinished is *unknown*, not *pass*.

This PR is green: 31 checks, 23 success, 7 skipped, 1 neutral (CodeQL, which is neutral on this repo's config), 0 failures.

JiRaska:
Duplicate — the same one-line fix already landed on main via #8778. Closing.

PR #8768, opened 2026-09-05
fix(party): pass partyHeader to exportPartyGdpr in DPO test

JiRaska:
## Summary
`PartyResource.exportPartyGdpr(id, partyHeader)` has no default value for `partyHeader`.
The DPO-caller test in `PartyResourceAuditTest` calls it with a single argument, so
`compileTestKotlin` for `openbank-party-service` has been failing to compile on `main`,
blocking every PR that touches the module — including unrelated ones (e.g. #8749).

The sibling test 20 lines above already calls it correctly as `exportPartyGdpr(partyId, null)`.

## Test plan
- [x] `./gradlew :openbank-party-service:compileTestKotlin` — BUILD SUCCESSFUL


JiRaska:
Duplicate of #8778, which already landed on main with the identical one-line fix. Closing.

Issue #8797, opened 2026-09-05
The six 2026-09-30 governance deadlines, measured: four cannot graduate, two are unknown


JiRaska:
Six advisory rules carry `target_enforce_date: 2026-09-30`. I measured each one; the picture is not in any single place today, so this is it.

**Four cannot graduate on that date. Two are unknown for reasons worth naming.**

| rule | verdict | evidence |
|---|---|---|
| `coverage.money_path_depth` | **blocked** | 5 services below the 70% bar |
| `change_requirements.mp_messaging_dotted_keys` | **retire, don't graduate** | 171 occurrences vs the enforced sibling's baselined 11 |
| `dependencies.artifact_integrity` | **unknown** | the sweep that would answer it could not run |
| `change_requirements.openapi_request_schema_conformance` | **unknown** | not measurable outside a service build |
| `finops_tiers` | **safe but near-empty** | would enforce over 3 of 68 services |
| `provenance` | **1 of 4 criteria confirmed** | the hard precondition passes in CI |

---

## money_path_depth — blocked by five services

Re-measured against the last weekly pitest run (33289385445, 2026-08-30). Corrected in #8312: the record named account and transaction, both of which have since **cleared**, and omitted three that are below.

```
below 70%:  balance 60  interest 60  ledger 63  sanctions 69  sepa-payment 69
at/above :  account 70  fx 72  domestic 74  sdd 75  transaction 76  sepa-instant 77  fraud 89
```

## mp_messaging_dotted_keys — the action is to retire the duplicate

Re-measured 2026-09-05 (#8780). Both issues the old note sequenced on (#3928, #3931) are **closed**, and the count is 171, not the 16 recorded.

171 is not "the debt grew 10×" — the two gates count different things:

- `kafka-dotted-key-ratchet` (**enforced**) baselines today's set as an operational decision (#2945) and fails only on a **new** occurrence. Clean: 11 baselined, 0 new.
- `dotted-mp-messaging-key-guard` (**advisory**, this rule's producer) has **no baseline at all**.

Flipping it would fail the build on 171 occurrences the enforced gate already decided to keep.

## artifact_integrity — the measurement itself failed

Metadata look substantial: **3293 components, 5458 sha256 entries**, `verify-metadata=true`. Graduating means flipping Gradle `lenient` → `strict`.

[...]

PR #8799, opened 2026-09-05
feat(gates): register finops-tier-drift so a collapse of its subject set is a hard failure

JiRaska:
## The gap

`rules.yaml` names `gate: finops-tier-drift` — **but no such gate existed in the manifest.** The checker ran only from `finops-lifecycle.yml`, where, by design (honouring `finops_tiers.enforced: advisory`), it returns 0 for every finding.

Measured 2026-09-05:

```
tier coverage: 26/68 services classified
   3 explicitly declared   (notification T2, card-issuance T1, product-catalog T0)
  23 money-path -> T0 baseline
  42 unclassified
```

**The script exits 0 whether that number is 3 or 0.** Nothing anywhere would have noticed the difference.

## What `mode: enforced` means here — not what it looks like

Tier **correctness stays advisory**: the script still returns 0 for findings, so this gate cannot fail on a wrong tier. The only red path is run-gates' `min_subjects` floor.

The split is deliberate and both halves are written into the `run:` block: **we do not yet enforce that tiers are right, we do enforce that the checker can still see them.**

## What that required

| | |
|---|---|
| `count_subjects()` extracted | so the self-test drives the **shipped** counter, not a copy |
| a `SUBJECTS=` line | the floor is the load-bearing output, and the floor can only read that line |
| `--self-test` | it falsifies the **count**, not the findings — because the count is the only thing that can go red |

The self-test uses four fixtures spanning 0..3 including the collapse case, **plus an assertion that the fixtures still disagree with each other**: a counter returning a constant would otherwise pass all of them.

## Known-positive, through the runner, both directions

Emptying `money_path_services`:

```
SUBJECTS=3
[run-gates] this gate examined 3 subject(s), below its declared floor of 15.
1 gates  FAIL=1
```

Restored: `PASS=1`.

## Notes

Floor at **15** against 26 observed — a collapse detector, not a target, in the same spirit as the rest of the manifest. `--report` still works; the weekly audit reads it.

`python-lint`, `gate-subject-floor`, `gate-lifecycle-metadata`, `gate-selftest-declaration`, `gate-observability-declarations`: all pass.

## Linked issues

Refs #8797 — where this gap was recorded while measuring the 2026-09-30 deadlines.


JiRaska:
Second-agent review of this PR found two defects in it; both are fixed in the commit above.

**1 — the floor was blind to the collapse the rationale is written about.** `SUBJECTS` was the 26-service union, but 23 of those come from `money_path_services`, which `journey-money-path-accountability` already floors at 20. So `declared: {}` (3 → 0) still left `SUBJECTS=23`, over the floor of 15, and the gate stayed **green** on the exact case cited. It now floors the declared side alone, `min_subjects: 2` (3 observed; 2 lets one service be demoted, a wipe cannot pass). Falsified both directions:

```
declared wiped  -> SUBJECTS=0  FAIL=1
restored        -> SUBJECTS=3  PASS=1
```

**2 — the self-test proved the wrong half.** `count_subjects()` is a one-line union that cannot plausibly drift; the collapse mechanism is the line-scanning parsers upstream, which no fixture touched. Mutating the money-path regex `[A-Za-z0-9_-]+` → `[A-Za-z0-9]+` collapsed `SUBJECTS` to 3 while the self-test stayed green. A parser fixture with a known answer now goes red on that mutation:

```
FAIL parser: money_path_services from a fixture: want [...], got []
```

Also added the `# N observed (date)` inline comment the file's convention expects.

Not changed, and deliberately: `mode: enforced` is about the floor, not findings — the `run:` block says so, and `check-advisory-gate-registration.py` correctly does not count this as an advisory gate, so `rules.yaml`'s `target_enforce_date: 2026-09-30` accounting is untouched.

JiRaska:
## The `Validate manifests` failure is `advisory-gate-registration`, and it is real

`Validate manifests` is only the aggregator (`One or more gate shards failed`). The gate that
fails is `advisory-gate-registration`, reproduced locally against this branch:

```
finops-tier-drift — the gate is `mode: enforced` and BLOCKS today, but rules.yaml `finops_tiers`
still declares `enforced: advisory` with a `target_enforce_date`.
```

So this PR registers the gate as enforced while `rules.yaml` still calls the same control
advisory. That is a genuine disagreement, not gate noise.

## I tried the obvious fix and it does not hold — the three artefacts use three vocabularies

Measured, all combinations:

- `enforced: enforce` → `advisory-gate-registration` moves on, but `check-finops-tiers.py`
  reports `finops_tiers.enforced is 'enforce' — expected advisory|block`.
- `enforced: block` → the tier script accepts it and **still** prints "not blocking until the
  classifier flips the gate to block" and returns 0, while `advisory-gate-registration` now
  objects that the workflow STEP is advisory.
- The step itself says so in its own comment: "the script always exits 0 for now".

And `check-finops-tiers.py` ends in an unconditional `return 0`. It cannot fail on a finding at
all, which is why nobody has seen the one it is currently reporting:

```
⚠️ openbank-card-issuance-service: money-path service declared 'T1' (below T0).
   Demoting a money-path service requires an ADR-0030 threat model + 2 approvals.
```

## What I did not do, and why

I reverted my `rules.yaml` edit rather than pushing it. Making this consistent means deciding
whether card-issuance at `T1` is correct — a money-path tier question needing an ADR-0030 threat
model and two approvals — and that is not a gate-registration change. Forcing it through here
would either redden every PR on that finding or paper over it.

Filed as #9678 with the full measurement. This PR's own contribution — the `min_subjects` floor,
so a collapsed declared set cannot read as a pass — is right and is untouched by any of it.

JiRaska:
## Merged `main` in, integrated both sides — and the remaining blocker is a governance decision, not a merge

### What the merge did

`main` and this branch both changed `check-finops-tiers.py`, for different and non-overlapping
reasons, so this integrates them rather than taking a side (taking either would have silently
reverted the other):

- `main` (#9678, PR #9733) split `evaluate()` into `evaluate_split()`, so a CONFIG bug — an
  unparseable `rules.yaml`, an undefined tier, a declared key that is not a service — blocks under
  every policy setting, while the money-path POLICY finding stays advisory. Its self-test drives
  seven cases through the real parser.
- This branch adds the other output: `count_subjects()`, the `SUBJECTS=` line that
  `gates.yaml: min_subjects` floors, and the fixtures that make that count falsifiable.

They fail differently, which is why both are needed: every case in main's self-test falsifies the
FINDINGS, and a check reasoning about zero services produces no findings at all — that reads as
health.

Integration: `evaluate_split()` returns a fourth element (the declared-subject count), `evaluate()`
and `self_test()` unpack it away, `main()` prints it; the coverage line now calls `count_subjects()`
instead of its own inline union, so the self-test drives the shipped counter; main's `self_test()`
gains the four count fixtures (spanning 0..3, so a constant cannot pass) plus the
`parse_money_path_services` fixture the count depends on.

`--self-test` is **12 of 12 green**, and the floor was falsified **by effect**: with `declared:`
wiped to `{}` the script prints `SUBJECTS=0`, below `min_subjects: 2`, so the gate goes red on
exactly the collapse it was registered for.

Also corrected the gate's `rationale` in `gates.yaml`, which still said the script "returns 0 for
every finding by design" — true when this branch was written, false since #9678.

### The remaining blocker

`python3 .github/scripts/run-gates.py --all` on the merged branch, differenced against the same
command on bare `origin/main`, leaves exactly one new finding:

```
::error title=Advisory gate registration::.github/gates/gates.yaml :: registry :: finops-tier-drift
— the gate is `mode: enforced` and BLOCKS today, but rules.yaml `finops_tiers` still declares
`enforced: advisory` with a `target_enforce_date`.
```

**Its stated remedy is wrong for this case.** The gate says "set `enforced: enforce` +
`enforced_since:` and drop the date", and doing that would make the money-path POLICY finding
blocking — and there is a live one: `openbank-card-issuance-service` is declared `T1`, below `T0`.
That would red-gate `main`, and demoting a money-path service is an ADR-0030 threat model plus two
approvals, not a side effect of landing a subject floor.

Two knobs are being conflated:

| knob | governs |
|---|---|
| `gates.yaml: finops-tier-drift.mode` | whether the **subject floor** blocks |
| `rules.yaml: finops_tiers.enforced` | whether the **tier policy finding** blocks |

This PR's whole thesis is that the first should be enforced while the second stays advisory.

Turning the gate to `mode: advisory` does not work either — measured in `run-gates.py`: a
`min_subjects` breach sets `rc = 1`, but `mode: advisory` maps a non-zero rc to `warned`, not
`failed`. So the floor would stop blocking and the PR would buy nothing.

The checker cannot currently express the split: `rules_by_producer()` builds
`out[script] = (rule, enforced, has_date)`, one entry per producer **script path**, so a second
`rules.yaml` rule naming the same `ci_producer` would silently overwrite the first rather than
describe a second lifecycle.

### What I did not do

I did not flip `finops_tiers.enforced`, and I did not weaken the gate to make the check pass. Both
would trade a correct refusal for a green tick. The choice between

1. giving the floor its own `rules.yaml` rule and teaching `check-advisory-gate-registration.py`
   that one producer can serve two lifecycles, or
2. resolving `openbank-card-issuance-service: T1` first (threat model + approvals) and then
   graduating `finops_tiers` properly,

is a governance call, not a merge conflict.

PR #8801, opened 2026-09-05
fix(governance): drain paid-off psd2 idempotency baseline entries

JiRaska:
## Summary
`gates (gitops-api)` fails fleet-wide on every PR right now: `idempotency-contract-parity`
reports 4 STALE baseline entries for `openbank-psd2-service` (2 name-mismatch, 2
bound-not-published). Re-running the checker at current `main` HEAD confirms all four
are no longer real findings — the psd2 contract/binding drift was already fixed
elsewhere; only the declared-backlog file was never drained.

Deleted the four stale lines and trimmed the prose block that described them, keeping
the still-live sca-service entry.

## Test plan
- [x] `python3 .github/scripts/check-idempotency-contract-parity.py --enforce --baseline .github/gates/idempotency-parity-baseline.txt` — `0 NEW, 0 stale`
- [x] same script `--self-test` — still passes (24 cases)

### DIFF
--- a/.github/gates/idempotency-parity-baseline.txt
+++ b/.github/gates/idempotency-parity-baseline.txt
@@ -8,24 +8,9 @@
 # Every entry below is one service's PIS/SCA surface and each is a real, live defect, not a
 # tidiness item. They are declared rather than fixed in the gate's own PR because correcting a
 # published header name is an api-contract change on a money-path service (two approvals plus a
-# threat model, rules.yaml money_path_services) and because the psd2 correction has to choose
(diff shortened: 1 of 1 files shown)


JiRaska:
Superseded — #8783 already drained these same 4 stale psd2 entries on `main` (data lines only; my PR additionally trimmed the now-stale prose comment, which is what conflicts). Re-verified against `origin/main`: `check-idempotency-contract-parity.py --enforce` reports `0 NEW, 0 stale`. Closing as duplicate.

PR #8802, opened 2026-09-05
fix(ci): the openapi request-schema gate has compared nothing for six weeks — wrong path, now fixed

JiRaska:
## The finding

The rule carried **no `blocked_on`**, so its 2026-09-30 date implied readiness was merely unmeasured. It is worse than unmeasured: **the gate examines nothing**, and its own *"Not a finding"* wording is what hides that.

It compares the committed `openapi.yaml` against the schema Quarkus emits during the build (`-Dquarkus.smallrye-openapi.store-schema-directory`, `_service-ci.yml`). **That generated document is not being produced.**

Across three independent CI runs, every service that *reaches the script* — a service with no committed `openapi.yaml` is skipped one step earlier by the workflow's own guard, as `openbank-simulation` was in `33949966927`:

| run | services | result |
|---|---|---|
| `33949966927` | billing, interest, swift | `no generated schema ... Not a finding` |
| `33949973656` | document-service | same |
| `33949914172` | customer-edge | same |

**On this path the count line is never printed at all** — the script `return 0`s before reaching its `"{checked} request schema(s) compared, {findings} finding(s)."` line. Zero request schemas have been compared.

(Note for anyone verifying: the string is **compared**, not *checked*. Grepping the wrong word returns empty across every run, which reads as confirmation while measuring nothing — the reviewer hit exactly that.)

## Why this is not "unmeasured readiness"

Flipping it to enforced would graduate a check that examines nothing — and its green would then be **load-bearing**.

A missing generated document currently reads **identically to a clean service**. That is the same *"could not look"* encoded as *"looked and found nothing"* that `check-fleet-attestations.sh` deliberately separates (it exits 2 and prints *"NOT a verdict"*). This script does not.

## What the date actually needs

1. A **build fix**. What is *measured* is only that nothing exists at the path the script is given, while `quarkusBuild` demonstrably executed with the flag set (`> Task :openbank-billing-service:quarkusBuild`, `27 executed`). The cause is **not** established, and one live alternative changes the fix entirely: Quarkus may resolve a relative `store-schema-directory` against the augmentation's working directory, so the document could be produced one level away. The extension is present in every named service, so "no OpenAPI generation at all" is not the explanation.
2. Making the missing-document case **distinguishable** from a clean one.

Only then is a false-positive measurement meaningful.

**I did not touch the script.** Whether a missing document should be a finding, a hard error, or an explicit unknown is a design call — and #2312 (aml-service publishing `triggerType`, `riskScore`, `notes`, fields that exist in no DTO) shows what this gate is guarding against.

## On the method

Getting here took five probes, four of them broken: a local Gradle build that fails on the environment (Java 26 vs CI's 25), a case-sensitive job-name filter that could not have matched the lowercase `build` job, a reusable workflow that has no runs of its own, and a sample of runs that built nothing.

The measurement is recorded **with its run ids**, so the next reader can re-run it rather than re-derive it.

## Scope

No behaviour change — `blocked_on` is prose no checker reads. `check-gate-graduation` still reports 18 advisory rules with live dates; `gen-rules-opa-data.py` no diff; `yamllint`, `gate-lifecycle-metadata`, `advisory-gate-registration` pass.

## Linked issues

Refs #8797.

### DIFF
--- a/.github/workflows/_service-ci.yml
+++ b/.github/workflows/_service-ci.yml
@@ -226,6 +226,15 @@ jobs:
       # -Dquarkus.smallrye-openapi.store-schema-directory writes the document Quarkus
       # actually serves, as a side effect of the SAME quarkusBuild task this :build already
       # runs. Read by the openapi-request-schema-conformance step below.
+      #
+      # The path is RELATIVE TO THE SUBPROJECT, not to the repo root. Passing
+      # "<svc>/build/openapi-generated" therefore wrote to <svc>/<svc>/build/..., the step
+      # below found nothing at the path it was given, and — because a missing document is
+      # "Not a finding" there — the gate reported clean for six weeks having compared zero
+      # schemas. Measured 2026-09-05 by running the same command locally:
+      #   openbank-interest-service/openbank-interest-service/build/openapi-generated/openapi.yaml
+      # With the path corrected, that one service alone yields "2 request schema(s) compared,
+      # 21 finding(s)". Keep this relative to the subproject.
       - name: Build + test (:${{ inputs.service }})
         run: |
           # Capture image/lifecycle/time plus a transient daemon id from this job's Docker
@@ -249,7 +258,7 @@ jobs:
           }
           trap cleanup_test_evidence EXIT
           ./gradlew :${{ inputs.service }}:build --build-cache --console=plain --stacktrace \
-            -Dquarkus.smallrye-openapi.store-schema-directory="${{ inputs.service }}/build/openapi-generated"
+            -Dquarkus.smallrye-openapi.store-schema-directory="build/openapi-generated"
 
       # A Docker Hub pull failure surfaces as NAMED TEST FAILURES, ~900 lines above the actual
       # cause, so the reader goes hunting for a test bug that is not there. Diagnostic only:
--- a/openbank-libs/governance/rules.yaml
+++ b/openbank-libs/governance/rules.yaml
@@ -267,7 +267,7 @@ change_requirements:
     #   PlaceHoldRequest.expiresAt
     # each declared in the spec, absent from the DTO -- sent by a generated client and silently
     # dropped. So the backlog behind this date is REAL and was never a false-positive-rate
-    # question. 122 is the number to burn down; the rule stays advisory until it is near zero.
+    # question. 122 is the number to burn down (issue #8828); advisory until it is near zero.
     #
     # So the blocker is not readiness and not a false-positive rate: flipping this to enforced
     # would graduate a check that examines nothing, and the green would then be load-bearing.
--- a/openbank-libs/governance/rules.yaml
+++ b/openbank-libs/governance/rules.yaml
@@ -235,12 +235,21 @@ change_requirements:
     # compared. Verification note: the word is "compared", not "checked" — grepping the wrong
     # one returns empty across every run and reads as a confirmation while measuring nothing.
     #
-    # What is MEASURED is only that nothing exists at the path the script is given, while
-    # quarkusBuild demonstrably executed with the flag set. The CAUSE is not established, and
-    # one live alternative changes the fix: Quarkus may resolve a relative
-    # store-schema-directory against the augmentation working directory, so the document could
-    # be produced one level away. The extension is present in every named service, so "no
-    # OpenAPI generation at all" is not the explanation.
+    # CAUSE SETTLED 2026-09-05 by running the same command locally (isolated Gradle home,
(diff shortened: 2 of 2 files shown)


JiRaska:
## Cause settled, and fixed in this PR

The reviewer flagged that "the document is not produced" was an inference, and that a relative `store-schema-directory` resolving elsewhere would need a different fix. That is exactly what it was.

Ran the same command locally with an isolated Gradle home — `BUILD SUCCESSFUL in 12m 24s` — and the document exists:

```
openbank-interest-service/openbank-interest-service/build/openapi-generated/openapi.yaml   (988 lines, 18 paths)
```

**The path is relative to the subproject, not the repo root.** Passing `<svc>/build/openapi-generated` wrote to `<svc>/<svc>/build/...`. The document was produced the whole time, one directory away. Fixed by passing `build/openapi-generated`.

## What the gate says the first time it actually looks

```
check-openapi-request-schema-conformance: openbank-interest-service: 2 request schema(s) compared, 21 finding(s).
```

`InterestRateConfig` is almost entirely disjoint between spec and DTO:

- **published, on no DTO:** `ratePercent`, `tierMin`, `tierMax`, `validFrom`, `validTo`, `calculationMethod`
- **on the DTO, unpublished:** `annualRate`, `dayCount`, `effectiveFrom`, `effectiveTo`, `minBalance`, `maxBalance`, `rateType`, `accountId`, `active`, `id`, `createdAt`, `updatedAt`

Plus `AccrualRequest.balance` / `.currency` / `.productId`, all unpublished. Same shape as kyc #5895 — a client generated from that document cannot call the endpoint.

So the backlog behind the 2026-09-30 date is **real and was never measured**, not a false-positive-rate question. The rule stays advisory until a run with the corrected path gives the fleet number.

The step exits 0 by design, so this cannot turn a build red — it only makes the gate start looking.

JiRaska:
## The fleet number, first time it has ever existed

Built all **52** services carrying an `openapi.yaml` (one Gradle invocation, `--continue`, `BUILD SUCCESSFUL in 25m 14s`, all 52 documents produced) and ran the checker over each:

```
121 request schemas compared    122 findings    17 of 52 services affected
 88 parsed-not-published   (a DTO property the spec omits)
 34 published-not-parsed   (a spec property no DTO has — the client sends it, it is ignored)
```

| service | findings | schemas |
|---|---|---|
| pid | 27 | 11 |
| interest | 21 | 2 |
| domestic-payment | 13 | 1 |
| dispute | 12 | 8 |
| standing-order | 10 | 1 |
| sca | 8 | 4 |
| consent | 8 | 2 |
| sepa-payment | 6 | 1 |
| clearing | 6 | 1 |
| balance | 3 | 1 |
| kyc | 2 | 4 |
| party · lending · document · transaction · sanctions · swift | 1 each | — |

**35 services are clean.**

## The money-path half

`published-not-parsed` is the half a client acts on — it sends the field and the service ignores it:

```
CreateSepaPaymentRequest.currencyCode          declared in the spec, absent from the DTO
CreateSepaPaymentRequest.requestedExecutionDate
CreateDomesticPaymentRequest.currencyCode
CreateDomesticPaymentRequest.message
PlaceHoldRequest.expiresAt
```

A generated client filling in `currencyCode` on a SEPA or domestic payment is sending a field that is silently dropped.

## What this settles about the 2026-09-30 date

The blocker was never a false-positive rate — the gate had **never produced a number at all**. **122** is the burn-down. The rule stays advisory until it is near zero; the corrected path in this PR is what makes the number visible on every future PR.

JiRaska:
**These gate failures are not this PR's, and a re-run will not clear them — the branch has to take `main`.**

`gates (lint-supplychain-security)`, `gates (registry-kotlin-data)` and `Validate manifests` were all failing on `source-service-convention` and `gate-observability-declarations`, which were red on a **clean `origin/main`** checkout for several hours. #8823 fixed both; I re-verified on current `main`, where each checker now exits 0.

The part worth knowing: **`gh run rerun` is not enough.** I tried it here and on my own PR — the checks went green briefly and then failed again on the next run. Verified for this branch specifically:

```
git merge-base --is-ancestor 2e9243887b origin/<this branch>   ->  NO
```

`2e9243887b` is #8823. The gates run against the merged tree, and until the branch actually contains that commit, the tree CI evaluates does not have the fix — regardless of when the run starts.

Merging `origin/main` in clears it. On my own PR the merge was clean, both checkers then exited 0 on the merged tree, and `git diff --name-only origin/main` still showed only my own files, so nothing was silently pulled in or dropped.

I have not touched this branch — a worktree is holding it, so it is someone's live work. Flagging so the red is not mistaken for something in the diff.

Issue #8745, opened 2026-09-05
Fleet sweep (generalises #8718): three reachable state transitions that write no outbox event while their siblings do


JiRaska:
Generalising #8718 (swift-service): a fleet sweep for **aggregates whose state machine emits an outbox event for SOME transitions and not for OTHERS**.

The shape is invisible to every outbox gate we have. Those gates compare a state row against its outbox row; here there is no second row to diverge from — the event simply never exists. Nothing is missing at the row level, so `check-outbox-*`, the atomicity ITs and the event-contract baseline are all green about it.

Read against `origin/main` @ `cdd9541d1576a7c32651ae5aa7288184b390d1e1`. 41 modules carry outbox code in `src/main`; all were examined. **Three reachable asymmetries on a topic with a real consumer**, ranked money-path first, plus three weaker cases recorded so nobody re-derives them.

## Findings

| # | Service | Money path | Silent transition | Emitting sibling | Reachable via | Topic consumers | Verdict |
|---|---------|-----------|-------------------|------------------|---------------|-----------------|---------|
| 1 | `openbank-transaction-service` | yes | `COMPLETED -> REVERSED` | `transaction.completed`, `transaction.failed` | `POST /api/v1/transactions/{id}/reverse` | audit, analytics-sink, fraud | **confirmed** |
| 2 | `openbank-dispute-service` | no | any status incl. `RESOLVED_CUSTOMER` / `RESOLVED_MERCHANT` / `WITHDRAWN` | `dispute.opened`, `dispute.resolved` | `PUT /api/v1/disputes/{id}` | audit, engagement | **confirmed** |
| 3 | `openbank-clearing-service` | yes | empty cycle: batch born `SETTLED` | `batch.settled` + `net_settlement.post` | `POST` trigger-cycle endpoint | audit, clearing | **confirmed, low impact** |
| 4 | `openbank-standing-order-service` | no | `pause` / `resume` / `cancel` | `standing-order.failed.v1` | REST | **none external** | weak — nobody would have consumed it |
| 5 | `openbank-document-service` | no | document `PENDING_SIGNATURE` / `SIGNED` / archived | `document.generated.v1`, `signature-ceremony.completed.v1` | REST + ceremony flow | audit, analytics-sink | weak — no status-changed event type exists at all |
[...]

PR #8805, opened 2026-09-05
fix(dispute): announce a dispute that ends through the update path — every withdrawal was silent

JiRaska:
## The defect

A dispute reaching a **terminal** status through `DisputeService.update` wrote no outbox message. `withdraw()` delegates to `update()`, so **every withdrawal was silent**.

```kotlin
disputeRepo.update(updated).flatMap { saved ->          // one-arg overload: no outbox
    val event = DisputeTimelineEvent(
        eventType = "STATUS_CHANGED",                    // internal timeline row
        ...
    )
    timelineRepo.save(event).map { saved }
}
```

That `eventType = "STATUS_CHANGED"` is easy to misread as an emission — the field is spelled the same as the real thing. It is a `DisputeTimelineEvent`, internal to this service, and reaches no consumer. The service's actual outbox messages are built in named helpers (`openedOutboxMessage`, `resolvedOutboxMessage`, `remediationRequestedOutboxMessage`) and carry dotted names.

So the same terminal state was reachable two ways with different observability:

| path | on the wire |
|---|---|
| `resolve()` → `doResolve()` | `dispute.resolved` ✔ |
| `update(status = RESOLVED_CUSTOMER)` | nothing ✘ |
| `withdraw()` → `update(...)` | nothing ✘ |

`update` explicitly handles `RESOLVED_CUSTOMER`, `RESOLVED_MERCHANT` and `WITHDRAWN` — it stamps `resolvedAt` for exactly those three — so this is not a path that merely *could* end a dispute.

## Why it matters, concretely

`openbank-engagement-service` applies an **ADR-0220 D3.5 targeting exclusion** when it sees `dispute.opened` and lifts it **only** on `dispute.resolved`:

> `dispute.opened` applies the exclusion, `dispute.resolved` lifts it
> — `DisputeOpenedEventConsumer.kt:26-27`

A customer whose dispute was withdrawn, or resolved through `update` rather than `resolve`, therefore stayed excluded from targeting **permanently**.

The KDoc on `resolvedOutboxMessage` already names this exact risk, which is what makes the gap an omission rather than a design choice:

> without it a consumer can learn that a customer entered dispute but never that they left it, so an ADR-0220 exclusion applied on open would never lift

## The change

1. `update()` passes `resolvedOutboxMessage(updated)` to the **two-arg** repository overload. That overload is `@WithTransaction` (`DisputeRepositories.kt:73-83`), so the status change and the event commit together; with an empty list it is byte-for-byte equivalent to the one-arg version, so the non-terminal path is unchanged.
2. Guarded on the **previous** status (`dispute.status !in TERMINAL_STATUSES`) so a repeated `PUT` of an already-terminal status does not re-announce.
3. `TERMINAL_STATUSES` extracted and used for **both** the `resolvedAt` stamp and the emission. They were the same three names written out twice before, one of which emitted nothing — exactly the drift that let this happen.
4. A null `remediationOutcome` now serialises as JSON `null` rather than the four-character string `"null"` in a quoted field. `WITHDRAWN` reaches the builder with no outcome; `resolve` always sets one, so **that payload is unchanged**. The engagement consumer reads only `eventType` and `partyId`, so nothing downstream depends on the old rendering.

## Verification — falsified in both directions

The two existing tests mocked the **one-arg** overload, which is precisely why they could not catch this. They now capture the outbox list and assert the event. A third test covers the counter-case.

| tree | result |
|---|---|
| fix in place | **13/13 green**, 0 failures, 0 errors, **0 skipped** (JUnit XML) |
| production reverted, tests kept | **3 red** — both existing tests plus the counter-test |
| guard removed (`if (true)`, emit on every update) | **exactly 1 red — the counter-test**, other 12 green |

That third row is the one worth reading. Reverting production reddens the counter-test too, but only because the mock shape no longer matches — that is not evidence it detects over-emission. Removing just the guard proves it discriminates: without it, "emits on terminal" would also pass for an implementation that emits on **every** update, which would tell engagement to lift an exclusion while the dispute is still open.

`ktlintCheck` + `detekt` green.

**Local verification note:** this machine's shared Gradle state broke mid-session — every build failed with 58 `libs.*`-unresolved errors in `openbank-agent-service/build.gradle.kts`, reproducible from a **fresh worktree on pristine `origin/main`** and on a module that had compiled an hour earlier, with no relevant commit in 24h and 7 concurrent Gradle daemons sharing `~/.gradle`. Everything above was therefore run under an isolated `--gradle-user-home`, not the shared one. The repo is fine; the machine is not.

## Scope

`openbank-dispute-service` is not in `rules.yaml: money_path_services`. No API surface change, no migration, no new event type — `dispute.resolved` already exists on `openbank.dispute.events` and the baseline pins the topic, not the payload. This adds an **occurrence** of an existing event on a path that had none.

Refs #8745

## Security checklist

      (license, CVE, source).

Checked against the diff: two files, no `@Suppress`/`as Any` added, no dependency file touched, no secret-shaped line. No auth, crypto or payment path changes. The event payload carries `partyId`, which it already carried on this exact event type from the `resolve` path — this change adds no field and no new recipient, it only makes an existing event fire on a path that was silent.

### DIFF
--- a/openbank-dispute-service/src/main/kotlin/com/openbank/dispute/application/usecase/DisputeService.kt
+++ b/openbank-dispute-service/src/main/kotlin/com/openbank/dispute/application/usecase/DisputeService.kt
@@ -107,19 +107,29 @@ class DisputeService(
                     resolution = request.resolution ?: dispute.resolution,
                     chargebackAmount = request.chargebackAmount ?: dispute.chargebackAmount,
                     resolvedBy = request.resolvedBy ?: dispute.resolvedBy,
-                    resolvedAt = if (request.status in listOf(
-                            DisputeStatus.RESOLVED_CUSTOMER,
-                            DisputeStatus.RESOLVED_MERCHANT,
-                            DisputeStatus.WITHDRAWN,
-                        )
-                    ) {
+                    resolvedAt = if (request.status in TERMINAL_STATUSES) {
                         OffsetDateTime.now(clock)
                     } else {
                         dispute.resolvedAt
                     },
                     updatedAt = OffsetDateTime.now(clock),
                 )
-                disputeRepo.update(updated).flatMap { saved ->
+                // A dispute that ENDS here must announce it exactly as `resolve` does.
+                // engagement-service applies an ADR-0220 D3.5 targeting exclusion on
+                // `dispute.opened` and lifts it ONLY on `dispute.resolved`
+                // (DisputeOpenedEventConsumer:26-27), so a dispute that reaches a terminal state
+                // through this path without the event leaves that customer excluded permanently.
+                // `withdraw()` delegates here, so every withdrawal took that path. The timeline row
+                // below is internal to this service and reaches no consumer.
+                // Guarded on the PREVIOUS status so a repeated PUT of an already-terminal status
+                // does not re-announce; the empty list makes this call identical to the one-arg
+                // overload, which the repository implements with the same @WithTransaction.
+                val messages = if (updated.status in TERMINAL_STATUSES && dispute.status !in TERMINAL_STATUSES) {
+                    listOf(resolvedOutboxMessage(updated))
+                } else {
+                    emptyList()
+                }
+                disputeRepo.update(updated, messages).flatMap { saved ->
                     val event = DisputeTimelineEvent(
                         disputeId = saved.id,
                         eventType = "STATUS_CHANGED",
@@ -300,7 +310,9 @@ class DisputeService(
         // would never lift. Additive, and `dispute.remediation_requested` already carries one.
         payload = """{"eventType":"dispute.resolved","disputeId":"${dispute.id}",""" +
             """"reference":"${dispute.reference}","partyId":"${dispute.partyId}",""" +
-            """"outcome":"${dispute.remediationOutcome}",""" +
+            // WITHDRAWN reaches this builder with no remediation outcome; interpolating it
+            // straight would put the four-character string "null" in a quoted field.
+            """"outcome":${dispute.remediationOutcome?.let { "\"$it\"" } ?: "null"},""" +
             """"status":"${dispute.status}","resolvedAt":"${dispute.resolvedAt}",""" +
             """"occurredAt":"${dispute.resolvedAt?.toInstant() ?: Instant.now(clock)}",""" +
             """"sourceService":"$SOURCE_SERVICE"}""",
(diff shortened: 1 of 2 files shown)


JiRaska:
**The three red checks here are not this PR — `main` is red on `source-service-convention`, and the two findings are already known false positives.**

`gates (lint-supplychain-security)`, `gates (registry-kotlin-data)` and `Validate manifests` all fail on the same gate id. Read from each run's `gate-results-*.json` artifact rather than the job log, since the gates job also prints its self-test fixtures:

```
gates (lint-supplychain-security) FAILED: source-service-convention
gates (registry-kotlin-data)      FAILED: source-service-convention
```

Neither finding is in `openbank-dispute-service`:

```
FAIL openbank-domestic-payment: .../DelegatedSpendFinalizedAbsentEvent.kt:33 UNRESOLVED
FAIL openbank-kyc-service:      .../KycEvent.kt:69 UNRESOLVED
2 non-conforming write site(s) ... across 28 producer(s).
```

Verified against a **pristine `origin/main` worktree**: the checker exits 1 there with the identical two failures, so this predates the branch. And they are false positives — `DelegatedSpendFinalizedAbsentEvent` declares `const val SOURCE_SERVICE = "domestic-payment"` in its own companion object, which *is* the module directory name without the prefix; the checker cannot resolve a constant declared in the same file.

Already tracked as **#8827**, with **#8825** and **#8826** in flight to fix it. Nothing to do here — this PR goes green once one of those lands.

The dispute change itself was verified under an isolated `--gradle-user-home` (this machine's shared Gradle state is broken): compile exit 0, `DisputeServiceTest` 13/13 with 0 skipped, `ktlintCheck` + `detekt` green, and falsified both directions — production reverted gives 3 red, guard removed gives exactly 1 red (the counter-test).

JiRaska:
**Closing — superseded by #9057, which landed the same fix. Two things it did not carry are now up as #9595.**

#9057 (`14fcbb4a6`, "fix(dispute): operator update to a terminal status emits dispute.resolved") implements the same behaviour this PR does, and its comment cites #8745 finding 2 and the ADR-0220 exclusion reasoning. Verified equivalent on the core fix:

- emits `resolvedOutboxMessage(updated)` on the transition into a terminal status ✔
- guards on the **previous** status (`dispute.status !in terminal`) so a repeated `PUT` does not re-announce ✔
- carries the counter-test — `update to a non-terminal status emits no outbox event` ✔

That is the whole of what mattered, and there is no reason to prefer this branch: it is 394 commits behind and conflicts in both files precisely because #9057 rewrote them.

**Two differences survived the comparison, and one is a live defect:**

1. **A withdrawal now emits `"outcome":"null"` — the four-character string in a quoted field, not JSON null.** `resolvedOutboxMessage` interpolates inside the quotes; `withdraw` delegates to `update`, whose `copy()` never sets `remediationOutcome`, and the domain default is null. Before #9057 a withdrawal emitted nothing, so this is new on the wire. Up as **#9595**, with the `withdraw` test now *parsing* the payload rather than asserting a substring — the existing test mocked `update(any(), any())` and never opened it, which is how it got through.

2. #9057 spells the terminal triple out **twice** inside `update` — once for the `resolvedAt` stamp, once for the emission guard. Those two disagreeing is exactly the defect #9057 fixed, so #9595 declares it once. Behaviour-identical, easy to drop if a reviewer prefers.

**Worth recording about the process, since this is the second time.** #8762 was closed the same way five days ago: I opened it three hours before the duplicate that merged, and the only difference was which one a reviewer reached. Here the gap was larger — this PR sat open from 09-05 to 09-10 — so a claim check at authoring time could not have prevented either. A claim check protects against a duplicate that **already exists**; it cannot protect against one opened later against an unreviewed PR. What actually costs the duplicated work is review latency, and #9057 is the better outcome regardless: it is on `main`.

PR #8791, opened 2026-09-05
docs(adr): ADR-0282 Lípa, a financial-health loyalty ecosystem across rewards, segments, catalog and 360

JiRaska:
## Summary

ADR-0282 proposes **Lípa**, a loyalty ecosystem whose unit (the Lístek) rewards financial health, never spend or credit. It ties together four islands that today tell each other nothing about the customer: engagement gamification (ADR-0220/0261), the campaign segment DSL (ADR-0201), the product-catalog kernel (ADR-0257/0259) and Customer 360 (ADR-0210). Docs-only; `proposed` / `planned`.

## Type of change

- [ ] feat (new functionality)
- [ ] fix (bug fix)
- [ ] refactor (no behavior change)
- [ ] perf (performance)
- [x] docs
- [ ] chore (build / tooling)
- [ ] security (private until disclosed)
- [ ] breaking change

## Linked issues

Refs #2891 (ACCOUNT events carry no partyId; consent not ingested) — named as the first data gap the ADR closes.

## Changes

- `docs/adr/0282-lipa-loyalty-ecosystem-financial-health-rewards-personal-offerings.md` — D1–D9: closed-loop Lístek unit (no cash-out, no P2P, no fiat price); new `openbank-loyalty-service` ledger; sealed financial-health earn catalogue outside credit; benefit catalogue delivered by existing engines (WaiverEvaluator, InterestTier, CardTier, fx); IFRS 15 provisioning via billing with a per-party annual cap; feature-store micro-segments plus the data enrichments they need; `PersonalOfferingOverlay` as a monotonic, maker-checker term delta; reciprocal transparency in 360 and the app; household (Háj) and community (Les) trees.
- `docs/adr/README.md`, `DIGEST.md`, `index.json` — regenerated by `gen-index.sh`.

Measured context in the ADR: 2 segments, 4 DSL rules (2 unsupported), 4 `EarnSource` variants (1 wired), 1 challenge, 12 analytics topics, transaction facts limited to `initiated`, feature store with a single consumer (fraud).

## Definition of Done

- [x] Hexagonal layering preserved (domain has zero framework imports). — N/A, docs only
- [x] Unit tests added/updated. — N/A, docs only
- [x] Integration tests added/updated (if service boundary involved). — N/A
- [x] Contract tests updated (if public API changed). — N/A
- [x] `./gradlew detekt ktlintCheck koverVerify build` passes. — N/A, no Kotlin touched
- [x] No new lint warnings.
[...]

PR #8806, opened 2026-09-05
docs(governance): correct the event_change and dotted-keys graduation evidence

JiRaska:
Two governance records merged yesterday (#8782, #8780) were falsified by a second-agent review. Both are corrected here; I re-derived each number independently before writing it.

## `event_change` — the verdict inverts

The record said **0 findings on 20 commits**, and concluded the rule was ready to graduate on 2026-09-15. Replaying **today's** checker as an imported module over the real population (22 commits in 90 days touching a file containing `DomainEvent(`, `git show c^:path` vs `git show c:path`):

```
population commits 22
findings total 40
{'fc33a87176': 27, 'da16beed4a': 8, 'f75fe6d3af': 3, '069039649b': 1, 'd6e95d78a7': 1}
```

All 40 are the same class — a field added as non-nullable `T` with no default, which breaks a consumer **replaying the topic**. Under `--enforce` those five commits would have been **red**. The rule is therefore *not* ready to graduate on evidence; what it needs first is a ruling on whether an added non-nullable field is a real break here, or whether the checker is too strict.

**How the zero happened, which is the reusable part:** the advisory path exits 0 and emits findings as `::warning::`. Any measurement that reads the exit code — or that greps a checked-out commit where `check-event-schema-compat.py` **did not yet exist** (`No such file or directory`, grep count 0) — reads clean. The reviewer reproduced that false-clean before catching it. Same vacuous-measurement failure the note's own caveats warn about, one layer up.

## `mp_messaging_dotted_keys` — number wrong, attribution wrong

```
::error::Found 175 dotted mp.messaging leaf key(s)   (record said 171)
```

And the record credited all of them to the enforced sibling's deliberate baseline. That sibling watches **`group.id` and `auto.offset.reset`, incoming only**:

```
38 incoming value.deserializer   37 outgoing client.id
32 outgoing value.serializer     24 outgoing key.serializer
15 incoming key.deserializer     12 incoming group.id
12 incoming auto.offset.reset     4 incoming client.id
 1 outgoing bootstrap.servers
```

Only **24** are in its watch set, **11** baselined. The other ~151 are keys it has never looked at — **unassessed, not accepted**. The overstatement is an order of magnitude. The action (retire the duplicate rather than flip it) is unchanged; the ~151 need a ruling first.

## Scope

Prose only — no gate reads these strings. `check-gate-graduation` passes (17 advisory rules, unchanged count), `gen-rules-opa-data.py` produces no diff.

Refs #8782, #8780.


JiRaska:
**These gate failures are not this PR's, and a re-run will not clear them — the branch has to take `main`.**

`gates (lint-supplychain-security)`, `gates (registry-kotlin-data)` and `Validate manifests` were all failing on `source-service-convention` and `gate-observability-declarations`, which were red on a **clean `origin/main`** checkout for several hours. #8823 fixed both; I re-verified on current `main`, where each checker now exits 0.

The part worth knowing: **`gh run rerun` is not enough.** I tried it here and on my own PR — the checks went green briefly and then failed again on the next run. Verified for this branch specifically:

```
git merge-base --is-ancestor 2e9243887b origin/<this branch>   ->  NO
```

`2e9243887b` is #8823. The gates run against the merged tree, and until the branch actually contains that commit, the tree CI evaluates does not have the fix — regardless of when the run starts.

Merging `origin/main` in clears it. On my own PR the merge was clean, both checkers then exited 0 on the merged tree, and `git diff --name-only origin/main` still showed only my own files, so nothing was silently pulled in or dropped.

I have not touched this branch — a worktree is holding it, so it is someone's live work. Flagging so the red is not mistaken for something in the diff.

JiRaska:
**Correction to my comment above: the last paragraph is wrong for this PR.** I wrote "a worktree is holding it, so it is someone's live work" — that is true of the two sibling PRs I posted the same note on (#8791, #8802) and **not** of this one. `git worktree list` shows nothing holding `docs/graduation-evidence-corrected`, and its last commit is ~3.5 hours old.

Everything above that paragraph is measured and stands for this branch: it does not contain #8823 (`git merge-base --is-ancestor 2e9243887b origin/docs/graduation-evidence-corrected` → NO), which is why a re-run does not clear the gates.

The reason I still have not merged `main` in: "no worktree in *this* checkout" is not the same as "nobody is working on it" — another agent could hold it elsewhere — so I left the decision to whoever owns the branch rather than pushing to it on a weaker signal than I claimed to have.

I generalised a per-PR observation across three PRs in one shot, which is exactly the mistake the note itself was written carefully to avoid. Flagging rather than quietly editing it.

Issue #8793, opened 2026-09-05
ADR-0282 phase 2 — openbank-loyalty-service: the Lístek ledger, IFRS 15 provisioning, and the first three benefits


JiRaska:
Phase 2 of ADR-0282 (Lípa). Depends on phase 1 for earn signals; the ledger itself does not.

## Why

`openbank-engagement-service` has an opaque `Points` counter that cannot be exchanged for anything, by design. Its KDoc states where the exchange must live when the business wants one: outside the domain layer, as an explicit, reviewed, independently rate-limited call into the ADR-0143 billing path — never a method on the type. This issue builds that outside.

Today the only reward the platform can give is cash, in two places: `openbank-referral-service` (fixed reward with a ledger handoff, ADR-0266) and `openbank-incentive-service` (promo offers). ADR-0220 D3 rule 4 calls marketing cash out as the thing reward economics must not be.

## Scope

- [ ] New `openbank-loyalty-service`: append-only `earn | burn | expire | reverse` ledger per party with its own transactional outbox, shaped like `openbank-referral-service`. It records obligation in Lístky and **never posts money**.
- [ ] Every earn carries `earnSourceId`, a frozen `ruleVersion` and the id of the durable domain event that triggered it — the ADR-0261 `GamificationAward` discipline reused, including its idempotency lesson (key on the achievement, not on the triggering event id).
- [ ] Sealed `LeafEarnSource` catalogue, financial-health signals only. No variant may reference spend volume, card usage, credit uptake, utilisation, overdraft, or any product whose eligibility ADR-0142 decides. Exhaustiveness asserted in a test, as `EarnSourceTest` already does.
- [ ] `BenefitCatalog` — a reviewed catalogue entry per benefit (name, Lístek price, validity, delivering engine). First three: a fee waiver via a new `WaiveCondition` in `openbank-libs-domain` evaluated by `openbank-billing-service`; an interest bonus expressed as an `InterestTier` on the catalog product and read by `openbank-interest-service` through `CatalogInterestProfile`; one FX conversion at the reference rate with no margin.
- [ ] Redemption is reserve-then-commit-or-release, keyed by an idempotency token — the semantics ADR-0266 already gives promo codes. A retry cannot burn twice.
- [ ] IFRS 15 provisioning: a daily summary from loyalty-service, a balanced journal posted by `openbank-billing-service` to a loyalty-provision account, released on expiry or reversal. Reconcilable against the loyalty ledger per ADR-0026.
[...]

PR #8807, opened 2026-09-05
feat(loyalty): the Lístek ledger — earn, redeem, expire, with the cap as an outcome

JiRaska:
## Summary

The first slice of ADR-0282 phase 2 (#8793): a new `openbank-loyalty-service` owning the Lístek ledger. It mints Lístky for financial-health achievements, burns them for reviewed catalogue benefits, expires them first-in-first-out, and publishes every entry through its own transactional outbox. It posts no money and calls no money-path service.

## Type of change

- [ ] fix (bug fix)
- [ ] perf (performance)
- [ ] docs
- [ ] chore (build / tooling)
- [ ] breaking change

## Linked issues

Refs #8793 (ADR-0282 phase 2), Refs #8791 (the ADR itself, still open)

## Changes

- **Domain** — `Leaves` (opaque, non-negative, no conversion toward money), a sealed `LeafEarnSource` of ten financial-health signals, `EarnCatalog` with a frozen `RULE_VERSION`, `AnnualCap` returning a sealed decision, `LeafLedger`'s pure FIFO allocation and expiry rules, `BenefitCatalog` and `BenefitGrant`, and the `EarnOutcome` / `RedemptionOutcome` hierarchies.
- **Application** — earn, redeem, expire, read-summary and provisioning-summary use cases.
- **Infrastructure** — reactive Panache repositories writing domain rows and outbox rows in one transaction, the outbox dispatcher and Kafka publisher, a metrics adapter, a `suspend` scheduled sweep, and the REST resource.
- **Resources** — `V1__loyalty.sql`, `V2__synthetic_outbox_taint.sql`, `application.yaml` (port 8157), `openapi.yaml` (info.version 1.0.0), `governance.yaml`, `Dockerfile`, `version.txt`.
- **Registration** — `release-please-config.json` and `.release-please-manifest.json` at 0.1.0.
- **`openbank-loyalty-service/CLAUDE.md`** — the traps in this code and, explicitly, what is not built.

## Three decisions worth a reviewer's attention

**A capped earn is not a success and not an error.** `EarnOutcome.Capped` is its own type, its own HTTP shape (200 with `outcome: CAPPED`, nothing written) and its own counter. A loyalty programme where every party has hit the annual cap looks identical to a quiet week from any success metric, and that is the state most worth alerting on. This platform has already shipped the opposite arrangement, where `PushResult.skipped()` carried `success = true` and every undelivered push was counted as delivered.

**`BenefitGrantStatus.GRANTED` means owed and published, never applied.** Billing, interest and fx are money-path and are deliberately not called from this slice, so no field in any response or event is a delivery receipt. Wiring an engine should add a status value, not reuse this one.

**`V2__synthetic_outbox_taint.sql` is separate on purpose.** The column could have gone into V1's `CREATE TABLE`, but `check-synthetic-outbox-taint.py` establishes the property by globbing `*__synthetic_outbox_taint.sql`, so a service that folded the column inline would be invisible to the guard while appearing to satisfy it. Following the fleet convention keeps the check meaningful here.

## Definition of Done

- [x] Hexagonal layering preserved (domain has zero framework imports). — the `domain` package imports only `java.time` and `java.util`
- [x] Unit tests added/updated. — 28 unit tests across domain and application
- [x] Integration tests added/updated (if service boundary involved). — 6 tests in `LoyaltyLedgerOutboxIT` against a real Postgres via Testcontainers
- [x] Contract tests updated (if public API changed). — N/A, no consumer exists yet
- [x] `./gradlew detekt ktlintCheck koverVerify build` passes. — `test`, `detekt`, `ktlintCheck` run green locally; see the verification section
- [x] No new lint warnings. — detekt clean after extracting `LeafOutboxMessages`; remaining warnings are the fleet-wide Panache deprecation
- [x] OpenAPI spec regenerated (if REST API changed). — `openapi.yaml` written for all four routes, `openapi-server-port` and `openapi-enum-domain-drift` green
- [x] Flyway migration added + rollback note (if DB changed). — both migrations carry rollback notes
- [x] Avro schema versioned backward-compatibly (if event changed). — N/A, new topic, no existing consumer
- [x] Docs updated (README, ADR if architectural). — ADR-0282 in #8791; service CLAUDE.md added

## Verification

Run against a warm isolated Gradle home, because the shared `~/.gradle` currently fails build-script compilation for every module in this repo (reproduced on an untouched checkout, so it is not this branch).

```
GRADLE_USER_HOME=~/.gradle-local ./gradlew :openbank-loyalty-service:test :openbank-loyalty-service:detekt :openbank-loyalty-service:ktlintCheck
```

34 tests, 0 failures, **0 skipped** — the skipped count matters here, because a service that cannot boot reports its integration tests as skipped rather than failed, and a skip count scans as a pass.

Gates run locally, all green: `entity-column-names`, `configproperty-supplied`, `configproperty-kotlin-defaults`, `no-runblocking-in-a-scheduled-body`, `nonnull-jaxrs-param-ratchet`, `outbox-dispatch-enabled-guard`, `outbox-has-writer`, `synthetic-outbox-taint-persistence`, `flyway-default-datasource`, `duplicate-yaml-key-guard`, `release-registration-consistency`, `openapi-server-port`, `openapi-enum-domain-drift`, `openapi-version-not-taken`.

**The tests were falsified, not merely run.** Each sabotage was applied, the diff confirmed, the suite re-run, and the change reverted:

| Sabotage | Result |
|---|---|
| Drop the outbox write from `appendEarn` | "an earn writes the lot and its outbox row in one transaction" fails: expected 1, was 0 |
| Make the earn idempotency guard return null | "a redelivered achievement awards once" fails: 409 from the unique index instead of 200 |
| Make `AnnualCap.evaluate` always grant | three tests fail across `AnnualCapTest` and `EarnLeavesUseCaseTest` |

The first attempt at this proved nothing: the patch script asserted on a stale anchor and exited before writing, so the suite passed against unmodified code. The diff above each run is what establishes the sabotage actually landed.

## Security checklist

- [x] No secrets, tokens, passwords, or PII in code, config, tests, logs. — the only credentials in `application.yaml` are the fleet's `CHANGE_ME_LOCAL_DEV_ONLY` placeholders, same as every sibling service
- [x] No new `@SuppressWarnings`, `as Any`, `@Suppress("...")` without justification (state it in this PR). — one `@Suppress("MaxLineLength")` on the outbox claim SQL constant, matching every other outbox repository in the fleet
- [x] No new third-party dependency, OR: dependency review attached (license, CVE, source). — every dependency is an existing version-catalog alias
- [x] Auth / crypto / payment code changed? → tag `security-review-required`. — no; this service posts no money and calls no money-path service
- [x] PII path changed? → tag `gdpr-review-required`. — a new store keyed by `party_id` and nothing else; no name, contact or financial detail is persisted

## Compliance impact

- [ ] PCI DSS
- [ ] DORA
- [x] GDPR
- [ ] PSD2 / SCA / RTS
- [ ] AML / 5AMLD
- [ ] CNB reporting
- [ ] None

The ledger stores `party_id` and ledger facts only. Erasure is not yet wired: this slice has no `PARTY_ERASED` consumer, and that belongs with the same phase that wires the earn producers. It is recorded in the service's CLAUDE.md under what is not built, not left implicit.

The closed-loop properties that keep a Lístek outside e-money (no cash-out, no transfer, no fiat price, no purchase) are enforced structurally and asserted by test. ADR-0282 D1 makes a legal review a precondition for shipping this phase; that review has not happened, which is one reason nothing here is deployed.

## Rollout / Rollback

- Deployment strategy: none in this PR. There is no gitops manifest, so `auto-deploy` will not build this service and `check-deploy-coverage.sh` is correctly silent about it: its subject is a released component that has a gitops workload but no build.
- Rollback plan: revert the PR. Nothing outside the new directory changes except the two release-please registration files.
- Data migration reversible: yes, both migrations carry rollback notes, and no database exists yet to migrate.

## Reviewer notes

The service is registered with release-please at 0.1.0 and therefore becomes a released component on merge, while remaining undeployed. That is deliberate: the alternative is to register it later, which is the state that produced six historical "released but never built" services.

`ProvisioningSummaryUseCase` computes the IFRS 15 obligation and publishes it; `openbank-billing-service` will own the journal. Keeping that split is what lets this slice ship without an ADR-0030 threat model, and it also avoids a second ledger.


JiRaska:
**Corrected a real failure of mine.** `adr-registry-integrity-check` was red because this branch's code cites ADR-0282, and that ADR exists only in the unmerged #8791. The gate is right: a reference to an ADR with no file on the branch is a dangling citation.

Reproduced on the pull request's own merge commit rather than guessed at, which is what named the cause — the check passes on clean `main` and fails on `refs/pull/<n>/merge` with one error per citing file.

Fixed the way this repository documents for a stacked pull request: #8791's branch is merged **into** this one, so the ADR is present here and CI can judge this branch independently. The shared files drop out of the diff as #8791 lands on `main`. The check now passes locally: 273 ADRs, derived artefacts fresh.

Merge order that follows from it: #8825 (both gate fixes) → #8791 (the ADR) → this one.

JiRaska:
**Seven enforced gates found what the first slice left out, and all of it was mine.** Together they are the surface a new event producer has to occupy before anyone can trust what it emits.

| gate | what was missing | fix |
|---|---|---|
| `scheduler-liveness-adoption` | the sweep had no heartbeat | `registerWorkflowLiveness` on `StartupEvent` |
| `source-service-presence` | the payload carried no `sourceService` | stamped on every message |
| `event-contract-coverage-ratchet` | no contract for the topic | AsyncAPI 3.0, three message types |
| `kafka-topic-existence` | no `KafkaTopic` CR | declared in gitops |
| `event-consumer-liveness` | nothing consumes the topic | allowlisted with a reason |
| `service-runbook-drift` | no runbook | generated, only the new file |
| `vex-overlay-coverage` | no VEX overlay | gap declared with an owner, #8830 |

Two deserve more than a row.

**`sourceService` is not decoration.** audit-service attributes a row by this field when present and *derives* it from the topic name when absent. `audit_entries` is append-only at the database and `source_service` is chain-hashed into `record_hash`, so a row attributed by derivation can never be corrected afterwards.

**The VEX overlay is deliberately not written.** The gate says it in its own header: a VEX statement is a security judgement about a specific CVE and a specific artifact, and fabricating one to make a gate green is strictly worse than the gap it hides. Nobody has triaged this service's resolved dependency set, so the gap is declared with an owner instead — the mechanism the gate provides for exactly this case. The service is not deployed anywhere, so nothing is exposed today; that lowers the urgency and does not remove the obligation.

The liveness registration hangs off `StartupEvent` rather than `@PostConstruct`, because `@ApplicationScoped` is lazy: a `@PostConstruct` would first run when the cron first fires, leaving the gauge **absent** until then, and absent is a different signal from stale. `recordSuccess` is called only where the work returned, never on the failure branch.

Verified locally: 34 tests, 0 failures, 0 skipped, detekt and ktlint clean, all seven gates green. The one gate still red on this branch is `source-service-convention`, which is red on `main` for two unrelated services and is fixed in #8825.

Issue #8794, opened 2026-09-05
ADR-0282 phase 3 — micro-segments, PersonalOfferingOverlay, and reciprocal transparency in 360 and the app


JiRaska:
Phase 3 of ADR-0282 (Lípa). Depends on phase 1 (the features exist) and phase 2 (a tier exists to segment on).

## Why

Measured on `main`, 2026-09-05: `SegmentCatalog` holds **two** segments, `actives` and `actives-tenured-30d`. Product personalisation is `EligibilitySegment`, six coarse values — `RETAIL`, `STUDENT`, `SENIOR`, `BUSINESS`, `PREMIUM`, `ALL`. Nothing anywhere produces a per-party variant of a product. The bank cannot express "this customer, this term".

## Scope

- [ ] Segment DSL gains `FeatureAtLeast(featureName, threshold)` and `LoyaltyTierIs(tier)`, evaluated against the ADR-0140 online store rather than silver. Segments stay versioned code reviewed in a PR (ADR-0201 D1) — no free-form SQL from any UI, ever.
- [ ] `PersonalOfferingOverlay` in the ADR-0257 catalog kernel: a bounded term delta (fee, rate tier, limit, card tier, benefit bundle) over a published offering revision, for one party, with a validity window and the **micro-segment version** that justified it.
- [ ] The overlay is **monotonic** — it may only improve a term relative to the published revision. Personalisation can then never become discriminatory pricing, and that is a construction-time invariant, not a review convention.
- [ ] Credit products are excluded from overlays entirely. Price and eligibility stay ADR-0142's; a pre-approved credit offer stays ADR-0220 D4's.
- [ ] Deterministic policy proposes the overlay; ADR-0259's AI may draft and explain but never decides; maker-checker publishes; the customer accepts in the app.
- [ ] Customer 360 Lípa panel: tree, tier, earn history, active benefits, micro-segment memberships **with their versions**, active overlay, and the reason each was chosen (ADR-0222 explanation over reason codes).
- [ ] The same panel in `openbank-app`, plus a one-tap opt-out of personalisation that keeps earned value — ADR-0220 D3 rule 2 extended from the hub to segments and overlays.
- [ ] Vulnerable customers (the ADR-0200 D6 adverse-state set) excluded from targeting and from overlays at the eligibility stage, as ADR-0220 D3 rule 5 already does for challenges. They keep earning and redeeming on their own initiative.

## Acceptance — by effect

[...]

PR #8820, opened 2026-09-05
feat(admin-ui): the Lípa console — teach the programme and administer it in one place

JiRaska:
## Summary

The Lípa console: one admin-UI workspace that teaches the loyalty programme and administers it, plus a Lípa panel on Customer 360. It explains what a Lístek is, why it is deliberately not money, what the bank owes because of it, and where AI is allowed to help. It reads the loyalty-service catalogues and a party's ledger through the ADR-0056 relay.

## Type of change

- [ ] fix (bug fix)
- [ ] perf (performance)
- [ ] docs
- [ ] chore (build / tooling)
- [ ] breaking change

## Linked issues

Refs #8794 (ADR-0282 phase 3, the transparency half), Refs #8791 (the ADR), Refs #8793 (the service)

## Changes

- **`/loyalty`** — five tabs: Principles, Catalogues, Customer, Finance and law, AI.
- **Lípa panel on Customer 360** — balance, earned this year, next expiry, mounted beside the adverse-state and portfolio panels.
- **BFF routes** — `/api/loyalty` (catalogues plus the obligation) and `/api/loyalty/party/[partyId]`.
- **`lib/loyalty/lipaContent.ts`** — the teaching content as bilingual data, so the page carries no hardcoded copy and the claims stay assertable.
- **Registration** — `loyalty:view` permission, the route prefix, the sidebar entry.
- **Tests** — `loyalty-console.guard.test.ts` and `customer-360-lipa.test.tsx`.

## Why the management surface is deliberately read-only

The earn and benefit catalogues are code, changed in a pull request, the same discipline segments and campaign templates already use. So the console shows what is in force and says plainly that changing it is a review, rather than offering an editor that would have to be disabled or, worse, would quietly become the second place a reward economy is defined. An operator who does not know that reads a refusal as a fault. That is why teaching and administering are one screen here and not two.

The console is otherwise honest in the same direction throughout:

- A **capped earn** and an **unaffordable redemption** are shown as the two endings that are not errors.
- A **granted benefit** is described as owed and published, never as delivered. `graceful-states` has a sibling problem in the push adapter that counted an APNs 200 as a delivery; a console that rendered "granted" as "delivered" would put that same mistake in the place an operator goes to check.
- The **legal tab states what is unsettled**: the EMD2 limited-network review has not happened, and erasure is not wired.
- Every **AI role is marked as a proposal**, because none is wired.

## Where AI fits, and where it stops

Four roles, all proposals, each carrying its own guardrail: a catalogue drafter whose output is a reviewable draft and not a write, an explainer that may only write from reason codes a decision actually produced, a watch that files findings on the states nobody alerts on (cap saturation, an expiry wave, an earn source nobody earned), and message ranking that runs in shadow. The red lines are on screen and asserted by test: AI proposes and never decides, no model sets a price, a term or credit eligibility, every proposal is two-person approved, and a vulnerable customer stays excluded from targeting while keeping the right to earn and redeem.

## Definition of Done

- [x] Hexagonal layering preserved (domain has zero framework imports). — N/A, no Kotlin in this PR
- [x] Unit tests added/updated. — two new test files, 37 tests
- [x] Integration tests added/updated (if service boundary involved). — N/A; the BFF contract is asserted in the guard test
- [x] `./gradlew detekt ktlintCheck koverVerify build` passes. — N/A, no Gradle module touched
- [x] No new lint warnings. — `eslint` clean on every new file, with a probe confirming the linter fires
- [x] OpenAPI spec regenerated (if REST API changed). — N/A here; the service endpoints are in #8807
- [x] Avro schema versioned backward-compatibly (if event changed). — N/A
- [x] Docs updated (README, ADR if architectural). — the console itself is the operator documentation; ADR-0282 is #8791

## Verification

- `npm test` (never a bare vitest run — the pretest hook bakes the governance and catalog artifacts the suite needs).
- The two new files on their own: 37 tests pass. With `graceful-states.guard.test.ts` alongside: 236 pass.
- `npx tsc --noEmit`: the 20 errors reported are all pre-existing, in generated modules the pretest step produces. None is in a file this PR adds.
- `eslint` on every new file: clean, exit 0. Established with a probe rather than assumed: an `any` annotation is not flagged by this config, but a conditional hook is, so the linter demonstrably fires.

**The mermaid probe was falsified before being trusted.** The diagram is parsed with the same mermaid the console renders with, and the test first requires that probe to reject three known-broken diagrams: the `@Scheduled` label form this repository was actually broken by, a semicolon in a sequence message, and a bare `@`. The first version of the probe used `A[Outbox @Scheduled every 5s]`, which mermaid **accepts** — it would have reported a clean parse while being structurally unable to detect the defect it was written for.

**One real finding against this PR came from an existing guard.** `graceful-states.guard.test.ts` failed on `LipaPanel.tsx` for building a raw `HTTP ${status}` string, which an operator reads as the app being broken. Fixed by reporting a kind instead. That is the guard doing exactly its job on new code.

## A note on the suite, since a reviewer will see it

Three full-suite runs on this branch, and they do not agree with each other:

| run | result | machine |
|---|---|---|
| 1 | 435 files passed, 2334 tests passed | moderate load |
| 2 | 10 files failed, 426 passed | load rising |
| 3 | 27 files failed, 409 passed | load average 95 |

The spread is the machine, not the branch. Every failure in runs 2 and 3 except two was in a file this PR never touches, with individual tests taking 9 to 51 seconds — the timeout signature of a saturated host, measured at load average 95 with eight Gradle daemons and 34 node processes belonging to other work on the same machine. It is also the non-determinism recorded on #5285, where two runs of a clean `main` worktree gave different single-test failures.

The two failures that were genuinely this PR's are fixed, and both are described above. To separate the branch from the host rather than assert it: this PR's two test files, all five Customer 360 tests, the i18n guard and the graceful-states guard were run together in isolation at load average 115 — **9 files, 461 tests, all passing**. That includes `customer-360-documents.test.tsx`, which failed in run 3 and passes here with this PR's change present.

A red on an untouched file in this repository's admin-ui suite is worth re-running before it is worth reading.

## Security checklist

- [x] No new third-party dependency, OR: dependency review attached (license, CVE, source). — none; mermaid, lucide and radix are already dependencies
- [x] Auth / crypto / payment code changed? → tag `security-review-required`. — no
- [x] PII path changed? → tag `gdpr-review-required`. — the party route takes a UUID, validates it as one before use, and returns ledger facts only. No name, contact or financial detail passes through it.

Both BFF routes enforce `loyalty:view` with `requireApiPermission` at the route boundary, not only in the page: edge middleware and the client guard keep navigation truthful but neither protects a direct call to an internal route.

## Compliance impact

- [ ] PCI DSS
- [ ] DORA
- [x] GDPR
- [ ] PSD2 / SCA / RTS
- [ ] AML / 5AMLD
- [ ] CNB reporting
- [ ] None

The console is the operator-facing half of ADR-0282 D8's reciprocal transparency: it shows the customer's own view and no operator-only field, which is what makes the GDPR Art. 22 position ("deterministic, explained, opt-out-able") checkable rather than asserted. The panel test pins that symmetry, so a future operator-only figure fails a test instead of quietly voiding the claim.

## Rollout / Rollback

- Deployment strategy: rolling, with the rest of admin-ui.
- Rollback plan: revert the PR. Nothing outside the new files changes except three registration lines and two lines in the Customer 360 page.

## Reviewer notes

`loyalty-service` is not deployed anywhere yet, so in every current environment this console renders the shared `DataUnavailable` panel in its data tabs, and the Principles, Finance-and-law and AI tabs still work because they do not depend on the service. The `not_deployed` and `unreachable` states are kept apart deliberately: mapping every error to "not deployed" is how a healthy service with one unregistered route was reported as absent for the whole life of that route (#3371).

The two service endpoints this page reads, `/earn-sources` and `/provisioning`, are added in #8807 and are not on `main` yet.


JiRaska:
**Corrected a real failure of mine.** `adr-registry-integrity-check` was red because this branch's code cites ADR-0282, and that ADR exists only in the unmerged #8791. The gate is right: a reference to an ADR with no file on the branch is a dangling citation.

Reproduced on the pull request's own merge commit rather than guessed at, which is what named the cause — the check passes on clean `main` and fails on `refs/pull/<n>/merge` with one error per citing file.

Fixed the way this repository documents for a stacked pull request: #8791's branch is merged **into** this one, so the ADR is present here and CI can judge this branch independently. The shared files drop out of the diff as #8791 lands on `main`. The check now passes locally: 273 ADRs, derived artefacts fresh.

Merge order that follows from it: #8825 (both gate fixes) → #8791 (the ADR) → this one.

PR #8821, opened 2026-09-05
ci(gates): declare the measured runtime budget on five gates that landed without one

JiRaska:
## Summary

`gate-observability-declarations` is failing on `origin/main`, so every open pull request is red on it and none of them caused it. Five gates landed without `budget_seconds`, taking the undeclared count from the baseline's 159 to 164. This declares a measured budget on each.

## Type of change

- [x] fix (bug fix)
- [ ] perf (performance)
- [ ] docs
- [ ] chore (build / tooling)
- [ ] breaking change

## Linked issues

Refs #8791 — found while the ADR-0282 pull request was red on this gate for a cause it did not create.

## Changes

Five lines in `.github/gates/gates.yaml`, each after that gate's `mode:` where the rest of the estate carries it:

| gate | measured | declared |
|---|---|---|
| `alert-rca-ledger-guards` | 4.1s | 30 |
| `standing-critical-digest-guards` | 4.2s | 30 |
| `agent-bounded-write-surface` | 5.0s | 30 |
| `source-service-convention` | 14.1s | 60 |
| `ruleset-bypass-actors` | network-bound | 60 |

The four local ones were timed on this machine at load average 95, so the real figures are lower. Budgets are deliberately generous: one that fails on a busy runner teaches people to ignore the signal, which costs more than a loose bound. `ruleset-bypass-actors` makes a single GitHub rulesets API call and does no local work, so its budget covers a slow API.

## Verification

```
python3 .github/scripts/check-gate-observability-declarations.py --enforce
```

Before: `::error::missing_budget_seconds: 164 gates, baseline allows 159`, exit 1. After: `OK -- both ratchets hold`, exit 0.

**The `--enforce` flag is the part worth stating.** Without it the script exits 0 and downgrades the same finding to `::warning::`. A verification run without the flag would have reported this fix as working while being structurally unable to fail — which is exactly how a check gets trusted for the wrong reason. The gate's own `run:` uses `--enforce`, so that is what was reproduced.

Falsified: removing one of the five declarations returns the gate to exit 1 naming that gate, and restoring it returns it to exit 0.

The baseline is untouched. The gate's message says a new gate must declare its budget rather than widen the allowance, so widening it would have been the wrong repair.

## Definition of Done

- [x] Hexagonal layering preserved (domain has zero framework imports). — N/A, manifest only
- [x] Unit tests added/updated. — N/A; the gate's own self-test covers the checker
- [x] Integration tests added/updated (if service boundary involved). — N/A
- [x] `./gradlew detekt ktlintCheck koverVerify build` passes. — N/A, no Gradle module touched
- [x] No new lint warnings.
- [x] OpenAPI spec regenerated (if REST API changed). — N/A
- [x] Avro schema versioned backward-compatibly (if event changed). — N/A
- [x] Docs updated (README, ADR if architectural). — N/A; the declarations are self-describing

## Security checklist

- [x] No new third-party dependency, OR: dependency review attached (license, CVE, source). — none
- [x] Auth / crypto / payment code changed? → tag `security-review-required`. — no
- [x] PII path changed? → tag `gdpr-review-required`. — no

## Compliance impact

- [ ] PCI DSS
- [ ] DORA
- [ ] GDPR
- [ ] PSD2 / SCA / RTS
- [ ] AML / 5AMLD
- [ ] CNB reporting
- [x] None

## Rollout / Rollback

- Deployment strategy: N/A, CI manifest only.
- Rollback plan: revert the PR; the gate returns to failing as it does today.

## Reviewer notes

`main` is red on a second enforced gate at the same time, and this PR does not touch it: `source-service-convention` reports two ambiguous producers, in `openbank-domestic-payment` and `openbank-kyc-service`, each declaring two spellings of its own service name. That is issue #5902 and it needs a decision about which spelling is canonical for a delegated-spend event, which is an attribution question rather than a tidy-up, so it is left alone here rather than guessed at.

### DIFF
--- a/.github/gates/gates.yaml
+++ b/.github/gates/gates.yaml
@@ -2902,6 +2902,32 @@ gates:
 
   # ==== api ===================================================================
 
+  # Issue #8351 (ROADMAP M2): "idempotency coverage" had no measured inventory — which
+  # mutating endpoints require an idempotency key existed nowhere as data. This gate parses
+  # every released service's openapi.yaml, classifies each mutating operation
+  # (required-header / required-body — both fleet idioms plus the Berlin Group X-Request-ID —
+  # / optional / absent), prints the full inventory, and enforces the hard rule on the one
+  # class where a retry double-books: money-path (rules.yaml) POST. Today's 94 uncovered
+  # money-path POSTs are shrink-only baselined with per-class reasons; a NEW uncovered one
+  # fails. Writing the gate caught real drift on day one: psd2's v2 PIS spec declared an
+  # X-Idempotency-Key the code never reads (#8772).
+  - id: idempotency-coverage-money-path
+    name: "money-path POST endpoints declare a required idempotency key (#8351, enforced)"
+    group: api
+    mode: enforced
+    budget_seconds: 20   # 52 OpenAPI parses via gatelib cache, measured ~2s
+    rationale: >-
+      A duplicate POST on a money-path endpoint is a double debit; without a ratchet, a new
+      endpoint ships without an idempotency key and nothing objects. The gate also prints the
+      fleet-wide inventory the M2 exit criterion asks for.
+    review_after: 2027-03-05
+    min_subjects: 40   # 52 released services with a spec today (2026-09-05); some slack
+    selftest: python3 .github/scripts/check-idempotency-coverage.py --self-test
+    selftest_expect: pass
+    selftest_inputs: [".github/scripts/check-idempotency-coverage.py"]
+    run: |
+      python3 .github/scripts/check-idempotency-coverage.py
+
   # The verification reconciler (.github/scripts/pact-reconcile-verifications.py)
   # dispatches verify-provider.yml with the PACTICIPANT NAME as its `service` input,
   # which only works because every pacticipant name is exactly a module directory —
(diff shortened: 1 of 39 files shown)


JiRaska:
**This PR cannot go green on its own, and that is not its fault.** Its own gate passes; it fails on `source-service-convention`, a second enforced gate that is red on `main` for an unrelated reason — the resolver reports two correct sites as ambiguous.

That fix is #8825, and #8825 was symmetrically blocked by this one. Measured on each branch: here `check-gate-observability-declarations.py --enforce` exits 0 and `check-source-service-convention.py` exits 1; on #8825's branch, the reverse.

To break the deadlock this branch has been merged **into** #8825, which now carries both fixes and passes both gates. Nothing here was changed or force-pushed.

Merge #8825 and this PR becomes empty; close it then, or merge it first and #8825 will simply carry a commit it already has.

JiRaska:
Absorbed and closing. #8825 carried this branch's commit and merged as `fce1bce`, so both gate fixes are on `main` and this pull request's content is now identical to it.

Verified by content rather than by the diff GitHub shows: `git diff origin/main origin/ci/gate-budget-declarations` is empty. The +1988/-56 in the header is computed against the stale merge-base, which is why a fully absorbed branch can still look like a large change.

The fix is confirmed live on `main`: `check-gate-observability-declarations.py --enforce` exits 0 there, as does `check-source-service-convention.py`. The queue is unblocked.

Issue #8808, opened 2026-09-05
Card platform (ADR-0283): scheme-agnostic capability ports, Visa + Mastercard adapters, control plane, AI agents


JiRaska:
## Why

The platform has a good **card issuer registry** (`openbank-card-issuance-service`, ADR-0113/0194/0262) and roughly none of the rest of a card ecosystem. Measured 2026-09-05 on `origin/main`:

- ADR-0190 declares authorisation, 3DS, PIN and any processor **out of scope** — no port to a processor or a scheme exists.
- `POST /api/v1/cards/{id}/authorizations` (`CardAuthorizationPolicy`) exists and **nothing calls it** (grep outside the service: zero). A money-gating decision engine with no input.
- `openbank.cards.events` is consumed only by campaign-service and audit-service. **The ledger never sees card spend.**
- No tokenisation (wallets, MDES/VTS), no card disputes/chargebacks (ADR-0117 is generic), no account updater, no push-to-card, no BIN management, no scheme reporting, no interchange/fees model.
- `CardNetwork { VISA, MASTERCARD, AMEX, UNIONPAY }` is an enum with no behaviour behind it.
- card-issuance was **not** in `rules.yaml: money_path_services` although it already ships an authorisation decision.

## Decision

ADR-0283 (supersedes ADR-0190): a **scheme-agnostic card platform** — one domain, N adapters. Every network capability is a port with an in-repo sandbox simulator and a real adapter behind it; the bank chooses the network **per product** from one control plane. Visa and Mastercard are the first two adapters (both developer portals offer free sandboxes; the commercial membership/BIN-sponsor track runs in parallel and does not block engineering).

## Phases (each a child issue)

- [ ] Phase 0 — governance: ADR-0283, card-issuance into `money_path_services`, SLOs, journey accountability, threat-model refresh (this epic's opening PR).
- [ ] Phase 1 — card-processing: authorisation → hold → clearing → ledger posting, sandbox acquirer simulator, accounting-day E2E. (#8809)
- [ ] Phase 2 — Visa Developer + Mastercard Developers sandbox adapters behind capability ports: BIN/merchant lookup, tokenisation, push payments, disputes. (#8810)
- [ ] Phase 3 — Card Center control plane in admin-ui + MCP providers: network per product, disputes desk, token lifecycle, capability matrix. (#8811)
[...]

PR #8826, opened 2026-09-05
fix(ci): unblock main — declare five gate budgets, disambiguate two sourceService constants

JiRaska:
## Summary

Two enforced gates fail on a clean `origin/main` checkout, so every open PR is red on them regardless of its own content. This fixes both. Reproduced by running the gate manifest in a detached worktree at `origin/main` with no local changes.

## Type of change

- [x] fix (bug fix)
- [ ] perf (performance)
- [ ] docs
- [ ] chore (build / tooling)
- [ ] breaking change

## Linked issues

Refs #8808 (found while running the full manifest for the card-platform phase-0 PR, #8819)

## Changes

**`gate-observability-declarations`** — five gates were added without `budget_seconds`, taking the undeclared count from 159 to 164 against a baseline that allows 159. The baseline is a ratchet whose whole purpose is that a new gate arrives declared, so the fix is to declare the five, not to raise the number. Each budget is set from the seconds measured in CI run 33956518443 with headroom, and the measurement is stated inline next to the value.

| Gate | Measured in CI | Declared |
|---|---|---|
| alert-rca-ledger-guards | 0.2 s | 20 s |
| standing-critical-digest-guards | 0.1 s | 20 s |
| ruleset-bypass-actors | 0.7 s | 30 s |
| agent-bounded-write-surface | 0.8 s | 20 s |
| source-service-convention | 0.8 s | 30 s |

**`source-service-convention`** — two modules each declare one constant name with two different values, so every reference to that simple name is ambiguous. The checker reports such a module UNRESOLVED rather than guessing, which is the right call: guessing would produce a confident wrong verdict in either direction.

- `openbank-domestic-payment` used `SOURCE_SERVICE` both for what it emits (`domestic-payment`) and for the producer whose snapshots it accepts (`delegation-service`). The second is a different fact and is renamed `EXPECTED_PRODUCER_SERVICE`.
- `openbank-kyc-service` had a metrics-tag `SERVICE = "kyc"` in the orphaned-party gauge and an event `SERVICE = "kyc-service"`. The event one is renamed `SOURCE_SERVICE`.

No emitted value changes in either module. Both compile, including the domestic-payment test tree that referenced the renamed constant.

## Definition of Done

- [ ] Unit tests added/updated — n/a, this is a rename plus a manifest declaration; the two gates are themselves the tests and both now pass.
- [ ] Integration tests added/updated — n/a.
- [ ] Contract tests updated — n/a; the Pact fixture reference was updated as part of the rename.
- [x] `./gradlew detekt ktlintCheck koverVerify build` passes — the two touched modules compile (`compileKotlin`, `compileTestKotlin`).
- [x] No new lint warnings.
- [ ] OpenAPI spec regenerated — n/a.
- [ ] Flyway migration added — n/a.
- [ ] Avro schema versioned backward-compatibly — n/a; no wire value changes.
- [x] Docs updated — the reason for each rename is written where the constant is.

## Security checklist

- [x] No secrets, tokens, passwords, or PII in code, config, tests, logs. Verified against the diff.
- [x] No new `@SuppressWarnings`, `as Any`, `@Suppress("...")`. Verified: `git diff origin/main...HEAD | grep -E '^\+.*(@Suppress|as Any)'` is empty.
- [x] No new third-party dependency. No `build.gradle*`, `libs.versions.toml` or `package.json` in the file list.
- [x] Auth / crypto / payment code changed? A constant in domestic-payment's delegated-spend projection was renamed; the value it compares against is unchanged, so the guard accepts and rejects exactly what it did before.
- [x] PII path changed? No.

## Compliance impact

- [ ] PCI DSS
- [ ] DORA
- [ ] GDPR
- [ ] PSD2 / SCA / RTS
- [ ] AML / 5AMLD
- [ ] CNB reporting
- [x] None

The kyc event's `sourceService` value is unchanged (`kyc-service`), so audit attribution and the record-hash chain are untouched.

## Rollout / Rollback

- Deployment strategy: rolling. No behaviour change to deploy.
- Rollback plan: revert. The gates return to failing on main, which is the state before this PR.

## Reviewer notes

The thing worth checking is that neither rename changes a value that reaches the wire or the audit chain. `domestic-payment` emits `domestic-payment` and accepts `delegation-service`, both unchanged; `kyc-service` emits `kyc-service`, unchanged. Only the Kotlin identifiers moved.

### DIFF
--- a/.github/gates/gates.yaml
+++ b/.github/gates/gates.yaml
@@ -203,6 +203,7 @@ gates:
     selftest_exempt: "is-a-test-suite — the run: IS alert-rca-ledger.py --self-test"
     name: "alert-RCA ledger guards (MTTR granularity, recurrence, observation coverage)"
     group: lint
+    budget_seconds: 20   # 0.2s in CI (run 33956518443), 7.3s through the local runner
     mode: enforced
     rationale: "The MTTR granularity refusal and the recurrence guard are the only things stopping ADR-0241 D1/D3 from publishing a number nobody measured, and the coverage outcome is what stops an empty week reporting as a clean one. They were wired only inside a scheduled workflow that has never once succeeded, so they had never run on any lane. Recheck for removal when a durable alert open/close feed exists (#5869) and the MTTR half is no longer a refusal."
     review_after: "2027-02-21"
@@ -213,6 +214,7 @@ gates:
     selftest_exempt: "is-a-test-suite — the run: IS standing-critical-digest.py --self-test"
     name: "standing-critical digest guards (severity filter, observation envelope)"
     group: lint
+    budget_seconds: 20   # 0.1s in CI (run 33956518443), 3.5s locally
     mode: enforced
     rationale: "The digest's severity filter and observation envelope are the producer half of the same control; if the envelope shape drifts, the weekly review silently folds nothing and reports a clean week. Same lane gap as its consumer above. Recheck when the digest stops being the ledger's only observation source."
     review_after: "2027-02-21"
@@ -1924,6 +1926,7 @@ gates:
                         # so the floor moves with the list rather than tracking it exactly.
     name: "agent bounded GitHub write surface (ADR-0031 D9 phase 3, enforced)"
     group: registry
+    budget_seconds: 20   # 0.8s in CI (run 33956518443), 3.6s locally
     mode: enforced
     rationale: "flaky-test-hunter is the only agent with a real GitHub write path; its non-money-path confinement and its inability to merge or approve held only because a human read a Kotlin constant, and money_path_services grows over time."
     review_after: "2027-02-20"
@@ -2565,6 +2568,7 @@ gates:
   - id: source-service-convention
     name: "sourceService == module directory name (issues #5256/#5902, enforced)"
     group: kotlin
+    budget_seconds: 30   # 0.8s in CI (run 33956518443), 2.7s locally; it walks every module src tree
     mode: enforced
     min_subjects: 20
     rationale: >-
@@ -4790,6 +4794,7 @@ gates:
   - id: ruleset-bypass-actors
     name: "who may bypass main-protection is pinned, and widening it fails a PR (Refs #4828, enforced)"
     group: lint
+    budget_seconds: 30   # 0.7s in CI (run 33956518443); it makes GitHub API calls, so the ceiling covers a slow response
     mode: enforced
     rationale: >
       The bypass surface of main-protection is a repository SETTING, not a file: it is edited
--- a/openbank-domestic-payment/src/main/kotlin/com/openbank/domestic/domain/model/DelegatedSpendBinding.kt
+++ b/openbank-domestic-payment/src/main/kotlin/com/openbank/domestic/domain/model/DelegatedSpendBinding.kt
@@ -49,7 +49,7 @@ data class DelegatedSpendReservationSnapshot(
     init {
         require(schemaVersion == SCHEMA_VERSION) { "Unsupported reservation schema version: $schemaVersion" }
         require(aggregateType == AGGREGATE_TYPE) { "Unexpected aggregateType: $aggregateType" }
-        require(sourceService == SOURCE_SERVICE) { "Unexpected sourceService: $sourceService" }
+        require(sourceService == EXPECTED_PRODUCER_SERVICE) { "Unexpected sourceService: $sourceService" }
         require(resourceType == ACCOUNT_RESOURCE_TYPE) { "Domestic payment reservation must target ACCOUNT" }
         require(operationType == OPERATION_TYPE) { "Unexpected reservation operationType: $operationType" }
         require(amount > BigDecimal.ZERO) { "Reservation amount must be positive" }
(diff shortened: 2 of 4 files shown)


JiRaska:
Superseded by #8823, which landed the same two fixes an hour or so earlier and is now on `main`.

Both halves overlap exactly:

- **Gate budgets.** #8823 declared `budget_seconds` for the same five gates, taking the undeclared count back to the 159 the ratchet allows.
- **`sourceService` disambiguation.** #8823 renamed domestic-payment's accepted-producer constant to `EXPECTED_SOURCE_SERVICE` (this PR called it `EXPECTED_PRODUCER_SERVICE`) and resolved the kyc-service ambiguity by renaming the metrics-tag constant in `OrphanedPartyGauge` rather than the event constant. Different names, same property: one simple name, one value, per module.

Verified against current `main` rather than assumed: both `gate-observability-declarations` and `source-service-convention` pass there now.

Closing rather than merging, because taking this branch's side would rename constants `main` has already settled and gain nothing. No unique content is lost.

JiRaska:
**Blast radius, measured: this PR alone takes 14 of the 41 open non-draft PRs from red to fully green — 34% of the queue.**

I went looking for the cause of my own PR's three red checks, found `origin/main` red exactly as you describe, and then measured the whole queue rather than assuming. Posting the numbers because they are an argument for reviewing this one ahead of the rest.

## Method

For every open non-draft PR, read the **failed gate ids from each run's `gate-results-*.json` artifact**, not from the job log — the gates job also prints its meta-gate self-test fixtures, which fail on purpose and are indistinguishable from real failures under a `grep`.

## Failed gate ids across the 26 PRs red on a gates / `Validate manifests` check

```
10  gate-observability-declarations     <- this PR
 8  source-service-convention           <- this PR
 4  security-checklist-money-path       (#8757)
 3  (gates check red, artifact lists no failed gate)
 1  each: threat-model-coverage, threat-model-claims-resolve, openapi-version-not-taken,
         openapi-enum-domain-drift, license-header-consistency, incoming-dlq-wiring,
         idempotency-contract-parity, governance-script-unit-tests, credential-deadline-ratchet,
         api-contract-gate, agent-pr-guard, advisory-gate-registration
```

## Resolved per PR, since a PR can fail several gates

| | count | PRs |
|---|---|---|
| **fully green once this lands** | **14** | #8534 #8629 #8715 #8791 #8799 #8802 #8804 #8806 #8807 #8819 #8820 #8821 #8829 (+ #8805, mine) |
| those two gates plus a red of their own | 3 | #5986 #8011 #8833 |
| entirely their own problem | 4 | #7192 #8334 #8457 #8544 |
| gates red but the artifact lists no failed gate | 4 | #5990 #8606 #8665 #8676 |

#8831 is red on these two **and** on the `dependency-review` / `Trivy` trio, so it needs a re-run as well — that trio is intermittent (it returns "Dependency review is not supported on this repository", and the same checks pass on other PRs minutes apart).

## One thing worth a second look

The **4 PRs whose gates check is red while the artifact lists no failed gate** are a different shape from everything above, and I have not chased them. Either the job died before writing a verdict, or it wrote one that says nothing failed — and the second case is the more interesting one, since a gates job that is red while reporting no failed gate is exactly the state that reads as "unknown" but renders as "failed". Worth a look independently of this PR.

## Independent confirmation of your premise

Reproduced in a detached worktree at pristine `origin/main`, no local changes:

```
python3 .github/scripts/check-source-service-convention.py --root .
exit 1
FAIL openbank-domestic-payment: .../DelegatedSpendFinalizedAbsentEvent.kt:33 UNRESOLVED
FAIL openbank-kyc-service:      .../KycEvent.kt:69 UNRESOLVED
2 non-conforming write site(s), 0 stale baseline entr(ies) across 28 producer(s).
```

And the domestic-payment one is a false positive in the way #8827 describes — that file declares `const val SOURCE_SERVICE = "domestic-payment"` in its own companion object, which already *is* the module directory name without the prefix. The checker cannot resolve a constant declared in the same file.

JiRaska:
**Correction to my numbers above — my probe read one shard of three. The conclusion survives; the figures move.**

I said "read the failed gate ids from each run's `gate-results-*.json` artifact" and then took only the **first** artifact per run. A gates run publishes one artifact **per shard**:

```
gate-results-registry-kotlin-data
gate-results-lint-supplychain-security
gate-results-gitops-api
```

So every classification above was made on roughly a third of the evidence, and a failure recorded in another shard read as absent. Re-measured across **all** shards of **every** failed gates run per PR:

| gate | I reported | actual |
|---|---|---|
| `source-service-convention` | 8 | **18** |
| `gate-observability-declarations` | 10 | **16** |
| `security-checklist-money-path` | 4 | **6** |

Resolved per PR:

| | was | now | PRs |
|---|---|---|---|
| **fully green once this lands** | 14 | **11** | #8534 #8629 #8715 #8791 #8802 #8804 #8806 #8819 #8820 #8821 (+ #8805) |
| gate-red only on these two, but also red on a non-gate check | — | 2 | #8665 #8831 (both need the intermittent `dependency-review`/`Trivy` re-run too) |
| these two **plus** a gate of their own | 3 | 5 | #8011 #8457 #8799 #8807 #8833 |
| entirely their own gates | 4 | 7 | #5986 #5990 #7192 #8334 #8544 #8606 #8676 |
| no verdict recovered | 4 | 1 | #8829 |

**So: 11 of 41 open non-draft PRs — 27%, not 34% — go fully green on this PR alone**, with 2 more needing a re-run on top. Still the highest-leverage single merge in the queue by a wide margin; I just can't claim the larger number.

**Withdrawing the "worth a second look" paragraph.** I flagged 4 PRs as "gates red while the artifact lists no failed gate" and suggested the interesting reading — a gates job red while reporting nothing failed. That was my probe, not the gates. #8606 for one records `openapi-enum-domain-drift` plainly:

```
##[error]openapi enum drifts from the domain enum `FunnelStage` it serves: spec omits ['KYC_DOCUMENTS_REQUIRED']
55 gates  PASS=54  FAIL=1
```

It was in a shard I never opened. Only #8829 still yields no verdict, and one case is not a pattern.

Worth recording for anyone else reading these artifacts: **the verdict is the union across shards, and `head -1` silently answers about one of them.** It is the same failure the gates job's own design guards against in the other direction — the log carries self-test fixtures that fail on purpose, so a `grep` over-reports where a single-shard artifact read under-reports.

Issue #8828, opened 2026-09-05
openapi request-schema conformance: 122 findings across 17 services — the burn-down behind the 2026-09-30 date


JiRaska:
`openapi_request_schema_conformance` carries `target_enforce_date: 2026-09-30`. Until today the gate had **never compared a single request schema** — it was pointed at a path the build never wrote to (fixed in #8802). With the path corrected, here is the number it produces.

## The fleet number

Built all **52** services carrying an `openapi.yaml` (one Gradle invocation, `--continue`, `BUILD SUCCESSFUL in 25m 14s`, all 52 documents produced) and ran the checker over each generated document:

```
121 request schemas compared    122 findings    17 of 52 services affected
 88 parsed-not-published   a DTO property the spec omits
 34 published-not-parsed   a spec property no DTO has
```

**35 services are clean.**

| service | findings | published-not-parsed | parsed-not-published | schemas compared |
|---|---:|---:|---:|---:|
| `openbank-pid-service` | **27** | 10 | 17 | 11 |
| `openbank-interest-service` | **21** | 6 | 15 | 2 |
| `openbank-domestic-payment` | **13** | 2 | 11 | 1 |
| `openbank-dispute-service` | **12** | 3 | 9 | 8 |
| `openbank-standing-order-service` | **10** | 3 | 7 | 1 |
| `openbank-consent-service` | **8** | 2 | 6 | 2 |
| `openbank-sca-service` | **8** | 3 | 5 | 4 |
| `openbank-clearing-service` | **6** | 0 | 6 | 1 |
| `openbank-sepa-payment` | **6** | 2 | 4 | 1 |
| `openbank-balance-service` | **3** | 1 | 2 | 1 |
| `openbank-kyc-service` | **2** | 2 | 0 | 4 |
| `openbank-document-service` | **1** | 0 | 1 | 6 |
| `openbank-lending-service` | **1** | 0 | 1 | 8 |
| `openbank-party-service` | **1** | 0 | 1 | 10 |
| `openbank-sanctions-service` | **1** | 0 | 1 | 3 |
| `openbank-swift-service` | **1** | 0 | 1 | 1 |
| `openbank-transaction-service` | **1** | 0 | 1 | 3 |

## The half that hurts

`published-not-parsed` means the spec declares a field, the DTO does not have it: **a client generated from that document sends the field and the service silently drops it.** Five of those 34 are on the money path:

```
CreateSepaPaymentRequest.currencyCode
[...]

PR #8833, opened 2026-09-05
fix(sepa-payment): publish the request schemas the resources actually parse (also domestic-payment, balance)

JiRaska:
## Summary

Three money-path request schemas disagreed with the DTO the resource actually deserializes into — in **both** directions — and the gate that exists to catch exactly this could not see it, because it was reading a path the build never wrote to (#8802). First fixes off the #8828 burn-down.

## Type of change

- [x] fix (bug fix)
- [ ] perf (performance)
- [ ] docs
- [ ] chore (build / tooling)
- [ ] breaking change

## Linked issues

Refs #8828, #8802, #5895.

## Changes

| schema | published, on no DTO | omitted from the spec |
|---|---|---|
| `CreateSepaPaymentRequest` | `currencyCode`, `requestedExecutionDate` | `type`, `debtorIban`, `debtorName` (**all non-nullable**), `currency` |
| `CreateDomesticPaymentRequest` | `currencyCode`, `message` | 9 properties, 5 of them non-nullable — `debtorAccountNumber`, `debtorBankCode`, `debtorName`, `creditorName`, `priority` |
| `PlaceHoldRequest` | `expiresAt` (the lifetime is `ttlSeconds`) | `referenceId` (non-nullable), `ttlSeconds` |

**A client generated from any of these three documents could not create the payment or the hold at all** — it omits fields the DTO requires and sends fields the service silently drops. Same shape as #5895.

- Each schema's property names and optionality now mirror its DTO exactly, with a comment in place saying so and naming the DTO.
- `info.version` bumped per ADR-0048 — sepa `1.7.0 → 1.8.0`, domestic `1.6.0 → 1.7.0`, balance `1.9.0 → 1.10.0`.

## Verification

Ran the conformance checker against the **generated** documents from a real `quarkusBuild`:

```
before:  sepa 6 findings · domestic 13 · balance 3      (22)
after:   sepa 0 findings · domestic 0  · balance 0      (0)
```

`check-api-contract.py --base origin/main` accepts all three and reclassifies as expected:

```
::notice:: openbank-sepa-payment: spec-only PR — reclassifying 4 breaking document change(s)
           as a CORRECTION ... Requiring MINOR instead.
api-contract gate: openbank-sepa-payment: correction change, 1.7.0 -> 1.8.0 — OK
```

## Definition of Done

- [ ] Unit tests added/updated. — no code changed; the check that proves this is the conformance gate itself, now wired correctly in #8802
- [ ] `./gradlew detekt ktlintCheck koverVerify build` passes. — no Kotlin touched
- [x] No new lint warnings.

## Security checklist

- [x] Auth / crypto / payment code changed? → tag `security-review-required`. — payment **documents**, not payment code; no server behaviour changes

## Compliance impact

- [ ] PCI DSS
- [ ] DORA
- [ ] GDPR
- [x] PSD2 / SCA / RTS
- [ ] AML / 5AMLD
- [ ] CNB reporting
- [ ] None

The published SEPA and domestic-payment contracts were unusable as written. Correcting them is what a PSD2 API consumer needs in order to build against this service at all; no compensating control is needed, because the server side is byte-identical.

## Rollout / Rollback

- Deployment strategy: rolling — documents only, no image change required
- Rollback plan: revert the commit; the served behaviour is unaffected either way

## Reviewer notes

**The DTOs are the authority here, not the specs.** Every change moves the document toward the code, never the reverse — no server behaviour changes, which is exactly why `check-api-contract.py` reclassifies the removed properties as a `correction` (breaking to `oasdiff`, unbreakable in fact) and requires MINOR rather than a URL major.

Worth a second look: whether `requestedExecutionDate` on SEPA and `expiresAt` on the hold were *intended* features that were specified and never implemented. I removed them because nothing parses them today, but if either was meant to exist, the fix is a DTO change and a follow-up.


JiRaska:
Rebased onto current `main` and **version-sequenced behind #9670** — please merge #9670 first.

Both PRs correct `openbank-sepa-payment/src/main/resources/openapi.yaml` and both originally claimed `info.version` 1.8.0 against `main`'s 1.7.0. They touch different schemas (#9670: the status vocabularies; this one: `CreateSepaPaymentRequest`) so there is no content conflict, but one number cannot be published twice. This PR now takes **1.9.0**, recorded in the file's own header with the condition for dropping back to 1.8.0 if #9670 is closed rather than merged.

Also resolved while rebasing: `openbank-balance-service` `info.version` collided with `main`'s 1.9.1 — took 1.10.0, verified never published (`git log -S` on that file returns nothing).

Verified the balance-service half against the real DTO rather than trusting the PR title: `BalanceResource.PlaceHoldRequest` is `(amount, currency, reason, referenceId: String, ttlSeconds: Long? = null)` — so `referenceId` genuinely is required, `ttlSeconds` genuinely is optional int64, and the spec's `expiresAt` is a field the code does not have. A client generated from the old document omitted a required property and sent one that does not exist.

PR #8838, opened 2026-09-05
fix(pid): publish the request schemas the resources actually parse (14 services, 100 findings → 0)

JiRaska:
## Summary

The fleet tail of #8828: **100 request-schema findings across 14 services → 0.** Every one of these documents disagreed with the DTO its resource deserializes into, in both directions, invisibly, because the gate that checks this was reading a path the build never wrote to (#8802).

## Type of change

- [x] fix (bug fix)
- [ ] perf (performance)
- [ ] docs
- [ ] chore (build / tooling)
- [ ] breaking change

## Linked issues

Refs #8828, #8802, #5962.

## Changes

**Method — the DTO is the authority, and only for the property SET:**

- Property names come from the document Quarkus generates from the DTOs.
- A property that survives keeps its committed `description`, `example` and `enum` **verbatim** — no descriptions were lost (0 removed lines contain either).
- A property the DTO does not have is removed; one it has and the spec omitted is added.
- **`required` is only ever intersected** with the surviving names, never extended. The generated document is *not* evidence about optionality — SmallRye does not read Kotlin nullability and marks nearly everything required.
- Generated wrapper schemas the committed documents have never defined (`UUID`, `PaymentType`) are **inlined**, not `$ref`'d — verified 0 dangling references across all 14 files.

**Worst cases:**

| service | published, on no DTO | omitted |
|---|---|---|
| `pid` `CreatePartyRequest` | `legalName`, `tradingName`, `taxId`, `dateOfBirth`, `nationality` | `givenName`, `familyName`, `birthdate`, + 6 more |
| `interest` `InterestRateConfig` | `ratePercent`, `tierMin`, `tierMax`, `validFrom`, `validTo`, `calculationMethod` | 12 properties |
| `standing-order` | `amount`, `currencyCode`, `debtorAccountId` | `amountMinorUnits`, `currency`, `debitAccountId`, + 4 |
| `sca` `VerifyScaRequest` | `code` | `partyId`, `otp` |
| `consent`, `dispute`, `kyc`, `clearing`, `document`, `lending`, `party`, `sanctions`, `swift`, `transaction` | see #8828 | |

A client generated from any of them could not call the endpoint.

**Also:** dropped the `openbank-pid-service:EXPIRED,PENDING,REJECTED,VERIFIED` enum-drift baseline. It was baselined explicitly as *"needs a schema fix, not an enum fix"* — `UpdateKycRequest` had no `kycStatus` property at all, so the matcher had paired the spec enum with an unrelated openid4vp `Status` on two coincidental values. **This PR is that schema fix**, so the pairing is gone and `openapi-enum-domain-drift` reports the entry as stale; the gate fails in both directions, so it has to go.

## Verification

```
per-service conformance, against the generated documents from a real quarkusBuild:
  all 14 services → "N request schema(s) compared, 0 finding(s)"

check-api-contract.py --base origin/main → 14 × "OK (required >= MINOR)", 0 errors
run-gates --only openapi-enum-domain-drift / -version-not-taken / -route-conformance / -server-port → PASS
schema-name and path sets are byte-identical to origin/main in every file (checked structurally)
0 dangling $ref
```

## Definition of Done

- [ ] Unit tests added/updated. — no code changed; the conformance gate is the check, wired correctly in #8802
- [ ] `./gradlew detekt ktlintCheck koverVerify build` passes. — no Kotlin touched
- [x] No new lint warnings.

## Security checklist

- [x] Auth / crypto / payment code changed? → tag `security-review-required`. — **No.** No code changes in this PR at all; the `sca`, `clearing` and `transaction` **documents** change to describe what those endpoints already parse. Nothing is newly permitted or refused, so there is no behaviour for a security review to review.
- [x] PII path changed? → tag `gdpr-review-required`. — **No.** The `pid` and `party` documents describe PII fields more accurately; no field is newly collected, stored, logged or transmitted.
- [x] Cardholder data path changed? → tag `pci-review-required`. — **No.** No PAN, CVV or cardholder data appears in any spec this PR touches.

## Compliance impact

- [ ] PCI DSS
- [ ] DORA
- [x] GDPR
- [x] PSD2 / SCA / RTS
- [x] AML / 5AMLD
- [ ] CNB reporting
- [ ] None

`pid`/`party` (GDPR), `sca`/`consent`/`clearing` (PSD2), `kyc`/`sanctions` (AML) all publish contracts that now match what the service parses. No compensating control needed — the server side is byte-identical in every case; these documents were describing endpoints inaccurately, which is the risk being removed.

## Rollout / Rollback

- Deployment strategy: rolling — documents only, no image change required
- Rollback plan: revert; served behaviour is unaffected either way

## Reviewer notes

Two things worth a human eye:

1. **Removed properties may be unimplemented features rather than mistakes.** `interest`'s entire `InterestRateConfig` tier vocabulary (`tierMin`/`tierMax`/`validFrom`/`validTo`) reads like a design that was specified and never built. I removed it because nothing parses it today; if it was intended, the fix is a DTO change and a follow-up, not this.
2. **`required` was only narrowed, never widened**, so some genuinely non-nullable DTO fields are still published as optional. That is deliberate — widening `required` from the generated document would be wrong for the reason above. Narrowing it per service against the Kotlin is a good follow-up.


JiRaska:
`verification-metadata` here is not this PR's defect — it is a gap on `main`. A cold-cache resolve of this PR's changed modules pulls the Kotlin **2.4.20** compiler/daemon stack, and `gradle/verification-metadata.xml` pins only the release candidates (`2.4.20-RC`, `-RC2`, `-RC3`), never the final release. #8838 and #9012 report an identical list of 20 missing artifacts.

Fixed on main by #9098, which adds the ten components with checksums computed from the Maven Central bytes and corroborated against the published SHA-1 sidecars. Once that merges this goes green with no change here.

PR #8839, opened 2026-09-05
fix(domestic-payment,kyc): rename same-named consts colliding with sourceService

### LINKED ISSUE #8827: source-service-convention gate: 2 false positives (constant not reaching a sourceService write site) block domestic-payment and kyc-service
## What's broken

`check-source-service-convention.py` (gate `source-service-convention`) fails fleet-wide on
`main` right now with two findings that are, on inspection, both false positives — the
constant it flags is never assigned to an actual `sourceService` audit/event field.

```
FAIL  openbank-domestic-payment: .../DelegatedSpendFinalizedAbsentEvent.kt:33 UNRESOLVED —
  references SOURCE_SERVICE, which this module declares with 2 different values
  (delegation-service, domestic-payment) — ambiguous
FAIL  openbank-kyc-service: .../KycEvent.kt:69 UNRESOLVED — references SERVICE, which this
  module declares with 2 different values (kyc, kyc-service) — ambiguous
```

Blocks `gates (registry-kotlin-data)` on every PR touching either service (found while
triaging an unrelated PR, #8749).

## Why each is a false positive

**domestic-payment**: two `SOURCE_SERVICE` constants exist in the module.
`DelegatedSpendFinalizedAbsentEvent.SOURCE_SERVICE = "domestic-payment"` is the module's own
emitted value (correct — matches the module directory name) and is the only one actually
assigned to `sourceService = SOURCE_SERVICE`.
`DelegatedSpendBinding.SOURCE_SERVICE = "delegation-service"` sits next to a comment reading
"Producer-owned discriminator accepted by the projection; this service does not emit it" — it
documents *delegation-service's own* value for matching incoming state, and is never read into
this module's `sourceService` field (`grep -rn 'sourceService\s*=\s*SOURCE_SERVICE'` in the
module has exactly one hit, in the event class). The gate's per-module scan appears to treat
any `SOURCE_SERVICE`-named constant in the module as a candidate value, rather than only ones
actually assigned to a `sourceService` field.

**kyc-service**: same shape. `KycEvent.kt`'s `private const val SERVICE = "kyc-service"` is
assigned via `"sourceService" to SERVICE` — the real, correct producer value (matches the
module directory name, per the file's own comment citing #3994/#5256).
`OrphanedPartyGauge.kt`'s `const val SERVICE = "kyc"` is used only as a Micrometer tag
(`.tag("service", SERVICE)`), never as an audit `sourceService` value.

Both cases are the same root shape: a same-named constant used for something other than the
`sourceService` write site trips the "declares 2 different values" ambiguity check.

## Evidence
- Failing run: https://github.com/JiRaska/open-bank-oss/actions/runs/33956847732/job/101281605764
- `openbank-domestic-payment/src/main/kotlin/com/openbank/domestic/domain/model/DelegatedSpendBinding.kt:109`
- `openbank-domestic-payment/src/main/kotlin/com/openbank/domestic/domain/event/DelegatedSpendFinalizedAbsentEvent.kt:38`
- `openbank-kyc-service/src/main/kotlin/com/openbank/kyc/domain/model/KycEvent.kt:98` (correct)
- `openbank-kyc-service/src/main/kotlin/com/openbank/kyc/infrastructure/observability/OrphanedPartyGauge.kt:176` (unrelated metrics tag)
- Script: `.github/scripts/check-source-service-convention.py`

I'm filing rather than patching the detector directly — this file's own history (CLAUDE.md)
documents two earlier false-positive regressions in this exact script (12 then ~40 false
findings) from attempts to derive "does this look like a producer" heuristically, so a
narrow, carefully-tested fix (only count a constant if it reaches an actual `sourceService =`
or `"sourceService" to` write site) deserves review rather than a quick patch under time
pressure.

JiRaska:
## Summary
`gates (registry-kotlin-data)` fails fleet-wide on every PR touching `openbank-domestic-payment`
or `openbank-kyc-service`: `source-service-convention` reports both as UNRESOLVED/ambiguous.

That's the gate's designed-in caution working correctly, not a bug in the check — both modules
genuinely declare two constants under the same simple name, and the gate deliberately refuses
to guess which one a write site's reference means (see the script's own header, which cites the
identical fx-service `SERVICE` vs `SOURCE_SERVICE` shape as the reason this ambiguity path
exists). The fix is in the modules, not the gate: rename the constant that ISN'T the real
`sourceService` write site so the collision goes away.

- **domestic-payment**: `DelegatedSpendBinding.SOURCE_SERVICE = "delegation-service"` documents
  the *upstream* producer's value for a `require(...)` validation, not this module's own
  emission — `DelegatedSpendFinalizedAbsentEvent.SOURCE_SERVICE = "domestic-payment"` is the
  real write site. Renamed the former to `EXPECTED_SOURCE_SERVICE`.
- **kyc-service**: `OrphanedPartyGauge.SERVICE = "kyc"` is an unrelated Micrometer tag, not the
  audit `sourceService` — `KycEvent.kt`'s `SERVICE = "kyc-service"` (assigned via
  `"sourceService" to SERVICE`) is the real write site. Renamed the gauge's constant to
  `METRIC_SERVICE_TAG`.

No values changed, no external callers of either renamed constant.

(Originally filed as #8827 diagnosing this as a gate false positive — on closer reading of the
gate's own design comments this is intentional ambiguity detection, so closing that issue with
this PR instead.)

## Linked issues
Closes #8827

## Test plan
- [x] `python3 .github/scripts/check-source-service-convention.py --root . --self-test` — `25 passed, 0 failed`
- [x] `python3 .github/scripts/check-source-service-convention.py --root .` — `OK — 28 producer(s) checked`
- [x] `./gradlew :openbank-kyc-service:compileKotlin :openbank-domestic-payment:compileKotlin` — `BUILD SUCCESSFUL`

### DIFF
--- a/openbank-domestic-payment/src/main/kotlin/com/openbank/domestic/domain/model/DelegatedSpendBinding.kt
+++ b/openbank-domestic-payment/src/main/kotlin/com/openbank/domestic/domain/model/DelegatedSpendBinding.kt
@@ -49,7 +49,7 @@ data class DelegatedSpendReservationSnapshot(
     init {
         require(schemaVersion == SCHEMA_VERSION) { "Unsupported reservation schema version: $schemaVersion" }
         require(aggregateType == AGGREGATE_TYPE) { "Unexpected aggregateType: $aggregateType" }
-        require(sourceService == SOURCE_SERVICE) { "Unexpected sourceService: $sourceService" }
+        require(sourceService == EXPECTED_SOURCE_SERVICE) { "Unexpected sourceService: $sourceService" }
         require(resourceType == ACCOUNT_RESOURCE_TYPE) { "Domestic payment reservation must target ACCOUNT" }
         require(operationType == OPERATION_TYPE) { "Unexpected reservation operationType: $operationType" }
         require(amount > BigDecimal.ZERO) { "Reservation amount must be positive" }
@@ -106,7 +106,7 @@ data class DelegatedSpendReservationSnapshot(
         /** Producer-owned discriminator accepted by the projection; this service does not emit it. */
         const val SOURCE_EVENT_TYPE = "DelegationSpendReservationStateChanged"
         const val AGGREGATE_TYPE = "DelegationSpendReservation"
-        const val SOURCE_SERVICE = "delegation-service"
+        const val EXPECTED_SOURCE_SERVICE = "delegation-service"
         const val ACCOUNT_RESOURCE_TYPE = "ACCOUNT"
         const val OPERATION_TYPE = "DOMESTIC_PAYMENT"
         const val SCHEMA_VERSION = 1L
--- a/openbank-kyc-service/src/main/kotlin/com/openbank/kyc/infrastructure/observability/OrphanedPartyGauge.kt
+++ b/openbank-kyc-service/src/main/kotlin/com/openbank/kyc/infrastructure/observability/OrphanedPartyGauge.kt
@@ -117,7 +117,7 @@ class OrphanedPartyGauge(
 
     private fun gauge(r: MeterRegistry, name: String, holder: AtomicLong) {
         Gauge.builder(name, holder) { it.get().toDouble() }
-            .tag("service", SERVICE)
+            .tag("service", METRIC_SERVICE_TAG)
             .strongReference(true)
             .register(r)
     }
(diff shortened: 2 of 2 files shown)


JiRaska:
Superseded — #8823 landed the identical fix (both renames + the same gates.yaml budget declarations) first. Verified clean on current `main`: both `check-source-service-convention.py` and `check-gate-observability-declarations.py` pass. Closing as duplicate.

PR #8837, opened 2026-09-05
feat(card-processing): the card money path — authorisation, hold, clearing, ledger posting (ADR-0283 phase 1)

JiRaska:
## Summary

`openbank-card-processing-service`: the card money path. It measures what a card has already spent, asks card-issuance for the decision, holds the funds an approval implies, applies clearing presentments, releases what is never presented, and posts cleared spend to the books on the `CARD` rail. ADR-0283 phase 1.

Stacked on #8819 (ADR-0283 itself), which is merged into this branch so its CI is independent.

## Type of change

- [x] feat (new functionality)
- [ ] fix (bug fix)
- [ ] refactor (no behavior change)
- [ ] perf (performance)
- [ ] docs
- [ ] chore (build / tooling)
- [ ] security (private until disclosed)
- [ ] breaking change

## Linked issues

Closes #8809. Refs #8808.

## The defect this closes

Measured on `origin/main` before writing any code:

- `POST /api/v1/cards/{id}/authorizations` decides approve or decline and **has no caller** anywhere in the fleet. The channel toggles a customer sets are enforced by a function nothing invokes.
- That endpoint takes the spend figures **as arguments**, so with no caller they were never anything. A limit evaluated against a number the requester supplies is not a limit.
- `openbank.cards.events` reaches campaign-service and audit-service and **never the ledger**. Every other instrument posts; card spend could not reach the books at all.

## Changes

**The service.** Hexagonal, money-path from its first commit. The authorisation aggregate with a derived hold amount, a pure transition policy whose refusals are values rather than exceptions, spend counters computed in the database over the authorisation rows, a transactional outbox with the atomic cross-pod claim, a hold-expiry sweep, and four events.

**Where each decision lives.** Whether to approve is card-issuance's (not re-implemented here: two copies of a control diverge and the customer sees the wrong one). How much has been spent is this service's. Whether money moves is the ledger's, and only on a clearing.

**Three properties worth reviewing specifically:**

- **Failing closed.** If card-issuance cannot be reached the authorisation is declined, under its own reason `ISSUER_UNAVAILABLE` — never one borrowed from the policy, because in a dispute "the issuer was down" and "you turned gambling off" are different facts. This is the opposite of VoP's deliberate fail-open, and the threat model says why.
[...]

Issue #8810, opened 2026-09-05
Visa Developer + Mastercard Developers sandbox adapters behind scheme capability ports (ADR-0283 phase 2)


JiRaska:
Refs #8808 (phase 2). ADR-0283.

## Scope

Capability ports in `openbank-libs-domain` (`com.openbank.libs.cards.scheme`) + one adapter module per network, each with a **sandbox simulator** (default) and a **real sandbox client** (opt-in by config, credentials from OpenBao, never in the repo):

| Port | Visa Developer | Mastercard Developers |
|---|---|---|
| `BinLookupPort` / `MerchantDataPort` | Visa BIN Attributes, Merchant Search / Locator | BIN Lookup, Merchant Identifier, Places |
| `TokenisationPort` | Visa Token Service (VTS) | MDES |
| `PushPaymentPort` | Visa Direct (OCT) | Mastercard Send |
| `DisputePort` | Visa Resolve Online (VROL) | Mastercom |
| `AccountUpdaterPort` | Visa Account Updater | Automatic Billing Updater |
| `NetworkControlsPort` | Visa Transaction Controls | In Control |
| `FxRatePort` | Visa FX Rates API | Currency Conversion |

Order: BIN/merchant lookup first (cheap, read-only, is the missing **writer** for the starved merchant catalogue in #8573), then tokenisation, push, disputes.

## Non-negotiables

- Auth per portal: Visa = mTLS + API key (+ MLE where the product requires it); Mastercard = OAuth 1.0a request signing with the portal-issued key. Both live behind the adapter; no service outside the adapter module sees them.
- Every adapter carries a Pact consumer test with a **literal** path (the `CLAUDE.md` rule), a recorded-fixture replay from the real sandbox, and a contract test that the simulator and the real client return the same shape.
- Capability registry (`openbank-libs/governance/card-capabilities.yaml`): which network offers which capability, at what certification cost — the source the knowledge base and the control plane are generated from (rule 6: derived data is never hand-edited).
- DORA: each real adapter registers the provider in the ICT third-party register.

Issue #8573, opened 2026-09-03
Transaction categorisation is built but starved: the merchant catalogue has ~30 rows and no writer, and non-card spend can never be categorised


JiRaska:
## Correcting the premise

I have twice written that transaction categorisation is "a backend feature with no engine". **That is wrong.** The pipeline is built end to end and built carefully:

- `MerchantDescriptor.normalise` turns a free-text acquirer descriptor into a lookup key, stripping case, punctuation, legal forms (`AS`, `SRO`, `GMBH`…), trailing town + district (`PRAHA 4`) and country tokens. Matching is **exact on the normalised key, deliberately** — its KDoc explains that fuzzy matching would let `ALZA VETERINARY` collect Alza.cz's logo and coordinates, and a wrong location presented as fact is worse than none.
- `MerchantCatalogRepository.findByDescriptors` does **one catalogue query per page**, not per row, so enrichment stays off the per-row critical path.
- `GET /api/v1/transactions` enriches every page; customer-edge forwards `merchant` and `merchantCategory` and **declares both** in its OpenAPI (ADR-0103 D5).
- The app consumes it in `TransactionApi`/`Transaction` and renders it in `TxDetailScreen`, `CardTxDetailSheet`, `PfmScreen` and `TxMapScreen`.

Nothing here needs building. What it needs is **data**, and there is no way to supply it.

### 1. The catalogue cannot grow

`merchant_catalog` is seeded by a single `INSERT` in `V16__create_merchant_catalog.sql` — about 30 Czech merchants — and **there is no writer anywhere in the codebase**. No ingestion job, no admin endpoint, no operator screen (admin-ui has ~20 sections; none is merchants). Adding one merchant today means a database migration and a deploy.

The edge's own schema already concedes the consequence: `merchant` is absent *"when the acquirer descriptor has no catalogue match, **which is most transactions**"*.

### 2. Non-card spend has no category at all, and cannot get one from `type`

`merchantCategory` is MCC/catalogue-derived, so it only ever exists for card spend. Everything else — transfers, direct debits, standing orders, SIPO — has no category and lands in PFM's "Other" bucket. For a Czech retail customer that is rent and utilities, i.e. a large share of outgoings.

[...]

PR #8855, opened 2026-09-05
feat(card-processing): Visa and Mastercard BIN adapters behind the capability port (ADR-0283 phase 2)

JiRaska:
## Summary

The first card capability with real vendor bindings: Visa and Mastercard BIN lookup behind `BinLookupPort`, plus the in-repo simulator that is the default. BIN lookup goes first because it is read-only, carries no cardholder data, and both networks expose it in a free developer sandbox.

Stacked on #8837 (the service), #8840 (the ports) and #8842 (the Card Center page), all merged into this branch so its CI stands alone.

## Type of change

- [ ] fix (bug fix)
- [ ] perf (performance)
- [ ] docs
- [ ] chore (build / tooling)
- [ ] breaking change

## Linked issues

Refs #8810, #8808. Refs #8573 — BIN and merchant reference data is the capability that feeds the starved merchant catalogue.

## Three bindings, selected by configuration

`openbank.card-processing.scheme.bin-lookup` is `simulator`, `visa` or `mastercard`. That setting is what makes "the bank chooses the network" a configuration change rather than a code change.

- **Simulator** — always available, the default, and a first-class binding rather than a test double. A platform whose only binding needs credentials nobody has is a platform where the ports are never exercised, which is exactly how the issuer's authorisation decision came to have no caller for a month. It answers from the networks' own published sandbox test ranges and **refuses an unknown BIN** instead of inventing an issuer, so the miss branch is exercised rather than theoretical.
- **Visa** — mutual TLS plus an API key. The TLS half is Quarkus configuration, so there is no crypto in the adapter to get wrong.
- **Mastercard** — OAuth 1.0a RSA-SHA256 request signing, which *is* code: the `Authorization` header is a signature over the request's own method, URL and parameters, so no filter can attach it.

## The signer, and why it is tested the way it is

Four properties are easy to get wrong and each one produces a header that looks valid and is always rejected — a 401 that reads like a credential problem when it is a construction problem. Each has its own test:

| Property | The mistake it prevents |
|---|---|
| RFC 3986 percent-encoding | `URLEncoder` gives `+` for a space, leaves `*`, escapes `~` |
| Parameters sorted **after** encoding | Sorting raw values orders by a different alphabet |
| The parameter block encoded **twice** | Encoding once passes any test whose values contain nothing that encodes |
| URL normalised: no query, no default port, lower-case scheme and host | The server rebuilds it that way |

The signature is then verified with the public key, which is what the server does. A golden-string test would prove only that the code still does what it did, not that it was ever right.

## The router does not fall back

A configured vendor binding that cannot answer returns **its own** failure. A silent fallback to the simulator would make an unconfigured integration indistinguishable from a working one at every call site, in every log line and in every stored row — the same failure shape as a skipped delivery counted as a success.

`SchemeFailure` earns its five values here: `NOT_BOUND` (nobody configured a credential — permanent until someone does) is never merged with `UNAVAILABLE` (the network is down — clears on its own), and `UNAUTHENTICATED` will not fix itself by retrying.

## A caller, not just a port

`MerchantDataPort` is wired into the authorisation path, so this PR does not add ports with no caller — the defect ADR-0283 exists to fix. It **fails open**: a merchant name is what a customer reads on a statement, not a control, so an unanswered lookup keeps the raw descriptor and the authorisation proceeds. That is the deliberate opposite of the issuer call, which fails closed, and the difference is that one authorises and the other only labels.

`BinLookupPort`'s first consumer is the reference-data writer in #8573, not this PR. The adapter ships now because it is the capability both networks expose for free, and it is exercised end to end by the router tests.

## Definition of Done

- [x] Hexagonal layering preserved — the ports are in libs-domain, the adapters in infrastructure.
- [x] Unit tests added — 63 total in the module, 0 failures, 0 skipped. New: simulator behaviour, signer construction, per-adapter failure classification, router selection.
- [x] Integration tests — the existing real-database ITs still pass with the merchant lookup on the authorisation path.
- [ ] Contract tests — n/a. There is no Visa- or Mastercard-side replay in this repository, which is why both vendor paths are literals reviewed as text and every failure branch is unit-tested.
- [x] `ktlintCheck`, `detekt`, `test`, `koverVerify` — each run with its exit code measured directly.
- [x] No new lint warnings.
- [ ] OpenAPI spec — n/a, no new REST surface.
- [ ] Flyway migration — n/a.
- [ ] Event schema — n/a.
- [x] Docs updated — the capability registry's `bindings` column now lists the two vendor adapters, the generated matrix follows, and the Card Center banner derives that sentence from the data instead of asserting it.

## Security checklist

- [x] No secrets, tokens, passwords, or PII in code, config, tests, logs. **No credential has a default and none appears in any `application.yaml`**: every one is an empty environment expansion supplied from OpenBao. The signer test generates its own throwaway key pair.
- [x] No new `@SuppressWarnings` / `as Any`. Two `@Suppress("TooGenericExceptionCaught")`, one per vendor adapter, each with the reason at the site: a vendor lookup must never be able to fail the money path it enriches.
- [x] No new third-party dependency. The signer uses the JDK's own `Signature` and `KeyFactory`.
- [x] Auth / crypto / payment code changed? **Yes** — this PR contains request-signing crypto. It signs, it never verifies, and it holds no key material: the private key arrives as a config value and the producer declines to build a signer if it is absent or unparseable.
- [x] PII path changed? No. A BIN is an issuer range, not a card number, and no port signature accepts a PAN.

## Compliance impact

- [x] PCI DSS
- [x] DORA
- [ ] GDPR
- [ ] PSD2 / SCA / RTS
- [ ] AML / 5AMLD
- [ ] CNB reporting
- [ ] None

PCI DSS: no scope change. A BIN is the issuer range, never a PAN, and the port signatures are what keep it that way. DORA: two ICT third-party providers now have adapter code. Neither is bound in any environment, so neither enters the register yet — that happens in the change that configures a credential, and the PR doing so should say so.

## Rollout / Rollback

- Deployment strategy: rolling. Behaviour is unchanged in every environment, because no credential is configured anywhere and the default binding is the simulator.
- Rollback plan: revert, or set `bin-lookup=simulator` — which is already the default.

## Reviewer notes

The judgement call worth challenging is the no-fallback rule in the router. It means a bank that configures Visa and then lets the credential lapse gets `NOT_BOUND` on every lookup rather than a simulated answer. I think that is right — a wrong-but-plausible answer is worse than no answer, and a merchant name is not worth pretending about — but it is a decision, not a fact.

Second, the two vendor paths are literals nothing in this repository can adjudicate, because there is no Visa or Mastercard replay here. That is stated in the client KDoc rather than glossed. If you know the current sandbox routes differ, that is the review this PR most needs.


JiRaska:
## `dependency-review` here is the fleet-wide submission break, not this PR

Measured rather than assumed. Across the last completed runs of that workflow it fails on exactly the two branches that **add a new Gradle module** — `feat/card-processing-service` and this one — and succeeds on every other card branch, which add no module:

```
failure  feat/card-scheme-sandbox-adapters
failure  feat/card-processing-service
success  feat/card-scheme-bulletin-agent
success  feat/card-center-admin-ui
success  feat/card-scheme-capability-ports
success  docs/card-platform-phase0
```

The job's own error is `Dependency review is not supported on this repository. Please ensure that Dependency graph is enabled` — which reads like a repository setting, and is not: the graph is enabled, which is why the other branches pass. What is broken is the **submission** that populates it. #8781 is open for exactly that: `Dependency submission` has failed on `main` and on every PR since setup-gradle v6 moved to `github-dependency-graph-gradle-plugin:1.4.2`, on artifact verification.

With no submitted snapshot, a PR that introduces a module has nothing to compare against and the action errors out. A PR that only changes existing files still resolves against the last good snapshot.

So this check goes green here when #8781 lands, and nothing in this PR can move it. I have not worked around it: the fix belongs to that PR, and papering over it here would hide a fleet-wide supply-chain gap behind one green tick.

PR #8825, opened 2026-09-05
ci(gates): resolve a sourceService constant in its own file first

JiRaska:
## Summary

`source-service-convention` is failing on `origin/main`, so every open pull request is red on it. The two sites it names are correct code: the resolver looks a bare constant up across the whole module, finds the same simple name declared twice, and gives up. This teaches it to resolve the reference in its own file first, which is what Kotlin does.

## Type of change

- [ ] feat (new functionality)
- [x] fix (bug fix)
- [ ] refactor (no behavior change)
- [ ] perf (performance)
- [ ] docs
- [ ] chore (build / tooling)
- [ ] security (private until disclosed)
- [ ] breaking change

## Linked issues

Refs #5902 (the sourceService convention this gate enforces), Refs #8791 (found while an unrelated pull request was red on it)

## What was actually wrong

A bare identifier in Kotlin resolves to the declaration in its own file before anything else, and a `private const val` is file-private besides, so a same-named constant in a sibling file cannot be what the line reads.

| module | constant | one value | the other |
|---|---|---|---|
| `openbank-kyc-service` | `SERVICE` | `"kyc"` — a metric label in `OrphanedPartyGauge` | `"kyc-service"` — the event source in `KycEvent` |
| `openbank-domestic-payment` | `SOURCE_SERVICE` | `"delegation-service"` — the **inbound** producer `DelegatedSpendBinding` validates | `"domestic-payment"` — its own outbound event |

In both, the file that emits the value declares the correct one beside it, and the other constant means something else entirely. Neither is a naming defect. A gate that cries wolf about correct code is worth less than nothing, so this fixes the resolver rather than renaming a constant in a money-path service or declaring the gate blind through its acknowledgement list.

## The preference cannot launder a wrong value

That is the obvious objection, so the self-test proves it rather than the comment asserting it. Three fixtures, three verdicts:

| fixture | shape | verdict |
|---|---|---|
| `openbank-samefile-service` | use beside the **right** value, wrong one elsewhere in the module | allowed |
| `openbank-samefile-bad-service` | use beside the **wrong** value, right one elsewhere | flagged |
| `openbank-ambiguous-service` | use in a file declaring **neither** | flagged |

The third fixture was edited: its use moved into a third file. Left beside one of the declarations it would have stopped testing ambiguity the moment the resolver improved, and quietly become a duplicate of the first case — a test that passes for a reason it was not written for.

## Verification

```
python3 .github/scripts/check-source-service-convention.py --root .
[...]

PR #8856, opened 2026-09-05
fix(ci): remove five duplicate budget_seconds keys that turn main red

JiRaska:
## Summary

`yamllint` fails on a clean `origin/main` checkout: five gates declare `budget_seconds` twice, `key-duplicates` fires, and every open PR inherits the failure through its merge commit. This removes the duplicates.

## Type of change

- [x] fix (bug fix)
- [ ] perf (performance)
- [ ] docs
- [ ] chore (build / tooling)
- [ ] breaking change

## Linked issues

Found while watching CI on the ADR-0283 stack (#8819, #8840), both of which went red on it without touching the file.

## What happened

Two changes declared a budget for the same gates independently. That is the duplicate-mapping-key shape this repo already documents: SnakeYAML and PyYAML keep only the **last** of a repeated key and drop the rest silently, so the file reads as saying one thing and parses as another.

## Which one was removed, and why it changes nothing

The **first** of each pair, never the second. PyYAML already keeps the last, so the effective budget today is the second value — this removes the ambiguity, not the setting. The surviving values are also the ones carrying a measurement.

| Gate | Surviving budget | Basis |
|---|---|---|
| alert-rca-ledger-guards | 30 s | 4.1 s measured under load |
| standing-critical-digest-guards | 30 s | 4.2 s measured under load |
| agent-bounded-write-surface | 30 s | 5.0 s measured under load |
| source-service-convention | 60 s | 14.1 s measured under load |
| ruleset-bypass-actors | 60 s | network-bound GitHub API read |

## Definition of Done

- [x] Hexagonal layering preserved — n/a, one manifest file.
- [ ] Unit tests — n/a. The gates below are the test, and they were red before and green after.
- [ ] Integration tests — n/a.
- [ ] Contract tests — n/a.
- [x] Verified after the change: `yamllint`, `duplicate-yaml-key-guard`, `gate-observability-declarations`, `gate-subject-floor`, `gate-lifecycle-metadata`, `gate-selftest-declaration`, `gate-script-registration` all pass. Verified before it too — `yamllint` fails on a detached checkout of `origin/main` with no local changes.
- [x] No new lint warnings.
- [ ] OpenAPI / Flyway / event schema — n/a.
- [x] Docs updated — the removed lines were duplicates; every surviving comment is the one with the measurement in it.

## Security checklist

- [x] No new `@SuppressWarnings` / `as Any` / `@Suppress`.
- [x] Auth / crypto / payment code changed? No. One of the five gates is `ruleset-bypass-actors`, which watches who may bypass main-protection; its budget is unchanged in effect and its logic is untouched.
- [x] PII path changed? No.

## Compliance impact

- [ ] PCI DSS
- [ ] DORA
- [ ] GDPR
- [ ] PSD2 / SCA / RTS
- [ ] AML / 5AMLD
- [ ] CNB reporting
- [x] None

## Rollout / Rollback

- Deployment strategy: none. A CI manifest, no runtime effect.
- Rollback plan: revert; `yamllint` returns to failing on main, which is the state before this PR.

## Reviewer notes

Worth confirming you agree the first of each pair is the right one to drop. My reasoning is that the parser already ignores it, so removing it cannot change behaviour, while removing the second would silently shorten four budgets and one network-bound one. If you would rather keep the shorter values, that is a deliberate tightening and belongs in its own change with its own measurements.

### DIFF
--- a/.github/gates/gates.yaml
+++ b/.github/gates/gates.yaml
@@ -200,7 +200,6 @@ gates:
   # it deliberately does not check REACHABILITY — a script invoked from a workflow that
   # never runs still counts as wired. Registering here is what puts them on a PR lane.
   - id: alert-rca-ledger-guards
-    budget_seconds: 15   # one script self-test run
     selftest_exempt: "is-a-test-suite — the run: IS alert-rca-ledger.py --self-test"
     name: "alert-RCA ledger guards (MTTR granularity, recurrence, observation coverage)"
     group: lint
@@ -212,7 +211,6 @@ gates:
       python3 .github/scripts/alert-rca-ledger.py --self-test
 
   - id: standing-critical-digest-guards
-    budget_seconds: 15   # one script self-test run
     selftest_exempt: "is-a-test-suite — the run: IS standing-critical-digest.py --self-test"
     name: "standing-critical digest guards (severity filter, observation envelope)"
     group: lint
@@ -1923,7 +1921,6 @@ gates:
   # 2026-08-20: a money-path prefix, an added `PUT /pulls/{n}/merge`, and a deleted bound each make
   # it exit 1; removing either half of `evaluate` makes the self-test go red.
   - id: agent-bounded-write-surface
-    budget_seconds: 20   # rules.yaml + agent charter scan
     min_subjects: 20   # the money_path_services entries the bound is checked against; 23 today
                         # (2026-08-20). A gate that resolves an empty list would pass everything,
                         # so the floor moves with the list rather than tracking it exactly.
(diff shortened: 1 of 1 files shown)


JiRaska:
Closing as absorbed: #8831 removed the same five duplicate `budget_seconds` keys and merged, so this branch's content is now identical to `main`.

Verified by content rather than by the diff header, which is computed against a stale merge-base and can still show a change for a fully absorbed branch: `git diff origin/main origin/fix/gates-duplicate-budget-keys` is empty.

Confirmed live on `main`: no gate carries a duplicated `budget_seconds`, and the real `yamllint` gate reports `ok` there.

For the record on cause, since it was mine: #8823 added budgets to those five gates, my #8825 added them again, and `yaml.safe_load` silently keeps the last of a duplicated key — so the file parsed, `gate-observability-declarations` saw a budget and passed, and only `yamllint` could see the duplication. Two pull requests then raced to fix it; this is the one that arrived second.

Issue #8811, opened 2026-09-05
Card Center: one control plane for products, networks, disputes, tokens and the capability matrix (ADR-0283 phase 3)


JiRaska:
Refs #8808 (phase 3). ADR-0283.

## Scope — "manage it from one place"

**Card Center** in `openbank-admin-ui` (extends `/cards`), persona-aware (ADR-0229), backed by MCP providers so the same data serves the agent channel (ADR-0228 pattern):

1. **Products & networks**: product-catalog gains card product attributes (network, tier, fee schedule, BIN range); an operator selects the network **per product**; the choice is what switches the adapter (phase 2 port binding by product).
2. **Capability matrix**: generated from `card-capabilities.yaml` — per capability, what each network offers, what it costs, what certification it requires. This is the "educate the bank" surface.
3. **Disputes desk**: chargeback state machine per scheme reason-code family, statutory and scheme deadlines, evidence chain (ADR-0117 integration), four-eyes on `dispute.decide`.
4. **Token lifecycle**: list/suspend/resume/delete tokens per card, wallet provisioning status.
5. **Processing & scheme ops**: authorisation stream, decline reasons (`DeclineReason` explained), holds ageing, clearing exceptions, adapter health (simulator vs real sandbox), scheme reporting placeholders.
6. **Card Knowledge Base** in Service Docs: four-party model, auth/clearing/settlement, tokenisation, EMV, 3DS, scheme rules, Visa vs Mastercard comparison — generated from the registry, mermaid parsed by the `mermaid-parses` gate.

RBAC guard tests (`cards-rbac.guard.test.ts` pattern), vitest via `npm test`.

PR #8858, opened 2026-09-05
feat(card-processing): tokenisation and dispute bindings (ADR-0283 phase 2)

JiRaska:
## Summary

Working bindings for the two remaining read-write card capabilities: tokenisation and disputes. Every capability in the registry that has a port now has a binding something can actually execute.

The vendor half is deliberately absent, and that is the part of this PR worth reviewing.

Stacked on #8855, merged in so its CI stands alone.

## Type of change

- [ ] fix (bug fix)
- [ ] perf (performance)
- [ ] docs
- [ ] chore (build / tooling)
- [ ] breaking change

## Linked issues

Refs #8810, #8808.

## Why there is no Visa or Mastercard client here

VTS, MDES, VROL and Mastercom are **contract-only**. The capability registry records `availability: contract` for all four, which means no free developer sandbox exists to build or verify an adapter against.

A client written from documentation nobody on this side can open would be code that can never be executed, never verified, and green in every test — the exact shape of the finrep call to a ledger route that had never existed (#2269). The BIN adapters in #8855 exist precisely because that capability *does* have a free sandbox.

So the vendor names stay **selectable and honest**. Choosing `visa` or `mastercard` returns `NOT_BOUND` with a reason naming the contract, attributed to the scheme the caller asked for rather than to the simulator — a `NOT_BOUND` blamed on the simulator would read as the simulator being broken. When a contract exists, the adapter lands there and the branch goes away.

If you would rather have speculative clients now, that is a reasonable different call and I will write them; I would want the PR to say plainly that no test can reach them.

## What the simulators do, and what they refuse to invent

**Tokenisation** — provision, list, suspend, resume, delete, over an in-memory store.

- Stateful on purpose: the port's contract is a *sequence*, and a stateless stub cannot exercise it at all.
- In memory on purpose: the network owns the token vault in a real deployment and this platform stores no token either. A simulator that persisted would teach a durability the real binding does not offer.
- `DELETED` is terminal, because it is terminal in every scheme's lifecycle. Allowing a resume would let a caller believe it had restored a wallet credential the network has already destroyed.
- An empty list is an **answer**, not a failure — otherwise a caller treats a normal state as an error and the "no wallets yet" branch never renders.
- Nothing it mints is PAN-shaped, and a test asserts that. The ports were designed so no signature accepts a PAN; the simulator must not smuggle one in through the back door.

**Disputes** — open, submit evidence, read status.

- The caller's reason code is carried through **unchanged**. The vocabulary is per-network, versioned and published under contract; inventing or normalising one would teach a caller a code the network rejects, months later, at a real chargeback deadline.
- The respond-by date is a flat 30 days and says in the code that it is a placeholder. Real windows depend on the reason code, the region and the calendar, and anything computed here would look authoritative and be wrong.

## On callers

These two ports have no production consumer yet, and I am not inventing one. Their real consumers are the Card Center token screen and the disputes desk, which are the rest of phase 3 (#8811). What this PR does give them is a binding that runs and a test suite that exercises every branch — which is the difference between a port and the decorative kind this ADR exists to stop.

## Definition of Done

- [x] Hexagonal layering preserved — ports in libs-domain, bindings in infrastructure.
- [x] Unit tests added — 78 in the module now, 0 failures, 0 skipped. New: the token lifecycle including the terminal-delete and empty-list cases, the dispute lifecycle including both refusals, and the contract-required routing for both.
- [x] Integration tests — the existing real-database ITs still pass.
- [ ] Contract tests — n/a, no vendor call exists to pact.
- [x] `ktlintCheck`, `detekt`, `test`, `koverVerify` — each exit code measured directly.
- [x] No new lint warnings.
- [ ] OpenAPI / Flyway / event schema — n/a.
- [x] Docs updated — the registry now states, per capability, that simulator-only is a decision rather than a backlog item, and the generated matrix follows.

## Security checklist

- [x] No secrets, tokens, passwords, or PII in code, config, tests, logs. The simulators mint fixed, obviously-simulated values.
- [x] No new `@SuppressWarnings` / `as Any` / `@Suppress`.
- [x] Auth / crypto / payment code changed? No. No credential is read and no network call is made by anything in this PR.
- [x] PII path changed? No.
- [x] Cardholder data path changed? **No, and a test enforces it** — the token simulator's output is asserted not to be PAN-shaped.

## Compliance impact

- [x] PCI DSS
- [ ] DORA
- [ ] GDPR
- [ ] PSD2 / SCA / RTS
- [ ] AML / 5AMLD
- [ ] CNB reporting
- [ ] None

PCI DSS: no scope change. Network tokenisation is the mechanism that *reduces* PAN exposure, and nothing here handles a PAN. DORA is unticked because no external provider is bound — that is the whole point of this PR's central decision.

## Rollout / Rollback

- Deployment strategy: rolling. Behaviour is unchanged: both routers default to the simulator and nothing calls either port in production yet.
- Rollback plan: revert.
- Data migration reversible: N/A — the simulators hold no database state.

## Reviewer notes

One decision, stated twice because it is the substance: **no speculative vendor clients**. I would rather ship a `NOT_BOUND` that names the missing contract than a client nobody can run. Disagree and I will write them.

Second, smaller: the dispute simulator's `respondByDate`. I made it a flat placeholder rather than modelling scheme windows. If you would prefer it absent entirely — null rather than a number that could be mistaken for real — say so; that is arguably more honest still.


JiRaska:
The red `build (openbank-consent-service)` here is not about this diff: coverage 35.14 % against a 69 floor, with the identical test set (130 tests, 1 skipped, 0 failed) that passes on #8864, and a truncated kover `test.ic` (`EOFException`). Same shape as 4 of the last 41 consent builds, two of them on `main` — tracked in #9919. The failed jobs are queued for a rerun once the run completes.

PR #8863, opened 2026-09-05
feat(kyb): legal-entity onboarding, representation mandates and profile switching (ADR-0284)

JiRaska:
## Summary

A sole trader or a company can become a customer, and the human who runs one switches between their personal and business profiles in the same app session. New `openbank-kyb-service` verifies the business identifier against a public register and derives who must sign from the register's own representation rule; party-service gains the entity party and the representation mandate that the customer edge switches profiles on. ADR-0284.

## Type of change

- [ ] fix (bug fix)
- [ ] perf (performance)
- [ ] docs
- [ ] chore (build / tooling)
- [ ] breaking change

## Linked issues

Refs #8866 — the ADR-0284 tail this PR deliberately does not build: the onboarding-service business funnel projection, the D8 owner-graph hooks (Customer 360, Lípa, catalog overlay, campaign segments) and the D9 agents.

## Changes

- **`openbank-kyb-service`** (port 8157, ns `kyb`, db `openbank_kyb`): `LegalEntityIdentifier` with per-scheme normalisation and check digits (IČO mod-11, NIP, SIREN Luhn, LEI ISO 7064), `RegistryExtract` + adapters for ARES (CZ), GLEIF (LEI) and manual attestation, and the `BusinessOnboardingCase` state machine whose signer count comes from the register, not from the applicant.
- **Country packs** (`country-packs/cz-v1.json`): versioned, effective-dated jurisdiction data in the ADR-0212 shape — schemes, register descriptor, UBO source and threshold, legal-form map with cs/en labels, evidence per form. The Czech overlay is that file plus one parser for the free-text *způsob jednání*, biased to the safe side (`UNKNOWN` routes to manual review, never a lower signer count).
- **Temporal case timers** in the ADR-0211 shape: invitation reminder and expiry, idle-case abandonment; build-time selected as lending does, worker switched by `openbank.kyb.worker.enabled`. A Temporal failure never fails a customer step — the case row is already committed and the miss is counted.
- **party-service**: entity party fields (`legalForm`, `registrationCountry`), the `PartyMandate` aggregate + migration + REST surface, `GET /parties/{agent}/acting-for`, and `PARTY_MANDATE_GRANTED`/`REVOKED` through the existing outbox.
- **kyc-service**: `subjectType` and the KYB check set (registry match, representative authority, UBO identification), so a company is not asked for an identity-document scan.
- **customer-edge**: `GET /profiles` and the `X-Acting-For` switch, resolved once at the identity chokepoint and fail-closed; business onboarding proxied under `/business/*` and always performed as the human.
- **Governance**: rules.yaml matrix grants + declared M2M writes, OPA bundle with its own rego extension and tests, AsyncAPI contract for `openbank.kyb.events` with one typed payload class per event type, gitops component (namespace, CNPG + pod identity, Kafka user and topics, msg-override, PodMonitor, Temporal namespace and ingress, network policies), release-please and auto-deploy registration, VEX overlay, runbook.

## Definition of Done

- [x] No new lint warnings.
- [x] Avro schema versioned backward-compatibly (if event changed).

## Security checklist

Every `@Suppress` added here is `TooManyFunctions` or `LongMethod` on a state machine, a contract builder or a per-register mapper, with the reason on the line above it. The new dependency is `io.temporal:temporal-sdk`, already used by lending, domestic-payment and campaign at the same pinned version.

The security-relevant surface is the profile switch. `X-Acting-For` is verified against an ACTIVE mandate in party-service at the single edge chokepoint and is **fail-closed**: no mandate, unreachable party-service, non-200, unparseable body, malformed header or the kill switch off all answer 403. `ActingForResolverTest` names each of those branches. This is deliberately the opposite of `PartyMergeResolver`, which fails open — an unhonoured merge shows a customer less, an unverified switch would show them someone else's company.

## Compliance impact

- [ ] PCI DSS
- [x] DORA
- [x] GDPR
- [x] PSD2 / SCA / RTS
- [x] AML / 5AMLD
- [ ] CNB reporting
- [ ] None

**DORA** — two new ICT third-party dependencies (ARES, GLEIF), both free public registers, declared in the external-feed register with the manual-attestation exit path; an outage becomes `MANUAL_REVIEW`, never a silent pass.
**GDPR** — representatives' names and dates of birth come from public registers under an AML legal obligation, are stored only in the case record and the mandate, are never on the event wire (the payload classes carry no date of birth, and `signerName` only on the invitation), and are erased with the party on the unchanged Art. 17 path.
**PSD2** — every signature and any profile switch that leads to a payment stays SCA-bound (ADR-0021); the acting-for header carries no authority of its own.
**AML/5AMLD** — KYB checks and UBO identification per AMLD5 Art. 30 and zákon č. 253/2008 Sb., recorded on the KYC case with the existing four-eyes review.

## Rollout / Rollback

- Deployment strategy: rolling. The first build lands via auto-deploy on merge (`openbank-kyb-service` is now in `ALL_SERVICES`); the gitops PR rewrites the image pin to the built sha.
- Rollback plan: the service is additive and nothing depends on it. Revert the ArgoCD application, or set `ACTING_FOR_ENABLED=false` on the edge to disable profile switching alone — every existing customer route then behaves exactly as it does today.
- Data migration reversible: yes. party V22 and kyb V1/V2 carry rollback notes; both are additive (new table, new nullable columns).

## Reviewer notes

Three things worth a careful read:

1. **`ActingForResolver`** — the fail-closed switch. Every refusal path returns 403 from one place so no branch can become a fall-open by accident, and the test asserts each branch by name rather than only the happy path.
2. **`CzechRepresentationRuleParser`** — heuristic over free register text, pinned against real phrasings. It resolves a threshold-conditional rule ("samostatně … nad X Kč dva společně") to the STRICTER count, and anything it does not recognise to `UNKNOWN`, which costs a manual review rather than a wrong signer count.
3. **The mandate is not a delegation.** A mandate is a fact sourced from a public register; employee-level access stays an ADR-0232 grant issued by a mandate holder acting for the entity. The ADR records why modelling one as the other was rejected.

Local verification: kyb 35 tests, party and kyc suites green (the one red in party, `PartyPactFolderProviderVerificationTest`'s "no caller identity" interaction, fails identically on a clean `origin/main` worktree here — OIDC is disabled in the local test profile, so a 401 is not reachable). `run-gates.py --all` was run repeatedly; every finding it raised about this change is fixed in the follow-up commits, each named in its message.


JiRaska:
**On the red `CodeQL` check:** one alert, `java/ssrf` (critical) at
`openbank-customer-edge/.../UpstreamClient.kt:405`. Deliberately left open, not dismissed.

The finding is a taint path, not an exploitable hole. The only caller-supplied value that reaches a
URL path here is the KYB invitation token, and it is guarded three times over: shape-checked
(`^[A-Za-z0-9_-]{8,128}$`) before use, URL-encoded, and appended to a **fixed** host prefix that
`UpstreamClient.validatedUri` then re-checks against the allowed-host regex. CodeQL does not model a
regex guard as a sanitizer, so the path stays flagged whatever the guard does.

Not a required context, so it does not block the merge. The structural fix — selecting the target
from a constant map instead of composing a URL string — is a change to a money-path service made for
a static analyser rather than for a defect, and belongs in its own PR if the red badge becomes
annoying enough to be worth that.

Issue #8812, opened 2026-09-05
Card AI agents: dispute evidence, scheme-bulletin compliance, card-ops copilot, agentic commerce PoC (ADR-0283 phase 4)


JiRaska:
Refs #8808 (phase 4). ADR-0283, ADR-0203 (business-plane agents: read-only, human disposition).

## Agents

1. **Dispute evidence agent** — assembles the evidence pack for a chargeback (ledger entries, merchant data from phase 2, 3DS/SCA outcome, cardholder history), proposes reason code and disposition; a human decides (`dispute.decide` is four-eyes). Charter in `agents.yaml`; bounded write surface = a *proposal*, never the decision.
2. **Scheme-bulletin compliance agent** — reads Visa / Mastercard release bulletins (twice-yearly mandates), maps each mandate to our capability ports and adapters, files a governance issue with the effective date. The real, recurring cost of being a card issuer, automated.
3. **Card-ops copilot** over MCP — "why was card X declined", "which cards hit a category limit this week", grounded on `DeclineReason` and the card-processing read model.
4. **Agentic commerce PoC** — Visa Intelligent Commerce / Trusted Agent Protocol and Mastercard Agent Pay (agentic tokens with a spend mandate). Maps onto the existing AP2 / MCP / delegation stack (ADR-0232): an agent holds a scoped, SCA-bound mandate, the network token carries it, card-processing enforces the ceiling. This is where the platform can be first.

## Constraints

- Every agent is read-only or proposal-only (ADR-0203, `check-agent-bounded-write-surface.py`), EU AI Act register regenerated (`gen-eu-ai-act.py`), OPA bundles restamped via `regen-derived.sh`.
- Model calls through the platform's existing LLM gateway and prompt registry; no card data (even synthetic PAN) in a prompt — masked PAN only.

PR #8864, opened 2026-09-05
feat(card-processing): network-token and dispute-case lifecycle — the callers for the phase-2 ports (ADR-0283 phase 3)

JiRaska:
## Summary

ADR-0283 phase 2 wired the tokenisation and dispute ports and their simulated bindings, and
nothing called them. That is the defect the whole ADR was written about: card-issuance had held
a complete authorisation decision since ADR-0194 that no caller ever invoked, so a control that
looked shipped was doing nothing. This PR is the caller for both ports, with the persistence,
events, contracts, policy and REST surface around them.

Stacked on #8858. Its diff against `main` spans the stack; read it after #8837, #8855 and #8858.

## What is here

- **`CardTokenService`** — provision, list and change the status of a network token through
  `TokenisationPort`, mirrored into `card_network_tokens`. The network owns the vault; this bank
  owns the record. Writes reach the scheme first and are mirrored only once it answered.
- **`CardDisputeService`** — open a chargeback, file evidence, refresh status through
  `DisputePort`, persisted in `card_dispute_cases`.
- **Five events** on the existing topic, in a second sealed hierarchy. A token provisioning has
  no `authorizationId`, and inventing one to satisfy the money-path base class would put a field
  on the wire that lies.
- **V3 migration**, two entities with every column name spelled out, two repositories writing the
  row and its outbox message in one transaction, a metrics adapter, DTOs, two resources.
- **openapi.yaml 1.0.0 -> 1.1.0** (eight added paths, no operation touched), **asyncapi** with the
  five messages, and **two authorisation actions** under their own allow reason.

## The properties the tests exist for, each falsified before being trusted

1. **A degraded token read is labelled.** Every answer carries `TokenReadSource`, so a mirror row
   is never presented as a live network state. The test asserts that the live and degraded reads
   of the *same rows* differ in provenance — an assertion on list contents alone cannot see it.
2. **Opening a dispute fails closed.** Replacing the refusal with a local case made exactly
   `a scheme that cannot answer leaves NO local case` fail, and nothing else. A case the network
   never opened renders as active while its representment window expires in silence.
3. **Provisioning is idempotent.** Deleting the idempotency read made exactly
   `a repeated idempotency key ... does not call the scheme` fail. Without it a retried wallet
   request mints a second credential the customer can see.
4. **The policy separation is real.** Narrowing the new allow rule to one action failed the test
   that asserts both, 11/12. The negative half is pinned too: a viewer may read a token list and
   may not suspend one, and the money-path reason must not stretch to cover these actions.

17 unit tests, 0 skipped. Both sabotages reverted and the suite re-run green.

## Gates

`threat-model-updated-on-trust-boundary-change`, `adversarial-contract-test`,
`identifier-intent-guard`, `idempotency-coverage-money-path`, `asvs-l3-mechanical-subset` and
`check-event-contract-code-agreement` all pass locally with `PR_DIFF_BASE` set. The three new
POSTs with no idempotency key are baselined with reasons that are properties of the endpoint: a
state set, a filing addressed at an existing case, and a read-through that writes only on a move.
The two where a retry would genuinely double-create require `Idempotency-Key` and are held by a
UNIQUE index.

## What is deliberately NOT here

No Visa or Mastercard tokenisation or dispute adapter. Both networks put these capabilities
behind a commercial contract, so an adapter written against documentation nobody here can open
would be code that can never be executed and green in every test. Choosing one of those bindings
returns `NOT_BOUND` with a reason naming the contract, which keeps the gap visible at the call
site rather than only in a document.

Refs #8812
Closes nothing on its own — the disputes desk and token screens in admin-ui land next.

## Security checklist

- [x] No secrets, tokens, passwords, or PII in code, config, tests, logs. Scheme credentials are
      `Optional<String>` config properties resolved from Kubernetes secrets; nothing is committed,
      and no card number, CVV or token credential is accepted, stored or logged anywhere in this
      change.
      this PR). The only `@Suppress` added is `ktlint:standard:filename` on files that hold several
      related declarations, the fleet convention for a ports or events file.
- [x] No new third-party dependency, OR: dependency review attached (license, CVE, source). No new
      dependency: every artifact used is already on the fleet's version catalogue.
- [x] Auth / crypto / payment code changed? → tag `security-review-required`. Yes, this is card
      money-path code; the label is applied and the PR needs two approvals under ADR-0030.
- [x] PII path changed? → tag `gdpr-review-required`. No. The card is referenced by its
      card-issuance id and the party by id; no name, address, document number or contact detail is
      read or written here.
- [x] Cardholder data path changed? → tag `pci-review-required`. No. Keeping the PAN outside this
      service is the explicit design: no port signature accepts one, and the network token
      reference is opaque and not derivable to a card number.


JiRaska:
## `build (openbank-sca-service)` is red, and this PR does not touch sca-service

```
Rule violated: lines covered percentage is 66.269400, but expected minimum is 82
```

`git diff --name-only origin/main...<head> | grep sca-service` → **0 files**. What the branch does
add is two files under `openbank-libs-domain`:
`domain/cards/scheme/CardScheme.kt` and `domain/cards/scheme/CardSchemePorts.kt`.

### Measured, both sides

| | covered | total | line % | floor |
|---|---|---|---|---|
| clean `origin/main` | 716 | 839 | **85.34** | 82 |
| this branch (CI) | — | — | **66.27** | 82 |

`main` was run locally on a detached worktree with a **fresh** `GRADLE_USER_HOME`, so this is not a
reused `koverCachedVerify` report — the exact false-green that voided an earlier attempt at this
same class of problem.

### Mechanism

Every service's Kover report measures `com/openbank/libs`, but a service's test task never runs the
libs module's own tests. So new shared code is uncovered *from every dependent's perspective*,
however well `SchemeResultTest` covers it inside libs-domain. 105 source lines become ~240 counted
lines once the data classes' generated `equals`/`hashCode`/`copy`/`componentN` are included, and
sca-service — 839 measured lines, the smallest base near its floor — goes under first.

### What I deliberately did NOT do

Both obvious fixes are known-bad here, and both are recorded from a previous attempt:

- **Excluding `com/openbank/libs` from the report.** Measured on the same head of #5719:
  libs included 60.50 %, libs excluded **30.91 %** against a floor of 65. Dependents' tests cover a
  lot of libs code, so libs holds the aggregate *up*; excluding it puts services further under.
  That change was reverted (`79a006153`).
- **Lowering sca-service's floor.** That is the one move that empties the ratchet of meaning, and it
  would hide the next regression in a service nobody is looking at.

### The actual question

`CardSchemePorts` is card-scheme domain code living in the fleet-wide `openbank-libs-domain`, so it
compiles into the coverage denominator of every service — including one whose subject is strong
customer authentication and which will never call it. The architecturally clean fix is for these
types to live in a card-scoped module rather than shared libs, which stops the fleet-wide drag at
source. That is a design decision spanning the ADR-0283 stack, not something to settle quietly
inside this PR, so I am reporting the measurement rather than picking one of the bad options.

Related: #6384 tracks the general "a libs addition lowers every dependent's coverage" behaviour.

PR #8865, opened 2026-09-05
fix(vop): replace a 401 pact negative the provider replay can never satisfy (main is red)

JiRaska:
## Summary

Both Verification of Payee consumer pacts pinned a negative case the provider replay **cannot
satisfy**, so provider verification has been failing on `main` in two services since #8552.

`AccountPactFolderProviderVerificationTest` and `PartyPactFolderProviderVerificationTest` are
`@QuarkusTest` classes whose requests are authenticated by construction. A pact that says "no
caller identity -> 401" therefore replays as:

```
1.1) status: expected status of 401 but was 200
```

Reproduced on a clean `origin/main` worktree before touching anything: account-service 1 of 6
interactions failing, party-service 1 of 11.

## Why nobody saw it

Path-scoped CI rebuilds a service when its own files change. #8552 touched `pacts/` and
vop-service only, so neither provider module was rebuilt after the pacts landed. Provider
verification is the gate that decides whether a contract is honoured, and it was red and unread.

## The replacement

A **404 for a key the provider does not know** — an unseeded IBAN, an unseeded party id — with a
provider state that deliberately seeds nothing. That is the refusal this hop can actually assert,
and it is the one VoP's behaviour turns on: `AccountHolderNameLookupAdapter` maps an unknown IBAN
to NO_DATA, which is a *valid* Verification of Payee answer, so a route that started answering
200-with-nulls instead of 404 would be indistinguishable from a genuine "we hold no name" in every
log, metric and unit test.

## Measured, not assumed

The first draft used an IBAN with wrong check digits and the replay answered **400, not 404** —
validation rejects it long before the lookup runs. The constant now carries correct mod-97 digits
and the comment records why, because a malformed IBAN tests the validator and only a well-formed
unknown one tests the lookup.

| | before | after |
|---|---|---|
| account-service interactions verified | 5 of 6 | 6 of 6 |
| party-service interactions verified | 10 of 11 | 11 of 11 |
| skipped | 0 | 0 |

Consumer pacts regenerated by re-running the consumer tests, as the drift gate requires.

## Blast radius

This unblocks any PR that rebuilds account-service or party-service — currently the ADR-0283 card
stack, where the same failure appears as a red `build (openbank-account-service)`.

Refs #8552

## Security checklist

- [x] No secrets, tokens, passwords, or PII in code, config, tests, logs. The change is two test
      constants and four provider-state handlers; the unseeded IBAN and party id are invented
      values that match no record.
- [x] No new `@SuppressWarnings`, `as Any`, `@Suppress("...")` without justification. None added.
- [x] No new third-party dependency, OR: dependency review attached (license, CVE, source). None.
- [x] Auth / crypto / payment code changed? → tag `security-review-required`. No production code
      changes at all — this is test and contract material only. What it restores is the ability of
      the provider replay to fail, which is the security property that was lost.
- [x] PII path changed? → tag `gdpr-review-required`. No. No party record is read or written; the
      new state handlers seed nothing.
- [x] Cardholder data path changed? → tag `pci-review-required`. No.


JiRaska:
Closing as absorbed: #8890 landed the same fix on `main` — verified by content, not by title.

`main` now carries, on both VoP pacts, exactly the negative case this PR proposed:

```
GET the account behind the payee IBAN, for its owning party id   -> 200
GET the account behind an IBAN the bank does not hold            -> 404
GET the party whose name VoP compares the payer-supplied name    -> 200
GET a party id the bank does not hold                            -> 404
```

and the same widening of `check-adversarial-contract-tests.py`, which had demanded a 401 or 403
that a `@TestSecurity`-authenticated provider replay can never produce:

```python
NEGATIVE = re.compile(
    r"\b(401|403|404)\b|UNAUTHORIZED|FORBIDDEN|NOT_FOUND|Unauthorized|unauthorized|"
    r"notFound|rejectsUnauthenticated|missingToken|expiredToken"
)
```

The wording of the interaction descriptions differs ("an IBAN the bank does not hold" vs "an IBAN
account-service does not know") and nothing else does. Merging this on top would put a second copy
of one fix in the tree, free to drift from the first — so it is closed rather than reconciled.

What this PR measured is still worth keeping in the record: the provider replay for both services
went 5/6 and 10/11 to **6/6 and 11/11** once the negative case was one the provider could actually
serve, and the 401 version had been red on `main` since #8552 without anyone seeing it, because
path-scoped CI does not rebuild account-service or party-service when only `pacts/` changes.

PR #8874, opened 2026-09-05
feat(transactions): let an operator maintain the merchant catalogue

JiRaska:
Refs #8573 — the backend half of option A.

## The problem, restated from measurement

The enrichment pipeline behind `merchant` / `merchantCategory` is **complete end to end**: descriptor normalisation, one catalogue query per page, customer-edge forwarding *and* declaring both fields, four app screens rendering them including a map. It is not missing. It is **starved**.

`merchant_catalog` held the ~30 rows one migration seeded and **had no writer anywhere in the codebase** — no ingestion job, no endpoint, no admin screen. Adding a merchant meant a database migration and a deploy, so in practice the catalogue never grew, and the edge's own schema concedes the result: `merchant` is absent *"when the acquirer descriptor has no catalogue match, **which is most transactions**"*.

## What this adds

| Route | |
|---|---|
| `GET /api/v1/merchants` | list, paged |
| `GET /api/v1/merchants/unmatched` | **the operator worklist** |
| `PUT /api/v1/merchants/{descriptorKey}` | create or replace, idempotent by key |
| `DELETE /api/v1/merchants/{descriptorKey}` | remove |

### `unmatched` is not in the issue's original shape, and it is the point

I proposed a CRUD editor. Building it made the omission obvious: **a bare CRUD screen is a blank form.** An operator has no way to know which merchants are worth adding, so the catalogue would stay at 30 rows with a nicer way to reach them.

`unmatched` reads a bounded window of recent transaction descriptions, normalises them **the way the lookup does**, drops what the catalogue already resolves, and ranks the rest by how many transactions carried it. Filling the catalogue becomes a worklist ordered by how many customers each entry affects.

Bounded rather than exhaustive deliberately: `MerchantDescriptor.normalise` lives in Kotlin, so the anti-join cannot run in SQL, and the value is in the top of the list, not a complete answer nobody reads to the end. `scan` and `limit` are capped server-side.

### Writes normalise the key

An operator pastes what they see on a statement line — `ALZA.CZ A.S. PRAHA 4` — while the lookup queries `ALZACZ`. Writing the raw string would create rows **nothing ever hits**: a catalogue that looks maintained and enriches nothing. This is the load-bearing property and the tests assert it against the real normaliser, not a mock.

## Boundaries kept

Operator-facing only. The entity's rule is unchanged and its KDoc updated from "read-only in this service" to say who writes it: **public business data only** — a trading name, a logo, a shop location — never anything customer-derived. Nothing here is proxied by customer-edge; every route requires `OPERATOR` or `ADMIN`.

`merchant.list` / `merchant.update` / `merchant.delete` are added to the role-action matrix, alphabetically, with rationale. **transaction-service runs `AUTHZ_ENFORCE=true`**, so an `@Authorize` action with no matching grant fails **closed** — a 403 for a correctly authorised operator that no unit test could see. I checked the mechanism is data-driven (`role_action_matrix[role].grant[]`) rather than bespoke rego before relying on it.

Half a coordinate pair is refused with a 400 rather than reaching the column constraint as a 500, mirroring the invariant the read path already documents: both coordinates or neither, never a pin at latitude 0.

## Verification

- `MerchantCatalogResourceTest` — **10 tests, 0 failures**; `ktlintCheck` clean, `BUILD SUCCESSFUL`.
- **Guard falsified:** removing normalisation from the write path reddens exactly `a raw acquirer descriptor is normalised before it is written` and `a key that normalises to nothing is refused, not written` — those two and nothing else. Restored and re-run green.
- Existing `@QuarkusTest` suites in this service need the IT Postgres/Redpanda stack, which is not available locally; these are pure unit tests over mocked repositories, with the real `MerchantDescriptor` so normalisation is exercised rather than stubbed. CI covers the rest.

## Not in this PR

**The admin-ui screen.** It is the larger half and it needs these endpoints first. Also still open from #8573: non-card spend cannot be categorised at all (`TransactionType` is a ledger taxonomy — everything a customer spends is `DEBIT`), and there is no customer-set override. Both are option B/C there and unchanged by this.

## Security checklist

      The catalogue is keyed by acquirer descriptor and holds public business data only; nothing in
      it is keyed by customer, card or transaction. `GET /unmatched` returns descriptors and
      counts, never the transactions or accounts they came from.
      None added. When detekt's `TooManyFunctions` fired on `PanacheTransactionRepository`, the
      method was moved to `TransactionDescriptorRepository` rather than suppressed.
- [x] Auth / crypto / payment code changed? → **No.** No money path is touched: no transaction is
      created, altered, reversed or posted. The write endpoints edit display metadata that the
      statement renders. Role and OPA gates are added, not relaxed —
      `merchant.list`/`update`/`delete` are new grants in `rules-opa-data.yaml`, so
      `AUTHZ_ENFORCE=true` fails closed on them.
- [x] PII path changed? → **No.** See the first box; the table is deliberately not keyed by any
      customer identifier, stated in `V16__create_merchant_catalog.sql` and the entity KDoc.
- [x] Cardholder data path changed? → **No.** No PAN, no cardholder identity, no card token. The
      acquirer descriptor is the merchant's own text, not the cardholder's.

Threat model updated in the same PR: §4d of `docs/threat-models/openbank-transaction-service.md`,
plus a change-log entry. It records one gap rather than claiming a control — the row stamps
`updated_at` but not who edited it, so attribution for a catalogue change comes only from the
access log.


JiRaska:
Merging with four known environmental failures, all on the GitHub side and tracked in #8690: `CodeQL (java-kotlin, manual)`, `Dependency review`, `dependency-review` (dependency-submission installation rate limit; the submission itself is fixed by #8781 and green on main) and `Trivy` (the scans themselves PASS — only the SARIF upload steps fail on the same rate limit). Everything substantive is green: all-green, Validate manifests, both affected service builds including openbank-transaction-service with the auth-negative pact replay (#8988).

Issue #8866, opened 2026-09-05
ADR-0284 follow-ups: business funnel projection, the owner-graph hooks, and the two KYB agents


JiRaska:
ADR-0284 lands the legal-entity onboarding, the representation mandate and the profile switch. Three parts of the decision are deliberately not in that first slice, and this issue is where they are tracked so the `partial` delivery status has somewhere to point.

**1. onboarding-service business funnel (D6 follow-up).**
`openbank.kyb.events` has a producer and, today, no consumer: the operator review queue is `GET /api/v1/kyb/cases?status=MANUAL_REVIEW` over REST, so nothing depends on the stream to function. The cockpit (ADR-0068) should project the business funnel from it the way it projects the individual one, at which point the topic's entry in `rules.yaml: event_consumer_liveness.allowlist` must be deleted — the allowlist entry is written to be removed by the first consumer PR, not to become permanent.

**2. The owner-graph hooks (D8).**
The mandate is published once (`PARTY_MANDATE_GRANTED`) and is meant to be consumed, never re-derived. Four consumers are named in the ADR and none is built:

- analytics-sink ingests the mandate + `openbank.kyb.events` into the silver layer, so Customer 360 (ADR-0210) gains an `acting_for` edge and an operator sees the businesses as part of one relationship;
- Lípa (ADR-0282) gains a business grove and the owner bridge — an owner's personal Lístky redeemable into their company's fee waivers and back, capped by D5's provisioning. Redemption, never a transfer: a transfer of value between two legal persons is a taxable, AML-relevant event and a bank-side benefit is not;
- the owner bundle as a per-party catalog overlay produced by ADR-0282 D7's deterministic engine from the owner graph, never a hand-priced special;
- `business-owner` / `business-representative` segment attributes in Campaign Studio (ADR-0221), with delivery consent-gated per party as it already is.

**3. The two agents (D9).**
`kyb-analyst` (drafts a disposition for a `MANUAL_REVIEW` case with cited evidence, and proposes a signer count for a *způsob jednání* the parser marked `UNKNOWN` — accepted proposals become pinned parser tests, so the judgement is captured as data rather than trusted at runtime) and `business-copilot` (the ADR-0089 charter extended with the acting-for scope). Both `plane: business`, read-only, proposal-only, through `agents.yaml`, the EU AI Act register and the OPA agent policy.

[...]

PR #8877, opened 2026-09-05
feat(onboarding): project the business onboarding funnel from kyb events

JiRaska:
## Summary

onboarding-service now consumes `openbank.kyb.events` and keeps a per-CASE business read model
beside the existing per-PARTY one, so the operator cockpit can see business onboarding at all
(ADR-0284 D6, first item of #8866). Stacked on `feat/legal-entity-onboarding` (#8863) — review
that one first; this PR's base is that branch, so its diff here is empty by construction.

## Type of change

- [ ] fix (bug fix)
- [ ] perf (performance)
- [ ] docs
- [ ] chore (build / tooling)
- [ ] breaking change

## Linked issues

Refs #8866

## Changes

- `BusinessOnboardingRecord` + `BusinessCaseStage` + `BusinessFunnelStage`: seven board columns,
  mapped totally from the case status. An unknown status is dropped rather than mapped onto a
  neighbouring one — a guessed column reads as a measurement.
- `kyb-events-in` channel with `failure-strategy: dead-letter-queue`, its own explicit topic
  `openbank.dlq.onboarding.kyb-events-in`, the `KafkaTopic` CR, the KafkaUser Read + Write ACLs
  and the `group.id` in the msg-override ConfigMap. All four, or the rethrow wedges on the DLQ
  send itself (#5745, #5751), and the dotted YAML key never reaches the connector (#686).
- `last_event_at` guards ordering: an event older than the stored row is ignored, so a redelivered
  `BUSINESS_SIGNER_INVITED` cannot walk a live customer back to `AWAITING_SIGNATURES`.
- Register facts merge forward instead of being overwritten with null by an event that does not
  carry them; the review reason is cleared once a case leaves `MANUAL_REVIEW`.
- GDPR Art. 17 anonymises the initiator link rather than deleting the case, which the bank must
  retain for AML.
- Three read endpoints under `/api/v1/onboarding/business`, `openapi.yaml` 1.1.0 → 1.2.0
  (additive only). A separate resource class rather than more methods on `OnboardingResource`:
  the two read models are keyed differently, so a shared endpoint returns counts that cannot say
  what they counted.
- Removes the `openbank.kyb.events` entry from `rules.yaml: event_consumer_liveness.allowlist`.
  Its own text said to delete it with the first consumer PR.

## Definition of Done

- [x] No new lint warnings.

14 unit tests, 0 skipped. Falsified rather than assumed:

- the merge-forward test fails when the merge is removed from the projection (sabotaged, 1 failure,
  restored);
- the consumer-liveness gate exits 1 under `--enforce` with the allowlist entry re-added, and 0
  without it.

No integration test: driving a real `@QuarkusTest` through Kafka needs a broker, and the reactive
repository cannot be called from a bare test thread. The projection and the wire parsing are the
parts with logic in them and both are covered; the persistence path is the same
`Panache.withTransaction` idiom as `OnboardingRepositoryImpl`.

## Security checklist

      One `@Suppress("TooGenericExceptionCaught")` on the consumer's parse guard: Jackson raises
      unchecked subclasses of both `IOException` and `RuntimeException` for malformed input and the
      handling is identical for all of them.

## Compliance impact

- [ ] PCI DSS
- [ ] DORA
- [x] GDPR
- [ ] PSD2 / SCA / RTS
- [x] AML / 5AMLD
- [ ] CNB reporting
- [ ] None

The read model carries the legal name of a company and the party id of the human who started the
case. Erasure anonymises that link (the initiator column is NOT NULL by design, so it is set to the
all-zero UUID rather than nulled, which reads as "a person who has been erased" instead of as
missing data) and leaves the case itself, which AML retention requires.

## Rollout / Rollback

- Rollback plan: the consumer is additive — reverting stops the projection and leaves the table.
  The channel, the DLQ topic and the ACLs are separate GitOps resources and can be reverted alone.
- Data migration reversible: yes — `V4` creates one table and one sequence, rollback note in the
  migration header.

## Reviewer notes

The table is deliberately separate from `onboarding_records` rather than a set of nullable columns
on it: that one is keyed per party and counts people, this one is keyed per case, exists before any
entity party does, and carries several humans. Folding them together makes every existing cockpit
count ambiguous.

### DIFF
--- a/.github/CLAUDE.md
+++ b/.github/CLAUDE.md
@@ -384,6 +384,18 @@
   destructive (`gh run rerun`, a deploy, a delete), stub the binary on `PATH` — **and validate
   the stub against a known-positive first**, or a silent passthrough runs the real thing.
 
+- **A new Kafka channel is TWO deploys, not one, and doing it in one is a hard boot failure.**
+  The msg-override ConfigMap that supplies `group.id` / `auto.offset.reset` (the #686 apparatus) is
+  applied by Argo BEFORE auto-deploy bumps the image, so a channel named there that the RUNNING
+  image has no `connector:` for is SRMSG00071 — SmallRye refuses to start, so the whole service
+  stops rather than the one consumer degrading. `msg-channel-image-parity` compares the ConfigMap
+  against the `application.yaml` inside the image that is actually deployed and blocks it at PR
+  time. Order: (1) code + `application.yaml` channel + the ACLs, merge, let the image ship;
+  (2) the ConfigMap lines. In between, the consumer runs on the fallback group id
+  (`quarkus.application.name`), so grant the KafkaUser Read on **both** group names or every poll
+  answers `GroupAuthorizationException` — which is silent, because a consumer that cannot join a
+  group looks exactly like a topic with nothing on it (#8877, #8882).
+
 ### main-protection: the bypass has two halves, and only one had a reader
 - **`rulesets/rule-suites` (the bypass LOG) is private and rejects fine-grained tokens;
   `rulesets/{id}` (the bypass CONFIG) is world-readable.** Same feature, opposite access, and it
--- a/openbank-infra/gitops/components/kafka/kafka-dlq-topics.yaml
+++ b/openbank-infra/gitops/components/kafka/kafka-dlq-topics.yaml
@@ -700,3 +700,18 @@ spec:
   config:
     retention.ms: 2592000000
     cleanup.policy: delete
+---
+# openbank-onboarding-service :: channel `kyb-events-in` (ADR-0284 D6)
+apiVersion: kafka.strimzi.io/v1
+kind: KafkaTopic
+metadata:
+  name: openbank.dlq.onboarding.kyb-events-in
+  namespace: messaging
+  labels:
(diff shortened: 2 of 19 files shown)


JiRaska:
`gates (gitops-api)` was right, and the finding is worth recording rather than just fixing.

`msg-channel-image-parity` compares every channel named in a service's msg-override ConfigMap
against the `application.yaml` **inside the image that is actually deployed**. Argo applies the
ConfigMap before auto-deploy bumps the image, so this PR would have named `kyb-events-in` in config
that the running onboarding-service image cannot serve — SRMSG00071, a hard boot failure of the
whole service, not a degraded consumer.

Fixed by moving the two `group.id` / `auto.offset.reset` lines out of this PR and into #8882, to be
added once an image carrying the channel is deployed. In the meantime the consumer runs on the
fallback group id (`quarkus.application.name`), so the KafkaUser now grants Read on **both** group
names — without the fallback grant every poll answers `GroupAuthorizationException`, which looks
exactly like a topic with nothing on it. That interim is acceptable only because the topic is new
and carries no backlog.

Verified locally: the gate reports `32 channel override(s) across 20 service(s); every one is
servable by the image actually deployed`.

PR #8879, opened 2026-09-05
feat(kyb): GB country pack, Companies House register and the UK PSC beneficial-ownership adapter

JiRaska:
## Summary

A second country pack, and a register adapter for it. The point is not the UK: it is that the
country pack from ADR-0284 D2 is an abstraction rather than a Czech special case, and only a
jurisdiction whose register answers a *different shape* can demonstrate that. Stacked on
`feat/legal-entity-onboarding` (#8863) — that PR is the base, so review it first.

## Type of change

- [ ] fix (bug fix)
- [ ] perf (performance)
- [ ] docs
- [ ] chore (build / tooling)
- [ ] breaking change

## Linked issues

Refs #8866

## Changes

- `gb-v1.json`: Companies House company types mapped onto the fleet legal-form classes, the PSC
  register as the UBO source, UK evidence lists, and the 2017 Money Laundering Regulations as the
  AML basis.
- `CompaniesHouseRegistryAdapter`: company profile plus officers, over the free (key-gated) API.
- The UK register carries no way-of-acting text, because UK law puts the power to bind a company in
  the articles. The pack declares that (`representationRuleParser: null`,
  `listsRepresentationRule: false`) and the rule is derived under the model articles: any single
  director binds. A company whose officers could not be read is `UNKNOWN`, never `SOLE`.
- Only officers who can bind are offered as signers. A resigned director is history and a secretary
  cannot sign; offering either produces a contract the register does not support.
- Companies House publishes a director's birth month and year only, so `dateOfBirth` stays null
  rather than being reconstructed from partial data.
- With no API key the adapter declines `GB_CRN` entirely and `RegistryRouter` falls through to
  manual attestation — the difference between "we could not check" and "the register said no". The
  class is `@Startup` so the "no key configured" warning is emitted at boot rather than on the
  first GB lookup, which for a bank with no UK customers is never.
- The new host is declared in `check-external-feeds.py` under `NOT_PROBED`, with the same reason
  ARES and GLEIF carry: a per-entity lookup API has no fixed document to probe.

## Definition of Done

- [x] No new lint warnings.

10 tests in the two mapping suites, 0 skipped, 5 of them new. The signer filter was falsified by
deleting it: the binding-roles assertion goes red, so the test can detect the absence of what it
tests. No integration test — the adapter's only I/O is a third-party register, and a test that
calls it would assert a real company's continued existence rather than our mapping.

## Security checklist

- [x] No secrets, tokens, passwords, or PII in code, config, tests, logs. No API key is configured
      in gitops and none is in this repo; the fixtures are invented companies.
      One `@Suppress("TooManyFunctions")`, with the same reason `AresRegistryAdapter` carries: the
      private mappers exist because folding them back reproduces the oversized function they came
      from. One `@Suppress("TooGenericExceptionCaught")` on the register call boundary.

## Compliance impact

- [ ] PCI DSS
- [ ] DORA
- [x] GDPR
- [ ] PSD2 / SCA / RTS
- [x] AML / 5AMLD
- [ ] CNB reporting
- [ ] None

The adapter reads names and roles of company officers from a public register, and deliberately does
not store a date of birth for GB officers because Companies House does not publish a full one.
`amlLegalBasis` in the pack names the UK regulation the customer due diligence rests on.

## Rollout / Rollback

- Rollback plan: additive — removing the pack from `index.json` returns GB to manual attestation.
  No schema, no event, no API change.

## Reviewer notes

The interesting review question is the derived representation rule. `SOLE` for a UK company is a
statement about the model articles, not about the extract, and a company that has adopted bespoke
articles requiring two signatures would be wrong here. That is why an empty officer list is
`UNKNOWN` rather than `SOLE`, and why the pack records that the register does not carry the rule:
the honest fix if this becomes a real exposure is an operator-confirmed rule on the case, not a
better parser.


JiRaska:
Pushed the UBO adapter onto this branch (ADR-0284 D5, the last adapter item of #8866) — it belongs
here because it is served by the same Companies House client and the same country-pack abstraction.

- `UboAdapter` / `BeneficialOwnershipPort`, separate from the company-register port: the UK serves
  companies and their PSCs from one API, the Czech `Evidence skutečných majitelů` is a different
  register with no public API, and folding them together makes "the register answered" ambiguous
  about *which* register.
- `CompaniesHousePscAdapter` reads persons-with-significant-control. Ownership is a **band**, never
  a percentage — the register publishes `ownership-of-shares-25-to-50-percent` and nothing finer, so
  a precise figure would be invented and then compared against by a threshold test. A person holding
  shares in one band and votes in another takes the higher one; a control with no figure at all is
  `UNQUANTIFIED`, not below-threshold, because the register named them for a reason.
- A filed statement ("no individual PSC identified") is carried separately from an empty owner list.
  It is an answer the company filed; collapsing it into the absence shows an analyst the same screen
  as a company that has not filed at all.
- `SelfDeclarationUboAdapter` is the fallback everywhere else, and deliberately does **not** derive
  an owner from the directors: a director is not a beneficial owner, the two coincide often enough
  for the mistake to look right in testing, and a fabricated UBO satisfies the evidence requirement
  while being unevidenced.
- `UboRouter` degrades an unreachable register to `UNAVAILABLE`, never to `SELF_DECLARATION` — the
  two need different work, and merging them makes a transient outage look like a customer who owes
  us a declaration, which nobody retries.
- `GET /api/v1/kyb/ubo`, action `kyb.ubo.read`, granted to `ROLE_OPERATOR` and `ROLE_KYC_REVIEWER`
  and absent from the edge principal's enumerated list: a beneficial-ownership extract is personal
  data about third parties who are not the caller. Always 200 — a `SELF_DECLARATION` or `UNAVAILABLE`
  finding is an answer, and 404 would read as "this company has no owners".
- `openapi.yaml` 1.0.0 → 1.1.0 (additive). The ~78 OPA bundle restamps are the documented
  consequence of touching `role_action_matrix`, which is in the policy read set; regenerated with
  `regen-derived.sh`, never by hand.

9 unit tests plus 2 rego tests, 0 skipped, both falsified: degrading an outage to
`SELF_DECLARATION` reddens the router test, and adding `kyb.ubo.read` to the edge list reddens the
rego test. `check-authz-pdp-parity` and `check-matrix-write-grants` pass (`kyb.ubo.read` ends in
`read`, so it is not a matrix write grant). The three `gitops` gates that fail locally
(`litellm-config-revision`, `credential-deadline-ratchet`, `loki-rule-load-test`) fail identically
on the base branch — they need PR context this checkout does not have.

PR #8880, opened 2026-09-05
feat(analytics): ingest kyb events into the Customer 360 silver layer

JiRaska:
## Summary

The business half of the customer graph reaches analytics (ADR-0284 D8): which legal entities
exist, who may act for them, and how far each onboarding case got. That is what lets Customer 360
see that a personal customer also runs two companies — the connection the ADR exists for. Stacked
on `feat/legal-entity-onboarding` (#8863), which is the base.

## Type of change

- [ ] fix (bug fix)
- [ ] perf (performance)
- [ ] docs
- [ ] chore (build / tooling)
- [ ] breaking change

## Linked issues

Refs #8866

## Changes

- `openbank.kyb.events` added to the sink's topic list and to the analytics-sink KafkaUser Read
  ACL, in one change: subscribing without the ACL is a `GroupAuthorizationException` on every poll,
  the ACL alone is a grant to read nothing, and neither half reports the other missing.
- `PayloadMasker` masks `signerName` and `legalName`. A company name is not personal data, but a
  sole trader's legal name is that person's own name and no field-name-keyed rule can tell the two
  apart. The identifier is deliberately left in the clear: it is what the warehouse joins on, and
  for a company it is not personal data.
- `RetentionPolicies` maps `KYB` onto the KYC category. The unmapped default holds it for the same
  period on the same non-erasable footing, so this changes the declared legal basis rather than the
  outcome — which is the point, because an unclassified aggregate reads exactly like a classified
  one in every retention report.
- Removes the `openbank.kyb.events` entry from `rules.yaml: event_consumer_liveness.allowlist`.

**Merge note:** #8877 removes that same allowlist entry, for its own consumer. Whichever lands
second will conflict on those lines; the resolution is to keep one comment, not to restore the
entry.

## Definition of Done

- [x] No new lint warnings.

4 tests, 0 skipped. Falsified rather than assumed: dropping the `legalName` masking rule turns the
masking test red, and the consumer-liveness gate exits 1 with the allowlist entry present and 0
without it, under `--enforce`.

## Security checklist

- [x] No new `@Suppress` of any kind.

## Compliance impact

- [ ] PCI DSS
- [ ] DORA
- [x] GDPR
- [ ] PSD2 / SCA / RTS
- [x] AML / 5AMLD
- [ ] CNB reporting
- [ ] None

A new personal-data stream enters the warehouse. The two name fields are masked on the way in, the
records are attributed to `KYB` and retained on the AML legal basis, and erasure follows the
existing aggregate-keyed path — `KYB` is non-erasable for the same reason `KYC` is, which is a
statutory hold and should be reviewed as such rather than assumed.

## Rollout / Rollback

- Rollback plan: remove the topic from the list; the ACL becomes an unused grant and no schema
  changes. Rows already written stay, masked as they were written.

## Reviewer notes

The masking decision is the one worth arguing with. Masking `legalName` for every legal form costs
readable company names in the silver layer and buys the sole-trader case, which cannot be separated
by field name. The alternative is to mask conditionally on `legalFormClass`, which the masker has no
concept of today. If the warehouse turns out to need plain company names, the honest fix is a
form-aware rule, not an exemption.


JiRaska:
**Why this is red, and why I am not fixing it here.**

`build (openbank-transaction-service)` fails on `main`, not on anything in this PR:

```
au.com.dius.pact.provider.junitsupport.MissingStateChangeMethod:
  Did not find a test class method annotated with @State("no valid M2M identity is presented")
  for Interaction "POST the MT103 settlement debit with a missing or expired token"
```

Three consumer pacts (swift, sdd, interest → transaction-service, added by #8697) declare that
provider state and `TransactionPactFolderProviderVerificationTest` has no handler for it, so
pact-jvm fails the interaction before it ever issues a request. Behind that sits the same defect
#8876 fixed for the VoP pacts: those interactions expect **401**, and the provider replay
authenticates every replayed request, so they could not pass even with a handler.

#8988 and #8983 are already open against exactly this, so I have not written a competing fix — same
reasoning as the duplicate-PR check this repo asks for. Once either lands, merging `main` into this
branch should clear it; auto-merge is armed and will take it from there.

Everything else red here is the installation API rate limit (`Trivy`, `dependency-review`), which is
not a required context.

Issue #8792, opened 2026-09-05
ADR-0282 phase 1 — the data Lípa needs: partyId on ACCOUNT events, a post-settlement transaction fact, a per-party profile


JiRaska:
Phase 1 of ADR-0282 (Lípa). **Nothing else in the ADR can start honestly until this lands** — ADR-0220 D5's rule applies: a rewards programme rendered from fabricated profiles is worse than a missing feature.

## Why

Measured on `main`, 2026-09-05:

- `openbank-campaign-service`'s segment DSL has four rules and **two cannot be constructed at all**. `SegmentRule.HasAccount` is rejected because no ACCOUNT event in the analytics layer carries a `partyId` (0 of 59 bronze rows, #2891). `SegmentRule.HasActiveConsentScope` is rejected because consent events are not ingested. So "holds an account" — the most basic segmentation predicate a bank has — is unanswerable.
- The only transaction fact in analytics is `openbank.transactions.transaction.initiated`. It carries no category, no MCC, no merchant, no amount band, no rail, and it is emitted before the money moves. Savings rate, buffer months and spend cadence are all underivable.
- `silver_current_state` is the latest event per `(aggregate_type, aggregate_id)`. There is no per-party profile anywhere; the admin-ui Customer 360 route re-derives what it needs by querying silver directly on every request.
- `OnlineFeatureStore` (ADR-0140) has exactly one consumer, `openbank-fraud-service`. The offline half is unbuilt, so no segment can be backtested.

## Scope

- [ ] ACCOUNT events carry the owning `partyId` (#2891). Producer change in `openbank-account-service`, consumer/ingest change in `openbank-analytics-sink`, and `SegmentRule.HasAccount.unsupportedReason` becomes `null` in the same PR.

  **Partly delivered since: `AccountInitialLoadSource` landed with #8905 (merged 2026-09-07 13:57Z)**, so the
  ingest side now has a real `INITIAL_LOAD` seed with a deterministic `event_id`. The producer half below is
  unchanged by it.

[...]

Issue #2891, opened 2026-07-31
Segment DSL queries columns the silver layer does not have — every campaign enrols 0 parties, silently


JiRaska:
## Summary

`Segment.toWhereClause()` generates SQL over columns that do not exist in
`openbank_analytics.silver_current_state`. Every segment evaluation therefore fails, and because
`SilverSegmentEvaluator` is fail-closed, the failure is reported to the caller as an **empty
cohort** rather than an error. `POST /api/v1/campaigns/{id}/enrol` returns `enrolled: 0` for every
campaign and every segment, which reads as "nobody matched".

Found while running the first sandbox rollout end-to-end (#2749). It blocks the rollout's stated
goal — a consent-gated journey with `enrolled >= 1` — and no segment can satisfy it today.

## Evidence

All four rules in the closed DSL reference columns the table does not have:

| Rule | Column it needs |
|------|-----------------|
| `PartyStatusIs` | `state` (as `jsonExtractString(state, 'status')`) |
| `TenureAtLeastDays` | `first_seen` |
| `HasAccount` | `has_account` |
| `HasActiveConsentScope` | `consent_scopes` |

Actual schema of `silver_current_state` — an event log, not a per-party projection:

```
aggregate_type, aggregate_id, event_id, aggregate_version,
event_type, occurred_at, source_service, schema_version, payload
```

Running exactly what the DSL renders for `PartyStatusIs("ACTIVE")`:

```
SELECT DISTINCT aggregate_id FROM openbank_analytics.silver_current_state
WHERE (aggregate_type = 'PARTY' AND jsonExtractString(state, 'status') = {p0_status:String})

Code: 47. DB::Exception: Unknown expression or function identifier 'state' (UNKNOWN_IDENTIFIER)
```

And no other table in the database supplies them:

```
SELECT table, name FROM system.columns
WHERE database='openbank_analytics'
  AND name IN ('state','first_seen','has_account','consent_scopes')

gold_screen_feedback_by_screen   first_seen
reload_proposals                 state
```

Both hits are unrelated tables. `has_account` and `consent_scopes` exist nowhere.

## Why nothing caught it

- `SilverSegmentEvaluator` catches the non-200 and returns `emptyList()` — correct as a fail-closed
  policy, but it converts a schema error into a plausible business outcome. The error is only in the
  service log; the API says `enrolled: 0`.
- `enrol()` has exactly one cohort source (`segmentEvaluation.evaluate(segment)`), so there is no
[...]

PR #8886, opened 2026-09-05
feat(campaign): HasActiveConsentScope works — and it is still not the consent control

JiRaska:
## Summary

The second segment rule disabled by a stale reason. `openbank.consent.events` has been ingested since 2026-06-26; the claim that consent events "are not ingested into the analytics layer at all" was written on 2026-07-31, in the same commit that disabled `HasAccount`. With this, every rule in the segment DSL is usable.

**Stacked on #8885** — it builds on that branch, so review that one first. The shared files drop out of this diff once it lands.

## Type of change

- [ ] fix (bug fix)
- [ ] perf (performance)
- [ ] docs
- [ ] chore (build / tooling)
- [ ] breaking change

## Linked issues

Refs #8792 (ADR-0282 phase 1), Refs #2891

## This is not the consent control, and the KDoc says so first

The binding check is the live per-send call in `LiveConsentCheckAdapter` (ADR-0198). This rule narrows a cohort so a send is not attempted against parties who would be refused anyway. A campaign without it is not one bit less consent-gated, and a campaign with it is not one bit more.

That distinction is concrete rather than ceremonial: silver is a projection with its own lag, so a party who revoked consent seconds ago can still appear in this cohort. The live check is what stops the message. Anyone reading a green cohort here as "these people have consented" is reading it wrong, which is why the warning is the first paragraph of the rule's documentation and not a footnote.

## Why this one reads silver when the previous one reads bronze

The two rules make opposite source choices one after the other, and the reason is the question, not a habit.

| rule | question | source | why |
|---|---|---|---|
| `HasAccount` | historical: has ever held an account | bronze | only `AccountCreated` carries `partyId`; a silver row for an account that changed status has lost it |
| `HasActiveConsentScope` | current: is a consent live now | silver | silver keeps the latest event, so a revoked, superseded or expired consent drops out by construction |

Reading bronze here would match every consent ever granted, including the ones the customer has since withdrawn. Reading silver in `HasAccount` would omit exactly the parties with the most account activity.

## Evidence the premise was stale

| fact | established by |
|---|---|
| the topic is ingested | `openbank.consent.events` in analytics-sink's `application.yaml` since 2026-06-26 |
| a CONSENT aggregate exists | `ConsentGranted` and siblings declare `aggregateType = "Consent"` |
| the party link is there | `ConsentGranted.partyId` |
| the scopes are there | `ConsentGranted.scopes: Set<ConsentScope>` |
| both survive ingest | `PayloadMasker.PII_KEYS` lists neither |

## What the rule can and cannot check

A mis-shaped scope (lower case, a space, empty) is rejected where the segment is **defined**, as the rest of the DSL does. A well-formed name that `ConsentScope` has never heard of cannot be rejected — campaign-service does not depend on consent-service — and it matches nobody rather than erroring. That limit is stated on the rule, and the vocabulary is pinned by a test that reads consent-service's own enum.

## A test was removed rather than made to pass

With both rules now supported, no input can reach `Segment.init`'s unsupported-rule branch, and `SegmentRule` is sealed precisely so a stand-in cannot be conjured in a test — the compiler refuses it. Faking one would have produced a test that proves nothing.

So the test is gone and the guard is documented in place instead: it is the landing pad for the next rule written ahead of its data, it re-arms the moment one declares `unsupportedReason`, and nobody should delete it for being green. Saying that in the code beats a test that cannot fail.

## Verification

```
./gradlew :openbank-campaign-service:test --tests '*SegmentEvaluationTest*' --tests '*ProducerPinTest*' \
          :openbank-campaign-service:detekt :openbank-campaign-service:ktlintCheck
```

18 tests, 0 failures, **0 skipped** — the skip count matters because both pin tests skip themselves when the sibling module is absent, and a skip reads as a pass.

**Falsified, not asserted:**

| sabotage | what goes red |
|---|---|
| drop the `event_type = 'ConsentGranted'` filter, so a revoked consent matches | `HasActiveConsentScope reads current state, and only a live grant counts` |
| interpolate the scope instead of binding it | `HasActiveConsentScope renders a parameterised scope, never interpolated` |

**What I could not verify:** the same limit as #8885. Cohort behaviour against real data needs a live ClickHouse this session does not reach. The SQL shape and the cross-module couplings are tested; the cohort is not.

## Definition of Done

- [x] Hexagonal layering preserved (domain has zero framework imports). — pure domain
- [x] Unit tests added/updated. — 3 new cases plus a second pin test
- [x] Integration tests added/updated (if service boundary involved). — N/A; the live half is called out above
- [x] `./gradlew detekt ktlintCheck koverVerify build` passes. — test, detekt, ktlintCheck green on the module
- [x] No new lint warnings.
- [x] OpenAPI spec regenerated (if REST API changed). — N/A
- [x] Avro schema versioned backward-compatibly (if event changed). — N/A, read-side only
- [x] Docs updated (README, ADR if architectural). — the reasoning is in the rule's KDoc

## Security checklist

- [x] No new third-party dependency, OR: dependency review attached (license, CVE, source). — none
- [x] Auth / crypto / payment code changed? → tag `security-review-required`. — no; consent-service is read about, not modified
- [x] PII path changed? → tag `gdpr-review-required`. — reads a party UUID and a scope name the sink already stores unmasked; adds no field and no store

## Compliance impact

- [ ] PCI DSS
- [ ] DORA
- [x] GDPR
- [ ] PSD2 / SCA / RTS
- [ ] AML / 5AMLD
- [ ] CNB reporting
- [ ] None

Consent is the Art. 7 basis for marketing, so this is squarely in scope — and the important compliance statement is the negative one: this rule does not become the consent gate, and the live per-send check remains the only thing that decides whether a message may be sent.

## Rollout / Rollback

- Deployment strategy: rolling, with campaign-service.
- Rollback plan: revert; the rule returns to being rejected at construction.
- Data migration reversible: N/A, read-side only.

## Reviewer notes

The judgement worth a second opinion: a consent that expired is excluded because its latest event is `ConsentExpired`, which relies on consent-service actually emitting that event when the clock passes the expiry rather than only computing expiry at read time. If expiry is lazily evaluated there, this cohort will include expired consents until something writes the event — and the live check would still refuse the send, so the failure is a wasted enrolment rather than an unlawful message.

### DIFF
--- a/openbank-campaign-service/src/main/kotlin/com/openbank/campaign/domain/model/Segment.kt
+++ b/openbank-campaign-service/src/main/kotlin/com/openbank/campaign/domain/model/Segment.kt
@@ -22,6 +22,13 @@ data class Segment(val name: String, val version: Int, val rules: List<SegmentRu
         // Fail where the segment is DEFINED, not at enrolment. An unsupported rule used to render
         // SQL that ClickHouse rejected, which the fail-closed evaluator turned into an empty cohort
         // — indistinguishable from "nobody matched" (#2891).
+        //
+        // CURRENTLY UNEXERCISED, and that is worth knowing rather than discovering: since #8792
+        // every rule in the sealed catalogue is supported, so no input can reach this branch and no
+        // test can construct one — `SegmentRule` is sealed precisely so a stand-in cannot be
+        // conjured in a test. The guard is not dead code; it is the landing pad for the next rule
+        // written ahead of its data, and it re-arms the moment one declares `unsupportedReason`.
+        // Nobody should delete it for being green.
         rules.forEach { rule ->
             require(rule.unsupportedReason == null) {
                 "segment rule ${rule::class.simpleName} cannot be evaluated: ${rule.unsupportedReason}"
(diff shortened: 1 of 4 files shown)


JiRaska:
Measured against the sandbox warehouse, because a fail-closed segment rule returns an empty cohort for a broken rule and for a genuinely empty one, and those must be told apart.

**The rule is correct. The cohort is empty today, and that is a fact about the data.**

Every consent aggregate in the warehouse has been revoked. Nine aggregates, nine granted, nine revoked, the same nine ids:

| granted ids | revoked ids | distinct ids |
|---|---|---|
| 9 | 9 | 9 |

So the latest-state fold holds only `ConsentRevoked`, and a rule asking for an active grant correctly matches nobody.

**Positive control, so the emptiness is not hiding a broken predicate.** The same predicate with the latest-state fold removed, run over the granted rows:

| matching aggregates | distinct parties |
|---|---|
| 5 | 2 |

`ConsentGranted` payloads do carry both keys the rule reads, confirmed by listing the payload keys: `partyId` and `scopes`, the latter holding values such as `MARKETING_COMMS_EMAIL`, `CREDIT_OFFERS`, `CREDIT_PROFILE_USE` and `CREDIT_AI_AGENT`.

**This also corrects the premise in #8792.** That issue says the rule was unsupported because consent events are not ingested. They are: 18 rows across two event kinds, the newest from 2026-09-04. The same staleness applied to `HasAccount`, whose blocking claim was written five weeks after the data landed.

Issue #8896, opened 2026-09-05
ADR-0213 D4: the credit policy is Kotlin literals — no four-eyes table store, and one risk appetite for every jurisdiction


JiRaska:
## Where ADR-0213 actually stands

D1–D3 ship: a pure evaluator over versioned decision tables, fail-closed to REFER, evidence pinned on every application. **D4 does not.** The only `CreditPolicyPort` binding is `StarterCreditPolicy`, whose tables are Kotlin literals — four rules for eligibility and affordability, two pricing bands, all `version = 1`, `effectiveFrom = EPOCH`.

Consequences, all visible on the new credit-risk console (it reports `codeSeeded: true` and says so on the page):

- A credit-risk officer cannot change a threshold. A PR, a review and a release can.
- The four-eyes activation D4 promises has nothing to activate; the ADR-0212 compliance packs have that machinery and the policy tables do not.
- **One policy for every jurisdiction.** `PolicyAttribute.JURISDICTION` exists and the pack's `mandatoryChecks` merge into ELIGIBILITY, so the country-specific *statutory floor* is per-jurisdiction — but the bank's own risk appetite (DSTI, DTI, pricing bands) is a single fleet-wide table. That is the real blocker for operating in more than one market.
- `effectiveFrom = EPOCH` with one version means "replay a historic decision against its pinned policy" is trivially true and proves nothing.

## What D4 needs

1. A stored, versioned table set — same governance shape as ADR-0212 packs: proposal, four-eyes activation, immutable once active, `effectiveFrom`.
2. A key that includes jurisdiction (and probably product type), so a market's appetite is its own row rather than a branch inside a shared table.
3. An `@Alternative @Priority` binding replacing `StarterCreditPolicy` with no change to the wiring (the port was designed for this).
4. Admin-ui authoring on top of the existing read view: the console already renders every table, rule, version and effective window, and counts how often each rule fires.

## Free once it lands

The console's charts already read their thresholds from `/api/v1/lending/risk/policy` rather than from constants, so a stored policy changes the overlays with no UI change. Rule hit counts already show a rule that never fires.

JiRaska:
Still true, and measured on `origin/main`:

```
CreditPolicyPort.kt:34   @ApplicationScoped @Default class StarterCreditPolicy : CreditPolicyPort
                         tables are Kotlin literals, version = 1, effectiveFrom = LocalDate.EPOCH
[...]

Issue #8895, opened 2026-09-05
Credit regulatory reporting gaps: F 18/F 19 (NPE, forbearance), F 12 (allowance movements), C 07.00, CNB DSTI/DTI/LTV limits, AnaCredit LOAN feed


JiRaska:
## What exists

- **AnaCredit** (ADR-0037): derive-only service, credit dataset for a monthly reference date. Instrument types `OVERDRAFT, CREDIT_CARD_CREDIT, REVOLVING_CREDIT, LOAN`; exposures arrive by REST upsert, and a `loan.stage_changed` consumer keeps a per-loan IFRS 9 stage projection.
- **FINREP/COREP** (ADR-0097): `F 01.01`, `F 01.02`, `F 01.03`, `F 02.00`, `C 01.00`, derived from the attested close, with missing data flagged as explicit zeroes.

## What is missing on the credit side

| Return | Why it is needed | Data already in the platform |
|---|---|---|
| **F 18.00** — performing/non-performing exposures | The core NPE and forbearance disclosure. The bank already records forbearance assessments (ADR-0215 D1) and a defaulted state at the CRR Art. 178 threshold, and reports neither. | `loan.status`, `ForbearanceAssessment`, `loan_provisioning.days_past_due` |
| **F 19.00** — forborne exposures | Companion to F 18; forbearance is a domain operation today with no return behind it. | as above |
| **F 12.00/F 12.01** — movements in allowances | Stage-to-stage movement of impairment. Every provisioning cycle already books a **delta**, which is the exact input this template wants. | `loan_provisioning` per (loan, period), `PostingKind.PROVISIONING` |
| **C 07.00** — credit risk, standardised approach | COREP stops at own funds (`C 01.00`); the RWA side of the credit book is absent. | exposure, collateral register, `LoanStatus` |
| **CNB DSTI/DTI/LTV limits** | The macroprudential limits are supervised as a share of new production, with a permitted exception quota. Nothing measures the share, so nothing can report it. | the decision evidence added by the credit-risk console (`decision_*` columns) |
| **AnaCredit instrument `LOAN` fed from lending** | The type exists in the enum; the loan book has no publisher into it, so the granular dataset covers overdrafts only. | `openbank-lending-service` loan lifecycle events |

## Sequencing suggestion

[...]

Issue #8894, opened 2026-09-05
Credit policy decision: does the DSTI floor move to total debt service (CNB definition), and per jurisdiction?


JiRaska:
## The finding

The affordability floor the engine applies is **not** the ratio the CNB definition means.

`OriginationDecisionService.affordabilityRatios` computes `DSTI = new installment / verified income`. `LoanApplication.existingDebtServiceMonthly` is collected at intake, persisted, and passed to the engine as its own attribute (`EXISTING_DEBT_SERVICE_MONTHLY`) — but the starter policy has no rule over it, so an applicant's existing obligations do not enter the ratio the `DSTI ≤ 0.45` rule tests.

Debt-service-to-income in supervisory usage is **total** debt service over income. An applicant already paying 6 000 CZK a month against a 60 000 CZK income arrives at the floor with 10 percentage points of headroom the bank does not see.

## Status now

Not a live defect and not silently wrong: the credit-risk console renders both figures side by side (`dsti` and `dstiIncludingExistingDebt`), the second labelled as the CNB definition the engine does **not** read, and the scatter has a toggle between them. So the gap is visible and quantified per application. The read side deliberately did not change the engine — which number the floor tests is a credit-policy decision, not a refactor.

## The decision to take

1. Does the affordability rule move to total debt service? That is a real tightening: applications that pass today would refer or decline.
2. If yes, it needs a new `PolicyAttribute` (or a computed `DSTI_TOTAL`), a table version bump, and — under ADR-0213 D4 — a four-eyes activation, which does not exist yet (see the D4 table-store issue).
3. Is the answer the same in every jurisdiction? The CNB applies DSTI/DTI/LTV limits; other jurisdictions in the compliance-pack set do not. This is exactly what the ADR-0212 pack is for, so the threshold likely belongs per-pack rather than in one fleet-wide table.

## Evidence

- Both ratios and the arithmetic: `CreditRiskInsightServiceTest`, `CreditRiskConsoleIT` (asserts the total exceeds the engine figure by exactly `existingDebtService / income`).
- The console overlays the threshold read from `/api/v1/lending/risk/policy`, so once the rule changes the chart follows with no code edit.

JiRaska:
Still an open decision, and nothing in the repo has recorded one. Measured on `origin/main`:

```
OriginationDecisionService.kt:149  "…`dstiIncludingExistingDebt` adds `existingDebtServiceMonthly`"
                            :170  computes it — for the VIEW only
[...]

Issue #8893, opened 2026-09-05
Warehouse knows nothing about credit: openbank.lending.events is not in the analytics-sink subscription


JiRaska:
## The gap

`openbank.lending.events` is not in the analytics-sink's subscription, so the warehouse holds **nothing** about credit. Measured on `origin/main`:

- `openbank-analytics-sink/src/main/resources/application.yaml` subscribes to twelve topics. The lending topic is not one of them (`openbank-lending-service/src/main/resources/application.yaml:293` publishes to `openbank.lending.events`).
- The only credit rows in ClickHouse come from `V11__credit_funnel.sql`, which reads `openbank.credit.funnel.events` — a customer-journey stream (consent, quote, application started/abandoned, ADR-0269 rule 8). It carries no decision, no outcome, no stage and no exposure, and it says so in its own header comment.
- So none of the nine lending event types reaches bronze: `credit.decision.evaluated`, `credit.application.transition`, `credit.loan.transition`, `loan.disbursed`, `loan.interest_accrued`, `loan.written_off`, `loan.rescheduled`, `loan.stage_changed`, `loan.provisioned`.

## Why it matters

Everything a credit-risk function does over time is unavailable: vintage curves beyond what one console page holds, roll rates between DPD buckets, stage migration matrices, PD backtesting against realised default, decision-outcome drift after a policy change, and any IFRS 9 disclosure that needs a history rather than a snapshot.

The console added in the credit-risk PR reads the OLTP service directly and is capped at 1000 rows on purpose. That is the right shape for a desk view and the wrong one for analysis — the warehouse is the sanctioned bulk path (ADR-0022).

## What it takes (all four, or the sink is green about nothing)

1. `topics:` in `openbank-analytics-sink/src/main/resources/application.yaml` gains `openbank.lending.events`.
2. `openbank-infra/gitops/components/analytics/kafka-analytics-sink-mtls.yaml` gains a `Read, Describe` ACL for that topic. The file's own comment records that a previous change updated one and not the other (#2598/#2629), and the sink then asked for a topic its principal had no Read on.
3. ClickHouse gold views for the credit domain, in the shape `V10`/`V11` already use: decision outcomes over time, stage migration, ECL by period, vintage. New migration file plus the ConfigMap sync the two gates check (`check-clickhouse-ddl-in-configmap.py`, `check-clickhouse-migration-configmap-sync.py`).
[...]

PR #8898, opened 2026-09-05
feat(lending): credit-risk & decisioning console — and the engine evidence that was never persisted

JiRaska:
## Summary

The credit decision engine (ADR-0213) has produced machine-readable evidence for every application since it shipped — outcome, reason codes with the rule that fired, pinned table versions, input snapshot hash — and nothing rendered any of it. A credit-risk analyst had two flat tables of enum strings. This adds a read-only credit-risk console over that evidence and over the IFRS 9 provisioning history, plus a notebook that reads its export.

While integration-testing it, the console came back empty against a real database. Cause: **the engine's evidence was never persisted.** That is fixed here (see "The defect this found").

## Type of change

- [x] fix (bug fix)
- [ ] perf (performance)
- [ ] docs
- [ ] chore (build / tooling)
- [ ] breaking change

## Linked issues

Refs #8893, Refs #8894, Refs #8895, Refs #8896

## The defect this found

`LendingService.claimTransition` writes status plus the three **human** decision fields. The ASSESSMENT leg ran the engine, put the result in the response, emitted `credit.decision.evaluated` — and then claimed the transition through that same method, which does not write any engine column. So `decision_outcome`, `decision_price_band`, `decision_reasons`, `decision_matched_rules`, `policy_versions`, `decision_input_hash` and `decided_engine_at` were **NULL on every row**, from `V11__decision_engine_inputs.sql` onward. A code comment in `claimTransition` described the gap and left it ("pre-existing and is NOT fixed here — it needs its own change and its own test").

Nothing could see it: the API response carried the values (computed in memory), the outbox event carried them, and no reader existed to notice the row did not. The evidence survived only as a Kafka event.

Fixed by `compareAndSetDecision`: the same conditional `where status = :from` claim, writing the evidence in the **same statement** as the transition, so a row can never hold a decided state with no decision behind it. `CreditRiskConsoleIT` is the test: it drives a real application through ASSESSMENT over HTTP and reads it back. Before the fix that read returned `[]` — that is how the defect was found, not by inspection.

## Changes

- **lending-service** — `GET /api/v1/lending/risk/{decisions,decisions/summary,portfolio,policy}`. Read-only; disposal stays in the approval inbox (ADR-0227 D4). Role-gated to the set that may read the ADR-0214 evidence bundle, OPA-annotated with the existing `lending.list` / `lending.read` actions, so no new `role_action_matrix` grant.
- Decision views decode the pinned evidence columns (`DecisionEvidenceCodec`, total by construction). The summary is a database `GROUP BY`, so a figure labelled a total is one (#3294's reasoning); the lists are clamped to 1000.
- `OriginationDecisionService.affordabilityRatios` becomes the single definition of DSTI/DTI, shared with the read side, so a console figure and an engine figure cannot diverge. It also returns total-debt-service DSTI, exposed and labelled as the CNB definition the engine does **not** read (#8894 decides whether that changes).
- Portfolio joins the latest `loan_provisioning` row per loan. A loan the cycle never assessed carries `assessment: null` — never Stage 1 with zero ECL.
- **admin-ui** `/lending/risk` — outcome trend, reason-code Pareto, DSTI×DTI scatter with thresholds **read from `/risk/policy`** rather than hard-coded, engine×disposition override matrix, jurisdiction split, decision tables with a per-rule hit count, IFRS 9 stage mix, DPD buckets, ECL coverage, vintage. CSV/JSON export.
- The page states what it cannot know: code-seeded policy, no-op bureau port (so `EXCLUSION` can never fire), placeholder PD/LGD with the model version shown next to every ECL.
- **docs/notebooks** — `credit-risk-decisioning.ipynb` reads the JSON export (pandas + matplotlib only). The export is the audit boundary: no notebook touches a money-path database.
- Threat-model supplement §10; two engineering notes in the service `CLAUDE.md`.

## Definition of Done

- [ ] Contract tests updated (if public API changed). — no cross-service consumer; these are console reads.
- [x] No new lint warnings.
- [ ] Flyway migration added + rollback note (if DB changed). — no schema change; the columns exist since V11 and were simply never written.
- [ ] Avro schema versioned backward-compatibly (if event changed). — no event change.

## Security checklist

- [x] No new `@SuppressWarnings`, `as Any`, `@Suppress("...")` without justification (state it in this PR). — none added.
- [x] No new third-party dependency, OR: dependency review attached (license, CVE, source). — none; recharts is already a dependency.
- [x] Auth / crypto / payment code changed? → tag `security-review-required`. — no auth code changed; the new surface is role-gated and read-only, and the money-path write path gains only the evidence columns in an existing conditional UPDATE.
- [x] PII path changed? → tag `gdpr-review-required`. — a new read exposes applicant income and affordability inputs to the credit desk; gated to the roles that already read the evidence bundle, capped at 1000 rows, no new storage and no new export target.
- [x] Cardholder data path changed? → tag `pci-review-required`. — no.

## Compliance impact

- [ ] PCI DSS
- [ ] DORA
- [x] GDPR
- [ ] PSD2 / SCA / RTS
- [ ] AML / 5AMLD
- [x] CNB reporting
- [ ] None

**GDPR.** The decisions read returns affordability inputs (income, existing debt service, age, residency, employment tenure) for evaluated applications. Compensating controls: role gate identical to the ADR-0214 evidence bundle, a server-side cap, no new persistence, and an export the analyst performs under their own identity rather than a background job.

**CNB.** The persistence fix restores the decision evidence in the database, which is what a supervisory reconstruction of a credit decision needs; until now only the event stream had it. The console also makes the DSTI definition gap visible and quantified (#8894), and the credit reporting gaps are inventoried in #8895.

## Rollout / Rollback

- Deployment strategy: rolling.
- Rollback plan: revert. The new endpoints are additive reads; the persistence fix writes columns that were NULL, so a rollback stops writing them again and leaves existing rows valid.
- Data migration reversible: N/A (no migration).

Not attached — the page needs a populated book to be worth a screenshot, and this environment's is empty. The render is covered by `lending-risk-page.test.tsx`.

## Reviewer notes

- **Where to look first:** `LendingService.claimDecision` and `DECISION_HQL`. Everything else is a read.
- **Falsification, not assertion.** The IT's decision assertions fail against the old persistence path (that is how the defect surfaced). Two console assertions were sabotage-checked: changing the policy threshold in the fixture reddens the "reads the threshold from policy" test, and widening the override denominator reddens the override-rate test.
- **Verified locally:** lending unit tests 21/21, `CreditRiskConsoleIT` against a real Postgres, ktlint + detekt clean, admin-ui `npm test` (my three files green; nine unrelated files failed on 5s timeouts under parallel load, the known non-determinism of that suite), `check-api-contract.py --base origin/main --enforce` → additive, 1.17.0 → 1.19.0.
- **Spec version:** 1.18.0 is claimed by #8838, so this takes 1.19.0. Whoever lands second re-checks.
- **One baseline entry added:** `ktlint-baseline.xml` gains the `package-name` violation every file in `port/in` carries (Kotlin keyword directory), matching the two entries already there.
- **Deliberately not here:** the warehouse has no lending stream at all (#8893), which is why this console reads the OLTP service directly and is capped.

### DIFF
--- a/docs/notebooks/README.md
+++ b/docs/notebooks/README.md
@@ -1,0 +1,25 @@
+# Notebooks
+
+Analyst notebooks that read an **export**, never a live database. The bank's OLTP databases are
+not an analytics surface (ADR-0022): a notebook pointed at `lending` would hold a session against
+a money-path database, bypass every role gate the service applies, and read PII with no audit
+trail. So the flow is:
+
+1. A credit-risk analyst opens the console (`/lending/risk`) with their own role.
+2. They export the decision set they are entitled to see (JSON or CSV).
+3. The notebook reads that file from disk.
+
+The export is the audit boundary. What is in the file is what the analyst was allowed to read at
+the moment they read it, and the file carries the policy bundle that produced the decisions, so a
+notebook cell can always answer "against which rules?".
+
+| Notebook | Reads | Answers |
+|---|---|---|
+| [`credit-risk-decisioning.ipynb`](credit-risk-decisioning.ipynb) | the console's JSON export | Outcome and reason-code distribution, affordability against the policy's own thresholds, engine-vs-human overrides, IFRS 9 stage/ECL and vintage |
+
+Dependencies are `pandas` and `matplotlib` only — deliberately not a project environment: the
+notebook must open on an analyst's laptop without this repo being built.
+
+**What a notebook must not become.** These are exploratory. A number that has to be defensible to
+the regulator belongs in a service or a regulatory return, where it is versioned, tested and
+gated — never in a notebook cell that only one person has run.
--- a/docs/notebooks/credit-risk-decisioning.ipynb
+++ b/docs/notebooks/credit-risk-decisioning.ipynb
@@ -1,0 +1,308 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Credit risk & decisioning",
+    "",
+    "Reads a JSON export from the admin console (`/lending/risk` \u2192 **JSON**). Nothing here connects to a",
+    "database or a service: the export is the audit boundary, and it carries the policy bundle that",
+    "produced the decisions, so every threshold drawn below is the bank's own rule, not a constant",
+    "typed into this notebook.",
+    "",
+    "**What the data can and cannot say.** The decision fields are the ADR-0213 evidence the engine",
+    "pinned (outcome, reason codes, matched rule ids, table versions, input hash). There is no score",
+    "and no probability \u2014 the engine does not produce one. The IFRS 9 figures come from the scheduled",
+    "provisioning cycle; the PD/LGD behind them is a flat placeholder set (see the model version cell),",
+    "so an ECL here is arithmetic on assumed parameters, not a calibrated loss estimate."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import json",
+    "from pathlib import Path",
+    "",
+    "import pandas as pd",
+    "import matplotlib.pyplot as plt",
+    "",
+    "# The file the console downloaded. Point EXPORT at your own download.",
+    "EXPORT = Path(\"credit-decisions.json\")",
+    "",
+    "payload = json.loads(EXPORT.read_text())",
+    "policy = payload.get(\"policy\")",
+    "decisions = pd.json_normalize(payload.get(\"decisions\", []))",
+    "portfolio = pd.json_normalize(payload.get(\"portfolio\", []))",
+    "",
+    "print(f\"{len(decisions)} decisions, {len(portfolio)} loans\")",
+    "print(\"policy:\", (policy or {}).get(\"source\"), \"code-seeded:\", (policy or {}).get(\"codeSeeded\"))"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## 1. What the engine decides",
+    "",
+    "Outcome mix over the exported set. Remember this is the export, not the book: the console's own",
+    "tiles take book-wide totals from a database aggregate, and the export is capped."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "outcomes = decisions[\"engineOutcome\"].value_counts()",
+    "display(outcomes.rename(\"count\").to_frame().assign(share=lambda d: d[\"count\"] / d[\"count\"].sum()))",
+    "",
+    "ax = outcomes.reindex([\"APPROVE\", \"REFER\", \"DECLINE\"]).fillna(0).plot.bar(",
+    "    color=[\"#22c55e\", \"#f59e0b\", \"#ef4444\"], rot=0, figsize=(6, 3))",
+    "ax.set_title(\"Engine outcomes (exported sample)\")",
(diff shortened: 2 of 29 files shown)


JiRaska:
## Reviewer brief

**Where to start:** `LendingService.claimDecision` and `DECISION_HQL` in `LendingRepositories.kt`. That is the only write in the whole PR. Everything else is a read.

### The defect this found

The ASSESSMENT leg evaluated the application, returned the result in the response, emitted `credit.decision.evaluated` — and then claimed the transition through `compareAndSetStatus`, which writes status and the three **human** decision fields only. So `decision_outcome`, `decision_price_band`, `decision_reasons`, `decision_matched_rules`, `policy_versions`, `decision_input_hash` and `decided_engine_at` were **NULL on every row** from `V11__decision_engine_inputs.sql` onward. The evidence survived only as a Kafka event.

Nothing could see it. The API response carried the values because they were computed in the same call, the outbox event carried them, and until this console there was no reader of the row to notice. The gap was described in `claimTransition`'s own KDoc and left for later — this is that change.

`compareAndSetDecision` is the same conditional claim (`where id = :id and status = :from`) with the evidence written in the **same statement**, so a row cannot hold a decided state with no decision behind it.

### How it was proven rather than asserted

`CreditRiskConsoleIT` drives a real application through ASSESSMENT over HTTP against a real Postgres, then reads it back. Against the old write path that read returned `[]` — that is how the defect surfaced, not by inspection.

Two console assertions were sabotage-checked before being trusted: changing the threshold in the policy fixture reddens the "reads the threshold from policy" test, and widening the denominator reddens the override-rate test.

### What to weigh in review

- **Breadth of one read.** A single call returns income and affordability inputs for every evaluated applicant. Gated to the roles that already read the ADR-0214 evidence bundle, capped server-side at 1000 rows, no new storage and no new export target. Threat-model supplement is §10.
- **The DSTI definition.** The engine's ratio is new installment over verified income only. The console shows the total-debt-service figure beside it, explicitly labelled as the CNB definition the engine does **not** read. Moving the floor to the total would tighten credit, so it is a policy decision (#8894), deliberately not made here.
- **Thresholds come from data.** The scatter overlays are read from `/api/v1/lending/risk/policy`; the page hard-codes no limit.
- **No schema change.** The columns have existed since V11 and were simply never written.

### Verified locally

| Check | Result |
|---|---|
| lending unit tests | 21/21 |
| `CreditRiskConsoleIT` (real Postgres) | 7/7 |
| ktlint, detekt | clean |
| admin-ui, the three new files | 16/16 |
| `check-api-contract.py --base origin/main --enforce` | additive, 1.17.0 → 1.19.0 |

Nine unrelated admin-ui files failed on 5s timeouts under parallel load — the known non-determinism of that suite, not this change.

### CI status at the time of writing

13 SUCCESS · 6 QUEUED · 4 SKIPPED · 3 FAILURE.

All three failures are `Trivy` and the two `dependency review` variants, and all three are the same infrastructure cause, quoted from the Trivy job:

```
##[error]API rate limit exceeded for installation.
```

Trivy scanned the manifests to completion and failed only when uploading its SARIF. None of the three is a required check. All three gate shards are green, which covers the api-contract, threat-model and security-checklist gates under real PR context.

Spec version is 1.19.0 because #8838 already claims 1.18.0 — whoever lands second re-checks.

### Follow-ups filed rather than smuggled in

- #8893 — the warehouse has no lending stream at all, which is why this console reads the OLTP service directly and is capped
- #8894 — should the DSTI floor move to total debt service, and per jurisdiction
- #8895 — F 18/F 19, F 12, C 07.00, CNB DSTI/DTI/LTV limits, AnaCredit LOAN feed
- #8896 — ADR-0213 D4: the policy is Kotlin literals, with one risk appetite for every jurisdiction

Merge is left to a human: the local guard refused it because the PR touches money-path paths.

PR #8904, opened 2026-09-05
feat(libs): declare the settled money-flow features, and the coverage signal that qualifies them

JiRaska:
## Summary

ADR-0282 phase 1 asks for feature definitions for savings rate and spend cadence, computed by the same pure function that serves both stores so the two materialisations cannot skew. Writing them showed the interface could not express them, and that is a measurement rather than a preference.

## Type of change

- [ ] fix (bug fix)
- [ ] perf (performance)
- [ ] docs
- [ ] chore (build / tooling)
- [ ] breaking change

## Linked issues

Refs #8792 · ADR-0140 · ADR-0201 D3 · ADR-0282 phase 1

## Why the interface had to change

`FeatureEvent` carried `(entityId, eventType, occurredAt)`. #8792 requires the fact to be **post-settlement**, and measured on the warehouse 2026-09-05 no single event carries both halves:

| event | says | carries |
|---|---|---|
| `TransactionCompleted` | it settled | a reference number only |
| `TransactionInitiated` | amount, currency, accounts | emitted before the money moves |

A flat event list with no correlation key cannot pair them. So the money features were not inconvenient to write against the old shape — they were inexpressible.

Three optional fields, each null for every event with no such fact, so all three existing construction sites are untouched: `amountMinor` (minor units, never a Double), `correlationId` (the transaction's own aggregate id, carried by both events), `direction` (IN or OUT for this entity, so an internal transfer is correctly both).

## Changes

Five features, declared as one list so a consumer cannot register the flattering half alone.

| feature | what it is |
|---|---|
| `settled_inflow_minor_d90` | time from the settlement, amount from the instruction |
| `settled_outflow_minor_d90` | same, outbound |
| `savings_rate_d90` | `(in - out) / in` |
| `spend_cadence_days_d90` | median gap between settled outbound payments |
| `money_events_without_amount_d90` | the coverage signal |

An instruction that never settled contributes nothing. An instruction older than the window still prices a settlement inside it, or a settled payment would carry no value.

## Undefined is not zero

An entity with no settled inflow has no savings rate, and `0.0` would report "saves nothing" — the least flattering answer for the customer we know least about, in a feature a rewards programme thresholds on. V10's `volatility_ratio` guards the identical divisor with NULL for the identical reason.

`compute` is non-null by contract, so the undefined case is `NaN`: every comparison against NaN is false, so a threshold test **fails closed** and no reward is granted on an unknown. Same choice for a cadence below two settlements, where `0.0` would read as "spends every day".

A negative savings rate is reported, not clamped. Spending more than you received is real, and clamping erases exactly the customers a financial-health programme most needs to see.

## The coverage signal is not decoration

Every flow above drops a settlement it cannot price, and a drop is invisible: the flow reads lower and nothing anywhere disagrees. That is the shape of the push adapter whose skipped result carried `success = true`. `money_events_without_amount_d90` gives the drop a number, so "spends little" and "we could not price this" stop being one reading.

The analogous warehouse join pairs 47 of 47 today, so a non-zero reading is a real regression in the event contract rather than noise.

## Falsified

Replacing both `NaN` returns with `0.0` fails exactly the two fail-closed tests and no others. 13 tests, 2 failed.

## Deliberately absent

Buffer months needs a balance **level** rather than a flow, so it reads `BALANCE_UPDATED` — which does carry `bookedAmount` on all 203 rows measured — and belongs in its own definition rather than bolted onto a flow feature. Category, MCC and merchant are absent from every transaction payload, so no spend-category feature can be declared honestly and none is.

## Definition of Done

- [x] No new lint warnings.

A libs change reaches every dependent, so the only existing feature-store consumer was compiled explicitly: `:openbank-fraud-service:compileKotlin` and `compileTestKotlin` both pass.

## Security checklist

- [x] No new `@SuppressWarnings`, `as Any`, `@Suppress("...")` without justification. Verified against the diff: zero added.
- [x] No new third-party dependency. Verified: no build or lock file in the diff.

These are pure functions over events passed in by a caller; they open no store and read no data of their own. Nothing here names a counterparty or a free-text message.

## Compliance impact

- [ ] PCI DSS
- [ ] DORA
- [x] GDPR
- [ ] PSD2 / SCA / RTS
- [ ] AML / 5AMLD
- [ ] CNB reporting
- [ ] None


JiRaska:
The one red on this PR is `build (openbank-swift-service)`, and it is not about this diff. Filed as #8916 with the measurements.

Three consecutive attempts against the same unchanged commit `fd0229f17`: 45m20s cancelled, 11m21s failed, 45m20s cancelled. `_service-ci.yml` sets `timeout-minutes: 45`, so two of the three are that wall.

`openbank-swift-service` has zero references to `FeatureEvent`, `FeatureDefinition` or `OnlineFeatureStore` — it builds only because a `openbank-libs-domain` change builds the fleet. The other 165 of 166 checks are green.

I stopped at three attempts rather than re-running until green: a job that passes on the fourth try still fails two thirds of the time, and that is the thing worth fixing rather than routing around.

Issue #8919, opened 2026-09-06
SwiftPactFolderProviderVerificationTest takes ~5-6 minutes against an 8-minute budget, and only PRs touching shared files ever run it


JiRaska:
## What was measured

`SwiftPactFolderProviderVerificationTest` — a **pure in-process message-pact verification**, not a
`@QuarkusTest`, with no HTTP, no database and no Testcontainers — takes **minutes**, and its
8-minute per-interaction budget has almost no headroom left.

Two runs of the identical command on this machine, nothing else building:

| branch | result | wall clock |
|---|---|---|
| a `main`-based branch (docs only) | PASS | **4m 54s** |
| `feat/card-center-admin-ui` (ADR-0283 stack) | PASS | **6m 23s** |

In CI the same test fails on two unrelated PRs of that stack with:

```
java.util.concurrent.TimeoutException: verifyPact(...) timed out after 8 minutes
```

So it is not a contract mismatch and not a logic regression — the verification simply does not
finish inside its budget on a loaded runner. The card stack pushes it ~30% slower (it enlarges
`openbank-libs-domain`, which is on every service's test classpath), and that is enough to cross
the line.

## Why nobody has seen this

Path-scoped CI. `main` builds a service only when a PR touches it, and pushes to `main` rarely touch
swift-service — its last Services CI runs on `main` contain **no swift job at all**. The test is
therefore exercised almost exclusively by PRs that touch a SHARED file (an OPA bundle restamp is
enough to pull swift into the changed set), which is how an ADR-0283 stack that never edits swift
ended up owning its timeout.

## Where the time probably goes

The test's own KDoc records that the no-arg `MessageTestTarget()` scans the whole test classpath and
threw `ClassGraphException` on swift-service, and that the fix was to scope the scan:

```kotlin
context?.target = MessageTestTarget(listOf("com.openbank.swift.contract"))
```

The package list narrows what is *reported*, not how much classpath ClassGraph walks. A scan whose
cost is proportional to the classpath explains both the original crash and the current runtime, and
it explains why adding classes to `openbank-libs-domain` moves it.

Worth measuring before fixing: whether the minutes are in the scan or in pact-jvm itself.

## What this issue asks for

1. Measure where the time goes (scan vs verification), rather than raising the timeout, which would
   hide a five-minute unit-style test.
2. If it is the scan, give pact-jvm the producer directly instead of asking it to discover one.
[...]

PR #8934, opened 2026-09-06
fix(swift): give the test JVM the same 2g ceiling account and lending already run

JiRaska:
## Problem

swift-service's test fork runs at Gradle's default 512m heap. The suite boots Quarkus repeatedly and runs `SwiftPactFolderProviderVerificationTest` (every pact naming swift-service as provider, one fork). On 2026-09-06 two unrelated PRs died identically:

- **#8781** — `OutOfMemoryError` across ClassGraph-worker / vert.x-eventloop / Finalizer threads ~3 min in, then GC-spiral hang until the 45-minute job timeout
- **#8697** — same OOM, then `SwiftPactFolderProviderVerificationTest ... status COMPLETED` failed with `TimeoutException` downstream of the GC storm

## Fix

`maxHeapSize = "2g"` on swift's test tasks — the ceiling account-service, lending-service and product-catalog already carry. account-service's comment explicitly asks to be told when a second module OOMs rather than applying an unmeasured fleet-wide ratchet; swift is that measured case. A ceiling is not an allocation.

Refs #8345 (the pact tail #8697 is landing), unblocks #8781 and #8697.

## Security checklist

- [x] No secrets, tokens, passwords, or PII in code, config, tests, logs.
- [x] No new `@SuppressWarnings`, `as Any`, `@Suppress("...")` without justification.
- [x] No new third-party dependency.
- [x] Auth / crypto / payment code changed? → no — test-JVM heap ceiling in a build file only.
- [x] PII path changed? → no.
- [x] Cardholder data path changed? → no.

## Compliance impact

None: CI build-file change (test heap ceiling), no runtime behaviour, no data path, no schema.

PR #8938, opened 2026-09-06
build(swift): give the test JVM 2g so the suite stops OOMing in CI

JiRaska:
## Summary

The `openbank-swift-service` test JVM ran at Gradle's 512m default and exceeded it. Seven open PRs
are currently red on `build (openbank-swift-service)`, and none of them touch this service. This
gives the task `maxHeapSize = "2g"`, matching account-service and lending-service.

## Type of change

- [ ] fix (bug fix)
- [ ] perf (performance)
- [ ] docs
- [x] chore (build / tooling)
- [ ] breaking change

## Linked issues

Refs #8919

## Changes

- `openbank-swift-service/build.gradle.kts`: `maxHeapSize = "2g"` on `tasks.withType<Test>`, with
  the measurement and the reason it is per-module rather than a fleet default.

## What was measured

From the failing job (run 34016248642, `Build + test (:openbank-swift-service)`):

```
09:52:00  job start
09:56:57  Type implements CloseableResource but not AutoCloseable: au.com.dius.pact.consumer.junit5.JUnit5MockServerSupport
09:56:57  Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "idle-connection-evictor-1"
          … 12 more OutOfMemoryError lines through 09:57:53
10:37:13  ##[error]The operation was canceled.
```

Step 10 is the only `cancelled` step; every other step is `success` or `skipped`, and no test
reports a failing assertion. That is why this reads as a cancellation rather than as a memory
fault, and why it was first filed as a slow test.

**The earlier diagnosis in #8919 was wrong and is corrected there.** The 8-minute
`TimeoutException` on `SwiftPactFolderProviderVerificationTest` and this cancellation are the same
defect at two stages: with the heap exhausted, verification advances only between full GCs. The
timing table in that issue was taken on a machine where the heap was never the constraint, so it
does not describe CI.

## Why per-module and not build-logic

`openbank-account-service` states the rule in place:

> Deliberately per-module rather than a fleet default: nothing measures test heap anywhere, so a
> global bump would be an unmeasured ratchet applied to 50 modules to fix one. **If a second module
> starts OOMing, that is the signal to raise it in build-logic** and to count the Quarkus boots per
> module rather than keep paying for them.

lending was the second, swift is the third, so that threshold is met. A fleet default is a
50-module change and wants the boot count the comment asks for; it is left to #8919 rather than
folded in here.

## Definition of Done

- [ ] Unit tests added/updated. — build configuration only; the suite itself is unchanged.
- [ ] `./gradlew detekt ktlintCheck koverVerify build` passes. — an A/B of the full swift suite
      with and without the override is running locally; the numbers go on this PR as a comment
      when it finishes. Opening now because seven PRs are blocked on it.
- [x] No new lint warnings.

## Security checklist

## Compliance impact

- [x] None

## Rollout / Rollback

- Deployment strategy: N/A — build configuration, no shipped artifact changes.

## Reviewer notes

The commit type is `build`, deliberately: nothing shipped changes, so this must not cut a
swift-service release.

### DIFF
--- a/openbank-swift-service/build.gradle.kts
+++ b/openbank-swift-service/build.gradle.kts
@@ -68,6 +68,23 @@ dependencies {
 // probe's answer first.
 // Pact: write generated consumer contracts to pacts/ and forward broker config.
 tasks.withType<Test> {
+    // Gradle's default test-JVM heap is 512m. This module exceeded it: on 2026-09-06 the CI job
+    // OOMed four minutes in, at the moment the pact consumer test starts its mock server
+    // (`JUnit5MockServerSupport`), then thrashed for forty minutes until the job was cancelled —
+    // thirteen `OutOfMemoryError` lines and not one failing assertion, so the job reads as a
+    // cancellation rather than as a memory fault. The earlier reading of the same defect was
+    // `SwiftPactFolderProviderVerificationTest` blowing its 8-minute budget: with the heap
+    // exhausted the verification only advances between full GCs, so it either overruns or dies.
+    // Locally, where the heap is not the constraint, the same test finishes in ~5-6 minutes, which
+    // is why this passes on a laptop and fails in CI on branches that never touch this service.
+    //
+    // `openbank-account-service` and `openbank-lending-service` already carry this override for
+    // the same failure, and account's comment sets the rule: a second OOMing module is the signal
+    // to raise the default in build-logic and to count the Quarkus boots per module. swift is the
+    // third, so that threshold is met — but a fleet default is a 50-module change and belongs in
+    // its own PR with those counts, not here (#8919).
+    maxHeapSize = "2g"
+
     // CI hang #3 (#2320) is GONE — measured, not assumed. `SwiftBootSmokeIT` used to hang 37+ min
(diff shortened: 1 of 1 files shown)


JiRaska:
Superseded by #8934, which merged to `main` while this was open. Not a judgement call — the two
changes are the same edit:

```
$ git show origin/main:openbank-swift-service/build.gradle.kts | grep -n maxHeapSize
81:    maxHeapSize = "2g"
```

Same task (`tasks.withType<Test>`), same value, same reason. #8934's comment cites the same
measurement from the other side — `OutOfMemoryError` ~3 minutes into the suite, then a GC hang
until either the 45-minute job cancellation (#8781) or a downstream pact timeout (#8697) — which
is the same failure I read off the #8855 job log.

Merging this too would put two `maxHeapSize` lines in one `tasks.withType<Test>` block, so it is
closed rather than rebased. Nothing here is lost: the correction to #8919's premise (the pact test
is not slow, the heap is exhausted) is posted on that issue, and it stands independently of which
PR carried the fix.

Issue #8757, opened 2026-09-05
Integrate security-checklist preparation into PR-body producers


JiRaska:
## Remaining work

Integrate the tracked security-checklist preparation helper into the tools that compose PR body files, and prove that integration at the producer boundary. Keep this issue open until adoption is demonstrated.

## Already delivered

- #9930 added `.github/scripts/append-security-checklist.py` and its enforced self-test gate. It copies the current PR template section, leaves boxes unticked and preserves an existing section byte-for-byte.
- #9909 added remediation guidance to the checker and contributor instructions.
- #8940 made the security-checklist checker read the live PR body. Correcting a body no longer needs an empty commit; the next gate execution can read the updated description.
- The helper's eight self-test cases passed again against the merged implementation on 2026-09-14, including actual template extraction, idempotence and untouched checklist attestations.

## Why the integration is still needed

`gh pr create --body-file` does not automatically apply GitHub's PR template. The body-file convention remains correct for shell safety, but the producer must explicitly include the checklist. Shipping a helper alone does not establish that any producer invokes it.

The original 2026-09-05 measurement (50 of 67 ready PRs lacking the section) is historical exposure, not a current count or proof that all those PRs were blocked. The gate applies to money-path changes; the original report did not classify every missing section by scope. Earlier comments retain the measurements and original diagnosis.

## Acceptance

- Inventory the actual PR-body producers, including any machine-local `/open-pr` implementation and tracked automation that creates money-path PRs. An unavailable local skill is unverified, not proof that no producer exists elsewhere.
- Invoke `python3 .github/scripts/append-security-checklist.py <body-file>` before PR creation, preserving the body-file workflow.
- Add a producer-level regression covering a body without a checklist, a body with an existing checklist and preservation of unchecked boxes. Do not automatically attest that security work was performed.
- Demonstrate a prepared PR body from the integrated producer, with author-substantiated checklist completion where required.

Refs #8590.

JiRaska:
Related half of this trap is now fixed: #8940 (merged 2026-09-06) makes the security-checklist gate read the LIVE PR body via the pulls API instead of the frozen github.event.pull_request.body snapshot — so 'edit the body to add the checklist' now takes effect on the next gate run without an empty commit (four PRs needed that dance on 2026-09-06: #8011, #8310, #8334, #8934). The first half (the mandated --body-file never applies the template, so the section is absent from the start) still stands — that one is a workflow/docs decision.

[...]

PR #8940, opened 2026-09-06
fix(ci): read the live PR body in the security-checklist gate, not the frozen event payload

JiRaska:
## Problem

`PR_BODY` is `github.event.pull_request.body` — frozen at the last synchronize event. The documented remediation for a missing `## Security checklist` is to edit the PR body, but the gate kept reading the frozen snapshot: measured on #8934, the body was edited at 11:45Z and the gate run at 11:46Z still failed on the old text. The only way through was an otherwise-pointless empty commit to re-emit the event — four money-path PRs needed that dance today (#8011, #8310, #8334, #8934).

## Fix

The security-checklist gate now fetches the **live** body via the pulls API (one contents-read call per shard run, using the `GH_TOKEN` the job already carries) and falls back to the frozen payload only when the API is unreachable.

Refs #8757 — this removes the second half of that trap ("I added the checklist and the gate still fails"). The first half (`--body-file` never applying the template) remains and is a workflow/docs decision, not a gate bug.

## Security checklist

- [ ] Auth / crypto / payment code changed? → no — CI gate input plumbing only.
- [ ] PII path changed? → no.
- [ ] Cardholder data path changed? → no.

## Compliance impact

None: changes which PR-body snapshot a CI gate reads; no product code, no data path.

### DIFF
--- a/.github/gates/gates.yaml
+++ b/.github/gates/gates.yaml
@@ -4034,7 +4034,14 @@ gates:
     selftest_expect: pass
     selftest_inputs: [".github/scripts/check-security-checklist.py", ".github/PULL_REQUEST_TEMPLATE.md"]
     run: |
-      python3 .github/scripts/check-security-checklist.py --body "${PR_BODY}" --base "${PR_DIFF_BASE}" --enforce
+      # PR_BODY is github.event.pull_request.body — frozen at the last synchronize event.
+      # Editing the body to add the checklist (the documented remediation) only reached this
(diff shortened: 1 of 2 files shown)


JiRaska:
Merging over `Dependency review`/`dependency-review`/`Trivy` failures that are environmental: the dependency-graph base snapshot is missing fleet-wide (#8738, fix in #8781) and the runs coincide with the installation rate-limit exhaustion tracked in #8690. All gates, including the checklist gate this PR changes, pass.

PR #8959, opened 2026-09-06
fix(notification): renumber the duplicate V14 to V15, and let the gate allow that

JiRaska:
main has **two V14 migrations** in notification-service:

| file | landed |
|---|---|
| `V14__synthetic_outbox_taint.sql` | 2026-08-24 (#6731) |
| `V14__notification_deduplication_key.sql` | 2026-09-06 (#8334) |

Flyway keys on the version, so once the older one is applied the newer is refused —
`Detected resolved migration not applied to database: 14` — and **notification-service fails
startup on every boot** (#5628).

Found while rebasing #8874, which inherited the red without touching a single notification-service
file.

## Why renumber rather than out-of-order

Confirmed with the service owner that #8334's migration has **not** been applied to any live
database. That is what makes renumbering correct: nothing has checksummed it yet.
`QUARKUS_FLYWAY_OUT_OF_ORDER=true` — the remedy the gate suggests — would permanently loosen a
safety property for the whole service to paper over a one-off collision.

## The blocker was two enforced gates contradicting each other

- `check-flyway-version-commit-order`: *"renumber it to the next free version"*
- `check-db-migration`: refuses **any** rename of a committed migration as an edit

Once a colliding migration reaches main, the remedy one gate demands is the thing the other
forbids, and the service stays unbootable with no legal way out.

So the rename ships with the narrowest carve-out that breaks the deadlock:

- **`R100` only** — git's byte-identical score, so it is a renumber and cannot smuggle in an edit
- **same service directory** — Flyway history is per service
- **the old version must genuinely collide** at base
- **the target version must be free** at base

**Not claimed:** whether the migration already reached a live database. Not machine-checkable, so
the gate prints the condition it is relying on rather than pretending to have verified it.

## Falsified, not asserted

Six cases, each removing one condition. Deleting the `R100` check makes **exactly one** fail:

```
::error::self-test: rename an edited rename is refused — R099 is not a pure renumber: expected False, got True
self-test FAILED (1 case(s))
```

Restored → green. The carve-out can say no, which is what separates it from a hole.

## Verified

- `check-db-migration.py --self-test` → 24 cases green
- `--base origin/main --enforce` → exit 0, printing why the rename was permitted
- `check-flyway-version-commit-order.py` → clean, 433 migrations across 57 services

## Blocks

#8874 and #8909 both inherit this red through main. Neither touches notification-service.


JiRaska:
Superseded: the V14→V15 renumber, the dedup-consumer fix, the baseline cleanup and the out-of-order flag removal all landed in #8953 (merged); the duplicate-version detection lives on in #8960 (rebased). Closing to keep one canonical change.

Issue #7645, opened 2026-08-31
ClickHouse warehouse has no idempotent migration mechanism: the init ConfigMap only runs on an empty data dir, so every post-boot schema change is inert on the live cluster


JiRaska:
## What

The ClickHouse warehouse has **no migration mechanism**. Its schema is applied by a ConfigMap mounted
into `/docker-entrypoint-initdb.d`, which ClickHouse executes *only when the data directory is empty*.
Every schema change made after the cluster's first boot is therefore inert, and three have
accumulated unapplied.

`openbank-infra/gitops/components/analytics/clickhouse.yaml` states the mechanism plainly in its own
header:

> The schema (V1 bronze/silver + V2 onboarding funnel gold views) is applied on first boot via the
> `clickhouse-init` ConfigMap mounted into `/docker-entrypoint-initdb.d` — ClickHouse runs those .sql
> files once, when the data dir is empty.

## The divergence

`openbank-analytics-sink/src/main/resources/clickhouse/` — the source of record — holds eleven files:

```
V1__analytics_bronze_silver.sql   V7__fold_aggregate_type_case.sql
V2__onboarding_funnel.sql         V8__fold_silver_layer_case.sql
V3__screen_feedback.sql           V9__party_credit_profile.sql
V4__screen_feedback_context.sql   V9__synthetic_provenance.sql
V5__party_accounts.sql            V10__credit_funnel.sql
V6__campaign_engagement.sql
```

`clickhouse-init-configmap.yaml` — what actually runs — stops at eight:

```
01-bronze-silver.sql  02-onboarding-funnel.sql  03-screen-feedback.sql
04-screen-feedback-context.sql  05-party-accounts.sql  06-campaign-engagement.sql
07-fold-aggregate-type-case.sql  08-fold-silver-layer-case.sql
```

Nothing scans the directory. Verified by content rather than by counting keys —
`grep -c synthetic` over the ConfigMap returns **0**.

## What it costs

`V9__synthetic_provenance.sql` adds `bronze_events.synthetic UInt8 DEFAULT 0` and redefines
`silver_current_state` and `gold_daily_event_volume` with `WHERE b.synthetic = 0`.

None of that is deployed. In the live warehouse the column does not exist, so **synthetic and test
events are counted in every silver and gold aggregate** — including the onboarding-funnel board the
admin cockpit reads. `V10__credit_funnel.sql` is likewise absent.

## Why it will not self-correct

[...]

PR #8962, opened 2026-09-06
fix(notification): renumber the unapplied V14 — two migrations claimed version 14

JiRaska:
## Summary

`openbank-notification-service` has two migrations numbered 14 on `main`. Flyway refuses to start over that, so the next deploy of that service would crashloop — and the enforced `flyway-version-commit-order` gate is red on **every open PR** until it is fixed.

One file rename. The reasoning is the part worth reading.

## Type of change

- [x] fix (bug fix)
- [ ] perf (performance)
- [ ] docs
- [ ] chore (build / tooling)
- [ ] breaking change

## Linked issues

Refs #5628

## The collision

```
V14__synthetic_outbox_taint.sql          reached main 2026-08-24
V14__notification_deduplication_key.sql  reached main 2026-09-06 (24e15de9)
```

Both in `openbank-notification-service/src/main/resources/db/migration/`. This is the war story the root `CLAUDE.md` already records: two branches claim the same number, the merge is textually clean, nothing conflicts, and the service then refuses to boot.

**The running pod is unaffected** — `notification-service` has been up 3h21m from an earlier image. That is the usual shape of this fault: the bomb is in git and the cluster is calm, so it surfaces at the next deploy or restart rather than now.

## Why NOT the remedy the gate prescribes

The gate's own message says:

> set `QUARKUS_FLYWAY_OUT_OF_ORDER=true` in that service's gitops Deployment env … then add an entry to this script's `KNOWN_VIOLATIONS`.

That is correct for a **lower** version arriving late, which is what out-of-order execution exists to allow. It cannot fix an **equal** one: duplicate versions are a different Flyway error (`Found more than one migration with version 14`), raised while resolving migrations, before ordering is even considered. Following the suggestion would have left the service unbootable **and** the gate green — strictly worse than the red gate.

The gate's message conflates the two cases ("lower-or-equal"). Worth splitting, but not in this PR; noted below as a follow-up.

## Why renumbering is safe here — measured, not assumed

Renaming a migration is dangerous only once it has been applied, because Flyway checksums the file. Read from the live sandbox database:

```
installed_rank | version | description            | success | installed_on
            14 | 14      | synthetic outbox taint | t       | 2026-08-24 14:00:01
            13 | 13      | notification failure reason | t  | 2026-08-09 19:27:39
```

Version 14 is recorded as the **taint** migration. `V14__notification_deduplication_key.sql` appears nowhere in `flyway_schema_history`, so it has never been applied in this environment and is free to take the next number. That is exactly the condition `CLAUDE.md` attaches to renumbering.

## Verified

| Check | Result |
|---|---|
| `check-flyway-version-commit-order.py --enforce` after the rename | exit 0, 433 migrations across 57 services, clean |
| Same gate with the collision restored | exit 1, names `V14__notification_deduplication_key.sql` |
| References to the old filename in code, config or docs | none |

## Definition of Done

- [x] Hexagonal layering preserved. — N/A, a file rename.
- [ ] Unit tests added/updated. — the gate falsification above is the evidence; there is no code.
- [ ] Integration tests added/updated. — N/A.
- [ ] Contract tests updated. — N/A.
- [x] `./gradlew detekt ktlintCheck koverVerify build` passes. — no code changed.
- [x] No new lint warnings.
- [ ] OpenAPI spec regenerated. — N/A.
- [x] Flyway migration added + rollback note. — no new migration; an existing unapplied one is renumbered. Rollback is renaming it back, which restores the collision.
- [ ] Avro schema versioned. — N/A.
- [x] Docs updated. — the commit message carries the reasoning.

## Security checklist

- [x] No new `@SuppressWarnings` / `@Suppress` without justification. — none.
- [x] No new third-party dependency. — none.
- [x] Auth / crypto / payment code changed? — no.
- [x] PII path changed? — no.
- [x] Cardholder data path changed? — no.

## Compliance impact

- [ ] PCI DSS
- [x] DORA
- [ ] GDPR
- [ ] PSD2 / SCA / RTS
- [ ] AML / 5AMLD
- [ ] CNB reporting
- [ ] None

**DORA.** An unbootable service on the next deploy is an availability fault in a tier-classified service. This removes it before it can be triggered.

## Rollout / Rollback

- Deployment strategy: rolling. The migration is unapplied everywhere, so the first deploy after this merge applies it as version 15.
- Rollback plan: rename back — which restores the collision, so the real rollback is reverting whatever needed the migration.
- Data migration reversible: the migration itself is unchanged; only its version number moves.

## Reviewer notes

- **Check the premise, it is one query:** `SELECT version, description FROM flyway_schema_history ORDER BY installed_rank DESC` on `notifications-db-1`. If version 14 there is *deduplication key* rather than *synthetic outbox taint*, this fix is wrong and the answer is a Flyway repair instead.
- **Two follow-ups I did not smuggle in:** the gate should report an EQUAL version as its own finding with its own remedy, since the current text sends you to a fix that cannot work; and its message references `check-db-migration.py`, which does not exist in `.github/scripts/`.


JiRaska:
Closing as a duplicate of #8960, which is the better change.

#8960 was opened at 15:25, this one at 15:35 — ten minutes apart, both renaming the same file to `V15__notification_deduplication_key.sql`. I did not check the open PR list before starting, which is the parallel-work trap this repo documents; I avoided it on the issue side an hour earlier (#7645) and walked into it here.

**#8960 is strictly larger and fixes what I explicitly deferred.** Beyond the rename it makes the gate report an EQUAL version as its own finding rather than folding it into the order message, and it teaches `check-db-migration.py` that a renumber is not an edit. Those are the exact two follow-ups I listed in this PR's reviewer notes as out of scope.

One thing measured here that may still be useful, recorded so it does not die with this PR: on the live sandbox, `flyway_schema_history` in `openbank_notifications` has version 14 as **synthetic outbox taint**, applied 2026-08-24, and the deduplication-key migration appears nowhere. That is the fact which makes renumbering safe rather than a checksum hazard, and it is a measurement of the database rather than an inference from git.

Also confirmed while investigating: the `openbank-notification-service` build failure is not caused by either PR. It fails on `main` too — at `ded789d3` four integration tests fail (`DeviceTokenSweepCronIT`, `ApprovalStoreWiringIT`, `OperatorMessageNullVariableIT`, `OversightWebhookIT`). #8964 already covers the deduplication test itself, with a sharper diagnosis than I had: 18 tests versus 212 is the difference between an application that could not boot and one that ran.

Issue #4363, opened 2026-08-09
A push to a party with no device is dropped, not re-routed: decide the per-template fallback policy


JiRaska:
## The gap

When a party has no ACTIVE device token, `NotificationConsumer.sendPush` marks the notification `FAILED` with `no_active_device` and **drops the message**. There is no second attempt on another channel, and no other component picks it up.

Measured 2026-08-08: 13 of 15 fully onboarded customers have no registered device (#4353). For them, every push-channel notification the platform has ever produced was written to a table and thrown away.

## Why this was deliberately not bundled with the observability work

#4349 and #4362 make the failure legible and alertable; #4351 made the channel observable; openbank-app#432 fixes the registration side. None of them changes what happens to a message that cannot be delivered — because that is a **customer-visible behaviour change** and deserves its own review rather than riding along in an observability PR.

## What needs deciding, not just implementing

1. **Which templates may fall back.** `SCA_APPROVAL` should almost certainly *not*: the push is the mechanism, and an emailed "a payment awaits approval" is both weaker and a phishing template. `TRANSACTION_COMPLETED` and the informational set are the obvious candidates.
2. **Which channel.** `IN_APP` is the conservative choice — no external egress, no new PII surface, and the app already fetches detail on tap via `GET /api/v1/notifications/{id}/self`. EMAIL reaches someone who never opens the app but re-opens the ADR-0135 §3 question of what may appear in the body.
3. **Consent.** The ADR-0198 D4 gate is per-channel (`marketingScopeFor`). A fallback that silently moves a MARKETING push onto EMAIL would bypass a channel-specific consent decision — the gate must be re-evaluated for the fallback channel, not inherited.
4. **Status semantics.** If a push fails and an in-app copy succeeds, what is the row? Probably a new row on the fallback channel plus the original left `FAILED` with its reason — one row cannot honestly carry two outcomes, and `failure_reason` (#4349) exists precisely so the original stays interpretable.

## Suggested shape

A per-template fallback policy in the domain model (closed, like the template enum itself) rather than a config flag, so adding a template forces the decision to be made and reviewed. Not a blanket "on failure, try email".

Related: #4353 (why so many parties have no device), #4349, #4362, ADR-0252, ADR-0253 (a control must be able to show it carried work — a dropped message currently carries nothing).

JiRaska:
## This is a product decision, and two of its inputs turn out to be blockers — measured, not assumed

[...]

PR #8963, opened 2026-09-06
fix(notification): renumber the duplicate V14 migration and correct the two gates that misdiagnosed it

### LINKED ISSUE #8961: main is red: two V14 migrations in notification-service trip the enforced flyway-version-commit-order gate
## What

`main` is red on the enforced `flyway-version-commit-order` gate, and has been since two `V14` migrations landed in `openbank-notification-service`:

```
openbank-notification-service: V14__notification_deduplication_key.sql (version 14) reached main AFTER
V14__synthetic_outbox_taint.sql (version 14) but carries a lower-or-equal version number.
```

Measured 2026-09-06 on the latest `main` CI run (34040576396, job `gates (registry-kotlin-data)`, 77/84 FAIL) and reproduced identically on an unrelated docs PR branch (run 34040887531). Every PR inherits it, so the red says nothing about the PR that carries it — which is the expensive part: a genuine finding in that gate is now indistinguishable from this one.

## Why it matters beyond CI

The gate's message is about the deployed pod, not the build: Flyway refuses the later-arriving lower version with `Detected resolved migration not applied to database: 14` and the service then crashes on every boot once the other V14 is applied. Nobody has checked here whether notification-service's live database has already applied one of them — that determines which of the two fixes is available.

## The two fixes, per the gate's own message

- Still open in a PR: renumber to the next free version.
- Already on `main` (this case): renaming is blocked by `db-migration-gate`, so set `QUARKUS_FLYWAY_OUT_OF_ORDER=true` in that service's gitops Deployment env (the `components/campaign/campaign-service.yaml` pattern) and add a `KNOWN_VIOLATIONS` entry to `.github/scripts/check-flyway-version-commit-order.py`.

Filed from an unrelated PR that this red is blocking; not picking the fix here, because which one is correct depends on the live database state.

JiRaska:
## Summary

`openbank-notification-service` carried two migrations at version 14, which stops Flyway before it reads the database at all, so the next deploy of that service would not have started. This renumbers the one that has never been applied, and corrects the two gates that misdiagnosed the case.

## Type of change

- [x] fix (bug fix)
- [ ] perf (performance)
- [ ] docs
- [ ] chore (build / tooling)
- [ ] breaking change

## Linked issues

Closes #8961

## Changes

- `V14__notification_deduplication_key.sql` becomes `V15__notification_deduplication_key.sql`. Content unchanged, including its rollback note.
- `check-flyway-version-commit-order.py` gives the duplicate-version case its own message. The old one was written for an out-of-order migration and prescribed `QUARKUS_FLYWAY_OUT_OF_ORDER` plus a baseline entry, which governs order and does nothing about duplicates. Following it would have left the service unable to boot while the gate went green.
- `check-db-migration.py` gains `resolves_version_collision()`. It previously classified every rename as an edit of an applied migration, which made the only available repair impossible: with a duplicate version there is no "add V(n+1)", because the pair itself is the error. The exception is narrow and decidable from the tree alone, with no database access: the old version must genuinely be duplicated in that service at the PR base, and the new version must be free there.

## How this was established

Which of the two is safe to renumber came from the live cluster, not from the git history:

```
notifications-db-1: select version, script from flyway_schema_history order by installed_rank desc limit 1;
14 | db/migration/V14__synthetic_outbox_taint.sql   (installed 2026-08-24)
```

The deduplication-key migration appears nowhere in that table, so it has never been applied and renumbering it changes no deployed database.

Flyway's behaviour was measured, not assumed. Both halves ran against this repo's own migration files in a throwaway Postgres 18.1 with Flyway 11:

```
# all 15 files as they are on main
ERROR: Found more than one migration with version 14
Offenders:
-> /flyway/sql/V14__synthetic_outbox_taint.sql (SQL)
-> /flyway/sql/V14__notification_deduplication_key.sql (SQL)

# the deployed state (14 files, no dedup migration), then the renumbered file added
Successfully applied 14 migrations to schema "public", now at version v14
Successfully applied 1 migration to schema "public", now at version v15
15 V15__notification_deduplication_key.sql success=true
uq_notifications_deduplication_key
```

That first result is what makes the old remedy wrong: the resolver refuses the set, so an ordering flag has nothing to act on.

Both gate changes were falsified before being trusted:

- Forcing `resolves_version_collision()` to always allow reddens five must-refuse self-test cases; forcing it to always refuse reddens the one must-allow case.
- Removing the duplicate branch from the order gate reddens its new self-test, which also asserts the strictly-lower case keeps its own out-of-order remedy rather than inheriting the new wording.
- Running `main`'s version of `check-db-migration.py` against this branch fails with the forward-only error, so the exception is load-bearing rather than decorative.

## Definition of Done

- [x] No new lint warnings.

No Kotlin changed, so the Gradle row does not apply here; the two gate self-tests plus the two live gate runs are the verification. The migration keeps the rollback note it shipped with, and the gate confirms it.

## Security checklist

One of the two changes relaxes a gate, so it is worth stating plainly what it can and cannot let through: only a rename inside one service, only when that version is duplicated at the base, only onto a free version. An ordinary renumber of an applied migration is still refused, and that is one of the must-refuse self-test cases.

## Compliance impact

- [ ] PCI DSS
- [ ] DORA
- [ ] GDPR
- [ ] PSD2 / SCA / RTS
- [ ] AML / 5AMLD
- [ ] CNB reporting
- [x] None

## Rollout / Rollback

- Deployment strategy: rolling, with the next notification-service image. The migration applies on boot as `V15`.
- Rollback plan: the migration carries its own note (`DROP INDEX`, `DROP COLUMN`). Reverting this PR alone would restore the duplicate pair and stop the service from starting, so a revert has to renumber instead.
- Data migration reversible: yes, per the note in the file.

## Reviewer notes

The running pod is on `sandbox-cf05a32f`, built before the second V14 merged, which is why nothing is failing right now. The breakage is scheduled for the next deploy of this service, not live. Worth a second opinion on the width of the rename exception and on whether the `KNOWN_VIOLATIONS` entries in the order gate should be re-examined now that the two cases are distinguished.


JiRaska:
Rebased onto `main` after #8967 merged, and this PR now also reverts that change. Worth stating why plainly, because #8967 did exactly what the gate's message told it to.

#8967 set `QUARKUS_FLYWAY_OUT_OF_ORDER=true` and baselined the file. Measured against Flyway 11 with this repo's own migration files in a throwaway Postgres 18.1:

```
ERROR: Found more than one migration with version 14
Offenders:
-> /flyway/sql/V14__synthetic_outbox_taint.sql (SQL)
-> /flyway/sql/V14__notification_deduplication_key.sql (SQL)
```

The resolver refuses the set before it reads the database, so an ordering flag has nothing to act on. After #8967 the gate reported green and the service still could not boot, which is a worse state than the red it replaced. That is the reason the gate's message is corrected here rather than only the symptom.

Two more defects surfaced once the boot-blocker was gone and this module's tests could actually run for the first time since the collision landed:

- The deduplication guard could never fire. The service persists through Hibernate Reactive, so a unique violation is not a `PSQLException`, and the type filter matched nothing. Every redelivery was nacked rather than skipped. CI evidence on this branch: the first run timed out waiting for an ack it was never going to get.
- The redelivery test asserted one row for the party, while one delivery legitimately leaves two: the PUSH row and the no-device EMAIL reroute (#4363), which this test class enables. It now asserts what its name claims, one row carrying the deduplication key, plus that the redelivery adds no row at all.

`build (openbank-notification-service)` is green on the current head. Both gates were run locally against the real tree and both self-tests were falsified in both directions before being trusted.

JiRaska:
Closing as superseded. #8953 and #8960 landed the same repair about half an hour before this PR's checks finished, and #8961 is closed by them. Nothing here is missing from `main`, so merging would put two copies of one fix in the tree.

Compared file by file against `main` rather than by title:

- The migration is renumbered to `V15__notification_deduplication_key.sql`.
- The `QUARKUS_FLYWAY_OUT_OF_ORDER` env added by #8967 is gone from the notification-service manifest, and the baseline list is back to its two long-standing entries.
- The ordering gate distinguishes the duplicate-version case, and the db-migration gate permits the renumber that repairs one.
- The deduplication guard no longer filters on `PSQLException`. `main` matches `java.sql.SQLException` on sqlState 23505 plus the constraint name in the message; this branch matched Hibernate's `ConstraintViolationException.constraintName`. Different route, same defect closed, and `main`'s is already merged.
- The redelivery test asserts the dedup-keyed fact and row stability instead of a total row count, with the same reasoning about the EMAIL fallback row.

One thing worth keeping from this branch's evidence, since neither merged PR states it: the two remedies are not interchangeable, and the wrong one is silent. Measured against Flyway 11 with this repo's own migration files, a duplicate version is refused before the database is touched — `Found more than one migration with version 14` — so `QUARKUS_FLYWAY_OUT_OF_ORDER` cannot help, and #8967 left the gate green while the service still could not boot. The live database is what settled which file was safe to renumber: `flyway_schema_history` in notifications-db records version 14 as `V14__synthetic_outbox_taint.sql`, and the deduplication-key migration appears nowhere in it.

PR #8966, opened 2026-09-06
fix(notification): the dedup guard matched an exception the reactive stack never throws (partial — does not yet make the test pass)

### LINKED ISSUE #8964: notification-service: the deduplication redelivery test fails once the service can actually boot
`NotificationConsumerIT` has a failing test that has never actually run in CI, because the defect
that hid it and the defect it tests arrived in the same PR (#8334).

**Measured, not inferred.** On `main` at `e93b59f0a`:

```
NotificationConsumerIT > an uncorrelated send still emits an outcome, with a null correlation id() FAILED
18 tests completed, 1 failed, 17 skipped
```

Seventeen skipped is the signature of an application that did not start: `main` carries two
migrations claiming version 14, so Flyway refuses the set and every `@QuarkusTest` is skipped rather
than run. With the duplicate renumbered (#8960) the same module runs for real:

```
NotificationConsumerIT > same durable first-use fact is persisted only once across redelivery() FAILED
212 tests completed, 1 failed
```

212 instead of 18. The failure is `java.util.concurrent.TimeoutException at
NotificationConsumerIT.kt:147` — `consumeAndAwait` waits 20 s for the connector's ack and does not
get one.

**Scope.** #8960 deliberately does not touch it: it unblocks the fleet (the duplicate-version gate
was failing on every open PR) and leaves this defect visible instead of merged-over. Fixing the
deduplication behaviour is a separate change against the feature #8334 introduced, and belongs with
whoever owns it.

**Worth carrying beyond this bug:** #8334 was merged with `all-green` and
`build (openbank-notification-service)` both red — red for the very defect it was introducing. No
gate can prevent a merge that overrides a red required context; the duplicate-version detector added
in #8960 only makes the *next* one of these impossible to merge green.

JiRaska:
## Summary

The notification redelivery guard could never fire: it matched the wrong exception type, so every
redelivered notification failed as though the duplicate were a real fault. Fixes #8964, the defect
that #8960 made visible.

## Type of change

- [x] fix (bug fix)
- [ ] perf (performance)
- [ ] docs
- [ ] chore (build / tooling)
- [ ] breaking change

## Linked issues

Closes #8964 · builds on #8960

## Changes

`isDeduplicationConflict()` walked the cause chain for `org.postgresql.util.PSQLException` and read
`serverErrorMessage.constraint`. This service runs **Hibernate Reactive over the Vert.x PG client**,
so the JDBC driver's exception is not in the chain at all:

```
org.hibernate.exception.ConstraintViolationException: ... duplicate key value violates
  unique constraint "uq_notifications_deduplication_key"
    at org.hibernate.reactive.pool.impl.SqlClientConnection.convertException
    at io.vertx.core.impl.future.FutureImpl$4.onFailure
```

The filter matched nothing, `recoverWithUni` never ran, and the redelivery path failed — the exact
case the unique index and this guard exist to absorb.

Now matched on what the reactive stack produces: `PgException.constraint`, then
`ConstraintViolationException.constraintName`, with the constraint **name** in the message as a last
fallback. Deliberately still only that one index: recovering any other unique violation would
silently drop a notification nobody meant to skip. The dead `PSQLException` import is removed rather
than left behind — a guard that cannot fire is worse than none, because it reads as protection.

**Why nothing caught it — corrected.** My first version of this paragraph claimed the service could
not boot and that its tests were reported as SKIPPED, citing `18 tests completed, 1 failed, 17
skipped` against `212`. That comparison was wrong: the 18 came from a LOCAL run with no OIDC, Kafka
or Testcontainers infrastructure available, and the 212 from CI. Measured properly, both from CI:

| where | result |
|---|---|
| `main`, run 34036121346 | `212 tests completed, 5 failed, 30 skipped` |
| #8960 (duplicate renumbered) | `212 tests completed, 1 failed` |

So notification-service **does** boot on `main` despite the two V14 files, and this test was not
hidden by a startup failure — it was one of five already failing there when #8334 merged with
`all-green` red. The renumber in #8960 clears four of them; this PR is about the fifth.

This branch carries the renumber from #8960 so the service can boot at all; #8960 is the PR that
fixes that for the fleet, and this one should land after it.

## Definition of Done

- [x] Unit tests added/updated. (The existing `NotificationConsumerIT` case is what this makes pass;
      it could not run before.)
- [x] No new lint warnings.

**Verification status, stated honestly:** the local `NotificationConsumerIT` run was still going when
this PR was opened and when this correction was written, so what rests on evidence is the diagnosis
— the stack trace above shows `ConstraintViolationException` over `PgException` with no
`PSQLException` anywhere, which the old filter could not match. A green run is not yet claimed. The
same test does execute locally (it failed there with `TimeoutException` before this change), so the
local run is a real signal rather than an environment artifact; CI on this PR is still the
authority.

## Security checklist

- [x] No new `@Suppress` of any kind.
- [x] No new third-party dependency. (`io.vertx.pgclient` is already on the runtime classpath — it
      is the driver this service uses; `account-service` and `billing-service` import it the same
      way.)
- [x] PII path changed? → tag `gdpr-review-required`. (Notification rows carry recipient data; this
      changes only which failures are recovered, not what is stored.)

## Compliance impact

- [ ] PCI DSS
- [x] DORA
- [ ] GDPR
- [ ] PSD2 / SCA / RTS
- [ ] AML / 5AMLD
- [ ] CNB reporting
- [ ] None

At-least-once delivery means a security notification can be redelivered indefinitely. With the guard
inert, each redelivery failed and was retried, so the customer-facing behaviour depended on the
consumer's failure handling rather than on the database fact that was designed to be the authority.

## Rollout / Rollback

- Deployment strategy: rolling.
- Rollback plan: revert; the guard returns to being inert.
- Data migration reversible: N/A — the renumber carried from #8960 changes a filename, not content.

## Reviewer notes

Worth checking that the message fallback is narrow enough for you: it matches only when the
constraint's own name appears in the text, so an unrelated unique violation still propagates.


JiRaska:
**Converted to draft: the diagnosis holds, the fix does not finish the job.**

Ran `NotificationConsumerIT` locally with this change. The failure **moved**, which is evidence the
exception-matching defect was real:

| branch | failure |
|---|---|
| without this change | `TimeoutException` at `NotificationConsumerIT.kt:147` — no ack, no nack; the handler hung |
| with this change | `AssertionFailedError` at `NotificationConsumerIT.kt:266` — `expected: 1L but was: 2L` |

So both deliveries are now acked cleanly, and the test gets as far as its real assertion. But **two
rows are persisted**, which means the unique constraint never fired at all — the recovery path this
PR fixes is therefore not even reached on the second delivery.

What that narrows it to, and what I have NOT established:

- the test uses a **fresh Testcontainers Postgres** (`PostgresTestResource`), with
  `flyway.migrate-at-start: true` and `hibernate-orm.database.generation: none` — so migrations do
  run and the partial index should exist;
- `NotificationEntity.deduplicationKey` is mapped to `deduplication_key`, and the request DTO
  carries the field;
- yet no `23505` occurs. Either the index is not present in that database, or the second insert
  carries `deduplication_key = NULL` — and the index is partial
  (`WHERE deduplication_key IS NOT NULL`), so a null makes it silently inapplicable.

Which of those two it is needs a look at the actual database during the run (the index, and the
second row's key), not more code reading. I stopped rather than guess: each local run of this class
currently takes ~1.5–2 h on this machine.

The change in this PR is worth keeping either way — a guard that matches `PSQLException` in a
reactive service can never fire, and leaving it in place would keep reading as protection — but it
does not close #8964 on its own, so the issue stays open and this stays a draft.

JiRaska:
Closing as superseded: the fix is already on `main`.

Measured in a clean worktree at `b79e4990` — `NotificationConsumerIT`: **18 tests, 0 failures, 0 skipped**, including `same durable first-use fact is persisted only once across redelivery`, the test that was failing this afternoon. Zero skipped matters as much as zero failures: it proves the application actually booted, so the duplicate-V14 collision is genuinely gone rather than hiding the run.

`main` carries a third variant of the guard — it walks the cause chain for `SQLException` and requires BOTH `sqlState == 23505` AND the constraint name in the message. That is narrower than matching the message alone and more robust than relying on `constraintName` being populated, which the code comment says it is not under Hibernate Reactive.

The migration rename this PR also carried is likewise already on `main` (`V15__notification_deduplication_key.sql`, byte-identical). Both PRs are additionally DIRTY after the merges that landed while they were open.

Nothing here says the analysis was wrong — #8966 in particular diagnosed the reactive-stack exception shape correctly. It simply arrived third.

PR #8969, opened 2026-09-06
fix(notification): remove duplicate-version V14 migration (#8961)

### LINKED ISSUE #8961: main is red: two V14 migrations in notification-service trip the enforced flyway-version-commit-order gate

JiRaska:
## What

Removes `V14__notification_deduplication_key.sql` from notification-service plus its `KNOWN_VIOLATIONS` baseline entry in `check-flyway-version-commit-order.py` (added by #8967 — the both-ways check would flag it stale the moment the file disappears).

## Why the #8967 remedy cannot be sufficient

Two migrations on main carry version 14 (`V14__synthetic_outbox_taint.sql` from #6731 landed first, this one from #8334 second). `QUARKUS_FLYWAY_OUT_OF_ORDER=true` relaxes *ordering*; it does not relax *resolution* — Flyway refuses to resolve two migrations with the same version at all, so any pod carrying both files still fails at startup.

That same fact is why deletion is safe: the deduplication migration has only ever coexisted with the synthetic one, so **no live database can ever have applied it** and renumbering rewrites no applied history.

## Plan

1. This PR: delete the duplicate + its baseline entry → the enforced `flyway-version-commit-order` gate (red on main for every PR) goes green.
2. Follow-up PR: re-add the identical DDL as `V15` with its rollback note, and retire the now-unneeded out-of-order env from `components/notifications/notification-service.yaml`.

Verified locally: `check-flyway-version-commit-order.py --enforce` — clean, 432 migrations, 2 known baseline entries.

Closes #8961


JiRaska:
Superseded by #8953 (already merged): it renumbered the migration V14→V15, removed the KNOWN_VIOLATIONS entry and the out-of-order env, fixed the dedup-conflict check in NotificationConsumer, and extended the IT. This PR was based on an older main and would revert parts of that fix.

PR #8973, opened 2026-09-06
fix(notification): match the dedup violation on PgException, not the JDBC exception

JiRaska:
The deduplication guard added by #8334 **could never fire**.

`NotificationConsumer` persists through **reactive** Panache (`io.quarkus.hibernate.reactive.panache`),
whose driver is the Vert.x PG client. The guard matched `org.postgresql.util.PSQLException` — the
**JDBC** driver's class, which never appears in that failure chain.

So a duplicate was re-thrown instead of skipped, the message was nacked, and redelivery retried it
forever. The exact opposite of what a deduplication guard is for.

Every other reactive service already gets this right — account, transaction, party, product-catalog.
party-service even documents it: *"PgException bubbles up unwrapped through Panache reactive"*. Of
the two `PSQLException` users in the tree, the other is `JdbcAgentAuditOutbox`, where it is correct.

## Why review and CI both missed it

The predicate was a **private method of an application class**, so only an integration test could
reach it — and that integration test passed **vacuously**. main carries two `V14` migrations in this
service, so the unique index never applied and the duplicate path was never exercised. Renumbering
to V15 (#8959) completed the schema, ran the path for real, and the timeout appeared.

That is the finding worth keeping: the migration collision was **masking a broken feature**, not
just breaking a boot.

## Change

Classification moved to the persistence boundary as `NotificationDeduplication`, mirroring
product-catalog's `PostgresConflicts`, so a unit test reaches it without booting a database.

Kept deliberately narrow: it is not enough that *some* unique index was violated. Only the dedup
index means "already recorded" — any other `23505` must keep failing loudly rather than being
silently acked as a duplicate and losing a notification without trace.

## Falsified, not asserted

Restoring the JDBC-shaped predicate turns **exactly the four recognition cases** red:

```
FAIL  a JDBC-shaped failure is not matched — this service persists reactively()
FAIL  the deduplication index violation is recognised()
FAIL  it is recognised through a wrapping cause chain()
FAIL  the constraint name is also read from detail()
pass  a different unique index is NOT a duplicate fact()
pass  a non-unique-violation sqlstate is not a duplicate()
pass  null and unrelated failures are not duplicates()
```

The three negative cases stay green — they guard against over-matching and are unaffected either
way. Restored: all seven pass.

## Verified

- 7 tests, 0 skipped, 0 failures, read from the JUnit XML
- ktlint + detekt clean
- exit code measured on the command, not on a pipeline

## Merge order

**Stacked on #8959.** Without the V15 renumber the schema does not apply and this path is
unreachable. Order: #8959 → this → #8874 → #8909.

## Note for the author of #8334

That PR was merged with a red `build (openbank-notification-service)` — 5 integration tests failing.
This PR and #8959 between them clear that.

### DIFF
--- a/openbank-notification-service/src/main/kotlin/com/openbank/notification/application/NotificationConsumer.kt
+++ b/openbank-notification-service/src/main/kotlin/com/openbank/notification/application/NotificationConsumer.kt
@@ -37,6 +37,7 @@ import com.openbank.notification.domain.model.PushSendOutcome
 import com.openbank.notification.domain.model.TemplateSensitivity
 import com.openbank.notification.infrastructure.client.PartyContactClient
 import com.openbank.notification.infrastructure.client.PartyMergeResolver
+import com.openbank.notification.infrastructure.persistence.NotificationDeduplication
 import com.openbank.notification.infrastructure.persistence.entity.NotificationEntity
 import com.openbank.notification.infrastructure.persistence.repository.DeviceTokenRepository
 import com.openbank.notification.infrastructure.persistence.repository.NotificationPreferenceRepository
@@ -58,7 +59,6 @@ import org.eclipse.microprofile.config.inject.ConfigProperty
 import org.eclipse.microprofile.reactive.messaging.Incoming
 import org.eclipse.microprofile.rest.client.inject.RestClient
 import org.jboss.logging.Logger
-import org.postgresql.util.PSQLException
 import java.time.Clock
 import java.time.Instant
 import java.util.UUID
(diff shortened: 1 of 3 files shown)


JiRaska:
Closing: superseded by #8953, and **my fix was wrong**.

#8953 landed the same diagnosis with a better-grounded answer. Hibernate Reactive adapts the
Vert.x `PgException` into a **plain `java.sql.SQLException`** (sqlState 23505, constraint named in
the message) beneath its `ConstraintViolationException` — so matching
`io.vertx.pgclient.PgException`, as this PR does, is just a second type that never appears in the
chain. Same class of defect I was fixing.

The evidence is direct, not a judgement call:

| | `NotificationConsumerIT` |
|---|---|
| #8953 (`java.sql.SQLException`) | **passes** |
| this PR (`io.vertx.pgclient.PgException`) | **fails** — 219 tests, 1 failed |

**Why my own verification did not catch it.** The unit tests construct a `PgException` and assert
the predicate matches it. That is circular: it proves the predicate recognises the input I invented,
not the input the runtime produces. The falsification I ran — reverting to a JDBC-shaped predicate
and watching four cases redden — only demonstrated the test can detect changes to *my* predicate.
Neither exercise could tell me the type was wrong, because both took my premise as given.

The integration test was the one check that could have told me, and it was red the whole time. I
read that red as "inherited from the base branch" and did not re-check it after the fix.

Nothing here is worth salvaging over #8953. The one idea that might still be: moving PostgreSQL
error classification out of a `private` method into a persistence-boundary object, so a unit test
can reach it without booting a database — that is what let the original defect survive review. But
a unit test over a fabricated exception would not have caught this one either, so it is a smaller
win than I claimed, and it needs the real type to be worth anything.

#8959 (the V15 renumber) is likewise superseded by #8953, which reached the same conclusion about
the two contradicting gates and carved out `check-db-migration.py` the same way.

PR #8977, opened 2026-09-06
fix(domestic-payment): slice the producer schema by indentation, not by its neighbour

JiRaska:
## Summary

`DomesticPaymentDelegationContractTest` fails on `main` and on every PR whose diff builds
`openbank-domestic-payment`. The guard it asserts is **not** broken — the test reads the wrong part
of the producer's AsyncAPI. This anchors the slice on indentation instead of on whichever schema
happens to be the next neighbour.

## Type of change

- [x] fix (bug fix)
- [ ] perf (performance)
- [ ] docs
- [ ] chore (build / tooling)
- [ ] breaking change

## Linked issues

Refs #8334

## What was measured

`origin/main`, no other commits, `:openbank-domestic-payment:test`:

```
DomesticPaymentDelegationContractTest > reservation consumer contract matches producer and retains only a domain-separated hash() FAILED
    java.lang.AssertionError at DomesticPaymentDelegationContractTest.kt:85
5 tests completed, 1 failed
```

Line 85 is `assertThat(producerSchema).doesNotContain("idempotencyKey:")`, and `producerSchema` was:

```kotlin
producerAsyncApi.substringAfter("    spendReservationState:")
                .substringBefore("\n    DelegationSpendReservationStateChanged:")
```

The first anchor is a **schema** (line 161 of the producer AsyncAPI); the second is a **message**
(line 420). The slice therefore spans 260 lines and swallows every schema in between. #8334 added
`spendReserved:` at line 217, and that schema carries the raw caller-chosen `idempotencyKey` **by
design** — the file says so eight lines above it:

> these are emitted for EVERY operation type, carry the raw (caller-chosen) idempotencyKey rather
> than its hash

**No contract regression, and no leak.** The schema the test means to police,
`spendReservationState` (lines 161-216), publishes `idempotencyKeyHash` and nothing else. The guard
was right the whole time; it was reading a neighbour.

## The fix

A `schemaBlock(yaml, key)` helper takes the block under `    <key>:` and stops at the next key of
the same indent, so a schema added later cannot drift into the span.

| tree | result |
|---|---|
| `origin/main`, unmodified | **FAILED** (the bug) |
| with this fix | **BUILD SUCCESSFUL** |
| with this fix **+ a raw `idempotencyKey:` injected into `spendReservationState`** | **FAILED** |

The third row is the one that matters: a fix that merely stopped the test failing would also have
disarmed it. The sabotage was reverted before committing; only the test file changes here.

## Definition of Done

- [x] Unit tests added/updated — this is the test.
- [x] `ktlintCheck` / `detekt` pass on the module.
- [x] No new lint warnings.

## Security checklist

- [x] Auth / crypto / payment code changed? → **no** — the change is one test file; no production
      code, no auth, crypto or payment path is touched, so no `security-review-required` tag.
- [x] PII path changed? → **no** — and the assertion this repairs is precisely the one that keeps a
      raw idempotency key out of the published contract. It was never leaking; the test was reading
      a neighbouring schema. No `gdpr-review-required` tag.
- [x] Cardholder data path changed? → **no** — domestic payments, no PAN anywhere in the diff, so no
      `pci-review-required` tag.

## Compliance impact

- [x] None — the published contract is unchanged; only the test's read of it.

## Rollout / Rollback

- Deployment strategy: N/A — test-only change.

## Reviewer notes

This is the third defect traced to #8334, which merged with a red build: the duplicate Flyway V14
and the dead `PSQLException` dedup check were the other two (both fixed in #8953). Worth a look at
why that PR was mergeable.

### DIFF
--- a/.github/scripts/check-adversarial-contract-tests.py
+++ b/.github/scripts/check-adversarial-contract-tests.py
@@ -56,6 +56,30 @@ SELFTEST_BAD = '''class FooPactConsumerTest {
     fun `returns the list`() { /* expects 200 */ }
 }'''
 
+# A `/contract/` file that never talks to a provider cannot express this gate's subject. The
+# rationale above is about authz: "the provider can stop enforcing authz entirely and every Pact
+# stays green". A test that asserts the SHAPE OF A PUBLISHED DOCUMENT — an asyncapi.yaml or an
+# openapi.yaml read as text — has no caller, no identity and no status code, so demanding a 401
+# of it can only be satisfied by writing one that tests nothing. Measured on
+# DomesticPaymentDelegationContractTest (#8977): zero HTTP calls, and the finding told it to add
+# an expectation it has no request to attach to.
+#
+# Pact-NAMED files stay in scope unconditionally — that is the population the debt was counted
+# over, and a pact test drives a provider even when the only marker is its name. A `/contract/`
+# file joins them when it actually exercises one.
+EXERCISES_PROVIDER = re.compile(
+    r"au\.com\.dius\.pact|io\.restassured|RestAssured|MockServer|PactDsl|"
+    r"@Provider\b|@PactFolder|@PactBroker|statusCode\(|\bgiven\(\)"
+)
+PACT_NAMED = re.compile(r"Pact[^/]*Test\.kt$")
+
+
+def in_scope(path: Path) -> bool:
+    """Pact-named always; a /contract/ file only when it drives a provider."""
+    if PACT_NAMED.search(str(path)):
+        return True
+    return bool(EXERCISES_PROVIDER.search(path.read_text(encoding="utf-8", errors="replace")))
+
 
 def changed_contract_files(since: str) -> list[Path]:
     res = subprocess.run(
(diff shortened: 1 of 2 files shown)


JiRaska:
Closing this. Both halves are superseded, and the second half was **wrong** — worth stating plainly
so the reasoning does not get reused.

## The test fix: same defect, same shape, landed first

#8991 fixed the slice on `main` while this was open, bounding the schema window by indent
(`Regex("\n    \\S")`) instead of by the neighbouring message key. Same diagnosis, equivalent
implementation. Nothing here to add.

## The gate change: my premise was false, and #8991 demonstrates it

I narrowed `adversarial-contract-test` to skip "document-shape" contract tests, arguing:

> A test that asserts the SHAPE OF A PUBLISHED DOCUMENT … has no caller, no identity and no status
> code, so demanding a 401 of it can only be satisfied by writing one that tests nothing.

That is not true, and the counter-example is the very PR that superseded this one. Facing the same
finding on the same file, #8991 added a real negative case:

```kotlin
fun `every published operation keeps the 403 rejection in the contract`() {
    …
        .`as`("operation %s documents the 403 rejection", opId)
        .contains("'403':")
}
```

An OpenAPI document **does** carry status codes, so "every published operation documents its 403"
is a genuine contract assertion — and a valuable one: it fails the moment an operation is published
without its rejection documented, which is exactly the drift the gate exists to catch. The gate was
correctly pushing for an assertion the file could express; I concluded it could not, and would have
disarmed the gate for a whole class of files on that basis.

The `--fleet-report` rescoping in the same commit rested on the same premise (20 files excluded as
"document-shape"), so it goes with it.

## What stands

Nothing from this branch needs to land. The measurement in the PR body is still accurate — there
was no leak, `spendReservationState` publishes only `idempotencyKeyHash`, and the failure was the
slice reading `spendReserved` next door — but #8991 carries the fix and the better answer to the
gate.

Issue #8351, opened 2026-09-03
Idempotency coverage inventory + ratchet gate (M2 exit criterion)


JiRaska:
## Context
M2 (Resilience) lists "idempotency coverage" as a remaining gap (ROADMAP.md). There is no measured inventory: which mutating endpoints require `Idempotency-Key`, which enforce it, which silently accept duplicates.

## Acceptance criteria
- [ ] Enumerate all mutating (POST/PUT/PATCH/DELETE) endpoints across released services from openapi.yaml files
- [ ] Classify each: idempotency-key required / optional / absent — with justification per class
- [ ] Money-path POST endpoints: 100% required + enforced, or an ADR-linked exception
- [ ] CI check (`check-idempotency-coverage.py` style) ratchets the inventory so a new money-path mutating endpoint without idempotency fails the gate
- [ ] The known trap is covered by the check design: `@QueryParam`/`@HeaderParam` nullability per `check-nonnull-jaxrs-params.py` precedent

Audit: docs/audit-2026-09-03-top50.md (B15). M2 exit criterion.

JiRaska:
## Measured idempotency inventory — and what a gate can honestly enforce

Inventory first, gate second, per the acceptance criteria. Every count below carries the command that produced it. Measured on `origin/main` @ `736647d2`, 2026-09-03.

Gate shipped in #8551.

---

### 1. The corpus

```
find . -name openapi.yaml -path '*/src/main/resources/*' | wc -l          # 52
```

Mutating operations (`post|put|patch|delete`) enumerated by parsing each spec with PyYAML, resolving path-level `parameters` and `$ref` chains:

| | count |
|---|---|
| specs | 52 |
| mutating operations | **375** |
| ops declaring an idempotency **header** | 30 |
| ops carrying the key as a request-**body** field | 9 |
| money-path mutating ops | 148 (113 of them `POST`) |

**Two probe corrections worth recording**, because both reported a clean number that was wrong:

- The first pass ignored `$ref`'d parameters. There are 240 of them (`grep -rn '\$ref.*components/parameters' openbank-*/src/main/resources/openapi.yaml | wc -l`); resolving them moved declared 27→28, required 16→17.
[...]

PR #8978, opened 2026-09-06
fix(sdd): replay-safe collection authorisation + ADR-0289 natural-key idempotency (#8351)

JiRaska:
## What

Burn-down #8351, SDD batch (money-path):

1. **Replay-safe authorise** — `POST /api/v1/sdd/collections/authorise` re-stamped the mandate and re-emitted the outbox event on every retry of the same collection. It now short-circuits when the instruction's `dueDate` equals the mandate's `lastCollectionDate`: the stored policy decision replays with no `save` and no outbox append. A different dueDate is a new collection and flows through unchanged.
2. **ADR-0285** — records the natural-key idempotency of both SDD creation POSTs: mandates dedup on (creditorId, UMR), authorise on (mandateId, UMR, dueDate) — the key the downstream collection-debit consumer already dedups under (`so-sdd-{mandateId}-{umr}-{dueDate}`). Tags use the closed vocabulary (`resilience`, `payments`) rather than coining `idempotency`.
3. **OpenAPI** — both operations' descriptions now state the idempotency contract (`info.version` 1.1.0 → 1.1.1, PATCH: descriptions only).
4. **Baseline** — the two SDD creation-POST entries re-pointed to ADR-0285 (lines kept; the gate is shrink-only and natural-key exceptions stay listed).

## Verification

- `SddMandateServiceTest` incl. the new replay test (0× save / 0× outbox on same-dueDate retry) — green locally.
- ktlint main+test — green locally.
- `check-adr-registry.sh` — green; index regenerated (275 ADRs).
- Threat-model change-log entry added (2026-09-06).

## Security checklist

- [x] Threat model reviewed — `docs/threat-models/openbank-sdd-service.md` change-log updated; no new caller, endpoint or privilege
- [x] Replay path fails closed: a retry returns the stored decision, never a fresh side effect
- [x] No authz/authn change; OPA and RBAC wiring untouched

Refs #8351


JiRaska:
**Renumbered.** This ADR was written as a number that another PR took on `main` while this branch
was open, so the number moved and every reference moved with it — the ADR file and its H1, the
threat model, the idempotency baseline, the OpenAPI descriptions, the service and test comments,
and the migration's header comment. `check-adr-registry.sh` is green on the result.

The migration comment was safe to edit because that migration is new on this branch and has never
been applied anywhere; the Flyway checksum rule only bites after a migration has run.

PR #8980, opened 2026-09-06
fix(balance): replay-safe hold placement + ADR-0287 natural-key idempotency (#8351)

JiRaska:
## What

Burn-down #8351, balance-service batch (money-path) — the one **real double-effect bug** the inventory surfaced here:

- **Hold placement was not replay-safe.** A retried `POST /api/v1/balances/{accountId}/holds` reserved twice — the amount stayed unavailable until expiry or release. `placeHold` now checks the caller-supplied natural key (accountId, currency, referenceId) first and replays the original hold with no second reservation and no second `HOLD_PLACED` event. New unique index `uq_balance_holds_reference` (V10, with rollback note) is the race backstop for two concurrent first attempts; a lost race is recovered by re-reading the winner's row (constraint-name-checked 23505, same shape as #8953).
- **ADR-0287** documents the natural-key idempotency of all five balance creation POSTs: credit/debit via the existing V8 movement ledger, initialize via (account, currency), holds via the new check-first + index, reconciliation as an operator trigger whose re-run records a fresh report by design.
- **OpenAPI**: the four operations' descriptions now state the contract (`info.version` 1.9.0 → 1.9.1, PATCH — descriptions only).
- **Baseline**: the five balance entries re-pointed to ADR-0287 (lines kept — shrink-only gate, natural-key exceptions stay listed).

## Verification

- `BalanceServiceTest` incl. the new replay test (retry returns the original hold id, exactly one `saveWithEvent`), `LedgerProjectionServiceTest`, ktlint main+test — green locally.
- `check-adr-registry.sh` — green; index regenerated (276 ADRs).
- Threat-model change-log entry added (2026-09-06).

## Security checklist

- [x] Threat model reviewed — `docs/threat-models/openbank-balance-service.md` change-log updated; no new caller, endpoint or privilege
- [x] Replay path fails closed: a retry returns the original hold, never a second reservation
- [x] Migration is additive (index only) with a rollback note; no existing column or constraint altered

Refs #8351


JiRaska:
Heads-up for whoever is working this branch from `tmp/wt-8351-balance`: a force-push at 14:14:28
dropped two commits that had been pushed a minute earlier (14:13:12), and I have restored them by
cherry-picking on top of your head rather than replacing it — your `ae44b2de0` is untouched and is
the parent of both.

What was lost, and why each matters:

1. **`V11` was claimed twice.** `V11__balance_outbox_claimed_at.sql` reached main while this branch
   was open, so `V11__balance_hold_reference_idempotency.sql` had to move. Flyway keys its schema
   history by version and refuses the whole set — `Found more than one migration with version 11` —
   so balance-service does not boot at all with both present. V12 and V13 are also taken; the file
   is now `V14`, and the references in `BalancePorts`, `BalanceService`, ADR-0287, the threat model
   and the idempotency baseline moved with it. `check-flyway-version-commit-order.py --enforce` is
   clean again over 434 migrations in 57 services; without it, that gate is the red
   `gates (registry-kotlin-data)` on this PR.

2. **`BalanceServiceClockTest` had no stub for `holdRepo.findByNaturalKey`.** Replay-safe
   `placeHold` asks for an existing hold before reserving, so mockk failed the call rather than the
   assertion — `io.mockk.MockKException at BalanceServiceClockTest.kt:52`. The stub answers "no
   prior hold", which is what that test's subject (the injected clock) needs.

Both were verified before pushing, not assumed: the Flyway gate exits 0, and the test file compiles
and passes locally.

JiRaska:
**Third application of the same two commits.** They were pushed at 14:13, dropped by a force-push at
14:14:28, restored, and dropped again at 15:00:17. This is not a disagreement about the content —
the rebuilds simply start from a base that predates them.

Both are cherry-picked on top of `569f21533`, so your commit is untouched and is their parent. Pushed
as a fast-forward; nothing of yours is replaced.

**If you rebuild this branch again, please carry these two — without them the branch is broken in a
way CI reports but a local run may not:**

1. **`V11` is claimed twice.** `V11__balance_outbox_claimed_at.sql` is on main;
   `V11__balance_hold_reference_idempotency.sql` is this branch's. Flyway keys its schema history by
   version and refuses the whole set — `Found more than one migration with version 11` — so
   balance-service does not boot at all. V12 and V13 are also taken, hence **V14**, with the
   references in `BalancePorts`, `BalanceService`, ADR-0287, the threat model and the idempotency
   baseline moved to match. This is what `gates (registry-kotlin-data)` fails on.

2. **`BalanceServiceClockTest` needs a stub for `holdRepo.findByNaturalKey`.** Replay-safe
   `placeHold` asks for an existing hold before reserving, so mockk fails the *call* rather than the
   assertion: `io.mockk.MockKException at BalanceServiceClockTest.kt:52`. The stub answers "no prior
   hold", which is what a test about the injected clock needs.

Verified on the pushed head before pushing, not assumed:

```
check-flyway-version-commit-order: 434 migration(s) across 57 service(s) checked — clean.  (exit 0)
BalanceServiceClockTest: tests="2" skipped="0" failures="0" errors="0"
```

A cheap check that would have caught the drop from your side too: after rebuilding, run
`git merge-base --is-ancestor <sha> HEAD` for anything recently pushed to the branch, or just
`git log --oneline origin/<branch>` before the force-push.

Issue #8984, opened 2026-09-06
ci: pacts/** maps to no service — a pact-only PR can break the provider's build unseen


JiRaska:
## What happened

#8697 changed ONLY files under `pacts/` (regenerated consumer pacts). Path-scoped CI built no service, so the provider replay (`TransactionPactFolderProviderVerificationTest`) never ran on the PR — and after merge, main's transaction-service build was red with no alert, until an unrelated PR's merge-ref exposed it.

This is the shape AGENTS.md already warns about: **a module CI never rebuilt stays red on main unseen.**

## Why it is systemic, not one-off

Every committed pact in `pacts/<consumer>-<provider>.json` names its provider in the filename. A pact-only change can break the provider's verification for ANY provider service, and no current path mapping causes that provider's build to run.

## Suggested fix (conceptual)

In the path-scoping logic, map a changed `pacts/*-<provider>.json` to the provider service's build (suffix after the LAST `-`... careful: service names contain hyphens — derive from the pact's `provider.name` field instead, or keep an explicit mapping the derive-pact-drift-scope.sh script already maintains). Alternatively, a cheap floor: any `pacts/**` change triggers builds of all PROVIDER services found in the changed files' `provider.name`.

Refs #8697, #8983

JiRaska:
Fixed by #8987 (merged): a changed pacts/*.json now builds BOTH sides on PRs (provider derived from the '<consumer>-openbank-<provider>' filename, falling back to full-fleet when the provider does not resolve to a module — the safe direction). Verified in the merged diff: .github/workflows/services-ci.yml.

PR #8988, opened 2026-09-06
test(transaction): serve the unauthenticated pact interactions from their own provider class

JiRaska:
## Summary

`main`'s transaction-service provider replay has been red since #8697. Three consumer pacts gained a "missing or expired token" interaction whose provider state has no handler, so pact-jvm fails them outright. This serves those interactions from a class that can actually produce a 401.

## Type of change

- [x] fix (bug fix)
- [ ] perf (performance)
- [ ] docs
- [ ] chore (build / tooling)
- [ ] breaking change

## Linked issues

Refs #8984, #8697

## Changes

- `TransactionNegativeAuthPactVerificationTest` — new, without `@TestSecurity`, filtered to the one negative state, with the state handler pact-jvm was looking for.
- `TransactionPactFolderProviderVerificationTest` — filtered to everything else.
- `TransactionPactStateCoverageTest` — new guard, see below.

## Why a state method alone does not work

The existing class carries a class-level `@TestSecurity(user = "pact-verifier", roles = ["ROLE_OPERATOR"])`, so every request it makes is authenticated. The three interactions expect **401**, and no state handler can undo a class-level security context — it would answer 200 where the pact demands 401. They need a class without `@TestSecurity`, so the request arrives anonymous and `TransactionResource`'s `@RolesAllowed` answers 401, which is what the consumers encoded.

## Why two classes are safe here

`CLAUDE.md` warns against two verification classes for one provider. That collision is two BROKER-sourced classes each fetching every pact the broker holds. These are both `@PactFolder` with disjoint state filters built from one shared literal, so no interaction lands in both or in neither.

Checked against the real state names rather than by reading the regex:

| provider state | authenticated class | negative class |
|---|---|---|
| a valid borrower account exists | yes | no |
| a valid source account exists | yes | no |
| no valid M2M identity is presented | no | yes |
| the transaction service is available | yes | no |
| transaction-service is reachable and holds no transactions for the pact account | yes | no |

pact-jvm 4.7.3 matches a filter value against the state name with `String.matches`, a full-match Java regex — read from `InteractionFilter$ByProviderState`'s bytecode, so the negative lookahead is supported rather than hoped for.

## The assumption, and its guard

`ByProviderState` is an `anyMatch` over the interaction's states, so an interaction with **no** state matches neither filter and would be verified by neither class, silently, with both green. All 13 committed interactions declare a state today. `TransactionPactStateCoverageTest` fails the build if that changes, and asserts its corpus is non-empty first, so a wrong path cannot make it pass by finding nothing.

## Definition of Done

- [x] No new lint warnings.

The Gradle row is unticked deliberately: this machine is running several agents and a single-module test run took two and a half hours earlier tonight, so CI on this PR is the verification. The partition table above and the pact-jvm matching semantics were established locally.

## Security checklist

- [x] Test-only: no production code, no endpoint, no dependency, no secret, no config.
- [x] Auth behaviour is exercised, not changed — the three interactions assert that an unauthenticated money-path debit is refused, which nothing verified before.
- [x] No PII and no cardholder data are read or written; the pacts carry synthetic ids.
- [x] No new `@SuppressWarnings`, `as Any` or `@Suppress` beyond one `UNCHECKED_CAST` on a JSON map read in the coverage test.

## Compliance impact

- [ ] PCI DSS
- [ ] DORA
- [ ] GDPR
- [ ] PSD2 / SCA / RTS
- [ ] AML / 5AMLD
- [ ] CNB reporting
- [x] None

## Rollout / Rollback

- Deployment strategy: test-only, no runtime change.
- Rollback plan: revert; main returns to the red replay.

## Reviewer notes

Another session appeared to be working the same defect in a local worktree tonight, with no branch pushed after more than an hour idle, so this may overlap with work you have seen elsewhere — compare content before merging both. The judgement call is the two-class split versus relaxing `@TestSecurity` on the existing class and authenticating per interaction, which pact-jvm gives no clean hook for.

### DIFF
--- a/docs/threat-models/openbank-transaction-service.md
+++ b/docs/threat-models/openbank-transaction-service.md
@@ -323,3 +323,18 @@ what the catalogue may hold.
   alter initiation); the span is assertion-backed by `TransactionApiIT`, which drives the real HTTP
   endpoint against PostgreSQL/Redpanda Testcontainers and the test Temporal terminal-write workflow.
   The contract proves this service boundary only — it does not claim a distributed downstream trace.
+- **2026-09-07** — Authentication-failure response shape. An unauthenticated call to a
+  `@RolesAllowed` endpoint answered with Quarkus's bare `Not Authorized` string, while every other
+  error from this service is an `ApiError` document: `io.quarkus.security.UnauthorizedException` is
+  not a `WebApplicationException`, so `WebApplicationExceptionMapper` — which already maps status
+  401 to `ErrorCode.UNAUTHORIZED` — never saw it. The service now registers the shared
+  `UnauthorizedExceptionMapper` from `openbank-libs-runtime` via a thin `@Provider` subclass, so a
+  401 carries the standard envelope. Risk class = **integrity of the client contract**, not
+  confidentiality: the trust boundary itself is unchanged, the endpoint is refused exactly as
+  before, and the envelope adds no detail about why — the message is a constant
+  (`Authentication required`), never the exception's own text, so nothing about the token, the
+  principal or the failure reason reaches the caller. What changes is that a caller parsing the
+  error body no longer breaks on the one response it is most likely to receive. Assertion-backed by
+  the swift, sdd and interest provider-replay interactions, which each require a 401 for a debit
+  presented without a valid M2M identity and could not be verified at all until this landed
+  (issues #8993, #8984).
--- a/openbank-libs-runtime/src/main/kotlin/com/openbank/libs/api/error/CommonExceptionMappers.kt
+++ b/openbank-libs-runtime/src/main/kotlin/com/openbank/libs/api/error/CommonExceptionMappers.kt
@@ -7,6 +7,7 @@ package com.openbank.libs.api.error
 import com.openbank.libs.approval.InvalidApprovalStateException
 import com.openbank.libs.approval.SelfApprovalNotAllowedException
 import com.openbank.libs.authz.PolicyDecisionException
+import io.quarkus.security.UnauthorizedException
 import jakarta.ws.rs.WebApplicationException
 import jakarta.ws.rs.core.Response
 import jakarta.ws.rs.ext.ExceptionMapper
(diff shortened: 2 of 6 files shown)


JiRaska:
CI has taken this two steps forward and stopped at a third, so here is exactly where it stands.

**Fixed and proven.** `TransactionPactFolderProviderVerificationTest` now runs 11 interactions with 0 failures. Before this branch it failed three of them with `MissingStateChangeMethod`, which is the red on `main`. The state filter partitions the pacts as intended and the state handler is found.

**What still fails.** The three negative interactions now reach the provider and get a 401, and pact-jvm then refuses the response body:

```
Request Failed - Invalid JSON (1:2), found unexpected character 'N'
```

The pact expects `{"status": 401}` and nothing else, so this is not a mismatch of expectations — it is pact-jvm parsing the actual body. The body is Quarkus's own `Not Authorized`, produced by the security layer for an anonymous request to an `@RolesAllowed` endpoint.

**Why that is a real defect and not a test artifact.** `io.quarkus.security.UnauthorizedException` is not a `WebApplicationException`, so `libs-runtime`'s `WebApplicationExceptionMapper` — which already maps 401 to the `UNAUTHORIZED` error envelope — never sees it. Every other error from these services is a JSON envelope; a 401 is a bare string. A client that parses the error body breaks on exactly the response it is most likely to receive, which is what the three consumers encoded and what nothing until now exercised.

**Why I stopped here.** The fix belongs in `openbank-libs-runtime` (a service-local mapper is forbidden, #526), so it changes the 401 body for every service in the fleet. That is not something to push unverified at 01:30 into a money-path repository, and another session appeared to be editing `CommonExceptionMappers.kt` tonight, so a second uncoordinated change would be worse than none.

Next step for whoever picks this up: add an `ExceptionMapper<io.quarkus.security.UnauthorizedException>` to `CommonExceptionMappers.kt` returning the standard envelope with `ErrorCode.UNAUTHORIZED`, and check the `ClassNotFoundException`-at-ArC-init hazard the file already documents for the Hibernate mappers — `io.quarkus.security` is on every service's classpath, but that is worth measuring rather than assuming. With that in place this branch should go green as it stands.

JiRaska:
One correction to my own "next step" above, because I checked before assuming.

`quarkus-security` is `compileOnly` in `openbank-libs-runtime` (build.gradle.kts:79). A `@Provider class ... : ExceptionMapper<io.quarkus.security.UnauthorizedException>` there is therefore the exact hazard that file already documents for the Hibernate mappers: services without the extension on their runtime classpath crash at ArC init with `ClassNotFoundException`, which is how agent-service, analytics-sink and ap2-service broke in #6240.

So the fix is not a five-line mapper. It is one of:

- the name-based route the same file already uses for persistence failures, classifying by class name while walking the cause chain, so no compile-time reference to the security type exists; or
- an explicit per-service registration, the pattern the Hibernate mappers use, which means touching every service that has security.

Either is a fleet-wide decision about what a 401 body looks like, and it deserves daylight and its own PR rather than being appended to this one. This branch stops at the point where it is provably useful: the provider replay's state handling is fixed and the negative interactions are isolated in a class that can serve them, which is the half `main` is red for today.

PR #9011, opened 2026-09-07
fix(consent): replay-safe suppression creation + ADR-0293 natural-key idempotency (#8351)

JiRaska:
## What

Burn-down #8351, consent-service batch:

- **Suppressions were not replay-safe** — a retried `POST /api/v1/suppressions` stacked a second identical active row. `SuppressionService.create` now checks the natural key (partyId, scope, value) over active rows and replays the original (no duplicate row, no second event). New partial unique index `uq_suppressions_active_natural` (V8, `WHERE revoked_at IS NULL`, value coalesced for scope=ALL; rollback note included) is the race backstop; a lost race re-reads the winner's row. A revoked suppression deliberately does not block a fresh one of the same value.
- **Consents** — the idempotency was already enforced in code (Berlin Group `tppTransactionId`, falling back to `X-Request-ID`, replayed from the idempotency store with `X-Idempotency-Replayed: true`) but the contract said nothing. OpenAPI now declares both keys, still optional (the pre-existing edge contract); `info.version` 1.9.0 → 1.9.1 (PATCH — descriptions/declared optional params only).
- **ADR-0288** records both decisions; the two baseline entries re-point to it.
- **Threat model** change-log updated (2026-09-07).

## Verification

- New `SuppressionServiceTest`: replay returns the original row with 0× save; a revoked suppression does not block a fresh create — green locally.
- ktlint main+test, `check-adr-registry.sh` — green locally.

## Security checklist

- [x] Threat model reviewed — change-log updated; no new caller, endpoint or privilege
- [x] Replay path fails closed: a retry returns the original row, never a duplicate
- [x] Migration is additive (partial unique index) with a rollback note

Refs #8351


JiRaska:
**Renumbered.** This ADR was written as a number that another PR took on `main` while this branch
was open, so the number moved and every reference moved with it — the ADR file and its H1, the
threat model, the idempotency baseline, the OpenAPI descriptions, the service and test comments,
and the migration's header comment. `check-adr-registry.sh` is green on the result.

The migration comment was safe to edit because that migration is new on this branch and has never
been applied anywhere; the Flyway checksum rule only bites after a migration has run.

JiRaska:
Correction to the PR body: it classifies the consent spec change as PATCH (`1.9.0 -> 1.9.1`) because it only adds descriptions and *optional* parameters. That reasoning is wrong — a parameter the document did not previously declare is additive whether or not it is required, and oasdiff classifies it that way. The gate said so directly, and the version is now `1.10.0`, read off live main.

PR #9037, opened 2026-09-07
fix(transaction): unparseable enum inputs are 400, never a silent null

### LINKED ISSUE #8699: transaction-service: an unparseable enum becomes null — a typo'd status filter returns EVERY transaction, and a typo'd rail is persisted as null with a 201

JiRaska:
Closes #8699

## Root cause
`runCatching { X.valueOf(it) }.getOrNull()` on six call sites turned an unparseable value into a **legal null**, and null then did two different wrong things:
- **Search filters**: `?status=FAILDE` dropped the condition and returned **every** transaction with 200 — a malformed date three lines below correctly 400s.
- **Initiate path**: an unknown `rail` persisted a money-path debit with `rail = null` under a 201 — routing information silently lost.

## Fix
- `TransactionResource`: strict `parseEnumParam` — absent stays null, unparseable throws `IllegalArgumentException` (libs-runtime maps it to 400, #526) with the allowed values in the message. Applied to `status`/`type` filters and the `rail`/`instructionType` write path.
- `PanacheTransactionRepository.toDomain()`: read-back of persisted `rail` / `instructionType` / `actorId` is strict via `strictPersisted` — a persisted value that does not parse is **data corruption** and must be loud, never silently nulled.
- The strictness immediately exposed a stale contract: the sepa-payment consumer pact sent `rail=SEPA` (not a `PaymentRail` value) and only passed because of the silent null. Production `SettlementAdapter` sends `SEPA_CT`; pact test + pact file now mirror the real wire.

## Tests
- New ITs in `TransactionApiIT`: typo `status` → 400 (not unfiltered 200), valid `status` → 200, typo `rail` → 400 (not persisted null), absent `rail` → still 201.
- Local gate green: `:openbank-transaction-service:test ktlintCheck detekt koverVerify` + `:openbank-sepa-payment:test ktlintCheck detekt`.

## Security checklist
- [x] No secrets, credentials, or PII added to code or logs (enum names in error messages are not PII)
- [x] Input validation tightened, not loosened (strict enum parsing on money-path write)
- [x] No new outbound edges or trust boundaries introduced
- [x] Money-path behavior change is fail-closed (unknown rail now rejected instead of persisted as null)


JiRaska:
CI note: `CodeQL (java-kotlin, manual)` is the known environment-bound check (manual trigger, GitHub-side) — every other gate is green (25 pass incl. all-green, adversarial-contract-test and threat-model gates fixed by 52c38a2b2). The earlier Dependency review / Trivy / security-regression-test failures were the tracked GitHub rate-limit/env failures (#8690), now passing.

PR #9054, opened 2026-09-07
fix(transaction): reversals record what they reverse

### LINKED ISSUE #8841: transaction-service: reversal_of and is_reversal are declared, documented, drawn in the ER diagram — and never assigned
## What

`openbank-transaction-service` records a reversal and cannot say **what it reversed**. Two columns exist for exactly that, are documented as compliance fields, are drawn in the ER diagram — and no code path ever assigns either one.

```
git grep -nE "reversalOf|isReversal" origin/main -- 'openbank-transaction-service/src/main/**/*.kt'

TransactionEntity.kt:111    var reversalOf: UUID? = null
TransactionEntity.kt:148    var isReversal: Boolean = false
```

Those two lines are the **entire** result: the declarations themselves. No mapper, no use case, no repository writes either field. So for every row ever written, `reversal_of` is NULL and `is_reversal` is FALSE — including on genuine reversals. `is_reversal` is `NOT NULL DEFAULT FALSE` (V3), so it is uniformly false and carries no information at all.

## Why the fields cannot be populated as things stand

`TransactionService.reverseTransaction` (`:257`) holds the original's id — it is `ReverseTransactionCommand.originalTransactionId`, and the method reads the original row with it. It then creates the reversal by delegating to `initiateTransaction`:

```kotlin
transactionRepository.update(original.reverse())          // original -> REVERSED
return initiateTransaction(
    InitiateTransactionCommand(
        type = TransactionType.REVERSAL,
        description = "Reversal: ${command.reason}",
        ...                                                // no field for the original id
    ),
)
```

`InitiateTransactionCommand` has no parameter for an original transaction, and the domain `Transaction` model has no such property either (`git grep -nE "original|reversal|reversed" .../domain/model/Transaction.kt` is empty). The id is available at the call site and is dropped one line later.

## What the repo asserts instead

- `V2__compliance_fields.sql:14` — `ADD COLUMN IF NOT EXISTS reversal_of UUID;`, in the migration named **compliance fields**
- `diagrams/02-er-schema.mmd:66` — `transactions ||--o| transactions : "reversal_of"`, a self-relation that has never held for a single row
- `docs/04-data.en.md:37` and `04-data.cs.md:37` list `reversal_of` among the V2 compliance fields

## The sibling service shows the intended shape

`openbank-ledger-service` has the same concept and the same column name, wired end to end:

```
JournalEntry.kt:105                 reversalOf = id
PanacheJournalRepository.kt:330     it.reversalOf = reversalOf
PanacheJournalRepository.kt:359     reversalOf = reversalOf
V12__unique_reversal_of.sql         CREATE UNIQUE INDEX uq_journal_entries_reversal_of
```

with a comment explaining the index is the backstop against double-credit downstream. That is what makes this read as an omission rather than a deliberate choice: transaction-service has the column, the migration, the diagram and the docs, and none of the code.

## Consequence

A consumer of `openbank.transactions.*` learns **that** a reversal happened — the new REVERSAL transaction emits `transaction.initiated` like any other (`TransactionService.kt:217-219`) — and cannot link it to the original except by parsing `description`, which carries the *reason*, not an id. The same is true of anyone querying the table directly for reconciliation or regulatory reporting.

## What this is NOT

- **Not** "reversals are invisible". They are announced; only the linkage is missing. I checked this first and withdrew the stronger claim on #8745.
- **Not** an open door to double-reversal. `reverseTransaction` requires `original.status == COMPLETED` and `reverse()` moves it to `REVERSED`, and there is an idempotency key, so the missing unique index is not currently load-bearing the way ledger's is.

## Suggested fix

Add `reversalOf` (and set `isReversal`) through `InitiateTransactionCommand` -> domain `Transaction` -> `TransactionEntity`, populated by `reverseTransaction` from the id it already holds. Then decide whether the ledger-style unique index is wanted as a second backstop, or whether the status check is considered sufficient — that is a judgement call and worth stating explicitly either way rather than leaving the column unindexed by default.

A test should assert the link on a real row rather than on a mock: the defect is precisely that a field the code never touches looks fine in every unit test that does not read it back.

Found while verifying finding 1 of #8745, which did not hold. Refs #8745

JiRaska:
Closes #8841

## Root cause
`reversal_of` / `is_reversal` are V2 **compliance** columns — documented in `docs/04-data.*.md`, drawn as a self-relation in the ER diagram — with no code path ever assigning either. `reverseTransaction` holds the original id in `command.originalTransactionId` and dropped it one line later, because `InitiateTransactionCommand` and the domain `Transaction` had no field for it.

## Fix
- `InitiateTransactionCommand.reversalOf` → domain `Transaction.reversalOf` / `isReversal` → entity mapper both directions.
- `reverseTransaction` passes `original.id`; `isReversal` derives from `type == REVERSAL`.

## Unique index — decision made explicitly (the issue asked)
No `uq_transactions_reversal_of`: `transactions` is `PARTITION BY RANGE (booking_date)` and Postgres requires every partitioning column in a unique constraint (measured — the migration fails at boot). The COMPLETED→REVERSED status transition (+ idempotency key) stays the sole double-reversal guard; recorded in the command KDoc.

## Tests
- `TransactionServiceTest` asserts `reversalOf`/`isReversal` on the reversal result.
- New `TransactionReversalLinkIT`: real REST initiate → reverse, then asserts the link on the **persisted row** via plain JDBC — the defect class is a field no test ever read back.

Local gate green: `test ktlintCheck detekt koverVerify`.

## Follow-up (not in this PR)
`reversalOf` is not yet exposed on the REST response DTO / `transaction.initiated` event payload — that is an additive API change (ADR-0048 versioning) worth its own PR.

## Security checklist
- [x] No secrets, credentials, or PII added to code or logs
- [x] Input validation unchanged; no validation loosened
- [x] No new outbound edges or trust boundaries introduced
- [x] Money-path behaviour strictly gains auditability (compliance linkage); no control removed

### DIFF
--- a/openbank-transaction-service/src/main/kotlin/com/openbank/transaction/application/port/in/TransactionPorts.kt
+++ b/openbank-transaction-service/src/main/kotlin/com/openbank/transaction/application/port/in/TransactionPorts.kt
@@ -33,6 +33,16 @@ data class InitiateTransactionCommand(
     val scaExemption: String? = null,
     /** Rail payment that triggered this booking (ADR-0108); null for operator/system commands. */
     val originatingPaymentId: UUID? = null,
+    /**
+     * The transaction this booking reverses (#8841). Set only by `reverseTransaction`, which
+     * already holds the original's id — without this parameter the id was available at the call
+     * site and dropped one line later, leaving the V2 compliance column `reversal_of` NULL on
+     * every row ever written. Deliberately no ledger-style unique index: `transactions` is
+     * PARTITION BY RANGE (booking_date), and Postgres requires every partitioning column in a
+     * unique constraint — the COMPLETED→REVERSED status transition stays the sole guard against
+     * a double reversal, and the decision is recorded here rather than left by default.
+     */
+    val reversalOf: UUID? = null,
     /** Which scheme carried the money (ADR-0103 D2). Null until stamped at origination. */
     val rail: PaymentRail? = null,
     /** How the movement was instructed (ADR-0103 D2) — orthogonal to [rail]. */
--- a/openbank-transaction-service/src/main/kotlin/com/openbank/transaction/application/usecase/TransactionService.kt
+++ b/openbank-transaction-service/src/main/kotlin/com/openbank/transaction/application/usecase/TransactionService.kt
@@ -209,6 +209,8 @@ class TransactionService(
             scaExemption = command.scaExemption,
             rail = command.rail,
             instructionType = command.instructionType,
+            reversalOf = command.reversalOf,
+            isReversal = command.type == TransactionType.REVERSAL,
         )
 
         val saved = try {
@@ -284,6 +286,7 @@ class TransactionService(
                 description = "Reversal: ${command.reason}",
                 valueDate = java.time.LocalDate.now(clock),
                 initiatedBy = UUID.fromString("00000000-0000-0000-0000-000000000001"),
+                reversalOf = original.id,
             ),
         )
     }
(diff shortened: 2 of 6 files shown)


JiRaska:
CI note: `CodeQL (java-kotlin, manual)` is the known environment-bound manual check; all-green passes (Dependency review + Trivy green this run).

Issue #7740, opened 2026-08-31
fleet-attestation placeholder allowlist cannot fail on a stale entry; all 5 are stale


JiRaska:
`.github/scripts/fleet-attestation-placeholders.txt` is a baseline that can only fail in one direction.

Its header says *"Remove an entry the moment the service is genuinely built + pushed."* Nothing enforced that. The allowlist is consulted only for an image ref that is DECLARED under `openbank-infra/gitops/`, so an entry whose ref is no longer declared suppresses nothing — it can never reach the ABSENT branch, appears in no run output, and survives only as prose asserting something about the fleet.

Measured on `origin/main` (2026-08-31): **all five entries** had outlived their subject. Every pin has moved to a real built tag — finrep and vop `sandbox-3b62a4a5`, delegation and referral `sandbox-c3f21ee6`, case-coordinator-agent `sandbox-e96c8e41`. Two entries additionally justify themselves with claims main contradicts: the vop entry cites *"no Dockerfile ... absent from auto-deploy's ALL_SERVICES"* and `openbank-vop-service/Dockerfile` exists while `openbank-vop-service` is listed in `auto-deploy.yml`'s `ALL_SERVICES`.

This is the repo's established convention, which this one file sat outside of: a declaration must fail in both directions, or it outlives its subject. Compare `rules.yaml: lineage_code_audit` ("an allowlist entry below whose edge is no longer declared ... is itself a finding — delete it") and `deploy-coverage-baseline.txt`'s `STALE_BASELINE`.

Fix: extend the existing checker with a staleness pass and drain the file.

JiRaska:
**STILL TRUE — re-verified on `origin/main` today (2026-09-01).** Nothing has landed; the fix is in open PR #7741, which is not resolution.

Both halves of the finding still hold. The allowlist still has all five entries:

```
$ git show origin/main:.github/scripts/fleet-attestation-placeholders.txt | grep -c '^265175468565'
5
```

and the checker still consults it in only one direction — an entry is read solely when an image ref appears in the gitops scan, so a ref that is no longer declared reaches no branch at all and is reported nowhere:

```
$ git show origin/main:.github/scripts/check-fleet-attestations.sh | sed -n '399,401p;454,455p'
is_allowed_placeholder() {
  [ -f "$PLACEHOLDER_FILE" ] || return 1
  grep -vE '^[[:space:]]*(#|$)' "$PLACEHOLDER_FILE" | grep -qxF "$1"
[...]

PR #8936, opened 2026-09-06
docs(adr): ADR-0285 Communication Studio — governed style and playbook layer

JiRaska:
## Summary

ADR-0285 proposes a **Communication Studio**: every conversational prompt splits into an immutable git **core** (safety, injection defence, tool routing, HITL/SCA) and a business-editable **style** + **playbook** layer served by a new `openbank-communication-service` and edited in the admin-ui under four-eyes with an evals replay on publish. Documentation only — no code, no behaviour change.

## Type of change

- [ ] feat (new functionality)
- [ ] fix (bug fix)
- [ ] refactor (no behavior change)
- [ ] perf (performance)
- [x] docs
- [ ] chore (build / tooling)
- [ ] security (private until disclosed)
- [ ] breaking change

## Linked issues

Refs #8935

## Changes

- `docs/adr/0285-communication-studio-governed-style-and-playbook-layer.md` — new ADR, `proposed` / `planned`.
- `docs/adr/README.md`, `DIGEST.md`, `index.json` — regenerated by `docs/adr/gen-index.sh` (derived, never hand-edited).

Decisions D1–D8: three-layer prompt with the core always first and never writable through any API; persona as the unit of style, one per channel, shared by bot and staff; a deterministic save-time linter plus `commstyle.publish` four-eyes (the ADR-0176 D5 pattern) so a business editor cannot weaken the core; immutable append-only versions with an evals replay against a golden set at publish time; consumers cache the published version and fall back to the git baseline when the service is down; roles from `rules.yaml` per ADR-0229, never granted to a service account; a `/communication` admin-ui workspace; marketing copy and core editing explicitly out of scope.

## Definition of Done

- [x] Hexagonal layering preserved (domain has zero framework imports).
- [ ] Unit tests added/updated.
- [ ] Integration tests added/updated (if service boundary involved).
- [ ] Contract tests updated (if public API changed).
- [x] `./gradlew detekt ktlintCheck koverVerify build` passes.
- [x] No new lint warnings.
- [ ] OpenAPI spec regenerated (if REST API changed).
- [ ] Flyway migration added + rollback note (if DB changed).
- [ ] Avro schema versioned backward-compatibly (if event changed).
- [x] Docs updated (README, ADR if architectural).

No code changes in this PR, so the test and schema rows are not applicable; the ADR registry check and `gen-index.sh` were run locally.

## Security checklist

- [x] No secrets, tokens, passwords, or PII in code, config, tests, logs.
[...]

PR #9060, opened 2026-09-07
ci(governance): unblock main — stale attestation placeholder + stale ADR registry

JiRaska:
## Summary

`Verify fleet attestations` is red on `main`, and therefore on every open PR. It surfaced on #8909
— a transaction-service categorisation PR with no connection to kyb — which is what a fleet-wide
baseline failure looks like from inside an unrelated branch.

The placeholder allowlist pins `openbank-kyb-service:sandbox-26f33679`. gitops now declares
`sandbox-4b9687f9`:

```
openbank-infra/gitops/components/kyb/kyb-service.yaml:60:
  image: 265175468565.dkr.ecr.eu-north-1.amazonaws.com/openbank-kyb-service:sandbox-4b9687f9
```

So the entry names an image nothing deploys — exactly the state the ratchet from #7740 was built to
fail on.

## Why this is a baseline fix and not a supply-chain finding

The gate says so itself: *"every declared image is attested, but the placeholder allowlist carries
entr(y/ies) whose image is no longer declared ... nothing is undeployable, the BASELINE is wrong,
and the fix is a file deletion rather than a rebuild."* Nothing is being suppressed here; the
strict-scope rule (an allowlist entry may only ever cover an image ABSENT from the registry, never
an unattested one) is untouched.

## Why the entry existed, and why it outlived its purpose

It was correct when written — the tag was #8863's own commit and no build existed for it. Its own
comment carried the exit condition: *"Remove this line with that gitops PR; the ratchet fails on a
stale entry, so it cannot outlive the first real build."* The auto-deploy gitops PR rewrote the pin
to the built sha and the line stayed. The ratchet then did precisely what that sentence promised.

The file returns to zero entries, which its header already documents as the correct reading rather
than a missing file.

## Verification

```
==> Placeholder allowlist: 0 entr(y/ies), 0 stale
PLACEHOLDER ALLOWLIST: PASS — every entry still names an image declared in openbank-infra/gitops.
```

Before the change the same command reported `1 entr(y/ies), 1 stale` and FAIL.

Note the attestation half of that script cannot be run locally (no registry credentials — it
reports `65 unknown (probe failed)` and says outright that this is a tooling failure, not a
finding). The allowlist half is decidable from the files alone, and that is the half this PR
touches.

Refs #8863


JiRaska:
**Correcting a claim in the PR description above.** It says the gate is red "on every open PR". That
is not what the data shows, and the real shape is more precise and more useful.

Measured across the 18 open PRs that carry a `Verify fleet attestations` result:

```
9051  FAILURE  2026-09-07T11:52:35Z     <- see note
9059  FAILURE  2026-09-07T12:29:21Z
9052  FAILURE  2026-09-07T12:31:28Z
8909  QUEUED   2026-09-07T12:43:25Z
9056  SUCCESS  2026-09-07T12:19:09Z     <- newest passing run
8920  SUCCESS  2026-09-07T12:11:26Z
... 13 more SUCCESS, all 2026-09-06 to 2026-09-07T11:26Z
```

The pin moved at **2026-09-07T12:19:29Z**:

```
8618af1cf chore(gitops): auto-deploy sandbox-4b9687f93cefe73517c9f75b8de41c7d70a9ac60 (#9041)
```

So the split is by **run start time, not by PR**: every run that started after that commit fails,
and every green row predates it. The fifteen SUCCESS results are stale, not healthy — none of them
has seen the current `main`, and each will flip the moment its PR is pushed to or rerun. Counting
"3 of 18 failing" and concluding the blast radius is small would be the wrong reading, which is why
the timestamps matter more than the tally here.

(#9051's failure at 11:52 predates the pin move, so it is a separate cause and is not evidence for
this one — I have not looked into it.)

JiRaska:
**#9064 folded into this PR, because the two were deadlocked.**

`main` carries two independent fleet-wide reds right now, and each of the two fixes inherited the
other's failure. Measured on both PRs:

```
#9060  gates (lint-supplychain-security)  SUCCESS      <- its own fix
       gates (registry-kotlin-data)       FAILURE      <- the ADR registry, from main
#9064  gates (registry-kotlin-data)       SUCCESS      <- its own fix
       gates (lint-supplychain-security)  FAILURE      <- the attestation placeholder, from main
```

Perfectly symmetric: neither could ever go green on its own, so neither could merge, so `main` would
have stayed red under both. Splitting them was the wrong call and this corrects it — #9064's commit
is cherry-picked here unchanged and #9064 is closed pointing at this PR.

Second fix, from #9064: `docs/adr/README.md` and `index.json` say 277 ADRs while the tree holds 278.
Two PRs each regenerated those derived files against a different ADR set (#8978 at 11:47 with
ADR-0289, #8936 at 14:16 with ADR-0285); git merged the two regenerated pairs as text without a
conflict and the result agrees with neither.

Both halves verified on this branch's head:

```
check-adr-registry: OK — 278 ADRs: ... derived artefacts fresh.
PLACEHOLDER ALLOWLIST: PASS — every entry still names an image declared in openbank-infra/gitops.
```

JiRaska:
Superseded by #9061, which landed both fixes at 15:35:20 — the same 8-line placeholder removal and the same one-line README/index regeneration, byte-for-byte the same end state (`fleet-attestation-placeholders.txt` now has 0 entries, README says 278 ADRs). Verified on live main: `check-adr-registry` and `check-fleet-attestations --check-placeholders` both exit 0.

This PR was a duplicate I should have caught — #9061 was already open when I opened this one, and it appeared in my own listing of attestation runs. Closing rather than merging; the branch is left in place.

PR #9064, opened 2026-09-07
docs(adr): regenerate the registry — main says 277 ADRs and holds 278

JiRaska:
## Summary

`adr-registry-integrity-check` fails on `main` itself, and therefore on every PR that reaches it.
`docs/adr/README.md` and `docs/adr/index.json` still carry the pre-ADR-0285 count:

```
-**Reading 277 ADRs:** don't. [DIGEST.md](DIGEST.md) is the entire decision history
+**Reading 278 ADRs:** don't. [DIGEST.md](DIGEST.md) is the entire decision history
```

## Cause

Two PRs each regenerated the same derived files against a different ADR set:

```
2026-09-07T14:16:18  13d99b078  docs(adr): ADR-0285 Communication Studio ... (#8936)
2026-09-07T11:47:48  b776e59c5  fix(sdd): replay-safe collection authorisation ... (#8978)
```

(#8978's squash title still says ADR-0285; the file it merged is `0289-sdd-...`. The number moved
during review and the title did not follow — worth knowing when reading that commit later.)

Git merged the two regenerated README/index pairs as text, without a conflict, and the result agrees
with neither. This is the derived-file version of the failure mode the repo already documents for
list files: a clean merge is not a correct merge.

## Order matters here

Regenerate → stage → commit, never regenerate → check → commit. `check-adr-registry.sh` restores the
three derived files to HEAD when it fails, so running the checker first silently reverts the
regeneration and the later `git add` stages the old content.

## Verification

```
check-adr-registry: OK — 278 ADRs: unique numbers, headings match filenames, front-matter valid,
supersession bidirectional, derived artefacts fresh.
```

Run against `origin/main` before the change, the same command exits 1 with README.md and index.json
reported stale.

Note this is one of **two** independent fleet-wide reds on `main` right now; the other is the stale
attestation placeholder in #9060. They are unrelated causes and neither fixes the other.


JiRaska:
Folded into #9060. The two PRs were deadlocked: this one was red on `gates (lint-supplychain-security)` (the stale attestation placeholder, from main) while #9060 was red on `gates (registry-kotlin-data)` (this PR's own subject, from main). Each fixed exactly the gate the other failed on, so neither could ever go green alone. The commit here is cherry-picked into #9060 unchanged; both gates pass on that head. Not deleting this branch.

Issue #9003, opened 2026-09-07
lending_outbox: 44 loan.provisioned rows are stamped 1970 and carry no occurredAt — #3272's defect, in another service


JiRaska:
## What

44 of the 45 rows in `lending_outbox` on the sandbox carry `created_at`, `updated_at` and `sent_at` of **1970-01-01**, and their payloads carry no `occurredAt` at all. Measured 2026-09-07 on `lending-db-1`:

```
          event_type           | status | count | oldest     | last_sent
 loan.provisioned              | SENT   |    44 | 1970-01-01 | 1970-01-01 00:00:00+00
 credit.application.transition | SENT   |     1 | 2026-09-03 | 2026-09-03 12:14:53+00
```

A sample payload — note there is no `occurredAt` key:

```json
{"loanId":"02b36126-...","period":"2026-07","stage":"STAGE_1",
 "expectedCreditLoss":"1869.63 CZK","delta":"1869.63 CZK"}
```

## This is the defect #3272 already fixed once, recurring in another service

`OutboxMessage.createdAt` on `main` now defaults to `Instant.now()`, and its KDoc records exactly this failure mode from ledger-service — 388 of 553 rows stamped 1970 because `toEntity()` assigns the field over the column's `DEFAULT now()`, so an omitted value is silently EPOCH rather than an error. `LendingOutboxRepositoryImpl.toEntity()` does the same assignment (`it.createdAt = createdAt`).

The 44 lending rows predate that default, so this is a **data** artefact rather than live code today — but nobody had looked, and the rows are still there being dispatched in the wrong order.

## Why it matters beyond tidiness

**Dispatch starvation.** The dispatcher claims with `ORDER BY created_at ASC`, so 44 epoch rows sort permanently ahead of every correctly-stamped event. On this instance they are all `SENT`, so nothing is stuck now; on any instance where one fails and retries, real traffic queues behind a 1970 backlog.

**Warehouse misplacement, which is new since yesterday.** `bronze_events` is `PARTITION BY toYYYYMM(occurred_at)`, and the credit marts merged in #8939 group by `toStartOfMonth(occurred_at)` and by disbursement month. An epoch-stamped provisioning event lands in the 1970-01 partition and shows up in a vintage curve 56 years from where it belongs. The marts cannot defend against this: they read the timestamp the producer wrote.

[...]

Issue #3272, opened 2026-08-02
libs-domain: OutboxMessage.createdAt defaults to Instant.EPOCH — 388/553 ledger outbox rows are stamped 1970 and jump the FIFO claim order


JiRaska:
### Affected service / module

openbank-libs-domain (`persistence/outbox/OutboxStatus.kt`) — consequences fleet-wide

### Version / commit

main @ 63b6cd8dd (measured live in sandbox, 2026-08-02)

### What happened?

`OutboxMessage.createdAt` defaults to `Instant.EPOCH`:

```kotlin
data class OutboxMessage(
    val eventId: UUID = UUID.randomUUID(),
    val aggregateId: UUID,
    val eventType: String,
    val payload: String,
    val createdAt: Instant = Instant.EPOCH,   // <- silent 1970 when the caller omits it
)
```

Most call sites omit it. Each service's `toEntity()` then assigns `it.createdAt = createdAt`,
which **overrides the column's `DEFAULT now()`** — so the row is stamped `1970-01-01 00:00:00+00`
and no error is raised anywhere.

This is not a cosmetic timestamp problem. The dispatcher claims work with

```sql
... ORDER BY created_at ASC LIMIT :claimLimit FOR UPDATE SKIP LOCKED
```

so every epoch-stamped row sorts **ahead of every correctly-stamped one, permanently**. Where a
service mixes both (see below), correctly-stamped events are starved behind a 1970 backlog rather
than dispatched FIFO. The `openbank.outbox.backlog` age signal is equally meaningless — a fresh row
reads as 56 years old.

### Measured, not inferred

Row counts from the live sandbox databases, `created_at < '2000-01-01'` over total:

| service | epoch / total |
|---|---|
| ledger | **388 / 553** |
| transaction | **176 / 304** |
| sanctions | **66 / 163** |
| aml | **31 / 108** |
| domestic-payment | 13 / 87 |

ledger is the instructive one: it is **mixed**. `FxRevaluationService` passes `createdAt`; the other
five ledger call sites do not. So within a single outbox the two populations coexist and the FIFO
order is not merely shifted — it is inverted for 70% of the table.

Static count across `openbank-*/src/main/kotlin`: roughly **48 `OutboxMessage(...)` call sites omit
`createdAt`, across ~17 services**. Only sepa-payment, dispute-service, consent-service, sca-service
and one ledger call site pass it. (The static figure is a grep-derived estimate and should be
[...]

PR #9079, opened 2026-09-07
fix(lending): correct the 44 epoch-stamped outbox rows and guard the column

JiRaska:
Refs #9003 (closes the code+data half; the sandbox row correction lands with the deploy).

## What
44 of 45 `lending_outbox` rows carried 1970-01-01 — the #3272 defect class recurring in lending. Two consequences the issue names: epoch rows sort permanently ahead of real work in dispatcher claims (`ORDER BY created_at ASC`), and a replay would land them in the 1970-01 `bronze_events` partition, 56 years from where the provisioning events belong.

## Verified first: the live half is closed
No `src/main` construction site of `OutboxMessage` passes an explicit `createdAt` anywhere in the fleet (grep-verified); `LendingOutboxMessage` is a typealias of the shared `OutboxMessage` whose default is now `Instant.now()` with the #3272 KDoc.

## Fix (V16)
- `UPDATE` the epoch rows to the correction instant (they are all `SENT`; true business time is unrecoverable and the UPDATE is itself the audit trail).
- `CHECK (created_at >= 2020-01-01)` — the issue's suggested guard, turning a future epoch row into an INSERT-time error.

## Falsification honoured (not "the default is now()")
`LendingOutboxWriteIT` now asserts the row written through the real emitter has `created_at > 2020-01-01`. Migration validated standalone on a fresh PostgreSQL 16 container: all 16 migrations apply cleanly, the CHECK rejects a 1970 insert, accepts the default `now()` insert.

## Local-env caveat
The lending IT suite could not finish locally: this machine runs ~8 Gradle daemons for parallel sessions and Quarkus boot died on OIDC/Vert.x timeouts twice (Flyway connected and applied without error). CI runs the real gate.

## Security checklist
- [x] No secrets, credentials, or PII added
- [x] No validation loosened — a DB constraint added
- [x] No new outbound edges
- [x] Data correction touches only demonstrably-wrong rows (created_at < 2020), all SENT

### DIFF
--- a/openbank-lending-service/src/main/resources/db/migration/V16__outbox_created_at_plausible.sql
+++ b/openbank-lending-service/src/main/resources/db/migration/V16__outbox_created_at_plausible.sql
@@ -1,0 +1,24 @@
+-- #9003: 44 of 45 lending_outbox rows on the sandbox carry created_at/updated_at/sent_at of
+-- 1970-01-01 — the #3272 defect class (toEntity() assigning over the column DEFAULT) from before
+-- OutboxMessage.createdAt gained its Instant.now() default. No src/main construction site passes
+-- an explicit createdAt today (grep-verified), so this migration is the data half plus the guard.
+--
+-- Data: the rows are all SENT, so correcting them changes no pending dispatch order; leaving them
+-- would keep 44 epoch rows sorting ahead of real work forever (dispatcher claims ORDER BY
+-- created_at ASC) and landing in the 1970-01 bronze_events partition if ever replayed into the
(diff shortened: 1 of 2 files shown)


JiRaska:
CI note: remaining failures are the known environment-bound ones tracked in #8690/#9031 — Dependency review (dependency graph not enabled on the repo), Trivy SARIF upload and security-regression-test (GitHub installation rate limit, exhausted by the ~434 workflow runs/hour fan-out). all-green and every enforced gate pass.

PR #9080, opened 2026-09-07
fix(onboarding): an APPROVED kyc no longer moves the funnel board backwards

### LINKED ISSUE #8951: onboarding: an APPROVED KYC derives to REGISTERED, moving the funnel board backwards
## What

`FunnelStage.derive` maps an **APPROVED** KYC on a not-yet-active party to `REGISTERED` —
the *first* column of the onboarding funnel. The cockpit board therefore moves a customer
**backwards**, from `KYC_UNDER_REVIEW` to `REGISTERED`, at the exact moment their KYC is
approved, and keeps them there until party-service emits `PARTY_STATUS_CHANGED`.

## Where

`openbank-onboarding-service/src/main/kotlin/com/openbank/onboarding/domain/model/OnboardingRecord.kt:62`

```kotlin
fun derive(party: PartyStage, kyc: KycStage?, scaEnrolled: Boolean): FunnelStage = when {
    party == PartyStage.SUSPENDED || party == PartyStage.CLOSED -> BLOCKED
    party == PartyStage.ACTIVE && scaEnrolled -> ACTIVE
    party == PartyStage.ACTIVE && !scaEnrolled -> SCA_PENDING
    kyc == KycStage.UNDER_REVIEW -> KYC_UNDER_REVIEW
    kyc == KycStage.DOCUMENTS_REQUIRED -> KYC_DOCUMENTS_REQUIRED
    kyc == KycStage.OPEN || kyc == null -> KYC_OPEN
    kyc == KycStage.REJECTED || kyc == KycStage.EXPIRED -> BLOCKED
    else -> REGISTERED
}
```

`KycStage` has six values — `OPEN`, `DOCUMENTS_REQUIRED`, `UNDER_REVIEW`, `APPROVED`,
`REJECTED`, `EXPIRED`. Five of them have a branch. **`APPROVED` has none**, so it falls
through to `else -> REGISTERED`.

## Why it is reachable, not an edge case

`APPROVED` + `PENDING_KYC` is the *normal* ordering, not a race: kyc-service approves, and
party-service activates afterwards. `OnboardingProjectionService.kt:227` derives on exactly
that transition:

```kotlin
funnelStage = FunnelStage.derive(existing.partyStatus, event.newStatus, existing.scaEnrolled),
```

So every approved application passes through the wrong stage. The KDoc immediately above
the enum declares the ordering as
`REGISTERED → KYC_OPEN → KYC_UNDER_REVIEW → SCA_PENDING → ACTIVE | BLOCKED`,
which the observed transition violates.

## Evidence

Observed from a real test run, not from reading alone. Two independent assertions in a
draft `FunnelStageDeriveTest` failed, and they fail in the two directions that pin the
behaviour:

- `an APPROVED kyc on a party party-service has not activated yet stays in the KYC column`
  failed — `derive(PENDING_KYC, APPROVED, false)` did not return `KYC_OPEN`.
- `derive is total - no combination of the three dimensions is left underived` failed on
  `assertThat(stages).doesNotContain(FunnelStage.REGISTERED)` — `REGISTERED` **is**
  produced by some input combination.

Together those two say `derive(PENDING_KYC, APPROVED, *) == REGISTERED`.

## What the fix is NOT

`KYC_OPEN` is not the right answer either — the KYC is approved, not open. There is no
stage in the current enum meaning *"KYC complete, awaiting party activation"*. So the fix
is a design decision, one of:

1. add an explicit branch `kyc == KycStage.APPROVED -> KYC_UNDER_REVIEW` (keeps the customer
   in the KYC column until activation — no enum change, no board column change), or
2. add a new `KYC_APPROVED` stage between `KYC_UNDER_REVIEW` and `SCA_PENDING` (honest, but
   it is a new board column and an admin-ui change), or
3. keep `REGISTERED` and correct the documented ordering — only defensible if someone
   deliberately wants the backwards jump, which seems unlikely.

Whichever is chosen, `else -> REGISTERED` should become unreachable-by-construction (an
exhaustive `when` over `KycStage`), so the next value added to the enum is a compile error
rather than a silent fall-through to the first funnel column. That fall-through is the
actual root cause: a catch-all `else` on a closed enum.

## Not fixed here

Found while adding unit tests in #8937, which is deliberately test-only. The draft test
that surfaced this asserted the wrong expected value (`KYC_OPEN`) and carried a comment
claiming `else -> REGISTERED` was unreachable — both wrong; only the *failure* was
informative. Those two assertions are removed from #8937 rather than committed red or
pinned to the buggy value, so nothing in that PR cements this behaviour.

Refs #8937

JiRaska:
Closes #8951

## Root cause
`FunnelStage.derive` had branches for five of six `KycStage` values — **APPROVED had none**, so `APPROVED` + not-yet-`ACTIVE` party (the *normal* ordering: kyc-service approves, party-service activates afterwards) fell through `else -> REGISTERED`, the first funnel column. The board moved a customer backwards at the exact moment of approval.

## Fix (issue's option 1)
- `APPROVED → KYC_UNDER_REVIEW`: `KYC_OPEN` would lie (the KYC is not open); a new `KYC_APPROVED` stage is a board/admin-ui change nobody asked for.
- The kyc dimension is now an **exhaustive `when`** — the catch-all `else` on a closed enum was the actual root cause; the next `KycStage` value is a compile error, not a silent fall-through.

## Tests
- APPROVED stays in the KYC column for a non-active party (both `scaEnrolled` states).
- Totality: `derive` never produces `REGISTERED` for any combination of the three dimensions — it is written once at `PartyCreated` and the funnel only moves forward.

Local: unit tests + ktlint + detekt green. IT suite could not finish locally (machine saturated by parallel agent sessions; the single failure is a 58s SocketTimeout in `OnboardingStageFilterIT`, pre-existing env flake) — CI runs the real gate.

## Security checklist
- [x] No secrets, credentials, or PII
- [x] No validation changed; pure domain derivation fix
- [x] No new outbound edges
- [x] Customer-facing correctness fix (board no longer regresses on approval)

### DIFF
--- a/openbank-onboarding-service/src/main/kotlin/com/openbank/onboarding/domain/model/OnboardingRecord.kt
+++ b/openbank-onboarding-service/src/main/kotlin/com/openbank/onboarding/domain/model/OnboardingRecord.kt
@@ -63,11 +63,21 @@ enum class FunnelStage {
             party == PartyStage.SUSPENDED || party == PartyStage.CLOSED -> BLOCKED
             party == PartyStage.ACTIVE && scaEnrolled -> ACTIVE
             party == PartyStage.ACTIVE && !scaEnrolled -> SCA_PENDING
-            kyc == KycStage.UNDER_REVIEW -> KYC_UNDER_REVIEW
-            kyc == KycStage.DOCUMENTS_REQUIRED -> KYC_DOCUMENTS_REQUIRED
-            kyc == KycStage.OPEN || kyc == null -> KYC_OPEN
-            kyc == KycStage.REJECTED || kyc == KycStage.EXPIRED -> BLOCKED
-            else -> REGISTERED
+            else ->
+                // Exhaustive on purpose (#8951): the previous catch-all `else -> REGISTERED` is
+                // the actual root cause — APPROVED had no branch and fell through to the FIRST
+                // funnel column, so the board moved a customer backwards at the exact moment
+                // their KYC was approved (APPROVED + not-yet-ACTIVE is the normal ordering:
+                // kyc-service approves, party-service activates afterwards). A new KycStage
+                // value is now a compile error here, not a silent fall-through.
+                when (kyc) {
+                    null, KycStage.OPEN -> KYC_OPEN
+                    KycStage.DOCUMENTS_REQUIRED -> KYC_DOCUMENTS_REQUIRED
+                    // APPROVED stays in the KYC column until party activation — KYC_OPEN would
+                    // lie (the KYC is not open) and there is no KYC_APPROVED board column.
+                    KycStage.UNDER_REVIEW, KycStage.APPROVED -> KYC_UNDER_REVIEW
+                    KycStage.REJECTED, KycStage.EXPIRED -> BLOCKED
+                }
         }
     }
 }
--- a/openbank-onboarding-service/src/test/kotlin/com/openbank/onboarding/application/usecase/OnboardingProjectionServiceTest.kt
+++ b/openbank-onboarding-service/src/test/kotlin/com/openbank/onboarding/application/usecase/OnboardingProjectionServiceTest.kt
@@ -50,6 +50,29 @@ class OnboardingProjectionServiceTest {
     }
 
     @Test
+    fun `an APPROVED kyc on a not-yet-active party stays in the KYC column`(): Unit = runBlocking {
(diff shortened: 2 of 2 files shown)


JiRaska:
CI note: remaining failures are the known environment-bound ones tracked in #8690/#9031 — Dependency review (dependency graph not enabled on the repo), Trivy SARIF upload and security-regression-test (GitHub installation rate limit, exhausted by the ~434 workflow runs/hour fan-out). all-green and every enforced gate pass.

PR #8994, opened 2026-09-06
chore(fleet): ratchet 16 kover floors to measured coverage

JiRaska:
This PR adds unit coverage across low-floor modules and raises 16 Kover minimums. It targets `main` and is the surviving PR for the test sweep formerly duplicated in #8937. The complete comparison is recorded at https://github.com/JiRaska/open-bank-oss/pull/8937#issuecomment-5654244332; #8937 is closed as superseded. This is a combined test-and-floor diff, not a floor-only stacked diff.

The proposed minimums are: agent-service 83, audit-service 87, authz-policy-auditor 85, control-liveness-sentinel 47, devops-agent 77, document-service 83, finops-agent 81, fx-service 80, governance-auditor 81, kyc-service 79, libs-domain 89, notification-service 72, onboarding-service 75, release-steward 80, security-scanner 96 and tpp-registry-service 76. All exceed their corresponding current-main values. These are the branch's actual settings; the earlier measurement table has not been regenerated for the entire current fleet.

The Temporal workflow test for flaky-test-hunter now uses a Kotlin-aware payload converter matching the production client's configuration. Previously, deserializing `TestScanSnapshot` failed repeatedly until the 45-minute CI job limit. A 20-second JUnit timeout now bounds each workflow test without removing its business assertions.

Validation of commit `6b38c4db652c016e1b4d89887bd81d4a624fa095`:

- Reproduced the original deserialization failure with the bounded test; it failed with `TimeoutException`.
- All four workflow scenarios pass after the converter fix; module ktlint and detekt pass.
- Full flaky-test-hunter suite: 105 tests discovered, 104 executed, zero failures/errors, one skipped broker-backed Pact test.
- Module `koverVerify` and `quarkusBuild` pass. The full command completed in 2m26s.
- This is module-level proof. Fresh fleet CI is still required for the whole PR. The preceding run also had a cancelled verification-metadata job; that is a separate unresolved execution gap.

The closed companion #8937 has no unique test behavior missing here: its residual differences are comments, formatting, import order and an equivalent JSON fixture. TPP module `ktlintCheck` also passes on this head unchanged. The coverage increase and test additions remain distinct review decisions even though this branch contains both. No production source, API, database or event behavior changes in this diff. The standalone scanner settings change is trailing whitespace only.

Refs #265.

## Security checklist

- [x] No secrets, tokens, passwords, or PII in code, config, tests, logs.
- [x] No new suppression without justification.
- [x] No new third-party dependency.
[...]

PR #9085, opened 2026-09-07
fix(libs): strip the openbank- prefix from the kyb topic-producer row

JiRaska:
## Summary

`TopicProducersTest`'s own invariant ("no row carries the openbank- prefix") has been red on `main` since #8920 landed both the table and that test in the same commit. Path-scoped CI never caught it because `openbank-libs-domain` only builds when that module's own directory changes or on a code-global (full-fleet) trigger. Found while an unrelated PR touching `openbank-libs-runtime` triggered exactly that full-fleet build.

## Type of change

- [x] fix (bug fix)
- [ ] perf (performance)
- [ ] chore (build / tooling)
- [ ] docs
- [ ] breaking change

## Linked issues

Refs #8920

## Changes

- `TopicProducers.kt` — the kyb row loses the `openbank-` prefix, matching every other row and the table's own documented rule (enforced fleet-wide by `check-source-service-convention.py`).
- `KybAnalyticsIngestTest.kt` — its assertion is corrected to match, with the misdiagnosis it was based on rewritten in the comment.

## Why the prefixed value was wrong, not just inconsistent

The row's comment cited `KybAnalyticsIngestTest.kt` as evidence the prefix was correct. That test asserts `TopicAttribution.sourceService` in `openbank-analytics-sink`, a different object that happens to delegate straight through to `TopicProducers.sourceService` — so the two ARE linked, and the comment was right about that, wrong about which side was authoritative.

Checked against the producer itself rather than assumed: `openbank-kyb-service`'s `KybEvent.SOURCE_SERVICE` — what it actually stamps into every event body — is `"kyb-service"`, no prefix. This table's own row is a fallback that never fires for kyb anyway (the body already carries the field), which is exactly the "harmless but pointless" case the table's own KDoc describes for a producer that self-stamps.

## Definition of Done

- [x] Hexagonal layering preserved (domain has zero framework imports) — no import added.
- [x] No new lint warnings.
- [x] Docs updated (README, ADR if architectural) — comment corrected in place.

`openbank-libs-domain`'s and `openbank-analytics-sink`'s own test suites are the verification; both modules build on this PR.

## Security checklist

- [x] No auth / crypto / payment code changed: the delegation-service commit is test-only (fixes a wall-clock time bomb in `SpendReservationServiceTest`, cherry-picked from #9086); the kyb-prefix commit touches a fallback-attribution string, never a decision path.
- [x] No PII path changed.
- [x] No cardholder-data path changed.

## Compliance impact

- [ ] PCI DSS
- [ ] DORA
- [ ] GDPR
- [ ] PSD2 / SCA / RTS
- [ ] AML / 5AMLD
- [ ] CNB reporting
- [x] None

This is a fallback-attribution string for analytics/audit event rows, not a new data category or processing purpose.

## Rollout / Rollback

- Deployment strategy: no runtime change — pure library code, exercised only by tests.
- Rollback plan: revert; the table returns to the (broken) prefixed value.
- Data migration reversible: N/A — this fallback has never fired for kyb events, since the producer self-stamps the field.

## Reviewer notes

Found and fixed while unblocking an unrelated PR (#9025) whose full-fleet trigger surfaced this pre-existing red.

### DIFF
--- a/openbank-analytics-sink/src/test/kotlin/com/openbank/analytics/application/KybAnalyticsIngestTest.kt
+++ b/openbank-analytics-sink/src/test/kotlin/com/openbank/analytics/application/KybAnalyticsIngestTest.kt
@@ -36,7 +36,13 @@ class KybAnalyticsIngestTest {
     fun `a kyb record attributes to the kyb service without an override`() {
         assertThat(TopicAttribution.domainOf("openbank.kyb.events")).isEqualTo("kyb")
         assertThat(TopicAttribution.aggregateType("openbank.kyb.events")).isEqualTo("KYB")
-        assertThat(TopicAttribution.sourceService("openbank.kyb.events")).isEqualTo("openbank-kyb-service")
+        // NOT "openbank-kyb-service": TopicProducers.sourceService (delegated to here) strips the
+        // module prefix like every other row, per its own documented rule and per
+        // check-source-service-convention.py -- and KybEvent.SOURCE_SERVICE, what kyb-service
+        // actually stamps in the body, is "kyb-service" too. This test's earlier value was wrong,
+        // and TopicProducersTest's "no row carries the openbank- prefix" was the invariant that
+        // caught it (issue found 2026-09-07).
+        assertThat(TopicAttribution.sourceService("openbank.kyb.events")).isEqualTo("kyb-service")
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


JiRaska:
Same priority note as #9086: this fixes `build (openbank-libs-domain)`, which is red on `main` and therefore on every PR whose diff reaches that module — observed tonight on #8909, #8994 and #9059, none of which touches the topic-producer table. Between this PR and #9086 (delegation), most of the red currently in the open queue is accounted for by two small fixes to main.

PR #9086, opened 2026-09-07
test(delegation): anchor the real-clock spend test to the real clock

JiRaska:
## Summary

`SpendReservationServiceTest > occurredAt is a real clock reading, not a default` **began failing
today at 12:00Z** and fails from now on — on `main` and on every PR that builds
openbank-delegation-service.

## The mechanism

The class runs on a fixed clock and `grant()` derives its validity window from it:

```kotlin
private val clock: Clock = Clock.fixed(Instant.parse("2026-08-08T12:00:00Z"), ZoneOffset.UTC)
...
validFrom = now.minusDays(1),
validTo   = now.plusDays(30),     // => 2026-09-07T12:00:00Z
```

One test in that class deliberately uses `Clock.systemUTC()`, because it exists to prove `occurredAt`
is a real reading rather than an `Instant.EPOCH` default (#3874/#3883) — and only recency can tell
those apart, so a fixed clock would make it vacuous.

Those two facts were fine together for thirty days. Once wall-clock now passed the fixed grant's
expiry, the reservation was refused before the assertion was ever reached:

```
SpendReservationServiceTest > occurredAt is a real clock reading, not a default() FAILED
    com.openbank.delegation.application.usecase.SpendReservationRefusedException
    at SpendReservationServiceTest.kt:303
```

The failure is an **expiry**, not the property under test. It is a time bomb that armed itself on the
day the fixture was written and fired 30 days later, and the arithmetic says so without needing a
test run: `2026-08-08T12:00Z + 30d = 2026-09-07T12:00Z`, which is behind now.

## The fix

`grant()` gains an `at` parameter defaulting to the fixed `now`, so every other test in the class is
unchanged in behaviour, and the real-clock test anchors its grant to
`OffsetDateTime.now(ZoneOffset.UTC)`.

The general rule the comment records in place: **a test that deliberately mixes a real clock with
fixed-clock fixtures has to anchor both to the same clock**, or the fixture's window silently becomes
a deadline.

## What this is not

#8909 surfaced it, and it is not #8909's defect: that PR changes two OPA checksum lines under
`openbank-infra/gitops/components/delegation/` and no Kotlin whatsoever. Path-scoped CI builds the
module because the directory changed.

## Verification

The diagnosis is established by the arithmetic above and by the exception in the CI log, both of
which are independent of any local run. A local `:openbank-delegation-service:test
--tests '*SpendReservationServiceTest*'` was still executing when this PR was opened (the machine is
running two other Gradle builds), so **the CI run on this PR is the check that settles it** — I am
not claiming a green local run I have not seen.

## Security checklist

- [x] No secrets, tokens, passwords, or PII in code, config, tests, logs. A test fixture's clock only.
- [x] No new `@SuppressWarnings`, `as Any`, `@Suppress("...")` without justification. None added.
- [x] No new third-party dependency. None.
- [x] Auth / crypto / payment code changed? → tag `security-review-required`. **No.** The diff is one
      test file; no production code, no delegation limit, window or authorisation logic changes. The
      grant fixture's validity window moves from a fixed clock to the wall clock **inside a single
      test**, which is what makes that test able to fail again rather than expire.
- [x] PII path changed? → tag `gdpr-review-required`. **No.**
- [x] Cardholder data path changed? → tag `pci-review-required`. **No.**

*(delegation-service is money-path, which is why this section is required; the change itself is
test-only. The section was missing because the PR was created with `--body-file`, which bypasses the
template — issue #8757.)*


JiRaska:
**Priority note for whoever reviews first thing:** this one-file test fix is what is keeping
`build (openbank-delegation-service)` red on `main`, and therefore on nearly every open PR that
touches anything delegation-adjacent. Observed tonight across my own queue alone:

```
#8888  #8904  #8909  #8927  #8937  #9049  #9059   -> build (openbank-delegation-service) FAILED
```

None of those PRs has anything to do with delegation; path-scoped CI builds the module because
their diff touches the directory or a dependency, and the module has been red since **2026-09-07
12:00Z**, when the fixed-clock grant in `SpendReservationServiceTest` expired against the wall clock
the one real-clock test deliberately uses.

The change is a test fixture parameter — `grant(at = ...)`, defaulting to the existing fixed clock so
every other test in the class is unchanged — and the PR is green: 31 checks, 0 failing.

`build (openbank-libs-domain)` is red on the same PRs for the unrelated `TopicProducers` prefix
contradiction, which #9085 fixes. The two together account for most of the red in the queue right
now.

The merge hook correctly refuses to let an agent merge this (delegation-service is money-path), which
is why it is sitting here rather than already landed.

JiRaska:
Superseded — #9093 merged an equivalent fix for the same wall-clock time-bomb (`fa6d976d4`), so this branch is now CONFLICTING against main with no remaining delta. Leaving this for a human to close rather than doing it myself.

JiRaska:
Superseded — #9093 merged at 2026-09-07T21:29Z and `main` now carries the same fix:

```kotlin
val wallClock = Clock.systemUTC()
currentGrant = grant(at = OffsetDateTime.now(wallClock))
```

Same mechanism as this PR (an `at` parameter on `grant()` defaulting to the fixed clock, with the wall-clock test anchoring its own window), and its comment adds something mine did not: it names #8310 and #9015 as the two PRs that went green either side of main going red, which is the clearest statement of why a dated fixture is worse than an ordinary broken test — it fails on a calendar date rather than on a change.

This branch is now DIRTY against that, and re-resolving it would only reintroduce a second copy of one fix. Closing rather than merging.

For the record: I closed #9093 as the duplicate yesterday and kept this one, on the grounds that it was opened first and had more CI history. That call was reversed by whoever reopened and merged it, and the outcome is the right one — the defect is fixed on main, which is what mattered. Not deleting this branch.

PR #9093, opened 2026-09-07
fix(delegation): stop the wall-clock test expiring on a calendar date

JiRaska:
`SpendReservationServiceTest > occurredAt is a real clock reading, not a default` **started failing
at 2026-09-07T12:00Z and would have kept failing for good.** No code changed — a date passed.

## The arithmetic

```
fixed clock in the suite : 2026-08-08T12:00:00Z
grant validTo (= +30d)   : 2026-09-07T12:00:00Z
now (when CI ran)        : 2026-09-07T16:57Z   -> expired -> reservation refused
```

Every grant's window is derived from the suite's fixed clock. That one test **deliberately** runs
the service on `Clock.systemUTC()` — it checks `occurredAt` is a real reading rather than an
`Instant.EPOCH` default (#3874/#3883) — but still used the fixed-clock grant. Once wall-clock time
passed the fixed clock's `validTo`, the grant was expired on the only clock the service can see, so
the reservation was refused: `SpendReservationRefusedException`, not an assertion failure.

## Why this looked impossible

#8310 and #9015 both show **SUCCESS** on this build. They merged before noon. main went red between
them without either being at fault, which is exactly the shape that makes a dated fixture expensive
to diagnose — the obvious readings (flake, bad interaction between two PRs) are both wrong.

It blocks **every** delegation-service build from now on, including unrelated PRs. #8909 is stuck
behind precisely this.

## The fix

`grant()` gains an `at` parameter defaulting to the fixed `now`; the wall-clock test passes
`OffsetDateTime.now(wallClock)`. The window now follows whichever clock the service under test
reads.

Deliberately **not** widening `validTo` to a large number of days — that moves the bomb rather than
removing it. A fixture can no longer outlive its own window because it no longer has a fixed one.

## Falsified, not assumed

Removing the single line that builds the grant on the wall clock:

```
EXIT=1  tests="14" failures="1"
FAIL  occurredAt is a real clock reading, not a default()
      <- com.openbank.delegation.application.usecase.SpendReservationRefusedException
(the other 13 pass)
```

Exactly this test, exactly the exception CI reported, nothing else disturbed. Restored: 14 tests,
0 failures, ktlint clean.

## Security checklist

- [x] No secrets, tokens, passwords, or PII in code, config, tests, logs. Test fixture only.
- [x] No new `@SuppressWarnings` / `@Suppress` without justification. None added.
- [x] Auth / crypto / payment code changed? → **No production code at all.** The diff is one test
      file: `grant()` gains an `at` parameter and the wall-clock test passes its own clock.
      `SpendReservationService` and every spend-ceiling rule are untouched.
- [x] PII path changed? → No.

**Why this counts as money-path for the gate and still carries no money risk:** the file lives under
`openbank-delegation-service`, whose spend ceilings are a money-path control, so the path rule fires
correctly. What changed is the fixture's validity window, which existed only inside the test. The
13 other tests in the suite — including every ceiling-arithmetic case — are unchanged and green.


JiRaska:
Duplicate of #9086 — same defect, same fix, same file. Both add an `at` parameter to `grant()` defaulting to the fixed `now` and anchor the wall-clock test's grant to the real clock; the diffs differ only in the wording of the comment.

Keeping #9086 because it is the earlier of the two (18:19 vs 19:02) and has the fuller CI history. Nothing here is lost: the mechanism your KDoc states — that `at` is the clock the service under test will read, so the window must come from the same clock — is the same conclusion, and #9086 records it in the same place.

Two parallel sessions landed on this within 45 minutes; noting it here rather than silently closing, so the overlap is visible.

Not deleting the branch.

JiRaska:
Confirming this diagnosis independently, and adding the measurement that pins the expiry instant — it matters because the failure is not intermittent, it is a cliff every PR falls off from a fixed moment onward.

## The instant

The fixture pins `clock = Clock.fixed(Instant.parse("2026-08-08T12:00:00Z"))` and builds the grant with `validTo = now.plusDays(30)`, i.e. **2026-09-07T12:00:00Z**. `occurredAt is a real clock reading, not a default` is the one test that deliberately swaps in `Clock.systemUTC()`, so from that instant on it sees a wall clock past `validTo` and `reserve()` throws `SpendReservationRefusedException` at line 303 — before the recency assertion it was written for is ever reached.

## Measured, on the unmodified tree

| where | result |
|---|---|
| local, detached worktree of `origin/main` (84fab0e27), `--tests '*SpendReservationServiceTest*'` | **rc=1**, `14 tests completed, 1 failed` |
| local, `feat/libs-money-flow-features` (#8904) | rc=1, same single failure |
| CI, #8904 `build (openbank-delegation-service)`, ran 20:02:29Z | FAILURE |
| CI, #8927 same job | FAILURE |
| CI, **this PR**, ran 20:32:27Z -> 20:42:06Z | **success** |

The last two rows are the control: both ran after the expiry instant, on the same runner pool, minutes apart. The only difference is that this branch carries the fix. That rules out load, flakiness and the 3g heap change, and rules in the dated fixture.

It also settles a question I had to ask, since #8904 and #8927 both change `openbank-libs-domain` and delegation-service is a dependent: **this is not a libs-change breaking a consumer.** `openbank-delegation-service/src` contains no reference to `FeatureDefinition`, `MoneyFlowFeatures` or `FeatureEvent`, and the failure reproduces on `origin/main` with neither PR applied.

## Why the failure reads as something else

The job log's most eye-catching lines are `Unable to find an EntityManagerFactory for persistence unit '<default>'` (x3), `IllegalStateException: db down`, and a run of `delegation.expiration.sweep FAILED ... [Error Occurred After Shutdown]`. All are deliberate negative-path fixtures and shutdown noise; none is the failure. The only line that matters is `SpendReservationRefusedException at SpendReservationServiceTest.kt:303`, and the exception type — *refused*, not an assertion failure — is what points at the grant window rather than at the clock reading.

## Blast radius while this is open

Every PR whose diff reaches `openbank-delegation-service` fails `all-green` from 2026-09-07T12:00:00Z onward, and it cannot self-heal — waiting makes it worse, not better. #8904 and #8927 are blocked on exactly this right now, both with auto-merge armed, so merging this releases them without further intervention.

JiRaska:
Filed #9107 for the shape this instance exposed — the sweep found 19 test files that pin a fixed clock and also reach for the real one, and a note on why enforcing that mechanically is harder than it looks. This PR stays the fix for the live breakage; #9107 is about preventing the next one.

PR #9012, opened 2026-09-07
test(fleet): adopt trace contracts in 7 money-path services

JiRaska:
## Summary

Adopts the existing `TraceContract` facility in 7 more money-path services (account, balance, clearing, consent, fx, sanctions, sca), so the Test Intelligence `trace` column stops being empty for them. 5 services already used it; `src/main` is untouched in all 7.

## Type of change

- [ ] feat
- [ ] fix
- [ ] refactor
- [ ] perf
- [ ] docs
- [ ] chore (build / tooling)
- [ ] security
- [ ] breaking change

## Linked issues

Refs #265

## Changes

See the commit message — it carries the measurements, the falsification, and what is deliberately NOT claimed.

## Definition of Done

- [x] Hexagonal layering preserved — `src/main` untouched.
- [x] Unit tests added/updated.
- [ ] Integration tests added/updated — see commit.
- [ ] Contract tests updated — N/A
- [ ] `./gradlew detekt ktlintCheck koverVerify build` passes — see Reviewer notes
- [x] No new lint warnings — ktlintCheck + detekt run on every touched module.
- [ ] OpenAPI spec regenerated — N/A
- [ ] Flyway migration — N/A
- [ ] Avro schema — N/A
- [ ] Docs updated — N/A

## Security checklist

- [x] No secrets, tokens, passwords, or PII in code, config, tests, logs.
- [x] No new `@Suppress` without justification.
- [x] No new third-party dependency.
- [x] Auth / crypto / payment code changed? — no production code changed.
- [x] PII path changed? — no.
- [x] Cardholder data path changed? — no.

## Compliance impact

- [x] None

## Rollout / Rollback

- Deployment strategy: N/A — test/tooling only, nothing deployable.
- Rollback plan: revert the commit.
- Data migration reversible: N/A

## Reviewer notes

**The assertion that carries the weight is `requiresSameTrace`**: it proves the INSERT that persisted the aggregate happened inside the request's trace. A write that escapes the request context still writes its row and still answers 201 — invisible to every assertion these modules already had.

**Falsified rather than assumed.** Pointing one contract at a non-existent table made the test fail AND emit zero markers, confirming both that the assertions can fail and that `verifiedAs()` being last stops evidence escaping a run that later goes red.

[...]

PR #9098, opened 2026-09-07
build(deps): pin the Kotlin 2.4.20 toolchain in verification-metadata

JiRaska:
build(deps): pin the Kotlin 2.4.20 toolchain in verification-metadata

`verification-metadata` is red on every PR whose changed-module set triggers a
cold-cache resolve — #8838 and #9012 today, both with the identical list. A
cold-cache build pulls the Kotlin 2.4.20 compiler/daemon stack and
gradle/verification-metadata.xml does not pin it.

What made this easy to misread: the file DOES contain "kotlin-build-tools-api-2.4.20"
— six times. All six are release candidates:

  kotlin-build-tools-api-2.4.20-RC.jar
  kotlin-build-tools-api-2.4.20-RC2.jar
  kotlin-build-tools-api-2.4.20-RC3.jar

The final 2.4.20 is absent, so a substring grep answers "already pinned" about a
version that is not there. Ten components, twenty artifacts.

## How the checksums were obtained

Not from Gradle. `--write-verification-metadata` was attempted four times and
wrote nothing each time, for three different reasons that all look like success:
two Gradle JVMs sat at 0.0% CPU for 29 minutes with the cache byte-count frozen;
a leftover daemon held the journal lock (`Timeout waiting to lock journal cache`);
and `testClasses UP-TO-DATE` means nothing is resolved, so there is nothing to
record — the same no-op as a warm dependency cache, one level up.

Instead each artifact was downloaded from Maven Central and hashed locally, which
is what Gradle itself does, then cross-checked a second way: the SHA-1 of the same
downloaded bytes matches the `.sha1` sidecar Maven Central publishes, for all 20.
(There are no `.sha256` sidecars — those URLs 404, which a naive check reports as
20 mismatches rather than as an absent file.)

The diff is 80 insertions and 0 deletions; the XML parses; no component is
duplicated; every one of the 20 artifacts the gate named is now present with the
hash computed above.

## Follow-up worth opening separately

`openbank-libs/gradle/libs.versions.toml` declares `kotlin = "2.3.20"`, so nothing
here asks for 2.4.20 by name — and the file's own history shows RC, RC2 and RC3
each being added as they appeared. That pattern is what a floating resolution
looks like, and pinning checksums does not address it. Naming it here rather than
silently normalising it: this PR unblocks the gate, it does not answer why the
toolchain version moves on its own.

Refs #8838, #9012

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

## Security checklist

- [x] No secrets, tokens, passwords, or PII in code, config, tests, logs. Checksums only.
- [x] No new `@SuppressWarnings`, `as Any`, `@Suppress("...")` without justification. None.
- [x] No new third-party dependency. **None added** — this pins artifacts the build already
      resolves today. Every hash was computed from the bytes Maven Central serves and
      independently corroborated against the published SHA-1 sidecar.
- [x] Auth / crypto / payment code changed? → tag `security-review-required`. **No.**
- [x] PII path changed? → tag `gdpr-review-required`. **No.**
- [x] Cardholder data path changed? → tag `pci-review-required`. **No.**


JiRaska:
Auto-merge is armed here but **cannot fire until #9086 lands**. `build (openbank-delegation-service)` fails on this PR for a reason that has nothing to do with it: `SpendReservationServiceTest` expired against the wall clock at 2026-09-07T12:00Z and is red on `main`, so `all-green` — which is a required context — stays red on anything whose diff reaches that module.

Not fixing it from here on purpose: cherry-picking an unrelated delegation test fix into a dependency-pinning PR would need the same human review #9086 already needs, and would make both harder to read. #9086 is green (31 checks, 0 failing) and one file.

Order to land: #9086 → this → #8838 and #9012 go green with no change of their own.

JiRaska:
Correcting my note above: this PR is gated by **two** main-side fixes, not one. `build (openbank-libs-domain)` has now failed here as well, which is the `TopicProducers` prefix contradiction that #9085 fixes — #8920 added the `openbank-kyb-service` row and, in the same PR, the test forbidding that prefix.

So the order is **#9086 and #9085 → this → #8838 / #9012**. Both blockers are one-file changes to main and neither has anything to do with pinning Kotlin checksums; the armed auto-merge here fires once `all-green` can go green, which needs both.

Nothing to change on this branch — flagging it so the dependency is not read as a single one.

Issue #9100, opened 2026-09-07
credential-deadline-ratchet fails closed on a base commit the shallow checkout lacks, and renders it as a policy finding


JiRaska:
`credential-deadline-ratchet` reports FAIL when it has not evaluated anything. The gate does not
produce a finding — `credential-inventory.py` raises, and the runner renders the non-zero exit as a
policy failure.

From #9098's `gates (gitops-api)` job:

```
Traceback (most recent call last):
  File ".github/scripts/credential-inventory.py", line 199, in <module>
    sys.exit(main())
  File ".github/scripts/credential-inventory.py", line 187, in main
    return enforce_new(args.enforce_new, today)
  File ".github/scripts/credential-inventory.py", line 121, in enforce_new
    out = subprocess.run(["git", "diff", "--name-only", "--diff-filter=A", f"{base}...HEAD"], ...
subprocess.CalledProcessError: Command '['git', 'diff', '--name-only', '--diff-filter=A',
  'f145b3297a15311504ad8abc0f30a48375d17090...HEAD']' returned non-zero exit status 128
```

Exit 128 from `git diff A...B` here means the runner's object store does not contain
`f145b3297` — `PR_DIFF_BASE` resolves to a `main` tip that the shallow PR checkout never fetched.
It is not a missing rotation deadline.

## Why this matters more than one red square

The same gate, on the same repository, is green for the same code when the base commit happens to be
present. Measured just now across open PRs: #9085, #9063, #9071, #9055 → `gates (gitops-api)`
SUCCESS; #9098 and #9077 → FAILURE. And run locally on #9098's exact head with `--enforce-new
origin/main`, the gate answers:

```
credential-deadline-ratchet: 0 finding(s)
```

So the verdict depends on the runner's fetch depth rather than on the diff, which makes both outcomes
uninformative: a green run does not establish that the credential inventory was checked, and a red
one does not establish that anything is wrong. That is the failure mode the repo's own gate
conventions are written against — a check that cannot distinguish "clean" from "could not look".

## Suggested direction (not a patch)

Two independent halves:

1. **Fetch what the gate needs.** Whatever resolves `PR_DIFF_BASE` should guarantee the object is in
   the checkout (deepen, or `git fetch origin <base-sha>`) before any gate diffs against it. Other
   gates that diff against the base are exposed to the identical failure.
2. **Fail loudly as infrastructure, not as a finding.** `enforce_new` should catch
[...]

PR #9101, opened 2026-09-07
fix(gates): credential-deadline-ratchet must not report "could not run" as a finding

JiRaska:
## What happened

`credential-inventory.py --enforce-new` ran:

```python
subprocess.run(["git", "diff", "--name-only", "--diff-filter=A", f"{base}...HEAD"], check=True)
```

The gates shard checks out at `fetch-depth: 1` (ci.yml says so in its own comment). A PR whose base commit has since scrolled out of the shallow window does not have that commit. git exits 128, `check=True` raised, and CI reported:

```
FAIL [gitops] new static ExternalSecret must declare a rotation deadline (ADR-0279 #18, enforced)
subprocess.CalledProcessError: Command '['git', 'diff', ... '2a5b004d...HEAD']' returned non-zero exit status 128
```

It had found no such ExternalSecret. **It had not looked.**

## How it was caught

#9095 is a two-file change — one Kotlin service's `application.yaml` plus a new unit test. It touches no manifest, and no ExternalSecret appears anywhere in its diff. Sibling PRs opened minutes later (#9091, #9092, #9094) passed the same gate, because their base was still inside the shallow window.

So the gate's verdict tracked **how long a branch had been open**, not what it contained. That is not a policy check.

## Why it matters beyond this one gate

A check that cannot distinguish *"nothing to report"* from *"could not run"* is not a check. This one failed closed — the right direction — but under the wrong sentence: it accused the author of a credential violation and sent them looking for a manifest that never existed. The direction is not the part that was broken; the claim was.

## The fix

- **`reach_base`** — fetch the base by name first (cheapest, and works even when it is not on main), then deepen in place. Same idiom and same reasoning as `check-flyway-version-commit-order.py`: 60 other gates do not need full history and should not pay for it.
- **Honest UNKNOWN** — when the base is genuinely unreachable, or the diff itself fails, say `UNKNOWN` in those words, state plainly that it is *not a finding against this pull request*, and name the likely cause. Still exit non-zero, because an unrunnable gate must not go green — but never under the policy's sentence.

## Falsification

The self-test now covers the gate's own blind spot, since that is precisely what broke.

| Sabotage | Result |
|---|---|
| unreachable-base path `return 0` | `self-test FAIL: unreachable base returned success — a gate that did not run must never report clean` |
| reword the message as a rotation-deadline finding | `self-test FAIL: unreachable base reported as a credential violation` |

Real-repo behaviour, both directions:

```
$ credential-inventory.py --enforce-new origin/main~3
credential-deadline-ratchet: 0 finding(s)          # exit 0

$ credential-inventory.py --enforce-new 000…000
::error::… could not reach the diff base …, so it scanned NOTHING. This is an
UNKNOWN result, not a clean one, and not a finding against this pull request.  # exit 1
```

## Unfixed, and worth a separate look

This gate was found only because its crash happened to be loud. **Any other gate calling `git diff <base>...HEAD` under `check=True` has the same latent fault**, and one that catches the exception and returns 0 would fail *open* — silently reporting clean. `check-adversarial-contract-tests.py`, `check-flyway-version-commit-order.py`, `check-loki-rules.py` and `check-security-checklist.py` all use the `...HEAD` idiom; flyway already handles shallowness explicitly, the others I have not audited. Happy to sweep them if useful — kept out of this PR to keep the falsified change small.

### DIFF
--- a/.github/scripts/credential-inventory.py
+++ b/.github/scripts/credential-inventory.py
@@ -270,80 +270,124 @@ def self_test() -> int:
     # that fixes the real CI symptom unproven, so it gets its own case: a base that is genuinely
     # absent from a shallow clone and genuinely recoverable from its origin. `file://` keeps it
     # offline and fast; the deepening path is the same one CI takes.
-    with tempfile.TemporaryDirectory() as tmp:
-        origin, clone = Path(tmp) / "origin", Path(tmp) / "clone"
-        git_o = ["git", "-C", str(origin)]
-        # A fixture must not depend on the ambient git configuration. CI built no commits at
-        # all here — `rev-parse HEAD~2` came back `fatal: ambiguous argument` — while the same
-        # code committed fine on a developer machine, which is the signature of a system or
-        # global gitconfig doing something the fixture never asked for (a commit template, a
-        # hooks path, signing, an identity policy). Rather than chase which one, the whole
-        # ambient layer is switched off: GIT_CONFIG_* to /dev/null, HOME inside the fixture,
-        # and the identity supplied through the environment instead of inherited.
-        env = {
-            **os.environ,
-            "GIT_CONFIG_GLOBAL": os.devnull,
-            "GIT_CONFIG_SYSTEM": os.devnull,
-            "GIT_CONFIG_NOSYSTEM": "1",
-            "HOME": str(Path(tmp) / "home"),
-            "GIT_AUTHOR_NAME": "fixture", "GIT_AUTHOR_EMAIL": "fixture@example.invalid",
-            "GIT_COMMITTER_NAME": "fixture", "GIT_COMMITTER_EMAIL": "fixture@example.invalid",
-        }
-        (Path(tmp) / "home").mkdir(exist_ok=True)
-        steps: list[subprocess.CompletedProcess[str]] = []
+    # Strip the inherited git environment for the WHOLE fixture, process-wide and restored
+    # after. Passing a clean env to the setup commands is not enough: the recovery case calls
+    # enforce_new, whose own subprocesses inherit os.environ, and GIT_DIR/GIT_INDEX_FILE
+    # override the working directory — so `os.chdir(clone)` is silently ignored and git keeps
+    # answering about the main checkout. That is what CI was doing, and it is invisible from
+    # inside the fixture: every command succeeds, against the wrong repository.
+    inherited_git = {k: v for k, v in os.environ.items() if k.startswith("GIT_")}
+    for k in inherited_git:
+        del os.environ[k]
+    try:
+        with tempfile.TemporaryDirectory() as tmp:
+            origin, clone = Path(tmp) / "origin", Path(tmp) / "clone"
+            git_o = ["git", "-C", str(origin)]
+            # A fixture must not depend on the ambient git configuration. CI built no commits at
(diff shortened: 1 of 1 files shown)


JiRaska:
### Update: the first draft of this fix was wrong, and its own self-test could not see it

CI ran the gate and it **passed** — `self-test: clean`, `0 finding(s)` — then failed the shard on the time budget:

```
[run-gates] this gate took 12.7s against a budget of 5s
```

Two separate defects behind that, both now fixed and falsified.

**1. The self-test ran against the real checkout.** It called `enforce_new`, which reached the network and tried `--unshallow` — the 12.7s. Worse than slow: a self-test that deepens the repository it runs in has altered the thing it was meant to observe. It now runs in a throwaway repo. 0.7s.

**2. `reach_base` did not actually work.** This is the one worth reading.

`base...HEAD` is a **three-dot** diff, so it needs a **merge base**, not just the base commit. Fetching the sha by name (`--depth=1`) makes `git cat-file -e` succeed while the histories stay disconnected — so the diff still exits 128. The first draft checked exactly that, object presence, declared the base reached, and **failed in precisely the place it was written to fix**.

Its self-test passed throughout. Both the unrecoverable base and the failed diff land on an honest UNKNOWN, so an outcome-only contract cannot tell a working repair from an almost-correct one. The success criterion is now the gate's *actual query*, which is the one criterion an almost-correct repair cannot satisfy.

That gap was found by a new self-test case, not by review:

| Sabotage | First draft | Now |
|---|---|---|
| unreachable base `return 0` | caught | caught |
| reword as a deadline finding | caught | caught |
| `reach_base` always returns `True` | **passed clean** | `a recoverable base was not recovered` |
| drop the deepening attempts | **passed clean** | `a recoverable base was not recovered` |

The new case builds a shallow clone of a local `file://` origin, asserts the base is genuinely absent first (otherwise the fixture proves nothing), and requires it to be recovered. Offline, under a second, same deepening path CI takes.

### One accepted cost

Normal run 0.1s, self-test 0.7s. The unreachable path costs ~2.7s against a real origin here and may cost more in a shallow CI checkout, since that is where deepening does real work — it could brush the 5s budget. Left as is deliberately: a slow honest answer beats a fast false accusation, and it only runs on the path that was previously broken. If it does start tripping the budget, raise the number rather than skip the recovery.

JiRaska:
Confirming this works, from its own failing run: the job log now carries your new message —

```
the diff base …, so it scanned NOTHING. This is an …
```

— alongside the underlying `git diff … '2a5b004d...HEAD'` returning `exit status 128`. So the script half of #9100 does what it says; the run is still red because the *cause* is upstream in the workflow, and this PR deliberately does not touch that.

#9102 is the other half: `resolve PR diff base` currently falls back to `EVENT_BASE_SHA` without re-checking it, and hands every gate a base they cannot diff against.

One measurement worth carrying over, because it cost me a wrong first patch: the property to assert is **`git merge-base <base> HEAD` resolving**, not the base object being present. Measured in a depth-1 clone against this remote — with the base fetched by sha and `git cat-file -e` answering yes, `git diff <base>...HEAD` still exits 128 while `<base>..HEAD` works, because three-dot needs the merge-base and the shared history is cut. Deepening is what fixes it.

The two PRs are complementary: this one makes a future recurrence legible, #9102 removes today's cause.

PR #9108, opened 2026-09-07
feat(transaction): serve merchant logos from this bank's own origin

JiRaska:
## Summary

Merchant logos are now stored and served by this bank, not linked from somebody else's CDN, and the
enrichment response carries an origin-relative, content-hashed URL for them. This is the first slice
of the merchant-enrichment programme (logos → honest geo → the card path → automated ingest).

## Type of change

- [ ] fix (bug fix)
- [ ] perf (performance)
- [ ] docs
- [ ] chore (build / tooling)
- [ ] breaking change

## Linked issues

Refs #8573

## Changes

- `V18__merchant_logo.sql` — `merchant_logo` (two pre-rendered square PNGs, content type, SHA-256,
  provenance: source URL, licence, attribution, uploader) plus `merchant_catalog.logo_etag`, the
  denormalised presence marker the per-page enrichment read consults so it never needs a join.
- `merchant_catalog.logo_url` changes meaning rather than getting populated: it is now the
  **provenance** record (where a logo was obtained), operator-facing, never sent to a customer.
- `LogoImages` — decode, dimension-check, fit into a transparent square, re-encode as PNG.
- `GET|PUT|DELETE /api/v1/merchants/{descriptorKey}/logo`, operator-write / reader-read, reusing the
  existing `merchant.list` / `merchant.update` / `merchant.delete` authz actions.
- `MerchantResponse.logoUrl` is now derived: `/api/v1/merchants/{key}/logo?size=64&v=<hash prefix>`,
  or absent when no logo has been ingested.
- OpenAPI 1.14.0 → 1.15.0 (additive).

## Definition of Done

- [ ] Integration tests added/updated (if service boundary involved). Deliberately not: the routes
      are exercised at the resource layer, and the reactive repository's session behaviour is the
      part a real-DB IT would add value for — that arrives with the geo/terminal slice, which
      changes the same read path.
- [ ] Contract tests updated (if public API changed). No consumer pact covers `merchant`; the edge
      forwards it byte-for-byte and gains the route in a follow-up PR.
- [x] No new lint warnings.
- [ ] Avro schema versioned backward-compatibly (if event changed). No event changed.
- [x] Docs updated (README, ADR if architectural). `docs/threat-models/openbank-transaction-service.md`
      §4e, plus the §4c spoofing row, which claimed `logoUrl` was expected to be bank-controlled and
      is now structurally unable to be anything else.

## Security checklist

      this PR). Two `@Suppress("ThrowsCount")` in `LogoImages`: each throw is a distinct,
      separately actionable rejection reason, and collapsing them would tell an operator only that
      the upload was refused. One `@Suppress("LongParameterList")` on `putLogo`, whose parameters
      are the path key, three provenance fields, the security context and the body.
- [x] No new third-party dependency, OR: dependency review attached (license, CVE, source). None —
      image decoding is `javax.imageio` from the JDK.
- [x] Auth / crypto / payment code changed? → tag `security-review-required`. Considered and the
      answer is **no**: no authentication, cryptographic or payment-execution code is touched. The
      new routes reuse the catalogue's existing roles and OPA actions, and the SHA-256 here is a
      content identifier for caching, not a security primitive. No tag.
- [x] PII path changed? → tag `gdpr-review-required`. The change **removes** an exposure that had
      not shipped yet rather than adding one; see below.
- [x] Cardholder data path changed? → tag `pci-review-required`. Considered and the answer is
      **no**: nothing added is keyed by card, customer or transaction. `merchant_logo` hangs off the
      acquirer descriptor, exactly as the catalogue it belongs to does. No tag.

## Compliance impact

- [ ] PCI DSS
- [ ] DORA
- [x] GDPR
- [ ] PSD2 / SCA / RTS
- [ ] AML / 5AMLD
- [ ] CNB reporting
- [ ] None

The design decision this PR exists to make: a third-party logo URL in the statement response would
have meant every render telling that host the customer's IP address and which merchant they paid, at
the moment they opened their transaction list — a spending profile reconstructable from an access
log, outside any consent or processor agreement. Serving the bytes from this origin removes the
third party from the request path entirely. Nothing customer-derived is stored: `merchant_logo` is
keyed by acquirer descriptor, like the catalogue it hangs off.

Logo bytes are re-encoded rather than stored as uploaded. That is a security control:

- SVG and HTML are refused (a logo a banking app renders must not be able to carry script);
- bytes appended after a valid image do not survive the re-encode (polyglot files);
- embedded metadata (EXIF, including GPS) is dropped, because pixels are all that is kept;
- a header declaring implausible dimensions is refused **before** any pixel buffer is allocated — a
  40 kB PNG can declare 30000×30000, which is 3.6 GB of `int[]` and an OOM kill of the pod.

`LogoImagesTest` asserts each of those refusals rather than the happy path, which would pass just as
well with every guard deleted.

## Rollout / Rollback

- Rollback plan: revert the service. The route disappears and `logoUrl` goes back to null; nothing
  else reads `merchant_logo` or `logo_etag`, and no other service's response shape changes.
- Data migration reversible: yes — `DROP TABLE merchant_logo;` and
  `ALTER TABLE merchant_catalog DROP COLUMN logo_etag;`. Both are additive, so the previous release
  runs unchanged against the new schema.

## Reviewer notes

- **The catalogue marker is maintained in the same transaction as the bytes**, through the same
  Hibernate session rather than a second repository call, because the hot path reads only the marker
  — a drift would advertise a logo that 404s or hide one that exists.
- The customer-facing URL is deliberately **origin-relative**, so whichever host served the statement
  serves the logo. customer-edge forwards `merchant` byte-for-byte, so exposing the same path there
  is what makes it resolve for the mobile app; that is a follow-up PR, together with the spec change
  on the edge side.
- Still to come in this programme, in order: geo honesty (today's seed pins a whole chain at one
  Prague coordinate, and per-store location properly comes from the POS terminal / ATM identifier,
  which nothing in the fleet currently carries); the card path (`MerchantDescriptor` into
  `openbank-libs-domain`, MCC fallback); SSRF-hardened automated ingest with per-source licence
  recording; the admin-ui upload screen.

### DIFF
--- a/docs/threat-models/openbank-transaction-service.md
+++ b/docs/threat-models/openbank-transaction-service.md
@@ -153,7 +153,7 @@ line and needs its own review.
 | **I**nfo disclosure | Enrichment leaks a *cardholder's* whereabouts rather than a shop's | The catalogue has no customer-, card- or transaction-scoped column; a row is per merchant descriptor and identical for every customer who shopped there. Nothing customer-derived is written back |
 | **T**ampering | A wrong or planted catalogue row attributes a payment to the wrong business — a lever for social engineering ("your payment to X") or for hiding one | Rows arrive only by migration, never from a request. Lookup is an **exact** match on the normalised key: no fuzzy or prefix matching, so a near-name cannot inherit another merchant's identity. `description` is passed through unmodified, so the raw acquirer text remains available and authoritative |
 | **R**epudiation | A dispute is raised against a prettified name that does not appear on the acquirer record | Enrichment is display-only and additive. Disputes and SPAYD consume `description`, which this change does not touch; `source: ENRICHED` labels anything the bank resolved |
-| **S**poofing | `logoUrl` points at attacker-controlled content rendered inside the bank app | URLs are catalogue-controlled and expected to be on a bank-controlled CDN; there is no request path that can set one |
+| **S**poofing | `logoUrl` points at attacker-controlled content rendered inside the bank app | Since §4e the field is DERIVED and origin-relative — it can only ever address this service's own logo route, and the catalogue column an operator writes is provenance that customers never receive. The mitigation is now structural rather than an expectation about what operators type |
 | **D**oS | Enrichment adds a per-row query to every statement page | One query per page: descriptors are normalised, de-duplicated and fetched together, so cost is bounded by distinct merchants on the page, not row count |
 
 **DFD update:** none. Same caller, same endpoint, same authorisation; one additional read of a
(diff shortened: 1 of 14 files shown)


JiRaska:
Both `gates (registry-kotlin-data)` findings here are real — I checked for the `exit status 128`
base-diff crash that is producing phantom findings on other PRs tonight (#9100) and this is not that.
Quoting them so they don't need digging out of the log:

```
##[error]new `Instant = Instant.EPOCH` default (var updatedAt: Instant = Instant.EPOCH)
  — make the caller supply the time, inject a Clock, or use Instant.now() at the owning call site (#3882, #4208)

##[error]New Flyway migration without a rollback note (rules.yaml: db_change requires one).
  Add a `-- Rollback:` comment saying how to undo this migration
```

The rollback note appears to be written already — the PR body says *"yes — `DROP TABLE
merchant_logo;`"* — but the gate reads the **migration file**, not the body, so it needs to be a
`-- Rollback:` comment in the `.sql`.

On the epoch default, the root `CLAUDE.md` has the history: `Instant.EPOCH` as a data-class default
is the one value that passes every `isNotNull()` assertion while claiming 1970-01-01, which is how
#3882 shipped 23 call sites of it. For a `updatedAt` on a persisted entity the fix is usually the
Clock the service already injects rather than `Instant.now()` inline.

Not touching the branch — it is three minutes old and clearly being actively worked.

PR #9158, opened 2026-09-08
fix(admin-ui): make identity decision dialog focus-safe

### LINKED ISSUE #9157: admin-ui: make Identity Cases decision dialog focus-safe
Identity Cases still uses a hand-rolled alert dialog and manual focus trap for high-stakes four-eyes vote and reopen actions. Replace it with the established Radix alert-dialog pattern, preserve endpoint/RBAC/maker-checker behavior, prevent dismissal while a decision is pending, restore cancel focus to the exact trigger, and move successful decisions to the stable workspace when the source row disappears. Add behavioral regression coverage.

JiRaska:
## Summary

- replace the hand-rolled Identity Cases decision modal with the established Radix alert-dialog pattern
- keep destructive vote/reopen actions open and non-dismissible while the request is pending
- restore Cancel/Escape focus to the exact trigger and successful decisions to the stable workspace when the source row disappears
- preserve the existing four-eyes endpoints, payloads, RBAC, maker-checker rules, retry behavior, and single-flight guard

## Verification

- focused Vitest: 2 files / 4 tests passed
- canonical Admin UI type-check passed
- changed-file ESLint: 0 errors (one unchanged inherited hook warning)
- Chromium identity decision recovery E2E: 1/1 passed on an isolated port
- independent exact-diff review: no blocker
- diff check clean

Closes #9157


JiRaska:
Superseded by #9345, which carries the same identity decision focus-safety intent forward with the shared Radix alert-dialog primitive, exact-trigger restoration, busy-state dismissal locking, and Chromium regression coverage.

Issue #9130, opened 2026-09-08
auto-deploy has failed on every run since 2026-09-07 21:27 — three services blocked on pact regressions


JiRaska:
Noticed while confirming that #9095 and #9104 had reached sandbox (they did — `sandbox-a43ace11` carries both).

`auto-deploy.yml` outcomes, newest first:

```
running  2026-09-08T04:20
failure  2026-09-08T03:33
failure  2026-09-08T02:54
failure  2026-09-08T01:03
failure  2026-09-07T21:27
success  2026-09-07T21:00
```

Four consecutive failures, all the same cause — `can-i-deploy` refusing three services:

```
can-i-deploy: openbank-sdd-service   NOT deployable [REGRESSION] — a published pact FAILED verification
can-i-deploy: openbank-vop-service   NOT deployable [REGRESSION]
can-i-deploy: openbank-swift-service NOT deployable [REGRESSION]
```

The gate says plainly what happens next, and it is the part that matters:

> this will not clear on its own; a reconcile tick will re-offer it forever (ADR-0092, #2549)

The failing pacts are sdd→transaction, vop→account, vop→party and swift→transaction, all against `main` at `2a5b004d` — nothing to do with the campaign changes that merged just before the run of failures began.

### Why this needs an issue rather than a retry

Deploys still partially land (the sandbox tag advanced to `a43ace11` at 02:54), so the estate does not look frozen — it looks like it is working while three services are quietly pinned to whatever they were. That is the same shape as #9004: one job covering the whole fleet, where a sibling's failure is indistinguishable from a slow day for everyone else. A red run every 40 minutes with nobody assigned is a broken pipeline that has started to read as background noise.

There is also a second, unrelated error in the same log worth a look while someone is in there:

```
Error making request to https://pact.open-bank.tech/environments status=400
{"errors":{"name":["name 'sandbox' is already used by an existing environment."]}}
```

That one is non-fatal today, but it means the step creates the environment unconditionally rather than reconciling it.

### Scope

- [ ] Fix or re-verify the three failing pacts (they will not clear on their own)
- [ ] Decide whether an all-red auto-deploy should alert — four silent failures over seven hours is the finding underneath the finding
- [ ] Make the pact environment creation idempotent

JiRaska:
## Traced the deadlock to its root — it is not the three services the error names

The `can-i-deploy` errors name sdd, vop and swift. Following the chain, all four blocked services (transaction-service included) terminate at one place:

```
[...]

PR #9166, opened 2026-09-08
fix(campaign): set INCENTIVE_SERVICE_URL so the incentive client resolves in-cluster

### LINKED ISSUE #9096: campaign-service: INCENTIVE_SERVICE_URL is not set in the Deployment
Found while fixing the REST client config keys (#9095).

`openbank-campaign-service/src/main/resources/application.yaml` resolves the incentive client as:

```yaml
quarkus:
  rest-client:
    incentive-service:
      url: ${INCENTIVE_SERVICE_URL:http://localhost:8156}
```

`openbank-infra/gitops/components/campaign/campaign-service.yaml` **never sets `INCENTIVE_SERVICE_URL`**. In-cluster the client therefore falls back to `http://localhost:8156` and cannot reach incentive-service.

Pre-existing; #9095 does not regress it (the `${…}` expression was already there, just on a key Quarkus never read). Splitting it out because the fix is a different change class: adding the env var makes `gen-network-policies.py` derive a new campaign -> incentive edge, so it needs a policy regen and a threat-model look, not a one-line yaml edit.

### Why this was invisible

Same shape as #9095: the call site fails closed, so a broken incentive client reads as "no incentive offer for this party" rather than as an error. Worth checking what the observable symptom actually was before assuming it never fired.

### Scope

- [ ] Set `INCENTIVE_SERVICE_URL` in the campaign Deployment
- [ ] Regenerate network policies (`gen-network-policies.py`) and confirm the campaign -> incentive edge appears in the live ingress allow-list
- [ ] Update the incentive-service threat model if the inbound caller set changes
- [ ] Extend `RestClientUrlConfiguredTest` (or a sibling) so a `${VAR:...}` in application.yaml whose `VAR` no Deployment sets is itself a failure — the current guard only proves the config key exists, not that the value survives deployment

The last item is the one that generalises. This class of bug has now appeared twice.

JiRaska:
Closes #9096.

## The defect

campaign-service resolves its incentive REST client as:

```yaml
quarkus:
  rest-client:
    incentive-service:
      url: ${INCENTIVE_SERVICE_URL:http://localhost:8156}
```

**Nothing has ever set `INCENTIVE_SERVICE_URL`.** In-cluster the client therefore pointed at its own pod's loopback, where nothing listens, and every call failed.

## Why nobody noticed

Same reason as #9095: the call site **fails closed**, so a broken incentive client reads as *"no offer for this party"* rather than as an error.

The two were siblings, one layer apart. #9095 moved the URL onto a config key Quarkus actually reads; this gives that key a value that exists outside a developer laptop. Fixing only the first would have left the client exactly as unreachable — which is worth stating plainly, because #9095 looked like the whole fix at the time.

## NetworkPolicy

`gen-network-policies.py` derives ingress edges from Deployment env URLs, so adding the variable also produces the rule that lets the call through. **Regenerated, never hand-edited.** The generator emitted exactly one new edge and touched nothing else:

```diff
       - from:
+          - namespaceSelector:
+              matchLabels:
+                kubernetes.io/metadata.name: campaign
```

## Falsification

Not just "gates are green". Reverting `incentive/network-policies.yaml` to main while keeping the env var — i.e. exactly the mistake of adding the URL and forgetting the regen — fails:

```
::error::NetworkPolicy drift detected — run gen-network-policies.py and commit the result
  FAIL  gen-network-policies-drift-gate  26.9s  [gitops]
```

With the regen committed, the whole `gitops` gate group passes (exit 0).

## Not in this PR

incentive-service has **no document under `docs/threat-models/`**, so there is no inbound-caller list to keep honest — 45 services have one and this is not among them. That deserves its own issue rather than a drive-by file here.

I have not verified the incentive call now succeeds end-to-end in sandbox: that needs the fix deployed, and auto-deploy is currently deadlocked (#9130). The claim here is scoped to what is checkable — the variable is set, the policy allows the edge, and omitting either is caught.


JiRaska:
## Three checks are red. One is mine, and it needs your call — I am not going to make it by editing the baseline.

### Mine: ASVS V9.1 — a new plaintext inter-service edge

```
::error::V9.1 campaign-service.yaml:154: plaintext http:// env value http://incentive-service.incentive.svc:8156
asvs-l3: 1 NEW finding(s) against the mechanical subset — correct the code, not the gate
```

The ratchet is doing exactly its job. The awkward part is that this is not really a new *call* — campaign-service has always called incentive-service; the edge existed in intent and simply never worked because the variable was unset. Making it work is what makes it visible to the gate.

Facts that bound the options:

- **https is not available.** incentive-service has no `internal-tls` secret and its Service exposes only `8156` http. I cannot point at a TLS endpoint that does not exist.
- **239 identical edges are already baselined**, including campaign's own siblings — `consent-service:8106` (line 216) and `lending-service:8126` (line 218). Internal traffic in this fleet is plaintext essentially everywhere.
- **The baseline file has no justification field.** Adding a line leaves no trace of why.

So the three real options:

1. **Baseline it** — consistent with the 239 siblings, but it is one more line on a ratchet whose entire purpose is to stop that number growing, and I would be adding it to unblock my own PR. That is the shape of defeating a control, even when the resulting config is no worse than its neighbours.
2. **Leave the client broken** — absurd, but it is what "no new plaintext edge, no exceptions" means here.
3. **Give incentive-service internal TLS** — the only option that actually improves anything, and far outside this PR.

My recommendation is (1) *with* (3) filed as a follow-up, because the alternative is a feature that stays broken to keep a number flat. But adding a line to a security ratchet to unblock my own change is your decision, not mine — so the PR stays red until you say which.

### Not mine: `gates (lint-supplychain-security)`

Fails identically on **main** (`74f0ca82`, and the two runs before it):

```
self-test verdict wrong — gate is unfalsified: a gate that wrongly accepts its known-positive
self-test verdict wrong — gate is unfalsified: a gate whose --self-test harness fails
gate failed: an enforced gate that fails
```

These come from harness *test fixtures* in group `t` — gates designed to fail — being picked up by the real run. main is red on this and has been for at least three runs; this PR inherits it. Worth its own issue if it is not already someone's.

### Probably not mine: `prod-readiness collectors agree`

`gates (gitops-api)` passes on main, so I did not dismiss this one as inherited — but it passes locally on this branch:

```
SUBJECTS=48
readiness collectors agree: identical scores, evidence and gates for every service
```

Its CI half compares a Node collector against the Python one, so I cannot fully reproduce the CI condition here. Recording it as unexplained rather than claiming it is fine.

JiRaska:
Superseded by #9139 — identical INCENTIVE_SERVICE_URL env var + NetworkPolicy fix for #9096, but #9139 also updates the ASVS L3 baseline for the newly-introduced plaintext http:// entry (which is why this PR's Validate manifests check was failing). Closing as duplicate, not merging.

PR #9187, opened 2026-09-08
fix(admin-ui): protect document template drafts

### LINKED ISSUE #9186: Modernize document template editor dialog
## Problem

The Admin UI document-template editor uses a hand-rolled modal. It does not provide a complete dialog focus contract and can discard long HTML or sample-data drafts without an explicit confirmation.

## Acceptance criteria

- Use the existing Radix dialog primitive with an accessible name and description.
- Focus the first relevant editor field and restore focus to the invoking control.
- Require explicit confirmation before discarding a dirty draft.
- Prevent dismissal while a save is in flight and preserve the existing API, payload, RBAC and preview behavior.
- Add behavior-level regression coverage for clean close, dirty discard and busy save.

## Scope

Admin UI only. No document-service API, authorization or persistence changes.

JiRaska:
## Summary

- replace the hand-rolled document-template editor overlay with the existing Radix dialog primitive
- preserve long HTML and sample-data drafts behind an explicit discard confirmation
- restore focus to the invoking control and block dismissal while a save is running
- retain the existing document-service endpoints, request payloads, preview behavior and RBAC

Closes #9186

## Verification

- `npm test -- src/test/document-template-editor-dialog.test.tsx src/test/document-template-lifecycle-dialog.test.tsx src/test/document-templates-accessibility.guard.test.ts` — 9/9 passed
- `npm run type-check` — passed
- changed-file ESLint — 0 errors; one pre-existing `useEffect(load)` warning
- `git diff --check` — passed
- independent exact-blob review — no blocker

## Risk

Admin UI presentation/state only. No API, authorization, persistence, document content, or lifecycle payload changes.


JiRaska:
## My earlier merge on this branch deleted the feature the PR exists for — fixed

Correcting my own mistake, and the mechanism is worth recording.

I resolved `document-templates/page.tsx` with `git checkout origin/main -- <file>`. That is
**whole-file, not per-hunk**: it took main's version entire and dropped 112 lines of this
branch's work, including the unsaved-draft guard. The branch was left titled "protect document
template drafts" while protecting nothing, and the diff looked like an ordinary conflict
resolution.

Both sides had independently rebuilt this editor on Radix Dialog — main added the `Tabs`
primitive and `modalReturnFocusRef`, this branch added the dirty-check and the discard
confirmation. Complementary, not competing. main's base is kept and the guard is restored on
top of it.

### What the test caught that I would have missed

Four ways out of that editor: Radix `onOpenChange` (overlay click, Escape), the header X, and the
footer "Close" button. The last one still called `setModalOpen(false)` directly, so one exit in
four walked straight past the guard — the kind of gap that reads as working every time you happen
to close it the other three ways.

```
document-template-editor-dialog.test.tsx    1 failed | 1 passed   ->   2 passed
full admin-ui suite                         519 files / 2810 passed, 1 skipped
```

### The procedural lesson

Last time I ran only the one guard test I had touched, saw it green, and called the merge done.
The suite was already red at that point — on a test file I had not opened. **Run the whole suite
after a merge**; the file you did not look at is exactly where a whole-file checkout lands.

Also: I twice deleted my own worktree in this session by running `git worktree remove` on the line
after a `git push` that had been rejected. Verify the push landed on the remote before cleaning up.

Issue #7654, opened 2026-08-31
feat(admin-ui): complete ADR-0208 operator design-system migration


JiRaska:
## Outcome

Make the bilingual banking operator console consistent, accessible, fast to scan, and educational without changing its security, RBAC, maker-checker, BFF, or money-path contracts. This is the delivery tracker for ADR-0208, not a cosmetic rewrite.

## Current baseline (main, 2026-08-31)

- 114 route pages
- 47 pages contain raw six-digit hexadecimal colours (1,069 occurrences)
- 5,294 inline-style blocks in application/component TSX
- 41 page-local status-colour-map candidates
- 29 pages use the shared status presentation
- Dark-mode configuration exists but no `.dark`/`prefers-color-scheme` rules exist

## Deliverables

1. A tracked design-system reference: tokens, scales, component a11y contracts, semantic colour map, and migration plan.
2. One token notation per concept, WCAG AA contrast guard for real token/surface pairs, and a deliberate theme decision: implement a tested dark theme or remove the undeclared capability.
3. Shared primitives for the recurring operator patterns (status/badge, button, form field, table, tabs, modal/drawer, empty/loading state, pagination/filter). Each must have real callers, accessible states, and bilingual-safe copy APIs.
4. CI ratchets: raw colours and new local status maps may only decrease; accepted exceptions are explicit and reviewed.
5. Ordered migration of the highest-density pages and regulated workflows, with visual/browser checks where CSS geometry matters.
6. Final evidence: keyboard and screen-reader contracts, responsive browser checks, performance/bundle evidence, GitOps sandbox rollout, and independent PR review for each merge.

## Safety invariants

- No change to authorization, approval separation, idempotency, backend payloads, or financial behaviour as part of a visual migration.
- Unknown status remains neutral; no status-to-colour map may claim success by guesswork.
- No new UI dependency unless it replaces demonstrated duplication.

## Initial linked work

- #7648 / #7649: semantic colour ratchet and approval identity badge
- #7651 / #7653: document-template status migration

This issue stays open until the requirements above have current, runnable evidence.

JiRaska:
**STILL TRUE as a tracker — but one baseline bullet has gone stale and should not be re-cited.**

The 2026-08-31 baseline says *"Dark-mode configuration exists but no `.dark`/`prefers-color-scheme` rules exist"*. That is no longer accurate on `origin/main`:

```
[...]

PR #9308, opened 2026-09-08
feat(admin-ui): unify bounded result pagination

JiRaska:
## Summary
- add a shared bounded-list footer that keeps the visible/total result extent understandable
- migrate account search and the masked card portfolio to one linked load-more interaction
- keep final extent visible after pagination and remove the exhausted action
- preserve backend query limits, account search behavior, card RBAC/lifecycle controls, and masked-PAN-only rendering

## Accessibility and performance
- link each action to its result surface with `aria-controls`
- announce incremental account progress while avoiding a duplicate live region on cards
- keep the initial DOM bounded to 25 rows and add only the requested remainder

## Verification
- focused Vitest: 3 files / 5 tests
- TypeScript type-check and changed-file ESLint
- production build: 97/97 routes
- Chromium E2E: 6/6 (both pagination flows, account request supersession and IBAN validation, masked-card outage and filter recovery)

Refs #7654


JiRaska:
## `load-more-control.spec.ts` describes client-side paging; the page does cursor paging

CI fails on the first assertion of `keeps account extent visible while incrementally rendering a
bounded result`:

```
Locator: getByText('Showing 25 of 30 accounts')
Error: element(s) not found
```

**This is not a merge artefact.** I checked the branch side of the merge commit against the
current head — `progressLabel` and `onLoadMore` are byte-identical before and after, so nothing
was lost in resolving the barrel conflict. The test has never matched the page it drives.

### The two models

The test expects the client to hold 30 rows and reveal 25, then the rest:

```
Showing 25 of 30 accounts     ->  click Load more  ->  Showing 30 of 30 accounts
ACCOUNT-26 not rendered yet
```

The page is cursor-paginated against the server:

```tsx
progressLabel={t(`Zobrazeno ${filtered.length} účtů`, `Showing ${filtered.length} accounts`)}
onLoadMore={() => void search(query, result?.pagination.nextCursor)}
```

It renders what it has — `Showing 30 accounts`, no "of" — and `Load more` fetches the next page
rather than revealing withheld rows. The mock returns all 30 with `hasNextPage: false` against a
request for `limit=25`, which a real `account-service` would not do.

`LoadMoreControl` itself is agnostic: `progressLabel` is a prop, and `hasMore` is documented for
exactly this split — *"A CURSOR-paginated caller must pass it: it has fetched one page, so `total`
is the size of what is on screen."* The component anticipated both callers; the test assumed the
other one.

### What needs deciding

Either:

1. **Keep cursor paging** (what the page and `account-service` actually do) and rewrite the test:
   mock `limit=25` returning 25 rows with `hasNextPage: true` and a cursor, assert
   `Showing 25 accounts`, click, then assert `Showing 30 accounts` after the second page. No
   "of N" — with a cursor the total is genuinely unknown, and inventing one would be the
   `PushResult.skipped()` mistake in a different costume: a number that looks authoritative and
   is not measured.
2. **Switch accounts to client-side paging** to match the test, which means fetching everything
   up front — a different product decision, and worse for a large estate.

I would take (1), but it is an authoring call about which paging model this page has, so I have
not rewritten someone else's test to match an implementation they may have intended to change.

The sibling `masked card extent` test in the same file will need the same treatment.

PR #9354, opened 2026-09-09
fix(admin-ui): preserve verified incident evidence

JiRaska:
## Summary

- validate the DORA incident envelope and every incident before rendering it as durable evidence
- reject HTTP failures, malformed dates and malformed protected fields instead of trusting a type cast
- retain the last verified register through transient or invalid refresh responses and label it as potentially stale
- clear retained protected incident data immediately when the current session is unauthorized
- bound the browser-side request and distinguish an unreachable source from an invalid response

## Verification

- TypeScript check passed
- focused ESLint: no errors (one pre-existing initial-load effect warning)
- incident response, recovery, UX, refresh and locale suites: 11 tests passed
- production build: 97/97 routes generated

Refs #7654


JiRaska:
Superseded by #9527, which carries the incident-evidence recovery on the current contract and additionally bounds requests, rejects stale responses, preserves only verified snapshots during transient failures, and clears protected evidence on authorization loss. #9527 is exact-head CI green and remains awaiting independent review.

PR #9381, opened 2026-09-09
fix(admin-ui): align AML case evidence

### LINKED ISSUE #9380: fix(admin-ui): align AML case evidence with service contract
## Problem
The AML admin page reads a UI-only model with customerName, customerType, score, and timestamp. The AML service instead returns customerReference, party ownership, screening type, categorical risk, alert evidence, assignment/decision provenance, and explicit audit timestamps. Existing browser coverage repeats the UI-only fixture, so it cannot catch the production mismatch.

The page also asserts that automated scanning is not configured even though the case-list endpoint provides no configuration evidence for that claim.

## Outcome
- Validate the actual AML case response before replacing the trusted snapshot.
- Present customer reference and party ownership, screening type, categorical risk, alert evidence, status, analyst provenance, and updated time.
- Remove invented numeric risk and unsupported environment claims.
- Preserve retained-snapshot recovery with a production-shaped browser fixture.

## Scope
Admin UI only; no AML decisions, service API, or authorization changes.

JiRaska:
## Summary
- validate the AML service's actual case contract before replacing the last trusted snapshot
- replace invented customer, score, and timestamp fields with customer reference, party ownership, screening type, categorical risk, alert evidence, analyst provenance, and updated time
- remove an unsupported environment-configuration claim
- update retained-snapshot browser coverage to use a production-shaped response

## Verification
- 5 focused Vitest checks passed
- TypeScript and ESLint passed
- production build generated 97/97 pages
- AML recovery Playwright E2E passed

Closes #9380


JiRaska:
## This is not a competing style — main's parser is falsified by the service contract

Resolving this conflict meant choosing between two parsers for `/api/v1/aml/cases`. The contract
settles it, and the answer is worse than "pick one".

`AmlCaseResponse` in `openbank-aml-service/src/main/resources/openapi.yaml`:

```
id  idempotencyKey  partyId  accountId  transactionId  customerReference
screeningType  riskLevel  status  alertCode  alertDetail  matchedEntity
decisionReason  assignedAnalyst  decidedBy  screenedAt  decidedAt  createdAt  updatedAt
```

main's `isAmlCase` requires `customerName`, `customerType`, `score` and `timestamp`. **None of
those four exists in the contract.**

Measured, with a response built field-by-field from that schema:

```
parseAmlCaseSnapshot(realResponse)  ->  { cases: [], excludedCount: 1 }
```

`/aml` therefore renders an empty case list for every operator, every time.

## Why nothing said so

Two reasons, and both generalise.

**The fail-soft posture converts the mismatch into the wrong diagnosis.** That parser's docstring
is "keep usable evidence while making partial upstream corruption measurable to the operator" —
a good design for partial corruption. Against a total contract mismatch it reports `excludedCount`,
which reads as *the upstream sent bad data*, and sends an investigator at the AML service rather
than at the parser. An empty compliance queue is also indistinguishable from a quiet day.

**The test cannot disagree with the parser.** `aml-case-evidence.test.ts` builds `validCase` by
hand in the shape `isAmlCase` expects, so the fixture and the predicate encode one mental model
and confirm each other. A parser and a test written from the same assumption are green by
construction; only the contract is an independent witness.

## Resolution

Takes this branch's `amlCaseContract.ts`: the field set matches the schema, and `AML_STATUSES`,
`AML_RISK_LEVELS`, `SCREENING_TYPES` are validated rather than accepted as free strings.

`lib/aml/cases.ts` and `aml-case-evidence.test.ts` are removed. Two contracts for one endpoint is
how this happened; keeping the falsified one is not an option.

Fail-fast is deliberate here — for compliance evidence an error the operator can see beats an
empty table they cannot tell from "no cases today".

Verified: full admin-ui suite, **513 files / 2786 passed, 1 skipped**. Nothing else imported the
removed module.

PR #9395, opened 2026-09-09
docs(sepa-payment): ADR-0300 returns idempotency contract (#8351)

JiRaska:
Refs #8351

Docs-only: re-verifies POST /api/v1/sepa-payments/returns as idempotent on the pacs.004 OrgnlEndToEndId natural key plus the RETURNED state guard (a re-delivered return replays, never reverses twice; the reversal leg carries sepa-reversal-{paymentId} into transaction-service). Records ADR-0300, re-annotates the baseline, openapi.yaml 1.7.0 -> 1.7.1 (description only). No code change.

## Security checklist

- [x] Docs-only change — no code, endpoint, role or privilege touched
- [x] Threat model: no trust-boundary change (verification + ADR record only)
- [x] Contract: openapi.yaml description-only bump


JiRaska:
**This PR's `Trivy` failure is a real advisory that `main` has already fixed — merging `main` clears it, with no change of your own.**

Its Trivy job reports:

```
sharp   GHSA-rgj7-g3m4-5g8c   HIGH   fixed   0.35.0 → 0.35.4
```

and the gate is `--ignore-unfixed`, i.e. "fail on any **fixed** HIGH/CRITICAL", so a fixable HIGH is exactly what reddens it.

`ee6af00f3` — `security(admin-ui): upgrade sharp to 0.35.4 (#9367)` — merged 2026-09-09T09:13:19Z and pins it through `overrides`. This branch does not contain that commit:

```
git merge-base --is-ancestor ee6af00f3 origin/<this branch>   ->  NO
```

Verified for this PR specifically rather than inferred from the branch date: the log names the advisory (2 mentions) and carries **zero** `rate limit exceeded` lines, so this is the real finding and not the installation rate-limit incident that is currently reddening Trivy across much of the queue (#8690). Those two look identical in the checks list and are not the same thing — a rate-limited Trivy produces no totals and names no advisory at all.

Nothing to change here; take `main` and it goes green.

PR #9402, opened 2026-09-09
fix(clearing): replay-safe payment submit + ADR-0298 natural-key idempotency (#8351)

JiRaska:
Refs #8351

## What

`POST /api/v1/clearing/submit` was the last money-path creation POST with no idempotency: a retried submit stacked a second PENDING `ClearingItem` for the same payment, and the clearing cycle sweeps all PENDING items into batches — the same payment would have settled twice.

- Natural key: `paymentId` (a payment enters clearing exactly once). `submit` now reads `findByPaymentId` first and replays the existing item.
- DB backstop: V9 `uq_clearing_items_payment` unique index; a lost true-concurrency race re-reads the winner instead of erroring the caller (23505 on that constraint).
- No synthetic Idempotency-Key: the designed callers already hold the stable paymentId, and no REST caller exists today (see the class comment).

## Security checklist

- [x] Money-path service: threat model updated (2026-09-09 entry)
- [x] No new route, caller, role or privilege; a retry sees the original item
- [x] Migration V9 additive (unique index) with rollback note + pre-duplicate detection query
- [x] Contract updated: openapi.yaml 1.4.0 → 1.4.1 (description only)
- [x] Tests: replay case (save 0×), unique-race recovery case, original save path

## Test evidence

:openbank-clearing-service:test --tests "*ClearingServiceTest*" + ktlint + detekt — BUILD SUCCESSFUL.


JiRaska:
**This PR's `Trivy` failure is a real advisory that `main` has already fixed — merging `main` clears it, with no change of your own.**

Its Trivy job reports:

```
sharp   GHSA-rgj7-g3m4-5g8c   HIGH   fixed   0.35.0 → 0.35.4
```

and the gate is `--ignore-unfixed`, i.e. "fail on any **fixed** HIGH/CRITICAL", so a fixable HIGH is exactly what reddens it.

`ee6af00f3` — `security(admin-ui): upgrade sharp to 0.35.4 (#9367)` — merged 2026-09-09T09:13:19Z and pins it through `overrides`. This branch does not contain that commit:

```
git merge-base --is-ancestor ee6af00f3 origin/<this branch>   ->  NO
```

Verified for this PR specifically rather than inferred from the branch date: the log names the advisory (2 mentions) and carries **zero** `rate limit exceeded` lines, so this is the real finding and not the installation rate-limit incident that is currently reddening Trivy across much of the queue (#8690). Those two look identical in the checks list and are not the same thing — a rate-limited Trivy produces no totals and names no advisory at all.

Nothing to change here; take `main` and it goes green.

PR #9408, opened 2026-09-09
security(admin-ui): bump sharp 0.35.0 → 0.35.4 (GHSA-rgj7-g3m4-5g8c)

JiRaska:
Refs #8351

## What

Trivy now fails every open PR on the fixed HIGH vulnerability in sharp's bundled libheif (GHSA-rgj7-g3m4-5g8c, fixed in 0.35.4). Bumps both the direct dependency and the matching override, lockfile regenerated.

## Security checklist

- [x] Dependency-only change; no code, endpoint or privilege touched
- [x] Fixes a Trivy-gate-blocking HIGH rather than ignoring it (no .trivyignore entry)
- [x] Lockfile regenerated via npm install --package-lock-only

### DIFF
--- a/openbank-admin-ui/package-lock.json
+++ b/openbank-admin-ui/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "openbank-admin-ui",
-  "version": "0.240.1",
+  "version": "0.241.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "openbank-admin-ui",
-      "version": "0.240.1",
+      "version": "0.241.0",
       "dependencies": {
         "@hookform/resolvers": "5.9.1",
         "@opentelemetry/api": "^1.9.1",
@@ -46,6 +46,7 @@
         "recharts": "3.10.1",
         "rehype-raw": "^7.0.0",
         "remark-gfm": "^4.0.0",
+        "sharp": "0.35.4",
         "sonner": "2.0.8",
         "tailwind-merge": "3.6.0",
         "xml2js": "^0.6.2",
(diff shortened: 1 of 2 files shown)


JiRaska:
**The advisory this closes is already closed on `main` — and the part of this diff that is not redundant looks like the wrong mechanism.**

## The security goal is already met

`ee6af00f3` — `security(admin-ui): upgrade sharp to 0.35.4 (#9367)` — merged 2026-09-09T09:13:19Z and sets

```json
"overrides": { "sharp": "0.35.4" }
```

so GHSA-rgj7-g3m4-5g8c is resolved on `main` today. This branch does not contain that commit (`git merge-base --is-ancestor ee6af00f3 origin/security/admin-ui-sharp-cve` → NO), which is why it still looks outstanding from here.

## What this branch adds beyond it

Reading the real squash delta (`git diff origin/main...origin/security/admin-ui-sharp-cve` — 2 files, +182/−131; the 2-dot diff shows 236 files and is the documented stale-branch artifact, not a revert):

| | `dependencies` | `overrides` |
|---|---|---|
| `main` | — | `sharp 0.35.4` |
| this branch | **`sharp 0.35.4`** | `sharp 0.35.4` |

So the only substantive difference is promoting sharp to a **direct dependency**, and I do not think that is wanted:

- **admin-ui never imports it.** `git grep "from ['\"]sharp['\"]\|require(['\"]sharp['\"])" -- 'openbank-admin-ui/**'` returns nothing outside `node_modules`.
- **It arrives through `next`, as an OPTIONAL dependency.** From `main`'s lockfile: `node_modules/sharp` → `version 0.35.4, optional: true`, and the only package declaring it is `node_modules/next`.

sharp ships platform-specific native binaries, which is why `next` takes it optionally — an install on a platform with no prebuilt binary can skip it and Next falls back. Declaring it in `dependencies` makes that install **required**, so a platform gap becomes a hard install failure rather than a degraded image pipeline. `overrides` pins the version without changing whether it must resolve, which is the mechanism that matches the actual relationship.

## Suggested disposition

Close as superseded by #9367, unless the direct dependency is deliberate — in which case it wants its own PR and a sentence about what imports it, because nothing in the tree does today.

## How I got here

Auditing which open PRs are red on Trivy purely because of the ongoing installation rate-limit incident (#8690). Two of them — #9402 and #9395 — turned out to be red on this **real** advisory instead, because their branches predate #9367:

```
sharp  GHSA-rgj7-g3m4-5g8c  HIGH  fixed  0.35.0 → 0.35.4
```

Of the 15 PRs red only on the Trivy/dependency-review trio, **8 carry #9367 and 7 do not**; the seven clear by merging `main`, no change of their own required. Worth noting my first probe missed this entirely — it grepped the logs for `CVE-[0-9]{4}-`, and a GitHub advisory id is `GHSA-…`, so a real finding read as "no findings" on both PRs.

JiRaska:
**Correction to the last line of my comment above: "the seven clear by merging `main`" is wrong — only two do.**

I inferred it from ancestry alone (7 of the 15 lack `ee6af00f3`) without checking what each one's Trivy job actually failed on. Reading all seven logs:

| PR | `GHSA-rgj7-g3m4-5g8c` mentions | `rate limit exceeded` | Trivy totals |
|---|---|---|---|
| #9402 | **2** | 0 | `Total: 1 (HIGH: 1, CRITICAL: 0)` |
| #9395 | **2** | 0 | `Total: 1 (HIGH: 1, CRITICAL: 0)` |
| #8941 | 0 | 6 | none |
| #8881 | 0 | 6 | none |
| #8806 | 0 | 6 | none |
| #8804 | 0 | 6 | none |
| #8802 | 0 | 6 | none |

**Lacking the fix is not the same as failing on it.** The rate limit kills the scan before it reaches the dependency tree at all — those five produce no totals and name no advisory, so merging `main` changes nothing for them. They are waiting on #8690 like the other eight.

Corrected split of the 15 PRs red only on the Trivy / dependency-review trio: **2 on a real advisory** (clear by merging `main`), **13 on the rate limit** (nothing to do on the branch).

Nothing above this correction changes — #9367 already resolves the advisory on `main`, and the direct-dependency question stands on its own.

JiRaska:
Superseded by #9367, which already merged the same sharp 0.35.4 security fix. Current main and this PR both resolve sharp to 0.35.4; rebasing this stale lockfile-only branch would add no security value and would consume another full CI cycle.

PR #9456, opened 2026-09-09
fix(admin-ui): expose complete payment evidence

### LINKED ISSUE #9455: fix(admin-ui): expose complete payment list evidence
## Problem
The admin Payments page converts every failed SEPA or domestic list request to an empty array and then renders "No payments found". It also requests only each service default page (50 records) and exposes no continuation, so a partial outage or truncated history looks complete.

## Scope
- validate and normalize the documented payment list arrays
- track SEPA and domestic availability independently
- preserve verified rows when only one source fails and label the snapshot as partial
- request explicit bounded limit/offset pages and expose continuation per source
- distinguish a genuinely empty available snapshot from unavailable data

## Acceptance criteria
- failures never become a genuine empty state
- one healthy source remains usable while the failed source is clearly named
- every source can advance until its returned page is shorter than the requested limit
- malformed rows do not enter operator-visible totals or navigation
- stale requests cannot replace a newer snapshot

JiRaska:
## Summary
- distinguish unavailable SEPA/domestic payment sources from genuine empty history
- strictly validate and normalize list evidence before rendering it
- forward the supported payment-list query contract through both BFF routes
- add independent offset pagination per payment source without discarding healthy partial results

## Verification
- 18 focused Vitest tests passed
- `npm run type-check` passed
- targeted ESLint: 0 errors (2 pre-existing hook warnings)
- `npm run build`: 97/97 routes
- `git diff --check` passed

Closes #9455


JiRaska:
## Sequenced behind #9209 — deliberately not resolved yet

This PR and #9209 both rewrite `openbank-admin-ui/src/app/payments/page.tsx`, and #9209 is not
merged. Resolving this conflict now means resolving against a `main` that is about to change
underneath it, in the same file, and the likeliest casualty is #9209's fix rather than anything
here.

That fix is not cosmetic, which is why the order matters: main's `onCloseAutoFocus` stored
`document.activeElement` as the return target, and that is `<body>` whenever nothing holds focus
(a form submitted programmatically, or by Enter after the field blurred). `<body>` is always
`isConnected`, so the fallback was unreachable and `.focus()` on it a no-op — the operator landed
nowhere. Merging this branch's version of the file first would quietly reinstate it.

The conflict here is also not mechanical. This branch replaces the whole load path — per-source
pagination, per-source evidence, an explicit 401 boundary that clears BOTH sources, and a
generation counter guarding against out-of-order responses — against main's simpler loader plus
`parseVopEvidence`. Taking either side wholesale drops real behaviour from the other.

**Order:** land #9209 (small, scoped to the dialog's focus handling), then rebase this one and
resolve the loader conflict once, against a file that already carries the focus fix.

No changes pushed here.

PR #9574, opened 2026-09-10
ci(admin-ui): realign raw colour subject floor

### LINKED ISSUE #9573: CI: realign raw-colour gate subject floor after UI cleanup
The enforced raw-colour-literal-ratchet reports 1,591 valid subjects against a stale min_subjects floor of 1,600. The ceiling still passes at 1,782, so the gate fails because successful UI cleanup crossed its corpus floor, not because raw colour literals increased. Deliberately realign the subject floor while preserving a meaningful no-op detector and add or retain falsifiable gate coverage.

JiRaska:
## Summary

- preserve the raw-colour ceiling at 1,782
- realign the stale subject floor after semantic-token cleanup reduced the valid corpus to 1,591
- retain a roughly five-percent no-op detection margin at 1,510 subjects

## Verification

- raw-colour checker self-test: 4 passed
- raw-colour checker: 1,591 subjects, ceiling passed
- run-gates.py --only raw-colour-literal-ratchet: passed
- git diff --check

Closes #9573

### DIFF
--- a/.github/gates/gates.yaml
+++ b/.github/gates/gates.yaml
@@ -3872,7 +3872,7 @@ gates:
       customer reported it (#4348). Baseline is empty by measurement, so this gate is prevention
       only — recheck whether the fleet still has no occurrences, not whether debt was paid down.
     review_after: '2027-02-22'
-    min_subjects: 1500
+    min_subjects: 1510
     selftest: python3 .github/scripts/check-noop-not-success.py --self-test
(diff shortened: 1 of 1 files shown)


JiRaska:
Superseded by merged #9486, which fixes the gate at the correct ownership boundary by separating scanned subjects from remaining colour debt. No waiver or threshold relaxation is needed.

PR #9595, opened 2026-09-10
fix(dispute): a withdrawal's outcome must be JSON null, not the string "null"

JiRaska:
## The defect

#9057 made a terminal `update` emit `dispute.resolved` — correct, and what #8745 finding 2 asked for. It also put every **withdrawn** dispute on the wire with a malformed field:

```kotlin
""""outcome":"${dispute.remediationOutcome}",""" +
```

The interpolation is **inside** the quotes. `withdraw` delegates to `update`, `update`'s `copy()` does not set `remediationOutcome`, and the domain default is `null` — so the payload carries the four-character string `"null"` in a quoted field instead of JSON `null`.

Traced end to end on `origin/main`:

| step | |
|---|---|
| `withdraw(id, actor)` | `update(id, UpdateDisputeRequest(status = WITHDRAWN, resolution, resolvedBy))` — no outcome |
| `update`'s `copy(...)` | sets `status`, `resolution`, `chargebackAmount`, `resolvedBy`, `resolvedAt`, `updatedAt` — **not** `remediationOutcome` |
| `DisputeModels.kt:52` | `val remediationOutcome: RemediationOutcome? = null` |
| `resolvedOutboxMessage` | renders `"outcome":"null"` |

Before #9057 a withdrawal emitted nothing at all, so this is **new on the wire** rather than long-standing.

## Blast radius

Small but real. `DisputeOpenedEventConsumer` reads only `eventType` and `partyId`, so the ADR-0220 exclusion still lifts correctly — the behaviour #9057 was written to fix is not affected. audit-service stores the payload as it arrives, so the bad field lands in the audit record for every withdrawal, and any consumer added later that reads `outcome` gets the string `"null"` where it expects an absent value or a `RemediationOutcome`.

## Scope of the fix

**Only `resolvedOutboxMessage` is changed.** `remediationRequestedOutboxMessage` has the identical shape and is deliberately left alone: it is reached solely from `doResolve`, where `remediationOutcome = request.outcome` is non-null *by type*, so it can never see a null.

**Also declares `TERMINAL_STATUSES` once.** #9057 spelled the same three names out twice inside `update` — once for the `resolvedAt` stamp, once for the emission guard. Those two disagreeing about what "terminal" means is precisely how a terminal transition came to stamp a resolution time while announcing nothing, which is the defect #9057 fixed. Behaviour-identical; happy to split it out if you would rather review the wire fix alone.

## Why the existing test did not catch it

`withdraw sets withdrawn status` mocked `disputeRepo.update(any(), any())` and never captured or opened the payload — it proved an event was emitted and nothing about what was in it.

It now **parses** the payload:

```kotlin
val node = ObjectMapper().readTree(outbox.captured.single().payload)
assertThat(node.get("outcome").isNull).isTrue()
```

A substring assertion (`contains("\"outcome\":null")`) would also pass on a payload that is not valid JSON at all, which is the failure mode one step along from this one.

## Verification — falsified, not just observed green

| tree | `DisputeServiceTest` |
|---|---|
| fix in place | **13/13**, 0 failures, 0 errors, **0 skipped** (JUnit XML) |
| production line reverted | **exactly 1 red** — `withdraw sets withdrawn status`, on the `outcome` assertion |

`ktlintCheck` + `detekt` green. The production diff is 3 hunks and nothing else — `ktlintFormat` was run for import ordering and reformatted no unrelated code.

## Scope

`openbank-dispute-service` is not in `rules.yaml: money_path_services`. No API surface change, no migration, no new event type — the event and its topic are unchanged, only one field's serialisation is corrected.

Refs #8745

## Security checklist

      (license, CVE, source).

Checked against the diff: two files, no `@Suppress`/`as Any` added, no dependency file touched, no secret-shaped line. `ObjectMapper` is a test-only import of Jackson, already on the classpath. No auth, crypto or payment path changes; the payload gains no field and no new recipient — one existing field stops being emitted as a quoted string.


JiRaska:
**`Trivy` and `Dependency review` here are the ongoing installation rate-limit incident, not findings against this diff.**

Read from both job logs on this PR's head:

| check | `rate limit exceeded` / "not supported on this repository" lines | CVE rows |
|---|---|---|
| Trivy | 6 | **0** |
| Dependency review | 1 | **0** |

Zero CVE rows in either — they died on the API, not on a finding. `main` itself is red on the same two workflows right now (head `1f4db5755`: `Security scan` and `Release Please`, 6 and 3 rate-limit lines respectively), so this is inherited, not introduced.

Tracked as #8690, which I updated today with the current numbers. The remedies are all owner-side (App/org rate budget, a dedicated token as a repository secret, or throttling `admin-ui-deploy`'s ~345 API requests per build), so there is nothing to do on this branch.

The diff itself is two Kotlin files and adds no dependency, so `Dependency review` has nothing of this PR's to review even when it runs.

Local verification stands as described in the body: `DisputeServiceTest` 13/13 with 0 skipped, `ktlintCheck` + `detekt` green, and falsified — reverting the one production line reddens exactly one test on the `outcome` assertion.

PR #9670, opened 2026-09-11
fix(sepa-payment): publish the real SepaPaymentStatus vocabulary

JiRaska:
## Summary

Neither published status enum in `openbank-sepa-payment` has ever matched `SepaPaymentStatus`.
Both advertised `RECALLED`, which has never existed in the code; the list filter also advertised
`PENDING`; and between them `RECEIVED`, `VALIDATED`, `RETURNED` and `CANCELLED` — all reachable,
all returned — went unpublished. A client generated from this contract can neither filter for four
real states nor send two of the six transitions the service accepts.

## Type of change

- [x] fix (bug fix)
- [ ] perf (performance)
- [ ] docs
- [ ] chore (build / tooling)
- [ ] breaking change

## Linked issues

Refs #5962

Replaces #5990, whose branch is an add/add conflict against the current
`check-openapi-enum-vs-domain.py` on `main`; the sepa half is rebuilt fresh here.

## Changes

- `listSepaPayments` `status` filter -> the full domain enum, all seven values. It now pairs
  exactly with `SepaPaymentStatus`, so it is no longer baselined at all.
- `TransitionStatusRequest.targetStatus` -> the six values `SepaPayment.canTransitionTo` can
  reach. `RECEIVED` is the entry state and is the target of no transition, so it is deliberately
  absent; recorded as a deliberate subset in the gate baseline rather than silently omitted.
- `.github/scripts/check-openapi-enum-vs-domain.py`: both stale `#5962` entries dropped, one
  deliberate-subset entry added with the reason above. The gate fails on a stale entry in either
  direction, so this has to land with the spec change.
- `info.version` 1.7.0 -> 1.8.0.

## Definition of Done

- [ ] Unit tests added/updated. — spec-only change; no Kotlin touched.
- [ ] `./gradlew detekt ktlintCheck koverVerify build` passes. — no Kotlin touched.
- [x] No new lint warnings.

Verified locally against current `main`:
- `check-openapi-enum-vs-domain.py --self-test` — passed
- `check-openapi-enum-vs-domain.py --enforce` — `SUBJECTS=219`, `15 baselined; no new drift`, 0 stale
- The six transition targets were read off `SepaPayment.canTransitionTo`, not assumed.
- `RECALLED` elsewhere in the tree belongs to `openbank-sepa-instant`'s `SctInstStatus`, a
  different service where it genuinely exists; untouched, including the admin-ui badge branch
  that renders it.

## Security checklist

- [x] Auth / crypto / payment code changed? — money-path SERVICE, but no auth/crypto/payment
      logic is touched; this is its published vocabulary only.
- [x] PII path changed?
- [x] Cardholder data path changed?

## Compliance impact

- [x] None

The document now states what the service already does; no behaviour changes.

## Rollout / Rollback

- Deployment strategy: rolling — spec-only, no runtime change.
- Rollback plan: revert the commit; the two baseline entries return with it.

## Reviewer notes

The version bump is MINOR under the ADR-0048 **correction** class, not MAJOR. oasdiff will read
the two removals as breaking — they are not, because `RECALLED` and `PENDING` could never be
produced or accepted by the running service, so no conformant client can regress. Per that class
this PR deliberately touches nothing else inside `openbank-sepa-payment/`.


JiRaska:
**Merge order matters: this one goes FIRST.**

#8833 corrects `CreateSepaPaymentRequest` in this same `openapi.yaml`. The two do not conflict in content — this PR touches the status vocabularies (list filter + `TransitionStatusRequest.targetStatus`), #8833 touches the create request schema — but `info.version` is a single number and both originally claimed 1.8.0 off `main`'s 1.7.0.

Sequenced rather than left to chance: this PR keeps **1.8.0**, #8833 has been moved to **1.9.0** and says so in its own header comment. If this PR is closed instead of merged, #8833 drops back to 1.8.0.

PR #9710, opened 2026-09-11
feat(kyb): a human confirms the representation rule, per IČO — no agreement bound by a parser

### LINKED ISSUE #9709: KYB: teach the Czech representation-rule parser the joint phrasings it reads as UNKNOWN (31% of live texts)
## Problem

`CzechRepresentationRuleParser` turns the register's free-text *způsob jednání* into a
`RepresentationRule`, and `RegistryExtract.signaturesRequired()` turns that into a signature count.
Anything it cannot parse is `UNKNOWN`, which `BusinessOnboardingCase` routes to `MANUAL_REVIEW` —
so an unparsed rule is safe, but it is also an onboarding that stops.

Measured 2026-09-11 against **142 live `způsob jednání` texts** pulled from ARES
(`ekonomicke-subjekty-vr/{ico}`, sample drawn from 15 name searches across several towns and legal
forms):

| verdict | count |
|---|---|
| SOLE(1) | 88 |
| JOINT_N(2) | 8 |
| JOINT_ALL | 2 |
| **UNKNOWN** | **44** (31%) |

**The safe direction holds today.** Zero of the 88 `SOLE` verdicts contain any second-signer marker
(`dva`, `dvou`, `oba`, `společně`, `a další`, `nejméně`, `alespoň`) — no joint rule was read as
solo, which is the failure that would let the bank contract a company on one signature where the
register demands two. That property currently rests on a single `!t.contains("spolecne")` guard,
which is narrower than the evidence: several real texts require two signatures without ever using
the word *společně*.

## What the 44 unknowns actually are

**1. Word order. The parser assumes noun-then-adverb; the register writes both orders.** One cause,
both modes:

- `Za společnost jedná samostatně jednatel.` — `samostatne` precedes the noun, so the `solo` regex
  (which needs `jednatel …samostatne`) misses it.
- `Jménem společnosti jednají společně dva členové představenstva` — `spolecne` precedes the count,
  so `jointlyN` (which needs `dva …clenove …spolecne`) misses it.

**2. The družstvo written-form clause**, which is the most common shape in the sample:

> Za družstvo jedná navenek předseda nebo místopředseda. Je-li však pro právní úkon … předepsána
> písemná forma, je třeba podpisu **alespoň dvou členů představenstva**.

A framework agreement is in written form, so the operative count is two — and this text never says
*společně*. It is also the exact shape that makes the current SOLE guard fragile: `předseda …
jedná` plus no `společně` is one regex tweak away from reading as SOLE.

**3. A role pair joined by `a`, not `spolu s`:** `předseda představenstva a další člen
představenstva společně`. `jointlyPair` requires the literal `společně s`.

**4. A parenthesised numeral:** `vždy dva (2) členové představenstva společně` — the number and the
noun are not adjacent.

**5. Role-specific rules, which are a different class and must not be flattened into a count.**
Kofola: *zastupují vždy společně předseda představenstva spolu s jedním členem představenstva.*
Emitting `JOINT_N(2)` here would be wrong in the dangerous direction: any two board members would
satisfy a count, while the register requires *the chair* plus one. The rule model carries no role
today, so a count is the only thing it can say.

## Proposal

1. **Make the safe direction structural, not incidental.** Add an explicit second-signature veto
   (`alespoň/nejméně dvou`, `a další člen`, `podpisu … a místopředsedy`, a numeral + signer noun)
   that forbids a `SOLE` verdict regardless of what else matched. Today the property holds because
   no sampled text happened to trip it; it should hold because nothing can.
2. **Accept both word orders** for solo and for joint-N, and allow a short gap between a numeral and
   its noun.
3. **Teach the družstvo written-form clause** as `JOINT_N(2)`.
4. **Carry the role constraint in the model** rather than flattening it. A rule that names *which*
   office must sign gets a flag; `BusinessOnboardingCase` routes it to `MANUAL_REVIEW` exactly as
   `UNKNOWN` does today — the outcome is unchanged, but it becomes a stated, countable category
   instead of a parse failure, and it can never be mistaken for "any two will do".

Point 4 is the one that matters for the question that prompted this: the system must not contract a
company without the signatures of the *correct* people, and a count alone cannot express that.

Refs ADR-0284, #9707

JiRaska:
## Summary

**No agreement is bound by a parser.** An operator confirms how an entity is represented, once per
IČO, and that confirmation is what the onboarding case uses. The heuristic parser still reads the
register's *způsob jednání* — better than it did, 44 UNKNOWN of 142 live texts down to 25 — but its
verdict is now a *suggestion on the review form*, never a decision.

## Why the parser alone was not enough

The previous commit here made the parser's dangerous verdict the guarded one: SOLE is withheld
whenever any second-person, condition, amount or negation marker appears. That is worth having, and
it guarantees only that a text carrying a **known** marker cannot read as solo. A second-signature
clause phrased with none of them would still parse SOLE, and nothing downstream could tell. For a
control that decides whether the bank contracts a company on one signature or two, that is the wrong
shape of guarantee.

So `registryVerified` now takes a `RepresentationDecision`, and only `Attested` proceeds:

```
GET  /api/v1/kyb/representation/{scheme}/{identifier}          the review form
POST /api/v1/kyb/representation/{scheme}/{identifier}          confirm
GET  /api/v1/kyb/representation/{scheme}/{identifier}/history  audit trail
```

The parser's verdict is still computed and stored **beside** the human's answer — as the form's
suggestion, and so the two can be compared later to measure whether the heuristic is improving.
`RepresentationAttestation.correctedTheParser` is that measurement. It never gates.

### Keyed by IČO *and* the rule text

A company that amends its způsob jednání keeps its IČO — and that amendment is the single event this
control exists for, because it is exactly when the old signature count becomes wrong. Keying on the
identifier alone would carry yesterday's confirmation onto today's different rule: the failure this
feature is meant to prevent, arriving through the control itself. The key is therefore the
identifier plus a SHA-256 of the register text, taken after the parser's own folding, so a re-fetch
differing only in case, diacritics or whitespace is the same rule and costs no second review.

Three outcomes, deliberately not two:

| state | meaning | case |
|---|---|---|
| `ATTESTED` | a human confirmed *this* text | proceeds on the human's numbers |
| `UNATTESTED` | never confirmed | review, parser's suggestion shown |
| `SUPERSEDED` | confirmed before, **text has changed** | review, quoting what was confirmed and by whom |

`SUPERSEDED` exists because answering "unattested" there would be safe but silent — a reviewer shown
a blank form for a company they recognise will confirm from memory. Submitting a hash that no longer
matches the register is a **409**: the operator must re-read what they are confirming.

### Offices are carried, not flattened

An attested rule naming offices makes `cosignersInvited` require each office filled by a **distinct**
signer. A count of two is met by two ordinary members; and one person whose register role contains
both office words is one signature, not two. A one-signature office rule invites nobody, so
`initiatorMatched` checks it there rather than skipping to `READY_TO_SIGN` on the wrong person.

## Every guard was falsified by removing it

A guard is only proven by what it prevents, so each was deleted and the suite re-run:

| remove | tests that go red |
|---|---|
| `Unattested` proceeds like `Attested` | 4 |
| `Superseded` reuses the old confirmation | 3 |
| office distinctness | 3 |
| office check skipped entirely | 5 |
| stale-hash check | 3 |
| SOLE whitelist (parser) | 3 |

One of these was written badly the first time and caught by this: the distinctness test passed
against the broken code because the *count* check upstream rejected the case first, so it proved
nothing. It now selects two signers, which satisfies the count, leaving distinctness as the only
thing that can fail.

## Two defects the integration test found, invisible to a unit test

- The attestation table had **no Hibernate sequence** — every insert died on
  `relation "kyb_representation_attestations_seq" does not exist`. A mocked repository issues no SQL.
- Then it had **two id sources**: a `BIGSERIAL` default and that sequence, both starting at 1. A
  raw-SQL insert took id 1 and collided with the app's first insert on the primary key. The column
  is `BIGINT` now, with the sequence as the single source — and the migration says why, because
  anything backfilling this table by hand would hit it again.

## Type of change

- [ ] fix (bug fix)
- [ ] perf (performance)
- [ ] docs
- [ ] chore (build / tooling)
- [ ] breaking change

## Linked issues

Closes #9709, Refs ADR-0284

## Changes

- `RepresentationRule.requiredRoles` + `isRoleConstrained`; `signaturesRequired()` answers null for
  a role-constrained rule, and `BusinessOnboardingCase` routes it to `MANUAL_REVIEW` naming the
  offices.
- Both Czech word orders for solo and joint-N; a short gap between a numeral and its noun
  (`dva (2) členové`).
- The družstvo written-form clause (`je třeba podpisu alespoň dvou členů představenstva`) as
  `JOINT_N(2)` — it demands two signatures without ever saying *společně*.
- A role pair joined by `a` as well as `spolu s`; a count qualified by `z nichž jeden musí být
  předsedou`.
- Two pre-existing defects the new tests found (below).

## Measured, 142 live ARES texts (2026-09-11)

`ekonomicke-subjekty-vr/{ico}`, sample drawn from 15 name searches across several towns and legal
forms.

| verdict | before | after |
|---|---|---|
| SOLE | 88 | 97 |
| JOINT_N (count) | 8 | 10 |
| JOINT_N (role-constrained) | — | 7 |
| JOINT_ALL | 2 | 2 |
| **UNKNOWN** | **44** | **25** |

**Zero** of the 97 SOLE verdicts demands a second signature (`společně`, `spolu s`, `alespoň dvou`,
`nejméně dvou`, `a místopředsedy`, a numeral + signer noun). That is the failure that matters here —
reading a joint rule as solo would let the bank contract a company on one signature where the
register demands two.

## Reviewer notes — where the safety actually comes from

**SOLE is a whitelist, not a fallback.** The first version of this PR had it the other way round —
SOLE was a positive pattern, and any joint phrasing the joint patterns did not know fell through to
the one verdict that can cost the bank a signature. I claimed no SOLE verdict in the corpus demanded
a second signature; that was true of the 142 sampled texts and false of the parser. Written
adversarially, **7 of 14** joint rules answered SOLE:

- `Jednatel jedná samostatně ve věcech do 100 000 Kč; jinak jednají oba jednatelé.`
- `Jednatelé nejednají samostatně; jednají vždy dva jednatelé.`
- `Jednatel jedná samostatně pouze se souhlasem druhého jednatele.`
- `Jednatel jedná samostatně vyjma věcí nad 1 000 000 Kč, kde jedná spolu s druhým jednatelem.`

Each is one signature short of what the register demands. Now SOLE is withheld whenever the text
carries **any** marker of a second person (a count, `oba`/`obou`, `druhý`, `další`, `jiný`), joint
wording in any inflection (`společně`/`společný`/`společným` — never `společnost`), `spolu`,
`souhlas`, a plural `podpis`, a condition (`-li`, `pokud`, `vyjma`, `však`), a threshold (any digit,
`Kč`, `částka`) or a negation (`nejedná`, `nemůže`, `není oprávněn`, `žádný`). Such a text is either
claimed by a joint pattern or goes to review. All 14 adversarial texts are a permanent test
(`no text that involves a second person or a condition can ever answer SOLE`); disabling the
whitelist reddens it.

Three solo shapes carry a harmless marker and are tolerated **per shape, never per word**: a `-li`
that introduces the head-count (`Je-li jednatelů více, jedná každý samostatně`), consent of an
internal body (`valná hromada`, `dozorčí rada`), and the register's own `Počet členů statutárního
orgánu: 1` line. `se souhlasem druhého jednatele` stays blocked. Pinned by a second test.

Joint verdicts still run **before** the solo branch, so a text matching a joint pattern never
reaches it — `a written-form second signature outranks the solo opening` fails if the branches are
ever reordered.

**The role constraint is not cosmetic.** Kofola's rule is *předseda představenstva spolu s jedním
členem představenstva*. Any two board members satisfy a count of two and do not satisfy that rule,
so emitting `JOINT_N(2)` would be wrong in the dangerous direction — the right number of the wrong
people. `signaturesRequired()` therefore returns null for these rather than a number a caller could
check and stop at; the case goes to `MANUAL_REVIEW` naming the offices. 7 of 142.

**Downstream, the chain is closed:** the initiator must be a listed representative (else review),
co-signers are chosen only by index into `extract.representatives`, and `signed()` requires an
`IDENTIFIED` signer with a `partyId` and enforces the count. This parser is the only place that
decides *how many* and *who* — which is why the dangerous verdict is the guarded one.

**Two pre-existing defects, both found by writing tests rather than by reading the code:**

- The numeral alternation had no word boundary, so `jednají` matched the numeral `jedna` and
  `Za společnost jednají vždy dva jednatelé společně` answered **1** required signature instead of 2.
  Bad direction, and it was live.
- The bare `Samostatně` form (the whole rule as one word) had no pattern and read as UNKNOWN.

## Definition of Done

- [x] No new lint warnings.

Domain-only change; no REST surface, no DB, no event, no new dependency. Module runs 59 tests,
0 skipped, 0 failures. `requiredRoles` defaults to empty, so every existing construction site is
unchanged.

## Security checklist

The one `@Suppress` is the pre-existing `MagicNumber` on the Czech numeral map, whose values *are*
the numbers. The test texts are real company rules from a public register; no personal data — the
one sampled text naming individuals is not reproduced here or in the tests.

## Compliance impact

- [ ] PCI DSS
- [ ] DORA
- [ ] GDPR
- [ ] PSD2 / SCA / RTS
- [x] AML / 5AMLD
- [ ] CNB reporting
- [ ] None

This is the control that decides how many — and now which — signatures a business agreement needs.
The change only ever moves a case toward stricter handling: previously-UNKNOWN texts become a count
or a named-office review, and no text that reviewed before now auto-proceeds on fewer signatures.

## Rollout / Rollback

- Rollback plan: revert the PR. Cases already in `MANUAL_REVIEW` are unaffected; the parser is
  evaluated at registry-verification time, not persisted as a decision.


JiRaska:
## Auto-merge turned OFF — an adversarial review found three false ACCEPTs. Fixed in 940a75b.

This PR had auto-merge on, which meant the control that decides *how many signatures, and from
which offices, a business agreement legally requires* would have landed with no human review. I
commissioned a review for exactly that reason. It found three ways a case could be bound by
signatures from the **wrong offices** — all reachable, none covered by the 95-test suite.

### 1. The offices were checked at INVITATION and never again

`cosignersInvited` deliberately permits more invitees than the rule needs ("a company may want every
director on it"). So the people who turn up are a *subset* of the people who were checked — and a
subset that reaches the **count** can miss an **office**:

- Rule: chair + vice-chair, 2 signatures.
- Invite chair, vice **and** an ordinary member → the invited set covers both offices → passes.
- The vice never onboards. Member identifies → `identified = 2 ≥ 2` → `READY_TO_SIGN`.
- Chair signs, member signs → `signedCount = 2` → `SIGNED` → `ACTIVE`.

The agreement is bound by **chair + ordinary member** where the register requires **chair +
vice-chair**. `signed()` completed on `signedCount` alone and never consulted the offices.

Now re-checked over the people who actually **signed**, and `recomputeReadiness` will not even
*open* signing until the identified people can cover them.

### 2. `mistopredseda` contains `predseda`

Matching was a bare substring test, and `fold("místopředseda představenstva")` **contains**
`predseda`. Two vice-chairs — boards routinely have more than one — satisfied a chair + vice rule
with no chair anywhere on the agreement. An office must now begin a word.

### 3. `reviewResolved` recorded the office constraint and nothing enforced it

It set `requiredSignerRoles`, then went `INITIATOR_MATCHED` → `recomputeReadiness`, which compares
counts only. A one-signature office rule could be completed by whoever happened to be the initiator,
while `CaseResponse` advertised the office it had just recorded.

### Also fixed

- **Greedy first-fit could REFUSE a coverable set** — offices `[predseda, mistopredseda]` with the
  vice listed first. Replaced with a real bipartite matching; the sets are tiny, so it costs nothing.
- **Unguarded `representatives[index]`** → unmapped 500 for a sole-trader case with a bad index.
- **The parser KDoc claimed a structural guarantee it does not have.** `soleForbidden` is lexical
  coverage, and the review produced joint rules that used none of its markers — *"Ředitel jedná
  samostatně vždy s prokuristou"*, *"…je nutná též účast prokuristy"* — and parsed **SOLE**. The
  vocabulary is wider now and it is still a list. The KDoc now says that, and says what limits the
  blast radius: since this PR the parser's verdict binds nothing, it is the *suggestion* on the
  operator's form.

Falsified: removing the signature-time check reddens 1 test, counting-only readiness 2, substring
matching 1, the boundary normalisation 1. **101 tests, 0 skipped, 0 failures.**

**I have left auto-merge off.** Finding 1 is the kind of thing a second pair of eyes should see
before this lands, and the review's own verdict was that it should not merge in its previous shape.

Issue #9707, opened 2026-09-11
KYB: find a company by name and town, not only by IČO


JiRaska:
## Problem

Business onboarding opens on an identifier field (IČO). A founder typically does not know their own
IČO — it lives on an invoice, not in anyone's head. Today there is no other way in: no identifier,
no case.

## Proposal

Add name (+ optional town) search over the country register, and let the customer pick the company
from the listed rows. CZ only to start — ARES is the only pack whose register exposes a search
endpoint; the country pack advertises the capability (`supportsNameSearch`) so the UI can hide the
box where it does not exist.

`GET /api/v1/kyb/registry/search?country=CZ&name=…&city=…`

## What this changes about risk

It moves the choice of legal entity from "the customer typed a number we then verified" to "the
customer picked a row". Picking the wrong `STAVBY s.r.o.` out of several hundred identically-named
ones is silent, and surfaces at signing or at the first payment. So a search hit binds nothing: it
carries an identifier, and that identifier still goes through the ordinary lookup + extract
confirmation.

## Measured against the live ARES service (2026-09-11)

The town filter must be `sidlo.textovaAdresa`. `sidlo.nazevObce` is the field name ARES puts in
every search RESPONSE, so it is the one a reader reaches for — and the search endpoint ignores it.
With `obchodniJmeno=Asseco`, six matches in total:

```
sidlo.textovaAdresa  Praha 4   Brno 0   Ostrava 0   Bratislava 2    (4 + 2 = 6, discriminates)
sidlo.nazevObce      Praha 6   Brno 6                Bratislava 6    (ignored — every town, 6)
```

A town box wired to `nazevObce` still returns results, so nothing downstream would notice — the
customer sees companies from the wrong end of the country while believing they filtered.

A third outcome exists next to hits and no-hits, and it is the common one for a short query:
`obchodniJmeno=stavby` matches 2 818 subjects and ARES answers `VYSTUP_PRILIS_MNOHO_VYSLEDKU` — an
error document, not a truncated list, and paging does not help because the cap is on the match
count rather than the page. It is not an outage and must not render as one; the customer is told to
add a town or more of the name.

Refs ADR-0284.

PR #9716, opened 2026-09-11
feat(customer-edge): expose business search by name and town to the app

JiRaska:
## Summary

The edge half of company search (#9707). `GET /customer/v1/business/search` proxies kyb-service's
registry search with the edge M2M token, so the app can offer a founder who does not know their own
IČO a way in that is not an identifier field.

## Type of change

- [ ] fix (bug fix)
- [ ] perf (performance)
- [ ] docs
- [ ] chore (build / tooling)
- [ ] breaking change

## Linked issues

Refs #9707, Refs ADR-0284

## Changes

- `GET /customer/v1/business/search?country&name&city&limit` — read-only, creates nothing, like
  `/lookup` and `/schemes`.
- openapi `1.56.0 -> 1.57.0` (additive).
- Three tests on the new route.

## Reviewer notes

**A hit binds nothing.** It carries an identifier, and that identifier still goes through `/lookup`
and the ordinary extract confirmation before a case is started. Picking the wrong
`STAVBY s.r.o.` out of several hundred identically-named ones is otherwise silent and surfaces at
signing.

**The name is encoded, not interpolated.** `Příklad & syn` is a legal company name; unencoded, the
ampersand would append a parameter of the caller's choosing to the upstream query. The test pins the
encoded form (`name=P%C5%99%C3%ADklad+%26+syn`) rather than merely asserting the request was made.

**Parameters are shape-checked before they reach a URL**, the same reasoning the invitation-token
route already documents in this class — a bad `limit` is a 400 here instead of an upstream round
trip, and the floor and caps mirror kyb-service's own so the two cannot disagree about what is too
short to be worth asking a public register.

**Nullable + `requireNotNull`, never a non-null JAX-RS parameter.** JAX-RS injects null for an absent
one and the declared type only decides where the failure lands — every landing is a 500 (gate
`nonnull-jaxrs-param-ratchet`). The test asserts a 400-shaped rejection for each of the six
malformed cases *and* that no upstream call was made.

**Depends on** #9708 for the upstream route. This proxy forwards HTTP and has no compile-time
dependency on kyb-service, so it can land independently — until #9708 merges it will forward to a
route that answers 404, which is also what a register without name search answers, and what the app
already treats as "hide the search box".

## Definition of Done

- [x] No new lint warnings.

No DB, no event, no new dependency. Thin proxy over an existing upstream.

## Security checklist

The route is `@RolesAllowed("ROLE_CUSTOMER")` by the class annotation and identifies the caller from
the token's own `party_id`, like every other route here. It reads a public register and returns only
what that register already publishes.

## Compliance impact

- [ ] PCI DSS
- [ ] DORA
- [ ] GDPR
- [ ] PSD2 / SCA / RTS
- [x] AML / 5AMLD
- [ ] CNB reporting
- [ ] None

KYB is an AML control surface, but this is read-only over a public register and changes no
verification step.

## Rollout / Rollback

- Rollback plan: revert the PR; the route is additive and nothing calls it until the app ships.


JiRaska:
## Independent review ran against this branch — the SSRF question is clean; three fixes in 36b670a

I had a second agent review this adversarially, since the merge hook blocks it on human review for
being a high-risk path. Its headline, which is the thing you care about:

> **I could not construct an SSRF or query-injection input.**

It tried `name=Acme&limit=9999` → `Acme%26limit%3D9999` (one parameter), `name=x#@evil.svc/` →
`%23%40evil.svc%2F` (no fragment, no authority), `name=../../q/dev/v1/` → fully escaped (no
traversal). `country` is additionally pinned to `^[A-Za-z]{2}$` via `Regex.matches`, and `limit` is a
JVM `Integer` that cannot carry a metacharacter.

**One thing it surfaced that I had stated wrongly, and I have corrected the KDoc:** `UpstreamClient.
validatedUri` is *not* what stops this — its host regex constrains the authority and permits any
path or query. Unlike the invitation-token route, which gets structural separation via
`structuredUri`, this route uses plain `get()`. So `enc()` is the single load-bearing control. It is
correct, but it is alone, and the comment now says so instead of claiming every parameter is
"shape-checked rather than merely encoded" — which cannot be true of a company name, since one
legitimately contains `&`, `/` and `#`.

### Three fixes

- **`MAX_TERM = 100` was a cap the upstream does not have.** `RegistrySearchQuery` requires only that
  the name is not blank. So the edge 400'd queries kyb-service would have served, and a founder whose
  company has a long registered name could not search for it at all — Czech cooperative and
  association names run past 100 routinely. Now 500, documented as a bound on what this edge puts in
  a URL rather than a claim about the register. `MIN_TERM` and `MAX_LIMIT` *do* mirror kyb-service
  exactly and keep that wording.
- **The `city` encoding assertion was vacuous.** It used `Praha`, which `URLEncoder` leaves untouched,
  so it passed against raw interpolation — and `city` is the one parameter with no shape regex, so it
  is exactly the one whose encoding deserves pinning. Now `Ústí nad Labem`, which fails unless both
  the diacritics and the spaces are encoded.
- Boundary cases added for both the name and city caps, plus a test that a long real-world company
  name is still searchable.

Falsified: interpolating `city` instead of encoding reddens 1 test, doing the same to `name` reddens
2, putting the cap back at 100 reddens 1.

### Two things the reviewer raised that I deliberately did **not** change

- `MIN_TERM = 3` matches kyb-service's floor numerically, but the two sides act differently: kyb
  answers an empty 200, the edge answers 400. That is a UX judgement, it is documented in the
  openapi as `minLength: 3`, and changing it is your call rather than mine.
- The openapi omits `default: 20` for `limit` and documents `503` but not the `502` that
  `UpstreamClient` produces. Doc nits; the sibling `/business/schemes` documents only `200`, so this
  is no worse than its neighbours.

### On the red check

`gates (registry-kotlin-data)` — the shard artifact shows 85 gates ran and one failed,
`adr-registry-integrity-check`: stale `docs/adr/CURRENT.md` on `main`, reproducible on a clean
`origin/main` worktree, fixed by #9719. Not this PR.

PR #9717, opened 2026-09-11
feat(admin-ui): business onboarding console — confirm how a company is represented

JiRaska:
## Summary

The operator console for business onboarding (#9711). A heuristic reads the register's *způsob
jednání*, but no agreement is bound by it: an operator confirms how each entity is represented, once
per IČO **and rule wording**, and that confirmation is what the onboarding case uses.

## Type of change

- [ ] fix (bug fix)
- [ ] perf (performance)
- [ ] docs
- [ ] chore (build / tooling)
- [ ] breaking change

## Linked issues

Refs #9711, Refs #9709, Refs ADR-0284

## Changes

- `/business-onboarding` — the manual-review queue, and per company: the register's own wording, the
  parser's reading labelled as a suggestion, the confirmation form, and the confirmation history.
- `kyb-service` registered in all three admin-ui service registries.
- `business-onboarding:view` / `:attest` permissions, route mapping, sidebar entry.
- 10 tests: two guards (RBAC, a11y + single-flight) and six behavioural.

## Reviewer notes — three things it does deliberately rather than the simpler thing

**It posts the hash of the text it rendered.** If the register moved while the form was open,
kyb-service answers 409 and the page says *the register text changed, re-read it* — not
"confirmation failed". Confirming otherwise records a confirmation of a rule nobody looked at.

**A `SUPERSEDED` decision is an alert, not a blank form.** It quotes the previous confirmation, who
made it, and what the rule used to say. A blank form for a company the reviewer recognises gets
confirmed from memory — which is the failure this whole control exists to prevent, arriving through
the control itself.

**Typing any office into the roles field warns** that the signatories go to manual review and that
each office needs a *different* person. A count of two is met by two ordinary board members and does
not meet a chair-plus-member rule.

## Every guarantee was falsified by removing it

| remove from the page | tests that go red |
|---|---|
| the rule-text hash from the POST body | 1 |
| the 409 branch (renders as a generic failure) | 1 |
| the `SUPERSEDED` previous-confirmation alert | 1 |
| the signature-count floor | 1 |
| the role warning | 1 |

One of these caught a bad test while I was writing it: the first version of the "shows the parser as
a suggestion" case waited only for the company name, which renders from the queue response — the
representation panel is a *second* fetch, so the assertion ran before it landed and would have
passed against a page that never rendered the rule at all. It now waits for the panel.

## Service registration

`kyb-service` had no entry in any of the three hand-kept registries, so the BFF would have answered
`Unknown service` in-cluster while working locally — the exact failure `service-registry.guard`
exists for (it is green, 14/14, and checks keys against the deployed workload names in
`openbank-infra/gitops`, where the Service is `kyb-service` on port 8157).

## RBAC

Mirrors kyb-service's own `@RolesAllowed` on the representation routes: `COMPLIANCE` may read the
queue and may **not** confirm a signing rule, so the console does not render a button that would 403
on click. Nav/route gating only — the BFF relays the operator's own bearer and kyb-service + OPA
decide the real answer.

## Definition of Done

- [x] Hexagonal layering preserved (N/A — UI).
- [x] `npm test` passes.
- [x] No new lint warnings.

**Full suite: 550 files, 2936 passed, 1 skipped.** No new dependency.

On lint, precisely: eslint reports two `react-hooks/set-state-in-effect` **warnings** (not errors)
on the page's load-on-mount effect. That is the same shape `identity-cases/page.tsx` already
produces for the identical pattern — I matched the house convention rather than diverge from it, and
am naming the two instances rather than claiming zero.

**Depends on** #9710 for the `/representation` routes and #9708's queue fields. Until those land the
page renders the standard `DataUnavailable` state for a service that answers 404, which is the same
thing it shows for a backend that is not deployed.

## Security checklist

The page renders representatives' names from the public register — the same data the register
publishes and the case API already returns. No new PII path.

## Compliance impact

- [ ] PCI DSS
- [ ] DORA
- [ ] GDPR
- [ ] PSD2 / SCA / RTS
- [x] AML / 5AMLD
- [ ] CNB reporting
- [ ] None

This is the console for the control that decides how many — and which — signatures a business
agreement needs. Every confirmation is attributed and kept; superseded rows are never deleted.

## Rollout / Rollback

- Rollback plan: revert the PR. The page is additive; no other screen links to it except the sidebar
  entry it adds.

## Screenshots / demo

Rendered states are pinned by the behavioural tests rather than by screenshots: unattested with the
parser suggestion, the 409 "register text changed" alert, the superseded alert quoting the previous
confirmation, and the role warning.


JiRaska:
## Independent review ran against this branch — six findings, all fixed in a6ba351

I had a second agent review this PR adversarially, since it is blocked on human review anyway.
It found six real problems. Every fix is falsified (reverting it reddens exactly one test).

**The one worth your attention is #2.** The other five are correctness and honesty.

1. **`business-onboarding:view` granted COMPLIANCE a page where every request 403s.** COMPLIANCE is
   absent from kyb-service's `@RolesAllowed` on the representation routes, from `requireOperator()`
   on `GET /cases`, from `rules.yaml`'s `role_action_matrix` for every `kyb.*` action, and from
   `kyb_rest_ext.rego`'s `operator-kyb-review` reason. All four agree; the console disagreed with all
   four. Worse, my own guard test asserted the mistake with a comment claiming compliance *"may READ
   the queue"*. Both corrected — it now gets neither permission.

2. **Re-confirming an ATTESTED entity silently reverted to the parser's reading.** For an `ATTESTED`
   decision the server sends the *parser's* numbers in `parserSuggests*` (`parsedSigners`, and
   `parserSuggestsRoles` is always empty there), and the form seeded from those. So an entity a human
   had confirmed as *2 signatures, `predseda` + `clen`* opened the re-confirm form pre-filled with
   **1 and no offices** — and one click on a button labelled "Re-confirm" superseded a
   role-constrained joint rule with a bare count of one. Silently, no diff shown. That is precisely
   the mistake human attestation exists to prevent, arriving through the confirmation UI itself.
   Now seeded from the standing attestation.

3. **The form was never re-keyed.** It seeds its fields once while `ruleTextHash` is read from the
   prop at submit time, so a refresh bringing back different wording left the operator's typed
   numbers beside a hash of text they had not read — and the 409 guard could not fire, because that
   hash was fresh. `key={decision.ruleTextHash}` couples them.

4. **A 409 was a dead end.** The message said "reload it and read the new wording" and offered no way
   to; `load` is keyed on scheme+identifier, so nothing re-ran and every further click re-posted a
   hash that could never match again.

5. **The queue was truncated at the service default of 20** while the label asserted that number was
   the whole queue, and the search box filtered only that page.

6. A failed history fetch left the button looking inert.

### On the two red checks, neither of which is this PR

- **`gates (registry-kotlin-data)`** — the shard's own `gate-results-*.json` artifact says 85 gates
  ran and exactly one failed: `adr-registry-integrity-check`. That is stale `docs/adr/CURRENT.md` on
  `main`, reproducible on a clean `origin/main` worktree, and #9719 fixes it.
- **`Admin UI build`** — the Playwright step. Four specs fail: `card-detail-recovery`,
  `sanctions-list-change-review`, and `sanctions-theme` ×2. I ran those same four against a clean
  `origin/main` checkout with its own `npm ci`: **identical four failures**. Pre-existing, and none
  touches business onboarding. (Three more failed only under parallel load — the dev server throws
  `ECONNRESET` — and pass when run with `--workers=1`.)

JiRaska:
## All three red checks on this PR are one pre-existing cause — filed as #9736

They look like three independent failures and they are not:

- `Admin UI build` — the `E2E tests (Playwright)` step.
- `Admin UI` — an aggregator that reads `needs.ui-build.result`.
- `Validate manifests` — its `Verify Admin UI gate (build + golden-path E2E)` step. This one is a
  **required** context, and its name gives no hint that the cause is an e2e suite rather than a
  manifest, which is why it is worth stating explicitly.

The root cause is four Playwright specs that fail on a clean `origin/main` checkout with its own
`npm ci` — `card-detail-recovery`, `sanctions-list-change-review`, and `sanctions-theme` ×2. None
of them touches business onboarding. I measured them on `main` precisely so this PR would not be
blamed for them.

They are not a shared trivial cause (element-not-found, unchecked checkbox, across three unrelated
features), so triaging test-stale-versus-screen-broken is real work in three areas I have no
business guessing at overnight — two of them compliance surfaces. Hence the issue rather than a
speculative fix.

Worth knowing from #9736: the job is `if: needs.changes-ui.outputs.changed == 'true'` and PR-scoped,
so `main` never runs it. The head of `main` carries 57 check-runs and none is `Admin UI`. A
regression can therefore sit there unobserved until the next admin-ui PR inherits it — which is
what happened here.

On this PR's own merits: `Lint`, `Type-check` and `Test (unit + integration)` all pass, and the full
local suite is 550 files / 2940 passed / 1 skipped.

PR #9718, opened 2026-09-11
feat(analytics-sink): fix party attribution for TRANSACTION, and give it a health signal

JiRaska:
## Summary

#2891 asks for `partyId` denormalised onto every ACCOUNT event. **This does not do that, and the reason is the point of the change.** Attribution through the account map already works; what was broken is that TRANSACTION attribution has never worked at all, and that the mechanism had no health signal to say so.

## Type of change

- [x] fix (bug fix) — TRANSACTION attribution has resolved zero rows since V12
- [ ] perf (performance)
- [ ] docs
- [ ] chore (build / tooling)
- [ ] breaking change

## Linked issues

Refs #8792, #2891

## Why not the producer change

1. **ADR-0022 twice states the value** that a later change be adoptable *"without changing the producer contract, because the bronze layer is already an append-only log"*. Bronze takes what producers emit; relationships resolve in silver (V5, V12, V16). Denormalising a foreign aggregate's field into every event inverts that.
2. **Ownership is a property of the ACCOUNT aggregate, not of a balance change.** Copied into each event it freezes a point-in-time owner — right at emission, wrong for any query meaning *"who owns this now"*, with nothing afterwards to separate the two readings.
3. **It would not have fixed the actual defect.** Attribution through the account map works: `AccountStatusChanged` resolves to **9 distinct parties** while carrying no `partyId` at all.

## What the actual defects were

**A. The account map is silently incomplete.** Of the 29 accounts appearing in ACCOUNT events, 19 are in `silver_party_accounts`; the other 10 carry **334 unattributed events**. The remedy is #8905's `INITIAL_LOAD` backfill — it exists, it is correctly wired (plain `@Alternative @Priority(1)`, no build-property gate; `NoOpBackfillSource` **fails closed**; the REST client is configured in both `application.yaml` and gitops). It needs **running**, not more code. That gap is why I built the coverage view.

**B. TRANSACTION attribution has never worked** — in V12, or in V16 which copied its predicate. Both claim `upper(aggregate_type) IN ('ACCOUNT', 'TRANSACTION')` and resolve **zero** transaction rows, because *both halves* are wrong for that type:

| probe | result |
|---|---|
| TRANSACTION events | 125 |
| carrying a `partyId` payload key | **0** |
| carrying `initiatedByPartyId` | **51** |
| `aggregate_id` matching a known account | **0** — it is a *transaction* id |
| `sourceAccountId` matching a known account | **9** |

A clause that reads as coverage and resolves nothing is worse than an absent one: **every party-scoped count over transactions has been silently zero since V12 shipped.** V16 is mine, and I introduced its half by copying the predicate without measuring it — the coverage view found that within a minute of existing, which is the argument for having it.

## Measured

On a **scratch database** reading the live source views — never on `openbank_analytics`, and verified afterwards that 0 objects of mine exist there and that neither live view carries this change:

| | before | after |
|---|---|---|
| TRANSACTION events attributed | **0** of 125 | **52** |
| TRANSACTION aggregates attributed | 0 of 64 | 52 |
| ACCOUNT (unchanged by design) | 103 of 437 | 103 of 437 |

## Disjointness is arithmetic, not assertion

Independently counted per view:

```
silver_party_events     leg1  56 + leg2 18 =  74 = view total
silver_party_activity   leg1 169 + leg2 85 = 254 = view total
```

with **0 empty `party_id`** in either.

**The new guard is load-bearing and falsified.** `initiatedByPartyId = ''` on leg 2: **8 rows** carry both that field and an owned `sourceAccountId`, and removing the guard takes the activity view **254 → 262**, restoring to 254. V12 guarded only `partyId`, which was harmless *solely* because its transaction leg matched nothing — with the join fixed, omitting it would double every attributed transaction, and a doubled count reads as a busier customer.

## Why all three objects are in one file

The coverage view uses **the same predicate** as the attribution views. That is load-bearing rather than tidy: one computing its own notion of "known owner" would report a gap the views do not have, or hide one they do, and the number would stop describing what it names. If the attribution key changes, this changes with it.

Scope is ACCOUNT + TRANSACTION, the two types attributed through the map. Widening it to every type would count PARTY and CONSENT events — which carry their own `partyId` by construction — diluting the ratio toward 100% while the accounts stream stayed broken. That is the failure mode of every aggregate metric this repo has been bitten by.

## Gates, with the parity one falsified both ways

| gate | result |
|---|---|
| `clickhouse-ddl-configmap-parity` | rc=0 |
| same, ConfigMap entry removed | **rc=1**, naming **both** the missing object **and** `silver_party_activity` *"defined differently … a fresh warehouse would build an older version"* — so the redefinitions are genuinely covered, not just the new object |
| `clickhouse-migration-configmap-sync` | rc=0 (17 migrations) |
| `aggregate-type-case-fold` | rc=0 (2931 files) |
| `flyway-version-commit-order` | rc=0 |

**A note for anyone reproducing the sync gate:** it exits **0 without `--enforce`** and **1 with it**; `gates.yaml` passes the flag and sets `mode: enforced`, so it is properly enforced. My first run omitted the flag and briefly looked like a blind spot in the gate — it was my probe's fault, and I am recording it so the next person does not repeat the mis-read.

## What merging does

Per ADR-0022 analytics-sink runs **no migration runner**, so merging applies nothing to a running warehouse. An operator applies `V17`; the ConfigMap copy is what a rebuilt cluster gets. That is why both had to change together.

**Note that this redefines two existing views** (`silver_party_events` from V12, `silver_party_activity` from V16). `gold_party_event_profile` and `gold_party_domain_activity` read the first, so their outputs change — they gain the transaction rows that were always missing. That is the fix, not a side effect.

## Definition of Done

- [x] Hexagonal layering preserved (domain has zero framework imports). — SQL + ConfigMap only
- [ ] Unit tests added/updated. — no code path; verification is the measured queries above
- [x] `./gradlew detekt ktlintCheck koverVerify build` passes. — no Kotlin touched
- [x] No new lint warnings.
- [x] Flyway migration added + rollback note (if DB changed). — ClickHouse DDL, not Flyway. Rollback is re-applying V12 and V16's definitions of the two silver views plus `DROP VIEW gold_party_attribution_coverage`; that restores the zero-transaction behaviour, which is worth stating plainly since it is a regression rather than a neutral revert.
- [x] Docs updated (README, ADR if architectural). — the reasoning, both defects and the remedy for A are in the migration header

## Security checklist

- [x] No new `@SuppressWarnings`, `as Any`, `@Suppress("...")` without justification. — none
- [x] Auth / crypto / payment code changed? → no. Read-side projections; no money moves through them.
- [x] PII path changed? → **more events are now keyed by data subject, which is the intent.** No new field is read and nothing is unmasked: `initiatedByPartyId` and `sourceAccountId` are already in the `PayloadMasker`-masked payload, and a view holds no rows of its own, so retention and erasure are unchanged.

## Compliance impact

- [ ] PCI DSS
- [ ] DORA
- [x] GDPR — it widens correct attribution of already-ingested pseudonymous data to a data subject. Flagged rather than waved through because re-keying more events by subject deserves a reviewer's eye even when the underlying rows do not change.
- [ ] PSD2 / SCA / RTS
- [ ] AML / 5AMLD
- [x] CNB reporting — `gold_party_event_profile` and `gold_party_domain_activity` gain transaction rows they always should have had. Any report built on their previous output was undercounting; flagging so nobody treats a step change as a new defect.
- [ ] None


JiRaska:
## Applied live, and the numbers the views actually produce

V16 and V17 are now applied to the sandbox warehouse (`clickhouse-0`, namespace `analytics`) — the
operator step this repo needs because analytics-sink runs **no** migration runner and the init
ConfigMap only executes on a cluster's first boot (ADR-0022). Both files piped in with
`--multiquery`, rc=0; the previous `silver_party_events` definition was captured with
`SHOW CREATE VIEW` beforehand. Objects present afterwards: `silver_party_activity`,
`gold_party_attribution_coverage`, `silver_party_events` (redefined).

`gold_party_attribution_coverage`, live:

| aggregate_type | events | aggregates | events_attributed | aggregates_attributed | events_unattributed | aggregates_unattributed |
|---|---|---|---|---|---|---|
| ACCOUNT | 437 | 29 | 103 | 19 | 334 | 10 |
| TRANSACTION | 125 | 64 | 52 | 52 | 73 | 12 |

TRANSACTION **0 → 52** on the live warehouse, matching the scratch-database measurement in the PR
body exactly. The 10 unattributed ACCOUNT aggregates (334 events) are the gap the view was built to
make visible: their `AccountCreated` never reached the warehouse, so `silver_party_accounts` has no
owner for them. The remedy is #8905's `INITIAL_LOAD` backfill — an operator request, not code.

## One thing worth stating so a zero is not read as a regression

`silver_party_activity` (history) resolves **52** TRANSACTION rows. `silver_party_events`
(current state) resolves **0**, and that is correct rather than a half-applied fix:

```
TRANSACTION rows in silver_current_state:            64
  ... carrying initiatedByPartyId:                    0
  ... carrying sourceAccountId:                       3
```

`silver_current_state` keeps only the LATEST event per aggregate, and for a transaction that is
`TransactionCompleted`, which carries neither a party nor a source account (it carries
`aggregateId`, `referenceNumber`, `occurredAt` — see V13's note). So the attribution key has nothing
to read there, by construction. The party-bearing transaction events are `TransactionInitiated`, and
they are only reachable through history — which is exactly the basis argument V16 exists for, now
confirmed from the other direction.

Practical consequence for callers: **ask `silver_party_activity` "what has this party done"**, never
`silver_party_events`. The latter answers "what is the party's current state" and for transactions
that state is a completion record with no party on it.

Issue #9736, opened 2026-09-12
admin-ui e2e is red on main, and the gate that would say so never runs there


JiRaska:
## The admin-ui e2e suite is red on `main`, and the gate that would say so never runs there

Four Playwright specs fail on a clean `origin/main` checkout. Because the job that runs them is
**PR-scoped and path-scoped**, nothing on `main` ever executes it — so the breakage is invisible
until it blocks somebody's unrelated PR, which is how it was found.

## Measured

Clean worktree at `origin/main`, its own `npm ci`, `npx playwright test --workers=1`:

```
✘ e2e/card-detail-recovery.spec.ts:27       keeps the masked card detail visible after a failed refresh
✘ e2e/sanctions-list-change-review.spec.ts:134  prunes an ambiguously applied disable from the reconciled manual-screening scope
✘ e2e/sanctions-theme.spec.ts:41            light theme preserves compliance evidence and WCAG A/AA
✘ e2e/sanctions-theme.spec.ts:41            dark theme preserves compliance evidence and WCAG A/AA
```

Same four, identically, on a feature branch that touches none of those areas — which is what
establishes they are pre-existing rather than caused by the PR that surfaced them.

They are not a shared trivial cause. Sample assertions that fail:

```
expect(locator).toBeVisible() failed — element(s) not found
  page.getByRole('heading', { name: '411111******4242' })

expect(locator).toBeChecked() failed
  page.getByRole('checkbox', { name: OTHER_SANCTIONS_LIST.displayName })
```

Element-not-found and unchecked-checkbox across three unrelated features. Someone has to decide,
per spec, whether the TEST is stale or the SCREEN is broken — and if it is the screen, two of these
are compliance surfaces.

**Three further specs fail only under parallel load** (`approvals-decision`, `docs-hub-theme`,
`term-deposit-account-opening`) and pass at `--workers=1`. The dev server logs `ECONNRESET` during
those runs, so they look like contention rather than defects — worth knowing so they are not
conflated with the four above.

## Why `main` never notices

`.github/workflows/ci.yml`:

```yaml
  ui-build:
    needs: changes-ui
    if: needs.changes-ui.outputs.changed == 'true'
```

The job is gated on an admin-ui path change, and the check only appears on pull requests. The head
of `main` carries 57 check-runs and **none** of them is `Admin UI`, `Admin UI build`, or the
`Verify Admin UI gate` step inside `Validate manifests`. So a regression merged through any route
[...]

PR #9737, opened 2026-09-12
test(admin-ui): give the card e2e fixtures RFC-4122 identifiers

JiRaska:
Refs #9736 — one of the four red specs, diagnosed and fixed. **The screen is right; the fixture is wrong.**

## What was happening

#9552 gave `parseCard` strict RFC-4122 validation:

```js
const UUID = /^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i
```

The spec's placeholders are repeated digits — `11111111-1111-1111-1111-111111111111` — so the
**variant nibble is `1`** where the grammar requires 8, 9, a or b. `parseCard` throws `invalid id`,
`useServiceResource` catches it in the same `catch` it uses for network failures, retries three
times, and renders the service-unavailable state.

So the assertion fails with `element(s) not found`, and nothing anywhere says an identifier was
rejected.

## How it was diagnosed

By instrumenting, not by reading. A probe spec logged the routes and responses:

```
ROUTE MATCHED http://localhost:3001/api/svc/card-issuance-service/api/v1/cards/1111…
RESP 200 …   (three times — once per wake retry)
```

while the page rendered *"Card-issuance-service is not responding — the service is deployed but did
not answer (timeout or upstream error)"*. A mock that matches and answers 200, against a page that
reports the service unreachable, is what pointed at the parser instead of at the route pattern.

Confirmed directly:

```
parseCard(<the spec's exact fixture>)  ->  Error: invalid id
```

## Verification, both directions

```
npx playwright test e2e/card-detail-recovery.spec.ts --workers=1
  1 passed (8.6s)

# put ONE repeated-digit id back:
  1 failed
```

So the fixture change is load-bearing rather than incidental. `card-transition-dialog-a11y.spec.ts`
and `cards-portfolio-recovery.spec.ts` pass untouched (4 passed), so nothing else depended on the
old values.

A comment above the constants records why the shape matters — the natural instinct is to tidy them
back into repeated digits.

## Not in this PR

The other three failures in #9736 (`sanctions-theme` ×2, `sanctions-list-change-review`) have a
different cause: `getByRole('checkbox', { name: 'European PEP List' })` renders **unchecked**, and
the initial selection comes from `loadLists`'s `listScopeInitialisedRef` branch in
`src/app/sanctions/page.tsx`. That is a genuine screen-versus-test question on a compliance surface,
and I am not guessing at it — see my comment on #9736 for what I measured.

**And a separate, larger finding that came out of this** — filed separately, not fixed here: that
same `[1-5]` version class **rejects UUIDv7**, and `Ids.newId()` in `openbank-libs-domain` returns
v7 as "the default for durable, indexed identifiers". Six admin-ui parsers carry the pattern.

### DIFF
--- a/openbank-admin-ui/e2e/card-detail-recovery.spec.ts
+++ b/openbank-admin-ui/e2e/card-detail-recovery.spec.ts
@@ -1,11 +1,14 @@
 import { expect, test } from '@playwright/test'
 import { signInAsOperator } from './helpers/auth'
 
-const cardId = '11111111-1111-1111-1111-111111111111'
+// RFC-4122 shape is load-bearing: parseCard (clientContract.ts) validates the version and
+// variant nibbles, so a repeated-digit placeholder is rejected and the page renders its
+// service-unavailable state instead of the card (#9736).
+const cardId = '11111111-1111-4111-8111-111111111111'
 const card = {
   id: cardId,
-  partyId: '22222222-2222-2222-2222-222222222222',
-  accountId: '33333333-3333-3333-3333-333333333333',
+  partyId: '22222222-2222-4222-8222-222222222222',
+  accountId: '33333333-3333-4333-8333-333333333333',
   productCode: 'DEBIT-CLASSIC',
   cardType: 'DEBIT',
   network: 'VISA',
--- a/openbank-admin-ui/src/app/sanctions/page.tsx
+++ b/openbank-admin-ui/src/app/sanctions/page.tsx
@@ -995,7 +995,13 @@ export default function SanctionsPage() {
                                 {isPep && <span style={{ fontSize: '9px', fontWeight: 700, color: 'var(--accent-text)', background: 'var(--accent-bg)', border: '1px solid var(--accent-border)', padding: '1px 4px', borderRadius: '3px' }}>PEP</span>}
                                 {!lst.enabled && <span style={{ fontSize: '9px', fontWeight: 700, color: 'var(--text-tertiary)', background: 'var(--surface-4)', padding: '1px 4px', borderRadius: '3px' }}>{t('vyp.', 'off')}</span>}
                               </div>
-                              <div style={{ fontSize: '10px', color: 'var(--text-tertiary)', fontFamily: 'var(--font-mono)', marginTop: '2px' }}>
+                              {/* A checked row paints --accent-bg/--info-bg behind this line, and
+                                  --text-tertiary (#64748b) on --accent-bg (#eef2ff) is 4.26:1 —
+                                  below the 4.5:1 AA floor the suite's axe assertion enforces.
+                                  --text-secondary is 6.78:1 on the same surface, so a selected row
+                                  steps up one token rather than the palette changing fleet-wide
+                                  (#9749). Unchecked rows keep the tertiary tone on --surface-2. */}
+                              <div style={{ fontSize: '10px', color: checked ? 'var(--text-secondary)' : 'var(--text-tertiary)', fontFamily: 'var(--font-mono)', marginTop: '2px' }}>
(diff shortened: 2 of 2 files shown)


JiRaska:
Second commit pushed to this PR — **the sanctions scope defect, which turned out not to be a
judgement call after all.**

I said earlier I would not decide screen-versus-test on the sanctions specs. Having instrumented it,
there is nothing to decide: the screen has a plain React bug.

`loadLists` decided first-load-versus-reconcile **inside** the `setSelectedListTypes` updater, and
mutated `listScopeInitialisedRef.current` there:

```js
setSelectedListTypes(current => {
  if (!listScopeInitialisedRef.current) {
    listScopeInitialisedRef.current = true
    return nextLists.filter(l => l.enabled).map(l => l.listType)
  }
  return retainEnabledSelectedListTypes(current, nextLists)
})
```

React invokes an updater more than once — StrictMode does it deliberately to surface impure
updaters — so the second invocation sees the ref already `true`, takes the reconcile branch against
an empty `current`, and discards the initial selection. On first load **no sanctions list is
selected**, which on a screening console decides what the operator's search actually runs against.

The decision now happens before the updater; the updater is pure.

Measured on one worker and one server:

```
before     sanctions-theme: 2 failed
after      sanctions-theme: 1 failed (light), 1 passed (dark)
reverted   sanctions-theme: 2 failed again
after      sanctions-list-change-review: 4 passed
after      sanctions-list-change-dialog.test.tsx: 7 passed
```

So three of the four specs in #9736 are green with this branch.

## The light-theme test still fails, and it is a different, real defect

Not the checkbox — an axe `color-contrast` violation:

```
fgColor #64748b   bgColor #eef2ff   contrastRatio 4.25   (AA needs 4.5:1)
= --text-tertiary on --accent-bg  (globals.css:37, :97)
```

**That pair only renders when a PEP row is CHECKED.** So the scope bug was hiding it: fixing the
selection is what made the contrast failure reachable at all. One defect masking another is worth
naming, because it also means the light-theme a11y assertion has never actually been exercised on
that state.

I have not touched it. The remedy is a token decision — darken `--text-tertiary`, lighten
`--accent-bg`, or stop using the tertiary token on an accent surface — and picking one is a design
call on a compliance surface, not a test fix.

JiRaska:
Third commit — **all four specs from #9736 now pass together.**

```
card-detail-recovery + sanctions-theme (light, dark) + sanctions-list-change-review
  -> 4 passed (18.0s)
```

The last one was the contrast violation I said I would leave alone. I changed my mind for a reason
worth stating: the required `Validate manifests` context reads `needs.ui.result`, i.e. the **whole**
Playwright suite, so this one assertion blocks every admin-ui PR in the repo — including the two in
this branch and everyone else's.

What I did is narrower than the palette decision in #9749. A **checked** row steps up one existing
token, `--text-secondary`, which measures 6.78:1 on `--accent-bg` against the 4.26:1 of
`--text-tertiary`. Unchecked rows keep the tertiary tone on `--surface-2`, where it was always fine.
No colour is invented and no other surface moves.

The fleet-wide question — whether `--text-tertiary` should simply be darkened, which would clear
this and every sibling surface at once — stays in #9749 with its measurement. This change does not
pre-empt it; if that lands, the conditional here becomes redundant and can go.

## A verification mistake worth recording

My first negative case *passed*, which would have meant the fix did nothing. It was the probe: a
`sed` revert silently failed to match the conditional expression, so the "reverted" run re-ran the
patched file. Re-done with the file content asserted on both sides:

```
reverted  (grep count 0) -> light theme: 1 failed
restored  (grep count 1) -> light + dark: 2 passed
```

Same class as the two sabotage misses earlier in this branch's history — a negative case is itself a
probe, and a probe that does not do what you think reports clean.

JiRaska:
Superseded by #9751 — same fixture fix, plus the other two failing specs (sanctions-theme x2, sanctions-list-change-review pruning case), so #9751 clears the whole #9736 gate on its own.

PR #9763, opened 2026-09-12
feat(ci): add the four probe shapes that produced false-clean answers tonight

JiRaska:
## Four more shapes for `probe.sh`, all measured in one session

`probe.sh` exists because **a probe fails by reporting CLEAN**. Four more shapes did exactly that
overnight, none of them covered by the library, all of them mechanical rather than a matter of care.

| probe | the false-clean it produces |
|---|---|
| `probe_each` | `for x in $LIST` does **not** word-split in zsh — the loop runs ONCE with the whole string. A merge watcher reported `merged=0/29` for forty minutes; the truth was 20/29. |
| `probe_edit_applied` | an in-place edit matching nothing is silent and exits 0. As a test's **negative case** it re-runs the FIXED code, the suite passes, and the reading is "this fix does nothing". |
| `probe_falsify` | a sabotage of the wrong **shape** leaves the subject untouched, so the gate looks inert. |
| `probe_gh_json` | `gh … --json` answers 502/504 with the error on stderr and **empty stdout**, so `\| jq length` reads 0 and a full queue reports as no PRs. |

## The one worth arguing for

`probe_falsify` is the interesting one, because it guards the repo's own remedy. "Verify by effect,
write the negative case first" is the rule — and **the sabotage is itself a probe**. Twice in one
session mine missed:

- `java.time.Instant.EPOCH` appended to a tracked file, against
  `check-no-epoch-instant-default.py` whose regex is `:\s*Instant\s*=\s*Instant\.EPOCH\b` — the
  import form. Gate stayed green. Reads exactly like a dead gate.
- a bare `private const val SOURCE_SERVICE = "…"`, against
  `check-source-service-convention.py`, which matches **emission sites**, not constant
  declarations. Gate stayed green again.

Both gates were fine. Both sabotages were wrong. Without a control, the conclusion available from
that evidence is "this gate does nothing" — and the write-up would have been confident and false.

`probe_falsify` runs the check three times — clean, mutated, restored — and demands
**pass / fail / pass**. A mutation that changes nothing cannot satisfy it, and it also refuses when
the check was already failing before the mutation, which is the other way a falsification reads as
success.

## Self-test

Both directions per case, run by the enforced `probe-lib-known-positive` gate:

```
[ok] probe_each yields one line per item (got 3, want 3)
[ok] probe_each collapses padding to one item
[ok] probe_each yields nothing for an empty list
[ok] probe_edit_applied accepts an edit that changes the file
[ok] probe_edit_applied REJECTS an expression that matches nothing
[ok] probe_falsify accepts a mutation the check can detect
[ok] probe_falsify REJECTS a mutation the check cannot see (the sabotage missed)
[ok] probe_falsify refuses when the check already fails before any mutation
[ok] probe_gh_json returns the payload when gh succeeds
[ok] probe_gh_json FAILS (never prints empty) when gh errors
[ok] probe_gh_json FAILS on an EMPTY success — 'no results' and 'could not ask' differ

SUBJECTS=40   (was 29)
```

## One honest limit

The `probe_each` case cannot reproduce the incident: **bash word-splits**, so the zsh behaviour is
unreachable from a bash self-test. What it pins instead is the property the caller needs in either
shell — one line per item — which is what makes the loop correct regardless of who reads the script.
Saying so here rather than letting the case imply more than it proves.


JiRaska:
Merged `origin/main` in — #9766 added `probe_commit_touches_path` to the same file, so both sides
appended to the self-test block and git conflicted.

Resolved as a **union**, not by taking a side. A union merge is exactly where a whole-file
resolution silently deletes the other side's work, so the counts are the evidence rather than the
verdict:

| ref | `probe_*` functions | `_check` assertion sites |
|---|---|---|
| merge-base `7e92be54b` | 8 | 29 |
| `origin/main` (after #9766) | 9 | 36 |
| this branch before the merge | 12 | 40 |
| **merged result** | **13** | **47** |

13 = 8 + 1 + 4 and 47 = 29 + 7 + 11 — both sides' additions survive, nothing was dropped.

`bash .github/scripts/lib/probe.sh --selftest` exits 0 with `SUBJECTS=47`, and
`shellcheck -S warning` is clean.

Still needs a human: the change is under `.github/scripts/lib/`, which agent PRs may not
self-merge. That is the correct final state, not something to work around.

PR #9788, opened 2026-09-12
fix(admin-ui): darken --text-tertiary so it clears AA on every light background

### LINKED ISSUE #9749: Light theme: --text-tertiary on --accent-bg is 4.25:1, below WCAG AA — and a second defect was hiding it
## `--text-tertiary` on `--accent-bg` is 4.25:1 in light theme — below WCAG AA, on the sanctions screening scope

Found while fixing #9736's sanctions specs (#9737). The e2e suite's own axe assertion catches it:

```
id: color-contrast   impact: serious
fgColor #64748b   bgColor #eef2ff   contrastRatio 4.25   expected 4.5:1
```

Those are tokens, not one-off colours:

```
openbank-admin-ui/src/app/globals.css:37   --text-tertiary: #64748b
openbank-admin-ui/src/app/globals.css:97   --accent-bg:     #eef2ff
```

The pair renders on the manual-screening list in `src/app/sanctions/page.tsx`: a row whose list is
**checked** and whose `displayName`/`listType` contains "pep" gets `background: var(--accent-bg)`,
and the metadata line under the name uses `--text-tertiary`.

## Why it has never been caught

The accent background is applied **only when the row is checked**, and until #9737 no row was ever
checked on first load — an impure state updater discarded the initial selection. So one defect was
masking the other, and the light-theme axe assertion has never been exercised against this state.

Dark theme is unaffected: `--text-tertiary` is `#94a3b8` and `--accent-bg` is `#1e1b4b` there
(globals.css:200, :206), which is a light-on-dark pair with plenty of margin.

## Scope is probably wider than this one screen

These are global tokens. Any light-theme surface that puts tertiary text on an accent background has
the same ratio; the sanctions list is simply where an axe assertion happened to be pointed. Worth a
sweep of `--text-tertiary` over `--accent-bg` / `--info-bg` / `--success-bg` / `--warning-bg` before
picking a fix.

## Not fixed here, deliberately

Three remedies, and choosing between them is a design decision rather than a test repair:

1. Darken `--text-tertiary` in light theme until it clears 4.5:1 on every `*-bg` surface — fixes the
   class, changes text colour everywhere.
2. Lighten `--accent-bg` — fixes the class, weakens the "this row is selected" signal.
3. Stop using `--text-tertiary` on accent surfaces; give selected rows their own text token —
   narrowest, but adds a token.

Option 1 is the only one that cannot leave a sibling surface failing, and it is also the one that
touches the most pixels.

Refs #9736, #9737.

JiRaska:
Closes #9749. Refs #9736.

## Wider than the issue said

I filed #9749 on one pair. Measured across the whole light palette, `--text-tertiary` (#64748b)
fails AA on **five of nine** backgrounds:

| background | ratio | |
|---|---:|---|
| surface-4 | **3.86** | ✗ |
| accent-bg | **4.26** | ✗ |
| surface-3 | **4.34** | ✗ |
| danger-bg | **4.35** | ✗ |
| info-bg | **4.37** | ✗ |
| success-bg | 4.52 | ✓ |
| surface-2 | 4.55 | ✓ |
| warning-bg | 4.59 | ✓ |
| surface-1 | 4.76 | ✓ |

`surface-3` and `surface-4` are ordinary console surfaces — this is not a niche accent row.

## The token's own comment explains how it happened

```
/* … #64748b keeps a clear hierarchy while meeting AA on both white (4.76:1)
   and surface-2 (4.55:1). */
```

White and surface-2 are **the two lightest backgrounds in the palette** — the easiest possible
comparison, and the only two anyone measured. The value was verified against the cases it could not
fail.

## The fix

`#586779` is the **lightest** value that clears 4.5:1 on all nine (worst case surface-4 at 4.69:1)
and stays lighter than `--text-secondary`, so the primary/secondary/tertiary hierarchy is unchanged.
Dark theme declares its own value and is untouched.

## The guard fixes the comparison set, not the value

27 pairs — every light text token against every light background token. Choosing the easy comparison
is what let the original through, so that is the part worth making impossible to repeat rather than
this one hex.

It carries two self-checks, because a contrast guard can pass vacuously in two ways:

```ts
// the lookups must resolve to real values
expect(BACKGROUNDS.map(token).every(v => /^#[0-9a-f]{6}$/i.test(v))).toBe(true)
// and the arithmetic is pinned against a KNOWN ratio, not against itself
expect(ratio('#64748b', '#e2e8f0')).toBeCloseTo(3.86, 1)   // the failure this replaces
expect(ratio('#ffffff', '#000000')).toBeCloseTo(21, 0)
```

## Verification, both directions

```
guard on the new token                        29 assertions pass
restore #64748b                               5 failures, naming exactly:
                                                surface-3, surface-4, accent-bg, danger-bg, info-bg
full admin-ui suite (npm test)                549 files, 2957 passed, 1 skipped
axe-asserting theme e2e (sanctions + docs)    4 passed, light and dark
```

## Relationship to #9765

#9765 already stepped the selected sanctions row up to `--text-secondary` for its own surface. With
this token fixed, that conditional becomes redundant — it is not wrong, just no longer load-bearing,
and can be simplified whenever that file is next touched. I have left it, rather than bundling an
unrelated edit into a palette change.


JiRaska:
The `Admin UI build` red here is **not this PR**. It is
`e2e/ui-primitives.spec.ts` → *the dark token theme…*, which fails on unmodified `origin/main`
4 runs out of 5. Measured on the same commit, in the same worktree:

| spec | globals.css | result |
|---|---|---|
| unchanged | `origin/main` (no token change) | 1 passed / 4 failed in 5 runs |

The test flips the theme and runs axe immediately, while `.stat-card`'s `transition: all 0.2s ease`
is still animating, so it measures a background that exists in neither palette (`#5a5f69` is light
`--bg` ~72% of the way to dark `--bg`). Because the crossfade *starts* from the light palette, the
sampled ratio also moves when a LIGHT token changes — which is exactly why this PR looked like it
had broken dark-theme contrast.

Fixed separately in #9809 (test-only). This PR needs no change; it should go green once that lands
or on a lucky re-run.

I got this wrong once on the way: my first check reverted `globals.css`, saw the test pass, and I
took that as "my change caused it". One run of a test that fails 80 % of the time is not evidence
either way — the repeat count is what settled it.

JiRaska:
**Interaction with #9825 — measured, and it decides the merge order.**

#9788 (open, `fix/text-tertiary-clears-aa`) changes `--text-tertiary` from `#64748b` to `#586779`
and adds its own `light-theme-token-contrast.guard.test.ts`. Measured against the same surface list:

| `--text-tertiary` | surfaces below AA |
|---|---|
| `#64748b` (today) | 6 — surface-3, surface-4, accent-bg, danger-bg, info-bg, sidebar-bg |
| `#586779` (#9788) | **1** — sidebar-bg |

So #9788 is the systemic remedy #9749 asked for, and it is the substance; this PR is the coverage.
They are complements, not competitors — but they **cannot merge in either order untouched**:

- If #9788 lands first, ~10 entries in this PR's `BELOW_AA` start passing, and the ratchet's *stale*
  arm fails the build. That is the guard working as designed, not a defect — but it means this PR
  must regenerate its baseline after that merge.
- If this lands first, #9788 turns 10 declared pairs green and hits the same assertion.

**Proposal: merge #9788 first.** It removes real user-facing debt; this PR then drops the stale
entries (a strictly stricter guard) and the two lists shrink to what remains.

**On the duplicate guard.** #9788's new file computes the same ratios, light theme only, from a
hand-listed token set — the shape this PR exists to replace. Whichever lands second, the repo should
end with **one** of them, or they will drift apart and both claim to be the contrast gate. I am not
deleting another author's file from this PR; flagging it so the second merge makes that call
deliberately.

PR #9791, opened 2026-09-12
fix(admin-ui): request the service-supported audit evidence window

### LINKED ISSUE #9428: fix(admin-ui): make audit trail evidence bounded and trustworthy
The Audit Log currently claims to show a full trail but silently requests only 100 newest records and blindly trusts the response shape. Modernize the workflow to request the service-supported evidence window, disclose its bound, reject malformed audit entries, and preserve a verified same-query snapshot on refresh failure without weakening audit RBAC.

JiRaska:
## Problem

`openbank-admin-ui/src/app/audit/page.tsx` requested `?limit=100` and rendered
`Latest N events (100 maximum)`. The service's own bound is five times that:

```kotlin
// openbank-audit-service/.../infrastructure/rest/AuditResource.kt
suspend fun getAuditTrail(
    @PathParam("aggregateId") aggregateId: String,
    @QueryParam("limit") @DefaultValue("100") limit: Int,
): Response = Response.ok(repo.findByAggregateId(aggregateId, limit.coerceIn(1, 500))).build()
```

So 400 retrievable events per aggregate were unreachable from the operator UI,
and the "maximum" disclosed to the operator was the client's hard-coded page
rather than the service's window. A count at the bound also could not be told
apart from a complete trail.

## Change

- `AUDIT_EVIDENCE_WINDOW = 500` declared once in `src/lib/audit/auditEvidence.ts`,
  used by both the request URL and the disclosure string. The two numbers drifting
  apart is what produced the wrong claim, so they now cannot.
- The disclosure splits the two cases: a **full** window discloses truncation
  ("older events may exist and are not shown"); a short response says the service
  returned everything it holds for that aggregate.
- The search hint no longer says "full audit trail".
- `e2e/audit-trail-recovery.spec.ts` route stubs follow the new bound.

## Out of scope, deliberately

Response validation (`parseAuditEvidenceList`) and the failed-refresh snapshot
both landed in #9371 and are untouched here. RBAC is unchanged — same role-gated
endpoint, no new parameters, no paging added (the service exposes `offset`, but
walking past the window is a separate design question, not a bound disclosure).

## Verification

- `src/test/audit-evidence-window.test.tsx` — 4 tests, all four **fail against the
  pre-fix page** (restored `origin/main:page.tsx` and re-ran) and pass after.
- Full admin-ui suite: `549 files, 2934 passed | 1 skipped`.
- `tsc --noEmit` clean; `npm run lint` 0 errors.
- `npx playwright test e2e/audit-trail-recovery.spec.ts` — 2 passed.

Closes #9428


JiRaska:
The `Admin UI build` red here is not this PR's change. It is
`e2e/ui-primitives.spec.ts` → *the dark token theme…*, a pre-existing flake that fails on
unmodified `origin/main` 4 runs out of 5: it flips the theme and scans with axe while
`.stat-card`'s `transition: all 0.2s ease` is still running, so it measures an interpolated
background (`#5a5f69`) that exists in neither palette.

Diagnosis and fix in #9809 / #9808. Nothing to change here.

PR #9793, opened 2026-09-12
fix(ci): attest SLSA provenance on the runner image, via the shared lib

JiRaska:
## Linked issues

Refs #9805 — the governance follow-up for the class.
Refs #9803 — the PolicyException that breaks the deadlock so this fix can be built.

## What

`runner-image.yml` never attested SLSA provenance for the runner image. This adds it, by sourcing the shared `cosign-attest.sh` instead of hand-rolling cosign.

**Rewritten.** The first version of this PR bumped the pinned digest in `arc-runners.tf`. That was the wrong fix — [the new digest is denied too](https://github.com/JiRaska/open-bank-oss/pull/9793#issuecomment-5645270916) — and the tf change is gone. This branch now touches one file, rebased onto current `main`.

## The defect

Every runner Pod in all four pools has been denied at admission since **2026-09-05 18:57**, the commit that added `verify-openbank-image-slsa-provenance` (#8847) scoped to `openbank-*` — which matches `openbank-ci-runner`. The runner image has no SLSA attestation, so it fails that policy, and Kyverno then reports `unverified image` on the two **Enforce** policies as well. The only error it logs is the SLSA one:

```
"error": ".attestations[0].attestors[0].entries[0].keys: attestions not found
          for predicate type https://slsa.dev/provenance/v0.2"          x37 in 3 min
```

The pools were never cleanly dead — `failurePolicy: Ignore` admits a Pod whenever the webhook times out — so CI ran intermittently at a fraction of capacity with nothing to point at. 865 denials/24 h.

## How every other cause was falsified

Measured against the live cluster and registry on 2026-09-12:

| candidate | probe | result |
|---|---|---|
| wrong/rotated key | `cosign public-key --key awskms:///alias/openbank-cosign-signing` vs the pinned PEM | **byte-identical** |
| artifacts gone from ECR | `aws ecr list-images` for both digests | `.sig` **and** `.att` present for both |
| artifacts unverifiable | `cosign verify` + `verify-attestation --type cyclonedx` with the pinned PEM | **both pass** |
| predicate shape | `bomFormat` in the attested predicate | `CycloneDX`, as the policy condition requires |
| Kyverno has no ECR creds | pod env | `AWS_CONTAINER_CREDENTIALS_FULL_URI` injected |
| IAM excludes the repo | `kyverno-image-verify-iam.tf` | `repository/openbank-*` — includes it |
| tag vs digest reference | same service image probed both ways | both admitted |

What survives is a single asymmetry between the image that is admitted and the one that is not:

```
openbank-ledger-service (admitted):  .sig OK   cyclonedx OK   slsaprovenance OK
openbank-ci-runner      (denied):    .sig OK   cyclonedx OK   slsaprovenance ABSENT
```

## The change

`cosign_sign_and_attest` from `openbank-infra/scripts/lib/cosign-attest.sh` does sign + CycloneDX + SLSA in one call, each step proving its own envelope landed with a `verify-attestation` bound to that run rather than trusting the append-only `.att` tag. Not hand-rolling this is the root `CLAUDE.md` rule, and this workflow was the violation.

**`continue-on-error: true` is removed.** Its comment justified it as *"a cosign/trivy hiccup must still never block a runner rebuild"*. That trade was wrong: an unsigned or unattested runner image is not degraded, it is one no Pod can run. The build that produces it has failed and must say so.

The independent verify step now checks **all three** predicates, one per ClusterPolicy. Checking only what the producer produced is precisely how the 2026-09-04 run went green while shipping an undeployable image — its log really does say `signature + SBOM attestation both verified present`, and that was true and insufficient.

## Merging this does not fix the cluster, and there is a bootstrap deadlock

```yaml
runs-on: openbank-deploy
```

The workflow that builds the runner image runs **on a runner pool that this same defect blocks**. Merging changes nothing until a rebuild actually runs, and the rebuild needs an admitted runner Pod.

Three ways out, none of them in this PR:

1. **Wait for a `failurePolicy: Ignore` timeout to admit one Pod.** This is how CI is limping along now — luck, not design.
2. **Runbook `0011-kyverno-admission-rollback.md`** — drop `verify-openbank-image-sbom-attestation` to Audit long enough for one rebuild, then restore. A deliberate, time-boxed weakening of an Enforce control.
3. **A narrow PolicyException for `arc-runners`.** There used to be exactly this (`arc-runner-image-exception`, removed in #908), so the shape is known.

Then: run `runner-image.yml`, take the new digest from its job summary, bump `local.runner_image` in `arc-runners.tf`, and `workflow_dispatch` `platform-tofu.yml` — which is manual-only, so a merge alone never applies it.

## Verification after the rebuild

This probe works at zero scale, which matters because `openbank-deploy` sits at 0 runners and so can never be verified by the absence of denials — a dry-run Pod create asks Kyverno for the same decision without needing a job:

    kubectl apply --dry-run=server -f - <<PROBE
    apiVersion: v1
    kind: Pod
    metadata: {name: kyverno-probe, namespace: arc-runners}
    spec:
      containers: [{name: c, image: <NEW_DIGEST_REF>}]
    PROBE

Expect `pod/kyverno-probe created (server dry run)`. Run it against the **old** digest in the same session as a control — it must still be denied, or the probe is not measuring anything.

### DIFF
--- a/.github/workflows/runner-image.yml
+++ b/.github/workflows/runner-image.yml
@@ -100,57 +100,61 @@ jobs:
           cosign version
 
       # Supply-chain provenance for the runner image itself (ADR-0030 D4 step 4, issue #310).
-      # The kyverno attestation policy (verify-openbank-image-sbom-attestation, Audit) matches
-      # openbank-* — the ci-runner image was the last fleet image with neither a signature nor
-      # a CycloneDX SBOM attestation, so the Audit PolicyReport could never reach 0 fail and the
-      # Enforce graduation stayed blocked. Same KMS key + trivy pattern as auto-deploy.yml.
+      # Sources openbank-infra/scripts/lib/cosign-attest.sh instead of hand-rolling trivy +
+      # cosign, which is the repo-wide rule (root CLAUDE.md) and is what this workflow got
+      # wrong: it signed and attested a CycloneDX SBOM and NEVER attested SLSA provenance.
+      # #8847 then added verify-openbank-image-slsa-provenance scoped to `openbank-*`, which
+      # matches openbank-ci-runner, so from 2026-09-05 every runner image failed that policy
+      # and every runner Pod was denied at admission — Kyverno reports "unverified image" on
+      # the two Enforce policies as well, with the missing SLSA predicate as the only logged
+      # error. Measured 2026-09-12: the passing control image (openbank-ledger-service) has
+      # cyclonedx AND slsaprovenance; openbank-ci-runner had cyclonedx only.
       #
-      # continue-on-error stays: a cosign/trivy hiccup must still never block a runner
-      # rebuild. What changes is the silence — the verify step below makes a real failure
-      # loud (job summary + a dedicated, separately-watchable step) instead of a
-      # single easy-to-miss ::warning buried in this step's log.
-      - name: Sign + attest runner image SBOM (cosign + KMS)
+      # cosign_sign_and_attest does all three — sign, CycloneDX SBOM, SLSA provenance — with
+      # the same KMS key, and each step proves its own envelope landed with a
+      # verify-attestation bound to THIS run rather than trusting an append-only .att tag.
+      #
+      # continue-on-error is GONE, deliberately. It was justified as "a cosign/trivy hiccup
+      # must never block a runner rebuild", and that trade was wrong: an unsigned or
+      # unattested runner image is not a degraded image, it is one no Pod can run. A build
+      # that produces it has failed and must say so.
+      - name: Sign + attest runner image (cosign + KMS, shared lib)
         id: attest
-        timeout-minutes: 10
-        continue-on-error: true
+        timeout-minutes: 15
         run: |
           set -uo pipefail
-          COSIGN_KEY="awskms:///alias/openbank-cosign-signing"
-          echo "==> cosign sign ${IMAGE_DIGEST_REF}"
-          COSIGN_YES=true cosign sign --key "$COSIGN_KEY" "$IMAGE_DIGEST_REF" \
-            || { echo "::warning::cosign sign failed"; exit 1; }
-          sbom="${RUNNER_TEMP}/ci-runner.cdx.json"
-          # --platform is REQUIRED: this image is built linux/arm64 (see the buildx step) and
-          # trivy defaults to linux/amd64 for REMOTE scans regardless of host arch — without it
-          # the scan fails with "no child with platform linux/amd64 in index" and the image is
-          # left unattested. Same bug that left several fleet images unattested (2026-07-12).
-          trivy image --platform linux/arm64 --format cyclonedx --output "$sbom" "$IMAGE_DIGEST_REF" \
-            || { echo "::warning::trivy SBOM generation failed"; exit 1; }
-          echo "==> cosign attest (cyclonedx) ${IMAGE_DIGEST_REF}"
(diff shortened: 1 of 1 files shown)


JiRaska:
## Falsified before merge — do not merge this as it stands

Converted to draft. The premise of this PR is wrong: **the new digest is rejected too.**

Server-side dry-run Pod create in `arc-runners`, against the digest this PR pins:

```
resource Pod/arc-runners/kyverno-probe-new-digest was blocked due to the following policies

verify-openbank-image-sbom-attestation:
  verify-cyclonedx-sbom-attestation: unverified image ...openbank-ci-runner@sha256:c008fc0acd3e...
verify-openbank-image-slsa-provenance:
  verify-slsa-provenance: unverified image ...openbank-ci-runner@sha256:c008fc0acd3e...
```

So re-pinning changes nothing at admission. Found only because the probe was built early to verify the `openbank-deploy` pool, which sits at 0 runners and so can never be verified by the absence of denials.

## What is established

1. **Both** runner digests are rejected today — the current `45c0408d…` and this PR's `c008fc0a…`.
2. Not a Kyverno, key or namespace problem: a current service image (`openbank-ledger-service:sandbox-8db73dbb`) is **admitted** by the same probe in the same namespace.
3. The 2026-09-04 build genuinely verified this digest at the time. Its log prints the result, not just the command:
   ```
   ✅ signature + SBOM attestation both verified present on ...@sha256:c008fc0acd3e...
   ```
   That step has no `continue-on-error` and its verify uses `cosign verify-attestation --key awskms:///alias/openbank-cosign-signing --type cyclonedx`.
4. The public key pinned in the policies has **never changed** in git — introduced in #311, same value today.
5. Both sign paths — `runner-image.yml` and `openbank-infra/scripts/lib/cosign-attest.sh` — use the same alias, `awskms:///alias/openbank-cosign-signing`.

## What that leaves

Something makes `openbank-ci-runner` artifacts unverifiable **against the pinned PEM** while service artifacts verify. Two candidates, not separable from inside the cluster:

- The artifacts are absent in ECR now (present on 2026-09-04, gone since).
- The KMS alias did not resolve to the pinned key at the moment the runner image was signed, so the signature is against a different key. Service images would be unaffected because every deploy re-signs.

Worth noting next door: `openbank-ci-runner` is plain `IMMUTABLE` — the `IMMUTABLE_WITH_EXCLUSION` carve-out for mutable `sha256-*` tags exists only for `openbank-admin-ui` (`ecr-service-repositories.tf`). And `cosign-attest.sh` carries the #8982 branch that *tolerates* a failed attest with a warning. Neither is proven to be the cause here, but both are the right place to look.

## The one command that settles it

```bash
cosign public-key --key awskms:///alias/openbank-cosign-signing
```

If that does not match the PEM in `verify-sbom-attestation-policy.yaml`, the cause is the key and every image signed before the repoint is affected — not just this one. If it does match, the cause is the artifacts, and:

```bash
aws ecr list-images --repository-name openbank-ci-runner --filter tagStatus=TAGGED \
  --query 'imageIds[?starts_with(imageTag, `sha256-c008fc0a`)]'
```

says whether `.sig` / `.att` still exist.

Holding this PR until one of those answers comes back. The comment block added to `arc-runners.tf` here is accurate about the *symptom* and worth keeping whatever the fix turns out to be, but the digest bump itself is not the fix.

JiRaska:
Governance follow-up for the class, not this incident: #9805 — the fleet attestation gate verifies only cyclonedx, cannot see the runner image, and nothing forbids hand-rolled cosign. This PR is the incident fix; that issue is what keeps it from being the third round.

